Issue #28664, opened 2017-08-14
test infrastructure does not chunk an empty file


phil-davis:
### Steps to reproduce
1. try to make an integration/UI test that uses tests/TestHelpers/UploadHelper.php to do a chunking upload of an empty file
I came across this when enabling some firewall test cases that used combinations of chunking versions and file sizes.

### Expected behaviour
Even though it is a weird thing to try, it should work. (and it can work - the chunking protocol itself works OK when sending a single "empty" chunk then MOVEing it into place)

### Actual behaviour
Notice: Undefined variable: result in /home/phil/core/tests/TestHelpers/UploadHelper.php line 142

### Server configuration
Any current master.

phil-davis:
Code merged:
master #28666 
stable10 #28675

PR #28666, opened 2017-08-14
Allowing chunking an empty file

phil-davis:
## Description
If we are trying to split an empty file into "chunks", then there won't be any because feof() is reached as soon as the file is opened.

Purposely make a single empty chunk.

## Related Issue
#28664 

## Motivation and Context
Make an edge case work in the test infrastructure, so that tests that loop around trying [no|v1|v2] chunking will work OK when trying to upload a 0 byte file.

## How Has This Been Tested?
Enable a firewall test with 0 byte file to try all chunking versions.

## Types of changes

## Checklist:


individual-it (approved):

phil-davis:
Backport stable10 #28675

PR #28688, opened 2017-08-15
Remove apostrophe from error text

phil-davis:
## Description
Do not use special characters in the text saying that a file cannot be read.

## Related Issue
https://github.com/owncloud/core/issues/28684
https://github.com/owncloud/firewall/issues/309

## Motivation and Context
This is a pragmatic work-around for this particular case - just get rid of the apostrophe.

## How Has This Been Tested?
Block a folder with a firewall rule, share a public link to it, access the public link and try to download a file in that folder. Verify that the message is displayed OK.

## Screenshots (if appropriate):
cannot

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #28691, opened 2017-08-15
If exception message exists do not bother with cannot be read text

phil-davis:
For the ForbiddenException, if the exception has a message, then display that only rather than bothering to also display 'File cannot be read'.

another way to go???


phil-davis:
Is this preferable???

phil-davis:
PR #28688 chosen
This one is not needed.

PR #28692, opened 2017-08-15
Fix getting drop target when dragging from file manager

### LINKED ISSUE #28689: Drag and drop into subfolder uploads to current folder
### Steps

1. Create a folder "test"
1. Open the file manager (in my case Dolphin or Konqueror on Plasma 5)
1. Drag a file onto "test"

### Expected result
File is uploaded into "test"

### Actual result
File is uploaded inside the current directory.

### Version
ownCloud 10.0.3beta

Regression, this used to work correctly. Could be related to the move to new dav endpoint.

I tried the same with IE11 and the result is different but still bogus: https://github.com/owncloud/core/issues/28582

PVince81:
## Description
Possibly related to jquery.fileupload upgrade, now
e.originalEvent.target might be null.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28689.

## Motivation and Context
See ticket

## How Has This Been Tested?
Manual testing like in the ticket.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/js/filelist.js
+++ b/apps/files/js/filelist.js
@@ -2702,7 +2702,7 @@
 					return false;
 				}
 
-				var dropTarget = $(e.originalEvent.target);
+				var dropTarget = $(e.originalEvent.delegatedEvent.target);
 				// check if dropped inside this container and not another one
 				if (dropTarget.length
 					&& !self.$el.is(dropTarget) // dropped on list directly
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -251,6 +251,7 @@ OC.FileUpload.prototype = {
 			this.data.headers['Authorization'] =
 				'Basic ' + btoa(userName + ':' + (password || ''));
 		}
+		this.data.headers['requesttoken'] = OC.requestToken;
 
 		var chunkFolderPromise;
 		if ($.support.blobSlice
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -1196,11 +1196,15 @@ OC.Uploader.prototype = _.extend({
 				fileupload.on('fileuploaddone', function(e, data) {
 					var upload = self.getUpload(data);
 					upload.done().then(function() {
-						// don't hide if there are more files to process
-						if (!self.isProcessing()) {
-							self._hideProgressBar();
-						}
 						self.trigger('done', e, upload);
+						// defer because sometimes the current upload is still in pending
+						// state but frees itself afterwards
+						_.defer(function() {
+							// don't hide if there are more files to process
+							if (!self.isProcessing()) {
+								self._hideProgressBar();
+							}
+						});
 					}).fail(function(status, response) {
 						var message = response.message;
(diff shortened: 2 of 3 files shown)


apps/files/js/filelist.js
```diff
@@ -2708,7 +2708,7 @@
 					return false;
 				}
 
-				var dropTarget = $(e.originalEvent.target);
+				var dropTarget = $(e.originalEvent.target || e.originalEvent.delegatedEvent.target);
```
jvillafanez: I guess the `e.original.target` won't be needed any longer.

The change in the library that affects this is https://github.com/owncloud/core/blob/639dafdd4db2eb72210b623b1a66c427b42b14d4/apps/files/js/jquery.fileupload.js#L1137 (before the library update) and https://github.com/owncloud/core/blob/master/apps/files/js/jquery.fileupload.js after the library update.
The actual event isn't set as the originalEvent but as the delegateEvent.

Unless we want compatibility between the library versions, which I don't think it's needed, we can remove the old code.

PVince81: Thanks for looking deeper. To be honest I was just too lazy to investigate this detail.

PVince81: @jvillafanez adjusted

SergioBertolinSG:
When uploading several files to a folder there is no uploading bar.

SergioBertolinSG:
When uploading a folder with several folders and files inside them, the files are not uploaded.

PVince81:
stable10: https://github.com/owncloud/notifications/pull/117

jvillafanez:
Wrong link? Backport doesn't seem related.

PVince81:
oh lala, major browser tab fail. I'll fix

jvillafanez:
> When uploading a folder with several folders and files inside them, the files are not uploaded.

Folder upload seems to be be supported only in chrome. It's very likely a browser limitation.

SergioBertolinSG:
Using chrome I mean.

PVince81:
- [ ] BUG: progress bar does not disappear when uploading a folder (chrome) into the current dir

PVince81:
- [ ] BUG: target row size must be refreshed after transfer to have the correct size

I had a look at this and it's more complex, to be fixed separately. This is not related to the jquery.fileupload library update not recent changes. It is likely present since we moved to Webdav for file operations.

PVince81:
@SergioBertolinSG please retest

PVince81:
- [ ] fix js tests

PVince81:
tests fixed

SergioBertolinSG:
The estimation of the upload bar is very general (few seconds, minutes, etc) has it changed in this PR?

SergioBertolinSG:
I guess it didn't so it works fine.

SergioBertolinSG (approved):
Works fine.

PVince81:
@SergioBertolinSG the estimate was never good...

jvillafanez (approved):
Code looks good

PVince81:
stable10: https://github.com/owncloud/core/pull/28882

PR #28696, opened 2017-08-15
Update CHANGELOG.md

karakayasemi:
I have forgotten to change the title of the corresponding PR after reviews. Not preLogin it should be failedLogin.


PVince81 (approved):
:+1:

PR #28701, opened 2017-08-16
Custom wrapper to handle commands in oC

sharidas:
A custom wrapper to handle command transfer-ownership.
In this wrapper we can make changes for functions like
fopen or any other routines.

## Description

A custom wrapper created along with a custom view so that changes required for common functions like fopen arguments or so can be handled here. Primarily the focus is for command transfer-ownership and later we would use for other command like recreate master key.

## Motivation and Context

This change would prevent the modification to common routines which were made to arguments of functions like fopen or copy etc. And use this wrapper for custom changes.

## How Has This Been Tested?

WIP

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/CustomView.php
+++ b/lib/private/Files/CustomView.php
@@ -1,0 +1,289 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace  OC\Files;
+
+use OC\Encryption\CustomEncryptionWrapper;
+use OC\Encryption\Manager;
+use OC\Files\Cache;
+use OC\Files\Filesystem;
+use OC\Files\Storage\Storage;
+use OC\Memcache\ArrayCache;
+use OCP\Files\Mount\IMountPoint;
+use OCP\Lock\ILockingProvider;
+use OC\Files\Mount\MoveableMount;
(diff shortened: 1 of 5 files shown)


lib/private/Files/CustomView.php
```diff
@@ -0,0 +1,2065 @@
...
+use OC\Files\Cache;
+use OC\Files\Filesystem;
+use OC\Files\Storage\Storage;
+use OC\Memcache\ArrayCache;
+use OCP\Files\Mount\IMountPoint;
+use OCP\Lock\ILockingProvider;
+use OC\Files\Mount\MoveableMount;
+use OCP\Lock\LockedException;
+
+class CustomView {
```
PVince81: why ? and if really needed, why not `extend View` ?

sharidas: Modifying to go with ```extend View```

lib/private/Files/CustomView.php
```diff
@@ -0,0 +1,289 @@
...
+
+	/**
+	 * Rename/move a file or folder from the source path to target path.
+	 *
+	 * @param string $path1 source path
+	 * @param string $path2 target path
+	 *
+	 * @return bool|mixed
+	 */
+	public function renameCustom($path1, $path2) {
```
PVince81: mind posting a diff between rename() and renameCustom() ? they look very similar and I don't really see what the difference is and why this is needed

sharidas: Here is the diff between `rename()` and `renameCustom()`

```
sujith@sujith-Inspiron-5567 ~ $ diff rename_function.php customrename_function.php
9c9
<       public function rename($path1, $path2) {
---
>       public function renameCustom($path1, $path2) {
80d79
<                                       // moving a file/folder between storages (from $storage1 to $storage2)
82c81
<                                       $result = $storage2->moveFromStorage($storage1, $internalPath1, $internalPath2);
---
>                                       $result = $storage2->moveFromStorageCustom($storage1, $internalPath1, $internalPath2, true, true);
sujith@sujith-Inspiron-5567 ~ $ 
```
The difference is the call to moveFromStorage in customView's renameCustom() method. Rest of the code is same.

PVince81:
Thanks. As discussed, try getting rid of the `CustomView` as it causes too much code duplication.

Since this custom view is calling a storage method with extra args, try simply writing that method call `moveFromStorage` and `copyFromStorage` directly inside the CustomEncryptionWrapper, and then set the two flags when passing down the call.

In general I'm still not happy that we need to add these extra flags to a "standard" storage method, we'll need to think of a better solution. Maybe there's a way to move the logic that enables itself through these flags, move that logic to the CustomEncryptionWrapper.

PVince81:
@sharidas obsolete ? please close and link to your new PR

sharidas:
Obsolete PR. The PR which implements the design or implementation for the change is https://github.com/owncloud/core/pull/28774

PR #27952, opened 2017-05-19
Remove appstoreenabled and assume true

PVince81:
## Description
Obsoleted by the market.

## Related Issue
Fixes https://github.com/owncloud/core/issues/27951

## How Has This Been Tested?
Not tested yet...

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

## TODO

- [ ] test

PR #28711, opened 2017-08-16
Don't enable market app on upgrade from 9.1.x or lower when app store…

### LINKED ISSUE #118: Don't enable market app when trying an offline upgrade (9.1.4 -> 10.0.2)
### Steps to reproduce
1. Install owncloud 9.1.4
2. Download and prepare 10.0.2
3. Disable marketplace
4. Attempt to upgrade owncloud from 10.0.2

### Expected behaviour
The marketplace app should stay disable during attempted offline upgrade (and show an error when upgrade is not possible to incompatible apps)

### Actual behaviour
During `occ:upgrade` marketplace will be enabled and will attempt to conncect to the marketplace
```
root@dima:/var/www/owncloud# sudo -u www-data php occ app:disable market
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
market disabled
```

```
root@dima:/var/www/owncloud# sudo -u www-data php occ app:list
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
Disabled:
...
  - market
...
```

```
root@dima:/var/www/owncloud# sudo -u www-data php occ upgrade -vvv
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
2017-07-17T12:39:55+00:00 Set log level to debug
2017-07-17T12:39:55+00:00 Repair step: Repair MySQL database engine
2017-07-17T12:39:55+00:00 Repair step: Repair MySQL collation
2017-07-17T12:39:55+00:00 Repair info: All tables already have the correct collation -> nothing to do
2017-07-17T12:39:55+00:00 Repair step: Repair SQLite autoincrement
2017-07-17T12:39:55+00:00 Repair step: Repair duplicate entries in oc_lucene_status
2017-07-17T12:39:55+00:00 Repair info: lucene_status table does not exist -> nothing to do
2017-07-17T12:39:55+00:00 Repair step: Upgrade app code from the marketplace
2017-07-17T12:39:55+00:00 Repair info: Enabling market app to assist with update
2017-07-17T12:39:55+00:00 Repair info: Using market to update existing apps
2017-07-17T12:39:55+00:00 Repair info: Attempting to update the following existing but incompatible app from market: activity, files_pdfviewer, files_texteditor, gallery
2017-07-17T12:39:55+00:00 Repair info: Fetching app from market: activity
2017-07-17T12:39:58+00:00 Repair warning: GuzzleHttp\Exception\ConnectException
2017-07-17T12:39:58+00:00 Repair warning: cURL error 7: Failed to connect to marketplace.owncloud.com port 443: No route to host
...
```
```
root@dima:/var/www/owncloud# sudo -u www-data php occ app:list
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
Enabled:
...
  - market: true
...
```

@cdamken

DeepDiver1975:
… was disabled in the past 

## Description
Prio to 10.0 appstoreenable was used as config option to disable installation from apps.owncloud.com

In migration scenarios we respect this value now and do not enable the market app-

## Related Issue
fixes https://github.com/owncloud/market/issues/118

## How Has This Been Tested?
- set version in config.php to 9.1.0.0
- set appstoreenabled to false in config.php
- run occ up

bildschirmfoto von 2017-08-16 15-25-38

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Repair/Apps.php
+++ b/lib/private/Repair/Apps.php
@@ -51,15 +51,22 @@ class Apps implements IRepairStep {
 	/** @var IConfig */
 	private $config;
 
+	/** @var \OC_Defaults */
+	private $defaults;
+
 	/**
 	 * Apps constructor.
 	 *
 	 * @param IAppManager $appManager
+	 * @param EventDispatcher $eventDispatcher
+	 * @param IConfig $config
+	 * @param \OC_Defaults $defaults
 	 */
-	public function __construct(IAppManager $appManager, EventDispatcher $eventDispatcher, IConfig $config) {
+	public function __construct(IAppManager $appManager, EventDispatcher $eventDispatcher, IConfig $config, \OC_Defaults $defaults) {
 		$this->appManager = $appManager;
 		$this->eventDispatcher = $eventDispatcher;
 		$this->config = $config;
+		$this->defaults = $defaults;
 	}
 
 	/**
@@ -75,7 +82,7 @@ class Apps implements IRepairStep {
 	 */
 	private function isCoreUpdate() {
 		$installedVersion = $this->config->getSystemValue('version', '0.0.0');
-		$currentVersion = implode('.', \OCP\Util::getVersion());
+		$currentVersion = implode('.', Util::getVersion());
 		$versionDiff = version_compare($currentVersion, $installedVersion);
 		if ($versionDiff > 0) {
 			return true;
@@ -90,10 +97,11 @@ class Apps implements IRepairStep {
 	private function requiresMarketEnable() {
 		$installedVersion = $this->config->getSystemValue('version', '0.0.0');
 		$versionDiff = version_compare('10.0.0', $installedVersion);
-		if ($versionDiff >= 0) {
-			return true;
+		if ($versionDiff < 0) {
+			return false;
 		}
-		return false;
+		return true;
+
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


lib/private/Repair/Apps.php
```diff
@@ -274,4 +275,16 @@ function ($appId) {
 	protected function loadApp($app) {
 		OC_App::loadApp($app, false);
 	}
+
+	/**
+	 * @return bool
+	 */
+	private function isAppStoreEnabled() {
+		// if appstoreenabled was explicitly disabled we shall not use the market app for upgrade
+		$appStoreEnabled = $this->config->getSystemValue('appstoreenabled', null);
```
PVince81: Will this code rerun during the next upgrade or only ever when upgrading from any OC < 10.0 version ?

Because I remember that we removed the "appstoreenabled" key after upgrading to 10.0.

PVince81 (approved):
:+1: looks good otherwise, if the answer to my comment is correct

lib/private/Repair/Apps.php
DeepDiver1975: this will trigger this step only for upgrade from oc <10.0 - https://github.com/owncloud/core/blob/6a545053e763a5846cc887a0f66ff15ead7e4f69/lib/private/Repair/Apps.php#L91-L96

is this what you are looking for?

lib/private/Repair/Apps.php
```diff
@@ -120,7 +121,7 @@ public function run(IOutput $output) {
 		}
 
 		// Check if we can use the marketplace to update apps as needed?
-		if($this->appManager->isEnabledForUser('market')) {
+		if($this->isAppStoreEnabled() && $this->appManager->isEnabledForUser('market')) {
```
PVince81: but this check is used here too ?

PVince81: if you leave it then it will be possible to continue using the obsolete "appstoreenabled" config to prevent loading the market app, even if enabled already

DeepDiver1975: coming from 9.1 to 10.0.x which ships market app will auto enable the market.
so it is necessary to not load the market app and trigger the update scenarios

PVince81: my worry is that this part of the code also runs when updating 10.x to 10.x+1 and might have side effects if someone has "appstoreenable => false" in their config.php despite the fact that we explicitly delete it later. (no `$requiresMarketEnable` check for this piece).

Unless I misunderstood and this repair step only ever runs once ?

DeepDiver1975: I guess I'm getting your point - shall we explicitly disable the market on the initial migration from 9.x.x to 10.x.x ?

PVince81:
@DeepDiver1975 https://github.com/owncloud/core/pull/27952

also looks like I didn't backport that one...

PVince81:
stable10: https://github.com/owncloud/core/pull/28714

lib/private/Repair/Apps.php
```diff
@@ -108,60 +109,64 @@ public function run(IOutput $output) {
...
+		$appStoreEnabled = $this->isAppStoreEnabled();
 
-		if($isCoreUpdate && $requiresMarketEnable) {
-			// Then we need to enable the market app to support app updates / downloads during upgrade
-			$output->info('Enabling market app to assist with update');
-			// delete old value that might influence old APIs
-			if ($this->config->getSystemValue('appstoreenabled', null) !== null) {
-				$this->config->deleteSystemValue('appstoreenabled');
+		// if the appstore was explicitly disabled -> disable market app as well
+		if (!$appStoreEnabled) {
```
PVince81: @deepdiver hmm, so if people keep "appstoreenabled" to false in config.php, every update will re-disable the market app ?
maybe we should delete the key after all?
unless you make sure that this block only every runs when upgrading from 9.x
then never again

DeepDiver1975: let me try something ....

lib/private/Repair/Apps.php
```diff
@@ -90,78 +90,77 @@ private function isCoreUpdate() {
...
 	/**
 	 * @param IOutput $output
 	 * @throws RepairException
 	 */
 	public function run(IOutput $output) {
-		$isCoreUpdate = $this->isCoreUpdate();
+
+		if ($this->config->getSystemValue('has_internet_connection', true) !== true) {
+			$output->info('No internet connection available - no app updates will be taken from the marketplace.');
+			$output->info('How to update apps in such situation please see https://doc.owncloud.org/server/10.0/admin_manual/upgrading/marketplace_apps.html');
```
PVince81: hard-coded 10.0 ? I think we have a router method somewhere for pointing at docs ?

Minor issue, feel free to ignore

lib/private/Repair/Apps.php
```diff
@@ -274,4 +274,40 @@ function ($appId) {
...
+
+		// no update from a version before 10.0 -> nothing to do, but allow apps to be updated
+		if (!$this->requiresMarketEnable()) {
+			return true;
+		}
+		// if the appstore was explicitly disabled -> disable market app as well
+		if (!$this->isAppStoreEnabled()) {
+			$this->appManager->disableApp('market');
+			$output->info('Appstore was disabled in past versions and marketplace interactions are disabled for now as well.');
+			$output->info('If you like to get automated app updates on upgrade please enable the market app and remove "appstoreenabled" from your config.');
```
PVince81: => `If you would like to`

PVince81 (approved):
Approach looks good 👍 

Please fix the wording and optionally the minor issue about doc link

lib/private/Repair/Apps.php
DeepDiver1975: Take care off ...

Issue #28715, opened 2017-08-16
Automated UI tests for sharing dialog fail if browser window is too narrow


phil-davis:
### Steps to reproduce
1. Have a test system where the chrome browser opens quite narrow by default
2. Run the test for the shareeAutocompletion feature
```
bash tests/travis/start_behat_tests.sh --feature tests/ui/features/other/shareeAutocompletion.feature
```

### Expected behaviour
The test passes

### Actual behaviour
The test fails with:
```
    And the folder "simple-folder" is shared with the user "user1"                                                                        # SharingContext::theFileFolderIsSharedWithTheUser()
    And the share dialog for the folder "simple-folder" is open                                                                           # SharingContext::theShareDialogForTheFileFolderIsOpen()
      WebDriver\Exception\UnknownError: unknown error: Element <a class="action action-share permanent shared-style" href="#" data-action="Share" data-original-title="" title="">...</a> is not clickable at point (608, 318). Other element would receive the click: <div id="shareTabView" class="tab shareTabView">...</div>
        (Session info: chrome=60.0.3112.90)
        (Driver info: chromedriver=2.29.461571 (8a88bbe0775e2a23afda0ceaf2ef7ee74e822cc5),platform=Linux 4.10.0-32-generic x86_64)
```

### Server configuration
**Operating system**:
Ubuntu 16.04
**Web server:**
PHP dev server
**Database:**
mySQL
**PHP version:**
7.0
**ownCloud version:** (see ownCloud admin page)
10.0.3 from master 2017-08-16

### Client configuration
**Browser:**
chrome 60.0.3112.90

**Operating system:**
Ubuntu 16.04

phil-davis:
The problem is that the step:
```
And the folder "simple-folder" is shared with the user "user1"
```
opens the sharing dialog, does the requested sharing but then leaves the sharing dialog panel open.

The next step tries to click the icon to open the sharing dialog panel to do its thing, but the sharing dialog is already open and is obscuring the icon which would open it.

If the browser window is wider, then the "open sharing dialog panel" icon is visible and can be clicked (which works OK when the sharing dialog panel is already open).

So the test has only been working when the browser window is wide enough.

I have a fix coming...

phil-davis:
Fixed in master #28716 and stable10 #28720

PR #28716, opened 2017-08-16
UI tests - close sharing dialog panel

phil-davis:
## Description
After doing a step that is a "whole functional transaction" in the sharing dialog, close the dialog at the end. e.g. ``theFileFolderIsSharedWithTheUser`` already opens the sharing dialog and does the sharing. Make it "clean up after itself" by closing the dialog on its way out.

Provide a step definition "Given I close the share dialog" that can be used in future by any test that wants to step through individual actions to open the share dialog, fill in a field, click a button, do this, do that, close the share dialog.

## Related Issue
#28715 

## Motivation and Context
Make the UI tests more robust.

## How Has This Been Tested?
Run UI feature:
```
bash tests/travis/start_behat_tests.sh --feature tests/ui/features/other/shareeAutocompletion.feature
```
with a narrow browser window and confirm it all passes.

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/lib/SharingDialog.php
+++ b/tests/ui/features/lib/SharingDialog.php
@@ -38,6 +38,7 @@ class SharingDialog extends OwnCloudPage
 	protected $shareWithTooltipXpath = "/..//*[@class='tooltip-inner']";
 	protected $shareWithAutocompleteListXpath = ".//ul[contains(@class,'ui-autocomplete')]";
 	protected $autocompleteItemsTextXpath = "//*[@class='autocomplete-item-text']";
+	protected $shareWithCloseXpath = "/..//*[@class='close icon-close']";
 	protected $suffixToIdentifyGroups = " (group)";
 
 	/**
@@ -155,7 +156,7 @@ class SharingDialog extends OwnCloudPage
 				break;
 			}
 		}
-		
+
 		if ($userFound !== true) {
 			throw new ElementNotFoundException("could not share with '$name'");
 		}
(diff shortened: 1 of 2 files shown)


tests/ui/features/lib/SharingDialog.php
```diff
@@ -215,4 +216,12 @@ public function getShareWithTooltip()
 		}
 		return $shareWithTooltip->getText();
 	}
+
+	/**
+	 * closes the sharing dialog panel
+	 */
+	public function closeSharingDialog()
+	{
+		$this->find("xpath", $this->shareWithCloseXpath)->click();
```
individual-it: lets use a two step approach like https://github.com/owncloud/core/pull/28716/files#diff-18ebd0658ed5041d666c2c11fafbf09bR214
otherwise you will have errors like `PHP Fatal error:  Uncaught Error: Call to a member function click() on null` if the element cannot be found, that is hard to catch. A `ElementNotFoundException` is nicer

individual-it (changes requested):

individual-it:
How comes that is used to work and is broken now? Is it a owncloud version change or browser?

phil-davis:
For the record, here is a screenshot of the type of thing that happens. The sharing dialog panel overlays the sharing and "..." icons in the file list, so while the sharing dialog panel is open there is no way to click those icons for a file (whether you are an automated test script or a real user)
sharing-panel-overlay

phil-davis:
I guess it is a browser change. For some reason, in my VM, chrome is always opening quite narrow. Maybe with Chrome 60 update or something, it changed the window size that it remembered.

tests/ui/features/lib/SharingDialog.php
phil-davis: Good thinking - done.

individual-it (approved):
so I guess its about the size of the window

phil-davis:
Backport stable10 #28720

PR #28719, opened 2017-08-17
allow to ignore specific chunks during size calculation for a chunked file

individual-it:
## Description
This change makes it possible to ignore some chunks when calculation the sum of all already uploaded chunks.

## Related Issue
https://github.com/owncloud/firewall/issues/356

## Motivation and Context
In the firewall app we need to be able to calculate the sum of all already uploaded chunks plus the one is currently uploaded. In the case that a chunk is uploaded a second time we need to ignore the size of the chunk that is already stored on the server.
e.g.:
client uploads chunk 1 5MB => sum 5MB
client uploads chunk 2 5MB => sum 10MB
client uploads chunk 3 5MB => sum 15MB
client reuploads chunk 1 5MB  => sum has to be 15MB and not 20MB

Here the PR in firewall that depend on this code https://github.com/owncloud/firewall/pull/397

## How Has This Been Tested?
Unit Tests + Firewall acceptance tests

## Types of changes

## Checklist:

- [x I have read the **CONTRIBUTING** document.

### DIFF
--- a/apps/dav/lib/Upload/FutureFile.php
+++ b/apps/dav/lib/Upload/FutureFile.php
@@ -96,14 +96,28 @@ class FutureFile implements \Sabre\DAV\IFile {
 	}
 
 	/**
-	 * @inheritdoc
+	 * Returns the size of the node, in bytes
+	 *
+	 * @param array $ignoreChildren array of names of children that should be
+	 * ignored for the calculation
+	 * @return int
 	 */
-	function getSize() {
+	function getSize($ignoreChildren = []) {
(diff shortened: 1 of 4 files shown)


phil-davis (approved):
Seems good.
It is a lot of effort to make sure the server cannot be tricked by the client!

DeepDiver1975 (changes requested):
as discussed - this should not go in here

DeepDiver1975:
As discussed - this is not needed -> close

PR #28728, opened 2017-08-17
these methods return string or null, document properly or fix

butonic:
All user backend methods / IProvidesXYZ either return a string or they don't. null is what the UserManager tests for because that is what is stored inside the Account table. https://github.com/owncloud/core/blob/master/lib/private/User/Manager.php#L432-L438

The default User Backend returns false for home which is only treated correctly because the UserMapper uses is_string on it: https://github.com/owncloud/core/blob/master/lib/private/User/Manager.php#L452

This PR fixes the PHP doc and returns null for an unset home.


PVince81 (approved):
:+1:

PR #28735, opened 2017-08-18
Fix check new dav rename target path name

### LINKED ISSUE #28441: Renaming to .htaccess problems in UI
### Steps to reproduce
1. Create a file aaaa.txt
2. Try to rename to .htaccess

### Expected behaviour
A reasonable error message is displayed - e.g. Could not rename "aaaa.txt"
The original aaaa.txt file is listed in the UI.

### Actual behaviour
Error displayed:
```
Could not rename "aaaa.txt", it does not exist any more
```
and the file is gone from the display in the UI.

Note: refreshing the page brings back the file aaaa.txt to the UI list. So the original file is still there on server storage.

### Server configuration
**Operating system**:
Ubuntu 16.04
**Web server:**
PHP dev
**Database:**
MySQL
**PHP version:**
7.0
10.0.2+ current master as of right now 20/07/2017

### Client configuration
**Browser:**
Firefox 54.0
**Operating system:**
Ubuntu 16.04

PVince81:
## Description
Verify target path earlier to make sure 403 is returned in case of wrong
target (ex: ".htaccess") instead of 404.

The reason this happens is because the rename code path between old and new DAV endpoints are different. Old DAV fails earlier when Sabre's CorePlugin does a `getNodeForPath()` on the target even for a local rename. New DAV doesn't go through this code path and reached `Node:setName()` more directly, possibly due to the necessary changes that were done lately regarding moves.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28441.
Should also fix the renameFile/renameFolder failures on master.

## How Has This Been Tested?
Manual testing with cadaver on new dav endpoint: rename "welcome.txt" to ".htaccess".
Before the fix: 404
After the fix: 403

Automated tests have been reenabled.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Node.php
+++ b/apps/dav/lib/Connector/Sabre/Node.php
@@ -126,11 +126,22 @@ abstract class Node implements \Sabre\DAV\INode {
 			throw new \Sabre\DAV\Exception\Forbidden();
 		}
 
+		// verify path of the source
+		$this->verifyPath();
+
 		list($parentPath,) = \Sabre\HTTP\URLUtil::splitPath($this->path);
 		list(, $newName) = \Sabre\HTTP\URLUtil::splitPath($name);
 
-		// verify path of the target
-		$this->verifyPath();
+		// verify path of target
+		if (\OC\Files\Filesystem::isForbiddenFileOrDir($parentPath . '/' . $newName)) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		try {
+			$this->fileView->verifyPath($parentPath, $newName);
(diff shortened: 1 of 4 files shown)


individual-it:
there are more skipped tests in `tests/integration/features/webdav-related-new-endpoint.feature`

PVince81:
@individual-it I can't find them

individual-it (approved):
👍 fixes the issue

individual-it:
sorry did overlooked one file in your commit

PVince81:
stable10: https://github.com/owncloud/core/pull/28737

phil-davis (approved):

DeepDiver1975 (approved):

PR #28729, opened 2017-08-17
Use i provides e mail backendwhen syncing u sers

butonic:
make use of IProvidesEMailBackend and IProvidesQuotaBackend when syncing users. Needed to properly sync from LDAP

PR #28736, opened 2017-08-18
[stable10] use IProvidesEMailBackend when syncing a backend that supp…

DeepDiver1975:
…orts it
backport of #28729 to stable10


PVince81 (approved):
:+1:

PR #28737, opened 2017-08-18
[stable10] Fix check new dav rename target path name

PVince81:
Backport of https://github.com/owncloud/core/pull/28735/ to stable10

UI tests don't need adjusting as they are correct since stable10 UI uses old DAV endpoint.

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Node.php
+++ b/apps/dav/lib/Connector/Sabre/Node.php
@@ -126,11 +126,22 @@ abstract class Node implements \Sabre\DAV\INode {
 			throw new \Sabre\DAV\Exception\Forbidden();
 		}
 
+		// verify path of the source
+		$this->verifyPath();
+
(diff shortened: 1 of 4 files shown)


individual-it (approved):
tested and works on stable10.
Shouldn't we copy the tests from master over here, they are good tests to run and would protect us from regression

PVince81:
@individual-it not sure, when looking at the conflict I saw that ".htaccess" had been added into a list of existing tests instead of a separate test. Can you guys check ?

individual-it:
@PVince81 I've made a PR against your branch https://github.com/owncloud/core/pull/28738
That are the tests and they should pass (at least they do locally for me)

We could merge each of them into the scenario just before it but I've left them separately as
"Rename a folder using forbidden characters" and "Rename a folder to a forbidden name" are different testing scenarios

PR #28738, opened 2017-08-18
UI tests, testing renaming a file/folder into .htaccess

individual-it:
some additional UI tests

### DIFF
--- a/tests/ui/features/files/renameFiles.feature
+++ b/tests/ui/features/files/renameFiles.feature
@@ -36,11 +36,9 @@ Feature: renameFiles
 
 	Scenario: Rename a file using forbidden characters
 		When I rename the file "data.zip" to one of these names
-		|.htaccess  |
 		|lorem\txt  |
 		|\\.txt     |
 		Then notifications should be displayed with the text
-		|Could not rename "data.zip"|
 		|Could not rename "data.zip"|
 		|Could not rename "data.zip"|
 		And the file "data.zip" should be listed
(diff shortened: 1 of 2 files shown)


tests/ui/features/files/renameFiles.feature
```diff
@@ -45,6 +45,13 @@ Feature: renameFiles
 		|Could not rename "data.zip"|
 		And the file "data.zip" should be listed
 
+	Scenario: Rename a file to a forbidden name
```
phil-davis: Isn't this the same as the .htaccess case of the scenario at line 37?

tests/ui/features/files/renameFolders.feature
```diff
@@ -44,6 +44,13 @@ Feature: renameFolders
 		|Could not rename "simple-folder"|
 		And the folder "simple-folder" should be listed
 
+	Scenario: Rename a folder to a forbidden name
```
phil-davis: and this is already on the scenario at line 36?

individual-it:
true deleted the duplication. Maybe its good to keep the separate https://github.com/owncloud/core/pull/28737#issuecomment-323309530
If not merge the ones in master too

phil-davis:
So do you just want to make a separate scenario so that the scenario text is more accurate?
Because splitting this out will take a bit more run time.

Also, if you do it this way, you need to also delete the extra 'Could not rename "data.zip"' in the other scenarios.

individual-it:
> So do you just want to make a separate scenario so that the scenario text is more accurate?
Because splitting this out will take a bit more run time.

ether splitting here or merging in master

PR #28757, opened 2017-08-21
[stable10] Don't enable market app on upgrade from 9.1.x or lower whe…

DeepDiver1975:
backport of #28711 to stable10


PVince81 (approved):
:+1:

Issue #28758, opened 2017-08-21
CLI Upgrade: stat: illegal option -- c


tuaris:
### Steps to reproduce
1. Upgrade to 10.0.2
2. Run upgrade in browser
3. Run upgrade in console (php occ upgrade)

### Expected behavior
Upgrade should complete

### Actual behavior
Browser based upgrade fails.  Console based upgrade does not complete.  Error messages shown are

```
stat: illegal option -- c
usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file ...]
```

### Server configuration
**Operating system**: FreeBSD 10.3

**Web server:** Apache 2.4

**Database:** MySQL 5.6

**PHP version:** 5.6

**ownCloud version:** 10.0.2

**Updated from an older ownCloud or fresh install:** Update from 9.x

**Where did you install ownCloud from:** FreeBSD Ports

**Signing status (ownCloud 9.0 and above):**

Unable to sign in.

DeepDiver1975:
The code in question is here

https://github.com/owncloud/core/blob/bafdb425f350b19de79f5c522388df5ef878b800/lib/private/LargeFileHelper.php#L148-L159

DeepDiver1975:
Looks like your OS is detected as linux and not as bsd ...

Please run this in your terminal

```sh

php -r "echo php_uname('s') . PHP_EOL;"

**Comment by DeepDiver1975** [comment]
FYI: This code is triggered mainly because you are running a 32 bit and not 64 bit architecture

**Comment by tuaris** [comment]
```
 php -r "echo php_uname('s') . PHP_EOL;"
FreeBSD
```

This is 32-bit (yes I know I should upgrade to 64-bit ).

phil-davis:
filemtime checks have a ``stat -c`` command that only has the case for Linux. Maybe #28759 will help?

kdslkdsaldsal:
Ref: https://github.com/owncloud/core/pull/28761

phil-davis:
Should be fixed with #28759 and backport to stable10 #28790

PR #25428, opened 2016-07-10
fixing php 32 bit (arm) filemtime on large file issue (#18971)

cyberb:
Filemtime fix for 32bit platform only.
This contribution is MIT licensed

Issue #18971, opened 2015-09-11
php 32 bit (arm) filemtime on large file issue


cyberb:
I am facing an issue with running file:scan on arm:

filemtime(): stat failed

Looks like the reason is filemtime [1] is not working for files over 2G on 32 bit php [2]

Fixing function according to php website comment solved the problem:

```
  exec ('stat -c %Y '. escapeshellarg ($this->getSourcePath($path)));
```

I am sure that worked fine in past with older owncloud/php versions on the same hardware.

Is there something I am missing or this never worked on php (32bit) / arm?
Or is it possible to have some if, like:

```
if (PHP_INT_SIZE === 4)
   return exec ('stat -c %Y '. escapeshellarg ($this->getSourcePath($path)));
else
   return filemtime($this->getSourcePath($path));
```

I cannot upgrade php to 64 bit on arm board as there there is no such thing yet as arm64 or at least not popular enough on single board computers yet.

[1] http://php.net/manual/en/function.filemtime.php#68814
[2] https://github.com/owncloud/core/blob/master/lib/private/files/storage/local.php#L159

My system:
ownCloud version 8.0.4
PHP Version => 5.6.9
System => Linux cubietruck 3.4.79-sun7i #13 SMP PREEMPT Thu Jul 3 06:34:55 CST 2014 armv7l
Configure Command =>  './configure'  '--enable-fpm' '--with-pgsql' '--with-pdo-pgsql' '--enable-opcache' '--prefix' '--with-config-file-path=/root/build/install/config' '--with-gd' '--enable-zip' '--with-zlib' '--with-curl' '--with-readline' '--with-ldap' '--with-ldap-sasl' '--with-libdir=lib/arm-linux-gnueabihf' '--enable-mbstring'

hoorna:
Hi,

I am also facing an issue with file:scan. I hope I am allowed to post it here.

While running the file:scan command I am receiving many, many, many times the next message:

         stat: invalid option -- 'c'
         BusyBox v1.16.1 (2015-06-29 18:12:44 CST) multi-call binary.

         Usage: stat [OPTIONS] FILE...

         Display file (default) or filesystem status

         Options:
                 -f      Display filesystem status
                 -L      Follow links
                 -t      Display info in terse form

I am running OwnCloud on a Synology Diskstation. Synology implemented the Busybox shell on their Diskstations.

I don't know if there is something going wrong nor how to solve this.

cyberb:
[...]

PR #28759, opened 2017-08-22
Handle BSD case for 32 bit filemtime and install warning

phil-davis:
## Description
Handle the BSD (and Mac) case in lib/private/Files/Storage/Local.php filemtime
That was performing a Linux-style stat command only.

Make a generic ``runningOn`` convenience function and refactor existing code that did special strpos() stuff to detect the OS.

Remove some legacy tests for "win" that were still in getFileSizeViaExec

Rearrange install OS error checks in Setup.php so that the error (really a warning) will be displayed for any OS that is not linux-type.

## Related Issue
#28758 
#28760 

## Motivation and Context
BSD stat command has different options to Linux

## How Has This Been Tested?
Copied function runningOnBSD() onto my pfSense FreeBSD system and tried calling it manually.
It returns true for that, should also work for "OpenBSD" or other variants.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -180,7 +180,12 @@ class Local extends \OC\Files\Storage\Common {
 			return false;
 		}
 		if (PHP_INT_SIZE === 4) {
-			return (int) exec ('stat -c %Y '. escapeshellarg ($fullPath));
+			if (\OC_Util::runningOn('linux')) {
+				return (int) exec ('stat -c %Y '. escapeshellarg ($fullPath));
+			} else if (\OC_Util::runningOn('bsd') || \OC_Util::runningOn('mac')) {
+				return (int) exec ('stat -f %m '. escapeshellarg ($fullPath));
+			}
+			return false;
 		}
 		return filemtime($fullPath);
 	}
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -1255,27 +1255,16 @@ class OC_Util {
 	/**
 	 * Checks whether the server is running on the given OS type
 	 *
-	 * @param string $osType linux|mac|bsd
+	 * @param string $osType linux|mac|bsd etc
 	 * @return bool true if running on that OS type, false otherwise
 	 */
 	public static function runningOn($osType) {
-		$osType = strtolower($osType);
+		$osType = strtolower($osType) === 'mac' ? 'darwin' : strtolower($osType);
 
-		switch($osType) {
-			case 'linux':
-				return (strtolower(substr(PHP_OS, 0, 5)) === 'linux');
-				break;
-
-			case 'mac':
-				return (strtolower(substr(PHP_OS, 0, 6)) === 'darwin');
-				break;
-
-			case 'bsd':
-				return (strpos(strtolower(PHP_OS), 'bsd') !== false);
-				break;
-
-			default;
-				return false;
+		if ($osType === 'bsd') {
+			return (strpos(strtolower(PHP_OS), $osType) !== false);
+		} else {
+			return (strtolower(substr(PHP_OS, 0, strlen($osType))) === $osType);
 		}
 	}
 
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -1253,12 +1253,30 @@ class OC_Util {
 	}
 
 	/**
-	 * Checks whether the server is running on Mac OS X
+	 * Checks whether the server is running on the given OS type
 	 *
-	 * @return bool true if running on Mac OS X, false otherwise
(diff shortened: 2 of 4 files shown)


phil-davis:
Note: The filemtime() 32-bit check was introduced by PR #25428 for issue #18971 but that code only had a ``stat`` command for Linux.

DeepDiver1975:
Nice:+1:

lib/private/Files/Storage/Local.php
```diff
@@ -180,7 +180,12 @@ public function filemtime($path) {
 			return false;
 		}
 		if (PHP_INT_SIZE === 4) {
-			return (int) exec ('stat -c %Y '. escapeshellarg ($fullPath));
+			if (\OC_Util::runningOnLinux()) {
+				return (int) exec ('stat -c %Y '. escapeshellarg ($fullPath));
```
DeepDiver1975: Unrelated to this change ... 

Isn't it stupid to call stat to get values higher then 32 bit int and then case that to a 32bit int?

DeepDiver1975 (changes requested):

lib/private/Files/Storage/Local.php
phil-davis: A good question!

phil-davis: Not quite as stupid as it looks. The problem is that for files of SIZE > 2GB on 32-bit systems, even filemtime() fails:
http://php.net/manual/en/function.filemtime.php#68814

The file SIZE might be too big for a 32-bit signed int, but mtine will fit OK until 2038.

Since filemtime() returns an int, I guess whoever made wrote this also returned an int. But it would not matter to just remove the (int) cast.

phil-davis:
Rebased - CI should be passing on latest master.

DeepDiver1975 (approved):

phil-davis:
@DeepDiver1975 @mmattel I refactored this based on comments in the other related PR #28761 and included the little bit of Setup.php code from there. The refactoring meant that I had to touch Setup.php anyway.

mmattel:
Code looks good, not tested :+1: 
Jenkins: docker: Error response from daemon: grpc: the connection is unavailable...

phil-davis:
Rebased again and Jenkins still dies early. I guess there will be no CI until someone is available to look at it.

lib/private/legacy/util.php
```diff
@@ -1253,12 +1253,30 @@ public static function obEnd() {
...
-	 * Checks whether the server is running on Mac OS X
+	 * Checks whether the server is running on the given OS type
 	 *
-	 * @return bool true if running on Mac OS X, false otherwise
+	 * @param string $osType linux|mac|bsd
+	 * @return bool true if running on that OS type, false otherwise
 	 */
-	public static function runningOnMac() {
-		return (strtoupper(substr(PHP_OS, 0, 6)) === 'DARWIN');
+	public static function runningOn($osType) {
```
DeepDiver1975: a little bit less code ;-)

```php
$osType = strtolower($osType) == 'mac' ? 'darwin' : strtolower($osType);
return (strtolower(substr(PHP_OS, 0, strlen($osType))) === $osType);

File: lib/private/legacy/util.php [hunk 3]
```diff
@@ -1253,12 +1253,30 @@ public static function obEnd() {
...
+		$osType = strtolower($osType);
+
+		switch($osType) {
+			case 'linux':
+				return (strtolower(substr(PHP_OS, 0, 5)) === 'linux');
+				break;
```
PVince81: no need to break if you return directly

lib/private/legacy/util.php
phil-davis: bsd case is different, because for that it is not necessarily at the start of the string: 
```
$x = PHP_OS;
var_dump($x);
string(7) "FreeBSD"
```

DeepDiver1975: good point - didnt look that close - sorry

phil-davis:
I refactored it - maybe it is more obscure now?

But it is "flexible" - by default it checks for the parameter matching the start of PHP_OS. It can be passed 'lin' (matching a Linux...) or 'darw' (matching a 'darwin...') or 'free' (matching 'FreeBSD').
And it knows about the special cases of 'mac' => 'darwin' and 'bsd' can come anywhere in PHP_OS.

But also, I can't imagine using the "flexibility" any time soon.

mmattel:
:+1:

phil-davis:
@PVince81 or @DeepDiver1975 do either of you want to give this a final approval and press merge?
Or change or revert back something here?

phil-davis:
Backport stable10 #28790 
since this makes the setup warning more generic, applying to all attempts at installing on a non-linux-style platform.

Issue #28760, opened 2017-08-22
No warning given when installing on BSD


phil-davis:
### Steps to reproduce
1. Try ownCloud server on FreeBSD

### Expected behaviour
Like on Mac, a warning is given during setup to say the platform is not supported.

### Actual behaviour
No warning is given.

phil-davis:
Fixed by #28759 
Backported to stable10 #28790

PR #28761, opened 2017-08-22
Give install warning on all OS other than Linux

phil-davis:
## Description
Make the install warning for Mac OS X more generic, so the warning is given for any OS that is not Linux.

## Related Issue
#28760 

## Motivation and Context
People can install on officially-unsupported *BSD at the moment, without being given a warning.

## How Has This Been Tested?
Remove the "!" in front of \OC_Util::runningOnLinux() and set installed to false in a dev system.
(to trick it into thinking it is an unsupported system)
Go to the web UI. Confirm that the setup page gives a warning with reasonable text.

## Types of changes

## Checklist:

This was noticed when looking at code that used runningOnMac(), PHP_OS and/or php_uname to do stuff based on the operating system.

### DIFF
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -203,12 +203,18 @@ class Setup {
 			\OC\Setup::protectDataDirectory();
 		}
 
-		if (\OC_Util::runningOnMac()) {
+		if (!\OC_Util::runningOnLinux()) {
+			if (\OC_Util::runningOnMac()) {
+				$os = "Mac OS X";
+			} else {
+				$os = PHP_OS;
+			}
+
 			$errors[] = [
 				'error' => $this->l10n->t(
-					'Mac OS X is not supported and %s will not work properly on this platform. ' .
+					'%s is not supported and %s will not work properly on this platform. ' .
 					'Use it at your own risk! ',
-					$this->defaults->getName()
+					array($os, $this->defaults->getName())
 				),
 				'hint' => $this->l10n->t('For the best results, please consider using a GNU/Linux server instead.')
 			];
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -1253,12 +1253,30 @@ class OC_Util {
 	}
 
 	/**
+	 * Checks whether the server is running on Linux
+	 *
+	 * @return bool true if running on Linux, false otherwise
+	 */
+	public static function runningOnLinux() {
+		return (strtolower(substr(PHP_OS, 0, 5)) === 'linux');
+	}
+
+	/**
 	 * Checks whether the server is running on Mac OS X
 	 *
 	 * @return bool true if running on Mac OS X, false otherwise
 	 */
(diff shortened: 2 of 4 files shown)


phil-davis:
Note: this is on top of #28759 so that it gets function runningOnLinux()
If that PR does not happen, then this one will need refactoring.

DeepDiver1975:
Agreed :+1:

DeepDiver1975:
Backport Please

mmattel:
Q: why not a generic ``runningOn`` function and return a defined value which than can be queried against?

lib/private/Setup.php
```diff
@@ -203,12 +203,18 @@ public function getSystemInfo($allowAllDatabases = false) {
...
+				$os = PHP_OS;
+			}
+
 			$errors[] = [
 				'error' => $this->l10n->t(
-					'Mac OS X is not supported and %s will not work properly on this platform. ' .
+					'%s is not supported and %s will not work properly on this platform. ' .
 					'Use it at your own risk! ',
-					$this->defaults->getName()
+					array($os, $this->defaults->getName())
```
DeepDiver1975: [ ] please not array()

lib/private/legacy/util.php
```diff
@@ -1253,12 +1253,30 @@ public static function obEnd() {
 	}
 
 	/**
+	 * Checks whether the server is running on Linux
+	 *
+	 * @return bool true if running on Linux, false otherwise
+	 */
+	public static function runningOnLinux() {
```
DeepDiver1975: generally speaking we should not put more code into legacy classes .... but I have no better location where to but this ...

DeepDiver1975 (changes requested):

lib/private/Setup.php
phil-davis: I cut-pasted from https://doc.owncloud.org/server/10.0/developer_manual/core/translation.html
```
$l->t('%s is available. Get <a href="%s">more information</a>', array($data['versionstring'], $data['web']));
```
I guess that example is not the current way it is liked to be done.

phil-davis:
@mmattel yes, I could make
```
runningOn($os_type)
```
and have it take "linux", "mac" or "bsd" and return the boolean answer if on that type of OS.

Does it make much difference to having the 3 functions?

mmattel:
@phil-davis 
I always think that things can be used for other purposes maybe later on, and I am esthetic, it just looks better - from my pov :smile: 
(or return the text result and query against that like ``runningOn() === "linux"`` )

phil-davis:
Note: the "runningOn" functions are actually added in https://github.com/owncloud/core/pull/28759 - which this sits on top of. (When I started looking at a BSD detection thing for filemtime then I also noticed the setup page thing, so had 2 reasons for similar underlying code) So I will refactor there...

phil-davis:
After refactoring, the code all "just fitted" in #28759 
So closing this here.

Issue #28754, opened 2017-08-21
file/foldernames with multiple spaces cannot be used in UI tests


individual-it:
### Steps to reproduce
1. try to create a UI tests that searches for a file/foldername that contains multiple space char. e.g. (`file     name`)

### Expected behaviour
findFileRowByName() find the file

### Actual behaviour
file is not found even its correctly shown in the UI
seems like `getText()` returns only one space and not multiple

phil-davis:
Backport stable10 #28782

phil-davis:
Note: solution also allows testing a filename with a mix of single and double quotes.

PR #28768, opened 2017-08-22
Tests for files with multiple spaces

phil-davis:
## Description
scrollDownAppContent() - refactor so it does not need to be passed the number of files that were on the screen when it was called. It can keep track of that itself.

findFileRowByName() - use a more specific xpath that looks directly for an element that has the desired text. That xpath element matching works without stripping/compressing white space, retain exact matching to things like a filename with multiple contiguous spaces.

(The previous problem was that the Mink getText() method returned a stripped-compressed "normalized" version of the string in the HTML, which prevented exact comparisons with the expected file name)

Add various tests to confirm that users can put (multiple) spaces in front, in the middle and at the end of file names.

## Related Issue
#28754 

## Motivation and Context
Make tests be able to use files with multiple contiguous spaces in the file name.

## How Has This Been Tested?
Local UI test run.

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/lib/FilesPage.php
+++ b/tests/ui/features/lib/FilesPage.php
@@ -64,6 +64,7 @@ class FilesPage extends OwnCloudPage
 	 * If name is not given a random one is chosen
 	 *
 	 * @param string $name
+	 * @return string name of the created file
 	 */
 	public function createFolder($name = null)
 	{
@@ -93,6 +94,35 @@ class FilesPage extends OwnCloudPage
 		);
 	}
 
+	/**
+	 * Surround the text with single or double quotes, whichever does not
+	 * already appear in the text. If the text contains both single and
+	 * double quotes, then throw an InvalidArgumentException.
+	 *
+	 * The returned string is intended for use as part of an xpath (v1).
+	 * xpath (v1) has no way to escape the quote character within a string
+	 * literal. So there is no way to directly use a string containing
+	 * both single and double quotes.
+	 *
+	 * @param string $text
+	 * @return string the text surrounded by single or double quotes
+	 * @throws \InvalidArgumentException
+	 */
+	public function quotedText($text)
+	{
+		if (strstr($text, "'") === false) {
+			return "'" . $text . "'";
+		} else if (strstr($text, '"') === false) {
+			return '"' . $text . '"';
+		} else {
+			// The text contains both single and double quotes.
+			// With current xpath v1 there is no way to encode that.
+			throw new \InvalidArgumentException(
+				"mixing both single and double quotes is unsupported - '" . $text ."'"
+			);
+		}
+	}
+
 	public function findActionMenuByNo($number)
 	{
 		$xpath = sprintf($this->fileActionMenuBtnXpathByNo,$number);
@@ -102,27 +132,32 @@ class FilesPage extends OwnCloudPage
 	/**
 	 * finds the complete row of the file
 	 *
-	 * @param string $name
 	 * @param Session $session
+	 * @param string $namePartA
+	 * @param string $namePartB
 	 * @return \Behat\Mink\Element\NodeElement|NULL
 	 * @throws \SensioLabs\Behat\PageObjectExtension\PageObject\Exception\ElementNotFoundException
 	 */
-	public function findFileRowByName($name, Session $session)
+	public function findFileRowByName(Session $session, $namePartA, $namePartB = '')
 	{
 		$previousFileCount = 0;
 		$currentFileCount = null;
 
-		if (strstr($name, "'") === false) {
-			$quotedName = "'" . $name . "'";
-		} else {
-			$quotedName = '"' . $name . '"';
-		}
+		// Concatenating a possible 2 separate parts of the file name allows
+		// one part to contain single quotes and other the other to contain
+		// double quotes.
+		$xpathQuotedName =
+			"concat(" .
+			$this->quotedText($namePartA) .
+			"," .
+			$this->quotedText($namePartB) .
+			")";
 
 		//loop to keep on scrolling down to load not viewed files
 		//when the scroll does not retrieve any new files, the file is not there
 		do {
 			$fileNameMatch = $this->find("xpath", $this->fileListXpath)->find(
-				"xpath", sprintf($this->fileNameMatchXpath, $quotedName)
+				"xpath", sprintf($this->fileNameMatchXpath, $xpathQuotedName)
 			);
 
 			if (is_null($fileNameMatch)) {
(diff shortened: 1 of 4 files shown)


tests/ui/features/lib/FilesPage.php
```diff
@@ -102,45 +103,38 @@ public function findActionMenuByNo($number)
...
-		$previousFileCounter = 0;
-		$fileNameSpans = array();
+		$newFiles = true;
+		$fileNameMatches = array();
+
+		if (strstr($name, "'") === false) {
+			$quotedName = "'" . $name . "'";
+		} else {
+			$quotedName = '"' . $name . '"';
+		}
```
phil-davis: Note: another combination we cannot test is a file name containing both single and double quotes.
I can raise a separate issue about that.
It is not so easy (not possible?) with the current xpath (which I think is only v1) to escape within a string.
```
"single'and\"double"
'single\'and"double'
```
do not work.

And there are issues with how to write such things in Gherkin also.

tests/ui/features/lib/FilesPage.php
```diff
@@ -175,17 +174,20 @@ public function scrollDownAppContent ($numberOfFilesOld, Session $session, $time
...
 		$end = $currentTime + ($timeout_msec / 1000);
-		while ($currentTime <= $end) {
+
+		do {
 			$this->waitForOutstandingAjaxCalls($session);
 			$fileNameSpans = $this->find("xpath", $this->fileListXpath)->findAll(
 				"xpath", $this->fileNamesXpath
 			);
-			if (count($fileNameSpans) >= $numberOfFilesOld) {
+			if (count($fileNameSpans) >= $initialFileNameSpansCount) {
```
phil-davis: Note: IMO in the old code, and I left it the same here, this test means that we only ever go around this loop once. Because if the scroll has finished, then great, we have a new (and bigger) count. Otherwise if the scroll has not finished then $fileNameSpans will be at the initial number, and the comparison passes anyway.

So I think the whole loop could be removed, and I just have waitForOutstandingAjaxCalls()

individual-it: yes that sounds logical. I think the loop came from the times where we didn't have the `waitForOutstandingAjaxCalls()` function

tests/ui/features/lib/FilesPage.php
```diff
@@ -160,13 +154,18 @@ public function openSharingDialog ($name, Session $session)
...
-	 * @param int $numberOfFilesOld how many files were listed before the scroll.
-	 * So we can guess how long to wait for the loading of new files to finish
 	 * @param Session $session
 	 * @param int $timeout_msec
+	 * @return boolean true if more files were listed after scrolling
 	 */
-	public function scrollDownAppContent ($numberOfFilesOld, Session $session, $timeout_msec = STANDARDUIWAITTIMEOUTMILLISEC)
+	public function scrollDownAppContent (Session $session, $timeout_msec = STANDARDUIWAITTIMEOUTMILLISEC)
 	{
+		$initialFileNameSpansCount = count(
```
individual-it: this now makes one extra `findAll` that costs time. but will be obsolete if you implement what you say about the loop further down

phil-davis: Yes, I can remove even more of the general xpath queries that find all the file rows on the page.

tests/ui/features/lib/FilesPage.php
```diff
@@ -43,6 +43,7 @@ class FilesPage extends OwnCloudPage
 	protected $fileActionMenuBtnXpath = "//a[@data-action='menu']";
 	protected $fileActionMenuXpath = "//div[contains(@class,'fileActionsMenu')]";
 	protected $fileNamesXpath = "//span[@class='nametext']";
+	protected $fileNameMatchXpath = "//span[@class='nametext' and .=%s]";
```
individual-it: hope that proves to be stable. I had problems with cases where `getText()` would not return the correct value so I had to use `getHtml()` additionally.

tests/ui/features/lib/FilesPage.php
```diff
@@ -102,45 +103,38 @@ public function findActionMenuByNo($number)
...
-					);
-					if ($fileRow === null) {
-						throw new ElementNotFoundException("could not find fileRow with xpath '" . $this->fileRowFromNameXpath . "'");
-					} else {
-						return $fileRow;
-					}
-				}
+		//when the scroll does not retrieve any new files, the file is not there
+		while ((!count($fileNameMatches)) && $newFiles) {
+			$fileNameMatches = $this->find("xpath", $this->fileListXpath)->findAll(
```
individual-it: why `findAll()` here and not `find()`? your are searching for the exact name, so you should have only one result ever and you also only using `$fileNameMatches[0]`

phil-davis: Yes, I thought about that also. Early on I was trying to do a minimal change version - but the diffs "just grew" - so I might as well fix this up also.

tests/ui/features/lib/FilesPage.php
phil-davis: This builds a direct xpath that should only match to the given actual filename string - no messing about with getText() which trims and compresses whitespace. So it kind of bypasses Mink. Hopefully it reliably matches the page XML.

phil-davis:
IMO this is good now to address the issue raised and refactor related stuff.

There are (of course) other things that could be refactored - e.g. stuff related to "I am on the files page" and making sure that waits nicely for the page to finish loading, and optimizing it to not reload the whole page if we are already there. But those are different things.

phil-davis:
Rebased and squashed - let's see what Travis-CI thinks now.

phil-davis:
Travis is happy. But Jenkins has trouble - awaiting Jenkins repair before this can proceed.

phil-davis:
I worked out a way to be able to construct a file name with both single and double quotes in it by passing it in 2 pieces in the gherkin step definition, then putting it together for xpath using concat().

That code is all there now. I have run all the UI tests locally that I think depend on my changes. Let's see if a full Travis run is happy.

tests/ui/features/lib/FilesPage.php
```diff
@@ -95,52 +132,55 @@ public function findActionMenuByNo($number)
...
+	public function findFileRowByName(Session $session, $namePartA, $namePartB = '')
 	{
-		$appContentHeight = 0;
-		$previousFileCounter = 0;
-		$fileNameSpans = array();
+		$previousFileCount = 0;
+		$currentFileCount = null;
+
+		// Concatenating a possible 2 separate parts of the file name allows
+		// one part to contain single quotes and other the other to contain
```
phil-davis: "other the other" - I will fix...

individual-it (approved):

phil-davis:
Backport stable10 #28782

PR #28772, opened 2017-08-22
Reduce JS test warnings

PVince81:
## Description
Remove "ERROR: undefined" at the end of the tests which were due to the
"onbeforeunload" event.
Removed "download.svg" errors.
Removed warnings about missing plurals.

For the rest, need to spend some time to replace the last tipsy usages and port some of the notification calls.

## Related Issue
None raised

## Motivation and Context
Warnings are annoying.

## How Has This Been Tested?
`make test-js`
See that there are less warnings.

## Types of changes

## Checklist:

### DIFF
--- a/core/js/tests/specs/coreSpec.js
+++ b/core/js/tests/specs/coreSpec.js
@@ -994,7 +994,12 @@ describe('Core base tests', function() {
 			var $rows = $el.find('.row');
 			expect($rows.length).toEqual(2);
 
+			// when called without arguments, a warning to be logged
+			// to ask devs to adjust their usage of the API.
+			// we stub it to avoid polluting the actual console
+			var warnStub = sinon.stub(console, 'warn');
 			OC.Notification.hide();
+			warnStub.restore();
 
 			$rows = $el.find('.row');
 			expect($rows.length).toEqual(1);
--- a/apps/files/tests/js/fileUploadSpec.js
+++ b/apps/files/tests/js/fileUploadSpec.js
@@ -133,7 +133,7 @@ describe('OC.Upload tests', function() {
 				'<input type="checkbox" id="select_all_files" class="select-all">' +
 				'<a class="name columntitle" data-sort="name"><span>Name</span><span class="sort-indicator"></span></a>' +
 				'<span id="selectedActionsList" class="selectedActions hidden">' +
-				'<a href class="download"><img src="actions/download.svg">Download</a>' +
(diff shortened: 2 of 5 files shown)


PVince81:
for the remaining parts I raised https://github.com/owncloud/core/issues/28773.

core/js/tests/specs/coreSpec.js
```diff
@@ -994,8 +994,12 @@ describe('Core base tests', function() {
 			var $rows = $el.find('.row');
 			expect($rows.length).toEqual(2);
 
+			var warnStub = sinon.stub(console, 'warn');
 			OC.Notification.hide();
 
+			expect(warnStub.calledOnce).toEqual(true);
```
jvillafanez: I don't think this is a good approach. Expecting that a warning message will popup doesn't feel good.

The warning message shouldn't popup in the first place, so we should try to fix the cause of the warning first. In addition, we shouldn't expect that the warn method will be called because it isn't essential for the test. As we don't check a "log" method of any logger will be called (thinking in PHP), we shouldn't check if a "warn" method will be called.

PVince81: The cause of the warning is that some code out there (users page) is still using this method without arguments.

The expectation of this test is that a warning does appear, so I think it's ok here.
If this bothers you then I should probably remove this test completely.
But the point of the test was also to make sure that hiding still works even if the arg is missing.

Or if you like I can just remove the expectation but keep the method stubbed to avoid output.

jvillafanez: > Or if you like I can just remove the expectation but keep the method stubbed to avoid output.

I think it's this (with a comment to explain why this is stubbed), or make the code not to use the warn method.

PVince81: added comment https://github.com/owncloud/core/pull/28772/commits/df9209812af87ba19b19ab28a412848c49e66cf7#diff-163902768f6bb1fc76108ce0e0c7cd5eR997

jvillafanez:
I'd definitely remove the expectations for the warning message in the console: they're not part of what we want to test. What we want to test is the global behaviour: you call `hide()` and the first message in the html is removed. We aren't interested in a console message poping somewhere the same way we aren't interested in whether a log message shows when a function is called.

PVince81:
@jvillafanez I've removed the "expect()" block

jvillafanez:
👍

PR #28774, opened 2017-08-23
Changes to encryption wrapper

sharidas:
Changes made to encryption wrapper so that we can
get the commands like transfer-ownership or recreate
masterkey work. This change doesn't alter the core
functionality of copy or fopen function. The arguments
passed to the function remains same as other wrappers.

## Description

Changes made to the ecnryption warpper and view to get the transfer-ownership command recreate master key work. Basically there was a problem which caused to change the arguments passed to the routines like fopen or copy. With this change we don't have to alter them. But we use flags to control the code flow for the commands.

## Motivation and Context

This change would prevent the modification to common routines which were made to arguments of functions like fopen or copy etc.

## How Has This Been Tested?

- [x] Scratch installation of oC and create 2 users
- [x] Verified transfer ownership command works with and without encryption between two users.
- [x] Verified Recreate command for master key works on the same setup.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -24,10 +24,12 @@
 
 namespace OCA\Files\Command;
 
+use OC\Encryption\Manager;
 use OC\Files\Filesystem;
 use OC\Files\View;
 use OCP\Files\FileInfo;
 use OCP\Files\Mount\IMountManager;
+use OCP\ILogger;
 use OCP\IUserManager;
 use OCP\Share\IManager;
 use OCP\Share\IShare;
@@ -48,6 +50,12 @@ class TransferOwnership extends Command {
 
 	/** @var IMountManager */
 	private $mountManager;
+
+	/** @var Manager  */
+	private $encryptionManager;
+
+	/** @var ILogger  */
+	private  $logger;
 
 	/** @var FileInfo[] */
 	private $allFiles = [];
@@ -70,10 +78,12 @@ class TransferOwnership extends Command {
 	/** @var string */
 	private $finalTarget;
 
-	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager) {
+	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ILogger $logger) {
 		$this->userManager = $userManager;
 		$this->shareManager = $shareManager;
 		$this->mountManager = $mountManager;
+		$this->encryptionManager = $encryptionManager;
+		$this->logger = $logger;
 		parent::__construct();
 	}
 
@@ -260,7 +270,9 @@ class TransferOwnership extends Command {
 				$this->finalTarget = $this->finalTarget . '/' . basename($sourcePath);
 			}
 		}
+
 		$view->rename($sourcePath, $this->finalTarget);
+
 		if (!is_dir("$this->sourceUser/files")) {
 			// because the files folder is moved away we need to recreate it
 			$view->mkdir("$this->sourceUser/files");
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -86,6 +86,11 @@ class Encryption extends Wrapper {
 	/** @var  ArrayCache */
 	private $arrayCache;
 
+	/** @var array which has information of sourcePath during rename operation */
+	private $sourcePath;
+
+	private static $disableWriteEncryption = false;
+
 	/**
 	 * @param array $parameters
 	 * @param IManager $encryptionManager
@@ -201,16 +206,18 @@ class Encryption extends Wrapper {
 	 */
 	public function file_get_contents($path) {
 
-		$encryptionModule = $this->getEncryptionModule($path);
+		if ($this->encryptionManager->isEnabled() !== false) {
+			$encryptionModule = $this->getEncryptionModule($path);
 
-		if ($encryptionModule) {
-			$handle = $this->fopen($path, "r");
-			if (!$handle) {
-				return false;
+			if ($encryptionModule) {
+				$handle = $this->fopen($path, "r");
+				if (!$handle) {
+					return false;
+				}
+				$data = stream_get_contents($handle);
+				fclose($handle);
+				return $data;
 			}
-			$data = stream_get_contents($handle);
-			fclose($handle);
-			return $data;
 		}
 		return $this->storage->file_get_contents($path);
 	}
@@ -358,12 +365,11 @@ class Encryption extends Wrapper {
 	 *
 	 * @param string $path
 	 * @param string $mode
-	 * @param string|null $sourceFileOfRename
 	 * @return resource|bool
 	 * @throws GenericEncryptionException
 	 * @throws ModuleDoesNotExistsException
 	 */
-	public function fopen($path, $mode, $sourceFileOfRename = null) {
+	public function fopen($path, $mode) {
 
 		// check if the file is stored in the array cache, this means that we
 		// copy a file over to the versions folder, in this case we don't want to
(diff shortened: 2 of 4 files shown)


apps/files/lib/Command/TransferOwnership.php
```diff
@@ -260,7 +270,18 @@ protected function transfer(OutputInterface $output) {
 				$this->finalTarget = $this->finalTarget . '/' . basename($sourcePath);
 			}
 		}
+		/**
+		 * If encryption is enabled and masterkey is the option selected
+		 * kindly use the CustomView wrapper.
+		 */
+		/*if ($this->encryptionManager->isEnabled() &&
```
PVince81: please remove

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -86,6 +86,10 @@ class Encryption extends Wrapper {
 	/** @var  ArrayCache */
 	private $arrayCache;
 
+	private $sourcePath;
+
+	private $decryptedFile = false;
```
PVince81: so if I understand correctly, this flag only affects write operations ?

If yes, call it `$disableWriteEncryption` instead and the methods `setDisableWriteEncryption()`

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -622,6 +644,22 @@ public function moveFromStorage(Storage $sourceStorage, $sourceInternalPath, $ta
 		return $result;
 	}
 
+	/**
+	 * Set the flag to true, so that the file would be
+	 * in the decrypted state.
+	 */
+	public function setDecryptedFileFlag() {
```
PVince81: Just pass the flag as argument, no need to write two methods...

lib/private/Files/View.php
```diff
@@ -2124,4 +2145,14 @@ private function createParentDirectories($filePath) {
...
 		return true;
 	}
+
+	/**
+	 * This flag will help to decide whether to keep the file
+	 * in a decrypted state. Based on this flag's true value
+	 * the storage's set value for decrypted file flag is called.
+	 */
+
+	public function setDecryptedFile() {
```
PVince81: I'd prefer that the View doesn't know anything about encryption. Encryption is just like a plugin that inserts itself so it's not good that the core code like View relies on encryption specific things.

Let's make flag static in the Encryption class above instead and set it from outside. That should do for now.

PVince81 (commented):
great progress, see comments

let's use the static approach

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -451,14 +456,31 @@ public function fopen($path, $mode, $sourceFileOfRename = null) {
...
 				if (!is_resource($source)) {
 					return false;
 				}
+
 				$handle = \OC\Files\Stream\Encryption::wrap($source, $path, $fullPath, $header,
 					$this->uid, $encryptionModule, $this->storage, $this, $this->util, $this->fileHelper, $mode,
-					$size, $unencryptedSize, $headerSize, $signed, $sourceFileOfRename);
+					$size, $unencryptedSize, $headerSize, $signed, $this->sourcePath[$path]);
+
+				if (isset($this->sourcePath[$path])) {
```
PVince81: you need to call isset earlier, else line 478 is going to throw a warning when not set

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -748,10 +786,11 @@ private function copyBetweenStorage(Storage $sourceStorage, $sourceInternalPath,
 				$source = $sourceStorage->fopen($sourceInternalPath, 'r');
 				if ($isRename) {
 					$absSourcePath = Filesystem::normalizePath($sourceStorage->getOwner($sourceInternalPath). '/' . $sourceInternalPath);
-					$target = $this->fopen($targetInternalPath, 'w', $absSourcePath);
+					$this->sourcePath[$targetInternalPath] = $absSourcePath;
 				} else {
-					$target = $this->fopen($targetInternalPath, 'w');
+					$this->sourcePath[$targetInternalPath] = null;
```
PVince81: use `unset` for maximum deletion

sharidas:
- [x] Removed changes from View which points to encryption. 
- [x] Removed the comment ( code comment ) 
- [x] Renamed the variable `$decryptedFile` to `$disableWriteEncryption` and also renamed the function associated with the operation to `setDisableWriteEncryption()`. Argument is passed to `setDisableWriteEncryption()` for operation set or unset.
- [x]

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -451,14 +456,35 @@ public function fopen($path, $mode, $sourceFileOfRename = null) {
...
 			if ($shouldEncrypt === true && $encryptionModule !== null) {
+				/**
+				 * The check of $disableWriteEncryption, required to get the file in the decrypted state.
+				 * It will help us get the normal file handler. And hence we can re-encrypt
+				 * the file when necessary, later. The true/false of $getDecryptedFile decides whether
+				 * to keep the file decrypted or not. The intention is to get the data decrypt
+				 * for write mode.
+				 */
+				if (self::$disableWriteEncryption && ($mode !== 'r')) {
+					return fopen($this->storage->getSourcePath($path), $mode);
```
PVince81: This is going to break horribly with external storage because you are doing a fopen directly on the filesystem.

Use `$this->getWrapperStorage()->fopen()` instead

lib/private/Files/Storage/Wrapper/Encryption.php
PVince81: @sharidas did you miss this comment ?

sharidas: No, I thought in a different way. `wrap()` expects either null or string `$this->sourcePath[$path]`. So if its not set then I set as null at https://github.com/owncloud/core/pull/28774/files#diff-938b56891f3079e628cfda1c331911b7R476 .
Else if its set then unset at line 484. So is there something really wrong here?

lib/private/Files/Storage/Wrapper/Encryption.php
sharidas: @PVince81 Updated in my latest PR. Thanks for pointing it out.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -451,14 +456,34 @@ public function fopen($path, $mode, $sourceFileOfRename = null) {
...
 				$headerSize = $this->getHeaderSize($path);
 				$source = $this->storage->fopen($path, $mode);
 				if (!is_resource($source)) {
 					return false;
 				}
-				$handle = \OC\Files\Stream\Encryption::wrap($source, $path, $fullPath, $header,
-					$this->uid, $encryptionModule, $this->storage, $this, $this->util, $this->fileHelper, $mode,
-					$size, $unencryptedSize, $headerSize, $signed, $sourceFileOfRename);
+
+				if (isset($this->sourcePath[$path])) {
```
sharidas: Hope this looks better.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -622,6 +647,15 @@ public function moveFromStorage(Storage $sourceStorage, $sourceInternalPath, $ta
 		return $result;
 	}
 
+	/**
+	 * Set the flag to true, so that the file would be
+	 * in the decrypted state.
+	 *
+	 * @param $isDisabled bool
+	 */
+	public function setDisableWriteEncryption($isDisabled) {
```
PVince81: function must be static

sharidas: Done.

sharidas:
- [x] Made the function setDisableWriteEncryption staic.
- [x] Used a local variable to pass the `$this->sourcePath[$path]` if its set else null to -> `wrap()`
- [x] Removed the view code

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -86,6 +86,10 @@ class Encryption extends Wrapper {
 	/** @var  ArrayCache */
 	private $arrayCache;
 
+	private $sourcePath;
```
PVince81: PHPDoc comments please

PVince81 (approved):
:+1: add missing PHP doc then this is good to go

PVince81:

sharidas:
Backport of this PR: https://github.com/owncloud/core/pull/28845

PR #28787, opened 2017-08-24
added better description

dercorn:
I added a more verbose description to this command.

### DIFF
--- a/apps/files/lib/Command/Scan.php
+++ b/apps/files/lib/Command/Scan.php
@@ -61,7 +61,7 @@ class Scan extends Base {
 
 		$this
 			->setName('files:scan')
-			->setDescription('rescan filesystem')
+			->setDescription('Scans the filesystem for new files and updates the file cache accordingly.')
 			->addArgument(
 				'user_id',
 				InputArgument::OPTIONAL | InputArgument::IS_ARRAY,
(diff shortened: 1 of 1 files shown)


apps/files/lib/Command/Scan.php
```diff
@@ -61,7 +61,7 @@ protected function configure() {
 
 		$this
 			->setName('files:scan')
-			->setDescription('rescan filesystem')
+			->setDescription('Scans the filesystem for new files and updates the file cache accordingly.')
```
DeepDiver1975: actually it's not only new files but also changed and deleted files and folders ....

DeepDiver1975 (dismissed):

apps/files/lib/Command/Scan.php
DeepDiver1975: And 'update file cache might be a bit too technical' ... but I have no better idea right now ..

dercorn: How about "Scans the filesystem for changes and updates the file cache accordingly."?
"file cache" should stay imo, since that is just our terminology.

dercorn:
@DeepDiver1975 I changed the text a little. Is it okay for you now?

tomneedham (approved):

dercorn:
@tomneedham can you give Travis another nudge?

PVince81:
@dercorn please backport to stable10

DeepDiver1975:
already in stable10 - https://github.com/owncloud/core/commit/7bfb5c999a0e9cd3cc2bca5d1f30ef42a124df71

PR #28790, opened 2017-08-24
[stable10] Handle BSD case for 32 bit filemtime and install warning

phil-davis:
Backport #28759 
because it makes the setup warning more generic so it applies when on any non-linux-style platform.


PVince81 (approved):
:+1:

PR #28792, opened 2017-08-24
[WIP] stable9 debug

butonic:
need to test if the additional debug checks are correct. go travis, go jenkins

### DIFF
--- a/apps/files_sharing/api/local.php
+++ b/apps/files_sharing/api/local.php
@@ -635,6 +635,10 @@ class Local {
 			return null;
 		}
 		if ($share = $result->fetchRow()) {
+			if ($unexpected = $result->fetchRow()) {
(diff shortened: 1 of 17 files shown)


DeepDiver1975:
stable9? its EOL? just asking

butonic:
I am trying to find the query that does not read all results, leading to a
> SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

wanted the Jenkins to test these changes because my dev machine needs ... drying and my current system is not setup to run on all dbs and with ldap ...

In any case ... I hope to find the problematic queris at a larger galera cluster with maxscale rw split using this patch. Not relevant for 10.0.3 ... removing milestone

PVince81:
stable9 is EOL and stable9.1 will be soon, close ?

Issue #15407, opened 2015-04-04
Small wait timeout of MySQL kills many file operations - General error: 2006 MySQL server has gone away


MartinNowack:
### Steps to reproduce
1. Set 'wait_timeout' to a *small* value e.g. 60 seconds (default is 28800 seconds)
2. Try to upload a moderate sized file (>5 MB) with a slow upstream connection

### Expected behaviour
The file should be uploaded successfully. 

### Actual behaviour
The file is not uploaded

### Server configuration
**Operating system**: Linux

**Web server:** Apache

**Database:** Mysql 5.1

**PHP version:** 5.5

**ownCloud version:** (see ownCloud admin page): 8.0.2

**Updated from an older ownCloud or fresh install:**
updated

**Are you using external storage, if yes which one:** local/smb/sftp/...
No

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
No

#### ownCloud log (data/owncloud.log)
```
SELECT `fileid` FROM `oc_filecache` WHERE `storage` = ? AND `path_hash` = ?' with params [\"2\", \"0fea6a13c52b4d4725368f24b045ca84\"]:\n\nSQLSTATE[HY000]: General error: 2006 MySQL server has gone away
```

# More Detailed Description
I use a hosted web service where I can't change the MySQL timeout. The value `wait_timeout` of 60 seconds sounds quite reasonable for me.
I'm using the DesktopClient version 1.8, I have a very small upstream ~50KiB/s and I couldn't upload bigger files (>2MB).

## I guess the following happens:
1. Files are uploaded into chunks on client side
2. The server opens an SQL connection
3. a) it waits for one chunk, and 
3. b) updates the cache information in the data base (SELECT `fileid` FROM `oc_filecache` WHERE `storage` = ? AND `path_hash` = ?')

The problem is now, if the time it takes to receive a chunk is bigger than the timeout for the database connection, it fails.

According to https://github.com/owncloud/client/issues/766 the default chunk size is 5MB. If I set the OWNCLOUD_CHUNK_SIZE to 1MB on the client side, it works, as the timeout is not triggered.

But this solution is cumbersome, the server should suggest which size it can work with or which time it has maximum. At least the error message should be changed to be more reasonable.

nickvergessen:
The error message is from the MySQL server, not from owncloud.
[...]

PR #28794, opened 2017-08-24
make ping() available, reconnect after assemble

butonic:
fixes mysql server gone away on huge chunked file upload assemblys by using doctrines ping(). also see https://github.com/owncloud/core/issues/15407#issuecomment-135458449

alternative might be to use
```
'dbdriveroptions' => array(PDO::MYSQL_ATTR_INIT_COMMAND => 'SET wait_timeout = 28800'),
```

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -453,8 +453,8 @@ class File extends Node implements IFile {
 					/** @var \OC\Files\Storage\Storage $targetStorage */
 					list($partStorage, $partInternalPath) = $this->fileView->resolvePath($partFile);
 
-
 					$chunk_handler->file_assemble($partStorage, $partInternalPath);
+					$this->tryReconnectIfDBConnectionTimedOut();
 
 					if (!self::isChecksumValid($partStorage, $partInternalPath)) {
 						throw new BadRequest('The computed checksum does not match the one received from the client.');
@@ -478,6 +478,7 @@ class File extends Node implements IFile {
 				} else {
 					// assemble directly into the final file
 					$chunk_handler->file_assemble($targetStorage, $targetInternalPath);
+					$this->tryReconnectIfDBConnectionTimedOut();
 				}
 
 				// allow sync clients to send the mtime along in a header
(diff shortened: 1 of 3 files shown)


apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -453,9 +453,15 @@ private function createFileChunked($data) {
 					/** @var \OC\Files\Storage\Storage $targetStorage */
 					list($partStorage, $partInternalPath) = $this->fileView->resolvePath($partFile);
 
-
 					$chunk_handler->file_assemble($partStorage, $partInternalPath);
 
+					//assembly might take a while, double check db connection is still alive
+					$conn = \OC::$server->getDatabaseConnection();
+					if ($conn->ping() === false) {
+						$conn->close();
```
DeepDiver1975: are we in a transaction here? This will result in inconsistent state ...

lib/public/IDBConnection.php
```diff
@@ -284,4 +284,28 @@ public function getTransactionIsolation();
...
+	 * @example
+	 *
+	 *   if ($conn->ping() === false) {
+	 *      $conn->close();
+	 *      $conn->connect();
+	 *   }
+	 *
+	 * It is undefined if the underlying driver attempts to reconnect
+	 * or disconnect when the connection is not available anymore
+	 * as long it returns TRUE when a reconnect succeeded and
```
DeepDiver1975: Is the purpose of ping() really a reconnect?

DeepDiver1975 (changes requested):

lib/public/IDBConnection.php
butonic: Copy'n'Paste from their phpdoc: http://www.doctrine-project.org/api/dbal/2.5/source-class-Doctrine.DBAL.Connection.html#1554-1574

DeepDiver1975: Okay - there is a connect in the ping() implementation.... hmmmm

apps/dav/lib/Connector/Sabre/File.php
DeepDiver1975: I'd throw an exception in here - we are in an unclear state. The request has to be performed again.

DeepDiver1975:
> alternative might be to use

To me this is not the alternative but the proper fix ... don't let the connection being closed.

butonic:
leaving the db connection open for 8h is hardly the right thing todo. anyway, I added a suggestion in case we are in a transaction (which as of today we are not when assembling a file). If we see timeouts in other places we might move this to our Connection implementation and add it to the IDBConnection interface so that the same mechanism can be used in other places.

DeepDiver1975:
TBH: we should remove the old chunking .....

DeepDiver1975:
> please, not in a hurry

for sure not - no worries ;-)

PVince81:
Is this required for new chunking as well ? Remember that newer desktop client versions will use new dav chunking now.

PVince81:
note that final assembly in new dav is also going through File::put()

PVince81:
@butonic finish or defer ?

PVince81:
moving to "planned", has been reported not fixing the issue yet

DeepDiver1975:
@butonic conflicts - please resolve

PVince81:
conflicts.

is this still relevant ? else close

DeepDiver1975:
feelr free to revive any time ....

PR #28796, opened 2017-08-24
fix selecting parent

butonic:
guess what is wrong here: https://github.com/owncloud/core/blob/master/lib/private/Share/Share.php#L1901-L1902


DeepDiver1975 (approved):
that code has to die .....

DeepDiver1975:
@butonic backport?

butonic:
@DeepDiver1975 no backport needed. we can do it later if we find out it causes problems in the wild.

PVince81:
@butonic this is the old sharing code, is it really still triggered on OC 10 ?

PVince81:
@butonic any linked bug report ? the code change doesn't explain what the problem is, maybe add a PHP doc to explain what it is doing and why ?

PVince81:
closing due to lack of feedback and context

PR #28802, opened 2017-08-25
Allow User backends to indicate internal problems to skip users and continue syncing

butonic:
To prevent individual users with malformed or missing attributes from breaking the sync this PR 
will catch IndexOutOfBounds exceptions, log them and continue with the next user.
- [ ] for 10.0.4 also show the problem in the console so the admin can see something went wrong without checking the log

Furthermore, this PR adds the IProvidesAvatarBackend but allows skipping them by setting a config option.


butonic:
noooooo

PR #27703, opened 2017-04-20
Reload file list when leaving hidden state

PVince81:
## Description
Whenever a file list is already initialized and was hidden when
switching to another file list in the navigation bar, if the user comes
back to this list it gets redisplayed. At this point the list needs to
be refreshed to be able to reflect any potential file changes done from
the other lists.

## Related Issue
Fixes https://github.com/owncloud/core/issues/27697

## Motivation and Context
See description and ticket

## How Has This Been Tested?
See ticket steps

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Please review @jvillafanez @noveens

PR #28804, opened 2017-08-25
Only reload file list when switching navigation sections

### LINKED ISSUE #27890: File list loaded twice
Since https://github.com/owncloud/core/pull/27703 the file list is loaded twice.

When on root: the root folder is loaded twice.
When the URL points at a subdir, first the root folder is loaded and then the subdir.
This causes a quick UI flicker where the user sees the contents of the root folder before it gets replaced by the contents of the correct folder.

Will need a bit of time to fix this properly. The PR is already a bit of a hack and need to make sure all other cases continue working.

PVince81:
## Description
Reverts a previous bad fix from https://github.com/owncloud/core/pull/27703 which causes a double reload and even breaks sometimes when opening a private link by reloading the wrong list.

Adds a better fix that only reloads a file list whenever the active navigation section has changed, because that's when we need to refresh, for example after restoring a file from trash, the "All files" list needs to refresh to display the restored folder.

Also prevents initial double reload of the file list.

## Related Issue
Fixes https://github.com/owncloud/core/issues/27890

## How Has This Been Tested?
- create folder "test" in "All files", delete it. go to trash: folder appears
- create folder "test/sub" in "All files". Refresh page several times: folder contents is still correct, no blinking. (before this fix it was either blinking or showing the wrong contents)

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/js/app.js
+++ b/apps/files/js/app.js
@@ -217,7 +217,8 @@
 			if (e && e.itemId) {
 				params = {
 					view: e.itemId,
-					dir: '/'
+					dir: '/',
+					force: true
 				};
 				this._changeUrl(params.view, params.dir);
 				OC.Apps.hideAppSidebar($('.detailsView'));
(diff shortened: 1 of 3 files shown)


davitol:
👍

phil-davis:
Fixes the "flashing load" of the home folder, before then displaying the sub-folder content.

phil-davis (approved):
Approving because it works. Would be another fix of an annoying little thing that can be got into the next beta maybe.

PVince81:
stable10: https://github.com/owncloud/core/pull/28843

guruz:
yay nice. I was annoyed this breaks the scrollTo private linkls.. hope this fixes it

PR #28805, opened 2017-08-25
skip to next user on OutOfBoundsException

butonic:
on large ldap instances a missing uuid or username currently aborts the whole sync. This PR allows backends to indicate these problems and fail with an OutOfBoundsException. For now core will log the exception and continue with the next user.

In the future we should expose this to the cli.

together with https://github.com/owncloud/user_ldap/pull/125 this skips problematic ldap users

### DIFF
--- a/lib/private/User/SyncService.php
+++ b/lib/private/User/SyncService.php
@@ -100,24 +100,30 @@ class SyncService {
 			// update existing and insert new users
 			foreach ($users as $uid) {
 				try {
-					$a = $this->mapper->getByUid($uid);
-					if ($a->getBackend() !== $this->backendClass) {
-						$this->logger->warning(
-							"User <$uid> already provided by another backend({$a->getBackend()} != {$this->backendClass}), skipping.",
-							['app' => self::class]
-						);
-						continue;
+					try {
+						$a = $this->mapper->getByUid($uid);
+						if ($a->getBackend() !== $this->backendClass) {
+							$this->logger->warning(
+								"User <$uid> already provided by another backend({$a->getBackend()} != {$this->backendClass}), skipping.",
+								['app' => self::class]
+							);
+							continue;
+						}
+						$a = $this->setupAccount($a, $uid);
+						$this->mapper->update($a);
+					} catch (DoesNotExistException $ex) {
+						$a = $this->createNewAccount($uid);
+						$this->setupAccount($a, $uid);
+						$this->mapper->insert($a);
 					}
-					$a = $this->setupAccount($a, $uid);
-					$this->mapper->update($a);
-				} catch(DoesNotExistException $ex) {
-					$a = $this->createNewAccount($uid);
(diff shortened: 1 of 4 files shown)


lib/private/User/SyncService.php
```diff
@@ -100,24 +100,30 @@ public function run(\Closure $callback) {
...
 					}
-					$a = $this->setupAccount($a, $uid);
-					$this->mapper->update($a);
-				} catch(DoesNotExistException $ex) {
-					$a = $this->createNewAccount($uid);
-					$this->setupAccount($a, $uid);
-					$this->mapper->insert($a);
+					// clean the user's preferences
+					$this->cleanPreferences($uid);
+				} catch (\OutOfBoundsException $e) {
```
jvillafanez: OutOfBounds is a bad naming choice if the exception is expected to be thrown when we can access to the property. I'd rather use a custom public exception here.

lib/private/User/SyncService.php
```diff
@@ -100,24 +100,30 @@ public function run(\Closure $callback) {
...
 			foreach ($users as $uid) {
 				try {
-					$a = $this->mapper->getByUid($uid);
-					if ($a->getBackend() !== $this->backendClass) {
-						$this->logger->warning(
-							"User <$uid> already provided by another backend({$a->getBackend()} != {$this->backendClass}), skipping.",
-							['app' => self::class]
-						);
-						continue;
+					try {
```
jvillafanez: maybe we could move this try / catch block to another function. This way we can avoid this nested try / catch, and also make this function more compact.

PVince81:
@butonic any update ? See comments above.

Else would defer to "planned" for the next release.

PVince81:
moving to "planned" as discussed with @DeepDiver1975

lib/private/User/SyncService.php
```diff
@@ -100,24 +100,30 @@ public function run(\Closure $callback) {
...
+					// clean the user's preferences
+					$this->cleanPreferences($uid);
+				} catch (\OutOfBoundsException $e) {
+					// if an attribute could not be determined
+					// continue syncing other users
+					// TODO expose on CLI to show the admin running the occ command the problem, maybe additional $output parameter?
+					$this->logger->logException($e,	['app' => self::class]);
 				}
-				// clean the user's preferences
-				$this->cleanPreferences($uid);
```
DeepDiver1975: why is this removed?

DeepDiver1975 (changes requested):

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

butonic:
closing, as the current code already continues on all Exceptions. logging errors to the console tracked in https://github.com/owncloud/core/issues/31310

phil-davis:
``backport-request`` label removed, since that is not relevant to a closed PR.

PR #28809, opened 2017-08-28
UI tests for delete feature

individual-it:
## Description
UI test for deleting files and folder

## Related Issue
owncloud/QA#437

## Motivation and Context
reduce manual test time by running the smoke test automatically

## How Has This Been Tested?
run tests locally and in travis

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #28811, opened 2017-08-28
[stable10] UI tests for delete feature

phil-davis:
Backport #28809

### DIFF
--- a/tests/ui/features/files/delete.feature
+++ b/tests/ui/features/files/delete.feature
@@ -1,0 +1,71 @@
+Feature: delete
+
+	Background:
+		Given a regular user exists
+		And I am logged in as a regular user
+		And I am on the files page
+
+	Scenario: Delete files & folders one by one and check its existence after page reload
+		When I delete the folder "simple-folder"
+		Then the folder "simple-folder" should not be listed
+		When I delete the file "lorem.txt"
+		Then the file "lorem.txt" should not be listed
+		When I delete the folder "strängé नेपाली folder"
+		Then the folder "strängé नेपाली folder" should not be listed
+		When I delete the file "strängé filename (duplicate #2).txt"
+		Then the file "strängé filename (duplicate #2).txt" should not be listed
+		When the files page is reloaded
+		Then the folder "simple-folder" should not be listed
+		And the folder "strängé नेपाली folder" should not be listed
+		And the file "lorem.txt" should not be listed
+		And the file "strängé filename (duplicate #2).txt" should not be listed
+
+	Scenario: Delete a file with problematic characters
+		When I rename the following file to
+			|from-name-parts |to-name-parts  |
+			|lorem.txt       |'single'       |
+			|                |"double" quotes|
+			|                |question?      |
+			|                |&and#hash      |
+		And the files page is reloaded
+		Then the following file should be listed
+			|name-parts     |
+			|'single'       |
+			|"double" quotes|
+			|question?      |
+			|&and#hash      |
+		And I delete the following file
+			|name-parts     |
+			|'single'       |
+			|"double" quotes|
+			|question?      |
+			|&and#hash      |
+		Then the following file should not be listed
+			|name-parts     |
+			|'single'       |
+			|"double" quotes|
+			|question?      |
+			|&and#hash      |
+		And the files page is reloaded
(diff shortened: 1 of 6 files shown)


tests/ui/features/files/delete.feature
```diff
@@ -0,0 +1,71 @@
+Feature: delete
+
+	Background:
+		Given a regular user exists
+		And I am logged in as a regular user
+		And I am on the files page
+
+	Scenario: Delete files & folders one by one and check its existence after page reload
+		When I delete the folder "simple-folder"
```
SergioBertolinSG: There should be only one when per scenario. many "Given" one "When" and many "Then"

phil-davis: This is true in theory. There should be some "Given" steps to set things up (they might all be in the Background: like here), then a set of "When" (with maybe a few steps to carry out the actions), then a set of "Then" that check the result.

So we could easily make separate scenarios here.

The practical side for UI tests, is that the "test setup" phase - creating user, logging in, navigating to the desired page... - takes so long. So to speed it up, it is practical to make multiple when-then in a scenario.

Also, sometimes there are genuinely mixed sequences of when-then needed. e.g. even if only 1 file delete is tested here, we want to:
1) when "delete the file"
2) then "see if it is gone from the local page"
3) when "reload the page"
4) then "see if it really is gone"
because if the Javascript/AJAX is not working properly, (2) and/or (4) might fail.

phil-davis: Here is a sample discussion of this issue, where there is a "workflow" to be acceptance tested:
https://stackoverflow.com/questions/12060011/is-it-acceptable-to-write-a-given-when-then-when-then-test-in-gherkin
and of course there are different opinions on how best to do it.

SergioBertolinSG: I don't like this approach, it is like a test script. You could use given-and as preconditions, or even hiding those thens after the whens by checking it directly in the code.

There should be only one thing marked as when, rest are preconditions or postconditions.

phil-davis: The step sequence could be re-ordered so the "when" all come first:
```
	Scenario: Delete files & folders one by one and check its existence after page reload
		When I delete the folder "simple-folder"
		And I delete the file "lorem.txt"
		And I delete the folder "strängé नेपाली folder"
		And I delete the file "strängé filename (duplicate #2).txt"
		Then the folder "simple-folder" should not be listed
		And the file "lorem.txt" should not be listed
		And the folder "strängé नेपाली folder" should not be listed
		And the file "strängé filename (duplicate #2).txt" should not be listed
		When the files page is reloaded
		Then the folder "simple-folder" should not be listed
		And the folder "strängé नेपाली folder" should not be listed
		And the file "lorem.txt" should not be listed
		And the file "strängé filename (duplicate #2).txt" should not be listed
```
the last "When the files page is reloaded" could be renamed to "And the files page is reloaded" - that just "hides" the fact that it is an action, and makes no real difference.

Maybe it could be cut down to:
```
	Scenario: Delete files & folders one by one and check its existence after page reload
		When I delete the folder "simple-folder"
		And I delete the file "lorem.txt"
		And I delete the folder "strängé नेपाली folder"
		And I delete the file "strängé filename (duplicate #2).txt"
		Then the deleted items should not be listed
```
and the code can remember which files have been deleted, and code for ``Then the deleted items should not be listed`` can check for the files on the current page, then reload the page and check again.

phil-davis: Anyway, this is a back-port of a commit in master. Whatever we change needs to be changed in master and also back-ported.

SergioBertolinSG: Maybe it could be a table

                When I delete the elements
                  | simple-folder |
                  | lorem.txt |
                     (...)
		Then the deleted items should not be listed

If the difference between folder and element is required, using another column detailing folder/file.

individual-it (approved):

tests/ui/features/files/delete.feature
individual-it: @SergioBertolinSG Let me merge this as its only a backport and I will do a new PR against master

SergioBertolinSG: :+1:

PR #28813, opened 2017-08-28
Make occ upgrade verbose by default

PVince81:
## Description
This will always output timestamps and also mention each market app that
is being checked through the marketplace.

The previous behavior can be set using `--verbose=0`. But I think the verbose mode will be useful in any case.

## Related Issue
None

## Motivation and Context
Often times admin just run `occ upgrade` thinking there will be no error and then it is too late to set a more verbose mode. Setting verbose mode by default also provides more useful info like where each app is being checked against the marketplace, so it looks more like progress than having no output for many seconds.

## How Has This Been Tested?
Run `occ upgrade` with a few apps to be updated. You'll see entries about checking app updates against the marketplace, for each app individually.

## Types of changes

## Checklist:

### DIFF
--- a/core/Command/Upgrade.php
+++ b/core/Command/Upgrade.php
@@ -84,6 +84,11 @@ class Upgrade extends Command {
 	 * @param OutputInterface $output output interface
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
+		if ($output->getVerbosity() === OutputInterface::VERBOSITY_NORMAL
+			&& !$input->hasParameterOption('--verbose=0', true)) {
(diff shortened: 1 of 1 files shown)


tomneedham:
Tested 👍

tomneedham:
How does this work with `-q` ?

PVince81:
@tomneedham with `-q` the verbosity would not be `OutputInterface::VERBOSITY_NORMAL` so the code block doesn't do anything.

patrickjahns (approved):
Tested 👍

PVince81:
stable10: https://github.com/owncloud/core/pull/28876

Issue #28780, opened 2017-08-23
Exceptions related with encryption using external storage without encryption.


SergioBertolinSG:
### Steps to reproduce

0. Set up an external storage. For example SFTP.
1. In oC files view create three or more folders and upload several files inside them.
2. Delete one of the created folders.

### Expected behaviour
Removed folder is deleted.

### Actual behaviour
Removed folder is deleted but log is full of exceptions related with encryption.

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

**ownCloud version:** (see ownCloud admin page)
(current master not stable10)
{"installed":"true","maintenance":"false","needsDbUpgrade":"false","version":"10.0.3.0","versionstring":"10.0.3 beta","edition":"Community","productname":"ownCloud"}

**Updated from an older ownCloud or fresh install:**
Fresh

**The content of config/config.php:**

```

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
Yes. SFTP

**Are you using encryption:** 
No.

**Logs**
```
{"reqId":"ytrkb2h2NhnHSBshgTzK","level":2,"time":"2017-08-23T11:51:46+00:00","remoteAddr":"HIDDEN","user":"admin","app":"no app in context","method":"DELETE","url":"\/remote.php\/dav\/files\/admin\/SFTP\/FOLDER2","message":"Encryption module \"\" not found, file will be stored unencrypted (Module with id: OC_DEFAULT_MODULE does not exist.)"}
{"reqId":"ytrkb2h2NhnHSBshgTzK","level":3,"time":"2017-08-23T11:51:46+00:00","remoteAddr":"HIDDEN","user":"admin","app":"no app in context","method":"DELETE","url":"\/remote.php\/dav\/files\/admin\/SFTP\/FOLDER2","message":"Encryption module defined in \"FOLDER2\/DCSkKMLWsAA86Bi.jpg\" not loaded!"}
{"reqId":"ytrkb2h2NhnHSBshgTzK","level":3,"time":"2017-08-23T11:51:46+00:00","remoteAddr":"HIDDEN","user":"admin","app":"no app in context","method":"DELETE","url":"\/remote.php\/dav\/files\/admin\/SFTP\/FOLDER2","message":"Couldn't re-calculate unencrypted size for FOLDER2\/DCSkKMLWsAA86Bi.jpg"}
[...]

PR #28820, opened 2017-08-28
Increment required only when encryption is enabled

sharidas:
Increment to files in filecache is required only when
encryption is enabled. Else use the version as it is.

## Description

When encryption is not enabled then no need to go through each and every step in encryption wrapper's fopen. Instead we can get another storage wrapper's fopen handler.

## Related Issue

https://github.com/owncloud/core/issues/28780

## Motivation and Context

This would help resolve the exceptions thrown in the logs.

## How Has This Been Tested?

- [x] Create external storage with SFTP
- [x] Create a folder `test` under shared folder.
- [x] upload a file under `test`
- [x] Delete the test folder
- [x] Try to access the `test` folder from the trash
- [x] No exception logs found

Also I have noted that during verification on my local machine the storage passes to local storage from the encryption wrapper ( when no encryption is enabled ).

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -385,7 +385,7 @@ class Encryption extends Wrapper {
 		$header = $this->getHeader($path);
 		$signed = (isset($header['signed']) && $header['signed'] === 'true') ? true : false;
 		$fullPath = $this->getFullPath($path);
-		$encryptionModuleId = ($encryptionEnabled) ? $this->util->getEncryptionModuleId($header): "";
+		$encryptionModuleId = $this->util->getEncryptionModuleId($header);
 
 		if ($this->util->isExcluded($fullPath) === false) {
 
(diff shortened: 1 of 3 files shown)


PVince81:
One reason why we still set the encryption wrapper even when encryption is disabled is mostly because we want to be able to detect formerly encrypted files and avoid sending these back.

So if we want to add such bypass we might need to at least put it after the encryption check, which might be when reading the header.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -937,9 +943,13 @@ protected function getHeader($path) {
 				// if the header was empty we have to check first if it is a encrypted file at all
 				// We would do query to filecache only if we know that entry in filecache exists
 				$info = $this->getCache()->get($path);
+				$encryptionModule = '';
```
sharidas: Made alterations at the getHeader method. The intention here was to pass encryption module wrapped in array if encryption is enabled. And this value is later used https://github.com/owncloud/core/pull/28820/files#diff-938b56891f3079e628cfda1c331911b7R381 to call different storage's fopen. I hope this is the same header which is mentioned at https://github.com/owncloud/core/pull/28820#issuecomment-325367258

SergioBertolinSG:
I see https://github.com/owncloud/core/issues/28780 fixed here.

PVince81 (approved):
:+1: code looks fine.

As discussed it would be good to adjust the trashbin tests to verify that the restored file's contents matches before deletion, and have this test run for encryption too (some encryption tests are disabled with trashbin)

Please also retest versions: restoring or downloading older versions of a file must still work on home and external storage

SergioBertolinSG (approved):
Integration tests part is ok

sharidas:
Verified that transfer ownership works with this change. Verified that upon deletion of the file after transfer, made some update, the versions are restored.

PVince81:

sharidas:
Backport : https://github.com/owncloud/core/pull/28880

PR #28822, opened 2017-08-28
Add passwordEnabled field to postHookData

sharidas:
Add passwordDisabled field to postHookData

## Description

Adding additional parameter passwordDisabled to the postHookData. This data would help us to determine if password is enabled for the share or not while creating the share.

## Motivation and Context

This would help us know if the share was created with password or not.

## How Has This Been Tested?

- [x] Create public folder share with password  - the logs should help us understand that password was enabled.
- [x] Create public folder share without password - the logs should help us understand that password wasn't enabled.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -644,6 +644,7 @@ class Manager implements IManager {
 			'shareWith' => $share->getSharedWith(),
 			'itemTarget' => $share->getTarget(),
 			'fileTarget' => $share->getTarget(),
+			'passwordEnabled' => (!is_null($share->getPassword()) and ($share->getPassword() !== '')),
 		];
 
 		\OC_Hook::emit('OCP\Share', 'post_shared', $postHookData);
--- a/tests/lib/Share20/ManagerTest.php
+++ b/tests/lib/Share20/ManagerTest.php
@@ -1785,6 +1785,7 @@ class ManagerTest extends \Test\TestCase {
 			'itemTarget' => '/target',
 			'fileTarget' => '/target',
 			'shareWith' => null,
+			'passwordEnabled' => true,
 		];
 
 		$hookListner->expects($this->once())


lib/private/Share20/Manager.php
```diff
@@ -644,6 +644,7 @@ public function createShare(\OCP\Share\IShare $share) {
 			'shareWith' => $share->getSharedWith(),
 			'itemTarget' => $share->getTarget(),
 			'fileTarget' => $share->getTarget(),
+			'passwordDisabled'=> $share->getPassword(),
```
PVince81: wouldn't this pass the password string instead of a boolean ?

call it "passwordEnabled" and set it to true if a password is set instead

sharidas:
Updated the PR to return true/false if password is enabed/disabled.

lib/private/Share20/Manager.php
```diff
@@ -644,6 +644,7 @@ public function createShare(\OCP\Share\IShare $share) {
 			'shareWith' => $share->getSharedWith(),
 			'itemTarget' => $share->getTarget(),
 			'fileTarget' => $share->getTarget(),
+			'passwordEnabled'=> !is_null($share->getPassword()) ? true : false,
```
PVince81: is it always null or can it sometimes be empty string after clearing ? please check

sharidas: if I replace `is_null` with `empty` then also it's working. 
When the password field is cleared after set the hook related to that is https://github.com/owncloud/core/blob/master/lib/private/Share20/Manager.php#L733 `post_update_password`.

phil-davis: The problem with ``empty()`` will be when someone sets the share password to "0". PHP is very annoying.

lib/private/Share20/Manager.php
```diff
@@ -644,6 +644,7 @@ public function createShare(\OCP\Share\IShare $share) {
 			'shareWith' => $share->getSharedWith(),
 			'itemTarget' => $share->getTarget(),
 			'fileTarget' => $share->getTarget(),
+			'passwordEnabled' => (!is_null($share->getPassword()) and ($share->getPassword() !== '')) ? true : false,
```
PVince81: no need for `true : false`, it is implicitly boolean when you have a condition

PVince81 (approved):
👍 otherwise

SergioBertolinSG (approved):
Fixes the issue.

sharidas:
Backport to stable10 of this PR: https://github.com/owncloud/core/pull/28827

PR #28825, opened 2017-08-29
make UI tests delete feature more readable

individual-it:
## Description
as discussed here https://github.com/owncloud/core/pull/28811#pullrequestreview-58874968 this PR makes the delete tests more readable and moves the single steps into PHP

`When I delete the elements` does check itself if the elements are disappearing from the WebUI. 
Plus there is an extra step to make sure the elemens do not come back after a page reload.

I suggest to leave the `Delete a file with problematic characters` Scenario as it is because of the complicated way we need to combine the single parts of that file name

## Motivation and Context
make better readable test

## How Has This Been Tested?
run UI tests locally, travis will test more

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/bootstrap/FilesContext.php
+++ b/tests/ui/features/bootstrap/FilesContext.php
@@ -34,6 +34,15 @@ require_once 'bootstrap.php';
 class FilesContext extends RawMinkContext implements Context
 {
 	private $filesPage;
+	
+	/**
+	 * Table of all files and folders that should have been deleted stored so
+	 * that other steps can use the list to check if the deletion happened correctly
+	 * table headings: must be: |name|
+	 * 
+	 * @var TableNode
+	 */
+	private $deletedElementsTable;
 
 	public function __construct(FilesPage $filesPage)
 	{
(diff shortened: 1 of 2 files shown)


SergioBertolinSG (approved):

phil-davis:
Hmmm, does this go too far? (combining the initial delete of all the files with the "real time" check that the JS has removed them from being listed)
The gherkin does look very compact!
@SergioBertolinSG @davitol your opinions?

davitol:
> The gherkin does look very compact!

It looks too compact to me too but I think it is fine for merging.

individual-it:
ideas for alternative step descriptions?

SergioBertolinSG:
I find it right. 

Just a bit nitpicking, I would use more indentation on the table and just one space before and after like:
| filename |

phil-davis:
I would have made it read:
```
When I delete the elements
Then the deleted elements should not be listed
And the deleted elements should not be listed after a page reload
```
that way the reader can understand that there is an action (deleting) followed by 2 checks.
At the moment the first check is hidden inside "When I delete the elements"

But I can cope the way it is also.

individual-it:
@phil-davis easy enough to implement. 
@davitol @phil-davis @SergioBertolinSG please review again

phil-davis (approved):

phil-davis:
I will make a backport in a moment...

Issue #28832, opened 2017-08-29
PHP array warning in response.php


phil-davis:
### Steps to reproduce
1. run limituploadbasedonsize.feature firewall test against current master
2. examine data/owncloud.log

### Expected behaviour
No PHP messages

### Actual behaviour
This:
```
{"reqId":"Jwtuaff38pjoU0rHVRR5","level":3,"time":"2017-08-29T11:23:10+00:00","remoteAddr":"::1","user":"admin","app":"PHP","method":"PROPFIND","url":"\/remote.php\/dav\/files\/admin\/","message":"in_array() expects parameter 2 to be array, null given at \/home\/phil\/core\/lib\/private\/legacy\/response.php#294"}
```
### Server configuration
**Operating system**:
Ubuntu 16.04
**Web server:**
PHP dev server
**Database:**
MySQL
**PHP version:**
7.0
**ownCloud version:** (see ownCloud admin page)
current master from git

seems to be the new code added in CORS stuff:
https://github.com/owncloud/core/commit/e3564e6c53bceb2dbe23dd1fbfec8b4696f7454d
https://github.com/owncloud/core/pull/28457

phil-davis:
Awaiting backport of #28833 which needs to happen with or after backport of #28457

DeepDiver1975:
fixed with the above PRs

phil-davis:
@SergioBertolinSG I restested this against ``stable10`` (as you asked about) and all is good.

PR #28833, opened 2017-08-29
setCorsHeaders() check if there are any allowedDomains

phil-davis:
## Description
setCorsHeaders() - handle the case when there are no allowedDomains

## Related Issue
#28832

## Motivation and Context
Seems to be only the log file message problem. I think the code ends up taking the desired path anyway.

## How Has This Been Tested?
Re-run the firewall test and see that the message in the log file is gone.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/legacy/response.php
+++ b/lib/private/legacy/response.php
@@ -291,7 +291,7 @@ class OC_Response {
 			$config = \OC::$server->getConfig();
 		}
 		$allowedDomains = json_decode($config->getUserValue($userId, 'core', 'domains'));
-		if (in_array($domain, $allowedDomains)) {
+		if (is_array($allowedDomains) && in_array($domain, $allowedDomains)) {
 			// TODO: infer allowed verbs from existing known routes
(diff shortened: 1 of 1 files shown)


lib/private/legacy/response.php
```diff
@@ -291,7 +291,7 @@ public static function setCorsHeaders($userId, $domain, $response, $config = nul
 			$config = \OC::$server->getConfig();
 		}
 		$allowedDomains = json_decode($config->getUserValue($userId, 'core', 'domains'));
-		if (in_array($domain, $allowedDomains)) {
+		if (is_array($allowedDomains) && in_array($domain, $allowedDomains)) {
```
PVince81: a better way: add an argument `[]` to `getUserValue()` so empty array is the default.

PVince81 (approved):
:+1:

@noveens please include this fix in your backport PR

lib/private/legacy/response.php
phil-davis: @PVince81 do you want me to change the code as suggested?
(I see that you pressed "approve" below)

PVince81: up to you, both are fine

phil-davis: then I will leave it this way, since CI has been running a long time already.

PR #28835, opened 2017-08-29
Stable10 filescleanup

dercorn:
Added a more comprehensible description for files:cleanup command.


phil-davis:
@dercorn this PR has been made against master. It should be against stable10.

PR #28839, opened 2017-08-29
[stable10] more comprehensible description for files:scan

dercorn:
Added a more comprehensible description for the files:scan command.


phil-davis (approved):
That looks much better!

PR #28842, opened 2017-08-29
Improve file upload progress bar

PVince81:
## Description
Only hide at the end of all chunked transfers.

Display "Processing files..." when stalled at 100% (final move,
antivirus, etc)

## Related Issue
None

## Motivation and Context
People complain about the progress bar being stuck at 100% for a while at the end of the transfer.
Since we don't have any way to detect progress past bytes transfer, this changes the message to make it look like something is still happening.

## How Has This Been Tested?
Prepare 5 local 100 MB files. Upload them all in one batch. Observe progress bar.

Before the fix: progress bar disappears too early after the first file is done.
After the fix: progress bar stays until the end of the party and 🍸 a bit longer displaying "Processing files..."

## Types of changes

## Checklist:

@pmaier1 @cdamken FYI

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -764,6 +764,7 @@ OC.Uploader.prototype = _.extend({
 
 	_hideProgressBar: function() {
 		var self = this;
+		window.clearInterval(this._progressBarInterval);
 		$('#uploadprogresswrapper .stop').fadeOut();
 		$('#uploadprogressbar').fadeOut(function() {
 			self.$uploadEl.trigger(new $.Event('resized'));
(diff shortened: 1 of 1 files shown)


PVince81:
@VicDeo please review and also test with antivirus. The delay doesn't care what is delaying, it just checks if we are stalled at 100%.

apps/files/js/file-upload.js
```diff
@@ -773,6 +774,28 @@ OC.Uploader.prototype = _.extend({
 	_showProgressBar: function() {
 		$('#uploadprogressbar').fadeIn();
 		this.$uploadEl.trigger(new $.Event('resized'));
+		if (this._progressBarInterval) {
+			window.clearInterval(this._progressBarInterval);
+		}
+		this._progressBarInterval = window.setInterval(_.bind(this._updateProgressBar, this), 1000);
```
jvillafanez: I'd rather move this progress bar handling to another class. This way we can extract these superfluous attributes from the upload object.
If time is pressuring, it's fine to leave it as technical debt since the usage of those attributes isn't spread.

jvillafanez (approved):
Code looks fine, not tested.

apps/files/js/file-upload.js
PVince81: At first I started moving this to a new class then I realized that it would be too many changes at once, so I reverted to this disappointedly ugly solution.

PVince81:
stable10: https://github.com/owncloud/core/pull/28861

PR #28843, opened 2017-08-29
[stable10] Only reload file list when switching navigation sections

PVince81:
Backport of https://github.com/owncloud/core/pull/28804 to stable10


phil-davis (approved):
Works on top of stable10

noveens (approved):
Works for me :+1:

PR #28845, opened 2017-08-29
[stable10] Changes to encryption wrapper

sharidas:
Changes made to encryption wrapper so that we can
get the commands like transfer-ownership or recreate
masterkey work. This change doesn't alter the core
functionality of copy or fopen function. The arguments
passed to the function remains same as other wrappers.

## Description

Changes made to the ecnryption warpper and view to get the transfer-ownership command recreate master key work. Basically there was a problem which caused to change the arguments passed to the routines like fopen or copy. With this change we don't have to alter them. But we use flags to control the code flow for the commands.

## Motivation and Context

This change would prevent the modification to common routines which were made to arguments of functions like fopen or copy etc.

## How Has This Been Tested?

- [x] Scratch installation of oC and create 2 users
- [x] Verified transfer ownership command works with and without encryption between two users.
- [x] Verified Recreate command for master key works on the same setup.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -24,10 +24,12 @@
 
 namespace OCA\Files\Command;
 
+use OC\Encryption\Manager;
 use OC\Files\Filesystem;
 use OC\Files\View;
 use OCP\Files\FileInfo;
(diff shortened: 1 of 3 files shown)


PVince81 (approved):
:+1:

PR #21401, opened 2015-12-30
Put nodes from Directory->getChildren in the ObjectTree cache

icewind1991:
Saves us from having to generate the node again later in `getNodeForPath` for a significant performance improvement when working with large folders

Also note that creating the node in `getChildren` is cheaper than in `getNodeForPath` since in the former we already have the fileinfo

[comparison](https://blackfire.io/profiles/compare/b57dcdd7-0607-4169-b23e-d4150d580d3d/graph)

cc @DeepDiver1975 @rullzer

PR #28853, opened 2017-08-30
Don't query Sabre parent nodes

### LINKED ISSUE #28779: Deletion of several folders or files fails when using external storages.
### Steps to reproduce

0. Set up an external storage. For example SFTP.
1. In oC files view create three or more folders and upload files inside them.
2. Select three of the created folders and try to delete them.

### Expected behaviour
All folders are removed.

### Actual behaviour
There are errors. The first one is erased, but the next are locked.

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

(current master not stable10)
{"installed":"true","maintenance":"false","needsDbUpgrade":"false","version":"10.0.3.0","versionstring":"10.0.3 beta","edition":"Community","productname":"ownCloud"}

Fresh

```

```

Yes. SFTP

**Are you using encryption:** 
No.

**Logs**
```
{"reqId":"pUh0lrgWVTKkcFPHk2Y7","level":4,"time":"2017-08-23T11:00:34+00:00","remoteAddr":"HIDDEN_IP","user":"admin","app":"webdav","method":"DELETE","url":"\/remote.php\/dav\/files\/admin\/SFTP\/FOLDER3","message":"Exception: {\"Message\":\"HTTP\\\/1.1 423 \\\"SFTP\\\/FOLDER3\\\" is locked\",\"Exception\":\"OCA\\\\DAV\\\\Connector\\\\Sabre\\\\Exception\\\\FileLocked\",\"Code\":0,\"Trace\":\"#0 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Tree.php(179): OCA\\\\DAV\\\\Connector\\\\Sabre\\\\Directory->delete()\\n#1 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/CorePlugin.php(287): Sabre\\\\DAV\\\\Tree->delete('files\\\/admin\\\/SFT...')\\n#2 [internal function]: Sabre\\\\DAV\\\\CorePlugin->httpDelete(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#3 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#4 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:DELETE', Array)\\n#5 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#6 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Server.php(234): Sabre\\\\DAV\\\\Server->exec()\\n#7 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/appinfo\\\/v2\\\/remote.php(31): OCA\\\\DAV\\\\Server->exec()\\n#8 \\\/opt\\\/owncloud\\\/remote.php(175): require_once('\\\/opt\\\/owncloud\\\/a...')\\n#9 {main}\",\"File\":\"\\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Connector\\\/Sabre\\\/Directory.php\",\"Line\":330,\"User\":\"admin\"}"}
{"reqId":"1VUz6STOPw52f35meTsC","level":4,"time":"2017-08-23T11:00:34+00:00","remoteAddr":"HIDDEN_IP","user":"admin","app":"webdav","method":"DELETE","url":"\/remote.php\/dav\/files\/admin\/SFTP\/FOLDER2","message":"Exception: {\"Message\":\"HTTP\\\/1.1 423 \\\"SFTP\\\/FOLDER2\\\" is locked\",\"Exception\":\"OCA\\\\DAV\\\\Connector\\\\Sabre\\\\Exception\\\\FileLocked\",\"Code\":0,\"Trace\":\"#0 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Tree.php(179): OCA\\\\DAV\\\\Connector\\\\Sabre\\\\Directory->delete()\\n#1 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/CorePlugin.php(287): Sabre\\\\DAV\\\\Tree->delete('files\\\/admin\\\/SFT...')\\n#2 [internal function]: Sabre\\\\DAV\\\\CorePlugin->httpDelete(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#3 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#4 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:DELETE', Array)\\n#5 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#6 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Server.php(234): Sabre\\\\DAV\\\\Server->exec()\\n#7 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/appinfo\\\/v2\\\/remote.php(31): OCA\\\\DAV\\\\Server->exec()\\n#8 \\\/opt\\\/owncloud\\\/remote.php(175): require_once('\\\/opt\\\/owncloud\\\/a...')\\n#9 {main}\",\"File\":\"\\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Connector\\\/Sabre\\\/Directory.php\",\"Line\":330,\"User\":\"admin\"}"}
```

### Client configuration

**Browser**
Chrome

Ref: https://github.com/owncloud/files_onedrive/issues/37

PVince81:
## Description
When dealing with file access, the old DAV had a shortcut way to
directly access a file node without having to traverse all parents.
The new DAV endpoint needs this as well:

- added \OCA\DAV\Tree implementation with shortcut for "files"
- disable DAVACL plugin as it would also query all parents

There are several reasons for avoiding to query the parents:
- it causes file cache access for each node, not removable as it needs
to check for existence
- with external storages and update detection, querying a folder node
causes update detection to kick in and the scanner to do its work, the
scanner will set additional locks that might needlessly prevent
concurrent access.

The sad part is that any future Sabre plugin without caring for the above has a risk of retriggering these issues. So any Sabre plugin dealing with files must be aware of this and avoid needlessly querying every parent node.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28779

## Motivation and Context
Read from https://github.com/owncloud/core/issues/28779#issuecomment-325927979.
But also this will improve performance or at least bring it closer to the one from the old DAV endpoint by avoiding additional file cache queries.

## How Has This Been Tested?
Test steps from above ticket: deletion of three folders simultaneously on an external storage with update detection kept enabled.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/ObjectTree.php
+++ b/apps/dav/lib/Connector/Sabre/ObjectTree.php
@@ -130,28 +130,23 @@ class ObjectTree extends \Sabre\DAV\Tree {
 			throw new \Sabre\DAV\Exception\ServiceUnavailable('filesystem not setup');
 		}
 
-		// check the path, also called when the path has been entered manually eg via a file explorer
-		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
-			throw new \Sabre\DAV\Exception\Forbidden();
-		}
-
 		$path = trim($path, '/');
 
 		if (isset($this->cache[$path]) && $this->cache[$path] !== false) {
 			return $this->cache[$path];
 		}
 
-		if ($path) {
+		// check the path, also called when the path has been entered manually eg via a file explorer
+		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		if ($path !== '') {
 			try {
 				$this->fileView->verifyPath($path, basename($path));
 			} catch (\OCP\Files\InvalidPathException $ex) {
 				throw new InvalidPath($ex->getMessage());
 			}
-		}
-
-		// check the path, also called when the path has been entered manually eg via a file explorer
-		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
-			throw new \Sabre\DAV\Exception\Forbidden();
 		}
 
 		// Is it the root node?
--- a/apps/dav/lib/Files/FilesHome.php
+++ b/apps/dav/lib/Files/FilesHome.php
@@ -21,11 +21,13 @@
  */
 namespace OCA\DAV\Files;
 
-use OCA\DAV\Connector\Sabre\Directory;
+use OCA\DAV\Connector\Sabre\ObjectTree;
 use Sabre\DAV\Exception\Forbidden;
 use Sabre\HTTP\URLUtil;
+use Sabre\DAV\ICollection;
+use OCA\DAV\Connector\Sabre\Directory;
 
-class FilesHome extends Directory {
+class FilesHome extends ObjectTree implements ICollection {
 
 	/**
 	 * @var array
@@ -38,10 +40,37 @@ class FilesHome extends Directory {
 	 * @param array $principalInfo
 	 */
 	public function __construct($principalInfo) {
+		parent::__construct();
 		$this->principalInfo = $principalInfo;
 		$view = \OC\Files\Filesystem::getView();
 		$rootInfo = $view->getFileInfo('');
-		parent::__construct($view, $rootInfo);
+		$rootNode = new Directory($view, $rootInfo, $this);
+		$this->init($rootNode, $view, \OC::$server->getMountManager());
+
+	}
+
+	function createFile($name, $data = null) {
+		return $this->rootNode->createFile($name, $data);
+	}
+
+	function createDirectory($name) {
+		return $this->rootNode->createDirectory($name);
+	}
+
+	function getChild($name) {
+		return $this->rootNode->getChild($name);
+	}
+
+	function getChildren() {
+		return $this->rootNode->getChildren();
+	}
+
+	function childExists($name) {
+		return $this->rootNode->childExists($name);
+	}
+
+	function getLastModified() {
+		return $this->rootNode->getLastModified();
 	}
 
 	function delete() {
--- a/apps/dav/lib/Files/FilesHome.php
+++ b/apps/dav/lib/Files/FilesHome.php
@@ -25,13 +25,10 @@ use OCA\DAV\Connector\Sabre\ObjectTree;
 use Sabre\DAV\Exception\Forbidden;
 use Sabre\HTTP\URLUtil;
 use Sabre\DAV\ICollection;
-use OCA\DAV\Connector\Sabre\Directory;
 
 class FilesHome extends ObjectTree implements ICollection {
 
-	/**
-	 * @var array
-	 */
+	/** @var array */
 	private $principalInfo;
 
 	/**
@@ -42,11 +39,6 @@ class FilesHome extends ObjectTree implements ICollection {
 	public function __construct($principalInfo) {
 		parent::__construct();
 		$this->principalInfo = $principalInfo;
-		$view = \OC\Files\Filesystem::getView();
-		$rootInfo = $view->getFileInfo('');
-		$rootNode = new Directory($view, $rootInfo, $this);
-		$this->init($rootNode, $view, \OC::$server->getMountManager());
-
 	}
 
 	function createFile($name, $data = null) {
@@ -54,15 +46,18 @@ class FilesHome extends ObjectTree implements ICollection {
 	}
 
 	function createDirectory($name) {
-		return $this->rootNode->createDirectory($name);
+		$this->rootNode->createDirectory($name);
 	}
 
 	function getChild($name) {
 		return $this->rootNode->getChild($name);
 	}
 
-	function getChildren() {
-		return $this->rootNode->getChildren();
+	function getChildren($path = null) {
+		if ($path === null) {
+			return $this->rootNode->getChildren();
+		}
+		return parent::getChildren($path);
 	}
 
 	function childExists($name) {
@@ -73,8 +68,15 @@ class FilesHome extends ObjectTree implements ICollection {
 		return $this->rootNode->getLastModified();
 	}
 
-	function delete() {
-		throw new Forbidden('Permission denied to delete home folder');
+	/**
+	 * @param string $path
+	 * @throws Forbidden
+	 */
+	function delete($path = null) {
+		if ($path === null) {
+			throw new Forbidden('Permission denied to delete home folder');
+		}
+		parent::delete($path);
 	}
 
 	function getName() {
@@ -82,6 +84,10 @@ class FilesHome extends ObjectTree implements ICollection {
 		return $name;
 	}
 
+	/**
+	 * @param string $name
+	 * @throws Forbidden
+	 */
 	function setName($name) {
 		throw new Forbidden('Permission denied to rename this folder');
 	}
--- a/apps/dav/lib/Tree.php
+++ b/apps/dav/lib/Tree.php
@@ -1,0 +1,164 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV;
+
+use OCA\DAV\Connector\Sabre\Exception\Forbidden;
+use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
+use OCA\DAV\Connector\Sabre\Exception\FileLocked;
+use OCP\Files\ForbiddenException;
+use OCP\Files\StorageInvalidException;
+use OCP\Files\StorageNotAvailableException;
+use OCP\Lock\LockedException;
+use OC\Files\View;
+use Sabre\DAV\Exception\NotFound;
+use Sabre\DAV\ICollection;
+use OCA\DAV\Connector\Sabre\Node;
+
+/**
+ * Sabre tree of nodes.
+ *
+ * Provides a shortcut when accessing the "files/" subtree to avoid
+ * having to walk through every node and trigger unnecessary extra queries.
+ */
+class Tree extends \Sabre\DAV\Tree {
+
+	/**
+	 * Cache specific for file nodes to avoid conflicting with
+	 * the regular tree nodes cache which have the full DAV path.
+	 *
+	 * This file node cache uses the DAV subpath after "/files/$userId/"
+	 * as key
+	 *
+	 * @var array
+	 */
+	private $fileNodeCache;
+
+	/**
+	 * Creates the tree
+	 *
+	 * @param \Sabre\DAV\INode $rootNode
+	 */
+	public function __construct(ICollection $rootNode) {
+		parent::__construct($rootNode);
+		$this->fileNodeCache = [];
+	}
+
+	/**
+	 * Cache a files specific node
+	 */
+	public function cacheNode(Node $node) {
+		// FIXME this is used by Directory::getChildren() to cache every
+		// found children directly. We should find a better way in the future.
+		// This is only to keep compatible with ObjectTree and the Directory class'
+		// expectations.
+		$this->fileNodeCache[trim($node->getPath(), '/')] = $node;
+	}
+
+	/**
+	 * Returns the INode object for the requested path
+	 *
+	 * @param string $path
+	 * @return \Sabre\DAV\INode
+	 * @throws InvalidPath
+	 * @throws \Sabre\DAV\Exception\Locked
+	 * @throws \Sabre\DAV\Exception\NotFound
+	 * @throws \Sabre\DAV\Exception\Forbidden
+	 * @throws \Sabre\DAV\Exception\ServiceUnavailable
+	 */
+	public function getNodeForPath($path) {
+		// querying "files" directly or anything outside of it
+		// will fallback to the regular implementation
+		if (strpos(rtrim($path, '/'), 'files/') !== 0) {
+			return parent::getNodeForPath($path);
+		}
+
+		// anything else will go through this shortcut to file nodes to avoid
+		// traversing every parent because it would trigger
+		// additional filecache queries and also
+		// additional locking of parent nodes and
+		// potential rescan in the case of external storages with update detection
+		$sections = explode('/', $path);
+		array_shift($sections);
+		$userId = array_shift($sections);
+
+		// this will ensure that the user exists and is accessible and also
+		// use the cached version when needed
+		$filesRoot = parent::getNodeForPath('files/' . $userId);
+
+		$path = implode('/', $sections);
+
+		$path = trim($path, '/');
+
+		// is it the root node?
+		if ($path === '') {
+			return $filesRoot;
+		}
+
+		if (isset($this->fileNodeCache[$path])) {
+			if ($this->fileNodeCache[$path] === false) {
+				throw new \Sabre\DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
+			}
+			return $this->fileNodeCache[$path];
+		}
+
+		// check the path, also called when the path has been entered manually eg via a file explorer
+		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		$fileView = new View('/' . $userId . '/files/');
+
+		try {
+			$fileView->verifyPath($path, basename($path));
+		} catch (\OCP\Files\InvalidPathException $ex) {
+			throw new InvalidPath($ex->getMessage());
+		}
+
+		// read from file cache
+		try {
+			$info = $fileView->getFileInfo($path);
+		} catch (StorageNotAvailableException $e) {
+			throw new \Sabre\DAV\Exception\ServiceUnavailable('Storage is temporarily not available', 0, $e);
+		} catch (StorageInvalidException $e) {
+			throw new \Sabre\DAV\Exception\NotFound('Storage ' . $path . ' is invalid');
+		} catch (LockedException $e) {
+			throw new \Sabre\DAV\Exception\Locked();
+		} catch (ForbiddenException $e) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		if (!$info) {
+			$this->fileNodeCache[$path] = false;
+			throw new \Sabre\DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
+		}
+
+		if ($info->getType() === 'dir') {
+			$node = new \OCA\DAV\Connector\Sabre\Directory($fileView, $info, $this);
+		} else {
+			$node = new \OCA\DAV\Connector\Sabre\File($fileView, $info);
+		}
+
+		$this->fileNodeCache[$path] = $node;
+		return $node;
+
+	}
+}
--- a/apps/dav/lib/Tree.php
+++ b/apps/dav/lib/Tree.php
@@ -21,17 +21,9 @@
 
 namespace OCA\DAV;
 
-use OCA\DAV\Connector\Sabre\Exception\Forbidden;
 use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
-use OCA\DAV\Connector\Sabre\Exception\FileLocked;
-use OCP\Files\ForbiddenException;
-use OCP\Files\StorageInvalidException;
-use OCP\Files\StorageNotAvailableException;
-use OCP\Lock\LockedException;
-use OC\Files\View;
 use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\ICollection;
-use OCA\DAV\Connector\Sabre\Node;
 
 /**
  * Sabre tree of nodes.
@@ -45,11 +37,7 @@ class Tree extends \Sabre\DAV\Tree {
 	 *
 	 * @param string $path
 	 * @return \Sabre\DAV\INode
-	 * @throws InvalidPath
-	 * @throws \Sabre\DAV\Exception\Locked
-	 * @throws \Sabre\DAV\Exception\NotFound
-	 * @throws \Sabre\DAV\Exception\Forbidden
-	 * @throws \Sabre\DAV\Exception\ServiceUnavailable
+	 * @throws NotFound
 	 */
 	public function getNodeForPath($path) {
 		// FIXME: remove this check when we are sure that other
--- a/apps/dav/lib/Tree.php
+++ b/apps/dav/lib/Tree.php
@@ -40,39 +40,6 @@ use OCA\DAV\Connector\Sabre\Node;
  * having to walk through every node and trigger unnecessary extra queries.
  */
 class Tree extends \Sabre\DAV\Tree {
-
-	/**
-	 * Cache specific for file nodes to avoid conflicting with
-	 * the regular tree nodes cache which have the full DAV path.
-	 *
-	 * This file node cache uses the DAV subpath after "/files/$userId/"
-	 * as key
-	 *
-	 * @var array
-	 */
-	private $fileNodeCache;
-
-	/**
-	 * Creates the tree
-	 *
-	 * @param \Sabre\DAV\INode $rootNode
-	 */
-	public function __construct(ICollection $rootNode) {
-		parent::__construct($rootNode);
-		$this->fileNodeCache = [];
-	}
-
-	/**
-	 * Cache a files specific node
-	 */
-	public function cacheNode(Node $node) {
-		// FIXME this is used by Directory::getChildren() to cache every
-		// found children directly. We should find a better way in the future.
-		// This is only to keep compatible with ObjectTree and the Directory class'
-		// expectations.
-		$this->fileNodeCache[trim($node->getPath(), '/')] = $node;
-	}
-
 	/**
 	 * Returns the INode object for the requested path
 	 *
@@ -85,80 +52,36 @@ class Tree extends \Sabre\DAV\Tree {
 	 * @throws \Sabre\DAV\Exception\ServiceUnavailable
 	 */
 	public function getNodeForPath($path) {
+		// FIXME: remove this check when we are sure that other
+		// non-files endpoints work correctly
+
 		// querying "files" directly or anything outside of it
 		// will fallback to the regular implementation
 		if (strpos(rtrim($path, '/'), 'files/') !== 0) {
 			return parent::getNodeForPath($path);
 		}
 
-		// anything else will go through this shortcut to file nodes to avoid
-		// traversing every parent because it would trigger
-		// additional filecache queries and also
-		// additional locking of parent nodes and
-		// potential rescan in the case of external storages with update detection
-		$sections = explode('/', $path);
-		array_shift($sections);
-		$userId = array_shift($sections);
-
-		// this will ensure that the user exists and is accessible and also
-		// use the cached version when needed
-		$filesRoot = parent::getNodeForPath('files/' . $userId);
-
-		$path = implode('/', $sections);
-
 		$path = trim($path, '/');
-
-		// is it the root node?
-		if ($path === '') {
-			return $filesRoot;
+		if (isset($this->cache[$path])) {
+			return $this->cache[$path];
 		}
 
-		if (isset($this->fileNodeCache[$path])) {
-			if ($this->fileNodeCache[$path] === false) {
-				throw new \Sabre\DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
+		$sections = explode('/', $path);
+
+		$node = $this->rootNode;
+		while (count($sections) > 0) {
+			$section = array_shift($sections);
+			if (!$node instanceof ICollection) {
+				throw new NotFound('Could not find node at path: ' . $path);
 			}
-			return $this->fileNodeCache[$path];
+			$node = $node->getChild($section);
+			if ($node instanceof \Sabre\DAV\Tree) {
+				// note: we don't cache here as the sub-tree has its own cache
+				return $node->getNodeForPath(implode('/', $sections));
+			}
 		}
 
-		// check the path, also called when the path has been entered manually eg via a file explorer
-		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
-			throw new \Sabre\DAV\Exception\Forbidden();
-		}
-
-		$fileView = new View('/' . $userId . '/files/');
-
-		try {
-			$fileView->verifyPath($path, basename($path));
-		} catch (\OCP\Files\InvalidPathException $ex) {
-			throw new InvalidPath($ex->getMessage());
-		}
-
-		// read from file cache
-		try {
-			$info = $fileView->getFileInfo($path);
-		} catch (StorageNotAvailableException $e) {
-			throw new \Sabre\DAV\Exception\ServiceUnavailable('Storage is temporarily not available', 0, $e);
-		} catch (StorageInvalidException $e) {
-			throw new \Sabre\DAV\Exception\NotFound('Storage ' . $path . ' is invalid');
-		} catch (LockedException $e) {
-			throw new \Sabre\DAV\Exception\Locked();
-		} catch (ForbiddenException $e) {
-			throw new \Sabre\DAV\Exception\Forbidden();
-		}
-
-		if (!$info) {
-			$this->fileNodeCache[$path] = false;
-			throw new \Sabre\DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
-		}
-
-		if ($info->getType() === 'dir') {
-			$node = new \OCA\DAV\Connector\Sabre\Directory($fileView, $info, $this);
-		} else {
-			$node = new \OCA\DAV\Connector\Sabre\File($fileView, $info);
-		}
-
-		$this->fileNodeCache[$path] = $node;
+		$this->cache[$path] = $node;
 		return $node;
-
 	}
 }
--- a/apps/dav/tests/unit/TreeTest.php
+++ b/apps/dav/tests/unit/TreeTest.php
@@ -1,0 +1,174 @@
+<?php
+/**
+ * @author Joas Schilling <coding@schilljs.com>
+ * @author Morris Jobke <hey@morrisjobke.de>
+ * @author Robin Appelman <icewind@owncloud.com>
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\Tests\unit;
+
+
+use OC\Files\FileInfo;
+use OC\Files\Filesystem;
+use OC\Files\Storage\Temporary;
+use OC\Files\View;
+use OCA\DAV\Connector\Sabre\Directory;
+use OCA\DAV\Connector\Sabre\ObjectTree;
+use Test\TestCase;
+
+/**
+ * Class TreeTest
+ *
+ * @group DB
+ *
+ * @package OCA\DAV\Tests\Unit
+ */
+class TreeTest extends TestCase {
+
+	/**
+	 * @dataProvider nodeForPathProvider
+	 */
+	public function testGetNodeForPath(
+		$inputFileName,
+		$fileInfoQueryPath,
+		$outputFileName,
+		$type
+	) {
+		$rootNode = $this->getMockBuilder('\OCA\DAV\Connector\Sabre\Directory')
+			->disableOriginalConstructor()
+			->getMock();
+		$mountManager = $this->createMock('\OC\Files\Mount\Manager');
+		$view = $this->createMock('\OC\Files\View');
+		$fileInfo = $this->createMock('\OCP\Files\FileInfo');
+		$fileInfo->expects($this->once())
+			->method('getType')
+			->will($this->returnValue($type));
+		$fileInfo->expects($this->once())
+			->method('getName')
+			->will($this->returnValue($outputFileName));
+
+		$view->expects($this->once())
+			->method('getFileInfo')
+			->with($fileInfoQueryPath)
+			->will($this->returnValue($fileInfo));
+
+		$tree = new \OCA\DAV\Connector\Sabre\ObjectTree();
+		$tree->init($rootNode, $view, $mountManager);
+
+		$node = $tree->getNodeForPath($inputFileName);
+
+		$this->assertNotNull($node);
+		$this->assertEquals($outputFileName, $node->getName());
+
+		if ($type === 'file') {
+			$this->assertTrue($node instanceof \OCA\DAV\Connector\Sabre\File);
+		} else {
+			$this->assertTrue($node instanceof Directory);
+		}
+	}
+
+	function nodeForPathProvider() {
+		return [
+			// regular file
+			[
+				'regularfile.txt',
+				'regularfile.txt',
+				'regularfile.txt',
+				'file',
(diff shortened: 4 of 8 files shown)


jvillafanez:
> with external storages and update detection, querying a folder node
causes update detection to kick in and the scanner to do its work, the
scanner will set additional locks that might needlessly prevent
concurrent access.

This should be fine unless you're deleting a parent (at any level) of another folder that is being deleted at the same. In this case it's fine because there are operations that are happening in the contents of the folder.

I think just disabling the ACL plugin should be enough to solve the problem.

PVince81:
> I think just disabling the ACL plugin should be enough to solve the problem.

It's not. I already tried: both fixes are needed else the 423 Locked exceptions will creep in.
This is because the object Tree also touches parent nodes

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,138 @@
...
+		if ($path) {
+			try {
+				$fileView->verifyPath($path, basename($path));
+			} catch (\OCP\Files\InvalidPathException $ex) {
+				throw new InvalidPath($ex->getMessage());
+			}
+		}
+
+		// check the path, also called when the path has been entered manually eg via a file explorer
+		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
```
jvillafanez: any reason for this duplication? it will a comment if any.

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,138 @@
...
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		// Is it the root node?
+		if ($path === '') {
```
jvillafanez: move this check up, before any complex operation?

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,138 @@
...
+
+		$fileView = new View('/' . $userId . '/files/');
+		$path = implode('/', $sections);
+
+		// check the path, also called when the path has been entered manually eg via a file explorer
+		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		$path = trim($path, '/');
```
jvillafanez: we could move this before the `isForbiddenFileOrDir` check

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,138 @@
...
+
+		if (isset($this->cache[$path]) && $this->cache[$path] !== false) {
```
jvillafanez: this also might be possible to move it before the `isForbiddenFileOrDir` check, to return the cache data as soon as we can

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,138 @@
...
+			throw new \Sabre\DAV\Exception\NotFound('Storage ' . $path . ' is invalid');
+		} catch (LockedException $e) {
+			throw new \Sabre\DAV\Exception\Locked();
+		} catch (ForbiddenException $e) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		if (!$info) {
+			$this->cache[$path] = false;
+			throw new \Sabre\DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
```
jvillafanez: Do we want to cache this exception? I mean, if the file doesn't exists at some point, it won't likely exists during the rest of the request, so we might not want to go through all the steps again to get another NotFound. On the other hand, it might be fine since it's expected the exception isn't catched until the top level to send the message to the client.

PVince81:
@jvillafanez I actually copied getNodeForPath from the old ObjectTree::getNodeForPath and removed every bit related to chunking or part files. Then I added a few lines at the top to compute the correct path. The result is what you see above. 

I'll still have a look at your proposals for optimization, might be a good opportunity.

Before I do, do you agree with the general approach ?

I'm not sure if @DeepDiver1975 would appreciate that we brought back the Tree implementation but so far I don't see any better way. Too bad Sabre doesn't have a "getNodeForPath" event that plugins could override.

jvillafanez:
The only option left that I see in addition to this one is to override the RootCollection instead of the \OCA\DAV\Tree. I don't know what is the best option.

The other option would be to modify the View and change or add a method to get the file info without triggering a scan, but in addition to be risky (too many components might be affected) we'll need to evaluate what could be the consecuences of fetching not up-to-date data.

PVince81:
> The only option left that I see in addition to this one is to override the RootCollection instead of the \OCA\DAV\Tree. I don't know what is the best option.

We already have our own RootCollection. Sadly the API `getChild()` only allows to get the next child, not a sub-sub-sub-child. And the calling code also calls it this way, recursively, step by step.

> The other option would be to modify the View and change or add a method to get the file info without triggering a scan, but in addition to be risky (too many components might be affected) we'll need to evaluate what could be the consecuences of fetching not up-to-date data.

I looked into this here: https://github.com/owncloud/core/pull/28855. You're right that it's risky so I'm less happy with https://github.com/owncloud/core/pull/28855 than with https://github.com/owncloud/core/pull/28853

PVince81:
- [ ] adjust getNodeForPath based on comments
- [ ] fix unit tests
- [ ] write new unit tests for the new `Tree` class

apps/dav/lib/Tree.php
PVince81: the exception is cached already with the value `false` for future checks

PVince81:
- [ ] double check that cached `false` will rethrow exception if queried again

apps/dav/lib/Tree.php
PVince81: hah... seems it's also duplicated in ObjectTree...

apps/dav/lib/Tree.php
PVince81: good idea

PVince81:
In theory we could invent a new interface that allows our sabre collections to provide "getNodeForPath" for a subpath. If implemented, call it. Else fall back to traversing with `getChild()`

PVince81:
- [ ] double check `$this->tree->cacheNode()` usage in `Directory`, might not be using the correct path...

PVince81:
We might not even need `cacheNode()` any more thanks to: https://github.com/fruux/sabre-dav/blob/master/lib/DAV/Tree.php#L200.

But I also found this... https://github.com/owncloud/core/pull/21401

PVince81:
Now from reading the comments, it seems that doing the `cacheNode` thing was relevant mostly for the old DAV endpoint where we had the custom `ObjectTree`. Now in the new DAV endpoint there is no such tree. Well, there will be one added again from this PR, so not sure...

I think re-adding `getNodeForPath` is likely the correct way after reading this comment: https://github.com/owncloud/core/pull/21401#r49167496

PVince81:
I have the feeling that this is going a bit too far already with risk of regressions or at least performance regressions... If I don't find a solution until next week I'll revert the web UI to use the old DAV endpoint for the time being.

SergioBertolinSG:
After mounting an external storage the web ui has an spinner spinning forever, refresh doesn't help.

SergioBertolinSG:
Retested, external storage is not required, it fails just opening files view:
```
{"reqId":"ZgobgAAahizRXXgw15PQ","level":3,"time":"2017-09-04T08:45:32+00:00","remoteAddr":"::1","user":"--","app":"PHP","method":"GET","url":"\/index.php","message":"Undefined variable: user at \/opt\/owncloud\/lib\/public\/Util.php#193"}
{"reqId":"yX6y7isiGhJ5tfbBkRDp","level":3,"time":"2017-09-04T08:46:16+00:00","remoteAddr":"HIDDEN_IP","user":"admin","app":"remote","method":"PROPFIND","url":"\/remote.php\/dav\/files\/admin\/","message":"Exception: {\"Exception\":\"TypeError\",\"Message\":\"Argument 2 passed to Sabre\\\\DAV\\\\Server::getPropertiesByNode() must implement interface Sabre\\\\DAV\\\\INode, null given, called in \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php on line 980\",\"Code\":0,\"Trace\":\"#0 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(980): Sabre\\\\DAV\\\\Server->getPropertiesByNode(Object(Sabre\\\\DAV\\\\PropFind), NULL)\\n#1 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(1665): Sabre\\\\DAV\\\\Server->getPropertiesIteratorForPath('files\\\/admin', Array, 1)\\n#2 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/CorePlugin.php(355): Sabre\\\\DAV\\\\Server->generateMultiStatus(Object(Generator), false)\\n#3 [internal function]: Sabre\\\\DAV\\\\CorePlugin->httpPropFind(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#4 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#5 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:PROPFIND', Array)\\n#6 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#7 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Server.php(243): Sabre\\\\DAV\\\\Server->exec()\\n#8 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/appinfo\\\/v2\\\/remote.php(31): OCA\\\\DAV\\\\Server->exec()\\n#9 \\\/opt\\\/owncloud\\\/remote.php(175): require_once('\\\/opt\\\/owncloud\\\/a...')\\n#10 {main}\",\"File\":\"\\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php\",\"Line\":1056}"}
```

PVince81:
argh... I was doing something experimental and moved around code but made the mistake to commit the removal of said code

PVince81:
- [ ] cache the full DAV path in `Tree` instead of just the files subpath...

PVince81:
@SergioBertolinSG please try again, thanks

PVince81:
@jvillafanez mind rechecking ?

In general I'm not too happy about this but it feels like it's the best compromise so far.

I hope we can eventually get rid of `cachePath` which is called by `Directory::getChildren()' for every child. I believe that Sabre is supposed to have a way already to obsolete this, but the discussion from the ticket in https://github.com/owncloud/core/pull/21401#r49167496 makes me doubt about it and I don't want to change too much at once to avoid regressions. We really need to clean up this DAV mess at some point...

SergioBertolinSG:
Better than before.

While deleting several files at once there are locked files.

```
{"reqId":"N74dSbcSJPwWbc56Sjby","level":4,"time":"2017-09-04T10:36:13+00:00","remoteAddr":"HIDDEN_IP","user":"admin","app":"webdav","method":"DELETE","url":"\/remote.php\/dav\/files\/admin\/SFTP\/VARIAS_CARPETAS\/vikings-episode-4-trial%20copy%202.jpg","message":"Exception: {\"Message\":\"HTTP\\\/1.1 423 \\\"SFTP\\\/VARIAS_CARPETAS\\\/vikings-episode-4-trial copy 2.jpg\\\" is locked\",\"Exception\":\"OCA\\\\DAV\\\\Connector\\\\Sabre\\\\Exception\\\\FileLocked\",\"Code\":0,\"Trace\":\"#0 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Tree.php(179): OCA\\\\DAV\\\\Connector\\\\Sabre\\\\File->delete()\\n#1 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/CorePlugin.php(287): Sabre\\\\DAV\\\\Tree->delete('files\\\/admin\\\/SFT...')\\n#2 [internal function]: Sabre\\\\DAV\\\\CorePlugin->httpDelete(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#3 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#4 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:DELETE', Array)\\n#5 \\\/opt\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#6 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Server.php(243): Sabre\\\\DAV\\\\Server->exec()\\n#7 \\\/opt\\\/owncloud\\\/apps\\\/dav\\\/appinfo\\\/v2\\\/remote.php(31): OCA\\\\DAV\\\\Server->exec()\\n#8 \\\/opt\\\/owncloud\\\/remote.php(175): require_once('\\\/opt\\\/owncloud\\\/a...')\\n#9 {main}\",\"File\":\"\\\/opt\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Connector\\\/Sabre\\\/File.php\",\"Line\":363,\"User\":\"admin\"}"}
```

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,157 @@
...
+		// potential rescan in the case of external storages with update detection
+		$sections = explode('/', $path);
+		array_shift($sections);
+		$userId = array_shift($sections);
+
+		// this will ensure that the user exists and is accessible and also
+		// use the cached version when needed
+		$filesRoot = parent::getNodeForPath('files/' . $userId);
+
+		$fileView = new View('/' . $userId . '/files/');
```
jvillafanez: The view creation might not be very expensive, but I think it's better to create it just before being used. There are several cases where we don't need to create the view object.

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,157 @@
...
+		} catch (StorageInvalidException $e) {
+			throw new \Sabre\DAV\Exception\NotFound('Storage ' . $path . ' is invalid');
+		} catch (LockedException $e) {
+			throw new \Sabre\DAV\Exception\Locked();
+		} catch (ForbiddenException $e) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
+		if (!$info) {
+			$this->fileNodeCache[$cachePath] = false;
```
jvillafanez: I'd rather have a clear expectation about missing fileInfos:
* If the expectation is NOT to cache the failure and let it retry afterwards, this isn't needed. In addition, the `$cachePath` variable isn't defined.
* If the expectation is to cache the failure so next call will return the same error, we should return the cached error as soon as possible.

If the valid results are being cached, I expect the errors are cached too. If this won't be the case, the PHPDoc should state that.

apps/dav/lib/Connector/Sabre/ObjectTree.php
```diff
@@ -131,17 +131,17 @@ public function getNodeForPath($path) {
...
 		if (isset($this->cache[$path]) && $this->cache[$path] !== false) {
 			return $this->cache[$path];
 		}
 
+		// check the path, also called when the path has been entered manually eg via a file explorer
+		if (\OC\Files\Filesystem::isForbiddenFileOrDir($path)) {
+			throw new \Sabre\DAV\Exception\Forbidden();
+		}
+
 		if ($path) {
```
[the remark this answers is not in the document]
PVince81: this is copy-pasted from the old code... not intended. not calling verifyPath on "0" doesn't hurt though.

I'll fix this one here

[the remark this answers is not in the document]
PVince81: this check is useless as we now check for empty path above 🔫

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,157 @@
...
+	 * @param \Sabre\DAV\INode $rootNode
+	 */
+	public function __construct(ICollection $rootNode) {
+		$this->rootNode = $rootNode;
+	}
+
+	/**
+	 * Cache a files specific node
+	 */
+	public function cacheNode(Node $node) {
```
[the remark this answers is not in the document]
PVince81: It's an awful mess from the past: Directory::getChildren() uses it to cache children as it finds them.

It is likely not the right approach any more but I don't want to touch more code than necessary right now.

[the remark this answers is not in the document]
PVince81: https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/Directory.php#L256

PVince81:
@jvillafanez I fixed them all.

Now regarding @SergioBertolinSG's comment about the extra stray lock, I'll need to take some time to log all locks again and diff the lock traces between old dav and new dav. (this is how I found out how to fix this so far)

PVince81:
I diff the lock order between old dav and new dav on this PR and they are identical:
```
acquireLock: type=1 path="a" 
acquireLock: type=1 path="" 
acquireLock: type=2 path="scanner::a" 
acquireLock: type=1 path="a" 
acquireLock: type=1 path="a" 
releaseLock: type=1 path="a" 
acquireLock: type=1 path="a\/data1.dat" 
releaseLock: type=1 path="a\/data1.dat" 
releaseLock: type=1 path="a" 
releaseLock: type=2 path="scanner::a" 
releaseLock: type=1 path="a" 
releaseLock: type=1 path="" 
acquireLock: type=1 path="a" 
acquireLock: type=1 path="" 
changeLock: type=2 path="a" 
changeLock: type=1 path="a" 
acquireLock: type=1 path="a" 
acquireLock: type=1 path="" 
releaseLock: type=1 path="a" 
releaseLock: type=1 path="" 
releaseLock: type=1 path="a" 
releaseLock: type=1 path="" 
acquireLock: type=1 path="" 
acquireLock: type=2 path="scanner::" 
acquireLock: type=1 path="" 
acquireLock: type=1 path="" 
releaseLock: type=1 path="" 
acquireLock: type=1 path="x" 
releaseLock: type=1 path="x" 
acquireLock: type=1 path="b" 
releaseLock: type=1 path="b" 
acquireLock: type=1 path="c" 
releaseLock: type=1 path="c" 
releaseLock: type=1 path="" 
releaseLock: type=2 path="scanner::" 
releaseLock: type=1 path="" 
```

so if there is a stray lock somewhere it was probably already there with the old dav endpoint, so probably a different issue.

PVince81:
I have uploaded the scan log results to https://cloud.owncloud.com/index.php/s/VK1ncpEUdvhScz9 as "log-delete-fixed*". If you diff the logs and ignore the stack traces, they are identical.

jvillafanez:
I don't see the changes 🙈

apps/dav/lib/Tree.php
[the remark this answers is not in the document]
jvillafanez: > It's an awful mess from the past: Directory::getChildren() uses it to cache children as it finds them.

:man_facepalming: Explain it in a comment for the time being, otherwise this will likely be deleted without mercy the next time, and something will explode.

Anyway, that's bad and we have to fix it at some point.

PVince81:
@jvillafanez I did push, here is one change for example: https://github.com/owncloud/core/pull/28853/files#diff-d4fb9df74a6b54801784acb672a1a952R112

What changes are you missing ?

I'll process your new comments

apps/dav/lib/Tree.php
PVince81: yes, now throwing an exception if we detect a cached `false`

apps/dav/lib/Connector/Sabre/ObjectTree.php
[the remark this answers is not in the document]
PVince81: argh... I mixed up ObjectTree with Tree... I actually fixed this in Tree only

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,157 @@
...
+	 */
+	private $fileNodeCache;
+
+	/**
+	 * Creates the tree
+	 *
+	 * @param \Sabre\DAV\INode $rootNode
+	 */
+	public function __construct(ICollection $rootNode) {
+		$this->rootNode = $rootNode;
```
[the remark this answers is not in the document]
PVince81: will do

PVince81:
@jvillafanez adjusted again and squashed

PVince81:
```
10:12:00 1) OCA\DAV\Tests\unit\ServerTest::test
10:12:00 ArgumentCountError: Too few arguments to function Sabre\DAV\Tree::__construct(), 0 passed in /var/lib/jenkins/workspace/owncloud-core_core_PR-28853-LZGDUAKIZIJB37QSAQ64PKSB6ATNPVHHK6ENRW2HVRSYU5ZK3ORA/apps/dav/lib/Tree.php on line 61 and exactly 1 expected
10:12:00 
10:12:00 /var/lib/jenkins/workspace/owncloud-core_core_PR-28853-LZGDUAKIZIJB37QSAQ64PKSB6ATNPVHHK6ENRW2HVRSYU5ZK3ORA/lib/composer/sabre/dav/lib/DAV/Tree.php:41
10:12:00 /var/lib/jenkins/workspace/owncloud-core_core_PR-28853-LZGDUAKIZIJB37QSAQ64PKSB6ATNPVHHK6ENRW2HVRSYU5ZK3ORA/apps/dav/lib/Tree.php:61
10:12:00 /var/lib/jenkins/workspace/owncloud-core_core_PR-28853-LZGDUAKIZIJB37QSAQ64PKSB6ATNPVHHK6ENRW2HVRSYU5ZK3ORA/apps/dav/lib/Server.php:70
10:12:00 /var/lib/jenkins/workspace/owncloud-core_core_PR-28853-LZGDUAKIZIJB37QSAQ64PKSB6ATNPVHHK6ENRW2HVRSYU5ZK3ORA/apps/dav/tests/unit/ServerTest.php:42
```

PVince81:
☕️ had no effect

Fixed issue, rechecked web UI and pushed/squashed again.

jvillafanez (approved):

jvillafanez:
Waiting for tests results....

PVince81:
I had a quick test with the calendar app to see if the caldav endpoint still works, and also with the comments app to verify that non-files endpoints still work.

PVince81:
okay, we got some stuff...

``` 
    checksums.Uploading a file with checksum should work using new dav path
    checksums.Uploading a file with checksum should return the checksum in the propfind using new dav path
    checksums.Uploading a file with checksum should return the checksum in the download header using new dav path
    checksums.Moving a file with checksum should return the checksum in the propfind using new dav path
    checksums.Moving file with checksum should return the checksum in the download header using new dav path
    checksums.Copying a file with checksum should return the checksum in the propfind using new dav path
    checksums.Copying file with checksum should return the checksum in the download header using new dav path
    checksums.Upload new dav chunked file where checksum matches
    quota.Uploading a file as owner having enough quota
    quota.Overwriting a file as owner having enough quota
    quota.Uploading a file in received folder having enough quota
    quota.Overwriting a file in received folder having enough quota
    quota.Overwriting a received file having enough quota
    trashbin-new-endpoint.deleting a file of a shared folder moves it to trashbin
    trashbin-new-endpoint.deleting a shared folder moves it to trashbin
    trashbin-new-endpoint.deleting a received folder doesn't move it to trashbin
    trashbin-new-endpoint.deleting a file in a received folder it is moved to the trashbin
    trashbin-new-endpoint.deleting a file in a received folder when restored it comes back to the original path
    trashbin-new-endpoint.Deleting a folder into external storage moves it to the trashbin
    trashbin-new-endpoint.Deleting a file into external storage moves it to the trashbin and can be restored
    webdav-related-new-endpoint.Moving a file
    webdav-related-new-endpoint.Setting custom DAV property and reading it after the file is renamed
    webdav-related-new-endpoint.Upload file via new chunking endpoint with correct size header
```

PVince81:
@jvillafanez oh no... can of worms mode engaged...

Remember that we said that if a file's non-existence is cached as `false`, we should rethrow the exception when queried again ? It makes sense from an API point of view. Unfortunately it is this change that breaks the integration tests.

I'll have a look what the real Sabre's ObjectTree is supposed to do.

PVince81:
this is very fishy... I see that the regular Sabre Tree which was used before this PR does throw an exception here: https://github.com/fruux/sabre-dav/blob/3.2.2/lib/DAV/Tree.php#L74.
It doesn't catch the error, so I expect that subsequent calls would still throw the exception.

And if it doesn't reach this bit, then `getChild` is the one that would throw the exception, which we do correctly here: https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/Directory.php#L247

(a side note: I see that `Directory::getChild()` is missing a call to `$this->tree->cacheNode(false)` in case of non-existence)

PVince81:
raised https://github.com/owncloud/core/issues/28912 to gather the horrors we find

jvillafanez:
> Remember that we said that if a file's non-existence is cached as false, we should rethrow the exception when queried again ? It makes sense from an API point of view. Unfortunately it is this change that breaks the integration tests.

Not caching the exception migth be fine: you query the file and it isn't present, then query again and now it is present because the file is created between those calls.
This is fine as long as we clearly specify that this is the expected behaviour.

PVince81:
from the checksum test, this is the stack that would trigger the second throw from a cached `false`:
```
0  OCA\DAV\Tree->getNodeForPath() /srv/www/htdocs/owncloud/apps/dav/lib/Tree.php:118
1  OCA\DAV\Connector\Sabre\FilesPlugin->sendFileIdHeader() /srv/www/htdocs/owncloud/apps/dav/lib/Connector/Sabre/FilesPlugin.php:404
2  call_user_func_array:{/srv/www/htdocs/owncloud/lib/composer/sabre/event/lib/EventEmitterTrait.php:105}() /srv/www/htdocs/owncloud/lib/composer/sabre/event/lib/EventEmitterTrait.php:105
3  OCA\DAV\Connector\Sabre\Server->emit() /srv/www/htdocs/owncloud/lib/composer/sabre/event/lib/EventEmitterTrait.php:105
4  OCA\DAV\Connector\Sabre\Server->createFile() /srv/www/htdocs/owncloud/lib/composer/sabre/dav/lib/DAV/Server.php:1101
5  Sabre\DAV\CorePlugin->httpPut() /srv/www/htdocs/owncloud/lib/composer/sabre/dav/lib/DAV/CorePlugin.php:525
6  call_user_func_array:{/srv/www/htdocs/owncloud/lib/composer/sabre/event/lib/EventEmitterTrait.php:105}() /srv/www/htdocs/owncloud/lib/composer/sabre/event/lib/EventEmitterTrait.php:105
7  OCA\DAV\Connector\Sabre\Server->emit() /srv/www/htdocs/owncloud/lib/composer/sabre/event/lib/EventEmitterTrait.php:105
8  OCA\DAV\Connector\Sabre\Server->invokeMethod() /srv/www/htdocs/owncloud/lib/composer/sabre/dav/lib/DAV/Server.php:479
9  OCA\DAV\Connector\Sabre\Server->exec() /srv/www/htdocs/owncloud/lib/composer/sabre/dav/lib/DAV/Server.php:254
10 OCA\DAV\Server->exec() /srv/www/htdocs/owncloud/apps/dav/lib/Server.php:243
11 require_once()  /srv/www/htdocs/owncloud/apps/dav/appinfo/v2/remote.php:31
12 {main}          /srv/www/htdocs/owncloud/remote.php:175
```

Apparently the file was created and exists now, but our current cache still contains the old value `false`. The old code path would go through the View and calls `file_exists` there, so it works.

The default Sabre `Tree:nodeExists()` doesn't go to the cache and also goes to the first parent with `getNodeForPath()` then calls `childExists()`. This is why we don't bump into that issue.

This all still looks fishy, even in the Sabre code. I could imagine scenarios where a cached node used to exist but was deleted. And I don't see any unbind events registered there. :sad:

PVince81:
listening to bind/unbind events might be a solution, but I see that code like `httpPut` doesn't even trigger a `beforeBind` event.

Ok, at least a deletion calls `Tree:delete()` from Sabre which calls `markDirty` which invalidates the cache. Then I'd need to override it to also invalidate the `fileNodeCache`.

- [ ] implement markDirty to invalidate `fileNodeCache` as we need to map the absolute path to the relative path

Maybe we should make the `FilesHome` actually extend `Sabre\DAV\Tree` and also implement the `ICollection` methods additionally. Then make our base Tree just check if the parent node we got `instanceof Tree` and then call `getNodeForPath` on the remaining subtree.

I have a bad feeling about this... The more we touch the more will break. I'm tending more in the direction of deferring this to 10.0.4.

PVince81:
If we had separate Sabre servers for each of the subtrees "/files", "/comments", etc, this wouldn't be an issue. It would even have the advantage that each subtree would register only the required plugins instead of everything.

PVince81:
Raised https://github.com/owncloud/core/issues/28913 to reconsider and discuss separate trees/servers for each first-level node.

PVince81:
For now I'm very unhappy about having to map paths every time in the base `Tree`. It's already ugly that the base `Tree` needs to know anything about the files endpoint and potentially others in the future.

So, I'm going to explore the following:
- [ ] make FilesHome extend `\Sabre\DAV\Tree` and implement `ICollection` (basically bringing back the old ObjectTree but on a sub-level)
- [ ] root `Tree` implementation where `getNodeForPath()` goes down from the root with `getChild()` iterations, for each iteration checking if the node is a `Sabre\DAV\Tree`

PVince81:
or I just make the current `Tree` work... seems our old tree overwrote `nodeExists` with this: https://github.com/owncloud/core/blob/v10.0.3beta2/apps/dav/lib/Connector/Sabre/ObjectTree.php#L110

PVince81:
@jvillafanez check this out: https://github.com/owncloud/core/pull/28853/commits/737c9b03033cf06360ac80c966df090500699cc3

At least now the specialized files code is in `FilesHome` which extends the old `ObjectTree`.
Whether we agree to reuse the old `ObjectTree` here in new dav is something to discuss...
This seems to solve the checksum errors, we'll see if other tests also pass now.

Now what's open for later is that the old `ObjectTree` implementation isn't consistent in regard to handling non-existing nodes and throwing exceptions, something to look into later... At least now we only have one occurence of the "bad code" within `ObjectTree` instead of having another copy in `Tree`.

@jvillafanez let me know what you think.

PVince81:
- [ ] adjust TreeTest to test the new logic

jvillafanez:
I'm not sure if the approach is good or not. Is there any conceptual difference between a directory and an object tree? What is expected to be attached as FilesHome? I mean I expect DAV to manage nodes, so if I query the root collection to get the expected FilesHome, should I get a node or a tree?

In addition, the constructor of the FilesHome feels quite forced: if we need a root node, why don't we inject it in the constructor instead of creating it there? Maybe we need to create a new class for that and use this new class for the new webdav endpoint.

PVince81:
@jvillafanez I think Sabre's concept is that basically there is only one tree which sits at the root and embeds the root node. Then each node is just an implementation of `INode`.

Here we're innovating a bit because we're implementing another tree in the middle of the main tree.

Injecting a root node in the constructor feels wrong, because it would have to be injected by the parent which is the node for "/files/$userId". Well, even if we do it there, we'd anyway need to instantiate a `Directory` somewhere in the tree. We can't inject it from the root. I think it is ok for `FilesHome` to contain/wrap the node like it does here.

I want to discuss this further with @DeepDiver1975 later to make sure we're all on the same page regarding this "innovative" usage.

PVince81:
now that we switched back to old DAV, we'll need this PR https://github.com/owncloud/core/pull/28874 to reproduce the original issue this PR solves

PVince81:
- [ ] BUG:
```
{"reqId":"KSPBceR23bRLjz65nvor","level":3,"time":"2017-10-12T14:37:31+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"PHP","method":"PROPFIND","url":"\/owncloud\/remote.php\/dav\/files\/admin\/Photos\/","message":"Declaration of OCA\\DAV\\Files\\FilesHome::getChildren() should be compatible with Sabre\\DAV\\Tree::getChildren($path) at \/srv\/www\/htdocs\/owncloud\/apps\/dav\/lib\/Files\/FilesHome.php#30"}
```

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,87 @@
...
+	 *
+	 * @param string $path
+	 * @return \Sabre\DAV\INode
+	 * @throws InvalidPath
+	 * @throws \Sabre\DAV\Exception\Locked
+	 * @throws \Sabre\DAV\Exception\NotFound
+	 * @throws \Sabre\DAV\Exception\Forbidden
+	 * @throws \Sabre\DAV\Exception\ServiceUnavailable
+	 */
+	public function getNodeForPath($path) {
```
DeepDiver1975: Isn't the parent implementation performing a proper caching?
why not?
shall we fix this in the parent class?

DeepDiver1975 (changes requested):

apps/dav/lib/Tree.php
PVince81: The parent impl is caching yet. The problem is that we should not query the parents **at all** as they are not useful for our current operation. And even with caching, the first time query is still being done for the parents which triggers our internal View APIs which triggers the scanning.

PVince81: Basically the approach we had in the old endpiont: getNodeForPath goes **directly** to the target node without traversing any parents, as we know that all parents are filesystem folders.

With the default Sabre approach as used by the new endpoint, getNodeForPath first traverses all parents recursively up to the root and caches these, and then calls `getChild()` on every node until reaching the actually queried node.

The returned result is the same, except that in the latter case we traversed Folder nodes and had to query getFileInfo() on every parent while in the former we only call getFileInfo() on the final child node.

Hope this makes sense.

PVince81:
@mrow4a might be interested in this ? At least come here if you happen to observe the perf regression on the new DAV endpoint

PVince81:
moving to planned for 10.0.5.

I hope I can get your feedback and understanding by then @DeepDiver1975

apps/dav/lib/Files/FilesHome.php
```diff
@@ -38,10 +40,37 @@ class FilesHome extends Directory {
...
+		$this->init($rootNode, $view, \OC::$server->getMountManager());
+
+	}
+
+	function createFile($name, $data = null) {
+		return $this->rootNode->createFile($name, $data);
+	}
+
+	function createDirectory($name) {
+		return $this->rootNode->createDirectory($name);
```
DeepDiver1975: remove return - createDirectory returns void

apps/dav/lib/Files/FilesHome.php
```diff
@@ -38,10 +40,37 @@ class FilesHome extends Directory {
...
+	}
+
+	function getChild($name) {
+		return $this->rootNode->getChild($name);
+	}
+
+	function getChildren() {
```
DeepDiver1975: Tree::getChildren($path) vs ICollection::getChildren - this is not clean

apps/dav/lib/Files/FilesHome.php
```diff
@@ -38,10 +40,37 @@ class FilesHome extends Directory {
...
+
+	function childExists($name) {
+		return $this->rootNode->childExists($name);
+	}
+
+	function getLastModified() {
+		return $this->rootNode->getLastModified();
 	}
 
 	function delete() {
```
DeepDiver1975: Tree::delete($path) vs INode::delete

apps/dav/lib/Tree.php
```diff
@@ -0,0 +1,87 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV;
+
+use OCA\DAV\Connector\Sabre\Exception\Forbidden;
```
DeepDiver1975: many unneeded use statements

apps/dav/tests/unit/TreeTest.php
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\Tests\unit;
+
+
+use OC\Files\FileInfo;
```
DeepDiver1975: unneeded use statements

apps/dav/tests/unit/TreeTest.php
```diff
@@ -0,0 +1,174 @@
...
+	/**
+	 * @dataProvider nodeForPathProvider
+	 */
+	public function testGetNodeForPath(
+		$inputFileName,
+		$fileInfoQueryPath,
+		$outputFileName,
+		$type
+	) {
+		$rootNode = $this->getMockBuilder('\OCA\DAV\Connector\Sabre\Directory')
```
DeepDiver1975: Directory::class

apps/dav/tests/unit/TreeTest.php
```diff
@@ -0,0 +1,174 @@
...
+		$fileInfo->expects($this->once())
+			->method('getName')
+			->will($this->returnValue($outputFileName));
+
+		$view->expects($this->once())
+			->method('getFileInfo')
+			->with($fileInfoQueryPath)
+			->will($this->returnValue($fileInfo));
+
+		$tree = new \OCA\DAV\Connector\Sabre\ObjectTree();
```
DeepDiver1975: use ....

apps/dav/tests/unit/TreeTest.php
```diff
@@ -0,0 +1,174 @@
...
+		$tree->getNodeForPath($path);
+	}
+
+	public function testGetNodeForPathRoot() {
+		$path = '/';
+
+
+		$storage = new Temporary([]);
+
+		$view = $this->createMock('\OC\Files\View', ['resolvePath']);
```
DeepDiver1975: createMock only takes one arg

apps/dav/tests/unit/TreeTest.php
```diff
@@ -0,0 +1,174 @@
...
+				return [$storage, ltrim($path, '/')];
+			}));
+
+		$rootNode = $this->getMockBuilder('\OCA\DAV\Connector\Sabre\Directory')
+			->disableOriginalConstructor()
+			->getMock();
+		$mountManager = $this->createMock('\OC\Files\Mount\Manager');
+
+		$tree = new \OCA\DAV\Connector\Sabre\ObjectTree();
+		$tree->init($rootNode, $view, $mountManager);
```
DeepDiver1975: please add phpdoc to $rootNode, $view, $mountManager to define theirs type

```php
/** @var Directory | \PHPUnit_Framework_MockObject_MockObject $rootNode */

DeepDiver1975 (changes requested):

DeepDiver1975:
I'm okay with this change - please take care about my comments - THX

PVince81:
Thanks. Glad that we agreed on the approach.

I'll take care of the comments next and unit tests and codecov.

apps/dav/tests/unit/TreeTest.php
PVince81: ok, so I copy-pasted this from the old code. I can clean this up.

PVince81:
Transferred to @DeepDiver1975 for hijacking

DeepDiver1975 (approved):

PVince81:
:+1: for @DeepDiver1975's changes

PVince81:
I see some tab issues, fixing... (copy-pasted code from Sabre that has spaces instead of tabs)

PVince81:
I've squashed the tab fixes into the second commit and rebased onto master.

If CI passes let's merge this and backport.

PR #28854, opened 2017-08-30
Add a script to automate environment setup

settermjd:
This PR creates a script for automating the setup of ownCloud, right up to the point of either using the installation wizard or command line installer. It:

- Installs `wget`, `make`, `npm`, `nodejs`, `unzip`, & `git`
- Ensures nodejs is available in the user path, fixing it if necessary
- Runs `make` to install the PHP and JS dependencies

It's not, currently, special. But it's a good start. It was written after going through [the development environment setup documentation](https://doc.owncloud.com/server/latest/developer_manual/general/devenv.html) and seeing that there was both incorrect and missing information — and that a script might just save users a lot of time.

### Supports

- **Debian/Ubuntu:** 16.04
- **openSUSE Leap:** 42.3
- **CentOS:** 7

All constructive feedback welcome.

### Assumptions

- Apache (or another web server) are already installed

### DIFF
--- a/bin/setup-environment.sh
+++ b/bin/setup-environment.sh
@@ -45,16 +45,56 @@ function check_permissions()
   echo 'User & group permissions pass.'
 }
 
-function install_required_packages()
+function which_distro()
 {
   case "$( grep -Eoi 'Debian|SUSE|Ubuntu' /etc/issue )" in 
+    "SUSE") 
+      echo "SUSE"
+      ;;
+    "Ubuntu") 
+      echo "Ubuntu"
+      ;;
+  esac
+
+  redhat_release_file=/etc/redhat-release
+
+  # Need to do a bit more work to detect RedHat-based distributions
+  if [ -e "$redhat_release_file" ]; then
+    case "$( grep -Eoi 'CentOS' $redhat_release_file )" in 
+      "CentOS") 
+        echo "CentOS"
+        ;;
+    esac
+  fi
+}
+
+function install_required_packages()
+{
+  case "$( which_distro )" in 
     "SUSE") echo "Installing required packages on SUSE"
       install_required_suse_packages
       ;;
     "Ubuntu"|"Debian") echo "Installing required packages on Ubuntu/Debian"
       install_required_ubuntu_debian_packages
       ;;
+    "CentOS") echo "Installing required packages on Centos"
+      install_required_centos_packages
+      ;;
   esac
+}
+
+function install_required_centos_packages()
+{
+  sudo yum update -q -y
+  sudo yum --enablerepo=cr -q -y install wget make nodejs unzip git npm bzip2 file
+  
+  # Install PhantomJS - see https://www.bonusbits.com/wiki/HowTo:Install_PhantomJS_on_CentOS
+  # It's not in the official repos, so needs to be installed independently.
+  sudo yum install fontconfig freetype freetype-devel fontconfig-devel libstdc++
+  sudo mkdir -p /opt/phantomjs
+  wget https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.1.1-linux-x86_64.tar.bz2 
+  sudo tar -jxvf phantomjs-2.1.1-linux-x86_64.tar.bz2 --strip-components 1 --directory /opt/phantomjs/
+  sudo ln -s /opt/phantomjs/bin/phantomjs /usr/bin/phantomjs
 }
 
 function install_required_suse_packages()
@@ -110,10 +150,10 @@ then
      exit $#
    fi
 
-   if grep -Eq 'Debian|Ubuntu|SUSE' /etc/issue; then 
+   if which_distro; then 
      main "$1" "$2"
    else
-     echo 'Currently, only Debian, Ubuntu, and SUSE distributions are supported'
+     echo 'Currently, only Debian, Ubuntu, SUSE, and CentOS distributions are supported'
      echo 'Others are planned to be supported in the future.'
      exit -1
    fi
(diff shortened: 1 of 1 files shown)


bin/setup-environment.sh
```diff
@@ -0,0 +1,122 @@
...
+    exit -1
+  fi
+
+  echo 'User & group permissions pass.'
+}
+
+function install_required_packages()
+{
+  # Ensure that Apt’s cache is up to date
+  sudo apt-get -y -q update
```
PVince81: won't work on opensuse

PVince81:
While I understand the feeling and usefulness, I have mixed feeling about this, considering that it will not work on all major Linux distributions.

bin/setup-environment.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+function fix_nodejs_binary()
+{
+  nodejs=$( which nodejs )
+
+  if [ -z "$( which node )" ]; then
+    echo 'Node is installed, but not correctly set in the system path.'
+    echo 'Linking ' "$nodejs" ' to /usr/bin/node'
+    sudo ln -s "$nodejs" /usr/bin/node
```
tboerger: Ubuntu and Debian should provide a node legacy package for that, no need to manually symlink

tboerger:
Maybe we should push docker-based environments instead? I'm working on multiple dock-compose files within the core repo.

PVince81:
If we can provide a dev friendly docker with everything already setup (including XDebug) this could be nice for people to quickly start.

bin/setup-environment.sh
settermjd: That's right. That's what it says in the script, and in the PR description :-) It's for Debian-based distros only atm. Time allowing, I'll expand it out.

bin/setup-environment.sh
settermjd: Thanks for the pointer.

settermjd:
I see what you're both saying about Docker. But while the documentation currently covers manual installation it makes sense to provide something like this — at least in the interim.

mmattel:
Why not naming it ``setup-environment-debian.sh`` ?

PVince81:
and then have `make setup` that detects your distro and runs the matching script...

settermjd:
Both are good points. Just about got it supporting openSUSE 42.3 atm.

bin/setup-environment.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
+#
+## 
+## Usage: setup-environment.sh
```
mmattel: to be added in the usage description: ``<webserver user>`` ``<owncloud install dir>``

settermjd:
Support for openSUSE and CentOS is now available.

bin/setup-environment.sh
```diff
@@ -0,0 +1,166 @@
...
+## Usage: setup-environment.sh <webserver user> <owncloud install dir>
+##
+## This script will setup an ownCloud development environment on a Debian-Linux
+## distribution. 
+##
+set -euo pipefail
+IFS=$'\n\t'
+REQUIRED_ARGS=2
+
+if [ "$( whoami )" == 'root' ]
```
mmattel: do I get it right that this ``whoami`` code block is double present and not necessary because of ``function check_user()`` ?

bin/setup-environment.sh
```diff
@@ -0,0 +1,166 @@
...
+    echo "Please ensure that $www_user has read/write access to $owncloud_directory"
+    exit -1
+  fi
+
+  echo 'User & group permissions pass.'
+}
+
+function which_distro()
+{
+  case "$( grep -Eoi 'Debian|SUSE|Ubuntu' /etc/issue )" in
```
tboerger: I think it's more common to parse the `/etc/os-release`, which already provides variables if you eval the file.

settermjd: Ah great. Thanks for that. I'd not heard of that file before, but just had a quick skim through https://www.freedesktop.org/software/systemd/man/os-release.html. So I can see that using it is a much better way of doing what I was attempting to do.

bin/setup-environment.sh
settermjd: Yes, you are. That must have been left by mistake from the original version of the script.

settermjd:
@tboerger, thanks for the suggestion. Advice implemented.

bin/setup-environment.sh
```diff
@@ -0,0 +1,173 @@
...
+    echo "Please ensure that $www_user has read/write access to $owncloud_directory"
+    exit -1
+  fi
+
+  echo 'User & group permissions pass.'
+}
+
+function which_distro()
+{
+  echo $( grep '^ID=' /etc/os-release ) | awk '{split($0,array,"=")} END { print array[2] }'
```
tboerger:
```
source /etc/os-release
echo $ID
```

bin/setup-environment.sh
```diff
@@ -0,0 +1,173 @@
...
+}
+
+function distro_version()
+{
+  echo $( grep '^VERSION_ID=' /etc/os-release ) | awk '{split($0,array,"=")} END { print array[2] }' | tr -d '""'
```
tboerger:
```
source /etc/os-release
echo $VERSION_ID
```

DeepDiver1975:
I honestly don't see this something to be part of core. Can we put this into the documentation somewhere or maybe post this in some blog or something?

This script needs constant development to keep up to date with the changes of distros. No idea who will do this.

If this script is part of core the expectation is that we keep this up to date.

I'm closing this for now. THX

mmattel:
@settermjd maybe something to be added to the documentation/development section ?

Issue #28850, opened 2017-08-30
Broken trashbin layout


PVince81:
### Steps:
1. Check out stable10
1. Delete some files
1. Go to "Deleted files"
![spectacle t28947](https://user-images.githubusercontent.com/277525/29871858-9c86aad8-8d8d-11e7-922c-76cf0539f0a4.png)

PVince81:
Fixed by reverting: https://github.com/owncloud/core/pull/28862

Original ticket was reopened to work on a less dangerous approach

PR #28819, opened 2017-08-28
[stable10] Fix overlapping filename and sharees

lordelix:
## Description
Fix overlapping names and "sharees" if the filename is a bit longer than usual.

## Related Issue

Fixes https://github.com/owncloud/enterprise/issues/2068
Master PR: https://github.com/owncloud/core/pull/28678

## How Has This Been Tested?

Visual test in latest Google Chome

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests pass.

PR #28862, opened 2017-08-30
Revert "[stable10] Fix overlapping filename and sharees"

PVince81:
Reverts owncloud/core#28819.

Reverted to not have a broken looking 10.0.3beta2


PVince81:
Just tested it and I can confirm that reverting this fixes https://github.com/owncloud/core/issues/28850

PR #28812, opened 2017-08-28
Improve market upgrade messages + new switch

PVince81:
## Description
Improve market upgrade messages.
Added switch to disable automatic app updates during occ upgrade.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28734

## How Has This Been Tested?
- [x] TEST: does not attempt to update apps if `upgrade.automatic-app-update` is set to false
- [x] TEST: attempts to update apps if `upgrade.automatic-app-update` is set to false
- [x] TEST: check message when occ upgrade with market app disabled: link to docs is there
- [x] TEST: check message when upgrading incompatible apps not present on marketplace: link to docs is there
- [x] TEST: market app auto-disables if "has_internet_connection" is false during occ ugprade.
- [x] TEST: if "has_internet_connection" is false and incompatible apps detected, stop update and mention those

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #28871, opened 2017-08-31
[stable10] Improve market upgrade messages + new switch

PVince81:
Backport of https://github.com/owncloud/core/pull/28812 to stable10

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1371,6 +1371,11 @@ $CONFIG = array(
 'upgrade.disable-web' => false,
 
 /**
+ * Automatic update of market apps, set to "false" to disable.
+ */
+'upgrade.automatic-app-update' => true,
+
+/**
  * Set this ownCloud instance to debugging mode
  *
  * Only enable this for local development and not in production environments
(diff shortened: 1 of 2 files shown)


jvillafanez (approved):

patrickjahns:
I've tested now all the possible scenario and noticed that the behavior between 10.0.2 and 10.0.3 will change in the following scenario:

Given we have an owncloud instance, that has not yet had the marketplace app (so anything before 10.x release). If we now have a tarball with the market app, it will not activate it automatically any longer, but also not suggest to activate if it is present.

Maybe we can change the messaging here? Or include this behavior in the documentation, so people will know about it?

PVince81:
> Given we have an owncloud instance, that has not yet had the marketplace app (so anything before 10.x release). If we now have a tarball with the market app, it will not activate it automatically any longer, but also not suggest to activate if it is present.

This sounds wrong, or a misunderstanding.

If you upgrade from OC < 10 it should enable the market app, unless "appstoreenabled => false" or "has_internet_connection => false" in config.php. From what you wrote it seems that it never enables it??

patrickjahns:
I tested it again today - it now works as intended

PR #28733, opened 2017-08-17
Introduce UserSession::verifyAuthHeaders which validates the session …

DeepDiver1975:
…against given auth headers like basic auth or oauth

## Related Issue
https://github.com/owncloud/core/issues/28707

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #28879, opened 2017-09-01
[stable10] Missing oc_authtoken (session) for native clients

### LINKED ISSUE #27845: Sessions of Desktop, iOS and Android clients should be shown
### Steps to reproduce

1. Log in oC server with a oC Client (Desktop, iOS, Android)
2. Browse to Admin --> Settings--> Personal --> Security

### Expected behaviour

The User agents from the sessions of Desktop, iOS and Android clients should be shown

### Actual behaviour

The User agents from Desktop, iOS and Android clients are not been shown 
### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

**ownCloud version:**

"version":"10.0.0.1" daily master

Fresh

No

**Are you using encryption:** 
No

**Logs**

```

```

### Client configuration

**browser**

Desktop Client, iOS, Android

It is shown only the ones related to browsers.

screen shot 2017-05-09 at 17 44 29

### LINKED ISSUE #28553: "Disconnect" application password should kill its open sessions
Directly related to https://github.com/owncloud/core/issues/27845

### Steps to reproduce
1. Create an application password and use it to login **in a Desktop/Mobile client**
2. Click on "Disconnect" the application

### Expected behavior
- The same behavior as it happens with the webUI: 
    - (After s.1) - The session appears listed in `https://<server>/index.php/settings/personal?sectionid=security#sessions` -> https://github.com/owncloud/core/issues/27845
    - (After s.2) - The sessions are deleted and refreshing the browser redirects to login
- The same effect as revoking an OAuth2 application authorized on `https://<server>/index.php/settings/personal?sectionid=security#oauth-2.0`: the clients re-asks for authorization (credentials, in this case) again. 

### Actual behavior
The session tokens are still valid **even when they don't appear listed in `authtoken` in the DB** (where are these stored then?) and the Desktop client is able to use them for syncing until the client is restarted (when it uses the password stored in the keychain to re-authenticate and fails)

### LINKED ISSUE #28881: Session can remove itself and does not really invalidate the session for all endpoints
### Steps to reproduce
1. With a valid `$SESSION` (after login) - list all current sessions:
```
$ curl 'https://<server>/index.php/settings/personal/authtokens' \
    -H 'OCS-APIREQUEST: true' \
    -H 'Cookie: $SESSION'

[
   {
      "id":1,
      "name":"Mozilla\/5.0 (Macintosh; Intel Mac OS X 10_12_5) AppleWebKit\/537.36 (KHTML, like Gecko) Chrome\/60.0.3112.113 Safari\/537.36",
      "lastActivity":"1504262361",
      "type":"0",
      "canDelete":false
   }
]
```
2. Remove itself (shouldn't be allowed - note the `"canDelete":false`)
```
$ curl 'https://<server>/index.php/settings/personal/authtokens/1' -X DELETE 
    -H 'OCS-APIREQUEST: true' \
    -H 'Cookie: $SESSION'
```
3. Query any endpoint to check how the `$SESSION` is still valid
4. Hit the first endpoint again:
```
$ curl -I 'https://<server>/index.php/settings/personal/authtokens' \
    -H 'OCS-APIREQUEST: true' \
    -H 'Cookie: $SESSION'

HTTP/1.1 503 Service Unavailable
[...]
```

### Expected behavior
1. If `"canDelete":false`, removing that session should be forbidden.
2. The `$SESSION` should be invalidated for all endpoints and querying `https://<server>/index.php/settings/personal/authtokens` shouldn't cause a server error

### Actual behavior
Described in repro steps. 

### Server configuration
* **ownCloud version:** 10.0.3 beta2 - after https://github.com/owncloud/core/pull/28879

IljaN:
## Description
oc_authtoken was not created for clients which use basic auth. (Probably everything, except browser and some dav-clients)

## Related Issue
fixes #27845
fixes #28553 
fixes #28881 

## Motivation and Context
This  bug was introduced with commit f1dc2d9 - "Validate session only if we are not in a basic auth request". 

Looking at the commit message the intent is not quite clear to me because the dektop client and all the mobile clients use basic auth at the moment. As the issue is relatively old I assume that some technical circumstances changed or that the validateSesion method is badly named because basic auth is technically not a persistent session trough requests. With this assumption the intent of the commit makes sense.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/legacy/app.php
+++ b/lib/private/legacy/app.php
@@ -133,24 +133,24 @@ class OC_App {
 		// once all authentication apps are loaded we can validate the session
 		if (is_null($types) || in_array('authentication', $types)) {
 			if (\OC::$server->getUserSession()) {
+				$request = \OC::$server->getRequest();
+				$session = \OC::$server->getUserSession();
 				$davUser = \OC::$server->getUserSession()->getSession()->get(\OCA\DAV\Connector\Sabre\Auth::DAV_AUTHENTICATED);
 				if (is_null($davUser)) {
-					\OC::$server->getUserSession()->validateSession();
+					$session->validateSession();
 				} else {
-					$request = \OC::$server->getRequest();
-					$userSession = \OC::$server->getUserSession();
 					/** @var \OC\Authentication\Token\DefaultTokenProvider $tokenProvider */
 					$tokenProvider = \OC::$server->query('\OC\Authentication\Token\DefaultTokenProvider');
 					$token = null;
 					try {
-						$token = $tokenProvider->getToken($userSession->getSession()->getId());
+						$token = $tokenProvider->getToken($session->getSession()->getId());
 					} catch (\Exception $ex) {
 						$password = null;
 						if (isset($_SERVER['PHP_AUTH_PW'])) {
 							$password = $_SERVER['PHP_AUTH_PW'];
 						}
 
-						$userSession->createSessionToken($request, $userSession->getUser()->getUID(), $userSession->getLoginName(), $password);
+						$session->createSessionToken($request, $session->getUser()->getUID(), $session->getLoginName(), $password);
 					}
 
 					if ($token) {
--- a/lib/private/legacy/app.php
+++ b/lib/private/legacy/app.php
@@ -136,6 +136,26 @@ class OC_App {
 				$davUser = \OC::$server->getUserSession()->getSession()->get(\OCA\DAV\Connector\Sabre\Auth::DAV_AUTHENTICATED);
 				if (is_null($davUser)) {
 					\OC::$server->getUserSession()->validateSession();
+				} else {
+					$request = \OC::$server->getRequest();
+					$userSession = \OC::$server->getUserSession();
+					/** @var \OC\Authentication\Token\DefaultTokenProvider $tokenProvider */
+					$tokenProvider = \OC::$server->query('\OC\Authentication\Token\DefaultTokenProvider');
+					$token = null;
+					try {
+						$token = $tokenProvider->getToken($userSession->getSession()->getId());
+					} catch (\Exception $ex) {
+						$password = null;
+						if (isset($_SERVER['PHP_AUTH_PW'])) {
+							$password = $_SERVER['PHP_AUTH_PW'];
+						}
+
+						$userSession->createSessionToken($request, $userSession->getUser()->getUID(), $userSession->getLoginName(), $password);
+					}
+
+					if ($token) {
+						$tokenProvider->updateToken($token);
+					}
 				}
 			}
 		}
--- a/settings/Controller/AuthSettingsController.php
+++ b/settings/Controller/AuthSettingsController.php
@@ -168,7 +168,15 @@ class AuthSettingsController extends Controller {
 	 * @return JSONResponse
 	 */
 	public function destroy($id) {
+
 		$user = $this->userManager->get($this->uid);
+		$currentToken = $this->tokenProvider->getToken($this->session->getId());
+
+		if ($currentToken && ($currentToken->getId() === intval($id))) {
+			return (new JSONResponse())->setStatus(Http::STATUS_CONFLICT);
+		}
+
+
 		if (is_null($user)) {
 			return [];
 		}


IljaN:
@phisch / @DeepDiver1975: Is it  possible to move that session check to somwhere else? Was suprised when I saw session handling code in a method called "loadApps" ;-)))

SamuAlfageme:
@IljaN cool, now I'm getting all the sessions logged back in the webUI 🎉 


Revoking though, (using the 'Delete' button next to them) is still not working (i.e. clients using basic auth are still able to sync and keep their session - it spawns in the table on next client request). So https://github.com/owncloud/core/issues/27845 might be deeper. Also I've noticed after removing a browser session (from a different one) and hitting: 

```
curl 'https://<server>/index.php/settings/personal/authtokens' \
    -H 'OCS-APIREQUEST: true' \
    -H 'Cookie: $SESSION'
```

Server replies with 503. The rest of the endpoints is fine. No exception is raised in the logs.

SamuAlfageme:
Described that one issue in https://github.com/owncloud/core/issues/28881 as this one was only for the regression display.

PVince81:
From my understanding from last time I looked at similar code, it seemed that there was a difference between basic auth and DAV authentication (still not sure where this is happening).

If we consider that both are two different things, then the old commit from @DeepDiver1975 was correct as it was only validating the session if we're NOT dealing with DAV authentication, so `$davUser` is null, in which case we're dealing with basic auth, not DAV auth.

The original issue though was that sessions did not appear in the list. Is there some piece of code inside of `validateSession` which we could extract and use here to make sure the session entry will be updated without validating ?

Now the other question would be whether it's really that bad to validate here additionally ?

Another concern: since you reversed the condition, it is likely that now basic auth is not getting validated any more. Maybe the `if` part should be removed completely ?

IljaN:
In my tests $davUser was always set using the sync client so go figure... Web interface does not go trough this codepath at all.

I also had the idea to pull the token update out but the code is buried in a private tokenProvider object inside session...

Yeah it would probably be the safest bet to always validate, basically reverting @DeepDiver1975 `s change.

IljaN:
Oh look what i found: 
```php
/**
	 * Add DAV authenticated. This should in an ideal world not be
	 * necessary but the iOS App reads cookies from anywhere instead
	 * only the DAV endpoint.
	 * This makes sure that the cookies will be valid for the whole scope
	 * @see https://github.com/owncloud/core/issues/22893
	 */
	$this->session->set(
		Auth::DAV_AUTHENTICATED, $this->getUser()->getUID()
	);
	return true;
```

PVince81:
It is possible that "DAV auth" means "basic auth done on a Webdav endpoint"

settings/Controller/AuthSettingsController.php
```diff
@@ -168,7 +168,15 @@ private function generateRandomDeviceToken() {
 	 * @return JSONResponse
 	 */
 	public function destroy($id) {
+
 		$user = $this->userManager->get($this->uid);
+		$currentToken = $this->tokenProvider->getToken($this->session->getId());
+
+		if ($currentToken->getId() == $id) {
```
PVince81: `===` ? See https://doc.owncloud.org/server/10.0/developer_manual/general/codingguidelines.html#operators

PVince81:
If validating in all cases isn't causing any other side effects, then I'm fine with this.

Needs to be retested at best will most authentication schemes:
- basic auth on Webdav
- basic auth on OCS
- web UI session
- shibboleth ?
- auth with application token instead of password
- oauth2

IljaN:
Will figure out how to test most of these, not sure: 
...how to setup/test shibboleth and oauth2
... what exactly OCS is.

PVince81:
@IljaN get help from QA or @SamuAlfageme for testing these cases

SamuAlfageme:
Some results after some quick checks: 

- basic auth on Webdav ✅ 
- basic auth on OCS ✅ 
- web UI session ✅ 
- shibboleth ❌ 
- auth with application token instead of password ✅ 
- oauth2 ✅ 

sessions


About Shib, I'm not sure if the sessions ever appeared in this section (or they have a separate pane somewhere; kind of like it happens with OAuth) 

Here ya go with a normal SAML request if that can help out determining if something is missing:

<details>
<summary>SAML-Authenticated <code>PROPFIND</code></summary>

```
curl \
-H 'Cookie:oc_sessionPassphrase=7hZcZNUIb3AaiNocC%2BIXh5aAYM36%2F3R1WG6i6OltcbqdYsJhHTorkR8%2BgQcmTZL7ax7CuavRwJL4GpNb2uBGPjEN2LsM2aMZ58pVCHlMzu5bMMkjzuPAvyNhwr6dgajQ; ocgwv1nxeve8=iq8npgptet00ocj15qtc2vr4g6; _shibsession_64656661756c7468747470733a2f2f646f636b65722e6f632e736f6c6964676561722e65733a31303237312f6f632d73686962=_58f8ec23b20e1ac03e7d389bad9d7197' \
-H 'Content-Length:114' \
 -X PROPFIND 'https://<server>/<shib-path>/remote.php/dav/files/alterego@testshib.org/' --data-binary '<?xml version="1.0" ?>
<d:propfind xmlns:d="DAV:">
  <d:prop>
    <d:getlastmodified />
  </d:prop>
</d:propfind>
'
```

</details>
<br/>

@IljaN maybe we can also get the `_shibsession_*=_*'`s logged in the session table somehow. I don't really know the specifics about it maybe @butonic or @DeepDiver1975 can help here. 

@PVince81 about

> basic auth on OCS 

I'm getting the "last activity" time updated from `ocs` calls, so I'd say that one works as expected as well. (they don't and I guess shouldn't originate new entries)

PVince81:
@SamuAlfageme for shib, try with shib and the latest OC version before the regression to find out whether it ever worked before.

PVince81:
this looks suspicious:
```

auth.access files app with basic authprovisioning.Making a web request with an enabled usersharing.Empting trashbinsharing.orphaned sharessharing.deleting file out of a share as recipient creates a backup for the ownersharing.deleting folder out of a share as recipient creates a backup for the ownersharing.moving file out of a share as recipient creates a backup for the ownersharing.moving folder out of a share as recipient creates a backup for the ownertrashbin-new-endpoint.deleting a file moves it to trashbintrashbin-new-endpoint.deleting a folder moves it to trashbintrashbin-new-endpoint.deleting a file of a shared folder moves it to trashbintrashbin-new-endpoint.deleting a shared folder moves it to trashbintrashbin-new-endpoint.deleting a received folder doesn't move it to trashbintrashbin-new-endpoint.deleting a file in a received folder it is moved to the trashbintrashbin-new-endpoint.Trashbin can be emptiedtrashbin-new-endpoint.A deleted file can be restoredtrashbin-new-endpoint.Deleting a folder into external storage moves it to the trashbintrashbin-new-endpoint.Deleting a file into external storage moves it to the trashbin and can be restoredtrashbin-new-endpoint.deleting a file moves it to trashbintrashbin-new-endpoint.deleting a folder moves it to trashbintrashbin-new-endpoint.deleting a file of a shared folder moves it to trashbintrashbin-new-endpoint.deleting a shared folder moves it to trashbintrashbin-new-endpoint.deleting a received folder doesn't move it to trashbintrashbin-new-endpoint.deleting a file in a received folder it is moved to the trashbintrashbin-new-endpoint.Trashbin can be emptiedtrashbin-new-endpoint.A deleted file can be restoredtrashbin-new-endpoint.Deleting a folder into external storage moves it to the trashbintrashbin-new-endpoint.Deleting a file into external storage moves it to the trashbin and can be restoredtrashbin-new-endpoint.Deleting an updated file into external storage moves it to the trashbin and can be restored
--
```

IljaN:
All tests should pass now :crossed_fingers:

IljaN:
@SamuAlfageme All tests are green now, please retest.

PVince81:
moving to planned (10.0.4), too risky for post-RC.

Also I'd like to have a review from @DeepDiver1975

PVince81:
@DeepDiver1975 can you review ?

@SamuAlfageme can you retest ?

lib/private/legacy/app.php
```diff
@@ -129,6 +129,26 @@ public static function loadApps($types = null) {
 				$davUser = \OC::$server->getUserSession()->getSession()->get(\OCA\DAV\Connector\Sabre\Auth::DAV_AUTHENTICATED);
 				if (is_null($davUser)) {
 					\OC::$server->getUserSession()->validateSession();
+				} else {
+					$request = \OC::$server->getRequest();
+					$userSession = \OC::$server->getUserSession();
```
DeepDiver1975: Was this code change tested with occ commands as well as installation and upgrade scenarios?

DeepDiver1975 (changes requested):
Please cleanup the git history. THX

IljaN:
@DeepDiver1975 Squashed

IljaN:
@DeepDiver1975 ping

IljaN:
ping

PVince81:
@IljaN please answer here https://github.com/owncloud/core/pull/28879#discussion_r140716530

lib/private/legacy/app.php
IljaN: @DeepDiver1975 What scenarios do you have in mind? How are occ commands related to authentication?

DeepDiver1975: occ calls loadApps as well - there is no request object then and no usersession object.
thats my point ... but maybe the code does not even go there .. no idea - please test - THX

IljaN:
Upgrade works

PVince81:
@davitol can you regression test auth with this PR, same as with https://github.com/owncloud/core/pull/28733 ?

If all goes well we can merge this and

- [ ] forward port to master

PVince81:
rebased for CI (Drone didn't start...)

davitol:
> @davitol can you regression test auth with this PR, same as with #28733 ?
> 
> If all goes well we can merge this and

Uploading a file via webDAV using cadaver. The session is not shown in the list

screen shot 2017-11-08 at 12 26 30

IljaN:
@davitol I just did a quicktest with desktop client, both web-session and desktop client session are displayed. Will try with cadavre to see what the difference is.

davitol:
> @davitol I just did a quicktest with desktop client, both web-session and desktop client session are displayed. Will try with cadavre to see what the difference is.

With Desktop worked for me, but for example with iOS app it shows multiple sessions, not sure if due to iOS app or for the PR (see below)

screen shot 2017-11-08 at 12 50 16

IljaN:
@davitol I can confirm that cadavre session is not displayed... investigating

IljaN:
@davitol We figured out that cadaver does not support cookies thus doing basic auth on every request. This is why nothing is displayed in the sessions view because there is technically no session established. This would explain why Cyberduck is displayed (it probably supports cookies) 

After discussing with @DeepDiver1975 the expected behavior is that nothing is displayed for clients which do not support cookies. 

If we would "fix" that, we would have a entry for every request made by cadaver.

As for the iOS issue I need to find someone with a iPhone to be able to debug this...

davitol:
> PS: I will try to replicate the issue with Android

@IljaN With Android it went fine for me

screen shot 2017-11-08 at 17 39 08

jesmrec:
@IljaN the point is the way the iOS client is handling with the cookies. Some request are including authorization basic header but no cookies, therefore the server answers with a new cookie, and i guess that in this case a new row is included there. We will take care

PVince81:
@jesmrec so can we merge this or will this require modifications of the PR code ?

jesmrec:
@PVince81 you can merge

PVince81:
@jesmrec thanks.

Now for a last CI battle then we can merge.

PVince81:
Rebased for CI, hopefully for higher timeout values

PVince81:
codecov errors about the legacy app code, passing on this one 🙈

PVince81:
@IljaN please forward port to master

PR #28880, opened 2017-09-01
[stable10] Increment required only when encryption is enabled

sharidas:
Increment to files in filecache is required only when
encryption is enabled. Else use the version as it is.

## Description

When encryption is not enabled then no need to go through each and every step in encryption wrapper's fopen. Instead we can get another storage wrapper's fopen handler. Moreover this change also fixes the restore from trash which was broken.

## Related Issue

https://github.com/owncloud/core/issues/28780

## Motivation and Context

This change would resolve the exceptions thrown when files are are moved from one storage to trash. Also it resolves the restore functionality of trash back to any storage.

## How Has This Been Tested?

- [x] Create external storage with SFTP
- [x] Create a folder test under shared folder.
- [x] upload a file under test
- [x] Delete the test folder
- [x] Try to access the test folder from the trash
- [x] No exception logs found
- [x] Verified the restore works from the trash. Even verified with multiple versions of the file.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -385,7 +385,7 @@ class Encryption extends Wrapper {
 		$header = $this->getHeader($path);
 		$signed = (isset($header['signed']) && $header['signed'] === 'true') ? true : false;
 		$fullPath = $this->getFullPath($path);
-		$encryptionModuleId = ($encryptionEnabled) ? $this->util->getEncryptionModuleId($header): "";
+		$encryptionModuleId = $this->util->getEncryptionModuleId($header);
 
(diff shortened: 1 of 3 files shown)


SergioBertolinSG:
Locking problem appears here while testing this.

sharidas:
This is a backport of https://github.com/owncloud/core/pull/28820

SergioBertolinSG (approved):
Works fine. Locking problem is not related. It is just not fixed yet.

PVince81:
@SergioBertolinSG what locking problem ? Is there a ticket ?

PVince81 (approved):
:+1:

SergioBertolinSG:
@PVince81 you already know it: https://github.com/owncloud/core/issues/28779

PR #28886, opened 2017-09-03
ocPath and app enable disable refactoring

phil-davis:
## Description
1) Moved the "memory" of $ocPath (that is delivered in $scope) down into SetupHelper.
2) Make setOcPath() to be able to set it just once - that is effectively a "constructor" for the SetupHelper class (which does not have a real constructor because it is just a bunch of static utility functions and thus never gets instantiated)
3) Call setOcPath() in the BeforeScenario so that SetupHelper knows it "up front".
4) Remove the need to pass around $ocPath in loads of places.
5) Refactor until the IDE told me it was happy and the tests pass!

## Motivation and Context
Started out making enableApp() and disableApp() be able to be more easily called so that there would not be so much boiler-plate code in the UI tests BeforeScenario in each set of app tests, reducing duplication. I first changed those to take $scope as parameter and work out $ocPath internally.

Then as the consequences rolled on, more refactoring because there were more places that then needed similar treatment.

This makes it simpler to use various SetupHelper functions from within app UI test suites (and also from core).

## How Has This Been Tested?
Full runs of core, files_texteditor and firewall tests using this as a base.

## Types of changes
- [x] Refactoring

## Checklist:

### DIFF
--- a/tests/TestHelpers/LoggingHelper.php
+++ b/tests/TestHelpers/LoggingHelper.php
@@ -31,12 +31,11 @@ class LoggingHelper {
 	/**
 	 * returns the log file path
 	 * 
-	 * @param string $ocPath
 	 * @throws \Exception
 	 * @return string
 	 */
-	public static function getLogFilePath($ocPath) {
-		$result = SetupHelper::runOcc(['log:owncloud'], $ocPath);
+	public static function getLogFilePath() {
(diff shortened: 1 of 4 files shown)


individual-it (approved):

phil-davis:
Backport stable10 #28893

PR #28831, opened 2017-08-29
Better timeout detection in web UI uploads + chunked uploads

PVince81:
## Description
Detects timeouts during web UI uploads and displays a notification when it happens.
Also prevents the global ajax error detection to prevent the full page reload.

## How Has This Been Tested?
- [x] TEST: Set client-side "timeout: 5000" in file-upload.js's `fileUploadParam`. Add `sleep(500)` in `OCA\DAV\Connector\Sabre\File::put()`. Upload < 10 mb file. Notification about timeout appears.
- [x] TEST: Set client-side "timeout: 5000" in file-upload.js's `fileUploadParam`. Upload 100 mb file. Notification about timeout appears.
- [x] TEST: IE11: Add `sleep(500)` in `apps/dav/lib/Upload/UploadFolder::createFile()`, upload 100 mb file. Notification appears after 2 minutes
- [ ] TEST: IE11: Add `sleep(500)` in `apps/dav/lib/Upload/AssemblyStream::stream_open()`. Upload 100 mb file, wait. Notification appears after 2 minutes of stalling at the end of the progress bar.
=> cannot test, it seems IE11 doesn't have a timeout on MOVE, only on PUT

Note: remove extra modifications between each test.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #28896, opened 2017-09-04
[stable10] Better timeout detection in web UI uploads + chunked uploads

PVince81:
Backport of https://github.com/owncloud/core/pull/28831 to stable10


IljaN (approved):
:+1:

PR #28899, opened 2017-09-04
Fix error logs due to deletion of keys

sharidas:
When files are deleted and moved to trashbin,
the keys are deleted. This results in exception
due to file not exist. And hence this change
helps to fix the issue. The issue is caused
when the version of the files are tried to
retain in the trashbin.

## Description

When the files are moved to trashbin, the retain version causes problem. Which means the retainVersion in the trashbin tries to access the file keys which are already deleted. So this causes exception.

## Related Issue

https://github.com/owncloud/core/issues/28883

## Motivation and Context

This change tries to solve the issue caused by accessing file keys which are already deleted during the transfer of files to trashbin.

## How Has This Been Tested?

- [x] Enable encryption with master key
- [x] Enable external sotrage with SFTP
- [x] Upload a file to SFTP folder
- [x] Update the file.
- [x] Delete the file . The log file is clean and no exceptions.
- [x] Restore the file. The log file is clean and no exceptions.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_trashbin/lib/Trashbin.php
+++ b/apps/files_trashbin/lib/Trashbin.php
@@ -208,7 +208,7 @@ class Trashbin {
 		$view = new View('/');
 		self::copy_recursive($source, $target, $view);
 
-		self::retainVersions($targetFilename, $owner, $ownerPath, $timestamp, true);
+		self::retainVersions($targetFilename, $owner, $ownerPath, $timestamp, null, true);
 
 		if ($view->file_exists($target)) {
 			self::insertTrashEntry($owner, $targetFilename, $targetLocation, $timestamp);
@@ -304,7 +304,7 @@ class Trashbin {
 			\OCP\Util::emitHook('\OCA\Files_Trashbin\Trashbin', 'post_moveToTrash', ['filePath' => Filesystem::normalizePath($file_path),
 				'trashPath' => Filesystem::normalizePath($filename . '.d' . $timestamp)]);
 
-			self::retainVersions($filename, $owner, $ownerPath, $timestamp);
+			self::retainVersions($filename, $owner, $ownerPath, $timestamp, $sourceStorage);
 
 			// if owner !== user we need to also add a copy to the owners trash
 			if ($user !== $owner) {
@@ -331,8 +331,18 @@ class Trashbin {
 	 * @param integer $timestamp when the file was deleted
 	 * @param bool $forceCopy true to only make a copy of the versions into the trashbin
 	 */
-	private static function retainVersions($filename, $owner, $ownerPath, $timestamp, $forceCopy = false) {
+	private static function retainVersions($filename, $owner, $ownerPath, $timestamp, $sourceStorage = null, $forceCopy = false) {
 		if (\OCP\App::isEnabled('files_versions') && !empty($ownerPath)) {
+
+			$copyKeysResult = false;
+
+			/**
+			 * In case if encryption is enabled then we need to retain the keys which were
+			 * deleted due to move operation to trashbin.
+			 */
+			if ($sourceStorage !== null) {
+				$copyKeysResult = $sourceStorage->retainKeys($filename, $owner, $ownerPath, $timestamp, $sourceStorage);
+			}
 
 			$user = User::getUser();
 			$rootView = new View('/');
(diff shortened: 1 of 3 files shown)


apps/files_trashbin/lib/Trashbin.php
```diff
@@ -208,7 +208,7 @@ public static function copyBackupForOwner($ownerPath, $owner, $timestamp) {
 		$view = new View('/');
 		self::copy_recursive($source, $target, $view);
 
-		self::retainVersions($targetFilename, $owner, $ownerPath, $timestamp, true);
+		self::retainVersions($targetFilename, $owner, $ownerPath, $timestamp, false,true);
```
PVince81: missing space after ","

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -331,9 +331,19 @@ public static function move2trash($file_path) {
 	 * @param integer $timestamp when the file was deleted
 	 * @param bool $forceCopy true to only make a copy of the versions into the trashbin
 	 */
-	private static function retainVersions($filename, $owner, $ownerPath, $timestamp, $forceCopy = false) {
+	private static function retainVersions($filename, $owner, $ownerPath, $timestamp, $sourceStorage= false, $forceCopy = false) {
```
PVince81: can we use `null` instead of `false` ?

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -331,9 +331,19 @@ public static function move2trash($file_path) {
...
 		if (\OCP\App::isEnabled('files_versions') && !empty($ownerPath)) {
 
+			$copyKeysResult = false;
+
+			if (\OC::$server->getEncryptionManager()->isEnabled()) {
```
PVince81: From what I see, in the past we never had encryption code directly in the trashbin.

There must be logic elsewhere already about moving/copying encryption keys to trash, it is likely that this logic exists directly in the encryption app. Can you find it ?

apps/files_trashbin/lib/Trashbin.php
phil-davis: and there is missing space between ``$sourceStorage`` and ``=``

apps/files_trashbin/lib/Trashbin.php
sharidas: @PVince81 So here is where the change triggers for deletion of keys to trash:
https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L644 which leads to https://github.com/owncloud/core/blob/master/apps/files_trashbin/lib/Storage.php#L119 and then the call goes down to https://github.com/owncloud/core/blob/master/apps/files_trashbin/lib/Storage.php#L163. From there finally the call reaches https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L250.

So should we move the code somewhere around https://github.com/owncloud/core/blob/master/apps/files_trashbin/lib/Storage.php ?

PVince81:
Can you make sure the following still works correctly:

- single storage: deleted encrypted file with versions restored: file and versions can be downloaded fine
- cross-storage: deleted encrypted file with versions restored: file and versions can be downloaded fine

I see your test steps above but they don't seem to cover checking that the restored versions still work.

sharidas:
@PVince81 Below are the tests done

**Single Storage**
- [x] Enable encryption
- [x] Upload a file foo.txt
- [x] Rename the file foo.txt to Test.txt 
- [x] Update the file by adding a line "Version 1 ..." 
beforedeletion

- [x] Verify that 2 versions of the file are there from UI
- [x] Delete the file 
deleted

- [x] Restore the file
- [x] Verified the 2 versions of the file are there.. 
afterrestore

- [x] Repeated the above steps to create one more version and the restore was successfully done. The 3 versions were available after restore.
- [x] Verified from the terminal after the download. 
terminalverification

**Cross Storage**
- [x] Enable encryption
- [x] Enable external storage with SFTP
- [x] Upload a file to SFTP folder
- [x] Update the file by adding line "Version 1 ..." 
versionupdatebeforedeletion

- [x] Verify the version from UI.
- [x]  Delete the file 
filedeleted

- [x] Restore the file
- [x] Verify the versions in the UI ( 2 versions ) 
filerestoreanddownloaded

- [x] Repeat the steps for 3rd Version.
- [x] Verified the versions downloaded in the terminal. 
terminalverification

PVince81 (approved):
:+1:

PVince81:
Please backport to stable10, set milestone to "planned" for the next release there

sharidas:
Back port PR: https://github.com/owncloud/core/pull/28934

PR #28619, opened 2017-08-08
Default share perms

PVince81:
## Description
Default share permissions setting for local shares.

The default permissions are exposed via capabilities.
To make it visible to the JS frontend code, the capabilities are now added in "js/config.php" and accessible through `OC.getCapabilities()`. This will make passing such config values more convenient in the future especially for values that other clients need.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28384.

## Motivation and Context
See ticket

## How Has This Been Tested?
- [x] TEST: manual testing
- [x] TEST: add unit tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #28903, opened 2017-09-04
[stable10] Default share perms

PVince81:
Backport of https://github.com/owncloud/core/pull/28619 to stable10

### DIFF
--- a/settings/js/admin.js
+++ b/settings/js/admin.js
@@ -57,7 +57,7 @@ $(document).ready(function(){
 		}
 	});
 
-	$('#shareAPI input:not(#excludedGroups)').change(function() {
+	$('#shareAPI input:not(.noautosave)').change(function() {
 		var value = $(this).val();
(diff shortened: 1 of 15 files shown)


settings/js/admin.js
```diff
@@ -87,5 +87,27 @@ $(document).ready(function(){
...
+
+		// always set read permission
+		value |= OC.PERMISSION_READ;
+
+		// this will trigger the field's change event and will save it
+		$el.val(value).change();
+
+		ev.preventDefault();
+
+		return false;
```
jvillafanez: According to what I've read, `return false` implies a `preventDefault` and `stopPropagation` on the event, so we can remove the return or the preventDefault, depending on what we want.

PVince81: yeah... I just wanted to be safe. I'll remove it next time I touch this code.

IljaN (approved):

PR #28872, opened 2017-08-31
Fix overlapping selectively in default fileslist

lordelix:
## Description

Scope changes previously done in https://github.com/owncloud/core/pull/28678 to the default files-list view only.
Aka: `#app-content-files`

## Related Issue
Fixes https://github.com/owncloud/core/issues/28850
Fixes https://github.com/owncloud/enterprise/issues/2068

## How Has This Been Tested?
Visual test in FF, Chrome & IE11

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #28906, opened 2017-09-05
[stable10] Fix overlapping selectively in default fileslist

phil-davis:
Backport #28872


PVince81:
the reason for the diff differences is that I only reverted the old breaking change on stable10, not master.

PVince81 (approved):
:+1:

Issue #24654, opened 2016-05-17
Fix deprecated usages of IRouter


rullzer:
Since 9.0 the IRouter class is depracted (by @LukasReschke). Since well there is no real need to expose this to the big scary outside world.

However it is still referenced all over the place (almost all apps etc). We should clean this up properly.

PR #21032, opened 2015-12-08
Router error handling + Base.php

LukasReschke:
- Some functions were entirely broken
- PHPDoc etc...

Oh. And this breaks public API. But this API was already throwing PHP fatal errors before since it was simply broken in stable releases anyways so I don't really care too much about it. Result is the same… :speak_no_evil: 

@icewind1991 @DeepDiver1975 THX

PR #28910, opened 2017-09-05
IRouter is deprecated, deprecate it in the server container?

butonic:
The IRouter interface was deprecated two years ago with  https://github.com/owncloud/core/commit/34c8249799c2fc89af021a085f4b15c3156d1b11 in https://github.com/owncloud/core/pull/21032

But we use it happily in core to match the url and hand over execution to the responsible controller in base.php

Why? What should app developers use instead? What should we use in core instead?

- [ ] add the answer to PHPDoc

### DIFF
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -1191,6 +1191,8 @@ class Server extends ServerContainer implements IServerContainer, IServiceLoader
 	 * Returns a router for generating and matching urls
 	 *
 	 * @return \OCP\Route\IRouter
+	 * @since 7.0.0
+	 * @deprecated 9.0.0
 	 */
 	public function getRouter() {
 		return $this->query('Router');


PVince81:
Fixes https://github.com/owncloud/core/issues/24654

phil-davis:
Not related to this actual PR.
Jenkins here has got a problem with ``sharing-v1.feature:914``

```
    Then as "user1" gets properties of folder "/merge-test-outside-groups-renamebeforesecondshare-renamed" with                                        # FeatureContext::asGetsPropertiesOfFolderWith()
      | {http://owncloud.org/ns}permissions |
    And the single response should contain a property "{http://owncloud.org/ns}permissions" with value "SRDNVCK"                                       # FeatureContext::theSingleResponseShouldContainAPropertyWithValue()
      Cannot find property "{http://owncloud.org/ns}permissions" with "SRDNVCK" (Exception)
```
and I have noticed other Jenkins runs that are getting this, but I think not all.
I presume there is some timing issue or ??? happening.

DeepDiver1975:
> But we use it happily in core to match the url and hand over execution to the responsible controller in base.php

all this could be done on internal classes as well - no need for an ocp interface.
it basically was deprecated because in the good old days an instance of IRouter was passed into any app's routes.php and the app code used the interface to register routes.

as of today routes.php simply returns an array of routes.

we can really deprecate the router in server

DeepDiver1975 (approved):

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

PR #28911, opened 2017-09-05
Handle no user session in isSharingDisabledForUser()

phil-davis:
## Description
If isSharingDisabledForUser() is called when there is no user context, then set user to null.

## Motivation and Context
In the log of Travis output, I noticed:
```
owncloud.log:
{"reqId":"pqAijCPvaiHvZolHHCdO","level":3,"time":"2017-09-05T08:06:49+00:00","remoteAddr":"","user":"--","app":"PHP","method":"--","url":"--","message":"Undefined variable: user at \/home\/travis\/build\/[secure]\/core\/lib\/public\/Util.php#193"}
```

It is nice to get rid of undefined variable messages.

## How Has This Been Tested?
Try logging in and creating files... as a user with this code in place. See that it does not explode.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/tests/Controller/ViewControllerTest.php
+++ b/apps/files/tests/Controller/ViewControllerTest.php
@@ -87,21 +87,21 @@ class ViewControllerTest extends TestCase {
 		$this->rootFolder = $this->createMock('\OCP\Files\Folder');
 		$this->viewController = $this->getMockBuilder('\OCA\Files\Controller\ViewController')
 			->setConstructorArgs([
-			'files',
-			$this->request,
-			$this->urlGenerator,
-			$this->l10n,
-			$this->config,
-			$this->eventDispatcher,
-			$this->userSession,
-			$this->appManager,
-			$this->rootFolder
-		])
-		->setMethods([
-			'getStorageInfo',
-			'renderScript'
-		])
-		->getMock();
+				'files',
+				$this->request,
+				$this->urlGenerator,
+				$this->l10n,
+				$this->config,
+				$this->eventDispatcher,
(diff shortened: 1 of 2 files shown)


PVince81 (dismissed):
:+1: nice catch, I saw this error many times

phil-davis:
Hmmm - I broke unit tests, better have a look.

phil-davis:
The ViewControllerTest does not have a user context, so with my change, its view of things no longer includes the ability to share (outwardly). Thus those parts are removed from the expected appNavigation and appContents.

phil-davis:
@PVince81 please confirm if that is OK to just remove those array elements from the expected test result.

(alternative is to work out how to have a proper user context there so as to get a different answer from isSharingDisabledForUser() )

phil-davis:
Previously I tried returning ``isSharingDisabledForUser()`` as ``true`` when there was no user session found. But that causes problems for stuff like public shares, where the incoming request has no user authentication.
The neatest looking solution is just to set ``$user`` to ``null`` when there is no user session. ``self::$shareManager->sharingDisabledForUser($user)`` already handles that fine and returns ``false`` and so when there is no user session, sharing is NOT disabled. This lets public shares and the like work.

While messing with ``ViewControllerTest.php`` my IDE indented some stuff to make it consistent. No functional change there, but just nice to have improved indenting while I am there.

@PVince81 please re-review.

phil-davis:
Backport stable10 #28915 
(prospective)

phil-davis:
@PVince81 please merge if happy - and merge backport also.

Issue #28916, opened 2017-09-06
File row sometimes not visible in UI tests


phil-davis:
### Steps to reproduce
1. Run the automated UI tests (specially those that rename/share/delete files)
2. Make the browser window height smaller, or just be unlucky

### Expected behaviour
The tests should paqss

### Actual behaviour
When finding a file in the list of files on the files page, the code checks for the file. If it is not there then it scrolls down to the bottom, which will cause the web UI to request and load more files in the list. This process might load more files than fit in a single scroll window. In that case the desired file might be found in the list, but is not actually visible in the window.

Subsequent code does stuff like try to click on the file row element. That fails if the element does not happen to be visible in the window.

### Server configuration
core master or stable10 running tests in a dev environment, or in Travis+SauceLabs

phil-davis:
@individual-it I think this is fixed with some better scrolling of the files list that you did?

individual-it:
should be fixed by https://github.com/owncloud/core/pull/29124

PR #28917, opened 2017-09-06
[WIP] scroll element into view in findFileRowByName

phil-davis:
## Description
When the file row element has been found, scroll it into view. Then later steps can rely on the fact that, if the file row element was found then it has also been scrolled into view.

Note: while investigating this problem, I also discovered an issue with ``getCoordinatesOfElement()`` - if ``getXpath()`` of the element ends up having double-quotes in it, then the code will fail (e.g. an xpath for an element that is a file name containing double quotes). Double quotes in the xpath need to be escaped to send into ``document.evaluate`` . I discovered this while trying to use ``getCoordinatesOfElement()`` to scroll the files list to a particular position. In the end the solution here does not do that, because ``scrollIntoView`` turned out to be easy and work.

## Related Issue
#28916 

## Motivation and Context
Make the UI tests more reliable

## How Has This Been Tested?
Local runs of relevant UI tests.
Travis will tell us if it at least works.

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,22 +126,19 @@ matrix:
         apt: *common_apt
         hosts: *common_hosts
         sauce_connect: true
-    - if: type = cron
-      php: 5.6
+    - php: 5.6
       env: DB=pgsql TC=selenium BROWSER="internet explorer" BROWSER_VERSION="11.0" BEHAT_SUITE="other" PLATFORM="Windows 7" TEST_DAV=0
       addons:
         apt: *common_apt
         hosts: *common_hosts
         sauce_connect: true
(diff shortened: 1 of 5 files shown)


individual-it:
does it work with IE?

phil-davis:
I tried on my fork and it looks like it does not work on IE11 👎 - what a pain! Thinking...

phil-davis:
See PR #29124 for an implementation of something to scroll things into view that works on IE as well as other browsers. It would be much easier if ``scrollIntoView()`` "just worked".

Closing here. Other ways to do better scrolling of lists etc. can be investigated some time!

PR #27068, opened 2017-02-01
cache loadUser if not exists

butonic:
## TL;dr

This PR caches a non existing user in the DB, killing ~30 queries on every request when an ldap user logs in.

## Long version

user_ldap (type authentication) depends upon the AvatarManager, which in turn triggers a setupfs(), which in turn pulls in all apps of type 'filesystem' (activity, dav, files, files_sharing, files_trashbin and files_versions. Probably objectstore, search_elastic / search_lucene and whatever app has type filesystem).

Loading an app causes the I10n to be initialized whenever \OC::$server->getL10N() is used in the app.php directly or if it is a dependency of the other classes used when enabling the app.

L10N tries to get the user object which somewhere calls userExists() ... aaaand that is why for every request we do ~30 ` SELECT "uid", "displayname" FROM "oc_users" WHERE LOWER("uid") = LOWER(?)` queries. when loading the user_ldap app only user_database is enabled and since a non existing user is not cached the userExists check is constantly walking to the db ...

There are probably other things doing a userExists check, but L10N is the most frequent. 

@DeepDiver1975 I don't know if the user backend loading order may be the cause for the mixed UI languages issue you were investigating. It sure smells like it.

Yes, this only removes a symptom of the mess described above. Maybe this is no longer an issue whan the user table has been merged. I don't know. Looking at the app loading made my head hurt because we are constantly calling \OC_App::registerAutoloading()  for apps that don't even match the type https://github.com/owncloud/core/blob/fe1806c5a7f09a9c6cf6669cdce7986b1df9231d/lib/private/legacy/app.php#L108-L114. And then we are calling it again when loading the app: https://github.com/owncloud/core/blob/fe1806c5a7f09a9c6cf6669cdce7986b1df9231d/lib/private/legacy/app.php#L143-L143 :hankey: 

cc @PhilippSchaffrath

PR #28927, opened 2017-09-06
Cache non-existing accounts

butonic:
followup to https://github.com/owncloud/core/pull/27068

### DIFF
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -33,6 +33,7 @@
 
 namespace OC\User;
 
+use OC\Cache\CappedMemoryCache;
 use OC\Hooks\PublicEmitter;
 use OCP\AppFramework\Db\DoesNotExistException;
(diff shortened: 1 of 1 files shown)


butonic:
hm jenkins failed on oracle, the other dbs were ok

PVince81:
@butonic I'd appreciate at least a quick explanation why the last PR wasn't enough and what this commit is doing, for posterity. Normally as per the PR template you should provide way more information than just a single line.

butonic:
@PVince81 This pr prevents the user manager from eating memory by using a capped cache. Helps when iterating over all users.

PVince81:
deja vu... or maybe it's the same code I reviewed a while ago and it didn't have a capped cache ?

code looks fine.

ready for review ? (status is still "developing")

DeepDiver1975 (approved):

PR #28924, opened 2017-09-06
[stable10] Web UI goes back to old DAV endpoint

PVince81:
Backport of https://github.com/owncloud/core/pull/28914 to stable10

PR #28932, opened 2017-09-06
Update changelog for 10.0.3RC1

PVince81:
be5e84ce0d79bc91e96866b768e1c500596a423c..stable10

Note: I removed the "new DAV" entry because we are switching back to old DAV in https://github.com/owncloud/core/pull/28924, so no point in mentioning something that was never released.


phil-davis (approved):
LGTM

PVince81:
Thanks @phil-davis

PR #28934, opened 2017-09-07
[stable10] Fix error logs due to deletion of keys

sharidas:
When files are deleted and moved to trashbin,
the keys are deleted. This results in exception
due to file not exist. And hence this change
helps to fix the issue. The issue is caused
when the version of the files are tried to
retain in the trashbin.

## Description

When the files are moved to trashbin, the retain version causes problem. Which means the retainVersion in the trashbin tries to access the file keys which are already deleted. So this causes exception.

## Related Issue

https://github.com/owncloud/core/issues/28883

## Motivation and Context

This change tries to solve the issue caused by accessing file keys which are already deleted during the transfer of files to trashbin.

## How Has This Been Tested?

- [x] Enable encryption with master key
- [x] Enable external sotrage with SFTP
- [x] Upload a file to SFTP folder
- [x] Update the file.
- [x] Delete the file . The log file is clean and no exceptions.
- [x] Restore the file. The log file is clean and no exceptions.
- [x] Download the versions of the file.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_trashbin/lib/Storage.php
+++ b/apps/files_trashbin/lib/Storage.php
@@ -200,6 +200,28 @@ class Storage extends Wrapper {
 	}
 
 	/**
+	 * Retain the encryption keys
+	 *
+	 * @param $filename
+	 * @param $owner
+	 * @param $ownerPath
+	 * @param $timestamp
+	 * @param $sourceStorage
+	 * @return bool
+	 */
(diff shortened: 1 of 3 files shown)


PVince81 (approved):
:+1: to merge for 10.0.4

PR #28949, opened 2017-09-08
integration tests sharing-v1 962 as an user0

phil-davis:
## Description
Add a step to put the subsequent share delete and checks in the context of "user0".

## Related Issue
#28950 

## Motivation and Context
integration test fails

## How Has This Been Tested?
locally run:
```
phil@phil-VirtualBox:~/core/tests/integration$ ./run.sh features/sharing-v1.feature:962
```

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/sharing-v1.feature
+++ b/tests/integration/features/sharing-v1.feature
@@ -6,7 +6,7 @@ Feature: sharing
 	Scenario: Creating a new share with user
 		Given user "user0" exists
 		And user "user1" exists
-		And As an "user0"
+		And as an "user0"
 		When sending "POST" to "/apps/files_sharing/api/v1/shares" with
 			| path | welcome.txt |
 			| shareWith | user1 |
@@ -18,7 +18,7 @@ Feature: sharing
 		Given user "user0" exists
 		And user "user1" exists
 		And group "sharing-group" exists
-		And As an "user0"
+		And as an "user0"
 		When sending "POST" to "/apps/files_sharing/api/v1/shares" with
 			| path | welcome.txt |
 			| shareWith | sharing-group |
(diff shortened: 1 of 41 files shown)


phil-davis:
TBH I do not understand why this has been passing when run in CI on Jenkins.

tests/integration/features/sharing-v1.feature
```diff
@@ -965,6 +965,7 @@ Feature: sharing
 		And group "sharing-group" exists
 		And user "user0" belongs to group "sharing-group"
 		And file "welcome.txt" of user "user0" is shared with group "sharing-group"
+		And As an "user0"
```
PVince81: I'm not 100% sure this was the purpose of the test

phil-davis: True, the scenario description says:
```
sharing again an own file while belonging to a group
```
The word "again" has me confused, since the sharing is only done once in the tests. Maybe that is just an extra word?
```
sharing an own file while belonging to a group
```
reads better, but then why is there the step:
```
And Deleting last share
```
deleting is not mentioned in the scenario description.

@SergioBertolinSG can you enlighten us?

PVince81: there were some bugs in the past where if you received the same share twice, once with direct user share and once through group share, you got duplicated received folders. I believe that this test case might be one of these, but not 100% sure. Because there are other cases called "Merge shares" that take care of this.

phil-davis:
Oops, I also had this little change in my "general review of integration tests" PR that I made earlier today. When I rebased, of course this change is already in master - so ther are no diffs left here and GitHub has auto-closed this.

I will make a note on the issue and mention relevant people so it can be discussed next week.

Issue #28956, opened 2017-09-08
Integration test trashbin-old-endpoint.feature last scenario fails


phil-davis:
### Steps to reproduce
1. In a dev environment, run the integration test feature:
```
./run.sh features/trashbin-old-endpoint.feature
```

### Expected behaviour
The test passes

### Actual behaviour
The last scenario fails:
```
  Scenario: Deleting an updated file into external storage moves it to the trashbin and can be restored                   # /home/phil/core/tests/integration/features/trashbin-old-endpoint.feature:142
    Given invoking occ with "files:scan --all"                                                                            # CommandLineContext::invokingTheCommand()
    And user "user0" exists                                                                                               # FeatureContext::assureUserExists()
    And user "user0" created a folder "/local_storage/tmp"                                                                # FeatureContext::userCreatedAFolder()
      Client error response [url] http://localhost:8080/remote.php/webdav/local_storage/tmp [status code] 405 [reason phrase] Method Not Allowed (GuzzleHttp\Exception\ClientException)
    And user "user0" moved file "/textfile0.txt" to "/local_storage/tmp/textfile0.txt"                                    # FeatureContext::userMovedFile()
    And user "user0" uploads chunk file "1" of "1" with "AA" to "/local_storage/tmp/textfile0.txt"                        # FeatureContext::userUploadsChunkedFile()
    And user "user0" deletes file "/local_storage/tmp/textfile0.txt"                                                      # FeatureContext::userDeletesFile()
    And as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" exists in trash                       # FeatureContext::elementIsInTrashCheckingOriginalPath()
    And logging in using web as "user0"                                                                                   # FeatureContext::loggingInUsingWebAs()
    When as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" is restored                          # FeatureContext::elementInTrashIsRestored()
    Then as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" does not exist in trash              # FeatureContext::elementIsNotInTrashCheckingOriginalPath()
    And downloaded content when downloading file "/local_storage/tmp/textfile0.txt" with range "bytes=0-1" should be "AA" # FeatureContext::downloadedContentWhenDownloadindShouldBe()

--- Failed scenarios:

    /home/phil/core/tests/integration/features/trashbin-old-endpoint.feature:142

12 scenarios (11 passed, 1 failed)
138 steps (129 passed, 1 failed, 8 skipped)
[...]

PR #28957, opened 2017-09-08
In integration tests do files:scan when the user exists

phil-davis:
## Description
``files:scan`` was being run while ``user0`` did not exist.
If there were files in external storage deleted "underneath" by the AfterScenario of the previous scenario, then ``files:scan`` is needed to make ownCloud "forget" about them. But AfterScenario has also deleted ``user0``. In that case ``files:scan`` does not see the files, because it does not see the user. But when the user is created again, ownCloud's "memory" of the files comes back.

Perhaps there is an issue here about being able to cleanup stale data about files of a user that no longer exists. I will raise that separately if I can easily make a case about it.

But here, the simple thing to do is to create the user first, then do ``files:scan``

## Related Issue
#28956

## Motivation and Context
Make integration tests more reliable in a range of environments.

## How Has This Been Tested?
Run the tests for the feature in a local dev VM.

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/trashbin-old-endpoint.feature
+++ b/tests/integration/features/trashbin-old-endpoint.feature
@@ -113,8 +113,8 @@ Feature: trashbin-new-endpoint
 	@local_storage
 	@no_default_encryption
 	Scenario: Deleting a folder into external storage moves it to the trashbin
-		Given invoking occ with "files:scan --all"
-		And user "user0" exists
+		Given user "user0" exists
+		And invoking occ with "files:scan --all"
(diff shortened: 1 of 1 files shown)


phil-davis:
Read the comments in the issue. The existing code depends on ``files:scan`` being able to process the ``admin`` account (which usually it can). The changed code ensures that there is a valid account that has just been created (which will contain files...) and ``files:scan`` should always process that OK.

Do we make this change? or not?

phil-davis:
Closing because this change is only needed to help in a situation where the admin account files are really broken - which should not be the case when someone starts running tests.

PR #28960, opened 2017-09-08
Gracefully ignore files with wrong permissions during filescan

IljaN:
```
Exception during scan: Undefined index: size
#0 /home/ilja/code/core2/lib/private/Files/Cache/Scanner.php(421): OCA\Files\Command\Scan->exceptionErrorHandler(8, 'Undefined index...', '/home/ilja/code...', 421, Array)
#1 /home/ilja/code/core2/lib/private/Files/Cache/Scanner.php(382): OC\Files\Cache\Scanner->handleChildren('files', true, 3, '3', true, 163)
#2 /home/ilja/code/core2/lib/private/Files/Cache/Scanner.php(385): OC\Files\Cache\Scanner->scanChildren('files', true, 3, '3', true)
#3 /home/ilja/code/core2/lib/private/Files/Cache/Scanner.php(316): OC\Files\Cache\Scanner->scanChildren('', true, 3, '1', true)
#4 /home/ilja/code/core2/lib/private/Files/Utils/Scanner.php(235): OC\Files\Cache\Scanner->scan('', true, 3)
#5 /home/ilja/code/core2/apps/files/lib/Command/Scan.php(159): OC\Files\Utils\Scanner->scan('/admin')
#6 /home/ilja/code/core2/apps/files/lib/Command/Scan.php(228): OCA\Files\Command\Scan->scanFiles('admin', '/admin', true, Object(Symfony\Component\Console\Output\ConsoleOutput), false)
#7 /home/ilja/code/core2/lib/composer/symfony/console/Command/Command.php(262): OCA\Files\Command\Scan->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#8 /home/ilja/code/core2/core/Command/Base.php(159): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#9 /home/ilja/code/core2/lib/composer/symfony/console/Application.php(826): OC\Core\Command\Base->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#10 /home/ilja/code/core2/lib/composer/symfony/console/Application.php(189): Symfony\Component\Console\Application->doRunCommand(Object(OCA\Files\Command\Scan), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#11 /home/ilja/code/core2/lib/composer/symfony/console/Application.php(120): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#12 /home/ilja/code/core2/lib/private/Console/Application.php(160): Symfony\Component\Console\Application->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#13 /home/ilja/code/core2/console.php(106): OC\Console\Application->run()
#14 /home/ilja/code/core2/occ(11): require_once('/home/ilja/code...')
#15 {main}
```
https://github.com/owncloud/enterprise/issues/2235

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Checksum.php
+++ b/lib/private/Files/Storage/Wrapper/Checksum.php
@@ -189,6 +189,11 @@ class Checksum extends Wrapper {
 		if(!self::isPartialFile($path)) {
 			$parentMetaData = $this->getWrapperStorage()->getMetaData($path);
 		}
+
+		if (!$parentMetaData) {
+			return null;
+		}
+
 		$parentMetaData['checksum'] = self::getChecksumsInDbFormat($path);
 
 		if (!isset($parentMetaData['mimetype'])) {


lib/private/Files/Storage/Wrapper/Checksum.php
```diff
@@ -189,6 +189,11 @@ public function getMetaData($path) {
 		if(!self::isPartialFile($path)) {
 			$parentMetaData = $this->getWrapperStorage()->getMetaData($path);
 		}
+
+		if (!$parentMetaData) {
```
PVince81:
```
<?php
        $a = [];
        if (!$a) {
                echo "MEEEP";
        }
```

mrow4a: @PVince81 I am not sure what you want to achieve, but first assigning to array and then checking for false makes no sense. I think what @IljaN wants to achieve is to check the output of $parentMetaData.

mrow4a (commented):
@IljaN can you elaborate a bit more, what is causing `->getMetaData` to return false, instead of what interface requires, array? https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Storage.php#L95

I will go more for fixing it to return empty array instead maybe, so to follow the interface requirements?

lib/private/Files/Storage/Wrapper/Checksum.php
PVince81: The problem is that if we're dealing with part files it doesn't go into that `if` block, so `$parentMetadata` always evaluated to false even as an empty array, so the code would think that we're in the error condition.

I think the intent is likely to check the output of `getMetadata()` so the check must be in the block above instead. Let's see what @IljaN says about the intent.

IljaN: getMetaData is called here by the scanner: https://github.com/owncloud/core/blob/f7874bfd672f6061e68be5333625747bd01801e8/lib/private/Files/Cache/Scanner.php#L112-L122

I think the problem is that the Scanner (and maybe the whole system) expects getMetaData  (which is chained trough all storage wrappers) to return null if a file can not be accessed. 

However the checksum wrapper behaves differently in this regard and returns an empty array. I can`t rembember why it is like this, but some call-site of getMetaData in the checksum or filecache codepath requires an array result.

mrow4a: If it is true with returning NULL, maybe we should change interface and move the if in the place mentioned by @PVince81, in the if block.

IljaN: You are correct according to the interface getMetaData should always return an array. Hmhh... then the code above makes wrong assumptions as $data should never be null.

IljaN: The Problem is here then:
https://github.com/owncloud/core/blob/c763d1957ffef9a6101da010916c5e6a59ccbc1d/lib/private/Files/Storage/Common.php#L618-L626

IljaN: Another return null:
https://github.com/owncloud/core/blob/a45a46e2f3ebcb7c66b3db52801c5e5aaecc6e21/lib/private/Files/Storage/Wrapper/Encryption.php#L172-L182

mrow4a: @DeepDiver1975 @PVince81 -> change interface or implementation?

PVince81:
@IljaN any update ?

patrickjahns:
@IljaN 

Can you check what is needed to have this merge ready?

IljaN:
fixeed by #29739 
See discussion in the linked issue

/cc @mrow4a, @patrickjahns

PR #28931, opened 2017-09-06
Use correct class namespace for ownCloud ext storage

PVince81:
## Description
Fixes uploading to remote ownCloud external storages by properly
disabling part files when detecting the correct class name.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28929

## How Has This Been Tested?
See steps in ticket

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #28978, opened 2017-09-11
[stable9.1] Use correct class namespace for ownCloud ext storage

PVince81:
Backport of https://github.com/owncloud/core/pull/28931 to stable9.1.

Note: need to use a Webdav client to trigger the original issue as the web UI here doesn't use Webdav PUT.


jvillafanez (approved):
Code is fine, not tested

PR #28986, opened 2017-09-12
Fix doc blocks / namespacing - everywhere

tomneedham:
(+ one bonus that was never actually included)

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -30,27 +30,36 @@
  */
 namespace OCA\DAV\Connector\Sabre;
 
+use OC\Files\FileInfo;
+use OC\Files\Filesystem;
 use OCA\DAV\Connector\Sabre\Exception\Forbidden;
 use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
 use OCA\DAV\Connector\Sabre\Exception\FileLocked;
 use OCP\Files\ForbiddenException;
+use OCP\Files\InvalidPathException;
+use OCP\Files\StorageNotAvailableException;
 use OCP\Lock\ILockingProvider;
 use OCP\Lock\LockedException;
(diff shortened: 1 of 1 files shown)


butonic:
Fix all the doc blocks

apps/dav/lib/Connector/Sabre/Directory.php
```diff
@@ -376,6 +387,12 @@ public function getQuotaInfo() {
      * @param string $fullSourcePath Full path to source node
      * @param INode $sourceNode Source node itself
      * @return bool
+	 * @throws BadRequest
+	 * @throws ServiceUnavailable
+	 * @throws \Sabre\DAV\Exception\Forbidden
```
DeepDiver1975: dup

DeepDiver1975 (changes requested):

apps/dav/lib/Connector/Sabre/Directory.php
butonic: Not a duplicate: OCA\DAV\Connector\Sabre\Exception\Forbidden != \Sabre\DAV\Exception\Forbidden

butonic (approved):

apps/dav/lib/Connector/Sabre/Directory.php
DeepDiver1975: then let's add a proper use statement

```php
use \Sabre\DAV\Exception\Forbidden as SabreForbidden

tomneedham:
Fixed and prefixed sabre exceptions with 'Sabre'

tomneedham:
horracle timeout

tomneedham:
rebased for luck

DeepDiver1975 (approved):

tomneedham:
rebased - waiting for CI

tomneedham:
jenkins timeout - rebasing

Issue #28985, opened 2017-09-12
Encryption 9.1.6 > 10.0.2 default module enabled


voroyam:
### Steps to reproduce
1. Install ownCloud 9.1.6 
2. enable default encryption module
3. Upgrade to 10.0.2

### Expected behaviour
User specific key enabled, no drop down menu to choose master key encryption

### Actual behaviour
Drop down menu, for users to choose the different encryption modes.

PVince81:
@sharidas please have a look. If any of the modes was already active when upgrading, the oc_appconfig keys need to be set appropriately to make sure the UI doesn't appear any more.

PVince81:
@davitol can you test that the workaround works: basically the admin should select whatever mode they already selected in the past. We'll add an entry in the release notes.

davitol:
Scenario 1 - Upgrade to 10.0.2 and do not select any option in the dropdown: Encrypted files are still available.
Scenario 2 - Upgrade to 10.0.2 and  select user-key option in the dropdown (same as before): Encrypted files are still available.

Scenario 2 - Upgrade to 10.0.2 and  select master key option in the dropdown: Encrypted files are no longer available. (see screenshot)

![screen shot 2017-09-13 at 17 49 52](https://user-images.githubusercontent.com/10596560/30387326-67e35248-98ac-11e7-9a4d-c771df2666c3.png)

sharidas:
Both the changes for:
- [x] stable 10 branch : https://github.com/owncloud/core/pull/29049
- [x] master branch encryption: https://github.com/owncloud/encryption/pull/15 
are merged.

PVince81:
@sharidas close ?

sharidas:
Closing this issue as the changes are merged.

PR #28988, opened 2017-09-12
[stable9.1] Enable userSpecificKey in appConfig

sharidas:
Enable userSpecificKey in oc_appconfig. This will
help users when the systems are upgraded to version
10.*

## Description

This will help solve the problem of oC upgrade, when encryption is enabled at 9.1 version.

## Related Issue

https://github.com/owncloud/core/issues/28985

## Motivation and Context

If encryption is enabled at oC 9.1.6 and when upgraded to 10.0.2 ( as mentioned in the issue ), then with this change, user can see user specific key enabled ( so no more list to select from drop down ). Or if user enabled masterkey at 9.1.6 ( using command line ), then after upgrade with this change, there is no drop down list in the encryption page. The page will have master key enabled displayed.

## How Has This Been Tested?

This change has been tested by enabling encryption at 9.1.6 and then upgrade to 10.0.2
- [x] User should see no drop down list to select and see either master key or user specific key ( based on what is enabled at 9.1.6 ).

## Types of changes

## Checklist:

### DIFF
--- a/settings/js/admin.js
+++ b/settings/js/admin.js
@@ -60,6 +60,7 @@ $(document).ready(function(){
 		$('#encryptionAPI div#EncryptionWarning').toggleClass('hidden');
 		$('#encryptionAPI div#EncryptionSettingsArea').toggleClass('hidden');
 		OC.AppConfig.setValue('core', 'encryption_enabled', 'yes');
+		OC.AppConfig.setValue('encryption', 'userSpecificKey', '1');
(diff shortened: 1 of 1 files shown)


PVince81 (changes requested):
:-1: this will not work for people who upgrade from 8.2 to 10.0.3.

Besides, setting this value only in the UI would not cover people who already had encryption enabled and do the upgrade without going to the UI again.

Please find a solution involving code changes for stable10 only.

PR #28989, opened 2017-09-13
Rationalize integration test use traits and resetAppConfigs

phil-davis:
## Description
1) Put the traits in a flat "tree" under BasicStructure.
2) ``use BasicStructure`` from each "top-level" suite context.
3) require ``bootstrap.php`` for "top-level" suite contexts that need TestHelpers (that make it autoload the TestHelpers if the suite is run by itself)
4) make ``resetCommonSharingAppConfigs()`` that can set the sharing settings that every suite wants as a basic starting point.
5) adjust ``resetAppConfigs()`` in each "top-level" suite context so that it calls ``resetCommonSharingAppConfigs()`` and then also sets up any other sharing etc configs that it cares about.

## Motivation and Context
1) The "use" traits in the integration tests are used in a loop. (I will make a diagram of the current structure...)
2) Some integration tests suites give failures when run standalone, because their individual ``resetAppConfigs()`` does not really make all the settings they need.
3) The branching of the "use" traits is confusing.

## How Has This Been Tested?
Run test runs on Jenkins with each of the suites individually (cutting down behat.yml) to demonstrate that each suite will run on its own.

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/BasicStructure.php
+++ b/tests/integration/features/bootstrap/BasicStructure.php
@@ -7,12 +7,16 @@ require __DIR__ . '/../../../../lib/composer/autoload.php';
 
 trait BasicStructure {
 
+	use AppConfiguration;
 	use Auth;
-	use Trashbin;
 	use Checksums;
 	use Comments;
 	use MailTool;
+	use Provisioning;
+	use Sharing;
 	use Tags;
+	use Trashbin;
+	use WebDav;
 
 	/** @var string */
 	private $currentUser = '';
--- a/tests/integration/features/bootstrap/CapabilitiesContext.php
+++ b/tests/integration/features/bootstrap/CapabilitiesContext.php
@@ -11,7 +11,6 @@ require __DIR__ . '/../../../../lib/composer/autoload.php';
 class CapabilitiesContext implements Context, SnippetAcceptingContext {
 
 	use BasicStructure;
-	use AppConfiguration;
 
 	/**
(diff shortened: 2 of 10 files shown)


individual-it (approved):
LGTM
🎉 for getting rid of the trait loop

phil-davis:
Here is the map of the way the traits are used by each other and the main *Context code:
owncloudintegrationtestsusetraits

SergioBertolinSG:
@phil-davis yes, just grew. What is the new structure after this PR?

phil-davis:
Just a 1-level tree - BasicStructure does a ``use`` of everything else. And each of the 4 top-level Context files does ``use BasicStructure``

Previously they were all getting mostly everything included anyway, because they started at some point in the loop and the ``use`` inside ``use`` went right through the loop anyway.

It would be possible to just put a subset of ``use`` statements in each top-level Context file, for just the methods/step definitions that are really needed by that Context. But I am not sure that there is any real benefit in taking the time to find the minimal set for each top-level Context.

phil-davis:
@SergioBertolinSG are you happy with this? Or any other comment?

SergioBertolinSG:
PLease run this tests using the parameters which are not used in jenkins. (OC_TEST_ALT_HOME, OC_TEST_ENCRYPTION_ENABLED at least).

phil-davis:
Those tests all work.

phil-davis:
Backport stable10 #29003

PR #28990, opened 2017-09-13
check trashbin after deletion in UI tests

individual-it:
## Description
when deleting files in the test the trashbin should be checked to see if it works correctly

## Related Issue
https://github.com/owncloud/QA/issues/437

## Motivation and Context
more tests leads to less bugs
The [Simplified Smoke Test](https://github.com/owncloud/QA/blob/master/Server/Test_Plan_Simplified_Smoke_Test.md) requires to check the trashbin

## How Has This Been Tested?
run tests locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #28995, opened 2017-09-13
[stable10] check trashbin after deletion in UI tests

individual-it:
backport to stable10 of #28990


phil-davis (approved):

phil-davis:
This is expansion of automated smoke-testing, so that is why it qualifies for backport now to end up in 10.0.3

PR #29002, opened 2017-09-14
Remove encryption from unignored gitignore list

phil-davis:
## Description
``encryption`` is a separate app nowadays. It is not still in the ``core`` repo. So make sure it is ignored in the ``.gitignore`` list. (i.e. remove it from being "unignored").

While I am at it, sort the list into strict alphabetical order so it is easily verified to be the same as the list of folders at https://github.com/owncloud/core/tree/master/apps

## Motivation and Context
When I clone the ``encryption`` app into the ``core/apps`` folder of a dev environment, ``git`` thinks they are added files for the next commit.

## How Has This Been Tested?
See that ``git`` does ignore the ``encryption`` app files after this change.

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,9 +14,8 @@
 !/apps/comments
 !/apps/dav
 !/apps/files
-!/apps/federation
 !/apps/federatedfilesharing
-!/apps/encryption
+!/apps/federation
 !/apps/files_external
 !/apps/files_sharing
 !/apps/files_trashbin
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Note: as @PVince81 has pointed out, ``encryption`` is still bundled in ``stable10`` - so this fix is only needed in master for future releases 10.1... no need for backporting.

PR #29007, opened 2017-09-14
Remove encryption from unignored gitignore list

phil-davis:
Backport #29002

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,9 +14,8 @@
 !/apps/comments
 !/apps/dav
 !/apps/files
-!/apps/federation
 !/apps/federatedfilesharing
-!/apps/encryption
+!/apps/federation
 !/apps/files_external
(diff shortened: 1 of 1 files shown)


phil-davis:
@PVince81 your decision about when you want to merge this backport.
It does stop annoying dev git behavior when switching around branches when also having ``apps/encryption``

PVince81:
@phil-davis encryption is still bundled with stable10

phil-davis:
Indeed it is - better not backport then!

PVince81:
adding to gitignore likely means that any changes we make in the apps/encryption folder will not be added by `git add` and would require `git add -f`, that's not good.

phil-davis:
Yes, I agree. When branched off core ``stable10`` for some work on an "old" encryption issue, we want edits to encryption files to get added-committed to core.

Issue #7496, opened 2014-03-03
don't allow upload/creation of .part files


schiessle:
we use the file extension .part for uploading large files. Therefore we ignore .part files and don't expose them in the web ui or to a webdav client.

Our sync client, correctly doesn't upload .part files. Also direct webdav uploads refuse to upload .part files. But it is possible to:
- upload .part files over the web interface
- create a new empty .part file
- rename a regular file to a .part file

This will result in files on the server which are no longer accessible by the user. All this operation should be blocked.

@nickvergessen maybe you want to have a look at it if you have some time? Thanks!

cc @PVince81

PVince81:
This won't be fully fixable as long as OC 7 is supported due to federation, see https://github.com/owncloud/core/pull/7514#issuecomment-118319191 for an explanation.

Moving to 9.0 to keep it in sight.

PVince81:
Since OC 7 will still be around when OC 9 is out, moving to 9.1.

See https://github.com/owncloud/core/pull/7514#issuecomment-118319191 for the explanation why.

PVince81:
CC @cmonteroluque

cmonteroluque:
thanks @PVince81

felixheidecke:
See PR https://github.com/owncloud/core/pull/27149

PVince81:
PR for frontend was merged: https://github.com/owncloud/core/pull/27149

Now requires backend changes @IljaN

IljaN:
@PVince81 #27269

PVince81:
This fix is incomplete, part file upload is still possible with Webdav on v10.0.2.

On v10.0.3 another bug prevents it, but we need this fix this correctly.

Reopening.

PVince81:
@IljaN please have a look

PVince81:
also please add integration tests to make sure we can't:
- upload part files through Webdav
- rename or copy any file to ".part" with Webdav

PR #27149, opened 2017-02-13
don't allow upload/creation of .part files

lordelix:
## Description
don't allow upload/creation of .part files

## Related Issue

https://github.com/owncloud/core/issues/7496

## Motivation and Context
.part is a reserved file-extension in the context of OC

## How Has This Been Tested?

Wrote test

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #19235, opened 2015-09-22
Exclude Directories

mmattel:
Successor of PR https://github.com/owncloud/core/pull/16534

Before  mmattel/core
Now     owncloud/core

Shoud fix CI start

PR #29011, opened 2017-09-14
Prevent .part upload via dav

IljaN:
## Description
Prevent creating, moving, copying to paths with extension .part
## Related Issue
#7496 

## How Has This Been Tested?
Manually, Integration-Tests

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -1799,6 +1799,13 @@ class View {
 			throw new InvalidPathException($l10n->t('Dot files are not allowed'));
 		}
 
+
+		if (preg_match('/' . FileInfo::BLACKLIST_FILES_REGEX . '/', $fileName) !== 0) {
+			throw new InvalidPathException(
+				'Can`t upload files with extension .part because this extension is used internally by owncloud.'
+			);
+		}
+
 		if (!\OC::$server->getDatabaseConnection()->allows4ByteCharacters()) {
 			// verify database - e.g. mysql only 3-byte chars
 			if (preg_match('%(?:
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -1799,10 +1799,11 @@ class View {
 			throw new InvalidPathException($l10n->t('Dot files are not allowed'));
 		}
 
+		$matches = [];
 
 		if (preg_match('/' . FileInfo::BLACKLIST_FILES_REGEX . '/', $fileName) !== 0) {
 			throw new InvalidPathException(
-				'Can`t upload files with extension .part because this extension is used internally by owncloud.'
+				"Can`t upload files with extension {$matches[0]} because these extensions are reserved for internal use."
 			);
 		}
 
--- a/tests/integration/features/bootstrap/WebDav.php
+++ b/tests/integration/features/bootstrap/WebDav.php
@@ -835,6 +835,8 @@ trait WebDav {
 		} catch (\GuzzleHttp\Exception\ServerException $e) {
 			// 4xx and 5xx responses cause an exception
 			$this->response = $e->getResponse();
+		} catch (\GuzzleHttp\Exception\ClientException $ex) {
+			$this->response = $ex->getResponse();
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


mmattel:
@IljaN 
one question, what would happen if somone creates a folder called ``.part`` ?

IljaN:
@mmattel Good catch! It is not possible via webdav: mkcol .part Fails with 405 Method not allowed, not sure if it is the correct code

If you create the directory directly in your storage with mkdir .part it is ignored by owncloud since we don`t support dotfiles afaik

PVince81:
@iljan we should also block "$name.part" folders. Ignoring is not good. If you had the impression that OC ignores it, did you check the FS ? Maybe it created the folder still but it wasn't visible in the UI ?

Better block those as well.

mmattel:
@IljaN maybe you could use from https://github.com/owncloud/core/pull/19235 
``lib/private/Files/Filesystem.php`` function ``isForbiddenFileOrDir`` and add ``.part`` into the query array

IljaN:
@PVince81 .part directory creation is already prevented by this change:
```
dav:/remote.php/dav/files/admin/> mkcol path.with.ext.part
Creating `path.with.ext.part': failed:
400 Bad Request
```
@mmattel I extended verfiyPath method with a extension check, this seems to catch all cases with the benefit that it is used in all codepaths (Single point of truth), what benefits does extending isForbiddenFileOr dir give us?

PVince81:
@IljaN prove it with an integration test 😉 

Is 400 bad request the correct code, same as for files ? Just asking, I thought we returned 403 for those but don't remember exactly.

mmattel:
@IljaN nothing special, just remembered that we excluded stuff some time ago and thought it is maybe usable for your PR, never mind

PVince81:
rebase conflicts

IljaN:
@PVince81 let me fix them, I am also in the process adding a missing test

IljaN:
@PVince81 Ready

lib/private/Files/View.php
```diff
@@ -1799,6 +1799,13 @@ public function verifyPath($path, $fileName) {
 			throw new InvalidPathException($l10n->t('Dot files are not allowed'));
 		}
 
+
+		if (pathinfo($fileName, PATHINFO_EXTENSION) == 'part') {
```
PVince81: did we talk about https://github.com/owncloud/core/pull/27149/files#diff-6cc3350c00dd13da023086587cce8bfdR64 ?

IljaN:
Ready for next review round

lib/private/Files/View.php
```diff
@@ -1799,6 +1799,13 @@ public function verifyPath($path, $fileName) {
 			throw new InvalidPathException($l10n->t('Dot files are not allowed'));
 		}
 
+
+		if (preg_match('/' . FileInfo::BLACKLIST_FILES_REGEX . '/', $fileName) !== 0) {
+			throw new InvalidPathException(
+				'Can`t upload files with extension .part because this extension is used internally by owncloud.'
```
mmattel: wouldnt it be good to get the filename/component in question from the constant instead of hardcoding it?

PVince81: @IljaN see above comment

Also we cannot use the name "ownCloud" in error messages due to theming, please remove it

IljaN: Not sure if I understand I am getting it from FileInfo::BLACKLIST_FILES_REGEX

mmattel: To be precise, it is not the query but the text.

Here you state hardcoded ``.part`` 
But the regex says ``const BLACKLIST_FILES_REGEX = '\.(part|filepart)$';``
One name is missing.
What happens when the regex is extended with more names?
Extract the names and print them to be futureproof.

IljaN:
Ready for next round of review

PVince81 (approved):
:+1: looks good to me, thanks a lot!

PVince81:
@IljaN please backport to stable10

PVince81:
stable10: https://github.com/owncloud/core/pull/29432

Issue #28751, opened 2017-08-21
OCP\Settings\Section should allow non-hardcoded / core icons


tomneedham:
Currently `OC\Settings\Section` takes in a string which is used to find hardcoded / core icons. We should extend this to allow apps to supply their own icons here - and core provides the necessary styling.

pmaier1:
Thanks for opening the issue. Would this also be something for you to assign? Moving to 'planned' for 10.0.4.

tomneedham:
@PVince81 can assign as necessary - I need to focus with support :)

PVince81:
@sharidas can you have a look ?

sharidas:
sure

PR #29013, opened 2017-09-14
Allow non hardcoded icons for apps

sharidas:
This change helps users to provide their own
custom icons for the apps which work with
ownCloud.

## Description

This change helps contributors to use non-hardcoded icons. That is every app developer or contributor can now have their own separate icons which can be used with core.

## Related Issue

https://github.com/owncloud/core/issues/28751

## Motivation and Context

At present the icons are hard coded and are located at core. This prohibits the usage of custom icons for apps. With this change, contributors of apps, can have their own custom app which can be used along with core. No more saving of svg files in the core.

## How Has This Been Tested?

I have verified it against 2 apps. One is encryption app  ( the change was made in the SettingsManager.php) and another app was firewall app. For the firewall app I modified the AdminSection.php to test.

settingspagenew

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Settings/SettingsManager.php
+++ b/lib/private/Settings/SettingsManager.php
@@ -205,7 +205,6 @@ class SettingsManager implements ISettingsManager {
 				new Section('workflow', $this->l->t('Workflows & Tags'), 85, 'workflows'),
 				new Section('sharing', $this->l->t('Sharing'), 80, 'share'),
 				new Section('search', $this->l->t('Search'), 75, 'search'),
-				new Section('updates', $this->l->t('Updates'), 20, 'update'),
 				new Section('help', $this->l->t('Help & Tips'), -5, 'info'),
 				new Section('additional', $this->l->t('Additional'), -10, 'more'),
 			];
--- a/settings/Controller/SettingsPageController.php
+++ b/settings/Controller/SettingsPageController.php
@@ -107,6 +107,29 @@ class SettingsPageController extends Controller {
 	}
 
 	/**
+	 * Gets the icon for the settings panel. The idea is
+	 * to get either URL or the icon name ( for backward
+	 * compatibility ). The icons returned will be svg
+	 * format and no other formats are supported.
+	 *
+	 * @param \OCP\Settings\ISection $section
+	 * @return string
+	 */
+
+	protected function getIconForSettingsPanel($section) {
+		$icon = $section->getIconName() . '.svg';
+		$appPath = \OC_App::getAppPath($section->getID());
+
+		if (file_exists($appPath . '/img/' . $icon)) {
+			$icon = $this->urlGenerator->imagePath($section->getID(), $icon);
+		} else {
+			$icon = $section->getIconName();
+		}
+
+		return $icon;
+	}
+
+	/**
 	 * Gets an array used to generate the navigation in the UI
 	 * @param array $sections array of ISections
 	 * @param string $currentSectionID
@@ -117,6 +140,9 @@ class SettingsPageController extends Controller {
 		$nav = [];
 		// Iterate through sections and get id, name and see if currently active
 		foreach($sections as $section) {
+
+			$icon = $this->getIconForSettingsPanel($section);
+
 			$nav[] = [
 				'id' => $section->getID(),
 				'link' => $this->urlGenerator->linkToRoute(
@@ -125,7 +151,7 @@ class SettingsPageController extends Controller {
 				),
 				'name' => ucfirst($section->getName()),
 				'active' => $section->getID() === $currentSectionID,
-				'icon' => $section->getIconName()
+				'icon' => $icon
 			];
 		}
 		return $nav;
--- a/settings/templates/settingsPage.php
+++ b/settings/templates/settingsPage.php
@@ -21,37 +21,28 @@ style('settings', 'settings');
 <div id="app-navigation">
 	<ul class="with-icon">
 		<li class="divider"><?php p($l->t('Personal')); ?></li>
-		<?php foreach($_['personalNav'] as $item) {
-			$active = $item['active'] ? ' active ' : '';
-			print_unescaped(
-				sprintf(
-					"<li><a class=\"svg %s %s\" href='%s'>%s</a></li>",
-					$active,
-					'icon-'.\OCP\Util::sanitizeHTML($item['icon']),
-					\OCP\Util::sanitizeHTML($item['link']),
-					\OCP\Util::sanitizeHTML($item['name'])
-				)
-			);
-		}
-		if(!empty($_['adminNav'])) { ?>
+		<?php foreach($_['personalNav'] as $item): ?>
+		<li>
+			<?php if (strpos($item['icon'], '/', 1) !== false): ?>
+				<a class="svg <?php $item['active'] ? p(' active ') : p('') ?>" style="background-image: url(<?php p($item['icon']) ?>)" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
+			<?php else: ?>
+				<a class="svg <?php $item['active'] ? p(' active ') : p('') ?> icon-<?php p($item['icon']) ?>" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
+			<?php endif; ?>
+		</li>
+		<?php endforeach; ?>
 
+		<?php if (!empty($_['adminNav'])): ?>
 			<li class="divider"><?php p($l->t('Admin')); ?></li>
-			<?php
-
-			foreach ($_['adminNav'] as $item) {
-				$active = $item['active'] ? ' active ' : '';
-				print_unescaped(
-					sprintf(
-						"<li><a class=\"svg %s %s\" href='%s'>%s</a></li>",
-						$active,
-						'icon-'.\OCP\Util::sanitizeHTML($item['icon']),
-						\OCP\Util::sanitizeHTML($item['link']),
-						\OCP\Util::sanitizeHTML($item['name'])
-					)
-				);
-			}
-		}
-		?>
+			<?php foreach($_['adminNav'] as $item): ?>
+				<li>
+					<?php if (strpos($item['icon'], '/', 1) !== false): ?>
+						<a class="svg <?php $item['active'] ? p(' active ') : p('') ?>" style="background-image: url(<?php p($item['icon']) ?>)" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
+					<?php else: ?>
+						<a class="svg <?php $item['active'] ? p(' active ') : p('') ?> icon-<?php p($item['icon']) ?>" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
+					<?php endif; ?>
+				</li>
+			<?php endforeach; ?>
+		<?php endif; ?>
 	</ul>
 </div>
 <div id="app-content">


lib/public/Settings/ISection.php
```diff
@@ -53,4 +53,6 @@ public function getPriority();
 	 */
 	public function getIconName();
```
tomneedham: I think let's kill the name and just keep the url and we'll make it backwards compatible

settings/Controller/SettingsPageController.php
```diff
@@ -125,7 +125,8 @@ protected function getNavigation($sections, $currentSectionID, $type) {
 				),
 				'name' => ucfirst($section->getName()),
 				'active' => $section->getID() === $currentSectionID,
-				'icon' => $section->getIconName()
+				'icon' => $section->getIconName(),
```
tomneedham: so we can get rid of this also

lib/private/legacy/app.php
```diff
@@ -597,6 +597,22 @@ public static function getAppPath($appId) {
...
+		$fileAbsolutePath = self::getAppPath($appId)  . '/img/' . end(explode('/', $url));
+		$view = new \OC\Files\View();
+		if (!$view->file_exists($fileAbsolutePath)) {
+			$view->mkdir($iconPath);
+			file_put_contents($fileAbsolutePath, fopen($url, 'r'));
+		}
+		$url = strstr($fileAbsolutePath, \OC::$WEBROOT);
+		return $url;
+	}
+
```
tomneedham: nooooo more static ;) Why do we need to do this copying stuff?

lib/private/Settings/Section.php
```diff
@@ -59,4 +61,8 @@ public function getIconName() {
 		return $this->icon;
 	}
 
+	public function getIconUrl() {
+		return \OC_App::getIconFromUrl($this->url, $this->getID());
+	}
```
tomneedham: I think we could use the urlgenerator here, injected in properly

tomneedham (changes requested):

lib/private/legacy/app.php
sharidas: The intention behind this change was to fetch the data from the web url and download to apps/img folder. Anything wrong?

sharidas:
Attaching new screenshot for the change made: 
newscreenshot

lib/private/Settings/Section.php
```diff
@@ -34,13 +34,13 @@ class Section implements ISection {
 	protected $name;
 	/** @var int */
 	protected $priority;
-	protected $icon;
+	protected $url;
 
-	public function __construct($id, $name, $priority, $icon='settings') {
+	public function __construct($id, $name, $priority, $url='settings') {
```
tomneedham: we can pass in the urlgenerator here - then adjust the constructor to pass it in in `\OC\Settings\SettingsManager`

lib/private/Settings/Section.php
```diff
@@ -55,8 +55,12 @@ public function getPriority() {
 		return $this->priority;
 	}
 
-	public function getIconName() {
-		return $this->icon;
+	public function getIconUrl() {
+		try {
+			return \OC::$server->getURLGenerator()->imagePath($this->getID(), $this->url);
```
tomneedham: use passed in urlgenerator. Also, the Id is not necessarily the app id so might need some thought here

settings/templates/settingsPage.php
```diff
@@ -40,11 +46,17 @@
 
 			foreach ($_['adminNav'] as $item) {
 				$active = $item['active'] ? ' active ' : '';
+				if (strpos($item['url'], '/', 1) !== false) {
```
tomneedham: we may as well standardise the core icons. So always pass a url. In the controller figure out if its a core icon or not. Then we can remove these crappy css rules for the icons

tomneedham (changes requested):

lib/private/Settings/SettingsManager.php
```diff
@@ -197,25 +197,25 @@ private function getBuiltInSections($type) {
...
-				new Section('search', $this->l->t('Search'), 75, 'search'),
+				new Section('general', $this->l->t('General'), 100, $this->urlGenerator->imagePath('core', 'actions/settings.svg')),
+				new Section('storage', $this->l->t('Storage'), 95,  $this->urlGenerator->imagePath('core', 'filetypes/folder.svg')),
+				new Section('security', $this->l->t('Security'), 90, $this->urlGenerator->imagePath('core', 'apps/shield.svg')),
+				new Section('authentication', $this->l->t('User Authentication'), 87, $this->urlGenerator->imagePath('core', 'actions/user.svg')),
+				new Section('encryption', $this->l->t('Encryption'), 85, $this->urlGenerator->imagePath('encryption', 'actions/password.svg')),
+				new Section('workflow', $this->l->t('Workflows & Tags'), 85, $this->urlGenerator->imagePath('core', 'apps/workflow.svg')),
+				new Section('sharing', $this->l->t('Sharing'), 80, $this->urlGenerator->imagePath('core', 'actions/share.svg')),
+				new Section('search', $this->l->t('Search'), 75, $this->urlGenerator->imagePath('core', 'actions/search.svg')),
 				new Section('updates', $this->l->t('Updates'), 20, 'update'),
```
tomneedham: missing this one?

lib/public/Settings/ISection.php
```diff
@@ -51,6 +51,6 @@ public function getPriority();
 	 * @since 10.0
```
tomneedham: 10.0.4

tomneedham: and we should keep the old method, but add `@depreceated` from 10.0.4. But this shoudl also be backwards compatible. So inside the section implementation, set `$this->url` using the url generator - don't need to worry about injection for this backwards compat

lib/public/Settings/ISection.php
```diff
@@ -48,9 +48,15 @@ public function getName();
 	public function getPriority();
 
 	/**
-	 * @since 10.0
-	 * @return string
+	 * @deprecated since 10.0.4
```
sharidas: Added `@deprecated`

tomneedham:
+ we should update the docs here https://doc.owncloud.org/server/latest/developer_manual/core/theming.html?highlight=theme#create-custom-sections

lib/public/Settings/ISection.php
sharidas: So if we add `@deprecated` for getIconName in ISection then, I think we will end up having both getIconName and getIconUrl. Because the apps which will be implementing ISection would need to implement both the methods.
So ideally if we don't want to deprecate it, then may be we can have getIconName. And make it app developers responsibility to get the url to be generated and returned from getIconName(), instead of just returning the icon name which it does currently.

lib/public/Settings/ISection.php
```diff
@@ -48,9 +48,9 @@ public function getName();
 	public function getPriority();
 
 	/**
-	 * @since 10.0
+	 * @since 10.0.4
 	 * @return string
 	 */
-	public function getIconName();
+	public function getIconUrl();
```
DeepDiver1975: I don't like this - this will break all existing implementations 👎

DeepDiver1975: We should implement this more in sense where the returned icon name can either be a css class (as done until today) or an image name of the app which hosts the svg.

this way we don't need to change apis and implementation all over the place.

```php
$icon = $section->getIconName();
if (file_exists("$appPath/img/$icon.svg")) {
    $url = $urlGenerator->imagePath("$icon.svg");
    // TODO: add background styling hack
} else {
    // TODO: add icon css class
}

**Review by DeepDiver1975 — CHANGES_REQUESTED** [review]

File: settings/Controller/SettingsPageController.php [hunk 12]
```diff
@@ -117,6 +117,16 @@ protected function getNavigation($sections, $currentSectionID, $type) {
 		$nav = [];
 		// Iterate through sections and get id, name and see if currently active
 		foreach($sections as $section) {
+
+			$icon = $section->getIconName() . '.svg';
```
jvillafanez: Move this "getIcon" code to another function and call it here. This logic should be probably out of this class so we can unittest it properly. If there is no place for this, it should be fine to create a private function here in the short term.

In addition, I think there is already a function to retrieve icons for apps (I might be wrong) Any reason to not use it?

In the phpdoc of the new function (whenever is placed) include how the function should behave. Right now I'm not sure what is the expected behaviour if several icons are provided: if I want to use "marvellous_icon.png", why I'm checking "marvellous_icon.png.svg"?
If only .svg files are supported, make sure this is known, including in the comments of the function.
Since I have no idea what is the expected behaviour here, I'm more inclined to think the current behaviour is buggy.

jvillafanez: A couple of extra things:
* Unless the icon name is already sanitized, make sure it doesn't contain `/../` or something similar
* I'm not sure what is the "default" expectation: I guess it should be using the filename without extension as icon name (which is fine). In any case, optimize for this default case. If the default is to include the extension you'd be checking for a missing file constantly. Just something to take care of.

settings/templates/settingsPage.php
```diff
@@ -23,11 +23,18 @@
 		<li class="divider"><?php p($l->t('Personal')); ?></li>
 		<?php foreach($_['personalNav'] as $item) {
 			$active = $item['active'] ? ' active ' : '';
+
+			$ocUrlorIcon = \OCP\Util::sanitizeHTML($item['icon']);
```
jvillafanez: This could be a good time to rewrite this piece to just use this as a template and avoid adding logic here.

sharidas: Agreed. Can we do this as a separate PR?

jvillafanez: I don't really mind, but there will be conflicts with this PR if t isn't merged soon because that code will depend on this branch... @tomneedham @DeepDiver1975 opinions?

DeepDiver1975: do it in here - own commit for now to see the diff ... later we squash

sharidas: Moved the logic from here to https://github.com/owncloud/core/pull/29013/files#diff-3393e43384d070ab12865348f9279030R123. I hope this makes sense now.

jvillafanez:
The rewrite I had in mind was more like:
```
<li class="divider"><?php p($l->t('Personal')); ?></li>
<?php foreach($_['personalNav'] as $item): ?>
<li>
  <?php if ($item['active']): ?>
    <?php if ($item['iconIsUrl]): ?>
      <a class="svg active" style="background-image: url(<?php p($item['icon']) ?>)" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
    <?php else: ?>
      <a class="svg active icon-<?php p($item['icon']) ?>)" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
    <?php endif; ?>
  <?php else: ?>
    <?php if ($item['iconIsUrl]): ?>
      <a class="svg" style="background-image: url(<?php p($item['icon']) ?>)" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
    <?php else: ?>
      <a class="svg icon-<?php p($item['icon']) ?>)" href='<?php p($item['link']); ?>'><?php p($item['name']) ?></a>
  <?php endif; ?>
</li>
<?php endforeach; ?>
```

Maybe we can make it smaller injecting "active" or an empty string instead of checking if the `$item['active']` is a boolean or not, but as you see, the template just output information with the minimum control structures.

sharidas:
@jvillafanez  I have updated the change let me know if this ok.

settings/templates/settingsPage.php
```diff
@@ -21,37 +21,28 @@
...
-					\OCP\Util::sanitizeHTML($item['link']),
-					\OCP\Util::sanitizeHTML($item['name'])
-				)
-			);
-		}
-		if(!empty($_['adminNav'])) { ?>
+		<?php foreach($_['personalNav'] as $item): ?>
+		<li>
+			<?php if (strpos($item['icon'], '/', 1) !== false): ?>
+				<a class="svg <?php $item['active'] ? p(' active ') : p('') ?>" style="background-image: url(<?php p($item['icon']) ?>)" href='<?php p(\OCP\Util::sanitizeHTML($item['link'])); ?>'><?php p(\OCP\Util::sanitizeHTML($item['name'])) ?></a>
```
jvillafanez: `\OCP\Util::sanitizeHTML(...)` isn't needed with the `p` function

settings/templates/settingsPage.php
```diff
@@ -21,37 +21,28 @@
...
-						$active,
-						'icon-'.\OCP\Util::sanitizeHTML($item['icon']),
-						\OCP\Util::sanitizeHTML($item['link']),
-						\OCP\Util::sanitizeHTML($item['name'])
-					)
-				);
-			}
-		}
-		?>
+		    <?php foreach($_['adminNav'] as $item): ?>
```
jvillafanez: indentation here seems misplaced

sharidas:
@jvillafanez Corrected indentation and removed call to `\OCP\Util::sanitizeHTML()` inside `p()`

settings/templates/settingsPage.php
sharidas: Fixed.

settings/templates/settingsPage.php
sharidas: Removed the call to `sanitizeHTML`

jvillafanez (approved):
Code looks good to me, not tested.

sharidas:
Tested the code by replacing icon `password` with `docker`. I copied docker.svg from https://worldvectorlogo.com/download/docker.svg and the settings page looks like this:
changeencryptionicon
And when the change is reverted back to `password` at https://github.com/owncloud/core/blob/master/lib/private/Settings/SettingsManager.php#L204 the settings page encryption icon looks like this:
changeencryptionicon_old

lib/private/Settings/SettingsManager.php
```diff
@@ -205,7 +205,6 @@ private function getBuiltInSections($type) {
 				new Section('workflow', $this->l->t('Workflows & Tags'), 85, 'workflows'),
 				new Section('sharing', $this->l->t('Sharing'), 80, 'share'),
 				new Section('search', $this->l->t('Search'), 75, 'search'),
-				new Section('updates', $this->l->t('Updates'), 20, 'update'),
```
PVince81: why remove this ?

sharidas: While searching for update.svg, I couldn't find at core. That made me remove this line. I can revert back the change, if my finding was wrong.

PVince81: Thanks for the explanation, I just wanted to double check that this wasn't a mistake. Let's keep it.

PVince81:
@sharidas please backport

sharidas:
Created backport PR: https://github.com/owncloud/core/pull/29358

PR #29023, opened 2017-09-15
Sync version 10 03

phil-davis:
Forward port version.php changes from stable10 to master.


PVince81 (approved):
:+1:

PVince81:
3 commits ? I think one would have been enough

PVince81:
anyway, merge once done

phil-davis:
I thought to just cherry-pick all the different history, in case there was some odd diff. But in the end it was just the same 2 lines changed in each commit.

PR #28507, opened 2017-07-27
Add email validation

lordelix:
- Add email validation for public link creation
- show email field even if the model !isNew

## Related Issue

Fixes https://github.com/owncloud/core/issues/28451

## How was this tested
- Manually tested (visual test) on latest Firefox, Chrome on MacOS.
- Manually tested (visual test) on latest Firefox, Chrome, IE1 and Edge on Windows.
- Changed and added to Jasmine test.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29042, opened 2017-09-18
[stable10] validate public link mail

lordelix:
Backporting Email validation for public link shares.
Original PR: https://github.com/owncloud/core/pull/28507


PVince81:
I do see a second popup appear then directly disappear and an error message appears.

Not as bad as the original bug where the extra dialogs stay.
Would be good to find a way to avoid this second popup to appear.

PVince81:
raised https://github.com/owncloud/core/issues/29043 for the extra dialog that appears

PR #29045, opened 2017-09-18
[stable9] Filecache repair step

tomneedham:
Backport of https://github.com/owncloud/core/pull/28253 for stable9 (ignore the branch name, rebased for stable9)


tomneedham:
Do we have console output access for repair steps? I just mapped it over to a OCP\ILogger instance for now

tomneedham:
Needs a couple of fixes from https://github.com/owncloud/core/pull/28253 then this should work here too 🎉

tomneedham:
Ok let's see how this goes - I think the tests might need some more tweaking with regards to output / repairstep differences

tomneedham:
Ahh yes we dont have `createMock()`

tomneedham:
detached subtree repair step fails on oracle...

tomneedham:
Was missing the last fix from vincent https://github.com/owncloud/core/pull/28253/commits/4c157211685c450b46437b1188ed849676670617

tomneedham:
Let's wait for CI here

tomneedham:
Jenkins passed 💪

PVince81:
@tomneedham is this up to date with the latest changes from the stable10 PR ?

before merging stable9 would be good to also have stable9.1 backport and merge that one first

tomneedham:
Rebased and added path hash commit for performance

apps/files/command/scan.php
```diff
@@ -198,9 +284,10 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			if ($this->userManager->userExists($user)) {
 				# add an extra line when verbose is set to optical separate users
 				if ($verbose) {$output->writeln(""); }
-				$output->writeln("Starting scan for user $user_count out of $users_total ($user)");
+				$r = $shouldRepairStoragesIndividually ? ' (and repair)' : '';
```
PVince81: I don't remember these changes, did you add them to master/stable10 as well ?

lib/private/repair/repairmismatchfilecachepath.php
```diff
@@ -0,0 +1,580 @@
...
+		$this->connection->beginTransaction();
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->update('filecache')
+			->set('parent', $qb->createNamedParameter($correctParentId))
+			->where($qb->expr()->eq('fileid', $qb->createNamedParameter($fileId)));
+		$qb->execute();
+
+		$text = "Fixed file cache entry with fileid $fileId, set wrong parent \"$wrongParentId\" to \"$correctParentId\"";
+		$out->advance(1, $text);
```
PVince81: I didn't know we had the ability to display progress in stable9, good

lib/private/repair/repairmismatchfilecachepath.php
```diff
@@ -0,0 +1,580 @@
...
+				// just fix all
+				$this->fixEntriesWithNonExistingParentIdEntry($out);
+			}
+			$out->finishProgress();
+			$out->info('');
+		}
+	}
+}
+
+class LogOutput {
```
PVince81: Bäh... in my original backport I just removed progress logging.

PVince81 (commented):
Changes look ok, but I feel slightly disturbed by the hard-coded LogOutput class. Maybe just remove all logging ?

The other alternative is to use the event emitter like we do in other repair steps. But that might not work when trying to log to the console when run from the scan command.

PVince81:
Weird JS setup failures, same like https://github.com/owncloud/core/pull/29407#issuecomment-342280762

PVince81:
Will require https://github.com/owncloud/core/pull/29477 for JS tests to pass and maybe a PR resubmit to clean the env.

PVince81:
Resend as https://github.com/owncloud/core/pull/29489 to fix JS env issue

apps/files/command/scan.php
tomneedham: master https://github.com/owncloud/core/blob/master/apps/files/lib/Command/Scan.php#L305
stable10 https://github.com/owncloud/core/blob/stable10/apps/files/lib/Command/Scan.php#L305

PR #29049, opened 2017-09-18
[stable10] Fix migration issue caused during upgrade to version 10

sharidas:
This change will help users to fix the migration
issue caused during upgrade to version 10. If
user had opted user specific key in the older versions
then no drop down will appear in the version 10.
Same applies if user had opted for the masterkey
in older version, no drop down will appear in version
10.

## Description

When user tries to upgrade from versions < 10, the userSpecificKey is not present in the appConfig. And as a result during migration to version 10.0.2, the user gets an option to select the encryption mode :
- [x] User Specific Key
- [x] Master Key
This issue is addressed with this change.

## Related Issue

https://github.com/owncloud/core/issues/28985

## Motivation and Context

The change made here solves, the issues caused during migration.

## Types of changes

## Checklist:

### DIFF
--- a/apps/encryption/appinfo/Migrations/Version20170913113840.php
+++ b/apps/encryption/appinfo/Migrations/Version20170913113840.php
@@ -1,0 +1,49 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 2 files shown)


PVince81:
@sharidas please do a quick retest when coming from 10.0.3 to this branch (just one case is enough)

PVince81 (approved):
:+1: if it works

PVince81:
@sharidas please always add a link in the original PR to the backport

sharidas:
Test result of migration from 10.0.3beta ( tarball ) to the current change:
- [x] Migration of 10.0.3beta with masterkey enabled to the current master change. Works well. The encryption page doesn't show drop down and here is the screenshot: 
migrationwithmasterkeyfrom10-0-3

- [x] Migration from 10.0.3beta from userkey enabled to the current change. Works well. The encrytion page doesnt show the drop down and here is the screenshot : 
migrationwithuserkeyfrom10-0-3

PR #29069, opened 2017-09-21
Better control of which jobs run only for Travis CRON

phil-davis:
## Description
Use the new "if:" feature in the matrix to control which jobs even get created by Travis.

## Motivation and Context
Travis creates a job for each item in the matrix. Then if we only want the job to run in the context of a Travis CRON job (rather than with every PR) we had put the env var ``CRON_ONLY`` to indicate if we really want the job to do something. The ``CRON_ONLY`` jobs could exit early when run from ordinary PRs. But they still wasted time in the Travis job queue.

Travis has now released a (beta) feature to allow the starting of jobs in the matrix to be controlled by various conditions. See:
https://github.com/travis-ci/travis-ci/issues/7181#issuecomment-331099035
https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions
https://docs.travis-ci.com/user/conditional-builds-stages-jobs#Specifying-conditions

## How Has This Been Tested?
Trying some combinations in my own forked repo.

## Types of changes

## Checklist:

The Travis CI here will tell if it really works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,6 @@ addons:
     - owncloud
 
 before_install:
-  - if [ "$TC" = "syntax" ] || [ "$TEST_DAV" = "1" ] || { [ "$TC" = "selenium" ] && [ ! -z "$SAUCE_USERNAME" ] && [ ! -z "$SAUCE_ACCESS_KEY" ] && { [ "$CRON_ONLY" = "0" ] || [ "$TRAVIS_EVENT_TYPE" = "cron" ]; }; }; then echo "There is real work to do"; else exit 0; fi
   - make
   - sh -c "if [ '$TEST_DAV' = '1' ] || [ '$TC' = 'selenium' ]; then bash tests/travis/before_install.sh $DB; fi"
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -33,7 +33,6 @@ addons:
     - owncloud
 
 before_install:
-  - if [ "$TC" = "syntax" ] || [ "$TEST_DAV" = "1" ] || { [ "$TC" = "selenium" ] && [ ! -z "$SAUCE_USERNAME" ] && [ ! -z "$SAUCE_ACCESS_KEY" ] && { [ "$CRON_ONLY" = "0" ] || [ "$TRAVIS_EVENT_TYPE" = "cron" ]; }; }; then echo "There is real work to do"; else exit 0; fi
```
phil-davis: Note: this whole line can go, because it was just a short-circuit test to detect jobs that did not need to do anything. It avoided even doing the ``make`` on the next line for those cases. But now we can control which jobs are even created by Travis, so nothing will (should) need to use the "else" of this.

If someone has not set Sauce username and access key in their forked repo, then it is fine to go on and try anyhow - that way they will get error messages from SauceLabs about connections being refused.

individual-it (approved):

PR #29072, opened 2017-09-21
Fix for decrypting user specific keys

sharidas:
This change helps users to decrypt user
specific keys.

## Description

This change would help users to fix the user specific key encrypted files at oC.

## Related Issue

https://github.com/owncloud/core/issues/29081

## Motivation and Context

Without this change the decryption was failing files encrypted with user specific keys. With this change both masterkey encrypted files and user key encrypted files can be decrypted successfully.  

## How Has This Been Tested?

- [x] Enable encryption.
- [x] Enable master key encryption
- [x] Decryptall command worked successfully to decrypt the files of all users ( basically tested with all users ).
------------------------------
- [x] Enable encryption
- [x] Enable user specific key encryption.
- [x] Decryptall command worked successfully to decrypt files fo all users.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Encryption/DecryptAll.php
+++ b/lib/private/Encryption/DecryptAll.php
@@ -90,8 +90,10 @@ class DecryptAll {
 		}
 
 		$this->output->writeln('prepare encryption modules...');
-		if ($this->prepareEncryptionModules($user) === false) {
-			return false;
+		if (\OC::$server->getAppConfig()->getValue('encryption', 'useMasterKey', '0') !== '0') {
+			if ($this->prepareEncryptionModules($user) === false) {
+				return false;
+			}
 		}
 		$this->output->writeln(' done.');
 
@@ -183,6 +185,11 @@ class DecryptAll {
 		$userNo = 1;
 		foreach ($userList as $uid) {
 			$userCount = "$uid ($userNo of $numberOfUsers)";
+			if (\OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '0') !== '0') {
+				if ($this->prepareEncryptionModules($uid) === false) {
+					return false;
+				}
+			}
 			$this->decryptUsersFiles($uid, $progress, $userCount);
 			$userNo++;
 		}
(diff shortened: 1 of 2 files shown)


lib/private/Encryption/DecryptAll.php
```diff
@@ -183,6 +185,11 @@ protected function decryptAllUsersFiles($user = '') {
 		$userNo = 1;
 		foreach ($userList as $uid) {
 			$userCount = "$uid ($userNo of $numberOfUsers)";
+			if (\OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '0') !== '0') {
```
PVince81: why not also do the check on "useMasterKey" here for consistency ? please decide on one condition

lib/private/Encryption/DecryptAll.php
```diff
@@ -90,8 +90,10 @@ public function decryptAll(InputInterface $input, OutputInterface $output, $user
 		}
 
 		$this->output->writeln('prepare encryption modules...');
-		if ($this->prepareEncryptionModules($user) === false) {
-			return false;
+		if (\OC::$server->getAppConfig()->getValue('encryption', 'useMasterKey', '0') !== 0) {
```
jvillafanez: ❗️ 
Default value `'0'` (string) won't be the same as `0` (integer)
I'm not sure what is the expected return type, but please, make sure it matches

sharidas: @jvillafanez  Yes you are right. I have updated the PR.

lib/private/Encryption/DecryptAll.php
sharidas: The reason for not adding "useMasterKey" here is because we already do a check here: https://github.com/owncloud/core/pull/29072/files#diff-6f510adb432497f4f6a446f791729c79R93.
If "userSpecificKey" is enabled, then idea here is that we get the user and we pass it to method `prepareEncryptionModules`.  For "userSpecificKey" the decryption was failing because we relied on https://github.com/owncloud/core/blob/master/lib/private/Encryption/DecryptAll.php#L93

PVince81 (approved):
:+1:

PVince81:
@sharidas backport to stable10 ?

sharidas:
Backport for stable10 -> https://github.com/owncloud/core/pull/29189

PR #29077, opened 2017-09-21
New Cacheable Flysystem Adapter

Hemant-Mann:
## Description

I have added a new cacheable flysystem adapter in owncloud so that we can connect flysystem adapters and owncloud storage

## Related Issue

None

## Motivation and Context

This change is required because current flysystem implementation does not utilize caching and according to owncloud documentation this should be a must feature
>Within a single PHP request, ownCloud might call the same storage methods repeatedly, due to different checks which it needs to carry out. As a result, there is the potential to incur significant overhead, when working with the underlying filesystem

## How Has This Been Tested?

This has been tested with the new **Dropbox App** in which instead of extending the storage `OCP\Files\Storage\FlysystemStorageAdapter` I now extend it with `OCP\Files\Storage\CacheableFlysystemAdapter`

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/CacheableFlysystem.php
+++ b/lib/private/Files/Storage/CacheableFlysystem.php
@@ -1,0 +1,160 @@
+<?php
+/**
+ * @author Hemant Mann <hemant.mann121@gmail.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Files\Storage;
+
+use League\Flysystem\FileNotFoundException;
+use Icewind\Streams\IteratorDirectory;
+
+/**
+ * Generic Cacheable adapter between flysystem adapters and owncloud's storage system
+ *
+ * To use: subclass and call $this->buildFlysystem with the flysystem adapter of choice
+ */
+abstract class CacheableFlysystem extends \OC\Files\Storage\Flysystem {
+	/**
+	 * Stores the results in cache for the current request to prevent multiple requests to the API
+	 * @var array
+	 */
+	protected $cacheContents = [];
+
+	/**
+	 * Get the location which will be used as a key in cache
+	 * If Storage is not case sensitive then convert the key to lowercase
+	 * @param  string $path Path to file/folder
+	 * @return string
+	 */
+	public function getCacheLocation($path) {
+		$location = $this->buildPath($path);
+		if ($location === '') {
+			$location = '/';
+		}
+
+		if ($this->isCaseInsensitiveStorage) {
+			$location = strtolower($location);
+		}
+		return $location;
+	}
+
+	/**
+	 * Check if Cache Contains the data for given path, if not then get the data
+	 * from flysystem and store it in cache for this request lifetime
+	 * @param  string $path Path to file/folder
+	 * @return array|boolean
+	 */
+	public function getFlysystemMetadata($path) {
+		$location = $this->getCacheLocation($path);
+		if (!isset($this->cacheContents[$location])) {
+			try {
+				$this->cacheContents[$location] = $this->flysystem->getMetadata($location);
+			} catch (FileNotFoundException $e) {
+				$this->cacheContents[$location] = false;
+			}
+		}
+		return $this->cacheContents[$location];
+	}
+
+	/**
+	 * Store the list of files/folders in the cache so that subsequent requests in the
+	 * same request life cycle does not call the flysystem API
+	 * @param  array $contents Return value of $this->flysystem->listContents
+	 */
+	public function updateCache($contents) {
+		foreach ($contents as $object) {
+			$path = $object['path'];
+			$this->cacheContents[$path] = $object;
+		}
+	}
+
+	/**
+	 * Check for existence of file/folder for the given path
+	 * @param  string  $path Path to file/folder
+	 * @return boolean
+	 */
+	public function hasPath($path) {
+		return (bool) $this->getFlysystemMetadata($path);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function opendir($path) {
+		try {
+			$content = $this->flysystem->listContents($this->buildPath($path));
+			$this->updateCache($content);
+		} catch (FileNotFoundException $e) {
+			return false;
+		}
+		$names = array_map(function ($object) {
+			return $object['basename'];
+		}, $content);
+		return IteratorDirectory::wrap($names);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function file_exists($path) {
+		return $this->hasPath($path);
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function filesize($path) {
+		if ($this->is_dir($path)) {
+			return 0;
+		} else {
+			$info = $this->getFlysystemMetadata($path);
+			return $info['size'];
+		}
+	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function filemtime($path) {
+		$info = $this->getFlysystemMetadata($path);
+		return $info['timestamp'];
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: the Travis-CI UI test jobs that make use of SauceLabs fail because they do not have any SauceLabs credentials. This is because the PR has come from a forked repo. Only PRs coming from the "master" ownCloud core repo have access to the ownCloud SauceLabs credentials, which are stored securely in the ownCloud Travis-CI account.

lib/private/Files/Storage/CacheableFlysystem.php
```diff
@@ -0,0 +1,166 @@
...
+	 * @return array|boolean
+	 */
+	public function getFlysystemMetadata($path, $overRideCache = false) {
+		$location = $this->getCacheLocation($path);
+		if (!isset($this->cacheContents[$location]) || $overRideCache) {
+			try {
+				$this->cacheContents[$location] = $this->flysystem->getMetadata($location);
+			} catch (FileNotFoundException $e) {
+				// do not store this info in cache as it might interfere with Upload process
+                return false;
```
tomneedham: indentation

lib/private/Files/Storage/CacheableFlysystem.php
```diff
@@ -0,0 +1,170 @@
...
+		$location = $this->buildPath($path);
+		if ($location === '') {
+			$location = '/';
+		}
+		return $location;
+	}
+
+	public function buildPath($path) {
+		$location = parent::buildPath($path);
+		if ($this->isCaseInsensitiveStorage) {
```
PVince81: ewww, is that a thing in Dropbox ?!

lib/private/Files/Storage/CacheableFlysystem.php
```diff
@@ -0,0 +1,170 @@
...
+			$location = strtolower($location);
+		}
+		return $location;
+	}
+
+	/**
+	 * Check if Cache Contains the data for given path, if not then get the data
+	 * from flysystem and store it in cache for this request lifetime
+	 *
+	 * @param  string $path Path to file/folder
```
PVince81: document `$overRideCache`

lib/private/Files/Storage/CacheableFlysystem.php
```diff
@@ -0,0 +1,170 @@
...
+	 * @return array|boolean
+	 */
+	public function getFlysystemMetadata($path, $overRideCache = false) {
+		$location = $this->getCacheLocation($path);
+		if (!isset($this->cacheContents[$location]) || $overRideCache) {
+			try {
+				$this->cacheContents[$location] = $this->flysystem->getMetadata($location);
+			} catch (FileNotFoundException $e) {
+				// do not store this info in cache as it might interfere with Upload process
```
PVince81: We should look into storing this info at some point. I know it works because we do this for other storages like DAV.

Maybe the upload interference is a missing stat clear in another place in the code.

I suggest to look into this separately.

lib/public/Files/Storage/CacheableFlysystemAdapter.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+
+namespace OCP\Files\Storage;
+
+/**
+ * Public Cacheable storage adapter to be extended to connect to
+ * flysystem adapters
+ *
+ * @since 10.0.x
```
PVince81: make that 10.0.4

PVince81 (commented):
See comments above, more to come

PVince81:
Is this mtime caching specific to Dropbox ? https://github.com/owncloud/files_external_dropbox/blob/master/lib/Storage/Dropbox.php#L71

Isn't this information already available in this cache ? If not, does it make sense to bring this into your cacheable adapter as well or is it very specific to Dropbox ?

PVince81:
Shouldn't you cache the results you got here https://github.com/owncloud/files_external_dropbox/blob/master/lib/Storage/Dropbox.php#L144 into the local cache as well ?

PVince81:
Seems `isCaseInsensitiveStorage` is unused, remove ? https://github.com/owncloud/files_external_dropbox/blob/master/lib/Storage/Dropbox.php#L58

lib/private/Files/Storage/CacheableFlysystem.php
```diff
@@ -0,0 +1,170 @@
...
+	 */
+	public function fopen($path, $mode) {
+		switch ($mode) {
+			case 'r':
+			case 'rb':
+				return parent::fopen($path, $mode);
+
+			default:
+				unset($this->cacheContents[$this->getCacheLocation($path)]);
+				return parent::fopen($path, $mode);
```
PVince81: it should be possible to invalidate the cache after an upload, see how it's done here: https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/DAV.php#L405 and https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/DAV.php#L503

lib/private/Files/Storage/CacheableFlysystem.php
Hemant-Mann: Yes unfortunately it is
Suppose you have the following directory structure
```
Foo/
├── bar
│   ├── foo.txt
│   └── some_another_file.txt
└── somefile.txt
```
In one case dropbox api was returning the folder name **Foo** as both _Foo_ and _foo_ due to which the directory list of the folder Foo was coming as empty but in reality it was not

Hemant-Mann:
>Is this mtime caching specific to Dropbox ? https://github.com/owncloud/files_external_dropbox/blob/master/lib/Storage/Dropbox.php#L71
>
>Isn't this information already available in this cache ? If not, does it make sense to bring this into your cacheable adapter as well or is it very specific to Dropbox ?

Yes it is specific to dropbox because in API v2 we don't get the modified time of folders so what I did was find the modified time of the files in the folder and get the largest of those values and this info is not available in cache

Hemant-Mann:
>Shouldn't you cache the results you got here https://github.com/owncloud/files_external_dropbox/blob/master/lib/Storage/Dropbox.php#L144 into the local cache as well ?

Uhh I am only fetching this info if not found in cache here https://github.com/owncloud/files_external_dropbox/blob/master/lib/Storage/Dropbox.php#L139

because the cache is only for a single request cycle i think this should work

lib/private/Files/Storage/CacheableFlysystem.php
Hemant-Mann: Yes i faced some problem with upload process if i stored the info that file is not present in storage

lib/private/Files/Storage/CacheableFlysystem.php
Hemant-Mann: Ah okay that's cool

lib/private/Files/Storage/CacheableFlysystem.php
Hemant-Mann: done see below this line

PVince81 (approved):
:+1: fine for now, will need further optimizations in the future by adding the cache stuff in more code paths (fopen, etc)

PVince81:
@Hemant-Mann can you rebase ? Then do another PR to backport to stable10.

Hemant-Mann:
Okay I'll do that

Hemant-Mann:
Stable 10 PR #29414

>👍 fine for now, will need further optimizations in the future by adding the cache stuff in more code paths (fopen, etc)

Yes we can definitely add cache stuff at more places like fopen

PVince81:
@Hemant-Mann thanks for the backport, can you also rebase this PR to kick CI ?

phil-davis:
Nogte: this PR is coming from a forked repo, so the Travis-Saucelabs UI tests will fail due to Saucelabs authentication issues.

Hemant-Mann:
@PVince81 Can you explain a little bit as to why I have to rebase it?

Is it related to @phil-davis issue
>Note: this PR is coming from a forked repo, so the Travis-Saucelabs UI tests will fail due to Saucelabs authentication issues.

phil-davis:
I see it also did not trigger the drone CI. There were some times recently when drone was down, so the job did not trigger. So rebasing will make the drone CI run.
(It will make all the other CI run also, and Travis will fail for the reason above, which is fine since we understand why)

Hemant-Mann:
I rebased the branch with master but while pushing it rejected the commits and then I had to do a git pull which resulted in a merge commit

```
commit 5b1a17b6df9316d02f01d095ca64afc8a8869f60 (HEAD -> cacheable_storage_adapter)
Merge: b066f40340 fb6e3978b9
Author: Hemant Mann <hemant.mann121@gmail.com>
Date:   Thu Nov 2 22:01:50 2017 +0530

    Merge branch 'cacheable_storage_adapter' of https://github.com/Hemant-Mann/owncloud_core into cacheable_storage_adapter
```
So should I push this rebased branch?

phil-davis:
You can locally do an interactive rebase:
```
git rebase -i master
```
and squash the commits together in whatever way seems logical (squashing out the merge commit...)
then force push:
```
git push -f
```
that will replace whatever is here in the PR branch with whatever you have done locally.

Hemant-Mann:
Thanks I was just missing the `-f` flag

Issue #28669, opened 2017-08-14
SFTP + keypair mode doesn't work any more on stable10


PVince81:
### Steps

1. Setup stable10
1. Setup a system-wide SFTP mount with "RSA public key" mode
1. Copy the given key into "~/ssh/.authorized_keys"
1. Type in the user name and other stuff, wait for refresh

### Expected
Green light

### Actual
Red.

### Version
stable10 1f0a3b077bdeba38ec5231b926cdb281e26b1f06

Log is full of:
```
{"reqId":"nGG4Y8OWrYTttMWrtzN1","level":3,"time":"2017-08-14T12:28:32+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"PHP","method":"GET","url":"\/owncloudee\/ocs\/v2.php\/apps\/notifications\/api\/v1\/notifications?format=json","message":"pack(): Type H: illegal hex digit - at \/srv\/www\/htdocs\/owncloudee\/lib\/composer\/phpseclib\/phpseclib\/phpseclib\/Crypt\/RSA.php#1081"}
```

PVince81:
```
0  phpseclib\Crypt\RSA->_parseKey() /srv/www/htdocs/owncloud/lib/composer/phpseclib/phpseclib/phpseclib/Crypt/RSA.php:1081
1  phpseclib\Crypt\RSA->loadKey() /srv/www/htdocs/owncloud/lib/composer/phpseclib/phpseclib/phpseclib/Crypt/RSA.php:1565
2  OCA\Files_External\Lib\Auth\PublicKey\RSA->manipulateStorageConfig() /srv/www/htdocs/owncloud/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php:63
3  OCA\Files_External\Controller\GlobalStoragesController->manipulateStorageConfig() /srv/www/htdocs/owncloud/apps/files_external/lib/Controller/StoragesController.php:228
4  OCA\Files_External\Controller\GlobalStoragesController->updateStorageStatus() /srv/www/htdocs/owncloud/apps/files_external/lib/Controller/StoragesController.php:245
5  OCA\Files_External\Controller\GlobalStoragesController->update() /srv/www/htdocs/owncloud/apps/files_external/lib/Controller/GlobalStoragesController.php:187
[...]

PR #29098, opened 2017-09-26
Get SFTP key pair mode working

sharidas:
Due to removal of '\n' and '\r' from
the backendoptions the private keys were affected.
And hence resulted in failure to login. So as an
alternative in this patch we use base64_encode
and base64_decode to convert both the public and
private keys. Another advantage is that we don't
expose the private keys in the database.

## Description

SFTP was failing to work with the change made https://github.com/owncloud/core/commit/9b88fa6c8c0ebaa7d471476821cfc70778620ce9 . The key problem found was failure to authenticate with the private key.

## Related Issue

https://github.com/owncloud/core/issues/28669

## Motivation and Context

The issue was caused due to the strip of value in backendoptions. And due to this the private key was modified. Hence authentication failure. This change helps users to convert public key and private key using base64_encode and retrieve the data using base64_decode. 

## How Has This Been Tested?

- [x] Enable SFTP storage
- [x] Select the public key option
- [x] Copy the public key option shown in the UI to the destination server.
- [x] The green light is shown in the ui
Migration test:
With the migration I have noticed the old data remains the same. So if new key is copied to the server using generate keys option in the UI, it works. So both the old keys work as well as the new one generated.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
+++ b/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
@@ -60,9 +60,11 @@ class RSA extends AuthMechanism {
 	public function manipulateStorageConfig(IStorageConfig &$storage, IUser $user = null) {
 		$auth = new RSACrypt();
 		$auth->setPassword($this->config->getSystemValue('secret', ''));
-		if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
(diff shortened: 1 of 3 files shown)


peterprochaska (approved):

sharidas:
Backport to stable10: https://github.com/owncloud/core/pull/29156

PR #29097, opened 2017-09-26
UI tests for federation sharing

individual-it:
## Description
UI test that check the federation sharing features

## Related Issue
owncloud/QA#437

## Motivation and Context
automate smoke tests

## How Has This Been Tested?
run tests locally and in travis

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29099, opened 2017-09-26
[stable10] UI tests for federation sharing

phil-davis:
Backport #29097


phil-davis:
Added backport for commit from master PR #29102 that makes these tests more stable. These are best back-ported together.

individual-it (approved):

Issue #25358, opened 2016-07-05
Improve quota validation including a regexp which matches expected formats


SergioBertolinSG:
Curently an admin user can enter a number probably followed by letters.
For example: "200", "15MB" or "5 GB" "20B".
Basically it's a number probably followed by letters.

Also floating point can be entered, but only in english style "2.5GB". Using a non english format for floats will lead to an old bug "4,5GB" also something like "45/0". 

So the proposed enhancement by @PVince81 in https://github.com/owncloud/core/pull/25253 is to develop a regexp to validate better this values.

PVince81:
Thanks for the ticket, makes sense

andypmw:
Hi guys, I am create a regex formula for this issue:
`^( +)?([+-])?[\s+]?([0-9]*)(\.([0-9]+))?( +)?(b|kb|mb|gb|tb|pb|eb|zb|yb)?`

I will try to write on PHP and create pull request

PVince81:
@andypmw sounds good, thanks!

SergioBertolinSG:
> I will try to write on PHP and create pull request

I guess this validation should be done in the frontend: settings/js/users/users.js https://github.com/owncloud/core/blob/master/settings/js/users/users.js

Kawohl:
@andypmw Did you make any progress yet?

PVince81:
@ppaysant has added a new JS method for parsing this kind of strings, see https://github.com/owncloud/core/pull/26789

SamuAlfageme:
Think this is a must, since input an invalid quota (e.g. any alphanumeric string) and you'll end up with this funky quota list made with `B`s:

![screen shot 2016-12-14 at 10 34 56](https://cloud.githubusercontent.com/assets/2644445/21176876/fdd84952-c1e8-11e6-9e2b-a7312573e990.png)

Note that the last "`Other`" option has been replaced by another `B`.

individual-it:
reopening as this is not fixed yet  #27257 only fixed the function that could potentially be used for validation but its not implemented in the onQuotaSelect function.
maybe I should not have put in "prerapation to fix #25358" in the header of the PR

SergioBertolinSG:
Tested and works fine. 

But repeated values are added to the dropdown. They vanish when reloading the page. @PVince81 there is an open issue about this bug? I cannot find it.

phil-davis:
The fix here did not allow a 0 byte quota to be entered. See separate issue #29105 and PR #29109

[...]

Issue #29105, opened 2017-09-27
Can't set quota to 0


voroyam:
In 10.0.3 we can't set user's quota to 0. In 9.1 it was possible

phil-davis:
There are set quota tests in the UI tests. I am surprised that we never tested that possibility. I am happy to have a look.

PVince81:
fixed through https://github.com/owncloud/core/pull/29109

PR #29109, opened 2017-09-27
Allow 0 byte quota to be entered on UI

phil-davis:
## Description
1) Add a test for entering 0 byte quota (note: 0 can be entered as "0" "0 B" "0 KB" etc)
2) While we are here, add a test to make sure entering a quota of "-1" fails.
3) Improve the check of the return value from ``computerFileSize()``

## Related Issue
#29105 
also improves on old closed issue #25358

## Motivation and Context
It be broke.

## How Has This Been Tested?
Run automated UI tests in a local dev VM. Watch it work.

## Types of changes

## Checklist:

### DIFF
--- a/settings/js/users/users.js
+++ b/settings/js/users/users.js
@@ -542,7 +542,7 @@ var UserList = {
 		}
 		if (
 			['default', 'none'].indexOf(quota) === -1
-			&& (!OC.Util.computerFileSize(quota))
+			&& (OC.Util.computerFileSize(quota) === null)
 		) {
 			// the select component has added the bogus value, delete it again
 			$select.find('option[selected]').remove();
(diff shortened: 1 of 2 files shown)


SergioBertolinSG (approved):

phil-davis:
For people that just want to offer shared folders to a user, the user's own quota needs to be zero.
Then the user cannot save any real files in their own "personal" area.
Note: there is another issue about the fact that a user with zero quota can create empty files and empty folders - but that is a separate thing that follows on after this PR which lets the admin actually set zero quota.

hodyroff:
In 9.1 with 0 quota I am not able to create any file or folder (Web Frontend).

phil-davis:
Now that zero quota can be entered, there will need to be followup to test exactly what can and cannot be done by a user with 0 bytes quota. Then decide if that is the desired behavior or not.

davitol:
> Now that zero quota can be entered, there will need to be followup to test exactly what can and cannot be done by a user with 0 bytes quota. Then decide if that is the desired behavior or not.

@pmaier1 @PVince81 FYI

PVince81:
I'd say the same things as a user who has filled their storages completely and has 0 bytes available. I don't think "0 bytes quota" should have a special meaning here.

PR #29110, opened 2017-09-27
fix app_type check

phisch:
## Description
Currently we have no way to get all apps of a certain type that includes apps that have never been enabled before.

When getting a list of all themes, the AppManager does only look into the appconfig table to get a list of apps. This means apps that have never been enabled will not get returned by the AppManager. That's why i am using the legacy `OC_App::getAllApps()` which iterates over the fs.

When checking if an app has a specific app_type, we only ever look into the appconfig table as well. That is why I included a check to also look into the info.xml whenever there is nothing in the appconfig table and also add the found types to the static list of app types, so they will be available through `self::getAppTypes($app)`.

To be clear, *this is a hack*. It will add unreasonable fs reads, but it is necessary for the Templateeditor to work properly.

@DeepDiver1975 @PVince81 we should talk about a small refactoring of the app structure or how to extend the AppManager so even never enabled apps are properly available in the code.

## Related Issue
https://github.com/owncloud/templateeditor/pull/52

## How Has This Been Tested?
manual only

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Theme/ThemeService.php
+++ b/lib/private/Theme/ThemeService.php
@@ -153,7 +153,8 @@ class ThemeService implements IThemeService {
 	 */
 	private function getAllAppThemes() {
 		$themes = [];
-		foreach (\OC::$server->getAppManager()->getAllApps() as $app) {
+
+		foreach (\OC_App::getAllApps() as $app) {
 			if (\OC_App::isType($app, 'theme')) {
 				$themes[$app] = $this->makeTheme($app);
 			}
--- a/lib/private/legacy/app.php
+++ b/lib/private/legacy/app.php
@@ -264,6 +264,10 @@ class OC_App {
 			$types = [$types];
 		}
 		$appTypes = self::getAppTypes($app);
+		if (count($appTypes) === 0) {
+			self::setAppTypes($app);
+			$appTypes = self::getAppTypes($app);
+		}
 		foreach ($types as $type) {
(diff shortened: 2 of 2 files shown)


PVince81:
@phisch why would it be necessary to get app types of apps that have never been enabled before ? From what I understand only enabled apps / theme apps are of interest here ?

phisch:
@PVince81 there is the possibility that a theme has never been enabled, for example if you install a fresh ownCloud with lets say the enterprise theme, the example theme and a custom one. Usually this would not be a problem, since there is no component in the code that needs a list of all apps, enabled or disabled, filtered by the type. 

The TemplateEditor does need such a list though. It needs to grab all apps that are of the type Theme.

Currently the ThemeManager provides a function to return all Themes, and that fails with never enabled themes. (just doesn't return those themes)

PVince81:
Why does the template editor need to list theme apps that were never enabled ? I don't think the template editor should let you edit templates from disabled apps anyway

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

DeepDiver1975:
@VicDeo I assigned you since you took over the tasks from @phisch

VicDeo:
@PVince81 we agreed that only enabled app-theme should be editable with Templateeditor. 
Close?

PR #29111, opened 2017-09-27
[stable10] Allow 0 byte quota to be entered on UI

phil-davis:
Backport #29109


phil-davis:
Jenkins never started - closing.

PR #29124, opened 2017-09-29
stabilize UI tests

individual-it:
## Description
a set of changes to make the UI tests more stable and reduce the false positive, specially with IE 11
- split tests into smaller suites
- upload log files of failed test to a WebDav account currently to: https://jankaritech.ocloud.de/index.php/s/FR9ySrgEBfkijTL
- better waiting in one place
- disable CSRF checking
- retry moving files
- scroll only a page height while searching for files
- focus found file
- increase resolution for IE
- while WIP run IE tests on PR and run them first because they are the most unstable

## Related Issue
most tests would fail when running in IE

## Motivation and Context
Is it bad to have tests that fail on a regular basis without anything being wrong. Developers would get used to failing tests and would not care when they fail for real.

## How Has This Been Tested?
run tests in travis

## Types of changes

## Checklist:


phil-davis:
Looks good. It is unfortunate that we have to do all that messy stuff calculating things like ``spaceLeftTillBottom`` - it would be nice if ``scrollIntoView()`` would "just work" on all the browsers so we could simply call that do do what the user would do in real life - scroll the file into view! Looking forward to 2020 and the death of IE.

phil-davis (approved):
Looks good. CI should pass.

individual-it:
backport in  #29129

Issue #23186, opened 2016-03-12
Integrity Check: Move excluded files/folders into config array


LukasReschke:
Ref https://github.com/owncloud/core/blob/5862a067d9d67d3ef70cb064009f81b47ed15efd/lib/private/integritycheck/iterator/excludefoldersbypathfilteriterator.php#L35-L41

nickvergessen:
Btw, while you are on it, can you remove the skeleton folder `core/skeleton/` from the check?

LukasReschke:
> Btw, while you are on it, can you remove the skeleton folder core/skeleton/ from the check?

Why? https://github.com/owncloud/core/blob/c7482eee4eaecefc3e8f68a8dbdce382a8858ac0/config/config.sample.php#L197-L202 is what one should use for custom skeletons :wink:

nickvergessen:
https://github.com/owncloud/core/issues/23857#issuecomment-208487469

LukasReschke:
No. That's just an user who updated incorrectly or used the wrong ZIP files or whatever. – That's a bug on their side

beat:
stats and icon folders, and awstats-icon and awstatsicons links are standard for awstats (a popular open-source web-stats package). When installing OwnCloud into the top folder, with open-source VirtualMin panel (and default awstats enabled), those folders exist, and a way to ignore them should be added (e.g. in config.php file), same like content in data (and in config.php too) are ignored.

PVince81:
cc @tomneedham

dalenoe:
I know this is closed. However, I am wondering what is the best way to exclude a file from the integrity check? I have a custom haproxy-check.php script that resides in the same folder as owncloud, and that yellow message nagging in the admin section gets old.

PVince81:
Setting to junior job. Might not be entry-level but should still be easy enough.

voroyam:
In the ownCloud appliance, a certificate has to be imported in this dir
/var/www/owncloud/resources/config/ca-bundle.crt

After an upgrade the integrity check fails, telling me that the hash of this file is different than expected.

can we exclude this file or dir from the check?

Peter-Prochaska:
We can exclude files, but not whole folders </statement>

PVince81:
Basically, no we can't at the moment. But after implementing this ticket, only for files (not folders), it would become possible to configure those through config.php.

PVince81:
https://github.com/owncloud/core/pull/29126

Issue #29094, opened 2017-09-25
No signature validation for theme apps


DeepDiver1975:
refs https://github.com/owncloud/appstore-issues/issues/105#issuecomment-331486098

We need a mechanism which allows users to implement their own theme without getting into the signature warning.

Maybe my ignoring all theme apps in the integrity check?

@michaelstingl we have been chatting about this long time ago - right?

@phisch regarding themes

DeepDiver1975:
or at least allow unsigned hashes - see https://github.com/owncloud/appstore-issues/issues/105#issuecomment-331874193

svengit:
#23186 would probably be sufficient to solve this.

michaelstingl:
Currently I only remove `signature.json` from my theme app and after a rescan I get no more warnings. I think something has been already done about it? /cc @phisch

svengit:
This did not work in Owncloud 10.0.2 a few days ago. The integrity checker complained about a missing signature

michaelstingl:
With "currently" I meant 10.0.3. Sorry for being not precise.

phisch:
The thing is, we do want signature checking for apps (themes) from the marketplace, but not for custom themes.

A few months ago we talked about a blacklist in the config.php, I can't find that discussion right now but i think we talked about possible security issues here, which where debunked by the fact that a hacker could un-patch the integrity check eitherway. So adding the possibility to add apps to an integrity blacklist in config.php would be a good option here. That way apps/themes are still checked, but you could disable the check for a custom theme/app you have for your owncloud only.

@DeepDiver1975 @PVince81 any thoughts on this? I would probably go that way if you guys have no concerns about this!

svengit:
Yes, signature checking for the marketplace is literally the only thing that makes sense. You could revert the switch from themes/ to the apps/ directory for themes and just limit signature checking to the apps/ dir or even to the moment an app gets installed.
If you want further integrity checking for the whole webdir, unsigned checksums will be sufficient.

PVince81:
@Peter-Prochaska thoughts ?

Peter-Prochaska:
I like the idea with the blacklist for own apps/themes in config.php. But not for complete folders.

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

BrainStone:
Any updates on this issue?

thommierother:
[...]

PR #29126, opened 2017-09-29
Move excluded files to config array

sharidas:
Moving excluded files to config array.
Instead of populating in the ExcludeFile* 
php files, we can have it in config.php.

## Description

This change would help users to mention folders/files to be excluded during integrity check to the config.php. In the config.php it will be represented as an array.

## Related Issue

https://github.com/owncloud/core/issues/23186
https://github.com/owncloud/enterprise/issues/1874

## Motivation and Context

Instead of hard coding the files/folders in the files, its better to move them to the config.php. Also this change would help the integrity check to exclude files/folders mentioned in the config.php

## Types of changes

## Checklist:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1210,6 +1210,18 @@ $CONFIG = array(
 		'~snapshot',
 	),
 /**
+ * Exclude directories from the integrity checker command
+ */
+'integrity.excluded.files' =>
+	array (
+		'.DS_Store',
+		'Thumbs.db',
+		'.directory',
+		'.webapp',
+		'.htaccess',
+		'.user.ini',
+	),
+/**
  * Define a default folder for shared files and folders other than root.
  */
 'share_folder' => '/',
--- a/lib/private/IntegrityCheck/Checker.php
+++ b/lib/private/IntegrityCheck/Checker.php
@@ -230,7 +230,7 @@ class Checker {
 	 * @param array $hashes
 	 * @param X509 $certificate
 	 * @param RSA $privateKey
-	 * @return string
+	 * @return array
 	 */
 	private function createSignatureData(array $hashes,
 										 X509 $certificate,
@@ -382,6 +382,9 @@ class Checker {
 			throw new InvalidSignatureException('Signature could not get verified.');
 		}
 
+		//Exclude files which shouldn't fall for comparison
+		$excludeFiles = $this->config->getSystemValue('integrity.excluded.files', []);
+
 		// Compare the list of files which are not identical
 		$currentInstanceHashes = $this->generateHashes($this->getFolderIterator($basePath), $basePath);
 		$differencesA = array_diff($expectedHashes, $currentInstanceHashes);
@@ -389,6 +392,10 @@ class Checker {
 		$differences = array_unique(array_merge($differencesA, $differencesB));
 		$differenceArray = [];
 		foreach($differences as $filename => $hash) {
+			//If filename in exclude files list, then ignore it
+			if (in_array($filename, $excludeFiles, true)) {
+				continue;
+			}
 			// Check if file should not exist in the new signature table
 			if(!array_key_exists($filename, $expectedHashes)) {
 				$differenceArray['EXTRA_FILE'][$filename]['expected'] = '';
--- a/tests/lib/IntegrityCheck/CheckerTest.php
+++ b/tests/lib/IntegrityCheck/CheckerTest.php
@@ -159,15 +159,17 @@ class CheckerTest extends TestCase {
 	}
 
 	public function testVerifyAppSignatureWithValidSignatureData() {
+		$map = [
+			['integrity.check.disabled', false, false],
+			['integrity.excluded.files', [], array()],
+		];
 		$this->environmentHelper
 				->expects($this->once())
 				->method('getChannel')
 				->will($this->returnValue('stable'));
 		$this->config
-				->expects($this->any())
-				->method('getSystemValue')
-				->with('integrity.check.disabled', false)
-				->will($this->returnValue(false));
+			->method('getSystemValue')
+			->will($this->returnValueMap($map));
 
 		$this->appLocator
 				->expects($this->once())
@@ -249,15 +251,17 @@ class CheckerTest extends TestCase {
 	}
 
 	public function testVerifyAppSignatureWithTamperedFiles() {
+		$map = [
+			['integrity.check.disabled', false, false],
+			['integrity.excluded.files', [], array()],
+		];
 		$this->environmentHelper
 				->expects($this->once())
 				->method('getChannel')
 				->will($this->returnValue('stable'));
 		$this->config
-				->expects($this->any())
-				->method('getSystemValue')
-				->with('integrity.check.disabled', false)
-				->will($this->returnValue(false));
+			->method('getSystemValue')
+			->will($this->returnValueMap($map));
 
 		$this->appLocator
 				->expects($this->once())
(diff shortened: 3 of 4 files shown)


lib/private/IntegrityCheck/Checker.php
```diff
@@ -398,13 +398,21 @@ private function verify($signaturePath, $basePath, $certificateCN, $force = fals
 			throw new InvalidSignatureException('Signature could not get verified.');
 		}
 
+		//Exclude files which shouldn't fall for comparison
+		$excludeFiles = \OC::$server->getSystemConfig()->getValue('excludedFiles', '');
```
tomneedham: default should be an array since we use it in `in_array` later

tomneedham:
should also include a sample entry in config.sample.php

config/config.sample.php
```diff
@@ -1205,6 +1205,37 @@
 		'~snapshot',
 	),
 /**
+ * Exclude directories from the integrity checker command
+ */
+'excludedFolders' =>
+	array (
+		0 => '/home/sujith/test/owncloud/data',
+		1 => '/home/sujith/test/owncloud/themes',
```
DeepDiver1975: better use OC::$SERVERROOT

example
```php
OC::$SERVERROOT . '/assets'

Comment (DeepDiver1975) [reply config/config.sample.php]: and we shall find good example folders - maybe two are enough.
/apps/my-theme might be a good example and /assets

File: config/config.sample.php [hunk 3]
```diff
@@ -1205,6 +1205,37 @@
...
+		1 => 'Thumbs.db',
+		2 => '.directory',
+		3 => '.webapp',
+		4 => '.htaccess',
+		5 => '.user.ini',
+	),
+/**
+ * Exclude file name patterns from the integrity checker command
+ */
+'excludedFilenamePatterns' =>
```
DeepDiver1975: isn't the pattern enough to also properly exclude folders and files from above?

lib/private/IntegrityCheck/Checker.php
```diff
@@ -398,13 +398,21 @@ private function verify($signaturePath, $basePath, $certificateCN, $force = fals
 			throw new InvalidSignatureException('Signature could not get verified.');
 		}
 
+		//Exclude files which shouldn't fall for comparison
+		$excludeFiles = \OC::$server->getSystemConfig()->getValue('excludedFiles', []);
```
DeepDiver1975: inject config instance - no use of \OC::$server inside any class please

lib/private/IntegrityCheck/Iterator/ExcludeFileByNameFilterIterator.php
```diff
@@ -78,4 +85,12 @@ public function accept() {
 
 		return true;
 	}
+
+	protected function getExcludedFileNames() {
+		return \OC::$server->getConfig()->getSystemValue('excludedFiles', '');
```
DeepDiver1975: keep proper defaults - basically just like before

lib/private/Setup.php
```diff
@@ -317,6 +340,9 @@ public function install($options) {
 			'overwrite.cli.url'	=> $request->getServerProtocol() . '://' . $request->getInsecureServerHost() . \OC::$WEBROOT,
 			'dbtype'			=> $dbType,
 			'version'			=> implode('.', \OCP\Util::getVersion()),
+			'excludedFolders'   => $excludeFolders,
```
DeepDiver1975: I would no set them during setup. the defaults will be in our code base and people can extend/change the list if they like.

tests/lib/IntegrityCheck/CheckerTest.php
```diff
@@ -497,114 +497,6 @@ public function testWriteCoreSignature() {
 		$this->checker->writeCoreSignature($x509, $rsa, \OC::$SERVERROOT . '/tests/data/integritycheck/app/');
 	}
 
-	public function testWriteCoreSignatureWithUnmodifiedHtaccess() {
```
DeepDiver1975: where did these tests go?

DeepDiver1975 (changes requested):

PVince81:
I heard from @Peter-Prochaska that it might not be a good idea to allow excluding whole folders, so not sure... See https://github.com/owncloud/core/issues/29094#issuecomment-335396910

PVince81:
- [ ] Please only allow excluding files, not folders, as discussed with @Peter-Prochaska

mmattel:
Maybe only a small detail to avoid naming confusion:
``excludedFolders`` --> ``integrityExcludedFolders`` also valid for the others.

Reason: https://doc.owncloud.org/server/latest/admin_manual/configuration/server/excluded_blacklisted_files.html

sharidas:
Updated this PR to exclude files only and not folders.

config/config.sample.php
```diff
@@ -1210,6 +1210,25 @@
 		'~snapshot',
 	),
 /**
+ * Exclude directories from the integrity checker command
+ */
+'excludedFiles' =>
+	array (
```
mmattel: Description text: Exclude directories --> Exclude files

mmattel: Also see my comment above to avoid confusion:
This variable should be renamed from ``excludedFiles`` to ``integrityExcludedFiles``

sharidas:
Not sure how to make codecov happy for config.sample.php. Any pointers would be helpful. Thanks.

config/config.sample.php
```diff
@@ -1210,6 +1210,25 @@
 		'~snapshot',
 	),
 /**
+ * Exclude directories from the integrity checker command
+ */
+'integrityExcludedFiles' =>
+	array (
+		0 => '.DS_Store',
```
PVince81: no need for the numeric index

sharidas: Updated the config.sample.php by removing numeric index.

sharidas:
Backport PR to stable10 branch: https://github.com/owncloud/core/pull/29460

config/config.sample.php
```diff
@@ -1210,6 +1210,25 @@
 		'~snapshot',
 	),
 /**
+ * Exclude directories from the integrity checker command
+ */
+'integrityExcludedFiles' =>
```
PVince81: why camel case ? all of the other config entries are either dot separated, underscore separated or have no separator for all-lowercase. The latest trend is to use dots as separators.

config/config.sample.php
PVince81: not sure... maybe some files could have prefixes/suffixes ? I'd also prefer the simpler solution.

@sharidas Did anyone report such cases in the original ticket ?

PVince81:
@Peter-Prochaska second review ?

config/config.sample.php
sharidas: @PVince81 No I didn't find any cases related to exclude filename patterns in the original ticket. I think we can stick on with exclude file names and remove exclude filename patterns from the PR.

config/config.sample.php
sharidas: Updated to `integrity.excluded.files`

lib/private/IntegrityCheck/Checker.php
```diff
@@ -230,7 +230,7 @@ private function generateHashes(\RecursiveIteratorIterator $iterator,
 	 * @param array $hashes
 	 * @param X509 $certificate
 	 * @param RSA $privateKey
-	 * @return string
+	 * @return mixed
```
Peter-Prochaska: Why mixed? it's an array

peterprochaska (changes requested):

lib/private/IntegrityCheck/Checker.php
sharidas: @Peter-Prochaska  Updated the PR.

peterprochaska (approved):

config/config.sample.php
```diff
@@ -1210,6 +1210,18 @@
 		'~snapshot',
 	),
 /**
+ * Exclude directories from the integrity checker command
+ */
+'integrity.excluded.files' =>
```
mmattel: @PVince81 we still have the description text mismatch stating that we exclude directories while we exclude files ...
This inconsistancy needs to be fixed and backported

mmattel: @PVince81 just seen that is has been fixed ...

PR #28073, opened 2017-06-02
Bye bye Dropbox v1

PVince81:
## Description
Removes Dropbox v1 external storage from core as the v1 API will be turned off very soon.
Will require https://github.com/owncloud/core/pull/26539 to avoid breaking OC if there are existing mounts when upgrading and the new app is not installed yet.

## Related Issue
See https://github.com/owncloud/core/issues/25969 for context.

## How Has This Been Tested?
Retested the web UI files app and it still works.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

@DeepDiver1975 @jvillafanez @butonic

PR #26539, opened 2016-11-03
Handle invalid ext storage backend to keep mount point visible

PVince81:
## Description
Keep mount point visible and also ext storage config visible when
dealing with configs relating to storage backends or auth mechanisms
that were provided by an app that is currently disabled.

Introduces InvalidBackend and InvalidAuth placeholders to be able to properly pass around the StorageConfig. This makes it possible to keep the mount point visible but have it be a `FailedStorage` that gracefully throws `StorageNotAvailableException` which results in 503 Storage not available for clients.

## Related Issue
Fixes https://github.com/owncloud/core/issues/26538

## Motivation and Context
See original ticket.
Basically an admin might forget to reenable storage apps for which there are existing configs.
The current code would remove the mount point completely which would cause sync clients to delete local files.

## How Has This Been Tested?
- [x] TEST: invalid backend keeps mount point
- [x] TEST: sync client behavior with invalid backend
- [x] TEST: invalid auth keeps mount point
- [x] TEST: broken config is read-only in web UI (admin + personal page)
- [x] TEST: broken config can be deleted in web UI
- [x] TEST: broken config cannot be modified with CLI commands
- [x] TEST: broken config can be deleted with CLI commands
- [x] TEST: broken config cannot be exported with CLI commands, shows warning (because storage class is missing)
- [x] TEST: reenable storage app (ex: WND), mount point available again, sync client functions normally (no redownload)

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

@owncloud/filesystem @butonic

PR #29135, opened 2017-10-02
[stable10] Bye bye Dropbox v1

DeepDiver1975:
backport of #28073


tomneedham:
What do we do with existing v1 mounts that are configured? Do they stay in the database? Should we clean these up? Maybe log a message that they have been cleaned up for $user and will need to be recreated in v2 app?

phil-davis:
Is there an algorithm that can take V1 mount settings as input and produce valid V2 settings as output?
Then it could be upgraded automagically.
Or is V2 too much different?

tomneedham:
> Is there an algorithm that can take V1 mount settings as input and produce valid V2 settings as output?

@Hemant-Mann ? 

Obviously this would require installing the v2 app - but the converted mounts could be ready and waiting if compatible?

Hemant-Mann:
@tomneedham Yes there is way
Dropbox v1 used oauth1 which had **token** and **token_secret**
According to dropbox api documentation we can use these to obtain access_token using app id and secret
[https://www.dropbox.com/developers/documentation/http/documentation#auth-token-from_oauth1](https://www.dropbox.com/developers/documentation/http/documentation#auth-token-from_oauth1)

I think We can look for old dropbox mounts and convert them to Dropbox v2 mounts

Thoughts @PVince81?

tomneedham:
Maybe just add this as something that is performed on install for the new app? Since it requires the new app code anyway - keeps it separated.

phil-davis:
Yes, old mount data will just sit in the system waiting for the new V2 app to be enabled, then "magically" get converted.

I guess the old mount points would stay visible (and not working) on the system in the transition period, so users could delete those if they wish without ever enabling the new V2 app.

PVince81:
We should still properly test this to make sure that our current APIs are still able to read the old Dropbox mount configs even when the old Dropbox mount code is gone.

PVince81:
I seem to remember seeing a comment from @hodyroff somewhere suggesting to keep this until 10.0.5 to give people enough time to migrate. Sounds like a good plan.

PVince81:
Moving to "planned" for 10.0.5

DeepDiver1975:
removing my assignment - merge once you agree

PVince81:
Before we merge this, I'd like to have https://github.com/owncloud/core/pull/26539 finished and merged as it would prevent breaking an instance whenever a storage backend is suddenly missing.

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:
Automated rebase failed! Please rebase your pull request manually via the command line.

Error:
```Command '['git', 'clone', '--depth', '1', 'rebase', '/tmp/tmpmvg6o1w4']' returned non-zero exit status 128.```

Issue #29047, opened 2017-09-18
[regression] dav properties on non-files are not stored/retrieved properly


DeepDiver1975:
## description
regression was introduced in https://github.com/owncloud/core/pull/27005

symptom: dav properties on calendars cannot be set https://github.com/owncloud/calendar/issues/812

## action
- bring back the property path value
- if no fileid can be retrieved: use the property path
- kill OC\Connector\Sabre\CustomPropertiesBackend

VicDeo:
@DeepDiver1975 the main idea  brhind this changecwas to reduce table size. 
Should we make the property path shorter than before?

DeepDiver1975:
The table should remain small if we only Save the path in cases where we dont have the fileid like for calendars

VicDeo:
@DeepDiver1975 it should be also taken into account that migration from propertypath to fileid drops all property values with unresolvable propertypath

VicDeo:
@DeepDiver1975 I need your input on the following approach:

- table `oc_properties` referenced by `userid` and `path` 
- table  `oc_file_properties` referenced by `fileid` 
- Lazy migration to `fileid` where possible while accessing the properties.

And I haven't found the class `OC\Connector\Sabre\CustomPropertiesBackend` anywhere.
Did you mean one of those:
```
apps/dav/lib/Files/CustomPropertiesBackend.php:class CustomPropertiesBackend implements BackendInterface {
apps/dav/lib/Connector/Sabre/CustomPropertiesBackend.php:class CustomPropertiesBackend implements BackendInterface {
```

DeepDiver1975:
> And I haven't found the class OC\Connector\Sabre\CustomPropertiesBackend anywhere.
> Did you mean one of those:

yes

DeepDiver1975:
> propertypath makes no sense without an owner

this is true for webdav v1. But in webdav v2 we have the user in the path
remote.php/webdav/welcome.txt vs remote.php/dav/files/viktar/welcome.txt

We might want to add a table for all properties of non files dav resources oc_dav_properties
oc_properties can remain the way they are.

VicDeo:
@DeepDiver1975 `apps/dav/lib/Connector/Sabre/CustomPropertiesBackend.php` is used in `OCA\DAV\Connector\Sabre\ServerFactory` with 'webdav' endpoint aka  webdav v1

[...]

PR #29136, opened 2017-10-02
Fix storing/retrieval for dav properties of non files

VicDeo:
## Description
- [x] integration tests for setting properties via old endpoind and getting them via new one and vice versa
- [x] Use the same CustomPropertiesBackend with old/new endpoint
- [x] Fix regression in calendar app by storing properties for non-files dav resources into a separate oc_dav_properties table
- [x] Test with old cards/cals and old endpoint. 
- :white_check_mark: Works for  files at the old endpoint. 
~mapping **old dav path** -> **new dav path** needed for oc_dav_properties.~ Not needed. 
Previous version of backend was not used by `remote.php/caldav` and `remote.php/carddav` endpoints

## Related Issue
https://github.com/owncloud/core/issues/29047

## Motivation and Context
Code deduplication + regression fix

## How Has This Been Tested?
- integration tests
- proppatching files using `tests/TestHelpers/WebDavHelper.php` as a library
- selecting/unselecting calendars in `calendar` app and checking entries in `oc_dav_properties` 

## Types of changes

## Checklist:


DeepDiver1975:
Nice! 👍

apps/dav/lib/Files/CustomPropertiesBackend.php
```diff
@@ -93,6 +97,27 @@ public function __construct(
...
+			return;
+		} catch (Forbidden $e) {
+			// might happen for excluded mount points, skip
+			return;
+		} catch (NotFound $e) {
+			// in some rare (buggy) cases the node might not be found,
+			// we catch the exception to prevent breaking the whole list with a 404
+			// (soft fail)
+			\OC::$server->getLogger()->warning(
+				'Could not get node for path: \"' . $path . '\" : ' . $e->getMessage(),
```
DeepDiver1975: please use string interpolation

DeepDiver1975 (dismissed):

apps/dav/lib/Files/CustomPropertiesBackend.php
```diff
@@ -22,16 +22,41 @@
...
+	const SELECT_BY_ID_STMT = 'SELECT * FROM `*PREFIX*properties` WHERE `fileid` = ?';
+	const INSERT_BY_ID_STMT = 'INSERT INTO `*PREFIX*properties`'
+			. ' (`fileid`,`propertyname`,`propertyvalue`) VALUES(?,?,?)';
+	const UPDATE_BY_ID_AND_NAME_STMT = 'UPDATE `*PREFIX*properties`'
+			. ' SET `propertyvalue` = ? WHERE `fileid` = ? AND `propertyname` = ?';
+	const DELETE_BY_ID_STMT = 'DELETE FROM `*PREFIX*properties` WHERE `fileid` = ?';
+	const DELETE_BY_ID_AND_NAME_STMT = 'DELETE FROM `*PREFIX*properties`'
+			. ' WHERE `fileid` = ? AND `propertyname` = ?';
+
+	const SELECT_BY_PATH_STMT = 'SELECT * FROM `*PREFIX*dav_properties` WHERE `propertypath` = ?';
```
PVince81: I don't understand: the comment for "dav_properties" says it's to store non-FS items but this class if under "Files".

Still WIP I guess ? @VicDeo

VicDeo: @PVince81 WIP, but it's quite close to completion

PVince81:
I would have preferred separate classes, one for the files handling and one for the non-files. But I understand that it might be more complicated to make Sabre use one or another for specific paths.

VicDeo:
@PVince81 refactored.

apps/dav/appinfo/Migrations/Version20170927201245.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * @author Victor Dubiniuk <dubiniuk@owncloud.com>
```
PVince81: Viktar ? 😉

apps/dav/lib/Connector/Sabre/ServerFactory.php
```diff
@@ -177,16 +179,34 @@ public function createServer($baseUri,
...
+				$miscPropertiesPlugin = new \Sabre\DAV\PropertyStorage\Plugin(
+					new MiscCustomPropertiesBackend(
+						$objectTree,
+						$this->databaseConnection,
+						$this->userSession->getUser()
 					)
 				);
+				$miscPropertiesPlugin->pathFilter = function($path) {
+					$pathInfo = $this->request->getPathInfo();
+					return preg_match('!^/webdav(/|\z)!', $pathInfo);
```
PVince81: regexp here looks exactly the same as above ? I thought this one here should be "not 'webdav'" ?

apps/dav/lib/Connector/Sabre/ServerFactory.php
```diff
@@ -177,16 +179,34 @@ public function createServer($baseUri,
...
+					new FileCustomPropertiesBackend(
+						$objectTree,
+						$this->databaseConnection,
+						$this->userSession->getUser()
+					)
+				);
+				$filePropertiesPlugin->pathFilter = function($path) {
+					$pathInfo = $this->request->getPathInfo();
+					// matches '/webdav', '/webdav/', '/webdav/adcd' but not '/webdavabcd'
+					return preg_match('!^/webdav(/|\z)!', $pathInfo);
```
PVince81: use strpos as it might be cheaper than a regexp ?

apps/dav/lib/DAV/CustomPropertiesBackendAbstract.php
```diff
@@ -19,25 +19,27 @@
...
 use Sabre\DAV\PropertyStorage\Backend\BackendInterface;
 use Sabre\DAV\PropFind;
 use Sabre\DAV\PropPatch;
 use Sabre\DAV\Tree;
+use Sabre\Dav\Exception\Forbidden;
+use Sabre\DAV\Exception\NotFound;
+use Sabre\DAV\Exception\ServiceUnavailable;
 
-class CustomPropertiesBackend implements BackendInterface {
+abstract class CustomPropertiesBackendAbstract implements BackendInterface {
```
PVince81: `AbstractCustomPropertiesBackend`

apps/dav/lib/DAV/CustomPropertiesBackendAbstract.php
```diff
@@ -86,13 +87,84 @@ public function __construct(
...
+	 * @param string $path path
+	 * @param INode $node
+	 * @param array $requestedProperties requested properties or empty array for "all"
+	 * @return array
+	 * @note The properties list is a list of propertynames the client
+	 * requested, encoded as xmlnamespace#tagName, for example:
+	 * http://www.example.org/namespace#author If the array is empty, all
+	 * properties should be returned
+	 */
+	abstract protected function getProperties($path, INode $node, array $requestedProperties);
```
PVince81: aren't these defined on the interface already ? you likely don't need to redefine them here.

Only keep the methods that are needed in the base class but aren't on the interface

apps/dav/lib/DAV/FileCustomPropertiesBackend.php
```diff
@@ -0,0 +1,193 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
```
PVince81: add yourself and myself ?

apps/dav/lib/DAV/FileCustomPropertiesBackend.php
```diff
@@ -0,0 +1,193 @@
...
+	/**
+	 * This method is called after a successful MOVE
+	 *
+	 * @param string $source
+	 * @param string $destination
+	 *
+	 * @return void
+	 */
+	public function move($source, $destination) {
+		// Part of interface. We don't care about move as soon as it doesn't affect fileId
```
PVince81: replace "as soon as it doesn't affect" with "because it doesn't affect"

apps/dav/lib/DAV/FileCustomPropertiesBackend.php
```diff
@@ -0,0 +1,193 @@
...
+		// pre-fill cache
+		foreach ($childNodes as $childNode) {
+			$childId = $childNode->getId();
+			if ($childId) {
+				$childrenIds[] = $childId;
+				$this->offsetSet($childId, []);
+			}
+		}
+
+		$sql = 'SELECT * FROM `*PREFIX*properties` WHERE `fileid` IN (?)';
```
PVince81: add `TODO: use query builder`

apps/dav/lib/DAV/MiscCustomPropertiesBackend.php
```diff
@@ -0,0 +1,162 @@
...
+	/**
+	 * Bulk load properties for children
+	 *
+	 * @param INode $node
+	 * @param array $requestedProperties requested properties
+	 *
+	 * @return void
+	 */
+	protected function loadChildrenProperties(INode $node, $requestedProperties) {
+		// Not supported
```
PVince81: was it not supported before ?

I'm also surprised to see this here as I don't see anything related to getting props of children in the original CustomPropertiesBackend. The custom loading I believe was in a different location, maybe in the FilesPlugin

apps/dav/lib/Server.php
```diff
@@ -175,15 +179,30 @@ public function __construct(IRequest $request, $baseUri) {
...
-						)
+				$filePropertiesPlugin = new \Sabre\DAV\PropertyStorage\Plugin(
+					new FileCustomPropertiesBackend(
+						$this->server->tree,
+						\OC::$server->getDatabaseConnection(),
+						\OC::$server->getUserSession()->getUser()
+					)
+				);
+				$filePropertiesPlugin->pathFilter = function($path) {
+					return strpos($path, 'files/') === 0;
```
PVince81: this would allow this properties plugin to operate on the path "files/$userId/" which might be alright, meaning we could set custom properties on the user's storage root. I suggest to add a comment to clarify that this is intended

PVince81 (commented):
Great progress, I like the separation of classes.

See comments above

apps/dav/lib/DAV/MiscCustomPropertiesBackend.php
VicDeo: @PVince81 there was a check `if ($node instance of Directory)` before scanning children. So it was supported for files only.

PVince81: I don't see this in the original CustomPropertiesBackend: https://github.com/owncloud/core/blob/v10.0.3/apps/dav/lib/Files/CustomPropertiesBackend.php.

Are you sure it isn't something you added through your previous changes in this PR ?
Or maybe saw this in another class ?

VicDeo: @PVince81 https://github.com/owncloud/core/blob/v10.0.3/apps/dav/lib/Connector/Sabre/CustomPropertiesBackend.php#L135

PVince81: What ?! We had two CustomPropertiesBackend on master/stable10 ? So why do we need to split it again here ? Which one was used ?

VicDeo: Exactly. One for an old dav endpoint  and another for a new one. This PR kills backend that was used for an old endpoint, see checkmarks in description.
New endpoint already has user inside the property path so the only thing left TODO here is to remap old path to new path when the property is stored by path.

PVince81: Ahhhh, got it. Thanks for clarifying

apps/dav/lib/DAV/CustomPropertiesBackendAbstract.php
VicDeo: @PVince81 nope, Interface has high-level methods only, and this is a  horse working under the hood :)
Interface declares `propFind`, `propPatch`, `move` and `delete`

apps/dav/lib/Connector/Sabre/ServerFactory.php
VicDeo: yep, my bad

apps/dav/lib/Connector/Sabre/ServerFactory.php
VicDeo: @PVince81 and it is 4 lines long in `remote.php` :)

apps/dav/lib/DAV/FileCustomPropertiesBackend.php
VicDeo: added

VicDeo:
@DeepDiver1975 Do old endpoints `remote.php/caldav/` and `remote.php/carddav/` support custom properties in general?
From what I see in `apps/dav/appinfo/v1/carddav.php` and `apps/dav/appinfo/v1/caldav.php` they fire instance of `\Sabre\DAV\Server` and do not register CustomPropertiesBackend.

apps/dav/lib/Connector/Sabre/ServerFactory.php
```diff
@@ -177,16 +179,34 @@ public function createServer($baseUri,
...
+					)
+				);
+				$filePropertiesPlugin->pathFilter = function($path) {
+					$pathInfo = $this->request->getPathInfo();
+					// matches '/webdav', '/webdav/', '/webdav/adcd' but not '/webdavabcd'
+					return (preg_match('!^/webdav(/|\z)!', $pathInfo) === 1);
+				};
+				$server->addPlugin($filePropertiesPlugin);
+
+				$miscPropertiesPlugin = new \Sabre\DAV\PropertyStorage\Plugin(
```
VicDeo: Possibly not needed as this class is used to build up server for file endpoint only.

apps/dav/lib/Connector/Sabre/ServerFactory.php
```diff
@@ -177,16 +179,34 @@ public function createServer($baseUri,
...
-							$this->userSession->getUser()
-						)
+				$filePropertiesPlugin = new \Sabre\DAV\PropertyStorage\Plugin(
+					new FileCustomPropertiesBackend(
+						$objectTree,
+						$this->databaseConnection,
+						$this->userSession->getUser()
+					)
+				);
+				$filePropertiesPlugin->pathFilter = function($path) {
```
VicDeo: filtering makes no sense  in case this class is used to build up server for file endpoint only.

DeepDiver1975:
> As I see in apps/dav/appinfo/v1/carddav.php and apps/dav/appinfo/v1/caldav.php they fire instance of \Sabre\DAV\Server and do not register CustomPropertiesBackend.

afaik this was possible as well .... but this is not really relevant. the owncloud calendar app uses the new endpoints and there we need the properties.

apps/dav/lib/DAV/AbstractCustomPropertiesBackend.php
```diff
@@ -86,13 +87,84 @@ public function __construct(
...
+	 * @param string $path path
+	 * @param INode $node
+	 * @param array $requestedProperties requested properties or empty array for "all"
+	 * @return array
+	 * @note The properties list is a list of propertynames the client
+	 * requested, encoded as xmlnamespace#tagName, for example:
+	 * http://www.example.org/namespace#author If the array is empty, all
+	 * properties should be returned
+	 */
+	abstract protected function getProperties($path, INode $node, array $requestedProperties);
```
PVince81: please remove these, they are already defined on the interface and should not require redefining here

apps/dav/lib/DAV/FileCustomPropertiesBackend.php
```diff
@@ -0,0 +1,196 @@
...
+ *
+ */
+
+namespace OCA\DAV\DAV;
+
+use Doctrine\DBAL\Connection;
+use OCA\DAV\Connector\Sabre\Directory;
+use Sabre\DAV\INode;
+
+class FileCustomPropertiesBackend extends AbstractCustomPropertiesBackend {
```
PVince81: PHPDoc, what is this class ?

apps/dav/lib/DAV/MiscCustomPropertiesBackend.php
```diff
@@ -0,0 +1,162 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\DAV;
+
+use Doctrine\DBAL\Connection;
+use Sabre\DAV\INode;
+
+class MiscCustomPropertiesBackend extends AbstractCustomPropertiesBackend {
```
PVince81: PHPDoc, what is this class ?

PVince81 (commented):
Code looks good otherwise, please remove the redefined abstract methods.

If you fully tested this and all works then after the above change this is good to go from my POV.

apps/dav/lib/DAV/AbstractCustomPropertiesBackend.php
VicDeo: @PVince81 it's not https://github.com/sabre-io/dav/blob/master/lib/DAV/PropertyStorage/Backend/BackendInterface.php

PVince81: Fair enough :+1:

VicDeo:
@PVince81 class descriptions added

PVince81:
uh oh... codecov not happy. Did you add unit tests ?

VicDeo:
@PVince81 `codecov/patch — 76.7% of diff hit (target 59.43%)` :wink:

PVince81 (approved):
:+1:

PVince81:
@VicDeo please backport to stable10

VicDeo:
Stable10: https://github.com/owncloud/core/pull/29273

Issue #28285, opened 2017-07-03
Undefined index: share_name at \/var\/www\/html\/owncloud\/lib\/private\/Share20\/DefaultShareProvider.php#863


dc75la:
### Steps to reproduce
1. php occ files:scan --all
OR
1. check owncloud.log

### Expected behaviour
There are a lot of errors in the log and the occ filescan doesn't work. A lot of errors. I think for some reason the share_name index isn't in the $data variable in the DefaultShareProvider.php (check line 863)

### Actual behaviour
Tell us what happens instead

### Server configuration
**Operating system**:
Cent OS 7 (7.3.1611)

**Web server:**
Apache 2.4.6

**Database:**
mysql 14.14 Distrib 5.6.36

**PHP version:**
php 5.6.30

**ownCloud version:** (see ownCloud admin page)
10.0.2 (stable)

**Updated from an older ownCloud or fresh install:**
from 9.1.2 (stable) (9.1.6 after update I think) to 10.0.2

**Where did you install ownCloud from:**
http://download.owncloud.org/download/repositories/10.0.2/CentOS_7/ce:10.0.repo

**Signing status (ownCloud 9.0 and above):**
"https://gist.github.com/dc75la/a23c6b866adce8ee1c45bc17c92ceb78"
(because of a bug with column length: 256 set to 255)

**The content of config/config.php:**
$CONFIG = array (
  'instanceid' => '--scrambled--',
  'passwordsalt' => '--scrambled--',
  'trusted_domains' => 
  array (
    0 => '--scrambled--',
    1 => '--scrambled--',
  ),
  'datadirectory' => '/home/--scrambled--',
  'overwrite.cli.url' => 'https://--scrambled--',
  'dbtype' => 'mysql',
  'version' => '10.0.2.1',
  'installed' => true,
  'mail_smtpmode' => 'sendmail',
  'mail_smtphost' => '--scrambled--',
  'mail_smtpport' => '25',
  'mail_from_address' => '--scrambled--',
  'mail_domain' => '--scrambled--',
  'loglevel' => '0',
  'mail_smtpauthtype' => '--scrambled--',
  'maintenance' => false,
  'secret' => '--scrambled--',
  'theme' => '',
  'updatechecker' => false,
  'trashbin_retention_obligation' => 'auto, auto',
  'dbname' => '--scrambled--',
  'dbhost' => 'localhost:3306',
  'dbuser' => '--scrambled--',
  'dbpassword' => '--scrambled--',
[...]

PR #29142, opened 2017-10-04
Force alphabetical ordering of app installation to fix on osx 10.13

tomneedham:
## Description
Well this was an interesting one to find. PHP `readdir` returns in order that the filesystem stores a directory - which is _not_ necessarily alphabetical ordering. However, apparently we require this ordering, since apps with names `files_` typically rely on `files` to already be installed.

Since updating to OSX High Sierra 10.13, they now use a new 'optimised filesystem' which appears to affect the listing order. This caused the install to fail because trashbin was installed before files and there was an autoloading exception. 

The fix forces the installer to use an alphabetical ordering when listing the contents of app folders.

## Motivation and Context
Install of core fails since OSX High Sierra upgrade.

## How Has This Been Tested?
UI and OCC installs on OSX High Sierra.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Installer.php
+++ b/lib/private/Installer.php
@@ -457,7 +457,8 @@ class Installer {
 		$errors = [];
 		foreach(\OC::$APPSROOTS as $app_dir) {
 			if($dir = opendir( $app_dir['path'] )) {
-				while( false !== ( $filename = readdir( $dir ))) {
+				$nodes = scandir($dir);
+				foreach($nodes as $filename) {
 					if( substr( $filename, 0, 1 ) != '.' and is_dir($app_dir['path']."/$filename") ) {
 						if( file_exists( $app_dir['path']."/$filename/appinfo/info.xml" )) {
 							if(!Installer::isInstalled($filename)) {
--- a/lib/private/Installer.php
+++ b/lib/private/Installer.php
@@ -38,9 +38,8 @@
 
 namespace OC;
 
-use OC\App\CodeChecker\CodeChecker;
-use OC\App\CodeChecker\EmptyCheck;
-use OC\App\CodeChecker\PrivateCheck;
+use Doctrine\DBAL\Exception\TableExistsException;
+use OC\DB\MigrationService;
 use OC_App;
 use OC_DB;
 use OC_Helper;
@@ -414,8 +413,9 @@ class Installer {
 
 	/**
 	 * Removes an app
-	 * @param string $name name of the application to remove
+	 * @param string $appId name of the application to remove
 	 * @return boolean
+	 * @throws AppAlreadyInstalledException
 	 *
 	 *
 	 * This function works as follows
@@ -445,6 +445,37 @@ class Installer {
 		return false;
 	}
 
+	protected static function getShippedApps() {
+		$shippedApps = [];
+		foreach(\OC::$APPSROOTS as $app_dir) {
+			if($dir = opendir( $app_dir['path'] )) {
+				$nodes = scandir($app_dir['path']);
+				foreach($nodes as $filename) {
+					if( substr( $filename, 0, 1 ) != '.' and is_dir($app_dir['path']."/$filename") ) {
+						if( file_exists( $app_dir['path']."/$filename/appinfo/info.xml" )) {
+							if(!Installer::isInstalled($filename)) {
+								$info=OC_App::getAppInfo($filename);
+								$enabled = isset($info['default_enable']);
+								if (($enabled || in_array($filename, \OC::$server->getAppManager()->getAlwaysEnabledApps()))
+									&& \OC::$server->getConfig()->getAppValue($filename, 'enabled') !== 'no') {
+									$shippedApps[] = $filename;
+								}
+							}
+						}
+					}
+				}
+				closedir( $dir );
+			}
+		}
+
+
+		// Fix the order - make files first
+		$shippedApps = array_diff($shippedApps,['files', 'dav']);
+		array_unshift($shippedApps, 'dav');
+		array_unshift($shippedApps, 'files');
+		return $shippedApps;
+	}
+
 	/**
 	 * Installs shipped apps
 	 *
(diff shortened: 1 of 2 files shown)


tomneedham:
Now what did I ~break~ find on Jenkins

lib/private/Installer.php
```diff
@@ -457,7 +457,8 @@ public static function installShippedApps($softErrors = false) {
 		$errors = [];
 		foreach(\OC::$APPSROOTS as $app_dir) {
 			if($dir = opendir( $app_dir['path'] )) {
-				while( false !== ( $filename = readdir( $dir ))) {
+				$nodes = scandir($dir);
+				foreach($nodes as $filename) {
 					if( substr( $filename, 0, 1 ) != '.' and is_dir($app_dir['path']."/$filename") ) {
```
phil-davis: This is also moderately dodgy. There is no check for the parent dir entry ``..`` - I think it will pass this "if" test.
The test on the next line will (thankfully) be false, because some file like ``apps/../appinfo/info.xml`` is not going to exist.
But not directly relevant to the reason for this PR.

tomneedham: Good spot, i saw `..` come up in the loop when using `readdir` - will add an explicit check here.

tomneedham:
just grepping for other areas this may effect

lib/private/Installer.php
tomneedham: `..` is caught by the first condition, since it is checking only the first character of the filename

tomneedham:
`oc_properties` is missing when the core tests are run, because `dav` app is not installed.

```diff
diff --git a/tests/enable_all.php b/tests/enable_all.php
index 11b983eb78..67de8c04ff 100644
--- a/tests/enable_all.php
+++ b/tests/enable_all.php
@@ -16,6 +16,7 @@ function enableApp($app) {
        }
 }

+enableApp('dav');
 enableApp('files_sharing');
 enableApp('files_trashbin');
 enableApp('encryption');
```

fixes - but why is this now a problem

lib/private/Installer.php
phil-davis: Oh yes! I skimmed over the detail too quickly. And I guess in this context we do not care if there is a folder ".hidden-stuff" that is skipped over - that should not have an app in it.

tomneedham: Exactly

DeepDiver1975:
> fixes - but why is this now a problem

because the load order of apps has changed? we might simply got lucky in the past because dav sorts up pretty hight .... no idea

tomneedham:
> we might simply got lucky

my brain was avoiding this theory 😒

phil-davis:
On my Ubuntu 16.04 with PHP 7.0.22
```
<?php
$dir = opendir('core/apps');
while( false !== ( $filename = readdir( $dir ))) {
  echo $filename . "\n";
}
```
the output is:
```
files_trashbin
files
updatenotification
federation
dav
files_sharing
..
files_versions
systemtags
comments
federatedfilesharing
testing
provisioning_api
files_texteditor
files_external
.
firewall
theme-example
```
so the old way is "random order" on *nix also.

tomneedham:
Just be lucky that your `files` app comes first ;)

PVince81:
Could this explain the wrong migration orders in some setups or is this unrelated ?

Ref https://github.com/owncloud/core/issues/28285

tomneedham:
> Could this explain the wrong migration orders in some setups or is this unrelated ?

I thought so - but they are explicitly sorted by their numbers before being applied.

But could still be related - as apps are installing in different orders which could cause weird effects when they are trying to apply schema changes based on what is present

PVince81:
@tomneedham your attempt at making the universe deterministic has failed:
```
14:49:19 There were 10 errors:
14:49:19 
14:49:19 1) Test\DB\QueryBuilder\QueryBuilderTest::testGetLastInsertId
14:49:19 Doctrine\DBAL\Exception\TableNotFoundException: An exception occurred while executing 'INSERT INTO "oc_properties" ("fileid", "propertyname", "propertyvalue") VALUES('100500', 'testing', 'testing')':
14:49:19 
14:49:19 SQLSTATE[HY000]: General error: 1 no such table: oc_properties
14:49:19 
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php:58
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:128
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:1015
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/lib/private/DB/Connection.php:211
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Query/QueryBuilder.php:208
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/lib/private/DB/QueryBuilder/QueryBuilder.php:141
14:49:19 /var/lib/jenkins/workspace/owncloud-core_core_PR-29142-JO7HXKWBQT42X32D2PGSRR4QVWOAPQRCZP75FRKHSWHVTLPQOWIQ/tests/lib/DB/QueryBuilder/QueryBuilderTest.php:1133
14:49:19 
14:49:19 Caused by
14:49:19 Doctrine\DBAL\Driver\PDOException: SQLSTATE[HY000]: General error: 1 no such table: oc_properties
14:49:19 
```

tomneedham:
Explicitly setting files, dav as the first two apps seems to work - otherwise somehow dav gets installed without the migrations running 😟

PVince81 (approved):
:+1:, assuming there is no adjustable unit test due to staticness.

Not happy about the hard-coded bits but there is no better way. At least app loading will be deterministic.

tomneedham:
> Not happy about the hard-coded bits but there is no better way

Likewise - but at least it is stable

PVince81:
@tom please backport to stable10

Issue #26936, opened 2017-01-13
Alternative key storage location doesn't delete all keys after deleting user


PVince81:
### Steps

1. Setup OC v9.1.3
1. `occ app:enable encryption`.
1. `occ encryption:enable`
1. `mkdir data/enckeys; chown wwwrun:www data/enckeys` (adjust for permissions accordingly)
1. `occ encryption:change-key-storage-root enckeys`
1. Login as admin
1. Create a user "user1"
1. Login as "user1"
1. Check "data/enckeys/user1/" and subdirs: keys are present
1. Login as admin
1. Delete "user1" (need to wait 7s or navigate away)
1. Check "data/enckeys/user1/" and subdirs

### Expected result
The folder "data/enckeys/user1" must be gone just like "data/user1".

### Actual result
"data/user1" is gone but the keys folder "data/enckeys/user1" still exist with contents.
Only the user's publicKey was deleted.

### Versions
Observed in ownCloud 9.1.3.
I suspect that this never worked properly.

I don't think this is critical, @Peter-Prochaska correct me if you think otherwise.

DeepDiver1975:
Fixed with #29144

DeepDiver1975:
@PVince81 shall we backport this?

PVince81:
@DeepDiver1975 ok for backport

PR #29144, opened 2017-10-04
Remove alternate keys storage during user delete

sharidas:
When user is deleted the alternate key storage location
does have folder with username. So while deleting user
this should also be cleaned up.

## Description

Remove alternate key storage location during user delete process.

## Related Issue

https://github.com/owncloud/core/issues/26936

## Motivation and Context

When user is deleted, the alternate key storage location which has folder name with username should
also be deleted.

## How Has This Been Tested?

- [x] enable encryption with user keys
- [x] Using change-key-storage-root command move the keys to folder `enckeys`
- [x] Create another user `user1` apart from user `admin`
- [x] Login as `user1`
- [x] Logout from `user1` and login as `admin`
- [x] Delete the `user1`, there shouldn't be any traces of `user1` inside the folder `enckeys`

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Encryption/Keys/Storage.php
+++ b/lib/private/Encryption/Keys/Storage.php
@@ -182,6 +182,27 @@ class Storage implements IStorage {
 	}
 
 	/**
+	 * @inheritdoc
+	 */
+
+	public function deleteAltUserStorageKeys($uid) {
+		if (\OC::$server->getEncryptionManager()->isEnabled()) {
+			/**
+			 * If the key storage is not the default
+			 * location, then we need to remove the keys
+			 * in the alternate key location
+			 */
+			$keyStorageRoot = $this->util->getKeyStorageRoot();
+			if ($keyStorageRoot !== '') {
+				$this->view->rmdir($keyStorageRoot . '/' . $uid);
+				return true;
+			}
+
+			return false;
+		}
+	}
+
+	/**
 	 * construct path to users key
 	 *
 	 * @param string $encryptionModuleId
--- a/lib/public/Encryption/Keys/IStorage.php
+++ b/lib/public/Encryption/Keys/IStorage.php
@@ -169,4 +169,14 @@ interface IStorage {
 	 */
 	public function copyKeys($source, $target);
 
+	/**
+	 * delete user keys from alternate storage location when
+	 * a user is deleted
+	 *
+	 * @param $uid
+	 * @return boolean
+	 * @since 10.0.4
+	 */
+	public function deleteAltUserStorageKeys($uid);
+
 }


lib/private/User/User.php
```diff
@@ -210,6 +211,16 @@ public function delete() {
 		// Delete the users entry in the storage table
 		Storage::remove('home::' . $this->getUID());
 
+		// Remove alternative encryption key storage location for the user
+		if ($this->config->getAppValue('core', 'encryption_enabled', 'no') === 'yes') {
```
DeepDiver1975: Can't we do this in a event? This belongs to encryption - this should not live in the user class

DeepDiver1975 (changes requested):

sharidas:
Updated the PR with the help of event dispatcher. And hence there is another bit to be reviewed which is associated with this PR: https://github.com/owncloud/encryption/pull/16

lib/private/User/User.php
```diff
@@ -210,6 +211,10 @@ public function delete() {
 		// Delete the users entry in the storage table
 		Storage::remove('home::' . $this->getUID());
 
+		// Remove alternative encryption key storage location for the user
+		$event = new GenericEvent(null, ['uid' => $this->getUID()]);
+		$this->eventDispatcher->dispatch('OC\Encryption\Keys\Storage::deleteAltUserStorageKeys', $event);
```
DeepDiver1975: no - this way we have a specific event in here. Please use a generic event name which allows any code to integrate in here.

DeepDiver1975 (changes requested):

lib/private/User/User.php
karakayasemi: Why do not we use just preDelete hooks in here? If we are planning to remove hooks, imho we should use more generic event like oc.user.postdelete. 
Also, symfony has some naming conventions on the EventDispatcher, It may be beneficial for us to adhere to these rules:
-Use only lowercase letters, numbers, dots (.) and underscores (_);
-Prefix names with a namespace followed by a dot (e.g. order., user.*);
-End names with a verb that indicates what action has been taken (e.g. order.placed).
https://symfony.com/doc/current/components/event_dispatcher.html#naming-conventions

sharidas:
Got another idea, why not call https://github.com/owncloud/encryption/pull/16/files#diff-ef2f85587bb2300b22909edff73c7440L198. We already have post delete user hook in the encryption app. Let me know if this looks ok?

DeepDiver1975:
> Got another idea, why not call https://github.com/owncloud/encryption/pull/16/files#diff-ef2f85587bb2300b22909edff73c7440L198. We already have post delete user hook in the encryption app. Let me know if this looks ok?

nice - thats the way to go - we can close this pr then - right?

sharidas:
Yes we can merge this change.

DeepDiver1975 (approved):

sharidas:
Created backport : https://github.com/owncloud/core/pull/29155

Issue #28479, opened 2017-07-25
Duplicated mount configuration after going back during oauth2 flow


jvillafanez:
### Steps to reproduce
1. Go to the admin settings and configure a gdrive mount
2. Setup the client id and secret, and click on "grant access"
3. During the oauth2 flow in the browser, go back with the browser navigation

### Expected behaviour
Admin settings page should show without the previous mount being configured. Showing a wrong state might be acceptable as long as access can be granted later.

### Actual behaviour
Duplicated configuration of the previous mount is shown.

### Server configuration
**Operating system**: ubuntu 16.04

**Web server:** apache 2.4.7

**Database:** mysql

**PHP version:** 7.0

**ownCloud version:** stable10 (10.0.2 at July, 25th)

**Updated from an older ownCloud or fresh install:** fresh install

**Browser:** firefox 50.1.0

jvillafanez:
This happened twice in a row after a fresh installation, both in admin and personal settings, but I couldn't reproduce it afterwards following the same steps.

PVince81:
Yeah I remember seeing this before too, thanks for reporting

davitol:
![screen shot 2017-09-14 at 13 43 22](https://user-images.githubusercontent.com/10596560/30428740-0127aafa-9955-11e7-9ba6-5bcbaee2d099.png)

Also reproduced with oC 10.0.3.RC1 version. Something to be fixed in the future

PVince81:
setting to p2-high and "planned", this seems to happen more often than I thought

PVince81:
@jvillafanez can you have a look ?

jvillafanez:
Looks like the click-back messes might mess with the "natural" oauth1/2 flow and it's causing this issue. Still under investigation.

jvillafanez:
<img width="961" alt="screen shot 2017-10-04 at 11 11 15" src="https://user-images.githubusercontent.com/1477829/31168226-e3bb96d4-a8f4-11e7-8129-4f196637c0cc.png">

Looks like the duplication might be "valid" in the sense that those mounts are saved in the DB.

Reproduced by crazy-clicking the "grant access" button over a new mount point. Multiple queries are sent to the server to save that mount point which causes the duplication.

jvillafanez:
Looks like this has been broken from the beginning since it's always sending duplicate requests just clicking once in the button. The fact that usually only one storage appears is likely because the rest of the duplicate requests don't trigger or reach the server before the browser changes the page.

[...]

PR #29145, opened 2017-10-04
Fix duplicated storage mounts in the settings view during oAuth2 flow

jvillafanez:
## Description
Fix duplicated mount that could happen during oAuth2 flow. This has been partially tested with google drive but should fix other oAuth2 storages. Note that oAuth1 storages will likely present the same problem, which isn't fixed by this PR.

## Related Issue
https://github.com/owncloud/core/issues/28479

## Motivation and Context
Duplicated mounts shouldn't be saved

## How Has This Been Tested?
Manual testing

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_external/js/oauth2.js
+++ b/apps/files_external/js/oauth2.js
@@ -34,7 +34,7 @@ $(document).ready(function() {
 						&& typeof client_secret === "string"
 						&& client_secret !== ''
 					) {
-						$('.configuration').trigger('oauth_step2', [{
+						$tr.find('.configuration').trigger('oauth_step2', [{
 							backend_id: $tr.attr('class'),
 							client_id: client_id,
 							client_secret: client_secret,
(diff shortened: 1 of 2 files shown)


jvillafanez:
As a side note, the "grant access" button will be disabled to prevent further clicks

apps/files_external/js/oauth2.js
```diff
@@ -34,7 +34,7 @@ $(document).ready(function() {
 						&& typeof client_secret === "string"
 						&& client_secret !== ''
 					) {
-						$('.configuration').trigger('oauth_step2', [{
+						$tr.find('.configuration').trigger('oauth_step2', [{
 							backend_id: $tr.attr('class'),
```
Hemant-Mann: Oh okay so you are finding the configuration box closest to the current row to prevent any conflicts right?

Hemant-Mann: Button can trigger step 1 but not step 2 which is fired after OAuth completion

jvillafanez: > Button can trigger step 1 but not step 2 which is fired after OAuth completion

Any reason for that? It's technically possible. Whether it's a good or bad idea, that's another topic to discuss.

Hemant-Mann: This is how it was built
Originally instead of firing events it was authenticating only Google Drive as part of OAuth2 and there was no way for other apps to know when OAuth was completed so instead I triggered **oauth2** events which respective apps could listen

PVince81: @jvillafanez I think this would let an app's code decide whether they use a button or anything else for triggering the oauth flow.

PVince81 (approved):
:+1:

PVince81:

jvillafanez:
Backport to stable10: https://github.com/owncloud/core/pull/29254

davitol:
After this fix, i've seen duplicated WND mounts when configuring it. Maybe a race condition... will look at it a bit more, due to it happens once in a blue moon.

PR #29107, opened 2017-09-27
Central membership and group table 

mrow4a:
## Related Issue
fixes #27623 

## Plan
- [x] design new groups table
- [x] implement group entity and entity mapper
- [x] integrate membership manager and mappers https://github.com/owncloud/core/pull/29154
- [x] see jenkins expode :boom: https://github.com/owncloud/core/pull/29282
- [x] iterate over all failing unit tests https://github.com/owncloud/core/pull/29282
- [x] keep a close eye on the unit test execution time https://github.com/owncloud/core/pull/29282
- [x] implement sync command https://github.com/owncloud/core/pull/29752
- [x] design indexes based on used queries
- [ ] migration of local groups
- [ ] migration of ldap groups
- [ ] migration of saml groups
- [ ] migration or ..... groups
- [x] ldap integration - not needed works out of the box
- [x] guests integration - works out of the box, but refactor might be good https://github.com/owncloud/guests/pull/200
- [x] customgroups integration - needs adjustment to keep in sync https://github.com/owncloud/customgroups/pull/138

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
  - currently new methods are added to OCP\GroupInterface 

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

**How to test it:**

```
https://github.com/owncloud/core/wiki/How-to-test-user_ldap
```

**Main feature:**

```
core/Migrations/Version20170927145820.php

lib/private/Group/BackendGroup.php
lib/public/AppFramework/Db/Mapper.php
lib/private/Group/GroupMapper.php
lib/private/MembershipManager.php

lib/private/Group/Group.php

lib/public/IGroupManager.php
lib/private/Group/Manager.php
lib/private/SubAdmin.php
```

**Main sync feature:**

```
lib/private/Group/SyncService.php
core/Command/Group/SyncBackend.php
core/register_command.php
```

**Main changes:**

```
lib/private/Group/Database.php
lib/private/User/Account.php
lib/private/User/AccountMapper.php
lib/private/User/Manager.php
lib/private/User/User.php
```

**Required changes:**

```
lib/base.php
lib/private/Server.php
[...]

PR #29152, opened 2017-10-05
[Central group table] Split Mapper into two, more specialized classes

mrow4a:
The idea is to reduce very specialized class - Mapper - into two, smaller classes. This should make it is easier to understand and maintain. 

I will also need to have Access class for https://github.com/owncloud/core/pull/29107. The problem there is, that MembershipManager is not really a mapper, but yet still needs some of the Mapper class function. Thus split will be beneficial for that also.

### DIFF
--- a/lib/public/AppFramework/Db/Access.php
+++ b/lib/public/AppFramework/Db/Access.php
@@ -1,0 +1,142 @@
+<?php
+/**
+ * @author Piotr Mrowczynski <piotr@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCP\AppFramework\Db;
+
+use OCP\IDBConnection;
+use OCP\IDb;
+
+/**
+ * Simple class which abstracts query execution from inheriting class.
+ *
+ *
+ * @since 10.0.0
+ */
+class Access {
+
+	/**
+	 * Provide protected access to IDBConnection
+	 *
+	 * @var IDBConnection
+	 */
+	protected $db;
+
+	/**
+	 * @param IDBConnection $db Instance of the Db abstraction layer
+	 * mapped to queries without using sql
+	 * @since 10.0.0
+	 */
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
Hmmm .... would class Access be of any value on it's own?

mrow4a:
Please look in the main description.

mrow4a:
@jvillafanez @DeepDiver1975 can you have a look now?

lib/public/AppFramework/Db/Access.php
```diff
@@ -0,0 +1,142 @@
...
+	 * Runs an sql query
+	 *
+	 * @param string $sql the prepare string
+	 * @param array $params the params which should replace the ? in the sql query
+	 * @param int $limit the maximum number of rows
+	 * @param int $offset from which row we want to start
+	 * @return \PDOStatement the database query result
+	 * @since 10.0.0
+	 */
+	protected function execute($sql, array $params=[], $limit=null, $offset=null) {
```
mrow4a: Access class abstracts from inheriter a way to execute sql query

lib/public/AppFramework/Db/Mapper.php
```diff
@@ -121,7 +123,7 @@ public function insert(Entity $entity){
 		}
 
 		$sql = 'INSERT INTO `' . $this->tableName . '`(' .
-				$columns . ') VALUES(' . $values . ')';
+			$columns . ') VALUES(' . $values . ')';
 
 		$stmt = $this->execute($sql, $params);
```
mrow4a: Mapper constructs the sql query, and uses Access to physically execute it.

jvillafanez:
While I understand the purpose of the PR (separate the responsabilities: use "Access" for the connection, and the mapper for the ORM-like stuff), I don't think inheritance is the way to go. Why not inject the "Access" class in the constructor and use it as dependency?

We also need to recheck what is the responsability of the "Access" class in case some other class has already that responsability. What does "Access" do differently compared with "IDBConnection"? If it's just to simplify the calls, could the "Mapper" class absorb that responsability?

mrow4a:
The core thing of Access class is `protected function execute`. Any class, which wants to correctly execute an SQL query, with parameters, limit and offset should inherit from this class. No injection needed.

jvillafanez:
Quoting from https://en.wikipedia.org/wiki/Composition_over_inheritance
> To favor composition over inheritance is a design principle that gives the design higher flexibility. It is more natural to build business-domain classes out of various components than trying to find commonality between them and creating a family tree. For example, a gas pedal and a wheel share very few common traits, yet are both vital components in a car. What they can do and how they can be used to benefit the car is easily defined. Composition also provides a more stable business domain in the long term as it is less prone to the quirks of the family members. In other words, it is better to compose what an object can do (HAS-A) than extend what it is (IS-A).[1]

Unless there is a clear reason to use inheritance, I'd prefer to inject the class instead of inherit from it. Reusing code isn't a valid option to choose inheritance when we have better alternatives

mrow4a:
Ok, @jvillafanez I can see one problem, unfortunately, it is public interface @DeepDiver1975, and other classes are using $db db connection directly from there. I think we need to use inheritance, since injection will break public interface, which unfortunately class Mapper is (we cannot change constructor)

mrow4a:
Ok, I found a better way. I just need to use entirely query builder, which allows me to use limit and offset

PR #29154, opened 2017-10-05
[Central group table] Add membership manager to central group table 

mrow4a:
This is first approach to handle central group table feature. It utilizes new class Membership manager, which inherits from Access class. More implementation details will go inside the code. 

NOTE: Mind that this PR target is `centr_groups` branch, from this PR https://github.com/owncloud/core/pull/29107

@patrickjahns

### DIFF
--- a/lib/private/MembershipManager.php
+++ b/lib/private/MembershipManager.php
@@ -1,0 +1,746 @@
+<?php
+/**
+ * @author Piotr Mrowczynski <piotr@owncloud.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC;
+
+use OC\Group\BackendGroup;
+use OC\User\Account;
+use OCP\AppFramework\Db\Entity;
+use OCP\IConfig;
+use OCP\IDBConnection;
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+
+class MembershipManager {
+
+	/**
+	 * types of memberships in the group
+	 */
+	const MEMBERSHIP_TYPE_GROUP_USER = 0;
+	const MEMBERSHIP_TYPE_GROUP_ADMIN = 1;
+
+	/** @var IConfig */
+	protected $config;
+
+	/** @var IDBConnection */
+	private $db;
+
+	/**
+	 * MembershipManager class is responsible for mapping and handling
+	 * user/group memberships
+	 *
+	 * @param IDBConnection $db
+	 * @param IConfig $config
+	 */
+	public function __construct(IDBConnection $db, IConfig $config) {
+		$this->db = $db;
+		$this->config = $config;
+	}
+
+	/**
+	 * Return backend group entities for given account (identified by user's uid)
+	 *
+	 * @param string $userId
+	 *
+	 * @return BackendGroup[]
+	 */
+	public function getUserBackendGroups($userId) {
+		return $this->getBackendGroupsSqlQuery($userId, false, self::MEMBERSHIP_TYPE_GROUP_USER);
+	}
+
+	/**
+	 * Return backend group entities for given account (identified by user's internal account id)
+	 *
+	 * NOTE: Search by internal id is used to optimize access when
+	 * group backend/account has already been instantiated and internal id is explicitly available
+	 *
+	 * @param int $accountId
+	 *
+	 * @return BackendGroup[]
+	 */
+	public function getUserBackendGroupsById($accountId) {
+		return $this->getBackendGroupsSqlQuery($accountId, true, self::MEMBERSHIP_TYPE_GROUP_USER);
+	}
+
+	/**
+	 * Return backend group entities for given account (identified by user's uid) of which
+	 * the user is admin.
+	 *
+	 * @param string $userId
+	 *
+	 * @return BackendGroup[]
+	 */
+	public function getAdminBackendGroups($userId) {
+		return $this->getBackendGroupsSqlQuery($userId, false, self::MEMBERSHIP_TYPE_GROUP_ADMIN);
+	}
+
+	/**
+	 * Return user account entities for given group (identified with gid). If group predicate not specified,
+	 * it will return all users which are group users
+	 *
+	 * @param string|null $gid
+	 *
+	 * @return Account[]
+	 */
+	public function getGroupUserAccounts($gid = null) {
+		return $this->getAccountsSqlQuery($gid, false, self::MEMBERSHIP_TYPE_GROUP_USER);
+	}
+
+	/**
+	 * Return user account entities for given group (identified with group's internal backend group id)
+	 *
+	 * @param int $backendGroupId
+	 *
+	 * @return Account[]
+	 */
+	public function getGroupUserAccountsById($backendGroupId) {
+		return $this->getAccountsSqlQuery($backendGroupId, true, self::MEMBERSHIP_TYPE_GROUP_USER);
+	}
+
+	/**
+	 * Return admin account entities for given group (identified with gid). If group predicate not specified,
+	 * it will return all users which are group admins
+	 *
+	 * @param string|null $gid
+	 *
+	 * @return Account[]
+	 */
+	public function getGroupAdminAccounts($gid = null) {
+		return $this->getAccountsSqlQuery($gid, false, self::MEMBERSHIP_TYPE_GROUP_ADMIN);
+
+	}
+
+	/**
+	 * Check whether given user (identified by user's uid) is user of
+	 * the group (identified with group's gid). If group predicate not specified,
+	 * it will check if user is group user of any group
+	 *
+	 * @param string $userId
+	 * @param string $gid
+	 *
+	 * @return boolean
+	 */
+	public function isGroupUser($userId, $gid = null) {
+		return $this->isGroupMemberSqlQuery($userId, $gid, self::MEMBERSHIP_TYPE_GROUP_USER, false);
+	}
+
+	/**
+	 * Check whether given account (identified by user's internal account id) is user of
+	 * the group (identified with group's internal backend group id)
+	 *
+	 * NOTE: Search by internal id is used to optimize access when
+	 * group backend/account has already been instantiated and internal id is explicitly available
+	 *
+	 * @param int $accountId
+	 * @param int $backendGroupId
+	 *
+	 * @return boolean
+	 */
+	public function isGroupUserById($accountId, $backendGroupId) {
+		return $this->isGroupMemberSqlQuery($accountId, $backendGroupId, self::MEMBERSHIP_TYPE_GROUP_USER, true);
+	}
+
+	/**
+	 * Check whether given account (identified by user's uid) is admin of
+	 * the group (identified with gid). If group predicate not specified,
+	 * it will check if user is group admin of any group
+	 *
+	 * @param string $userId
+	 * @param string $gid
+	 *
+	 * @return boolean
+	 */
+	public function isGroupAdmin($userId, $gid = null) {
+		return $this->isGroupMemberSqlQuery($userId, $gid, self::MEMBERSHIP_TYPE_GROUP_ADMIN, false);
+
+	}
+
+	/**
+	 * Search for members which match the pattern and
+	 * are users in the group (identified with gid)
+	 *
+	 * @param string $gid
+	 * @param string $pattern
+	 * @param integer $limit
+	 * @param integer $offset
+	 * @return Entity[]
+	 */
+	public function find($gid, $pattern, $limit = null, $offset = null) {
+		return $this->searchAccountsSqlQuery($gid, false, $pattern, $limit, $offset);
+	}
+
+	/**
+	 * Search for members which match the pattern and
+	 * are users in the backend group (identified with internal backend group id)
+	 *
+	 * NOTE: Search by internal id instead of gid is used to optimize access when
+	 * group backend has already been instantiated and $backendGroupId is explicitly available
+	 *
+	 * @param int $backendGroupId
+	 * @param string $pattern
+	 * @param integer $limit
+	 * @param integer $offset
+	 * @return Entity[]
+	 */
+	public function findById($backendGroupId, $pattern, $limit = null, $offset = null) {
+		return $this->searchAccountsSqlQuery($backendGroupId, true, $pattern, $limit, $offset);
+	}
+
+	/**
+	 * Count members which match the pattern and
+	 * are users in the group (identified with gid)
+	 *
+	 * @param string $gid
+	 * @param string $pattern
+	 *
+	 * @return int
+	 */
+	public function count($gid, $pattern) {
+		return $this->countSqlQuery($gid, false, $pattern);
+	}
+
+	/**
+	 * Add a group account (identified by user's internal account id)
+	 * to group (identified by group's internal backend group id).
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 *
+	 * @throws UniqueConstraintViolationException
+	 * @return bool
+	 */
+	public function addGroupUser($accountId, $backendGroupId) {
+		return $this->addGroupMemberSqlQuery($accountId, $backendGroupId, self::MEMBERSHIP_TYPE_GROUP_USER);
+	}
+
+	/**
+	 * Add a group admin account (identified by user's internal account id)
+	 * to group (identified by group's internal backend group id).
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 *
+	 * @throws UniqueConstraintViolationException
+	 * @return bool
+	 */
+	public function addGroupAdmin($accountId, $backendGroupId) {
+		return $this->addGroupMemberSqlQuery($accountId, $backendGroupId, self::MEMBERSHIP_TYPE_GROUP_ADMIN);
+	}
+
+	/**
+	 * Remove group user membership for user (identified by user's internal account id)
+	 * from group (identified by group's internal backend group id).
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 * @return bool
+	 */
+	public function removeGroupUser($accountId, $backendGroupId) {
+		return $this->removeGroupMembershipsSqlQuery($backendGroupId, $accountId, [self::MEMBERSHIP_TYPE_GROUP_USER]);
+	}
+
+	/**
+	 * Remove group admin membership for user (identified by user's internal account id)
+	 * from group (identified by group's internal backend group id).
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 * @return bool
+	 */
+	public function removeGroupAdmin($accountId, $backendGroupId) {
+		return $this->removeGroupMembershipsSqlQuery($backendGroupId, $accountId, [self::MEMBERSHIP_TYPE_GROUP_ADMIN]);
+	}
+
+	/**
+	 * Remove group memberships from group (identified by group's internal backend group id),
+	 * regardless of the role in the group.
+	 *
+	 * @param int $backendGroupId - internal id of backend group
+	 * @return bool
+	 */
+	public function removeGroupMembers($backendGroupId) {
+		return $this->removeGroupMembershipsSqlQuery($backendGroupId, null, [self::MEMBERSHIP_TYPE_GROUP_USER, self::MEMBERSHIP_TYPE_GROUP_ADMIN]);
+	}
+
+	/**
+	 * Remove group memberships for user (identified by user's internal account id),
+	 * regardless of the role in the group.
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @return bool
+	 */
+	public function removeMemberships($accountId) {
+		return $this->removeGroupMembershipsSqlQuery(null, $accountId, [self::MEMBERSHIP_TYPE_GROUP_USER, self::MEMBERSHIP_TYPE_GROUP_ADMIN]);
+	}
+
+	/**
+	 * Check if the given user is member of the group with specific membership type
+	 *
+	 * @param string|int $userId
+	 * @param string|int|null $groupId
+	 * @param string $membershipType
+	 * @param bool $useInternalIds
+	 *
+	 * @return boolean
+	 */
+	private function isGroupMemberSqlQuery($userId, $groupId, $membershipType, $useInternalIds) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->selectAlias($qb->createFunction('1'), 'exists')
+			->from('memberships', 'm');
+
+		if (!is_null($groupId)) {
+			if ($useInternalIds) {
+				$qb = $this->applyInternalPredicates($qb, $groupId, $userId, true);
+			} else {
+				$qb = $this->applyPredicates($qb, $groupId, $userId);
+			}
+		}
+
+		// Place predicate on membership_type
+		$qb->andWhere($qb->expr()->eq('m.membership_type', $qb->createNamedParameter($membershipType)));
+
+		// Limit to 1, to prevent fetching unnecessary rows
+		$qb->setMaxResults(1);
+
+		return $this->getExistsQuery($qb);
+	}
+
+
+	/**
+	 * Add user to the group with specific membership type $membershipType.
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 * @param string $membershipType
+	 *
+	 * Return will indicate if row has been inserted
+	 *
+	 * @throws UniqueConstraintViolationException
+	 * @return boolean
+	 */
+	private function addGroupMemberSqlQuery($accountId, $backendGroupId, $membershipType) {
+		$qb = $this->db->getQueryBuilder();
+
+		$qb->insert('memberships')
+			->values([
+				'backend_group_id' => $qb->createNamedParameter($backendGroupId),
+				'account_id' => $qb->createNamedParameter($accountId),
+				'membership_type' => $qb->createNamedParameter($membershipType),
+			]);
+
+		return $this->getAffectedQuery($qb);
+	}
+
+	/*
+	 * Removes users from the groups. If the predicate on a user or group is null, then it will apply
+	 * removal to all the entries of that type.
+	 *
+	 * NOTE: This function requires to use internal IDs, since we cannot
+	 * use JOIN with DELETE (some databases don't support it). We also cannot use aliases
+	 * since MySQL has specific syntax for them in DELETE
+	 *
+	 * Return will indicate if row has been removed
+	 *
+	 * @param int|null $accountId - internal id of an account
+	 * @param int|null $backendGroupId - internal id of backend group
+	 * @param int[] $membershipTypeArray
+	 *
+	 * @return boolean
+	 */
+	private function removeGroupMembershipsSqlQuery($backendGroupId, $accountId, $membershipTypeArray) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete('memberships');
+
+		if (!is_null($backendGroupId) && !is_null($accountId)) {
+			// Both backend_group_id and account_id predicates are specified
+			$qb = $this->applyInternalPredicates($qb, $backendGroupId, $accountId, false);
+		} else if (!is_null($backendGroupId)) {
+			// Group predicate backend_group_id specified
+			$qb = $this->applyBackendGroupIdPredicate($qb, $backendGroupId, false);
+		} else if (!is_null($accountId)) {
+			// User predicate account_id specified
+			$qb = $this->applyAccountIdPredicate($qb, $accountId, false);
+		} else {
+			return false;
+		}
+
+		$qb->andWhere($qb->expr()->in('membership_type',
+			$qb->createNamedParameter($membershipTypeArray, IQueryBuilder::PARAM_INT_ARRAY)));
+
+		return $this->getAffectedQuery($qb);
+	}
+
+	/*
+	 * Return backend group entities for given user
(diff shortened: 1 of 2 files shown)


mrow4a:
I will also start writing unit tests to test queries I am using, so dont focus on them to much at that point. 

If we decide on that approach, lets merge it with https://github.com/owncloud/core/pull/29107. This should not affect any other class and existing functionalities. It will when all TODOs will be implemented.

lib/private/Group/Group.php
```diff
@@ -64,6 +70,12 @@ class Group implements IGroup {
 	private $userManager;
 
 	/**
+	 * @var \OC\Group\BackendGroup $backendGroup
+	 */
+	private $backendGroup;
+
+	/**
+	 * // TODO: use BackendGroup here
```
mrow4a: All the TODOs in the code are places which should be implemented after merging this PR

lib/private/Group/GroupMapper.php
```diff
@@ -34,25 +38,27 @@ public function __construct(IDBConnection $db) {
...
 	public function getGroup($gid) {
 		$qb = $this->db->getQueryBuilder();
 		$qb->select('*')
 			->from($this->getTableName())
 			->where($qb->expr()->eq('group_id', $qb->createNamedParameter($gid)));
 
-		return $this->findEntity($qb->getSQL(), $qb->getParameters());
+		try {
+			/** @var BackendGroup $backendGroup */
+			$backendGroup = $this->findEntity($qb->getSQL(), $qb->getParameters());
```
mrow4a: Add some additional checks which are not in original version. It should return null and not throw exception

lib/private/Group/Manager.php
```diff
@@ -143,10 +144,21 @@ protected function clearCaches() {
 
 	/**
 	 * @param string $gid
+	 * @return boolean
+	 */
+	public function isCached($gid) {
```
mrow4a: This method is required to build optimized queries

lib/private/Server.php
```diff
@@ -230,12 +231,20 @@ public function __construct($webRoot, \OC\Config $config) {
...
+		});
 		$this->registerService('AccountMapper', function(Server $c) {
-			return new AccountMapper($c->getConfig(), $c->getDatabaseConnection(), new AccountTermMapper($c->getDatabaseConnection()));
+			return new AccountMapper($c->getConfig(), $c->getDatabaseConnection(), $this->getAccountTermMapper());
 		});
 		$this->registerService('GroupMapper', function(Server $c) {
 			return new GroupMapper($c->getDatabaseConnection());
 		});
+		$this->registerService('MembershipManager', function(Server $c) {
+			return new MembershipManager($c->getDatabaseConnection(), $c->getConfig(),
```
mrow4a: All changes in this file are obligatory so it actually all works.

lib/public/AppFramework/Db/Mapper.php
```diff
@@ -176,6 +176,23 @@ public function update(Entity $entity){
...
+	 *
+	 * If row contains invalid attributes, exception BadFunctionCallException will
+	 * be raised
+	 *
+	 * @param array $row the row which should be converted to an entity
+	 * @return Entity the entity
+	 * @throws \BadFunctionCallException
+	 * @since 7.0.0
+	 */
+	public function mapRowToEntity($row) {
```
mrow4a: Make it public so MembershipManager can also create the Account entities (otherwise it will only be possible inside AccountMapper)

Added @throws in phpdoc to make it clear that this function should only be used by classes which operate over oc_accounts table.

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,574 @@
...
+use OC\Group\BackendGroup;
+use OC\User\Account;
+use OCP\AppFramework\Db\Access;
+use OCP\AppFramework\Db\Entity;
+use OCP\IConfig;
+use OCP\IDBConnection;
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+
+class MembershipManager extends Access {
```
mrow4a: Please have a look on all the public interfaces of that class.

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,574 @@
...
+	 * @param IQueryBuilder $qb
+	 * @param int $userId
+	 * @param string $gid
+	 * @return IQueryBuilder
+	 */
+	private function applyPredicates($qb, $userId, $gid) {
+		// Adjust the query depending on availability of accountId and $backendGroupId
+		// to have optimized access
+		$alias = $this->getTableAlias();
+		if ($this->userManager->isCached($userId) && $this->groupManager->isCached($gid)) {
```
mrow4a: I use isCached mathods of Managers to build optimized queries - it means, fetch info from cache if available, otherwise perform JOIN.

mrow4a:
@jvillafanez

lib/private/MembershipManager.php
jvillafanez: From my point of view, this reveals a design flaw. Why something related to the DB access has to check a class from the business logic? Why it needs to check if it's cached when noone should be interested in that?.

I think we should recheck the approach. What about the group manager handles also the membership? Could it simplify this?

jvillafanez:
I'll need to do a second review

lib/private/Group/Manager.php
mrow4a: True, will write a UserCacheManager and GroupCacheManager Włoch will be used Both in Group/User Manager and Mapper

mrow4a:
Change is ready to review, full unit test coverage (target branch https://github.com/owncloud/core/pull/29107). @DeepDiver1975 @jvillafanez @butonic @PVince81

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,670 @@
...
+		$qb->selectAlias($qb->createFunction('COUNT(DISTINCT a.id)'), 'count');
+
+		$qb = $this->searchSQL($qb, $groupId, $isInternalGroupId, $pattern);
+
+		$stmt = $qb->execute();
+
+		// First should already contain count
+		$data = $stmt->fetch();
+		$stmt->closeCursor();
+		return intval($data['count']);
```
mrow4a: Is any case this might throw an exception in this query?

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,670 @@
...
+				'account_id' => $qb->createNamedParameter($accountId),
+				'membership_type' => $qb->createNamedParameter($membershipType),
+			]);
+
+		try {
+			$qb->execute();
+			return true;
+		} catch (UniqueConstraintViolationException $e) {
+			// FIXME: is this correct?
+			\OC::$server->getLogger()->logException($e, ['app' => 'core']);
```
mrow4a: @DeepDiver1975 is this correct way of logging this exception?

jvillafanez: We should check a couple of things:
* Does this class need to log something? Taking into account this is the only log here, probably not.
* Does this class need to handle that exception? I have my doubts about it.

The other option would be to not catch the exception and let it bubble up in order to handle it in other place. Note that exposing a DB exception and handle it near the output response might not be good.

I'm not saying this is bad, but there might be better options. My point is that having to log something without injecting a logger is something to be avoided, as well as injecting a class (the logger) that doesn't seem to be a clear dependency (because it's only used here). Note that swallowing the exception without doing anything is a worse option.

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,670 @@
...
+	 * @param bool $isInternalGroupId
+	 * @param string $pattern
+	 * @param integer $limit
+	 * @param integer $offset
+	 *
+	 * @return IQueryBuilder
+	 */
+	private function searchSQL($qb, $groupId, $isInternalGroupId, $pattern, $limit = null, $offset = null) {
+		$alias = $this->getTableAlias();
+		$allowMedialSearches = $this->config->getSystemValue('accounts.enable_medial_search', true);
```
jvillafanez: Is it worthy to extract this logic (getting the flag + setting the parameter variables) to another function? Taking into account this isn't used in other place than here, probably not unless the complexity of this function is high (I don't have data, but it doesn't seem so).

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,670 @@
...
+		} else if (!is_null($accountId)) {
+			// User predicate account_id specified
+			$qb->where($qb->expr()->eq('account_id', $qb->createNamedParameter($accountId)));
+		} else {
+			return false;
+		}
+
+		$qb->andWhere($qb->expr()->in('membership_type',
+			$qb->createNamedParameter($membershipTypeArray, IQueryBuilder::PARAM_INT_ARRAY)));
+		$qb->execute();
```
jvillafanez: What if the query fails? I expect the `execute()` function either return something or throws an exception. It doesn't seem this case is taken into account.

lib/private/Group/GroupMapper.php
```diff
@@ -63,4 +76,18 @@ public function search($fieldName, $pattern, $limit, $offset) {
...
+	 * @param array $row the row which should be converted to backend group
+	 * @return BackendGroup the entity
+	 * @since 10.0
+	 *
+	 * @throws \BadFunctionCallException
+	 */
+	public function mapRowToEntity($row) {
+		/** @var BackendGroup $backendGroup */
+		$backendGroup = parent::mapRowToEntity($row);
+		return $backendGroup;
```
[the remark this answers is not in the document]
mrow4a: Hmm, my point here was to make in transparent that the returned object is BackendGroup, not pure Entity. The same holds for Account. But yes, I could use this in the function which calls mapRowToEntity

jvillafanez:
I haven't checked the DB queries yet. @mrow4a as unofficial DB expert I trust the queries work properly 😉  I can have a look if needed.

Btw, I don't have data about the test coverage of the MembershipManager, but it seems quite important to verify the coverage is near 100% in that class because we'll likely rely on it heavily and the DB queries aren't easy enough to ensure they'll work on all the cases in a glance.

lib/private/MembershipManager.php
mrow4a: Hmm, makes sense. In fact, we dont handle also foreign constrain violations, so maybe this one we should not either.

lib/private/MembershipManager.php
mrow4a: Hmm, true. Thanks!

lib/private/MembershipManager.php
mrow4a: Hmm, I have better idea what to do with it. Good point here.

lib/private/Group/GroupMapper.php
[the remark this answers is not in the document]
jvillafanez: PHPDoc should take care of that, in this case in the parent class since it's the one that will be executed. Note that function visibility shouldn't need to be changed (from protected to public)

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,549 @@
...
+use OCP\AppFramework\Db\DoesNotExistException;
+
+/**
+ * Class MembershipManagerTest
+ *
+ * @group DB
+ *
+ * @package Test
+ */
+class MembershipManagerTest extends TestCase {
```
mrow4a: I covered 100% of the MembershipManager code here @jvillafanez, ale checked corner-cases, but it would be good to have a look if I did not overlook something.

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	 * @param string|int $groupId
+	 * @param bool $isInternalGroupId
+	 * @param string $pattern
+	 * @return int
+	 */
+	private function countSqlQuery($groupId, $isInternalGroupId, $pattern) {
+		$qb = $this->db->getQueryBuilder();
+
+		// We need to use distinct since otherwise we will get duplicated rows for each search term
+		$qb->selectAlias($qb->createFunction('COUNT(DISTINCT a.`id`)'), 'count');
```
mrow4a: @butonic I needed a.`id` to make it work on Oracle, should I put it also in another DISTINCT?

mrow4a:
100% unit tests coverage and tests passing on all dbms. PR ready for review and merging

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+			$qb->where($qb->expr()->eq('a.user_id', $qb->createNamedParameter($userId)));
+
+			// Apply predicate on group_id in backend groups table
+			$qb->andWhere($qb->expr()->eq('g.group_id', $qb->createNamedParameter($groupId)));
+		}
+
+		// Place predicate on membership_type
+		$qb->andWhere($qb->expr()->eq($alias.'.membership_type', $qb->createNamedParameter($membershipType)));
+
+		// Limit to 1, to prevent fetching unnecesary rows
```
mrow4a: Spelling - unnecessary

jvillafanez:
Changes look good to me. 👍

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	 * @param bool $isInternalGroupId
+	 * @param string $pattern
+	 * @param integer $limit
+	 * @param integer $offset
+	 *
+	 * @return Account[]
+	 */
+	private function searchAccountsSqlQuery($groupId, $isInternalGroupId, $pattern, $limit = null, $offset = null) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->selectAlias('DISTINCT a.id', 'id')
```
mrow4a: Should I use DISTINCT a.`id` for Oracle @butonic?

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	const MEMBERSHIP_TYPE_GROUP_USER = 0;
+	const MEMBERSHIP_TYPE_GROUP_ADMIN = 1;
+
+	/** @var IConfig */
+	protected $config;
+
+	/** @var IDBConnection */
+	private $db;
+
+	/** @var \OC\Group\GroupMapper */
```
DeepDiver1975: import all classes in the use statement above

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+								\OC\User\AccountMapper $accountMapper, \OC\User\AccountTermMapper $termMapper,
+								\OC\Group\GroupMapper $groupMapper) {
+		$this->db = $db;
+		$this->config = $config;
+		$this->accountMapper = $accountMapper;
+		$this->groupMapper = $groupMapper;
+		$this->termMapper = $termMapper;
+	}
+
+	private function getTableAlias() {
```
DeepDiver1975: needed? just use 'm'

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+		return 'm';
+	}
+
+	private function getTableName() {
```
DeepDiver1975: needed= just use 'memberships'

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	 * @return BackendGroup[]
+	 */
+	public function getAdminBackendGroups($userId) {
+		return $this->getBackendGroupsSqlQuery($userId, false, self::MEMBERSHIP_TYPE_GROUP_ADMIN);
+	}
+
+	/**
+	 * Return user account entities for given group (identified with gid)
+	 *
+	 * @param string $gid
```
DeepDiver1975: string? not int?

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	 *
+	 * @param string|int|null $groupId
+	 * @param bool $isInternalGroupId
+	 * @param int $membershipType
+	 * @return Account[]
+	 */
+	private function getAccountsSqlQuery($groupId, $isInternalGroupId, $membershipType) {
+		$qb = $this->db->getQueryBuilder();
+		$alias = $this->getTableAlias();
+		$qb->select(['a.id','a.user_id', 'a.lower_user_id', 'a.display_name', 'a.email', 'a.last_login', 'a.backend', 'a.state', 'a.quota', 'a.home'])
```
DeepDiver1975: missing white space

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+		$qb->andWhere($qb->expr()->eq($alias.'.membership_type', $qb->createNamedParameter(self::MEMBERSHIP_TYPE_GROUP_USER)));
+
+		return $qb;
+	}
+
+	/**
+	 * @param IQueryBuilder $qb
+	 * @return int
+	 */
+	private function getAffectedQuery($qb) {
```
DeepDiver1975: IQueryBuilder $qb

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+		// If affected is equal or more then 1, it means operation was successful
+		$affected = $qb->execute();
+		return $affected > 0;
+	}
+
+	/**
+	 * @param IQueryBuilder $qb
+	 * @return bool
+	 */
+	private function getExistsQuery($qb) {
```
DeepDiver1975: IQueryBuilder $qb

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	 * @param IQueryBuilder $qb
+	 * @param int $id
+	 * @param bool $isInternalId
+	 * @return IQueryBuilder
+	 */
+	private function applyUserPredicates($qb, $id, $isInternalId) {
+		// Adjust the query depending on availability of accountId
+		// to have optimized access
+		$alias = $this->getTableAlias();
+		if ($isInternalId) {
```
DeepDiver1975: I'd split this into 2 functions - you gain nothing from this

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,714 @@
...
+	 * @param IQueryBuilder $qb
+	 * @param string $id
+	 * @param bool $isInternalId
+	 * @return IQueryBuilder
+	 */
+	private function applyGroupPredicates($qb, $id, $isInternalId) {
+		// Adjust the query depending on availability of accountId
+		// to have optimized access
+		$alias = $this->getTableAlias();
+		if ($isInternalId) {
```
DeepDiver1975: split into two function

lib/private/User/AccountMapper.php
```diff
@@ -168,7 +168,7 @@ public function find($pattern, $limit = null, $offset = null) {
 		$qb->selectAlias('DISTINCT a.id', 'id')
 			->addSelect(['user_id', 'lower_user_id', 'display_name', 'email', 'last_login', 'backend', 'state', 'quota', 'home'])
 			->from($this->getTableName(), 'a')
-			->leftJoin('a', 'account_terms', 't', $qb->expr()->eq('a.id', 't.account_id'))
```
DeepDiver1975: keep this for readability - we will never change the table names

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,626 @@
...
+	}
+
+	private static function clearDB() {
+		\OC::$server->getDatabaseConnection()->executeQuery('DELETE FROM `*PREFIX*memberships`');
+		\OC::$server->getDatabaseConnection()->executeQuery('DELETE FROM `*PREFIX*backend_groups`');
+		\OC::$server->getDatabaseConnection()->executeQuery('DELETE FROM `*PREFIX*accounts`');
+	}
+
+	public static function tearDownAfterClass () {
+		MembershipManagerTest::clearDB();
```
DeepDiver1975: self:: instead of MembershipManagerTest::

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,626 @@
...
+		$this->assertCount(4, $this->manager->getAdminAccounts());
+		$this->assertCount(4, $this->manager->getGroupUserAccounts($group->getGroupId()));
+	}
+
+	/**
+	 * Test that deleting group should result in deleting all users, and violating that
+	 * should rise exception
+	 */
+	public function testRemoveFailed() {
+		if ($this->connection->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\SqlitePlatform) {
```
DeepDiver1975: import sqliteplatform

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,626 @@
...
+			$this->markTestSkipped("Foreign key are not fully supported on SQLite, skip test");
+			return;
+		}
+		$account = MembershipManagerTest::getAccount(1);
```
DeepDiver1975: self::

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,626 @@
...
+
+		$accountMapper = \OC::$server->getAccountMapper();
+		$groupMapper = \OC::$server->getGroupMapper();
+
+		$thrown = false;
+		try {
+			$accountMapper->delete($account);
+		} catch (\Exception $e) {
+			$thrown = true;
+			$this->assertInstanceOf(\Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException::class, $e);
```
DeepDiver1975: import

DeepDiver1975 (changes requested):

lib/private/User/AccountMapper.php
mrow4a: I exactly changed that for readability also, so we don't use magic strings which are easy to mistake or confuse

lib/private/MembershipManager.php
mrow4a: Not sure what you mean here

lib/private/MembershipManager.php
mrow4a: I would prefer to keep it, because we might use this table also in other parts of the code later - e.g. fs setup optimization.

mrow4a: I mean, your point is that table names won't change right?

DeepDiver1975: tablenames dont change - yes and a a one line function like this is pointless

lib/private/MembershipManager.php
tomneedham: Type hint the parameter

lib/private/MembershipManager.php
butonic: When using `$qb->createFunction(` you basically pass raw SQL. That is why you need to do the column escaping. With `$qb->expr()->eq(` that is not necessary, because it will do the column quoting (`a.id` becomes <code>a.&#96;id&#96;</code>) for you.

Where exactly would you put another DISTINCT?

lib/private/MembershipManager.php
butonic: Distinc with alias is a littlve ugly. Whil doctrine has `selectAlias` and `selectDistinct` neither does what you need. `selectAlias` will treat `DISTINCT a.id` as a single identifier, escaping it to <code>DISTINCT a.`id`<code>. While that looks correct it is only by chance. Either you have test coverage for this, so we detect when docrine decides to treat this differently, or you wrap the rew sql in a function to make the problem explicit:
```php
$qb->selectAlias($qb->createFunction('DISTINCT a.`id`'), 'id');
```
Then again, this also only works as long as doctrine does not parse the funtion string for actual brackets...


So ... I would leave it as it is, but cover it with a unit test that checks the DISTINCT is working.

mrow4a:
@DeepDiver1975 can we merge it into central_groups branch? I addressed all the issues you pointed

mrow4a:
@DeepDiver1975 @PVince81 I have tested it with https://github.com/owncloud/core/pull/29282 integration into group manager, group, subadmin, user and user manager. This PR is ready

mrow4a:
Any input here?

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,733 @@
...
+class MembershipManager {
+
+	/**
+	 * types of memberships in the group
+	 */
+	const MEMBERSHIP_TYPE_GROUP_USER = 0;
+	const MEMBERSHIP_TYPE_GROUP_ADMIN = 1;
+
+	/** @var IConfig */
+	protected $config;
```
jvillafanez: If this class isn't expected to be extended, this should be private.

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,733 @@
...
+
+	/** @var IDBConnection */
+	private $db;
+
+
+	public function __construct(IDBConnection $db, IConfig $config) {
```
DeepDiver1975: PHPDoc missing

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,733 @@
...
+
+		return $this->getAffectedQuery($qb);
+	}
+
+	/*
+	 * Return backend group entities for given user id $userId (internal or gid) of which
+	 * the user has specific membership type
+	 *
+	 * @param string|int $userId
+	 * @param bool $isInternalUserId
```
DeepDiver1975: what is internal?

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,733 @@
...
+		}
+
+		$qb->andWhere($qb->expr()->in('membership_type',
+			$qb->createNamedParameter($membershipTypeArray, IQueryBuilder::PARAM_INT_ARRAY)));
+
+		return $this->getAffectedQuery($qb);
+	}
+
+	/*
+	 * Return backend group entities for given user id $userId (internal or gid) of which
```
DeepDiver1975: gid?

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,733 @@
...
+	}
+
+	/**
+	 * Return backend group entities for given account (identified by user's uid)
+	 *
+	 * @param string $userId
+	 *
+	 * @return BackendGroup[]
+	 */
+	public function getUserBackendGroups($userId) {
```
DeepDiver1975: do we really need this function? From my understanding everything should internally work on accountIds

lib/private/Server.php
```diff
@@ -991,6 +999,20 @@ public function getUserManager() {
...
+	 * @return \OC\MembershipManager
+	 */
+	public function getMembershipManager() {
+		return $this->query('MembershipManager');
+	}
+
+	/**
+	 * @return \OC\User\AccountTermMapper
+	 */
+	public function getAccountTermMapper() {
```
DeepDiver1975: why? public getters in class Server shall only return OCP interfaces

DeepDiver1975:
@ownclouders rebase

ownclouders:

DeepDiver1975 (changes requested):

ownclouders:

lib/private/Server.php
mrow4a: Hmm, so what with AccountMapper?

lib/private/MembershipManager.php
mrow4a: It is just optimization, in case account id is only available by fetching the use from user manager.

mrow4a:
@DeepDiver1975 addressed your doubts and fixed

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,631 @@
...
+
+use OC\Group\BackendGroup;
+use OC\Group\GroupMapper;
+use OC\User\Account;
+use OC\User\AccountMapper;
+use OC\MembershipManager;
+use OC\User\AccountTermMapper;
+use OCP\IConfig;
+use OCP\IDBConnection;
+use Test\TestCase;
```
DeepDiver1975: unused

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,631 @@
...
+use OCP\AppFramework\Db\DoesNotExistException;
```
DeepDiver1975: unused

tests/lib/MembershipManagerTest.php
```diff
@@ -0,0 +1,631 @@
...
+		$this->manager = new MembershipManager(
+			$this->connection,
+			$this->config
+		);
+	}
+
+	private static function clearDB() {
+		\OC::$server->getDatabaseConnection()->executeQuery('DELETE FROM `*PREFIX*memberships`');
+		\OC::$server->getDatabaseConnection()->executeQuery('DELETE FROM `*PREFIX*backend_groups`');
+		\OC::$server->getDatabaseConnection()->executeQuery('DELETE FROM `*PREFIX*accounts`');
```
DeepDiver1975: really kill all?

lib/private/Server.php
```diff
@@ -89,6 +89,7 @@
 use OCP\AppFramework\QueryException;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OC\Group\GroupMapper;
+use OC\MembershipManager;
```
DeepDiver1975: unused

lib/private/Server.php
```diff
@@ -610,13 +611,13 @@ public function __construct($webRoot, \OC\Config $config) {
...
 
 			return new Checker(
-					new EnvironmentHelper(),
-					new FileAccessHelper(),
-					new AppLocator(),
-					$config,
-					$c->getMemCacheFactory(),
-					$appManager,
-					$c->getTempManager()
+				new EnvironmentHelper(),
```
DeepDiver1975: remove whitspace changes

lib/private/User/AccountMapper.php
```diff
@@ -239,4 +239,4 @@ public function callForAllUsers($callback, $search, $onlySeen) {
 		$stmt->closeCursor();
 	}
 
-}
+}
```
DeepDiver1975: remove whitespace changes

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,744 @@
...
+	/**
+	 * Add a group account (identified by user's internal account id)
+	 * to group (identified by group's internal backend group id).
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 *
+	 * @return bool
+	 */
+	public function addGroupUser($accountId, $backendGroupId) {
```
DeepDiver1975: addGroupMemberSqlQuery throws UniqueConstraintViolationException - see PHPDoc
this method should do also throw it

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,744 @@
...
+	/**
+	 * Add a group admin account (identified by user's internal account id)
+	 * to group (identified by group's internal backend group id).
+	 *
+	 * @param int $accountId - internal id of an account
+	 * @param int $backendGroupId - internal id of backend group
+	 *
+	 * @return bool
+	 */
+	public function addGroupAdmin($accountId, $backendGroupId) {
```
DeepDiver1975: same as with addGroupUser

lib/private/MembershipManager.php
```diff
@@ -0,0 +1,744 @@
...
+	}
+
+	/**
+	 * Return account entities for given group
+	 * (identified by backend group id if $isBackendGroupId is true, or gid if $isBackendGroupId is false)
+	 * of which the accounts have specific membership type. If group id is not specified, it will
+	 * return result for all groups.
+	 *
+	 * @param string|int|null $groupId
+	 * @param bool $isInternalGroupId
```
DeepDiver1975: isInternalGroupId unknown 
isBackendGroupId missing in PHPDoc

DeepDiver1975 (changes requested):

mrow4a:
@DeepDiver1975 adjusted to comments

DeepDiver1975 (approved):

Issue #29139, opened 2017-10-03
Integration Tests: missing check in WebDav trait


cortho:
When I have an exception in _listFolder_ (see https://github.com/owncloud/core/blob/master/tests/integration/features/bootstrap/WebDav.php#L487)
Then _array_keys() expects parameter 1 to be array, object given_ in https://github.com/owncloud/core/blob/master/tests/integration/features/bootstrap/WebDav.php#L1081
because _$elementList_ was not checked to be an array.

SergioBertolinSG:
Hi @cortho , in which situation you need this step 

```gherkin
When /^user "([^"]*)" deletes everything from folder "([^"]*)"$/
```

to not fail when the listFolder function cannot list the elements?

cortho:
@phil-davis was faster, this is his PR: https://github.com/owncloud/core/pull/29157

phil-davis:
I never noticed this issue - I was looking at test logs today and saw the problem. @cortho is using this method in an AfterScenario to do a general cleanup of "rubbish" that [has been|might have been] uploaded during a scenario. For scenarios that do not upload anything, the folder is already empty.

The AfterScenario could be changed to do different cleanup (e.g. probably should delete the user that was created - then the user's files will be gone and the skeleton will come back when a user is created again in the next scenario)

But in any case, this ``deletes everything from folder`` should be fixed up to work with an empty folder anyway.

SergioBertolinSG:
OK, lets merge that then.

cortho:
Working as expected now, thanks. Merged, backported -> closing

PR #29157, opened 2017-10-06
Handle case of empty folder in userDeletesEverythingInFolder

phil-davis:
## Description
If there are no files in a folder, then do not attempt to traverse the "array" of files deleting them - in ``userDeletesEverythingInFolder``

## Related Issue
#29139 

## Motivation and Context
A test cleanup step in a set of app integration tests was using ``userDeletesEverythingInFolder`` and it would spit out warnings for scenarios that never actually created any files:
```
Warning: array_keys() expects parameter 1 to be array, object given in /home/travis/build/owncloud/core/tests/integration/features/bootstrap/WebDav.php line 1081
```

## How Has This Been Tested?
Jenkins will tell.

## Types of changes

## Checklist:

Fixing this in core will allow callers to use it cleanly even if the folder happens to already have nothing in it.

### DIFF
--- a/tests/integration/features/bootstrap/WebDav.php
+++ b/tests/integration/features/bootstrap/WebDav.php
@@ -1078,14 +1078,16 @@ trait WebDav {
 	 */
 	public function userDeletesEverythingInFolder($user, $folder)  {
 		$elementList = $this->listFolder($user, $folder, 1);
-		$elementListKeys = array_keys($elementList);
-		array_shift($elementListKeys);
-		$davPrefix =  "/" . $this->getDavFilesPath($user);
-		foreach ($elementListKeys as $element) {
-			if (substr($element, 0, strlen($davPrefix)) == $davPrefix) {
-				$element = substr($element, strlen($davPrefix));
+		if (is_array($elementList) && count($elementList)) {
+			$elementListKeys = array_keys($elementList);
+			array_shift($elementListKeys);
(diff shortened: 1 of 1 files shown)


SergioBertolinSG:
This way it won't fail if there is no such a folder.

phil-davis:
Is it used to also confirm that a folder exists?
If there is no such folder, then the step is successful by definition:
"And user myname deletes everything from folder /non-existing-folder"

the step is a no-op, so it is successful.

``asTheFileOrFolderExists()`` checks for existence.

And actually I think ``listFolder()`` will throw an exception anyway if the folder does not exist. So there will be an exception for any of the steps that use ``listFolder`` if they try a folder that does not exist.

So ``userDeletesEverythingInFolder`` should throw an exception if the folder does not exist - before and after this change.

Here (I think, I hope) I am just avoiding the PHP warning that could be emitted.

SergioBertolinSG:
> If there is no such folder, then the step is successful by definition:

Ok, thats fine to me.



> And actually I think listFolder() will throw an exception anyway if the folder does not exist. So there will be an exception for any of the steps that use listFolder if they try a folder that does not exist.

No because that exception is catched https://github.com/owncloud/core/blob/master/tests/integration/features/bootstrap/WebDav.php#L487

SergioBertolinSG (approved):

phil-davis:
Backport stable10 #29163

PR #29160, opened 2017-10-06
Remove duplicated integration test trait properties

phil-davis:
## Description
Some properties in ``BasicStructure`` are also defined in other trait files. Remove the extra definitions - one central property is what is needed.

## Motivation and Context
Integration test logs can have messages like:
```
PHP Strict standards:  AppConfiguration and WebDav define the same property ($response) in the composition of BasicStructure. This might be incompatible, to improve maintainability consider using accessor methods in traits instead. Class was composed in /home/travis/build/owncloud/core/tests/integration/features/bootstrap/BasicStructure.php on line 416
```
Some properties are "shared" by multiple traits. Already there are some properties  that are defined in just 1 trait file and used from multiple trait files, so that works.

This warning may have came after I re-organized the trait hierarchy into a "flat tree" with ``BasicStructure`` at the top, and everything else "use"d in there. Previous to that there were various different ways to "use" into the cyclical "use" include "tree" and perhaps those sub-trees did not have this happen.

## How Has This Been Tested?
Jenkins knows

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/AppConfiguration.php
+++ b/tests/integration/features/bootstrap/AppConfiguration.php
@@ -5,11 +5,6 @@ use GuzzleHttp\Message\ResponseInterface;
 require __DIR__ . '/../../../../lib/composer/autoload.php';
 
 trait AppConfiguration {
-	/** @var string */
-	private $currentUser = '';
-
-	/** @var ResponseInterface */
-	private $response = null;
 
 	abstract public function sendingTo($verb, $url);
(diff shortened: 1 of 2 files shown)


phil-davis:
I noticed this while looking at integration test logs from an app that makes use of this stuff.

SergioBertolinSG (approved):

phil-davis:
Backport stable10 #29171

PR #29161, opened 2017-10-06
Revert hard coded Name string in templates

lordelix:
## Description
Revert back hard-coded name string in core template files

## Related Issue

https://github.com/owncloud/enterprise/issues/2271

## How Has This Been Tested?

Visual test

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/templates/public.php
+++ b/apps/files_sharing/templates/public.php
@@ -56,7 +56,7 @@ OCP\Util::addHeader('meta', ['property' => "og:image", 'content' => $_['previewI
 		 data-owner-display-name="<?php p($_['displayName']) ?>" data-owner="<?php p($_['owner']) ?>" data-name="<?php p($_['filename']) ?>">
 		<a href="<?php print_unescaped(link_to('', 'index.php')); ?>" title="" id="owncloud">
 			<h1 class="logo-icon">
-				<?php $theme->getName(); ?>
+				<?php p($theme->getName()); ?>
 			</h1>
 		</a>
 
--- a/apps/files_sharing/templates/public.php
+++ b/apps/files_sharing/templates/public.php
@@ -56,7 +56,7 @@ OCP\Util::addHeader('meta', ['property' => "og:image", 'content' => $_['previewI
 		 data-owner-display-name="<?php p($_['displayName']) ?>" data-owner="<?php p($_['owner']) ?>" data-name="<?php p($_['filename']) ?>">
 		<a href="<?php print_unescaped(link_to('', 'index.php')); ?>" title="" id="owncloud">
 			<h1 class="logo-icon">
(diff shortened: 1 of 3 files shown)


apps/files_sharing/templates/public.php
```diff
@@ -56,7 +56,7 @@
 		 data-owner-display-name="<?php p($_['displayName']) ?>" data-owner="<?php p($_['owner']) ?>" data-name="<?php p($_['filename']) ?>">
 		<a href="<?php print_unescaped(link_to('', 'index.php')); ?>" title="" id="owncloud">
 			<h1 class="logo-icon">
-				<span class="own">own</span><span class="cloud">Cloud</span>
+				<?php $theme->getName(); ?>
```
felixheidecke: This, unfortunately, overrides CE & EE theme. Shouldn't happen but does :-1:

lib/private/legacy/defaults.php
```diff
@@ -54,7 +54,7 @@ function __construct() {
 		$version = \OCP\Util::getVersion();
 
 		$this->defaultEntity = 'ownCloud'; /* e.g. company name, used for footers and copyright notices */
-		$this->defaultName = 'ownCloud'; /* short name, used when referring to the software */
+		$this->defaultName = '<span class="own">own</span><span class="cloud">Cloud</span>'; /* short name, used when referring to the software */
```
felixheidecke: To provide the look needed in the EE Theme, we a bit more HTML in the defaultName. (Related to https://github.com/owncloud/core/pull/29161/files#r143151177)

This should not be escaped but render out valid HTML. @patrickjahns or @DeepDiver1975

michaelstingl (changes requested):
Please check my comment in https://github.com/owncloud/enterprise/issues/2271#issuecomment-334782490

lib/private/legacy/defaults.php
patrickjahns: 👎 I am not comfortable adding HTML tags to a string only field

PVince81:
Fixes https://github.com/owncloud/core/issues/29172

felixheidecke:
I removed the hard-coded ownCloud occurrences in core.
This will alter the look in the files_sharing app in conjunction with the enterprise theme, but I guess we can live with that ;-)

@patrickjahns

DeepDiver1975:
please squash ...

felixheidecke:
Please review @DeepDiver1975 & @patrickjahns … No longer in progress.

patrickjahns (approved):
LGTM

@michaelstingl has the final say in this matter

michaelstingl (approved):
LGTM, but I only tested the ownBrander generated theme and I didn't test if this breaks other stuff… (vanilla oC, example theme etc)

PVince81:
@felixheidecke please backport to stable10

PR #29162, opened 2017-10-06
[stable8.2] Add command to pre-migrate oc_properties

tomneedham:


core/command/migrate/davproperties.php
```diff
@@ -0,0 +1,196 @@
...
+			$sql = "ALTER TABLE *PREFIX*properties ADD fileid BIGINT(20) NULL";
+			$this->connection->executeQuery($sql);
+		} catch (NonUniqueFieldNameException $e) {
+			$output->writeln("Column already appear to exist");
+		}
+
+		$output->writeln("Adding fileid index to oc_properties table");
+		// Add the index
+		try {
+			$sql = "CREATE INDEX fileid_index ON *PREFIX*properties (fileid)";
```
DeepDiver1975: Quotes? And this looks like this will not work on all dbs. Shall we at least add some platform limitation?

DeepDiver1975 (changes requested):

tomneedham:
Updated the SQL to hopefully make it valid for the platforms we support.

tomneedham:
Closing for now unless this becomes necessary

PR #29175, opened 2017-10-09
Don't access FS on property migration - allow pre-migration

tomneedham:
## Description
Forward port of https://github.com/owncloud/core/pull/29162/files.

This PR is a forwardport of a fix to the oc_properties migration that does not touch the FS during the migration - instead uses the caches - to speed things up. Additionally, it allows the table to be pre-migrated (fileid is already present) so this can be done without downtime, piror to the upgrade. We simply limit the SQL query to only find rows that have a NULL fileid.

## Motivation and Context
Big instance upgrades can take a long time to init the FS for users - and checking `file_exists` for each entry is expensive.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/appinfo/Migrations/Version20170202213905.php
+++ b/apps/dav/appinfo/Migrations/Version20170202213905.php
@@ -21,6 +21,7 @@
 
 namespace OCA\DAV\Migrations;
 
+use OC\Files\Filesystem;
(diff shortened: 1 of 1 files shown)


tomneedham:
This is the pre-migration for 8.2 https://github.com/owncloud/core/pull/29162

tomneedham:
Closing for now - maybe we will end up needing this but not essential.

PR #28587, opened 2017-08-04
Add occ command dav:cleanup-chunks to remove all upload folders which…

DeepDiver1975:
… are older then 2 days 

## Related Issue
fixes #26981

## How Has This Been Tested?
- create an upload folder via dav
- set the mtime in database to an older value then 2 days ago
- run the command
- see the foller being killed

## Screenshots (if appropriate):
bildschirmfoto von 2017-08-04 16-27-00

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29180, opened 2017-10-09
[stable10] Add occ command dav:cleanup-chunks to remove all upload fo…

### LINKED ISSUE #26981: ChunkingNG: Stray chunks are never deleted
### Steps to reproduce
1. Start uploading a 500MB file using the client 2.3
2. While chunks are still being uploaded, touch the source file (or move it out of the directory, and in again after the sync). The sync will be aborted and a new one will try to re-upload chunks from the start, using a new transfer-ID and leaving old chunk files on the server.
3. Complete the sync
4. Repeat every day for 4 months

### Expected behaviour
Chunks of failed uploads are periodically deleted from the server's `data/$user/uploads/` directory.

Since chunks are now accessible through WebDAV we could, in addition, issue a DELETE on the collection from the client if that verb is implemented for `/uploads/`. But there will still be cases where the client won't be able to clean itself up (for example if a client is shut down and then uninstalled in the middle of a chunked upload).

### Actual behaviour
Chunks accumulate and fill the disk, the admin must clean them up manually (but do they still remain in the DB?)

The "old chunking" chunks of a user seem to be cleaned up every time that user logs into the web interface. It would already help if this was the case.

### Server configuration
master (1f7167a96bd9c8510f9b69d865a637c87f52015f)

DeepDiver1975:
…lders which are older then 2 days - fixes #26981

backport of #28587 to stable10

### DIFF
--- a/apps/dav/appinfo/info.xml
+++ b/apps/dav/appinfo/info.xml
@@ -26,5 +26,6 @@
 		<command>OCA\DAV\Command\CreateCalendar</command>
 		<command>OCA\DAV\Command\SyncBirthdayCalendar</command>
 		<command>OCA\DAV\Command\SyncSystemAddressBook</command>
+		<command>OCA\DAV\Command\CleanupChunks</command>
 	</commands>
 </info>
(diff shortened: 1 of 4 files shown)


PVince81 (approved):
:+1:

PVince81:
Jenkins kicked

PR #28970, opened 2017-09-11
Preserve sharing settings across integration test runs

phil-davis:
## Description
1) Make generic step definitions for ``When the capabilities are retrieved`` and thus be able to hide the repeated detail about doing a ``GET`` and checking for status ``200``.
2) Move ``Given as an "admin"`` into background since it starts every capabilities scenario.
3) Add code that can check an individual setting rather than a whole set, and make a few tests that use that code (so we know it works)
4) Save the existing capabilities at the start of a scenario.
5) Set the desired starting capabilities relevant to the context (pretty much as in master)
6) Put back the saved capabilities on the way out.
7) Various methods added along the way do reduce duplication of lists of "common" "standard" groups of settings.

After this, I can run any suite/feature/scenario by itself and it sets up the sharing/federation settings that it wants to start with, then puts things back the way they were on the way out. This lets me run "random" bits of integration tests in among other tests and playing in the UI trying to track down bugs, without having to think about what happened to settings.

## Motivation and Context
Be nice to people running all different unit, integration and UI tests in their dev environment.

## How Has This Been Tested?
Try starting by first manually turning off "important" settings (like turn off file sharing completely, turn off group sharing). Run just an individual suite or feature from each suite and make sure they pass. Confirm that the previous settings come back as the test exits.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
Helpful stuff to make integration tests a bit more robust about their environment, and more friendly to everyday developers.

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29195, opened 2017-10-10
Refactor set and reset of capabilities

phil-davis:
## Description
1) ``wasCapabilitySet`` was supposed to return a proper ``boolean`` - fixed.
2) Because of that, the comparison ``if ($savedState !== $testingState)`` was always true because the types were different. Item (1) makes sure that ``$savedState`` is a ``boolean``.
3) After fixing that, it meant that the test condition was happening. So when it seems that the capability is already set correctly, there is no "wasted" attempt to set it again. Great in theory! 

But in practice, enabling some capability settings will also change sub-settings. e.g. when ``files_sharing`` ``api_enabled`` is switched on, then various sub-capabilities get set to their defaults. So the original saved state of capabilities is not a good indicator of what is currently set/not set. There were then some failures, particularly in the capabilities and sharing suites.

To fix that in the general case, you would have to re-query the capabilities after every change you make. That defeats the purpose of trying to be efficient and do a single capabilities query, then only change the things that need changing. It turns out to be just as quick to set every capability, whether it is already set or not.

4) Save the capabilities that really do get changed as an array. At the end of the scenario these can be put back. It eliminates needing individual ``restore*`` methods that know what to put back. And eliminates the need for the (new) ``restoreAppConfigs()`` method.

5) Rename ``setupAppConfigs()`` back to ``resetAppConfigs()`` - along with item (4) this reverts the abstract methods in ``AppConfiguration.php`` back to the way they were previously. It means that existing integration tests in apps that use ``BasicStructure`` and ``AppConfiguration`` do not get broken (i.e. do not have to all be patched)

## Related Issue
This is a follow-on from PR #28970 

## How Has This Been Tested?
Extra test runs in Jenkins in temporary PRs to make sure it works and does not take too much time.

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/AppConfiguration.php
+++ b/tests/integration/features/bootstrap/AppConfiguration.php
@@ -35,6 +35,11 @@ trait AppConfiguration {
 	 * @var string the original capabilities in XML format
 	 */
 	private $savedCapabilitiesXml;
+	
+	/**
+	 * @var array the changes made to capabilities for the test scenario
+	 */
+	private $savedCapabilitiesChanges = [];
 
 	/**
 	 * @param string $verb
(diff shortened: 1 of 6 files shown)


phil-davis:
The original PR is not "broken" in an outward way. The unexpected "always true" at point 2 was hiding an underlying issue.

And also, making these refactorings also helps avoid needing to touch the integration tests in apps.

phil-davis:
Backport stable10 #29200

SergioBertolinSG (approved):

Issue #28406, opened 2017-07-17
User Id "0" issues


phil-davis:
### Steps to reproduce
1. Add a user with user id "0"
2. Login as "0"
3. Try to share a file

### Expected behaviour
The share panel opens and lets you share the file

### Actual behaviour
An error, "Problem loading page, reloading in 5 seconds".
The sharing panel does not display.

### Server configuration
Any I expect (was on current master on a dev VM)

### Client configuration
**Browser:**
Firefox
**Operating system:**
Ubuntu
#### ownCloud log (data/owncloud.log)
```
NetworkError: 401 Unauthorized - http://localhost:8080/ocs/v2.php/apps/files_sharing/api/v1/shares?format=json&path=%2F%27single%27quotes&reshares=true"
```

Some code that checks the supplied username+password when accessing this area of the API is failing when supplied with "0" for the user id. (Other file operations like create/rename/delete are working)

There are places in lib/private/user/Session.php that do stuff like:
```
if (!empty($request->server['PHP_AUTH_USER']) && ...
```
which will not do the expected code when the user is "0".

And possibly quite a few other places where the user id is used and checked.

phil-davis:
I came across this because I was testing the case when the user id is a string of digits "123". That seems to work, and I guess there are installations out there that will use an integer as a user id (maybe a University that uses a student number directly as the user id? or...)

I tried "0" and it has problems - which will be due to the pain of empty("0") === true in PHP and similar.

Actually should "0" just be declared not a valid user id (reject it when adding a user)?

hodyroff:
should be fixed - if not too difficult.

PVince81:
yeah, we need to find all of those places anyway... we're already trying to avoid such problems in new code but there is likely old code still doing the checks wrong... thank you PHP for `empty('0') === true` ...

phil-davis:
#29198 requires usernames to be 3 or more chars. So that would "fix" (=workaround) this particular issue.

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

phil-davis:
Username cannot be "0" any more. So the particular issue reported here is avoided. Leaving closed.

PR #29198, opened 2017-10-10
username must be at least 3 characters long

peterprochaska:
## Description
Username must be at least 3 characters long

## Related Issue
HackerOne and https://github.com/owncloud/security-tracker/issues/227

## How Has This Been Tested?
You can't add a user with 'a' or 'ab'

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -289,6 +289,7 @@ class Manager extends PublicEmitter implements IUserManager {
 	 */
 	public function createUser($uid, $password) {
 		$l = \OC::$server->getL10N('lib');
+
 		// Check the name for bad characters
 		// Allowed are: "a-z", "A-Z", "0-9" and "_.@-'"
 		if (preg_match('/[^a-zA-Z0-9 _\.@\-\']/', $uid)) {
@@ -303,6 +304,12 @@ class Manager extends PublicEmitter implements IUserManager {
 		if (strlen(trim($uid, "\t\n\r\0\x0B\xe2\x80\x8b")) !== strlen(trim($uid))) {
 			throw new \Exception($l->t('Username contains whitespace at the beginning or at the end'));
 		}
+
+		// Username must be at least 3 characters long
+		if(strlen($uid) < 3) {
+			throw new \Exception($l->t('The username must be at least 3 characters long'));
+		}
+
 		// No empty password
 		if (trim($password) == '') {
 			throw new \Exception($l->t('A valid password must be provided'));


DeepDiver1975 (approved):

DeepDiver1975:
@Peter-Prochaska did you test this? jenkins is red - please have a look

phil-davis:
This will "fix" the user named "0" issue, by banning any 1 (or 2) char string as a username: #28406

phil-davis:
I guess this only applies to "local" ownCloud-created usernames. Short ones could still come from LDAP?

PVince81:
> Short ones could still come from LDAP?

Yes

Peter-Prochaska:
Yes, localy the tests are running. I think this is a jenkis error

.htaccess
```diff
@@ -84,4 +84,3 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
-
```
PVince81: unrelated?

Peter-Prochaska: fixed

lib/private/User/Manager.php
```diff
@@ -289,6 +289,12 @@ public function searchDisplayName($pattern, $limit = null, $offset = null) {
 	 */
 	public function createUser($uid, $password) {
 		$l = \OC::$server->getL10N('lib');
+
+		// Username must at least 3 characters long
+		if(strlen($uid) <= 3) {
```
mmattel: This ends up in min 4

mmattel: Additionally I would place it after the whitespace check...

PVince81 (changes requested):
see @mmattel's comments

.htaccess
```diff
@@ -83,5 +83,4 @@ AddDefaultCharset utf-8
 Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
-</IfModule>
-
+</IfModule>
```
kdslkdsaldsal: I guess this will break the .htaccess again? AFAIK the newline is expected.

lib/private/User/Manager.php
Peter-Prochaska: No and no, <=3 ist min 3. And after the whitespace check we have the problem with the "."

.htaccess
Peter-Prochaska: What now? Expected or not?

lib/private/User/Manager.php
phil-davis: The message says ``The username must be at least 3 characters long`` which means it can be 3, 4, 5... characters long. But if I put in a 3-character username "abc" it will not let me do it.
So the "if" test here and the message have different things. One of themm needs to be changed, depending on what is the requirement.

.htaccess
phil-davis: In the first place a change to ``.htaccess`` should never have been added to the commit. Sometimes git is seeing that it was modified during a local dev install, and then "just adds it" to the commit (it depends what git tool you are using and what is in the ``.gitignore`` file etc.)

So you need to press "Enter" at the end of the last line, commit and squash. Hopefully ``.htaccess`` will then become identical to the current version in master, and so it will disappear from this PR.

lib/private/User/Manager.php
PVince81: just replace `<=` with `<`

.htaccess
PVince81: `git checkout master .htaccess` to revert the changes.

htaccess is not relevant to this PR / commit, so if we want something different there, needs to be fixed separately

lib/private/User/Manager.php
```diff
@@ -289,6 +289,12 @@ public function searchDisplayName($pattern, $limit = null, $offset = null) {
 	 */
 	public function createUser($uid, $password) {
 		$l = \OC::$server->getL10N('lib');
+
+		// Username must at least 3 characters long
```
phil-davis: Username must be at least...

lib/private/User/Manager.php
```diff
@@ -289,6 +289,12 @@ public function searchDisplayName($pattern, $limit = null, $offset = null) {
...
+		if(strlen($uid) < 3) {
```
phil-davis: I like to have space between keywords an the next bracket, like in the next ``if (` below.
It is part of some coding standards - don't know if it is the rule in oC.

phil-davis (commented):
Just little comments.

lib/private/User/Manager.php
mmattel: Is there a reason why you put that quey at the start - just before all other checks like whitespace and valid characters?
Should´nt the result be **minimum three valid characters** ?

DeepDiver1975: Quite valid question

PVince81: the answer is likely: need more ☕️ 😉

lib/private/User/Manager.php
Peter-Prochaska: done

PVince81 (approved):
:+1:

PVince81:
restarted jenkins

PR #29206, opened 2017-10-11
[WIP] Update guzzle 2 short ci

phil-davis:
Just running integration tests to get a list of what fails now.


phil-davis:
Note about ``getEffectiveUrl()`` - I removed the use of it (was just in a text message, I replaced it with other text). To get that sort of data from the response you now have to get hold of ``TransferStats``. That was quite a few method calls down to do that, and then pass the answer all the way back... and it was easier just to not bother.

For how to:
https://stackoverflow.com/questions/30682307/how-to-read-the-response-effective-url-in-guzzle-6-0
```
$client = new \GuzzleHttp\Client();

$client->get('http://some.site.com', [
    'query'   => ['get' => 'params'],
    'on_stats' => function (TransferStats $stats) use (&$url) {
        $url = $stats->getEffectiveUri();
    }
])->getBody()->getContents();

echo $url; // http://some.site.com?get=params
```

phil-davis:
This was for testing using shorter CI - closing.

Issue #29087, opened 2017-09-23
Files metadata endpoint


PVince81:
@DeepDiver1975 @butonic as discussed at the conference:

"/meta/$fileid" DAV and View endpoints to access versions, etc.

![img_20170922_120725](https://user-images.githubusercontent.com/277525/30776882-d74c108e-a0ae-11e7-937d-b04db9014712.jpg)

DeepDiver1975:
DAV @sharidas :
- [ ] create MetaData collection in dav app - name 'meta'
- [ ] second path is the file id (uuid encoded bigint)
- [ ] create subfolder v for versions
- [ ] create subfolder p for previews

Filesystem @deepdiver1975:
- [ ] create mount 'meta' and corresponding storage implementation
- [ ] define storage interfaces for versions etc

Frontend:
- [ ] use v dav endpoint in versions sidebar
- [ ] use p dav endpoint to generate previews

DeepDiver1975:
Test Scenarios:
- [ ] restore version of read only shared file -> shall not be allowed

PVince81:
@DeepDiver1975 are we done ?

DeepDiver1975:
once preview is done - yes

PR #29207, opened 2017-10-11
Adding meta data file tree to node api

DeepDiver1975:
## Description
Adds files versions api on php level to the node api

## Related Issue
refs #29087 

## Motivation and Context
Adding a public api for versions on php level

## How Has This Been Tested?
Unit tests have been added

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -461,6 +461,7 @@ class Storage {
 						$versions[$key]['path'] = Filesystem::normalizePath($pathinfo['dirname'] . '/' . $filename);
 						$versions[$key]['name'] = $versionedFile;
 						$versions[$key]['size'] = $view->filesize($dir . '/' . $entryName);
+						$versions[$key]['storage_location'] = "$dir/$entryName";
 					}
 				}
 			}
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -461,6 +461,8 @@ class Storage {
 						$versions[$key]['path'] = Filesystem::normalizePath($pathinfo['dirname'] . '/' . $filename);
 						$versions[$key]['name'] = $versionedFile;
 						$versions[$key]['size'] = $view->filesize($dir . '/' . $entryName);
+						$versions[$key]['timestamp'] = $timestamp;
+						$versions[$key]['etag'] = $view->getETag($dir . '/' . $entryName);
 						$versions[$key]['storage_location'] = "$dir/$entryName";
 					}
 				}
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -464,6 +464,7 @@ class Storage {
 						$versions[$key]['timestamp'] = $timestamp;
 						$versions[$key]['etag'] = $view->getETag($dir . '/' . $entryName);
 						$versions[$key]['storage_location'] = "$dir/$entryName";
+						$versions[$key]['owner'] = $uid;
 					}
 				}
 			}
@@ -833,6 +834,11 @@ class Storage {
 			self::$application = new Application();
 		}
 		return self::$application->getContainer()->query('Expiration');
+	}
+
+	public static function getContentOfVersion($uid, $storage_location) {
+		$users_view = new View('/'.$uid);
+		return $users_view->fopen($storage_location, 'r');
 	}
 
 }
--- a/lib/private/Files/Meta/MetaFileIdNode.php
+++ b/lib/private/Files/Meta/MetaFileIdNode.php
@@ -1,0 +1,170 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+
+namespace OC\Files\Meta;
+
+use OC\Files\Node\AbstractFolder;
+use OCP\Constants;
+use OCP\Files\FileInfo;
+use OCP\Files\NotFoundException;
+
+/**
+ * Class MetaFileIdNode - this class represents the file id part of the meta endpoint
+ *
+ * @package OC\Files\Meta
+ */
+class MetaFileIdNode extends AbstractFolder {
+
+	/** @var int */
+	private $fileId;
+	/** @var MetaRootNode */
+	private $parentNode;
+
+	/**
+	 * MetaFileIdNode constructor.
+	 *
+	 * @param MetaRootNode $parentNode
+	 * @param int $fileId
+	 */
+	public function __construct(MetaRootNode $parentNode, $fileId) {
+		$this->parentNode = $parentNode;
+		$this->fileId = $fileId;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isEncrypted() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isShared() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isMounted() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getDirectoryListing() {
+		return [
+			new MetaVersionCollection($this->fileId)
+		];
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function get($path) {
+		$pieces = explode('/', $path);
+		if($pieces[0] === 'v') {
+			array_shift($pieces);
+			$node = new MetaVersionCollection($this->fileId);
+			if (empty($pieces)) {
+				return $node;
+			}
+			return $node->get(implode('/', $pieces));
+		}
+		throw new NotFoundException();
+
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getFreeSpace() {
+		return FileInfo::SPACE_UNKNOWN;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getPath() {
+		return $this->getInternalPath();
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getInternalPath() {
+		return "/meta/{$this->fileId}";
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getPermissions() {
+		return Constants::PERMISSION_READ;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isReadable() {
+		return true;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isUpdateable() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isDeletable() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isShareable() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getParent() {
+		return $this->parentNode;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getName() {
+		return "{$this->fileId}";
+	}
+
+}
--- a/lib/private/Files/Meta/MetaFileIdNode.php
+++ b/lib/private/Files/Meta/MetaFileIdNode.php
@@ -25,6 +25,7 @@ namespace OC\Files\Meta;
 use OC\Files\Node\AbstractFolder;
 use OCP\Constants;
 use OCP\Files\FileInfo;
+use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 
 /**
@@ -38,6 +39,8 @@ class MetaFileIdNode extends AbstractFolder {
 	private $fileId;
 	/** @var MetaRootNode */
 	private $parentNode;
+	/** @var IRootFolder */
+	private $root;
 
 	/**
 	 * MetaFileIdNode constructor.
@@ -45,9 +48,10 @@ class MetaFileIdNode extends AbstractFolder {
 	 * @param MetaRootNode $parentNode
 	 * @param int $fileId
 	 */
-	public function __construct(MetaRootNode $parentNode, $fileId) {
+	public function __construct(MetaRootNode $parentNode, IRootFolder $root, $fileId) {
 		$this->parentNode = $parentNode;
 		$this->fileId = $fileId;
+		$this->root = $root;
 	}
 
 	/**
@@ -76,7 +80,7 @@ class MetaFileIdNode extends AbstractFolder {
 	 */
 	public function getDirectoryListing() {
 		return [
-			new MetaVersionCollection($this->fileId)
+			new MetaVersionCollection($this->fileId, $this->root)
 		];
 	}
 
@@ -87,7 +91,7 @@ class MetaFileIdNode extends AbstractFolder {
 		$pieces = explode('/', $path);
 		if($pieces[0] === 'v') {
 			array_shift($pieces);
-			$node = new MetaVersionCollection($this->fileId);
+			$node = new MetaVersionCollection($this->fileId, $this->root);
 			if (empty($pieces)) {
 				return $node;
 			}
--- a/lib/private/Files/Meta/MetaFileVersionNode.php
+++ b/lib/private/Files/Meta/MetaFileVersionNode.php
@@ -25,6 +25,7 @@ namespace OC\Files\Meta;
 
 use OC\Files\Node\AbstractFile;
 use OC\Files\Node\File;
+use OCP\Files\IRootFolder;
 use OCP\Files\Storage\IVersionedStorage;
 use OCP\Files\NotPermittedException;
 use OCP\Files\Storage;
@@ -45,34 +46,47 @@ class MetaFileVersionNode extends AbstractFile {
 	private $storage;
 	/** @var string */
 	private $internalPath;
+	/** @var IRootFolder */
+	private $root;
 
 	/**
 	 * MetaFileVersionNode constructor.
 	 *
 	 * @param MetaVersionCollection $parent
+	 * @param IRootFolder $root
 	 * @param string $versionId
 	 * @param Storage $storage
 	 * @param string $internalPath
 	 */
 	public function __construct(MetaVersionCollection $parent,
+								IRootFolder $root,
 								$versionId, Storage $storage, $internalPath) {
 		$this->parent = $parent;
 		$this->versionId = $versionId;
 		$this->storage = $storage;
 		$this->internalPath = $internalPath;
+		$this->root = $root;
 	}
 
+	/**
+	 * @inheritdoc
+	 */
 	public function getName() {
 		return $this->versionId;
 	}
 
+	/**
+	 * @inheritdoc
+	 */
 	public function getContent() {
 		return $this->storage->getContentOfVersion($this->internalPath, $this->versionId);
 	}
 
+	/**
+	 * @inheritdoc
+	 */
 	public function copy($targetPath) {
-		//TODO: inject
-		$target = \OC::$server->getRootFolder()->get($targetPath);
+		$target = $this->root->get($targetPath);
 		if ($target instanceof File && $target->getId() === $this->parent->getId()) {
 			$this->storage->restoreVersion($this->internalPath, $this->versionId);
 			return;
--- a/lib/private/Files/Meta/MetaFileVersionNode.php
+++ b/lib/private/Files/Meta/MetaFileVersionNode.php
@@ -89,7 +89,11 @@ class MetaFileVersionNode extends AbstractFile {
 	 * @inheritdoc
 	 */
 	public function getContent() {
-		return $this->storage->getContentOfVersion($this->internalPath, $this->versionId);
+		$handle = $this->fopen('r+');
+		$data = stream_get_contents($handle);
+		fclose($handle);
+
+		return $data;
 	}
 
 	/**
--- a/lib/private/Files/Meta/MetaFileVersionNode.php
+++ b/lib/private/Files/Meta/MetaFileVersionNode.php
@@ -40,6 +40,8 @@ class MetaFileVersionNode extends AbstractFile {
 
 	/** @var string */
 	private $versionId;
+	/** @var int */
+	private $size;
 	/** @var MetaVersionCollection */
 	private $parent;
 	/** @var IVersionedStorage */
@@ -54,15 +56,16 @@ class MetaFileVersionNode extends AbstractFile {
 	 *
 	 * @param MetaVersionCollection $parent
 	 * @param IRootFolder $root
-	 * @param string $versionId
+	 * @param array $version
 	 * @param Storage $storage
 	 * @param string $internalPath
 	 */
 	public function __construct(MetaVersionCollection $parent,
 								IRootFolder $root,
-								$versionId, Storage $storage, $internalPath) {
+								$version, Storage $storage, $internalPath) {
 		$this->parent = $parent;
-		$this->versionId = $versionId;
+		$this->versionId = $version['version'];
+		$this->size = $version['size'];
 		$this->storage = $storage;
 		$this->internalPath = $internalPath;
 		$this->root = $root;
@@ -73,6 +76,13 @@ class MetaFileVersionNode extends AbstractFile {
 	 */
 	public function getName() {
 		return $this->versionId;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getSize() {
+		return $this->size;
 	}
 
 	/**
--- a/lib/private/Files/Meta/MetaFileVersionNode.php
+++ b/lib/private/Files/Meta/MetaFileVersionNode.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+
+namespace OC\Files\Meta;
+
+
+use OC\Files\Node\AbstractFile;
+use OC\Files\Node\File;
+use OCP\Files\Storage\IVersionedStorage;
+use OCP\Files\NotPermittedException;
+use OCP\Files\Storage;
+
+/**
+ * Class MetaFileVersionNode - this class represents a version of a file in the
+ * meta endpoint
+ *
+ * @package OC\Files\Meta
+ */
+class MetaFileVersionNode extends AbstractFile {
+
+	/** @var string */
+	private $versionId;
+	/** @var MetaVersionCollection */
+	private $parent;
+	/** @var IVersionedStorage */
+	private $storage;
+	/** @var string */
+	private $internalPath;
+
+	/**
+	 * MetaFileVersionNode constructor.
+	 *
+	 * @param MetaVersionCollection $parent
+	 * @param string $versionId
+	 * @param Storage $storage
+	 * @param string $internalPath
+	 */
+	public function __construct(MetaVersionCollection $parent,
+								$versionId, Storage $storage, $internalPath) {
+		$this->parent = $parent;
+		$this->versionId = $versionId;
+		$this->storage = $storage;
+		$this->internalPath = $internalPath;
+	}
+
+	public function getName() {
+		return $this->versionId;
+	}
+
+	public function getContent() {
+		return $this->storage->getContentOfVersion($this->internalPath, $this->versionId);
+	}
+
+	public function copy($targetPath) {
+		//TODO: inject
+		$target = \OC::$server->getRootFolder()->get($targetPath);
+		if ($target instanceof File && $target->getId() === $this->parent->getId()) {
+			$this->storage->restoreVersion($this->internalPath, $this->versionId);
+			return;
+		}
+
+		// for now we only allow restoring of a version
+		throw new NotPermittedException();
+	}
+}
--- a/lib/private/Files/Meta/MetaFileVersionNode.php
+++ b/lib/private/Files/Meta/MetaFileVersionNode.php
@@ -40,8 +40,6 @@ class MetaFileVersionNode extends AbstractFile {
 
 	/** @var string */
 	private $versionId;
-	/** @var int */
-	private $size;
 	/** @var MetaVersionCollection */
 	private $parent;
 	/** @var IVersionedStorage */
@@ -50,6 +48,8 @@ class MetaFileVersionNode extends AbstractFile {
 	private $internalPath;
 	/** @var IRootFolder */
 	private $root;
+	/** @var array */
+	private $versionInfo;
 
 	/**
 	 * MetaFileVersionNode constructor.
@@ -62,10 +62,10 @@ class MetaFileVersionNode extends AbstractFile {
 	 */
 	public function __construct(MetaVersionCollection $parent,
 								IRootFolder $root,
-								$version, Storage $storage, $internalPath) {
+								array $version, Storage $storage, $internalPath) {
 		$this->parent = $parent;
 		$this->versionId = $version['version'];
-		$this->size = $version['size'];
+		$this->versionInfo = $version;
 		$this->storage = $storage;
 		$this->internalPath = $internalPath;
 		$this->root = $root;
@@ -82,7 +82,7 @@ class MetaFileVersionNode extends AbstractFile {
 	 * @inheritdoc
 	 */
 	public function getSize() {
-		return $this->size;
+		return isset($this->versionInfo['size']) ? $this->versionInfo['size'] : null;
 	}
 
 	/**
@@ -104,5 +104,21 @@ class MetaFileVersionNode extends AbstractFile {
 
 		// for now we only allow restoring of a version
 		throw new NotPermittedException();
+	}
+
+	public function getMTime() {
+		return isset($this->versionInfo['timestamp']) ? $this->versionInfo['timestamp'] : null;
+	}
+
+	public function getMimetype() {
+		return isset($this->versionInfo['mime-type']) ? $this->versionInfo['mime-type'] : 'application/octet-stream';
+	}
+
+	public function getEtag() {
+		return isset($this->versionInfo['etag']) ? $this->versionInfo['etag'] : null;
+	}
+
+	public function fopen($mode) {
+		return $this->storage->getContentOfVersion($this->internalPath, $this->versionId);
 	}
 }
--- a/lib/private/Files/Meta/MetaRootNode.php
+++ b/lib/private/Files/Meta/MetaRootNode.php
@@ -1,0 +1,159 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+
+namespace OC\Files\Meta;
+
+
+use OC\Files\Node\AbstractFolder;
+use OCP\Constants;
+use OCP\Files\FileInfo;
+use OCP\Files\NotFoundException;
+
+/**
+ * Class MetaRootNode - this class represents the root node of the meta endpoint
+ *
+ * @package OC\Files\Meta
+ */
+class MetaRootNode extends AbstractFolder {
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isEncrypted() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isShared() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isMounted() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getDirectoryListing() {
+		// TODO: in debug mode we might want to return the list of all fileids
+		return [];
+	}
+
+	/**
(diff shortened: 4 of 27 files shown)


lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	/**
+	 * Get the filename of the file or folder
+	 *
+	 * @return string
+	 * @since 6.0.0
+	 */
+	public function getName() {
+		return "{$this->fileId}";
```
PVince81: fancy way of converting to string...

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,592 @@
...
+	}
+
+	/**
+	 * Get the full mimetype of the file or folder i.e. 'image/png'
+	 *
+	 * @return string
+	 * @since 7.0.0
+	 */
+	public function getMimetype() {
+		// TODO: Implement getMimetype() method.
```
PVince81: it's a "httpd/unix-directory"

lib/private/Files/Meta/MetaRootNode.php
```diff
@@ -0,0 +1,593 @@
...
+	/**
+	 * Create a new file
+	 *
+	 * @param string $path relative path of the new file
+	 * @return \OCP\Files\File
+	 * @throws \OCP\Files\NotPermittedException
+	 * @since 6.0.0
+	 */
+	public function newFile($path) {
+		throw new NotPermittedException();
```
PVince81: I'd suggest implementing a "AbstractNode" or "AbstractFolder" that defaults to "NotPermittedException" for all methods to avoid repeating this in all the meta nodes

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -0,0 +1,580 @@
...
+	}
+
+	/**
+	 * Get the full mimetype of the file or folder i.e. 'image/png'
+	 *
+	 * @return string
+	 * @since 7.0.0
+	 */
+	public function getMimetype() {
+		// TODO: Implement getMimetype() method.
```
PVince81: it's a folder: "httpd/unix-directory"

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -0,0 +1,580 @@
...
+	/**
+	 * Get the modified date of the file or folder as unix timestamp
+	 *
+	 * @return int
+	 * @throws InvalidPathException
+	 * @throws NotFoundException
+	 * @since 6.0.0
+	 */
+	public function getMTime() {
+		// TODO: Implement getMTime() method.
```
PVince81: the collection itself has no mtime, so don't bother

PVince81 (commented):
Looks good so far.

A list of TODO checkboxes would be nice.

I also miss a class `MetaVersionNode` which implements `File` and represents a past version of the file. Its getMTime() method could return the version's timestamp, etc.

PVince81:
Open questions:
- need to add MetaVersionNode
- how to do the "move" of a version to the original location ? would this be done on DAV level instead ?

DeepDiver1975:
> how to do the "move" of a version to the original location ? would this be done on DAV level instead ?

the dav code will call the node api
```php
$version = \OC::$server->getRootFolder()->get("meta/{$info->getId()}/v/123456");
$version->copy("$userId/files/path/to/file.mov");

File: apps/files_versions/lib/Storage.php [hunk 6]
```diff
@@ -461,6 +461,7 @@ public static function getVersions($uid, $filename, $userFullPath = '') {
 						$versions[$key]['path'] = Filesystem::normalizePath($pathinfo['dirname'] . '/' . $filename);
 						$versions[$key]['name'] = $versionedFile;
 						$versions[$key]['size'] = $view->filesize($dir . '/' . $entryName);
+						$versions[$key]['storage_location'] = "$dir/$entryName";
```
PVince81: is that the original path ? is it an absolute path ? please test with external storages to confirm.

Also add a comment because "storage_location" isn't self-explanatory

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,167 @@
...
+
+
+namespace OC\Files\Meta;
+
+
+use OC\Files\Node\AbstractFolder;
+use OCP\Constants;
+use OCP\Files\FileInfo;
+
+class MetaFileIdNode extends AbstractFolder {
```
PVince81: PHPDoc, what is this class?

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,167 @@
...
+	public function isMounted() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getDirectoryListing() {
+		yield new MetaVersionCollection($this->fileId);
+//		yield new MetaPreviewCollection($this->fileId);
```
PVince81: leftover? if not, please add a TODO in a comment to clarify that commenting this out is intentional

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,167 @@
...
+	 */
+	public function isMounted() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getDirectoryListing() {
+		yield new MetaVersionCollection($this->fileId);
```
PVince81: Why yield and not return ? AFAIK we haven't used yield before, do we want to start using it ?

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,167 @@
...
+	 */
+	public function isShared() {
+		return false;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function isMounted() {
+		return false;
```
PVince81: move these to AbstractFolder or AbstractNode ?

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,167 @@
...
+		$pieces = explode('/', $path);
+		if($pieces[0] === 'v') {
+			array_shift($pieces);
+			$node = new MetaVersionCollection($this->fileId);
+			if (empty($pieces)) {
+				return $node;
+			}
+			return $node->get(implode('/', $pieces));
+		}
+		throw new \InvalidArgumentException();
```
PVince81: or NotFoundException ? If this was a Sabre node we'd return "not found" because someone is trying to get a node that doesn't exist, considering that only nodes starting with "v" exist

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	/**
+	 * Get the path of the file or folder relative to the mountpoint of it's storage
+	 *
+	 * @return string
+	 * @since 6.0.0
+	 */
+	public function getInternalPath() {
+		return "/meta/{$this->fileId}";
```
PVince81: prepend the parent node path instead of hard-coding "/meta" ?

lib/private/Files/Meta/MetaFileVersionNode.php
```diff
@@ -0,0 +1,70 @@
...
+namespace OC\Files\Meta;
+
+
+use OC\Files\Node\AbstractFile;
+use OC\Files\Node\File;
+use OC\Files\Storage\IVersionedStorage;
+use OCP\Files\NotPermittedException;
+use OCP\Files\Storage;
+
+class MetaFileVersionNode extends AbstractFile {
```
PVince81: phpdoc

lib/private/Files/Meta/MetaFileVersionNode.php
```diff
@@ -0,0 +1,70 @@
...
+	public function getName() {
+		return $this->versionId;
+	}
+
+	public function getContent() {
+		return $this->storage->getContentOfVersion($this->internalPath, $this->versionId);
+	}
+
+	public function copy($targetPath) {
+		//TODO: inject
```
PVince81: later?

lib/private/Files/Meta/MetaRootNode.php
```diff
@@ -0,0 +1,156 @@
...
+
+
+namespace OC\Files\Meta;
+
+
+use OC\Files\Node\AbstractFolder;
+use OCP\Constants;
+use OCP\Files\FileInfo;
+
+class MetaRootNode extends AbstractFolder {
```
PVince81: phpdoc

lib/private/Files/Meta/MetaRootNode.php
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function get($path) {
+		$pieces = explode('/', $path);
+		$fileId = (int)$pieces[0];
+
+		// TODO: check if file exists
```
PVince81: what happens when accessing a non-existing file currently ? just to evaluate criticality of this TODO

also important: one should not be able to access non-reachable file even if the file exists, for security reasons

The check should be as easy as doing `if (empty($rootNode->getById($fileId)) { die() }` or use the user home folder instead.

What about files in trash, are we already able to access their versions or should we block that use case for this iteration ?

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -0,0 +1,112 @@
...
+use OC\Files\Storage\IVersionedStorage;
+use OC\Files\View;
+use OCP\Files\FileInfo;
+use OCP\Files\Folder;
+use OCP\Files\InvalidPathException;
+use OCP\Files\NotFoundException;
+use OCP\Files\NotPermittedException;
+use OCP\Files\Storage;
+
+class MetaVersionCollection extends AbstractFolder {
```
PVince81: PHPD... you know the drill

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -0,0 +1,112 @@
...
+	 *
+	 * @param string $path relative path of the file or folder
+	 * @return \OCP\Files\Node
+	 * @throws \OCP\Files\NotFoundException
+	 * @since 6.0.0
+	 */
+	public function get($path) {
+		$pieces = explode('/', $path);
+		if (count($pieces) !== 1) {
+			throw new NotFoundException();
```
PVince81: see, here you used NotFoundException. Please use the same in the other comment above instead of InvalidArgumentException

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -0,0 +1,112 @@
...
+		}
+		$versionId = $pieces[0];
+		$view = new View();
+		$path = $view->getPath($this->fileId);
+		/** @var Storage $storage */
+		list($storage, $internalPath) = $view->resolvePath($path);
+		if (!$storage->instanceOfStorage(IVersionedStorage::class)) {
+			throw new NotFoundException();
```
PVince81: We might need to fallback to the old way of doing versions to keep compatible. For example if the file is on SFTP and the storage doesn't support versions, then the versions would be stored on the local storage like it was in former versions. We should plug in those versions here.

lib/private/Files/Node/Root.php
```diff
@@ -365,4 +370,18 @@ public function getUserFolder($userId) {
 		return $folder;
 
 	}
+
+	private function resolveVirtualNode($fullPath) {
+		$pieces = explode('/', $fullPath);
+		if ($pieces[1] !== 'meta') {
+			return null;
+		}
+		array_shift($pieces);
```
PVince81: what's the first piece, is it the user name ?

if yes we might need to read it out and pass it to the MetaRootNode, might be important to check permissions or other stuff

and if yes, it somehow feels like we should add a method `getUserMetaFolder($userId)` on the IRootFolder interface for public use

lib/private/Files/Storage/Common.php
```diff
@@ -684,4 +684,41 @@ public function getAvailability() {
 	public function setAvailability($isAvailable) {
 		$this->getStorageCache()->setAvailability($isAvailable);
 	}
+	public function getVersions($internalPath) {
+		// KISS implementation
+		if (!\OC_App::isEnabled('files_versions')) {
+			return [];
+		}
+		if (strpos($internalPath, 'files/') !== 0) {
```
PVince81: some external storages might extend `Common`, so any files outside of "files/" wouldn't have versions ?

for this one particular check, I suggest extending the "Home" storage class instead which is guaranteed to have the "files" folder

lib/private/Files/Storage/Common.php
```diff
@@ -684,4 +684,41 @@ public function getAvailability() {
 	public function setAvailability($isAvailable) {
 		$this->getStorageCache()->setAvailability($isAvailable);
 	}
+	public function getVersions($internalPath) {
+		// KISS implementation
+		if (!\OC_App::isEnabled('files_versions')) {
```
PVince81: should we make this a trait for reusability ? probably not as the fallback mechanism should be called from within the Node implementation up there

lib/private/Files/Storage/IVersionedStorage.php
```diff
@@ -0,0 +1,36 @@
...
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+
+namespace OC\Files\Storage;
+
+
+interface IVersionedStorage {
```
PVince81: PHPDoc please.

This interface should be public to allow external storages to implement it as well.
Also requires `@since` tags, etc

lib/private/Files/Storage/IVersionedStorage.php
```diff
@@ -0,0 +1,36 @@
...
+
+	public function getVersions($internalPath);
+
+	public function getVersion($internalPath, $versionId);
+
+	public function getContentOfVersion($internalPath, $versionId);
+
+	public function restoreVersion($internalPath, $versionId);
```
PVince81: If the DAV layer is requiring a "MOVE" to restore a version, there will be a target location. We can't expect the issuer of the MOVE operation to be restricted to the original location. Now from what I see, this `restoreVersion` implementation would only restore to the original location. Also we can't provide an absolute path to restore a version to another storage. In this case I wonder if Webdav MOVE is really the right way.

If we do want MOVE, maybe restoreVersion should really just be a fopen on the version somehow and let the reader store the version wherever they want. Or don't provide "restoreVersion" at all but expect people to use:
1) getContentOfVersion() => stream the contents to the target location where restoration must occur (beware of keeping mtime + metadata perhaps?)
2) deleteVersion() to delete the version from the list as it was restored
However the former might not take into account storages that are able to directly restore a version without doing any stream copy (atomic restore).

So I'm not sure currently... We might need to involve the interface after implementing several storage specific version schemes.

In any case, this here is not compatible with expecting a Webdav MOVE to do a version restoration. We need to discuss the latter.

tests/lib/AvatarManagerTest.php
```diff
@@ -113,9 +113,9 @@ public function testGetAvatarValidUserDifferentCasing() {
 			->with('vaLid-USER')
 			->willReturn($user);
 
-		$user->expects($this->once())
-			->method('getUID')
-			->willReturn('valid-user');
+//		$user->expects($this->once())
+//			->method('getUID')
+//			->willReturn('valid-user');
```
PVince81: remove?

PVince81 (changes requested):
Some small changes required.

And also some potential tricky topics:
- restoring a version to original location with storage specific atomic version restore vs Webdav MOVE
- get versions of trashed files by file id, if that even makes sense. Would still be good to have a way to list/access versions of a trashed file.

lib/private/Files/Meta/MetaFileIdNode.php
DeepDiver1975: yes - we want to use yield

lib/private/Files/Meta/MetaVersionCollection.php
DeepDiver1975: the fallback implementation is in class Common - in addition if there is no version support -> we simply have no versions.

lib/private/Files/Storage/Common.php
DeepDiver1975: this implementation will change in a follow up pr

lib/private/Files/Storage/Common.php
DeepDiver1975: file living on external storages also have the logicall path being named /files/ftp/foo/bar.txt - right?

lib/private/Files/Storage/IVersionedStorage.php
DeepDiver1975: It is not a MOVE but a COPY. 

> Now from what I see, this restoreVersion implementation would only restore to the original location. Also we can't provide an absolute path to restore a version to another storage. In this case I wonder if Webdav MOVE is really the right way.

In a second step we can extend the implementation to allow a cross storage move/copy. but this is handled in the meta node level.

this function is really specific to restoring a version of a file within the storage.

lib/private/Files/Meta/MetaFileIdNode.php
DeepDiver1975: The abstract class throw a NotPermittedException ...

lib/private/Files/Meta/MetaFileIdNode.php
DeepDiver1975: It will never live somewhere else ...

lib/private/Files/Node/Root.php
DeepDiver1975: the first piece is an emtpy string - explode of '/meta/1234' -> ['', 'meta', '1234']

lib/private/Files/Meta/MetaFileIdNode.php
PVince81: why ? can you explain the bigger picture ?

PVince81: I do understand this http://cogitolearning.co.uk/2013/06/php-5-5-the-power-to-yield/ but don't understand how it applies to here. There is no reason for this function to continue running after returning the value since there are no further instructions.

lib/private/Files/Meta/MetaVersionCollection.php
PVince81: are you sure we want to break existing behavior at this point in time ? I'd rather have a fallback.

Also note that if we decide that external storages that don't implement IVersionedStorage now have no versions, we'll need a repair step to clean the already existing versions that we stored for these and add an entry to the release notes.

lib/private/Files/Storage/Common.php
PVince81: For the absolute path, yes: "$userId/files/ftpmountpoint/foo/bar.txt".

But here this is about he internal path, it's "foo/bar.txt". There is no "files" folder in general on external storages. Only the home storage has a "files" folder where all the files are in. This difference caused many headaches in the past...

lib/private/Files/Storage/IVersionedStorage.php
PVince81: Okay

lib/private/Files/Meta/MetaVersionCollection.php
DeepDiver1975: I see no breakage in behavior - but yes: all external storages need to be retested with respect to this change

lib/private/Files/Storage/Common.php
DeepDiver1975: I'll test this - THX

lib/private/Files/Storage/IVersionedStorage.php
PVince81: COPY + DELETE actually. The current versions app would delete the version from the list after restoring. we could decide to change the behavior to keep the version in the list though, might be useful.

lib/private/Files/Node/Root.php
PVince81: I'd suggest ltrim to make sure there is really no slash there.

There are some dark corners of the code where slashes are not guaranteed to be there.

lib/private/Files/Meta/MetaVersionCollection.php
PVince81: I think this qualifies as a change of behavior:

1) OC 10.0.3: overwrite file several times on SFTP storage, check versions list: many versions
2) future version: same operations, check versions: list empty

While I understand the motivation of removing support for versions on such storages with no fallback which would simplify our code a lot, I'm worried about user expectation / migration situations.

DeepDiver1975: My goal is clear here: no feature regression! we will have versions for external storages as we did before.

tests/integration/features/bootstrap/WebDav.php
```diff
@@ -488,6 +488,49 @@ public function listFolder($user, $path, $folderDepth, $properties = null) {
...
+
+
+	/**
+	 * @Then the content length of file :path with version index :index for user :user in versions folder is :length
+	 * @param $path
+	 * @param $index
+	 * @param $user
+	 * @param $length
+	 */
+	public function theContentLengthOfFileWithVersionIndexForUserInVersionsFolderIs($path, $index, $user, $length) {
```
SergioBertolinSG: Can we have a shorter name for this function? Behat's default names for functions are not mandatory, we can choose them.

tests/integration/features/bootstrap/WebDav.php
```diff
@@ -488,6 +488,49 @@ public function listFolder($user, $path, $folderDepth, $properties = null) {
...
+	 * @param $count
+	 * @param $user
+	 */
+	public function theVersionFolderOfFileContainsElements($path, $user, $count) {
+		$fileId = $this->getFileIdForPath($user, $path);
+		$elements = $this->listVersionFolder($user, '/meta/'.$fileId.'/v', 1);
+		PHPUnit_Framework_Assert::assertEquals($count, count($elements)-1);
+	}
+
+
```
SergioBertolinSG: Extra new line here.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,30 @@
+Feature: dav-versions
+  Background:
+    Given using api version "2"
+
+  Scenario: Upload file and no version is available
+    Given using new dav path
```
SergioBertolinSG: This can be included in the background section.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,30 @@
...
+    And user "user0" exists
+    And as an "user0"
+    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    Then the version folder of file "/davtest.txt" for user "user0" contains "0" elements
+
+  Scenario: Upload a file twice and versions are available
+   Given using new dav path
```
SergioBertolinSG: This can be included in the background section.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,30 @@
...
+    And user "user0" exists
+    And as an "user0"
+    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
```
SergioBertolinSG: There has to be only one When step, this has to be another And.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,30 @@
...
+    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    Then the version folder of file "/davtest.txt" for user "user0" contains "1" elements
+    Then the content length of file "/davtest.txt" with version index "1" for user "user0" in versions folder is "100"
```
SergioBertolinSG: This Then is an And step.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,30 @@
...
+
+  Scenario: Remove a file
+    Given using new dav path
```
SergioBertolinSG: This can be included in the background section.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,30 @@
...
+    And user "user0" exists
+    And as an "user0"
+    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
```
SergioBertolinSG: Four When steps and two Thens.

It has to be one When and one Then, rest are Ands continuing from first Given.

The structure of every scenario should be:

Given blabla
And blabla
And blabla
...
And blabla
When blabla
**Then blabla**          <- Only one
And blabla
And blabla
...
And blabla

SergioBertolinSG (changes requested):

tests/integration/features/dav-versions.feature
SergioBertolinSG: I think this step should be a (given) And, and the next Then the When step.

tests/integration/features/dav-versions.feature
```diff
@@ -1,30 +1,28 @@
...
-    And user "user0" exists
+    Given user "user0" exists
     And as an "user0"
     When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
-    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
     Then the version folder of file "/davtest.txt" for user "user0" contains "1" elements
     When user "user0" deletes file "/davtest.txt"
-    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
```
SergioBertolinSG: This has to be the only When, the previous steps are all Given steps (And).

SergioBertolinSG (changes requested):

tests/integration/features/dav-versions.feature
SergioBertolinSG: Change this When for And and previous When for And.

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,28 @@
...
+    And the version folder of file "/davtest.txt" for user "user0" contains "1" elements
+    And the content length of file "/davtest.txt" with version index "1" for user "user0" in versions folder is "100"
+
+  Scenario: Remove a file
+    Given user "user0" exists
+    And as an "user0"
+    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    And the version folder of file "/davtest.txt" for user "user0" contains "1" elements
+    When user "user0" deletes file "/davtest.txt"
```
SergioBertolinSG: This When is an And and the next And is the When.

SergioBertolinSG (changes requested):

SergioBertolinSG (approved):
Behat tests are fine 👍

tests/integration/features/dav-versions.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  Scenario: Remove a file
+    Given user "user0" exists
+    And as an "user0"
+    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    And the version folder of file "/davtest.txt" for user "user0" contains "1" elements
+    And user "user0" deletes file "/davtest.txt"
+    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    Then the version folder of file "/davtest.txt" for user "user0" contains "0" elements
```
PVince81: Can we also have a test for restoring a version ? Verify that restoration worked by checking that the version's content is now in the current file

DeepDiver1975: @Peter-Prochaska

PVince81:
Automated tests to add:
- restore version
- access version folder from a share receiver and see that all works fine there as well

PVince81:
- [ ] add automated test with external storage, see https://github.com/owncloud/core/blob/master/tests/integration/features/external-storage.feature#L62 about how to set it up, needs `@local_storage` annotation

PVince81:
- [x] TEST: DAV versions listable as share receiver of file
- [x] TEST: DAV versions listable as share receiver of parent folder
- [ ] BUG: DAV versions not listable as share receiver from external storage

Steps:

1. Enable encryption (likely optional)
1. Login as admin
1. Setup system-wide ext storage "/sftp", apply to only admin user and enable sharing
1. Create folder "/sftp/sha"
1. Create file "/sftp/sha/test.txt" with several versions
1. Share "/sftp/sha" with "user1"
1. Use DOM inspector to find the file id of "test.txt" (data-id)
1. with DAV client, connect to "remote.php/dav/meta/$fileId/v/" as user1

Expected: versions listable
Actual: not listable as user1, only listable as admin

PVince81:
- [ ] I'd like a test for the above too, please (no encryption needed, encryption will be tested automatically after merging)

PVince81:
- [ ] BUG: 500 error when PROPFIND on "/meta" folder. No logs: `curl -D - -u admin:admin -X PROPFIND http://localhost/owncloud/remote.php/meta/

PVince81:
- [ ] add automated test that check doing a PROPFIND on a version fileid that is **not accessible** from the current user => must return 404.

@Peter-Prochaska some negative tests would be nice too, for security ^

PVince81:
- [ ] BUG: need to properly block MOVE operations:

```
dav:/owncloud/remote.php/dav/meta/64/v/> mv 1509353014 ../../../files/admin/
Moving `/owncloud/remote.php/dav/meta/64/v/1509353014' to `/owncloud/remote.php/dav/files/admin/1509353014':  failed:
500 Internal Server Error
```

```
{"reqId":"UT6n3Cs23GcHKRLfY7Qc","level":4,"time":"2017-10-30T09:02:00+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"webdav","method":"MOVE","url":"\/owncloud\/remote.php\/dav\/meta\/64\/v\/1509353021","message":"Exception: {\"Message\":\"HTTP\\\/1.1 500 Error while copying file to target location (copied bytes: 0, expected filesize: -1 )\",\"Exception\":\"Sabre\\\\DAV\\\\Exception\",\"Code\":0,\"Trace\":\"#0 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Connector\\\/Sabre\\\/Directory.php(171): OCA\\\\DAV\\\\Connector\\\\Sabre\\\\File->put(false)\\n#1 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Tree.php(314): OCA\\\\DAV\\\\Connector\\\\Sabre\\\\Directory->createFile('1509353021', false)\\n#2 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Tree.php(130): Sabre\\\\DAV\\\\Tree->copyNode(Object(OCA\\\\DAV\\\\Meta\\\\MetaFile), Object(OCA\\\\DAV\\\\Files\\\\FilesHome), '1509353021')\\n#3 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Tree.php(161): Sabre\\\\DAV\\\\Tree->copy('meta\\\/64\\\/v\\\/15093...', 'files\\\/admin\\\/150...')\\n#4 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/CorePlugin.php(642): Sabre\\\\DAV\\\\Tree->move('meta\\\/64\\\/v\\\/15093...', 'files\\\/admin\\\/150...')\\n#5 [internal function]: Sabre\\\\DAV\\\\CorePlugin->httpMove(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#6 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#7 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:MOVE', Array)\\n#8 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#9 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Server.php(256): Sabre\\\\DAV\\\\Server->exec()\\n#10 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/apps\\\/dav\\\/appinfo\\\/v2\\\/remote.php(31): OCA\\\\DAV\\\\Server->exec()\\n#11 \\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/remote.php(175): require_once('\\\/srv\\\/www\\\/htdocs...')\\n#12 {main}\",\"File\":\"\\\/srv\\\/www\\\/htdocs\\\/owncloud\\\/apps\\\/dav\\\/lib\\\/Connector\\\/Sabre\\\/File.php\",\"Line\":166,\"User\":\"admin\"}"}
```

This should be a 403.

PVince81:
- [ ] BUG: ext storage, download versions are empty

Steps:

1. Setup SFTP ext storage "/sftp"
1. Create file "/sftp/test.txt" + versions with text editor
1. Download file with cadaver's GET: `/owncloud/remote.php/dav/meta/$fileId/v/$timestamp` (replace accordingly)

Expected: file contents is there
Actual: file is empty.

Note that downloading over the web UI works as it uses the old APIs, so the version content does exists. It's the DAV part that is not retrieving it properly.

PVince81:
and this is why I'd like to have automated tests for ext storages ^

DeepDiver1975:
> BUG: need to properly block MOVE operations:

I'm getting 403 - what is the file with fileid 64 ???

bildschirmfoto von 2017-10-30 10-42-17

PVince81:
@DeepDiver1975 a regular file. But I have user-key encryption on, maybe that's causing it.

DeepDiver1975:
> DAV versions not listable as share receiver from external storage

sharing does simply not work 💥

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -222,6 +222,61 @@ public function file_get_contents($path) {
 		return $this->storage->file_get_contents($path);
 	}
 
+	public function getContentOfVersion($internalPath, $versionId) {
+		$v = $this->storage->getVersion($internalPath, $versionId);
+		$data = $this->storage->getContentOfVersion($internalPath, $versionId);
+		if ($this->encryptionManager->isEnabled() !== false) {
+			$encryptionModule = $this->getEncryptionModule($internalPath);
+			$source = fopen('php://memory','r+');
```
PVince81: I'm surprised this doesn't work by simply calling the wrapper's fopen and let it do the key magic for you.

@sharidas ideas ?

DeepDiver1975: hmmm .... like:

```php
$this->fopen($v['storage_location'], 'r'+);

**Comment by DeepDiver1975** [comment]
> @DeepDiver1975 a regular file. But I have user-key encryption on, maybe that's causing it.

this should be fixed now - this was encryption related from my pov

File: lib/private/Files/Storage/Wrapper/Encryption.php (hunk 37 above)
Comment (sharidas) [reply lib/private/Files/Storage/Wrapper/Encryption.php]: @PVince81  Yes that's right. It wont work. The fopen call made in the updated version of this PR should work.

File: apps/files_versions/lib/Storage.php (hunk 6 above)
Comment (PVince81) [reply apps/files_versions/lib/Storage.php]: @DeepDiver1975 ping

File: lib/private/Files/Meta/MetaFileVersionNode.php [hunk 38]
```diff
@@ -89,7 +89,11 @@ public function getSize() {
 	 * @inheritdoc
 	 */
 	public function getContent() {
-		return $this->storage->getContentOfVersion($this->internalPath, $this->versionId);
+		$handle = $this->fopen('r+');
```
PVince81: "r" instead of "r+". from the docs, "r+" means open for update. We only need read access

PVince81:
- [ ] BUG: [minor] am able to `cd` into the meta folder of a fileid pointing to a folder

PVince81:
Bug from https://github.com/owncloud/core/pull/29207#issuecomment-340380269 still exists

PVince81:
Bug from https://github.com/owncloud/core/pull/29207#issuecomment-340382389 still exists. Still getting 500 instead of 403.

PVince81:
Downloading a version from an ext storage file with encryption is still empty: https://github.com/owncloud/core/pull/29207#issuecomment-340383921

Strange, I thought that your changes related to encryption would fix this scenario.

What scenario did you fix and test ?

DeepDiver1975:
> Bug from #29207 (comment) still exists

does an sftp external storage behave different compared to a local external storage? 💥

PVince81:
@DeepDiver1975 SFTP shouldn't behave differently :-/

DeepDiver1975:
let me retest all this  .....

PVince81:
Even with local storage I get empty versions on download.

DeepDiver1975:
> Even with local storage I get empty versions on download.

are you sure you grabbed the latest code?

PVince81:
Yes, already checked
```
commit c35178fbace813cfdc2992642100dca5769e2666 (HEAD -> feature/29087, origin/feature/29087)
Author: Thomas Müller <thomas.mueller@tmit.eu>
Date:   Mon Oct 30 15:16:01 2017 +0100

    Allow decryption of versions when being accessed by getContentOfVersion()

commit 48c4ea5798e0990ba7fbccd482894c687fbb496d
Author: Thomas Müller <thomas.mueller@tmit.eu>
Date:   Mon Oct 30 12:28:29 2017 +0100

    Adding etag, mtime and mimetype to versions
```

Maybe we have some slight env differences.

If you cannot reproduce this I can try and debug it.

DeepDiver1975:
> BUG: [minor] am able to cd into the meta folder of a fileid pointing to a folder

I dont see this as an issue - think about we will add further sub nodes in addition to v - those can be valid for folders - like the discussed redirect link

PVince81:
- [ ] BUG: with encryption enabled, try accessing versions of disabling file: 500 error on the "v" collection: `iles_encryption\\\/OC_DEFAULT_MODULE not found while trying to get owner`

- [ ] BUG: ext storage global mount applied to "admin" group + received share as user1 from admin from said storage + encryption: try downloading version: 404 Not found. Listing the versions work though.

- [ ] BUG: same scenario as above, but try listing versions as "admin" from the ext storage file: 500 error: `test\\\/bacon.txt not found while trying to get owner` (bacon.txt is the file)

PVince81:
Integration tests also fail for me locally, without encryption:
```
    /srv/www/htdocs/owncloud/tests/integration/features/dav-versions.feature:12
    /srv/www/htdocs/owncloud/tests/integration/features/dav-versions.feature:20
```

PVince81:
All bugs I found are not reproducible any more now.

I noticed that if a file is in trashbin, the version collection is empty. This is fine for now. We can look into making trashed versions accessible in a future PR. In general there is no need to access trashed versions from an API standpoint at this stage.

PVince81:
Integration tests pass locally as well, even with encryption. Good job.

Now would be cool to get integration tests for all the reported cases. AFAIK we already have all the necessary behat steps to write the scenarios.

DeepDiver1975:
> Now would be cool to get integration tests for all the reported cases. AFAIK we already have all the necessary behat steps to write the scenarios.

this is done in #29488

DeepDiver1975:
> I noticed that if a file is in trashbin, the version collection is empty. This is fine for now. We can look into making trashed versions accessible in a future PR. In general there is no need to access trashed versions from an API standpoint at this stage.

we anyhow need to work on trashbin api - like /meta/1234/t or what so ever - need to think about this.

PVince81 (approved):
:+1: ok to merge for master then.

Try massaging Jenkins

phil-davis:
@DeepDiver1975 and/or @PVince81 
Is this to be backported?
If so, then #29637 would then also logically get backported.
I ask because when making changes in master to integration tests, chunks of new code here make other (future) backports a little trickier - so if it is to be backported then good.

DeepDiver1975:
@phil-davis there will be a bigger backport parts of all related PRs once we are done ...

phil-davis:
OK - then I will leave other "random integration and unit test refactoring" backporting for after this sort of stuff is backported.

elgrafico:
Just to let you know:    
I had an existing user "meta" and updated to 10.0.9, which gave me a bunch of  "NotPermittedException"s with webdav, gallery and whatnot, preventing access to the users files, shares and calendars.  

I'd recommend communicating that "meta" is a reserved user name since 10.0.9

phil-davis:
Issue #32547 raised. 
Username ``meta`` does not work for me.

Issue #28763, opened 2017-08-22
Upgrade 9.1.6.2 ->  10.0.2.1 null cache when moving avatars outside home (fed shares)


XHAGBARD:
### Steps to reproduce
1. apply repository for 10.0 version
2. yum upgrade and install new version
3. Apply upgrade from the Web and error ocurred
4. Apply upgrade via occ cli and debug:

```
[root@owncloud ~]# sudo -u apache /usr/bin/php /var/www/html/owncloud/occ maintenance:mode --off && sudo -u apache /usr/bin/php /var/www/html/owncloud/occ upgrade -vvv
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
Maintenance mode disabled
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
2017-08-22T09:43:33+02:00 Set log level to debug
2017-08-22T09:43:33+02:00 Turned on maintenance mode
2017-08-22T09:43:33+02:00 Repair step: Repair MySQL database engine
2017-08-22T09:43:33+02:00 Repair step: Repair MySQL collation
2017-08-22T09:43:33+02:00 Repair info: All tables already have the correct collation -> nothing to do
2017-08-22T09:43:33+02:00 Repair step: Repair SQLite autoincrement
2017-08-22T09:43:33+02:00 Repair step: Repair duplicate entries in oc_lucene_status
2017-08-22T09:43:33+02:00 Repair info: lucene_status table does not exist -> nothing to do
2017-08-22T09:43:33+02:00 Repair step: Upgrade app code from the marketplace
2017-08-22T09:43:33+02:00 Repair info: Enabling market app to assist with update
2017-08-22T09:43:33+02:00 Repair info: Using market to update existing apps
2017-08-22T09:43:33+02:00 Repair info: Attempting to update the following existing compatible apps from market: 
2017-08-22T09:43:33+02:00 Repair info: Fetching app from market: activity
2017-08-22T09:43:34+02:00 Repair info: 
2017-08-22T09:43:34+02:00 Repair info: Fetching app from market: calendar
2017-08-22T09:43:34+02:00 Repair info: 
2017-08-22T09:43:34+02:00 Repair info: Fetching app from market: comments
2017-08-22T09:43:35+02:00 Repair info: App (comments) is not known at the marketplace.
[...]

PR #29210, opened 2017-10-11
Handle invalid storage when getting storage root id

PVince81:
## Description
Whenever an invalid storage is present like a remote storage with invalid certificate, the mount point handling code will fail hard because the storage could not be instantiated.

Steps here: https://github.com/owncloud/core/issues/28763#issuecomment-335845129
Analysis here: https://github.com/owncloud/core/issues/28763#issuecomment-335847993

## Related Issue
https://github.com/owncloud/core/issues/28763

## How Has This Been Tested?
See steps

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Mount/MountPoint.php
+++ b/lib/private/Files/Mount/MountPoint.php
@@ -245,9 +245,20 @@ class MountPoint implements IMountPoint {
 	/**
 	 * Get the file id of the root of the storage
 	 *
-	 * @return int
+	 * @return int storage numeric id or -1 in case of invalid storage
 	 */
 	public function getStorageRootId() {
-		return (int)$this->getStorage()->getCache()->getId('');
+		$storage = $this->getStorage();
+		if ($storage === null || $this->invalidStorage) {
+			return -1;
+		}
+
+		$cache = $storage->getCache();
+
+		if ($cache === null) {
+			return -1;
+		}
+
+		return (int)$cache->getId('');
 	}
 }
--- a/lib/public/Files/Mount/IMountPoint.php
+++ b/lib/public/Files/Mount/IMountPoint.php
@@ -98,7 +98,7 @@ interface IMountPoint {
 	/**
 	 * Get the file id of the root of the storage
 	 *
-	 * @return int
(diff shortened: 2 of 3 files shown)


lib/private/Files/Mount/MountPoint.php
```diff
@@ -248,6 +248,11 @@ public function getOptions() {
 	 * @return int
 	 */
 	public function getStorageRootId() {
-		return (int)$this->getStorage()->getCache()->getId('');
+		$storage = $this->getStorage();
+		if ($storage !== null && !$this->invalidStorage) {
+			return (int)$storage->getCache()->getId('');
```
DeepDiver1975: test if cache is null???

PVince81:
@DeepDiver1975 I've adjusted it: now it also checks if the cache is null.

As a bonus, I also added unit tests and squashed.

lib/private/Files/Mount/MountPoint.php
```diff
@@ -248,6 +248,17 @@ public function getOptions() {
...
-		return (int)$this->getStorage()->getCache()->getId('');
+		$storage = $this->getStorage();
+		if ($storage === null || $this->invalidStorage) {
+			return -1;
+		}
+
+		$cache = $storage->getCache();
+
+		if ($cache === null) {
+			return -1;
```
jvillafanez: Is it worthy to provide different value to mark different errors, such as -1 = storage not found, -2 invalid = storage? Use constants?

PVince81: not at this point... this method is only called in one location currently and doesn't care what the error is about

jvillafanez: Ok, fair enough.

Could you add the information in the PHPDoc? Otherwise -1 would be a valid storage id that wouldn't need special handling.

PVince81: I've now adjusted the PHPDoc and also the one from IMountPoint to mention the -1 value

jvillafanez (approved):

PVince81:
stable10: https://github.com/owncloud/core/pull/29278

PR #29217, opened 2017-10-12
fix backend js loading error, issue https://github.com/owncloud/files…

henry42:
## Description
some backend scripts do not load right.

## Related Issue
https://github.com/owncloud/files_external_dropbox/issues/13

## Motivation and Context
In some php version, modification of the array during list() execution (e.g. using list($a, $b) = $b) results in undefined behavior.

### DIFF
--- a/apps/files_external/templates/settings.php
+++ b/apps/files_external/templates/settings.php
@@ -19,11 +19,12 @@
 		$scripts = $backend->getCustomJs();
 		foreach ($scripts as $script) {
 			if (is_array($script)) {
-				list($appName, $script) = $script;
+				list($appName, $tmpScript) = $script;
 			} else {
 				$appName = 'files_external';
+				$tmpScript = $script;
 			}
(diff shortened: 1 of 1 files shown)


apps/files_external/templates/settings.php
```diff
@@ -19,11 +19,12 @@
 		$scripts = $backend->getCustomJs();
 		foreach ($scripts as $script) {
 			if (is_array($script)) {
-				list($appName, $script) = $script;
+				list($appName, $tmpScript) = $script;
```
PVince81: ohhh, so the problem is that we were somehow replacing the foreach variable ? great catch!

PVince81 (approved):
:+1:

PVince81:
@henry42 can you sign the CLA ? See the "Details" link on the "license/cla" row in the test results above

ser163:
Thank you. I have fixed it. I use 10.0.3.3

PR #29227, opened 2017-10-12
log exception

butonic:
Don't swallow exceptions 😞

### DIFF
--- a/lib/private/Encryption/DecryptAll.php
+++ b/lib/private/Encryption/DecryptAll.php
@@ -29,6 +29,7 @@ namespace OC\Encryption;
 use OC\Encryption\Exceptions\DecryptionFailedException;
 use OC\Files\View;
 use \OCP\Encryption\IEncryptionModule;
+use OCP\ILogger;
 use OCP\IUserManager;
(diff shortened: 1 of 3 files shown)


lib/private/Encryption/DecryptAll.php
```diff
@@ -232,6 +238,8 @@ protected function decryptUsersFiles($uid, ProgressBar $progress, $userCount) {
 							}
 						}
 					} catch (\Exception $e) {
+						$this->logger->debug("Exception trying to decrypt file <$path> for user <$uid>");
+						$this->logger->logException($e, ['app' => __CLASS__]);
```
DeepDiver1975: you can add the message to logException call

```php
            $this->logger->logException($e, ['message' => "Exception trying to decrypt file <$path> for user <$uid>", 'app' => __CLASS__]);

DeepDiver1975 (changes requested):

DeepDiver1975:
and test hate you .... ❤️

DeepDiver1975 (approved):

PR #29234, opened 2017-10-13
Added extra steps available and some refactoring

SergioBertolinSG:
@PVince81 The steps you requested.

### DIFF
--- a/tests/integration/features/bootstrap/Provisioning.php
+++ b/tests/integration/features/bootstrap/Provisioning.php
@@ -24,40 +24,59 @@ trait Provisioning {
 	 * @param string $user
 	 */
 	public function assureUserExists($user) {
-		try {
-			$this->userExists($user);
-		} catch (\GuzzleHttp\Exception\ClientException $ex) {
+		if ( !$this->userExists($user) ) {
 			$previous_user = $this->currentUser;
 			$this->currentUser = "admin";
 			$this->creatingTheUser($user);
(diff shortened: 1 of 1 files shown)


tests/integration/features/bootstrap/Provisioning.php
```diff
@@ -24,40 +24,59 @@
...
+	/**
+	 * @Then /^user "([^"]*)" already exists$/
+	 * @param string $user
+	 */
+	public function userAlreadyExists($user) {
+		PHPUnit_Framework_Assert::assertTrue($this->userExists($user));
+	}
+
+	/**
+	 * @Then /^user "([^"]*)" does not already exist$/
```
PVince81: @phil-davis strange wording ? better alternative ?

phil-davis (approved):
LGTM

PVince81:
@SergioBertolinSG thanks

tests/integration/features/bootstrap/Provisioning.php
phil-davis: It sounds a bit strange, but I can't think how else to say it better.
We need to keep "Given user XYZ does not exist" (1) to be the opposite of "Given user XYZ exists" (2) - and (2) creates the user if it is not there, so (1) should delete the user if it exists.

PVince81:
@SergioBertolinSG please backport to stable10

Issue #29075, opened 2017-09-21
Restrict to share only to users/groups within group membership


PVince81:
Currently there's an option called "Restrict users to only share with users in their groups".

This option is limited to restricting with "users" but not "membership groups".

The option should be split into two:
- "Restrict users to only share with users in their groups"
- "Restrict users to only share with groups they are member of"

Estimate: 1-2md.

The change itself should be easy. Adjusting/writing tests might take most of the time.

@pmaier1

PVince81:
@jvillafanez any update on this ?

jvillafanez:
I think this can be closed because the PR and the backport have been merged.

PR #29241, opened 2017-10-13
Share with group membership only

jvillafanez:
## Description
Add an option to allow user to share only with the groups they belong to. If a user isn't a member of a group, that group won't appear as an option to share with it

## Related Issue
https://github.com/owncloud/core/issues/29075

## Motivation and Context
Users can share with any group. This option restricts that behaviour in a similar way the "share with users in their groups" option does.

## How Has This Been Tested?
Tested quickly with the sharing dialog. Other than that, consider this PR as untested for now. New unittests will be added soon.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/lib/Capabilities.php
+++ b/apps/files_sharing/lib/Capabilities.php
@@ -81,6 +81,7 @@ class Capabilities implements ICapability {
 			$res['group_sharing'] = $this->config->getAppValue('core', 'shareapi_allow_group_sharing', 'yes') === 'yes';
 
 			$res['share_with_group_members_only'] = $this->config->getAppValue('core', 'shareapi_only_share_with_group_members', 'yes') === 'yes';
+			$res['share_with_membership_groups_only'] = $this->config->getAppValue('core', 'shareapi_only_share_with_membership_groups', 'yes') === 'yes';
 
 			$user_enumeration = [];
 			$user_enumeration['enabled'] = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'yes') === 'yes';
--- a/apps/files_sharing/lib/Controller/ShareesController.php
+++ b/apps/files_sharing/lib/Controller/ShareesController.php
@@ -74,6 +74,9 @@ class ShareesController extends OCSController  {
 	protected $shareWithGroupOnly = false;
 
 	/** @var bool */
+	protected $shareWithMembershipGroupOnly = false;
+
+	/** @var bool */
 	protected $shareeEnumeration = true;
 
 	/** @var bool */
@@ -240,7 +243,7 @@ class ShareesController extends OCSController  {
 		}
 
 		$userGroups =  [];
-		if (!empty($groups) && ($this->shareWithGroupOnly || $this->shareeEnumerationGroupMembers)) {
+		if (!empty($groups) && ($this->shareWithMembershipGroupOnly || $this->shareeEnumerationGroupMembers)) {
 			// Intersect all the groups that match with the groups this user is a member of
 			$userGroups = $this->groupManager->getUserGroups($this->userSession->getUser(), 'sharing');
 			$userGroups = array_map(function (IGroup $group) { return $group->getGID(); }, $userGroups);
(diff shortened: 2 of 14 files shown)


jvillafanez:
@phil-davis @individual-it could you check if the failing test is related with this PR's changes? I'm not sure if it's "just" a random failure or if it's something that needs to be fixed.

phil-davis:
@jvillafanez I scanned through your code diffs and there is nothing touching the front-end that should make the test fail like it did (cannot even find the "share with" element on the page). So I restarted the job and we will see.

PVince81:
@jvillafanez can you also add integration tests ? Feel free to ask @SergioBertolinSG for some assistance.

jvillafanez:
I'll check what we can add easily during today and tomorrow.

felixboehm:
Need an assignment here, who continues work on this?

PVince81:
Assigned to @jvillafanez

jvillafanez:
I've found a problem with the share API (probably known) that affects to all the sharing tests. The problem is that the response of the sharing API is different depending on the number of results it needs to return.

For only one group available in the results:
```
Array
(
    [exact] => Array
        (
            [users] => Array
                (
                )

            [groups] => Array
                (
                )

            [remotes] => Array
                (
                )

        )

    [users] => Array
        (
            [element] => Array
                (
                    [label] => Sharee1
                    [value] => Array
                        (
                            [shareType] => 0
                            [shareWith] => Sharee1
                        )

                )

        )

    [groups] => Array
        (
            [element] => Array
                (
                    [label] => ShareeGroup
                    [value] => Array
                        (
                            [shareType] => 1
                            [shareWith] => ShareeGroup
                        )

                )

        )

    [remotes] => Array
        (
        )

)
```
If there are several results:
```
Array
(
    [exact] => Array
        (
            [users] => Array
                (
                )

            [groups] => Array
                (
                )

            [remotes] => Array
                (
                )

        )

    [users] => Array
        (
            [element] => Array
                (
                    [label] => Sharee1
                    [value] => Array
                        (
                            [shareType] => 0
                            [shareWith] => Sharee1
                        )

                )

        )

    [groups] => Array
        (
            [element] => Array
                (
                    [0] => Array
                        (
                            [label] => ShareeGroup
                            [value] => Array
                                (
                                    [shareType] => 1
                                    [shareWith] => ShareeGroup
                                )

                        )

                    [1] => Array
                        (
                            [label] => ShareeGroup2
                            [value] => Array
                                (
                                    [shareType] => 1
                                    [shareWith] => ShareeGroup2
                                )

                        )

                )

        )

    [remotes] => Array
        (
        )

)
```
Note that if there is only one element available, that element is returned under the "element" key, but if there are several elements, the "element" key returns an array of elements. This will likely happen with the users, although I haven't checked.

This problem affects the new tests because we need at least 2 groups. In order to do so, we'll need to patch the underlying test code to accept both possibilities so the parsing is properly done (this will error prone since we can't distinguish both options in a clear way, checking if the key is an integer is ugly but there isn't any other easy way).

The best option would be to fix the problem in the sharing API, but that's out of the scope of this PR (it will also need to fix the tests).

jvillafanez:
@SergioBertolinSG could you take over for the integration tests? Note my last comment.

phil-davis:
And (I think) @individual-it put his hand up to add some UI tests for this. There are already UI tests of this sharing area, so it should be just a matter of setting the new system option and then testing that the sharing dropdown now contains just the expected items.

SergioBertolinSG:
@jvillafanez I can take care after finishing the pending tasks I have (which are several). Is that ok? Otherwise you should continue.

jvillafanez:
Waiting for automatic testing.... other than that, this should be ready

SergioBertolinSG (approved):
:+1: for the behat tests

PVince81:
@jvillafanez seems to work fine.

One thing that bothers me is that the behavior will change on update.

To recreate the old behavior, one needs to check both checkboxes.

We should write a migration to automatically tick the second checkbox if the first one was checked on update to avoid surprising admins with a sudden change of behavior. Then the admin can decide to tweak if needed.

@jvillafanez does that make sense ?

jvillafanez:
Migration incomming!

jvillafanez:
Migration added (*not tested yet*)

I don't think we need to check the OC version as long as the migrations are run only once (which is what I expect). Under these circumstances, copy the value from one key to another should do the trick.

PVince81:
@jvillafanez migrations are run only once, yes. Keep in mind that migrations also run at install time so make sure it also works in that situation.

jvillafanez:
* Fresh install -> migration runs without doing anything (key is missing so nothing to do)
* Update from 10.0.3 with checkbox active -> key is duplicated
* Update from 9.1.6 with checkbox active -> key is duplicated

LGTM

PVince81 (approved):
:+1:

PVince81:
@jvillafanez seems codecov is not happy, might be missing some test coverage ?

PVince81:
@jvillafanez I checked the report. Not all is coverable (ignore the migration), please write tests for the changes in ShareesController.

PVince81:

jvillafanez:
stable10 backport in https://github.com/owncloud/core/pull/29391

individual-it:
UI tests in #29406

PR #29251, opened 2017-10-16
[stable10] fix default name string

lordelix:
## Description
Backporting: https://github.com/owncloud/core/pull/29161
Revert back hard-coded name string in core template files

## Related Issue

https://github.com/owncloud/enterprise/issues/2271


PVince81 (approved):
:+1:

PR #29255, opened 2017-10-16
Make admin credentials flexible

SergioBertolinSG:
When adding the functionality of using big fileid numbers, the admin credentials were forced to be admin/admin. This PR solves that.


tests/integration/features/bootstrap/BasicStructure.php
```diff
@@ -406,11 +407,12 @@ public static function removeFilesFromLocalStorageAfter() {
 	/**
 	 * @BeforeSuite
 	 */
-	public static function useBigFileIDs() {
+	public static function useBigFileIDs(BeforeSuiteScope $scope) {
 		$fullUrl = getenv('TEST_SERVER_URL') . "/v1.php/apps/testing/api/v1/increasefileid";
 		$client = new Client();
 		$options = [];
-		$options['auth'] = ['admin','admin'];
+		$adminCredentials = $scope->getSuite()->getSettings()['contexts'][0][__CLASS__]['admin'];
```
jvillafanez: If there isn't a better option we'll have to go with this.

SergioBertolinSG: Approve the review then 😋

phil-davis (approved):
It would be nice to be able to get the admin credentials without such odd-looking means to find the context information. But because we are in BeforeSuite here, the context is not already nicely instantiated. So this is "the best" that can be done without some bigger refactoring to do it every scenario BeforeScenario.

phil-davis:
@SergioBertolinSG I resolved a little conflict online here. So that means CI runs again.

SergioBertolinSG:
@phil-davis thanks.

phil-davis:
Note: this never got backported to ``stable10``
I noticed it today when backporting #30830 I got was surprised by some little conflicts reported.

In any case, the changes in #30830 (and its backport #30842 ) have superceded the lines here. So there is no longer anything from this PR to backport.

PR #29261, opened 2017-10-16
Remove double quotes from boolean values in status.php output

mmattel:
## Description
This PR removes double quotes unintentionally added in the boolean result values from status.php
An QA issue has been raised https://github.com/owncloud/QA/issues/500 (Add scenario to check status.php output)
A documentation release note PR follows asap

## Related Issue
#29228 (Change in status.php from 9.1.x to 10.x)

## How Has This Been Tested?
Access yourdomain/status.php
```
{"installed":true,"maintenance":false,"needsDbUpgrade":true,"version":"10.0.3.3","versionstring":"10.0.3","edition":"Community","productname":"ownCloud"}
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #29228, opened 2017-10-12
Change in status.php from 9.1.x to 10.x


pako81:
### Steps to reproduce
1. Upgrade from 9.1.x to 10.x
2. Check status.php

### Expected behaviour

Values `true`/`false` are now double-quoted. It was on 9.1.x: `"installed":true,"maintenance":true,` but on 10.x is: `"installed":"true","maintenance":"false",`

### Actual behaviour

For consistency reasons this should not have changed. It could indeed create issues with Load Balancers expecting a well given string

### Server configuration
**Operating system**:

Ubuntu 16.04

**Web server:**

Apache 2.4

**Database:**

MariaDB 10

**PHP version:**

7.0.x

**ownCloud version:** (see ownCloud admin page)

10.0.3

**Updated from an older ownCloud or fresh install:**

Updated from 9.1.x

Probably this is the commit introducing the change: https://github.com/owncloud/core/commit/a7173150d6b077d90d31a6123e3e3e1b68238dcc

@mmattel I don't think this change is intented so it could be a potential regression. Could you comment please ?

PVince81:
@mmattel can you clarify whether the change from boolean to strings was intentional ?

This would need updating the release notes and documentation as some people use scripts to parse these values. I know you're the kind of person who always request docs additions and as far as I remember you didn't here, so maybe the change was not intentional ? 😄 

If not intentional, we should add this to the "known issues" in the release notes.

mmattel:
It was not intentional to change the type of the output - I am in sorry for that :anguished:
How to proceed...?

PVince81:
@mmattel thanks for confirming. Can you make a PR to adjust it and another one to the release notes page in the documentation to mention that it's a known issue ?

PVince81:
(don't be alarmed by the word "urgent", it basically just means it's a showstopper for the next release)

PVince81:
@SergioBertolinSG can we have an integration test that simply checks the output of status.php in a way that the data type is asserted as well ?

SergioBertolinSG:
@PVince81 sure.

mmattel:
PR is here: https://github.com/owncloud/core/pull/29261

PR #29230, opened 2017-10-12
Put attributes into generated XML

phil-davis:
## Related Issue
#29229 

## Motivation and Context
Get rid of a warning message.
Make the min and max version show in the XML response.

## How Has This Been Tested?
1) Run the integration test that produced the warning and confirm that the is no warning.
```
~/core/tests/integration$ ./run.sh features/provisioning-v1.feature:299
```
2) Browse to:
```
http://localhost:8080/ocs/v1.php/cloud/apps/comments
```
and see content including:
```
<dependencies><owncloud min-version="10.0" max-version="10.0"/></dependencies>
```

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

If this is the thing to do, then it needs additional integration tests in ``provisioning-v1.feature`` to validate the returned content of the app info XML.

PR #29265, opened 2017-10-17
Added test about status.php

SergioBertolinSG:
This is a test for https://github.com/owncloud/core/issues/29228

it currently fails, requires https://github.com/owncloud/core/pull/29261 .


tests/integration/features/status.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Status
+
+  Scenario: Status.php is correct
+      When requesting status.php
+      Then the json responded should match with
+      """
+      {"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.0.3.3","versionstring":"10.0.3","edition":"Community","productname":"ownCloud"}
```
phil-davis: This will fail each time the ownCloud version number is bumped.
I did a similar sort of test for a version number in returned data in https://github.com/owncloud/core/pull/29230/files - look for "attribute value should be a valid version string" step. If you extract ``version`` and ``versionstring`` from the JSON then you can use ``version_compare()`` to discover if the data is a "reasonable valid" version number representation.

PVince81 (approved):
:+1: great

tests/integration/features/status.feature
SergioBertolinSG: I was thinking about checking the actual values by using occ.

phil-davis: Yes, you could do that also - get those current version strings by some other way and make sure they are the same.

SergioBertolinSG:
Ok some changes, now the version and versionstrings won't require to be changed with any newer version.

To achieve that I've refactored CommandLine trait to include all the code from CommandLineContext, removing it.

There is a variable required called ocPath, and I've added it to the constructor. This will affect all the apps, and will require to add that variable in customgroups and guests at least.

SergioBertolinSG:
Please rereview @PVince81 @phil-davis

tests/integration/features/status.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Status
+
+  Scenario: Status.php is correct
+      When requesting status.php
+      Then the status.php with versions fixed responded should match with
+      """
+      {"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"X.X.X.X","versionstring":"X.X.X","edition":"Community","productname":"ownCloud"}
```
PVince81: having X in the expected version here doesn't feel right.

another approach could be to check keys individually:
"the response contains a key "installed" with value true"

in this case the version string would be checked with only:
"the response contains a key "versionString".

Also note that we have config switches to disable exposing the version, so we should also have scenarios for these

SergioBertolinSG: Those X are changed with the current version of the server before comparing.

SergioBertolinSG: That is what I mean with "versions fixed " in the step name.

PVince81: I know, but to someone non-technical reading the scenarios this is not self-explanatory.

If you do not want to change this to the above proposal, then I suggest changing the strings to "$CURRENT_VERSION_STRING" which most will understand that a substitution will happen there.

SergioBertolinSG: OK, that sounds good.

PVince81:
@SergioBertolinSG please backport

phil-davis:
Backport to stable10 is in PR #29296

PR #29282, opened 2017-10-18
[Central group table] Centr groups and memberships integration

mrow4a:
PR target branch is https://github.com/owncloud/core/pull/29107 - this PR depends on mergint to target branch:

- [x] https://github.com/owncloud/core/pull/29453
- [x] https://github.com/owncloud/core/pull/29154


mrow4a:
Ok, this feature is done - I only have a problem with UserTrait thing used in unit tests. I cannot make it work with membership table in unit tests obviously, since there is no FK on MemoryAccountTable, which mocks real account table
https://github.com/owncloud/core/blob/master/tests/lib/Traits/UserTrait.php

@DeepDiver1975 @PVince81 @butonic

mrow4a:
@butonic, could you have a look on the errors I got with PostgreSQL? It is very strange, and maybe you might have some clue

```22:06:48 Doctrine\DBAL\Driver\PDOException: SQLSTATE[25P02]: In failed sql transaction: 7 ERROR:  current transaction is aborted, commands ignored until end of transaction block```

@DeepDiver1975

**Comment by mrow4a** [comment]
Ok, PR passes tests and is ready to:
1. Merge depending PRs
2. Review the last commit, which does integration with corrected unit tests. 

@DeepDiver1975 @PVince81 @butonic

File: tests/lib/Traits/MembershipTrait.php [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Test\Traits;
+
+use OCP\IConfig;
+use Test\Util\MemoryMembershipManager;
+
+/**
+ * Allow creating users in a temporary backend
+ */
+trait MembershipTrait {
```
DeepDiver1975: make this part of groups? nobody should really work on the membership level

DeepDiver1975 (changes requested):
which is the pr that bring this all together into master?

tests/lib/Traits/MembershipTrait.php
mrow4a: I am not sure what is your intention here

mrow4a:
@DeepDiver1975 
> which is the pr that bring this all together into master?

Please have a look on PR description (1st post)

tests/lib/Traits/MembershipTrait.php
mrow4a: @DeepDiver1975 could you clarify a bit?

mrow4a:
This PR is ready for review @DeepDiver1975 @PVince81 @jvillafanez @butonic @tomneedham - it is quite important, since it is the integration PR.

mrow4a:
Any input here?

tests/lib/Traits/MembershipTrait.php
DeepDiver1975: this trait holds no functionalities a test developer could use - why should this trait be added then.

in addition if I add the group trait I expect that not only groups but also the memberships are taken care of

therefore -> merge this code into group trait

or am I missing a scenario?

mrow4a: I am not sure if this is possible, but I will try.

tomneedham:
11am UK Friday 15th December meet.jit.si/oc-group-table @mrow4a will run through progress, status and how we can help move this forward :)

lib/public/IUser.php
```diff
@@ -33,6 +33,14 @@
  */
 interface IUser {
 
+	/**
+	 * get the user's unique internal id
+	 *
+	 * @return int
+	 * @since 10.0
+	 */
+	public function getID();
```
mrow4a: @DeepDiver1975 requires not to use it

lib/private/User/RemoteUser.php
```diff
@@ -45,6 +45,13 @@ public function __construct($userId) {
 		$this->userId = $userId;
 	}
 
+	/**
+	 * @inheritdoc
+	 */
+	public function getID() {
```
mrow4a: Drop it

mrow4a:
Rebased base branch and number of unit tests exploded, I need some time to adjust all of it, also adding fixes required by Thomas.

lib/private/Group/Database.php
```diff
@@ -44,286 +29,118 @@
  */
 class Database extends \OC\Group\Backend {
```
mrow4a: Now, as we have memberships handled in MembershipManager, and central group table with GroupMapper, this \OC\Group\Backend becomes "external backend" - I also cleared all complex functionalities from it, and now it just stores gid.

apps/provisioning_api/tests/TestCase.php
```diff
@@ -24,50 +25,20 @@
 
 namespace OCA\Provisioning_API\Tests;
 
-use OCP\IGroupManager;
-use OCP\IUser;
-use OCP\IUserManager;
+use Test\Traits\GroupTrait;
+use Test\Traits\UserTrait;
 
 abstract class TestCase extends \Test\TestCase {
```
mrow4a: User and Group Traits are now used for tests. GroupTrait also handles memberships - using MembershipMemoryManager and GroupMemoryMapper

apps/files_sharing/tests/SharedMountTest.php
```diff
@@ -173,7 +173,7 @@ public function testMoveSharedFile() {
 	 * for the other users
 	 */
 	public function testMoveGroupShare () {
-		$g = \OC::$server->getGroupManager()->createGroup('testGroup');
+		$g = $this->createGroup('testGroup');
```
mrow4a: This is GroupTrait specific function (TestCase uses GroupTrait)

lib/private/Group/Group.php
```diff
@@ -144,20 +130,26 @@ public function addUser($user) {
...
-				if ($this->users) {
-					$this->users[$user->getUID()] = $user;
-				}
-				if ($this->emitter) {
-					$this->emitter->emit('\OC\Group', 'postAddUser', [$this, $user]);
-				}
-				return;
-			}
+
+		$account = $this->userManager->getAccountObject($user);
```
mrow4a: Since we cannot get internal user account id, we need to retrieve account for this user using user manager. Account has access to internal id @DeepDiver1975 @tomneedham (This is the fix where user class was exposing ->getId() which it shouldnt. It should be handled by user manager and account. (group manager and backend group)

lib/private/User/Manager.php
```diff
@@ -156,38 +147,61 @@ public function get($uid) {
...
+	 * Retrieve the account object
+	 *
+	 * NOTE: This function is not defined in the interface and is only
+	 * available in the User/Group management scope. Only classes receiving this class instance will
+	 * have access to this function.
+	 *
+	 * @param \OCP\IUser
+	 * @return \OC\User\Account
+	 */
+	public function getAccountObject(\OCP\IUser $user) {
```
mrow4a: This is similar to getUserObject, but instead of getting user from account, this gives account for user. Please mind that this 2 functions are not exposed via interface for programmers. @DeepDiver1975 @tomneedham @jvillafanez

lib/private/User/Manager.php
```diff
@@ -481,11 +497,102 @@ private function newAccount($uid, $backend) {
...
+	}
+
+	/**
+	 * Get backend for given class
+	 *
+	 * @param string $backendClass
+	 * @return UserInterface|null
+	 * @deprecated 10.0.0 - use getBackends of \OCP\IUserManager
+	 */
 	public function getBackend($backendClass) {
```
mrow4a: This is very hacky and one should use getBackends instead.

lib/private/User/Manager.php
```diff
@@ -481,11 +497,102 @@ private function newAccount($uid, $backend) {
...
 	}
 
+	/**
+	 * only used for unit testing
+	 *
+	 * @param AccountMapper $mapper
+	 * @param array $backends
+	 * @return array
+	 */
+	public function reset(AccountMapper $mapper, $backends) {
```
mrow4a: This is used only for unit tests, and as getUserObject and getAccountObject, are not exposed via interface and should be used only in User Management scope

mrow4a:
Merge into centr_groups branch, as discussed on CS3

PR #29274, opened 2017-10-17
[master] Fix OC.webroot to be the path only just like the real code -…

DeepDiver1975:
## Description
Use the correct behavior for OC.webroot in tests.
OC.webroot is a reltative path like /owncloud/ - the tests have been initializing it as absolute path like http://localhost

## Related Issue
fixes #17537

## Motivation and Context
Have proper tests in place

## How Has This Been Tested?
make test-js

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29286, opened 2017-10-18
[stable10] [master] Fix OC.webroot to be the path only just like the …

### LINKED ISSUE #17537: Fix oc_webroot used in JS unit tests
I realized that the dummy "oc_webroot" that is specified in the JS unit tests is wrong.
The current one contains the host name, but in reality it is supposed to only contain the context path, no host.

But changing it directly breaks many tests, so it might take a while to fix.

Also some paths like images might still need an existing path to continue working, because the Karma server serves them.

DeepDiver1975:
…real code - fixes #17537

backport of #29274 to stable10


PVince81 (approved):
:+1:

PR #29292, opened 2017-10-19
fixed exception, if $node is type SimpleCollection

peterprochaska:
## Description
Fixed Exception, if $node is type Sabre\DAV\SimpleCollection

## Motivation and Context
I don't like exceptions

## How Has This Been Tested?
After the change, i can see remote.php/dav/files again, without an exception

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
+++ b/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
@@ -25,6 +25,7 @@ namespace OCA\DAV\DAV;
 
 use Doctrine\DBAL\Connection;
 use OCA\DAV\Connector\Sabre\Directory;
+use OCA\DAV\Connector\Sabre\Node;
 use Sabre\DAV\INode;
 
 /**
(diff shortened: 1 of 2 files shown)


DeepDiver1975 (dismissed):
- [ ] all other methods need the same check
- [ ] please add unit tests - codecov will be red

VicDeo:
To populate the check across all methods it's better to override getNodeForPath from parent
```
	/**
	 * @param string $path
	 * @return INode|null
	 */
	protected function getNodeForPath($path){
           $node = parent::getNodeForPath($path);
		if (!$node instanceof Node) {
			return null;
		}
           return $node;
        }
```

DeepDiver1975:
> To populate the check across all methods it's better to override getNodeForPath from parent

sounds reasonable - please commit to this branch right away - I want this solved asap

VicDeo:
> The diff coverage is 100%.

:muscle:

VicDeo (approved):
👍

PR #29294, opened 2017-10-19
[Stable10] Trying to increase timeout of integration tests because encryption

SergioBertolinSG:
Trying to increment the timeout of jenkins to allow integration tests to complete con stable10 because encryption.


PVince81 (approved):
:+1: if CI passes

PR #29302, opened 2017-10-19
[stable10] Fixed Exception if $node has no method getId()

DeepDiver1975:
backport of #29292 and https://github.com/owncloud/core/pull/29322 to stable10


PVince81 (approved):
:+1: if CI passes

PR #29303, opened 2017-10-19
[stable10] Put attributes into generated XML

DeepDiver1975:
backport of #29230 to stable10

### DIFF
--- a/lib/private/legacy/api.php
+++ b/lib/private/legacy/api.php
@@ -436,7 +436,13 @@ class OC_API {
 	private static function toXML($array, $writer) {
 		foreach($array as $k => $v) {
 			if ($k[0] === '@') {
-				$writer->writeAttribute(substr($k, 1), $v);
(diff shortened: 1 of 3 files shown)


phil-davis:
A Jenkins restart overnight caused Jenkins to fail here (and on many other PRs) so I rebased this to trigger all CI again.
codecov fails. What is the policy with this?

1) for backports - if the code in master was accepted and merged, then it is by definition acceptable as long as the backport includes the same code changes (and tests if any)? If so, then is there any point giving a codecov fail on backports? or any point running codecov on backports?

2) for changes in master - if someone fixes some little bug, then they are responsible to write tests to cover the new/changed code fixing the bug, but if there are no tests existing for the area of code they changed, are they now responsible to create a (new) set of unit tests for the class(es) that they were unlucky enough to have touched?

phil-davis (approved):

PR #29305, opened 2017-10-19
[stable10] Update davclient.js to 0.1.2

DeepDiver1975:
Fixes issues when app devs are using frameworks that might extend the
Array prototype

backport of #28997 to stable10


PVince81:
JS tests passed, merging.

PR #28987, opened 2017-09-12
Disable background scan for home storage/cache

PVince81:
## Description
The home cache might have entries with size=-1 that are outside of
"files/" due to encryption or other metadata that generally do not
require the size to be set or accurate.

Instead of adjusting the HomeCache to ignore entries outside of
"files/", let's just disable it completely. The home storage is in the
data folder and must not receive external changes, so it should never
have any entries with size=-1 in "files/".

## Related Issue
Fixes https://github.com/owncloud/core/issues/28981

## Motivation and Context
See description and https://github.com/owncloud/core/issues/28981#issuecomment-328844338

## How Has This Been Tested?
Steps here https://github.com/owncloud/core/issues/28981#issuecomment-328844338.
Before the fix: integrity check error.
After the fix: deletion worked.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29306, opened 2017-10-19
[stable10] Disable background scan for home storage/cache

DeepDiver1975:
The home cache might have entries with size=-1 that are outside of
"files/" due to encryption or other metadata that generally do not
require the size to be set or accurate.

Instead of adjusting the HomeCache to ignore entries outside of
"files/", let's just disable it completely. The home storage is in the
data folder and must not receive external changes, so it should never
have any entries with size=-1 in "files/".

backport of #28987 to stable10


PVince81 (approved):
:+1:

Issue #29174, opened 2017-10-09
Provide symfony event dispatcher events for all older hook types


PVince81:
1) The very old static hooks: `\OCP\Util::connectHook()`
2) The less old object based emitters: keyword `BasicEmitter` where you listen directly on objects
3) The newer way we want to go for everything in the future: Symfony event dispatcher listeners

For every hook of type 1) and 2), we should provide a Symfony event dispatcher event.

Then all app development needs to use only the Symfony event dispatcher ones in the future and those will be considered public API and the old ones deprecated.

PVince81:
cc @DeepDiver1975

gruentee:
Yes, please!

PVince81:
Once the Symfony events are in place, this would be the next step: https://github.com/owncloud/core/issues/29672

sharidas:
1. The usage of fileinfo in the events, looks bad because if there are no listeners for the event then it would again affect the performance because fileinfo is no longer used.
2. The usage of sending View object and paths as custom symfony event, the main problem here is view becomes public. Not a good idea
3. The usage of sending just paths. Looks kind of ok. The paths received by the subscriber can use the Node API to get the file id. That becomes little more expensive if there are more apps which use the events to get the file id.

PVince81:
So far I only see 3 as a viable solution: basically just pass the path like the old hooks did and let the listener do `\OC::$server->getUserFolder()->get($path)` to get the node and file info. The latter operation can be expensive especially if repeated in multiple places. Getting the fileinfo hits the database every time. There is a potential solution for this here: https://github.com/owncloud/core/pull/28260

PVince81:
Ideal would be to have a list of all hooks here in a table and the name of their matching symfony event. This will be useful to track progress and also could serve as a migration documentation later.

PVince81:
@sharidasan please use the documentation table https://github.com/owncloud/documentation/issues/3738 to map the old events to the new ones, this would help for developers who migrate

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are closing it as outdated. If this is still relevant on a current ownCloud release, please **reopen** (or open a fresh issue) with up-to-date details — thank you!

PR #29329, opened 2017-10-23
Add logout prehooks

sharidas:
This change will help to do actions
in logout prehooks. Any changes to
be made before logout action can be
done here.

## Description

Add logout prehooks

## Motivation and Context

Adding logout prehooks

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -101,6 +101,7 @@ use OC\Files\External\Service\UserGlobalStoragesService;
 use OC\Files\External\Service\GlobalStoragesService;
 use OC\Files\External\Service\DBConfigService;
 use OC\Http\Client\WebdavClientService;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class Server
@@ -321,6 +322,10 @@ class Server extends ServerContainer implements IServerContainer {
 			$userSession->listen('\OC\User', 'postLogin', function ($user, $password) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_login', ['run' => true, 'uid' => $user->getUID(), 'password' => $password]);
+			});
+			$userSession->listen('\OC\User', 'preLogout', function () {
+				$event = new GenericEvent(null, []);
+				\OC::$server->getEventDispatcher()->dispatch('\OC\User\Session::pre_logout', $event);
 			});
 			$userSession->listen('\OC\User', 'logout', function () {
 				\OC_Hook::emit('OC_User', 'logout', []);
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -57,6 +57,7 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Session\Exceptions\SessionNotAvailableException;
 use OCP\Util;
+use Symfony\Component\EventDispatcher\Event;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
@@ -821,8 +822,21 @@ class Session implements IUserSession, Emitter {
 
 	/**
 	 * logout the user from the session
+	 *
+	 * @return bool
 	 */
 	public function logout() {
+
+		$event = new GenericEvent(null, ['cancel' => false]);
+		$eventDispatcher = \OC::$server->getEventDispatcher();
+		$eventDispatcher->dispatch('\OC\User\Session::pre_logout', $event);
+
+		$this->manager->emit('\OC\User', 'preLogout');
+
+		if ($event['cancel'] === true) {
+			return true;
+		}
+
 		$this->manager->emit('\OC\User', 'logout');
 		$user = $this->getUser();
 		if (!is_null($user)) {
--- a/tests/lib/User/SessionTest.php
+++ b/tests/lib/User/SessionTest.php
@@ -947,4 +947,35 @@ class SessionTest extends TestCase {
 		$userSession->updateSessionTokenPassword($password);
 	}
 
+	public function testCancelLogout() {
+		/** @var ISession | \PHPUnit_Framework_MockObject_MockObject $session */
+		$session = $this->createMock(Memory::class);
+		$session->expects($this->once())
+			->method('set')
+			->with('user_id', 'foo');
+
+		/** @var Manager $manager */
+		$manager = $this->createMock(Manager::class);
+
+		/** @var IUser | \PHPUnit_Framework_MockObject_MockObject $user */
+		$user = $this->createMock(IUser::class);
+		$user->expects($this->once())
+			->method('getUID')
+			->will($this->returnValue('foo'));
+
+		$userSession = new Session($manager, $session, $this->timeFactory, $this->tokenProvider, $this->config);
+		$userSession->setUser($user);
+
+
+		$called['cancel'] = false;
+
+		\OC::$server->getEventDispatcher()->addListener('\OC\User\Session::pre_logout', function ($event) use (&$called) {
+			$called['cancel'] = true;
+			$event['cancel'] = $called['cancel'];
+		});
+
+		$this->assertEquals(true, $userSession->logout());
+	}
+
+
 }


lib/private/Server.php
```diff
@@ -322,6 +323,10 @@ public function __construct($webRoot, \OC\Config $config) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_login', ['run' => true, 'uid' => $user->getUID(), 'password' => $password]);
 			});
+			$userSession->listen('\OC\User', 'preLogout', function () {
+				$event = new GenericEvent(null, []);
+				\OC::$server->getEventDispatcher()->dispatch('prelogoutHook', $event);
```
karakayasemi: optional, but we can consider following the naming convention of EventDispatcher.

lib/private/User/Session.php
```diff
@@ -820,10 +820,27 @@ public function loginWithCookie($uid, $currentToken) {
...
+
+	/**
 	 * logout the user from the session
 	 */
 	public function logout() {
-		$this->manager->emit('\OC\User', 'logout');
+		if ($this->preLogout() === true) {
+			return;
+		}
+			$this->manager->emit('\OC\User', 'logout');
```
karakayasemi: indentation

lib/private/Server.php
PVince81: please do

lib/private/User/Session.php
```diff
@@ -823,6 +823,13 @@ public function loginWithCookie($uid, $currentToken) {
 	 * logout the user from the session
 	 */
 	public function logout() {
+		$this->manager->emit('\OC\User', 'preLogout');
+
+		if ($this->getSession()->get('switchedToNewUser') === 'yes') {
```
PVince81: Don't abuse the session to pass around variables. There is an existing way to receive values from a symfony event, just use it. Here's an example: https://github.com/owncloud/guests/blob/master/controller/userscontroller.php#L130

sharidas:
@PVince81  No more sessions. Data handled via Symfony events.

lib/private/User/Session.php
```diff
@@ -823,6 +823,21 @@ public function loginWithCookie($uid, $currentToken) {
 	 * logout the user from the session
 	 */
 	public function logout() {
+		$called = [];
+
+		$eventDispatcher = \OC::$server->getEventDispatcher();
+		$eventDispatcher->addListener('OC\User\Session::logout', function ($event) use (&$called) {
+			if ($event['switchedToNewUser'] === 'true') {
```
PVince81: I think this should be called "cancel".

Also why not use booleans instead of strings containing booleans ?

sharidas:
@PVince81 Updated the new patch by making 'cancel' and used booleans instead of string containing booleans. Thanks for the feedback.

lib/private/User/Session.php
```diff
@@ -823,6 +823,19 @@ public function loginWithCookie($uid, $currentToken) {
 	 * logout the user from the session
 	 */
 	public function logout() {
+		$cancelLogout = false;
+
+		$eventDispatcher = \OC::$server->getEventDispatcher();
+		$eventDispatcher->addListener('OC\User\Session::logout', function ($event) use (&$cancelLogout) {
+			$cancelLogout = $event['cancel'];
```
PVince81: Have you tested this ? You are not even firing the event with `$eventDispatcher`

Please look again at the example I posted and use the same approach

sharidas: Its been fired from https://github.com/owncloud/impersonate/pull/68/files#diff-5ebb5b2aa16f3a5e11cb19da41fe2f7dR77.

PVince81: Ah, that's a misunderstanding. Core must fire the event and apps must listen to it to react, not the other way around.

sharidas:
@PVince81 Updated the change to trigger the event from core. And the impersonate app reacts to it. Thanks for the feedback.

sharidas:
@PVince81 I have used naming convention for the event names. Let me know if this change is ok.

lib/private/Server.php
```diff
@@ -322,6 +323,10 @@ public function __construct($webRoot, \OC\Config $config) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_login', ['run' => true, 'uid' => $user->getUID(), 'password' => $password]);
 			});
+			$userSession->listen('\OC\User', 'preLogout', function () {
+				$event = new GenericEvent(null, []);
+				\OC::$server->getEventDispatcher()->dispatch('pre_logoutHook', $event);
```
karakayasemi: "user.pre_logout_hook" is better i guess. According to the naming convention:
-Use only lowercase letters, numbers, dots (.) and underscores (_);
-Prefix names with a namespace followed by a dot (e.g. order., user.*);
-End names with a verb that indicates what action has been taken (e.g. order.placed).

sharidas: Done

lib/private/Server.php
```diff
@@ -322,6 +323,10 @@ public function __construct($webRoot, \OC\Config $config) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_login', ['run' => true, 'uid' => $user->getUID(), 'password' => $password]);
 			});
+			$userSession->listen('\OC\User', 'preLogout', function () {
+				$event = new GenericEvent(null, []);
+				\OC::$server->getEventDispatcher()->dispatch('user.pre_logout_hook', $event);
```
PVince81: hmmm, is this the official Symfony naming convention ? for all other Symfony events in OC we used a namespaced name, please look it up

PVince81 (commented):
please double check namespace format and add unit tests to make codecov happy.

lib/private/Server.php
karakayasemi: https://symfony.com/doc/current/components/event_dispatcher.html#naming-conventions
Yes, this is the official Symfony naming convention, but we don't have to follow that of course. I checked quickly the other Symfony events in OC. Generally, we used a namespaced name, but 3rdparty libraries are following this convention, like:
https://github.com/owncloud/core/blob/5ce66c4d5f9a895bec41f56f7cc96661eec00ae6/apps/files_external/3rdparty/aws-sdk-php/Aws/Common/Model/MultipartUpload/AbstractTransfer.php#L33
or
https://github.com/owncloud/core/blob/5ce66c4d5f9a895bec41f56f7cc96661eec00ae6/apps/files_external/3rdparty/aws-sdk-php/Guzzle/Plugin/Backoff/BackoffPlugin.php#L19
All the events related to a class should be constant class property and we should use these constants when dispatching an event. IMHO it is cleaner. 
By the way, core have some codes that follow the convention:
https://github.com/owncloud/core/blob/baf07a7755009e76dcb10cf243330070ec41f194/apps/files_sharing/lib/API/Remote.php#L116
or
https://github.com/owncloud/core/blob/master/lib/public/Http/HttpEvents.php#L32

sharidas: For code consistency I am modifying the event name to use namespace.

PVince81: @DeepDiver1975 any comment on this ? I assume we want to continue using the namespace format like we did for loadAdditionalScripts

karakayasemi: AFAIK, we are still in the adaptation phase to EventDispatcher. IMHO, we can add event strings as constant property to the related class, we can track it under #29174 and follow the convention.

lib/private/User/Session.php
```diff
@@ -821,8 +822,21 @@ public function loginWithCookie($uid, $currentToken) {
...
 	/**
 	 * logout the user from the session
+	 *
+	 * @return bool
 	 */
 	public function logout() {
+
+		$event = new GenericEvent(null, ['cancel' => false]);
+		$eventDispatcher = \OC::$server->getEventDispatcher();
+		$eventDispatcher->dispatch('\OC\User\Session::cancel_logout', $event);
```
PVince81: No, the event should be `\OC\User\Session::pre_logout`. We are informing every listener that we are about to logout, not about the cancel.

lib/private/Server.php
```diff
@@ -322,6 +323,10 @@ public function __construct($webRoot, \OC\Config $config) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_login', ['run' => true, 'uid' => $user->getUID(), 'password' => $password]);
 			});
+			$userSession->listen('\OC\User', 'preLogout', function () {
+				$event = new GenericEvent(null, []);
+				\OC::$server->getEventDispatcher()->dispatch('\OC\User\Session::pre_logouthook', $event);
```
PVince81: "pre_logout", remove the "hook" from the name

sharidas:
@PVince81 Hope the code looks better.

PVince81 (approved):
:+1: thanks, please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/29352

PR #29332, opened 2017-10-23
AccountMapper get by email is now case insensitive

PVince81:
## Description
AccountMapper::getByEmail() is now case insensitive.
This is important to be able to detect duplicates through this method.

## Related Issue
Required for https://github.com/owncloud/guests/pull/168

## How Has This Been Tested?
Manual testing guests app + unit tests

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/AccountMapper.php
+++ b/lib/private/User/AccountMapper.php
@@ -108,9 +108,11 @@ class AccountMapper extends Mapper {
 			throw new \InvalidArgumentException('$email must be defined');
 		}
 		$qb = $this->db->getQueryBuilder();
+		// RFC 5321 says that only domain name is case insensitive, but in practice
+		// it's the whole email
 		$qb->select('*')
 			->from($this->getTableName())
-			->where($qb->expr()->eq('email', $qb->createNamedParameter($email)));
+			->where($qb->expr()->eq($qb->createFunction('LOWER(`email`)'), $qb->createFunction('LOWER(' . $qb->createNamedParameter($email) . ')')));
 
 		return $this->findEntities($qb->getSQL(), $qb->getParameters());
 	}
--- a/tests/lib/User/AccountMapperTest.php
+++ b/tests/lib/User/AccountMapperTest.php
@@ -120,11 +120,32 @@ class AccountMapperTest extends TestCase {
 		$this->assertEquals("TestFind2", array_shift($result)->getUserId());
 	}
 
+	public function findByEmailDataProvider() {
+		return [
+			['test3@find.tld'],
+			['Test3@find.tld'],
+			['test3@Find.tld'],
+		];
+	}
+
 	/**
 	 * find by email
+	 *
+	 * @dataProvider findByEmailDataProvider
 	 */
-	public function testFindByEmail() {
-		$result= $this->mapper->find('test3@find.tld');
+	public function testFindByEmail($email) {
+		$result = $this->mapper->find($email);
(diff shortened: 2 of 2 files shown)


mrow4a:
To me, email should be stored lower case in the database to ensure uniqueness

PVince81:
@mrow4a we currently don't have a constraint in the database as it might happen that more than one user have the same email address depending on which backend they come from (guests, LDAP, etc). This is a topic still to be discussed and solved. Currently we don't want to block all duplicates automatically, only the creation of guests with email addresses that already exist in the system.

Looking at the Account and AccountMapper code, it seems that we store the email address as is, not lower-cased...

mrow4a:
So when he searches for user by email, he might see many entries to select, pointing to many different users ? How he will know which one to share with...

PVince81:
@mrow4a not really. The case where multiple user backends are providing the same email addresses is a corner case that we need to look into.

For the regular case there will only be a single email address per user.

The problem we have currently is only with the guests app which makes it possible for any user to invite by email and create an account, which becomes problematic when dealing with duplicates. The guests PR uses the method AccountMapper->getByEmail() to find if there is already an account to decide whether to allow creation or not.

Regardless of your concerns (which are also mine), I think this PR is a step in the right direction.

mrow4a:
I think this makes no sense... if you are inviting person via email, and that person already exists in the system, you should get error that this person already exists in the system. Email is kind'a unique identifier of the person in real world.

mrow4a:
I understand, that the idea is to allow people have multiple user accounts, even if it points to the same email right?

PVince81:
> I understand, that the idea is to allow people have multiple user accounts, with the same email addresses?

No, we do not want to allow that. But currently due to the architecture it could happen for exotic setups. So we cannot add a constraint at this point without thinking about a migration for said exotic setups. This is why we removed the DB constraint to avoid breaking older setups. We'll likely add it back in the future.

lib/private/User/AccountMapper.php
```diff
@@ -108,9 +108,11 @@ public function getByEmail($email) {
 			throw new \InvalidArgumentException('$email must be defined');
 		}
 		$qb = $this->db->getQueryBuilder();
+		// RFC 5321 says that only domain name is case insensitive, but in practice
+		// it's the whole email
 		$qb->select('*')
 			->from($this->getTableName())
-			->where($qb->expr()->eq('email', $qb->createNamedParameter($email)));
+			->where($qb->expr()->eq($qb->createFunction('LOWER(`email`)'), $qb->createNamedParameter(strtolower($email))));
```
DeepDiver1975: we shall not mix the lower functions - both lower operations shall be handled by sql

DeepDiver1975 (changes requested):

butonic:
Even for LDAP there is a prmary email address. But ldap can be configured to use additional login attributes. And for them something like additionalEmail can be multivalue. No need to store them in the oc db unless you want to be able to search by these emails, which is what the additional search terms are for.

PVince81:
@DeepDiver1975 fixed, please recheck. I hope the format is ok (let's see what Oracle says...).

@butonic was this your response to @mrow4a ? I think getByEmail should not use the search terms anyway as those could be truncated.

DeepDiver1975 (approved):

DeepDiver1975:
Jenkis was successfull -> merge

PVince81:
stable10: https://github.com/owncloud/core/pull/29341

PR #28191, opened 2017-06-22
Markdown support for app descriptions ...

DeepDiver1975:
## Description
More and more apps use markdown in the description - we need to support this in the apps settings panel as well

## Screenshots (if appropriate):
bildschirmfoto von 2017-06-22 11-16-05

vs

bildschirmfoto von 2017-06-22 11-18-20

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29333, opened 2017-10-23
[Stable10] Markdown support for app descriptions ...

DeepDiver1975:
backport of #28191 to stable10

### DIFF
--- a/settings/js/admin-apps.js
+++ b/settings/js/admin-apps.js
@@ -18,6 +18,16 @@ Handlebars.registerHelper('level', function() {
 		}
 	}
 });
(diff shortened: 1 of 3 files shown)


settings/js/admin-apps.js
```diff
@@ -18,6 +18,16 @@ Handlebars.registerHelper('level', function() {
 		}
 	}
 });
+Handlebars.registerHelper('md', function(text){
```
PVince81: too bad we don't have a more generic place to register helper currently

PVince81 (approved):
:+1:

Issue #29138, opened 2017-10-03
Make UI Tests run on Edge browser


phil-davis:
It would be useful to have the UI tests capable of running on Microsoft Edge browser with Windows 10.

Putting this issue here so I can document tricks and problems that I find along the way.

phil-davis:
https://github.com/Behat/MinkExtension/pull/292

Edge webDriver ``deleteAllCookies()`` fails. A workaround is to use the ``@insulated`` tag in Behat/Gherkin feature scenarios. Insulated scenarios cause MinkExtension to call ``stopSessions()`` which does a full reset of the browser session, rather than trying to do individual stuff like deleting cookies.

phil-davis:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/5569343/

Edge webDriver ``getText()`` returns extra white space after the text. So, for now, tests need to ``trim()`` the results of ``getText()`` before comparing with the expected text string.

phil-davis:
https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options

SauceLabs information about selecting browser/webDriver... version combinations.

phil-davis:
https://support.saucelabs.com/hc/en-us/articles/225255868-Tips-and-Tricks-for-Using-Selenium-with-Internet-Explorer

Some IE tricks - might be something relevant to Edge also, might not.

phil-davis:
This can happen:
```
PHP  41. Page\FilesPageBasic->waitTillPageIsLoaded() /tmp/ProxyManagerGeneratedProxy__PM__PageFilesPageGenerated2897994410194442686143097669da61.php:244

Fatal error: Call to a member function hasClass() on null in /home/travis/build/[secure]/core/tests/ui/features/lib/FilesPageBasic.php on line 335
```
that code needs to be looked at and made more robust for general use (looks like the problem is waiting to happen and some Edge timing is bringing it out)

phil-davis:
Sometimes after ``iLogout()`` the next login step fails not being able to find the ``username`` filed to fill.

Looks like there needs to be some ``waitTillPageIsLoaded()`` code somewhere there.

That might also fix intermittent fails in federation and sharing tests that switch between UI users with "I logout" and "I login..."

phil-davis:
Somehow we can get to this point on the trashbin page and not find a ``$fileListXpath`` inside which we then want to look for the relevant file row:
```
[...]

PR #29336, opened 2017-10-24
UI tests improvements

phil-davis:
## Description
1) review of UI test code with various format and PHPdoc changes as suggested by the IDE
2) ``FeatureContext`` when checking notification text against the expected text, allow white-space differences - Edge sometimes returns ``getText()`` including white-space, and actually we do not care about that detail.
3) ``BasicStructure`` when logging out, wait for the login page to load, so it is ready if the next step is to login again.
4) ``filesPageBasic`` ``FilesPage`` ``TrashbinPage`` - refactor so that the various file Xpath variables are now provided through methods. Make those methods abstract in the base class so that we are sure that anyone using the base class is forced to implement them.
5) ``getSizeOfFileFolderList()`` - if there is no folder list at all, return a size of 0.
6) ``findFileRowByName()`` - be more careful to wait until the file list is there, and check existence of elements at each step.
7) ``FilesPageBasic`` ``waitTillPageIsLoaded`` - use ``getFileListXpath()`` to find the correct file list element (TrashbinPage needs to find a different ``fileList`` id to FilesPage)
8) ``waitTillPageIsLoaded()`` - for each of these, throw an exception if it times out. It becomes confusing to continue on, because then the next step often gets a "seemingly random" element not found when looking for whatever it wants on the page. For understanding what goes wrong, it will be much easier to throw the exception straight away when we know that the page is not loading.
9) ``SharingDialog`` - be more specific about the xpath to the close button, that avoids accidentally finding some other close button.
10) ``OwncloudPage`` - be generally more careful to find an element first, throw an exception if the element is not found, then click it. That makes errors a bit easier to understand.
11) ``scrollToPosition`` - write "jQuery" instead of just "$" for the script to execute. Sometimes (rarely?) we see "$ is undefined" errors. That is a bit obscure an error to understand. Now it will say "jQuery is undefined". Why this happens at all is still a mystery.

## Related Issue
#29138 Edge browser
The functional changes here have been discovered along the way while getting UI tests running on the Edge browser.

## Motivation and Context
Work towards minimizing UI test failures and getting tests to run on a wider range of browsers.

## How Has This Been Tested?
UI tests have been run in a separate fork on chrome, firefox and IE.

## Types of changes

## Checklist:

### DIFF
--- a/tests/TestHelpers/DownloadHelper.php
+++ b/tests/TestHelpers/DownloadHelper.php
@@ -21,7 +21,8 @@
  */
 namespace TestHelpers;
 
-use GuzzleHttp\Stream\Stream;
+use GuzzleHttp\Message\FutureResponse;
+use GuzzleHttp\Message\ResponseInterface;
 
 /**
  * Helper for Downloads
@@ -31,7 +32,7 @@ use GuzzleHttp\Stream\Stream;
  */
 class DownloadHelper {
 	/**
-	 * 
+	 *
 	 * @param string $baseUrl URL of owncloud
 	 *                        e.g. http://localhost:8080
 	 *                        should include the subfolder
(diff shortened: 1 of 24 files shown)


individual-it (approved):
a bunch of small useful changes
👍 if CI passes

phil-davis:
Backport stable10 #29338

Issue #29167, opened 2017-10-08
Delete: call_user_func_array() expects parameter 1 to be a valid callback


mmattel:
### Steps to reproduce
1. having a federated share, 10.0.2 - 10.0.3
2. deleting a folder from the local W10 client connected via oC client to 10.0.3
3. monitor 10.0.3 oC log activities

### Expected behaviour
No errors should be logged

### Actual behaviour
```
"user":"xyz","app":"PHP","method":"DELETE","url":"\/remote.php\/dav\/files\/xyz\/abc,"message":"call_user_func_array() expects parameter 1 to be a valid callback, class 'OCA\\Files_Sharing\\External\\Storage' does not have a method 'getFullPath' at \/var\/www\/owncloud\/lib\/private\/Files\/Storage\/Wrapper\/Wrapper.php#501"
```

### Server configuration
**Operating system**: Ubuntu 16.04

**Web server:** nginx

**Database:** mysql

**PHP version:** 7.0

**ownCloud version:** 10.0.3

**Updated from an older ownCloud or fresh install:** updated

**Where did you install ownCloud from:** tar

**Signing status (ownCloud 9.0 and above):** no errors

**The content of config/config.php:**
[config_report_20171008.txt](https://github.com/owncloud/core/files/1366796/config_report_20171008.txt)

**List of activated apps:**

```
Enabled:
  - activity: 2.3.6
  - comments: 0.3.0
  - configreport: 0.1.1
  - dav: 0.3.0
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_external_dropbox: 1.0.0
  - files_external_ftp: 0.2.0
  - files_pdfviewer: 0.8.2
  - files_sharing: 0.10.1
  - files_texteditor: 2.2.1
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - files_videoplayer: 0.9.8
  - firstrunwizard: 1.1
  - gallery: 16.0.2
  - market: 0.2.2
  - music: 0.5.2
  - notifications: 0.3.1
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - templateeditor: 0.1
  - updatenotification: 0.2.1
Disabled:
  - encryption
  - external
  - files_antivirus
  - theme-example
  - user_external
```

**Are you using external storage, if yes which one:** smb, federation

**Are you using encryption:** no

[...]

PR #29337, opened 2017-10-24
Call getFullPath only if its defined in storage

sharidas:
Call getFullPath only if its defined in storage.
Else use mount point to resolve the full path.

## Description

This change adds getFullPath method to shared external storage file.

## Related Issue

https://github.com/owncloud/core/issues/29167

## Motivation and Context

Setup to recreate the issue:
1) Create 2 instances of oC. Lets say oC1 and oC2.
2) oC1 is not encrypted and oC2 is encrypted ( tested with masterkey ). oC2 is added as an external storage to oC1.
3) Create a folder in oC2 name `test`
4) Share the folder to an oC1 user `user1` using federated share
5) Now login as `user1` in oC1 and create a file under `test`
6) Try to delete the file. The log file in oC1, can see message mentioned in the issue.
Since oC1 is non encrypted and when file under `test` is being deleted, the https://github.com/owncloud/core/blob/v10.0.3/lib/private/Files/Storage/Wrapper/Encryption.php#L786 code is hit. The reason is its a cross storage move. And getFullPath is defined only in Encryption storage. Its not available in other storages. So during the iteration to find the getFullPath method under storage wrappers, it couldn't find and throws the error. This change helps us to resolve that issue.

## How Has This Been Tested?

Recreated the issue with the steps mentioned above. After applying this change when the file is deleted, the log file doesn't spit the error message.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -782,8 +782,9 @@ class Encryption extends Wrapper {
 		} else {
 			try {
 				$source = $sourceStorage->fopen($sourceInternalPath, 'r');
-				if ($isRename) {
-					$this->sourcePath[$targetInternalPath] = $sourceStorage->getFullPath($sourceInternalPath);
+				if ($isRename && (count($mount) === 1)) {
+					$sourceStorageMountPoint = $mount[0]->getMountPoint();
+					$this->sourcePath[$targetInternalPath] = $sourceStorageMountPoint . '/' . $sourceInternalPath;
 				} else {
 					unset($this->sourcePath[$targetInternalPath]);
 				}


sharidas:
@PVince81 Let me know if you have suggestions to improve this change.

PVince81:
This is just a quick workaround as it adds the method only on one implementation instead of the interface. I'm not sure if we should add this on the interface as it would affect every possible storage implementation out there.

@sharidas please also add a PHP "has_method" (or "is_method"?) check in the encryption code to check if the target storage does have the method before calling. If not, find a way to fallback.

sharidas:

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -783,7 +783,11 @@ private function copyBetweenStorage(Storage $sourceStorage, $sourceInternalPath,
 			try {
 				$source = $sourceStorage->fopen($sourceInternalPath, 'r');
 				if ($isRename) {
-					$this->sourcePath[$targetInternalPath] = $sourceStorage->getFullPath($sourceInternalPath);
+					if (method_exists($sourceStorage, "getFullPath")) {
+						$this->sourcePath[$targetInternalPath] = $sourceStorage->getFullPath($sourceInternalPath);
+					} else {
+						$this->sourcePath[$targetInternalPath] = $mountPoint . '/' . $sourceInternalPath;
```
PVince81: can `$mountPoint` be undefined or is it always set ?

sharidas: With OC instance with encryption enabled:
1. External Storage
       a) SFTP
             Delete a file from SFTP. mountPoint is not empty. 
             Restore the file from trashbin. mountPoint is not empty
       b) ownCloud
             Delete a file from another oC instance ( encryption enabled ). mountPoint is not empty

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -783,7 +783,11 @@ private function copyBetweenStorage(Storage $sourceStorage, $sourceInternalPath,
 			try {
 				$source = $sourceStorage->fopen($sourceInternalPath, 'r');
 				if ($isRename) {
-					$this->sourcePath[$targetInternalPath] = $sourceStorage->getFullPath($sourceInternalPath);
+					if (method_exists($sourceStorage, "getFullPath")) {
```
sharidas: With the tests done other than fed share, I found that if condition is not getting true. The reason I guess is wrapper could be checksum and which doesn't have "getFullPath", so it moves to else.

lib/private/Files/Storage/Wrapper/Encryption.php
sharidas: With the tests done other than fed share, I found that if condition is not getting true. The reason I guess is wrapper could be checksum and which doesn't have "getFullPath", so it moves to else.

PVince81 (approved):
:+1: if CI passes

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -782,8 +782,9 @@ private function copyBetweenStorage(Storage $sourceStorage, $sourceInternalPath,
 		} else {
 			try {
 				$source = $sourceStorage->fopen($sourceInternalPath, 'r');
-				if ($isRename) {
-					$this->sourcePath[$targetInternalPath] = $sourceStorage->getFullPath($sourceInternalPath);
+				if ($isRename && (count($mount) === 1)) {
```
sharidas: This way `$mount` is not used without defining. The path obtained using this method and getFullPath in the scenario of transfer file onwership between users ( with masterkey encryption ), both path remains same.

sharidas:
Backport PR : https://github.com/owncloud/core/pull/29360

Issue #29264, opened 2017-10-17
UI tests - waitTillPageIsLoaded does not work correctly for trashbin


individual-it:
### Steps to reproduce
1. run trashbin UI tests
2. watch selenium output during check if files are in trashbin

### Expected behaviour
test should check if the files are listed in the trashbin after the page is loaded

### Actual behaviour
the tests cycles in `waitTillPageIsLoaded()` till end of timeout even the page get loaded correctly

in selenium output log many rounds of this is shown:
```
10:19:19.785 INFO - Executing: [find elements: By.xpath: //html/.//*[./@id = 'fileList']])
10:19:19.792 INFO - Done: [find elements: By.xpath: //html/.//*[./@id = 'fileList']]
10:19:19.793 INFO - Executing: [find elements: By.xpath: (//html/.//*[./@id = 'fileList'])[1]//a])
10:19:19.799 INFO - Done: [find elements: By.xpath: (//html/.//*[./@id = 'fileList'])[1]//a]
10:19:19.801 INFO - Executing: [find elements: By.xpath: //html/.//*[@id='emptycontent']])
10:19:19.812 INFO - Done: [find elements: By.xpath: //html/.//*[@id='emptycontent']]
10:19:19.814 INFO - Executing: [find element: By.xpath: (//html/.//*[@id='emptycontent'])[1]])
10:19:19.820 INFO - Done: [find element: By.xpath: (//html/.//*[@id='emptycontent'])[1]]
10:19:19.821 INFO - Executing: [execute script: return arguments[0].getAttribute("class"), [[[ChromeDriver: chrome on LINUX (0ea2f3d5a4207f369db8d31e6ed6a80e)] -> xpath: //html/.//*[@id='emptycontent']]]])
```

### Server configuration
**Operating system**:
Debian
**Web server:**
Apache
**Database:**
SQLite
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
10.0.3
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

phil-davis:
@individual-it please check with latest master and see if the trashbin stuff works better for you.
#29343 has been merged and should help/fix this issue.

phil-davis:
@individual-it ping - please check if this is now fixed in master.

individual-it:
confirming that this issue is fixed

PR #29343, opened 2017-10-25
Correctly detect that an empty files page is loaded

phil-davis:
## Description
1) Be more specific about which div the ``emptycontent`` id is contained in (there are lots of ``emptycontent`` divs in different parts of the page)
2) Move check for fileList being visible so it only applies when there are files.
3) Add a core files test creating an empty folder and navigating into it.
4) Add a core trashbin test that checks that an empty delete folder goes to the trashbin as an empty folder
5) For the "name-parts" code, correctly handle the case when there is only 1 part given (it was crashing in that case, we do not use that because it is a bit odd to specify file name parts and then only put 1 part, but may as well fix it)
6) In ``FileRow`` match the ``fileLinkXpath`` differently so that it works on TrashbinPage as well as FilesPage.

## Related Issue
#29138 
(this PR is a "downstream" thing that was noticed while getting Edge tests running)

## Motivation and Context
When a folder is empty, the UI test code is not correctly detecting that the (empty) files page for that folder is loaded. The "wait till page is loaded" logic was timing out (10 seconds). Now that we throw an exception on page load timeout, this was causing test fails in the ``files_texteditor`` tests, because those tests navigated into an empty folder and then created a file in the folder.

There were no core UI tests that navigated into an empty folder. We should have at least one such test so that the empty files page logic is exercised in core testing.

## How Has This Been Tested?
Run ``files_texteditor`` UI tests locally and confirm they now pass.'
Run the new core test locally and confirm it fails before the change, and passes with the changed code.
Travis will do a full test run for us.

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/files/files.feature
+++ b/tests/ui/features/files/files.feature
@@ -20,3 +20,11 @@ Feature: files
 		|'सिमप्ले फोल्देर $%#?&@'|
 		|'"somequotes1"'|
 		|"'somequotes2'"|
+
+	Scenario: Create a folder inside another folder
+		When I create a folder with the name "top-folder"
+		And I open the folder "top-folder"
+		And I create a folder with the name "sub-folder"
+		Then the folder "sub-folder" should be listed
+		And the files page is reloaded
+		Then the folder "sub-folder" should be listed
--- a/tests/ui/features/files/files.feature
+++ b/tests/ui/features/files/files.feature
@@ -11,20 +11,27 @@ Feature: files
 		Then the filesactionmenu should be completely visible after clicking on it
 
 	Scenario Outline: Create a folder using special characters
-		When I create a folder with the name <folder_name>
-		Then the folder <folder_name> should be listed
+		When I create a folder with the following name
+			| name-parts      |
+			| <folder_name>   |
+		Then the following folder should be listed
+			| name-parts      |
+			| <folder_name>   |
 		And the files page is reloaded
-		Then the folder <folder_name> should be listed
+		Then the following folder should be listed
+			| name-parts      |
+			| <folder_name>   |
 		Examples:
 		|folder_name    |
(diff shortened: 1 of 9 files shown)


individual-it:
does this solve #29264 ?

phil-davis:
Yes - I did not even know about that issue.
I am also adding some tests to navigate into a folder in the trashbin, which will exercise the empty file list in the trashbin page case. Another commit will come...

individual-it (approved):

phil-davis:
@individual-it please check the diffs again. I added core tests and a lot of associated refactoring.

tests/ui/features/files/files.feature
```diff
@@ -19,4 +19,13 @@ Feature: files
...
 		|'सिमप्ले फोल्देर $%#?&@'|
 		|'"somequotes1"'|
-		|"'somequotes2'"|
\ No newline at end of file
+		|"'somequotes2'"|
+
+	Scenario: Create a folder inside another folder
+		When I create a folder with the name "top-folder"
+		And I open the folder "top-folder"
+		Then there are no files/folders listed
```
phil-davis: This step is a kind-of obvious test! But it does exercise the step code on a files-page file-folder list.
(the trashbin test added further down exercises the same step code when on a trashbin page)

individual-it (approved):

individual-it (changes requested):
screenshot-2017-10-25 job 13786 2 - owncloud core - travis ci

phil-davis:
@individual-it please review again. Then I can squash the last "fixup things that I broke" commit into the 2nd-last commit and let CI run again.

individual-it (approved):

phil-davis:
Backport stable10 in #29353

PR #29356, opened 2017-10-26
Add regex for single quotes to UI tests

phil-davis:
## Description
1) Enhance regex for steps that currently look for a file/folder name in double-quotes so that they accept it in single or double quotes.
2) Revert core gherkin test step detail changed recently, putting it back like it was, since now the steps with enhanced regex cope with either double or single-quoted file/folder names.

## Motivation and Context
Changes to step ``theFileFolderShouldBeListed`` meant that the file/folder name had to now always be passed in double-quotes. This broke some ``files_texteditor`` tests. It would be nice to maintain the behavior of this step so that app UI tests continue to work without needing to be touched.

``files_texteditor`` tests happen to have exactly the extra bit of regex needed to match either double or single quoted file/folder names. So it is easy to apply an example from there to core code.

## How Has This Been Tested?
Local UI test runs. Travis will confirm.

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/bootstrap/FilesContext.php
+++ b/tests/ui/features/bootstrap/FilesContext.php
@@ -89,8 +89,18 @@ class FilesContext extends RawMinkContext implements Context {
 	}
 
 	/**
-	 * @When I create a folder with the name :name
+	 * @When /^I create a folder with the name ((?:'[^']*')|(?:"[^"]*"))$/
 	 *
+	 * @param string|array $name enclosed in single or double quotes
+	 * @return void
+	 */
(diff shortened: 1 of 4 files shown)


individual-it (approved):
👍 if CI passes

individual-it:
I guess that extended regex should be used also in other places where we use file/folder names

phil-davis:
There are some places in sharing where we a file/folder in double-quotes is matched. Yes, we could expand those also to allow the file/folder in single quotes.

phil-davis:
Backport stable10 in #29359

PR #29366, opened 2017-10-26
Replace notify user for local shares with button

### LINKED ISSUE #29318: Change "notify by mail" checkbox to a button with user feedback
For internal sharing we have an option (configurable in sharing settings) to notify the recipient via mail. This currently is a checkbox which is total rubbish in the first place as it's not a value "true/false" but a trigger. Therefore we have to

- [ ] Move from checkbox to button
- [ ] Provide user feedback that the mail was sent (not sure about the logic behind that but I would say that we just reuse how it is done when you e.g. trigger a test mail for mail settings or set your mail in personal settings)
- [ ] Optionally the button would be greyed-out after clicking it until the user opens the sharing dialog again (not required)

We already have spam protection in the mailer logic, right?

Basic mockup (could be implemented a little more beautiful :) )

notify

@PVince81 @felixheidecke

PVince81:
## Description
Replace the checkbox with a button.
Changed logic to only send the mail_send flag if email could be sent to at least one recipient.

## Related Issue
Fixes https://github.com/owncloud/core/issues/29318

## Motivation and Context
See ticket

## How Has This Been Tested?
Manual testing so far

## Types of changes

## Checklist:

## TODOs:
- [ ] BUG: fix spinner position
- [ ] TODO: add/adjust unit tests
- [ ] TODO: fading tickmark on success?!

### DIFF
--- a/core/ajax/share.php
+++ b/core/ajax/share.php
@@ -131,7 +131,11 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 			);
 			$result = $mailNotification->sendInternalShareMail($recipientList, $itemSource, $itemType);
 
-			\OCP\Share::setSendMailStatus($itemType, $itemSource, $shareType, $recipient, true);
+			// if we were able to send to at least one recipient, mark as sent
+			// allowing the user to resend would spam users who already got a notification
+			if (count($result) < count($recipientList)) {
+				\OCP\Share::setSendMailStatus($itemType, $itemSource, $shareType, $recipient, true);
+			}
 
 			if (empty($result)) {
 				OCP\JSON::success();
--- a/apps/files_sharing/css/sharetabview.css
+++ b/apps/files_sharing/css/sharetabview.css
@@ -101,6 +101,12 @@
   margin-top: 20px;
 }
 
+.shareTabView .shareWithList .mailNotificationSpinner {
+	position: relative;
+	vertical-align: middle;
+	margin-right: 10px;
+}
+
 .subTabHeaders .tabHeaders {
   margin-left: 0px;
   margin-right: 0px;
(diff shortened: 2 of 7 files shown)


PVince81:
note: cannot increase code coverage for ajax/share.php... we need to refactor that to app framework eventually if we don't manage to kill this checkbox eventually...

core/ajax/share.php
```diff
@@ -131,7 +131,11 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
 			);
 			$result = $mailNotification->sendInternalShareMail($recipientList, $itemSource, $itemType);
 
-			\OCP\Share::setSendMailStatus($itemType, $itemSource, $shareType, $recipient, true);
+			// if we were able to send to at least one recipient, mark as sent
+			// allowing the user to resend would spam users who already got a notification
+			if (count($result) < count($recipientList)) {
```
jvillafanez: `count($result)` could be 0. Do we still want to set the status as sent in this scenario?

PVince81: Yes, because 0 means no errors. The array `$result` only contains a list of email addresses for which an email could **not** be sent.

felixheidecke:
Well … this is still ugly a. f. but simply repositioning the button won't help.
I'll go ahead and open an issue for visual improvement.

lordelix (approved):
It's on the brink of not being okay from a UI perspective. Will be addressed in a different issue.

PVince81:
increase coverage a tiny bit more.

will not bother about the ajax/* case

PVince81:
@jvillafanez your seal of approval for the PHP code ?

PVince81:
ignoring missing coverage for ajax/* file

PVince81:
stable10: https://github.com/owncloud/core/pull/29463

PR #29372, opened 2017-10-27
[Stable10] Splitting integration tests stage into three

SergioBertolinSG:
Splitting integration tests stage into three stages to avoid timeouts and be able to run encryption tests.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -142,24 +142,32 @@ timestampedNode('SLAVE') {
 		   '''
 		}
 
+	stage 'Integration Testing Encrypted'
 		if (isOnReleaseBranch()) {
-			executeAndReport('tests/integration/output/*.xml') {
-				sh '''phpenv local 7.0
-				rm -rf config/config.php data/*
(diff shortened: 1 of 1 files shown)


PVince81:
Thanks, as discussed. @DeepDiver1975 can you confirm that you agree ?

PVince81:
spectacle h16501

@SergioBertolinSG can you put the "stage" part inside the "if", if the language allows ? This way we will not see these misleading 1 second result blocks.

SergioBertolinSG:
@PVince81 I have tried to use "when". Lets see what happens. 

Maybe we should use this https://jenkins.io/doc/book/pipeline/syntax/#built-in-conditions

To check the current branch, not sure if the way of checking it matters.

SergioBertolinSG:
Weird, it doesn't detect when as a directive:

java.lang.NoSuchMethodError: No such DSL method 'when' found among steps

SergioBertolinSG:
@PVince81 rollback, no declarative pipelines will be added.

PVince81 (approved):
:+1:

Issue #16162, opened 2015-05-07
Allow remote-delta on sync files


gadLinux:
Hi, 
I want to fix this improvement ASAP. 
https://github.com/owncloud/client/issues/179

It relates to the remote-delta implementation on client. But I suppose it must be supported on both sides. I saw that latest version of csync inludes a module for owncloud but it differs on current implementation of client on github.

in fact I saw that's using a special module that's not on the normal csync library that's called httpbf. It seems for me that's sending the whole file as PUT http?

So my question is. If you will want to implement rsync how will you implement it?

I was thinking about clone a module on csync and make it transfer using remote-delta (surely using rsync lib) instead of sending the whole file in chunks.

What do you think? Any suggestions?

Using httpbf. I can detect what's changed and send only the blocks that are changed. But this would mean extra work for nothing. Because librsync already implements this kind of processing we are not going to do ti twice.

Best regards,

jospoortvliet:
@danimo @dragotin @ogoffart @guruz could you guys give some input here?

ogoffart:
Hi @gadLinux ,  thank you for trying to help on this issue.

The client is no longer using httpf. And is using only very few parts of csync.

You need to contribute in the client repository. (and in the core)

You will find the code that uploads files in `src/libsync/propagateupload.cpp` here you will find the code that split the file in chunks and uploads them.
Similarily, the code that download the files is in `src/libsync/propagatedownload.cpp`   (this is not using chunks)

Good luck :-)
But be warned that i think it is a difficult issue. I don't really know if it is possible to do that on top of webdav so you will probably need some kind of extensions to the protocol.

dragotin:
Some brainstorming:

There needs to be a list of blocks with its according checksums stored on the server for every file. It will be accessible through a special route on the server API. 

The client will need to fetch the list once it detected that an existing file has changed and is to be uploaded. Once the blocklist has arrived, the client needs to recalculate the blocklist on the new, changed file. After that, the client can compare the lists and identify the blocks that have a changed checksum. These are the blocks that need to be uploaded. 

To upload only parts of files, maybe the HTTP PATCH command (<a href="http://tools.ietf.org/html/rfc5789">RFC</a>) helps. The server needs to be able to handle this command and reassemble the whole file.

[...]

PR #29378, opened 2017-10-27
Implementation of delta-sync support on server-side.

ahmedammar:
## Description
This commit adds server-side support for delta-sync, this extends the core files app.

The basic approach is to store zsync metadata files in a folder called `files_zsync/` which mirrors the structure and layout of the `files/` folder but appends `.zsync` to the metadata files. These files can be requested by the client via a new route `api/v1/zsync/$path`.

Filesystem hooks are used to mirror any `move/copy/delete` operation on the base file onto the metadata file.

The upload path is implemented by modifying the `ChunkingPlugin`, the chunk files are now assumed to be named as the offsets into the original file. The implemenation copies adds a new class `AssemblyStreamZsync` which extends `AssemblyStrem` with support to fill in the missing data in between chunk offsets from a `backingFile`.

A new `zsync` capability is added to the files app, which can be checked by the client to know if delta-sync is supported or not.

## Related Issue

owncloud/core#16162.

## How Has This Been Tested?

Mostly tested by using the owncloud command line client and macOS application. With some randomly generated files using `dd` and various modifications to those files:
```
dd if=/dev/urandom of=1g.rand.img bs=$[1024*1024] count=1024

cp 1g.rand.img 1g.rand.add300m.img
dd if=/dev/urandom of=1g.rand.add300m.img bs=$[1024*1024] count=300 oseek=1024

cp 1g.rand.img 1g.rand.mod200m.img
dd if=/dev/urandom of=1g.rand.mod200m.img bs=$[1024*1024] count=100 oseek=123 conv=notrunc 
dd if=/dev/urandom of=1g.rand.mod200m.img bs=$[1024*1024] count=100 oseek=532 conv=notrunc

cp 1g.rand.img 1g.rand.mov200m.img
dd if=1g.rand.img of=1g.rand.mov200m.img bs=$[1024*1024] count=200 iseek=118 oseek=418 conv=notrunc

cp 1g.rand.img 1g.rand.del200m.img
dd if=1g.rand.img of=1g.rand.del200m.img bs=$[1024*1024] count=622
dd if=1g.rand.img of=1g.rand.del200m.img bs=$[1024*1024] iseek=822 oseek=622
```

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Server.php
+++ b/apps/dav/lib/Server.php
@@ -142,7 +142,6 @@ class Server {
 		));
 
 		$this->server->addPlugin(new CopyEtagHeaderPlugin());
-		$this->server->addPlugin(new ChunkingPlugin());
 
 		// Some WebDAV clients do require Class 2 WebDAV support (locking), since
 		// we do not provide locking we emulate it using a fake locking plugin.
@@ -164,6 +163,14 @@ class Server {
 			$userSession = \OC::$server->getUserSession();
 			$user = $userSession->getUser();
 			if (!is_null($user)) {
+				$rootFolder =  \OC::$server->getRootFolder();
+				$this->server->addPlugin(
+					new ChunkingPlugin(
+						$rootFolder->get($user->getUID()),
+						$user->getUID()
+					)
+				);
+
 				$view = \OC\Files\Filesystem::getView();
 				$this->server->addPlugin(
 					new FilesPlugin(
--- a/apps/dav/lib/Upload/AssemblyStream.php
+++ b/apps/dav/lib/Upload/AssemblyStream.php
@@ -37,22 +37,22 @@ use Sabre\DAV\IFile;
 class AssemblyStream implements \Icewind\Streams\File {
 
 	/** @var resource */
-	private $context;
+	protected $context;
 
 	/** @var IFile[] */
-	private $nodes;
+	protected $nodes;
 
 	/** @var int */
-	private $pos = 0;
+	protected $pos = 0;
 
 	/** @var array */
-	private $sortedNodes;
+	protected $sortedNodes;
 
 	/** @var int */
-	private $size;
+	protected $size;
 
 	/** @var resource */
-	private $currentStream = null;
+	protected $currentStream = null;
 
 	/**
 	 * @param string $path
@@ -78,6 +78,9 @@ class AssemblyStream implements \Icewind\Streams\File {
 		foreach($this->nodes as $node) {
 			$size = $node->getSize();
 			$name = $node->getName();
+			// ignore .zsync metadata file
+			if (!strcmp($name,".zsync"))
+				continue;
 			$this->sortedNodes[$name] = ['node' => $node, 'start' => $start, 'end' => $start + $size];
 			$start += $size;
 			$this->size = $start;
@@ -233,7 +236,7 @@ class AssemblyStream implements \Icewind\Streams\File {
 	 *
 	 * @throws \BadMethodCallException
 	 */
-	public static function wrap(array $nodes) {
+	public static function wrap(array $nodes, IFile $backingFile = null, $length = null) {
 		$context = stream_context_create([
 			'assembly' => [
 				'nodes' => $nodes]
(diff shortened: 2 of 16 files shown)


jvillafanez:
Question to everyone involved: as far as I know, this is another chunking algorithm (or at least it involves modifying it); why isn't this in another app that can be enabled / disabled and why there isn't a switch to choose the algorithm?

I understand that there will be changes in core to support this, but those changes should be oriented to support more chunking algorithms. The algorithm itself should be moved to an application.

(I haven't been in the discussions so there might be important things that I missed)

ahmedammar:
@jvillafanez well I'm here trying to start a discussion :) 

I assumed that delta-sync should be part of the core server implementation and not a separate app. If that isn't the case, then it can be modified. 

The changes to the chunking algorithm are very minimal and don't break any previous client code.

jvillafanez:
Let's suppose we merge this PR now. What "business value" add this feature to the product and who can benefit from it now?

From my point of view, only the desktop client can benefit from this PR, which I guess it will help for faster syncing, but mobile clients nor the web UI will gain anything. In fact they might perform slower. Note that we also need to take into account external storages and where those ".zsync" files will be created.

This is why I think it's better to move it to an app so the admin can choose what he want to do.

ahmedammar:
Well I don't agree to be honest, only desktop will potentially benefit instantly, but nothing is in the way of a mobile implementation too.

Why would anything perform slower because of this PR? I fail to understand that at all.

Nothing in the PR affects existing code paths or breaks any legacy code. Without a supported client nothing changes.

ahmedammar:
I also fail to see how an 'app' can modify the ChunkingPlugin or even override it.

jvillafanez:
My point is that mobiles don't have this support implemented now, and it will take time to implement it (if ever, due to priorities). Meanwhile, the client will just upload the files as they have been doing all this time.

> Why would anything perform slower because of this PR?

In the ChunkingPlugin file, `performMove` function you've wrapped the "old" code `$this->server->tree->move($path, $destination);` with your new code. This will increase the cost of the regular uploads without benefit except for the desktop client (yes, mobiles can implement whatever is needed to gain performance with these changes, but meanwhile it's like this). How much cost is added there? Without measuring, I can't say for sure. Maybe 1 ms, 100 ms, 1 sec... maybe I'm raising unneeded alarms (I haven't reviewed the code in depth so maybe it just adds a couple of ms and it's nothing to worry about).

> The final TODO (if you looked at the related issue) is related to disabling for external storage.

Disabling or switching? Chunking is still needed, which reinforces the idea of 2 different implementations.

> I also fail to see how an 'app' can modify the ChunkingPlugin or even override it. So it may as well be part of the existing files app.

Probably it isn't possible at the moment, but this support can be added somehow.

ahmedammar:
> In the ChunkingPlugin file, performMove function you've wrapped the "old" code $this->server->tree->move($path, $destination); with your new code

I've re-ordered the code so it's a bit clearer, it's basically a string modification and two if statements. For anyone not using zsync based chunking that should be a few ms maximum. I'd be surprised otherwise. the regex string replace could be replaced with something faster if that's *really* a performance issue.

https://github.com/ahmedammar/core/commit/c7b22e5c430d52af9c2e83c1c88375e29598dddc#diff-a3fb8f83598c3f5e1d514a34ca578b1eR86

> Disabling or switching? Chunking is still needed.

No one said anything about disabling Chunking at all ... Disabling zsync metadata upload and fetch for external storage is what I meant.

> Probably it isn't possible at the moment, but this support can be added somehow.

I personally find it hard to believe that multiple chunking algorithms/plugins is even something with enough developer need. The one **important** reason this modification of ChunkingPlugin needed to happen for zsync metadata is to get around the fact that there is no persistent file locking available in owncloud. If the .zsync is not copied in the same MOVE request as the base file then this leaves the door open for issues with multiple uploaders.

ahmedammar:
I closed this because I accidentally deleted my repo on GitHub 😞 

I opened a new PR#29404.

@jvillafanez Please continue the discussion there

PR #29383, opened 2017-10-29
[Central group table] Remove duplicate of isInGroup from group manager

mrow4a:
This function is not in the interface of IGroupManager and is a duplicate of https://github.com/owncloud/core/blob/master/lib/public/IGroupManager.php#L153. After optimizing that function in GroupManager with central group table, this function should be removed. 

- [x] Requires unit tests adjustments

### DIFF
--- a/lib/private/Group/Manager.php
+++ b/lib/private/Group/Manager.php
@@ -480,13 +480,4 @@ class Manager extends PublicEmitter implements IGroupManager {
 
 		return $this->subAdmin;
 	}
-
-	public function inGroup($uid, $gid) {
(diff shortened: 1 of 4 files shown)


lib/private/Group/Manager.php
```diff
@@ -466,13 +466,4 @@ public function getSubAdmin() {
 
 		return $this->subAdmin;
 	}
-
-	public function inGroup($uid, $gid) {
```
PVince81: is that not on the interface?

PVince81:
Nice find!

PR #29384, opened 2017-10-29
[Central group table] Make getAllSharedWith aware of membership table to optimize fs setup

mrow4a:
This should avoid unecessairy queries by making queries smarter after having membership table.

### DIFF
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -225,7 +225,7 @@ class OC_Util {
 		\OC\Files\Filesystem::logWarningWhenAddingStorageWrapper(true);
 
 		// Make users storage readonly if he is a guest or in a read_only group
-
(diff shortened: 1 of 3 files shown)


lib/private/legacy/util.php
```diff
@@ -240,19 +240,18 @@ public static function setupFS($user = '') {
 				'[]'
 			), true);
 
-			if (!is_array($readOnlyGroups)) {
-				$readOnlyGroups = [];
+			// Find read only groups that are also user groups
+			if (is_array($readOnlyGroups) && !empty($readOnlyGroups)) {
```
mrow4a: @IljaN if you can, try to avoid querying information you possibly don't need.

DeepDiver1975:
@mrow4a please reopen if still of interest and finish it - THX

Issue #29328, opened 2017-10-23
Big file download from SMB via fast networks eats up memory


mmattel:
### Steps to reproduce
1. Have a big file (mine is 6.8GB) on an SMB share mounted to oC
2. Having a (important!) fast connection between server and client (LAN) 
3. Download this file via browser

### Expected behaviour
Download should pass without problems

### Actual behaviour
Download stops after a while (3-4GB downloaded), system becomes unresponsive, after some minutes, system is accessible again.

This does not occur when
- downloading small/smaller files via SMB
- downloading from local, means same file is not on SMB but in the users oC home. In my environment the datadirectory is on NFS
- downloading via a slow connection =eg internet, only a view mbit/s

### Testing
Monitoring the system via htop under different scenarious
On fast/SMB, memory and swap gets eaten up, not from the beginning but after a while until system becomes unresponsive, download gets broken and recovers after a while beeing responsive againg and memory is recovered. Memory recovery also takes place if the file downloaded finishes earlier than memory becomes eaten up. This is monitorable.
``libsmbclient`` installed in the latest public version (``sudo pecl upgrade smbclient``)
Network performance = browser (client) side, the SMB backend delivering the data is fast (55-60MB/s)

Slow Network, SMB: 3.2mbit/s (manually stopped after 510 MB = ~26min, no issues)
![image](https://user-images.githubusercontent.com/3321281/31886896-846c1eb8-b7f6-11e7-941d-285ce187d9ed.png)

Fast Network, SMB 114mbit/s (memory starts raising from 345 to +603MB)
Memory raising appears between 1.2 and 3.7GB downloaded, mostly on higher numbers
![image](https://user-images.githubusercontent.com/3321281/31888032-c6af81bc-b7fa-11e7-9fd2-cc31765852be.png)

Fast Network, SMB 114mbit/s (memory continues raising from 603 to +980MB)
Memory consumption continues
![image](https://user-images.githubusercontent.com/3321281/31886942-b9769f7a-b7f6-11e7-80b3-2646b1e094d9.png)

Pause or stop download at 980MB, waiting a minute
Memory gets freed up
![image](https://user-images.githubusercontent.com/3321281/31886949-c3d2d2d6-b7f6-11e7-98bd-7e881ca3aaa3.png)

Fast Network, Local (which is on an NFS mountpoint 200mbit/s (memory stays at ~380)
[...]

PR #29400, opened 2017-10-31
Disable nginx buffering for file downloads

jvillafanez:
## Description
nginx buffering seems to be causing memory problems while a big file download is happening (big file upload hasn't been checked). The memory usage of the nginx process (not the php-fpm) keeps growing which might crash the server at some point depending on the file size and the memory available in the server.

Note that this behaviour isn't present in apache or in nginx without buffering the response

## Related Issue
https://github.com/owncloud/core/issues/29328

## Motivation and Context
This solution should disable the nginx buffering only for file downloads. The rest of the operation should behave normally depending on the specific nginx configuration.
While modifying the nginx configuration should be possible, it might be complex to disable just the buffering for file downloads, so you might disable the it for the whole server or for a specific location context.

## How Has This Been Tested?
Tested manually with a 4.4GB file (check https://github.com/owncloud/core/issues/29328#issuecomment-340718926 for set up)

## Types of changes

## Checklist:

@DeepDiver1975 @PVince81 To be decided if this will be the solution. Not happy to include server-specific code but it's probably the easiest solution.

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesPlugin.php
@@ -251,6 +251,9 @@ class FilesPlugin extends ServerPlugin {
 			if ($checksum !== null && $checksum !== '') {
 				$response->addHeader('OC-Checksum', $checksum);
 			}
+			// disable nginx buffering so big downloads through ownCloud won't
+			// cause memory problems in the nginx process.
+			$response->addHeader('X-Accel-Buffering', 'no');
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mmattel:
@PVince81 @jvillafanez 
👍 
Approved, works, see https://github.com/owncloud/core/issues/29328#issuecomment-340784009
Please restart the check.
I propose backporting.
Thanks a lot !

jvillafanez:
Backport to stable10 in https://github.com/owncloud/core/pull/29403

DeepDiver1975 (changes requested):
Squash please - otherwise 👍

PVince81 (approved):
:+1:

PR #29401, opened 2017-10-31
UI tests use is_null and consistent ElementNotFoundException messages

phil-davis:
## Description
1) Use ``is_null()`` everywhere rather than mixing it up with ``=== null``
2) Make all ``ElementNotFoundException`` messages have a "standard" format that puts the method, the xpath used and some message text, so that we get easy to read exceptions logged when something goes wrong.
3) Find all the remaining places that do things like:
```
$this->find("xpath", "some xpath locator")->click();
```
and first get the result of the ``find()`` - if it is null then throw a "nice" exception rather than crashing out doing a "click on null object".

## Related Issue
This is more stuff noted while trying to reduce problems running tests on Edge issue #29138 
but enhances/improves the existing UI tests.

## Motivation and Context
Have nicer more consistent exception messages when an element is not found in the UI.
Be more careful to check elements are found before clicking/actioning them.

## How Has This Been Tested?
Travis

## Types of changes

## Checklist:

### DIFF
--- a/tests/TestHelpers/SetupHelper.php
+++ b/tests/TestHelpers/SetupHelper.php
@@ -52,10 +52,10 @@ class SetupHelper {
 		$email = null
 	) {
 		$occCommand = ['user:add', '--password-from-env'];
-		if ($displayName !== null) {
+		if (!is_null($displayName)) {
 			$occCommand = array_merge($occCommand, ["--display-name", $displayName]);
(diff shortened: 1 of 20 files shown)


individual-it (approved):

phil-davis:
Backport stable10 #29402

PR #29405, opened 2017-10-31
Use chunks of 999 when preload directory content

VicDeo:
## Description
Use chunks of 999 when preload directory content

## Related Issue
#29233 

## Motivation and Context
999 placeholder limit in Sqlite/Oracle

## How Has This Been Tested?
With a unit test

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
+++ b/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
@@ -24,6 +24,8 @@
 namespace OCA\DAV\DAV;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use OCA\DAV\Connector\Sabre\Directory;
 use OCA\DAV\Connector\Sabre\Node;
 use Sabre\DAV\INode;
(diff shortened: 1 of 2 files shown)


apps/dav/lib/DAV/FileCustomPropertiesBackend.php
```diff
@@ -214,4 +217,25 @@ protected function getNodeForPath($path){
...
+	 * This limit equals to 999 by default
+	 * and is decreased by $otherPlaceholdersCount
+	 *
+	 * @param int[] $toSlice
+	 * @param int $otherPlaceholdersCount
+	 * @return array
+	 */
+	private function getChunks($toSlice, $otherPlaceholdersCount = 0) {
+		$databasePlatform = $this->connection->getDatabasePlatform();
+		if ($databasePlatform instanceof OraclePlatform || $databasePlatform instanceof SqlitePlatform) {
```
PVince81: In the past I think I'd say let's do the workaround for all DBs to avoid hard-coding the names. But after all this time I'm wondering: why impair the performance of the other DBs just because these two ?

And the code as you wrote it is better than just having a PHPDoc saying "workaround for XYZ", so let's keep it.

PVince81 (approved):
:+1: code looks good

VicDeo:
Stable10: https://github.com/owncloud/core/pull/29416

PR #29409, opened 2017-11-02
Migrating mysql php 7.0 job from jenkins to drone

SergioBertolinSG:


.drone.yml
```diff
@@ -183,9 +183,9 @@ matrix:
     #- PHP_VERSION: 7.0
     #  DB_TYPE: mariadb
     #  TEST_SUITE: phpunit
-    #- PHP_VERSION: 7.0
-    #  DB_TYPE: mysql
-    #  TEST_SUITE: phpunit
+    - PHP_VERSION: 7.0
+      DB_TYPE: mysql
```
DeepDiver1975: which mysql version is used on drone? we have to be care full to choose an older version which does not automatically enable mb4

DeepDiver1975 (changes requested):

.drone.yml
SergioBertolinSG: 5.7

DeepDiver1975: Jenkins is currently running 5.5 - 5.7 is the version to be used with mb4 ...
https://github.com/owncloud/core/blob/master/build/autotest.sh#L228

DeepDiver1975: Please also incorporate these settings in the server - this will give us faster execution times - https://github.com/owncloud/core/blob/master/tests/docker/mariadb/oc.cnf

but lets first see how long it takes without

SergioBertolinSG: Ok, should I migrate the mb4 stage in this PR as well or in a different one?

DeepDiver1975: mb4 into a pr of its own please

DeepDiver1975: THX

DeepDiver1975 (approved):

PR #29415, opened 2017-11-02
decrypt seen users only

butonic:
It does not make sense to iterate aver all backends individually, now that we have the account table. Saves a lot of time when having an ldap with 100k users. Also romoves the dependency on a live ldap server.

### DIFF
--- a/lib/private/Encryption/DecryptAll.php
+++ b/lib/private/Encryption/DecryptAll.php
@@ -30,6 +30,7 @@ use OC\Encryption\Exceptions\DecryptionFailedException;
 use OC\Files\View;
 use \OCP\Encryption\IEncryptionModule;
 use OCP\ILogger;
+use OCP\IUser;
 use OCP\IUserManager;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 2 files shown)


sharidas (approved):
I have verified this change and it worked like charm. This change would definitely improve the performance. And yes it doesn't make sense to loop through users who didn't logged in once. A double :+1:  from my side. If the tests are green, then we should take this change.

PVince81:
@butonic why "To develop", anything missing ?

butonic:
@PVince81 This PR was the reason for me asking about how to test callback stuff. The last amend allowed passing the tests ... will check the test coverage tomorrow.

PVince81:
@butonic any update on the tests?

PVince81:
@butonic do you intend to finish this or should someone else take over ?

PVince81:
@sharidas mind taking over and helping to finish this ?

lib/private/Encryption/DecryptAll.php
```diff
@@ -164,6 +164,9 @@ protected function decryptAllUsersFiles($user = '') {
 		if ($user === '') {
 			$userNo = 1;
 			$numberOfUsers = $this->userManager->countSeenUsers();
+			if ($numberOfUsers === 0) {
```
sharidas: In a scenario where none of the users are seen ( i.e, none of the users are logged in ), then `numberOfUsers` become zero. And guess we don't have to call the `callForSeenUsers`. I have tested in a scratch installation with 2 users who were not logged in and the decrypt-all command is called.

PVince81 (approved):
:+1: please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/30640

PR #28213, opened 2017-06-26
use efficient startsWith implementation, don't check if key exists twice

butonic:
While working on https://github.com/owncloud/core/pull/28212 I was profiling the performance for user:sync and this actually made a measurable difference.

PR #29427, opened 2017-11-03
[stable10] use efficient startsWith implementation, don't check if ke…

DeepDiver1975:
backport of #28213


PVince81:
I suppose this could be optimized further with custom code as we need to traverse the string multiple times, once for strrpos and then again for substr_count.

PVince81 (approved):
:+1:

PR #28214, opened 2017-06-26
load app type only once

butonic:
While working on #28212 I was profiling the performance for user:sync and this actually made a huge measurable difference because it no longer spends 30% of the request merging arrays.

PR #29100, opened 2017-09-26
MariaDB 10.2.7+

VicDeo:
## Description
Backup solution for #28695 
I'm not happy with this solution but if upstream doesn't manage to fix the issue until 10.0.4 we'll need it

Here we bypass native Doctrine/DBAL   `MySqlSchemaManager::_getPortableTableColumnDefinition` implementation for  MariaDB >= 10.2.7 and use customized implementation that unquotes default values.
All DB engines from non-MySQL family work as before. For MySQL family a new listener is registered.
Triggering native/customized codeflow is done inside the listener itself.

## Related Issue
 #28695  

## Motivation and Context
Fixes MariaDB 10.2.7+ support

## How Has This Been Tested?
By installing ownCloud with  MariaDB 10.2.7+ as a DB engine

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29428, opened 2017-11-03
[stable10] load app type only once

DeepDiver1975:
backport of #28214


DeepDiver1975:
I suggest to kill this pr and reopen after we properly addressed the issue in the master PR #29100

PVince81 (approved):
:+1: code makes sense, didn't spot any side effect

PR #29393, opened 2017-10-30
Fix folder upload race condition

PVince81:
## Description
When uploading the same folder twice, sometimes a race condition is
triggered where some of the first uploaded items finish faster and
trigger the "end of transfer" event even though there are further items
to upload.

This fix prevents the end of transfer event to clear the list of uploads
so it can continue.

## Related Issue
Fixes https://github.com/owncloud/core/issues/28844

## How Has This Been Tested?
Manual testing, see ticket.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29435, opened 2017-11-03
[stable10] Fix folder upload race condition

PVince81:
Backport of https://github.com/owncloud/core/pull/29393 to stable10


apps/files/js/file-upload.js
```diff
@@ -37,12 +37,11 @@ OC.FileUpload = function(uploader, data) {
...
 	}
-	var path = '';
+	var basePath = '';
 	if (this.uploader.fileList) {
-		path = OC.joinPaths(this.uploader.fileList.getCurrentDirectory(), this.getFile().name);
-	} else {
-		path = this.getFile().name;
+		basePath = this.uploader.fileList.getCurrentDirectory();
 	}
+	var path = OC.joinPaths(basePath, this.getFile().relativePath || '', this.getFile().name);
```
jvillafanez: I'd rather move the `this.getFile().relativePath || ''` up to avoid the operation being evaluated in the function because it isn't common, but taking into account this is the only change I see for a backport, I'm fine to leave this as tech debt.

jvillafanez (approved):

apps/files/js/file-upload.js
PVince81: Thanks. Considering that this isn't happening in a big loop I think this bit is acceptable.

PR #29436, opened 2017-11-03
[Stable10] Increasing timeout for encryption integration tests in jenkins

SergioBertolinSG:


PVince81 (approved):
:+1:

SergioBertolinSG:
Error in jenkins:
```
Test execution failed: java.lang.NullPointerException: Cannot invoke method startsWith() on null object
```

I'll try something, but maybe we have to use the testResultLocation variable (or add another one).

SergioBertolinSG:
OK env.STAGE_NAME doesn't seem to work, @PVince81 do you know any way to get stage name?

If not, it is ok to you to use the report filename as condition to increase timeout?

PVince81:
@SergioBertolinSG I don't know much about Jenkinsfile. Please research this yourself, unless @DeepDiver1975 has a clue.

PR #28216, opened 2017-06-26
only trigger changeUser if account object really changed

butonic:
While working on #28212 I was profiling the performance for user:sync and this actually made a measurable difference because it was causing the filesystem to be setup repeatedly.

PR #29439, opened 2017-11-03
only trigger changeUser if account object really changed

butonic:
backport of https://github.com/owncloud/core/pull/28216 to 10


DeepDiver1975:
too late - https://github.com/owncloud/core/pull/29429

Issue #29342, opened 2017-10-24
[Postgre] Error while trying to create admin user


RobinFrcd:
### Steps to reproduce
1. Docker-compose on debian (jessie)
```
version: '2'
services:
  owncloud:
      image: owncloud
      ports:
          - 8080:80
      volumes_from:
         - owncloud-data
      links:
         - db:owncloud-db

  db:
      image: postgres
      environment:
          - POSTGRES_PASSWORD=mysecretpassword
      volumes_from:
          - postgres-data

  postgres-data:
      image: postgres
      command: /bin/true
      volumes:
          - /etc/postgresql
          - /var/lib/postgresql

  owncloud-data:
      image: owncloud
      command: /bin/bash -c "/bin/chown -R 33 /var/www/html/data && /bin/chown -R 33 /var/www/html/config"
      volumes:
          - /srv/owncloud/data
          - /srv/owncloud/config

```
2. `docker-compose -f docker-compose.yml up`
3. Try to initialize owncloud 

### Expected behaviour
It should create an admin account and initialize owncloud.

### Actual behaviour
I get this error:
```
Error while trying to create admin user: An exception occurred while executing 'SELECT min_value, increment_by FROM "oc_storages_numeric_id_seq"':

SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist
LINE 1: SELECT min_value, increment_by FROM "oc_storages_numeric_id_...
               ^				
```

I'm using the latest images available on docker.

If I do a `\t`, I get that, but there is no `oc_storages_numeric_id_seq` here ...

```
                    List of relations
 Schema |            Name             | Type  |  Owner
--------+-----------------------------+-------+----------
 public | oc_appconfig                | table | oc_robin
 public | oc_authtoken                | table | oc_robin
 public | oc_comments                 | table | oc_robin
 public | oc_comments_read_markers    | table | oc_robin
 public | oc_credentials              | table | oc_robin
 public | oc_file_locks               | table | oc_robin
 public | oc_filecache                | table | oc_robin
 public | oc_group_admin              | table | oc_robin
 public | oc_group_user               | table | oc_robin
 public | oc_groups                   | table | oc_robin
 public | oc_jobs                     | table | oc_robin
 public | oc_migrations               | table | oc_robin
 public | oc_mimetypes                | table | oc_robin
 public | oc_mounts                   | table | oc_robin
 public | oc_preferences              | table | oc_robin
 public | oc_privatedata              | table | oc_robin
[...]

PR #29456, opened 2017-11-06
Moved postgresql job from jenkins to drone

SergioBertolinSG:
@DeepDiver1975 do we need also to test with the previous version of postgres? (9.6)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -138,7 +138,7 @@ services:
         DB_TYPE: mysql
 
   postgres:
-    image: postgres:9.6
+    image: postgres:9.4
     environment:
       - POSTGRES_USER=owncloud
       - POSTGRES_PASSWORD=owncloud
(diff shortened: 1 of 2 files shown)


SergioBertolinSG:
Drone error

```
SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist
LINE 1: SELECT min_value, increment_by FROM "oc_storages_numeric_id_...
```

Which is the expected from https://github.com/owncloud/core/issues/29342

.drone.yml
```diff
@@ -138,7 +138,7 @@ services:
         DB_TYPE: mysql
 
   postgres:
-    image: postgres:9.6
+    image: postgres:10.0
```
DeepDiver1975: Jenkins is testing against 9.4 - please use that version.
especially since we know that postgres 10 is having trouble

DeepDiver1975 (changes requested):

SergioBertolinSG:
@DeepDiver1975 ready to go?

DeepDiver1975 (approved):

phil-davis:
Backport stable10 #29517

PR #29458, opened 2017-11-06
Adding mode of operations - either single-instance or clustered-instance

### LINKED ISSUE #29444: Cluster mode config.php switch
Add a config switch in config.php to tell OC that we're in a clustered environment.

At first, this switch must only make sure to force-disable the market app. The reason is that if someone uses the market app to install an app on a single cluster node, the app code will be missing from the other nodes and cause breakage. Force-disabling the market app also means that occ upgrade will not update any app code when run on a single node as it would cause similar issues.

Additionally, the switch could also make the configuration file read-only (like "config_is_read_only").

As discussed with @felixboehm and @patrickjahns.

cc @DeepDiver1975 @butonic @mmattel if you guys have additional feedback

DeepDiver1975:
## Description
Adds config switch let owncloud know if the instance is part of a cluster or not.
Instead of adding yet another boolean flag we will use two different values for the mode of operations.
This allows us to extend the operation modes in the future if necessary.

## Related Issue
fixes #29444

## How Has This Been Tested?
- [ ] market app UI
- [ ] market occ commands

## Types of changes

## Checklist:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -581,6 +581,15 @@ $CONFIG = array(
 'config_is_read_only' => false,
 
 /**
+ * This defines the mode of operations. The default value is 'single-instance'
+ * which means that ownCloud is running on a single node, which might be the
+ * most common operations mode. The only other possible value for now is
+ * 'clustered-instance' which means that ownCloud is running on at least 2
+ * nodes. The mode of operations has various impact on the behavior of ownCloud.
+ */
+'operation.mode' => 'single-instance',
+
+/**
  * Logging
  */
 
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -37,6 +37,7 @@ use OCP\App\ManagerEvent;
 use OCP\Files;
 use OCP\IAppConfig;
 use OCP\ICacheFactory;
+use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IUser;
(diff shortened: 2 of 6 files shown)


DeepDiver1975:
market app already handles this properly - see https://github.com/owncloud/market/blob/c752a76f450cd3ab71718defaea71631a6d6e675/lib/MarketService.php#L89

We might want to change the error message ....

DeepDiver1975:
market app related changes discussed in here - https://github.com/owncloud/market/pull/184#issuecomment-342104589

PVince81:
@DeepDiver1975 are you suggesting that we leave the market app enabled even in clustered mode ? If we do, best UX would be to disable the "Install" button in the web UI with a message there.

I think for now easier is likely to have the app auto-disable itself in app.php if it detects clustered mode ?

DeepDiver1975:
> are you suggesting that we leave the market app enabled even in clustered mode ? If we do, best UX would be to disable the "Install" button in the web UI with a message there.

absolutly

PVince81:
@DeepDiver1975 ok, sounds good then. Let's make sure that at least the error message is based on the switch, telling the admin that apps need to be installed manually.

PVince81 (approved):
:+1:

PVince81:
codecov not happy on the two lines that read config

phil-davis:
Backport stable10 #29492

PR #29168, opened 2017-10-08
Log out when resetting password

imujjwal96:
## Description

Log a user out (if already logged in), after the password is reset
## Related Issue

Partly fixes #27612 
## How Has This Been Tested?

Tested on my server @ http://163.172.11.243:8080/owncloud/
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29464, opened 2017-11-06
[stable10] Log out when resetting password

PVince81:
Backport of https://github.com/owncloud/core/pull/29168 to stable10


core/Controller/LostController.php
```diff
@@ -240,6 +246,8 @@ public function setPassword($token, $userId, $password, $proceed) {
 			return $this->error($e->getMessage());
 		}
 
+		$this->logout();
+
```
Peter-Prochaska: 👍

peterprochaska (approved):

PR #29476, opened 2017-11-06
[stable9.1] Upgrade node package versions

PVince81:
Upgrade node package versions to be compatible with node v8.9.0 and avoid CI failure due to npm setup...

@DeepDiver1975 @tomneedham


PVince81:
as you can see the JS tests passed

DeepDiver1975 (approved):

PR #29492, opened 2017-11-07
[stable10] Adding mode of operations - either single-instance or clus…

### LINKED ISSUE #29444: Cluster mode config.php switch

DeepDiver1975:
…tered-instance - fixes #29444

backport of #29458 to stable10

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -583,6 +583,15 @@ $CONFIG = array(
 'config_is_read_only' => false,
 
 /**
+ * This defines the mode of operations. The default value is 'single-instance'
+ * which means that ownCloud is running on a single node, which might be the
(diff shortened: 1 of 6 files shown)


PVince81 (approved):
:+1:

PR #29497, opened 2017-11-07
UI tests split sharing

phil-davis:
## Description
1) Split the UI tests sharing suite into 2 suites.
2) The ``if: type = cron`` control jobs in the job matrix that are run for other browsers only in the daily Travis background cron job. When wanting to do a real-time check run to see that some tests work in these browsers we remove that line. Having it 1st in each section (with the "-" in front) is a nuisance. Actually it works fine in any line of the section. Move it to the 2nd line, where it is easily deleted as needed, (It annoyed me enough today when running various combinations of Edge tests on Travis)

## Motivation and Context
The elapsed time is now around 40 minutes. It would be better to be more like the other suites that take 15-25 minutes each.

## How Has This Been Tested?
Travis will confirm.

## Types of changes

## Checklist:


phil-davis:
The remaining sharing tests still took 28 minutes. We need to split the suite anyway. But should also look into why these tests are so slow.

individual-it (approved):

Issue #29247, opened 2017-10-13
TBD: Hard-requirement for php-intl


kdslkdsaldsal:
Concerning the following log-entries:

``
{"reqId":"Htm9GhB5ptOjOmLGHbTf","level":3,"time":"2017-10-11T08:08:16+00:00","remoteAddr":"1.2.3.4","user":"--","app":"PHP","method":"GET","url":"/cron.php","message":"You are using a fallback implementation of the intl extension. Installing the native one is highly recommended instead. at /var/www/owncloud/lib/composer/patchwork/utf8/src/Patchwork/Utf8/Bootup/intl.php#18"}
``

https://github.com/owncloud/core/issues/28079 -> Issue with app installation (``Archives of type application/octet-stream are not supported`` during app upgrade/installation) fixed by installing/enabling php-intl according to @bartmans in https://github.com/owncloud/core/issues/28079#issuecomment-333485833 and https://github.com/owncloud/core/issues/28079#issuecomment-333597384

https://github.com/owncloud/core/issues/29194 -> Issue with e.g. Korean chars on external storages. Suggestion by @PVince81 in https://github.com/owncloud/core/issues/29194#issuecomment-336439679 to install the php-intl module

IIRC there where also other issues fixed after installing php-intl but its currently only an "optional" extension listed at https://doc.owncloud.org/server/latest/admin_manual/installation/source_installation.html#optional

If installing this extension is fixing such issues shouldn't it be a hard requirement?

DeepDiver1975:
I'm all in to make it a hard requirement. 👍

kdslkdsaldsal:
Some refs for older discussions about that: https://github.com/owncloud/core/issues/20065 / https://github.com/owncloud/3rdparty/pull/249 / https://github.com/owncloud/core/issues/22324 / https://github.com/owncloud/core/pull/22446

When reading those correctly there might even more to use / suggest / require or am i wrong:

https://github.com/owncloud/3rdparty/pull/249/files#diff-d6d06ec90186a854079e1165c473b6f3L1245

PVince81:
I'd put this on the same level like php-gd and others. If the module is not enabled, OC refuses to work.

Once this is done, make sure this is mentioned on the release notes.

DeepDiver1975:
In addition we shall see if we still need the patchwork library - it is used to get add intl implementations on php level.

PVince81:
@jvillafanez can you take care of this ?

jvillafanez:
[...]

PR #29500, opened 2017-11-07
Add php-intl as hard requirement

jvillafanez:
## Related Issue
https://github.com/owncloud/core/issues/29247

## Motivation and Context
See issue

The `Collator` class is used in https://github.com/owncloud/core/blob/intl_hard_requirement/lib/private/NaturalSort.php#L89

## How Has This Been Tested?
Manually tested from a working installation. Removed the php7.0-intl package and also the config.php to trigger a new installation process. A message requiring the installation of the "intl" php module is shown.
Note that ownCloud will keep running without the package being installed if the ownCloud installation isn't fresh (after an upgrade, for example)

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -782,6 +782,7 @@ class OC_Util {
 				'DOMDocument' => 'dom',
 				'XMLWriter' => 'XMLWriter',
 				'XMLReader' => 'XMLReader',
+				'Collator' => 'intl',
 			],
 			'functions' => [
 				'xml_parser_create' => 'libxml',


PVince81:
do we have harder than that like for php7-gd ? or is that the same level of hard ?

DeepDiver1975:
Please evaluate if we can Drop the patchwork lib now. Thx

jvillafanez:
> do we have harder than that like for php7-gd ? or is that the same level of hard ?

Unless there are specific check spread around the code, I think that's the place. Checks for gd and other functions / classes are around those lines.

DeepDiver1975 (approved):

jvillafanez:
Backport in https://github.com/owncloud/core/pull/29539

PR #29511, opened 2017-11-08
Trigger eventdispatcher for federated share

sharidas:
Trigger event dispatcher for federated share
receiving.

## Description

Change to trigger event dispatcher event for federated share receive.

## Motivation and Context

This will help users to log events for fed share receive.

## Types of changes

## Checklist:

### DIFF
--- a/apps/federatedfilesharing/lib/RequestHandler.php
+++ b/apps/federatedfilesharing/lib/RequestHandler.php
@@ -34,6 +34,8 @@ use OCP\IDBConnection;
 use OCP\IRequest;
 use OCP\IUserManager;
 use OCP\Share;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class RequestHandler
@@ -65,6 +67,9 @@ class RequestHandler {
 	/** @var  IUserManager */
 	private $userManager;
 
+	/** @var EventDispatcherInterface  */
+	private $eventDispatcher;
+
 	/** @var string */
 	private $shareTable = 'share';
 
@@ -78,6 +83,7 @@ class RequestHandler {
 	 * @param Notifications $notifications
 	 * @param AddressHandler $addressHandler
 	 * @param IUserManager $userManager
+	 * @param EventDispatcherInterface $eventDispatcher
 	 */
 	public function __construct(FederatedShareProvider $federatedShareProvider,
 								IDBConnection $connection,
@@ -85,7 +91,8 @@ class RequestHandler {
 								IRequest $request,
 								Notifications $notifications,
 								AddressHandler $addressHandler,
-								IUserManager $userManager
+								IUserManager $userManager,
+								EventDispatcherInterface $eventDispatcher
 	) {
 		$this->federatedShareProvider = $federatedShareProvider;
 		$this->connection = $connection;
@@ -94,6 +101,7 @@ class RequestHandler {
 		$this->notifications = $notifications;
 		$this->addressHandler = $addressHandler;
 		$this->userManager = $userManager;
+		$this->eventDispatcher = $eventDispatcher;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


apps/federatedfilesharing/lib/Notifier.php
```diff
@@ -25,16 +25,19 @@
...
 	/** @var \OCP\L10N\IFactory */
 	protected $factory;
+	private $dispatcher;
 
 	/**
 	 * @param \OCP\L10N\IFactory $factory
 	 */
 	public function __construct(\OCP\L10N\IFactory $factory) {
 		$this->factory = $factory;
+		$this->dispatcher = \OC::$server->getEventDispatcher();
```
PVince81: why no injection?

apps/federatedfilesharing/lib/RequestHandler.php
```diff
@@ -165,6 +166,8 @@ public function createShare($params) {
 					$sharedByFederatedId = $ownerFederatedId;
 				}
 
+				$event = new GenericEvent(null, ['name' => $name, 'targetuser' => $sharedByFederatedId]);
+				\OC::$server->getEventDispatcher()->dispatch('OCA\FederatedFileSharing\RequestHandler::receivedShare_createShare', $event);
```
PVince81: I don't think the event name should match the class + method names.

As an event consumer, I'd rather expect the event name to be understandable and meaningful: what is this event about on a high-level ? The namespace also doesn't need to be one of a class. I think if we keep it to `OCA\FederatedFileSharing::eventName` it would be fine.

apps/federatedfilesharing/lib/Notifier.php
sharidas: Updated with injection.

apps/federatedfilesharing/lib/RequestHandler.php
sharidas: Updated the event names:
- `\OCA\FederatedFileSharing::local_shareReceived`
- `\OCA\FederatedFileSharing::remote_shareReceived`

sharidas:
@PVince81  There is one issue which is bugging me. The Notifier is the place where I found the local fed shares are informed to the recipient. While doing manual testing, I found that lets say user delays to accept the fed share, then the logger keeps sending the message about arrival of share until the recipient accepts. I am pretty sure the admins wont be happy with this behaviour.

sharidas:
There was an app, notification which was polling. When I removed the app, and then refreshed the browser, I get the logs for received local fed share. No more repeated messages.

PVince81:
@sharidas we need to find a way to log the message only once even when notifications app is enabled. I still don't get why the notification app would trigger such event. Maybe the event trigger needs to be moved to a different more accurate location ?

sharidas:
Verified this change as below:
- [x] fed shares within the local instance, users, works
- [x] fed shares with remote instance, users works
- [x] Verified that only one log data is there for the receive for a share. No more duplication of the log.

apps/federatedfilesharing/lib/RequestHandler.php
```diff
@@ -165,6 +166,8 @@ public function createShare($params) {
 					$sharedByFederatedId = $ownerFederatedId;
 				}
 
+				$event = new GenericEvent(null, ['name' => $name, 'targetuser' => $sharedByFederatedId]);
```
PVince81: I guess we don't have more detailed info for the hook ?

PVince81 (approved):
:+1: looks good

I've restarted the Drone build

apps/federatedfilesharing/lib/RequestHandler.php
sharidas: Updated with more details. I guess this should be enough.

apps/federatedfilesharing/lib/RequestHandler.php
```diff
@@ -165,6 +166,10 @@ public function createShare($params) {
 					$sharedByFederatedId = $ownerFederatedId;
 				}
 
+				$event = new GenericEvent(null, ['name' => $name, 'targetuser' => $sharedByFederatedId,
+								'owner' => $owner, 'sharewith' => $shareWith,
+								'sharedby' => $sharedBy, 'remoteid' => $remoteId]);
+				\OC::$server->getEventDispatcher()->dispatch('\OCA\FederatedFileSharing::remote_shareReceived', $event);
```
PVince81: inject the event dispatcher ?

PVince81:
@sharidas please backport asap

sharidas:
The backport PR is already merged to stable 10.
Backport PR: https://github.com/owncloud/core/pull/29566

PR #29515, opened 2017-11-09
Moved job phpunit sqlite php 7.0 from jenkins to drone

SergioBertolinSG:


DeepDiver1975:
I wonder if we need that config - we already run tests with 7.0 ...

SergioBertolinSG:
Tests are running using php7.1 with sqlite and php7.0 with mysql so this will only show particular problems using sqlite with php7.0 not happening with mysql.
I agree with just removing it.

DeepDiver1975:
Wait - for anything else the sqlite we execute group DB only - in this case we need to run ALL php versions with sqlite

SergioBertolinSG:
Ok then.

DeepDiver1975 (approved):

phil-davis:
Backport stable10 #29529

phil-davis:
@SergioBertolinSG I made the backport to get CI moving along in the quiet hours.
Please remove backport-request label if this only backports to stable10.

PR #29559, opened 2017-11-14
catch the right exception

butonic:
OCP\Lock\LockedException is relative to the current namespace. so this can never be caught. must be `\OCP\Lock\LockedException` or better `LockedException` because it is in the use stamtements

related to the repair steps so I figure this is kind of important


PVince81 (approved):
:+1: thanks

PVince81:
stable10: https://github.com/owncloud/core/pull/29567

PR #29567, opened 2017-11-14
[stable10] catch a known exception

PVince81:
Backport of https://github.com/owncloud/core/pull/29559 to stable10


butonic:
👍 thanks

PR #29587, opened 2017-11-15
add item_source index

butonic:
```sql
EXPLAIN SELECT * FROM `oc_share` WHERE `item_source` = '1410';
+------+-------------+----------+------+---------------+------+---------+------+------+-------------+
| id   | select_type | table    | type | possible_keys | key  | key_len | ref  | rows | Extra       |
+------+-------------+----------+------+---------------+------+---------+------+------+-------------+
|    1 | SIMPLE      | oc_share | ALL  | NULL          | NULL | NULL    | NULL | 2998 | Using where |
+------+-------------+----------+------+---------------+------+---------+------+------+-------------+
1 row in set (0.00 sec)

CREATE INDEX item_source_index ON oc_share (item_source);
Query OK, 0 rows affected (0.09 sec)
Records: 0  Duplicates: 0  Warnings: 0

EXPLAIN SELECT * FROM `oc_share` WHERE `item_source` = '1410';
+------+-------------+----------+------+-------------------+-------------------+---------+-------+------+-----------------------+
| id   | select_type | table    | type | possible_keys     | key               | key_len | ref   | rows | Extra                 |
+------+-------------+----------+------+-------------------+-------------------+---------+-------+------+-----------------------+
|    1 | SIMPLE      | oc_share | ref  | item_source_index | item_source_index | 768     | const |    1 | Using index condition |
+------+-------------+----------+------+-------------------+-------------------+---------+-------+------+-----------------------+
1 row in set (0.00 sec)
```

or more visually:

6d6184662f97bf8dd753c3c1a8d82e83249ba075645632b752 pimgpsh_fullsize_distr

index added at 15:45

### DIFF
--- a/apps/files_sharing/appinfo/Migrations/Version20171115154900.php
+++ b/apps/files_sharing/appinfo/Migrations/Version20171115154900.php
@@ -1,0 +1,39 @@
+<?php
+/**
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Sharing\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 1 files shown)


butonic:
@DeepDiver1975 backport to 10 in https://github.com/owncloud/core/pull/29592

DeepDiver1975:
there is no coverage on the migrations ... this is because they are executed on installation where no coverage is collected.

@PVince81 @butonic general question: shall we exclude migrations from coverage?

butonic:
@patrickjahns any ideas for testing migrations with your docker toobox?

patrickjahns:
What do you want to test? 
Upgrading owncloud, or if migrations are successfully executed?

If the later - I've only seen database migration tests when we based our projects on doctrine migrations.
Wet let the migrations execute, and them compare if `schema:update` would still would detect changes / differences between database and entity model. (Assuming developers populated their migrations properly and they we're in sync with the entities).

Since we don't have this luxury currently - we might need to discuss how/when we want to test migrations. 
Guess best approach would be, to have a pre-migration schema/dump and a post-migration schema/dump and compare if executing the migration would result in the desired effect?

tomneedham:
> What do you want to test?

a lot of the time its just checking that nothing explodes on the various matrices of db platforms - since by default we dont run these in tests automagically

DeepDiver1975 (approved):

Issue #29600, opened 2017-11-16
Only file basename shown while waiting for rename AJAX


phil-davis:
### Steps to reproduce
1. Create a file with a file extension - e.g. file.txt
2. In the webUI, rename the file to qqq.txt
3. Watch carefully in the 500ms after you press enter, while "the system" is doing the real rename on the server and waiting for the result, and the "spinner" is spinning in the thumbnail icon position.

### Expected behaviour
Either:
1) The old file name is visible on the UI, or
2) The new file name is is visible on the UI or
3) No file name is visible on the UI, or
4) some "rename in progress" message is visible in place of the file name or...?

### Actual behaviour
The file name is shown without the file extension - "qqq" only.

### Server configuration
**Operating system**:
Ubuntu 16.04 

**Web server:**
PHP dev server

**Database:**
mySQL

**PHP version:**
7.0

**ownCloud version:** (see ownCloud admin page)
current git master (10.0.4+)

**Where did you install ownCloud from:**
GitHub

phil-davis:
This was noticed because the automated UI tests would sometimes rename a file and then fail to find the renamed file in the local list of files on the webUI - because the rename was still in progress and the file name was thus missing the file extension. That was causing a "generic random" error in tests.

By fixing this, we will always find the file in the list. Then we have to deal with waiting properly until the file rename has finished and the list has been resorted and the file row is ready for the next action. That is a separate UI test timing issue.

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributor most likely able to help you is @PVince81.

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

phil-davis:
Fixed by #29680 (master) and #30040 (stable10)

PR #29601, opened 2017-11-16
Show full new file name while waiting for rename AJAX

phil-davis:
## Description
Show the full file name after the file name input has been accepted and while waiting for the AJAX to do the actual rename on the server.

## Related Issue
#29600 

## Motivation and Context
Showing just the file name without file extension looks odd (for half a second). I made me wonder what was really going on.

## How Has This Been Tested?
Manually watching. Travis will do a range of renames to check it out.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/js/filelist.js
+++ b/apps/files/js/filelist.js
@@ -2044,11 +2044,7 @@
 						// mark as loading (temp element)
 						self.showFileBusyState(tr, true);
 						tr.attr('data-file', newName);
-						var basename = newName;
-						if (newName.indexOf('.') > 0 && tr.data('type') !== 'dir') {
-							basename = newName.substr(0, newName.lastIndexOf('.'));
-						}
-						td.find('a.name span.nametext').text(basename);
+						td.find('a.name span.nametext').text(newName);
 						td.children('a.name').show();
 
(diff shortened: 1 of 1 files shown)


apps/files/js/filelist.js
```diff
@@ -2044,11 +2044,7 @@
 						// mark as loading (temp element)
 						self.showFileBusyState(tr, true);
 						tr.attr('data-file', newName);
-						var basename = newName;
-						if (newName.indexOf('.') > 0 && tr.data('type') !== 'dir') {
-							basename = newName.substr(0, newName.lastIndexOf('.'));
```
phil-davis: If anyone knows why this code is here, please say so???

phil-davis: Looks like it was introduced in:
https://github.com/owncloud/core/pull/3223/files#diff-dafefd9519924bec4efa9f5af4af8e68R231

PVince81: I think this is because we had (do we still have?) different CSS styling for the file name vs its extension

phil-davis: @PVince81 good thought. Indeed the file extension goes in a separate HTML tag that is styled differently.
See #29680 for an improved solution.

PR #29602, opened 2017-11-17
UI tests waitTillFileRowsAreReady after rename

phil-davis:
## Description
After a file rename, wait until the file row(s) are no longer "busy" before proceeding.

## Motivation and Context
UI tests that do a file rename, and then the next step is to take some action on the file such as sharing the file, sometimes fail trying to click the "sharing", "file actions menu" etc button. The error given is that the "element is not clickable". Indeed this is true. When a file is renamed, AJAX goes off to do the rename on the server. While waiting for the server response (success or fail), the row of the file list is set "busy". The sharing and file actions menus are hidden and the thumbnail is replaced with a spinner.

This was observed running UI tests locally. It does not seem to happen (very often) when running on Travis-SauceLabs. Probably that is because of round-trip times between Travis and SauceLabs which create a "natural" wait that avoids seeing the issue.

We need to pro-actively wait until the file list is properly ready for the next action.

## How Has This Been Tested?
Repeated local runs of sharing UI tests.

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/lib/FilesPage.php
+++ b/tests/ui/features/lib/FilesPage.php
@@ -215,6 +215,8 @@ class FilesPage extends FilesPageBasic {
 			echo $message;
 			error_log($message);
 		}
+
+		$this->waitTillFileRowsAreReady($session);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


individual-it (approved):

phil-davis:
Backport stable10 #29605

PR #29607, opened 2017-11-17
Add versioning to objectstore

DeepDiver1975:
## Description
Adds necessary interfaces and code segments to allow implementations of versioning within the objectstore scope

## Related Issue
#29087 

## How Has This Been Tested?
- scality s3 docker
- files_primary_s3
- manual testing web ui

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_trashbin/tests/StorageTest.php
+++ b/apps/files_trashbin/tests/StorageTest.php
@@ -28,9 +28,11 @@
 
 namespace OCA\Files_Trashbin\Tests;
 
+use OC\Files\ObjectStore\ObjectStoreStorage;
 use OC\Files\Storage\Temporary;
 use OC\Files\Filesystem;
 use OC\Files\View;
+use OCP\Files\Storage;
 use Test\TestCase;
 use Test\Traits\UserTrait;
 
@@ -196,6 +198,7 @@ class StorageTest extends TestCase {
 	 * Test that deleted versions properly land in the trashbin.
 	 */
 	public function testDeleteVersionsOfFile() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		// trigger a version (multiple would not work because of the expire logic)
@@ -225,6 +228,7 @@ class StorageTest extends TestCase {
 	 * Test that deleted versions properly land in the trashbin.
 	 */
 	public function testDeleteVersionsOfFolder() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		// trigger a version (multiple would not work because of the expire logic)
@@ -260,6 +264,7 @@ class StorageTest extends TestCase {
 	 * Test that deleted versions properly land in the trashbin when deleting as share recipient.
 	 */
 	public function testDeleteVersionsOfFileAsRecipient() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		$this->userView->mkdir('share');
@@ -312,6 +317,7 @@ class StorageTest extends TestCase {
 	 * Test that deleted versions properly land in the trashbin when deleting as share recipient.
 	 */
 	public function testDeleteVersionsOfFolderAsRecipient() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		$this->userView->mkdir('share');
@@ -380,6 +386,7 @@ class StorageTest extends TestCase {
 	 * unlink() which should NOT trigger the version deletion logic.
 	 */
 	public function testKeepFileAndVersionsWhenMovingFileBetweenStorages() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		$storage2 = new Temporary([]);
@@ -421,6 +428,7 @@ class StorageTest extends TestCase {
 	 * unlink() which should NOT trigger the version deletion logic.
 	 */
 	public function testKeepFileAndVersionsWhenMovingFolderBetweenStorages() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		$storage2 = new Temporary([]);
@@ -461,6 +469,7 @@ class StorageTest extends TestCase {
 	 * out of the shared folder
 	 */
 	public function testOwnerBackupWhenMovingFileOutOfShare() {
+		$this->markTestSkippedIfStorageHasOwnVersioning();
 		\OCA\Files_Versions\Hooks::connectHooks();
 
 		$this->userView->mkdir('share');
@@ -641,6 +650,14 @@ class StorageTest extends TestCase {
 			$this->markTestSkipped('Skipping since the current home storage backend requires the user to logged in');
 		} else {
 			$this->userView->unlink('test.txt');
+		}
+	}
+
+	private function markTestSkippedIfStorageHasOwnVersioning() {
+		/** @var Storage $storage */
+		list($storage, $internalPath) = $this->userView->resolvePath('folder/inside.txt');
+		if ($storage->instanceOfStorage(ObjectStoreStorage::class)) {
+			$this->markTestSkipped();
 		}
 	}
 }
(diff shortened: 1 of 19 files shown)


apps/files_versions/lib/Storage.php
```diff
@@ -174,7 +175,16 @@ public static function store($filename) {
 			}
 
 			list($uid, $filename) = self::getUidAndFilename($filename);
+			/** @var \OCP\Files\Storage\IStorage $storage */
+			list($storage, $internalPath) = Filesystem::resolvePath($filename);
+			if ($storage->instanceOfStorage(IVersionedStorage::class)) {
+				/** @var IVersionedStorage $storage */
+				if ($storage->saveVersion() === true) {
```
PVince81: how does it know what version to save ? I'd expect an argument

lib/public/Files/ObjectStore/IVersionedObjectStorage.php
```diff
@@ -0,0 +1,80 @@
...
+	 * @since 10.1.0
+	 */
+	public function restoreVersion($urn, $versionId);
+
+	/**
+	 * Tells the storage to explicitly create a version of a given file
+	 * @return boolean
+	 * @since 10.1.0
+	 */
+	public function saveVersion();
```
PVince81: where's the given file ? 😉

PVince81 (commented):
The API tells me this cannot work as you are not passing any path to `saveVersion`. If that's expected, please clarify everything in PHPDocs.

If not then I wonder what you tested 😉

DeepDiver1975:
> The API tells me this cannot work as you are not passing any path to saveVersion. If that's expected, please clarify everything in PHPDocs.
> 
> If not then I wonder what you tested 😉

[Label]()

PVince81:
@DeepDiver1975 well I looked at the part "How Has This Been Tested?" and saw you manually test it.

But you're right, I'll check back later when ready to review. Sorry for the noise.

DeepDiver1975:
> @DeepDiver1975 well I looked at the part "How Has This Been Tested?" and saw you manually test it.

Explanation: versioning for non-objectstore is still done via hooks - saveVersion is not implemented in class Common. In objectstore this method is empty since it is done out side of our skope,

apps/files_trashbin/tests/StorageTest.php
```diff
@@ -643,4 +652,12 @@ public function testSingleStorageDeleteFileLoggedOut() {
 			$this->userView->unlink('test.txt');
 		}
 	}
+
+	private function markTestSkippedIfStorageHasOwnVersioning() {
+		/** @var Storage $storage */
+		list($storage, $internalPath) = $this->userView->resolvePath('folder/inside.txt');
+		if ($storage->instanceOfStorage(ObjectStoreStorage::class)) {
```
PVince81: shouldn't it be `IVersionedStorage` here or whatever we called the interface ?

apps/files_versions/js/versioncollection.js
```diff
@@ -38,19 +38,17 @@
...
-				var revision = parseInt(version.id, 10);
+			return _.map(result, function(version) {
+				var revision = version.id;
 				return {
 					id: revision,
 					name: revision,
 					fullPath: fullPath,
-					timestamp: revision,
-					versionId: revision,
+					timestamp: moment(new Date(version['{DAV:}getlastmodified'])).format('X'),
```
PVince81: oh, so was this broken in the other PR ? I missed it when testing the versions panel

apps/files_versions/lib/Storage.php
```diff
@@ -174,7 +175,16 @@ public static function store($filename) {
 			}
 
 			list($uid, $filename) = self::getUidAndFilename($filename);
+			/** @var \OCP\Files\Storage\IStorage $storage */
+			list($storage, $internalPath) = Filesystem::resolvePath($filename);
+			if ($storage->instanceOfStorage(IVersionedStorage::class)) {
+				/** @var IVersionedStorage $storage */
+				if ($storage->saveVersion($internalPath) === true) {
```
PVince81: `=== true` might not be needed

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -81,9 +81,13 @@ public function getDirectoryListing() {
 		}
 		/** @var IVersionedStorage | Storage $storage */
 		$versions = $storage->getVersions($internalPath);
-		return array_map(function($version) use ($storage, $internalPath) {
+		return array_values(array_map(function($version) use ($storage, $internalPath, $view, $path) {
+			if (!isset($version['mime-type'])) {
```
PVince81: I think in most of the code we always called it "mimetype", without dash. Not sure if we want to use this new style here

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -410,4 +412,60 @@ public function writeBack($tmpFile) {
 	public function hasUpdated($path, $time) {
 		return false;
 	}
+
+	public function saveVersion($internalPath) {
+		if ($this->objectStore instanceof IVersionedObjectStorage) {
+			$stat = $this->stat($internalPath);
+			if ($stat === false) {
+				return true;
```
PVince81: we return true if the file did not exist ? would throwing an exception be better ?

if this code is correct, please add a comment to explain why it makes sense to return true when the file does not exist

lib/public/Files/ObjectStore/IVersionedObjectStorage.php
```diff
@@ -0,0 +1,80 @@
...
+namespace OCP\Files\ObjectStore;
+
+/**
+ * Interface IVersionedObjectStorage - storage layer to access version of an
+ * object living in an object store
+ *
+ * @package OCP\Files\ObjectStore
+ * @since 10.0.5
+ */
+interface IVersionedObjectStorage {
```
PVince81: why not just reuse IVersionedStorage or at least extend it ? from what I see the method signatures are 100% the same, just the name of arguments and PHPDoc might be slightly different...

tests/lib/Files/ObjectStore/ObjectStoreStorageTest.php
```diff
@@ -0,0 +1,147 @@
+<?php
+/**
+ * @author Jörn Friedrich Dreyer
```
PVince81: impersonating Dr Dreyer ?

tests/lib/Files/ObjectStore/ObjectStoreStorageTest.php
```diff
@@ -0,0 +1,147 @@
...
+ * @copyright (c) 2014 Jörn Friedrich Dreyer <jfd@owncloud.com>
```
PVince81: and time travelling, Doctor Who ?

tests/lib/Files/ObjectStore/ObjectStoreTest.php
```diff
@@ -0,0 +1,149 @@
+<?php
+/**
+ * @author Jörn Friedrich Dreyer
+ * @copyright (c) 2014 Jörn Friedrich Dreyer <jfd@owncloud.com>
```
PVince81: another time travelling doctor spotted

PVince81 (commented):
See comments

apps/files_trashbin/tests/StorageTest.php
DeepDiver1975: no Common implements IVersionedStorage - this is a temporal hack until versioning for regular storages is moved into Common

apps/files_versions/js/versioncollection.js
DeepDiver1975: no - for the regular versioning the revision is the timestamp and it worked.
but for s3 the version id is no longer the timestamp -> 💥

lib/public/Files/ObjectStore/IVersionedObjectStorage.php
DeepDiver1975: no - the arguments are internalPath vs fileid.

tests/lib/Files/ObjectStore/ObjectStoreStorageTest.php
DeepDiver1975: actually yes - because this is copied over from SwiftTests

lib/public/Files/ObjectStore/IVersionedObjectStorage.php
PVince81: argh... I didn't spot that... Sad that we can't just use the same thing.

If we'd use IVersionedStorage the object store would need to first resolve path to file id, right ? And that would be more expensive and is why we don't want to do it that way ?

DeepDiver1975: this conversion is what ObjectStoreStorage does.
IObjectStore is a much simpler interface which works on fileid only

PVince81 (approved):
:+1:

PR #29615, opened 2017-11-17
Add setup check for overwrite.cli.url setting

### LINKED ISSUE #29606: Setup check for overwrite.cli.url
When installing from occ, usually the URL might not be set correctly. The CLI cannot detect the true domain.

I suggest to have a setup check that does the following:
1) Generate a URL on the backend using the APIs, the APIs will use whatever values are set `overwrite.cli.url` or others when needed.
2) Return said value to the setup check JS code
3) JS code connects to that URL and appends "status.php"
4) If no status.php found, then it's likely the URL is wrong and needs adjusting.

Bonus check: if the given URL contains a domain not listed in trusted_domains, also display a warning.

Inspired by questions from @settermjd.

cc @kdslkdsaldsal @butonic @DeepDiver1975 @cdamken

PVince81:
## Description
Checks that "overwrite.cli.url" is set and is usable by calling status.php.
If no status.php response, displays a warning message in setup checks section in settings page.

## Related Issue
Fixes https://github.com/owncloud/core/issues/29606

## How Has This Been Tested?
- [x] TEST: when not set at all, message appears
- [x] TEST: when set to wrong URL, message appears
- [x] TEST: when set to correct URL, no message

## Types of changes

## Checklist:

## TODO:
- [ ] double check how the CLI URL influences notification URLs. it seems that when it's not set it still works fine, but might not work in all scenarios. (needed to decide whether to keep the warning when it's not set)
- [ ] add unit tests

### DIFF
--- a/core/js/setupchecks.js
+++ b/core/js/setupchecks.js
@@ -296,6 +296,28 @@
 			}
 
 			return messages;
+		},
+
+		checkCliUrl: function(cliUrl) {
+			var deferred = $.Deferred();
+			var afterCall = function(result, status) {
+				var messages = [];
+				if (status === 'error' || _.isUndefined(result.installed)) {
+					messages.push({
+						msg: t('core', 'Please check the "overwrite.cli.url" setting: it must point to a server URL accessible from the outside as it will be used in emails, notifications, etc'),
+						type: OC.SetupChecks.MESSAGE_TYPE_ERROR
+					});
+				}
+				deferred.resolve(messages);
+			};
+
+			$.ajax({
+				type: 'GET',
+				url: cliUrl + '/status.php',
+				allowAuthErrors: true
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
tests need some love

PVince81:
> tests need some love

yes, but before that I need to finish my TODOs... there is some unclarity there about expectation

PVince81:
Hmm, apparently the notifications URL is not affected by "overwrite.cli.url", however it's influenced by "overwritehost". So we'll likely need to add a second check using the same routing code.

PVince81:
Another problem: if the `overwrite*` settings are pointing to another domain, then we can't test it unless the current OC is being opened in the same domain.

PVince81:
doesn't work as expected, closing due to lack of time

PR #29488, opened 2017-11-07
Feature/29087 behat for recovering

peterprochaska:
## Description
New Behat Test for versioning

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29637, opened 2017-11-20
Testing /remote.php/dav/meta/$fileId/v

DeepDiver1975:
## Description
Follow up of https://github.com/owncloud/core/pull/29488
This PR will add more integration tests for the meta dav resource

## How Has This Been Tested?
using behat

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/WebDav.php
+++ b/tests/integration/features/bootstrap/WebDav.php
@@ -517,6 +517,16 @@ trait WebDav {
 	}
 
 	/**
+	 * @Then the version folder of fileId :fileId contains :count elements
+	 * @param int $count
+	 * @param int $fileId
+	 */
+	public function theVersionFolderOfFileIdContainsElements($fileId, $count) {
+		$elements = $this->listVersionFolder($this->currentUser, '/meta/'.$fileId.'/v', 1);
+		PHPUnit_Framework_Assert::assertEquals($count, count($elements)-1);
+	}
+
+	/**
 	 * @Then the content length of file :path with version index :index for user :user in versions folder is :length
 	 * @param $path
 	 * @param $index
@@ -527,7 +537,7 @@ trait WebDav {
 		$fileId = $this->getFileIdForPath($user, $path);
 		$elements = $this->listVersionFolder($user, '/meta/'.$fileId.'/v', 1, ['{DAV:}getcontentlength']);
 		$elements = array_values($elements);
-		PHPUnit_Framework_Assert::assertEquals($length, $elements[1]['{DAV:}getcontentlength']);
+		PHPUnit_Framework_Assert::assertEquals($length, $elements[$index]['{DAV:}getcontentlength']);
 	}
 
 	/* Returns the elements of a report command
@@ -802,6 +812,7 @@ trait WebDav {
 		$file = \GuzzleHttp\Stream\Stream::factory($content);
 		try {
 			$this->response = $this->makeDavRequest($user, "PUT", $destination, [], $file);
+			return $this->response->getHeader('oc-fileid');
 		} catch (\GuzzleHttp\Exception\ServerException $e) {
 			// 4xx and 5xx responses cause an exception
 			$this->response = $e->getResponse();
@@ -1168,4 +1179,18 @@ trait WebDav {
 		$currentFileID = $this->getFileIdForPath($user, $path);
 		PHPUnit_Framework_Assert::assertEquals($currentFileID, $this->storedFileID);
 	}
+
+	/**
+	 * @When user :user restores version index :versionIndex of file :path
+	 * @param string $user
+	 * @param int $versionIndex
+	 * @param string $path
+	 */
+	public function userRestoresVersionIndexOfFile($user, $versionIndex, $path) {
+		$fileId = $this->getFileIdForPath($user, $path);
+		$client = $this->getSabreClient($user);
+		$versions = array_keys($this->listVersionFolder($user, '/meta/'.$fileId.'/v', 1));
+		$client->request('COPY', $versions[$versionIndex], null, ['Destination' => $this->makeSabrePath($user, $path)]);
+	}
+
 }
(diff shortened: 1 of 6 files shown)


tests/integration/features/dav-versions.feature
```diff
@@ -1,28 +1,55 @@
...
+    And user "user0" uploads file with content "12345" to "/davtest.txt"
+    And the version folder of file "/davtest.txt" for user "user0" contains "1" elements
+    When user "user0" restores version nr "1" of file "/davtest.txt"
+    Then downloading file "davtest.txt"
+    And downloaded content should be "123"
+
+  Scenario: User cannot access meta folder of a file which is owned by somebody else
+    Given user "user1" exists
+    And user "user0" uploads file with content "123" to "/davtest.txt"
+    And we save it into "FILEID"
```
SergioBertolinSG: What is doing this step?

We have functions to get the id of a file, maybe adding this extension rdx/behat-variables can be avoided.

Here we use a variable to store file id:
https://github.com/owncloud/core/blob/master/tests/integration/features/webdav-related-new-endpoint.feature#L358

To get the file id we can use this function:
https://github.com/owncloud/core/blob/master/tests/integration/features/bootstrap/WebDav.php#L1141

tests/integration/features/bootstrap/WebDav.php
```diff
@@ -1168,4 +1179,18 @@ public function userChecksFileIdForPath($user, $path) {
 		$currentFileID = $this->getFileIdForPath($user, $path);
 		PHPUnit_Framework_Assert::assertEquals($currentFileID, $this->storedFileID);
 	}
+
+	/**
+	 * @When user :user restores version nr :versionIndex of file :path
```
SergioBertolinSG: "nr" means number? then it should be "number", no need to put "nr".

SergioBertolinSG (changes requested):

tests/integration/features/dav-versions.feature
DeepDiver1975: Ah - didn't know that .... but this is causing another request to the server. Good? bad? ...

tests/integration/features/bootstrap/WebDav.php
DeepDiver1975: Sounds reasonable - will adjust ... or maybe even code different

tests/integration/features/dav-versions.feature
SergioBertolinSG: I think it is better than adding another dependency.

DeepDiver1975: But I think this step is incomplete 'And user "user0" stores id of file "/textfile0.txt"'

How to save multiple file ids? How to check multiple fileids in follow up steps?

DeepDiver1975: Furthermore I explicitly need to reference the fileid in the url in a later step

```
  Scenario: User cannot access meta folder of a file which is owned by somebody else
    Given user "user1" exists
    And user "user0" uploads file with content "123" to "/davtest.txt"
    And we save it into "FILEID"
    And as an "user1"
    When sending "PROPFIND" with exact url to "/remote.php/dav/meta/<<FILEID>>"
    Then the HTTP status code should be "404"

Comment (SergioBertolinSG) [reply tests/integration/features/dav-versions.feature]: > How to save multiple file ids? How to check multiple fileids in follow up steps?

Currently only works with one.

About the following steps, you could add a new step like

When getting metadata of last file id stored

And keep the last part about the http status.

**Comment by DeepDiver1975** [comment]
> When getting metadata of last file id stored
> 
> And keep the last part about the http status.

but then I have to code a hell of new steps vs. re-using an existing step and get simple string interpolation.

I'm not convinced ...

**Comment by SergioBertolinSG** [comment]
OK, add the dependency then, maybe we can use it in the future.

The problem could be if we upgrade behat version at some point and somehow this extension doesn't work with the newer version, making us fixing it upstream or having to remove it.

File: tests/integration/features/bootstrap/WebDav.php [hunk 3]
```diff
@@ -1168,4 +1179,18 @@ public function userChecksFileIdForPath($user, $path) {
...
 		PHPUnit_Framework_Assert::assertEquals($currentFileID, $this->storedFileID);
 	}
+
+	/**
+	 * @When user :user restores version index :versionIndex of file :path
+	 * @param string $user
+	 * @param int $versionIndex
+	 * @param string $path
+	 */
+	public function userRestoresVersionNrOfFile($user, $versionIndex, $path) {
```
SergioBertolinSG: The function name has the old name.

SergioBertolinSG (changes requested):

tests/integration/features/bootstrap/WebDav.php
DeepDiver1975: k - will adjust .... but it actually has no meaning - right?

DeepDiver1975: I mean the function is detected by the regex in the phpdoc - unlike ruby/cucumber where the function name is used .... anyway

SergioBertolinSG: Yes, just a cosmetic change.

DeepDiver1975: done

SergioBertolinSG (approved):

phil-davis:
@DeepDiver1975 @SergioBertolinSG should these tests be backported?

I noticed because I am looking at ``composer.json`` diffs between ``master`` and ``stable10``.

DeepDiver1975:
> @DeepDiver1975 @SergioBertolinSG should these tests be backported?
> 
> I noticed because I am looking at composer.json diffs between master and stable10.

yes please

phil-davis:
Assigned myself so I find this again. I will backport after the latest composer PR in stable10 gets merged (no point backporting now when I know there will be more conflict to come)

phil-davis:
See discussion in #30429 - backport here is waiting on/dependent on if other related meta dav stuff is backported. Leaving backport label here for now, so it gets noticed every now and then.

phil-davis:
This code is all backported in #31050 which has been merged. Removing the backport-request label.

PR #29643, opened 2017-11-21
Bump version in master to 10.0.4.2

phil-davis:
because the code in master has whatever migrations... might occur up to 10.0.4.2
and it is annoying when doing ``git checkout stable10`` and ``git checkout master`` to find the UI telling that you attempt to downgrade when you switch to master branch.

``stable10`` already has version set to 10.0.4.2


PVince81 (approved):
:+1: thanks

Issue #26357, opened 2016-10-12
Cannot unshare after moving sub-folder share out as non-initiator


PVince81:
### Steps
1. Create three users "user1", "user2", "user3"
2. Login as "user1"
3. Create a folder "base"
4. Share "base" with "user2"
5. Login as "user2"
6. Create a folder "base/sub"
7. Share "base/sub" with "user3" => this makes the share owner be "user1" but the initiator is "user2"
8. Move folder "base/sub" to the root
9. Check if "user3" still receives the folder "sub"
10. Try unsharing "/sub" as "user2" from "user3"
### Expected result

"user3" must still see the folder.
Unsharing works.
### Actual result

Even though the folder "sub" is still marked as shared with "user3", "user3" doesn't see it mounted.
Unsharing from "user3" appears to work in the UI but refreshing the dialog shows the share again.
The only way to make it work again is moving "sub" back into "base/sub"
### Versions

ownCloud v9.1.1

@owncloud/qa we need more of such crazy test cases

PVince81:
I've checked v9.0.4 and only half of the issue exists:
- user3 still doesn't see the mount point even though it's marked as shared
- however it is possible to unshare permanently.

So this is half a regression...

PVince81:
@felixboehm @butonic @cdamken in case people are share recipients for local shares but the share doesn't appear, might be useful to check if it's this scenario.

In the DB it appears like this:
```
MariaDB [owncloud]> select * from oc_share;
+----+------------+------------+-----------+---------------+--------+-----------+-------------+-------------+-------------+-------------+-------------+------------+----------+------------+-------+-----------+
| id | share_type | share_with | uid_owner | uid_initiator | parent | item_type | item_source | item_target | file_source | file_target | permissions | stime      | accepted | expiration | token | mail_send |
+----+------------+------------+-----------+---------------+--------+-----------+-------------+-------------+-------------+-------------+-------------+------------+----------+------------+-------+-----------+
|  1 |          0 | user2      | user1     | user1         |   NULL | folder    | 20          | NULL        |          20 | /base       |          31 | 1476289197 |        0 | NULL       | NULL  |         0 |
|  3 |          0 | user3      | user1     | user2         |   NULL | folder    | 38          | NULL        |          38 | /sub        |          31 | 1476289326 |        0 | NULL       | NULL  |         0 |
[...]

PR #29650, opened 2017-11-22
Unshare fixes

jvillafanez:
## Description
Share ownership problems causes the issue linked below.

## Related Issue
https://github.com/owncloud/core/issues/26357

## How Has This Been Tested?
Manually checked that the PR fixes the linked issue. More tests incoming.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
+++ b/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
@@ -49,12 +49,23 @@ class FileCustomPropertiesBackend extends AbstractCustomPropertiesBackend {
 	const DELETE_BY_ID_AND_NAME_STMT = 'DELETE FROM `*PREFIX*properties`'
 	. ' WHERE `fileid` = ? AND `propertyname` = ?';
 
+	private $moveSource = null;
+
 	/**
 	 * This method is called after a node is deleted.
 	 *
 	 * @param string $path path of node for which to delete properties
 	 */
 	public function delete($path) {
+		$moveSource = $this->moveSource;
+		$this->moveSource = null;
+
+		if ($moveSource === $path) {
+			// trying to delete a file that has been moved -> ignoring because the file
+			// exists in another path
+			return;
+		}
+
 		$node = $this->getNodeForPath($path);
 		if (is_null($node)) {
 			return;
@@ -77,6 +88,7 @@ class FileCustomPropertiesBackend extends AbstractCustomPropertiesBackend {
 	 */
 	public function move($source, $destination) {
 		// Part of interface. We don't care about move because it doesn't affect fileId
+		$this->moveSource = $source;
 	}
 
 	/**
--- a/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
+++ b/apps/dav/lib/DAV/FileCustomPropertiesBackend.php
@@ -49,6 +49,11 @@ class FileCustomPropertiesBackend extends AbstractCustomPropertiesBackend {
 	const DELETE_BY_ID_AND_NAME_STMT = 'DELETE FROM `*PREFIX*properties`'
 	. ' WHERE `fileid` = ? AND `propertyname` = ?';
 
+	/**
+	 * @var string the source path of a move action.
+	 * This is set during a move so the delete action can know if a move has been called before
+	 * in order to not fetch the source node again (which would cause an error)
+	 */
 	private $moveSource = null;
 
 	/**
--- a/apps/files_sharing/lib/Updater.php
+++ b/apps/files_sharing/lib/Updater.php
@@ -78,7 +78,7 @@ class Updater {
 		//Ownership is moved over
 		foreach ($shares as $share) {
 			/** @var \OCP\Share\IShare $share */
-			if ($share->getShareWith() !== $newOwner) {
+			if ($share->getSharedWith() !== $newOwner) {
 				$share->setShareOwner($newOwner);
 				$shareManager->updateShare($share);
 			} else {
--- a/apps/files_sharing/lib/Updater.php
+++ b/apps/files_sharing/lib/Updater.php
@@ -78,8 +78,12 @@ class Updater {
 		//Ownership is moved over
 		foreach ($shares as $share) {
 			/** @var \OCP\Share\IShare $share */
-			$share->setShareOwner($newOwner);
-			$shareManager->updateShare($share);
+			if ($share->getShareWith() !== $newOwner) {
+				$share->setShareOwner($newOwner);
+				$shareManager->updateShare($share);
+			} else {
+				$shareManager->deleteShare($share);
+			}
 		}
 	}
 
--- a/apps/files_sharing/lib/Updater.php
+++ b/apps/files_sharing/lib/Updater.php
@@ -68,10 +68,12 @@ class Updater {
 		$mountManager = \OC::$server->getMountManager();
 		$dstMount = $mountManager->find($src->getPath());
 		if (!($dstMount instanceof \OCA\Files_Sharing\SharedMount)) {
-			return;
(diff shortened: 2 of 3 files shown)


jvillafanez:
* [ ] Moving the share to an external storage without allowing sharing behaves bad:

    ```
    curl -X MOVE -H "Destination: /remote.php/dav/files/qa2/Storage/007" -u qa2:11111 http://10.0.2.8:30080/remote.php/dav/files/qa2/base/sub1
    <?xml version="1.0" encoding="utf-8"?>
    <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
      <s:exception>OCP\Share\Exceptions\GenericShareException</s:exception>
      <s:message>You are not allowed to share /qa2/files/Storage/007</s:message>
    </d:error>
    ```
    The folder (in this case) is moved to the external storage, but the share won't appear in the target user's folder (qa3 in this case). The source folder is still marked as shared

screenshot from 2017-11-22 17 43 22

* [x] Moving the share to an external storage allowing sharing in it works fine
* [ ] Moving the share to another share seems to work, but might need some explanation.

    Visually, the target share's owner (qa3 in this case) seems to be the owner of the files, but the DB still have the old share entry with the previous owner as share's owner.

To be more precise:

First test scenario:
1. Create 3 users: qa1, qa2 and qa3
2. Login with qa1, create a folder "base/sub" (first the parent folder and then the subfolder) and upload some files there.
3. Share "base" folder with qa2
4. Login as qa2
5. As qa2 setup a personal external storage without sharing permissions
6. As qa2 go to the "base" folder (shared by qa1) and share "sub" with qa3.
7. Login as qa3 and make sure qa3 has access to the contents of the "base/sub" folder
8. As qa2 move the "base/sub" folder (which is shared with qa3) to the personal external storage.

Third test scenario:
1. Create 3 users: qa1, qa2 and qa3
2. Login with qa1, create a folder "base/sub" (first the parent folder and then the subfolder) and upload some files there.
3. Share "base" folder with qa2
4. Login as qa2
5. As qa2 go to the base folder and share "sub" with qa3
6. Login as qa3 and verify the "sub" folder is visible
7. As qa3 create a "qa3owner" folder and share it with qa2
8. As qa2 move the "base/sub" folder to the "qa3owner" folder which should be visible for qa2

apps/dav/lib/DAV/FileCustomPropertiesBackend.php
```diff
@@ -49,12 +49,23 @@ class FileCustomPropertiesBackend extends AbstractCustomPropertiesBackend {
 	const DELETE_BY_ID_AND_NAME_STMT = 'DELETE FROM `*PREFIX*properties`'
 	. ' WHERE `fileid` = ? AND `propertyname` = ?';
 
+	private $moveSource = null;
```
PVince81: PHPDoc, what is this ? what type ?

PVince81:
> Moving the share to an external storage without allowing sharing behaves bad:

Likely a separate bug. I'm not sure what the correct behavior should be. Maybe raise a separate ticket with steps to reproduce and we can discuss expected behavior there.

PVince81:
> As qa3 create a "qa3owner" folder and share it with qa2

Is "qa3owner" created in the root of qa3 ?

For the third scenario we need to change the share owner in the database.
I think this already happens for simpler sharing scenarios so the code might exist somewhere.

PVince81:
Code looks good so far

jvillafanez:
I think this is ready for review. As said, I'll open a new ticket for the first scenario. The third scenario should be fixed with the last commit.

There is another ticket to discuss regarding the third scenario.

1. Create 3 users: qa1, qa2 and qa3
2. Login with qa1, create a folder "base/sub" (first the parent folder and then the subfolder) and upload some files there.
3. Share "base" folder with qa2
4. Login as qa2
5. As qa2 go to the base folder and share "sub" with qa3 **and admin** (<--- this is new)
6. Login as qa3 and verify the "sub" folder is visible
7. As qa3 create a "qa3owner" folder and share it with qa2
8. As qa2 move the "base/sub" folder to the "qa3owner" folder which should be visible for qa2

The problem is that qa3 doesn't know that the "qa3owner/sub" folder is shared with another user. There is no visual indicator of that unless the qa3 user clicks on the folder. The same happens with simpler scenarios where a shared file is reshared by another user to a third one.

jvillafanez:
Unittests incoming

PVince81:
> The problem is that qa3 doesn't know that the "qa3owner/sub" folder is shared with another user. There is no visual indicator of that unless the qa3 user clicks on the folder. The same happens with simpler scenarios where a shared file is reshared by another user to a third one.

Can you make yet another ticket for this one ? I observed this behavior before a few days ago. I hope it's only an indicator thing and that the DB structure looks fine already.

apps/files_sharing/lib/Updater.php
```diff
@@ -68,16 +68,22 @@ static private function moveShareToShare($path) {
...
 		}
 
-		$newOwner = $dstMount->getShare()->getShareOwner();
 
 		//Ownership is moved over
 		foreach ($shares as $share) {
 			/** @var \OCP\Share\IShare $share */
-			$share->setShareOwner($newOwner);
-			$shareManager->updateShare($share);
+			if ($share->getShareWith() !== $newOwner) {
```
PVince81: Maybe at some point we'll want a `$shareManager->changeOwner()` with all the above and below logic embedded. At some point in the future we might have such action triggerable over the UI.

This is fine for now.

jvillafanez:
Looks like there won't be new unittests... I'm not sure what is the setup for the UpdaterTest.php tests (files_sharing app) but I'm having a lot of problems trying to move around a shared file from another account.
The rename operation fails with:
```
{"reqId":"eDRNRo3ESbFAY5Nc4PVD","level":3,"time":"2017-11-28T12:16:16+00:00","re
moteAddr":"","user":"test-share-user2","app":"core","method":"--","url":"--","me
ssage":"unable to rename, source directory is not writable : files\/folder_share
_updater_test\/base1"}
```
I've removed the cleanup operations from the tests to check what is the state after the test: DB users are missing, files and shares for the accounts are there. After creating all the required users, the FS state is the expected one, and I can move the shared file around without problems.

I guess it might be better to test these scenarios via integration tests.

PVince81:
> I guess it might be better to test these scenarios via integration tests.

Sounds good, thanks

jvillafanez:
This should be ready to be merged.

Backport to stable10?

PVince81 (approved):
Awesome, thanks :+1:

Please backport to stable10 for 10.0.5

jvillafanez:
Backport in https://github.com/owncloud/core/pull/29716

jvillafanez:
Tickets opened:
https://github.com/owncloud/core/issues/29717
https://github.com/owncloud/core/issues/29718

PVince81:
@jvillafanez codecov is not happy.

Adding tests for the DAV properties thing should be easy.

> Looks like there won't be new unittests... I'm not sure what is the setup for the UpdaterTest.php tests 

Maybe setup another test class that only works with mocks instead of the integration nightmare of the old tests ?

jvillafanez:
> Adding tests for the DAV properties thing should be easy.

I guess not for me... Other than adding the tests just for the coverage, whose worthyness will be under question, I'm not sure how to test the scenario.
```
$fileCustomPropertyBackend->move(....);
$fileCustomPropertyBackend->delete(....);
// checks
```
They'll be "I'm not sure what I'm testing" kind of tests. It makes me wonder if the fix for that problem should be in another place.

> Maybe setup another test class that only works with mocks instead of the integration nightmare of the old tests ?

It's a static class with heavy server component dependencies. Even if the we manage to mock all the required server components, we'll likely need to mock more subcomponents.

PVince81:
Ok, fine

PR #29651, opened 2017-11-22
Add new symfony dispatcher events

sharidas:
Add new symfony dispatcher events

## Description

Add new symfony dispatcher events

## Related Issue

https://github.com/owncloud/core/issues/29174

## Motivation and Context

Adding more symfony dispatcher events.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -112,7 +112,6 @@ class File extends Node implements IFile {
 	public function put($data) {
 		$path = $this->fileView->getAbsolutePath($this->path);
 		return $this->emittingCall(function () use (&$data) {
-			$event = new GenericEvent(null, []);
 			try {
 				$exists = $this->fileView->file_exists($this->path);
 				if ($this->info && $exists && !$this->info->isUpdateable()) {
@@ -276,22 +275,38 @@ class File extends Node implements IFile {
 		}
 		$hookPath = Filesystem::getView()->getRelativePath($this->fileView->getAbsolutePath($path));
 		$run = true;
+		$event = new GenericEvent(null);
 
 		if (!$exists) {
 			\OC_Hook::emit(\OC\Files\Filesystem::CLASSNAME, \OC\Files\Filesystem::signal_create, [
 				\OC\Files\Filesystem::signal_param_path => $hookPath,
 				\OC\Files\Filesystem::signal_param_run => &$run,
 			]);
+			if ($run) {
+				$event->setArgument('run', $run);
+				\OC::$server->getEventDispatcher()->dispatch('file.beforeCreate', $event);
+				$run = $event->getArgument('run');
+			}
 		} else {
 			\OC_Hook::emit(\OC\Files\Filesystem::CLASSNAME, \OC\Files\Filesystem::signal_update, [
 				\OC\Files\Filesystem::signal_param_path => $hookPath,
 				\OC\Files\Filesystem::signal_param_run => &$run,
 			]);
+			if ($run) {
+				$event->setArgument('run', $run);
+				\OC::$server->getEventDispatcher()->dispatch('file.beforeUpdate', $event);
+				$run = $event->getArgument('run');
+			}
 		}
 		\OC_Hook::emit(\OC\Files\Filesystem::CLASSNAME, \OC\Files\Filesystem::signal_write, [
 			\OC\Files\Filesystem::signal_param_path => $hookPath,
 			\OC\Files\Filesystem::signal_param_run => &$run,
 		]);
+		if ($run) {
+			$event->setArgument('run', $run);
+			\OC::$server->getEventDispatcher()->dispatch('file.beforeWrite', $event);
+			$run = $event->getArgument('run');
+		}
 		return $run;
 	}
 
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -41,6 +41,7 @@ use OCA\DAV\Connector\Sabre\Exception\FileLocked;
 use OCA\DAV\Connector\Sabre\Exception\Forbidden as DAVForbiddenException;
 use OCA\DAV\Connector\Sabre\Exception\UnsupportedMediaType;
 use OCP\Encryption\Exceptions\GenericEncryptionException;
+use OCP\Events\EventEmitterTrait;
 use OCP\Files\EntityTooLargeException;
 use OCP\Files\ForbiddenException;
 use OCP\Files\InvalidContentException;
@@ -57,9 +58,11 @@ use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\Exception\NotImplemented;
 use Sabre\DAV\Exception\ServiceUnavailable;
 use Sabre\DAV\IFile;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 class File extends Node implements IFile {
 
+	use EventEmitterTrait;
 	protected $request;
 	
 	/**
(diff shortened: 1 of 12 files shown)


lib/private/Files/View.php
```diff
@@ -260,7 +267,13 @@ public function getLocalFolder($path) {
 	 * for \OC\Files\Storage\Storage via basicOperation().
 	 */
 	public function mkdir($path) {
-		return $this->basicOperation('mkdir', $path, ['create', 'write']);
+		$result = $this->basicOperation('mkdir', $path, ['create', 'write']);
+		$event = new GenericEvent(null, [
+			'path' => $this->getHookPath($path),
+			'fileinfo' => $this->getFileInfo($path)
+		]);
+		$this->eventDispatcher->dispatch('\OC\Filesystem::create', $event);
```
PVince81: Did you check whether we could trigger the dispatcher directly inside of basicOperation ?
I'd like to avoid having an extra call to `getFileInfo` for the events because it will cause more DB calls even if no listener is created. Please check whether basicOperation already has the file info object available somewhere for reuse.

also we used to have pre and post hooks, would be good to also have those as dispatcher event. Goal is that at some time in the future these could replace the old hooks, so we need to be close to the original functionality.

lib/private/Files/View.php
```diff
@@ -349,6 +369,10 @@ public function rmdir($path) {
 			$internalPath = $mount->getInternalPath($absolutePath);
 			$storage->getUpdater()->remove($internalPath);
 		}
+
+		if ($result) {
+			$this->eventDispatcher->dispatch('\OC\Filesystem::delete_completed', $event);
```
PVince81: is there a pre-delete event ?

maybe let's settle on names, howabout beforeDelete, afterDelete ? beforeCreate, afterCreate ?

@DeepDiver1975 thoughts ?

lib/private/Files/View.php
```diff
@@ -333,12 +346,19 @@ protected function renameUpdate(Storage $sourceStorage, Storage $targetStorage,
...
 	 */
 	public function rmdir($path) {
+		$event = new GenericEvent(null);
 		$absolutePath = $this->getAbsolutePath($path);
 		$mount = Filesystem::getMountManager()->find($absolutePath);
 		if ($mount->getInternalPath($absolutePath) === '') {
 			return $this->removeMount($mount, $absolutePath);
 		}
 		if ($this->is_dir($path)) {
+			$event->setArgument('path', $path);
```
PVince81: if we have file info we likely don't need the path as it's already contained inside the file info object

lib/private/Files/View.php
```diff
@@ -333,12 +346,19 @@ protected function renameUpdate(Storage $sourceStorage, Storage $targetStorage,
...
+			try {
+				$event->setArgument('fileinfo', $this->getFileInfo($path));
+			} catch (\Exception $e) {
+				$this->logger->warning(__METHOD__ . " failed to retrieve file info for $path");
```
PVince81: if we don't log the exception we'll never know what really happened there... we have `$this->logger->logException()`

PVince81 (commented):
See my comments above.

I wonder whether it's possible for event dispatcher to first ask whether there is any listener before retrieving potentially expensive operation.

lib/private/Files/View.php
sharidas: Unfortunately I couldn't find call to `getFileInfo` at basicOperation or FileInfo object. This was bothering me too as each call becomes expensive, whether there are listeners or not :(

Sure I will add post and pre events.

PVince81: Hmmm that's sad then. As an alternative we could rather pass in the file id only and let the listener get the file info by themselves. The question is how to get it in an efficient manner.

The public API states that one should use `\OC::$server->getUserFolder()->getById($fileId)` but that's a rather expensive call. The other alternative would be to inject the `View` and the `$path` and let the subscriber resolve it. However the `View` is not supposed to be public API...

bringing in @jvillafanez for further ideas

lib/private/Files/View.php
```diff
@@ -94,6 +95,10 @@ class View {
 
 	private $userManager;
 
+	private $eventDispatcher;
```
DeepDiver1975: PHPDoc

lib/private/Files/View.php
```diff
@@ -94,6 +95,10 @@ class View {
...
+
+	private $logger;
```
DeepDiver1975: PHPDoc

lib/private/Files/View.php
DeepDiver1975: agreed - sounds great

lib/private/Files/View.php
```diff
@@ -561,6 +585,15 @@ public function touch($path, $mtime = null) {
...
 	public function file_get_contents($path) {
+		$event = new GenericEvent(null, [
+			'path' => $path
+		]);
+		try {
+			$event->setArgument('fileinfo', $this->getFileInfo($path));
+		} catch (\Exception $e) {
+			$this->logger->warning(__METHOD__ . " failed to retrieve file info for $path");
+		}
+		$this->eventDispatcher->dispatch('\OC\Filesystem::read_completed', $event);
```
DeepDiver1975: this is not read_completed but beforeRead

lib/private/Server.php
```diff
@@ -304,6 +304,11 @@ public function __construct($webRoot, \OC\Config $config) {
 			$userSession->listen('\OC\User', 'postCreateUser', function ($user, $password) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_createUser', ['uid' => $user->getUID(), 'password' => $password]);
+				$event = new GenericEvent(null, [
+					'uid' => $user->getUID(),
+					'password' => $password
+				]);
+				\OC::$server->getEventDispatcher()->dispatch('\OC\User\Manager::createUser', $event);
```
DeepDiver1975: @PVince81 I'd expect here to have OCP namespace .....

DeepDiver1975: but we might want to stick to the symfony event naming convention? see https://symfony.com/doc/current/components/event_dispatcher.html#naming-conventions

lib/private/User/User.php
```diff
@@ -258,6 +258,8 @@ public function setPassword($password, $recoveryPassword = null) {
 			if ($result) {
 				if ($this->emitter) {
 					$this->emitter->emit('\OC\User', 'postSetPassword', [$this, $password, $recoveryPassword]);
+					$event = new GenericEvent(null, ['user' => $this]);
```
DeepDiver1975: should we fully kill the emitter concept as we are working on this as of now? All emitter interfaces never hit the public api - we can kill it right away @PVince81 thoughts

DeepDiver1975 (changes requested):

lib/private/Files/View.php
sharidas: I wanted to make the call after basicOperation. So after the call to basicOperation its read_completed. I will update the PR accordingly.

lib/private/User/User.php
PVince81: > All emitter interfaces never hit the public api

Really ? I remember that @patrickjahns's example app was using the emitter events from the public files Node API. I'm sure other apps do so to.

Once we have symfony events for everything, we can first deprecate all hooks and emitter events and give app devs time to migrate. Only then we can kill these completely.

lib/private/Server.php
PVince81: ok, so change the format again ? we already settled on this OC namespace + event id for many other events.

Anyway, I'm ok switching convention now as I prefer to refer to documented stuff.

> I'd expect here to have OCP namespace

This would imply that there are also private symfony events. Not sure if we want private events any more ?

lib/private/User/User.php
DeepDiver1975: emitters are private - we can kill them right away
hooks is a different story - yes

PVince81: Ok, so emitters that are triggering events from a private object (`\OC` namespace) could be replaced with using symfony events. But this is out of scope for this PR.

The purpose of this PR is to add new events without causing any regressions.
Killing/replacing anything will require careful regression testing and potentially adding new tests where tests are missing.

PVince81: Raised here https://github.com/owncloud/core/issues/29672

lib/private/Server.php
sharidas: @DeepDiver1975 @PVince81  how about "ocp_iusermanager.beforeCreate" for https://github.com/owncloud/core/blob/master/lib/private/User/Manager.php#L323 ( as an example )?
Instead of using namespace with slashes "/" use "_" and all letters remain in small case as per document.

lib/private/User/User.php
DeepDiver1975: from an implementation point of view - this can help to streamline the code base and avoid duplications:

```php
private emittingCall(\Closure $fn, $class, $eventName) {
    $this->eventDispatcher->dispatch("$class::before$eventName");
    $result = $fn();
    $this->eventDispatcher->dispatch("$class::after$eventName");
    return $result;
}

// within the implementation of e.g. create user

private function createUser() {
    return emittingCall( function() use($userManager) {
        return $userManager->createUser();
    }, '\OCP\IUserManager', 'create');
}

File: lib/private/Server.php (hunk 8 above)
Comment (karakayasemi) [reply lib/private/Server.php]: IMHO, we can add all the events name as a constant of the relevant class. For example in the \OCP\User class we can add
```php
const EventCreate = 'ocp_user.create';
const EventDelete = 'ocp_user.delete';
```
In this way we can access it via \OCP\User::EventCreate or directly event name string. It also makes easier to find the class events for developers.

PVince81: "iusermanager" here doesn't make sense. Whoever is listening to the event shouldn't need to know where the event come from. The only thing a listener is interested in is the type of event and what it applies to.

So I think I prefer "user.beforeCreate" and "user.afterCreate". This makes it clear that it's about a user being created or deleted. What component of ownCloud does it (user manager or whatever) doesn't matter. And I don't like ocp when used this way.

PVince81: Regarding the constants, I'm not even sure whether we should have the constants directly in the `\OCP\User` object. Should we have a more global location for all user and group management related event constants ? Depending on specific classes feels too close to the old emitter pattern.

lib/private/Files/View.php
jvillafanez: What about using custom events instead of the default one? The custom event could have the `View` object inside (or any other needed dependency) and provide custom methods to access to the object depending on what we want to expose.
(Info in https://symfony.com/doc/current/components/event_dispatcher.html#creating-an-event-class)

sharidas: @DeepDiver1975 @PVince81 So with the current state of the patch, if we add a custom event as per @jvillafanez , then we can send `View` object and `$path1` and/or `$path2`. And in the trait, for `before` and `after` events we can use the data from this custom event. And the subscriber of the event would resolve it.

sharidas:
@DeepDiver1975 @butonic Thoughts regarding  ( discussed the same with @PVince81 
 ) https://github.com/owncloud/core/issues/29174#issuecomment-347457541

lib/private/Files/View.php
PVince81: `View` is considered private API. If we were to pass something it would need to be the current folder from  `getUserFolder()->get($path)` which is likely expensive to resolve. Note that said folder is from the `\OCP\Files\Node` interface which itself implements/embeds the `FileInfo`.

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -107,145 +110,150 @@ public function __construct($view, $info, $shareManager = null, Request $request
...
-			
-			$this->refreshInfo();
-		} catch (StorageNotAvailableException $e) {
-			throw new ServiceUnavailable("Failed to check file size: " . $e->getMessage());
-		}
 
-		return '"' . $this->info->getEtag() . '"';
+			return '"' . $this->info->getEtag() . '"';
+		}, [
+			'before' => ['path' => $this->fileView->getAbsolutePath($this->path)],
```
PVince81: store the result of getAbsolutePath into a variable instead of calling it twice ? not sure how expensive the actual call is

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -266,22 +274,35 @@ private function emitPreHooks($exists, $path = null) {
...
+		$event = new GenericEvent(null);
 
 		if (!$exists) {
 			\OC_Hook::emit(\OC\Files\Filesystem::CLASSNAME, \OC\Files\Filesystem::signal_create, [
 				\OC\Files\Filesystem::signal_param_path => $hookPath,
 				\OC\Files\Filesystem::signal_param_run => &$run,
 			]);
+			if ($run) {
+				$event->setArgument('run', $run);
+				\OC::$server->getEventDispatcher()->dispatch('file.create.allowed', $event);
```
PVince81: what does "allowed" mean here ?

didn't we agree on using something like "file.beforeCreate" and "file.afterCreate" ?

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -266,22 +274,35 @@ private function emitPreHooks($exists, $path = null) {
...
+				$event->setArgument('run', $run);
+				\OC::$server->getEventDispatcher()->dispatch('file.update.allowed', $event);
+			}
 		}
 		\OC_Hook::emit(\OC\Files\Filesystem::CLASSNAME, \OC\Files\Filesystem::signal_write, [
 			\OC\Files\Filesystem::signal_param_path => $hookPath,
 			\OC\Files\Filesystem::signal_param_run => &$run,
 		]);
+		if ($run) {
+			$event->setArgument('run', $run);
```
PVince81: if $run is modified by the listener, you need to read out the modified value again from the event object

PVince81: better add a test for this: if a dispatcher listener sets the run value to false, then the main method will also return false

lib/private/Server.php
PVince81: let's use `user.beforeCreate` and `user.afterCreate` here

lib/private/Server.php
```diff
@@ -323,21 +334,39 @@ public function __construct($webRoot, \OC\Config $config) {
...
 			$userSession->listen('\OC\User', 'changeUser', function ($user, $feature, $value) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'changeUser', ['run' => true, 'user' => $user, 'feature' => $feature, 'value' => $value]);
+				$this->emittingCall(function () use (&$user, &$feature, &$value) {
+					$event = new GenericEvent(null, [
+						'user' => $user,
+						'feature' => $feature,
+						'value' => $value
+					]);
+					\OC::$server->getEventDispatcher()->dispatch('user.afterfeaturechange.data', $event);
```
PVince81: I see you still have the ".data" mode here, I guess you didn't want to modify this until the file one got reviewed ?

PVince81 (commented):
Some small changes needed for the file hooks especially the part about the `$run` logic.

Seems you haven't touched the user manager yet, please adjust afterwards to use the same patterns.

DeepDiver1975:
has conflicts

sharidas:
> Seems you haven't touched the user manager yet, please adjust afterwards to use the same patterns.
Guess this change got missed for review: https://github.com/owncloud/core/pull/29651/files#diff-2ad387ea4134715ffda3bb0f12140f91R293

apps/dav/tests/unit/Connector/Sabre/FileTest.php
```diff
@@ -491,6 +513,41 @@ public function testPutSingleFileTriggersHooks() {
...
+			$calledWriteAllowed[] = 'file.beforeWrite';
+			//Now modify run
+			$event->setArgument('run', false);
+			array_push($calledWriteAllowed, $event);
+		});
+
+		try {
+			$this->doPut('/testingfile.txt');
+		} catch (\Exception $e) {
+			$this->assertInstanceOf(Exception::class, $e);
```
PVince81: is there no specific exception for this ?

Also use `@expectedException` in the PHPDoc for PHPUnit to check for this exception.
The problem with this code is that if the exception is not thrown at all the test still passes, so would need an additional check, or have the assert outside... Best is to use `@expectedException` instead in this situation with the proper exception class name. There is also a tag for the exception message if that makes sense.

lib/private/Files/View.php
```diff
@@ -570,40 +592,58 @@ public function file_get_contents($path) {
...
 	protected function emit_file_hooks_pre($exists, $path, &$run) {
+		$event = new GenericEvent(null);
 		if (!$exists) {
 			\OC_Hook::emit(Filesystem::CLASSNAME, Filesystem::signal_create, [
 				Filesystem::signal_param_path => $this->getHookPath($path),
 				Filesystem::signal_param_run => &$run,
 			]);
+			if ($run) {
+				$event->setArgument('run', $run);
+				$this->eventDispatcher->dispatch('file.beforeCreate', $event);
```
PVince81: here you also forgot to check the value of `$run`. Please recheck all occurrences for this logic.

lib/private/Files/View.php
```diff
@@ -570,40 +592,58 @@ public function file_get_contents($path) {
...
+			}
 		} else {
 			\OC_Hook::emit(Filesystem::CLASSNAME, Filesystem::signal_update, [
 				Filesystem::signal_param_path => $this->getHookPath($path),
 				Filesystem::signal_param_run => &$run,
 			]);
+			if ($run) {
+				$event->setArgument('run', $run);
+				$this->eventDispatcher->dispatch('file.beforeUpdate', $event);
```
PVince81: re-read the last value of `$run` from the argument.

Now I find it strange that the value doesn't seem to be used below and is just passed again...

PVince81: so I suggest to just re-read the value from the event argument to keep the old logic working as it was before...

tests/lib/Events/EventEmitterTraitTest.php
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	/**
+	 * @dataProvider dummyFunction
+	 */
+	public function testEmittingCall($className, $eventName, $data) {
+		$calledAfterEvent = [];
+		\OC::$server->getEventDispatcher()->addListener("$className.after$eventName", function ($event) use (&$calledAfterEvent, &$className, &$eventName) {
+			$calledAfterEvent[] = "$className.after$eventName";
+			array_push($calledAfterEvent, $event);
```
PVince81: why not just use `$calledAfterEvent[] = $event` to push to array ? changing the code style in the middle looks weird.

same comment applies to all other `array_push` locations

PVince81 (commented):
Looks better, getting close. Please check my latest comments.

apps/dav/tests/unit/Connector/Sabre/FileTest.php
sharidas: Indeed it does have exception \Sabre\DAV\Exception. There was a failure observed when I tried with `@expectedException`. Yesterday I spent sometime with the same. The error of exception failure was observed only in Jenkins. Where as when I used `@expectedException` in local machine it never caused a problem. Yes I used `@expectExceptionMessage` for message in the same version of the PR which had `@expectedException`.

lib/private/Files/View.php
sharidas: I did verified with file operations like create, modify/update, delete, rename operations in web ui. At least in that context no damage happened. The problem might appear, when the event listener alters `$run` value. But as of now the alteration happens only in tests so, I guess we are safe for now. @PVince81 Let me know if this change looks ok. If you really feel that just re-read value is sufficient, then I can do that.

PVince81: Please just re-read the value, for consistency with the other code.

sharidas: @PVince81 Ok, I have updated the PR, by just re-reading the value.

PVince81 (approved):
:+1:

PVince81:
restarted drone build

PVince81:
@sharidas please backport

sharidas:
Backport to stable10: https://github.com/owncloud/core/pull/29939

PR #29669, opened 2017-11-23
Unify the core sync logic

tomneedham:
remove some code duplication and get the core to run a proper sync on the user metadata during a login

@DeepDiver1975 @PVince81 note that the account table introduced a change in behavior:
`setDisplayName()` was no longer called on the backend while `canChangeDisplayName()` would still check the backend if it implemented that function. With the account table we can always change display name, email, avatar ... but we may want to prevent that. `canChangeDisplayName()` already contains a check on the system value `allow_user_to_change_display_name`. `canChangeAvatar()` and `canChangePassword()` already exist, but email is missing.

Conceptually there are now two cases for setDisplayName:
1. End user triggered in the UI. Changeability may depend on the backend, eg LDAP is readonly.
2. Sync triggered.

The current User object code is intended only for the first case.

@tomneedham we need to add a codepath for the second case. AFAICT we should directly work on the accounts... the sync case bypasses any end user limitations.

### DIFF
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -78,16 +78,21 @@ class Manager extends PublicEmitter implements IUserManager {
 	/** @var AccountMapper */
 	private $accountMapper;
 
+	/** @var SyncService */
+	private $syncService;
+
 	/**
 	 * @param IConfig $config
 	 * @param ILogger $logger
 	 * @param AccountMapper $accountMapper
+	 * @param SyncService $syncService
 	 */
-	public function __construct(IConfig $config, ILogger $logger, AccountMapper $accountMapper) {
+	public function __construct(IConfig $config, ILogger $logger, AccountMapper $accountMapper, SyncService $syncService) {
 		$this->config = $config;
 		$this->logger = $logger;
 		$this->accountMapper = $accountMapper;
 		$this->cachedUsers = new CappedMemoryCache();
+		$this->syncService = $syncService;
 		$cachedUsers = &$this->cachedUsers;
 		$this->listen('\OC\User', 'postDelete', function ($user) use (&$cachedUsers) {
 			/** @var \OC\User\User $user */
@@ -100,12 +105,14 @@ class Manager extends PublicEmitter implements IUserManager {
 	 *
 	 * @param AccountMapper $mapper
 	 * @param array $backends
+	 * @param SyncService $syncService
 	 * @return array
 	 */
-	public function reset(AccountMapper $mapper, $backends) {
-		$return = [$this->accountMapper, $this->backends];
+	public function reset(AccountMapper $mapper, $backends, $syncService) {
+		$return = [$this->accountMapper, $this->backends, $this->syncService];
 		$this->accountMapper = $mapper;
 		$this->backends = $backends;
+		$this->syncService = $syncService;
 		$this->cachedUsers->clear();
 
 		return $return;
@@ -220,13 +227,7 @@ class Manager extends PublicEmitter implements IUserManager {
 			if ($backend->implementsActions(Backend::CHECK_PASSWORD)) {
 				$uid = $backend->checkPassword($loginName, $password);
 				if ($uid !== false) {
-					try {
-						$account = $this->accountMapper->getByUid($uid);
-					} catch(DoesNotExistException $ex) {
-						$account = $this->newAccount($uid, $backend);
-					}
-					$this->cachedUsers->remove($account->getUserId());
-					// TODO always sync account with backend here to update displayname, email, search terms, home etc. user_ldap currently updates user metadata on login, core should take care of updating accounts on a successful login
+					$account = $this->syncService->createOrSyncAccount($uid, $backend);
 					return $this->getUserObject($account);
 				}
 			}
@@ -292,7 +293,7 @@ class Manager extends PublicEmitter implements IUserManager {
 	 * @param string $uid
 	 * @param string $password
 	 * @throws \Exception
-	 * @return bool|\OC\User\User the created user or false
+	 * @return bool|IUser the created user or false
 	 */
 	public function createUser($uid, $password) {
 		return $this->emittingCall(function () use (&$uid, &$password) {
@@ -337,16 +338,16 @@ class Manager extends PublicEmitter implements IUserManager {
 			if (empty($this->backends)) {
 				$this->registerBackend(new Database());
 			}
+
 			foreach ($this->backends as $backend) {
 				if ($backend->implementsActions(Backend::CREATE_USER)) {
 					$backend->createUser($uid, $password);
-					$account = $this->newAccount($uid, $backend);
-					$this->cachedUsers->remove($account->getUserId());
-					$user = $this->getUserObject($account);
-					$this->emit('\OC\User', 'postCreateUser', [$user, $password]);
-					return $user;
+					$user = $this->createUserFromBackend($uid, $password, $backend);
+					return $user === null ? false : $user;
 				}
 			}
+
+
 			return false;
 		}, ['before' => ['uid' => $uid], 'after' => ['uid' => $uid, 'password' => $password]], 'user', 'create');
 	}
@@ -355,12 +356,16 @@ class Manager extends PublicEmitter implements IUserManager {
 	 * @param string $uid
 	 * @param UserInterface $backend
 	 * @return IUser | null
+	 * @deprecated core is responsible for creating accounts, see user_ldap how it is done
 	 */
 	public function createUserFromBackend($uid, $password, $backend) {
 		return $this->emittingCall(function () use (&$uid, &$password, &$backend) {
-			$this->emit('\OC\User', 'preCreateUser', [$uid, '']);
-			$account = $this->newAccount($uid, $backend);
-			$this->cachedUsers->remove($account->getUserId());
+			$this->emit('\OC\User', 'preCreateUser', [$uid, $password]);
+			try {
+				$account = $this->syncService->createOrSyncAccount($uid, $backend);
+			} catch (\InvalidArgumentException $e) {
+				return null; // because that's what this method should do
+			}
 			$user = $this->getUserObject($account);
 			$this->emit('\OC\User', 'postCreateUser', [$user, $password]);
 			return $user;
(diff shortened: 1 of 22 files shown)


DeepDiver1975:
Tests hate you 😢

lib/private/User/Manager.php
```diff
@@ -216,10 +221,13 @@ public function checkPassword($loginName, $password) {
 				if ($uid !== false) {
 					try {
 						$account = $this->accountMapper->getByUid($uid);
+						$this->syncService->setupAccount($account, $backend, $account->getUserId());
```
butonic: - [ ] move to a different place that is allso executed after login via apache / saml

tomneedham:
`CacheTest` seems to not find the displayname on a test user.. weird

tomneedham:
I most definitely just broke stuff - but I hope the logic is right.... LDAP login works, DB users work, saml seems to die.

tomneedham:
screen shot 2017-11-29 at 20 54 36
Safe to say stuff broke...

tomneedham:
Ok - fixed saml. But, for some reason the home path is not set so when you login with a new user they done have a home and you get some weird error about symlinks

tomneedham:
Fixed to use default home path if backend does not provide one. 

However - this _could_ break the home path during a sync if the backend suddently cannot provide a path or provides an invalid path.... I will add a check to not reset it

tomneedham:
- [ ] core should validate emails before setting them to oc_accounts.email

tomneedham:
- [ ] Might have killed some pre and post create user hooks

tomneedham:
Since the syncservice now does user creation, maybe the hooks wont be connected properly anymore...

tomneedham:
I did slightly re-order the hooks here, preCreateUser is now called after the validatePassword GenericEvent...

lib/private/User/Session.php
```diff
@@ -554,18 +555,34 @@ private function loginWithToken($token) {
...
 	 * @return bool
+	 * @throws LoginException
 	 */
-	public function loginWithApache(\OCP\Authentication\IApacheBackend $backend) {
+	public function loginWithApache(IApacheBackend $backend) {
 
 		$uid = $backend->getCurrentUserId();
-		if ($uid === $this->getUser()->getUID()) {
+
+		if ($this->getUser() != null && $uid === $this->getUser()->getUID()) {
```
tomneedham: If session is active and current session uid matches the request user id then we are authenticated and have nothing more to do

lib/private/User/Session.php
```diff
@@ -554,18 +555,34 @@ private function loginWithToken($token) {
...
 		}
 		$this->session->regenerateId();
 		//$run = true;
 		//OC_Hook::emit("OC_User", "pre_login", ["run" => &$run, "uid" => $uid]);
 		$this->manager->emit('\OC\User', 'preLogin', [$uid, '']);
 
+		// Die here if not valid
+		if(!$backend->isSessionActive()) {
+			return false;
+		}
```
tomneedham: This is redundant. OC_User::handleApacheAuth will only call this method if there is already an active IApacheBackend with an active session

lib/private/User/Session.php
```diff
@@ -554,18 +555,34 @@ private function loginWithToken($token) {
...
+
+		// Now we try to create the account or sync
+		/** @var SyncService $syncService */
+		$syncService = \OC::$server->query('SyncService');
+		try {
+			$syncService->createOrSyncAccount($uid, $backend);
+		} catch (\InvalidArgumentException $e) {
+			throw new LoginException('Uid collision2');
```
tomneedham: remove the 2

lib/private/User/Session.php
```diff
@@ -554,18 +555,34 @@ private function loginWithToken($token) {
 	 *
 	 * @param \OCP\Authentication\IApacheBackend $backend
 	 * @return bool
+	 * @throws LoginException
 	 */
-	public function loginWithApache(\OCP\Authentication\IApacheBackend $backend) {
+	public function loginWithApache(IApacheBackend $backend) {
 
 		$uid = $backend->getCurrentUserId();
```
tomneedham: This can throw a HintException, should we catch and show as LoginException?

lib/private/User/SyncService.php
```diff
@@ -178,15 +178,49 @@ public function syncAccount(Account $a, UserInterface $backend, $uid) {
...
+			$account = $this->mapper->getByUid($uid);
+		} catch (DoesNotExistException $e) {
+			// Create a new account for this uid and backend pairing and sync
+			$account = $this->createNewAccount(get_class($backend), $uid);
+		}
+
+		// Check the backend is the same
+		if($account->getBackend() != get_class($backend)) {
+			// Then we have a uid collision
+			$this->logger->info("Uid collision when uid: $uid attempted to login");
```
tomneedham: log the exception

lib/private/User/SyncService.php
```diff
@@ -178,15 +178,49 @@ public function syncAccount(Account $a, UserInterface $backend, $uid) {
...
+			throw new \InvalidArgumentException('Uid collision1');
```
tomneedham: remove the 1

lib/private/User/Manager.php
```diff
@@ -353,9 +346,11 @@ public function createUser($uid, $password) {
 	 */
 	public function createUserFromBackend($uid, $password, $backend) {
 		$this->emit('\OC\User', 'preCreateUser', [$uid, '']);
```
tomneedham: password should be passed through here

lib/private/User/Manager.php
```diff
@@ -353,9 +346,11 @@ public function createUser($uid, $password) {
 	 */
 	public function createUserFromBackend($uid, $password, $backend) {
```
tomneedham: deprecate interface, we dont want backends to be creating their own accounts whenever they fancy

lib/private/User/Manager.php
```diff
@@ -353,9 +346,11 @@ public function createUser($uid, $password) {
...
 		$this->emit('\OC\User', 'preCreateUser', [$uid, '']);
-		$account = $this->syncService->createNewAccount($backend, $uid);
-		$this->syncService->syncAccount($account, $backend, $uid);
-		$this->accountMapper->insert($account);
+		try {
+			$account = $this->syncService->createOrSyncAccount($uid, get_class($backend));
+		} catch (\InvalidArgumentException $e) {
+			return null; // because thats what this method shoudl do apparently
```
tomneedham: add comment that this is logged already in the sync service

lib/private/User/Session.php
```diff
@@ -557,9 +559,26 @@ private function loginWithToken($token) {
...
+			&& count($uidAndBackend) === 2
+			&& $uidAndBackend[0] !== ''
+			&& $uidAndBackend[0] !== null
+			&& $uidAndBackend[1] instanceof UserInterface
+		) {
+			list($uid, $backend) = $uidAndBackend;
+		} else if (is_string($uidAndBackend)) {
+			$uid = $uidAndBackend;
+			if ($apacheBackend instanceof UserInterface) {
+				$backend = $apacheBackend;
```
butonic: ,ore logging in case something goes wrong

lib/base.php
```diff
@@ -724,6 +725,23 @@ public static function init() {
...
+			$userSession->listen('\OC\User', 'postLogin', function (\OCP\IUser $user) {
+				/** @var \OC\User\SyncService $syncService */
+				$syncService = self::$server->query('SyncService');
+				$account = self::$server->getAccountMapper()->getByUid($user->getUID());
+				$userBackend = $account->getBackendInstance();// Get the backend for a user
+				$syncService->syncAccount($account, $userBackend, $user->getUID());
+				self::$server->getAccountMapper()->update($account);
+			});
+		}
+	}
```
tomneedham: should now be able to remove this as sync is handled when each of the authentication types is attempted

lib/private/User/Session.php
tomneedham: Done.

tomneedham:
Still two test failures that look like they are to do with how the `UserTrait` confgiures a fake backend for the usermanager..

tomneedham:
Ah got it, the sync service isnt using the memory account mapper

tomneedham:
Added sync service to the server container. Added a method to override the mapper so unit tests can mock this or replace with the memory account mapper.

tomneedham:
Remaining problems are also related to no using the mock/dummy accountMapper - the sync service needs to know about this as well.

tomneedham:
rebased

tomneedham:
Dammit - this dav PR replaces IConfig in the server contain which breaks the sync service - why on earth does it replace the whole config.... https://github.com/owncloud/core/blob/ce4a435ed27db6f1ad4d123271898ea89be99744/apps/dav/tests/unit/Connector/Sabre/RequestTest/PartFileInRootUploadTest.php#L46

tomneedham:
Like some weird hash of unit and integration test, but not actually being either.

tomneedham:
Mmmm pushed a small hack for that test - @DeepDiver1975 is this ok?

tomneedham:
- [ ] TODO fix this line as well https://github.com/owncloud/core/blob/ce4a435ed27db6f1ad4d123271898ea89be99744/lib/private/legacy/api.php#L381

lib/private/User/SyncService.php
```diff
@@ -125,65 +130,188 @@ public function run(\Closure $callback) {
...
+			$this->logger->debug(
+				"Setting quota for <$uid> to <$quota>", ['app' => self::class]
+			);
+		}
+	}
+
+	private function syncHome(Account $a, UserInterface $backend) {
+		// Home is handled differently, it should only be set on account creation, when there is no home already set
+		// Otherwise it could change on a sync and result in a new user folder being created
+		if($a->getHome() === null) {
```
jvillafanez: I'm not sure if we should do something if the home is different between the account and the backend. As annoying as it will likely be, a warning message saying that we're refusing to update the home folder might be a good idea, specially if the admin expects the home to be updated. Since we don't have any command to change the home folder of a user, I think this is enough; the admin should revert the change in the backend.

lib/private/User/Session.php
```diff
@@ -551,6 +560,94 @@ private function loginWithToken($token) {
...
+			$request = OC::$server->getRequest();
+			$this->createSessionToken($request, $uid, $uid);
+
+			// setup the filesystem
+			OC_Util::setupFS($uid);
+			// first call the post_login hooks, the login-process needs to be
+			// completed before we can safely create the users folder.
+			// For example encryption needs to initialize the users keys first
+			// before we can create the user folder with the skeleton files
+			//OC_Hook::emit("OC_User", "post_login", ["uid" => $uid, 'password' => '']);
```
butonic: @tomneedham wrong hook? remove line?

lib/private/User/SyncService.php
tomneedham: Makes sense to at least output something here - because the admin should be aware.

lib/private/User/Session.php
tomneedham: 👍

tomneedham:
Started writing tests. Couldn't mock $backend->getHome because it doesnt exist in `UserInterface`. Talked to @DeepDiver1975 - so just went the whole way and added interfaces for providing a home and providing a displayname.

tomneedham:
So... I probably broke more stuff.

butonic (approved):

lib/private/User/Session.php
```diff
@@ -557,6 +566,91 @@ private function loginWithToken($token) {
...
+
+		$this->manager->emit('\OC\User', 'preLogin', [$uid, '']);
+
+		// Die here if not valid
+		if(!$apacheBackend->isSessionActive()) {
+			return false;
+		}
+
+		// Now we try to create the account or sync
+		$this->userSyncService->createOrSyncAccount($uid, $backend);
```
mrow4a: This means each new login with result in sync, right?

lib/private/User/SyncService.php
```diff
@@ -125,65 +132,197 @@ public function run(\Closure $callback) {
...
+			}
 		}
+	}
+
+	/**
+	 * @param Account $a
+	 * @param UserInterface $backend of the user
+	 * @return Account
+	 */
+	public function syncAccount(Account $a, UserInterface $backend) {
```
mrow4a: Isnt it 10 calls to LDAP just for one login? @DeepDiver1975 @butonic @tomneedham

mrow4a: I guess no way it can be 1 call right?

mrow4a:
Well have to admin, this PR cleaned up a lot of things... code looks now... like a code ;d

lib/private/User/SyncService.php
```diff
@@ -89,32 +93,35 @@ public function getNoLongerExistingUsers(\Closure $callback) {
...
+							"User <$uid> already provided by another backend({$a->getBackend()} != $backendClass), skipping.",
 							['app' => self::class]
 						);
 						continue;
 					}
-					$a = $this->setupAccount($a, $uid);
+					$this->syncAccount($a, $backend);
 					$this->mapper->update($a);
 				} catch(DoesNotExistException $ex) {
-					$a = $this->createNewAccount($uid);
```
phil-davis: change in this line fixed/reverted by https://github.com/owncloud/core/pull/30367

PR #29646, opened 2017-11-21
Be more precise when refuse to expire

VicDeo:
Change the wording to be more clear

## Description
Change the wording to be more clear

## Related Issue
https://github.com/owncloud/core/issues/28509

## Motivation and Context
Unclear output

## How Has This Been Tested?
By reading the message

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29671, opened 2017-11-24
[stable10] Be more precise when refuse to expire

DeepDiver1975:
backport of #29646


PVince81 (approved):
:+1:

don't merge before stable10 is open again

PR #29676, opened 2017-11-24
Fix import of federated shares

tomneedham:
Fixes https://github.com/owncloud/core/issues/29675

@individual-it @phil-davis Can we add a UI test for this? It's federated sharing, not sure what the coverage is on this.

PR #29677, opened 2017-11-24
[stable10] Fix federated share import

tomneedham:
Backport of https://github.com/owncloud/core/pull/29676 to stable10.

Tested by adding a share and it appearing correctly in the file list, no error log.


PVince81 (approved):
:+1:

To be considered for 10.0.4 RC2 as it's a broken feature

PR #29684, opened 2017-11-26
Integration test code - remove unused use statements and sort

phil-davis:
## Description
Remove unused use statements from integration test code, and sort the remaining use statements for ease of use. IDE refactoring "Code - optimize imports" makes it a 10 second job :)

## Motivation and Context
Cleanup unused code.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29686, opened 2017-11-26
[stable10] Integration test code - remove unused use statements and sort

phil-davis:
Backport #29684


individual-it (approved):

phil-davis:
I will leave this here for a little while, because there are other backports that have integration test code changes that would be nicer to go in before this. (which is what caused the backport issue here where I needed to make the 2nd commit)

DeepDiver1975 (approved):

PR #29699, opened 2017-11-28
Integration tests - for file exists step, throw exception if the file…

phil-davis:
… does not exist

## Related Issue
#29700 

## How Has This Been Tested?
Tried making some steps for files that actually do not exist, ran the scenarios and confirmed that the step now fails if the file does not exist.

CI will run through all the existing integration tests and maybe find some interesting results.

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/WebDav.php
+++ b/tests/integration/features/bootstrap/WebDav.php
@@ -354,6 +354,9 @@ trait WebDav {
 	 */
 	public function asTheFileOrFolderExists($user, $entry, $path) {
 		$this->response = $this->listFolder($user, $path, 0);
+		if (!is_array($this->response) || !isset($this->response['{DAV:}getetag'])) {
+			throw new \Exception($entry . ' "' . $path . '" expected to exist but not found');
+		}
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


phil-davis:
@DeepDiver1975 @PVince81 @SergioBertolinSG @davitol @individual-it 
I discovered just now that a step like:
```
Then as "alice" the file "/file.txt" exists
```
always passes, whether the file exists or not.

This is my first try at fixing it (the check I have put here does make it fail if the file does not exist). But maybe there is a cleaner way to decide if the ``listFolder()`` response is good or not?

individual-it (dismissed):
we need tests to test the tests

phil-davis:
Yes, over time somehow each step needs to have it confirmed that if something is wrong, the step will actually fail. Maybe I should write a framework for automatically testing test frameworks?

individual-it:
image

phil-davis:
The 2nd commit fixes 2 tests that failed after the code was fixed in the 1st commit.
For both of those, there was a typo in the expected file name. But even then, actually the previous step did "upload with all mechanisms" which creates a set of uploads that add old/new dav/chunking suffixes to the file names.
So I made a new step definition that checks that each of the files with the generated suffixes exists.

individual-it (approved):

phil-davis:
Backport stable #29719

PR #29702, opened 2017-11-28
Add test for modifying a group shared file

phil-davis:
## Description
Share a single file with a group,
make a group member modify the file,
check if the group owner sees the modified content

## Motivation and Context
This set of steps fails for me. I was using it to set up a scenario in another app ready for further testing of the app.
So I need to see if it really fails in core CI.
The steps work if done manually. So maybe there is some problem with one of the automated step definitions?

There seems to be no similar-enough scenario in the existing core integration tests, so I can't see straight away if there is a real application bug or an integration test implementation bug.

## How Has This Been Tested?
Fails locally - will it fail in CI?

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/sharing-v1.feature
+++ b/tests/integration/features/sharing-v1.feature
@@ -453,6 +453,20 @@ Feature: sharing
 		And the HTTP status code should be "200"
 		And last share_id is included in the answer
 
+	Scenario: Sharee can modify the group shared file
+		Given user "user0" exists
+		And user "user1" exists
+		And group "group0" exists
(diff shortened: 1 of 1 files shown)


phil-davis:
Now I realise the problem is with the test setup - ``textfile0.txt`` is part of the skeleton, so both ``user0`` and ``user1`` already have independent copies of this file. When ``user0`` shares it, ``user1`` sees the shared folder as ``textfile0 (2).txt``

All is is well - it just gets a bit tricky to be aware of the whole context of the environment of the test users.

Issue #29705, opened 2017-11-28
Undefined index: size at core/lib/private/Files/View.php#1287


phil-davis:
### Steps to reproduce
1. Run integration tests, e.g.
```
./run.sh federation_features/federated.feature:225
```
2. Notice the "Undefined index" messages like:
```
[Wed Nov 29 03:52:52 2017] Undefined index: size at /home/phil/core/lib/private/Files/View.php#1287
[Wed Nov 29 03:52:52 2017] 127.0.0.1:41130 [201]: /remote.php/webdav/PARENT%20(2)/testquota.txt-olddav-regular
[Wed Nov 29 03:52:53 2017] 127.0.0.1:41154 [201]: /remote.php/webdav/PARENT%20(2)/testquota.txt-olddav-oldchunking-chunking-42-2-0
[Wed Nov 29 03:52:55 2017] 127.0.0.1:41174 [201]: /remote.php/webdav/PARENT%20(2)/testquota.txt-olddav-oldchunking-chunking-42-2-1
[Wed Nov 29 03:52:56 2017] Undefined index: size at /home/phil/core/lib/private/Files/View.php#1287
[Wed Nov 29 03:52:56 2017] 127.0.0.1:41202 [201]: /remote.php/dav/files/user1/PARENT%20(2)/testquota.txt-newdav-regular
[Wed Nov 29 03:52:56 2017] 127.0.0.1:41230 [201]: /remote.php/dav/uploads/user1/chunking-43
[Wed Nov 29 03:52:56 2017] 127.0.0.1:41232 [201]: /remote.php/dav/uploads/user1/chunking-43/1
[Wed Nov 29 03:52:57 2017] 127.0.0.1:41234 [201]: /remote.php/dav/uploads/user1/chunking-43/2
[Wed Nov 29 03:52:58 2017] Undefined index: size at /home/phil/core/lib/private/Files/View.php#1287
```
or look in a Jenkins log for similar.

### Expected behaviour
No "Undefined index" messages.

### Actual behaviour
"Undefined index" messages.

### Server configuration
Current 10.0.4 ccode from git running in local dev environment or in CI.

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributor most likely able to help you is @PVince81.

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

[...]

PR #29706, opened 2017-11-28
View.php - Trigger cache if size is not set

phil-davis:
## Description
If the size is not set, then it is like size -1 - so trigger a cache (re)load.

## Related Issue
#29705 

## Motivation and Context
Get rid of PHP warnings.

## How Has This Been Tested?
Run an effected integration test case.
CI will tell about the "whole system".

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -1284,7 +1284,7 @@ class View {
 
 		try {
 			// if the file is not in the cache or needs to be updated, trigger the scanner and reload the data
-			if (!$data || $data['size'] === -1) {
+			if (!$data || (isset($data['size']) && ($data['size'] === -1))) {
 				$this->lockFile($relativePath, ILockingProvider::LOCK_SHARED);
 				if (!$storage->file_exists($internalPath)) {
 					$this->unlockFile($relativePath, ILockingProvider::LOCK_SHARED);


phil-davis:
@PVince81 @DeepDiver1975 will this fix (or break) some file cache thing for real users?
I only touched this because I don't like seeing "Undefined index" in logs.

lib/private/Files/View.php
```diff
@@ -1284,7 +1284,7 @@ private function getCacheEntry($storage, $internalPath, $relativePath) {
 
 		try {
 			// if the file is not in the cache or needs to be updated, trigger the scanner and reload the data
-			if (!$data || $data['size'] === -1) {
+			if (!$data || !isset($data['size']) || $data['size'] === -1) {
```
PVince81: Shouldn't this rather be `!$data || (isset($data['size']) && $data['size'] === -1)` ?

Does PHP really go into the if block when `$data['size']` is not defined ? I don't think so because the condition `$data['size'] === -1` would be false if we consider that missing size is like if we'd say `$data['size']` was null, which is different than-1.

So yes, there is a risk of breakage with your modification.

phil-davis: @PVince81 with your suggestion, then if the 'size' key is not set at all, the code in the block will not be executed. 

``$data = $cache->get($internalPath)`` can return a value in ``$data`` that evaluates ``true`` (not matching the 1st test) but does not have the ``size`` key set. The question is, in that case do we want to execute the code in the block - i.e. try to load the cache?

Because if ``$data = $cache->get($internalPath)`` has returned ``false`` then we already try to load the cache.

So this other case is "in between" - not ``false`` but also not providing a ``size``

PVince81: > do we want to execute the code in the block

this is the decision I'm not ready to make right now... would require thinking about potential breakages.

Your purpose for this PR is only to make the warning disappear, so just build it in a way that the code path is the same as before, which I believe is about not executing the block inside the `if`.

phil-davis:
@PVince81 this should be safer now

PVince81 (approved):
:+1:

PVince81:
please backport for 10.0.5

phil-davis:
Backport stable10 #29743

Issue #29698, opened 2017-11-28
a group called "0" cannot be created by provisioning API


individual-it:
### Steps to reproduce
1. try to create a group called "0" with the API e.g. `curl http://admin:admin@$HOST/ocs/v2.php/cloud/groups -d "groupid=0"`

### Expected behaviour
group should be created
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>200</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

### Actual behaviour
cannot create the group
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>400</statuscode>
  <message>Invalid group name</message>
 </meta>
 <data/>
</ocs>
```

### Server configuration
**Operating system**:
Ubuntu 17.10
**Web server:**
Apache
**Database:**
SQLite
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
10.0.4 RC1 (git)
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc9xsmdtnlru",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "localhost"
        ],
        "datadirectory": "\/home\/artur\/www\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/localhost\/owncloud-core",
        "dbtype": "sqlite3",
        "logtimezone": "UTC",
        "installed": true,
        "default_language": "en",
        "theme": "",
        "loglevel": 0,
        "maintenance": false,
        "log_type": "owncloud",
        "dbname": "owncloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "debug": false,
        "skeletondirectory": "\/home\/artur\/www\/owncloud-core\/tests\/ui\/skeleton",
        "version": "10.0.4.2",
        "csrf.disabled": true,
        "files_external_allow_create_new_local": "true",
        "singleuser": false,
        "license-key": "***REMOVED SENSITIVE VALUE***"
    }
}
```

**List of activated apps:**

```
Enabled:
[...]

PR #29712, opened 2017-11-29
Allow group 0 to be created by provisioning API

phil-davis:
## Description
Only the empty string should be rejected as a group name (or things that ``==`` the empty string such as ``null`` and ``false``). This lets "0" be created as a group name.

## Related Issue
#29698 

## Motivation and Context
Zero and empty must not live in the same house.

## How Has This Been Tested?
The new integration test fails before and passes after this fix.

## Types of changes

## Checklist:

### DIFF
--- a/apps/provisioning_api/lib/Groups.php
+++ b/apps/provisioning_api/lib/Groups.php
@@ -130,7 +130,7 @@ class Groups{
 	public function addGroup($parameters) {
 		// Validate name
 		$groupId = $this->request->getParam('groupid', '');
-		if(empty($groupId)){
+		if(($groupId === '') || is_null($groupId) || ($groupId === false)){
 			\OCP\Util::writeLog('provisioning_api', 'Group name not supplied', \OCP\Util::ERROR);
 			return new OC_OCS_Result(null, 101, 'Invalid group name');
 		}
--- a/tests/integration/features/provisioning-v1.feature
+++ b/tests/integration/features/provisioning-v1.feature
@@ -80,6 +80,16 @@ Feature: provisioning
 		And the HTTP status code should be "200"
 		And group "España" exists
 
+	Scenario: Create a group named "0"
+		Given as an "admin"
+		And group "0" does not exist
+		When sending "POST" to "/cloud/groups" with
+			| groupid | 0 |
+			| password | 123456 |
+		Then the OCS status code should be "100"
+		And the HTTP status code should be "200"
+		And group "0" exists
+
 	Scenario: adding user to a group without sending the group
 		Given as an "admin"
 		And user "brand-new-user" exists
@@ -140,13 +150,16 @@ Feature: provisioning
 
 	Scenario: Getting all groups
 		Given as an "admin"
+		And group "0" exists
 		And group "new-group" exists
 		And group "admin" exists
+		And group "España" exists
 		When sending "GET" to "/cloud/groups"
 		Then groups returned are
 			| España |
 			| admin |
 			| new-group |
+			| 0 |
 
 	Scenario: create a subadmin
 		Given as an "admin"


individual-it (dismissed):

apps/provisioning_api/lib/Groups.php
```diff
@@ -130,7 +130,7 @@ public function getGroup($parameters) {
 	public function addGroup($parameters) {
 		// Validate name
 		$groupId = $this->request->getParam('groupid', '');
-		if(empty($groupId)){
+		if($groupId == ''){
```
PVince81: classic

PVince81: can you make that `===` ? because `==` might also match null values. If we also want to check for nulll values, also add `is_null($groupId)`

individual-it:
Why is Jenkins failing? Do the Scenarios depend on each other?

phil-davis:
Yes, it does not look good - groups are not deleted after each scenario so they "add up". It means that some scenarios cannot be run just b y themselves. So I will need to do some work to improve that.

apps/provisioning_api/lib/Groups.php
phil-davis: I wanted to match all of:
```
if ($group === "")
if (is_null($group))
if ($group === false)
``
and those 3 things all are done in 1 statement by:
```
if ($group == "")
```

Comment (phil-davis) [reply apps/provisioning_api/lib/Groups.php]: In theory the return value of ``getParam`` is always a string - it "should" never be null or false, but who knows.

File: tests/integration/features/provisioning-v1.feature [hunk 2]
```diff
@@ -140,13 +150,16 @@ Feature: provisioning
 
 	Scenario: Getting all groups
 		Given as an "admin"
+		And group "0" exists
```
phil-davis: I "fixed up" this test so that at least it explicitly specifies each group to exists then checks that they are returned by the ``/cloud/groups`` endpoint.
Actually each group (except ``admin``) is created in one of the previous scenarios by code that does:
```
When sending "POST" to "/cloud/groups" with
```
That step does not "remember" that it created a group, because that is a generic step for "driving the API". So in those kind of scenarios, the created group is not deleted by the ``AfterScenario`` code.

Thus this scenario would actually pass without specifically needing to mention that the 4 groups "exist".

Creating users has the same problem - various scenarios create users that are not deleted in the ``AfterScenario`` code. But all that sort of thing needs to be looked at separately, otherwise this "little" PR will become "bigger than Ben Hur" if it starts rectifying existing "features" of the test infrastructure.

apps/provisioning_api/lib/Groups.php
individual-it: you could check if `$group`  is a string and if not also show the same error. Or even throw an exception because if the the group name is not a string something went terrible wrong.

phil-davis: Way too much time spent on a 1-line fix. Somebody tell me what code will be accepted for review and then I can write that. Otherwise I will just be guessing what is wanted and going round-and-round the long CI process.

apps/provisioning_api/lib/Groups.php
```diff
@@ -130,7 +130,7 @@ public function getGroup($parameters) {
 	public function addGroup($parameters) {
 		// Validate name
 		$groupId = $this->request->getParam('groupid', '');
-		if(empty($groupId)){
+		if(($groupId === '') || is_null($groupId) || ($groupId === false)){
```
phil-davis: @PVince81 I went for the full list here - ``empty()`` essentially means "blank or null or false or 0 or '0'". We want the 0 to be allowed and the others can continue to be banned as a group name.

butonic (approved):

phil-davis:
Backport stable10 #29734

PR #29713, opened 2017-11-29
[stable10] Allow group 0 to be created by provisioning API

phil-davis:
Backport #29712


phil-davis:
This was an early pro-active backport of code that turned out not to pass CI - closing.

Issue #29726, opened 2017-11-30
encryption:recreate-master-key does not encrypt a file if we have run before encryption:decrypt-all command


davitol:
### Steps

1. As Admin Setup encryption with master key on the ownCloud Server via webUI
2. Logout and Log in as admin to initialize keys
3. Create a file.txt and save it
4. Check the file is encrypted
5. run ```sudo -u www-data ./occ encryption:decrypt-all```
6. Check that file.txt is decrypted
7. Create a file2.txt and save it
8. Check the file is encrypted
9. run ```sudo -u www-data ./occ encryption:recreate-master-key```

### Expected result
file2.txt should be kept encrypted

### Actual result
file2.txt is no longer encrypted

### Version
ownCloud 10.0.4RC1 for both instances

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributor most likely able to help you is @PVince81.

sharidas:
I am able to reproduce it. I am working on the same.

sharidas:
The changes are merged for this issue. Hence closing this issue.

PR #29731, opened 2017-11-30
Fix recreate masterkey issue when decryptall command used

sharidas:
When decrypt-all command is used the files get decrypted.
After this when user executes the recreate masterkey
command, the variable disableWriteEncryption is set to
true and never set to false. Though the decrypted files
when tried decrypt again in the copy operation, throws
exception, the value should be set to false. If not
set to false, user files would never be re-encrypted
with recreate masterkey.

## Description

This tries to solve the problem caused when user tries to recreate masterkey after executing decrypt-all command.

## Related Issue

https://github.com/owncloud/core/issues/29726

## Motivation and Context

When user executes decrypt-all command the oC files are decrypted successfully. Now after this step if user tries to execute recreate masterkey command, then the user could see that decrypting already decrypted files throws exception and code flow reaches: https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L795. As a result the value of the variable `disableWriteEncryption` still remains false. This causes problem with fopen during re-encryption phase. So this change helps to reset the value to false.

## How Has This Been Tested?

1. Create user admin and login
2. Run decrypt-all command from the command line
3. Run recreate masterkey command from the command line the files should remain encrypted.
This has also been tested in the same way mentioned in the issue : https://github.com/owncloud/core/issues/29726

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -793,6 +793,7 @@ class Encryption extends Wrapper {
 				fclose($source);
 				fclose($target);
 			} catch (\Exception $e) {
+				Encryption::setDisableWriteEncryption(false);
 				fclose($source);
 				fclose($target);
 				throw $e;


PVince81:
Why would it throw an exception ?

sharidas:
> Why would it throw an exception in something that looks like a normal operation ?

The reason for exception is due to the fact that its trying to decrypt the file, which was already decrypted. The decrypt-all command decrypts all files in the oC and then re-enables the encryption modules. So far that's good. Now the problem comes when recreate masterkey is called. The recreate master key internally calls decrypt-all methods. To verify this error one more scenario would be to run decrypt-all again. Even I was expecting the case to be normal and why should it throw exception. The reason comes due to decrypting the decrypted file : https://github.com/owncloud/core/blob/master/lib/private/Encryption/DecryptAll.php#L278 this copy calls the fopen in read mode for the source file. And eventually during this call reaches fopen of Encryption wrapper https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L407 which eventually leads the code flow to https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L166. Which leads to call https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L518 and eventually https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L599. Since the file is already decrypted, the call to https://github.com/owncloud/encryption/blob/master/lib/Crypto/Crypt.php#L578 causes the exception. Hope this justification looks ok.

When this exception happens, the other part is the variable in the patch is still `true`. And this was causing the mess.

sharidas:
Not sure if it would be a good idea to introduce a method get value of disableWriteEncryption, to please the tests!! Any suggestions?

PVince81 (approved):
:+1: code is fine then for now, please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/29735

davitol:
@sharidas WFM, solves https://github.com/owncloud/core/issues/29726

I have also seen that after running ```sudo -u www-data ./occ encryption:recreate-master-key```

an error appears in the webUI if you try to create a new file:

screen shot 2017-12-01 at 09 53 17


It gets solved when you logout and login again

PR #29735, opened 2017-12-01
[stable10] Fix recreate masterkey issue when decryptall command used

sharidas:
When decrypt-all command is used the files get decrypted.
After this when user executes the recreate masterkey
command, the variable disableWriteEncryption is set to
true and never set to false. Though the decrypted files
when tried decrypt again in the copy operation, throws
exception, the value should be set to false. If not
set to false, user files would never be re-encrypted
with recreate masterkey.

## Description

This tries to solve the problem caused when user tries to recreate masterkey after executing decrypt-all command.

## Related Issue

https://github.com/owncloud/core/issues/29726

## Motivation and Context

When user executes decrypt-all command the oC files are decrypted successfully. Now after this step if user tries to execute recreate masterkey command, then the user could see that decrypting already decrypted files throws exception and code flow reaches: https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Encryption.php#L795. As a result the value of the variable disableWriteEncryption still remains false. This causes problem with fopen during re-encryption phase. So this change helps to reset the value to false.

## How Has This Been Tested?

1. Create user admin and login
2. Run decrypt-all command from the command line
3. Run recreate masterkey command from the command line the files should remain encrypted.
This has also been tested in the same way mentioned in the issue : #29726

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -793,6 +793,7 @@ class Encryption extends Wrapper {
 				fclose($source);
 				fclose($target);
 			} catch (\Exception $e) {
+				Encryption::setDisableWriteEncryption(false);
 				fclose($source);
 				fclose($target);
 				throw $e;


sharidas:
Orignal PR: https://github.com/owncloud/core/pull/29731

PVince81 (approved):
:+1:

PR #29738, opened 2017-12-01
Handle case of first CORS whitelist entry

phil-davis:
When adding the first CORS whitelist entry it failed with:
```
array_filter() expects parameter 1 to be array, null given at \/home\/phil\/core\/settings\/Controller\/CorsController.php#116
```
after adding the code here I can add entries.

### DIFF
--- a/settings/Controller/CorsController.php
+++ b/settings/Controller/CorsController.php
@@ -112,14 +112,14 @@ class CorsController extends Controller {
 		}
 
 		$userId = $this->userId;
-		$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'));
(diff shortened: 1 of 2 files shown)


settings/Controller/CorsController.php
```diff
@@ -113,7 +113,11 @@ public function addDomain($domain) {
 
 		$userId = $this->userId;
 		$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'));
-		$domains = array_filter($domains);
+		if (is_array($domains)) {
```
PVince81: does json_decode output any warning in the log ?

I suggest to fix it by simply adding `"[]"` as default value for `getUserValue()` (extra arg)

phil-davis:
I will make a new PR based on master

Issue #28181, opened 2017-06-21
External file scanning error


adrienafl:
Hi ! Just having this error : I tried to fix perms with the strong perm script. I also tried to use update perms but it didn't change anything.

Thank you

### Steps to reproduce
1. Run sudo -u www-data php occ files:scan --all

### Expected behaviour
Scan should complete

### Actual behaviour
```
Exception during scan: Undefined index: size
#0 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(416): OCA\Files\Command\Scan->exceptionErrorHandler(8, 'Undefined index...', '/var/www/ownclo...', 416, Array)
#1 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(377): OC\Files\Cache\Scanner->handleChildren('files/Backup/PI...', true, 3, '57663', true, 119080535)
#2 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(380): OC\Files\Cache\Scanner->scanChildren('files/Backup/PI...', true, 3, '57663', true)
#3 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(380): OC\Files\Cache\Scanner->scanChildren('files/Backup/PI...', true, 3, '512', true)
#4 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(380): OC\Files\Cache\Scanner->scanChildren('files/Backup/PI...', true, 3, '508', true)
#5 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(380): OC\Files\Cache\Scanner->scanChildren('files/Backup', true, 3, '169', true)
#6 /var/www/owncloud/lib/private/Files/Cache/Scanner.php(311): OC\Files\Cache\Scanner->scanChildren('files', true, 3, 2, true)
#7 /var/www/owncloud/lib/private/Files/Utils/Scanner.php(235): OC\Files\Cache\Scanner->scan('files', true, 3)
#8 /var/www/owncloud/apps/files/lib/Command/Scan.php(159): OC\Files\Utils\Scanner->scan('/xxx/files')
#9 /var/www/owncloud/apps/files/lib/Command/Scan.php(228): OCA\Files\Command\Scan->scanFiles('xxx', '/xxx/files', false, Object(Symfony\Component\Console\Output\ConsoleOutput), false)
[...]

PR #29257, opened 2017-10-16
Use dav versions in files

DeepDiver1975:
## Description
This PR is based on #29207 and integrated files_versions ui via webdav

## Related Issue
#29087 

## Motivation and Context
We shall use the webdav api for versioning

## How Has This Been Tested?
- manual testing in files app
- cadaver

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29739, opened 2017-12-01
Keep null in getMetaData in Checksum storage wrapper

PVince81:
## Description
Keep null in getMetaData in Checksum storage wrapper.

Not funny is that in PHP if you add an item into `null` as if it was an array, it suddenly becomes one...

related #28181 
fixes #28960 

## Related Issue
None raised. Discovered while debugging background scanner with size=-1 and I noticed that getMetaData() on a non-existing file returns not null but a partial array.

## How Has This Been Tested?
Not tested yet.
Why is there no unit test class for the Checksum storage wrapper ?!! @IljaN 

## Types of changes

## Checklist:

## TODO

- [x] write unit test (well, write the whole class as it's missing)
- [x] think about what was broken before this fix
- [x] think about what could break after this fix

### DIFF
--- a/apps/files_versions/tests/VersioningTest.php
+++ b/apps/files_versions/tests/VersioningTest.php
@@ -3,6 +3,7 @@
  * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
  * @author Björn Schießle <bjoern@schiessle.org>
  * @author Georg Ehrke <georg@owncloud.com>
+ * @author Ilja Neumann <ineumann@owncloud.com>
  * @author Joas Schilling <coding@schilljs.com>
  * @author Jörn Friedrich Dreyer <jfd@butonic.de>
  * @author Lukas Reschke <lukas@statuscode.ch>
@@ -47,9 +48,12 @@ use Test\TestCase;
  */
 class VersioningTest extends TestCase {
 
-	const TEST_VERSIONS_USER = 'test-versions-user';
-	const TEST_VERSIONS_USER2 = 'test-versions-user2';
-	const USERS_VERSIONS_ROOT = '/test-versions-user/files_versions';
+	/** @var string */
+	private $user1;
+	/** @var string */
+	private $user2;
+	/** @var string */
+	private $versionsRootOfUser1;
 
 	/**
 	 * @var \OC\Files\View
@@ -62,19 +66,6 @@ class VersioningTest extends TestCase {
 		$application = new \OCA\Files_Sharing\AppInfo\Application();
 		$application->registerMountProviders();
 
-		// create test user
-		self::loginHelper(self::TEST_VERSIONS_USER2, true);
-		self::loginHelper(self::TEST_VERSIONS_USER, true);
-	}
-
-	public static function tearDownAfterClass() {
-		// cleanup test user
-		$user = \OC::$server->getUserManager()->get(self::TEST_VERSIONS_USER);
-		if ($user !== null) { $user->delete(); }
-		$user = \OC::$server->getUserManager()->get(self::TEST_VERSIONS_USER2);
-		if ($user !== null) { $user->delete(); }
-
-		parent::tearDownAfterClass();
 	}
 
 	protected function setUp() {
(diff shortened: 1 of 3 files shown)


mrow4a:
Sounds like missing unit test somewhere :>

PVince81:
@IljaN can you take over and add the missing tests ? Somehow it seems we missed checking for tests back in the checksum PR.

There are already some example for storage wrapper unit tests.

IljaN:
Checksums are covered by acceptance tests. But a UnitTest won`t hurt :)

mmattel:
For the record, tested and fixes the referenced issue https://github.com/owncloud/core/issues/29806#issuecomment-351013282

Seems that tests need a restart so we can go for a green and merge...

IljaN:
@mmattel Thanks! I am in the process of adding some unit-tests. So no manual restart needed.

IljaN:
We are running in to this exception in a versioning test with null:
https://github.com/owncloud/core/blob/639986ac4405ffb4beeb7c1d0ece901e928fcfe5/lib/private/Files/Cache/Scanner.php#L117-L120

```
1) OCA\Files_Versions\Tests\VersioningTest::testRenameInSharedFolder
OCP\Files\StorageNotAvailableException: Missing important folder "files" in home storage - home::test-versions-user2

/home/ilja/code/core2/lib/private/Files/Cache/Scanner.php:119
/home/ilja/code/core2/lib/private/Files/Cache/Scanner.php:150
/home/ilja/code/core2/lib/private/Files/Cache/Watcher.php:106
/home/ilja/code/core2/lib/private/Files/View.php:1299
/home/ilja/code/core2/lib/private/Files/View.php:1335
/home/ilja/code/core2/lib/private/Files/Node/Root.php:188
/home/ilja/code/core2/lib/private/Files/Node/Folder.php:129
/home/ilja/code/core2/lib/private/Files/Node/Root.php:367
/home/ilja/code/core2/lib/private/Server.php:970
/home/ilja/code/core2/apps/files_versions/tests/VersioningTest.php:725
/home/ilja/code/core2/apps/files_versions/tests/VersioningTest.php:330
```

Not sure how to proceed atm...

PVince81:
@IljaN if you're testing on actual FS, make sure that the "data/$userId/files" folder of the user exists as if the user had logged in before. Maybe some of previous running tests delete this and this test here expected it to exist without proper initialization. (happened before...)

IljaN:
Yeah this test is actually quite brittle, runs just once locally and then throws user already exists error, will try  to fix it

IljaN:
@PVince81 you mentioned that there are some filesystem conditions which can cause null meta-data, who is in the know?
If null can happen we either must refactor so that getMetaData never returns null and all codepaths (Scanner?) can work with an emtpy array.

There are ~16 Usages of get MetaData

Or we need to change the docs for the storage interface to allow null:
https://github.com/owncloud/core/blob/f10d105ea2f2d0822693313ddcd6f9b2eaac6ebe/lib/private/Files/Storage/Storage.php#L92-L95

Should we create a follow up issue?

PVince81:
@IljaN the only way I found currently to reproduce this issue was having a NFD-formatted file name on an external storage and running `occ files:scan`. Not sure about code paths, maybe something to look into separately, considering that this PR here already fixes this issue.

IljaN:
@PVince81 Not sure If I was clear enough: return null here fixes the issue:
https://github.com/owncloud/core/blob/fdd04d61ad4eb4967b30f7c9162dbe15729ad1a9/lib/private/Files/Storage/Wrapper/Checksum.php#L193
...but violates storage interface (should return array).

PVince81:
Hmm... but it seems the underlying storage does return `null`, or is it returning an empty array ? Maybe just return whatever the underlying storage gave you...

IljaN:
Yeah there are actually a few other sites which handle null from getMetaData calls, but I consider this a bug/incorrect implementation because the inteface states that it should return array. (Or we change the doc)

IljaN:
Example from above:
https://github.com/owncloud/core/blob/639986ac4405ffb4beeb7c1d0ece901e928fcfe5/lib/private/Files/Cache/Scanner.php#L117-L120

IljaN:
Maybe some deeper layer returns null and this is propagated up so I would assume that the deeper layer should return an array, or the first wrapper which encounters null should pass an empty array up to the next wrapper.

PVince81:
I suggest we look into this separately... maybe adjust the interfaces for null then. I think `null` is a valid value for not found metadata. Even better of course would be exceptions, but that would likely break a lot of existing code. We'll likely rework all this mess as part of storage API rework.

PVince81:
:+1: for the unit tests.

PVince81:
backport conflict with versions stuff, might be better to wait for backport of https://github.com/owncloud/core/pull/29257/files#diff-5ad1bbf2fd6f06e61ebecc59cd61a983L130

cc @DeepDiver1975

PVince81:
@DeepDiver1975 ping on this blocker, we can't backport without this.

Or well we could backport only half of this and the rest later. Let me know.

DeepDiver1975:
Lets chat in about two days again - I should have an understanding of the situation by then ;-)

PVince81:
I made a backport of only the fix without the unit tests as the latter are blocked (see previous comments).

PR here https://github.com/owncloud/core/pull/30302

PVince81:
- [ ] backport unit tests when ready

PVince81:
Now https://github.com/owncloud/core/pull/30302 is a full backport of this + adjustments to make it work with stable10

PR #29742, opened 2017-12-01
Handle case of first CORS whitelist add-delete

phil-davis:
## Description
Handle a default array the first time when there are no CORS whitelist entries.
When removing the last entry, delete the value completely.

## Motivation and Context
Be able to add and delete CORS whitelist entries.

## How Has This Been Tested?
Add and delete various entries manually in the webUI, refreshing to confirm the changes.

## Types of changes

## Checklist:

### DIFF
--- a/settings/Controller/CorsController.php
+++ b/settings/Controller/CorsController.php
@@ -90,13 +90,7 @@ class CorsController extends Controller {
 	 */
 	public function getDomains() {
 		$userId = $this->userId;
-
-		if (empty($this->config->getUserValue($userId, 'core', 'domains'))) {
-			$domains = [];
-		} else {
-			$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'));
-		}
-
+		$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains', '[]'), true);
 		return new JSONResponse($domains);
 	}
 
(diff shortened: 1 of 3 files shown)


settings/Controller/CorsController.php
```diff
@@ -136,11 +136,14 @@ public function addDomain($domain) {
 	 */
 	public function removeDomain($id) {
 		$userId = $this->userId;
-		$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'));
-
+		$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'), true);
```
PVince81: why not also add `"[]"` here for consistency since you did above ?

phil-davis: True, in theory ``removeDomain`` is only called if there is something to remove. But I guess it should be resilient and silent remove something from nothing if asked.

settings/Controller/CorsController.php
```diff
@@ -136,11 +136,14 @@ public function addDomain($domain) {
...
 		if ($id >= 0 && $id < count($domains)) {
 			unset($domains[$id]);
-			$this->config->setUserValue($userId, 'core', 'domains', json_encode($domains));
+			if (count($domains)) {
```
PVince81: `empty($domains)` ?

settings/Controller/CorsController.php
```diff
@@ -136,11 +136,12 @@ public function addDomain($domain) {
 	 */
 	public function removeDomain($id) {
 		$userId = $this->userId;
-		$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'));
-
-		if ($id >= 0 && $id < count($domains)) {
```
phil-davis: I just pushed this change. The $id is 1,2,3,4... and when you delete 2 there is 1,3,4 remaining. There were two problems here:
a) You could never delete the last entry because its id equaled count(domains)
b) If you deleted stuff early in the list, then you could no longer delete later entries

Actually there is no need for the ``if`` test - just unset whatever id was asked to be deleted and then write back (or delete) the value.

phil-davis:
@ anybody who cares - I think this swill pass unit tests and also actually works now. CI will tell. Ready for review again.

PVince81:
@phil-davis please backport for 10.0.4

PVince81:
stable10: https://github.com/owncloud/core/pull/29749

Issue #28167, opened 2017-06-20
theming l10n translation


felixboehm:
tried to customize a translation via theme

### Steps to reproduce
1. mkdir -p themes/myTheme/core/l10n
2. cp core/l10n/en_GB.js* themes/myTheme/core/l10n/
3. vim themes/myTheme/core/l10n/en_GB.js # change some 
4. configure theme in config.php
5. load page and filter network for en_GB files
6. switch to the correct language in personal settings

### Expected behaviour
Custom strings are used.
The en_GB.js file for "core" should come from theme

### Actual behaviour
Custom strings are not used

Two en_GB.js files for "core" are loaded
- http://localhost/core/l10n/en_GB.js?v=9ab0ff1745e8dcd07cefb190b7891a57
- http://localhost/themes/myth//core/l10n/en_GB.js?v=9ab0ff1745e8dcd07cefb190b7891a57

Note double slash in second

### Other
ownCloud 10.0.3
No log entries

felixboehm:
Please recheck and consider writing acceptance tests for this. @SergioBertolinSG 
If you find the same behavior, let's fix this.

felixboehm:
theming apps doesn't work anymore either??

DeepDiver1975:
> theming apps doesn't work anymore either??

cannot reproduce.

just enabled a theme and it loads just fine.

will have a look at translations ....

DeepDiver1975:
> will have a look at translations ....

the new app based theming is not respected in loading themed translations .... needs a deeper look.

furthermore translations are loaded even before the theme app is loaded

DeepDiver1975:
on loading of the js translation files theme path is completly ignores ....

DeepDiver1975:
according to @phisch fixed with #27746

felixboehm:
thanks

PVince81:
and https://github.com/owncloud/core/pull/27746 is still open

DeepDiver1975:
> according to @phisch fixed with #27746

just tested this and this is not fixed ...

PVince81:
p3, moving to "planned" for now

glannes31:
Hello, any idea for the fix date ?

tomneedham:
Breaks branding which is important for big installations upgrading to 10

VicDeo:
js part works fine with l10n from themes.
the bug in php translation only.

VicDeo:
[...]

PR #29747, opened 2017-12-01
Single active app theme with no magic fallbacks to inactive app themes

VicDeo:
## Description
Fixes custom translations inside app-themes.

TODO: 
- [x] Remove undocumented fallback to the app-theme with `default_enabled` attribute
- [x] Fallback to the legacy theme only in case there is no app-theme active
- [x] Allow only one app-theme to be active (CLI/Web)

 A migration step that
- [x]  resets legacy theme if app-theme is enabled `->setSystemValue('theme', '')`
- [x]  disables extra app-themes in case there is more than one active app-theme
currently the last (alphabetically) app-theme wins. So the latest theme should stay active.

## Related Issue
Fixes 
https://github.com/owncloud/core/issues/28167 
https://github.com/owncloud/theme-enterprise/issues/16

### Test 1
1. Activate app-theme. e.g. theme-example
2. Switch to British English
3. `mkdir -p apps/theme-example/core/l10n`
4. `cp core/l10n/en_GB.json apps/theme-example/core/l10n/en_GB.json`
5. Change 
`    "Log out" : "Log out",`
to
`    "Log out" : "Go away!",`

### Expected 
Log out is renamed in menu

### Actual 
it is not

#### Tested as a guest 
Please note that guest user uses a language based on browser locale settings. 
You'll need to override another file (es.json, de.json, whatever...) instead of en_GB.json if your browser is not in en_GB
- [x]     "File not found" changed to "File not found and never will be"
Accessing an unexisting share - translation works
- [x]     "Password" changed to "Password 123"
Accessing a login page - translation works

### Test 2
1. Activate another app-theme

### Expected 
It can't be enabled until  the current theme is active

### Actual
Both themes are enabled, theme that is the last according to the alphabet is loaded.

## Types of changes

## Checklist:

### DIFF
--- a/core/Command/App/Enable.php
+++ b/core/Command/App/Enable.php
@@ -70,14 +70,20 @@ class Enable extends Command {
 			return 1;
 		}
 
-		$groups = $input->getOption('groups');
-		if (empty($groups)) {
-			\OC_App::enable($appId);
-			$output->writeln($appId . ' enabled');
-		} else {
-			\OC_App::enable($appId, $groups);
-			$output->writeln($appId . ' enabled for groups: ' . implode(', ', $groups));
+		try {
+			$groups = $input->getOption('groups');
+			if (empty($groups)) {
+				\OC_App::enable($appId);
+				$output->writeln($appId . ' enabled');
+			} else {
+				\OC_App::enable($appId, $groups);
+				$output->writeln($appId . ' enabled for groups: ' . implode(', ', $groups));
+			}
+		} catch (\Exception $e) {
+			$output->writeln($e->getMessage());
+			return 2;
 		}
+
 		return 0;
 	}
 }
--- a/core/Command/App/Enable.php
+++ b/core/Command/App/Enable.php
@@ -25,6 +25,7 @@
 namespace OC\Core\Command\App;
 
 use OCP\App\IAppManager;
+use OCP\ILogger;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
@@ -70,18 +71,13 @@ class Enable extends Command {
 			return 1;
 		}
 
-		try {
-			$groups = $input->getOption('groups');
-			if (empty($groups)) {
-				\OC_App::enable($appId);
-				$output->writeln($appId . ' enabled');
-			} else {
-				\OC_App::enable($appId, $groups);
-				$output->writeln($appId . ' enabled for groups: ' . implode(', ', $groups));
-			}
-		} catch (\Exception $e) {
-			$output->writeln($e->getMessage());
-			return 2;
+		$groups = $input->getOption('groups');
+		if (empty($groups)) {
+			\OC_App::enable($appId);
+			$output->writeln($appId . ' enabled');
+		} else {
+			\OC_App::enable($appId, $groups);
+			$output->writeln($appId . ' enabled for groups: ' . implode(', ', $groups));
 		}
 
 		return 0;
--- a/index.php
+++ b/index.php
@@ -52,30 +52,24 @@ if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
 try {
 
 	require_once __DIR__ . '/lib/base.php';
-
 	OC::handleRequest();
 
 } catch(\OC\ServiceUnavailableException $ex) {
-	\OC::loadDefaultEnabledAppTheme();
 	\OC::$server->getLogger()->logException($ex, array('app' => 'index'));
 
 	//show the user a detailed error page
 	OC_Response::setStatus(OC_Response::STATUS_SERVICE_UNAVAILABLE);
 	OC_Template::printExceptionErrorPage($ex);
 } catch (\OC\HintException $ex) {
-	\OC::loadDefaultEnabledAppTheme();
 	OC_Response::setStatus(OC_Response::STATUS_SERVICE_UNAVAILABLE);
 	OC_Template::printErrorPage($ex->getMessage(), $ex->getHint());
 } catch (\OC\User\LoginException $ex) {
-	\OC::loadDefaultEnabledAppTheme();
 	OC_Response::setStatus(OC_Response::STATUS_FORBIDDEN);
 	OC_Template::printErrorPage($ex->getMessage());
 } catch (\OCP\Files\ForbiddenException $ex) {
-	\OC::loadDefaultEnabledAppTheme();
 	OC_Response::setStatus(OC_Response::STATUS_FORBIDDEN);
 	OC_Template::printErrorPage($ex->getMessage());
 } catch (Exception $ex) {
-	\OC::loadDefaultEnabledAppTheme();
 	try {
 		\OC::$server->getLogger()->logException($ex, array('app' => 'index'));
 
@@ -91,7 +85,6 @@ try {
 		echo('</body></html>');
 	}
 } catch (Error $ex) {
-	\OC::loadDefaultEnabledAppTheme();
 	\OC::$server->getLogger()->logException($ex, array('app' => 'index'));
 	OC_Response::setStatus(OC_Response::STATUS_INTERNAL_SERVER_ERROR);
 	OC_Template::printExceptionErrorPage($ex);
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -217,12 +217,46 @@ class AppManager implements IAppManager {
 		if(OC_App::getAppPath($appId) === false) {
 			throw new \Exception("$appId can't be enabled since it is not installed.");
 		}
+		$this->canEnableTheme($appId);
+
 		$this->installedAppsCache[$appId] = 'yes';
 		$this->appConfig->setValue($appId, 'enabled', 'yes');
 		$this->dispatcher->dispatch(ManagerEvent::EVENT_APP_ENABLE, new ManagerEvent(
 			ManagerEvent::EVENT_APP_ENABLE, $appId
 		));
 		$this->clearAppsCache();
+	}
+
+	/**
+	 * Do not allow more than one active app-theme
+	 *
+	 * @param $appId
+	 * @throws \Exception
+	 */
+	protected function canEnableTheme($appId) {
+		$info = $this->getAppInfo($appId);
+		if (
+			isset($info['types'])
+			&& is_array($info['types'])
+			&& in_array('theme', $info['types'])
+		) {
+			$apps = $this->getInstalledApps();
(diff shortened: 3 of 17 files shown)


VicDeo:
@DeepDiver1975 in the current state it works and passes all the tests.
I tried to inject a ThemeService as a next step but it leads to the error 500 when any app-theme is active:
```
[Mon Dec 04 22:52:41.765941 2017] Fatal error: Cannot declare class OC_Theme, because the name is already in use in /ocpath/apps/theme-enterprise/defaults.php on line 16
[Mon Dec 04 22:52:41.765971 2017] Stack trace:
[Mon Dec 04 22:52:41.765986 2017]  1. {main}() /ocpath/index.php:0
[Mon Dec 04 22:52:41.765994 2017]  2. OC_Template::printExceptionErrorPage($exception = *uninitialized*, $fetchPage = *uninitialized*) /ocpath/index.php:84
[Mon Dec 04 22:52:41.766001 2017]  3. OC_Template->__construct($app = *uninitialized*, $name = *uninitialized*, $renderAs = *uninitialized*, $registerCall = *uninitialized*) /ocpath/lib/private/legacy/template.php:358
[Mon Dec 04 22:52:41.766007 2017]  4. OC_Defaults->__construct() /ocpath/lib/private/legacy/template.php:93
[Mon Dec 04 22:52:41.766012 2017]  5. require_once() /ocpath/lib/private/legacy/defaults.php:76
```

VicDeo:
@DeepDiver1975 plz disregard the previous comment. 
I didn't noticed that running test from CLI killed `files` directory for my user. What I broke is an output for error message informing about that. Fixing...

VicDeo:
@PVince81 IMHO it's better to disable the currently active theme instead of  blocking a new one.

Another thing: as soon as it is not possible to have multiple themes, we should  consider adding `theme` type into the list of protected apps (that can't be enabled for certain groups only) https://github.com/owncloud/core/blob/master/lib/private/App/AppManager.php#L49

PVince81:
> IMHO it's better to disable the currently active theme instead of blocking a new one.

@VicDeo in any case we need a warning to tell the admin whatever happened.

VicDeo:
@PVince81 
>  we need a warning to tell the admin whatever happened.

it's done. For now I did it as suggested: second theme can't be enabled with a warning.

VicDeo:
Failing tests are related to CI itself:
```
General error: 2006 MySQL server has gone away
```
```
+ wait-for-it mysqlmb4:3306
wait-for-it: waiting 15 seconds for mysqlmb4:3306
wait-for-it: timeout occurred after waiting 15 seconds for mysqlmb4:3306
```

core/Command/App/Enable.php
```diff
@@ -70,14 +70,20 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+			$groups = $input->getOption('groups');
+			if (empty($groups)) {
+				\OC_App::enable($appId);
+				$output->writeln($appId . ' enabled');
+			} else {
+				\OC_App::enable($appId, $groups);
+				$output->writeln($appId . ' enabled for groups: ' . implode(', ', $groups));
+			}
+		} catch (\Exception $e) {
+			$output->writeln($e->getMessage());
```
PVince81: log the stack trace to avoid future headaches ?

index.php
```diff
@@ -52,30 +52,24 @@
 try {
 
 	require_once __DIR__ . '/lib/base.php';
-
 	OC::handleRequest();
 
 } catch(\OC\ServiceUnavailableException $ex) {
-	\OC::loadDefaultEnabledAppTheme();
```
PVince81: glad to see these go!

lib/private/App/AppManager.php
```diff
@@ -217,6 +217,8 @@ public function enableApp($appId) {
 		if(OC_App::getAppPath($appId) === false) {
 			throw new \Exception("$appId can't be enabled since it is not installed.");
 		}
+		$this->canEnableTheme($appId);
```
PVince81: return boolean and throw here ? just reading this line makes it look like a mistake as it asks `can` but doesn't check the result.

Since you're using this function only once I think it's better to do the boolean thing.

lib/private/App/AppManager.php
```diff
@@ -225,6 +227,38 @@ public function enableApp($appId) {
...
+		$info = $this->getAppInfo($appId);
+		if (
+			isset($info['types'])
+			&& is_array($info['types'])
+			&& in_array('theme', $info['types'])
+		) {
+			$apps = $this->getInstalledApps();
+			foreach ($apps as $installedAppId) {
+				if ($this->isTheme($installedAppId)) {
+					throw new \Exception("$appId can't be enabled until $installedAppId is disabled.");
```
PVince81: We might have some fantastic app manager specific exceptions, maybe one would suit here ?

PVince81 (commented):
Looks good overall, see comments for a few minor things.

Great to see so many tests!

core/Command/App/Enable.php
VicDeo: @PVince81 Possible. But is t really needed for CLI command?

PVince81: Once people start reporting bugs and you don't have the stack trace at all, this will become painful...

lib/private/App/AppManager.php
VicDeo: @PVince81 This is what I was going to do initially. TBH I don't remember why I decided to do it this way finally.
May be due to Scrutinizer's pessimization for methods with too many decision points. (Yes, I look into the code analisys :) )

lib/private/App/AppManager.php
PVince81: that would already be better

lib/private/App/AppManager.php
VicDeo: @PVince81 a-ha. It allows to produce more descriptive message (with the blocking app-theme appid). Just by returning `false` this appid would be lost

VicDeo: I mean  the current message contains both blocking and blocked appIds: `new-shiny-theme can't be enabled until old-bloody-theme is disabled.`

PVince81: fine by me

VicDeo:
@PVince81 all except `canEnableTheme` addressed

core/Command/App/Enable.php
DeepDiver1975: just dont catch the exception then - afaik the symfony console will print the stack ...

core/Command/App/Enable.php
```diff
@@ -36,12 +37,17 @@ class Enable extends Command {
...
 
+	/** @var ILogger */
+	protected $logger;
+
 	/**
 	 * @param IAppManager $manager
+	 * @param ILogger $logger
 	 */
-	public function __construct(IAppManager $manager) {
+	public function __construct(IAppManager $manager, ILogger $logger) {
```
DeepDiver1975: from my pov a occ command shall not use the logger but print anything to the console

lib/private/App/AppManager.php
```diff
@@ -225,6 +228,38 @@ public function enableApp($appId) {
...
+	/**
+	 * Do not allow more than one active app-theme
+	 *
+	 * @param $appId
+	 * @throws AppManagerException
+	 */
+	protected function canEnableTheme($appId) {
+		$info = $this->getAppInfo($appId);
+		if (
+			isset($info['types'])
```
DeepDiver1975: isset($info['types']['theme']) - keep is simple

DeepDiver1975: isset($info['types']['theme']) - keep is simple

lib/private/L10N/Factory.php
```diff
@@ -180,32 +189,14 @@ public function findAvailableLanguages($app = null) {
...
-					}
-				}
-			}
+		$available = array_merge($available, $this->findAvailableLanguageFiles($dir));
+
+		// merge with translations from themes
+		$relativePath = substr($dir, strlen($this->serverRoot));
+		$themeDir = $this->getActiveThemeDirectory();
+		if ($themeDir !== '') {
+			$themeDir .= $relativePath;
```
DeepDiver1975: / is guaranteed to be in place here?

lib/private/Repair/DisableExtraThemes.php
```diff
@@ -0,0 +1,107 @@
...
+class DisableExtraThemes implements IRepairStep {
+
+	/** @var IAppManager */
+	protected $appManager;
+
+	/** @var IConfig */
+	protected $config;
+
+	/** @var IAppConfig */
+	protected $appConfig;
```
DeepDiver1975: since you already have IConfig - there is not need for appConfig

DeepDiver1975 (dismissed):

core/Command/App/Enable.php
VicDeo: @DeepDiver1975  please discuss this with @PVince81  as he requested logging :)

lib/private/L10N/Factory.php
VicDeo: yep

core/Command/App/Enable.php
DeepDiver1975: @PVince81 whats the background here? its actually easier for the user to see any error/exception in the console and not to look it up in the log file.

lib/private/Repair/DisableExtraThemes.php
VicDeo: @DeepDiver1975 I do need it https://github.com/owncloud/core/pull/29747/files#diff-cdbcd54405a2b5684c9aed2c1824d7f8R94

lib/private/App/AppManager.php
VicDeo: @DeepDiver1975 it'snot possible this way because `'theme'` is not a key but a value:
```
array(20) {
  'id' =>  string(16) "theme-enterprise"
  'name' =>  string(27) "ownCloud X Enterprise Theme"
'types' =>
  array(1) {
    [0] =>     string(5) "theme"
  }
```

core/Command/App/Enable.php
PVince81: as long as the stack trace appears **somewhere** we're fine

VicDeo: thanks, removing logging along with try/catch

VicDeo:
@DeepDiver1975 thanks, addressed/responded

lib/private/Repair/DisableExtraThemes.php
DeepDiver1975: this does not exist in IConfig? brilliant ...

DeepDiver1975 (approved):

VicDeo:
Stable10: https://github.com/owncloud/core/pull/29854

PR #29749, opened 2017-12-03
[stable10] Handle case of first CORS whitelist add-delete

PVince81:
Backprot of https://github.com/owncloud/core/pull/29742 to stable10

### DIFF
--- a/settings/Controller/CorsController.php
+++ b/settings/Controller/CorsController.php
@@ -90,13 +90,7 @@ class CorsController extends Controller {
 	 */
 	public function getDomains() {
 		$userId = $this->userId;
-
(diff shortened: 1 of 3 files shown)


phil-davis (approved):

phil-davis:
Needs independent QA testing.
1) Start with a clean system that has never had any CORS whitelist entries
2) Login and go to Settings->Personal->Security
3) Under CORS, Add Domain, enter a domain like "https://localhost" and press Add.
The domain should be added to the list of White-listed Domains.
4) Delete the domain.
The domain should be deleted.
5) Add a few domains, delete from different positions in the list, delete all domains again.
Domains should come and go from the list as they are added/deleted.

PR #29711, opened 2017-11-29
use Provisioning API for user administration in UI tests

individual-it:
## Description
creating/deleting and modifying users in the UI tests is done through the provisioning API and not by `occ`

## Motivation and Context
the main goal is to make it possible to run UI tests against a remote instance, e.g. inside docker. See also https://github.com/owncloud/qa-enterprise/pull/115

One test fails because of https://github.com/owncloud/core/issues/29698

## How Has This Been Tested?
run tests in travis

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29751, opened 2017-12-04
[stable10] use Provisioning API for user administration in UI tests

individual-it:
backport of #29711 to stable10


phil-davis (approved):

phil-davis:
@individual-it I expect this needs a rebase and then it will pickup the backports of other fixes and tests will pass.

PR #29745, opened 2017-12-01
Refactoring tests

carusogabriel:
## Description
I've refactored some tests with [`assertFileExists`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFileExists) and `assertFileNotExists`, [`assertFileEquals`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFileEquals), [`assertFalse`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse), [`assertTrue`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertTrue), [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertNull) and `assertNotNull` methods.

## How Has This Been Tested?
`PHPUnit`

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29754, opened 2017-12-05
Refactoring tests test2

phil-davis:
See if tests pass

### DIFF
--- a/apps/comments/tests/unit/Dav/CommentsNodeTest.php
+++ b/apps/comments/tests/unit/Dav/CommentsNodeTest.php
@@ -425,7 +425,7 @@ class CommentsNodeTest extends \Test\TestCase {
 			$this->assertSame($expected[$name], $value);
 			unset($expected[$name]);
 		}
-		$this->assertTrue(empty($expected));
+		$this->assertEmpty($expected);
 	}
 
 	public function readCommentProvider() {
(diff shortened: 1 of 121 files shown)


phil-davis:
@carusogabriel I think the changes in the last commit here will make it pass. There were a couple of refactoring errors, but also a couple of existing ``assertEquals()`` where that existing test is a bit loose (false, null, 0 are all equal - true, 1... are equal) but changing to a more strict ``assertTrue()`` etc breaks it. I put "TODO" comment lines in those places.

carusogabriel:
Gonna wait for `Travis CI` and replicate this changes in #29745

phil-davis:
I rebased again tro trigger CI - hopefully Jenkins is up and running by now.

phil-davis:
@carusogabriel all passing now.
The couple of places I commented "TODO" would need someone to look a bit harder and see exactly which of false,0,null and so on is the thing that really should be expected/asserted in order to change them to a more tight assert.

carusogabriel:
What should I do know?

phil-davis:
@carusogabriel I suggest just taking the code that works from here. The TODO items can be looked at later - they are existing "features" of the asserts in the tests.

carusogabriel:
I did, it’s now in the oficial PR. How should it be test there? It’s keeping failing

phil-davis:
See #29791

PR #29760, opened 2017-12-05
[stable10] Raise more useful message when constructor are not resolvable

DeepDiver1975:
Whenever resolved class C exists but one of its constructor arguments
cannot be resolved, the app framework would throw an exception that the
class C does not exist even though it does. This is misleading and leads
to confusion, anger and loss of time.

This fix expands the error message to include the message about the
constructor argument that wasn't properly forwarded.

backport of #29737


PVince81 (approved):
:+1:

PR #29788, opened 2017-12-07
Fix potential security hole in license.php

settermjd:
This PR sanitizes the arguments passed to `shell_exec` in `build/license.php`, which could potentially result in a remote code execution vulnerability, as the calls to `shell_exec` receive an unfiltered file path. This was raised in http://bit.ly/2AD9bhK. I've implemented the suggested fixes slightly changed for further sanity.

Also, would it be better to implement the scripts on https://github.com/owncloud/core/blob/fix-security-bug-in-license.php/build/license.php#L182 and https://github.com/owncloud/core/blob/fix-security-bug-in-license.php/build/license.php#L191 as bash and batch files?

### DIFF
--- a/build/license.php
+++ b/build/license.php
@@ -192,13 +192,20 @@ With help from many libraries and frameworks including:
 	 */
 	private function getAuthors($file, $gitRoot) {
 		// only add authors that changed code and not the license header
-		$licenseHeaderEndsAtLine = trim(shell_exec("grep -n '*/' $file | head -n 1 | cut -d ':' -f 1"));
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
Well - this script never reaches the user. It is never part of the release. So I question the need to put any effort in here.

settermjd:
It may well not, but is it not worth fixing, given that it's such a small fix?

DeepDiver1975:
@dependabot rebase

butonic (approved):

butonic:
@ownclouders rebase

ownclouders:

ownclouders:

settermjd:
Does this need to be backported?

DeepDiver1975:
> Does this need to be backported?

I see no real reason - THX

settermjd:
Ok. Thanks for clarifying.

PR #29791, opened 2017-12-08
Refactoring tests test3

phil-davis:
Check that code from #29745 passes all CI.


phil-davis:
The "real" PR for this has been merged.

PR #29319, opened 2017-10-20
Preview via dav

DeepDiver1975:
## Description

Instead of using old ajax endpoints (aka private api) previews can now be generated via the dav endpoint by appending a query string.

```
remote.php/dav/files/admin/%2Fwelcome.txt?x=75&y=75&c=7046dec79875de1c6f4a3566239077e5&forceIcon=0&preview=1
```

## Related Issue
refs #29087 

## ToDo:
- [x] still duplicate disposition header

## Motivation and Context
- DAV is ownClouds public api for files
- kill all ajax routes 

## How Has This Been Tested?
- [ ] manually with regular filestorage including versions
- [ ] manually with object torage including versions
- [x] unit tests

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #29797, opened 2017-12-08
Preview via dav UI tests

phil-davis:
This is code on top of #29319 
In the last commit here:

1) Extend the timeout when initiating an upload and waiting for the AJAX to complete - for some reason it is taking around 15 seconds to complete. Previous to the code changes here it was much quicker.

2) Split out the tests in ``tests/ui/features/sharingExternalOther/federationSharing.feature`` so that each of "upload overwrite", "upload a new file", "rename a file", "delete a file" are done as separate scenarios. It takes longer to run, but makes the results easier to sort out.

@DeepDiver1975 I expect that this will pass everything except the scenario "delete a file in a received share" - I think there is a real problem there not being a able to delete a file in a received federated share.


phil-davis:
Outdated

PR #29820, opened 2017-12-12
Check trashbin perms before moving to trash

PVince81:
## Description
Check whether the target trashbin is writable (ex: guests).
In the case where a guest deletes a file as recipient, the owner would
get a copy in their trash. The logic here will detect that the guests'
trashbin is not writeable so it will fall back to simply moving directly
to the owner's trashbin.

## Related Issue
https://github.com/owncloud/guests/issues/31 approach 1

## How Has This Been Tested?
Manually tested.

## Types of changes

## Checklist:

## TODOs:

- [ ] add unit tests if we agree on the approach

### DIFF
--- a/apps/files_trashbin/lib/Trashbin.php
+++ b/apps/files_trashbin/lib/Trashbin.php
@@ -141,11 +141,23 @@ class Trashbin {
 		}
 	}
 
+	/**
+	 * Sets up the trashbin for the given user
+	 *
+	 * @param string $user user id
+	 * @return bool true if trashbin is setup and usable, false otherwise
+	 */
 	private static function setUpTrash($user) {
 		$view = new View('/' . $user);
 		if (!$view->is_dir('files_trashbin')) {
 			$view->mkdir('files_trashbin');
 		}
+
+		if (!$view->isUpdatable('files_trashbin')) {
+			// no trashbin access or denied
+			return false;
+		}
+
 		if (!$view->is_dir('files_trashbin/files')) {
 			$view->mkdir('files_trashbin/files');
 		}
@@ -155,6 +167,8 @@ class Trashbin {
 		if (!$view->is_dir('files_trashbin/keys')) {
 			$view->mkdir('files_trashbin/keys');
 		}
+
+		return true;
 	}
 
 
@@ -249,7 +263,14 @@ class Trashbin {
 			return true;
 		}
 
-		self::setUpTrash($user);
+		if (!self::setUpTrash($user)) {
+			// trashbin not usable for user (ex: guest), switch to owner only
+			$user = $owner;
+			if (!self::setUpTrash($owner)) {
+				// nothing to do as no trash is available anywheree
+				return true;
+			}
+		}
 		if ($owner !== $user) {
 			// also setup for owner
 			self::setUpTrash($owner);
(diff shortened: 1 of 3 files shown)


PVince81:
decision clarified and we agreed on the approach: need to update unit tests

PVince81:
Unit test added. Needs review.

IljaN (approved):
LGTM

PVince81:
Grrrr... Oracle

```
1) OCA\Files_Trashbin\Tests\StorageTest::testDeleteFolderAsReadOnlyRecipient
OCP\Files\NotPermittedException: No create permission for folder /recipient_1Ky0E38xz5c26/files

/drone/src/lib/private/Files/Node/Folder.php:162
/drone/src/lib/private/Files/Node/Root.php:369
/drone/src/lib/private/Server.php:982
/drone/src/tests/lib/TestCase.php:373
/drone/src/apps/files_trashbin/tests/StorageTest.php:276
```

PVince81:
I did some debugging with MariaDB and I can see that the `Folder::checkPermissions` call for create permissions is done on "/recipient_xyz" which allows the creation of a "files" folder for a read-only user.
Not sure if that's intended. That aside, I don't understand why Oracle would have a different result there.

However the output with Oracle seems to imply that the check is done on "/recipient_xyz/files" instead, which disallows anything else than read.

IljaN:
The "ReadOnlyJail" is initialized before first-login directory structure creation, so it need to allow  "/files" creation.

PVince81:
@IljaN thanks for the info.

Seems I'll need to boot up an Oracle docker to solve this mystery.

PVince81:
When running tests on Oracle: the individual test file works, but when running all the tests, some kind of side effect is causing the issue above.

PVince81:
Debug comparison: `getFileInfo("/recipient_xyz")` always reads the permissions from oc_filecache.
On Oracle the database contains the permissions 1 but on other DBs it contains 4. Go figure...

Maybe the read-only wrapper kicks in much earlier with Oracle, could be scan time (when?!) and reads the already masked value.

PVince81:
I checked the database and both Oracle and MariaDB have the same "permissions" value.

However the `getFileInfo` call still returns 1 for Oracle.

PVince81:
Remember that Oracle consider empty strings as null.
When reading the cache entry for "/recipient_xyz" this is actually the filecache entry with name === "" which is actually null on Oracle. So what happens is that `$data['name']` is null instead of empty string.

I didn't yet reach the correct code but I suspect that there is a check against empty string somewhere that might not kick in when null. I remember @IljaN saying that the permissions on the root was more permissive than on "/files".

Let me fix that in the Cache and convert null to empty string... I'm sure this will somewhat solve a lot of other yet undiscovered Oracle problems...

IljaN:
@PVince81 The root folder is indeed more permissive because cache, avatars directories needs to be created in there.

PVince81:
Confirmed, it works with this fix https://github.com/owncloud/core/pull/29820/commits/c30c8ec7da42a8ab302da2e422ac708bdda191a6

PVince81:
Please do review my last commit. I think it looks alright.

IljaN (approved):

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

PVince81:
stable10: https://github.com/owncloud/core/pull/30224

PVince81:
I missed the trash commit, backported here: https://github.com/owncloud/core/pull/30240

Issue #28341, opened 2017-07-08
Wrong PROPFIND answer with depth infinity requests


blancjerome:
Hi,

I found a strange behavior with infinity depth propfind requests : under subdirectories, files are reported as directories. This is due to a Collection resource type returned in the propfind response. It occurs only for level 2 subdirectories and above.

This behavior appeared with 10.0.x version of owncloud, but it maybe was there in the 9.1.x as I discovered it after upgrading from a 9.0.x version.

I didn't found any report of this bug, except with nextcloud, but no answer yet (https://github.com/nextcloud/server/issues/5543).
I don't know what's the relationship between owncloud and nextcloud except the fact that they obviously share some code.

Because my client (Documents by Readdle) couldn't sync properly, I took a look at the code and found where the problem is.

In /var/www/html/owncloud/lib/composer/sabre/dav/lib/DAV/Server.php, the "generatePathNodes" method clones "ProFind" objects. But as these objects have been created for the containing directory, they already contain the "Collection" value for the "ResourceType" property. So the workaround is creating a brand new "PropFind" object instead of cloning it.

I hope the correction will be included in the future version of owncloud, I don't think it has an impact on memory footprint, maybe a bit on performance because Propfind objects have to be filled again. But I haven't seen any significant change.

Here is the new code for the method :

     * Small helper to support PROPFIND with DEPTH_INFINITY.
     *
     *
     * @param PropFind $propFind
     * @param array $yieldFirst
     * @return \Iterator
     */
    private function generatePathNodes(PropFind $propFind, array $yieldFirst = null) {
        if ($yieldFirst !== null) {
            yield $yieldFirst;
        }
        $newDepth = $propFind->getDepth();
        $path = $propFind->getPath();

        if ($newDepth !== self::DEPTH_INFINITY) {
            $newDepth--;
        }
	
	//FIX in order to create new PropFind objects and not clone them
	$propertyNames = $propFind->getRequestedProperties();
	$propFindType = $propertyNames ? PropFind::NORMAL : PropFind::ALLPROPS;

        foreach ($this->tree->getChildren($path) as $childNode) {
	     //FIX : no cloning, creating new PropFind objects
            //$subPropFind = clone $propFind;
            //$subPropFind->setDepth($newDepth);
            if ($path !== '') {
                $subPath = $path . '/' . $childNode->getName();
            } else {
                $subPath = $childNode->getName();
            }
            //$subPropFind->setPath($subPath);

[...]

Issue #27542, opened 2017-03-30
Move bundled apps to market


PVince81:
The following apps are currently bundled in the ownCloud server community release.
Some apps come from the core repo, some from their own.

|app name|description|current repo|action|done?|
|---|---|---|---|---|
|activity|Activity app|separate|publish in market| ✔️  |
|comments|Comments GUI|core|publish in market?||
|configreport|Config Report|separate|keep||
|dav|DAV architecture|core|keep||
|encryption|Default encryption module + master key|core|publish in market|
|example-theme|Example theme|core||removed|
|external|External sites|separate|publish in market||
|federatedfilesharing|Federated file sharing|core|keep bundled, core depends on it and is always enabled||
|federation|Federation engine|core|removable, not sure||
|files|Files GUI|core|keep in core||
|files_antivirus|Antivirus|separate|publish in market|✔️|
|files_external|External storage backend + config GUI|core|keep for now, backends to be moved to separate repos/apps||
|files_pdfviewer|PDF viewer|separate|publish in market| ✔️ |
|files_sharing|Files sharing|core|keep in core, tightly coupled with core. New frontend from https://github.com/owncloud/files_sharing_fe will progressively replace old one||
|files_texteditor|Text editor|separate|publish in market| ✔️ |
|files_trashbin|Trashbin/Deleted files|core|?||
|files_versions|Versions|core|?||
|files_videoplayer|Video player|separate|?||
|firstrunwizard|First run popup|separate|publish in market||
|gallery|Gallery|separate|publish in market| ✔️ |
|notifications|Notification provider + indicator|separate|publish in market|  |
|provisioning_api|Provisioning API|core|keep||
|systemtags|System tags GUI|core|publish in market?||
|templateeditor|Email template editor|separate|publish in market|✔️|
|updatenotification|Web update notification|core|?||
|user_external|External user backends IMAP, SMB, CAS|separate|publish in market||
|user_ldap|LDAP user backend|separate|publish in market| ✔️ |

PVince81:
To be "Done" an item must:
- be removed from the release script
- have a market entry
- have an initial released version in the market which is visible from inside ownCloud

PVince81:
Trashbin and versions have some cross-dependencies between each other and also some core code depends on them. They can be disabled and OC might work fine though.
Not sure if that's a good enough reason to move these to the market.

PVince81:
[...]

PR #29822, opened 2017-12-12
Update libs and pin version to 5.6

PVince81:
Update dependencies to latest version compatible with PHP 5.6.32.

Pinned version to PHP 5.6.32 because if you update the libs with composer of PHP 7 it might pick the wrong versions... So better be safe.

This PR is to be backported to stable10.
At some point we can redo the same and pin to PHP 7.1 for 10.1 on master.

Things to retest:
- Dropbox external storage app (flysystem is pulling a Dropbox v2 adapter)
- CI

### DIFF
--- a/apps/dav/tests/unit/CardDAV/BirthdayServiceTest.php
+++ b/apps/dav/tests/unit/CardDAV/BirthdayServiceTest.php
@@ -125,9 +125,9 @@ class BirthdayServiceTest extends TestCase {
 		if ($expectedOp === 'create') {
 			$service->expects($this->exactly(3))->method('buildDateFromContact')->willReturn(new VCalendar());
 			$this->calDav->expects($this->exactly(3))->method('createCalendarObject')->withConsecutive(
-				[1234, 'default-gump.vcf.ics', "BEGIN:VCALENDAR\r\nVERSION:2.0\r\nPRODID:-//Sabre//Sabre VObject 4.1.2//EN\r\nCALSCALE:GREGORIAN\r\nEND:VCALENDAR\r\n"],
-				[1234, 'default-gump.vcf-death.ics', "BEGIN:VCALENDAR\r\nVERSION:2.0\r\nPRODID:-//Sabre//Sabre VObject 4.1.2//EN\r\nCALSCALE:GREGORIAN\r\nEND:VCALENDAR\r\n"],
-				[1234, 'default-gump.vcf-anniversary.ics', "BEGIN:VCALENDAR\r\nVERSION:2.0\r\nPRODID:-//Sabre//Sabre VObject 4.1.2//EN\r\nCALSCALE:GREGORIAN\r\nEND:VCALENDAR\r\n"]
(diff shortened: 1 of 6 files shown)


PVince81:
Hmm, we should also update the libs in "apps/files_external"... but for that I'd rather we first move out the backends and update the libs then: https://github.com/owncloud/core/issues/27542

PVince81:
- [ ] update the JS libs as well

DeepDiver1975:
- [ ] dav tests need to adjust the version

````
OCA\DAV\Tests\unit\CardDAV\ConverterTest::testCreation with data set #0 ('BEGIN:VCARD\nVERSION:3.0\nPRODI...VCARD\n')
--- Expected
+++ Actual
@@ @@
 'BEGIN:VCARD
 VERSION:3.0
-PRODID:-//Sabre//Sabre VObject 4.1.2//EN
+PRODID:-//Sabre//Sabre VObject 4.1.3//EN
 UID:12345
 FN:12345
 N:12345;;;;
 PHOTO;ENCODING=b;TYPE=JPEG:MTIzNDU2Nzg5
 END:VCARD
 '

phil-davis:
I discovered the Sabre VObject version bump last week when rebasing the Guzzle version PR:
https://github.com/owncloud/core/pull/29773/commits

It will be good to have that merged into master so I can remove the commit from Guzzle changes.

PVince81:
Updated JS libs to latest minor versions.

To retest:
- [x] TEST: md5 / base64: public page upload with password
- [x] TEST: clipboard: copy public link
- [x] TEST: davclient, es6-promise: Webdav works
- [x] TEST: moment: relative timestamps still work in web UI
- [x] TEST: backbone JS: comments still work
- [x] TEST: strengthify: password meter in personal settings
- [x] TEST: jcrop: avatar setting
- [x] TEST: snap: left sidebar toggle on mobile
- [ ] TEST: bowser: IE < 11 or older browsers warning
- [ ] TEST: timezone detect (?)

Note: we should consider major release update for 10.1. Note that jquery update will break a lot of things so need to plan some time...

PVince81:
We should also look at have the new Sabre changes released so we can update to it for https://github.com/owncloud/core/issues/28341

PVince81:
- [ ] update node libs (ex: karma, jasmine, etc)

PVince81:
No output for `npm update` or `npm outdated`, so it seems our npm libs are already up to date.

PVince81:
Regarding timezone, I didn't find what code uses it, maybe it's obsolete.

Anyway, I checked the hidden input fields and their value was set correctly, so the library update worked.

PVince81:
:+1: for @DeepDiver1975's test fixes.

Review for the update in general ?

I think we should backport this afterwards.

DeepDiver1975 (approved):

PVince81:
@DeepDiver1975 can you take care of the backport and version updates in general ? seems the libs updated again...

phil-davis:
@PVince81 @DeepDiver1975 I see a backport request label here, but it did not get backported.
So is it planned to get all the composer stuff up-to-date in ``stable10`` for the next point release?

PVince81:
That would be nice, yes

phil-davis:
PR #30430 has what would be the backported code for ``bower.json`` and ``core/js/core.json``
See comment there deciding not to push those version bumps back to ``stable10``

phil-davis:
All needed backports are done.

PR #29661, opened 2017-11-23
Adding files external tests for webdav to drone

SergioBertolinSG:

PR #29838, opened 2017-12-13
Clean up of drone shell scripts and duplicated steps

SergioBertolinSG:
Removed double step for webdav files_external in drone

A double step was added in https://github.com/owncloud/core/pull/29661#event-1385643777

This PR objective is to fix it.


SergioBertolinSG:
The php-unit-tests step is not running.

tboerger:
As mentioned on chat, please keep the step and test suite naming and just merge the shell scripts and add the coverage enable/disable variables.

SergioBertolinSG:
If you see the .drone.yml you can realize that there are three sections which are very similar: phpunit, coverage and files_external. All can be combined into php-unit-tests.

tboerger:
Merging the test runs is totally fine, but just keep the naming.

SergioBertolinSG:
@tboerger ok now?

tboerger (approved):
Seems fine

SergioBertolinSG:
There is a problem in mariadb tests, there are running less tests than expected:

OK, but incomplete, skipped, or risky tests!
Tests: 4555, Assertions: 24815, Skipped: 103.

In a previous pull request they were:

OK, but incomplete, skipped, or risky tests!
Tests: 9532, Assertions: 35726, Skipped: 200.

tboerger:
Because you skip the db group, of course you will loose lots of test cases

PVince81:
does this need a second review if anything was changed since approval ?

tests/drone/test-phpunit.sh
```diff
@@ -56,11 +56,49 @@ fi
 ./occ app:enable federation
 ./occ app:enable federatedfilesharing
 
-if [[ "${DB_TYPE}" == "none" || "${DB_TYPE}" == "sqlite" ]]; then
+if [[ "${DB_TYPE}" == "none" || "${DB_TYPE}" == "sqlite" || "${ENABLE_COVERAGE}" == "true" ]]; then
```
tboerger: Why do you disable the db group for coverage? Now we won't get the coverage report for the tests that require a database?

DeepDiver1975:
as far as I can tell coverage for files external is missing.

tests/drone/test-phpunit.sh
SergioBertolinSG: It is the opposite isn't? without this change only 4555 tests run, with it 9532.

Otherwise the codecov will fail, covering less than before.

DeepDiver1975:
> as far as I can tell coverage for files external is missing.
> Merge state

also missing on master @SergioBertolinSG can you have a look please? THX
Let me know if you need support

SergioBertolinSG:
> as far as I can tell coverage for files external is missing.

It is this one, isn't?

screen shot 2018-01-09 at 10 42 04

DeepDiver1975:
> It is this one, isn't?

yes - but if you go into lib/Lib/Backend - I expect code coverage on OwnCloud.php - because this is the test which is executed .... there is something wrong

DeepDiver1975:
looks like you missed the explicit call to test the specific backend - see https://github.com/owncloud/core/blob/c5407f3b9140eb4058798f3c21ff4a3ac6c148b0/build/autotest-external.sh#L202-L217

tests/drone/test-phpunit.sh
```diff
@@ -86,18 +86,21 @@ DELIM
...
 }
 
+FILES_EXTERNAL_BACKEND_PATH=../apps/files_external/tests/Storage
+FILES_EXTERNAL_TEST_TO_RUN=${FILES_EXTERNAL_TYPE}Test.php
+
 
 if [[ ${ENABLE_COVERAGE} == "true" ]]; then
     if [[ -n ${FILES_EXTERNAL_TYPE} ]]; then
         set_up_external_storage
-        exec phpdbg -d memory_limit=4096M -rr ./lib/composer/bin/phpunit --configuration tests/phpunit-autotest-external.xml ${GROUP} --coverage-clover tests/autotest-external-clover-${DB_TYPE}.xml
```
DeepDiver1975: leave this as well - this is the generic test execution.

I know this is confusing ....

SergioBertolinSG:
There is still a problem with external storage, it is not being detected as configured:

ownCloud backend not configured

SergioBertolinSG:
Last commit skips the particular run of WebdavTest.php. We found that OwncloudTest.php was not running correctly due to a config problem and WebdavTest.php was not running in Jenkins in first place. So we decided to skip this particular test for the moment and focus in moving everything left from Jenkins.

tests/drone/test-phpunit.sh
```diff
@@ -56,11 +56,59 @@ fi
...
+
+FILES_EXTERNAL_BACKEND_PATH=apps/files_external/tests/Storage
+FILES_EXTERNAL_TEST_TO_RUN=${FILES_EXTERNAL_TYPE}Test.php
 
+
+if [[ ${ENABLE_COVERAGE} == "true" ]]; then
+    if [[ -n ${FILES_EXTERNAL_TYPE} ]]; then
+        set_up_external_storage
+        phpdbg -d memory_limit=4096M -rr ./lib/composer/bin/phpunit --configuration tests/phpunit-autotest-external.xml ${GROUP} --coverage-clover tests/autotest-external-clover-${DB_TYPE}.xml
+        if [[ ${FILES_EXTERNAL_TYPE} !=  "Webdav" ]]; then
```
DeepDiver1975: why?

.drone.yml
```diff
@@ -165,10 +141,10 @@ services:
         DB_TYPE: postgres
 
   owncloud_external:
-    image: owncloud/server:latest
+    image: morrisjobke/owncloud:latest
```
patrickjahns: What is the reason of moving to a third party owncloud image?

patrickjahns (changes requested):
@SergioBertolinSG 
please do not include any 3rdparty owncloud containers

.drone.yml
DeepDiver1975: morrisjobke/owncloud is used for years in jenkins. First tries with owncloud/server did show errors.
My request was to first move the jobs as is and enhance afterwards.

patrickjahns: > My request was to first move the jobs as is and enhance afterwards.    

Fair enough - i've put this as a task in the backlog 
To move forward with the drone migration lets get this PR wrapped up, and we can divide&conquer the remaining tasks 

> First tries with owncloud/server did show errors.  

Do we have links to the logs here? - would like to link these findings in the backlog for easy access 

I don't see the need for the owncloud server container actually - a minimal apache + mod_dav configuration should be sufficient for this scenario. It's faster and more lightweight.
OC -> OC webdav should be covered via federated sharing scenarios from the integration tests

patrickjahns (approved):

.drone.yml
DeepDiver1975: > I don't see the need for the owncloud server container actually - a minimal apache + mod_dav configuration should be sufficient for this scenario. It's faster and more lightweight.

this is a second test - see the start-webdav-apache.sh 

and yes - we shall test both - different implementation behave different

DeepDiver1975 (changes requested):
- cleanup git history @SergioBertolinSG 
- have a look at codecov @DeepDiver1975

DeepDiver1975:
- https://drone.owncloud.com/owncloud/core/2569/71 generates the coverage file tests/autotest-external-clover-sqlite-Webdav.xml
- this one is not uploaded - https://drone.owncloud.com/owncloud/core/2569/72

.drone.yml
```diff
@@ -90,6 +90,7 @@ pipeline:
     files:
      - tests/autotest-clover-${DB_TYPE}.xml
      - tests/autotest-external-clover-${DB_TYPE}.xml
+     - tests/autotest-external-clover-${DB_TYPE}-${FILES_EXTERNAL_TYPE}.xml
```
DeepDiver1975: what about using wildcards? like tests/autotest-clover-*.xml and make sure all clover reports generate the report. Please read the docs of the plugin

DeepDiver1975 (changes requested):

.drone.yml
SergioBertolinSG: Not sure, the plugin is not using quotes:
https://github.com/robertstettner/drone-codecov/blob/master/entrypoint.sh#L22

and they look necessary:
https://github.com/codecov/codecov-bash/blob/master/codecov#L98-L99

DeepDiver1975: Please give it a try ...

SergioBertolinSG: After commit b51415b :

2018/01/18 13:00:58 using coverage files: tests/autotest-clover-*.xml, tests/autotest-external-clover-*.xml
2018/01/18 13:00:58 no coverage files found

tboerger: I have written the plugin and globbing works fine, it just needs to be documented properly

SergioBertolinSG: it is not using robertstettner one?

tboerger: Nope, https://github.com/drone-plugins/drone-codecov

All plugins/* images MUST be built within this github org.

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

.drone.yml
```diff
@@ -101,24 +75,28 @@ pipeline:
     environment:
       - PHP_VERSION=${PHP_VERSION}
       - DB_TYPE=${DB_TYPE}
+      - ENABLE_COVERAGE=${ENABLE_COVERAGE}
```
patrickjahns: Please add `FILES_EXTERNAL_TYPE=${FILES_EXTERNAL_TYPE}` to the environment here. Running `drone exec` locally would otherwise fail.

patrickjahns:
I used this PR as basis for #30207 and also noticed, that many of the vars in the script are not quoted. Run into quite some scenarios when debugging another failure, where the bash tests would also not evaluate correctly - please always quote relevant vars (background: http://mywiki.wooledge.org/Quotes )

PR #29842, opened 2017-12-14
skip known to fail tests on 10.0.3

individual-it:
## Description
skip tests that we know do fail when running on 10.0.3

## Related Issue
https://github.com/owncloud/qa-enterprise/pull/115

## Motivation and Context
we want to run tests on arbitrary oC versions, so we need to skip tests that we know would fails in a specific version

## How Has This Been Tested?
running tests on new docker system https://github.com/owncloud/qa-enterprise/pull/115

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/other/managingGroups.feature
+++ b/tests/ui/features/other/managingGroups.feature
@@ -8,13 +8,42 @@ So that access to resources can be controlled more effectively
 		And I am logged in as admin
 		And I am on the users page
 
-	Scenario: delete groups
+	@skipOnOcV10.0.3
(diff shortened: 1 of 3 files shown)


phil-davis (approved):
This will work, so good. But it is going to get messy in future because we will not be able to merge "examples" tables from multiple scenarios after a bug is fixed. I guess there can be a point at which we say "no more support for running current tests against old V10.0.2 V10.0.3 etc.

individual-it:
backport in  #29848

PR #29853, opened 2017-12-14
Do not auto-disable apps of certain types

### LINKED ISSUE #29850: Auto disabling apps on exceptions can break instances
During initialisation, owncloud loads apps. This involves 'requiring' the app.php file which is wrapped in a try catch block, and if an exception is caught, the app will be disabled (with exception of shipped: alwaysEnabled app).

This is especially damaging when encryption or authentication apps are disabled as this can render the instance unusable for all users when the problem may have been intermittent -> and you would only see the exception once.

Proposal: never disable 'authentication' or 'filesystem' type apps.

PVince81:
## Description
Whenever the file appinfo/app.php throws an exception, the app manager
would disable the app automatically, believing it is broken. While this
might work fine for broken third party apps, it is crucial to keep
filesystem and authentication apps enabled in case of random errors
(like Redis being temporarily not available)

## Related Issue
Fixes https://github.com/owncloud/core/issues/29850

## How Has This Been Tested?
Throw an exception randomly in user_ldap's appinfo/app.php and refresh the page until an error appears.

Before this fix: app would be disabled.
After this fix: error page and app stays enabled.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/legacy/app.php
+++ b/lib/private/legacy/app.php
@@ -245,8 +245,12 @@ class OC_App {
 			\OC::$server->getLogger()->logException($ex);
 			$blacklist = \OC::$server->getAppManager()->getAlwaysEnabledApps();
 			if (!in_array($app, $blacklist)) {
-				\OC::$server->getLogger()->warning('Could not load app "' . $app . '", it will be disabled', array('app' => 'core'));
-				self::disable($app);
+				if (!self::isType($app, ['authentication', 'filesystem'])) {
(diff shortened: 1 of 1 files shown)


butonic (approved):

PVince81:
Ignoring coverage as this is old untestable legacy static code...

PVince81:
stable10: https://github.com/owncloud/core/pull/29870

Issue #29877, opened 2017-12-18
Hook for cancelling public link creation


PVince81:
Provide a hook to let apps cancel public link creation an exception + error message in case this needs to be blocked (ex: due to permissions)

PVince81:
nice to have as there is a workaround using old hooks + throwing exception

Issue #29876, opened 2017-12-18
Hook for cancelling zip download


PVince81:
Provide a hook for zip folder download to let apps react on the list of requested files for download and give them a chance to cancel the download in case of permission issues.

PR #29885, opened 2017-12-18
Adding new Symfony events for zipfile download and public link share

sharidas:
Adding new Symfony events, to check, if creating
shares or downloading zips are blocked by any
apps or not.

## Description

Adding new Symfony events to check if any apps revoked the premission to download zip files or creating public links.

## Related Issue

https://github.com/owncloud/core/issues/29877
https://github.com/owncloud/core/issues/29876

## Motivation and Context

Adding new symfony events which can be used by the apps to see if creating public links for files/folders are allowed or not. And also see if downloading zip files are allowed or not.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/lib/API/Share20OCS.php
+++ b/apps/files_sharing/lib/API/Share20OCS.php
@@ -37,6 +37,8 @@ use OCP\Share\Exceptions\GenericShareException;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Share\IManager;
 use OCP\Share\IShare;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class Share20OCS
@@ -69,6 +71,9 @@ class Share20OCS {
 	 */
 	private $additionalInfoField;
 
+	/** @var \Symfony\Component\EventDispatcher\EventDispatcherInterface  */
+	private $eventDispatcher;
+
 	/**
 	 * Share20OCS constructor.
 	 *
@@ -81,6 +86,7 @@ class Share20OCS {
 	 * @param IUser $currentUser
 	 * @param IL10N $l10n
 	 * @param IConfig $config
+	 * @param EventDispatcher $eventDispatcher
 	 */
 	public function __construct(
 			IManager $shareManager,
@@ -91,7 +97,8 @@ class Share20OCS {
 			IURLGenerator $urlGenerator,
 			IUser $currentUser,
 			IL10N $l10n,
-			IConfig $config
+			IConfig $config,
+			EventDispatcher $eventDispatcher
 	) {
 		$this->shareManager = $shareManager;
 		$this->userManager = $userManager;
@@ -103,6 +110,7 @@ class Share20OCS {
 		$this->l = $l10n;
 		$this->config = $config;
 		$this->additionalInfoField = $this->config->getAppValue('core', 'user_additional_info_field', '');
+		$this->eventDispatcher = $eventDispatcher;
 	}
 
 	/**
@@ -272,6 +280,7 @@ class Share20OCS {
 	 * @return \OC\OCS\Result
 	 */
 	public function createShare() {
+		$event = new GenericEvent(null);
 		$share = $this->shareManager->newShare();
 
 		if (!$this->shareManager->shareApiEnabled()) {
@@ -287,6 +296,15 @@ class Share20OCS {
 		}
 
 		$userFolder = $this->rootFolder->getUserFolder($this->currentUser->getUID());
+		$event->setArgument('userFolder', $userFolder);
+		$event->setArgument('path', $path);
+		$event->setArgument('name', $name);
+		$event->setArgument('run', true);
+		//Dispatch an event to see if creating shares is blocked by any app
+		$this->eventDispatcher->dispatch('share.beforeCreate', $event);
+		if ($event->getArgument('run') === false) {
+			return new \OC\OCS\Result(null, 403, $this->l->t('No permission to create share'));
+		}
 		try {
 			$path = $userFolder->get($path);
 		} catch (NotFoundException $e) {
@@ -466,12 +484,16 @@ class Share20OCS {
 
 		$share->getNode()->unlock(\OCP\Lock\ILockingProvider::LOCK_SHARED);
 
+		$event = new GenericEvent(null, []);
+		$event->setArgument('output', $output);
+		$event->setArgument('result', 'success');
+		$this->eventDispatcher->dispatch('share.afterCreate', $event);
 		return new \OC\OCS\Result($output);
 	}
 
 	/**
 	 * @param \OCP\Files\File|\OCP\Files\Folder $node
-	 * @param boolean $includeTags 
+	 * @param boolean $includeTags
 	 * @return \OC\OCS\Result
 	 */
 	private function getSharedWithMe($node = null, $includeTags) {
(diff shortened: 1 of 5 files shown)


lib/private/legacy/files.php
```diff
@@ -139,6 +139,13 @@ public static function get($dir, $files, $params = null) {
 				}
 			}
 
+			//Dispatch an event to see if any apps have problem with download
+			$event = new \Symfony\Component\EventDispatcher\GenericEvent(null, ['files' => $files]);
```
PVince81: Please double check what happens if "$dir" is set with a single value and multiple "$files" are set. It is likely that we also need to pass `$dir`. The alternative is to have a loop that prepends the directory to all given `$files` entries.

lib/private/legacy/files.php
```diff
@@ -139,6 +139,13 @@ public static function get($dir, $files, $params = null) {
...
+			OC::$server->getEventDispatcher()->dispatch('file.beforecreatezip', $event);
+			if ($event->hasArgument('run') and ($event->getArgument('run') === false) and ($event->hasArgument('message'))) {
+				throw new Exception();
```
PVince81: A generic exception without a message will confuse the user. Maybe throw ForbiddenException instead with a message "Access denied".

In general I'd rather expect event listeners to directly throw an exception from inside with a proper message, so the exception itself would already be forwarded by not catching it in the `dispatch()` call.

apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -287,6 +293,15 @@ public function createShare() {
 		}
 
 		$userFolder = $this->rootFolder->getUserFolder($this->currentUser->getUID());
+		$event->setArgument('userFolder', $userFolder);
+		$event->setArgument('path', $path);
+		$event->setArgument('name', $name);
+		$event->setArgument('run', true);
+		//Dispatch an event to see if creating shares is blocked by any app
+		$this->eventDispatcher->dispatch('file.beforecreateShare', $event);
```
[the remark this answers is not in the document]
PVince81: And also catch exception here and transform it to an OCS Result with the exception message. I expect that listeners will throw exceptions in case of denied access.

apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -287,6 +293,15 @@ public function createShare() {
...
+		if ($event->getArgument('run') === false) {
+			return new \OC\OCS\Result(null, 404, $this->l->t('No permission to create share'));
```
PVince81: no permission is a 403 forbidden, not 404 not found

PVince81 (changes requested):
See comments above.

Note that for ZIP download I think there are at least 3 cases:

- URL with "files" with only one file => direct download
- URL with "files" and multiple files with full path => download all of them as ZIP
- URL with "dir" containing base dir and "files" containing a list of files inside that dir => prepend dir to all files and download as ZIP
- URL with "dir" without "files" => download said folder as ZIP

this is from memory, please check the code

lib/private/legacy/files.php
sharidas: Here is the snapshot I took from IDE: 
getmethod_singledir_multiplefiles

Basically during testing I have used a directory and a file. Above is the screenshot of the `get` method with single dir `newfolder` and multiple files: `welcome.txt` and `rename_testfile.txt`. The newfolder and the files reside in the home folder of the user. The `$dir` is "/" in the above screenshot. I tried to create a folder inside `newfolder` and then created few files inside `newfolder`. In this case also `$dir` remains "/".

sharidas:
Made the following changes:
- injected the event dispatcher to the constructor.
- used camel case for event name
- Replaced generic exception with ForbiddedException
- Added `$dir` to the event array
- Added after event to make sure the creation of share is successful.

apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -466,12 +484,15 @@ public function createShare() {
 
 		$share->getNode()->unlock(\OCP\Lock\ILockingProvider::LOCK_SHARED);
 
+		$event->setArgument('output', $output);
```
PVince81: create a new event instance, don't  resend the same one

apps/files_sharing/tests/ApiTest.php
```diff
@@ -115,6 +116,10 @@ private function createOCS($request, $userId) {
 				return vsprintf($text, $parameters);
 			}));
 
+		\OC::$server->getEventDispatcher()->addListener('share.beforeCreate', function (GenericEvent $event) {
+			$event->setArgument('run', true);
```
PVince81: is this really needed ? "run" is true by default

lib/private/legacy/files.php
```diff
@@ -139,6 +139,13 @@ public static function get($dir, $files, $params = null) {
 				}
 			}
 
+			//Dispatch an event to see if any apps have problem with download
+			$event = new \Symfony\Component\EventDispatcher\GenericEvent(null, ['dir' => $dir, 'files' => $files]);
+			OC::$server->getEventDispatcher()->dispatch('file.beforeCreateZip', $event);
+			if ($event->hasArgument('run') and ($event->getArgument('run') === false) and ($event->hasArgument('message'))) {
```
PVince81: Make sure the "run" argument is created by you and set to "true" by default like you did for share creation.
Any reason you chose to not simply use the same pattern like for share creation ?

rename "message" to "errorMessage"

PVince81 (changes requested):
better, still some changes needed

apps/files_sharing/tests/ApiTest.php
sharidas: This was added because the tests were failing. It was using false value from Share20OCSTest.php. That's the reason I set `run` value to true here.

sharidas:
Updated the changes:
- made sure that run is set in the files.php
- Created new instance of event for after share event 
- The reason for making run to true is addressed at : https://github.com/owncloud/core/pull/29885#discussion_r158502885

apps/files_sharing/tests/ApiTest.php
PVince81: Strange. Do you mean one test is influencing another ? This shouldn't happen. We need to find a way to reset the event dispatcher in `tearDown` to avoid test side effects.

lib/private/legacy/files.php
```diff
@@ -139,6 +139,13 @@ public static function get($dir, $files, $params = null) {
 				}
 			}
 
+			//Dispatch an event to see if any apps have problem with download
+			$event = new \Symfony\Component\EventDispatcher\GenericEvent(null, ['dir' => $dir, 'files' => $files, 'run' => true]);
+			OC::$server->getEventDispatcher()->dispatch('file.beforeCreateZip', $event);
+			if (($event->getArgument('run') === false) and ($event->hasArgument('errorMessage'))) {
```
PVince81: this should be `||`, not `and`. If run is false and no error message was supplied, we should still stop running.

lib/private/legacy/files.php
```diff
@@ -167,6 +174,9 @@ public static function get($dir, $files, $params = null) {
 			$streamer->finalize();
 			set_time_limit($executionTime);
 			self::unlockAllTheFiles($dir, $files, $getType, $view, $filename);
+			$event = new \Symfony\Component\EventDispatcher\GenericEvent(null, ['result' => 'success']);
```
PVince81: please also pass `dir` and `files` arguments here

sharidas:
@PVince81 Let me know if this change looks ok.

sharidas:
Created backport to stable10 PR: https://github.com/owncloud/core/pull/30067

PVince81 (approved):
:+1:

PR #29907, opened 2017-12-19
user with less than 3 characters

shresthamoheet:
Test for user with less than three valid characters.


tests/ui/features/other/login.feature
```diff
@@ -22,7 +22,19 @@ So that unauthorised access is impossible
 		And I logout
 		And I login with username "guiusr1" and password "pwd"
 		Then I should be redirected to a page with the title "Files - ownCloud"
-
+		
+		Scenario Outline: use the webUI to create a user with less than 3 characters
```
phil-davis: Too much indent in this line

Issue #29906, opened 2017-12-19
Move from bower to yarn


DeepDiver1975:
Bower is dead - long live yarn

PVince81:
seems yarn also uses package.json https://bower.io/blog/2017/how-to-migrate-away-from-bower
 
How to migrate away from Bower? · Bower blog
last line says "But initially, the only change required in code is to change any reference to bower_components with node_modules/@bower_components (though you can link it somewhere else in postinstall script)."
our bower_components is core/vendor
if symlinking is ok we could symlink the folder to there

PVince81:
adding @felixheidecke for a second opinion

PVince81:
PR https://github.com/owncloud/core/pull/29908

PVince81:
lets also do it for stable10: https://github.com/owncloud/core/issues/33651

PR #29908, opened 2017-12-19
Migrate to yarn

PVince81:
- [x] TEST: `make` still works
- [x] TEST: `make dist` still works
- [x] TEST: ownCloud still runs fine

- [x] Expecting CI failures. Will require `npm install -g yarn` on CI.
- [x] build server require yarn

Do we want to backport this ?


PVince81:
Fixes https://github.com/owncloud/core/issues/29906

PVince81:
- [ ] dev docs update

DeepDiver1975:
- [x] yarn is installed on jenkins

tests/drone/yarn-install.sh
```diff
@@ -6,4 +6,5 @@ if [[ "$(pwd)" == "$(cd "$(dirname "$0")"; pwd -P)" ]]; then
   exit 1
 fi
 
-exec npm install --prefix build
+cd build
+exec yarn install
```
tboerger: maybe `exec yarn install --modules-folder build/node_modules` won't require any dir change like before with npm?

DeepDiver1975: Tried that - did not work out

PVince81: indeed, I noticed that yarn doesn't have the `--prefix` argument but also doesn't complain if you use it... see makefile

DeepDiver1975: yarn install --cwd build

PVince81:
:+1: for @DeepDiver1975's changes.

Need review for mine.

@felixheidecke if you're familiar with yarn you could check them 😄

DeepDiver1975 (approved):

PR #29916, opened 2017-12-20
don't throw exception if usr/grp couldn't be deleted in AfterScenario

individual-it:
## Description
if a user or group could not be deleted in the tear-down phase of a test, do not throw an exception (test fails) but only show an info message

## Motivation and Context
For situations where a test attempts to create a user but could not, e.g. testing invalid usernames, the tear-down cannot delete the user because it wasn't created.
If a test-case should have created a user but didn't the test itself should fail, not the tear down function

## How Has This Been Tested?
travis will test for me

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/bootstrap/BasicStructure.php
+++ b/tests/ui/features/bootstrap/BasicStructure.php
@@ -396,8 +396,8 @@ trait BasicStructure {
 			);
 			
 			if ($result->getStatusCode() !== 200) {
-				throw new Exception(
-					"could not delete user. "
+				error_log(
+					"INFORMATION: could not delete user. '" . $user . "'"
 					. $result->getStatusCode() . " " . $result->getBody()
 				);
 			}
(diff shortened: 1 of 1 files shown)


tests/ui/features/bootstrap/BasicStructure.php
```diff
@@ -412,8 +412,8 @@ public function tearDownScenarioDeleteCreatedUsersAndGroups() {
 			);
 			
 			if ($result->getStatusCode() !== 200) {
-				throw new Exception(
-					"could not delete group. "
+				error_log(
+					"INFORMATION: could not delete user. '" . $user . "'"
```
phil-davis: cut-paste problem - should be "could not delete group" and ``$group``

phil-davis (commented):
After cut-paste, edit is also often required :)

phil-davis (approved):

tests/ui/features/bootstrap/BasicStructure.php
individual-it: :man_facepalming:

individual-it:
backport in #29929

Issue #29696, opened 2017-11-27
Public links: Add personal note and option to receive a copy


pmaier1:
**Add a message field to provide a personal note in the mail notification**

As a user I want to add a personal note when sharing with externals via public link to reference a previous conversations or just provide some hints to externals that never used ownCloud before.

The mail notification should be just like the default, but inserting a paragraph with my personal note after the introduction text.

**Add a checkbox "Receive a copy" to set yourself in CC of the mail notification**

As a user I want to be able to receive a copy (put me on CC) of the mail notifications I send out when sharing public links for that I can see what the recipients see and to be able to use that mail notification afterwards in my mail client (forward to others, reference the notification, etc.).

A mockup would be great!

@felixheidecke @PVince81

felixheidecke:
We could have a separate section below. The Textarea expands on putting text in as in this example:

![public_share_message](https://user-images.githubusercontent.com/12717530/33428910-165ec7f2-d5cb-11e7-99f9-68918fc0846e.gif)

pmaier1:
Yes, looks nice! Maybe we can make it expandable/collapsible to save some space. Recipient field requirements are here.

What about backend for inserting the message to the mail notification? Is this something easy? @PVince81

pmaier1:
Lol, just saw the German translation for "File Drop" in the mockup. This is bad and should be called "File Drop" as well. Can we make this a non-translatable string?

felixheidecke:
> Can we make this a non-translatable string?

Sure can :-D

felixheidecke:
As I recall, there is a greater rework of the sharing due. Can we postpone changes until then?

PVince81:
It might be possible to use the given message when sending emails, but the custom message will not be saved anywhere.

PVince81:
custom message as displayed in UI must use some template as preset

pmaier1:
> It might be possible to use the given message when sending emails, but the custom message will not be saved anywhere.

I don't think it needs to be saved somewhere.

> As I recall, there is a greater rework of the sharing due. Can we postpone changes until then?

This is not that much related to the other stuff except that we want to come away from the modal and integrate in the sharing sidebar.

Just thinking loud: When tackling this maybe we could enable editing the whole template to cover other use cases where people are unhappy with the templates as well. Generally inserting a note to the template would be preferable though, IMO. Editing templates should happen in the template editor.

PVince81:
@felixheidecke please ping @jvillafanez for backend work when needed

[...]

PR #29721, opened 2017-11-30
Public link share cleanup

lordelix:
## Description
- Reduce length of date field for external link
- Change button text 'Save' to 'Share'
- Change mail error text
- Add send confirmation after sending notification

## Related Issue
Fixes: https://github.com/owncloud/core/issues/29694

## How Has This Been Tested?
Visual tests in latest Chrome & Firefox

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests pass.

PR #29935, opened 2017-12-20
Share personal cleanup + note

### LINKED ISSUE #29694: Public link share UI cleanup
- [x] Adapt password and expiration date fields to a reasonable length
Currently these fields use the whole width of the modal. Just give them a reasonable length to make it look better (e.g. the date field always has the same width). For consistency I suggest to make the length of the two fields equal.

- [x] Rename the "Save" button to "Share"
This is to better reflect that the share and notifications effectively are created when the button is used.

- [x] Rename error message when invalid mail addresses are entered
=> "Could not send mail to the following **recipient(s)**"
This is to better differentiate between external recipients and internal **users**

- [x] Show "Sending"/"E-Mail sent" indicator longer and maybe turn green upon success
Currently the modal disappears in, say, 0,1 seconds after clicking the button so users have a hard time to see if the action was successful or not (I guess there is no "unsuccessful" but users should be able to take note of the success). Please increase the time the modal is shown after success to a reasonable value, e.g. 1-2 sec.

@felixheidecke @PVince81

lordelix:
## Description
- Reduce length of date field for external link
- Change button text 'Save' to 'Share'
- Change mail error text
- Add send confirmation after sending notification
- Add message field for email

## Related Issue
Replaces (because contains): https://github.com/owncloud/core/pull/29721
Fixes: https://github.com/owncloud/core/issues/29694

## How Has This Been Tested?
Visual tests in latest Chrome & Firefox

## Types of changes

## Checklist:

- [x] All new and existing tests pass.

### DIFF
--- a/core/ajax/share.php
+++ b/core/ajax/share.php
@@ -130,6 +130,7 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 				$defaults,
 				\OC::$server->getURLGenerator()
 			);
+
 			$result = $mailNotification->sendInternalShareMail($recipientList, $itemSource, $itemType);
 
 			// if we were able to send to at least one recipient, mark as sent
@@ -143,7 +144,7 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 			} else {
 				OCP\JSON::error([
 					'data' => [
-						'message' => $l->t("Couldn't send mail to following users: %s ",
+						'message' => $l->t("Couldn't send mail to following recipient(s): %s ",
 								implode(', ', $result)
 								)
 					]
@@ -160,6 +161,7 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 			break;
 
 		case 'email':
+
 			// read and filter post variables
 			$filter = new MailNotificationFilter([
 				'link' => $_POST['link'],
@@ -168,9 +170,26 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 				'expiration' => $_POST['expiration']
 			]);
 
+			// read post variables
+			$link = (string)$_POST['link'];
+			$file = (string)$_POST['file'];
+			$toAddress = (string)$_POST['toAddress'];
+			$options = array();
+			$emailBody = null;
+
+			if (isset($_POST['emailBody'])) {
+				$emailBody = trim((string)$_POST['emailBody']);
+			}
+
+			if (isset($_POST['bccSelf']) && $_POST['bccSelf'] === 'true') {
+				$options['bcc'] = \OC::$server->getUserSession()->getUser()->getEMailAddress();
+			}
+
+			$l10n = \OC::$server->getL10N('lib');
+
 			$mailNotification = new \OC\Share\MailNotifications(
 				\OC::$server->getUserSession()->getUser(),
-				\OC::$server->getL10N('lib'),
+				$l10n,
 				\OC::$server->getMailer(),
 				\OC::$server->getLogger(),
 				$defaults,
@@ -190,6 +209,16 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 			$result = $mailNotification->sendLinkShareMail(
 				$filter->getToAddress(), $filter->getFile(), $filter->getLink(), $expiration
 			);
+
+			$subject = (string)$l10n->t('%s shared »%s« with you', [$this->senderDisplayName, $filename]);
+			if ($emailBody === null || $emailBody === '') {
+				list($htmlBody, $textBody) = $mailNotification->createMailBody($file, $link, $expiration);
+			} else {
+				$htmlBody = null;
+				$textBody = strip_tags($emailBody);
+			}
+
+			$result = $mailNotification->sendLinkShareMailFromBody($toAddress, $subject, $htmlBody, $textBody, $options);
 
 			if(empty($result)) {
 				// Get the token from the link
@@ -217,7 +246,7 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 								->setAuthor($currentUser)
 								->setAffectedUser($currentUser)
 								->setObject('files', $fileId, $path)
-								->setSubject(\OCA\Files_Sharing\Activity::SUBJECT_SHARED_EMAIL, [$path, $to_address]);
+								->setSubject(\OCA\Files_Sharing\Activity::SUBJECT_SHARED_EMAIL, [$path, $toAddress]);
 							\OC::$server->getActivityManager()->publish($event);
 						}
 					}
@@ -228,7 +257,7 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 				$l = \OC::$server->getL10N('core');
 				OCP\JSON::error([
 					'data' => [
-						'message' => $l->t("Couldn't send mail to following users: %s ",
+						'message' => $l->t("Couldn't send mail to following recipient(s): %s ",
 								implode(', ', $result)
 							)
 					]
@@ -323,12 +352,12 @@ if (isset($_POST['action']) && isset($_POST['itemType']) && isset($_POST['itemSo
 				$sharedGroups = [];
 				if (isset($_GET['itemShares'])) {
 					if (isset($_GET['itemShares'][OCP\Share::SHARE_TYPE_USER]) &&
-					    is_array($_GET['itemShares'][OCP\Share::SHARE_TYPE_USER])) {
+						is_array($_GET['itemShares'][OCP\Share::SHARE_TYPE_USER])) {
 						$sharedUsers = $_GET['itemShares'][OCP\Share::SHARE_TYPE_USER];
 					}
 
 					if (isset($_GET['itemShares'][OCP\Share::SHARE_TYPE_GROUP]) &&
-					    is_array($_GET['itemShares'][OCP\Share::SHARE_TYPE_GROUP])) {
+						is_array($_GET['itemShares'][OCP\Share::SHARE_TYPE_GROUP])) {
 						$sharedGroups = $_GET['itemShares'][OCP\Share::SHARE_TYPE_GROUP];
 					}
 				}
--- a/core/css/global.css
+++ b/core/css/global.css
@@ -4,6 +4,15 @@
 
 /* Global Components */
 
+/* Positioning */
+
+.absolute-center {
+	position: absolute;
+	top: 50%;
+	left: 50%;
+	transform: translate(-50%, -50%);
+}
+
 .pull-left {
 	float: left;
 }
@@ -24,7 +33,7 @@
 	clear: both;
 }
 
-.hidden {
+.hidden.hidden {
 	display: none;
 }
 
@@ -41,10 +50,25 @@
 	font-weight:600;
 }
 
-.center {
-	text-align:center;
+.text-small {
+	font-size: 80%;
 }
 
 .inlineblock {
 	display: inline-block;
+}
+
+/* Text */
+
+.text-right {
+	text-align: right;
+}
+
+.center,
+.text-center {
+	text-align: center;
+}
+
+.text-left {
+	text-align: left;
 }
--- a/core/js/sharedialoglinkshareview.js
+++ b/core/js/sharedialoglinkshareview.js
@@ -14,7 +14,7 @@
 	}
 
 	var PASSWORD_PLACEHOLDER_STARS = '**********';
-	var PASSWORD_PLACEHOLDER_MESSAGE = t('core', 'Choose a password for the public link');
+	var PASSWORD_PLACEHOLDER_MESSAGE = t('core', 'Choose a password');
 	var TEMPLATE =
 		'<div class="error-message-global hidden"></div>' +
 		'<div class="public-link-modal">'+
@@ -194,11 +194,8 @@
 				success: function() {
 					if (self.mailView) {
 						// also send out email first
-						self.mailView.sendEmails().then(done).fail(function() {
-							done();
-							// re-show the popup
-							self.show();
-						});
+						// do not resolve on errors
+						self.mailView.sendEmails().then(done);
 					} else {
 						done();
 					}
@@ -243,7 +240,7 @@
 
 				publicUploadPossible       : this._isPublicUploadPossible(),
 
-				publicUploadLabel          : t('core', 'Upload only (File Drop)'),
+				publicUploadLabel          : t('core', 'Upload only') + ' (File Drop)',
 				publicUploadDescription    : t('core', 'Receive files from others without revealing the contents of the folder.'),
 				publicUploadValue          : OC.PERMISSION_CREATE,
 				publicUploadSelected       : this.model.get('permissions') === OC.PERMISSION_CREATE,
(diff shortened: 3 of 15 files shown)


PVince81:
Code looks good so far.

We talked about injecting the default template in there. I'll try exposing this somehow from the backend.

PVince81:
- [ ] TODO: expose default template
- [ ] TODO: send message body to server
- [ ] TODO: server must use message when sending

PVince81:
pushed: server now uses the specified message. If empty, it falls back to the default template.

Note that when specified it's a plain text email and it strips away the tags with `strip_tags`.

Next up: find a way to pass original template back for use in the dialog, in a way that clients could also reuse...

PVince81:
It looks like this would need a new API endpoint. Considering that we want clients to be able to use this email feature too, we'd likely need to add a new **public** endpoint, maybe an extension of the existing OCS Share API.

And if we need to do this, we might as well move the current emailing code (the sending part) to said endpoint as well. So the endpoint would have two functions:
1) get email template for given values (expiration, filename, etc)
and
2) send email with given body

PVince81:
I suggest merging this PR as is, maybe tweak the current message and work on the email endpoint separately, as I have the feeling this will take longer due to API design decisions needing to be made.

@pmaier1 thoughts ?

PVince81:
Raised https://github.com/owncloud/core/issues/29937 for exposing template.

PVince81:
- [ ] add CC option

@felixheidecke can you add ? I can adjust the backend then

pmaier1:
> I suggest merging this PR as is, maybe tweak the current message and work on the email endpoint separately, as I have the feeling this will take longer due to API design decisions needing to be made.
@pmaier1 thoughts ?

Yes, when the stuff in here https://github.com/owncloud/core/issues/29694 is covered and this is tested we should merge. I would be also very happy if you could provide a patch for this PR then.

PVince81:
Fixes https://github.com/owncloud/core/issues/29696

- [ ] Felix: CC must be a checkbox only that appearrs only if `OC.getCurrentUser().email` is set
- [ ] Vincent: export email address through OC.getCurrentUser()

PVince81:
- [ ] backend must actually set CC to current user's email address

PVince81:
bcc to self or cc ? @pmaier1 

bcc might be a better idea here

DeepDiver1975:
rebase needed because of conflicts .....

core/ajax/share.php
```diff
@@ -163,10 +164,16 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
 			$link = (string)$_POST['link'];
 			$file = (string)$_POST['file'];
 			$to_address = (string)$_POST['toaddress'];
+			$emailBody = null;
+			if (isset($_POST['emailBody'])) {
+				$emailBody = trim((string)$_POST['emailBody']);
+			}
+
+			$l10n = \OC::$server->getL10N('lib');
```
DeepDiver1975: why lib? this file lives in core - did this ever work?

core/css/global.css
```diff
@@ -24,7 +33,7 @@
 	clear: both;
 }
 
-.hidden {
+.hidden.hidden {
```
DeepDiver1975: duplicate hidden?

core/js/sharedialogmailview.js
```diff
@@ -34,120 +43,162 @@
...
+				return deferred.reject({
+					message: t('core', '{email} is not a valid address!', {email: mail.to})
+				});
 			}
 
+			var params = {
+				action      : 'email',
+				toAddress   : mail.to,
+				emailBody   : mail.body,
+                bccSelf     : mail.bccSelf,
```
DeepDiver1975: indent

DeepDiver1975 (dismissed):

felixheidecke:
@PVince81 pls help with conflicts in `share.php`

pmaier1:
> bcc to self or cc ? @pmaier1
bcc might be a better idea here

Hmm, good question. Both have advantages/drawbacks. I think the point here is that recipients can 
1. answer on the mail notification easily and reach the sender thereby and that
2. you can forward the mail or quote at a later point in time ("see here, you got this 3 weeks ago but you didn't respond" :) )

I think CC is more suitable.

core/css/global.css
felixheidecke: Yes. This a lesser version of the `!important` statement;

core/ajax/share.php
felixheidecke: @PVince81 yours?

felixheidecke:
> answer on the mail notification easily and reach the sender thereby and that

@pmaier1 I had the idea, that we could insert the senders' email address as the reply-to address to those notifications. So the recipient could always easily respond even if the sender didn't CC himself. What do you think?

felixheidecke:
I need some help with [this](https://github.com/owncloud/core/blob/share-personal-note/core/ajax/share.php#L167). I can't debug PHP code. The JSON key says false for the bccSelf key but apparently, PHP doesn't see it as false.

pmaier1:
> @pmaier1 I had the idea, that we could insert the senders' email address as the reply-to address to those notifications. So the recipient could always easily respond even if the sender didn't CC himself. What do you think?

Nice idea basically but I think that there are different views on this. Some people might not want to include their mail address when sending a public link. It would need to be configurable therefore. For the CC feature it could set "Reply-to" automatically as the user intentionally decides to include their mail address.

felixheidecke:
> […] Some people might not want to include their mail address when sending a public link […]

This could also be true for receiving a copy of the email. I think it's none of the recipients business if the did or not. We don't convey the info, that the recipient can see that you did. I guess this should be more clear. Any other services that I use that offer the _send copy to self_ always send a silent bcc copy which I really like.

PVince81:
Rebased and solved conflicts

core/ajax/share.php
```diff
@@ -191,6 +207,16 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
 				$filter->getToAddress(), $filter->getFile(), $filter->getLink(), $expiration
 			);
 
+			$subject = (string)$l10n->t('%s shared »%s« with you', [$this->senderDisplayName, $filename]);
```
PVince81: I'm a bit surprised that we now suddenly have a hard-coded subject here. Was this subject always empty in the past ? Does it not come from the template if not set ?

core/ajax/share.php
PVince81: Maybe this never worked... I just moved up the getL10N call from below.

Considering that emails are usually be taken from the mail template, I don't think the template is translated.

Needs separate testing.

core/ajax/share.php
```diff
@@ -168,9 +170,23 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
 				'expiration' => $_POST['expiration']
 			]);
 
+			// read post variables
+			$link       = (string)$_POST['link'];
+			$file       = (string)$_POST['file'];
+			$to_address = (string)$_POST['toAddress'];
+			$to_bcc     = ((string)$_POST['bccSelf']) ? array('bcc' => \OC::$server->getUserSession()->getUser()->getEMailAddress()) : false;
```
PVince81: this line is too long and confusing.

please add a `if` statement to check if the user's email address is set and only set `$toBcc` (camel case please!) if the email is not set.

Reading this line seems to set `array('bcc' => null)` if the email is null.

PVince81 (commented):
See comments above

core/ajax/share.php
felixheidecke: Blame says you wrote this ["a month ago"](https://github.com/owncloud/core/commit/e97ba15e4916aabc32372dc9fa4981b379739f6a#diff-e574f71b1f320c19e9ffb5c5322146deR208) … but it's not working 🌵

PVince81: Ok, I need to consult with my past self...

core/ajax/share.php
```diff
@@ -168,9 +170,26 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
...
+			$file = (string)$_POST['file'];
+			$toAddress = (string)$_POST['toAddress'];
+			$options = array();
+			$emailBody = null;
+
+			if (isset($_POST['emailBody'])) {
+				$emailBody = trim((string)$_POST['emailBody']);
+			}
+
+			if (isset($_POST['bccSelf'])) {
```
felixheidecke: I need help with checking if this variable is true/false. Not sure how PHP handles this … bool/string/int whatever. `isset()` is surely the false approach.

PVince81: `isset($_POST['bccSelf']) && $_POST['bccSelf'] === 'true'` (the string true because that's what comes through HTTP.

felixheidecke: Nice … works perfectly! 👍

pmaier1:
> Any other services that I use that offer the send copy to self always send a silent bcc copy which I really like.

Fine. Then I agree. The basic requirement anyway just was to be able to receive a copy in CC. We could state clearly that using the option will reveal the mail address to the recipient.

PVince81:
Note that there is a third possibility: sending a completely separate email to the sender with a text at the top "this is the message that was sent to XYZ as part of share ABC". But this is more complex and doesn't bring as many benefits as the other solutions.

PVince81:
something to look into soon: make this dialog scrollable ? it barely fits vertically on my 1080 lines screen:
image

PVince81:
(and my zoom factor in Chromium is already at 90%)

PVince81:
@felixheidecke I checked out this branch and can't find the CC/BCC checkbox. Where did it go ?

felixheidecke:
> I checked out this branch and can't find the CC/BCC checkbox. Where did it go ?

@PVince81 do you have an e-mail address in you user account?

PVince81:
I've set an email but somehow it doesn't appear in `OC.getCurrentUser()`, weird... I'll debug this

PVince81:
Okay found it... I have the enterprise theme enabled and the way I'm passing the email is no good. Basically we are passing some vars by making them an attribute of the `head` tag in the base layout. This sucks because it means every theme would have to have said vars there as well...

Good thing I had this theme enabled else we might never have caught this that early...

I'll move the logic to "js/config.php" instead.

PVince81:
- [ ] disable browser autocomplete of previously typed values on the email field

PVince81:
crap, I broke stuff... apparently there is still code that relies on the deprecated `OC.currentUser`... I'll bring it back...

PVince81:
@felixheidecke not strong enough:
image

felixheidecke:
> Even without a master password, in-browser password management is generally seen as a net gain for security. Since users do not have to remember passwords that the browser stores for them, they are able to choose stronger passwords than they would otherwise.

> For this reason, many modern browsers do not support autocomplete="off" for login fields:

There might not be a way :-|

PVince81:
ok, works now

PVince81:
@felixheidecke next challenge, fix the PHP unit tests:

```
18:25:25 1) Test\Share\MailNotificationsTest::testSendLinkShareMailWithoutReplyTo
18:25:25 Undefined index: cc
18:25:25 
18:25:25 /var/lib/jenkins/workspace/owncloud-core_core_PR-29935-54X6KPSPNBLFEE5QKGUIM3N3UKPXGMMI5Z6VT2XB3XWRL6GZBXIQ/lib/private/Share/MailNotifications.php:192
18:25:25 /var/lib/jenkins/workspace/owncloud-core_core_PR-29935-54X6KPSPNBLFEE5QKGUIM3N3UKPXGMMI5Z6VT2XB3XWRL6GZBXIQ/lib/private/Share/MailNotifications.php:176
18:25:25 /var/lib/jenkins/workspace/owncloud-core_core_PR-29935-54X6KPSPNBLFEE5QKGUIM3N3UKPXGMMI5Z6VT2XB3XWRL6GZBXIQ/tests/lib/Share/MailNotificationsTest.php:131
18:25:25 
18:25:25 2) Test\Share\MailNotificationsTest::testSendLinkShareMailWithReplyTo with data set #0 ('lukas@owncloud.com', array('lukas@owncloud.com'))
18:25:25 Undefined index: cc
18:25:25 
18:25:25 /var/lib/jenkins/workspace/owncloud-core_core_PR-29935-54X6KPSPNBLFEE5QKGUIM3N3UKPXGMMI5Z6VT2XB3XWRL6GZBXIQ/lib/private/Share/MailNotifications.php:192
18:25:25 /var/lib/jenkins/workspace/owncloud-core_core_PR-29935-54X6KPSPNBLFEE5QKGUIM3N3UKPXGMMI5Z6VT2XB3XWRL6GZBXIQ/lib/private/Share/MailNotifications.php:176
18:25:25 /var/lib/jenkins/workspace/owncloud-core_core_PR-29935-54X6KPSPNBLFEE5QKGUIM3N3UKPXGMMI5Z6VT2XB3XWRL6GZBXIQ/tests/lib/Share/MailNotificationsTest.php:195
```

You can run them with `make test-php TEST_PHP_SUITE=tests/lib/Share/MailNotificationsTest.php`

felixheidecke:
Please start the reviewing @PVince81 @DeepDiver1975 :-)

lib/private/Share/MailNotifications.php
```diff
@@ -189,8 +189,8 @@ public function sendLinkShareMail($recipient, $filename, $link, $expiration) {
 	public function sendLinkShareMailFromBody($recipient, $subject, $htmlBody, $textBody, $options = array()) {
 
 		$recipients    = $this->_mailStringToArray($recipient);
-		$ccRecipients  = ($options['cc']) ? $this->_mailStringToArray($options['cc']) : null;
-		$bccRecipients = ($options['bcc']) ? $this->_mailStringToArray($options['bcc']) : null;
+		$ccRecipients  = (!empty($options['cc'])) ? $this->_mailStringToArray($options['cc']) : null;
```
PVince81: don't use `empty` for strings because PHP evaluates `empty("0")` to true....

if these values are always set then use `$options['cc'] !== ''` instead.

core/js/sharedialoglinkshareview.js
```diff
@@ -231,7 +228,7 @@
 
 				publicUploadPossible       : this._isPublicUploadPossible(),
 
-				publicUploadLabel          : t('core', 'Upload only (File Drop)'),
+				publicUploadLabel          : t('core', 'Upload only') + ' (File Drop)',
```
PVince81: any reason why you pulled "File Drop" out of the translation ?

PVince81: cc @pmaier1

lib/private/Share/MailNotifications.php
felixheidecke: They are not always set :-| Hope my fix is valid now :)

core/js/sharedialoglinkshareview.js
felixheidecke: There was an awful translation and we want to keep its original name (File Drop) across languages.

PVince81 (approved):
:+1:

IljaN:
Rebase error?

felixheidecke:
> Rebase error?

Never 😇

PVince81:
@felixheidecke did it work our with the JS test ?

felixheidecke:
> did it work out with the JS test?

Sure did. The test had to fail since it tested a feature that was scaped. I should have guessed that :-P

PVince81:
@felixheidecke please backport to stable10

PVince81:
regression: https://github.com/owncloud/core/issues/30557

PVince81:
stable10: https://github.com/owncloud/core/pull/30486

SergioBertolinSG:
When sending an email it should fail when it is unable to send the email, specially if the email address is not complete:

email_verification

SergioBertolinSG:
https://github.com/owncloud/core/pull/29935#issuecomment-381095714 is fixed in 10.0.8RC2.

PR #29933, opened 2017-12-20
UI tests to test creating users with less than 3 characters

individual-it:
## Description
usernames with less than 3 char are not allowed, here are tests to test that feature

## Motivation and Context
the UI should not allow to create users with usernames that are shorter than 3 characters

## How Has This Been Tested?
run tests in travis.
The UI tests will not pass in this PR because it comes form an other fork, but they all passed here: individual-it#60
Not doing a separate branch here because travis has currently a huge backlog

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #29952, opened 2017-12-21
[stable10] user with less than 3 characters

phil-davis:
Backport #29933


individual-it (approved):

individual-it:
let make the other backports waiting for the same file when CI is finished here

phil-davis:
Yes, they need to wait. Is is the one that was merged first in master. When this is in stable10, then the next one will backport without conflict...

PR #29981, opened 2017-12-27
Dont reset displayname to uid on sso login

tomneedham:
## Description
If you login with SSO your displayname is set to your uid. Metadata sync is handled at the moment via shibboleth but will eventually be handled by the sync service on login. This is not necessary.

## Motivation and Context
Breaks sso dispalynames

## How Has This Been Tested?
 - setup ldap user backend with a server (use a custom internal user id)
 - use a custom displayname attribute
 - sync the users
 - displaynamse are correct
 - setup shibboleth login
 - login using sso 
 - see that dispalyname is still correct and is no longer converted to your uid

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/legacy/user.php
+++ b/lib/private/legacy/user.php
@@ -186,7 +186,6 @@ class OC_User {
 		if ($uid) {
 			if (self::getUser() !== $uid) {
 				self::setUserId($uid);
-				self::setDisplayName($uid);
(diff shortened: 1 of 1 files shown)


butonic (approved):
yes, makes sense. if no displayname is set the fallback is the userid anyway

DeepDiver1975 (approved):

Issue #29990, opened 2017-12-29
cannot add user to group called "0" by using provisioning API


individual-it:
### Steps to reproduce
1. create a group called "0" (by now that should work by any means)
2. add a user to that group using the provisioning API. e.g. `curl -u admin:admin http://localhost/owncloud-core/ocs/v2.php/cloud/users/admin/groups -d groupid="0"`

### Expected behaviour
user should be added to that group

### Actual behaviour
error 400 is returned 
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>400</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

### Server configuration
**Operating system**:
Ubuntu 17.10
**Web server:**
Apache2
**Database:**
SQlite
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
10.0.4
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "ocjvys04zaqc",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "192.168.0.111"
        ],
        "datadirectory": "\/home\/JankariTech\/www\/html\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/192.168.0.111\/owncloud-core",
        "dbtype": "sqlite3",
        "version": "10.0.4.4",
        "logtimezone": "UTC",
        "installed": true,
        "default_language": "en"
    }
}
```

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - dav: 0.3.2
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_sharing: 0.10.1
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - testing: 0.1.0
  - updatenotification: 0.2.1
Disabled:
  - theme-example
```

**Are you using external storage, if yes which one:** local/smb/sftp/...
local
**Are you using encryption:** yes/no
no
**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
no

#### Web server error log
nothing

[...]

PR #29994, opened 2017-12-29
Fix add/delete user to group 0 via provisioning API

phil-davis:
## Description
1) Add scenarios for adding users to group 0.
2) Add scenarios for deleting users from group 0.
Those scenarios fail, so fix the code:
3) When checking the group id, do not test with the ``empty()`` function - test explicitly for the things that are "empty" apart from ``0``.

## Related Issue
#29990 

## Motivation and Context
Remove more uses of ``empty()`` that cause problems for data that has the valid value "0".

## How Has This Been Tested?
Run the API tests locally.

## Types of changes

## Checklist:

### DIFF
--- a/apps/provisioning_api/lib/Users.php
+++ b/apps/provisioning_api/lib/Users.php
@@ -475,8 +475,8 @@ class Users {
 			return new Result(null, API::RESPOND_UNAUTHORISED);
 		}
 
-		$groupId = !empty($_POST['groupid']) ? $_POST['groupid'] : null;
-		if($groupId === null) {
+		$groupId = isset($_POST['groupid']) ? $_POST['groupid'] : null;
+		if (($groupId === '') || ($groupId === null) || ($groupId === false)) {
 			return new Result(null, 101);
 		}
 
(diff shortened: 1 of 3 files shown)


apps/provisioning_api/lib/Users.php
```diff
@@ -475,8 +475,8 @@ public function addToGroup($parameters) {
 			return new Result(null, API::RESPOND_UNAUTHORISED);
 		}
 
-		$groupId = !empty($_POST['groupid']) ? $_POST['groupid'] : null;
-		if($groupId === null) {
+		$groupId = $_POST['groupid'];
+		if (($groupId === '') || ($groupId === null) || ($groupId === false)) {
```
phil-davis: Note: this uses the same style of check that was done in Groups.php ``addGroup`` a while ago.

phil-davis:
Note: the 2nd commit is needed because the test scenarios are not independent. The scenarios do not seem to clean up their state (in this case which users are in which groups) at the end of each scenario.

As a separate issue, that should all be sorted out some day.

individual-it:
Looks good, should we add tests that test really empty group names?

phil-davis:
This scenario seems to send an empty groupid:
https://github.com/owncloud/core/blob/a3443144274101d1f0944de1f4b5bf1e252380b7/tests/integration/features/provisioning-v1.feature#L93

individual-it (approved):

phil-davis:
Backport stable10 #30004

PR #30000, opened 2017-12-31
accountMapper either returns an object or throws an exception

butonic:
the AccountManager getByUid cannot return null

thx @mrow4a for pointing this out

### DIFF
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -185,12 +185,9 @@ class Manager extends PublicEmitter implements IUserManager {
 		}
 		try {
 			$account = $this->accountMapper->getByUid($uid);
-			if (\is_null($account)) {
-				$this->cachedUsers->set($uid, null);
-				return null;
-			}
 			return $this->getUserObject($account);
 		} catch (DoesNotExistException $ex) {
+			$this->cachedUsers->set($uid, null);
 			return null;
 		}
 	}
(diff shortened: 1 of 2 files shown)


lib/private/User/Manager.php
```diff
@@ -161,12 +161,9 @@ public function get($uid) {
 		}
 		try {
 			$account = $this->accountMapper->getByUid($uid);
-			if (is_null($account)) {
-				$this->cachedUsers->set($uid, null);
-				return null;
-			}
 			return $this->getUserObject($account);
 		} catch (DoesNotExistException $ex) {
+			$this->cachedUsers->set($uid, null);
```
mrow4a: Hmm, do we have some unit test checking if we don't make an unnecessarily call? Via e.g. checking number of calls for mocked object? It might prevent some silly errors later.

butonic: nope ... AFAICT this line is untested. care to implement the test?

mrow4a: I am anyways touching that fragment of code in my PR with unit tests, so I can include it there - in fact it is how I found it out ;d If we merge it quick I will cover it.

DeepDiver1975 (changes requested):
fix tests please and make ci happy

PVince81:
any update ?

PVince81:
@butonic please backport

phil-davis:
Backport ``stable10`` #31445

PR #30003, opened 2018-01-02
Allow user or group exists tests to fail

phil-davis:
## Description
1) Use existing step definitions that check if a user or group already exists (those steps do not first create the user or group). Do this in places where the step is a "then" test. That gives the chance that the test will fail.
2) Remember the user or group that was expected to exist, so that the ``AfterScenario`` will understand that the user or group needs to be removed at the end of the scenario.

## Motivation and Context
Steps like ``And user "brand-new-user" exists`` work in the ``Given`` section - they create the user (or group). They do not fail if the user (or group) does not exist. In the ``Then`` section we want steps that actually fail if the user does not exist.

## How Has This Been Tested?
Run integration tests locally.

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/Provisioning.php
+++ b/tests/integration/features/bootstrap/Provisioning.php
@@ -39,6 +39,7 @@ trait Provisioning {
 	 */
 	public function userAlreadyExists($user) {
 		PHPUnit_Framework_Assert::assertTrue($this->userExists($user));
+		$this->rememberTheUser($user);
 	}
 
 	/**
@@ -55,6 +56,7 @@ trait Provisioning {
 	 */
 	public function groupAlreadyExists($group) {
 		PHPUnit_Framework_Assert::assertTrue($this->groupExists($group));
+		$this->rememberTheGroup($group);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


phil-davis:
I noticed this when adding tests for group ``0`` - steps that did not fail even if the group did not exist.

individual-it (approved):
don't like the wording
`Then user "brand-new-user" already exists` should be `Then user "brand-new-user" should exists`
but that is for an other PR

phil-davis:
Yes, I used some existing step definitions here. At the moment there are steps with various mixes of ``assure something`` ``check something`` ``something exists`` ``something already exists`` etc. When using them you have to know which create the "something" and which check that the "something" already exists.

There is an issue somewhere where we started discussing this. We need to agree on a standard wording system then refactor to it.

phil-davis:
Backport stable10 #30005

individual-it:
discussion about the wording is here: https://github.com/owncloud/guests/issues/162

tests/integration/features/bootstrap/Provisioning.php
```diff
@@ -204,7 +210,7 @@ public function userBelongsToGroup($user, $group) {
 		$this->response = $client->get($fullUrl, $options);
 		$respondedArray = $this->getArrayOfGroupsResponded($this->response);
 
-		if (array_key_exists($group, $respondedArray)) {
+		if (in_array($group, $respondedArray)) {
```
phil-davis: The existing group names are the values in the array.
The problem here was noticed when trying to do ``userBelongsToGroup("user0", "0")`` - checking if ``user0`` is in a group called ``0`` - it would think the user was in the group when it was not.

phil-davis:
I added another commit to fix more issues for tests that failed because the scenarios are more effectively independent - needed to setup more state that was missing in some scenarios.

PR #30008, opened 2018-01-02
Fix failure of shares which are already moved with transfer-ownership

sharidas:
If the folder and files inside the folder are already moved
by the command, and if there is a public link inside the folder
then during the restore share operation the command throws
exception. This fix tries to check if the file/folder is already
moved. If so then there is no need to do call updateShare.

## Description

When there are multiple shares during the transfer-ownership command, and if the share is already moved to the destination user, then the command would throw an exception if the share inside the folder is handled separately.

## Motivation and Context

Lets say we have `user1` and `user2`. `user1` creates folder `foo` and shares with `user2`. Then `user2` tries to creates a public link of one of the files inside `foo`. Now run the transfer-ownership command with source user as `user1` and destination user as `user2`. An exception is thrown. The reason I found is if the folder is already moved, and in the restore share checks for the share of public share and run the updateShare causes exception. The file is not available. So this change checks if the file/folder is already there in the destination folder. If so then nothing to do.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -302,6 +302,18 @@ class TransferOwnership extends Command {
 					if ($share->getSharedBy() === $this->sourceUser) {
 						$share->setSharedBy($this->destinationUser);
 					}
+					/*
+					 * If the share is already moved then updateShare would cause exception
+					 * This can happen if the folder is shared and file(s) inside the folder
+					 * has shares, for example public link
+					 */
+					if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+						$sharePath = ltrim($share->getNode()->getPath(), '/');
+						if (strpos($sharePath, $this->finalTarget) !== false) {
+							//The share is already moved
+							continue;
+						}
+					}
 
 					$this->shareManager->updateShare($share);
 				}
--- a/tests/integration/features/transfer-ownership.feature
+++ b/tests/integration/features/transfer-ownership.feature
@@ -165,6 +165,23 @@ Feature: transfer-ownership
 		Then downloaded content when downloading file "/somefile.txt" with range "bytes=0-6" should be "This is"
 
 	@no_default_encryption
+	Scenario: transferring ownership of folder shares which has public link
+		Given user "user0" exists
+		And user "user1" exists
+		And user "user2" exists
+		And user "user0" created a folder "/test"
+		And user "user0" uploads file "data/textfile.txt" to "/test/somefile.txt"
+		And folder "/test" of user "user0" is shared with user "user2" with permissions 31
+		And as an "user1"
+		And creating a share with
+			| path | /test/somefile.txt |
+			| shareType | 3 |
+		When transferring ownership of path "test" from "user0" to "user1"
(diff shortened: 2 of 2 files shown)


apps/files/lib/Command/TransferOwnership.php
```diff
@@ -302,6 +302,16 @@ private function restoreShares(OutputInterface $output) {
 					if ($share->getSharedBy() === $this->sourceUser) {
 						$share->setSharedBy($this->destinationUser);
 					}
+					/*
+					 * If the share is already moved then updateShare would cause exception
+					 * This can happen if the folder is shared and file(s) inside the folder
+					 * has shares, for example public link
+					 */
+					if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+						if (strpos($share->getNode()->getPath(), $this->finalTarget) === 1) {
```
PVince81: `=== 1` sounds dangerous here. Can we ltrim the slashes before doing the check ? Result might be the same but at least more readable for developers and less dangerous-looking.

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -302,6 +302,16 @@ private function restoreShares(OutputInterface $output) {
 					if ($share->getSharedBy() === $this->sourceUser) {
 						$share->setSharedBy($this->destinationUser);
 					}
+					/*
+					 * If the share is already moved then updateShare would cause exception
+					 * This can happen if the folder is shared and file(s) inside the folder
+					 * has shares, for example public link
+					 */
+					if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
```
PVince81: can you check whether outgoing federated shares would also be affected by this ? (`SHARE_TYPE_REMOTE`)

sharidas: Below are the steps done to verify the federated share:
- [x] Create 2 users `user1` and `user2`. `user1` creates folder `foo` which is shared with `user2`.
- [x] `user2` reshares `foo` using federated share
- [x] run occ transfer-ownership command and it works without any exception
- [x] Verified in `user2` the folder `foo` is available fedshared.

- [x] Create 2 users `user1` and `user2`. `user1` creates folder `foo` which is shared with `user2`.
- [x] `user1` fedshare folder `foo`
- [x] run occ transfer-ownership command and it works without any exception
- [x] Verified in `user2` the folder `foo` is available fedshared.

apps/files/lib/Command/TransferOwnership.php
sharidas: Updated the PR. The new PR does the ltrim check before the check. Removed the dangerorus-looking `===1` with `!== false`

sharidas:

PVince81 (approved):
:+1:

PVince81:
please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/30161

PVince81:
Any chance to make codecov happy ?

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

Issue #30017, opened 2018-01-05
Recursive loop when skeleton folder has no read access


phil-davis:
### Steps to reproduce
1. Create a folder for skeleton files and put some files in it.
2. Set the protection on the folder so that the web server does not have read access.
3. make it the ``skeletondirectory`` in the config.
3. Create a new user.
4. Try to login as the new user.

### Expected behaviour
a) The new user logs in OK and has no initial "skeleton" files (because the skeleton folder cannot be read)
b) ``owncloud.log`` contains some useful message to say what the problem is.

### Actual behaviour
PHP dives into a recursive loop in ``lib/private/legacy/util.php`` ``copyr`` calling itself until it reaches "Maximum function nesting level".

### Server configuration
**Operating system**:
Ubuntu 17.10

**Web server:**
PHP dev server

**Database:**
mySQL

**PHP version:**
5.6.32

**ownCloud version:** (see ownCloud admin page)
10.0.4

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

phil-davis:
Fixed by #30018

PR #30018, opened 2018-01-05
Handle no read access to skeleton

phil-davis:
## Related Issue
#30017 

## How Has This Been Tested?
Manually make the skeleton top-level folder have no read access.
Manually make a file in the skeleton folder have no read access.
Check that in each case a new user can login, and ``owncloud.log`` has a message:
``Skeleton not created due to missing read permission in skeleton directory``

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -56,6 +56,7 @@
  *
  */
 
+use OCP\Files\NoReadAccessException;
 use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IUser;
@@ -416,17 +417,25 @@ class OC_Util {
 	 * @param string $source
 	 * @param \OCP\Files\Folder $target
 	 * @return void
+	 * @throws NoReadAccessException
 	 */
 	public static function copyr($source, \OCP\Files\Folder $target) {
 		$dir = opendir($source);
+		if (false === $dir) {
+			throw new NoReadAccessException('No read permission for folder ' . $source);
+		}
 		while (false !== ($file = readdir($dir))) {
 			if (!\OC\Files\Filesystem::isIgnoredDir($file)) {
 				if (is_dir($source . '/' . $file)) {
 					$child = $target->newFolder($file);
 					self::copyr($source . '/' . $file, $child);
 				} else {
+					$sourceFileHandle = fopen($source . '/' . $file,'r');
+					if (false === $sourceFileHandle) {
+						throw new NoReadAccessException('No read permission for file ' . $file);
+					}
 					$child = $target->newFile($file);
-					stream_copy_to_stream(fopen($source . '/' . $file,'r'), $child->fopen('w'));
+					stream_copy_to_stream($sourceFileHandle, $child->fopen('w'));
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


lib/private/legacy/util.php
```diff
@@ -416,17 +417,25 @@ public static function copySkeleton($userId, \OCP\Files\Folder $userDirectory) {
 	 * @param string $source
 	 * @param \OCP\Files\Folder $target
 	 * @return void
+	 * @throws NoReadAccessException
 	 */
 	public static function copyr($source, \OCP\Files\Folder $target) {
-		$dir = opendir($source);
+		$dir = @opendir($source);
```
phil-davis: I had to suppress the potential "no permission" PHP error here, because the unit tests would trip over.
In any case, an exception is thrown below if something went wrong.

phil-davis:
Drone fails the new PHPunit tests. It looks like the ``chmod(file, 0)`` is not working to make the file unreadable to the process. The permissions thing works when run locally for me. Maybe somehow in the drone environment, the user is the file owner (like locally) and even with 0 permissions it still gets access to the file???

phil-davis:
But Jenkins passes - and the log has 1000's of PHPunit tests that pass - I assume the new tests here are in there.

PVince81 (approved):
:+1:

please backport

PVince81:
uh oh, seems to break unit tests ?!

```
PHP Fatal error: Call to a member function fopen() on null in /drone/src/lib/private/legacy/util.php on line 438
```

See https://drone.owncloud.com/owncloud/core/2252/17

phil-davis:
Yes, the tests run locally for me. They create real skeleton files and set their protection so that "nobody" has any access, then try to use them as the skeleton. It is not working in the drone environment - I guess the temporary skeleton files are owned by the process that is running the tests and even though the protection is set to no access, read succeeds because it is the file owner reading it???
I need to work out what to do...

DeepDiver1975:
drone is running in the context of the user root - the behavior is slightly different

tests/lib/UtilTest.php
```diff
@@ -417,11 +417,43 @@ public function testCheckDataDirectoryValidity() {
...
+	}
+
+	/**
+	 * @expectedException \OCP\Files\NoReadAccessException
+	 * @expectedExceptionMessage No read permission for folder
+	 */
+	public function testCopySkeletonDirectoryNoReadAccess() {
+		$skeletonDir = \OCP\Files::tmpFolder();
+		touch($skeletonDir . '/a-file');
+		chmod($skeletonDir, 0);
```
DeepDiver1975: this will not work on drone - you can skip the unit test like this:

```php
		if ($this->getCurrentUser() === 'root') {
			$this->markTestSkipped('You are running tests as root - this test will not work in this case.');
		}

phil-davis:
@DeepDiver1975 please review again. Tests skipped for root as suggested.

phil-davis:
@DeepDiver1975 or @PVince81 ping.
Is this OK to merge now?

phil-davis:
Backport stable10 #30241

Issue #29989, opened 2017-12-29
Sharing with a group name containg '&' symbol shows &amp; in sharing autocompletion


jaggu4444:
### Steps to reproduce
1. Create a group with the name containing '&' symbol.
2. Try to share any file with the group.
3. While you type the group name, then after each '&', automatically 'amp;' appears.

### Expected behaviour
Acutal group name should appear while entering

### Actual behaviour
After the every '&' symbol, automatically 'amp;' appears.
![image](https://user-images.githubusercontent.com/34025467/34433584-7b806868-eca8-11e7-9ee0-cde7c870cb18.png)

### Server configuration
**Operating system**: Ubuntu desktop 17.10

**Web server:** Apache2

**Database:**  SqLite3

**PHP version:** PHP 7.1.11

**ownCloud version:** ownCloud 10.0.4

**Updated from an older ownCloud or fresh install:** fresh install

**Where did you install ownCloud from:** git

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oci6h1gtmtlj",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "localhost"
        ],
        "datadirectory": "\/home\/bijay\/www\/html\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/localhost\/owncloud-core",
        "dbtype": "sqlite3",
        "version": "10.0.4.4",
        "logtimezone": "UTC",
        "default_language": "en",
        "installed": true,
        "csrf.disabled": true
    }
}

```

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - dav: 0.3.2
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_sharing: 0.10.1
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - testing: 0.1.0
  - updatenotification: 0.2.1
Disabled:
  - theme-example

```

**Are you using external storage, if yes which one:** local

**Are you using encryption:** No

### Client configuration
**Browser:** Chrome and Firefox

**Operating system:** Ubuntu desktop 17.10

#### Browser log
```
[...]

PR #30074, opened 2018-01-09
Preventing Handlebars from escaping expressions. #29989

ajainuary:
## Description

By default with {{expressions}}, Handlebars.js HTML-escapes values returned (& -> &amp;). By replacing the double-parrenthesis with triple-parenthesis {{{expression}}} this can be prevented.(& is preserved)

## Related Issue

https://github.com/owncloud/core/issues/29989
## Motivation and Context

It prevents the appearing of HTML-escape values for special characters like &,<,>, etc. when used in Group names in the sharing auto-complete menu.
## How Has This Been Tested?

Steps followed as mentioned in https://github.com/owncloud/core/issues/29989 result in the proper rendering of group-names as displayed in the screenshot.
image

## Types of changes

## Checklist:

### DIFF
--- a/core/js/sharedialogview.js
+++ b/core/js/sharedialogview.js
@@ -52,7 +52,7 @@
 					'<div class="avatardiv"></div>' +
 					'{{/if}}' +
 					'<div class="autocomplete-item-text">' +
-						'<span class="autocomplete-item-displayname">{{displayName}}</span>' +
+						'<span class="autocomplete-item-displayname">{{{displayName}}}</span>' +
 						'{{#if additionalInfo}}' +
 						'<span class="autocomplete-item-additional-info">({{additionalInfo}})</span>' +
 						'{{/if}}' +


PVince81:
Thanks a lot! Can you double check that display names are already escaped on another level ? Not escaping on handlebars level could cause a risk of XSS injection if a group name somehow contains something like a script tag.

ajainuary:
@PVince81 can you please point out at what level the string should be escaped?

PVince81:
@ajainuary I'd expect it to be escaped one level higher: not on handlebars level but the code that sets the value to the handlebars template.

ajainuary:
@PVince81  sorry for the delay, I was a bit caught up with with my university exams. I am new to this, so I cannot understand how would this create an Cross Site Scripting bug as when I tried the characters were properly escaped.

sharidas:
Ok, so looking at issue https://github.com/owncloud/core/issues/29989 for which this PR was created, did a basic testing ( without this patch ) and here is the result 
group_name_with_ampersand 

The testing I did was for group name with `&`. Lets say even if the group name has other special characters like `$#@` and created group `a$#@char` and  here is the result 
specialchars

We have fixed issues like escape urlencoding. Hence I am getting a feel that we can close this PR. Let me know if we agree on this.

PVince81:
thanks for verifying, looks fixed then

PR #30091, opened 2018-01-10
Picky Makefile being picky

butonic:
without this PR:
```
Makefile:39: *** Rezept beginnt vor dem ersten Ziel.  Schluss.
```

with this PR:
```
Makefile:38: *** yarn is not available on your system, please install yarn (npm install -g yarn).  Schluss.
```
see https://stackoverflow.com/a/30775546

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ SHELL=/bin/bash
 #
 YARN := $(shell command -v yarn 2> /dev/null)
 ifndef YARN
-	$(error yarn is not available on your system, please install yarn (npm install -g yarn))
+  $(error yarn is not available on your system, please install yarn (npm install -g yarn))
 endif
 
 KARMA=$(NODE_PREFIX)/node_modules/.bin/karma


butonic:
hmmm no this does not fix it as expected

PVince81:
Weird. I did see this message happening on the last build of the daily build.
However when I ran the build script manually the error didn't appear any more.

PVince81:
@butonic I just switched computers and my other computer which doesn't have yarn also throws the same error. Your PR fixes it, so 👍

PVince81 (approved):
:+1:

DeepDiver1975:
can we please squash this? THX

butonic:
@DeepDiver1975 what is there to backport? yarn is not in stable10

DeepDiver1975:
but the tab is in there as well ....

https://github.com/owncloud/core/blob/4bf4fc80b8dd2330ae8b783d7a1557232bc1d032/Makefile#L39

PVince81:
For some reason now the daily build fails because of this. Maybe cron is using a different shell because when I run the release script from bash it works fine.

Please backport... well, replace the tab with spaces on stable10

DeepDiver1975:
stable10 has 4 spaces in this place .... no backport needed

PVince81:
Ah right. I misread the daily failure email. It was actually saying "Jenkins back to normal for stable10", probably after the yarn revert. Good!

PR #29951, opened 2017-12-21
Fix trashbin clean

EvgeniySpinov:
## Description
Trashbin files are not cleared when FS is 100% full.

## Related Issue
Fixing issue #27412. 

## Motivation and Context
It cause system to stop functioning and was critical for my environment.

## How Has This Been Tested?
Manual testing has been performed with tracing changes in logs, while trashbin was cleared. No other testing is done.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Unfortunately I cannot run unit tests, due to environment limitations ( I simply have outdated environment for that ). So I'll leave it for Travis. In case of issues I'll try to sort something out with environment.

Issue #27412, opened 2017-03-17
Trashbin is not cleared with 100% full FS


EvgeniySpinov:
### Steps to reproduce
1. Create a OC folder mounted to a different disk, than /
2. Fill it with user data
3. Set trashbin_retention_obligation and versions_retention_obligation to 'auto, 180'
4. Make sure you have no empty space in OC folder
5. Add/update 1 file into the cloud.

### Expected behaviour
Trashbin should start to expire and then delete files due to low disk space condition as described in docs, i.e. till 50% of user quota will be reached. Probably important: I do not have user quota set.

### Actual behaviour
Trashbin have files 180 days old. More space is not freed. Only files that are older than 180 days are expired and deleted.

### Server configuration
**Operating system**:
CentOS 6
**Web server:**
Apache2 + PHP as Apache module
**Database:**
MySQL
**PHP version:**

PHP 5.4.42 (cli) (built: Jun 24 2015 13:41:14)
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.5, Copyright (c) 1999-2015, by Zend Technologies

**ownCloud version:** (see ownCloud admin page)

ownCloud 8.2.10 (production)

**Updated from an older ownCloud or fresh install:**

Live system with older updates since 8.0

**Where did you install ownCloud from:**

From official website zip.

**List of activated apps:**

```
Enabled:
  - activity: 2.1.4
  - announcementcenter: 1.0.3
  - bookmarks: 0.6.2
  - calendar: 0.8.2
  - contacts: 0.5.0.0
  - documents: 0.11.2
  - external: 1.2
  - files: 1.2.1
  - files_external: 0.3.0
  - files_pdfviewer: 0.7.1
  - files_sharing: 0.7.0
  - files_texteditor: 2.0
  - files_trashbin: 0.7.0
  - files_versions: 1.1.0
  - files_videoviewer: 0.1.3
  - firstrunwizard: 1.1
  - gallery: 14.2.0
  - notifications: 0.1.0
  - provisioning_api: 0.3.0
  - search_lucene: 0.7.0
  - tasks: 0.8.1
  - templateeditor: 0.1
  - updater: 0.6
  - user_webdavauth: 1.1.0.2
Disabled:
  - calendarplus
  - encryption
  - files_antivirus
  - files_locking
  - files_w2g
  - mail
  - tasksplus
  - user_external
[...]

PR #30095, opened 2018-01-10
Fix deleted items auto expiration for users with no quota

VicDeo:
## Description
Fix regression introduced a long time ago in https://github.com/owncloud/core/commit/a3999036f720e73727dac07a337578f3e25f92aa#diff-0dc914768208fc24b2aa2f2a369bb684
It is based on a wrong assumption that FS free space could be negative. 

## Related Issue
https://github.com/owncloud/core/issues/27412
obsoletes https://github.com/owncloud/core/pull/29951
## Motivation and Context
Trashbin auto clean up doesn't work when a user has no quota

## How Has This Been Tested?
1. Set `trashbin_retention_obligation` to `auto`
2. Create a user with no quota
3. Upload some files and delete uploaded files for that user
4. Keep uploading until all partition for `datadir` is occupied

### Expected 
Deleted files are purged to save space

### Actual
They are not

## Types of changes

## Checklist:

To document: 
New config.php `'trashbin_purge_limit'` that was just a constant before.
Introduced to fine tune purging in cases like https://github.com/owncloud/core/issues/2936

### DIFF
--- a/apps/files_trashbin/lib/AppInfo/Application.php
+++ b/apps/files_trashbin/lib/AppInfo/Application.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * @author Roeland Jago Douma <rullzer@owncloud.com>
- * @author Victor Dubiniuk <dubiniuk@owncloud.com>
+ * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
  *
  * @copyright Copyright (c) 2018, ownCloud GmbH
  * @license AGPL-3.0
@@ -23,6 +23,7 @@
 namespace OCA\Files_Trashbin\AppInfo;
 
 use OCA\Files_Trashbin\Expiration;
+use OCA\Files_Trashbin\Quota;
 use OCP\AppFramework\App;
 
 class Application extends App {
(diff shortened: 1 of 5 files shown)


PVince81:
@VicDeo that's fantastic ! Can you quickly explain what approach you used to fix it ?

The commit is called "Refactor Trashbin::calculateFreeSpace to be testable" which doesn't imply a fix and I can't seem to quickly find the actual location in the code where the fix is happening.

Suggestion: have one commit for the testable refactor and another one with the actual fix.

VicDeo:
@PVince81 
I just reverted the changes from the previous test and moved the method into another non-static class

As soon as it will reintroduce purging of deleted files described in  https://github.com/owncloud/core/issues/2936 
`trashbin_purge_limit` could be set to 30%, 20%, 10%, etc in case there are no users with hard quota

PVince81 (approved):
Makes sense, thanks 👍

PVince81:
Please backport

VicDeo:
Stable10: https://github.com/owncloud/core/pull/30163

PR #29977, opened 2017-12-27
Bump latest symfony and sabre/vobject point versions

phil-davis:
## Description
1) Do ``composer update`` and it bumps symfony and sabre/vobject point versions
2) Change the mentions of Sabre VObject 4.1.* in tests to the current 4.1.4

## Motivation and Context
Other PRs that bump things with ``composer`` are also finding these point version changes of other stuff - so get them into core.

## How Has This Been Tested?
CI knows

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30102, opened 2018-01-11
[stable10] Bump latest symfony and sabre/vobject point versions

phil-davis:
Backport #29977


phil-davis:
Note: this also bumps minor versions of other stuff in ``composer.lock`` up to similar versions that are in master. That is because the only reasonable way to resolve the backport conflicts for ``composer.lock`` is to run ``composer update`` again in ``stable10`` to make a nice new ``composer.lock`` - and of course that bumps everything it can find to the latest compatible versions.

composer.lock
```diff
@@ -2875,42 +2889,43 @@
...
             ],
-            "time": "2017-01-28T02:37:08+00:00"
+            "time": "2018-01-04T15:09:34+00:00"
         }
     ],
     "packages-dev": [
         {
             "name": "behat/behat",
-            "version": "v3.3.0",
+            "version": "v3.4.3",
```
DeepDiver1975: I prefer to only update libs which are related to this change.

phil-davis: To resolve conflicts in trying to backport ``composer.lock`` it is really necessary to run ``composer update`` and let it regenerate ``composer.lock`` again. That file is not the sort of thing to hand-edit.
So this here is the result. 
What to do?

DeepDiver1975: to update the lock file try running
```sh
composer update --lock
```

PVince81: Try `make update-composer` instead, it should regenerate the lock file. Not sure if that's the same...

phil-davis: I used ``make update-composer`` this time while doing a rebase to resolve conflicts. That produces a similar result - plenty of bumps of versions like this.

phil-davis:
See #30266 for another try at this.

patrickjahns:
@phil-davis 
To update just one library - it is straight forward `composer update <library> ` - in this case `composer update sabre/vobject `

phil-davis:
@patrickjahns #30266 has a better attempt at this.
The other version bumps that happen here if you do ``composer update`` are things that have already happened in ``master`` and need/should/could be bumped in ``stable10`` also.

phil-davis:
Done in #30266

Issue #19153, opened 2015-09-18
User Can Accept Shares from Other Users


MTRichards:
User Accept Shares: As a user, I want to confirm acceptance of files shared with me before I add them to my account so that I can ensure audit controls (based on notifications app)
Acceptance Criteria:
-   This confirmation is optional, and the overall capability is an on/off switch in the admin panel
-   When files are shared, the owners checks a box “confirm acceptance” which sends a notification and waits on an acceptance of the share
-   If the share is accepted, it appears in the recipients account. 
-   If the share is not accepted, the share owner is notified in their activity stream.

PVince81:
@schiesbn @jancborchardt this was discussed a long time ago :smile:

jancborchardt:
Yup, as already mentioned way back in https://github.com/owncloud/core/issues/4437

> When files are shared, the owners checks a box “confirm acceptance” which sends a notification and waits on an acceptance of the share

Not another one of these options though – at least not being the default. Shares should always need to be accepted, for reasons of storage space, unexpected files turning up in your account etc. Yes maybe we’ll have a special option or plugin which allows people to force shares, but an option like this is not needed by default.

Part of this is also having the overview page of pending & rejected shares: https://github.com/owncloud/core/issues/13495

cc @owncloud/designers

cdamken:
@MTRichards 

00004636

PVince81:
This could also solve some performance issues: whenever a share is created, there is code that will try and resolve file name conflicts for example if you received a folder "test" but already had a folder with this name locally. The conflict resolution can be slow when sharing with big groups.

If every user needs to accept their share separately, the resolution can be done at that time instead of at the time the share is created.

Is that actually still the case in 9.0 though ? @rullzer @schiesbn

MTRichards:
This is a pretty big feature, but certainly one that has been out there for a long time. Interested in the state in 9.0, and also with federated sharing in 9.0, then we can discuss if we want to do this / how we want to do this.

There are plenty of use cases where this is not wanted or needed, and some where it is wanted.

MTRichards:
I agree completely with this statement above:
>Not another one of these options though – at least not being the default. Shares should always need to be accepted, for reasons of storage space, unexpected files turning up in your account etc. Yes maybe we’ll have a special option or plugin which allows people to force shares, but an option like this is not needed by default.

We do have lots of users who are happy without this feature, and we can't change the default behavior, or clutter it up.

rullzer:
[...]

PR #30106, opened 2018-01-11
Accept user share notification

PVince81:
## Description
New notification to accept shares and pending shares list
    
- Reusing the "oc_share.accepted" field to store share states.
- Added checkbox in settings page to enable automatically accepting shares, this defaults to enabled currently. Unchecking this option will prompt users to accept or reject shares through notifications
- Added pending and rejected shares in the "Shared with you" view
- The "unshare from self" action now changes the state of shares to "Rejected" instead of deleting directly

## Related Issue
https://github.com/owncloud/core/issues/19153

## How Has This Been Tested?
TODO - for testcases see https://github.com/owncloud/core/pull/30106#issuecomment-369961079

fireshot capture 21 - mit ihnen geteilt - owncloud_ - http___localhost_8000_index php_apps_files_

## Types of changes

## Checklist:

## TODOs:

- [x] expose share state value in PHP API
- [x] add setting in share settings page to switch on this feature
- [x] send notification to user
- [x] send notification to group members
- [x] implement endpoint for accepting share
- [x] implement endpoint for declining share
- [x] discard notification if share is deleted
- [x] discard notification if target user is not member of the group any more
- [x] add missing unit tests
- [x] do not render folder favorite start


PVince81:
To test you need to untick the new auto accept option in the admin settings for sharing

lib/private/Share20/Share.php
```diff
@@ -76,6 +79,7 @@ class Share implements \OCP\Share\IShare {
 	public function __construct(IRootFolder $rootFolder, IUserManager $userManager) {
 		$this->rootFolder = $rootFolder;
 		$this->userManager = $userManager;
+		$this->state = Constants::STATE_ACCEPTED;
```
jvillafanez: I'm not sure if setting "accepted" as default value for the state is a good idea. Maybe setting null as default and enforcing that the state has a valid value just before saving it in the DB is a better idea.

apps/files_sharing/lib/Notifier.php
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		// Read the language from the notification
+		$l = $this->factory->get('files_sharing', $languageCode);
+
+		// TODO: discard if share does not exist any more
+		// TODO: discard if target user is not in target group any more
+
+		switch ($notification->getSubject()) {
+			case 'local_share':
```
jvillafanez: We should move this block to a private function. If it's implemented as "case" I expect more cases (public shares, federated shares, etc) that will eventually be implemented. Those cases will follow what is implemented here, so there will be more big blocks increasing the lines of this function making it more difficult to read.

apps/files_sharing/lib/Notifier.php
```diff
@@ -0,0 +1,92 @@
...
+				$params = $notification->getSubjectParameters();
+				if ($params[0] !== $params[1] && $params[1] !== null) {
+					$notification->setParsedSubject(
+						(string) $l->t('User %1$s shared "%3$s" with you (on behalf of %2$s)', $params)
+					);
+				} else {
+					$notification->setParsedSubject(
+						(string)$l->t('User %1$s shared "%3$s" with you', $params)
```
jvillafanez: Missing space after the typecasting (for consistency with the rest of the code)

apps/files_sharing/lib/Notifier.php
jvillafanez: My point isn't refactor, it's just move the code block to a private function in order to make this swicth block as compact as possible.

PVince81:
- [ ] if autoaccept disabled, do send a notification to the receiver that a new share was added (note: we could even do this as a separate PR as this can be standalone)
- [ ] if autoaccept enabled, send notification to share sender about acceptation/rejection

apps/files_sharing/lib/Notifier.php
PVince81: moved

PVince81:
Added notification when autoaccept is disabled. It includes a private link for the accepted share.

Also added notification icon.

PVince81:
- [ ] TODO: migrate permissions=0 to accepted=1 and make sure old permissions zero code still works in case of old data

PVince81:
- [ ] needs extensive testing with reshares...

PVince81:
- [ ] TODO: feedback after rejecting/accepted button:
    - refresh file view
    - display yellow notification

PVince81:
I've pushed some changes.

It is already possible to accept and reject shares from the notification buttons.
Unshare from self now implicitly triggers a rejection of a share. This means that user or group share are not getting deleted but just marked as rejected.

This is still very rough and I'm afraid we might have side effects. I'm not happy by the fact that we have a single endpoint "/shares" for both outgoing and received shares and also common code for `formatShare()` which is supposed to cover both cases. This is all still very messy and might need another round of refactoring, but probably another time as we don't want to break more.

I've added some todos above, we need to make sure this works with reshare / share initiator. It will anyway require a full retest of the sharing feature for user/group and also the "shared with me" and "shared with you" file lists, which seem to be affected by the share state somehow.

The endpoint "/shares?shared_with_me=true" now has an additional arguments "state":
- if omitted, it defaults to 0 (accepted). This is to prevent breaking any clients which expect only receiving a list of already accepted shares
- if set to empty string `state=` or `state=all`, it will retrieve shares with all possible states
- if set to a specific value, it will return only the shares with the given state. (0 = accepted, 1 = pending, 2 = rejected)

From here on we'll be able to adjust the "shared with you" list to also visually display pending and rejected shares.

jvillafanez:
> if set to empty string state= or state=all, it will retrieve shares with all possible states

I don't think that `state=` and `state=all` should have the same behaviour. To have consistency across all the API calls, `state=` should either cause an error due to invalid arguments, or use the default value.

It might be nice if the state can accept a list, something like `state=0,1` or `state=0,2`, but taking into account that I don't think there will be more states, it's probably too much work for a little gain. I don't have a use case where this could be useful at the moment.

PVince81:
> I don't think that state= and state=all should have the same behaviour.

Thanks for your comment. I wasn't sure either.

- [ ] "state=" should use the default value. Keep "all" for all.

I don't want to add muultiple state values as input as it would introduce a lot more logic for not much gain as you said.

PVince81:
Now struggling with the fileactions part: the "Shared with you" list already displays the pending shares now. We need to adjust the file actions menu to display "accept" and "reject" actions or others, but not the default actions.

Currently, the FileActions object was designed based on its legacy: to define file actions based on mime type + permissions. Ideally it should be refactored to provide custom actions with custom filtering behavior instead of mime/permissions. This way we could have the custom handler filter out actions based on the current file row that was clicked. Whatever change we do, we need to keep it compatible with the old way to avoid breaking apps that register their actions.

PVince81:
I've made a new PR to look into making file actions more flexible: https://github.com/owncloud/core/pull/30478

PVince81:
Pending shares can now be accepted in "shared with you", but with many bugs:

- [ ] prevent sidebar to open for pending/rejected shares
- [ ] implement custom sort by state
- [ ] prevent "unshare from self" from removing element: redisplay it in rejected mode instead
- [ ] do we really need three states ?
- [ ] missing star placeholder / alignment when pending share is the only entry

PVince81:
the favorite star is missing so the first column is not wide enough for pending shares:
image

I wonder if we could reuse that space to display the icon as a state ? @felixheidecke

PVince81:
update:
image

PVince81:
Hmm we should probably also display the originator of the share like for the accepted one:

- [ ] display originator (share owner)

PVince81:
- [ ] share owner indicator must not be clickable as there is no matching action. or needs visual distinction

PVince81:
added more unit tests and fixed some. still not enough, codecov will tell.

PVince81:
- [ ] add acceptance tests...

SamuAlfageme:
- [ ] add API documentation for client consumption

PVince81:
To clarify: scenario when deleting parent folder:

1. user1 received a share and put it into "Shared/received_share"
1. user1 deletes the folder "Shared"

In past versions the folder "received_share" would automatically get "unshared from self" and the share is deleted.
Now with this PR and the new approach, the code marks the share as rejected and would still appear under "pending shares" so the user can accept it again. When accepting the share would then be mounted in "/received_share".

Is that acceptable behavior or should it be different ? @pmaier1

PVince81:
- [ ] add option to hide rejected shares (or make a ticket to add this later) as a user cannot completely delete rejected shares, only the owner can

PVince81:
fixed all failing unit tests

pmaier1:
> Now with this PR and the new approach, the code marks the share as rejected and would still appear under "pending shares" so the user can accept it again. When accepting the share would then be mounted in "/received_share".

Basically that sounds reasonable. One thing: Why would the share appear under "pending shares"? The user accepted the share before and then unshared so it should appear in the "rejected shares" section, IMO. The user can then still restore it triggering the flow you described.
In my perception the "pending shares" section is only for shares that have not been accepted nor rejected yet.

PVince81:
added missing unit tests for Share20OCS::getShares() for the regular case and also "shared_with_me" case, the latter has been touched due to the introduction of the state filter

PVince81:
> Why would the share appear under "pending shares"? 

@pmaier1 I mistyped. You are right, the shares in question become rejected so will appear under "rejected shares". Good, no code changes needed as this is the current behavior in this PR. Thanks!

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:
Automated rebase failed! Please rebase your pull request manually via the command line.

Reason:
```
Command: git rebase base/master

Exit Code: 128

Cause:

error: Failed to merge in the changes.


```

PVince81:
all unit tests passed, next up is fixing the existing acceptance test. (later on add more...) and then code cov results will appear and tell us how much we missed

PVince81:
something is wrong with this branch... rebasing locally seems to discard my changes or thinks my older commits are empty ?!

PVince81:
what, even squashing the changes in the current branch doesn't work ?

@felixheidecke in the future please avoid using merge commits on PRs, the problem is likely related to that

PVince81:
Squashed, solved conflict and rebased onto master.

PVince81:
- [ ] sharing with a big group likely to be slow as we need to create one notification entry for every member: should we move this creation to a background job and do this in batches then ?

PVince81:
## Acceptance tests to implement
- [ ] notifications
    - [ ] share to user sends notification
    - [ ] share to group sends notification to every member
    - [ ] accept share from notification makes share visible in "all files", and visible in "shared with you" as "accepted"
    - [ ] reject share from notification makes it not visible in "all files", and visible in "shared with you" as "rejected"
- [ ] Shared with you view
    - [ ] unshare user-share from self makes it dissapear, still visible in "shared with you" as "rejected"
    - [ ] unshare group-share from self makes it dissapear, still visible in "shared with you" as "rejected"
    - [ ] accept pending share from "shared with you" makes it accepted after page reload, visible also in "all files"
    - [ ] reject pending share from "shared with you" makes it accepted after page reload, not visible in "all files"
    - [ ] accept an already rejected share
    - [ ] reject an already accepted share
- [ ] migration
    - [ ] on older 10.0.x version unshare group share from self, upgrade to here, check that share appears "rejected". accept "rejected" share works. (last step is important due to internal data structure that gets refreshed from old to new)

PVince81:
- [ ] BUG: rejected a received group share that was renamed will reset the folder name to the original. this might confuse users as the name shown in the UI is still the renamed one. So better preserve it if possible.

felixheidecke:
> @felixheidecke in the future please avoid using merge commits on PRs, the problem is likely related to that

will do.

I think we should get rid of the favorite star in this section since it unnecessarily indents and is of no importance here. Agreed?

PVince81:
> I think we should get rid of the favorite star in this section since it unnecessarily indents and is of no importance here. Agreed? If so … I'll do it.

I'm not sure I'm convinced yet about the new indicators and whether they are properly understood by people. If we are to keep the indicators we could get rid of the star icon as the user can also star using the sidebar.

felixheidecke:
Let's fall back to the click dummy and refrain from using the icons in the front. I'll go ahead and change the pending / removed share to light gray.

PVince81:
> Let's fall back to the click dummy and refrain from using the icons in the front. I'll go ahead and change the pending / removed share to light gray.

Ok, sounds good. Once we have a stable version we could do some usability testing to find out how clear or confusing this is, not only "shared with you" visually but the whole sharing process.

felixheidecke:
I updated the description and added a screenshot of the current state.

PVince81:
- [ ] keep auto-accept mode only for updated systems, but unset it for new installs ? (to avoid surprises on update)

@pmaier1 thoughts ?

PVince81:
Updated some PHP tests.

Added JS tests for file actions additions and also the special accept/reject action handlers.

Still missing:
- [ ] JS tests for shared file list

PVince81:
note: to discard an obsolete notification, need to call `$notificationManager->markProcessed($notification)` from the code path that makes it obsolete: https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/External/Manager.php#L253

PVince81:
I wonder if we should move the notification triggering logic to the ShareManager: this way when an app creates a share through PHP API it would also produce a notification. Currently it only works through the OCS API... This would likely be better because in the case of cleaning up notifications it would have to happen through events: because sometimes the share deletion is triggered by a file operation, not OCS.

- [ ] move notification handling to ShareManager ?

PVince81:
or have the notification handling be an independent component that reacts on share events, this is likely the better approach!

- [ ] move notification handling to a hooks listener
    - [ ] when share created: send notification
    - [ ] when share deleted completely: delete notifications related to the share

PVince81:
- [ ] regular folder icon broken in "shared with you" view, for some reason it points to "/core" instead of "/owncloud" on my setup

PVince81:
Added proper sorting by state, the column can now be clicked. Also it's sorted by state by default.
The sort by shareState will first sort by state (accepted, pending, rejected) and then by file name.

SamuAlfageme:
Nice job! Was checking how the notifications look like in the client (for https://github.com/owncloud/client/issues/6378) and the wording be polished a bit: 

screenshot 2018-03-09 10 16 42

- They're not "remote shares" and we don't really need to know the federated share id of the originator 
- The name of the share is relative to the mountpoint - i.e. includes the "`/`" as the prefix. This is irrelevant to the end user.

jvillafanez:
> They're not "remote shares" and we don't really need to know the federated share id of the originator

Those notifications come from the remote sharing, they aren't generated by this PR

PVince81:
@felixheidecke please fix https://github.com/owncloud/core/pull/30106#issuecomment-370813863

PVince81:
- [ ] in case of update: keep "auto accept" ticket. else unticked by default
- [ ] change wording: "reject" => "decline"
- [ ] sort by state: pending first, then accepted, then rejected
- [ ] add link to notification to point to the "Shared with you" list with the entry highlighted

apps/files_sharing/css/sharedfilelist.css
```diff
@@ -1,3 +1,23 @@
...
+tr[class*='share-state-'] .action-favorite {
+	display: none !important;
+}
+
+.share-state-pending .thumbnail {
+	background-image: url("../img/folder-waiting.svg") !important;
+}
+
+.share-state-rejected .thumbnail {
+	background-image: url("/core/img/filetypes/folder.svg") !important;
```
individual-it: that url does not work if owncloud is not installed in the root of the domain

individual-it:
please rebase again on master, to get the current version of acceptance tests (API & UI)

individual-it:
There seem to be problems with federation shares:

1. share from "admin@localhost/owncloud-core" to "user2@http://192.168.0.106/owncloud-core/" (same server)
2. as "user2" accept share by notification button
3. new share is visible in "all files"
4. but in "shared with you" it's still pending 
image
5. click on "Accept" there and reload page gives this image:
image
6. unsharing throws:
```
{"reqId":"aotXYEzD6DRGNOxM9rka","level":0,"time":"2018-03-14T06:47:00+00:00","remoteAddr":"127.0.0.1","user":"--","app":"webdav","method":"PROPFIND","url":"\/owncloud-core\/public.php\/webdav\/","message":"Exception: {\"Message\":\"HTTP\\\/1.1 401 Username or password was incorrect\",\"Exception\":\"Sabre\\\\DAV\\\\Exception\\\\NotAuthenticated\",\"Code\":0,\"Trace\":\"#0 [internal function]: Sabre\\\\DAV\\\\Auth\\\\Plugin->beforeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#1 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#2 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(466): Sabre\\\\Event\\\\EventEmitter->emit('beforeMethod', Array)\\n#3 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#4 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/appinfo\\\/v1\\\/publicwebdav.php(95): Sabre\\\\DAV\\\\Server->exec()\\n#5 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/public.php(86): require_once('\\\/home\\\/artur\\\/www...')\\n#6 {main}\",\"File\":\"\\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Auth\\\/Plugin.php\",\"Line\":168,\"User\":false}"}
{"reqId":"J5eZwDNGo2SqMbwPqVDt","level":3,"time":"2018-03-14T06:47:00+00:00","remoteAddr":"192.168.0.106","user":"user2","app":"no app in context","method":"DELETE","url":"\/owncloud-core\/remote.php\/dav\/files\/user2\/simple-folder%20(2)","message":"Exception: {\"Exception\":\"Sabre\\\\HTTP\\\\ClientHttpException\",\"Message\":\"Unauthorized\",\"Code\":401,\"Trace\":\"#0 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Client.php(231): Sabre\\\\HTTP\\\\Client->send(Object(Sabre\\\\HTTP\\\\Request))\\n#1 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Storage\\\/DAV.php(198): Sabre\\\\DAV\\\\Client->propFind('http:\\\/\\\/localhos...', Array, 1)\\n#2 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Storage\\\/Wrapper\\\/Wrapper.php(97): OC\\\\Files\\\\Storage\\\\DAV->opendir('')\\n#3 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Storage\\\/Wrapper\\\/Availability.php(105): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Wrapper->opendir('')\\n#4 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Storage\\\/Wrapper\\\/Wrapper.php(97): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Availability->opendir('')\\n#5 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Storage\\\/Wrapper\\\/Wrapper.php(97): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Wrapper->opendir('')\\n#6 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Cache\\\/Scanner.php(354): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Wrapper->opendir('')\\n#7 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Cache\\\/Scanner.php(406): OC\\\\Files\\\\Cache\\\\Scanner->getNewChildren('')\\n#8 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Cache\\\/Scanner.php(386): OC\\\\Files\\\\Cache\\\\Scanner->handleChildren('', false, 3, '481', true, 0)\\n#9 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/Cache\\\/Scanner.php(319): OC\\\\Files\\\\Cache\\\\Scanner->scanChildren('', false, 3, '481', true)\\n#10 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/View.php(1369): OC\\\\Files\\\\Cache\\\\Scanner->scan('', false)\\n#11 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/private\\\/Files\\\/View.php(1410): OC\\\\Files\\\\View->getCacheEntry(Object(OCA\\\\Files_Trashbin\\\\Storage), '', 'simple-folder (...')\\n#12 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/lib\\\/Connector\\\/Sabre\\\/ObjectTree.php(179): OC\\\\Files\\\\View->getFileInfo('\\\/user2\\\/files\\\/si...')\\n#13 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/lib\\\/Tree.php(67): OCA\\\\DAV\\\\Connector\\\\Sabre\\\\ObjectTree->getNodeForPath('simple-folder (...')\\n#14 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/lib\\\/DAV\\\/AbstractCustomPropertiesBackend.php(238): OCA\\\\DAV\\\\Tree->getNodeForPath('files\\\/user2\\\/sim...')\\n#15 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/lib\\\/DAV\\\/FileCustomPropertiesBackend.php(230): OCA\\\\DAV\\\\DAV\\\\AbstractCustomPropertiesBackend->getNodeForPath('files\\\/user2\\\/sim...')\\n#16 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/lib\\\/DAV\\\/FileCustomPropertiesBackend.php(74): OCA\\\\DAV\\\\DAV\\\\FileCustomPropertiesBackend->getNodeForPath('files\\\/user2\\\/sim...')\\n#17 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/PropertyStorage\\\/Plugin.php(126): OCA\\\\DAV\\\\DAV\\\\FileCustomPropertiesBackend->delete('files\\\/user2\\\/sim...')\\n#18 [internal function]: Sabre\\\\DAV\\\\PropertyStorage\\\\Plugin->afterUnbind('files\\\/user2\\\/sim...')\\n#19 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#20 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/CorePlugin.php(288): Sabre\\\\Event\\\\EventEmitter->emit('afterUnbind', Array)\\n#21 [internal function]: Sabre\\\\DAV\\\\CorePlugin->httpDelete(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#22 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/event\\\/lib\\\/EventEmitterTrait.php(105): call_user_func_array(Array, Array)\\n#23 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(479): Sabre\\\\Event\\\\EventEmitter->emit('method:DELETE', Array)\\n#24 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/dav\\\/lib\\\/DAV\\\/Server.php(254): Sabre\\\\DAV\\\\Server->invokeMethod(Object(Sabre\\\\HTTP\\\\Request), Object(Sabre\\\\HTTP\\\\Response))\\n#25 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/lib\\\/Server.php(279): Sabre\\\\DAV\\\\Server->exec()\\n#26 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/apps\\\/dav\\\/appinfo\\\/v2\\\/remote.php(31): OCA\\\\DAV\\\\Server->exec()\\n#27 \\\/home\\\/artur\\\/www\\\/owncloud-core\\\/remote.php(175): require_once('\\\/home\\\/artur\\\/www...')\\n#28 {main}\",\"File\":\"\\\/home\\\/artur\\\/www\\\/owncloud-core\\\/lib\\\/composer\\\/sabre\\\/http\\\/lib\\\/Client.php\",\"Line\":160}"}
{"reqId":"J5eZwDNGo2SqMbwPqVDt","level":3,"time":"2018-03-14T06:47:00+00:00","remoteAddr":"192.168.0.106","user":"user2","app":"files_external","method":"DELETE","url":"\/owncloud-core\/remote.php\/dav\/files\/user2\/simple-folder%20(2)","message":"Unauthorized"}
```

PVince81:
> There seem to be problems with federation shares:

This PR is not supposed to affect anything related to federated shares. It also doesn't bring in any logic to display pending federated shares.

Your finding tells me that some of the refactoring might have broken it.

- [ ] BUG: make sure federated sharing + shared with you view works as before with federated shares (no pending entries)

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

PVince81:
- [ ] dispatch symfony events on accept/reject like for federated shares

PVince81:
- [ ] make private link to pending share open the "Shared with you" panel
    - [ ] REQUIRES: https://github.com/owncloud/core/pull/30795

individual-it:
I think there is a problem when receiving shares from multiple users when the name of the share is the same:
1. create user1, user2, user3
2. as user1 create folder "new" and "new/user1"
3. as user2 create folder "new" and "new/user2"
4. as user1 share "new" with user3
5. as user2 share "new" with user3
6. as user3 check that you see 2 shares pending
7. accept the share created by "user**2**"
8. check that you can see the folder "new" in "All files" and its the correct one
9. accept the share created by "user**1**"
10. check "All files" => the share you have accepted first (from user2) got renamed to "new (2)" and the share you have accepted last is now "new"

PVince81:
> looks like the naming depends on the time of the share creation, the receiver does get the newest shares presented at the top of the list and would naturally accept them first

I'd rather expect the final name to be based not on share creation time but at the time it is accepted.

PVince81:
- [ ] check case with duplicate names

individual-it:
"Shared with you" page is not updated on a simple reload, need to click on the link again to get current information
no-refresh-on-reload

individual-it:
name of an already existing folder only updates after reload on the "shared-with-you" page
1. receive a folder with a name that already exists in the own structure
2. accept
3. the folder name didn't got updated in "Share with you"
4. click again on "Share with you" to see the new name

PVince81:
@individual-it please add checkboxes for the bugs you find for easier tracking. I've edited your last comments for that. Thanks!

individual-it:
- [ ] BUG: "Shared with others" and "Shared by link" also show "State" column
image
image

PVince81:
Arghh there's some tricky stuff in there.

When accepting a share, the share changes state in the controller. So far so good.
However the attached nodes are still inaccessible even after accepting a share because within that one HTTP request, the filesystem was still initialized without the share attached. So the response of an "accept" operation is empty and the log shows:
```
{"reqId":"N7ns0T6PxqmLJ61Odbzg","level":3,"time":"2018-03-19T15:46:37+00:00","remoteAddr":"127.0.0.1","user":"user1","app":"PHP","method":"POST","url":"\/owncloud\/ocs\/v1.php\/apps\/files_sharing\/api\/v1\/shares\/pending\/1?format=json","message":"OCP\\Files\\NotFoundException:  at \/srv\/www\/htdocs\/owncloud\/apps\/files_sharing\/lib\/API\/Share20OCS.php#181"}
```

To be able to return a response and also show the correct name in the UI (in case of appended "(2)" after accept), we need to tear down the FS and set it up again before returning the response.

This is the kind of thing that usually needs calling horrible private static functions...

individual-it:
@ownclouders rebase

ownclouders:

ownclouders:

individual-it:
- [ ]  BUG: declining immediately after accepting does not work on the webUI (works only after page reload)
    1. accept a share
    2. decline the same share (e.g. because you have clicked the wrong button)
    3. the share will not be declined because the id is missing in the request 
    REQUEST: `ocs/v1.php/apps/files_sharing/api/v1/shares/pending/?format=json`
    RESPONSE: `{"ocs":{"meta":{"status":"failure","statuscode":999,"message":"Invalid query, please check the syntax. API specifications are here: http:\/\/www.freedesktop.org\/wiki\/Specifications\/open-collaboration-services. DEBUG OUTPUT:\n"},"data":[]}}`

individual-it:
- [ ] BUG: it is possible to "enter" pending and declined shares, resulting in listing all accepted shares
enter-pending-declined-shares

PVince81:
Fixed some of the bugs above.

- [ ] proper error handling when changing share state (currently it crashes into the console)
- [ ] proper tearDown API, don't use private APIs

individual-it:
rebased and solved conflicts with master to get all the new goodies in the acceptance tests

PVince81:
@individual-it I've solved most of the bugs today, please try again.

- [ ] adjust unit tests for last commit related to name deduplication (they'll fail currently)

individual-it:
@PVince81 looks much better, most tests pass. I'm keep on adding more of them

- [ ] BUG: visual problem in Firefox:
image

individual-it:
- [ ] BUG: wrong pointer when hover "Accept Share" & "Decline Share"
wrong-mouse-pointer

individual-it:
@PVince81 that test fails on the API side https://github.com/owncloud/core/pull/30106/files#diff-740f91c66cd89f180412448a68c66a13R154
Please check if the logic of the test is what is intended behaviour

PVince81:
@individual-it the logic of the test is the intended behavior, yes. When re-accepting it would mount it back to where it was before, given that "/PARENT" is still available. This is akin to restoring a file from trashbin.
If "/PARENT" is gone it would slide into "/" then. (maybe needs another test)

PVince81:
@individual-it if you think you have covered some of the mentionned cases please tick the boxes here: https://github.com/owncloud/core/pull/30106#issuecomment-369961079

PVince81:
- [ ] TEST: configure default receive shared folder `'share_folder' => '/PARENT'` and make sure received share lands there by default
     - [ ] case 1: "/PARENT" folder does not exist => it gets automatically created
     - [ ] case 2:  "/PARENT" folder already exists => it is just reused

I'm not sure whether we need this in an acceptance test but this will definitely be in the unit test I'm adding.

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

PVince81:
private links now point to the "Shared with you" pending share entry: https://github.com/owncloud/core/pull/30106#issuecomment-373452006

individual-it:
- [ ] BUG: share notifications show username only and not displayname

PVince81:
- [ ] remove notifications in case of auto-accept (for now)

PVince81:
- [ ] REQUIRES https://github.com/owncloud/core/pull/30930 for discarding notifications in case a share is deleted (the listener is already there, just need to test it)

PVince81:
As discussed with @pmaier1 today for the first version we do not send any notifications when auto-accept is enabled (which matches old behavior). We only send when auto-accept is disabled.
I've adjusted the test plan accordingly.

felixheidecke:
@individual-it I can't find the page where this bug occurs: https://github.com/owncloud/core/pull/30106#issuecomment-374867432

jvillafanez:
https://github.com/owncloud/core/pull/30996 ready in order to progress with the "discard notification if target user is not member of the group any more" task

jvillafanez:
Looks like the "discard notification if target user is not member of the group any more" task will be more complex than initially thought:

In order to remove the notification we need the userId and the shareId in order to create a notification that can be used as a filter. Although the userId is easy, the real problem is in the shareId:
1. The "group.preRemoveUser" event doesn't provide any information about the share (it doesn't know anything about sharing), so we must somehow fetch all the shares of the user that is going to be deleted. First look it seems quite a heavy task to fetch all the shares the user can access and filter them. Hopefully using `getAllSharedWith` and filter by group should help.
2. Point 1 MUST be done in the "group.preRemoveUser" event. There is another hook to remove the shares of the user once he is removed from the group. Since we can ensure any event order, we should consider that such callback happens before our one, which means that when we try to fetch the shares they are already gone.
3. As said, fetching the share information must be done in the "group.preRemoveUser", but the actual action to remove the notification should be done after the user is removed, which mean we need to use a global variable (or any other trick) to communicate the information fetched from the "group.preRemoveUser" callback to the "group.postRemoveUser" one.

Another (discarded) solution is to move the new "share.afterDelete" (and other) events down to the DefaultSharingProvider so we make sure the events are triggered when the shares are deleted, specially with the `userDeletedFromGroup` method of the IDefaultProvider (which is the method that is triggered when a user is deleted from the group). The problem with this approach is that it's quite intrusive because the events are thrown nearly at the DB layer.

PVince81:
@jvillafanez can we simply delete the notification at the time it is prepared by the "Notifier" ? The Notifier would detect that the share isn't accessible any more for the current user so instead of preparing the notification, return `null` and delete it (mark as processed). I believe we already do something similar in other apps.

individual-it:
- [ ] BUG: declining an remote share does not work from the "Shared with you" page
1. share from `admin@localhost` to `uuu1@127.0.0.1`
2. as `uuu1` accept the remote share from the notification
3. on the "Shared with you" page decline the remote share
4. Share cannot be declined
image
DELETE requests send to `ocs/v1.php/apps/files_sharing/api/v1/shares/pending/17?format=json`
response: `{"ocs":{"meta":{"status":"failure","statuscode":404,"message":"Wrong share ID, share doesn't exist"},"data":[]}}`

jvillafanez:
Rebased to fix the conflicts

jvillafanez:
Looks like there are problems with the parameters being sent by the event. There are some mismatches between what is expected to be sent and what is really sent.
Following patch fixes the unittest, but it's needed a second look in order to decide what do we exactly want to send in the events:
```
diff --git a/apps/files_sharing/tests/ApiTest.php b/apps/files_sharing/tests/ApiTest.php
index 0d4b357..b808de3 100644
--- a/apps/files_sharing/tests/ApiTest.php
+++ b/apps/files_sharing/tests/ApiTest.php
@@ -158,7 +158,7 @@ class ApiTest extends TestCase {
 
 		$this->assertEquals('share.beforeCreate', $calledBeforeCreate[0]);
 		$this->assertInstanceOf(GenericEvent::class, $calledBeforeCreate[1]);
-		$this->assertTrue($calledBeforeCreate[1]->getArgument('shareData')['run']);
+		//$this->assertTrue($calledBeforeCreate[1]->getArgument('shareData')['run']);
 		$this->assertTrue($result->succeeded());
 		$data = $result->getData();
 		$this->assertEquals(19, $data['permissions']);
diff --git a/lib/private/Share20/Manager.php b/lib/private/Share20/Manager.php
index 7372ab1..b06dc31 100644
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -876,7 +876,7 @@ class Manager implements IManager {
 
 		// Emit post hook
 		\OC_Hook::emit('OCP\Share', 'post_unshare', $hookParams);
-		$afterEvent = new GenericEvent(null, ['share' => $hookParams['deletedShares']]);
+		$afterEvent = new GenericEvent(null, ['share' => $hookParams['deletedShares'], 'shareObject' => $share]);
 		$this->eventDispatcher->dispatch('share.afterDelete', $afterEvent);
 	}
 
```

individual-it:
- [ ]  BUG: no data in HTTP response when deleting shares by API
1. create share `curl -u admin:admin http://localhost/owncloud-core/ocs/v1.php/apps/files_sharing/api/v1/shares -d "path=welcome.txt&shareType=0&shareWith=uuu1"`
2. delete share `curl -u admin:admin http://localhost/owncloud-core/ocs/v1.php/apps/files_sharing/api/v1/shares/98 -X DELETE`
3. result should be 
```<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```
but is empty

owncloud log: 
```
{"reqId":"VohgmJhDF5tH5OXOdMvp","level":3,"time":"2018-04-11T10:12:53+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"PHP","method":"DELETE","url":"\/owncloud-core\/ocs\/v1.php\/apps\/files_sharing\/api\/v1\/shares\/98","message":"InvalidArgumentException: Argument \"shareObject\" not found. at \/home\/artur\/www\/owncloud-core\/lib\/composer\/symfony\/event-dispatcher\/GenericEvent.php#63"}
```

jvillafanez:
Unittests should have been fixed by https://github.com/owncloud/core/pull/30106/commits/4d405b9f696f763d5a3c84fbd30fd5848e5eb5fb , as well as the bug above (https://github.com/owncloud/core/pull/30106#issuecomment-380402128). Waiting for CI to confirm.

jvillafanez:
Share autocompletition (https://example/ocs/v1.php/apps/files_sharing/api/v1/sharees) call seems to be broken. I'm not sure if it's related to the changes in this PR.

jvillafanez:
I'll need some expectations.
* For the acceptance scenario "Scenario: move accepted share, decline it, accept again" (https://drone.owncloud.com/owncloud/core/5783/204) when the user accepts the file again, the movement is lost, so the share appears again in the root. The behaviour looks acceptable to me, not sure what we want. @PVince81 @pmaier1 

The other failing scenario is a bug (confirmed). The share seems to be renamed again to "shared (2)" instead of keeping the same name.

jvillafanez:
Just to confirm the change in the behaviour for the documentation: the shares will be permanent from the receiver's point of view, so the "auto share deletion" that happened when the receiver rejected the share won't happen any longer. I think this is a good improvement.

jvillafanez:
screen shot 2018-04-12 at 10 01 09

I'm not sure if showing the preview is possible, but we should detect whether it's file or folder to show a better icon. Right now it seems that all of those shares are folder. @felixheidecke

jvillafanez:
* [ ] BUG: Actions executed from the notifications app don't remove the notification, so it pops up again and again. The actual action seems to be executed.
* [ ] ???: Duplicated information shown to the user with the notification app. When the user clicks in the link inside the email, it shows the "shared with you" with the shared file to be accepted or rejected as well as the notification from the notifications app. In addition, the notification isn't removed when an action is taken from the files view.

PVince81:
> For the acceptance scenario "Scenario: move accepted share, decline it, accept again" when the user accepts the file again, the movement is lost, so the share appears again in the root. The behaviour looks acceptable to me, not sure what we want. @PVince81 @pmaier1

Acceptable IMO

jvillafanez:
Adjusted the acceptance scenario "Scenario: move accepted share, decline it, accept again" to match the expectations

jvillafanez:
Failing tests (https://jenkins.owncloud.org/job/owncloud-core/job/core/job/PR-30106/66/console) seems unrelated. @individual-it could you confirm? Is it a known issue or fixed in master?

apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -505,12 +524,18 @@ private function getSharedWithMe($node = null, $includeTags) {
 
 		$formatted = [];
 		foreach ($shares as $share) {
-			if ($this->canAccessShare($share)) {
-				try {
-					$formatted[] = $this->formatShare($share);
-				} catch (NotFoundException $e) {
-					// Ignore this share
-				}
+			if ($stateFilter !== null && $share->getState() !== $stateFilter) {
```
jvillafanez: move these conditions to a private method?

jvillafanez:
Still not reached half way of the changes to review... 😓

individual-it:
@jvillafanez Jenkins can not run the tests that need the notifications app. drone does skip them and I'm working on making them run in drone. But jenkins tries to run them and fails, it would need some fixing to make jenkins work correctly.
I was hoping we could kill jenkins before merging this PR
@patrickjahns ?

apps/files_sharing/lib/Service/NotificationPublisher.php
```diff
@@ -0,0 +1,150 @@
...
+					yield $user->getUID();
+				}
+			}
+		} else if ($share->getShareType() === \OCP\Share::SHARE_TYPE_USER) {
+			yield $share->getSharedWith();
+		}
+	}
+
+	/**
+	 * Send notification for accepting share
```
jvillafanez: There are cases where the notification for the share won't be sent (share by link, auto-accepted share, etc). We should document those cases here.

apps/files_sharing/lib/Service/NotificationPublisher.php
```diff
@@ -0,0 +1,150 @@
...
+		$endpointUrl = $this->urlGenerator->getAbsoluteURL(
+			$this->urlGenerator->linkTo('', 'ocs/v1.php/apps/files_sharing/api/v1/shares/pending/' . $share->getId())
+		);
+
+		foreach ($this->getAffectedUsers($share) as $userId) {
+			$notification = $this->notificationManager->createNotification();
+			$notification->setApp('files_sharing')
+				->setUser($userId)
+				->setDateTime(new \DateTime())
+				->setObject('local_share', $share->getFullId());
```
jvillafanez: Comment needed here: the fullId is used here to be able to retrieve the share using the core's share manager; using just the id won't do because core needs to distinguish between several share providers.

apps/files_trashbin/js/filelist.js
```diff
@@ -251,14 +251,6 @@
 			);
 		},
 
-		_onClickFile: function(event) {
```
jvillafanez: any reason for this change? It seems unrelated.

lib/private/Share20/Manager.php
```diff
@@ -917,29 +917,17 @@ public function deleteFromSelf(\OCP\Share\IShare $share, $recipientId) {
...
-			$sharedWith = $this->groupManager->get($share->getSharedWith());
-			if (is_null($sharedWith)) {
-				throw new \InvalidArgumentException('Group "' . $share->getSharedWith() . '" does not exist');
-			}
-			$recipient = $this->userManager->get($recipientId);
-			if (!$sharedWith->inGroup($recipient)) {
-				throw new \InvalidArgumentException('Invalid recipient');
-			}
-		}
+		return $this->updateShareForRecipient($share, $recipientId);
```
jvillafanez: ~Shouldn't the manager call the provider's `move` method here and let it do whatever it needs? Otherwise, the `move` method in the `IShareProvider` interface is completely pointless. This needs a second thought.~
Better add a comment for clarification. The `move` method is deprecated in favor of the `updateShareForRecipient`, so technically this is fine, although I'm not sure if the implementation should be the one changing.

jvillafanez:
* [ ] Update "since" tags <- **important**

apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -810,13 +843,131 @@ public function updateShare($id) {
...
+		}
+
+		$pathAttempt = \OC\Files\Filesystem::normalizePath($parentDir . '/' . $share->getTarget());
+
+		$pathinfo = pathinfo($pathAttempt);
+		$ext = (isset($pathinfo['extension'])) ? '.'.$pathinfo['extension'] : '';
+		$name = $pathinfo['filename'];
+
+		$i = 2;
+		while ($userFolder->nodeExists($pathAttempt)) {
```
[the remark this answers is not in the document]
jvillafanez: It will need some research at least to know where this can be placed. Leaving this for now.

apps/files_sharing/lib/Service/NotificationPublisher.php
jvillafanez: Done just below

apps/files_sharing/lib/Service/NotificationPublisher.php
jvillafanez: Done just below

apps/files_sharing/lib/Hooks.php
```diff
@@ -60,4 +106,67 @@ public static function unshareChildren($params) {
 			}
 		}
 	}
+
+	public function registerListeners() {
```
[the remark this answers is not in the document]
jvillafanez: Delaying this for the time being because we'll likely need to refactor some parts of the code. It seems a big task for this PR.

jvillafanez:
Rebased to fix conflicts.

To be rechecked due to conflicts:

apps/files_sharing/lib/MountProvider.php
lib/private/Share20/DefaultShareProvider.php
lib/private/Share20/Manager.php
tests/acceptance/features/bootstrap/BasicStructure.php
tests/acceptance/features/bootstrap/WebUIFilesContext.php
tests/acceptance/features/bootstrap/WebUISharingContext.php
tests/acceptance/features/lib/FilesPageBasic.php
tests/acceptance/features/lib/SharedWithYouPage.php
tests/acceptance/run.sh

@individual-it could you recheck the acceptance tests to verify everything is correct? I'm not sure if I fixed the conflicts correctly

individual-it:
@jvillafanez conflict solving is fine from the acceptance tests point of view. Some extra spaces etc. but nothing that changes functionality.
still I cannot decline remote shares from "shared-with-you" page
image

PVince81:
@jvillafanez how far are we from completion (code complete / feature complete) ? Can you go through the checkboxes and also add some new ones if anything is open ?

@individual-it have you been be able to cover all scenarios in the tests ?

PVince81:
also please use checkboxes for bugs...

individual-it:
@PVince81 I think all cases are covered https://github.com/owncloud/core/pull/30106#issuecomment-369961079 
except of checking the default setting and upgrade, we haven't automated that yet

jvillafanez:
The main goal from this point on should be stabilize the PR and merge it the sooner the better, so any non-critical problem should be moved to another ticket in order to fix it once this PR is merged.

So my idea is:
* https://github.com/owncloud/core/pull/30106#issuecomment-359007815 - refresh files view from the notification + yellow message - move it out (you'll need to reload the page manually) 👉 
* https://github.com/owncloud/core/pull/30106#issuecomment-369544055 - hide rejected shares - move it out. 👉 
* https://github.com/owncloud/core/pull/30106#issuecomment-369611920 - share with a big group might be slow - move it out 👉 
* https://github.com/owncloud/core/pull/30106#issuecomment-372746060 - auto-accept enabled for updates and disabled for new installations - I'd keep it enabled in any case, otherwise move it out. 👉 
* https://github.com/owncloud/core/pull/30106#pullrequestreview-103697631 - background image not showing on non-root installations - to be investigated 🔍 
* https://github.com/owncloud/core/pull/30106#issuecomment-373424579 - dispatch accept / decline events - not sure about how important it is to move it out ❓ 
* https://github.com/owncloud/core/pull/30106#issuecomment-375880161 - tests with a different recieve shared folder - to be rechecked 🔍 
* https://github.com/owncloud/core/pull/30106#issuecomment-376412938 - several bugs - to be rechecked 🔍 
* https://github.com/owncloud/core/pull/30106#issuecomment-380731006 - duplicated information shown - move it out / do nothing. 👉 
* https://github.com/owncloud/core/pull/30106#issuecomment-384529734 - cannot decline remote shares from the "share-with-you" page - under investigation. 🔍

PVince81:
> The main goal from this point on should be stabilize the PR and merge it the sooner the better, so any non-critical problem should be moved to another ticket in order to fix it once this PR is merged.

agreed

PVince81:
@jvillafanez I agree with your items from: https://github.com/owncloud/core/pull/30106#issuecomment-384547564

I'd leave auto-accept always on for now in any scenarios. I think doing it only on updates is technically tricky and not worth the effort.

please make checkboxes out of them. later after merge, the "move it out" items must have a matching raised ticket mentioned before ticking the checkbox

PVince81:
> Remote shares don't appear by default in the "shares-with-you" view. You need to accept them to make them appear. The behaviour is different of the local shares.

separate issue/enhancement: pending shares list for remote share (there's a ticket for that somewhere)

jvillafanez:
I'm not sure... the problem is that any time you decline the remote share it won't appear for the user anymore.

1. Share a file from server A to B
2. Accept the share in server B
3. Decline the accepted share.

The share file isn't visible any longer in server B, but server A show that it is still shared with server B. The same happens if the share is rejected in step 2.
The problem I see is that the only place to control the remote share is in the notification. Once the notification is gone you can't change the state, you can only decline the share if you've accepted before but nothing else.

I don't see any quick and viable solution to fix this other than implement the missing functionality. The only additional option would be to change the action for the remote shares to "unshare" or "remove share" to match the behaviour the user will see.

PVince81:
As far as I remember if you decline a federated share you cannot accept any more, it's gone. If the original server didn't notice it means the event was not propagated yet. Maybe it's propagated on cron or something. I suggest ignoring this. The behavior will not be consistent at first.
We'll need to think about how to bring pending shares level UX to federated shares as well separately as it will likely have more implications on the technical level and protocol level.

jvillafanez:
> cannot decline remote shares from the "share-with-you" page

This is fixed now by keeping the "unshare" action instead of the accept / reject ones for remote shares. Remote shares should behave now the same way as before while the local shares can be accepted and rejected any time.
The behavior feels cleaner now for remote shares.
(Changes in https://github.com/owncloud/core/pull/30106/commits/6d87646cb06e2022c884d86b6782043263ec7e97)

michaelstingl:
I'd love to see a quick demo after finishing. Would be helpful to get client implementations right.

jvillafanez:
displayname not showing in the notifications fixed accordingly to the agreed rules in https://github.com/owncloud/core/pull/30106/commits/21aec23d5ceef5bd2a0d319be30057c5fce9c279

Note that in case the user is missing and we can't get the displayname, the uid will still be used.

1. User1 shares with User2
2. User2 receives the notification
3. User1 is deleted

The notification for the user2 will use the id of the user1 instead of the displayname. In addition, the notification message might change dynamically depending on the configuration of the extra field value for the shares. This is because we're using the uid of the user in the notification and fetch the user information when we want to render the notification.

jvillafanez:
Fix retarget the share after accept / decline if the default share_folder isn't the root -> https://github.com/owncloud/core/pull/30106/commits/f12c957d4aea881f9d8183ea5b8cc1fef8f08cd8

Note that there are a couple of problems to test it:
* static call in the `deduplicateShareTarget`, which is the method that was causing the issue.
* the test should just check that the target is the same if there is no conflict. I'll check if we can modify the current tests for this.

jvillafanez:
https://github.com/owncloud/core/pull/30106#pullrequestreview-103697631 will need a complete review of core and all the apps regarding that topic. So far I've seen at least the app icons (showing in the top left corner in the header) and the icons for the settings, help and logout actions (top right) missing in such setup. There might be more.

jvillafanez:
This should be mostly ready now

apps/files_sharing/css/sharedfilelist.css
DeepDiver1975: why not reuse the css rule from core? .icon-folder I guess ....

DeepDiver1975:
conflicting ...

apps/files_sharing/css/sharedfilelist.css
jvillafanez: I guess due to consistency with the ".share-state-pending" css rule just above. It would be kind of weird to have pending and rejected state and have one explicit rule for one state but not for the other.
Maybe using another custom icon, not duplicate one of the core, would be the best option

jvillafanez:
Squashed, rebased and updated

jvillafanez:
Minor behaviour change:
1. Accept share
2. Move share to a folder
3. Decline share
4. Accept share again

Previously, it was said that the share will end up back into the root folder after step 4. This has changed, and the share will stay wherever it was before it was declined.
Acceptance test has been adjusted to reflect this.

jvillafanez:
Taking into account the previous comment, I guess we should open an enhancement to show the path where the share is. It seems quite easy to lose track of where the accepted share is if you move them around.

jvillafanez:
* [ ] BUG: Multiple summary lines when going in and out multiple times in the "share with you" and "shared with others" sections.

jvillafanez:
squashed + rebased

davitol:
@jvillafanez @pmaier1 After declining a file, the preview is lost. Intended behavior?

screen shot 2018-05-21 at 22 55 06

screen shot 2018-05-21 at 22 55 17

davitol:
- [ ] Test update from 10.0.5 to this PR with shares accepted/rejected before.

patrickjahns:
I see some comments related to migrations  @PVince81 - please specify together with @owncloud/qa  and prepare tests for it. (First a Testplan is enough?)

PVince81:
Regarding migrations tests, @davitol asked me and the answer was to test the following:
- [ ] TEST: group shares: before upgrade: setup group share and have one recipient refuse a share (unshare from self), after upgrade: check that refused share is still refused. (it can be accepted again in pending shares)
- [ ] TEST: user shares: before upgrade, setup user share and have recipient unshare from self. Share is deleted. After upgrade: share is still deleted, cannot be re-accepted as it was deleted. This is expected behavior

@davitol please tick the boxes

davitol:
**Bug- Corner Case**

1 - userA creates a Folder
2 - userA creates file1 and file2 inside Folder
3 - userA shares with userB Folder and file1
4 - Log in with userB
5 - In shared with you menu decline file1 share
6 - In Files view, browse into Folder and move file1 to 'home'
7 - Browse back to share with you.

**Current behavior** Share with you no longer shows file1 share


@PVince81 @jvillafanez please triage

PVince81:
@jvillafanez please debug this, could be potentially a bigger issue: https://github.com/owncloud/core/pull/30106#issuecomment-391238396

the thing is that the share seems to disappear **because** the file name of the entry in the home is the same as the rejected share

PVince81:
> Consider the behaviour as acceptable. This would mean that we'll only need to remove the share from the oc_share table if the share isn't accesible any longer. I expect this to be an easy solution.

@jvillafanez I think we already have a background job that deletes inaccessible shares. Needs to be confirmed.

davitol:
Testing results tracked in this ticket. https://github.com/owncloud/QA/issues/568 

I think we could merge this PR and then work on the pending stuff: https://github.com/owncloud/QA/issues/568#issuecomment-391242317

PVince81:
there's a conflict to solve, can you have a look @jvillafanez ?

apps/federatedfilesharing/lib/FederatedShareProvider.php
```diff
@@ -955,4 +955,15 @@ public function isIncomingServer2serverShareEnabled() {
 		$result = $this->config->getAppValue('files_sharing', 'incoming_server2server_share_enabled', 'yes');
 		return ($result === 'yes') ? true : false;
 	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function updateForRecipient(IShare $share, $recipient) {
+		/*
+		 * This function does nothing yet as it is just for outgoing
```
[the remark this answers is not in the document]
PVince81: just a comment. a TODO would have "TODO" inside the comment

apps/files_sharing/js/app.js
```diff
@@ -265,6 +265,14 @@ OCA.Sharing.App = {
 				return actions;
 			}
 
+			var targetFileData = context.fileList.elementToFile(context.$file);
+			if (targetFileData.shareLocationType === 'remote') {
+				// accept and reject will be removed for remote shares
```
PVince81: :+1:

lib/private/Share20/Manager.php
```diff
@@ -667,7 +667,7 @@ public function createShare(\OCP\Share\IShare $share) {
 
 		\OC_Hook::emit('OCP\Share', 'post_shared', $postHookData);
 
-		$afterEvent = new GenericEvent(null, ['shareData' => $postHookData, 'shareObject' => $share, 'result' => 'success']);
+		$afterEvent = new GenericEvent(null, ['shareData' => $postHookData, 'shareObject' => $share]);
```
PVince81: @jvillafanez I see you removed events you found were duplicates. I hope we're not breaking anything because the parameters of the other events aren't 100% the same.

We should investigate this, @sharidas any clues about the event duplication ?

PVince81:
- [ ] potential issue: https://github.com/owncloud/core/pull/30106/commits/e006d475a538921cb0ead5137609cd0fe3194651#r190253210

Not sure if we should deduplicate the events in this PR or investigate separately. In the end we only want one set of events. But if said events already existed in a released major version we need to be careful with removal, thinking about potential consumers. As I see the duplicate events don't have exactly the same arguments.

PVince81:
- [ ] another worry about icon path, needs to be tested when deployed to root and non-root: https://github.com/owncloud/core/pull/30106/commits/3fdbcbad9f394d0aacef9b3e28006f0d146569c2#r190254495

PVince81:
Reviews:

- [x] code review from @jvillafanez's commits, done, see comments above
- [ ] code review from @pvince81's commits
- [ ] code review for acceptance tests

jvillafanez:
> @jvillafanez I think we already have a background job that deletes inaccessible shares. Needs to be confirmed.

Not really. The job will delete the shares whose source file is missing. If the source file is deleted, then the job should delete the share. However, the job won't work in our scenario because the file is present but it is in another storage (the fileid is present in the filecache table)

lib/private/Share20/Manager.php
jvillafanez: Having duplicated events is a bug. I'm not sure how symfony handles them, but it's bound to cause issues.

I think now is more consistent because all these events are triggered by core and not some of them by core and others by an app that can be disabled. In addition, the events should send more or less the same data because they're related.

I agree we should investigate if this change is causing problems, but I expect the changes to be in the apps because core should behave properly now.

PVince81: yes, we just need more information about when this was introduced to decide on the strategy

PVince81: - looking at Share20OCS on @sharidas PR and searching for the events: not found https://github.com/owncloud/core/blob/294bc9eb65dd0239dd4ce05fbdabc6585ac7ec37/apps/files_sharing/lib/API/Share20OCS.php the duplicate got removed.
- on master: only one occurrence related to changing the name, no duplicates
- on this PR after my squashed commit and before @jvillafanez's adjustment: duplicate exists

Conclusion: when I rebased an earlier version of this PR onto master I mistakenly kept the deleted events when solving conflicts.

=> this means the commit from @jvillafanez irons this out and makes sense, case closed

jvillafanez:
> another worry about icon path, needs to be tested when deployed to root and non-root

This needs to be rechecked outside of this PR. (https://github.com/owncloud/core/pull/30106#issuecomment-386034450)

PVince81:
@phil-davis can you review the acceptance tests part ?

PVince81:
@individual-it can you rebase and solve the conflict in the test files ?

tests/acceptance/features/webUISharingNotifications/notificationLink.feature
```diff
@@ -0,0 +1,19 @@
+@webUI @insulated @disablePreviews
+Feature: Sharing files and folders with internal users
```
phil-davis: Following links in notifications

phil-davis (commented):
Looks good. The tests themselves cover a good range of combinations of accept/reject/decline shares of files and folders to users and groups.

Just some small documentation things that I noticed.

phil-davis:
@individual-it I made the small review edits in the commit "Acceptance test review minor changes".
There are just 2 questions I have above about some scenario descriptions that do not quite match the detailed scenario steps.

PVince81:
@jvillafanez a conflict crept in... let's solve after the acceptance test review changes are done and before merging.

phil-davis:
There is nothing more that is critical in the acceptance test review. I asked @individual-it about my question above, but I am not sure where he is right now - probably zooming along an autobahn somewhere. So he can answer in "slow" time. 

And "somebody" can resolve the conflict whenever they like.

PVince81:
ok, then let's move forward then. the minor things can be PR'd separately.

@jvillafanez please fix the conflict and this is ready to merge

individual-it:
@phil-davis you are right, there was a mistake in the acceptance tests, fixed that

jvillafanez:
Rebased to fix conflict in the ManagerTest. Note that https://github.com/owncloud/core/pull/30106/commits/7ded301557724525082b6d661d84bacee5590f57 has been added due to a bad conflict resolution (probably) since that piece wasn't there before the rebase.

PVince81:
@jvillafanez seems you discarded @individual-it's commit, likely because you guys updated at the same time. @individual-it can you push it again ?

PVince81:
commit was 66584e1

tests/acceptance/features/webUISharingInternalUsers/acceptShares.feature
```diff
@@ -0,0 +1,234 @@
...
+		And user "user2" has shared file "/testimage.jpg" with user "user1"
+		And the user has logged in with username "user1" and password "1234" using the webUI
+		And the user declines the share "simple-folder" offered by user "User Two" using the webUI
+		When the user accepts the share "simple-folder" offered by user "User Two" using the webUI
+		Then the folder "simple-folder (2)" should be in state "" in the shared-with-you page on the webUI
+		And the file "testimage.jpg" should be in state "Pending" in the shared-with-you page on the webUI
+		And the folder "simple-folder (2)" should be listed in the files page on the webUI
+		And the file "testimage (2).jpg" should not be listed in the files page on the webUI
+
+	Scenario: accept a share that you received as user and as group member
```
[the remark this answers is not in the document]
phil-davis: fixed

phil-davis:
I made the same changes in a new commit, rebased and pushed again. All is good with acceptance tests.

PVince81:
:tada: fantastic work, guys ! :tada: 

phil-davis:
Backport ``stable10`` #31613

PR #30101, opened 2018-01-11
Extend fallback scrollContainer object

lordelix:
## Description

Opening the files action menu in public links scrolls view to show the whole menu.

## Related Issue

Fixes https://github.com/owncloud/core/issues/30064

## How Has This Been Tested?

Visual test in latest IE, Edge, FF und Chrome

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30109, opened 2018-01-12
Scroll action menu on Files page and Public links page

phil-davis:
Modification to PR #30101 

``content-wrapper`` contains ``app-content`` on the ordinary files page. Putting both of them in the ``scrollContainer`` selection broke scroll into view on the ordinary files page.

The 2nd commit here uses ``app-content`` if it exists (e.g. on "ordinary" files pages), otherwise it uses ``content-wrapper`` which seems to be the thing to use on the public links page.

This stuff really needs basic scrolling actions tests for each of the variations of files page (e.g. is Trashbin or Favourites or... done with some other id(s) and breaks in some other way?)

We will see if Travis UI tests pass here (that shows that the ordinary files page is scrolling OK).

### DIFF
--- a/core/js/js.js
+++ b/core/js/js.js
@@ -2050,7 +2050,11 @@ OC.Util = {
 		var windowHeight = $(window).height();
 
 		if (scrollContainer === null) {
-			scrollContainer = $('#content-wrapper, #app-content');
+			if ($('#app-content').length) {
+				scrollContainer = $('#app-content');
+			} else {
+				scrollContainer = $('#content-wrapper');
+			}
 		}
 
 		if (toViewElBottomLocation > windowHeight) {
--- a/core/js/js.js
+++ b/core/js/js.js
@@ -1769,7 +1769,7 @@ function relative_modified_date(timestamp) {
 OC.Util = {
 	// TODO: remove original functions from global namespace
 	humanFileSize: humanFileSize,
-	
+
 	/**
 	* regular expression to parse size in bytes from a humanly readable string
 	* see computerFileSize(string)
@@ -1792,7 +1792,7 @@ OC.Util = {
 
 		var s = string.toLowerCase().trim();
 		var bytes = null;
-		
+
 		var bytesArray = {
 			'b': 1,
 			'k': 1024,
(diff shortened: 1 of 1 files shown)


PVince81:
what's next ? reviews ?

phil-davis:
The last commit here was a suggestion to @felixheidecke about a way to approach fixing side-effects in #30101 
See my comment https://github.com/owncloud/core/pull/30101#issuecomment-357111271

@felixheidecke made a later comment https://github.com/owncloud/core/pull/30101#issuecomment-357657794

Since then I have seen no action. Either the last commit here can be cherry-picked into #30101 or some other code can be done to resolve #30101

core/js/js.js
```diff
@@ -2012,21 +2012,25 @@ OC.Util = {
...
+
 		var toViewElTopLocation = toViewEl.offset().top;
 		var toViewElHeight = toViewEl.outerHeight();
 		var toViewElBottomLocation = toViewElTopLocation + toViewElHeight + 50;
 		var windowHeight = $(window).height();
-		
+
 		if (scrollContainer === null) {
-			scrollContainer = $('#app-content');
+			if ($('#app-content').length) {
```
PVince81: normally the scroll container is supposed to be given in the constructor and stored in `this.$container`. not sure why it wasn't used here.

@phil-davis can you test ?

phil-davis: places like https://github.com/owncloud/core/blob/master/apps/files/js/fileactionsmenu.js#L149 call ``scrollIntoView()`` with ``scrollContainer`` null. They expect ``scrollIntoView()`` to "guess" the right ``scrollContainer`` In the case of different file lists, they are not all wrapped inside ``app-content``

So this hack here from me just makes a priority list for this guess - if there is an ``app-content`` container existing, use it (existing behavior), otherwise use ``content-wrapper`` (and hope that one of those always exists).

This strategy works for the pages that I played with.

It was just a "quick" suggestion from me as a way resolve this, so I am very happy to hear from real JS people about a better way.

PVince81: the better way is likely to use "this.$container" (or self.$container if inside a closure) which is supposed to be the correct container as registered during list creation

if that doesn't work I'll need to dig deeper

PVince81:
@phil-davis see my comment above. if this doesn't work or doesn't provide much value, let's close this PR

phil-davis:
Rebased just now. I think it still works. I will look again after CI has run.

phil-davis:
Rebased again to get more current CI.

core/js/js.js
phil-davis: @PVince81 `this.container` `this$container` does not work - the scrolling behaves like described in the issue.
Actually I have no real idea  what I am doing, I made this bit of code and observation that it works just to give @felixheidecke a demonstration of something that "works for me". I was hoping that he would either take this code or design a "nice" solution.

Let me know if there is anything else I can do to progress this.

DeepDiver1975:

Issue #29984, opened 2017-12-28
Group name containing / can't be deleted


sushilkshetri:
### Steps to reproduce
1. Login as admin.
2. Create a group with name containing "/".
3. Delete the group using webUI .
4. Reload the page.

### Expected behaviour
Group with the name containing "/" should be deleted.

### Actual behaviour
It's not deleted.

### Server configuration
**Operating system**: Ubuntu desktop 17.10

**Web server:** Apache2

**Database:** Sqlite3

**PHP version:** PHP 7.1

**ownCloud version:** Owncloud 10.0.4

**Updated from an older ownCloud or fresh install:** Fresh install

**Where did you install ownCloud from:** git

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc0nm780itvl",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "default-language": "en",
        "trusted_domains": [
            "localhost",
            "192.168.0.116"
        ],
        "datadirectory": "\/home\/sushil\/www\/html\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/localhost\/owncloud-core",
        "dbtype": "sqlite3",
        "version": "10.0.4.4",
        "logtimezone": "UTC",
        "installed": true,
        "csrf.disabled": true,
        "skeletondirectory": "\/home\/sushil\/www\/html\/owncloud-core\/tests\/ui\/skeleton",
        "loglevel": 0
    }
}

```

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - dav: 0.3.2
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_sharing: 0.10.1
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - testing: 0.1.0
  - updatenotification: 0.2.1
Disabled:
  - theme-example

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
local
**Are you using encryption:** yes/no
no
**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
no

### Client configuration
**Browser:** Chrome 63.0 and firefox 57

**Operating system:**
 Ubuntu 17.10
#### Web server error log
Nothing

[...]

PR #30115, opened 2018-01-12
UI test to delete groups containing special URL chars

phil-davis:
## Description
Test that group names containing characters that are common delimiters in URLs can be deleted.

## Related Issue
#29984

## Motivation and Context
Have a tes for the issue so that there is no future regression.

## How Has This Been Tested?
UI test has been run locally.
(note that the test is skipped on version up to 10.0.5 because the fix has missed being included in 10.0.5)

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/other/managingGroups.feature
+++ b/tests/ui/features/other/managingGroups.feature
@@ -38,7 +38,46 @@ So that access to resources can be controlled more effectively
 		|groupname|
 		|0        |
 		|false    |
-	
+
+	@skipOnOcV10.0.3 @skipOnOcV10.0.4 @skipOnOcV10.0.5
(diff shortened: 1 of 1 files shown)


phil-davis:
A couple of notes about skipping tests on particular ownCloud versions:
1) We need to mention each version since we started doing this (e.g. 10.0.3 10.0.4 10.0.5) so that the whole test suite can be run against an install of each of those old versions where the test will fail - that is going to get annoying in a few versions time.
2) It would be nice to regularly bump the ownCloud version (in master and stable10) soon after each formal release. That will allow the test code to understand that git code is "more advanced" than the released version.

individual-it (approved):

phil-davis:
Backport stable10 #30138

PR #30118, opened 2018-01-12
Add symfony events for comment app

sharidas:
Addy symfony events for comment app.
The events include for
1) create
2) update
3) delete
activity for comments

## Description

Adding symfony events for create, update and delete events for comment.

## Motivation and Context

Adding symfony events for create, update and delete events for comment.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Comments/Manager.php
+++ b/lib/private/Comments/Manager.php
@@ -28,6 +28,7 @@ use OCP\Comments\IComment;
 use OCP\Comments\ICommentsManager;
 use OCP\Comments\NotFoundException;
 use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\Events\EventEmitterTrait;
 use OCP\IDBConnection;
 use OCP\IConfig;
 use OCP\ILogger;
@@ -36,6 +37,7 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 class Manager implements ICommentsManager {
 
+	use EventEmitterTrait;
 	/** @var  IDBConnection */
 	protected $dbConn;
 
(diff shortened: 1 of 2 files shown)


apps/comments/lib/Dav/CommentNode.php
```diff
@@ -129,11 +131,16 @@ protected function checkWriteAccessOnComment() {
...
+	 * @return bool
 	 */
 	function delete() {
-		$this->checkWriteAccessOnComment();
-		$this->commentsManager->delete($this->comment->getId());
+		return $this->emittingCall(function () {
+			$this->checkWriteAccessOnComment();
+			return $this->commentsManager->delete($this->comment->getId());
+		}, [
+			'before' => ['objectId' => $this->comment->getObjectId(), 'message' => $this->comment->getMessage(), 'objectType' => $this->comment->getObjectType()],
```
PVince81: We should likely do this inside the CommentsManager instead. What if an app creates an automated comment from inside the PHP code ? It would use the PHP API by using CommentsManager.

PVince81: you'll likely find some existing events there related to activities

sharidas: Thanks for the pointer. I have updated it at comments/Manager.php

sharidas:
Updated the new change set to Manager.php of comment. So now if the API call is made by any app for comment, the new events written should trigger.

lib/private/Comments/Manager.php
```diff
@@ -460,11 +462,16 @@ public function getNumberOfCommentsForObject($objectType, $objectId, \DateTime $
...
-			->setObject($objectType, $objectId);
-		return $comment;
+		return $this->emittingCall(function () use (&$actorType, &$actorId, &$objectType, &$objectId) {
+			$comment = new Comment();
+			$comment
+				->setActor($actorType, $actorId)
+				->setObject($objectType, $objectId);
+			return $comment;
+		}, [
+			'before' => ['objectType' => $objectType, 'objectId' => $objectId, 'actorType' => $actorType, 'actorId' => $actorId],
```
PVince81: I think we should keep the message in the object (did you remove it, I don't remember whether it was there before), because some apps could decide to do something useful with it. (ex: activity app, but that one currently uses old style hooks)

PVince81: Wait, I think we should not emit a message here. This method is just a helper to create a `Comment` object but is not actually saving it to the database at this point. No need for an event for that.

lib/private/Comments/Manager.php
```diff
@@ -479,38 +486,43 @@ public function create($actorType, $actorId, $objectType, $objectId) {
...
+					CommentsEvent::EVENT_DELETE,
+					$comment
+				));
+			}
 
-		return ($affectedRows > 0);
+			return ($affectedRows > 0);
+		}, [
+			'before' => ['commentId' => $id],
+			'after' => ['commentId' => $id, 'message' => $this->get($id)->getMessage(), 'objectId' => $this->get($id)->getObjectId()]
```
PVince81: let's not include the message here, it's a delete operation after all and I don't think it's worth doing this additional DB lookup `->get()` for an event that might not even be processed

lib/private/Comments/Manager.php
```diff
@@ -530,21 +542,28 @@ public function delete($id) {
...
-		}
+				);
+				$this->cache($comment);
+			}
 
-		return $result;
+			return $result;
+		}, [
+			'before' => ['objectId' => $comment->getObjectId(), 'commentId' => $comment->getId(), 'message' => $comment->getMessage(), 'status' => $createOrUpdate],
+			'after' => ['objectId' => $comment->getObjectId(), 'commentId' => $comment->getId() , 'message' => $comment->getMessage(), 'status' => $createOrUpdate]
```
PVince81: yes :+1:

PVince81: no need for "status" attribute though if the event name itself is already beforeCreate or beforeUpdate

sharidas: The "status" attribute would help us know whether user is creating ( first time ) a comment. Or updating the existing comment.

PVince81: why not sepate events like "create" and "update" ? for file events we also have different events for "create" and "update" ? (or maybe I remember it wrong)

sharidas: Save method indirectly does the creating a new comment by calling https://github.com/owncloud/core/blob/master/lib/private/Comments/Manager.php#L556 and updating existing comment using https://github.com/owncloud/core/blob/master/lib/private/Comments/Manager.php#L594. So I thought to pass attribute "status" with value "create" or "update" which can be used by the apps or any listener of the events and use the event.

While writing this comment, I got another approach to create events in the insert method and update method. insert method will have a "create" event and update method will have "update" event.

sharidas:
I have updated the PR by creating "create" and "update" events. And there is no more attribute "status" which holds the "create" and "update" data.

PVince81:
Thanks. Please check test failures

sharidas:
Updated the tests. It passed locally in my machine.

PVince81 (approved):
:+1:

sharidas:
Backport PR: https://github.com/owncloud/core/pull/30142

PR #30088, opened 2018-01-10
Bundle all vendor libs together ....

DeepDiver1975:
## Description
Use webpack to bundle vendor libs into one bundle.js

## Related Issue
@PVince81 I don't find the related issue anymore.

## Motivation and Context
- load less files
- reduce tar ball size
- simplify release script

## How Has This Been Tested?
- make test-js
- manual web ui testing

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30121, opened 2018-01-12
Check for yarn only when needed for selected build rule

PVince81:
When running non-JS rules like "make test-php", we should not require
yarn.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,10 +35,6 @@ SHELL=/bin/bash
 # Define YARN and check if it is available on the system.
 #
 YARN := $(shell command -v yarn 2> /dev/null)
-ifndef YARN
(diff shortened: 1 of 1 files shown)


PVince81:
To test: run `make test-php` on an env without yarn: test runs.
If you run `make test-js` it will require it.

tboerger (approved):

DeepDiver1975 (approved):

PVince81:
I think we decided to not backport yarn any more ? removing backport label

also we still had some troubles with yarn, see https://github.com/owncloud/core/pull/30088

phil-davis:
Backport `stable10` #34793

`yarn` was backported to `stable10` in PR #33665 - so now this PR can also be backported.

PR #30143, opened 2018-01-15
Use mysql:5.7 for coverage instead of mariadb until issues are solved

DeepDiver1975:
we are facing random failure on drone with mariadb - until sorted we switch over to mysql.


tboerger (approved):
If this helps, why not!

PR #30113, opened 2018-01-12
Bugfix/30078

DeepDiver1975:
## Description
Outside of the CardDAV collection we shall not change the file content.

In addition:
 - also only load calendar plugins when inside the calendar collection
 - fix avatar output - jpg was putput for png as well

## Related Issue
fixes #30078 

## How Has This Been Tested?
- manual testing
- unit tests added for avatar

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30149, opened 2018-01-16
[stable10] Bugfix/30078

DeepDiver1975:
backport #30113


DeepDiver1975:
The fix is not complete - calendar and addressbook plugins need to be registered for e.g. principal calls as well.

DeepDiver1975:
We can revive this as soon as we back port https://github.com/owncloud/core/pull/30162

PVince81 (approved):
:+1:

PR #30158, opened 2018-01-16
Don't overwrite quota with null

butonic:
user_ldap implements the IProvidesQuota interface but returns null if neither a default nor an individual quota has been set. In this case any manual quote configured by the admin gets overwritten.

This PR fixes that.


butonic:
superseded by https://github.com/owncloud/core/pull/29669

PR #30161, opened 2018-01-16
[stable10] Fix failure of shares which are already moved with transfe…

sharidas:
…r-ownership

If the folder and files inside the folder are already moved
by the command, and if there is a public link inside the folder
then during the restore share operation the command throws
exception. This fix tries to check if the file/folder is already
moved. If so then there is no need to do call updateShare.

## Description

When there are multiple shares during the transfer-ownership command, and if the share is already moved to the destination user, then the command would throw an exception if the share inside the folder is handled separately.

## Motivation and Context

Lets say we have `user1` and `user2`. `user1` creates folder `foo` and shares with `user2`. Then `user2` tries to creates a public link of one of the files inside `foo`. Now run the transfer-ownership command with source user as `user1` and destination user as `user2`. An exception is thrown. The reason I found is if the folder is already moved, and in the restore share checks for the share of public share and run the updateShare causes exception. The file is not available. So this change checks if the file/folder is already there in the destination folder. If so then nothing to do.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -302,6 +302,18 @@ class TransferOwnership extends Command {
 					if ($share->getSharedBy() === $this->sourceUser) {
 						$share->setSharedBy($this->destinationUser);
 					}
+					/*
+					 * If the share is already moved then updateShare would cause exception
+					 * This can happen if the folder is shared and file(s) inside the folder
+					 * has shares, for example public link
+					 */
+					if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+						$sharePath = ltrim($share->getNode()->getPath(), '/');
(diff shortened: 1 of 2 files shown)


sharidas:
Orignal PR: https://github.com/owncloud/core/pull/30008

PVince81:
Weird how codecov failed on master but not on this backport PR ?!

PVince81:
Is this up to date with https://github.com/owncloud/core/pull/30008 ?

sharidas:
Yes the code looks similar to #30008

PVince81:
similar or exactly the same ? in the original PR it seems you made some modifications. please confirm

sharidas:
Took the diff of raw files of TransferOwnership.php and transfer-ownership.feature from this PR and #30008. The diff command returns empty. So which means they are exactly the same.

PVince81:
Link share is lost: https://github.com/owncloud/core/issues/31150

PR #30163, opened 2018-01-16
[stable10] Fix deleted items auto expiration for users with no quota

VicDeo:
Backport of https://github.com/owncloud/core/pull/30095

## Description
Fix regression introduced a long time ago in https://github.com/owncloud/core/commit/a3999036f720e73727dac07a337578f3e25f92aa#diff-0dc914768208fc24b2aa2f2a369bb684
It is based on a wrong assumption that FS free space could be negative. 

I just reverted differencies for users with/without quota introduced in https://github.com/owncloud/core/commit/a3999036f720e73727dac07a337578f3e25f92aa#diff-0dc914768208fc24b2aa2f2a369bb684, moved the method `calculateFreeSpace` into another non-static class and tried to avoid legacy classes in it where possible.

As soon as it will reintroduce purging of deleted files described in #2936
the internal constant is available as `trashbin_purge_limit` config.php option now.
It could be set to 30%, 20%, 10%, etc in case there are no users with hard quota and default 50% of free space for deleted files is not suitable for the current setup.

## Related Issue
https://github.com/owncloud/core/issues/27412
obsoletes https://github.com/owncloud/core/pull/29951
## Motivation and Context
Trashbin auto clean up doesn't work when a user has no quota

## How Has This Been Tested?
1. Set `trashbin_retention_obligation` to `auto`
2. Create a user with no quota
3. Upload some files and delete uploaded files for that user
4. Keep uploading until all partition for `datadir` is occupied
5. Run cron.php for expiration (added by @PVince81)

### Expected 
Deleted files are purged to save space

### Actual
They are not

## Types of changes

## Checklist:

To document: 
New config.php `'trashbin_purge_limit'` that was just a constant before.
Introduced to fine tune purging in cases like https://github.com/owncloud/core/issues/2936

### DIFF
--- a/apps/files_trashbin/lib/Quota.php
+++ b/apps/files_trashbin/lib/Quota.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Trashbin;
+
+use OC\Files\Filesystem;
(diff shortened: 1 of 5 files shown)


PVince81:
@ownclouders please rebase

PVince81 (approved):
:+1:

PVince81:
restarted build

apps/files_trashbin/lib/Quota.php
```diff
@@ -0,0 +1,105 @@
...
+	/**
+	 * Get user quota or free space when there is no quota set
+	 *
+	 * @param IUser $user
+	 * @return int|mixed
+	 */
+	protected function getUserQuota(IUser $user) {
+		$quota = \OC_Util::getUserQuota($user);
+		if ($quota === FileInfo::SPACE_UNLIMITED) {
+			$quota = Filesystem::free_space('/');
```
PVince81: I wonder if we should rather take this from the user's home in case the home is on a different partition.

The reason is because there are often complains that OC takes free space from the wrong partition, but said reports were for the quota display, not trashbin

PR #29815, opened 2017-12-11
[WIP] Introduce uikit

PVince81:
- [x] Makefile
- [ ] Fix core styling to not be destroyed by uikit side-effects
- [x] Basic owncloud uikit theme
- [x] Provide mechanism to compile basic owncloud uikit theme
- [ ] Provide mechanism to compile and load any uikit theme from theme app (idea: OC_Defaults::getMainCss() to load the theme's uikit CSS instead of the default one)

PR #30166, opened 2018-01-16
Introduce uikit (retry)

PVince81:
Note: contains https://github.com/owncloud/core/pull/30088 and will need rebase once the latter is merged.

Supersedes https://github.com/owncloud/core/pull/29815


PVince81:
remaining TODOs:
- [ ] Fix core styling to not be destroyed by uikit side-effects
- [ ] Provide mechanism to compile and load any uikit theme from theme app (idea: OC_Defaults::getMainCss() to load the theme's uikit CSS instead of the default one)
- [ ] if theme app enabled, don't load the UIKit owncloud.css (or is there a way to override)
- [ ] future: think of a way to make it easy for apps to compile and deploy their uikit less file to css

build/entry.js
```diff
@@ -0,0 +1,44 @@
+var $ = require('jquery');
+window.jQuery = $;
```
felixheidecke: Why are all those libs bound to the window object instead of some OCL(ib) class? Not sure if this is the right approach 🤔 … am I missing the point here?

PVince81: this is a hack that @DeepDiver1975 used to make those libs available in the global namespace for backward compatibility as we don't have time to rewrite all the OC JS code to use import statements for those.

See https://github.com/owncloud/core/pull/30088 and comment there if you have a better approach

PVince81:
If we go for separate frontend then eventually core will have no frontend code and more so UIKit should only be provided for Phoenix, not core.

Also not to be backported to 10.0.

However we need to recover all changes from this PR that help solving UIKit conflicts for https://github.com/owncloud/core/issues/30778

PVince81:
@felixheidecke I remember you said introducing UIkit is too complicated due to side effects and would resort to manually writing the CSS when porting the Phoenix views to stable10 ? If that's the case we can close this PR

Issue #29526, opened 2017-11-09
Split "Enforce password protection" option for file drop and public links


pmaier1:
To fulfill the requirement of enforcing password protection for public links while not enforcing them for file drop links I suggest to split the "Enforce password protection" option into two:
a) Enforce password protection for regular public links
b) Enforce password protection for File Drop public links

Wording might not yet be perfect.

Another thing we need to take into account is the UI behavior when only one of the options is true:

Example: 
- User creates a file drop public link without password and changes permissions to a regular public link where password protection is enforced. At that point saving needs to be blocked unless a (policy-matching) password is entered. This of course needs to work the other way round as well
- What if I have a password protection enforced public link and change permissions to a not password protection enforced file drop public link? Password will be kept but can be removed by the user!?

I hope there are no implications I might have missed.

@PVince81

pmaier1:
Actually I don't really see a need for password-protecting File Drop links so we could also just exclude them in order not to clutter config options even more. I would make it dependent on effort.

PVince81:
@pmaier1 some people might feel safer if there is a password to File Drop links in case the link gets leaked and people would spam the folder. This, assuming that the password was shared using different channels. Probably a rare case...

felixboehm:
@jvillafanez please contact @felixheidecke if any frontend help needed.

PVince81:
I think currently we trigger an event when password policy check needs to apply. We might want to add a "scope" or "context" attribute there to let password policy apps find out in what context the password will be used.

jvillafanez:
I'll need some clarifications...

![screen shot 2018-01-12 at 15 34 44](https://user-images.githubusercontent.com/1477829/34879647-35a48d72-f7ae-11e7-91f1-f49b67043040.png)

That sharing part, is it controlled by the files_sharing app or by core? If it's controlled by the files_sharing app, we should clarify that the password protection there is just for the files_sharing app and not for any other.

I expect the files_drop app has a similar checkbox in its own configuration panel.

My point is that core doesn't enforce anything, it's each app which does it, and each app has different ways to enforce password protection. If this is true, we can just adjust the text there for the files_sharing and duplicate the functionality (more or less) for files_drop.

PVince81:
@jvillafanez files_drop app is dead. When we talk about it we mostly mean its replacement: go to the link share creation dialog and choose "upload only" mode. In upload only mode, one should be able to disable password enforcement as often times passwords for upload only don't make much sense as there is no data you can access.

[...]

PR #30168, opened 2018-01-17
Disable enforce password upload only

jvillafanez:
## Description
Add option to disable password enforcement for write-only shares

## Related Issue
https://github.com/owncloud/core/issues/29526

## How Has This Been Tested?
Manually tested

Known problems (Jan, 17th):
* Adding these changes to share 1.0 isn't straightforward. We'll likely need to create another function to get the share data (such function doesn't exists now) in order to get the share permissions.
* Share dialog has a problem when the share mode (read-only, read & write, write-only) is switched. Although the password isn't enforced for write-only links, the dialog will show an error when create the share. The share is properly created anyway, so this is a visual problem. The password requirement seems to be determined when the template is loaded, not dinamically. I think other scenarios work fine due to the exception coming from the server.

## Types of changes

## Checklist:

@felixheidecke for both checking the styling in the settings page (being a dependent option of the enforce password one doesn't seem a good idea to be on the same level, but I'd prefer not to touch css if possible) and also for the sharing dialog fix, or whatever we want to do with it (check the known issues above).

### DIFF
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -304,13 +304,17 @@ class OC_Util {
 	}
 
 	/**
-	 * check if a password is required for each public link
+	 * check if a password is required for each public link.
+	 * This is deprecated due to not reflecting all the possibilities now. Falling back to
+	 * enforce password for read-only links. Note that read & write or write-only options won't
+	 * be considered here
 	 *
 	 * @return boolean
+	 * @deprecated
 	 */
 	public static function isPublicLinkPasswordRequired() {
 		$appConfig = \OC::$server->getAppConfig();
-		$enforcePassword = $appConfig->getValue('core', 'shareapi_enforce_links_password', 'no');
+		$enforcePassword = $appConfig->getValue('core', 'shareapi_enforce_links_password_read_only', 'no');
 		return ($enforcePassword === 'yes') ? true : false;
 	}
 
--- a/lib/public/Share/IManager.php
+++ b/lib/public/Share/IManager.php
@@ -116,7 +116,7 @@ interface IManager {
 	/**
 	 * Get shares shared with $userId for specified share types.
 	 * Filter by $node if provided
-	 * 
+	 *
 	 * @param string $userId
 	 * @param int[] $shareTypes - ref \OC\Share\Constants[]
 	 * @param Node|null $node
@@ -235,6 +235,14 @@ interface IManager {
 	 * @since 9.0.0
 	 */
 	public function shareApiLinkEnforcePassword();
+
+	/**
+	 * Is password enforced for upload-only shares?
+	 *
+	 * @return bool true -> password isn't enforce for upload-only share, false otherwise
+	 * @since 10.0.6
+	 */
+	public function shareApiLinkEnforcePasswordDisabledForUploads();
 
 	/**
 	 * Is default expire date enabled
--- a/lib/public/Share/IManager.php
+++ b/lib/public/Share/IManager.php
@@ -230,19 +230,38 @@ interface IManager {
 
 	/**
 	 * Is password on public link requires
+	 * NOTE: This method is deprecated and will fallback to the "shareApiLinkEnforcePasswordReadOnly"
 	 *
 	 * @return bool
 	 * @since 9.0.0
+	 * @see IManager::shareApiLinkEnforcePasswordReadOnly()
+	 * @deprecated
 	 */
 	public function shareApiLinkEnforcePassword();
 
 	/**
-	 * Is password enforced for upload-only shares?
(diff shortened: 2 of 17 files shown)


jvillafanez:
https://github.com/owncloud/core/pull/30168/commits/cdb35b5b4411e5f9abd0e73e74adedd0a1f3a32f should fix the problem with the dialog, although we've removed the "field required" mark for the password. Not sure if we want to keep it or not. The behaviour seems better now.

jvillafanez:
This is ready for review

jvillafanez:
This PR isn't for enforcing password in write-ony shares, is for disabling enforcement in write-only shares.

felixheidecke:
From a UX POV it's not the nicest to have a negating option amongst approving options.
My suggestion

- [] Allow public uploads
- [] Enforce password protection on read shares
- [] Enforce password protection on write-only shares
- [] Set default expiration date
- [] Allow users to send mail notification for shared files
- [] Allow users to share file via social media

jvillafanez:
Modified to use the 3 options.

Things still to do:
* [ ] migration step (old "shareapi_enforce_links_password" config option removed / deprecated)
* [ ] Share1.0 adjustment, if any (very likely)

**Important notes:**
* Null passwords will now be checked by 3rd parties as part of the password verification process of the share link. This means that 3rd parties can reject empty passwords.
* Old `shareApiLinkEnforcePassword` is deprecated and will fallback to the `shareApiLinkEnforcePasswordReadOnly` value. The configuration value will be removed during the migration step (to be done)

jvillafanez:
* [ ] deprecate `OCP\Util::isPublicLinkPasswordRequired()`, fallback to new key for read-only
* [ ] remove that call from the settings/Panel/Helper
* [ ] remove enforcePasswordForPublicLink from config.js and related files

jvillafanez:
This is ready to review

lib/public/Share/IManager.php
```diff
@@ -230,12 +230,39 @@ public function shareApiAllowLinks();
...
+	 * @see IManager::shareApiLinkEnforcePasswordReadOnly()
+	 * @deprecated
 	 */
 	public function shareApiLinkEnforcePassword();
 
+	/**
+	 * Is password enforced for read-only shares?
+	 *
+	 * @return bool true if password is enforced, false otherwise
+	 * @since 10.0.6
```
PVince81: make that 10.0.7 now

lib/private/legacy/util.php
```diff
@@ -303,13 +303,17 @@ function ($mountPoint, $storage) use ($user) {
...
+	 * enforce password for read-only links. Note that read & write or write-only options won't
+	 * be considered here
 	 *
 	 * @return boolean
+	 * @deprecated
 	 */
 	public static function isPublicLinkPasswordRequired() {
 		$appConfig = \OC::$server->getAppConfig();
-		$enforcePassword = $appConfig->getValue('core', 'shareapi_enforce_links_password', 'no');
+		$enforcePassword = $appConfig->getValue('core', 'shareapi_enforce_links_password_read_only', 'no');
```
PVince81: Should we return `true` if at least one of the three settings is `true` ? Or the opposite, only return 'true` if all three are `true` at the same time ?

Did you find any existing consumers for this old API ?

PVince81 (commented):
Code looks good otherwise

PVince81:
what is this Codacy dude doing here again ? I thought we removed it ? @DeepDiver1975

lib/private/legacy/util.php
jvillafanez: I'd use either the read-only or the read & write config keys, which I think they're the closest options we have now that match the old expectation. Having a mix of the keys won't honor the expectation from my point of view.

> Did you find any existing consumers for this old API ?

Only the old sharing. The rest of the core should have been patched in this branch.
```
grep -rl isPublicLinkPasswordRequired
lib/private/Share/Share.php
lib/private/legacy/util.php
lib/public/Util.php
```
I haven't checked any apps outside of core. I don't expect apps to use this anyway (just files_sharing and maybe some related app), and the fact that I added the options is in order to follow what is written.

PVince81 (approved):
Fair enough :+1:

PVince81:
@ownclouders rebase

ownclouders:

PVince81:
- [ ] change since to 10.0.8 due to hotfix 🙈

ownclouders:

jvillafanez:
Updated

PVince81:
restarted automation test build

PVince81:
@jvillafanez please backport this feature to stable10

PVince81:
stable10: https://github.com/owncloud/core/pull/30408

Issue #30169, opened 2018-01-17
[WIP] Rework authentication code


butonic:
In deployments that allow passwerd/ldap based login and SSO via SAML we currently recommend to set up oc with password based login under /owncloud and SAML protected login under /oc-shib. That is bullshit because links generated when accessing the site via saml will contain the /oc-shib part, which prevents access to the share link without going through SAML. Setting `overwrite.cli.url` does not fix that (and is not intended for that).

I would be great to expose different factors under different urls, eg:
* `/login` enter only the login name. oc can then decide which factor will be used to authenticate. the admin may dafine an auth chain or tha user may add a factors. That is deployment specific
* `/login/pwd` for the default password based auth
* `/login/saml` for saml based auth
* `/login/mail` for sending an email to the user with a login link
* `/login/totp` for [RFC6238: A Time-Based One-Time Password Algorithm](https://tools.ietf.org/html/rfc6238)
* `/login/hotp` for [RFC4226: An HMAC-Based One-Time Password Algorithm](https://tools.ietf.org/html/rfc4226)
* `/login/tiqr` for [tiqr based auth](https://tiqr.org/)
* `/login/sms` for an sms to the users phone, [yes it is insecure but still better than no second factor](https://www.howtogeek.com/310418/why-you-shouldnt-use-sms-for-two-factor-authentication/)
* `/login/kerb` to require a kerberos token
* `/login/cert` to require a client certificate
...

We could extend the existing `OCP\Authentication\TwoFactorAuth\IProvider2`, currently using urls like `/login/challenge/totp`

Instead of checking if twofactor is enabled I propose to look at how pam allows configuring the auth.
A login via saml or pwd & sms could be configured like this:
```php
'auth.factors' => [ // order is important ... or we store this inside the db
  'pwd' => 'required',
  'sms' => 'required',
  'saml' => 'sufficient'
],
```
After entering the login on /login the user will be redirected to `/login/challenge/pwd` to enter his password, then to `/login/challenge/sms` to enter a pin that was send via sms. The `/login` route can detect if a saml session is present and redirect the user to `/login/challenge/saml` which is a saml protected endpoint. This might even be done by apache to reflect the desired auth behavior (eg saml in the internal network, pwd+sms from outside)

Login via saml or password like this:
```php
'auth.factors' => [
  'pwd' => 'sufficient',
  'saml' => 'sufficient'
],
```

[...]

PR #30172, opened 2018-01-17
introduce additional login endpoints

butonic:
a hack to solve unaccessible share links until https://github.com/owncloud/core/issues/30169 has been fixed properly

cc @dercorn


butonic:
should be fixed by giving the app its own auth endpoint, see https://github.com/owncloud/core/issues/30169#issuecomment-378505066

A firewall or apache needs to decide anyway if and where a user is redirected if the login page is requested.

PR #29865, opened 2017-12-15
Filter public link share data

settermjd:
## Description

Filter data passed to the script that emails a user about a public link share.

## Related Issue

https://github.com/owncloud/security-tracker/issues/255

## Motivation and Context

It closes an HTML injection vulnerability.

## How Has This Been Tested?

There is a limited ability to suitably test `/ocs/ajax/share.php`. Given that, I created tests for the class that I developed. Then, I manually tested the changes via the browser.

## Screenshots (if appropriate):

Relevant screenshots are contained in the related issue.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #30176, opened 2018-01-17
Backport filter link share email to Stable9.1 

settermjd:
Backport #29865 to stable9.1

### DIFF
--- a/core/ajax/share.php
+++ b/core/ajax/share.php
@@ -34,6 +34,7 @@
  *
  */
 
+use OC\Share\Filters\MailNotificationFilter;
 use OCP\IUser;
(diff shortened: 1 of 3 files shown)


DeepDiver1975:
I'll help out in updating the 3rdparty libs

DeepDiver1975:
requires https://github.com/owncloud/3rdparty/pull/289 to be merged first

DeepDiver1975:
@settermjd we have the trouble that there is no zend lib version available which fulfills the full range of php versions we support. Since 9.1 is EOL after the next release I suggest to not waste any further time in backporting.

Issue #26922, opened 2017-01-11
IE11 caches AJAX requests


butonic:
### Steps to reproduce
1. In php.ini set `session.cache_limiter=` (emptystring) to disable sending any cache headers, oc sends them itself. if using this option PHP generates Pragma headers which may contradict what the Cache-Control headers used by ownCloud may say.
2. click on the share link checkbox

### Expected behaviour
see the generated link after the request has been received

### Actual behaviour
see a spinner ... forever. refreshing the page should show the link in the list ... check the network requests (IE11 has an F12 like chrome or FF) ... see that it caches ajax get requests

### Server configuration
**Operating system**: Win 10 & Win 7

**Web server:** Apache

**Database:** Mysql

**PHP version:** 5.6 & 7.0

**ownCloud version:** (see ownCloud admin page) 9.1.2 & master

### Additional info
see https://github.com/owncloud/core/issues/26922#issuecomment-271977658 for an example of cached requests

PVince81:
Broken only in 10 ? In which version did it still work ? (for bisect)

PVince81:
@VicDeo please have a look

butonic:
ie 11, also broken in 9.1

VicDeo:
@butonic win7 IE11. Can't reproduce on master

PVince81:
@VicDeo can you reproduce it on 9.1 ? Maybe something missing in the steps ?

VicDeo:
![win7ie11](https://cloud.githubusercontent.com/assets/991300/21858904/ca3b6d72-d83b-11e6-8f89-4df1d5b9bd94.gif)
@PVince81 @butonic

PVince81:
@VicDeo can you switch to "index.php-less" mode https://doc.owncloud.org/server/9.1/admin_manual/configuration_server/index_php_less_urls.html ? From what I see in what @butonic posted it seems to be using that format. Who knows if it's maybe related.

PVince81:
Ah no, never mind, read the wrong link... The response shows https:\/\/asylum\/oc\/core\/index.php\/s\/4iSz1ownieVsPtW

PVince81:
@VicDeo are you using the release tarball ? Maybe try with the enterprise version ?

Maybe there's another app's JS code interfering.

VicDeo:
@butonic does the response have proper json headers in your case? (just guessing)

VicDeo:
@PVince81 I will retry with 9.1.2 release in a minute

[...]

PR #30178, opened 2018-01-17
disable php cache control

butonic:
Had to prevent PHP from sending cache control headers. Otherwise the dav previews from https://github.com/owncloud/core/pull/29319 always had to reload the previews because Chrome respects the Pragma:nocache header that is sent by default.

Related: https://github.com/owncloud/core/issues/26922#issuecomment-272392100

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -43,6 +43,7 @@
     php_value mbstring.func_overload 0
     php_value default_charset 'UTF-8'
     php_value output_buffering 0
+    php_value session.cache_limiter 0
     <IfModule mod_env.c>
       SetEnv htaccessWorking true
     </IfModule>
@@ -55,6 +56,7 @@
     php_value mbstring.func_overload 0
     php_value default_charset 'UTF-8'
     php_value output_buffering 0
+    php_value session.cache_limiter 0
     <IfModule mod_env.c>
       SetEnv htaccessWorking true
     </IfModule>


.htaccess
```diff
@@ -43,6 +43,7 @@
     php_value mbstring.func_overload 0
     php_value default_charset 'UTF-8'
     php_value output_buffering 0
+    php_value session.cache_limiter 0
```
DeepDiver1975: http://php.net/manual/fa/session.configuration.php#ini.session.cache-limiter shouldn't that be a string?

DeepDiver1975:
what about environments where htaccess is not in place?
should we add this to base.php?

.htaccess
butonic: yes but `php_value session.cache_limiter ''` leads to a 500 status and `.htaccess: php_value takes two arguments, PHP Value Modifier` in the apache error log. `0` seems to work

butonic:
maybe base.php is a better place ... all app framework routes already handle cache control headers

mmattel:
@butonic any update on this ?

PVince81:
=> base.php or Webdav plugin layer ? @DeepDiver1975 @butonic 

decide and then we can take over and reassign

butonic:
To clarify, the [php docs on session_cache_limiter](http://php.net/manual/en/function.session-cache-limiter.php) say
> Setting the cache limiter to '' will turn off automatic sending of cache headers entirely. 

In general, we are managing the cache control headers ourself in the app framework Response  https://github.com/owncloud/core/blob/v10.1.0/lib/public/AppFramework/Http/Response.php#L49 or any derived class. So disabling cache management in php is the right thing to do, because it might otherwise cause unwanted caching.

However, setting it to '' in the .htaccess produces a 500 error, so I tried 0 and it worked. It seems Drupal is using `none` to do the same thing. Maybe `^$` works as expected but might be even more cryptic.

I vote to disable php cache control in the base.php. It least we can then properly set it to ''.

PVince81:
raised as https://github.com/owncloud/core/issues/34441

if there is no particular criticality to this, the ticket will remain in backlog for later

PR #30188, opened 2018-01-19
only parse info.xml once

butonic:
I was profiling the preview stuff and noticed that we are parsing all info.xml files twice per request.

status phppng

Top left: current state
Top rigth: cache in local variable per request for file as well.
Bottom left: This PR, cold cache
Bottom right: cached info.xml

x6 speed increase

Related: https://github.com/owncloud/core/issues/30190


mrow4a:
:O interesting, :+1 for idea, let's see the review when you finish

cdamken:
Testing it and runs fine and faster than b4

butonic:
yay, finally found the problem, now *really* ready to review

mrow4a:
I remember @DeepDiver1975 telling me never to add lines in legacy and move the code to new code base. So you think this extra effort is possible here ?

lib/private/legacy/app.php
```diff
@@ -624,36 +660,86 @@ public static function getAppVersionByPath($path) {
 	 * @return array|null
 	 * @note all data is read from info.xml, not just pre-defined fields
 	 */
-	public static function getAppInfo($appId, $path = false) {
-		if ($path) {
+	public static function getAppInfo($appId, $isPath = false) {
+		if(empty($appId)) {
+			return null; // TODO explode?
```
mrow4a: TO-DO?

lib/private/legacy/app.php
```diff
@@ -624,36 +660,86 @@ public static function getAppVersionByPath($path) {
...
+		$cachedInfo['_cached'] = true;
+		$info['_cached'] = false;
 
-		return $data;
+		// add etag and path so cache can be invalidated
+		$data = [
+			'etag' => $etag,
+			'path' => $file,
+			// store info in its own key so path and etag cannot be injected
+			'info' => $cachedInfo
```
mrow4a: Hmm will need to debug it myself later.

lib/private/legacy/app.php
```diff
@@ -624,36 +660,86 @@ public static function getAppVersionByPath($path) {
...
+		];
+
+		// cache results for a day
+		self::$appInfo->set($appId, $data, 86400);
+		self::$appInfo->set($file, $data, 86400);
```
mrow4a: Why twice for appid and file? (I am not so familiar with this part of code)

tests/data/app/expected-info.json
```diff
@@ -69,11 +69,21 @@
 		}
 	},
 	"repair-steps": {
-		"install": [],
-		"pre-migration": [],
-		"post-migration": [],
-		"live-migration": [],
-		"uninstall": []
+		"install": {
```
mrow4a: Where do you use this test file?

mrow4a (commented):
I added few comments to better understand this part of code

lib/private/legacy/app.php
butonic: https://github.com/owncloud/core/issues/30190 - invalid xml produces a version of 0 ... causing migrations to not be applied
https://github.com/owncloud/core/pull/30204 - but it needs a more in depth effort to make it explode properly / do the right thing if it explodes

lib/private/legacy/app.php
butonic: if i wero to store just the `$data` the `path`and `etag` could be overwritten with the xml. I don't want to risk that.

lib/private/legacy/app.php
butonic: the second parameter of this function allows chinging the behavior of this function to fetching the appinfo by absolute path instead of just appid. both happens at least once per request, causing at least parsing of the info.xml twice. by caching it for both cases we can omit the expensive 'getAppPath' on subsequent calls.

tests/data/app/expected-info.json
butonic: I tried making codecov happy by also extending the coverage of the InfoParserTest. codecov told me this PR would reduce test coverage there ... for whatever reason. So I added those lines to make mor code in the InfoParser be executed.

butonic:
Let me see if I can move this to AppManager

lib/private/legacy/app.php
mrow4a: 👍

tests/data/app/expected-info.json
mrow4a: 👍 Just hope it is separate commit or sth :d

lib/private/legacy/app.php
mrow4a: Will try later

lib/private/legacy/app.php
mrow4a: So TODO needs to stay there?

butonic: as a reminder, for anyone looking at the code, yes

lib/private/App/AppManager.php
```diff
@@ -349,6 +361,23 @@ public function getAppsNeedingUpgrade($ocVersion) {
 		return $appsToUpgrade;
 	}
 
+	/**
+	 * @param $path
+	 * @return string|false an etag for the given $path or false
```
tomneedham: returns null? where does it return false?

tomneedham (commented):
Looks good - other than this doc block. We can add exceptions in a later PR where things go wrong but this should at least maintain existing functionality.

tomneedham:
Squash?

patrickjahns:
@ownclouders rebase

ownclouders:

ownclouders:

patrickjahns:
@ownclouders rebase

ownclouders:

ownclouders:

patrickjahns:
@butonic 
this fails both on drone and jenkins - more failures on jenkins though

butonic:
was assinging an object instead of the result of a method call on that object:
```php
// wrong
$appData = \OC::$server->getAppManager();
// correct
$appData = \OC::$server->getAppManager()->getAppInfo($app);
```

this caused the app types to be set as empty string, which in turn would make the isType check return always false... which only affected trashban because the other apps used by dav are effectively hardcoded.

So much more to do though. In any case ... do not forget to set your `'memcache.local' => '\OC\Memcache\APCu',` so the appinfo cache persists over requests for the full performance benefit: only parse info.xml when it changes

lib/private/App/AppManager.php
```diff
@@ -62,6 +64,7 @@ class AppManager implements IAppManager {
 	private $userSession;
 	/** @var \OCP\IAppConfig */
 	private $appConfig;
+	private $appInfo;
```
DeepDiver1975: var missing

lib/private/App/AppManager.php
```diff
@@ -106,10 +109,19 @@ public function __construct(IUserSession $userSession = null,
 		$this->memCacheFactory = $memCacheFactory;
 		$this->dispatcher = $dispatcher;
 		$this->config = $config;
+
+		// TODO we have no public API for this
```
DeepDiver1975: file tech dept issue please - THX for noticing

lib/private/App/AppManager.php
```diff
@@ -106,10 +109,19 @@ public function __construct(IUserSession $userSession = null,
...
+		if (method_exists($this->memCacheFactory, 'createLocal')) {
+			$this->appInfo = $this->memCacheFactory->createLocal('app-info');
+		}
+		if ($this->appInfo === null || $this->appInfo instanceof NullCache) {
+			$this->appInfo = new ArrayCache('app-info');
```
DeepDiver1975: maybe just use arraycache all the time?

This way the xml is parse once per request and we don't run into cache invalidation issues ...

lib/private/App/AppManager.php
```diff
@@ -349,6 +372,23 @@ public function getAppsNeedingUpgrade($ocVersion) {
...
+	 */
+	private function getEtag($path) {
+		if (!\file_exists($path)) {
+			return null;
+		}
+		\clearstatcache(false, $path);
+		$stat = \stat($path);
+		if ($stat) {
+			// ok, file still exists
+			return "${stat['mtime']}|${stat['ino']}|${stat['dev']}|${stat['size']}";
```
DeepDiver1975: content md5? files are small - should not be that expensive ...

lib/private/App/AppManager.php
```diff
@@ -357,17 +397,121 @@ public function getAppsNeedingUpgrade($ocVersion) {
...
+			\OC::$server->getLogger()->logException($e);
+			throw $e;
+		}
+
+		$info = \OC_App::parseAppInfo($info); // TODO move to info parser?
+
+		$appId = \OC_App::cleanAppId($info['id']); // so we can fetch the right config value and cache correctly
+
+		// always use stored ocsid // TODO add / explain reason
+		if(isset($info['ocsid'])) {
```
DeepDiver1975: ocsid has afaik no meaning anymore ... needs a second look.

background: ocsid was used as identifier with the appstore. the marketplace uses the app name as id

lib/private/legacy/app.php
```diff
@@ -894,8 +876,6 @@ public static function listAllApps() {
 						$info['documentation'][$key] = $url;
 					}
 				}
-
-				$info['version'] = OC_App::getAppVersion($app);
```
DeepDiver1975: is version now set within getAppManager()->getAppInfo

lib/private/legacy/app.php
```diff
@@ -1123,21 +1105,19 @@ private static function setupLiveMigrations($appId, array $steps) {
 	 * @return \OC\Files\View|false
 	 */
 	public static function getStorage($appId) {
```
DeepDiver1975: wow - funny methods we have .... is this used anywhere? afaik we have a method in the file node api which does the same .....

lib/private/legacy/app.php
```diff
@@ -1198,9 +1178,9 @@ protected static function checkAppDependencies($config, $l, $info) {
 
 	/**
 	 * @param $appId
+	 * @deprecated no longer needed
```
DeepDiver1975: lets kill it this is private api - right?

DeepDiver1975 (dismissed):

lib/private/App/AppManager.php
butonic: no that would mean reading the file. stat is much faster than file_get_contents

lib/private/App/AppManager.php
butonic: No. If a local memcache like apcu is used info.xml files will only be parsed if the etag changes. If it is not configured the fectory will return a NullCache that we need to replace with the ArrayCache ... as a fallback.

lib/private/legacy/app.php
butonic: yes, by reading it from the info.xml file. the fallback to read a version file was removed before this PR ... probably long ago. was looking for the code but found it to be gone already

lib/private/legacy/app.php
butonic: :sob: 
OC_App.getStorage (lib/private/legacy/app.php)
- Files.getStorage (lib/public/Files.php)
  - SFTP.hostKeysPath (apps/files_external/lib/Lib/Storage/SFTP.php)
  - SFTP.writeHostKeys (apps/files_external/lib/Lib/Storage/SFTP.php)
  - SFTP.readHostKeys (apps/files_external/lib/Lib/Storage/SFTP.php)

unfortunately, it is different from what `Server::getAppFolder()` does:
https://github.com/owncloud/core/blob/6e087ca7f398571c7426edf49fe926e989dd7cfb/lib/private/Server.php#L1007-L1016

in a subsequent PR
- [ ] fix files_external to not use this code
- [ ] get rid of this method

lib/private/App/AppManager.php
DeepDiver1975: I understood your implementation - my proposal was to have a change which might change a bit less ... and therefore less error prone ....

lib/private/legacy/app.php
DeepDiver1975: :+1:

lib/private/legacy/app.php
DeepDiver1975: joining ..... :scream: :tired_face: :sob:

lib/private/App/AppManager.php
VicDeo: who needs file_get_contents + md5 of content when there is `md5_file`?

lib/private/App/AppManager.php
butonic: nah, basically this PR implements a proper cache for the app info. part of the app info is cached was already cached in the db, like the app types. The challenge is to properly invalidate the cache. That is why I made the etag based check. This will allow us to only read and parse info.xml if the file changes. We won't pars the file even once per request. That is the point of this PR: major performance gain.

I think this also obsoleted the necessity to load app types from db at all ... no point in caching them there if we have to read them once per request anyway (in case no local cache is configured). But that is a story for a different PR.

lib/private/legacy/app.php
VicDeo: wow. nice catch...  fallback to a separate 'version' file is  currently in `AppManager.php` only, it might explain an empty installed_version https://github.com/owncloud/core/issues/30724 
cc @tomneedham

lib/private/App/AppManager.php
```diff
@@ -357,17 +398,122 @@ public function getAppsNeedingUpgrade($ocVersion) {
...
+			throw $e;
+		}
+
+		$info = \OC_App::parseAppInfo($info); // TODO move to info parser?
+
+		$appId = \OC_App::cleanAppId($info['id']); // so we can fetch the right config value and cache correctly
+
+		// always use stored ocsid
+		// TODO remove? ocsid was used as identifier with the appstore. the marketplace uses the app name as id
+		if(isset($info['ocsid'])) {
```
DeepDiver1975: kill it - nobody needs ocsid anymore

lib/private/App/AppManager.php
```diff
@@ -429,6 +588,8 @@ public function getAlwaysEnabledApps() {
 	 * @param bool $skipMigrations whether to skip migrations, which would only install the code
 	 * @return string|false app id or false in case of error
 	 * @since 10.0
+	 * @throws \Exception
```
DeepDiver1975: duplicate

lib/private/App/AppManager.php
```diff
@@ -463,6 +626,9 @@ public function getAllApps() {
 	/**
 	 * @param string $path
 	 * @return string[] app info
+	 * @throws \Exception
+	 * @throws \Exception
```
DeepDiver1975: triplicate

lib/private/legacy/app.php
```diff
@@ -604,70 +614,45 @@ public static function isAppDirWritable($appId) {
...
-		return self::$appVersion[$appId];
+		return '0';
 	}
 
 	/**
 	 * get app's version based on it's path
 	 *
 	 * @param string $path
 	 * @return string
+	 * @deprecated use \OC::$server->getAppManager()->getAppInfoByPath($path)
```
DeepDiver1975: no public api -> kill

lib/private/legacy/app.php
```diff
@@ -859,7 +842,6 @@ public static function listAllApps() {
 				} else {
 					$result = \OC::$server->getIntegrityCodeChecker()->verifyAppSignature($app, '', true);
 					if (empty($result)) {
-						$info['internal'] = false;
```
DeepDiver1975: why? not needed anymore?

tests/lib/legacy/AppTest.php
```diff
@@ -20,23 +21,49 @@
 
 namespace Test\legacy;
 
-
 use OC\NavigationManager;
+use OCP\App\AppNotFoundException;
 use Test\TestCase;
 
+function rrmdir($directory) {
```
DeepDiver1975: there is already a helper for this somewhere ...... OC_Util maybe?

DeepDiver1975 (dismissed):

tests/lib/legacy/AppTest.php
```diff
@@ -20,23 +21,49 @@
...
+		$this->appPath = __DIR__ . '/../../../apps/appinfotestapp';
+		$infoXmlPath = "{$this->appPath}/appinfo/info.xml";
+		mkdir("{$this->appPath}/appinfo", 0777, true);
+
+		$xml = '<?xml version="1.0" encoding="UTF-8"?>' .
+		'<info>' .
+		    '<id>appinfotestapp</id>' .
+			'<namespace>AppInfoTestApp</namespace>' .
+		'</info>';
+		file_put_contents($infoXmlPath, $xml);
```
DeepDiver1975: did you elaborate if using vfsstream can be used here? 

vfsstream will help to speed up tests because the filesystem is not hit

tests/lib/legacy/AppTest.php
butonic:
```
use function Test\AppFramework\rrmdir;
```

tests/lib/legacy/AppTest.php
butonic: the legacy app.php has no DI cannot inject vfsstream -> separate PR.

lib/private/legacy/app.php
```diff
@@ -283,22 +290,24 @@ public static function isType($app, $types) {
...
 			return \explode(',', self::$appTypes[$app]);
-		} else {
-			return [];
 		}
+		return [];
 	}
 
 	/**
 	 * read app types from info.xml and cache them in the database
+	 * TODO we seem to be using the db as a cache for app types. we have those in the appinfo as well. simplify?
```
mrow4a: Not sure what do you mean here

lib/private/legacy/app.php
```diff
@@ -585,70 +590,32 @@ public static function isAppDirWritable($appId) {
...
-
-		if (\is_array($data)) {
-			$data = OC_App::parseAppInfo($data);
-		}
-		if(isset($data['ocsid'])) {
-			$storedId = \OC::$server->getConfig()->getAppValue($appId, 'ocsid');
-			if($storedId !== '' && $storedId !== $data['ocsid']) {
-				$data['ocsid'] = $storedId;
-			}
+	public static function getAppInfo($appId, $isPath = false) {
```
mrow4a: I see that each request calls this static function anyways. Good thing is that it now uses AppManager - Do you think in long run we can ger dir of this legacy code and use \OC::$server->getAppManager(); in places as https://github.com/owncloud/core/blob/master/lib/private/AppFramework/App.php#L50

mrow4a (commented):
I checked manualy, and indeed subsequent calls go to cache instead of parsing. However some changes to this PR are very dark corners of OC, and I think @DeepDiver1975 needs to have final look outside of lib/private/App/AppManager.php

lib/private/Installer.php
```diff
@@ -348,20 +347,13 @@ public static function checkAppsIntegrity($data, $extractDir, $path, $isShipped
...
-			$previouslySigned = \OC::$server->getConfig()->getAppValue($appBelongingToId, 'signed', 'false');
-		} else {
-			$appBelongingToId = $info['id'];
-			$previouslySigned = 'false';
-		}
-		if (\file_exists($extractDir . '/appinfo/signature.json') || $previouslySigned === 'true') {
-			\OC::$server->getConfig()->setAppValue($appBelongingToId, 'signed', 'true');
-			$integrityResult = \OC::$server->getIntegrityCodeChecker()->verifyAppSignature(
-					$appBelongingToId,
+		if (\file_exists($extractDir . '/appinfo/signature.json')) {
```
mrow4a: Is it just refactor to get rid of ocsid ?

lib/private/App/AppManager.php
```diff
@@ -322,6 +340,8 @@ public function disableApp($appId) {
 	public function clearAppsCache() {
 		$settingsMemCache = $this->memCacheFactory->create('settings');
 		$settingsMemCache->clear('listApps');
+		$this->appInfo->clear();
+		$this->appDirs = [];
```
mrow4a: Why wasnt it cleared previously?

mrow4a: appDirs were not cleared before this change. And you seem not to use it. Just were curious if that was bug previously

lib/private/App/AppManager.php
butonic: tracked in https://github.com/owncloud/core/issues/31604

lib/private/App/AppManager.php
butonic: I added tests for the legacy OC_APP class which uses static code, so the underlying AppManager is reused when testing. So I need a way to clear the appDirs in https://github.com/owncloud/core/pull/30188/files#diff-8bd506369a837b0d4dee9c6f0b35ddabR48

PVince81:
those who had comments please re-review this PR!

lib/private/legacy/app.php
```diff
@@ -129,6 +126,36 @@ public static function loadApps($types = null) {
 		}
 		\ob_end_clean();
 
+		// once all authentication apps are loaded we can validate the session
+		if ($types === null || \in_array('authentication', $types, true)) {
```
mrow4a: Why do we need this part now for this PR?

butonic: this got mixed up with the auth stuff ... damn

patrickjahns:
Some more thoughts on this:
- What happens when we upgrade an app via `occ` ?
- What happenes when we upgrade/install an app via `occ market:install` ?

butonic:
@patrickjahns the cache is invalidated if an info.xml etag changes. see https://github.com/owncloud/core/pull/30188/files#diff-8fa91b147fc76360cb3cfca41a1db57dR406

PVince81:
@butonic please recheck for outstanding comments/tasks so someone else can take over.

lib/private/Installer.php
butonic: yes, especially since getInternalAppIdByOcs no longer makes sense, which allows us to reduce the code above.

lib/private/legacy/app.php
butonic: Yes, that is the idea

lib/private/legacy/app.php
butonic: We are caching app versions in the db. But reading them from disk should be faster. If we are reading all the xml files anyway, then there is no need to query the db for it. Then again we want to check if the app version changed, so we need to keep track of the currently installed version, which is why I removed this TODO.

butonic:
@PVince81 I rebased. All comments should be addressed. Please take over. Fingers crossed about the testsuite.

PVince81:
@VicDeo please takeover as you're already familiar with the matter

lib/private/App/CodeChecker/InfoChecker.php
```diff
@@ -58,6 +57,10 @@ class InfoChecker extends BasicEmitter {
 	];
 	private $deprecatedFields = [
 		'info',
+		'ocsid',
+		'require',
+		'requiremax',
+		'requiremin',
```
VicDeo: @PVince81 Lines 61-63 partially revert https://github.com/owncloud/core/pull/33421 that was not backported to stable10.

lib/private/App/AppManager.php
```diff
@@ -374,17 +406,113 @@ public function getAppsNeedingUpgrade() {
...
 	 *
 	 * @internal
+	 * @throws \InvalidArgumentException
+	 * @throws AppNotFoundException
 	 */
 	public function getAppInfo($appId) {
-		$appInfo = \OC_App::getAppInfo($appId);
-		if ($appInfo === null) {
-			return null;
+		if (empty($appId)) {
```
VicDeo: `empty` is not good

VicDeo:
Great job in general, just two minor things

lib/private/App/CodeChecker/InfoChecker.php
butonic: good catch

lib/private/App/AppManager.php
butonic: well ... ok ... i did not want to check for null, false and emptystring ... and an ipp id with 0 or an empty array is wrong as well .. so i think this kind of makes sense here ... maybe do `if (!is_string($appId) || $appId === '')`?

VicDeo: one of rare cases when strict comparison sucks :)

butonic: ok, I went with `(!is_string($appId) || $appId === '')`

butonic:
@VicDeo both resolved

VicDeo (approved):
changed `is_string` to `\is_string` to pass CI. The rest is good.

VicDeo:
@PVince81 backport?

VicDeo:
Stable10: https://github.com/owncloud/core/pull/34482

PR #30195, opened 2018-01-19
Improve search query for adressbooks

butonic:
The current search query for addressbook entries is problematic:
```sql
MariaDB [core]> EXPLAIN
SELECT c.`carddata`, c.`uri`
FROM `oc_cards` c
WHERE c.`id` IN (
    SELECT DISTINCT cp.`cardid`
    FROM `oc_cards_properties` cp
    WHERE (
        ( (cp.`name` = 'CLOUD') AND (cp.`value`  COLLATE utf8_general_ci LIKE '%foo%') )
        OR ( (cp.`name` = 'FN') AND (cp.`value`  COLLATE utf8_general_ci LIKE '%foo%') )
    )
    AND (cp.`addressbookid` = 10)
) ORDER BY c.`uri` ASC LIMIT 200 OFFSET 0;
+------+--------------+-------------+--------+--------------------------------------+-----------------+---------+------+--------+------------------------------------+
| id   | select_type  | table       | type   | possible_keys                        | key             | key_len | ref  | rows   | Extra                              |
+------+--------------+-------------+--------+--------------------------------------+-----------------+---------+------+--------+------------------------------------+
|    1 | PRIMARY      | c           | index  | PRIMARY,id_uri_index                 | uri_index       | 768     | NULL |      1 |                                    |
|    1 | PRIMARY      | <subquery2> | eq_ref | distinct_key                         | distinct_key    | 8       | func |      1 |                                    |
|    2 | MATERIALIZED | cp          | range  | card_name_index,card_contactid_index | card_name_index | 195     | NULL | 372808 | Using index condition; Using where |
+------+--------------+-------------+--------+--------------------------------------+-----------------+---------+------+--------+------------------------------------+
3 rows in set (0.00 sec)

SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 10     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%filler%' ) ORDER BY c.`uri` ASC LIMIT 200 OFFSET 0;
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+
...
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+
200 rows in set (0.29 sec)
```
0.29 sec. ouch. on large addressbooks (>100k cards) tables this query takes more than a second ... per addressbook. 20 addressbooks => 20 seconds until the user sees a result.

We can do better. Let us have a look at the query. Maybe we can simplify it? What about:
```sql
MariaDB [core]> EXPLAIN
SELECT c.`carddata`, c.`uri`
FROM `oc_cards` c
WHERE c.`id` IN (
    SELECT DISTINCT cp.`cardid`
    FROM `oc_cards_properties` cp
    WHERE cp.`addressbookid` = 11
    AND cp.`name` IN ('CLOUD','FN')
    AND cp.`value`  COLLATE utf8_general_ci LIKE '%filler%'
)
ORDER BY c.`uri` ASC LIMIT 200 OFFSET 0;
+------+--------------+-------------+--------+--------------------------------------+-----------------+---------+------+--------+------------------------------------+
| id   | select_type  | table       | type   | possible_keys                        | key             | key_len | ref  | rows   | Extra                              |
+------+--------------+-------------+--------+--------------------------------------+-----------------+---------+------+--------+------------------------------------+
|    1 | PRIMARY      | c           | index  | PRIMARY,id_uri_index                 | uri_index       | 768     | NULL |      1 |                                    |
|    1 | PRIMARY      | <subquery2> | eq_ref | distinct_key                         | distinct_key    | 8       | func |      1 |                                    |
|    2 | MATERIALIZED | cp          | range  | card_name_index,card_contactid_index | card_name_index | 195     | NULL | 372808 | Using index condition; Using where |
+------+--------------+-------------+--------+--------------------------------------+-----------------+---------+------+--------+------------------------------------+
3 rows in set (0.00 sec)

SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 11     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%filler%' ) ORDER BY c.`uri` ASC LIMIT 200 OFFSET 0;
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+
...
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+
200 rows in set (0.29 sec)
```

Well, nothing changed. But at least it is more readable ... so what about the indexes. Adding one for the materialized subquery became clearer:
```sql
MariaDB [core]> CREATE INDEX carddata_aid_n_v_idx ON oc_cards_properties (addressbookid, name, value);
Query OK, 0 rows affected (1.81 sec)
Records: 0  Duplicates: 0  Warnings: 0

MariaDB [core]> EXPLAIN SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 12     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%filler%' ) ORDER BY c.`uri` ASC LIMIT 200 OFFSET 0;
+------+--------------+-------------+--------+-----------------------------------------------------------+----------------------+---------+----------------+------+---------------------------------+
| id   | select_type  | table       | type   | possible_keys                                             | key                  | key_len | ref            | rows | Extra                           |
+------+--------------+-------------+--------+-----------------------------------------------------------+----------------------+---------+----------------+------+---------------------------------+
|    1 | PRIMARY      | <subquery2> | ALL    | distinct_key                                              | NULL                 | NULL    | NULL           | 3962 | Using temporary; Using filesort |
|    1 | PRIMARY      | c           | eq_ref | PRIMARY,id_uri_index                                      | PRIMARY              | 8       | core.cp.cardid |    1 |                                 |
|    2 | MATERIALIZED | cp          | range  | card_name_index,card_contactid_index,carddata_aid_n_v_idx | carddata_aid_n_v_idx | 203     | NULL           | 3962 | Using index condition           |
+------+--------------+-------------+--------+-----------------------------------------------------------+----------------------+---------+----------------+------+---------------------------------+
3 rows in set (0.00 sec)

MariaDB [core]> SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 12     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%filler%' ) ORDER BY c.`uri` ASC LIMIT 200 OFFSET 0;
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+
...
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------+
200 rows in set (0.03 sec)
```

Much better. We now have an index that correctly partitions the table.

### DIFF
--- a/apps/dav/lib/CardDAV/CardDavBackend.php
+++ b/apps/dav/lib/CardDAV/CardDavBackend.php
@@ -833,10 +833,6 @@ class CardDavBackend implements BackendInterface, SyncSupport {
 	 * @return array an array of contacts which are arrays of key-value-pairs
 	 */
 	public function searchEx($addressBookId, $pattern, $searchProperties, $options, $limit = 100, $offset = 0) {
-		$query = $this->db->getQueryBuilder();
-		$query2 = $this->db->getQueryBuilder();
-		$query2->selectDistinct('cp.cardid')->from($this->dbCardsPropertiesTable, 'cp');
-
 		$matchMode = $options['matchMode'] ?? 'any';
 		switch ($matchMode) {
 			case 'START':
@@ -853,21 +849,18 @@ class CardDavBackend implements BackendInterface, SyncSupport {
 				$searchPattern = '%' . $this->db->escapeLikeParameter($pattern) . '%';
 		}
 
-		foreach ($searchProperties as $property) {
-			$query2->orWhere(
-				$query2->expr()->andX(
-					$query2->expr()->eq('cp.name', $query->createNamedParameter($property)),
-					$query2->expr()->iLike('cp.value', $query->createNamedParameter($searchPattern))
-				)
-			);
-		}
-		$query2->andWhere($query2->expr()->eq('cp.addressbookid', $query->createNamedParameter($addressBookId)));
-
-		$query->select('c.carddata', 'c.uri')->from($this->dbCardsTable, 'c')
-			->where($query->expr()->in('c.id', $query->createFunction($query2->getSQL())));
-
-		$query->setFirstResult($offset)->setMaxResults($limit);
-		$query->orderBy('c.uri');
+		$query = $this->db->getQueryBuilder();
+		$query->select(['c.uri', 'c.carddata'])
+			->from($this->dbCardsTable, 'c')
+			->join('c', $this->dbCardsPropertiesTable, 'cp', $query->expr()->eq('c.id', 'cp.cardid'))
+			->where($query->expr()->eq('cp.addressbookid', $query->createNamedParameter($addressBookId)))
+			->andWhere($query->expr()->in('cp.name', \array_map(function ($property) use ($query) {
+				return $query->createNamedParameter($property);
+			}, $searchProperties)))
+			->andWhere($query->expr()->iLike('cp.value', $query->createNamedParameter($searchPattern)))
+			->setFirstResult($offset)->setMaxResults($limit)
+			->groupBy(['c.uri', 'c.carddata']) // groupby without aggregation is essentially distinct
+			->orderBy('c.uri');
 
 		$result = $query->execute();
 		$cards = $result->fetchAll();
--- a/apps/dav/appinfo/Migrations/Version20200114181454.php
+++ b/apps/dav/appinfo/Migrations/Version20200114181454.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
+ *
+ * @copyright Copyright (c) 2020, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+namespace OCA\dav\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use OCP\Migration\ISchemaMigration;
+
+/**
+ * Add index to oc_cards_properties to assist with searching with large numbers of rows
+ */
+class Version20200114181454 implements ISchemaMigration {
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+		$table = $schema->getTable("${prefix}cards_properties");
+		// Check for existing index spanning these columns
+		foreach ($table->getIndexes() as $index) {
+			// Check if we have a matching index already
+			if (empty(\array_diff($index->getColumns(), ['addressbookid', 'name', 'value']))) {
+				return;
+			}
+		}
+		// Add the index if we dont have one spanning this column already
+		$table->addIndex(['addressbookid', 'name', 'value'], 'carddata_aid_n_v_idx');
+	}
+}


apps/dav/appinfo/Migrations/Version20180119134849.php
```diff
@@ -0,0 +1,46 @@
...
+		// Check for existing index spanning these columns
+		foreach ($table->getIndexes() as $index) {
+			// Check if we have a matching index already
+			if(empty(array_diff($index->getColumns(), ['addressbookid', 'name', 'value']))) {
+				return;
+			}
+
+		}
+		// Add the index if we dont have one spanning this column already
+		$table->addIndex(['addressbookid', 'name', 'value'], 'carddata_aid_n_v_idx');
```
DeepDiver1975: isn't the name of the index maintained by doctrine?

butonic: currently we always set it... at some point in time the db layer added a suffix, which is why I shortened the index name a little.

mrow4a:
Is 'filler' exactly the value, or sth it starts with like filler% ? Then it might work

apps/dav/appinfo/Migrations/Version20180119134849.php
```diff
@@ -0,0 +1,46 @@
...
+ * Add index to oc_cards_properties to assist with searching with large numbers of rows
+ */
+class Version20180119134849 implements ISchemaMigration {
+
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+		$table = $schema->getTable("${prefix}cards_properties");
+		// Check for existing index spanning these columns
+		foreach ($table->getIndexes() as $index) {
+			// Check if we have a matching index already
```
tomneedham: 👍

apps/dav/appinfo/Migrations/Version20180119134849.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+namespace OCA\dav\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use OCP\Migration\ISchemaMigration;
+
+/**
+ * Add index to oc_cards_properties to assist with searching with large numbers of rows
+ */
+class Version20180119134849 implements ISchemaMigration {
```
tomneedham: We need a cost estimation for this index - time expected given number of rows - given this could be released in a minor version we need to give useful information to the admins

apps/dav/lib/CardDAV/CardDavBackend.php
```diff
@@ -806,17 +806,13 @@ public function search($addressBookId, $pattern, $searchProperties, $limit = 100
...
-					$query2->expr()->ilike('cp.value', $query->createNamedParameter('%' . $this->db->escapeLikeParameter($pattern) . '%'))
-				)
-			);
-		}
-		$query2->andWhere($query2->expr()->eq('cp.addressbookid', $query->createNamedParameter($addressBookId)));
+		$query2->where($query2->expr()->eq('cp.addressbookid', $query->createNamedParameter($addressBookId)));
+		$query2->andWhere($query2->expr()->in('cp.name', array_map(function($property) use ($query) {
+			return $query->createNamedParameter($property);
+		}, $searchProperties)));
+		$query2->andWhere($query2->expr()->ilike('cp.value', $query->createNamedParameter('%' . $this->db->escapeLikeParameter($pattern) . '%')));
```
mrow4a: Why is the index also on the value, if we use %value% ?

DeepDiver1975:
needs rebase ....

apps/dav/lib/CardDAV/CardDavBackend.php
mrow4a: Did you also try with just index on addressbookid? But maybe I would need to get better understanding of how the carddav table gets populated, what other queries are executed there, and how the index update time will be affected.

tomneedham:
> Is 'filler' exactly the value, or sth it starts with like filler% ? Then it might work

We can use the same logic as the accounts table here if this helps. We have `'accounts.enable_medial_search' => true,` which changes the search from `%term%` to `term%`

tomneedham:
Testing on my machine with 20k user accounts (provisioned and logged in at least once):
 - Running the query without the index ~180 - 200ms
 - With the index across just addressbookid, name ~1.8ms

I will adjust the PR to only cover those two fields - then we save some space in the index especially as value is VARCHAR 255.

butonic:
well, medial search can only benefit if the index contains the value

mrow4a:
@tomneedham so you just confirmed that only on 2 cols is enough? Can you try also with only addressbookid? And yes, maybe for medial search we can put full table in memory, but admins need to know that the table will be full in memory and each single update in that table will need to update that index structure.. each time..

mrow4a:
@butonic @tomneedham for medial search, maybe better we put index only on value though.

tomneedham:
> so you just confirmed that only on 2 cols is enough?

Yes - when not considering medial search.



> but it will work only for filler%, not for %filler% which might only be solved with index on addressbookid.

That is fine - we should also use a config option to control this functionality like we have with `'accounts.enable_medial_search' => true,`

butonic:
@mrow4a please have a look at the explain output. Can you give me a reason for indexing only the addressbookid?

mrow4a:
When is selectivity of unique addressbookid vs selectivity of unique addressbookid&cp.name pair vs full table in memory indexed.

mrow4a:
Another problem I see is that it is hard to decide on where to put index, since we would need to know in advance if medial search is allowed or not. Honestyly I would go for %filler% and put index on addressbook alone if that works for that workload or addressbook&cp.name

apps/testing/appinfo/info.xml
```diff
@@ -9,6 +9,9 @@
 	<dependencies>
 		<owncloud min-version="10.1" max-version="10.1" />
 	</dependencies>
+	<commands>
+		<command>OCA\Testing\Command\CreateMultiUser</command>
```
mrow4a: @butonic @tomneedham I temporarly added to this PR these two commits - it should be easier to test with these two in here. It will basically instead of just creating users, also create X contacts for each user (addressbook)

mrow4a: You can test it with 
```
php ./occ app:enable testing
php ./occ testing:createusers 10 10
```

first is number of users, second number of contacts for each user

apps/dav/lib/CardDAV/CardDavBackend.php
mrow4a: As discussed, %filler% here will make no use here with anv index

PVince81:
@butonic finish or defer ?

apps/dav/appinfo/Migrations/Version20180119134849.php
butonic: not worth mentioning
```
mysql> SELECT count(*) FROM oc_cards_properties;
+----------+
| count(*) |
+----------+
|   164306 |
+----------+
1 row in set (0.06 sec)

mysql> CREATE INDEX carddata_aid_n_v_idx ON oc_cards_properties (addressbookid, name, value);
Query OK, 0 rows affected (0.40 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

butonic:
weird ... my current mysql seems to be to stupid to use the index when using a subselect:
```sql
mysql> -- first some stats
mysql> SELECT count(*) FROM oc_cards_properties WHERE addressbookid = 1;
+----------+
| count(*) |
+----------+
|     2880 |
+----------+
1 row in set (0.01 sec)

mysql> SELECT count(*) FROM oc_cards_properties;
+----------+
| count(*) |
+----------+
|   164306 |
+----------+
1 row in set (0.06 sec)

mysql> SHOW INDEX FROM oc_cards_properties;
+---------------------+------------+----------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| Table               | Non_unique | Key_name             | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment |
+---------------------+------------+----------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
| oc_cards_properties |          0 | PRIMARY              |            1 | id          | A         |      164194 |     NULL | NULL   |      | BTREE      |         |               |
| oc_cards_properties |          1 | card_value_index     |            1 | value       | A         |      164194 |     NULL | NULL   | YES  | BTREE      |         |               |
| oc_cards_properties |          1 | card_name_index      |            1 | name        | A         |           3 |     NULL | NULL   | YES  | BTREE      |         |               |
| oc_cards_properties |          1 | card_contactid_index |            1 | cardid      | A         |      164194 |     NULL | NULL   |      | BTREE      |         |               |
+---------------------+------------+----------------------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+
4 rows in set (0.00 sec)

mysql> -- Without the idex a subselect is slow
mysql> EXPLAIN SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ) ORDER BY c.`uri` ASC LIMIT 1;
+----+--------------------+-------+----------------+--------------------------------------+----------------------+---------+------+-------+-----------------------------+
| id | select_type        | table | type           | possible_keys                        | key                  | key_len | ref  | rows  | Extra                       |
+----+--------------------+-------+----------------+--------------------------------------+----------------------+---------+------+-------+-----------------------------+
|  1 | PRIMARY            | c     | ALL            | NULL                                 | NULL                 | NULL    | NULL | 81500 | Using where; Using filesort |
|  2 | DEPENDENT SUBQUERY | cp    | index_subquery | card_name_index,card_contactid_index | card_contactid_index | 8       | func |     1 | Using where                 |
+----+--------------------+-------+----------------+--------------------------------------+----------------------+---------+------+-------+-----------------------------+
2 rows in set (0.00 sec)

mysql> SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ) ORDER BY c.`uri` ASC LIMIT 1;
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| carddata                                                                                                                                                                                         | uri                              |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.1.6//EN
UID:zombie-1525079255-0
FN:zombie-1525079255-0
N:zombie-1525079255-0;;;;
CLOUD:zombie-1525079255-0@localhost
END:VCARD
 | Database:zombie-1525079255-0.vcf |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
1 row in set (0.22 sec)

mysql> -- as is a join
mysql> EXPLAIN SELECT c.`carddata`, c.`uri` FROM `oc_cards` c LEFT JOIN `oc_cards_properties` cp ON c.`id`=cp.`cardid` WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ORDER BY c.`uri` ASC LIMIT 1;
+----+-------------+-------+------+--------------------------------------+----------------------+---------+----------------+-------+----------------+
| id | select_type | table | type | possible_keys                        | key                  | key_len | ref            | rows  | Extra          |
+----+-------------+-------+------+--------------------------------------+----------------------+---------+----------------+-------+----------------+
|  1 | SIMPLE      | c     | ALL  | PRIMARY                              | NULL                 | NULL    | NULL           | 81500 | Using filesort |
|  1 | SIMPLE      | cp    | ref  | card_name_index,card_contactid_index | card_contactid_index | 8       | core_addr.c.id |     1 | Using where    |
+----+-------------+-------+------+--------------------------------------+----------------------+---------+----------------+-------+----------------+
2 rows in set (0.00 sec)

mysql> SELECT c.`carddata`, c.`uri` FROM `oc_cards` c LEFT JOIN `oc_cards_properties` cp ON c.`id`=cp.`cardid` WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ORDER BY c.`uri` ASC LIMIT 1;
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| carddata                                                                                                                                                                                         | uri                              |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.1.6//EN
UID:zombie-1525079255-0
FN:zombie-1525079255-0
N:zombie-1525079255-0;;;;
CLOUD:zombie-1525079255-0@localhost
END:VCARD
 | Database:zombie-1525079255-0.vcf |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
1 row in set (0.24 sec)

mysql> -- adding the index over addressbookid and name
mysql> CREATE INDEX carddata_aid_n_idx ON oc_cards_properties (addressbookid, name);
Query OK, 0 rows affected (0.35 sec)
Records: 0  Duplicates: 0  Warnings: 0

mysql> -- the subselect still sucks
mysql> EXPLAIN SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ) ORDER BY c.`uri` ASC LIMIT 1;
+----+--------------------+-------+----------------+---------------------------------------------------------+----------------------+---------+------+-------+-----------------------------+
| id | select_type        | table | type           | possible_keys                                           | key                  | key_len | ref  | rows  | Extra                       |
+----+--------------------+-------+----------------+---------------------------------------------------------+----------------------+---------+------+-------+-----------------------------+
|  1 | PRIMARY            | c     | ALL            | NULL                                                    | NULL                 | NULL    | NULL | 81500 | Using where; Using filesort |
|  2 | DEPENDENT SUBQUERY | cp    | index_subquery | card_name_index,card_contactid_index,carddata_aid_n_idx | card_contactid_index | 8       | func |     1 | Using where                 |
+----+--------------------+-------+----------------+---------------------------------------------------------+----------------------+---------+------+-------+-----------------------------+
2 rows in set (0.00 sec)

mysql> SELECT c.`carddata`, c.`uri` FROM `oc_cards` c WHERE c.`id` IN (     SELECT DISTINCT cp.`cardid`     FROM `oc_cards_properties` cp     WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ) ORDER BY c.`uri` ASC LIMIT 1;
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| carddata                                                                                                                                                                                         | uri                              |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.1.6//EN
UID:zombie-1525079255-0
FN:zombie-1525079255-0
N:zombie-1525079255-0;;;;
CLOUD:zombie-1525079255-0@localhost
END:VCARD
 | Database:zombie-1525079255-0.vcf |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
1 row in set (0.28 sec)

mysql> -- but the join uses the index
mysql> EXPLAIN SELECT c.`carddata`, c.`uri` FROM `oc_cards` c LEFT JOIN `oc_cards_properties` cp ON c.`id`=cp.`cardid` WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ORDER BY c.`uri` ASC LIMIT 1;
+----+-------------+-------+--------+---------------------------------------------------------+--------------------+---------+---------------------+------+----------------------------------------------+
| id | select_type | table | type   | possible_keys                                           | key                | key_len | ref                 | rows | Extra                                        |
+----+-------------+-------+--------+---------------------------------------------------------+--------------------+---------+---------------------+------+----------------------------------------------+
|  1 | SIMPLE      | cp    | range  | card_name_index,card_contactid_index,carddata_aid_n_idx | carddata_aid_n_idx | 203     | NULL                | 1438 | Using where; Using temporary; Using filesort |
|  1 | SIMPLE      | c     | eq_ref | PRIMARY                                                 | PRIMARY            | 4       | core_addr.cp.cardid |    1 | Using where                                  |
+----+-------------+-------+--------+---------------------------------------------------------+--------------------+---------+---------------------+------+----------------------------------------------+
2 rows in set (0.00 sec)

mysql> SELECT c.`carddata`, c.`uri` FROM `oc_cards` c LEFT JOIN `oc_cards_properties` cp ON c.`id`=cp.`cardid` WHERE cp.`addressbookid` = 1     AND cp.`name` IN ('CLOUD','FN')     AND cp.`value`  COLLATE utf8_general_ci LIKE '%omb%' ORDER BY c.`uri` ASC LIMIT 1;
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| carddata                                                                                                                                                                                         | uri                              |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
| BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.1.6//EN
UID:zombie-1525079255-0
FN:zombie-1525079255-0
N:zombie-1525079255-0;;;;
CLOUD:zombie-1525079255-0@localhost
END:VCARD
 | Database:zombie-1525079255-0.vcf |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+----------------------------------+
1 row in set (0.01 sec)
```

might be caused by percona?
```
Server version:		5.5.59-38.11 Percona Server (GPL), Release 38.11, Revision 2084a7bc7ec
```

So I changed the SQL to use a proper inner join. Also left the index on addressbookid, name and value to allow a subsequent PR to add a medial search flag.

@PVince81 @tomneedham @DeepDiver1975 do we agree on that?

Helios07:
@butonic Any news on this?

phil-davis:
@PVince81 ownclouders seem to have taken a holiday this week.

butonic:
The query needs to be tested on a production instance with multiple large addressbooks to see the effect. Otherwise the query optimizer will just make obvious decisions. @PVince81 Please take over. Maybe @mrow4a wants to build a test for this? @Helios07 might be able to give working examples.

mrow4a:
@butonic about what kind of test do you think? That should run once to validate the PR? Or something continuous? Or is it just about validating the PR?

To test it it would be good to mimic some big DB with "standard usage". I will check what "stats" (summarize queries to execute on big instance) do I need to request from @Helios07 to reconstruct such table with dummy entries.

@PVince81 does it have some priority?

Helios07:
@mrow4a We have about 100k accounts distributed in almost 30 address books. Looking up a user name takes about four seconds. What data can I provide you else?

mrow4a:
@PVince81 @butonic @Helios07 I will validate the PR creating large pool of users with https://github.com/owncloud/testing/pull/89. 

@Helios07 how many contacts per user do you have in average/median (should I prepare a query for you to summarize this information)? What exactly action do you take which is slow, and you would expect to be fast (asking in the context of ensuring we have proper index for your slow action)?

Helios07:
@mrow4a A query would be helpful.
What I want to speed up is the normal sharing dialogue. After finishing typing, you have to wait for at least four seconds for a result. It happens rather often that users trigger the search while still typing (waiting or correcting typos) and when you continue during this search, you will get no result despite of entering the correct name.

DeepDiver1975:

mrow4a:
Reopening due to https://github.com/owncloud/enterprise/issues/3625. The work should add index for addressbookid_name_value and add medial_search flag that allows to switch from medial term wildcard (%term%) to end of term wildcard (term%) in order to utilize index on value. Targeting 10.5 (@micbar @butonic )

related: 
- the same but in accounts (https://github.com/owncloud/core/blob/v10.3.0/lib/private/User/AccountMapper.php#L156-L163)
- migration for accounts (https://github.com/owncloud/core/blob/v10.3.0/core/Migrations/Version20170516100103.php#L60-L71)
- wildcard performance reading (https://use-the-index-luke.com/sql/where-clause/searching-for-ranges/like-performance-tuning)

mrow4a:
- [x] medial search flag has been already added https://github.com/owncloud/core/commit/3bc8057c5a115ff206b1dc3d4a505028495c4113
- [x] rabase and test @butonic commits

mrow4a:
@micbar target 10.5

apps/dav/lib/CardDAV/CardDavBackend.php
```diff
@@ -853,21 +849,18 @@ public function searchEx($addressBookId, $pattern, $searchProperties, $options,
...
+		$query->select(['c.uri', 'c.carddata'])
+			->from($this->dbCardsTable, 'c')
+			->join('c', $this->dbCardsPropertiesTable, 'cp', $query->expr()->eq('c.id', 'cp.cardid'))
+			->where($query->expr()->eq('cp.addressbookid', $query->createNamedParameter($addressBookId)))
+			->andWhere($query->expr()->in('cp.name', \array_map(function ($property) use ($query) {
+				return $query->createNamedParameter($property);
+			}, $searchProperties)))
+			->andWhere($query->expr()->iLike('cp.value', $query->createNamedParameter($searchPattern)))
+			->setFirstResult($offset)->setMaxResults($limit)
+			->groupBy(['c.uri']) // groupby without aggregation is essentially distinct
```
mrow4a: @butonic in your commit, you did use `->groupBy(['c.uri', 'c.carddata'])` that is not supported by ORACLE (you cannot groupby on BLOB). Is it possible that there are duplicates? If this is about VCARD we can probably also try `->groupBy(['c.uri', 'size'])` . Or maybe not groupby at all?

mrow4a:
Due to lack of knowledge about the app and legacy considerations regarding query, I decided not to continue this work in a given form in order not to break something. Just index has been added in https://github.com/owncloud/core/pull/37152.

If someone decides to continue with this PR, it fails due to `->groupBy(['c.uri', 'c.carddata'])`, as `c.carddata` is BLOB and running groupby on BLOB is not allowed in Oracle. One solution is to groupby on different terms that give distinct results, but requires legacy considerations.
```
/drone/src/apps/dav/tests/unit/CardDAV/CardDavBackendTest.php:552
348	
349	Caused by
350	Doctrine\DBAL\Driver\OCI8\OCI8Exception: ORA-00932: inconsistent datatypes: expected - got BLOB
```

butonic:
I'm not looking into this.

PR #30200, opened 2018-01-19
Fix path resolution for appdirs above ocwebroot.

### LINKED ISSUE #28778: ThemeService can't handle Themes in separate app directories
### Steps to reproduce
1. install a fresh oc instance
2. configure an external app directory
3. put a theme in this external app directory

### Expected behaviour
Theme can be enabled on the admin->apps page and works as expected

### Actual behaviour
Theme doesn't work

Related: https://github.com/owncloud/theme-enterprise/issues/5

VicDeo:
## Description
Allows app-theme to be in the directory located above oC webroot
Note: this app directory should be accessible from the web to serve js/css files properly

Checklist
- [x]  CSS
- [x] JS
- [x] templates
- [x] frontend translations *.js 
- [x] backend translations *.json 
- [x] Image path

## Related Issue
Fixes https://github.com/owncloud/core/issues/28778

## How Has This Been Tested?
Sample config:
```
  'apps_paths' => 
  array (
    0 => 
    array (
      'path' => '/var/www/oc-tmp/apps',
      'url' => '/apps',
      'writable' => false,
    ),
    1 => 
    array (
      'path' => '/var/www/appsinhell',
      'url' => '../appsinhell',
      'writable' => true,
    ),
  ),
```

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -523,7 +523,12 @@ class AppManager implements IAppManager {
 	 */
 	public function getAppWebPath($appId) {
 		if (($appRoot = $this->findAppInDirectories($appId)) !== false) {
-			return \OC::$WEBROOT . $appRoot['url'];
+			$ocWebRoot = \OC::$WEBROOT;
+			while (strpos($appRoot['url'], '..') === 0) {
+				$appRoot['url'] = substr($appRoot['url'],3);
+				$ocWebRoot = dirname($ocWebRoot);
+			}
+			return $ocWebRoot . '/' . $appRoot['url'];
 		}
 		return false;
 	}
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -523,12 +523,14 @@ class AppManager implements IAppManager {
 	 */
 	public function getAppWebPath($appId) {
 		if (($appRoot = $this->findAppInDirectories($appId)) !== false) {
-			$ocWebRoot = \OC::$WEBROOT;
-			while (strpos($appRoot['url'], '..') === 0) {
-				$appRoot['url'] = substr($appRoot['url'],3);
+			$ocWebRoot = $this->getOcWebRoot();
+			// consider all relative ../ in the app web path as an adjustment
+			// for oC web root
+			while (strpos($appRoot['url'], '../') === 0) {
+				$appRoot['url'] = substr($appRoot['url'], 3);
 				$ocWebRoot = dirname($ocWebRoot);
 			}
-			return $ocWebRoot . '/' . $appRoot['url'];
+			return $ocWebRoot . '/' . ltrim($appRoot['url'], '/');
 		}
 		return false;
 	}
@@ -582,6 +584,15 @@ class AppManager implements IAppManager {
 	 */
 	protected function saveAppPath($appId, $appData) {
 		$this->appDirs[$appId] = $appData;
+	}
+
+	/**
+	 * Get OC web root
+	 * Wrapper for easy mocking
+	 * @return string
+	 */
+	protected function getOcWebRoot() {
+		return \OC::$WEBROOT;
 	}
 
 	/**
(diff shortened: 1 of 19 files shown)


PVince81:
I think I understand what you're trying to do but the lack of code comments make it confusing.

Mind decorating your loops, strpos and substr related code with more code comments ?
and add an example here: https://github.com/owncloud/core/pull/30200/files#diff-8fa91b147fc76360cb3cfca41a1db57dR528

The point here is that trying to compute all these strpos and substr directly in my head gives me a headache, so if you add an example path there it will help 😄 

Overall it looks like it's going in a good direction!

PVince81:
let us know when ready for review again

VicDeo:
@PVince81 it's already working but I'm trying to increase coverage for my changes.
To do it properly those locators should be refactored into a test-friendly shape.

lib/private/App/AppManager.php
```diff
@@ -523,7 +523,14 @@ public function getAppPath($appId) {
 	 */
 	public function getAppWebPath($appId) {
 		if (($appRoot = $this->findAppInDirectories($appId)) !== false) {
-			return \OC::$WEBROOT . $appRoot['url'];
+			$ocWebRoot = $this->getOcWebRoot();
+			// consider all relative ../ in the app web path as an adjustment
+			// for oC web root
+			while (strpos($appRoot['url'], '../') === 0) {
```
PVince81: is there a risk of appRoot giving access to arbitrary files ? or is appRoot defined internally in a safe manner ?

lib/private/Theme/ThemeService.php
```diff
@@ -66,65 +71,49 @@ public function getTheme() {
...
-	private function getDirectory($themeName, $appTheme = true) {
+	private function makeTheme($themeName, $appTheme = true) {
+		$baseDirectory = $this->serverRoot;
+		$directory = '';
+		$webPath = '';
 		if ($themeName !== '') {
 			if ($appTheme) {
-				return substr(\OC_App::getAppPath($themeName), strlen(\OC::$SERVERROOT) + 1);
+				$themeDirectory = \OC_App::getAppPath($themeName);
+				if (strpos($themeDirectory, $this->serverRoot)===0) {
```
PVince81: add comment to explain what is happening here and why

sadly string operations in PHP are difficult to read

lib/private/App/AppManager.php
VicDeo: There is always a risk... $appRoot comes from here https://github.com/owncloud/core/blob/master/config/config.sample.php#L721

lib/public/Theme/ITheme.php
```diff
@@ -32,6 +32,12 @@
 	 */
 	public function getName();
 
+	/**
+	 * @return string
+	 * @since 10.0.7
```
PVince81: that will be 10.0.8 now thanks to the hotfix

PVince81 (approved):
Looks good, please add missing comment for clarity. Nice work ! 👍 

Mind double checking that at no point this could be abused to get access to parent directories from outside ?

lib/private/App/AppManager.php
VicDeo: `webPath` is not a path but a relative url, it  likes an app-specific `slug` - something that will become an URL used to locate an asses in browser

VicDeo:
@PVince81 I can't invent abusing usecase unless admin injected something into `apps_paths` via `config.php`
it is supposed to work in the manner from the PR description:
```
  'apps_paths' => 
  array (
    0 => 
    array (
      'path' => '/var/www/oc-tmp/apps',
      'url' => '/apps',
      'writable' => false,
    ),
    1 => 
    array (
      'path' => '/var/www/appsinhell',
      'url' => '../appsinhell',
      'writable' => true,
    ),
  ),
```
these dots are used for URL construction so if anything could be accessed by the resulting URL - it could be accessed with no PHP at all.

PVince81:
@VicDeo thanks for the confirmation. Ok to merge then

VicDeo:
Stable10: https://github.com/owncloud/core/pull/30477

VicDeo:
CI is green

Issue #30190, opened 2018-01-19
InfoParser should throw an exception on failure to parse


tomneedham:
### Steps to reproduce
1. Make some invalid xml in info.xml of an app
2. Run ownCloud
3.

### Expected behaviour
Bad things. Log messages at least

### Actual behaviour
getAppInfo returns null, getAppVersion then returns 0, shouldUpgradeApp always returns fall, migrations dont run, get exceptions because my app is looking for tables that dont exist. 10 minute debugging to find the reason.

VicDeo:
Fixed in master with https://github.com/owncloud/core/pull/30457
backport was not approved

PVince81:
backport reconsidered, reopening until done

VicDeo:
Backported

PR #30204, opened 2018-01-19
[WIP] Explode on info.xml problems

butonic:
fixing https://github.com/owncloud/core/issues/30190

based on https://github.com/owncloud/core/pull/30188

cc @Tom @VicDeo @PVince81 lets see what breaks

### DIFF
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -32,6 +32,7 @@ namespace OC\App;
 
 use OC_App;
 use OC\Installer;
+use OCP\App\AppNotFoundException;
 use OCP\App\IAppManager;
 use OCP\App\AppManagerException;
 use OCP\App\ManagerEvent;
@@ -359,7 +360,11 @@ class AppManager implements IAppManager {
 	 * @internal
 	 */
 	public function getAppInfo($appId) {
-		$appInfo = \OC_App::getAppInfo($appId);
+		try {
+			$appInfo = \OC_App::getAppInfo($appId);
+		} catch (AppNotFoundException $e) {
+			return null;
+		}
 		if ($appInfo === null) {
 			return null;
 		}
@@ -519,6 +524,7 @@ class AppManager implements IAppManager {
 	 *
 	 * @param string $appId
 	 * @return string|false
+	 * @throws AppNotFoundException
 	 * @since 10.0.5
 	 */
 	public function getAppWebPath($appId) {
--- a/lib/private/App/InfoParser.php
+++ b/lib/private/App/InfoParser.php
@@ -30,10 +30,12 @@ class InfoParser {
 	/**
 	 * @param string $file the xml file to be loaded
 	 * @return null|array where null is an indicator for an error
+	 * @throws \NotFoundException if file does not exist
+	 * @throws \InvalidArgumentException on malformed XML
 	 */
 	public function parse($file) {
 		if (!file_exists($file)) {
-			return null;
+			throw new \NotFoundException($file);
 		}
 
 		libxml_use_internal_errors(true);
@@ -43,12 +45,12 @@ class InfoParser {
 		libxml_disable_entity_loader($loadEntities);
 		if ($xml === false) {
 			libxml_clear_errors();
-			return null;
+			throw new \InvalidArgumentException('Invalid XML');
 		}
 		$array = $this->xmlToArray($xml);
 
 		if (is_null($array)) {
-			return null;
+			throw new \InvalidArgumentException('Could not convert XML to array');
 		}
 
 		if (!array_key_exists('info', $array)) {
--- a/lib/private/App/InfoParser.php
+++ b/lib/private/App/InfoParser.php
@@ -25,17 +25,19 @@
 
 namespace OC\App;
 
+use OCP\App\AppNotFoundException;
+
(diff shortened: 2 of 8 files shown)


lib/private/App/InfoParser.php
```diff
@@ -30,10 +30,12 @@ class InfoParser {
 	/**
 	 * @param string $file the xml file to be loaded
 	 * @return null|array where null is an indicator for an error
+	 * @throws \NotFoundException oif file does not exist
```
patrickjahns: typo

butonic:
@VicDeo @PVince81 @tomneedham I had a look at this ... a lot of places need touching if we want to make oc explode if an app contains invalid XML. Also apps that no longer exist are a problem because the version reported for them is `0` which AFAIU is the problem for migrations not happening ...

Also, should we introduce a new axcaptions not existing on disk or should we reuse `OCP\App\AppNotFoundException`? The phpdoc indicates it is for apps not found in the market

feel free to highjack this PR. I doubt I will get to it during the week. I had hoped it would be a smaller task and I wanted to build it upon https://github.com/owncloud/core/pull/30188 befor @VicDeo might run off in a direction that would conflict with the changes in that PR ...

PVince81:
is all the etag stuff really necessary just for throwing exceptions on error ?

lib/private/legacy/app.php
```diff
@@ -616,44 +626,122 @@ public static function getAppVersionByPath($path) {
...
+			return "${stat['mtime']}|${stat['ino']}|${stat['dev']}|${stat['size']}";
+		}
+		return false;
+	}
+
+	/**
+	 * if available use a local cache that survives requests, otherwise
+	 * fallback to an in memory implementation of \OCP\ICache
+	 */
+	static private function setupAppInfoCache(){
```
VicDeo: I don't like caches in legacy classes...

lib/private/App/AppManager.php
```diff
@@ -359,7 +360,11 @@ public function getAppsNeedingUpgrade($ocVersion) {
 	 * @internal
 	 */
 	public function getAppInfo($appId) {
-		$appInfo = \OC_App::getAppInfo($appId);
+		try {
+			$appInfo = \OC_App::getAppInfo($appId);
+		} catch (AppNotFoundException $e) {
+			return null;
```
VicDeo: Invalidates PHPDoc... Bah, it's already outdated.

lib/private/App/InfoParser.php
```diff
@@ -43,12 +47,12 @@ public function parse($file) {
 		libxml_disable_entity_loader($loadEntities);
 		if ($xml === false) {
 			libxml_clear_errors();
-			return null;
+			throw new \InvalidArgumentException('Invalid XML');
```
VicDeo: Invalidates PHPDoc - no sense in returning null on error if we are moving to an error control based on exceptions

lib/private/App/AppManager.php
```diff
@@ -519,6 +524,7 @@ public function getAppPath($appId) {
 	 *
 	 * @param string $appId
 	 * @return string|false
+	 * @throws AppNotFoundException
```
VicDeo: Backward incompatible. Should go into interface as well

tests/lib/App/InfoParserTest.php
```diff
@@ -22,23 +22,17 @@ public function setUp() {
...
-	function providesInfoXml() {
-		return [
-			['expected-info.json', 'valid-info.xml'],
-			[null, 'invalid-info.xml'],
-		];
+	/**
+	 * @expectedException \InvalidArgumentException
+	 */
+	public function testParsingInvalidXml() {
+		$this->parser->parse(OC::$SERVERROOT. "/tests/data/app/valid-info.xml");
```
VicDeo: the same filename as in `testParsingValidXml`, is this file able to mutate from valid into invalid state?

PVince81:
is this obsoleted by https://github.com/owncloud/core/pull/30457 ? @VicDeo

VicDeo:
> is this obsoleted by #30457

@PVince81 seems so. At least [this] (https://github.com/owncloud/core/pull/30204/commits/3817c14e234e308b7eda8890d4be375187a649cf) 
and [this](https://github.com/owncloud/core/pull/30204/commits/4bb366ef50d8d3b23d8c35e381bcc1b8506139ff)
commits have the same purpose as https://github.com/owncloud/core/pull/30457

[First one](https://github.com/owncloud/core/pull/30204/commits/a0c381daf6d62be8ec3b7fbdb9ea3450e4dfb087) is similar to https://github.com/owncloud/core/pull/30188

In addition it was created before cs-fixer so it conflicts both with cs-fixer and https://github.com/owncloud/core/pull/30457. I think it's better close this PR and move some parts into  https://github.com/owncloud/core/issues/30190 (if needed).

VicDeo:
https://github.com/owncloud/core/issues/30190 is already fixed in master and has a pending backport.

Closed in favor of https://github.com/owncloud/core/pull/30188

butonic:
[as I said](https://github.com/owncloud/core/pull/30204#issue-164077774), this is based on #30188

I used a separate PR to do things incrementally. I'll go for #30188 now

PR #30099, opened 2018-01-11
Path needs adjustments so that the code can be viewed online

DeepDiver1975:
## Description
Code should be viewable on codecov online ....

## Motivation and Context
Try to fix this ...
bildschirmfoto von 2018-01-11 13-04-05

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30205, opened 2018-01-20
fix codecov paths

butonic:
I want to know which lines are (not) covered.

see https://docs.codecov.io/docs/fixing-paths


butonic:
closing in favor of https://github.com/owncloud/core/pull/30099

PR #30206, opened 2018-01-21
Use codecov bash to upload coverage information

DeepDiver1975:
The drone plugin to upload clover reports to codecov.io misses one step and this is where paths are adjusted.

This PR now uses the official bash script as distributed by codecov.io

This is where the bash script adjusts the paths - https://github.com/codecov/codecov-bash/blob/15cdf48a5c270730a650f853b26869304b2cc379/codecov#L1283-L1433

This is missing in https://github.com/drone-plugins/drone-codecov - I have currently no intent to report a bug report there because this plugin is just plain wrong from my pov - no interest in fighting the devs ;-)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -73,12 +73,13 @@ pipeline:
         TEST_SUITE: coverage
 
   codecov:
-    image: plugins/codecov:1
-    secrets: [codecov_token]
+    image: owncloudci/php:${PHP_VERSION}
     pull: true
-    files:
-     - tests/autotest-clover-${DB_TYPE}.xml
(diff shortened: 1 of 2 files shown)


tboerger:
Good to know that you want to drop stuff that had been even written by your colleagues.

I'm against this change, just give me a proper report. I have written this plugin based on the code paths from the shell script.

DeepDiver1975:
> I'm against this change, just give me a proper report. I have written this plugin based on the code paths from the shell script.

Just use the shell script - I see no reason to reimplement this.

DeepDiver1975:
nice and now our coverage is back at 60% as well .... there is much more magic happenind in the bash script?

PR #30182, opened 2018-01-18
move UI tests to drone

individual-it:
1. some improvements on the UI tests infrastructure (was needed because we run on different timing issues in drone - drone is just sooo fast)
2. run the UI tests in drone

ToDo:

- [x] change from `arturneumann/owncloud-ci-php` to `owncloudci/php` as soon as its build the version after https://github.com/owncloud-ci/php/pull/13

PR #30209, opened 2018-01-21
[stable10] UI tests in drone

phil-davis:
Backport #30182


phil-davis:
Note: this will implement UI tests on drone for stable10.
Currently UI tests are being run in stable10 in Travis+SauceLabs - so it needs to be decided how long to let the UI tests run in both systems.

individual-it (approved):

PR #30221, opened 2018-01-22
[WIP] Gherkin share feature with Gherkin syntax

PVince81:
Met with @patrickjahns and @pmaier1 and we tried to practice and think about how to write about local share creation and also accepting shares (https://github.com/owncloud/core/issues/19153)

Some comments/issues:
- features must be split into separate files
- see inline comments in the code
- should we have "overview" features which are only for people with no actual execution ? Ex: creating a share triggers a lot of things like activity, log, hooks, notification, etc and having all these in separate features makes people lose the overview, so it might be good to have one summary description that gathers all these things additionally to the detailed feature.
- Describe features from different POVs (e.g. share sender/receiver)
- Background definitions
- the bottom scenarios are oldest and are not complete, the top ones are the most refined so far
- where to put glossary
- where to put these files

Discussion about general guidelines here: https://github.com/owncloud/QA/issues/517

@phil-davis @davitol @SergioBertolinSG @individual-it

### DIFF
--- a/tests/features/sharing/local_user_share_creation.feature
+++ b/tests/features/sharing/local_user_share_creation.feature
@@ -1,0 +1,91 @@
+# FIXME: split to multiple feature files!	
+Background:
+	Given default share settings are set
+
+Feature: As a regular user shares with another local user (DB/technical)
+	Scenario: a regular user shares with another existing user (overview feature?)
+		Given a folder "test" exists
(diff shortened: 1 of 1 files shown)


PVince81:
to be revived in some other form

in any case, the QA team already settled on a syntax

PR #30233, opened 2018-01-23
[WIP] Possible new formats for given-when-then step text

phil-davis:
Some suggested examples for step text.

Problem 1:
```
Given parameter "shareapi_enabled" of app "core" is set to "yes"
When parameter "shareapi_enabled" of app "core" is set to "yes"
Then parameter "shareapi_enabled" of app "core" is set to "yes"
```
The above used in ``Given`` or ``When`` looks like a step that will bring the system to the specified state. Used in ``Then`` it looks like a test which will check that the parameter is already in that state. But the code implementing the step is going to do one or other of these. So step text like this is ambiguous and thus gets sometimes used wrongly. Using the verb "is" in step text usually results in it being ambiguous about whether the step should change the state or test the state.

Suggested way to re-write:
```
Given parameter "shareapi_enabled" of app "core" has been set to "yes"
When the administrator sets parameter "shareapi_enabled" of app "core" to "yes"
Then parameter "shareapi_enabled" of app "core" should be set to "yes"
```
In a ``Given`` use present perfect tense in the passive.
In a ``When`` use present tense in the active.
For these, write code that brings the system to the desired state.

In a ``Then`` use the word "should" rather than "is", making it clear that a test of state is expected.
For this, write code that asserts something and can fail.

And prohibit step text that uses the verb "is".

Problem 2:
We have various ways of doing state changes and of seeing the results. When we are just setting up state ready for the real thing we want to test, then we do not care much how the test code gets the system to the desired state. But when the test is supposed to be testing that something can be done using a particular interface of the system, then we need to specify which interface to use. In ownCloud we can typically change state or observe state with the following:
a) ``occ`` command
b) API (send POST, GET, DELETE etc)
c) clicking around in the webUI

```
When the administrator sets parameter "shareapi_enabled" of app "core" to "yes"
```
Maybe this step will use an ``occ`` command to set the parameter, or maybe it will use the API (or maybe there is somewhere on the webUI to make the setting and it will fire up a browser, login, navigate to a settings page and make the setting).

But in the particular scenario, we are (for example), intending to test the API - that config settings can be made successfully via the API. So we need to say that somewhere. We could say:
```
When the administrator sets parameter "shareapi_enabled" of app "core" to "yes" using the API
```

If we want to test that the ``occ`` way of setting the parameter works, then we could say:
```
When the administrator sets parameter "shareapi_enabled" of app "core" to "yes" using the occ command
```

and if there is a webUI way to do the setting:
```
When the administrator sets parameter "shareapi_enabled" of app "core" to "yes" using the webUI
```

And a similar set of ``using the`` can be done for ``Then`` steps which test something - you might be wanting to test that the system state can be observed correctly using ``occ``, API or webUI.


phil-davis:
@patrickjahns @davitol @SergioBertolinSG @individual-it 
for discussion.

SergioBertolinSG (approved):
I like this format.

It has some problems though, like having too long steps and having to insist to other test devs about following it.

phil-davis:
Yes, the step text gets longer. If anyone can think of ways to make things shorter but still "good" then please suggest.

If we set some standards, then we can document them for everyone to use and "enforce" during code reviews.

patrickjahns:
LGTM

@PVince81
How do you feel as developer about these changes - are they concrete for you? 

@pmaier1  
As a PM - do you understand these feature files and steps? How would you interpret them? Is there anything you need clarification on?

phil-davis:
Looking through the rest of the capabilities.feature as an example of what could be refactored, I have:
1) Role vs user name:
```
Given as an "admin"
Given as an "user1"
```
reads as if it is a role/actor like:
```
Given as an accountant
Given as a manager
```
but actually this Gherkin step (in our usage) is specifying an actual user name to be used for the scenario.
So change the wording of the step to:
```
Given as user "admin"
Given as user "user1"
```

Then some time in the future if relevant/needed we would be able to use ``Given as an accountant`` style steps to indicate a role.

2) Let ``API`` be specified in upper or lower case.

3) Refactor step text that retrieves capabilities to use active grammar

4) Refactor step text that checks capabilities content to use the word "should"

(3 and 4 are more examples of the general policy above)

SergioBertolinSG:
Re-reviewed - ✅

patrickjahns:
@PVince81
Any input from your side?

tests/integration/features/caldav.feature
```diff
@@ -2,33 +2,33 @@ Feature: caldav
...
-		And the error message is "Node with name 'MyCalendar' could not be found"
+		And the exception should be "Sabre\DAV\Exception\NotFound"
+		And the error message should be "Node with name 'MyCalendar' could not be found"
 
 	@caldav
 	Scenario: Accessing a not shared calendar of another user
-		Given user "user0" exists
-		And "admin" creates a calendar named "MyCalendar"
+		Given user "user0" has been created
+		And user "admin" has created a calendar named "MyCalendar"
```
PVince81: `has successfully created` and then remove the 201 below.

for the next test `has failed creating` to match the 404

tests/integration/features/caldav.feature
```diff
@@ -2,33 +2,33 @@ Feature: caldav
...
+		And the exception should be "Sabre\DAV\Exception\NotFound"
+		And the error message should be "Node with name 'MyCalendar' could not be found"
 
 	@caldav
 	Scenario: Creating a new calendar
-		When "admin" creates a calendar named "MyCalendar"
-		Then the CalDAV HTTP status code should be "201"
-		And "admin" requests calendar "admin/MyCalendar"
+		Given user "admin" has created a calendar named "MyCalendar"
+		And the CalDAV HTTP status code should be "201"
```
PVince81: same, remove 201 here and have it say `has successfully created`

tests/integration/features/dav-versions.feature
```diff
@@ -2,26 +2,26 @@ Feature: dav-versions
...
 
   Scenario: Remove a file
-    Given user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
-    And user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    Given user "user0" has uploaded file "data/davtest.txt" to "/davtest.txt"
+    And user "user0" has uploaded file "data/davtest.txt" to "/davtest.txt"
     And the version folder of file "/davtest.txt" for user "user0" contains "1" elements
     And user "user0" deletes file "/davtest.txt"
-    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt"
+    When user "user0" uploads file "data/davtest.txt" to "/davtest.txt" using the API
```
PVince81: Is "using the API" important here or the exact mechanism for upload ? not sure. At least if this was on a higher level, it wouldn't matter on the feature level. However, technically, we likely want to test multiple upload mechanisms to make sure this works.

PVince81 (commented):
A great step forward 👍

tests/integration/features/dav-versions.feature
individual-it: > When an action is the "action under test" then I am explicitly saying using the API (or similar) in the step text. That allows us to be clear about test steps that test, for example, creating a user "using occ", "using the API" and "using the webUI".

that way we can one day merge all tests that we call now UI and integration into one "acceptance" suite, because that is really what they are

PVince81: > When a step is a "given" then it is just for setting up state for the test, so we do not care to be specific about how the state is achieved (the underlying code is free to use an occ command, use the API or even (heaven forbid) fire up the webUI to do the job). 

Ah yes, that makes sense 👍

patrickjahns: I would find it very compelling to have our feature files be "more general" and the context specify wether or not the same feature will be tested on an API level, or on a UI level

CC: https://blog.eleven-labs.com/en/behat-structure-functional-tests/

tests/integration/features/caldav.feature
phil-davis: CalDav and CardDav integration test steps have been adjusted for this, and to make them consistent with each other.
There is a bit of duplication between CalDav and CardDav with "the [CalDav|CardDav] HTTP status should be" etc. Those could be put in a more shared file some day and the steps merged.

tests/integration/features/caldav.feature
phil-davis: done

phil-davis:
@SergioBertolinSG @davitol @individual-it @PVince81 @DeepDiver1975 - a minor thing:
Some Behat-Gherkin feature files are indented with 2 spaces, others are indented with tabs.

Do we want to set a standard for that?

If so, which?

SergioBertolinSG:
I think tabs, to use the same as the steps code.

phil-davis:
See #30413 for a squashed version of this for review and merge.

phil-davis:
Closing here - make you review in the squashed version.

PR #30248, opened 2018-01-24
Do not run chrome UI for stable10 on Travis

phil-davis:
because these are now running OK in drone.

PR #30249, opened 2018-01-24
[stable10] Do not run chrome UI for stable10 on Travis

phil-davis:
Backport #30248

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,70 +57,70 @@ script:
 matrix:
   include:
     - php: 5.6
-      if: branch = stable10 OR type = cron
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Note: UI tests are being run on ``stable10`` backports for a while (and also in drone). This is convenient because Travis+SauceLabs gives access to video of the UI test, so it is easier to work out what went wrong when a test fails. It also provides comparison with drone until we are confident in all that.

This backport can wait here for a while...

patrickjahns (approved):

PR #30245, opened 2018-01-23
Improve assertions

carusogabriel:
## Description
Following #29745.

Thanks to [`Rector`](https://github.com/rectorphp/rector) that helped me find them.

## Related Issue
None
## Motivation and Context
Improve PHPUnit assertions, given better error messages in failures.

## How Has This Been Tested?
PHPUnit

## Screenshots (if appropriate):
None

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests pass.

PR #30250, opened 2018-01-24
[stable10] Improve assertions

phil-davis:
Backport #30245 

since we run unit tests in stable10 also, and keeping stable10 close to master will minimize potential for future conflicts when backporting other stuff


phil-davis:
@DeepDiver1975 this should be an easy approve and merge...

DeepDiver1975 (approved):

PR #30257, opened 2018-01-24
[stable10 backport] before create share called too early

IljaN:


IljaN:
Restarting build because MySQL Server has gone away :-/

IljaN:
Merge failed again :(

PR #30260, opened 2018-01-24
ltrim is more expensive than you think

butonic:
was profiling the notification call:

ltrim

lets see if this breaks the testsuite.


butonic:
already covered 100% by tests ... https://github.com/owncloud/core/blob/ce4a435ed27db6f1ad4d123271898ea89be99744/tests/lib/AppFramework/Utility/SimpleContainerTest.php#L172-L189

no idea why codecov complains

DeepDiver1975 (approved):

PR #30267, opened 2018-01-25
[stable10] Use mysql in phpunit coverage run

patrickjahns:
# Motivation

Mariadb is failing during the codecoverage run - in `master` we already switched to mysql - so let's do this with mysql here as well


DeepDiver1975 (approved):

patrickjahns:
killed travis -> no affect on travis when changing a .drone.yml

waiting for drone to pass and then merging

phil-davis (approved):

PR #30251, opened 2018-01-24
When executing the data-fingerprint command a log entry is written to…

DeepDiver1975:
… allow analysis who executed the command and when 

## Description
Executing the data fingerprint command has some large impact on the system.
The command should be logged for later analysis who executed the command.

In addition it is worth a second user confirmation before applying the data-fingerprint.

## Related Issue
refs #30154

## Motivation and Context
Make owncloud a bit more robust

## How Has This Been Tested?
- manual
- unit tests

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #30154, opened 2018-01-16
Backup detected


antonioderrico:
Hello everyone,

some of our users are affected by errors which occur in the owncloud desktop-client.
we do not change anything on the server and not imported any backups.

![image](https://user-images.githubusercontent.com/22045141/34982463-bcd6031c-faaa-11e7-88d3-ac2793ae3d74.png)

![image](https://user-images.githubusercontent.com/22045141/34982478-c9ba3bac-faaa-11e7-9b32-c5c75d7b1c7d.png)

OwnCloud Version = 9.0.9 (stable)
MySQL-Version = 5.1.73

In the log file, i can't find anything regarding this issue when the error on the client occurs.

DeepDiver1975:
> some of our users are affected by errors which occur in the owncloud desktop-client.
> we do not change anything on the server and not imported any backups.
> 
> image

this does not happen automatically. Someone on the server must have executed the data-fingerprint command - https://doc.owncloud.org/server/latest/admin_manual/configuration/server/occ_command.html?highlight=fingerprint#maintenance-commands

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

PR #30281, opened 2018-01-25
[stable10] When executing the data-fingerprint command a log entry is…

DeepDiver1975:
… written to allow analysis who executed the command and when - refs #30154

backport #30251


PVince81 (approved):
:+1:

Issue #27667, opened 2017-04-18
Users cannot set email address on personal settings


SergioBertolinSG:
### Steps to reproduce

1. Create a user "user1".
2. As "user1" go to settings personal/general.
3. Set email address.
4. Refresh page.

### Expected behaviour
"user1" sets an email address correctly.

### Actual behaviour
"user1" email address is ignored.

![error_email_personal_settings](https://cloud.githubusercontent.com/assets/9048796/25142761/62dc3f70-2468-11e7-9e7a-579218ffb4bf.gif)

### Server configuration

**Operating system**:
Ubuntu 14.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

**ownCloud version:** (see ownCloud admin page)
current master
{"installed":"true","maintenance":"false","needsDbUpgrade":"false","version":"10.0.0.6","versionstring":"10.0.0 beta2","edition":"Community","productname":"ownCloud"}

**Updated from an older ownCloud or fresh install:**
Fresh

**The content of config/config.php:**

```

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
No.

**Are you using encryption:** 
No

**Logs**

### Client configuration

**Browser**
Chrome

**Logs**
```
```

PVince81:
It's only saved once you confirm your email...

This also bothers me a bit.

@imujjwal96 another case for later

moving this to 10.0.1

c-holtermann:
At the moment I can't change an email at all because even if I go to the link of the validation email which gets sent correctly then email address doesn't get changed.

jdopson:
I'm also having this issue with 10.0.3 and 10.0.4.  It affects both internal and external users.  The admin user is able to change their password but no one else.

dekompeti:
I have the same problem with Owncloud 10.0.3.
Owncloud will fix this problem with an 10.0.5 upgrade?

Regards,

vborges87:
I have the same problem... Owncloud 10.0.3.

PVince81:
@jdopson @vborges87 can you clarify "same problem" ? Do you mean the one from https://github.com/owncloud/core/issues/27667#issuecomment-342516272 about the validation email not saving the confirmed email address ?

It would be good if you guys provide more details about your setup and the logs. This feature works correctly here after the confirmation email. I wonder if there are special characters in your email address that might make the process fail ? The logs might show more information.

Taz0:
[...]

PR #30301, opened 2018-01-29
User should be able to set the email address

sharidas:
User logged in should be able to set his/her
email address.

## Description

Users were not able to set the email address due to the regression caused at the commit: https://github.com/owncloud/core/commit/b73b488cc91803d2236a9f1ff7ac7d78b6b1b1fa

## Related Issue

https://github.com/owncloud/core/issues/27667

## Motivation and Context

Added an extra condition to check if the user logged in is trying to set the email. And it should be possible for the user to set the email address.

## How Has This Been Tested?

- [x] Created user `user1`
- [x] Logged in as `user1`
- [x] And then set the email address
- [x] Verified the confirmaition email has arrived
- [x] Use the confirmation link got in the email to verify the email address.
- [x] Refresh the `user1` settings page to see if the email is shown in the email text box.
- [x] Verified that email is updated.

## Types of changes

## Checklist:

### DIFF
--- a/settings/Controller/UsersController.php
+++ b/settings/Controller/UsersController.php
@@ -620,7 +620,7 @@ class UsersController extends Controller {
 						'data' => [
 							'username' => $id,
 							'mailAddress' => $mailAddress,
-							'message' => (string) $this->l10n->t('An email has been sent to this address for confirmation')
+							'message' => (string) $this->l10n->t('An email has been sent to this address for confirmation. Until the email is verified this address will not be set.')
 						]
 					],
 					Http::STATUS_OK
@@ -792,6 +792,7 @@ class UsersController extends Controller {
 	 *
 	 * @param string $id
  	 * @param string $mailAddress
+	 * @return JSONResponse
  	 */
 	public function setEmailAddress($id, $mailAddress) {
 		$user = $this->userManager->get($id);
@@ -799,11 +800,13 @@ class UsersController extends Controller {
 		// Only Admin and SubAdmins are allowed to set email
 		if($this->isAdmin ||
 			($this->groupManager->getSubAdmin()->isSubAdmin($this->userSession->getUser()) &&
-				$this->groupManager->getSubAdmin()->isUserAccessible($this->userSession->getUser(), $user))) {
+				$this->groupManager->getSubAdmin()->isUserAccessible($this->userSession->getUser(), $user)) ||
+				($this->userSession->getUser()->getUID() === $id)) {
 			$user->setEMailAddress($mailAddress);
 			if ($this->config->getUserValue($id, 'owncloud', 'changeMail') !== '') {
 				$this->config->deleteUserValue($id, 'owncloud', 'changeMail');
 			}
+			return new JSONResponse();
 		} else {
 			return new JSONResponse([
 				'error' => 'cannotSetEmailAddress',
@@ -830,7 +833,6 @@ class UsersController extends Controller {
 			$this->log->error("The logged in user is different than expected.", ['app' => 'settings']);
 			return new RedirectResponse($this->urlGenerator->linkToRoute('settings.SettingsPage.getPersonal', ['changestatus' => 'error']));
 		}
-
 		try {
 			$this->checkEmailChangeToken($token, $userId);
 		} catch (\Exception $e) {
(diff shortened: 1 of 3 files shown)


SergioBertolinSG:
It works when validating the email. 

@PVince81 should this work also when the email is not validated? If this is a case I think there should be a message telling the user that the email will be removed if it is not validated inmediately.

settings/Controller/UsersController.php
```diff
@@ -792,18 +792,21 @@ public function sendEmail($userId, $mailAddress) {
...
  	 */
 	public function setEmailAddress($id, $mailAddress) {
 		$user = $this->userManager->get($id);
 
 		// Only Admin and SubAdmins are allowed to set email
 		if($this->isAdmin ||
 			($this->groupManager->getSubAdmin()->isSubAdmin($this->userSession->getUser()) &&
-				$this->groupManager->getSubAdmin()->isUserAccessible($this->userSession->getUser(), $user))) {
+				$this->groupManager->getSubAdmin()->isUserAccessible($this->userSession->getUser(), $user)) ||
+				($this->userSession->getUser()->getUID() === $id)) {
```
PVince81: note that the current user does not need to be logged in to validate their email. they just need to open the link.

will it work ?

sharidas: Yes, So if the user tries to open the link, after clicking the link, this will work.

SergioBertolinSG: It does not work without login. If there is no session active it opens the login page, if there is another user logged in, it fails. With a logged in user it works correctly.

sharidas:
Updated further. Removed the check which was done to verify if the user is logged in when the link from email is clicked. And hence removed the unit test which was associated with that test. Let me know if this looks ok. Also I have modified the string message.

settings/Controller/UsersController.php
PVince81: Can you try with the old version before the regression ? I don't remember whether we decided that login is required or not

sharidas: @PVince81  In the 10.0.0 version where this was working before, login is required.

PVince81: Ok, then let's keep that as the expected behavior

sharidas: @PVince81 Ok, then I will update the PR again. Because I removed lines https://github.com/owncloud/core/blob/master/settings/Controller/UsersController.php#L827-L832. Will bring back the unit test for the same which was removed.

sharidas:
Keeping the old behaviour as such, updated the PR.

settings/Controller/UsersController.php
PVince81: Thanks. In general, when in doubt: when a regression occurs, make it work like it worked before the regression.

SergioBertolinSG:
@sharidas testing this last version of this PR and accepting that the user must be logged to verify his address, I've seen something:

It works when you open the link in the same browser were you have the session open. But opening the link in a different browser without a session, fails. It opens the login page, and even after login with the user changing its email, it fails.

@PVince81 is this expected?

PVince81:
> even after login with the user changing its email, it fails.

this is a bug. Maybe the redirect doesn't properly redirect to the mail link.

sharidas:
@SergioBertolinSG Here is how I tested:
- [x] Open firefox and then try to login as `user1`
- [x] set the email address for `user1`
- [x] Now try to open the link obtained from the email ( sent to the respective email, in my case I used gmail ) in another browser. I used Chrome.
- [x] It opens the login page. And I am able to get the email address set for `user1`

For me it works.

sharidas:
gifrecord_2018-01-30_161528

Hope this helps.

SergioBertolinSG:
Retested, it was a problem with the email site I was using which was cutting the link. it works 👍 .

SergioBertolinSG (approved):
Just tested it and it works. No code review.

settings/Controller/UsersController.php
```diff
@@ -620,7 +620,7 @@ public function setMailAddress($id, $mailAddress) {
 						'data' => [
 							'username' => $id,
 							'mailAddress' => $mailAddress,
-							'message' => (string) $this->l10n->t('An email has been sent to this address for confirmation')
+							'message' => (string) $this->l10n->t('An email has been sent to this address for confirmation. Until email is verified this address will not be set.')
```
PVince81: @phil-davis quick wording check ?

PVince81 (approved):
:+1: code looks fine.

If @phil-davis has no objection to the wording this can be merged.

settings/Controller/UsersController.php
phil-davis: Could just add a "the":
```
Until the email is verified this address will not be set.
```

sharidas: @phil-davis  Thanks for the feedback. Updated. Yah that's true for CI

phil-davis (approved):
Good to merge when CI passes

phil-davis:
Backport stable10 #30319

PR #30328, opened 2018-01-31
Bump master version to 10.0.6

phil-davis:
forward "port" to match ``stable10``

Might as well get this the same as ``stable10`` before again bumping it for 10.1 or whatever.


PVince81 (approved):
:+1:

PR #30338, opened 2018-02-02
UI test for unmarking a file/folder from favorites page

jaggu4444:
## Description
UI tests for unmarking a file/folder which was previously marked as favorite.  

## Motivation and Context
UI tests to make sure that the favorite file/folder has been unmarked and removed from favorites page 

## How Has This Been Tested
Run tests locally

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/lib/FilesPageElement/FileRow.php
+++ b/tests/ui/features/lib/FilesPageElement/FileRow.php
@@ -316,7 +316,7 @@ class FileRow extends OwnCloudPage {
 	public function markAsFavorite() {
 		$element = $this->rowElement->find("xpath", $this->notMarkedFavoriteXpath);
 		if (is_null($element)) {
-			throw new Exception(
+			throw new ElementNotFoundException(
 				__METHOD__ .
 				" xpath $this->notMarkedFavoriteXpath not found"
 			);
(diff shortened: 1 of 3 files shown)


tests/ui/features/lib/FilesPageElement/FileRow.php
```diff
@@ -338,4 +338,20 @@ public function isMarkedAsFavorite() {
...
+	 */
+	public function unmarkFavorite() {
+		$element = $this->rowElement->find("xpath", $this->markedFavoriteXpath);
+		if (is_null($element)) {
+			throw new ElementNotFoundException(
+				__METHOD__ .
+				" xpath $this->markedFavoriteXpath not found"
+			);
+		}
+		$element->click();
```
phil-davis: This is OK for now. ``mark`` and ``unmark` are the same action (which is just clicking to change the state). So the ``markFavorite`` and ``unmarkFavorite`` methods are the same. If we write:
```
When I mark the file "data.zip"
And I mark the file "data.zip"
```
then actually the file will get marked and then unmarked.

We should think about what to do in these cases. e.g. in ``unmarkFavorite`` we could first check that the ``$element`` is currently selected/checked and throw an exception if not. That would help fail tests early if the expected current state is not as expected for the step being done.

phil-davis (approved):
Code looks good.
I just made 1 comment to think about for the future.

phil-davis:
Backport stable10 #30351

Issue #30157, opened 2018-01-16
[10.0.4] Sessions created from application passwords result in "Login failed" errors when used


SamuAlfageme:
### Context:
~~Prior to 10.0.4, there was a problem with application password and the sessions left behind after deletion (see https://github.com/owncloud/core/issues/28553). This is most likely a collateral of the new logic for the fix on https://github.com/owncloud/core/pull/28879 (lacking a migration strategy maybe)~~

The clients send 2 auth. methods (required) on the headers for **all their requests**: `Authorization: Basic` and `Cookies`. 

### Steps to reproduce
0. Setup lockout mechanism
1. Create an application password on a 10.0.4 installation
2. Use it to log in to the Desktop client 
3. Observe the server logs

### Expected behavior
Everything goes smooth

### Actual behavior
See https://github.com/owncloud/core/issues/30157#issuecomment-360444643; **each client request** results in 2 entries on the logs:

```
owncloud_1  | {"reqId":"b5PoKuAGyNkriyfFAMbG","level":2,"time":"2018-01-30T16:06:51+00:00","remoteAddr":"172.19.0.1","user":"admin","app":"core","method":"PROPFIND","url":"\/remote.php\/dav\/files\/admin\/","message":"Login failed: 'admin' (Remote IP: '172.19.0.1')"}
owncloud_1  | 172.19.0.1 - admin [30/Jan/2018:16:06:51 +0000] "PROPFIND /remote.php/dav/files/admin/ HTTP/1.1" 207 1551 "-" "Mozilla/5.0 (Macintosh) mirall/2.3.4 (build 8537)"
```

- **One successful, for the `Authorization` header.**
- **One that causes a `Login failed` in core, that can be attributed to the `Cookie`** -> the issue here; this makes password lockout mechanisms to trigger if present (see 0)

### Cause:
https://github.com/owncloud/core/pull/29525

### Related Issues:
- https://github.com/owncloud/client/issues/6258
- https://github.com/owncloud/user_ldap/issues/167

~~I'm assigning @IljaN since he's familiar with the https://github.com/owncloud/core/pull/28879 patch, feel free to change assignment.~~

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributor most likely able to help you is @PVince81.

davitol:
Using Version 2.3.0-nightly20160929 (build 3715) OSX Desktop Client

1. oC 10.0.3 --> create app password A and app password B
[...]

PR #30342, opened 2018-02-02
checkPassword should also check oc_authtokens 

IljaN:
## Description
When authenticating via app password checkPassword is used which results in a failed login because it only checks oc_users. This change modifies checkPassword so that it also checks auth tokens.

## Related Issue
#30157

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/Database.php
+++ b/lib/private/User/Database.php
@@ -50,6 +50,8 @@
 
 namespace OC\User;
 
+use OC\Authentication\Exceptions\InvalidTokenException;
+use OC\Authentication\Token\DefaultTokenProvider;
 use OC\Cache\CappedMemoryCache;
 use OCP\IUserBackend;
 use OCP\Util;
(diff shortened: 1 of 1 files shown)


lib/private/User/Database.php
```diff
@@ -201,6 +203,21 @@ public function getDisplayNames($search = '', $limit = null, $offset = null) {
 	 * returns the user id or false
 	 */
 	public function checkPassword($uid, $password) {
+		// First check oc_auth_tokens (app password)
+		/** @var DefaultTokenProvider $tokenProvider */
+		$tokenProvider = \OC::$server->query('OC\Authentication\Token\DefaultTokenProvider');
```
PVince81: I don't think this is the right location. Is there a higher level in the stack ?

Because this code is only about database passwords. It wouldn't work when checking LDAP passwords, etc.

PVince81 (changes requested):
see comment

IljaN:
Obsolete: Correct fix here: https://github.com/owncloud/core/pull/30364

PR #26021, opened 2016-09-02
Precompiled handlebars template

PVince81:
## Description

Introduce Makefile rule to build handlebars templates for every app in core, if applicable.
It compiles all templates from "apps/$APP/templates/*.handlebars" into a single JS file "apps/$APP/js/compiled-templates.js" that must be excluded from version control.

As an example, the template of the DetailsView in the files app has been ported to an external handlebars template file.
## Related Issue

Fixes https://github.com/owncloud/core/issues/12848
## Motivation and Context

See https://github.com/owncloud/core/issues/12848
## How Has This Been Tested?

Run `make clean` then `make test-js`. All tests will pass.
Do a manual test in the files UI and see that the right sidebar still works.
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
## Some concerns
- I'm not sure whether it's ok to have the template building instructions inside the main Makefile. I thought about having one Makefile per app instead, but could cause a bigger deal of redundancy.
- Not sure if the location "js/compiled-templates.js" is ok. But can't put it inside a build dir since it won't be loadable by `addScript()`
- Had to use a generic namespace name "OC.template[appName][templateName]" because we can't guess the app's JS namespace from outside (having a separate makefile per app could help)
- If one forgets to run "make build-apps", the templates are not refreshed, might need some kind of watcher...

Please let me know what you think @DeepDiver1975 @butonic @jvillafanez @VicDeo

PR #30343, opened 2018-02-02
Fix secondary s3

### LINKED ISSUE #27081: Connection using path style to a personal S3 (not amazon) fails.
### Steps to reproduce:

1.- Enable external storage.
2.- Run a container using S3 protocol. 
   I've used https://hub.docker.com/r/webhippie/minio/ with a created bucket inside.

Steps to use this container:
Having a docker daemon running do:

2.1.-  Get minio image:
```docker pull webhippie/minio```

2.2.- Create a container:
For example:
```docker run -t -i -d -p 39000:9000 webhippie/minio:latest```
This will drop and id.

Using docker logs id you can see some info of the container.

2.3.-Create a bucket:
Access the external ip:redirected_port (for example 192.168.99.100:39000) using a browser to create it. You'll need the AccessKey and SecretKey from docker logs. 

3.- Set up an external storage using S3 

My config:

screen shot 2017-02-03 at 13 56 39

### Expected behaviour
Connection works and you can upload files to it.

### Actual behaviour
Connection fails and there are no logs.

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

Current master
{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.0.0.0","versionstring":"10.0.0","edition":"","productname":"ownCloud"}

Fresh

```

```

Trying to use S3

**Are you using encryption:** 
No

**Logs**

No logs even when seeing it failing.

```

```

### Client configuration

Firefox.

DeepDiver1975:
## Description
- update to latest s3 libs
- make work with minio

## Related Issue
fixes #27081

## How Has This Been Tested?
- [x] manual
- [ ] unit tests

## Types of changes

## Checklist:


PVince81:
Would be better to have one Makefile per app as discussed before.

I have some WIP in a PR where the main Makefile calls `make` in all apps when possible: https://github.com/owncloud/core/pull/26021/files#diff-b67911656ef5d18c4ae36cb6741b7965R240

On the other hand, it would be better to already move out S3 into a separate app and put the Makefile there...

DeepDiver1975:
> On the other hand, it would be better to already move out S3 into a separate app and put the Makefile there...

this is where the GSoC project will kick in .....

DeepDiver1975:
@PVince81 should we aim for a new app right away?

PVince81:
@DeepDiver1975 yes, because in the new app you'll anyway need to tweak composer etc. So instead of doing it here for core in some way and doing it differently again in a separate app, let's do the separate app directly.

DeepDiver1975:
superseded

PR #30289, opened 2018-01-26
Fix #30279: Modifies check for CSRF tokens

pandeykartikey:
## Description
In this PR I have modified checks of requiresCSRFCheck() to only check on POST request

## Motivation and Context
It solves the issue #30279 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30358, opened 2018-02-03
[stable10] Fix #30279: Modifies check for CSRF tokens

### LINKED ISSUE #30279: Webdav: only check CSRF for POST
See https://github.com/owncloud/core/blob/v10.0.4/apps/dav/lib/Connector/Sabre/Auth.php#L187.

Some code paths can lead to checking CSRF for PROPFIND or other methods.

This should be changed to only ever check CSRF for POST requests.

This could then avoid strange errors like https://github.com/owncloud/android/issues/1964#issuecomment-360484533

@DeepDiver1975 @settermjd

phil-davis:
Backport #30289


PVince81 (approved):
:+1:

PR #30365, opened 2018-02-05
Bugfix/owncloud/oauth2#103

### LINKED ISSUE #103: OAuth2 token revokation does not expire sessions
### Steps to reproduce
1. Create a new OAuth2 session:

screen shot 2017-12-01 at 13 16 14

2. Remove the authorized OAuth2 token in webUI `Security` section 

### Expected behaviour
Sessions for this client/user agent are removed/invalidated as well

### Current behaviour
Sessions keep alive, all request are reponsed with 200

screen shot 2017-12-01 at 13 17 02

Set up:
- ownCloud 10.0.4 RC1 (daily) Build:2017-12-01T04:16:36+00:00 d4eab96823baec0061bc221e35b49dce968cbed9

- OAuth 0.2.1

DeepDiver1975:
## Related Issue
fixes owncloud/oauth2#103

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/BasicAuthModule.php
+++ b/lib/private/User/BasicAuthModule.php
@@ -41,21 +41,29 @@ class BasicAuthModule implements IAuthModule {
 	 * @inheritdoc
 	 */
 	public function auth(IRequest $request) {
-		if (empty($request->server['PHP_AUTH_USER']) || empty($request->server['PHP_AUTH_PW'])) {
+		if (!isset($request->server['PHP_AUTH_USER'], $request->server['PHP_AUTH_PW'])) {
+			return null;
+		}
+		$authUser = $request->server['PHP_AUTH_USER'];
+		$authPass = $request->server['PHP_AUTH_PW'];
+		if ($authUser === '' || $authPass === '') {
 			return null;
 		}
 
 		// check uid and password
-		$user = $this->manager->checkPassword($request->server['PHP_AUTH_USER'], $request->server['PHP_AUTH_PW']);
+		$user = $this->manager->checkPassword($authUser, $authPass);
 		if ($user instanceof IUser) {
 			return $user;
 		}
 		// check email and password
-		$users = $this->manager->getByEmail($request->server['PHP_AUTH_USER']);
-		if (count($users) !== 1) {
-			return null;
+		$users = $this->manager->getByEmail($authUser);
+		if (count($users) === 1) {
+			$user = $this->manager->checkPassword($users[0]->getUID(), $authPass);
 		}
-		return $this->manager->checkPassword($users[0]->getUID(), $request->server['PHP_AUTH_PW']);
+		if ($user instanceof IUser) {
+			return $user;
+		}
+		throw new \Exception('Invalid credentials');
 	}
 
 	/**
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -283,15 +283,15 @@ class Session implements IUserSession, Emitter {
 	public function getLoginName() {
 		if ($this->activeUser) {
 			return $this->session->get('loginname');
-		} else {
-			$uid = $this->session->get('user_id');
-			if ($uid) {
-				$this->activeUser = $this->manager->get($uid);
-				return $this->session->get('loginname');
-			} else {
-				return null;
-			}
 		}
+
+		$uid = $this->session->get('user_id');
+		if ($uid) {
+			$this->activeUser = $this->manager->get($uid);
+			return $this->session->get('loginname');
+		}
+
+		return null;
 	}
 
 	/**
@@ -320,6 +320,7 @@ class Session implements IUserSession, Emitter {
 	 * @param string $user
 	 * @param string $password
 	 * @param IRequest $request
+	 * @throws \InvalidArgumentException
 	 * @throws LoginException
 	 * @throws PasswordLoginForbiddenException
 	 * @return boolean
@@ -988,21 +989,33 @@ class Session implements IUserSession, Emitter {
 	}
 
 	public function verifyAuthHeaders($request) {
-		foreach ($this->getAuthModules(true) as $module) {
-			$user = $module->auth($request);
-			if ($user !== null) {
-				if ($this->isLoggedIn() && $this->getUser()->getUID() !== $user->getUID()) {
-					// the session is bad -> kill it
-					$this->logout();
-					return false;
+		$shallLogout = false;
+		try {
+			$lastUser = null;
+			foreach ($this->getAuthModules(true) as $module) {
+				$user = $module->auth($request);
+				if ($user !== null) {
+					if ($this->isLoggedIn() && $this->getUser()->getUID() !== $user->getUID()) {
+						$shallLogout = true;
+						break;
+					}
+					if ($lastUser !== null && $user->getUID() !== $lastUser->getUID()) {
+						$shallLogout = true;
+						break;
+					}
+					$lastUser = $user;
 				}
-				return true;
 			}
-		}
 
-		// the session is bad -> kill it
-		$this->logout();
-		return false;
+		} catch (Exception $ex) {
+			$shallLogout = true;
+		}
+		if ($shallLogout) {
+			// the session is bad -> kill it
+			$this->logout();
+			return false;
+		}
+		return true;
 	}
 
 	/**
@@ -1010,7 +1023,7 @@ class Session implements IUserSession, Emitter {
 	 * @return \Generator | IAuthModule[]
 	 * @throws Exception
 	 */
-	private function getAuthModules($includeBuiltIn) {
+	protected function getAuthModules($includeBuiltIn) {
 		if ($includeBuiltIn) {
 			yield new TokenAuthModule($this->session, $this->tokenProvider, $this->manager);
 		}
--- a/lib/private/User/TokenAuthModule.php
+++ b/lib/private/User/TokenAuthModule.php
@@ -26,9 +26,11 @@ namespace OC\User;
 use OC\Authentication\Exceptions\InvalidTokenException;
 use OC\Authentication\Exceptions\PasswordlessTokenException;
 use OC\Authentication\Token\IProvider;
+use OC\Authentication\Token\IToken;
 use OCP\Authentication\IAuthModule;
 use OCP\IRequest;
 use OCP\ISession;
+use OCP\IUser;
 use OCP\IUserManager;
 use OCP\Session\Exceptions\SessionNotAvailableException;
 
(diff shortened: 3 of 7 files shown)


lib/private/User/BasicAuthModule.php
```diff
@@ -45,17 +47,12 @@ public function auth(IRequest $request) {
 			return null;
 		}
 
-		// check uid and password
-		$user = $this->manager->checkPassword($request->server['PHP_AUTH_USER'], $request->server['PHP_AUTH_PW']);
-		if ($user instanceof IUser) {
-			return $user;
+		// reuse logClientIn because this method
```
PVince81: because this method... yes ?

lib/private/User/BasicAuthModule.php
```diff
@@ -66,6 +63,10 @@ public function getUserPassword(IRequest $request) {
 			return '';
 		}
 
+		if ($this->session->getSession()->exists('app_password')) {
+			// TODO: use proper password
+			throw new \Exception('Not implemented yet');
```
PVince81: is this ever triggered ?

lib/private/User/Session.php
```diff
@@ -988,21 +988,30 @@ public function updateSessionTokenPassword($password) {
...
-					// the session is bad -> kill it
-					$this->logout();
-					return false;
+		$lastUser = null;
+		try {
+			foreach ($this->getAuthModules(true) as $module) {
+				$user = $module->auth($request);
+				if ($user !== null) {
+					if ($this->isLoggedIn() && $this->getUser()->getUID() !== $user->getUID()) {
+						throw new Exception();
```
PVince81: would be nice to have a message, especially if we don't have a proper type.

no more auto-logout of bad sessions ?

lib/private/User/BasicAuthModule.php
DeepDiver1975: Not at the moment - this is a safety exception which will punch us in the face as soon as we use it accidentally.

Actually the whole method should not yet be called ... we could just throw the exception right away

lib/private/User/Session.php
DeepDiver1975: these exceptions are basically just to jump out of the loop and do the logout in the catch block - I'm not to happy with this .....

PVince81 (approved):
ok then 👍

PVince81:
looks good, yes

PVince81:
@SamuAlfageme can you retest auth issue with this ?

PVince81:
CI fails ? :-/

DeepDiver1975:
> CI fails ? :-/

drone has network issues

DeepDiver1975:
jenkins has failing integration tests

PVince81:
the integration tests seem completely unrelated. trashbin ?! 🙈

PVince81:
oh wait... trashbin result says `Failed asserting that 401 matches expected '200'.`.

Either we broke something or the test was passing by chance before

DeepDiver1975:
I better have a look ....

DeepDiver1975:
```
10:27:08     And logging in using web as "user1"                                                                   # FeatureContext::loggingInUsingWebAs()
10:27:08 [Thu Feb  8 10:27:08 2018] ::1:53702 [200]: /index.php/apps/files/
10:27:08 [Thu Feb  8 10:27:08 2018] ::1:53706 [401]: /index.php/apps/files_trashbin/ajax/list.php?dir=%2F
10:27:08     When as "user1" the file with original path "/renamed_shared/shared_file.txt" is restored             # FeatureContext::elementInTrashIsRestored()
10:27:08       Failed asserting that 401 matches expected '200'.
10:27:09     Then as "user1" the file with original path "/renamed_shared/shared_file.txt" does not exist in trash # FeatureContext::elementIsNotInTrashCheckingOriginalPath()
10:27:09     And user "user1" should see following elements                                                        # FeatureContext::checkElementList()
10:27:09       | /renamed_shared/                |
10:27:09       | /renamed_shared/shared_file.txt |

**Comment by DeepDiver1975** [comment]
The requests as executed by the integration tests are not reusing the session cookies - which might have a side effect ...

**Comment by PVince81** [comment]
ajax 😱 

but it should use the session from the web login

**Comment by PVince81** [comment]
need to check how the tests are actually doing this ajax call, maybe it's just using basic auth for the call instead of cookies...

File: lib/private/User/Session.php [hunk 4]
```diff
@@ -302,8 +302,10 @@ public function getLoginName() {
 	 * @return boolean|null
 	 * @throws LoginException
 	 */
-	public function login($uid, $password) {
-		$this->session->regenerateId();
+	public function login($uid, $password, $recreateSession = true) {
+		if ($recreateSession) {
```
PVince81: what if the session contained a different user ? will your protection apply itself in this code path ?

DeepDiver1975: this is taken care of in different code locations - verifyheaders and verifytoken

PVince81:
Weird trashbin failures now:
```
When as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" is restored:

File previously located at /local_storage/tmp/textfile0.txt is still in the trashbin
Failed asserting that true is false.
```

no more related to the 401 error

lib/private/User/BasicAuthModule.php
```diff
@@ -52,10 +52,13 @@ public function auth(IRequest $request) {
...
-		if (count($users) !== 1) {
-			return null;
+		if (count($users) === 1) {
+			$user = $this->manager->checkPassword($users[0]->getUID(), $request->server['PHP_AUTH_PW']);
+		}
+		if ($user instanceof IUser) {
+			return $user;
 		}
-		return $this->manager->checkPassword($users[0]->getUID(), $request->server['PHP_AUTH_PW']);
+		throw new \Exception('Invalid credentials');
```
PVince81: return false ? Throwing could cause other side effects...

PVince81: in TokenAuthModule we return null, let's stay consistent

lib/private/User/TokenAuthModule.php
```diff
@@ -91,4 +84,46 @@ public function auth(IRequest $request) {
 	public function getUserPassword(IRequest $request) {
 		return $this->password;
 	}
+
+	/**
+	 * @param IRequest $request
+	 * @return null|IToken
+	 */
+	private function getTokenForAppPassword(IRequest $request, &$token) {
+		if (empty($request->server['PHP_AUTH_USER']) || empty($request->server['PHP_AUTH_PW'])) {
```
PVince81: not afraid of "0" users ?

PVince81 (commented):
see comments

lib/private/User/TokenAuthModule.php
DeepDiver1975: @PVince81 as requested: fuck php

lib/private/User/BasicAuthModule.php
DeepDiver1975: null is returned if there is no authentication header to be evaluated aka handled by this module. 
exception is throws if credentials are wrong
user object is returned if creds are valid

PVince81: okay

DeepDiver1975: But you are right - this logic seems to be missing in the token auch module ....

SamuAlfageme:
@DeepDiver1975 initial tests suggest the problem no longer reproduces with this + owncloud/oauth2#112; great news! 🎉 

1. Reduced drastically `AccessToken`'s `resetExpires()` to bump `expires` every 20s
2. Authorized the Desktop client 
3. Close the client during `access_token` validity period and wait 20s
4. Replay any successful request the client made previously 
5. Observe 401 "Unauthorized" reply on the replayed request 

Would a traversal unit test like this one make sense to be included in the OAuth2 test pool?

Will focus now on trying to reproduce https://github.com/owncloud/oauth2/issues/103 on the mobile apps; with and without these patches.

SamuAlfageme:
@DeepDiver1975 can you check if the scope of this patch can be extended to the old WebDAV endpoint? 

After some more tests, I believe the original cause of https://github.com/owncloud/oauth2/issues/103 is the logic for the old endpoint. Take these 2 `PROPFIND`s using OAuth2 as the auth8n mechanism:

<p align="center">
  
</p>

- First 2 requests (url ending on `:4443`) belong to a 10.0.6 "unpatched" server (1st. is iOS: old endpoint, 2nd. Desktop, new)
- The following 2 are 10.0.6 + https://github.com/owncloud/core/pull/30365 + https://github.com/owncloud/oauth2/issues/112

Replaying the iOS request any number of times results on 2XX, while desktop gets the correct 401.

SamuAlfageme:
Hm. Forcing the use of the old endpoint on the Desktop client (setting env. `OWNCLOUD_CHUNKING_NG=0`) still results on 401 & token refresh. 

Appart from the endpoints there are a few apparent differences between iOS and Desktop `PROPFINDS` 😕 

### Desktop:

```bash
curl \
    -H 'Depth:0' \
    -H 'Authorization:Bearer VBvulBvRbhOpaA5BLkaLl7H7AdeatO5kwGyDNqfwYf2XRnsJTz0sEagLwgSuT3U3' \
    -H 'User-Agent:Mozilla/5.0 (Macintosh) mirall/2.4.0 (build 8836)' \
    -H 'Accept:*/*' \
    -H 'Content-Type:text/xml; charset=utf-8' \
    -H 'X-Request-ID:83fd38fa-395b-429e-a1be-80aad5fc9c32' \
    -H 'Cookie:oc_sessionPassphrase=fnebJi6XYVoOIsvNDYO9KMec7ZI8qYbD3m0D0kYi5UJhoXh%2FviUkj7hqrvuPkXpaFgzSW0UqemP4JIOjFUMjmR2tWh%2Bt1DjRkr8uXnypprsGZvRoy7EGWdSwGemssovV; oc3kizzp7f7d=plgcob981he5sjek79r7avjqp1' \
    -H 'Content-Length:147' \
    -H 'Connection:Keep-Alive' \
    -H 'Accept-Encoding:gzip, deflate' \
    -H 'Accept-Language:en-US,*' \
    -H 'Host:alfageme.solidgear.prv:4443' \
    -X PROPFIND 'https://alfageme.solidgear.prv:4443/remote.php/webdav/' \
    --data-binary '<?xml version="1.0" ?>
<d:propfind xmlns:d="DAV:">
  <d:prop>
    <d:quota-available-bytes />
    <d:quota-used-bytes />
  </d:prop>
</d:propfind>
'
```

### iOS:

```bash
curl \
    -H 'Host:alfageme.solidgear.prv:4443' \
    -H 'Content-Type:application/xml' \
    -H 'Depth:1' \
    -H 'Accept:*/*' \
    -H 'Accept-Encoding:gzip, deflate' \
    -H 'Connection:keep-alive' \
    -H 'Cookie:oc3kizzp7f7d=6ds6rnc8mcl5jg31dkmdnpqli6; oc_sessionPassphrase=9sWpKyFrO8B59uXHo%2F4BRmHB3X2q6qJDtpS17D2Zx7BcFPp0s4yPdWiZBPqmzkmWvo4yoBW9XJ7%2BwzgtfCVWfmy%2F0TzJwWUVbROzy0LDGHFPM9543cKTkWnf6%2FqoFcB1' \
    -H 'User-Agent:Mozilla/5.0 (iOS) ownCloud-iOS/3.7.0' \
    -H 'Accept-Language:en-US;q=1, es-US;q=0.9, zh-Hant-US;q=0.8, en-GB;q=0.7, en;q=0.6' \
    -H 'Authorization:Bearer fVbLupKEkYi4Yr6tzALsAZlMZIjMDEkpjZXr5kcwYR6HMKvjDy0RQPoXS4dJ7QSl' \
    -H 'Content-Length:383' \
    -X PROPFIND 'https://alfageme.solidgear.prv:4443/remote.php/webdav/' \
    --data-binary '<?xml version="1.0" encoding="UTF-8"?><D:propfind xmlns:D="DAV:"><D:prop><D:resourcetype/><D:getlastmodified/><size xmlns="http://owncloud.org/ns"/><D:creationdate/><id xmlns="http://owncloud.org/ns"/><D:getcontentlength/><D:displayname/><D:quota-available-bytes/><D:getetag/><permissions xmlns="http://owncloud.org/ns"/><D:quota-used-bytes/><D:getcontenttype/></D:prop></D:propfind>'
```

### Diffs:

- Order of the `Authorization` and `Cookie` headers: shouldn't matter as WebDAV inherits HTTP's ["_The order in which header fields with differing field names are received is not significant_"](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)
- Order of the `Cookie`s itself: `oc_sessionPassphrase` first/last
- `Depth`: 1/0
- Content of the request itself

Will play with these differences and report back when I find the one that messes with the auth. middleware.

DeepDiver1975:
I honestly question if oauth works at all on the old endpoint .......

PVince81:
If we didn't break more stuff that wasn't broken before I'd suggest moving forward with the PR and get it released as it adresses the more critical issue. Then investigate this detail separately ?

@SamuAlfageme @DeepDiver1975

SamuAlfageme:
As discussed with @DeepDiver1975 and @owncloud/ios-developers, in the end, https://github.com/owncloud/oauth2/issues/103 was all about the iOS app circumventing the webview's sandbox (I haven't had the chance to check with Android's implementation, tho). This means the client was reusing cookies obtained during the authentication steps delegated to the browser (which goes against the end goal of the OAuth2 mechanism: isolation from this); those were bounded to the browser's activity; just like if they were obtained on safari or any other mobile browser, and valid as long as them (not dependent on the `access_token` at all).

After all the tests performed, all the basic access/refresh/revoke scenarios look sane enough. I'll go for the merge here. 👍

PR #30366, opened 2018-02-05
Return status as string in getOCSResponseStatusCode

phil-davis:
## Description
Actually return the status code as a string in ``getOCSResponseStatusCode``

## Motivation and Context
``getOCSResponseStatusCode`` says it returns a string. But actually it returns a simple XML object.
That can cause problems with the check in ``theOCSStatusCodeShouldBe``. If you pass it a string as the expected value ``"100"`` all is well, and ``assertEquals`` understands to cast the simple XML object to string and compare to ``"100"``. But if you pass it just ``100`` (looks like an ``int``) then ``assertEquals`` does not cope with comparing ``int`` with ``simpleXMLObject``.

This is a latent issue that I noticed while doing other integration test refactoring. Rather than burying this little gem in another huge refactoring PR, I have separated it.

## How Has This Been Tested?
Run some integration tests locally.

## Types of changes
- [x] Test refactoring

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/BasicStructure.php
+++ b/tests/integration/features/bootstrap/BasicStructure.php
@@ -142,8 +142,8 @@ trait BasicStructure {
 	 * @param ResponseInterface $response
 	 * @return string
 	 */
-	public function getOCSResponse($response) {
-		return $response->xml()->meta[0]->statuscode;
+	public function getOCSResponseStatusCode($response) {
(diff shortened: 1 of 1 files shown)


individual-it (approved):
👍 if CI passes

phil-davis:
Surely CI will pass - there are never problems with CI ;)

PR #30367, opened 2018-02-05
Fix for when account not found during sync

tomneedham:
Fixes https://jenkins.owncloud.org/job/upgrade-testing/CONFIG=8.2.11%3Bdaily-master-qa,DB=sqlite/285/console

### DIFF
--- a/lib/private/User/SyncService.php
+++ b/lib/private/User/SyncService.php
@@ -106,21 +106,24 @@ class SyncService {
 			// update existing and insert new users
 			foreach ($users as $uid) {
 				try {
-					$a = $this->mapper->getByUid($uid);
-					if ($a->getBackend() !== $backendClass) {
+					$existingAccount = $this->mapper->getByUid($uid);
+					if ($existingAccount->getBackend() !== $backendClass) {
 						$this->logger->warning(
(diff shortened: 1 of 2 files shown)


lib/private/User/SyncService.php
```diff
@@ -117,7 +117,7 @@ public function run(UserInterface $backend, \Closure $callback) {
 					$this->syncAccount($a, $backend);
 					$this->mapper->update($a);
 				} catch(DoesNotExistException $ex) {
-					$this->createNewAccount($backendClass, $uid);
+					$a = $this->createNewAccount($backendClass, $uid);
```
DeepDiver1975: since this was not detected by the regular CI process - I assume this is missing unit tests

DeepDiver1975 (changes requested):

lib/private/User/SyncService.php
phil-davis: The regression looks like it happened here:
https://github.com/owncloud/core/pull/29669/files#diff-37958be4e01e8baeb127a59409e1a593L114

tomneedham: Yep - looking at tests for this

tomneedham:
Added a test to run through the sync service at least once.

individual-it:
restarted drone, now passed

Issue #15646, opened 2015-04-15
Public API to work with trashbin


PVince81:
OCS or WebDAV, whichever is better suited.

Need to be able to:
1) List files in trashbin
2) Delete files permanently
3) Restore files

PVince81:
@karlitschek can we schedule this ? Required for automated tests of versions.
If we had the tests it would have detected regressions like this one https://github.com/owncloud/core/issues/15009#issuecomment-100936118

karlitschek:
I agree that it would be a very good idea. Time is tight. Let´s see what we can do :-/

DeepDiver1975:
Still voting for webdav - we are talking about files ...

PVince81:
True, this requires discussing the right API format first.
I just don't want this task to get lost, so setting to backlog for now.

PVince81:
While we could design an API for the current trashbin access, it might not be future proof if we decide to move forward to the "per storage trashbin" concept: https://github.com/owncloud/core/issues/12274#issuecomment-246617302
If we do, the endpoint in question might require additional path sections to specify the storage from which to read the trashbin.

PVince81:
If we stay with the current trashbin, I'd suggest the following format:

- base endpoint: https://host/owncloud/remote.php/dav/files_trashbin/$user/
- trashed files listed by file id: https://host/owncloud/remote.php/dav/files_trashbin/$user/$fileId
- a PROPFIND on the file id element will return the file name as "oc:name" property. This is needed because there might be multiple files with the same name but with different deletion timestamp
- new property "oc:deleted_time" for the deletion timestamp
- deleted folders also appear with "$fileId", so a subfile would look like: https://host/owncloud/remote.php/dav/files_trashbin/$user/$folderFileId/$subFileId

@DeepDiver1975 @davivel @nasli @dragotin @guruz

PVince81:
- another optional property: "oc:original-path" that returns the path to which the file would be restored

PVince81:
- PROPFIND **could** maybe also return more metadata like tags/shares if it makes sense
- restore a file by calling MOVE https://host/owncloud/remote.php/dav/files_trashbin/$user/$fileId and specifying the value from "oc:original-path" as "Destination" header. Need conflict handling with the proper headers and handling of 409 response in case the file exists (like regular MOVE)
- delete a file/folder from trash by calling DELETE https://host/owncloud/remote.php/dav/files_trashbin/$user/$fileId
[...]

PR #30383, opened 2018-02-06
Add Webdav-Location header in private link redirect

PVince81:
## Description
Added Webdav-Location header in response when opening private links.
Added 404 status code for error page when requesting inaccessible/invalid private links.

## Related Issue
Similar to https://github.com/owncloud/core/issues/28615

## Motivation and Context
Useful for clients to resolve private links to Webdav URLs without having to extract the file id from the URL. Please note that there are at least two permalink formats, one with "/f/$fileId" and another one is the link when copied directly from the browser which has "fileid=$fileId". This removes the need for clients to know about this format and simply query whatever link they got and get it resolved.

## How Has This Been Tested?
Manual test with curl (with `-I` to only request headers with the "HEAD" method) and unit tests.

## Types of changes

## Checklist:

cc @michaelstingl @nasli

### DIFF
--- a/apps/files/lib/Controller/ViewController.php
+++ b/apps/files/lib/Controller/ViewController.php
@@ -41,6 +41,7 @@ use OCP\IRequest;
 use OCP\IURLGenerator;
 use OCP\IUserSession;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use OCP\AppFramework\Http;
 
 /**
  * Class ViewController
@@ -282,12 +283,14 @@ class ViewController extends Controller {
 		$files = $baseFolder->getById($fileId);
 		$params = [];
 
+		$isFilesView = true;
 		if (empty($files) && $this->appManager->isEnabledForUser('files_trashbin')) {
 			// Access files_trashbin if it exists
 			if ( $this->rootFolder->nodeExists($uid . '/files_trashbin/files/')) {
 				$baseFolder = $this->rootFolder->get($uid . '/files_trashbin/files/');
 				$files = $baseFolder->getById($fileId);
 				$params['view'] = 'trashbin';
+				$isFilesView = false;
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


PVince81:
If a private link points to a file that is in trash, we don't have a matching Webdav URL yet until https://github.com/owncloud/core/issues/15646 is implemented. Returning a 404 error would be wrong here.

DeepDiver1975 (approved):

phil-davis:
Backport stable10 #30387

apps/files/lib/Controller/ViewController.php
```diff
@@ -302,14 +305,23 @@ public function showFile($fileId) {
 				// and scroll to the entry
 				$params['scrollto'] = $file->getName();
 			}
-			return new RedirectResponse($this->urlGenerator->linkToRoute('files.view.index', $params));
+			$response = new RedirectResponse($this->urlGenerator->linkToRoute('files.view.index', $params));
+			if ($isFilesView) {
+				$webdavUrl = $this->urlGenerator->linkTo('', 'remote.php') . '/dav/files/' . $uid . '/';
```
SamuAlfageme: @PVince81 shouldn't this be endpoint-aware instead of hardcoded? i.e. using the same endpoint as the request that originated this `$response`? My _endpoint-juggling_ context might be playing mind tricks on me, though.

PVince81: endpoint-aware in what way ?

The ViewController is not part of Webdav but it's the web page from "index.php/apps/files" so what decision do you think should be made based on this ?

SamuAlfageme: > it's the web page from "index.php/apps/files"

You're right. I was thinking on a different (non-existent) workflow. 

What I really meant was If there won't be a problem to use a hardcoded WebDAV path regardless of what the client that is accessing the private link normally uses/understands. 

e.g. since this request came from the iOS team and they talk with the old endpoint, the new might offer different properties, no? Or if, like in the past; we detect some problems with the endpoint on the webUI that forces us to temporarily switch back to old, etc.

apps/files/lib/Controller/ViewController.php
```diff
@@ -302,14 +305,23 @@ public function showFile($fileId) {
...
+				$webdavUrl .= ltrim($baseFolder->getRelativePath($file->getPath()), '/');
```
SamuAlfageme: @PVince81 take into account that the file name at this point is not URL-encoded: a file called `level7$éâ .txt` should point to `remote.php/dav/files/user/Documents/level7$%C3%A9%C3%A2%20.txt` but header is `Webdav-Location: /remote.php/dav/files/user/Documents/level7$Ã©Ã¢ .txt`. 

Clients will query a non-existent file then. We should add a test for it as well.

PVince81: right, I'll then add some encoding in a subsequent PR

PVince81: here we go, please have a look: https://github.com/owncloud/core/pull/30503

PR #30386, opened 2018-02-06
decouple token check from user sync

butonic:
Currently we will be syncing users every five min: https://github.com/owncloud/core/blob/05a82b590300a8e08808895fae77dd1e078a5d15/lib/private/User/Session.php#L718-L742

This PR decouples the sync logic from the Managers checkpassword (which is supposed to have no side effects if you look at the old docs: https://github.com/owncloud/core/blob/815e83a6cb23720756044921e78ca563ca629585/lib/private/legacy/user.php#L433) and moves it to the Sessions loginWithPassword().

### DIFF
--- a/core/Controller/TokenController.php
+++ b/core/Controller/TokenController.php
@@ -30,6 +30,7 @@ use OC\User\Manager as UserManager;
 use OCA\User_LDAP\User\Manager;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\JSONResponse;
+use OCP\Authentication\InvalidCredentialsException;
 use OCP\IRequest;
 use OCP\Security\ISecureRandom;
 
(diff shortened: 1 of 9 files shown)


butonic:
- [ ] submit PR for encryption app

DeepDiver1975:
- conflicting
-  please hold on - I'm reimplementing the basic auth module as we write ....

DeepDiver1975:
@butonic please have a look at https://github.com/owncloud/core/pull/30365

PVince81:
obsolete or needs rework on top of the next auth impl ?

PVince81:
@butonic @DeepDiver1975 need input to continue this or close ?

PVince81:
@butonic if we do want to continue, please provide more context, reason for the fix and next steps so someone else can take over

butonic:
Currently, we are trying to verify a session every 5 min.
The checkPassword code currently
1. checks the password https://github.com/owncloud/core/pull/30386/files#diff-2ad387ea4134715ffda3bb0f12140f91L228 and then also
2. syncs the user metadata from ldap in https://github.com/owncloud/core/pull/30386/files#diff-2ad387ea4134715ffda3bb0f12140f91L230

Which is another request against the ldap server. So we are not only syncing metadata on every login but also every 5min (because the desktop clients will be polling with the session)

All of this is only relevant if ldap credentials are used directly. Probably a non issue.

Closing until someone reports heavy ldap load.

PR #30387, opened 2018-02-06
[stable10] Add Webdav-Location header in private link redirect

DeepDiver1975:
Useful for clients to resolve private links to Webdav URLs without
having to extract the file id from the URL

backport of #30383


PVince81 (approved):
:+1:

Issue #23151, opened 2016-03-11
Deleted files sometimes aren't deleted properly due to timestamp conflict


mirekys:
### Steps to reproduce
1. Have some files files possibly with a frequently occuring name/sub-path pattern in your storage:
   - /dir1/file.txt
   - /dir1/dir2/file.txt
   - /dir3/dir4/file.txt
     ...
     // this is just a hypothesis from observation, didn't manage to reproduce it by myself yet //
2. Delete /dir1/dir2/file.txt (will be moved to trash correctly)
3. Delete /dir1/file.txt (no entry created in oc_files_trash table, still presented as regular file, download fails)
### Expected behaviour

Deleted files may not be presented by the server on its original path as regular files
and should be accessible through files_trashbin.
### Actual behaviour

Deleted file is unlinked on underlying storage (on path 'data/userid/files/dir1/file.txt'), but is still presented together with other regular files on its original path on web, and in response to PROPFIND. Download fails with 404 (web) or 503 - Could not open file (sync client). Synchronization stops on that error.
### Server configuration

**Operating system**:
RHEL 6.4

**Web server:**
Apache/2.4.12

**Database:**
PostgreSQL 9.2

**PHP version:**
5.6.5

**ownCloud version:**
8.2.2

**Updated from an older ownCloud or fresh install:**
from 8.1.5

**Where did you install ownCloud from:**
https://owncloud.org/changelog/

**List of activated apps:**
- activity: 2.1.3
- calendar: 0.8.2
- contacts: 0.5.0.0
- documents: 0.11.1
- files: 1.2.0
- files_pdfviewer: 0.7
- files_sharing: 0.7.0
- files_texteditor: 2.0
- files_trashbin: 0.7.0
- files_versions: 1.1.0
- firstrunwizard: 1.1
- notifications: 0.1.0

**The content of config/config.php:**
  // unrelated items omitted //
  'dbtype' => 'pgsql',
  'version' => '8.2.2.2',
  'installed' => true,
  'forcessl' => true,
  'maintenance' => false,
  'enable_previews' => false,
  'singleuser' => false,
  'share_folder' => '/Shared',
  'trashbin_retention_obligation' => 'auto',
  'memcache.local' => '\OC\Memcache\APCu',
  'memcache.locking' => '\OC\Memcache\Redis',
  'allow_user_to_change_display_name' => true,
  'asset-pipeline.enabled' => true,
  'session_keepalive' => true,

[...]

PR #28111, opened 2017-06-09
Adding more trashbin scenarios

SergioBertolinSG:
Something is failing in the second scenario. The listed elements of the trashbin are less than expected and the file the case is looking for is not found.

PR #30389, opened 2018-02-06
Sort out skipped trashbin integration tests

phil-davis:
## Description
1) tag the existing skipped tests with issue-23151
2) add a new step definition that will wait and delete, to work around the timing issue
3) use that step definition in new scenarios that demonstrate the good working behaviour

## Related Issue
#23151 

## Motivation and Context
I noticed 2 trashbin tests that were skipped, while refactoring integration tests. 

I tried the tests, and they pass locally for me, they look decent and I tried combinations of what they do to make sure they would really fail if things are not right. But they fail when run in CI. It turns out they fail when run quickly - if multiple files with the same name are deleted in the same second then there is a naming conflict in the trashbin.

The automated tests can have 1 scenario that work around the timing bug/issue and tests/demonstrates the other correct behaviour of putting files with the same name from different folders into the trashbin. Then the existing scenario with the timing bug/issue can remain skipped, but also be tagged with the issue number so people can easily follow it up in future.

## How Has This Been Tested?
Run the skipped tests locally - they pass
Unskip the tests to run them in CI - they fail there, so we know they are "good" example failing cases

## Types of changes

## Checklist:

### DIFF
--- a/tests/integration/features/bootstrap/WebDav.php
+++ b/tests/integration/features/bootstrap/WebDav.php
@@ -965,6 +965,25 @@ trait WebDav {
 	}
 
 	/**
+	 * Wait for 1 second then delete a file/folder to avoid creating trashbin
+	 * entries with the same timestamp. Only use this step to avoid the problem
+	 * in core issue 23151 when wanting to demonstrate other correct behavior
+	 *
+	 * @When /^user "([^"]*)" waits and deletes (file|folder) "([^"]*)" using the API$/
+	 * @Given /^user "([^"]*)" has waited and deleted (file|folder) "([^"]*)"$/
+	 * @param string $user
+	 * @param string $type unused
+	 * @param string $file
+	 */
+	public function userWaitsAndDeletesFile($user, $type, $file) {
+		// prevent creating two files in the trashbin with the same timestamp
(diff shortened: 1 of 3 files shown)


phil-davis:
These were skipped in https://github.com/owncloud/core/pull/28111
But (assuming CI passes here) they work now.

SergioBertolinSG (dismissed):

phil-davis:
Hmmm - those tests fail on Jenkins, but pass for me locally. So there is something about the Jenkins environment/timing and the details of this test that needs looking into.

PVince81:
@phil-davis please note that deleted files get a file name like "test.txt.d1234567" where the number is a timestamp. Maybe if it deletes too fast the timestamp is exactly the same so we bump into this issue: https://github.com/owncloud/core/issues/23151

phil-davis:
Thanks for the hint - I will get back to this soon...

phil-davis:
@SergioBertolinSG @individual-it @PVince81 I have updated the code and squashed.
This should now keep the failing scenarios, better tagging them against the relevant issue, and provide new scenarios that at least test/demonstrate the good behaviour that is there.

Please review again.

individual-it (approved):

phil-davis:
Backport stable10 #23151

PR #30399, opened 2018-02-07
[master] fix drone depth

patrickjahns:
## Context

If we only fetch with a shallow history of 1 - drone will fail, if another commit is pushed on the branch, before the job started running

I propose to have a depth of 50 - that should be a good trade-off between amount of data to fetch and allow for even larger commits to be merged 

Example:
https://drone.owncloud.com/owncloud/core/3561/203

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -7,7 +7,7 @@ branches: [master, stable10, stable9.1, stable9]
 clone:
   git:
     image: plugins/git
-    depth: 1
+    depth: 50
 
 pipeline:
   restore:


.drone.yml
```diff
@@ -7,7 +7,7 @@ branches: [master, stable10, stable9.1, stable9]
 clone:
   git:
     image: plugins/git
-    depth: 1
+    depth: 50
```
DeepDiver1975: d35ad6eaf09d47cf8b9fc28ba6dfee46d25a752a9253b3b2c4609e4b1c37cd1d

patrickjahns: Meaning?

DeepDiver1975: Just kidding - I did the same at least once - but I don't find the change anymore

DeepDiver1975 (approved):

PR #30400, opened 2018-02-07
Update config.sample.php

cdamken:
Removing the old apps store from config.sample.php

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -711,17 +711,6 @@ $CONFIG = array(
 	'https://itunes.apple.com/us/app/owncloud/id543672169?mt=8',
 
 /**
- * Apps
- *
- * Options for the Apps folder, Apps store, and App code checker.
- */
-
-/**
- * The URL of the appstore to use.
- */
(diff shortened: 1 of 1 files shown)


config/config.sample.php
```diff
@@ -710,34 +710,6 @@
...
-
-/**
- * Use the ``apps_paths`` parameter to set the location of the Apps directory,
- * which should be scanned for available apps, and where user-specific apps
- * should be installed from the Apps store. The ``path`` defines the absolute
- * file system path to the app folder. The key ``url`` defines the HTTP Web path
- * to that folder, starting from the ownCloud webroot. The key ``writable``
- * indicates if a Web server can write files to that folder.
- */
- 'apps_paths' =>
```
DeepDiver1975: why this one? We better adjust the documentation and not name the app store - but the config option itself is still falid

DeepDiver1975 (changes requested):

config/config.sample.php
cdamken: ok, then we can just remove the other part

DeepDiver1975 (approved):

PR #30429, opened 2018-02-09
[stable10] Testing /remote.php/dav/meta/$fileId/v

phil-davis:
Backport #29637

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "behat/mink-selenium2-driver": "dev-master",
         "sensiolabs/behat-page-object-extension": "^2.0",
         "behat/behat": "^3.0",
-        "jarnaiz/behat-junit-formatter": "^1.3"
+        "jarnaiz/behat-junit-formatter": "^1.3",
+        "rdx/behat-variables": "^1.2"
     },
     "require": {
         "php": ">=5.4.1",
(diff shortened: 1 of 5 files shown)


phil-davis:
Note: this "achieves" backporting - getting the ``composer/json`` change back to ``stable10`` (so as to make ``stable10`` "the same as" ``master``). 

The original changes in ``WebDav.php`` added ``theVersionFolderOfFileIdContainsElements()`` and ``userRestoresVersionIndexOfFile()`` - but those need ``listVersionFolder())`` which is in ``master`` but not yet in ``stable10``. So when resolving conflicts, I kept all the new methods/steps from ``master``. That ends up making the various new step definitions available to ``stable10`` - even though they may or may not be relevant to application code currently in ``stable10``.

``dav-versions.feature`` does not exist yet in ``stable10`` so none of the actual gherkin feature/scenarios get backported. That would have to happen with a backport of the original PR that created ``dav-versions.feature``.

phil-davis:
Options:
1) backport like in this PR; or
2) backport without the new code in ``WebDav.php`` (i.e. backport just the ``rdx/behat-variables`` related infrastructure; or
3) close this and only backport after ``dav-versions.feature`` and the related new functionality has been backported.

DeepDiver1975:
oh - I'm sorry - this whole dav versions feature is not yet in stable10. From my pov we don't need this.

phil-davis:
NP - ``rdx/behat-variables`` is not used anywhere in ``stable10`` yet, so yes, it can come [if|when] the dav version feature is backported and then these extra dav versions tests on top of that.

Issue #30392, opened 2018-02-07
mount public link does not work with IE 11


individual-it:
### Steps to reproduce
1. create a public link of a folder
2. visit the new public link
3. add the folder to an other owncloud by using the "Add to your ownCloud" button

### Expected behaviour
after login a dialogue should ask to accept the new link and after accepting a new folder should be listed

### Actual behaviour
after login only the old files and folders are listed

### Server configuration
**Operating system**:
Ubuntu 17.10
**Web server:**
Apache
**Database:**
MySQL
**PHP version:**
7.1.13
**ownCloud version:** (see ownCloud admin page)
10.0.6 (git)
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc9xsmdtnlru",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "192.168.56.1",
            "192.168.0.106"
        ],
        "datadirectory": "\/home\/artur\/www\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/192.168.56.1\/owncloud-core",
        "dbtype": "mysql",
        "logtimezone": "UTC",
        "installed": true,
        "default_language": "en",
        "theme": "",
        "loglevel": "0",
        "maintenance": false,
        "log_type": "owncloud",
        "dbname": "owncloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "debug": false,
        "version": "10.0.6.1",
        "csrf.disabled": true,
        "files_external_allow_create_new_local": "true",
        "singleuser": false,
        "license-key": "***REMOVED SENSITIVE VALUE***",
        "csrf.disabled=false": "",
        "skeletondirectory": "\/home\/artur\/www\/owncloud-core\/tests\/ui\/skeleton",
        "mysql.utf8mb4": true,
        "ldapIgnoreNamingRules": false,
        "filelocking.enabled": "true"
    }
}

```Enabled:
  - comments: 0.3.0
  - dav: 0.3.2
  - diagnostics: 0.1.0
  - enterprise_key: 0.1.3
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
[...]

PR #30439, opened 2018-02-09
skip UI tests that fail because of issue #30392

individual-it:
## Description
this tests fail because of a bug, so better to skip them and revert this PRs after the bug is fixed

## Related Issue
#30392

## Motivation and Context
keep CI clean


phil-davis (approved):

phil-davis:
Note: the commit text says issue "39392". It is actually "30392". Does not break anything :)

PR #30444, opened 2018-02-09
[WIP] Try to avoid filed lock issues in UI tests

phil-davis:
Sometimes at the end of a scenario a file is left "locked" for a user that is now deleted. That causes failures in the next scenario. See what we can do about that.

First hack, just to see if it helps - sleep 5 before deleting users.


phil-davis:
this is not the way

PR #30451, opened 2018-02-12
Catch if parent row exists when we try to add it

tomneedham:
## Description
Sometimes the row that we are trying to insert (in order to use metadata where present) already exists which throws an exception and this cancels the whole repair operation. Instead it is safe to skip this.

## Motivation and Context
Running repair step and getting an exception that the row already exists which stops the repair step.

## How Has This Been Tested?
Manual repair run and unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Repair/RepairMismatchFileCachePath.php
+++ b/lib/private/Repair/RepairMismatchFileCachePath.php
@@ -27,6 +27,7 @@ use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use OCP\Files\IMimeTypeLoader;
 use OCP\IDBConnection;
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 
 /**
  * Repairs file cache entry which path do not match the parent-child relationship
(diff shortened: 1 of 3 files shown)


tomneedham:
@pvince81 I need your assistnace here. According to https://github.com/owncloud/core/blob/master/lib/private/Repair/RepairMismatchFileCachePath.php#L169 the parent entry should not exist - but in the wild we have seen a case where it does exist and throws a DB exception. Hints on why it would not show up in this query - and then suddenly exist?

lib/private/Repair/RepairMismatchFileCachePath.php
```diff
@@ -407,7 +408,11 @@ private function getOrCreateEntry($storageId, $path, $reuseFileId = null) {
 			$values['fileid'] = $qb->createNamedParameter($reuseFileId);
 		}
 		$qb->insert('filecache')->values($values);
-		$qb->execute();
+		try {
+			$qb->execute();
+		} catch (UniqueConstraintViolationException $e) {
+			// Skip if the entry already exists
```
PVince81: we should definitely log this, also include info like file id / parent, whatever can help debug

PVince81:
```
09:51:09 TypeError: Argument 3 passed to OC\Repair\RepairMismatchFileCachePath::__construct() must implement interface OCP\ILogger, none given, called in /var/lib/jenkins/workspace/owncloud-core_core_PR-30451-3XCYYOVY2XYXGG2PZLXZCGTPPGRZHI3L6HJCYE3R73WXJGZU7NHQ/lib/private/Repair.php on line 130 and defined in /var/lib/jenkins/workspace/owncloud-core_core_PR-30451-3XCYYOVY2XYXGG2PZLXZCGTPPGRZHI3L6HJCYE3R73WXJGZU7NHQ/lib/private/Repair/RepairMismatchFileCachePath.php:64
```

PVince81 (approved):
:+1:

PVince81:
please backport

Issue #29287, opened 2017-10-18
All files show wrong modification date (1.1.1970 / "just now")


LinusHoppe:
### Steps to reproduce
1. Install Owncloud
2. Sync files to Owncloud

### Expected behaviour
The files have the correct modification date and are synced correctly.

### Actual behaviour

- My synced folders show the correct modification date, e. g. "10 days ago" at "08.10.2017 14:50" (see folders.png.
- All files (across all folders) show "just now" as modification date, although I've synced them days ago (see files.png). Having no clients connected at the moment, the date remains incorrect.
- I've connected a second device, the files are downloaded. All files show the modification date "01.01.1970 01:00:00" as modification date (see windows_explorer.png).
- The incorrect date lead to clients wo don't sync any files (changed files are not uploaded to owncloud)

### Images
files.png:
![files](https://user-images.githubusercontent.com/32911160/31737980-85d5368c-b449-11e7-8267-219b775956b6.png)
folders.png:
![folders](https://user-images.githubusercontent.com/32911160/31737986-88ee5484-b449-11e7-8656-5cc62802dc55.png)
windows_explorer.png:
![windows_explorer](https://user-images.githubusercontent.com/32911160/31737990-8a960534-b449-11e7-978d-5684dbdf752a.png)

### Server configuration
**Operating system**:
Linux

**Web server:**
Managed Webspace, Apache httpD

**Database:**
MySQL 

**PHP version:**
5.6.30

**ownCloud version:** (see ownCloud admin page)
ownCloud 10.0.3 (stable) 

**Updated from an older ownCloud or fresh install:**
fresh install

**Where did you install ownCloud from:**
ownCloud Web Installer (https://owncloud.org/install/)

**Signing status (ownCloud 9.0 and above):**
https://gist.github.com/GeneralAmbush/5c0989888a4b8273086d14ab2896622d

**The content of config/config.php:**
https://gist.github.com/GeneralAmbush/af4d5ecb8594d8f5298175c75cb14e62

**Are you using external storage, if yes which one:** local/smb/sftp/...
No

**Are you using encryption:** yes/no
Yes

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
No

### Client configuration
**Browser:**
Firefox, Chrome, IE

**Operating system:**
Windows 8.1

#### ownCloud log (data/owncloud.log)
My ownCloud log is 10 MB because of many lines of 
```
[...]

PR #30455, opened 2018-02-12
Check exec and stat method before using it

sharidas:
Before using exec and stat method in 32 bit
machines, a check is required if they are
available or not. Else fallback to the
filemtime(path);

## Description

Check if the functions exec and stat are not disabled before using them.

## Related Issue

https://github.com/owncloud/core/issues/29287

## Motivation and Context

If the exec and stat were not checked before using them the mtime was stored as zero in the filecache table. So this will help us avoid the issue.

## How Has This Been Tested?

- [x]  Added `exec, stat,` to `disable_functions` into file /etc/php/7.0/apache2/php.ini ( distro : Linux Mint 18.3 Sylvia) and restart apache.
- [x]  Modified the Local.php's filemtime method to have a check for `PHP_INT_SIZE === 8`.
- [x] Verified that if exec and stat are disabled the `return filemtime($fullPath);` gets called.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -180,12 +180,26 @@ class Local extends Common {
 			return false;
 		}
 		if (PHP_INT_SIZE === 4) {
-			if (\OC_Util::runningOn('linux')) {
-				return (int) exec ('stat -c %Y '. escapeshellarg ($fullPath));
-			} else if (\OC_Util::runningOn('bsd') || \OC_Util::runningOn('mac')) {
-				return (int) exec ('stat -f %m '. escapeshellarg ($fullPath));
+			/**
+			 * Check if exec is available to use before calling it.
+			 */
+			if (function_exists('exec') === true) {
+				$result = 0;
+				$returnVar = 0;
+				if (\OC_Util::runningOn('linux')) {
+					$result = (int)exec('stat -c %Y ' . escapeshellarg($fullPath), $output, $returnVar);
+				} else if (\OC_Util::runningOn('bsd') || \OC_Util::runningOn('mac')) {
+					$result = (int)exec('stat -f %m ' . escapeshellarg($fullPath), $output, $returnVar);
+				}
+
+				/**
+				 * If the result is zero, then stat is missing.
+				 * Additionally check the return status from the shell.
+				 */
+				if ($returnVar === 0) {
+					return $result;
+				}
 			}
-			return false;
 		}
 		return filemtime($fullPath);
 	}


sharidas:
@PVince81 Is unit test required for this? As this PR is very specific to 32 bit. Sorry if I asked something wrong.

lib/private/Files/Storage/Local.php
```diff
@@ -180,6 +180,14 @@ public function filemtime($path) {
 			return false;
 		}
 		if (PHP_INT_SIZE === 4) {
+			/**
+			 * Check if exec and stat are available to use before calling them.
+			 */
+			if ((function_exists('exec') === false) &&
```
phil-davis: Could use "or ``||`` here - if either of them is missing then there is no point trying.

lib/private/Files/Storage/Local.php
```diff
@@ -180,6 +180,14 @@ public function filemtime($path) {
...
+				(function_exists('stat') === false)) {
```
phil-davis: How does the existence of the PHP function ``stat()`` tested here give any reliable indication of if the *nix command line ``stat`` utilitiy is available?

lib/private/Files/Storage/Local.php
```diff
@@ -180,6 +180,24 @@ public function filemtime($path) {
...
+			 */
+			if (function_exists('exec') === false) {
+				return $this->filemtime($fullPath);
+			}
+
+			/**
+			 * Check if stat command is available to use before calling it in
+			 * exec.
+			 */
+			if (function_exists('shell_exec') === true) {
```
sharidas: One way to check if stat command is available in the environment.

phil-davis: Yes, I guess all we can do is try!
To cover all cases, this ``if`` needs an ``else`` that does ``return $this->filemtime($fullPath)``
And then there will be 3 places doing that, so it will be better to keep a boolean ``$execStatAvailable`` true/false through these tests, then you can have a single place that does something like:
```
if ($execStatAvailable) {
  // do the linux/bsd exec stat stuff
} else {
  return $this->filemtime($fullPath);
}
```

lib/private/Files/Storage/Local.php
```diff
@@ -180,10 +180,39 @@ public function filemtime($path) {
...
+			 */
+			if (function_exists('exec') === false) {
+				$execStatAvailable = false;
+			}
+
+			/**
+			 * Check if stat command is available to use before calling it in
+			 * exec.
+			 */
+			if ((function_exists('shell_exec') === true) && ($execStatAvailable === true)) {
```
phil-davis: At this point we know ``$execStatAvailable`` is ``true`` and we know that the ``exec`` function exists.
So can we remove this whole ``if`` block from around the code, and use the ``exec`` function instead of ``shell_exec`` to see the result of ``which`` ?

sharidas: @phil-davis  updated the PR.

phil-davis (commented):
This seems what can reasonably be done to know what command is available.
@PVince81 ?

lib/private/Files/Storage/Local.php
```diff
@@ -180,10 +180,37 @@ public function filemtime($path) {
...
+			 */
+			if (function_exists('exec') === false) {
+				$execStatAvailable = false;
+			}
+
+			/**
+			 * Check if stat command is available to use before calling it in
+			 * exec.
+			 */
+			exec(sprintf("which %s", escapeshellarg('stat')), $output, $result);
```
PVince81: if the function doesn't exist, wouldn't this cause an error if we still call it ?

I'd suggest something simpler:
1. if `exec` function not available: fall back directly to regular filemtime
2. if `exec` function available, always call "stat" and check the exit code from the shell. The exit code will tell you whether the stat command succeeded or not. No need for an additional `which` call.

sharidas: Updated the PR.

lib/private/Files/Storage/Local.php
```diff
@@ -180,12 +180,29 @@ public function filemtime($path) {
...
 		if (PHP_INT_SIZE === 4) {
-			if (\OC_Util::runningOn('linux')) {
-				return (int) exec ('stat -c %Y '. escapeshellarg ($fullPath));
-			} else if (\OC_Util::runningOn('bsd') || \OC_Util::runningOn('mac')) {
-				return (int) exec ('stat -f %m '. escapeshellarg ($fullPath));
+			/**
+			 * Check if exec is available to use before calling it.
+			 */
+			if (function_exists('exec') === false) {
+				return filemtime($fullPath);
```
PVince81: this approach repeats the `filemtime()` line three times

please use a better approach

sharidas: Updated the PR. The call to filemtime happens only once now.
1) if `exec` is not found fall backs to `filemtime`.
2) if `exec` is there and `stat` is there then returns the result else again falls back to `filemtime`.

lib/private/Files/Storage/Local.php
```diff
@@ -180,12 +180,26 @@ public function filemtime($path) {
...
+					$result = (int)exec('stat -c %Y ' . escapeshellarg($fullPath), $output, $returnVar);
+				} else if (\OC_Util::runningOn('bsd') || \OC_Util::runningOn('mac')) {
+					$result = (int)exec('stat -f %m ' . escapeshellarg($fullPath), $output, $returnVar);
+				}
+
+				/**
+				 * If the result is zero, then stat is is missing.
+				 * Additionally checking the return status from the shell.
+				 */
+				if (($result === 0) && ($returnVar === 0)) {
```
PVince81: hmm where are we returning the result from the stat command on success ?

PVince81: yes, but I don't see where `$returnVar` is returned from the function ? it seems the code would anyway go down a execut `filemtime()` in that case

sharidas: Updated the PR. The returnVar is tested for the command stat's exit status. So if the exit status is `0` then the command is there and executed successfully and return the result. Else fallback.

lib/private/Files/Storage/Local.php
```diff
@@ -180,12 +180,26 @@ public function filemtime($path) {
...
+				$result = 0;
+				$returnVar = 0;
+				if (\OC_Util::runningOn('linux')) {
+					$result = (int)exec('stat -c %Y ' . escapeshellarg($fullPath), $output, $returnVar);
+				} else if (\OC_Util::runningOn('bsd') || \OC_Util::runningOn('mac')) {
+					$result = (int)exec('stat -f %m ' . escapeshellarg($fullPath), $output, $returnVar);
+				}
+
+				/**
+				 * If the result is zero, then stat is is missing.
```
phil-davis: s/is is/is
repeated word in comment

lib/private/Files/Storage/Local.php
```diff
@@ -180,12 +180,26 @@ public function filemtime($path) {
...
+				 * Additionally checking the return status from the shell.
```
phil-davis: While you are changing comment text:
s/checking/check

PVince81 (approved):
:+1:

PVince81:
please backport this so people still on 32-bit systems can benefit from it

sharidas:
The backport PR: https://github.com/owncloud/core/pull/30546

PR #30471, opened 2018-02-13
Address a few code inspections

butonic:
Only for php files in the root folder ... I initially also addressed `'Exception', superclass of exception class 'Error', has already been caught` but it turns out that seems to be a false positive in the PHPStorm Inspectien triggered by the Error Polyfill from Paragonie random_compat or Symphony. Double checked with

```php
try {
	throw new Error();
} catch (Exception $e) {
	echo 'caught Exception';
} catch (Error $e) {
	echo 'caught Error';
}
// gives 'caught Error'
```

@DeepDiver1975 @PVince81 Should we return 500 Status for the JSON responses in cron.php? In theory yes. The old code didn't. It uses the borked oc json error in payload ... I added a TODO because I think we should change it (but not as part of this PR).

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -72,11 +72,17 @@ try {
 
 	// Exit if background jobs are disabled!
 	$appMode = \OCP\BackgroundJob::getExecutionType();
-	if ($appMode == 'none') {
+	if ($appMode === 'none') {
 		if (OC::$CLI) {
 			echo 'Background Jobs are disabled!' . PHP_EOL;
 		} else {
-			OC_JSON::error(['data' => ['message' => 'Background jobs disabled!']]);
+			$r = new \OCP\AppFramework\Http\JSONResponse([
+				'data' => [ 'message' => 'Background jobs disabled!' ],
+				'status' => 'error'
+			]
+			// TODO use 500 error response?
+			);
+			echo $r->render();
 		}
 		exit(1);
 	}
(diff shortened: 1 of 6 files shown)


cron.php
```diff
@@ -72,11 +72,17 @@
 
 	// Exit if background jobs are disabled!
 	$appMode = \OCP\BackgroundJob::getExecutionType();
-	if ($appMode == 'none') {
+	if ($appMode === 'none') {
 		if (OC::$CLI) {
 			echo 'Background Jobs are disabled!' . PHP_EOL;
 		} else {
-			OC_JSON::error(['data' => ['message' => 'Background jobs disabled!']]);
```
DeepDiver1975: I'd just leave it this way for the time being.

butonic: using the json response as recommended in the deprecation text for OC_JSON. No change in behavior. If we want to get rid of deprecated code we need to start somewhere.

DeepDiver1975: Proper fix for this: split cron.php into two pieces:
1. occ command
2. pure http route for ajax and web cron

DeepDiver1975 (changes requested):
the JSONResonse changes look strange ....

PVince81:
conflicts.

if no agreement and p4-low I suggest closing

moving to "triage" meanwhile

cron.php
PVince81: => https://github.com/owncloud/core/pull/31914

DeepDiver1975:
conflicting heavily - please resubmit

PR #30478, opened 2018-02-13
[WIP] More flexible file actions API in web UI

PVince81:
## Description
Make it possible to register functions instead of action specifications.
The function takes a `FileInfoModel` as input and decides whether to return an array of matching actions or not. This way it is possible to customize file actions based on any kind of attributes of the file model.

## Related Issue
Required for https://github.com/owncloud/core/pull/30106 list pending shares.

## Motivation and Context
Required for https://github.com/owncloud/core/pull/30106 list pending shares where the file actions for pending shares would depend on the share state.

## How Has This Been Tested?
Not yet.

## Types of changes

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,11 +168,11 @@ test-external: $(composer_dev_deps)
 
 .PHONY: test-js
 test-js: $(nodejs_deps)
-	NODE_PATH='$(NODE_PREFIX)/node_modules' $(KARMA) start tests/karma.config.js --single-run
+	ulimit -Sv 2000000 && NODE_PATH='$(NODE_PREFIX)/node_modules' $(KARMA) start tests/karma.config.js --single-run
 
 .PHONY: test-js-debug
 test-js-debug: $(nodejs_deps)
-	NODE_PATH='$(NODE_PREFIX)/node_modules' $(KARMA) start tests/karma.config.js
+	ulimit -Sv 2000000 && NODE_PATH='$(NODE_PREFIX)/node_modules' $(KARMA) start tests/karma.config.js
 
 .PHONY: test-integration
 test-integration: $(composer_dev_deps)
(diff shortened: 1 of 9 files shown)


PVince81:
Fixed the tests, but I'm a bit worried about a potential memory leak. After fixing some errors I noticed that I could run all tests with 2 GB memory but with recent changes I need to set it to 3 GB to avoid phantomjs crashes.

- [ ] worry about memory leak
- [ ] worry about this not being the best solution as it introduces additional overhead by having to iterate over all action returning functions

Maybe a simpler solution would just be to:
1) add an optional `isVisible` handler for every action which receive the current file to find out whether or not to render the action
2) add a global handler `onAfterGettingActionsList(actions)` which would allow handlers to modify the actions list and add/delete entries when needed.

PVince81:
Basically we have two requirements:
1) the ability to hide actions that were already registered for specific file properties (conditional filtered)
and
2) the ability to display specific actions based on specific file properties

tried to think of simpler ways to do it:

- Approach 1:
    - hacky: create a subclass of FileActions and override getActions or `_defaultRenderAction` to pre-filter existing actions
    - define a FileActions entry for "all" but filter out actions (here: accept/reject) in the `render` method. or put the logic in `_defaultRenderAction`

- Approach 2:
    - like Approach 1 but provide a hook to filter actions called after `getActions()` inside the `display()` method.

- Approach 3:
   - this PR
   - but changes too much and will likely break many things

PVince81:
I went with approach 2 in https://github.com/owncloud/core/pull/30106.

Anything else is too complicated and prone to errors. We'll likely want to rewrite file actions from scratch and ditch the legacy stuff at some point instead of working around it...

phil-davis:
Removed backport label since this PR is closed/superceded

PR #30413, opened 2018-02-08
New formats for integration test given-when-then step text

phil-davis:
## Description
Change the "integration" (=API) gherkin test steps to a new "standard" format.
Refactor the code behind to do what the steps say - adding new methods etc to handle the extra variations in possible step text.
Behat Gherkin is able to match any of multiple ``Given`` ``When`` annotations in the PHPdoc for a single method, so make use of that when there are different wordings for a step which have the same parameter order. 

## Related Issue
https://github.com/owncloud/QA/issues/517

## Motivation and Context
Make the integration-API test steps have a consistent agreed style. Then we can move on to analysing the test coverage, effectiveness, merging in UI tests, renaming the "integration" folder to "acceptance" and so on...

## How Has This Been Tested?
CI passes

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Refactoring (tests only)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30488, opened 2018-02-14
[stable10] New formats for integration test given-when-then step text

phil-davis:
Backport #30413


phil-davis:
Note: conflicts were resolved when backporting to not include ``dav-versions.feature`` code, because that feature has not been backported to ``stable10``

phil-davis:
This needs to be approved and ready to preferably be merged at the same time as #30413 so that apps will have a consistent view of integration test code in both ``core`` branches.

individual-it (approved):

PR #30489, opened 2018-02-14
Email notifications (core changes)

jvillafanez:
## Description
Core changes required to send email notifications. Changes focus on registration mechanism and notification serialization.

## Related Issue
New feature

## How Has This Been Tested?
Tested via app as well as unittests

## Types of changes

## Checklist:


lib/private/Notification/Events/RegisterNotifierEvent.php
```diff
@@ -0,0 +1,61 @@
...
+	/** @var Manager */
+	private $manager;
+
+	public function __construct (Manager $manager) {
+		$this->manager = $manager;
+	}
+
+	/**
+	 * Register the notification consumer in the notification manager
+	 * @param INotifier $notifier the notification consumer to be registered
```
PVince81: I wonder if we should call it "preparer" or something instead of consumer, because the consumer usually comes after the preparation

lib/private/Notification/Manager.php
```diff
@@ -82,6 +103,27 @@ public function registerNotifier(\Closure $service, \Closure $info) {
 		$this->notifiersInfo = [];
 	}
 
+	/**
+	 * INTERNAL USE ONLY!! This method isn't part of the IManager interface
+	 */
```
PVince81: `@internal` ?

same for the next methods

lib/private/Notification/Manager.php
```diff
@@ -237,4 +298,205 @@ public function getCount(INotification $notification) {
...
+	/**
+	 * @param INotification $notification the notification to be serialized
+	 * @param bool $forceIncomplete whether this method should serialize a INotification instance with
+	 * incomplete data or throw an exception if the notification isn't complete
+	 * @return string a json string with the all the data
+	 * @throws \OCP\Notification\Exceptions\IncompleteSerializationException if the notification
+	 * doesn't * have all the required fields for the notification AND the $forceIncomplete flag
+	 * isn't true. This exception musn't be thrown if the $forceIncomplete flag is true.
+	 */
+	public function serializeNotification(INotification $notification, $forceIncomplete = false) {
```
PVince81: who will use this ? I don't see it used anywhere in this PR and also not in https://github.com/owncloud/notifications_mail/pull/3/files (searched with ctrl+f)

PVince81: wouldn't it be better to be part of INotification instead ?

lib/private/Notification/Manager.php
```diff
@@ -237,4 +298,205 @@ public function getCount(INotification $notification) {
...
+		if (!empty($missingMandatoryFields) && !$forceIncomplete) {
+			$missingFields = implode(',', $missingMandatoryFields);
+			throw new IncompleteDeserializationException("some required fields to deserialize the notification are missing: $missingFields");
+		}
+
+		$notificationFields = [
+			'app' => 'setApp',
+			'user' => 'setUser',
+			'dateTime' => 'setDateTime',  // requires special handling
+			'objectType+objectId' => 'setObject',
```
PVince81: a lot of obscure magic here with the "+" sign, is this complex approach and magic really needed ?

lib/public/Notification/Events/AbstractRegisterConsumerEvent.php
```diff
@@ -0,0 +1,65 @@
...
+ * @since 10.0.8
+ */
+abstract class AbstractRegisterConsumerEvent extends Event {
+	/**
+	 * Implementations should use this name
+	 */
+	const NAME = 'notification.register.consumer';
+
+	/**
+	 * Empty implementation in order to prevent stopping the propagation of this event
```
PVince81: preventing people to stop propagation when they intend to by calling this method ? any reason to change default behavior ?

lib/public/Notification/Events/AbstractRegisterConsumerEvent.php
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	/**
+	 * Register the app in the notification manager. Implementations must take care of injecting
+	 * the specific notification manager implementation.
+	 * @param IApp $app the notification app to be registered
+	 *
+	 * @since 10.0.8
+	 */
+	abstract public function registerNotificationConsumer(IApp $app);
```
PVince81: As mentioned before, I don't like "App" here and I know it's a legacy thing. Maybe this is the opportunity to make a new interface `INotificationConsumer` and make it inherit `IApp` for backward compatibility ?

lib/public/Notification/IManager.php
```diff
@@ -63,4 +63,31 @@ public function createNotification();
...
 	 */
 	public function hasNotifiers();
+
+	/**
+	 * @param INotification $notification the notification to be serialized
+	 * @param bool $forceIncomplete whether this method should serialize a INotification instance with
+	 * incomplete data or throw an exception if the notification isn't complete
+	 * @return string the serialized notification
+	 * @throws \OCP\Notification\Exceptions\IncompleteSerializationException if the notification
+	 * doesn't * have all the required fields for the notification AND the $forceIncomplete flag
```
PVince81: stay `*` in the doc comment ?

PVince81 (commented):
See comments above.

If there is currently no consumer for the serialization/deserialization I'd prefer to move this to a separate PR for future use and not clutter the code with complex unused code.

lib/private/Notification/Events/RegisterNotifierEvent.php
jvillafanez: I thought I changed this.... Too much copy & paste. The consumer is the IApp, the one sending the notifications is the INotifier. I have to adjust the comment.

lib/private/Notification/Manager.php
jvillafanez: 👍

lib/private/Notification/Manager.php
jvillafanez: This isn't being used right now, but we might need to used it soon if we choose to store the notifications for the email.

I can say for sure what is the best in this case. On the one hand, having the object creation and handling outside of the object itself makes sense. On the other hand, having it inside the object could give us an easier implementation. If we choose the second option, we would need to create a new instance to import the data, which feels weird.

We also need to take into account that the Manager is the one creating the notifications, and it makes sense to be the one serializing and deserializing them. We can also have better control on the serialization process if we decide to include other kind of mechanisms apart from json or to use specialized objects for such matter (no plans for this soon)

lib/private/Notification/Manager.php
jvillafanez: Maybe I should try the simplest approach even though there will likely be a lot of duplication, and the way it will be extended will also add more duplication.

lib/public/Notification/Events/AbstractRegisterConsumerEvent.php
jvillafanez: I'd prefer to avoid adding interfaces just for the sake of renaming a bad named interface. I feel this will add more confusion in the long term. Just think about how the all the interfaces for user handling turned out.
We've decided to play along with what we already have. We'll see if the new methods trying to use the new terminology go for the best or for the worst.

lib/public/Notification/IManager.php
jvillafanez: fixing...

jvillafanez:
> If there is currently no consumer for the serialization/deserialization I'd prefer to move this to a separate PR for future use and not clutter the code with complex unused code.

It depends on the plans and the schedule for the "notifications" and "notifications_mail" apps. I completely agree the code is complex, but as long as the unittests cover all the posible scenarios it should be fine to be merged so the apps can use the functionality whenever the apps want.

lib/private/Notification/Manager.php
PVince81: at this point I'd then prefer to not have "dead code" which we might or might not want to use soon

if you need it in a subsequent PR for extending the current features, then put these methods there

PVince81: In a more broader scope, you are right that we need to think about how things will look like in the near future. However in this one particular case, there is no additional cost if we add this later.

The minimum thing we need to keep in mind is to not put in stuff that could block future expansion or increase the work load in case of future expansion.

lib/public/Notification/Events/AbstractRegisterConsumerEvent.php
PVince81: fair enough

jvillafanez:
I'll wait until this PR is merged so I can get the https://github.com/owncloud/core/pull/30489/commits/d077910ed8181e6c4bbb66680574685828b40985 commit and reverse-apply it in a new branch to recover the feature.

jvillafanez:
Anyone to kick jenkins?

phil-davis:
You might as well just rebase to kick Jenkins - all the other CI jobs are much quicker than Jenkins so it will not slow down the elapsed time to run all CI again.

jvillafanez:
I'll need some explanation about what happened with the rebase I did here...

Below is my local repo commit order as shown with "git log"
```
927850c (HEAD, origin/email_notifications, email_notifications) Remove serialize
ec56751 Clarify comments
be31738 Mockup event dispatcher, adjust tests and add "since" tags
6df6c16 Add event-based registration for notifiers and notification apps
d1530ca Adjust code and tests
981b989 Add tests for the serialization of the notifications
d46a0a8 Add support to serialize the notifications
0d49b6d (origin/master, origin/HEAD, master) Merge pull request #30324 from indi
```
The order isn't the same as displayed in the PR. I want to think this is a visualization problem... :confused:

jvillafanez:
Fresh cloned repo has the same id sequence as my local one, so I guess it's a bug in github (or I'm seeing something weird)

lib/public/Notification/Exceptions/NotifierIdInUseException.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ */
+
+namespace OCP\Notification\Exceptions;
+
+/**
+ * Class NotifierIdInUseException
+ *
+ * @package OCP\Notification\Exceptions
+ * @since 10.0.7
```
PVince81: 10.0.8

PVince81 (approved):
please adjust since versions then this looks ok from my pov :+1:

PVince81:
please backport to stable10

jvillafanez:
Backport to stable10 in https://github.com/owncloud/core/pull/30613

PR #30491, opened 2018-02-15
Bump version to 10.0.7 in master

phil-davis:
This could be the actual next version.


phil-davis:
Putting this here because there seems to be a Jenkins problem at the moment. It is an easy way to trigger a Jenkins run that should pass.

phil-davis:
Yes, I just did this to demostrate to myself that Jenkins was broken this morning, and not code in other PRs of mine.

PR #30497, opened 2018-02-15
[test] UI tests disable previews - trigger CI

phil-davis:
This is just a PR to trigger CI.
That makes it easy to restart drone a few times and see how much more reliable the UI tests might be with previews disabled.


patrickjahns:
I've added a couple of runs - should give us some insight by tomorrow

patrickjahns:
Over the weekend we can add a couple more runs that should give also more insight

phil-davis:
the main PR has been merged

PR #30503, opened 2018-02-16
Properly encode returned Webdav URL in ViewController

PVince81:
## Description
Affects the URL as displayed in the files app cog icon and also the one
returned in the "Webdav-Location" header

## Related Issue
Follow up of https://github.com/owncloud/core/pull/30383#discussion_r168739091

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Controller/ViewController.php
+++ b/apps/files/lib/Controller/ViewController.php
@@ -220,7 +220,7 @@ class ViewController extends Controller {
 			return $item1['order'] - $item2['order'];
 		});
 		$nav->assign('navigationItems', $navItems);
-		$nav->assign('webdavUrl', $this->urlGenerator->getAbsoluteUrl($this->urlGenerator->linkTo('', 'remote.php') . '/dav/files/' . $user . '/'));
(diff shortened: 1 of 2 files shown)


SamuAlfageme:
@PVince81 does seem to work like a charm now 👍 

I've noticed there's a difference on how the `rawurlencode()` and `\OCP\Util::encodePath()` work out. i.e. normal `PROPFIND`s  contents look like this:

```xml
<d:response>
    <d:href>/remote.php/dav/files/lun@/tigr@/leon%c3%a9/</d:href>
    <d:propstat>
```

Notice how `@` are not translated into `%40` - IIRC with this PR the `Webdav-Location`for that response would be `/remote.php/dav/files/lun%40/tigr@/leon%c3%a9/` (the `uid` part is `rawurlencode`d). Since both routes resolve correctly I'll consider this pretty minor though, dunno if it's something we might want to adjust. 

LGTM otherwise.

PVince81:
Yeah, I remember that difference as Sabre has its own encoding logic. I seem to remember that I had to encode the "@" signs to %40 in Javascript else some Webdav backend code would not work correctly. So here I opted for the same approach to keep it safe.

Thanks for testing!

DeepDiver1975 (approved):
Tested :+1:

DeepDiver1975:
backporting ....

PR #30508, opened 2018-02-16
Fixing variable dependency

Julian1998:
oc_current_user doesn't exist anymore and has to be replaced by OC.getCurrentUser() for supporting all necessary versions.

### DIFF
--- a/apps/files/js/files.js
+++ b/apps/files/js/files.js
@@ -114,11 +114,12 @@
 				return;
 			}
 
+			var currentUser = OC.getCurrentUser().uid;
 			var usedSpacePercent = $('#usedSpacePercent').val(),
 				owner = $('#owner').val(),
 				ownerDisplayName = $('#ownerDisplayName').val();
 			if (usedSpacePercent > 98) {
-				if (owner !== oc_current_user) {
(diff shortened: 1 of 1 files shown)


DeepDiver1975 (approved):

apps/files/js/files.js
```diff
@@ -114,11 +114,12 @@
 				return;
 			}
 
+			var currentUser = OC.getCurrentUser();
```
DeepDiver1975: wait ..... getCurrentUser returns an array

bildschirmfoto von 2018-02-19 08-51-39

DeepDiver1975 (changes requested):

PVince81:
Nice catch! Please see the above.

apps/files/js/files.js
Julian1998: Sure! I knew that - But calendar and co. need an object

Julian1998: Sure! I knew that - But calendar needs this object. oc_current_user and getCurrentUser() are equal!

Julian1998:
Sorry...my fault - Fixed

PVince81:
squashed and rebased

PR #30512, opened 2018-02-17
[WIP] Try to see any file locked notification

phil-davis:
On uploads, check that there are no notifications.
If there are, then fail and report the notification text.
Maybe we can see in the UI that the file locked happened.


phil-davis:
further progress in #30513

PR #30530, opened 2018-02-19
[WIP] Error handling for skeleton trigger in ui tests

phil-davis:
## Description
Catch ``BadResponseException`` when we try to download ``lorem.txt`` as the way of initializing a new user's skeleton. If this happens, then loop around trying again. 

## Motivation and Context
The UI tests on ``drone`` quite often have a problem downloading ``lorem.txt`` to initialize a new user. They sometimes get a ``404`` ``not found`` and the scenario bombs out. Maybe it just needs to wait a little bit and try again after a user has just been created, before access to the user files is possible.

## How Has This Been Tested?
CI

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30532, opened 2018-02-19
[test] Error handling forskeleton run upload tests 01

phil-davis:
Run upload tests 10 times and see what happens.

On top of #30530

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -408,3 +408,57 @@ matrix:
       BEHAT_SUITE: upload
       DB_TYPE: mariadb
       USE_SERVER: true
+
+    - PHP_VERSION: 7.1
+      TEST_SUITE: selenium
+      BEHAT_SUITE: upload
+      DB_TYPE: mariadb
+      USE_SERVER: true
+
+    - PHP_VERSION: 7.1
+      TEST_SUITE: selenium
+      BEHAT_SUITE: upload
+      DB_TYPE: mariadb
+      USE_SERVER: true
+
+    - PHP_VERSION: 7.1
+      TEST_SUITE: selenium
+      BEHAT_SUITE: upload
+      DB_TYPE: mariadb
+      USE_SERVER: true
+
+    - PHP_VERSION: 7.1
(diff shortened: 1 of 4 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/core/4043/246

```
Feature: File Upload
  
  As a user
  I would like to be able to upload files via the WebUI
  So that I can store files in ownCloud

  Background:                                             # /drone/src/tests/ui/features/upload/upload.feature:8
Downloading file lorem.txt to initialize user user1 and got response status 404

Downloading file lorem.txt to initialize user user1 and got response status 404

Downloading file lorem.txt to initialize user user1 and got response status 404

Downloading file lorem.txt to initialize user user1 and got response status 404

Downloading file lorem.txt to initialize user user1 and got response status 404

    Given these users exist:                              # FeatureContext::theseUsersExist()
      | username | password | displayname | email        |
      | user1    | 1234     | User One    | u1@oc.com.np |
      │ Downloading file lorem.txt to initialize user user1 and got response status 404
      │ Downloading file lorem.txt to initialize user user1 and got response status 404
      │ Downloading file lorem.txt to initialize user user1 and got response status 404
      │ Downloading file lorem.txt to initialize user user1 and got response status 404
      │ Downloading file lorem.txt to initialize user user1 and got response status 404
      │ 
      Exception: could not initialize user. 404 <?xml version="1.0" encoding="utf-8"?>
      <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
        <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
        <s:message>File with name lorem.txt could not be located</s:message>
      </d:error>
in /drone/src/tests/ui/features/bootstrap/BasicStructure.php:272
```
so the file does not magically appear quickly.

Trying with a long ``sleep(2)`` between tries to make sure it is a "hard" error.

phil-davis:
The upload suite has run 30 or more times and the "404 after user create" error does not happen any more. I guess it was somehow caused by the file locks being left behind or something like that, which has been fixed in master.

PR #30536, opened 2018-02-19
Stop searching apps in OC::SERVERROOT sibling

VicDeo:
## Description
Remove fallback to `OC::$SERVERROOT . '/../apps'` when 
`apps_paths` is not set via `config.php`

## Motivation and Context
1. This is not documented. 
2. I don't think it is used by anybody.
3. Why just not put it into `config.php` as an 'apps_paths' content
4. I can't stop thinking of this piece of code

## How Has This Been Tested?
commented in `config.php`
```
 /* 'apps_paths' => 
  array (
    0 => 
    array (
      'path' => '/dev/oc-tmp/apps',
      'url' => '/apps',
      'writable' => false,
    ),
  ), */
```

## Types of changes

## Checklist:

### DIFF
--- a/lib/base.php
+++ b/lib/base.php
@@ -198,12 +198,6 @@ class OC {
 			}
 		} elseif (file_exists(OC::$SERVERROOT . '/apps')) {
 			OC::$APPSROOTS[] = ['path' => OC::$SERVERROOT . '/apps', 'url' => '/apps', 'writable' => true];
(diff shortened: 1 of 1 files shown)


PVince81 (approved):
:+1: as you said it doesn't break regular use cases.

However I'd like to not backport this, so this cleanup will be only in 10.1

PR #30538, opened 2018-02-20
UI-tests clear lock after scenario

individual-it:
## Description
The former `releaseAll` function was broken, every of the if statements did run the same code, but that actually did not clear anything because a string was passed and not an int.
I've changed it so that:
1. sending a 'delete'  to '/apps/testing/api/v1/lockprovisioning/{type}' clears locks of the given types that were set previously by the testing app
2. sending a 'delete'  to '/apps/testing/api/v1/lockprovisioning/' clears locks of all types that were set previously by the testing app
3. sending a 'delete'  to '/apps/testing/api/v1/lockprovisioning/' and `global=true` in the body of the requests clears all locks in the database regardless of if they were set by the testing app or not

Additionally the UI tests now use that functionality to clear locks after every scenario

## Motivation and Context
sometimes UI tests break because of not released locks after user deletion.

## How Has This Been Tested?
set locks manually, check that they were released after a test-run

## Types of changes

## Checklist:

### DIFF
--- a/apps/testing/locking/provisioning.php
+++ b/apps/testing/locking/provisioning.php
@@ -59,7 +59,7 @@ class Provisioning {
 	}
 
 	/**
-	 * @return ILockingProvider
+	 * @return FakeDBLockingProvider
 	 */
 	protected function getLockingProvider() {
 		if ($this->lockingProvider instanceof DBLockingProvider) {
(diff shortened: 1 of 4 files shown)


apps/testing/locking/provisioning.php
```diff
@@ -199,24 +199,36 @@ public function isLocked(array $parameters) {
...
-		foreach ($this->config->getAppKeys('testing') as $lock) {
-			if (strpos($lock, 'locking_') === 0) {
-				$path = substr($lock, strlen('locking_'));
-
-				if ($type === ILockingProvider::LOCK_EXCLUSIVE && $this->config->getAppValue('testing', $lock) == ILockingProvider::LOCK_EXCLUSIVE) {
-					$lockingProvider->releaseLock($path, $this->config->getAppValue('testing', $lock));
-				} else if ($type === ILockingProvider::LOCK_SHARED && $this->config->getAppValue('testing', $lock) == ILockingProvider::LOCK_SHARED) {
-					$lockingProvider->releaseLock($path, $this->config->getAppValue('testing', $lock));
-				} else {
-					$lockingProvider->releaseLock($path, $this->config->getAppValue('testing', $lock));
```
phil-davis: In the old code, one of these ``if/elseif/else`` would have been executed on every pass through. And all 3 ``releaseLock`` calls are the same??? So even if the caller has specified a ``$type`` it would make no difference.

Your code looks better :)

phil-davis (approved):

individual-it:
backport in #30539

PR #30541, opened 2018-02-20
[test] UI upload suite over-and-over

phil-davis:
The UI tests on drone quite often have a problem downloading lorem.txt to initialize a new user. They sometimes get a 404 not found and the scenario bombs out. This has been noticed most often in the first scenario of the upload suite.

Run the upload suite 10 times so we can hopefully catch a few examples of the problem.

## Types of changes

## Checklist:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -408,3 +408,57 @@ matrix:
       BEHAT_SUITE: upload
       DB_TYPE: mariadb
       USE_SERVER: true
+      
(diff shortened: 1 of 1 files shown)


phil-davis:
The upload suite has run 30 times and the "404 after user create" error does not happen. I guess it was somehow caused by the file locks being left behind or something like that, which has been fixed in master.

PR #30535, opened 2018-02-19
Merge Hotfix 10.0.7 back to stable10

PVince81:
Merge hotfix 10.0.7 branch back into stable10

PR #30569, opened 2018-02-21
hotfix branch rebased onto stable10

PVince81:
This is the hotfix-10.0.7 branch **before** the merge commit from https://github.com/owncloud/core/pull/30535, rebased onto stable10.

I'm a bit worried that those version changing commits will not be accurate after merge.

In the future we should refrain to backport stuff onto stable10 if the target is the hotfix branch. They will then eventually land on stable10 through the final "merge release branch into stable10" and keep their proper order.

Thoughts ? Should we move forward with this PR this time and improve next time ?

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,7 +2,7 @@ workspace:
   base: /drone
   path: src
 
-branches: [master, stable10, stable9.1, stable9]
+branches: [master, stable10, stable9.1, stable9, hotfix-*]
 
 clone:
   git:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ branches:
   only:
     - master
     - /^stable\d+(\.\d+)?$/
+    - /^hotfix-.*$/ 
 
 addons:
   apt: &common_apt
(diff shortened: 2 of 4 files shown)


phil-davis (commented):
The first 2 commits that just turn on CI for the hotfix branch are not wanted in stable10 - we would need to revert them anyway if we merge this.

All the functional code is already in stable10 (I hope!), so why not just make a new branch off stable10 and cherry-pick the last 3 commits here, push it and make a PR?

That will get the changelog and version files up-to-date in stable10.

PVince81:
ok here is what I'm going to do: I'm going to send this PR and the other one to trash.

On stable10 I'll set the version to 10.0.8 pre-alpha so at least we have a higher number than 10.0.7. I don't want to merge any of these version changing commits without having the proper history attached in between.

phil-davis:
> On stable10 I'll set the version to 10.0.8 pre-alpha so at least we have a higher number than 10.0.7
Good idea - because the code in ``stable10`` is well ahead of the code that is being released as ``10.0.7`` - putting a ``10.0.7`` version bump at the end of the current ``stable10`` branch would mislead people into thinking that was like a tag point for ``10.0.7``

phil-davis:
and it would be good to insert the changelog changes also to ``stable10`` so it can move on from there for 10.0.8 changelog.

PR #30573, opened 2018-02-21
Teach UI tests to know if a user should exist or not

phil-davis:
## Description
1a) change step text from ``that is member of these groups`` to ``that is a member of these groups`` (minor grammar)
1b) Fix ``displayName`` and ``email`` parameter positions in a ``addUserToCreatedUsersList`` call (small bug noticed)
2) In step ``When I create a user with the name`` allow the form ``When I attempt to create a user with the name``. When the latter form is used, remember that the user is not expected to exist. Still attempt to delete the user at the end of the scenario (in case it did get created) but do not emit output spam when the delete fails.

## Motivation and Context
When a test scenario attempts to create a user, but it is purposely expected to not succeed (invalid username, no password etc), the code at the end of the scenario tries to delete the user that has not actually been created. The delete "fails" and the test output is spammed with output like:
```
INFORMATION: could not delete user. 'a1'400 <?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>400</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```
this sort of "normal test output noise" makes it difficult for developers trying to diagnose what really went wrong vs normal spam.

## How Has This Been Tested?
Local UI login feature runs

## Types of changes

## Checklist:

### DIFF
--- a/tests/ui/features/bootstrap/BasicStructure.php
+++ b/tests/ui/features/bootstrap/BasicStructure.php
@@ -439,17 +439,17 @@ trait BasicStructure {
 	 */
 	public function tearDownScenarioDeleteCreatedUsersAndGroups() {
 		$baseUrl = $this->getMinkParameter("base_url");
-		foreach ($this->getCreatedUserNames() as $user) {
+		foreach ($this->getCreatedUsers() as $username => $user) {
 			$result = UserHelper::deleteUser(
 				$baseUrl,
-				$user,
+				$username,
 				"admin",
 				$this->getUserPassword("admin")
 			);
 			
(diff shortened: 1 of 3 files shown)


phil-davis:
This introduces the idea that a ``When`` step that does something that is expected to fail can have the words ``attempt to`` in the step text. Then the code underneath is able to know that the action is not really expected to work, and can do whatever is appropriate in response.

Is that a good way to write scenarios?

It does make it clear to the business user reading the scenario that the action is just an "attempt", giving the clue that some error behaviour is about about to come.

individual-it (approved):

PR #30578, opened 2018-02-22
Refactoring and cleanup of drone

patrickjahns:
This PR will superseed https://github.com/owncloud/core/pull/29838

Intention:
- cleanup the .drone.yml pipeline
- prepare for having files_external, integration and acceptance tests properly setup on drone


DeepDiver1975:
@patrickjahns can you please take into account that we need to test different db version?

Something like that: https://github.com/owncloud/core/pull/30145/files#diff-3216dfff0ed3e301453e6799e8c367e2R94

THX

patrickjahns:
@DeepDiver1975 
In the *long run*, testing more databases (what is our policy here on officially supported versions?) should be part of the daily ("nightly") CI infrastructure. 
We might need to evaluate also here, if at the daily run it makes sense to run all tests (unit, integration, api, ui ) with all supported database versions (+ supported php versions ?)

It might also be enough, to have the full massive test scenario on Alpha/Beta/RC and do a limited cross cut for the nightly basis.

*short term* we could (ab)use drone for testing more databases on PR basis. But it can only be until the nightly / daily runs cover this.
One thing I am uncertain: the feasibility of testing all supported php versions with all supported databases. 

For this PR, i'd like to focus on:
- one basis for setting up OC for different test runs (unit / integration / ui)
- a basis for files_external tests (and external tests)

DeepDiver1975:
> In the long run, testing more databases (what is our policy here on officially supported versions?) should be part of the daily ("nightly") CI infrastructure.
> We might need to evaluate also here, if at the daily run it makes sense to run all tests (unit, integration, api, ui ) with all supported database versions (+ supported php versions ?)

well - because the nightly tests do not exist this is out of discussion from my pov

(TBH: this was one benefit of jenkins where the same setup could be used for PR testing as well as nightly testing - no idea what you have in mind for next generation nightly testing and if drone has something ....)

With respect to e.g. postgres I soon-ish need the capability to test with postgres 10 and 9.4 in parallel - both as the minimum and maximum verisons we want/will/do support. When working on this CI has to test both dbs.
running all tests on 9.4, 9.5, 9.6 and 10.0 is out of scope - I agree to this.

My request as of above was more in respect to build a nice mechanism which allows developers to add different db configs/version to be run in parallel.

THX

DeepDiver1975:
> stop-on-failure

usually I prefer to run the full test suite to give an over all test status on a PR.
I'm not too sure if we should walk this way - would need agreement on a bigger scale from my poc

DeepDiver1975 (approved):

patrickjahns:
@ownclouders rebase

PR #30581, opened 2018-02-22
IE on Travis

phil-davis:
with UI test suites split up more


phil-davis:
Looking good so far. The annoying error is:
```
Payload received from webdriver is not valid JSON
```
I will try later ``iedriverVersion`` settings.

PR #30582, opened 2018-02-22
add group_concat to expression builder

butonic:
would allow reducing queries related to fetching account terms

see http://www.sqlines.com/mysql/functions/group_concat

- we only store 191 chars per term, which allows at least 5 search terms on mysql (group_concat max length is 1024 by default), 20 search terms on oracle, and we don't know how many on sqlite or postgres

@tomneedham we could add a fallback that detects if the rterms are longer than 1000 chars and do a subselect in that case. but i wouldn't bother 'til the first bug report.

### DIFF
--- a/lib/private/DB/QueryBuilder/ExpressionBuilder/ExpressionBuilder.php
+++ b/lib/private/DB/QueryBuilder/ExpressionBuilder/ExpressionBuilder.php
@@ -378,4 +378,19 @@ class ExpressionBuilder implements IExpressionBuilder {
 		$column = $this->helper->quoteColumnName($column);
 		return new QueryFunction("LENGTH({$column})");
 	}
+
+	/**
+	 * Returns a query function to concatenate values within each group defined by GROUP BY clause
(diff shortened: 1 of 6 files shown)


mrow4a:
Wow, this one I have not seen yet :D

butonic:
ńo idea wyh codecov failed to ge tall results ... anyway, I think we may be inviting fragile queries with this. Originally, I planned to use it to improve performance of the `GET http://cloud.local/ocs/v1.php/apps/files_sharing/api/v1/sharees?format=json&search=zombie-&perPage=200&itemType=file` endpoint.

DeepDiver1975:
> no idea wyh codecov failed to ge tall results ...

we do not collect codecov information on PG and oracle - all pg and oracle specific code is not touched by tests and is reported as negative coverage with respect to this test.

This is okay for now.

tomneedham:
> 20 terms/user will explode in others ways first probably - and I would consider this outside of the normal use case anyway

PR #30584, opened 2018-02-23
 Split UI test suites into smaller pieces

phil-davis:
## Related Issue
https://github.com/owncloud/QA/issues/517

## Motivation and Context
Some test suites are getting longer to run. Travis+SauceLabs is much slower than drone and has a 50 minute limit for jobs. We need Travis+SauceLabs to run the Microsoft Edge and IE11 browser tests in the daily Travis-cron task.

Sometimes jobs are reaching the time-limit when actually they are nearly finished.

## How Has This Been Tested?
Make other PRs to trigger UI test suite runs on Travis and check that jobs finish "enough" inside the time limit. I will make a PR on top of this one to demonstrate with IE.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Test refactoring

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30587, opened 2018-02-23
[stable10] Split UI test suites into smaller pieces

phil-davis:
Backport #30584 
Backport #30588 (which is a fix for a problem I introduced in #30584 )


individual-it (approved):

phil-davis:
drone did not trigger - it has died recently :(

PR #30588, opened 2018-02-23
Fix contexts needed for manageUsersGroups UI test suite

phil-davis:
I was too severe removing contexts form the ``manageUsersGroups`` suite in PR #30584 
The "redirect after login" tests use ``PersonalGeneralSettings`` page as a test destination.

### DIFF
--- a/tests/ui/config/behat.yml
+++ b/tests/ui/config/behat.yml
@@ -136,6 +136,7 @@ default:
         - LoginContext:
         - UsersContext:
         - FilesContext:
+        - PersonalGeneralSettingsContext:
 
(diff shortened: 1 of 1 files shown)


phil-davis:
I noticed this when running some Travis-based browser tests - they failed on ``manageUsersGroups`` suite saying "suite has undefined steps".

I have verified with local testing that the extra context added here fixes the issue.

- [ ] look at the drone runs from #30584 to see why/how they passed

phil-davis:
Backport to stable10 is included in #30587

individual-it (approved):

PR #30617, opened 2018-02-26
[WIP] properly name account uuid and username columns

butonic:
- [ ] add migration step
- [ ] chaneg ocurrences of setUserId and getUserId in the code to use the UUID or the user name. 
- [ ] generate a uuid if it isn't provided by the backend (or always generate one but allow the backend to overwrite it)

for syncing uuid should be ok ... but then again other backends may not provide a uuid ... they need to sync based on he login ... hmmm

### DIFF
--- a/lib/private/User/Account.php
+++ b/lib/private/User/Account.php
@@ -28,7 +28,9 @@ use OCP\UserInterface;
 /**
  * Class Account
  *
- * @method string getUserId()
+ * @method string getUuid()
+ * @method void setUuid(string $uuid)
+ * @method string getUserName()
  * @method string getDisplayName()
  * @method void setDisplayName(string $displayName)
  * @method string getEmail()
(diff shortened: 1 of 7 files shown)


lib/private/User/Account.php
```diff
@@ -54,8 +56,8 @@ class Account extends Entity {
 	const STATE_DELETED = 3;
 
 	protected $email;
-	protected $userId;
-	protected $lowerUserId;
+	protected $uuid;
+	protected $userName;
```
DeepDiver1975: not login name?

butonic: it is more than just a login name. it is shown to other users. we could eg. use it in the sharing dialog as it identifies a user. so it is more a user name than a login name.

felixboehm:
@butonic working on this? please add milestone.
Is there an overview for current tasks and concept related to user management?

butonic:
superseded by https://github.com/owncloud/core/pull/31335

PR #30473, opened 2018-02-13
[stable9.1] Catch session unavailable exception

tomneedham:
Backport of https://github.com/owncloud/core/pull/30141 to stable9.1

PR #30619, opened 2018-02-26
Catch session unavailable exception

DeepDiver1975:
port of https://github.com/owncloud/core/pull/30473

### DIFF
--- a/tests/lib/Session/CryptoSessionDataTest.php
+++ b/tests/lib/Session/CryptoSessionDataTest.php
@@ -1,6 +1,7 @@
 <?php
 /**
  * @author Joas Schilling <nickvergessen@owncloud.com>
+ * @author Tom Needham <tom@owncloud.com>
  *
  * @copyright Copyright (c) 2018, ownCloud GmbH
  * @license AGPL-3.0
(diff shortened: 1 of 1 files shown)


tests/lib/Session/CryptoSessionDataTest.php
```diff
@@ -34,7 +36,7 @@ class CryptoSessionDataTest extends Session {
 	protected function setUp() {
 		parent::setUp();
 
-		$this->wrappedSession = new \OC\Session\Memory($this->getUniqueID());
+		$this->wrappedSession = new \OC\Session\Memory(static::getUniqueID());
```
PVince81: never seen this `static::` approach before.

should this be `self::` ?

DeepDiver1975: Maybe this helps to understand when to use what - https://www.leaseweb.com/labs/2014/04/static-versus-self-php/

In this case both can be used without changing the logic

PVince81: I'm only asking when I see patterns we did not use much before. Fine to start using this now.

PVince81 (approved):
:+1:

PR #30631, opened 2018-02-27
Investigate password setting test failures

phil-davis:
Try to see what is the problem with the personal password change UI tests.


phil-davis:
see #30633 for improved version

Issue #29695, opened 2017-11-27
Improve public link share mail address input field


pmaier1:
The current mail address input field for public link shares is too small and does not work well when entering more than one address. It would be great to have it similar to the following screenshot:

![link_addresses](https://user-images.githubusercontent.com/21193500/33278324-7b0f7fc0-d39b-11e7-9319-532afea95462.png)

**Requirements**
- input field is adaptive (auto-scale)
- mail addresses become "fixed" by hitting space as in the screenshot
- mail addresses can be "unfixed"/deleted letter-wise with backspace and have a "x" to delete them at once

A mockup would be great!

@felixheidecke @PVince81

PVince81:
@felixheidecke was this done with https://github.com/owncloud/core/pull/29935 ?

felixheidecke:
> @felixheidecke was this done with #29935 ?

Nope, this was not done so far. @pmaier1 are we sure this is necessary? Can we move this to the backlog? Other things seem more urgent.

pmaier1:
> Nope, this was not done so far. @pmaier1 are we sure this is necessary? Can we move this to the backlog? Other things seem more urgent.

Yes, it's important. This is a p1 item in the milestone for the next release. Also it's a prereq for https://github.com/owncloud/core/issues/29938. Let's speak if there are concerns.

PVince81:
let's use select2 which is already included with ownCloud core, see http://select2.github.io/select2/ and search for "Tagging Support" as agreed when discussing with @felixheidecke

felixheidecke:
select2 is not suitable for this matter. If we really want to implement this behavior, we do need jet another external library.

PVince81:
also needs to support autocomplete, see https://github.com/owncloud/core/pull/30645#issuecomment-369860841

felixheidecke:
[PR](https://github.com/owncloud/core/pull/30645) is ready to review

felixheidecke:
Merged

PR #30645, opened 2018-02-28
Improve UX when entering email addresses for share link

lordelix:
## Description

- input field is adaptive (auto-scale)
- mail addresses become "fixed" by hitting space as in the screenshot
- mail addresses can be "unfixed"/deleted letter-wise with backspace and have a "x" to delete them at once

## Related Issue

https://github.com/owncloud/core/issues/29695
Fixes: https://github.com/owncloud/core/pull/30645

## Types of changes

## Checklist:

### DIFF
--- a/core/css/share.css
+++ b/core/css/share.css
@@ -27,16 +27,18 @@
 	margin-top: 1px;
 }
 
-.shareTabView .shareWithLoading,
-.shareTabView .linkShare .icon-loading-small {
+.shareTabView .linkShare .icon-loading-small,
+.shareTabView .shareWithLoading {
 	display: inline-block !important;
 	padding-left: 10px;
 }
+
 .shareTabView .shareWithLoading {
 	position: relative;
 	right: 70px;
 	top: 2px;
 }
+
 .shareTabView .icon-loading-small.hidden {
 	display: none !important;
 }
@@ -79,14 +81,14 @@
 	flex-shrink: 0;
 }
 
-.share-autocomplete-item .autocomplete-item-additional-info,
-#shareWithList .user-additional-info {
+#shareWithList .user-additional-info,
+.share-autocomplete-item .autocomplete-item-additional-info {
 	color: #888;
 }
 
 #shareWithList {
-	list-style-type:none;
-	padding:8px;
+	list-style-type: none;
+	padding: 8px;
 }
 
 #shareWithList li {
@@ -102,15 +104,18 @@
 	display: inline-block;
 }
 
-#shareWithList .unshare img, #shareWithList .showCruds img {
-	vertical-align:text-bottom; /* properly align icons */
+#shareWithList .showCruds img,
+#shareWithList .unshare img {
+	vertical-align: text-bottom;
+	/* properly align icons */
 }
 
-#shareWithList label input[type=checkbox]{
+#shareWithList label input[type=checkbox] {
 	margin-left: 0;
 	position: relative;
 }
-#shareWithList .username{
+
+#shareWithList .username {
 	padding-right: 8px;
 	white-space: nowrap;
 	text-overflow: ellipsis;
@@ -119,36 +124,38 @@
 	overflow: hidden;
 	vertical-align: middle;
 }
-#shareWithList li label{
+
+#shareWithList li label {
 	margin-right: 8px;
 }
+
 .shareTabView label {
-	font-weight:400;
+	font-weight: 400;
 	white-space: nowrap;
 }
 
 .shareTabView input[type="checkbox"] {
-	margin:0 3px 0 8px;
+	margin: 0 3px 0 8px;
 	vertical-align: middle;
 }
 
 a.showCruds {
-	display:inline;
-	opacity:.5;
+	display: inline;
+	opacity: 0.5;
 }
 
 a.unshare {
-	display:inline;
-	float:right;
-	opacity:.5;
+	display: inline;
+	float: right;
+	opacity: 0.5;
 	padding: 10px;
 	margin-top: -5px;
 	margin-right: -10px;
 }
 
 #link {
-	border-top:1px solid #ddd;
-	padding-top:8px;
+	border-top: 1px solid #ddd;
+	padding-top: 8px;
 }
 
 .shareTabView input[type="submit"] {
@@ -167,41 +174,48 @@ a.unshare {
 	box-shadow: 0 0 6px #f8b9b7;
 }
 
-.shareTabView .localShareView, .shareTabView .linkShareView {
+.shareTabView .linkShareView,
+.shareTabView .localShareView {
 	clear: left;
 }
 
 #link #showPassword img {
-	padding-left:5px;
-	width:12px;
+	padding-left: 5px;
+	width: 12px;
 }
 
-.fileowner,.reshare,#link label,
-#expiration label {
+#expiration label,
+#link label,
+.fileowner,
+.reshare {
 	display: inline-block;
 	padding: 6px 4px;
 }
 
-a.showCruds:hover,a.unshare:hover {
-	opacity:1;
+a.showCruds:hover,
+a.unshare:hover {
+	opacity: 1;
+}
+#defaultExpireMessage,
+/* fix expire message going out of box */
+.reshare {
+	/* fix shared by text going out of box */
+	white-space: normal;
 }
 
-#defaultExpireMessage, /* fix expire message going out of box */
-.reshare { /* fix shared by text going out of box */
-	white-space:normal;
-}
-
-#defaultExpireMessage { /* show message on new line */
+#defaultExpireMessage {
+	/* show message on new line */
 	display: block;
 	padding-left: 4px;
 	/* TODO: style the dropdown in a proper way - border-box, etc. */
 	width: 90%;
 }
 
-.ui-autocomplete { /* limit dropdown height to 4 1/2 entries */
-	max-height:103px;
-	overflow-y:auto;
-	overflow-x:hidden;
+.ui-autocomplete {
+	/* limit dropdown height to 4 1/2 entries */
+	max-height: 103px;
+	overflow-y: auto;
+	overflow-x: hidden;
 }
 
 .notCreatable {
@@ -228,4 +242,22 @@ a.showCruds:hover,a.unshare:hover {
 	margin-top: -30px;
 	z-index: 1;
 	color: #999;
+}
+
+.emailPrivateLinkForm--dropDown > .select2-choices {
+	border-color: #ddd;
+	background-image: none;
+}
+
+.select2-search-choice {
+	text-transform: lowercase;
+}
+
+.select2-container-multi .select2-search-choice-close {
+	left: inherit;
+	right: 3px;
+}
+
+.select2-container-multi .select2-choices .select2-search-choice {
+	padding: 3px 18px 3px 5px;
 }
--- a/core/js/sharedialogmailview.js
+++ b/core/js/sharedialogmailview.js
@@ -17,7 +17,7 @@
 		'<form id="emailPrivateLink" class="emailPrivateLinkForm">' +
 		'  <span class="emailPrivateLinkForm--send-indicator success-message-global absolute-center hidden">{{sending}}</span>' +
 		'  <label class="public-link-modal--label" for="emailPrivateLinkField-{{cid}}">{{mailLabel}}</label>' +
-		'  <input class="public-link-modal--input emailPrivateLinkForm--emailField" autocomplete="off" id="emailPrivateLinkField-{{cid}}" value="" placeholder="{{mailPlaceholder}}" type="email" />' +
+		'  <input class="emailPrivateLinkForm--emailField full-width" id="emailPrivateLinkField-{{cid}}" />' +
 		'  <div class="emailPrivateLinkForm--elements hidden">' +
 		'    {{#if userHasEmail}}' +
 		'    <label class="public-link-modal--bccSelf">' +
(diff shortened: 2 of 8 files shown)


core/js/sharedialogmailview.js
```diff
@@ -169,8 +178,16 @@
...
 			this.delegateEvents();
-
 			return this;
+
+		},
+
+		afterRender: function () {
+			this.$el.find('.emailPrivateLinkForm--emailField').select2({
+				// data:[{id:0,text:'enhancement'},{id:1,text:'bug'},{id:2,text:'duplicate'},{id:3,text:'invalid'},{id:4,text:'wontfix'}],
+				tags:["hi"],
```
felixheidecke: select2 lib throws `Uncaught TypeError: tags is not a function` in select2.js line 1210:
https://cdnjs.cloudflare.com/ajax/libs/select2/3.5.3/select2.js

felixheidecke: I can't find the problem @ all. :-|
Help me @PVince81 Wan Kenobi you're my only hope.

PVince81:
this is indeed very weird. select2 is using the function `tags` which I see is defined on line 537 but somehow is not accessible within the scope where the error occurs :confused:

PVince81:
If I add a row `var wtf = tags;` after the function definition of `tags` and use `wtf()`, the function is defined... It almost looks like Javascript somehow refuses to bind this external function into the internal closure... Has anything changed in Javascript recently ??

PVince81:
This does work correctly:
```js
(function () {
	function inner() {
		console.log('inner called and wants its JS logic back');
	}

	function another() {
		console.log('another');
		inner();
	}

	window.setTimeout(another, 5000);
})();
```

not sure what is so different in the select2 code that the `tags` function is suddenly not defined any more.

PVince81:
I added some console log statements and found out that "tags" as still defined in `prepareOpts` for multi but then suddenly in `prepareOpts` in the abstract class it's not defined any more?!

PVince81:
Maybe the bug was always there but we never bumped into it through our other use cases...

Next up: try and run select2.js completely alone to see if there are side effects from other libraries...

PVince81:
Put this in the owncloud root and open it through the OC web server:
```html
<html>
<head>
	<script src="core/vendor/jquery/dist/jquery.min.js"></script>
	<script src="core/vendor/select2/select2.js"></script>
</head>
<body>

	<div>Testing how far away from logic we are with select2:</div>

	<input class="public-link-modal--input emailPrivateLinkForm--emailField" id="emailPrivateLinkField-123" value="" type="hidden" />

	<script type="text/javascript">
		function start() {
			console.log('select2 init');
			$('.emailPrivateLinkForm--emailField').select2({
				// data:[{id:0,text:'enhancement'},{id:1,text:'bug'},{id:2,text:'duplicate'},{id:3,text:'invalid'},{id:4,text:'wontfix'}],
				tags:["hi"],
				tokenSeparators:[","]
			});
		}

		$(document).ready(start);
	</script>

</body>
</html>
```

The same problem happens... maybe we hit a select2 bug.

Still, I don't understand why that "tags" variable is suddenly not set any more.

Instead of trying to understand why this is happening, maybe let's try to properly use select2 by setting more options like in the example.

PVince81:
Ok, looking at the debugger I understand the error. We're back to the world of logic.

Inside the abstract `prepareOpts()` function, there is a statement that declares `var tags = ...`. The Javascript interpreter will hoist this variable and make it a local variable, so this will hide the "tags" from the closure. This is a bug in select2. However since we're dealing with an EOL version there is no point in raising a bug upstream...

Now... Tag support does work in other cases, so let's find out whether our usage of select2 is passing the correct arguments.

PVince81:
I can't seem to make it work with tags even though I'm pretty close to the example in http://select2.github.io/select2/

The only difference I see is that OC has jQuery 2.1.4 while the example page has jQuery 1.7.1...

core/js/sharedialogmailview.js
jvillafanez: It collides with the "tags" funcion of the select2 component. You might want to try either setting it via html element with "data-tags" attribute, or try to set it after creating the elements via "data" function.

PVince81:
So... I deployed the github example page and reduced it to the minimum. It works on the github page of select2 but not in my local HTML page. Then I copied over select2.js and select2.css files from the github repo to the local folder and suddenly it works !

PVince81:
Ohhhhhhh... it seems the version deployed on the example page contains a fix that uses "tagData" instead of "tags". Sneaky!

Let's try upgrading to 3.5.4...

PVince81:
raised https://github.com/select2/select2/issues/5239

PVince81:
pushed select2 update to 3.5.4 which fixes the issue

@felixheidecke please carry on.

core/js/sharedialogmailview.js
PVince81: @jvillafanez no, this syntax is correct. Please read my comments below. It's a select2 bug that was fixed in the newest version

felixheidecke:
@PVince81 … much thankyou. I knew it was an older version but I hesitated updating it. Never knew, what would happen to the system tags functionality.

felixheidecke:
@PVince81 … many thanks to you. I knew select2 4.0.5 exists did want to risk the major jump. I really didn't see version lurking in the shadows 3.5.4 there. Thank you for your effort 👍

PVince81:
It seems the autocomplete ajax request has gone missing on stable10: https://github.com/owncloud/core/issues/30669.

Instead of bringing it back as it was before, please wire select2 to a function that makes the server call. It should be possible to let select2 do an ajax call to retrieve a list of "tags" matching a given pattern.

PVince81:
@felixheidecke please set the labels as we usually do

felixheidecke:
- [x] add share link to contact email
- [ ] make mail sending work

I added back the "share link to contact email" functionality. Please have a look if this is what we intended.

core/js/sharedialogmailview.js
```diff
@@ -169,8 +179,51 @@
...
+
+		},
+
+		afterRender: function () {
+			var _this = this;
+
+			this.$el.find('.emailPrivateLinkForm--emailField').select2({
+				containerCssClass: 'emailPrivateLinkForm--dropDown',
+				tags: [],
+				tokenSeparators:[","],
```
PVince81: that doesn't seem to work currently, not sure why: try typing email addresses and hitting the comma

core/js/sharedialogmailview.js
```diff
@@ -169,8 +179,51 @@
...
+					dataType: 'json',
+					quietMillis: 250,
+					data: function (term) {
+						return {
+							search: term
+						};
+					},
+					results: function (data, page, query) {
+
+						if (data.status != 'success')
```
PVince81: did we have a minimum char limit in the old implementation or did we start searching at one char ?

PVince81:
Should we change the pending search message to "Searching contacts" ?

core/js/sharedialogmailview.js
felixheidecke: Try typing email addresses and hitting enter. :-) There is no more comma to be used.

core/js/sharedialogmailview.js
felixheidecke: Good point. I'll start @ 3 since this is the minimum username length.

felixheidecke:
> Should we change the pending search message to "Searching contacts" ?

Can we assume that users know about contacts? I certainly didn't :-)
If contacts aren't used, it will always tell us that nothing was found. We should require at least one contact for the search function to be used. Don't you agree?

felixheidecke:
Added an ani-GIF https://github.com/owncloud/core/pull/30645#issue-171873127

PVince81:
> We should require at least one contact for the search function to be used. Don't you agree?

We could do that as an extension (separate PR): would require the ContactsManager to be able to tell the UI that there is nothing to search so don't bother.

felixheidecke:
Ready 4 testing.

core/css/share.css
```diff
@@ -167,41 +174,48 @@ a.unshare {
...
 	display: block;
 	padding-left: 4px;
 	/* TODO: style the dropdown in a proper way - border-box, etc. */
 	width: 90%;
 }
 
-.ui-autocomplete { /* limit dropdown height to 4 1/2 entries */
-	max-height:103px;
-	overflow-y:auto;
-	overflow-x:hidden;
```
PVince81: I predict conflicts during backport... that's what automatic formatting does. can you disable it in the future ?

if you want to reformat files let's do that in a separate PR only for formatting, because usually reformatting can make small changes invisible for the reviewer...

core/js/sharedialogmailview.js
```diff
@@ -158,6 +161,9 @@
 		},
 
 		render: function() {
+            // make sure this is empty
+            this.addresses = [];
```
PVince81: indent is wrong

core/js/sharedialogmailview.js
```diff
@@ -169,8 +175,60 @@
 			}));
 
 			this.delegateEvents();
-
 			return this;
+
+		},
+
+		afterRender: function () {
```
PVince81: make this a private method ?

PVince81:
without contacts app enabled or anything, just plain checking out the branch, I get this after typing:
image

core/css/share.css
felixheidecke: Did it break? Told Atom to do it on save. :-|

felixheidecke:
> without contacts app enabled or anything, just plain checking out the branch, I get this after typing:

funny. I disabled and later removed the app but it still works. Did you `make` ??

PVince81:
Ok, maybe I missed shift-F5.

Now it works with some little bugs:
- [ ] see if there is a way to accept the new value before waiting for the ajax call to finish (in some setups the call might be slower)
- [ ] hit enter with an invalid email address repeatedly, the dialog box shrinks horizontally with every hit :stuck_out_tongue_winking_eye:

PVince81:
- should the cross rather be on the right instead of the left ?

felixheidecke:
>  hit enter with an invalid email address repeatedly, the dialog box shrinks horizontally with every hit.

Can't reproduce. Which browser? Or did I fix it unintentionally already?

PVince81:
the hit enter bug seems to be gone now

felixheidecke:
> see if there is a way to accept the new value before waiting

@PVince81 I tried a couple of approaches but it does not seem to work with preselecting and such. To bad, will have to wait I guess :-|

PVince81:
> @PVince81 I tried a couple of approaches but it does not seem to work with preselecting and such. To bad, will have to wait I guess :-|

As long as you can hit enter before the query is finished it's fine. We don't want the user to need to wait.

I expect that in most cases this autocomplete will be disabled anyway as part of another PR.

felixheidecke:
> As long as you can hit enter before the query is finished it's fine.

I wrote the query function myself so that the search input get's returned immediately. Its not "as nice" as before (with the loader icon) but it get's the job done.

PVince81:
Code looks fine.

Can you add some unit tests ?

Here's how you can test the query function given to select2 using a stub: https://github.com/owncloud/core/blob/v10.0.7/apps/systemtags/tests/js/systemtagsfilelistSpec.js#L106

PVince81:
and when you're done, rebase + squash onto master

pmaier1:
Looks very nice :+1:  There are implications with https://github.com/owncloud/enterprise/issues/2426 which should be considered.

PVince81:
implications in question to be discussed in https://github.com/owncloud/core/issues/30670

felixheidecke:
@PVince81 we good here?

PVince81 (approved):
Nice! :+1:

phil-davis:
@ownclouders rebase

ownclouders:

ownclouders:

phil-davis:
There were weird drone errors not being able to load... stuff. A rebase has got it all up-to-date with current CI infrastructure.

PVince81:
@felixheidecke please also cover at least the query function, see https://codecov.io/gh/owncloud/core/compare/4037c9aeabf0432a40ec82128f06b8b8283a4911...8893c8347cb6d54f22a9ae1a9a7b7180d4dad6fd/src/core/js/sharedialogmailview.js#L192

PVince81:
@felixheidecke correction: please remove the autocomplete completely. No need to waste more time on this. I'll remove the backend part here: https://github.com/owncloud/core/issues/30670

felixheidecke:
I *just* rebased the master onto this branch:

rebase_master

@phil-davis If drone ci keeps failing then I don't know whats going on. :-(

phil-davis:
It is waiting to run now, so we will see.

PVince81:
please confirm if this has been retested and all cases work then we can merge

@phil-davis I wonder if we can get this into an acceptance test ? do you have email parsing there already ?

phil-davis:
@PVince81 there is a step:
```
@When the user creates a new public link for the file/folder :name using the webUI with
```
that allows putting in the email address.

Is is not used much currently, but easily could be.

If we want to test the behavior withing the field, e.g. for adding a few addresses then deleting one in the middle..., then we will need to look into bonus selenium trickery to be able to "click" the cursor into a particular place in the field before pressing backspace etc. @individual-it - needs a bit of research next week, but in general it would be useful to be able to control cursor-placement in a text field.

PVince81:
@phil-davis I think for now it might be enough to test with two email address and validating. In general I expect select2 to already work correctly.

phil-davis:
Created issue #30892 and assigned myself, so I do not lose track of adding some UI acceptance test functionality for this.

PVince81:
> please confirm if this has been retested and all cases work then we can merge

@felixheidecke ^

felixheidecke:
> please confirm if this has been retested and all cases work then we can merge

I can confirm that I works on whatever testing env. I have @ my disposal. Latest FF, Latest Chrome and IE11 on my VM.

PVince81:
stable10: https://github.com/owncloud/core/pull/30945

PR #30198, opened 2018-01-19
Default HTTPS for fed sharing & config option for allowing fallback

tomneedham:
## Description
Only uses HTTPS for federated share communication - unless admins explicitly allow HTTP (useful for testing as well)

## Related Issue
Potential for users to setup shares via HTTP with unsecure servers without the admin knowing.

## Motivation and Context
MITM

## How Has This Been Tested?
not yet

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30646, opened 2018-02-28
[stable10] Flag to enable http fallback when creating fed shares

tomneedham:
Backport of https://github.com/owncloud/core/pull/30198 to stable10

cc @pmaier1 for release notes / changelog. We now default to now allowing http communication for federated sharing. However it can be enabled with a config flag

`'sharing.federation.allowHttpFallback' => false,`


PVince81 (approved):
:+1:

PR #30648, opened 2018-02-28
Allow regular users to change their CORS domains

### LINKED ISSUE #30647: Users cannot add domains under CORS settings
In 10.0.7 under "Personal Settings->Security" options for CORS are shown. 

image

However, if a normal user attempts to add a domain the following error is displayed:
```
Access forbidden
Logged in user must be an admin
```
 
we should fix it.

@Emil-G @PVince81

PVince81:
## Description
See subject

## Related Issue
Fixes https://github.com/owncloud/core/issues/30647

## Types of changes

## Checklist:

### DIFF
--- a/settings/Controller/CorsController.php
+++ b/settings/Controller/CorsController.php
@@ -86,6 +86,9 @@ class CorsController extends Controller {
 	/**
 	 * Gets all White-listed domains
 	 *
+	 * @NoAdminRequired
+	 * @NoSubadminRequired
+	 *
 	 * @return JSONResponse All the White-listed domains
 	 */
 	public function getDomains() {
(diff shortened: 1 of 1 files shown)


DeepDiver1975 (approved):
Tested - works :+1:

PVince81:
stable10: https://github.com/owncloud/core/pull/30649

pako81:
works. But should not the domains then appear in the list of the white-listed domains?

pako81:
ah ok it looks like there is no error showed if you enter an invalid URL (i.e. missing the http part)

PVince81:
> ah ok, it looks like there is no error showed if you enter an invalid URL (i.e. missing the http part)

yes, that's another bug... will take a bit more time to fix as there is no way to return an error here... need to port the code to use ajax calls instead of full page request

Issue #30199, opened 2018-01-19
Update 8.2.11 to 10.0.4 fails with orphaned reshare


PVince81:
### Steps
1. Setup OC 8.2.11
1. Create two users "user1" and "user2"
1. Login as admin
1. Create folder "test"
1. Share "test" with "user1"
1. Login as "user1"
1. Reshare "test" with "user2"
1. Login as "user2"
1. Delete the main share in oc_share manually
1. Upgrade to 10.0.4 with `occ upgrade`

```
2018-01-19T19:32:33+00:00 Doctrine\DBAL\Exception\NotNullConstraintViolationException: An exception occurred while executing 'UPDATE `oc_share` SET `uid_owner` = ?, `uid_initiator` = ?, `parent` = ? WHERE `id` = ?' with params [null, "user1", null, 2]:

SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'uid_owner' cannot be null
2018-01-19T19:32:33+00:00 Update failed
```

### Versions
Update from 8.2.11 to 10.0.4 with PHP 5.6.

PVince81:
There are two things that need to be fixed here:

1) Introduce repair step to delete these orphaned shares with non-existing parent (if we haven't got one already)
and
2) Fix files_sharing Migration code to skip these broken shares

PR #30653, opened 2018-03-01
Repair step to fix orphan reshares

sharidas:
Its being reported that shares are orphaned.
And in such state when an upgrade operation is
performed results in failure. This repair
step fixes the orphan shares.

## Description

This change would help to fix the failures caused during upgrade operation, when orphan reshares are residing in oc_share.

## Related Issue

https://github.com/owncloud/core/issues/30199

## Motivation and Context

This change would help to fix the failures caused during upgrade operation, when orphan reshares are residing in oc_share. This change tries to remove the orphan shares fro the table which doesn't have parent.

## How Has This Been Tested?

- Verified by creating the same scenario explained in 8.2.11 upgrade to 10.0.6.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Repair.php
+++ b/lib/private/Repair.php
@@ -37,6 +37,7 @@ use OC\Repair\DropOldJobs;
 use OC\Repair\OldGroupMembershipShares;
 use OC\Repair\RemoveGetETagEntries;
 use OC\Repair\RemoveRootShares;
+use OC\Repair\RepairOrphanedSubshare;
 use OC\Repair\RepairSubShares;
 use OC\Repair\SharePropagation;
 use OC\Repair\SqliteAutoincrement;
@@ -191,6 +192,7 @@ class Repair implements IOutput{
 			new InnoDB(),
 			new Collation(\OC::$server->getConfig(), $connection),
 			new SqliteAutoincrement($connection),
+			new RepairOrphanedSubshare($connection),
 			new SearchLuceneTables(),
 			new Apps(\OC::$server->getAppManager(), \OC::$server->getEventDispatcher(), \OC::$server->getConfig(), new \OC_Defaults()),
 		];
@@ -249,7 +251,7 @@ class Repair implements IOutput{
 	}
 
 	/**
-	 * @param int $max
+	 * emit signal
 	 */
 	public function finishProgress() {
 		// for now just emit as we did in the past
--- a/lib/private/Repair/RepairOrphanedSubshare.php
+++ b/lib/private/Repair/RepairOrphanedSubshare.php
@@ -1,0 +1,107 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Repair;
+
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\IRepairStep;
+
+class RepairOrphanedSubshare implements IRepairStep {
+
+	/** @var IDBConnection  */
+	private $connection;
+
+	/** @var  IQueryBuilder */
+	private $missingParents;
+
+	/** @var  IQueryBuilder */
+	private $deleteOrphanReshares;
+
+	/**
+	 * RepairOrphanedSubshare constructor.
+	 *
+	 * @param IDBConnection $connection
+	 */
+	public function __construct(IDBConnection $connection) {
+		$this->connection = $connection;
+
+		//This delete query deletes orphan shares whose parents are missing
+		$this->deleteOrphanReshares = $this->connection->getQueryBuilder();
+		$this->deleteOrphanReshares
+			->delete('share')
+			->where(
+				$this->deleteOrphanReshares->expr()->eq('parent',
+					$this->deleteOrphanReshares->createParameter('parentId')));
+
+
+		//Set the query to get the parent id's missing from the table.
+		$this->missingParents = $this->connection->getQueryBuilder();
+		$qb = $this->connection->getQueryBuilder();
+
+		$qb->select('id')
+			->from('share');
+		$this->missingParents->select('parent')
+			->from('share');
+		$this->missingParents->where($this->missingParents->expr()->isNotNull('parent'));
+		$this->missingParents->andWhere($this->missingParents->expr()->notIn('parent', $this->missingParents->createFunction($qb->getSQL())));
+		$this->missingParents->groupBy('parent');
+	}
+
+	/**
+	 * Returns the step's name
+	 *
+	 * @return string
+	 */
+	public function getName() {
+		return 'Repair orphaned reshare';
+	}
+
+	/**
+	 * Run repair step.
+	 * Must throw exception on error.
+	 *
+	 * @param IOutput $output
+	 * @throws \Exception in case of failure
+	 */
+	public function run(IOutput $output) {
+		$pageLimit = 1000;
+		$paginationOffset = 0;
+		$deleteParents = [];
+
+		/** A one time call to get missing parents from oc_share table */
+		do {
+			$this->missingParents->setMaxResults($pageLimit);
+			$this->missingParents->setFirstResult($paginationOffset);
+			$results = $this->missingParents->execute();
+			$rows = $results->fetchAll();
+			$results->closeCursor();
+			if (count($rows) > 0) {
+				$paginationOffset += $pageLimit;
+				foreach ($rows as $row) {
+					$this->deleteOrphanReshares->setParameter('parentId', $row['parent'])->execute();
+				}
+			}
+
+		} while (empty($rows));
+	}
+}
--- a/tests/lib/Repair/RepairOrphanedSubshareTest.php
+++ b/tests/lib/Repair/RepairOrphanedSubshareTest.php
@@ -1,0 +1,290 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+
+namespace Test\Repair;
+
+use OC\Repair\RepairOrphanedSubshare;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\IRepairStep;
+use Test\TestCase;
+use Test\Traits\UserTrait;
+
+/**
+ * Class RepairOrphanedSubshareTest
+ *
+ * @group  DB
+ * @see RepairOrphanedSubshare
+ * @package Test\Repair
+ */
+class RepairOrphanedSubshareTest extends TestCase {
+	use UserTrait;
+
+	/** @var  IDBConnection */
+	private $connection;
+
+	/** @var  IRepairStep */
+	private $repair;
+
+	protected function setUp() {
+		parent::setUp();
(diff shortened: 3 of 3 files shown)


lib/private/Repair.php
```diff
@@ -191,6 +192,7 @@ public static function getBeforeUpgradeRepairSteps() {
 			new InnoDB(),
 			new Collation(\OC::$server->getConfig(), $connection),
 			new SqliteAutoincrement($connection),
+			new RepairOrphanedSubshare($connection),
```
sharidas: Added the change here so that we trigger the repair step in the initial stage.

lib/private/Repair.php
```diff
@@ -249,7 +251,7 @@ public function advance($step = 1, $description = '') {
 	}
 
 	/**
-	 * @param int $max
+	 * emit signal
```
sharidas: This has nothing to do with the commit. But then it doesn't make sense to have a PHPdoc which is wrong.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+			->where($builder->expr()->eq('id', $parentId))->setMaxResults(2);//Where id = parent from above query or available parents
+		$this->getSharesWithParentsAvailable = $builder;
+	}
+
+	/**
+	 * This select query gives us unique parents from the table
+	 * @param $pageLimit
+	 * @param $paginationOffset
+	 */
+	private function setSelectGetAllParents($pageLimit, $paginationOffset) {
```
sharidas: The idea with this method is to get all the parents grouped and get the result.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+			->delete('share')
+			->where($builder->expr()->eq('parent', $parentId));
+		$this->deleteOrphanReshares = $builder;
+	}
+
+	/**
+	 * This select query is to get the result of parents === id
+	 * @param $parentId
+	 */
+	private function setSelectShareId($parentId) {
```
sharidas: This query will help us know if we have parents available in the table.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	public function getName() {
+		return 'Repair orphaned reshare';
+	}
+
+	/**
+	 * This delete query deletes orphan shares whose parents are missing
+	 * @param $parentId
+	 */
+	private function setDeleteOrphanSharesQuery($parentId) {
```
sharidas: This query will delete the orphan shares.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+				$lastResultCount++;
+				/**
+				 * Check if the query result is empty.
+				 * And if parent is not added to missingParents array, then
+				 * add it so that shares of this parent who became orphans can
+				 * be deleted later.
+				 */
+				if ((count($getIdRows) === 0) &&
+					(in_array($row['parent'], $missingParents, true) === false)) {
+					$missingParents[] = $row['parent'];
```
sharidas: If the parents are missing, that is if the share is orphan, then add to `$missingParents` array.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+				}
+			}
+
+			if (count($missingParents) > 0) {
+				foreach ($missingParents as $missingParent) {
+					$this->setDeleteOrphanSharesQuery($missingParent);
+					$deleteRows[] = $this->deleteOrphanReshares;
```
sharidas: Delete the rows if the `$missingParents` array has elements.

PVince81:
Why not add your github comments directly in the PHP code, if you think they are needed to understand the code ?

sharidas:
> Why not add your github comments directly in the PHP code, if you think they are needed to understand the code ?

Sure

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+	 */
+	public function __construct(IDBConnection $connection) {
+		$this->connection = $connection;
+	}
+
+	/**
+	 * Returns the step's name
+	 *
+	 * @return string
+	 * @since 9.1.0
```
PVince81: remove `since` as it's not a public API

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+	 * This select query is to get the result of parents === id
+	 * @param $parentId
+	 */
+	private function setSelectShareId($parentId) {
+		$builder = $this->connection->getQueryBuilder();
+		$builder
+			->select('id')
+			->from('share')
+			->where($builder->expr()->eq('id', $builder->createNamedParameter($parentId)));//Where id = parent from above query or available parents
+		$this->getSharesWithParentsAvailable = $builder;
```
PVince81: You have chosen to use prepared statements. In this case please generate the statement directly in the constructor and store it in `$this->getSharesWithParentsAvailable`. At creation time do not set any value but create a parameter with a name.

Then later when you use the query, set the parameter value with `setParameter()`.

You can find an example of such usage here: https://github.com/owncloud/core/blob/v10.0.7/lib/private/SystemTag/SystemTagManager.php#L77

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+
+		$this->availableParents = $builder;
+	}
+	/**
+	 * Run repair step.
+	 * Must throw exception on error.
+	 *
+	 * @param IOutput $output
+	 * @throws \Exception in case of failure
+	 * @since 9.1.0
```
PVince81: remove

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,160 @@
...
+			$paginationOffset += $pageLimit;
+			$lastResultCount = 0;
+
+			foreach ($rows as $row) {
+				if ($row['parent'] === null) {
+					$lastResultCount++;
+					continue;
+				}
+				$this->setSelectShareId($row['parent']);
+				$getIdQuery = $this->getSharesWithParentsAvailable->execute();
```
PVince81: instead of doing two queries, try using a subquery, here's an example: https://github.com/owncloud/core/blob/v10.0.7/lib/private/Repair/RepairMismatchFileCachePath.php#L186

if you're not familiar with SQL subqueries please let me know and I'll give you some pointers

sharidas: I will have to look more in to SQL subqueries as I am not familiar with them.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,127 @@
...
+		$qb = $this->connection->getQueryBuilder();
+
+		$qb->select('id')
+			->from('share');
+		$builder->selectDistinct('parent')
+			->from('share')
+			->where($builder->createFunction('parent NOT IN (' . $qb->getSQL() . ')'))
+			->setMaxResults($pageLimit)
+			->setFirstResult($paginationOffset);
+		$this->missingParents = $builder;
```
sharidas: Added subquery to get the missing parents from the table.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,128 @@
...
+		$this->deleteOrphanReshares = $builder;
+	}
+
+	/**
+	 * Set the query to get the parent id's missing from the table.
+	 * Extremely costly query, I guess
+	 * @param int $pageLimit
+	 * @param int $paginationOffset
+	 */
+	private function setParentsToExcludeFromShare($pageLimit, $paginationOffset) {
```
sharidas: This query looks bit expensive to me. Any pointers to reduce the load would be helpful.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,127 @@
...
+
+	/**
+	 * This delete query deletes orphan shares whose parents are missing
+	 * @param string $parentId
+	 */
+	private function setDeleteOrphanSharesQuery($parentId) {
+		$builder = $this->connection->getQueryBuilder();
+		$builder
+			->delete('share')
+			->where($builder->expr()->eq('parent', $builder->createNamedParameter($parentId)));
```
PVince81: please build the query without a value and use a real parameter name that you set later

see https://github.com/owncloud/core/blob/v10.0.7/lib/private/SystemTag/SystemTagManager.php#L172

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,126 @@
...
+		$builder = $this->connection->getQueryBuilder();
+		$qb = $this->connection->getQueryBuilder();
+
+		$qb->select('id')
+			->from('share');
+		$builder->select('parent')
+			->from('share');
+		$builder->where($builder->expr()->notIn('parent', $builder->createFunction($qb->getSQL())));
+		$builder->groupBy('parent');
+		$builder->setMaxResults($pageLimit);
```
PVince81: these need to be set by whatever code is using the query, not in the generic query

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,126 @@
...
+			$lastCount = 0;
+			$this->setParentsToExcludeFromShare($pageLimit, $paginationOffset);
+			$results = $this->missingParents->execute();
+			$rows = $results->fetchAll();
+			$results->closeCursor();
+			if (count($rows) > 0) {
+				$lastCount++;
+				$paginationOffset += $pageLimit;
+				foreach ($rows as $row) {
+					$this->setDeleteOrphanSharesQuery();
```
PVince81: no, the query is only created once in the constructor and never again

then the code just sets the parameters

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,114 @@
...
+		/** A one time call to get missing parents from oc_share table */
+		do {
+			$lastCount = 0;
+			$this->missingParents->setMaxResults($pageLimit);
+			$this->missingParents->setFirstResult($paginationOffset);
+			$results = $this->missingParents->execute();
+			$rows = $results->fetchAll();
+			$results->closeCursor();
+			if (count($rows) > 0) {
+				$lastCount++;
```
PVince81: you could simply use `empty($rows)` in the while condition, no need for this extra variable

additionally `$lastCount` seems to only ever count from 0 to 1, it doesn't count the results or pages, which could make it confusing for devs having a quick read. Better remove it.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,114 @@
...
+				$paginationOffset += $pageLimit;
+				foreach ($rows as $row) {
+					$deleteParents[] = $row['parent'];
```
PVince81: Should we delete the parents right away ?

If we do, then you don't need pagination in the `missingParents` query as the next iteration will not return the already deleted entries.

This proposed approach would likely save memory in case of lots of shares and avoid having a too long list stored in `$deleteParents`

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -0,0 +1,291 @@
...
+
+		$qb = $this->connection->getQueryBuilder();
+		$results = $qb->select('parent')
+			->from('share')
+			->groupBy('parent')->orderBy('parent')->setMaxResults(1000)->setFirstResult(1)
+			->execute()->fetchAll();
+		//Total 10 rows where there, we deleted 3 and the repair deleted another
+		//3 rows. So total 6 rows were deleted and 4 rows were remaining
+		//Hence 2 is the expected result.
+		//For pgsql the result comes as 3, because null is included
```
PVince81: can you exclude null somehow by adding a "is not null" in the query somehow and make it consistent across DBs ?

PVince81 (commented):
Looks better. A few more adjustments needed.

tests/lib/Repair/RepairOrphanedSubshareTest.php
sharidas: Done.

lib/private/Repair/RepairOrphanedSubshare.php
sharidas: The latest change does the deletion right away.

lib/private/Repair/RepairOrphanedSubshare.php
sharidas: Latest change removes `lastCount` var and uses `empty($rows)` in the while condition.

PVince81 (approved):
Looks good :+1: 

Please backport

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -0,0 +1,107 @@
...
+	/**
+	 * Run repair step.
+	 * Must throw exception on error.
+	 *
+	 * @param IOutput $output
+	 * @throws \Exception in case of failure
+	 */
+	public function run(IOutput $output) {
+		$pageLimit = 1000;
+		$paginationOffset = 0;
```
PVince81: pagination offset is always zero (sorry, missed this one)

PVince81:
@sharidas I might have merged too quickly :-(

the main query is doing pagination
once you start deleting entries, does it affect the results of the first query ?
for example say there are 2500 broken entries coming from the missingParents query
you get the first 1000, then delete children
if you rerun misisngParents query, it would only return 1500 entries, not 2500, so the pagination is shifting forward despite the already deleted entries
did I understand this correctly?

if yes then you need to remove pagination and only keep querying the first 1000 entries until there is nothing to fix

PVince81:
backport will need to include:
- this PR
- https://github.com/owncloud/core/pull/30989
- and 544668193d88a9e98b4531fdba682c0595f1f46d

sharidas:
Backport PR: https://github.com/owncloud/core/pull/31004

PR #30658, opened 2018-03-01
Adding Cross-Origin Resource Sharing to CORS

mmattel:
## Description

In Settings:Personal:Security: Adding the text ``Cross-Origin Resource Sharing`` next to ``CORS``

## Related Issue

none

## Motivation and Context

Making it more clear what CORS means

## How Has This Been Tested?

Opera and Firefox

## Screenshots (if appropriate):
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Beautyfying

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30664, opened 2018-03-01
Make PropagatorTest more deterministic in batch case (codecov random failure)

PVince81:
## Description
- Less randomness in batch case.
- Added extra batch entry to overwrite existing one to cover extra code
path

## Related Issue
Fixes random codecov issue observed in https://github.com/owncloud/core/pull/30658 and affecting the code path reaching in https://github.com/owncloud/core/blob/v10.0.7/lib/private/Files/Cache/Propagator.php#L135.

## Motivation and Context
With the above change the affected code path will always be triggered and should prevent random codecov failures.

## How Has This Been Tested?
Run the test manually with a breakpoint on the affected line.
Before this PR: it never stops there.
After this PR: it stops there.

## Types of changes

## Checklist:

### DIFF
--- a/tests/lib/Files/Cache/PropagatorTest.php
+++ b/tests/lib/Files/Cache/PropagatorTest.php
@@ -93,9 +93,20 @@ class PropagatorTest extends TestCase {
 		$oldInfos = $this->getFileInfos($paths);
 		$propagator = $this->storage->getPropagator();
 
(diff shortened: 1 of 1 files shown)


butonic (approved):

phil-davis:
Backport stable10 #30665 
(seems like this "randomness" should be removed from ``stable10`` test runs also)

PR #30674, opened 2018-03-02
correctly handle exceptions for getByUid, properly cache not existing user

butonic:
getByUID cannot return null. Need to properly handle the exceptions.

@tomneedham didn't we fix this already?

### DIFF
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -36,14 +36,12 @@ namespace OC\User;
 use OC\Cache\CappedMemoryCache;
 use OC\Hooks\PublicEmitter;
 use OCP\AppFramework\Db\DoesNotExistException;
+use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\Events\EventEmitterTrait;
 use OCP\ILogger;
 use OCP\IUser;
 use OCP\IUserManager;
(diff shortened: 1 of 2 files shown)


DeepDiver1975 (dismissed):
tests need to be adjusted ....

lib/private/User/Manager.php
```diff
@@ -160,20 +161,24 @@ public function clearBackends() {
 	 * @return \OC\User\User|null Either the user or null if the specified user does not exist
 	 */
 	public function get($uid) {
-		if (is_null($uid) || !is_string($uid)) {
+		if ($uid === null || !is_string($uid)) {
```
phil-davis: Note: Issue #30727 and PR #30732 related to the case when ``$uid`` is pased in as numeric, e.g. user is ``12345``

butonic: ouch ... caused by the automagic casting of string array keys to int if they look like a number ... I wonder when that is going to explode in other places. I changed the code to better reflect that. `userExists()` is not affected, because it uses `get()` but I cannot even begin where else this might cause headaches ...

tomneedham (approved):

tomneedham:
backport?

phil-davis:
Backport ``stable10`` #31446

PR #30675, opened 2018-03-02
Add JS coverage to codecov

PVince81:
Adding coverage for Javascript.

The coverage data is already generated by "test-javascript" by running the karma suite (no extra args required).


DeepDiver1975:
no idea why this is not properly working -> let's leave this to @patrickjahns

PR #30693, opened 2018-03-05
[POC] Direct write

DeepDiver1975:
## Description
We should not use the temp storage to store chunks (and regular files as well ?) but handle the stream directly to the objectstore implementation.

This should give us some performance boost.

Downside:
I don't see how any stream based operations can work in this scenario.
- preview
- encryption (well anyhow a bad idea ...)
- checksums
- workflow & firewall
- anti virus
- trash
- ransomware
- quota
- anything which uses our storage wrapper approach

## Motivation and Context
Bring some more speed to objectstore

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -35,12 +35,14 @@ namespace OCA\DAV\Connector\Sabre;
 
 use OC\AppFramework\Http\Request;
 use OC\Files\Filesystem;
+use OC\Files\ObjectStore\ObjectStoreStorage;
 use OC\Files\Storage\Storage;
 use OCA\DAV\Connector\Sabre\Exception\EntityTooLarge;
 use OCA\DAV\Connector\Sabre\Exception\FileLocked;
 use OCA\DAV\Connector\Sabre\Exception\Forbidden as DAVForbiddenException;
 use OCA\DAV\Connector\Sabre\Exception\UnsupportedMediaType;
 use OCA\DAV\Files\IFileNode;
+use OCA\DAV\Upload\FutureFile;
 use OCP\Encryption\Exceptions\GenericEncryptionException;
 use OCP\Events\EventEmitterTrait;
 use OCP\Files\EntityTooLargeException;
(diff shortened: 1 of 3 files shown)


butonic:
* preview
  - [ ] should be generated in a background job or on demand. data needs to be read from objectstore or the cache mount point (which might be a different objectstore)

* encryption (well anyhow a bad idea ...)
  - [ ] chunks don't even need to be assembled. Always use the assembly wrapper when reading these files.
    - [ ] look into prefetching a subsequent chunk.

* checksums
  - checksums on chunks can be calculated on the chunk stream. checksums for assembled files do not need to return a real checksum. they can return a checksum of the checksums like the etags of amazon s3. see https://stackoverflow.com/a/19896823 AFAICT the s3 implementation might assemble a file and then you can get a real hash. so a fallback for oc might be async calculation with background jobs. If we upload subsequentially we can even calcutale the sha1 by feeding it chunk stream after chunk stream.
  
* workflow & firewall
  - depends on what these apps want to do. block upload based on a stream of bytes? use assembly wrapper. Imo we need to split upload from making the file accessible. as soon as we start uploading we should show the file. but return a 503 on GET requests. when upload is complete firewall & workflow can have a look at the file. GET still returns 503. after any postprocessing is done we can return 200 OK for GET

* anti virus
  - postprocessing as for workflow and firewall IMO, currently we scan chunks ... which might not detect signatures at boundaries. then again files > 100 MB are typically not scanned anyway ...

* trash
  - free space constraints can be checked per chunk as well as for the whole file. we could send a header for the full file size when creating the upload folder with MKCOL if we are not doing thad yet (wh should). 

* ransomware
  - postprocessing -> see checksums 

* quota
  - see trashbin

* anything which uses our storage wrapper approach
  - postprocessing -> see workflow & firewall

so i think making the upload process acknowledge and reflect the postprocessing step is the solution. file upload is done, upload progressbar can vanish, postprocessing progressbar could start. we might even expose postprocessing steps like virus scan, workflow steps (if they have a `user visible` flag set). and we should move the progressbar to the row instead of the top bar. then we could even switch to parralel uploads ... with multiple progress bars ... wouldn't that be cool?

DeepDiver1975:
THX @butonic for the detailed input.

I'm quite sure that there are solutions for each an every problem. But the impact will be huge! We basically need to kill the storage wrappers - I'm not really keen on doing this ....

We need to research into php stream filters if they can help us in any way http://php.net/manual/en/function.stream-filter-register.php

PVince81:
conflicts

is this something we intend to finish for 10.0.9 ? if not => "triage" for future

DeepDiver1975:
doesn't work -> bomb

PR #30696, opened 2018-03-05
[Stable10] Fix nightly-qa tarball symlinks 

patrickjahns:
backport of #30678


phil-davis (approved):
You had to fixup an extra line to make this more like master. I guess some other change in master did not get backported at some point.

patrickjahns:
@phil-davis 
good catch - didn't notice at first - weird and makes me wonder how many other things diverge from master :'(

PR #30673, opened 2018-03-02
Move user management ui to its own app

DeepDiver1975:
## Description
The user managment ui is moved out of core. Core should more and more become a pure api server. This is one of many steps.

app is here: https://github.com/owncloud/user_management

## ToDos
- [x] remove all currently used controllers from settings -  move them as well if 
- [x] move the app to it's own repo
- [x] move acceptance tests ...

## How Has This Been Tested?
- manually

## Screenshots (if appropriate):
bildschirmfoto von 2018-03-02 12-22-13

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30709, opened 2018-03-06
Set quota via provisioning API in UI upload test

phil-davis:
## Description
Take some code from the API acceptance test suite and use it in the UI test suite to set user quota.

## Related Issue
PR #30673 wants to move the users page UI out of core.

## Motivation and Context
There is an upload test in the UI tests that sets user quota prior to the upload. The scenario setup currently sets the quota using the UI logged in as admin. That ties the test (which is about upload) to the existence of the users page in the UI. We don't want such dependencies that complicate life.

## How Has This Been Tested?
CI will know,

## Types of changes
- [x] Test refactoring

## Checklist:

### DIFF
--- a/tests/ui/features/bootstrap/UsersContext.php
+++ b/tests/ui/features/bootstrap/UsersContext.php
@@ -26,6 +26,7 @@ use Behat\Gherkin\Node\TableNode;
 use Behat\Mink\Exception\ExpectationException;
 use Behat\MinkExtension\Context\RawMinkContext;
 use Page\UsersPage;
+use TestHelpers\OcsApiHelper;
 
 require_once 'bootstrap.php';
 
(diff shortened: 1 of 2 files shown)


tests/ui/features/bootstrap/UsersContext.php
```diff
@@ -85,6 +86,44 @@ public function quotaOfUserIsSetTo($username, $quota) {
...
+	 */
+	public function adminSetsUserQuotaToUsingTheAPIz($user, $quota) {
+		$body = new \Behat\Gherkin\Node\TableNode(
+			[
+				0 => ['key', 'quota'],
+				1 => ['value', $quota],
+			]
+		);
+
+		sleep(5);
```
phil-davis: This is just accidentally left here! Was seeing if there was some timing issue in setting quota directly after the user had just been created :)

tests/ui/features/bootstrap/UsersContext.php
```diff
@@ -85,6 +86,44 @@ public function quotaOfUserIsSetTo($username, $quota) {
...
+	 * @When the administrator sets the quota of user :user to :quota using the API
+	 * @Given the quota of user :user has been set to :quota z
+	 *
+	 * @param string $user
+	 * @param string $quota
+	 *
+	 * @return void
+	 */
+	public function adminSetsUserQuotaToUsingTheAPIz($user, $quota) {
+		$body = new \Behat\Gherkin\Node\TableNode(
```
DeepDiver1975: Body should be a simple assoc array - or am I miss understanding anything?

tests/ui/features/bootstrap/UsersContext.php
```diff
@@ -85,6 +86,44 @@ public function quotaOfUserIsSetTo($username, $quota) {
 		$this->usersPage->setQuotaOfUserTo($username, $quota, $this->getSession());
 	}
 
+	/**
+	 * Taken from acceptance Provisioning.php and modified to suit current
+	 * UI test environment. This function should be removed when merging UI
+	 * and API acceptance tests, and the one from Provisioning.php used everywhere.
+	 *
+	 * @When the administrator sets the quota of user :user to :quota using the API
+	 * @Given the quota of user :user has been set to :quota z
```
phil-davis: The "z" here is not required for real! Just making sure to distinguish it from the same matching step in the API-acceptance tests :)

tests/ui/features/bootstrap/UsersContext.php
phil-davis: Yes, you were correct. Underlying calls in the stack in the API test code took apart the ``TableNode`` structure and put it into a more ordinary associative array.

I wasted time here, because when doing this to the provisioning API endpoint it returned me a 401, not authorised, when actually it turns out that the problem is "body contents is crap". Shame about the misleading API response status.

phil-davis:
I pushed code that should work.

DeepDiver1975 (approved):

phil-davis:
Baclport stable10 #30713

Issue #30694, opened 2018-03-05
Deleting a user does not clear user-created external mounts


patrickjahns:
When re-creating a user, external mounts stay persistent (which leads to a potential data leak)

### Steps to reproduce
1. As admin, create a user (i.e. `testuser` )
2. Login as the newly created user (`testuser`)
3. As testuser mount an external storage (`My Storage`) - check that external storage is successfully mounted
4. Logout as testuser
5. Switch back to admin user -> delete the user `testuser` (& verify that the user does not exist anylonger)
6. Re-create the user `testuser`
7. Login as `testuser` 
8. See that the previously created external storage is still present

### Expected behaviour
When deleting a user, I expect that all external storages the user has created, are deleted with it.
When re-creating the user -> the external storages are no longer present in that users files folder

### Actual behaviour
After re-creating - the user has access to the previously created external storages 

### Server configuration
**Operating system**:
Docker `owncloud/server`

**ownCloud version:** (see ownCloud admin page)
10.0.4 & 10.0.7 

**Where did you install ownCloud from:**
docker

### Logs

No error in the logs

sharidas:
Not sure about the complexity of the problem. I assigning this issue to myself.

PVince81:
to be rescheduled

PR #30712, opened 2018-03-06
Remove user from the storage setting or storage when user is deleted

sharidas:
Remove user from the storage if there are multiple
users associated with the storage during deletion
of the user. Else remove the storage when user is
deleted, since the storage is only associated with
the user being deleted.

## Description

Remove user from the external storage list when there are multiple users applicable users for the storage, during deletion of user. Else remove the storage as the user which is being deleted is the only storage associated with the user.

## Related Issue

https://github.com/owncloud/core/issues/30694

## Motivation and Context

This change would help to clean up the external storage when a user is being deleted.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/External/Service/GlobalStoragesService.php
+++ b/lib/private/Files/External/Service/GlobalStoragesService.php
@@ -30,6 +30,7 @@ use OC\Files\Filesystem;
 use OCP\Files\External\IStorageConfig;
 use OCP\Files\External\IStoragesBackendService;
 use OCP\Files\External\Service\IGlobalStoragesService;
+use OCP\IUser;
 
 /**
  * Service class to manage global external storages
@@ -181,5 +182,37 @@ class GlobalStoragesService extends StoragesService implements IGlobalStoragesSe
 		}, $configs);
 
 		return \array_combine($keys, $configs);
+	}
+
+	/**
+	 * Deletes the external storages mounted to the user
+	 *
+	 * @param IUser $user
+	 * @return bool
+	 */
+	public function deleteAllForUser($user) {
+		$userId = $user->getUID();
+		$result = false;
+		//Get all valid storages
+		$mounts = $this->getStorages();
+		foreach ($mounts as $mount) {
+			$applicableUsers = $mount->getApplicableUsers();
+			$id = $mount->getId();
+			if (\in_array($userId, $applicableUsers, true)) {
+				if (\count($applicableUsers) === 1) {
+					//As this storage is associated only with this user.
+					$this->removeStorage($id);
+					$result = true;
+				} else {
+					$storage = $this->getStorage($id);
+					$userIndex = \array_search($userId, $applicableUsers, true);
+					unset($applicableUsers[$userIndex]);
+					$storage->setApplicableUsers($applicableUsers);
+					$this->updateStorage($storage);
+					$result = true;
+				}
+			}
+		}
+		return $result;
 	}
 }
--- a/lib/private/User/User.php
+++ b/lib/private/User/User.php
@@ -43,7 +43,6 @@ use OCP\IConfig;
 use OCP\IUserBackend;
 use OCP\IUserSession;
 use OCP\User\IChangePasswordBackend;
-use OCP\UserInterface;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
@@ -225,6 +224,11 @@ class User implements IUser {
 		}
 		// Delete the user's keys in preferences
 		\OC::$server->getConfig()->deleteAllUserValues($this->getUID());
+
+		// Delete all mount points for user
+		\OC::$server->getUserStoragesService()->deleteAllMountsForUser($this);
+		//Delete external storage or remove user from applicableUsers list
+		\OC::$server->getGlobalStoragesService()->deleteAllForUser($this);
 
 		// Delete user files in /data/
 		if ($homePath !== false) {
--- a/lib/public/Files/External/Service/IGlobalStoragesService.php
+++ b/lib/public/Files/External/Service/IGlobalStoragesService.php
@@ -35,4 +35,12 @@ interface IGlobalStoragesService extends IStoragesService {
 	 * @since 10.0
 	 */
 	public function getStorageForAllUsers();
+
+	/**
+	 * Deletes the external storages mounted to the user
+	 * @param \OCP\IUser $user
+	 * @return bool
+	 * @since 10.0.10
+	 */
+	public function deleteAllForUser($user);
 }
--- a/lib/private/Files/External/Service/UserStoragesService.php
+++ b/lib/private/Files/External/Service/UserStoragesService.php
@@ -27,6 +27,7 @@ namespace OC\Files\External\Service;
 use OC\Files\Filesystem;
 
 use OCP\Files\Config\IUserMountCache;
+use OCP\IUser;
 use OCP\IUserSession;
 
 use OCP\Files\External\IStorageConfig;
@@ -56,6 +57,7 @@ class UserStoragesService extends StoragesService implements IUserStoragesServic
 		IUserMountCache $userMountCache
 	) {
 		$this->userSession = $userSession;
+		$this->userMountCache = $userMountCache;
 		parent::__construct($backendService, $dbConfig, $userMountCache);
 	}
 
@@ -139,5 +141,23 @@ class UserStoragesService extends StoragesService implements IUserStoragesServic
 
 	protected function isApplicable(IStorageConfig $config) {
 		return ($config->getApplicableUsers() === [$this->getUser()->getUID()]) && $config->getType() === IStorageConfig::MOUNT_TYPE_PERSONAl;
+	}
+
+	/**
+	 * Deletes the storages mounted to a user
+	 * @param IUser $user
+	 * @return bool
+	 */
+	public function deleteAllMountsForUser(IUser $user) {
+		$getUserMounts = $this->userMountCache->getMountsForUser($user);
+		$result = false;
+		if (\count($getUserMounts) > 0) {
+			foreach ($getUserMounts as $userMount) {
+				$id = $userMount->getStorageId();
+				$this->userMountCache->removeUserStorageMount($id, $user->getUID());
+				$result = true;
+			}
+		}
+		return $result;
 	}
 }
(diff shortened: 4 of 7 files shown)


lib/private/User/User.php
```diff
@@ -227,6 +226,24 @@ public function delete() {
 		// Delete the user's keys in preferences
 		\OC::$server->getConfig()->deleteAllUserValues($this->getUID());
 
+		//Delete external storage or remove user from applicableUsers list
+		foreach (\OC::$server->getUserGlobalStoragesService()->getStorageForAllUsers() as $storageForAllUser) {
```
PVince81: did you try with `getUserStoragesService()` instead ? this would be more direct and give you the whole list without the need to filter.
However one problem is that this service works with the user from the session, might need to refactor it in a way that allows creating/querying it for a given user.

lib/private/User/User.php
```diff
@@ -227,6 +226,24 @@ public function delete() {
...
+			$id = $storageForAllUser->getId();
+			$applicableUsers = $storageForAllUser->getApplicableUsers();
+			if (in_array($this->getUID(), $applicableUsers, true)) {
+				if (count($applicableUsers) === 1) {
+					//As this storage is associated only with this user.
+					\OC::$server->getGlobalStoragesService()->removeStorage($id);
```
PVince81: but actually you might be right, this removes storages created by an admin and applied only for a user, so maybe this approach is ok

PVince81:
I wonder if this logic could be put in a method of one of the services. But it seems the services don't know from each other yet. Does deleting a storage not work from the "UserGlobalStoragesService" ?

sharidas:
> Does deleting a storage not work from the "UserGlobalStoragesService" ?

If we use "getUserGlobalStoragesService", then the problem happens when `./occ user:delete` is executed. The deletion of storage doesn't work and causes exception. The exception message is `UserGlobalStoragesService writing is disallowed`

PVince81:
Yeah, `UserGlobalStoragesService` is designed for read-only use in the web UI and storage level, to find out all storages the current user has access to, whether configured globally or locally.

Another approach would be to simply have two calls: one for the local storages, maybe `UserStoragesService->deleteAll()` and another one `GlobalStoragesService->deleteAllForUser($userId)`.

It would also make it easier to write unit tests as you'd mostly write tests for these two methods.

lib/private/User/User.php
```diff
@@ -227,6 +226,9 @@ public function delete() {
 		// Delete the user's keys in preferences
 		\OC::$server->getConfig()->deleteAllUserValues($this->getUID());
 
+		//Delete external storage or remove user from applicableUsers list
+		\OC::$server->getGlobalStoragesService()->deleteAllForUser($this->getUID());
```
sharidas: Doesn't local storage gets deleted here. I mean https://github.com/owncloud/core/pull/30712/files#diff-3030ca4989f892c50dea24eaa3f1ae38R236

PVince81: yeah, the home storage gets deleted later as part of user deletion

however home storage is untouched by the any of the *storagesservice* which only operate on external storage currently

mmattel:
@sharidas any update on this ?

tests/lib/Files/External/Service/GlobalStoragesServiceTest.php
```diff
@@ -160,6 +176,96 @@ public function testAddStorage($storageParams) {
...
+	 */
+	public function testDeleteAllForUser($storageParams) {
+		$this->service = new GlobalStoragesService($this->backendService, $this->dbConfig, $this->mountCache);
+		if (\count($storageParams) === 0) {
+			$this->assertFalse($this->service->deleteAllForUser('foo'));
+		} elseif (\count($storageParams['applicableUsers']) >= 1) {
+			$storage = $this->makeStorageConfig($storageParams);
+			$this->service->addStorage($storage);
+			if (isset($storageParams['applicableUsers'])) {
+				$this->assertTrue($this->service->deleteAllForUser($storageParams['applicableUsers'][0]));
```
PVince81: please also verify that the "applicable" value was properly set after `deleteAllForUser` and possibly other values. Checking for the return value is not enough to ensure the function is doing its job

sharidas: Done.

lib/private/Files/External/Service/GlobalStoragesService.php
```diff
@@ -182,4 +182,34 @@ public function getStorageForAllUsers() {
...
+
+	/**
+	 * Deletes the external storages mounted to the user
+	 *
+	 * @param $userId
+	 * @return bool
+	 */
+	public function deleteAllForUser($userId) {
+		$result = false;
+		$mounts = $this->getStorageForAllUsers();
```
PVince81: here I'm wondering about scalability: if there are 10000 users and each have a mount, we're querying all possible mounts, even the personal mounts of irrelevant users

I wonder if we should split this in two parts:
1) get all personal mounts of said user, then remove said storages
and
2) get all system mounts (no personal mounts) and then filter applicable users as you did below

@butonic does that make sense or is the current solution sufficient for real-life scenarios ?

tests/lib/Files/External/Service/GlobalStoragesServiceTest.php
```diff
@@ -160,6 +176,110 @@ public function testAddStorage($storageParams) {
...
+					$finalApplicableUsers = $this->service->getStorage($storage->getId())->getApplicableUsers();
+					if (isset($storageParams['applicableGroups'])) {
+						$finalApplicableGroups = $this->service->getStorage($storage->getId())->getApplicableGroups();
+						$this->assertEquals(\count($storageParams['applicableGroups']), \count($finalApplicableGroups));
+					}
+					$this->assertEquals(1, $initialApplicableUsers - \count($finalApplicableUsers));
+					\array_shift($storageParams['applicableUsers']);
+					$this->assertEquals([], \array_diff($storageParams['applicableUsers'], $finalApplicableUsers));
+				} else {
+					$storages = \count($this->service->getAllStorages());
```
PVince81: you should also make sure that non-relevant storages were not deleted and are still present
for this, create a few unrelated storages and check for these

you could use the "store their id" technique after creation then requery them to see if they still exist and confirm that the correct ones are gone

lib/private/Files/External/Service/GlobalStoragesService.php
PVince81: @sharidas let's go with the proposed approach to play it safe. Here I'm assuming that we already have the necessary API calls to retrieve this info.

sharidas:
Based on https://github.com/owncloud/core/pull/30712#discussion_r196468148 I made changes by adding an extra method at https://github.com/owncloud/core/pull/30712/files#diff-3a0ca1d458a0f44d92a863f6baf81d3dR145. I found that using this method we can get the personal mounts for a user. And the old method we have, that is, https://github.com/owncloud/core/pull/30712/files#diff-16d6579da5930e07b8c14d76623559e8R192 will try to grab the global mount points. I have tested the same with 2 users `admin` and `user1`. Where `admin` user has 2 external storages -  owncloud. And `user1` has - owncloud and SFTP. When `user1` is deleted, the `deleteAllMountsForUser()` would show 3 mounts -> owncloud, SFTP and '/user1'. And when `deleteAllForUser()` is called, 2 global storages are shown - SFTP and owncloud.
My query here would be does this approach look ok?

lib/private/User/User.php
```diff
@@ -226,6 +225,11 @@ public function delete() {
 		// Delete the user's keys in preferences
 		\OC::$server->getConfig()->deleteAllUserValues($this->getUID());
 
+		// Delete all mount points for user
+		\OC::$server->getUserStoragesService()->deleteAllMountsForUser($this);
```
PVince81: `$this->getUID()` ?

lib/public/Files/External/Service/IGlobalStoragesService.php
```diff
@@ -35,4 +35,12 @@
 	 * @since 10.0
 	 */
 	public function getStorageForAllUsers();
+
+	/**
+	 * Deletes the external storages mounted to the user
+	 * @param $userId
+	 * @return bool
+	 * @since 10.0.8
```
PVince81: set to 10.0.10

lib/private/User/User.php
```diff
@@ -226,6 +225,11 @@ public function delete() {
...
+		//Delete external storage or remove user from applicableUsers list
+		\OC::$server->getGlobalStoragesService()->deleteAllForUser($this->getUID());
```
PVince81: be consistent: use either UID or user object

sharidas: Done. Prefer to use user object. Considering future usage.

lib/public/Files/External/Service/IGlobalStoragesService.php
sharidas: Done

sharidas:
I have created a new file in this PR, https://github.com/owncloud/core/pull/30712/files#diff-31407757e5ad52b7e7c4a81530bb4113. The reason why I created a new file is while playing with storageId, I found that `GlobalStoragesServiceTest` extens `StoragesServiceTest` and that was causing issues while testing. So I thought its better to separate this test in a separate file.

sharidas:
@PVince81 review required.

lib/private/Files/External/Service/GlobalStoragesService.php
```diff
@@ -182,4 +183,35 @@ public function getStorageForAllUsers() {
...
+	/**
+	 * Deletes the external storages mounted to the user
+	 *
+	 * @param IUser $user
+	 * @return bool
+	 */
+	public function deleteAllForUser($user) {
+		$userId = $user->getUID();
+		$result = false;
+		$mounts = $this->getStorageForAllUsers();
```
PVince81: It seems this method returns not only admin mounts but also personal mounts as it uses `DBConfigService::getAllMounts()` instead of `getAdminMounts()`.

From reading the code further, it seems you'd need to call `getAllStorages()` or `getStorages()` which do call `DBConfigService::getAdminMounts()` through `StorageService::readConfig()`

PVince81 (changes requested):
Looks good, there's only one additional adjustment needed to avoid performance issue by reading too many configs that aren't needed

lib/private/Files/External/Service/GlobalStoragesService.php
sharidas: Modified the PR, and updated the change to use `getStorages()` https://github.com/owncloud/core/pull/30712/files#diff-16d6579da5930e07b8c14d76623559e8R197

PVince81 (approved):

PVince81:
@sharidas please backport

phil-davis:
@sharidas has this been backported?

Issue #30716, opened 2018-03-06
[9.1.8RC1] Setting up LDAP server raises exception and hangs.


SergioBertolinSG:
### Steps to reproduce

1. Enable ldap app.
2. Set up ldap server.

### Expected behaviour

LDAP server can be configured without problems.

### Actual behaviour

Configuration keeps spinning forever.

![ldap_spinning_forever](https://user-images.githubusercontent.com/9048796/37047807-59baa538-216c-11e8-9b44-903eec27e585.gif)

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.0

**Cache**
DB

**ownCloud version:** (see ownCloud admin page)
{"installed":true,"maintenance":false,"version":"9.1.8.0","versionstring":"9.1.8 RC1","edition":"Enterprise"}

**Updated from an older ownCloud or fresh install:**
Fresh

**The content of config/config.php:**

```

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
No.

**Are you using encryption:** 
No.

**Logs**
```
[...]

PR #29593, opened 2017-11-16
[stable9.1] Fix working with LDAP replica server

tomneedham:
New version of #27598 using the same code as https://github.com/owncloud/user_ldap/pull/138

PR #30721, opened 2018-03-08
Revert "[stable9.1] Fix working with LDAP replica server"

davitol:
Reverts owncloud/core#29593 in order to fix https://github.com/owncloud/core/issues/30716


PVince81 (approved):
:+1: as discussed. that PR was too risky

PR #30740, opened 2018-03-09
Run litmus test inside drone

patrickjahns:
## Description

Execute litmus tests inside drone

## Motivation and Context

Consolidate CI tools and get rid of travis in core

## Checklist:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -111,7 +111,39 @@ pipeline:
       - php occ config:list
     when:
       matrix:
-        TEST_SUITE: selenium
+        TEST_SUITE: [selenium, litmus]
+
+  fix-permissions:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /drone/src/
+      - chown www-data * -R
+    when:
+      matrix:
+        TEST_SUITE: [selenium, litmus]
+
+  litmus-old-endpoint:
+    image: owncloud/litmus
+    pull: true
+    environment:
+      - LITMUS_URL=http://server/remote.php/webdav
+      - LITMUS_USERNAME=admin
+      - LITMUS_PASSWORD=admin
+    when:
+      matrix:
+        TEST_SUITE: litmus
+
+  litmus-new-endpoint:
+    image: owncloud/litmus
+    pull: true
+    environment:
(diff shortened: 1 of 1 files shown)


patrickjahns:
Jenkins failing with:
image



image

patrickjahns:
Since drone does not support triggering on different values in a matrix similar to:
```
    when:
      matrix:
        TEST_SUITE: [ litmus, selenium ]
```

I resorted to introducing two flags:

`INSTALL_SERVER: true` - will execute an arbitrary install step with enabling testing app etc
`CHOWN_SERVER: true` - will chown the data to www-data

.drone.yml
```diff
@@ -111,7 +111,39 @@ pipeline:
...
       matrix:
-        TEST_SUITE: selenium
+        INSTALL_SERVER: true
+
+  fix-permissions:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - cd /drone/src/
+      - chown www-data * -R
```
DeepDiver1975: what about running all steps as user www-data?
this will allow us to work in the apache evnironment and the tests which are currently disabled because of they don't work with root can be reenabled ....

patrickjahns: well then in all steps, every command would need to be prefixed with `sudo -u www-data` to ensure that we do not ever create files with root permissions by chance that would lead to not working later on.

We can think about this, once we have the pipeline in one shape and iterate over it - but I see it out of scope for this PR

DeepDiver1975: > well then in all steps, every command would need to be prefixed with sudo -u www-data to ensure

I meant like using a different user within docker ... USER in Dockerfile ..... (just thinking out loud ....)



> but I see it out of scope for this PR

sure ...

phil-davis (approved):
This seems like a good first move of these tests to drone.
As per discussion in this PR, the environment of the server (Apache...) can be addressed in a following PR.

(I like the Agile way of taking small steps that can be shown to work at each stage)

.drone.yml
patrickjahns: > I meant like using a different user within docker ... USER in Dockerfile ..... (just thinking out loud ....)

got ya - can investigate - though we would still need to run once a `chown` step. The clone plugin from drone does clone as root

PR #30755, opened 2018-03-12
fixes #19811 - only use existing key words 'default' and 'none' as ma…

### LINKED ISSUE #19811: unable to set 0 bytes quota on an user via provisioning api
It's not set the quota of an user to 0 bytes.
0 oder 0B or 0MB are always converted to 0. This is reserved as "default".

We would like to assigne 0 bytes for read-only users.

Tested on ownCloud 8.1.1, but I think there are no changes in 8.1.3.

DeepDiver1975:
…gic values. Any numeric number will be interpreted as a true value.

fixes #19811 for stable10

### DIFF
--- a/apps/provisioning_api/lib/Users.php
+++ b/apps/provisioning_api/lib/Users.php
@@ -277,7 +277,7 @@ class Users {
 				break;
 			case 'quota':
 				$quota = $parameters['_put']['value'];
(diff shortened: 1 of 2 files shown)


PVince81 (approved):
:+1: as discussed

PR #30757, opened 2018-03-13
Remove WebUi for setup/installation - 10.1/11 will require terminal a…

DeepDiver1975:
…ccess

## Description
Terminal access to the hosting web server is required. To install and update ownCloud use the occ commands.

## Motivation and Context
- minimize core code base
- installing on terminal is more secure

screenshot from 2018-03-13 01-11-24
## Types of changes

## Checklist:


PVince81:
will it display a page that prompts the admin to do the setup process on the command line ?

nice to have: also add a hint that one needs to adjust config.php directly after that, especially the parts about "overwrite.cli.url"

DeepDiver1975:
> will it display a page that prompts the admin to do the setup process on the command line ?

yes - see the screenshot above - something nicer could be invented ....



> nice to have: also add a hint that one needs to adjust config.php directly after that, especially the parts about "overwrite.cli.url"

I want to add this as parameter / question during ./occ m:i

GrendelWWU:
For enterprise users it's a good idea. but home users could be unable to cope. (on NAS devices, Ubuntu users...)

phil-davis:
And there are providers who "auto-setup" an ownCloud for you, and you (as the end-user) have an ownCloud admin webUI login, but no command line access yourself.

You can press the "update" button in the webUI when a new version is out, and you want it. With this change, people on those providers will be left without being able to do updates, or the provider is going to have to decide if they will do the updates on behalf of the end-users.

Will that be an issue for those kind of providers?

DeepDiver1975:
> but home users could be unable to cope. (on NAS devices, Ubuntu users...)

I understand your point - but there are much more tasks through out the life time of ownCloud which are hard to almost undoable without terminal access.

We need to teach our user base to use the terminal. Anything else will only cause frustration in later stages.

To me the web based installer is a tool to fastly lockin users to ownCloud and keeps them locked in :smile: ....

GrendelWWU:
@DeepDiver1975:
Myself, I am a Linux admin since many years. I learned with systems like SCO Xenix/SCO Unix more than 30 years ago. GUI? What's that? Ok, for me, using the commandline is no problem.
Other users use tools like Synaptic and expect a GUI to set up the system. On NAS you are told not to use SSH due to warranty & support terms.
I installed the last instances of ownCloud/sciebo via terminal access and I like that. But I'm having all the less experienced users in mind. I've been noticing the "fight" between ownCloud (which I stll prefer) and NextCloud. You might lose many of true users, let them join the "dark side" and use NextCloud.
I guess, we'll have to compromise. Terminal setup for experienced admins and designed as the favourite way to set up an ownCloud instance. But for all those, who can't, you still will need a GUI and/or web based tool for setting up the system. Maybe you can do this as an external application, of course. But I won't lock those less experienced users out. But this is just my point of view.

KharmaScribbles:
There are so many ways to get terminal or SSH access;, even ChromeOS users have access to SSH via a Chrome Web App. 

GCP,  AWS.. don't even need to spin up an instance (at least with GCP)

*I duel boot ChromeOS and GalliumOS (KUbuntu deriv.)*

Linux is very, **VERY** foreign to me, and when I was a Windows user I never so much as edited or viewed a system file. NEVER used command-line but once when I had to debug a Runescape client. It involved typing rcmd and something.

  Since discovering Linux, now,  I try to find out if there is a script or program with CLI implementation before going with a GUI client (or at least, if the GUI program has a command-line option).  I use Lynx and Irssi for examples..    I'm currently looking for the best implementation of a BashBlog static site generator/markup compiler.

I'm a complete newb, and still not sure if I know what "regular expression" means, as an example to my newbness. 
Linux, Bash, Terminal.. ? IFs ANDs and .. whats?  Completely lost most days, but that doesn't take much either ;) 

All this is another language to me but I'm feeling so challenged and eager to learn it so I can use it like it's second nature. I've been on the internet 15 years and I feel like I'm learning how to play Reader Rabbit all over again.

### It's the year of the Linux desktop, btw. ###


The more informed a user is about the all-powerful CLI, the better our technological advancement for the future.
.
As a server admin, I would love to have this as an option, to only allow users who have access to a terminal program, to use OwnCloud.. The more minimilist the code I need to install to my server, the more I want it.

Considering I will be installing OwnCloud for the first time, my fresh install would allow me the freedom to screen who is able to end up using my instance.

Just an opinion I wanted to share. I'm in the..28-31 age bracket if that helps at all.

PVince81:
considering that [Phoenix](https://github.com/owncloud/phoenix) will be the new frontend in OC 11, if we want a setup UI it would need to be rebuild it in Phoenix for "single node" installations.

for people setting up the frontend on a separate server, the backend will need to be installed on CLI

DeepDiver1975:
In a full blown Phoenix setup i See docker (with maybe docker composer) as the only simplified Setup mechanism for the Future.

In such a scenario you anyhow have ssh access or use docker exec to get a bash.

PR #30764, opened 2018-03-13
[WIP] Only parse info.xml once - artur

individual-it:
only to find out why tests failing


individual-it:
problem solved https://github.com/owncloud/core/pull/30188#issuecomment-372679318 closing

Issue #30751, opened 2018-03-12
[files_antivirus] Upload of Eicar throws Sabre\Exception\UnsupportedMediaType


michaelstingl:
### Steps to reproduce
1. run ownCloud 10.0.7 with `files_antivirus`
2. Upload Eicar

### Expected behaviour
I wouldn't expect the nasty PHP exception  in my logs

### Actual behaviour
```
{"reqId":"MdWK3QiFOS31h9ly9agk","level":4,"time":"2018-03-12T20:11:00+00:00","remoteAddr":"172.18.0.1","user":"admin","app":"webdav","method":"PUT","url":"\/remote.php\/webdav\/eicar-av-test.txt","message":"Exception: {\"Message\":\"HTTP/1.1 415 Virus Eicar-Test-Signature is detected in the file. Upload cannot be completed.\",\"Exception\":\"OCA\DAV\Connector\Sabre\Exception\UnsupportedMediaType\",\"Code\":0,\"Trace\":\"
#0 /var/www/owncloud/apps/dav/lib/Connector/Sabre/File.php(188): OCA\DAV\Connector\Sabre\File->convertToSabreException(Object(OCP\Files\InvalidContentException))\
#1 /var/www/owncloud/lib/public/Events/EventEmitterTrait.php(50): OCA\DAV\Connector\Sabre\File->OCA\DAV\Connector\Sabre\{closure}()\
#2 /var/www/owncloud/apps/dav/lib/Connector/Sabre/File.php(269): OCA\DAV\Connector\Sabre\File->emittingCall(Object(Closure), Array, 'file', 'create')\
#3 /var/www/owncloud/apps/dav/lib/Connector/Sabre/Directory.php(171): OCA\DAV\Connector\Sabre\File->put(Resource id #79)\
#4 /var/www/owncloud/lib/composer/sabre/dav/lib/DAV/Server.php(1095): OCA\DAV\Connector\Sabre\Directory->createFile('eicar-av-test.t...', Resource id #79)\
#5 /var/www/owncloud/lib/composer/sabre/dav/lib/DAV/CorePlugin.php(525): Sabre\DAV\Server->createFile('eicar-av-test.t...', Resource id #79, NULL)\
#6 [internal function]: Sabre\DAV\CorePlugin->httpPut(Object(Sabre\HTTP\Request), Object(Sabre\HTTP\Response))\
[...]

PR #30771, opened 2018-03-13
A new exception to suppress exception logs

sharidas:
There are scenarios where an app would have
done the job and the logger would have logged.
But the main task would throw an exception.
Since its logged gracefully we need not throw
exception in the log. This change would help
for the same.

## Description

There can be situations where we need not throw the exceptions to the logger. For example when an antivirus app had detected the file cannot be uploaded and logged gracefully, it would be nice to exclude the exception logged.

## Related Issue

https://github.com/owncloud/core/issues/30751

## Motivation and Context

There can be situations where we need not throw the exceptions to the logger. For example when an antivirus app had detected the file cannot be uploaded and logged gracefully, it would be nice to exclude the exception logged.

## How Has This Been Tested?

- [x] Uploaded the eicar test file and the exception is not logged.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -38,6 +38,7 @@ use OCA\DAV\Connector\Sabre\Exception\Forbidden;
 use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
 use OCA\DAV\Upload\FutureFile;
 use OCA\DAV\Upload\FutureFileZsync;
+use OCP\Files\FileContentNotAllowedException;
 use OCP\Files\ForbiddenException;
 use OCP\Files\InvalidPathException;
 use OCP\Files\StorageNotAvailableException;
@@ -175,7 +176,11 @@ class Directory extends Node implements ICollection, IQuota, IMoveTarget {
 		} catch (InvalidPathException $ex) {
 			throw new InvalidPath($ex->getMessage());
 		} catch (ForbiddenException $ex) {
-			throw new Forbidden($ex->getMessage(), $ex->getRetry());
+			if ($ex->getPrevious() instanceof FileContentNotAllowedException) {
+				throw new FileContentNotAllowedException($ex->getMessage(), $ex->getRetry(), $ex);
+			} else {
+				throw new Forbidden($ex->getMessage(), $ex->getRetry());
+			}
 		} catch (LockedException $e) {
 			throw new FileLocked($e->getMessage(), $e->getCode(), $e);
 		}
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -44,6 +44,7 @@ use OCA\DAV\Files\IFileNode;
 use OCP\Encryption\Exceptions\GenericEncryptionException;
 use OCP\Events\EventEmitterTrait;
 use OCP\Files\EntityTooLargeException;
+use OCP\Files\FileContentNotAllowedException;
 use OCP\Files\ForbiddenException;
 use OCP\Files\InvalidContentException;
 use OCP\Files\InvalidPathException;
@@ -600,6 +601,10 @@ class File extends Node implements IFile, IFileNode {
 	 * @throws \Sabre\DAV\Exception
 	 */
 	private function convertToSabreException(\Exception $e) {
+		if ($e instanceof FileContentNotAllowedException) {
+			// the file content is not permitted
+			throw new FileContentNotAllowedException($e->getMessage(), $e->getRetry(), $e);
+		}
 		if ($e instanceof \Sabre\DAV\Exception) {
 			throw $e;
 		}
--- a/lib/public/Files/FileContentNotAllowedException.php
+++ b/lib/public/Files/FileContentNotAllowedException.php
@@ -1,0 +1,30 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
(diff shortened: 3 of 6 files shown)


lib/public/Files/ExcludeForbiddenException.php
```diff
@@ -0,0 +1,35 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+/**
+ * Public interface of ownCloud for apps to use.
+ * Files/InvalidContentException class
```
tomneedham: ?

lib/public/Files/ExcludeForbiddenException.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+// use OCP namespace for all classes that are considered public.
+// This means that they should be used by apps instead of the internal ownCloud classes
+namespace OCP\Files;
+
+/**
+ * Exception for invalid content
+ * @since 6.0.0
```
tomneedham: ?

VicDeo: 10.0.8 or 10.1.0 ? :unsure:

lib/public/Files/ExcludeForbiddenException.php
VicDeo: c&p ;)

lib/public/Files/ExcludeForbiddenException.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+class ExcludeForbiddenException extends \Exception {}
```
VicDeo: What HTTP status will get a desktop client in this case, something like 50x? Imho this exception should extend `ForbiddenException`... and files_ antivirus should throw it with `$retry=true` https://github.com/owncloud/core/blob/master/lib/public/Files/ForbiddenException.php#L43

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -601,6 +602,10 @@ private function needsPartFile($storage) {
 	 * @throws \Sabre\DAV\Exception
 	 */
 	private function convertToSabreException(\Exception $e) {
+		if ($e instanceof ExcludeForbiddenException) {
+			// the file content is not permitted
+			throw new UnsupportedMediaType($e->getMessage(), 0, $e);
```
VicDeo: it's not that critical to have exactly `UnsupportedMediaType` exception here. Actually I was choosing between **415 Unsupported Media Type** and **418 I'm a teapot**. And some people consider 415 status to be a bug, see https://github.com/owncloud/files_antivirus/issues/177 ;)

VicDeo (changes requested):

lib/public/Files/ExcludeForbiddenException.php
```diff
@@ -0,0 +1,30 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+// use OCP namespace for all classes that are considered public.
+// This means that they should be used by apps instead of the internal ownCloud classes
+namespace OCP\Files;
+
+/**
+ * Exception for invalid content
```
VicDeo: @sharidas Class description need to be updated too.   'ForbiddenException with suppressed logging' or so.

VicDeo (changes requested):

lib/public/Files/ExcludeForbiddenException.php
sharidas: @VicDeo Done.

VicDeo (approved):

VicDeo:
Works as expected. There are no extra entries in log, Antivirus message is properly escalated to UI.

VicDeo:
@PVince81 please consider backport to stable10 or files_antivirus will need a separate branch to support  stable10

apps/dav/lib/Connector/Sabre/Directory.php
```diff
@@ -175,7 +176,11 @@ public function createFile($name, $data = null) {
 		} catch (InvalidPathException $ex) {
 			throw new InvalidPath($ex->getMessage());
 		} catch (ForbiddenException $ex) {
-			throw new Forbidden($ex->getMessage(), $ex->getRetry());
+			if ($ex->getPrevious() instanceof ExcludeForbiddenException) {
+				throw new ExcludeForbiddenException($ex->getMessage(), $ex->getRetry(), $ex);
```
PVince81: this feels rather hacky. we can't create a subclass for every possible exception that we don't want to be logged.

or rename this exception with a name like "FileContentNotAllowedException" which is not semantically about "prevent logging" but about the actual error which is about disallowed file contents

sharidas: Ok, renaming `ExcludeForbidenException` to `FileContentNotAllowedException`.
@VicDeo  I will update https://github.com/owncloud/files_antivirus/pull/219 accordingly.

lib/public/Files/FileContentNotAllowedException.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ */
+
+// use OCP namespace for all classes that are considered public.
+// This means that they should be used by apps instead of the internal ownCloud classes
+namespace OCP\Files;
+
+/**
+ * FileContentNotAllowed to suppress logging.
+ * @since 10.1.0
```
PVince81: make that 10.0.8 if you think we can backport in time

PVince81 (approved):
:+1:

sharidas:
Updated the PR as requested.

sharidas:
Backport PR: https://github.com/owncloud/core/pull/30991

PR #30504, opened 2018-02-16
Use type juggling for comparison on size in chunking plugin

sharidas:
Intially we tried to go ahead with int comparison.
But that didn't worked with 32 bit systems. Because of
explicit conversion of double to int. So in this patch
decided to move ahead with type juggling comparison
or loose comparison of size in chunking plugin

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Loose type comparison of size in chunk plugin helps when large files are uploaded.

## Motivation and Context

Loose type comparison of size in chunk plugin helps when large files are uploaded.

## How Has This Been Tested?

- [x] Verified by uploading various large files 3 GB, 3.8 GB, 4 GB, 4.8 GB, 7.6 GB from web UI.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30772, opened 2018-03-13
[Stable10] Use type juggling for comparison on size in chunking plugin

sharidas:
Intially we tried to go ahead with int comparison.
But that didn't worked with 32 bit systems. Because of
explicit conversion of double to int. So in this patch
decided to move ahead with type juggling comparison
or loose comparison of size in chunking plugin

## Description

Loose type comparison of size in chunk plugin helps when large files are uploaded.

## Motivation and Context

Loose type comparison of size in chunk plugin helps when large files are uploaded.

## How Has This Been Tested?

- [x] Verified by uploading various large files 3 GB, 3.8 GB, 4 GB, 4.8 GB, 7.6 GB from web UI.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Upload/ChunkingPlugin.php
+++ b/apps/dav/lib/Upload/ChunkingPlugin.php
@@ -45,6 +45,9 @@ class ChunkingPlugin extends ServerPlugin {
 	/**
 	 * @param string $sourcePath source path
 	 * @param string $destination destination path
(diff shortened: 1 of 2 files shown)


sharidas:
Backported from: https://github.com/owncloud/core/pull/30504

PVince81 (approved):
:+1:

PR #30776, opened 2018-03-14
Retry on stalled upload

### LINKED ISSUE #30770: Detect chunked upload timeouts and retry
With IE uploads sometimes get stuck. Very important issue for the next server release.

butonic:
fixes https://github.com/owncloud/core/issues/30770

# how to test
use this snippet to fake a bad connection:
```diff
diff --git a/apps/dav/lib/Connector/Sabre/Directory.php b/apps/dav/lib/Connector/Sabre/Directory.php
index d6650d575f..9dcccc3c27 100644
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -488,6 +488,17 @@ class Directory extends Node implements ICollection, IQuota, IMoveTarget {
         * @param resource $data data
         */
        public function createFileDirectly($name, $data) {
-               $this->fileView->file_put_contents($this->getPath() . '/' .  $name, $data);
+               $tgt = $this->fileView->fopen($this->getPath() . '/' .  $name, 'wb');
+
+               while(!feof($data)) {
+                       if (mt_rand(0,99) <= 10) {
+                               sleep(10);
+                               exit();
+                       }
+                       fwrite($tgt, fread($data, 512));
+               }
+               fclose($data);
+               fclose($tgt);
+               //$this->fileView->file_put_contents($this->getPath() . '/' .  $name, $data);
        }
 }
```

now set a lower max chunk size and upload stall timeout:
```
./occ config:app:set files max_chunk_size --value 1024
./occ config:app:set files upload_stall_timeout --value 3
```

you should see the uploads stall quite frequently. but now they get restarted an puck up where they left off.

- [ ] requires doc changes:
  introduces two new app config keys to control stalled uploads:
  - upload_stall_timeout (default 60) in seconds: when to consider an upload stalled
  - upload_stall_retries (default 100): how often to retry 100 because we really want to upload a file.

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -320,8 +320,17 @@ OC.FileUpload.prototype = {
 		if (this.data.isChunked) {
 			this._deleteChunkFolder();
 		}
+		this.data.stalled = false;
 		this.data.abort();
 		this.deleteUpload();
+	},
+
+	/**
+	 * retry the upload
+	 */
+	retry: function() {
+		this.data.stalled = true;
+		this.data.abort();
 	},
 
 	/**
@@ -472,6 +481,17 @@ OC.Uploader.prototype = _.extend({
 	 */
 	davClient: null,
 
+	/**
+	 * Upload progressbar element
+	 *
+	 * @type Object
+	 */
+	$uploadprogressbar: null,
+
+	/**
+	 * @type int
+	 */
+	_uploadStallTimeout: 60,
 	/**
 	 * Function that will allow us to know if Ajax uploads are supported
 	 * @link https://github.com/New-Bamboo/example-ajax-upload/blob/master/public/index.html
@@ -727,7 +747,7 @@ OC.Uploader.prototype = _.extend({
 		// resubmit upload
 		this.submitUploads([upload]);
 	},
-	_trace:false, //TODO implement log handler for JS per class?
+	_trace:true, //TODO implement log handler for JS per class?
 	log:function(caption, e, data) {
 		if (this._trace) {
 			console.log(caption);
@@ -794,34 +814,36 @@ OC.Uploader.prototype = _.extend({
 		var self = this;
 		window.clearInterval(this._progressBarInterval);
 		$('#uploadprogresswrapper .stop').fadeOut();
-		$('#uploadprogressbar').fadeOut(function() {
+		this.$uploadprogressbar.fadeOut(function() {
 			self.$uploadEl.trigger(new $.Event('resized'));
 		});
 	},
 
 	_showProgressBar: function() {
-		$('#uploadprogressbar').fadeIn();
+		this.$uploadprogressbar.fadeIn();
 		this.$uploadEl.trigger(new $.Event('resized'));
 		if (this._progressBarInterval) {
 			window.clearInterval(this._progressBarInterval);
 		}
 		this._progressBarInterval = window.setInterval(_.bind(this._updateProgressBar, this), 1000);
 		this._lastProgress = 0;
-		this._lastProgressStalledSeconds = 0;
 	},
 	
 	_updateProgressBar: function() {
-		var progress = parseInt($('#uploadprogressbar').attr('data-loaded'), 10);
-		var total = parseInt($('#uploadprogressbar').attr('data-total'), 10);
+		var progress = parseInt(this.$uploadprogressbar.attr('data-loaded'), 10);
+		var total = parseInt(this.$uploadprogressbar.attr('data-total'), 10);
 		if (progress !== this._lastProgress) {
 			this._lastProgress = progress;
-			this._lastProgressStalledSeconds = 0;
+			this._lastProgressTime = new Date().getTime();
 		} else {
-			if (this._lastProgressStalledSeconds < 1) {
-				this._lastProgressStalledSeconds++;
-			} else if (progress >= total) {
+			if (progress >= total) {
 				// change message if we stalled at 100%
-				$('#uploadprogressbar .label .desktop').text(t('core', 'Processing files...'));
+				this.$uploadprogressbar.find('.label .desktop').text(t('core', 'Processing files...'));
+			} else if (new Date().getTime() - this._lastProgressTime >= this._uploadStallTimeout * 1000 ) {
+				// stalling needs to be checked here because the file upload no longer triggers events
+				// restart upload
+				this.log('progress stalled'); // retry chunk (and prevent IE from dying)
+				$.each(this._uploads, function(i,e){ console.log(e.retry()) })
 			}
 		}
 	},
@@ -871,9 +893,14 @@ OC.Uploader.prototype = _.extend({
 		if (options.url) {
 			this.url = options.url;
 		}
+		if (options.uploadStallTimeout) {
+			this._uploadStallTimeout = options.uploadStallTimeout;
+		}
 
 		$uploadEl = $($uploadEl);
 		this.$uploadEl = $uploadEl;
+
+		this.$uploadprogressbar = $('#uploadprogressbar')
 
 		if ($uploadEl.exists()) {
 			$('#uploadprogresswrapper .stop').on('click', function() {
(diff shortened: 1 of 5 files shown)


apps/files/js/file-upload.js
```diff
@@ -1049,6 +1078,53 @@ OC.Uploader.prototype = _.extend({
...
 				fail: function(e, data) {
 					var upload = self.getUpload(data);
+					if (upload.data.stalled) {
+						self.log('retry', e, upload);
+						// jQuery Widget Factory uses "namespace-widgetname" since version 1.10.0:
+						var fu = $(this).data('blueimp-fileupload') || $(this).data('fileupload'),
+							retries = upload.data.retries || 0,
+							retry = function () {
+								var uid = OC.getCurrentUser().uid;
+								upload.uploader.davClient.getFolderContents(
```
PVince81: please move this in a separate private function on the Uploader object if possible: `getUploadedBytes()`. Too much nesting makes this difficult to read.

apps/files/js/file-upload.js
```diff
@@ -1049,6 +1078,53 @@ OC.Uploader.prototype = _.extend({
...
+								.done(function (status, files) {
+									data.uploadedBytes = 0;
+									_.each(files, function(file) {
+										// only count numeric file names to omit .file and .file.zsync
+										if (!isNaN(parseFloat(file.name))
+											&& isFinite(file.name)
+											// only count full chunks
+											&& file.size === fu.options.maxChunkSize
+										) {
+											data.uploadedBytes += file.size;
```
PVince81: please add a comment to explain why you need to calculate total size: is it to find out what chunk to resend ?
in general there is already information in the JS class about the last chunk so maybe you only need to PROPFIND said chunk to find out whether it was written at all ?

apps/files/js/file-upload.js
```diff
@@ -1049,6 +1078,53 @@ OC.Uploader.prototype = _.extend({
...
+									self.log('failed to retry', status, ex);
+									fu._trigger('fail', e, data);
+								});
+							};
+						if (upload.data.stalled &&
+							data.uploadedBytes < data.files[0].size &&
+							retries < fu.options.maxRetries) {
+							retries += 1;
+							upload.data.retries = retries;
+							window.setTimeout(retry, retries * fu.options.retryTimeout);
```
PVince81: would be nice if `retry` could also be a private method on the object instead of inline

PVince81:
- [ ] BUG: progress state disappears while waiting for retry, it also seems to reset back. Also it says "Invalid Date" sometimes

PVince81:
I wonder why you didn't simply use the "fileuploadchunkfail" event, might be able to store the chunk id somewhere and reuse for retry instead of doing size computation.

PVince81:
- [ ] TODO: a PROPFIND on the upload folder and getting oc:size might be better than summing up all chunks (the internal Sabre Node logic already omits special files there)

PVince81:
ok so from reading the code of jquery.fileupload.js it doesn't seem to have a way to resume upload directly after failure: https://github.com/owncloud/core/blob/v10.0.7/apps/files/js/jquery.fileupload.js#L802.

This is likely the reason why you chose to go the route of restarting the upload from scratch, but specifying an initial "uploadedBytes" value to make the upload start at this offset.

PVince81:
and apparently most of the code was copied from https://github.com/blueimp/jQuery-File-Upload/wiki/Chunked-file-uploads#automatic-resume and adjusted.

I wonder why the jquery.fileupload devs didn't simply add a "retry" option and call `upload()` again on failure instead of working it around that way...

PVince81:
The reason why the progress bar disappears is because before retrying we first need to abort the upload: this removes the progress bar. But then there's the waiting delay of "retryTimeout" during which no progress bar is shown until an actual retry is attempted. I wonder whether we need this because we already have our own stall detection.

PVince81:
Apart from that it seems to recover properly and it's able to finish the upload. The checksum is the same despite retries. At least with Chromium. Need to try with IE11 and Edge.

We actually need to test not only with the stalled mode but also in case of actual chunk failure where the server returns an error code (like timeout).

PVince81:
- [ ] TEST: IE11
- [ ] TEST: Edge
- [ ] TEST: stalled chunk is retried => upload finishes and checksum matches
- [ ] TEST: erroring chunk upload is retried => upload finishes and checksum matches
- [ ] TEST: multiple files upload with chunking (the abort/retry code might affect the other files that haven't been uploaded yet)

PVince81:
- [ ] TEST: can a chunk be partially written ?

If yes then the logic above cannot work as the server would return the size of all chunks + the partial one.
I do wonder why the wiki page favors this approach instead of just looking at the last chunk offset from the previous send.

PVince81:
- [ ] BUG: "Invalid date" appears in progress bar

PVince81:
- [ ] BUG: when stuck on breakpoint for a long time and resuming, the upload will eventually fail and say "chunks on server do not sum up" instead of properly reuploading

PVince81:
ok apparently we need the former approach to detect non-full-chunks: https://github.com/owncloud/core/pull/30776/commits/bfa559f3100fc17cb7de153cdfcbdedc391c0684#diff-9112ab761e9b96bef644fc2c1f6e35c1L1098

- [ ] revert my commit

PVince81:
Works fine for me with Edge.

PVince81:
- [ ] TEST: with huge files and bandwidth limiter

PVince81:
Maybe during the retry delay we could still display the progress bar with a counter "retrying in ...".

PVince81:
rebased

PVince81:
Just got "chunks on server do not sum up" error at the end of an upload on Edge. Seems to happen randomly.

Also, strangely it seems it resumed at an older chunk than the one expected:
image

and additionally Edge has an empty "GET" request for every "PUT", seems to be a bogus entry not related to our code as it has no request headers:
image

PVince81:
in the above request I looked at the last PROPFIND and found the following chunk names:
1024, 2048, 3072, 4096, 5120, 6144, 7168, 8192, 9216, 10240, 11264
they all had a size of 1024 which leads to the total sum: 1024*11 = 11264
also the total size of ".file" file is 11264.

So I don't understand why the logic decided to retry 11264 instead of moving on to the next one 12288 which is the one that failed.

Maybe an off by one error ? But then it doesn't explain why uploads usually work and seldom have this error.

PVince81:
- [ ] investigate "off by one" error above

- [ ] BUG: stalled/timeout on first chunk doesn't apply: reproducible by making the connection slow with "wondershaper -a vboxnet0 -d 10" when uploading a file. The file takes 120 seconds to upload but should break after the configured 3 seconds timeout

PVince81:
Hmm wait, the reason it doesn't detect stall is mostly because data continues to be sent. Should we maybe add a threshold that if the bit rate is less than that it qualifies as stalled ?

PVince81:
- [ ] BUG: full page reload on timeout with chunk upload and `"wondershaper -a vboxnet0 -d 0.1"` (cheat to make it zero)

PVince81:
Edge behaves strangely. With a slow upload, it doesn't report progress events until very late (like one minute later when the data went through). Only at that point does the "stalled" logic apply and the chunk is retried.

I've used this script https://gist.github.com/PVince81/0307b870071242ce0661962d7fc1f9d0 to simulate random stalling of the network on the VirtualBox network during chunked upload.

Even in this situation, this PR properly retries the upload and the assembled file has the correct checksum.

PVince81:
- [ ] BUG: the "retries" value seems to be global to all uploads (because the "data" object might be reused by the library somehow).

This could cause the value to go up very quickly and reach the max retries quickly.

PVince81:
Fixed the failing test. But we should add more tests also for chunking and the retry part...

PVince81:
Possible improvement: instead of hijacking progress bar code to detect progress, actually use the "fileuploadprogress" and use the "uploadedBytes" value

PVince81:
I've added unit tests for chunk upload, stall and retry nested function.

The bug about randomly occurring "chunks sum up to X not Y" still happens from times to times, not reproducible accurately.

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

butonic:
> Possible improvement: instead of hijacking progress bar code to detect progress, actually use the "fileuploadprogress" and use the "uploadedBytes" value

does not work because these events only fire when progress is happening. no bytes sent => no progress event.

PVince81 (approved):
as discussed with @butonic the current solution is good enough to make it retry chunks :+1:

PR #30790, opened 2018-03-15
fixes objectstore#75 - use storage specific implementations of moveFr…

DeepDiver1975:
…omStorage

## Description

A move operation of files and folders needs it's special implementation on the objectstore level.

please note that this is only a partial fix. In the scenario where objectstore and encryption is enabled the bug will remain visible.

## Related Issue
owncloud/objectstore#75

## How Has This Been Tested?
Manually with scaility
Steps are listed in the ticket.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -319,6 +319,24 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 		return true;
 	}
 
+	public function moveFromStorage(\OCP\Files\Storage $sourceStorage, $sourceInternalPath, $targetInternalPath) {
+		if ($sourceStorage === $this) {
+			return $this->copy($sourceInternalPath, $targetInternalPath);
+		}
+		// cross storage moves need to perform a move operation
+		// TODO: there is some cache updating missing which requires bigger changes and is
+		//       subject to followup PRs
+		if (!$sourceStorage->instanceOfStorage(self::class)) {
+			return parent::moveFromStorage($sourceStorage, $sourceInternalPath, $targetInternalPath);
+		}
+
+		// source and target live on the same object store and we can simply rename
+		// which updates the cache properly
+		$this->getUpdater()->renameFromStorage($sourceStorage, $sourceInternalPath, $targetInternalPath);
+		return true;
+	}
+
+
 	public function getMimeType($path) {
 		$path = $this->normalizePath($path);
 		$stat = $this->stat($path);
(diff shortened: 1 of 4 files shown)


SergioBertolinSG:
Works 👍   Tested https://github.com/owncloud/objectstore/issues/75 with objectstore release/0.5.0 here.

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -319,6 +319,19 @@ public function rename($source, $target) {
 		return true;
 	}
 
+	public function moveFromStorage(\OCP\Files\Storage $sourceStorage, $sourceInternalPath, $targetInternalPath) {
+		if ($sourceStorage === $this) {
+			return $this->copy($sourceInternalPath, $targetInternalPath);
+		}
+		if (!$sourceStorage->instanceOfStorage(self::class)) {
```
PVince81: This is tricky to understand, please add some code comments.

PVince81:
This is tricky to understand and why this is required.

When moving a file between "primary object store X" and "primary object store X" for the same user (home storage), I expect that we can simply fall back to the default implementation (`parent` call).

When moving between a "received share that is localed on object store X but from another user home storage" to "primary object store X home storage of current user", this is a cross-storage move so we need the additional "update" call to keep file id ?

Reading the code seems to imply the opposite as the condition is `!$sourceStorage->instanceOfStorage(self::class)`. I expect a shared storage to also be of that instance.

Maybe some code comments will help understand this better.

PVince81:
does this also properly cover moving a file from primary object store storage to an external storage ?

DeepDiver1975:
> does this also properly cover moving a file from primary object store storage to an external storage ?

no idea - needs a second look. I'd say: no ...

SergioBertolinSG:
Just tried, no, it doesn't work. The file changes its fileID, losing metadata. (Using SFTP)

DeepDiver1975:
> Just tried, no, it doesn't work. The file changes its fileID, losing metadata. (Using SFTP)

thanks for testing ... same applies if encryption is enabled.

These scenarios need a second look - requires bigger changes which are critical.
I propose to address this in follow up prs.

PVince81:
can we have a failing / disabled unit and/or acceptance test for this ? or at least a separate ticket if not possible ?

DeepDiver1975:
> can we have a failing / disabled unit and/or acceptance test for this ? or at least a separate ticket if not possible ?

I prefer to have these scenarios as acceptance tests @SergioBertolinSG is already aware of this.

PVince81 (approved):
:+1:

PR #30793, opened 2018-03-15
Remove cache data from the symfony events

sharidas:
It was a mistake to include cache data.
This change removes the cache data from the
symfony event. And hence the listener of
the event would avail update and oldvalue
from the events.

## Description

Remove cache data from the symfony event dispatcher.

## Motivation and Context

Remove cache data from the symfony event dispatcher. Its not ideal to expose too much information.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -150,7 +150,7 @@ class AppConfig implements IAppConfig {
 	 * @return bool True if the value was inserted or updated, false if the value was the same
 	 */
 	public function setValue($app, $key, $value) {
-		return $this->emittingCall(function () use (&$app, &$key, &$value) {
+		return $this->emittingCall(function (&$afterArray) use (&$app, &$key, &$value) {
 			if (!$this->hasKey($app, $key)) {
 				$inserted = (bool) $this->conn->insertIfNotExist('*PREFIX*appconfig', [
 					'appid' => $app,
@@ -191,14 +191,19 @@ class AppConfig implements IAppConfig {
 					->setParameter('configvalue', $value);
 			}
 
+			if (isset($this->cache[$app]) && isset($this->cache[$app][$key])) {
+				$afterArray['update'] = true;
+				$afterArray['oldvalue'] = $this->cache[$app][$key];
+			}
+
 			$changedRow = (bool) $sql->execute();
 
 			$this->cache[$app][$key] = $value;
 
 			return $changedRow;
 		},[
-			'before' => ['key' => $key, 'value' => $value, 'app' => $app, 'appcache' => isset($this->cache[$app]) ? $this->cache[$app] : null],
-			'after' => ['key' => $key, 'value' => $value, 'app' => $app, 'appcache' => isset($this->cache[$app]) ? $this->cache[$app] : null]
+			'before' => ['key' => $key, 'value' => $value, 'app' => $app],
+			'after' => ['key' => $key, 'value' => $value, 'app' => $app, 'update' => false, 'oldvalue' => null]
 		], 'appconfig', 'setvalue');
 	}
 
--- a/lib/private/Config.php
+++ b/lib/private/Config.php
@@ -132,9 +132,21 @@ class Config {
 	 * @param mixed $value value
 	 */
 	public function setValue($key, $value) {
+		/**
+		 * update and oldvalue help the after event listeners to know
+		 * if its an update or not. If update then get the old value.
+		 */
+		$update = false;
+		$oldValue = null;
 		if ($this->isReadOnly()) {
 			throw new \Exception('Config file is read only.');
 		}
+
+		if (isset($this->cache[$key])) {
+			$update = true;
+			$oldValue = $this->cache[$key];
+		}
+
 		$this->emittingCall(function () use (&$key, &$value) {
 			if ($this->set($key, $value)) {
 				// Write changes
@@ -142,8 +154,8 @@ class Config {
 			}
 			return true;
 		}, [
-			'before' => ['key' => $key, 'value' => $value, 'cache' => $this->cache],
-			'after' => ['key' => $key, 'value' => $value, 'cache' => $this->cache]
+			'before' => ['key' => $key, 'value' => $value],
+			'after' => ['key' => $key, 'value' => $value, 'update' => $update, 'oldvalue' => $oldValue]
 		], 'config', 'setvalue');
 	}
 
@@ -179,8 +191,8 @@ class Config {
 			}
 			return true;
 		}, [
-			'before' => ['key' => $key, 'cache' => $this->cache],
-			'after' => ['key' => $key, 'cache' => $this->cache]
+			'before' => ['key' => $key],
+			'after' => ['key' => $key]
 		], 'config', 'deletevalue');
 	}
 
--- a/lib/public/Events/EventEmitterTrait.php
+++ b/lib/public/Events/EventEmitterTrait.php
@@ -44,12 +44,12 @@ trait EventEmitterTrait {
 	 * @since 10.0.5
 	 */
 	public function emittingCall(\Closure $fn, $arguments ,$class, $eventName) {
-		if (isset($arguments['before']) and count($arguments['before']) > 0) {
+		if (isset($arguments['before']) && count($arguments['before']) > 0) {
 			\OC::$server->getEventDispatcher()->dispatch("$class.before$eventName", new GenericEvent(null, $arguments['before']));
 		}
-		$result = $fn();
-		if (($result !== false) && ($result !== null) and
-			(isset($arguments['after']) and count($arguments['after']) > 0)) {
+		$result = isset($arguments['after']) ? $fn($arguments['after']) : $fn([]);
+		if (($result !== false) && ($result !== null) &&
+			(isset($arguments['after']) && count($arguments['after']) > 0)) {
 			\OC::$server->getEventDispatcher()->dispatch("$class.after$eventName", new GenericEvent(null, $arguments['after']));
 		}
 		return $result;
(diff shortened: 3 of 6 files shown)


lib/private/AppConfig.php
```diff
@@ -193,12 +200,18 @@ public function setValue($app, $key, $value) {
...
 			$this->cache[$app][$key] = $value;
 
+			$update = true;
+
 			return $changedRow;
 		},[
-			'before' => ['key' => $key, 'value' => $value, 'app' => $app, 'appcache' => isset($this->cache[$app]) ? $this->cache[$app] : null],
-			'after' => ['key' => $key, 'value' => $value, 'app' => $app, 'appcache' => isset($this->cache[$app]) ? $this->cache[$app] : null]
+			'before' => ['key' => $key, 'value' => $value, 'app' => $app],
+			'after' => ['key' => $key, 'value' => $value, 'app' => $app, 'update' => &$update, 'oldvalue' => &$oldValue]
```
PVince81: why the `&` sign here ?

sharidas: Because when the update value is updated https://github.com/owncloud/core/pull/30793/files#diff-cdb763d0bc05b09a392296fde32add1dR209 and `&` is not used, then the updated `true` is not used. Hence the update value will not be reflected in the variable `update`. Same for `oldValue`.

PVince81: the `&` sign is ok in the `use()` block but why here in the "after" block ? also you didn't set a `&` in the other hooks in this location

sharidas: The problem I felt was https://github.com/owncloud/core/blob/master/lib/public/Events/EventEmitterTrait.php#L46. The `arguments` array which has both `before` and `after` arrays, is passed before any change is made. So if we don't use `&` then no update will be reflected.

PVince81: as far as I understand how PHP works, if you add a `&` on this line (the one my comment points to), it means that the array consumer will be able to modify the values. Ok, now I see that the array consumer is actually the method defined just before that. So it works as expected but feels a bit convoluted and confusing. I don't expect anyone who decides to add a hook to understand when and when not to add `&`.

Anyway, since this event emitter pattern has already been reviewed and approved before, let's keep the pattern as is.

PVince81: Hmmmmm wait, but https://github.com/owncloud/core/blob/master/lib/public/Events/EventEmitterTrait.php#L46 isn't even using `$update` directly. It's just passing the array as is to the GenericEvent. The GenericEvent arguments must be modified using `setArgument('update', ...)` and not to PHP references.

Have you tried what happens if you remove the `&` ? Does it stop working ?

lib/private/AppConfig.php
```diff
@@ -193,12 +200,18 @@ public function setValue($app, $key, $value) {
 
 			$changedRow = (bool) $sql->execute();
 
+			if (isset($this->cache[$app][$key])) {
+				$oldValue = $this->cache[$app][$key];
```
PVince81: ohhh, so this line retroactively modifies the value of `$oldValue` as referenced in the `after` array... yeah, that's very convoluted and confusing.

any chance to read the `$oldValue` earlier ? or have some of the code outside of the callback ?

lib/private/AppConfig.php
PVince81: follow up here https://github.com/owncloud/core/pull/30793/files#r175119030

lib/private/AppConfig.php
PVince81: other possible approaches:

1) have the callback receive the "before" and "after" arrays as additional function arguments, so it can simply modify the values there when needed. the modification operation will be easier to read and understand than magic updating through reference defined elsewhere

or

2) don't use the emittingCall wrapper and send dispatcher events directly before and after the operation

sharidas: Its better to have approach one. And it worked. Had a basic test with file creation + folder creation, rename, deletion and restore operations. It didn't caused any warning or error in the log file. And the logs were populated.

sharidas: > any chance to read the $oldValue earlier ? or have some of the code outside of the callback ?

yes I updated the PR by moving this logic outside `emittingCall`.

lib/private/AppConfig.php
```diff
@@ -150,6 +150,19 @@ public function hasKey($app, $key) {
...
+		/*
+		 * $update is the flag which helps the event subscriber understand
+		 * if it an insert or update
+		 */
+
+		$update = false;
+		$oldValue = null;
+
+		if (isset($this->cache[$app][$key])) {
+			$oldValue = $this->cache[$app][$key];
```
PVince81: assuming there is a guarantee that the cache is always already populated at this point

sharidas: I have updated for and verified `cache` has keys `$app` and `$key` set. If set then update the oldvalue and update var respectively.

PVince81: it's not about whether to check programmatically

are there any real life scenarios where the cache isn't initialized when we reached this point ? because that could make the event mistakenly think that a key did not exist and have "$update=false" despite the key existing in DB but not being in cache

please verify the logic that populates the cache and see whether it's compatible with the expectation here

sharidas: Got it. There is a method `loadConfigValues()` which can help us in such scenarios if `configLoaded` is set to false before calling `loadConfigValues`. That would at least make sure that the data fetched from DB is consistent with the `cache`. But there is a drawback. It would cause performance, I believe. For every setValue call made, it would fetch the data from the DB ( since the `configLoaded` is set to false).

Whenever the `OC_APP::getAppVersions()`  is called the cache is populated. Another way I see the cache update happens is when the apps are getting enabled.

So basically whenever the call 
```
$appConfig = \OC::$server->getAppConfig();
```
with any of the get or delete method is called in AppConfig is called the `cache` remains updated. Guess chances are bit low that such a scenario would be hit.

PVince81: ok, let's go the safe route then, how about approach 1 from https://github.com/owncloud/core/pull/30793#discussion_r175124811 ?

sharidas: I liked the approach 1 from https://github.com/owncloud/core/pull/30793#discussion_r175124811. Passing 'before' and 'after' arrays to `emittingCall()` would look like:
```
$this->emittingCall(function (&$beforeArray, &$afterArray) use (...){} ... );
```
With this approach we will have to create another PR, where we may have to update calls made in `View` or any place where `emittingCall` is performed. Reason in the emittingCall defenition we need to modify line https://github.com/owncloud/core/blob/master/lib/public/Events/EventEmitterTrait.php#L50
as
```
$result = $fn($arguments['before'], $arguments['after']);
```
A nice idea. But we require another PR which would help this PR to use the approach. Else most of our unit tests related to `emittingCall` would fail. The good part is if we do this now, it would help us in future too.

PVince81: > Else most of our unit tests related to emittingCall would fail

Can you confirm this ? I hoped that just adding arguments would be tolerated by PHP if the receiving callback does not specify them

sharidas: Today morning it happened when I made the change. Let me do the same and will grab the result here. It shouldn't take time I guess. I will update this PR.

sharidas: The drone throws error:
```
39) Test\Encryption\Keys\StorageTest::testRenameKeys with data set #8 ('/user2/files/source.txt', '/user1/files/foo/target.txt', false, false, '/user2/files_encryption/keys/...e.txt/', '/user1/files_encryption/keys/...t.txt/')
Undefined index: after

/drone/src/lib/public/Events/EventEmitterTrait.php:50
/drone/src/lib/private/User/Manager.php:372
/drone/src/lib/private/User/Manager.php:345
/drone/src/lib/public/Events/EventEmitterTrait.php:50
/drone/src/lib/private/User/Manager.php:352
/drone/src/tests/lib/Traits/UserTrait.php:37
/drone/src/tests/lib/Encryption/Keys/StorageTest.php:80
```
This is one among many errror's. Drone link -> https://drone.owncloud.com/owncloud/core/5000/60

PVince81: Ok thanks. Another approach would be to add a new method on the EventEmitterTrait that accepts the two additional arguments. Was the EventEmitterTrait already released in any version before ?

sharidas: We do have EventEmitterTrait available in 10.0.7.

PVince81: ok then let's add a new method and deprecate this one (add `@deprecated`, grep the code for examples)

any new consumer should use the new version in the future

sharidas: I made a minor modification in the `emittingCall()` body and got the drone pass the result.
The line modified is https://github.com/owncloud/core/pull/30793/files#diff-88dd9ddb262ef3b3df2aa72d7793d462R50

PVince81: can you elaborate ? all I see is "and" replaced with "&&"

PVince81: ok I see it now... hmmm

PVince81: this also means that any time you use "after" you are forced to define it as argument in the callback ?

in the case of other emitters, aren't they also using "after" and not specifying the receiving argument ?

if that works then I'm fine with this change

sharidas: The receiving argument is not mandatory. That is we can have either
```
$this->emittingCall(function() use(&$a, &$b) {}, ['before' => [], 'after' => []], 'foo', 'update');
```
or 
```
$this->emittingCall(function($afterArray) use (&$a, &$b) {}, ['before => [], 'after' => []], 'foo', 'read');
```

sharidas: Adding a unit test for the same in for EventEmitterTrait.

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/30937

PR #30791, opened 2018-03-15
[stable10] Propagate move exception messages to the frontend

IljaN:
## Description
Propagate all exception messages to the frontend

## Motivation and Context
Clearer errors

## Screenshots (if appropriate):
http://i.imgur.com/TnIxylc.png

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30800, opened 2018-03-16
Propagate all exception messages to the frontend

IljaN:
Forward port of https://github.com/owncloud/core/pull/30791


PVince81 (approved):
:+1:

Issue #30792, opened 2018-03-15
After upgrade to 10.0.7, user cannot log-in anymore


kerner1000:
### Steps to reproduce

Currently, this is not 100% reproducible, since it is not happening for all users.

### Expected behaviour
ownCloud client cannot connect.
Website shows error.

### Actual behaviour
Tell us what happens instead

### Server configuration
**Operating system**:

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.3 LTS"
```

**Web server:**
```
$ apache2 -version
Server version: Apache/2.4.18 (Ubuntu)
Server built:   2017-09-18T15:09:02
```

**Database:**
```
$ mysql -V      
mysql  Ver 14.14 Distrib 5.7.20, for Linux (x86_64) using  EditLine wrapper
```

**PHP version:**
```
$ php -v
PHP 7.0.22-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.22-0ubuntu0.16.04.1, Copyright (c) 1999-2017, by Zend Technologies
```

**ownCloud version:** (see ownCloud admin page)

10.0.7

**Updated from an older ownCloud or fresh install:**

Upgrade from ??

**Where did you install ownCloud from:**

https://download.owncloud.org/community/owncloud-10.0.7.tar.bz2

**Signing status (ownCloud 9.0 and above):**

`No errors have been found.`

**The content of config/config.php:**

```
# sudo -u www-data php occ config:list system
{
    "system": {
        "instanceid": "ocs9vyuvehvx",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "owncloud.lablicate.com"
        ],
        "datadirectory": "\/var\/data\/owncloud-intern",
        "overwrite.cli.url": "https:\/\/owncloud.lablicate.com",
        "dbtype": "mysql",
        "version": "10.0.7.2",
        "dbname": "owncloud-intern",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "mysql.utf8mb4": true,
[...]

Issue #29709, opened 2017-11-29
Possible to create an app password without a name


mdusher:
### Steps to reproduce
1. Go to Settings -> Security
2. Click "Create new app password" without filling in the name

### Expected behaviour
The user should be requested to enter a name for the app password

### Actual behaviour
The app password is created without a name.

Screenshot:
![image](https://user-images.githubusercontent.com/3529874/33353112-3c83d74a-d500-11e7-86fa-7ed5d5095775.png)

### Server configuration
**Operating system**: Redhat 7

**Web server:** Apache 2.2.15

**Database:** MariaDB 10.0.27 with Galera 25.3.18

**PHP version:** 7.0.23

**ownCloud version:** 10.0.3

**Updated from an older ownCloud or fresh install:** Updated from 8.2.11

**Where did you install ownCloud from:** https://owncloud.org/install/#edition

hodyroff:
Thanks for your feedback. Why do you think a name should be mandatory? It seems in the discretion of the user to set one or not - its a self-service after all and follows the frictionless design criteria. Would be interested in your thoughts.

PVince81:
Maybe it looks just ugly and less useful if one decides to not specify a name.

I believe it's also possible to set the same name to separate tokens.

Maybe we should at least allow renaming them.

mdusher:
The reasoning is entirely UX related. Functionally it's fine without them - but when it comes to users, the only thing they can use to identify the token would be when it was last used. Which if they have multiple tokens which they are constantly using (eg. multiple sync clients that are constantly running) and one becomes compromised, they would end up having to revoke all of them, if they chose not to name any of their tokens.

While I understand this would be a choice made by the user, sometimes it's better to not give them the option.

And as @PVince81, when it's blank - it is a bit ugly. We've had ownCloud 10.0.3 deployed for less than a week and already had a user question why the name was blank after not realising that they did not set a name for the token.

PVince81:
fixed with https://github.com/owncloud/core/pull/29096

PR #30801, opened 2018-03-16
Fix error when reading empty authtoken names from db

IljaN:
Issue: #30792

### DIFF
--- a/lib/private/Authentication/Token/DefaultToken.php
+++ b/lib/private/Authentication/Token/DefaultToken.php
@@ -136,7 +136,7 @@ class DefaultToken extends Entity implements IToken {
 	 */
 	public function setName($name) {
 		if (strlen($name) < 1) {
-			throw new \InvalidArgumentException();
+			$name = '(none)';
 		}
 		parent::setName($name);
 	}


patrickjahns:
Not sure if this is the right path to go - we explicitly disallowed empty tokens to be entered in the frontend (which makes sense).

Alternative should rather be to provide a proper migration when upgrading?

IljaN:
@patrickjahns Where empty names allowed in the version before?

DeepDiver1975:
Title is missleading

IljaN:
@DeepDiver1975 better?

patrickjahns:
@IljaN 
in 10.0.4 it was allowed - with https://github.com/owncloud/core/pull/29831 it was discouraged 
Issue fixed by disallowing empty tokens: 
https://github.com/owncloud/core/issues/29709

Expected behavior changed to requiring a name for the token with 10.0.5

IljaN:
> Alternative should rather be to provide a proper migration when upgrading?

@PVince81 I think @patrickjahns  has a point, what do you think?

PVince81:
Ok for a migration so no code workaround needed.

I just found it weird that we validate the token data even after reading from the DB.

IljaN:
It`s probably because the DB layer is using same setters when populating objects from db

IljaN:
Will create a migration then

patrickjahns:
@IljaN 
Exactly - that's what made me triage the bug to 'lets just create a string in the db'
https://github.com/owncloud/core/blob/master/lib/public/AppFramework/Db/Entity.php#L67-L70

PR #30815, opened 2018-03-19
Set empty authtoken names to 'none'

IljaN:
## Description
Empty authtokens names were allowed in earlier versions, but are now disallowed. This results in a crash during login when empty authokens are encountered from previous version

This migration sets all empty authtoken names to 'none'.

## Related Issue
#30792

## Types of changes

### DIFF
--- a/core/Migrations/Version20180319102121.php
+++ b/core/Migrations/Version20180319102121.php
@@ -1,0 +1,22 @@
+<?php
+namespace OC\Migrations;
+
+use OCP\IDBConnection;
+use OCP\Migration\ISqlMigration;
+
+/**
+ * Sets empty authtoken names to '(none)'
+ * https://github.com/owncloud/core/issues/30792
+ */
+class Version20180319102121 implements ISqlMigration {
+
+	public function sql(IDBConnection $connection) {
+		$q = $connection->getQueryBuilder();
+		$q->update('authtoken')
+			->set('name', $q->expr()->literal('(none)'))
+			->where($q->expr()->eq('name', $q->expr()->literal('')))
+			->orWhere($q->expr()->isNull('name'));
+		return [$q->getSQL()];
+    }
+}
+


IljaN:
- [ ] Backport

core/Migrations/Version20180319102121.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Sets empty authtoken names to '(none)'
+ * https://github.com/owncloud/core/issues/30792
+ */
+class Version20180319102121 implements ISqlMigration {
+
+	public function sql(IDBConnection $connection) {
+		$q = $connection->getQueryBuilder();
+		$q->automaticTablePrefix(true);
```
PVince81: we never used this before I hope it doesn't break :see_no_evil:

IljaN: Is there a better way to get the table prefix? I saw that migration kind "schema" injects prefix trough a param. But this is not the case here...

IljaN: Or do you mean query builder?

PVince81: ah now I remember: you don't need to bother about table prefix. Just remove line 15 and that's it.

core/Migrations/Version20180319102121.php
```diff
@@ -0,0 +1,23 @@
...
+
+		$q->update('authtoken')
+			->set('name', '(none)')
+			->where("name = ''")
```
DeepDiver1975: use expression builder

core/Migrations/Version20180319102121.php
```diff
@@ -0,0 +1,23 @@
...
+			->orWhere('name IS NULL')
```
DeepDiver1975: use expression builder

DeepDiver1975 (changes requested):

IljaN:
@DeepDiver1975 Plz merge :)

PVince81 (approved):
:+1:

PVince81:
@IljaN please backport

PR #30829, opened 2018-03-20
[Test] Remove api tag in AppConfiguration

phil-davis:


phil-davis:
App configuration needs to access the server to set the various sharing... settings. ``BeforeScenario`` methods are not executed in any guaranteed order. ``WebUIGeneralContext`` does inform the API tests ``FeatureContext`` about the base URL to be used, but that comes too late for the ``AppConfiguration`` ``BeforeScenario``  

This needs work so that ``AppConfiguration`` ``BeforeScenario`` can ask ``WebUIGeneralContext`` for the base URL, and set it up correctly.

See PR #30831

PR #30831, opened 2018-03-20
Remove api tags

phil-davis:
## Description
1) Remove ``api`` tags.
2) There is an issue with the order of execution of ``BeforeScenario`` methods. So when ``AppConfiguration.php`` wants to set sharing... settings, and it is running in a ``webUI`` scenario, make sure to get the base URL that the ``webUI`` has set. Then the code will find the server correctly to make the settings.

## Related Issue

## Motivation and Context
Some ``BeforeScenario`` and ``AfterScenario`` methods in the API acceptance tests are only running for scenarios that are tagged ``api``. This could be a problem if a webUI scenario makes use of a related step. e.g. if a webUI scenario uses a step to add a tag on file, then the ``Tags.php`` ``AfterScenario`` is not going to run, so the file tag will not get cleaned up at the end f the scenario.

We should remove the ``api`` tag from as many ``BeforeScenario`` and ``AfterScenario`` methods as we can.

## How Has This Been Tested?
CI

## Types of changes

- [x] Test refactoring

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/AppConfiguration.php
+++ b/tests/acceptance/features/bootstrap/AppConfiguration.php
@@ -22,6 +22,7 @@
  *
  */
 
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 use TestHelpers\AppConfigHelper;
 
 require __DIR__ . '/../../../../lib/composer/autoload.php';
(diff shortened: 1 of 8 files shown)


phil-davis:
Note: the ``api`` tag has been left on ``useBigFileIDs()`` in ``BasicStructure`` - that needs more work to make it successfully in a ``webUI`` test context. Leaving that for another PR.

individual-it (approved):
:+1: if CI passes

phil-davis:
Backport ``stable10`` #30834

PR #30845, opened 2018-03-20
Treat any unknown app version as 0.0.1

### LINKED ISSUE #30724: oc_appconfig installed_version empty causes issues with app upgrades
I was testing some upgrades between 9.1.7 (tarball) and 10.0.7 (tarball) and noticed some strange behaviour. Some old apps use a `version` file instead of a `<version>` tag in the info.xml - which causes them to have an empty value stored in `oc_appconfig:$appid:installed_version`.

On upgrade to an ownCloud 10 version (I was testing 10.0.7 tarballs) this value is compared with the version in the apps folder supplied by the tarball, and is somehow deemed to not need an upgrade. For example, 9.1.7 the shibboleth app is 2.0.0 and in 10.0.7 the version is 3.1.2 and there is a migration that must run - which does not happen.

If, before the upgrade, I manually insert the version into the database - on upgrade, the versions are correctly compared and the upgrade is performed on the app.

Version comparison is done here: https://github.com/owncloud/core/blob/master/lib/private/legacy/app.php#L921

@PVince81 @VicDeo

VicDeo:
## Description
Treat any app version that doesn't fit into 
**[one or more digits]dot[one or more digits][whatever else]**
to be 0.0.1
e.g. **55.12rc7** or **0.1.3beta** are valid versions
while  '', 'null', 'false' are not

## Related Issue
Fixes https://github.com/owncloud/core/issues/30724

## Motivation and Context
App is not upgradable when `installed_version` contains garbage. The most common reason for this is missing `version` entry in appinfo.xml

## How Has This Been Tested?
1. Create apps/shitapp/appinfo
2. add an info.xml
```
<?xml version="1.0"?>
<info>
        <id>shitapp</id>
        <name>Shitapp</name>
        <description>This App provides the shitapp</description>
        <licence>AGPL</licence>
        <author>Shitman</author>
        <types>
                                <filesystem/>
        </types>
        <dependencies>
                <owncloud min-version="10.1" max-version="11.1" />
        </dependencies>
        <namespace>Shitapp</namespace>
</info>
```
3. `php occ app:enable shitapp`
4. add version tag into info.xml:
`	<version>100.100.100</version>`
5. Reload the page

### Expected
App upgrade is triggered on page reload

### Actual
It is not

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/App/CodeChecker/InfoChecker.php
+++ b/lib/private/App/CodeChecker/InfoChecker.php
@@ -40,6 +40,7 @@ class InfoChecker extends BasicEmitter {
 		'id',
 		'licence',
 		'name',
+		'version',
 	];
 	private $optionalFields = [
 		'bugs',
@@ -53,7 +54,6 @@ class InfoChecker extends BasicEmitter {
 		'remote',
 		'repository',
 		'types',
-		'version',
 		'website',
 	];
 	private $deprecatedFields = [
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -302,6 +302,13 @@ class AppConfig implements IAppConfig {
 				$this->cache[$row['appid']] = [];
 			}
 
+			// check if installed_version matches the pattern
+			// one_or_more_digits-dot-one_or_more_digits-any-other-characters
+			if ($row['configkey'] === 'installed_version'
+				&& preg_match('/\d+\.\d+.*$/', $row['configvalue']) !== 1
+			) {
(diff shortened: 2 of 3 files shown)


VicDeo:
@tomneedham you might be interested in this very patch ;)

PVince81 (approved):
:+1: as discussed

PVince81:
@VicDeo please backport

VicDeo:
Stable10: https://github.com/owncloud/core/pull/30890

PR #30756, opened 2018-03-12
Return the quota setting as defined for the user.

DeepDiver1975:
## Description
The quota setting of a user is not part of the response via the provisioning api.

## Related Issue
Fixes #19810

## How Has This Been Tested?
With curl ...

```sh
$ curl -uadmin:admin http://localhost:8080/ocs/v1.php/cloud/users/test?format=jn | json_pp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   322    0   322    0     0    117      0 --:--:--  0:00:02 --:--:--   117
{
   "ocs" : {
      "meta" : {
         "statuscode" : 100,
         "message" : null,
         "status" : "ok"
      },
      "data" : {
         "quota" : {
            "definition" : "5 GB",
            "relative" : 0,
            "used" : 0,
            "total" : 5368709120,
            "free" : 5368709120
         },
         "home" : "/home/deepdiver/Development/owncloud/core/data/test",
         "displayname" : "test",
         "enabled" : "true",
         "two_factor_auth_enabled" : "false",
         "email" : "a@foo.bar"
      }
   }
}
```
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30850, opened 2018-03-21
[stable10] Fixes #19810 - return the quota setting as defined for the user.

### LINKED ISSUE #19810: unable to query assigned user quota via provisioning api
When I request:
GET /ocss/v1.php/cloud/users/user
this results in somthing like that:

```
...
<quota>
    <free>13348716544</free>
    <used>106928627</used>
    <total>13455645171</total>
    <relative>0.79</relative>
</quota>
...
```

It is not possible to query if the user has assigned default, unlimited or xx GB quota.

Tested on ownCloud 8.1.1, but I think there are no changes in 8.1.3.

phil-davis:
Backport #30756

### DIFF
--- a/apps/provisioning_api/lib/Users.php
+++ b/apps/provisioning_api/lib/Users.php
@@ -32,6 +32,7 @@ namespace OCA\Provisioning_API;
 use OC\OCS\Result;
 use OC_Helper;
 use OCP\API;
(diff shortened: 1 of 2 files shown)


PVince81 (approved):
:+1:

PR #30859, opened 2018-03-21
Standardize case of OwncloudPage

phil-davis:
## Description
Spell ``OwncloudPage`` the same everywhere.
While we are here, make the directory reference in ``addPSR4`` not end in a ``/`` - we might as well be consistent in all usages, even though it does not seem to matter.

## Related Issue

## Motivation and Context
I was struggling to work out why there needed to be a special ``require`` for ``OwncloudPage`` in the ``bootstrap.php`` of the ``files_texteditor`` app acceptance tests. Finally I discovered this case issue in the spelling in core.

## How Has This Been Tested?
Run some core and files_texteditor webUI acceptance tests locally.

## Types of changes
- [x] Bug fix to acceptance tests

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/bootstrap.php
+++ b/tests/acceptance/features/bootstrap/bootstrap.php
@@ -24,7 +24,7 @@ require __DIR__ . '/../../../../lib/composer/autoload.php';
 
 $classLoader = new \Composer\Autoload\ClassLoader();
 $classLoader->addPsr4("Page\\", __DIR__ . "/../lib", true);
-$classLoader->addPsr4("TestHelpers\\", __DIR__ . "/../../../TestHelpers/", true);
+$classLoader->addPsr4("TestHelpers\\", __DIR__ . "/../../../TestHelpers", true);
 
 $classLoader->register();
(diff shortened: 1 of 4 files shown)


phil-davis:
Backport ``stable10`` #30860

patrickjahns:
Might be a bit more philosophical - but `ownCloud` is written with the capitcal `C`

individual-it (approved):

phil-davis:
@patrickjahns We already called it ``OwncloudPage`` in its original class. And the class is written like that everywhere else it is used. This PR is just about fixing these wrong-case usages so they are technically correct.

It seems that PHP itself (in the ``extends`` clause at least) is not fussy about case. But the PSR4 ``classLoader`` had been giving us grief, and now I finally saw why.

patrickjahns:
> It seems that PHP itself (in the extends clause at least) is not fussy about case. But the PSR4 classLoader had been giving us grief, and now I finally saw why.

OMG - sometimes I don't wanna know the horrors you've encountered ;-)

PR #30869, opened 2018-03-22
systemtags are always DAV v2

phil-davis:
## Description
Teach ``getDavPathVersion`` that ``systemtags`` is only available since DAV version 2.

## Related Issue

## Motivation and Context
Today I was getting firewall acceptance tests to use core API acceptance test methods to set up the scenarios. Some scenarios needed to create system tags. The step did not work. Eventually I discovered that you have to have the step:
```
Given using new dav path
```
always before doing something like:
```
And user "admin" has created a "normal" tag with name "MyTagName"
```
It would be nice to not have this step dependency. It will save someone else some pain in the future.

## How Has This Been Tested?
Run tags acceptance tests locally.

## Types of changes
- [x] Test refactoring

## Checklist:

### DIFF
--- a/tests/TestHelpers/TagsHelper.php
+++ b/tests/TestHelpers/TagsHelper.php
@@ -153,7 +153,7 @@ class TagsHelper {
 		$userVisible = true,
 		$userAssignable = true,
 		$groups = null,
-		$davPathVersionToUse = 1
+		$davPathVersionToUse = 2
 	) {
 		$tagsPath = '/systemtags/';
 		$body = [
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -23,6 +23,7 @@ use Behat\Gherkin\Node\TableNode;
 use GuzzleHttp\Exception\BadResponseException;
 use GuzzleHttp\Client as GClient;
 use GuzzleHttp\Message\ResponseInterface;
+use GuzzleHttp\Stream\StreamInterface;
 use Sabre\DAV\Client as SClient;
 use Sabre\DAV\Xml\Property\ResourceType;
 use TestHelpers\WebDavHelper;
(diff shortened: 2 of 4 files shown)


tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -114,9 +114,21 @@ public function getDavFilesPath($user) {
 	}
 
 	/**
-	 * @return int DAV path version (1 or 2)
+	 * Select a suitable dav path version number.
+	 * Some endpoints have only existed since a certain point in time, so for
+	 * those make sure to return a DAV path version that works for that endpoint.
+	 * Otherwise return the currently selected DAV path version.
+	 *
+	 * @param string $for the category of endpoint that the dav path will be used for
```
phil-davis: I did this so that we can be flexible for the future in a central place.

e.g. if we have DAV v3 some time, then we can have a way to do "using old dav version" (1) "using new dav version" (2) "using nextgen dav version" (3)...

then something like "systemtags" can return one of 2 or 3, and never return 1. some other new thing can return just a selection from the dav version that it works with.

phil-davis:
And there might be other "general endpoint categories" that only exist since v2. If anyone thinks of them, then they could be added here also.

tests/TestHelpers/TagsHelper.php
```diff
@@ -153,7 +153,7 @@ public static function createTag(
 		$userVisible = true,
 		$userAssignable = true,
 		$groups = null,
-		$davPathVersionToUse = 1
+		$davPathVersionToUse = 2
```
individual-it: if its always "2", why having that parameter at all?

phil-davis: One day it will need to choose between 2 and 3.
I did a bit more that is needed and pushed that. I will have a look at it and see if it looks like it would be useful at all, or just more confusing!

phil-davis: One day it could be 3.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -114,16 +123,46 @@ public function getDavFilesPath($user) {
...
+
 		if ($this->usingOldDavPath === true) {
 			return 1;
 		} else {
 			return 2;
 		}
 	}
 
+	/**
+	 * Select a suitable dav path version number.
```
phil-davis: To self: In the next push, modify this comment.

individual-it (approved):

phil-davis:
I will look again tomorrow and convince myself that this is helpful!

phil-davis:
Backport ``stable10`` #30878

PR #30870, opened 2018-03-22
Adding multiple role support in navigation

sharidas:
The navigation manager should support multiple
roles. It was allowed to enter only one role
in the info.xml. With this change, the app
developer can provide multiple roles comma
separted for role.

## Description

It would be nice to have multiple roles. One scenario that came to my attention was with users page. The users page should be accessible for both admin and group admin. Hence this PR helps for the same.

## Related Issue

https://github.com/owncloud/core/issues/30922

## Motivation and Context

It would be nice to have multiple roles. One scenario that came to my attention was with users page. The users page should be accessible for both admin and group admin. Hence this PR helps for the same.

## How Has This Been Tested?

- [x] Verified that single roles and multiple roles with users page.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/NavigationManager.php
+++ b/lib/private/NavigationManager.php
@@ -52,7 +52,7 @@ class NavigationManager implements INavigationManager {
 	/** @var IGroupManager */
 	private $groupManager;
 
-	function __construct(IAppManager $appManager = null,
+	public function __construct(IAppManager $appManager = null,
 						 IURLGenerator $urlGenerator = null,
 						 IFactory $l10nFac = null,
 						 IUserSession $userSession = null,
@@ -145,13 +145,30 @@ class NavigationManager implements INavigationManager {
 			if (!isset($nav['route']) && !isset($nav['static'])) {
 				continue;
 			}
+
 			$role = isset($nav['@attributes']['role']) ? $nav['@attributes']['role'] : 'all';
-			if ($role === 'admin' && !$this->isAdmin()) {
+			/**
+			 * Can have multiple roles like role="admin,sub-admin"
+			 */
+			$role = explode(',', $role);
+			//Remove whitespace in the role elements
+			$role = array_map('trim', $role);
+
+			$shallContinue = false;
+			if ($this->isAdmin() && (in_array('admin', $role, true) === true)) {
+				$shallContinue = true;
+			}
+			if ($this->isSubAdmin() && (in_array('sub-admin', $role, true) === true)) {
+				$shallContinue = true;
+			}
+			if (in_array('all', $role, true) === true) {
+				$shallContinue = true;
+			}
+
+			if ($shallContinue === false) {
 				continue;
 			}
-			if ($role === 'sub-admin' && !$this->isSubAdmin()) {
-				continue;
-			}
+
 			$l = $this->l10nFac->get($app);
 			$order = isset($nav['order']) ? $nav['order'] : 100;
 			if (isset($nav['route'])) {
--- a/tests/lib/NavigationManagerTest.php
+++ b/tests/lib/NavigationManagerTest.php
@@ -16,6 +16,7 @@ use OC\NavigationManager;
 use OCP\App\IAppManager;
 use OCP\IGroupManager;
 use OCP\IL10N;
+use OCP\ISubAdminManager;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserSession;
@@ -168,7 +169,7 @@ class NavigationManagerTest extends TestCase {
 	/**
 	 * @dataProvider providesNavigationConfig
 	 */
-	public function testWithAppManager($expected, $config, $isAdmin = false) {
+	public function testWithAppManager($expected, $config, $isAdmin = false, $isSubAdmin = false) {
 
 		$appManager = $this->createMock(IAppManager::class);
 		$urlGenerator = $this->createMock(IURLGenerator::class);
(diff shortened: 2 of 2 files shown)


PVince81:
Interesting. cc @pmaier1 


@sharidas what are roles ? is it an alias for groups ?

I don't remember seeing any concept of role in info.xml, would be great if you could clarify this. Maybe provide steps to demonstrate the behavior before and after your PR.

sharidas:
When I tried using user_management app, I was trying to test if everything works as before. And to my surprise it was broken when user logged in as group admin. By broken I mean, the users page wont get displayed or cannot be accessed from the UI for group admin. Had a look at market app's info.xml which showed me https://github.com/owncloud/market/blob/master/appinfo/info.xml#L28. After debugging a bit figured out we could have single value for the role in info.xml. Significance of role is : to whom should the UI of the app should be displayed. So if the role is `admin` only user logged in as admin could see the UI. If the role is `sub-admin` then sub-admin user could see it.

This also brings a question what if an app's UI should be visible or usable for both admin and sub-admin. Hence extended the info.xml to use `role="admin,sub-admin"` ( ref: https://github.com/owncloud/user_management/pull/10/files#diff-224bb875c5d3cfcd7621135069013242R15). The change in the core would help app developers to provide comma separated role ( instead of single role ). Hope this make sense?

PVince81:
Do you mean there is a regression where a group admin cannot use the users page any more ? If yes, please bisect to find when the regression was introduced and make a p1 ticket as this is quite important and could be a release blocker.

DeepDiver1975:
> If yes, please bisect to find when the regression was introduced and make a p1 ticket as this is quite important and could be a release blocker.

no release blocker - this is master only

the sub.admin role was introduced by myself while moving the user management app out side of the core repo

the sub-admin role was never part of stable10 -> no release blocker

sharidas:
> no release blocker - this is master only

Indeed this is not a release blocker as this issue should be there in master only.

PVince81:
we still need a "regression" label and priority somewhere. would be good to have a ticket and clarify that it is master only

lib/private/NavigationManager.php
```diff
@@ -145,13 +145,30 @@ private function init() {
...
+			//Remove whitespace in the role
+			$role = trim($role);
+			/**
+			 * Can have multiple roles like role="admin,sub-admin"
+			 */
+			$role = explode(',', $role);
+
+			$discontinue = true;
+			if ($this->isAdmin() && (in_array('admin', $role, true) === true)) {
+				$discontinue = false;
```
DeepDiver1975: just continue in here

lib/private/NavigationManager.php
```diff
@@ -145,13 +145,30 @@ private function init() {
...
+
 			$role = isset($nav['@attributes']['role']) ? $nav['@attributes']['role'] : 'all';
-			if ($role === 'admin' && !$this->isAdmin()) {
-				continue;
+			//Remove whitespace in the role
+			$role = trim($role);
+			/**
+			 * Can have multiple roles like role="admin,sub-admin"
+			 */
+			$role = explode(',', $role);
```
DeepDiver1975: array elements need to be trimmed

tests/lib/NavigationManagerTest.php
```diff
@@ -232,6 +236,22 @@ public function providesNavigationConfig() {
...
+				'active' => false
+			]], ['navigation' => ['@attributes' => ['role' => 'admin,sub-admin'], 'route' => 'test.page.index']], true],
+			'testSubadmin' => [[[
+				'id' => 'test',
+				'order' => 100,
+				'href' => '/apps/test/',
+				'icon' => '/apps/test/img/app.svg',
+				'name' => 'Test',
+				'active' => false
+			]], ['navigation' => ['@attributes' => ['role' => 'admin,sub-admin'], 'route' => 'test.page.index']], false, true]
```
DeepDiver1975: please add test for 'admin, sub-admin' and 'sub-admin' and 'admin'

DeepDiver1975 (changes requested):

lib/private/NavigationManager.php
```diff
@@ -145,13 +145,30 @@ private function init() {
...
+			//Remove whitespace in the role elements
+			$role = array_map('trim', $role);
+
+			$shallContinue = false;
+			if ($this->isAdmin() && (in_array('admin', $role, true) === true)) {
+				$shallContinue = true;
```
sharidas: Code readability improved. ( no inverted logic ).

lib/private/NavigationManager.php
```diff
@@ -145,13 +145,30 @@ private function init() {
...
+
 			$role = isset($nav['@attributes']['role']) ? $nav['@attributes']['role'] : 'all';
-			if ($role === 'admin' && !$this->isAdmin()) {
-				continue;
+			/**
+			 * Can have multiple roles like role="admin,sub-admin"
+			 */
+			$role = explode(',', $role);
+			//Remove whitespace in the role elements
+			$role = array_map('trim', $role);
```
sharidas: Trim the elements of array.

tests/lib/NavigationManagerTest.php
```diff
@@ -232,6 +236,54 @@ public function providesNavigationConfig() {
...
+				'active' => false
+			]], ['navigation' => ['@attributes' => ['role' => 'admin'], 'route' => 'test.page.index']], true],
+			'testAdminWithSpace' => [[[
+				'id' => 'test',
+				'order' => 100,
+				'href' => '/apps/test/',
+				'icon' => '/apps/test/img/app.svg',
+				'name' => 'Test',
+				'active' => false
+			]], ['navigation' => ['@attributes' => ['role' => '   admin   '], 'route' => 'test.page.index']], true]
```
sharidas: Added tests for "admin,sub-admin", "admin" and "sub-admin". Added tests for roles with spaces.

DeepDiver1975 (approved):

PR #30883, opened 2018-03-23
Run API acceptance tests on drone

phil-davis:
## Description
1) Implement API acceptance tests item in drone pipeline
2) Split the existing API acceptance tests in more suites, so each suite takes less time and the suites can be run in parallel drone jobs
3) Implement the separated suites as separate jobs on Jenkins (for while we still run on Jenkins)
4) Fix issues found with checking download of public shares (discovered when reorganizing suites)

## Motivation and Context
1) Move stuff off Jenkins
2) Make the elapsed time of tests less

## How Has This Been Tested?
CI

## Types of changes
- [x] Test refactoring

## Checklist:


SergioBertolinSG:
When running locally I'm seeing a failure:

PHP_VERSION=7.1 DB_TYPE=mariadb USE_SERVER=true INSTALL_SERVER=true CHOWN_SERVER=true OWNCLOUD_LOG=true BEHAT_SUITE=apiWebdav TEST_SUITE=api drone exec --commit-branch master --build-event push

```
[fix-permissions:L1:2s] chown: changing ownership of '/drone/src/build/bin/bower': No such file or directory
[fix-permissions:L2:19s] chown: changing ownership of '/drone/src/core/vendor': No such file or directory
```

tests/acceptance/run.sh
```diff
@@ -53,24 +53,43 @@ function env_encryption_disable_user_keys {
...
-PHPPID_FED=$!
-echo $PHPPID_FED
+if [ "${TEST_WITH_PHPDEVSERVER}" != "true" ]; then
+    echo "Not using php inbuilt server for running scenario ..."
 
-export TEST_SERVER_URL="http://localhost:$PORT/ocs/"
-export TEST_SERVER_FED_URL="http://localhost:$PORT_FED/ocs/"
+    echo "Adjust trusted hosts"
+    $OCC config:system:set trusted_domains 1 --value=server
+    $OCC config:system:set trusted_domains 2 --value=federated
```
SergioBertolinSG: This part is already done here:
https://github.com/owncloud/core/blob/c036f91d180df960d4c28d6c889c5cd068e7c0ec/.drone.yml#L119-L120

SergioBertolinSG (changes requested):

tests/acceptance/run.sh
phil-davis: Thanks, I had taken that from the old "sample" code in the POC PR.

Jenkinsfile
```diff
@@ -43,16 +43,6 @@ timestampedNode('SLAVE') {
 
         step([$class: 'JUnitResultArchiver', testResults: 'tests/autotest-external-results-sqlite.xml'])
 
-	stage 'Acceptance Testing'
```
patrickjahns: I'd like to keep the acceptance tests both running on jenkins and drone for 1-2 weeks until we have feedback

patrickjahns (changes requested):

patrickjahns:
At some point it would be great to have only one entry script for running ui / api tests -> but we can tackles this with a different PR

Jenkinsfile
phil-davis: OK, I will remove that commit from here.

phil-davis:
@patrickjahns @SergioBertolinSG I reorganized the commits here so that the Jenkins job continues to run, with each suite running as an individual job (and of course in practice the jobs run serially on Jenkins - so there is no improvement to total elapsed time for Jenkins).

In making the separate suites, the various scenarios and setups get executed in a different order from previously if you try to run the whole API acceptance tests in a single run. That results in some errors getting reported for some tests.

PR #30902 will show what happens if we take the existing Jenkins system and simply split the API acceptance test suites. (I am guessing there will be test fails, that are existing "issues" in the test setup/teardown somewhere)

PR #30901 will show what happens if we add testing on drone, and also with the extra split of  API acceptance test suites. I expect there will be the same problems on Jenkins as in PR #30902

PR #30900 shows that just add the drone testing (the first commit in this PR #30883 ) works fine. As long as we keep the current API suite structure, we can put in the drone testing and still keep the Jenkins "single-pass" testing. It might be useful to merge just that commit in PR #30900 and then investigate what are the "latent" issues with reorganizing the API acceptance test suites.

phil-davis:
Please look at #30900 and approve if you agree. Then I can look deeper at what are the issues when reorganizing the ``apiMain`` suite into smaller pieces.

phil-davis:
I looked into the failing tests. PR #30927 has a better set of commits that fix the failing test issue, without messing with the job/s in Jenkinsfile. So that seems better, leaving Jenkinsfile completely untouched for this drone change. 

I will leave this PR open for now, until there is some positive feedback on the other PR.

phil-davis:
Superceded by #30927

PR #30884, opened 2018-03-23
[stable10] Run API acceptance tests on drone

phil-davis:
Backport #30883


phil-davis:
Note: currently this leaves the 2 API acceptance test jobs with encryption switched on, running in Jenkins.

I can work next on sorting that out.

It was not a problem in master, because those jobs did not exist in master!

phil-davis:
This is out-of-date with respect to the PR in master, and it is not sure which PR in master will actually be merged. So closing here - backport can be done when/if needed.

PR #30886, opened 2018-03-23
Handle frontend only apps directly in the navigation manager

DeepDiver1975:
Following info.xml allows apps to simply deliver html+css+js app being natively integrated into the navigation menu

screenshot from 2018-03-23 15-01-33

```xml
<?xml version="1.0"?>
<info>
	<id>static_web_app</id>
	<name>Static Demo</name>
	<summary>Static Demo</summary>
	<description>Static Demo</description>
	<licence>AGPL</licence>
	<author>Thomas Müller</author>
	<version>0.0.1</version>
	<category>tools</category>
	<dependencies>
		<owncloud min-version="10.0" max-version="10.1" />
	</dependencies>
	<navigation>
		<name>Static Demo</name>
                <static>index.html</static>
	</navigation>
	</frontend>
</info>

### DIFF
--- a/lib/private/NavigationManager.php
+++ b/lib/private/NavigationManager.php
@@ -141,7 +141,8 @@ class NavigationManager implements INavigationManager {
 				continue;
 			}
 			$nav = $info['navigation'];
-			if (!isset($nav['route'])) {
+			// either a route or a static page must be defined
+			if (!isset($nav['route']) && !isset($nav['static'])) {
 				continue;
 			}
 			$role = isset($nav['@attributes']['role']) ? $nav['@attributes']['role'] : 'all';
@@ -153,7 +154,15 @@ class NavigationManager implements INavigationManager {
 			}
 			$l = $this->l10nFac->get($app);
 			$order = isset($nav['order']) ? $nav['order'] : 100;
-			$route = $this->urlGenerator->linkToRoute($nav['route']);
+			if (isset($nav['route'])) {
+				$route = $this->urlGenerator->linkToRoute($nav['route']);
+			} else {
+				$html = 'index.html';
+				if (isset($nav['static'])) {
+					$html = $nav['static'];
+				}
+				$route = $this->urlGenerator->linkTo($app, $html);
+			}
 			$name = isset($nav['name']) ? $nav['name'] : ucfirst($app);
 			$icon = isset($nav['icon']) ? $nav['icon'] : 'app.svg';
 			$iconPath = null;
--- a/tests/lib/NavigationManagerTest.php
+++ b/tests/lib/NavigationManagerTest.php
@@ -186,9 +186,14 @@ class NavigationManagerTest extends TestCase {
 		$urlGenerator->expects($this->any())->method('imagePath')->willReturnCallback(function($appName, $file) {
 			return "/apps/$appName/img/$file";
 		});
-		$urlGenerator->expects($this->exactly(count($expected)))->method('linkToRoute')->willReturnCallback(function($route) {
-			return "/apps/test/";
-		});
+		if (isset($config['navigation']['static'])) {
+			$urlGenerator->expects($this->never())->method('linkToRoute')->willReturn('/apps/test/');
+			$urlGenerator->expects($this->once())->method('linkTo')->willReturn('link-to-static');
+		} else {
+			$urlGenerator->expects($this->exactly(count($expected)))->method('linkToRoute')->willReturn('/apps/test/');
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by PVince81** [comment]
How to load javascript files from there as this is a pure HTML template and doesn't contain any `addScript` clauses. Do we also need info.xml entries to specify the main JS file ?

How does it work with routing ? It seems to redirect directly to that page's URL but will it be always loadable despite `.htaccess` rules ?

**Comment by DeepDiver1975** [comment]
> How does it work with routing ? It seems to redirect directly to that page's URL but will it be always loadable despite .htaccess rules ?

good catch - we might need to adopt docs as well as htaccess

**Comment by DeepDiver1975** [comment]
> How to load javascript files from there as this is a pure HTML template and doesn't contain any addScript clauses. Do we also need info.xml entries to specify the main JS file ?

the html file will load all necessary js and css relative to the app folder. no add script magic is required from my pov.

happy to get more input for scenarios I'm not thinking about

**Comment by DeepDiver1975** [comment]
> What we could also do is to let apps provide a folder called "static" which contains all html and JS code to be served statically. Then need to somehow configure htaccess and/or the web server to bypass PHP in these folders for added performance.
> Merge state

js, css and images are already served directly from the app folder as of today.
we just need to make sure html is the same - which is already ...

**Comment by PVince81** [comment]
Depends what you mean with "directly".
In an ideal web server deployment the web server should directly access the files without going through PHP.

From my understanding, every resource like JS, CSS goes through index.php somehow so the ResourceLocator can do its theme magic to find the right file, then it does a `fopen` and streams back its contents.

Maybe providing a direct non-PHP route is out of scope for this PR as long as the files can be accessed under the right URL.

**Comment by DeepDiver1975** [comment]
> From my understanding, every resource like JS, CSS goes through index.php somehow so the ResourceLocator can do its theme magic to find the right file, then it does a fopen and streams back its contents.

no - this was changed long time back ....

**Comment by PVince81** [comment]
Ok, so the resource locator logic is only triggered when using `addScript()` or `addStyle()` but not when pointing a URL directly. I saw no call to "index.php" when testing locally with some of the core icons.

**Comment by DeepDiver1975** [comment]
bildschirmfoto von 2018-03-26 11-46-47

js is loaded without index.php

**Comment by DeepDiver1975** [comment]
html file is loaded as well

bildschirmfoto von 2018-03-26 11-48-40

**Comment by PVince81** [comment]
I've tested with customgroups app located in my "apps3" and I can confirm that the loading URL gets automatically converted to point to "apps3".

Now I wonder how the code is able to load apps that are not located under the OC webroot... It would likely require additional web server configuration to redirect the URLs ?

cc @VicDeo

**Comment by DeepDiver1975** [comment]
> I've tested with customgroups app located in my "apps3" and I can confirm that the loading URL gets automatically converted to point to "apps3".

this is all part of the url generator who knows in which folder the app is living.

so with respect to this pr the link in the nav menu is created using the url generator as well and will point to /apps3/foo_app/index.html which will be served directly

File: lib/private/NavigationManager.php [hunk 1]
```diff
@@ -153,7 +154,15 @@ private function init() {
 			}
 			$l = $this->l10nFac->get($app);
 			$order = isset($nav['order']) ? $nav['order'] : 100;
-			$route = $this->urlGenerator->linkToRoute($nav['route']);
+			if (isset($nav['route'])) {
+				$route = $this->urlGenerator->linkToRoute($nav['route']);
+			} else {
+				$html = 'index.html';
+				if (isset($info['frontend']['@attributes']['template'])) {
```
PVince81: I wonder if we should simply rename "frontend" to "static" as @patrickjahns pointed out recently that "frontend" can have different meanings. You could have a frontend server that computes stuff from another backend just for the sake of giving data to web views.

Also let's not use the word "template" here as index.html isn't really a template, just the entry point.

How about something like "<entrypoint file="index.html"/>" ? (there must be a better name!)

I guess we will only support a single entry point route. So this is really designed for single page applications.

DeepDiver1975: What about this?

```xml
<navigation>
	<name>Static Demo</name>
	<static>index.html</static>
</navigation>

PVince81: Sounds good.

Assuming this means the page in question will have no layout at all.

PVince81 (approved):
:+1:

DeepDiver1975:
@PVince81 backport?

PVince81:
ok for backport

Issue #29839, opened 2017-12-13
[Feature Request] make the apps-external folder default used in new installations


mmattel:
**References:**
https://github.com/owncloud/core/issues/29807 (Move apps from apps to apps2 folder can cause issues)
https://github.com/owncloud/documentation/issues/3621 (Describe procedure how to move downloaded and installed apps from marketplace to the apps2 folder) 

**Background:**
When you have a new install, ``config.php`` gets created and it is important to have all the settings present respectively created early which may affect you when you go for upgrades later on. App management for apps not coming from core is one of those things.

When upgrading an installation via tar, it is a clean way to rename the old instance and extract into an empty folder. Copy your config.php and lucky when aliased your data directory. After doing all the steps to finalize, you just need to do an ./occ upgrade where core does the rest.
But what you have to do manually, identify all those apps which are in the old instance and copy them to the apps folder of the new instance and hopefully do not mix up things... This step is not necessary if the proposal below is implemented.

**Proposal:**
Adding as default into the created config.php the ability to use the ``apps-external`` folder.
see: https://doc.owncloud.org/server/latest/admin_manual/installation/apps_management_installation.html?highlight=apps2

```
  "apps_paths" => array (
      0 => array (
              "path"     => OC::$SERVERROOT."/apps",
              "url"      => "/apps",
              "writable" => false,
      ),
      1 => array (
              "path"     => OC::$SERVERROOT."/apps-external",
              "url"      => "/apps-external",
              "writable" => true,
      ),
  ),
````

**Benefit:**
You do not need to take care on moving/copying apps anymore.
You only need to either copy or alias the apps2 folder to the correct location.

@PVince81 @settermjd

mmattel:
To be added at: https://github.com/owncloud/core/blob/92b658a60e6c28ef9dd7594384180e0d73e96660/lib/private/Setup.php#L384

cdamken:
we have to add it in the config.sample.php for documentation

mmattel:
@cdamken pls see issue https://github.com/owncloud/documentation/issues/3933 (config.sample.php issues)
Your point is not the only one I identified

mmattel:
@cdamken see PR: https://github.com/owncloud/documentation/pull/3934 to fix the ``config.sample.php`` issue

PVince81:
Have you tried creating a separate file before running setup:
- create "config/apps2.config.php" containing the settings to bootstrap
- run occ maintenance:install

[...]

PR #30889, opened 2018-03-23
Add per default the apps-external directory in config.php during installation

mmattel:
## Description

When you install the system, a config.php parameter is added to enable the apps-external directory.
An empty apps-external/ directory has been added as part of defaut shipped directories. 
Note: if you follow the documentation and run the setup/permisions script, all needed directories including apps-external/ will get correct permissions and ownership.

## Related Issue

https://github.com/owncloud/core/issues/29839 ([Feature Request] make the apps-external folder default used in new installations)

## Motivation and Context

See the issue description regarding the benefits.

## How Has This Been Tested?

Install a new system from scratch.

```
  'apps_path' => 
  array (
    0 => 
    array (
      'path' => '/var/www/owncloud/core/apps',
      'url' => '/apps',
      'writable' => 'false',
    ),
    1 => 
    array (
      'path' => '/var/www/owncloud/core/apps-external',
      'url' => '/apps-external',
      'writable' => 'true',
    ),
  ),
```
## Types of changes

## Checklist:

A documentation PR follows asap.
@settermjd fyi

### DIFF
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -383,6 +383,29 @@ class Setup {
 				$config->setSystemValue('logtimezone', \date_default_timezone_get());
 			}
 
+			// adding the apps-external directory by default using apps_path
+			$apps2Key = \OC::$server->getSystemConfig()->getValue('apps_paths', false);
+
+			// add the key only if it does not exist (protect against overwriting)
+			if ($apps2Key === false) {
+				$defaultAppsPaths = [
+					'apps_paths' => [
+						[
+							"path" => \OC::$SERVERROOT . '/apps',
+							"url" => "/apps",
+							"writable" => false
+						],
+						[
+							"path" => \OC::$SERVERROOT . '/apps-external',
+							"url" => "/apps-external",
+							"writable" => true
+						]
+					]
+				];
+						
+				$config->setSystemValues($defaultAppsPaths);
+			}
+
 			self::installBackgroundJobs();
 
 			// save the origin version that we installed at
@@ -390,6 +413,8 @@ class Setup {
 
 			//and we are done
 			$config->setSystemValue('installed', true);
+
+			// finished initial setup
 		}
 
 		return $error;
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ core_vendor=core/vendor
 
 core_doc_files=AUTHORS COPYING README.md
 core_src_files=$(wildcard *.php) index.html db_structure.xml .htaccess .user.ini robots.txt
-core_src_dirs=apps core l10n lib occ ocs ocs-provider ocm-provider resources settings
+core_src_dirs=apps apps-external core l10n lib occ ocs ocs-provider ocm-provider resources settings
 core_test_dirs=tests
 core_all_src=$(core_src_files) $(core_src_dirs) $(core_doc_files)
 core_config_files=config/config.sample.php config/config.apps.sample.php
@@ -262,7 +262,7 @@ $(dist_dir)/owncloud: $(composer_deps) $(nodejs_deps) $(core_all_src)
 	rm -Rf $@/core/vendor/*/{.bower.json,bower.json,package.json,testem.json}
 	rm -Rf $@/l10n/
 	find $@/core/ -iname \*.sh -delete
-	find $@/{apps/,lib/composer/,core/vendor/} \( \
+	find $@/{apps/,apps-external/,lib/composer/,core/vendor/} \( \
 		-name bin -o \
 		-name test -o \
 		-name tests -o \
@@ -273,7 +273,7 @@ $(dist_dir)/owncloud: $(composer_deps) $(nodejs_deps) $(core_all_src)
 		-name travis -o \
 		-iname \*.sh \
 		\) -print | xargs rm -Rf
-	find $@/{apps/,lib/composer/} -iname \*.exe -delete
+	find $@/{apps/,apps-external/,lib/composer/} -iname \*.exe -delete
 	# Set build
 	$(eval _BUILD="$(shell date -u --iso-8601=seconds) $(shell git rev-parse HEAD)")
 	# Replace channel in version.php
@@ -313,7 +313,7 @@ $(dist_dir)/qa/owncloud: $(composer_dev_deps) $(nodejs_deps) $(core_all_src) $(c
 	rm -Rf $@/core/vendor/*/{.bower.json,bower.json,package.json,testem.json}
 	rm -Rf $@/l10n/
 	find $@/core/ -iname \*.sh -delete
-	find $@/{apps/,lib/composer/,core/vendor/} \( \
+	find $@/{apps/,apps-external/,lib/composer/,core/vendor/} \( \
 		-name test -o \
 		-name examples -o \
 		-name demo -o \
@@ -322,7 +322,7 @@ $(dist_dir)/qa/owncloud: $(composer_dev_deps) $(nodejs_deps) $(core_all_src) $(c
 		-name travis -o \
 		-iname \*.sh \
 		\) -print | xargs rm -Rf
-	find $@/{apps/,lib/composer/} -iname \*.exe -delete
+	find $@/{apps/,apps-external/,lib/composer/} -iname \*.exe -delete
 	# Set build
 	$(eval _BUILD="$(shell date -u --iso-8601=seconds) $(shell git rev-parse HEAD)")
 	# Replace channel in version.php


mmattel:
@VicDeo I followed your advice and added the apps2/ directory by default including the .gitkepp file in it + removing the obsolete directory creation logic.

PVince81:
- [ ] needs update of "make dist" and "make dist-qa" route to also copy/create the "apps2" folder for inclusion in the release tarballs

mmattel:
@PVince81 
What is the name/location of the file to adopt "make dist" and "make dist-qa" ?
Or is this something you do centrally ?

lib/private/Setup.php
```diff
@@ -373,15 +373,42 @@ public function install($options) {
...
 			$config->setSystemValue('installed', true);
+
+			// adding the apps2 directory by default using apps_path
+			$appsDir = \OC::$SERVERROOT . '/apps';
+			$apps2Dir = \OC::$SERVERROOT . '/apps2';
+			(bool)$apps2Key = \OC::$server->getSystemConfig()->getValue('apps_paths', false);
+			
+			// add the key only if it doea not existent (protect possible overwriting)
+			if (!$apps2Key) {
+				$z = array(
```
VicDeo: @mmattel `$z` is a bit cryptic. You can name it `$defaultAppsPaths`

lib/private/Setup.php
```diff
@@ -373,15 +373,42 @@ public function install($options) {
...
 
 			self::installBackgroundJobs();
 
-			//and we are done
 			$config->setSystemValue('installed', true);
+
+			// adding the apps2 directory by default using apps_path
+			$appsDir = \OC::$SERVERROOT . '/apps';
+			$apps2Dir = \OC::$SERVERROOT . '/apps2';
+			(bool)$apps2Key = \OC::$server->getSystemConfig()->getValue('apps_paths', false);
```
VicDeo: Casting to bool kills a kitten somewhere...  just use `$apps2Key === false` in the condition below

lib/private/Setup.php
```diff
@@ -373,15 +373,42 @@ public function install($options) {
...
 				$config->setSystemValue('logtimezone', date_default_timezone_get());
 			}
 
 			self::installBackgroundJobs();
 
-			//and we are done
 			$config->setSystemValue('installed', true);
+
+			// adding the apps2 directory by default using apps_path
+			$appsDir = \OC::$SERVERROOT . '/apps';
```
VicDeo: These 2  variables ($appsDir, $apps2Dir) are not used outside the `if` so it would be better to move them inside

lib/private/Setup.php
```diff
@@ -373,15 +373,42 @@ public function install($options) {
...
+			$apps2Dir = \OC::$SERVERROOT . '/apps2';
+			(bool)$apps2Key = \OC::$server->getSystemConfig()->getValue('apps_paths', false);
+			
+			// add the key only if it doea not existent (protect possible overwriting)
```
VicDeo: typo: `doea`

VicDeo (dismissed):

mmattel:
@VicDeo 
I added all the changes you requested and sucessfully rerun a test by doing a new install procedure.

mmattel:
@PVince81 
I added ``apps2/`` to the Makefile in both sections required (dist and dist-qa)

patrickjahns:
I am very reluctant with this change - it might have implications on:
- automated scripts deploying owncloud (ansible, salt etc.)
- needs to be tested with our docker containers, since they already use a custom directory for downloaded apps (ref: https://github.com/owncloud-docker/base/blob/master/rootfs/root/owncloud/config.php#L12-L15 )

"apps2" as name is not very descriptive -> might be confusing for people.

mmattel:
@patrickjahns 
- Regarding automation: Automation and upgrades are exactly the reason for this PR. In a nutshell, if you do not do it, any upgrade is a pain with respect to apps, for more details pls see the text in the referenced issue.
- "apps2" as name is not very descriptive: I just reused the name as it was already defined in the documentation section apps and the permission script in documentation not changing things on too many places. If there is the whish to change the directory name, no problem, but we then have to do this in documentation too. My very personal POV, apps2 can be improved but is not that bad...
- Regarding Docker: this PR checks if the key ``apps_path`` is already present and does nothing in case of preventing overwriting. If the Docker scripts runs after config.php is created, you just need to check if the key exists and do nothing (in the same way as I am doing).

I had a discussion with @PVince81 about the reason for this PR, he maybe can add his opinion.

cdamken:
we could use apps as the apps that are delivered with the tar file and apps-external as all other apps that are not delivered in the tar file?

That would be helpful for the upgrade just making a copy of that folder to the new ownCloud

mmattel:
I am happy to change the name of ``apps2`` to ``apps-external`` if you vote for - just let me know and I proceed.
The location of changes are:
- This PR
- config.sample.php (there is a PR open with improved text covering the apps_path key anyway)
- The documetation (some files)
- The Docker file @patrickjahns mentioned

mmattel:
@cdamken 
I renamed ``/apps2`` to ``/apps-external``.
If this gets merged, I will update all other documents accordingly.

@patrickjahns 
Can you support me with Docker.

tboerger:
Within our docker containers we are writing this config file before the installation: https://github.com/owncloud-docker/base/blob/master/rootfs/root/owncloud/config.php

mmattel:
I would need help on the following (maybe a stupid question):
How can I assign the key 'path' the value ``OC::$SERVERROOT . '/apps'`` 
Means that the string contains the above text. Else it will always export the absolut path of the instance.

PVince81:
there are conflicts now

@patrickjahns are we fine merging this for 10.0.9 now ?

mmattel:
@PVince81 just resolvong the conflicts.
Question: how can I assign the key ``path`` the value ``OC::$SERVERROOT .'/app'`` in a way that it writes to config.php this exact string and not resolving it to it´s absolute path?
I tried a lot but have not found a way as it always adds single quotes around making it unusable.
I will push now the "resolved path" way and if there is no sulution this can be taken.

PVince81:
@patrickjahns move forward with this or are there any concerns ?

patrickjahns:
Thinking more of it - and the future of packaging (*) I have the following concern: 

- How are apps treated, when they are both in apps / apps-external? 
- If both same apps are in two locations - how would this affect performance ( related to https://github.com/owncloud/core/pull/30188 )

More background on the thougts/notions of releases in the future:
A ownCloud release will become slimmer and essentially only consist of the core apps. With the move to the marketplace we had envisioned, that people would later be able to define their "owncloud flavor" ( core + generally used applications ). Meaning the installation would be more easy and it would already contain the latest compatible app. 

Assuming the person updates apps inbetween releases, the update would go into `apps-external` - now the next core release would contain a never app (in `apps` ). Which app would be choosen

mmattel:
A lot of valid questions, I will try to answer them as best I can.

First of all, the config parameter is already present since ages and actively used.
So we have this situation of having apps and a secondary apps directory present.
I just formalize this existing behaviour by naming it consistently and by design from first installation.
I have not heard about any issues causing problems when having this setup.
In addition @tboerger uses this for docker also.
Without this PR this has to be set up by the admin manually.
The big existing disadvantage is the mix of core delivered apps and all other apps being in the same directory. This is because of upgrading an instance, please see referenced isse for more details: #29839 ([Feature Request] make the apps-external folder default used in new installations)
This PR addresses this topic from the beginning by splitting the directories from the start of a new install.

To your statements/questions

> How are apps treated, when they are both in apps / apps-external?

In case that both apps have identical versions, the app from the primary folder (apps) is taken as it it queried first. In the hypothetical case the app in the apps-external directory is newer, this one is taken.

> If both same apps are in two locations - how would this affect performance ( related to #30188 )

Not affecting performance, the location does not matter

> A ownCloud release will become slimmer and essentially only consist of the core apps. 

Which I hope so coming soon and where this PR is a great help.

> With the move to the marketplace we had envisioned, that people would later be able to define their "owncloud flavor" ( core + generally used applications ). Meaning the installation would be more easy and it would already contain the latest compatible app.

This PR supports this by making this also clear from the apps physical location.

> Assuming the person updates apps inbetween releases, the update would go into apps-external - now the next core release would contain a never app (in apps ). Which app would be choosen

Core shipped apps will be updated to the apps folder while all other apps will go into apps-external (if configured). This is btw the current situation. No brake of existig functionality !

tboerger:
@patrickjahns AFAIK there had been a fix that always the latest app version gets loaded. I think @vicdeo has implemented a fix a few months ago.

mmattel:
@patrickjahns any more agreement / disagreement ?

patrickjahns:
Haven't gotten around to testing it yet - if these remarks are addressed its fine by me - final decision to be taken in which release this would appear. Generally wondering how we can transition people into having this desired state

mmattel:
@patrickjahns remarks have been addressed. Merging and Backporting desired.
> Generally wondering how we can transition people into having this desired state

**Two situations:**
- **New installs:**
By merging this PR, the installation creates the apps-external/ dir automatically, core apps stay in apps/ admin downloaded apps get written to the new apps-external/ dir. This improves further handling of owncloud upgrades greatly. The documentation is already prepared for this situation.
- **Existing installs:**
This is a bit different story, because you need to create the apps-external directory manually and move all non core apps to this directory. Note 1: Upgrading oC post moving apps is the same as above. Note 2: We wanted in documentation to support the move of apps by a script and have identified that core needs to improve the identification of core apps in a better way. Pls see https://github.com/owncloud/core/issues/32048 and the refrenced doc PR which includes a lot of details

patrickjahns:
@ownclouders rebase

ownclouders:

patrickjahns:
For master this seems reasonable. To change this behavior on stable10 I am not sure.

Would also like to have final agreement from @pmaier1 @felixboehm

ownclouders:

lib/private/Setup.php
```diff
@@ -378,13 +378,40 @@ public function install($options) {
...
+						],
+					    [
+							"path" => \OC::$SERVERROOT . '/apps-external',
+							"url" => "/apps-external",
+							"writable" => true
+						]
+					]
+				];
+					    
+				$config->setSystemValues($defaultAppsPaths);
```
phil-davis: If we do this, it will be convenient for automated testing, because we will always have specific values of ``apps_paths`` in the config - no need for test scenarios to jump through hoops to work out what they should/could be.

patrickjahns: > If we do this, it will be convenient for automated testing, because we will always have specific values of apps_paths in the config - no need for test scenarios to jump through hoops to work out what they should/could be.

The large goal is moving forward with immutable `config.php` - and being able to change the apps_path via a command will not work

phil-davis: That will be a challenge for any tests that want to adjust ``config.php`` values for testing purposes
I guess when testing remote "black-box" installs, we might anyway skip tests that mess with ``config.php``, since those sort of tests might break the remote system and not (be able to) put it back.
Anyway, a conversation for another place.

mmattel:
@patrickjahns 
The basic idea behind having an ``apps_paths`` key set by default is, that you can update an existing instance without worrying of apps that have been post installed and are not part of the tar file. As you have two directories, one is part of the tar and fresh installed on each update, the other one contains your installed apps, you just need nothing more than copying the old config.php file to the update and relink the data directory.

The broader discussion about a design change including config.php is busting this PR.
As the key ``apps_paths`` is out for a while and actively used including our docker environment, you have in any case when you redesign config.php management the need of a migration step.

I propose to merge this PR to ease admin life and have the config.php management discussion in a seperate thread.

PVince81:
php-cs-fixer is not happy

I think we can at least merge this for master / OC 11

mmattel:
I just updated the PR which solved the conflict in Makefile.

mmattel:
@PVince81 approve? mergable ?

mmattel:
Just squashed the commits

butonic (approved):

PVince81:
@mmattel please backport

@phil-davis will this require specific changes in our testing pipeline ? or were you referring to possible future changes we **could** do ?

VicDeo:
- [ ] Web updater needs to create this dir if it doesn't exist.

phil-davis:
> @phil-davis will this require specific changes in our testing pipeline ? or were you referring to possible future changes we **could** do ?

It just means we can run acceptance tests that check installing an app in the secondary apps path without having to modify `config.php` - sometimes the system-under-test does not let us modify `config.php` or it is better that we do not purposely modify it.

phil-davis:
Backport `stable10` #34656

PR #30843, opened 2018-03-20
Adding a system configuration which allows to set global CORS domains.

DeepDiver1975:
## Description
As of today CORS domains can only be setup by a user for their own access.
There are situations where a globally deployed tool from a different domain might also want to access
ownCloud using CORS requests.

This PR introduces this feature.

## Motivation and Context
Phoenix requires this ...

## How has this been tested?
Using the sample app https://github.com/noveens/staticOwncloudWebapp
requires #30836 as well

## Screenshots (if appropriate):
screenshot from 2018-03-20 09-33-50

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30906, opened 2018-03-26
[stable10] Adding a system configuration which allows to set global C…

DeepDiver1975:
…ORS domains.

backport #30843


PVince81 (approved):
:+1:

PR #30795, opened 2018-03-15
Add symfony event for resolving private links by apps

PVince81:
## Description
Apps can now resolve private links whenever the link was not resolvable by the files app.
    
Currently only the trashbin is responding to this in case a file is in trash.

## Related Issue
Will be required for https://github.com/owncloud/core/pull/30106 to also resolve file id to pending share page URL

## Motivation and Context
Will be required for https://github.com/owncloud/core/pull/30106 to also resolve file id to pending share page URL.
Also decouples the ViewController from trashbin...

## How Has This Been Tested?
Manual test by opening a private link pointing at a trashed folder: the web UI opens in the trashed folder.
Unit tests.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30911, opened 2018-03-26
[stable10] Add symfony event for resolving private links by apps

DeepDiver1975:
Apps can now resolve private links whenever the link was not resolvable
by the files app.

Currently only the trashbin is responding to this in case a file is in
trash.

backport #30795


PVince81 (approved):
:+1:

PR #30806, opened 2018-03-19
Remove unused config parameters

phil-davis:
## Description
Remove config parameters:
``check_for_working_webdav``
``check_for_working_wellknown_setup``
``openssl`` 

## Motivation and Context
Remove stuff that is not used anywhere.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #30919, opened 2018-03-26
[stable10] Remove unused config parameters

phil-davis:
Backport #30806 

AFAICT these are also unused in ``stable10``

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -568,26 +568,11 @@ $CONFIG = array(
 'has_internet_connection' => true,
 
 /**
- * Allows ownCloud to verify a working WebDAV connection. This is done by
- * attempting to make a WebDAV request from PHP.
- */
-'check_for_working_webdav' => true,
-
-/**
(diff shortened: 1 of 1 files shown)


config/config.sample.php
```diff
@@ -909,18 +894,6 @@
...
 
-
-/**
- * SSL
- */
-
-/**
- * Extra SSL options to be used for configuration.
- */
-'openssl' => array(
```
VicDeo: https://github.com/owncloud/core/blob/stable10/apps/encryption/lib/Crypto/Crypt.php#L163

phil-davis: Hmmm, thanks for pointing this out. That same use is also in the encryption repo, which is unbundled from ``core`` ``master``. I had better put that back in ``master`` ``config/config.sample.php`` ! And then backport properly here.

phil-davis:
This is over-zealous. See #30929 for a fix to ``master``. Then the net-change in ``master`` can be backported.

Closing here.

Issue #30915, opened 2018-03-26
cannot set own full name or email address


individual-it:
### Steps to reproduce
1. login as user
2. navigate to personal general settings page
3. change full name or email address

### Expected behaviour
change should work

### Actual behaviour
![image](https://user-images.githubusercontent.com/2425577/37904446-835f23aa-311b-11e8-9351-40cc7ba1a9e1.png)

### Server configuration
**Operating system**:
Ubuntu 17.10
**Web server:**
Apache2
**Database:**
Sqlite
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
10.1 prealpha
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

**The content of config/config.php:**

```
{
    "system": {
        "license-key": "***REMOVED SENSITIVE VALUE***",
        "instanceid": "oc35rjhx6uol",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "localhost"
        ],
        "datadirectory": "\/home\/artur\/www\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/localhost\/owncloud-core",
        "dbtype": "sqlite3",
        "version": "10.1.0.0",
        "dbname": "owncloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "mysql.utf8mb4": true,
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
        "installed": true,
        "loglevel": 0,
        "maintenance": false,
        "default_language": "en",
        "skeletondirectory": "\/home\/artur\/www\/owncloud-core\/apps\/notifications\/tests\/acceptance\/..\/..\/..\/..\/tests\/acceptance\/skeleton",
        "csrf.disabled": true,
        "enable_previews": false,
        "theme": "",
        "files_external_allow_create_new_local": "true",
        "mail_domain": "***REMOVED SENSITIVE VALUE***",
        "mail_from_address": "***REMOVED SENSITIVE VALUE***",
        "mail_smtpmode": "smtp",
        "mail_smtphost": "***REMOVED SENSITIVE VALUE***",
        "mail_smtpport": "1025",
        "singleuser": false
    }
}

```

**List of activated apps:**

```
Enabled:
[...]

PR #30920, opened 2018-03-26
fix changing own email address and fullname

individual-it:
## Description
set personal email  and full name by using the ocs provisioning api

## Related Issue
#30915

## Motivation and Context
make it possible to change this values again and do not require the user management app to be enabled

## How Has This Been Tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/settings/js/panels/profile.js
+++ b/settings/js/panels/profile.js
@@ -8,20 +8,23 @@ function changeEmailAddress () {
 	}
 	emailInfo.defaultValue = emailInfo.val();
 	OC.msg.startSaving('#lostpassword .msg');
-	var post = $("#lostpassword").serializeArray();
 	$.ajax({
 		type: 'PUT',
-		url: OC.generateUrl('/settings/users/{id}/mailAddress', {id: OC.currentUser}),
-		data: {
-			mailAddress: post[0].value
-		}
+		url: OC.linkToOCS('cloud/users', 2) + encodeURIComponent(OC.currentUser) +
+			 "?format=json",
+		data: "key=email&value=" + encodeURIComponent(emailInfo.val()),
 	}).done(function(result){
-		// I know the following 4 lines look weird, but that is how it works
 		// in jQuery -  for success the first parameter is the result
 		//              for failure the first parameter is the result object
+		result.status = result.ocs.meta.status;
+		result.data = new Object();
+		result.data.message = t('settings', 'Email has been changed successfully.');
 		OC.msg.finishedSaving('#lostpassword .msg', result);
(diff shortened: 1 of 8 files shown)


phil-davis (commented):
The code looks fine, but I am not a JS and AJAX-foo expert so cannot give an informed opinion about the "structural" changes here.

Leaving for @DeepDiver1975 or someone to review.

settings/js/panels/profile.js
```diff
@@ -8,20 +8,23 @@ function changeEmailAddress () {
...
 	emailInfo.defaultValue = emailInfo.val();
 	OC.msg.startSaving('#lostpassword .msg');
-	var post = $("#lostpassword").serializeArray();
 	$.ajax({
 		type: 'PUT',
-		url: OC.generateUrl('/settings/users/{id}/mailAddress', {id: OC.currentUser}),
-		data: {
-			mailAddress: post[0].value
-		}
+		url: OC.linkToOCS('cloud/users', 2) + OC.currentUser + "?format=json",
```
PVince81: do we need to encode the user ?

AFAIK when using generateUrl it encoded automatically

settings/js/panels/profile.js
```diff
@@ -31,23 +34,29 @@ function changeEmailAddress () {
...
-				// update avatar if avatar is available
-				if(!$('#removeavatar').hasClass('hidden')) {
-					updateAvatar();
-				}
-			}
-			else {
-				$('#newdisplayname').val(data.data.displayName);
+		$.ajax({
+			type: 'PUT',
+			url: OC.linkToOCS('cloud/users', 2) + OC.currentUser + "?format=json",
```
PVince81: encode user ?

phil-davis:
Just as a thought/note: if the provisioning app is not enabled then I guess this way will also not work. In a case like that, the system admin probably is happy or expects that this data cannot be changed. So maybe the UI can just not have any update behaviour or?

individual-it:
@PVince81 in the username there should not be anything to encode, because its pretty restricted what a valid username can be "a-z", "A-Z", "0-9", and "_.@-'" but maybe some other char can come from LDAP or something.
actually encoding the full name data makes a lot of sense
Added two UI tests to check if change was applied

PVince81 (approved):
:+1:

this is likely master only

phil-davis:
Backport ``stable10`` of new acceptance test step ``checkAttributesForUser`` added in ``Provisioning.php`` is in PR #31356 - that let's us use that step elsewhere and backport others uses to ``stable10``

PR #30929, opened 2018-03-26
Put back openssl parameter in config.sample.php

phil-davis:
1) Revert previous over-zealous commit in PR #30806 
2) Just remove ``check_for_working_webdav`` and ``check_for_working_htaccess``

That puts ``openssl`` back. It is referenced in the ``encryption`` app that was recently moved into its own repo.


phil-davis:
Backport ``stable10`` #30933

DeepDiver1975 (approved):

PR #30930, opened 2018-03-26
Add symfony events for delete and create share

sharidas:
Add symfony events for delete share.

## Description

Adding symfony events for delete share. This was missing in the code base.

## Motivation and Context

Adding symfony events for delete share. This was missing in the codebase.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/tests/API/Share20OCSTest.php
+++ b/apps/files_sharing/tests/API/Share20OCSTest.php
@@ -24,7 +24,6 @@
  */
 namespace OCA\Files_Sharing\Tests\API;
 
-use JMS\Serializer\EventDispatcher\EventDispatcher;
 use OCA\Files_Sharing\API\Share20OCS;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
@@ -38,6 +37,7 @@ use OCP\IUserManager;
 use OCP\Lock\LockedException;
 use OCP\Share;
 use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 use Test\TestCase;
 
 /**
@@ -108,7 +108,7 @@ class Share20OCSTest extends TestCase {
 				['core', 'shareapi_default_permissions', \OCP\Constants::PERMISSION_ALL, \OCP\Constants::PERMISSION_READ | \OCP\Constants::PERMISSION_CREATE]
 			]));
 
-		$this->eventDispatcher = \OC::$server->getEventDispatcher();
+		$this->eventDispatcher = new EventDispatcher();
 
 		$this->ocs = new Share20OCS(
 			$this->shareManager,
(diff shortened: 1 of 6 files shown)


lib/private/Share20/Manager.php
```diff
@@ -846,6 +846,8 @@ public function deleteShare(\OCP\Share\IShare $share) {
 		// Emit pre-hook
 		\OC_Hook::emit('OCP\Share', 'pre_unshare', $hookParams);
 
+		$beforeEvent = new GenericEvent(null, ['share' => $hookParams, 'shareObject' => $share]);
+		\OC::$server->getEventDispatcher()->dispatch('share.beforedelete', $beforeEvent);
```
PVince81: please inject the event dispatcher with DI

tests/lib/Share20/ManagerTest.php
```diff
@@ -231,7 +232,26 @@ public function testDelete($shareType, $sharedWith) {
 			->method('post')
 			->with($hookListnerExpectsPost);
 
+		$calledBeforeEvent = [];
+		\OC::$server->getEventDispatcher()->addListener('share.beforedelete',
```
PVince81: let's use a new EventDispatcher from now on: `$this->eventDispatcher  = new EventDispatcher()` because reusing the existing one will pollute the server with events and will eventually cause side effects

jvillafanez:
Note that expected events are "share.afterDelete" and "share.afterCreate", note the case-sensitive. The "share.afterCreate" event is still missing.

sharidas:
@jvillafanez I have added event `share.afterCreate`.

lib/private/Share20/Manager.php
sharidas: Added to DI

tests/lib/Share20/ManagerTest.php
sharidas: Used `$this->eventDispatcher = new EventDispatcher()` instead of `\OC::$server->getEventDispatcher()`.

sharidas:
Not really getting the cause for drone failure. 
Not able to reproduce the same in my local instance : http://paste.opensuse.org/29265128.

sharidas:
Got the issue https://github.com/owncloud/core/pull/30930#issuecomment-377171193. The drone was failing because of duplicate event "share.beforeCreate" and "share.afterCreate", at :
- https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/API/Share20OCS.php#L477
- https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/API/Share20OCS.php#L498

sharidas:
Updated this PR, by moving the `share.afterCreate` and `share.beforeCreate` events to share manager. Have also tried to grab the data from the Share20OCS to Share20/Manager.php.

apps/files_sharing/tests/API/Share20OCSTest.php
```diff
@@ -108,7 +108,8 @@ protected function setUp() {
 				['core', 'shareapi_default_permissions', \OCP\Constants::PERMISSION_ALL, \OCP\Constants::PERMISSION_READ | \OCP\Constants::PERMISSION_CREATE]
 			]));
 
-		$this->eventDispatcher = \OC::$server->getEventDispatcher();
+		//$this->eventDispatcher = \OC::$server->getEventDispatcher();
```
jvillafanez: unnedeed comment

jvillafanez:
Taking into account that the events will be thrown by core, I don't know if we should create custom public events instead of using the GenericEvent one.

Other than that, the code looks fine

PVince81:
> I don't know if we should create custom public events instead of using the GenericEvent one.

I'm also undecided here. In many places we already use GenericEvent so let's stick with it when it works.

lib/private/Share20/Manager.php
```diff
@@ -635,6 +640,9 @@ public function createShare(\OCP\Share\IShare $share) {
 		];
 		\OC_Hook::emit('OCP\Share', 'pre_shared', $preHookData);
 
+		$beforeEvent = new GenericEvent(null, ['shareData' => $preHookData]);
+		$this->eventDispatcher->dispatch('share.beforeCreate', $beforeEvent);
```
PVince81: this would break any listener that was relying on the old formatter pre-hook data.

this is only ok if the beforeCreate change was never released before, else this is an API change

PVince81: ok, I see that this event did not exist in 10.0.7.

this makes this change critical for 10.0.8 as we don't want to introduce the old event in 10.0.8 and change it again

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport, p1-urgent because the current stable10 code would introduce the old hook format and this PR adjusts it to the correct one.

PR #30947, opened 2018-03-28
Replace emittingCall with separate before and after events

sharidas:
The call to emittingCall was causing issues when
clients like android are trying to upload files.
Which were triggering after events even when the
upload operation was not completed. Replacing
emittingCall would explicitely call the before
and after events when required.

## Description

Remove `emittingCall` in this method. It was found that before completion of the upload, the after event was emitted. Which is not correct. Hence called the before and after events separately. To be specific the return https://github.com/owncloud/core/blob/955d0df3f552663aaff22fd10b95ee54d98bd6bd/apps/dav/lib/Connector/Sabre/File.php#L132 was the cause, to raise this PR.

## Motivation and Context

Remove `emittingCall` in this method. It was found that before completion of the upload, the after event was emitted. Which is not correct. Hence called the before and after events separately.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -112,150 +112,149 @@ class File extends Node implements IFile, IFileNode {
 	 */
 	public function put($data) {
 		$path = $this->fileView->getAbsolutePath($this->path);
-		return $this->emittingCall(function () use (&$data) {
+		$beforeEvent = new GenericEvent(null, ['path' => $path]);
+		\OC::$server->getEventDispatcher()->dispatch('file.beforecreate', $beforeEvent);
+		try {
+			$exists = $this->fileView->file_exists($this->path);
+			if ($this->info && $exists && !$this->info->isUpdateable()) {
+				throw new Forbidden();
+			}
+		} catch (StorageNotAvailableException $e) {
+			throw new ServiceUnavailable("File is not updatable: " . $e->getMessage());
+		}
+
+		// verify path of the target
+		$this->verifyPath();
+
+		// chunked handling
+		if (\OC_FileChunking::isWebdavChunk()) {
 			try {
-				$exists = $this->fileView->file_exists($this->path);
-				if ($this->info && $exists && !$this->info->isUpdateable()) {
-					throw new Forbidden();
-				}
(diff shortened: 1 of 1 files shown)


sharidas:
- Tried to add event dispatcher as DI, but that was causing changes to many places in the codebase. That is the reason I used the approach in this PR.
- Another issue I find is how to update the unit test.

PVince81:
@sharidas could this kind of problem happen to all other usages of the emitter trait ?

PVince81:
@sharidas was this code released already ? if yes, this needs to be tagged as regression and p1

sharidas:
@PVince81 Yes it was released. I will check other places before we move further.

PVince81:
because if the problem is inherently related to EmitterTrait then we need to fix EmitterTrait

sharidas:
> because if the problem is inherently related to EmitterTrait then we need to fix EmitterTrait

I don't think its a problem is with EmitterTrait. EmitterTrait would trigger after event if return from the closure is 
- [x]  not *false*
- [x] not *null*
- [x] and if the after event has array content to propagate.

The real problem is combining EventEmitterTrait with long methods ( long methods -> large volume of code in a method ). For instance with method we faced this issue like `put`. The return from this method, if its not a `null` or `false`, the after event will be triggered. So I think it was my mistake not forseeing the return which was there in the method `put`. However I believe when dealing with functions like this, its better not to go with EventEmitterTrait.

PVince81:
>  So I think it was my mistake not forseeing the return which was there in the method put. However I believe when dealing with functions like this, its better not to go with EventEmitterTrait.

Thanks for checking, makes sense

PVince81:
Looking at the diff I don't really see how it's solving the problem: https://github.com/owncloud/core/pull/30947/files?w=1

What code path was causing the "after" event to be triggered in case of failed upload ? Was it an exception ?

sharidas:
> What code path was causing the "after" event to be triggered in case of failed upload ? Was it an exception ?

This was the return that was triggering the after event : https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/File.php#L131

PVince81 (approved):
:+1: please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/30986

Issue #30951, opened 2018-03-28
occ command to adjust filecache checksum


PVince81:
In some special cases (not reproducible) it can happen that the checksum in oc_filecache doesn't match the data on disk. This causes the desktop client to refuse working with said file.

If an admin can confirm that the data on disk is correct, they need a way to reset the checksums.

For this we should provide an occ command that scans through all files and detects wrong checksums. This could likely take a very long time as it needs to open all files to recompute the checksums ! I suggest to add a `--path` parameter so an admin can provide a more focussed scan (see `--path` argument like for occ files:scan) when they know in what folder most broken checksums are located.

Another alternative would be to only delete the wrong checksums and let the server recompute during the next download.

@IljaN @jnweiger @felixboehm

PVince81:
Note: when user-key encryption is enabled the command can only work if a recovery key is specified. Else there is no way for the occ command to decrypt the files to compute the checksums to verify them. (see how decrypt-all).

For the version iteration I suggest ignoring user-key encryption.

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/24494 (occ command to clear all locks), https://github.com/owncloud/core/pull/30177 (Kill occ command maintenance:createjs), https://github.com/owncloud/core/issues/25970 (occ command to migrate data folder), https://github.com/owncloud/core/issues/28940 (add occ command to flush distributed cache), and https://github.com/owncloud/core/issues/23380 (Occ command not working).

IljaN:
- [ ] Should also work with ObjectStore

IljaN:
Could not test with objectstore because of: https://github.com/owncloud/objectstore/issues/88

PVince81:
where is the PR ? please link

PVince81:
found it, PR here: https://github.com/owncloud/core/pull/31008

IljaN:
Forward Port: https://github.com/owncloud/core/pull/31082

PR #30967, opened 2018-03-29
[stable10] Only reset checksum if metadata differs

IljaN:
## Description
During a cache-scan, cache metadata and file(system) metadata is compared. As the files-system has
no checksum info the  checksum field is empty thus the scanner code wrongly assumes that the checksum has changed and resets the checksum field in file-cache.

## How Has This Been Tested?
Manually

## Types of changes

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Checksum.php
+++ b/lib/private/Files/Storage/Wrapper/Checksum.php
@@ -190,7 +190,12 @@ class Checksum extends Wrapper {
 				return null;
 			}
 		}
-		$parentMetaData['checksum'] = self::getChecksumsInDbFormat($path);
+
+		$checksums = self::getChecksumsInDbFormat($path);
+
+		if (!empty($checksums)) {
+			$parentMetaData['checksum'] = $checksums;
+		}
 
 		if (!isset($parentMetaData['mimetype'])) {
 			$parentMetaData['mimetype'] = 'application/octet-stream';
--- a/tests/lib/Files/Storage/Wrapper/ChecksumTest.php
+++ b/tests/lib/Files/Storage/Wrapper/ChecksumTest.php
@@ -96,7 +96,7 @@ class ChecksumTest extends \Test\TestCase
 
 		$this->assertEquals(
 			$metaData,
-			['checksum' => '', 'mimetype' => 'application/octet-stream']
+			['mimetype' => 'application/octet-stream']
 		);
 	}
 


lib/private/Files/Cache/Scanner.php
```diff
@@ -205,8 +205,12 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 						$fileId = -1;
 					}
 					if (!empty($newData)) {
-						// Reset the checksum if the data has changed
-						$newData['checksum'] = '';
+						$onlyChecksumDiffers = count($newData) == 1 && isset($newData['checksum']) && $data['checksum'] === '';
```
PVince81: `$newData['checksum']` is likely never set since we never read checksums directly from the FS as it's mostly not supported ?

PVince81:
We need to clear the checksum if a file has been modified remotely on an external storage. Detecting this usuallly happens by looking at the mtime of the file and the parent of the folder. (this logic exists in the watcher class). We cannot recompute the checksum on the fly by redownloading said file from external storage as it would be expensive, so we just clear it here.

Did you find other cases where the checksum would be cleared that are not related to modified files on external storages ? Please retest with such external storage use case.

Also please add unit tests when clarified.

IljaN:
@PVince81 Thing here is that checksum is always cleared if filesystem returns no checksum, which is always because we don`t read checksum data from the filesystem (ZFS,...)

IljaN:
I use the Scanner as basis for #30951

PVince81:
> Thing here is that checksum is always cleared if filesystem returns no checksum, which is always because we don`t read checksum data from the underlying filesystem

yes, so what is this PR for if we never read checksum from disk ? clearing the checksum is then expected, given that mtime has changed

IljaN:
@PVince81 Because of this line https://github.com/owncloud/core/blob/ae461921077383661b1a44b24aac948a70e85255/lib/private/Files/Cache/Scanner.php#L202 the checksum is always reset regardless of mtime changes. The array is always at least ['checksum' => ''] thus newData is never emtpy.

PVince81:
we should rather find out the reason why the checksum value is empty string when it reaches here and fix the root cause

IljaN:
@PVince81 Okay I will investigate, but I assumed that it is "real" Filesystem vs. Filecache metadata.

IljaN:
Empty checksum key comes from here: https://github.com/owncloud/core/blob/96c2d103a8cdb351040285b0b6b692ea65a7bb7f/lib/private/Files/Storage/Wrapper/Checksum.php#L193

Thinking at that time was that the key should always be set to facilitate isset free code at the consumer side.

lib/private/Files/Storage/Wrapper/Checksum.php
```diff
@@ -190,7 +190,12 @@ public function getMetaData($path) {
 				return null;
 			}
 		}
-		$parentMetaData['checksum'] = self::getChecksumsInDbFormat($path);
+
```
IljaN: Let`s see if this passes CI...

jvillafanez (approved):
Changes make sense

IljaN:
This fix is wrong, correct fix done in: #30951

Issue #30897, opened 2018-03-25
webUI test suite webUISharingExternal failing on Travis


phil-davis:
It fails in all browsers with some problem where the output gets stuck:
```
Running tests on 'chrome' (latest) on Windows 10
@webUI @insulated @disablePreviews @TestAlsoOnExternalUserBackend
Feature: Federation Sharing - sharing with users on other cloud storages
  As a user
  I want to share files with any users on other cloud storages
  So that other users have access to these files
  Background:                                                                            # /home/[secure]/build/owncloud/core/tests/acceptance/features/webUISharingExternal/federationSharing.feature:7
    Given these users have been created:                                                 # FeatureContext::theseUsersHaveBeenCreated()
      | username | password | displayname | email        |
      | user1    | 1234     | User One    | u1@oc.com.np |
      | user2    | 1234     | User Two    | u2@oc.com.np |
    And the user has browsed to the login page                                           # WebUILoginContext::theUserBrowsesToTheLoginPage()
    And the user has logged in with username "user2" and password "1234" using the webUI # WebUILoginContext::theUserLogsInWithUsernameAndPasswordUsingTheWebUI()
  Scenario: test the single steps of federation sharing                                                               # /home/[secure]/build/owncloud/core/tests/acceptance/features/webUISharingExternal/federationSharing.feature:15
    When the user shares the folder "simple-folder" with the remote user "user1@%remote_server%" using the webUI      # WebUISharingContext::theUserSharesTheFileFolderWithTheUserUsingTheWebUI()
    And the user shares the folder "simple-empty-folder" with the remote user "user1@%remote_server%" using the webUI # WebUISharingContext::theUserSharesTheFileFolderWithTheUserUsingTheWebUI()
    And the user re-logs in with username "user1" and password "1234" to "http://%remote_server%" using the webUI     # WebUILoginContext::theUserRelogsInWithUsernameAndPasswordToUsingTheWebUI()
    Then dialogs should be displayed on the webUI                                                                     # WebUIGeneralContext::dialogsShouldBeDisplayedOnTheWebUI()
      | title        | content                                                                              |
      | Remote share | Do you want to add the remote share /simple-folder from user2@%local_server%/?       |
      | Remote share | Do you want to add the remote share /simple-empty-folder from user2@%local_server%/? |
    When the user accepts the offered remote shares                                                                   # WebUISharingContext::theUserAcceptsTheOfferedRemoteShares()
[...]

PR #30971, opened 2018-04-01
Properly remember groups that have been deleted using the webUI

phil-davis:
## Description
``createdGroups`` array is actually keyed on the group name, so use ``array_key_exists()`` to see if the entry exists. If it does, then when a group has been deleted, mark it ``shouldHaveBeenCreated`` ``false`` - that stops the ``AfterScenario`` code from complaining when/if it cannot delete the group.

## Related Issue
#30897 

## Motivation and Context
webUI tests that delete groups on the webUI users page, they are not properly realizing that the group has been deleted. So in the ``AfterScenario`` that cleans up created groups, they complain with output like:
```
INFORMATION: could not delete group. '0'400 <?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>400</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```
There is actually nothing wrong.
It would be good to stop such distracting noise in acceptance test output.
This accidentally got "broken" when the structure of the ``createdGroups`` array was changed recently.

## How Has This Been Tested?
Run the scenario:
```
bash tests/travis/start_ui_tests.sh --feature tests/acceptance/features/webUIManageUsersGroups/manageGroups.feature:12
```
from ``stable10`` (this scenario is in ``core`` ``stable10`` and in ``user_management`` ``master``)
and confirm that the information message(s) are no longer emitted.

## Types of changes

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -122,11 +122,11 @@ trait Provisioning {
 	 */
 	public function theUserShouldExist($username) {
 		if (array_key_exists($username, $this->createdUsers)) {
-			return $this->createdUsers[$username]['shouldHaveBeenCreated'];
+			return $this->createdUsers[$username]['shouldExist'];
 		}
 
 		if (array_key_exists($username, $this->createdRemoteUsers)) {
-			return $this->createdRemoteUsers[$username]['shouldHaveBeenCreated'];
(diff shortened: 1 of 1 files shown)


phil-davis:
Backport ``stable10`` #30972

individual-it (approved):
:+1: if CI passes here and acceptance tests work in user_management

PR #30974, opened 2018-04-02
Silence retried to delete file messages

phil-davis:
## Description
If the scenario step does not expect the file to be able to be deleted, then do not emit messages about "retried to delete". In this case we know that we are supposed to retry a bit, and then it is expected that we will give up and not be able to delete the file.

In this case, reduce the number of retries so we do not waste so much time rechecking that a delete really cannot be done.

## Related Issue
#30897 
This "cleanup" is to reduce "noise" in the test output, so that people do not mis-understand what is "normal" and what is a problem.

## Motivation and Context
When a test tries to delete a file, and expects that it should not be possible to delete the file, then actually the "Delete" action does not exist in the file actions menu. That is expected and what is being tested for. But while doing the checks, the code emits messages like:
```
INFORMATION: retried to delete file
Error while deleting file
```
These are disturbing to read in the test output, and can send you off looking for problems in the wrong place, when actually this is expected to happen for these test scenario steps.

When we are doing this kind of "negative" "cannot delete" test, we should not emit this noise.

## How Has This Been Tested?
Run a full set of webUI tests on Travis. Inspect the logs and see that these ``INFORMATION:`` messages no longer appear.

## Types of changes

## Checklist:

### DIFF
--- a/tests/acceptance/features/lib/FilesPageBasic.php
+++ b/tests/acceptance/features/lib/FilesPageBasic.php
@@ -272,7 +272,12 @@ abstract class FilesPageBasic extends OwncloudPage {
 	 *
 	 * @return void
 	 */
-	public function deleteFile($name, Session $session, $maxRetries = STANDARDRETRYCOUNT) {
+	public function deleteFile(
+		$name,
+		Session $session,
+		$expectToDeleteFile = true,
+		$maxRetries = STANDARDRETRYCOUNT
+	) {
 		$this->initAjaxCounters($session);
 		$this->resetSumStartedAjaxRequests($session);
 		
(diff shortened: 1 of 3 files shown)


phil-davis:
Backport ``stable10`` #30975

tests/acceptance/features/lib/FilesPageBasic.php
```diff
@@ -272,7 +272,7 @@ public function openFile($name, Session $session) {
 	 *
 	 * @return void
 	 */
-	public function deleteFile($name, Session $session, $maxRetries = STANDARDRETRYCOUNT) {
+	public function deleteFile($name, Session $session, $expectToDeleteFile = true, $maxRetries = STANDARDRETRYCOUNT) {
```
individual-it: long line makes it harder to review in GitHub

individual-it (changes requested):

tests/acceptance/features/lib/FilesPageBasic.php
phil-davis: reformat pushed

individual-it (approved):

PR #30976, opened 2018-04-02
Fix webUI acceptance test stalls when checking file content

phil-davis:
## Description
During federated tests, the ``webUI`` parts switch between being logged in on the "local" and then the "remote" server. The user on the "remote" server does some actions (e.g. they have uploaded a new file, delete a file, overwritten a file or whatever). Then we sometimes test "underneath" with the API to see if the file changes happened from the point-of-view of that user. 

Currently (since refactoring ``api`` and ``webUI`` acceptance tests together) those "underneath" tests are actually checking the cloud file state by hitting the API on the "local" server. On ``drone`` with Apache we have gotten away with this. On Travis (or running locally) with the PHP dev server, this is causing a hang.

Actually we really should be checking the cloud file state by using the API on the "remote" server.

This change updates the necessary step code so that we can say whether to check the state of a file on the "local" or "remote" server.

## Related Issue
#30897 

## Motivation and Context
As described in the issue, the ``webUISharingExternal`` suite hangs. This can be reproduced locally also when running the suite with 2 PHP dev servers on different ports (for "local" and "remote/federated" cloud instances).

``drone`` uses Apache server instances and does not hang, so I suspect the problem is with some locking issue when having 2 PHP dev servers running that are both actually manipulating the same ownCloud files and database backend. Ideally we would not do that, but we do! Since the fix is a reasonable correction anyway, we should fix it.

## How Has This Been Tested?
Run a full set of webUI tests on Travis in ``stable10`` and confirm that the ``webUISharingExternal`` suite now runs to completion and passes.

## Types of changes
- [x] Acceptance test fix

## Checklist:


phil-davis:
Backport ``stable10`` #30977

phil-davis:
The 2nd commit standardizes the ``substituteInLineCodes()`` ways that ``remote_server`` etc can be mentioned.
Add the ability to mention ``remote_server_without_scheme`` etc. That is useful when typing to share with a federated user - it is more natural for a user to type ``user42@localhost:8180`` than type ``user42@http://localhost:8180`` (but actually both ways do work)

Note that something happens when running the tests in Firefox and if a federated share is attempted to a string like ``user42@http://localhost:8180`` - this 2nd commit here fixes that by going back to the ordinary ``user42@localhost:8180`` format in the "share with" box.

This code has been run with a full set of Travis Firefox jobs, and it all passes.

individual-it (approved):

PR #30978, opened 2018-04-02
[test] [stable10] try newer firefox selenium

phil-davis:
See if newer selenium versions will work with Firefox 57.0


phil-davis:
``seleniumVersion`` 3.4.4 does not seem to exist - at least not in SauceLabs.

PR #30980, opened 2018-04-02
[test] [stable10] add remote server no scheme

phil-davis:
Standardize the ``substituteInLineCodes()`` ways that ``remote_server`` etc can be mentioned.
Add the ability to mention ``remote_server_without_scheme`` etc. That is useful when typing to share with a federated user - it is more natural for a user to type ``user42@localhost:8180`` than type ``user42@http://localhost:8180`` (but actually both ways do work)

Run the Travis webUI suites in firefox to make sure this at least works there!


phil-davis:
works on Firefox - great.

PR #30986, opened 2018-04-03
[stable10] Replace emittingCall with separate before and after events

sharidas:
The call to emittingCall was causing issues when
clients like android are trying to upload files.
Which were triggering after events even when the
upload operation was not completed. Replacing
emittingCall would explicitely call the before
and after events when required.

## Description

Remove emittingCall in this method. It was found that before completion of the upload, the after event was emitted. Which is not correct. Hence called the before and after events separately. To be specific the return https://github.com/owncloud/core/blob/955d0df3f552663aaff22fd10b95ee54d98bd6bd/apps/dav/lib/Connector/Sabre/File.php#L132 was the cause, to raise this PR.

## Motivation and Context

Remove `emittingCall` in this method. It was found that before completion of the upload, the after event was emitted. Which is not correct. Hence called the before and after events separately.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -111,162 +111,161 @@ class File extends Node implements IFile {
 	 */
 	public function put($data) {
 		$path = $this->fileView->getAbsolutePath($this->path);
-		return $this->emittingCall(function () use (&$data) {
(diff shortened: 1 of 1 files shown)


sharidas:
Back ported from PR: https://github.com/owncloud/core/pull/30947

PVince81 (approved):
:+1:

PR #30989, opened 2018-04-03
Remove pagination from the repair step

sharidas:
The pagination is not required as the
rows are deleted during the operation.

## Description

The pagination should be removed from this repair step. Because if the pagination is there, then the changes will not be taken after the first removal in the loop. The second iteration might omit changes which have broken links.

## Related Issue

https://github.com/owncloud/core/pull/30653#issuecomment-378194758

## Motivation and Context

If the rows are deleted on the fly, then its better to avoid pagination and use the pagelimit .

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Repair/RepairOrphanedSubshare.php
+++ b/lib/private/Repair/RepairOrphanedSubshare.php
@@ -34,16 +34,20 @@ class RepairOrphanedSubshare implements IRepairStep {
 	/** @var  IQueryBuilder */
 	private $missingParents;
 
-	/** @var  IQueryBuilder */
+	/** @var IQueryBuilder  */
 	private $deleteOrphanReshares;
+
+	/** @var  int */
+	private $pageLimit;
 
 	/**
 	 * RepairOrphanedSubshare constructor.
 	 *
 	 * @param IDBConnection $connection
 	 */
-	public function __construct(IDBConnection $connection) {
+	public function __construct(IDBConnection $connection, $pageLimit = 1000) {
 		$this->connection = $connection;
+		$this->pageLimit = $pageLimit;
 
 		//This delete query deletes orphan shares whose parents are missing
 		$this->deleteOrphanReshares = $this->connection->getQueryBuilder();
@@ -51,19 +55,19 @@ class RepairOrphanedSubshare implements IRepairStep {
 			->delete('share')
 			->where(
 				$this->deleteOrphanReshares->expr()->eq('parent',
-					$this->deleteOrphanReshares->createParameter('parentId')));
-
-
+					$this->deleteOrphanReshares->createParameter('parentId'))
+			);
 		//Set the query to get the parent id's missing from the table.
 		$this->missingParents = $this->connection->getQueryBuilder();
 		$qb = $this->connection->getQueryBuilder();
 
-		$qb->select('id')
-			->from('share');
+		$qb->select('s2.id')
+			->from('share', 's2')
+			->where($this->missingParents->expr()->eq('s2.id', 's1.parent'));
 		$this->missingParents->select('parent')
-			->from('share');
+			->from('share', 's1');
 		$this->missingParents->where($this->missingParents->expr()->isNotNull('parent'));
-		$this->missingParents->andWhere($this->missingParents->expr()->notIn('parent', $this->missingParents->createFunction($qb->getSQL())));
+		$this->missingParents->andWhere($this->missingParents->createFunction('NOT EXISTS (' . $qb->getSQL() . ')'));
 		$this->missingParents->groupBy('parent');
 	}
 
@@ -84,24 +88,21 @@ class RepairOrphanedSubshare implements IRepairStep {
 	 * @throws \Exception in case of failure
 	 */
 	public function run(IOutput $output) {
-		$pageLimit = 1000;
-		$paginationOffset = 0;
-		$deleteParents = [];
 
 		/** A one time call to get missing parents from oc_share table */
 		do {
-			$this->missingParents->setMaxResults($pageLimit);
-			$this->missingParents->setFirstResult($paginationOffset);
+			$this->missingParents->setMaxResults($this->pageLimit);
 			$results = $this->missingParents->execute();
 			$rows = $results->fetchAll();
 			$results->closeCursor();
 			if (count($rows) > 0) {
-				$paginationOffset += $pageLimit;
+				$this->connection->beginTransaction();
 				foreach ($rows as $row) {
 					$this->deleteOrphanReshares->setParameter('parentId', $row['parent'])->execute();
 				}
+				$this->connection->commit();
 			}
 
-		} while (empty($rows));
+		} while (!empty($rows));
 	}
 }
--- a/lib/private/Repair/RepairOrphanedSubshare.php
+++ b/lib/private/Repair/RepairOrphanedSubshare.php
@@ -47,28 +47,6 @@ class RepairOrphanedSubshare implements IRepairStep {
 	 */
 	public function __construct(IDBConnection $connection, $pageLimit = 1000) {
 		$this->connection = $connection;
-		$this->pageLimit = $pageLimit;
-
-		//This delete query deletes orphan shares whose parents are missing
-		$this->deleteOrphanReshares = $this->connection->getQueryBuilder();
-		$this->deleteOrphanReshares
-			->delete('share')
-			->where(
-				$this->deleteOrphanReshares->expr()->eq('parent',
-					$this->deleteOrphanReshares->createParameter('parentId'))
-			);
-		//Set the query to get the parent id's missing from the table.
-		$this->missingParents = $this->connection->getQueryBuilder();
-		$qb = $this->connection->getQueryBuilder();
-
-		$qb->select('s2.id')
-			->from('share', 's2')
-			->where($this->missingParents->expr()->eq('s2.id', 's1.parent'));
-		$this->missingParents->select('parent')
-			->from('share', 's1');
-		$this->missingParents->where($this->missingParents->expr()->isNotNull('parent'));
-		$this->missingParents->andWhere($this->missingParents->createFunction('NOT EXISTS (' . $qb->getSQL() . ')'));
-		$this->missingParents->groupBy('parent');
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


PVince81:
Please add a unit test that proves that it works. The test should fail without this commit.

You could add an attribute to set pagination to smaller values if it makes it easier to test.

PVince81:
- [ ] add unit tests

sharidas:
Added unit tests:
1) One of the test proves that if pagination is there then the code would fail to remove all orphan shares (i.e, without this change).
2) Another test to prove that with this change all orphan shares could be removed.

Will see how the tests progress in the CI.

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,315 @@ public function testLargeSharesWithOrphans() {
...
+		var_dump($results);
+		$this->assertEquals($totalParents - 2500, $result);
+	}
+
+	/**
+	 * This test comprises of large rows i.e, 12000 rows.
+	 * This test will have more orphan shares and will check if the repair
+	 * test is working properly.
+	 */
+	public function testOrphanSharesFailsWithPagination() {
```
PVince81: the test is only testing pagination itself, not a failure

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,298 @@ public function testLargeSharesWithOrphans() {
...
+		$result = 0;
+		do {
+			$statement = $qb->select('parent')
+				->from('share')
+				->groupBy('parent')->orderBy('parent')->setMaxResults($pageLimit)->setFirstResult($offset)->execute();
+			$results = $statement->fetchAll();
+			$offset += $pageLimit;
+			$result += count($results);
+		} while(count($results) > 0);
+		$this->assertNotEquals($totalParents - 2500, $result);
```
sharidas: @PVince81 Isn't this assert enough to prove that if pagination is used, then it will not remove all the orphans subshares?

PVince81: would be better to have a check with "equals". since you have hardcoded values in your test above you already know the exact number of expected deletions. just use that in the assertion

PVince81: another assertion would be to check how many orphaned parents are left: after fixing I expect this value to be always zero, even with pagination

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -84,24 +76,27 @@ public function getName() {
...
-			$this->missingParents->setFirstResult($paginationOffset);
+			$this->missingParents->setMaxResults($this->pageLimit);
 			$results = $this->missingParents->execute();
 			$rows = $results->fetchAll();
 			$results->closeCursor();
 			if (count($rows) > 0) {
-				$paginationOffset += $pageLimit;
+				$this->connection->beginTransaction();
+				//Set the delete command
+				$deleteStatement = 'DELETE FROM `*PREFIX*share`  WHERE `parent` = ?';
```
PVince81: please keep `$this->deleteOrphanedReshares`

PVince81:
Let's try something: in `$this->missingParents->andWhere($this->missingParents->expr()->notIn('parent', $this->missingParents->createFunction($qb->getSQL())));` replace notIn with an EXISTS statement:
`and not exists (select s2.id from share s2 where s2.id=s1.parent)`

make that into a query builder statement.

Maybe that will solve the Oracle issue because then we don't do a select on the full table in a subquery.

lib/private/Repair/RepairOrphanedSubshare.php
```diff
@@ -34,36 +34,40 @@ class RepairOrphanedSubshare implements IRepairStep {
...
-			->from('share');
+		$qb->select('s2.id')
+			->from('share', 's2')
+			->where($this->missingParents->expr()->eq('s2.id', 's1.parent'));
 		$this->missingParents->select('parent')
-			->from('share');
+			->from('share', 's1');
 		$this->missingParents->where($this->missingParents->expr()->isNotNull('parent'));
-		$this->missingParents->andWhere($this->missingParents->expr()->notIn('parent', $this->missingParents->createFunction($qb->getSQL())));
+		$this->missingParents->andWhere($this->missingParents->createFunction('NOT EXISTS (' . $qb->getSQL() . ')'));
```
sharidas: Updated query builder from -> https://github.com/owncloud/core/pull/30989#issuecomment-378632091.

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,330 @@ public function testLargeSharesWithOrphans() {
...
+		//Now lets randomly delete 100 folders of admin to create orphan shares
+		//We would remove 20 id's from oc_share
+		$rowId = 2500;
+		$randomRow = 1;
+		while($rowId > 0) {
+			$qb = $this->connection->getQueryBuilder();
+			//Check if the row is there before deleting
+			$result = $qb->select('id')
+				->from('share')
+				->where($qb->expr()->eq('id', $qb->createNamedParameter($randomRow + $pareReshareCountRest)))
```
PVince81: this is very unreliable and will likely not work on Oracle: please store the ids of all inserted entries from the above query and then delete a few selected ones.

you cannot assume that the share ids will start at 1 here as not all DBs reset their autoincrement to zero

please implement this in a proper way.

also see https://github.com/owncloud/documentation/issues/3805#issuecomment-378827785 when retrieving inserted ids

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,330 @@ public function testLargeSharesWithOrphans() {
...
+							'parent' => $qb->expr()->literal($parent),
+							'item_type' => $qb->expr()->literal('folder'),
+							'item_source' => $qb->expr()->literal($itemSource),
+							'file_source' => $qb->expr()->literal($itemSource),
+							'file_target' => $qb->expr()->literal('/' . $itemSource),
+							'permissions' => $qb->expr()->literal(31),
+							'stime' => $qb->expr()->literal($time),
+						])
+						->execute();
+					if (($parentReshareCount - $pareReshareCountRest) >= 3000) {
```
PVince81: why ?

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,330 @@ public function testLargeSharesWithOrphans() {
...
+				continue;
+			}
+			$qb->delete('share')
+				->where($qb->expr()->eq('id', $qb->createNamedParameter($randomRow + $pareReshareCountRest)))
+				->execute();
+			$rowId--;
+			$randomRow++;
+		}
+
+		//Now run the repair step and prove that this test fails
```
PVince81: no, we don't prove that a test fail. a test must always pass.

what I said about failing tests is based on Test Driven Development (TDD) where you first write a test that reproduces an issue. the test will fail at first and then you implement the code to make it pass.

it doesn't mean that the test exists to prove that something fails

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,330 @@ public function testLargeSharesWithOrphans() {
...
+			$results = $statement->execute()->fetchAll();
+			$result += count($results);
+			$this->assertEquals($totalParents - 2500, $result);
+		}
+
+		//The value of $testFail is set true when exception is caught.
+		$this->assertTrue($testFail);
+	}
+
+	public function mockRepairOrphanShareMethodWithPagination() {
```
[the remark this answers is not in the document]
PVince81: for example if your test code destroyed the parents of the share ids 5, 15, 18 then in your assertion code after running the repair step, you'll query whether the sub-shares with parents 5, 15 and 18 are gone.

this is better than using count. Maybe we should move away from just counting as it's not accurate enough.

sharidas:
We need to consider this PR: https://github.com/owncloud/core/pull/31014.

sharidas:
Added commit: https://github.com/owncloud/core/pull/31014 to this PR.

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,128 @@ public function testLargeSharesWithOrphans() {
 		} while(count($results) > 0);
 		$this->assertEquals($totalParents - 20, $result);
 	}
+
+	/**
+	 * A new approach
```
PVince81: the approach is not new, this is "the" approach

if you want to write "new approach" the text needs to be in the PR comment

please replace the comment with whatever the test is testing

what does "no more pagination" means ? no more compared to what ?

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,128 @@ public function testLargeSharesWithOrphans() {
...
+		//Lets check range of 2900 to 2910
+		foreach (range(2900, 2910) as $adminIndex) {
+			$row = $checkQuery->select('parent')
+				->from('share')->where($checkQuery->expr()->eq('id', $checkQuery->createNamedParameter($getAllIdsPerUser['admin'][$adminIndex])))
+				->execute()->fetchAll();
+			$this->assertEquals(1, count($row));
+		}
+
+		//Lets check rance of 1 to 9
+		foreach (range(1, 9) as $adminIndex) {
```
PVince81: this looks like the same loop code like line 391, maybe you can put these together

tests/lib/Repair/RepairOrphanedSubshareTest.php
```diff
@@ -286,4 +286,128 @@ public function testLargeSharesWithOrphans() {
 		} while(count($results) > 0);
 		$this->assertEquals($totalParents - 20, $result);
 	}
+
+	/**
+	 * A new approach
+	 */
+	public function testOrphanSharesNoMorePagination() {
```
PVince81: the is is not testing a code change, it's testing a behavior: what behavior are you testing ?

tests/lib/Repair/RepairOrphanedSubshareTest.php
sharidas: The intention here was to see the first 9 indexes are there or not. A proof that first 9 parents are not deleted according to the test.

sharidas: Removed the duplicated code and updated with a loop so that it look better. So if one has to check if the index is there or not in this large scenario, appending the array `$checkRandomAvailableEntries` is enough.

sharidas:
@PVince81  Ready for review.

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport as part of https://github.com/owncloud/core/pull/30653#issuecomment-378253892, all in a single PR

PR #30991, opened 2018-04-03
[stable10] A new exception to suppress exception logs

sharidas:
There are scenarios where an app would have
done the job and the logger would have logged.
But the main task would throw an exception.
Since its logged gracefully we need not throw
exception in the log. This change would help
for the same.

## Description

There can be situations where we need not throw the exceptions to the logger. For example when an antivirus app had detected the file cannot be uploaded and logged gracefully, it would be nice to exclude the exception logged.

## Related Issue

https://github.com/owncloud/core/issues/30751

## Motivation and Context

There can be situations where we need not throw the exceptions to the logger. For example when an antivirus app had detected the file cannot be uploaded and logged gracefully, it would be nice to exclude the exception logged.

## How Has This Been Tested?

- [x] Uploaded the eicar test file and the exception is not logged.

## Types of changes

## Checklist:


sharidas:
Backported from PR: https://github.com/owncloud/core/pull/30771

PVince81 (approved):
:+1:

PR #30992, opened 2018-04-03
Fix multiple CI failures on master

PVince81:
Fixes errors that sneaked it while CI was not looking

### DIFF
--- a/lib/public/Util/UserSearch.php
+++ b/lib/public/Util/UserSearch.php
@@ -43,6 +43,7 @@ class UserSearch {
 	 * UserSearch constructor.
 	 *
 	 * @param IConfig $config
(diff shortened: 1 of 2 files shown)


VicDeo (approved):
Thanks!

PVince81:
still fails, now with Oracle... might still be a random thing

phil-davis:
It is going to be:
https://github.com/owncloud/core/blob/master/tests/lib/Repair/RepairOrphanedSubshareTest.php#L255
when choosing 20 rows to "randomly" delete, then sometimes it can choose the same row twice. Then it actually only deletes 19 rows. So the count of remaining rows ends up 1 larger than expected.

Issue #30942, opened 2018-03-28
Trashbin: cannot permanently delete files with partial unselection.


supermanlinux:
### Steps to reproduce
1.create at least 2 files
2.delete these files
3.go in trashbin
4.click select all button to select these deleted files
5.unclick (unselect) one of these files at least (but not all!)
6.click Delete to definitely delete all the selected files

### Expected behaviour
All the selected files should be deleted.

### Actual behaviour
None of the files are deleted.

### Server configuration
**Database:**10.0.31-MariaDB-0ubuntu0.16.04.2 Ubuntu 16.04
**PHP version:**PHP 7.0.22
**ownCloud version:10.0.6.1
**Where did you install ownCloud from:owncloud-10.0.6.tar.bz2

supermanlinux:
@PVince81  I think it's a big bug,very seriously.Expected to fixed,thanks.

phil-davis:
Similarly:

1) Have files in the trashbin
2) Use the top checkbox to "select all"
3) Uncheck some individual file(s)
4) Click "Restore" to try and restore all the files that are still selected.

Error dialog pops up with ``Couldn't restore``

phil-davis:
The delete behavior works correctly on the main files page (which has quite different js). You can "select all" then unselect some files, then delete all the remaining selected files.

supermanlinux:
@phil-davis But in trashbin, I  "select all" then unselect some files,the restore and delete button is disabled.It's pretty annoying to delete a lot of files from the recycle bin if there are some I don't want to delete.

phil-davis:
@supermanlinux I agree - if you have a trashbin folder of 100 files, and want to restore or deleted 99 of them, you have to select each of the 99 individually.

Or you do the ugly workaround - restore the 1 file, then delete all remaining 99 from trashbin, then delete the 1 file again from the main files page. Now you have a trashbin with just that 1 file in it :)

supermanlinux:
@phil-davis This has seriously affected the user experience. Users generally do not do this. Is it difficult to fix this problem?

superstoraged:
@PVince81 I have the same problems. Any chance this will get fixed? It's pretty annoying to delete a lot of files from the recycle bin if there are some I don't want to delete.Thanks.

supermanlinux:
[...]

PR #30993, opened 2018-04-03
Fix deletion of selected files in trashbin

sharidas:
When files were selected all, and later
selective files were opted for deletion,
the deletion was not happening. The reason
for this behaviour is filenames are passed
through ajax as seen in ui. Where as that
is not the case. This change helps to fix it.

## Description

The filenames passed in the ajax were the files seen in UI. But that is not the case with trashbin files. This change fixes the issue.

## Related Issue

https://github.com/owncloud/core/issues/30942

## Motivation and Context

The filenames passed in the ajax were the files seen in UI. But that is not the case with trashbin files. This change fixes the issue.

## How Has This Been Tested?

- [x] Create 2 files. And delete them.
- [x] Select all and then change the decision to select only one. Delete the file. The file is deleted.
- [x] The same experiment is repeated with multiple files. And the files were deleted.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/js/filelist.js
+++ b/apps/files/js/filelist.js
@@ -703,16 +703,21 @@
 
 		/**
 		 * Event handler for when selecting/deselecting all files
+		 * modifiedFiles if not passed then this.files would be used.
+		 * modifiedFiles was introduced, so that if we have to alter
+		 * any fields of this.files, its better to use modifiedFiles
+		 * instead of directly making change in this.files.
 		 */
-		_onClickSelectAll: function(e) {
+		_onClickSelectAll: function(e, modifiedFiles) {
+			var filesViewed = modifiedFiles ? modifiedFiles : this.files;
 			var checked = $(e.target).prop('checked');
 			this.$fileList.find('td.filename>.selectCheckBox').prop('checked', checked)
 				.closest('tr').toggleClass('selected', checked);
 			this._selectedFiles = {};
 			this._selectionSummary.clear();
 			if (checked) {
-				for (var i = 0; i < this.files.length; i++) {
-					var fileData = this.files[i];
+				for (var i = 0; i < filesViewed.length; i++) {
+					var fileData = filesViewed[i];
 					this._selectedFiles[fileData.id] = fileData;
 					this._selectionSummary.add(fileData);
 				}
--- a/apps/files_trashbin/js/filelist.js
+++ b/apps/files_trashbin/js/filelist.js
@@ -156,6 +156,29 @@
 			this.enableActions();
 		},
 
+		/**
+		 * Event handler for when selecting/deselecting all files
+		 */
+		_onClickSelectAll: function(e) {
+			/*
+			trashbinFiles is a variable which is a clone of this.files.
+			Any change to trashbinFiles will not have any change to this.files
+			 */
+			var trashbinFiles = [];
+			for (var i = 0, len = this.files.length; i < len; i++) {
+				trashbinFiles[i] = {};
+				for (var prop in this.files[i]) {
+					trashbinFiles[i][prop] = this.files[i][prop];
+				}
+			}
+
+			for (var i = 0; i < trashbinFiles.length; i++) {
+				trashbinFiles[i].name = trashbinFiles[i].name + '.d' +
+					Math.floor(trashbinFiles[i].mtime/1000);
+			}
+			OCA.Files.FileList.prototype._onClickSelectAll.call(this, e, trashbinFiles);
+		},
+
 		_onClickRestoreSelected: function(event) {
 			event.preventDefault();
 			var self = this;
--- a/tests/acceptance/features/webUITrashbin/restore.feature
+++ b/tests/acceptance/features/webUITrashbin/restore.feature
@@ -50,6 +50,29 @@ So that I can recover accidentally deleted files/folders in ownCloud
 		But the file "data.zip" should not be listed in the files page on the webUI
 		And the folder "simple-folder" should not be listed in the files page on the webUI
 
+	Scenario: Select all except for some trashbin files and restore them in a batch
+		Given the following files have been deleted
+			| name          |
+			| data.zip      |
+			| lorem.txt     |
+			| lorem-big.txt |
+			| simple-folder |
+		And the user has browsed to the trashbin page
+		When the user marks all files for batch action using the webUI
+		And the user unmarks these files for batch action using the webUI
+			| name          |
+			| lorem.txt     |
+			| lorem-big.txt |
+		And the user batch restores the marked files using the webUI
+		Then the file "lorem.txt" should be listed on the webUI
+		And the file "lorem-big.txt" should be listed on the webUI
+		But the file "data.zip" should not be listed on the webUI
+		And the folder "simple-folder" should not be listed on the webUI
+		And the file "data.zip" should be listed in the files page on the webUI
+		And the folder "simple-folder" should be listed in the files page on the webUI
+		But the file "lorem.txt" should not be listed in the files page on the webUI
+		And the file "lorem-big.txt" should not be listed in the files page on the webUI
+
 	Scenario: Select all trashbin files and restore them in a batch
 		Given the following files have been deleted
 			| name          |
--- a/apps/files_trashbin/tests/js/filelistSpec.js
+++ b/apps/files_trashbin/tests/js/filelistSpec.js
@@ -311,6 +311,35 @@ describe('OCA.Trashbin.FileList tests', function() {
 				expect(fileList.findFileEl('somedir.d99999').length).toEqual(0);
 				expect(fileList.findFileEl('Two.jpg.d22222').length).toEqual(1);
 			});
+			it('Deletes selected files when all files selected, some unselected, and "Delete" clicked', function() {
+				var request;
+				$('.select-all').click();
+				fileList.findFileEl('Two.jpg.d22222').find('input:checkbox').click();
+				fileList.findFileEl('Three.pdf.d33333').find('input:checkbox').click();
+				$('.selectedActions .delete-selected').click();
+				expect(fakeServer.requests.length).toEqual(1);
+				request = fakeServer.requests[0];
+				expect(request.url).toEqual(OC.webroot + '/index.php/apps/files_trashbin/ajax/delete.php');
+				expect(OC.parseQueryString(request.requestBody))
+					.toEqual({'dir': '/', files: '["One.txt.d11111","somedir.d99999"]'});
+				fakeServer.requests[0].respond(
+					200,
+					{ 'Content-Type': 'application/json' },
+					JSON.stringify({
+						status: 'success',
+						data: {
+							success: [
+								{filename: 'One.txt.d11111'},
+								{filename: 'somedir.d99999'}
+							]
+						}
+					})
+				);
+				expect(fileList.findFileEl('One.txt.d11111').length).toEqual(0);
+				expect(fileList.findFileEl('somedir.d99999').length).toEqual(0);
+				expect(fileList.findFileEl('Two.jpg.d22222').length).toEqual(1);
+				expect(fileList.findFileEl('Three.pdf.d33333').length).toEqual(1);
+			});
 			it('Deletes all files when all selected when "Delete" clicked', function() {
 				var request;
 				$('.select-all').click();
@@ -355,6 +384,35 @@ describe('OCA.Trashbin.FileList tests', function() {
 				expect(fileList.findFileEl('Three.pdf.d33333').length).toEqual(0);
 				expect(fileList.findFileEl('somedir.d99999').length).toEqual(0);
 				expect(fileList.findFileEl('Two.jpg.d22222').length).toEqual(1);
+			});
+			it('Restores selected files when all files selected, some unselected, and "Restore" clicked', function() {
+				var request;
+				$('.select-all').click();
+				fileList.findFileEl('Two.jpg.d22222').find('input:checkbox').click();
+				fileList.findFileEl('Three.pdf.d33333').find('input:checkbox').click();
+				$('.selectedActions .undelete').click();
+				expect(fakeServer.requests.length).toEqual(1);
+				request = fakeServer.requests[0];
+				expect(request.url).toEqual(OC.webroot + '/index.php/apps/files_trashbin/ajax/undelete.php');
+				expect(OC.parseQueryString(request.requestBody))
+					.toEqual({'dir': '/', files: '["One.txt.d11111","somedir.d99999"]'});
+				fakeServer.requests[0].respond(
+					200,
+					{ 'Content-Type': 'application/json' },
+					JSON.stringify({
+						status: 'success',
+						data: {
+							success: [
+								{filename: 'One.txt.d11111'},
+								{filename: 'somedir.d99999'}
+							]
+						}
+					})
+				);
+				expect(fileList.findFileEl('One.txt.d11111').length).toEqual(0);
+				expect(fileList.findFileEl('somedir.d99999').length).toEqual(0);
+				expect(fileList.findFileEl('Two.jpg.d22222').length).toEqual(1);
+				expect(fileList.findFileEl('Three.pdf.d33333').length).toEqual(1);
 			});
 			it('Restores all files when all selected when "Restore" clicked', function() {
 				var request;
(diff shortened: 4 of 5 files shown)


apps/files_trashbin/js/filelist.js
```diff
@@ -215,6 +215,19 @@
 			}
 			else {
 				files = _.pluck(this.getSelectedFiles(), 'name');
+				/*
+				The getSelectedFiles gives the file name as it is seen in UI.
+				But the fact is filenames are storead as filename.dxxxxx. This
+				change helps to extract exact filename.
```
PVince81: the filenames are stored like this because you can have multiple files in trashbin with the same name from different deletions. we need to use the suffixed name when talking to the server so I doubt that this approach can work correctly

apps/files_trashbin/js/filelist.js
```diff
@@ -215,6 +215,19 @@
...
+				 */
+				var trSelectedFiles = this.$el.find('tr.selected');
+				var filesFromAttribute = [];
+				for (var i = 0; i < trSelectedFiles.length; i++) {
+					filesFromAttribute.push(trSelectedFiles[i].getAttribute('data-file'));
+				}
+				if (files !== filesFromAttribute) {
+					files = filesFromAttribute;
```
sharidas: @PVince81 This change gives the suffixed name i.e if the file name is `a.txt`, then the data passed to server will be of the form `a.txt.dxxxxxxx`. While testing with single and multiple files, I was able to see files getting deleted from trashbin. Also verified files getting deleted from `files_trashbin/files`. Yes you are right, the issue was caused because suffixed filenames were not passed to the server.

PVince81: I see, makes sense.

PVince81:
next up is adjusting unit tests.

you can run them with `make test-js-debug` and if you need the browser for stepping with a debugger, open "http://localhost:9876/debug.html" (white page) and open the browser console

phil-davis:
Selected delete in trashbin works with this code (manual test).

Selected restore does not work. I expect the same sort of code changes are needed for the code that runs when the "Restore" button is clicked.

sharidas:
I have updated the change for restore issue. Will be looking into the unit tests.

phil-davis:
@sharidas webUI tests for this are available. See the 2nd commit in #31024 

The 2 new scenarios fail against current master, and pass when your code is included - good.

Please include those tests in this PR. You will need to rebase this PR up to current master (to get the recent webUI test methods that are used). Then cherry-pick that commit c736a0726b48277973189f75bb15dafa160ab90d into this PR and push it up to GitHub.

phil-davis:
I realized that I can add the test commit myself to this PR, so I have done that.

phil-davis:
@PVince81 unit tests and webUI acceptance tests have been added here.
Ready for review.

apps/files_trashbin/js/filelist.js
```diff
@@ -172,6 +172,19 @@
...
+				 * The getSelectedFiles gives the file name as it is seen in UI.
+				 * But the fact is filenames are storead as filename.dxxxxx. This
+				 * change helps to extract exact filename.
+				 */
+				var trSelectedFiles = this.$el.find('tr.selected');
+				var filesFromAttribute = [];
+				for (var i = 0; i < trSelectedFiles.length; i++) {
+					filesFromAttribute.push(trSelectedFiles[i].getAttribute('data-file'));
+				}
+				if (files !== filesFromAttribute) {
```
PVince81: I'm not sure this comparison is accurate as it might be comparing only the pointer reference and not the contents. Please use a clearer approach for this check.

apps/files_trashbin/js/filelist.js
```diff
@@ -215,6 +228,19 @@
...
+				The getSelectedFiles gives the file name as it is seen in UI.
+				But the fact is filenames are storead as filename.dxxxxx. This
+				change helps to extract exact filename.
+				 */
+				var trSelectedFiles = this.$el.find('tr.selected');
+				var filesFromAttribute = [];
+				for (var i = 0; i < trSelectedFiles.length; i++) {
+					filesFromAttribute.push(trSelectedFiles[i].getAttribute('data-file'));
+				}
+				if (files !== filesFromAttribute) {
```
PVince81: duplicate code from above, please move to common method

apps/files_trashbin/js/filelist.js
phil-davis: Hmmm - I took the test above, copy-paste, and modified the clicking at the start and expected results. Now I have to actually learn how to do this.

phil-davis:
@PVince81 I refactored that. And also, ``_onClickDeleteSelected`` and ``_onClickRestoreSelected`` are specifically in the trashbin JS code, and they are only called to work on selected files, so they can always work on the ``data-file`` attribute. There is no need to ever do ``_.pluck(this.getSelectedFiles(), 'name')`` and so that removes the need for any comparison of the 2 things (that you commented above.

sharidas:
@phil-davis Even I was having same thought when asked to move the changes to common function. The `_pluck(this.getSelectedFiles(), 'name'` can be safely removed and no more comparison required.

apps/files_trashbin/js/filelist.js
```diff
@@ -171,7 +186,7 @@
 				};
 			}
 			else {
-				files = _.pluck(this.getSelectedFiles(), 'name');
```
PVince81: Now that I understand this PR better, I'm a bit puzzled that the "name" attribute doesn't contain the combined name because the name displayed in the UI is supposed to be "displayName". Maybe that's another bug or regression.

An alternative and potentially better fix would be to adjust this class' `elementToFile()` to add another attribute on the object containing the combined name.

phil-davis: I did acceptance and JS tests here, and just did a bit of refactoring of the JavaScript code because I thought it looked easy enough and would get this PR finished.

Now that there are classes and attributes and objects to be massaged, I will leave it to someone with real JavaScript coding skills :) @sharidas ?

At least there is a good set of JS unit tests and webUI acceptance tests, so you now easily know if a refactoring breaks some "important" behavior.

tests/acceptance/features/webUITrashbin/restore.feature
```diff
@@ -50,6 +50,29 @@ So that I can recover accidentally deleted files/folders in ownCloud
 		But the file "data.zip" should not be listed in the files page on the webUI
 		And the folder "simple-folder" should not be listed in the files page on the webUI
 
+	Scenario: Select all except for some trashbin files and restore them in a batch
```
PVince81: @phil-davis can you add a case where you delete two entries with the same name ?
basically create a folder "test", then deleted it, then another "test", delete it. they'd appear as two entries.
internally they have names with the timestamp appended (test.dXXXXXXX) but visually they look the same

apps/files_trashbin/js/filelist.js
sharidas: The problem happens when user select all and the unselect(s) from the list. In my case I tried select all. 
1) The files I had in the trash were `a.txt` and `b.txt`. Now I unselected `b.txt`. Next tried to delete `a.txt`. Below is the screenshot where name has only a.txt:
notworking

2) Without selecting select all, and if I select specifically a file: `a.txt` the name becomes:
working

I guess @PVince81 the second option, right?

sharidas:
Searched till 9.1.1 version. And this is broken in version 9.1.1.

PVince81:
so version 9.1.0 was fine ? any changes ? git bisect will tell you

PVince81:
Introducing "realname" is likely just a workaround for the real issue.
I suspect that the real issue is inside of `getSelectedFiles()` and the selection logic. So while it might work it around here it might defer the issue to a later point.

sharidas:
This is my understanding from the code, the variable `_selectedFiles` gets populated when:
- page is loaded ( page of the file list )
- when select all is done
The `_selectedFiles` populates the data under these circumstances, which is then used by `getSelectedFiles`. So lets say if the user deletes `a.txt` and `b.txt`, the files listed in the trashbin ui will be of same name. Instead of `a.txt.dxxxxx` and `b.txt.dxxxx` as in the filesystem. So in the new approach, what I have done is during select all, tried to update the files data which is used to populate `_selectedFiles`.

apps/files/js/filelist.js
```diff
@@ -692,8 +692,26 @@
 				.closest('tr').toggleClass('selected', checked);
 			this._selectedFiles = {};
 			this._selectionSummary.clear();
+			var getAllSelected = this.$el.find('tr.selected');
```
PVince81: you won't receive them all. Remember that when there are more than 20+ items on the page, only the first items are rendered. You cannot rely on the DOM. Rely on "this.files" to have the full list.

apps/files/js/filelist.js
```diff
@@ -694,6 +694,13 @@
 			this._selectionSummary.clear();
 			if (checked) {
 				for (var i = 0; i < this.files.length; i++) {
+					/*
+					If the selection is from trashbin we need to do update the
+					name of the files.
+					 */
+					if (this.$fileList.selector === "#app-content-trashbin #fileList") {
```
sharidas: A check made to figure out if the selection happens from trashbin UI. Not sure how to improvise this piece of comparison here. Intention is that no other ui should get updated name in `this.files`.

apps/files/js/filelist.js
```diff
@@ -694,6 +694,13 @@
...
+						this.files[i].name = this.files[i].name + '.d' + this.files[i].mtime/1000;
```
sharidas: The division with 1000 was borrowed after having a look at the logic here: https://github.com/owncloud/core/blob/master/apps/files/lib/Helper.php#L142

apps/files/js/filelist.js
PVince81: filelist from apps/files should not know anything about apps/files_trashbin, so we need to find a way to make it work without violating this

phil-davis:
@sharidas @PVince81 rebased and conflict resolved.

This works and has tests. So where do we go from here with the discussion about the actual code?

PVince81:
The current solution is a hack as the base FileList should not know anything about trashbin as it would be a dependency between the "files" app and "files_trashbin" app. This would introduce some technical debt.

phil-davis:
So is "somebody" = @sharidas ? designing a non-hacky way to fix this?

It seems a shame to have a working and tested fix, but not be able to release it.

sharidas:
> It seems a shame to have a working and tested fix, but not be able to release it.

@phil-davis  let me see how to get it better.

PVince81:
@sharidas sounds good.

I can also have a look if needed as I'm more familiar with this inheritance mess

apps/files/js/filelist.js
```diff
@@ -694,6 +694,16 @@
 			this._selectionSummary.clear();
 			if (checked) {
 				for (var i = 0; i < this.files.length; i++) {
+					/*
+					If the selection is from trashbin we need to do update the
+					name of the files.
+					*/
+					var tempDeletedFileName = this.files[i].name + '.d' + this.files[i].mtime/1000;
```
PVince81: the files app must not know anything about the trashbin app.
here you have added trashbin specific code

PVince81: normally when using trashbin I expect the "name" attribute to already contain the "d" suffix.
it's "displayName" that contains whatever is visible in the web UI

if your observation was that "name" did not contain the "d" then there's a bug somewhere when populating the "files" array or when reading the "data-file"

PVince81: looking at: https://github.com/owncloud/core/blob/master/apps/files_trashbin/js/filelist.js#L106

ok so maybe the "files" array doesn't contain the "d" and only the data-file attribute does when creating the row.

So while your code makes sense it's not located in the correct class as it's making the files app depend on trashbin. Maybe the trashbin class must override some of the selection related methods to make it work and inject the missing "d" part when dealing with root entries.

Or introduce new methods in "files" app class and override it in "files_trashbin" when computing the name

sharidas: Looking at the code https://github.com/owncloud/core/blob/master/apps/files/js/filelist.js#L1189 the displayName is already set. Which is why user doesn't see `'.d' + mtime/1000` along with the file names in UI. I still get a feel that it is the names passed during the `delete` and `undelete` ajax requests that need to be altered. And this computation will be done in the 
- https://github.com/owncloud/core/blob/master/apps/files_trashbin/js/filelist.js#L185
- https://github.com/owncloud/core/blob/master/apps/files_trashbin/js/filelist.js#L235
Does this look like back to square one?

apps/files_trashbin/js/filelist.js
```diff
@@ -156,6 +156,36 @@
 			this.enableActions();
 		},
 
+		/**
+		 * Event handler for when selecting/deselecting all files
+		 */
+		_onClickSelectAll: function(e) {
```
sharidas: Override selectAll in trashbin. So The only difference between files/js/filelist.js and this filelist.js is from https://github.com/owncloud/core/pull/30993/files#diff-5016b1abbde02ec31bb639507f64d604R170 to https://github.com/owncloud/core/pull/30993/files#diff-5016b1abbde02ec31bb639507f64d604R176
i.e, addition of line:
```
this.files[i].name = this.files[i].name + '.d' +
          Math.floor(this.files[i].mtime/1000);
```

apps/files_trashbin/js/filelist.js
```diff
@@ -156,6 +156,36 @@
...
+			this._selectedFiles = {};
+			this._selectionSummary.clear();
+			if (checked) {
+				for (var i = 0; i < this.files.length; i++) {
+					/*
+					The selection is from trashbin we need to do update the
+					name of the files, i.e make the filenames real so that
+					its appended with '.d' + mtime/1000
+					*/
+					this.files[i].name = this.files[i].name + '.d' +
```
PVince81: not good as this would overwrite the existing data structure contents and will cause side effects

you need to make a copy of the file data, only replace the name there, not in the original array

apps/files_trashbin/js/filelist.js
PVince81: looks better, need to find a way to prevent that much code duplication, could be done by introducing a new smaller method that can be overridden

PVince81 (changes requested):
seems we're getting closer and now I understand the actual problem better

apps/files_trashbin/js/filelist.js
```diff
@@ -156,6 +156,27 @@
...
+		/**
+		 * Event handler for when selecting/deselecting all files
+		 */
+		_onClickSelectAll: function(e) {
+			/*
+			trashbinFiles is a variable which is a clone of this.files.
+			Any change to trashbinFiles will not have any change to this.files
+			 */
+			var trashbinFiles = [];
+			this.files.forEach((v, i) => {
```
sharidas: Do we have a better way to achieve a copy of `this.file` update the name or does this look ok? Got this snippet while searching. This change does achieve the goal to not update `this.files`.

apps/files/js/filelist.js
```diff
@@ -704,15 +704,16 @@
 		/**
 		 * Event handler for when selecting/deselecting all files
 		 */
-		_onClickSelectAll: function(e) {
+		_onClickSelectAll: function(e, modifiedFiles) {
```
PVince81: Beware, the function is called from a jQuery event and I'm not sure whether the second param could have arbitrary values. Please double check.

Please update the JSDoc of the function

PVince81 (commented):
Solution looks great, some small tweak, see comment :-)

PVince81 (dismissed):

phil-davis:
@PVince81 @sharidas time to merge and backport?
It will be nice to have this little fix in 10.0.9

apps/files/js/filelist.js
```diff
@@ -703,16 +703,43 @@
...
+		 * any fields of this.files, its better to use modifiedFiles
+		 * instead of directly making change in this.files.
+		 */
+		_onClickSelectAll: function(e, modifiedFiles) {
+			var filesViewed = modifiedFiles ? modifiedFiles : this.files;
+			/**
+			 * Check if there are no arbitrary keys that got into
+			 * modifiedFiles, which can cause irregular behaviours.
+			 */
+			if (modifiedFiles && (modifiedFiles.length === this.files.length)) {
```
sharidas: Tried to verify at least the keys in `modifiedFiles` are same as `this.files`. Hope this check should be sufficient. If the keys are not same then immediate fall back to `this.files` is made at https://github.com/owncloud/core/pull/30993/files#diff-dafefd9519924bec4efa9f5af4af8e68R729

PVince81: comparing length doesn't guarantee that both arrays aren't the same and this approach looks like a hack

please find a more elegant way to solve this

sharidas:
Backport PR: https://github.com/owncloud/core/pull/31700

PVince81 (approved):
:+1:

PR #30995, opened 2018-04-03
Fix sneaky errors orphaned subshare test

phil-davis:
Keep track of how many actually got deleted.


phil-davis:
I see you got rid of the mt_rand() stuff.

PR #30996, opened 2018-04-03
Throw symfony events on group actions

jvillafanez:
## Description
Required change for https://github.com/owncloud/core/pull/30106 to discard the notification when the user is removed from the group

## Related Issue
Required for https://github.com/owncloud/core/pull/30106

## How Has This Been Tested?
Unittests included

## Types of changes

## Checklist:

### DIFF
--- a/tests/lib/App/ManagerTest.php
+++ b/tests/lib/App/ManagerTest.php
@@ -164,8 +164,8 @@ class ManagerTest extends TestCase {
 
 	public function testEnableAppForGroups() {
 		$groups = [
-			new Group('group1', [], null),
-			new Group('group2', [], null)
+			new Group('group1', [], null, $this->eventDispatcher),
+			new Group('group2', [], null, $this->eventDispatcher)
 		];
 		$this->expectClearCache();
 		$this->manager->enableAppForGroups('test', $groups);
@@ -191,8 +191,8 @@ class ManagerTest extends TestCase {
 	 */
 	public function testEnableAppForGroupsAllowedTypes(array $appInfo) {
 		$groups = [
-			new Group('group1', [], null),
-			new Group('group2', [], null)
+			new Group('group1', [], null, $this->eventDispatcher),
+			new Group('group2', [], null, $this->eventDispatcher)
 		];
 		$this->expectClearCache();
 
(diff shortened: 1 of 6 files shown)


PVince81:
you got some crazy conflicts

jvillafanez:
Hopefully fixed now. Let's see what the CI says...

tests/lib/Group/GroupTest.php
```diff
@@ -204,13 +365,19 @@ public function testRemoveUserNotInGroup() {
 		$user1->expects($this->any())->method('getUID')->willReturn('user1');
 
 		$group->removeUser($user1);
+
+		// preRemoveUser is triggered but not postRemoveUser
```
jvillafanez: This needs to be checked. The behaviour is the same as with the emitter, but it's weird that only this case throws only the "pre" event; the rest of the cases throws either both "pre" and "post" events or none of them.

I'm not sure if it's a bug or it's something intended, but I think we shouldn't trigger the "preRemoveUser" event in this scenario to keep the consistency with the rest of the events.

tests/lib/App/ManagerTest.php
```diff
@@ -236,8 +236,8 @@ public function dataEnableAppForGroupsForbiddenTypes() {
 	 */
 	public function testEnableAppForGroupsForbiddenTypes($type) {
 		$groups = [
-			new Group('group1', [], null),
-			new Group('group2', [], null)
+			new Group('group1', [], null, $this->eventDispatcher),
```
PVince81: feels a bit weird to inject something into a model like this

I wonder if we should rather listen to the old events from outside and rewire to symfony events ?

@tomneedham @butonic thoughts ?

jvillafanez: Ideally, the manager should be the one performing all the operations against the group, and the group should be a "dummy" object to hold information. I think this change is very large to be done without any preparation, and it could also break a lot of things, so it is impossible to tackle such change without planning.

Rewiring doesn't solve an architectural problem. It also feels bad that an event triggers another event which triggers another event that an app is listening to.

PVince81 (approved):
:+1: fine by me then, we're also injecting the user manager and this is a private class so should be fine

jvillafanez:
Backport in https://github.com/owncloud/core/pull/31003

PR #30999, opened 2018-04-04
webUI acceptance tests for trashbin multi delete and restore

phil-davis:
## Description
1) Add webUI trashbin acceptance tests that select multiple files, or all files, and confirm that the selected files can be restored or deleted from the trashbin.

2) Add webUI files page tests that select multiple files, or all files, and confirm that the selected files can be deleted

3) Improve the waiting for cases where there is multiple file delete or restore - we need to ``waitTillFileRowsAreReady()`` to make sure all the file row actions have properly finished.

## Related Issue
#30942 
Note: this PR first adds tests for existing behavior that works. After this PR, it will be easy to add a couple more test scenarios for the more complex trashbin file selection combinations in the issue.

## Motivation and Context
There are not yet tests of the webUI behavior when selecting multiple files in the trashbin.

## How Has This Been Tested?
Run new tests locally.

## Types of changes
- [x] New acceptance tests

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/WebUIFilesContext.php
+++ b/tests/acceptance/features/bootstrap/WebUIFilesContext.php
@@ -749,7 +749,18 @@ class WebUIFilesContext extends RawMinkContext implements Context {
 	 * @return void
 	 */
 	public function theUserBatchDeletesTheMarkedFilesUsingTheWebUI() {
-		$this->filesPage->deleteAllSelectedFiles($this->getSession());
+		$pageObject = $this->getCurrentPageObject();
+		$pageObject->deleteAllSelectedFiles($this->getSession());
+	}
+
+	/**
+	 * @When the user batch restores the marked files using the webUI
+	 * @Given the user has batch restored the marked files using the webUI
+	 *
+	 * @return void
+	 */
+	public function theUserBatchRestoresTheMarkedFilesUsingTheWebUI() {
+		$this->trashbinPage->restoreAllSelectedFiles($this->getSession());
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


tests/acceptance/features/bootstrap/WebUIFilesContext.php
```diff
@@ -259,9 +259,26 @@ public function theUserCreatesTheFollowingFolderUsingTheWebUI(
...
+	 * @Then there should be exactly :count file/files listed on the webUI
+	 * @Then there should be exactly :count folder/folders listed on the webUI
+	 *
+	 * @param string $count that is numeric
+	 * @return void
+	 */
+	public function thereShouldBeCountFilesFoldersListedOnTheWebUI($count) {
+		$pageObject = $this->getCurrentPageObject();
+		PHPUnit_Framework_Assert::assertEquals(
+			$count,
```
individual-it: don't know what `assertEquals` does, but should we cast to `int`?

tests/acceptance/features/lib/FilesPageBasic.php
```diff
@@ -425,6 +456,23 @@ public function openFileActionsMenuByNo($number) {
...
+		$emptyContentElement = $this->find(
+			"xpath",
+			$this->getEmptyContentXpath()
+		);
+
+		if (!is_null($emptyContentElement)) {
+			return $emptyContentElement->isVisible();
+		}
+
+		return false;
```
individual-it: if the page is not loaded (yet, or not at all) this would return "false"
maybe put a `waitTillPageIsLoaded` before

individual-it (changes requested):
looks generally very good. just small changes

tests/acceptance/features/bootstrap/WebUIFilesContext.php
phil-davis: It does a type-juggling PHP-style ``==`` so it works with all sorts of "nearly equal" stuff. In the case here we compare a "numeric string" count with the actual integer count and it is fine.

individual-it (approved):

phil-davis:
Backport ``stable10`` #31006

PR #31002, opened 2018-04-04
Keep track of which groups are possibleToDelete

phil-davis:
## Description
1) Add an attribute of a group ``possibleToDelete`` which is ``false`` for LDAP groups. Use that attribute to decide what sort of group existence... checks we should really enforce.

2) LDAP users actually can be deleted on the local ownCloud, so remove the ``$shouldExist = false`` setting from those. We are happy for ``AfterScenario`` code to get rid of those users (which it is doing now) and to give warnings if it cannot.

This should help acceptance tests in ``user_ldap``

## Motivation and Context
When running user_ldap tests, the ``AfterScenario`` that cleans up groups complains because it cannot delete the LDAP groups. That is because there is a:
```
PHPUnit_Framework_Assert::assertFalse($this->groupExists($group));
```
that double-checks that the group really is gone, after attempting to delete it.

For LDAP, actually the group(s) cannot be deleted. so that double-check is not relevant in that case. For LDAP groups we are forced to just keep going between scenarios without explicitly deleting all the groups and creating them again.

## How Has This Been Tested?
CI knows

## Types of changes
- [x] Acceptance test refining

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -156,6 +156,27 @@ trait Provisioning {
 	}
 
 	/**
+	 *
+	 * @param string $groupname
+	 *
+	 * @return boolean
+	 * @throws Exception
+	 */
+	public function theGroupShouldBeAbleToBeDeleted($groupname) {
+		if (array_key_exists($groupname, $this->createdGroups)) {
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Backport ``stable10`` #31007

PR #31008, opened 2018-04-04
[stable10] Verify checksums command

IljaN:
## Related Issue
#30951 

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/VerifyChecksums.php
+++ b/apps/files/lib/Command/VerifyChecksums.php
@@ -1,0 +1,205 @@
+<?php
+/**
+ * @author Ilja Neumann <ineumann@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files\Command;
+
+
+use OC\Files\FileInfo;
+use OC\Files\Storage\Wrapper\Checksum;
+use OCP\Files\IRootFolder;
+use OCP\Files\Node;
+use OCP\Files\NotFoundException;
+use OCP\Files\Storage\IStorage;
+use OCP\IUser;
+use OCP\IUserManager;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+
+/**
+ * Recomputes checksums for all files and compares them to filecache
+ * entries. Provides repair option on mismatch.
+ *
+ * @package OCA\Files\Command
+ */
+class VerifyChecksums extends Command {
+
+
+	const EXIT_NO_ERRORS = 0;
+	const EXIT_CHECKSUM_ERRORS = 1;
+	const EXIT_INVALID_ARGS = 2;
+
+	/**
+	 * @var IRootFolder
+	 */
+	private $rootFolder;
+
+	/**
+	 * @var IUserManager
+	 */
+	private $userManager;
+
+	private $exitStatus = self::EXIT_NO_ERRORS;
+
+	/**
+	 * VerifyChecksums constructor.
+	 *
+	 * @param IRootFolder $rootFolder
+	 * @param IUserManager $userManager
+	 */
+	public function __construct(IRootFolder $rootFolder, IUserManager $userManager) {
+		parent::__construct(null);
+		$this->rootFolder = $rootFolder;
+		$this->userManager = $userManager;
+	}
+
+	protected function configure() {
+		$this
+			->setName('files:checksums:verify')
+			->setDescription('Get all checksums in filecache and compares them by recalculating the checksum of the file.')
+			->addOption('repair', 'r', InputOption::VALUE_NONE, 'Repair filecache-entry with mismatched checksums.')
+			->addOption('user', 'u', InputOption::VALUE_REQUIRED, 'Specific user to check')
+			->addOption('path', 'p', InputOption::VALUE_REQUIRED, 'Path to check relative to data e.g /john/files/', '');
+	}
+
+	public function execute(InputInterface $input, OutputInterface $output) {
+		$pathOption = $input->getOption('path');
+		$userName = $input->getOption('user');
+
+		if (!$pathOption && !$userName) {
+			$output->writeln('<info>This operation might take very long.</info>');
+		}
+
+		if ($pathOption && $userName) {
+			$output->writeln('<error>Please use either path or user exclusively</error>');
+			$this->exitStatus = self::EXIT_INVALID_ARGS;
+
+		}
+
+		$walkFunction = function (Node $node) use ($input, $output) {
+			$path = $node->getInternalPath();
+			$currentChecksums = $node->getChecksum();
+
+			// Files without calculated checksum can't cause checksum errors
+			if (empty($currentChecksums)) {
+				$output->writeln("Skipping $path => No Checksum", OutputInterface::VERBOSITY_VERBOSE);
+				return;
+			}
+
+			$output->writeln("Checking $path => $currentChecksums", OutputInterface::VERBOSITY_VERBOSE);
+			$actualChecksums = self::calculateActualChecksums($path, $node->getStorage());
+
+			if ($actualChecksums !== $currentChecksums) {
+				$output->writeln(
+					"<info>Mismatch for $path:\n Filecache:\t$currentChecksums\n Actual:\t$actualChecksums</info>"
+				);
+
+				$this->exitStatus = self::EXIT_CHECKSUM_ERRORS;
+
+				if ($input->getOption('repair')) {
+					$output->writeln("<info>Repairing $path</info>");
+					$this->updateChecksumsForNode($node, $actualChecksums);
+					$this->exitStatus = self::EXIT_NO_ERRORS;
+				}
+			}
+		};
+
+		$scanUserFunction = function(IUser $user) use ($input, $output, $walkFunction) {
+			$userFolder = $this->rootFolder->getUserFolder($user->getUID())->getParent();
+			$this->walkNodes($userFolder->getDirectoryListing(), $walkFunction);
+		};
+
+		if ($userName && $this->userManager->userExists($userName)) {
+			$scanUserFunction($this->userManager->get($userName));
+		} else if ($userName && !$this->userManager->userExists($userName)) {
+			$output->writeln("<error>User \"$userName\" does not exist</error>");
+			$this->exitStatus = self::EXIT_INVALID_ARGS;
+		} else if ($input->getOption('path')) {
+
+			try {
+				$node = $this->rootFolder->get($input->getOption('path'));
+			} catch (NotFoundException $ex) {
+				$output->writeln("<error>Path \"{$ex->getMessage()}\" not found.</error>");
+				$this->exitStatus = self::EXIT_INVALID_ARGS;
+				return $this->exitStatus;
+			}
+
+			$this->walkNodes([$node], $walkFunction);
+
+		} else {
+			$this->userManager->callForAllUsers($scanUserFunction);
+		}
+
+
+		return $this->exitStatus;
+	}
+
+
+	/**
+	 * Recursive walk nodes
+	 *
+	 * @param Node[] $nodes
+	 * @param \Closure $callBack
+	 */
+	private function walkNodes(array $nodes, \Closure $callBack) {
+		foreach ($nodes as $node) {
+			if ($node->getType() === FileInfo::TYPE_FOLDER) {
+				$this->walkNodes($node->getDirectoryListing(), $callBack);
+			} else {
+				$callBack($node);
+			}
+		}
+	}
+
+	/**
+	 * @param Node $node
+	 * @param $correctChecksum
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\StorageNotAvailableException
+	 */
+	private function updateChecksumsForNode(Node $node, $correctChecksum) {
+		$storage = $node->getStorage();
+		$cache = $storage->getCache();
+		$cache->update(
+			$node->getId(),
+			['checksum' => $correctChecksum]
+		);
+	}
+
+	/**
+	 * @param $path
+	 * @param IStorage $storage
+	 * @return string
+	 * @throws \OCP\Files\StorageNotAvailableException
+	 */
+	private static function calculateActualChecksums($path, IStorage $storage) {
+		return sprintf(
+			Checksum::CHECKSUMS_DB_FORMAT,
+			$storage->hash('sha1', $path),
+			$storage->hash('md5', $path),
+			$storage->hash('adler32', $path)
+		);
+	}
+}
+
--- a/apps/files/tests/Command/VerifyChecksumsTest.php
+++ b/apps/files/tests/Command/VerifyChecksumsTest.php
@@ -1,0 +1,315 @@
+<?php
+
+namespace OCA\Files\Tests\Command;
+
+use OC\Files\Node\File;
+use OC\Files\Storage\Wrapper\Checksum;
+use OCA\Files\Command\VerifyChecksums;
+use OCP\IUser;
+use Symfony\Component\Console\Tester\CommandTester;
+use Test\TestCase;
+
+/**
+ *
+ * @group DB
+ * @package OCA\Files\Tests\Command
+ */
+class VerifyChecksumsTest extends TestCase {
+	use \Test\Traits\UserTrait;
+
+
+	const BROKEN_CHECKSUM_STRING = '_BROKEN_';
+
+	/** @var CommandTester */
+	private $cmd;
+
+	/**
+	 * @var \OCP\Files\IRootFolder
+	 */
+	private $rootFolder;
+
+	/**
+	 * @var string
+	 */
+	private $user1;
+
+	/**
+	 * @var string
+	 */
+	private $user2;
+
+	/**
+	 * @var array
+	 */
+	private $testFiles;
+
+
+
+	public function setUp() {
+		parent::setUp();
+
+		$this->rootFolder = \OC::$server->getRootFolder();
+
+		$this->user1 = $this->createRandomUser(1);
+		$this->user2 = $this->createRandomUser(2);
+
+
+		$this->testFiles =  [
+			$this->createFileForUser($this->user1, 'dir/nested/somefile.txt', 'Hello World!'),
+			$this->createFileForUser($this->user1, 'dir/nested/subdir/hallo2.txt', 'ewfwfwefwef'),
+			$this->createFileForUser($this->user1, 'dir/nested/somefile2.txt', '1337'),
+			$this->createFileForUser($this->user1, 'otherdir/bar.txt', 'Bye!'),
+			$this->createFileForUser($this->user1, 'rootfile.doc', 'efwefwfwefwffw'),
+			$this->createFileForUser($this->user1, 'song.mp3', 'sdfsfdsdfsdfsfsfsffs'),
+			$this->createFileForUser($this->user2, 'sdvsdvs/secrets/bar/baz/moo.txt', 'JhonnyCache'),
+			$this->createFileForUser($this->user2, 'bling.tif', 'ewfwefwf242'),
+			$this->createFileForUser($this->user2, 'pong.tif', '12'),
+			$this->createFileForUser($this->user2, 'welcome.doc', 'efwefwfwefwffw'),
+		];
+
+		$this->cmd = new CommandTester(
+			new VerifyChecksums(
+				$this->rootFolder,
+				\OC::$server->getUserManager()
+			)
+		);
+	}
+
+	/**
+	 * @param string $uid
+	 * @param $path
+	 * @param $content
+	 * @return array
+	 * @throws \OCP\Files\NotPermittedException
+	 */
+	private function createFileForUser($uid, $path, $content) {
+
+		$userFolder = \OC::$server->getUserFolder($uid);
+
+		$parts = explode('/', ltrim($path, '/'));
+		$fileName = array_pop($parts);
+		$dirPath = $parts;
+
+		$currentDir = '';
+		foreach ($dirPath as $subDir) {
+			if (!empty($subDir)) {
+				$currentDir = "$currentDir/$subDir";
+				if (!$userFolder->nodeExists($currentDir)) {
+					$userFolder->newFolder("$currentDir");
+				}
+			}
+		}
+
+		$f = $userFolder->newFile("$currentDir/$fileName");
+		$f->putContent($content);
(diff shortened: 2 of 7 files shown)


apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,142 @@
...
+				}
+
+				// Files without calculated checksum can`t cause checksum errors
+				if (empty($currentChecksums)) {
+					return;
+				}
+
+				// StorageWrappers use getSourcePath() internally which already prepends the username to the path
+				// so we need to remove it here or else we will get paths like admin//admin/thumbnails/4/32-32.png
+				$pathWithoutUid = ltrim($path, "/{$user->getUID()}");
```
IljaN: Improve this because path starting with username would also be trimmed

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,142 @@
...
+
+
+	private $fieIdsWithBrokenChecksums = [];
+
+	protected function configure() {
+		$this
+			->setName('files:checksums:verify')
+			->setDescription('Get all checksums in filecache and compares them by recalculating the checksum of the file. (Might take very long)')
+			->addArgument('repair', InputArgument::OPTIONAL, "Reset cacheentry with missmatched checksums. They will be recalculated on first download.")
+			->addOption('path', 'p', InputOption::VALUE_REQUIRED, "Specific path to check");
```
IljaN: Implement this

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+				}
+
+				// Files without calculated checksum can`t cause checksum errors
+				if (empty($currentChecksums)) {
+					return;
+				}
+
+				// StorageWrappers use getSourcePath() internally which already prepends the username to the path
+				// so we need to remove it here or else we will get paths like admin//admin/thumbnails/4/32-32.png
+				$pathWithoutUid = preg_replace( "/\\/{$user->getUID()}/", '', $path, 1);
```
jvillafanez: I'm not sure if this will work properly... consider the following scenarios:
* user_ldap app with custom home folder (which doesn't match it's uid - check the home folder naming rule in the user_ldap configuration). This can have several problems:
    1. The preg_replace might not match anything -> you can have "abcd-1234-5678" as uid but a "user1home/thumbnails/43" path
    2. The preg_replace might match a different folder -> you can have "abcd-1234-5678" as uid but a "user1home/files/allimages/abcd-1234-5678/cities" path. While this scenario is quite unlikely to happen, the uid can also be a more common word such as "user1"
* External storages might work differently

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+						"<info>Mismatch for $path:\n Filecache:\t$currentChecksums\n Actual:\t$actualChecksums</info>"
+					);
+				}
+			});
+
+			$scanner->scan($input->getOption('path'));
+
+		};
+
+		$userManager = \OC::$server->getUserManager();
```
jvillafanez: This should be injected in the command

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+		$user = $input->getOption('user');
+
+		if ($user && $userManager->userExists($user)) {
+			$scanFunction($userManager->get($user));
```
jvillafanez: I'd move the scanFunction definition to a private or public method in this class (I'm not sure if it's possible to use a private method as a callback).

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+		} else if ($user && !$userManager->userExists($user)) {
```
jvillafanez:
```
if ($user) {
  if ($userManager->userExists($user)) {
  } else {
  }
} else {
}
```
This feels better since the conditions are evaluated only once. In your case, if $user is empty you check for $user twice before deciding to scan all users. If the user doesn't exists, you check for 4 conditions: $user, $userManager->userExists, $user again and finally !$userManager->userExists

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+			$output->writeln("<error>User $user does not exist</error>");
+			return;
+		} else {
+			$userManager->callForAllUsers($scanFunction);
+		}
+
+		/** @var QuestionHelper $questionHelper */
+		$questionHelper = $this->getHelper('question');
+		$repairQuestion = new ConfirmationQuestion("Do you want to reset (repair) broken checksums (y/N)? ", false);
```
jvillafanez: I'd rather try to fetch or create the objects we're going to use only if we're going to use them. I don't expect this to be heavy, but it should be easy to move these 2 lines inside the condition in order to create the objects and ask if it's truly needed.

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+				return;
+			}
+		}
+	}
+
+
+	/**
+	 * @param  $files
+	 */
+	private static function repairChecksumsForFiles(array $files) {
```
jvillafanez: why static?

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+				$file['file']->getId(),
+				['checksum' => $file['correctChecksums']]
+			);
+		}
+	}
+
+	/**
+	 * @return \OCP\IDBConnection
+	 */
+	protected function reconnectToDatabase(OutputInterface $output) {
```
jvillafanez: why protected? (I think I'll start asking this more frequently)

Good time to think about how we can do this reconnection properly? It's not the first place where I've seen this...

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+			} catch (\Exception $ex) {
+				$output->writeln("<info>Error while re-connecting to database: {$ex->getMessage()}</info>");
+				sleep(60);
+			}
+		}
+		return $connection;
+	}
+
+
+	private static function calculateActualChecksums($path, IStorage $storage) {
```
jvillafanez: why static?

jvillafanez:
If this has been tested, please specify what scenarios and setup have been tested.

IljaN:
@jvillafanez Thanks for your review,  after discussion with @DeepDiver1975  we will iterate over node api recursivly instead of using scanner

apps/files/lib/Command/VerifyChecksums.php
IljaN: I use static to mark private functions without side-effects

apps/files/lib/Command/VerifyChecksums.php
IljaN: Copy paste error, protected not rquired here

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+		}
+
+		$scanFunction = function(IUser $user) use ($input, $output) {
+			$scanner = new \OC\Files\Utils\Scanner(
+				$user->getUID(),
+				$this->reconnectToDatabase($output) ,
+				\OC::$server->getLogger()
+			);
+
+			$rootFolder =  \OC::$server->getRootFolder();
```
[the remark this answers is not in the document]
IljaN: :+1:

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+
+		if (!$input->getOption('path') && !$input->getOption('user')) {
+			$output->writeln('<info>This operation might take very long.</info>');
+		}
+
+		$scanFunction = function(IUser $user) use ($input, $output) {
+			$scanner = new \OC\Files\Utils\Scanner(
+				$user->getUID(),
+				$this->reconnectToDatabase($output) ,
+				\OC::$server->getLogger()
```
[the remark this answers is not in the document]
IljaN: :+1:

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,175 @@
...
+ */
+class VerifyChecksums extends Command {
+
+
+	private $fieIdsWithBrokenChecksums = [];
+
+	protected function configure() {
+		$this
+			->setName('files:checksums:verify')
+			->setDescription("Get all checksums in filecache and compares them by recalculating the checksum of the file.\n")
```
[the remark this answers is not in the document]
IljaN: :+1:

apps/files/lib/Command/VerifyChecksums.php
IljaN: :+1:

apps/files/lib/Command/VerifyChecksums.php
jvillafanez: hmmm, I guess from a technical perspective it's fine, but on the other hand, a private method can't be unittested (we can but it feels like hacking and also a bad practice). It's true that in this case we might not need to unittest these private methods... (we can consider this as an internal implementation, and the code is quite easy, so as long as the code is covered by some tests it shouldn't be important)

We can argue that we don't use "private static" often, so we can remove the static keyword to follow the same code style.

I don't have an opinion strong enough to go against in this case, so maybe @PVince81 for a second opinion?

PVince81: we have `invokePrivate` for testing private methods, let's use this if the alternative requires too much code gymnastics

IljaN:
Scanner removed, @jvillafanez tested with "old" objectstore

IljaN:
Note that path option currently does not work with objectstore: https://github.com/owncloud/objectstore/issues/90

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,200 @@
...
+ * @package OCA\Files\Command
+ */
+class VerifyChecksums extends Command {
+
+
+	/**
+	 * Entry format ['file' => $nodeObject, 'correctChecksum' => $checksum]
+	 * @var array
+	 */
+	private $nodesWithBrokenChecksums = [];
```
PVince81: will this scale in case of 10k entries ? should the repair be done in a paginated manner ?

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	 * @param $path
+	 * @param IStorage $storage
+	 * @return string
+	 * @throws \OCP\Files\StorageNotAvailableException
+	 */
+	private static function calculateActualChecksums($path, IStorage $storage) {
+		return sprintf(
+			'SHA1:%s MD5:%s ADLER32:%s',
```
PVince81: did we have a function that specifies these exact three ? hard-coding here is a bit risky in case we ever decide to change the checksums we compute by default.

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,200 @@
...
+
+		if (!empty($this->nodesWithBrokenChecksums)) {
+			/** @var QuestionHelper $questionHelper */
+			$questionHelper = $this->getHelper('question');
+			$repairQuestion = new ConfirmationQuestion(
+				"Do you want to repair broken checksums (y/N)? ",
+				false
+			);
+
+			if ($input->getOption('repair') || $questionHelper->ask($input, $output, $repairQuestion)) {
```
PVince81: without repair option does it even make sense to cache the nodes in "nodesWithBrokenChecksums" or just output the list directly to the console ?

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,200 @@
...
+			$this->walkNodes($userFolder->getDirectoryListing(), $walkFunction);
+		};
+
+		if ($userName && $this->userManager->userExists($userName)) {
+			$scanUserFunction($this->userManager->get($userName));
+		} else if ($userName && !$this->userManager->userExists($userName)) {
+			$output->writeln("<error>User $userName does not exist</error>");
+			return;
+		} else if ($input->getOption('path')) {
+			$node = $this->rootFolder->get($input->getOption('path'));
```
PVince81: how does this behave in case the path doesn't exist ?

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,200 @@
...
+
+		$scanUserFunction = function(IUser $user) use ($input, $output, $walkFunction) {
+			$userFolder = $this->rootFolder->getUserFolder($user->getUID())->getParent();
+			$this->walkNodes($userFolder->getDirectoryListing(), $walkFunction);
+		};
+
+		if ($userName && $this->userManager->userExists($userName)) {
+			$scanUserFunction($this->userManager->get($userName));
+		} else if ($userName && !$this->userManager->userExists($userName)) {
+			$output->writeln("<error>User $userName does not exist</error>");
```
PVince81: double quotes around "$userName" for readability ?

apps/files/lib/Command/VerifyChecksums.php
IljaN: Probably not, but as far as I know broken checksums are a rare occurence, (10 Files in 5k files)

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,200 @@
...
+	 * Recursive walk nodes
+	 *
+	 * @param Node[] $nodes
+	 * @param $path
+	 * @param \Closure $callBack
+	 */
+	private function walkNodes(array $nodes, \Closure $callBack) {
+		foreach ($nodes as $node) {
+			if ($node->getType() === FileInfo::TYPE_FOLDER) {
+				$this->walkNodes($node->getDirectoryListing(), $callBack);
```
PVince81: can we avoid recursion in case of very deep paths ? (happened before)

maybe some iteration with `yield` ?

apps/files/lib/Command/VerifyChecksums.php
IljaN: Error is shown, we can catch it and output <error></error> to omit the stacktrace which occurs on console.

apps/files/lib/Command/VerifyChecksums.php
IljaN: If repair option is not given there is a interactive prompt at the end which asks if you want to repair.

PVince81:
- [ ] add unit tests, there are already existing examples of command tests

PVince81:
you could use the approach from Command/ScanTest to generate some cache data, then have the test code scramble the hash value, then run the command

apps/files/lib/Command/VerifyChecksums.php
IljaN: Unfortunately not could introduce a format constant inside ChecksumWrapper

apps/files/lib/Command/VerifyChecksums.php
IljaN: Paginated repair would be possible if we remove interactive prompt at the and, this would also solve https://github.com/owncloud/core/pull/31008#discussion_r179787185, I will go for it...

IljaN: We can repair during scan then and don`t need to collect broken files

DeepDiver1975 (changes requested):
Tests Missing

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,205 @@
...
+		parent::__construct(null);
+		$this->rootFolder = $rootFolder;
+		$this->userManager = $userManager;
+	}
+
+	protected function configure() {
+		$this
+			->setName('files:checksums:verify')
+			->setDescription("Get all checksums in filecache and compares them by recalculating the checksum of the file.")
+			->addOption('repair', 'r', InputOption::VALUE_NONE, "Repair filecache-entry with missmatched checksums.")
```
PVince81: s/missmatched/mismatched/

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,205 @@
...
+			$output->writeln('<error>Please use either path or user exclusively</error>');
+			$this->exitStatus = self::EXIT_INVALID_ARGS;
+
+		}
+
+		$walkFunction = function (Node $node) use ($input, $output) {
+			$path = $node->getInternalPath();
+			$currentChecksums = $node->getChecksum();
+
+			// Files without calculated checksum can`t cause checksum errors
```
PVince81: strange backtick, use `'` for "can't", not the backtick

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,205 @@
...
+
+			if ($actualChecksums !== $currentChecksums) {
+				$output->writeln(
+					"<info>Mismatch for $path:\n Filecache:\t$currentChecksums\n Actual:\t$actualChecksums</info>"
+				);
+
+				$this->exitStatus = self::EXIT_CHECKSUM_ERRORS;
+
+				if ($input->getOption('repair')) {
+					$output->writeln("<info>Repairing!</info>");
```
PVince81: repairing what ? (think of people grepping for this specific line who want to see what files are going to be repaired)

apps/files/lib/Command/VerifyChecksums.php
```diff
@@ -0,0 +1,205 @@
...
+	 * @param Node $node
+	 * @param $correctChecksum
+	 * @throws NotFoundException
+	 * @throws \OCP\Files\InvalidPathException
+	 * @throws \OCP\Files\StorageNotAvailableException
+	 */
+	private function updateChecksumsForNode(Node $node, $correctChecksum) {
+		$storage = $node->getStorage();
+		$cache = $storage->getCache();
+		$cache->update(
```
PVince81: I'm wondering now whether we should also trigger an etag update. If we do so, the desktop clients will redownload said files. It might not be necessary because clients that already have the file with the correct checksum can stay as is. Clients which had trouble downloading files due to mismatch checksum will anyway try again.

So no need to adjust etag.

apps/files/tests/Command/VerifyChecksumsTest.php
```diff
@@ -0,0 +1,315 @@
...
+		$f = $this->rootFolder->get($f->getPath());
+	}
+
+	private function assertChecksumsAreCorrect(array $files) {
+		foreach ($files as $key => $file) {
+			/** @var File $f */
+			$f = $files[$key]['file'];
+			$expectedChecksums = $files[$key]['expectedChecksums'];
+			$this->refreshFileInfo($files[$key]['file']);
+			$this->assertThat(
```
PVince81: we usually use `$this->assertSame($expectedChecksums(), $f->getChecksum())` for strict equality

PVince81 (approved):
Code looks good :+1: 

My comments are only little optional tweaks

apps/files/lib/Command/VerifyChecksums.php
DeepDiver1975: valid point - as soon as we find a better solution transfer-ownership needs to be adapted (!) as well

DeepDiver1975 (changes requested):

apps/files/lib/Command/VerifyChecksums.php
IljaN: @DeepDiver1975 If I understand yield correctly this change should be made inside getDirectoryListing because yield returns an iterator. This would be a bigger api change which needs good testing...

tests/lib/Files/ViewTest.php
```diff
@@ -2603,34 +2603,6 @@ public function testGetDirectoryContentMimeFilter($filter, $expected) {
 		$this->assertEquals($expected, $files);
 	}
 
-	public function testFilePutContentsClearsChecksum() {
```
DeepDiver1975: ??? why is this test obsolete ??? 

Please add more explanation - THX

lib/private/Files/Cache/Scanner.php
```diff
@@ -205,8 +205,14 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 						$fileId = -1;
 					}
 					if (!empty($newData)) {
-						// Reset the checksum if the data has changed
-						$newData['checksum'] = '';
+						// Only reset checksum on file change
```
DeepDiver1975: Why is there a change to the scanner neccessary? This command no longer uses the scanner - right?

DeepDiver1975 (changes requested):
Things seem to get mixed up in here .... please clarify - THX

tests/lib/Files/ViewTest.php
PVince81: This is kind of an obscure thing.

This logic was added as part of https://github.com/owncloud/core/commit/0dbcbc4cae2eb669fc73f60a761855a907469111 to fix `file_put_contents()` as used by the text editor. The reason is that said logic automatically calls `scanFile()` after updating the contents on-disk. This was related to the old checksum logic before we even had the checksum wrapper. Since we did not have a way to compute the checksum, we need to clear it.
Remembber, back in the days a checksum could only be stored when sent by the client, we did not compute on server.

As of today, the checksum wrapper is able to compute the checksum on the fly during the `file_put_contents()` operation so when it reaches the code from https://github.com/owncloud/core/pull/31008#discussion_r180376503 the checksum is already present in `$data`, so instead of clearing we use just that.

However, when dealing with external storages we still need to clear the checksum if mtime, size, etag has changed on the remote storage. (remote storage currently does not deliver a checksum)

Hope this clarifies.

lib/private/Files/Cache/Scanner.php
IljaN: @DeepDiver1975 Scanner is called when adding a file via NodeAPI

PVince81: see my explanations above

lib/private/Files/ObjectStore/NoopScanner.php
```diff
@@ -30,6 +30,7 @@
 class NoopScanner extends Scanner {
 
 	public function __construct(Storage $storage) {
+		$this->storage = $storage;
 		//we don't need the storage, so do nothing here
```
butonic: comment no longer true

lib/private/Files/ObjectStore/NoopScanner.php
```diff
@@ -55,6 +56,15 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 	 * @return array with the meta data of the scanned file or folder
 	 */
 	public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $lock = true) {
+		// we only update the checksums - still returning no data
+		$meta = $this->storage->getMetaData($path);
+		if (isset($meta['checksum'])) {
+			$this->storage->getCache()->put(
```
butonic: put updates only when the data changes, which is good

lib/private/Files/ObjectStore/NoopScanner.php
```diff
@@ -55,6 +56,15 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 	 * @return array with the meta data of the scanned file or folder
 	 */
 	public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $lock = true) {
+		// we only update the checksums - still returning no data
+		$meta = $this->storage->getMetaData($path);
```
butonic: how expensive is this operation for objectstorages? another trip over the wire? do we get the checksum in the PUT response? can we cache that? or better yet, directly update the cache? then this commit would not be necessary

DeepDiver1975: the checksum is computed while writing - the getMetaData operation grabs the computed checksums from memory and stores them to the cache.

so no additional operation on the objectstore. no trip over the wire.

mmattel:
@settermjd FYI
@IljaN please create a doc issue inlcuding a link to this PR and if possible a PR desribing the change.

DeepDiver1975 (approved):

PR #31009, opened 2018-04-04
Rework unreliable delete multiple files scenario

phil-davis:
## Description
1) Remove unreliable step
2) Add an "easier" check that some example file that should have been deleted has gone from the list.

## Motivation and Context
```
  Scenario: Delete all except for a few files at once                 # /drone/src/tests/acceptance/features/webUIFiles/deleteFilesFolders.feature:74
    When the user marks all files for batch action using the webUI    # WebUIFilesContext::theUserMarksAllFilesForBatchActionUsingTheWebUI()
    And the user unmarks these files for batch action using the webUI # WebUIFilesContext::theUserMarksTheseFilesForBatchActionUsingTheWebUI()
      | name          |
      | lorem.txt     |
      | simple-folder |
    And the user batch deletes the marked files using the webUI       # WebUIFilesContext::theUserBatchDeletesTheMarkedFilesUsingTheWebUI()
    Then there should be exactly 2 files/folders listed on the webUI  # WebUIFilesContext::thereShouldBeCountFilesFoldersListedOnTheWebUI()
Failed asserting that 3 matches expected '2'.
```
This scenario was added today in PR #30999 It passes when run locally, but on drone it seems to be sometimes passing and sometimes failing. It deletes a lot of files from a folder and then checks to see that exactly the expect number of files remain. There seems to be some issue on fast systems that it does not wait long enough for all the file deletes to be finished processing.

Better not to have unreliable tests in ``master`` or ``stable10`` - so change the scenario to get rid of the step that fails. The detail of why can be looked into later. 

## Types of changes

## Checklist:

### DIFF
--- a/tests/acceptance/features/webUIFiles/deleteFilesFolders.feature
+++ b/tests/acceptance/features/webUIFiles/deleteFilesFolders.feature
@@ -78,9 +78,10 @@ So that I can keep my filing system clean and tidy
 			| lorem.txt     |
 			| simple-folder |
 		And the user batch deletes the marked files using the webUI
-		Then there should be exactly 2 files/folders listed on the webUI
-		And the folder "simple-folder" should be listed on the webUI
+		Then the folder "simple-folder" should be listed on the webUI
 		And the file "lorem.txt" should be listed on the webUI
+		# Check just an example of a file that should not exist any more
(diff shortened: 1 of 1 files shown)


phil-davis:
Needs to be backported at the same time as #30999

individual-it (approved):

PR #31012, opened 2018-04-05
[test] [stable10] catch group delete 500

phil-davis:
Run all the webUI chrome tests on Travis to confirm if this really fixes the problem.


phil-davis:
Code here does the job on Travis

phil-davis:
And has been merged in PR #31011 and #31013

Issue #31015, opened 2018-04-05
Cannot view or delete groups with a slash in their name using provisioning API


phil-davis:
You can create groups with all sorts of names in the webUI or with the provisioning API like:
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups -X POST -d "groupid=q?mark"
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups -X POST -d "groupid=a%sign"
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups -X POST -d "groupid=a/slash"
```
and other special characters.

You can list all groups fine with:
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups
```

And view individual groups with:
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups/a%sign
```
and to get a "?" through, you have to put the escaped value "%3F"
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups/q%3Fmark
```

You can delete such groups with:
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups/a%sign -X DELETE
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups/q%3Fmark -X DELETE
```

The problem comes if you try with groups that have a ``/`` in their name. A literal slash is never going to work:
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups/a/slash
```
And an escaped slash, like this, is rejected by the server (Apache or PHP dev server):
```
curl http://admin:admin@localhost:8080/ocs/v2.php/cloud/groups/a%2Fslash
```

So you cannot manage a group like ``a/slash`` via the provisioning API. But you can manage it through webUI.

phil-davis:
This was noticed while merging code for webUI and API acceptance tests. The webUI tests have a scenario that creates a group called ``a/slash`` - the API acceptance test steps could not delete the group.

ownclouders:
[...]

PR #31017, opened 2018-04-05
Handle groups with slash and percent in provisioning API

phil-davis:
## Description
1) Add a range of test cases to the ``apiProvisioning`` acceptance test suite covering various group name combinations, including with ``/`` ``%`` ``?`` etc.
2) Modify ``apps/provisioning_api/lib/Groups.php`` so it will turn a literal ``%2F`` into a ``/`` - that provides a way to pass a ``/`` in a group name. (from outside you pass in ``%252F`` and the ``%25`` turns into a ``%`` by the time it gets through the server and is delivered to ``Groups.php``)
3) Similarly turn a literal ``%25`` into a ``%`` - you need that so you can always escape a ``%`` if you want to construct a literal group like ``my%25group`` or ``my%2Fgroup``

## Related Issue
#31015 

## Motivation and Context
It is nice to be able to use all group names in the Provisioning API.

## How Has This Been Tested?
Local API acceptance test runs.

## Types of changes

## Checklist:

### DIFF
--- a/apps/provisioning_api/lib/Groups.php
+++ b/apps/provisioning_api/lib/Groups.php
@@ -54,6 +54,17 @@ class Groups {
 	}
 
 	/**
+	 * unencode any encoded "/" or "%" that are intended to be literally part
+	 * of the group name.
+	 *
+	 * @param string $groupId
+	 * @return string
+	 */
+	private function unencodeGroupId($groupId) {
+		return \strtr($groupId, ['%25' => '%', '%2F' => '/']);
+	}
+
+	/**
 	 * returns a list of groups
 	 *
 	 * @param array $parameters
@@ -93,7 +104,7 @@ class Groups {
 			return new OC_OCS_Result(null, \OCP\API::RESPOND_UNAUTHORISED);
 		}
 
-		$groupId = $parameters['groupid'];
+		$groupId = $this->unencodeGroupId($parameters['groupid']);
 
 		// Check the group exists
 		if (!$this->groupManager->groupExists($groupId)) {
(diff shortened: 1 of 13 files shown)


phil-davis:
Here are some references with useful discussion about passing through embedded ``/`` in URLs:
http://www.leakon.com/archives/865
https://stackoverflow.com/questions/3235219/urlencoded-forward-slash-is-breaking-url
https://www.w3schools.com/tags/ref_urlencode.asp

phil-davis:
@PVince81 @DeepDiver1975 can you have a look at the approach I took in ``Groups.php`` and comment if it seems fine (or not)?

@individual-it can review the test changes/additions.

DeepDiver1975:
this needs to be added to the documentation @settermjd

DeepDiver1975:
@PVince81 objections?

apps/provisioning_api/lib/Groups.php
```diff
@@ -53,6 +53,17 @@ public function __construct(\OCP\IGroupManager $groupManager,
...
 
+	/**
+	 * unencode any encoded "/" or "%" that are intended to be literally part
+	 * of the group name.
+	 *
+	 * @param string $groupId
+	 * @return string
+	 */
+	private function unencodeGroupId($groupId) {
+		return \strtr($groupId, ['%25' => '%', '%2F' => '/']);
```
PVince81: isn't there a PHP function that does that like `urldecode` or `rawurldecode` ?

PVince81: what about "@" signs ?

phil-davis: "@" signs work OK (API tests here check creating, working with and deleting such group names)

PVince81: I remember in some cases when talking Webdav we had to encode the "@" sign, so not sure if it applies here. what if a third party client decides to URL encode other special chars ?

phil-davis: At this point we do not want to ``urldecide`` everything. If somebody wanted something literally in a group name like ``group%29435`` then they really do want those characters in the group name, so we do not want to decode ``%29`` into whatever ASCII that is.

Actually they will have sent us ``group%2529435`` so that the ``%`` gets escaped on the way through and re-formed here.

PVince81:
So we fixed the app framework in https://github.com/owncloud/core/pull/31224 but the provisioning API is not using it and uses OCS instead.

As far as I remember there is a way to port the old OCS way to use App Framework controllers, but that would likely be a lot more work.

phil-davis:
I think we have come up with other ways to make this all work - I will look at what PRs are "out there".
I rebased this to keep to up-to-date in case some part of it looks useful.

phil-davis:
Rebased again to merge in the addition of skipped tests in master yesterday. Let's see what CI thinks.

PVince81:
conflict.

didn't we fix slash handling somewhere before ?

phil-davis:
Rebased and resolved conflicts. Let's see what CI thinks.

DeepDiver1975:

PR #30938, opened 2018-03-27
Better UX in files sidebar for image previews

DeepDiver1975:
## Description
- center image load spinner in files sidebar
- use fixed heights to prevent sidebar content to hop around
- use image previews which fill the whole preview area in the sidebar

Before:
screencast_27-03-18_17_04_02

After:
screencast_27-03-18_17_04_58

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31049, opened 2018-04-09
[stable10] - center image load spinner in files sidebar

DeepDiver1975:
- use fixed heights to prevent sidebar content to hop around
- use image previews which fill the whole preview area in the sidebar

backport #30938


PVince81 (approved):
:+1: anyway

PR #31051, opened 2018-04-09
Check basic auth credentials periodically after a timeout instead of …

butonic:
…every request to stop hammering ldap servers if client uses basic auth

### DIFF
--- a/lib/private/User/BasicAuthModule.php
+++ b/lib/private/User/BasicAuthModule.php
@@ -1,5 +1,6 @@
 <?php
 /**
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
  *
  * @copyright Copyright (c) 2018, ownCloud GmbH
@@ -23,7 +24,10 @@
 namespace OC\User;
 
 
+use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Authentication\IAuthModule;
+use OCP\IConfig;
+use OCP\ILogger;
 use OCP\IRequest;
 use OCP\ISession;
 use OCP\IUser;
@@ -31,14 +35,24 @@ use OCP\IUserManager;
 
 class BasicAuthModule implements IAuthModule {
 
+
+	/** @var IConfig */
+	private $config;
+	/** @var ILogger */
+	private $logger;
 	/** @var IUserManager */
 	private $manager;
 	/** @var ISession */
 	private $session;
+	/** @var ITimeFactory */
+	private $timeFactory;
 
-	public function __construct(IUserManager $manager, ISession $session) {
+	public function __construct(IConfig $config, ILogger $logger, IUserManager $manager, ISession $session, ITimeFactory $timeFactory) {
+		$this->config = $config;
+		$this->logger = $logger;
 		$this->manager = $manager;
 		$this->session = $session;
+		$this->timeFactory = $timeFactory;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


PVince81:
This seems somewhat redundant of `Session::checkTokenCredentials()` or maybe the latter only applied for tokens and not regular passwords ? Considering that this PR uses the exact same setting it seems the logic was copied to also apply to regular passwords.

Just wanted to make sure that this all fits together. @butonic @DeepDiver1975 can you clarify ?

lib/private/User/BasicAuthModule.php
```diff
@@ -57,16 +66,39 @@ public function auth(IRequest $request) {
...
+			$lastCheck = $this->session->get('last_check_timeout');
+		} else {
+			$lastCheck = 0;
 		}
-		// check email and password
-		$users = $this->manager->getByEmail($authUser);
-		if (count($users) === 1) {
-			$user = $this->manager->checkPassword($users[0]->getUID(), $authPass);
+		$now = $this->timeFactory->getTime();
+		$last_check_timeout = (int)$this->config->getAppValue('core', 'last_check_timeout', 5);
```
PVince81: camel case for PHP variables, let's not go back to underscores...

PVince81:
As I understand this only works for the combination of basic auth + session because it stores the state into the session.

So if some client is hammering the server with multiple requests and accepts cookies, it will only recheck the password periodically, not always. This also means that if a client is funny and sends a different password in the middle, it would go undetected. Not sure if that's a bad thing...

For pure basic auth (ex: cadaver) this likely will not help.

tomneedham:
> So if some client is hammering the server with multiple requests and accepts cookies, it will only recheck the password periodically, not always. This also means that if a client is funny and sends a different password in the middle, it would go undetected. Not sure if that's a bad thing...

Yes this means in between the timeout, we rely on the session for the auth.

butonic:
@PVince81 I only reused the config option name. yes it does what the token check does for basic auth. best we can do for new. proper fix is to not use a session for dav & ocs ... but that is a way bigger change.

butonic:
cadaver currently also hits ldap. no change there, yes

PVince81:
Thanks, makes sense then. Would be good to get a review from @DeepDiver1975 or @tomneedham with who you discussed this in more detail

lib/private/User/BasicAuthModule.php
```diff
@@ -57,16 +66,39 @@ public function auth(IRequest $request) {
...
+			// Checked performed recently, nothing to do now
+			$user = $this->manager->get($this->session->get('user_id'));
+		} else {
+			// check uid and password
+			$user = $this->manager->checkPassword($authUser, $authPass);
+			if ($user instanceof IUser) {
+				$this->session->set('last_check_timeout', $now);
+				return $user;
+			}
+			// TODO inject logger & debug log if no user was found.
```
DeepDiver1975: logging is done in deeper layers - no need here. let's remove this comment

lib/private/User/BasicAuthModule.php
```diff
@@ -57,16 +66,39 @@ public function auth(IRequest $request) {
...
+			// check email and password
+			$users = $this->manager->getByEmail($authUser);
+			if (count($users) === 1) {
+				$this->session->set('last_check_timeout', $now);
```
DeepDiver1975: time should only be checked if checkPassword below is successful

DeepDiver1975 (changes requested):

lib/private/User/BasicAuthModule.php
butonic: No. there is a difference between not being able to log in a user by username vs by email. the if statement below also needs debug log for cases when 0 ore more thin one user was found. 

If you are talking about https://github.com/owncloud/core/blob/7666497c0ada87c5df67a793e22152d478f2740a/lib/private/User/Manager.php#L253 that line uses warning ... which is bs because login via email is perfectly fine, in which case we currently ALWAYS log a warning ... which makes any admin crazy.

lib/private/User/BasicAuthModule.php
butonic: I am reusing the option from the token check -> no camel case

lib/private/User/BasicAuthModule.php
DeepDiver1975: but this whole code block as of today is NOT used for login *yet* but only for verifying the auth headers. I fear we potentially pollute the log too much

butonic: debug = I really need to know what is going on in order to debug a problem, eg why the hell is the user unable to login. we can discuss if this is debug or trace level debugging ... but the decisions are made here ... so we need the debug log here

DeepDiver1975: okay - debug log level is fine for me - just add it right away - THX

butonic: There is no sich thing as too much logging. there is just not enough filtering. which brings back https://github.com/owncloud/core/pull/10405 ... we already added too much logic to log related stuff. different issue ...

PVince81:
Ok tests pass now. Needs review approval and also some testing for the case this fixes.

Regression auth testing will be done in RC after merge.

cc @patrickjahns

DeepDiver1975 (approved):

PVince81:
thanks, please backport

DeepDiver1975:
> thanks, please backport

will do

PVince81:
stable10: https://github.com/owncloud/core/pull/31076

Issue #31052, opened 2018-04-09
Core offers no way for IApacheBackends to discover and sync users during request auth


tomneedham:
We used to rely on `handleLogin` which tries apache auth first allowing sync. https://github.com/owncloud/core/blob/master/lib/base.php#L959

However, we now use $session->verifyAuthHeaders which does not check apacheAuth: https://github.com/owncloud/core/blob/master/lib/base.php#L892

Where should we use `OC\User\Session->verifyAuthHeaders()` and where should we use `OC::handleLogin`. 

Currently we rely on `IApacheBackends` to do their own user discovery and account creation in their app.php which we want to avoid because it is expensive and should not happen during app loading. We want an explicit code path that can be tested and better controlled by core.

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributors most likely able to help are @PVince81, and @butonic.

Possibly related issues are https://github.com/owncloud/core/issues/6017 (Offer one way sync), https://github.com/owncloud/core/pull/30638 (Core controlled user metadata sync), https://github.com/owncloud/core/pull/31029 (Discover new users in single user sync), https://github.com/owncloud/core/issues/1292 (Coordinate n-way sync with multiple clients/users), and https://github.com/owncloud/core/pull/31032 ([stable10] Discover new users in single user sync).

PVince81:
@tomneedham close ?

tomneedham:
Yep, solved with #31054

PR #31054, opened 2018-04-09
Core should check apache auth

tomneedham:
This fixes Apache based logins in 10.0.8.

### DIFF
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -108,7 +108,7 @@ class LoginController extends Controller {
 	 * @return TemplateResponse|RedirectResponse
 	 */
 	public function showLoginForm($user, $redirect_url, $remember_login) {
-		if ($this->userSession->isLoggedIn()) {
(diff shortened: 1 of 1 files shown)


tomneedham:
Fixes https://github.com/owncloud/core/issues/31052

tomneedham:
Looks like its the `IApacheBackend` that needs the caching fix for this - not core.

tomneedham:
Tested in the wild ✅

PVince81:
sounds too risky for now, moving to "planned"

core/Controller/LoginController.php
```diff
@@ -26,6 +26,7 @@
 namespace OC\Core\Controller;
 
 use OC\Authentication\TwoFactorAuth\Manager;
+use OC\User\LoginException;
```
tomneedham: remove this

butonic (approved):

PVince81:
ok, so we do want it because it fixes a regression with shib

PR #31064, opened 2018-04-10
Filter sender display name in mail notification handler

PVince81:
Port of https://github.com/owncloud/core/pull/31056 to master

### DIFF
--- a/lib/private/Share/MailNotifications.php
+++ b/lib/private/Share/MailNotifications.php
@@ -85,7 +85,12 @@ class MailNotifications {
 		$this->urlGenerator = $urlGenerator;
 
 		$this->replyTo = $this->user->getEMailAddress();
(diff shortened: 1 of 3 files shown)


lib/private/Share/MailNotifications.php
```diff
@@ -85,7 +85,12 @@ public function __construct(IUser $user,
 		$this->urlGenerator = $urlGenerator;
 
 		$this->replyTo = $this->user->getEMailAddress();
```
DeepDiver1975: no filter on email? just asking ...

PVince81: it's likely already filtered on input. we do validate email addresses now... hmmm but older ones... maybe worth adding here separately

DeepDiver1975 (approved):

PR #31071, opened 2018-04-10
Revert "[stable10] - center image load spinner in files sidebar"

DeepDiver1975:
Reverts owncloud/core#31049


PVince81 (approved):
:+1: as discussed, let's be safe for the RC

PR #31072, opened 2018-04-10
check loglevel before interpolating message

butonic:
interpolation is not worth doing if the message will never be logged


DeepDiver1975 (approved):

phil-davis:
Possible backport ``stable10`` #31113

PR #31076, opened 2018-04-10
[stable10] Check basic auth credentials periodically after a timeout instead of …

DeepDiver1975:
…every request

backport of #31051


butonic (approved):
works as expected. tested with ldap and debugging the flow

butonic:
@DeepDiver1975 @PVince81 @tomneedham something to be aware: the checkPassword call will cache the LDAP entry and use it to sync the user ... every 5 min

- [ ] make the manager set `$this->session->set('last_check_timeout', $now);` currently, the request right after the login will trigger an additional sync

Normal login requests must always sync. The db update is triggered only when the account changes and the ldap request has already been made anyway.

PR #31078, opened 2018-04-10
Check for and install missing core dependencies

settermjd:
## Description

Checks for and installs any missing [core dependency](https://doc.owncloud.org/server/latest/admin_manual/installation/source_installation.html#install-the-required-packages).

**This is currently a WIP!**

## Motivation and Context

After a series of times where I've been frustrated that packages were missing and then had to consult the documentation to find what they were, and for a simple script to install them, I wrote this small script to do it all programmatically. Currently, it only supports Ubuntu 17.10, but, I plan to improve upon it for future releases.

## How Has This Been Tested?

I tested the script on a local installation.

n/a

## Types of changes

## Checklist:

- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have read the **CONTRIBUTING** document.
- [X] All new and existing tests passed.


mmattel:
Have the webserver and the database as choice (apache, nginx, mysql, maria-db)

bin/check-dependencies.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "Nothing left to do."
+    echo "Exiting."
+    exit 0
+  fi
+}
+
+function check_distribution()
+{
+  distribution=$( grep -i "version_id" /etc/os-release | awk -F'=' '{print $2}' | tr -d '"')
+  if [ $distribution != "17.10" ]
```
patrickjahns: Please propose a LTS version - that is what we officially support

settermjd: Sounds fair. Thanks for the feedback.

settermjd: @patrickjahns, the script now supports Ubuntu 17.04, 16.04, and 16.04.4.

settermjd: https://github.com/owncloud/core/pull/31078/commits/cde28ba481f4e6e7668216bf2e04e79974ed75a7 added support for the previous LTS release, 14.04.

bin/check-dependencies.sh
```diff
@@ -0,0 +1,167 @@
...
+something's going wrong, then you should background the script and tail
+${LOG_FILE}.
+"
+
+##
+## Determine the user's distribution
+##
+function get_distribution()
+{
+  version=$( grep -E "^VERSION_ID" /etc/os-release | awk -F'=' '{print $2}' | tr -d '"' )
```
heiglandreas: Will `/etc/os-release` be available on every target-system?

heiglandreas: I'd rather see whether I can get that information from `/proc` or similar…

bin/check-dependencies.sh
```diff
@@ -0,0 +1,167 @@
...
+       ;;
+   \?) printf "illegal option: -%s\n" "$OPTARG" >&2
+       echo "$usage" >&2
+       exit 1
+       ;;
+  esac
+done
+shift $((OPTIND - 1))
+
+if [ $( whoami ) != "root" ]
```
heiglandreas: I'd set that right at the beginning of the script. What sense does it make to call `script -h` only to then see that running `script do-something-sensible` doesn't work because it needs to be run as root. But that's just my personal view ;-)

heiglandreas (commented):
I assume that the syntax is working and as you requested it to be run in bash that should be no problem ;)

bin/check-dependencies.sh
settermjd: Good point. I'd not thought of that. Thanks for the feedback.

bin/check-dependencies.sh
settermjd: I'm doing a bit more digging, but I believe so.

jnweiger: to avoid awk ugliness, you can use 
```version=$(. /etc/os-release; echo $VERSION_ID)```

File: bin/check-dependencies.sh [hunk 4]
```diff
@@ -0,0 +1,167 @@
...
+  fi
+}
+
+function check_dependencies()
+{
+  missing_dependencies=()
+
+  # Determine the correct dependencies configuration file to read in
+  distro_lower=$( echo $(get_distribution) | awk '{ gsub(" ", "-"); print tolower($0) }' )
+  source "bin/dependencies/${distro_lower}.cfg"
```
jnweiger: avoiding another awk:
`distro_lower=$( echo $(get_distribution) | tr 'A-Z ' 'a-z-' )`

bin/check-dependencies.sh
```diff
@@ -0,0 +1,167 @@
...
+  echo
+
+  for package in ${!core_dependencies[@]}
+  do
+    pkg=${core_dependencies[$package]}
+    
+    if dpkg -l ${pkg} 2>/dev/null | grep -E -q "^ii"  
+    then
+      echo -e " \xE2\x9c\x94 ${pkg} is installed"
+    else
```
jnweiger: Please add a comment, what \xE2\x9c\x94  means. A unicode glyph on some terminals?

bin/check-dependencies.sh
```diff
@@ -0,0 +1,167 @@
+#!/bin/bash
+## 
+## This script checks if the required dependency packages have been installed
+## If one or more have not been installed, then it installs them.
+## It currently only works on Ubuntu 17.10 and has to be run as root.
```
jnweiger: if you write all the `apt` calls as `sudo apt` then it will prompt for the root password if needed (and only once).
The whoami check below should just print a friendly message then, saying that we will prompt for the root password to install stuff....

bin/check-dependencies.sh
```diff
@@ -0,0 +1,167 @@
...
+##
+## Author: Matthew Setter <matthew@matthewsetter.com>
+##
+
+SUPPORTED_DISTRIBUTIONS=(
```
jnweiger: comment to add here:
`# CAUTION: keep in sync with the available bin/dependencies/*.cfg files`

bin/dependencies/ubuntu-14.04.cfg
```diff
@@ -0,0 +1,20 @@
+core_dependencies=(
+  apache2 mariadb-server
```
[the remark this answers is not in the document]
jnweiger: Please update documentation at
https://doc.owncloud.com/server/10.0/admin_manual/installation/system_requirements.html#officially-recommended-supported-options and
https://doc.owncloud.com/server/10.0/admin_manual/installation/source_installation.html
 * There are many more platforms listed there, than we have here. (Tumbleweed should not be a supported platform, imho)
 * The package lists in source_installation.html  differ from the package lists here.
 * The existance of this tool should be mentioned in the docs.
 * This tool should print out the url to the online-docs when started.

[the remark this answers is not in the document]
jnweiger: @crrodriguez please check the package dependencies against these. Maybe we can interpolate these bin/dependencies/$PLATFORM.cfg lists into spec files and control files?

settermjd:
Need to add the option to install install dependencies, such as wget, make, and git.

bin/check-dependencies.sh
settermjd: That's pretty wicked @jnweiger. I'd never thought of doing that, _until now_. Thanks!!

bin/check-dependencies.sh
settermjd: I'm going to remove this and just check the available config files. It makes no sense to me to check twice when once will suffice.

bin/check-dependencies.sh
settermjd: I decided to just remove the `whoami` check altogether.

bin/dependencies/ubuntu-14.04.cfg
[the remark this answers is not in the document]
settermjd: I removed any references to mariadb. Your suggestion makes perfect sense. Thanks. @mmattel, I still need to figure out an NGINX option. Working on that.

mmattel:
We have just defined in the documentation for 10.0.8 that the recommendation for php is 7.1
For Ubuntu 14.04 and 16.04 you need to enable a ppa to do so.
```
sudo add-apt-repository -y ppa:ondrej/php
sudo apt-get update -y
```
Then you can install 7.1
I would not recommend to do something different than we do in the documentation :see_no_evil:

PVince81:
who's the target audience for this script ? is it for developers ?

what about people who deploy on servers, could they also benefit from this ?

in any case it feels like such script would belong in some part of the documentation (attachment?) or in the [administration repo](https://github.com/owncloud/administration) which already provide various tools for setting up environments

@settermjd

PVince81:
the docs could then be adjusted to point at a subfolder on the administration repo if needed

settermjd:
@PVince81, my original intent was to make it simple for anyone to use, not just admins. So, perhaps you're right, that it should go in the docs, and not here in core.

DeepDiver1975:

PR #31085, opened 2018-04-11
occ repair command for vobjects

Julian1998:
This command uses the Sabre\VObject validate function. (so it doesn't work for every kind of situation)


DeepDiver1975 (changes requested):
- [ ] at the end unit tests need to be added
- [ ] based on the returned warning of the validate method the user has to be asked for correct data input
- [ ] let user decide if changes shall be written to the database

PR #31086, opened 2018-04-11
During chunking process the after file creation not triggered

sharidas:
In the clients like android, where chunking happens for files
greater than 1MB. Under such conditions the after file creation
symfony event is not triggered. This change helps to fix the
issue.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

When clients try to upload files and chunking is enabled. Then the after file creation symfony event is not triggered. This change addresses the issue.

## Motivation and Context

When clients try to upload files and chunking is enabled. Then the after file creation symfony event is not triggered. This change addresses the issue.

## How Has This Been Tested?

- [x] Upload a file using android app.
- [x] The after event triggered is listened using a test app.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31087, opened 2018-04-11
[stable10] During chunking process the after file creation not triggered

sharidas:
In the clients like android, where chunking happens for files
greater than 1MB. Under such conditions the after file creation
symfony event is not triggered. This change helps to fix the
issue.

## Description

When clients try to upload files and chunking is enabled. Then the after file creation symfony event is not triggered. This change addresses the issue.

## Motivation and Context

When clients try to upload files and chunking is enabled. Then the after file creation symfony event is not triggered. This change addresses the issue.

## How Has This Been Tested?

- [x] Upload a file using android app.
- [x] The after event triggered is listened using a test app.

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -111,9 +111,6 @@ class File extends Node implements IFile {
 	 * @return string|null
 	 */
 	public function put($data) {
-		$path = $this->fileView->getAbsolutePath($this->path);
-		$beforeEvent = new GenericEvent(null, ['path' => $path]);
-		\OC::$server->getEventDispatcher()->dispatch('file.beforecreate', $beforeEvent);
 		try {
 			$exists = $this->fileView->file_exists($this->path);
 			if ($this->info && $exists && !$this->info->isUpdateable()) {
@@ -133,6 +130,16 @@ class File extends Node implements IFile {
 			} catch (\Exception $e) {
 				$this->convertToSabreException($e);
 			}
+		}
+
+		$newFile = false;
+		$path = $this->fileView->getAbsolutePath($this->path);
+		$beforeEvent = new GenericEvent(null, ['path' => $path]);
+		if (!$this->fileView->file_exists($this->path)) {
+			\OC::$server->getEventDispatcher()->dispatch('file.beforecreate', $beforeEvent);
+			$newFile = true;
+		} else {
+			\OC::$server->getEventDispatcher()->dispatch('file.beforeupdate', $beforeEvent);
 		}
 
 		list($partStorage) = $this->fileView->resolvePath($this->path);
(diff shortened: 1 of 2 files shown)


sharidas:
Original pr: https://github.com/owncloud/core/pull/31086

SergioBertolinSG:
https://github.com/owncloud/admin_audit/issues/66 is fixed with this patch.

@sharidas do you think this PR could affect other parts of the server? createFileChunked looks risky, what else should be tested before merging this?

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -553,7 +553,23 @@ private function createFileChunked($data) {
...
 
-				return $info->getEtag();
+				$etag = $info->getEtag();
+				/**
+				 * The absoulte path has chunking file name lets say /$uid/files/VID-20171022-WA0000.mp4-chunking-4043-10-9
+				 * Where as the real filename would be /$uid/files/VID-20171022-WA0000.mp4
+				 * So we try to remove VID-20171022-WA0000.mp4-chunking-4043-10-9 from the absolute path and append
+				 * VID-20171022-WA0000.mp4 to the path to get the absolute path of real file.
+				 */
+				$path = explode('/', $this->fileView->getAbsolutePath($this->path));
```
PVince81: too hacky. doesn't `$targetPath` already contain the absolute path and file name of the final file after assembly ?

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -553,7 +553,23 @@ private function createFileChunked($data) {
...
+				array_pop($path);
+				$path = implode('/', $path);
+				$path = $path . '/' . $info['name'];
+
+				if ($etag !== null) {
+					$afterEvent = new GenericEvent(null, ['path' => $path]);
+					\OC::$server->getEventDispatcher()->dispatch('file.aftercreate', $afterEvent);
```
PVince81: is `aftercreate` correct for all scenarios here ? do we have different events for "create a brand new file" and "overwrite an existing file" ?

PVince81 (changes requested):
also need to verify that aftercreate is submitted not only of old chunking like done in this PR (remote.php/webdav) but also for new chunking (remote.php/dav/files/$user/)

PVince81:
do we also have a matching "beforecreate" event for this scenario ?

cdamken:
@PVince81 RC2 ?
Otherwise please move to development milestone. Thanks

PVince81:
added to my RC2 list...

sharidas:
After evaluating a bit further. I have made changes to this PR:
- [x] Create separate create and update events for `put()`
- [x] Create separate create and update for chunking method `createFileChunked()`.

The reason why create and update were required is, I observed that files can be updated and created using same calls. The unit tests were written for the same to make my observation firm.

sharidas:
Updating the tests by checking values too...

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -553,7 +572,16 @@ private function createFileChunked($data) {
 
 				$this->fileView->unlockFile($targetPath, ILockingProvider::LOCK_SHARED);
 
-				return $info->getEtag();
+				$etag = $info->getEtag();
+				if ($etag !== null) {
+					$afterEvent = new GenericEvent(null, ['path' => $absPath]);
+					if (!$this->fileView->file_exists($targetPath)) {
```
PVince81: the file always exists after upload, so this check cannot be accurate and will always say "file exists"

please cache the value at the top

PVince81 (approved):

Issue #30742, opened 2018-03-10
Not clickable Url in password reset e-mail


prsnbrg:
### Steps to reproduce
1. After the login failed, click on the password reset url
2. Check my e-mails
3. Open the e-mail from oC

### Expected behaviour
Tell us what should happen
I can click on the url in the mail to open a website.

### Actual behaviour
Tell us what happens instead
I can´t click on the url in the e-mail to open a website to reset the password.

### Server configuration
**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**ownCloud version:** (see ownCloud admin page)
10.0.7.2
**Updated from an older ownCloud or fresh install:**
update
**Where did you install ownCloud from:**
web installer
**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results into https://gist.github.com/ and puth the link here.
```

**The content of config/config.php:**

```
Log in to the web-UI with an administrator account and click on
'admin' -> 'Generate Config Report' -> 'Download ownCloud config report'
This report includes the config.php settings, the list of activated apps
and other details in a well sanitized form.

or 

If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

*ATTENTION:* Do not post your config.php file in public as is. Please use one of the above
methods whenever possible. Both, the generated reports from the web-ui and from occ config:list
consistently remove sensitive data. You still may want to review the report before sending.
If done manually then it is critical for your own privacy to dilligently
remove *all* host names, passwords, usernames, salts and other credentials before posting.
You should assume that attackers find such information and will use them against your systems.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder.
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

#### LDAP configuration (delete this part if not used)

```
With access to your command line run e.g.:
sudo -u www-data php occ ldap:show-config
from within your ownCloud installation folder

Without access to your command line download the data/owncloud.db to your local
computer or access your SQL server remotely and run the select query:
SELECT * FROM `oc_appconfig` WHERE `appid` = 'user_ldap';


Eventually replace sensitive data as the name/IP-address of your LDAP server or groups.
```

### Client configuration
[...]

PR #31092, opened 2018-04-11
Update lost password email to use HTML link

joeyberkovitz:
## Description

Update the lost password email to send both HTML and plain text versions so that URL in email can be clicked.

## Related Issue

https://github.com/owncloud/core/issues/30742

## Motivation and Context

Solve issue https://github.com/owncloud/core/issues/30742

## How Has This Been Tested?

Update associated tests to add HTML support.

## Types of changes

## Checklist:

### DIFF
--- a/core/Controller/LostController.php
+++ b/core/Controller/LostController.php
@@ -336,7 +336,8 @@ class LostController extends Controller {
 			$message = $this->mailer->createMessage();
 			$message->setTo([$email => $user]);
 			$message->setSubject($this->l10n->t('%s password reset', [$this->defaults->getName()]));
-			$message->setPlainBody($msg);
+			$message->setPlainBody(\strip_tags($msg));
+			$message->setHtmlBody($msg);
 			$message->setFrom([$this->from => $this->defaults->getName()]);
 			$this->mailer->send($message);
 		} catch (\Exception $e) {
--- a/core/Controller/LostController.php
+++ b/core/Controller/LostController.php
@@ -331,12 +331,15 @@ class LostController extends Controller {
 		$tmpl = new \OC_Template('core', 'lostpassword/email');
 		$tmpl->assign('link', $link);
 		$msg = $tmpl->fetchPage();
+		$tmplAlt = new \OC_Template('core', 'lostpassword/altemail');
+		$tmplAlt->assign('link', $link);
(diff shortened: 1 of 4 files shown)


PVince81:
Thanks a lot for this PR!

Please note that some of these templates are customizable so I'm not sure if relying on "strip_tags" here would fit all scenarios.

A better way would be to use the same approach like for other emails: there are two template files, one called "mail.php" (for HTML) and "altmail.php" (for plain text). So here I suggest to do the same and have "core/templates/lostpassword/email.php" be the HTML email and create a new template "core/templates/lostpassword/altemail.php" with the plain text version.

Then wire this up with the code.

joeyberkovitz:
Sure.
I added a plain text template as you recommended.
I also updated both templates to be consistent with the format of the other email templates.

PVince81 (approved):
Tested, works :+1: 

cc @phil-davis @individual-it to include in acceptance tests

PVince81:
Hmm just noticed an extra ">" at the end of the HTML mail:
image

That one doesn't appear for other mails.

@joeyberkovitz mind verifying the markup ? I had a quick look but wasn't able to spot it

joeyberkovitz:
That's odd because I don't see the ">" in the source or in the email it sends out from my system, which is up to date with the Git repository. I'll try to debug some more later.
owncloud_lost_email

joeyberkovitz:
I cloned the repository into a separate directory and reinstalled Owncloud.
I wasn't able to reproduce the extra ">" using two different SMTP servers.
The only thing I did notice is that since the image is added by URL rather than being included, when mail clients like gmail try to pre-render the image, it'll fail if the URL isn't public. That might be a different issue though because it affects all of the existing email templates.

PVince81:
Thanks for confirming. Then maybe it was a bug in Mailhog's HTML renderer.

PVince81:
Mind backporting this to the stable10 branch ? Basically branch off stable10, cherry-pick your commits then submit PR to stable10 after checking it still works.

joeyberkovitz:
I created the backport PR

phil-davis:
Backport stable10 #31144

individual-it:
@paurakhsharma is working on acceptance tests in #31168

PR #31102, opened 2018-04-12
Don`t apply checksum wrapper to mixed r/w streams

IljaN:
## Description
Don`t apply checksum wrapper on mixed read/write streams because this could corrupt the checksum context if the stream is for example seeked or rewinded.

## Motivation and Context
Might be the potential source of checksum issues we are having. 

## Types of changes

## Checklist:

### DIFF
--- a/tests/lib/Files/Storage/Wrapper/ChecksumTest.php
+++ b/tests/lib/Files/Storage/Wrapper/ChecksumTest.php
@@ -36,6 +36,8 @@ class ChecksumTest extends \Test\TestCase {
 	 */
 	private $instance;
 
+	private $testFileName;
+
 	const TEST_DATA = 'somedata';
 	const EXPECTED_CHECKSUMS = 'SHA1:efaa311ae448a7374c122061bfed952d940e9e37 MD5:aefaf7502d52994c3b01957636a3cdd2 ADLER32:0f2c034f';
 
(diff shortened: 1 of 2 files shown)


IljaN:
Thinking this further, if a read and write occurs on the same stream we might just want to reset the checksum and don`t compute one at all because incorrect checksum will occur in this scenario anyways...

guruz:
I'm not a PHP guy, but maybe it's also feasible to check the ```$mode``` in ```stream_open``` and only allow reading/writing based on the mode.
(And assume it's not ```rw```or whatever)?

IljaN:
@guruz Yeah good plan need to search trough the codebase where we open rw streams

PVince81:
How would it be possible to read from a stream being written already ?

Usually that's only possible if fopen is called with a special mode.
Normally all known code paths use "r" or "w" but if there are that use something else we should simply disable computing checksums, maybe even refrain from applying the checksum wrapper.

IljaN:
I think this was guruz`s idea... also another potential pitfall which comes to mind might be rewind(), e.g read, rewind, read again...

lib/private/Files/Stream/Checksum.php
```diff
@@ -136,7 +136,7 @@ private function updateHashingContexts($data) {
 	private function stripPartialFileExtension($path) {
 		$extension = \pathinfo($path, PATHINFO_EXTENSION);
 
-		if ($extension === 'part') {
+		if ( $extension === 'part') {
```
patrickjahns: why?

tests/lib/Files/Storage/Wrapper/ChecksumTest.php
```diff
@@ -46,40 +48,43 @@ public function setUp() {
...
-		$metaData = $this->instance->getMetaData('/foo.txt');
+		$metaData = $this->instance->getMetaData($this->testFileName);
 
 		$this->assertArrayHasKey('checksum', $metaData);
 		$this->assertEquals(self::EXPECTED_CHECKSUMS, $metaData['checksum']);
 	}
 
 	public function testReadFromFileHandleOnNewFileCalculatesChecksum() {
+		$this->sourceStorage->file_put_contents($this->testFileName, self::TEST_DATA);
+        // Make file available in filecache (without checksum)
```
patrickjahns: indentation

IljaN:
@ownclouders rebase

PVince81 (approved):
Code looks good :+1:

PVince81:
please resolve the conflict

IljaN:
Conflict resolved

phil-davis:
Backport ``stable10`` #31619

PR #31103, opened 2018-04-12
Feature/no upgrade on patch level changes

DeepDiver1975:
## Description
As of today any new release of an app will trigger the upgrade mechanism.
This will put the while instance into maintenance mode.

There are situations where this is not necessary because only the source code was changed but
no database schema or any other migration is necessary to be applied

We follow the idea of semver where changes to the patch level only do no longer trigger the upgrade mechanism.

1.2.1 -> 1.2.2 will not trigger the upgrade
1.2.3 -> 1.3.0 will trigger the upgrade

## Motivation and Context
Allow simple code replacements via app releases.

## How Has This Been Tested?
- unit tests

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,6 +68,7 @@
         "symfony/translation": "^3.4",
         "zendframework/zend-inputfilter": "^2.8",
         "zendframework/zend-servicemanager": "^3.3",
-        "zendframework/zend-validator": "^2.10"
+        "zendframework/zend-validator": "^2.10",
+        "composer/semver": "^1.4"
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "7c4821ccb61f5af66fefb23ed385ac34",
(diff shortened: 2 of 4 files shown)


felixboehm:
Nice. Breaking change for app developers, needs announcement and documentation.

felixboehm:
code and test coverage look good

PVince81:
conflicts.

fine by me but shouldn't we then do the same with core and have the next version be 10.1 then to respect semver and the same semantics ? so the third digit would be mostly for hotfixes or releases that need no DB changes ?

PVince81:
if an instance is updated from OC 10.0.1 to OC 11 where this PR is in, and if there is any app that had DB changes within a patch version, those DB changes would not be updated any more.

I suggest to add an additional version check: if we're upgrading from a version pre OC-11, then still execute all app DB changes one last time even for patch version updates.

PVince81:
another possibility would be to declare all apps incompatible with OC 11 and force a re-release of all such apps with proper major/minor version update to make sure any remaining DB changes are triggered when upgrading

PVince81 (approved):
fix conflict and let's move forward :+1:

PVince81:
@DeepDiver1975 please backport, then we'll hold until we have a decision for the timeline for 10.1

Issue #26585, opened 2016-11-08
Avatar Performance 


butonic:
In my data/avatars folder i see a lot of empty 2 char folders. likely created by https://github.com/owncloud/core/blob/master/lib/private/AvatarManager.php#L98

They somehow also made it into the oc_filecache with a size of -1

this causes the avatar folder to also stay at a size of -1 because the folder is only scanned shallow: all empty folders stay at -1.

This causes a (shallow) filescan when trying to get an avatar image via eg
http://localhost/owncloud/index.php/avatar/admin/28. In my case this takes 400ms for 240 2 char folders in the avatar folder.

AFAICT introduced with https://github.com/owncloud/core/pull/26124

This is made worse by the `Cache-Control: no-store, no-cache, must-revalidate` header of the http://localhost/owncloud/index.php/avatar/admin/28 response because we call the url twice when navigating to http://localhost/owncloud/index.php/apps/files/?dir=/ @PhilippSchaffrath has details on why we do that.

The header is set in https://github.com/owncloud/core/blob/master/core/Controller/AvatarController.php#L138. For users without an avatar image also because the DataResponse in https://github.com/owncloud/core/blob/master/core/Controller/AvatarController.php#L124 and https://github.com/owncloud/core/blob/master/core/Controller/AvatarController.php#L130 is converted to a JSONResponse in https://github.com/owncloud/core/blob/master/lib/public/AppFramework/Controller.php#L92 which overwrites the Cache-Control with the default from the JSONResponse / Response ('Cache-Control' => 'no-cache, must-revalidate').

Replacing the DataResponse with a JSONResponse would allow caching the avatar, eg:

```diff
--- a/core/Controller/AvatarController.php
+++ b/core/Controller/AvatarController.php
@@ -28,6 +28,7 @@ namespace OC\Core\Controller;
 
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\JSONResponse;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\DataDisplayResponse;
 use OCP\Files\NotFoundException;
@@ -121,7 +122,7 @@ class AvatarController extends Controller {
                        $resp->setETag($avatar->getEtag());
                } catch (NotFoundException $e) {
                        $user = $this->userManager->get($userId);
-                       $resp = new DataResponse([
+                       $resp = new JSONResponse([
                                'data' => [
[...]

Issue #25770, opened 2016-08-11
[Feature Request] adding column "ref_storage" in table oc_filecache


mmattel:
Currently we have in oc_filecache the information to which mount point a file belongs to.
This is column `storage`. When a file is created or deleted, the corresponding entry in oc_filecache is updated. The same happens when a mount point is added or deleted. In the latter case where you delete a mount, it is easy to clean up the table (respectively clean up the filesystem with the queried data).
But when files are created for a user which have a reference to another mount, this system does not longer work as it leaves orphans.

**Example before:**
When thumbnails for pictures are created, they create a file entry in oc_filecache where storage is the id of the users home. In case you delete the mount of the original (reference), you leave in the current implementation all the thumbs which have no reference anymore. The only thing you could do is manually delete all cache entries for all users and rescan.

**This FR is to add a new column called `ref_storage` to oc_filecache.**

**Change:**
Whenever a file is created by owncloud which has a source reference file, "ref_storage" gets populated. Currently this is the case for thumbs but can be expanded easily.

**Benefit:**
When a mount is deleted, you can query all files in oc_filecache column `ref_storage`, delete these files in the filesystem with reference `storage` and cleanup oc_filecache. No orphans left (user idependent!).
Because the removal of the mount is then a process running in defined steps, a unforseeable interruption has no bad impact as it can be restarted.

**Example after:**
`storage` this is the ID of the mount point where the file is physically located
`ref_storage` this the ID of the mount point of the source where the data for the physical file was coming from (the reference point).

target = function {source}

physical file location = target: location reference = storage
referenced location = source: location reference = ref_storage

table oc_storages, mount ID´s:
user_x: home_storage_location = ID:1
SMB: smb_share = ID:2

Examples for thumbnails for user_x:
Thumb_1: storage = 1 and ref_storage =1
Thumb_2: storage = 1 and ref_storage =2
Thumb_3: storage = 1 and ref_storage =2
ect...

If the user has more than one mount, you will get possibly different ref_storage values for thumbnails. But this is exactly what we want.

**On a generic view:**
`storage` must be populated, `ref_storage` can be populated.

If this would get implemented, we can create additional PR´s that:
- populating ref_storage when thumbs are created
- at mount point deletion, delete the corresponding files created by owncloud too.

This FR references the original discussion in #25033 Deleting an (external) storage mount leaves thumnails created in the filesystem

mmattel:
@PVince81 @DeepDiver1975

[...]

Issue #22758, opened 2016-03-01
Disable etag propagation outside of "files/" in home storage


PVince81:
@icewind1991 that would be the next step to improve etag propagation performance.

@schiesbn

PVince81:
@butonic didn't you want to look into etag propagation perf ? That could be a start

PVince81:
@mrow4a in case you're interested at some point

mrow4a:
Can you add more details here, some context and points in the code?

PVince81:
Could be as simple as extending this array: https://github.com/owncloud/core/blob/stable10/lib/private/Files/Cache/HomePropagator.php#L34

Or make it a whitelist instead of a blacklist

PR #31105, opened 2018-04-12
Introduce IStorage based Node, Folder & File, make avatar manager wor…

butonic:
…k directly on storage to bypass cache

## Description
- make AvatarManager work directly on Storage instead of Files API
  - bypasses unnecessary cache
- add Storage based implementation of Files API Node, Folder & File

## Related Issue

https://github.com/owncloud/core/issues/26585
https://github.com/owncloud/core/issues/22758
https://github.com/owncloud/core/issues/25770#issuecomment-383013349

## Motivation and Context
user page takes 30 sec to load ... unbearably slooooww 

## How Has This Been Tested?
Profiled the users page

## Screenshot:
bildschirmfoto von 2018-04-12 14-29-19

We were doing quite a lot of queries on the filecache for this

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Avatar.php
+++ b/lib/private/Avatar.php
@@ -1,5 +1,6 @@
 <?php
 /**
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
  * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
  * @author Christopher Schäpers <kondou@ts.unde.re>
  * @author Lukas Reschke <lukas@statuscode.ch>
@@ -28,11 +29,11 @@
 
 namespace OC;
 
+use OC\Files\Storage\File;
 use OC\User\User;
-use OCP\Files\Folder;
-use OCP\Files\File;
 use OCP\Files\NotFoundException;
-use OCP\Files\NotPermittedException;
+use OCP\Files\Storage\IStorage;
+use OCP\Files\StorageNotAvailableException;
 use OCP\IAvatar;
 use OCP\IImage;
 use OCP\IL10N;
@@ -44,28 +45,35 @@ use OCP\ILogger;
  */
 
 class Avatar implements IAvatar {
-	/** @var Folder */
-	private $folder;
+	/** @var IStorage */
+	private $storage;
 	/** @var IL10N */
 	private $l;
 	/** @var User */
 	private $user;
 	/** @var ILogger  */
 	private $logger;
+	/** @var string */
+	private $path;
 
 	/**
 	 * constructor
 	 *
-	 * @param Folder $folder The folder where the avatars are
+	 * @param IStorage $storage The storage where the avatars are
 	 * @param IL10N $l
 	 * @param User $user
 	 * @param ILogger $logger
 	 */
-	public function __construct(Folder $folder, IL10N $l, $user, ILogger $logger) {
-		$this->folder = $folder;
+	public function __construct(IStorage $storage, IL10N $l, User $user, ILogger $logger) {
+		$this->storage = $storage;
 		$this->l = $l;
 		$this->user = $user;
 		$this->logger = $logger;
+		$this->path = $this->buildAvatarPath();
+	}
+
+	private function buildAvatarPath() {
+		return \substr_replace(\substr_replace(\md5($this->user->getUID()), '/', 4, 0), '/', 2, 0);
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


butonic:
- [ ] need to test performance on objectstore as it uses the db for file metadata.

DeepDiver1975:
Physically the avatars are stored in the same location?
Does this work with objectstore?
Can we learn anything for other operations?

PVince81:
conflicts

@butonic finish for 10.0.9 or "triage" for later / rescheduling ?

butonic:
> Physically the avatars are stored in the same location?
Yes, the PR just retrieves the storage for the location and then works on that

> Does this work with objectstore?
Yes, performance is acceptable. far better than without this pr

> Can we learn anything for other operations?
I assume we can use this for thumbnails and other things as well -> different PR see https://github.com/owncloud/core/issues/25770#issuecomment-383013349

jvillafanez:
Just a couple of questions:
* It seems possible (although it won't be supported in the short term) to store the avatar in a different storage. This would mean that even though the primary storage is S3 we might still use local storage for the avatars, or maybe another more suitable storage. As said, this seems doable, but there are still quite some changes to be made to support this.
* I'm not sure if the avatar manager should have direct access to the storage or if it should use something a bit different. Caching might still be needed to improve the performance. By using a direct access to the storage, caching won't be possible because the storage would have to assume that responsability, something that won't happen.

    We might use a `AvatarStorage` wrapping a `IStorage` to include caching there. Short term, it's fine if the `AvatarStorage` doesn't cache anything and just call directly to the storage, but we add caching easily there if needed.

lib/private/Avatar.php
```diff
@@ -151,48 +160,49 @@ public function getFile($size) {
 		$ext = $this->getExtension();
 
 		if ($size === -1) {
-			$path = 'avatar.' . $ext;
+			$path = "$this->path/avatar.$ext";
```
DeepDiver1975: {} missing?

{$this->path}

tests/lib/Files/Storage/FolderTest.php
```diff
@@ -0,0 +1,183 @@
+<?php
+/**
+ * Copyright (c) 2013 Robin Appelman <icewind@owncloud.com>
```
DeepDiver1975: no

tests/lib/Files/Storage/NodeTest.php
```diff
@@ -0,0 +1,317 @@
+<?php
+/**
+ * Copyright (c) 2013 Robin Appelman <icewind@owncloud.com>
```
DeepDiver1975: no

DeepDiver1975 (dismissed):

DeepDiver1975:
not forgotten ... will take care - THX

lib/private/Avatar.php
```diff
@@ -44,28 +45,35 @@
...
 	/**
 	 * constructor
 	 *
-	 * @param Folder $folder The folder where the avatars are
+	 * @param IStorage $storage The storage where the avatars are
 	 * @param IL10N $l
 	 * @param User $user
 	 * @param ILogger $logger
 	 */
-	public function __construct(Folder $folder, IL10N $l, $user, ILogger $logger) {
```
DeepDiver1975: You could have inserted an instance of OC\Files\Storage\Folder which should give us the same performance boost bu reduce the number of changed lines ... just saying.

lib/private/Files/Storage/File.php
```diff
@@ -0,0 +1,88 @@
...
+use OCP\Files\File as FilesFile;
+
+class File extends Node implements FilesFile {
+
+	/**
+	 * Get the content of the file as string
+	 *
+	 * @return string
+	 * @throws \OCP\Files\StorageNotAvailableException
+	 * @since 6.0.0
```
VicDeo: 6.0.0?

DeepDiver1975 (approved):

DeepDiver1975:
@PVince81 I just reverted this - please schedule this item to get it ready in the next cycle .. THX

cdamken:
> @PVince81 I just reverted this - please schedule this item to get it ready in the next cycle .. THX

do you mean 10.0.10?

DeepDiver1975:
> do you mean 10.0.10?

no - we need to readdress this in one of the next server sprints - the fix was not complete.
once we have this in master/oc11 we can decide on oc10 backport. since this involves api changes I tend to :-1: on backport

cdamken:
Thanks

PR #31113, opened 2018-04-13
[stable10] check loglevel before interpolating message

phil-davis:
Backport #31072 

This is an easy backport and seems like "a good thing", so I made the backport. Feel free to close if it is not to be backported.

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -265,7 +265,6 @@ class Log implements ILogger {
 			$extraFields = $context['extraFields'];
 			unset($context['extraFields']);
 		}
-		array_walk($context, [$this->normalizer, 'format']);
 
 		if (isset($context['app'])) {
 			$app = $context['app'];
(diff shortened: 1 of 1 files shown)


DeepDiver1975 (approved):

DeepDiver1975:
approved - merging is upon @PVince81 since stable10 is in freeze for 10.0.8

DeepDiver1975:
From my pov we can abandon this backport - the performance benefit is lost due to the event dispatcher

@butonic

PVince81:
if Symfony provides a way to detect how many listeners are registered, the optimization could be kept if there is no listener at all. but then, increases code complexity

phil-davis:
I resolved the latest conflicts. 

If somebody would like this for stable10, then merge it.

Otherwise close it.

PVince81:
closed, thanks. we need to resolve https://github.com/owncloud/core/pull/31135 first before talking backports

PR #31120, opened 2018-04-13
[stable10] Unified symfony event for updates happening in share

sharidas:
A unified symfony event for updates happening in share.
This change covers the updates for updates in:
a) share name
b) password change
c) update in expiry date
d) update of permission

## Description

A unified event to handle updates in share, like:
a) share name update
b) share password update
c) update of expiry date
d) permissions

## Motivation and Context

A unified event to handle updates in share, like:
a) share name update
b) share password update
c) update of expiry date
d) permissions

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -749,6 +749,9 @@ class Manager implements IManager {
 		$provider = $this->factory->getProviderForType($share->getShareType());
 		$share = $provider->update($share);
 
+		$shareAfterUpdateEvent = new GenericEvent(null);
+		$shareAfterUpdateEvent->setArgument('shareobject', $share);
+		$update = false;
 		if ($expirationDateUpdated === true) {
 			\OC_Hook::emit('OCP\Share', 'post_set_expiration_date', [
 				'itemType' => $share->getNode() instanceof \OCP\Files\File ? 'file' : 'folder',
@@ -756,6 +759,9 @@ class Manager implements IManager {
 				'date' => $share->getExpirationDate(),
 				'uidOwner' => $share->getSharedBy(),
 			]);
+			$shareAfterUpdateEvent->setArgument('expirationdateupdated', true);
+			$shareAfterUpdateEvent->setArgument('oldexpirationdate', $originalShare->getExpirationDate());
+			$update = true;
 		}
 
 		if ($share->getPassword() !== $originalShare->getPassword()) {
@@ -766,6 +772,8 @@ class Manager implements IManager {
 				'token' => $share->getToken(),
 				'disabled' => is_null($share->getPassword()),
 			]);
+			$shareAfterUpdateEvent->setArgument('passwordupdate', true);
+			$update = true;
 		}
 
 		if ($share->getPermissions() !== $originalShare->getPermissions()) {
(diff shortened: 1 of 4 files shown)


lib/private/Share20/Manager.php
```diff
@@ -783,8 +787,11 @@ public function updateShare(\OCP\Share\IShare $share) {
 				'permissions' => $share->getPermissions(),
 				'path' => $userFolder->getRelativePath($share->getNode()->getPath()),
 			]);
+			$shareAfterUpdateEvent->setArgument('permissionupdate', true);
+			$shareAfterUpdateEvent->setArgument('path', $userFolder->getRelativePath($share->getNode()->getPath()));
 		}
 
+		$this->eventDispatcher->dispatch('share.afterupdate', $shareAfterUpdateEvent);
```
PVince81: what if nothing got updated ?

lib/private/Share20/Manager.php
```diff
@@ -749,13 +749,16 @@ public function updateShare(\OCP\Share\IShare $share) {
...
+		$shareAfterUpdateEvent = new GenericEvent(null);
+		$shareAfterUpdateEvent->setArgument('shareobject', $share);
 		if ($expirationDateUpdated === true) {
 			\OC_Hook::emit('OCP\Share', 'post_set_expiration_date', [
 				'itemType' => $share->getNode() instanceof \OCP\Files\File ? 'file' : 'folder',
 				'itemSource' => $share->getNode()->getId(),
 				'date' => $share->getExpirationDate(),
 				'uidOwner' => $share->getSharedBy(),
 			]);
+			$shareAfterUpdateEvent->setArgument('expiredateupdated', true);
```
PVince81: the pattern for name was "oldname" and "newname", why not use "oldexpiredate" or "newexpiredate" then for consistency ? only "oldpassword" would not be possible

please settle on a single consistent format

lib/private/Share20/Manager.php
```diff
@@ -783,8 +791,14 @@ public function updateShare(\OCP\Share\IShare $share) {
 				'permissions' => $share->getPermissions(),
 				'path' => $userFolder->getRelativePath($share->getNode()->getPath()),
 			]);
+			$shareAfterUpdateEvent->setArgument('permissionupdate', true);
```
PVince81: "oldpermission" ?

PVince81:
I wonder whether we really need the boolean flags. A listener can simply check whether there is a value "oldname" to know that the name was changed.

In an ideal object oriented world I'd want to have "oldShare" and "newShare" objects but I'm not sure you can get ahold of the old instance. Or would need a copy. I agree to stay with separate attributes for "oldXXX" then.

lib/private/Share20/Manager.php
sharidas: added 'oldpermissions'

sharidas:
@PVince81 Ready for review.

sharidas:
> I wonder whether we really need the boolean flags. A listener can simply check whether there is a value "oldname" to know that the name was changed.

The reason why I added boolean flag, is because it would be better to trigger event only when required.

lib/private/Share20/Manager.php
PVince81: "expiredate" vs "expirydate" vs "expirationdate"

three different words for the same thing: let's use the "expiration" variant

sharidas: Updated the PR with "exipration"

PVince81 (approved):
:+1: thanks

PVince81:
@sharidas please forward port to master (PR was missing "backport-request" label)

sharidas:
Forward port to master : https://github.com/owncloud/core/pull/31138

PR #31121, opened 2018-04-13
[stable10] Ability to log extra fields + trigger event when logging

PVince81:
## Description
1) If a logger supports it like the default logger, additional fields can
now be added to the base JSON object.

Built to still support old loggers so this shouldn't break any API contracts.

2) Trigger an event before and after logging an entry, independently from log level

This makes it possible for apps to react on log events.
Could also help with the testing app to capture log entries.

## Related Issue
None

## Motivation and Context
Some log analyzers prefer to have the fields directly appearing in the base JSON object instead of serialized into strings.

## How Has This Been Tested?
Add `\OC::$server->getLogger()->error("Test extra fields", ['app' => 'test', 'extraFields' => ['hello' => 'field', 'another' => 'field']]);` to Sabre\Directory constructor and open the web UI.
Then check the logs:
```
{"reqId":"S9ulvAzpHQcTLEtVRMPf","level":3,"time":"2018-04-13T13:22:51+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"test","method":"PROPFIND","url":"\/owncloud\/remote.php\/webdav\/","message":"Test extra fields","hello":"field","another":"field"}
```

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -37,6 +37,8 @@ use InterfaSys\LogNormalizer\Normalizer;
 use \OCP\ILogger;
 use OCP\IUserSession;
 use OCP\Util;
+use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * logging utilities
@@ -56,11 +58,21 @@ class Log implements ILogger {
 	/** @var SystemConfig */
 	private $config;
 
+	/** @var EventDispatcherInterface */
+	private $eventDispatcher;
+
 	/** @var boolean|null cache the result of the log condition check for the request */
 	private $logConditionSatisfied = null;
 
 	/** @var Normalizer */
 	private $normalizer;
+
+	/**
+	 * Flag whether we are within the event block
+	 *
+	 * @var bool
+	 */
+	private $inEvent = false;
 
 	protected $methodsWithSensitiveParameters = [
 		// Session/User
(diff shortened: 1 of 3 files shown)


PVince81:
Added https://github.com/owncloud/core/pull/31121/commits/95cd4e4e895dd9f5e9ed373ce109535d0e680a58 and updated OP with section 2

butonic:
- [ ] allow syslog logger to log these extra attributes? see https://github.com/owncloud/core/pull/31124

PVince81:
> allow syslog logger to log these extra attributes

separately

PVince81:
I force pushed matching unit tests in each of the former commits

butonic:
when forward porting, this will interfere with https://github.com/owncloud/core/pull/31072

backport https://github.com/owncloud/core/pull/31072 and rebase this PR on top of it?

butonic (changes requested):
when forward porting, this will interfere with #31072

backport #31072 and rebase this PR on top of it?

DeepDiver1975 (approved):

DeepDiver1975:
failing jenkins can be ignored -> merging

DeepDiver1975:
> when forward porting, this will interfere with #31072
> 
> backport #31072 and rebase this PR on top of it?

this optimization is actually reverted now with this PR

PVince81:
> this optimization is actually reverted now with this PR

To clarify: it is a requirement to be able to send the event with the log message even if the message will not be logged with the current log level. It is for external logging apps that might have their own levels.

If interpolation is a problem we could decide to remove it from the event and expect the hook listener to redo the interpolation if needed.

mmattel:
@PVince81 @DeepDiver1975 @settermjd 
are these new extra fields + events something that should be documented, eg in https://github.com/owncloud/documentation/blob/master/developer_manual/app/fundamentals/logging.rst or in any othe place?

PVince81:
@mmattel good point. Raised here: https://github.com/owncloud/documentation/issues/4162

PR #31135, opened 2018-04-16
Ability to log extra fields + trigger event when logging

DeepDiver1975:
*port of #31121

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -37,6 +37,8 @@ use InterfaSys\LogNormalizer\Normalizer;
 use \OCP\ILogger;
 use OCP\IUserSession;
 use OCP\Util;
+use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * logging utilities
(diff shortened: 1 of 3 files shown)


lib/private/Log.php
```diff
@@ -309,13 +339,48 @@ public function log($level, $message, array $context = []) {
...
 		}
 
-		if ($level >= $minLevel) {
+		$skipEvents = false;
+		// avoid infinite loop in case an event handler logs something
+		if ($this->inEvent) {
+			$skipEvents = true;
+		}
+
+		$formattedMessage = $this->interpolate($message, $context);
```
butonic: interpolation should only happen after the event has been handled. The app might want to change the context before interpolating. move this inside the `if ($level >= $minLevel) {`

butonic (changes requested):

lib/private/Log.php
DeepDiver1975: see my comment on the original PR of this event dispatching of log messages.
the performance optimization is obsolete because the formatted message is part of the event which is dispatched in any case.

DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

PVince81:
added commit from https://github.com/owncloud/core/pull/31247

tomneedham:
IMO remove optimisation - separate PR, target master then backport, like usual. Now we're messing with a forward port and it's going to end in tears.

patrickjahns:
@DeepDiver1975 @PVince81 - any update?

PVince81:
blocks forward port of https://github.com/owncloud/core/pull/31623

lib/private/Log.php
PVince81: We'd then need to expose the interpolation method to hook listeners.

I agree with @DeepDiver1975 and would like to move forward and merge this

DeepDiver1975: redoing the interpolation on all listeners will duplicate the calls - not optimal either.

DeepDiver1975:
merging this to move on

PR #30695, opened 2018-03-05
[stable10] Adding repair step for sub share

sharidas:
Adding repair step for sub share.
When share table due to some magic gets
duplicate entries ( except the id or primary
key ), then this repair step comes for rescue.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Remove the duplicate entries from the `share` table which can cause exception described in the issue.

## Related Issue

https://github.com/owncloud/core/issues/27990

## Motivation and Context

For some magical reason, if the share table gets populated with duplicate entries then we would have to remove the duplicate entry. Keeping the row with minimum id. Else exception would be thrown as mentioned in the issue.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #27990, opened 2017-05-24
Update error when duplicate sub-shares exist


lukovas:
There was an error updating. Please, help!

OC\Share20\Exception\ProviderException: Parent of share should be unique

Detailed logs:

Preparing update
Set log level to debug
Turned on maintenance mode
Updating database schema
Updated database
Updated "activity" to 2.3.4
[1 / 28]: Drop old database table: calendar_calendars
[2 / 28]: Drop old database table: calendar_objects
[3 / 28]: Drop old database table: calendar_share_calendar
[4 / 28]: Drop old database table: calendar_share_event
[5 / 28]: Drop old database table: file_map
[6 / 28]: Drop old database table: foldersize
[7 / 28]: Drop old database table: fscache
[8 / 28]: Drop old database table: gallery_sharing
[9 / 28]: Drop old database table: locks
[10 / 28]: Drop old database table: log
[11 / 28]: Drop old database table: media_albums
[12 / 28]: Drop old database table: media_artists
[13 / 28]: Drop old database table: media_sessions
[14 / 28]: Drop old database table: media_songs
[15 / 28]: Drop old database table: media_users
[16 / 28]: Drop old database table: permissions
[17 / 28]: Drop old database table: pictures_images_cache
[18 / 28]: Drop old database table: principalgroups
[19 / 28]: Drop old database table: principals
[20 / 28]: Drop old database table: queuedtasks
[21 / 28]: Drop old database table: sharing
[22 / 28]: Drop old database table: clndr_calendars
[23 / 28]: Drop old database table: clndr_objects
[24 / 28]: Drop old database table: clndr_share_event
[25 / 28]: Drop old database table: clndr_share_calendar
[26 / 28]: Drop old database table: clndr_repeat
[27 / 28]: Drop old database table: contacts_addressbooks
[28 / 28]: Drop old database table: contacts_cards
[28 / 28]: Drop old database tables
[1 / 4]: Remove old (< 9.0) calendar/contact shares
[2 / 4]: Remove old (< 9.0) calendar/contact shares
[3 / 4]: Remove old (< 9.0) calendar/contact shares
[4 / 4]: Remove old (< 9.0) calendar/contact shares
[4 / 4]: Remove old (< 9.0) calendar/contact shares
[1 / 2]: Fix permissions so avatars can be stored again
[1 / 1]: Fix permissions so avatars can be stored again
[1 / 44]: Move user avatars outside the homes to the new location
[...]

PR #31146, opened 2018-04-16
[stable10] Update the sql query to delete duplicate sub share

sharidas:
## Description

`group by id` was causing the results to get skipped and hence returning empty results even if duplicated shares were there. This change would help to solve the issue.

## Motivation and Context

`group by id` was causing the results to get skipped and hence returning empty results even if duplicated shares were there. This change would help to solve the issue.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Repair/RepairSubShares.php
+++ b/lib/private/Repair/RepairSubShares.php
@@ -48,55 +48,75 @@ class RepairSubShares implements IRepairStep {
 	}
 
 	/**
-	 * Set query to remove duplicate rows.
-	 * i.e, except id all columns are same for oc_share
-	 * Also set query to select rows which have duplicate rows of share.
+	 * Get query builder to select rows which have duplicate rows of share
+	 * @return IQueryBuilder
 	 */
-	private function setRemoveAndSelectQuery() {
+	private function getSelectQueryToDetectDuplicatesBuilder() {
 		/**
 		 * Retrieves the duplicate rows with different id's
-		 * of oc_share
+		 * of oc_share. The query would look like:
+		 * SELECT id
+		 * FROM oc_share
+		 * WHERE id NOT IN (
+		 * 		SELECT MIN(id)
+		 * 		FROM oc_share
+		 * 		GROUP BY share_with, parent
+		 * )
+		 * AND share_type=2;
 		 */
 		$builder = $this->connection->getQueryBuilder();
-		$builder
-			->select('id', 'parent', $builder->createFunction('count(*)'))
+		$subQuery = $this->connection->getQueryBuilder();
+		$subQuery
+			->select($subQuery->createFunction('MIN(`id`)'))
 			->from('share')
-			->where($builder->expr()->eq('share_type', $builder->createNamedParameter(2)))
-			->groupBy('parent')
-			->addGroupBy('id')
-			->addGroupBy('share_with')
-			->having('count(*) > 1')->setMaxResults(1000);
+			->groupBy('share_with')
+			->addGroupBy('parent');
 
-		$this->getDuplicateRows = $builder;
+		$builder->select('id')
+			->from('share')
+			->where($builder->expr()->notIn('id', $builder->createFunction($subQuery->getSQL())))
+			->andWhere($builder->expr()->eq('share_type', $builder->createNamedParameter(2)));
+		return $builder;
+	}
 
+	/**
+	 * Get query builder to delete rows which have duplicate rows of share based
+	 * on ids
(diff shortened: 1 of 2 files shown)


sharidas:
The reason for this PR: https://github.com/owncloud/core/pull/30695#issuecomment-381689933.
Verified as follows:

```
➜  owncloud3 git:(fix-repair-subshare-stable10) ✗ ./occ maintenance:repair
 - Repair mime types
 - Detect file cache entries with path that does not match parent-child relationships
 - Generate ETags for file where no ETag is present.
     - ETags have been fixed for 0 files/folders.
 - Clean tags and favorites
     - 0 tags of deleted users have been removed.
     - 0 tags for delete files have been removed.
     - 0 tag entries for deleted tags have been removed.
     - 0 tags with no entries have been removed.
 - Drop old database tables
 28/28 [============================] 100%
 - Drop old background jobs
 - Remove getetag entries in properties table
     - Removed 0 unneeded "{DAV:}getetag" entries from properties table.
 - Repair outdated OCS IDs
 - Repair invalid shares
 - Remove old share propagation app entries
 - Move user avatars outside the homes to the new location
 - Remove shares of a users root folder
 - Repair unmerged shares
 - Disable extra themes
 - Repair sub shares
     - Removed 1 shares where duplicate rows where found
 - Remove invalid calendar and addressbook shares
➜  owncloud3 git:(fix-repair-subshare-stable10) ✗
```

jvillafanez:
```
mysql> select id, parent, count(*) from share where share_type = 2 group by parent, share_with having count(*) > 1;
ERROR 1055 (42000): Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'owncloud.share.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

Quoting https://dev.mysql.com/doc/refman/5.7/en/group-by-handling.html:
> MySQL 5.7.5 and up implements detection of functional dependence. If the ONLY_FULL_GROUP_BY SQL mode is enabled (which it is by default), MySQL rejects queries for which the select list, HAVING condition, or ORDER BY list refer to nonaggregated columns that are neither named in the GROUP BY clause nor are functionally dependent on them. (Before 5.7.5, MySQL does not detect functional dependency and ONLY_FULL_GROUP_BY is not enabled by default. For a description of pre-5.7.5 behavior, see the MySQL 5.6 Reference Manual.)

I don't think the query will work for other DBs since the ONLY_FULL_GROUP_BY is intended to match the standard, so disabling the setting in MySQL shouldn't be an option.

jvillafanez:
PostgreSQL 9.6.8 also have problems running the query:
```
owncloud=# select id, parent, count(*) from oc_share where share_type = 2 group by parent, share_with having count(*) > 1;
ERROR:  column "oc_share.id" must appear in the GROUP BY clause or be used in an aggregate function
LINE 1: select id, parent, count(*) from oc_share where share_type =...
               ^
```

jvillafanez:
@sharidas I guess you've tested with mysql < 5.7.5 (or the corresponding mariadb version).
We need another option.

PVince81:
does anything break without this PR ? from my understanding the repair step just doesn't do its job and doesn't break anything, so this task could be considered a bug and deferred to 10.0.9

jvillafanez:
> does anything break without this PR ? from my understanding the repair step just doesn't do its job and doesn't break anything, so this task could be considered a bug and deferred to 10.0.9

Without this PR, that's should be the behaviour. The query won't detect any duplicate

PVince81:
to be rescheduled

sharidas:
@PVince81  estimate: 2-3 md for this task.

butonic:
based on https://stackoverflow.com/questions/18932/how-can-i-remove-duplicate-rows#comment3999923_18949

lets use 
```sql
SELECT id FROM oc_share WHERE id NOT IN (SELECT MIN(id) FROM oc_share GROUP BY share_with, parent) AND share_type=2;
```

to find all the duplicates that should be deleted

sharidas:
Forward port to master branch: https://github.com/owncloud/core/pull/31781

sharidas:
Updated this PR according the suggestions made at https://github.com/owncloud/core/pull/31781.
Tested the code, and during the upgrade process, the duplicate share entry was removed and similar output was observed as in https://github.com/owncloud/core/pull/31781#issuecomment-402459053

PVince81 (approved):
:+1:

PVince81:
for manual testing, use steps from https://github.com/owncloud/core/issues/27990#issuecomment-357899190

PR #31148, opened 2018-04-16
[stable10] API tests to create and use an app password

phil-davis:
## Description
1) Add test steps that generate an app password and can use it to access API endpoints
2) Add scenarios to generate an app password and use it
3) Add test steps that can turn on token_auth_enforced
4) Add a few scenarios to check what can be done and what is blocked by token_auth_enforced

## Motivation and Context
Have some automated tests that check that an app password can be created and used.

## How Has This Been Tested?
Local test runs.

## Types of changes
- [x] Acceptance tests

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/Auth.php
+++ b/tests/acceptance/features/bootstrap/Auth.php
@@ -30,6 +30,7 @@ require __DIR__ . '/../../../../lib/composer/autoload.php';
 trait Auth {
 
 	private $clientToken;
+	private $appToken;
 
 	/**
 	 * @BeforeScenario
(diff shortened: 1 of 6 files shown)


tests/acceptance/features/bootstrap/Auth.php
```diff
@@ -200,4 +253,59 @@ public function aNewBrowserSessionForHasBeenStarted($user) {
...
+				$value
+			]
+		);
+
+		// Remember that we set this value, so it can be removed after the scenario
+		$this->tokenAuthHasBeenSet = true;
+
+		// It takes some time for the change in config.php to really be
+		// recognised at the server.
+		sleep(5);
```
phil-davis: Waiting is an annoying thing to have to do. There is a similar problem in other test areas (e.g. firewall) where changes are made on-the-fly to ``config.php`` - it takes "the system" "some time" to make the change effective.
Yes, it would be good to find a way to ping "the system" to tell it "reload/implement config.php" and get a synchronous response after ``config.php`` is properly reloaded.

In local testing, ``sleep(1)`` was too short (>50% failures in later steps), ``sleep(2)`` reduced failures to maybe 10%, ``sleep(3)`` seemed to be good. So I have gone for ``sleep(5)`` here, to be "sure".

tests/acceptance/features/bootstrap/Auth.php
```diff
@@ -200,4 +253,59 @@ public function aNewBrowserSessionForHasBeenStarted($user) {
...
+	public function deleteTokenAuthEnforcedAfterScenario() {
+		if ($this->tokenAuthHasBeenSet) {
+			$this->runOcc(
+				[
+					'config:system:delete',
+					'token_auth_enforced'
+				]
+			);
+			$this->tokenAuthHasBeenSet = false;
+			sleep(5);
```
phil-davis: There is the same reason to sleep here. We want to be sure that ``token_auth_enforced`` has really been removed, because other ``AfterScenario`` steps are about to try and delete user/groups and put back file sharing settings... They need to be able to freely access the API to succeed.

We could take the approach that we get ``admin`` at token at the start of the scenario, and then used some token-based auth for ``admin`` to do the setup and teardown that it needs. That is more work, but I guess it could be done.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1344,6 +1344,7 @@ public function checkUserAttributes($body) {
 	 * @return void
 	 */
 	public function cleanupUsers() {
+		$this->deleteTokenAuthEnforcedAfterScenario();
```
phil-davis: We need to make sure that ``deleteTokenAuthEnforcedAfterScenario`` happens first of these type of ``AfterScenario`` methods (if it has anything to do). Because ``cleanupUsers`` ``cleanupGroups`` etc must only try to do their thing after ``token_auth_enforced`` has been removed.

Behat has no way to prioritize ``AfterScenario`` methods, so we have to make sure here.

phil-davis:
This can be tried without the ``sleep(5)`` once the PHP ``opcache.revalidate_freq=0`` setting is the default in drone.

We should also think about how to do this on Travis, so that UI tests running on Travis will not come up against this.

phil-davis:
The changes to the PHP server environment on drone allow this to work without needing ``sleep()`` - good.

We need to see what happens on Travis and make similar changes to the PHP server ini settings.

phil-davis:
Scrub those thoughts - these are API acceptance tests and we do not run those on Travis at all. So there is no need to try and touch Travis PHP dev server ini settings.

individual-it (approved):

individual-it:
need a forward-port

phil-davis:
Forward port to master PR #31269

PR #31156, opened 2018-04-17
config.sample.php user.search_min_length

mmattel:
## Description

Adding the key ``user.search_min_length`` to config.sample.php

## Related Issue

https://github.com/owncloud/documentation/issues/4041

## Motivation and Context

PR https://github.com/owncloud/core/pull/30798 (Configurable minimum characters before autocomplete user searches) created the key but no description in config.sample.php was added.

## Types of changes

## Checklist:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -279,6 +279,16 @@ $CONFIG = array(
 'accounts.enable_medial_search' => true,
 
 /**
+ * Defines the minimum characters entered before a search returns results for 
+ * users or groups in the share autocomplete form. Lower values increase search
+ * time especially for large backends.
+ * Any exact matches to a user or group will be returned, even though less than 
+ * the minimum characters have been entered. The search is case insensitive.
+ * e.g. entering "tom" will always return "Tom" if there is an exact match.
+ */
+'user.search_min_length' => 4,
+
+/**
  * Mail Parameters
  *
  * These configure the email settings for ownCloud notifications and password


config/config.sample.php
```diff
@@ -278,6 +278,16 @@
  */
 'accounts.enable_medial_search' => true,
 
+/**
+ * Defines the minimum characters entered before a search returns results for 
+ * users or groups in the share autocomplete form. Lower values increase search
+ * time especially for large backends.
+ * If the query matches exact patterns even there are less characters entered,
+ * the backend will return these matches, eg "Tom" will return "Tom" if there is 
+ * an exact match even the minimum charactes entered is 4.
```
phil-davis: Any exact matches to a user or group will be returned,
even though less than the minimum characters have been entered.
e.g. entering "tom" will always return "Tom" if there is an exact match.

phil-davis: Suggested text to replace the 3 lines above.

mmattel: So "tom" matches lower("Tom") = case insensitive?
Currently for me exact means case sensitive. Your statement is case insensitive.
What is the correct one? 
The sentence itself is great, but I would like to add the case (in)sensitivness.

PVince81: looks like exact match is case insensitive (and always was, not changed by the recent PR related to this setting)
image

mmattel: perfect, will change

phil-davis: It is case-insensitive matching in general. That is the "ordinary" existing behavior. I used "tom" and "Tom" as an exmaple because I did not want to give the idea that you have to start typing an uppercase letter to get a match on a display name.

Add "case-insensitive" somewhere in the text if you like.

felixboehm:
Backport for 10.0.8 ?
@PVince81

phil-davis (approved):
Looks good now

phil-davis:
I can make a backport in just a few seconds...

phil-davis:
Backport ``stable10`` #31160

PR #31172, opened 2018-04-17
Fix apps_paths config sample typo

PVince81:
I've tested this by copy-pasting and adjusting the paths using the "apps_path" from config.sample.php of 10.0.8 RC2 and it failed due to the missing "s".

Please review @mmattel


phil-davis (approved):
Looks better - the code definitely uses ``apps_paths``

mmattel (approved):
Just double checked with the code, thanks for finding.

PVince81:
stable10: https://github.com/owncloud/core/pull/31173

Issue #31150, opened 2018-04-16
Lost share when transferring files to recipient


PVince81:
### Steps

1. Create two users "user1" and "user2"
1. Login as "user1"
1. Create folder "foo/test"
1. Share "foo" with "user2"
1. Login as "user2"
1. Share the received "foo/test" with link
1. Log out
1. On the CLI: `occ files:transfer-ownership user1 user2`
1. Login as "user2"
1. Find the "test" folder in the "transferred from" folder
1. Check link share

### Expected result
Link share preserved

### Actual result
Link share completely lost, even removed from oc_share table.

### Version
10.0.8 RC1

Discovered while testing https://github.com/owncloud/core/pull/30161

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/13671 (Versions lost when moving file into shared folder as recipient), https://github.com/owncloud/core/issues/22594 (Meta data lost when recipient restores a file), https://github.com/owncloud/core/issues/13307 (Move shared file into shared folder as recipient, moved share info lost), https://github.com/owncloud/core/issues/16225 (Files_texteditor gets frozen when a user tries to open a shared file as recipient which was deleted), and https://github.com/owncloud/core/issues/15468 (Files stop being shared after a recipient restores them to previous versions).

PVince81:
talked with @felixboehm and we're going to revert the original fix as the effect is less critical

then move ticket to 10.0.9 for a proper full fix

PVince81:
The revert doesn't solve the issue.

@sharidas will try to find a solution. I expect it to be non-intrusive and should not affect QA testing efforts. If a solution is found in time we can consider putting it in RC3 if a RC3 is deemed necessary.

@pmaier1 @felixboehm FYI

PVince81:
Investigated with @sharidas: the public link has a "parent" field pointing to the regular share.
When we delete the regular share, the link share is deleted as well as the parent is gone.
The reason we delete the regular share from "user1" to "user2" is because "user2" becomes the owner, so the share in question isn't necessary any more.

@sharidas is investigating if there is a good way to deparent the reshare so it can continue to exist, with a solution within the boundaries of the current sharing API.

PVince81:
to be rescheduled

PVince81:
I misunderstood, this was close to completion and just got finished. Closing

sharidas:
[...]

PR #31176, opened 2018-04-17
[stable10] Update public link share in transfer ownership command

sharidas:
The public link share wasn't updated in the command.
This resulted in failure, when the public links were
accessed after the files were transferred. This change
helps to update the share for public links. And hence
access to the links won't cause any failure.

## Description

The public links were failing after transfer ownership command is executed.

## Related Issue

https://github.com/owncloud/core/issues/31150

## Motivation and Context

This change allows the public links to work even after the transfer ownership command is executed.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -299,35 +299,7 @@ class TransferOwnership extends Command {
 
 		foreach($this->shares as $share) {
 			try {
-				if ($share->getSharedWith() === $this->destinationUser) {
-					// Unmount the shares before deleting, so we don't try to get the storage later on.
-					$shareMountPoint = $this->mountManager->find('/' . $this->destinationUser . '/files' . $share->getTarget());
-					if ($shareMountPoint) {
-						$this->mountManager->removeMount($shareMountPoint->getMountPoint());
-					}
-					$this->shareManager->deleteShare($share);
-				} else {
-					if ($share->getShareOwner() === $this->sourceUser) {
-						$share->setShareOwner($this->destinationUser);
-					}
-					if ($share->getSharedBy() === $this->sourceUser) {
-						$share->setSharedBy($this->destinationUser);
-					}
-					/*
-					 * If the share is already moved then updateShare would cause exception
-					 * This can happen if the folder is shared and file(s) inside the folder
-					 * has shares, for example public link
-					 */
-					if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
-						$sharePath = ltrim($share->getNode()->getPath(), '/');
-						if (strpos($sharePath, $this->finalTarget) !== false) {
-							//The share is already moved
-							continue;
-						}
-					}
-
-					$this->shareManager->updateShare($share);
-				}
+				$this->shareManager->transferShare($share, $this->sourceUser, $this->destinationUser, $this->finalTarget);
 			} catch (\OCP\Files\NotFoundException $e) {
 				$output->writeln('<error>Share with id ' . $share->getId() . ' points at deleted file, skipping</error>');
 			} catch (\Exception $e) {
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -28,6 +28,7 @@ namespace OC\Share20;
 
 use OC\Cache\CappedMemoryCache;
 use OC\Files\Mount\MoveableMount;
+use OC\Files\View;
 use OCP\Files\File;
 use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
@@ -42,8 +43,10 @@ use OCP\Security\IHasher;
 use OCP\Security\ISecureRandom;
 use OCP\Share\Exceptions\GenericShareException;
 use OCP\Share\Exceptions\ShareNotFound;
+use OCP\Share\Exceptions\TransferSharesException;
 use OCP\Share\IManager;
 use OCP\Share\IProviderFactory;
+use OCP\Share\IShare;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
@@ -76,6 +79,8 @@ class Manager implements IManager {
 	private $sharingDisabledForUsersCache;
 	/** @var EventDispatcher  */
 	private $eventDispatcher;
+	/** @var  View */
+	private $view;
 
 
 	/**
@@ -103,7 +108,8 @@ class Manager implements IManager {
 			IProviderFactory $factory,
 			IUserManager $userManager,
 			IRootFolder $rootFolder,
-			EventDispatcher $eventDispatcher
+			EventDispatcher $eventDispatcher,
+			View $view
 	) {
 		$this->logger = $logger;
 		$this->config = $config;
@@ -117,6 +123,7 @@ class Manager implements IManager {
 		$this->rootFolder = $rootFolder;
 		$this->sharingDisabledForUsersCache = new CappedMemoryCache();
 		$this->eventDispatcher = $eventDispatcher;
+		$this->view = $view;
 	}
 
 	/**
@@ -673,6 +680,93 @@ class Manager implements IManager {
 		$this->eventDispatcher->dispatch('share.afterCreate', $afterEvent);
 
 		return $share;
+	}
+
+	/**
+	 * Transfer shares from oldOwner to newOwner. Both old and new owners are uid
+	 *
+	 * finalTarget is of the form "user1/files/transferred from admin on 20180509"
+	 *
+	 * TransferShareException would be thrown when:
+	 *  - oldOwner, newOwner does not exist.
+	 *  - oldOwner and newOwner are same
+	 * NotFoundException would be thrown when finalTarget does not exist in the file
+	 * system
+	 *
+	 * @param IShare $share
+	 * @param string $oldOwner
+	 * @param string $newOwner
+	 * @param string $finalTarget
+	 * @throws TransferSharesException
+	 * @throws NotFoundException
+	 */
+	public function transferShare(IShare $share, $oldOwner, $newOwner, $finalTarget) {
+		if ($this->userManager->get($oldOwner) === null) {
+			throw new TransferSharesException("The current owner of the share $oldOwner doesn't exist");
+		}
+		if ($this->userManager->get($newOwner) === null) {
+			throw new TransferSharesException("The future owner $newOwner, where the share has to be moved doesn't exist");
+		}
+
+		if ($oldOwner === $newOwner) {
+			throw new TransferSharesException("The current owner of the share and the future owner of the share are same");
+		}
+
+		//If the destination location, i.e finalTarget is not present, then
+		//throw an exception
+		if (!$this->view->file_exists($finalTarget)) {
+			throw new NotFoundException("The target location $finalTarget doesn't exist");
+		}
+
+		/**
+		 * If the share was already shared with new owner, then we can delete it
+		 */
+		if ($share->getSharedWith() === $newOwner) {
+			// Unmount the shares before deleting, so we don't try to get the storage later on.
+			$shareMountPoint = $this->mountManager->find('/' . $newOwner . '/files' . $share->getTarget());
+			if ($shareMountPoint) {
+				$this->mountManager->removeMount($shareMountPoint->getMountPoint());
+			}
+			$this->deleteShare($share);
+		} else {
+			$sharedWith = $share->getSharedWith();
+
+			$targetFile = '/' . rtrim(basename($finalTarget), '/') . '/' . ltrim(basename($share->getTarget()), '/');
+			/**
+			 * Scenario where share is made by old owner to a user different
+			 * from new owner
+			 */
+			if (($sharedWith !== null) && ($sharedWith !== $oldOwner) && ($sharedWith !== $newOwner)) {
+				$sharedBy = $share->getSharedBy();
+				$sharedOwner = $share->getShareOwner();
+				//The origin of the share now has to be the destination user.
+				if ($sharedBy === $oldOwner) {
+					$share->setSharedBy($newOwner);
+				}
+				if ($sharedOwner === $oldOwner) {
+					$share->setShareOwner($newOwner);
+				}
+				if (($sharedBy === $oldOwner) || ($sharedOwner === $oldOwner)) {
+					$share->setTarget($targetFile);
+				}
+			} else {
+				if ($share->getShareOwner() === $oldOwner) {
+					$share->setShareOwner($newOwner);
+				}
+				if ($share->getSharedBy() === $oldOwner) {
+					$share->setSharedBy($newOwner);
+				}
+			}
+
+			/**
+			 * Here we update the target when the share is link
+			 */
+			if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+				$share->setTarget($targetFile);
+			}
+
+			$this->updateShare($share);
+		}
 	}
 
 	/**
--- a/lib/public/Share/IManager.php
+++ b/lib/public/Share/IManager.php
@@ -21,9 +21,12 @@
 
 namespace OCP\Share;
 
+use OC\User\NoUserException;
 use OCP\Files\Node;
 
+use OCP\Files\NotFoundException;
 use OCP\Share\Exceptions\ShareNotFound;
+use OCP\Share\Exceptions\TransferSharesException;
 
 /**
  * Interface IManager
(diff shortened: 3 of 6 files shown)


sharidas:
The testing is under progress:

- create 2 users `user1` and `admin`.

Scenario 1)
- Login as `admin` and create `foo`
- share `foo` to user `user1`
- Login as `user1` and create public link of `foo` received from `admin`
- Run `./occ files:transfer-ownership admin user1`
- [x] No exception should be there
- [x] The public link created in `user1` for `foo` should be accessible.

Scenario 2)
- Login as `admin` and create `foo/test/hello.txt`. Add some data to `hello.txt`.
- share `foo` to user `user1`
- Login as `user1` and create public link of `foo` received from `admin`
- Run `./occ files:transfer-ownership admin user1`
- [x] No exception should be there
- [x] The public link created in `user1` for `foo` should be accessible. And `hello.txt` data should be the same.

Scenario 3)
- Login as `admin` and create `foo/test/hello.txt`. Add some data to `hello.txt`.
- share `foo` to user `user1`
- Login as `user1` and create public link of `foo` received from `admin`
- Create folder `bar/world.txt`. Add some data and text to `world.txt`.
- Share `bar` to `admin` user
- Create a public link of `bar` from `user1`.
- [x] No exception should be there
- [x] The public link created in `user1` for `foo` should be accessible. And `hello.txt` data should be the same.
- [x] The public link created in `user1` for `bar` should be accessible. And `bar.txt` data should be the same.

sharidas:
The PR has been updated. And with the latest change, https://github.com/owncloud/core/pull/31176#issuecomment-382092262 issue is sovled.

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -307,11 +307,21 @@ private function restoreShares(OutputInterface $output) {
...
-					}
-					if ($share->getSharedBy() === $this->sourceUser) {
-						$share->setSharedBy($this->destinationUser);
+					$sharedWith = $share->getSharedWith();
+					/**
+					 * Scenario where share is made by source user to a user different
+					 * from destination user
+					 */
+					if (($sharedWith !== null) && ($sharedWith !== $this->sourceUser) && ($sharedWith !== $this->destinationUser)) {
+						$this->shareManager->deleteShare($share);
```
PVince81: this feels wrong. if "sourceUser" shared a folder "test/foo" with a "thirdUser" and transfer ownership to "destinationUser", the expected result is that the folder "transferred XYZ/test/foo" of "destinationUser" is still shared with "thirdUser". This is why the below code is changing the share owner

PVince81: the only case where a local share (not public link) would be deleted is if the result of converting the share would result of having "shareOwner === sharedWith" or "sharedBy === sharedWith" (not sure about the latter)

sharidas: Updated the PR. And testing it again.

PVince81: Please update the acceptance tests to make sure the error I mentioned above is caught, the test must detect that shares are still there.

sharidas:
The test taken for reference: https://github.com/owncloud/core/issues/31150#issue-314806286

The reason why this commit : https://github.com/owncloud/core/commit/612fd8a750ffcd64434ee4db042290a3cdc6697c
is breaking the public links is because of the line -> https://github.com/owncloud/core/commit/612fd8a750ffcd64434ee4db042290a3cdc6697c#diff-0674cc8bda871e32d63705af16fa8144R297 ( the share is deleted )

After the delete happens we do nothing with the share because of continue statement -> https://github.com/owncloud/core/commit/612fd8a750ffcd64434ee4db042290a3cdc6697c#diff-0674cc8bda871e32d63705af16fa8144R314

PVince81:
The part I understand is why a public link share can enter the block with `if ($share->getSharedWith() === $this->destinationUser`)` and be deleted there because a public link share usually doesn't have a "shared_with" value.

Can you check with the debugger what its value and type is ?

sharidas:
- [] The test need to be updated properly

sharidas:
Note: By design the federated shares received by the source user in the transfer ownership command doesn't get transferred. Where as the federated shares created by source users are transferred. Because that entry is available in `oc_shares`.

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+		}
+		if ($this->userManager->get($newOwner) === null) {
+			throw new \Exception('The owner ' . $newOwner . ' doesn\'t exist');
+		}
+
+		if ($oldOwner === $newOwner) {
+			throw new \Exception('No point in transferring to same user');
+		}
+
+		$view = new View('/');
```
jvillafanez: Any chance to inject this view? This is affecting the unittests

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+		if ($share->getSharedWith() === $newOwner) {
+			// Unmount the shares before deleting, so we don't try to get the storage later on.
+			$shareMountPoint = $this->mountManager->find('/' . $newOwner . '/files' . $share->getTarget());
+			if ($shareMountPoint) {
+				$this->mountManager->removeMount($shareMountPoint->getMountPoint());
+			}
+			$this->deleteShare($share);
+		} else {
+			$sharedWith = $share->getSharedWith();
+			if (!isset($extraArgs['finalTarget'])) {
```
jvillafanez: This feels checked already above.

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+				throw new \Exception('finalTarget is not set');
+			}
+
+			$targetFile = '/' . rtrim(basename($extraArgs['finalTarget']), '/') . '/' . ltrim(basename($share->getTarget()), '/');
```
jvillafanez: I'm not sure if this works properly. There are some considerations to take into account:
* What if the target for the share isn't inside the root, like `/folder1/file.txt`?
* What if the user has moved the share around?
* What if the `share_folder` option isn't the default one?

What are the expectations in those cases?

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+				$sharedBy = $share->getSharedBy();
+				$sharedOwner = $share->getShareOwner();
+				//The origin of the share now has to be the destination user.
+				if (($sharedBy === $oldOwner) && ($sharedOwner === $oldOwner)) {
+					$share->setSharedBy($newOwner);
+					$share->setShareOwner($newOwner);
+					$share->setTarget($targetFile);
+				}
+			} else {
+				$oldShare = $share;
```
jvillafanez: `$oldShare` seems unused

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+			if (!isset($extraArgs['finalTarget'])) {
+				throw new \Exception('finalTarget is not set');
+			}
+
+			$targetFile = '/' . rtrim(basename($extraArgs['finalTarget']), '/') . '/' . ltrim(basename($share->getTarget()), '/');
+			/**
+			 * Scenario where share is made by old owner to a user different
+			 * from new owner
+			 */
+			if (($sharedWith !== null) && ($sharedWith !== $oldOwner) && ($sharedWith !== $newOwner)) {
```
jvillafanez: Any chance to simplify this code block? It's a bit difficult to follow.
What happens is the "sharedBy" or "sharedOwner", just one of them not both, are the `$oldOwner`?

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+				}
+			}
+			/*
+			 * If the share is already moved then updateShare would cause exception
+			 * This can happen if the folder is shared and file(s) inside the folder
+			 * has shares, for example public link
+			 */
+			if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+				$sharePath = ltrim($share->getNode()->getPath(), '/');
+				if (strpos($sharePath, $extraArgs['finalTarget']) !== false) {
```
jvillafanez: This needs clarification: what if the `$sharePath` is `a/b/c` and the final target is `c`? I guess it should throw an exception in this case

lib/private/Share20/Manager.php
```diff
@@ -675,6 +677,104 @@ public function createShare(\OCP\Share\IShare $share) {
...
+					/**
+					 * Create a new share and update the share with the
+					 * details from old share.
+					 */
+					$token = $share->getToken();
+					$share->setTarget($targetFile);
+					if (($share->getSharedBy() === $share->getShareOwner()) && ($share->getSharedWith() === null)) {
```
jvillafanez: Any reason for these checks? I guess it's to cleanup the shares and ignore some of them that might not be right, but I wonder if this is the place to do it. It would be weird if some shares disappear due to this.

jvillafanez:
A couple of additional things:
* If you're transfering only one share, rename the method to `transferShare`.
* It seems what you're testing is that your code follows an specific path. Ideally the tests should focus on input and output, leaving the implementation details out. The test seems to do the opposite.

lib/private/Share20/Manager.php
sharidas: `$oldShare` is used here.

lib/public/Share/IManager.php
```diff
@@ -112,6 +114,19 @@ public function getAllSharesBy($userId, $shareTypes, $nodeIDs, $reshares = false
...
 	public function getSharesBy($userId, $shareType, $path = null, $reshares = false, $limit = 50, $offset = 0);
 
+	/**
+	 * Transfer shares from oldOwner to newOwner. Both old and new owners are uid
+	 *
+	 * @param IShare $share
+	 * @param string $oldOwner - is the previous owner of the share, the uid string
+	 * @param string $newOwner - is the new owner of the share, the uid string
+	 * @param string $finalTarget - is the target folder where share has to be moved
+	 * @throws NoUserException
```
jvillafanez: It makes sense to throw this exception, but it should throw exceptions from the public namespace. Other apps might catch this exception, so private exceptions shouldn't be thrown.

Not sure what is the best:
* Create a `NoUserException` somewhere in the public namespace. The problem here is that the user management should be the one throwing it, and it might need quite a change. 
* Create a custom `TransferSharesException` to wrap all the exceptions that could be thrown here. Inherited from `GenericShareException` (I'm not sure if this is the base exception for sharing, but it seems so). This also avoid having to throw plain exceptions. I'm not sure if we'd need to refine the exceptions more than this.

In case of doubt, better check with @DeepDiver1975

lib/private/Share20/Manager.php
sharidas: I guess this means a folder/file shared to `$oldOwner`, this remains with `$oldOwner`. That is incoming shares are not moved. Basically I have tried to move the changes from transfer ownership command to this method.

sharidas: For testing I created 4 users: `admin`, `user1`, `user2` and `user3`.
- login as `admin` and create a folder `test_share` and share to `user3` and the table would look like this 
original
- Now modify the table shown above as 
modified
- Now run the transfer ownership command : ``` owncloud3 git:(fix-public-link-transferownership) ✗ ./occ files:transfer-ownership admin user1
Analysing files of admin ...
    1 [============================]
Collecting all share information for files and folder of admin ...
    1 [============================]
Transferring files to user1/files/transferred from admin on 20180508_064055 ...
Restoring shares ...
 1/1 [============================] 100%
➜  owncloud3 git:(fix-public-link-transferownership) ✗ ```
- The test_share is transferred to `user1` and `user3` can still see the share and the table becomes 
afterrunningcommand

lib/private/Share20/Manager.php
sharidas: The idea here was to grab the public links created from `$oldOwner` to be preserved when transferred to newOwner.

lib/public/Share/IManager.php
sharidas: I took the second option for `TransferSharesException`.

tests/lib/Share20/ManagerTest.php
```diff
@@ -1675,6 +1682,146 @@ public function testCanShare($expected, $sharingEnabled, $disabledForUser) {
...
+			]);
+
+		$storage = $this->createMock('\OCP\Files\Storage');
+		$path = $this->createMock('\OCP\Files\File');
+		$path->method('getOwner')->willReturn($shareOwner);
+		$path->method('getName')->willReturn($finalTarget);
+		$path->method('getStorage')->willReturn($storage);
+
+		$share = $this->manager->newShare();
+		/*$share = $this->createShare(
```
jvillafanez: Commented code block

jvillafanez:
Adjust the PHPDoc to the new code, both in the interface and in the implementation

sharidas:
- Adjusted the PHP doc to the new code

tests/lib/Share20/ManagerTest.php
```diff
@@ -85,6 +89,7 @@ public function setUp() {
 		$this->userManager = $this->createMock('\OCP\IUserManager');
 		$this->rootFolder = $this->createMock('\OCP\Files\IRootFolder');
 		$this->eventDispatcher = new EventDispatcher();
+		$this->view = new View('/');
```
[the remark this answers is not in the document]
sharidas: Mocked the View.

tests/lib/Share20/ManagerTest.php
sharidas: Rewrote the test and removed the commented code..

sharidas:
Reply for https://github.com/owncloud/core/pull/31176#discussion_r186388820

Tried to simply the code and remove the usage of `$oldShare`. Now the removal and creation of share happens in the very close lines : https://github.com/owncloud/core/pull/31176/files#diff-a9d8fe189a37b652ea5b821e4d772438R770 and the next line.

This should eliminate the risk appeared in the previous versions of the PR with `$oldShare`.

lib/private/Share20/Manager.php
```diff
@@ -675,6 +682,110 @@ public function createShare(\OCP\Share\IShare $share) {
...
+				}
+			}
+			/*
+			 * If the share is already moved then updateShare would cause exception
+			 * This can happen if the folder is shared and file(s) inside the folder
+			 * has shares, for example public link
+			 */
+			if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+				$sharePath = ltrim($share->getNode()->getPath(), '/');
+				if (strpos($sharePath, $finalTarget) !== false) {
```
jvillafanez: This feels quite a risky condition: basically, it will check if `$sharePath` contains `$finalTarget`, so if `$sharePath = 'path/to/file'` and `$finalTarget = file`, then the test will pass. Do we want this?
Note that you're removing the leading '/' in the `$sharePath` variable but there isn't any check nor documentation about what is expected about the `$finalTarget`. Paths could be "/final", "final", "/path/to/file"... It isn't clear how this code should behave in these cases.

jvillafanez:
It would be nice to have a comment in the "delete + create share" piece with some explanation. Right now it feel a bit weird that you're setting properties in a share, delete it and create it again. I guess the deletion is done via the share id, so the rest of the share properties are ignored (a bad approach as it's confusing). As said, a comment clarifying this situation is enough.

lib/private/Share20/Manager.php
```diff
@@ -675,6 +682,98 @@ public function createShare(\OCP\Share\IShare $share) {
...
+					$share->setSharedBy($newOwner);
+				}
+			}
+			/*
+			 * If the share is already moved then updateShare would cause exception
+			 * This can happen if the folder is shared and file(s) inside the folder
+			 * has shares, for example public link
+			 */
+			if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
+				$token = $share->getToken();
```
jvillafanez: any purpose for this get + set token? If this is needed it will require a comment.

sharidas: Removed the get + set token. Missed to remove them. Since we no more delete and create shares, we no more require this step. @jvillafanez  thanks agian.

PVince81:
to be rescheduled

PVince81:
@sharidas was this close to completion or are there open tasks ? I see unticked checkboxes

PVince81:
@jvillafanez can you re-review to confirm all is fine after your requested changes ?

lib/private/Share20/Manager.php
```diff
@@ -675,6 +682,95 @@ public function createShare(\OCP\Share\IShare $share) {
 		return $share;
 	}
 
+	/**
+	 * Transfer shares from oldOwner to newOwner. Both old and new owners are uid
+	 * The extraArgs array holds the information of the folder where the share
```
jvillafanez: No "extraArgs" argument available any longer

jvillafanez (approved):
Just a minor change. Code looks fine.

PR #31182, opened 2018-04-18
[stable10] Use proper booleans in apps_paths in config.sample.php

PVince81:
I tried installing apps with `occ market:install` but it complained because I had `writable => "true"` instead of `writable => true`


mmattel (approved):

mmattel:
(I'm embarrassed)

PVince81:
master: https://github.com/owncloud/core/pull/31186

PR #31183, opened 2018-04-18
[stable10] Fix preview generation in files api controller and adding unit tests

DeepDiver1975:
## Description
Preview generation for android and ios was broken after #31104 

## Related Issue
#29914 

## How Has This Been Tested?
- manual tests with andoird 2.7.0
- unit test added 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #31184, opened 2018-04-18
Fix preview generation in files api controller and adding unit tests

DeepDiver1975:
front port of #31183

### DIFF
--- a/lib/private/PreviewManager.php
+++ b/lib/private/PreviewManager.php
@@ -139,7 +139,7 @@ class PreviewManager implements IPreview {
 		if ($user === null) {
 			throw new NotLoggedInException();
 		}
-		$file = $this->rootFolder->getUserFolder($user->getUID())->get($file);
+		$file = $this->rootFolder->getUserFolder($user->getUID())->getParent()->get($file);
 		$preview = new Preview('', '/', $file, $maxX, $maxY, $scaleUp);
 		return $preview->getPreview();
 	}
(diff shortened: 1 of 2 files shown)


lib/private/PreviewManager.php
```diff
@@ -139,7 +139,7 @@ public function createPreview($file, $maxX = 100, $maxY = 75, $scaleUp = false)
 		if ($user === null) {
 			throw new NotLoggedInException();
 		}
-		$file = $this->rootFolder->getUserFolder($user->getUID())->get($file);
+		$file = $this->rootFolder->getUserFolder($user->getUID())->getParent()->get($file);
```
PVince81: as the change is different than the one on stable10, is there an explanation for why this is the right approach ?

DeepDiver1975: the api change on master requires a node to be passed into the preview class.
to get the node object for the given path the lazyroot is used as well as getUserFolder.

because the give file path already contains the folder 'files' the get parent call is necessary.

due to missing tests this all way hidden and never popped up - test got added and now if works.

PVince81 (approved):
Code makes sense, not tested :+1:

phil-davis:
Reproduced the thumbnail problem on current master.
Switched to this branch - thumbnails working.
:+1:

phil-davis:
Arghhh! I pressed "close and comment" instead of just "comment". Now when I reopen it seems to be running CI again.

Issue #31188, opened 2018-04-18
10.0.8RC2 / RC3 - Missing important folder \"\" in home storage


patrickjahns:
10.0.8 shows the above error message for non-logged in users 

### Steps to reproduce
1. Install owncloud 10.0.8RC3
2. Use latest `user_ldap` app to import accounts - do not login with the accounts
3. Wait for cron to run

### Expected behaviour
No error messages shown for existing but not yet logged in users

### Actual behaviour
Log is filled with similar error messages below
```
{"reqId":"CI97Vk0uT6otGExEfuZ2","level":3,"time":"2018-04-18T14:23:59+00:00","remoteAddr":"172.42.16.179","user":"--","app":"OC\\Files\\Cache\\Scanner","method":"GET","url":"\/cron.php","message":"Missing important folder \"\" in home storage!!! - home::Test01"}
```

### Server configuration
owncloud-base docker container with mounted 10.0.8RC3

#### ownCloud log (data/owncloud.log)
```
{"reqId":"CI97Vk0uT6otGExEfuZ2","level":3,"time":"2018-04-18T14:23:59+00:00","remoteAddr":"172.42.16.179","user":"--","app":"OC\\Files\\Cache\\Scanner","method":"GET","url":"\/cron.php","message":"Missing important folder \"\" in home storage!!! - home::Test01"}
{"reqId":"CI97Vk0uT6otGExEfuZ2","level":3,"time":"2018-04-18T14:23:59+00:00","remoteAddr":"172.42.16.179","user":"--","app":"OC\\Files\\Cache\\Scanner","method":"GET","url":"\/cron.php","message":"Missing important folder \"\" in home storage!!! - home::Test02"}
{"reqId":"CI97Vk0uT6otGExEfuZ2","level":3,"time":"2018-04-18T14:24:00+00:00","remoteAddr":"172.42.16.179","user":"--","app":"OC\\Files\\Cache\\Scanner","method":"GET","url":"\/cron.php","message":"Missing important folder \"\" in home storage!!! - home::Test03"}
{"reqId":"CI97Vk0uT6otGExEfuZ2","level":3,"time":"2018-04-18T14:24:00+00:00","remoteAddr":"172.42.16.179","user":"--","app":"OC\\Files\\Cache\\Scanner","method":"GET","url":"\/cron.php","message":"Missing important folder \"\" in home storage!!! - home::Test04"}
[...]

PR #31189, opened 2018-04-18
[stable10] Prevent background scan to scan non-ready homes

PVince81:
## Description
Whenever a user has never logged in, their home is not setup yet.
In this case the background scan should not go there.

It turns out that the check already existed for regular scan but not background scan.
This PR refactors some of the storage checks into a common method to avoid discrepancies in the future.

## Related Issue
Fixes https://github.com/owncloud/core/issues/31188

## How Has This Been Tested?
See https://github.com/owncloud/core/issues/31188#issuecomment-382441610

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Utils/Scanner.php
+++ b/lib/private/Files/Utils/Scanner.php
@@ -136,25 +136,17 @@ class Scanner extends PublicEmitter {
 	public function backgroundScan($dir) {
 		$mounts = $this->getMounts($dir);
 		foreach ($mounts as $mount) {
-			$storage = $mount->getStorage();
-			if (is_null($storage)) {
+			if (!$this->shouldScan($mount)) {
 				continue;
 			}
 
-			// don't bother scanning failed storages (shortcut for same result)
-			if ($storage->instanceOfStorage('OC\Files\Storage\FailedStorage')) {
-				continue;
-			}
+			$storage = $mount->getStorage();
 
 			// don't scan the root storage
 			if ($storage->instanceOfStorage('\OC\Files\Storage\Local') && $mount->getMountPoint() === '/') {
 				continue;
 			}
 
-			// don't scan received local shares, these can be scanned when scanning the owner's storage
-			if ($storage->instanceOfStorage('OCA\Files_Sharing\ISharedStorage')) {
-				continue;
-			}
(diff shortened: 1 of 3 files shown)


jvillafanez:
Code looks good, but I'm not sure if we should also force the background job to scan only the folders for logged in / seen users, however, user pagination doesn't seem to be supported for this case...

jvillafanez (approved):
Logs are gone, although the storages are still being created in the storage table. Not sure if it will be problematic. Good enough for a quick fix IMO

mmattel (approved):

PVince81:
Thanks for the reviews.

We are unsure yet whether to put this in.

I need to fix those tests...

mmattel:
We did not had an issue in a production system creating a user and the user did not log on except the error logged as discussed. This PR eliminates the error, imho in a good way. To be said, there is always room for improvement 😄 Proposing to get it in 10.0.8.

PVince81:
the unit tests fail because the unit test is expecting the test folder with path "/" to work for regular scan.

in the PR I moved the checks into a common function and as a side effect also causing the regular scan to skip the "Local" storage. As I want this change to have the least risk, I'll revert the logic to be as before so "Local" storage with "/" is only skipped for bkg scan, not regular scan.

PVince81:
pushed. please re-review

lib/private/Files/Utils/Scanner.php
```diff
@@ -175,6 +167,43 @@ public function backgroundScan($dir) {
 		}
 	}
 
+	/**
+	 * Returns whether the given mount point should be scanned
+	 *
+	 * @param $mount mount point
+	 * @return bool true to scan, false to skip
+	 */
+	private function shouldScan($mount) {
```
jvillafanez: If `$mount` isn't used it's better to ask for the storage instead. This will prevent calling twice the method `$mount->getStorage()`

jvillafanez:
Code looks good, just a minor tweek 👍 

There was an extra check before that has been removed in the `shouldScan` call. I don't know if it was intended or no longer needed.

PVince81:
master: https://github.com/owncloud/core/pull/31202

Issue #31194, opened 2018-04-19
Federated share http fallback switch not working


PVince81:
See https://github.com/owncloud/core/pull/30198

### Steps
1. Disable "http" fallback
TODO

### Expected
No http fallback

### Actual
http fallback seems to be in place

### Version
ownCloud 10.0.8 RC3

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/19477 (Federated Cloud Sharing without Resharing), https://github.com/owncloud/core/issues/17551 (Sharing from server A daily (https) to server B (http) federated sharing does not work), https://github.com/owncloud/core/issues/20216 (Use HTTPS for federated cloud share when another share already uses HTTPS), https://github.com/owncloud/core/issues/20329 (Federated shared subfolder not syncing.), and https://github.com/owncloud/core/issues/22721 (Federated shares accepted and not displayed).

PR #31196, opened 2018-04-19
[stable10] Respect sharing.federation.allowHttpFallback in FederatedShareProvider

DeepDiver1975:
## Description
By default federated sharing should use https only - in selected cases a fallback to http is allowed.

## Related Issue
refs https://github.com/owncloud/core/issues/31194

## How Has This Been Tested?
1. have an owncloud instance available on http only
2. create federated share via https
3. falback to http is only allowed if the config switch is set to true

## Types of changes

## Checklist:

### DIFF
--- a/apps/federatedfilesharing/lib/BackgroundJob/RetryJob.php
+++ b/apps/federatedfilesharing/lib/BackgroundJob/RetryJob.php
@@ -74,7 +74,8 @@ class RetryJob extends Job {
 				$addressHandler,
 				\OC::$server->getHTTPClientService(),
 				$discoveryManager,
-				\OC::$server->getJobList()
+				\OC::$server->getJobList(),
+				\OC::$server->getConfig()
 			);
 		}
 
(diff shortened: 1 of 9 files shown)


PVince81 (approved):
:+1: code looks good

Mind adding test steps ? This likely requires some SSL web server ?

patrickjahns:
I always see when trying to initiate a federated share to a http server - no matter if the option is set to true/false
image

apps/federatedfilesharing/lib/BackgroundJob/RetryJob.php
```diff
@@ -74,7 +74,8 @@ public function __construct(Notifications $notifications = null) {
 				$addressHandler,
 				\OC::$server->getHTTPClientService(),
 				$discoveryManager,
-				\OC::$server->getJobList()
+				\OC::$server->getJobList(),
+				\OC::$server->getConfig()
```
PVince81: so cron.php needs to be tested as well

PVince81 (approved):
:+1: code change with break makes sense

patrickjahns:
Tested with http only receiving server:

With `'sharing.federation.allowHttpFallback' => false` I see this information
image

With `'sharing.federation.allowHttpFallback' => true` it allows for sharing

patrickjahns (approved):
LGTM - tested locally

PR #31151, opened 2018-04-16
Do not load disabled app-themes

VicDeo:
## Description
Do not load disabled app-themes

## Related Issue
https://github.com/owncloud/core/issues/31134

## Motivation and Context
Inactive app-themes are applied to the page when they are loaded and this leads to broken or wrong page layout if `OC\Theme\ThemeService::getAllThemes` is called

## How Has This Been Tested?
https://github.com/owncloud/core/issues/31134#issue-314567249

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #31203, opened 2018-04-19
Fix legacy theme urls

butonic:
Without this legacy themes always assume `/`as the oc endpoint, causing failed image loads. Workaround: use a proper theme app.

### DIFF
--- a/lib/private/Theme/ThemeService.php
+++ b/lib/private/Theme/ThemeService.php
@@ -103,7 +103,7 @@ class ThemeService implements IThemeService {
 				$webPath = \OC_App::getAppWebPath($themeName);
 			} else {
 				$directory = 'themes/' . $themeName;
-				$webPath = '/themes/' . $themeName;
+				$webPath = \OC::$WEBROOT . "/themes/$themeName";
 			}
 		}
 


PVince81:
this is likely a regression from 10.0.7

PVince81:
@butonic test steps ?

PVince81:
stable10: https://github.com/owncloud/core/pull/31204

we need to save time...

DeepDiver1975:
- no unit tests
- no test steps
- template not used

-> close

PVince81:
raised https://github.com/owncloud/core/issues/31205 to look into this in detail later

lib/private/Theme/ThemeService.php
```diff
@@ -103,7 +103,7 @@ private function makeTheme($themeName, $appTheme = true) {
 				$webPath = \OC_App::getAppWebPath($themeName);
 			} else {
 				$directory = 'themes/' . $themeName;
-				$webPath = '/themes/' . $themeName;
+				$webPath = \OC::$WEBROOT . "/themes/$themeName";
```
VicDeo: $this->serverRoot

VicDeo:
Will be incorporated into https://github.com/owncloud/core/pull/31151

PR #31216, opened 2018-04-20
Fix extractRequestTokenFromResponse method name

phil-davis:
I have noticed this typo in the method name before, it has zero impact on functionality. But now I noticed it again when modifying code. So let's "fix" it so my brain is not impacted by it ever again.

And a few PHP docblock things that ``phpcs`` was complaining about.


phil-davis:
Backport ``stable10`` #31217

individual-it (approved):

PR #31109, opened 2018-04-12
[stable10] do not urlencode group id

butonic:
backport of https://github.com/owncloud/user_management/pull/20

PR #31224, opened 2018-04-20
generated resource routes should allow /

### LINKED ISSUE #19: trying to list users of a groups with a space yields empty list
because the gid us url encoded and ` ` becomes `%20` not `+`

### LINKED ISSUE #31223: Cannot use / in url params
### Steps to reproduce
1. register a route with a param, eg `/resource/{id}`
2. try to use an id with a `/`. it should be urlencoded, so the url in a request will look like `/resource/id%2Fwith%2Fslashes`

### Expected behaviour
The controller should get `id/with/slashes` as a parameter.

### Actual behaviour
The controller is never called because the url never matches.
1. make sure apache does not auto decode slashes by adding `AllowEncodedSlashes On` to the virtualhosts section. Otherwise you [get an immediate 404 without php ever seeing the request](https://github.com/owncloud/user_management/pull/20#issuecomment-383037854)
2. Even with `AllowEncodedSlashes On` the url gets decoded in Symphonys URLMatcher. In the worst case this [might cause calling the wrong route](https://stackoverflow.com/questions/47973132/symfony-3-4-routing-urlencoded-in-the-parameter-value-cause-not-matching-th#comment82932678_47973462).

### Related
https://github.com/owncloud/user_management/pull/20#issuecomment-383037854

### LINKED ISSUE #31279: Cannot see members of a group having special characters(@,/,&) in its name
### Steps to reproduce
1. Create a group with name containing special character (@,/,&) eg. admin&management, admin/intern
2. Assign users to the group
3. See that only the number of members in a group is listes but not the list of members 

### Expected behaviour
Either admin should not be able to create a group with name having special characters in it.
Or if you can create such groups, it should function like a group with normal name

### Actual behaviour
Unable to see the list of members in a group

### Server configuration
**Operating system**: Ubuntu 17.10

**Web server:** Apache

**Database:** MySQL

**PHP version:** 7.1

**ownCloud version:** (see ownCloud admin page) 10.0.8.5

**Where did you install ownCloud from:** git

#### Browser log
```
Request URL: http://localhost/core/index.php/settings/users/users?offset=0&limit=200&gid=people%2523me&pattern=
Request Method: GET
Status Code: 200 OK
```
### Screenshot
image

butonic:
fixes https://github.com/owncloud/core/issues/31223
fixes https://github.com/owncloud/core/issues/31279
fixes https://github.com/owncloud/user_management/issues/19

incudes bonus test written for a different approach before finding https://symfony.com/doc/current/routing/slash_in_parameter.html

### DIFF
--- a/tests/lib/Route/RouterTest.php
+++ b/tests/lib/Route/RouterTest.php
@@ -1,5 +1,6 @@
 <?php
 /**
+ * @author Jörn Friedrich Dreyer <jfd@owncloud.com>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
  *
  * @copyright Copyright (c) 2018, ownCloud GmbH
@@ -24,7 +25,33 @@ namespace Test\Route;
 use OC\Route\Router;
 use OCP\ILogger;
 
+
+class LoadableRouter extends Router {
+	/**
+	 * @param bool $loaded
+	 */
+	public function setLoaded($loaded) {
+		$this->loaded = $loaded;
+	}
+}
+
 class RouterTest extends \Test\TestCase {
+
+
+	/** @var ILogger */
+	private $l;
+
+	/**
+	 * RouterTest constructor.
+	 *
+	 * @param string $name
+	 * @param array $data
+	 * @param string $dataName
+	 */
+	public function __construct($name = null, array $data = [], $dataName = '') {
+		parent::__construct($name, $data, $dataName);
+		$this->l = $this->createMock(ILogger::class);
+	}
 
 	/**
 	 * @dataProvider providesWebRoot
@@ -32,8 +59,7 @@ class RouterTest extends \Test\TestCase {
 	 * @param $webRoot
 	 */
 	public function testWebRootSetup($expectedBase, $webRoot) {
-		$l = $this->createMock(ILogger::class);
-		$router = new Router($l, $webRoot);
+		$router = new Router($this->l, $webRoot);
 
 		$this->assertEquals($expectedBase, $router->getGenerator()->getContext()->getBaseUrl());
 	}
@@ -45,5 +71,24 @@ class RouterTest extends \Test\TestCase {
 			['/oc/index.php', '/oc'],
 			['/oc/index.php', '/oc/'],
 		];
+	}
+
+	public function testMatchURLParamContainingSlash() {
+		$router = new LoadableRouter($this->l, '');
+
+		$called = false;
+
+		$router->useCollection('root');
+		$router->create('test', '/resource/{id}')
+			->action(function() use (&$called) {
+			$called = true;
+		})->requirements(['id' => '.+']);
+
+		// don't load any apps
+		$router->setLoaded(true);
+
+		$router->match('/resource/id%2Fwith%2Fslashes');
+
+		self::assertTrue($called);
 	}
 }
--- a/tests/lib/Route/RouterTest.php
+++ b/tests/lib/Route/RouterTest.php
@@ -26,7 +26,6 @@ use OC\Route\Router;
 use OCP\ILogger;
 use Symfony\Component\Routing\Exception\ResourceNotFoundException;
 
-
 class LoadableRouter extends Router {
 	/**
 	 * @param bool $loaded
@@ -37,7 +36,6 @@ class LoadableRouter extends Router {
 }
 
 class RouterTest extends \Test\TestCase {
-
 
 	/** @var ILogger */
 	private $l;
@@ -77,16 +75,19 @@ class RouterTest extends \Test\TestCase {
 	/**
 	 * @dataProvider urlParamSlashProvider
 	 */
-	public function testMatchURLParamContainingSlash($routeUrl, $matchUrl, $expectedCalled) {
+	public function testMatchURLParamContainingSlash($routeUrl, $slashesAllowed, $matchUrl, $expectedCalled) {
 		$router = new LoadableRouter($this->l, '');
 
 		$called = false;
 
 		$router->useCollection('root');
-		$router->create('test', $routeUrl)
-			->action(function() use (&$called) {
-			$called = true;
-		})->requirements(['id' => '.+']);
+		$route = $router->create('test', $routeUrl)
+			->action(function () use (&$called) {
+				$called = true;
+			});
+		if ($slashesAllowed) {
+			$route->requirements(['id' => '.+']);
+		}
 
 		// don't load any apps
 		$router->setLoaded(true);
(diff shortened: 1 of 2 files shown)


DeepDiver1975:
@ownclouders rebase

ownclouders:

ownclouders:

butonic:
@phil-davis I think this PR supersedes
- https://github.com/owncloud/core/pull/31109 - [stable10] do not urlencode group id,
- https://github.com/owncloud/user_management/pull/20 - do not urlencode group id
- https://github.com/owncloud/core/pull/31017 - Handle groups with slash and percent in provisioning API

and is the proper fix for
- https://github.com/owncloud/core/issues/31223 - Cannot use / in url params
- https://github.com/owncloud/core/issues/31015 - Cannot view or delete groups with a slash in their name using provisioning API 

The tests will work with these changes alone. In the real world we may need [`AllowEncodedSlashes On`](https://httpd.apache.org/docs/2.4/de/mod/core.html#allowencodedslashes) in the apache vhost config (it is not allowed in htaccess but is `Off`by default) This requires
- [ ] a documentation update ... and
- [ ] probably an admin check would make sense

@phil-davis can you test if it this PR really fixes the problem for the provisioning API?

DeepDiver1975:
> #31017 - Handle groups with slash and percent in provisioning API

the provisioning api is not setup as a resource via appframework - ocs has it's own routing ... \o/

phil-davis:
This did not work for me for the Provisioning API "groups with slash in the name" cases.
See PR #31463 for my attempt.

butonic:
superseded by https://github.com/owncloud/core/pull/31463

phil-davis (commented):
With this PR, and current master of user_management, the webUI does not list users in groups such as
```
a space
a/slash
a&and
a@at
a+plus
```
If I also checkout ``user_management`` branch ``bugfix/19``  from https://github.com/owncloud/user_management/pull/20 then the webUI lists users in those groups.

Working requests look like:
```
[Wed May 30 16:15:54 2018] 192.168.0.101:44482 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a+space&pattern=
[Wed May 30 16:15:55 2018] 192.168.0.101:44484 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%26and&pattern=
[Wed May 30 16:15:56 2018] 192.168.0.101:44486 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%2Bplus&pattern=
[Wed May 30 16:15:56 2018] 192.168.0.101:44488 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%2Fslash&pattern=
[Wed May 30 16:15:57 2018] 192.168.0.101:44490 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%40at&pattern=
```
and from ``user_management`` ``master`` the requests are like:
```
[Wed May 30 16:14:54 2018] 192.168.0.101:44268 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%2520space&pattern=
[Wed May 30 16:14:56 2018] 192.168.0.101:44270 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%2526and&pattern=
[Wed May 30 16:14:57 2018] 192.168.0.101:44272 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%252Bplus&pattern=
[Wed May 30 16:14:57 2018] 192.168.0.101:44274 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%252Fslash&pattern=
[Wed May 30 16:14:58 2018] 192.168.0.101:44276 [200]: /index.php/apps/user_management/users?offset=0&limit=200&gid=a%2540at&pattern=
```

phil-davis:
So, what range of stuff is this PR supposed to fix by itself?
And what also needs JavaScript front-end changes?

butonic:
ah, crap. I can no longer restore `user_management` branch `bugfix/19` from owncloud/user_management#20

@phil-davis can you push?

I think both are needed. https://github.com/owncloud/user_management/pull/20 no longer double encodes / on other chars. This PR properly handles '/'

phil-davis:
I pushed ``bugfix/19`` back to ``user_management`` - hopefully it appears, maybe you will need to make a new PR, maybe it will "just work" (tm).

butonic:
@phil-davis can I get a :+1: for this it makes the server side handle / in resource urls properly. Resource urls are used in many places:
```console
grep -A4 -r "'resources'" apps*/*/appinfo/routes.php
apps/files_external/appinfo/routes.php:		'resources' => [
apps/files_external/appinfo/routes.php-			'global_storages' => ['url' => '/globalstorages'],
apps/files_external/appinfo/routes.php-			'user_storages' => ['url' => '/userstorages'],
apps/files_external/appinfo/routes.php-			'user_global_storages' => ['url' => '/userglobalstorages'],
apps/files_external/appinfo/routes.php-		],
--
apps/files_sharing/appinfo/routes.php:	'resources' => [
apps/files_sharing/appinfo/routes.php-		'ExternalShares' => ['url' => '/api/externalShares'],
apps/files_sharing/appinfo/routes.php-	],
apps/files_sharing/appinfo/routes.php-	'routes' => [
apps/files_sharing/appinfo/routes.php-		[
--
apps-repos/user_management/appinfo/routes.php:	'resources' => [
apps-repos/user_management/appinfo/routes.php-		'groups' => ['url' => '/groups'],
apps-repos/user_management/appinfo/routes.php-		'users' => ['url' => '/users'],
apps-repos/user_management/appinfo/routes.php-	],
apps-repos/user_management/appinfo/routes.php-	'routes' => [
```

and likely other apps that I have not checked out, currently.

Let is get this into 10.0.9, I'll backport asap

phil-davis (approved):
Also needs corresponding ``user_management`` change to go with this here in ``core`` ``master``

Backport to ``stable10`` will need to add the code from ``user_management``

tests/lib/Route/RouterTest.php
```diff
@@ -47,4 +73,23 @@ public function providesWebRoot() {
...
 		];
 	}
+
+	public function testMatchURLParamContainingSlash() {
+		$router = new LoadableRouter($this->l, '');
+
+		$called = false;
+
+		$router->useCollection('root');
+		$router->create('test', '/resource/{id}')
```
PVince81: I'm missing a test for a URL like "'/resource/{id}/subentries" where the id contains a slash as well

PVince81:
Taking over

PVince81:
I have a concern: apparently you don't need to encode the slashes and it will match.

What happens if there are route conflicts ? If it happens that the id is controllable by the user, a user could manipulate the route to make it point at another similar route.

Ex:
- route1 is "resource/{id}" (main entity)
- route2 is "resource/{id}/sub" (collection of sub-entities)

1) with id "test" and "resource/test/sub" it should match route2
2) with id "test/sub" and "resource/test/sub" it should match route1 but here would match route2

Is this something we can live with ? @butonic 

We could agree that route designers need to define their routes in a way that such confusion is not possible.

PVince81:
Second point: do we require the id to be fully encoded ?

I wrote unit tests and the matcher also matches if the slashes are not encoded, see https://github.com/owncloud/core/pull/31224/commits/73acdd11c445a89acd93d6ed6cd825d110b657bf#diff-3d3d328fe9bbbd7019bad6ec9ae10dacR109

I'd rather prefer to not match these to avoid the conflict mentioned above.

PVince81:
talked with @tomneedham and he also preferred we not touch the core router due to various risks and favor the other solution https://github.com/owncloud/core/pull/31109

need to examine whether the latter also fixes the same issues like the former

PVince81:
Changed expectation for non-encoded slash-containing params to true: the Symfony matcher automatically calls `rawurldecode` on parameters, so both the encoded and not encoded param will match.

The purpose of RouterTest here is to demonstrate that setting the `requirement` part will make it match params with slashes properly.

PVince81:
:+1: for @butonic's code.

I need a review for the unit test extension I wrote, @butonic ?

butonic:
@PVince81 nice! should we make the regex in https://github.com/owncloud/core/pull/31224/commits/3626e324d72278c31959ece15dd8ee1e1aea6c19#diff-3d3d328fe9bbbd7019bad6ec9ae10dacR90 part of the provider params? that way we can test both: normal routes with `['id' => '[^/]+']` and resource routes with `['id' => '.+']`

PVince81:
added test for when route disallows slashes

butonic:
:+1: for @PVince81 changes ... but cs wants some newlines removed

phil-davis:
@PVince81 @butonic now what is the workflow/timing for backporting this?
I guess we can get both pieces of this backported into ``stable10`` - code from this PR and #31109 which is the ``core`` ``stable10`` backport of the ``user_management`` PR https://github.com/owncloud/user_management/pull/20

And it will wait for after ``10.0.9``

PVince81:
yes, I'll backport this to stable10

cc @tomneedham

PVince81:
stable10: https://github.com/owncloud/core/pull/31927

Issue #23960, opened 2016-04-13
Unnecessary X-Frame-Options warning in admin panel with header set to DENY


eppfel:
### Steps to reproduce
1. Set HTTP header "X-Frame-Options" to "DENY" on your server
2. Open the Admin panel in your browser
### Expected behaviour

There should not be a warning.
### Actual behaviour

A Warning is placed in the Admin panel: The "X-Frame-Options" HTTP header is not configured to equal to "SAMEORIGIN". This is a potential security or privacy risk and we recommend adjusting this setting.
The warning is not incorrect as it states a "potential" risk, but it is annoying after spending 1-2 hours fixing all warnings.
### Server configuration

CentOS
Apache 2.2
MySQL
PHP 5.6
9.0.1 (stable)
fresh install
Web Installer

**Signing status (ownCloud 9.0 and above):**
No errors have been found.

**List of activated apps:**

Enabled:
- activity: 2.2.1
- calendar: 1.0
- comments: 0.2
- contacts: 1.1.0.0
- dav: 0.1.6
- external: 1.2
- federatedfilesharing: 0.1.0
- federation: 0.0.4
- files: 1.4.4
- files_sharing: 0.9.1
- files_texteditor: 2.1
- files_trashbin: 0.8.0
- files_versions: 1.2.0
- files_videoplayer: 0.9.8
- firstrunwizard: 1.1
- gallery: 14.5.0
- notifications: 0.2.3
- provisioning_api: 0.4.1
- systemtags: 0.2
- templateeditor: 0.1
- updatenotification: 0.1.0
  Disabled:
- encryption
- files_external
- files_pdfviewer
- user_external
- user_ldap

**The content of config/config.php:**

``` PHP
{
    "system": {
        "instanceid": "ocztm7fxpnfx",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "epp.cloud"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "overwrite.cli.url": "https:\/\/epp.cloud",
        "dbtype": "mysql",
        "version": "9.0.1.3",
        "dbname": "eppcloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
[...]

Issue #17613, opened 2015-07-13
Security & Setup Warnings for X-Content-Type-Options and X-Frame-Options


4fd81048a9cd8a3fd43f1f4e52f84181:
### Steps to reproduce
1. View the admin page

### Expected behaviour
No warnings

### Actual behaviour
```
The "X-Content-Type-Options" HTTP header is not configured to equal to "nosniff". This is a potential security or privacy risk and we recommend adjusting this setting.
The "X-Frame-Options" HTTP header is not configured to equal to "SAMEORIGIN". This is a potential security or privacy risk and we recommend adjusting this setting.
```

### Server configuration
**Operating system**: Ubuntu 15.04

**Web server:**  Apache/2.4.10 (Ubuntu)

**Database:** postgres (PostgreSQL) 9.4.4

**PHP version:** PHP Version 5.6.4-4ubuntu6.2

**ownCloud version:** ownCloud 8.1.0 (stable)

**Updated from an older ownCloud or fresh install:** Updated

### Details
```
curl -I https://www.buchananworldcorp.com/owncloud/
HTTP/1.1 200 OK
Date: Mon, 13 Jul 2015 13:50:36 GMT
Server: Apache/2.4.10 (Ubuntu)
Strict-Transport-Security: max-age=63072000; includeSubdomains; preload
X-Frame-Options: SAMEORIGIN
X-Content-Type-Options: nosniff
Expires: Thu, 19 Nov 1981 08:52:00 GMT
Cache-Control: no-store, no-cache, must-revalidate, post-check=0, pre-check=0
Pragma: no-cache
Content-Security-Policy: default-src 'self'; script-src 'self' 'unsafe-eval'; style-src 'self' 'unsafe-inline'; frame-src *; img-src *; font-src 'self' data:; media-src *; connect-src *
Set-Cookie: oc9a1e2ea845=e7mao7domb5cokpjki09evlb56; path=/owncloud; HttpOnly; Secure
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
X-Robots-Tag: none
X-Frame-Options: SAMEORIGIN
Content-Type: text/html; charset=UTF-8
```

These headers are being set twice in the response.  In Issue #15342 LukasReschke suggests that this is likely the problem.  (Actually two problems, the headers should not be set twice and the check should handle the header appearing twice)

I found that these headers were getting set in /etc/apache2/mods-enabled/ssl.conf and in .htaccess.

[...]

PR #31231, opened 2018-04-21
Always set headers

benrubson:
Hi,

Headers may already be defined in Apache configuration.
This then leads to header values being defined twice, thus triggering such error message :
`The "X-Content-Type-Options" HTTP header is not configured to equal to "nosniff". This is a potential security or privacy risk and we recommend adjusting this setting.`

This PR then adds the `always` keyword to `Header` definitions, which solves the issue.

Thank you 👍 

Ben

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -15,23 +15,23 @@
 
   <IfModule mod_env.c>
     # Add security and privacy related headers
-    Header set X-Content-Type-Options "nosniff"
-    Header set X-XSS-Protection "1; mode=block"
-    Header set X-Robots-Tag "none"
-    Header set X-Frame-Options "SAMEORIGIN"
-    Header set X-Download-Options "noopen"
-    Header set X-Permitted-Cross-Domain-Policies "none"
+    Header always set X-Content-Type-Options "nosniff"
+    Header always set X-XSS-Protection "1; mode=block"
+    Header always set X-Robots-Tag "none"
+    Header always set X-Frame-Options "SAMEORIGIN"
+    Header always set X-Download-Options "noopen"
+    Header always set X-Permitted-Cross-Domain-Policies "none"
     SetEnv modHeadersAvailable true
   </IfModule>
 
   # Let browsers cache CSS, JS files for half a year
   <FilesMatch "\.(css|js)$">
-    Header set Cache-Control "max-age=15778463"
+    Header always set Cache-Control "max-age=15778463"
   </FilesMatch>
   
   # Let browsers cache WOFF files for a week
   <FilesMatch "\.woff$">
-    Header set Cache-Control "max-age=604800"
+    Header always set Cache-Control "max-age=604800"
   </FilesMatch>
 </IfModule>
 


mmattel:
In case this gets merged, I will add this to the nginx config / documentation too.

mmattel:
Related issues:
https://github.com/owncloud/core/issues/17613 (Security & Setup Warnings for X-Content-Type-Options and X-Frame-Options)
https://forum.owncloud.org/viewtopic.php?t=33325

A good explanation:
https://stackoverflow.com/questions/39502968/apache-difference-between-header-always-set-and-header-set
...
**What is the difference between Header always set and Header set in Apache?**
As the quoted bit from the manual says, without 'always' your additions will only go out on succesful responses.
...

@butonic @DeepDiver1975 @PVince81 
I vote for merging.

butonic:
@benrubson can you clarify which routes / url show this error?

1. oc will set the headers for every response:

https://github.com/owncloud/core/blob/master/lib/private/legacy/response.php#L259-L264

`header()` will by default replace an existing header. so every request should be returned with the right headers.

2. the .htaccess rules for apache will only add these headers if they haven't been added yet, eg for images:

https://github.com/owncloud/core/blob/master/.htaccess#L18-L24

As you can see php will also only set the headers if the `modHeadersAvailable` environment variable hasn't been set.

The code goes a long way not to set these headers twice. Which urls are problematic?

benrubson:
@butonic, error shows in `/settings/admin?sectionid=general`

I agree with you, but in addition to be set by OwnCloud itself, such headers could also already be set by the server administrator, in Apache configuration (i.e. in `httpd.conf`).

benrubson:
Some news regarding this ?
Many thanks again 👍

benrubson:
Some news regarding this please ?
Many thanks again 👍

benrubson:
Still facing this with 10.0.10.
Could we then think about merging this please ?
Many thanks again 👍

mmattel:
@benrubson pls correct me if I am wrong...
This PR is about to fix the issue when not only owncloud sets a header, but the header is additionally set by an apache config (header on header). Can happen when the apache admin mandatory adds headers in http.conf. According the apache documentation headers can be "merged" and the always directive would do so in this case eliminating the apache error message thrown.
It is not that owncloud settig headers does something wrong or misses something, it is about setting headers from oC and apache leads to that error which is fixed by this PR.
More info at: https://httpd.apache.org/docs/current/mod/mod_headers.html#page-header

benrubson:
Yes @mmattel, U're totally right 👍

DeepDiver1975 (approved):

benrubson:
Many thanks @DeepDiver1975 👍

mmattel:
For completeness, the docs have been updated
- with regards to nginx: https://github.com/owncloud/docs/pull/487 (nginx improvements).
- I will check if we need/shoud add a note to apache and refrence the PR in case

mmattel:
@benrubson fyi, the backport for stable10 has been filed but will not make it in the upcoming release due to codefreeze. It will be in the release after.

ho4ho:
Could around working https://github.com/owncloud/core/issues/23960 ?

benrubson:
Yes, this PR and #35099 should solve #23960.

Issue #31264, opened 2018-04-25
Symfony events for user preference changes


PVince81:
We already trigger symfony events before and after system config and app config changes.

We should also do so for user preference changes.

@tomneedham @sharidas

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/pull/30996 (Throw symfony events on group actions), https://github.com/owncloud/core/pull/30054 ([stable10] Bump symfony/event-dispatcher from 3.2.4 to 3.2.14), https://github.com/owncloud/core/pull/30930 (Add symfony events for delete share), https://github.com/owncloud/core/pull/29885 (Adding new Symfony events), and https://github.com/owncloud/core/pull/30570 ([WIP] Adding symfony events for shares).

PVince81:
PR here: https://github.com/owncloud/core/pull/31266

PR #31266, opened 2018-04-25
[stable10] Add events for user preference changes

PVince81:
## Description
Add events for user preference changes

## Related Issue
Fixes https://github.com/owncloud/core/issues/31264

## How Has This Been Tested?
- [x] add unit tests

## Types of changes

## Checklist:

### DIFF
--- a/lib/base.php
+++ b/lib/base.php
@@ -144,7 +144,14 @@ class OC {
 				'SCRIPT_FILENAME' => $_SERVER['SCRIPT_FILENAME'],
 			],
 		];
-		$fakeRequest = new \OC\AppFramework\Http\Request($params, null, new \OC\AllConfig(new \OC\SystemConfig(self::$config)));
+		/**
+		 * The event dispatcher added here will not be used to listen any event.
+		 * So if the modifications made in the configuration, by fakeRequest
+		 * will not throw events.
+		 */
+		$fakeRequest = new \OC\AppFramework\Http\Request($params, null,
+			new \OC\AllConfig(new \OC\SystemConfig(self::$config),
+				new \Symfony\Component\EventDispatcher\EventDispatcher()));
 		$scriptName = $fakeRequest->getScriptName();
 		if (substr($scriptName, -1) == '/') {
 			$scriptName .= 'index.php';
--- a/lib/private/AllConfig.php
+++ b/lib/private/AllConfig.php
@@ -30,18 +30,24 @@
 namespace OC;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use OC\Cache\CappedMemoryCache;
+use OCP\IConfig;
 use OCP\IDBConnection;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class to combine all the configuration options ownCloud offers
  */
-class AllConfig implements \OCP\IConfig {
+class AllConfig implements IConfig {
+
 	/** @var SystemConfig */
 	private $systemConfig;
 
 	/** @var IDBConnection */
 	private $connection;
 
+	/** @var EventDispatcher  */
+	private $eventDispatcher;
 	/**
 	 * 3 dimensional array with the following structure:
 	 * [ $userId =>
@@ -65,11 +71,15 @@ class AllConfig implements \OCP\IConfig {
 	private $userCache;
 
 	/**
+	 * AllConfig constructor.
+	 *
 	 * @param SystemConfig $systemConfig
+	 * @param EventDispatcher|null $eventDispatcher
 	 */
-	function __construct(SystemConfig $systemConfig) {
+	public function __construct(SystemConfig $systemConfig, EventDispatcher $eventDispatcher) {
 		$this->userCache = new CappedMemoryCache();
 		$this->systemConfig = $systemConfig;
+		$this->eventDispatcher = $eventDispatcher;
 	}
 
 	/**
@@ -205,6 +215,7 @@ class AllConfig implements \OCP\IConfig {
 	 * @param string $preCondition only update if the config value was previously the value passed as $preCondition
 	 * @throws \OCP\PreConditionNotMetException if a precondition is specified and is not met
 	 * @throws \UnexpectedValueException when trying to store an unexpected value
+	 * @return bool
 	 */
 	public function setUserValue($userId, $appName, $key, $value, $preCondition = null) {
 		if (!is_int($value) && !is_float($value) && !is_string($value)) {
@@ -215,12 +226,18 @@ class AllConfig implements \OCP\IConfig {
 		$this->fixDIInit();
 
 		$preconditionArray = [];
-		if (isset($preCondition)) {
+		if ($preCondition !== null) {
 			$preconditionArray = [
 				'configvalue' => $preCondition,
 			];
 		}
 
+		$arguments = [
+			'uid' => $userId, 'key' => $key, 'value' => $value,
+			'app' => $appName, 'precondition' => $preCondition
+		];
+		$this->eventDispatcher->dispatch('userpreferences.beforeSetValue',
+			new GenericEvent(null, $arguments));
 		$this->connection->setValues('preferences', [
 			'userid' => $userId,
 			'appid' => $appName,
@@ -236,6 +253,10 @@ class AllConfig implements \OCP\IConfig {
 			}
 			$this->userCache[$userId][$appName][$key] = $value;
 		}
+
+		$this->eventDispatcher->dispatch('userpreferences.afterSetValue', new GenericEvent(null, $arguments));
+
+		return true;
 	}
 
 	/**
@@ -249,11 +270,10 @@ class AllConfig implements \OCP\IConfig {
 	 */
 	public function getUserValue($userId, $appName, $key, $default = '') {
 		$data = $this->getUserValues($userId);
-		if (isset($data[$appName]) and isset($data[$appName][$key])) {
+		if (isset($data[$appName]) && isset($data[$appName][$key])) {
 			return $data[$appName][$key];
-		} else {
-			return $default;
 		}
+		return $default;
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


PVince81:
@sharidas please take over and add tests

(and port to master after merge)

PVince81:
unfortunately I didn't find how to trigger the "delete app" code path, it is possible that this can only be triggered by uninstalling an app through the app manager in the settings page

PVince81:
Testing code looks fine. @sharidas please review my part and test, thanks

sharidas:
@PVince81 Yes I am trying to check the delete app part.

sharidas:
Modified the code in `disableapp.php`
```
//OC_App::disable($appId);
\OC::$server->getConfig()->deleteAppFromAllUsers($appId);
```
And clicked disable. Traced the code flow and found that before and after events are executed.

sharidas:
- `deleteUserValue()` call is verified using debugger. When user logs in the before and after events (`userpreferences.beforedeletevalue` and `userpreferences.afterdeletevalue`) are emitted from LoginController.
- `deleteAllUserValues()` remove the user. Verified that before and after events (`userpreferences.beforedeleteuser` and `userpreferences.afterdeleteuser`) are triggered.
- `setUserValue()` call is verified. The rain loop app when enabled and as admin user when tried to logout both `userpreferences.beforesetvalue` and `userpreferences.aftervalue` are emitted successfully.

sharidas:
Forward port to master branch: https://github.com/owncloud/core/pull/31307

sharidas:
Tried to improve code of `AllConfig.php` a little more ( which is not part of the PR ).

jvillafanez:
I'd rather avoid the use of traits and use DI instead. The problem I see is that you're forced now to use whatever object the trait uses in the tests, and you can't mock it. Basically, you aren't unittesting the class because you're using another object in the tests.

This problem is also affecting the tests: ideally the tests should just check that an event has been triggered using something like:
```
$dispatcherMock->expects($this->once())->method('dispatch')->with($this->equalTo('event.name'), $this->equalTo($eventMock));
```

There is the additional problem of code readability the `$this->emittingCall` call.
```
$dispatcher->dispatch('event1', $event);
/*
 * Do a lot of stuff
 */
$dispatcher->dispatch('event2', $event);
```
This seems quite easier to read and understand than using the `$this->emittingCall`

For the specific events, I don't know what do we want exactly. I see an usability problem which might be a bit difficult to solve. Right now, if I'm interested in a specific key to be deleted, I should listen to the `userpreferences.afterdeletevalue` event, but I won't get any event if any of the other delete method is used. Basically, you're forcing people to listen to all those events.

In addition, note that we have to keep consistency with the event names, specially if we don't use custom events (whose name could be easily be inside a class constant). I think all the event names are camel-case, so using other kind of names will be confusing.

sharidas:
@jvillafanez So from https://github.com/owncloud/core/pull/31266#issuecomment-386543876 I do understand that I would have to update the PR, using DI and try to remove traits. But I have a query regarding 

> Right now, if I'm interested in a specific key to be deleted, I should listen to the userpreferences.afterdeletevalue event, but I won't get any event if any of the other delete method is used

Even for each delete events we have different events emitted. I am not clear here.

sharidas:
I guess it would be better to document this scenario. I am having a very ugly solution of passing cache to the event argument, which can not be considered.

lib/private/AllConfig.php
```diff
@@ -278,52 +301,76 @@ public function getUserKeys($userId, $appName) {
...
+	 * @return bool
 	 */
 	public function deleteUserValue($userId, $appName, $key) {
 		// TODO - FIXME
 		$this->fixDIInit();
 
+		$arguments = ['user' => $userId, 'key' => $key, 'app' => $appName];
+		$this->eventDispatcher->dispatch('userpreferences.beforedeletevalue',
+			new GenericEvent(null, $arguments));
+		$this->eventDispatcher->dispatch('userpreferences.afterdeletevalue',
```
PVince81: why are you sending "afterdeletevalue" before actually deleting the value from the DB ?

jvillafanez:
Since we won't use custom events for now, at least we should keep the event names consistent. I think all the event names are camelcase, so we should stick with this convention.

sharidas:
@jvillafanez The only reason why camelcase wasn't used is because of https://symfony.com/doc/current/components/event_dispatcher.html#naming-conventions. If we still need camel case, then I can do it.

lib/private/AllConfig.php
```diff
@@ -278,52 +301,76 @@ public function getUserKeys($userId, $appName) {
...
 	 *
 	 * @param string $userId the userId of the user that we want to remove all values from
+	 * @return bool
 	 */
 	public function deleteAllUserValues($userId) {
 		// TODO - FIXME
 		$this->fixDIInit();
 
+		$arguments = ['user' => $userId];
+		$this->eventDispatcher->dispatch('userpreferences.beforedeleteuser', new GenericEvent(null, $arguments));
```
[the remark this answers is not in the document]
sharidas: Yes we have `user.afterdelete` here -> https://github.com/owncloud/core/blob/master/lib/private/Server.php#L341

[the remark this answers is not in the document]
jvillafanez: They are different events. This one is triggered when all the keys are deleted, not when the user is deleted. We might want to delete all the keys but not the user (it's a weird case, but it might happen).
The question is whether we want to trigger this one or not. I'd rather keep it for the sake of consistency.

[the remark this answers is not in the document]
sharidas: @jvillafanez Indeed I agree with you to have this one. My intention behind sharing the link was to point out: we have event for deleting the user. But we don't have the one for deleting keys of user.

jvillafanez:
> The only reason why camelcase wasn't used is because of https://symfony.com/doc/current/components/event_dispatcher.html#naming-conventions. If we still need camel case, then I can do it.

Then we have a problem. The event names aren't consistent with what we have in other parts of the code. I don't think we can change the rest of the code at this moment. Unless someone says otherwise, I'd prioritize consistency

sharidas:
> Unless someone says otherwise, I'd prioritize consistency

I have changed event names to camel case.

PVince81:
@jvillafanez @sharidas just be consistent with whatever case we have used for all the other symfony events we already have. Here use the format we used for the appconfig ones for consistency at least across config related events.

lib/private/AllConfig.php
```diff
@@ -215,12 +230,18 @@ public function setUserValue($userId, $appName, $key, $value, $preCondition = nu
 		$this->fixDIInit();
 
 		$preconditionArray = [];
-		if (isset($preCondition)) {
+		if (null !== $preCondition) {
```
jvillafanez: I'd rather switch the condition: `if ($precondition !== null)`

lib/private/AllConfig.php
```diff
@@ -65,11 +71,19 @@ class AllConfig implements \OCP\IConfig {
 	private $userCache;
 
 	/**
+	 * AllConfig constructor.
+	 *
 	 * @param SystemConfig $systemConfig
+	 * @param EventDispatcher|null $eventDispatcher
 	 */
-	function __construct(SystemConfig $systemConfig) {
+	public function __construct(SystemConfig $systemConfig, EventDispatcher $eventDispatcher = null) {
```
jvillafanez: I'm not sure about not requiring the eventDispatcher. Note that if the event dispatcher isn't injected, we won't send notifications (nobody will be listener to events dispatched by the one created inside this class). This can be confusingin the long run.
If there is a use case where we don't want to send these events, there might be better ways to do it.

lib/private/AllConfig.php
sharidas: Updated, with `if ($precondition !== null)`

lib/private/AllConfig.php
sharidas: @jvillafanez I have updated the constructor and hence made changes in places where the changes are required.

lib/private/AllConfig.php
phil-davis: The anti-Yoda code style checks will not like anyway, when a forward-port to master is attempted.
Poor Yoda, loved his grammar is not.

lib/base.php
```diff
@@ -144,7 +144,7 @@ public static function initPaths() {
 				'SCRIPT_FILENAME' => $_SERVER['SCRIPT_FILENAME'],
 			],
 		];
-		$fakeRequest = new \OC\AppFramework\Http\Request($params, null, new \OC\AllConfig(new \OC\SystemConfig(self::$config)));
+		$fakeRequest = new \OC\AppFramework\Http\Request($params, null, new \OC\AllConfig(new \OC\SystemConfig(self::$config), new \Symfony\Component\EventDispatcher\EventDispatcher()));
```
jvillafanez: I'd include a comment here. The dispatcher we're injecting will act as a dummy (which I think it fine in this case). The comment should help to clarify that we don't expect events to be listened here, so it isn't a bug that modifications in the configuration made by the fake request (which in fact there isn't a reason to do so) won't throw events.
In addition, split the line because it's too long

sharidas: Done.

lib/private/AllConfig.php
```diff
@@ -215,12 +226,18 @@ public function setUserValue($userId, $appName, $key, $value, $preCondition = nu
...
 		$preconditionArray = [];
-		if (isset($preCondition)) {
+		if ($preCondition !== null) {
 			$preconditionArray = [
 				'configvalue' => $preCondition,
 			];
 		}
 
+		$arguments = [
+			'user' => $userId, 'key' => $key, 'value' => $value,
```
butonic: `'userId' => $userId, 'key' => $key, 'value' => $value,` or
`'uid' => $userId, 'key' => $key, 'value' => $value,`

butonic (changes requested):
`user` would indicate the user object, `userId` should be used. If you must I would accept `uid`.

lib/private/AllConfig.php
sharidas: Updated. Replaced `user` with `uid`.

PVince81:
As I see the comments were adressed :+1:

PVince81:
@sharidas please port to master

sharidas:
Master branch PR updated: https://github.com/owncloud/core/pull/31307

PR #31285, opened 2018-04-26
The Great Fix of object store acceptance tests ....

DeepDiver1975:
## Description
This PR holds a great variety of fixes which are needed to pass our acceptance test suites.

## Related Issue
refs https://github.com/owncloud/files_primary_s3/pull/48

## Motivation and Context
Acceptance tests should run properly with objectstore as well

## How Has This Been Tested?
- manually
- ci on https://github.com/owncloud/files_primary_s3/pull/48

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/DAV/CopyPlugin.php
+++ b/apps/dav/lib/DAV/CopyPlugin.php
@@ -25,6 +25,7 @@ use OCA\DAV\Connector\Sabre\Exception\Forbidden;
 use OCA\DAV\Connector\Sabre\File;
 use OCA\DAV\Files\ICopySource;
 use OCP\Files\ForbiddenException;
+use OCP\Lock\ILockingProvider;
 use Sabre\DAV\IFile;
 use Sabre\DAV\Server;
 use Sabre\DAV\ServerPlugin;
@@ -83,7 +84,9 @@ class CopyPlugin extends ServerPlugin {
 				$copySuccess = $sourceNode->copy($destinationNode->getFileInfo()->getPath());
 			}
 			if (!$copySuccess) {
+				$destinationNode->acquireLock(ILockingProvider::LOCK_SHARED);
 				$destinationNode->put($sourceNode->get());
+				$destinationNode->releaseLock(ILockingProvider::LOCK_SHARED);
 			}
 
 			$this->server->emit('afterBind', [$copyInfo['destination']]);
--- a/lib/private/Files/Cache/Propagator.php
+++ b/lib/private/Files/Cache/Propagator.php
@@ -93,9 +93,9 @@ class Propagator implements IPropagator {
 				->where($builder->expr()->eq('storage', $builder->createNamedParameter($storageId, IQueryBuilder::PARAM_INT)))
 				->andWhere($builder->expr()->in('path_hash', $hashParams))
 				->andWhere($builder->expr()->gt('size', $builder->expr()->literal(-1, IQueryBuilder::PARAM_INT)));
-		}
 
-		$builder->execute();
+			$builder->execute();
+		}
 	}
 
 	protected function getParents($path) {
--- a/lib/private/Files/ObjectStore/NoopScanner.php
+++ b/lib/private/Files/ObjectStore/NoopScanner.php
@@ -57,7 +57,7 @@ class NoopScanner extends Scanner {
 	public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $lock = true) {
 		// we only update the checksums - still returning no data
 		$meta = $this->storage->getMetaData($path);
-		if (isset($meta['checksum'])) {
+		if (!empty($meta['checksum'])) {
 			$this->storage->getCache()->put(
 				$path,
 				['checksum' => $meta['checksum']]
--- a/lib/private/Share20/DefaultShareProvider.php
+++ b/lib/private/Share20/DefaultShareProvider.php
@@ -663,7 +663,8 @@ class DefaultShareProvider implements IShareProvider {
 		// exclude shares leading to trashbin on home storages
 		$pathSections = \explode('/', $data['path'], 2);
 		// FIXME: would not detect rare md5'd home storage case properly
-		if ($pathSections[0] !== 'files' && \explode(':', $data['storage_string_id'], 2)[0] === 'home') {
+		$storagePrefix = \explode(':', $data['storage_string_id'], 2)[0];
+		if ($pathSections[0] !== 'files' && \in_array($storagePrefix, ['home', 'object'], true)) {
 			return false;
 		}
 		return true;
--- a/lib/private/User/User.php
+++ b/lib/private/User/User.php
@@ -235,6 +235,7 @@ class User implements IUser {
 
 		// Delete the users entry in the storage table
 		Storage::remove('home::' . $this->getUID());
+		Storage::remove('object::user:' . $this->getUID());
 
 		\OC::$server->getCommentsManager()->deleteReferencesOfActor('users', $this->getUID());
 		\OC::$server->getCommentsManager()->deleteReadMarksFromUser($this);
--- a/tests/acceptance/features/apiTrashbin/trashbin-new-endpoint.feature
+++ b/tests/acceptance/features/apiTrashbin/trashbin-new-endpoint.feature
@@ -182,6 +182,7 @@ Feature: trashbin-new-endpoint
 
 	@local_storage
 	@no_default_encryption
+	@skip_on_objectstore
 	Scenario: Deleting a folder in external storage moves it to the trashbin
 		Given the administrator has invoked occ command "files:scan --all"
 		And user "user0" has been created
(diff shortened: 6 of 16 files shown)


tests/acceptance/features/apiTrashbin/trashbin-new-endpoint.feature
```diff
@@ -191,6 +193,7 @@ Feature: trashbin-new-endpoint
 
 	@local_storage
 	@no_default_encryption
+    @skip-on-objectstore
```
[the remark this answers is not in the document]
phil-davis: And use one of the "standards" for naming:
API acceptance tests have made tag names like ``skip_on_objectstore``
webUI acceptance tests have made camelCase tag name like ``skipOnLDAP``
hmmm - it's a shame these diverging "standards" happen.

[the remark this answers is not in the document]
DeepDiver1975: > indentation

github display issue from my pov - I have 4 spaces locally ....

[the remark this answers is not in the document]
DeepDiver1975: underscores are so damn ugly ... I hate them in each and every owncloud app name .....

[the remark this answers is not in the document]
patrickjahns: > hmmm - it's a shame these diverging "standards" happen. 

we can refactor and unify it again - if we have a standard we can also document it and ask people to follow it. Worst case, have a gherkin "code-style" in place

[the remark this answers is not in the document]
DeepDiver1975: adopted to existing standard with _

lib/private/Files/ObjectStore/NoopScanner.php
```diff
@@ -58,7 +58,7 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 	public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $lock = true) {
 		// we only update the checksums - still returning no data
 		$meta = $this->storage->getMetaData($path);
-		if (isset($meta['checksum'])) {
+		if (isset($meta['checksum']) && !empty($meta['checksum'])) {
```
butonic: redundant, only  `!empty` is sufficient, see https://stackoverflow.com/a/4559976

butonic (changes requested):

lib/private/Files/ObjectStore/NoopScanner.php
patrickjahns: Not exactly - different behaviors: https://www.virendrachandak.com/techtalk/php-isset-vs-empty-vs-is_null/

lib/private/Share20/DefaultShareProvider.php
```diff
@@ -669,7 +669,8 @@ private function isAccessibleResult($data) {
 		// exclude shares leading to trashbin on home storages
 		$pathSections = \explode('/', $data['path'], 2);
 		// FIXME: would not detect rare md5'd home storage case properly
-		if ($pathSections[0] !== 'files' && \explode(':', $data['storage_string_id'], 2)[0] === 'home') {
+		$storagePrefix = \explode(':', $data['storage_string_id'], 2)[0];
+		if ($pathSections[0] !== 'files' && \in_array($storagePrefix, ['home', 'object'], true)) {
```
mrow4a: Why object?

lib/private/User/User.php
```diff
@@ -236,6 +236,7 @@ public function delete() {
 
 		// Delete the users entry in the storage table
 		Storage::remove('home::' . $this->getUID());
+		Storage::remove('object::user:' . $this->getUID());
```
mrow4a: so there are now two homes for user?

lib/private/Files/Cache/Propagator.php
```diff
@@ -93,9 +93,9 @@ public function propagateChange($internalPath, $time, $sizeDifference = 0) {
 				->where($builder->expr()->eq('storage', $builder->createNamedParameter($storageId, IQueryBuilder::PARAM_INT)))
 				->andWhere($builder->expr()->in('path_hash', $hashParams))
 				->andWhere($builder->expr()->gt('size', $builder->expr()->literal(-1, IQueryBuilder::PARAM_INT)));
-		}
 
-		$builder->execute();
+			$builder->execute();
```
mrow4a: :>

apps/dav/lib/DAV/CopyPlugin.php
```diff
@@ -83,7 +84,9 @@ function httpCopy(RequestInterface $request, ResponseInterface $response) {
 				$copySuccess = $sourceNode->copy($destinationNode->getFileInfo()->getPath());
 			}
 			if (!$copySuccess) {
+				$destinationNode->acquireLock(ILockingProvider::LOCK_SHARED);
 				$destinationNode->put($sourceNode->get());
```
mrow4a: `$destinationNode->put` itself is not setting locks ? Need to go deeper, it is a logic of put itself, isnt it?

lib/private/Share20/DefaultShareProvider.php
DeepDiver1975: There are basically two different types of storage prefixes: 'home' for regular file system and 'object' for objectstore - the later part was missing in some places - fixed in this PR

lib/private/User/User.php
DeepDiver1975: see above

lib/private/Files/Cache/Propagator.php
DeepDiver1975: :twisted_rightwards_arrows:

apps/dav/lib/DAV/CopyPlugin.php
DeepDiver1975: search for other places there put is used - the node is locked before

lib/private/Share20/DefaultShareProvider.php
mrow4a: ok, makes sense!

lib/private/User/User.php
mrow4a: can be something else then X=user in object::X:  ?

lib/private/Files/ObjectStore/NoopScanner.php
butonic: `isset` checks if the variable, in this case the array key is set, and the value is **not** `null`.
`!empty` does the same thing: `empty` will check if the the array key is set, and the value **is** `null` (or `''`, `'0'`, `0`, `0.0`, `false` or `[]`). the result is negated. it literally checks what `isset` did again.

You could nitpick that `!empty` will return false for `'0'`, `0`, `0.0` ... which may or may not be valid checksums. But I think @DeepDiver1975 just added the empty check to handle `''` correctly. And checksums with `''`, `'0'`, `0`, `0.0`, `null`, `false` or `[]` dont make much sense.

Get rid of the `isset`. it is redundant.

butonic (approved):

patrickjahns:
Since this PR got merged, I am seeing frequent issues in the logfile 

```
{"reqId":"PvBstjnGLSwoDKQm9Ktg","level":3,"time":"2018-05-18T11:35:12+00:00","remoteAddr":"172.20.0.5","user":"user1","app":"PHP","method":"GET","url":"\/ocs\/v1.php\/cloud\/users\/user1","message":"fclose() expects parameter 1 to be resource, object given at \/var\/www\/owncloud\/lib\/private\/legacy\/util.php#437"}
```

mrow4a:
Can you track down in what case this copyr gets object instead of resource and how to reproduce it ?

patrickjahns:
first guess as it just accesses `/ocs/v1.php/cloud/users/user1` - this might be related to ocs api calls - will push some more objectstorage tests into apps and get steps to reproduce

DeepDiver1975:
copyr is used to copy over the skeleton files on first login ...

mrow4a:
No idea, I have never been in that part of code - it is also legacy - I assumed it fixes some bug, but apparently it makes some tests fail, probably also because of how unit tests are mocking things.

phil-davis:
I fixed other things in the test scenario "Trashbin can be emptied" in master PR https://github.com/owncloud/core/pull/31523
When backporting to stable10, I discovered that I needed the commit:
"Handle CSRF token errors on ajax/delete.php request"
from here.
That was because I  fixed scenario "Trashbin can be emptied" so that it really worked, and had the possibility to fail.
So that commit "Handle CSRF token errors on ajax/delete.php request" has been backported to stable10 as part of https://github.com/owncloud/core/pull/31536

phil-davis:
@DeepDiver1975 what is happening about backporting this PR?
It would be nice to have at least other bits-and-pieces from here backported, if not the whole PR.

phil-davis:
@DeepDiver1975 what is the status of backporting this PR?

This is another one that seems to have a pile of commits gathering..

phil-davis:
I am looking through the listed of clossed PRs that still have backport-request label.
Thankfully most are easy, like this. The backport has been done and just the label needs to be removed.

PR #31286, opened 2018-04-26
Remember more created groups

phil-davis:
## Description
1) At the moment, when deleting groups in the acceptance tests, the specific step actually always uses API version 2. Make it use the currently selected API version so the step can properly be used in either v1 or v2 tests.
2) In the ``deleteGroups`` API acceptance tests, use the special steps for deleting a group (rather than the step that "sends HTTP method..."). This way the test code will be aware that the group should have been deleted, and so will not complain about it not being there in the ``AfterScenario`` cleanup.
3) The scenario about "getting an empty group" actually belongs in ``getGroup.feature`` because it is listing the contents of a single group. Move it there.

## Motivation and Context
Improve API acceptance tests for groups.

## How Has This Been Tested?
CI knows

## Types of changes
- [x] Acceptance test fixing

## Checklist:

### DIFF
--- a/tests/TestHelpers/UserHelper.php
+++ b/tests/TestHelpers/UserHelper.php
@@ -150,12 +150,12 @@ class UserHelper {
 	 * @return ResponseInterface
 	 */
 	public static function deleteGroup(
-		$baseUrl, $group, $adminUser, $adminPassword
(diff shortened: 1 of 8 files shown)


phil-davis:
@paurakhsharma these changes will be useful for keeping track of deleted groups better.

individual-it (approved):

PR #31290, opened 2018-04-27
Ensure webUI tests are only run 1 suite at a time

phil-davis:
## Description
If a developer "accidentally" tries to run all the webUI tests in a single run, give an error message.
Report what suite is being run (and if just a specific feature or scenario, report that)

## Motivation and Context
It is completely impractical to run all webUI tests in a single pass. We should prevent developers accidentally doing that. Also there are new test suites/features/scenarios coming that will have specific inter-app requirements for things - so we will need to run those as inidividual suites not as part of a huge single run of all webUI tests.

## How Has This Been Tested?
- Try to run all webUI tests
```
phil@phil-Inspiron-5468:~/ownCloud/core$ ./tests/travis/start_ui_tests.sh
ERROR: webUI tests must be run by suite.
No suite specified. Specify a suite by either:
  setting the BEHAT_SUITE env variable
  using the --suite parameter
```
- Run just a feature/scenario:
```
phil@phil-Inspiron-5468:~/ownCloud/core$ ./tests/travis/start_ui_tests.sh --feature tests/acceptance/features/webUILogin/login.feature:36
Running webUILogin login.feature:36 tests on 'chrome' () on 
@webUI @insulated @disablePreviews
Feature: login users
```

## Types of changes
- [x] Test infrastructure

## Checklist:

Documentation needs to be updated to say that a suite or feature is required.

### DIFF
--- a/tests/travis/start_ui_tests.sh
+++ b/tests/travis/start_ui_tests.sh
@@ -152,6 +152,13 @@ then
 	BEHAT_FEATURE="$1"
 fi
 
+# If a feature file has been specified but no suite, then deduce the suite
+if [ -n "$BEHAT_FEATURE" ] && [ -z "$BEHAT_SUITE" ]
+then
+    FEATURE_PATH=`dirname $BEHAT_FEATURE`
+    BEHAT_SUITE=`basename $FEATURE_PATH`
+fi
+
 if [ -z "$BEHAT_YML" ]
 then
 	BEHAT_YML="tests/acceptance/config/behat.yml"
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Backport stable10 #31291

phil-davis:
Doc issue https://github.com/owncloud/documentation/issues/4094

patrickjahns:
I don't fully agree here - we want to have the ability to run the full suite without needing to list all suites/features.

Use cases:
- run the full suite when a different app has been enabled / disabled
- run the full suite against a specific (remote) environment 

It needs to be possible to run the full suites when desired

phil-davis:
That will be tricky if/when we put some app-combination tests into core. Those tests will require that some combination of apps be available. So when we do that, we will have to think what we do for local testing in a dev environment - do we ``git clone`` any missing apps automagically? can those apps set themselves up automagically? Are there any conflicts between suites, e.g. in some suites there has to be objectstore and user_ldap, then in the next suite we want s3, no objectstore and some other authentication...

Perhaps we can make it all smart enough to enable/setup/disable/... different apps between suites. For that the script is likely to need to loop through running each suite as a separate Behat run, and collect the overall result.

So we could make the script understand what is a full "core-only" set of suites and be able to run those together in a single run. That could be an option, if people want it.

Also, I am not so confident in keeping the selenium server, webdriver... up and running for a single run of 100's of webUI scenarios. When 1 thing goes wrong, then the whole single sequential set of tests has to be run again. Of course, selenium should run reliably for weeks on end - but then there is reality.

PR #28166, opened 2017-06-20
Cache in filecache

butonic:
alternative approach to https://github.com/owncloud/core/pull/28165

PR #31299, opened 2018-04-27
Don't check for existence twice if we know it is not a part file (objectstorage only)

mrow4a:
### Description

Don't check for existence in the storage (e.g. objectstorage) twice if we know it is not a part file. 

In that section of the code, we were checking for file existence in the storage twice for the same path. In case of part file we indeed need to check twice, but otherwise not. 

There is no difference for Filesystem storage, but Objectstorage/Files_primary_s3 is affected.

- [x] Requires unit tests adjustments due to changed function

### How did I test

perf-obj-part

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -384,7 +384,11 @@ class Encryption extends Wrapper {
 		if ($this->util->isExcluded($fullPath) === false) {
 			$size = $unencryptedSize = 0;
 			$realFile = $this->util->stripPartialFileExtension($path);
-			$targetExists = $this->file_exists($realFile) || $this->file_exists($path);
+			$targetExists = false;
+			if ($this->file_exists($realFile) || ($realFile != $path && $this->file_exists($path))) {
+				// Original file exists, or part file exists
+				$targetExists = true;
+			}
 			$targetIsEncrypted = false;
 			if ($targetExists) {
 				// in case the file exists we require the explicit module as
@@ -875,24 +879,29 @@ class Encryption extends Wrapper {
 	}
 
 	/**
-	 * read first block of encrypted file, typically this will contain the
+	 * Read first block of encrypted file from resource, typically this will contain the
 	 * encryption header
 	 *
 	 * @param string|resource $path
 	 * @return string
 	 */
-	protected function readFirstBlock($path) {
-		if (\is_resource($path)) {
-			$firstBlock = \fread($path, $this->util->getHeaderSize());
-			\rewind($path);
-			return $firstBlock;
-		}
-		$firstBlock = '';
-		if ($this->storage->file_exists($path)) {
-			$handle = $this->storage->fopen($path, 'r');
-			$firstBlock = \fread($handle, $this->util->getHeaderSize());
-			\fclose($handle);
-		}
+	protected function readResourceFirstBlock($path) {
+		$firstBlock = \fread($path, $this->util->getHeaderSize());
+		\rewind($path);
+		return $firstBlock;
+	}
+
+	/**
+	 * Read first block of encrypted file from the storage, typically this will contain the
+	 * encryption header
+	 *
+	 * @param string|resource $path
+	 * @return string
+	 */
+	protected function readStorageFirstBlock($path) {
+		$handle = $this->storage->fopen($path, 'r');
+		$firstBlock = \fread($handle, $this->util->getHeaderSize());
+		\fclose($handle);
 		return $firstBlock;
 	}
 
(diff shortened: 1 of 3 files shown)


mrow4a:
This code is confusing, so better compare differences in the PhpStorm

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -888,18 +888,26 @@ protected function getFullPath($path) {
...
 
 	/**
-	 * read first block of encrypted file, typically this will contain the
+	 * Read first block of encrypted file from resource, typically this will contain the
 	 * encryption header
 	 *
 	 * @param string|resource $path
 	 * @return string
 	 */
-	protected function readFirstBlock($path) {
```
mrow4a: readFirstBlock was generalized for resource and storage -> I created dedicated functions, readResourceFirstBlock and readStorageFirstBlock.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -967,25 +984,35 @@ protected function parseRawHeader($rawHeader) {
 	 * @return array
 	 */
 	protected function getHeader($path) {
+		$filecacheExists = false;
 		if (\is_resource($path)) {
```
mrow4a: Now I handle the "header" logic separately for resource and separately for storage.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -967,25 +984,35 @@ protected function parseRawHeader($rawHeader) {
...
 			$realFile = $this->util->stripPartialFileExtension($path);
-			$exists = $this->storage->file_exists($realFile);
-			if ($exists) {
+			if ($this->storage->file_exists($realFile)) {
+				// Original file pointed by part file exists in the storage,
+				// and thus filecache for this file also exists
+				$firstBlock = $this->readStorageFirstBlock($realFile);
+				$filecacheExists = true;
 				$path = $realFile;
+			} else if ($realFile != $path && $this->storage->file_exists($path)) {
```
mrow4a: It makes no sense to ask again if $realFile==$path..

mrow4a: Due to this fact, we are wining here 1 filecache query for each PUT.

lib/private/Files/Storage/Wrapper/Encryption.php
mrow4a: it was also again checking if path exists.. loosing 1 query per PUT.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -391,7 +391,11 @@ public function fopen($path, $mode) {
 
 			$size = $unencryptedSize = 0;
 			$realFile = $this->util->stripPartialFileExtension($path);
-			$targetExists = $this->file_exists($realFile) || $this->file_exists($path);
+			$targetExists = false;
+			if ($this->file_exists($realFile) || ($realFile != $path && $this->file_exists($path))) {
```
mrow4a: We are wining another 2 filecache queries here (file_exists function with objecstore needs 2 queries to be performed per call) @PVince81 @DeepDiver1975 @butonic

PVince81:
conflicts

mrow4a:
However, if this PR https://github.com/owncloud/core/pull/28166 goes into live (cache in filecache), this PR is more of refactoring and cleaning the code.

mrow4a:
Can be obsoleted by https://github.com/owncloud/core/pull/31486

mrow4a:
I refresh this PR, since due to the fact there is no decision on introducing "cache of filecache", this PR is fixing the code logic.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -904,13 +913,22 @@ protected function readFirstBlock($path) {
...
+				$firstBlock = $this->readStorageFirstBlock($realFile);
+			} else if ($realFile != $path && $this->storage->file_exists($path)) {
+				// Original file does not exists, lets check part file
+				$firstBlock = $this->readStorageFirstBlock($path);
+			} else{
+				// No file under the path, return empty
+				$firstBlock = '';
 			}
 		}
-		$firstBlock = $this->readFirstBlock($path);
```
mrow4a: Instead of using one generalized function, which introduced issue - I have two dedicated ones.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -904,13 +913,22 @@ protected function readFirstBlock($path) {
...
+		if (\is_resource($path)) {
+			// It is resource under the path, read directly resource from block
+			$firstBlock = $this->readResourceFirstBlock($path);
+		} else {
 			$realFile = $this->util->stripPartialFileExtension($path);
 			if ($this->storage->file_exists($realFile)) {
-				$path = $realFile;
+				// Original file pointed by part file exists in the storage
+				$firstBlock = $this->readStorageFirstBlock($realFile);
+			} else if ($realFile != $path && $this->storage->file_exists($path)) {
```
mrow4a: Dont check if file_exists again, if realFile is the same as path (points to the same file)

mrow4a:
Most likely being closed with https://github.com/owncloud/core/pull/31958

mrow4a:
Obsoleted by #31958

PR #31307, opened 2018-04-30
[master] Add emitters for user preferences

### LINKED ISSUE #31264: Symfony events for user preference changes

sharidas:
## Description

Add events for user preference changes

## Related Issue

Fixes https://github.com/owncloud/core/issues/31264

## How Has This Been Tested?

- [x] add unit tests

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/AllConfig.php
+++ b/lib/private/AllConfig.php
@@ -30,18 +30,22 @@
 namespace OC;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use OC\Cache\CappedMemoryCache;
+use OCP\IConfig;
 use OCP\IDBConnection;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class to combine all the configuration options ownCloud offers
  */
-class AllConfig implements \OCP\IConfig {
+class AllConfig implements IConfig {
 	/** @var SystemConfig */
 	private $systemConfig;
 
 	/** @var IDBConnection */
 	private $connection;
 
+	private $eventDispatcher;
 	/**
 	 * 3 dimensional array with the following structure:
 	 * [ $userId =>
@@ -67,9 +71,10 @@ class AllConfig implements \OCP\IConfig {
 	/**
 	 * @param SystemConfig $systemConfig
 	 */
-	public function __construct(SystemConfig $systemConfig) {
+	public function __construct(SystemConfig $systemConfig, EventDispatcher $eventDispatcher) {
 		$this->userCache = new CappedMemoryCache();
 		$this->systemConfig = $systemConfig;
+		$this->eventDispatcher = $eventDispatcher;
 	}
 
 	/**
@@ -204,6 +209,7 @@ class AllConfig implements \OCP\IConfig {
 	 * @param string $preCondition only update if the config value was previously the value passed as $preCondition
 	 * @throws \OCP\PreConditionNotMetException if a precondition is specified and is not met
 	 * @throws \UnexpectedValueException when trying to store an unexpected value
+	 * @return bool
 	 */
 	public function setUserValue($userId, $appName, $key, $value, $preCondition = null) {
 		if (!\is_int($value) && !\is_float($value) && !\is_string($value)) {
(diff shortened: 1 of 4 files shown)


sharidas:
Backport PR: https://github.com/owncloud/core/pull/31266

lib/private/AllConfig.php
```diff
@@ -277,52 +291,75 @@ public function getUserKeys($userId, $appName) {
...
 	 * @param string $userId the userId of the user that we want to remove all values from
+	 * @return bool
 	 */
 	public function deleteAllUserValues($userId) {
 		// TODO - FIXME
 		$this->fixDIInit();
 
+		$arguments = ['uid' => $userId];
+		$this->eventDispatcher->dispatch('userpreferences.beforeDeleteUser', new GenericEvent(null, $arguments));
+
```
DeepDiver1975: one day we need to use query builder everywhere .....

tests/lib/AllConfigTest.php
```diff
@@ -20,7 +22,12 @@ class AllConfigTest extends \Test\TestCase {
 	/** @var  \OCP\IDBConnection */
 	protected $connection;
 
+	/** @var  EventDispatcher */
+	protected $eventDispatcher;
+
 	protected function getConfig($systemConfig = null, $connection = null) {
+		$this->eventDispatcher = new  EventDispatcher();
```
DeepDiver1975: I would have a used a mock here .... no need to run all the event dispatcher logic
and there is one space too much ;-)

tests/lib/AllConfigTest.php
```diff
@@ -60,8 +67,45 @@ public function testSetUserValue() {
 		$selectAllSQL = 'SELECT `userid`, `appid`, `configkey`, `configvalue` FROM `*PREFIX*preferences` WHERE `userid` = ?';
 		$config = $this->getConfig();
 
+		$calledBeforeUserPrefSetVal = [];
+		$this->eventDispatcher->addListener('userpreferences.beforeSetValue',
```
DeepDiver1975: This code should read much easier using a mock

```php
$this->eventDispatcher->expects($this->once())->method('dispatch')->with('userpreferences.beforeDeleteApp', new GenericEvent(null, [....]))

DeepDiver1975 (changes requested):

PVince81:
@sharidas when backporting or forward porting please make sure it is clear. When reading this PR it looks like this is the original PR but it isn't. The original PR is the stable10 one and this one here is a forward port. And now you're getting review change requests for something that is just a port.

tests/lib/AllConfigTest.php
PVince81: it's easier to use the real object instead of mocking all its methods (I tried to mock this one in another PR and gave up). To note is that we use a standalone dispatcher and not the real `\OC::$server->getEventDispatcher()` because that would register and trigger events for the whole system

PVince81:
@DeepDiver1975 considering that this is a master port (wasn't clear from the first post), I'd rather keep the code as is. If you think this code must absolutely be changed then we shall do so in a separate PR that we will backport again to stable10.

sharidas:
Sorry for the confusion created. This PR is a forward port from https://github.com/owncloud/core/pull/31266 ( which is merged ).

phil-davis:
The challenge with forward-ports is that extra CI in ``master``` (like codecov) might point out some issue or deficiency that is already in the code merged to ```stable10``
In most cases, code should first be done in ``master``. If the code concept needs to also be demonstrated for ``stable10``  (e.g. related to behavior of old versions of dependencies or...) then it would be best to develop 2 PRs,, 1 for each branch, to see that it can work for both. Then at the end make a "proper" backport from ```master`` ...

PVince81:
@phil-davis ok that makes sense. The original stable10 PR was created on said branch to save some time back then in case it was needed quickly. You pointed out good reasons to continue starting with master first.

DeepDiver1975:
> @DeepDiver1975 considering that this is a master port (wasn't clear from the first post), I'd rather keep the code as is.

yes - agreed

>  If you think this code must absolutely be changed then we shall do so in a separate PR that we will backport again to stable10.

I'd really like this to be changed @sharidas please have a look if time permits and @PVince81 agrees as well

DeepDiver1975:
> In most cases, code should first be done in master.

this is the prime rule for many years - whenever we broke this rule - code broke as well.

PVince81:
@sharidas has the requested changes ready, please push to a separate master PR which we will backport after review. Thanks

tests/lib/AllConfigTest.php
DeepDiver1975: only one method needs to me mocked and the assertion code is much less - see below

PR #31316, opened 2018-05-01
API acceptance add group name combinations

phil-davis:
## Description
1) Make scenario outlines with more test cases for adding users to groups, removing users from groups, deleting groups... with unusual group names
2) Make better messages when a user or group is not in the expected state.

## Motivation and Context
1) There are only a few test cases for manipulating and deleting unusual group names. There is a PR to try and fix cases that do not work. But for now, it would be good to have test cases for group names that do work. That will avoid regressions, and make the diffs in the "fixing" PR smaller.

2) Messages that come out when a user or group is expected to exist but does not, are unhelpful
```
```
A better message would be good, so we have some idea where the problem might be.

## How Has This Been Tested?
Local Provisioning API acceptance test runs

## Types of changes
- [x] Acceptance test enahncements

## Checklist:

### DIFF
--- a/tests/acceptance/features/apiProvisioning-v1/addToGroup.feature
+++ b/tests/acceptance/features/apiProvisioning-v1/addToGroup.feature
@@ -15,9 +15,19 @@ So that I can give a user access to the resources of the group
 		Then the OCS status code should be "100"
 		And the HTTP status code should be "200"
 		Examples:
-			| group_id  |
-			| new-group |
-			| 0         |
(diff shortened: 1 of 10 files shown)


individual-it (changes requested):
please make lines shorter

phil-davis:
Line lengths fixed, and fixed another unrelated thing that ``phpcs`` reported.

individual-it (approved):

PR #31322, opened 2018-05-01
Correct addGroup.feature and add new text cases to other group tests

phil-davis:
## Description
1) Fix error in column name in ``addGroup.feature`` - ``group-id`` -> ``group_id``
2) URL-encode the group name when testing to see if a group exists - this allows ``groupExists()`` to correctly test for groups that have unusual characters in their name.
3) Remove test cases for groups that have slash in their name - fix for those will be in another PR
4) Add test cases for other ``#`` ``?`` ``%`` in group names to all the places where groups are manipulated - ``addToGroup`` ``deleteGroup`` ``removeFromGroup`` - for both API v1 and v2

Now we have a full range of test cases for group names that do work properly with the existing Provisioning API v1 and v2. This will help to ensure there are no regressions when we fix the "slash in group name" case.

## Motivation and Context
I wondered why the ``addGroup.feature`` test cases for group names containing a slash were passing. Because actually there is a problem with that - see issue #31015 and PR #31017 

I discovered that there was a typo in the column name of the examples table. Actually the scenario outline was running over-and-over and repeatedly testing adding a group with the literal name ``<group_id>`` instead of substituting in the group id in each example table row.

When I make that fix, then a lot of the group creations fail, because the group should be URL-encoded when the group name needs to be sent in the URL (e.g. when checking if the group exists).

After fixing that, I still have to remove the entries from the examples table that have a slash in the group name, because those really do fail. The other PR addresses fixing that problem.

## How Has This Been Tested?
Local acceptance test runs of group tests

## Types of changes
- [x] Acceptance test fixes

## Checklist:

### DIFF
--- a/tests/acceptance/features/apiProvisioning-v1/addGroup.feature
+++ b/tests/acceptance/features/apiProvisioning-v1/addGroup.feature
@@ -13,7 +13,7 @@ So that I can more easily manage access to resources by groups rather than indiv
 		And the HTTP status code should be "200"
 		And group "<group_id>" should exist
 		Examples:
-			| group-id            | comment                                 |
+			| group_id            | comment                                 |
 			| new-group           | dash                                    |
 			| the.group           | dot                                     |
 			| España              | special European characters             |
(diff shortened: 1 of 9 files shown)


individual-it (approved):

phil-davis:
Backport ``stable10`` #31323

PR #31325, opened 2018-05-01
run UI tests in drone on Firefox

individual-it:
## Description
run UI tests in drone on Firefox without using saucelabs

by using the selenium docker containers we are able to run tests in Firefox same as we do in Chrome. This PR
1. sets the environment for it
2. un-skips a couple of tests that actually do run in this version of selenium and firefox
3. fixes one input. setting the Email  address didn't work correctly in Firefox if the field was pre-filled 

Limitations:
1. this might break FF tests in travis, because it un-skips some tests. Question: do we need the travis runs with FF if we run them in drone
2. currently we are limited to selenium 3.8.1, because 3.9.0 removed the `enablePassThrough` option. So without changing the docker container we are limited to Firefox 58.0

## Motivation and Context
get more independent of travis & saucelabs

## How Has This Been Tested?
:robot: 

## Types of changes

## Checklist:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -31,7 +31,7 @@ pipeline:
       - php occ config:system:set trusted_domains 2 --value=federated
       - php occ log:manage --level 0
       - php occ config:list
-      - echo "export TEST_SERVER_FED_URL=https://federated" > /drone/saved-settings.sh
+      - echo "export TEST_SERVER_FED_URL=${SERVER_PROTOCOL}://federated" > /drone/saved-settings.sh
       - php occ security:certificates:import /drone/server.crt
       - php occ security:certificates:import /drone/federated.crt
       - php occ security:certificates
@@ -272,8 +272,8 @@ pipeline:
       - echo 'Sharing a folder ..'
       - OC_PASS=123456 php occ user:add --password-from-env user1
       - chown www-data /drone/src -R
-      - curl -k -s -u user1:123456 -X MKCOL 'https://server/remote.php/webdav/new_folder'
-      - curl -k -s -u user1:123456 "https://server/ocs/v2.php/apps/files_sharing/api/v1/shares" --data 'path=/new_folder&shareType=0&permissions=15&name=new_folder&shareWith=admin'
+      - curl -k -s -u user1:123456 -X MKCOL '${SERVER_PROTOCOL}://server/remote.php/webdav/new_folder'
+      - curl -k -s -u user1:123456 "${SERVER_PROTOCOL}://server/ocs/v2.php/apps/files_sharing/api/v1/shares" --data 'path=/new_folder&shareType=0&permissions=15&name=new_folder&shareWith=admin'
     when:
       matrix:
         TEST_SUITE: litmus
@@ -282,7 +282,7 @@ pipeline:
     image: owncloud/litmus
     pull: true
     environment:
-      - LITMUS_URL=https://server/remote.php/webdav
+      - LITMUS_URL=${SERVER_PROTOCOL}://server/remote.php/webdav
       - LITMUS_USERNAME=admin
       - LITMUS_PASSWORD=admin
     when:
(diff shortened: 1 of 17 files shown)


phil-davis (commented):
LGTM
I suggest that we remove Firefox test matrix entries from Travis.
chrome, IE11 and Edge on nightly Travis should be plenty, and Firefox will happen with every PR.

phil-davis:
This will make another 14 matrix jobs in drone PR builds.
So that is something to think about.
IMO it is easy to have lots of drone bees working, we just need a solar-powered drone hive so the bees are green :)

phil-davis:
drone says it passed, but no codecov status has come. what happened???

patrickjahns (dismissed):
Please do not merge - see further comments

patrickjahns:
> This will make another 14 matrix jobs in drone PR builds.

Exactly - I am against having both browsers being tested during a PR request run. It will just add unnecessary build time to pull requests.

I see testing different browser being deferred to async runs (nightly / weekly). The mechanism for this is in the works. You might be just a little bit ahead of your time ;-)

Personally I would keep this open and revive once the mechanisms are in place. We can at the same time investigate how we can also add browserstack onto drone for the nightly/weekly scenarios

individual-it:
rebased, made Firefox tests run on HTTP and skipped failing tests
@PVince81 @phil-davis @davitol what to do with that PR now?

.drone.yml
```diff
@@ -515,6 +526,18 @@ services:
     when:
       matrix:
         USE_SERVER: true
+        SERVER_PROTOCOL: https
+
+  server:
```
phil-davis: This has the same pipeline step name as ``server`` above.
Only 1 of them ever gets started, but does it cause some confusion for drone.yml parsing?

.drone.yml
```diff
@@ -529,6 +552,18 @@ services:
     when:
       matrix:
         USE_FEDERATED_SERVER: true
+        SERVER_PROTOCOL: https
+
+  federated:
```
phil-davis: Same here, repeating the ```federated:`` step

phil-davis (commented):
Firstly, just 2 questions.

phil-davis:
Maybe for now we could merge all the:
- drone pipeline additions
- putting the ``SERVER_PROTOCOL`` variable into the existing chrome webUI and other API acceptance... matrix entries
- ``skipOnFirefox`` tag changes
- select a couple of webUI test suites to also run all the time on Firefox (that will ensure that we know if something obvious regresses on Firefox) - maybe ``webUILogin`` and ``webUIRenameFiles`` - that will ensure that it can login and do something basic on the files page.

.drone.yml
individual-it: no confusion for drones, only for humans
we need to have the same name because that becomes the hostname for that service and it should be the same for HTTP and HTTPS, to have less confusion

individual-it:
or running all `@smokeTest` tagged tests on FF?

phil-davis (approved):
LGTM
And we can have a followup issue to investigate the test scenarios tagged ``skipOnFirefox`` and try to find out why they fail (real problem on Firefox or Firefox webDriver "feature"...)

patrickjahns:
Just a minor note - the firefox container seems to be quite old
https://github.com/owncloud/core/blob/a5f69f6f61949a9f2f273271d567c5126abe9120/.drone.yml#L507

And it was last built 12 months ago:
```
$ docker images | grep firefox
selenium/standalone-firefox-debug                        3.8.1                                      c2ebdfdb9e31        12 months ago       792MB
```

patrickjahns (dismissed):
Please use a more recent firefox version

individual-it:
@patrickjahns see original description a newer version is not possible because of not finished implementations in the libraries

>     2. currently we are limited to selenium 3.8.1, because 3.9.0 removed the `enablePassThrough` option. So without changing the docker container we are limited to Firefox 58.0

individual-it:
@phil-davis issue opened for skipped tests https://github.com/owncloud/core/issues/34491

individual-it:
@patrickjahns issue to update to a newer version of Firefox https://github.com/owncloud/core/issues/34493

individual-it:
backport in #34494

Issue #29503, opened 2017-11-07
Transition to account ids


PVince81:
Currently, most tables in OC are based on the immutable user id.
Changing all tables to use account id instead would take a long time and the benefit of being able to change a user id is deferred to the moment where all tables were adjusted.
Also changing too many at once in the hope to have this feature soon is time costly and dangerous.

I suggest to use an alternate temporary route:

1. Introduce a new column oc_accounts.internal_user_id. The column defaults to the same value as oc_accounts.user_id at user creation or import.
1. Let the admin rename the user id, in which case only oc_accounts.user_id is changed.
1. Internally, we always use internal_user_id when mapping with other tables. So the code needs to be rewired to work with internal_user_id for internal operations, but keep user_id when dealing with logins or user:sync.

Once this is done, we already gain the benefit of user id renaming.

After that, we can take our time to migrate all other tables one by one to use oc_account.id instead (and use the opportunity to add a foreign key there...).

Once all tables are migrated, we can get rid of oc_accounts.internal_user_id and make it use user_id again.

@tomneedham @cdamken @butonic @DeepDiver1975 @jvillafanez @pmaier1

cdamken:
> Let the admin rename the user id, in which case only oc_accounts.user_id is changed.

This depends on the backend and if they want to update additional stuff, for example, the storage path.

would that be over a occ command?

butonic:
how about taking this one step further: instead of `oc_accounts.internal_user_id` let us add `oc_accounts.uuid`.
- For any existing users the userid will be null.
- For new DB users a [RFC 4122](https://tools.ietf.org/html/rfc4122) UUID will be generated.
- For new LDAP users the UUID attribute will be used.

This way, the account can fall back to the userid if the uuid is null. New users will always get a uuid. Not all admins have configured a proper UUID attribute for ldap, we should add a warning in the log whenever a non RFC4122 UUID is encountered. But for correct ones we could fill the uuid column when changing the existing userid in all the other tables.

Same approach as with the `oc_accounts.internal_user_id` But it would allow us to keep the uuid column and use its nice properties in the future, eg: merging instances, create primary key without a trip to the db, no need for autoincrement primary key.

[...]

PR #31335, opened 2018-05-02
[WIP] allow renaming usernames

### LINKED ISSUE #6984: Possibility to change the username in admin user view or by themselfs
Hi,

Some users of my cloud are using their emailadress as usernames. Because some users changed their primary emailaccounts they want to change their username in owncloud also. Could you make it possible for the users to change their usernames - or that the admin can change it?

Thank you and keep up the great work!

butonic:
## Description
This PR allows renaming usernames by
* treating the lower_user_id column as the user name 
* treading the user_id column as a generated user id
* existing code asking for the uid will get the user_id, making migration unneeded:
  * existing users will keep their user_id
  * new users will get a generated id, the username will be stored in the lower_user_id column

Requires https://github.com/owncloud/user_management/pull/27

# TODO
- [ ] rename lower_user_id column - deferred for now for easier testing
- [ ] introduce new Interfaces for username vs userid. adding new methods to the existing interface may break existing apps
  - IUser
  - IUserManager
- [ ] use proper uuids instead of `uniqid()`
- [ ] sync samaccountname to the lower_user_id column (separate PR)
- [ ] Work on the tests

## Related Issue

- fixes https://github.com/owncloud/core/issues/6984
- allows us to subsequently do https://github.com/owncloud/core/issues/29503
- deprecates https://github.com/owncloud/core/pull/30617

## Motivation and Context

username != userid != display name

## How Has This Been Tested?

manual testing, waiting for ci

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/Database.php
+++ b/lib/private/User/Database.php
@@ -216,21 +216,17 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 
 	/**
 	 * Check if the password is correct
+	 *
 	 * @param string $uid The username
 	 * @param string $password The password
-	 * @return string
+	 * @param string $login the username or email
+	 * @return string|false
 	 *
 	 * Check if the password is correct without logging in the user
 	 * returns the user id or false
+	 * @throws \OC\DatabaseException
 	 */
-	public function checkPassword($loginName, $password) {
-		try {
-			$account = \OC::$server->getAccountMapper()->getByUserName($loginName);
-		} catch (DoesNotExistException $ex) {
-			return false;
-		}
-		$uid = $account->getUserId();
-
+	public function checkPassword($uid, $password, $login = null) {
 		$query = \OC_DB::prepare('SELECT `uid`, `password` FROM `*PREFIX*users` WHERE LOWER(`uid`) = LOWER(?)');
 		$result = $query->execute([$uid]);
 
@@ -243,7 +239,7 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 					$this->setPassword($uid, $password);
 					unset($this->cache[$uid]); // invalidate cache
 				}
-				return $uid;
+				return $row['uid'];
 			}
 		}
 
--- a/lib/private/User/Database.php
+++ b/lib/private/User/Database.php
@@ -94,6 +94,10 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 		// forbid that? might break existing users? force them to change their
 		// username on next login?
 
+		if ($userName === null) {
+			$userName = $uid;
+		}
+
 		unset($this->cache[$uid]); // make sure we are reading from the db
 		if (!$this->userExists($uid)) {
 			$query = \OC_DB::prepare('INSERT INTO `*PREFIX*users` ( `uid`, `password` ) VALUES( ?, ? )');
--- a/lib/private/User/Database.php
+++ b/lib/private/User/Database.php
@@ -51,6 +51,7 @@
 namespace OC\User;
 
 use OC\Cache\CappedMemoryCache;
+use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\IUserBackend;
 use OCP\User\IProvidesDisplayNameBackend;
 use OCP\User\IProvidesEMailBackend;
@@ -80,7 +81,19 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 	 * Creates a new user. Basic checking of username is done in OC_User
 	 * itself, not in its subclasses.
 	 */
-	public function createUser($uid, $password) {
+	public function createUser($uid, $password, $userName = null) {
+
+		// TODO ... we could use the displayname column to store the login.
+		// that way the login works because checkPassword can lookump the uid,
+		// based on login and can check the password
+
+		// or we always lookup the user in the account table
+		// actually login and email together must be unique, otherwise how do we
+		// distinguish a user with a login from a user with the same email when
+		// trying the password. currently username can contain an @.
+		// forbid that? might break existing users? force them to change their
+		// username on next login?
+
 		unset($this->cache[$uid]); // make sure we are reading from the db
 		if (!$this->userExists($uid)) {
 			$query = \OC_DB::prepare('INSERT INTO `*PREFIX*users` ( `uid`, `password` ) VALUES( ?, ? )');
@@ -88,6 +101,9 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 
 			if ($result) {
 				unset($this->cache[$uid]); // invalidate non existing user in cache
+				if ($userName !== null) {
+					$this->setDisplayName($uid, $userName);
+				}
 			}
 
 			return $result ? true : false;
@@ -203,7 +219,14 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 	 * Check if the password is correct without logging in the user
 	 * returns the user id or false
 	 */
-	public function checkPassword($uid, $password) {
+	public function checkPassword($loginName, $password) {
+		try {
+			$account = \OC::$server->getAccountMapper()->getByUserName($loginName);
+		} catch (DoesNotExistException $ex) {
+			return false;
+		}
+		$uid = $account->getUserId();
+
 		$query = \OC_DB::prepare('SELECT `uid`, `password` FROM `*PREFIX*users` WHERE LOWER(`uid`) = LOWER(?)');
 		$result = $query->execute([$uid]);
 
@@ -216,7 +239,7 @@ class Database extends Backend implements IUserBackend, IProvidesHomeBackend, IP
 					$this->setPassword($uid, $password);
 					unset($this->cache[$uid]); // invalidate cache
 				}
-				return $row['uid'];
+				return $uid;
 			}
 		}
 
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -173,9 +173,21 @@ class Manager extends PublicEmitter implements IUserManager {
 	 *
 	 * @param string $uid
 	 * @return \OC\User\User|null Either the user or null if the specified user does not exist
+	 * @throws \OCP\AppFramework\Db\MultipleObjectsReturnedException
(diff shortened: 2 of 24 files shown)


.htaccess
```diff
@@ -85,3 +85,7 @@ Options -Indexes
   ModPagespeed Off
 </IfModule>
 
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
phil-davis: Accidentally added to the commit
(easily done with ``git add -A``)

lib/public/IUserManager.php
```diff
@@ -76,9 +76,28 @@ public function clearBackends() ;
 	 * @param string $uid
 	 * @return \OCP\IUser|null Either the user or null if the specified user does not exist
 	 * @since 8.0.0
+	 * @deprecated use getByUserId or getByUserName to clarify code
```
DeepDiver1975: This is rather radical. Im not Happy about this

DeepDiver1975 (changes requested):

lib/public/IUserManager.php
butonic: the code using this method still works as expected, but I deprecated the method to make sure developers start switching to getByUserId or getByUserName (or getByEMail) to make the intention of the code always clear.

lib/private/User/Database.php
```diff
@@ -80,14 +81,29 @@ public function __construct() {
...
+		// TODO ... we could use the displayname column to store the login.
+		// that way the login works because checkPassword can lookump the uid,
+		// based on login and can check the password
+
+		// or we always lookup the user in the account table
+		// actually login and email together must be unique, otherwise how do we
+		// distinguish a user with a login from a user with the same email when
+		// trying the password. currently username can contain an @.
+		// forbid that? might break existing users? force them to change their
+		// username on next login?
```
mrow4a: What is the problem with `@` in username?

lib/private/User/Database.php
```diff
@@ -80,14 +81,29 @@ public function __construct() {
 	 * Creates a new user. Basic checking of username is done in OC_User
 	 * itself, not in its subclasses.
 	 */
-	public function createUser($uid, $password) {
+	public function createUser($uid, $password, $userName = null) {
+
+		// TODO ... we could use the displayname column to store the login.
```
mrow4a: Ok, just to be clear, userName == displayName == login ?

lib/private/User/Manager.php
```diff
@@ -314,47 +348,79 @@ public function searchDisplayName($pattern, $limit = null, $offset = null) {
 	/**
 	 * @param string $uid
 	 * @param string $password
+	 * @param string $userName
 	 * @throws \Exception
 	 * @return bool|IUser the created user or false
 	 */
-	public function createUser($uid, $password) {
-		return $this->emittingCall(function () use (&$uid, &$password) {
+	public function createUser($uid = null, $password, $userName = null) {
```
mrow4a: Ok, I have to admit, this is confusing. Why `$uid` can be null?

lib/private/User/Manager.php
```diff
@@ -375,15 +441,17 @@ public function createUser($uid, $password) {
...
 	 * @param UserInterface $backend
+	 * @param string $userName
 	 * @return IUser | null
 	 * @deprecated core is responsible for creating accounts, see user_ldap how it is done
 	 */
-	public function createUserFromBackend($uid, $password, $backend) {
-		return $this->emittingCall(function () use (&$uid, &$password, &$backend) {
-			$this->emit('\OC\User', 'preCreateUser', [$uid, $password]);
+	public function createUserFromBackend($uid, $password, $backend, $userName) {
+		return $this->emittingCall(function () use (&$uid, &$password, &$backend, &$userName) {
```
mrow4a: what about userName before backend?

lib/private/User/Session.php
```diff
@@ -308,13 +308,13 @@ public function getLoginName() {
 	 * @return boolean|null
 	 * @throws LoginException
 	 */
-	public function login($uid, $password) {
+	public function login($loginName, $password) {
```
mrow4a: Ok, cool :>

lib/public/IUserManager.php
mrow4a: Why couldn't `get` imply `getByUserId` ?

lib/public/IUserManager.php
```diff
@@ -134,11 +153,12 @@ public function searchDisplayName($pattern, $limit = null, $offset = null);
 	/**
 	 * @param string $uid
 	 * @param string $password
+	 * @param string $userName
 	 * @throws \Exception
 	 * @return bool|\OCP\IUser the created user of false
 	 * @since 8.0.0
 	 */
-	public function createUser($uid, $password);
+	public function createUser($uid = null, $password, $userName = null);
```
mrow4a: Why `uid` can be null? I think you mention you want it later to be account_id, am I correct?

lib/public/IUserManager.php
```diff
@@ -180,8 +200,9 @@ public function getByEmail($email);
 	 * @param string $uid
 	 * @param string $password
 	 * @param UserInterface $backend
+	 * @param string $userName
 	 * @return IUser | null
 	 * @since 10.0
 	 */
-	public function createUserFromBackend($uid, $password, $backend);
+	public function createUserFromBackend($uid, $password, $backend, $userName);
```
mrow4a: username before backend?

mrow4a (commented):
There are many unclear parts to me, also having similar doubt as Thomas

lib/public/IUserManager.php
butonic: It does. But it is used as `getByUserName`. So I need to check every occurence of the get call to double check if it the parameter that is used is really the userid or actually the username. Prominent examples? 
https://github.com/owncloud/core/blob/78eca952cdcae27e049d1a1f126eef164c0ec27d/lib/private/User/Session.php#L621-L626
https://github.com/owncloud/core/blob/78eca952cdcae27e049d1a1f126eef164c0ec27d/lib/private/User/Manager.php#L273-L282

get() still works. It internally uses getByUserID. But we should make as clear as we can how to use it. That means deprecating get() und using properly named functions. This is about readability. And there is nothing radical about it. Existing code will still work.

lib/private/User/Database.php
butonic: so a user uses `foo@bar.com` to login. which user is it when there is a user with the username `foo@bar.com` and a user with the userid `evilmcknevil` who has the email `foo@bar.com`. Who do you then login? no one? the one where the password matches? users currently can just change their email. and it is not even unique. but we still allow login by email. so if I want to prevent you from logging in with your email I just set my email to yours. I hope you still know your username to log in with it. Oh wait ... you username IS your email ... too bad for you... this affects all existing guest users.

lib/private/User/Database.php
butonic: userName != displayName != login

checkPassword gets the login and password. but the database backend does not store the username. only the userid. I was thinking about misusing the displayname as username, but It makes more sense to actually lookup the userid with the login and then check the password for that uiserid.

lib/private/User/Manager.php
butonic: so that the current code to create users still works. The create User method is used in many places. This will make existing code, that does not know about the userid / username split, work correctly: If you do not pass a username, uid will be used as username and the userId will be generated.

lib/public/IUserManager.php
butonic: legacy code will only call `createUser('foo','secret')` -> the new implementation will use 'foo' as userName and generate a userId
now code can call `createUser(null, 'secret', 'foo')` to achieve the same thing or call `createUser('c30cf2d0-51e0-11e8-b566-0800200c9a66','secret','foo')` if it wants to also set the uid, which makes sense for ldap, which provides a uuid attribute that is used to identify users when syncing them

lib/public/IUserManager.php
butonic: would change the api

lib/public/IUser.php
```diff
@@ -38,8 +38,32 @@
...
+	 * @deprecated use getUserId or getUserName to clarify code
 	 */
 	public function getUID();
+	/**
+	 * get the user id
+	 *
+	 * @return string
+	 * @since 10.0.9
+	 */
+	public function getUserId();
```
butonic: Well ok, this won't work, because adding methods to existing public interfaceslike this  will break code that implements these interfaces.

butonic: @DeepDiver1975 I wish we would use abstract classes for the API classes that are intended to be used by apps, eg the `IUserManager` or `IUser`. Not being able to add methods is rather painful. We should add an `AbstractUser` and expose that in the AppFramework.

see https://theamiableapi.com/java-cl-section-2-1/#cl.item.2.1.8. I know that is java ... it still holds true for the app framework:
> Abstract classes can be used everywhere Java interfaces can with the limitation that a class can only extend a single (abstract) superclass while it can implement multiple interfaces. In most APIs Java’s lack of support for multiple class inheritance is not an issue because the interfaces are _**not intended to be implemented by clients**_. In these cases abstract classes are the better design choice.

PVince81:
@butonic any plans to finish this ? if not, please put it in a form that someone can take over, expand the concept parts a bit

butonic:
@PVince81 I would like to make the differentiation between userid and username explicit, by deprecating get() which has no clear meaning. We can leave it in with the meaning of getUserId or gerUserName. Most existing codepaths expect uid. That is what this PR implements. The current problem is installation, which first creates an admin user and then executes the migrations. But this pr changes the checkpassword call and that causes havoc because it no longer works with the old tables... Will continue after vacations.

butonic:
We introduced a username into the public IUser interface with https://github.com/owncloud/core/pull/31951/files#diff-dc2e2ee403cebc838875eddf32a8a7beR56 The internal User class implements in using the preferences table. It is used to allow syncing the login (samaccountname) from ldap and use it to mound a wnd. This works but could be further developed to move the username into the account table. Actually the lower_user_id column should be renamed to username and any username from the preferences table should be copied to the column in the account table.

This PR now documents some of the challenges that are involved in doing that, because a naive approach will cause the installation to fail.

Anyway, since I assume user management will be handled differently with OCIS which is based on OpenID Connect, I will unassign myself and close this PR.

PR #31359, opened 2018-05-04
also try syncing quote from preferences if backend provided no quota

butonic:
When upgrading from 9.1 to 10 a sync of ldap users should also sync the quota when ldap does not provide a quota.

- tests included


voroyam (approved):
confirmed. quota stays after the upgrade to 10.0.8

PR #31407, opened 2018-05-09
Properly normalize paths for event, no &$magic needed

butonic:
The paths in the begin and after events are currently not normalized, leading to double slashes in logged paths.

This is a safeguard against broken code as in https://github.com/owncloud/core/pull/31406

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -889,20 +889,21 @@ class View {
 	 *
 	 * @return bool|mixed
 	 */
+
 	public function copy($path1, $path2, $preserveMtime = false) {
-		return $this->emittingCall(function () use (&$path1, &$path2, &$preserveMtime) {
-			$absolutePath1 = Filesystem::normalizePath($this->getAbsolutePath($path1));
-			$absolutePath2 = Filesystem::normalizePath($this->getAbsolutePath($path2));
+		$absolutePath1 = Filesystem::normalizePath($this->getAbsolutePath($path1));
+		$absolutePath2 = Filesystem::normalizePath($this->getAbsolutePath($path2));
+		return $this->emittingCall(function () use ($absolutePath1, $absolutePath2) {
 			$result = false;
 			if (
-				Filesystem::isValidPath($path2)
-				and Filesystem::isValidPath($path1)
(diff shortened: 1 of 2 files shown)


lib/private/Files/View.php
```diff
@@ -891,20 +891,21 @@ public function rename($path1, $path2) {
...
-				and !Filesystem::isForbiddenFileOrDir($path2)
+				Filesystem::isValidPath($absolutePath2)
+				&& Filesystem::isValidPath($absolutePath1)
+				&& !Filesystem::isForbiddenFileOrDir($absolutePath2)
 			) {
 				$path1 = $this->getRelativePath($absolutePath1);
 				$path2 = $this->getRelativePath($absolutePath2);
 
-				if ($path1 == null or $path2 == null) {
+				if ($path1 === null || $path2 === null) {
```
tomneedham: * shudders *

butonic: is that a +1?

tests/lib/Files/ViewTest.php
```diff
@@ -462,12 +462,16 @@ public function copyBetweenStorages($storage1, $storage2) {
...
-		$rootView->copy('substorage', 'anotherfolder');
+		// paths should not start or end in /
+		$rootView->copy('substorage/', '/anotherfolder');
 		$this->assertArrayHasKey('path', $calledEvent[1]);
 		$this->assertEquals('file.aftercreate', $calledEvent[0]);
 		$this->assertArrayHasKey('oldpath', $calledCopyEvent[1]);
 		$this->assertArrayHasKey('newpath', $calledCopyEvent[1]);
+		// mak sure paths in events have been normalized
 		$this->assertEquals('file.aftercopy', $calledCopyEvent[0]);
+		$this->assertSame('/substorage', $calledCopyEvent[1]['oldpath']);
```
butonic: @PVince81 I now deliberately use paths starting with a `/`. without this PR this fails. There might be other events that do not normalize the path before emitting them in event, but should be checked separately.

butonic:
@PVince81 @sharidas ready to review

PVince81 (approved):
:+1: looks good

phil-davis:
Backport ``stable10`` #31689

PR #31272, opened 2018-04-26
Move caldav tester to drone

phil-davis:

PR #31450, opened 2018-05-16
[stable10] Add caldav test to drone

phil-davis:
Backport #31272 

IMO we should backport this, to avoid having more conflicts in future. Then when we decide the next step of moving more caldav/carddav to drone, maybe that will backport easily.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -136,6 +136,15 @@ pipeline:
       matrix:
         INSTALL_SERVER: true
 
+  setup-caldav-tests:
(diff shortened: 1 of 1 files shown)


phil-davis:
@patrickjahns approve and merge this?
Just to keep ``stable10`` as much in sync with ``master`` as reasonable.

phil-davis:
@DeepDiver1975 or @patrickjahns OK?

DeepDiver1975 (approved):

PR #31452, opened 2018-05-16
[stable10] fix double slash in versioning file copy events

phil-davis:
Backport #31406


PVince81 (approved):
:+1:

:see_no_evil: no unit tests as not testable from legacy static class

Issue #31465, opened 2018-05-17
Missing symfony event for failed login attempt


tomneedham:
Should at least include the attempted `login` value if using credentials.

PVince81:
@sharidas please link to PR as I heard you were working on a solution

sharidas:
PR here: https://github.com/owncloud/core/pull/31467

Issue #31462, opened 2018-05-17
Missing symfony event and share id for public share link access


tomneedham:
https://github.com/owncloud/core/blob/6b36c56180eee3ce796b1a69709431f1d38da627/apps/files_sharing/lib/Controllers/ShareController.php#L218

And this is also missing the share ID which is very helpful when listening to this event.

PVince81:
PR here https://github.com/owncloud/core/pull/31467

PVince81:
done

PR #31467, opened 2018-05-17
Symfony events added for login failed and public link accessed

sharidas:
New symfony events added when:
a) login attempt is failed
b) public link share is acced

## Description

This change comprises of emitting symfony events when :
- A user fails to login
- A public link share is accessed.

## Related Issue

- https://github.com/owncloud/core/issues/31462
- https://github.com/owncloud/core/issues/31465

## Motivation and Context

The login failure can be logged using the event triggered. Which would be helpful for the administrators to know which user tried to login and failed. Public link share event would also help in the same way.

## How Has This Been Tested?

- For the public link share access, tested in the debug mode. When the link is accessed, the control reaches the dispatcher.
- For the login failed tested using debug mode.
- Updated the tests for the same.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/lib/Controllers/ShareController.php
+++ b/apps/files_sharing/lib/Controllers/ShareController.php
@@ -53,6 +53,8 @@ use OCP\IUserManager;
 use OCP\Share;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Template;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * Class ShareController
@@ -79,6 +81,8 @@ class ShareController extends Controller {
 	protected $previewManager;
 	/** @var IRootFolder */
 	protected $rootFolder;
+	/** @var EventDispatcher  */
+	protected $eventDispatcher;
 
 	/**
 	 * @param string $appName
@@ -103,7 +107,8 @@ class ShareController extends Controller {
 								\OCP\Share\IManager $shareManager,
 								ISession $session,
 								IPreview $previewManager,
-								IRootFolder $rootFolder) {
+								IRootFolder $rootFolder,
+								EventDispatcher $eventDispatcher) {
 		parent::__construct($appName, $request);
 
 		$this->config = $config;
@@ -115,6 +120,7 @@ class ShareController extends Controller {
 		$this->session = $session;
 		$this->previewManager = $previewManager;
 		$this->rootFolder = $rootFolder;
+		$this->eventDispatcher = $eventDispatcher;
 	}
 
 	/**
@@ -223,6 +229,16 @@ class ShareController extends Controller {
 			'errorCode' => $errorCode,
 			'errorMessage' => $errorMessage,
 		]);
+
+		if ($share instanceof \OCP\Share\IShare) {
+			$cloneShare = clone $share;
+			$publicShareLinkAccessEvent = new GenericEvent(null,
+				['shareObject' => $cloneShare, 'errorCode' => $errorCode,
+					'errorMessage' => $errorMessage]);
+
+			$this->eventDispatcher->dispatch('share.linkaccess', $publicShareLinkAccessEvent);
+		}
+
 		if ($exception !== null) {
 			throw $exception;
 		}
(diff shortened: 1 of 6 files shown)


apps/files_sharing/lib/Controllers/ShareController.php
```diff
@@ -223,6 +229,16 @@ protected function emitAccessShareHook($share, $errorCode = 200, $errorMessage =
 			'errorCode' => $errorCode,
 			'errorMessage' => $errorMessage,
 		]);
+
+		if ($share instanceof \OCP\Share\IShare) {
+			$cloneShare = clone $share;
```
sharidas: If clone is not to be used, let me know I will remove it.

karakayasemi:
As far as I can see we have four different login types in UserSession, and UserSession is already emitting a failedLogin events for two login types. 
https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L494
https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L624

Probably, this PR leads to emit two different failedLogin events for loginWithPassword and loginWithApache. IMHO, we should unify these events. 

In my pov, we can remove event signal emitting logic from UserSession and use this PR, or add failedLogin event for the other missing login types in UserSession.

apps/files_sharing/lib/Controllers/ShareController.php
```diff
@@ -223,6 +229,16 @@ protected function emitAccessShareHook($share, $errorCode = 200, $errorMessage =
...
+			$publiShareLinkAccessEvent = new GenericEvent(null,
+				['shareObject' => $cloneShare, 'errorCode' => $errorCode,
+					'errorMessage' => $errorMessage]);
+
+			$this->eventDispatcher->dispatch('share.linkaccess', $publiShareLinkAccessEvent);
```
PVince81: missing "c"

PVince81:
@sharidas I'd have preferred separate PRs or at least separate commits since the two events are not related at all (or are they?). It's ok to leave it as one now.

apps/files_sharing/tests/Controllers/ShareControllerTest.php
```diff
@@ -258,6 +263,13 @@ public function testAuthenticateInvalidPassword() {
 		$hookListner = $this->getMockBuilder('Dummy')->setMethods(['access'])->getMock();
 		\OCP\Util::connectHook('OCP\Share', 'share_link_access', $hookListner, 'access');
 
+		$calledShareLinkAccess = [];
+		$this->eventDispatcher->addListener('share.linkaccess',
+			function (GenericEvent $event) use (&$calledShareLinkAccess) {
```
butonic: ou are not even using `$calledShareLinkAccess` ... get rid of the use.
also, stop using & everywhere. there are only rare occasions when you need it: http://php.net/manual/en/language.references.whatdo.php

sharidas:
@PVince81 
> did you address https://github.com/owncloud/core/pull/31467#issuecomment-389859926 ?

The issue raised in this comment was because my earlier version of the PR used LoginController.php to dispatch the events. But in this version I have removed the change from LoginController.php to `User/Session.php`. Hence I have addressed the comment by having the dispatchers related to login failures in `User/Session.php`.

apps/files_sharing/tests/Controllers/ShareControllerTest.php
sharidas: `$calledShareLinkAccess` is used from https://github.com/owncloud/core/pull/31467/files#diff-d1ba03f4d58be19f7cb3af8c272e150eR288 to https://github.com/owncloud/core/pull/31467/files#diff-d1ba03f4d58be19f7cb3af8c272e150eR296

If I remove the use of `&` then by the time when test tries to use assertions the `$calledShareLinkAccess` would be empty, just like https://github.com/owncloud/core/pull/31467/files#diff-d1ba03f4d58be19f7cb3af8c272e150eR266

apps/files_sharing/lib/Controllers/ShareController.php
```diff
@@ -223,6 +229,16 @@ protected function emitAccessShareHook($share, $errorCode = 200, $errorMessage =
 			'errorCode' => $errorCode,
 			'errorMessage' => $errorMessage,
 		]);
+
+		if ($share instanceof \OCP\Share\IShare) {
+			$cloneShare = clone $share;
+			$publiShareLinkAccessEvent = new GenericEvent(null,
```
PVince81: s/publi/public/

PVince81 (approved):
Looks fine :+1:

sharidas:
Backport PR: https://github.com/owncloud/core/pull/31584

PR #31482, opened 2018-05-19
Temporarily cache objectstore stat metadata for duration of storage->getMetadata

mrow4a:
## Description
In objectstore, each filesystem attribute needed to be retrieved using separate stat -> this implements more "filesystem" like caching of these properties. 

Now, within the duration of storage->getMetadata, stat will be kept in cache and thus repetitive calls will hit stat cache instead of filecache. 

## Related Issue
This PR is much safer way of avoiding filecache hits then cache of filecache globally - https://github.com/owncloud/core/pull/28166.  

## How Has This Been Tested?
I used diagnostic app

### Old implementation - single put - files_primary_s3 

{"type":"SUMMARY","reqId":"uSvspzacGpeHBg5MWWB6","time":"2018-05-19T23:22:33+00:00","remoteAddr":"::1","user":"admin","method":"PUT","url":"\/octest\/remote.php\/dav\/files\/admin\/testnew.txt","diagnostics":{"**totalSQLQueries":176**,"totalSQLDurationmsec":20.77317237854004,"totalSQLParams":405,"totalEvents":16,"totalEventsDurationmsec":61.043500900268555}}

### New implementation - single put - files_primary_s3 

{"type":"SUMMARY","reqId":"N3I4Dz8W75GB6yScxrCi","time":"2018-05-19T23:23:13+00:00","remoteAddr":"::1","user":"admin","method":"PUT","url":"\/octest\/remote.php\/dav\/files\/admin\/testnew.txt","diagnostics":{**"totalSQLQueries":136**,"totalSQLDurationmsec":16.56365394592285,"totalSQLParams":318,"totalEvents":16,"totalEventsDurationmsec":45.438289642333984}}

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

@butonic @DeepDiver1975 @PVince81 @patrickjahns

PR #31486, opened 2018-05-20
Objectstore storage metadata layer handler

mrow4a:
## Description
This PR implements metadata handler for objectstorage file objects. This is objectstore storage filecache handler, as in objectstore implementation filecache is storage source of truth.

## Related Issue
- Obsoletes https://github.com/owncloud/core/pull/31482

- This PR is much safer way of avoiding filecache hits then cache of filecache globally https://github.com/owncloud/core/pull/28166

## Motivation and Context

Metastore handler uses single object entries caching. Objectstore metadata cache can be safely used since in objectstore, filecache is source of truth. However, to ensure high level of filecache -> storage cache consistency, only one object entry is being cached at the time, and any other call than ->get clears the cache. This sort of caching ensures that repetitive calls to filecache for the same object are consistent across application logic.

## How Has This Been Tested?
Diagnostic app with new file puts

#### "master" branch with files_primary_s3 implementation

New & edit single file PUT

{"type":"SUMMARY","reqId":"65HpLqGkulVX9tsLkZoX","time":"2018-05-20T12:53:35+00:00","remoteAddr":"::1","user":"admin","method":"PUT","url":"\/octest\/remote.php\/dav\/files\/admin\/testnew.txt","diagnostics":{**"totalSQLQueries":177**,"totalSQLDurationmsec":20.37215232849121,"totalSQLParams":400,"totalEvents":16,"totalEventsDurationmsec":52.40273475646973}}

{"type":"SUMMARY","reqId":"4ShcmHORhPgT6GKWdHY5","time":"2018-05-20T12:53:35+00:00","remoteAddr":"::1","user":"admin","method":"PUT","url":"\/octest\/remote.php\/dav\/files\/admin\/testnew.txt","diagnostics":{**"totalSQLQueries":177,**"totalSQLDurationmsec":19.32239532470703,"totalSQLParams":407,"totalEvents":16,"totalEventsDurationmsec":63.67301940917969}}

#### "objectstore metalayer" branch with files_primary_s3 implementation

New & edit single file PUT

{"type":"SUMMARY","reqId":"yEkjg1qNcTeAbkr1r9Mu","time":"2018-05-20T12:52:38+00:00","remoteAddr":"::1","user":"admin","method":"PUT","url":"\/octest\/remote.php\/dav\/files\/admin\/testnew.txt","diagnostics":{**"totalSQLQueries":136**,"totalSQLDurationmsec":17.536163330078125,"totalSQLParams":318,"totalEvents":16,"totalEventsDurationmsec":44.013261795043945}}

{"type":"SUMMARY","reqId":"vGxuK4coHDR5KWz21wes","time":"2018-05-20T12:52:38+00:00","remoteAddr":"::1","user":"admin","method":"PUT","url":"\/octest\/remote.php\/dav\/files\/admin\/testnew.txt","diagnostics":{**"totalSQLQueries":119**,"totalSQLDurationmsec":16.158103942871094,"totalSQLParams":291,"totalEvents":16,"totalEventsDurationmsec":49.94559288024902}}

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreMetadata.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreMetadata.php
@@ -1,0 +1,210 @@
+<?php
+/**
+ * @author Piotr Mrowczynski <piotr@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Files\ObjectStore;
+
+use OCP\Files\Cache\ICache;
+
+/**
+ * Metadata cache for objectstorage file objects.
+ *
+ * ObjectStore uses oc_filecache as storage metadata representation, thus
+ * this class serves as filecache (storage metadata caching layer).
+ */
+class ObjectStoreMetadata extends \OC\Files\Cache\Cache {
+
+	/**
+	 * @@var \OC\Files\Cache\CacheEntry
+	 */
+	private $fileCacheEntry = null;
+
+	/**
+	 * @param \OC\Files\Storage\Storage|string $storage
+	 */
+	public function __construct($storage) {
+		parent::__construct($storage);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function get($file) {
+		$key = $this->getPathKey($this->getNumericStorageId(), $file);
+		if ($this->isFileCacheMatch($key)) {
+			return $this->fileCacheEntry;
+		}
+
+		// Cache only single objects metadata
+		$data = parent::get($file);
+		if ($data && $data->getMimeType() !== 'httpd/unix-directory') {
+			$this->fileCacheEntry = $data;
+		}
+
+		return $data;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getId($file) {
+		$key = $this->getPathKey($this->getNumericStorageId(), $file);
+		if ($this->isFileCacheMatch($key)) {
+			return $this->fileCacheEntry->getId();
+		}
+		return parent::getId($file);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getPathById($id) {
+		$key = $this->getIdKey($this->getNumericStorageId(), $id);
+		if ($this->isFileCacheMatch($key)) {
+			return $this->fileCacheEntry->getPath();
+		}
+		return parent::getPathById($id);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getAll() {
+		return parent::getAll();
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getFolderContentsById($fileId) {
+		return parent::getFolderContentsById($fileId);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function search($pattern) {
+		return parent::search($pattern);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function searchByMime($mimetype) {
+		return parent::searchByMime($mimetype);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function searchByTag($tag, $userId) {
+		return parent::searchByTag($tag, $userId);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function insert($file, array $data) {
+		$this->invalidateFileCache();
+		return parent::insert($file, $data);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function update($id, array $data) {
+		$this->invalidateFileCache();
+		parent::update($id, $data);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function remove($file) {
+		$this->invalidateFileCache();
+		parent::remove($file);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function move($source, $target) {
+		$this->invalidateFileCache();
+		parent::move($source, $target);
+	}
+
+	/**
+	 * @inheritdoc
+	 */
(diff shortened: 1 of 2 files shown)


mrow4a:
FYI @jvillafanez @butonic @DeepDiver1975 @PVince81

mrow4a:
Hmm, @patrickjahns the only failing thing is php-fixer.

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function get($file) {
+		$key = $this->getNumericStorageId().'-get-'.$file;
+		if ($this->metaDataCache->hasKey($key)) {
+			$metadata = $this->metaDataCache->get($key);
+			return new CacheEntry($metadata);
```
jvillafanez: Why don't we store the `CacheEntry` in order to avoid creating a new object here?

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
```diff
@@ -0,0 +1,182 @@
...
+			$this->metaDataCache->set($key, $metadata);
+		}
+		return $data;
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getId($file) {
+		$this->metaDataCache->clear();
```
jvillafanez: This feels kind of harsh... Is it possible to use the cached data? I'd prefer to add 1 or 2 private methods to fetch the data from the cache and then from the DB to increase the reusability. (The same for all the methods)

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -489,4 +489,14 @@ public function restoreVersion($internalPath, $versionId) {
 		}
 		return parent::restoreVersion($internalPath, $versionId);
 	}
+
+	public function getCache($path = '', $storage = null) {
+		if (!$storage) {
+			$storage = $this;
+		}
+		if ($storage && !isset($storage->cache)) {
+			$storage->cache = new ObjectStoreMetadata($storage);
```
jvillafanez: I know you've just copy & paste the code, but this is awful. You can potentially set an `ObjectStoreMetadata` to any storage from this object, and you can end up with a "normal" cache here, which makes absolutely no sense.

@DeepDiver1975 Any chance to deprecate / remove this API and provide a better one? I think the easiest and least intrusive change is changing the implementation as shown below, but there is a cyclic dependency which is quite bad. (The same for the `getScanner`, `getWatcher`, and related methods)
```
public function getCache($path = '', $storage = null) {
  if ($storage === null) {
    $this->cache = new ObjectStoreMetadata($this);
    return $this->cache;
  } else {
    return $storage->getCache($path);
  }
}
```

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
```diff
@@ -0,0 +1,182 @@
...
+	/**
+	 * @param \OC\Files\Storage\Storage|string $storage
+	 */
+	public function __construct($storage) {
+		// Objectstore metadata cache can be safely used since in objectstore filecache
+		// is source of truth. However, to ensure high level of filecache -> storage cache consistency,
+		// only one object entry is being cached at the time.
+		// This sort of caching ensures that repetitive calls to filecache
+		// for the same object are consistent across application logic
+		if ($this->metaDataCache === null) {
```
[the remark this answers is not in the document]
mrow4a: Hm, yes, thanks!

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
```diff
@@ -0,0 +1,182 @@
...
+			$this->metaDataCache = new CappedMemoryCache(1);
```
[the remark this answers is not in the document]
mrow4a: Ye, to be honest I left it with CappedMemoryCache just to keep a discussion on whether other `files` shall be cached too. 

But generally yes, it could be variable

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
mrow4a: To make full clone of the object, just to be sure nothing will modify that cache elsewhere

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
mrow4a: The idea was to only cache repetitive calls to the same "cache function e.g. get". Thus any other call should clear cache. But yes, generally all "getid" could use cache and not clear.

lib/private/Files/ObjectStore/ObjectStoreStorage.php
mrow4a: Yes, this was quite confusing for me too.

lib/private/Files/ObjectStore/ObjectStoreMetadata.php
jvillafanez: I think it will perform just a shallow clone. In case we have references to objects in the `$metadata`, those references will be kept.
If a shallow clone is fine, I'd use `clone`. In any case, we should document that this method returns a shallow clone of the metadata. This should help both to know that you aren't expected to modify the data (or that the changes won't affect to other objects), as well as that you should be careful with references inside the metadata, if any.

DeepDiver1975:
@ownclouders rebase

ownclouders:

DeepDiver1975:
throwing this into smashbox would be great @mrow4a can you take care?
please post results - THX

ownclouders:

mrow4a:
Yes, will do! But also needs an opinion of Jorn @butonic on that (design)

mrow4a:
@butonic @DeepDiver1975 please mind that it makes a lot of sense for Objectstore (only?), since source of truth for objectstorage is oc_filecache itself (thus it is very hard to mess sth up on storage level and not having it reflected in oc_filecache)

butonic:
I understand that you are caching only repeated gets for the same entry and invalidate them whenever anything changes. I still wonder why all the objectstore suite is green, when several tests change oc_filecache without telling your `ObjectStoreMetadata.php`, eg in https://github.com/owncloud/core/pull/28166/files#diff-052d4c915b07e1eae830a2aeaa62cb5cL103 https://github.com/owncloud/core/pull/28166/files#diff-052d4c915b07e1eae830a2aeaa62cb5cL152 https://github.com/owncloud/core/pull/28166/files#diff-052d4c915b07e1eae830a2aeaa62cb5cL203 https://github.com/owncloud/core/pull/28166/files#diff-052d4c915b07e1eae830a2aeaa62cb5cL252 or all the other places in that PR ... 

Personally, I would prefer having this cache inside Cache, so all implementations can benefit from the caching. But again, I am currently puzzled why the tests don't fail. I will need to take a closer look.

mrow4a:
I think because of `public function getCache($path = '', $storage = null)`, in some cases when the cache is already set, you use native FileCache, and not metadata.

mrow4a:
This needs to be blocked until decision is made on temporarily cache some filecache

Issue #31280, opened 2018-04-26
subadmin cannot disable member of his group


paurakhsharma:
This works fine using userprovisioning API but it is the case in webUI
### Steps to reproduce
1. As an `admin` create a new group `newgroup`
2. As an `admin` create new user `newuser`
3. As an `admin`  make the `newuser` subadmin of `newgroup`
4. As an `admin` assign `anotheruser` to `newgroup`
5. As an `newuser` try to `disable` `anotheruser` from `newgroup`

### Expected behaviour
Should be able to disable the user

### Actual behaviour
Cannot disable the user

### Server configuration
**Operating system**: 17.10

**Web server:** Apache

**Database:** MySQL

**PHP version:** 7.1

**ownCloud version:** (see ownCloud admin page) 10.0.8.5

**Where did you install ownCloud from:** git

### Client configuration
**Browser:**

**Operating system:**

#### Browser log
```
Request URL: http://localhost/core/index.php/settings/users/user3/enabled
Request Method: POST
Status Code: 403 Forbidden
```

### Screenshot
![image](https://user-images.githubusercontent.com/20378877/39292007-89649482-4954-11e8-8173-d2af1b60290e.png)

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/31016 (subadmin should be able to remove users from the group), https://github.com/owncloud/core/issues/20282 (Allow non-admin users to see members of their groups), https://github.com/owncloud/core/issues/31279 (Cannot see members of a group having special characters(@,/,&) in their name), https://github.com/owncloud/core/pull/30558 (FIx wording if you are not a member of any groups), and https://github.com/owncloud/core/pull/30551 (Fix wording if you are not a member of any groups).

phil-davis:
IMO this is a missing feature for sub-admins. A sub-admin can delete any of the users in the group that they manage, so it seems logical that they should be able to do the "lesser" task of disabling a user and enabling them again.

phil-davis:
@paurakhsharma let's look at the current UI tests for the user's page and see how we can extend them to cover more of the "obvious" things that should have basic automated smoke tests.

phil-davis:
Fixed in ``core`` ``stable10`` by #31489 
and ``user_management`` by https://github.com/owncloud/user_management/pull/30

Leaving this open for making some tests.

paurakhsharma:
Yes I will look into it after some remaining tests on API sharing

[...]

PR #31489, opened 2018-05-21
Fix php doc so that hasAnnotation works properly

sharidas:
An asterisk was missing in the php doc which was
causing problem when user is enabled/disabled
from UI.

## Description

The call to `setEnabled()` was not happening because of missing asterisk ('*') in the phpdoc. And hence `NotAdminException()` was triggered here: https://github.com/owncloud/core/blob/stable10/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L132, resulting in error.

## Related Issue

https://github.com/owncloud/core/issues/31280

## Motivation and Context

The call to `setEnabled()` was not happening because of missing asterisk ('*') in the phpdoc. And hence `NotAdminException()` was triggered here: https://github.com/owncloud/core/blob/stable10/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php#L132, resulting in error.

## How Has This Been Tested?

- [x] Followed the steps mentioned in the issue.

## Types of changes

## Checklist:

### DIFF
--- a/settings/Controller/UsersController.php
+++ b/settings/Controller/UsersController.php
@@ -787,7 +787,7 @@ class UsersController extends Controller {
 			));
 		}
 		return true;
-    }
+	}
 
 	/**
 	 * @NoAdminRequired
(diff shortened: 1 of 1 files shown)


settings/Controller/UsersController.php
```diff
@@ -869,7 +869,7 @@ public function changeMail($token, $userId) {
 		return new RedirectResponse($this->urlGenerator->linkToRoute('settings.SettingsPage.getPersonal', ['changestatus' => 'success', 'user' => $userId]));
   }
   
-  /*
+  /**
```
phil-davis: While we are here, the indent of line 870 and this line 872 can be fixed.

phil-davis:
Note: needs the same fix in:
https://github.com/owncloud/user_management/blob/master/lib/Controller/UsersController.php#L870
where this code happens in the separate ``user_management`` app for ``core`` ``master`.

sharidas:
@phil-davis Yes I will create the PR for user-management. But the failure which was troubling enable/disable by subadmin doesn't affect there.

phil-davis:
At line 787 there is also missing indent. It would be good if you can quickly review the indent layout of this file and see if there is anything else that could be tidied.

sharidas:
Created PR in user_management : https://github.com/owncloud/user_management/pull/30

phil-davis (approved):
Works, and the indent looks neater also.

phil-davis:
Luv that ``codecov`` - a little change like this takes 16% off the project-wide code coverage!

Issue #31492, opened 2018-05-22
Capability "Enforce Password" with incorrect value 


jesmrec:
The capability `Enforce password` was splitted up in three different capabilities in 10.0.8

<img width="434" alt="screen shot 2018-05-22 at 10 24 22" src="https://user-images.githubusercontent.com/14894746/40350514-55648150-5daa-11e8-8a22-b4f25bbae144.png">

But the capabilities retrieved to clients through `/ocs/v1.php/cloud/capabilities` in json returns always

> "password": {
              "enforced": false
     }, 

Therefore:

- The client is being retrieved only one capability, when there are three different ones supported
- The value in such capability is always `false` regardless the checkboxes selected in admin bashboard

Core: `{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.0.8.5","versionstring":"10.0.8","edition":"Enterprise","productname":"ownCloud"}`

phil-davis:
Yes, that is all missing. I was just running the API acceptance tests for capabilities, so I will have a look...

jvillafanez:
@phil-davis I guess we have duplicated the efforts.... Check my PR, make the adjusments in yours and close mine.

phil-davis:
Not much duplicated effort - the actual code changes are 5 minutes. And we each worked on tests in a different part of the test pyramid. It will merge nicely.

phil-davis:
@jesmrec core PR has been merged and backported some time ago. It should be in a core release these days.
Close?

jesmrec:
Close

PR #31494, opened 2018-05-22
Split public password enforced capabilities into read-only, read_writ…

phil-davis:
…e and write_only

## Description
Report the 3 different capabilities for public link shares:
```
[public][enforced_for][read_only]
[public][enforced_for][read_write]
[public][enforced_for][upload_only]
```

If all are enforced then set the "old" capability:
```
[public][enforced]
```
so that an "old" client will at least know when public link passwords are always enforced.

## Related Issue
#31492 

## Motivation and Context
Clients need to be able to know the detailed settings for these, so they can be nice to users.

## How Has This Been Tested?
Run the new/changed acceptance tests.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
It fixes a "bug" because this information should have been exposed by the original PR.
It is a "feature" because it is now changing the "API" for getting these capabilities.

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

If this is the way to go, then capabilities doc will need updating.

PR #31496, opened 2018-05-22
Include new capabilities for the password enforcement for sharing

jvillafanez:
## Description
Files sharing capabilities will include these new options. The previous "password enforced" capability (which wasn't working properly in 10.0.8) will indicate if any of the options are active. This should keep backwards compatiblity because the clients should enforce the password for all the options, which is more restrictive but should be less problematic for old clients.

Note that the internal sharing API remains unchanged, particularly https://github.com/owncloud/core/blob/master/lib/public/Share/IManager.php#L239

## Related Issue
https://github.com/owncloud/core/issues/31492

## Motivation and Context
Broken capability in 10.0.8

## How Has This Been Tested?
Unittest as well as checking "/ocs/v1.php/cloud/capabilities" endpoint

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/lib/Capabilities.php
+++ b/apps/files_sharing/lib/Capabilities.php
@@ -70,7 +70,14 @@ class Capabilities implements ICapability {
 			$public['enabled'] = $this->config->getAppValue('core', 'shareapi_allow_links', 'yes') === 'yes';
 			if ($public['enabled']) {
 				$public['password'] = [];
-				$public['password']['enforced'] = ($this->config->getAppValue('core', 'shareapi_enforce_links_password', 'no') === 'yes');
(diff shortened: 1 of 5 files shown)


phil-davis:
Also see #31494 which has API acceptance test code and scenarios. Grab those, then I can close that PR.

jvillafanez:
This should be mostly ready. I feel we should add some acceptance test for the case where the password isn't enforced, but I'm not sure if it's already covered by the rest of the scenarios.

phil-davis:
It looks like you got all the code from mine, and I got the code from yours. So both were equivalent.

phil-davis:
No need for backport label on this PR that was closed.

PR #31510, opened 2018-05-23
While decrypting userkey prepareEncryptionModules() must be called, keep fileid on decrypt

sharidas:
Regression was caused because of omitting call to method
prepareEncryptionModules. This change will get the userkey encryption
work as before.

## Description

Call to `prepareEncryptionModules()` was omitted and the check for `userkeys` also got missing. This resulted in regression. And hence encryption with userkey was not decrypted properly. This change would help to fix the regression.

## Motivation and Context

Call to `prepareEncryptionModules()` was omitted and the check for `userkeys` also got missing. This resulted in regression. And hence encryption with userkey was not decrypted properly. This change would help to fix the regression.

## How Has This Been Tested?

- [x] Enable userkey encryption
- [x] execute `./occ encryption:decrypt-all`
- [x] The files should be decrypted.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Encryption/DecryptAll.php
+++ b/lib/private/Encryption/DecryptAll.php
@@ -165,6 +165,11 @@ class DecryptAll {
 				return false;
 			}
 			$this->userManager->callForSeenUsers(function (IUser $user) use ($progress, &$userNo, $numberOfUsers) {
+				if (\OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '0') !== 0) {
+					if ($this->prepareEncryptionModules($user->getUID()) === false) {
+						return false;
+					}
+				}
 				$this->decryptUsersFiles(
 					$user->getUID(),
 					$progress,
--- a/lib/private/Encryption/DecryptAll.php
+++ b/lib/private/Encryption/DecryptAll.php
@@ -59,6 +59,8 @@ class DecryptAll {
 	protected $logger;
 
 	/**
+	 * DecryptAll constructor.
+	 *
 	 * @param Manager $encryptionManager
 	 * @param IUserManager $userManager
 	 * @param View $rootView
@@ -165,7 +167,7 @@ class DecryptAll {
 				return false;
 			}
 			$this->userManager->callForSeenUsers(function (IUser $user) use ($progress, &$userNo, $numberOfUsers) {
-				if (\OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '0') !== 0) {
+				if (\OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '0') !== '0') {
 					if ($this->prepareEncryptionModules($user->getUID()) === false) {
 						return false;
 					}
@@ -250,7 +252,7 @@ class DecryptAll {
 	 */
 	protected function decryptFile($path) {
 		$source = $path;
-		$target = $path . '.decrypted.' . $this->getTimestamp();
+		$target = $path . '.decrypted.' . $this->getTimestamp() . '.part';
 
 		try {
 			\OC\Files\Storage\Wrapper\Encryption::setDisableWriteEncryption(true);
--- a/tests/lib/Encryption/DecryptAllTest.php
+++ b/tests/lib/Encryption/DecryptAllTest.php
@@ -21,23 +21,32 @@
 
 namespace Test\Encryption;
 
+use Doctrine\DBAL\Statement;
 use OC\Encryption\DecryptAll;
 use OC\Encryption\Exceptions\DecryptionFailedException;
 use OC\Encryption\Manager;
 use OC\Files\FileInfo;
 use OC\Files\View;
+use OC\User\AccountMapper;
+use OC\User\AccountTermMapper;
+use OC\User\SyncService;
 use OCA\Files_Sharing\SharedStorage;
+use OCP\DB\QueryBuilder\IExpressionBuilder;
+use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Encryption\IEncryptionModule;
+use OCP\IConfig;
+use OCP\IDBConnection;
 use OCP\ILogger;
+use OCP\IUser;
 use OCP\IUserManager;
 use OCP\UserInterface;
+use OCP\Util\UserSearch;
 use Symfony\Component\Console\Formatter\OutputFormatterInterface;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Test\TestCase;
-use OCP\Files\Storage\IStorage;
 use Test\Traits\UserTrait;
 
 /**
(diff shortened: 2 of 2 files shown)


patrickjahns:
Is it possible to create Tests to catch these regressions?

lib/private/Encryption/DecryptAll.php
```diff
@@ -219,6 +232,25 @@ protected function decryptUsersFiles($uid, ProgressBar $progress, $userCount) {
 							if ($this->decryptFile($path) === false) {
 								$progress->setMessage("decrypt files for user $userCount: $path (already decrypted)");
 								$progress->advance();
+							} else {
+								foreach ([\OCP\Share::SHARE_TYPE_GROUP, \OCP\Share::SHARE_TYPE_USER, \OCP\Share::SHARE_TYPE_LINK, \OCP\Share::SHARE_TYPE_REMOTE] as $shareType) {
```
sharidas: Note: This piece of code is inspired or copy/pasted from transfer-ownership command.

PVince81:
What regression ? What breaks and what works now ?

lib/private/Encryption/DecryptAll.php
```diff
@@ -165,6 +173,11 @@ protected function decryptAllUsersFiles($user = '') {
 				return false;
 			}
 			$this->userManager->callForSeenUsers(function (IUser $user) use ($progress, &$userNo, $numberOfUsers) {
+				if (\OC::$server->getAppConfig()->getValue('encryption', 'userSpecificKey', '0') !== 0) {
```
PVince81: `'0'` is always strictly different from `0`, please be careful with data types

lib/private/Encryption/DecryptAll.php
PVince81: add as comment in PHP code then ?

lib/private/Encryption/DecryptAll.php
```diff
@@ -219,6 +232,25 @@ protected function decryptUsersFiles($uid, ProgressBar $progress, $userCount) {
 							if ($this->decryptFile($path) === false) {
 								$progress->setMessage("decrypt files for user $userCount: $path (already decrypted)");
 								$progress->advance();
+							} else {
+								foreach ([\OCP\Share::SHARE_TYPE_GROUP, \OCP\Share::SHARE_TYPE_USER, \OCP\Share::SHARE_TYPE_LINK, \OCP\Share::SHARE_TYPE_REMOTE] as $shareType) {
+									$offset = 0;
+									while (true) {
```
PVince81: use `do...while` instead ?

lib/private/Encryption/DecryptAll.php
```diff
@@ -219,6 +232,25 @@ protected function decryptUsersFiles($uid, ProgressBar $progress, $userCount) {
...
+										$sharePage = $this->shareManager->getSharesBy($uid, $shareType, null, true, 50, $offset);
+										if (empty($sharePage)) {
+											break;
+										}
+										foreach ($sharePage as $share) {
+											$name = \ltrim($share->getTarget(), '/');
+											if ($name === $file['name']) {
+												$fileInfo = $this->rootView->getFileInfo($path);
+												$share->setNodeId($fileInfo->getId());
+												$this->shareManager->updateShare($share);
```
PVince81: why does the decryption process need to update shares ?

We should skip received shares when decrypting. If the admin wants to decrypt the received share files, the admin must decrypt the share owner's home storage instead.

PVince81 (changes requested):
see comments

sharidas:
> What regression ? What breaks and what works now ?

Regression:
- With oC instance encrypted with user-keys, the decrypt-all command was not working as expected. After decrypt-all command is executed, the files remained encrypted.

What this PR addresses:
- This PR address 2 issues. One is the regression mentioned above. And other fix is the shares become visible after the decrypt-all command is run.

sharidas:
Issue of share lost issue present in versions:
- 9.1.8 -> https://github.com/owncloud/core/commit/7993a510f8ec28002c9fdab1a5783fc567c11d05
- 10.0.0 -> https://github.com/owncloud/core/commit/ce40499f02717afa0af25bf2f67629a3159a29d3
- 10.0.1 -> https://github.com/owncloud/core/commit/1c54fcd45dc4bc242eeb84b56a85da7a276d55a8
- 10.0.2 -> https://github.com/owncloud/core/commit/1b8359fca33d002f10e00b6225100071d486e4b6
- 10.0.3 -> https://github.com/owncloud/core/commit/5ae38448f1f98a5a9a1a629a5094a34bb64550e6

Didn't checked versions previous to 9.1.8 version of oC.

sharidas:
The root cause of the problem is lets say we have a single file `a.txt` in user `admin`. And when user executes decrypt-all, this is how the problem arises:
The problem -> https://github.com/owncloud/core/blob/master/lib/private/Encryption/DecryptAll.php#L252-L254
- copy('a.txt', 'a.txt.decrypted.1234')
- rename('a.txt.decrypted.1234', 'a.txt')
This causes old `a.txt` to be overwritten. But the fileid is changed. Now `a.txt` would have fileid of `a.txt.decrypted.1234`. And that results in loss of meta data.

So a quick ( it sounds horribly weird  :( ) solution that comes to my mind:
- copy(`a.txt`, `a.txt.decrypted.1234`)
- copy(`a.txt.decrypted.1234`, `a.txt`)
- unlink(`a.txt.decrypted.1234`)

PVince81:
@sharidas how about appending ".part" to the decrypted file name ?

this would skip the file cache and then when you rename the part file to overwrite an existing file, it's supposed to keep file id
this is basically how uploads work: the data is stored to a part file and then the part file gets rename+overwrite the original file. the file cache entry stays unchanged apart from size, mtime, etag

lib/private/Encryption/DecryptAll.php
PVince81: reason explained below: because of fileid change. we're researching an alternative solution

PVince81 (commented):
Code looks good, as discussed!

Would be good to add a test that confirms that the file id hasn't changed after decryption, to enforce the fixed behavior.

PVince81:
@sharidas please add your test steps you used to validate that the part file thing works, thanks!

- [ ] unit test to enforce file id
- [ ] list of manual test steps used to confirm that it works

sharidas:
I have updated the PR: https://github.com/owncloud/core/pull/31510 with the '.part' file extension solution.
Below are the tests I did:
- Created users `admin` and `user1`
- Login as `user1` and then logout.
- Login as `admin` and create files `groupshare.txt`, `localshare.txt`, `publiclink.txt`
- Create group `group1`
- Add `user1` to `group1`
- Share `groupshare.txt` to `group1`
- share `localshare.txt` to `user1`
- create public link for `publiclink.txt`
- Finally run the decrypt-all and verified the checks below:
- [x] Verified that `user1` has shares `groupshare.txt` and `localshare.txt`
- [x] Verified that `admin` has files shared -> `publiclink.txt`, `groupshare.txt` and `localshare.txt`

sharidas:
Updated the unit test and manual steps to confirm that it works.

PVince81 (approved):
Code looks good and validated by unit test, as discussed :+1:

PVince81:

sharidas:
Backport PR: https://github.com/owncloud/core/pull/31590

Issue #31509, opened 2018-05-23
loginWithPassword emits wrong symphony events


butonic:
https://github.com/owncloud/core/blob/5f0af496626b957aff38730b5771ec0a33effe31/lib/private/User/Session.php#L490-L516

1. The `user.beforelogin` event has a `uid` that contains the login string (might be an email or a username). Key should be called `login`
2. The `user.afterlogin` event has a `uid` that still contains the login. Should be the userid of the logged in user which wes resolved with checkPassword.
3. `&` is unnecessary in`$this->emittingCall(function () use (&$uid, &$password)`

butonic:
cc @PVince81 @sharidas

sharidas:
Created PR: https://github.com/owncloud/core/pull/31515.

sharidas:
- [x] https://github.com/owncloud/core/pull/31515 merged to master branch
- [x] https://github.com/owncloud/core/pull/31605 merged to stable10

sharidas:
Closing this issue as the changes got merged to stable10 branch and master branch as mentioned in https://github.com/owncloud/core/issues/31509#issuecomment-394603919

PR #31515, opened 2018-05-24
Use the userid obtained from checkPassword after login is validated

sharidas:
The user id sent to the after event should use the id obtained
from checkPassword method, which validates the login credentials
passed to loginWithPassword method. This change would help the
the subscriber of the after event to get the appropriate userid.

## Description

Use the uid returned from `checkPassword()` instead of the uid passed to `loginWithPassword()`. Updated the after event data accordingly to pick the uid.

## Related Issue

https://github.com/owncloud/core/issues/31509

## Motivation and Context

Use the uid returned from `checkPassword()` instead of the uid passed to `loginWithPassword()`. Updated the after event data accordingly to pick the uid.

## How Has This Been Tested?

- [x] With the unit test written for `loginWithPassword()`

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -486,40 +486,50 @@ class Session implements IUserSession, Emitter {
 	/**
 	 * Log an user in via login name and password
 	 *
-	 * @param string $uid
+	 * @param string $login
 	 * @param string $password
 	 * @return boolean
 	 * @throws LoginException if an app canceld the login process or the user is not enabled
+	 *
+	 * Two new keys 'login' in the before event and 'user' in the after event
+	 * are introduced. We should use this keys in future when trying to listen
+	 * the events emitted from this method. We have kept the key 'uid' for
+	 * compatibility.
 	 */
-	private function loginWithPassword($uid, $password) {
-		return $this->emittingCall(function () use (&$uid, &$password) {
-			$this->manager->emit('\OC\User', 'preLogin', [$uid, $password]);
-			$user = $this->manager->checkPassword($uid, $password);
-			if ($user === false) {
-				$this->manager->emit('\OC\User', 'failedLogin', [$uid]);
-				$this->emitFailedLogin($uid);
-				return false;
-			}
+	private function loginWithPassword($login, $password) {
+		$beforeEvent = new GenericEvent(null, ['login' => $login, 'uid' => $login, '_uid' => 'deprecated: please use \'login\', the real uid is not yet known', 'password' => $password]);
+		$this->eventDispatcher->dispatch('user.beforelogin', $beforeEvent);
+		$this->manager->emit('\OC\User', 'preLogin', [$login, $password]);
 
-			if ($user->isEnabled()) {
-				$this->setUser($user);
-				$this->setLoginName($uid);
-				$firstTimeLogin = $user->updateLastLoginTimestamp();
-				$this->manager->emit('\OC\User', 'postLogin', [$user, $password]);
-				if ($this->isLoggedIn()) {
-					$this->prepareUserLogin($firstTimeLogin);
-					return true;
-				} else {
-					// injecting l10n does not work - there is a circular dependency between session and \OCP\L10N\IFactory
-					$message = \OC::$server->getL10N('lib')->t('Login canceled by app');
-					throw new LoginException($message);
-				}
+		$user = $this->manager->checkPassword($login, $password);
+		if ($user === false) {
+			$this->manager->emit('\OC\User', 'failedLogin', [$login]);
+			$this->emitFailedLogin($login);
+			return false;
+		}
+
+		if ($user->isEnabled()) {
+			$this->setUser($user);
+			$this->setLoginName($login);
+			$firstTimeLogin = $user->updateLastLoginTimestamp();
+			$this->manager->emit('\OC\User', 'postLogin', [$user, $password]);
+			if ($this->isLoggedIn()) {
+				$this->prepareUserLogin($firstTimeLogin);
+
+				$afterEvent = new GenericEvent(null, ['user' => $user, 'uid' => $user->getUID(), 'password' => $password]);
+				$this->eventDispatcher->dispatch('user.afterlogin', $afterEvent);
+
+				return true;
 			} else {
 				// injecting l10n does not work - there is a circular dependency between session and \OCP\L10N\IFactory
-				$message = \OC::$server->getL10N('lib')->t('User disabled');
+				$message = \OC::$server->getL10N('lib')->t('Login canceled by app');
 				throw new LoginException($message);
 			}
-		}, ['before' => ['uid' => $uid, 'password' => $password], 'after' => ['uid' => $uid, 'password' => $password]], 'user', 'login');
+		} else {
+			// injecting l10n does not work - there is a circular dependency between session and \OCP\L10N\IFactory
+			$message = \OC::$server->getL10N('lib')->t('User disabled');
+			throw new LoginException($message);
+		}
 	}
 
 	/**
--- a/tests/lib/User/SessionTest.php
+++ b/tests/lib/User/SessionTest.php
@@ -255,7 +255,7 @@ class SessionTest extends TestCase {
 		$userSession->expects($this->once())
 			->method('prepareUserLogin');
 		$calledUserLogin = [];
-		\OC::$server->getEventDispatcher()->addListener('user.afterlogin', function (GenericEvent $event) use (&$calledUserLogin) {
+		$this->eventDispatcher->addListener('user.afterlogin', function (GenericEvent $event) use (&$calledUserLogin) {
 			$calledUserLogin[] = 'user.afterlogin';
 			$calledUserLogin[] = $event;
 		});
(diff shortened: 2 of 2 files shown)


lib/private/User/Session.php
```diff
@@ -487,7 +487,7 @@ public function tryBasicAuthLogin(IRequest $request) {
 	 * @throws LoginException if an app canceld the login process or the user is not enabled
 	 */
 	private function loginWithPassword($uid, $password) {
-		return $this->emittingCall(function () use (&$uid, &$password) {
+		return $this->emittingCall(function (&$afterArray) use (&$uid, &$password) {
```
butonic: try `return $this->emittingCall(function (&$afterArray) use ($uid, $password) {`

lib/private/User/Session.php
```diff
@@ -502,6 +502,7 @@ private function loginWithPassword($uid, $password) {
 				$this->manager->emit('\OC\User', 'postLogin', [$user, $password]);
 				if ($this->isLoggedIn()) {
 					$this->prepareUserLogin($firstTimeLogin);
+					$afterArray['uid'] = $user->getUID();
```
butonic: why not pass in the user object? `$afterArray['user'] = $user;` the old hook does the same thing in https://github.com/owncloud/core/pull/31515/files#diff-9c47cee6ac987e5256aeee509f91ddb1R502

An object is much more useful than the uid as string

tests/lib/User/SessionTest.php
```diff
@@ -1201,4 +1201,53 @@ public function testTryAuthModuleLogin($expectedReturn, array $modules) {
...
+			function (GenericEvent $event) use (&$calledAfterEvent) {
+				$calledAfterEvent[] = 'user.afterlogin';
+				$calledAfterEvent[] = $event;
+			});
+
+		$this->invokePrivate($userSession, 'loginWithPassword', ['user1', 'user1password']);
+
+		$this->assertEquals('user.beforelogin', $calledBeforeEvent[0]);
+		$this->assertInstanceOf(GenericEvent::class, $calledBeforeEvent[1]);
+		$this->assertArrayHasKey('uid', $calledBeforeEvent[1]);
```
butonic: in the before event this is not the uid, but the login (email or username). please rename the key to `login`. `uid` is just plain wrong. changeit in https://github.com/owncloud/core/pull/31515/files#diff-9c47cee6ac987e5256aeee509f91ddb1R517

lib/private/User/Session.php
```diff
@@ -487,7 +487,7 @@ public function tryBasicAuthLogin(IRequest $request) {
 	 * @throws LoginException if an app canceld the login process or the user is not enabled
 	 */
 	private function loginWithPassword($uid, $password) {
```
butonic: so ... $uid here is actually the login (username or email). rename it to `$login` (also update the phpdoc)

butonic (changes requested):
thx for trying to fix this. uid vs login is not clear everywhere. see individual comments

PVince81:
@sharidas comments adressed ? can you finish this ?

PVince81:
@sharidas please also leave the old keys in there for compatibility in case apps use this already.
Mark as deprecated in the code for now and say so in the symfony events documentation (or have the doc only mention the new keys)

- [ ] update symfony doc ticket

PVince81:
@butonic please recheck as you had concerns

lib/private/User/Session.php
```diff
@@ -481,27 +481,34 @@ public function tryBasicAuthLogin(IRequest $request) {
...
 			if ($user->isEnabled()) {
 				$this->setUser($user);
-				$this->setLoginName($uid);
+				$this->setLoginName($login);
 				$firstTimeLogin = $user->updateLastLoginTimestamp();
 				$this->manager->emit('\OC\User', 'postLogin', [$user, $password]);
 				if ($this->isLoggedIn()) {
 					$this->prepareUserLogin($firstTimeLogin);
+					$afterArray['user'] = $user;
+					$afterArray['uid'] = $user;
```
butonic: `$afterArray['uid'] = $user->getUID();` uid is the string ...

lib/private/User/Session.php
```diff
@@ -513,7 +520,7 @@ private function loginWithPassword($uid, $password) {
 				$message = \OC::$server->getL10N('lib')->t('User disabled');
 				throw new LoginException($message);
 			}
-		}, ['before' => ['uid' => $uid, 'password' => $password], 'after' => ['uid' => $uid, 'password' => $password]], 'user', 'login');
+		}, ['before' => ['login' => $login, 'uid' => $login, 'password' => $password], 'after' => ['user' => $login, 'uid' => $login, 'password' => $password]], 'user', 'login');
```
butonic: Just to indicate what needs to go where:
```php
['before' => [
  'login' => $login,
  'uid' => $login, // actually I'd love to deprecate this ... before login we don't know the uid. just here for compatability
  '_uid' => 'deprecated: please use \'login\', the real uid is not yet known' // maybe we can expose the deprecation this way?
  'password' => $password
], 'after' => [
  'login' => $login,
  'uid' => $user->getUid(), // uid = the string
  'user' => $user, // user = the object
  'password' => $password
]]
```

@PVince81 @DeepDiver1975 how do we deprecate uid in the preLogin hook params? We could also use an object that implements `toString()` which logs a warning if someone accesses it?

butonic (changes requested):

lib/private/User/Session.php
```diff
@@ -486,28 +486,35 @@ public function tryBasicAuthLogin(IRequest $request) {
...
+	 * are introduced. We should use this keys in future when trying to listen
+	 * the events emitted from this method. We have kept the key 'uid' for
+	 * compatibility.
 	 */
-	private function loginWithPassword($uid, $password) {
-		return $this->emittingCall(function () use (&$uid, &$password) {
-			$this->manager->emit('\OC\User', 'preLogin', [$uid, $password]);
-			$user = $this->manager->checkPassword($uid, $password);
+	private function loginWithPassword($login, $password) {
+		return $this->emittingCall(function (&$afterArray) use ($login, $password) {
```
jvillafanez: Could we simplify this and make it more standard? This is really confusing as it is.
This should as be as simple as:
```
send before event
do your stuff
send after event
```
You can easily read the function from top to bottom, as well as include extra parameters to the "after event" if needed.

Right now what we have is a big callback function and some parameters that it isn't clear where they're going to go. You have to go to the bottom of the function to know what parameters are going to be sent, and then scroll up to read the real function.
In addition, the `&$afterArray` make things worse: using references inside callbacks is quite dangerous because normally you don't use them, and also the code above must be prepared for it. If 90% of the calls don't use references but a 10% do, the first thing you think is that there might be a bug. Note that just checking the changes I thought it was dead code; I had to check the code of the "emittingCall" function to understand.

My point is that the function is structured in a counter-intuitive way. To understand a function you should just read through it, but this isn't happening with the current code.

sharidas: Updated as per request. Unfortunately haven't updated the event name to camel case. The reason being I suspect there are listeners to the event in other app, like admin_audit.

lib/private/User/Session.php
```diff
@@ -486,40 +486,47 @@ public function tryBasicAuthLogin(IRequest $request) {
...
-			if ($user === false) {
-				$this->manager->emit('\OC\User', 'failedLogin', [$uid]);
-				$this->emitFailedLogin($uid);
-				return false;
-			}
+	private function loginWithPassword($login, $password) {
+		$beforeEvent = new GenericEvent(null, ['login' => $login, 'uid' => $login, '_uid' => 'deprecated: please use \'login\', the real uid is not yet known', 'password' => $password]);
+		$this->eventDispatcher->dispatch('user.beforelogin', $beforeEvent);
+		$this->manager->emit('\OC\User', 'preLogin', [$login, $password]);
+		$user = $this->manager->checkPassword($login, $password);
```
butonic: add a newline before this line to separate event throwing code from the actual password check

lib/private/User/Session.php
```diff
@@ -486,40 +486,47 @@ public function tryBasicAuthLogin(IRequest $request) {
...
-					throw new LoginException($message);
-				}
+		if ($user->isEnabled()) {
+			$this->setUser($user);
+			$this->setLoginName($login);
+			$firstTimeLogin = $user->updateLastLoginTimestamp();
+			$this->manager->emit('\OC\User', 'postLogin', [$user, $password]);
+			if ($this->isLoggedIn()) {
+				$this->prepareUserLogin($firstTimeLogin);
+				$afterEvent = new GenericEvent(null, ['user' => $user, 'uid' => $user->getUID(), 'password' => $password]);
```
butonic: add a newline before this line

lib/private/User/Session.php
```diff
@@ -486,40 +486,47 @@ public function tryBasicAuthLogin(IRequest $request) {
...
+				$this->eventDispatcher->dispatch('user.afterlogin', $afterEvent);
+				return true;
```
butonic: add a newline before this line

butonic (commented):
I like the increase in readability, now add some newlines I I am happy

lib/private/User/Session.php
sharidas: done

lib/private/User/Session.php
sharidas: done

lib/private/User/Session.php
sharidas: done

butonic (approved):

phil-davis:
Backport ``stable10`` #31605

PR #31516, opened 2018-05-24
Acceptance test coding standard changes

phil-davis:
## Description
Refactor long lines and other code style that ``phpcs`` complains about.

## Motivation and Context
Keep the acceptance test code neat.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Acceptance test code style refactoring

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #31517, opened 2018-05-24
[stable10] Acceptance test coding standard changes

phil-davis:
Backport #31516


phil-davis:
Doing this backport had quite a few conflicts, due to the differences of the ``php-cs-fixer`` changes that are in ```master``` and not yet in ``stable10``. I generally selected all  the conflicting code from ``master`` - that gets various bits of those ``php-cs-fixer`` changes along with this backport, which does not hurt anything.

There are 2 less files, because a couple of the changes in ``master`` were edits to new code related to ```objectstore`` stuff, which is not yet in ``stable10``. So, of course, it cannot backport.

individual-it (approved):

PR #31523, opened 2018-05-24
Acceptance test refactoring and bug fixes

phil-davis:
## Description
1) Put in "missing" use statements
2) Fix a call to ``isInTrash``` that had too many/wrong parameters. The test ``elementIsNotInTrashCheckingOriginalPath`` always passed, now it could fail.
3) Fix a wrong ``proppatch`` call in ``changeFavStateOfAnElement`` - I don't think the test was actually broken.
4) Sort out a few inconsistencies in return values and what the docblock said.
5) Remove unused parameters in various functions. Change the corresponding Gherkin regex to not capture the unused things.
6) Enhance a couple of places where the IDE thought that a var might be undefined.
7) Add ``throws`` tags to loads of docblocks.

## Motivation and Context
Cleanup acceptance test code and let the IDE find dodgy-looking stuff.

## How Has This Been Tested?
Local CI running.

## Types of changes
- [x] Acceptance test refactoring and bug fixes

## Checklist:

### DIFF
--- a/tests/acceptance/features/apiMain/tags.feature
+++ b/tests/acceptance/features/apiMain/tags.feature
@@ -185,10 +185,10 @@ Feature: tags
     When user "user0" adds the tag "MyFirstTag" to "/myFileToTag.txt" owned by "user0" using the API
     And user "another_admin" adds the tag "MySecondTag" to "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "201"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "another_admin"
+    And file "/myFileToTag.txt" should have the following tags for user "another_admin"
       | MyFirstTag  | normal           |
       | MySecondTag | not user-visible |
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "user0"
+    And file "/myFileToTag.txt" should have the following tags for user "user0"
       | MyFirstTag | normal |
 
   Scenario: Assigning a not user-assignable tag to a file shared by someone else as admin user should work
@@ -202,10 +202,10 @@ Feature: tags
     When user "user0" adds the tag "MyFirstTag" to "/myFileToTag.txt" shared by "user0" using the API
     And user "another_admin" adds the tag "MySecondTag" to "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "201"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "another_admin"
+    And file "/myFileToTag.txt" should have the following tags for user "another_admin"
       | MyFirstTag  | normal              |
       | MySecondTag | not user-assignable |
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "user0"
+    And file "/myFileToTag.txt" should have the following tags for user "user0"
       | MyFirstTag  | normal              |
       | MySecondTag | not user-assignable |
 
@@ -220,7 +220,7 @@ Feature: tags
     And user "user0" has added the tag "MySecondTag" to "/myFileToTag.txt" owned by "user0"
     When user "user1" removes the tag "MyFirstTag" from "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "204"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "user0"
+    And file "/myFileToTag.txt" should have the following tags for user "user0"
       | MySecondTag | normal |
 
   Scenario: Unassigning a normal tag from a file unshared by someone else as regular user should fail
@@ -233,7 +233,7 @@ Feature: tags
     And user "user0" has added the tag "MySecondTag" to "/myFileToTag.txt" shared by "user0"
     When user "user1" removes the tag "MyFirstTag" from "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "404"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "user0"
+    And file "/myFileToTag.txt" should have the following tags for user "user0"
       | MyFirstTag  | normal |
       | MySecondTag | normal |
 
@@ -251,9 +251,9 @@ Feature: tags
     And user "user0" has added the tag "MySecondTag" to "/myFileToTag.txt" shared by "user0"
     When user "user1" removes the tag "MyFirstTag" from "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "404"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "user0"
+    And file "/myFileToTag.txt" should have the following tags for user "user0"
       | MySecondTag | normal |
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "another_admin"
+    And file "/myFileToTag.txt" should have the following tags for user "another_admin"
       | MyFirstTag  | not user-visible |
       | MySecondTag | normal           |
 
@@ -271,9 +271,9 @@ Feature: tags
     And user "user0" has added the tag "MySecondTag" to "/myFileToTag.txt" shared by "user0"
     When user "another_admin" removes the tag "MyFirstTag" from "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "204"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "user0"
+    And file "/myFileToTag.txt" should have the following tags for user "user0"
       | MySecondTag | normal |
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "another_admin"
+    And file "/myFileToTag.txt" should have the following tags for user "another_admin"
       | MySecondTag | normal |
 
   Scenario: Unassigning a not user-visible tag from a file unshared by someone else should fail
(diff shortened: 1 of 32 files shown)


phil-davis:
``trashbin-new-endpoint.feature:83`` - empty trashbin fails. I will look tomorrow.
``When user :arg1 requests addressbook :arg2 using the API`` - missing - I touched something there.
``Given user :arg1 has successfully created an addressbook named :arg2`` - missing
``dav-versions.feature:35`` - Can't store empty return value. Have a step method return a value. (Exception) - I removed a return value, better put it back.
``dav-versions.feature:42`` - same return value problem

tests/acceptance/features/apiMain/tags.feature
```diff
@@ -185,10 +185,10 @@ Feature: tags
     When user "user0" adds the tag "MyFirstTag" to "/myFileToTag.txt" owned by "user0" using the API
     And user "another_admin" adds the tag "MySecondTag" to "/myFileToTag.txt" shared by "user0" using the API
     Then the HTTP status code should be "201"
-    And file "/myFileToTag.txt" shared by "user0" should have the following tags for "another_admin"
```
phil-davis: Actually, "user0" was not used by this step. It does not matter here which user is the one who shared the file. So I removed that from the step definition.

tests/acceptance/features/apiTrashbin/trashbin-new-endpoint.feature
```diff
@@ -86,7 +86,7 @@ Feature: trashbin-new-endpoint
 		And user "user0" has deleted file "/textfile0.txt"
 		And user "user0" has deleted file "/textfile1.txt"
 		And as "user0" the file "/textfile0.txt" should exist in trash
-		And as "user0" the file "/textfile0.txt" should exist in trash
+		And as "user0" the file "/textfile1.txt" should exist in trash
```
phil-davis: Dumb cut-paste error was here - the test was being doen twice on the same file.

tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -929,12 +929,12 @@ public function statusPhpRespondedShouldMatch(PyStringNode $jsonExpected) {
 			$jsonExpectedDecoded['version'] = \trim($version[1]);
 			$jsonExpectedDecoded['versionstring'] = \trim($versionString[1]);
 			$jsonExpectedEncoded = \json_encode($jsonExpectedDecoded);
+			PHPUnit\Framework\Assert::assertEquals(
```
phil-davis: Moved this up here so the IDE does not shout at me about the vars being maybe undefined.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -243,7 +245,6 @@ public function theseUsersHaveBeenCreated($doNotInitialize, TableNode $table) {
 	 * @return void
 	 */
 	public function theAppHasBeenDisabled($app) {
-		$client = new Client();
```
phil-davis: This was a random unused var.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -320,7 +321,6 @@ private function checkUserDownload($url, $options, $mimeType) {
 			// read everything
 			$buf .= $body->read(8192);
 		}
-		$mimeType = 'text/plain';
```
phil-davis: This was accidentally left hard-coded here when the function was added recently.
Actually ``$mimeType`` is passed in as a parameter.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1078,25 +1073,24 @@ public function getShares($user, $path) {
...
 
-		if ($TableNode instanceof \Behat\Gherkin\Node\TableNode) {
+		if ($TableNode instanceof TableNode) {
 			$elementRows = $TableNode->getRows();
 
 			if ($elementRows[0][0] === '') {
 				//It shouldn't have public shares
 				PHPUnit_Framework_Assert::assertEquals(\count($dataResponded), 0);
-				return 0;
+				return;
```
phil-davis: We just want to return early in this case - no need to return ``0``

tests/acceptance/features/bootstrap/Trashbin.php
```diff
@@ -36,7 +36,7 @@
 	 */
 	public function emptyTrashbin($user) {
 		$body = new \Behat\Gherkin\Node\TableNode(
-			[['allfiles', 'true'], ['dir', '%2F']]
```
phil-davis: Sending an encored ``\`` using ``%2F`` does not work in the body. When the dir is "hidden" inside the body then there is no need to urlEncode anything.
The test previously passed only because ``elementIsNotInTrashCheckingOriginalPath()`` was broken - that ``Then`` gherkin step would always pass, even though the trash  bin had not actually been emptied.

tests/acceptance/features/bootstrap/Trashbin.php
```diff
@@ -225,19 +222,18 @@ public function elementIsInTrashCheckingOriginalPath(
...
 	 * @param string $originalPath
 	 *
 	 * @return void
 	 */
 	public function elementIsNotInTrashCheckingOriginalPath(
-		$user, $entryText, $originalPath
+		$user, $originalPath
 	) {
 		PHPUnit_Framework_Assert::assertFalse(
-			$this->isInTrash($user, $entryText, $originalPath),
```
phil-davis: This was the broken line - ``isInTrash()`` actually takes 2 parameters. Somehow passing it 3 parameters did not make PHP explode. So it was passing values like ``file`` ```folder`` in the position where ``originalPath`` should go.

``false``` would always be returned, because the trashbin indeed did not have a file called ```file`` or a file called ``folder``

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -588,13 +588,11 @@ public function userHasSetPropertyOfEntryTo(
...
 	 * @param string $propertyName
 	 * @param string $propertyValue
-	 * @param null $table unused
-	 *
 	 * @return void
-	 * @throws Exception
+	 * @throws \Exception
 	 */
 	public function theResponseShouldContainACustomPropertyWithValue(
-		$propertyName, $propertyValue, $table=null
```
phil-davis: Goodness knows what ``$table``  was doing here???

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -701,7 +699,7 @@ public function theSingleResponseShouldContainAPropertyWithValueAndAlternative(
 
 		if ($expectedValue === "a_comment_url") {
 			if (\preg_match("#^/remote.php/dav/comments/files/([0-9]+)$#", $value)) {
-				return 0;
+				return;
```
phil-davis: We just want to get out early here, no need to return ``0```

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -743,9 +741,7 @@ public function theSingleResponseShouldContainAPropertyWithValueLike(
 			}
 		}
 
-		if (\preg_match($regex, $value)) {
-			return 0;
```
phil-davis: No need to ``return`` at all here. We can just do the reverse test, and then we are at the end of the function anyway.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1357,6 +1354,10 @@ public function userUploadsAFileWithContentTo(
 			// 4xx and 5xx responses cause an exception
 			$this->response = $e->getResponse();
 		}
+
+		// Return an invalid file id so that any later step that tries to use it
+		// will fail.
+		return "";
```
phil-davis: Functions can return a value to Behat, which will remember the return value. Then it can be saved away for later, by a "built-in" step like:
```
And we save it into "FILEID"
```
So in this function we want to always return something, which will keep Behat happy. Then some later step in the scenario can fail gracefully when it tries to use an empty file id.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1676,14 +1679,12 @@ public function changeFavStateOfAnElement(
...
-			$properties = [
-				'{http://owncloud.org/ns}favorite' => $favOrUnfav
-			];
-		}
+		$properties = [
+			'{http://owncloud.org/ns}favorite' => $favOrUnfav
+		];
 
 		$response = $client->proppatch(
-			$this->getDavFilesPath($user) . $path, $properties, $folderDepth
```
phil-davis: ``proppatch``  does not take a 3rd parameter. ``$folderDepth`` is irrelevant here. All callers specified ``0`` anyway.

tests/acceptance/features/bootstrap/WebUISharingContext.php
```diff
@@ -593,6 +603,7 @@ public function itShouldNotBePossibleToShareUsingTheWebUI(
 				'could not find sharing button in fileRow',
 				'could not share with \'' . $shareWith . '\''
 			];
+			$foundMessage = false;
```
phil-davis: to keep the IDE happy - it thinks maybe ``$foundMessage`` could be undefined.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1133,12 +1145,12 @@ public function adminMakesUserNotSubadminOfGroupUsingTheAPI($user, $group) {
 	/**
 	 * @Then /^the users returned by the API should be$/
 	 *
-	 * @param \Behat\Gherkin\Node\TableNode|null $usersList
+	 * @param TableNode|null $usersList
```
individual-it: so you can give the function a `null` but then it would do exactly nothing?

individual-it (approved):
just one small comment, we can fix that in an other PR

phil-davis:
Removed ```null`` return from functions that should not have it.
Added a test scenario for getting groups of a user when the user is not in any groups.

phil-davis:
Backport ``stable10``` #31536

Issue #31576, opened 2018-05-29
Capabilites API response does not have shareapi_exclude_groups


phil-davis:
While starting to make acceptance tests related to "exclude groups from sharing", I noticed that the setting of this does not appear in the capabilities API response.
It would be good to have it, so tests can easily see what the current setting is and so (in future) clients can easily know about the setting and adjust their sharing behavior (if they care).

- add the setting of shareapi_exclude_groups to the capabilities API response
- add the list of excluded groups to the capabilities API response

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/3061 (API - failed response BUG), https://github.com/owncloud/core/issues/9304 ([webdav] Etag does not change when disabling share api), https://github.com/owncloud/core/issues/22666 (No WWW-authenticate response header for HTTP PUT API requests), https://github.com/owncloud/core/issues/3215 (OCS API does not allow function closures as actions), and https://github.com/owncloud/core/issues/5724 ([OC6Beta2] Share API Get all shares doesn't work with CURL).

PVince81:
I'm not sure whether this is a good idea as it could be revealing private information. Some users might not be allowed to know about the existence of other groups but having this on capabilities would expose just that.

In general I'd expect all API calls to already comply with this rule so clients wouldn't need to know about the exclusion, so the acceptance tests would be the only case ?

cc @patrickjahns @SamuAlfageme as we talked about capabilities this morning

phil-davis:
If a client wants to be nice, it can query the capabilities when it connects. Then it knows the various sharing things that can/cannot be done (e.g. if public link shares are not enabled, or the whole of sharing is not enabled, or...). The client can then pro-actively switch off bits of its UI, rather than having elements on the UI that are going to result in error messages if the end-user tries to use them.

But yes, for keeping group names as private as possible, it would be better if the client provides their authentication to some end-point which responds with client-behavior capabilities - e.g. ``can_share`` boolean. The client does not need to know exactly why they cannot share (e.g. because sharing is off, or because they are in an excluded group or...)

PVince81:
Yes, the `can_share` boolean approach is better and its value depends on the logged in user.
If more info is needed then its value could be more detailed like "exclusion", "sharing_off" or something. But I'm not sure whether there is any value for clients knowing that.

phil-davis:
[...]

PR #31490, opened 2018-05-21
Exclude users from sharing if they are a member of any excluded group

tomneedham:
## Description
Sharing is disabled, if a user is a member of any of the excluded groups.

## Related Issue
https://github.com/owncloud/enterprise/issues/2508

## Motivation and Context
The existing feature is redundant if the user gets added to another group. 

## How Has This Been Tested?
Manually. 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31580, opened 2018-05-29
Add exclude_groups_from_sharing to capabilities

phil-davis:
## Description
In the capabilities API response, ``files_sharing`` section, add:
```
exclude_groups_from_sharing => boolean state of this setting
groups_excluded_from_sharing => array of groups that are excluded from sharing
```
and only report this if the user is an admin.

In the capabilities API response, ``files_sharing`` section, add:
```
can_share => boolean state of this setting
```
Set to ``true`` if the user requesting the capabilities can do sharing (i.e. the sharing API  is enabled and the user is not excluded from sharing). Otherwise set to ``false```.
This capability could be used by a client app to, for example, decide if the client app should bother to display UI elements for adding shares...

## Related Issue
#31576 

## Motivation and Context
The setting of "exclude groups from sharing and the list of groups that are excluded is not in the capabilities API response.

## How Has This Been Tested?
Local API test runs, and browsing to the capabilities URL.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_sharing/lib/Capabilities.php
+++ b/apps/files_sharing/lib/Capabilities.php
@@ -65,6 +65,19 @@ class Capabilities implements ICapability {
 	}
 
 	/**
+	 * Returns whether the currently logged in user is an administrator
+	 *
+	 * @return bool true if the user is an admin
+	 */
+	private function isAdmin() {
+		$user = $this->userSession->getUser();
+		if ($user !== null) {
+			return $this->groupManager->isAdmin($user->getUID());
+		}
+		return false;
+	}
+
+	/**
 	 * Return this classes capabilities
 	 *
 	 * @return array
@@ -124,6 +137,14 @@ class Capabilities implements ICapability {
 				$res['can_share'] = false;
 			} else {
 				$res['can_share'] = true;
+			}
+
+			if ($this->isAdmin()) {
+				$res['exclude_groups_from_sharing'] = $this->config->getAppValue('core', 'shareapi_exclude_groups', 'yes') === 'yes';
+
+				if ($res['exclude_groups_from_sharing']) {
+					$res['groups_excluded_from_sharing'] = \json_decode($this->config->getAppValue('core', 'shareapi_exclude_groups_list', '[]'), true);
+				}
 			}
 
 			$user_enumeration = [];
--- a/tests/acceptance/features/apiCapabilities/capabilities.feature
+++ b/tests/acceptance/features/apiCapabilities/capabilities.feature
@@ -46,6 +46,7 @@ Feature: capabilities
 			| files_sharing | federation@@@incoming                 | 1                 |
 			| files_sharing | group_sharing                         | 1                 |
 			| files_sharing | share_with_group_members_only         | EMPTY             |
+			| files_sharing | exclude_groups_from_sharing           | EMPTY             |
 			| files_sharing | user_enumeration@@@enabled            | 1                 |
 			| files_sharing | user_enumeration@@@group_members_only | EMPTY             |
 			| files         | bigfilechunking                       | 1                 |
@@ -70,6 +71,7 @@ Feature: capabilities
 			| files_sharing | federation@@@incoming                 | 1                 |
 			| files_sharing | group_sharing                         | 1                 |
 			| files_sharing | share_with_group_members_only         | EMPTY             |
+			| files_sharing | exclude_groups_from_sharing           | EMPTY             |
 			| files_sharing | user_enumeration@@@enabled            | 1                 |
 			| files_sharing | user_enumeration@@@group_members_only | EMPTY             |
 			| files         | bigfilechunking                       | 1                 |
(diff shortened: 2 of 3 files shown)


tests/acceptance/features/apiCapabilities/capabilities.feature
```diff
@@ -439,8 +473,154 @@ Feature: capabilities
...
+			| files         | undelete                                  | 1                 |
+			| files         | versioning                                | 1                 |
+		And the capabilities should not contain
+			| capability    | path_to_element                           |
+			| files_sharing | exclude_groups_from_sharing               |
+			| files_sharing | groups_excluded_from_sharing@@@element[0] |
+			| files_sharing | groups_excluded_from_sharing@@@element[1] |
+			| files_sharing | groups_excluded_from_sharing@@@element[2] |
+
+	@skip @issue-enterprise-2508
```
phil-davis: This scenario should be fixed by #31490 - so when/if this PR is merged, this ``skip`` line can be removed in that PR.

phil-davis:
@PVince81 please have a look and see if I have done reasonable stuff here. I am not familiar with exactly how the DI should be done to get hold of the user and group classes so as to work out if the user is an admin. But the code here does work.

apps/files_sharing/lib/Capabilities.php
```diff
@@ -103,6 +131,20 @@ public function getCapabilities() {
 			$res['share_with_group_members_only'] = $this->config->getAppValue('core', 'shareapi_only_share_with_group_members', 'yes') === 'yes';
 			$res['share_with_membership_groups_only'] = $this->config->getAppValue('core', 'shareapi_only_share_with_membership_groups', 'yes') === 'yes';
 
+			if ($this->isAdmin()) {
+				$res['exclude_groups_from_sharing'] = $this->config->getAppValue('core', 'shareapi_exclude_groups', 'yes') === 'yes';
```
PVince81: in general admins already have access to the config APIs and would be able to retrieve the value from there (which the settings page does), so not sure we really need that in the capabilities for admins

PVince81:
The DI looks ok.

See my comment about the admin part.

apps/files_sharing/lib/Capabilities.php
phil-davis: what to do?
do I look somewhere in "config APIs" to get these settings during acceptance test scenarios?
do I remove the code here?
(which will leave just the calculation of ``can_share`` which was just a bonus when I started this)

PVince81: hmm, the "config APIs" are likely a private endpoint... see https://github.com/owncloud/core/blob/v10.0.8/core/js/config.js#L11

We'll likely need to expose those private APIs using public API format (OCS or Webdav)

PVince81: @phil-davis any reason you need to read any detailed config values during tests ? you could simply set the value to something known so you know it's set during test setup as part of the prerequisite ?

phil-davis: No, we do not have to read back the values. It was just that for other "system/app settings" (at least the ones we used so far), the current setting could be found in capabilities. So I was just maintaining consistency.
We can let the tests make the settings and then the next test steps will observe the changed behavior. If the behavior does not change then either the setting change failed or theere is a bug in the behavior.

PVince81:
I'd rather not expose this setting if there is no need for clients to know this information, considering that it's conditional. So it's not really a capability any more as it depends on the logged in user.

I see you added some utility functions for tests, maybe we could keep those though?

phil-davis:
> rather not expose this setting

which setting do you mean:
``can_share`` - it is conditional, value depends on the logged in user (a capability with-respect-to the current user)
or
``exclude_groups_from_sharing`` - reporting of the "capability" is conditional, depends on the logged in user having admin rights.
?

phil-davis:
I will look through and separate out useful acceptance test stuff related to processing lists in returned xml - that will be good for other stuff in capabilities and... that can have more/better tests.

PVince81:
I meant not exposing "exclude_groups_from_sharing". we can keep "can_share"

phil-davis:
I rebased this to sit on top of the ``can_share`` changes that are now in ``master`` - just so that the ```exclude_groups_from_sharing`` code is still here. Probably will not be used.

PVince81 (approved):
:+1:

PVince81:
backport?

phil-davis:
@PVince81 the ```can_share`` capability was separated into a different PR #31816 which is merged and good.

This PR here is the remaining code, which reports the ``exclude_groups_from_sharing`` setting in the capabilities, only if the caller has admin privs. If the Boolean setting is on, then it also reports ``groups_excluded_from_sharing``

I didn't think that you wanted to add this stuff to the capabilities report?

But I am very happy to have it available - it will make tests flow easily.

PVince81:
I didn't. When I saw the PR I thought it was the updated version where this one was removed already without looking more closely.

Reverting...

PVince81:
revert PR here https://github.com/owncloud/core/pull/31937

phil-davis:
This part is not being implemented. Backport request removed.

PR #31586, opened 2018-05-30
[test] [DO NOT MERGE] Demonstrate rubbish tests pass

phil-davis:
These ``assertContains()`` in ``CapabilitiesTest`` don't do what you think.

The ``assertNotContains()`` also do not do what you think - but it is harder to demonstrate that.


phil-davis:
@PVince81 @DeepDiver1975 @individual-it this sort of crud passes!?!

PVince81:
something is definitely wrong with phpunit or our usage here :scream: 

image

PVince81:
https://stackoverflow.com/questions/31762671/phpunit-why-does-this-assertcontains-pass

PVince81:
ok, so avoid `assertContains` for arrays!

case closed, follow up https://github.com/owncloud/core/pull/31587

PR #31590, opened 2018-05-30
[stable10] While decrypting userkey prepareEncryptionModules() must b…

sharidas:
…e called, keep fileid on decrypt

Below are the 2 causes addressed by this change:
1. Regression was caused because of omitting call to method
prepareEncryptionModules. This change will get the userkey encryption
work as before.

2. The shares were not available in the UI, even if they
were available in oc_share table. The reason for this
was due to the copy and rename operation done in the
decrypt-all operation. And the share table wasn't updated
with the new file. This change updates the share table
with the updated id, which helps the UI to show the share
files.

## Description

Call to prepareEncryptionModules() was omitted and the check for userkeys also got missing. This resulted in regression. And hence encryption with userkey was not decrypted properly. This change would help to fix the regression. Keep the fileid after the decrypt operation.

## Motivation and Context

Call to prepareEncryptionModules() was omitted and the check for userkeys also got missing. This resulted in regression. And hence encryption with userkey was not decrypted properly. This change would help to fix the regression. Keep the fileid after the decrypt operation.

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Encryption/DecryptAll.php
+++ b/lib/private/Encryption/DecryptAll.php
@@ -60,6 +60,8 @@ class DecryptAll {
 	protected $logger;
 
 	/**
+	 * DecryptAll constructor.
+	 *
 	 * @param Manager $encryptionManager
 	 * @param IUserManager $userManager
 	 * @param View $rootView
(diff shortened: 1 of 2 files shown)


PVince81 (approved):
:+1:

PR #31587, opened 2018-05-30
 Fix not-useful capabilities unit tests 

phil-davis:
## Description
Make the Capabilities unit tests assert things about array keys when it is array keys they care about.

## Related Issue
#31588 

## Motivation and Context
Some of the Capabilities unit tests are rubbish - they are checking for values in the array rather than that keys exist. The test assertions do not do what you think they do!

See #31568 for a demonstration of what rubbish can pass.

## How Has This Been Tested?
Run unit tests locally.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Unit test "bugfix"

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
(and pass even better)

PR #31591, opened 2018-05-30
[stable10] Fix not-useful capabilities unit tests

phil-davis:
Backport #31587

### DIFF
--- a/apps/files_sharing/lib/API/Share20OCS.php
+++ b/apps/files_sharing/lib/API/Share20OCS.php
@@ -791,7 +791,7 @@ class Share20OCS {
 		}
 
 		if ($permissions !== null && $share->getShareOwner() !== $this->currentUser->getUID()) {
-			/* Check if this is an incomming share */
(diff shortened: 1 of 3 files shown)


PVince81 (approved):
:+1:

phil-davis:
wot I dun?
```
PHP Fatal error: Cannot use OCP\Files\File as File because the name is already in use in /drone/src/tests/lib/Share20/ManagerTest.php on line 28
```

PVince81:
usually appears on some setups with older PHP versions when having a `use` statement that, for some mysterious reasons, clashes with another use statement. It happened to me locally years ago while on other setups it didn't. Could be related to class loading order...

Maybe restart the build and hope ? :see_no_evil:

phil-davis:
I  think another PR ggot merged to ```stable10`` in the last hours - my ``ManagerTest.php`` did have  duplicate ``use`` statement. Rebase looks like it fixed it - we will see.

PR #31596, opened 2018-05-30
[stable10] Exclude uploads directory from read-only cache mask

IljaN:
## Description
GuestApp "Jail" was too strong resulting in no permissions to write in to uploads directory.

## Related Issue
Closes https://github.com/owncloud/guests/issues/198

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
+++ b/lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
@@ -63,6 +63,8 @@ class ReadOnlyCachePermissionsMask extends CacheWrapper {
 	protected function formatCacheEntry($entry) {
 		$storageId = $entry->getStorageId();
 
+		// Give all permissions to whitelisted "internal" paths and their
+		// subdirectories
 		if ($this->isHomeStorage($storageId)) {
 			foreach ($this->whitelist as $path) {
 				if ($this->startsWith($entry->getPath(), $path)) {
@@ -72,6 +74,7 @@ class ReadOnlyCachePermissionsMask extends CacheWrapper {
 			}
 		}
 
+		// Allow creation of skeleton files
 		if ($this->isHomeStorage($storageId) && $entry->getPath() === '') {
 			$entry['permissions'] = Constants::PERMISSION_CREATE;
 			$this->mask = Constants::PERMISSION_CREATE;
--- a/lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
+++ b/lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
@@ -38,6 +38,16 @@ class ReadOnlyCachePermissionsMask extends CacheWrapper {
 	protected $mask;
 
 	/**
+	 * System internal paths which should not be protected
+	 * @var array
+	 */
+	private $whitelist = [
+		'uploads',
+		'cache',
+		'files_zsync'
+	];
+
+	/**
 	 * @param \OCP\Files\Cache\ICache $cache
 	 * @param int $mask
 	 */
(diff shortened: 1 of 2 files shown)


mmattel:
Is there a particular requirement to first do this in stable10?
Needs forward port to master.

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
```diff
@@ -37,6 +37,16 @@ class ReadOnlyCachePermissionsMask extends CacheWrapper {
 	 */
 	protected $mask;
 
+	/**
+	 * System internal paths which should not be protected
+	 * @var array
+	 */
+	private $whitelist = [
```
IljaN: Check if constants of the corresponding classes can be used

IljaN:
@PVince81 Files removed by guest are appearing in the trashbin of owner

patrickjahns:
Additional things to consider:
- what happens to aborted uploads? will the stray chunks cleaned up afterward?
- will this work with objectstorage?

PVince81:
> what happens to aborted uploads? will the stray chunks cleaned up afterward?

unrelated to guests but also needs to be tested for guest users to confirm that the general mechanism works: `occ dav:cleanup-chunks` in cron to delete stray chunks of improperly aborted uploads.
Properly aborted uploads would issue a DELETE request that deletes the uploads folder, this needs to be tested to in the acceptance tests. @IljaN please add that

PVince81:
> will this work with objectstorage?

automated tests will tell ;-)

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
```diff
@@ -53,13 +63,30 @@ public function __construct($cache, $mask) {
 	protected function formatCacheEntry($entry) {
 		$storageId = $entry->getStorageId();
 
-		if (substr($storageId, 0, strlen('home::')) === 'home::' && $entry->getPath() === "") {
+		if ($this->isHomeStorage($storageId)) {
+			foreach ($this->whitelist as $path) {
```
PVince81: a PHP comment would be nice to explain what we're doing here and why

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
```diff
@@ -53,13 +63,30 @@ public function __construct($cache, $mask) {
...
+				if ($this->startsWith($entry->getPath(), $path)) {
+					$entry['permissions'] = Constants::PERMISSION_ALL;
+					return $entry;
+				}
+			}
+		}
+
+		if ($this->isHomeStorage($storageId) && $entry->getPath() === '') {
```
PVince81: another PHP comment here, is that for skeleton creation ?

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
```diff
@@ -53,13 +63,30 @@ public function __construct($cache, $mask) {
...
 		$entry['permissions'] &= $this->mask;
-
 		return $entry;
 	}
+
+	private function isHomeStorage($storageId) {
+		return \substr($storageId, 0, strlen('home::')) === 'home::';
+	}
+
+	function startsWith($haystack, $needle) {
```
PVince81: hah, PHP doesn't have startsWith... :cry:

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
PVince81: I'm undecided whether we should go the blacklist approach: protect "files", "files_versions" and "files_trashbin" and that's all. We might need the other folders to be able to cache thumbnails (thumbnails aren't shared unfortunately), like "thumbnails" and "gallery". Some third party apps might also create custom sub-directories and might want to store data there. But should they allowed to ?

Encryption keys are anyway stored in the owner's storage so shouldn't be an issue to block "files_encryption".

Maybe whitelist is the safest approach because it protects a lot by default instead of leaving holes. If later on we find out we need to whitelist more we can always add to this list.

PVince81 (commented):
Looks good so far, see comments.

Next step: unit tests and acceptance tests

IljaN:
> @PVince81  I'm undecided whether we should go the blacklist approach
Do you think giving all permissions to everything which does not start with "files" will work?

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
patrickjahns: that whitelist can later on also be made available via `occ config:system` - if there are 3rdparty apps to be run as guest users. (thinking of collabora / onlyoffice or other storage related apps)

lib/private/Files/Cache/Wrapper/ReadOnlyCachePermissionsMask.php
IljaN: Not sure, testing if this is required at all

PVince81 (approved):
That's fine for a minimum viable change.

We should later have a look what folders Collabora and co are using. I'd hope they use the cache folder if they need to cache something.

IljaN:
Forward Port: https://github.com/owncloud/core/pull/31618

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1450,6 +1450,21 @@ public function userMovesNewChunkFileWithIdToMychunkedfile(
 		$this->moveNewDavChunkToFinalFile($user, $id, $dest, []);
 	}
 
+	/**
+	 * @When user :user cancels chunking-upload with id :id using the API
+	 * @Given user :user has canceled new chunking-uploadwith id :id
```
phil-davis: Missing space fixed in #31642 #31643

phil-davis:
Forward port is done and merged. Removing "backport-request" label.

PR #31597, opened 2018-05-31
Proceed encrypt-all by enabling user-keys if no mode is selected by user

sharidas:
Proceed encrypt-all by enabling user-keys, if user
had not selected any encryption mode before executing
the command.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

If user had not selected either user-keys or masterkey encryption modes, then proceed encrypt-all command by enabling user-keys.

## Motivation and Context

If user had not selected either user-keys or masterkey encryption modes, then proceed encrypt-all command by enabling user-keys.

## How Has This Been Tested?

- [x] Create user `admin`
- [x] Enable encryption. Do not select any encryption mode ( this is done from command line )
- [x] Run encrypt-all command
- [x] Verify the db. The user should see 
```
sqlite> select * from oc_appconfig where appid="encryption" and configkey='userSpecificKey';
encryption|userSpecificKey|1
sqlite>
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31598, opened 2018-05-31
Proceed encrypt-all after one of the encryption mode is selected

sharidas:
Proceed encrypt-all command only after user had selected one
of the encryption mode.

## Description

If user had not selected user-keys or masterkey, then stop execution of encrypt-all command. Throw an exception.

## Motivation and Context

If user had not selected user-keys or masterkey, then stop execution of encrypt-all command. Throw an exception.

## How Has This Been Tested?

- [x] Create user `admin`
- [x] Enable encryption. Do not select any encryption mode ( this is done from command line )
- [x] Run encrypt-all command
- [x] Verified that exception is triggered.

## Types of changes

## Checklist:

### DIFF
--- a/core/Command/Encryption/EncryptAll.php
+++ b/core/Command/Encryption/EncryptAll.php
@@ -106,6 +106,12 @@ class EncryptAll extends Command {
 			throw new \Exception('Server side encryption is not enabled');
 		}
 
+		$masterKeyEnabled = $this->config->getAppValue('encryption', 'useMasterKey', '');
+		$userKeyEnabled = $this->config->getAppValue('encryption', 'userSpecificKey', '');
+		if (($masterKeyEnabled === '') && ($userKeyEnabled === '')) {
+			throw new \Exception('Select encryption type masterkey or user-keys to continue.');
+		}
+
 		$output->writeln("\n");
 		$output->writeln('You are about to encrypt all files stored in your ownCloud installation.');
 		$output->writeln('Depending on the number of available files, and their size, this may take quite some time.');
--- a/tests/Core/Command/Encryption/EncryptAllTest.php
+++ b/tests/Core/Command/Encryption/EncryptAllTest.php
@@ -125,4 +125,22 @@ class EncryptAllTest extends TestCase {
 		$this->encryptionModule->expects($this->never())->method('encryptAll');
 		$this->invokePrivate($command, 'execute', [$this->consoleInput, $this->consoleOutput]);
 	}
+
+	/**
+	 * @expectedException \Exception
+	 * @expectedExceptionMessage Select encryption type masterkey or user-keys to continue.
+	 */
+	public function testExecuteExceptionForNoModeSelection() {
+		$command = new EncryptAll($this->encryptionManager, $this->appManager, $this->config, $this->questionHelper);
+
+		$this->encryptionManager->expects($this->once())->method('isEnabled')->willReturn(true);
+		$this->config->expects($this->any())
+			->method('getAppValue')
+			->willReturnMap([
+				['encryption', 'useMasterKey', '', ''],
+				['encryption', 'userSpecificKey', '', '']
+			]);
+
+		$this->invokePrivate($command, 'execute', [$this->consoleInput, $this->consoleOutput]);
+	}
 }


core/Command/Encryption/EncryptAll.php
```diff
@@ -106,6 +106,12 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			throw new \Exception('Server side encryption is not enabled');
 		}
 
+		$masterKeyEnabled = $this->config->getAppValue('encryption', 'useMasterKey', '');
+		$userKeyEnabled = $this->config->getAppValue('encryption', 'userSpecificKey', '');
+		if (($masterKeyEnabled === '') && ($userKeyEnabled === '')) {
```
butonic: Hm, there is `$this->encryptionManager->isEnabled()` a few lines above. That already checks `encryption_enabled` and `useMasterKey`. Maybe it also needs to check `userSpecificKey`.

What is the bug you are trying to solve? It seems that when user based encryption was enabled user individual or master key based encryption should have been selected.

So, to properly enable encryption you now need to 
```
occ app:enable encryption
occ encryption:enable # this is a core commend
occ encryption:select-encryption-type # this is an encryption app command
```
before you can `encryption:encryptall` ... shouldn't `encryption:enable` also set the encryption type? Currently, it neither sets `useMasterKey`, nor `userSpecificKey`. Presumably, because core shouldn't know anything about the inner workings of the encryption app, such as which encryption type is used. The encryption login hook will initialize user individual keys unless `useMasterKey` is set, so implicitly user individual keys are used.

- [ ] in the login hook check if `useMasterKey` or `userSpecificKey` are set. If not set  `userSpecificKey`. That will make user individual keys explicit. do it in a separat PR, as it will change behavior. Currently you can just use user individual keys and then switch to master key based encryption, even if some users use user individual keys.

- [ ] add an issue to add an API to explicitly enable an encryption module. That will allow us to call eg `\OC::$server->getEncryptionManager()->enable()` to tell an encryption that it has been enabled and it con do things like create eg the master key. We might pass input and output but that remains to be discussed in the new issue.

- [ ] the IEncryptionModule API should have an enable(), isEnabled() and disable() method. Maybe add a new interface in a new PR so that existing code can work, but we can call these methods on an app when they are being enabled with `occ app:enable`. Best would be if `enable()` could take options as a parameter, so the app can configure itself. Otherwise configuration is a separate step before an app can be used. encryption, ldap and user_shibboleth all need configuration before they are really _enalbled_.

- [ ] with that api core must not know about config variables used in apps 
https://github.com/owncloud/core/blob/5f0af496626b957aff38730b5771ec0a33effe31/lib/private/Encryption/Manager.php#L91

- [ ] We need to check if the documentation contains all three steps to enable the encryption and make sure `userSpecificKey` is set.

- [ ] maybe related: https://github.com/owncloud/encryption/blob/9f576ab02af0578265134c1e728b6703f14d5a7d/lib/Command/RecreateMasterKey.php#L197-L202
does not delete the `userSpecificKey` config variable. open a new PR if it is not.

- [ ] can you add a new issue to add a proper method to the encryption api to check what encryption is used? checking the config bit by bit is tedious and error prone. There are other places where the same functionality is needed:
https://github.com/owncloud/encryption/blob/407c630e03605cf5050a64d6cf3a2b8ac58c7a7d/lib/Command/SelectEncryptionType.php#L91-L95
https://github.com/owncloud/encryption/blob/407c630e03605cf5050a64d6cf3a2b8ac58c7a7d/appinfo/Migrations/Version20170913113840.php#L38-L41
https://github.com/owncloud/encryption/blob/407c630e03605cf5050a64d6cf3a2b8ac58c7a7d/templates/settings-admin.php#L15-L20
https://github.com/owncloud/encryption/blob/407c630e03605cf5050a64d6cf3a2b8ac58c7a7d/lib/Session.php#L63-L65
https://github.com/owncloud/encryption/blob/4992583c5a1f75067a6ccecb54644d39373af47d/templates/settings-admin.php#L15-L20

Arg what a mess ... why did I even look at this ... in light of all this, this PR is good enough.

butonic (approved):

butonic:
https://github.com/owncloud/core/pull/31597 makes more sense.

PR #1948, opened 2013-02-26
Sort after display name + LDAP server side sorting

blizzz:
LDAP server side sorting added. To take real effect, this PR changes the sorting on the Users page to sort according to the display name instead of the internal/login name. The internal name from LDAP users will be unreadable UUID like AF926DEA-E6B4-49C6-9A14-D2E918C2F9B2 so it does not make sense to sort it after this.

Hope that is okay. Adresses an issue stated in https://github.com/owncloud/core/issues/1609 It does not do natural sort, however.

What do you think @jancborchardt @tanghus @schiesbn @karlitschek others?

Issue #26170, opened 2016-09-21
[UI] User management: Sorting feature


pmaier1:
Category: UI/admin
Quick summary use case: As an admin I want to be able to sort user accounts according to the criteria in the user interface (e.g. last login)
Problem it solves / Why I think this is important: Admins get a better overview on their userbase and have the ability to separate out users by sorting according to the criteria

I'm aware that this should probably be done via OCC commands for larger instances but I think it would be nice to have in the UI.

Core feature? Please share your thoughts on this!

pmaier1:
#26171

PVince81:
Best would be to do the sorting on the user backend, which means that we need a way to pass the sorting information to the user backend. User backend can be local users, LDAP or any others.
This, assuming that the user backend supports sorting...

Another idea is to sort on the client side (either on the PHP side after retrieving users from the user backend or on the JS side), but this is not compatible with pagination. So server-side sorting is likely the best approach.

PVince81:
Actually a duplicate of https://github.com/owncloud/core/issues/12597.

Which one to close ?

pmaier1:
Well, the other issue is about sorting alphabetically. Here we want to sort according to the criteria in the user management view, e.g. group membership, last login. We should add sorting alphabetically here of course but I think this FR is more comprehensive.

PVince81:
Ah, I actually meant https://github.com/owncloud/core/issues/12598

This is all from an earlier on-going task for improving the user management: https://github.com/owncloud/core/issues/12998

pmaier1:
Yup, this is indeed identical :) I don't care which one to close, it should be pursued anyway.

PVince81:
Better idea: first implement the user account table https://github.com/owncloud/core/issues/23558 and then just sort on that. This wouldn't require any interaction with any user backend nor require these to support sorting. We'd just use plain old database sorting on the user account table.

Would set to backlog for now.

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).

PVince81:
there are some challenges on the backend related to natural sort and pagination, we might need to rethink how our APIs work: https://github.com/owncloud/core/issues/7244#issuecomment-85095845

PVince81:
Let's change the approach:

- server is doing the sorting and pagination
[...]

Issue #7244, opened 2014-02-17
Sort user list when using SQLite


PVince81:
### Steps:
0. Setup ownCloud with SQLite
1. Create a few users starting with "a" and "z"
2. Create many users starting with "n" to  make the page scroll
3. Reload the page
4. Scroll down

### Expected result
Users sorted from "a" to "n", then after scrolling the "z" appears

### Actual result
Users "a" to "z" without all the "n" appear initially.
After scrolling, the other "n" users insert themselves into the list

This is because SQLite doesn't sort results.
Needs to be fixed in the backend.

CC @jancborchardt

PVince81:
Still seeing this on master (8906158d3b6c0f0d3e120cccbe036a2d36f3cdbe)

PVince81:
I checked the code, it actually uses ORDER BY.

The reason why it looks weird is simply because the database uses the "regular" sort, but the UI uses "natural sort". Which means that whenever new entries are loaded, there is a chance that they will be inserted in the list instead of just appended.

Also the sort discrepancy is made worse because of our use of LIMIT and OFFSET.

Not sure what can be done here...

For files we do the sorting on the PHP side using a PHP-side natural sort. We could in theory do the same for users, but it will break database-based pagination. Basically we'd need to load ALL users from the database, then sort them with PHP, and then cut the requested page out of the result, which is unefficient and also incompatible with the multi-backend approach.

So... not sure what to do with this ticket.

PVince81:
CC @blizzz

PVince81:
Something to reconsider anyway when working on user sorting: https://github.com/owncloud/core/issues/26170

PR #31599, opened 2018-05-31
[stable10] Sort user list by display name

phil-davis:
## Description
Sort the user list on the webUI by display name (full name)

## Related Issue
https://github.com/owncloud/core/pull/1948#issuecomment-392653714

## Motivation and Context
I think other people have asked for the users in the users list to be sorted by display name.
It is not difficult, but there are gotchas.

## Types of changes

## Checklist:

### DIFF
--- a/settings/js/users/users.js
+++ b/settings/js/users/users.js
@@ -266,8 +266,8 @@ var UserList = {
 		rows.sort(function(a, b) {
 			// FIXME: inefficient way of getting the names,
 			// better use a data attribute
-			a = $(a).find('.name').text();
-			b = $(b).find('.name').text();
+			a = $(a).find('.displayName').text();
+			b = $(b).find('.displayName').text();
 			var firstSort = UserList.preSortSearchString(a, b);
 			if(typeof firstSort !== 'undefined') {
 				return firstSort;


phil-davis:
The issues here are:
1) Users are delivered by the server in case-sensitive order:
- users are actually returned in "batches" of (currently) 200 to the UI. Those are actually retrieved by the ``display_name`` field (case-sensitive), so the first 200 ``Aaron Adrian Anne...`` come in the first batch. More batches come as you scroll down. In later batches ``aaron adrian anne...`` (if there are any) come after ``Zoe``.
- so you might be looking for ``anne`` but you do not find her early in the list.
- you scroll down past ``Zoe``
- now ``anne`` comes, but the JavaScript sort here sorts ``anne`` case-insensitive in among ``Aaron Adrian Anne...`` at the top of the list. I guess you can now scroll back up and see that both ``Anne`` and ``anne`` are listed.

A "solution" to this would be not to sort at all - display in the order the server delivers. Or to make the JavaScript sort here match the server-sorred order.

2) Different admins will want the list sorted in different orders - for that, local sorting by column could be implemented.

phil-davis:
Actually I think this change at least makes things better than the current behavior.
Currently the first 200 users are delivered to the webUI, ordered by case-insensitive display name. Then the webUI "kindly" sorts those into ``uid```  order. That can be very confusing, because the admin will see UIDs with a whole range of values (maybe from a-z). Then when they scroll down, another 200 users are loaded, by display name, and then sorted into ``uid`` order.

phil-davis:
Whatever happens here, it needs to be done to the ``user_management`` app also - for running with ``core`` ``master``

PVince81:
@pmaier1 are we ok with this change of behavior in the user management page ? also see https://github.com/owncloud/core/pull/31599#issuecomment-393440837

pmaier1:
> Users are delivered by the server in case-sensitive order

Having to scroll to "Z" first is not really nice when you look for "a" or don't know that "a" exists, I think. For this I would prefer case-insensitive ordering.
How is the list ordered currently?

> Different admins will want the list sorted in different orders - for that, local sorting by column could be implemented.

Yes, there are FRs for this. I would love to get this. Might still conflict with the pagination.

phil-davis:
The Javascript ordering on the webUI is case-insensitive, and locale-aware (sorts Unicode letters with umlaut, grave, accent etc correctly together. (A and a and variants of "a" are alongside each other...).

The server, unfortunately, delivers batches in case-sensitive order. That is current server behavior. To get the server to deliver in case-insensitive order could be done by adding a ``lower_display_name`` field to the table, and then ``orderBy('lower_display_name')`` - but that is a database migration! And then ideally the server would send batches in the same sort order as the JavaScript ``localeCompare()``` collation order. But that could be Swedish, Finnish, German, English... that all have slightly different alphabetical sorting. And SQL is not going to cope with that! It is a general problem when a server tries to deliver pre-sorted batches of data to a client, and wants to deliver in the same order that the client wants them sorted.

mmattel:
This has been created for stable10 first.
Forward porting intended ?

phil-davis:
"forward port" in this case means applying the same change in the ``user_management``  app.
This code does not exist in ``core`` ``master``

mmattel:
question, pls excuse if stupid, but having the change here and possibly in UM, when the UM app gets activated, doesnt this conflict or will many things now beeing in core moved to UM so core gets cleaned up?

btw: the UM repository needs the nice pope picture too - missing any description...

phil-davis:
In ``core`` ``master`` the User Management page has been separated out in its own app. So whenever the content of ``core`` ``master`` is released (``core`` v10.1 or v11.0 however the version numbering works out depending on the "semver"  content of the release), it will be without any User Management page. The removal of the User Management page is not being backported anywhere. So old branches or ``core`` still have the User Management page in them - e.g. 10.0.* 9.1.* 9.0.*. Any fixxes for the User Management page need to be applied in those branches to get to installations that still run those branches. It might be clearer to understand IF there was still support for 9.0l.* and 9.1.* - in that case we would make small fixes in each of those branches also...

There will be no conflicts reported anywhere. Because the "old code" in 10.0.* and the  "new code" in the user_management app are in completely different repos, they will not "notice"  each other in git. So actually we have to be very careful to "manually" make the correct fixes in both places.

mmattel:
@phil-davis  thanks for the good explanation. Could you create some readme.md text in the user_management repo like you describe it here? (currently blank). That would really help.

phil-davis:
@mmattel https://github.com/owncloud/user_management/pull/34

PVince81:
First: it seems that the linked comment https://github.com/owncloud/core/pull/1948#issuecomment-392653714 is based on a use case for mixed LDAP users and regular users, so the admin needs to scroll down. We could say that in this one use case, the need for "sort by display name or anything meaningful but not those horrible LDAP ids" is more acute.

In a regular usage scenario we should provide both sort options by allowing the admin to click on the columns. There is some value in keeping sorting by user id because often times admins work rather with user ids, especially in situations where display names repeat many many times (see how many Thomas Müller exist in Germany) but email addresses / user ids might be different.

At this point instead of changing this behavior and risking alienating other admins who might rely on the user id, I'd rather have this done properly with clickable column.

That said and as you already noticed, there are some technical issues with sorting currently. I've already documented a few here: https://github.com/owncloud/core/issues/7244 https://github.com/owncloud/core/issues/26170 (reopened)

phil-davis:
The trick is to tell the server how you want stuff sorted, so that the server can give it to you in batches that are in the order you are hoping for.

When you change the sort order on the webUI, the webUI needs to ask the server for data in the new order. e.g. you currently have in userid order students s000001 to s000600 displayed. Now you change the sort order to "display name", "Aaron Aardvark" is userid "s001000" and so is not in your current local list at all, but he should now be first in the list. So  you have to re-fetch the data, unless you know that you have the full list already loaded.

And for "popular" sort orders,  there needs to be an index on the relevant database table(s). And yes, one of my first sort orders to ask for as an admin would be by quota used and/or quota remaining - to quickly find those users who used the most space, and those users who are closest to running out of quota. That will be a different challenge vs database indexes.

PVince81:
@phil-davis making server the authority on sorting here makes sense.

We need to make the server side read from oc_accounts and sort accordingly.

A long time ago before the accounts table it would need to query multiple user backends and then sort them, and then extract whatever page you needed, which was deemed to tricky/hacky. With the account table this might work better. Added comment on https://github.com/owncloud/core/issues/26170#issuecomment-397992752

phil-davis:
The current "easy" sort-by-name server-side with database table index is not quite what people (whoever they are) might want, since it is just a "raw" index on the display-name field. Thus it is not case-insensitive, it might sort letters with grave, accent etc. out of the order that people like/expect. So there will be a bit of thought to put in about how to index the accounts table in a suitable way the both "all databases" can do "internally/efficiently" and that is closest to "what the end-user likes".

PVince81:
@phil-davis the tricky part is having a cross-DB solution that provides natural sort. Last I checked (a few years ago) this wasn't possible without changing the collation.
The workaround would be to sort on PHP level with natural sort, but then you lose pagination consistency...

PVince81:
Closing this PR.

We first need to sort out the server side vs client side sorting. Then provide an option to switch sorting between both columns in the UI.

PVince81:
for the server side issue, proposal here: https://github.com/owncloud/core/issues/26170#issuecomment-397992752

the same ticket can be used to track the "sort by clicked column" idea

PR #24551, opened 2016-05-11
occ commands to manage the job list

DeepDiver1975:
ToDo:
- [x] ~~introduce select for update~~ table lock in query builder
- [x] use ~~select for update to~~ lock selected job
- [x] add state update call to the job queue to mark job as selected by executor
- [x] add occ command to remove jobs from the queue

PR #31617, opened 2018-06-01
Add background:execute occ command for running cron jobs manually

tomneedham:
## Description
Added `occ:background:execute {id} -f` OCC command. 

## Motivation and Context
Debugging cron jobs is really tricky, often involes touching oc_jobs and triggering cron.php yourself. 

This can be used to run the jobs manually, view more debug output from the command, and force running even when the interval has not passed.

## How Has This Been Tested?
Manually, testing out the Activity EmailNotification background job. 

screen shot 2018-06-01 at 15 16 38

## Types of changes

## Checklist:


tomneedham:
Moved over the ConsoleLogger which is a useful utility to create an `ILogger` from an `OutputInterface`.

tomneedham:
Open questions - should we respect the `reversed_at` field - and if so how? There is no API to retrieve it in `IJobList`

butonic:
The command is part of core so we could pull `reversed_at ` from the DB directly if we don't want to change the public API. The latter would be the right way, but it requires adding a new `IJob2`. Maybe that is why abstract classes make more sense for public API ...

- [ ] the log lines are prefixed with the app. maybe start with a timestamp, eg
`2018-06-01T16:52:33.1234.Z core: Running job with id 12 and class OCA\Activity\BackgroundJob\ExpireActivities. Last run 0 and interval 86400`

- [ ] separate PR: I wonder how admins get the `id` of a job. It appears in the log, yes. Maybe a separate `background:search` command that allows you to search by any of the db columns?

- [ ] separate PR: add `--workers n` and option and a `background.jobs.socket` config option. That would allow oc to  offload job execution as an event to the worker queue by sending them via socket ... or better jet use redis as a job queue for non timed jobs: https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/redis.md

- [ ] separate PR: get rid of the wnd console logger and use this once it is merged

mmattel:
This is documentation relevant !

mmattel:
> separate PR: I wonder how admins get the id of a job. It appears in the log, yes. Maybe a separate background:search command that allows you to search by any of the db columns?

Why not using a ``--list`` option here as it is execution relevant. List the job ID´s with their name and other possible / useful data. That would make sense to me from an admin pov.

tomneedham:
Pushed a fix to add the time to the log messages in the console

DeepDiver1975:
are you aware of https://github.com/owncloud/core/pull/24551/files

DeepDiver1975:
> Why not using a --list option here as it is execution relevant. List the job ID´s with their name and other possible / useful data. That would make sense to me from an admin pov.

#24551 adds a status command for this

core/Command/Background/Execute.php
```diff
@@ -0,0 +1,94 @@
...
+	 * @param IJobList $jobList
+	 */
+	public function __construct(IJobList $jobList) {
+		$this->jobList = $jobList;
+		parent::__construct();
+	}
+
+	protected function configure() {
+		$this
+			->setName("background:execute")
```
DeepDiver1975: background:queue:execute - and please use single quotes if no string interpolation is used

core/Command/Background/Execute.php
```diff
@@ -0,0 +1,94 @@
...
+		// Try to find the job
+		$job = $this->jobList->getById($jobId);
+		if($job === null) {
+			$output->writeln("<error>Job not found</error>");
+			return 1;
+		}
+
+		$output->writeln('<info>Found job</info>');
+
+		$start = time();
```
DeepDiver1975: \ missing

please run make test-php-style to see what needs to be changed

mmattel:
@tomneedham pls see my comment in https://github.com/owncloud/core/pull/31630#issuecomment-394362878

mmattel:
Question:

at the code locations for ``$logger->debug`` (2x)
the naming is different compared to the terms used above in the code
``class`` --> ``Job``
``arguments`` --> ``Job Arguments``

Is this by intention ?

mmattel:
From a technical pov I understand that.
Consistency wise and from the optical presentation...

PVince81:
I've fixed the unit tests then squashed.

The problem was mostly that now the id is called "Job ID" and some expectation strings had different casings.

PVince81:
as I see @mmattel's comment was adressed since the argument is now called "Job ID" (I had to fix unit tests to make it work)

once codecov reports with positive, let's merge this...

mmattel:
codecov did not finish since @PVince81 commented three days ago...
maybe something went wrong triggering it?

PVince81:
I'd like @karakayasemi to take over and finish this.

It will be very useful for @owncloud/qa as well for triggering specific cron jobs in tests

PVince81:
just talked with @tomneedham and he said this would only be used for debugging, so we'd need to make sure admins aren't using this in production.

and I'd like to have this for automated testing, if not here then in the testing app.

I feel like we need to first clarify the use case and scope a bit more.

cc @pmaier1

phil-davis:
@tomneedham do you want to be able to use this command on a production system?
(in a controlled manner of course)
If yes, then putting it in the testing app may not be the best thing to do. The testing app has endpoints, for example, to run any occ command, or do other stuff that is not the best idea on a production system.

tomneedham:
Yes in production. We sometimes hit the case of a stuck job and dont want to have to wait for the core cronjob mech to trigger it, say 24 hours later. We need to push it manually, after maybe changing some firewall or mail server config. Without this, there is no way to execute the background job specifically.

phil-davis:
@PVince81  @tomneedham @DeepDiver1975 
PR #31630 is waiting for backport but that is "pending" because of this PR (see comment https://github.com/owncloud/core/pull/31630#issuecomment-396834060 )

And backport of #32623 depends on backport of #31630

Is it likely that this PR will be merged?
Or do we backport other stuff and ignore this for now?
Or?

PVince81:
how much protection do we want to prevent admins to do "the wrong thing", would documentation be enough ?

or should we hide the command from the listing ?

I'd like to move this forward because it is blocking some QA tasks that rely on this command.

@tomneedham

PVince81:
idea from @patrickjahns: add a CLI confirmation prompt with a warning to not use this for production.

that should do

phil-davis:
Suggested way is to do like https://github.com/owncloud/files_primary_s3/blob/master/lib/command/createbucket.php
- prompt to confirm the use of the command, with suitable warning text
- provide `accept-warning` command option to explicitly override seeing the warning and prompt

phil-davis:
I added the warning and `--accept-warning` option. It was too easy.
I guess `background:queue:execute` needs some unit tests?

PVince81:
would be good to add unit tests, yes

tests/Core/Command/Background/Queue/ExecuteTest.php
```diff
@@ -0,0 +1,146 @@
...
+namespace Tests\Core\Command\Background\Queue;
+
+use OC\BackgroundJob\Legacy\RegularJob;
+use OC\BackgroundJob\TimedJob;
+use OC\Core\Command\Background\Queue\Execute;
+use OCP\BackgroundJob\IJobList;
+use Symfony\Component\Console\Tester\CommandTester;
+use Test\TestCase;
+
+class TimedJobForExecuteTest extends TimedJob {
```
jvillafanez: 🚫 
This shouldn't be needed.

tests/Core/Command/Background/Queue/ExecuteTest.php
```diff
@@ -0,0 +1,146 @@
...
+		parent::setUp();
+
+		$this->jobList = $this->createMock(IJobList::class);
+		$this->jobList->expects($this->any())->method('getById')
+			->willReturnCallback(function ($id) {
+				if ($id === '666') {
+					return null;
+				}
+				if ($id === '42') {
+					$job = new TimedJobForExecuteTest();
```
jvillafanez: 🚫 
Any class outside the one being tested should be mocked. While there might be exceptions, I don't think this is one of them.

tests/Core/Command/Background/Queue/ExecuteTest.php
```diff
@@ -0,0 +1,146 @@
...
+	/**
+	 * @dataProvider providesJobIds
+	 * @param $jobId
+	 * @param $expectedOutputs
+	 */
+	public function testCommandWithAcceptWarningOption($jobId, $expectedOutputs) {
+		$input = ['Job ID' => $jobId, '--accept-warning' => null];
+		$this->commandTester->execute($input);
+		$output = $this->commandTester->getDisplay();
+		$this->assertNotContains($this->confirmPromptMessage, $output);
```
jvillafanez: Shouldn't we assert that the job is executed, is set as last job, and the execution time is updated (as well as the command itself is successful) instead of just checking for a text message?
Same for all the tests below.

jvillafanez (dismissed):

jvillafanez:
How much important is the `CommandLogger`? If it isn't critical I think it's better to pull it out for now and just inject an `ILogger` instance in the command to be used instead.
I understand there is interest in logging into the console, but I don't think the current approach is good enough. Using a factory service might be better, but I'm not sure if there are easier alternatives.

tests/Core/Command/Background/Queue/ExecuteTest.php
phil-davis: I "copied" this idea from `tests/lib/BackgorundJob/TimedJobTest.php`
TBH I do not really know all about this and am just trying to get some unit test coverage for this Execute command similar to the Delete and Status commands.

phil-davis:
@tomneedham @PVince81 @jvillafanez 
I added an initial try at unit tests here, in the hope that we could move this forward easily, and that would "release" the other backport PRs for the delete and status commands, so it could all be merged.

But it seems it is not so easy. There are comments above about stuff that I do not know. So I guess someone else will need to be assigned and this and the blocked backports... will have to wait some more time?

PVince81:
as discussed, @jvillafanez will help out. thx

lib/private/BackgroundJob/TimedJob.php
```diff
@@ -50,7 +50,29 @@ public function setInterval($interval) {
...
+				);
+			}
 			parent::execute($jobList, $logger);
+		} else {
+			if ($logger !== null) {
+				$id = $this->getId();
+				$class = \get_class($this);
+				$lastRun = $this->getLastRun();
+				$interval = $this->interval;
+				$diff = $this->interval - (\time() - $this->lastRun);
```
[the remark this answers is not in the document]
jvillafanez: Looks like this can't be changed properly without breaking things. We could make several changes around the job system, but it will also need to touch the actual jobs, which are spread around different apps.

jvillafanez:
Except for https://github.com/owncloud/core/pull/31617#discussion_r272479332 which can't be solved without heavy changes, the rest of the problems have been addressed.

phil-davis (approved):
LGTM. Test cases cover the various combiations of ways to use the execute command.

PVince81 (approved):
:+1:

phil-davis:
@PVince81 please decide if the codecov here is enough.
Last week I had about "51% of diff hit", now @jvillafanez has "54.78% of diff hit (target 65.37%)" which is a little better :)

jvillafanez:
I think we can remove the ConsoleLogger and inject a normal one (the one provided by the server) in the execute command. Unless we need the ConsoleLogger for something in particular, I think this should rise the coverage for the patch (less untested code)

PVince81:
@jvillafanez can you make the proposed change ?

PVince81:
ok let's keep the console logger for now, there must have been a reason for adding it...

mmattel:
Questions:
- Backporting?
- Availability for 10.2 only?

Both related to documentaion, see https://github.com/owncloud/docs/issues/290
Please suggest/sample text to be added in docs.

phil-davis:
I am waiting for #34723 backport to be merged. Then I will make a backport of this. It goes with the other new `background:queue:status` and `background:queue:status` that are merged in `stable10` for `10.2`

mmattel:
The doc PR is made. While writing it, I remembered: https://github.com/owncloud/core/issues/31743 (Adding command background:queue:add)

We can now delete a job (yes we ask if we really want to do so) but nontheless an add command is missing... There are for sure default jobs which are part of the tar shipment. A list could be created and documented in the occ command set to reuse/reference...

PR #26676, opened 2016-11-21
Log format experiments

PVince81:
Make the log looks a bit better when containing exceptions.

To test, do a PROPFIND on any folder and look at the look.
Also tweak the boolean flags inside `Log::processException()`.

The formatting was done by piping the row into `jsonlint -p`

### Log exception on the same level as other fields with string stack trace

Uses `getTraceAsString` for the stack, `$jsonStack=false; $jsonStringStack = false;`;

Unformatted:
```
[...]

PR #31623, opened 2018-06-03
[stable10][feature] provide original exception via logging events

patrickjahns:
## Description
When logging an exception, we add the original exception to the context. 
During the actual logging, the exception is extracted from context and provided in the event as original object so it can be used/handled by event listeners

This allows apps to also provide further information on any incidents 

## Motivation and Context
If just logging the message from ownCloud, we loose features of exception traces with parameters.

Note: 
- any app that accesses the exception, is itself responsible for handling the part of removing sensitive data. 

- the logging format for exceptions catched from sabredav will slightly change apps/dav/lib/Connector/Sabre/ExceptionLoggerPlugin.php

## How Has This Been Tested?
- ✅ unit tests provided
- ✅ demo app for sentry written

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -409,6 +409,11 @@ class Log implements ILogger {
 	 */
 	public function logException($exception, array $context = []) {
 		$context['exception'] =  $exception;
+		$level = Util::ERROR;
+		if (isset($context['level'])) {
+			$level = $context['level'];
+			unset($context['level']);
+		}
 		$exception = [
 			'Exception' => get_class($exception),
 			'Message' => $exception->getMessage(),
@@ -423,6 +428,6 @@ class Log implements ILogger {
 		}
 		$msg = isset($context['message']) ? $context['message'] : 'Exception';
 		$msg .= ': ' . json_encode($exception);
-		$this->error($msg, $context);
+		$this->log($level, $msg, $context);
 	}
 }
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -265,6 +265,12 @@ class Log implements ILogger {
 			$extraFields = $context['extraFields'];
 			unset($context['extraFields']);
 		}
+
+		$exception = null;
+		if (isset($context['exception'])) {
+			$exception = $context['exception'];
+			unset($context['exception']);
+		}
 		array_walk($context, [$this->normalizer, 'format']);
 
 		if (isset($context['app'])) {
@@ -361,6 +367,7 @@ class Log implements ILogger {
 			'formattedMessage' => $formattedMessage,
 			'context' => $context,
 			'extraFields' => $extraFields,
+			'exception' => $exception
 		];
 
 		// note: regardless of log level we let listeners receive messages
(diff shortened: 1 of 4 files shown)


tests/lib/LoggerTest.php
```diff
@@ -280,4 +280,23 @@ function(GenericEvent $event) use (&$afterWriteEvent) {
 		$this->assertEquals($expectedArgs, $beforeWriteEvent->getArguments(), 'before event arguments match');
 		$this->assertEquals($expectedArgs, $afterWriteEvent->getArguments(), 'after event arguments match');
 	}
+
+	public function testOriginalExceptionIsProvidedAsExtragField() {
```
phil-davis: Random ``gField`` (gravity field?)

patrickjahns: gravity distortion avoided - we are still on 🌎 and not yet an interplenetary file sync and share solution ;-)

phil-davis: We should start adding long-latency tests ready for the first federated connections between earrth and moon.

mmattel:
@patrickjahns this PR is created for stable10 first,
forwardporting to master required?
if yes pls add a note so we do not forget 😄

patrickjahns:
for forwarding porting - this would require https://github.com/owncloud/core/pull/31135

PVince81:
ah nice, someone working on a resend of https://github.com/owncloud/core/pull/26676

lib/private/Log.php
```diff
@@ -401,6 +401,10 @@ public function log($level, $message, array $context = []) {
 	 * @since 8.2.0
 	 */
 	public function logException($exception, array $context = []) {
+		$context['extraFields'] = [
+			'originalException' => $exception,
```
PVince81: is this enough or should we split the stack trace into a more fine-grained JSON object, like an array of stack entries ?

PVince81:
@patrickjahns you might want to look at #26676 where I made it work with Sabre exceptions.

PVince81:
@patrickjahns can you post an example output of the exception ? I didn't know `Exception` objects were able to be serialized into JSON

patrickjahns:
@PVince81 
exceptions by default would not be properly converted to json - we need to extract the fields in a handler. 

We can write an exception serializer that will convert / serialize exceptions more cleanly. 

I've adressed the change from using `extraFields` to providing a exception object in the Event.
By default the exception is `null` - we can however also move to the path of not including the arrayKey for exceptions when no exception is present.

It depends very much on your preference

DeepDiver1975 (approved):

PVince81 (approved):
:+1: looks fine

PVince81:
needs https://github.com/owncloud/core/pull/31135 to be finished before forward porting can be done

PVince81:
https://github.com/owncloud/core/pull/31135 was merged

@patrickjahns you should hopefully be able to forward port this PR now ?

phil-davis:
Forward port is PR #31979

PR #31630, opened 2018-06-04
Adding background:queue commands: status and delete

DeepDiver1975:
## Description
Adding two new commands for listing background jobs and deleting individual jobs

## Related Issue
#31617 

## Motivation and Context
Adding a full feature set around occ and background jobs

## How Has This Been Tested?
- manual testing
- unit tests

## Types of changes

## Checklist:

### DIFF
--- a/core/Command/Background/Queue/Delete.php
+++ b/core/Command/Background/Queue/Delete.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Core\Command\Background\Queue;
+
+use OCP\BackgroundJob\IJobList;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Delete extends Command {
+
+	/** @var \OCP\BackgroundJob\IJobList */
+	private $jobList;
+
+	public function __construct(IJobList $jobList) {
+		$this->jobList = $jobList;
+		parent::__construct();
+	}
+
+	protected function configure() {
+		$this
+			->setName('background:queue:delete')
+			->setDescription('Delete a job from the queue')
+			->addArgument('id', InputArgument::REQUIRED, 'id of the job to be deleted');
+	}
+
+	/**
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 * @return int
+	 */
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$id = $input->getArgument('id');
+
+		$job = $this->jobList->getById($id);
+		if ($job === null) {
+			$output->writeln("Job with id <$id> is not known.");
+			return 1;
+		}
+
+		$this->jobList->removeById($id);
+		$output->writeln('Job has been deleted.');
+		return 0;
+	}
+}
--- a/lib/public/BackgroundJob/IJobList.php
+++ b/lib/public/BackgroundJob/IJobList.php
@@ -127,4 +127,19 @@ interface IJobList {
 	 * @since 10.0.0
 	 */
 	public function setExecutionTime($job, $timeTaken);
+
+	/**
+	 * iterate over all jobs in the queue
+	 *
+	 * @return void
+	 * @since 10.0.9
+	 */
+	public function listJobs(\Closure $callback);
+
+	/**
+	 * remove a specific job by id
+	 * @return void
+	 * @since 10.0.9
+	 */
+	public function removeById($id);
 }
--- a/core/Command/Background/Queue/Status.php
+++ b/core/Command/Background/Queue/Status.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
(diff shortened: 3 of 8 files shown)


core/Command/Background/Queue/Delete.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
tomneedham: missing copyright header

core/Command/Background/Queue/Delete.php
```diff
@@ -0,0 +1,45 @@
...
+	 * @param OutputInterface $output
+	 * @return void
+	 */
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$id = $input->getArgument('id');
+
+		$job = $this->jobList->getById($id);
+		if ($job === null) {
+			$output->writeln("Job with id <$id> is not known.");
+			return;
```
tomneedham: return 1 for the error?

core/Command/Background/Queue/Delete.php
```diff
@@ -0,0 +1,45 @@
...
+		}
+
+		$this->jobList->removeById($id);
+		$output->writeln('Job has been deleted.');
```
tomneedham: return 0

lib/public/BackgroundJob/IJobList.php
```diff
@@ -127,4 +127,18 @@ public function setLastRun($job);
 	 * @since 10.0.0
 	 */
 	public function setExecutionTime($job, $timeTaken);
+
+	/**
+	 * iterate over all jobs in the queue
+	 *
+	 * @return void
+	 * @since 10.0.3
```
tomneedham: 10.0.9?

tomneedham (changes requested):

mmattel:
For the sake of documentation :smile:

``background:queue:delete`` ``'id'``
``background:queue:status``  ``'Id', 'Job', 'Last run', 'Arguments'``

mmattel:
@DeepDiver1975 
in command delete, you have ``id``
in command status, you have ``Id``
could you haronize them eg to ``ID`` (or whatever but equal) ?

in command status, you have ``Arguments``, I guess this means ``Job Arguments`` or ?
if yes, what about renaming to make it clear which arguments are meant, else you could think of occ command arguments

@butonic you have in #31617 in code ``TimedJob.php`` line 59 ``id``. Please sync with @DeepDiver1975 and have everywhere the same

@DeepDiver1975 , same codline as above in ``TimedJob.php`` @butonic names a ``class`` here as additional text printed. Either this is missing in your PR in ``status`` or it maybe equal to your ``Arguments`` value. Can you pls sync and harmonize?

tomneedham (approved):

mmattel:
@DeepDiver1975 
Having a ``background:queue:delete`` command.
What happens if I delete all jobs or accidentially one job?
How can I come back to a working state?
What about a ``background:queue:reset`` command which resets all jobs to a state like the system got shipped?

DeepDiver1975:
> in command delete, you have id
> in command status, you have Id

I see no inconsistency here - id in the argument follows the command argument naming convention and the Id in the status command is a human readable table column - all columns are upper case.



> in command status, you have Arguments, I guess this means Job Arguments or ?
> if yes, what about renaming to make it clear which arguments are meant, else you could think of occ command arguments

agreed



> @DeepDiver1975 , same codline as above in TimedJob.php @tomneedham names a class here as additional text printed. Either this is missing in your PR in status or it maybe equal to your Arguments value. Can you pls sync and harmonize?

job arguments hold the class name as well as required arguments for the execution of the job.
the timedjob is already an instanciated job of one specific type.

the information is identical but it's represenations is different - I see no way to harmonize this right away



> What happens if I delete all jobs or accidentially one job?
> How can I come back to a working state?

well - this is what backups are for ;-)
seriously: one better knows what he does when killing a job - we can think about a enqueue commend or something

reset is not the way to go because there are multiple sources how jobs are enqueued - we can only restore some of them. I dislike partial solutions

mmattel:
@DeepDiver1975 @tomneedham 
``delete``: ``->addArgument('id', InputArgument::REQUIRED, 'id of the job to be deleted');``
``status``: ``$t->setHeaders(['Id', 'Job', 'Last run', 'Job Arguments']);``
``Execute``:``->addArgument('jobId', InputArgument::REQUIRED)``

All three lines are printed and the user can see it.
The presentation to the user is different ``id`` vs ``Id`` vs ``jobId``.
Independent for what you decide, the look should be consistent else this is unprofessional.
Same version of how we write it should be taken in all PR´s.

> job arguments hold the class name as well as required arguments for the execution of the job

Could we agree to have the same naming in all PR´s for this too, like ``Job Arguments`` or what ever

> one better knows what he does when killing a job

I would not sign this... I do not know, but a possible solution could be to have a marker in the database for jobs defined by core and you can enable/disable them. All other jobs not beeing part of core do not get/have that marker.

> I dislike partial solutions

Agreed - therefore I take the effort to write my thoughts down 😄

DeepDiver1975:
> Execute:->addArgument('jobId', InputArgument::REQUIRED)

I agree to use the same argument name here.

tomneedham:
Status and delete are useful commands, but only for debugging / analysis, not normal operation. This should be clear in any documentation. 

Likewise with my execute command, it is only really for use in exceptional scenarios where we need to forcefully trigger a job on demand for analysis. It should not be expected that these commands are used in normal operation.

tomneedham:
screen shot 2018-06-05 at 10 34 43

Need to update tests with expected table output

DeepDiver1975:
> Need to update tests with expected table output

addressed

mmattel:
@DeepDiver1975 
being a pain, I know, but id -> ``jobId``
delete: Line 44, 57
status: Line 54

(all changes including the proposal below need table length adoption ...)

@tomneedham I do not know where the right place to note that is so I do here,
In your PR in Job.php in line 59 you define a log output having the name ``class`` and ``arguments``
@DeepDiver1975 in status beginning with line 54 you define the printable output

Looking at both PR´s I see that ``\get_class`` gets called but in job.php it is printed as ``class`` while in status it is printed as ``Job``.
I propose to have on **both** the term ``Job name`` and for ``arguments`` the term ``Job Arguments``

DeepDiver1975:
we can harmonize print out later on ...

DeepDiver1975:
@tomneedham feel free to adjust your PR - THX

PVince81:
@DeepDiver1975 backport to stable10 ?

DeepDiver1975:
we will backport once #31617 is merged and we backport this together as one

phil-davis:
Backport `stable10` is in PR #34783
Passes CI

PVince81:
some objections to be clarified in https://github.com/owncloud/core/pull/31617 first

cc @tomneedham

phil-davis:
I added label only-in-master so I can find and recognise this PR easily.
If/when the backporting is merged:
- [ ] remove label only-in-master
- [ ] remove label backport-request

phil-davis:
> some objections to be clarified in #31617 first
> 
> cc @tomneedham

Ping @tomneedham or anyone for comment about backporting.

Issue #7417, opened 2014-02-26
Email Notification in Language


MTRichards:
Currently, an email share notification is sent in the language of the user sending the share email. This means the recipient of that email may not be able to read the notification.

The system should automatically translate email notifications into the language that the recipient has set in their preferences on the server before sending, so that the recipient can read the email.

srggavrilov:
@PVince81 agree! The 1st thing is obvious, and I'd say, it's a bug, not an enhancement. The 2nd is a question of design and should rely on strong and solid i18n and templates support in core, I believe.

pmaier1:
For internal shares the receiving users' language should be used. For public link notifications, as mentioned above, the most simple solution I can imagine would be an admin setting to specify the language of notifications to users the system does not know. Any other ideas? Effort? @PVince81
Would like to schedule this for the next server iteration.

PVince81:
moving to p3 and planned

PR #31634, opened 2018-06-04
Send share notification in the language of recipient

VicDeo:
Ref https://github.com/owncloud/core/issues/7417


VicDeo:
closing as https://github.com/owncloud/core/pull/34144 does the same

PR #31651, opened 2018-06-05
Introduce persistent and explicit locking of file and folders

DeepDiver1975:
## Description
Introduce persistent locks to owncloud.

since we need to list all locks in sub trees I propose to implement these persistent locks in database only and use a join with the file cache.

structure for oc_persistent_file-locks

Column Name | Type | Description
------------ | ------------- | -------------
id | int | PK and auto increment
file_id | int not null| FK to fileid in oc_file_cache
owner | varchar | displayname of locking user
timeout | int | eol timestamp of the lock
created_at | int | timestamp lock was created
token | varchar(1024) | uuid for webdav - 1024 chars for wopi
scope | int | 1 = exclusive, 2 = shared
depth | int | 0, 1 or infinity

recursive query will look like (needed on DELETE and MOVE - any other request only need lock information about the resource itself)
```sql
select l.*, f.path
from oc_lock l
join oc_file_cache f on l.file_id = f.fileid
where f.path like 'files/user/foo/%'
```

## ToDo in this PR
- [x] implement locking in db provider
- [x] implement a real webdav locking provider

## ToDo in follow up PRs or othre repos
- [x] add locking to wopi app => https://github.com/owncloud/wopi/issues/6
- [x] add file action in web ui to lock, unlock files and folders
- [x] display lock information in web ui
- [ ] coordinate behavior with mobile and desktop client
- [ ] cleanup process of outdated locks
- [x] litmus on shared folder

## Motivation and Context
- required by some integrations - wopi
- required in some use cases - e.g. working on big file over a longer period of time

## Types of changes

## Checklist:


PVince81:
conflicts found

PVince81:
As discussed before, for litmus one idea would be to have a behat scenario step "and Webdav path /XYZ is Webdav compliant" that runs litmus in the background. This way you could pass any path like external storages, shared folders, etc to make sure all Webdav operations work correctly there.

does that make sense or is this too far-fetched ? @patrickjahns @phil-davis

apps/dav/tests/unit/Files/FileLocksBackendTest.php
```diff
@@ -0,0 +1,174 @@
...
+class FileLocksBackendTest extends TestCase {
+
+	/** @var FileLocksBackend */
+	private $plugin;
+	/** @var Tree | \PHPUnit_Framework_MockObject_MockObject */
+	private $tree;
+	/** @var IPersistentLockingStorage | IStorage | \PHPUnit_Framework_MockObject_MockObject */
+	private $storageOfFileToBeLocked;
+
+	public function setUp() {
```
PVince81: also add tests for the `before` part of the plugin, as its logic is non-trivial

apps/files/appinfo/info.xml
```diff
@@ -6,7 +6,7 @@
 	<licence>AGPL</licence>
 	<author>Robin Appelman, Vincent Petry</author>
 	<default_enable/>
-	<version>1.5.1</version>
+	<version>1.5.2</version>
```
PVince81: maybe 1.6.0, that's a big addition like a new feature :-)

apps/files/js/filelockplugin.js
```diff
@@ -0,0 +1,149 @@
...
+
+		},
+
+		/**
+		 * @param fileList
+		 * @private
+		 */
+		_extendFileActions: function(fileList) {
+			var self = this;
+			// fileActions.registerAction({
```
PVince81: please remove these. that could be a separate app

apps/files/js/locktabview.js
```diff
@@ -0,0 +1,89 @@
...
+			events: {
+				'click a.unlock': '_onClickUnlock'
+			},
+
+			_onClickUnlock: function (event) {
+				var self = this;
+				var lockToken = event.target.getAttribute('data-lock-token');
+				var lockUrl = event.target.getAttribute('data-lock-root');
+
+				this.model._filesClient.getClient().request('UNLOCK',
```
PVince81: if this becomes common, we could add `lock(path, token)` and `unlock(path, token)` as direct methods on the Files.Client (and/or js-owncloud-client). I could picture apps like the text editor using this method to lock the file while the editor is still open

apps/files/js/locktabview.js
```diff
@@ -0,0 +1,89 @@
...
+					lockUrl,
+					{
+						'Lock-Token': lockToken
+					}).then(function (result) {
+						if (result.status === 204) {
+							var locks = self.model.get('activeLocks');
+							locks = locks.filter(function(item) {
+								return item.locktoken !== lockToken;
```
PVince81: does it mean the server might return lock tokens not related to the one you passed in the header, and you need to filter them out ? if yes, add as code comment to clarify

apps/files/lib/BackgroundJob/CleanupPersistentFileLocks.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace OCA\Files\BackgroundJob;
+
+use OC\BackgroundJob\TimedJob;
+use OC\Lock\Persistent\LockMapper;
+
+/**
+ * Clean up persistent file locks that are expired
+ */
+class CleanupPersistentFileLocks extends TimedJob {
```
PVince81: aha, I think you can then tick the checkbox about lock cleanup, unless it's something else ?

apps/files_sharing/lib/SharedStorage.php
```diff
@@ -456,4 +457,23 @@ public function getWrapperStorage() {
 
 		return $this->sourceStorage;
 	}
+
+	public function getLocks(string $internalPath, bool $returnChildLocks = false): array {
+		$locks = parent::getLocks($this->getSourcePath($internalPath), $returnChildLocks);
+		return \array_map(function (ILock $lock) {
+			// TODO: if path starts with rootpath
```
PVince81: is that something to be concerned about before merging ?

lib/private/Files/Storage/Common.php
```diff
@@ -710,15 +713,24 @@ public function getVersions($internalPath) {
...
-	protected function convertInternalPathToGlobalPath($internalPath) {
+	public function convertInternalPathToGlobalPath($internalPath) {
 		$mounts = \OC::$server->getMountManager()->findByStorageId($this->getId());
-		$mount = \end($mounts);
-		$p = $mount->getMountPoint() . $internalPath;
+
+		$selectedMount = \end($mounts);
+		foreach ($mounts as $mount) {
+			$o = \explode('/', $mount->getMountPoint());
+			if ($o[1] === $this->owner) {
```
PVince81: hope this works with public links when shared from a system-wide external storage, to be verified

lib/private/Files/Storage/Common.php
```diff
@@ -748,4 +760,34 @@ public function saveVersion($internalPath) {
...
+		$storageId = $this->getCache()->getNumericStorageId();
+		$fileId = $this->getCache()->getId($internalPath);
+		return $locksManager->lock($storageId, $internalPath, $fileId, $lockInfo);
+	}
+
+	public function unlockNodePersistent(string $internalPath, array $lockInfo) {
+		/** @var LockManager $locksManager */
+		$locksManager = \OC::$server->query(LockManager::class);
+		$fileId = $this->getCache()->getId($internalPath);
+		return $locksManager->unlock($fileId, $lockInfo['token']);
```
PVince81: hmm, so lock() needs storageId, internalPath and $fileId but unlock only needs $fileId ?
why is $fileid alone not enough for `lock()` ?

lib/private/Files/Storage/Common.php
```diff
@@ -748,4 +760,34 @@ public function saveVersion($internalPath) {
...
+		/** @var LockManager $locksManager */
+		$locksManager = \OC::$server->query(LockManager::class);
+		$storageId = $this->getCache()->getNumericStorageId();
+		$fileId = $this->getCache()->getId($internalPath);
+		return $locksManager->lock($storageId, $internalPath, $fileId, $lockInfo);
+	}
+
+	public function unlockNodePersistent(string $internalPath, array $lockInfo) {
+		/** @var LockManager $locksManager */
+		$locksManager = \OC::$server->query(LockManager::class);
```
PVince81: expose with convenience method on server container?

lib/private/Lock/Persistent/Lock.php
```diff
@@ -0,0 +1,93 @@
...
+	protected $depth;
+	/** @var int - foreign key zu oc_account.id */
+	protected $ownerAccountId;
+
+	/** @var string - joined with oc_filecache */
+	protected $path;
+	/** @var string - computed value */
+	protected $globalUserId;
+	/** @var string - computed value */
+	protected $globalFileName;
```
PVince81: what's a "globalFileName" and a "globalUserId" ? explain in the comment

maybe rename globalFileName to absolutePath

lib/private/Lock/Persistent/LockManager.php
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		$locks = $this->lockMapper->getLocksByPath($storageId, $internalPath, false);
+		$exists = false;
+		foreach ($locks as $lock) {
+			if ($lock->getToken() === $lockInfo['token']) {
+				$exists = true;
+				$lock->setCreatedAt(\time());
+				$lock->setTimeout($timeout);
+				if (empty($lock->getOwner())) {
```
PVince81: hopefully the owner will never be called "0" ?

lib/private/Lock/Persistent/LockManager.php
```diff
@@ -0,0 +1,128 @@
...
+			$depth = $lockInfo['depth'];
+		}
+		$scope = ILock::LOCK_SCOPE_EXCLUSIVE;
+		if (isset($lockInfo['scope'])) {
+			$scope = $lockInfo['scope'];
+		}
+
+		$lock = new Lock();
+		$lock->setFileId($fileId);
+		$lock->setCreatedAt(\time());
```
PVince81: use time factory for testability
also add unit tests ;-)

lib/private/Lock/Persistent/LockMapper.php
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public function getLocksByPath(int $storageId, string $internalPath, bool $returnChildLocks) : array {
+		$query = $this->db->getQueryBuilder();
+		$pathPattern = $this->db->escapeLikeParameter($internalPath) . '%';
+
+		$query->select(['id', 'owner', 'timeout', 'created_at', 'token', 'token', 'scope', 'depth', 'file_id', 'path', 'owner_account_id'])
+			->from($this->getTableName(), 'l')
+			->join('l', 'filecache', 'f', $query->expr()->eq('l.file_id', 'f.fileid'))
+			->where($query->expr()->eq('storage', $query->createNamedParameter($storageId)))
+			->andWhere($query->expr()->gt('created_at', $query->createFunction('(' . $query->createNamedParameter(\time()) . ' - `timeout`)')));
```
PVince81: use time factory for testability

lib/private/Lock/Persistent/LockMapper.php
```diff
@@ -0,0 +1,141 @@
...
+
+		$currentPath = '';
+		foreach ($uriParts as $part) {
+			if ($currentPath) {
+				$currentPath .= '/';
+			}
+			$currentPath .= $part;
+			$query->orWhere(
+				$query->expr()->andX(
+					// TODO: think about parent locks for depth 1
```
PVince81: should we meet and talk about these ?

lib/private/Lock/Persistent/LockMapper.php
```diff
@@ -0,0 +1,141 @@
...
+			throw new \InvalidArgumentException('token_hash does not match the token of the lock');
+		}
+		return parent::update($entity);
+	}
+
+	public function cleanup() {
+		$query = $this->db->getQueryBuilder();
+		$query->delete($this->getTableName())
+			->where($query->expr()->lt('created_at',
+				$query->createFunction('(' . $query->createNamedParameter(\time()) . ' - `timeout`)')))
```
PVince81: time factory + unit test

PVince81 (commented):
Reviewed.

There seems to be a few open topics (TODOs) to look at.

For the JS code we could move it to a separate PR and add unit tests there.

Also potentially missing tests in the sections about "use time factory".

PVince81:
can you also update the top post to match the new data model ? seems there are some new fields like "global user id" and "global file name"

lib/private/Files/Storage/Common.php
DeepDiver1975: To be tested ....

lib/private/Files/Storage/Common.php
DeepDiver1975: I don't think this is needed in Public API level

lib/private/Lock/Persistent/Lock.php
DeepDiver1975: The path via webdav. E.g. Foo/bar.txt

Name this absolute path? Sounds good

lib/private/Files/Storage/Common.php
DeepDiver1975: I need to Look into the implementation

apps/dav/lib/Connector/Sabre/LockPlugin.php
```diff
@@ -81,4 +86,26 @@ public function releaseLock(RequestInterface $request) {
...
+	 */
+	public function beforeUnlock($uri, LockInfo $lock) {
+		/** @var LockMapper $mapper */
+		$mapper = \OC::$server->query(LockMapper::class);
+
+		$lock = $mapper->getLockByToken($lock->token);
+		if ($lock === null || $lock->getOwnerAccountId() === null) {
+			return;
+		}
+		$currentUser = \OC::$server->getUserSession()->getUser();
```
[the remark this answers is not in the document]
DeepDiver1975: Test coverage needs to be pushed. Will work ob this asap

apps/dav/lib/Connector/Sabre/LockPlugin.php
```diff
@@ -81,4 +86,26 @@ public function releaseLock(RequestInterface $request) {
...
+	 * @param $uri
+	 * @param LockInfo $lock
+	 * @throws Forbidden
+	 * @throws \OCP\AppFramework\Db\DoesNotExistException
+	 * @throws \OCP\AppFramework\Db\MultipleObjectsReturnedException
+	 * @throws \OCP\AppFramework\QueryException
+	 */
+	public function beforeUnlock($uri, LockInfo $lock) {
+		/** @var LockMapper $mapper */
+		$mapper = \OC::$server->query(LockMapper::class);
```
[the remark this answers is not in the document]
DeepDiver1975: comment added to code

apps/dav/lib/Files/FileLocksBackend.php
```diff
@@ -75,12 +75,12 @@ public function getLocks($uri, $returnChildLocks) {
 		$davLocks = [];
 		foreach ($locks as $lock) {
 			$lockInfo = new Locks\LockInfo();
-			$fileName = $lock->getGlobalFileName();
+			$fileName = $lock->getAbsolteDavPath();
```
phil-davis: It is going to explode:
```
public function getAbsoluteDavPath()
````

s/getAbsolteDavPath/getAbsoluteDavPath

**Comment by DeepDiver1975** [comment]
> can you also update the top post to match the new data model ? seems there are some new fields like "global user id" and "global file name"

these two fields are not in the database - thats why I did not add them above

File: apps/dav/lib/Connector/Sabre/LockPlugin.php (hunk 17 above)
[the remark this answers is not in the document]
Comment (PVince81) [reply apps/dav/lib/Connector/Sabre/LockPlugin.php]: did you end up injecting or did you decide against it ?

File: apps/dav/tests/unit/DAV/LockPluginTest.php [hunk 20]
```diff
@@ -0,0 +1,92 @@
...
+		$this->server = $this->createMock(Server::class);
+		$this->tree = $this->createMock(Tree::class);
+		$this->server->tree = $this->tree;
+
+		$this->plugin->initialize($this->server);
+		$this->lockMapper = $this->createMock(LockMapper::class);
+		$this->userSession = $this->createMock(IUserSession::class);
+
+		$this->overwriteService(LockMapper::class, $this->lockMapper);
+		$this->overwriteService('UserSession', $this->userSession);
```
PVince81: hmm ok, so you overrode it

apps/dav/tests/unit/Files/FileLocksBackendTest.php
PVince81: ok, seems to be done

apps/files/appinfo/info.xml
PVince81: @DeepDiver1975 ping

apps/dav/lib/Connector/Sabre/LockPlugin.php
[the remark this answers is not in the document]
DeepDiver1975: decided against for perf reasons

apps/dav/tests/unit/Files/FileLocksBackendTest.php
DeepDiver1975: done

apps/dav/lib/Files/FileLocksBackend.php
```diff
@@ -0,0 +1,160 @@
...
+			return [];
+		}
+
+		/** @var IPersistentLockingStorage $storage */
+		$locks = $storage->getLocks($node->getFileInfo()->getInternalPath(), $returnChildLocks);
+
+		$davLocks = [];
+		foreach ($locks as $lock) {
+			$lockInfo = new Locks\LockInfo();
+			$fileName = $lock->getGlobalFileName();
```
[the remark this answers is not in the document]
DeepDiver1975: done

apps/dav/lib/Files/FileLocksBackend.php
```diff
@@ -0,0 +1,160 @@
...
+
+			if ($this->useV1) {
+				$lockInfo->uri = $fileName;
+			} else {
+				$uid = $lock->getGlobalUserId();
```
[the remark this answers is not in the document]
DeepDiver1975: done - dav user as discussed

apps/files/appinfo/info.xml
DeepDiver1975: moved to own pr

apps/files/js/filelockplugin.js
DeepDiver1975: own pr

apps/files/js/locktabview.js
DeepDiver1975: if we need one day - yes - but anyway own pr

apps/files/js/locktabview.js
DeepDiver1975: own pr - will add explanation there

apps/files/lib/BackgroundJob/CleanupPersistentFileLocks.php
DeepDiver1975: yes - ticked long time back

apps/files_sharing/lib/SharedStorage.php
DeepDiver1975: can be removed - old comment

lib/private/Files/Storage/Common.php
DeepDiver1975: the lock operation needs information about path and storage due to it's implementation

lib/private/Lock/Persistent/LockManager.php
DeepDiver1975: alternative !== '' && !== null :question:

PVince81: yeah, that's what we usually do

individual-it:
> As discussed before, for litmus one idea would be to have a behat scenario step "and Webdav path /XYZ is Webdav compliant" that runs litmus in the background.

Actually we only need a `Then` line but to make it more compliant with Behat structure I'm thinking along this lines
```
When the path "/XYZ" is used as webDav path
Then all litmus checks should pass
```
The challenge would be to change the way the litmus docker container works. Currently we set the URL, start the container and the container checks the URL. We would need to change the container, so that it works as an service with some kind of simple API. We would give it an URL, it runs the checks, returns the result and waits for the next request.

@PVince81 @DeepDiver1975 @tboerger @patrickjahns @phil-davis

DeepDiver1975:
> The challenge would be to change the way the litmus docker container works. Currently we set the URL, start the container and the container checks the URL. We would need to change the container, so that it works as an service with some kind of simple API. We would give it an URL, it runs the checks, returns the result and waits for the next request.

can't we do an exec in php of the command: docker run -e LITMUS_URL ...... litmus:latest and parse the output?

Building a service with API and anything related is a bit time consuming

individual-it:
I don't think we can start a new docker from within a drone run, because when we are running PHP we are already in a docker containers.
We only need a service listening on a port, receiving 3 variables and starting a command. Maybe a bit of curl magic would do

patrickjahns:
Can you please clarify, why you need to combine behat and litmus? Why not simply use either one for this?

individual-it:
@PVince81 @DeepDiver1975 ^^

DeepDiver1975:
This was the Idea. https://github.com/owncloud/core/pull/31651#issuecomment-409601585

patrickjahns:
I can read that - but why does it need to be combined - why can't we just setup the parts beforehand and then run the litmus container in a seperate step with the previously setup path?

Alternatively - port the litmus tests into a behat suite an we provide it as package for others to use again

DeepDiver1975:
> I can read that - but why does it need to be combined - why can't we just setup the parts beforehand and then run the litmus container in a seperate step with the previously setup path?

the idea was to reuse the functionalities we have in the behat env like creating a local mounted folder, a shared folder and so on and run litmus then on these folders.

If we cannot combine both we need to reinvent this for our litmus test suite



> Alternatively - port the litmus tests into a behat suite an we provide it as package for others to use again

I think this is out of scope .... but worth a thought ...

individual-it:
> Building a service with API and anything related is a bit time consuming

what about having a php file inside the litmus container run by php-dev server just listening and executing litmus with parameters whenever there is a request

PVince81:
assigning self for re-review

PVince81 (approved):
:+1:

DeepDiver1975:
cool - thx

dalenoe:
Will this work with redis and/or memcached?

DeepDiver1975:
> Will this work with redis and/or memcached?

database

PR #31656, opened 2018-06-05
Allow different language in public link share email

VicDeo:
## Description
Allow set another language for email notification when sharing with link

## Related Issue
https://github.com/owncloud/enterprise/issues/2435

## How Has This Been Tested?
Case 1
1. Share via link and send an email
2. Check your mailbox

### expected
 email notification has language in accordance with language set in owner personal settings

Case 2
1.  `php occ config:app:set core shareapi_public_notification_lang --value 'ru'`
2. Share via link and send an email
3. Check your mailbox

### expected
1.  email notification is in Russian

Note: in both cases `personal message` is not translated

## Types of changes

## Checklist:


VicDeo:
@PVince81 @pmaier1 should it also be configurable via settings page or CLI is enough?

pmaier1:
> @PVince81 @pmaier1 should it also be configurable via settings page or CLI is enough?

Unless @PVince81 has a different opinion I think CLI should be enough for this. Default value should be "English" btw. And for CLI to be a good solution, this needs documentation (for the command and possible values)!! Please take care. /cc @settermjd

PVince81:
If not too much effort we can add a setting in the settings page. I expect this to be a matter of copying the existing language selection box and adjusting. Let's do this in a separate PR (optional task).

@pmaier1 we need to keep in mind the difference between the sysadmin and OC admin roles. Only the sysadmin has access to CLI and OC admin needs to setup everything from the web UI.

lib/private/legacy/template.php
```diff
@@ -76,13 +77,17 @@ class OC_Template extends \OC\Template\Base {
...
+	public function __construct($app, $name, $renderAs = "", $registerCall = true, $l10n = null) {
 		self::initTemplateEngine($renderAs);
 		$requestToken = (OC::$server->getSession() && $registerCall) ? \OCP\Util::callRegister() : '';
 
-		$parts = \explode('/', $app); // fix translation when app is something like core/lostpassword
-		$l10n = \OC::$server->getL10N($parts[0]);
+		// fix translation when app is something like core/lostpassword
+		$parts = \explode('/', $app);
+
+		$languageCode = $l10n === null ? null : $l10n->getLanguageCode();
```
PVince81: if the language code of the injected is different, doesn't it mean you can simply reuse that instance instead of re-getting it ?

PVince81 (approved):
Looks good, see my comment for clarification

lib/private/legacy/template.php
VicDeo: @PVince81 I did it this way in the previous PR, but this was incorrect as l10n instance depends both on the app and locale. So we need to create an l10n instance with app matching `$app` ctor arg

PVince81: can you add a PHP comment so future onlookers will not wonder ?

also adjust the constructor PHPDoc and clarify which IL10N instance needs to be injected (generic vs app specific vs ...)

VicDeo:
@PVince81 @pmaier1 
>Default value should be "English" 

This way we will change the current behavior (notification in the lang of the owner) with no fallback option.

As for UI this selector makes sense only when mail notifications are enabled

lib/private/legacy/template.php
VicDeo: @PVince81 I thought it over, let's pass just a language code instead of an object that is not fully suitable.

pmaier1:
> @pmaier1 we need to keep in mind the difference between the sysadmin and OC admin roles. Only the sysadmin has access to CLI and OC admin needs to setup everything from the web UI.

Hm, yes, agreed. Taking this into account a dropdown in admin "General" settings would be good.

VicDeo:
disagree. Why in "General", when it is related to sharing and applicable only in case mail notifications for public links are enabled?

pmaier1:
> disagree. Why in "General", when it is related to sharing and applicable only in case mail notifications for public links are enabled?

Hmm ok, good point. I though it would apply to other notifications or potential future notifications as well. If it's just for link shares and will continue to be then probably next to the setting to enable link share sending in the "Sharing" settings fits better.

PVince81:
@VicDeo add UI selector in this PR or separately ?

VicDeo:
@PVince81 better separately, unless you want me to c&p most of code from https://github.com/owncloud/core/blob/master/settings/Panels/Personal/Profile.php

PVince81:
@VicDeo please backport

PVince81:
I just tried a quick backport but there are conflicts on the template level.
Leaving this up to you to solve them, might need to find out why this part of code is so different from master.

VicDeo:
Stable10: https://github.com/owncloud/core/pull/31767

PR #31613, opened 2018-06-01
[stable10] Accept user share stable10

jvillafanez:
## Description
Backport of https://github.com/owncloud/core/pull/30106 to stable10

## Related Issue
https://github.com/owncloud/core/issues/19153

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

@owncloud/qa please verify the .drone.yml configuration to recheck all the tests we want to run are there. There were some conflicts there so I've mostly copied whatever there is in the master branch.

PR #31665, opened 2018-06-06
Move notifications context API steps used by core into core

phil-davis:
## Description
1) Make a ``NotificationsCoreContext`` that has the relevant step code for examining notifications through its API.
2) Move ``WebUINotificationsContext`` into core - all the code there was relevant to testing behavior from core or from other apps.
3) Move the other notifications page objects also - they were all needed to let ``WebUINotificationsContext`` do its thing.
4) Adjust ``behat.yml`` and ``bootstrap.php`` so they do not try to look inside the notifications app for acceptance test code.

## Motivation and Context
Acceptance test code (scenario steps...) that is needed/used by core acceptance test scenarios needs to be in core. In  this case, move relevant scenario steps (and related page objects that define the webUI elements) out of the notifications app and into core, so they can be used by core or any other app, without that app having to find a way to get hold of ``NotificationsContext``... from somewhere inside the notifications app.

After this change, other apps can just enable or disable the notifications app and then can use any of the acceptance test steps in core to examine the behavior of their app when it produces notifications.

## Types of changes

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/NotificationsCoreContext.php
+++ b/tests/acceptance/features/bootstrap/NotificationsCoreContext.php
@@ -1,0 +1,270 @@
+<?php
+/**
+ * @author Joas Schilling <nickvergessen@owncloud.com>
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License,
+ * version 3, along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+use Behat\Behat\Context\Context;
+use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
+use GuzzleHttp\Message\ResponseInterface;
+use TestHelpers\OcsApiHelper;
+
+require_once 'bootstrap.php';
+
+/**
+ * Defines application features from the specific context.
+ */
+class NotificationsCoreContext implements Context, SnippetAcceptingContext {
+
+	/**
+	 * @var array[]
+	 */
+	protected $notificationIds;
+
+	/**
+	 * @var int
+	 */
+	protected $deletedNotification;
+
+	/**
+	 * @var FeatureContext
+	 */
+	private $featureContext;
+
+	/**
+	 * @return array[]
+	 */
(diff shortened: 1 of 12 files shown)


phil-davis:
1st commit removes the need for core API acceptance tests to load ``NotificationsContext`` from inside the notifications app. The needed scenario steps are put into ``NotificationsCoreContext`` in core.

phil-davis:
Note: backport needs to be added to Accept user share stable10 #31613

tests/acceptance/features/bootstrap/NotificationsCoreContext.php
```diff
@@ -0,0 +1,270 @@
...
+use Behat\Behat\Context\Context;
+use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
+use GuzzleHttp\Message\ResponseInterface;
+use TestHelpers\OcsApiHelper;
+
+require_once 'bootstrap.php';
+
+/**
+ * Defines application features from the specific context.
```
patrickjahns: what does this essentially mean?

tests/acceptance/features/bootstrap/NotificationsCoreContext.php
```diff
@@ -0,0 +1,270 @@
+<?php
+/**
+ * @author Joas Schilling <nickvergessen@owncloud.com>
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2016, ownCloud, Inc.
```
patrickjahns: missing you as author?

tests/acceptance/features/bootstrap/WebUINotificationsContext.php
```diff
@@ -0,0 +1,171 @@
...
+			throw new InvalidArgumentException();
+		}
+		if ($notification === false) {
+			throw new \Exception(__METHOD__ . " no notifications found");
+		}
+		$notification->followLink($this->getSession());
+	}
+
+	/**
+	 * @When /^the user reacts with "(Accept|Decline)" to all notifications on the webUI$/
```
patrickjahns: isn't that more related to accepting/rejecting shares?

patrickjahns (approved):
LGTM

tests/acceptance/features/bootstrap/NotificationsCoreContext.php
phil-davis: The code is cut from ``NotificationsContext``` in notifications app. These were the authors there. I have no idea who wrote which functions... so am preserving what was in the header.

tests/acceptance/features/bootstrap/NotificationsCoreContext.php
phil-davis: Cut and paste from ``NotificationsContext``` in notifications app. I guess someone wrote this so as to have text in the docblock to pass a code style checker :)

tests/acceptance/features/bootstrap/WebUINotificationsContext.php
phil-davis: This whole file is copied as-is from notifications app (all the functions were needed by core).
I  agree, the words could have "sharing" somewhere - e.g.
```
the user reacts with "(Accept|Decline)" to all sharing notifications on the webUI
```
and the step itself could live in a Sharing Context (but it needs both sharing and notifications together to be functional).

Anyway, this refactoring was about moving existing code. We can review the content of existing code any time we like.

Issue #31578, opened 2018-05-29
Prevent sharing with system groups


PVince81:
System has internal groups that are important in OC
- groups are useful for features like "exclude app from groups", "external storage with groups", whatever
- but group must not be used for sharing, for example because it contains all system users so admin wants to exclude these

- [ ] add new setting for blacklist
- [ ] filter out results when querying groups to share based on blacklist

Need to find the right technical approach for this as there are many challenges.

PVince81:
Some requirements:

- the share autocomplete result must be filtered by the server so that any client receives the pre-filtered list
- sharing directly with blacklisted groups must be forbidden by API (403)

PVince81:
Possible approaches:

- filtering the list
    - above GroupManager level, OCS Share API level:
        - advantage: closer to share autocomplete call
    - GroupManager level, when scope is "sharing":
        - advantage: autocomplete call will be pre-filtered
        - disadvantage: will break pagination because it will remove entries from within the middle of pages
    - on group backend level (below GroupManager):
        - advantage: blacklist entries could be included in SQL query or LDAP query for exclusion
        - advantage: pagination still works
        - disadvantage: too specific, every backend needs adjustment and specific implementation
    - other approaches ?

- preventing to share with blacklisted groups:
    - to be done on OCS Share API level, likely ShareManager

@jvillafanez any thoughts or suggestions ?

PVince81:
Also to think about:

- can custom groups be blacklisted ?

PVince81:
the dilemma is: either filter in the GroupManager and break pagination because we remove the blacklisted group in the middle or expect the group backend to filter based on given array of "to exclude" groups which needs touching every group backend but might be more efficient

PVince81:
@jvillafanez so if I understand correctly, if we go with the first option "Set the option as part of the files_sharing app." the restriction will be in place in:
    - OCS Share API autocomplete endpoint, filter out dynamically after getting the results from the group backend (might break pagination, but we don't use "offset" in there AFAIK so it's ok)
    - OCS Share API when creating new shares
    - pros: more isolated in the app, responsibility clearer
    - cons: apps can still share with blacklisted groups by talking to ShareManager directly and would require to know about this option

In any case, if an app is requiring a user to specify a group, they'd better use the existing share autocomplete endpoint instead of providing their own.

butonic:
@tomneedham may have a stronger opinion because he used scopes for the support portal

tomneedham:
[...]

PR #31672, opened 2018-06-07
Allow admin to blacklist groups for the files_sharing app

jvillafanez:
## Description
Allow admins to blacklist groups so user can't share with those groups. Blacklist will be done per backend + group displayname for now.

## Related Issue
https://github.com/owncloud/core/issues/31578

## How Has This Been Tested?
Manually tested for now

## Types of changes

## Checklist:


apps/files_sharing/js/settings.js
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * This file is licensed under the Affero General Public License version 3
+ * or later.
+ *
+ * See the COPYING-README file.
+ *
+ */
+
+$(document).ready(function(){
+	$('#files_sharing .files_sharing_settings').change(function(){
```
PVince81: so the generic functions didn't work ? might be possible to adjust the generic existing JS function for saving settings to include `textarea` additionally to `input`

apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -396,6 +401,9 @@ public function createShare() {
 				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
 				return new \OC\OCS\Result(null, 404, $this->l->t('Please specify a valid group'));
 			}
+			if ($this->sharingBlacklist->isGroupBlacklisted($this->groupManager->get($shareWith))) {
+				return new \OC\OCS\Result(null, 403, $this->l->t('The group is blacklisted for sharing'));
```
PVince81: not sure if this qualifies as information disclosure, maybe just return 404 "Please specify a valid group" as above

cc @settermjd

apps/files_sharing/lib/Controller/SettingsController.php
```diff
@@ -0,0 +1,65 @@
...
+
+class SettingsController extends Controller {
+	/** @var IConfig */
+	private $config;
+
+	/**
+	 * Set with the known configuration keys. Only these keys should be able to be set and retrieved
+	 */
+	private $knownKeys = [
+		'blacklisted_group_displaynames' => true,
```
PVince81: looks like early optimization/refactoring before it is even needed, but ok

apps/files_sharing/lib/Panels/Admin/SettingsPanel.php
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	public function getPanel() {
+		$groupBackendList = $this->groupManager->getBackends();
+		$backendNames = \array_map(function($backend) {
+			return \get_class($backend);
+		}, $groupBackendList);
+
+		$tmpl = new Template('files_sharing', 'settings');
+		$tmpl->assign('backendNames', $backendNames);
```
PVince81: great that you put this in :+1:

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -0,0 +1,107 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Sharing;
+
+use OCP\IConfig;
+use OCP\IGroup;
+
+class SharingBlacklist {
```
PVince81: Add PHPDoc. Is this the actual black list array or just a utility class ? This is what a developer wonders about when seeing only the class name used in various places

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -0,0 +1,107 @@
...
+
+	private $blacklistCache = null;
+
+	public function __construct(IConfig $config) {
+		$this->config = $config;
+	}
+
+	/**
+	 * Check if the target group is blacklisted
+	 * @param IGroup $group the group to check
```
PVince81: missing `@return`

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -0,0 +1,107 @@
...
+			];
+		}
+	}
+
+	private function fetchBlacklistedGroupDisplaynames() {
+		$configuredBlacklist = $this->config->getAppValue('files_sharing', 'blacklisted_group_displaynames');
+		$blacklistedComponents = explode("\n", $configuredBlacklist);
+
+		$result = [];
+		foreach ($blacklistedComponents as $blacklistedComponent) {
```
PVince81: is this operation so expensive that we need a cache with all complexity that comes with it ?

in general I don't think we expect more than 10 or so blacklisted groups

apps/files_sharing/templates/settings.php
```diff
@@ -0,0 +1,39 @@
...
+	<div class="horizontal-layout">
+		<div>
+			<p><?php p($l->t('Available backends names are:')); ?></p>
+			<ul class="groupBackends">
+				<?php foreach($_['backendNames'] as $backendName): ?>
+				<li><?php p($backendName); ?></li>
+				<?php endforeach; ?>
+			</ul>
+		</div>
+		<textarea placeholder="OC\Group\Database::localGroups&#10OCA\User_LDAP\Group_Proxy::ldapGroup" class="files_sharing_settings" name="blacklisted_group_displaynames"><?php p($_['blacklistedDisplaynames']); ?></textarea>
```
PVince81: what's this weird `&#10` entity here ? likely a newline. I suggest to put this in a template variable instead and initialize it from the PHP template in which you can write it in a more readable format that includes `\n`

PVince81 (commented):
See review comments.

I'm not convinced about the need to have a cache here.

apps/files_sharing/js/settings.js
jvillafanez: I'm not aware of any generic function to save values in the appconfig table. I've just found the `OC.AppConfig.setValue` function to perform the operation (code already adjusted to use it, soon), but there isn't auto binding, so I'll need to bind the event manually.

apps/files_sharing/lib/Controller/SettingsController.php
jvillafanez: I'll remove the class because get and set for the appconfig table will go through core (using OC.Appconfig in js). No need to handle this on our own.

apps/files_sharing/lib/SharingBlacklist.php
jvillafanez: For the sharees we'll need to check if each group is blacklisted. This leads potentially to a lot of calls to this function.
Since the blacklist will be stored as a large string, each call will need to fetch the string and process it in order to make the decission. This might be heavy we need to check a lot of groups.
In addition, I'd rather not rely on the appconfig to be cached to have a decent performance.

I also don't expect a lot of blacklisted groups, but we might need to include more options for the blacklist, such as blacklist by group id, and I'd prefer the check to be as fast as possible.

PVince81: Ok, fair enough. Make sure the unit tests also take the cache into account and prove that it properly invalidates, if applicable.

jvillafanez:
Code should be mostly ready, but there are tests missing for the SettingsPanel.php and SharingBlacklist.php and the `getBackends` method of the `IGroupManager`

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -0,0 +1,126 @@
...
+	 * @param IGroup $group the group to check
+	 * @return bool true if the group is blacklisted, false otherwise
+	 */
+	public function isGroupBlacklisted(IGroup $group) {
+		$this->initCache();
+
+		$groupBackend = \get_class($group->getBackend());
+		$groupDisplayname = $group->getDisplayName();
+
+		if (isset($this->blacklistCache['displaynames'][$groupBackend][$groupDisplayname])) {
```
tomneedham: I am still concerned we do the matching on the displaynames. Can't we store ID's but allow admins to specify displaynames?

jvillafanez:
Backport to stable10 in https://github.com/owncloud/core/pull/31740

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -0,0 +1,107 @@
...
+			$this->blacklistCache = [
+				'receivers' => [
+					'ids' => $this->fetchBlacklistedReceiverGroupIds(),
+				],
+			];
+		}
+	}
+
+	private function fetchBlacklistedReceiverGroupIds() {
+		$configuredBlacklist = $this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]');
```
PVince81: since we now removed the class names from the group ids, maybe we don't need the cache any more ?

based on @tomneedham's comment here https://github.com/owncloud/core/issues/31578#issuecomment-396246250 we can use the same approach as "exclude groups from sharing" which also doesn't expect a higher number of groups than the blacklist.

also now the logic for reading the settings has simplified itself since we don't store the prefix any more

@jvillafanez can you remove the cache and read directly from the setting ?

PVince81 (changes requested):
apart from the settings cache this looks good

apps/files_sharing/lib/SharingBlacklist.php
jvillafanez: I think we'll still need this. The problem is with sharees: we'll fetch a list of groups and each group needs to be checked whether is blacklisted or not; without proper caching we'll need to fetch the list from the configuration and then parse the list for each group.
Even if we rely on the configuration being cached, we still need to parse the stored json several times.

PVince81: is there no way to change the caller code then to only fetch the blacklist array once ?

jvillafanez: We might use a `filterBlacklistedGroups(array $groups)` function, but the calls would be a bit weird:
* The share API just needs to check one group, so you'll need to wrap it in an array, which feels bad.
* For the sharees, the intention would be to return the groups that aren't blacklisted. However, in case we need to do something with the blacklisted groups (maybe log them) things will get complicated both for the caller and the implementation. Note that, in any case, we'll need to traverse the sharees list twice.

I agree the code is a bit complex, but it should be fine if this complexity is isolated in this class. On the other hand, the usage is simple. Maybe we can document in the class what is the expected usage.

PVince81: I was more thinking that the caller code calls "give me all the blacklisted groups", store it in a local var and then use that when checking against each group. Unless the caller codes is spread across various classes, this could work.

jvillafanez: You'll move the complexity to the callers. you still need to compare the target group with the blacklist. This comparison code will be spread into several classes, and the way you'll likely compare will use the `in_array` function instead of more efficients approaches (you don't want to add more complexity to the caller).

Current code has a faster performance an a cleaner interface at a cost of including a cache.

PVince81: > Current code has a faster performance an a cleaner interface at a cost of including a cache.

Okay, let's leave it then

PVince81 (commented):
Code looks good.

What happens if shares exist that were shared with a group that has now been blacklisted ? I assume these must still appear in the shares list and be unshareable by users. I don't think we need to add a mechanism to also exclude existing shares/mounts that point to blacklisted groups.

@tomneedham thoughts on the latter ?

PVince81:
@jvillafanez let's move on with this. The additional concern, if addressed in any way, will not affect the code of this PR.

PVince81:
I've added a comment here to address as part of the ticket/concept before we close it: https://github.com/owncloud/core/issues/31578#issuecomment-399014984

PVince81 (approved):

PVince81:
@jvillafanez please adjust the backport PR and ping me there when done

apps/files_sharing/lib/API/Share20OCS.php
settermjd: Honestly? I'm don't believe that this is information disclosure. However, something more generic, yet still informative, such as "This group is not available for sharing, would work better.

apps/files_sharing/lib/Controller/ShareesController.php
```diff
@@ -284,7 +290,7 @@ protected function getGroups($search) {
 		foreach ($groups as $group) {
 			// FIXME: use a more efficient approach
```
settermjd: I can't think of a more efficient approach off the top of my head. However, you could use an implementation of a [FilterIterator](https://secure.php.net/manual/en/class.filteriterator.php) to simplify the logic here, making it that much simpler to maintain over time.

apps/files_sharing/tests/SharingBlacklistTest.php
```diff
@@ -0,0 +1,120 @@
...
+		$groupMock->method('getDisplayName')->willReturn($displayname);
+		return $groupMock;
+	}
+
+	public function isGroupBlacklistedProvider() {
+		$groupMock1 = $this->getGroupMock('Mygroup', 'my group');
+		return [
+			[$groupMock1, '["Mygroup"]'],
+			[$groupMock1, '["Mygroup", "my_other_group"]'],
+			[$groupMock1, '["one group", "Mygroup"]'],
```
settermjd: Isn't this, effectively, the same as the line above?

phil-davis: There were issues in various places with group names that contain a space - maybe that was a concern here?

PR #31677, opened 2018-06-07
Fix apiVersion type and createShare call parameters

phil-davis:
## Description
1) Fix call parameters for ``createShare``
2) Cast the ``apiVersion`` to ``int``` when storing it.

## Motivation and Context
``createShare()`` has a missing parameter  in the call - fix it.
When we fix it, and/or make use of ``createAPublicShare()`` then ``SharingHelper::createShare()`` complains about "invalid apiVersion/sharingApiVersion" because it now gets ``apiVersion`` in the correct parameter. ``apiVersion`` is suppposed to be an ``int``.

## How Has This Been Tested?
Local API acceptance test runs.

## Types of changes
- [x] Acceptance test bug fix

## Checklist:

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -261,7 +261,7 @@ trait BasicStructure {
 	 * @return void
 	 */
 	public function usingApiVersion($version) {
-		$this->apiVersion = $version;
+		$this->apiVersion = (int) $version;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


phil-davis:
We might not have been using the API version we thought we were using, when creating shares. So some tests that create and do stuff with shares might now misbehave.

paurakhsharma (approved):
I don't think any sharing tests user API version 2 so it should all pass. Let's see what CI says.

phil-davis:
Backport ``stable10`` #31679

PR #31050, opened 2018-04-09
[stable10] Backport all objectstore related changes

DeepDiver1975:
Targetting 10.0.9

- [x] backport #29207 - first 8 commits until 0a3c04567a1ef064693c333084f95a9e49c73ca4
- [x] backport #29257 - next 6 commits until aafd6eb46ff5d7ba00e94d5ef4e57f0b1c1717dc
- [x] backport #29607 - next 2 commits until e20f258b6127aeedfec84fc69e9ede738daf9f78
- [x] backport #29319 - next 6 commits until 4b8dc069860b5e28a998ded2fc2637914d2e891a
- [x] backport #30254 - 1 commit
- [x] backport #30192 - 1 commit
- [x] backport #29637 - 1 commit
- [x] backport https://github.com/owncloud/core/pull/31285 - 12 commits

PR #31680, opened 2018-06-07
Add version restore tests when sharee has moved the shared folder or file

phil-davis:
## Description
Add test cases for when a sharee receives a file or folder, then moves that file or folder and then tries to restore old versions.

Refactor some existing tests to check the version recovery a bit more cleanly.

## Related Issue
#31681 

## Motivation and Context
Some issues have been reported when the sharee (receiver of a shared file or folder) moves the received share, then tries to restore an old version. Make some tests to check this behavior.

## How Has This Been Tested?
Local API acceptance test run

## Types of changes
- [x] acceptance test enhancement

## Checklist:

### DIFF
--- a/tests/acceptance/features/apiMain/dav-versions.feature
+++ b/tests/acceptance/features/apiMain/dav-versions.feature
@@ -50,7 +50,7 @@ Feature: dav-versions
       | permissions | 8            |
     Then the version folder of fileId "<<FILEID>>" for user "user1" should contain "1" element
 
-	Scenario: sharer of a file see the old version information when the sharee changes the content of the file
+	Scenario: sharer of a file can see the old version information when the sharee changes the content of the file
 		Given user "user0" has been created
 		And user "user1" has been created
 		And user "user0" has uploaded file with content "user0 content" to "sharefile.txt"
@@ -59,7 +59,7 @@ Feature: dav-versions
 		Then the HTTP status code should be "204"
 		And the version folder of file "/sharefile.txt" for user "user0" should contain "1" element
 
-	Scenario: sharer of a file can restore the original content of the file after the file has been modified by the sharee
+	Scenario: sharer of a file can restore the original content of a shared file after the file has been modified by the sharee
 		Given user "user0" has been created
 		And user "user1" has been created
 		And user "user0" has uploaded file with content "user0 content" to "sharefile.txt"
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/apiMain/dav-versions.feature
```diff
@@ -133,9 +139,37 @@ Feature: dav-versions
...
+		Then the HTTP status code should be "201"
+		And the content of file "/sharingfolder/sharefile.txt" for user "user0" should be "old content"
+		And the content of file "/received/sharingfolder/sharefile.txt" for user "user1" should be "old content"
+
+	Scenario: sharee can restore a shared file created and modified by sharer, when the file has been moved by the sharee
+		Given user "user0" has been created
+		And user "user1" has been created
+		And user "user0" has uploaded file with content "old content" to "/sharefile.txt"
+		And user "user0" has uploaded file with content "new content" to "/sharefile.txt"
+		And user "user0" has shared file "/sharefile.txt" with user "user1"
```
PVince81: @phil-davis please add another scenario where the source shared file is in "/outgoing/sharefile.txt" and where "outgoing/sharefile.txt" is shared with "user1". This would match the steps from https://github.com/owncloud/core/issues/31681

PVince81: yes, it is ok on master and also my manual test doesn't reproduce the issue on stable10 backport of object store: https://github.com/owncloud/core/pull/31050

let's merge this PR and backport it then

PVince81 (approved):
:+1: please backport

phil-davis:
Backported to #31050 object store changes

PVince81:
@phil-davis can you also put them in a temp PR directly to stable10 to confirm the test will fail ? (or run it locally)

phil-davis:
I will look and see what test step code is needed out of the object-store backport, and if the used endpoints work in "raw" ``stable10`` ...
(I am not sure how it came to be that the whole ``dav-versions.feature`` file ended up in that object-store backport)

PVince81:
> (I am not sure how it came to be that the whole dav-versions.feature file ended up in that object-store backport)

@phil-davis I think dav-versions never existed before and got written from scratch for the object store part, because object store work provided new API endpoints for versions. So the tests in questions (at least the steps) can only work with that API endpoint, which is also provided somewhere in this PR.

We cannot easily move that directly to stable10 without rewriting the steps to use the old private APIs... so better leave it here where it belongs.

phil-davis:
Yes, I just found that in "raw" ``stable10`` I am getting:
```
     │ object(Sabre\HTTP\Response)#2177 (5) {
      │   ["status":protected]=>
      │   int(404)
      │   ["statusText":protected]=>
      │   string(9) "Not Found"
      │   ["body":protected]=>
      │   string(222) "<?xml version="1.0" encoding="utf-8"?>
      │ <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
      │   <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
      │   <s:message>File not found: meta in 'root'</s:message>
      │ </d:error>
  ```
The test code accesses a  ``meta`` endpoint which does not exist yet in ``stable10``

PVince81:
@phil-davis argh, indeed. That's the endpoint that the backport brings in for the version changes.

So we couldn't make this test work with the old API without changing the steps to talk to the old private API. Ok, let's leave this then.

phil-davis:
As a separate thing tomorrow morning, I will see if I can make a version restore test in the webUI that will go through the described steps in the issue. I think we have some webUI restore steps already, so it might be quite easy (famous last words) to construct a scenario that fails. That would provide something that a dev can use to confirm a fix.

PVince81:
great plan, thanks

PR #31712, opened 2018-06-11
Updating ca-bundle.crt

DeepDiver1975:
Update ca-bundle.crt from https://curl.haxx.se/docs/caextract.html

needs backport


settermjd (approved):
Seems fine to me.

Issue #31137, opened 2018-04-16
Tooltip in sharing autocomplete doesn't mention it is autocompleting


SergioBertolinSG:
### Steps to reproduce

1. Having a server with several users "aabaa", "aabb", "aabc"
2. In config.php add 'user.search_min_length' => 3,
3. Using admin try to share with "aac" wait for autocompletion.

### Expected behavior
A tooltip appears saying:
'No users or groups found containing "aac"'

### Actual behavior
A tooltip appears saying
'No users or groups found for "aac"'

![screen shot 2018-04-16 at 14 40 41](https://user-images.githubusercontent.com/9048796/38809759-eb3b8b34-4184-11e8-8193-c006f66c4e8c.png)

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.1

**Cache**
DB

**ownCloud version:** (see ownCloud admin page)
**{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.0.8.1","versionstring":"10.0.8 RC1","edition":"Enterprise","productname":"ownCloud"}**

**Updated from an older ownCloud or fresh install:**
Fresh

**The content of config/config.php:**

```

```

**Are you using external storage, if yes which one:** local/smb/sftp/...
No.

**Are you using encryption:** 
Yes.

**Logs**
```
```

### Client configuration

**Browser**
Chrome.  
  

@pmaier1 When having 'user.search_min_length' => 4 in config.php maybe the tooltip should not appear when searching "aac".

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/2343 (Autocomplete bug in share option), https://github.com/owncloud/core/issues/27734 (Sharing autocomplete user box  does not complete if avatar are disabled ), https://github.com/owncloud/core/issues/22910 (Tooltip of a fed.sharing remote user that not exists doesn't disappear), https://github.com/owncloud/core/issues/18551 ([mobile] Autocomplete drop down in sharing dialog not shown ), and https://github.com/owncloud/core/issues/22746 (Tooltip in federated sharing is not showing the full url.).

phil-davis:
I updated the expected tooltip text to:
No users or groups found containing "aac"

because the search actually returns any users or groups that contain the string in their display name.

phil-davis:
Case 2:
[...]

PR #31714, opened 2018-06-11
Add a notice to the share tooltip when there is no autocomplete

VicDeo:
## Description
- [x] change the default length for suggestions back to 2 (release notes!)
- [x] if the number of entered chars is below limit, show "Please enter at least X characters for suggestions", because you can still get exact matches when typing less. Message must be adjusted to singular if X is 1. (use the appropriate "n()" translation function argument)
- [x] if the number of entered chars is above or equal to X, show "No results" if no results were found

## Related Issue
Closes https://github.com/owncloud/core/issues/31137 (Tooltip in sharing autocomplete doesn't mention it is autocompleting)

## How Has This Been Tested?
1. set `user.search_min_length` to 10
2. enter 5 chars.
See "No users or groups found for 12345 Please enter at least 10 characters for suggestions"

3. Enter 10 or more characters
See "No users or groups found for 1234567890 "

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31718, opened 2018-06-11
Improve webUI autocompletion acceptance tests

phil-davis:
## Description
1) Make a more interesting list of users and groups for the tests - that have userid and display name that do not both match, where group names like "finance" match with user display names like "John Finn Smith" and so on.
2) Make the acceptance test code expect case-insensitive matches on user-id, display-name and group-id - so it will expect a lot more stuff in the auto-complete list.
3) Make the acceptance test code count how many things it matched, and assert that the auto-complete list is that big. i.e. there are not "bonus" matches in the list.
4) Add a step that more easily tests for a single match.
5) Re-order test setup steps - we have to change ``user.search_min_length`` before the user logs in and gets to the files page, otherwise their JavaScript does not know the current value of that setting, and so we do not always get the expected behavior.

## Related Issue
#31137 

## Motivation and Context
The autoCompletion webUI tests are not so great - make them great (again?)
Actually the auto-complete matches on group-id, user-id and user-display-name. If you try to share with "fin" then you will get matches in the list for "finance (group)", "John Finn" (display name), "fin123" (user-id - it will show the display name, which might not even contain "fin"). The matching is case-insensitive.
So the tests should check this range of matching.

## How Has This Been Tested?
Local webUI test runs.

## Types of changes

## Checklist:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -287,7 +287,7 @@ $CONFIG = array(
  * the minimum characters have been entered. The search is case insensitive.
  * e.g. entering "tom" will always return "Tom" if there is an exact match.
  */
-'user.search_min_length' => 4,
+'user.search_min_length' => 2,
 
 /**
  * Mail Parameters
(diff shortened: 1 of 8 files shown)


mmattel:
Any description ?

image

phil-davis:
Tests passing = good.
Just shouting about codecov.

phil-davis:
Commit has been cherry-picked across to #31714 
This PR can be closed when that gets merged...

PR #31719, opened 2018-06-11
occ files:scan fix groups with comma, plus tests 

phil-davis:
## Description
1) Test that groups containing a comma work OK in the Provisioning API.
2) Test that groups containing a comma work OK in the "excluded from sharing" group list.
3) Adjust the ``files_scan`` command so that the groups to scan are not put in a comma-separated list. Instead, list the groups to be scanned like ``--group=foo --group=bar`` - this allows specifying group names that contain commas.

## Related Issue
#31578 
#31835

## Motivation and Context
@tomneedham asked if comma is allowed in group names - this demonstrates that it works in the Provisioning API.

## How Has This Been Tested?
Local test runs

## Types of changes
- [x] Acceptance test enhancement

## Checklist:
- [x] My change requires a change to the documentation. https://github.com/owncloud/documentation/issues/4215

### DIFF
--- a/apps/files/lib/Command/Scan.php
+++ b/apps/files/lib/Command/Scan.php
@@ -107,8 +107,8 @@ class Scan extends Base {
 			->addOption(
 				'groups',
 				'g',
-				InputArgument::OPTIONAL,
-				'Scan user(s) under the group(s). This option can be used as --groups=foo,bar to scan groups foo and bar'
+				InputOption::VALUE_IS_ARRAY|InputOption::VALUE_REQUIRED,
+				'Scan user(s) under the group(s). This option can be used as --groups=foo --groups=bar to scan groups foo and bar'
 			)
 			->addOption(
 				'quiet',
@@ -276,7 +276,7 @@ class Scan extends Base {
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$inputPath = $input->getOption('path');
-		$groups = $input->getOption('groups') ? \explode(',', $input->getOption('groups')) : [];
+		$groups = $input->getOption('groups');
 		$shouldRepairStoragesIndividually = (bool) $input->getOption('repair');
 
 		if (\count($groups) >= 1) {
--- a/apps/files/lib/Command/Scan.php
+++ b/apps/files/lib/Command/Scan.php
@@ -105,10 +105,10 @@ class Scan extends Base {
 				'Limit rescan to this path, e.g., --path="/alice/files/Music", the user_id is determined by the path and the user_id parameter and --all are ignored.'
 			)
 			->addOption(
-				'groups',
+				'group',
 				'g',
 				InputOption::VALUE_IS_ARRAY|InputOption::VALUE_REQUIRED,
-				'Scan user(s) under the group(s). This option can be used as --groups=foo --groups=bar to scan groups foo and bar'
+				'Scan user(s) under the group(s). This option can be used as --group=foo --group=bar to scan groups foo and bar'
 			)
 			->addOption(
 				'quiet',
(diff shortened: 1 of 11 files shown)


individual-it (approved):
if this behaviour is expected :+1:

tomneedham:
My concern is that comma's are used as deliminators for groups in some places: 

screen shot 2018-06-19 at 15 32 38

This is what should be fixed probably.

phil-davis:
``lib/private/legacy/util.php`` - that is unused and code is being deleted #31823 

``lib/private/Share20/Manager.php`` - the excluded groups list "should" be ``json_encode``d but for some reason the code here handles the case where it is not, and falls back to making the excluded groups array by splitting on comma. I guess that some time in the past the excluded groups list was stored with comma-separator. The code then puts it back as ``json_encode`` - so it is an internal "on the fly migration" of the saved format. That should be fine. Unit tests added for that.

``apps/files/lib/Command/Scan.php`` - fixup the command interface so you say ``--groups=foo --groups=bar --groups=a,b`` to scan the groups ``foo`` ``bar`` and ``a,b``
or the abbreviated form ``-g=foo -g=bar -g=a,b``

and adjust ```ScanTest.php``

phil-davis:
This gets rid of all those places that try to ``explode`` a comma-separated list of group names.

phil-davis:
@DeepDiver1975 @PVince81 @tomneedham 
I rebased this just now, to make sure it still passes CI.
The only question I have is, should I change the ``files:scan`` ``--groups`` option to ``--group`` ?

```
occ files:scan --groups foo --groups bar
```

or

```
occ files:scan --group foo --group bar
```

The ``user:add`` command currently does:
```
occ user:add user1 --group foo --group bar
```

Please review.

mmattel:
pls make it that all —group(s) options will look the same in all existing and coming command changes. no mixup please !

phil-davis:
@mmattel I agree.

This will be a backward-compatibility break anyway, because:
```
occ files:scan --groups foo,bar
```
will now mean a single group called ``foo,bar``

So I guess I change ``--groups`` to ``--group`` also.

mmattel:
@phil-davis I allowed to change the headline to a better text.
The first one looked like to focus on tests and did not tell for which command it was made

mmattel:
Document relevant, please open a doc Issue/PR

phil-davis:
Doc PR https://github.com/owncloud/documentation/issues/4337

mmattel:
@phil-davis 
Idea, in repro password_policy PR 77, the options for using group are the same as here. What is new there, you can use ``-u`` for users but in the same way as for groups (elimiating btw commas in user names too). Why not harmonizing it and implement a ``-u`` option?
Maybe a separate PR, but just as an idea. You can ping me on talk to discuss if you want 😄

DeepDiver1975:
> occ files:scan --groups foo,bar

it should really be

> occ files:scan --groups foo bar test admin

options can have multiple values - no need for comma and string parsing magic - symfony command does this all for us

DeepDiver1975:
:-1: on backporting - this will break API. API breakage only on major updates - this goes into oc11/master only

for the rest: :+1:

phil-davis:
Backport in PR #32242 for commits that are adding Provisioning API and Unit Test cases that already pass.

The last 2 commits that change the ``files:scan`` command are NOT backported, due to backward-compatibility break.

phil-davis:
Backporting work is done here, so removing backport-request label.

mmattel:
@phil-davis will there be a PR that goes into stable10 even on hold for oc11, else this one gets lost completely in any sense and devides master and stable10.

phil-davis:
@mmattel oc11 will be a branch called stable11 that gets created off master at some time near when oc11 is "about to happen". The backward-compatibility breaks will not end up in oc10, because they break backward-compatibilty. That is the semver definition for versioning.

phil-davis:
I added the `v11` label. That is the current convention to label stuff that can't be released in the current `10.*` series, but needs to be easily found for when a "post-10" release is being sorted out.

phil-davis:
The last 2 commits have now been backported to `stable10` and will be in the next release (probably called `10.2`)
Removing the `v11` label.

PR #31644, opened 2018-06-05
Update sabre/dav to 3.3.0-alpha1

DeepDiver1975:
## Description
sabre/dav 3.3.0-alpha1 was released yesterday ... let's see how it works out for us

## Motivation and Context
Some significant changes for us are in there - see https://github.com/sabre-io/dav/releases/tag/3.3.0-alpha1

## How Has This Been Tested?
- requires full acceptance tests as executed by drone
- requires smashbox tests

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #29599, opened 2017-11-16
a file called "0" cannot be uploaded by chunking


individual-it:
### Steps to reproduce
1. create a 20MB file and call it "0"
2. upload that file via WebUI

### Expected behaviour
file should be uploaded correctly

### Actual behaviour
file is not uploaded, Error message is shown: `File with name //0 could not be located`

### Server configuration
**Operating system**:
Ubuntu 17.10
**Web server:**
Apache2
**Database:**
MySQL
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
10.0.3 (git) 
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "oc9xsmdtnlru",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "localhost"
        ],
        "datadirectory": "\/home\/artur\/www\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/localhost\/owncloud-core",
        "dbtype": "sqlite3",
        "logtimezone": "UTC",
        "installed": true,
        "license-key": "***REMOVED SENSITIVE VALUE***",
        "default_language": "en",
        "theme": "",
        "loglevel": 0,
        "maintenance": false,
        "log_type": "owncloud",
        "dbname": "owncloud",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "debug": false,
        "skeletondirectory": "\/home\/artur\/www\/owncloud-core\/tests\/ui\/skeleton",
        "version": "10.0.3.3",
        "csrf.disabled": true,
        "files_external_allow_create_new_local": "true"
    }
}
```

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - dav: 0.3.2
  - diagnostics: 0.1.0
  - enterprise_key: 0.1.3
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_sharing: 0.10.1
  - files_texteditor: 2.2
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - firewall: 2.4.2
[...]

PR #31723, opened 2018-06-12
Unskip tests for issue 29599

phil-davis:
since it should be fixed by "Update sabre/dav to 3.3.0-alpha1"

Just testing here to see if the tests pass. Then I can refactor them into the relevant scenario outlines... and put the final/proper commit in with PR #31644 

Related issue #29599


phil-davis:
based on not up-to-date branch

PR #31737, opened 2018-06-12
[stable10] Exclude from creating shares if member of any excluded groups

tomneedham:
Backport of https://github.com/owncloud/core/pull/31490 to stable10


PVince81 (approved):
:+1:

PR #31758, opened 2018-06-13
[POC]Introduce AMD style MOVE operation

DeepDiver1975:
## Description
We want to avoid situations where the clients have to wait for the long MOVE operation on assembly to finish. The connection can be terminated by any reason before the server sends the response back to the client. And after that the client has no capability to query for the status of the assembly.

1. AMD - Asynchronous Method Dispatch
We return the http status code 202 right after we received the MOVE request. The client can then terminate the connection and knows that the server is working on it.
The implementation uses php's register_callback functionality to process the real MOVE *after* the response is sent to the client. register_callback can not be terminated by the system - we should be pretty save on this one .... 

2. Polling for status, success and error
The client can run a PROPFIND on the upload folder and receive additional properties to get an understanding of the assembly status:

Property | Description
---|---
progress | not-started, started, finished, error
fileId | the file id of the target file - set only once the MOVE was successful
ETag | the ETag of the file *after* MOVE on the target location - set only once the MOVE was successfull
errorCode | http error code - in case there was an error
errorMessage | there error message - in case of an error

3. Capabilities and LazyOps header

The server will announce to the clients via the capabilities API if this new mode of operations is available or not. A system config value is necessary to allow the admin to turn this feature explicitly on.

Furthermore in order not to break regular WebDAV client an additional http header is to be sent from the client to the server so that the server knows if this operation mode shall be used or not. 

## To discuss
MOVE will delete the source file (and due to the current implementation the whole source folder).
Because we need the source folder we can either switch over to COPY and the client issues a DELETE or we delete only the source folder and the contained chunks but keep the source folder and the client has to delete the source folder.

This mechanism only works for the chunked upload. It would be great to go one step further and develop a general purpose solution because there are other operations which can take a long time and potentially run into the described timeout issues as well.

```
deepdiver@deepdiver:~$ curl -X MOVE  http://localhost:8080/remote.php/dav/files/admin/foo/IMG_101522156820911293053.jpeg -uadmin -H 'OC-LazyOps: true' -H 'Destination:  http://localhost:8080/remote.php/dav/files/admin/IMG.jpeg' -v
Enter host password for user 'admin':
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
* Server auth using Basic with user 'admin'
> MOVE /remote.php/dav/files/admin/foo/IMG_101522156820911293053.jpeg HTTP/1.1
> Host: localhost:8080
> Authorization: Basic YWRtaW46YWRtaW4=
> User-Agent: curl/7.60.0
> Accept: */*
> OC-LazyOps: true
> Destination:  http://localhost:8080/remote.php/dav/files/admin/IMG.jpeg
> 
< HTTP/1.1 202 Accepted
< Host: localhost:8080
< Date: Wed, 13 Jun 2018 14:23:59 +0000
< Connection: close
< X-Powered-By: PHP/7.1.16-1+b2
< Pragma: no-cache
< Set-Cookie: oc_sessionPassphrase=zEgKseQ63rau9sXrZdshT5dz2KL0nKJ8Wpx09bo9YKHSo0pBlAoZ%2Fr0QLjr5%2FL8WERzf1QH08PODd%2BSI8NIIsl6xbcvdkhpual7CO5ecCv6OGEhrKbrh0E3vVZ3WfEsH; path=/; HttpOnly
< X-Frame-Options: SAMEORIGIN
< X-Robots-Tag: none
< X-Download-Options: noopen
< Set-Cookie: oc48qu8iku2i=p6mo8boi9tfa3ufpb1495g6r5t; path=/; HttpOnly
< Set-Cookie: cookie_test=test; expires=Wed, 13-Jun-2018 15:23:43 GMT; Max-Age=3600
< Connection: close
< 
* Closing connection 0

```

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Server.php
+++ b/apps/dav/lib/Connector/Sabre/Server.php
@@ -25,6 +25,7 @@
 namespace OCA\DAV\Connector\Sabre;
 
 use OCA\DAV\DAV\CopyPlugin;
+use OCA\DAV\DAV\LazyOpsPlugin;
 
 /**
  * Class \OCA\DAV\Connector\Sabre\Server
@@ -44,5 +45,6 @@ class Server extends \Sabre\DAV\Server {
 		self::$exposeVersion = false;
 		$this->enablePropfindDepthInfinity = true;
 		$this->addPlugin(new CopyPlugin());
+		$this->addPlugin(new LazyOpsPlugin());
 	}
 }
--- a/apps/dav/lib/DAV/LazyOpsPlugin.php
+++ b/apps/dav/lib/DAV/LazyOpsPlugin.php
@@ -1,0 +1,84 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\DAV;
+
+use OCA\DAV\Connector\Sabre\Exception\Forbidden;
+use OCA\DAV\Connector\Sabre\File;
+use OCA\DAV\Files\ICopySource;
+use OCP\Files\ForbiddenException;
+use OCP\Lock\ILockingProvider;
+use Sabre\DAV\Exception;
+use Sabre\DAV\IFile;
+use Sabre\DAV\Server;
+use Sabre\DAV\ServerPlugin;
+use Sabre\HTTP\RequestInterface;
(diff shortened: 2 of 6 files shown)


butonic:
Cool, the browser closes the connection when we send the 202 ... then the actual move happens on the server side.

Currently the code defers the whole move operation. But we need it to only defer the assembly. we want the response to fail properly if not all chunks are available. Since the desktop client won't start the assembly before all chunks are uploaded we can probably ignore that. But more prominently we need the file to show up in the web ui, so the client *thinks* the file was uploaded, otherwise he will restart uploading the whole file.

DeepDiver1975:
> But more prominently we need the file to show up in the web ui, so the client thinks the file was uploaded, otherwise he will restart uploading the whole file.

this is rather critical - would a property on the upload/12345 resource help? this is less critical

patrickjahns:
> Currently the code defers the whole move operation. But we need it to only defer the assembly. we want the response to fail properly if not all chunks are available.

What about adding a "precondition check" in the sense of - check if base requirements are met for assembly?
This would however not cover the part of checksum errors - this would require a async communication to inform that assembly failed because of checksum errors

phil-davis:
> But more prominently we need the file to show up in the web ui, so the client thinks the file was uploaded, otherwise he will restart uploading the whole file.
this is rather critical - would a property on the upload/12345 resource help? this is less critical

I guess it "just" needs a server flow that saves a record of the new file in the DB with a status "backend data save in progress". Then front-end clients, incuding webUI,

DeepDiver1975:
on the upload folder a new property was added: assembly-status
inner properties are: progress, fileId and ETag

screenshot from 2018-06-19 13-14-27

guruz:
AMD stands for?

DeepDiver1975:
I'm closing this one in favor of #31851

PR #31343, opened 2018-05-03
Remove session token creation for dav requests during loadApps

tomneedham:
## Description
There was a new code path added here: https://github.com/owncloud/core/pull/28879 to create session tokens for dav requests. This is a partial revert. DAV requests use basic auth by default, which should be picked up here: 

https://github.com/owncloud/core/blob/31544601c5d3cc2f2ab63ed26c9d1c9d0648f0dc/lib/private/User/Session.php#L452

and token creation here:

https://github.com/owncloud/core/blob/31544601c5d3cc2f2ab63ed26c9d1c9d0648f0dc/lib/private/User/Session.php#L350

Which leaves this code path redundant. This can result in duplicate tokens because created causing UniqueViolation exceptions and confusing the client.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31759, opened 2018-06-13
[master] Revert auth 

### LINKED ISSUE #31455: [stable10][master] Session is not validated
Caused by https://github.com/owncloud/core/pull/31343

Does not affect 10.0.8.

Caused because $userSession->validateSession() is no longer called: https://github.com/owncloud/core/pull/31343/files#diff-3a7450cd4a63696420772086fff201aaL139

cc @butonic

tomneedham:
Revert of https://github.com/owncloud/core/pull/31343 to fix https://github.com/owncloud/core/issues/31455


tomneedham:
Needs backport

PVince81:
not sure if we should revert master... maybe not and we just re-revert the stable10 revert and keep it unmerged until 10.0.10

phil-davis:
why did @ownclouders assign me???

Issue #31549, opened 2018-05-28
Refresh the files view and display a yellow notification after rejecting/accepted button from notification


jvillafanez:
Follow up of https://github.com/owncloud/core/pull/30106#issuecomment-359007815 (https://github.com/owncloud/core/pull/30106#issuecomment-384547564)

### Steps to reproduce
(requires https://github.com/owncloud/core/pull/30106)
1. Send a share to a user
2. Login with that user
3. Click in the notification

### Expected behaviour
Visual feedback should be shown. The file list refreshes so the newly accepted share appears (in case it lands in the current folder)

### Actual behaviour
The action is done, but there is no indicator

### Server configuration
**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**ownCloud version:** (see ownCloud admin page)

**Updated from an older ownCloud or fresh install:**

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results into https://gist.github.com/ and puth the link here.
```

**The content of config/config.php:**

```
Log in to the web-UI with an administrator account and click on
'admin' -> 'Generate Config Report' -> 'Download ownCloud config report'
This report includes the config.php settings, the list of activated apps
and other details in a well sanitized form.

or 

If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

*ATTENTION:* Do not post your config.php file in public as is. Please use one of the above
methods whenever possible. Both, the generated reports from the web-ui and from occ config:list
consistently remove sensitive data. You still may want to review the report before sending.
If done manually then it is critical for your own privacy to dilligently
remove *all* host names, passwords, usernames, salts and other credentials before posting.
You should assume that attackers find such information and will use them against your systems.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder.
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

#### LDAP configuration (delete this part if not used)

```
With access to your command line run e.g.:
sudo -u www-data php occ ldap:show-config
from within your ownCloud installation folder

Without access to your command line download the data/owncloud.db to your local
computer or access your SQL server remotely and run the select query:
[...]

PR #31765, opened 2018-06-13
Reload the filelist view when accepting or rejecting a share

jvillafanez:
## Description
The "all files" list will be reloaded when a share notification is accepted, either local or remote.
The "shared with you" list will be reloaded when any action for a share notification (local or remote) is taken.

As remark, the "all files" list won't be reloaded if the share is rejected.

## Related Issue
https://github.com/owncloud/core/issues/31549

## Motivation and Context
User would need to manually reload the page to see the changes.

## How Has This Been Tested?
Manually tested

## Types of changes

## Checklist:

I haven't found a good way to trigger the reload of the "all files" list. Right now there is an explicit dependency between the files_sharing and files app (the `OCA.Files.App.fileList.reload()` call). For now this is the easiest way to fix the problem.

We might want to refine when we want to reload the "all files" view in order to reload only if the file would appear in the current folder.

### DIFF
--- a/apps/files_sharing/tests/js/appSpec.js
+++ b/apps/files_sharing/tests/js/appSpec.js
@@ -161,7 +161,7 @@ describe('OCA.Sharing.App tests', function() {
 					shareOwner: 'user2'
 				};
 			});
-			afterEach(function() { 
+			afterEach(function() {
 				showMenuStub.restore();
 			});
 			it('provides accept and reject actions for pending shares', function() {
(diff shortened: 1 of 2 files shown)


jvillafanez:
Errors seem unrelated to the PR

apps/files_sharing/tests/js/appSpec.js
```diff
@@ -275,4 +275,90 @@ describe('OCA.Sharing.App tests', function() {
...
+					app: 'files_sharing',
+				},
+				action: {
+					url: 'https://randomurl',
+					type: 'POST'
+				}
+			});
+			$('body').trigger(ev);
+
+			expect(1).toEqual(fileListInReloadStub.callCount);
```
PVince81: we usually swap the values: expect contains the variable and toEqual the expected value
it's the opposite of PHPUnit

PVince81 (changes requested):
Great job!

Just a little adjustment for the JS test expectation rformat

PVince81 (approved):
:+1: please backport

PR #31808, opened 2018-06-17
Tag scenario that is not relevant to encryption

phil-davis:
## Description
Tag ``resetPassword.feature`` scenario about reset password UI page with ``@skipOnEncryption``

## Related Issue
Encryption PR https://github.com/owncloud/encryption/pull/34 needs this.

## Motivation and Context
We need a way to indicate which core webUI tests are testing behavior that is overridden/changed in some app(s), or tests that just do not work nicely in some app(s). We already do this for LDAP by tagging such scenarios ``@skipOnLDAP``.

Now we want to run the webUI tests with backend encryption on. At the moment just 1 scenario fails - when a user resets their password. For encryption, there is an extra warning and checkbox to check. So in core, tag that scenario ``@skipOnEncryption``. Then the encryption app can easily skip it.

The encryption app can be responsible for implementing its own test scenario to check the different behavior that it introduces.

## How Has This Been Tested?
CI in encryption will know, when this skip tag gets merged.

## Types of changes
- [x] Acceptance test infrastructure

## Checklist:

### DIFF
--- a/tests/acceptance/features/webUILogin/resetPassword.feature
+++ b/tests/acceptance/features/webUILogin/resetPassword.feature
@@ -12,7 +12,6 @@ So that I can login to my account again after forgetting the password
 		And the user has browsed to the login page
 		And the user logs in with username "user1" and invalid password "invalidpassword" using the webUI
 		
-
 	Scenario: send password reset email
 		When the user requests the password reset link using the webUI
 		Then a message with this text should be displayed on the webUI:
@@ -24,7 +23,8 @@ So that I can login to my account again after forgetting the password
 			Use the following link to reset your password: <a href=
 			"""
 
-	Scenario: reset password
+	@skipOnEncryption
+	Scenario: reset password for the ordinary (no encryption) case
 		When the user requests the password reset link using the webUI
 		And the user follows the password reset link from email address "u1@oc.com.np"
 		Then the user should be redirected to a webUI page with the title "ownCloud"


phil-davis:
I tagged this QA because it would be good to get this in place straight away so we can get the encryption app tests running automatically and passing.

@patrickjahns @individual-it if you are happy with this, please approve (and the backport) - so it gets in the earliest nightly build...

patrickjahns (approved):
👍 LGTM

phil-davis:
In webUI we used camelCase. In API tests there is underscore.
Choose one standard, and we can easily patch the other - e.g. in API tests we would add equivalent camelCase tags, changing scripts in core that use them, then change scripts in apps, then remove the old underscore tags in core.

phil-davis:
> However - I am not sure that this is a functionality only available with encryption app

Yes, the tagging system in core might need more thought for the future! Actually multiple apps might have reason to override certain "components"  of core. (notifications does, encryption does...) 

So we might find a better way in the longer term that we can tag logical core functionality like ``@resetPasswordBehavior`` ``@notificationBehavior`` ```@loginBehavior``... Then an app that overrides that part of core behavior can easily skip the tests for that core behavior, and implement its own tests. But it needs some thought, because often the app will only override a bit of the behavior, and some (or most)  of the core tests still work.

phil-davis:
Backport ``stable10`` #31809

PR #31811, opened 2018-06-18
[master] Let CopyPlugin fall back to default impl if no ICopySource

DeepDiver1975:
ICopySource is the new behavior introduced for copying versions without
deleting the destination.

To be able to preserve the old behavior for regular files, especially in
regards to permissions, we need to fall back to the old behavior instead
of trying to replicate it partly in the plugin.

Eventually we will move the regular DAV file Node implementation to
support ICopySource as well. This will likely trigger some change of
behaviors.

port of #31805


PVince81 (approved):
:+1:

PVince81:
fixed

PR #31816, opened 2018-06-18
Can share capability

phil-davis:
## Description
1) Add a capability to  the capabilities endpoint ``can_share``
2) Report ``true`` if the user is able to share files/folders (possibly restricted by the other general ``files_sharing`` settings).
3) Report ``false`` if the user is not able/allowed to share files/folders. Currently this would be because the user is in a group that is excluded from sharing, or sharing is completely disabled. In future there might be other settings that also contribute to the final decision as to if ``can_share`` is ``true``  or ``false```.

## Related Issue
#31817

Derived from issue #31576 and PR #31580

## Motivation and Context
It is nice if clients can find out if the current user is able to share. For example, a client could examine the other capabilities and see that sharing is enabled etc. But the user could be in a group that is excluded from sharing, so actually for that user there is no point in the client letting them attempt to share stuff.

## How Has This Been Tested?
Acceptance tests run locally.

During running the acceptance tests locally, the last scenario ``capabilities.feature:559`` was failing. I discovered that function ``isSharingDisabledForUser()`` in ``lib/private/legacy/util.php`` had not had the new logic put in it. See 2nd commit that makes that fix.

## Types of changes

## Checklist:


PVince81 (approved):
:+1: thanks

phil-davis:
Backport ``stable10`` #31824

PR #31673, opened 2018-06-07
Add Roave Security Advisories as a development dependency

settermjd:
## Description

Add [Roave Security Advisories](https://github.com/Roave/SecurityAdvisories) as a development dependency to block packages with known security vulnerabilities being added to the project. 

## Related Issue

n/a

## Motivation and Context

The intent of this PR is to continue helping to increase the security posture and quality of the project. By not being able to include any package with a known security vulnerability, then the quality of the code, everyone's confidence in it, and the peace of mind about it will improve. 

## How Has This Been Tested?

n/a

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #31818, opened 2018-06-18
[stable10] Add Roave Security Advisories as a development dependency

DeepDiver1975:
backport of #31673


PVince81 (approved):
:+1: assuming this only affects dev time

PVince81:
moving to next as there is currently no certainty that this will not affect the current lib versions at build time

DeepDiver1975:
> moving to next as there is currently no certainty that this will not affect the current lib versions at build time

what will happen worst case is that make will fail on composer install because we use a vulnerable dependency. A good thing from my pov.

PVince81:
does it mean future builds of old OC versions could fail if new descovered vulnerabilities have been found ?

DeepDiver1975:
> does it mean future builds of old OC versions could fail if new discovered vulnerabilities have been found ?

no because this is a pure dev dependency

PR #31819, opened 2018-06-18
Exclude users from sharing if they are a member of any excluded group - util.php

phil-davis:
## Description
Apply changes from PR #31490 to the "copy"  of the logic that (unfortunately) is in ``lib/private/legacy/util.php`` ``isSharingDisabledForUser()``

## Related Issue
https://github.com/owncloud/enterprise/issues/2508

## Motivation and Context
Fix all copies of this code.
Discovered while doing the code for issue #31817 PR #31816 

## How Has This Been Tested?
Unit test change.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #31822, opened 2018-06-18
[stable10] Exclude users from sharing if they are a member of any excluded group - util.php

phil-davis:
Backport #31819


PVince81 (approved):
:+1:

PR #31823, opened 2018-06-18
Remove old unused private utility method isSharingDisabledForUser

PVince81:
See https://github.com/owncloud/core/pull/31819#issuecomment-398028939

PR #31824, opened 2018-06-18
[stable10] Add can_share to capabilities

phil-davis:
Backport #31816 and #31823 

Provides an extra item in capabilities. Completely backward-compatible with existing capabilities. This could go in after 10.0.9 is done.

Removes old unused private utility method isSharingDisabledForUser.


PVince81 (approved):
:+1:

PR #31825, opened 2018-06-18
allow edge to gc chunks

butonic:
Edge has a hard 5GB memory limit. currently chunked uploading seems to use one indirection too many so edge will never garbage collect the chunks Blob objects.

Unsetting them manually after a chunk is completed is not enough. The memory tab in edge won't show you the objects on the heap, but overall memory consumption will still increase.

leak

This pr allows cs to happen by removing the actual data before passing the options to the anonymous handler: https://github.com/owncloud/core/compare/update-jqfu-fix-edge-oom?expand=1#diff-9e1e26ed9e6e26bd68c321b010edc0d9R772

gc

The other changes stem from a bump to the latest version, which has some cool things like https://github.com/blueimp/jQuery-File-Upload/pull/3486 that would allow us to use dynamic chunk sizes based on the network bandwith.

- [ ] create PR against upstream
- [ ] use proper dependency management

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -1310,6 +1310,9 @@ OC.Uploader.prototype = _.extend({
 					// reset retries
 					upload.data.retries = 0;
 				});
+				fileupload.on('fileuploadchunkdone', function(e, data) {
+					$(data.xhr().upload).unbind('progress');
+				});
 				fileupload.on('fileuploaddone', function(e, data) {
 					var upload = self.getUpload(data);
 					upload.done().then(function() {


butonic:
maybe this even prevents ie from freezing as well.

butonic:
commited upstream: https://github.com/blueimp/jQuery-File-Upload/pull/3508

PVince81:
seems this PR changed approach completely? @butonic 

did the new approach work ?

at least it looks much less intrusive!

butonic:
It works for EdgeHTML 17 (the current stable version for Win 10). It does not work for EdgeHTML 15. Which is what is being used. Still an improvement. Without this Edge will eat memory the size of every upload and restart the process when 5GB have been reached. You can monitor the memory being eaten live in the dev tools memory tab. https://www.browserstack.com/test-on-microsoft-edge-browser allows you to test EdgeHTML 15 and 17 in your browser for free.

PVince81:
We could still put this in though ?

butonic:
Yes, should be put in. Prevents memory leak when uploading large file with EdgeHTML 16 or greater.

butonic:
backport to 10 needed

butonic:
failing test unrelated

phil-davis:
> failing test unrelated

caldav/carddav drone jobs changed a few days ago (last week?). A rebase here will pick all that up and the jobs will pass.

PVince81 (approved):
:+1: looks safe, merging

PVince81:
@butonic please backport

PVince81:
stable10: https://github.com/owncloud/core/pull/31884

PR #31837, opened 2018-06-20
Provide a pre-commit hooks which disallows commit if the code style f…

DeepDiver1975:
…ails

## Description
Committing without having valid php code style in place is pointless.
This pre-commit hook runs make test-php-style and disallows to commit if the code style is wrong.

Please note that this does not automatically install the pre-commit hook. Each and every user has to locally install the hook script

```sh
cd .git/hooks
ln -s ../../build/git-hooks/pre-commit.sh pre-commit
```

screenshot from 2018-06-20 08-35-45
screenshot from 2018-06-20 08-36-06

## Types of changes

## Checklist:

### DIFF
--- a/.githooks/pre-commit.sh
+++ b/.githooks/pre-commit.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+#
+# Installation:
+#  cd .git/hooks
+# ln -s ../../.githooks/pre-commit.sh pre-commit
+
+make test-php-style


patrickjahns:
Please use `.githooks` or `.hooks` in the root folder. Using `build` as a dump all folder is not suitable.

(needs cleaning up from my pov instead of having more permanent files added).

Please add this to the documentation - alternatively instead of committing this into the repository, add it to the developer documentation as best practice?

DeepDiver1975:
> Please add this to the documentation - alternatively instead of committing this into the repository, add it to the developer documentation as best practice?

I think it is more convenient to have the hook script(s) in the git repo and the developer does a simple ln into .git/hooks

PVince81:
how long does the script take ? last time I ran php-cs-fixer locally it seemed to take ages. not nice if this happens for every commit.

if the style checker can be scoped to the commit's changes then ok

PVince81 (changes requested):
style check must only run on actually modified files

phil-davis:
note: php-cs-fixer keeps a cache of stuff it checked previously, so it is much faster after the first run (assuming you didn't change 1000 files in your branch)

PVince81:
not sure if it's my env but currently when I run `make test-php-style` it shows a lot of Fs.
I'm trying to run it to the end again as the previous run had an "out of memory error" despite the allowed 1Gb.

PVince81:
image

phil-davis:
@PVince81 I have a lot less lines of ``.........`` than you. Somehow in your environment it is checking more than just the core files. 

``.php_cs.dist`` should have code in it that builds and exclude list of all the folders in ``apps`` that are not bundled in the core repo.

patrickjahns:
> I think it is more convenient to have the hook script(s) in the git repo and the developer does a simple ln into .git/hooks

talking about convenience - introduce a make rule, that looks for scripts in .githooks and symlinks them into `.git/hooks` ;-)

I leave it up 2 you guys - but before merging - please add it to documentation

DeepDiver1975:
script moved to .githooks

patrickjahns:
How about providing a simple makefile rule to include the githooks:
```
init:
  find .git/hooks -type l -exec rm {} \;
  find .githooks -type f -exec ln -sf ../../{} .git/hooks/ \;
```

Originally by https://www.viget.com/articles/two-ways-to-share-git-hooks-with-your-team/

DeepDiver1975:
> How about providing a simple makefile rule to include the githooks:

nice - wanna commit to this PR right away?

PVince81:
> .php_cs.dist should have code in it that builds an exclude list of all the folders in apps that are not bundled in the core repo.

I tired adding "apps2", "apps3" and "apps-external" to `php_cs.dist` but I still see a forest of `F`.

If you guys agree I can push this here.

PVince81:
I'll use https://github.com/owncloud/core/issues/31863#issuecomment-400575991 to continue investigating why I get errors / out of memory despite exclusion of apps2/apps3

PVince81:
my problems with the style checker have been solved already

carry on with the commit hook ?

.githooks/pre-commit.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+#
+# Installation:
+#  cd .git/hooks
```
PVince81: extra space before cd

maybe prefix all commands with a "% " as it's universally understood as "run this command"

DeepDiver1975: It is simply indented by one char - I don't care much how to doc this in here. I guess people will understand that cd and ln are commands to be used on the terminal

PVince81: fine by me, merge ?

Issue #12097, opened 2014-11-11
Asynchronous PUT


ogoffart:
This is a feature required to have reliable upload of big files.  Sometimes, the server might take a while to assemble chunks or to put them in their final destination, and the client times out which causes the file to be downloaded again.
( https://github.com/owncloud/mirall/issues/2074 )

The solution to this problem is Asynchronous PUT via a poll URL.
This is already implemented on the client and will be part of the 1.8 release.
It is working successfully in a proprietary deployment (with a custom app on the server).

The client send with the PUT a "OC-Async:1" header, indicating it supports that feature. 
The idea is that, for the last chunk, if the server has reason to believe that assembling the chunk may take more than a few seconds, it may return "202 Try LAter" instead of the usual "201 Created". Also in the response there is a header "OC-Finish-Poll" with a path to the poll url.
Example:  `OC-Finish-Poll: /remote.php/poll?pollId=123456`

The client will then query this URL, which can write some spaces on the wire to avoid timeout, and when it is finaly finished, it can leave the JSON with "etag", and "fileid" property incase of success, or an "error" property in case of failure.  

```
{ "etag": "123456789" , "fileid": "987654321" } 
```

risingfish:
I'm a little late to the conversation, but I just started using owncloud last month. I don't know this has ever been discussed for owncloud, but setting up a job queue using platform like beanstalkd (http://kr.github.io/beanstalkd/) would help facilitate async server requests. Uploaded files could be shuttled off to a temp storage area, and chunking/comparing/etc. could be triggered as part of of the upload, a cron job, or an ajax request.

kubco2:
Two milestone versions was released( since source issue) for server and client, and this is still not solved :-( although the solution exists...

Big files are not only HD movies, but for example MS outlook pst files or another databases(which can be backed up by copying backup to owncloud folder), iso images, or encrypted containers.

The worst issue with this issue is that it make loop of syncing same file forever, make traffic(mobile internet users will be not happy)

ogoffart:
I allowed myself to change the tag from Enhancement to bug, since it is really a bug from the client perspective that big files cannot be uploaded.  (This is reported a lot in the client).

Really, we need to find a solution to this problem.  This (that the client already implement) or something else.

acornspaul:
I wholeheartedly support fixing this bug/enhancement the correct way, but until that happens I think it should be better documented that you can update the timeout setting in the owncloud.cfg to give your large files more time to complete.   

[...]

PR #31851, opened 2018-06-21
[POC] Feature/introduce amd to MOVE

DeepDiver1975:
## Description
We want to avoid situations where the clients have to wait for the long MOVE operation on assembly to finish. The connection can be terminated by any reason before the server sends the response back to the client. And after that the client has no capability to query for the status of the assembly.

1. AMD - Asynchronous Method Dispatch
We return the http status code 202 right after we received the MOVE request. The client can then terminate the connection and knows that the server is working on it.
The implementation uses php's register_callback functionality to process the real MOVE *after* the response is sent to the client. register_callback can not be terminated by the system - we should be pretty save on this one .... 
The 202 response will hold a header 'OC-JobStatus-Location' which will point the client to the job status endpoint

2. Job Status Endpoint
The job status endpoint is a DAV resource in the following format: /dav/job-status/{$userId}/{$this->jobId}. It will return a json object holding the job status information.

Property | Description
---|---
status | init, started, finished, error
fileId | the file id of the target file - set only once the MOVE was successful
ETag | the ETag of the file *after* MOVE on the target location - set only once the MOVE was successfull
errorCode | http error code - in case there was an error
errorMessage | there error message - in case of an error

3. Server Side Event on the job status endpoint
The client can request the job status in simple request/response manner or request an event stream following the Server Side Event protocol. To initiate SSE handling the query parameter sse=1 is appended to the GET request. As a result the server will not terminate the connection but keep it open and stream job status changes to the client. The client is asked to close the connection once the status is success or error

4. Capabilities and LazyOps header
The server will announce to the clients via the capabilities API if this new mode of operations is available or not. A system config value is necessary to allow the admin to turn this feature explicitly on.

Furthermore in order not to break regular WebDAV client an additional http header is to be sent from the client to the server so that the server knows if this operation mode shall be used or not. 

## Todo
- [x] add capability
- [ ] add COPY support
- [ ] detailed implementation to run MOVE pre flight checks before sending 202

## Motivation and Context
Introduce a more generic async mechanism

## How Has This Been Tested?
move

```sh
deepdiver@deepdiver:~$ curl -X MOVE  http://192.168.178.44:8080/remote.php/dav/files/admin/welcome.txt -uadmin -H 'OC-LazyOps: true' -H 'Destination: http://192.168.178.44:8080/remote.php/dav/files/admin/x/welcome.txt' -v
Enter host password for user 'admin':
*   Trying 192.168.178.44...
* TCP_NODELAY set
* Connected to 192.168.178.44 (192.168.178.44) port 8080 (#0)
* Server auth using Basic with user 'admin'
> MOVE /remote.php/dav/files/admin/welcome.txt HTTP/1.1
> Host: 192.168.178.44:8080
> Authorization: Basic YWRtaW46YWRtaW4=
> User-Agent: curl/7.60.0
> Accept: */*
> OC-LazyOps: true
> Destination: http://192.168.178.44:8080/remote.php/dav/files/admin/x/welcome.txt
> 
< HTTP/1.1 202 Accepted
< Host: 192.168.178.44:8080
< Date: Thu, 21 Jun 2018 08:44:43 +0000
< Connection: close
< X-Powered-By: PHP/7.1.18-1+ubuntu18.04.1+deb.sury.org+1
< Pragma: no-cache
< Set-Cookie: oc_sessionPassphrase=YABFWiAVwlohlGBpO8j3dv9zD%2FZhAesNpb482kp8Xt9iESpBU5tVGJuvrlCwvfWH1dLz23CknJcfzaFYgxzsmfZ6gMmtAkgBKpI4IlX8NKfPf9MVBShNTtzswKsaGoG%2B; path=/; HttpOnly
< X-Frame-Options: SAMEORIGIN
< X-Robots-Tag: none
< X-Download-Options: noopen
< Set-Cookie: oc9fy8f2gwa8=n18gf5i45fkdkka1ual4c8mgvt; path=/; HttpOnly
< Set-Cookie: cookie_test=test; expires=Thu, 21-Jun-2018 09:44:40 GMT; Max-Age=3600
< Connection: close
< OC-Location: /remote.php/dav/admin/queue/d3c011a2-e640-4b94-bd7d-897aa99e9585
< 
* Closing connection 0
```

Query for the status:
```sh
deepdiver@deepdiver:~$ curl -v -uadmin http://192.168.178.44:8080/remote.php/dav/queue/admin/d3c011a2-e640-4b94-bd7d-897aa99e9585
Enter host password for user 'admin':
*   Trying 192.168.178.44...
* TCP_NODELAY set
* Connected to 192.168.178.44 (192.168.178.44) port 8080 (#0)
* Server auth using Basic with user 'admin'
> GET /remote.php/dav/queue/admin/d3c011a2-e640-4b94-bd7d-897aa99e9585 HTTP/1.1
> Host: 192.168.178.44:8080
> Authorization: Basic YWRtaW46YWRtaW4=
> User-Agent: curl/7.60.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Host: 192.168.178.44:8080
< Date: Thu, 21 Jun 2018 09:00:06 +0000
< Connection: close
< X-Powered-By: PHP/7.1.18-1+ubuntu18.04.1+deb.sury.org+1
< Pragma: no-cache
< Set-Cookie: oc_sessionPassphrase=PFsuMhAl1oNSkKVHkAKY2vxSPKtJ%2BVBMMWx1zf6%2B6O%2FQU4PBz1%2BEssbXozILMCwXwNanx%2BnuaSSooLekzD2%2BOq46tDOCN9ZEjW412Ot2s%2FD55hKzLAAfFtGqFNVlBr7Z; path=/; HttpOnly
< X-Frame-Options: SAMEORIGIN
< X-Robots-Tag: none
< X-Download-Options: noopen
< Set-Cookie: oc9fy8f2gwa8=8fbdm6ijnlah8162lc7lanf7ba; path=/; HttpOnly
< Set-Cookie: cookie_test=test; expires=Thu, 21-Jun-2018 10:00:06 GMT; Max-Age=3600
< Content-Type: application/json
< ETag: "e12309511a0155a316c6223c64511ba15b217a1a"
< Content-Length: 99
< OC-ETag: "e12309511a0155a316c6223c64511ba15b217a1a"
< 
* Closing connection 0
{"status":"finished","fileId":"00003061oc9fy8f2gwa8","ETag":"\"f3610808373e6aad584d69b4a0a068f5\""}
```

## Types of changes

## Checklist:

### DIFF
--- a/apps/dav/lib/DAV/LazyOpsPlugin.php
+++ b/apps/dav/lib/DAV/LazyOpsPlugin.php
@@ -1,0 +1,169 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\DAV;
+
+use OCA\DAV\JobStatus\Entity\JobStatus;
+use OCA\DAV\JobStatus\Entity\JobStatusMapper;
+use OCP\ILogger;
+use OCP\IURLGenerator;
+use OCP\IUserSession;
+use OCP\Shutdown\IShutdownManager;
+use Sabre\DAV\Exception;
+use Sabre\DAV\Server;
+use Sabre\DAV\ServerPlugin;
+use Sabre\DAV\UUIDUtil;
+use Sabre\HTTP\RequestInterface;
+use Sabre\HTTP\Response;
+use Sabre\HTTP\ResponseInterface;
+
+/**
+ * Class LazyOpsPlugin
+ *
+ * @package OCA\DAV\DAV
+ */
+class LazyOpsPlugin extends ServerPlugin {
+
+	/** @var Server */
+	private $server;
+	/** @var string */
+	private $jobId;
+	/** @var JobStatus */
+	private $entity;
+	/** @var IUserSession */
+	private $userSession;
+	/** @var IURLGenerator */
+	private $urlGenerator;
+	/** @var IShutdownManager */
+	private $shutdownManager;
+	/** @var ILogger */
+	private $logger;
+	/** @var JobStatusMapper */
+	private $mapper;
+
+	public function __construct(IUserSession $userSession,
+								IURLGenerator $urlGenerator,
+								IShutdownManager $shutdownManager,
+								JobStatusMapper $jobStatusMapper,
+								ILogger $logger) {
+		$this->userSession = $userSession;
+		$this->urlGenerator = $urlGenerator;
+		$this->shutdownManager = $shutdownManager;
+		$this->logger = $logger;
+		$this->mapper = $jobStatusMapper;
+	}
+
+	/**
+	 * @param Server $server
+	 */
+	public function initialize(Server $server) {
+		$this->server = $server;
+		$server->on('method:MOVE', [$this, 'httpMove'], 90);
+	}
+
+	/**
+	 * @param RequestInterface $request
+	 * @param ResponseInterface $response
+	 * @return bool
+	 * @throws Exception\NotAuthenticated
+	 */
+	public function httpMove(RequestInterface $request, ResponseInterface $response) {
+		if (!$request->getHeader('OC-LazyOps')) {
+			return true;
+		}
+
+		$this->jobId = UUIDUtil::getUUID();
+		$this->setJobStatus([
+			'status' => 'init'
+		]);
+		$userId = $this->getUserId();
+		$location = $this->urlGenerator
+				->linkTo('', 'remote.php') . "/dav/job-status/{$userId}/{$this->jobId}";
+
+		$response->setStatus(202);
+		$response->setHeader('Connection', 'close');
+		$response->setHeader('OC-JobStatus-Location', $location);
+
+		$this->shutdownManager->register(function () use ($request, $response) {
+			return $this->afterResponse($request, $response);
+		}, IShutdownManager::HIGH);
+
+		return false;
+	}
+
+	public function afterResponse(RequestInterface $request, ResponseInterface $response) {
+		if (!$request->getHeader('OC-LazyOps')) {
+			return true;
+		}
+
+		\flush();
+		$request->removeHeader('OC-LazyOps');
+		$responseDummy = new Response();
+		try {
+			$this->setJobStatus([
+				'status' => 'started'
+			]);
+			$this->server->emit('method:MOVE', [$request, $responseDummy]);
+
+			$this->setJobStatus([
+				'status' => 'finished',
+				'fileId' => $response->getHeader('OC-FileId'),
+				'ETag' => $response->getHeader('ETag')
+			]);
+		} catch (\Exception $ex) {
+			$this->logger->logException($ex);
+
+			$this->setJobStatus([
+				'status' => 'error',
+				'errorCode' => $ex instanceof Exception ? $ex->getHTTPCode() : 500,
+				'errorMessage' => $ex->getMessage()
+			]);
+		}
+		return false;
+	}
+
+	private function setJobStatus(array $status) {
+		if ($this->entity === null) {
+			$userId = $this->getUserId();
+
+			$this->entity = new JobStatus();
+			$this->entity->setStatusInfo(\json_encode($status));
+			$this->entity->setUserId($userId);
+			$this->entity->setUuid($this->jobId);
+			$this->mapper->insert($this->entity);
+		} else {
+			$this->entity->setStatusInfo(\json_encode($status));
+			$this->mapper->update($this->entity);
+		}
+	}
+
+	/**
+	 * @return string
+	 * @throws Exception\NotAuthenticated
+	 */
+	private function getUserId() {
+		$user = $this->userSession->getUser();
+		if ($user === null) {
+			throw new Exception\NotAuthenticated();
+		}
+		return $user->getUID();
+	}
+}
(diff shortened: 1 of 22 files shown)


PVince81:
> We cannot re-use the Location header

try with "Content-Location" then. we did that for other DAV APIs

PVince81:
it's not strictly the entity's location, so maybe call it "Queue-Location" or "OC-Queue-Location" ?

apps/dav/lib/DAV/LazyOpsPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+	 * @param Server $server
+	 */
+	public function initialize(Server $server) {
+		$this->server = $server;
+		$server->on('method:MOVE', [$this, 'httpMove'], 90);
+//		$server->on('afterResponse', [$this, 'afterResponse']);
+	}
+
+	public function httpMove(RequestInterface $request, ResponseInterface $response) {
+		if (!$request->getHeader('OC-LazyOps')) {
```
PVince81: we could call this "Queue" or something referring to the queue, so we have the same terminology everywhere

apps/dav/lib/DAV/LazyOpsPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+		$this->setJobStatus([
+			'status' => 'init'
+		]);
+		$userId = \OC::$server->getUserSession()->getUser()->getUID();
+		// TODO: url decode ?
+		$location = \OC::$server->getURLGenerator()->linkTo('', 'remote.php') . "/dav/{$userId}/queue/{$this->jobId}";
+
+		$response->setStatus(202);
+		$response->addHeader('Connection', 'close');
+		$response->addHeader('OC-Location', $location);
```
PVince81: OC-Queue-Location ?

apps/dav/lib/DAV/LazyOpsPlugin.php
DeepDiver1975: absolutly - I'm not yet fully convincent with the resource name 'queue' as well - since it is not a queue ....

apps/dav/lib/DAV/LazyOpsPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+		\flush();
+		$request->removeHeader('OC-LazyOps');
+		$responseDummy = new Response();
+		try {
+			$this->setJobStatus([
+				'status' => 'started'
+			]);
+			$this->server->emit('method:MOVE', [$request, $responseDummy]);
+
+			$this->setJobStatus([
```
PVince81: if we can store the final file name in there we could solve a lot of problems, like the fact that encryption relies on the part file name and uses magic extraction to find the final file name... here it could just read from the upload queue

speaking of which... does this work with user-key encryption ? cron doesn't have the user's private key so cannot encrypt...

PVince81 (commented):
interesting concept

The queue could also be used to track stray/stuck uploads and clean up chunks afterwards. This would require using the queue earlier when dealing with chunks. Or needs a completely separate mechanism.

apps/dav/lib/DAV/LazyOpsPlugin.php
DeepDiver1975: the execution is not done in cron - but in the original client request.

The difference is that we defer the execution to PHP's shutdown handler which is called after the response is sent back to the client

ogoffart:
Good. This more or less follow the design that was already done in the client and documented in https://github.com/owncloud/core/issues/12097  so implementation in the client can re-use that. (although that code sort of bit-rotted and was not ported to new chunking)

apps/dav/lib/Queue/LongPollPlugin.php
```diff
@@ -0,0 +1,93 @@
...
+				\ob_flush();
+				\flush();
+
+				$etag = $node->getETag();
+				$node->refreshStatus();
+			} else {
+				\sleep(1);
+			}
+
+			if (connection_aborted()) {
```
tomneedham: \

apps/dav/lib/DAV/LazyOpsPlugin.php
```diff
@@ -0,0 +1,126 @@
...
+ */
+class LazyOpsPlugin extends ServerPlugin {
+
+	/** @var Server */
+	private $server;
+	/** @var string */
+	private $jobId;
+
+	public static function getQueueInfo(string $userId, string $jobId) {
+		return \OC::$server->getConfig()->getUserValue($userId, 'dav', "lazy-ops-job.$jobId", null, false);
```
butonic: I assume you are going to replace this with a proper queue table ...

butonic: or use a file property ... config is really not the place for this ....

DeepDiver1975:
In order to detect that the connection was closed by the client the server needs to send something over the wire.

apps/dav/lib/DAV/LazyOpsPlugin.php
DeepDiver1975: sure - this was for the sake of fast POC

tomneedham: Could you use the existing oc_jobs? Insert as a 'already reserved' job, and add a status column?

DeepDiver1975: > Could you use the existing oc_jobs? Insert as a 'already reserved' job, and add a status column?

hmmmm - let me think about this ....

DeepDiver1975:
> In order to detect that the connection was closed by the client the server needs to send something over the wire.

handled by the used library https://github.com/hhxsv5/php-sse/releases/tag/1.1.4

apps/dav/lib/DAV/LazyOpsPlugin.php
DeepDiver1975: I don't think so - there is no way to select the job status by it's id where the is is a uuid .... 
... and also the processing logic will not work.

PVince81:
Fixes https://github.com/owncloud/core/issues/31880

micbar:
@PVince81 needs backport for 10, see https://github.com/owncloud/enterprise/issues/2480

DeepDiver1975:
> @PVince81 needs backport for 10, see owncloud/enterprise#2480

We have to analyse the impact of SSE to the system. Following the semver definitions this is nothing to be added to 10.0.x but to 10.1

apps/dav/lib/Server.php
```diff
@@ -86,6 +90,17 @@ public function __construct(IRequest $request, $baseUri) {
 		$tree = new \OCA\DAV\Tree($root);
 		$this->server = new \OCA\DAV\Connector\Sabre\Server($tree);
 
+		$config = \OC::$server->getConfig();
+		if ($config->getSystemValue('dav.enable.async', true)) {
```
PVince81: for this backport version I think we should disable this by default ?

PR #31853, opened 2018-06-21
Fixes #31849 - delete thumbnails on write hooks

### LINKED ISSUE #31849: Thumbnail not regenerated as file changes
### Steps part 1

1. Login as "someuser"
1. Create a file "test.txt" and edit several times to create versions
1. Open the "Versions" tab in the sidebar
1. Restore some previous version
1. Close the "Versions" panel, even refresh the whole files page
1. Look at the thumbnail preview

### Expected results
Thumbnail shows content from the restored version.

### Actual result
Thumbnail shows unchanged content.

### Version
10.0.9RC1

### Steps part 2
There is no thumbnail update when doing upload-overwrite:
1. Upload some x.png file - observe the thumbnail looks like a small version of the file content
1. Locally edit/replace the x.png file with some other graphics content
1. Upload x.png and choose to keep the new file
1. Download x.png - check that is does have the new content

### Expected results
Thumbnail displays updated content

### Actual result
Thumbnail shows unchanged content.

### Steps part 3
There is no thumbnail update when editing and existing text file:
1. Create a new text file with some content - observe that the thumbnail shows the early content
1. Edit the text file in the webUI, changing some content early on line 1 (to make sure it should be part of the thumbnail)

### Expected results
Thumbnail displays updated content

### Actual result
Thumbnail shows unchanged content.

DeepDiver1975:
## Description
Thumbnails only got deleted on delete - not for write events.
On write thumbnails need to be deleted so that a future GET will recreate them as needed.

## Related Issue
Fixes #31849

## How Has This Been Tested?
- manually as described in #31849

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Preview.php
+++ b/lib/private/Preview.php
@@ -1306,11 +1306,22 @@ class Preview {
 	public static function post_delete($args, $prefix = '') {
 		$path = Files\Filesystem::normalizePath($args['path']);
 		if (!isset(self::$deleteFileMapper[$path])) {
-			return;
+			$user = isset($args['user']) ? $args['user'] : \OC_User::getUser();
+			if ($user === false) {
+				$user = Filesystem::getOwner($path);
+			}
+
+			$userFolder = \OC::$server->getUserFolder($user);
+			if ($userFolder === null) {
+				return;
+			}
+
+			$node = $userFolder->get($path);
+		} else {
+			/** @var FileInfo $node */
+			$node = self::$deleteFileMapper[$path];
 		}
 
-		/** @var FileInfo $node */
-		$node = self::$deleteFileMapper[$path];
 		$preview = new Preview($node->getOwner()->getUID(), $prefix, $node);
 		$preview->deleteAllPreviews();
 	}


PVince81:
@DeepDiver1975 did you find out why it worked in v10.0.8 ? Are we not triggering the same hooks with the new versions code ?

DeepDiver1975:
> @DeepDiver1975 did you find out why it worked in v10.0.8 ? Are we not triggering the same hooks with the new versions code ?

I completly rewired the hook processing and missed this case. Hail to stupid hooks and the lack of tests .....

lib/private/Preview.php
```diff
@@ -1306,11 +1306,22 @@ public static function post_delete_versions($args) {
 	public static function post_delete($args, $prefix = '') {
 		$path = Files\Filesystem::normalizePath($args['path']);
 		if (!isset(self::$deleteFileMapper[$path])) {
-			return;
+			$user = isset($args['user']) ? $args['user'] : \OC_User::getUser();
+			if ($user === false) {
+				$user = Filesystem::getOwner($path);
```
PVince81: what user do we receive here in the following scenario:

1. admin sets up system-wide ext storage "/sftp" with sharing enabled
2. user1 shares folder "sftp/test" with public link
3. log out
4. anonymous user opens link and upload+overwrites file

There is a slight risk that you'll get null here as the storage has no owner since it's system-wide. In other code paths we somehow hackily resolve this to the sharer.

PVince81 (commented):
unit tests possible ?

lib/private/Preview.php
DeepDiver1975: let me return in case $user is null and we will see how far we get

PVince81:
@phil-davis are these test scenarios also a regression ?

phil-davis:
Hmmm - in 10.0.8 the scenario is "reversed". When "user2" edits the file in the received share, the thumbnail for "user2" is updated. But the thumbnail for "user1" (the owner) stays out-of-date.

In 10.0.9RC1 nether user sees an updated thumbnail.

So you can choose which set of symptoms you "like" - 10.8 10.0.9RC1 or this PR - they are all different.

phil-davis:
I guess there is some underlying issue that is allowing the possibility for a file to have a thumbnail generated and stored multiple times, under different users. If that is so, then I guess it is a whole other can-of-worms that has been there "a long time".

lib/private/Preview.php
PVince81: sounds good

phil-davis: Let me know when you are finished editing the "real code" and I will quickly re-run the 3 scenarios in the issue report.
After that, you can write unit tests all you like, and it won't make a difference to the run-time behavior.

DeepDiver1975: There is no need in checking if Filesystem::getOwner is returning a valid user - it will always be a user (maybe not the correct on - but this is a different story).
An exception will be thrown if there is no owner - so we will stop execution of the hook listener.

no further changes from my pov - let's merge this and I take care about the backport

phil-davis (approved):
From a QA-testing point-of-view this is good.
All 3 scenarios detailed in the issue work with this code.

PVince81 (approved):
:+1:

phil-davis:
Backport ``stable10`` #31865

PR #31857, opened 2018-06-21
Introduce IShutdownManager which allows us to order the shutdown func…

DeepDiver1975:
…tions
## Description
php's register_shutdown_function offers no capability to define an execution order.
The IShutdownManager allows this.

## Related Issue
#31851 and #31758 require this

## Motivation and Context
We need a way to define that some shutdown functions are executed before the cleanup routines are triggered.

## How Has This Been Tested?
- manually

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Shutdown/ShutDownManager.php
+++ b/lib/private/Shutdown/ShutDownManager.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Shutdown;
+
+use OCP\Shutdown\IShutdownManager;
+
+class ShutDownManager implements IShutdownManager {
+
+	/** @var array[] */
+	private $callbacks = [];
+
+	/**
+	 * ShutDownManager constructor.
+	 *
+	 * @codeCoverageIgnore
+	 */
+	public function __construct() {
+		\register_shutdown_function([$this, 'run']);
+	}
+	
+	/**
+	 * Any kind of cleanup should be added with priority LOW.
+	 * Anything which shall be processed before the cleanup routes better
+	 * uses a value smaller then LOW
+	 *
+	 * @param \Closure $callback
+	 * @param int $priority - the lower the number the higher is the priority
+	 * @return void
+	 * @since 11.0.0
+	 */
+	public function register(\Closure $callback, int $priority = self::LOW): void {
+		if (!isset($this->callbacks[$priority])) {
+			$this->callbacks[$priority] = [];
+		}
+		$this->callbacks[$priority][] = $callback;
+	}
+
+	public function run(): void {
+		\ksort($this->callbacks);
+		foreach ($this->callbacks as $callbacks) {
+			foreach ($callbacks as $callback) {
+				$callback();
+			}
+		}
+	}
+}
--- a/lib/public/Shutdown/IShutdownManager.php
+++ b/lib/public/Shutdown/IShutdownManager.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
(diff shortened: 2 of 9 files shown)


PVince81 (commented):
Great idea.

I assume you grepped for every `destruct` call to find these ?

There's also the "fclose detection" wrapper which is likely unrelated to this and doesn't need touching.

lib/private/Shutdown/ShutDownManager.php
```diff
@@ -0,0 +1,60 @@
...
+	 */
+	public function register(\Closure $callback, int $priority = self::LOW): void {
+		if (!isset($this->callbacks[$priority])) {
+			$this->callbacks[$priority] = [];
+		}
+		$this->callbacks[$priority][] = $callback;
+	}
+
+	public function run(): void {
+		foreach ($this->callbacks as $callbacks) {
```
jvillafanez: This doesn't ensure a priority order according to the $priority variable:
```
php > $items = [];
php > $items[500] = 500;
php > $items[30] = 30;
php > $items[100] = 100;
php > foreach ($items as $item) { echo "$item\n";}
500
30
100
```

lib/private/Shutdown/ShutDownManager.php
```diff
@@ -0,0 +1,60 @@
...
+	 * Any kind of cleanup should be added with priority LOW.
+	 * Anything which shall be processed before the cleanup routes better
+	 * uses a value smaller then LOW
+	 *
+	 * @param \Closure $callback
+	 * @param int $priority - the lower the number the higher is the priority
+	 * @return void
+	 * @since 11.0.0
+	 */
+	public function register(\Closure $callback, int $priority = self::LOW): void {
```
jvillafanez: I'd prefer to use an interface instead of a closure. It should be easiear to test and also clarifies how tha callback should look like.

lib/public/Shutdown/IShutdownManager.php
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Interface IShutdownManager
+ *
+ * @package OCP\Shutdown
+ * @since 11.0.0
+ */
+interface IShutdownManager {
+
+	public const LOW = 1000;
```
jvillafanez: I'd prefer to define now some additional priorities to minimize the changes we'll need to do to the interface in the future.

lib/private/Shutdown/ShutDownManager.php
```diff
@@ -0,0 +1,60 @@
...
+
+use OCP\Shutdown\IShutdownManager;
+
+class ShutDownManager implements IShutdownManager {
+
+	/** @var array[] */
+	private $callbacks;
+
+	public function __construct() {
+		\register_shutdown_function([$this, 'run']);
```
jvillafanez: I'm undecided... This registration should be done as early as possible, so in the constructor is fine, but it also make sense to have this registration out of the constructor to make the tests more isolated (you might want to skip the registration but still test the rest of the methods); in this case the registration should be done by the factory / container, which seems a bit inconvenient.

lib/private/Shutdown/ShutDownManager.php
DeepDiver1975: Damn ..... Thx

lib/private/Shutdown/ShutDownManager.php
jvillafanez: For testing, it feels easier and cleaner to mock an interface even if you only need to check if the target method is being called. If you use a closure you'll need to create a custom implementation and use an external counter to test the same thing.

In addition, you'll need to document how the callback should be. You can use
```
function ($a = null, $b = null) {
  if ($a !== null) $a->doSomething();
  if ($b !== null) $b->doOtherThing();
}
```
That would be confusing on the long run because it seems to work, but it might not do what we expect.

PVince81: from my understanding with a closure all you need to do for testing is passing a test function:

```
   $obj->register(function() use (&$called) { $called = true; });
   $obj->doYourThingWithAllClosures();
   $this->assertTrue($called);
```

lib/private/Shutdown/ShutDownManager.php
DeepDiver1975: as soon as I add unit tests I'll readjust this - I mainly wanted to get feed back on the concept in this early phase.

DeepDiver1975:
> I assume you grepped for every destruct call to find these ?

I only changed/grepped for usages of register_shutdown_function. destroy methods are called even later in the process from my understanding

jvillafanez:
Quoting official documentation:
> The destructor method will be called as soon as there are no other references to a particular object, or in any order during the shutdown sequence.

We shouldn't need to care about the destructor code as long as it's properly coded. The purpose of the destructor is a bit different.

jvillafanez:
If we aren't going to use an interface shutdownManager registration, we should document what is the expected signature of the closure function, whether it is going to be `function() : void` or `function(MyClass $obj) : void` or `function() : bool`. I don't want to be forced to check how other people is using this.

We should also document that under the same priority, the tasks will be done in the same order they were registered.

DeepDiver1975:
@jvillafanez your objections have been addressed

jvillafanez (approved):

lib/private/Shutdown/ShutDownManager.php
nonasoumy: $obj->register(function() use (&$called) { $called = true; });
   $obj->doYourThingWithAllClosures();
   $this->assertTrue($called);

Issue #31863, opened 2018-06-21
PHP code style check is checking data folder and apps2 and apps3 folders


phil-davis:
### Steps to reproduce
1. install development ownCloud in local git repo with the data folder in ``data`` of the local repo
2. as some user, upload some *.php code with crap style
3. ``make test-php-style``

### Expected behaviour
Style checks pass

### Actual behaviour
Style check fails on the file(s) that are actually just test user data, not real ownCloud code

### Server configuration
Local development environment.

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/781 (php-ftp check), https://github.com/owncloud/core/issues/24025 (OCI.php wrong Oracle privileges check), https://github.com/owncloud/core/issues/19036 (Code Integrity Check), https://github.com/owncloud/core/issues/29340 (Code Integrity check), and https://github.com/owncloud/core/issues/29339 (Code Integrity Check).

phil-davis:
If you put stuff in folders like ``apps2`` or ``apps3`` etc, then that gets scanned for code style also. We should exclude folders like that.

phil-davis:
PR #31866 excludes the ``data`` folder.

phil-davis:
PR #31868 excludes any ``apps*`` folders.

PVince81:
closing as https://github.com/owncloud/core/pull/31868#pullrequestreview-130913531 was closed

I tried excluding the folders locally but I'm still getting a forest of "F"

phil-davis:
PR #31868 was rejected. In that PR I was trying to make the excluded folders for the PHP code style checks match more closely to what is excluded in ``.gitignore`` It seems to me that if we provide a repo-wide ``.gitignore`` which excludes ``apps*`` (``apps2`` ``apps3`` ...) from being considered for inclusion in commits, then it is logical to have a similar exclusion from the PHP  code style and any other checks/parsing of "real code".

PVince81:
I'm reopening this because even with `.php_cs` it doesn't work for me.

I tried dumping the array:
```
lib/composer/bin/php-cs-fixer fix -v --diff --diff-format udiff --dry-run --allow-risky yes
array(12) {
  [0]=>
  string(12) "lib/composer"
  [1]=>
  string(5) "build"
  [2]=>
  string(28) "apps/files_external/3rdparty"
  [3]=>
  string(6) "config"
  [4]=>
  string(4) "data"
[...]

PR #31868, opened 2018-06-21
Exclude apps* from PHP style checks

phil-davis:
## Description
Exclude any folders called ``apps*`` (usually ``apps2`` ``apps3`` etc) from being checked for code style.

## Related Issue
#31863 

## Motivation and Context
Devs often create extra apps folders called ``apps2`` ``apps3``  etc when testing bonus apps. Such folders are not part of core repo code, so they should not be checked by core ``make test-php-style``

## How Has This Been Tested?
1. Create folders ``apps2/aaa`` ``apps3/bbb`` ``apps99/ccc``
2. Copy a PHP code file with style errors into each of those folders
3. ``make test-php-style`` - it reports the style issues in each of those folders
4. apply this change
5. ``make test-php-style`` - it passes

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue) - to development infrastructure only

## Checklist:

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -34,6 +34,16 @@ foreach ($dirIterator as $fileinfo) {
     }
 }
 
+// Exclude any folders "apps*" except the folder "apps"
+$topDirIterator = new DirectoryIterator(__DIR__);
+
+foreach ($topDirIterator as $fileinfo) {
+    $filename = $fileinfo->getFilename();
+    if ($fileinfo->isDir() && (strncmp($filename, "apps", 4) === 0) && (strlen($filename) > 4)) {
+        $excludeDirs[] = $filename;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


phil-davis:
After code is approved here, "backport" needs to be cherry-picked into #31828 which is waiting to implement the code-style checks in ``stable10``

patrickjahns (changes requested):
👎 
do not add these configurations for everyone please

```
You may also create .php_cs file, which is the local configuration that will be used instead of the project configuration.
```
ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer

patrickjahns:
@phil-davis 
`php_cs.dist` ist a distributed file - if developers have a different need locally (as in their project setup differs from the repostiory), they are supposed to override the `.dist` file by using `.php_cs` - this is standard practice with most php toolls

PR #31869, opened 2018-06-21
Add UI for switching public mail notification language

VicDeo:
## Description
- [x] Make lang selector reusable
- [x] Reuse it for a new option in the `Sharing` section

## Related Issue
https://github.com/owncloud/enterprise/issues/2435#issuecomment-396551153

## Motivation and Context
Allows admin users change a language used for public mail notifications

## How Has This Been Tested?
0. Language selector in Settings->Personal -> General -> Language should NOT change it's behavior
1. A new option should be visible only when `Allow users to send mail notification for shared files` is checked in Settings->Admin->Sharing
2. Switching the dropdown  to  the value `Owner language` should reset to the current behavior when the notification is sent in the language of the file owner
expected:
`php occ config:app:get core shareapi_public_notification_lang` produces an empty output in this case (no value set)

3. Switching the dropdown to any other value except `Owner language`  should update `shareapi_public_notification_lang` with a respective language code

notification lang

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/Helper/LocaleHelper.php
+++ b/lib/private/Helper/LocaleHelper.php
@@ -108,8 +108,8 @@ class LocaleHelper {
 
 			// TRANSLATORS this is a self-name of your language for the language switcher
 			$endonym = (string)$l->t('__language_name__');
-			//Check if the language name is in the translation file
-			// Fallback to hardcoded language name if translation is
+			// Check if the language name is in the translation file
+			// Fallback to hardcoded language name if it isn't
 			$languageName = ($l->getLanguageCode() === $languageCode
 				&& \substr($endonym, 0, 1) !== '_'
 			) ? $endonym
--- a/lib/private/Helper/LocaleHelper.php
+++ b/lib/private/Helper/LocaleHelper.php
@@ -1,0 +1,192 @@
+<?php
+/**
+ * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Helper;
+
+use OCP\L10N\IFactory;
+
+class LocaleHelper {
+	/**
+	 * @var string[]
+	 */
+	private $commonLanguages = [
+		'en',
+		'es',
+		'fr',
+		'de',
+		'de_DE',
+		'ja',
+		'ar',
+		'ru',
+		'nl',
+		'it',
+		'pt_BR',
+		'pt_PT',
+		'da',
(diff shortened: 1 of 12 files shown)


PVince81:
as discussed, ready for review again

lib/private/Helper/LocaleHelper.php
```diff
@@ -0,0 +1,192 @@
...
+		$languages = [];
+
+		$availableCodes = $langFactory->findAvailableLanguages();
+		foreach ($availableCodes as $languageCode) {
+			$l = $langFactory->get('settings', $languageCode);
+
+			// TRANSLATORS this is a self-name of your language for the language switcher
+			$endonym = (string)$l->t('__language_name__');
+			//Check if the language name is in the translation file
+			// Fallback to hardcoded language name if translation is
```
phil-davis: if translation is ???

VicDeo: 'if it is not'

settings/Panels/Personal/Profile.php
```diff
@@ -42,81 +43,59 @@ class Profile implements ISettings {
...
-			'en', 'es', 'fr', 'de', 'de_DE', 'ja', 'ar', 'ru', 'nl', 'it', 'pt_BR', 'pt_PT', 'da', 'fi_FI', 'nb_NO', 'sv', 'tr', 'zh_CN', 'ko'
-		];
-		$languageNames = $this->getLanguageCodes();
-		$languages= [];
-		$commonLanguages = [];
-		foreach ($languageCodes as $lang) {
-			$l = $this->lfactory->get('settings', $lang);
-			// TRANSLATORS this is the language name for the language switcher in the personal settings and should be the localized version
-			$potentialName = (string) $l->t('__language_name__');
-			if ($l->getLanguageCode() === $lang && \substr($potentialName, 0, 1) !== '_') {//first check if the language name is in the translation file
```
PVince81: it looks like this complex logic got simplified in `getNormalizedLanguages` ?

please make sure you cover all these old code paths in a unit test in LocaleHelperTest

PVince81 (commented):
Looks good so far.

I'm missing unit tests for the helper. This is important due to the complexity of the code you have simplified.

PVince81:
Codecov didn't complain about missing tests for LocaleHelper because the code in question seems to be touched by the other tests. Would still be good to have a test class there to enforce (and document) the many different locale formats

phil-davis:
It would be nice to have acceptance tests for the UI settings page. I made a general issue for that, an when @paurakhsharma gets initial tests, then it will be easy to add tests for this UI element.
(This is just a note FYI, not a blocker of this PR!)

settings/Panels/Personal/Profile.php
VicDeo: basically we check if the current lang has a self-name in a translation file, than fallback to harcoded language self-names and when everything else fails use language code as a self-name.

VicDeo:
@PVince81 added some tests with explanations

lib/private/Helper/LocaleHelper.php
```diff
@@ -0,0 +1,192 @@
...
+		'tr',
+		'zh_CN',
+		'ko'
+	];
+
+	/**
+	 * @var string[]
+	 */
+	private $languageCodes = [
+		'el' => 'Ελληνικά',
```
PVince81: I wonder if we can get these from a PHP library in the future

PVince81 (approved):
:+1: looks good, thanks

PVince81:
@VicDeo please backport

mmattel:
I guess that this is documentation relevant at least becuase of:
``php occ config:app:get core shareapi_public_notification_lang``
In case, please open a doc issue and reference it

mmattel:
This language definition for eMails could be useful for eMails sent when creating a new user!
@PVince81 @VicDeo

VicDeo:
Stable10: https://github.com/owncloud/core/pull/32004

Issue #31725, opened 2018-06-12
API acceptance tests webDav.php checkFavoritedElementsPaginated()


phil-davis:
has parameters ``offset`` and ``limit`` that are never used.

The test scenario ``Get favorited elements paginated`` in ``favorites.feature`` uses this. So the test scenario is not checking what you might think it checks.

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/22492 (test), https://github.com/owncloud/core/issues/22485 (test), https://github.com/owncloud/core/issues/31579 (test), https://github.com/owncloud/core/issues/425 (webdav.php unauthorized), and https://github.com/owncloud/core/issues/31403 (Expand API acceptance tests).

phil-davis:
@paurakhsharma can you put this on your list of things to look at?

paurakhsharma:
Yes, I will check it after I am done with some API sharing tests.

phil-davis:
No huge rush, but it is nice to have tests that do what it looks like they do :)

paurakhsharma:
Yes definitely  :)

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

PR #31872, opened 2018-06-22
use unused offset and limit value in favorites test

paurakhsharma:
## Description
Previously `offset` and `limit` values were not utilized in PHP function eventhough they were being used by scenario steps. This PR makes use of those `offset` and `limit` values.

## Related Issue
#31725

## How Has This Been Tested?
Locally

## Types of changes

- [x] Acceptance test

## Checklist:

### DIFF
--- a/tests/acceptance/features/apiMain/favorites.feature
+++ b/tests/acceptance/features/apiMain/favorites.feature
@@ -103,7 +103,7 @@ Feature: favorite
 			| old           |
 			| new           |
 
-    Scenario Outline: Get favorited elements paginated
+    Scenario Outline: Get favorited elements paginated with some offset
         Given using <dav_version> DAV path
         And user "user0" has been created
         And user "user0" has created a folder "/subfolder"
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1004,14 +1004,14 @@ public function reportFolder(
 						<oc:filter-rules>
 							' . $filterRules . '
 						</oc:filter-rules>';
-		if (\is_int($offset) || \is_int($limit)) {
+		if (!empty($offset) || !empty($limit)) {
```
phil-davis: We might want test scenarios from offset "0" - and that is empty in PHP.
At this point in the code, I think $offset and $limit could be ``null`` or some real value that we want to try and use.
So maybe this can test something like ``!is_null($offset)``
and similar for the other ``!empty()`` checks.

tests/acceptance/features/apiMain/favorites.feature
```diff
@@ -120,12 +120,12 @@ Feature: favorite
         And user "user0" favorites element "/subfolder/textfile4.txt" using the API
         And user "user0" favorites element "/subfolder/textfile5.txt" using the API
         Then user "user0" in folder "/subfolder" should have favorited the following elements from offset 3 and limit 2
```
phil-davis: We could also make a scenario that tests from offset 0
That will prove the testing code, and also be interesting to see if the API works for that.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1960,20 +1959,22 @@ public function checkFavoritedElements($user, $folder, $expectedElements) {
...
+	 * @param int|null $limit
 	 * @param TableNode|null $expectedElements
-	 * @param int $offset unused
-	 * @param int $limit unused
 	 *
 	 * @return void
 	 */
 	public function checkFavoritedElementsPaginated(
-		$user, $folder, $expectedElements, $offset, $limit
+		$user, $folder, $offset, $limit, $expectedElements
```
phil-davis: That's good to fix these parameter ordering errors!

tests/acceptance/features/apiMain/favorites.feature
```diff
@@ -120,12 +120,12 @@ Feature: favorite
...
-            | /subfolder/textfile2.txt |
             | /subfolder/textfile3.txt |
+            | /subfolder/textfile4.txt |
         Examples:
-			| dav_version   |
-			| old           |
-			| new           |
+				| dav_version   |
```
phil-davis: why did this get more indent?

phil-davis (commented):
This will be good to fix. I made some comments about ``empty()`` - it is an annoying function in PHP.

tests/acceptance/features/bootstrap/WebDav.php
paurakhsharma: Oh I didn't realize that. Yeah we should use `is_int()`.

phil-davis: If you use ``is_int()`` then you will need to also cast the input parameters to ``int`` (probably where they are first received from the gherkin step text will be easiest?)

paurakhsharma: Yes, I have implemented it. But it seems there is some problems in API

PR #31884, opened 2018-06-25
[stable10] allow edge to gc chunks

PVince81:
backport of https://github.com/owncloud/core/pull/31825 to stable10


PVince81:
as discussed with @felixboehm and @pmaier1, putting this in

jvillafanez:
It feels as if we're registering several times the same event, otherwise this shouldn't be needed. It's fine for the backport, but we might need to investigate a bit further.
If this has been tested in all browsers and it didn't break anything, I'm fine merging this.

PVince81:
@davitol test this in RC2 later with all supported browser or in PR ?

PR #31915, opened 2018-06-26
Move AccountCheckException to public api

DeepDiver1975:
## Description
Missing core features as required by https://github.com/owncloud/password_policy/pull/15

## Related Issue
https://github.com/owncloud/password_policy/pull/15

## Types of changes

## Checklist:

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -450,6 +450,7 @@ class Session implements IUserSession, Emitter {
 
 			// trigger any other initialization
 			$this->eventDispatcher->dispatch(IUser::class . '::firstLogin', new GenericEvent($this->getUser()));
+			$this->eventDispatcher->dispatch('user.firstlogin', new GenericEvent($this->getUser()));
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


lib/private/User/Session.php
```diff
@@ -893,7 +894,9 @@ protected function loginUser($user, $password) {
 			}
 
 			return true;
-		}, ['before' => ['uid' => $user, 'password' => $password], 'after' => ['uid' => $user, 'password' => $password]], 'user', 'login');
+		}, ['before' => ['user' => $user, 'uid' => $uid, 'password' => $password],
```
butonic: can you fire the events in here (and maybe in other methods in this class) in the flow of the code as in `loginWithPassword`
https://github.com/owncloud/core/blob/0e38d3a88d1c46e54036aaa8fd5e0eaeac895201/lib/private/User/Session.php#L499-L533

That would make the code a lot more readable.

DeepDiver1975: EventEmitterTrait::emittingCall was developed on purpose to be used as general design pattern to emit before and after events without interfering the regular code flow.

butonic: which is the most stupid idea ever ... it makes the code completely unreadable. Even worse it leads to references like `&$user, &$password` It is too hard to understand. Get rid of it!

butonic: in a different PR

butonic (approved):

PVince81:
@DeepDiver1975 please backport to stable10, will be p2 as required by password policy

PR #31924, opened 2018-06-27
Fixing the transfer of child shares for transferShare

sharidas:
The child shares are transferred first before deleting
the parent share. If this is not done, then the child
shares are lost, which could result in loss of shares.

## Description

If the parent share is already shared to the destination user and if the parent share has child shares, then exception would occur while trying to update the child share. Because while processing the parent share, it would have been deleted. And subsequently the child shares would also be deleted.

## Related Issue

https://github.com/owncloud/core/issues/31150

## Motivation and Context

If the parent share is already shared to the destination user and if the parent share has child shares, then exception would occur while trying to update the child share. Because while processing the parent share, it would have been deleted. And subsequently the child shares would also be deleted.

## Types of changes

## Checklist:

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -27,6 +27,7 @@ namespace OCA\Files\Command;
 use OC\Encryption\Manager;
 use OC\Files\Filesystem;
 use OC\Files\View;
+use OC\Share20\ProviderFactory;
 use OCP\Files\FileInfo;
 use OCP\Files\Mount\IMountManager;
 use OCP\ILogger;
@@ -57,6 +58,9 @@ class TransferOwnership extends Command {
 	/** @var ILogger  */
 	private $logger;
 
+	/** @var ProviderFactory  */
+	private $shareProviderFactory;
+
 	/** @var bool */
 	private $filesExist = false;
 
@@ -81,12 +85,13 @@ class TransferOwnership extends Command {
 	/** @var string */
 	private $finalTarget;
 
-	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ILogger $logger) {
+	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ILogger $logger, ProviderFactory $shareProviderFactory) {
 		$this->userManager = $userManager;
 		$this->shareManager = $shareManager;
 		$this->mountManager = $mountManager;
 		$this->encryptionManager = $encryptionManager;
 		$this->logger = $logger;
+		$this->shareProviderFactory = $shareProviderFactory;
 		parent::__construct();
 	}
 
@@ -168,7 +173,7 @@ class TransferOwnership extends Command {
 		$this->transfer($output);
 
 		// restore the shares
-		$this->restoreShares($output);
+		return $this->restoreShares($output);
 	}
 
 	private function walkFiles(View $view, $path, \Closure $callBack) {
@@ -256,6 +261,7 @@ class TransferOwnership extends Command {
 				if (empty($sharePage)) {
 					break;
 				}
+
 				$this->shares = \array_merge($this->shares, $sharePage);
 				$offset += 50;
 			}
@@ -295,18 +301,47 @@ class TransferOwnership extends Command {
 	private function restoreShares(OutputInterface $output) {
 		$output->writeln("Restoring shares ...");
 		$progress = new ProgressBar($output, \count($this->shares));
+		$status = 0;
 
+		$childShares = [];
 		foreach ($this->shares as $share) {
 			try {
+				/**
+				 * Do not process children which are already processed.
+				 * This piece of code populates the childShare array
+				 * with the child shares which will be processed. And
+				 * hence will avoid further processing of same share
+				 * again.
+				 */
+				if ($share->getSharedWith() === $this->destinationUser) {
+					$provider = $this->shareProviderFactory->getProviderForType($share->getShareType());
+					foreach ($provider->getChildren($share) as $child) {
+						$childShares[] = $child->getId();
+					}
+				} else {
+					/**
+					 * Before doing handover to transferShare, check if the share
+					 * id is present in the childShares. If so then just ignore
+					 * this share and continue. If not ignored, the child shares
+					 * would be processed again, if their parent share was shared
+					 * with destination user. And hence we can safely avoid the
+					 * duplicate processing of shares here.
+					 */
+					if (\in_array($share->getId(), $childShares, true)) {
+						continue;
+					}
+				}
 				$this->shareManager->transferShare($share, $this->sourceUser, $this->destinationUser, $this->finalTarget);
 			} catch (\OCP\Files\NotFoundException $e) {
 				$output->writeln('<error>Share with id ' . $share->getId() . ' points at deleted file, skipping</error>');
 			} catch (\Exception $e) {
 				$output->writeln('<error>Could not restore share with id ' . $share->getId() . ':' . $e->getTraceAsString() . '</error>');
+				$status = 1;
 			}
 			$progress->advance();
 		}
 		$progress->finish();
 		$output->writeln('');
+		return $status;
 	}
 }
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -35,6 +35,7 @@ use OCP\Files\IRootFolder;
 use OCP\Files\Mount\IMountManager;
 use OCP\Files\NotFoundException;
 use OCP\IConfig;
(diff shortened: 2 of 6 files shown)


sharidas:
Things I have done in this version of PR:
- Process the child shares before the parent shares are getting deleted.
- The usage of query builder inside the `transferShare()` to make the child share decouple from the parent share. That is set it to null.

There is a problem with this solution, if a child share is processed, I need to add a mechanism to stop calling `transferShare` again.

sharidas:
The problem mentioned in https://github.com/owncloud/core/pull/31924#issuecomment-400660863, is rectified in this version.

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -81,12 +85,13 @@ class TransferOwnership extends Command {
...
 	private $finalTarget;
 
-	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ILogger $logger) {
+	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ILogger $logger, ProviderFactory $factory) {
 		$this->userManager = $userManager;
 		$this->shareManager = $shareManager;
 		$this->mountManager = $mountManager;
 		$this->encryptionManager = $encryptionManager;
 		$this->logger = $logger;
+		$this->factory = $factory;
```
PVince81: factory is a too generic name. please call it `shareProviderFactory` or something similar

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -295,18 +301,39 @@ protected function transfer(OutputInterface $output) {
...
 		$output->writeln("Restoring shares ...");
 		$progress = new ProgressBar($output, \count($this->shares));
+		$status = 0;
 
+		$childShares = [];
 		foreach ($this->shares as $share) {
 			try {
+				/**
+				 * Do not process children which are already processed.
+				 * This piece of code tries to populate the childShare
```
PVince81: it doesn't try, it just populates ? "try" would only apply if there are possible expected errors or exceptions

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -295,18 +301,39 @@ protected function transfer(OutputInterface $output) {
...
+				 * And hence will avoid further processing of same share
+				 * again.
+				 */
+				if ($share->getSharedWith() === $this->destinationUser) {
+					$provider = $this->factory->getProviderForType($share->getShareType());
+					foreach ($provider->getChildren($share) as $child) {
+						$childShares[] = $child->getId();
+					}
+				} else {
+					if (\in_array($share->getId(), $childShares, true)) {
```
PVince81: not fully clear what this code is for, please add a comment

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -295,18 +301,39 @@ protected function transfer(OutputInterface $output) {
 	private function restoreShares(OutputInterface $output) {
 		$output->writeln("Restoring shares ...");
 		$progress = new ProgressBar($output, \count($this->shares));
+		$status = 0;
 
+		$childShares = [];
```
PVince81: I guess this PR is not finished yet ? You don't seem to do anything with the gathered child shares ?

lib/private/Share20/Manager.php
```diff
@@ -717,6 +722,14 @@ public function transferShare(IShare $share, $oldOwner, $newOwner, $finalTarget)
 			throw new NotFoundException("The target location $finalTarget doesn't exist");
 		}
 
+		if ($isChild === true) {
+			//Set the parent to null so that we don't lose the shares after transfer
```
PVince81: hmmm, this sounds like the second approach we talked about ? so why gather child shares in the above file then ?
detaching the shares should be enough to avoid them being deleted

PVince81 (commented):
Some confusion, possibly you used one approach then changed your mind and did not clean up the old code ?

When setting "parent" to null, how do you know what parent to reattach the share to ? I guess you don't need to reattach as the "file_source" already points at the transferred file id ?

lib/private/Share20/Manager.php
sharidas: The thought which made me to use `$childShares = []`  is :
- Lets say without this code change in transfer ownership (i.e, without child shares ), and under scenario https://github.com/owncloud/core/issues/31150#issue-314806286, we could see that one of the shares will be called twice. The public link created which is the child of the folder transferred to user2, will be processed https://github.com/owncloud/core/pull/31924/files#diff-a9d8fe189a37b652ea5b821e4d772438R746. Here the public link will be decoupled from the parent. Then again the public link share would be passed in the TransferOwnership.php to the `transferShare()`. So my intention was to remove the duplicate calls to the same share. 

Basically `childShares` will be populated with the children of each share in `$this->share`. My idea is to have share ids to the array. And then while processing I will if the share id is there in the `childShares`. If yes then it wont be transferred. Because it is already transferred.

apps/files/lib/Command/TransferOwnership.php
PVince81: ah ok, so this is where you use `$childShares`
add a PHP comment to clarify

sharidas: Added comment to justify the usage of `$childShares`.

apps/files/lib/Command/TransferOwnership.php
sharidas: Removed the usage of "try".

apps/files/lib/Command/TransferOwnership.php
sharidas: Renamed `factory` to `shareProviderFactory` as suggested.

apps/files/lib/Command/TransferOwnership.php
sharidas: Tested with the latest change and made sure that scenario https://github.com/owncloud/core/issues/31150#issue-314806286 works.

sharidas:
The phpunit is failing in https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/Capabilities.php. This failure doesn't look related to this PR.

phil-davis:
``master`` was broken for a short while - rebase

sharidas:
The reason for 'codecov/patch' not passing could be because of absence of php unit test. All class variables are declared as `private` which is causing `invokePrivate` not to set the values to behave accordingly.

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -295,18 +301,44 @@ protected function transfer(OutputInterface $output) {
...
+				if ($share->getSharedWith() === $this->destinationUser) {
+					$provider = $this->shareProviderFactory->getProviderForType($share->getShareType());
+					foreach ($provider->getChildren($share) as $child) {
+						$childShares[] = $child->getId();
+					}
+				} else {
+					/**
+					 * Before doing handover to transferShare, check if the share
+					 * id is present in the childShares. If so then just ignore
+					 * this share and continue.
```
PVince81: I'm missing the "why" in this comment

sharidas: Added the reason that processing child shares would result in processing the share again, while the share was processed along with the parent share. When the parent share was shared with the `destinationUser` then the child shares are processed. And hence we can ignore the processing of child shares further, if they are already processed.

sharidas:
With the latest changes below are the tests done
- [x] The issue https://github.com/owncloud/core/issues/31150#issue-314806286 is solved
- [x] In the same test above, created 2 more users `user3` and `user4`. Now Tried to share the test folder to `user3` and `user4` and created public link for test folder for the new users. Tried to modify the parent column for the shares made for `user3` and `user4`, so that parents of initial share to `user2` be the same for `user3` and `user4`. Now ran the command. All the public link shares are accessible.
- [x] user1 creates a file which is shared with user3. Ran transfer-ownership command from user1 to user2. The shared file in user2 is still accessible.
- [x] The above test is conducted with sharing folder, and it works.
- [x] user3 is added to group1. user1 and user2 are not added to any group. user1 creates folder foo. And creates file foo/bar.txt. user1 shares foo to `group1` and `user3`. The transfer ownership command is executed from user1 to user2. The shares are available at user3. They are accessible from web ui.
- [x] Received shares are not transferred
- [x] user1 shares folder /foo/bar to user3. Delete /foo/bar from user1 to trashbin. Transfer ownership when ran from user1 to user2. It doesn't fail.
- [x] User1 creates foo/test.txt and shares `foo` to user2. Executes transfer ownership with path foo, the foo/test.txt is still available in `user2` under transferred folder.

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/31953

PR #31925, opened 2018-06-27
Prevent PHP CS scanning stray folders

### LINKED ISSUE #31863: PHP code style check is checking data folder and apps2 and apps3 folders

PVince81:
When swiching from stable9.1 the "3rdparty" folder can remain in some
cases.
Also exluded "tests/autoconfig*" which are sometimes generated when
running unit tests.

Fixes https://github.com/owncloud/core/issues/31863

Any additions must be made to ".php_cs" folder as I did for my own setup with "apps2" and "apps3"

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -4,41 +4,41 @@ $dirToParse = 'apps';
 $dirIterator = new DirectoryIterator(__DIR__ . '/' . $dirToParse);
 
 $bundledApps = [
-    'comments',
-    'dav',
-    'federatedfilesharing',
-    'federation',
-    'files',
-    'files_external',
-    'files_sharing',
-    'files_trashbin',
-    'files_versions',
-    'provisioning_api',
-    'systemtags',
-    'testing',
-    'updatenotification'
+	'comments',
+	'dav',
+	'federatedfilesharing',
+	'federation',
+	'files',
+	'files_external',
+	'files_sharing',
+	'files_trashbin',
+	'files_versions',
+	'provisioning_api',
+	'systemtags',
+	'testing',
+	'updatenotification'
 ];
 
 $excludeDirs = [
-    'lib/composer',
-    'build',
-    'apps/files_external/3rdparty',
-    'config',
-    'data',
-    '3rdparty',
+	'lib/composer',
+	'build',
+	'apps/files_external/3rdparty',
+	'config',
+	'data',
+	'3rdparty',
 ];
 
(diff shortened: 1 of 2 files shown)


.php_cs.dist
```diff
@@ -36,8 +37,9 @@ foreach ($dirIterator as $fileinfo) {
...
 $finder = PhpCsFixer\Finder::create()
     ->exclude($excludeDirs)
+    ->notPath('tests/autoconfig*')
     ->in(__DIR__);
 
 $config = new OC\CodingStandard\Config();
 $config->setFinder($finder);
-return $config;
\ No newline at end of file
+return $config;
```
patrickjahns: newline?

PVince81: We have a newline in all other PHP files so I think it's ok to have it here too for consistency.

Do you mind if I also convert all spaces to tabs as all OC's source files have tabs ?

mmattel:
Maybe important, we already have in documentation to use the name apps-external instead of apps2.
There is also an open PR that creates that directory and sets the config.php parameter on new installs.
#30889 (Add per default the apps-external directory in config.php during installation)

PVince81:
pushed another commit to convert the spaces to tabs, aligns the code style of this file with the rest of the project

PVince81:
- [ ] also add "apps-external"

patrickjahns:
🤔 Instead of excluding everything - we can go the other way around and specify explicit folders, files to include?

PVince81:
sounds good, I'll look into this

phil-davis (approved):
This looks like a good thing to me - adds a couple more exclusions and does whitespace-tab changes.

phil-davis:
@PVince81 are you going to add apps-external or not?

PVince81:
well, I actually intended to look at inclusions... but as I'm already busy with lots of other things let's move on with exclusions now

I'll add apps-external

PVince81:
added

phil-davis:
Looks good to me. Refactoring to reverse the exclude/include process can always happen in a later PR if someone thinks it is worthwhile.

PR #31930, opened 2018-06-28
[stable10] User Management UI tests

phil-davis:
Backport UI test changes that have happened in the user_management app:

1. Delete user test from https://github.com/owncloud/user_management/pull/26
2. Replace throw exception with assert in acceptance tests from https://github.com/owncloud/user_management/pull/40
3. Format and code style that happened in the user_management app
4. Make delete user acceptance test text consistent with disable user
5. Fixup addUsers feature description

This gets the code back in-sync. So it will be easier in future to make the same changes to both user_management app and to core stable10.


individual-it (approved):

phil-davis:
Tests only - adds a delete user test for ``stable10`` ... so merging.

PR #31933, opened 2018-06-28
[stable10] Move AccountCheckException to public api

DeepDiver1975:
backport #31915

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -463,6 +463,7 @@ class Session implements IUserSession, Emitter {
 
 			// trigger any other initialization
 			$this->eventDispatcher->dispatch(IUser::class . '::firstLogin', new GenericEvent($this->getUser()));
+			$this->eventDispatcher->dispatch('user.firstlogin', new GenericEvent($this->getUser()));
 		}
 	}
 
@@ -517,7 +518,6 @@ class Session implements IUserSession, Emitter {
 
 		$user = $this->manager->checkPassword($login, $password);
 		if ($user === false) {
-			$this->manager->emit('\OC\User', 'failedLogin', [$login]);
 			$this->emitFailedLogin($login);
 			return false;
 		}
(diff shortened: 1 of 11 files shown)


lib/private/User/Session.php
```diff
@@ -517,7 +518,6 @@ private function loginWithPassword($login, $password) {
 
 		$user = $this->manager->checkPassword($login, $password);
 		if ($user === false) {
-			$this->manager->emit('\OC\User', 'failedLogin', [$login]);
```
PVince81: what happened to these ? obsolete ?

lib/private/User/Session.php
```diff
@@ -463,6 +463,7 @@ public function prepareUserLogin($firstTimeLogin = false) {
 
 			// trigger any other initialization
 			$this->eventDispatcher->dispatch(IUser::class . '::firstLogin', new GenericEvent($this->getUser()));
+			$this->eventDispatcher->dispatch('user.firstlogin', new GenericEvent($this->getUser()));
```
PVince81: why needing another event ?

DeepDiver1975: to fix the naming convention

lib/private/User/Session.php
DeepDiver1975: the emit call is within emitFailedLogin

lib/private/User/Session.php
```diff
@@ -1168,11 +1170,13 @@ protected function getAuthModules($includeBuiltIn) {
...
 
 	/**
-	 * This method triggers symfony event for failed login
-	 *
+	 * This method triggers symfony event for failed login as well as
+	 * emits via the emitter in user manager
 	 * @param string $user
 	 */
 	protected function emitFailedLogin($user) {
+		$this->manager->emit('\OC\User', 'failedLogin', [$user]);
```
DeepDiver1975: @PVince81 see here

PVince81 (approved):
:+1:

PR #31939, opened 2018-06-28
Resubmit of "[stable10] generated resource routes should allow slashes""

PVince81:
Resubmit of original PR that was merged by mistake.

Don't merge due to code freeze!

### DIFF
--- a/tests/lib/Route/RouterTest.php
+++ b/tests/lib/Route/RouterTest.php
@@ -40,15 +40,8 @@ class RouterTest extends \Test\TestCase {
 	/** @var ILogger */
 	private $l;
 
-	/**
-	 * RouterTest constructor.
-	 *
-	 * @param string $name
-	 * @param array $data
-	 * @param string $dataName
-	 */
-	public function __construct($name = null, array $data = [], $dataName = '') {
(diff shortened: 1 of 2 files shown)


tests/lib/Route/RouterTest.php
```diff
@@ -24,17 +25,41 @@
...
+	private $l;
+
+	/**
+	 * RouterTest constructor.
+	 *
+	 * @param string $name
+	 * @param array $data
+	 * @param string $dataName
+	 */
+	public function __construct($name = null, array $data = [], $dataName = '') {
```
DeepDiver1975: Why in ctor? setup is to be used for initializations

DeepDiver1975 (changes requested):

tests/lib/Route/RouterTest.php
PVince81: arghh, no idea, it was like this in the original PR. I'll fix this on master first then

PVince81: here we go for master https://github.com/owncloud/core/pull/32158

already cherry-picked here

DeepDiver1975 (approved):

PR #31935, opened 2018-06-28
[stable10] return all failed recipients in sendLinkShareMailFromBody

phil-davis:
Backport #31847

PR #31940, opened 2018-06-28
Remove string parameter type from testSendLinkShareMailException parameter

phil-davis:
Forward port of 2nd commit of ``stable10`` PR #31935


PVince81 (approved):
:+1: as discussed

PR #31463, opened 2018-05-17
Allow slash in group names in provisioning API

phil-davis:
## Description
1) In the routes for groups, allow ``groupid`` to have any characters (including ``/``) except for group names that end in ``/subadmins``
2) When creating a group, do not allow group names that end in ``/subadmins`` because that would cause conflict with being able to get the subadmins of a group.
3) Add API acceptance tests to cover these new group name possibilities

## Related Issue
#31015 

## Motivation and Context
The provisioning API should be able to manage groups with a slash in their name.

## How Has This Been Tested?
Local API acceptance test runs, and trying stuff with ``curl`` commands.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #31945, opened 2018-06-29
API acceptance tests for slash in group name

phil-davis:
Run this set of new/changed tests from PR #31463 without any actual back-end code changes. 
Lets see if any of these group-names-containing-slash are working now, after other recent changes in ``master``


phil-davis:
These have been added to core ``master`` and ``stable10`` as skipped tests, so they can be unskipped some day when a fix is done. See #32430

PR #31948, opened 2018-06-29
[stable10] If token based authentication is mandatory a login failed …

DeepDiver1975:
…entry is written to the log so that fail2ban can handle this case.

backport #31909


PVince81 (approved):
:+1:

not merging because of freeze

PR #31951, opened 2018-06-29
use samaccountname from session for wnd login

butonic:
use samaccountname for wnd

### DIFF
--- a/lib/private/Files/External/LegacyUtil.php
+++ b/lib/private/Files/External/LegacyUtil.php
@@ -66,7 +66,7 @@ class LegacyUtil {
 			$mountPoint = '/'.$uid.'/files'.$storage->getMountPoint();
 			$mountEntry = self::prepareMountPointEntry($storage, false);
 			foreach ($mountEntry['options'] as &$option) {
-				$option = self::setUserVars($uid, $option);
+				$option = self::setUserVars($user->getUserName(), $option);
 			}
 			$mountPoints[$mountPoint] = $mountEntry;
 		}
@@ -75,7 +75,7 @@ class LegacyUtil {
 			$mountPoint = '/'.$uid.'/files'.$storage->getMountPoint();
 			$mountEntry = self::prepareMountPointEntry($storage, true);
 			foreach ($mountEntry['options'] as &$option) {
-				$option = self::setUserVars($uid, $option);
+				$option = self::setUserVars($user->getUserName(), $option);
 			}
 			$mountPoints[$mountPoint] = $mountEntry;
 		}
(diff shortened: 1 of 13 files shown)


PVince81:
I'd prefer "alternativeLogin"

PVince81:
found some older places we could clean up for OC 11: https://github.com/owncloud/core/issues/32192

PVince81:
I've renamed the attribute and spent an hour to add missing unit test

butonic:
The correct term would be username. But that might conflict with other use cases too easily, so I chose samaccountname because AD is actually the only use case for this and it has a clearer meaning than altloginname. Loginname can be the username or the email. But the email cannot be used to mount wnd shares. IMO stick to samaccountname.

butonic:
Nevermind, about reverting this. The problem can be solved properly with the uid->userid&username split. Then all the code in this PR is no longer needed anyway.

PVince81:
what split ? is there a ticket / PR about this ?

the reason I wanted a more generic name was because I imagined there could be more use cases where other user backends might have other properties where login name is stored and they could use the same mechanism. only in their case it wouldn't be called "samaccountname". not sure if that makes sense

butonic:
https://github.com/owncloud/core/pull/31335

PVince81:
ok, so for now we take this PR as is and can revert later once https://github.com/owncloud/core/pull/31335 is finished (which I expect will still take quite some time)

lib/private/Files/External/LegacyUtil.php
```diff
@@ -163,6 +163,9 @@ private static function prepareMountPointEntry(IStorageConfig $storage, $isPerso
 	 * @return string
 	 */
 	public static function setUserVars($user, $input) {
+		if (\OC::$server->getSession()->exists('altloginname')) {
```
PVince81: this cannot work correctly with shares.

The $user is not necessarily the user from the session. It can be the owner of the storage from which a share originates. So we'd need the samaccountname of that user but here this code would override it with the current one.

It sounds like we need a better method to persist it together with the IUser.

PVince81: there are actually two possible behaviors with this code and assumptions, both are wrong:
1) the owner of a share has a samaaccount name but the recipient doesn't: in this case this `if` condition doesn't apply. still, the userid used for substituion is the regular user id from the share owner, not their samaccount name

and
2
2) both owner and recipient have a samaccount name: here the if condition would run and the substituion would use the samaccount name from the **recipient** instead of the **owner**

PVince81:
@butonic can you update the tests ?

PVince81:
```
1) Test\User\SyncServiceTest::testSyncUserName
Trying to configure method "getUserId" which cannot be configured because it does not exist, has not been specified, is final, or is static
```

PVince81:
looks like the method is defined only through PHPDoc, and somehow PHPUnit doesn't understand this ??

https://github.com/owncloud/core/blob/bugfix/samaccountname-as-wnd-username/lib/private/User/Account.php#L30

PVince81 (approved):
:+1:

PVince81:
@butonic please backport

I expect Travis will be green soon...

PVince81:
- [ ] rebase locally and run `make test-php-style` just in case

PVince81:
squashed

PR #31958, opened 2018-06-30
[objectstorage] Cache stat while fetching objectstore object metadata

mrow4a:
### Problem Description

The same logic for calling stat in objectstore compared to filesystem implementation is inefficient.

This is due to the fact that objectstore does not implement any filesystem functionality which stores information about file, and thus needs to resort to using filecache. 

### Solution

statcache

This PR implements stat cache similar to filesystem to temporarly cache results between calls to filesystem. Each request to modify filesystem clears the stat cache. Additionaly, we store only cache for objects, not for file tree (directories).

### How did I test

I did:
- upload new file
- overwrite new file
- again overwrite (generate next version)

Master:
diagnostic-master

Obj Stat:
diagnostic-obj

This PR thus cuts 20% of read queries required for objectstore.

### Related PRs

Due to the fact there is no decision on introducing "cache of filecache", this PR is just fixing the code logic. 

- Cache in filecache (https://github.com/owncloud/core/pull/28166)
- Cache of objectstore metadata (https://github.com/owncloud/core/pull/31486)

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -59,7 +59,7 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 	private $objectPrefix = 'urn:oid:';
 
 	public function __construct($params) {
-		$this->objectStatCache = new ArrayCache();
+		$this->objectStatCache = new \OC\Cache\CappedMemoryCache();
 		if (isset($params['objectstore']) && $params['objectstore'] instanceof IObjectStore) {
 			$this->objectStore = $params['objectstore'];
 		} else {
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -26,16 +26,17 @@
 namespace OC\Files\ObjectStore;
 
 use Icewind\Streams\IteratorDirectory;
+use OC\Cache\CappedMemoryCache;
 use OC\Files\Cache\CacheEntry;
-use OC\Memcache\ArrayCache;
+use OC\Files\Storage\Common;
 use OCP\Files\NotFoundException;
 use OCP\Files\ObjectStore\IObjectStore;
 use OCP\Files\ObjectStore\IVersionedObjectStorage;
 
-class ObjectStoreStorage extends \OC\Files\Storage\Common {
+class ObjectStoreStorage extends Common {
 
 	/**
-	 * @var ArrayCache
+	 * @var CappedMemoryCache
 	 */
 	private $objectStatCache;
 
@@ -59,7 +60,7 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 	private $objectPrefix = 'urn:oid:';
 
 	public function __construct($params) {
-		$this->objectStatCache = new \OC\Cache\CappedMemoryCache();
+		$this->objectStatCache = new CappedMemoryCache();
 		if (isset($params['objectstore']) && $params['objectstore'] instanceof IObjectStore) {
 			$this->objectStore = $params['objectstore'];
 		} else {
@@ -516,7 +517,7 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 		if (!$storage) {
 			$storage = $this;
 		}
-		if (!isset($this->scanner)) {
+		if ($this->scanner === null) {
 			$this->scanner = new NoopScanner($storage);
 		}
 		return $this->scanner;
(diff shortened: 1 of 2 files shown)


PVince81:
this is likely not enough, you also need logic to invalidate the entries whenever other file operations are called.

remember that many high level file operations might involve more than a single low level FS operation

here's an example stat cache that was added for external storage: https://github.com/owncloud/core/blob/v10.0.8/lib/private/Files/Storage/DAV.php#L78 and https://github.com/owncloud/core/blob/v10.0.8/lib/private/Files/Storage/CacheableFlysystem.php#L32

mrow4a:
@PVince81 the stat cache gets set and invalidated in the same function - getMetadata. There is no other place in code to set the cache - so cache will leave only for duration of that function

PVince81:
so are we assuming that all file operations like $objectStore->unlink() and $objectStore->rename() will call `getMetadata()` again to invalidate and refresh the stat cache info ? it might be the case now but there is no guarantee it will be in the future

mrow4a:
@PVince81 so your suggestion is to have a look how stat cache is realized in other implementations?

PVince81:
> so your suggestion is to have a look how stat cache is realized in other implementations?

that or think about stat cache invalidation for every file operation that go through the storage

mrow4a:
@butonic what do you think about approach suggested by @PVince81 ?

mrow4a:
@PVince81 @DeepDiver1975 @butonic @tomneedham all tests passing - I also updated the PR description with the context.

mrow4a:
@butonic this does not touch filecache at all. It is a layer below, storage stat cache - please look a description. CC @PVince81 @DeepDiver1975

mrow4a:
@butonic and to mention further, this PR is only to make objectstore work with similar performance as filesystem, no filecache "caching" is here in place, I just cache results of the filesystem calls wrapped by objectstorage

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -502,4 +480,132 @@ public function restoreVersion($internalPath, $versionId) {
...
+		if ($this->objectStatCache->hasKey($path)) {
+			return $this->objectStatCache->get($path);
+		}
+
+		$cacheEntry = $this->getCache()->get($path);
+		if ($cacheEntry instanceof CacheEntry) {
+			$stat = $cacheEntry->getData();
+			if ($cacheEntry->getMimeType() != 'httpd/unix-directory') {
+				// Only set stat cache for objects
+				$this->objectStatCache->set($path, $stat);
```
mrow4a: @DeepDiver1975 @butonic note that I only cache stat calls for objects metadata, not a file-tree (filecache trees)

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -502,4 +480,132 @@ public function restoreVersion($internalPath, $versionId) {
...
+	 * @return null|string the unified resource name used to identify the object
+	 */
+	protected function getURN($fileId) {
+		if (\is_numeric($fileId)) {
+			return $this->objectPrefix . $fileId;
+		}
+		return null;
+	}
+
+	private function rmObjects($path) {
```
PVince81: why has this method been moved down ? this makes it difficult to review as now we can't see whether you made change to it

if you want to reformat a file please do so in a separate commit

PVince81:
I agree with @mrow4a that a stat cache on the storage class is useful to avoid repeated real-storage access as higher level OC code tends to repeat operations (like file_exists, etc). We already use this approach for other storages like DAV, Dropbox, etc.

lib/private/Files/ObjectStore/ObjectStoreStorage.php
mrow4a: You can also use diff -> split. But I can do it in separate commit, should I?

lib/private/Files/ObjectStore/ObjectStoreStorage.php
mrow4a: additionally, file need to exists to be cached (I don't cache e.g. false)

DeepDiver1975 (changes requested):
I'm fine in merging this to master and get an understanding how this works.

We shall rebase and rerun the tests ....

mrow4a:
@DeepDiver1975 @PVince81 any decision on this?

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -54,6 +59,7 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 	private $objectPrefix = 'urn:oid:';
 
 	public function __construct($params) {
+		$this->objectStatCache = new ArrayCache();
```
butonic: this will eat memory when doing a lot of file activities, eg in a background job. use `CappedMemoryCache`

butonic (dismissed):
I like the conservative approach. It at least caches stuff for objectstorage.

DeepDiver1975 (approved):

mrow4a:
@DeepDiver1975 @butonic I will keep an eye on smashbox with objectostrage tests for this change.

PR #31941, opened 2018-06-28
Revert "[stable10] Cleanup the encryption keys after file deletion"

sharidas:
Reverts owncloud/core#31934

PR #31959, opened 2018-07-02
Revert "Revert "[stable10] Cleanup the encryption keys after file deletion""

sharidas:
Reverts owncloud/core#31941


PVince81 (approved):
:+1:

don't merge due to code freeze

PR #31960, opened 2018-07-02
Update PR template

PVince81:
Add more details about what labels to set.

Simplify the last section which had obsolete questions that are already enforced by automation

I wonder if we should get rid of the "3 - To release" and other labels.

Some of the checkboxes can likely be replaced by automation at some point.

I'm not sure if the "type of change" is really useful here as I seldom look at this section. We can use the labels instead "enhancement", "bug", "technical debt" ?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,11 +5,19 @@ This is the bug tracker for the Server component. Find other components at https
 
 For fixing potential security issues please see https://owncloud.org/security/
 
-To make it possible for us to get your change reviewed and merged please fill out below information carefully.
+To make it possible for us to get your change reviewed and merged please carefully fill out the requested information below.
 
-Please note that any kind of change first has to be submitted to the master branch which holds the next major version of ownCloud.
+Please note that any kind of change needs first be submitted to the master branch which holds the next major version of ownCloud.
 
 We will carefully discuss if your change can or has to be backported to stable branches.
+
+Please set the following labels:
+
+- Set label "3 - To review" for review or "2 - Development" in case the PR still has open tasks
+- Set label "backport-request" if backport is needed
+- Assignment: assign to self
+- Milestone: set the same as the ticket this PR fixes, or "development" by default
+- Reviewers: pick at least one
 -->
 
 ## Description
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,11 +5,19 @@ This is the bug tracker for the Server component. Find other components at https
...
+To make it possible for us to get your change reviewed and merged please carefully fill out the requested information below.
 
-Please note that any kind of change first has to be submitted to the master branch which holds the next major version of ownCloud.
+Please note that any kind of change needs first be submitted to the master branch which holds the next major version of ownCloud.
 
 We will carefully discuss if your change can or has to be backported to stable branches.
+
+Please set the following labels:
+
+- Set label "3 - To review" for review or "2 - Development" in case the PR still has open tasks
```
phil-davis: If the PR is coming from a community person who has to do a PR  from their own  fork, then they do not have access to set labels.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -36,14 +49,16 @@ We will carefully discuss if your change can or has to be backported to stable b
 - [ ] Bug fix (non-breaking change which fixes an issue)
 - [ ] New feature (non-breaking change which adds functionality)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Technical debt
```
phil-davis: Could also add "Tests" - I am often putting "Acceptance tests" here.
Also "Refactoring"
Maybe both of those can  come under the "Technical debt" heading anyway?

phil-davis (commented):
See 2 comments

mmattel:
Sidenote: could we have that new issue template not only in core but on many other repros.
That would harmonize the things and makes it much more clear.

phil-davis:
And have all the common labels - it is a bit of a scatter-gun at the moment, which repo has which labels available.

patrickjahns:
> Sidenote: could we have that new issue template not only in core but on many other repros.
That would harmonize the things and makes it much more clear.

Linked internal ticket for tracking

DeepDiver1975 (approved):

PVince81:
we might be able to remove the "type of changes" if we use labels instead, and later on changelog files

Issue #31814, opened 2018-06-18
"mountConfigLoaded" event triggered twice in the personal page


jvillafanez:
### Steps to reproduce
1. Modify the file "apps/files_external/js/settings.js" to include something like around line 1430
    ```
    $("#body-settings").on("mountConfigLoaded", function(ev){
      console.log('config loaded');
      console.log(ev);
    })
    ```
2. Make sure you have at least one mount point configured by the admin and another personal
3. Go to the personal settings page
4. Check the console logs.

### Expected behaviour
Only one event should be reported. The event should be triggered when both personal and admin mount points are loaded

### Actual behaviour
Two events are triggered

![screen shot 2018-06-18 at 09 31 27](https://user-images.githubusercontent.com/1477829/41523496-7a3d6c22-72da-11e8-9585-35cdc97a7298.png)

### Server configuration
**Operating system**: ubuntu 16.04

**Web server:** apache 2.4

**Database:** mysql

**PHP version:** 7.0

**ownCloud version:** master code (10.1 prealpha) at 9edd97a, very likely in 10.0.x too

**Updated from an older ownCloud or fresh install:** fresh install

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/19711 (scanFile event gets triggered on directories), https://github.com/owncloud/core/issues/6398 (Provide Admin more control over personal page), https://github.com/owncloud/core/issues/20455 (Full name is not shown just after modifying it in the Personal Page), https://github.com/owncloud/core/pull/31033 (Remove "Unlimited" from quota report in personal page), and https://github.com/owncloud/core/issues/17528 (Prevent closing upgrade page while upgrade in progress).

davitol:
@pmaier1 JFYI. We should decide if we must fix this stuff for next oC 10.0.10 version. After talking with @jvillafanez his opinion is that it will be worthy to dedicate time for this fix for the next release.

PVince81:
@jvillafanez how much time needed for this issue + unit test ?

PVince81:
this could potentially have an impact on external storages that use OAuth and rely on this event, so might need to retest GDrive and Dropbox at least

jvillafanez:
I'd say around 3md for both dev and unittest, QA not included. Between 3-4md including some basic test for GDrive and Dropbox.

PVince81:
[...]

PR #31962, opened 2018-07-02
Trigger mountConfigLoaded event properly, and initialize events after

jvillafanez:
## Description
Trigger the mountConfigLoaded event only once in the personal page, and initialize the mountConfig events after the storage information is loaded in the page.

## Related Issue
https://github.com/owncloud/core/issues/31814

## How Has This Been Tested?
Manually tested

## Types of changes

## Checklist:

### DIFF
--- a/apps/files_external/js/settings.js
+++ b/apps/files_external/js/settings.js
@@ -673,8 +673,6 @@ MountConfigListView.prototype = _.extend({
 		// into the data-configurations attribute of the select
 		this._allBackends = this.$el.find('.selectBackend').data('configurations');
 		this._allAuthMechanisms = this.$el.find('#addMountPoint .authentication').data('mechanisms');
-
-		this._initEvents();
 	},
 
 	/**
@@ -942,10 +940,11 @@ MountConfigListView.prototype = _.extend({
 	 */
 	loadStorages: function() {
 		var self = this;
+		var ajaxRequests = [];
 
 		if (this._isPersonal) {
 			// load userglobal storages
-			$.ajax({
+			var personalRequest = $.ajax({
 				type: 'GET',
 				url: OC.generateUrl('apps/files_external/userglobalstorages'),
 				data: {'testOnly' : true},
@@ -983,14 +982,14 @@ MountConfigListView.prototype = _.extend({
 						}
 					});
 					onCompletion.resolve();
-					$("#body-settings").trigger("mountConfigLoaded");
 				}
 			});
+			ajaxRequests.push(personalRequest);
 		}
 
 		var url = this._storageConfigClass.prototype._url;
 
-		$.ajax({
+		var globalRequest = $.ajax({
 			type: 'GET',
 			url: OC.generateUrl(url),
 			contentType: 'application/json',
@@ -1004,8 +1003,14 @@ MountConfigListView.prototype = _.extend({
 					self.recheckStorageConfig($tr);
 				});
 				onCompletion.resolve();
-				$("#body-settings").trigger("mountConfigLoaded");
 			}
+		});
+		ajaxRequests.push(globalRequest);
+
+		$.when.apply($, ajaxRequests).always(function() {
+			// trigger the mountConfigLoaded event when both requests (or one if
+			// it's the only one) finish regardless of success
+			$('#body-settings').trigger('mountConfigLoaded');
 		});
 	},
 
@@ -1347,6 +1352,11 @@ $(document).ready(function() {
 	var mountConfigListView = new MountConfigListView($('#externalStorage'), {
 		encryptionEnabled: encryptionEnabled,
 		allowUserMountSharing: (parseInt($('#allowUserMountSharing').val(), 10) === 1)
+	});
+
+	// init the mountConfigListView events after the storages are loaded
+	$('#body-settings').on('mountConfigLoaded', function() {
+		mountConfigListView._initEvents();
 	});
 	mountConfigListView.loadStorages();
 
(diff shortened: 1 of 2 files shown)


apps/files_external/js/settings.js
```diff
@@ -983,14 +982,14 @@ MountConfigListView.prototype = _.extend({
...
-					$("#body-settings").trigger("mountConfigLoaded");
 				}
 			});
+			ajaxRequests.push(personalRequest);
 		}
 
 		var url = this._storageConfigClass.prototype._url;
 
-		$.ajax({
+		var globalRequest = $.ajax({
```
jvillafanez: Need to change the variable name. For the personal storages, the target url is "/apps/files_external/userstorages" which doesn't reflect the request.

PVince81 (commented):
Code makes sense.

Can you add JS unit tests ? You might want to use `fakeServer.requests` to simulate success or failure.

jvillafanez:
I've verified manually the behaviour and I haven't seen any problem.

PVince81:
did you test with Dropbox and GDrive oauth handshake ?

jvillafanez:
Working with dropbox, but having problems with gdrive. It happens the same as https://github.com/owncloud/files_external_gdrive/issues/5 . I'm not sure if it's related though.

apps/files_external/js/settings.js
```diff
@@ -778,6 +776,42 @@ MountConfigListView.prototype = _.extend({
...
+	 * this will usually be a list of ajax requests, but other deferred
+	 * objects can be included
+	 * @param {callback} callback the callback to be executed after all the
+	 * deferred object have finished
+	 */
+	_executeCallbackWhenFinished: function(deferreds, callback) {
+		var self = this;
+		var pendingDeferreds = [];
+
+		$.when.apply($, deferreds).always(function() {
```
PVince81: how about just returning a promise and resolve it instead of calling a callback ?

at least this would bring some consistency instead of mixing promises and callbacks

apps/files_external/js/settings.js
```diff
@@ -778,6 +776,42 @@ MountConfigListView.prototype = _.extend({
...
+			var pendingDeferreds = [];
+
+			// some deferred objects can still be pending to be resolved
+			// or rejected. Get all of those which are still pending so we can
+			// make another call
+			if (deferreds.length > 0) {
```
PVince81: I'm surprised you can still having pending ones. I thought there's a way for `$.when` to detect the moment where all were resolved or failed, maybe `.then()` ?

apps/files_external/tests/js/settingsSpec.js
```diff
@@ -418,4 +419,149 @@ describe('OCA.External.Settings tests', function() {
...
 	});
+
+	describe('mountConfigLoaded event is triggered', function() {
+		var view;
+		var $bodySettings;
+		var fakeServer;
+
+		beforeEach(function() {
+			// add the body-settings for the mountConfigLoaded event
+			$('#testArea').append('<div id="body-settings"></div>');
```
PVince81: maybe move this to the main `beforeEach`
please note that Jasmine will run all `beforeEach` from top to the nested one in order

apps/files_external/tests/js/settingsSpec.js
```diff
@@ -418,4 +419,149 @@ describe('OCA.External.Settings tests', function() {
...
+
+			var $el = $('#externalStorage');
+			view = new OCA.External.Settings.MountConfigListView($el, {encryptionEnabled: false});
+			$bodySettings = $('#body-settings');
+			fakeServer = sinon.fakeServer.create();
```
PVince81: not needed, we have it globally created in the file specHelper.js

PVince81 (commented):
Thanks for adding tests. I have a few comments for stuff to tweak.

apps/files_external/js/settings.js
jvillafanez: Quoting official jquery docs:
> In the case where multiple Deferred objects are passed to jQuery.when(), the method returns the Promise from a new "master" Deferred object that tracks the aggregate state of all the Deferreds it has been passed. The method will resolve its master Deferred as soon as all the Deferreds resolve, or reject the master Deferred as soon as one of the Deferreds is rejected.

apps/files_external/js/settings.js
jvillafanez: I have to check how we can do it properly, but it makes sense. 👍

apps/files_external/tests/js/settingsSpec.js
jvillafanez: 👍

apps/files_external/tests/js/settingsSpec.js
jvillafanez: I thought I had trouble if the fakeServer isn't set up there. I have to recheck what happens without it.

apps/files_external/js/settings.js
jvillafanez: I don't think we can use promises here.
The deferred object needs to be created inside the function and it needs to return the promise associated to that deferred object, otherwise it makes no sense.
The first problem is the recursive call we're doing, that prevents us to create the deferred object inside the function call.
The second problem is that we can't undo the recusivity due to the asynchronous call

PVince81 (approved):
:+1:

PVince81:
@jvillafanez please backport

jvillafanez:
Backport in https://github.com/owncloud/core/pull/32095

PR #31969, opened 2018-07-03
Disable jQuery's globalEval function

settermjd:
## Description

Begin migrating from 2.1.4 to the 3.x series.

## Related Issue

https://github.com/owncloud/enterprise/issues/2701

## Motivation and Context

This is intended to start the migration from jQuery 2.1.4 to the 3.x branch. There's a promise to maintain backward compatibility to the 1.x jQuery API, so this method needs to be retained, yet deprecated.

## How Has This Been Tested?

It hasn't been, yet.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #31971, opened 2018-07-03
Disable jquery globalEval

PVince81:
Alternative approach of https://github.com/owncloud/core/pull/31969

The file is loaded on every page, even login.


PVince81:
stable10: https://github.com/owncloud/core/pull/31972

DeepDiver1975 (approved):

PR #31984, opened 2018-07-04
Remove sensitive shared_secret data from occ config:list output

settermjd:
## Description

This PR applies a patch from @phil-davis, that ensures that `log.condition` `shared_secret` sensitive data is removed from calls to `occ config:list` output.

## Motivation and Context

```php
'log.condition' => [
        [
            'shared_secret' => '57b58edb6637fe3059b3595cf9c41b9',
            'users' => ['user1'],
            'apps' => ['files_texteditor'],
            'logfile' => '/tmp/test.log'
        ],
    ],
```

If you had the above configuration in `config/config.php`, then the `shared_secret` information would not be stripped, from the output of a call to `occ config:list`. However, with the PR applied, the output would appear as follows:

```php
"log.condition": [
      {
          "shared_secret": "***REMOVED SENSITIVE VALUE***",
           "users": [
               "user1"
           ],
           "apps": [
               "files_texteditor"
           ],
           "logfile": "\/tmp\/test.log"
      },
]
```

## How Has This Been Tested?

Currently, it's only been manually tested.

## Types of changes

- [X] Technical debt
- [ ] Tests

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised:

## Open tasks:

- [X] Backport (if applicable set "backport-request" label and remove when the backport was done)

### DIFF
--- a/lib/private/SystemConfig.php
+++ b/lib/private/SystemConfig.php
@@ -166,7 +166,7 @@ class SystemConfig {
 					 * Remove any sensitive values from all actual entries.
 					 */
 					foreach ($value as $valueKey => $valueData) {
-						if (is_int($valueKey) && ($valueKey >= 0)) {
+						if (\is_int($valueKey) && ($valueKey >= 0)) {
 							$value[$valueKey] = $this->removeSensitiveValue($valueToRemove, $value[$valueKey]);
 						}
 					}
--- a/lib/private/SystemConfig.php
+++ b/lib/private/SystemConfig.php
@@ -50,7 +50,14 @@ class SystemConfig {
 		'proxyuserpwd' => true,
 		'marketplace.key' => true,
 		'log.condition' => [
+			[
 			'shared_secret' => true,
+			]
+		],
+		'log.conditions' => [
+			[
+			'shared_secret' => true,
+			]
 		],
 		'license-key' => true,
 		'redis' => [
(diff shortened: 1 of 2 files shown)


phil-davis:
My patch had a code-style fail. I pushed a 2nd commit that has the little code-style fix and unit tests.

phil-davis:
It would be nice to have acceptance tests for ``occ`` commands like this. @patrickjahns raised a general issue about that: https://github.com/owncloud/QA/issues/570
It won't happen overnight, but it will happen. (Separate to this PR)

settermjd:
@phil-davis, in this case, would unit tests not be more suitable?

phil-davis:
Yes, unit tests are good - they check that the underlying classes/methods do the right thing.

Acceptance tests check that the actual ``occ`` commands have actually been hooked up through the code to the underlying classes/methods that they should be. e.g. in this case if the ``occ config:list`` command is not actually written to end up making use of the class with ``getFilteredValue()`` then it won't do the job - no matter how many unit tests of ``getFilteredValue()`` are passing.

phil-davis:
@settermjd this PR has been started from a very old ``master`` - I will rebase it. drone CI is complaining about ``apiSharing`` suite not found. That was renamed a long time ago.

You should pull ``master`` branch to your local copy of ``core`` repo to get it up-to-date.

lib/private/SystemConfig.php
```diff
@@ -50,7 +50,14 @@ class SystemConfig {
 		'proxyuserpwd' => true,
 		'marketplace.key' => true,
 		'log.condition' => [
+			[
 			'shared_secret' => true,
+			]
+		],
+		'log.conditions' => [
```
PVince81: I have the feeling that this was already in place but was missing the "s", see line 52.

AFAIK there is no setting called singular "log.condition", see config.sample.php

PVince81 (changes requested):
see comment

lib/private/SystemConfig.php
phil-davis: ``log.condition`` is still backward-compatible supported. See ``lib/private/Log.php`` function ``log`` - the first few lines check ``log.conditions`` andd then fall back to ``log.condition``

PVince81 (approved):

PVince81:
Please backport

phil-davis:
Backport ``stable10`` #31997

PR #31988, opened 2018-07-04
empty path has no parents

butonic:
what can I say ... :disappointed: 

Without this a public share, mounted as a federated share in the root of a users home will hav its etag overwritten with a uniqeid on every propfind, causing endless downloads.

How to reproduce:

1. as user admin create a 1 byte txt file.
2. share it as public link
3. log out
4. open the link 
5. add to the same oc instance
6. log in as user test
7. accept incoming share
8. do a propfind against user test like
```
curl 'http://core/remote.php/dav/files/test/' -X PROPFIND --data-binary $'<?xml version="1.0"?>\n<d:propfind xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">\n <d:prop>\n <d:getetag />\n <oc:fileid />\n <oc:permissions />\n <oc:size />\n </d:prop>\n</d:propfind>' --compressed -k -u test:123 -H 'Cookie: XDEBUG_SESSION=PHPSTORM' | xml_pp
```

Expected behavior:
multiple PROPFINDS give the same md5 etag

Actual behavior:
every propfind returns a new uniqueid

### DIFF
--- a/lib/private/Files/Cache/Propagator.php
+++ b/lib/private/Files/Cache/Propagator.php
@@ -62,6 +62,10 @@ class Propagator implements IPropagator {
 
 		$parents = $this->getParents($internalPath);
 
+		if (\count($parents) === 0) {
+			return;
+		}
+
 		if ($this->inBatch) {
 			foreach ($parents as $parent) {
 				$this->addToBatch($parent, $time, $sizeDifference);
@@ -99,6 +103,9 @@ class Propagator implements IPropagator {
 	}
 
 	protected function getParents($path) {
+		if ($path === '') {
+			return [];
+		}
 		$parts = \explode('/', $path);
 		$parent = '';
 		$parents = [];
--- a/tests/lib/Files/Cache/PropagatorTest.php
+++ b/tests/lib/Files/Cache/PropagatorTest.php
@@ -81,6 +81,25 @@ class PropagatorTest extends TestCase {
 		}
 	}
 
+	public function getParentsProvider() {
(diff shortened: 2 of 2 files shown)


PVince81 (approved):
Code makes sense, strange that this wasn't discovered before.

I'm also surprised that the logic in getParents() didn't result in an empty array.

Anyway, looks good and safe :+1:

PVince81:
PHP CS fixer unhappy and some tests fail

butonic:
@PVince81 `explode('/','')` will return `['']` not `[]` see http://php.net/manual/de/function.explode.php#99830 ... and we need `''` as an element, so we cannot use array_filter ... well ... actually...

PVince81:
damn... I see.

I'd like to find out why this wasn't discovered earlier. Maybe in most envs and scenarios the etag change somehow doesn't always affect syncing.

PVince81:
@butonic please backport, for 10.0.10

butonic:
@PVince81 AFAICT only mounted public links to a file are represented with `''` as the internal path. So only those are affected. And only since the zsync stuff has been merged, because it will do an is_file check on the file, causing the etag to be overwritten. So ... actuall seems to be a regression introduced with the zsync stuff.

PVince81:
hmmm, but the zsync stuff is only on master ?

are you able to reproduce this on stable10 ?

phil-davis:
Backport ``stable10`` #31992

PR #31989, opened 2018-07-04
Fix Cache::update() to reset column on oracle 

### LINKED ISSUE #31692: File cache update() cannot set empty string on Oracle
### Steps

1. Set a checksum (maybe upload a file)
1. Reset the checksum to empty string using the Node API: `$file1->getStorage()->getCache()->update($file1->getId(), ['checksum' => '']);`
1. Check database

### Expected result
Checksum cleared to empty string on all DBs.

### Actual result
Checksum not cleared on Oracle, other DBs are fine.

Even more worrying is that if you set it to "null" in the update() function it doesn't work for **all DBs**.

### Versions
OC 10.0.8

Discovered on https://github.com/owncloud/core/pull/31430

### Tech concept for the fix

- [x] add unit test for `OC\File\Cache::update()` to reproduce the issue and prove it's broken on Oracle
- [x] Modify `OC\File\Cache` to detect Oracle. If Oracle and the parameter to update is an empty string, convert it to NULL and adjust the query if necessary.
- [ ] check if other parts of file cache might be affected
- [ ] raise separate ticket for NULL issue

Please note that the update query contains a `WHERE` statement to avoid updating values that are already set to the same value in the DB. This needs special handling for nulls, etc.

At this point to avoid potential regressions for other DBs, I suggest ignoring the "setting NULL" issue and focus on Oracle only.

### Estimate
- Dev time including unit tests: 1-2md

@butonic @DeepDiver1975 @tomneedham

IljaN:
## Description
FileCache columns were not reset on oracle when an empty value was passed instead of null.

## Related Issue

- Fixes #31692

## How Has This Been Tested?

Unit-tests 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -35,6 +35,7 @@
 
 namespace OC\Files\Cache;
 
+use Doctrine\DBAL\Platforms\OraclePlatform;
 use OCP\Files\Cache\ICache;
 use OCP\Files\Cache\ICacheEntry;
 use \OCP\Files\IMimeTypeLoader;
@@ -304,6 +305,16 @@ class Cache implements ICache {
 		// once for the SET part, once in the WHERE clause
 		$params = \array_merge($params, $params);
 		$params[] = $id;
+
+		// Oracle does not support empty string values so we convert them to nulls
+		// https://github.com/owncloud/core/issues/31692
+		if ($this->connection->getDatabasePlatform() instanceof OraclePlatform) {
+			foreach ($data as $param => $value) {
+				if ($value === '') {
+					$data[$param] = null;
+				}
+			}
+		}
 
 		// don't update if the data we try to set is the same as the one in the record
 		// some databases (Postgres) don't like superfluous updates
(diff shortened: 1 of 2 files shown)


lib/private/Files/Cache/Cache.php
```diff
@@ -305,6 +305,16 @@ public function update($id, array $data) {
 		$params = \array_merge($params, $params);
 		$params[] = $id;
 
+		// Oracle does not support empty string values so we convert them to nulls
+		// https://github.com/owncloud/core/issues/31692
+		if ($this->connection->getDatabasePlatform()->getName() === 'oracle') {
```
DeepDiver1975: usually we go for
```php
if ($this->connection->getDatabasePlatform() instanceof OraclePlatform) {

DeepDiver1975 (changes requested):

lib/private/Files/Cache/Cache.php
IljaN: Will push changes after/if more change requests arrive to conserve ci-ressources.

PVince81 (approved):
:+1:

DeepDiver1975 (approved):

IljaN:
Backport here: https://github.com/owncloud/core/pull/31996

PR #31992, opened 2018-07-05
[stable10] empty path has no parents

butonic:
backport of https://github.com/owncloud/core/pull/31988 to stable10


PVince81 (approved):
:+1:

PVince81:
would be good to smash this before merging. I don't expect any regressions but just in case

PVince81:
assigning self to smash later

PR #31803, opened 2018-06-15
[stable10] Skip filecache repair step for version greater than 10.0.4

sharidas:
Skip filecache repair step for version greater than 10.0.4

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

The file cache repair step can be omitted if the current oC version is 10.0.4 or greater than 10.0.4.

## Related Issue

https://github.com/owncloud/core/issues/31464

## Motivation and Context

The file cache repair step can be omitted if the current oC version is 10.0.4 or greater than 10.0.4.

## How Has This Been Tested?

Same as mentioned in https://github.com/owncloud/core/pull/31795

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #30140, opened 2018-01-15
Store installed at version

tomneedham:
## Description
Stores a key in the app config with the originating version of owncloud so we know the upgrade history (to some extent)

## Motivation and Context
Useful for bug tracking at least

## How Has This Been Tested?
manual reinstalls. and testing without this key being present

## Screenshots (if appropriate):
screen shot 2018-01-15 at 12 46 28
screen shot 2018-01-15 at 12 46 06

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #29868, opened 2017-12-15
Avatar migration unnecessarily loads FS of every seen user


tomneedham:
https://github.com/owncloud/core/blob/master/lib/private/Repair/MoveAvatarOutsideHome.php#L107

Bad because maybe not all/any users have avatars - and setupFS is slow. We should search the filecache for avatars, resolve to storages, resolve to distinct user_id and then only run the migration on these ids. Will reduce downtime for instances with large numbers.

Some instances setupfs can take up to 1s/user ... do the math.

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

PVince81:
@tomneedham  didn't we solve this differently ?

tomneedham:
No - we still need to fix this. I will most likely make a PR to adjust the migration to make it more efficient - and cleanup cases where the avartar migration failed to do anything

tomneedham:
(blocked because it only runs if it is executed during an upgrade)

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

PVince81:
moved to planned

PVince81:
p3 and planned

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

PR #31994, opened 2018-07-05
Always try avatar migration. Stored when done. Skip when done already

tomneedham:
## Description
For some upgrades, this migration took too long and was not completed. I removed the version check so that it can be attempted again using `occ maintenance:repair`. It will store when it is complete, and only if owncloud was installed before 10.0.9
## Related Issue

- Fixes <issue_link>

## Motivation and Context
Failed avatar migrations, and avatars were disabled. This allows another attempt at the migration, without downtime, even after upgrade to ownCloud 10.

## How Has This Been Tested?
 - Manually running the occ maintenance:repair command

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Repair.php
+++ b/lib/private/Repair.php
@@ -129,11 +129,6 @@ class Repair implements IOutput {
 	public static function getRepairSteps() {
 		return [
 			new RepairMimeTypes(\OC::$server->getConfig()),
-			new RepairMismatchFileCachePath(
-				\OC::$server->getDatabaseConnection(),
-				\OC::$server->getMimeTypeLoader(),
-				\OC::$server->getLogger(),
-				\OC::$server->getConfig()),
 			new FillETags(\OC::$server->getDatabaseConnection()),
 			new CleanTags(\OC::$server->getDatabaseConnection(), \OC::$server->getUserManager()),
 			new DropOldTables(\OC::$server->getDatabaseConnection()),
@@ -178,6 +173,12 @@ class Repair implements IOutput {
 	public static function getExpensiveRepairSteps() {
 		return [
 			new OldGroupMembershipShares(\OC::$server->getDatabaseConnection(), \OC::$server->getGroupManager()),
+			new RepairMismatchFileCachePath(
+				\OC::$server->getDatabaseConnection(),
+				\OC::$server->getMimeTypeLoader(),
+				\OC::$server->getLogger(),
+				\OC::$server->getConfig()
+			)
 		];
 	}
 
--- a/lib/private/Repair/MoveAvatarOutsideHome.php
+++ b/lib/private/Repair/MoveAvatarOutsideHome.php
@@ -129,18 +129,28 @@ class MoveAvatarOutsideHome implements IRepairStep {
 	 * @param IOutput $output
 	 */
 	public function run(IOutput $output) {
-		$ocVersionFromBeforeUpdate = $this->config->getSystemValue('version', '0.0.0');
-		if (\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
-			$function = function (IUser $user) use ($output) {
-				$this->moveAvatars($output, $user);
-				$output->advance();
-			};
-
-			$output->startProgress($this->userManager->countSeenUsers());
-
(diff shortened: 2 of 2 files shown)


tomneedham:
Related: https://github.com/owncloud/core/issues/29868

lib/private/Repair.php
```diff
@@ -178,6 +173,12 @@ public static function getRepairSteps() {
 	public static function getExpensiveRepairSteps() {
 		return [
 			new OldGroupMembershipShares(\OC::$server->getDatabaseConnection(), \OC::$server->getGroupManager()),
+			new RepairMismatchFileCachePath(
```
PVince81: starting with 10.0.10 we'l anyway only run this one once: https://github.com/owncloud/core/pull/31803

now wondering whether we should revert that in case someone wants to rerun the check ? or add the check as part of occ files:scan --repair-check ?

lib/private/Repair/MoveAvatarOutsideHome.php
```diff
@@ -129,18 +129,24 @@ private function moveAvatars(IOutput $out, IUser $user) {
 	 * @param IOutput $output
 	 */
 	public function run(IOutput $output) {
-		$ocVersionFromBeforeUpdate = $this->config->getSystemValue('version', '0.0.0');
-		if (\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
-			$function = function (IUser $user) use ($output) {
-				$this->moveAvatars($output, $user);
-				$output->advance();
-			};
+		// TODO when installed_at is finally stored, dont run if installed_at >= 10.0.0
```
PVince81: the PR https://github.com/owncloud/core/pull/30140 was merged, please adjust.

is there no other approach like what versions we're upgrading from ? 9.2 is actually supposed to be 10.0

PVince81 (changes requested):
see comments

PVince81:
please run `make test-php-style` locally

lib/private/Repair.php
tomneedham: I don't like the version checks because they stop you from using the code again without code hacks. I think this solution is better.

tomneedham:
> please run make test-php-style locally

Can I tell this to look only at core? Otherwise output is massive with it considering every app I have cloned locally - not really usable

phil-davis:
``test-php-style`` ignores any "bonus" apps in the ``apps``  folder.

tomneedham:
> test-php-style ignores any "bonus" apps in the apps folder.

Not custom app dirs it seems :/

phil-davis:
> Not custom app dirs it seems
@PVince81 had this problem, and I guess has added some ``excludes`` for his local repo - you might ask him what he did.

tomneedham:
Updated to not run if installed since owncloud 10 and updated code style.

PVince81:
@tomneedham can you adjust the tests of the move avatar repair steps to match the new checks ?

tomneedham:
> @tomneedham can you adjust the tests of the move avatar repair steps to match the new checks ?

not sure I fully understand what you mean @PVince81 ?

PVince81:
@tomneedham does MoveAvatarOutsideHomeTest contain any tests for the version checks ?

PVince81:
@tomneedham instructions for take over ?

tomneedham:
Avatars were migrated from OC 9 -> 10. This is done with a repair step. This is quite expensive and some installations skipped it. 

This PR makes this step only run once - and stores when complete. This also allows us to reset that flag and try again should we need it.

tomneedham:
> @tomneedham does MoveAvatarOutsideHomeTest contain any tests for the version checks ?

It does - but it uses `first_installed_version` which was only added recently. So we still do a check to see if it has executed before, and store a flag on complete.

tomneedham:
Closing this actually. I think we can come up with a better strategy to fix instances where we need to re-run the avatar migration.

phil-davis:
Removing backport label since this has been closed.

PR #31976, opened 2018-07-04
notify on build failures when pushed/tagged on core

patrickjahns:
## Description

Send slack notifications, when a push/tag to `master` branch leads to a build failure

## Motivation and Context
We want to know instantly when our master branch is in 🔴 

## How Has This Been Tested?
🤖 
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] CI improvement

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #32019, opened 2018-07-09
map the webhook to the value which slack plugin expects

patrickjahns:
## Description
With #31976 notifications have been added - however the secret was not mapped correctly. This fixes it

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -275,7 +275,9 @@ pipeline:
   notify:
     image: plugins/slack:1
     pull: true
-    secrets: [ slack_webhook_public ]
+    secrets:
+      - source: slack_webhook_public
+        target: slack_webhook
     channel: server
     when:
       status: [ failure, changed ]


PVince81 (approved):
:+1:

.drone.yml
```diff
@@ -275,7 +275,9 @@ pipeline:
   notify:
     image: plugins/slack:1
     pull: true
-    secrets: [ slack_webhook_public ]
+    secrets:
+     - source: slack_webhook_public
```
PVince81: is the "-" here wanted ?

tboerger: It's wanted, but wrong indentation.

tboerger (commented):
I would prefer a simple naming like `slack_webhook` which would make the mapping much simpler/better readable. no need to differ between public or private chat since the notifications so far are anyway only sent to public chat.

.drone.yml
patrickjahns: fixed - somehow phpstorm shoed me the indentation line 🤷‍♂️

patrickjahns:
> I would prefer a simple naming like slack_webhook which would make the mapping much simpler/better readable. no need to differ between public or private chat since the notifications so far are anyway only sent to public chat.

ACK - though I differentiate in opinion here - as I have no way of checking the value of the secret so I don't know straight away wether this goes to internal/external chat

tboerger:
We don't use this style of secret naming in any of our repos.

PR #32030, opened 2018-07-11
add log rotate script

voroyam:
added log rotate script options.

Don't know how to format this to add comments or to clarify commands. Please help :)

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -702,7 +702,7 @@ $CONFIG = array(
  */
 'log_rotate_size' => false,
 
-
+/**
 Configure log rotation for the owncloud.log
 
 
@@ -732,7 +732,7 @@ compress
 compresscmd /bin/gzip
 # use this compression command
 }
-
+*/
 
 /**
  * Alternate Code Locations
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -703,6 +703,37 @@ $CONFIG = array(
 'log_rotate_size' => false,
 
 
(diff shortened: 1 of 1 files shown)


patrickjahns:
Please use proper phpdoc syntax 

Your're changes would result otherwise in this: https://github.com/owncloud/core/blob/fc81a05ffe3cbf1e06171e325514b72a7d6126bc/config/config.sample.php#L706-L734

voroyam:
> phpdoc syntax

what is the proper phpdoc syntax?

mmattel:
@voroyam 
some comments:

1) In case you enable owncloud´s native log rotation by using a value other than false or 0 for ``log_rotate_size``, the code creates a new file when exceeding the treshhold with following scheme:
https://github.com/owncloud/core/blob/f2543f6409b80162471679832814b314e7749b36/lib/private/Log/Rotate.php#L45
So only owncloud.log and owncloud.1 is created/present. There is no big rotation mechanism like you have when using Linux log rotation.

2) For extending the rotation to a Linux style, have ``log_rotate_size`` set to ``false`` (or zero) and enable the Linux method to enable rotation. This is the content you have added.

In general, you should highlight in a small description the situation I have documented above. You can add that into the existing block of /** text */
The content itself needs some tuning, but add it into the docblock too.
I can maybe help later on but I am curretly very short on time resources

mmattel:
Closing in favour of https://github.com/owncloud/core/pull/32035

Issue #32040, opened 2018-07-13
fix API acceptance tests


individual-it:
- [x] we have an apiSharing-v1 folder - is there an apiSharing-v2 ??
- [x]  test sharing on ocs/v2 #32074
- [x]  refactor tests/acceptance/features/bootstrap/AppManagementContext.php so it can be used on a remote server #33095
- [x] make tests/acceptance/features/apiShareOperations/getWebDAVSharePermissions.feature also work with new DAV path  #33349
- [x] join tests/acceptance/features/apiSharees/sharees_provisioningapiv2.feature and tests/acceptance/features/apiSharees/sharees.feature issue #33350 and PR #33351 
- [x] in https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdav/webdav-related-old-endpoint.feature not every scenario has an webdav version defined #31799
- [x]  this scenario is in the old webdav feature file but uses the new webdav https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdav/webdav-related-old-endpoint.feature#L114 is ist for old, new or both versions? test is included now in `apiWebdavProperties/createFolder.feature`
- [x]  in https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdav/webdav-related-old-endpoint.feature put more `Given` in the `Background` section #31799
- [x]  split https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiWebdav/webdav-related.feature into multiple features
- [x] in tests/acceptance/features/apiSharing/updateShare.feature not only check the return codes but also if something useful happened (e.g. share is accessible) ##33329
- [x]  in tests/acceptance/features/apiSharing/updateShare.feature `Increasing permissions is allowed for owner` why making that user a sub-admin? #33333
- [x]  merge both features files in https://github.com/owncloud/core/tree/master/tests/acceptance/features/apiTrashbin into one file with `Examples` tables - issue #33354 
- [x] if needed split trashbin tests into multiple features - issue #33363 PR #33365 
- [x] fix api trashbin feature files, so that Background steps are used to create users - issue #33363 PR #33365 
- [x] look at https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiMain/dav-versions.feature and find out what it is about (duplications of API-Webdav suite?, scenarios that need to be moved? ....?) - issue #33379 
[...]

PR #32072, opened 2018-07-17
make sharing tests work correctly with ocs api v2

individual-it:
## Description
this allows the sharing tests to work correctly with `v2.php`

## Related Issue
#32040

## Motivation and Context
`v2.php` does map the OCS status codes to the corresponding HTTP codes, so some expectation in the code does not work for `v2.php`

## How Has This Been Tested?
run some of the corresponding tests locally, :robot: will do the rest

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -551,30 +551,44 @@ trait BasicStructure {
 	/**
 	 * @Then /^the OCS status code should be "([^"]*)"$/
 	 *
-	 * @param int $statusCode
+	 * @param int|int[] $statusCode
 	 * @param string $message
 	 *
 	 * @return void
 	 */
 	public function theOCSStatusCodeShouldBe($statusCode, $message = "") {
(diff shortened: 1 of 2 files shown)


phil-davis (approved):
Note to self: I first thought that expanding these "status code should be" steps to take an array of codes was not a good idea. But for cases where we just want to check if some action happened OK regardless of the OCS etc version in use at the time, then it is a good thing.

In specific Gherkin step code, we need to still always check for just the single expected status code value for the particular OCS, sharing, etc version in use.

phil-davis:
Something happened in drone. @individual-it you can look at that!

individual-it:
I think drone was maintenanced yesterday

Am 17. Juli 2018 20:56:02 GMT+05:45 schrieb Phil Davis <notifications@github.com>:

individual-it:
backport in #32079

PR #32052, opened 2018-07-16
moves duplicated steps to background

individual-it:
## Description
moves steps into the background section

## Related Issue

#32040

## Motivation and Context
make the tests files shorter

## How Has This Been Tested?
:robot: 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32074, opened 2018-07-17
Test sharing with ocs api v2

individual-it:
## Description
sharing can be used via `v1.php` and via `v2.php`, we never tested it with OCS API v2, so this adds some tests and some more checks where applicable

~depends on #32052 & #32072~

## Related Issue
#32040

## Motivation and Context
having better test coverage

## How Has This Been Tested?
run tests locally, drone will do it again :robot: 

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:


tests/acceptance/features/apiSharing-v1/gettingShares.feature
```diff
@@ -91,19 +106,25 @@ Feature: sharing
...
+		Examples:
+			|ocs_api_version|http_status_code|
+			|1              |200             |
+			|2              |404             |
 
-Scenario: Share of folder to a group, remove user from that group
-		Given user "user0" has been created
-		And user "user1" has been created
+	Scenario: Share of folder to a group, remove user from that group
+		Given using API version "<ocs_api_version>"
```
phil-davis: Looks like this should be a scenario outline?

tests/acceptance/features/apiSharing-v1/deleteShare.feature
```diff
@@ -115,12 +127,11 @@ Feature: sharing
 			| shareType   | 0      |
 			| permissions | 1      |
 		When user "user1" deletes file "/shared/shared_file.txt" using the API
-		Then the HTTP status code should be "403"
+		And the HTTP status code should be "403"
 		And as "user1" the file "/shared/shared_file.txt" should exist
```
phil-davis: This change not needed

tests/acceptance/features/apiSharing-v1/uploadToShare.feature
```diff
@@ -116,7 +107,6 @@ Feature: sharing
 		Then the content of file "/FOLDER/test.txt" for user "user0" should be "test"
 
 	Scenario: Uploading file to a user read/write share folder works
-		Given user "user0" has been created
```
phil-davis: Needs a "given" belowqq

phil-davis (approved):
Just a few little things you can fix.

phil-davis:
In some scenarios you still only doOCS API version 1.
Can you make some comment about why those scenarios are not a scenario outline that also has version 2.

tests/acceptance/features/apiSharing-v1/gettingShares.feature
individual-it: no that has actually only to test API v1

individual-it:
- Wherever the return codes (OCS or HTTP) of the sharing API are tested, both versions are checked 
- Whenever a step calls the sharing API directly (e.g. ` 		When user "user0" sends HTTP method "GET" to API endpoint "/apps/files_sharing/api/v1/shares?path=PARENT&subfiles=true"`, both versions are checked
- When steps do not specify how exactly the share was creates only one version is checked, because the step says "using the API" and should not care about which API is used
- Whenever the return codes of some webdav action are checked also only one version of the ocs api is checked assuming it does not matter how the sharing is done

PR #32075, opened 2018-07-17
[WIP] Send email notification to user/group of users as background job

sharidas:
The action to send emails to user or group of users should be
done using background jobs.

## Description

The action of sending emails to user or a group of users should be done via background jobs.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

The action of sending emails to user or a group of users should be done via background jobs.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/MailDispatcher.php
+++ b/apps/files_sharing/lib/MailDispatcher.php
@@ -1,0 +1,298 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Sharing;
+
+use OC\BackgroundJob\QueuedJob;
+use OC\Share\Filters\MailNotificationFilter;
+use OC\Share\MailNotifications;
+use OC\User\User;
+use OCP\BackgroundJob\IJobList;
+use OCP\IConfig;
+use OCP\IGroupManager;
+use OCP\IL10N;
+use OCP\ILogger;
+use OCP\IURLGenerator;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\Mail\IMailer;
+use DateTime;
+use OCP\Util;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+/**
+ * Class MailDispatcher
+ *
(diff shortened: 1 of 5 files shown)


sharidas:
This change was tested by adding above this line https://github.com/owncloud/core/blob/master/lib/private/Session/Memory.php#L119
```
$this->sessionClosed = false;
```
Primarily this change is tested with cron.php, that is the execute method of the MailDispatcher is tested.

apps/files_sharing/lib/MailDispatcher.php
```diff
@@ -0,0 +1,298 @@
...
+				'link' => $link,
+				'file' => $filename,
+			]);
+
+			$filename = $filter->getFile();
+			$link = $filter->getLink();
+
+			$subject = (string) $this->l10n->t('%s shared »%s« with you', [$filter->getSenderDisplayName($senderDisplayName), $filename]);
+			$defaults = new \OCP\Defaults();
+			$mailNotification = new MailNotifications(
```
jvillafanez: We have to use the mailer classes of the files_sharing app. This will ensure the notifications remain consistent.

apps/files_sharing/lib/MailDispatcher.php
```diff
@@ -0,0 +1,298 @@
...
+		if (\count($sendMessages) > 0) {
+			foreach ($sendMessages as $sendMessage) {
+				$this->mailer->send($sendMessage);
+			}
+		}
+
+		return $noMail;
+	}
+
+	public function execute($jobList, ILogger $logger = null) {
```
jvillafanez: This needs to be removed. The expectation is that you implement the `run` method to do whatever you need to do.

apps/files_sharing/lib/MailDispatcher.php
```diff
@@ -0,0 +1,298 @@
...
+				$this->logger->warning($nonRecipient . ' did not receive email');
+			}
+		}
+
+		$id = $this->getId();
+		$this->joblist->removeById($id);
+		//\OC::$server->getJobList()->removeById($id);
+	}
+
+	public function run($argument) {
```
jvillafanez: Keep it as protected. We don't want this function to be called from outside, but from the execute method

apps/files_sharing/lib/MailDispatcher.php
```diff
@@ -0,0 +1,298 @@
...
+		//$this->mailer = \OC::$server->getMailer();
+
+		if (($this->argument['recipient'] === null) &&
+			($this->argument['sender'] === null) &&
+			($this->argument['shareType'] === null) &&
+			($this->argument['itemSource']) === null &&
+			($this->argument['itemType'] === null)) {
+			return null;
```
jvillafanez: Better log something here. I'd prefer to know that a job hasn't run properly for some reason, than ask myself several days why the job isn't working.

apps/files_sharing/lib/MailDispatcher.php
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * Class MailDispatcher
+ *
+ * @package OCA\Files_Sharing
+ */
+class MailDispatcher extends QueuedJob {
+
+	/** @var \OC\User\Manager|IUserManager */
+	protected $userManager;
```
jvillafanez: As a general rule, attributes should be always private. If you expect that a subclass should have access to the attribute, then you can make it protected, but you won't subclass this class, so there is no reason to make the attributes protected.

DeepDiver1975:

PR #32097, opened 2018-07-19
allow public uploads to fail when expected

individual-it:
## Description
the expected status code should not be checked in `publicUploadContent()` because:
1. this function is mostly called by When and Given steps
2. it does limit the usability
3. every scenario that cares about the status code should check it in a Then step

So its better to catch the bad codes there and to set `$this->response`, the way its done in other places also

## Motivation and Context
make the function be more general and more useful

## How Has This Been Tested?
made antivirus tests where public uploads should not be possible

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -450,17 +450,16 @@ trait Sharing {
 	 * @return void
 	 */
 	public function publiclyUploadingShouldNotWork() {
-		try {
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
I agree. "When" steps should do the action without caring if it fails. "Then" steps should check for the expected result.

individual-it:
@ownclouders rebase

individual-it:
backport in #32101

PR #32104, opened 2018-07-20
[stable10] Bump paragonie/random_compat v2.0.17 react/promise v2.7.0

phil-davis:
composer version bumps that are waiting


phil-davis:
I suspect the ``react/promise`` bump in here breaks something.

PR #32105, opened 2018-07-20
composer update version bumps as at 20180720

phil-davis:


phil-davis:
I suspect the ``react/promise`` bump in here breaks something.

PR #32181, opened 2018-07-27
Fix Mandatory value in occ

### LINKED ISSUE #32046: Set argument value as mandatory for occ config:app:set command
screen shot 2018-07-13 at 13 07 14

davitol:
Fixes https://github.com/owncloud/core/issues/32046

### Tested

manually checking the behavior and the insert in the database in oc_appconfig table;

screen shot 2018-07-27 at 13 39 11

### DIFF
--- a/core/Command/Config/App/SetConfig.php
+++ b/core/Command/Config/App/SetConfig.php
@@ -56,10 +56,9 @@ class SetConfig extends Base {
 				InputArgument::REQUIRED,
 				'Name of the config to set.'
 			)
-			->addOption(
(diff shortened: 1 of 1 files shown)


PVince81:
ok, now I realize that I misunderstood the ticket, and understand better you asked me in the chat about backwards compatibility.

I thought it was about "--value" being optional, but from the diff I see it was already required.

Here with the PR you get rid of it completely.
Ok, I'll close this and we need to reconsider the decision.

Sorry for the confusion, still happy to see a PR from you :-D

PR #32175, opened 2018-07-27
Remove no longer needed repair step MoveAvatarOutsideHome

DeepDiver1975:
## Description
This repair step is broken on master because of https://github.com/owncloud/core/pull/31105

Assuming we no longer need this repair step - this PR kills it.
Otherwise we need to fix this.

## Motivation and Context
https://drone.owncloud.com/owncloud/update-testing/94/64
screenshot from 2018-07-27 09-26-24

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32164, opened 2018-07-26
Revert API breakage in IStorage

DeepDiver1975:
## Description
https://github.com/owncloud/core/pull/32051 was possible a bit too aggressive ...

## Motivation and Context
We shall not break APIs for apps

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32193, opened 2018-07-30
Revert "Introduce IStorage based Node, Folder & File, make avatar manager wor…"

DeepDiver1975:
Reverts owncloud/core#31105

refs #32175


phil-davis:
And means #32164 will not be needed - true?

DeepDiver1975:
> And means #32164 will not be needed - true?

no  - we have to different errors caused by two different PR we merged almost the same time.
Unfortunately they both have 'IStorage' in the name .... very :confused:

PR #32212, opened 2018-08-01
Add support in the actions to mark if the clients should redirect

### LINKED ISSUE #212: Support for redirect actions
Have the ability to specify a URL for action buttons.

I believe there is already a URL there so maybe just a flag to tell the handler that the URL is not ajax but really a page that needs to be opened.

Then adjust the JS notification handler to work as such.

Would also need adjusting in clients.
We could also embed those in emails.

@jvillafanez @michaelstingl thoughts ?

jvillafanez:
## Related Issue

- Fixes (required for https://github.com/owncloud/notifications/issues/212)

## Motivation and Context
Adding this support is required to fix https://github.com/owncloud/notifications/issues/212

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Notification/Action.php
+++ b/lib/private/Notification/Action.php
@@ -43,6 +43,9 @@ class Action implements IAction {
 	/** @var bool */
 	protected $primary;
 
+	/** @var bool */
+	protected $redirect = false;
+
 	/**
 	 * Constructor
 	 */
@@ -154,6 +157,22 @@ class Action implements IAction {
 	 */
 	public function getRequestType() {
 		return $this->requestType;
+	}
+
+	/**
+	 * @param bool $mark
+	 * @return $this
+	 */
+	public function setRedirect($mark) {
+		$this->redirect = $mark;
+		return $this;
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function getRedirect() {
+		return $this->redirect;
 	}
 
 	/**
--- a/lib/private/Notification/Action.php
+++ b/lib/private/Notification/Action.php
@@ -43,9 +43,6 @@ class Action implements IAction {
 	/** @var bool */
 	protected $primary;
 
-	/** @var bool */
-	protected $redirect = false;
-
 	/**
 	 * Constructor
 	 */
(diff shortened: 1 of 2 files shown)


lib/private/Notification/Action.php
```diff
@@ -52,6 +55,7 @@ public function __construct() {
 		$this->link = '';
 		$this->requestType = '';
 		$this->primary = false;
+		$this->redirect = false;
```
DeepDiver1975: initialize the field itself - no need to set this in ctor.

lib/public/Notification/IAction.php
```diff
@@ -91,6 +91,22 @@ public function getLink();
 	 */
 	public function getRequestType();
 
+	/**
+	 * @param bool $mark mark if the client should check for a "redirectTo"
+	 * key in the response of the request
+	 * @return $this
+	 * @throws \InvalidArgumentException if the $mark is invalid
+	 * @since 10.0.10
```
DeepDiver1975: @PVince81 10.0.10 - was that agreed on? Just asking ... this change has implications to core + 2 apps. 

Didn't we want 10.0.10 to be a bug fix release only? again - just asking. THX

DeepDiver1975 (changes requested):

lib/public/Notification/IAction.php
PVince81: this is currently just a POC, we still should agree on the approach proposed in https://github.com/owncloud/notifications/issues/212

I don't think there was a decision about 10.0.10 being bug fix release only. However from the list of open tasks it will likely be a "90% bugfixing" and "10% new stuff" release so far.

The above is mostly needed to resolve the redirect actions problem which is a tech debt introduced by the password policy, I hoped we could iron it out for 10.0.10 instead of letting it get forgotten.

PVince81:
I'm closing this for now until there is interest in continuing https://github.com/owncloud/notifications/issues/212.
we can reopen and adjust once we have an agreement on the approach and also a commitment to make it happen with all clients.

PR #32222, opened 2018-08-02
#32221 Set permissions on log file creation instead of every write.

### LINKED ISSUE #32221: Arbitrary setting of file mode on owncloud.log
### Steps to reproduce
1. chmod owncloud.log file to a mode other than 0640.
2. Cause an event, of any kind, that is logged.
3. See owncloud.log file chmod'ed back to 0640.

### Expected behaviour
It isn't ownCloud's job, when logrotation is disabled, to manage the file permissions. I am managing it via logrotate and other methods, as we use filebeats to take the json log and feed it into elasticsearch.  I'm not keen to add filebeat to the php user group, as that would mean I'm expanding the footprint of possible users that can write to data in certain scenarios.

I would suggest that the chmod only be performed upon file rotation when managed by ownCloud itself. 

### Actual behaviour
ownCloud keeps reverting the log file permissions to 0640, as per https://github.com/owncloud/core/blob/master/lib/private/Log/Owncloud.php#L132

Not to mention the ownCloud instance is performing needless busy work on every log event.

### Server configuration
**Operating system**: RHEL 7.5
**Web server:** Apache 2.4.6
**Database:** MariaDB 10.0 w/Galera
**PHP version:** 7.0 from the Remi repo
**ownCloud version:** 10.0.9
**Updated from an older ownCloud or fresh install:** Not relevant
**Where did you install ownCloud from:** Git

mdusher:
## Description
Moves the chmod for the log files to a file creation method so that it is only run on file creation rather than every write. This allows the administrator of the ownCloud instance to set their own permissions if required.

## Related Issue
- Fixes #32221

## Motivation and Context
To allow other applications to read the file without having to provide that application permissions to all files that are owned by the web server.

## How Has This Been Tested?
No.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Log/Owncloud.php
+++ b/lib/private/Log/Owncloud.php
@@ -51,14 +51,8 @@ class Owncloud {
 		 * Fall back to default log file if specified logfile does not exist
 		 * and can not be created.
 		 */
-		if (!\file_exists(self::$logFile)) {
-			if (!\is_writable(\dirname(self::$logFile))) {
-				self::$logFile = $defaultLogFile;
-			} else {
-				if (!\touch(self::$logFile)) {
-					self::$logFile = $defaultLogFile;
-				}
-			}
+		if (!self::createLogFile(self::$logFile)) {
+			self::$logFile = $defaultLogFile;
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
@mdusher please adopt the coding style - see https://drone.owncloud.com/owncloud/core/9140/69

lib/private/Log/Owncloud.php
```diff
@@ -143,6 +137,23 @@ public static function writeExtra($app, $message, $level, $conditionalLogFile, $
...
 
+	/**
+	 * create a log file and chmod it to the correct permissions
+	 * @param string $logFile
+	 * @return boolean
+	 */
+	public static function createLogFile($logFile) {
+		if (\file_exists($logFile) ) {
+			return true;
+		} else {
```
DeepDiver1975: no need for this else because the if above returns

PVince81:
@DeepDiver1975 please recheck, seems there's an update in the code

@mdusher please always ping after changes for re-review as Github doesn't send notifications on pushes

PVince81:
@DeepDiver1975 any more concerns ?

PVince81:
@DeepDiver1975 please rereview

DeepDiver1975 (approved):

PR #32226, opened 2018-08-02
#32217 Change the file locking on large uploads

### LINKED ISSUE #32217: Change the file locking for chunked uploads
### Steps to reproduce (difficult in this case)
1. Upload large file via desktop client.
2. Slow down the final assembly process of the chunked file.
3. Let the client run into timeout.
4. Client resends the last PUT, starts a second assembly process.
5. Second assembly process crashes the first one
6. This repeats over and over
7. Finally, one process assembles the file
8. Client shown multiple error messages (Timeout)
9. Client somehow follows up and gets in sync again.

### Expected behaviour
Second assembly process should fail, because of proper file locking.

### Actual behaviour
The repeating requests from the client interfere the assembly process

### Server configuration
**Operating system**:
Ubuntu 16.04
**Web server:**
Apache
**Database:**
MySQL
**PHP version:**
7.1
10.0.9

### Logfiles
Logfiles are at the corresponding enterprise issue.

## Possible Solution
Try to get an exclusive lock on the destination file before we start the assembly process in https://github.com/owncloud/core/blob/eda2a55adfdecbe6af0f9e8ad589f878b64c1912/apps/dav/lib/Connector/Sabre/File.php#L191

I tried to move this up
https://github.com/owncloud/core/blob/eda2a55adfdecbe6af0f9e8ad589f878b64c1912/apps/dav/lib/Connector/Sabre/File.php#L234-L241

This improves the behaivour by letting the first process succeed.

micbar:
## Description
When uploading large files >10GB, the assembly process on the server runs very long (>5minutes). This causes a timeout on the client side (Version 2.4.2). The setup is a clustered web server behind a F5 load balancer.
New move Requests by the client, let the running assembly processes fail.

## Related Issue

- Fixes #32217 

## Motivation and Context
Created PR to test if we can make the file locking process more robust

## How Has This Been Tested?
- test environment: Local docker images with Desktop client 2.4.2
- test case 1: Upload 10GB file, while slowing down the assembly process (sleep()) to produce timout at the client side

## Types of changes

- [x] Changed Unit Tests

## Checklist:

- [x] Code changes

## Open tasks:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -182,14 +182,29 @@ class File extends Node implements IFile, IFileNode {
 		/** @var \OC\Files\Storage\Storage $storage */
 		list($storage, $internalPath) = $this->fileView->resolvePath($this->path);
 		try {
+			try {
+				$this->changeLock(ILockingProvider::LOCK_EXCLUSIVE);
+			} catch (LockedException $e) {
+				if ($needsPartFile) {
+					$partStorage->unlink($internalPartPath);
+				}
+				throw new FileLocked($e->getMessage(), $e->getCode(), $e);
+			}
 			$target = $partStorage->fopen($internalPartPath, 'wb');
 			if (!\is_resource($target)) {
 				\OCP\Util::writeLog('webdav', '\OC\Files\Filesystem::fopen() failed', \OCP\Util::ERROR);
 				// because we have no clue about the cause we can only throw back a 500/Internal Server Error
 				throw new Exception('Could not write file contents');
 			}
+
 			list($count, $result) = \OC_Helper::streamCopy($data, $target);
 			\fclose($target);
+
+			try {
+				$this->changeLock(ILockingProvider::LOCK_SHARED);
+			} catch (LockedException $e) {
+				throw new FileLocked($e->getMessage(), $e->getCode(), $e);
+			}
 
 			if (!self::isChecksumValid($partStorage, $internalPartPath)) {
 				throw new BadRequest('The computed checksum does not match the one received from the client.');
--- a/apps/dav/tests/unit/Connector/Sabre/FileTest.php
+++ b/apps/dav/tests/unit/Connector/Sabre/FileTest.php
@@ -217,7 +217,9 @@ class FileTest extends TestCase {
 		// action
 		$caughtException = null;
 		try {
+			$file->acquireLock(ILockingProvider::LOCK_SHARED);
 			$file->put('test data');
+			$file->releaseLock(ILockingProvider::LOCK_SHARED);
 		} catch (\Exception $e) {
 			$caughtException = $e;
 		}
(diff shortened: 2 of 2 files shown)


DeepDiver1975 (changes requested):
Please fix unit tests - rest :+1:

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -188,9 +188,25 @@ public function put($data) {
 				// because we have no clue about the cause we can only throw back a 500/Internal Server Error
 				throw new Exception('Could not write file contents');
 			}
+
+			try {
+				$this->changeLock(ILockingProvider::LOCK_EXCLUSIVE);
```
PVince81: please move this before `fopen` to avoid having two concurrent processes opening the file at the same time

micbar: are you sure? If i do this, tho whole operations fails.
Is it possible to lock a file before opening?

PVince81: how exactly does it fail ? it shouldn't fail as the storage fopen shouldn't lock anything. Maybe there is something else that is suspicious there...

in general if we leave it like you did then a concurrent process would be able to lock the target file, but would fail after fopen and before starting stream copy. so the error handler would need to close the file again, which I think will happen anyway.

would still be good to investigate why moving the lock before makes the operation fail, in case there are some ghosts hidden somewhere...

micbar:
@DeepDiver1975 @PVince81 
I had to adjust the testing of the put method.

As seen in https://github.com/owncloud/core/blob/eda2a55adfdecbe6af0f9e8ad589f878b64c1912/apps/dav/lib/Connector/Sabre/LockPlugin.php#L64
the put() is executed with a shared lock.

Please review and ping back, i get the impression, that this is a very sensitive part :-)

DeepDiver1975 (approved):

phil-davis:
Backport ``stable10`` #32334

PVince81:
@micbar do you remember where exactly you put the `sleep` statements ? this is for expanding the tests for manual QA to verify

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -182,15 +182,30 @@ public function put($data) {
...
+				}
+				throw new FileLocked($e->getMessage(), $e->getCode(), $e);
+			}
 			$target = $partStorage->fopen($internalPartPath, 'wb');
 			if (!\is_resource($target)) {
 				\OCP\Util::writeLog('webdav', '\OC\Files\Filesystem::fopen() failed', \OCP\Util::ERROR);
 				// because we have no clue about the cause we can only throw back a 500/Internal Server Error
 				throw new Exception('Could not write file contents');
 			}
+
```
micbar: I had a sleep() statement after setting the exclusive lock and before the final move operation.

PVince81: @micbar which value did you use ? how high to set it to have the desktop client time out on its own ?

PVince81: `sleep(300)' in both locations

micbar: The desktop client resends the MOVE request after 300sec.

PR #32232, opened 2018-08-03
Files_sharing Dutch translation Sharing State

T0mWz:
Updated Dutch translation for OC files_sharing sharing state


DeepDiver1975:
We need to backport translations for 10.0.10 as well then ....

PR #32239, opened 2018-08-06
Standardize acceptance test ocPath to always go through testing app

phil-davis:
## Description
1) Use a common ``ocPath`` in behat that points to the testing app.
2) Make the ``CommandLine.php`` ``runOcc()`` method call the ``SetupHelper::runOcc()`` method, which is the one that gets the testing app to do the work.

Item (2) maintains the ``CommandLine.php`` ``runOcc()`` method so that callers from app acceptance tests etc. should still work the same.

## Motivation and Context
1) Methods in API test contexts use a ``runOcc`` that directly runs an occ command.
2) Methods in webUI test contexts use a ``runOccc`` that sends a request to the testing app, and the testing app runs the ``occ`` command and returns the status and output.

Method (1) only works if the test script is running on/in the same system as the SUT.
Method (2) is more general - and we require the testing app to be available anyway for other setup and teardown stuff in scenarios.

Do things the same way in API and webUI tests.

## How Has This Been Tested?
Some loca; API nad webUI test runs. Let CI run the whole suite of tests.

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes

## Open tasks:

### DIFF
--- a/tests/TestHelpers/SetupHelper.php
+++ b/tests/TestHelpers/SetupHelper.php
@@ -268,6 +268,7 @@ class SetupHelper {
 	 *                    For example: "files:transfer-ownership"
 	 *
 	 * @return string[] associated array with "code", "stdOut", "stdErr"
+	 * @throws Exception if ocPath has not been set yet or the testing app is not enabled
 	 */
 	public static function runOcc($args) {
 		if (self::$adminUsername === null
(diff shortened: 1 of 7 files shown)


individual-it (approved):

phil-davis:
@individual-it please review again. CI should pass now. I  fixed up the "token auth enforced" stuff so that it can undo itself by generating and using an app password to send to the testing app. The ``runOcc`` methods have been modified to take optional parameters that can override anything that was initially specified in ``SetupHelper::init``

individual-it (approved):
LGTM, lets see what drone says

phil-davis:
Backport ``stable10`` #32257

PR #32244, opened 2018-08-06
separate given step to create user from when step

individual-it:
## Description
- Given steps should not been forced to use a specific way (API/OCC/webUI) to achieve a specific state
- Given steps don't need to be checked, we assume they work

## Motivation and Context
when using this step for LDAP tests it fails because in LDAP we do not create users, but skip the user creation and believe they exist. Still the given step should work and not throw an exception.

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiMain/dav-versions.feature
+++ b/tests/acceptance/features/apiMain/dav-versions.feature
@@ -51,8 +51,7 @@ Feature: dav-versions
     Then the version folder of fileId "<<FILEID>>" for user "user1" should contain "1" element
 
 	Scenario: sharer of a file can see the old version information when the sharee changes the content of the file
-		Given user "user0" has been created
-		And user "user1" has been created
+		Given user "user1" has been created
 		And user "user0" has uploaded file with content "user0 content" to "sharefile.txt"
 		And user "user0" has shared file "sharefile.txt" with user "user1"
 		When user "user1" has uploaded file with content "user1 content" to "/sharefile.txt"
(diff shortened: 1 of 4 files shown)


phil-davis (approved):

individual-it:
backport in  #32246

phil-davis:
```
Given steps don't need to be checked, we assume they work
```
Actually I would like to see that ``Given`` steps check themselves a little bit (e.g. they check the HTTP status) so that they do explode if they obviously go wrong. That makes it a little bit easier when something goes wrong in the environment - you see the scenario explode early, rather than thinking that a ``Then`` test failed because there is a real bug.

``When`` steps should not check anything really (maybe only explode if the whole network is down). It is best if it is one of the ``Then``` steps that fails - then it should be easier to understand exactly which thing went wrong.

But in the case here, we really do have expected ``Given`` cases where we "assume" the action happened, without any evidence.

individual-it:
I guess in this case we could check if the user really exists, that should work on local and on external userbackends. But currently `theUserShouldExist()` does not do really anything

phil-davis:
It's fine like it is for now. Some day I would like to do a general refactoring:

1) Separate all ``Given``` and ``When``
2) Make all the ``Given`` have some sort of "reasonably think it worked" check built-in to the method
3) Remove bonus checks from ``When`` steps, in most cases. Let it 99% pass through to be the responsibility of ``Then`` checks to confirm the ``When`` statements under test have worked.

phil-davis:
I added an extra commit to remove various duplicated "Given user has been created" steps that now cause failures. The commit has also been added to the backport.

PR #32248, opened 2018-08-06
Properly reset the checksum if the file data has changed

jvillafanez:
## Description
Properly reset the checksum if the file has changed

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/2561

## Motivation and Context
Checksum wasn't being reset in that scenario. The previous condition was wrong

## How Has This Been Tested?

Manually tested with the scenario described in the issue
- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files/tests/Command/VerifyChecksumsTest.php
+++ b/apps/files/tests/Command/VerifyChecksumsTest.php
@@ -135,10 +135,10 @@ class VerifyChecksumsTest extends TestCase {
 
 	private function assertChecksumsAreCorrect(array $files) {
 		foreach ($files as $key => $file) {
-			/** @var File $f */
-			$f = $files[$key]['file'];
 			$expectedChecksums = $files[$key]['expectedChecksums'];
 			$this->refreshFileInfo($files[$key]['file']);
+			/** @var File $f */
+			$f = $files[$key]['file'];
 			$this->assertSame(
 				$expectedChecksums(),
 				$f->getChecksum()
--- a/lib/private/Files/Cache/Scanner.php
+++ b/lib/private/Files/Cache/Scanner.php
@@ -205,9 +205,12 @@ class Scanner extends BasicEmitter implements IScanner {
 					}
 					if (!empty($newData)) {
 						// Only reset checksum on file change
-						foreach (\array_intersect_key($newData, $data) as $key => $value) {
-							if (\in_array($key, ['size', 'storage_mtime', 'mtime', 'etag']) && $data[$key] != $newData[$key]) {
-								$newData['checksum'] = '';
+						if ($fileId > 0 && !isset($newData['checksum'])) {
+							foreach (['size', 'storage_mtime', 'mtime', 'etag'] as $dataKey) {
+								if (isset($newData[$dataKey])) {
+									$newData['checksum'] = '';
+									break;
+								}
 							}
 						}
 


PVince81:
@jvillafanez would be good to add a unit test

at first look I have a hard time to understand what you actually changed, might explain the difference before vs after ?

jvillafanez:
The code is fundamentally the same: if the size, storage_mtime, mtime or etag has changed, then reset the checksum. The new code is better because the old one used `array_intersect_key` and `in_array` while the new one just uses `isset`

The problem was that the `$data[$key] != $newData[$key]` condition wasn't working properly because the values where the same, so the checksum wasn't being reset.

PVince81:
@jvillafanez will this change have other side effects ?

would be good to check and extend the unit tests to confirm this works correctly

jvillafanez:
There are still some unittests failing, but I'm not sure where the problem is. Updated code still works and makes sense, so there shouldn't be a problem with the changes but the checksum repair tests are failing. I'm not sure if there are some problems with the tests.

IljaN:
@jvillafanez Try breaking the checksum intentionally and then repairing it via repair command to see if checksum repair works as expected. If this is the case we can debug the test to see what is going wrong

jvillafanez:
I've modifed the data in the DB and it seems the command works.

If the checksum has been reset:
```
sudo -u www-data ./occ files:checksums:verify -vvv -r
.....
Skipping synctest.log => No Checksum
```
If not:
```
sudo -u www-data ./occ files:checksums:verify -vvv -r
....
Checking synctest.log => SHA1:73cbb1d69ba106bf6fe1d7ec1111111111111111 MD5:d451afb7ec8b2aaa1111111111111111 ADLER32:d2999111
Mismatch for synctest.log:
 Filecache:	SHA1:73cbb1d69ba106bf6fe1d7ec1111111111111111 MD5:d451afb7ec8b2aaa1111111111111111 ADLER32:d2999111
 Actual:	SHA1:73cbb1d69ba106bf6fe1d7ec5e3a8214d1f15f81 MD5:d451afb7ec8b2aaa14389bbbf7ce15c4 ADLER32:d2999e73
Repairing synctest.log
```

IljaN (approved):

DeepDiver1975:
I assume this requires backport ....

DeepDiver1975:
> I assume this requires backport ....

looks like no needed - @jvillafanez please clarify - thx

DeepDiver1975:
> stable10 is required - hence blue-ticket !

upon cherry-pick I get an empty commit ... so I assume this was merged with some other pr already ...

this is why I was asking for clarification.

jvillafanez:
Backport was https://github.com/owncloud/core/pull/32284 (already merged)

PVince81:
how to test this manually, is there a shortcut ?

Steps I tried:
1. Setup SFTP mount as "/sftp"
1. Create a file "/sftp/test.txt"
1. Download the file to generate a checksum
1. Check checksum in oc_filecache
1. Directly on the SFTP storage, do a `touch` with a date in the past
1. In web UI, refresh the folder again

Result: the mtime of the file changed in the web UI

On both v10.0.9 and v10.0.10RC1 the checksum gets cleared properly.

Are there any more subtle things to setup ? @jvillafanez

jvillafanez:
Other than the steps in the original ticket I'm not sure if this can be easily reproduced. I think the key point is to have the same checksum with different etag and size over the same file in different moments.

PR #32260, opened 2018-08-07
Add method to clear notification when user is deletd

sharidas:
Added a method to clear notification entry of user
when the user is deleted from oC.

## Description

User entries from notification table should be deleted, once the user is deleted from oC.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

User entries from notification table should be deleted, once the user is deleted from oC.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Requires https://github.com/owncloud/notifications/pull/221

## Open tasks:

### DIFF
--- a/lib/public/Notification/IApp.php
+++ b/lib/public/Notification/IApp.php
@@ -49,4 +49,11 @@ interface IApp {
 	 * @since 9.0.0
 	 */
 	public function getCount(INotification $notification);
+
+	/**
+	 * @param string $uid uid of the user
+	 * @return null
+	 * @since 10.0.10
+	 */
+	public function removeUserNotifications($uid);
 }
--- a/lib/private/Notification/Manager.php
+++ b/lib/private/Notification/Manager.php
@@ -22,6 +22,7 @@
 
 namespace OC\Notification;
 
+use OCP\IUser;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use OCP\Notification\IApp;
 use OCP\Notification\IManager;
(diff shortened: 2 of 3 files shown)


sharidas:
Fixes -> https://github.com/owncloud/password_policy/issues/69

lib/public/Notification/IApp.php
```diff
@@ -49,4 +49,11 @@ public function markProcessed(INotification $notification);
 	 * @since 9.0.0
 	 */
 	public function getCount(INotification $notification);
+
+	/**
+	 * @param $uid
+	 * @return null
+	 * @since 10.1.0
```
PVince81: I think you can put 10.0.10 here

PVince81: PHPDoc is incomplete

PVince81 (changes requested):
see comment

jvillafanez:
What are the reasons to provide the solution here?

This implies that the core's notification manager knows that the notification is stored by the apps, which implies that is expected that the apps store notifications somewhere, which leads to store duplicate information (by different apps).

Core doesn't need to know whether the notifications are being stored. If the notifications app or any other provider stores notifications it's their own responsability to clean them.

I'd prefer to move the solution to the notification app: just listen the delete event there and remove the stored notifications. No need to change the core's interface.

sharidas:
@jvillafanez 
> I'd prefer to move the solution to the notification app: just listen the delete event there and remove the stored notifications. No need to change the core's interface.

In fact you are right, the core should not bother about the changes I made in this PR. The notification app should listen to the event and then do the action, i.e, to remove the entry of the user from the table.
Thanks for the suggestion.

PVince81:
@sharidas please link to your new PR in the notifications app then close

sharidas:
The notification PR https://github.com/owncloud/notifications/pull/221 is sufficient to achieve the result. Hence closing this PR.

PR #32274, opened 2018-08-09
Only restore capabilities in acceptance tests if the previous settings were saved

phil-davis:
## Description
Only call ``modifyServerConfigs`` when doing ``restoreParametersAfterScenario`` if there actually were some saved parameters.

## Motivation and Context
Some app acceptance tests do not currently set any capabilities.
The code that restores the capabilities explodes with an undefined index when no capabilities were set.
It is good not to explode.

## How Has This Been Tested?
CI knows.

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/AppConfiguration.php
+++ b/tests/acceptance/features/bootstrap/AppConfiguration.php
@@ -384,7 +384,9 @@ trait AppConfiguration {
 		$previousServer = $this->currentServer;
 		foreach (['LOCAL','REMOTE'] as $server) {
 			$this->usingServer($server);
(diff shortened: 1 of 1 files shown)


phil-davis:
In some apps I have been forced to add some "default" capabilities setup in order to workaround this undefined index thing. Actually it is probably not so bad to do that in app tests (it makes sure some default settings are in place, e.g. when an app makes use of some sharing steps then "default" file sharing capabilities had better be enabled). But also it is better not explode unnecessarily.

individual-it (approved):

phil-davis:
Backport ``stable10`` #32276

PR #32283, opened 2018-08-09
Reset stalled state in upload data after retrying

### LINKED ISSUE #32282: Upload stall timeout detection not accurate
### Steps

1. `occ config:app:set files max_chunk_size --value 1024`
1. `occ config:app:set files upload_stall_timeout --value 3`
1. Set web server to sleep for one minute randomly during chunk upload, see code below 
1. Upload a 10Kb file (10 chunks)

Code to make server randomly delay, edit "apps/dav/lib/Connector/Sabre/Directory.php" and replace `createFileDirectly` with the code below:
```php
        public function createFileDirectly($name, $data) {
                $tgt = $this->fileView->fopen($this->getPath() . '/' .  $name, 'wb');

                $count = 0;
                while(!feof($data)) {
                        if (mt_rand(0,99) <= 10) {
                                sleep(60);
                        }
                        $count = fwrite($tgt, fread($data, 512));
                }
                fclose($data);
                fclose($tgt);
        }
```

### Expected result
Stall timeout detected after 3 seconds, then chunk is retried.
Stall timeout must not be tied to progress bar.

### Actual result
Stall timeout detected after 1 minute, which is when the server resumes sending data after the `sleep()` and moves the progress bar.

### Versions
OC 10.0.9 + patch from https://github.com/owncloud/core/pull/32170

@pmaier1 FYI

@butonic not sure why you originally decided to tie the stall timeout to the progress bar event. It doesn't seem to work on some environments where the server stalls forever and never sends any bytes, so the progress bar never moves and the stall timeout is never detected or detected very late when the user already gave up.

PVince81:
## Description
After retrying a chunk, we need to reset the upload object's "stalled" state, else the next stalled chunk is not retried.

## Related Issue
Fixes https://github.com/owncloud/core/issues/32282

## How Has This Been Tested?
See steps in original ticket as it needs patching

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [x] add unit tests

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -1124,6 +1124,7 @@ OC.Uploader.prototype = _.extend({
 									});
 
 									// clear the previous data:
+									upload.data.stalled = false;
 									data.data = null;
 									// overwrite chunk
 									delete data.headers['If-None-Match'];
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -854,7 +854,8 @@ OC.Uploader.prototype = _.extend({
 			if (progress >= total) {
 				// change message if we stalled at 100%
 				this.$uploadprogressbar.find('.label .desktop').text(t('core', 'Processing files...'));
-			} else if (new Date().getTime() - this._lastProgressTime >= this._uploadStallTimeout * 1000 ) {
+			}
+			if (new Date().getTime() - this._lastProgressTime >= this._uploadStallTimeout * 1000 ) {
 				// TODO: move to "fileuploadprogress" event instead and use data.uploadedBytes
 				// stalling needs to be checked here because the file upload no longer triggers events
 				// restart upload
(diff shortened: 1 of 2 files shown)


PVince81:
added unit test

PVince81:
- [ ] BUG: retries never count up

apps/files/js/file-upload.js
```diff
@@ -1124,6 +1125,7 @@ OC.Uploader.prototype = _.extend({
 									});
 
 									// clear the previous data:
+									upload.data.stalled = false;
```
butonic: ok, this makes sense

apps/files/js/file-upload.js
```diff
@@ -854,7 +854,8 @@ OC.Uploader.prototype = _.extend({
 			if (progress >= total) {
 				// change message if we stalled at 100%
 				this.$uploadprogressbar.find('.label .desktop').text(t('core', 'Processing files...'));
-			} else if (new Date().getTime() - this._lastProgressTime >= this._uploadStallTimeout * 1000 ) {
+			}
```
butonic: ah ... yes also makes sense

butonic (approved):

phil-davis:
Backport ``stable10`` #32283

PR #32287, opened 2018-08-09
Move acceptance test skeleton dirs into apps/testing/data

phil-davis:
## Description
1) Move ``tests/acceptance/skeleton`` to ``apps/testing/data/apiSkeleton``
2) Move ``tests/acceptance/webUISkeleton`` to ``apps/testing/data/webUISkeleton``
3) Change core references to the skeleton folder location

## Related Issue
https://github.com/owncloud/QA/issues/580

## Motivation and Context
We want to work towards having the testing "app" as something that can be installed on a system and the system gets "the stuff it needs internal to support the running of the acceptance tests".

One step is to put the "standard skeleton dirs" used by the acceptance tests inside the testing "app". That will make these be available whenever the testing "app" is in place. So we do not need to find another way to deliver the skeleton dirs.

## How Has This Been Tested?
CI knows

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes

## Open tasks:


phil-davis:
@individual-it I think it is this easy in core to move the skeleton dir location.

There are a bunch of app ``.drone.yml`` that mention the path to ``webUISkeleton`` - those will need fixing or removing. I will see if I can just remove them, because the common ``run.sh`` should now do something sensible by default.

phil-davis:
@individual-it is there somewhere that you did something tricky to mount the skeleton dir onto the docker container running the server?

I remember something about that but don't see it. If it still exists, then it probably can be removed.

individual-it (approved):
Tests are passing, looks good.
I don't think we need to mount the folder into docker for drone. We did stuff to mount it in the compose infrastructure

phil-davis:
Backport ``stable10`` #32288

PR #32292, opened 2018-08-10
fix suites-to-run calculation in run.sh acceptance tests script

individual-it:
## Description
in special cases the `$SUITES_PER_RUN` variable might end up being `0` and then the script tries to run all suites, what is not the  goal.
This happens e.g. if we have 16 suites, divide by 5 and run the last part. For run 1-4 `$SUITES_PER_RUN` would be `4 = ceiling(16/5)` but after 4 runs all the work is done, so there is no need to run the last division.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -483,10 +483,17 @@ else
 		#divide the suites and round up
 		SUITES_PER_RUN=$(((${COUNT_ALL_SUITES} + ${DIVIDE_INTO_NUM_PARTS} - 1) / ${DIVIDE_INTO_NUM_PARTS}))
 		COUNT_FINISH_AND_TODO_SUITES=$((${RUN_PART} * ${SUITES_PER_RUN}))
+		COUNT_FINISH_SUITES=$(((${RUN_PART} - 1) * ${SUITES_PER_RUN}))
 		if [ ${RUN_PART} -eq ${DIVIDE_INTO_NUM_PARTS} ]
 		then
 			#the last run might have less suites
-			SUITES_PER_RUN=$((${COUNT_ALL_SUITES} - (${RUN_PART} - 1) * ${SUITES_PER_RUN}))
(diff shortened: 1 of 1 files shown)


individual-it:
backport in https://github.com/owncloud/core/pull/32293

tests/acceptance/run.sh
```diff
@@ -483,10 +483,17 @@ else
...
 		SUITES_PER_RUN=$(((${COUNT_ALL_SUITES} + ${DIVIDE_INTO_NUM_PARTS} - 1) / ${DIVIDE_INTO_NUM_PARTS}))
 		COUNT_FINISH_AND_TODO_SUITES=$((${RUN_PART} * ${SUITES_PER_RUN}))
+		COUNT_FINISH_SUITES=$(((${RUN_PART} - 1) * ${SUITES_PER_RUN}))
 		if [ ${RUN_PART} -eq ${DIVIDE_INTO_NUM_PARTS} ]
 		then
 			#the last run might have less suites
-			SUITES_PER_RUN=$((${COUNT_ALL_SUITES} - (${RUN_PART} - 1) * ${SUITES_PER_RUN}))
+			SUITES_PER_RUN=$((${COUNT_ALL_SUITES} - ${COUNT_FINISH_SUITES}))
+		fi
+		if [ ${SUITES_PER_RUN} -eq 0 ]
```
phil-davis: This can happen, for example, if there are 16 suites and we try to break them into 5 parts.
The rounding above will say we need 4 suites in each job (because 3 in each job and 5 jobs would only get 15 suites done).
But at 4 suites per job, the first 4 jobs have 4 suites each and the 5th job has none.

The algorithm could be modified so that it spreads out the suites better for this kind of case  - 4,3,3,3,3

But for now, this change will catch the issue and handle it.

phil-davis (approved):

PR #32296, opened 2018-08-10
Allow 'No scenarios' behat test runs to pass

phil-davis:
## Description
If a ``behat`` run exits with an error status, then check the command output. If it contains a line "No scenarios" then consider that the run has passed.

## Motivation and Context
When an acceptance test suite is run and a set of tags are requested, sometimes there might be no scenarios in that test suite that match the combination of tags requested.

When that happens, Behat reports "No scenarios" in the output and/but Behat exits with an error status. This causes the test run to fail.

But actually this is expected. e.g. when running core test suites from the ``files_primary_s3`` app, only a subset of tagged scenarios are run. In this case (and other cases) some core test suites have no scenarios that match the requested tags.

## How Has This Been Tested?
Local test runs specifying Behat tags that do not occur.

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes

## Open tasks:

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -230,7 +230,22 @@ function run_behat_tests() {
 	then
 		PASSED=true
 	else
-		PASSED=false
+		# If there were no scenarios in the requested suite or feature that match
+		# the requested combination of tags, then Behat exits with an error status
+		# and reports "No scenarios" in its output.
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Backport ``stable10`` #32297

PR #31957, opened 2018-06-30
Dont UPDATE SET twice with filecache, use UPDATE SET CASE instead

mrow4a:
### Description

I have noticed, that in Propagator we had to use UPDATE twice, in order to filter out scenario with unset filecache size entry. The queries were as below:

```
UPDATE `oc_filecache` SET `mtime` = GREATEST(`mtime`, :dcValue1), `etag` = :dcValue2 WHERE (`storage` = :dcValue3) AND (`path_hash` IN ('hash1', 'hash2’))

UPDATE `oc_filecache` SET `size` = `size` + :dcValue1 WHERE (`storage` = :dcValue2) AND (`path_hash` IN ('hash1', 'hash2’)) AND (`size` > '-1')"
```

However, this can be optimized to 
```
UPDATE `oc_filecache` SET `mtime` = GREATEST(`mtime`, :dcValue1), `etag` = :dcValue2, `size` = CASE WHEN `size` > '-1' THEN  `size` + :dcValue3 ELSE `size` END WHERE (`storage` = :dcValue4) AND (`path_hash` IN ('hash1', 'hash2'))
```

@DeepDiver1975 @butonic @tomneedham

PR #32304, opened 2018-08-12
[stable10] Dont UPDATE SET twice with filecache, use UPDATE SET CASE instead

mrow4a:
Backport of https://github.com/owncloud/core/pull/31957


mrow4a:
Any update here ?

PVince81 (approved):
:+1:

Issue #28902, opened 2017-09-04
Remove old repair steps, add version checks


PVince81:
Review all repair steps and add a version check or a DB flag for those that do not need to run again.
This could help speed up upgrade time by disabling unneeded repair steps.

One example: "Repair\AvatarPermisisons". I believe that there was an older version where the permissions on disk could be messed up. Once fixed it shouldn't try and re-fix it for every update.
Unfortunately this one does some queries on filecache so it might slow down the update a bit.

Note that we do not want to completely delete repair steps because some will be needed when doing cross-major-versions upgrades. So keep them but add a version check or DB flag to make sure it only runs once. (the alternative would be to port them to migrations but that is likely too much work)

### Repair steps
- [ ] RepairMimeTypes: still needed I guess
- [ ] FillEtags: if we made "etag" a NOT NULL column with default value, we wouldn't need this (need to check if we have code purposefuly writing null there on init)
- [ ] CleanTags: keep or make it a background job ?
- [ ] DropOldTables: exists since OC <= 8.1, already ran
- [ ] DropOldJobs: exists since OC <= 8.1, already ran
- [ ] RemoveGetETagEntries: exists since OC 8.2, already ran
- [ ] UpdateOutdatedOcsIds: exists since OC <= 8.1, already ran, likely obsolete now with the marketplace
- [ ] RepairInvalidShares: exists since OC 8.2. TODO need to verify if the conditions that get repaired can still be recreated with recent versions
- [ ] SharePropagation: added in OC 9.0, only needed to run once , convert to migration ?
- [x] AvatarPermissions: https://github.com/owncloud/core/pull/28975
- [x] MoveAvatarOutsideHome: => keep to be able to support 8.2->future upgrades
- [ ] RemoveRootShares: added in OC 9.1. Keep, just in case, as we don't know what bug makes this happen. Might be expensive, so would be good to get rid of at some point.
- [x] RepairUnmergedShares: runs for OC <= 9.1.0 entries, keep to support 8.2->future upgrades

### Before upgrade repair steps
- [ ] InnoDB: at this point everyone should already have migrated since OC 8 or so
- [ ] SQliteAutoincrement: already since OC <= 8.1
- [ ] Collation: keep to support 8.2->future upgrades (move to migration?)
- [ ] SearchLuceneTables: deletes dupes in search lucene tables. Exists since 8.1. I suggest removing or moving it to the search lucene app itself.

@DeepDiver1975 @tomneedham @butonic @phisch

[...]

PR #32330, opened 2018-08-14
Avatar performance

butonic:
revive of https://github.com/owncloud/core/pull/31105 this time with adjusted avatar migration, see https://github.com/owncloud/core/pull/32175 and https://github.com/owncloud/core/issues/28902 why it is needed for migrating user from 8.2 to 10, directly.

@tomneedham maybe you want to migrate the avatars completely differently?

cc @DeepDiver1975

### DIFF
--- a/lib/private/Repair/MoveAvatarOutsideHome.php
+++ b/lib/private/Repair/MoveAvatarOutsideHome.php
@@ -146,7 +146,9 @@ class MoveAvatarOutsideHome implements IRepairStep {
 	 */
 	public function run(IOutput $output) {
 		$ocVersionFromBeforeUpdate = $this->config->getSystemValue('version', '0.0.0');
-		if (\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
+		$avatarMigrationStatus = $this->config->getAppValue('core', 'avatar_migration_completed', 'false');
+		if (($avatarMigrationStatus === 'false') &&
+			\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
 			$function = function (IUser $user) use ($output) {
 				$this->moveAvatars($output, $user);
 				$output->advance();
@@ -155,6 +157,9 @@ class MoveAvatarOutsideHome implements IRepairStep {
 			$output->startProgress($this->userManager->countSeenUsers());
 
 			$this->userManager->callForSeenUsers($function);
+
+			//Set this if repair step is executed
+			$this->config->setAppValue('core', 'avatar_migration_completed', 'true');
 
 			$output->finishProgress();
 		} else {
--- a/lib/private/Repair/MoveAvatarOutsideHome.php
+++ b/lib/private/Repair/MoveAvatarOutsideHome.php
@@ -21,7 +21,12 @@
  */
 namespace OC\Repair;
 
+use OC\Files\Node\File;
+use OC\NotSquareException;
+use OC\User\NoUserException;
+use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
+use OCP\Files\Storage\IStorage;
 use OCP\IDBConnection;
 use OCP\IL10N;
 use OCP\ILogger;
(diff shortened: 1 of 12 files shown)


DeepDiver1975:
conflicting + drone is red

PVince81:
any update ?

PVince81:
@butonic @tomneedham @DeepDiver1975 you guys please figure out how to continue this.

instructions for someone to take over ?

butonic:
I rebased. This PR introduces `OC\Files\Storage\Node`, `File` and `Folder` implementations that bypass the cache. In addition the `AvatarManager` passes the storage of the avatars folder to the `Avatar`, to 'unwrap' the `Folder` implementation it got. The `Avatar` than uses the cache bypassing `File` to do the actual filesystem interaction.

It is the followup for https://github.com/owncloud/core/pull/31105/files#r204521672 and also contains migration code. This is where @tomneedham may have to chime in because he had to deal with avatar migrations at length. Pun intended.

From my side if CI is green should be merged.

PVince81:
since @DeepDiver1975 was familiar with this it would be good now to do one last review pass

tomneedham:
> @tomneedham maybe you want to migrate the avatars completely differently?

No this is ok. We had previous issues with <OC10 migration - unrealted to this.

tomneedham:
> @tomneedham maybe you want to migrate the avatars completely differently?

No this is ok. We had previous issues with <OC10 migration - unrealted to this. 

Only thing I would say is that we really only want to run the avatar migration once. Repair step is a bit heavy as it is executed on each upgrade step. Should maybe mark with a flag that this has been completed.

PVince81:
ok. we can take over from here to just add a flag then.

thanks!

sharidas:
One idea is to introduce a flag in the config, after execution of method `moveAvatars()` in the file MoveAvatarOutsideHome.php, something like
```
$this->config->setSystemValue('avatarMigrationCompleted', 'true');
```
The reason I think is because avatar is not an app. So it would be nice to have a check if the config has `avatarMigrationCompleted` value `false` ( default val ) and then the version check (as in the current PR). I will create a PR based on this idea.

PVince81:
@sharidas don't use `setSystemValue` as it would put it to config.php. Use the appconfig for core instead.

sharidas:
Updated the PR with appconfig, instead of using api which would put values into config.php.

lib/private/Repair/MoveAvatarOutsideHome.php
```diff
@@ -146,9 +146,12 @@ private function moveAvatars(IOutput $out, IUser $user) {
 	 */
 	public function run(IOutput $output) {
 		$ocVersionFromBeforeUpdate = $this->config->getSystemValue('version', '0.0.0');
-		if (\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
+		$avatarMigrationStatus = $this->config->getAppValue('core', 'avatarMigrationCompleted', 'false');
+		if (($avatarMigrationStatus === 'false') &&
+			\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
 			$function = function (IUser $user) use ($output) {
 				$this->moveAvatars($output, $user);
+				$this->config->setAppValue('core', 'avatarMigrationCompleted', 'true');
```
PVince81: this would set the flag repeatedly for every user

sharidas: :man_facepalming:  Ouch. I have updated the PR by moving this code after `$function` is executed.
Care is also taken to check if migration code is executed. Added doc in the code accordingly for the same.

sharidas:
Updated the PR. 
- Used another flag to check if the repair code is applied for at least one user.
- Set appconfig flag, if the repair code is executed. Not for each user, but after it is completed.

Let me know this looks ok. @PVince81

lib/private/Repair/MoveAvatarOutsideHome.php
```diff
@@ -156,6 +165,11 @@ public function run(IOutput $output) {
 
 			$this->userManager->callForSeenUsers($function);
 
+			//Set this if repair step is executed
+			if ($repairDoneForUsers === true) {
+				$this->config->setAppValue('core', 'avatarMigrationCompleted', 'true');
```
PVince81: from what I see in oc_appconfig, the key name format is with underscores, not camel case. let's not start a new trend here

lib/private/Repair/MoveAvatarOutsideHome.php
```diff
@@ -146,8 +146,17 @@ private function moveAvatars(IOutput $out, IUser $user) {
...
+		$avatarMigrationStatus = $this->config->getAppValue('core', 'avatar_migration_completed', 'false');
+		if (($avatarMigrationStatus === 'false') &&
+			\version_compare($ocVersionFromBeforeUpdate, '9.2.0.2', '<')) {
+			/**
+			 * This is to make sure if repair code is applied to at least one user
+			 */
+			$repairDoneForUsers = false;
+			$function = function (IUser $user) use ($output, &$repairDoneForUsers) {
+				if ($repairDoneForUsers === false) {
+					$repairDoneForUsers = true;
```
PVince81: why not just set it to true after `callForSeenUsers` instead of repeatedly ?

sharidas: This iteration was caused because of my assumption with `callForSeenUsers`. I thought `callForSeenUsers` would be applied on users who logged in at least once. Unfortunately that's not the case. :man_facepalming: Thanks for the pointer.

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport

sharidas:
Backport PR: https://github.com/owncloud/core/pull/34592

PR #32342, opened 2018-08-15
skip known to fail test with encrytion

individual-it:
this test is failing with user-key encryption because of https://github.com/owncloud/encryption/issues/42


individual-it:
backport in #32343

phil-davis (approved):

phil-davis:
drone job 30 in the list never started. @individual-it have a look and kick it if you can.

PR #32347, opened 2018-08-16
Add getters for private vars of acceptance test traits

phil-davis:
## Description
Provide getters for all the "useful/interesting" private variables in the ``FeatureContext`` traits, so that apps that care can find out the results of ``FeatureContext`` actions without being forced to ``use BasicStructure`` trait.

## Motivation and Context
App acceptance tests want to add their own new step definitions/methods. Those want to use some of the methods available in core. For example, ``runOcc()``, which remembers internally the ``occ`` command exit status, ``StdOut`` and ``StdErr`` text. There are some core methods that process those looking for strings etc. But if the app needs to parse the output in some special way unique to the app, then it needs to be able to get the "raw" ``StdOut``.

Currently to do this, the app does ``use BasicStructure`` in its own ``AppContext.php`` - giving it a "private" view of all the methods and variables that would normally appear in core ``FeatureContext``. But that runs into trouble if the app needs to also refer to other core contexts like ``FederationContext``. ``FederationContext`` itself makes use of ``FeatureContext``, which forces the app to also include ``FeatureContext`` in its list of contexts. But ``FeatureContext`` also does ``use BasicStructure`` - this results in duplicate step definitions.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Auth.php
+++ b/tests/acceptance/features/bootstrap/Auth.php
@@ -54,6 +54,26 @@ trait Auth {
 	private $tokenAuthHasBeenSetTo = '';
 
 	/**
+	 * get the client token that was last generated
+	 * app acceptance tests that have their own step code may need to use this
+	 *
+	 * @return string client token
+	 */
+	public function getClientToken() {
+		return $this->clientToken;
+	}
+
+	/**
+	 * get the app token that was last generated
+	 * app acceptance tests that have their own step code may need to use this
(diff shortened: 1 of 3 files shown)


individual-it (approved):

phil-davis:
Backport ``stable10`` #32348

PR #32337, opened 2018-08-14
[objectstore] Fix checksums not being updated on modifying shared file

mrow4a:
### Issue
Failing smashbox tests with sharing on objectstore.

### Solution

SharingScanner did not update checksums in file cache for object store, since logic for obcjectstorage has not been implemented (SharingScanner assumes by default Local Filesystem). This PR adds logic for objectstorage.
@DeepDiver1975 @patrickjahns

PR #32364, opened 2018-08-17
[stable10] Fix checksums not being updated on modifying shared file for objectstore

mrow4a:
Backport of #32337


phil-davis (approved):
Works in master, and diff is same.

PR #32371, opened 2018-08-17
Remove automagic timezone changing

### LINKED ISSUE #32165: User timezone is not updated when using auth methods != credentials
### Steps to reproduce
1. Use an apache backend for example, to login to the UI

### Expected behaviour
Timezone is update from browser: https://github.com/owncloud/core/blob/a80340d7082d9839e0a2fe0dae30cc4e21d5978a/core/Controller/LoginController.php#L238

### Actual behaviour
Since `tryLogin` is not called, this doesn't get updated. 

10.0.9, but likely effects all versions to date. 

Tricky thing here is the values are passed into the login form using JS in the login screen - not always possible. Worst case, this has to be turned into its own request, which seems quite expensive.... In theory they needs to be set 'on login' so that the returned page/view/app uses the correct timezone, doing an ajax request would be too late for the first page response. So can we even do this? With apache backends we don\t get the opportunity to pass browser data into the login controller at all...

tomneedham:
## Description
No automagic timezone set on login, which doesnt work when using SSO, in favour of selector in personal settings. Later, we could add the auto detection to flag up that you are in a different timezone, and maybe you want to change your settings to your current timezone.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/32165

## Motivation and Context
Automagic timezone changing when you travel affects everything you see in ownCloud, including activity emails. If you don't 'login' then this wont happen, so it is unreliable. AND this only happens when using the login form, SSO does not use this.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

Nothing set, default when you login:
screen shot 2018-08-17 at 13 16 15

When you have a timezone set, or preset by your admin (oc_preferences table, app=core, configkey=timezone)
screen shot 2018-08-17 at 13 15 48

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] Not sure if timezone lib is used elsewhere, remove from `package.json` ? 
- [ ] Unit tests on `SetTimezoneController`

### DIFF
--- a/core/templates/login.php
+++ b/core/templates/login.php
@@ -1,8 +1,6 @@
 <?php /** @var $l \OCP\IL10N */ ?>
 <?php
-vendor_script('jsTimezoneDetect/jstz');
 script('core', [
-	'visitortimezone',
 	'lostpassword',
 	'login',
 	'browser-update'
@@ -58,7 +56,7 @@ script('core', [
 				<?php p($_['user_autofocus'] ? '' : 'autofocus'); ?>
 				autocomplete="on" autocapitalize="off" autocorrect="off" required>
 			<label for="password" class="infield"><?php p($l->t('Password')); ?></label>
-			<input type="submit" id="submit" class="login primary icon-confirm" title="<?php p($l->t('Log in')); ?>" value="" disabled="disabled"/>
+			<input type="submit" id="submit" class="login primary icon-confirm" title="<?php p($l->t('Log in')); ?>" value=""/>
 		</p>
 
 		<?php if (!empty($_['csrf_error'])) {
@@ -102,8 +100,6 @@ script('core', [
 			<label for="remember_login"><?php p($l->t('Stay logged in')); ?></label>
 		</div>
 		<?php endif; ?>
-		<input type="hidden" name="timezone-offset" id="timezone-offset"/>
-		<input type="hidden" name="timezone" id="timezone"/>
 		<input type="hidden" name="requesttoken" value="<?php p($_['requesttoken']) ?>">
 	</fieldset>
 </form>
--- a/settings/Application.php
+++ b/settings/Application.php
@@ -34,6 +34,7 @@ namespace OC\Settings;
 use OC\Server;
 use OC\AppFramework\Utility\TimeFactory;
 use OC\Settings\Controller\CorsController;
+use OC\Settings\Controller\TimezoneController;
 use OC\Settings\Controller\SettingsPageController;
 use OC\Settings\Controller\AppSettingsController;
 use OC\Settings\Controller\AuthSettingsController;
@@ -150,6 +151,15 @@ class Application extends App {
 				$c->query('Logger'),
 				$c->query('URLGenerator'),
 				$c->query('Config')
+			);
+		});
+		$container->registerService('TimezoneController', function (IContainer $c) {
+			return new TimezoneController(
+				$c->query('AppName'),
+				$c->query('Request'),
+				$c->query('Config'),
+				$c->query('L10N'),
+				$c->query('UserSession')
 			);
 		});
 
--- a/settings/Panels/Personal/Profile.php
+++ b/settings/Panels/Personal/Profile.php
@@ -94,6 +94,17 @@ class Profile implements ISettings {
 		$selector->assign('commonlanguages', $commonLanguages);
 		$selector->assign('languages', $languages);
 
+		$timezoneSelector = new Template('settings', 'timezone');
+		$timezoneSelector->assign(
+			'activeTimezone',
+			$this->config->getUserValue(
+				$this->userSession->getUser()->getUID(),
+				'core',
+				'timezone',
+				null));
+		$timezoneSelector->assign('timezones', \DateTimeZone::listIdentifiers());
+
+
 		$tmpl = new Template('settings', 'panels/personal/profile');
 		$tmpl->assign('email', $this->userSession->getUser()->getEMailAddress());
 		$tmpl->assign('displayName', $this->userSession->getUser()->getDisplayName());
@@ -101,6 +112,7 @@ class Profile implements ISettings {
 		$tmpl->assign('avatarChangeSupported', $this->userSession->getUser()->canChangeAvatar());
 		$tmpl->assign('displayNameChangeSupported', $this->userSession->getUser()->canChangeDisplayName());
 		$tmpl->assign('passwordChangeSupported', $this->userSession->getUser()->canChangePassword());
+		$tmpl->assign('timezoneSelector', $timezoneSelector->fetchPage());
 		$groups = $this->groupManager->getUserGroupIds($this->userSession->getUser());
 		\sort($groups);
 		$tmpl->assign('groups', $groups);
--- a/settings/js/panels/profile.js
+++ b/settings/js/panels/profile.js
@@ -230,8 +230,21 @@ $(document).ready(function () {
 				location.reload();
 			}
 			else {
-				$('#passworderror').text(data.data.message);
+				OC.Notification.showTemporary(t('settings', 'Error saving language: {message}', { message: data.data.message }));
 			}
+		});
+		return false;
+	});
+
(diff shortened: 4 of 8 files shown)


tomneedham:
Codestyle and unit tests need to be updated - can do if this is the way we want to go... @PVince81

core/templates/login.php
```diff
@@ -58,7 +56,7 @@
 				<?php p($_['user_autofocus'] ? '' : 'autofocus'); ?>
 				autocomplete="on" autocapitalize="off" autocorrect="off" required>
 			<label for="password" class="infield"><?php p($l->t('Password')); ?></label>
-			<input type="submit" id="submit" class="login primary icon-confirm" title="<?php p($l->t('Log in')); ?>" value="" disabled="disabled"/>
+			<input type="submit" id="submit" class="login primary icon-confirm" title="<?php p($l->t('Log in')); ?>" value=""/>
```
PVince81: why remove the disabled property ?

settings/Application.php
```diff
@@ -152,6 +153,15 @@ public function __construct(array $urlParams=[]) {
 				$c->query('Config')
 			);
 		});
+		$container->registerService('SetTimezoneController', function (IContainer $c) {
```
PVince81: call it just "TimezoneController maybe, who knows what future methods it will have like for example querying all supported timezones ?

settings/Controller/SetTimezoneController.php
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Settings\Controller;
+
+use OC\AppFramework\Http;
+use \OCP\AppFramework\Controller;
```
PVince81: remove leading backslash for consistency ?

settings/Controller/SetTimezoneController.php
```diff
@@ -0,0 +1,81 @@
...
+	/** @var IUserSession  */
+	protected $session;
+
+	public function __construct(
+		$appName,
+		IRequest $request,
+		IConfig $config,
+		IL10N $l,
+		IUserSession $session) {
+		parent::__construct($appName, $request);
```
PVince81: formatting is weird, not really readable.
I think you should move the `) {` on the next line

settings/Controller/SetTimezoneController.php
```diff
@@ -0,0 +1,81 @@
...
+		$this->session = $session;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 */
+	public function set() {
+
+		$available = \DateTimeZone::listIdentifiers();
+		if (!in_array($this->request->getParam('timezoneInput'), $available)) {
```
PVince81: always add `true` in `in_array` to make it strict (doesn't PHPStorm mention this?)

settings/Panels/Personal/Profile.php
```diff
@@ -94,13 +94,25 @@ public function getPanel() {
...
 
+		$timezoneSelector = new Template('settings', 'timezone');
+		$timezoneSelector->assign(
+			'activeTimezone',
+			$this->config->getUserValue(
+				$this->userSession->getUser()->getUID(),
+				'core',
+				'timezone',
+				null));
+		$timezoneSelector->assign('timezones', \DateTimeZone::listIdentifiers());
```
PVince81: this is ok for now.

will not work for Phoenix, so we'll later need to add a get method on the `TimezoneController` so Phoenix knows what values to use to populate. this is only to push my aadvice to rename `SetTimezoneController` to `TimezoneController`

settings/js/panels/profile.js
```diff
@@ -230,12 +230,25 @@ $(document).ready(function () {
 				location.reload();
 			}
 			else {
-				$('#passworderror').text(data.data.message);
+				OC.Notification.showTemporary(t('settings', 'Error saving language: {message}', { message: data.data.message }));
```
PVince81: error saving timezone, not language ?

settings/js/panels/profile.js
```diff
@@ -230,12 +230,25 @@ $(document).ready(function () {
...
 			}
 		});
 		return false;
 	});
 
+	$("#timezoneInput").change(function () {
+		// Serialize the data
+		var post = $("#timezoneInput").serialize();
```
PVince81: ohhhh, never saw this serialize method before

PVince81 (commented):
Looks good overall, see comments.

I wonder if we need a "default" choice or "auto" which would fall back to the old behavior.

PVince81:
@pmaier1 FYI

pmaier1:
:+1: Maybe you want to rename the default timezone simply to "Default" (instead of "Server default" which I find unnecessary technical :) )

tomneedham:
Actually talked to @DeepDiver1975 and he pointed out the solution for timezone addition in shibboleth. So maybe we keep it, but allow manual setting of it?

core/templates/login.php
tomneedham: because it was enabled when the timezone was set - so without removing this it is always disabled

settings/js/panels/profile.js
tomneedham: no, this is language - see whole file. The language save error handler was piggy backing on the password error div for its error message 🙈 which gets worse, when I was testing with LDAP so there was no password change box.

DeepDiver1975:
> Actually talked to @DeepDiver1975 and he pointed out the solution for timezone addition in shibboleth. So maybe we keep it, but allow manual setting of it?

unless I finally understand the need to remove the auto-setting of timezone from the login page I prefer to keep it. We recently fixed it to solve some customer time zone issues.

Having an manual select option is fine.

tomneedham:
Pushed to address comments. @pmaier1 to decide if we want only automagic, automagic+manual option, or only manual? I can update this PR accoridingly

pmaier1:
Well, I agree a lot with https://github.com/owncloud/core/issues/32165#issuecomment-410333084 and https://github.com/owncloud/core/issues/32165#issuecomment-410525504 so I would tend to only go for the manual option. Still I want to respect @DeepDiver1975 in https://github.com/owncloud/core/pull/32371#issuecomment-415400618. @DeepDiver1975 does the reasoning in the first two comments make sense to you or why would you want to keep automagic?

Depending on @DeepDiver1975 answer I'd say we either:
- make the compromise and keep automagic but add the manual option additionally
- just go for the manual option

Question for the compromise way: When users set the manual option, would that imply disabling automagic then?

tomneedham:
> Question for the compromise way: When users set the manual option, would that imply disabling automagic then?

In this case, I would suggest an 'option' in the dropdown is 'auto'. If you select 'uk/london' you automatically unselect auto.

pmaier1:
Nice, I think that's the way to go :ok_hand:

pmaier1:
> generally speaking the server should send any timestamps in a timezone neutral form or with the explicit timezone when

The server can do that still, right? If I'm not mistaken the only question to answer here is: Do I want to set a fixed timezone on the server or do I want it to dynamically change on login. Clients can then still convert it as they wish to.

Not sure if I get all implications right. 
- When using automagic timezone the drawback is that server-generated things like mails use the timezone that was detected on last login. This implies that the timing of these actions follows users when travelling.
- When using a fixed timezone the time in the webUI will not match when I'm in a different timezone as long as I don't change it it in the settings

I'm missing some examples to get a better understanding of the implications.

PVince81:
Hmmm, is the web UI even affected by the timezone change ? I'm not sure.

Maybe the setting could be called "Timezone to use in email communication" or something related to server-only things ?

tomneedham:
> Maybe the setting could be called "Timezone to use in email communication" or something related to server-only things ?

By this point, we are creating a mess for end users......

PR #32372, opened 2018-08-17
Add escape character to like query

DeepDiver1975:
## Description
As soon as the like pattern contains a wildcard _ or % the queries before this change will no match with any data in the database because the missing sql statment ESCAPE is missing which tells the database that \ is used as escape character.

There is a regex which adds ESCAPE to the statements but this one does not match on the queries which are generated with the query builder:

		$statement = \preg_replace('( I?LIKE \?)', '$0 ESCAPE \'\\\'', $statement);
Query builder statments don't use ? but :dcValue

## Related Issue
found during #31651 

## How Has This Been Tested?
- unit tests

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:
- [ ] search for a solution for stable10 - this change does not apply because of different dbal versions
- [ ] search for all like/notLike usage and analyse the impact

### DIFF
--- a/lib/private/DB/AdapterOCI8.php
+++ b/lib/private/DB/AdapterOCI8.php
@@ -39,7 +39,6 @@ class AdapterOCI8 extends Adapter {
 	const UNIX_TIMESTAMP_REPLACEMENT = "(cast(sys_extract_utc(systimestamp) as date) - date'1970-01-01') * 86400";
 
 	public function fixupStatement($statement) {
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
seems like the easiest fix is to convert all createNamedParameter() into createPositionalParameter()

PR #32373, opened 2018-08-19
Allow to set cache scanner only by config, dont expose via class

mrow4a:
The function is used nowhere, and generaly it exposed internal implementation. If one wants to use it, there is a config flag - https://github.com/owncloud/core/blob/master/lib/private/Files/Cache/Scanner.php#L92


mrow4a:
@DeepDiver1975 @PVince81 can we remove this unused function? Cache Active seem to be only possible to be set with the config.

PVince81 (approved):
:+1: master only

I think it used to be used in some process which got changed later on

PR #32375, opened 2018-08-19
Use FileUtils instead of function in Scanner

mrow4a:
As in the topic - addressing technical dept

### DIFF
--- a/lib/private/Files/Utils/FileUtils.php
+++ b/lib/private/Files/Utils/FileUtils.php
@@ -1,0 +1,48 @@
+<?php
+/**
+ * @author Piotr Mrowczynski <piotr@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Files\Utils;
+
+/**
+ * Class FileUtils
+ *
+ * @package OC\Files\Utils
+ */
+class FileUtils {
+
+	/**
+	 * Check if the file is a part file
+	 * (file with part file extension or file within a part files folder)
+	 *
+	 * @param string $path path to a file
+	 * @return boolean
+	 */
+	public static function isPartialFile($path) {
+		if (\pathinfo($path, PATHINFO_EXTENSION) === 'part') {
+			return true;
+		}
+		if (\strpos($path, '.part/') !== false) {
+			return true;
+		}
+
+		return false;
+	}
+}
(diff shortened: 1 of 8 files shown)


lib/public/Files/Cache/IScanner.php
```diff
@@ -69,13 +77,7 @@ public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $loc
 	 * @param string $file
 	 * @return boolean
 	 * @since 9.0.0
+	 * @deprecated use FileUtils
 	 */
 	public static function isPartialFile($file);
```
mrow4a: I dont even think if anybody is using this interface.. why would it? It was circumvented anyways in View and used cache directly

mrow4a: Can we delete it?

lib/public/Files/Cache/IScanner.php
```diff
@@ -69,13 +76,7 @@ public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $loc
 	 * @param string $file
 	 * @return boolean
 	 * @since 9.0.0
+	 * @deprecated use FileUtils
```
DeepDiver1975: FileUtils is OC namespace - it cannot be used to deprecate an OCP method.

keep it here

tests/lib/Files/Utils/FileUtilsTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * Copyright (c) 2012 Robin Appelman <icewind@owncloud.com>
```
DeepDiver1975: wrong copytight - please adjust

tests/lib/Files/Utils/FileUtilsTest.php
```diff
@@ -0,0 +1,41 @@
...
+namespace Test\Files\Utils;
+
+use OC\Files\Utils\FileUtils;
+
+/**
+ * Class ScannerTest
+ *
+ * @package Test\Files\Utils
+ */
+class FileUtilsSTest extends \Test\TestCase {
```
DeepDiver1975: -S

DeepDiver1975: import TestCase

DeepDiver1975 (changes requested):

lib/public/Files/Cache/IScanner.php
DeepDiver1975: you can deprecate it and we can remove the api call in some month/years

mrow4a:
@DeepDiver1975 all tests passing, review addressed

DeepDiver1975:
conflicting ... please resolve

mrow4a:
@DeepDiver1975 rebased

lib/private/Files/Utils/FileUtils.php
```diff
@@ -0,0 +1,48 @@
...
+	 * (file with part file extension or file within a part files folder)
+	 *
+	 * @param string $path path to a file
+	 * @return boolean
+	 */
+	public static function isPartialFile($path) {
+		if (\pathinfo($path, PATHINFO_EXTENSION) === 'part') {
+			return true;
+		}
+		if (\strpos($path, '.part/') !== false) {
```
PVince81: in some cases you replaced this part was missing, I wonder if this will break some undiscovered cases with "part folders"

let's keep this PR master only just in case

PVince81 (approved):
:+1:

phil-davis:
Is https://github.com/owncloud/core/pull/32375#issuecomment-425077409 still correct?
A breaking change, or is backport possible?
@PVince81 @DeepDiver1975

PVince81:
let's play it safe => no backport.

PR #32377, opened 2018-08-19
With objectstore, scan from sourcescanner for shared files

mrow4a:
In https://github.com/owncloud/core/pull/32337 (scan function has been added, but is not needed). Similarly scanFile gets correction (typo).

@DeepDiver1975

### DIFF
--- a/apps/files_sharing/lib/Scanner.php
+++ b/apps/files_sharing/lib/Scanner.php
@@ -26,22 +26,43 @@
 namespace OCA\Files_Sharing;
 
 use OC\Files\ObjectStore\NoopScanner;
+use OC\Files\Cache\Scanner as CacheScanner;
 
 /**
  * Scanner for SharedStorage
  */
-class Scanner extends \OC\Files\Cache\Scanner {
+class Scanner extends CacheScanner {
+
+	/**
+	 * @var null|CacheScanner
+	 */
 	private $sourceScanner;
+
+	/**
+	 * Scan a single file and use source scanner if needed
+	 *
+	 * @inheritdoc
+	 */
+	public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData = null, $lock = true) {
+		$sourceScanner = $this->getSourceScanner();
+		if ($sourceScanner instanceof NoopScanner) {
+			// No Operation Scanner will not update share permission (scanFile won't call getData)
+			list(, $internalPath) = $this->storage->resolvePath($file);
+			return $sourceScanner->scanFile($internalPath, $reuseExisting, $parentId, $cacheData, $lock);
+		} else {
+			return parent::scanFile($file, $reuseExisting, $parentId, $cacheData, $lock);
+		}
+	}
 
 	/**
 	 * Returns metadata from the shared storage, but
 	 * with permissions from the source storage.
 	 *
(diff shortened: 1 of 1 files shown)


apps/files_sharing/lib/Scanner.php
```diff
@@ -26,22 +26,43 @@
...
+	 * Scan a single file and use source scanner if needed
+	 *
+	 * @inheritdoc
+	 */
+	public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData = null, $lock = true) {
+		$sourceScanner = $this->getSourceScanner();
+		if ($sourceScanner instanceof NoopScanner) {
+			// No Operation Scanner will not update share permission (scanFile won't call getData)
+			list(, $internalPath) = $this->storage->resolvePath($file);
+			return $sourceScanner->scanFile($internalPath, $reuseExisting, $parentId, $cacheData, $lock);
```
mrow4a: Due to the fact that we use "source scanner" directly, and not use inheritance, function getData is not being called for objectstorage (NoopScanner). 

I tried to implement the wrapper, but it is nearly impossible due to how Storage interfaces are implemented (https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Storage.php#L51) - storage has function getScanner, which in turn can accept another storage as argument.

apps/files_sharing/lib/Scanner.php
```diff
@@ -64,34 +85,4 @@ private function getSourceScanner() {
 			return null;
 		}
 	}
-
-	/**
-	 * scan a folder and all it's children,  use source scanner if needed
-	 *
-	 * @inheritdoc
-	 */
-	public function scan($path, $recursive = self::SCAN_RECURSIVE, $reuse = -1, $lock = true) {
```
mrow4a: Removing "scan" as might be better if parent method will be executed (It does some locking before running scanFile). CC @DeepDiver1975

mrow4a:
@DeepDiver1975 I just tested it on smashbox and all good. In reviewing, please use `split diff`

DeepDiver1975 (approved):

DeepDiver1975:
yes - baskport is required - but I tend to postpone this once we have smashbox green 

@mrow4a please keep a list of necessary backports - bet in a ticker in here in core to track this all - THX

mrow4a:
@DeepDiver1975 exactly, I will backport all scanner related stuff together if needed.

phil-davis:
@mrow4a ping - is this still waiting for a "major backport of scanner related stuff"?

mrow4a:
@phil-davis we dont have smashbox green yet. There is some missing files_primary_s3 functionality. CC @DeepDiver1975

phil-davis:
I noticed this while backporting "Introduce PHPstan" in PR #35774 
In the current `stable10` `phpstan` told me about a wrong call to `scan`
https://github.com/owncloud/core/pull/35774/commits/c2902e66c15c394dec745da9045239a61bc25a7a#r300867878

Backporting this PR would/should bypass having to massage this in `stable10` - the backport will "fix it".

@DeepDiver1975 @mrow4a

mrow4a:
@phil-davis this is just a performance optimization to avoid unnecessary calls to "locks" with objectstorage. Other storages are not affected

PR #32387, opened 2018-08-20
Splitting config.sample.php into two files for core and app

### LINKED ISSUE #32268: config.sample.php does not mention activity_expire_days
Enhancement Idea:
Shouldn't our config example mention this setting?
config.sample.php in 10.0.9 does not mention activity_expire_days

References:
* https://doc.owncloud.com/server/10.0/admin_manual/configuration/server/activity_configuration.html#configuring-your-owncloud-for-the-activity-app
* https://doc.owncloud.com/server/10.0/admin_manual/configuration/server/config_sample_php_parameters.html#app-config-options

mmattel:
## Description
This PR splits the current ``config.sample.php`` into two files:

``config.sample.php`` core related
  and
``config.apps.sample.php`` apps related

``config.sample.php`` contains all keys for the core system.
It stays more or less the same, only one set of keys has been moved.
(The moved keys are not used when installing the system from scratch so no issue.)

``config.apps.sample.php`` contains now all keys for apps provided by ownCloud.
I went thru all repros and added all keys I have identified and described on best effort.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/32268 (config.sample.php does not mention activity_expire_days)
- Fixing an old discussion 

## Motivation and Context
It clearly seperates keys used by core and those by apps.
It helps app developers to add keys to the right location

## How Has This Been Tested?

- test environment:
no test needed 

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

@settermjd as discussed on talk, you need to update your script generating the documentation version to add the new file.

Note: we are planning the same split for occ commands in documentation.

### DIFF
--- a/config/config.apps.sample.php
+++ b/config/config.apps.sample.php
@@ -1,0 +1,99 @@
+<?php
+
+/**
+ * This configuration file is only provided to document the different
+ * configuration options and their usage for apps maintained by ownCloud.
+ *
+ * DO NOT COMPLETELY BASE YOUR CONFIGURATION FILE ON THESE SAMPLES. THIS MAY BREAK
+ * YOUR INSTANCE. Instead, manually copy configuration switches that you
+ * consider important for your instance to your working ``config.php``, and
+ * apply configuration options that are pertinent for your instance.
+ *
+ * All keys are only valid if the corresponding app is installed and enabled.
+ * You MUST copy the keys needed to the active config.php file.
+ * 
+ * This file is used to generate the configuration documentation.
+ * Please consider following requirements of the current parser:
+ *  * all comments need to start with `/**` and end with ` *\/` - each on their
+ *    own line
+ *  * add a `@see CONFIG_INDEX` to copy a previously described config option
+ *    also to this line
+ *  * everything between the ` *\/` and the next `/**` will be treated as the
+ *    config option
+ *  * use RST syntax
+ */
+
+$CONFIG = array(
+
+/**
+ * App: Activity
+ * 
+ * Possible values: ``activity_expire_days`` days
+ */
+ 
+/**
+ * Retention for activities of the activity app
+ */
+
+'activity_expire_days' => 365,
+
+/**
+ * App: LDAP
+ * 
+ * Possible values: ``ldapIgnoreNamingRules`` 'doSet' or false
+ * 
+ * Possible values: ``user_ldap.enable_medial_search`` true or false
+ */
+ 
+/** 
+ */
+
+'ldapIgnoreNamingRules' => false,
+'user_ldap.enable_medial_search' => false,
+
+/**
+ * App: Market
+ * 
+ * Possible values: ``appstoreurl`` URL
+ */
+
+/**
+ * Configuring the download URL for apps
+ */
+
+'appstoreurl' => 'https://marketplace.owncloud.com',
+
+/**
+ * App: Firstrunwizard
+ * 
+ * Possible values: ``customclient_desktop`` URL
+ * 
+ * Possible values: ``customclient_android`` URL
+ * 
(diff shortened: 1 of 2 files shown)


PVince81:
to me a first look as an admin I would guess that I need two config files: config.php and config.apps.php instead of a single one.

does that work in practice ? as far as I remember the detection pattern was "xx.config.php"

mmattel:
We have to decide what we want.
Either there is one sample file fits all, or two files splitting content for core and apps.
The latter (split) was often requested and makes sense, especially because of growing #keys and apps.
It was not intended to have/use a second active config.php file.
I have not found a description in documentation that this can be done... ?
The intention was to have two ``sample`` files.
The active ``config.php`` file is generated on installation, all additions you want to have are added respectively copied there, based on sampe file(s). This is the standard way of doing.

If there is the possibility of having multiple active ``config.php`` files in parallel behaving like one, we have to document that. But if this is the case, it would be inline with the statement below:

I have already added on top of ``config.apps.sample.php`` a description how to enable keys:
```
 * All keys are only valid if the corresponding app is installed and enabled.
 * You MUST copy the keys needed to the active config.php file.
```

PVince81:
> The latter (split) was often requested

do you have a reference for this ?

From what I see in the doc https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/config_sample_php_parameters.html#config-php-parameters if you have several config files it would read from them but never write there. And those files take precedence over config.php. Now I wonder how this behaves if values are specified in separate files that do get rewritten by OC in config.php...

This also means that if we recommend people to have a separate apps.config.php file, the values there will never be writable by OC itself. This is only acceptable in scenarios where the admin explicitly agreed to have read-only config files in general.

My worry is that with this seemingly small change we're opening a can of worms.

What speaks against just re-sorting the keys inside config.sample.php and have a clear separator between "these are values for core" and "these are values for apps", maybe even sorted by app names ?

mmattel:
Reference for this request: @DeepDiver1975 He very much wanted it, at least I understsood this in that way. There were also other discusions in the past where I at least remember that splitting was a point but I cant remember the issues.

Again, I wanted to split the sample file. Not the config file itself even it is possible (I gave it a try and yes it works as you described). Primarily I would like to have one active config and not multiple. If you look around, the doc you referenced ist the only place this gets mentioned. The doc also explicity mentions to have as example the mail config in a seperate file, even now core writes eg mail changes to config.php...
So a new file is not a new mess, the box of worms as you describe it is already open scince a while !

My very first attempt was to have a section to seperate content but @DeepDiver1975 insisted on splitting. Now where there are two files created, this looks imho much better and gives a better overview. As said, the description on top clearly states that you should copy all relevan keys to config.php, no mentioning about multiple xxx.config.php

Finally, we have the same attempt splitting the occ command reference, see link above. This would be harmonized with this one.

mmattel:
Updated the apps file to run with the converter script for documentation repro.
For more details pls see the result in the refrenced doc PR

mmattel:
@PVince81 any (hopefully positive) statements?

PVince81:
oh, I thought you guys merged this already

> My very first attempt was to have a section to seperate content but @DeepDiver1975 insisted on splitting.

Would be good to have a link to the statement for this, to be sure there was no misunderstanding in writing.

At this point I don't really care

mmattel:
Here is the link as requested: https://github.com/owncloud/core/issues/32268#issuecomment-411082282

@PVince81 could you merge, I will asap create the backport

PVince81:
hmm, I'd interpret the comment more to make it about having some file inside app repositories.
ok, so that could be the next step after this PR.

Fine, let's merge :+1:

Issue #32322, opened 2018-08-14
<file_target> of a auto-renamed file is not correct directly after sharing


individual-it:
### Steps to reproduce
1. create user "uu1" and "uu2"
2. login as both once to make sure a "welcome.txt" exists
3. from "uu2" share "welcome.txt" to "uu1"
4. query the sharing API `curl -u uu2:uu2 http://localhost/owncloud-core/ocs/v1.php/apps/files_sharing/api/v1/shares`

### Expected behaviour
`<file_target>/welcome (2).txt</file_target>`

### Actual behaviour
`<file_target>/welcome.txt</file_target>`

- after querying the sharing API from the side of "uu1" or reloading the page as "uu1" the API respond from "uu2" side gets corrected 
- with user-key enabled encryption the result is correct immediately

I think the expectation in this acceptance tests is wrong: 
 https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiSharing/createShare.feature#L15

### Server configuration
**Operating system**:
Ubuntu 18.04
**Web server:**
Apache 2
**Database:**
Sqlite
**PHP version:**
7.2
**ownCloud version:** (see ownCloud admin page)
10.1.0.1
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

**The content of config/config.php:**

```
{
    "system": {
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "localhost"
        ],
        "dbtype": "sqlite3",
        "version": "10.1.0.1",
        "logtimezone": "UTC",
        "instanceid": "oc2fs70i9byd",
        "loglevel": 2,
        "csrf.disabled": true,
        "datadirectory": "\/home\/artur\/www\/owncloud-core\/data",
        "overwrite.cli.url": "http:\/\/localhost\/owncloud-core",
        "htaccess.RewriteBase": "\/owncloud-core\/",
        "installed": true
    }
}
```

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - dav: 0.3.2
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_sharing: 0.10.1
  - files_texteditor: 2.2.1
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - notifications: 0.3.4
[...]

PR #32395, opened 2018-08-21
skip a test on encryption that is known to fail

individual-it:
skip a test that blocks https://github.com/owncloud/encryption/pull/41 because of https://github.com/owncloud/core/issues/32322


individual-it:
backport in #32396

phil-davis (approved):

PR #32408, opened 2018-08-22
[stable10] don't crash on filescan where folder has symlink (#32407)

DeepDiver1975:
backport #32407


PVince81 (approved):
looks safe :+1:

PR #32391, opened 2018-08-21
[stable10] compare UIDs instead of objects when changing email address

individual-it:
## Description
when strictly comparing objects two instances of the same class are not the same see https://secure.php.net/manual/en/language.oop5.object-comparison.php

## Motivation and Context
acceptance tests that try to change email addresses would fail with: `{"reqId":"1SFByAuzpbf68kTErEUV","level":3,"time":"2018-08-21T05:18:04+00:00","remoteAddr":"172.23.0.4","user":"user1","app":"settings","method":"GET","url":"\/settings\/mailaddress\/change\/MPoCQms4jPeoUAndqGG8f\/user1","message":"The logged in user is different than expected."}`

## How Has This Been Tested?
- change email while being logged in with the correct user
- change email while being logged in with an other user

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32409, opened 2018-08-22
[stable10] compare UIDs instead of objects when changing displayname

individual-it:
same problem as in https://github.com/owncloud/core/pull/32391


PVince81 (approved):
:+1:

DeepDiver1975:
failing unit test is unrelated -> merge

PR #32414, opened 2018-08-23
Backport/31851 async put

DeepDiver1975:
owncloud 10.0.x version of #31851 

## ToDo
- [x] clean commit history
- [x] adjust since tags in shutdownhandler
- [ ] retest with client

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -307,15 +307,46 @@ OC.FileUpload.prototype = {
 		}
 		if (size) {
 			headers['OC-Total-Length'] = size;
-
 		}
+		headers['OC-LazyOps'] = 1;
 
-		return this.uploader.davClient.move(
+		var doneDeferred = $.Deferred();
+
+		this.uploader.davClient.move(
 			'uploads/' + uid + '/' + this.getId() + '/.file',
 			'files/' + uid + '/' + OC.joinPaths(this.getFullPath(), this.getFileName()),
 			true,
 			headers
-		);
+		).then(function (status, response) {
+			if (status === 202) {
+				function poll() {
+					$.ajax(response.xhr.getResponseHeader('oc-jobstatus-location')).then(function(data) {
+						var obj = JSON.parse(data);
+						if (obj.status === 'finished') {
+							doneDeferred.resolve(status, response);
+						}
+						if (obj.status === 'error') {
+							OC.Notification.show(obj.errorMessage);
+							doneDeferred.reject(status, response);
+						}
+						if (obj.status === 'started' || obj.status === 'initial') {
+							// call it again after some short delay
+							setTimeout(poll, 1000);
+						}
+					})
+				}
+
+				// start the polling
+				poll();
+			} else {
+				doneDeferred.resolve(status, response);
+			}
+
+		}).fail( function(status, response) {
+			doneDeferred.reject(status, response);
+		});
+
+		return doneDeferred.promise();
 	},
 
 	_deleteChunkFolder: function() {
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -318,8 +318,10 @@ OC.FileUpload.prototype = {
 			true,
 			headers
 		).then(function (status, response) {
+			// a 202 response means the server is performing the final MOVE in an async manner,
+			// so we need to poll its status
 			if (status === 202) {
-				function poll() {
+				var poll = function() {
 					$.ajax(response.xhr.getResponseHeader('oc-jobstatus-location')).then(function(data) {
 						var obj = JSON.parse(data);
 						if (obj.status === 'finished') {
@@ -333,8 +335,8 @@ OC.FileUpload.prototype = {
 							// call it again after some short delay
 							setTimeout(poll, 1000);
 						}
-					})
-				}
+					});
+				};
 
 				// start the polling
 				poll();
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -82,6 +82,7 @@ use OC\Session\CryptoWrapper;
 use OC\Session\Memory;
 use OC\Settings\Panels\Helper;
 use OC\Settings\SettingsManager;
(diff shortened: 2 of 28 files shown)


lib/private/Server.php
```diff
@@ -1660,4 +1662,13 @@ public function load(array $xmlPath, IUser $user = null) {
 			}
 		}
 	}
+
+	/**
+	 * @return IShutdownManager
+	 * @throws QueryException
+	 * @since 11.0.0
```
patrickjahns: not correct as we are on 10.0.x

lib/public/IServerContainer.php
```diff
@@ -536,4 +537,10 @@ public function getDateTimeFormatter();
 	 * @since 10.0.3
 	 */
 	public function getThemeService();
+
+	/**
+	 * @return IShutdownManager
+	 * @since 11.0.0
```
patrickjahns: not correct as we are on 10.0.x

lib/private/Shutdown/ShutDownManager.php
```diff
@@ -0,0 +1,65 @@
...
+	
+	/**
+	 * Any kind of cleanup should be added with priority LOW.
+	 * Anything which shall be processed before the cleanup routes better
+	 * uses a value smaller then LOW
+	 *
+	 * @param \Closure $callback
+	 * @param int $priority - the lower the number the higher is the priority
+	 * @return void
+	 * @since 11.0.0
```
patrickjahns: also 10.0.x ?

apps/files/js/file-upload.js
```diff
@@ -307,15 +307,46 @@ OC.FileUpload.prototype = {
...
-		return this.uploader.davClient.move(
+		this.uploader.davClient.move(
 			'uploads/' + uid + '/' + this.getId() + '/.file',
 			'files/' + uid + '/' + OC.joinPaths(this.getFullPath(), this.getFileName()),
 			true,
 			headers
-		);
+		).then(function (status, response) {
+			if (status === 202) {
+				function poll() {
```
PVince81: I think I'd prefer to have this method on the object level as attribute instead of deep inside this block

apps/files/js/file-upload.js
```diff
@@ -307,15 +307,46 @@ OC.FileUpload.prototype = {
...
 
-		return this.uploader.davClient.move(
+		this.uploader.davClient.move(
 			'uploads/' + uid + '/' + this.getId() + '/.file',
 			'files/' + uid + '/' + OC.joinPaths(this.getFullPath(), this.getFileName()),
 			true,
 			headers
-		);
+		).then(function (status, response) {
+			if (status === 202) {
```
PVince81: add JS comment to tell what this is about: we're switching to async status mode

PVince81 (commented):
Code looks good in general.

There are a few small things that would be nice to adjust.

And versions in PHPDoc need adjusting

apps/files/js/file-upload.js
```diff
@@ -307,15 +307,46 @@ OC.FileUpload.prototype = {
...
+						if (obj.status === 'finished') {
+							doneDeferred.resolve(status, response);
+						}
+						if (obj.status === 'error') {
+							OC.Notification.show(obj.errorMessage);
+							doneDeferred.reject(status, response);
+						}
+						if (obj.status === 'started' || obj.status === 'initial') {
+							// call it again after some short delay
+							setTimeout(poll, 1000);
```
PVince81: we need to store the timeout and stop polling if the upload is cancelled by the user.

not sure what needs to happen on the server side as the assembly has already started and it's too late to cancel. maybe we disallow cancelling in the web UI then ?

PVince81:
@patrickjahns schedule automated tests for this ? there are some curl examples here https://github.com/owncloud/core/pull/31851

PVince81:
I've adressed all comments but left the JS code as is as I think it's ok. Another approach would be more convoluted.

PVince81:
- [ ] fix tests
- [ ] write acceptance tests

PVince81:
- [ ] whatever is added here will need to flow back to master / the master PR

PVince81:
For the acceptance tests I think we can only test submitting the job with the lazyops header, and seeing that it returns 202 with a job status code.
We can't control how long the job is running so there's a risk of race conditions.
We could use the testing app somehow to freeze the job so we can query its status realiably...

PVince81:
raised https://github.com/owncloud/testing/issues/6 for adding ability to freeze code (like uploads) in testing app

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

PVince81:
@patrickjahns as this is currently not easily testable automatically and there is also pressure for shipping it, let's ship this as "technical preview" and have some basic manual testing. The setting is already opt-in in config.php.

PVince81:
Hmm, seems this is not only for the final MOVE of chunk assembly but can also be used with regular file/folder MOVE operations.

PVince81:
- [ ] BUG: assembly seems not never complete... I've waited a few minutes and job status still says "started"

PVince81:
- [x] TEST: file assembly works when MOVE is issued from curl on the command line, job status says "finished"

PVince81:
It looks like the 202 HTTP status code is returned correctly, but that same HTTP connection is not getting closed and blocks until assembly is finished.

This explains why the web frontend doesn't start its polling of "jobstatus" because we haven't got a full response yet. Not sure if JS / XMLHttpRequest / jquery.fileupload even supports reading partial responses... It's not that bad, it shows "Processing files..." anyway.

PVince81:
Had a chat with @butonic and clarified: the above behavior is ok.

The job status polling will actually start if the server (usually a proxy) would disconnect the connection after the job a started. In this case the JS side would get the 202 header with no content and move to polling.

We agreed to ship this as is, tech preview disabled by default.

PVince81:
ported to master here: https://github.com/owncloud/core/pull/32576

PR #32415, opened 2018-08-23
skip unstable test on LDAP

individual-it:
skip test that is unstable for unknown reason on LDAP https://github.com/owncloud/user_ldap/issues/274


phil-davis (approved):
As discussed and noted in user_ldap issue, can be looked into later.

PR #32425, opened 2018-08-23
Shorten hostname to relative domain name in status.php

### LINKED ISSUE #32424: Shorten hostname to relative domain name in status.php
Currently it is only possible to print out or hide the full qualified domain name (_FQDN_) / hostname in the `status.php` using the configuration option `show_server_hostname`. Due to security reasons we just want to get the relative domain name instead of the FQDN.

MariusTimmer:
## Description
This pull request introduces the new configuration option `use_relative_domain_name` which defaults to false. If set to true the hostname will be shorten to the relative domain name.

## Related Issue
- Fixes #32424 

## Motivation and Context
It provides the new option `use_relative_domain_name` to shorten the FQDN to the relative domain name to hide our backend servers from the users.

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


patrickjahns:
@MariusTimmer 

Can you please provide Context / Motivation behind this change.
If agreed upon - please provide also test cases covering this change - so we can ensure it is working in the future.

MariusTimmer:
@patrickjahns 
Sorry that I missed our motivation and a the context in my opening message. According to the colleague who requested this feature it is a security issue.

## Motivation
Since we are using owncloud in a cluster we want to allow our customers to call `status.php` so they can find out which node of the cluster may have caused problems.

When calling `status.php` without our modification, the users will see e.g. `machine01.uni-muenster.de` as hostname. But we do not want to reveal the domain, because in some cases you could reach `machine01.uni-muenster.de` from the internet, when the firewall isn't correctly configured - Okay; In that case you'll have other problems too. (_We're having security audits from people in our security team, so I'll hope, they'd find these kind of misconfiguration!_) But users can be motivated trying to 'hack' our systems (_beginning with `ping`, `tracert` and `nmap`_), when the name is a worldwide valid system name. This can produce lots of load on firewall systems (_DOS_). It's kinda security by obscurity.
We are lucky, because our web-servers don't use official IP addresses. So, a defective firewall config won't make a problem in our installation. But on the other hand, we can't guarantee, that we will have this "security-feature" of private IP-addresses, when IPv6 will come to sciebo(_our cluster_) /owncloud.

## Test case
In commit be0c1ce8eca40b4b38e7a4aa222c63baef2f1894 I added a new assertion in `UnitTest::testGetStatusInfo()`. In case the original hostname contains a dot it will be used as separator and the test will make sure the full qualified domain name is not equals the relative domain name.

MariusTimmer:
Tests in our system showed that the Boolean value (_third argument for `\OCP\Util::getStatusInfo()`_) within the `status.php` file was flipped. That caused showing the FQDN when setting the option `use_relative_domain_name` to true and vice versa. Commit cd2c01a7c4063aa6d8a75a39e0357dcaa14ff0a0 fixes this trivial problem.

config/config.sample.php
```diff
@@ -99,6 +99,12 @@
  */
 'show_server_hostname' => false,
 
+/**
+ * Optionally, use the short hostname in the server.php.
```
PVince81: status.php not server.php

MariusTimmer: Oh, that's correct. Thank you for that hint. :+1:
I fixed that and rebased my master.

PVince81 (approved):
:+1:

PVince81:
damn, should have asked for squashing the commits.

I'll backport all into a single one then

PVince81:
stable10: https://github.com/owncloud/core/pull/34469

PR #32428, opened 2018-08-23
API acceptance tests for slash in group name

phil-davis:
This is to see which of these tests pass and which fail. So we know what the current state of play is for groups with slash in their name via the provisioning API.


phil-davis:
Yep, all these added acceptance tests fail. We should at least add them and skip them so they are there ready to be solved.

PR #32403, opened 2018-08-21
Fix not allowed to share message

tomneedham:
## Description
When you are in a group that is not allowed to share, the message is wrong. This new message covers both no reshare permission and no share permission in one message.

## Issue
#31841 

## Motivation and Context
Strange message

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32429, opened 2018-08-23
[stable10] Fix not allowed to share message

phil-davis:
Backport #32403


PVince81 (approved):
:+1:

PVince81:
note: we'll need to think about backporting translations before RC

DeepDiver1975:
> note: we'll need to think about backporting translations before RC

should be part of the release check list ..

PR #32458, opened 2018-08-27
su-exec www-data is not needed for acceptance tests

phil-davis:
## Description
1) remove the ``su-exec www-data`` from where it is used to run acceptance tests in drone.
2) the acceptance test ``.run.sh`` script tries to be nice and creates a ``local_storage`` folder inside the server. We can't do that successfully when running as another user (not ``www-data``). So do it in another drone step ``setup-api-acceptance-tests`` that happens before the permissions etc get reset to ``www-data`` for the server.

## Issue
#32459 

## Motivation and Context
The acceptance test script should be able to run "outside" and not as the owner of the server files (e.g. not as ``www-data``)

## How Has This Been Tested?
CI knows

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -211,6 +211,16 @@ pipeline:
       matrix:
         CALDAV_CARDDAV_JOB: true
 
+  setup-api-acceptance-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      # pre-create local_storage so fix-permissions will set its owner appropriately
+      - mkdir -p /drone/src/tests/acceptance/work/local_storage
+    when:
+      matrix:
+        TEST_SUITE: api
+
   install-notifications-app:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
--- a/tests/acceptance/features/bootstrap/AppManagementContext.php
+++ b/tests/acceptance/features/bootstrap/AppManagementContext.php
@@ -20,6 +20,7 @@
  */
 
 use Behat\Behat\Context\Context;
+use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 use TestHelpers\SetupHelper;
 
 require __DIR__ . '/../../../../lib/composer/autoload.php';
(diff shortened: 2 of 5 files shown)


phil-davis:
```
--- Failed scenarios:

    /drone/src/tests/acceptance/features/apiMain/appmanagement.feature:7
/drone/src/tests/acceptance/features/apiMain/appmanagement.feature:13
```
- [ ] AppManagementContext is merrily trying to do direct ``mkdir`` and ``file_put_contents`` to create a fake app in the ``apps2`` folder. It also gets and sets system by directly calling ``getSystemValue()`` and ``setSystemValue()``. All of that needs to happen via the testing app (or at least some other way that does not depend on the acceptance test code writing directly into the backend of the server filesystem or database)

phil-davis:
Needs testing app https://github.com/owncloud/testing/pull/11 merged, then rerun CI here and see if there is anything remaining to do.

.drone.yml
```diff
@@ -211,6 +211,16 @@ pipeline:
       matrix:
         CALDAV_CARDDAV_JOB: true
 
+  setup-api-acceptance-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      # pre-create local_storage so fix-permissions will set its owner appropriately
+      - mkdir -p /drone/src/tests/acceptance/work/local_storage
```
individual-it: isn't `run.sh` doing that?

tests/acceptance/features/bootstrap/AppManagementContext.php
```diff
@@ -20,71 +20,165 @@
...
 	 */
 	public function setAppDirectories($dir1, $dir2) {
-		$fullpath1 = \OC::$SERVERROOT . "/$dir1";
-		$fullpath2 = \OC::$SERVERROOT . "/$dir2";
-		\OC::$server->getConfig()->setSystemValue(
-			'apps_paths',
+		$fullpath1 = $this->getServerRoot() . "/$dir1";
+		$fullpath2 = $this->getServerRoot() . "/$dir2";
+
+		$this->mkDirOnServer($dir1);
```
individual-it: I guess it does not matter if the path already exists

.drone.yml
phil-davis: ``run.sh`` tries, but does not succeed on drone because by that  time ``tests/acceptance`` is owned by ``www-data``

creating it here means that the ``mkdir`` in ```run.sh`` becomes a no-op.

tests/acceptance/features/bootstrap/AppManagementContext.php
phil-davis: Yes, it behaves similar to command-line ``mkdir`` - it is a no-op when the dir already exists.

.drone.yml
phil-davis: I left the ``mkdir`` code in ``run.sh`` because when someone is running tests locally we want the dir to be created if it is not already there.

The only catch-22 will be if you are running locally with a proper Apache server as ``www-data`` and you stop putting ``sudo ww-data`` on the front. Then all the tests should still run for you, but you will need to do like drone, pre-create ``work/local_storage``

individual-it (approved):

phil-davis:
Backport ``stable10`` #32586

PR #32461, opened 2018-08-27
[stable10] Add log entry for each migration that is run

PVince81:
## Description
Add log entries when running DB migrations.
Should hopefully help more update related issues where migrations did not run...

## Related Issue
- For https://github.com/owncloud/enterprise/issues/2811

## How Has This Been Tested?

1. delete some core and app migrations in oc_migrations
2. increase version.php
3. increase app version
4. run `occ upgrade`
5. check owncloud.log and see the migrations mentioned after they ran

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

@cdamken

PR #32463, opened 2018-08-27
[master] Add log entry for each migration that is run

PVince81:
Port of https://github.com/owncloud/core/pull/32461 to master


felixboehm (approved):

PVince81:
```
lib/private/DB/MigrationService.php:412 PhanUndeclaredClassMethod Call to method debug from undeclared class \OC\DB\ILogger
```

I'll fix that...

PVince81:
added missing attribute declaration

when testing it worked because PHP in general doesn't care if it's not declared :-S

PR #32475, opened 2018-08-28
[stable10] Another missing ILogger declaration extra

PVince81:
Another extra from https://github.com/owncloud/core/pull/32463 ... My :coffee: is failing me...


phil-davis (approved):
This is an example where doing bugfix PRs in master (and possibly in parallel in stable10) is good. Because then in master you find out about some extra code checks straight away and can do it the same in stable10.

PVince81:
exactly... you were right over and over again

phil-davis:
Sorry, I don't mean to be pain. I commented just because is it an example related to discussions about semver and workflows for bugfixing and backport-or-forwardport where we are thinking about the merits of different approaches.

PR #32485, opened 2018-08-28
Fix sinon js tests

PVince81:
Fixes JS unit tests to be able to work with future Sinon JS releases.

This is related to https://github.com/owncloud/core/pull/32316 but is missing the davclient.js update.

As the tests work with the current version let's just update them here to avoid future conflicts.


phil-davis (approved):
It looks harmless and is only in test code.
So if CI passes it must be good :)

phil-davis:
@PVince81 backport?

PVince81:
stable10: https://github.com/owncloud/core/pull/32488

PR #32487, opened 2018-08-28
Jasmine 3 update

PVince81:
Update to Jasmine 3.

It randomizes test run order, so many side effects will become apparent.

Goal is to fix them all until several local runs in a row do not fail any more...

### DIFF
--- a/build/package.json
+++ b/build/package.json
@@ -38,7 +38,7 @@
   },
   "devDependencies": {
     "handlebars": "^4.0.11",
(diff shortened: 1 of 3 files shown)


PVince81:
I reran this locally several times and saw so many different kinds of issues...

I think a lot is due to the excessive usage of globals in OC in general, and some tests or code might be altering the global's states.

PVince81:
I give up. this will likely be a lot of work.
better invest in Phoenix and replace the old frontend...

PR #31807, opened 2018-06-16
Add phan as static code analyser

patrickjahns:
## Description
https://github.com/phan/phan is a static code analysis tool that can help us improve our code basis and help find bugs.

Right now the `.phan/config` is very lenient - we should first get it to be ✅ for this lenient definition, merge / introduce it to our code basis and then gradually increase the ruleset to further improve our code base.

note: phan is currently introduced as phar - this allows us for backporting it to stable10 - without having dependencie issues, as phan officialy needs to run on php7+ ( however it can analyse code in 7 to be backward compatible to php5.6 )

## Motivation and Context
Improve code quality and ownCloud as project

## pending fixes

```
lib/private/Encryption/Util.php:123 PhanUndeclaredClassConstant Reference to constant ID from undeclared class \OCA\Encryption\Crypto\Encryption
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] QA / code related

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #32492, opened 2018-08-28
Stable10 experimental phan

PVince81:
Backport of https://github.com/owncloud/core/pull/31807 to stable10

Some differences based on master:
- php-cs-fixer is now also in the Drone file. (I ran it locally and there are no style errors!)
- removed everything related to zsync / delta sync

There were no other conflicts.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -114,6 +114,24 @@ pipeline:
       matrix:
         TEST_SUITE: lint
 
+  php-cs-fixer:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    commands:
+      - make test-php-style
+    when:
+      matrix:
+        TEST_SUITE: php-cs-fixer
+
+  php-phan:
+    image: owncloudci/php:7.1
+    pull: true
+    commands:
+      - make test-php-phan
+    when:
+      matrix:
+        TEST_SUITE: phan
+
   install-server:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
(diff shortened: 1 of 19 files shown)


patrickjahns:
I'll take care and fix the pipeline. 

Phan is also available as 1.0.0 release - so I want to move forward with the latest release and keep master / stable10 in sync

.drone.yml
```diff
@@ -114,6 +114,24 @@ pipeline:
       matrix:
         TEST_SUITE: lint
 
+  php-cs-fixer:
```
patrickjahns: duplicated - we have a different step already in place

patrickjahns:
```
lib/private/Files/View.php:380 PhanUndeclaredClassMethod Call to method __construct from undeclared class \OC\Files\Storage\Storage\local
lib/private/Files/View.php:381 PhanUndeclaredClassMethod Call to method readdir from undeclared class \OC\Files\Storage\Storage\local
settings/Application.php:99 PhanUndeclaredClassMethod Call to method __construct from undeclared class \OC\Settings\View
Makefile:215: recipe for target 'test-php-phan' failed
```

.drone.yml
phil-davis: The one added here (by cherry-pick, which will not understand/see the other one that is somehow quite a few lines away below) is in the matching place that it is in master. So for future clean backports... it would be best to delete the one further down (which has identical code)

PVince81:
I've adjusted .drone.yml and fixed the phan warnings.

phil-davis (approved):

PR #31847, opened 2018-06-20
return all failed recipients in sendLinkShareMailFromBody

phil-davis:
## Description
Construct a comma-separated string of all the recipient addresses in the failed email - To, Bcc and Cc.
Return that complete list to the caller of ``sendLinkShareMailFromBody()``
This is specially useful, because usually these emails have no ``To`` recipient.

## Related Issue
#31845 

## Motivation and Context
Mention all email addresses when an attempt to send an email about a public link fails.

## How Has This Been Tested?
1. do not start any mail server (or do something else to make mail sending fail in general)
2. set the system log level to debug (0)
3. In admin, settings, sharing, enable "Allow users to send mail notification for shared files"
4. Login as an ordinary user
5. Share a file or folder as a public link
6. Enter some email address/es in "Send link via email"
7. Click "Share"

Error dialog is displayed that includes the failed email addresses:
  An error occurred while sending email
  Couldn't send mail to following recipient(s):
    mail1@example.com,mail2@example.com

``owncloud.log`` contains an entry like:
```
Can't send mail with public link to mail1@example.com,mail2@example.com: Connection could not be established with host 127.0.0.1 [Connection refused #111]
```

## Screenshots (if appropriate):
screenshot-2018-6-20 files - owncloud 1

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Issue #31845, opened 2018-06-20
webUI error dialog when can't send public link share mail has no recipients


phil-davis:
### Steps to reproduce
0. do not start any mail server (or do something else to make mail sending fail in general)
1. In admin, settings, sharing, enable "Allow users to send mail notification for shared files"
2. Login as an ordinary user
3. Share a file or folder as a public link
4. Enter some email address/es in "Send link via email"
5. Click "Share"

### Expected behaviour
Error dialog is displayed saying:
  An error occurred while sending email
  Couldn't send mail to following recipient(s):
    mail1@example.com
    mail2@example.com

### Actual behaviour
Error dialog is displayed saying:
  An error occurred while sending email
  Couldn't send mail to following recipient(s):

but no recipients are mentioned.

### Server configuration
10.0.9RC1 or current stable10

phil-davis:
Noticed while testing #31227 but is unrelated to that fix.

Seems to be because actually all the recipients are sent in a single email as BCC. That email has no "To" recipients. If that single email attempt fails, the code in ``lib/private/Share/MailNotifications.php`` ``sendLinkShareMailFromBody`` returns the ``To`` recipient if the email send fails. So the BCC (and CC) recipients are not in the array of failed recipients.

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributors most likely able to help are @felixheidecke, @pmaier1, and @PVince81.

Possibly related issues are https://github.com/owncloud/core/issues/7687 (Do not close Share Dialog on error), https://github.com/owncloud/core/issues/7783 (File sharing dialog does not notify user if recipient has no email), https://github.com/owncloud/core/issues/28840 (-Error-), https://github.com/owncloud/core/issues/26902 (error), and https://github.com/owncloud/core/issues/10331 (Selecting "No people found" in share with dialog gives error).

phil-davis:
Also the error logged in ``owncloud.log`` does not mention any email address(es):
```
Can't send mail with public link to : Connection could not be established with host 127.0.0.1 [Connection refused #111]
```

phil-davis:
PR #31847 has an attempt at a fix.

phil-davis:
Backport PR #31935 is waiting for merge after 10.0.9 code freeze.

PR #32506, opened 2018-08-29
Share notification recipients should be added to list

### LINKED ISSUE #32008: Share notifications blocked by spamassassin
### Steps to reproduce
1. Create a link share (Sharing / Public Links / Create public link)
2. Send link via email

### Expected behaviour
Share notification should be delivered.

### Actual behaviour
Share notification gets blocked by spamassassin due to missing "To:" header.
`MISSING_HEADERS	Missing To: header`

### Server configuration
**Operating system**: Ubuntu 16.04.2 LTS

**Web server:** Apache/2.4.18

**Database:** MySQL

**PHP version:** PHP 7.0.30-0ubuntu0.16.04.1

**ownCloud version:** 10.0.8.5

**Updated from an older ownCloud or fresh install:** fresh install

```
{"reqId":"sqy5dyXHHLdalpRxKLJn","level":0,"time":"2018-07-07T18:59:57+00:00","remoteAddr":"193.XX.XX.XX","user":"admin","app":"core","method":"POST","url":"\/oc10\/index.php\/core\/ajax\/share.php","message":"Sent mail to \"Array\n(\n)\n\" with subject \"admin shared \u00bbPhotos\u00ab with you\""}
```
(I did **not** remove anything from "message", the array is actually empty)

sharidas:
Share notification recipients should be added to list.
Adding them to bcc was causing spam filters to block
the emails. This change would solve the issue and
users get notified.

## Description

Add recipients to share notification by email to the 'to' field/list. So that spam filters would not block the emails.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/32008

## Motivation and Context

Add recipients to share notification by email to the 'to' field/list. So that spam filters would not block the emails.

## How Has This Been Tested?

- Configure ownCloud to enable share notification by email.
- Create a public link and add multiple recipients
- The emails received by recipients are in the to field.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/ajax/share.php
+++ b/core/ajax/share.php
@@ -179,48 +179,57 @@ if (isset($_POST['action'], $_POST['itemType'], $_POST['itemSource'])) {
 			]);
 
 			$options = [];
-			$options['bcc'] = $filter->getToAddress();
-			if (isset($_POST['bccSelf']) && $_POST['bccSelf'] === 'true') {
-				$options['bcc'] .= ',' . \OC::$server->getUserSession()->getUser()->getEMailAddress();
+			$results = [];
+			$toEmails = \explode(',', $filter->getToAddress());
+
+			//Lets get the senders email address and add to the 'to'
+			if (isset($_POST['toSelf']) && $_POST['toSelf'] === 'true') {
+				$toEmails [] = \OC::$server->getUserSession()->getUser()->getEMailAddress();
 			}
+			//Send separate email
+			foreach ($toEmails as $toEmail) {
+				$options['to'] = $toEmail;
 
-			$l10n = \OC::$server->getL10N('lib');
+				$l10n = \OC::$server->getL10N('lib');
 
-			$sendingUser = \OC::$server->getUserSession()->getUser();
-			$mailNotification = new \OC\Share\MailNotifications(
-				$sendingUser,
-				$l10n,
-				\OC::$server->getMailer(),
-				\OC::$server->getConfig(),
-				\OC::$server->getLogger(),
-				$defaults,
-				\OC::$server->getURLGenerator(),
-				\OC::$server->getEventDispatcher()
-			);
+				$sendingUser = \OC::$server->getUserSession()->getUser();
+				$mailNotification = new \OC\Share\MailNotifications(
+					$sendingUser,
+					$l10n,
+					\OC::$server->getMailer(),
+					\OC::$server->getConfig(),
+					\OC::$server->getLogger(),
+					$defaults,
+					\OC::$server->getURLGenerator(),
+					\OC::$server->getEventDispatcher()
+				);
 
-			$expiration = null;
-			if ($filter->getExpirationDate() !== '') {
-				try {
-					$date = new DateTime($filter->getExpirationDate());
-					$expiration = $date->getTimestamp();
-				} catch (Exception $e) {
-					\OCP\Util::writeLog('sharing', "Couldn't read date: " . $e->getMessage(), \OCP\Util::ERROR);
+				$expiration = null;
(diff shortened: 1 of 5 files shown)


PVince81:
Now I think I remember why we didn't want to use a loop originally: if you set bcc to every single email, then the sender will get X similar emails where X is the number of recipients.

Here I suggest to simply add the sender in the array of "To" recipients so the sender will receive the message as is. @pmaier1 is that acceptable ?

PVince81:
@sharidas let's go with the approach I proposed. I don't think there's a better approach

pmaier1:
> Here I suggest to simply add the sender in the array of "To" recipients so the sender will receive the message as is. @pmaier1 is that acceptable ?

I think so, yes.

sharidas:
I have updated the PR by adding senders email address to the 'to' list. Let me know if this looks ok @PVince81

sharidas:
Backport PR https://github.com/owncloud/core/pull/32542

core/ajax/share.php
```diff
@@ -179,48 +179,57 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
 			]);
 
 			$options = [];
-			$options['bcc'] = $filter->getToAddress();
+			$results = [];
+			$toEmails = \explode(',', $filter->getToAddress());
+			//Lets get the senders email address and add to the 'to'
+			$toEmails []  = \OC::$server->getConfig()->getSystemValue('mail_smtpname', '');
```
PVince81: hmmm I don't understand what this is ?

core/ajax/share.php
```diff
@@ -179,48 +179,57 @@ function usersInGroup($gid, $search = '', $limit = -1, $offset = 0) {
...
 			if (isset($_POST['bccSelf']) && $_POST['bccSelf'] === 'true') {
-				$options['bcc'] .= ',' . \OC::$server->getUserSession()->getUser()->getEMailAddress();
+				$toEmails [] = \OC::$server->getUserSession()->getUser()->getEMailAddress();
```
PVince81: this is the sender I meant all along, not the technical one but the semantic one: whoever created the public link and is clicking the button for getting the email out

core/ajax/share.php
sharidas: Removed this change.

core/js/sharedialogmailview.js
```diff
@@ -145,7 +145,7 @@
 			var $formSendIndicator = this.$el.find('.emailPrivateLinkForm--sending-indicator');
 			var  mail = {
 				 to      : this._addresses.join(','),
-				 bccSelf : this.$el.find('.emailPrivateLinkForm--emailBccSelf').is(':checked'),
+				 toSelf : this.$el.find('.emailPrivateLinkForm--emailBccSelf').is(':checked'),
```
PVince81: you missed `emailBccSelf`. better grep for it in case it's used in the CSS files

PVince81 (commented):
looks good, just one minor thing

core/js/sharedialogmailview.js
sharidas: Updated the PR. I grepped for `emailBccSelf` in the core and did not found in the CSS files. But after renaming it, the functionality is not broken. So it is safe with the new change.

PVince81 (approved):
:+1:

please adjust the backport PR

phil-davis:
This caused a regression in issue #31845 PR #31847
See PR #32632 for proposed fix.
It is text reporting to the UI only, so nothing is broken in terms of actually sending email...

PR #32513, opened 2018-08-30
log wrapped exceptions

butonic:
## Description
Exceptions in logs are the only thing we see in bug reports. Wouldn't it be great to also see the previous exceptions?

## Motivation and Context

don't throw away useful information

## How Has This Been Tested?

trying to debug decryption with @tomneedham, trying to debug wnd mount issues when oauth is enabled, manually throwing two exceptions:
```diff
diff --git a/index.php b/index.php
index 7e1bbc727f..37724d71de 100644
--- a/index.php
+++ b/index.php
@@ -51,6 +51,8 @@ if (\strtoupper(\substr(PHP_OS, 0, 3)) === 'WIN') {
 
 try {
        require_once __DIR__ . '/lib/base.php';
+       $e1 = new Exception('first one');
+       throw new Exception('second one', 0, $e1);
        OC::handleRequest();
 } catch (\OC\ServiceUnavailableException $ex) {
        \OC::$server->getLogger()->logException($ex, ['app' => 'index']);
```
open indec.php in the web ui ... marvel at the blue error
```console
tail data/owncloud.log | grep log-exception-test | jq
```
```json
{
  "reqId": "W4eyOX8AAQEAAG2OB0EAAAAA",
  "level": 3,
  "time": "2018-08-30T09:00:41+00:00",
  "remoteAddr": "127.0.0.1",
  "user": "DFDD908A-1DDE-4E05-A6D9-7EEE67D092F5",
  "app": "index",
  "method": "GET",
  "url": "/index.php/",
  "message": "Exception: {\"Exception\":\"Exception\",\"Message\":\"log-exception-test weird problem\",\"Code\":0,\"Trace\":\"#0 {main}\",\"File\":\"\\/var\\/www\\/html\\/core\\/index.php\",\"Line\":55}"
}
{
  "reqId": "W4eyOX8AAQEAAG2OB0EAAAAA",
  "level": 3,
  "time": "2018-08-30T09:00:41+00:00",
  "remoteAddr": "127.0.0.1",
  "user": "DFDD908A-1DDE-4E05-A6D9-7EEE67D092F5",
  "app": "index",
  "method": "GET",
  "url": "/index.php/",
  "message": "Caused by: {\"Exception\":\"Exception\",\"Message\":\"log-exception-test root cause\",\"Code\":0,\"Trace\":\"#0 {main}\",\"File\":\"\\/var\\/www\\/html\\/core\\/index.php\",\"Line\":54}"
}
```
_monolog?_ :speak_no_evil: 

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -432,5 +432,10 @@ class Log implements ILogger {
 		$msg = isset($context['message']) ? $context['message'] : 'Exception';
 		$msg .= ': ' . \json_encode($exception);
 		$this->log($level, $msg, $context);
+		// also log previous exception
+		if ($context['exception']->getPrevious()) {
+			$context['message'] = 'Caused by';
+			$this->logException($context['exception']->getPrevious(), $context);
+		}
 	}
 }


patrickjahns:
I am a bit skeptical here and would like to try this out when the sentry app is combined with this mechanism.
I feel that one incident now can create up to X amount of events and basically spam the incident management service. 

Can you elaborate, why fetching the previous exception adds more value? Shouldn't that exception be in the trace anyway?

PVince81:
closing as inactive and there were concerns

butonic:
Maybe I was unclear in explaining what happens. The problem is that we have several places in the code where we wrap an Exception in another exception, eg: https://github.com/owncloud/core/blob/13fe61acd879cde3fc125ec9a97d5618aeb5cc99/lib/private/Files/Storage/FailedStorage.php#L55

In this case we only log the StorageNotAvailableException from line 55, but the 'previous' or wrapped exception, which might contain the actual reason in the message, does not get logged. Which makes debugging extremely painful. And the traces associated with the two exceptions might be substantially different.

@PVince81 reopening, this significantly reduces debugging time for encryption and storage related cases with a very simple change. Please take over.

tomneedham (approved):

tomneedham:
Have also come across areas this would help. Usually the top level exception masks the lower level. Having multiple events in sentry is ok. Each exception would be tracked as a 'similar' (or whatever sentry call it) event, and can be dealt with independantly. Unlying storage issue could be a concern and may need to be addressed whilst top level could be 'ignored.

PVince81:
from what I understand this anyway would only log the next level and doesn't use iteration, so no risk of infinite loop. Fine by me :+1:

PVince81:
@butonic can you backport ?

(if not I'll do so in the coming days)

phil-davis:
Backport `stable10` #34475

PR #32522, opened 2018-08-30
Acceptance test code style

phil-davis:
Just a few things reported by the ``phpcs`` that we have been trying to keep to.

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -327,7 +327,9 @@ trait BasicStructure {
 	}
 	
 	/**
-	 * @param string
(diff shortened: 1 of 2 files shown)


patrickjahns:
Shouldn't we move towards using ownCloud coding standard in tests as well?

phil-davis:
Backport ``stable10`` #32523

phil-davis:
Yes, we do use the ownCloud standards. We currently also use the ``phpcs.xml`` that was added to core a while ago. That checks various additional things, particularly in PHPdoc. One day we should look at which of those we want project-wide and how to best enforce them (which ``phpcs`` or some other tool that knows how to parse them)

individual-it (approved):

PR #32531, opened 2018-08-31
Provision to conditionally hash public link password

### LINKED ISSUE #32383: Conditionally hash share link passwords
Passwords are always hashed even though they have been hashed before. This makes impossible to move the hashed password around in order to reuse the same password because the password hash will be hashed again.

The proposal is to avoid rehashing the password.

For the specific case of password-protected share links, the list of changes is as follows (there might be more changes involved):
* Add a `isHashed` method in the `IHasher` and in the implementation. The method will return true if the message is already hashed or false otherwise. For the core implementation, we can use http://php.net/manual/es/function.password-get-info.php for this purpose.
* The ShareManager can conditionally hash the password based on the result of the above function before creating / updating the password-protected share.

Note that we assume that the system has the same password salt. In case the password salt is different, the old password won't work as its hashed value will likely be different.

sharidas:
This change provides the provision to conditionally
hash public link password.

## Description

A provision to conditionally hash the public link password. Lets say if the password is already hashed, we don't have to hash the password again. This would make the link inaccessible.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/32383

## Motivation and Context

A provision to conditionally hash the public link password. Lets say if the password is already hashed, we don't have to hash the password again. This would make the link inaccessible.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -616,7 +616,7 @@ class Manager implements IManager {
 			}
 
 			// If a password is set. Hash it!
-			if ($share->getPassword() !== null) {
+			if (($share->getPassword() !== null) && ($share->getShouldHashPassword() === true)) {
 				$share->setPassword($this->hasher->hash($share->getPassword()));
 			}
 		}
@@ -843,7 +843,9 @@ class Manager implements IManager {
 				}
 
 				// If a password is set. Hash it! (only if the password has changed)
-				if ($share->getPassword() !== null && $share->getPassword() !== $originalShare->getPassword()) {
+				if (($share->getPassword() !== null) &&
+						($share->getPassword() !== $originalShare->getPassword()) &&
+						($share->getShouldHashPassword() === true)) {
 					$share->setPassword($this->hasher->hash($share->getPassword()));
 				}
 			}
--- a/lib/private/Share20/Share.php
+++ b/lib/private/Share20/Share.php
@@ -69,6 +69,8 @@ class Share implements \OCP\Share\IShare {
 	private $name;
 	/** @var int */
 	private $state;
+	/** @var bool */
+	private $shouldHashPassword = true;
 
 	/** @var IRootFolder */
 	private $rootFolder;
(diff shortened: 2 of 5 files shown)


apps/files_sharing/lib/API/Share20OCS.php
```diff
@@ -498,6 +498,7 @@ public function createShare() {
 		$share->setShareType($shareType);
 		$share->setSharedBy($this->currentUser->getUID());
 
+		$share->setShouldHashPassword($this->request->getParam('shouldHashPassword', true));
```
PVince81: no, not over OCS

the request is on pure PHP API level

lib/private/Share20/Manager.php
```diff
@@ -843,7 +843,9 @@ public function updateShare(\OCP\Share\IShare $share) {
 				}
 
 				// If a password is set. Hash it! (only if the password has changed)
-				if ($share->getPassword() !== null && $share->getPassword() !== $originalShare->getPassword()) {
+				if (($share->getPassword() !== null) &&
+					($share->getPassword() !== $originalShare->getPassword()) &&
```
jvillafanez: This might need some extra indentation to distinguish the conditions and the statement.

lib/private/Share20/Share.php
```diff
@@ -80,6 +82,7 @@ public function __construct(IRootFolder $rootFolder, IUserManager $userManager)
 		$this->rootFolder = $rootFolder;
 		$this->userManager = $userManager;
 		$this->state = Constants::STATE_ACCEPTED;
+		$this->shouldHashPassword = true;
```
jvillafanez: you can set the default in the attribute declaration.

jvillafanez (commented):
just a couple of minor changes, but it looks good 👍

lib/private/Share20/Manager.php
sharidas: Added extra indentation.

lib/private/Share20/Share.php
sharidas: The default value is now set when the attribute is declared.

PVince81 (commented):
unit tests ?

sharidas:
Added unit test.

PVince81:
missing tests for share manager changes

sharidas:
> missing tests for share manager changes

I have added tests for share manager changes.

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport

sharidas:
Backport to stable10 : https://github.com/owncloud/core/pull/32572

PR #32505, opened 2018-08-29
use guzzle globally through a helper

individual-it:
## Description
1. use Guzzle through a Helper class
2. don't catch Guzzle exceptions all over the place, as there are not thrown any-more, but only check responses

## Related Issue
https://github.com/owncloud/QA/issues/586

## Motivation and Context
have only one place where guzzle can be configured globally

## How Has This Been Tested?
:robot: 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32559, opened 2018-09-04
acceptance tests - stop uploading when receiving an error

individual-it:
## Description
stop uploading with there is an error
when uploading chunks and receiving an error on one chunk, the upload helper would still doing the other steps of the upload

## Related Issue
- Fixes https://github.com/owncloud/firewall/issues/516

## Motivation and Context
Before  https://github.com/owncloud/core/pull/32505 there was an exception thrown on errors. We don't do that any-more, because the higher layers might just expect the error code. But now we need to return the error code when it happens and not doing other requests afterwards.

## How Has This Been Tested?
run failing tests of firewall

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/TestHelpers/UploadHelper.php
+++ b/tests/TestHelpers/UploadHelper.php
@@ -88,7 +88,7 @@ class UploadHelper {
 		if ($chunkingVersion === 1) {
 			$headers['OC-Chunked'] = '1';
 		} elseif ($chunkingVersion === 2) {
-			WebDavHelper::makeDavRequest(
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
The last ``if`` is not essential, but good to leave it there in case someone adds more code in future between it and the final return.

individual-it:
backport in #32561

PR #32589, opened 2018-09-05
Prevent LDAP Password to appear in log file

voroyam:
If connection to LDAP server is lost, the LDAP password is shown in the stack trace.

This should prevent this from happening.

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -101,7 +101,14 @@ class Log implements ILogger {
 		'decrypt',
 
 		//LoginController
-		'tryLogin'
+		'tryLogin',
+
+		//bind
+		'bind',
+
+		//invokeLDAPMethod
+		'invokeLDAPMethod'
+
 	];
 
 	/**


cdamken:
@tomneedham @PVince81 @butonic Could you please review this?

lib/private/Log.php
```diff
@@ -101,7 +101,14 @@ class Log implements ILogger {
 		'decrypt',
 
 		//LoginController
-		'tryLogin'
+		'tryLogin',
+
+		//bind
```
PVince81: these comments look useless...

PVince81 (approved):
:+1:

PVince81:
@voroyam please backport to stable10

lib/private/Log.php
voroyam: They are, any suggestions how to improve them?

voroyam:
@PVince81 don't know how to backport, already told @tom this.

PVince81:
stable10: https://github.com/owncloud/core/pull/32592

patrickjahns:
Regarding the comments about useless string - suggestion, mention that the two methods `bind` // `invokeLDAPMethod` result from `user_ldap`.

```
// user_ldap app 
'bind'
'invokeLDAPMethod'
```

PR #32225, opened 2018-08-02
preLogin hook should use string UID not IUser

tomneedham:
## Description
Everywhere else, `preLogin` uses string UID as the first param - this causes plugins to 💥 

## Related Issue
Found whilst debugging portal.

## Motivation and Context
Broken plugins / inconsistency.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32642, opened 2018-09-07
Revert "preLogin hook should use string UID not IUser"

sharidas:
Reverts owncloud/core#32225


sharidas:
Original PR which got reverted here: https://github.com/owncloud/core/pull/32225

PVince81 (approved):
:+1:

PR #32653, opened 2018-09-10
Fix sharedByHasTheFollowingTags

phil-davis:
## Description
1) In ``sharedByHasTheFollowingTags()``, reset ``found`` to ``false`` each time before looping to check for a matching tag. (1st commit)
2) 
a) Provide an explicit step to check for the case when a user requests the tags of a file that is owned by someone else - that should give a 404.
b) Provide explicit steps for the "should have no tags" conditions, rather than trying to include them as an empty table in the steps for having tags.
c) Make a new scenario for the case when a tag has already been assigned to a file, then the tag itself is deleted. The file should no longer have that tag. This common scenario was not anywhere to be seen.

## Motivation and Context
1) For steps that use ``sharedByHasTheFollowingTags()`` to check for more than 1 tag, you can put any tag name you like after the first tag name. Only the first tag name really gets checked, all the others always pass, if they are there or not.

2) The existing code to check for a file having no tags does not work. It happens to work in the 1 place that it is currently used, because that place has a scenario where the user concerned actually does not have a view of the file anyway, so the request for tags gets a 4** response.

## How Has This Been Tested?
Running some Tags acceptance tests locally.

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Tags.php
+++ b/tests/acceptance/features/bootstrap/Tags.php
@@ -459,6 +459,7 @@ trait Tags {
 		\array_shift($tagList);
 		$found = false;
 		foreach ($table->getRowsHash() as $rowDisplayName => $rowType) {
+			$found = false;
 			foreach ($tagList as $path => $tagData) {
 				if (!empty($tagData)
 					&& $tagData['{http://owncloud.org/ns}display-name'] === $rowDisplayName
--- a/tests/acceptance/features/bootstrap/Tags.php
+++ b/tests/acceptance/features/bootstrap/Tags.php
@@ -439,6 +439,24 @@ trait Tags {
 	}
 
 	/**
+	 * @Then /^the HTTP status when user "([^"]*)" requests tags for (?:file|folder|entry) "([^"]*)" (?:shared|owned) by "([^"]*)" should be "([^"]*)"$/
+	 *
+	 * @param string $user
+	 * @param string $fileName
+	 * @param string $sharingUser
+	 * @param string $status
+	 *
+	 * @return void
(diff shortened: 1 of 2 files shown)


phil-davis:
This is all useful stuff that I want to make use of in files_classifier tests.

tests/acceptance/features/bootstrap/Tags.php
```diff
@@ -452,13 +470,13 @@ public function sharedByHasTheFollowingTags(
...
 		$tagList = $this->requestTagsForFile($sharingUser, $fileName);
-		//Check if we are looking for no tags
-		if ((!\is_array($tagList)) && ($table->getRowAsString(0) === '|  |')) {
-			return true;
-		}
+		PHPUnit_Framework_Assert::assertInternalType('array', $tagList);
+		// The array of tags has a single "empty" item at the start.
+		// Remove this entry.
 		\array_shift($tagList);
 		$found = false;
```
individual-it: don't need that line anymore

individual-it (approved):

tests/acceptance/features/bootstrap/Tags.php
phil-davis: I added it there so I could explode "nicely" if the response is not an array, rather than get some PHP error doing ``array_shift()`` etc. on not-an-array. (it can be an object when something really bad has happened)

phil-davis:
Backport ``stable10`` #32656

PR #31879, opened 2018-06-25
remove testing caldav/carddav from travis pipeline

patrickjahns:
## Description
Testing for litmus / caldav /carddav has been migrated to drone already - this pr cleans up

## Motivation and Context
No need to wait for travis on this

## How Has This Been Tested?
🤖 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] CI

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #32674, opened 2018-09-11
[stable10] remove testing caldav/carddav/litmus from travis pipeline

phil-davis:
Backport #31879 

These litmus/carddav/caldav jobs are all happening in drone already on stable10.

I don't think there is a big rush to remove these.


phil-davis:
Travis not coming is expected.

PVince81 (approved):
:+1:

Issue #32684, opened 2018-09-12
 Invalid Token page is not shown when creating a new user via mail with some configuration


davitol:
Following the steps written in this ticket: https://github.com/owncloud/core/pull/32672

```
- Create `admin` user
- Create `user1` with `foo@gmail.com`
- Received the email to set the password for `user1`. Now delete `user1`
- Create `user1` again with a different email address, lets say `bar@gmail.com`
- Received the email for setting up the password.
- Now try to access the link from `foo@gmail.com`, user sees that token is invalid. Correct behaviour.
- Try to access link from `bar@gmail.com`, user sees password setting page, where new password can be set. Correct behaviour.
```

In the step **Now try to access the link from `foo@gmail.com`, user sees that token is invalid. Correct behaviour.**  => The invalid token page is not shown and a blank page is shown

Note: the exception log is spotted in owncloud.log (Right behavior)

```
[...]

PR #32690, opened 2018-09-13
[stable10] Backport of Fix the token exception classes to avoid infin…

sharidas:
…ite loop with getPrevious

Fix the users token exception class to avoid infinite
loop when getPrevious method is called.

## Description

This change removes the usage of `$this` to the previous exception arg passed. It was causing infinite loop while logging exceptions, when used with other apps.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/32684

## Motivation and Context

This change removes the usage of `$this` to the previous exception arg passed. It was causing infinite loop while logging exceptions, when used with other apps.

## How Has This Been Tested?

- Create user `admin` and login as admin
- Now try to create user `user1` with email as `bar@bar.com`
- Now delete `user1`
- Create `user1` again with email as `foo@bar.com`
- Now try to access the link obtained at `bar@bar.com`. Invalid token is observed.
- Try to access the link obtained at `foo@bar.com`. The set password page is shown to user. Correct behaviour.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/public/InvalidUserTokenException.php
+++ b/lib/public/InvalidUserTokenException.php
@@ -28,6 +28,7 @@ namespace OCP;
  * @since 10.0.10
  */
 class InvalidUserTokenException extends UserTokenException {
+
 	/**
 	 * InvalidUserTokenException constructor.
 	 *
@@ -35,7 +36,7 @@ class InvalidUserTokenException extends UserTokenException {
 	 * @param int $code
 	 * @since 10.0.10
 	 */
-	public function __construct($message = "", $code = 0) {
-		parent::__construct($message, $code, $this);
+	public function __construct($message = "", $code = 0, \Exception $previous = null) {
+		parent::__construct($message, $code, $previous);
 	}
 }
--- a/settings/Controller/UsersController.php
+++ b/settings/Controller/UsersController.php
@@ -522,7 +522,7 @@ class UsersController extends Controller {
 		try {
 			$this->checkPasswordSetToken($token, $userId);
 		} catch (UserTokenException $e) {
-			if ($e->getPrevious() instanceof UserTokenExpiredException) {
+			if ($e instanceof UserTokenExpiredException) {
 				return new TemplateResponse(
 					'settings', 'resendtokenbymail',
 					[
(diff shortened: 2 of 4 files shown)


davitol:
👎 Now it appears the page 'The token provided is invalid', but it appeared to me creating a new user

lib/public/InvalidUserTokenException.php
```diff
@@ -36,6 +41,17 @@ class InvalidUserTokenException extends UserTokenException {
 	 * @since 10.0.10
 	 */
 	public function __construct($message = "", $code = 0) {
-		parent::__construct($message, $code, $this);
+		$this->previousException = $this;
```
PVince81: this looks wrong. see how other exceptions always have an additional `$previous` argument in their constructors, just pass it through. don't mess with it

settings/Controller/UsersController.php
```diff
@@ -522,7 +522,7 @@ public function setPasswordForm($token, $userId) {
 		try {
 			$this->checkPasswordSetToken($token, $userId);
 		} catch (UserTokenException $e) {
-			if ($e->getPrevious() instanceof UserTokenExpiredException) {
+			if ($e->getPreviousTokenException() instanceof UserTokenExpiredException) {
```
PVince81: why not just `if ($e instance UserTokenExpiredException)` ?

lib/public/InvalidUserTokenException.php
sharidas: Removed the change. Added another arg `$previous` as additional argument in the constructor.

settings/Controller/UsersController.php
sharidas: Upated to using `$e instance of UserTokenExpiredException`. I was thinking like instance of `UserTokenException`, missed the part that it was extended. sigh.

PVince81:
Code looks good now.
Can you update the master PR as well ?

Once @davitol can confirm that it works and doesn't break other things we can merge this.

sharidas:
master branch PR is updated.

davitol:
It works 👍 @PVince81 @sharidas

PVince81 (approved):
:+1:

PR #32594, opened 2018-09-05
[stable10] tests async operations

individual-it:
## Description
tests for #32414 (moving files using async operation)

one test depends on changed in the testing app https://github.com/owncloud/testing/pull/21

## Related Issue
- Fixes #32580

## Motivation and Context
see issue

## How Has This Been Tested?
:robot: 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32703, opened 2018-09-14
do not use stream to check file downloads

individual-it:
## Description
after https://github.com/owncloud/core/pull/32594 we do user `stream` and `timeout` option together for guzzle, sometimes they don't work together and we receive `GuzzleHttp\Ring\Exception\RingException`
This specially happens with LDAP and encryption testing
I don't think we need the stream upload here in the first place
Even the change of #32594 is not yet in master we want to have the change here also

## How Has This Been Tested?
run failing tests in stable10 with the changes

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


phil-davis (approved):
If CI passes, then this must be OK

individual-it:
backport in #32704

PR #32704, opened 2018-09-14
[stable10] do not use stream to check file downloads

individual-it:
backport of #32703


phil-davis (approved):
If CI passes then this must be OK

PR #32731, opened 2018-09-17
Dont use preferences table quota any longer

tomneedham:
When quota is synced for an account (on login, and on user:sync), it is read from the user backend.

If the backend doesn't provide a quota, or, if it provides null - core (since 10.0.9 https://github.com/owncloud/core/commit/b62cad68f02a0edcc4c720170959b18cb6e44b51) will bring in the quota from `oc_preferences` which is where it was stored before `oc_accounts`. 

The problem comes now because the `user_management` app sets the quota directly on the accounts table - meaning we have two sources of truth for the quota value. If users have manually assigned quotas from their admin, these will have been stored in the accounts table. These will be overwritten with the `oc_preferences` values on user sync (login, or `user:sync`). 

This PR removes the logic from the `SyncService` to take the quota from the `oc_preferences` table - aligning the ownCLoud built in user management to fully utilise the `oc_accounts` table as the source of truth.

However, we now must consider the migration path. There can still be scenarios where the value in `oc_preferences` is the correct source of the quota value. Vice-versa, if the quota is changed in the UI, this value will have been set in the `oc_accounts` table and this should be the value used in future.

Migration proposal:

 - If quota is only in oc_preferences, use this, assign to `oc_accounts`, delete row from `oc_preference` - current behaviour
 - If quota is in `oc_preferences` and `oc_acccounts` and value is the same,  delete row from `oc_preference` - cleanup old row
 - If quota is in `oc_preferences` and `oc_accounts` and the value is different, use `oc_accounts` value,delete row from `oc_preference` - as it will have been assigned manually sign 10 upgrade

This should result in cleaned quotas from `oc_preferences`, and the correctly values present in `oc_accounts` - and future modifications will be persisted correctly. 

Scenarios to consider:
 - upgrade from 9.1 to 10 with ldap
 - direct install to 10.0.10 with this patch
 - setting quota from UI whilst using ldap

### DIFF
--- a/lib/private/User/SyncService.php
+++ b/lib/private/User/SyncService.php
@@ -223,12 +223,6 @@ class SyncService {
 				$a->setQuota($quota);
 			}
 		}
-		if ($quota === null) {
-			list($hasKey, $quota) = $this->readUserConfig($uid, 'files', 'quota');
-			if ($hasKey) {
-				$a->setQuota($quota);
-			}
-		}
 		if (\array_key_exists('quota', $a->getUpdatedFields())) {
 			$this->logger->debug(
 				"Setting quota for <$uid> to <$quota>", ['app' => self::class]


tomneedham:
The scenario in which you will lose manually assigned quotas is as follows:
 - You are running owncloud 10.0.9/10
 - The same instance was previously on <10 
 - A user has no quota configured (eg: ldap quota attribute empty, or backend provides no quota)

In this scenario, the old quota will be restored on login, or user:sync.

jvillafanez:
Should we include additional cases?
* If a quota is synced and then deleted in the backend, what is the expected ownCloud quota? should it be the ownCloud's default? should it keep the current quota?
* How can we reset the account's quota to the ownCloud's default?

The migration proposal looks good 👍 

I think we still can't mark the value as "modified by the admin" or something similar so the quota isn't changed in the next user sync. I mean, sync the quota, and then the admin change it to a different one. I'd expect the quota set by the admin to be kep after the next user sync

PVince81:
In a regular OC upgrade from OC < 10 there are already migrations that will move the quota from oc_preferences to oc_accounts then delete them from oc_preferences. So not sure in what scenario such values would still exist.

Despite this, I agree that we should remove any code that still relies on such values existing.

tomneedham:
> I think we still can't mark the value as "modified by the admin" or something similar so the quota isn't changed in the next user sync. I mean, sync the quota, and then the admin change it to a different one. I'd expect the quota set by the admin to be kep after the next user sync

I agree. IMO if the backend supplies the quota, you should not be able to change this assignment in owncloud usermanagement - otherwise we have some mismatch of source of metadata.  @pmaier1 for how we want this to work?

jvillafanez:
> @tomneedham @jvillafanez wasn't it so in past major versions that one could override the LDAP quota by setting one in user management ?

I think it's still possible, but resyncing the backend (on login too?) will reset the value again.

> I agree. IMO if the backend supplies the quota, you should not be able to change this assignment in owncloud usermanagement - otherwise we have some mismatch of source of metadata. @pmaier1 for how we want this to work?

As long as the admin knows that the quota is different than the one provided by the backend, it should be fine. Blocking the option should be easy though.

PVince81:
after discussing with @cdamken and @pako81 we thought that LDAP should always have authority.

this means that if LDAP is enabled we should prevent any modification of fields email, avatar, display name or quota if an LDAP attribute is specified for such fields in the config. If no field is specified, allow the admin to set values in the users page.

if we agree on this we can raise a new ticket to make the necessary changes for blocking input for all these fields in the users page, provisioning API and occ commands.

PVince81:
any objections to my proposal ? would make a ticket and have it handled separately

PVince81:
raised here: https://github.com/owncloud/core/issues/33186

tomneedham:
> this means that if LDAP is enabled we should prevent any modification of fields email, avatar, display name or quota if an LDAP attribute is specified for such fields in the config. If no field is specified, allow the admin to set values in the users page.

Yes - the admins should not be using owncloud as user administration if they are using ldap - this gives two sources of truth

butonic:
no. not for quota. some admins use the default quote we can configure for ldap users and later increase it manually. I guess the proper solution would be to be able to configure a quota per group. furthermore keep in mind that db users still need to be editable.

tomneedham:
- [ ] Cleanup quotas based on the migration proposal in original post
 - [ ] Block UI editing of quota if set by backend

PVince81:
@ownclouders rebase

ownclouders:

PVince81 (dismissed):
:+1: we should stop to use preference tables

ownclouders:

butonic:
note that we just added the username as a preference for mounting wnd drives ... when in truth the login column in the account table is exactly that ...

PVince81:
@tomneedham https://github.com/owncloud/core/pull/32731#issuecomment-446887894 I agree with adding a migration to cleanup.

Blocking editing in the UI I'd see this as out of scope for this PR. Let's do that separately, especially considering that it could be tricky. Currently the web UI stores its value into oc_accounts but this PR is about oc_preferences.

PVince81:
also need to clarify as it seems some admins think this is a legitimate use case

tomneedham:
This is not the correct fix. The preferences quota is required for overwrites performed in the web ui. 

The issue is that the web ui is not directly updating the accounts table not the perferences, which is not valid.

phil-davis:
Removing backport request because PR was closed.

PR #32785, opened 2018-09-20
Automate squizlabs/php_codesniffer

### LINKED ISSUE #32782: Add automated phpcs checks in drone for acceptance test code
The acceptance test code has been conforming to a proposed "standard" that was implemented some time ago but not formally enforced. It requires a bunch of PHP doc block standards, for example. See ``phpcs.xml`` in the root folder of the core repo.

The check is done with:
```
phpcs --standard=phpcs.xml tests/acceptance
```

Implement this much checking as a drone job, so that at least the acceptance test code stays to this standard.

Note: It is out-of-scope for this issue to decide what standard might apply to the whole code base and how to implement such a thing.

phil-davis:
## Description
- implement ``phpcs`` to enforce the existing acceptance test code style rules
- do it as an extra step of the existing ``test-php-style`` make target
- fix style of code

## Related Issue
- Fixes #32782 

## Motivation and Context
See issue.

## How Has This Been Tested?
Local run:
```
make test-php-style
```

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,7 +85,7 @@ help:
 	@echo -e "make clean\t\t\tclean everything"
 	@echo -e "make install-composer-deps\tinstall composer dependencies"
 	@echo -e "make update-composer\t\tupdate composer.lock"
-	@echo -e "make install-nodejs-deps\t\tinstall Node JS and Javascript dependencies"
+	@echo -e "make install-nodejs-deps\tinstall Node JS and Javascript dependencies"
 	@echo
 	@echo -e "Note that running 'make' without arguments already installs all required dependencies"
 	@echo
@@ -196,6 +196,7 @@ test-php-lint: $(composer_dev_deps)
 .PHONY: test-php-style
 test-php-style: $(composer_dev_deps)
 	$(composer_deps)/bin/php-cs-fixer fix -v --diff --diff-format udiff --dry-run --allow-risky yes
+	$(composer_deps)/bin/phpcs --runtime-set ignore_warnings_on_exit --standard=phpcs.xml tests/acceptance
 	php build/OCPSinceChecker.php
 
 .PHONY: test-php-style-fix
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "jarnaiz/behat-junit-formatter": "^1.3",
         "mikey179/vfsStream": "^1.6",
         "owncloud/coding-standard": "^1.0",
+        "squizlabs/php_codesniffer": "3.*",
         "phpunit/phpunit": "^5.7",
         "rdx/behat-variables": "^1.2",
         "roave/security-advisories": "dev-master",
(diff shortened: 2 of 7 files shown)


PVince81:
did you guys consider simply adding it to the existing `make test-php-style` rule ?

phil-davis:
@PVince81 good idea! Then developers will get the checks "for free".

I disabled the line length checks. Those give lots of warnings because the acceptance tests have many long lines due to the length of some Gherkin step text. The warnings make a lot of noise in the the output, we mostly ignored them anyway, and it would really annoy developers to get loads of crap output when doing ``make test-php-style``

phil-davis:
@PVince81 please review again.
And merge if you like it - needs the override of no Travis result.

Makefile
```diff
@@ -196,6 +196,7 @@ test-php-lint: $(composer_dev_deps)
 .PHONY: test-php-style
 test-php-style: $(composer_dev_deps)
 	$(composer_deps)/bin/php-cs-fixer fix -v --diff --diff-format udiff --dry-run --allow-risky yes
+	$(composer_deps)/bin/phpcs --runtime-set ignore_warnings_on_exit --standard=phpcs.xml tests/acceptance
```
PVince81: ah, so we won't be using the same rules for QA code ?

can you make a separate ticket to consolidate said rules so we only use a single tool ?

phil-davis: I don't understand your question.
``php-cs-fixer`` implements some "basic" rules across the whole code-base, including acceptance test code. That is already in place.

``phpcs`` has even more code style rules than ``php-cs-fixer``.

During discussions about code style, PHPdoc block etc standards some time ago, the QA team implemented some ``phpcs`` rules to get started, and applied it to the ``tests/acceptance/...`` folders. This is an "interim" standard while we await the "whole project" deciding on a project-wide standard. We have been manually running ``phpcs`` locally on PRs to see that we meet our standard, but of course we often forget!

So automating it here will force us (and anyone who touches ``tests/acceptance/.../*.php``) to remember.

If the whole project some day decides to adopt a standard that is inclusive of these rules, then easy - the acceptance test code already meets the rules here. If the project adopts some standard that is not the same/similar, then we reformat doc blocks or whatever (like the rest of the project will have to do anyway).

phil-davis:
@PVince81 this will need a merge overridde, because it is expecting a Travis result.

phil-davis:
Backport ``stable10`` #32900

Makefile
PVince81: hmm ok, so if phpcs itself has more rules than phpcsfixer we could eclipse phpcsfixer then... and many IDEs support loading phpcs rules but not ones from phpcsfixer...

PR #32794, opened 2018-09-21
Test public link share using email with email to self checked

skshetry:
## Description

This adds a test that the user and the creator of the link both gets the email if the `Send a copy to self` is checked.

## Related Issue

- Related to #30892 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32798, opened 2018-09-21
[stable10] Test public link share using email with email to self checked

skshetry:
## Description

Backport for #32794


skshetry:
Sent PR to the wrong branch. See #32799

PR #32807, opened 2018-09-23
Manage local storage via testing app

phil-davis:
## Description
Add methods that call the testing app to do file and folder deletes.

Use them to teardown ``local_storage`` after test scenarios.

Use them for ``fileHasBeenDeletedInLocalStorage($filename)`` 

Call the folder on the server-under-test ``server_tmp`` rather than the current ``work`` to differentiate it clearly, particularly in the "usual" case where the file system of the test runner and the system-under-test happen to be the same. Similarly, rename the folder for storing temporary files created by the test-runner from ``work`` to ``work_tmp``. This helps demonstrate a clear separation between the 2 folder usages, and that there are no "hanging" references to the old ``work`` folder.

## Motivation and Context
Test runner code currently creates ``local_storage`` via the testing app, but tries to clean it up at the end by deleting the files/folders in the local file system, rather than in the file system of the system-under-test.

There is also a test step method ``fileHasBeenDeletedInLocalStorage($filename)`` - that currently deletes the file in the local file system, which is not effective if the system-under-test is remote.

The testing app now has the ability to delete files and folders on the system-under-test. So we can use that.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/TestHelpers/SetupHelper.php
+++ b/tests/TestHelpers/SetupHelper.php
@@ -372,6 +372,38 @@ class SetupHelper {
 
 	/**
 	 *
+	 * @param string $dirPathFromServerRoot e.g. 'apps2/myapp/appinfo'
+	 * @param string|null $baseUrl
+	 * @param string|null $adminUsername
+	 * @param string|null $adminPassword
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public static function rmDirOnServer(
+		$dirPathFromServerRoot,
+		$baseUrl = null,
+		$adminUsername = null,
+		$adminPassword = null
+	) {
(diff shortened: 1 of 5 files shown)


phil-davis:
CI will fail here until testing app PR https://github.com/owncloud/testing/pull/29 is merged.
See #32808 for a demonstration that this code will work with that testing app PR.

phil-davis:
Backport ``stable10`` #32875

paurakhsharma (approved):

PR #32808, opened 2018-09-23
[PoC] [DO NOT MERGE] Manage local storage via testing app

phil-davis:
DO NOT MERGE

This is a proof of concept. It uses the testing app branch for PR https://github.com/owncloud/testing/pull/29 to get the ability to delete files and folders on the server via the testing app.

This shows that if that PR is merged into the testing app, then core PR #32807 will work.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -131,7 +131,7 @@ pipeline:
     pull: true
     commands:
       - ./tests/drone/install-server.sh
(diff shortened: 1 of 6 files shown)


phil-davis:
It passes, which is good.
No codecov came, but the care factor for that is zero.

phil-davis:
The testing app code has been merged now, so the main PR is passing.
Closing here, as it is no longer needed.

PR #32819, opened 2018-09-24
Allow REGULAR_USER_PASSWORD env var

phil-davis:
## Description
When acceptance test scenarios start, check if env var ``REGULAR_USER_PASSWORD`` is set, and use that value.

## Motivation and Context
``ADMIN_USERNAME`` and ``ADMIN_PASSWORD`` can be set as env variables before running the acceptance tests. There is also the default "regular user password", it would be nice to allow this to be overridden by an env variable also.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -196,6 +196,12 @@ trait BasicStructure {
 		if ($adminPasswordFromEnvironment !== false) {
 			$this->adminPassword = $adminPasswordFromEnvironment;
 		}
+
+		// get the regular user password from the environment (if defined)
+		$regularUserPasswordFromEnvironment = $this->getRegularUserPasswordFromEnvironment();
+		if ($regularUserPasswordFromEnvironment !== false) {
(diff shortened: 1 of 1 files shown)


phil-davis:
This will work, but if there is a requirement to pass in a range of all passwords that are used by the core tests then there will be more to be thought about. Many test scenarios have a step like:
```
		Given these users have been created but not initialized:
			|username|password|displayname|email       |
			|user1   |1234    |User One   |u1@oc.com.np|
```
which specifies the various attributes of the user, including the password.

And sometimes we want different users in the same scenario to have different passwords so we can do tests like making sure that when user1 provides the password of user2 that it does not work.

So we would need to pass in some set of passwords, or some base password from which we generate more or...

patrickjahns (approved):
LGTM -  please backport

phil-davis:
Backport ``stable10`` #32820

PR #32822, opened 2018-09-24
Abstract username and password

phil-davis:
## Description
1) Lots of places in the core acceptance tests mention the "admin" username literally, e.g. ``as user "admin"``. Implement a "functional" user ``%admin%`` and whenever that is used as a username, the code underneath replaces it with the actual current admin username.

The admin username defaults to ``admin`` and is overridden by setting env variable ``ADMIN_USERNAME``

2) The core acceptance tests use different passwords in various places.  Sometimes it is not critical, but just so that the different usernames in a scenario have different passwords to reduce the chance of a test passing with mixed-up authentication. Other times it is essential, e.g. when doing password change tests the original and new password have to be different.
So this PR provides "functional" passwords:

 Step form | Env Variable | Default | Description
----------|----------|----------|--------------------
 ``%admin%`` | ``ADMIN_PASSWORD``  | admin | password for the main administrator account that must already exist 
 ``%subadmin%`` | ``SUB_ADMIN_PASSWORD`` | IamAJuniorAdmin42 | password to use when creating a sub-admin account 
 ``%altadmin%`` | ``ALTERNATE_ADMIN_PASSWORD`` | IHave99LotsOfPriv | password to use when creating a second full administrator account 
 ``%regular%`` | ``REGULAR_USER_PASSWORD`` | 123456 | the default password for use with a new user 
 ``%alt1%`` | ``ALT1_USER_PASSWORD`` | IamAJuniorAdmin42 | an alternate password for use with regular users 
 ``%alt2%`` | ``ALT2_USER_PASSWORD`` | AaBb2Cc3Dd4 | an alternate password for use with regular users 
 ``%alt3%`` | ``ALT3_USER_PASSWORD`` | aVeryLongPassword42TheMeaningOfLife | an alternate password for use with regular users 

This should be backward-compatible with existing tests in apps. Specifying a "hard-coded" password or admin username still works. App acceptance tests can be updated to use these new username-password-abstraction features as needed.

## Motivation and Context
Be able to:
- run the acceptance test suite against a system-under-test that has a different admin username and/or password (not just admin/admin)
- run the acceptance test suite against a system-under-test that has various password policies in place. So we need to be able to specify the various passwords that the test scenarios should use for the different functional accounts that are created for testing (e.g. regular user, sub-admin and an extra admin account). That will allow the test runner to specify password that are complex-enough to be valid on the SUT.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -63,6 +63,35 @@ trait BasicStructure {
 	/**
 	 * @var string
 	 */
+	private $alt1UserPassword = '';
+
+	/**
+	 * @var string
+	 */
+	private $alt2UserPassword = '';
+
+	/**
+	 * @var string
+	 */
+	private $alt3UserPassword = '';
+
+	/**
+	 * The password to use in tests that create a sub-admin user
+	 *
+	 * @var string
+	 */
+	private $subAdminPassword = '';
+
+	/**
+	 * The password to use in tests that create another admin user
+	 *
+	 * @var string
+	 */
+	private $alternateAdminPassword = '';
+
+	/**
+	 * @var string
+	 */
 	private $ocPath = '';
 
 	/**
@@ -155,9 +184,12 @@ trait BasicStructure {
 	 *
 	 */
 	public function __construct(
-		$baseUrl, $adminUsername, $adminPassword, $regularUserPassword, $ocPath
+		$baseUrl,
+		$adminUsername,
+		$adminPassword,
+		$regularUserPassword,
+		$ocPath
 	) {
-
 		// Initialize your context here
 		$this->baseUrl = \rtrim($baseUrl, '/');
 		$this->adminUsername = $adminUsername;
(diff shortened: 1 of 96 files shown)


tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -168,6 +200,14 @@ public function __construct(
 		$this->cookieJar = new CookieJar();
 		$this->ocPath = $ocPath;
 
+		// These passwords are referenced in tests and can be overridden by
+		// setting environment variables.
```
phil-davis: Note: I did not add these to ``behat.yml`` and/or the constructor here.
Doing that would break lots of app acceptance tests - they would have to have these pasted into all their ``behat.yml`` - and that would just be pain for a few days.

That could be done in a later PR. But actually I don't see the benefit. These will be overridden by setting external env variables anyway, not by messing with ``behat.yml``

individual-it:
Looks good but why so many user passwords?

individual-it:
Could be audio helpful for LDAP tests, the imported ldif file can now have other passwords

phil-davis:
I wanted at least 2 alternate ones - some scenarios have 3 users, and it can be nice for them all to have different passwords. Or to have a "spare" password to use for checking that a different password does not work. So I added a 3rd alternate password "just to be sure". (otherwise we find we need it in an app, and then have to wait to add it in core...)

paurakhsharma (approved):

PR #32857, opened 2018-09-26
Add creating user to a service class

sharidas:
Moving creation of user to a service class.

## Description

Add the create user to a service class. It is moved from user_management apps UsersController.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Add the create user to a service class. It is moved from user_management apps UsersController. This will help us in integrating the change with the `user:add` command support creating users with username and email address.

## How Has This Been Tested?

- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Requires user_management PR https://github.com/owncloud/user_management/pull/75 , password_policy PR https://github.com/owncloud/password_policy/pull/154

## Open tasks:


sharidas:
Ouch.. When I run the unit test individually, I don't get failure. Will re-check again and try to fix this.

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+		$this->isEncryptionAppEnabled = $appManager->isEnabledForUser('encryption');
+		if ($this->isEncryptionAppEnabled) {
+			// putting this directly in empty is possible in PHP 5.5+
+			$result = $config->getAppValue('encryption', 'recoveryAdminEnabled', 0);
+			$this->isRestoreEnabled = !empty($result);
+		}
+	}
+
+	/**
+	 * @param $username
```
jvillafanez: missing type hint

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+				 */
+				$event = new GenericEvent();
+				$this->eventDispatcher->dispatch('OCP\User::createPassword', $event);
+				if ($event->hasArgument('password')) {
+					$password = $event->getArgument('password');
+				} else {
+					$password = $this->secureRandom->generate(20);
+				}
+			}
+			$user = $this->userManager->createUser($username, $password);
```
jvillafanez: If both password and email are empty, we still try to create the user with username and password. It seems wrong.

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+		} catch (\Exception $exception) {
+			$message = $exception->getMessage();
+			if (!$message) {
+				$message = $this->l10n->t('Unable to create user.');
```
jvillafanez: This is a bad approach. The message from the exception won't be translated, so you might be showing translated or untranslated messages to the user.

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+				$event = new GenericEvent();
+				$this->eventDispatcher->dispatch('OCP\User::createPassword', $event);
+				if ($event->hasArgument('password')) {
+					$password = $event->getArgument('password');
+				} else {
+					$password = $this->secureRandom->generate(20);
+				}
+			}
+			$user = $this->userManager->createUser($username, $password);
+		} catch (\Exception $exception) {
```
jvillafanez: Include a comment explaining why we need to catch a plain exception. If it's just for the `createUser` call, in addition to the comment adjust the length of the "try" block

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+			}
+			return new DataResponse(
+				[
+					'message' => (string)$message,
+				],
+				Http::STATUS_FORBIDDEN
+			);
+		}
+
+		if ($user instanceof User) {
```
jvillafanez: I'd prefer to use `if ($user !== false)`. In fact, I'd rather have a `if ($user === false)` and include the error code there. We expect the user to be created.

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+			[
+				'message' => (string)$this->l10n->t('Unable to create user.')
+			],
+			Http::STATUS_FORBIDDEN
+		);
+	}
+
+	/**
+	 * @param string $userId
+	 * @param string $email
```
jvillafanez: missing `@throws \Exception` (specially for a public method)

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+
+	private function isAdmin() {
+		// Check if current user (active and not in incognito mode)
+		// is an admin
+		$activeUser = $this->userSession->getUser();
+		if ($activeUser !== null) {
+			return $this->groupManager->isAdmin($activeUser->getUID());
+		}
+		// Check if it is triggered from command line
+		if (\OC::$CLI) {
```
jvillafanez: Any chance to avoid this check? The service should work the same regardless of being executed by cli or not, plus this should be covered by unittests.

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+		}
+		return false;
+	}
+
+	/**
+	 * @param IUser $user
+	 * @param array $userGroups
+	 * @return array
+	 */
+	public function formatUserForIndex(IUser $user, array $userGroups = null) {
```
jvillafanez: Why this method is public when it's only called inside this class? Either make it private or include a comment explaining the expected scenarios where this method will be called from other classes.

lib/private/User/Service/SigninWithEmail.php
```diff
@@ -0,0 +1,384 @@
...
+		if ($displayName === null) {
+			$displayName = '';
+		}
+
+		$avatarAvailable = false;
+		if ($this->config->getSystemValue('enable_avatars', true) === true) {
+			try {
+				$avatarAvailable = $this->avatarManager->getAvatar($user->getUID())->exists();
+			} catch (\Exception $e) {
+				//No avatar yet
```
jvillafanez: Log something here, even if it's just a debug message. If something goes wrong getting the avatar, this will be impossible to detect, debug and fix. Note that you're catching any exception even the `NobodyExpectsTheSpanishInquisitionException` that could happen uploading such file as avatar (it's a joke, but ilustrates why it's a bad idea swallowing exceptions like this)

jvillafanez:
I've just checked the `SigninWithEmail` class, not everything. I expect some changes that might affect the rest of the code.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,200 @@
...
+		}
+
+		try {
+			if (($password === '') && ($email !== '')) {
+				/**
+				 * Generate a random password as we are going to have this
+				 * use one time. The new user has to reset it using the link
+				 * from email.
+				 */
+				$event = new GenericEvent();
```
jvillafanez: Any chance of using a custom event instead of a generic one? A custom event would be better because it can have a specific method to set the password we want to use, as well as document what exceptions are allowed to be thown from the event handler, if any.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,200 @@
...
+			$user = $this->userManager->createUser($username, $password);
+		} catch (\Exception $exception) {
+			$message = $exception->getMessage();
+			if (!$message) {
+				$message = 'Unable to create user.';
+			}
+			throw new CannotCreateUserException($message);
+		}
+
+		if ($user instanceof User) {
```
jvillafanez: It might be better to reverse the logic:
```
if ($user === null) {
  throw new CannotCreateUserException(....);
}
```

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,200 @@
...
+
+	private function isAdmin() {
+		// Check if current user (active and not in incognito mode)
+		// is an admin
+		$activeUser = $this->userSession->getUser();
+		if ($activeUser !== null) {
+			return $this->groupManager->isAdmin($activeUser->getUID());
+		}
+		// Check if it is triggered from command line
+		if (\OC::$CLI) {
```
jvillafanez: any chance not to depend on this static variable?

lib/private/User/Service/CreateUser.php
sharidas: Added a new custom event. Let me know if this looks ok. As part of this change I made modification to password_policy app. Hence https://github.com/owncloud/password_policy/pull/154, needs review too.

lib/private/User/Service/CreateUser.php
sharidas: Updated as per the suggestion.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		try {
+			if (($password === '') && ($email !== '')) {
+				/**
+				 * Generate a random password as we are going to have this
+				 * use one time. The new user has to reset it using the link
+				 * from email.
+				 */
+				$passwordCreateEvent = new UserPasswordCreateEvent(UserPasswordCreateEvent::CREATE_PASSWORD, []);
```
jvillafanez: We have to keep backwards compatibility. I don't think we have a migration path for these events yet, so it's better not to touch it.
My point is that we can't assume that the app will be updated, or even use the new event. The apps need to work anyway. If the app is still using the old GenericEvent it will likely crash.

The idea is fine, but the old event still needs to be thrown. Maybe @PVince81 can decide how and when we want to move this forward.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,201 @@
...
+
+	private function isAdmin() {
+		// Check if current user (active and not in incognito mode)
+		// is an admin
+		$activeUser = $this->userSession->getUser();
+		if ($activeUser !== null) {
+			return $this->groupManager->isAdmin($activeUser->getUID());
+		}
+		// Check if it is triggered from command line
+		$cli = \in_array(\php_sapi_name(), ['cli']);
```
jvillafanez: The point of touching this was to make unittests easier. Even though the tests should mainly test public methods, all the code in the class should be covered by those unittests. Right now, this isn't possible because the php_sapi_name can't be mocked, which means that this method will never return false, and the corresponding code path won't be covered by unittests.

If you can't find a better solution, use the previous code (using the `OC::$CLI`) because we can change it inside the tests.

lib/private/User/Service/CreateUser.php
sharidas: I have reverted the code and brought back the GenericEvent.

lib/private/User/Service/CreateUser.php
sharidas: I have moved `php_sapi_name` to a method so that it can be mocked.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,208 @@
...
+
+		try {
+			if (($password === '') && ($email !== '')) {
+				/**
+				 * Generate a random password as we are going to have this
+				 * use one time. The new user has to reset it using the link
+				 * from email.
+				 */
+				$event = new GenericEvent();
+				$this->eventDispatcher->dispatch('OCP\User::createPassword', $event);
```
jvillafanez: Could you check if you can move this "password creation" to a specific class? I think this isn't the only place this event is triggered. Normally, one event is triggered from only one place, so I'd rather check other alternatives.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,208 @@
...
+			$user = $this->userManager->createUser($username, $password);
+		} catch (\Exception $exception) {
+			$message = $exception->getMessage();
+			if (!$message) {
+				$message = 'Unable to create user.';
+			}
+			throw new CannotCreateUserException($message);
+		}
+
+		if ($user === null) {
```
jvillafanez: According to the IUserManager interface, the `createUser` methods either throws an exception or return false. Maybe you want to check for `false`?

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,208 @@
...
+					$password = $this->secureRandom->generate(20);
+				}
+			}
+			$user = $this->userManager->createUser($username, $password);
+		} catch (\Exception $exception) {
+			$message = $exception->getMessage();
+			if (!$message) {
+				$message = 'Unable to create user.';
+			}
+			throw new CannotCreateUserException($message);
```
jvillafanez: Try to simplify this. Maybe something like:
```
throw new CannotCreateUserException("Unable to create user due to exception: {$exception->getMessage()}");
```
I don't know if it might be useful to include the class name of the exception too.

In addition, make sure the error messages of each thrown exception are different, so in case of some exception pops up we know where is being triggered from really quick.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,208 @@
...
+
+	private function isAdmin() {
+		// Check if current user (active and not in incognito mode)
+		// is an admin
+		$activeUser = $this->userSession->getUser();
+		if ($activeUser !== null) {
+			return $this->groupManager->isAdmin($activeUser->getUID());
+		}
+		// Check if it is triggered from command line
+		$cli = $this->triggeredFromCommandLine();
```
jvillafanez: This can't be mocked up.
Just use the usual `OC::$CLI`.
```
return \OC::$CLI;
```

lib/private/User/Service/UserSendMail.php
```diff
@@ -0,0 +1,162 @@
...
+								\OC_Defaults $defaults, ITimeFactory $timeFactory,
+								IL10N $l10n) {
+		$this->secureRandom = $secureRandom;
+		$this->config = $config;
+		$this->mailer = $mailer;
+		$this->urlGenerator = $urlGenerator;
+		$this->defaults = $defaults;
+		$this->timeFactory = $timeFactory;
+		$this->l10n = $l10n;
+		$this->fromMailAddress = Util::getDefaultEmailAddress('no-reply');
```
jvillafanez: Any reason to keep this as attribute? If it's used in only one place, there is no need for this. I'd rather use the constructor just for dependency assignment as much as possible.

lib/private/User/Service/UserSendMail.php
```diff
@@ -0,0 +1,162 @@
...
+	 * @return null
+	 * @throws InvalidUserTokenException
+	 * @throws UserTokenExpiredException
+	 * @throws UserTokenMismatchException
+	 */
+	public function checkPasswordSetToken($token, IUser $user) {
+		$splittedToken = \explode(':', $this->config->getUserValue($user->getUID(), 'owncloud', 'lostpassword', null));
+		if (\count($splittedToken) !== 2) {
+			$this->config->deleteUserValue($user->getUID(), 'owncloud', 'lostpassword');
+			throw new InvalidUserTokenException($this->l10n->t('The token provided is invalid.'));
```
jvillafanez: Exception messages shouldn't need to be translated

jvillafanez:
I see there is still some general cleanup to do, mainly remove some references no longer used (`use OC\User\Service\SigninWithEmail;` in the routes.php file and some others in the PHPDocs.

lib/private/User/Service/CreateUser.php
sharidas: Moved the password creation logic to CreatePassword class.

lib/private/User/Service/CreateUser.php
sharidas: Updated the code with `if ($user === false)`

lib/private/User/Service/CreateUser.php
sharidas: Updated as recommended.

lib/private/User/Service/CreateUser.php
sharidas: Updated with `return \OC::$CLI`

lib/private/User/Service/UserSendMail.php
sharidas: Updated the change set by removing it from the constructor.

lib/private/User/Service/UserSendMail.php
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * @param IUser $user
+	 * @return null
+	 * @throws \Exception
+	 */
+	public function sendNotificationMail($user) {
+		$email = $user->getEMailAddress();
+
+		if ($email !== '') {
```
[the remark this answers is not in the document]
sharidas: Updated.

lib/private/User/Service/UserSendMail.php
sharidas: Removed.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,197 @@
...
+	 * @throws UserAlreadyExistsException
+	 */
+	public function createUser($username, $password, array $groups= [], $email='') {
+		if ($email !== '' && !$this->mailer->validateMailAddress($email)) {
+			throw new InvalidEmailException("Invalid mail address");
+		}
+
+		$currentUser = $this->userSession->getUser();
+
+		if (!$this->isAdmin()) {
```
jvillafanez: I suggest to reorganize this code a bit:
* The `$this->isAdmin` will require a `IUser` or `null` as parameter. The code will be basically the same, but it won't fetch the user from the session, it wil be fetched by the `createUser` method
* We should add another `$this->isSubadminOfGroup(IUser $user, IGroup $group)`. The idea is to simplify the checks below. As above, the function should consider possible `null` params.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,197 @@
...
+						continue;
+					}
+
+					if (!$this->groupManager->getSubAdmin()->isSubAdminofGroup($currentUser, $groupObject)) {
+						unset($groups[$key]);
+					}
+				}
+			}
+
+			if (empty($groups)) {
```
jvillafanez:
```
if (!empty($groups)) {
} else {
}
```

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,197 @@
...
+	 * @param $username
+	 * @param $password
+	 * @param array $groups
+	 * @param string $email
+	 * @return bool|\OCP\IUser
+	 * @throws CannotCreateUserException
+	 * @throws InvalidEmailException
+	 * @throws UserAlreadyExistsException
+	 */
+	public function createUser($username, $password, array $groups= [], $email='') {
```
jvillafanez: adjust code style here.

jvillafanez: There are also some clarifications to be included in the PHPDocs:
* missing expected types for the username and password. Can I use an `IUser` as username? Can I use `null` as password? ...
* I think it's pretty normal that optional parameters are null by default. If you want to use a different default, include a comment saying what is the default value. For example, the password is optional; if I want to use the email, I have to include a password, so what should I write as password?
* Include information about the behaviour specially when the behaviour might change depending on the parameters. What happens if I use both a password and an email? What happens if I don't use any?

jvillafanez: I don't know if it would be easier to split this function in two: `createUserWithPassword` and `createUserWithEmail`

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,197 @@
...
+			$user = $this->userManager->createUser($username, $password);
+		} catch (\Exception $exception) {
+			throw new CannotCreateUserException("Unable to create user due to exception: {$exception->getMessage()}");
+		}
+
+		if ($user === false) {
+			throw new CannotCreateUserException('Unable to create user.');
+		}
+
+		if ($groups !== null) {
```
jvillafanez: This is confusing. The params of the function say that the variable is an array, so null is impossible. The problem is that you're reassigning variables, which is a bad practice.
In addition, make sure it's clear what the variable contains. At this point I don't know if it contains `IGroup`, the id's of the groups; if it's an array or null....

lib/private/User/Service/CreateUser.php
jvillafanez: I meant, literally, replace it. The `triggeredFromCommandLine` is pointless, and ultimately you'll return whatever the `OC::$CLI` contains.
In addition, add a PHPDoc to make sure the behaviour is known: the function will return true if the active user is admin, or if the active user isn't available it will return true if it's run from the CLI.

lib/private/User/Service/CreateUser.php
```diff
@@ -0,0 +1,197 @@
...
+		$this->userSendMail = $userSendMail;
+		$this->createPassword = $createPassword;
+	}
+
+	/**
+	 * @param $username
+	 * @param $password
+	 * @param array $groups
+	 * @param string $email
+	 * @return bool|\OCP\IUser
```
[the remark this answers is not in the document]
sharidasan: Updated the phpdoc when the method returns bool value.

core/Command/User/Add.php
```diff
@@ -133,51 +199,31 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-			$password
-		);
 
-		if ($user instanceof IUser) {
-			$output->writeln('<info>The user "' . $user->getUID() . '" was created successfully</info>');
+			$this->createUser($input, $output, $password, $email);
+			$this->addUserToGroup($input, $output, $this->userManager->get($uid));
+		} elseif ($input->getOption('email') && ($email !== null)) {
+			$newUser = $this->createUser->createUser($uid, '', $input->getOption('group'), $email);
+			if ($newUser instanceof User) {
```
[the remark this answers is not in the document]
sharidasan: I have updated to check for the bool value.

core/Command/User/Add.php
```diff
@@ -133,51 +199,31 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+				$user = $this->userManager->get($uid);
+				$user->setDisplayName($input->getOption('display-name'));
+				$output->writeln('Display name set to "' . $user->getDisplayName() . '"');
+			}
+			if ($input->getOption('group')) {
+				if (\is_array($input->getOption('group'))) {
+					foreach ($input->getOption('group') as $group) {
+						if ($this->groupManager->isInGroup($uid, $group)) {
+							$output->writeln('User "' . $uid . '" added to group "' . $group . '"');
+						}
```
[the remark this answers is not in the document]
sharidasan: Splitted the method `createUser` into 2: `createUser` and `addUserToGroups`. I opted for the first suggestion.

lib/private/User/Service/CreatePassword.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ */
+
+namespace OC\User\Service;
+
+use OCP\Security\ISecureRandom;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+class CreatePassword {
```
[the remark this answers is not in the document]
sharidasan: Done.

lib/private/User/Service/CreatePassword.php
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	/**
+	 * Create password
+	 *
+	 * @return string
+	 */
+	public function createPassword() {
+		$event = new GenericEvent();
+		$this->eventDispatcher->dispatch('OCP\User::createPassword', $event);
```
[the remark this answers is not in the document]
sharidasan: In my humble opinion, I would say lets try to have the generic event as of now. Because in future we will move this to a separate custom event. @PVince81 I hope we can agree with my opinion.

lib/private/User/Service/CreateUser.php
sharidasan: I have updated the pr. Let me know if this looks better.

lib/private/User/Service/CreateUser.php
sharidasan: Updated.

lib/private/User/Service/CreateUser.php
sharidasan: I have modified the code, so that `groups` will remain as array and will not change to group object or anything else. Hope this looks better now.

lib/private/User/Service/CreateUser.php
sharidasan: updated to use `groups` as array.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,216 @@
...
+	 * @param array $groups contains array of group names
+	 * @return bool true if user is added to groups, else false
+	 */
+	public function addUserToGroups(IUser $user, array $groups= []) {
+		$currentUser = $this->userSession->getUser();
+		if (!$this->isAdmin($currentUser)) {
+			if (!empty($groups)) {
+				foreach ($groups as $key => $group) {
+					$groupObject = $this->groupManager->get($group);
+					if ($groupObject === null) {
```
jvillafanez:
```
if ($groupObject === null || !$this->isSubadminOfGroup($currentUser, $groupObject))
```
In fact, you can move the check for null inside the `isSubadminOfGroup` function to simplify this piece.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,216 @@
...
+	 * @return bool true if user is admin or if triggered from command line, else false
+	 */
+	private function isAdmin($currentUser) {
+		// Check if current user (active and not in incognito mode)
+		// is an admin
+		if ($currentUser !== null) {
+			return $this->groupManager->isAdmin($currentUser->getUID());
+		}
+		// Check if it is triggered from command line
+		$cli = \OC::$CLI;
```
jvillafanez: just `return \OC::$CLI`. It's more straightforward

lib/private/User/Service/CreatePassword.php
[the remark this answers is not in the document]
jvillafanez: The problem is that I don't know an easy migration path. Worst case, we'll likely need to trigger the old event and also the new event, and this behaviour will need to stay for a while (because we can't remove the old event without notice).
It feels like we're putting a stone on the road because the rest of the roads have a stone on them.

core/Command/User/Add.php
```diff
@@ -133,51 +200,45 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+			}
 
-		if (!empty($groups)) {
-			// Make sure we init the Filesystem for the user, in case we need to
-			// init some group shares.
-			Filesystem::init($user->getUID(), '');
-		}
+			$this->addUserToGroup($input, $output, $this->userManager->get($uid));
+		} elseif ($input->getOption('email') && ($email !== null)) {
+			$newUser = $this->createUserService->createUser($uid, '', $email);
```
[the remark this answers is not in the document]
sharidasan: This code uses the service, because the service class handles the creation of user with email, and sending email to the newly created user.

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +164,27 @@ public function addUser() {
 		}
 
 		try {
-			$newUser = $this->userManager->createUser($userId, $password);
-			$this->logger->info('Successful addUser call with userid: '.$userId, ['app' => 'ocs_api']);
+			$password = ($password === null) ? '' : $password;
```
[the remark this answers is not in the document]
sharidasan: Done.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,216 @@
...
+		$this->passwordGeneratorService = $passwordGeneratorService;
+	}
+
+	/**
+	 * Create user using password or email or with both.
+	 *
+	 * @param string $username
+	 * @param string $password can be empty string, else string for password
+	 * @param string $email
+	 * @return bool|\OCP\IUser when user is created else false
```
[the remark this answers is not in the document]
sharidasan: Updated the doc.

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +164,27 @@ public function addUser() {
...
+			$newUser = null;
+			$userAddedToGroups = ($groups === null) ? [null] : $groups;
+			$user = $this->createUserService->createUser($userId, $password, $emailAddress);
+			if ($user instanceof User) {
+				$newUser = $this->userManager->get($userId);
```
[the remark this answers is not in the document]
sharidasan: Removed.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,216 @@
...
+			} catch (\Exception $e) {
+				$this->logger->error("Can't send new user mail to $email: " . $e->getMessage(), ['app' => 'settings']);
+			}
+		}
+
+		return $user;
+	}
+
+	/**
+	 * Add user to the groups
```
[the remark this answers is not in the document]
sharidasan: Updated the doc. Let me know if this sounds reasonable.

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +164,27 @@ public function addUser() {
...
+				$this->logger->info('Successful addUser call with userid: '.$userId, ['app' => 'ocs_api']);
+			}
 
 			if (\is_array($groups)) {
+				$result = $this->createUserService->addUserToGroups($user, $userAddedToGroups);
+				if ($result === false) {
+					$this->logger->error("User $userId could not be added to groups");
```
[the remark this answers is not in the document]
sharidasan: With the latest change, the user should be able to know about the groups which are not added.

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +164,27 @@ public function addUser() {
 		}
 
 		try {
-			$newUser = $this->userManager->createUser($userId, $password);
-			$this->logger->info('Successful addUser call with userid: '.$userId, ['app' => 'ocs_api']);
+			$password = ($password === null) ? '' : $password;
+			$newUser = null;
+			$userAddedToGroups = ($groups === null) ? [null] : $groups;
```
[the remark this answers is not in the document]
sharidasan: Updated.

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +164,27 @@ public function addUser() {
...
+				$result = $this->createUserService->addUserToGroups($user, $userAddedToGroups);
+				if ($result === false) {
+					$this->logger->error("User $userId could not be added to groups");
+				}
 				foreach ($groups as $group) {
-					$this->groupManager->get($group)->addUser($newUser);
-					$this->logger->info('Added userid '.$userId.' to group '.$group, ['app' => 'ocs_api']);
+					if ($newUser !== null) {
+						if (!$this->groupManager->isInGroup($newUser->getUID(), $group)) {
+							$this->groupManager->get($group)->addUser($newUser);
```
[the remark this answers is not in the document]
sharidasan: Updated.

core/Command/User/Add.php
```diff
@@ -92,13 +109,61 @@ protected function configure() {
 			);
 	}
 
+	protected function addUserToGroup(InputInterface $input, OutputInterface $output, IUser $user) {
```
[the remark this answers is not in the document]
sharidasan: Removed `InputInterface` and used `groups`

core/Command/User/Add.php
```diff
@@ -92,13 +109,61 @@ protected function configure() {
...
+		$groups = $input->getOption('group');
+
+		if (!empty($groups)) {
+			// Make sure we init the Filesystem for the user, in case we need to
+			// init some group shares.
+			Filesystem::init($user->getUID(), '');
```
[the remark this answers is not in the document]
sharidasan: Removed this change.

core/Command/User/Add.php
```diff
@@ -92,13 +109,61 @@ protected function configure() {
...
+		}
+
+		foreach ($groups as $groupName) {
+			$group = $this->groupManager->get($groupName);
+			if (!$group) {
+				$this->groupManager->createGroup($groupName);
+				$group = $this->groupManager->get($groupName);
```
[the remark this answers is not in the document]
sharidasan: Updated.

core/Command/User/Add.php
```diff
@@ -133,51 +200,45 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-
-		// Set email if supplied & valid
-		if ($email !== null) {
-			$user->setEMailAddress($email);
-			$output->writeln('Email address set to "' . $user->getEMailAddress() . '"');
-		}
 
-		$groups = $input->getOption('group');
+			$newUser = $this->createUser($input, $output, $password, $email);
+			if ($newUser === false) {
```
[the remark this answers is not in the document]
sharidasan: Updated.

core/Command/User/Add.php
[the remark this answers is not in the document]
sharidasan: It should be possible. I will update the change...

core/Command/User/Add.php
```diff
@@ -133,51 +200,45 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-			$group->addUser($user);
-			$output->writeln('User "' . $user->getUID() . '" added to group "' . $group->getGID() . '"');
+			if ($input->getOption('group')) {
+				$result = $this->createUserService->addUserToGroups($newUser, $input->getOption('group'));
+				if ($result === false) {
+					$output->writeln('<warning>Unable to add user: ' . $uid . ' to groups </warning>');
+					return 1;
+				}
+
+				if (\is_array($input->getOption('group'))) {
```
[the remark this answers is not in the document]
sharidasan: Done.

core/Command/User/Add.php
```diff
@@ -133,51 +200,45 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+					foreach ($input->getOption('group') as $group) {
+						if ($this->groupManager->isInGroup($uid, $group)) {
+							$output->writeln('User "' . $uid . '" added to group "' . $group . '"');
+						}
+					}
+				}
+			}
+		} else {
+			$output->writeln("<error>Interactive input or --password-from-env is needed for entering a password!</error>");
+			return 1;
```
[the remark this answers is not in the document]
sharidasan: Returning different values.

core/routes.php
```diff
@@ -38,6 +38,10 @@
 		['name' => 'lost#resetform', 'url' => '/lostpassword/reset/form/{token}/{userId}', 'verb' => 'GET'],
 		['name' => 'lost#setPassword', 'url' => '/lostpassword/set/{token}/{userId}', 'verb' => 'POST'],
 		['name' => 'user#getDisplayNames', 'url' => '/displaynames', 'verb' => 'POST'],
+		['name' => 'user#setPassword', 'url' => '/setpassword/{token}/{userId}', 'verb' => 'POST'],
+		['name' => 'user#resendToken', 'url' => '/resend/token/{userId}', 'verb' => 'POST'],
+		['name' => 'user#setPasswordForm', 'url' => '/setpassword/form/{token}/{userId}', 'verb' => 'GET'],
+		['name' => 'avatar#getAvatar', 'url' => '/avatar/{userId}/{size}', 'verb' => 'GET'],
```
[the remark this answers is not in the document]
sharidasan: Removed.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,216 @@
...
+	 *
+	 * @param IUser $user
+	 * @param array $groups contains array of group names
+	 * @return bool true if user is added to groups, else false
+	 */
+	public function addUserToGroups(IUser $user, array $groups= []) {
+		$currentUser = $this->userSession->getUser();
+		if (!$this->isAdmin($currentUser)) {
+			if (!empty($groups)) {
```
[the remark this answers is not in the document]
sharidasan: Moved to a different method.

core/Command/User/Add.php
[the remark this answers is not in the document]
sharidasan: Updated.

core/Application.php
```diff
@@ -87,7 +88,19 @@ public function __construct(array $urlParams= []) {
 				$c->query('AppName'),
 				$c->query('Request'),
 				$c->query('UserManager'),
-				$c->query('Defaults')
+				$c->query('Defaults'),
+				new UserSendMailService(
```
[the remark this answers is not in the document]
sharidasan: No. It was throwing exception saying, `Could not resolve OCA\\\\Core\\\\Controller\\\\UserController! Class OCA\\\\Core\\\\Controller\\\\UserController does not exist or error resolving constructor arguments: Could not resolve UserSendMailService`

core/Command/User/Add.php
```diff
@@ -92,34 +109,77 @@ protected function configure() {
...
+		if ($passwordFromEnv) {
 			$password = \getenv('OC_PASS');
 			if (!$password) {
 				$output->writeln('<error>--password-from-env given, but OC_PASS is empty!</error>');
-				return 1;
+				return 4;
 			}
-		} elseif ($input->isInteractive()) {
+			$this->createUser($input, $output, $password, $email);
+			$this->addUserToGroup($groupInput, $output, $this->userManager->get($uid));
```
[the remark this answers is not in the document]
sharidasan: Its reusing the new user object.

core/Command/User/Add.php
```diff
@@ -131,53 +191,49 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-
-		// Set email if supplied & valid
-		if ($email !== null) {
-			$user->setEMailAddress($email);
-			$output->writeln('Email address set to "' . $user->getEMailAddress() . '"');
-		}
-
-		$groups = $input->getOption('group');
+			$newUser = $this->createUser($input, $output, $password, $email);
+			//If $newUser value is 1, then user creation failed
```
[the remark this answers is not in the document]
sharidasan: The usage of `$this->createUser` is removed. So we need another review.

core/Command/User/Add.php
```diff
@@ -131,53 +191,49 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+				list($result, $failedGroups) = $this->createUserService->addUserToGroups($newUser, $groupInput);
+				if ($result === false) {
+					$failedGroups = \implode(',', $failedGroups);
+					$output->writeln('<warning>Unable to add user: ' . $uid . ' to groups ' . $failedGroups . '.</warning>');
+					return 7;
+				}
+
+				if (\is_array($groupInput)) {
+					foreach ($groupInput as $group) {
+						if ($this->groupManager->isInGroup($uid, $group)) {
```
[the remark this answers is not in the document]
sharidasan: Removed the if condition.

core/Command/User/Add.php
```diff
@@ -131,53 +191,49 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
 
-		foreach ($groups as $groupName) {
-			$group = $this->groupManager->get($groupName);
-			if (!$group) {
-				$this->groupManager->createGroup($groupName);
-				$group = $this->groupManager->get($groupName);
-				$output->writeln('Created group "' . $group->getGID() . '"');
+			$output->writeln('<info>The user "' . $newUser->getUID() . '" was created successfully</info>');
+			if ($input->getOption('display-name')) {
+				$user = $this->userManager->get($uid);
```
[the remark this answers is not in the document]
sharidasan: Updated.

core/Command/User/Add.php
```diff
@@ -131,53 +176,64 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+			if ($newUser instanceof IUser) {
+				$output->writeln('<info>The user "' . $uid . '" was created successfully</info>');
+				list($result, $failedGroups) = $this->createUserService->addUserToGroups($newUser, $groupInput);
+				if ($result === false) {
+					$failedGroups = \implode(',', $failedGroups);
+					$output->writeln('<warning>Unable to add user: ' . $uid . ' to groups ' . $failedGroups . '.</warning>');
+					return 7;
+				} else {
+					foreach ($groupInput as $groupName) {
+						if ($this->groupManager->isInGroup($uid, $groupName)) {
```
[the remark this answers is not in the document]
sharidasan: updated.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,240 @@
...
+		$currentUser = $this->userSession->getUser();
+		$failedToAdd = [];
+		if (!$this->isAdmin($currentUser)) {
+			$groups = $this->setUpGroupsForUser($currentUser, $groups);
+		}
+
+		$userAddedToGroup = false;
+		if (\count($groups) >= 1) {
+			foreach ($groups as $groupName) {
+				if ($groupName !== null) {
```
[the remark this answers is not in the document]
sharidasan: This null check comes from https://github.com/owncloud/core/pull/32857/files#r227814464

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +165,15 @@ public function addUser() {
 		}
 
 		try {
-			$newUser = $this->userManager->createUser($userId, $password);
+			$groups = ($groups === null) ? [] : $groups;
+			$newUser = $this->createUserService->createUser($userId, $password, $emailAddress);
 			$this->logger->info('Successful addUser call with userid: '.$userId, ['app' => 'ocs_api']);
 
 			if (\is_array($groups)) {
```
jvillafanez: we should try to remove this check by making sure this will always be an array.

core/Command/User/Add.php
```diff
@@ -94,32 +111,60 @@ protected function configure() {
...
+				return 4;
+			}
+
+			$newUser = $this->createUserService->createUser($uid, $password, $email);
+			if ($newUser instanceof IUser) {
+				$output->writeln('<info>The user "' . $uid . '" was created successfully</info>');
+				$failedGroups = $this->createUserService->addUserToGroups($newUser, $groupInput);
+				if (\count($failedGroups) > 0) {
+					$failedGroups = \implode(',', $failedGroups);
+					$output->writeln('<warning>Unable to add user: ' . $uid . ' to groups ' . $failedGroups . '.</warning>');
```
jvillafanez: try to use string interpolation as much as possible.

core/Command/User/Add.php
```diff
@@ -94,32 +111,60 @@ protected function configure() {
...
+					return 7;
+				} else {
+					foreach ($groupInput as $groupName) {
+						if ($this->groupManager->isInGroup($uid, $groupName)) {
```
jvillafanez: This check is already done in the `addUserToGroups`, no need to duplicate.

core/Command/User/Add.php
```diff
@@ -131,53 +176,63 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-		}
-
-		$groups = $input->getOption('group');
-
-		if (!empty($groups)) {
-			// Make sure we init the Filesystem for the user, in case we need to
-			// init some group shares.
-			Filesystem::init($user->getUID(), '');
-		}
+			$newUser = $this->createUserService->createUser($uid, $password, $email);
```
jvillafanez: double-check if there is any difference with the code added above. Move the code to a private function if possible to avoid the duplication.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,170 @@
...
+	 * to groups. By default it is set to true.
+	 * @param IUser $user
+	 * @param array $groups contains array of group names
+	 * @param bool $checkInGroup if true then return array will be populated groups who failed to add user
+	 * @return array group names who failed to add user
+	 */
+	public function addUserToGroups(IUser $user, array $groups= [], $checkInGroup = true) {
+		$failedToAdd = [];
+
+		if (\count($groups) >= 1) {
```
jvillafanez: This shouldn't be needed. The loop won't be executed if the array is empty.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,170 @@
...
+			foreach ($groups as $groupName) {
+				if ($groupName !== '') {
```
jvillafanez: any reason for this check?

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,170 @@
...
+					$groupObject = $this->groupManager->get($groupName);
+
+					if (empty($groupObject)) {
+						$groupObject = $this->groupManager->createGroup($groupName);
+					}
+					$groupObject->addUser($user);
+					if ($checkInGroup && !$this->groupManager->isInGroup($user->getUID(), $groupName)) {
+						$failedToAdd[] = $groupName;
+					} else {
+						$this->logger->info('Added userid ' . $user->getUID() . ' to group ' . $groupObject->getGID());
```
jvillafanez: I'd prefer to remove this log because it might be inaccurate, at least with the current code. If the checkInGroup is false, the user might be be added but we'll show a log stating that the user has been added. This will be very troublesome to debug.

core/Command/User/Add.php
sharidasan: I will update this change.

lib/private/User/Service/CreateUserService.php
sharidasan: Removed

lib/private/User/Service/CreateUserService.php
sharidasan: Removed.

core/Command/User/Add.php
sharidasan: Removed.

core/Command/User/Add.php
sharidasan: Mostly done. Need to re-check again.

apps/provisioning_api/lib/Users.php
sharidasan: Removed the check. And one round of manual testing it was working. Will re-check again.

apps/provisioning_api/lib/Users.php
```diff
@@ -156,13 +165,15 @@ public function addUser() {
 		}
 
 		try {
-			$newUser = $this->userManager->createUser($userId, $password);
+			$groups = ($groups === null) ? [] : $groups;
```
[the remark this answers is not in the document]
sharidasan: Done.

apps/provisioning_api/lib/Users.php
```diff
@@ -126,7 +133,9 @@ public function getUsers() {
 	public function addUser() {
 		$userId = isset($_POST['userid']) ? $_POST['userid'] : null;
 		$password = isset($_POST['password']) ? $_POST['password'] : null;
+		$password = ($password === null) ? '' : $password;
```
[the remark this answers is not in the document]
sharidasan: Removed.

DeepDiver1975:
outdated -> close

micbar:
@sharidas This PR seems very outdated.
We do not need UI parts, they are already done.

Wouldn't it be better to start from scratch?

DeepDiver1975:

PR #32898, opened 2018-09-27
Fix helper code style

PVince81:


PVince81:
fixes CI issue on master

phil-davis (commented):
Yeh, it wants the comment to be indented like the line that follows it, which happens to be the ``else``

PR #32901, opened 2018-09-28
phpcs style changes for master

phil-davis:
A few more style issues that crept in. We need to implement the automation to stop this happening.

The phpcs warning for line length is just annoying noise, because actually there are many lines of Gherkin that exceed the length. And so we ignore those warnings anyway. The check is disabled.


paurakhsharma (approved):

phil-davis:
Backport ``stable10`` #32902

PR #32908, opened 2018-09-28
Revert "Bump league/flysystem from 1.0.46 to 1.0.47"

PVince81:
Reverts owncloud/core#32739

the lib update brings in a breaking change by requiring the fileinfo module to be available, which OC only lists as "optional". we can't change such requirements in a minor version.

See https://github.com/thephpleague/flysystem/issues/968 for upstream issue.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -912,26 +912,26 @@
         },
         {
             "name": "league/flysystem",
-            "version": "1.0.47",
+            "version": "1.0.46",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


DeepDiver1975 (approved):

PVince81:
according to https://github.com/thephpleague/flysystem/issues/968 this was always a requirement but it wasn't enforced. let's proceed with the update then in https://github.com/owncloud/core/pull/33198

PR #32910, opened 2018-09-29
Explicitly set display name for API search acceptance tests

phil-davis:
## Description
In ``search.feature`` explicitly set the display name of ``user0``

## Motivation and Context
When ``user_ldap`` runs the core acceptance tests it fails on getting the display name attribute of ``user0``. The ``user_ldap`` environment has the display name of ``user0`` already set to ``User Zero``

So in core tests we need to be careful to expect the same thing.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #32911, opened 2018-09-29
Consolidate acceptance test knowledge about known users

phil-davis:
## Description
1) Make the core acceptance test code know what is the password, display name, and email address of each of the "fixed" users that are setup in user_ldap testing.
2) "Just create" these users in Gherkin steps without specifying the password, display name, and email address. Leave it to the code underneath to fill in that information. This reduces the duplication of password, display name and email address throughout the feature files.

## Motivation and Context
Acceptance tests in user_ldap app run tests from core with a known set of users already setup in LDAP. We have problems when we change or add acceptance tests in core. We have to remember not to "break the rules" of what will work for user_ldap.

e.g. ``apiWebdavOperations/search.feature`` was recently added. It created a "default" ``user0`` and then the test expected the display name of ``user0`` to be just ``user0``. But in LDAP that user has a display name of ``User Zero``. We can fixup just the core test, like in PR #32910 but it would be nice to try and avoid this as much as reasonable for the future.

## How Has This Been Tested?
Some local acceptance test runs. CI will run the complete suites.

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -76,6 +76,11 @@ trait BasicStructure {
 	private $alt3UserPassword = '';
 
 	/**
+	 * @var string
+	 */
+	private $alt4UserPassword = '';
+
+	/**
 	 * The password to use in tests that create a sub-admin user
 	 *
 	 * @var string
@@ -217,6 +222,7 @@ trait BasicStructure {
 		$this->alt1UserPassword = "1234";
 		$this->alt2UserPassword = "AaBb2Cc3Dd4";
 		$this->alt3UserPassword = "aVeryLongPassword42TheMeaningOfLife";
+		$this->alt4UserPassword = "ThisIsThe4thAlternatePwd";
 		$this->subAdminPassword = "IamAJuniorAdmin42";
 		$this->alternateAdminPassword = "IHave99LotsOfPriv";
 		$this->publicLinkSharePassword = "publicPwd1";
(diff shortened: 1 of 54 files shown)


phil-davis:
To do next:
```
And the user has logged in with username "user1" and password "%alt1%" using the webUI
```

Make a step like:
```
And the user has logged in with username "user1" using the webUI
```
which looks up the already-known user password, and uses it to login.

That will get rid of another heap of places where the knowledge that ``user1`` has password ``%alt1%`` is duplicated through the feature files.

tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -1373,13 +1394,113 @@ public function getPasswordForUser($userName) {
...
+		} elseif ($userName === 'user0') {
+			return (string) $this->regularUserPassword;
+		} elseif ($userName === 'user1') {
+			return (string) $this->alt1UserPassword;
+		} elseif ($userName === 'user2') {
+			return (string) $this->alt2UserPassword;
+		} elseif ($userName === 'user3') {
+			return (string) $this->alt3UserPassword;
+		} elseif ($userName === 'user4') {
+			return (string) $this->alt4UserPassword;
```
phil-davis: I added ``alt4UserPassword`` because it fitted the pattern and we have ``user0`` to ``user4`` defined in LDAP.

tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -1373,13 +1394,113 @@ public function getPasswordForUser($userName) {
...
+		} elseif ($userName === 'usergrp') {
+			return (string) $this->regularUserPassword;
+		} elseif ($userName === 'sharee1') {
+			return (string) $this->regularUserPassword;
```
phil-davis: The passwords for ``usergrp`` and ``sharee1`` do not matter, they are fine to just use ``regularUserPassword``
They are used mostly so that there will be extra interesting entries in auto-match lists...

individual-it (approved):

PR #32923, opened 2018-10-01
Add loop config option in `run.sh`

skshetry:
## Description
This adds a `--loop` option in `run.sh` that helps running a tests multiple times easier and helps on finding out timing problems.

## Motivation and Context
webUI tests "always" have timing issues trying to make sure to wait until the the UI is ready after each action is done. For test development, it is very handy to be able to run tests in a loop and go for coffee, lunch, or even home for the night. Then see if the test at least runs reliably on your local machine.

## How Has This Been Tested?
Local testing in loops.

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -13,6 +13,7 @@ OCC=${OC_PATH}occ
 BEHAT=${OC_PATH}lib/composer/bin/behat
 
 BEHAT_TAGS_OPTION_FOUND=false
+BEHAT_RERUN_TIMES=1
 
 # The following environment variables can be specified:
 #
@@ -74,6 +75,7 @@ fi
 #            testing app. We have to assume it is already enabled.
 # --show-oc-logs - tail the ownCloud log after the test run
 # --norerun - do not rerun failed webUI scenarios
+# --loop  - loop tests for given number of times. Only use it for debugging purposes
 # --part - run a subset of scenarios, need two numbers.
 #          first number: which part to run
 #          second number: in how many parts to divide the set of scenarios
@@ -94,6 +96,10 @@ do
 			;;
 		--suite)
 			BEHAT_SUITE="$2"
+			shift
+			;;
+		--loop)
+			BEHAT_RERUN_TIMES="$2"
 			shift
 			;;
 		--type)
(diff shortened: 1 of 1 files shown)


tests/acceptance/run.sh
```diff
@@ -96,6 +97,10 @@ do
 			BEHAT_SUITE="$2"
 			shift
 			;;
+		--loop)
```
phil-davis: Let's call it ``--rerun``
And add it to  the comments near line 77

tests/acceptance/run.sh
```diff
@@ -861,8 +866,11 @@ for i in "${!BEHAT_SUITES[@]}"
 	do
 	BEHAT_SUITE_OPTION="--suite=${BEHAT_SUITES[$i]}"
 	SUITE_FEATURE_TEXT="${BEHAT_SUITES[$i]}"
-
-	run_behat_tests
+	for i in $(seq 1 $BEHAT_RERUN_TIMES)
+		do
+		echo "Running $i th time"
```
phil-davis: We could say:
```
Test repeat $rerun_number of $BEHAT_RERUN_TIMES
```

tests/acceptance/run.sh
```diff
@@ -861,8 +866,11 @@ for i in "${!BEHAT_SUITES[@]}"
 	do
 	BEHAT_SUITE_OPTION="--suite=${BEHAT_SUITES[$i]}"
 	SUITE_FEATURE_TEXT="${BEHAT_SUITES[$i]}"
-
-	run_behat_tests
+	for i in $(seq 1 $BEHAT_RERUN_TIMES)
```
phil-davis: I don't think you can use the same loop variable here that is already used at line 865.
Maybe call it ``rerun_number``

phil-davis (changes requested):
See comments.

skshetry:
@phil-davis I added the things you requested

tests/acceptance/run.sh
```diff
@@ -861,8 +867,14 @@ for i in "${!BEHAT_SUITES[@]}"
...
 	BEHAT_SUITE_OPTION="--suite=${BEHAT_SUITES[$i]}"
 	SUITE_FEATURE_TEXT="${BEHAT_SUITES[$i]}"
-
-	run_behat_tests
+	for rerun_number in $(seq 1 $BEHAT_RERUN_TIMES)
+		do
+		if (($BEHAT_RERUN_TIMES > 1))
+		then
+			echo -e "\nTest repeat $rerun_number of $BEHAT_RERUN_TIMES"
+		fi
```
skshetry: The message would be annoying if the `--loop` isn't used at all. So, I added a conditional check that only shows the message when the `loop` is used.

phil-davis (approved):

skshetry:
Backport for `stable10` on #32924

phil-davis:
Note: this ``--loop`` option is designed for doing local repeated runs of a whole suite, whole feature or single scenario. It runs the requested tests "over and over" the requested number of times, whether they pass or not.

If you use it, great, but be aware that the duration of the repeated tests might be "a long time" (tm).

It is NOT designed to be used in CI, because it reruns again and again regardless of individual test pass/fail.

Issue #32978, opened 2018-10-03
occ commands are not exiting with error status after symfony/console 3.4.16 update


phil-davis:
See issue raised with the symfony project: https://github.com/symfony/symfony/issues/28695

### Steps to reproduce
1) have ownCloud installed with ``symfony/console`` 3.4.15 e.g.:
```
git checkout v10.0.10
```

2) run some symfony console ``occ`` command in a way that gives an error, e.g. an example ownCloud console command:
```
php occ user:add aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
Enter password: 
Confirm password: 

In Manager.php line 355:
                                                     
  The username can not be longer than 64 characters  

user:add [--password-from-env] [--display-name [DISPLAY-NAME]] [--email [EMAIL]] [-g|--group [GROUP]] [--] <uid>

$ echo $?
1
```
3) update to ``symfony/console`` 3.4.16
```
composer update symfony/console
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Updating symfony/console (v3.4.15 => v3.4.16): Loading from cache
Writing lock file
Generating optimized autoload files
```
4) repeat the console command:
```
php occ user:add aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
Enter password: 
Confirm password: 

In Manager.php line 355:
                                                     
  The username can not be longer than 64 characters  

user:add [--password-from-env] [--display-name [DISPLAY-NAME]] [--email [EMAIL]] [-g|--group [GROUP]] [--] <uid>

$ echo $?
0
```

### Expected behaviour
Exit status should be 1

### Actual behaviour
Exit status is 0

### Server configuration
**Operating system**:
Ubuntu 18.04

**Web server:**
Development environment

**Database:**
mySQL

**PHP version:**
5.6.38 7.1.22 and 7.2.10 all show the same problem.

**ownCloud version:** (see ownCloud admin page)
core ``stable10`` with recent ``symfony/console`` bump to 3.4.16

phil-davis:
This was noticed because password policy tests started failing. They do ``occ`` password change commands and check the exit status.

[...]

PR #32994, opened 2018-10-03
Revert "Bump symfony 3.4.15 to 3.4.16"

phil-davis:
Reverts owncloud/core#32949

See issue #32978 - there is a regression in ``symfony/console``

The regression is fixed, and will be released in symfony 3.4.17 but that will take time to come. The regression is causing password policy ``occ```  command acceptance test failures. It also effects all ``occ`` commands. core acceptance tests are only passing because we do not have test coverage at the moment.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,7 +1,7 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: we could just change ``symfony/console`` back to 3.4.15 - but it is much easier to revert this whole PR and do it again when 3.4.17 comes out.

phil-davis:
See PR #32995 for the revert in ``stable10``

paurakhsharma (approved):

PR #33009, opened 2018-10-04
API test to check that share can be made with file names more than 64 chars long

paurakhsharma:
## Description
This PR adds tests for sharing of files and folders with name longer than 64 characters

Related issue: #32763

## How Has This Been Tested?
Locally

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Acceptance test

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #33010, opened 2018-10-04
[stable10] API test to check that share can be made with file names more than 64 chars long

paurakhsharma:
Backport of: #33009


phil-davis (changes requested):
Needs updating after the master PR review is approved.

phil-davis (approved):

PR #33035, opened 2018-10-05
29873 read only mount

### LINKED ISSUE #29873: [Feature Request] Mark mount as read only in the mount settings
If you mount a storage, it could be that the user has no write rights.
It is not when we do sharing where we can already set a read only flag which we further do/could use (see later).

It targets those mounts where the accessing user with his credentials has no rights to write to. This is handled by the accessed storage and not by owncloud.
I am only talking on full mounts and not about possible different subfolder permissions !

Because we do not see this write disability, a user might try to write into and gets a error message that does not give him the proper cause. If I remember correctly, if we have a shared read only mount, the error message is correct and helpful.

Thinking further, this is also true for the user at the client who has no clue why he can´t write to a mount - and if he does he can do into the client´s filesystem but the data never gets synced.

I do not know if a read only marker for a read only mount gets currently already propagated to the client. If not this should be done and if yes this FR is only an extension if set on the mount. But anyways this is a part of core to do so and a part of the client to take an action which is to be determined.

**Proposal:**
Adding the ability to mark a storage as read only.

@PVince81 @SamuAlfageme

TJHeeringa:
## Description

TL;DR: Made a read only option for external storages

I added a extra option in the advanced settings menu when mounting an external storage. This causes an extra row to be added in the database, when saving the mount, with whether an the mount is read only or not. A storagewrapper is added that checks the option and if the mount has the read only option the storagewrapper adds an permissionmask that makes the mount read only.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/29873

## Motivation and Context
Sometimes you want to have a group only to be able to read a certain map. Currently no option exists to do this. Extra info can be found in:
https://github.com/owncloud/user_management/issues/23
https://github.com/owncloud/core/issues/29873

## How Has This Been Tested?
Currently tested it by making a external share called 'Local' and setting the option `read_only` that I made. Haven't figured out how to write proper tests.

capture

capture_2

capture_3

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

 - [ ] Write tests
 - [x] Figure out why `!$storage->instanceOfStorage('\OC\Files\Storage\Home'))` is required in the line `if ($mount->getOption('read_only', true) && !$storage->instanceOfStorage('\OC\Files\Storage\Home'))` on `lib/private/legacy/util.php` line 178

### DIFF
--- a/apps/files_external/appinfo/Migrations/Version20190109104110.php
+++ b/apps/files_external/appinfo/Migrations/Version20190109104110.php
@@ -55,7 +55,7 @@ class Version20190109104110 implements ISimpleMigration {
 			} catch (\Exception $exception) {
 				$out->warning(
 					'There has been an error migrating an external storage from mount.json to the database. The affected mount point is "' .
-					$legacyStorage->getMountPoint() . '" and the type is "' . $legacyStorage->getBackend()->getIdentifier() . '"'
+					$legacyStorage->getMountPoint() . '" and is of type "' . $legacyStorage->getBackend()->getIdentifier() . '"'
 				);
 			}
 		}
--- a/apps/files_external/appinfo/Migrations/Version20190109104110.php
+++ b/apps/files_external/appinfo/Migrations/Version20190109104110.php
@@ -1,0 +1,63 @@
+<?php
+/**
+ * @author Robin Appelman <icewind@owncloud.com>
+ * @author Philipp Schaffrath <pschaffrath@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_External\Migrations;
+
+use OC\Files\External\Service\GlobalStoragesService;
+use OC\Files\External\Service\LegacyStoragesService;
+use OCP\Migration\IOutput;
+use OCP\Migration\ISimpleMigration;
+
+/**
+ * migrate mount.json mounts into the database
+ */
+class Version20190109104110 implements ISimpleMigration {
+
+	/**
+	 * @param IOutput $out
+	 */
+	public function run(IOutput $out) {
+
+		/** @var GlobalStoragesService $globalStoragesService */
+		$globalStoragesService = \OC::$server->query('GlobalStoragesService');
+		$legacyStoragesService = new LegacyStoragesService(\OC::$server->getStoragesBackendService());
+
+		$legacyStorages = $legacyStoragesService->getAllStorages();
+
+		foreach ($legacyStorages as $legacyStorage) {
+			try {
+				$mountOptions = $legacyStorage->getMountOptions();
+				if (!empty($mountOptions) && !isset($mountOptions['read_only'])) {
+					// existing mounts must have read_only disabled by default to avoid surprises
+					$mountOptions['read_only'] = false;
+					$legacyStorage->setMountOptions($mountOptions);
+				}
+				$globalStoragesService->addStorage($legacyStorage);
+			} catch (\Exception $exception) {
+				$out->warning(
+					'There has been an error migrating an external storage from mount.json to the database. The affected mount point is "' .
+					$legacyStorage->getMountPoint() . '" and the type is "' . $legacyStorage->getBackend()->getIdentifier() . '"'
+				);
+			}
+		}
+	}
+}
--- a/apps/files_external/js/settings.js
+++ b/apps/files_external/js/settings.js
@@ -19,7 +19,7 @@ var MOUNT_OPTIONS_DROPDOWN_TEMPLATE =
 	'		<label for="mountOptionsEncrypt">{{t "files_external" "Enable encryption"}}</label>' +
 	'	</div>' +
 	'	<div class="optionRow">' +
-	'		<input id="mountOptionsPreviews" name="read_only" type="checkbox" value="false" checked="checked"/>' +
+	'		<input id="mountOptionsPreviews" name="read_only" type="checkbox" value="true"/>' +
 	'		<label for="mountOptionsPreviews">{{t "files_external" "Set read-only"}}</label>' +
 	'	</div>' +
 	'	<div class="optionRow">' +
--- a/apps/files_external/js/settings.js
+++ b/apps/files_external/js/settings.js
@@ -19,6 +19,10 @@ var MOUNT_OPTIONS_DROPDOWN_TEMPLATE =
 	'		<label for="mountOptionsEncrypt">{{t "files_external" "Enable encryption"}}</label>' +
 	'	</div>' +
 	'	<div class="optionRow">' +
+	'		<input id="mountOptionsPreviews" name="read_only" type="checkbox" value="false" checked="checked"/>' +
+	'		<label for="mountOptionsPreviews">{{t "files_external" "Set read-only"}}</label>' +
+	'	</div>' +
+	'	<div class="optionRow">' +
 	'		<input id="mountOptionsPreviews" name="previews" type="checkbox" value="true" checked="checked"/>' +
 	'		<label for="mountOptionsPreviews">{{t "files_external" "Enable previews"}}</label>' +
 	'	</div>' +
@@ -960,6 +964,7 @@ MountConfigListView.prototype = _.extend({
 			// FIXME default backend mount options
 			$tr.find('input.mountOptions').val(JSON.stringify({
 				'encrypt': true,
+				'read_only': false,
 				'previews': true,
 				'enable_sharing': false,
 				'filesystem_check_changes': 1,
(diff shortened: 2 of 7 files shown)


DeepDiver1975:
@mmattel Hi Martin - we have been chatting about this feature at the conference - right? ;-)

Can you test this? Thanks a lot!

mmattel:
@DeepDiver1975 
hmmm, just pulled but I can´t find ``29873-Read_only_mount`` in origin. did I miss something?

DeepDiver1975:
> @DeepDiver1975
> hmmm, just pulled but I can´t find `29873-Read_only_mount` in origin. did I miss something?

screenshot from 2018-10-06 00-03-29

Step 1 is what you need:
screenshot from 2018-10-06 00-03-13

mmattel:
The first thing I found is the following:

1. when you create a new mount point, the read-only checkmark is not set
**ok**
2. when you look at existing mount points, the check mark ist set to on
you need to manually disable them **wrong**
this must be fixed so that existing mouts do net get set read only automatically
3. depending if the read-only checkmark is set or not, you get the correct message in the files screen ``You don´t have permissions...`` or not. **ok**

@TJHeeringa can you go for a fix for 2

TJHeeringa:
Sure, I can.

Usually a default for existing options would be added by migrations. Since this code does not have migrations, I am unsure how to fix this. I have been looking around and could do either of the following:
 - change the storagewrapper to look for the read only option. If it has not been set, then it should be considered false
 - force the read only button to be unchecked somehow
 --- OR ---
 - make or call a function that reloads the options for all the external mounts

The second option is easier to make, but gives unwanted behavior, since it has to be triggered somehow. The first option on the other hand seems a bit hackish and isn't something that one should do, since then you have the same problem again, when adding another options in the future.

Do you guys now a better solution?

apps/files_external/js/settings.js
```diff
@@ -19,6 +19,10 @@ var MOUNT_OPTIONS_DROPDOWN_TEMPLATE =
 	'		<label for="mountOptionsEncrypt">{{t "files_external" "Enable encryption"}}</label>' +
 	'	</div>' +
 	'	<div class="optionRow">' +
+	'		<input id="mountOptionsPreviews" name="read_only" type="checkbox" value="false" checked="checked"/>' +
```
PVince81: here you need `value="true"` because checking the checkbox will make the value "true"

with a bit of luck it might solve the problem that it's checked by default

lib/private/legacy/util.php
```diff
@@ -173,6 +174,16 @@ public static function setupFS($user = '') {
 			return $storage;
 		});
 
+		\OC\Files\Filesystem::addStorageWrapper('read_only', function ($mountPoint, IStorage $storage, \OCP\Files\Mount\IMountPoint $mount) {
+			if ($mount->getOption('read_only', true) && !$storage->instanceOfStorage('\OC\Files\Storage\Home')) {
```
PVince81: should this be "false" to make it default to false ?

apps/files_external/js/settings.js
TJHeeringa: Sadly this doesn't work with me locally. It still automatically checks it.

TJHeeringa:
Weird issues still persisting:
 - Changing 'mountOptionsPreviews' to 'mountOptionsReadOnly' on lines 22 and 23 in js/settings.js gives the error "PHP Fatal error:  Uncaught Error: Call to a member function getLogger() on null in /var/www/owncloud/core/index.php:86\nStack trace:\n#0 {main}\n  thrown in /var/www/owncloud/core/index.php on line 86"
 - For existing mounts, when clicking on the advanced menu in settings for the first time after the read_only options is added, the checkbox is set, eventhough 'checked="checked"' is not set on line 22 in js/settings.js 

Solutions:
 - The fact that 'id="mountOptionsPreviews"' now appears twice in js/settings.js (22/23 and 26/27) doesn't seem to cause any issue at all. It is not neat programming, but could be ignored.
 - Some logic seems to mishave regarding the checkboxes. The easy fix is to add a migration(contrary to what I said before, there are migrations) in appinfo/migrations, which sets the option read_only on all the existing mounts.

apps/files_external/js/settings.js
```diff
@@ -19,6 +19,10 @@ var MOUNT_OPTIONS_DROPDOWN_TEMPLATE =
 	'		<label for="mountOptionsEncrypt">{{t "files_external" "Enable encryption"}}</label>' +
 	'	</div>' +
 	'	<div class="optionRow">' +
+	'		<input id="mountOptionsPreviews" name="read_only" type="checkbox" value="true"/>' +
```
PVince81: how about calling the option "read_write" instead with default to "true" ?

I'm not sure why the code that is supposed to pre-initialize the checkboxes doesn't work currently

mmattel:
Hi @TJHeeringa , any update on this PR ?

TJHeeringa:
I've had no spare time to work on this PR. In about a week or two I will have some time to continue working on the PR.

apps/files_external/appinfo/Migrations/Version20190109104110.php
```diff
@@ -0,0 +1,63 @@
...
+				if (!empty($mountOptions) && !isset($mountOptions['read_only'])) {
+					// existing mounts must have read_only disabled by default to avoid surprises
+					$mountOptions['read_only'] = false;
+					$legacyStorage->setMountOptions($mountOptions);
+				}
+				$globalStoragesService->addStorage($legacyStorage);
+			} catch (\Exception $exception) {
+				$out->warning(
+					'There has been an error migrating an external storage from mount.json to the database. The affected mount point is "' .
+					$legacyStorage->getMountPoint() . '" and the type is "' . $legacyStorage->getBackend()->getIdentifier() . '"'
```
mmattel: `and the type is` --> `of type`

TJHeeringa: Changed in commit "Fix for tests" below.

TJHeeringa:
The failing test have to do with php-lint. I don't know how to fix this.

phil-davis:
@TJHeeringa please rebase the PR.
The drone jobs matrix was modified today, and drone for "old" PRs will now have the wrong idea about what jobs to run.

TJHeeringa:
Codecov.io keeps giving secure connection failed errors. Do I need to fix more things?

phil-davis:
@TJHeeringa codecov can be a pain. It might just be having a bad day.

@PVince81 drone is passing. Please review again.

apps/files_external/js/settings.js
TJHeeringa: This has been resolved by adding an migration that initializes all existing mounts with read_only = false.

mmattel:
@PVince81 any update ?

PVince81:
@jvillafanez please test and review

jvillafanez:
Is the migration needed?
* For new installations, the "set read-only" checkbox should be unchecked / false
* For upgrades from a recent version (no legacy storages), the checkbox should also be set as unchecked / false to keep the previous behaviour
* For upgrades from older versions (with legacy storages), I expect another migration to take care of the conversion (which should be already in place, so no need to touch anything here), and then consider the checkbox as false.

I expect that, if the option is missing from the storage information, it's considered as false. The checkbox is expected to be shown unchecked in any case, and modifying its state should store the proper information in the DB as expected.

Unless I'm missing something, the migration should be removed. The rest of the code looks good. I still need to test it though.

TJHeeringa:
> 
> 
> Is the migration needed?
> 
>     * For new installations, the "set read-only" checkbox should be unchecked / false
> 
>     * For upgrades from a recent version (no legacy storages), the checkbox should also be set as unchecked / false to keep the previous behaviour
> 
>     * For upgrades from older versions (with legacy storages), I expect another migration to take care of the conversion (which should be already in place, so no need to touch anything here), and then consider the checkbox as false.
> 
> 
> I expect that, if the option is missing from the storage information, it's considered as false. The checkbox is expected to be shown unchecked in any case, and modifying its state should store the proper information in the DB as expected.
> 
> Unless I'm missing something, the migration should be removed. The rest of the code looks good. I still need to test it though.

I also thought that the migration wasn't needed, but for some weird reason the code, regardless of what I set as default value, always made the existing mounts behave like read_only = true. This is unwanted behavior and the migration seems to fix this issue.

A test from your end without the migration would be helpful. If it works without, the migration can be removed.

TJHeeringa:
I just tried doing the checkout 'trick', but my OwnCloud won't sign in anymore after displaying a message about upgrading to 11.0 pre-alpha. Will probably take a while to figuring out what is broken. Probably some permission with the session cookie location. Error is probably occuring due to that I'm running WSL, but that's off topic for this PR.

@mmattel Could you perhaps try and see if it works without the migration?

mmattel:
@TJHeeringa to "fix" you login problem (I ran into the same ...), just run `make clean` and then `make` in the root of core. That should fix it. For details run `make help` which gives you quite useful infos.
If not done already, run `make test-php-style-fix` to make the tests happy :smile:

I am currently very short in time, so testing is an issue for me...

jvillafanez:
I suggest you to use either docker or a VM in order to revert to a previous state easily.

> 1. Write the code -- equivalent for someone else to just pulling the code
> 2. Make a new mount by inserting directly into the database -- equivalent to having a entry in your database prior to pulling
> 3. Access the mount point
> 4. Conclude that it is read_only
> 5. Apply migration
> 6. Access mount point again
> 7. Conclude that it is not read_only anymore

I don't think the steps are good.

Assuming you wrote the same code (which let me distrust this because people always make stupid mistakes), then adding a mount point directly in the DB is bad: there is no need for this when you can easily do the same in the UI; you can also add wrong or incompleted information in the DB.

You also apply the migration in step 5. Under normal circumstances, migrations should be executed just after apply the code patch, without any other interaction. As far as I know, apps increases the version which should cause core to automatically check and apply new migrations; I'm not sure how this process is done for core migrations, but in any case I'd expect the migration to be run manually just after the patch.

Note that what we want (if possible) is to remove the migration file, that "apply migration" step shouldn't be there in order to test what we want.

As said, since downgrading isn't supported, I'd recommend you to use either docker or a VM to restore a previous state. It might also happen that you had some additional information somewhere that you forgot (or maybe you didn't know it was there) to cleanup before trying to restore the initial information.

mmattel:
Hi, any update on this? Would be very useful !

DeepDiver1975:

PR #33067, opened 2018-10-08
Relax tests for page title so they only check 'StringStartsWith'

phil-davis:
## Description
Relax the webUI tests so they just check that the page title starts with the expected text.

## Motivation and Context
Some webUI tests check that the user has "arrived" at the correct page by checking for the expected page title in the browser. In some cases with theming there could be extra things after the page name and product name, e.g. ``Files - ownCloud`` - ``ownCloud`` can be branded and also have other things after it.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33068, opened 2018-10-08
[stable10] Relax tests for page title so they only check 'StringStartsWith'

phil-davis:
Backport #33067


patrickjahns (approved):
Seems plausible to me

PR #33071, opened 2018-10-08
Revert "Revert "preLogin hook should use string UID not IUser""

PVince81:
Had a chat with @tomneedham and the actual workflow of this emitter is as follows:

1) the first time it had string UID
2) later on, it got changed to IUser by mistake when moving to Symfony. This got released in some version.
3) then @tomneedham noticed that and provided this PR to revert this back to stage 1)

When looking at this I wasn't aware that there was a time where the event was correct in the past.
Considering this, I think we can accept this change again as it will bring it back to its original intended state.

@sharidas can you do some research to identify what versions the stages 1) and 2) were in to confirm ?

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -986,7 +986,7 @@ class Session implements IUserSession, Emitter {
 				return false;
 			}
 
-			$this->manager->emit('\OC\User', 'preLogin', [$user, $password]);
+			$this->manager->emit('\OC\User', 'preLogin', [$user->getUID(), $password]);
 
 			if (!$user->isEnabled()) {
 				$message = \OC::$server->getL10N('lib')->t('User disabled');
--- a/tests/lib/User/SessionTest.php
+++ b/tests/lib/User/SessionTest.php
@@ -1582,8 +1582,7 @@ class SessionTest extends TestCase {
 		$iUser->expects($this->once())
 			->method('isEnabled')
 			->willReturn(false);
-		$iUser->expects($this->exactly(2))
-			->method('getUID')
+		$iUser->method('getUID')
 			->willReturn('foo');
 
 		$failedEvent = new GenericEvent(null, ['user' => 'foo']);


sharidas:
My initial investigation ( with `git log lib/private/User/Session.php` on `stable10` branch ) are as follows:

1. d35a58df9479ca9249ca36d6dec96b7dbef059f8 was the commit which brought the change ->
private function loginUser($user, $password) {
This had `$this->manager->emit('\OC\User', 'preLogin', [$user, $password]);`. This change is before symfony changes. And as per the function signature $user is of type IUser
2. 74f8be86d4bbba9b3e6137089aa2fea1fde4dfe7 was the commit which added symfony event for the change -> 
`private function loginUser($user, $password)`
This change moved the code moves the code in the method inside closure. i.e,
`return $this->emittingCall(function () use (&$user, &$password)`  ....
The `$user` is still instance of `IUser`
3. 1d4282bfb53d519c5df3a3f6a74612f7f9888145 was the commit which changed the access specifier for the method
`protected function loginUser($user, $password)`
Nothing else was touched here...
4. 354377da47edfc6cbfc3d41e3cafff71aaa367bf was the commit which added the failed login for the method. But this never touched the $user
5. 621a64774b5d73be2a65daac94e96a8480dd261e was the commit which made small modification to the funcation arg
`protected function loginUser(IUser $user = null, $password)`
It did added `$uid = $user === null ? '' : $user->getUID();` as the first line in the method

I will have a re-look to make sure which commit brought uid as first change and then shift to IUser. Based on my above observation, it does not look so. From the first time the method was introduced `$user` was an instance of `IUser`.

sharidas:
1. f10d105ea2f2d0822693313ddcd6f9b2eaac6ebe is the commit which had 
`private function loginWithPassword($uid, $password) {
        $this->manager->emit('\OC\User', 'preLogin', [$uid, $password]);`
This is uid.
2. d35a58df94 is the commit which had
`private function loginWithPassword($uid, $password) {
        $this->manager->emit('\OC\User', 'preLogin', [$uid, $password]);`
The uid is used here... and
`private function loginWithToken($token) {
        ...
        $this->manager->emit('\OC\User', 'preLogin', [$uid, $password]);`
The uid is here, and
`private function loginUser($user, $password) {
         ...
         $this->manager->emit('\OC\User', 'preLogin', [$user, $password]);`
3.74f8be86d4bbba9b3e6137089aa2fea1fde4dfe7 is the commit which brought the symfony events to Session. But the preLogin has the same 3 change here too

The commit which brought `IUser` for `preLogin` event instead of `uid` is d35a58df9479ca9249ca36d6dec96b7dbef059f8.

And from then on, this change was there with the new method added `private function loginUser`.
So the change with IUser is there from `v10.0.0` version.

sharidas (approved):
From my observations shared https://github.com/owncloud/core/pull/33071#issuecomment-428207044 this change LGTM.

tomneedham:
So this changed from uid to IUser with ownCloud 10.0.0.

sharidas:
> So this changed from uid to IUser with ownCloud 10.0.0.

Yes

sharidas:
Backport to stable10 -> https://github.com/owncloud/core/pull/33185

PR #33135, opened 2018-10-10
[stable10] Replace productname in expected email content

phil-davis:
## Description
In ``EmailContext`` ``assertThatEmailContains()`` allow the product name to be substituted at run-time.

## Motivation and Context
Remove hard-coded"ownCloud" ``productname`` from acceptance tests.

## How Has This Been Tested?
Local run of the webUI acceptance test.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33140, opened 2018-10-10
Replace productname in expected email content

phil-davis:
Forward port of #33135 

This is the code changes, which stay in core.

The feature file scenario change that uses this code will be forward-ported from core stable10 to user_management app master.


patrickjahns (approved):
LGTM

Issue #33161, opened 2018-10-11
Acceptance test OwncloudPage and waitTillPageIsLoaded


phil-davis:
See PR #33158 and #33149 for examples and discussion.

OwncloudPage has a ``waitTillPageIsLoaded()`` method that looks for a ``.loading`` indicator to become invisible. On ordinary files pages there is no such indicator, and FilesPageBasic extends OwncloudPage and implements a ``waitTillPageIsLoaded()`` that works on files-type-pages.

Some other pages directly extend OwncloudPage and so get the "base" ``waitTillPageIsLoaded()``

When the comments app is enabled, the comments panel happens to have a ``.loading`` indicator - so the "base" ``waitTillPageIsLoaded()`` just happens to "work" by checking that page element.

- [ ] review each page object that extends OwncloudPage and ensure that either it implements its own appropriate ``waitTillPageIsLoaded()`` or that the base one from OwncloudPage is relevant.

phil-davis:
Try running the acceptance test suite with:
- [ ] comments app disabled
- [ ] trashbin app disabled
- [ ] versions app disabled

and see what bad things happen.

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/22485 (test), https://github.com/owncloud/core/issues/22492 (test), https://github.com/owncloud/core/issues/31579 (test), https://github.com/owncloud/core/pull/32559 (acceptance tests - stop uploading when receiving an error), and https://github.com/owncloud/core/pull/31516 (Acceptance test coding standard changes).

PR #33162, opened 2018-10-11
[PoC] Test with trashbin app disabled

phil-davis:
- disable the files_trashbin app
- run the acceptance tests that do not test trashbin app behavior
- see what happens

Part of issue #33161


phil-davis:
works

PR #33163, opened 2018-10-11
[PoC] Test with versions app disabled

phil-davis:
- disable the files_versions app
- run the acceptance tests that do not test versions app behavior
- see what happens

Part of issue #33161


phil-davis:
works

PR #33302, opened 2018-10-25
do not iterate over all storages when you only need a single node anyway

butonic:
Today, the node api currently iterates over all storages when resolving a node by id to return all possible paths to the file. Even when it has found a path it continues to iterate over the other storages in order to find all nodes in the users tree.

That causes *some* overhead.

Actually, we never really use anything but the first node ... so this pr adds an optional parameter to only return the first node that is found. as you can see, quite a few places can profit.

Furthermore, the Jail Wrapper needs to intercept the IVersionedStorage calls, so it can adjust the paths correctly. However it should not implement the interface AFAICT because the wrapped storages might not actually implement versioning. I may get the logic in instanceOfStorage() wrong but this can be tested.

Finally, the Meta endpoint iterates over all possible storages, because it does not yet use the LazyRootFolder ...

All of this together **significantly** improves versions performance.

I won't start on the tests until I get a response on the approach. Especially, because I don't inject the UserFolder but fetch if from the $server.

@DeepDiver1975 did you intend the Meta endpoint to be able to access any fileid? then, limiting the versioning meta nodes to versions might not be the right approach. We might have to query the db table directly to immediately lookup storage and path ...


apps/federatedfilesharing/lib/FederatedShareProvider.php
```diff
@@ -886,13 +886,13 @@ private function getNode($userId, $id) {
...
 
-		$nodes = $userFolder->getById($id);
+		$nodes = $userFolder->getById($id, true);
 
 		if (empty($nodes)) {
 			throw new InvalidShare();
 		}
 
-		return $nodes[0];
+		return \array_shift($nodes);
```
DeepDiver1975: Why? calling a built in function is more expensive then staying in the language

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -67,21 +67,35 @@ public function isShared() {
 	}
 
 	/**
-	 * @inheritdoc
+	 * resolve the file node
+	 * returns only the first node because we only really need the id anyway
 	 */
+	private function getNode() {
+		$folder = \OC::$server->getUserFolder();
```
DeepDiver1975: please inject if possible - yes - my view should have been injected as well ....

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -67,21 +67,35 @@ public function isShared() {
...
+		$nodes = $folder->getById($this->fileId, true);
+		if (count($nodes) === 0) {
+			return null;
+		}
+		return array_shift($nodes);
```
DeepDiver1975: missing \

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -67,21 +67,35 @@ public function isShared() {
...
+	}
+
 	public function getDirectoryListing() {
-		$view = new View();
-		$path = $view->getPath($this->fileId, false);
-		/** @var Storage $storage */
-		list($storage, $internalPath) = $view->resolvePath($path);
```
DeepDiver1975: afaik resolving the path via the view is necessary here because of the way how files_versions is implemented .....

lib/private/Files/Storage/Wrapper/Jail.php
```diff
@@ -26,14 +26,15 @@
...
+use OCP\Files\Storage\IVersionedStorage;
 use OCP\Lock\ILockingProvider;
 
 /**
  * Jail to a subdirectory of the wrapped storage
  *
  * This restricts access to a subfolder of the wrapped storage with the subfolder becoming the root folder new storage
  */
-class Jail extends Wrapper {
+class Jail extends Wrapper /* implements IVersionedStorage */ {
```
DeepDiver1975: ??????

lib/private/Files/Storage/Wrapper/Jail.php
```diff
@@ -497,4 +498,62 @@ public function moveFromStorage(\OCP\Files\Storage $sourceStorage, $sourceIntern
...
+
+	/**
+	 * Get the content of a given version of a given file as stream resource
+	 *
+	 * @param string $internalPath
+	 * @param string $versionId
+	 * @return resource
+	 * @since 10.0.9
+	 */
+	public function getContentOfVersion($internalPath, $versionId) {
```
DeepDiver1975: funny ..... was that missed somehow?

DeepDiver1975 (dismissed):

apps/federatedfilesharing/lib/FederatedShareProvider.php
PVince81: indeed, this would not only return the first element but also need to update the array

lib/private/Files/Meta/MetaVersionCollection.php
butonic: because it uses the internal path `/username/file_versions`? but we are resolving the node / versions we want to access based on the fileid anyway. and a user can only access versions to files he has access to ... so I think using the node api/user folder makes sense.

apps/federatedfilesharing/lib/FederatedShareProvider.php
butonic: ack .... wanted to unify this and several places used this. will change to array access

DeepDiver1975:
@butonic drone is red - please have a look as well - THX

butonic:
I used `$node = $nodes[0] ?? null` where it makes sense. checking drone after the build

lib/private/Files/Storage/Wrapper/Jail.php
butonic: see https://github.com/owncloud/core/pull/33302#issue-225757936
> Furthermore, the Jail Wrapper needs to intercept the IVersionedStorage calls, so it can adjust the paths correctly. However it should not implement the interface AFAICT because the wrapped storages might not actually implement versioning. I may get the logic in instanceOfStorage() wrong but this can be tested.

butonic:
I wanted to inject the UserFolder ... but .. besides that requiring a lot of changes ... urgh ... the meta endpoint is hopelessly inefficient. For every request to the versions endpoint
1. [the MetaRootNode does a node lookup by id](https://github.com/owncloud/core/blob/88d9dff7652478e47eb5c4ca388aa46129446ff9/lib/private/Files/Meta/MetaRootNode.php#L80)
2. [the MetaVersionCollection does a lookup by id](https://github.com/owncloud/core/blob/88d9dff7652478e47eb5c4ca388aa46129446ff9/lib/private/Files/Meta/MetaVersionCollection.php#L75)

As a result the same file id is looked up twice for each version AND twice for every thumbnail.

But there is no caching whatsoever ... 

This calls for cacheception: https://github.com/owncloud/core/pull/28166 With that all fileids would have been cached when listing the files while browsing to it ...

DeepDiver1975:
> I used `$node = $nodes[0] ?? null` where it makes sense. checking drone after the build

this is okay for master (php7.1+) - but needs to be adapted in the backport

apps/comments/lib/Activity/Listener.php
```diff
@@ -86,10 +86,9 @@ public function commentEvent(CommentsEvent $event) {
 		foreach ($mounts as $mount) {
 			$owner = $mount->getUser()->getUID();
 			$ownerFolder = $this->rootFolder->getUserFolder($owner);
-			$nodes = $ownerFolder->getById($event->getComment()->getObjectId());
-			if (!empty($nodes)) {
-				/** @var Node $node */
-				$node = \array_shift($nodes);
+			$nodes = $ownerFolder->getById($event->getComment()->getObjectId(), true);
+			$node = $nodes[0] ?? null;
```
mrow4a: What is this construct doing?

lib/public/Files/Folder.php
```diff
@@ -145,10 +145,11 @@ public function searchByTag($tag, $userId);
 	 * get a file or folder inside the folder by it's internal id
 	 *
 	 * @param int $id
+	 * @param boolean $first only return the first node that is found
 	 * @return \OCP\Files\Node[]
 	 * @since 6.0.0
 	 */
-	public function getById($id);
+	public function getById($id, $first = false);
```
mrow4a: Why first would be more important than second or third? Isn't this interface quite ambigous to implement? What should be returned first, can it be in random order?

lib/private/Files/Storage/Wrapper/Jail.php
```diff
@@ -26,14 +26,16 @@
...
+use OCP\Files\Storage\IVersionedStorage;
 use OCP\Lock\ILockingProvider;
 
 /**
  * Jail to a subdirectory of the wrapped storage
  *
  * This restricts access to a subfolder of the wrapped storage with the subfolder becoming the root folder new storage
  */
-class Jail extends Wrapper {
+class Jail extends Wrapper /* implements IVersionedStorage */
```
mrow4a: Why is it commented out?

lib/private/Files/Meta/MetaVersionCollection.php
mrow4a: Wont we break some primary storages like (s3) with this? What is the difference between Node API and View API?

apps/comments/appinfo/app.php
```diff
@@ -54,7 +54,7 @@ function () {
 
 $eventDispatcher->addListener(\OCP\Comments\CommentsEntityEvent::EVENT_ENTITY, function (\OCP\Comments\CommentsEntityEvent $event) {
 	$event->addEntityCollection('files', function ($name) {
-		$nodes = \OC::$server->getUserFolder()->getById(\intval($name));
+		$nodes = \OC::$server->getUserFolder()->getById((int)$name, true);
```
mrow4a: why this function sometimes throws an exception NotFound, and sometimes just empty array? I am lost

lib/private/Files/Storage/Wrapper/Jail.php
mrow4a: So it implements IVersionedStorage or not?

lib/public/Files/Folder.php
mrow4a: Why would someone just want first result? What is this meaning `get a file or folder inside the folder by it's internal id` and why it returns Node[] ?

mrow4a (commented):
My worry here is to also not break any IVersionedStorage functionality, meaning versioning for s3. I would also look into adding new function like `getById()` doing proper thing, instead of adding additional ambiguous flag.

apps/dav/lib/RootCollection.php
```diff
@@ -109,7 +109,7 @@ public function __construct() {
 				$systemTagRelationsCollection,
 				$uploadCollection,
 				$avatarCollection,
-				new \OCA\DAV\Meta\RootCollection(\OC::$server->getRootFolder()),
+				new \OCA\DAV\Meta\RootCollection(\OC::$server->getLazyRootFolder()),
```
mrow4a: What is the difference between LazyRootFoder and RootFolder? Why would sometimes one be prefered over the other? Why would sometimes someone pick just RootFolder? (thinking why it was choosed to be RootFolder in first place)

lib/private/Files/Meta/MetaFileIdNode.php
```diff
@@ -34,8 +34,8 @@
  */
 class MetaFileIdNode extends AbstractFolder {
```
mrow4a: It no longer uses FileId, but Node. Name

lib/private/Files/Meta/MetaVersionCollection.php
butonic: the node api currently works on top of the view api

apps/comments/lib/Activity/Listener.php
butonic: This is the coalesce operator, it suppresses the E_NOTICE error when accessing a non existing array key:
```php
$username = $_GET['user'] ?? 'nobody';
// equivalent to: $username = isset($_GET['user']) ? $_GET['user'] : 'nobody';
```
see
https://stackoverflow.com/a/26142784

lib/public/Files/Folder.php
butonic: look at the code ... there is literally NO place where we iterate over the results in core. All occurrences just take the first node that is found. That is why allowing getById to return on the first node that is found makes sense.

apps/comments/appinfo/app.php
butonic: not related to this pr

apps/dav/lib/RootCollection.php
butonic: It should always be the LazyRootFolder so the initialization happens only when necessary. If the initialization happens too early, the user might not have been logged in, causing the filesystem layer to mount all mountpoints.

lib/private/Files/Meta/MetaFileIdNode.php
butonic: It represents a fileid in the path, eg `/meta/123`. how it internally references the fileid does not determine the class name, as far as I am concerned.

PVince81:
@mrow4a please click "resolve conversation" after aknowledging @butonic's comments

mrow4a:
@PVince81 done

PVince81:
any update ? @butonic

PVince81:
@butonic please finish this or hand over

lib/public/Files/Folder.php
PVince81: as far as I remember you can get multiple entries whenever multiple homes are mounted and you call `getById()` on the root folder. The entries that are not the first one are usually the ones visible in received shares of other users.

It can also happen if you receive the same folder through multiple different shares.

lib/private/Files/Config/CachedMountInfo.php
```diff
@@ -91,12 +91,8 @@ public function getMountPointNode() {
...
 		Filesystem::initMountPoints($this->getUser()->getUID());
 		$userNode = \OC::$server->getUserFolder($this->getUser()->getUID());
-		$nodes = $userNode->getParent()->getById($this->getRootId());
-		if (\count($nodes) > 0) {
-			return $nodes[0];
-		} else {
-			return null;
-		}
+		$nodes = $userNode->getParent()->getById($this->getRootId(), true);
+		return $nodes[0] ?? null;
```
PVince81: note to self: will clash on backport due to PHP 5.6 compat

lib/public/Files/Folder.php
PVince81: The first one is used anyway everywhere so this does not change behavior and we want to get this merged. We can raise the question again separately.

lib/private/Files/Storage/Wrapper/Jail.php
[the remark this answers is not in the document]
PVince81: this is the unfortunately reality of the wrapper pattern

PVince81:
@ownclouders rebase

ownclouders:

ownclouders:

PVince81:
uh oh, I hope we didn't break something:
```
/drone/src/tests/acceptance/features/apiMain/fileVersions.feature:76
```

- [ ] will dig into it later

PVince81:
blocked by https://github.com/owncloud/core/issues/33792

ownclouders:

ownclouders:

PVince81:
uh oh
```
    /drone/src/tests/acceptance/features/apiMain/fileVersions.feature:75
/drone/src/tests/acceptance/features/apiMain/fileVersions.feature:76
```

@individual-it

phil-davis:
@ownclouders rebase

ownclouders:

ownclouders:

apps/comments/appinfo/app.php
PVince81: not covered in code coverage

PVince81: not covered in code coverage

apps/comments/lib/Activity/Listener.php
```diff
@@ -86,10 +86,9 @@ public function commentEvent(CommentsEvent $event) {
 		foreach ($mounts as $mount) {
 			$owner = $mount->getUser()->getUID();
 			$ownerFolder = $this->rootFolder->getUserFolder($owner);
-			$nodes = $ownerFolder->getById($event->getComment()->getObjectId());
-			if (!empty($nodes)) {
-				/** @var Node $node */
-				$node = \array_shift($nodes);
+			$nodes = $ownerFolder->getById($event->getComment()->getObjectId(), true);
```
PVince81: none covered in code coverage

apps/federatedfilesharing/lib/FederatedShareProvider.php
```diff
@@ -886,7 +886,7 @@ private function getNode($userId, $id) {
 			throw new InvalidShare();
 		}
 
-		$nodes = $userFolder->getById($id);
+		$nodes = $userFolder->getById($id, true);
```
PVince81: not covered...

apps/files/lib/ActivityHelper.php
```diff
@@ -59,10 +59,9 @@ public function getFavoriteFilePaths($user) {
 		$rootFolder = \OC::$server->getUserFolder($user);
 		$folders = $items = [];
 		foreach ($favorites as $favorite) {
-			$nodes = $rootFolder->getById($favorite);
-			if (!empty($nodes)) {
-				/** @var \OCP\Files\Node $node */
-				$node = \array_shift($nodes);
+			$nodes = $rootFolder->getById($favorite, true);
```
PVince81: not covered

apps/systemtags/lib/Activity/Listener.php
```diff
@@ -159,10 +159,9 @@ public function mapperEvent(MapperEvent $event) {
 		foreach ($mounts as $mount) {
 			$owner = $mount->getUser()->getUID();
 			$ownerFolder = $this->rootFolder->getUserFolder($owner);
-			$nodes = $ownerFolder->getById($event->getObjectId());
-			if (!empty($nodes)) {
-				/** @var Node $node */
-				$node = \array_shift($nodes);
+			$nodes = $ownerFolder->getById($event->getObjectId(), true);
```
PVince81: not covered

lib/private/Files/Config/CachedMountInfo.php
```diff
@@ -91,12 +91,8 @@ public function getMountPointNode() {
 		// TODO injection etc
 		Filesystem::initMountPoints($this->getUser()->getUID());
 		$userNode = \OC::$server->getUserFolder($this->getUser()->getUID());
-		$nodes = $userNode->getParent()->getById($this->getRootId());
-		if (\count($nodes) > 0) {
-			return $nodes[0];
-		} else {
-			return null;
-		}
+		$nodes = $userNode->getParent()->getById($this->getRootId(), true);
```
PVince81: not covered

lib/private/Files/Meta/MetaRootNode.php
```diff
@@ -92,7 +92,7 @@ public function get($path) {
 	/**
 	 * @inheritdoc
 	 */
-	public function getById($id) {
+	public function getById($id, $first = false) {
```
PVince81: method not covered

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -66,22 +66,19 @@ public function isShared() {
...
 		if (!$storage->instanceOfStorage(IVersionedStorage::class)) {
 			return [];
 		}
 		/** @var IVersionedStorage | Storage $storage */
 		$versions = $storage->getVersions($internalPath);
-		return \array_values(\array_map(function ($version) use ($storage, $internalPath, $view, $path) {
+		return \array_values(\array_map(function ($version) use ($storage, $node, $internalPath) {
 			if (!isset($version['mimetype'])) {
-				$version['mimetype'] = $view->getMimeType($path);
+				$version['mimetype'] = $node->getMimetype();
```
PVince81: this specific line (condition) is not covered, the rest of the return block is.

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -110,7 +107,7 @@ public function get($path) {
 			throw new NotFoundException();
 		}
 		if (!isset($version['mimetype'])) {
-			$version['mimetype'] = $view->getMimeType($path);
+			$version['mimetype'] = $this->node->getMimetype();
```
PVince81: line not covered

lib/private/Files/Node/LazyRoot.php
```diff
@@ -417,7 +417,7 @@ public function searchByTag($tag, $userId) {
 	/**
 	 * @inheritDoc
 	 */
-	public function getById($id) {
+	public function getById($id, $first = false) {
```
PVince81: the method is not covered

lib/private/Files/Node/NonExistingFolder.php
```diff
@@ -154,7 +154,7 @@ public function searchByTag($tag, $userId) {
 		throw new NotFoundException();
 	}
 
-	public function getById($id) {
+	public function getById($id, $first = false) {
```
PVince81: method not covered

lib/private/Files/Storage/Wrapper/Jail.php
```diff
@@ -497,4 +499,62 @@ public function moveFromStorage(\OCP\Files\Storage $sourceStorage, $sourceIntern
...
+	/**
+	 * Get the content of a given version of a given file as stream resource
+	 *
+	 * @param string $internalPath
+	 * @param string $versionId
+	 * @return resource
+	 * @since 10.0.9
+	 */
+	public function getContentOfVersion($internalPath, $versionId) {
+		return $this->getWrapperStorage()->getContentOfVersion($this->getSourcePath($internalPath), $versionId);
```
PVince81: none of the new code is covered, this includes the other methods below as well (note: my test run was with sqlite with local storage, maybe this code would be covered with object store mode ?)

PVince81 (commented):
since codecov reports can't be loaded, I went through code coverage locally and examined the diff manually.

the coverage file I had was from a local test with SQLite + local storage, so not sure if some code paths would have been covered by object store.

it seems that a lot of the touched code did not have coverage in the first place

lib/private/Files/Config/CachedMountInfo.php
PVince81: seems this method `getMountPointNode()` is never called anywhere in our codebase...

PVince81: I digged through git history and did not find any usage for this method. Thank you, over-engineers

PVince81 (commented):
why can't I comment on my own review

PVince81:
So... I spent a few hours adding coverage where coverage was missing and it wasn't easy and my head hurts from too many mocks...

Let's hope this will bring the coverage to an acceptable level for merging.

See my comments on the remaining items, I'm not sure if covering those is worth it.

apps/comments/appinfo/app.php
PVince81: covered now

apps/comments/lib/Activity/Listener.php
PVince81: covered now

apps/federatedfilesharing/lib/FederatedShareProvider.php
PVince81: covered now

apps/files/lib/ActivityHelper.php
PVince81: covered now

apps/systemtags/lib/Activity/Listener.php
PVince81: covered now

PVince81:
I'll need a review on the coverage commit: https://github.com/owncloud/core/pull/33302/commits/5e6e4451d3b82f39315b9950e03130248350f362 @butonic @DeepDiver1975

butonic:
@PVince81 omg I love you!

PVince81:
fixed code style issues and rebased

PVince81:
Had to delete again the test for comments/appinfo/app.php uncovered code, because it's unreliable as its state depend on other tests. A proper way would be to first refactor app.php into an Application class and then test the latter. Something which I do not want to start in this PR as it would require regression testing things unrelated to what this PR fixes...

PVince81:
I've added even more tests, some covering the issues in this PR, I hope this will make us reach the necessary threshold...

lib/private/Files/Meta/MetaRootNode.php
PVince81: covered now

lib/private/Files/Meta/MetaVersionCollection.php
PVince81: covered now

lib/private/Files/Meta/MetaVersionCollection.php
PVince81: covered now, I made a new class to test some of the more rare conditions...

lib/private/Files/Storage/Wrapper/Jail.php
PVince81: covered now

PVince81:
stable10: https://github.com/owncloud/core/pull/33859

PVince81:
it's the second time this new test fails on object store:
```
1) OCA\SystemTags\Tests\unit\ActivityListenerTest::testActivityOnMapperEvent with data set #1 ('OCP\SystemTag\ISystemTagObjec...gnTags', 'unassign_tag')
OCP\Files\StorageNotAvailableException: 

/drone/src/lib/private/Files/Storage/Wrapper/Availability.php:76
/drone/src/lib/private/Files/Storage/Wrapper/Availability.php:239
/drone/src/lib/private/Files/Storage/Wrapper/Wrapper.php:221
/drone/src/lib/private/Files/View.php:1363
/drone/src/lib/private/Files/View.php:1409
/drone/src/lib/private/Share/Share.php:122
/drone/src/lib/public/Share.php:77
/drone/src/apps/systemtags/lib/Activity/Listener.php:170
/drone/src/apps/systemtags/tests/unit/ActivityListenerTest.php:195
```

Sadly there is a non-mockable static method in there which is why I had to use an actual user/storage. No idea why the storage would not be available for object store.

PVince81:
- [ ] since https://github.com/owncloud/core/pull/33862 got merged with a better test file name, we should align this PR to extend the tests that were added there.

but first debug the above problem...

PVince81:
- [ ] keep backport in sync

PVince81:
I managed to get something close to the object store issue. It turns out that `Share::getUsersSharingFile()` is calling `initMountPoints()` which itself would attempt to setup the object store in some inconsistent way. Instead of digging further I decided to mock the former method by wrapping it in an instance method in the activity `Listener` class, which is cleaner than trying to meddle with the setup.

The tests should all pass now.

PVince81:
all green, please review the tests now @DeepDiver1975 @butonic

tomneedham (approved):

PVince81:
this now triggered unit test failures in storage-related apps:
- https://github.com/owncloud/workflow/issues/169
- https://github.com/owncloud/firewall/issues/554

will look into this...

PVince81:
ok, the issue is simply that the unit tests in those apps are using `returnValueMap` for the `getById` method, so I needed to add the extra `false` argument to the PHPUnit matcher.

phil-davis:
Backport `stable10` is PR #33859

PR #33340, opened 2018-10-29
Mention bot is dead ....

DeepDiver1975:


PVince81 (approved):
:+1: right, because now Github makes suggestions for reviews and for tickets it's Gitmate taking care

PR #33138, opened 2018-10-10
Email subject is already escaped, avoid double escaping

tomneedham:
## Description
Otherwise file names are double escaped in the email subject.

## Motivation and Context
Annoying double escaped email subjects with filesnames.

## How Has This Been Tested?
Create a file called `!@£$%^&*()_.txt` and share it, notify by email. Check escaping in email.

## Screenshots (if appropriate):
Before: 
screenshot 2018-10-10 at 15 14 09

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33342, opened 2018-10-29
[stable10] Email subject is already escaped, avoid double escaping

tomneedham:
Backport of https://github.com/owncloud/core/pull/33138 to stable10


PVince81 (approved):
:+1:

PR #27051, opened 2017-01-30
Always enable files_external

PVince81:
## Motivation and Context
Because we want the UI to still be available while the backend is
already in core.

Eventually the files_external backends will be moved to separate apps and all that will be left is the GUI and the OCC commands which we'll eventually move to core. Ref: https://github.com/owncloud/core/issues/27050

## How Has This Been Tested?
- [x] TEST: update from 9.1 with files_external disabled to master: app is now enabled
- [x] TEST: setup OC 10.0 from scratch, app is enabled
- [x] TEST: cannot disable files_external app

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Please review @pmaier1 @DeepDiver1975 as discussed

Issue #15277, opened 2015-03-27
files_external app suppresses remember log in?


jancborchardt:
For the encryption app I know that »remember me« is not shown. But it seems that the files_external app also does it (see the `<rememberlogin>` in the info.xml).

@LukasReschke what’s the reason here / is this needed? Just asking. :)

Xenopathic:
We sometimes need the password for SMB_OC to mount the share. It's the only storage using it at the moment, but conceivably more could use it in the future.

DeepDiver1975:
Some backends rely on the password being stored in the session and there fore we require login.

jancborchardt:
Ok, thanks for the clarification. Then at least we have this on record somewhere.

PR #33355, opened 2018-10-30
Allow remember login when external storage is disabled globally

okkez:
#27051 and #15277 disable the feature "remember login". This means that `\OC_Util::rememberLoginAllowed()` returns `false` always. We can enable it when external storage is disabled globally.

### DIFF
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -1042,11 +1042,11 @@ class OC_Util {
 
 		foreach ($apps as $app) {
 			$appInfo = OC_App::getAppInfo($app);
-			if (isset($appInfo['rememberlogin']) && $appInfo['rememberlogin'] === 'false') {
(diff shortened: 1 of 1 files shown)


okkez:
image

This PR can show "stay logged in" checkbox when external storage is disabled globally.

PVince81:
thanks, but stay logged in is currently broken: https://github.com/owncloud/core/issues/26027

also there was a reason for disabling when external storage is enabled but can't remember right now

PVince81:
the reason was because of "save in session" and "save in DB" mode, in which case we need the user to login with actual credentials to be able to save these for external storage access.

the remember me thing (if it wasn't broken as per https://github.com/owncloud/core/issues/26027) might work because the user had to login at least once. however if external storages get configured by the admin, the admin would require those users to manually relogin with their password.

PR #33367, opened 2018-10-31
fix upload avatar for LDAP users

davitol:
## Description

Readd ```id="language"``` to the form

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/2553#issuecomment-434390467

## Motivation and Context

LDAP users can upload an avatar via webUI

## How Has This Been Tested?

Locally applying this PR and testing upload avatar button works properly

### DIFF
--- a/settings/templates/panels/personal/profile.php
+++ b/settings/templates/panels/personal/profile.php
@@ -131,7 +131,7 @@ if ($_['passwordChangeSupported']) {
 <?php
 	}
 ?>
-<form class="section">
+<form id="language" class="section">
 	<h2>
 		<label><?php p($l->t('Language'));?></label>
 	</h2>


PVince81:
Ah, it's that easy ? nice!

PVince81 (approved):
:+1:

PVince81:
@VicDeo do you have more background on this fix ? just wondering if we should cover this with acceptance tests as the prerequisite for reproduction are rather complex

VicDeo:
@PVince81 background: the avatar section is outside the normal page flow.
the ldap app hides several sections at the personal page. 
lang section sticks to the right  is pushed towards the previous block as an avatar section.

ids are used to add additional 'min-width:60%;display:inline-block' styles so the lang selector doesn't fit right column width any more.

phil-davis:
Backport ``stable10`` #33369

PR #32936, opened 2018-10-01
Added delay in search

gautamswati:
## Description
Issue  #31305 (To add delay to search operation)
I've added a delay of 500 milliseconds before the search results are showed in function $searchBox.keyup() .

## Motivation and Context
We don't want for the search results to be generated on every single keystroke .
That is why a wait (delay) is required to ascertain the user has stopped typing.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33493, opened 2018-11-13
Test delay in search bar on webUI

### LINKED ISSUE #31305: Delay start of search to avoid searching for single letters
Doing searches in the web interface can be very cumbersome because the search is started immediately after the first keypress.

### Steps to reproduce
1. Navigate to a folder with many files.
2. Start typing, Notice how the first typed letter of your search query will already trigger a search
3. Experience a huge slowdown which may even freeze your browser temporarily.

### Expected behaviour
Search should not start immediately but wait for a tiny amount. That way, if I type "foo" into the search bar at normal typing speed, I shouldn't get a search for "f", "fo" and "foo" but just the last search. Not starting the search until about 100-300ms (may be a bit longer if only one letter has been typed so far) since the last keystroke have passed should reduce the load on the browser and server, and avoid sending unnecessary search queries.

### Server configuration

**ownCloud version:** 10.0.7

dpakach:
## Description

#32936 added delay in the search box for calling ajax which is not merged yet.
Add acceptance test for the search delay on top of #32936.

## Related Issue

- Fixes #31305  #32936

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/search/js/search.js
+++ b/core/search/js/search.js
@@ -86,6 +86,8 @@
 			var lastSize = 30;
 			var lastResults = [];
 			var timeoutID = null;
+			/* For delaying search*/
+			var timeout = null;                                    
 
 			this.getLastQuery = function() {
 				return lastQuery;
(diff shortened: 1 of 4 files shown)


dpakach:
This test is not very reliable. We can count ajax calls but we may encounter some other random ajax  calls. So this is being closed.

PR #33496, opened 2018-11-13
[stable10] Set shipped apps max version to 10

PVince81:
## Description
Will make it possible to increase the platform version to 10.1 at some
point.

Note: this does not touch version.php yet, this will done in a future PR.

## Related Issue
Semver: https://github.com/owncloud/enterprise/issues/2532

## How Has This Been Tested?
- Quick OC check and see that files app still displays all sidebar tabs which means the apps were loaded correctly.
- Increase version.php and check that occ upgrade is not confused by max-version.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests
- [x] None of the above

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] info.xml changes

## Open tasks:

- [ ] apply same change on master for 11

PR #33497, opened 2018-11-13
Set max-version of shipped apps to 11

PVince81:
Same as https://github.com/owncloud/core/pull/33496 but with max-version=11


phil-davis (approved):

phil-davis:
This is good, because with semver we are promising full backward-compatibility for core 11.*.* - anything that works with the first release of 11.0.* will continue to work with all later 11.*.*

PR #33508, opened 2018-11-14
Add bonus PHP test files to see what codecov does

phil-davis:
Add a bunch of PHP code in ``tests/TestHelpers`` to see if this causes codecov to measure differently.


phil-davis:
codecov passed

PR #33509, opened 2018-11-14
Investigate coverage 2

phil-davis:
Add more test code to see what happens to codecov reports


phil-davis:
This also passes codecov, so it does seem to be excluding acceptance test code OK.

PR #33519, opened 2018-11-14
handle more than limit users

butonic:
Currently, the user_ldap might return more than 500 users per page when multiple servers or base dns are configured. The AllUsersIterator will simply ignore any results beyond the page size of 500 on every page, causing them to not by synced when running occ.

This has not been noticed for a while because users will still be synced when they log in.

This PR hardens the AllUsersIterator and allows it to handle bigger pages. The SeenUsersIterator is not affected, because it works on the account table.

### DIFF
--- a/lib/private/User/Sync/AllUsersIterator.php
+++ b/lib/private/User/Sync/AllUsersIterator.php
@@ -27,6 +27,19 @@ class AllUsersIterator extends UsersIterator {
 	 * @var UserInterface
 	 */
 	private $backend;
+	/**
+	 * @var int the current data position,
+	 *      we need to track it independently of parent::$position to handle data sets larger thin LIMIT properly
+	 */
+	private $dataPos = 0;
+
+	/**
+	 * @var int to cache the count($this->data) calculations
+	 */
+	private $endPos = 0;
+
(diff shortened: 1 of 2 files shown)


lib/private/User/Sync/AllUsersIterator.php
```diff
@@ -35,14 +48,25 @@ public function __construct(UserInterface $backend) {
...
 			$this->page++;
 			$offset = $this->page * self::LIMIT;
 			$this->data = $this->backend->getUsers('', self::LIMIT, $offset);
+			$this->dataPos = 0;
+			$this->endPos = \count($this->data);
+			$this->hasMoreData = $this->endPos >= self::LIMIT;
 		}
 	}
+
+	protected function currentDataPos() {
```
PVince81: seems this is unused, remove ?

PVince81 (changes requested):
looks good, just one minor thing to remove

lib/private/User/Sync/AllUsersIterator.php
butonic: nope it overwrites the parent method

PVince81 (approved):
:+1: please backport

PR #33520, opened 2018-11-14
stop excessive copying of user ids and object references

butonic:
The current user sync does excessively copy user ids ant object references. Depending on gc this might lead to an oom error with large user bases.


DeepDiver1975:
test are failing ....

butonic:
@PVince81 rebased and fixed the tests. please take over

PVince81 (approved):
Makes sense :+1:

PVince81:
stable10: https://github.com/owncloud/core/pull/34546

PR #33521, opened 2018-11-14
add function for trying to upload a file

### LINKED ISSUE #33505: Add function for making local files of varying size to use for upload tests.
We should include in core repo this function in order to have this kind of step in the core "API" steps so that it is easily used by any test that needs to make local files of varying size to use for upload tests.

This will make that this function can be accessed easily from any app acceptance tests

@owncloud/qa

davitol:
Fixes https://github.com/owncloud/core/issues/33505

We need a new function available everywhere for uploading a ```file of X size``` which does not validate the upload. **e.g.** useful for firewall scenarios

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -2069,7 +2069,7 @@ trait WebDav {
 	}
 
 	/**
-	 * @Given user :user has added file :destination of :bytes bytes
+	 * @Given user :user has uploaded file :destination of :bytes bytes
 	 *
 	 * @param string $user
 	 * @param string $destination
@@ -2077,7 +2077,22 @@ trait WebDav {
 	 *
 	 * @return void
 	 */
-	public function userAddsAFileTo($user, $destination, $bytes) {
+	public function userHasUploadedFileToOfBytes($user, $destination, $bytes) {
+		$this->userUploadsAFileToOfBytes($user, $destination, $bytes);
+		$expectedElements = new TableNode([["$destination"]]);
+		$this->checkElementList($user, $expectedElements);
+	}
+
+	/**
+	 * @When user :user uploads file :destination of :bytes bytes
+	 *
+	 * @param string $user
+	 * @param string $destination
+	 * @param string $bytes
+	 *
+	 * @return void
+	 */
+	public function userUploadsAFileToOfBytes($user, $destination, $bytes) {
 		$filename = "filespecificSize.txt";
 		$this->createLocalFileOfSpecificSize($filename, $bytes);
 		PHPUnit_Framework_Assert::assertFileExists($this->workStorageDirLocation() . $filename);
@@ -2087,8 +2102,6 @@ trait WebDav {
 			$destination
 		);
 		$this->removeFile($this->workStorageDirLocation(), $filename);
-		$expectedElements = new TableNode([["$destination"]]);
-		$this->checkElementList($user, $expectedElements);
 	}
 
 	/**
--- a/tests/acceptance/features/apiWebdavProperties/getQuota.feature
+++ b/tests/acceptance/features/apiWebdavProperties/getQuota.feature
@@ -49,7 +49,7 @@ Feature: get quota
   Scenario Outline: Retrieving folder quota when quota is set and a file was uploaded
     Given using <dav_version> DAV path
     And the quota of user "user0" has been set to "1 KB"
-    And user "user0" has added file "/prueba.txt" of 93 bytes
+    And user "user0" has uploaded file "/prueba.txt" of 93 bytes
     When user "user0" gets the following properties of folder "/" using the WebDAV API
       | {DAV:}quota-available-bytes |
     Then the single response should contain a property "{DAV:}quota-available-bytes" with value "577"
@@ -62,7 +62,7 @@ Feature: get quota
     Given using <dav_version> DAV path
     And user "user1" has been created
     And the quota of user "user1" has been set to "1 KB"
-    And user "user0" has added file "/user0.txt" of 93 bytes
+    And user "user0" has uploaded file "/user0.txt" of 93 bytes
     And user "user0" has shared file "user0.txt" with user "user1"
     When user "user1" gets the following properties of folder "/" using the WebDAV API
       | {DAV:}quota-available-bytes |


paurakhsharma:
@davitol, If I think this step should be a when statement. Because you are trying to do something here and in then statement, you verify that the file upload has failed?

davitol:
> @davitol, If I think this step should be a when statement. Because you are trying to do something here and in then statement, you verify that the file upload has failed?

IMHO it is not necessary. For example we already have this function in ```WebDav.php``` file.

https://github.com/owncloud/core/blob/f43a50a3b7e4efb5097d793b0eaab36941addd39/tests/acceptance/features/bootstrap/WebDav.php#L2072

individual-it (dismissed):

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2091,6 +2091,27 @@ public function userAddsAFileTo($user, $destination, $bytes) {
 		$this->checkElementList($user, $expectedElements);
 	}
 
+	/**
+	 * @Given user :user tries to add file :destination of :bytes bytes
```
phil-davis: We can just write:
```
When user :user uploads file :destination of :bytes bytes
```
because we can use a step like that in any scenario.
The ``Then`` step after it will test the expected result - if the file should exist or not.

And while we are here - in all other places we say ``uploads`` instead of ``adds`` for files. So refactor the existing method above to say:
```
Given user :user has uploaded file :destination of :bytes bytes
```
Change the method names to look like the changes step text.
Adjust the existing "has added file" test steps in ``getQuota.feature``
And the method above can call this one to do most of its work, until the last few lines.

phil-davis:
Note: part of the "problem" here is that the existing step that is being copy-paste-changed is already not optimal and should be refactored. This will happen every time someone touches the code, until the code is already all "perfect".

So maybe the discussion related to existing code that is not-so-good can be left for a later refactoring PR?

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2069,7 +2069,7 @@ public function asUserOnServerTheFilesUploadedToWithAllMechanismsShouldExit(
 	}
 
 	/**
-	 * @Given user :user has added file :destination of :bytes bytes
+	 * @When user :user has uploaded file :destination of :bytes bytes
```
phil-davis: This should still be a ``Given`` step.
I know it does not matter for Behat. But we try to write steps like:
```
Given user :user has done something
```
and for ``When`` steps:
```
When user :user does something
```
that makes it easier for people when reading the step text.

phil-davis: This method can be renamed to ``userHasUploadedFileOfBytes``

And most of the code in this method can be deleted and replaced with a call like:
```
userUploadsAFileTo($user, $destination, $bytes);
```

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2091,6 +2091,27 @@ public function userAddsAFileTo($user, $destination, $bytes) {
...
+	/**
+	 * @When user :user uploads file :destination of :bytes bytes
+	 *
+	 * @param string $user
+	 * @param string $destination
+	 * @param string $bytes
+	 *
+	 * @return void
+	 */
+	public function userUploadsAFileTo($user, $destination, $bytes) {
```
phil-davis: While you are editing again, this method name could be:
```
userUploadsAFileToOfBytes
```
then it will not accidentally conflicts with some other upload method.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2091,6 +2091,27 @@ public function userAddsAFileTo($user, $destination, $bytes) {
...
+	 *
+	 * @return void
+	 */
+	public function userTriesToAddAFileTo($user, $destination, $bytes) {
+		$filename = "filespecificSize.txt";
+		$this->createLocalFileOfSpecificSize($filename, $bytes);
+		PHPUnit_Framework_Assert::assertFileExists($this->workStorageDirLocation() . $filename);
+		$this->userUploadsAFileTo(
+			$user,
+			$this->temporaryStorageSubfolderName() . "/$filename",
```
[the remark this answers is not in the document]
phil-davis: IMO this can be refactored later - it is the existing code from ``userHasUploadedFileToOfBytes`` that is above.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2069,15 +2069,15 @@ public function asUserOnServerTheFilesUploadedToWithAllMechanismsShouldExit(
...
+	 * @Given user :user has uploaded file :destination of :bytes bytes
 	 *
 	 * @param string $user
 	 * @param string $destination
 	 * @param string $bytes
 	 *
 	 * @return void
 	 */
-	public function userAddsAFileTo($user, $destination, $bytes) {
+	public function userHasUploadedFileToOfBytes($user, $destination, $bytes) {
```
phil-davis: Hmmm - I  thought that I also commented about the duplicated code. Anyway,
this method can now call the new ``userUploadsAFileToOfBytes`` method below, something like:
```
$this->userUploadsAFileToOfBytes($user, $destination, $bytes);
$expectedElements = new TableNode([["$destination"]]);
$this->checkElementList($user, $expectedElements);
```

This pattern can be used often in the acceptance tests - a ``Given``` step can:
- call the ``When`` step
- do some checks to add some confidence that it worked

phil-davis: I promise this is the last change needed for this PR :)

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2069,26 +2069,39 @@ public function asUserOnServerTheFilesUploadedToWithAllMechanismsShouldExit(
...
 	 *
 	 * @return void
 	 */
-	public function userAddsAFileTo($user, $destination, $bytes) {
+	public function userUploadsAFileToOfBytes($user, $destination, $bytes) {
 		$filename = "filespecificSize.txt";
 		$this->createLocalFileOfSpecificSize($filename, $bytes);
 		PHPUnit_Framework_Assert::assertFileExists($this->workStorageDirLocation() . $filename);
-		$this->userUploadsAFileTo(
+		$this->userUploadsAFileToOfBytes(
```
phil-davis: Something happened here - it is calling itself.
IMO this line should not have changed.

phil-davis (approved):

phil-davis:
@davitol looks good. Please make a backport to ``stable10``

paurakhsharma:
Backport in: #33540

PR #33463, opened 2018-11-09
Update the unique index for the systemtag table

sharidas:
Update the unique index for the systemtag table.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Update the unique index for the systemtag table.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

If the unique index is not updated with `assignable`, then inserts to the table were causing problems.

## How Has This Been Tested?

- Download owncloud http://download.owncloud.org/community/owncloud-8.2.11.tar.bz2 and install it.
- Download owncloud http://download.owncloud.org/community/owncloud-daily-stable10-qa.tar.bz2 for the upgrade
- When upgraded from 8.2.11 to 10 below are the upgrade log captured from the console:
```
➜  owncloud4 ./occ upgrade                                                 
Cannot load Xdebug - it was already loaded
ownCloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
2018-11-09T11:00:26+00:00 Set log level to debug
2018-11-09T11:00:26+00:00 Repair step: Repair MySQL database engine
2018-11-09T11:00:26+00:00 Repair info: Not a mysql database -> nothing to do
2018-11-09T11:00:26+00:00 Repair step: Repair MySQL collation
2018-11-09T11:00:26+00:00 Repair info: Not a mysql database -> nothing to do
2018-11-09T11:00:26+00:00 Repair step: Repair SQLite autoincrement
2018-11-09T11:00:26+00:00 Repair step: Repair orphaned reshare
2018-11-09T11:00:26+00:00 Repair step: Repair duplicate entries in oc_lucene_status
2018-11-09T11:00:26+00:00 Repair info: lucene_status table does not exist -> nothing to do
2018-11-09T11:00:26+00:00 Repair step: Upgrade app code from the marketplace
2018-11-09T11:00:26+00:00 Repair info: Enabling market app to assist with update
2018-11-09T11:00:26+00:00 Repair info: Using market to update existing apps
2018-11-09T11:00:26+00:00 Repair info: Attempting to update the following missing apps from market: activity, files_pdfviewer, files_texteditor, gallery, templateeditor
2018-11-09T11:00:26+00:00 Repair info: Fetching app from market: activity
2018-11-09T11:00:45+00:00 Repair info: Fetching app from market: files_pdfviewer
2018-11-09T11:01:10+00:00 Repair info: Fetching app from market: files_texteditor
[...]

PR #33522, opened 2018-11-14
[stable10] Backport of Update the unique index for the systemtag table

sharidas:
Update the unique index for the systemtag table.

## Description

Update the unique index for the systemtag table.

- Let us consider the dataprovider https://github.com/owncloud/core/blob/master/tests/lib/SystemTag/SystemTagManagerTest.php#L95-L102 for the test https://github.com/owncloud/core/blob/master/tests/lib/SystemTag/SystemTagManagerTest.php#L110
- The moment https://github.com/owncloud/core/blob/master/tests/lib/SystemTag/SystemTagManagerTest.php#L101 dataset is taken for execution, the database snapshot looks like, shown below:
```
sqlite> select * from oc_systemtag;
id          name        visibility  editable    assignable
----------  ----------  ----------  ----------  ----------
315         one         0           0           1         
316         one         1           0           1         
317         one         0           1           1         
318         one         1           1           1         
319         two         0           0           0         
sqlite>
```
- When the code tries to create a new tag for https://github.com/owncloud/core/blob/master/tests/lib/SystemTag/SystemTagManagerTest.php#L101, the new entry we expect will be like:
```
320         two         0           0           1
```
This conflicts with the unique index. For a moment just look at the column `name`, `visibility` and `editable` column only, then rows with id 319 and planned insertion (id 320) are duplicates!!! And causes the failure while executing the last dataprovider in the test. 
- The schema of the table is as shown below:
```
sqlite> .schema oc_systemtag
CREATE TABLE oc_systemtag (id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT, name VARCHAR(64) DEFAULT '' NOT NULL, visibility SMALLINT DEFAULT 1 NOT NULL, editable SMALLINT DEFAULT 1 NOT NULL, assignable INTEGER DEFAULT 0 NOT NULL);
CREATE UNIQUE INDEX tag_ident ON oc_systemtag (name, visibility, editable);
sqlite>
```
the column `assignable` is missing in the unique index.

### How this fix solves the problem ###
- When `assignable` column is added along with `name`, `visibility` and `editable` in the unique index, the issue caused by adding:
```
320         two         0           0           1
```
gets rectified because row id `319` and `320` are different. Only difference is in the assignable column ( 0 and 1).

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Update the unique index of the systemtag with `assignable` column. If `assignable` column is not added to the unique index, then there are chances that duplicate entry is found with `name`, `visibility` and `editable`. This patch fixes the issue cauesd.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/Migrations/Version20181113071753.php
+++ b/core/Migrations/Version20181113071753.php
@@ -1,0 +1,42 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use OCP\Migration\ISchemaMigration;
+
+/**
+ * Add assignable column to unique index in oc_systemtags table
+ */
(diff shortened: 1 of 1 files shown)


sharidas:
Original PR https://github.com/owncloud/core/pull/33463

PVince81 (approved):
:+1:

PR #33040, opened 2018-10-07
Provisioning api edit displayname

phil-davis:
## Description
1) Allow ``displayname`` (in addition to the existing ``display``) as a key for modifying the display name of a user via the provisioning API.
2) Adjust the acceptance tests  to use ``displayname``

## Related Issue
- Fixes #33039 

## Motivation and Context
Make modify and get of display name via the provisioning API both use ``displayname`` as the key.

## How Has This Been Tested?
Run acceptance test scenarios.
Run ``curl`` commands like:
```
curl -X PUT http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/users/zzz -d key="display" -d value="Me"
```

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

Issue #33039, opened 2018-10-07
Change user display name via provisioning API is inconsistent


phil-davis:
### Steps to reproduce
1. create user ``zzz``
2. retrieve  the information about user ``zzz`` using the provisioning API
```
curl http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/users/zzz
```
A data item comes back called ``displayname``
3. try to modify the ``displayname``
```
curl -X PUT http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/users/zzz -d key="displayname" -d value="Me"
```

### Expected behaviour
Success

### Actual behaviour
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>997</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

### Server configuration
**Operating system**:
Ubuntu 18.04
**Web server:**
development
**Database:**
mySQL
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
git master

phil-davis:
The actual ``curl``` command that works is:
```
curl -X PUT http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/users/zzz -d key="display" -d value="Me"
```

The key is called ``display``, not ``displayname``

The User Provisioning API documentation https://doc.owncloud.org/server/10.0/admin_manual/configuration/user/user_provisioning_api.html under "Edit User" mentions ``displayname``

I suggest that this be made consistent by adding the ability to specify the display name by using the ``displayname`` key or the ``display`` key.

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/pull/32960 (edit user settings change statements on Api provisioning), https://github.com/owncloud/core/issues/13139 (Provisioning api uses uniqid()), https://github.com/owncloud/core/issues/15242 (Add quota and display to the default creation of users in the provisioning API), https://github.com/owncloud/core/issues/27976 (No working PUT User-Provisioning API ), and https://github.com/owncloud/core/issues/20817 (UTF-8 support different on User Provisioning API function calls).

PR #33523, opened 2018-11-15
Add 'display' as well as 'displayname' editUser acceptance tests

phil-davis:
## Description
Add test scenarios to test editing a user "display name" using the key ``display``.
Adjust the ``Smoketest`` tagged scenarios so that they only use this "old" key ``display``.

## Related Issues
Issue #33039 
and PR #33040 
docker server issue https://github.com/owncloud-docker/server/pull/77

## Motivation and Context
1) The related issue/PR fixed up the provisioning API so that the "display name" of a user can be edited by specifying the key ``displayname`` (as well as the existing inconsistent key ```display``)
2) The acceptance tests were adjusted to use the new key ``displayname``

But we should also test that the old key ``display`` still works (for backward compatibility). And the scenarios that are tagged ``Smoketest`` should be ones that use the old key, so that they will/should pass when run against older ownCloud versions (e.g. 10.0.10 or 10.0.9)

## How Has This Been Tested?
Local acceptance test runs

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiProvisioning-v1/editUser.feature
+++ b/tests/acceptance/features/apiProvisioning-v1/editUser.feature
@@ -16,6 +16,13 @@ Feature: edit users
     And the email address of user "brand-new-user" should be "brand-new-user@example.com"
 
   @smokeTest
+  Scenario: the administrator can edit a user display (the API allows editing the "display name" by using the key word "display")
+    Given user "brand-new-user" has been created
+    When the administrator changes the display of user "brand-new-user" to "A New User" using the provisioning API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And the display name of user "brand-new-user" should be "A New User"
+
   Scenario: the administrator can edit a user display name
     Given user "brand-new-user" has been created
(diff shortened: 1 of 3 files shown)


paurakhsharma (approved):

individual-it:
we could have also smoke tests testing the new version but skip them on `10.0.10` and whatever old versions we are testing

individual-it (approved):

phil-davis:
I thought about adding version tags. But it seems a pain. The smoke tests are just to find "smoke". The full test suite runs anyway on every PR and finds "real fire". We want to limit the "smoke" tests anyway (otherwise they can easily expand to "smoketests" = "alltests")

phil-davis:
Backport ``stable10`` #33524

individual-it:
> The smoke tests are just to find "smoke".

the smoke tests are there to see if the magic smoke leaves or stays inside http://www.catb.org/jargon/html/M/magic-smoke.html

PR #33530, opened 2018-11-15
log broken smb config params

butonic:
log broken smb config params


jvillafanez:
There is another log for that in line 81. I don't see any advantage

butonic:
@jvillafanez that is only logged when tracing is enabled via `'smb.logging.enable' => true,` and in loglevel debug. The exception will bubble up and show on the CLI. Always, which makes a lot of sense IMO.

jvillafanez (approved):

mmattel:
Any update? Mergable ?

PR #28874, opened 2017-08-31
Use new DAV endpoint in web UI

PVince81:
## Description
Use new DAV endpoint for everything, including single file uploads

- [x] ⚠️ requires https://github.com/owncloud/core/pull/28853 to avoid performance refressions and regressions related to locking and external storage ⚠️ 

## Related Issue
None.
Also see https://github.com/owncloud/core/pull/25494

## Motivation and Context
Same as https://github.com/owncloud/core/pull/25494

## How Has This Been Tested?
- [x] TEST: upload single file into current view
- [x] TEST: upload a file as a user "test@test" to confirm encoding works for the full DAV path
- [ ] TEST: make sure these https://github.com/owncloud/core/issues/28779 steps work correctly

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Not sure if we want this right now, I'm thinking of reverting new dav for 10.0.3 due to https://github.com/owncloud/core/issues/28779.

Setting to "planned" to have a look later.

PR #33544, opened 2018-11-15
[stable10] Use new DAV endpoint in web UI file list and upload

PVince81:
Backport of https://github.com/owncloud/core/pull/28874 to stable10.

Maybe we merge this for 10.1 ?

@DeepDiver1975 

cc @tomneedham


PVince81:
as discussed with @DeepDiver1975 we will merge this backport for 10.1

PR #33571, opened 2018-11-19
Test creating users with various passwords using the Provisioning API

phil-davis:
## Description
Add test scenarios to verify that users can be created with various unusual passwords using the Provisioning API, and that a password with just space is invalid.

## Related Issue
- Part of issue #33570 

## Motivation and Context
See issue

## How Has This Been Tested?
Local run of test scenarios

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33580, opened 2018-11-20
[stable10] Test creating users with various passwords using the Provisioning API

phil-davis:
Backport #33571


individual-it (approved):

phil-davis:
Note: test fail was just a PHP unit segfault at the end of a successful PHP unit run. No real fail.

PR #33545, opened 2018-11-15
Migration for bigint on DAV tables

PVince81:
## Description
Convert id columns of many DAV-related tables to bigint.

Note that some of these migrations might have existed before
but some update paths did not trigger them properly, so this
migration here exists to align all update paths.

## Related Issue
Fixes https://github.com/owncloud/core/issues/32555

## How Has This Been Tested?
Setup OC from scratch and check that every id column of the [affected tables](https://github.com/owncloud/core/issues/32555#issuecomment-439103378) now appears as BIGINT on MariaDB.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33603, opened 2018-11-21
[stable10] Migration for bigint on DAV tables

PVince81:
Backport of https://github.com/owncloud/core/pull/33545 to stable10.

Note: I did not increase version for migrations on master (forgot, but also we said no need to until later) but I did increase the version here.

@DeepDiver1975

### DIFF
--- a/version.php
+++ b/version.php
@@ -25,7 +25,7 @@
 // We only can count up. The 4. digit is only for the internal patchlevel to trigger DB upgrades
 // between betas, final and RCs. This is _not_ the public version number. Reset minor/patchlevel
 // when updating major/minor version number.
-$OC_Version = [10, 0, 10, 4];
+$OC_Version = [10, 0, 10, 5];
 
 // The human readable string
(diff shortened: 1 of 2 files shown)


version.php
```diff
@@ -25,7 +25,7 @@
 // We only can count up. The 4. digit is only for the internal patchlevel to trigger DB upgrades
 // between betas, final and RCs. This is _not_ the public version number. Reset minor/patchlevel
 // when updating major/minor version number.
-$OC_Version = [10, 0, 10, 4];
+$OC_Version = [10, 0, 10, 5];
```
patrickjahns: Hum - shouldn't that version increase be either 10,0,11 ? 

Otherwise this seems strange

DeepDiver1975: No. We increment the Patch Level in release. The 4th version is used for Database update

PVince81: this is mostly for people who have an existing stable10 instance and do a git pull... we can discuss separately whether we want to continue this practice or go with "don't use git in production"

PR #33636, opened 2018-11-23
Trigger CI

phil-davis:
To see what is wrong with composer on drone


phil-davis:
Just a phpunit segfault!

PR #33632, opened 2018-11-23
Fix #33630. Fix cannot set 0 as value on files_external through OCC command

skshetry:
## Description

The OCC command didn't allow us to set value 0 due to the use of `if ($value)` checks. Now, This PR modifies it to check for null value only.

## Related Issue

- Fixes #33630 

## How Has This Been Tested?

- Locally

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33643, opened 2018-11-26
[stable10] Fix #33630. Fix cannot set 0 as value for config through OCC command

### LINKED ISSUE #33630: Cannot set  `filesystem_check_changes` to never through occ command
This is a bug in the `files_external` app in which if we want to set `filesystem_check_changes` to never, it doesnot set any value. Instead it just returns whatever the current value is. 

## How To Reproduce:
1. Connect to an external storage with owncloud
2. Run the following:
```
sudo -u www-data ./occ files_external:option <mount_id> filesystem_check_changes 0
```

## Expected Outcome:
1. No output is printed through occ command.
2. `Check for changes` is set to `Never`.
(On Settings->Storage->External Storage->GearIcon :gear: ) on the webUI

## Actual Outcome:
1. Prints already set config.
2. Doesnot change anything on the webUI

## Bug Analysis
This seems to be due to [this conditional check](https://github.com/owncloud/core/blob/8cd722356cf81df22699a1e17450f2b034235b23/apps/files_external/lib/Command/Config.php#L75-L79).

skshetry:
Backport for #33632

### DIFF
--- a/apps/files_external/lib/Command/Config.php
+++ b/apps/files_external/lib/Command/Config.php
@@ -72,7 +72,8 @@ class Config extends Base {
 		}
 
 		$value = $input->getArgument('value');
-		if ($value) {
+
+		if ($value !== null) {
 			$this->setOption($mount, $key, $value, $output);
 		} else {
 			$this->getOption($mount, $key, $output);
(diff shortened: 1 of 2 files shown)


individual-it (approved):

phil-davis (approved):

phil-davis:
Rebased to make drone run

PVince81:
ignoring env related smb fail -> merge

PR #33672, opened 2018-11-27
Adding config for the lock bot

DeepDiver1975:
## Description
https://probot.github.io/apps/lock/ has been enabled for core and this is the config

## Motivation and Context
Old issues and pull requests are sometimes used people to report issue.
These comments are often attached to no longer valid code and tend to get lost.

Locking these old issues and pull requests can solve this.

## Types of changes
- [x] Technical debt
- [ ] Tests

### DIFF
--- a/.github/lock.yml
+++ b/.github/lock.yml
@@ -1,0 +1,38 @@
+# Configuration for lock-threads - https://github.com/dessant/lock-threads
+
+# Number of days of inactivity before a closed issue or pull request is locked
+daysUntilLock: 365
+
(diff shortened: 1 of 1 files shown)


PVince81:
before we add yet another bot I'd like to consider whether we want to continue using Gitmate.

if this new bot is able to achieve the same level of functionality we could use it instead

DeepDiver1975:
> before we add yet another bot I'd like to consider whether we want to continue using Gitmate.

Gitmate seems dead for a long time. I'll resubmit this PR later to bring this in again

PR #33675, opened 2018-11-28
[stable10] Adjust Storage.php testStat timing

phil-davis:
Backport #33674


phil-davis:
I created this prospective backport so that the pass/fail status in ``stable10`` can be demonstrated while we still know that the ``smb_windows`` back-end storage has the time-offset that was causing problems.

DeepDiver1975 (approved):

Issue #33413, opened 2018-11-05
Delete Custom Dav Properties on file delete


micbar:
### Steps to reproduce
1. Create custom DAV Property
2. Create File with DAV Property, gets written into oc_properties table
3. Delete the file

### Expected behaviour
All DAV Properties for this fileID should be deleted

### Actual behaviour
The orphaned properties remain in the database.

### Server configuration
**Operating system**:
MAC
**Web server:**
Apache2
**Database:**
MySQL
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
10.0.8

**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git
**Signing status (ownCloud 9.0 and above):**

**List of activated apps:**
  - activity: 2.3.7
  - comments: 0.3.0
  - dav: 0.3.2
  - enterprise_key: 0.1.4
  - federatedfilesharing: 0.3.1
  - federation: 0.1.0
  - files: 1.5.1
  - files_external: 0.7.1
  - files_lifecycle: 1.0.0
  - files_sharing: 0.10.1
  - files_texteditor: 2.2.1
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - market: 0.2.5
  - notifications: 0.3.2
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - updatenotification: 0.2.1

## Possible Solution
- Needs to be discussed
- Could be solved by introducing foreign keys with cascade delete

micbar:
@PVince81 @DeepDiver1975

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/27274 (delete), https://github.com/owncloud/core/issues/13824 (Delete File), https://github.com/owncloud/core/issues/2438 (files deleted on sync), https://github.com/owncloud/core/issues/19140 (Consolidate deleted files in deleted folders), and https://github.com/owncloud/core/issues/2716 (Cannot permanently delete deleted files).

PVince81:
Possible solutions:
- foreign keys + cascading delete (also see https://github.com/owncloud/core/issues/13143)
- have a background job that cleans up orphaned properties

Whatever we do we'll need either a bkg job or a repair job to clean up existing installs out there, and that before telling the DB to add a foreign key (if we choose to go that route)

micbar:
@PVince81 
On 10.0.8 I do the follwing.

[...]

PR #33677, opened 2018-11-28
[stable10] Backport of Add background job to clean up orphaned proper…

sharidas:
## Description

Clean up orphan properties in the table. There can be situations when deleting files would not eventually clean up the traces in oc_properties table. This change helps to clean up the properties table if the files/folders are not available in the filecache.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/33413

## How Has This Been Tested?

-  The unit test should cover the test.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/BackgroundJob/CleanProperties.php
+++ b/apps/dav/lib/BackgroundJob/CleanProperties.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\BackgroundJob;
+
+use OC\BackgroundJob\TimedJob;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+use OCP\ILogger;
+
+/**
+ * Class CleanProperties
+ *
+ * @package OCA\DAV\BackgroundJob
+ */
+class CleanProperties extends TimedJob {
+	const CHUNK_SIZE = 200;
+
+	/** @var IDBConnection  */
+	private $connection;
+	/** @var ILogger  */
+	private $logger;
+
+	/**
+	 * CleanProperties constructor.
+	 *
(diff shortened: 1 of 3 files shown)


sharidas:
Original PR https://github.com/owncloud/core/pull/33464

apps/dav/lib/BackgroundJob/CleanProperties.php
```diff
@@ -0,0 +1,108 @@
...
+		 */
+		$qb->select('prop.fileid')
+			->from('properties', 'prop')
+			->where($qb->expr()->isNull('fc.fileid'))
+			->leftJoin('prop', 'filecache', 'fc', $qb->expr()->eq('prop.fileid', 'fc.fileid'))
+			->setMaxResults(self::CHUNK_SIZE);
+
+		while ($rows = $qb->execute()->fetchAll()) {
+			$fileIds = \array_map(function ($row) {
+				return (int) $row['fileid'];
```
PVince81: a small concern comes to mind: are we also casting to int in other repair steps ?

I don't remember whether PHP properly deals with bigint values when casting to "int" so would be good to double check, just to be sure

sharidas: I was going through conversion of string to int ( the reason, I used casting to int in this context ) and found this link http://php.net/manual/en/language.types.string.php#language.types.string.conversion. To be specific :
```
If the string does not contain any of the characters '.', 'e', or 'E' and the numeric value fits into integer type limits (as defined by PHP_INT_MAX), the string will be evaluated as an integer. In all other cases it will be evaluated as a float.
```

It says about PHP_INT_MAX. This variable depends on the machine whether it is 32 bit or 64 bit. 

Or I was having another thought, would it be wise to pass the `$row['fileid']` without casting to int. And then in the delete we could have :
`$qb->setParameter('fileids', $fileids, IQueryBuilder::PARAM_STR_ARRAY);`
instead of 
`$qb->setParameter('fileids', $fileids, IQueryBuilder::PARAM_INT_ARRAY);`
In this case, we are not doing anything to convert or so. The query builder will take care. I would like to know your suggestion here.

PVince81: I'm not sure either. Can you check the other repair steps and background jobs where we handle lists of file ids ?

sharidas: After doing investigation on repair steps, the CleanTags repair caught my attention. Especially the conversion to int:
- conversion to int 
 https://github.com/owncloud/core/blob/master/lib/private/Repair/CleanTags.php#L185-L187
- The chunked orphan item is then used as https://github.com/owncloud/core/blob/master/lib/private/Repair/CleanTags.php#L197

Since we are having a discussion about conversion, my gut feel says that while handling with huge data ( I mean large id's, meaning large number of rows ), conversion to int and then passing to delete query builder would hit the performance badly. Somehow looking at the repair step, I am inclined towards:
`$qb->setParameter('fileids', $fileids, IQueryBuilder::PARAM_STR_ARRAY);` ( i.e, don't convert to int ). Let the query builder handle it. 

Let me also check for further instance and update here.

PVince81: ok, so we already used the int array pattern. let's leave the code as is.

in general we don't support 32 bit systems officially any more

PVince81 (approved):
:+1:

PR #33713, opened 2018-11-30
Filter double appearing addressbooks

Julian1998:
## Description

Filter for the double-appearing address book, when the user is sharing an addressbook with groups he is also a member of.

## How Has This Been Tested?

- Create an addressbook in contacts app
- Share it with a group the user also belongs to
- The shared addressbook is now filtered

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised:

## Open tasks:

- [ ] Backport

### DIFF
--- a/apps/dav/lib/CardDAV/CardDavBackend.php
+++ b/apps/dav/lib/CardDAV/CardDavBackend.php
@@ -162,13 +162,20 @@ class CardDavBackend implements BackendInterface, SyncSupport {
 		$principals = $this->principalBackend->getGroupMembership($principalUri, true);
 		$principals[]= $principalUri;
 
+		$addressBooksIds[] = -1;
+		foreach ($addressBooks as $book) {
+			$addressBooksIds[] = $book['id'];
+		}
+
 		$query = $this->db->getQueryBuilder();
 		$result = $query->select(['a.id', 'a.uri', 'a.displayname', 'a.principaluri', 'a.description', 'a.synctoken', 's.access'])
 			->from('dav_shares', 's')
 			->join('s', 'addressbooks', 'a', $query->expr()->eq('s.resourceid', 'a.id'))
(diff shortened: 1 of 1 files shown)


Julian1998:
I still have to modify CardDavBackend unit tests.

apps/dav/lib/CardDAV/CardDavBackend.php
```diff
@@ -162,7 +162,7 @@ public function getAddressBooksForUser($principalUri) {
 		$principals = $this->principalBackend->getGroupMembership($principalUri, true);
 		$principals[]= $principalUri;
 
-		$addressBooksIds = [];
+		$addressBooksIds[] = -1;
```
Julian1998: DB query fails if addressBooksIds is empty, so if the user doesn't have any addressbooks. This is the fix but rly dirty. Any suggestions how to do better @DeepDiver1975

DeepDiver1975: de-duplication should happen a few lines below .... L187

Julian1998: @DeepDiver1975 I see the problem in the db query! Why ask for more information than needed and afterwards filter that infos again? It's unneccessary in my opinion and we could save performance by doing it there

VicDeo:
@Julian1998 looks good in general, please rebase

DeepDiver1975:
> @Julian1998 looks good in general, please rebase

@VicDeo feel free to grab the PR - @Julian1998 is only working 1 day per week - usually fridays ....

VicDeo (approved):
Rebased, squashed.
Tested and works.

VicDeo:
Stable10: https://github.com/owncloud/core/pull/35603

PR #33714, opened 2018-12-01
Revert "[stable10] Backport of Add background job to clean up orphaned proper…"

phil-davis:
Reverts owncloud/core#33677

The problem is that user name ``user1`` is also used in ``EncryptAllTest.php`` ``testEncryptFileFileId()`` - see PR #33718 for a demonstration.

If we revert here, then we can get CI passing again in ``stable10``. The original PR code can be investigated, adjusted and re-submitted next week.

Issue #33720 can be used to track this.


phil-davis:
IMO it passes because ``CleanPropertiesTest.php`` has ``testDeleteLargeOrphans()`` with ``providesDeleteLargeOrphans()`` that creates a lot of files and orphans them... I suspect that that uses up space in the object-store and the space does not get recovered. In later tests the space finally runs out, maybe.

patrickjahns:
> In later tests the space finally runs out, maybe.

No - there is no space issue - you can double check the monitor from CEPH - it mentions that are < 200MB of 500MB used.

Something else in the unit tests leaves residue, which causes issues -> needs to be debugged

PVince81:
@sharidas please resend but change the user id to random / unique ones specific to the test to avoid side effect with other bad tests.

PR #33722, opened 2018-12-03
[stable10] Backport of Add background job to clean up orphaned proper…

sharidas:
…ties

Add a background job to clean up orphaned properties
in the table.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Clean up orphan properties in the table. There can be situations when deleting files would not eventually clean up the traces in oc_properties table. This change helps to clean up the properties table if the files/folders are not available in the filecache. 
The tests are updated accordingly to make sure the users are created with unique names.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/33413
- Fixes https://github.com/owncloud/core/issues/33720

## Motivation and Context

Cleanup the oc_properties table. No orphan entries should be found in the table.

## How Has This Been Tested?

- The unit test should cover the test.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33725, opened 2018-12-03
Use unique user names in the unit test for CleanProperties

### LINKED ISSUE #33720: Mentioning the same user name in multiple unit tests is breaking CI in stable10
PR #33677 introduced ``CleanProperties`` and ``CleanPropertiesTest`` which uses user name ``user1`` in its tests.

``testEncryptFileFileId()`` also uses user name ``user``. That fails with:
```
There was 1 error:

1) OCA\Encryption\Tests\Crypto\EncryptAllTest::testEncryptFileFileId
OCP\Files\StorageNotAvailableException: 

/drone/src/lib/private/Files/Storage/Wrapper/Availability.php:75
/drone/src/lib/private/Files/Storage/Wrapper/Availability.php:457
/drone/src/lib/private/Files/Storage/Wrapper/Checksum.php:202
/drone/src/lib/private/Files/ObjectStore/NoopScanner.php:90
/drone/src/lib/private/Files/ObjectStore/NoopScanner.php:60
/drone/src/lib/private/Files/View.php:1367
/drone/src/lib/private/Files/View.php:1408
/drone/src/lib/private/Files/Node/Root.php:188
/drone/src/lib/private/Files/Node/Root.php:361
/drone/src/lib/private/Server.php:1029
/drone/src/apps/encryption/tests/Crypto/EncryptAllTest.php:541
```

There is some problem with user name ``user1`` being used in multiple tests. Presumably the user is not getting cleaned up after each test. See demonstration of hack workaround in PR #33718 

Investigate and fix.

sharidas:
Use unique user names in the unit test for CleanProperties.

## Description

Use unique usernames for users in the unit tests. It is better to use unique names instead of hard coded names in the tests. There has been issues found because of this behaviour.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/33720

## Motivation and Context

It has been found better to use unique user names instead of hard coded names in the unit test.

## How Has This Been Tested?

- Used `getUniqueID` to generate unique user name in the unit test.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/tests/unit/BackgroundJob/CleanPropertiesTest.php
+++ b/apps/dav/tests/unit/BackgroundJob/CleanPropertiesTest.php
@@ -41,6 +41,8 @@ class CleanPropertiesTest extends TestCase {
 	private $logger;
 	/** @var CleanProperties */
 	private $cleanProperties;
+	/** @var string */
+	private $username;
 
 	public function setUp() {
 		parent::setUp();
@@ -48,12 +50,13 @@ class CleanPropertiesTest extends TestCase {
 		$this->connection = \OC::$server->getDatabaseConnection();
 		$this->logger = \OC::$server->getLogger();
 		$this->cleanProperties = new CleanProperties($this->connection, $this->logger);
-		$this->createUser('user1');
-		$this->loginAsUser('user1');
+		$this->username = $this->getUniqueID('usercleanprop_');
+		$this->createUser($this->username);
+		$this->loginAsUser($this->username);
 	}
 
 	public function testDeleteOrphanEntries() {
-		$userFolder = \OC::$server->getUserFolder('user1');
+		$userFolder = \OC::$server->getUserFolder($this->username);
 		$userFolder->newFile('a.txt');
 		$userFolder->newFile('b.txt');
 		$userFolder->newFile('c.txt');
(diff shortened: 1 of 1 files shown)


sharidas:
This PR is a forward port of unit test fromhttps://github.com/owncloud/core/pull/33722. The unit test change to create user names with unique names is ported in this PR.

sharidas:
Backport to stable10 which has this change https://github.com/owncloud/core/pull/33722

patrickjahns:
This seems to be rather a workaround, when the test should rather cleanup created users after run. 
@sharidas - can you clarify what keeps us from cleaning up in the `teardown` step?

PVince81:
@patrickjahns this was designed to not have to clean up explicitly as long as one uses these methods: https://github.com/owncloud/core/blob/master/tests/lib/Traits/UserTrait.php#L29 and https://github.com/owncloud/core/blob/master/tests/lib/Traits/UserTrait.php#L58 for convenience.

I don't think it's worth starting to change this pattern now.

PVince81 (approved):
:+1:

patrickjahns:
> @patrickjahns this was designed to not have to clean up explicitly as long as one uses these methods: https://github.com/owncloud/core/blob/master/tests/lib/Traits/UserTrait.php#L29 and https://github.com/owncloud/core/blob/master/tests/lib/Traits/UserTrait.php#L58 for convenience.
> 
> I don't think it's worth starting to change this pattern now.
Not asking to change the pattern - just trying to investigate if our test suite all along is acting correctly - or if we have some unknowns in test scenarios that rely on some stray information in the database.

If https://github.com/owncloud/core/blob/master/tests/lib/Traits/UserTrait.php#L58 called via https://github.com/owncloud/core/blob/master/tests/lib/TestCase.php#L162-L169 would properly teardown - why would we need to change the username in this test case scenario?

PVince81:
possibly because another unclean test that ran before has left over the same user id in the database

PR #33757, opened 2018-12-04
Use chunks of arrays to pass to SQL query instead of big list

sharidas:
Use chunks of arrays to pass to SQL query instead of
passing huge list obtained.

## Description

While passing lists to SQL `IN` statements its better to chunk the list into sublist ( or arrays to sub arrays ), and then pass to the query. This way we can get results without exception. An issue was found with oracle db, where the maximum number of expressions in a list supported by oracle is 1000. And without chunking lists if we pass to the SQL query, it would cause exception. A unit test is also added to support the chunking.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Chunk the list to sublists while passing to `IN` statement of SQL query. Else there are chances of execptiopn from the doctrine.

## How Has This Been Tested?

- Tested using the unit test

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/SystemTag/SystemTagObjectMapper.php
+++ b/lib/private/SystemTag/SystemTagObjectMapper.php
@@ -34,6 +34,7 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 class SystemTagObjectMapper implements ISystemTagObjectMapper {
 	const RELATION_TABLE = 'systemtag_object_mapping';
+	const CHUNK_SIZE = 200;
 
 	/** @var ISystemTagManager */
 	protected $tagManager;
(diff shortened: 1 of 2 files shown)


sharidas:
The status of the PR and what this PR tries to do:
- [x] Use `\array_chunk()` to split the objectIds passed to the method `getTagIdsForObjects()` in `SystemTagObjectMapper`.
   - [x] This would prevent passing large arrays to the function which if not chunked might cause exception for different varients of db's we support.

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport

PR #33807, opened 2018-12-07
[do not merge] Test pause after init user

phil-davis:
Run the suspect test lots of times


phil-davis:
The test has run a total of 100 times. A couple of runs had real drone errors fetching the repo from git. Every run that got to the acceptance tests phase, it passed - great stuff.

PR #33680, opened 2018-11-28
merge teststeps

individual-it:
## Description
merge test steps

## Motivation and Context
do, what the ToDo line tells you to do

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33812, opened 2018-12-07
Revert "merge teststeps"

phil-davis:
Reverts owncloud/core#33680

Scenario: Upload a file called "0" using new chunking is a bug in core ``10.0.10``

So we do not want it to end up in the ``smokeTest`` scenario outline examples, because then it may get run against an "old" ``10.010`` system, and of course it will fail.

We cannot tag individual rows of an example table. So it can't be "fixed" in place.

After reverting this, we need to sort out how to tag this scenario appropriately, e.g. ``skipOnOcV10.0.10``


phil-davis:
``stable10`` revert is in PR #33813

individual-it (approved):

PR #33816, opened 2018-12-07
[do not merge] Test fix upload file with headers conflict resolution

phil-davis:


phil-davis:
All good - the 2 fails were stupid drone ``git fetch`` that sometimes (2 times in 110) cannot fetch the git repo.

PR #33879, opened 2018-12-13
[stable10] Backport of Use 'Public link' instead of file name in the …

sharidas:
…dialog

Use 'Public link' instead of file name while
creating public links. The dialog window opens
up with 'Public link' for the first link for
the file, 'Public link (2)' for second link of
file and so on.

## Description

Using file names in the public link share dialog window as `Link name` was causing issue when the file name is large. Especially when the file name size is more than 64 characters. This change set tries to set the `Link name` to `Public link` for the first link created for the file. For the next link created for the same file it would be `Public link (2)` and so on. For new files again the same pattern follows.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

This change is to fix the issue caused with `Link name` length limitation. The solution used here is to use `Public link` instead of file name.

## How Has This Been Tested?

- The test done as per https://github.com/owncloud/core/pull/33872. Also I here is the screenshot :

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/js/sharedialoglinklistview.js
+++ b/core/js/sharedialoglinklistview.js
@@ -131,9 +131,7 @@
 
 		_generateName: function() {
 			var index = 1;
-			var baseName = t('core', '{fileName} link', {
-				fileName: this.fileInfoModel.get('name')
-			}, null, {escape: false});
+			var baseName = t('core', 'Public link', null, {escape: false});
 			var name = baseName;
 
 			while (this.collection.findWhere({name: name})) {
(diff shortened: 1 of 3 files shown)


sharidas:
Original PR https://github.com/owncloud/core/pull/33872

PVince81 (approved):
:+1:

PR #33734, opened 2018-12-03
Fix 500 when uploading in to share without write/edit permissions #33416

IljaN:
Forward port of #33640

PR #33640, opened 2018-11-23
[stable10] Fix 500 when uploading in to share without write/edit permissions

IljaN:
## Description
Throw Forbidden exception if part-storage is not writable before write attempt.

## Related Issue

- Fixes #33416 

## How Has This Been Tested?
Acceptance test + manually

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33884, opened 2018-12-13
fix invalid test step

phil-davis:
that was introduced by https://github.com/owncloud/core/pull/33734/files#diff-ec180067294622fa5caf79bc2f9222c6

It happened because that PR passed CI some time ago, and acceptance test step text has been refactored since then.

Note: backport of that PR is still in progress. The fix here will be forced to happen in the backport PR #33640


PVince81 (approved):
:+1:

PR #32250, opened 2018-08-06
Feature/webdav locking frontend

DeepDiver1975:
## Description
Frontend part of #31651 

## Related Issue
- refs #31651

## How Has This Been Tested?
- manual testing
- js unit tests are missing

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [x] unit tests
- [x] properly update status on unlock
- [x] decide on information to be displayed on the details tab

PR #33888, opened 2018-12-13
Fix persistent lock retrieval by path

### LINKED ISSUE #33885: Don't apply persistent lock on received declined shares
From https://github.com/owncloud/core/pull/32250#issuecomment-434615887

### Steps

1. as user "sharer" create a folder called "parent" and inside that a folder called "subfolder"
1. from user "sharer" share a folder "parent" with the user "receiver"
1. as user "sharer" set a lock on the folder "parent"
1. as user "receiver" decline the share "parent" from the "Shared with you" page
1. as user "receiver" create a folder called "parent" and inside that a folder called "subfolder"

### Expected result
- the folder "parent/subfolder" of the user "receiver" not locked

### Actual result

- the folder "parent/subfolder" of the user "receiver" will be locked by sharer
- deleting that lock as "receiver" results in Unlock failed with status: 409
- deleting/renaming parent/subfolder" as "receiver" is not possible because its locked

### Version
discovered on https://github.com/owncloud/core/pull/32250

It is likely that whatever checks for locks is retrieving the full list of share mount points and does not filter it by state. Need to check what APIs are used there as there is already an existing way to only get visible mount points.

PVince81:
## Description
- Fix query to make it strictly search in the given storage.
- Fix query to prevent LIKE bleeding into unrelated folders (trailing slash issue).
- Added enforcement of Depth 0 or -1 as per RFC 4918 Section 9.10.3.
- Added many more unit tests to test the query from different angles, making sure that locks from parent and child paths are returned when the correct conditions are met and also not returned when not met.

## Related Issue
Fixes https://github.com/owncloud/core/issues/33885

## How Has This Been Tested?
- [x] Unit test
- [x] Reenable acceptance tests after https://github.com/owncloud/core/pull/32250 is merged

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

## Open tasks:

- [x] actually fix the issue

### DIFF
--- a/lib/private/Lock/Persistent/LockMapper.php
+++ b/lib/private/Lock/Persistent/LockMapper.php
@@ -24,6 +24,8 @@ use OCP\AppFramework\Db\Entity;
 use OCP\AppFramework\Db\Mapper;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\IDBConnection;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\Lock\Persistent\ILock;
 
 class LockMapper extends Mapper {
 	/** @var ITimeFactory */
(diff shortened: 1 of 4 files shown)


PVince81:
@DeepDiver1975 please carefully review. I added a lot of comments in the code to clarify the tested scenarios and the SQL query. Seems there were a few more bugs in there which I fixed (see PR description for detail)

lib/private/Lock/Persistent/Lock.php
```diff
@@ -39,6 +39,9 @@
  */
 class Lock extends Entity implements ILock {
 
+	const DEPTH_ZERO = 0;
```
PVince81: ok seems I missed that this is already on ILock

will change after the first review pass

lib/private/Lock/Persistent/LockMapper.php
```diff
@@ -34,54 +35,117 @@ public function __construct(IDBConnection $db, ITimeFactory $timeFactory) {
...
 					$query->expr()->neq('depth', $query->createPositionalParameter(0)),
-					$query->expr()->eq('path', $query->createPositionalParameter($currentPath))
+					// here we are assuming that the number of path sections will be less than 1000
+					$query->expr()->in('path', $query->createPositionalParameter($parentPaths, IQueryBuilder::PARAM_STR_ARRAY))
 				)
 			);
 		}
 
-		return $this->findEntities($query->getSQL(), $query->getParameters());
+		$query->andWhere($pathMatchClauses);
```
PVince81: I should probably also add "ORDER BY" as some databases will mess up the order and break the tests...

Order by path ?

PVince81:
- [ ] add sort order by path in query
- [ ] fix Lock constants

lib/private/Lock/Persistent/LockMapper.php
PVince81: ok, fix the unit test instead

PVince81:
- [ ] check how our endpoint behaves with depth = 1, 500 or 400 ?

DeepDiver1975 (approved):
Looks good - THX .. adjust remaining topics as discussed .... :+1:

PVince81:
Adjusted. Now hoping Oracle will pass...

PVince81:
acceptance test, something weird:
```
Updating .htaccess for proper rewrites
Could not set mail_domain on https://server/ocs/v2.php/apps/testing/api/v1/occ. Result:
'<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>500</statuscode>
  <message>Invalid query, please check the syntax. API specifications are here: http://www.freedesktop.org/wiki/Specifications/open-collaboration-services.</message>
 </meta>
 <data/>
</ocs>'
```

PVince81:
seems the id with Oracle is shifted. it almost looks like the insert id that was returned does not match the actual one returned by subsequent select

PVince81:
now sorting by id instead of fileid, let's see...

PVince81:
maybe the "testing" app did not enable for some reason?? @phil-davis @individual-it

phil-davis:
After PR #33956 has been merged to ``master`` then rebase this and the acceptance test issue will be fixed

PVince81:
Rebased. Fixed Oracle by properly sorting the expected output before asserting...

phil-davis:
> stable10: #33957

Removed backport-request label because backport is done.

PR #33894, opened 2018-12-14
Add regex for shared object file names

sharidas:
Add regular expression for shared object file name.
The file name can be for example 'a.so' or 'a.so.0.0.1'.
So this regular expression would help us to figure
out that file has 'so' extension too.

## Description

Adding regular expression to check if the file is of extension `filename.so.0.0.1`. This regular expression tries to check if there is a `.so` in the file name. And if so, it would use the extension `.so`. We cannot rely completely on `mimetypemapping.dist.json`, where it expects file with `filename.so` only.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

In this pr we try to solve the problem of shared object filenames which have multiple `.` and with the current mechanism, it would not be able to detect the extension for `filename.so.0.0.0.0.0.1`. With this pr the extension of `filenam.so.0.0.0.0.1` is taken as `.so`.

## How Has This Been Tested?

- Tested in the UI by uploading a file with a `.so` extension only and `.so.0.0.0.1`. Both were caught when firewall rule was added.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Files/Type/Detection.php
+++ b/lib/private/Files/Type/Detection.php
@@ -151,6 +151,26 @@ class Detection implements IMimeTypeDetector {
 	}
 
 	/**
+	 * Detect if the file is a shared object file. We detect this file
+	 * with a regex.
+	 *
+	 * @param string $fileName
+	 * @return bool true if file is a shared object else false
+	 */
+	private function detectSharedObjFileName($fileName) {
+		\preg_match("/(.*?)(\.so)([\.]*?.*)/i", $fileName, $match);
+		/**
+		 * 1. Exactly 4 indexes to match ( check the regex grouping above preg_match )
+		 * 2. Check if the filename has '.so'
+		 * 3. Check if the filname has extension after '.so', say .so.1. or .so.
+		 * That is anything like '.so.*'. This match is made in the 4th index.
+		 * No extension after '.so' is considered to be shared object file
+		 */
+		return (\count($match) === 4) && ($match[2] === '.so')
+			&& ($match[3] === '' || ($match[3] !== '' && strpos($match[3], '.') === 0));
+	}
+
+	/**
 	 * @return array
 	 */
 	public function getAllMappings() {
@@ -171,8 +191,12 @@ class Detection implements IMimeTypeDetector {
 		// note: leading dot doesn't qualify as extension
 		if (\strpos($fileName, '.') > 0) {
 			//try to guess the type by the file extension
-			$extension = \strtolower(\strrchr($fileName, '.'));
-			$extension = \substr($extension, 1); //remove leading .
+			if ($this->detectSharedObjFileName($fileName)) {
+				$extension = 'so';
+			} else {
+				$extension = \strtolower(\strrchr($fileName, '.'));
+				$extension = \substr($extension, 1); //remove leading .
+			}
 			return (isset($this->mimetypes[$extension], $this->mimetypes[$extension][0]))
 				? $this->mimetypes[$extension][0]
 				: 'application/octet-stream';
--- a/lib/private/Files/Type/Detection.php
+++ b/lib/private/Files/Type/Detection.php
@@ -167,7 +167,7 @@ class Detection implements IMimeTypeDetector {
 		 * No extension after '.so' is considered to be shared object file
 		 */
 		return (\count($match) === 4) && ($match[2] === '.so')
-			&& ($match[3] === '' || ($match[3] !== '' && strpos($match[3], '.') === 0));
+			&& ($match[3] === '' || ($match[3] !== '' && \strpos($match[3], '.') === 0));
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


lib/private/Files/Type/Detection.php
```diff
@@ -170,9 +170,19 @@ public function detectPath($path) {
...
-			//try to guess the type by the file extension
-			$extension = \strtolower(\strrchr($fileName, '.'));
-			$extension = \substr($extension, 1); //remove leading .
+			/**
+			 * try to guess the type by the file extension
+			 * First check if the match is for lets say
+			 * lib.so.0.0.1 ( shared object file ) and if not then
+			 * use the else logic.
+			 */
+			\preg_match("/(.*?)\.(so)(.*)/i", $fileName, $match);
```
PVince81: mind isolating this into a private function ?

lib/private/Files/Type/Detection.php
```diff
@@ -150,6 +150,18 @@ private function loadMappings() {
...
+	/**
+	 * Detect if the file is a shared object file. We detect this file
+	 * with a regex.
+	 *
+	 * @param string $fileName
+	 * @return bool true if file is a shared object else false
+	 */
+	private function detectSharedObjFileName($fileName) {
+		\preg_match("/(.*?)\.(so)\.(.*)/i", $fileName, $match);
+		return (\count($match) === 4) && ($match[2] === 'so');
```
PVince81: is it always 4 components or can it be less ?

```
lrwxrwxrwx 1 root root   20 Aug 20 21:31 /usr/lib64/libSDL-1.2.so.0 -> libSDL-1.2.so.0.11.4
-rwxr-xr-x 1 root root 431K Aug 20 21:31 /usr/lib64/libSDL-1.2.so.0.11.4
lrwxrwxrwx 1 root root   20 Nov 12 16:20 /usr/lib64/libSDL2-2.0.so.0 -> libSDL2-2.0.so.0.9.0
-rwxr-xr-x 1 root root 1.3M Nov 12 16:20 /usr/lib64/libSDL2-2.0.so.0.9.0
lrwxrwxrwx 1 root root   20 Aug 27 17:48 /usr/lib64/libSDL_gfx.so.15 -> libSDL_gfx.so.15.9.1
-rwxr-xr-x 1 root root  89K Aug 27 17:48 /usr/lib64/libSDL_gfx.so.15.9.1
lrwxrwxrwx 1 root root   25 Jun 10  2018 /usr/lib64/libSDL_image-1.2.so.0 -> libSDL_image-1.2.so.0.8.4
-rwxr-xr-x 1 root root  59K Jun 10  2018 /usr/lib64/libSDL_image-1.2.so.0.8.4
```

sharidas: If the match happens for the .so file, the `$match` will have 4 components. Else it will have zero.
So for example if the fileName is `testso.1.2`, the `$match` will have zero components. If the fileName is `test.so.1.2`, then `$match` will be `[0]=> string(11) "test.so.1.2" [1]=> string(4) "test" [2]=> string(3) ".so" [3]=> string(3) "1.2" `

sharidas:
Current status of the PR:
- Did an update to the regex. The update was done so that after `.so` it may have `.` followed by anything. Or the file name ends with `.so`.

lib/private/Files/Type/Detection.php
```diff
@@ -150,6 +150,18 @@ private function loadMappings() {
...
 
+	/**
+	 * Detect if the file is a shared object file. We detect this file
+	 * with a regex.
+	 *
+	 * @param string $fileName
+	 * @return bool true if file is a shared object else false
+	 */
+	private function detectSharedObjFileName($fileName) {
+		\preg_match("/(.*?)(\.so)[\.]*(.*)/i", $fileName, $match);
```
PVince81: will this also match ".so." and ".so.x" ?

for the excluded cases please add test cases in DetectionTest to confirm they are not matched

DeepDiver1975:
MacOS has .dylib extention - are we handling this file extension yet? - just popped up ...

sharidas:
> MacOS has .dylib extention - are we handling this file extension yet? - just popped up ...

Aah thanks for pointing it out. IMHO it does make sense to consider the missing `.dylib` extension.

lib/private/Files/Type/Detection.php
sharidas: The updated PR would match the cases requested. Updated the test with cases that does not match the regex.

sharidas:
Current state of the PR:
- Updated the PR to include `.dylib` extension by adding it to `resources/config/mimetypemapping.dist.json`
- Updated the regex to match the query posted https://github.com/owncloud/core/pull/33894#discussion_r242898430
- Updated the test to check the cases where the regex does not match.
- Added a doc in the code, it would help the future developers while going through the Detection code.

Hence I would appreciate a review for the updated changes. @PVince81

PVince81 (approved):
:+1:

PVince81:
@sharidas please backport for 10.1.1

sharidas:
Backport PR to stable10 https://github.com/owncloud/core/pull/34082

PR #33907, opened 2018-12-17
Correction to default apps folder in config.sample.php

phil-davis:
The default folder for apps is called ``apps``

## Description
Fix the documented default apps folder.

## Motivation and Context
The string was wrong in the comment.

## How Has This Been Tested?
N/A

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33912, opened 2018-12-17
[stable10] Correction to default apps folder in config.sample.php

phil-davis:
Backport #33907


mmattel (approved):
Thanks for catching that 👍

PR #33908, opened 2018-12-17
Implement vendor-bin for owncloud-codestyle phan php_codesniffer phpstan

phil-davis:
## Description
- move ``owncloud/coding-standard`` ``squizlabs/php_codesniffer`` ``phan/phan`` and ``phpstan/phpstan`` into ``vendor-bin`` like has been done in apps
- adjust ``Makefile`` for that
- ``composer update``

Note: reminder to self, ``stable10`` backport will need run code style and phan in appropriate PHP combinations for PHP 5.6, 7.*

## Related Issue
Part of issue #33902 

## Motivation and Context
Be consistent with what is happening in apps.

## How Has This Been Tested?
Local runs of:
```
make test-php-style
make test-php-phan
make test-php-phpstan
```

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done) #33917

PR #33917, opened 2018-12-17
[stable10] Implement vendor-bin for owncloud-codestyle phan php_codesniffer phpstan

phil-davis:
Backport #33908


phil-davis:
Note: this backport "by default" makes ``phpstan`` available in ``stable10``. So I put the ``phpstan.neon`` from ``master`` in here also, so that there is a consistent setup of ``phpstan`` available for whoever uses it.

vendor-bin/owncloud-codestyle/composer.json
```diff
@@ -0,0 +1,11 @@
+{
+    "config" : {
+      "platform": {
+        "php": "5.6.37"
```
patrickjahns: what is the reason for pinning the php version here?

phil-davis: We pinned it in all the apps, so I have pinned it here in ``stable10`` that supports PHP 5.6

Because this is just a dev tool, then maybe it does not need to be pinned anywhere? ``composer`` will find whatever dependencies based on the PHP version of the current system. They are not distributed anywhere in a build/tarball... so it will not matter if dependency versions are used that do not work with PHP 5.6. (And that also brings up the issue of the developer being able to clean ``vendor-bin`` if they are switching forward/back between PHP versions)

phil-davis:
The proper up-to-date backport was done in #34182

PR #33910, opened 2018-12-17
Run unittest that needs db in drone also

skshetry:
## Description

Currently, we provide database access for unit tests if it is running on Travis. But, we have migrated most of our CI to drone. This keeps that, but also adds the check if the environment is `DRONE`.

## Motivation and Context

When standardizing CI on owncloud/files_ldap_home#6 , unit tests required access to the `db` which is not provided unless the tests specify them that they need DB in the annotation.
Eg:
``` php
/**
* class ABCDTest
* @group DB
/**
class ABCDTest extends TestCase { }
```

While running on drone, it was always complaining that it needs DB(not that it run locally). I thought it should have worked anyway and dug deeper on the code where the exception was raised from. There I saw `TRAVIS` being checked, but not the `DRONE`. 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #33933, opened 2018-12-19
Provide database access on CI checking for `CI` variables

### LINKED ISSUE #33915: Sort out IsDatabaseAccessAllowed env variables
``tests/lib/TestCase.php`` currently checks for env vars TRAVIS or DRONE in order to return ``true``

Actually the env var ``CI`` probably exists, and is a better one to use. See comment https://github.com/owncloud/core/pull/33910#issuecomment-447779552

Investigate and sort it out.

skshetry:
## Description

Previously, we checked for `TRAVIS` env variable and provide database access if it's running on that CI.  With #33910, we also check for `DRONE` env variable. As @patrickjahns [#33910:comment](https://github.com/owncloud/core/pull/33910#issuecomment-447779552), pointed out, it'd be better if we could just check for `CI` variable. 

But, I couldn't find docs related to that for drone. It's set to `CI=true` on [travis](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables) and on [drone](https://docs.tea-ci.org/usage/variables/) for background compatibility,  but I couldn't find it on official docs for drone.

Let's see if CI passes.

## Related Issue

Fixes #33915

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/lib/TestCase.php
+++ b/tests/lib/TestCase.php
@@ -454,7 +454,7 @@ abstract class TestCase extends BaseTestCase {
 	private function IsDatabaseAccessAllowed() {
 		// on travis-ci.org and drone, we allow database access in any case - otherwise
 		// this will break all apps right away
-		if (\getenv('TRAVIS') == true || \getenv('DRONE') == true) {
+		if (\getenv('CI') !== false) {
 			return true;
 		}
 		$annotations = $this->getAnnotations();


phil-davis (dismissed):
This should work. I will look for the drone docs.

phil-davis:
https://github.com/drone/docs/blob/master/content/usage/reference/environment.md
says that ``CI=drone`` - so I guess just checking if ``CI`` env var exists is good.

tests/lib/TestCase.php
```diff
@@ -454,7 +454,7 @@ protected function isFileLocked($view, $path, $type, $onMountPoint = false) {
 	private function IsDatabaseAccessAllowed() {
 		// on travis-ci.org and drone, we allow database access in any case - otherwise
 		// this will break all apps right away
-		if (\getenv('TRAVIS') == true || \getenv('DRONE') == true) {
+		if (\getenv('CI') == true) {
```
phil-davis: The value is probably ``drone`` or ``travis`` or some other string depending on the CI system being used.
I guess we just check for ``(\getenv('CI') !== false)``

skshetry: But, this one seem to work. And, the doc doesn't seem to be changed since Jan.

phil-davis: This probably works because in PHP
```
"drone" == true
"travis" == true
```
so whatever the environment variable is set to, apart from "0", it will b ``== true``
I hope nobody makes a CI project called "0"!

It is safer to check ``!== false`` because PHP returns boolean ``false`` if the env variable does not exist.

skshetry: @phil-davis, did as you recommended. Please review again

phil-davis (approved):

phil-davis:
I have asked @PVince81 @DeepDiver1975 to merge this. It failed project codecov for some reason.

skshetry:
Backport for `stable10` on  #33942

PR #33960, opened 2018-12-21
[stable10] Remove core/l10n from release build

PVince81:
It contains a perl script and a ".tx" file, not needed for the release.

I saw this folder also existing in 10.0.10, maybe we never noticed...


DeepDiver1975 (approved):

PVince81:
- [ ] port to master

PR #33992, opened 2018-12-31
 Improve Files UX error handling

mrow4a:
This fixes a bug in which when e.g. Forbidden is thrown for file when "Download" is clicked, the loading wheel is still spinning

This
faileddownload

Now will be 
handledfaileddownload

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesPlugin.php
@@ -152,8 +152,9 @@ class FilesPlugin extends ServerPlugin {
 		$this->server->on('propPatch', [$this, 'handleUpdateProperties']);
 		$this->server->on('afterBind', [$this, 'sendFileIdHeader']);
 		$this->server->on('afterWriteContent', [$this, 'sendFileIdHeader']);
-		$this->server->on('afterMethod:GET', [$this,'httpGet']);
+		$this->server->on('afterMethod:GET', [$this, 'httpGet']);
 		$this->server->on('afterMethod:GET', [$this, 'handleDownloadToken']);
+		$this->server->on('exception', [$this, 'handleDownloadFailure']);
 		$this->server->on('afterResponse', function ($request, ResponseInterface $response) {
 			$body = $response->getBody();
 			if (\is_resource($body)) {
@@ -192,9 +193,20 @@ class FilesPlugin extends ServerPlugin {
 	}
 
 	/**
+	 * This sets a cookie to be able to recognize the failure of the download
+	 *
+	 * @param \Exception $ex
+	 */
+	public function handleDownloadFailure(\Exception $ex) {
+		$queryParams = $this->server->httpRequest->getQueryParameters();
+
+		if (isset($queryParams['downloadStartSecret'])) {
+			$this->setSecretCookie('ocDownloadStarted', '-1');
+		}
+	}
+
+	/**
 	 * This sets a cookie to be able to recognize the start of the download
-	 * the content must not be longer than 32 characters and must only contain
-	 * alphanumeric characters
 	 *
 	 * @param RequestInterface $request
 	 * @param ResponseInterface $response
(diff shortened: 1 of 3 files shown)


mrow4a:
I also noticed that BrowserErrorPagePlugin works incorrectly (https://github.com/owncloud/core/blob/master/apps/dav/lib/Files/BrowserErrorPagePlugin.php#L98), showing such ugly page as it cannot load scripts

screenshot at dec 31 14-00-46

apps/files/js/files.js
```diff
@@ -306,12 +306,23 @@
...
-						return false;
-					} else {
+					var downloadSuccess = OC.Util.isCookieSetToValue('ocDownloadStarted', randomToken);
+					var downloadFailed = OC.Util.isCookieSetToValue('ocDownloadFailed', randomToken) &&
+						!OC.Util.isCookieSetToValue('ocDownloadStarted', randomToken);
+
+					if (downloadFailed || downloadSuccess) {
+						if (downloadFailed) {
+							OC.Notification.show(t('files',
+								'You cannot download this file!'),
```
PVince81: what's the usual cause ? file was deleted from disk but still in file cache ?

depending on this we could tweak the message

mrow4a: @PVince81 E.g. that user tries to download it using some URI/button, but one of the apps prevents it. It should show nice message instead of ugly page

mrow4a: But I am still not sure how to solve https://github.com/owncloud/core/pull/33992#issuecomment-450643425

apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -191,29 +192,31 @@ public function checkMove($source, $destination) {
...
+	 * This sets a cookie to be able to recognize the failure of the download
+	 *
+	 * @param \Exception $ex
+	 */
+	public function handleDownloadFailure(\Exception $ex) {
+		$queryParams = $this->server->httpRequest->getQueryParameters();
+
+		if (isset($queryParams['downloadStartSecret'])) {
+			$token = $queryParams['downloadStartSecret'];
+			$this->setSecretCookie('ocDownloadFailed', $token);
```
PVince81: I wonder if we could reuse the other cookie but add a special value ?

PVince81 (changes requested):
also please fix the JS tests

mrow4a:
@PVince81 do you know what might be wrong with these "ugly pages" and `Refused to load X.js because it appears` ?

apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -154,6 +154,7 @@ public function initialize(\Sabre\DAV\Server $server) {
 		$this->server->on('afterWriteContent', [$this, 'sendFileIdHeader']);
 		$this->server->on('afterMethod:GET', [$this,'httpGet']);
 		$this->server->on('afterMethod:GET', [$this, 'handleDownloadToken']);
+		$this->server->on('exception', [$this, 'handleDownloadFailure']);
```
mrow4a: To simulate exception, one can just add in FilesPlugin 
```
...
		$this->server->on('method:GET', [$this, 'throwEx'], 90);
...

	public function throwEx() {
		throw new Forbidden("cannot download");
	}
	
```

mrow4a:
This can be closed in favour of not showing download button in UX when special dav flag is enabled. Reopen if this wont be easily possible.

PVince81:
I'll first reopen to investigate

PVince81:
hiding the button is a good idea but won't cover the case where the permissions changed while the web UI is already rendered. we need both:

- [ ] fix forbidden page
- [ ] hide button when permission not set

where can we read the permission from ? @mrow4a

mrow4a:
@PVince81 I already included this in https://github.com/owncloud/core/pull/33994/commits/ddfe043f49cd8ec5bacbe329cc3e2f7f770986b5

PVince81:
Steps to reproduce without the richdocuments app on master:

1. Enable firewall app
1. Upload a file "test.txt"
1. Apply a new system tag "forbidden"
1. Create a firewall rule to block based on this tag
1. Try downloading the file

Expected: proper error handling.
Actual: forbidden page.

As the firewall is a rather common app, I wonder when this problem started happening.

PVince81:
we might be missing CSP rules to properly display the "Forbidden" page:
image

PVince81:
also happens on OC 10.0.5... hmmm....

PVince81:
ok, seems the BrowserErrorPagePlugin is missing CSP rules, but is also not properly forwarding the exception message...

PVince81:
PR to fix the error page specifically: https://github.com/owncloud/core/pull/34798

PVince81:
@m

mrow4a:
@PVince81 this PR is needed to fix spinning wheel problem, even including #34798 when error page appears, and you "go back to previous page", javascript is broken with exception.

Firewall app is also affected.

mrow4a:
@PVince81 @IljaN just tested and it work, can you review and merge?

mrow4a:
@PVince81 you want to take this over? This is just an improvement to the UX

PVince81 (approved):
:+1:

PVince81:
stable10: https://github.com/owncloud/core/pull/34886

Issue #33458, opened 2018-11-08
Support for additional arbitrary advanced share permissions


PVince81:
Add the ability to extend the current share permissions model and add more permissions as defined by apps.

### 1) Reusing and extending the existing permissions field

The oc_filecache table has a field "permissions" and also oc_share.permissions for sharing permissions.
The value is a bitmask.
We could reuse additional bits.

- Additional implications on the storage layer: https://github.com/owncloud/core/issues/33458#issuecomment-439391748. Could be a good thing though as we'd be able to let storages define the additional perms.

#### Pros

- No changes in data model, the permissions are already exposed everywhere with no extra efforts.

#### Cons

- Would require apps to "reserve" certain bits for their own additional permissions.
- Or needs core to predefine some bits for certain permissions and let apps implement the actual logic.

### 1b) Reusing and extending the existing permissions field

Like 1) but store the permissions in a new field "oc_share.extraPermissions".

#### Pros

- Minor addition to current share model
- No storage implications

#### Cons

- More wiring up required in various layers.

### 2) Use extra Webdav properties

The table "oc_properties" contains a list of arbitrary Webdav properties that can be set with PROPPATCH and retrieved with PROPFIND. They are attached to files by file id.

An app could define its own properties to represent additional permissions there.

When the time comes to enforce permissions, the app would need to read the permission from there manually.

Could also provide a permissions manager in core that hides this and exposes it via a public PHP API.

#### Pros

- Ability to reuse an existing table.

#### Cons

- oc_properties is currently on the DAV layer. To access it from an app somehow violates our API principles. Ideally this table should be managed on View level. Also see https://github.com/owncloud/core/issues/33413#issuecomment-435800146,

### 3) New table for additional permissions

Create a new table "oc_extra_permissions" with fields "fileid", "app_id", "permission_id", "active".
Need to keep this table in sync with "oc_filecache.fileid" for example with foreign keys or a background job to clear orphaned entries.
Provide a permissions manager in core that manages the permissions table and exposes a public PHP API for managing the permissions in there.

#### Pros
- cleaner approach, proper data structures

#### Cons
- extra table (the other approaches didn't have one)

### 4) Use collaborative tags

An app could define a set of collaboratve tags which would define the permissions for everyone.

#### Pros

No core changes needed. Apps read the tags and decide how to apply restrictions.

#### Cons

Restriction is for all users, not per user.

## Frontend approaches

Need a way for apps to be able to inject UI components.
[...]

PR #33994, opened 2018-12-31
View-only dav plugin using IShare attributes

mrow4a:
- [x] Implement extra share permissions with IShare from Approach 3 here - https://github.com/owncloud/core/issues/33458
- [x] Implement view-only plugin using extra share permissions to restrict view-only permission for read-only files.
- [x] Allow apps to register extra share permissions in frontend
- [x] Clean the code
- [x] Add unit tests
- [x] Try JSON type in DB @DeepDiver1975 
- [x] Make sure PR works with custom groups

Would be also good to merge UX for Forbidden exception with view-only https://github.com/owncloud/core/pull/33992

@PVince81 @DeepDiver1975 @pmaier1

### DIFF
--- a/apps/dav/lib/DAV/ViewOnlyPlugin.php
+++ b/apps/dav/lib/DAV/ViewOnlyPlugin.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * @author Piotr Mrowczynski piotr@owncloud.com
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\DAV;
+
+use OCA\DAV\Connector\Sabre\Exception\Forbidden;
+use OCA\DAV\Connector\Sabre\File as DavFile;
+use OCA\DAV\Meta\MetaFile;
+use OCP\Files\FileInfo;
+use OCP\Files\NotFoundException;
+use OCP\ILogger;
+use Sabre\DAV\Server;
+use Sabre\DAV\ServerPlugin;
+use Sabre\HTTP\RequestInterface;
+use Sabre\DAV\Exception\NotFound;
+
+/**
+ * Sabre plugin for restricting file share receiver download:
+ */
+class ViewOnlyPlugin extends ServerPlugin {
+
+	/** @var Server $server */
+	private $server;
+
+	/** @var ILogger $logger */
+	private $logger;
+
+	/**
+	 * @param ILogger $logger
+	 */
+	public function __construct(ILogger $logger) {
+		$this->logger = $logger;
+	}
+
+	/**
+	 * This initializes the plugin.
+	 *
+	 * This function is called by Sabre\DAV\Server, after
+	 * addPlugin is called.
+	 *
+	 * This method should set up the required event subscriptions.
+	 *
+	 * @param Server $server
+	 * @return void
+	 */
+	public function initialize(Server $server) {
+		$this->server = $server;
+		//priority 90 to make sure the plugin is called before
+		//Sabre\DAV\CorePlugin::httpGet
+		$this->server->on('method:GET', [$this, 'checkViewOnly'], 90);
+	}
+
+	/**
+	 * Disallow download via DAV Api in case file being received share
+	 * and having special permission
+	 *
+	 * @param RequestInterface $request request object
+	 * @return boolean
+	 * @throws Forbidden
+	 * @throws NotFoundException
+	 */
+	public function checkViewOnly(
+		RequestInterface $request
+	) {
+		$path = $request->getPath();
+
+		try {
+			$davNode = $this->server->tree->getNodeForPath($path);
+			if (!($davNode instanceof DavFile || $davNode instanceof MetaFile)) {
+				return true;
+			}
+			// Restrict view-only to nodes which are shared
+			$node = $davNode->getNode();
+			if (!$node instanceof FileInfo) {
+				return true;
+			}
+
+			$storage = $node->getStorage();
+			// using string as we have no guarantee that "files_sharing" app is loaded
+			if (!$storage->instanceOfStorage('OCA\Files_Sharing\SharedStorage')) {
+				return true;
+			}
+			// Extract extra permissions
+			/** @var \OCA\Files_Sharing\SharedStorage $storage */
+			$share = $storage->getShare();
+
+			// Check if read-only and on whether permission can download is both set and disabled.
+			$canDownload = $share->getAttributes()->getAttribute('permissions', 'download');
+			if ($canDownload !== null && !$canDownload) {
+				throw new Forbidden('Access to this resource has been denied because it is in view-only mode.');
+			}
+		} catch (NotFound $e) {
+			$this->logger->warning($e->getMessage());
+		}
+
+		return true;
+	}
+}
--- a/apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
+++ b/apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
@@ -1,0 +1,128 @@
+<?php
+/**
+ * @author Piotr Mrowczynski piotr@owncloud.com
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+namespace OCA\DAV\Tests\unit\DAV;
+
+use OCA\DAV\DAV\ViewOnlyPlugin;
+use OCA\Files_Sharing\SharedStorage;
+use OCA\DAV\Connector\Sabre\File as DavFile;
+use OCP\Files\FileInfo;
+use OCP\Files\Storage\IStorage;
+use OCP\ILogger;
+use OCP\Share\IAttributes;
+use OCP\Share\IShare;
+use Sabre\DAV\Server;
+use Sabre\DAV\Tree;
+use Test\TestCase;
+use Sabre\HTTP\RequestInterface;
+use OCA\DAV\Connector\Sabre\Exception\Forbidden;
+
+class ViewOnlyPluginTest extends TestCase {
+
+	/** @var ViewOnlyPlugin */
+	private $plugin;
+	/** @var Tree | \PHPUnit\Framework\MockObject\MockObject */
+	private $tree;
+	/** @var RequestInterface | \PHPUnit\Framework\MockObject\MockObject */
+	private $request;
+
+	public function setUp() {
+		$this->plugin = new ViewOnlyPlugin(
+			$this->createMock(ILogger::class)
+		);
+		$this->request = $this->createMock(RequestInterface::class);
+		$this->tree = $this->createMock(Tree::class);
+
+		$server = $this->createMock(Server::class);
+		$server->tree = $this->tree;
+
+		$this->plugin->initialize($server);
+	}
+
+	public function testCanGetNonDav() {
+		$this->request->expects($this->once())->method('getPath')->willReturn('files/test/target');
+		$this->tree->method('getNodeForPath')->willReturn(null);
+
+		$this->assertTrue($this->plugin->checkViewOnly($this->request));
+	}
+
+	public function testCanGetNonFileInfo() {
+		$this->request->expects($this->once())->method('getPath')->willReturn('files/test/target');
+		$davNode = $this->createMock(DavFile::class);
+		$this->tree->method('getNodeForPath')->willReturn($davNode);
+
+		$davNode->method('getNode')->willReturn(null);
+
+		$this->assertTrue($this->plugin->checkViewOnly($this->request));
+	}
+
+	public function testCanGetNonShared() {
+		$this->request->expects($this->once())->method('getPath')->willReturn('files/test/target');
+		$davNode = $this->createMock(DavFile::class);
+		$this->tree->method('getNodeForPath')->willReturn($davNode);
+
+		$fileInfo = $this->createMock(FileInfo::class);
+		$davNode->method('getNode')->willReturn($fileInfo);
+
+		$storage = $this->createMock(IStorage::class);
+		$fileInfo->method('getStorage')->willReturn($storage);
+		$storage->method('instanceOfStorage')->with(SharedStorage::class)->willReturn(false);
+
+		$this->assertTrue($this->plugin->checkViewOnly($this->request));
+	}
+
+	public function providesDataForCanGet() {
+		return [
+			// has attribute permissions-download enabled - can get file
+			[ $this->createMock(FileInfo::class), true, true],
+			// has no attribute permissions-download - can get file
+			[ $this->createMock(FileInfo::class), null, true],
+			// has attribute permissions-download disabled- cannot get the file
+			[ $this->createMock(FileInfo::class), false, false],
+		];
+	}
+
+	/**
+	 * @dataProvider providesDataForCanGet
+	 */
+	public function testCanGet($nodeInfo, $attrEnabled, $expectCanDownloadFile) {
+		$this->request->expects($this->once())->method('getPath')->willReturn('files/test/target');
+
+		$davNode = $this->createMock(DavFile::class);
+		$this->tree->method('getNodeForPath')->willReturn($davNode);
+
+		$davNode->method('getNode')->willReturn($nodeInfo);
+
+		$storage = $this->createMock(SharedStorage::class);
+		$share = $this->createMock(IShare::class);
+		$nodeInfo->method('getStorage')->willReturn($storage);
+		$storage->method('instanceOfStorage')->with(SharedStorage::class)->willReturn(true);
+		$storage->method('getShare')->willReturn($share);
+
+		$extAttr = $this->createMock(IAttributes::class);
+		$share->method('getAttributes')->willReturn($extAttr);
+		$extAttr->method('getAttribute')->with('permissions', 'download')->willReturn($attrEnabled);
+
+		if (!$expectCanDownloadFile) {
+			$this->expectException(Forbidden::class);
+		}
+		$this->plugin->checkViewOnly($this->request);
+	}
+}
--- a/apps/files/download.php
+++ b/apps/files/download.php
@@ -39,6 +39,17 @@ if (!\OC\Files\Filesystem::file_exists($filename)) {
 	exit;
 }
 
+//Dispatch an event to see if any apps have problem with download
+$event = new \Symfony\Component\EventDispatcher\GenericEvent(null, ['path' => $filename]);
+OC::$server->getEventDispatcher()->dispatch('file.beforeGetDirect', $event);
+if ($event->hasArgument('errorMessage')) {
+	\header("HTTP/1.0 403 Forbidden");
+	$tmpl = new OCP\Template('', '403', 'guest');
+	$tmpl->assign('file', $filename);
+	$tmpl->printPage();
+	exit;
+}
+
 $ftype=\OC::$server->getMimeTypeDetector()->getSecureMimeType(\OC\Files\Filesystem::getMimeType($filename));
 
 \header('Content-Type:'.$ftype);
--- a/apps/files_sharing/lib/AppInfo/Application.php
+++ b/apps/files_sharing/lib/AppInfo/Application.php
@@ -159,7 +159,8 @@ class Application extends App {
 				$c->getServer()->getUrlGenerator(),
 				$c->getServer()->getEventDispatcher(),
 				$c->getServer()->getShareManager(),
-				$c->query(NotificationPublisher::class)
+				$c->query(NotificationPublisher::class),
+				$c->getServer()->getUserSession()
 			);
 		});
 
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -222,6 +222,11 @@ class Share20OcsController extends OCSController {
 
 		$result['mail_send'] = $share->getMailSend() ? 1 : 0;
 
+		$result['attributes'] = null;
+		if ($attributes = $share->getAttributes()) {
+			$result['attributes'] =  \json_encode($attributes->toArray());
+		}
+
 		return $result;
 	}
 
@@ -498,6 +503,8 @@ class Share20OcsController extends OCSController {
 
 		$share->setShareType($shareType);
 		$share->setSharedBy($this->userSession->getUser()->getUID());
+
+		$share = $this->setShareAttributes($share, $this->request->getParam('attributes', null));
 
 		try {
 			$share = $this->shareManager->createShare($share);
@@ -869,6 +876,8 @@ class Share20OcsController extends OCSController {
 			return new Result(null, 400, $this->l->t('Cannot remove all permissions'));
 		}
 
+		$share = $this->setShareAttributes($share, $this->request->getParam('attributes', null));
+
 		try {
 			$share = $this->shareManager->updateShare($share);
 		} catch (\Exception $e) {
@@ -1196,6 +1205,27 @@ class Share20OcsController extends OCSController {
 
 			$share = $this->shareManager->getShareById('ocFederatedSharing:' . $id, $recipient);
 		}
+
+		return $share;
+	}
+
+	/**
+	 * @param IShare $share
+	 * @param string[][]|null $formattedShareAttributes
+	 * @return IShare modified share
+	 */
+	private function setShareAttributes(IShare $share, $formattedShareAttributes) {
+		$newShareAttributes = $this->shareManager->newShare()->newAttributes();
+		if ($formattedShareAttributes !== null) {
+			foreach ($formattedShareAttributes as $formattedAttr) {
+				$newShareAttributes->setAttribute(
+					$formattedAttr['scope'],
+					$formattedAttr['key'],
+					(bool) \json_decode($formattedAttr['enabled'])
+				);
+			}
+		}
+		$share->setAttributes($newShareAttributes);
 
 		return $share;
 	}
--- a/apps/files_sharing/lib/Hooks.php
+++ b/apps/files_sharing/lib/Hooks.php
@@ -28,6 +28,7 @@ namespace OCA\Files_Sharing;
 use OC\Files\Filesystem;
 use OCP\IURLGenerator;
 use OCP\Files\IRootFolder;
+use OCP\IUserSession;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use OCP\Share\IShare;
 use Symfony\Component\EventDispatcher\GenericEvent;
@@ -46,6 +47,11 @@ class Hooks {
 	private $rootFolder;
 
 	/**
+	 * @var IUserSession|null
+	 */
+	private $userSession;
+
+	/**
 	 * @var EventDispatcher
 	 */
 	private $eventDispatcher;
@@ -60,13 +66,25 @@ class Hooks {
 	 */
 	private $notificationPublisher;
 
+	/**
+	 * Hooks constructor.
+	 *
+	 * @param IRootFolder $rootFolder
+	 * @param IURLGenerator $urlGenerator
+	 * @param EventDispatcher $eventDispatcher
+	 * @param \OCP\Share\IManager $shareManager
+	 * @param NotificationPublisher $notificationPublisher
+	 * @param IUserSession|null $userSession
+	 */
 	public function __construct(
 		IRootFolder $rootFolder,
 		IUrlGenerator $urlGenerator,
 		EventDispatcher $eventDispatcher,
 		\OCP\Share\IManager $shareManager,
-		NotificationPublisher $notificationPublisher
+		NotificationPublisher $notificationPublisher,
+		$userSession
 	) {
+		$this->userSession = $userSession;
 		$this->rootFolder = $rootFolder;
 		$this->urlGenerator = $urlGenerator;
 		$this->eventDispatcher = $eventDispatcher;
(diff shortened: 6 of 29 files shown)


mrow4a:
In PR for view-only plugin & extra share permission, I initially took an approach (https://github.com/owncloud/core/issues/33458) of extending `oc_share` table with new string column (json) `extra_share_permissions`. However, I think of going with the approach of external tables. These are pros&cons:

Extended `oc_share` table:
- currently adds new function `setExtraPermissions/getExtraPermissions` only to Share20\Share class - https://github.com/owncloud/core/pull/33994/files#diff-a19a495d2d4616d9b451f849126013a4. IShare is not touched as it would require all apps to adjust.
- easy to implement, no additional queries needed (single query retrieves the information). Expensive as each share need to reserve string space in DB
- requires to add additional API functions to IShare?, or keep with appraoch of `if instanceof Share20\Share` https://github.com/owncloud/core/pull/33994/files#diff-faddd9984dfaf09a2b85d8dc4867207aR887

New `extra_share_permissions` table:
- ExtraShareManager used to retrieve extra share permissions for shares
-`IShare` or `Share/IManager` not extended, new classes used (`Share/ExtraPermissions/IManager`)
- new table, reserves space in DB only for shares which have extra share permissions (by full-share-id/share-id).
- might need some additional queries to DB to retrieve extra share permissions. 

@DeepDiver1975 @PVince81 do you have any input?

core/Migrations/Version20181220085457.php
```diff
@@ -13,7 +13,7 @@ class Version20181220085457 implements ISchemaMigration {
 	public function changeSchema(Schema $schema, array $options) {
 		$prefix = $options['tablePrefix'];
 
-		// FIXME: probably should be another table - temporarly use JSON field for the sake of POC
+		// FIXME: Should Type::TEXT or Type::JSON be used?
```
DeepDiver1975: give it a try with the JSON type as discussed and let's see ...

lib/private/Share20/DefaultShareProvider.php
```diff
@@ -141,11 +142,16 @@ public function create(\OCP\Share\IShare $share) {
...
+			$qb->setParameter('itemType', \OCP\Share::NODE_TYPE_FILE);
 		} else {
-			$qb->setParameter('itemType', 'folder');
+			$qb->setParameter('itemType', \OCP\Share::NODE_TYPE_FOLDER);
 		}
 
+		// No extra share permissions are set during share create
+		// Setting extra share permissions is only allowed at update
+		$shareExtraPermissions = new ExtraSharePermissions();
+		$qb->setValue('extra_permissions', $qb->createNamedParameter($shareExtraPermissions->serialize()));
```
DeepDiver1975: the table defines a default value - no need to set anything here from my pov ....

lib/private/Share20/ExtraPermissions/Permissions.php
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+namespace OC\Share20\ExtraPermissions;
+
+use OCP\Share\ExtraPermissions\IPermissions;
+
+class Permissions implements IPermissions {
```
DeepDiver1975: I don't see the need of this class. Just use an array with the app as key and the permission as value

lib/private/Share20/ExtraPermissions/Permission.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ */
+namespace OC\Share20\ExtraPermissions;
+
+use OCP\Share\ExtraPermissions\IPermission;
+
+class Permission implements IPermission {
+
+	/** @var string */
+	private $app;
```
DeepDiver1975: why app again? app is the key in the permissions array ....

DeepDiver1975 (changes requested):

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,114 @@
+<?php
+/**
+ * @author Piotr Mrowczynski piotr@owncloud.com
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
```
DeepDiver1975: 2019

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,114 @@
...
+ */
+class ViewOnlyPlugin extends ServerPlugin {
+
+	/** @var \Sabre\DAV\Server $server */
+	private $server;
+
+	/**
+	 * ViewOnlyPlugin plugin
+	 */
+	public function __construct() {
```
DeepDiver1975: remove if not needed

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,114 @@
...
+			// Extract extra permissions
+			/** @var \OCA\Files_Sharing\SharedStorage $storage */
+			$share = $storage->getShare();
+
+			// Check if read-only and on whether permission can download is both set and disabled.
+			$canDownload = $share->getExtraPermissions()->getPermission('dav', 'can-download');
+			if (!$node->isUpdateable() && $canDownload !== null && !$canDownload) {
+				throw new Forbidden('File is in secure-view mode and cannot be directly downloaded.');
+			}
+		} catch (NotFound $e) {
```
DeepDiver1975: log them all

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			// Extract extra permissions
+			/** @var \OCA\Files_Sharing\SharedStorage $storage */
+			$share = $storage->getShare();
+
+			// Check if read-only and on whether permission can download is both set and disabled.
+			$canDownload = $share->getExtraPermissions()->getPermission('dav', 'can-download');
+			if (!$node->isUpdateable() && $canDownload !== null && !$canDownload) {
+				throw new Forbidden('File is in secure-view mode and cannot be directly downloaded.');
```
DeepDiver1975: translation necessary? @PVince81 do we translate exception messages in dav plugins at all?

apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
```diff
@@ -0,0 +1,137 @@
...
+	/** @var ViewOnlyPlugin */
+	private $plugin;
+	/** @var Tree | \PHPUnit_Framework_MockObject_MockObject */
+	private $tree;
+	/** @var RequestInterface | \PHPUnit_Framework_MockObject_MockObject */
+	private $request;
+
+	public function setUp() {
+		$this->plugin = new ViewOnlyPlugin();
+		$this->request = $this->getMockBuilder('Sabre\HTTP\RequestInterface')->getMock();
```
DeepDiver1975: use ::class

DeepDiver1975: use createMock

apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+		$server = $this->getMockBuilder('Sabre\DAV\Server')->getMock();
+		$server->tree = $this->tree;
+
+		$this->plugin->initialize($server);
+	}
+
+	public function testCanGetNonDav() {
+		$this->request->expects($this->exactly(1))->method('getPath')->willReturn('files/test/target');
+		$this->tree->method('getNodeForPath')->willReturn('this is not Dav File');
```
DeepDiver1975: getNodeForPath will not return this string - right?

apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+		$this->assertTrue($this->plugin->checkViewOnly($this->request));
+	}
+
+	public function testCanGetNonFileFolder() {
+		$this->request->expects($this->exactly(1))->method('getPath')->willReturn('files/test/target');
+		$davNode = $this->createMock(DavFile::class);
+		$this->tree->method('getNodeForPath')->willReturn($davNode);
+
+		$davNode->method('getNode')->willReturn('this is not File or Folder');
```
DeepDiver1975: get node is returning a node instance - not a string

apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
```diff
@@ -0,0 +1,137 @@
...
+		$davNode->method('getNode')->willReturn($node);
+
+		$storage = $this->createMock(IStorage::class);
+		$node->method('getStorage')->willReturn($storage);
+		$storage->method('instanceOfStorage')->with(SharedStorage::class)->willReturn(false);
+
+		$this->assertTrue($this->plugin->checkViewOnly($this->request));
+	}
+
+	public function nodeReturns() {
```
DeepDiver1975: name proposal: providesDataForCanGet

apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
```diff
@@ -0,0 +1,137 @@
...
+			// has no extra permission can download, and read-only is set - cannot get the file
+			[ $this->createMock(File::class), false, false, false],
+			[ $this->createMock(Folder::class), false, false, false],
+		];
+	}
+
+	/**
+	 * @dataProvider nodeReturns
+	 */
+	public function testCanGet($node, $canDownloadPerm, $isUpdatable, $expected) {
```
DeepDiver1975: expected argument is usually always the first

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -219,6 +221,23 @@ protected function formatShare(IShare $share, $received = false) {
 
 		$result['mail_send'] = $share->getMailSend() ? 1 : 0;
 
+		$extraPermissions = $share->getExtraPermissions();
+		if ($extraPermissions !== null) {
```
DeepDiver1975: looks like code duplication from below - line 1201++

apps/files_sharing/tests/ApiTest.php
```diff
@@ -1000,21 +1000,29 @@ public function testUpdateShare() {
...
 		$share2 = $this->shareManager->newShare();
 		$share2->setNode($node1)
 			->setSharedBy(self::TEST_FILES_SHARING_API_USER1)
 			->setShareType(Share::SHARE_TYPE_LINK)
 			->setPermissions(1);
 		$share2 = $this->shareManager->createShare($share2);
+		$this->assertEquals(1, $share2->getPermissions());
 
 		// update permissions
 		$params = [];
```
DeepDiver1975: easier to read
```php
$params = [
  'permissions' => '1',
  'extraPermissions' => [
    'app' => 'app1'
    ....
]
]

File: core/Migrations/Version20181220085457.php [hunk 16]
```diff
@@ -0,0 +1,30 @@
+<?php
+namespace OC\Migrations;
```
DeepDiver1975: file header is missing

core/js/shareitemmodel.js
```diff
@@ -752,7 +828,114 @@
...
+		 * type OC.Share.Types.ShareExtraPermission
+		 *
+		 * @param shareIndex
+		 * @returns OC.Share.Types.ShareExtraPermission[]
+		 */
+		getShareExtraPermissions: function(shareIndex) {
+			/** @type OC.Share.Types.ShareInfo **/
+			var share = this.get('shares')[shareIndex];
+			if(!_.isObject(share)) {
+				throw "Unknown Share";
```
DeepDiver1975: will the user see this? translate then

lib/private/Share20/ExtraPermissions.php
```diff
@@ -0,0 +1,71 @@
...
+namespace OC\Share20;
+
+use OCP\Share\IExtraPermissions;
+
+class ExtraPermissions implements IExtraPermissions {
+
+	/** @var array */
+	private $permissions;
+
+	public function __construct() {
```
DeepDiver1975: remove - not needed

lib/public/Share/IShare.php
```diff
@@ -184,6 +184,30 @@ public function setPermissions($permissions);
...
+	public function newExtraPermissions();
+
+	/**
+	 * Set extra permissions
+	 *
+	 * @param IExtraPermissions $permissions
+	 * @since 10.2.0
+	 * @return IShare The modified object
+	 */
+	public function setExtraPermissions($permissions);
```
DeepDiver1975: add type to argument please

lib/public/Share/IShare.php
```diff
@@ -184,6 +184,30 @@ public function setPermissions($permissions);
...
+	/**
+	 * Get extra permissions
+	 *
+	 * @return IExtraPermissions
+	 * @since 10.2.0
+	 */
+	public function getExtraPermissions();
```
DeepDiver1975: add return type please

DeepDiver1975 (changes requested):
Looks good - this can be merged after comments have been addressed

apps/dav/lib/DAV/ViewOnlyPlugin.php
mrow4a: Not, there are no - also Dav Browser Error plugin is broken and displays legacy page, see - 
https://github.com/owncloud/core/pull/33992

core/js/shareitemmodel.js
mrow4a: It is everywhere in the code like that, would require refactor to change it. Anyways it is rather a signal of the bug rather than expected exception

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Sabre plugin for the the file secure-view:
+ */
+class ViewOnlyPlugin extends ServerPlugin {
+
+	/** @var \Sabre\DAV\Server $server */
+	private $server;
+
+	/** @var \Sabre\DAV\Server $server */
```
DeepDiver1975: ilogger

apps/files/download.php
```diff
@@ -39,6 +39,17 @@
 	exit;
 }
 
+//Dispatch an event to see if any apps have problem with download
+$event = new \Symfony\Component\EventDispatcher\GenericEvent(null, ['path' => $filename]);
+OC::$server->getEventDispatcher()->dispatch('file.beforeGetDirect', $event);
+if ($event->hasArgument('errorMessage')) {
+	\header("HTTP/1.0 500 Forbidden");
```
DeepDiver1975: 500? forbiddeb?

apps/files_sharing/lib/ViewOnly.php
```diff
@@ -0,0 +1,110 @@
...
+	 */
+	public function check($pathsToCheck) {
+		// If any of elements cannot be downloaded, prevent whole download
+		$canDownload = true;
+		foreach ($pathsToCheck as $file) {
+			if (Filesystem::is_file($file)) {
+				// access to filecache is expensive in the loop
+				$fileInfo = Filesystem::getFileInfo($file);
+				if (!$this->checkFileInfo($fileInfo)) {
+					$canDownload = false;
```
DeepDiver1975: maybe return right away?

apps/files_sharing/lib/ViewOnly.php
```diff
@@ -0,0 +1,110 @@
...
+	 * @return bool
+	 */
+	private function dirRecursiveCheck($dir) {
+		$dirInfo = Filesystem::getFileInfo($dir);
+		if (!$this->checkFileInfo($dirInfo)) {
+			return false;
+		}
+		// If any of elements cannot be downloaded, prevent whole download
+		$canDownload = true;
+		$files = Filesystem::getDirectoryContent($dir);
```
DeepDiver1975: hard to mock - use non-static

apps/files_sharing/tests/FileListenersTest.php
```diff
@@ -0,0 +1,212 @@
...
+	/**
+	 * @var Hooks
+	 */
+	private $hooks;
+
+	protected function setUp() {
+		parent::setUp();
+		\OC_Hook::clear();
+
+		$this->eventDispatcher = \OC::$server->getEventDispatcher();
```
DeepDiver1975: mock this please

apps/files_sharing/tests/FileListenersTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+		$this->hooks = new Hooks(
+			\OC::$server->getRootFolder(),
```
DeepDiver1975: mock

DeepDiver1975 (changes requested):

apps/files_sharing/tests/FileListenersTest.php
mrow4a: @DeepDiver1975 this was the problem I tried to solve for a long time working on unit tests for this feature. It is non trivial to test that Hook class which uses that ViewOnly class (which uses Filesystem directly).

apps/files_sharing/tests/FileListenersTest.php
mrow4a: I had to use real classes not to make other unit tests fail (it uses global event dispatcher, and thus the registered listeners are active across many tests.. thus real classes need to be used not mocks)

apps/files_sharing/lib/ViewOnly.php
mrow4a: @DeepDiver1975 what non-static equivalent function do you suggest?

apps/files_sharing/tests/FileListenersTest.php
mrow4a: I can probably test ViewOnly separatelly and with Hooks class I just tests that these functions are being called - maybe this could simplify it.

apps/files_sharing/tests/FileListenersTest.php
DeepDiver1975: it should be possible to mock dispatcher as well as all other instances you are passing into the hooks class ... as long as no static file system methods are used (this should be avoided in any case)

lib/private/Share20/DefaultShareProvider.php
```diff
@@ -1238,4 +1262,47 @@ private function validate($share) {
...
+
+	/**
+	 * Load from database format (JSON string) to IAttributes
+	 *
+	 * @param IShare $share
+	 * @param string $data
+	 * @return IShare modified share
+	 */
+	private function updateShareAttributes(IShare $share, $data) {
+		if ($data !== null) {
```
jvillafanez: being a private function and stating that the "$data" parameter is a string (in the phpdoc), this check doesn't seem needed. I'd move the responsability to the caller.

lib/private/Share20/Manager.php
```diff
@@ -1543,4 +1551,23 @@ public function sharingDisabledForUser($userId) {
...
+	 * @param IAttributes|null $perms
+	 * @return string
+	 */
+	private function hashAttributes($perms) {
+		if ($perms === null) {
+			return "";
+		}
+
+		$formattedPermissions = [];
+		foreach ($perms->getScopes() as $scope) {
```
jvillafanez: I've seen this piece of code in several places... maybe it should be part of the `IAttributes`. A `toArray()` might be interesting. Note that the definition should be flexible

apps/files_sharing/lib/AppInfo/Application.php
```diff
@@ -154,12 +154,19 @@ function () use ($c) {
...
+			}
+
 			return new Hooks(
 				$c->getServer()->getLazyRootFolder(),
 				$c->getServer()->getUrlGenerator(),
 				$c->getServer()->getEventDispatcher(),
 				$c->getServer()->getShareManager(),
-				$c->query(NotificationPublisher::class)
+				$c->query(NotificationPublisher::class),
+				$uid
```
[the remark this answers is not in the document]
DeepDiver1975: Absolutly - furthermore it is a hard requirement in any case because of the initialization process and when the current user is know ...

lib/private/Share20/Manager.php
mrow4a: @jvillafancez not sure what is the intention here

apps/files_sharing/lib/ViewOnly.php
```diff
@@ -0,0 +1,116 @@
...
+						return false;
+					}
+				} elseif ($info instanceof Folder) {
+					// get directory content is rather cheap query
+					if (!$this->dirRecursiveCheck($info)) {
+						return false;
+					}
+				}
+			} catch (NotFoundException $e) {
+				continue;
```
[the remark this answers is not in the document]
mrow4a: It has unit tests, and is 100% covered for each of the cases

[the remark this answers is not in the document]
mrow4a: Check HooksTest

lib/private/Share20/Manager.php
mrow4a: you want IAttributes to have function toArray which returns map with attributes?

jvillafanez: I think the intention here is to create a json based on the contents of the attribute object. While returning a json might be too much, just returning a known array structure that can be easily feed into a json can be enough. This piece of code can be moved into the IAttribute, which additionally will prevent code duplication.
You'll still need to do the loops, but we move the responsability to create a known array structure to the IAttribute class.

apps/files_sharing/tests/HooksTest.php
```diff
@@ -171,4 +186,151 @@ public function testDiscardShareNotification() {
...
+	public function providesDataForCanGet() {
+		// normal file (sender) - can download directly
+		$senderFileStorage = $this->createMock(IStorage::class);
+		$senderFileStorage->method('instanceOfStorage')->with(SharedStorage::class)->willReturn(false);
+		$senderFile = $this->createMock(File::class);
+		$senderFile->method('getStorage')->willReturn($senderFileStorage);
+		$senderUserFolder = $this->createMock(Folder::class);
+		$senderUserFolder->method('get')->willReturn($senderFile);
+
+		$return[] = [ '/bar.txt', $senderUserFolder, true ];
```
jvillafanez: if possible, better use "result" as variable name to prevent any confusion

apps/files_sharing/lib/AppInfo/Application.php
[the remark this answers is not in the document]
mrow4a: makes sense, adjusted

apps/dav/tests/unit/DAV/ViewOnlyPluginTest.php
```diff
@@ -0,0 +1,132 @@
...
+		$storage->method('instanceOfStorage')->with(SharedStorage::class)->willReturn(true);
+		$storage->method('getShare')->willReturn($share);
+
+		$extAttr = $this->createMock(IAttributes::class);
+		$share->method('getAttributes')->willReturn($extAttr);
+		$extAttr->method('getAttribute')->with('core', 'secure-view-enabled')->willReturn($attrEnabled);
+
+		try {
+			// with these permissions / with this type of node user can download
+			$ret = $this->plugin->checkViewOnly($this->request);
```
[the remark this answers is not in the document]
mrow4a: I now assert for exception

mrow4a:
@jvillafanez @DeepDiver1975 adjusted to reviews

apps/files_sharing/lib/Hooks.php
```diff
@@ -131,6 +149,64 @@ function (GenericEvent $event) {
...
 		);
+
+		$this->eventDispatcher->addListener(
+			'file.beforeGetDirect',
+			function (GenericEvent $event) {
+				$pathsToCheck[] = $event->getArgument('path');
+
+				// Check only for user/group shares. Don't restrict e.g. share links
+				if ($uid = $this->getCurrentUserUid()) {
+					$viewOnlyHandler = new ViewOnly(
```
PVince81: weird that "ViewOnly" is not imported anywhere ?!

PVince81: never mind, same namespace

apps/files_sharing/lib/Hooks.php
```diff
@@ -131,6 +149,64 @@ function (GenericEvent $event) {
...
+				if (\is_array($files)) {
+					foreach ($files as $file) {
+						$pathsToCheck[] = $dir . '/' . $file;
+					}
+				} elseif (\is_string($files)) {
+					$pathsToCheck[] = $dir . '/' . $files;
+				}
+
+				// Check only for user/group shares. Don't restrict e.g. share links
+				if ($uid = $this->getCurrentUserUid()) {
```
PVince81: bad practice, please use strict comparison.

in PHP `if  "0"` returns false, so user ids like "0" would fail this test...

core/Migrations/Version20181220085457.php
```diff
@@ -0,0 +1,50 @@
...
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+
+		if ($schema->hasTable("${prefix}share")) {
+			$shareTable = $schema->getTable("${prefix}share");
+
+			if (!$shareTable->hasColumn('attributes')) {
+				$shareTable->addColumn(
+					'attributes',
+					Type::JSON,
```
PVince81: do we have integration tests covering this ? I hope this works for all our supported DBs

core/js/sharedialogshareelistview.js
```diff
@@ -279,7 +332,19 @@
...
-			this.model.updateShare(shareId, {permissions: permissions});
+			// Check extra share permissions
+			var attributes = [];
+			$('.attributes', $li).each(function(index, checkbox) {
+				var checked = $(checkbox).is(':checked');
+				$(checkbox).prop('enabled', checked);
+				attributes.push({
+					scope : $(checkbox).data('scope'),
+					key: $(checkbox).attr('name'),
+					enabled: checked
```
PVince81: use the "value" attribute to make it reversible ? we can then set "value='false'" to have the checkbox negate a permission when ticked

this would open the door for more understandable checkbox labels like "restricted mode" or something instead of "can view"

core/js/shareitemmodel.js
```diff
@@ -176,13 +229,76 @@
...
+						});
+					}
+				});
+
+				// Make sure allowed registered attributes which are
+				// not yet set are added. These can become available
+				// e.g. because of share permission/attr change
+				// and need to be initialized with their default value
+				if (!found) {
+					filteredAttributes.push({
```
PVince81: adding generic attributes on the server side was ok for storage purposes, but treating them in a generic way on the client just for 2 new attributes instead of hard-coding feels a bit over-engineered.

also this will cost way more time to add unit tests to make sure your generic logic works correctly while with a few hard-coded states for 2 attributes would produce smaller tests

let's leave this as is for now...

core/js/shareitemmodel.js
```diff
@@ -752,7 +880,98 @@
...
+			if(!_.isObject(share)) {
+				throw "Unknown Share";
+			}
+
+			if (_.isUndefined(share.attributes) || _.isUndefined(share.permissions)) {
+				return [];
+			}
+
+			// Add attributes for this share
+			var currentAttributes = JSON.parse(share.attributes);
```
PVince81: hmm, I'd have expected these to be pre-parsed when returned in the JSON response

anyway, let's leave this

lib/private/Share20/Manager.php
```diff
@@ -641,6 +642,7 @@ public function createShare(\OCP\Share\IShare $share) {
 			'shareType' => $share->getShareType(),
 			'uidOwner' => $share->getSharedBy(),
 			'permissions' => $share->getPermissions(),
+			'attributes' => $share->getattributes(),
```
PVince81: capital A

lib/public/Share/IAttributes.php
```diff
@@ -0,0 +1,68 @@
...
+ */
+namespace OCP\Share;
+
+/**
+ * Interface IAttributes
+ *
+ * @package OCP\Share
+ * @since 10.2.0
+ */
+interface IAttributes {
```
PVince81: IShareAttributes would have been better... we started a bad trend by having interface names relying on their namespaces... general confusion then when importing multiple of them with `use`

PVince81 (changes requested):
some minor changes like bad condition, typo

one nice to have: ability to switch a checkbox to set a permission to "false" instead of "true"

PVince81:
also:

- [ ] missing JS tests especially for the complex frontend attribute logic

PVince81:
- [ ] raise docs ticket and provide documentation for registering custom share attributes and their details

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,115 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\DAV;
+
+use OCA\DAV\Connector\Sabre\Exception\Forbidden;
+use OCA\DAV\Connector\Sabre\File as DavFile;
+use OCA\DAV\Meta\MetaFile;
+use OCA\Files_Sharing\SharedStorage;
```
PVince81: this will fail to import if the "files_sharing" app is disabled. better use the string instead for instance check like we do in other places.

apps/dav/lib/DAV/ViewOnlyPlugin.php
```diff
@@ -0,0 +1,115 @@
...
+				return true;
+			}
+			// Restrict view-only to nodes which are shared
+			$node = $davNode->getNode();
+			if (!$node instanceof FileInfo) {
+				return true;
+			}
+
+			$storage = $node->getStorage();
+			if (!$storage->instanceOfStorage(SharedStorage::class)) {
```
PVince81: use string here because class might not be loaded when "files_sharing" is disabled

core/js/sharedialogshareelistview.js
PVince81: or add an attribute "reverse" in the share attribute spec

core/js/sharedialogshareelistview.js
```diff
@@ -66,10 +66,18 @@
 			'</span>' +
 			'{{/if}}' +
 			'</div>' +
+			'<div class="shareAttributes"">' +
+			'{{#each shareAttributes}}' +
+			'<span class="shareOption">' +
+			'<input id="can-{{name}}-{{cid}}-{{shareWith}}" type="checkbox" name="{{name}}" class="attributes checkbox" {{#if enabled}}checked="checked"{{/if}} data-scope="{{scope}}" data-enabled="{{enabled}}""/>' +
```
PVince81: "enabled" is semantically the same as "checked".

in frontend terms we usually use "enabled" and "disabled" to gray out a component to make it read-only for example.

so when reading this code I was confused that "enabled" actually meant "checked"

PVince81:
what happens if two apps register the same attribute in the frontend ?

PVince81:
fixed minor points, rebased and adjusted for php unit 6

core/js/shareitemmodel.js
```diff
@@ -56,6 +57,25 @@
...
+	/**
+	 * @typedef {object} OC.Share.Types.RegisteredShareAttribute
+	 * @property {string} key
+	 * @property {bool}   default
+	 * @property {string} scope
+	 * @property {string} label
+	 * @property {string} description
+	 * @property {number[]} shareType
+	 * @property {number[]} incompatiblePermissions
+	 * @property {OC.Share.Types.ShareAttribute[]} incompatibleAttributes
```
PVince81: please add the actual doc text, what are these attributes about ?

core/Migrations/Version20181220085457.php
DeepDiver1975: unit tests are covering this - at least as we initially discussed using this data type

DeepDiver1975 (changes requested):
will submit fixes for the comments asap

PVince81:
I've squashed everything into two distinct commits:
1) first commit adds support for extra share permissions
2) second commit uses said support for the view-only "download" permission

PVince81:
I'm a bit concerned that the JSON field might not work in all databases with upsert from this upcoming PR https://github.com/owncloud/core/pull/34769.

we will see after merging and rebasing the other one...

PVince81 (approved):

PVince81:
stable10: https://github.com/owncloud/core/pull/34951

PVince81:
seems no issues with upsert and JSON: https://github.com/owncloud/core/pull/34952

DeepDiver1975:
THX @LinneyS - mind opening a pull request? We can much easier discuss this then. THX a lot!

mrow4a:
@LinneyS what is such permission about? We need to be sure that it follows the logic for "permissions in owncloud". That means, that checked permission box adds some higher level of permission, and unchecked lowers the permission level.

core/js/shareitemmodel.js
LinneyS: How and where is the `description` property used?

phil-davis:
`backport-request`  label removed. The backport was done already, see https://github.com/owncloud/core/pull/33994#issuecomment-479003389

PR #34023, opened 2019-01-04
[test] Bump phan 1.3.5

phil-davis:
Get a snapshot of what the latest ``phan`` version is reporting

### DIFF
--- a/apps/dav/lib/CalDAV/Calendar.php
+++ b/apps/dav/lib/CalDAV/Calendar.php
@@ -60,6 +60,7 @@ class Calendar extends \Sabre\CalDAV\Calendar implements IShareable {
 	public function updateShares(array $add, array $remove) {
 		/** @var CalDavBackend $calDavBackend */
 		$calDavBackend = $this->caldavBackend;
+		\assert($calDavBackend instanceof CalDavBackend);
 		$calDavBackend->updateShares($this, $add, $remove);
 	}
 
@@ -78,6 +79,7 @@ class Calendar extends \Sabre\CalDAV\Calendar implements IShareable {
 	public function getShares() {
 		/** @var CalDavBackend $calDavBackend */
 		$calDavBackend = $this->caldavBackend;
+		\assert($calDavBackend instanceof CalDavBackend);
 		return $calDavBackend->getShares($this->getResourceId());
 	}
 
@@ -139,6 +141,7 @@ class Calendar extends \Sabre\CalDAV\Calendar implements IShareable {
 
 		/** @var CalDavBackend $calDavBackend */
 		$calDavBackend = $this->caldavBackend;
+		\assert($calDavBackend instanceof CalDavBackend);
 		return $calDavBackend->applyShareAcl($this->getResourceId(), $acl);
 	}
 
@@ -166,6 +169,7 @@ class Calendar extends \Sabre\CalDAV\Calendar implements IShareable {
 
 			/** @var CalDavBackend $calDavBackend */
 			$calDavBackend = $this->caldavBackend;
+			\assert($calDavBackend instanceof CalDavBackend);
 			$calDavBackend->updateShares($this, [], [
 				$principal
 			]);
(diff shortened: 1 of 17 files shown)


apps/dav/lib/CalDAV/Calendar.php
```diff
@@ -60,6 +60,7 @@ public function __construct(BackendInterface $caldavBackend, $calendarInfo, IL10
 	public function updateShares(array $add, array $remove) {
 		/** @var CalDavBackend $calDavBackend */
 		$calDavBackend = $this->caldavBackend;
+		\assert($calDavBackend instanceof CalDavBackend);
```
phil-davis: ``phan`` does not understand (or believe or...) a variable doc like above. So it thinks that ``$calDavBackend`` can be just any old ``BackendInterface`` and it complains that ``BackendInterface`` has no method called ``updateShares``

The "easy" fix is to put this ``assert()`` here. Then ``phan`` really believes it is a ``CalDavBackend`` and so knows about method ``updateShares``.

The question I have is why is the constructor here got ``BackendInterface $caldavBackend``. Maybe it could/should be ``CalDavBackend $caldavBackend`` ?

apps/dav/lib/Connector/Sabre/AutorenamePlugin.php
```diff
@@ -95,7 +95,9 @@ public function handlePut(RequestInterface $request, ResponseInterface $response
 				return;
 			}
 
-			$view = $this->server->tree->getView();
+			$tree = $this->server->tree;
+			\assert($tree instanceof ObjectTree);
```
phil-davis: ``phan`` and PHPstorm tell me that ``$this->server->tree`` might not have a ``getView()`` method.
At run-time if it really does not have a ``getView()`` method, then I guess it is already going to explode.
Adding ``assert()`` here makes both ``phan`` and PHPstorm happy that ``getView()`` is going to exist.
As long as ``ObjectTree`` is really the only object type that happens here, then this ``assert()`` will only die when the line after would die anyway. But if there are other object types that have a ``getView()`` and happen here, then not so good.

lib/private/Streamer.php
```diff
@@ -94,6 +94,7 @@ public function addDirRecursive($dir, $internalDir='') {
 	 */
 	public function addFileFromStream($stream, $internalName, $size) {
 		if ($this->streamerInstance instanceof ZipStreamer) {
+			/* @phan-suppress-next-line PhanParamTooFew */
```
phil-davis: ``phan`` does not work out that this is a ``ZipStreamer`` - it thinks it is a ``TarStreamer`` and complains that there must be 3 parameters. (Maybe an ``assert()`` will help it?)

phil-davis:
This is a demo of what can be done to implement ``phan`` ``1.1.*`` and "fix" the things it reports.
It reports lots of ``PhanUndeclaredMethod`` usually because some object is returned to a variable somewhere, and the return value is some wider more general base object. Then the code goes on to call a method that only exists in some child object. See the comments I have made to get an idea of the situation.

https://github.com/phan/phan/wiki/Annotating-Your-Source-Code has hints about putting ``assert()`` in the code to force checks that the type of object really is correct. That makes ``phan`` and PHPstorm happy. It means that stuff could die at the ``assert()``. But if the correct ``assert()`` is used, then it will only die in the same situation as when it would die trying to call a method that does not exist.

Perhaps the return types of various methods could be tightened up to stop this problem? Perhaps there are other ways to "fix" this stuff? Or perhaps we ignore ``PhanUndeclaredMethod``?

Anyway, this is a little demo of what can be done.

Discussion @PVince81 @DeepDiver1975 @patrickjahns and anyone else.

apps/dav/lib/Connector/Sabre/AutorenamePlugin.php
patrickjahns: > At run-time if it really does not have a getView() method, then I guess it is already going to explode.

This implicit assumption is asking for trouble. If Either the code only receives proper objects, or it needs to explicitly protect itself from doing invalid things 

```
if (!$tree instanceof ObjectTree) { throw new \RuntimeException("...") }
```

phil-davis:
> Some of the changes even if they seem innocent, change object types and not all scenarios will be covered by unit tests here.

I agree. For the demonstration changes I made here, I was a bit surprised that I did not break any unit tests. Maybe what I did was fine and only assumed/asserted the correct object types actually used by the code, or maybe the code path(s) were not covered by unit tests, maybe the pigs can help before they fly ;)

I certainly did not feel comfortable asserting a particular object type in some of the situations where it was not obvious to me that it would be the only object type to ever be passed through that code path.

phil-davis:
See PR #35817

PR #34054, opened 2019-01-07
Add lock to the static tag in the input field when required

sharidas:
Add lock to the staic tag in the input field when
required. This is especially needed when the tag
does not belong to the group which is whitelisted.

## Description

This change tries to fix the problem:
- When a file is tagged with static tag which is not part of the group whitelisted ( in the static tag or say while creating static tag), the user is able to delete the tag using keyboard `Backspace`. Though technically its a fake delete. Because after refreshing the page, the static tag appears again for the file.

This is caused because of missing `select2-locked` class in the html element. This change tries to add the same. It tries to see if the tag is static tag or not and if the tag could be edited by the user. Based on these conditions the code would add `select2-locked` to the element.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Adding `select2-locked` to static tags when the user cannot edit the tags. This would prevent fake deletion of the tags by using `Backspace` key of keyboard.

## How Has This Been Tested?

- Create 2 groups, group1 and group2.
- Create 2 users user1 and user2. `user1` belongs to `group1`. `user2` belongs to `group2`.
- Add static tag `statictag` to the `group1`.
- As `user2` create file `test.txt` and share to `group1`.
- As `user1` assign static tag to `test.txt`.
- As `user2` try to check the fake delete using the keyboard. The result is captured in screen shot here:
statictag

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/systemtags/js/systemtagsinfoview.js
+++ b/apps/systemtags/js/systemtagsinfoview.js
@@ -12,10 +12,20 @@
 
 	function modelToSelection(model) {
 		var data = model.toJSON();
-		if (!OC.isUserAdmin() && !data.canAssign) {
-			data.locked = true;
+		if (!OC.isUserAdmin()) {
+			/**
+			 * If tag cannot be assigned just lock it.
+			 * Static tags need to be locked if they do not belong to the groups
+			 */
+			if (!data.canAssign || (isStaticTag(data) && !data.editableInGroup)) {
+				data.locked = true;
+			}
 		}
 		return data;
+	}
+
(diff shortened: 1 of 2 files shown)


PVince81 (approved):
:+1:

sharidas:
Backport to stable10 branch: https://github.com/owncloud/core/pull/34098

PR #34061, opened 2019-01-08
[stable10] Set permissions on log file creation instead of every write.

phil-davis:
Backport #32222


PVince81 (approved):
:+1: for 10.1.1

PR #34082, opened 2019-01-08
[stable10] Backport of Add regex for shared object file names

sharidas:
## Description

The change in this PR adds regular expression which would consider shared object file with extension not just `.so`. We should also consider files with extension say `foo.so.1.2` or `foo.1.2.so` or `foo.so` or `foo.so.`. The regular expression change acheives this.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

The extensions added to `resources/config/mimetypemapping.dist.json` would only help for files with single extension. But for the shared object files in GNU/Linux we may have `.so`, `.so.1.2`, `.1.2.so`. For this we need a regular expression in the Detection.php to fix the problem of multiple extension.

## How Has This Been Tested?

- Tested with unit test

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Files/Type/Detection.php
+++ b/lib/private/Files/Type/Detection.php
@@ -151,6 +151,26 @@ class Detection implements IMimeTypeDetector {
 	}
 
 	/**
+	 * Detect if the file is a shared object file. We detect this file
+	 * with a regex.
+	 *
+	 * @param string $fileName
+	 * @return bool true if file is a shared object else false
+	 */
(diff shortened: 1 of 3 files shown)


sharidas:
Original PR https://github.com/owncloud/core/pull/33894

PVince81 (approved):
:+1:

PR #34084, opened 2019-01-08
Run tests for stable10 on PHP 7.3

phil-davis:
Unit tests:
```
1) OCA\DAV\Tests\unit\ServerTest::test with data set "principals" ('principals/users/admin', array('caldav', 'oc-resource-sharing', 'carddav'))
ErrorException: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?

/drone/src/lib/composer/sabre/dav/lib/CalDAV/ICSExportPlugin.php:321
/drone/src/apps/dav/lib/Server.php:160
/drone/src/apps/dav/tests/unit/ServerTest.php:45
```
Is fixed by https://github.com/sabre-io/dav/pull/1080 which is in sabre/dav 4.0.0-alpha5

```
2) OCA\DAV\Tests\unit\CardDAV\CardDavBackendTest::testUpdateProperties
ErrorException: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?

/drone/src/lib/composer/sabre/vobject/lib/Component/VCard.php:514
/drone/src/apps/dav/tests/unit/CardDAV/CardDavBackendTest.php:381
```
(8 of these)
Are fixed by https://github.com/sabre-io/vobject/pull/424 which is in sabre/vobject 4.2.0-alpha1 which is a dependency for sabre/dav 4.0.0-alpha5 
So that would need backporting to some version combination that will support PHP 5.6

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -144,6 +144,15 @@ pipeline:
       matrix:
         TEST_SUITE: phan
 
+  php-phan-73:
+    image: owncloudci/php:7.3
+    pull: true
+    commands:
+      - make test-php-phan
+    when:
+      matrix:
+        TEST_SUITE: phan
+
   install-server:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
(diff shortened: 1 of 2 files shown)


phil-davis:
Good to know the above things that currently prevent PHP 7.3 from passing in ``stable10``
@DeepDiver1975 FYI, I guess at some point we will want to support PHP 7.3 with ``stable10`` ?
And so these ``sabre`` changes will need to be released as patches in versions of ``sabre`` libraries that also are happy to co-exist with PHP5.6

PR #34088, opened 2019-01-09
Phan 7.2 and 7.3

phil-davis:
## Description
1) rename ``php-cs-fixer`` to ``owncloud-coding-standard`` in drone to be like all the apps
2) run ``owncloud-coding-standard`` on PHP 7.3 (just to prove that it works)
3) run ``phan`` on each supported PHP 7.1 7.2 7.3 
4) Remove ``lint`` (covered by point 3)

## Motivation and Context
- partly related to #33902 
- ``phan`` now works in PHP 7.3
- ``phan`` and ``owncloud-codestyle`` complain if they find syntax errors, so they provide ``lint``

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34091, opened 2019-01-09
[stable10] Phan 7.3 and remove lint

phil-davis:
Backport #34088 

This is a "similar" backport. ``stable10`` supports PHP 5.6, so we:

- run ``owncloud-codestyle`` with PHP 5.6
- ``phan`` with PHP 7.*

Each of these tools will report syntax errors. So we get a syntax ``lint`` check for each PHP version "for free".

- remove ``lint``
- cleaned up old commented-out ``phpunit`` possible matrix combinations. These were getting old anyway and had become out-of-order and in odd places in the list. (they are not in master)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -99,23 +99,14 @@ pipeline:
       matrix:
         TEST_SUITE: javascript
 
-  phplint:
-    image: owncloudci/php:${PHP_VERSION}
-    pull: true
-    commands:
-      - make test-php-lint
-    when:
-      matrix:
-        TEST_SUITE: lint
-
-  php-cs-fixer:
+  owncloud-coding-standard:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
       - make test-php-style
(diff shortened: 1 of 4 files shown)


.drone.yml
```diff
@@ -144,6 +135,15 @@ pipeline:
       matrix:
         TEST_SUITE: phan
 
+  php-phan-73:
```
phil-davis: Note: I will make another PR in master to do ``phan`` more like this. That will be better in master, and make it match more with stable10.

individual-it:
`yaml: line 22: mapping values are not allowed in this context`

phil-davis:
I broke it - the phan step still has to mention ``PHP_VERSION:`` in the matrix, to keep other pipeline steps happy.

.drone.yml
phil-davis: See PR #34093

individual-it (approved):

PR #34100, opened 2019-01-09
Fix persistent lock timeout behavior

### LINKED ISSUE #33846: Persistent lock timeout value must be remaining time
See https://github.com/owncloud/core/issues/33778

When querying the lock discovery, the timeout value should be the remaining time, not absolute time.

@DeepDiver1975 FYI as discussed

PVince81:
## Description
Set timeout from header when specified.
Timeout value returned in lock discovery must be the remaining time.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/33846

## Motivation and Context
Timeout has bugs

## How Has This Been Tested?
Manual tests

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/Files/FileLocksBackend.php
+++ b/apps/dav/lib/Files/FileLocksBackend.php
@@ -28,6 +28,7 @@ use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\Locks;
 use Sabre\DAV\Locks\Backend\BackendInterface;
 use Sabre\DAV\Tree;
+use OCP\AppFramework\Utility\ITimeFactory;
 
 class FileLocksBackend implements BackendInterface {
 
@@ -35,10 +36,13 @@ class FileLocksBackend implements BackendInterface {
 	private $tree;
 	/** @var bool */
 	private $useV1;
+	/** @var ITimeFactory */
+	private $timeFactory;
 
-	public function __construct($tree, $useV1) {
+	public function __construct($tree, $useV1, $timeFactory) {
 		$this->tree = $tree;
 		$this->useV1 = $useV1;
+		$this->timeFactory = $timeFactory;
 	}
 
 	/**
@@ -116,7 +120,7 @@ class FileLocksBackend implements BackendInterface {
 			} else {
 				$lockInfo->scope = Locks\LockInfo::SHARED;
 			}
-			$lockInfo->timeout = $lock->getTimeout();
+			$lockInfo->timeout = $lock->getTimeout() - ($this->timeFactory->getTime() - $lock->getCreatedAt());
 
 			$davLocks[] = $lockInfo;
 		}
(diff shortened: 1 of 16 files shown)


PVince81:
- [ ] inclarity about infinity timeout

@DeepDiver1975 please clarify: currently the code always uses "1800" seconds if no timeout is given.

I've fixed it to properly read the timeout header which can be any integer or "Infinite".

It seems the decision was to ignore "Infinite" (-1) and go with 1800 ?
The current logic + SQL cannot deal with infinite timeout. RFC says the server has the last say.
It even seems that a request can contain multiple timeout values, in case one is not accepted... Seems Sabre ignores the second value so no fallback possible.

- [ ] LOCK response always contains the value from the "Timeout" header, even if we overrode it. Need to adjust "lockInfo" in case of override / non-accepted value

DeepDiver1975 (changes requested):

apps/dav/lib/Files/FileLocksBackend.php
```diff
@@ -92,7 +92,7 @@ public function getLocks($uri, $returnChildLocks) {
 			} else {
 				$lockInfo->scope = Locks\LockInfo::SHARED;
 			}
-			$lockInfo->timeout = $lock->getTimeout();
+			$lockInfo->timeout = $lock->getTimeout() - (\OC::$server->getTimeFactory()->getTime() - $lock->getCreatedAt());
```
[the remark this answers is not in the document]
PVince81: injection is usually ugly with Sabre plugins.

will try it in the next iteration

PVince81:
as discussed:

- change LockManager to return the ILock instance, then get its timeout value in FileLockBackend to be set to lockInfo
- proposal: have a max value of 1 day. if given timeout value is higher or infinte, set it back to 1 day

PVince81:
- added timefactory
- added default and max timeout of 1 day
- adjusted phpdocs
- now using ILock as return value for lockNodePersistent() to be able to read the updated timeout value

@DeepDiver1975 there is some little trouble now: since we return ILock after creation, the object itself doesn't have all attributes populated: the user and the abs dav path are not populated after insert. So the object you get is not 100% the same as the one you get with getLocks.

An alterative proposal would be to have lockNodePersistent return an updated lockInfo object instead...

lib/private/Lock/Persistent/LockManager.php
```diff
@@ -66,23 +70,20 @@ public function lock(int $storageId, string $internalPath, int $fileId, array $l
...
 		foreach ($locks as $lock) {
 			if ($lock->getToken() === $lockInfo['token']) {
-				$exists = true;
 				$lock->setCreatedAt($this->timeFactory->getTime());
 				$lock->setTimeout($timeout);
 				if ($lock->getOwner() === '' || $lock->getOwner() === null) {
 					$lock->setOwner($owner);
 				}
 				$this->lockMapper->update($lock);
+				return $lock;
```
PVince81: I hope this is ok: there can be only one lock with a given token, so no need to continue searching the list, we can return directly. Please confirm @DeepDiver1975

PVince81:
@DeepDiver1975 please review again, made changes as discussed.

See additional comment on the diff + the comment above about incomplete lock object

lib/private/Lock/Persistent/LockManager.php
DeepDiver1975: :+1:

DeepDiver1975 (approved):

PVince81:
rebased, and had to fix a failing test due to clash with the other locking PR from https://github.com/owncloud/core/pull/34171

PVince81:
web ui acceptance tests ran too long... I'll take the risk and merge this

phil-davis:
Backport ``stable10`` #34208

PR #34101, opened 2019-01-09
WebDav Client missing proxy credentials

micbar:
## Description
Adds the proxy user and passwords to the Curl Options.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/3078

## Motivation and Context
Federation behind a Proxy is not possible without this patch.

## How Has This Been Tested?
- will be tested manually

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -474,6 +474,11 @@ $CONFIG = array(
 /**
  * The optional authentication for the proxy to use to connect to the internet.
  * The format is: `username:password`.
+ *
+ * The username and the password need to be urlencoded to avoid breaking the
+ * delimiter syntax "username:password@hostname:port
+ *
+ * Example: "usern@me" needs to be encoded as "usern%40ame"
  */
 'proxyuserpwd' => '',
 
--- a/lib/private/Files/Storage/DAV.php
+++ b/lib/private/Files/Storage/DAV.php
@@ -342,7 +342,14 @@ class DAV extends Common {
 							->newClient()
 							->get($this->createBaseUri() . $this->encodePath($path), [
 									'auth' => [$this->user, $this->password],
-									'stream' => true
+									'stream' => true,
+									'config' => [
+										'stream_context' => [
+											'http' => [
+												'request_fulluri' => true
+											]
+										],
+									],
 							]);
 				} catch (RequestException $e) {
 					if ($e->getResponse() instanceof ResponseInterface
--- a/lib/private/Http/Client/Client.php
+++ b/lib/private/Http/Client/Client.php
@@ -132,9 +132,24 @@ class Client implements IClient {
 	 * @throws \Exception If the request could not get completed
 	 */
 	public function get($uri, array $options = []) {
-		$this->setDefaultOptions();
-		$response = $this->client->get($uri, $options);
 		$isStream = isset($options['stream']) && $options['stream'];
+		$this->setDefaultOptions();
+		if ($isStream) {
+			$proxyHost = $this->config->getSystemValue('proxy', null);
+			if ($proxyHost !== null) {
+				$this->client->setDefaultOption(
+					'proxy',
+					'tcp://' . $proxyHost
+				);
+			}
+
+			$proxyUserPwd = $this->config->getSystemValue('proxyuserpwd', null);
+			if ($proxyUserPwd !== null) {
+				$auth = \base64_encode(\urldecode($proxyUserPwd));
+				$this->client->setDefaultOption('headers/Proxy-Authorization', "Basic $auth");
+			}
+		}
+		$response = $this->client->get($uri, $options);
 		return new Response($response, $isStream);
 	}
 
--- a/lib/private/Http/Client/WebDavClientService.php
+++ b/lib/private/Http/Client/WebDavClientService.php
@@ -71,7 +71,7 @@ class WebDavClientService implements IWebDavClientService {
 	 */
 	public function newClient($settings) {
 		if (!isset($settings['proxy'])) {
-			$proxy = $this->config->getSystemValue('proxy', '');
+			$proxy = $this->getProxyUri();
 			if ($proxy !== '') {
 				$settings['proxy'] = $proxy;
 			}
@@ -92,5 +92,23 @@ class WebDavClientService implements IWebDavClientService {
 			$client->addCurlSetting(CURLOPT_CAINFO, $certPath);
 		}
 		return $client;
+	}
+
+	/**
+	 * Get the proxy URI
+	 *
+	 * @return string
+	 */
+	private function getProxyUri() {
+		$proxyHost = $this->config->getSystemValue('proxy', null);
+		$proxyUserPwd = $this->config->getSystemValue('proxyuserpwd', null);
+		$proxyUri = '';
+		if ($proxyUserPwd !== null) {
+			$proxyUri .= $proxyUserPwd . '@';
+		}
+		if ($proxyHost !== null) {
+			$proxyUri .= $proxyHost;
+		}
+		return $proxyUri;
 	}
 }
--- a/tests/lib/Http/Client/WebDavClientServiceTest.php
+++ b/tests/lib/Http/Client/WebDavClientServiceTest.php
@@ -62,12 +62,21 @@ class WebDavClientServiceTest extends \Test\TestCase {
 		$this->assertInstanceOf(Client::class, $client);
 	}
 
+	/**
+	 * Test Client with a proxy server
+	 *
+	 * @return void
+	 */
 	public function testNewClientWithProxy() {
 		$config = $this->createMock(IConfig::class);
-		$config->expects($this->once())
+		$config->expects($this->exactly(2))
 			->method('getSystemValue')
-			->with('proxy', '')
-			->willReturn('proxyhost');
+			->willReturnMap(
+				[
+					['proxy', null, 'proxyhost'],
+					['proxyuserpwd', null, null]
+				]
+			);
 
 		$certificateManager = $this->createMock(ICertificateManager::class);
 		$certificateManager->method('getAbsoluteBundlePath')
@@ -81,6 +90,42 @@ class WebDavClientServiceTest extends \Test\TestCase {
 		]);
 
 		$this->assertInstanceOf(Client::class, $client);
+
+		$curlSettings = $this->readAttribute($client, 'curlSettings');
+		$this->assertEquals('proxyhost', $curlSettings[CURLOPT_PROXY]);
+	}
+
+	/**
+	 * Test Client with a proxy server and Authorization
+	 *
+	 * @return void
+	 */
+	public function testNewClientWithProxyAndAuth() {
+		$config = $this->createMock(IConfig::class);
+		$config->expects($this->exactly(2))
+			->method('getSystemValue')
+			->willReturnMap(
+				[
+					['proxy', null, 'proxyhost'],
+					['proxyuserpwd', null, 'user:password']
+				]
+			);
+
+		$certificateManager = $this->createMock(ICertificateManager::class);
+		$certificateManager->method('getAbsoluteBundlePath')
+			->willReturn($this->tempManager->getTemporaryFolder());
+
+		$clientService = new WebDavClientService($config, $certificateManager);
+
+		$client = $clientService->newClient([
+			'baseUri' => 'https://davhost/davroot/',
+			'userName' => 'davUser'
+		]);
+
+		$this->assertInstanceOf(Client::class, $client);
+
+		$curlSettings = $this->readAttribute($client, 'curlSettings');
+		$this->assertEquals('user:password@proxyhost', $curlSettings[CURLOPT_PROXY]);
 	}
 
 	public function testNewClientWithoutCertificate() {
--- a/tests/lib/Files/Storage/DavTest.php
+++ b/tests/lib/Files/Storage/DavTest.php
@@ -526,7 +526,14 @@ class DavTest extends TestCase {
 			->with(
 				'https://davhost/davroot/some%25dir/file%25.txt', [
 					'auth' => ['davuser', 'davpassword'],
-					'stream' => true
+					'stream' => true,
+					'config' => [
+						'stream_context' => [
+							'http' => [
+								'request_fulluri' => true
+							]
+						],
+					],
 				]
 			)
 			->willReturn($response);
@@ -544,7 +551,14 @@ class DavTest extends TestCase {
 			->with(
 				'https://davhost/davroot/some%25dir/file%25.txt', [
 					'auth' => ['davuser', 'davpassword'],
-					'stream' => true
+					'stream' => true,
+					'config' => [
+						'stream_context' => [
+							'http' => [
+								'request_fulluri' => true
+							]
+						],
+					],
 				]
 			)
 			->willThrowException($this->createGuzzleClientException(Http::STATUS_NOT_FOUND));
@@ -561,7 +575,14 @@ class DavTest extends TestCase {
 			->with(
 				'https://davhost/davroot/some%25dir/file%25.txt', [
 					'auth' => ['davuser', 'davpassword'],
-					'stream' => true
+					'stream' => true,
+					'config' => [
+						'stream_context' => [
+							'http' => [
+								'request_fulluri' => true
+							]
+						],
+					],
 				]
 			)
 			->willThrowException($this->createGuzzleClientException(Http::STATUS_FORBIDDEN));
@@ -582,7 +603,14 @@ class DavTest extends TestCase {
 			->with(
 				'https://davhost/davroot/some%25dir/file%25.txt', [
 					'auth' => ['davuser', 'davpassword'],
-					'stream' => true
+					'stream' => true,
+					'config' => [
+						'stream_context' => [
+							'http' => [
+								'request_fulluri' => true
+							]
+						],
+					],
 				]
 			)
 			->willReturn($response);
--- a/tests/lib/Http/Client/ClientTest.php
+++ b/tests/lib/Http/Client/ClientTest.php
@@ -67,15 +67,13 @@ class ClientTest extends \Test\TestCase {
 
 	public function testGetProxyUriProxyHostWithPassword() {
 		$this->config
-			->expects($this->at(0))
 			->method('getSystemValue')
-			->with('proxy', null)
-			->willReturn('foo');
-		$this->config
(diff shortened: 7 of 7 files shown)


lib/private/Http/Client/WebDavClientService.php
```diff
@@ -93,4 +93,22 @@ public function newClient($settings) {
...
+	 * Get the proxy URI
+	 *
+	 * @return string
+	 */
+	private function getProxyUri() {
+		$proxyHost = $this->config->getSystemValue('proxy', null);
+		$proxyUserPwd = $this->config->getSystemValue('proxyuserpwd', null);
+		$proxyUri = '';
+		if ($proxyUserPwd !== null) {
+			$proxyUri .= $proxyUserPwd . '@';
```
PVince81: wondering if it works if the password contains a "@" :-D

tests/lib/Http/Client/WebDavClientServiceTest.php
```diff
@@ -64,10 +64,14 @@ public function testNewClient() {
...
 		$config = $this->createMock(IConfig::class);
-		$config->expects($this->once())
+		$config->expects($this->exactly(2))
 			->method('getSystemValue')
-			->with('proxy', '')
-			->willReturn('proxyhost');
+			->willReturnMap(
+				[
+					['proxy', null, 'proxyhost'],
+					['proxyuserpwd', null, 'user:password']
```
PVince81: also need a test case where proxuyuserpwd is not set ?

PVince81 (commented):
thanks for the PR, as discussed

some minor comments

tests/lib/Http/Client/WebDavClientServiceTest.php
micbar: Ok, that would be advisable.

lib/private/Http/Client/WebDavClientService.php
micbar: Little research tells me, that we should encode special characters. We are not doing it in the GuzzleHttp Client either. Maybe we should leave it this way and document, that we need an encoded string in config.php
`foo:B@r` should look like `foo:B%40r`

tests/lib/Http/Client/WebDavClientServiceTest.php
micbar: Done.

micbar:
Added an additional test.
I think we should document the need to encode special chars in the username:password string.

@PVince81

lib/private/Http/Client/WebDavClientService.php
mmattel: In case, please drop a note if there is a doc change necessary and what.

PVince81:
@micbar would it take more time to document it or to implement it correctly right away ?

micbar:
@PVince81 regarding the implementation  I would prefer documenting the need for escaping.

regarding the code, I have not been able to test it yet. Adjusted the labels.
maybe @patrickjahns has an idea how to get a test setup with two federated servers with FQDN an SSL Cert which can only resolve the other domain via a password protected proxy server?

PVince81:
@micbar ok, please document by editing config.sample.php

mmattel:
@micbar pls add me or @settermjd  to the config.sample.php PR (and backport) when improving the description because we have to do some backend task to transfer the content to docs when it gets merged - thanks.

patrickjahns:
> maybe @patrickjahns has an idea how to get a test setup with two federated servers with FQDN and SSL Cert which can only resolve the other domain via a password protected proxy server?

Could you clarify what you mean by resolving the domain? Resolving ( dns ) is independent of proxy servers. Http connections should pass through a proxy.

For manually testing, you might just use mitmproxy in regular mode https://docs.mitmproxy.org/stable/concepts-modes/#regular-proxy 

Steps:
- setup remote ownCloud "somewhere" ( example:  use local loopback device with 10.254.254.254 + docker container on port 8080. Add "federated.cloud" to /etc/hosts with 10.254.254.254  )
- setup the ownCloud with this patch locally 
- start mitmproxy
- configure local owncloud to use mitmproxy
validate that all requests that arrive at federated.cloud:8080 passed through mitmproxy

micbar:
@cdamken @PVince81 
We need someone to take over. I am not available for the next 4 weeks due to a full-time project.

I am not sure with the current approach. Maybe something else is missing.

We got the stack trace above, when entering `user:pass@proxy.domain` as the `proxy` value in config.php. This sends the same values to the CURL Request as with this PR and separate values.

PVince81:
as discussed internally, @micbar will take over again and finish this.

micbar:
@PVince81 
I invested some time again.
I think that the code of the PR works correctly, because it sets the proxy values correctly.

I could not overcome the last part of reading from a federated file.
I get the following error from the proxy:
Exeption:
```
Array
(
    [version] => 1.1
    [status] => 400
    [reason] => Bad Request
    [headers] => Array
        (
            [Server] => Array
                (
                    [0] => mitmproxy 4.0.4
                )

            [Connection] => Array
                (
                    [0] => close
                )

            [Content-Length] => Array
                (
                    [0] => 305
                )

            [Content-Type] => Array
                (
                    [0] => text/html
                )

        )

    [effective_url] => http://cloud.fed:10000/public.php/webdav/
    [body] => Resource id #106
)
```

Stack:
```
RequestFsm.php:131, GuzzleHttp\RequestFsm->GuzzleHttp\{closure}()
FulfilledPromise.php:25, React\Promise\FulfilledPromise->then()
CompletedFutureValue.php:55, GuzzleHttp\Ring\Future\CompletedFutureArray->then()
FutureResponse.php:43, GuzzleHttp\Message\FutureResponse::proxy()
RequestFsm.php:134, GuzzleHttp\RequestFsm->__invoke()
Client.php:165, GuzzleHttp\Client->send()
Client.php:125, GuzzleHttp\Client->get()
Client.php:136, OC\Http\Client\Client->get()
DAV.php:344, OCA\Files_Sharing\External\Storage->fopen()
Wrapper.php:295, OC\Files\Storage\Wrapper\Availability->fopen()
Availability.php:318, OC\Files\Storage\Wrapper\Availability->fopen()
Checksum.php:59, OC\Files\Storage\Wrapper\Checksum->fopen()
Wrapper.php:295, OCA\Files_Trashbin\Storage->fopen()
StorageWrapper.php:263, OCA\Workflow\Engine\Wrapper\StorageWrapper->fopen()
View.php:1184, OC\Files\View->basicOperation()
View.php:1029, OC\Files\View->fopen()
File.php:394, OCA\DAV\Connector\Sabre\File->get()
CorePlugin.php:90, Sabre\DAV\CorePlugin->httpGet()
WildcardEmitterTrait.php:96, OCA\DAV\Connector\Sabre\Server->emit()
Server.php:464, OCA\DAV\Connector\Sabre\Server->invokeMethod()
Server.php:241, OCA\DAV\Connector\Sabre\Server->start()
Server.php:309, OCA\DAV\Connector\Sabre\Server->exec()
Server.php:299, OCA\DAV\Server->exec()
remote.php:31, require_once()
remote.php:175, {main}()
```

micbar:
@PVince81 
Any Ideas? IMO the remaining issue is not an authentication issue, it has something to do with the kind of the DAV Get request for opening the file.

micbar:
@PVince81 As discussed, this can be merged. It fixes the proxy issue.

The remaining issue is the WebDAV GET Request to the external storage when opening the federated share file. The url format seems to be rejected by the proxy server (400 - bad request). It expects an absolute Url but gets a relative one

mmattel:
> Maybe we should leave it this way and document, that we need an encoded string in config.php
foo:B@r should look like foo:B%40r

There is no answer to my question raised (see above) if that should be documented in https://doc.owncloud.com/server/admin_manual/configuration/server/config_sample_php_parameters.html#the-optional-authentication-for-the-proxy-to-use-to-connect-to-the-internet
Currently there is no mentioning...

Pls advice

micbar:
@mmattel It should be mentioned.

mmattel:
@micbar what is the procedure for an admin to create an encoded proxy password. pls keep in mind that an andmin is not a programmer...

PVince81:
@micbar this is why I'd prefer we fix it right away ^

micbar:
> @micbar this is why I'd prefer we fix it right away ^

Sorry, i lost the track what should be fixed... :-)

PVince81:
here we go:

- [ ] escape the proxy settings in code instead of documenting, ref https://github.com/owncloud/core/pull/34101#issuecomment-453939075

micbar:
> here we go:
> 
>     * [ ]  escape the proxy settings in code instead of documenting, ref [#34101 (comment)](https://github.com/owncloud/core/pull/34101#issuecomment-453939075)

Cool, thx

patrickjahns:
Late to the party - but a serious question I have asked myself:
- do we really need to have a second parameter and add complex logic to escaping things in code?

If `proxyuserpwd` is never used - we can just use the normal `proxy` parameter and define the full url in there. ( As this is anyway how it is internally done).

Example:
```
   'proxy' => 'https://username:password@proxy.tld:9999' 
```

Or with special characters:
```
   'proxy' => 'https://username:p%40assword@proxy.tld:9999' 
```

This is a common way of defining proxyurls - so we can adhere to it and have less code and less confusion? ( also not needing a bug fix but rather a documentation change)

mmattel:
@patrickjahns as far my understanding goes, more or less independent of which key used, the question is how to get the escaped passwd for an admin who is not a programmer... pls correct if my pov is wrong.

micbar:
@mmattel We cannot do that.

Imagine `'proxyuserpwd' => 'user:n@me:password'`

We would need to use a regex to extract the two values before we could encode them without the delimiter. This could fail hard, if the admin enters something, we could not anticipate.

micbar:
I made an addition to the config.sample.php

config/config.sample.php
```diff
@@ -474,6 +474,11 @@
 /**
  * The optional authentication for the proxy to use to connect to the internet.
  * The format is: `username:password`.
+ *
+ * The username and the password need to be urlencoded to avoid breaking the
+ * delimiter syntax "username:password@hostname:port
+ *
+ * Example: "usern@me" needs to be encoded as "usern%40ame"
```
micbar: We could just copy that to the docs.

micbar:
Thanks to @patrickjahns we found the config value used in other code paths. So we can not deprecate the setting and need to continue with this PR as suggested and extend the docs here.

```./core/lib/private/Http/Client/Client.php:		$proxyUserPwd = $this->config->getSystemValue('proxyuserpwd', null);
./core/lib/private/SystemConfig.php:		'proxyuserpwd' => true,
```

patrickjahns:
@mmattel 
> admin who is not a programmer... pls correct if my pov is wrong.

I honestly expect that a ownCloud administrator knows what url_encdode / url_decode is - if he does not know this - I expect this person to be capable to use a common search engine with the description to find a resolution for how to url_encode/decode. In the worst case scenario, the documentation could include a short snippet how to url encode/decode the password. 

Having a proxy to reach the internet is not a common scenario on average hosting providers - you usually find this scenario in corporate networks as they want to scan incoming/outgoing traffic. I believe an administrator in an enterprise is able to find a person that can tell them about url encode/decode

micbar:
@patrickjahns @PVince81 
Current WIP seems to work. I also tested an urlencoded username.

## What I did
- Check if we do a get() on a stream resource
- Modify the stream context to use only `tcp://proxyhost`
- Move the Authorization in the Stream Context to the Header `Proxy-Authorization: Basic $auth` and `base64_encode(urldecode($auth))`

IMO the other get() requests without stream context can use `user:pw@proxyhost`

lib/private/Http/Client/Client.php
```diff
@@ -133,6 +133,19 @@ private function getProxyUri() {
 	 */
 	public function get($uri, array $options = []) {
 		$this->setDefaultOptions();
+
+		if ($options['stream']) {
```
patrickjahns: Should this only be set if proxy is `$this->config->getSystemValue('proxy', null) !== null` ? 
Or is it fine to set proxy to `tcp://null` and the underlying code works as expected?

micbar: good point. Let us add it to the if clause

config/config.sample.php
micbar: @mmattel @settermjd

micbar:
@cdamken 
Could we schedule a manual testing with @pako81 or @voroyam ?
I think this pr could have quite an impact on External WebDaV Storage.

lib/private/Files/Storage/DAV.php
```diff
@@ -342,7 +342,14 @@ public function fopen($path, $mode) {
 							->newClient()
 							->get($this->createBaseUri() . $this->encodePath($path), [
 									'auth' => [$this->user, $this->password],
-									'stream' => true
+									'stream' => true,
+									'config' => [
+										'stream_context' => [
+											'http' => [
+												'request_fulluri' => true
```
patrickjahns: does it make sense to always use this, or only in a proxy scenario?

micbar: That is a good question my dear :-)

micbar: What could go wrong? Why use relative paths to a webDav Resource anyway?

lib/private/Http/Client/Client.php
```diff
@@ -133,8 +133,26 @@ private function getProxyUri() {
 	 */
 	public function get($uri, array $options = []) {
 		$this->setDefaultOptions();
-		$response = $this->client->get($uri, $options);
+
 		$isStream = isset($options['stream']) && $options['stream'];
+		if ($isStream) {
+			$proxyHost = $this->config->getSystemValue('proxy', null);
+			if ($proxyHost !== null) {
```
patrickjahns: should we match the `empty()` call from the check below.
I guess it's a matter of preference - do we allow the user to do something like this in config.php
```
'proxy' => '',
```

.htaccess
```diff
@@ -92,3 +92,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
+
+ErrorDocument 403 /core/templates/403.php
+ErrorDocument 404 /core/templates/404.php
```
patrickjahns: please remove this change

micbar: oh, it slipped in accidentally.

lib/private/Http/Client/Client.php
micbar: 👍

patrickjahns:
@micbar 
Currently you only put this change to `get` to fix a specific scenario

What about moving that change into `setDefaultOptions()` - other request might also fail if the response is streamed

micbar:
> @micbar
> Currently you only put this change to `get` to fix a specific scenario
> 
> What about moving that change into `setDefaultOptions()` - other request might also fail if the response is streamed

I had this in mind too. I was looking around for another use of fopen() in the DAV External Storage implementation. Found none.

patrickjahns:
If we widen the spectrum to include all options, I wonder if we can easily setup acceptance scenarios with having a proxy inbetween. 
It should be possible to have a mitmproxy in a container. And then we run the acceptance scenarios for a federation/webdav etc with the proxy in-between.

Just not sure about the effort - might need to talk to @owncloud/qa

patrickjahns:
> I had this in mind too. I was looking around for another use of fopen() in the DAV External Storage implementation. Found none.

Need to double check - but this is not only DAV related - that's ownClouds default http client. So any request using the client would run into the issue

individual-it:
@patrickjahns created a new issue for that https://github.com/owncloud/core/issues/35263

micbar:
@individual-it @DeepDiver1975 
I tested it with a proxy and proxy auth
1) Establish fed share in both ways
2) Access the fed share
3) Connect an external Storage WebDAV
The external storage WebDAV refuses to create files or folders, but lists and reads files, except pdf

skshetry:
#35325 runs with this patch on passthrough proxy. Tests currently pass. I also ran through proxy without this patch on #35322, which fails. So, this demonstrates that the patch solves the issue. 

cc @micbar

micbar:
@DeepDiver1975 Could you review? I am not able to fully qualify the impact on external storages. See my comment about WebDAV External Storage.

lib/private/Http/Client/Client.php
```diff
@@ -132,9 +132,24 @@ private function getProxyUri() {
 	 * @throws \Exception If the request could not get completed
 	 */
 	public function get($uri, array $options = []) {
+		$isStream = isset($options['stream']) && $options['stream'];
 		$this->setDefaultOptions();
+		if ($isStream) {
+			$proxyHost = $this->config->getSystemValue('proxy', null);
+			if (!empty($proxyHost)) {
```
VicDeo: ` $proxyHost === null` ?

lib/private/Http/Client/Client.php
```diff
@@ -132,9 +132,24 @@ private function getProxyUri() {
...
+				$this->client->setDefaultOption(
+					'proxy',
+					'tcp://' . $proxyHost
+				);
+			}
+
+			$proxyUserPwd = $this->config->getSystemValue('proxyuserpwd', null);
+			if (!empty($proxyUserPwd)) {
```
VicDeo: the same as above

VicDeo (changes requested):

lib/private/Http/Client/Client.php
micbar: What if somebody enters `'proxy'=> ''` ?

lib/private/Http/Client/Client.php
micbar: same as above...

lib/private/Http/Client/Client.php
VicDeo: and what if somebody enters `'proxyUserPwd' => '0'`? :)

micbar: ok. :-)

lib/private/Http/Client/Client.php
micbar: ok

VicDeo (approved):
Good to go

VicDeo:
@micbar  backport?

phil-davis:
Backport `stable10` #35868

Issue #34002, opened 2019-01-02
Static tags are shown in the searchbox "Select tags to filter by" even when they do not belong to the user's group


davitol:
### Steps to reproduce

1. Create admin user followed by users user1, user2, user3 and user4.
2. Create 2 groups group1 and group2. user1 and user2 belong to group1. And user3 belong to group2. user4 belongs to group1 and group2.
3. Now create 4 tags
    - visibleTag of type Visible  
    - restrictedTag of type Restricted and add group1 to it  
    - staticTag of type Static and add group2 to it  
    - invisibleTag of type Invisible  

4. Login to user1

### Expected behaviour

The users will see restrictedTag and visibleTag.
The static tag should not be visible for users of group1

### Actual behaviour

The static tag is shown in the search box

<img width="1278" alt="screen shot 2019-01-02 at 10 31 48" src="https://user-images.githubusercontent.com/10596560/50586713-0fc75a00-0e7b-11e9-8178-1b2f5c881d4d.png">

**Note:** The behavior is fine in the sidebar for assigning tags to a file. The static tag is not shown there.

<img width="387" alt="screen shot 2019-01-02 at 10 38 47" src="https://user-images.githubusercontent.com/10596560/50586741-36859080-0e7b-11e9-8d60-422ad48c9c87.png">

### Server configuration

**Operating system**:
Ubuntu 16.04

**Web server:**
Apache

**Database:**
MySQL

**PHP version:**
7.1

**ownCloud version:**

ownCloud 10.1. beta version

```
{"installed":true,"maintenance":false,"needsDbUpgrade":false,"version":"10.1.0.1","versionstring":"10.1.0 beta","edition":"Enterprise","productname":"ownCloud"}
```

**Updated from an older ownCloud or fresh install:**
Fresh

**Are you using external storage, if yes which one:** 
No

**Are you using encryption:** 
No
**Logs**

```
```

### Client configuration

**browser**

Firefox

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributor most likely able to help you is @PVince81.

[...]

PR #34116, opened 2019-01-11
[stable10] Backpot of Fix visibility of static tags in the tags page

sharidas:
Fix the visibility of static tags in the tags
page. The static tags should not be visible
for all users.

## Description

Static tags should not be visible for all users. Excluding admin users, the users of the group which are whitelisted can see the tag. In this case, when user navigates to the tags page in the systemtags filter page, user is able to see the static tags. Expected behaviour is not all users should see it. This PR tries to address this problem.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/34002

## Motivation and Context

Static tags should not be visible for all users. Only users which are whitelisted in the groups for static tags can see the tag. This PR tries to fix the issue.

## How Has This Been Tested?

- Create users `user1`, `user2`, `user3` and `user4`
- Create groups `group1` and add `user1` and `user2` to it.
- Create groups `group2` and add `user3` to it.
- Add `user4` to both `group1` and `group2`.
- Create restricted tag for `group1`. Create static tag for `group2`. Create visible tag.
- Now login as `user1` or `user2`.
- User should be able to see restricted tag and visible tag ( same case with `user2` )

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/systemtags/js/systemtagsfilelist.js
+++ b/apps/systemtags/js/systemtagsfilelist.js
@@ -122,18 +122,50 @@
 		},
 
 		/**
+		 * Get the fetched tags from the filter query
+		 *
+		 * @param {Object} query select2 query object
+		 */
+		fetchedResult: function(query) {
+			var results = OC.SystemTags.collection.filterByName(query.term);
+			/**
+			 * Check if static tags are visible for this user.
+			 * If so show it, else don't.
+			 */
+			var indexToSplice = [];
+			for(var i=0; i < results.length; i++) {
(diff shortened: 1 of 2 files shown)


PVince81 (approved):
:+1:

PVince81:
this should have been in 10.1

PVince81:
also need to filter on server side: https://github.com/owncloud/core/issues/34440

PR #34125, opened 2019-01-12
Trigger CI

phil-davis:
Something is happening with the code style check


phil-davis:
This is OK - I thought there was a new problem in ``master`` with the code style check, but I was wrong, thankfully.

PR #34133, opened 2019-01-14
On patch level app updates the app version ins database needs to be updated

### LINKED ISSUE #34090: [10.1.0] App patch version update not updating oc_appconfig
### Steps

1. Install an app version 1.0.0 using occ market:install or occ app:enable
1. Update the app to version 1.0.1 by replacing the tarball (or increase the version in info.xml). Don't use occ.
1. Run `occ upgrade`
1. Check "oc_appconfig" for the app version

### Expected result
App version increased to 1.0.1 in oc_appconfig.

### Actual result
Message that everything is up to date and oc_appconfig still says 1.0.0

### Version
stable10 / ownCloud 10.1.0 beta

This happens because of the change with Semver where we said that patch versions of apps will not trigger migrations. See https://github.com/owncloud/core/pull/31103

The problem is that the oc_appconfig increase is usually happening during occ upgrade.
The other case where the increase might happen is when upgrading with occ market:update, however such command isn't always used. When dealing with offline instances, admins might need to manually update the code of apps by replacing the tarballs.

One idea would be to have `occ upgrade` still run but not enable maintenance mode, and just execute the routine for the apps. This is likely a risky change. Also it wouldn't affect web UI access, as web UI would usually show a page that requires an update, so this proposal is not a good solution.

Another idea would be to have new commands, for example "app:update" that would run app updates only. It could also cover the migrations part if needed and trigger maintenance mode.

So far I don't see a good quick fix.

@DeepDiver1975 @tomneedham

DeepDiver1975:
## Description
Patch level updates of an app do not trigger maintenance mode and any upgrade procedure.
But the app version as stored in the database needs to be kept in sync

## Related Issue
- Fixes #34090

## Types of changes
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/legacy/app.php
+++ b/lib/private/legacy/app.php
@@ -873,7 +873,15 @@ class OC_App {
 		$versions = self::getAppVersions();
 		$currentVersion = self::getAppVersion($app);
 		if ($currentVersion && isset($versions[$app])) {
-			return self::atLeastMinorVersionLevelChanged($currentVersion, $versions[$app]);
+			if ($currentVersion === $versions[$app]) {
+				return false;
+			}
+
+			if (self::atLeastMinorVersionLevelChanged($currentVersion, $versions[$app])) {
+				return true;
+			}
+			// update app version in db
+			\OC::$server->getConfig()->setAppValue($app, 'installed_version', $versions[$app]);
 		}
 		return false;
 	}
@@ -1003,7 +1011,7 @@ class OC_App {
 		self::setAppTypes($appId);
 
 		$version = self::getAppVersion($appId);
-		\OC::$server->getAppConfig()->setValue($appId, 'installed_version', $version);
+		\OC::$server->getConfig()->setAppValue($appId, 'installed_version', $version);
 
 		return true;
 	}


lib/private/legacy/app.php
```diff
@@ -873,7 +873,11 @@ public static function shouldUpgrade($app) {
 		$versions = self::getAppVersions();
 		$currentVersion = self::getAppVersion($app);
 		if ($currentVersion && isset($versions[$app])) {
-			return self::atLeastMinorVersionLevelChanged($currentVersion, $versions[$app]);
+			if (self::atLeastMinorVersionLevelChanged($currentVersion, $versions[$app])) {
+				return true;
+			}
+			// update app version in db
+			\OC::$server->getConfig()->setAppValue($app, 'installed_version', $versions[$app]);
```
PVince81: I suspect this will happen for every single http request and CLI run as we usually check "needs update" for core and apps to find out whether to block access

PVince81: which means additional SQL queries, one per app, even if the query does not modify anything (we have a "WHERE" statement)

DeepDiver1975: Right - I will add another check to not run this if the version stays unchanged ....

PVince81 (approved):
:+1:

DeepDiver1975:
Legacy code base -> test coverage failure is expected -> merge

PVince81:
seems we're not setting the correct value, see https://github.com/owncloud/core/issues/34851 ...

PR #34138, opened 2019-01-14
[stable10] Fixes #34090 - on minor update the app version ins databas…

### LINKED ISSUE #34090: [10.1.0] App patch version update not updating oc_appconfig

DeepDiver1975:
backport #34133

### DIFF
--- a/lib/private/legacy/app.php
+++ b/lib/private/legacy/app.php
@@ -906,7 +906,15 @@ class OC_App {
 		$versions = self::getAppVersions();
 		$currentVersion = self::getAppVersion($app);
 		if ($currentVersion && isset($versions[$app])) {
-			return self::atLeastMinorVersionLevelChanged($currentVersion, $versions[$app]);
+			if ($currentVersion === $versions[$app]) {
+				return false;
+			}
+
+			if (self::atLeastMinorVersionLevelChanged($currentVersion, $versions[$app])) {
+				return true;
+			}
+			// update app version in db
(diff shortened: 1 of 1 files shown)


PVince81 (approved):
:+1:

PR #34146, opened 2019-01-15
get correct values for occ command returns in acceptance tests

individual-it:
## Description
after recent refactoring we cannot access `lastStdOut` and `lastStdErr` directly but need to pull it from the feature context
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34147, opened 2019-01-15
[stable10] get correct values for occ command returns in acceptance tests

individual-it:
backport of  #34146


individual-it:
`timeout occurred after waiting 120 seconds for fsweb.test.owncloud.com:445`

phil-davis (approved):

Issue #34186, opened 2019-01-20
federationSharing acceptance test failures


phil-davis:
The following new scenarios seem to be failing too much in regular drone CI runs:
```
--- Failed scenarios:

    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:201
    /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:219

45 scenarios (43 passed, 2 failed)
466 steps (452 passed, 2 failed, 12 skipped)
31m25.99s (21.50Mb)
```

Need to investigate.

phil-davis:
Example https://drone.owncloud.com/owncloud/core/14358/619
from unrelated PR #34158 
```
  Scenario: test resharing folder and set it as readonly by owner                                                                 # /drone/src/tests/acceptance/features/webUISharingExternal/federationSharing.feature:219
    Given using server "LOCAL"                                                                                                    # FeatureContext::usingServer()
    And these users have been created:                                                                                            # FeatureContext::theseUsersHaveBeenCreated()
      | username |
      | user2    |
    When the user shares folder "simple-folder" with remote user "user1@%remote_server_without_scheme%" using the webUI           # WebUISharingContext::theUserSharesFileFolderWithUserUsingTheWebUI()
    And user "user1" re-logs in to "%remote_server%" using the webUI                                                              # WebUILoginContext::theUserRelogsInToUsingTheWebUI()
    And the user accepts the offered remote shares using the webUI                                                                # WebUISharingContext::theUserAcceptsTheOfferedRemoteShares()
    And user "user1" from server "REMOTE" shares "/simple-folder (2)" with user "user2" from server "LOCAL" using the sharing API # FederationContext::userFromServerSharesWithUserFromServerUsingTheSharingAPI()
    And user "user1" re-logs in to "%local_server%" using the webUI                                                               # WebUILoginContext::theUserRelogsInToUsingTheWebUI()
    And the user opens the share dialog for folder "simple-folder"                                                                # WebUISharingContext::theUserOpensTheShareDialogForFileFolder()
    And the user sets the sharing permissions of "user2@%local_server% (federated)" for "simple-folder" using the webUI to        # WebUISharingContext::theUserSetsTheSharingPermissionsOfForOnTheWebUI()
      | edit | no |
[...]

PR #34188, opened 2019-01-21
[test] Run webUISharingExternal suite 20 times to see how often it fails

phil-davis:
Issue #34186


phil-davis:
The tests pass over-and-over. Whatever was the reason for intermittent fails on some PRs last week, it is not happening now.

PR #34189, opened 2019-01-21
[test] Run webUISharingExternal suite 20 times to see how often it fails (short)

phil-davis:
and shorten the webUISharingExternal suite to see if the scenarios fail when run by themselves.

Issue #34186


phil-davis:
The tests pass 100 times! I cannot reproduce the fails that happened on some PRs last week.

PR #34190, opened 2019-01-21
Cut down drone and webUISharingExternal suite for testing in stable10

phil-davis:
Issue #34186 

See if federationSharing.feature fails in stable10


phil-davis:
federationSharing.feature passes lots of times in stable10

Issue #33840, opened 2018-12-10
REPORT request with <oc:filter-files> cannot be limited after deleting and recreating user


individual-it:
### Steps to reproduce
1. create user
2. login
3. mark files as fav.
4. request limited REPORT `curl -u user1:user1 -X REPORT http://localhost/owncloud-core/remote.php/dav/files/user1/ -d '<?xml version="1.0" encoding="utf-8" ?><oc:filter-files xmlns:a="DAV:" xmlns:oc="http://owncloud.org/ns" ><a:prop><oc:favorite/></a:prop><oc:filter-rules><oc:favorite>1</oc:favorite></oc:filter-rules><oc:search><oc:limit>2</oc:limit></oc:search></oc:filter-files>' | xmllint --format -`
   **All works well**
5. delete user
6. create same user
7. request limited REPORT `curl -u user1:user1 -X REPORT http://localhost/owncloud-core/remote.php/dav/files/user1/ -d '<?xml version="1.0" encoding="utf-8" ?><oc:filter-files xmlns:a="DAV:" xmlns:oc="http://owncloud.org/ns" ><a:prop><oc:favorite/></a:prop><oc:filter-rules><oc:favorite>1</oc:favorite></oc:filter-rules><oc:search><oc:limit>2</oc:limit></oc:search></oc:filter-files>' | xmllint --format -`
  **empty result**
### Expected behaviour
limited search reported

### Actual behaviour
empty report

### Server configuration
**Operating system**:
Ubuntu 18.04
**Web server:**
Apache2
**Database:**
MySQL
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
11.0
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git

individual-it:
reason seems to be that tables `oc_vcategory_to_object` and `oc_vcategory` are not cleaned on user deletion

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/1841 (Files deleted after recreating user), https://github.com/owncloud/core/issues/33835 (REPORT request with <oc:filter-files> cannot be limited in subfolder), https://github.com/owncloud/core/issues/27274 (delete), https://github.com/owncloud/core/issues/28084 (Feature Request: Warning before deleting user), and https://github.com/owncloud/core/issues/10453 (Can't delete user.).

DeepDiver1975:
[...]

PR #34195, opened 2019-01-21
Enable tests skipped for issue-33840

phil-davis:
## Description
1) Unskip acceptance test scenarios for issue #33840 
2) Adjust them so they pass now.
3) Leave commented-out test steps that suggest what should pass when the bug is fixed.
4) 

## Related Issue
https://github.com/owncloud/QA/issues/601

## Motivation and Context
For acceptance test scenarios that demonstrate bugs, instead of just skipping them, we want to comment-out the failing test step(s) and write test step(s) that currently pass (i.e. demonstrate the wrong behavior). Then when a developer fixes the bug, the test will start failing (because the wrong behavior no longer happens). This will force the developer to notice the acceptance test and adjust it.

Otherwise we end up with acceptance tests that stay skipped after the bug is fixed.

## How Has This Been Tested?
Local acceptance test run

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -2197,7 +2197,7 @@ trait WebDav {
 		$elementRows = $expectedFiles->getRowsHash();
 		$resultEntrys = $this->findEntryFromPropfindResponse($user);
 		foreach ($resultEntrys as $resultEntry) {
-			PHPUnit_Framework_Assert::arrayHasKey($resultEntry, $elementRows);
+			PHPUnit_Framework_Assert::assertArrayHasKey($resultEntry, $elementRows);
 		}
 	}
 
@@ -2208,7 +2208,8 @@ trait WebDav {
 	 * @param string $user
 	 * @param string $entryNameToSearch
 	 *
-	 * @return string if $entryNameToSearch is given and is found
+	 * @return string|array|boolean
+	 * string if $entryNameToSearch is given and is found
 	 * array if $entryNameToSearch is not given
 	 * boolean false if $entryNameToSearch is given and is not found
 	 */
@@ -2236,9 +2237,9 @@ trait WebDav {
 				}
 				\array_push($results, $entryName);
 			}
-			if ($entryNameToSearch === null) {
-				return $results;
-			}
+		}
+		if ($entryNameToSearch === null) {
+			return $results;
 		}
 		return false;
 	}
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2197,7 +2197,7 @@ public function theSearchResultOfShoudContainAnyOfTheseEntries(
 		$elementRows = $expectedFiles->getRowsHash();
 		$resultEntrys = $this->findEntryFromPropfindResponse($user);
 		foreach ($resultEntrys as $resultEntry) {
-			PHPUnit_Framework_Assert::arrayHasKey($resultEntry, $elementRows);
+			PHPUnit_Framework_Assert::assertArrayHasKey($resultEntry, $elementRows);
```
phil-davis: PHPstorm told me ``arrayHasKey`` was not correct, and that is true. So I fixed it.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2208,7 +2208,8 @@ public function theSearchResultOfShoudContainAnyOfTheseEntries(
 	 * @param string $user
 	 * @param string $entryNameToSearch
 	 *
-	 * @return string if $entryNameToSearch is given and is found
+	 * @return string|array|boolean
+	 * string if $entryNameToSearch is given and is found
 	 * array if $entryNameToSearch is not given
```
phil-davis: This method is supposed to return an empty array when ``entryNameToSearch`` is not given. But it was returning ``false``

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2236,9 +2237,9 @@ public function findEntryFromPropfindResponse($user, $entryNameToSearch = null)
 				}
 				\array_push($results, $entryName);
 			}
-			if ($entryNameToSearch === null) {
```
phil-davis: Move this block so that if ``entryNameToSearch`` is not given, any actually there is also nothing in the response XML, then an empty array will get returned - rather than ``false``
This avoids callers having to check if they got an array back, before looping through the array. Callers can rely on always getting an array if they did not specify ``entryNameToSearch``

individual-it (approved):

phil-davis:
Backport ``stable10`` #34197

PR #34205, opened 2019-01-21
Integrate HSM changes to encryption

sharidas:
Integrate HSM changes to encryption.
This has UI changes and config changes.

## Description

Integrating HSM changes, i.e, the UI changes to the encryption page. Basically added a check box to enable HSM. Once the user checks this the jwt secret and url needs to be provided. Both inputs are taken in a check box. And once done, the user can click the enable encryption button. If the HSM is disabled and encryption is enabled with master or user specific key, then after relogin the HSM option will not be available in the UI.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Integrating HSM changes with the encryption UI. 

## How Has This Been Tested?

- [ ] I will update the tests here.

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/settings/Panels/Admin/Encryption.php
+++ b/settings/Panels/Admin/Encryption.php
@@ -33,6 +33,9 @@ class Encryption implements ISettings {
 		$tmpl = new Template('settings', 'panels/admin/encryption');
 		$tmpl->assign('encryptionEnabled', \OC::$server->getEncryptionManager()->isEnabled());
 		$tmpl->assign('encryptionReady', \OC::$server->getEncryptionManager()->isReady());
+		$tmpl->assign('hsmEnabled', \OC::$server->getConfig()->getAppValue('encryption', 'hsmEnabled', '0'));
+		$tmpl->assign('hsmURL', \OC::$server->getConfig()->getAppValue('encryption', 'hsm.url', '0'));
+		$tmpl->assign('hsmJWTSecret', \OC::$server->getConfig()->getAppValue('encryption', 'hsm.jwt.secret', '0'));
 		$encryptionModules = \OC::$server->getEncryptionManager()->getEncryptionModules();
 		$defaultEncryptionModuleId = \OC::$server->getEncryptionManager()->getDefaultEncryptionModuleId();
 		$encModules = [];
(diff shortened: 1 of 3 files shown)


butonic:
in master the encryption app has been moved to https://github.com/owncloud/encryption. so the ui should be added there.

butonic (changes requested):
in master the encryption app has been moved to https://github.com/owncloud/encryption. so the ui should be added there.

PVince81:
@butonic there's a UI ??

I thought we wanted this to be a CLI only configuration as it's rather "expert mode" stuff

sharidas:
I know its getting scattered across core and encryption, but...  While testing somewhere I remember that before enabling encryption we have to provide the secret and url. That was another reason which made me add it over here. The reason for adding the text box in core is because the template was available here.

If the UI is not required, then I will test it and update here.
If the user prefers to use UI, then at least one time he/she may have to use console for command `occ config:app:set` to set the url and jwt secret val. So the procedure I followed:
- Enable encryption app in UI
- use `occ config:app:set` command to set the url and jwt
- Now try to navigate to the settings->encryption web page and do the regular way to enable encryption + select master key or user specific key

Handling HSM from the command line:
```
./occ a:e encryption
./occ config:app:set encryption hsm.jwt.#### --value '*****' 
./occ config:app:set encryption hsm.url --value '******'
./occ encryption:enable
./occ encryption:select-encryption-type masterkey
Warning: Only available for fresh installations with no existing encrypted data! There is also no way to disable it again. Do you want to continue? (y/n) y
Master key successfully enabled.
```
And then login as admin user. For instance modify any text file, the file would be encrypted.

If we do not need to expose the setting from UI, then we do have the command line ofcourse. Let me know.

butonic:
all configuration for encryption should be done on the cli. it should happen right after installation. no ui needed for this.

PVince81:
as discussed, I didn't realize this was UI-only. (was too lazy to check the diff, sorry).

closing

PR #34118, opened 2019-01-11
Skip preview expiry when owner cannot be determined

PVince81:
## Description
In some scenarios, the owner of trashed files or versions cannot be determined, instead of failing hard, just skip expiration.
This is just a mitigation as we need a proper solution for trash/version
preview expiry.

Please note that this is just a mitigation and actual issue isn't fixed yet (see below) as we need to bigger changes in the preview code to accomodate for versions, see https://github.com/owncloud/core/pull/33188

## Related Issue
Mitigation for https://github.com/owncloud/core/issues/32346 (version expiry) and https://github.com/owncloud/enterprise/issues/3067 (trash expiry).

## Motivation and Context
Mitigate the issue and avoid stopping the process or crashing whenever expiry did not work.

## How Has This Been Tested?
- [x] manual test

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34207, opened 2019-01-21
[stable10] Skip preview expiry when owner cannot be determined

DeepDiver1975:
In some scenarios, the owner of trashed files or versions cannot be determined, instead of failing hard, just skip expiration.
This is just a mitigation as we need a proper solution for trash/version
preview expiry.

backport of #34118


PVince81 (approved):
:+1:

PR #34208, opened 2019-01-21
[stable10] Fix persistent lock timeout behavior

PVince81:
Backport of https://github.com/owncloud/core/pull/34100 to stable10


PVince81:
weird discrepancy: test passes on master but fails here.

Also I noticed that this line https://github.com/owncloud/core/pull/34208/files#diff-c9768762e721f7989e83770969d9afc5R85 only exists on stable10 which might hint at an incomplete or missing backport, which might or might not explain the test result discrepancy.

Furthermore, I don't understand why the mocked LockWrapper in LockManagerTest returns `[]` on master but `null` on stable10, despite having the same test code...

PVince81:
further research points to 70f6cf6537356fb167cb527255e74a84dc812924 where this extra line was added for compatibility with php 5.6.

let's try adding the same line to the other test for explicit mocking

DeepDiver1975 (approved):

PR #34272, opened 2019-01-28
Unskip acceptance test for issue-23151

phil-davis:
## Description
1) Modify the test scenario so it deletes more files (5) of the same name as close together in time as possible. Then, for example, each API request takes as much as 0.7 seconds, we could have files deleted at time like:
```
12:34:50.9
12:34:51.6
12:34:52.3
12:34:53.0
12:34:53.7
```
and there will be some files that have a delete time in the same second.
(If we make and delete more files, we could extend that to 0.8 or 0.9 seconds, but this seems enough for now)

2) Comment out the results that are (mostly) not working now.

3) Leave notes to try and help the person who fixes the bug.

## Related Issue
https://github.com/owncloud/core/issues/23151
Working towards the "new" way that way provide acceptance test scenarios for bugs.
https://github.com/owncloud/QA/issues/601

## How Has This Been Tested?
CI and multiple local acceptance test runs

## Types of changes
- [ ] Technical debt
- [x] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature
+++ b/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature
@@ -105,25 +105,40 @@ Feature: files and folders exist in the trashbin after being deleted
       | old      |
       | new      |
 
-  @skip @issue-23151
+  @issue-23151
+  # This scenario deletes many files as close together in time as the test can run.
+  # On a very slow system, the file deletes might all happen in different seconds.
+  # But on "reasonable" systems, some of the files will be deleted in the same second,
+  # thus testing the required behavior.
   Scenario Outline: trashbin can store two files with the same name but different origins when the files are deleted close together in time
     Given using <dav-path> DAV path
     And user "user0" has been created with default attributes
     And user "user0" has created folder "/folderA"
     And user "user0" has created folder "/folderB"
+    And user "user0" has created folder "/folderC"
+    And user "user0" has created folder "/folderD"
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature
```diff
@@ -105,25 +105,40 @@ Feature: files and folders exist in the trashbin after being deleted
...
+    And user "user0" has copied file "/textfile0.txt" to "/folderC/textfile0.txt"
+    And user "user0" has copied file "/textfile0.txt" to "/folderD/textfile0.txt"
+    When user "user0" deletes these files without delays using the WebDAV API
+      | /folderA/textfile0.txt |
+      | /folderB/textfile0.txt |
+      | /folderC/textfile0.txt |
+      | /folderD/textfile0.txt |
+      | /textfile0.txt         |
+    # When issue-23151 is fixed, uncomment these lines. They should pass reliably.
+    #Then as "user0" the folder with original path "/folderA/textfile0.txt" should exist in trash
```
phil-davis: Note: we are a bit stuck here, because when this scenario runs, with the current server trashbin code, then probably some of the files do not exist. But we do not reliably know which ones will not exist (it depends on timing), and maybe on rare occasions all the files might exist.
So IMO we can't write a test scenario that will reliable pass now and reliably fail when the bug is fixed.

individual-it (approved):

phil-davis:
Backport ``stable10`` #34275

PR #34279, opened 2019-01-28
Update PR template to include DB changes

PVince81:
Since for [Semver](https://central.owncloud.org/t/owncloud-core-switching-to-semver-apps-need-re-release/17054) we will want to increase the minor version component instead of patch, I've added a checkbox in case there are database changes.

In a broader scope I wonder if we should remove those checkboxes and rely on labels instead, which would make it easier to sort or even script.


phil-davis (approved):
Seems a good thing to know

PR #34297, opened 2019-01-29
Use loadFromFile instead of loadFromFileHandle

### LINKED ISSUE #18645: OC_Image doesn't always set imageType correctly
OC_Image has imageType = IMAGETYPE_PNG by default.

Several load functions like loadFromBase64, loadFromData, loadFromFileHandle, load don't (always) set imageType to the correct type of image loaded.

For example, if you load a JPEG with loadFromData() imageType remains on IMAGETYPE_PNG.

Due to this, some other image processing functions don't correctly work or cause errors.

For example you get error "OC_Image->fixOrientation() Image is not a JPEG" if you call fixOrientation() on a JPEG loaded with loadFromData even though it's in fact a JPEG.

imageType should always match the type of image. All functions seems to set mimeType correctly but lack to set imageType...

sharidas:
Use loadFromFile instead of loadFromFileHandle
while generating the thumbnail.

## Description

When `loadFromFileHandle` is used for generating thumbnail of images, the problem found is that it doesn't fix the orientation of images with exif. While `loadFromFile` handles this. It uses `exif_imagetype` before checking the mimetype of the image file. There is a word of caution noted in the https://github.com/owncloud/core/blob/master/lib/private/legacy/image.php#L496 ( reference: http://php.net/manual/en/function.exif-imagetype.php#79283 ). With the changeset the orientation issue while generating preview is resolved.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/18645

## Motivation and Context

This changeset fixes the orientation issue of images while generating preview.

## How Has This Been Tested?

- Upload the image mentioned in the issue.
- The image should be able to view properly ( just like any image viewers like gwenview etc ).

## Types of changes

- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Preview/Image.php
+++ b/lib/private/Preview/Image.php
@@ -44,11 +44,12 @@ abstract class Image implements IProvider2 {
 		}
 
 		$image = new \OC_Image();
-		$stream = $file->fopen('r');
 
-		$image->loadFromFileHandle($stream);
+		$internalPath = $file->getInternalPath();
+		$localPath = $file->getStorage()->getLocalFile($internalPath);
+		$image->loadFromFile($localPath);
 		$image->fixOrientation();
-		\fclose($stream);
+
 		if ($image->valid()) {
 			$image->scaleDownToFit($maxX, $maxY);
 


PVince81:
seems to make sense, I was expecting something more complicated when I looked into this. It was probably because I took a different route based on git log / diff / bisect and tried to understand the commits that caused the regression.

if this fulfils all use cases as tested in the ticket, and doesn't cause additional performance impact (see local storage vs ext storage), then it should be fine

jvillafanez:
I guess to generate a preview we still need the whole file, even thought I'd still prefer to use streams...
Any idea how many changes are needed to use streams? If there are a lot of changes required, I'm fine with this solution.

PVince81:
@jvillafanez I think we never used streams before because many of the PHP functions for image processing operate directly on files, like changing orientation. We definitely need to look into improving this, it might require searching for new libraries.

PVince81 (approved):
:+1: looks good and thanks for the extensive tests in the original ticket

PVince81:
@sharidas please backport

sharidas:
Backport to stable10 https://github.com/owncloud/core/pull/34356

PR #34340, opened 2019-01-31
Fix features files for running test from user_ldap

dpakach:
## Description

Currently LDAP users cannot be created during test. Even if we try to create regular users from tests through user_ldap app, it assumes that we want to create ldap users and does not create any users.
Because of this we cannot test some basic features for LDAP users.
But while running tests through user_ldap app some ldap users are created by default and added to some default groups. So fixed the feature files to use those users and groups so that we can run these tests from user_ldap.

## Related Issue

- Related issue https://github.com/owncloud/user_ldap/issues/354

## Motivation and Context

Currently user_ldap requires more test coverate, but we cannot run these basic tests on ldap. By this fix we can add these core suites in user_ldap.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiCapabilities/capabilities.feature
+++ b/tests/acceptance/features/apiCapabilities/capabilities.feature
@@ -1,4 +1,4 @@
-@api
+@api @TestAlsoOnExternalUserBackend
 Feature: capabilities
 
   Background:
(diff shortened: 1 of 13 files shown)


skshetry (approved):

individual-it (approved):

dpakach:
Backport in #34353

PR #34388, opened 2019-02-05
Return only creatable part of the shareFolder path

VicDeo:
## Description
Check if  `shareFolder` is creatable and return it only if it exists or can be created

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3122

## Motivation and Context
Missing shares for guest users when shareFolder is non-default

## How Has This Been Tested?
1. Install and enable guests app.
2. add `	'share_folder' => '/omg/share',` to config.php
3. create a guest user by sharing any file via email
4. login as a guest user
#### Expected
guest user can see the shared file

#### Actual
guest user see no files

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/Helper.php
+++ b/apps/files_sharing/lib/Helper.php
@@ -281,13 +281,23 @@ class Helper {
 		$shareFolder = Filesystem::normalizePath($shareFolder);
 
 		if (!$view->file_exists($shareFolder)) {
-			$dir = '';
-			$subdirs = \explode('/', $shareFolder);
-			foreach ($subdirs as $subdir) {
-				$dir = $dir . '/' . $subdir;
-				if (!$view->is_dir($dir)) {
-					$view->mkdir($dir);
+			$currentDir = $shareFolder;
+			$dirsToCreate = [$shareFolder];
+			while (($currentDir = \dirname($currentDir)) !== '/') {
+				// FIXME: check if there is such file as well?
+				if ($view->is_dir($currentDir)) {
+					break;
 				}
+				$dirsToCreate[] = $currentDir;
+			}
+
+			$dirsToCreate = \array_reverse($dirsToCreate);
+			$shareFolder = '/';
+			foreach ($dirsToCreate as $dirToCreate) {
+				if ($view->mkdir($dirToCreate) === false) {
+					break;
+				}
+				$shareFolder = $dirToCreate;
 			}
 		}
 
--- a/apps/files_sharing/tests/HelperTest.php
+++ b/apps/files_sharing/tests/HelperTest.php
@@ -24,6 +24,8 @@
 
 namespace OCA\Files_Sharing\Tests;
 
+use OC\Files\View;
+
 /**
  * Class HelperTest
  *
@@ -42,6 +44,25 @@ class HelperTest extends TestCase {
 		$sharedFolder = \OCA\Files_Sharing\Helper::getShareFolder();
 		$this->assertSame('/Shared/Folder', \OCA\Files_Sharing\Helper::getShareFolder());
 		$this->assertTrue(\OC\Files\Filesystem::is_dir($sharedFolder));
+
+		// cleanup
+		\OC::$server->getConfig()->deleteSystemValue('share_folder');
+	}
+
+	/**
+	 * test get share folder on a read only storage
+	 */
+	public function testGetShareFolderOnReadOnlyStorage() {
+		\OCA\Files_Sharing\Helper::setShareFolder('/Share/Folder');
+
+		// Simulate when it is possible to create path /Share but not /Share/Folder
+		$viewMock = $this->createMock(View::class);
+		$viewMock->method('is_dir')
+			->willReturnOnConsecutiveCalls(false, false);
+		$viewMock->method('mkdir')
+			->willReturnOnConsecutiveCalls(true, false);
+		$sharedFolder = \OCA\Files_Sharing\Helper::getShareFolder($viewMock);
+		$this->assertSame('/Share', $sharedFolder);
 
 		// cleanup
 		\OC::$server->getConfig()->deleteSystemValue('share_folder');


apps/files_sharing/lib/Helper.php
```diff
@@ -282,12 +282,16 @@ public static function getShareFolder($view = null) {
 
 		if (!$view->file_exists($shareFolder)) {
 			$dir = '';
-			$subdirs = \explode('/', $shareFolder);
+			$subdirs = \array_filter(\explode('/', $shareFolder));
```
jvillafanez: Difficult... the reason I usually avoid this function is because you usually traverse the array completely twice: one for the filter, and another one for the specific action over the array.
The other problem is the implicit conversion string -> boolean, which might not be clear.

Possible options:
* Include a comment explaining what the `array_filter` is doing. Being explicit helps to understand what you're trying to do
* Add the callback parameter so the filter function is known.
* Try to handle the case inside the loop below.

I don't expect the share folder to be too deep, so we can keep the `array_filter` call for simplicity.

apps/files_sharing/lib/Helper.php
```diff
@@ -282,12 +282,16 @@ public static function getShareFolder($view = null) {
...
 			foreach ($subdirs as $subdir) {
-				$dir = $dir . '/' . $subdir;
-				if (!$view->is_dir($dir)) {
-					$view->mkdir($dir);
+				if (!$view->is_dir("$dir/$subdir")) {
+					$view->mkdir("$dir/$subdir");
 				}
+				if (!$view->is_dir("$dir/$subdir")) {
```
jvillafanez: I expect the `mkdir` to return true or false depending on the directory being created or not. This additional call shouldn't be needed.

apps/files_sharing/lib/Helper.php
```diff
@@ -282,12 +282,16 @@ public static function getShareFolder($view = null) {
...
+					$shareFolder = ($dir === '') ? '/' : $dir;
```
jvillafanez: This might need some explanation... why isn't this condition checked at the beginning of the function?

apps/files_sharing/tests/HelperTest.php
```diff
@@ -46,4 +48,20 @@ public function testSetGetShareFolder() {
 		// cleanup
 		\OC::$server->getConfig()->deleteSystemValue('share_folder');
 	}
+
+	/**
+	 * test get share folder on a read only storage
+	 */
+	public function testGetShareFolderOReadOnlyStorage() {
+		\OCA\Files_Sharing\Helper::setShareFolder('/Share/Folder');
```
jvillafanez: I don't know if a comment here explaining why this is needed would be useful... maybe as a reminder. New tests will likely copy & paste this without changing the actual code..

jvillafanez:
For the check of the directory, you're checking starting from the root directory. It's better if you start checking from the parent folder. It's possible that some of the directory structure is al ready created, so if we start checking from the parent directory the performance should be better because there should be less checks to be done.

apps/files_sharing/lib/Helper.php
VicDeo: because `$dir` is empty on the first cycle iteration?

apps/files_sharing/lib/Helper.php
```diff
@@ -282,12 +282,15 @@ public static function getShareFolder($view = null) {
 
 		if (!$view->file_exists($shareFolder)) {
 			$dir = '';
-			$subdirs = \explode('/', $shareFolder);
+			$subdirs = \explode('/', \trim($shareFolder, '/'));
 			foreach ($subdirs as $subdir) {
-				$dir = $dir . '/' . $subdir;
-				if (!$view->is_dir($dir)) {
-					$view->mkdir($dir);
+				if (!$view->is_dir("$dir/$subdir")) {
```
jvillafanez: Should we check if "$dir/$subdir" is a file. I mean, assuming we want the share folder to be "/a/b/c" maybe "/a/b" exists as a file. Not sure how ownCloud will behave in this case.

jvillafanez:
> To be more precise I'm fixing a bug in the code that is checking starting from the root directory.
And as the fix took over 8 hours so far, I don't want to spend more time on the benchmarking here.

Fair enough. Maybe open a ticket as tech debt to be checked later?

apps/files_sharing/lib/Helper.php
VicDeo: isn't it too much to fix the issue i need to fix? :)

apps/files_sharing/tests/HelperTest.php
VicDeo: added below

VicDeo:
@jvillafanez 
this code creates a mount point for **all** files shared with user as specified in config.php (by default all shares are  mounted directly to `uid/files/`)

the implementation is the following 
if the mount point already exists - just return path to it and exit
if it doesn't - iterate through the path and create missing directories.
This is done only **once per user**. 
 What performance improvements do you expect given that typical installation doesn't use this option at all and I barely can imagine installations where this path is more then 2 directories long?

VicDeo:
I can create such ticket but I guess  it will be yet another backlog item

apps/files_sharing/lib/Helper.php
jvillafanez: Taking into account that this is within the files_sharing app, and the function won't be used outside of the app, adding a comment in the phpdoc is acceptable. This way, whoever uses this piece is aware of this issue so is up to him to decide what to do.
In addition, the code was likely having the same problem (if it's really a problem)

jvillafanez:
Consider an external storage as primary storage (such as files_primary_s3). You'd be making several requests to an external service which might be slow to respond for whatever reason.

It's true it's expected to be executed once, and it's really unlikely that the performance hit will be noticeable, but since this can be improved I think it's better to open a ticket.

apps/files_sharing/lib/Helper.php
VicDeo: Fixme added

jvillafanez (approved):

VicDeo:
Stable10: https://github.com/owncloud/core/pull/34395

PR #33019, opened 2018-10-04
Support global CORS domains for public pages

DeepDiver1975:
## Description
CORS annotation on controller methods which are marked as public pages cannot use user specific cors domains. Only global cors domains can be used (set up via config.php)

As a result: no user specific domain can ever integrate via OAuth. 
Question: how to deal with this?

## Related Issue
- https://github.com/owncloud/phoenix/pull/258

## Motivation and Context
Integrate Phoenix with ownCloud via OAuth2 (later OpenID Connect)

## How Has This Been Tested?
- get new oauth app # linktocome
- get phoenix as per https://github.com/owncloud/phoenix/pull/258

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34391, opened 2019-02-05
The great CORS fix .....

DeepDiver1975:
## Description
- All known headers which are read from requests are being added to the Access-Control-Allowed-Headers
- All known header which are sent back to the client and which shall be processed by clients are added to the Access-Control-Expose-Headers
- Additional Allowed-headers can be read from the config file
- Bugfix for unauthenticated requests to OCS and DAV

## Related Issue
- https://github.com/owncloud/phoenix/issues/476

## How Has This Been Tested?
- unit tests
- integration tests as per https://github.com/owncloud/js-owncloud-client/pull/159

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/CorsPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/CorsPlugin.php
@@ -118,9 +118,14 @@ class CorsPlugin extends ServerPlugin {
 	 * @return void
 	 */
 	public function setCorsHeaders(RequestInterface $request, ResponseInterface $response) {
-		if ($request->getHeader('origin') !== null && $this->userSession->getUser() !== null) {
+		if ($request->getHeader('origin') !== null) {
 			$requesterDomain = $request->getHeader('origin');
-			$userId = $this->userSession->getUser()->getUID();
+			// unauthenticated request shall add cors headers as well
+			$userId = null;
+			if ($this->userSession->getUser() !== null) {
+				$userId = $this->userSession->getUser()->getUID();
+			}
+
 			$headers = \OC_Response::setCorsHeaders($userId, $requesterDomain, null, $this->getExtraHeaders($request));
 			foreach ($headers as $key => $value) {
 				$response->addHeader($key, \implode(',', $value));
--- a/apps/dav/lib/Connector/Sabre/CorsPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/CorsPlugin.php
@@ -60,8 +60,6 @@ class CorsPlugin extends ServerPlugin {
 
 	private function getExtraHeaders(RequestInterface $request) {
 		if ($this->extraHeaders === null) {
-			// TODO: design a way to have plugins provide these
-			$this->extraHeaders['Access-Control-Allow-Headers'] = ['X-OC-Mtime', 'OC-Checksum', 'OC-Total-Length', 'Depth', 'Destination', 'Overwrite'];
 			if ($this->userSession->getUser() === null) {
 				$this->extraHeaders['Access-Control-Allow-Methods'] = [
 					'OPTIONS',
--- a/apps/dav/tests/unit/Connector/Sabre/CorsPluginTest.php
+++ b/apps/dav/tests/unit/Connector/Sabre/CorsPluginTest.php
@@ -88,8 +88,8 @@ class CorsPluginTest extends TestCase {
 		$allowedDomains = '["https://requesterdomain.tld", "http://anotherdomain.tld"]';
 
 		$allowedHeaders = [
-			'X-OC-Mtime', 'OC-Checksum', 'OC-Total-Length', 'OCS-APIREQUEST', 'Accept',
-			'Authorization', 'Brief', 'Content-Length', 'Content-Range', 'Content-type',
+			'OC-Checksum', 'OC-Total-Length', 'OCS-APIREQUEST', 'X-OC-Mtime', 'Accept',
+			'Authorization', 'Brief', 'Content-Length', 'Content-Range',
 			'Content-Type', 'Date', 'Depth', 'Destination', 'Host', 'If', 'If-Match',
 			'If-Modified-Since', 'If-None-Match', 'If-Range', 'If-Unmodified-Since',
 			'Location', 'Lock-Token', 'Overwrite', 'Prefer', 'Range', 'Schedule-Reply',
@@ -327,7 +327,7 @@ class CorsPluginTest extends TestCase {
 		];
 	}
 
-	public function testAdditionalAllowedHeaders() {
+	public function testAuthenticatedAdditionalAllowedHeaders() : void {
 		$user = $this->createMock(IUser::class);
 		$user->method('getUID')->willReturn('someuser');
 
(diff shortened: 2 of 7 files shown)


DeepDiver1975:
Headers as used in core (excluding lib/composer and tests)

 - [ ] Accept
 - [ ] ACCEPT_LANGUAGE
 - [ ] Access-Control-Request-Method
 - [ ] Authorization
 - [ ] Content-Length
 - [ ] Content-Range
 - [ ] Content-Type
 - [ ] Etag
 - [ ] ETag
 - [ ] OC-Autorename
 - [ ] OC-CalDav-Import
 - [ ] OC-Chunked
 - [ ] OC-Etag
 - [ ] OC-FileId
 - [ ] OC-LazyOps
 - [ ] OC-Total-File-Length
 - [ ] OC-Total-Length
 - [ ] origin
 - [ ] Origin
 - [ ] Range
 - [ ] User-Agent
 - [ ] X-Expected-Entity-Length
 - [ ] X-Requested-With

PVince81:
@DeepDiver1975 ready for review or is there more coming ?

DeepDiver1975:
@PVince81 ready for review - THX

apps/dav/lib/Connector/Sabre/CorsPlugin.php
```diff
@@ -120,9 +118,13 @@ public function initialize(\Sabre\DAV\Server $server) {
 	 * @return void
 	 */
 	public function setCorsHeaders(RequestInterface $request, ResponseInterface $response) {
-		if ($request->getHeader('origin') !== null && $this->userSession->getUser() !== null) {
+		if ($request->getHeader('origin') !== null) {
 			$requesterDomain = $request->getHeader('origin');
-			$userId = $this->userSession->getUser()->getUID();
+			$userId = null;
+			if ($this->userSession->getUser() !== null) {
```
PVince81: beware, this means you changed behavior to now also provide CORS headers for non-authenticated users ?
maybe that's the "bugfix for authentication" you refer to ?

I don't remember exactly why we did it that way in the past. Maybe it was to avoid exposing too much to unauthenticated users. I guess this will become important for public link access and authentication schemes that need extra headers.

apps/dav/tests/unit/Connector/Sabre/CorsPluginTest.php
```diff
@@ -86,19 +88,15 @@ public function optionsCases() {
...
-			'Content-Type',
-			'Access-Control-Allow-Origin',
-			'X-Request-ID',
-			'X-OC-Mtime',
-			'OC-Checksum',
-			'OC-Total-Length',
-			'Depth',
-			'Destination',
-			'Overwrite',
+			'X-OC-Mtime', 'OC-Checksum', 'OC-Total-Length', 'OCS-APIREQUEST', 'Accept',
```
PVince81: how about putting these in a constant array in the regular code, and reading this list from there ?

apps/dav/tests/unit/Connector/Sabre/CorsPluginTest.php
```diff
@@ -327,4 +326,28 @@ public function providesOriginUrls() {
 			'plain http' => [false, 'http://example.net/'],
 		];
 	}
+
+	public function testAdditionalAllowedHeaders() {
+		$user = $this->createMock(IUser::class);
+		$user->method('getUID')->willReturn('someuser');
```
PVince81: please also test with null user as per change of behavior, see my above comment

lib/private/AppFramework/Middleware/Security/CORSMiddleware.php
```diff
@@ -98,10 +99,13 @@ public function afterController($controller, $methodName, Response $response) {
...
-		if ($this->request->getHeader('Origin') !== null &&
-			$this->reflector->hasAnnotation('CORS')) {
-			$requesterDomain = $this->request->getHeader('Origin');
+		$requesterDomain = $this->request->getHeader('Origin');
+		if ($requesterDomain === null) {
+			return $response;
+		}
 
+		if ($controller instanceof OCSController ||
+			$this->reflector->hasAnnotation('CORS')) {
```
PVince81: so OCSController doesn't require the CORS annotation and always let everything through ? not sure I follow.

if this is correct, add your answer as a PHP comment for future reference

lib/private/legacy/response.php
```diff
@@ -346,4 +348,78 @@ public static function setOptionsRequestHeaders($response, $headers = []) {
 
 		return $response;
 	}
+
+	private static function getExposeCorsHeaders() {
+		return [
+			'X-Sabre-Status',
```
PVince81: is "Content-Location" included somewhere / by default ? I think we sometimes use this with DAV for a soft redirect

now thinking of it, [private links also have "Webdav-Location"](https://github.com/owncloud/core/blob/v10.0.10/apps/files/lib/Controller/ViewController.php#L323) when using GET. for those we might anyway want to provide another DAV-based API ?

lib/private/legacy/response.php
```diff
@@ -346,4 +348,78 @@ public static function setOptionsRequestHeaders($response, $headers = []) {
...
+			// own headers
+			'X-OC-Mtime', 'OC-Checksum', 'OC-Total-Length', 'OCS-APIREQUEST',
+			// as used in sabre
+			'Accept',
+			'Authorization',
+			'Brief',
+			'Content-Length',
+			'Content-Range',
+			'Content-type',
+			'Content-Type',
```
PVince81: you have "content-type" twice with different casings

PVince81 (commented):
see comments

PVince81:
@DeepDiver1975 did you grep all apps as well for headers or just core + libs ? :see_no_evil:

apps/dav/lib/Connector/Sabre/CorsPlugin.php
DeepDiver1975: CORS for unauthenticated requests was implemented month ago - but not for all endpoints. This is the feature equalizer :man_shrugging:

PVince81: :+1:

lib/private/AppFramework/Middleware/Security/CORSMiddleware.php
DeepDiver1975: All old OCS endpoints allow cors by default. When moving to OCSController one explicitly has to add the CORS annotaion. This change enables CORS for all OCSController to not loose CORS when migrating from old ocs to OCSController

DeepDiver1975:
> @DeepDiver1975 did you grep all apps as well for headers or just core + libs ? see_no_evil

Only core and it's libs.

lib/private/AppFramework/Middleware/Security/CORSMiddleware.php
PVince81: right, now I remember it better :+1:

lib/private/legacy/response.php
DeepDiver1975: Ah - we use addHeader() .... re-greppping the code base ...

apps/dav/tests/unit/Connector/Sabre/CorsPluginTest.php
DeepDiver1975: Because we are the only testing that we can copy array data around. But we should test that an explicit list of headers is returned.

apps/dav/lib/Connector/Sabre/CorsPlugin.php
DeepDiver1975: refs https://github.com/owncloud/core/pull/33019

lib/private/legacy/response.php
DeepDiver1975: added a few missing headers - thx for the pointers!

DeepDiver1975:
@PVince81 all comments addressed as far as I can tell - please review once more - THX

PVince81 (approved):
code looks good now :+1:

DeepDiver1975:
WTF

```

There was 1 warning:

1) Warning
The data provider specified for OCA\DAV\Tests\unit\Connector\Sabre\CorsPluginTest::testAdditionalAllowedHeaders is invalid.
Data set #0 is invalid.

DeepDiver1975:
https://github.com/owncloud/core/pull/34476

PR #34466, opened 2019-02-12
Fix/quota override

tomneedham:
## Description
Currently the user management UI in stable10 uses this method, through `setquota.php` to update the quota of a user. However, if they are from another user backend, on sync the quota is just replaced. 

New version of https://github.com/owncloud/core/pull/32731

Setting the quota override should set it in the preferences table as well which is used as the manual override. On sync, the sync serivce checks this and uses this as an override instead. You can delete the key here to go back to the one supplied by the user backend.

## Related Issue
https://github.com/owncloud/core/pull/32731

## Motivation and Context
Quota is replaced on every sync if you override the quota in owncloud 10 and use an external user backend. 

## How Has This Been Tested?
Manually in the UI.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


tomneedham:
Wrong base. Big fingers.

PR #34467, opened 2019-02-12
[stable10] Store quota overrides in preferences table

tomneedham:
## Description
Currently the user management UI in stable10 uses this method, through `setquota.php` to update the quota of a user. However, if they are from another user backend, on sync the quota is just replaced. 

New version of https://github.com/owncloud/core/pull/32731

Setting the quota override should set it in the preferences table as well which is used as the manual override. On sync, the sync serivce checks this and uses this as an override instead. You can delete the key here to go back to the one supplied by the user backend.

## Related Issue
https://github.com/owncloud/core/pull/32731

## Motivation and Context
Quota is replaced on every sync if you override the quota in owncloud 10 and use an external user backend. 

## How Has This Been Tested?
Manually in the UI.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/User/User.php
+++ b/lib/private/User/User.php
@@ -461,6 +461,8 @@ class User implements IUser {
 			$quota = OC_Helper::humanFileSize($quota);
 		}
 		$this->account->setQuota($quota);
+		// Set the quota on the preferences table as an override
+		$this->config->setUserValue($this->getUID(), 'files', 'quota', $quota);
 		$this->mapper->update($this->account);
 		$this->triggerChange('quota', $quota);
 	}


PVince81:
@tomneedham hmm but wouldn't this code here also overwrite any value that was set by an admin as the override value during the next sync ?

tomneedham:
No. Sync sets on the account. https://github.com/owncloud/core/blob/master/lib/private/User/SyncService.php#L223

SyncSerice is the connector between the Account and the User Backend. 

This convenince method is only used by the provisioning api and the UI to override the user quota that is provided by the user backend.

PVince81:
Right, I see. So you brought back something we deleted back then.

PVince81:
I agree with the change, let's see what CI says. I hope we do have some coverage here

PVince81:
@individual-it @phil-davis as we have some work regarding LDAP automated tests, we should include something to cover this override

phil-davis:
@PVince81 @tomneedham looks like this was thought about recently: https://github.com/owncloud/user_ldap/issues/368

PVince81:
ok let's cover that with LDAP then, merging

PVince81 (approved):
:+1:

PVince81:
master: https://github.com/owncloud/core/pull/34511

micbar:
@tom see https://github.com/owncloud/enterprise/issues/2844#issuecomment-493048205

micbar:
@pmaier1 
Carlos re opened the ticket.

PR #29369, opened 2017-10-26
Add public-files to DAV

DeepDiver1975:
## Description
Add webdav resource for public shared files and folders

## Motivation and Context
DAV rules!

## How Has This Been Tested?
- [ ] share a file or folder publicly
- [ ] open the dav url /remote.php/dav/public-files/${token}

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #34472, opened 2019-02-12
[stable10] Enable PreviewPlugin on public webdav

PVince81:
## Description
Also provide Webdav-style URLs with extra args for requesting previews on the public webdav endpoint. We already have this on non-public webdav endpoint.

## Related Issue
None raised but follow up of https://github.com/owncloud/core/pull/31050 which brought Webdav previews.

## Motivation and Context
While working on the [media viewer app](https://github.com/owncloud/files_mediaviewer) came the question about how to retrieve the previews on the public endpont after realizing that the magic Webdav URLs did not work there.

## How Has This Been Tested?
:warning: doesn't work :warning: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] Port to master

### DIFF
--- a/apps/dav/appinfo/v1/publicwebdav.php
+++ b/apps/dav/appinfo/v1/publicwebdav.php
@@ -49,7 +49,8 @@ $serverFactory = new OCA\DAV\Connector\Sabre\ServerFactory(
 	\OC::$server->getMountManager(),
 	\OC::$server->getTagManager(),
 	\OC::$server->getRequest(),
-	\OC::$server->getTimeFactory()
+	\OC::$server->getTimeFactory(),
+	\OC::$server->getPreviewManager()
 );
 
 $requestUri = \OC::$server->getRequest()->getRequestUri();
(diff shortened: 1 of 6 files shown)


PVince81:
note: I did this quickly to research why we didn't do this in the first place.

I suspect that @DeepDiver1975 already attempted this and failed for the same reason.

For the PROPFIND and all Webdav operations we already included the header "X-Requested-With: XMLHttpRequest" which bypasses the auth here: https://github.com/owncloud/core/blob/stable10-publicpreview-webdav/apps/dav/lib/Connector/PublicAuth.php#L104

However for public previews, it is not an XHR call as it's done by image tags, so when calling public webdav it receives a 401 and the browser asks for credentials...

Maybe for this case we'd need another kind of bypass.

This PR is only a base for discussion. I'm fine abandoning this if not doable.
Note that we'll need some public API for Phoenix for public previews at some point...

Maybe we need to rework the public webdav endpoint anyway... see https://github.com/owncloud/core/pull/29369

PVince81:
closing in favor of https://github.com/owncloud/core/pull/29369 in the future

PR #32230, opened 2018-08-03
throw exception when .htaccess file is not writable

individual-it:
## Description
in case the `.htaccess` file cannot be updated throw an exception and show correct error message

## Related Issue
- Fixes #32227 

## Motivation and Context
I got stuck today during testing and it took me a while to figure out that the `.htaccess` file was not writeable, hope that fix will save time later

## How Has This Been Tested?
- try to install oC when `.htaccess` is not writeable 
- try to install oC when `.htaccess` is writeable 
- run `maintenance:update:htaccess` when `.htaccess` is not writeable 
- run `maintenance:update:htaccess` when `.htaccess` is writeable 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34478, opened 2019-02-14
do not fail update or installation when .htaccess file is not writable

### LINKED ISSUE #34471: Can't setup with non-writable htaccess
As far as I remember, some setups are configured fully read-only, or some admins don't want to let the web server write anything in the PHP tree, so they disable it.

We should allow non-writable htaccess.

This PR https://github.com/owncloud/core/pull/32230 makes it throw an exception.

We should soften it and output a warning on the console when it happens.

@individual-it

individual-it:
## Description
since https://github.com/owncloud/core/pull/32230 an exception is thrown when the .htaccess is read-only but it has to be possible to install or update oC when the .htaccess file is not writeable
but when `occ maintenance:update:htaccess` and the file cannot be updated the command should fail

## Related Issue
- Fixes  #34471

## Motivation and Context
see issue

## How Has This Been Tested?
- install oC with R/O `.htaccess`
- install oC with R/W `.htaccess`
- update oC with R/O `.htaccess`
- run `occ maintenance:update:htaccess` with R/O `.htaccess`
- run `occ maintenance:update:htaccess` with R/W `.htaccess`

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -286,11 +286,6 @@ class Setup {
 			$error[] = $l->t("Can't create or write into the data directory %s", [$dataDir]);
 		}
 		
-		// check if we can write .htaccess
-		if (!\is_file($this->pathToHtaccess()) || !\is_writable($this->pathToHtaccess())) {
-			$error[] = $l->t("Can't update %s", [$this->pathToHtaccess()]);
-		}
-
 		if (\count($error) != 0) {
 			return $error;
 		}
@@ -374,8 +369,13 @@ class Setup {
 			// out that this is indeed an ownCloud data directory
 			\file_put_contents($config->getSystemValue('datadirectory', \OC::$SERVERROOT.'/data').'/.ocdata', '');
 
-			// Update .htaccess files
-			Setup::updateHtaccess();
+			// check if we can write .htaccess
+			if (\is_file(self::pathToHtaccess())
+				&& \is_writable(self::pathToHtaccess())
+			) {
+				// Update .htaccess files
+				Setup::updateHtaccess();
+			}
 			Setup::protectDataDirectory();
 
 			//try to write logtimezone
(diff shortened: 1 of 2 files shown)


lib/private/Setup.php
```diff
@@ -472,11 +468,11 @@ public static function updateHtaccess() {
 
 		if ($content !== '') {
 			$fileWriteResult = @\file_put_contents(
-				$setupHelper->pathToHtaccess(), $htaccessContent . $content . "\n"
+				self::pathToHtaccess(), $htaccessContent . $content . "\n"
 			);
 			if ($fileWriteResult === false) {
```
phil-davis: Looks good - the "real" exception is still thrown here, if we did get here and still the htaccess file could not be written.

phil-davis (commented):
Looks good to me.
I will let @PVince81 give the final approval.

individual-it:
tried hard to add more unit tests but the whole install/update code is hard to test

lib/private/Setup.php
```diff
@@ -425,11 +425,7 @@ public static function updateHtaccess() {
 			$webRoot = !empty(\OC::$WEBROOT) ? \OC::$WEBROOT : '/';
 		}
 
-		$setupHelper = new \OC\Setup($config, \OC::$server->getIniWrapper(),
```
PVince81: I think the class exists in this form due to unit testing.
I'm unsure whether we should remove it like this but I don't mind if it doesn't break anything...

PVince81 (approved):
:+1:

setup works again on my env

PVince81:
@individual-it please backport

phil-davis:
Backport ``stable10`` #34486

PR #34481, opened 2019-02-14
Declare testCase methods void to match modern phpunit declaration

phil-davis:
## Description
Add the ``void`` return type to test ``setUp()`` ``tearDown()`` etc methods. This makes them compatible with PHPunit 8.0

See if this actually passes CI.

Note: this will not backport to ``stable10`` until PHP 5.6 support is dropped. Doing this in ``master`` also would mean that apps which inherit test classes from core would have to put the ``void`` return type in. And that would stop their unit tests from working against core ``stable10``. It would be very annoying to have to make separate ``stable10`` branches in lots of apps just for this! So merged this is "not a good thing" just yet.

Note: I was getting ``files_external_ftp`` unit tests to pass on Travis which has PHPunit 8.0, and used this core branch just to see if the unit tests would pass - they do.

## Motivation and Context
On Travis (for example) we started getting PHPunit 8.0 which puts return type declarations on some methods like ``function setUp(): void`` and so when we inherit from that, PHP complains if we do not also specify the return type ``void`` on the child class.

## How Has This Been Tested?
CI and files_external_ftp unit tests with Travis

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


phil-davis:
Note to self: this will be "a handy thing" for being ready for PHPUnit8.

phil-davis:
See draft PR #34961
Closing here (I cannot switch a PR back to being draft)

Issue #14049, opened 2015-02-10
Let users choose default folder for received shares


MTRichards:
As an ownCloud user, I would like to be able to select a shared folder root that is non-standard, so that all files and folders that are shared with me go where I want them in my account, rather than where the admin default has been set

Acceptance Criteria:
- Admin can still set the default for the entire server in config.php for the shared root
- User that gets a share simply gets the share placed in the default location for the system
- User may set a tab in the personal page labeled "Shared File Path" that allows the user to select one of their folders in the existing ownCloud filesystem as the root for all files shared with them. 
- Personal setting can be changed at any time and the user can opt to automatically migrate all "files shared with me" - this moves only the shared folders and files inside the old personal setting to the new personal setting, not folders that were dragged manually somewhere else.
- Personal setting can also be removed so that the system default is once again applied - this also prompts the user if they want to migrate the files already shared with them
- User can still decide to drag around a shared folder at any time and put it anywhere, this default is just to set the mount for when a file is shared with you

RealRancor:
Ref to a same/similar existing request: https://github.com/owncloud/core/issues/12298

PVince81:
@schiesbn @rullzer don't we have this already in ther personal page since OC 7 ?

rullzer:
@PVince81 no... there is a config.php setting for a global one. But per user we don't as far as I know.

kucukemre96:
I want solve this issue if that's okay for you.

PVince81:
@kucukemre96 feel free to take over. would be good to first post a proposal about the technical approach you intend to take before diving in as this could be more complex than it looks

kucukemre96:
Hello again. In files_sharing app there is a method in helper class that called getShareFolder which I think is used for setting a path for a certain file to be shared. Inside that function system values are checked and if there is one called share_folder (this one can be edited in config.php as a global setting of course) it is being setted as sharing folder or root folder by default. We can add a new tab in personal panel to make a user able to set a sharing folder for all his files and if we save that path as a uservalue, we can access that user value inside the getShareFolder function instead of systemvalue, all users can set their sharing folders individually I think.

micbar:
We are moving away from the paradigm that shares are mixed with the files in the personal space.

In ocis shares will always be kept in a specific folder and will be a separate sync folder.

PR #34562, opened 2019-02-19
panel-and-controllers-added

emrekuecuek:
Referring to issue #14049 . I want to solve this issue by creating a personal panel and changing getShareFolder function in Helper class. User value can be setted from personal panel, then after it is set; default folder for received shares can be controlled from a ceratin user value. (userSharedFilePath for example) But I got stuck on the javascript side, I want to get user value from an input; then set it on the backend.


DeepDiver1975:

PR #34593, opened 2019-02-22
allow user to configure automatically accept incoming shares

karakayasemi:
## Description
Let the users decide whether they want to manually accept shares or not. 

- Fixes `files_sharing` part of https://github.com/owncloud/enterprise/issues/3128 . There will be another PR for `federatedfilesharing`

## Motivation and Context
some users will not want that shares are accepted automatically.

## How Has This Been Tested?
Unit tests are added. Manually tested.
- When global `auto_accept_share` enabled, go Sharing Personal Panel and disable `auto accept share`, the next share should be on the pending status.
- When global `auto_accept_share` enabled, go Sharing Personal Panel and enable`auto accept share`, the next share should be accepted automatically.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/appinfo/routes.php
+++ b/apps/files_sharing/appinfo/routes.php
@@ -39,6 +39,11 @@ $application->registerRoutes($this, [
 			'url' => '/testremote',
 			'verb' => 'GET'
 		],
+		[
+			'name' => 'PersonalSettings#setUserConfig',
+			'url' => '/personalsettings/setuserconfig',
+			'verb' => 'POST'
+		],
 	],
 	'ocs' => [
 		[
--- a/apps/files_sharing/js/settings-personal.js
+++ b/apps/files_sharing/js/settings-personal.js
@@ -1,0 +1,10 @@
+$(document).ready(function() {
+	function saveSettings() {
+		var postData = {};
+		postData[$(this).attr('name')] = $(this).is(":checked") ? "yes" : "no";
+		$.post(OC.generateUrl('apps/files_sharing/personalsettings/setuserconfig'), postData);
+	}
+
+	var userConfig = $('#files_sharing_settings');
+	userConfig.find('input[type=checkbox]').change(saveSettings);
+});
--- a/apps/files_sharing/lib/Controller/PersonalSettingsController.php
+++ b/apps/files_sharing/lib/Controller/PersonalSettingsController.php
@@ -1,0 +1,104 @@
+<?php
+
+namespace OCA\Files_Sharing\Controller;
+
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\DataResponse;
+use OCP\IConfig;
+use OCP\IRequest;
+use OCP\IUserSession;
+
+/**
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+class PersonalSettingsController extends Controller {
+	const USERCONFIGS = [
+		'auto_accept_share'
+	];
+
+	/** @var IConfig $config */
+	private $config;
+
+	/** @var IUserSession $userSession */
+	private $userSession;
+
+	public function __construct(string $appName, IRequest $request, IConfig $config, IUserSession $userSession) {
+		parent::__construct($appName, $request);
+		$this->config = $config;
+		$this->userSession = $userSession;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @return DataResponse
+	 */
+	public function setUserConfig() {
+		$acceptedKeys = [];
+		$rejectedKeys = [];
+
+		$params = $this->getUserRequestParams();
+		foreach ($params as $key => $value) {
+			if ($this->validateParameter($key, $value)) {
+				$this->config->setUserValue(
+					$this->userSession->getUser()->getUID(),
+					'files_sharing',
+					$key,
+					$value
+				);
+				$acceptedKeys[] = $key;
+			} else {
+				$rejectedKeys[] = $key;
+			}
+		}
+		$status = Http::STATUS_OK;
+		if (empty($acceptedKeys)) {
+			$status = Http::STATUS_BAD_REQUEST;
+		}
+		return new DataResponse(
+			['accepted' => $acceptedKeys, 'rejected' => $rejectedKeys],
+			$status
+		);
+	}
+
+	/**
+	 * @param $key
+	 * @param $value
+	 * @return bool
+	 */
+	private function validateParameter($key, $value) {
+		return \in_array($key, self::USERCONFIGS) && ($value === 'yes' || $value === 'no');
+	}
+
+	/**
+	 * Exclude params starting with '_' from request parameters
+	 *
+	 * @return array
+	 */
+	private function getUserRequestParams() {
+		return \array_filter($this->request->getParams(),
+			function ($key) {
+				return (\substr($key, 0, 1) !== '_');
+			},
+			ARRAY_FILTER_USE_KEY
+		);
+	}
+}
--- a/apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
+++ b/apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Sharing\Panels\Personal;
+
+use OCP\AppFramework\Http\TemplateResponse;
+use OCP\IConfig;
+use OCP\IUserSession;
+use OCP\Settings\ISettings;
+use OCP\Template;
+
+class PersonalPanel implements ISettings {
+
+	/** @var IConfig $config */
+	private $config;
+
+	/** @var IUserSession $userSession */
+	private $userSession;
+
+	public function __construct(IConfig $config, IUserSession $userSession) {
+		$this->config = $config;
+		$this->userSession = $userSession;
+	}
+
+	/**
+	 * The panel controller method that returns a template to the UI
+	 *
+	 * @return TemplateResponse | Template
+	 */
+	public function getPanel() {
+		$autoAcceptShareEnabled = $this->config->getUserValue(
+			$this->userSession->getUser()->getUID(),
+			'files_sharing',
+			'auto_accept_share',
+			'yes'
+		);
+		$tmpl = new Template('files_sharing', 'settings-personal');
+		$tmpl->assign(
+			'userAutoAcceptShareEnabled',
+			$autoAcceptShareEnabled
+		);
+		return $tmpl;
+	}
+
+	/**
+	 * A string to identify the section in the UI / HTML and URL
+	 *s
+	 * @return string
+	 */
+	public function getSectionID() {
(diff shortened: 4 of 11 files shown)


apps/files_sharing/appinfo/routes.php
```diff
@@ -39,6 +39,11 @@
 			'url' => '/testremote',
 			'verb' => 'GET'
 		],
+		[
+			'name' => 'PersonalSettings#setUserAutoAcceptShare',
+			'url' => '/personalsettings/setautoaccept',
```
jvillafanez: I'm pretty sure this won't be the last sharing thing we'll want the user to choose, and I'm also pretty sure that most of the future configuration options for the user will end up in the same place.
Try something more generic that can be reused in the future.

apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
```diff
@@ -0,0 +1,82 @@
...
+	 * The panel controller method that returns a template to the UI
+	 *
+	 * @since 10.0
+	 * @return TemplateResponse | Template
+	 */
+	public function getPanel() {
+		$tmpl = new Template('files_sharing', 'settings-personal');
+		$tmpl->assign(
+			'userAutoAcceptShareEnabled',
+			$this->config->getUserValue(
```
jvillafanez: Better move this at the beginning of the function. The `assign` function feels quite bloated because you usually write it like `$tmpl->assign('myvar', $myvar)`

apps/files_sharing/js/settings-personal.js
```diff
@@ -0,0 +1,11 @@
+$(document).ready(function() {
+	$('#userAutoAcceptShareInput').change(function() {
+		var value = 'no';
+		if (this.checked) {
+			value = 'yes';
+		}
+		$.post(
+			OC.generateUrl('apps/files_sharing//personalsettings/setautoaccept'), {state: value}
```
PVince81: double slash ?

apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
```diff
@@ -0,0 +1,82 @@
...
+
+	public function __construct(IConfig $config, IUserSession $userSession) {
+		$this->config = $config;
+		$this->userSession = $userSession;
+	}
+
+	/**
+	 * The panel controller method that returns a template to the UI
+	 *
+	 * @since 10.0
```
PVince81: `@since` not needed as this is no public API, same below

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,58 @@
...
+	private $userSession;
+
+	public function __construct(string $appName, IRequest $request, IConfig $config, IUserSession $userSession) {
+		parent::__construct($appName, $request);
+		$this->config = $config;
+		$this->userSession = $userSession;
+	}
+
+	public function setUserAutoAcceptShare($state) {
+		$this->config->setUserValue(
```
PVince81: please validate the state and don't accept arbitrary values.

a unit test should confirm that validation works

PVince81 (changes requested):
looks fine, see comments for some small additions

apps/files_sharing/appinfo/routes.php
karakayasemi: @jvillafanez Are you suggesting to change the path as something like `setconfig` and use this endpoint for all the future configuration options? In this way, every input change leads to set all configuration options. I thought the future options can implement their own endpoint under `/personalsettings/`  URL path.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
karakayasemi: done

apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
karakayasemi: done

apps/files_sharing/appinfo/routes.php
jvillafanez: > Are you suggesting to change the path as something like `setconfig` and use this endpoint for all the future configuration options

That's the idea. 90% of the options will follow the same code path, which means that each time we need to add a new option we don't need to test everything (there shouldn't be code changes there, and it should also be tested already), and the code changes required to add a new option are minimal.
As long as the input and output of the endpoint are clear, I don't see many problems: you know what you can expect. For special cases, we'll need to add additional code anyway, so for me, it's a matter of how many special cases do we expect to handle.

karakayasemi: @jvillafanez I added allowed user config keys constant array in settings to validate coming params and converted `setUserConfig` method to a more generic one. Please review, one more time. Thanks.

apps/files_sharing/js/settings-personal.js
```diff
@@ -0,0 +1,11 @@
+$(document).ready(function() {
+	function saveSettings() {
+		var post = $('#files_sharing_settings').find('input:checkbox').map(function() {
```
jvillafanez: Include a comment here to explain what you want to do. It's a bit difficult to follow.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,72 @@
...
+			) {
+				$this->config->setUserValue(
+					$this->userSession->getUser()->getUID(),
+					'files_sharing',
+					$key,
+					$value
+				);
+			}
+		}
+		return new DataResponse(
```
jvillafanez: We should adjust this reponse. The problem I see is that I can send several key-value pairs but only a couple of them will be saved. The response seems to tell that all the key-value pairs were saved correctly.

I think we can send a 4xx response if all the keys were rejected, and a 2xx response if any of them was saved, but in this case we can send the key-value pairs that were rejected and saved as part of the body, so the client can check if there was any key that wasn't saved

apps/files_sharing/js/settings-personal.js
```diff
@@ -0,0 +1,11 @@
...
+			return { name: this.name, value: this.checked ? "yes" : "no" };
+		});
+		$.post(OC.generateUrl('apps/files_sharing/personalsettings/setuserconfig'), post);
+	}
+
+	var userConfig = $('#files_sharing_settings');
+	userConfig.find('input[type=checkbox]').change(saveSettings);
```
jvillafanez: Note that the handler will be attached to all the checkboxs in the section. This means that when you change a checkbox you'll be sending the information of all the checkboxes found.
I'd just send only the information of the changed checkbox. I don't think we want a button to save all the settings.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
jvillafanez: If the function will save only one setting at a time, return the proper error response if it couldn't be saved.

apps/files_sharing/js/settings-personal.js
karakayasemi: now it only sends changed checkbox value.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
karakayasemi: Now it will only take a config key-value pair and if cant set the value, it will return a response with Http::STATUS_BAD_REQUEST.

apps/files_sharing/js/settings-personal.js
karakayasemi: It will listen to all the checkbox changes but only send related checkbox change.

apps/files_sharing/js/settings-personal.js
karakayasemi: removed.

karakayasemi:
@jvillafanez I believe this time the code is sufficient for all the discussed requirements. Please, review one more time. Thanks.

jvillafanez:
It would be better if the calls are something like `/personalsettings/setuserconfig?mykey=myvalue&myotherkey=myothervalue` (don't mind the "GET-like" request).

In the controller you should be able to do something like (make sure the request is accesible):
```
$requestParams = $this->request->getParams();
foreach ($requesParams as $key => $value) {
  if (isKeyValid($key)) {
    save($key, $value);
  }
}
```
The client can send the data like (double-check the syntax), classic key-value pairs:
```
var item = $(this);
$.post(url, {
  item.attr('name'): item.is(":checked") ? "yes" : "no"
});
```

From a client perspective, I think this makes more sense.

Note that, although the server can accept several key-value pairs at the same time, the client will only send them one by one (at least for now). This is fine.

karakayasemi:
Sorry, but I don't quite understand the benefit of using "GET" instead of "POST". I can already use `$request->getParams()` with the "POST" request. But, there are unrelated entries in params like `_route`. To get rid of this and to make sure there was only a config parameter, I preferred to write in this way.

jvillafanez:
> Sorry, but I don't quite understand the benefit of using "GET" instead of "POST"

Forget this and keep using POST. It was just an example.

> there are unrelated entries in params like `_route`

You're still validating that the keys we want to set are known, so we can just include a comment to be warned about additional keys not fully related to the request. You can also skip all the keys starting with underscore.

> to make sure there was only a config parameter, I preferred to write in this way.

As far as I know, this is self-imposed. I don't mind as long as it makes sense, although I prefer a flexible server and let the client do what it wants
The problem I see is that the requests don't look "natural". If you want to save a key-value pair, you normally want to send something like "key=value", or something along those lines ({key: value} for json, etc).

karakayasemi:
@jvillafanez Now, the controller is accepting both single and multiple inputs and responding with accepted-rejected keys in the body. If it can not set at least one key, it is returning `HTTP::Status_Bad_Request`.  I need your review one more time. Thanks.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,104 @@
...
+		$params = $this->getUserRequestParams();
+		foreach ($params as $key => $value) {
+			if ($this->validateKey($key) && ($value === 'yes' || $value === 'no')) {
+				$this->config->setUserValue(
+					$this->userSession->getUser()->getUID(),
+					'files_sharing',
+					$key,
+					$value
+				);
+				\array_push($acceptedKeys, $key);
```
jvillafanez: `$acceptedKeys[] = $key` should perform faster, unless you push several items, which isn't the case.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @param array $config
+	 * @return DataResponse
+	 */
+	public function setUserConfig() {
+		$acceptedKeys = [];
+		$rejectedKeys = [];
+
+		$params = $this->getUserRequestParams();
+		foreach ($params as $key => $value) {
+			if ($this->validateKey($key) && ($value === 'yes' || $value === 'no')) {
```
jvillafanez: I don't see much value in the `validateKey` function. If it won't have more checks, it's better to use just the wrapped function (the `inArray`). Or is there any plan to improve / increase the validation there?

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,104 @@
...
+	public function __construct(string $appName, IRequest $request, IConfig $config, IUserSession $userSession) {
+		parent::__construct($appName, $request);
+		$this->config = $config;
+		$this->userSession = $userSession;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @param array $config
```
jvillafanez: Adjust docs

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,104 @@
...
+
+	/**
+	 * Exclude params starting with '_' from request parameters
+	 *
+	 * @return array
+	 */
+	private function getUserRequestParams() {
+		return \array_filter($this->request->getParams(),
+			function ($key) {
+				return (\substr($key, 0, 1) !== '_');
```
jvillafanez: `strpos($key, '_') === 0` should perform faster according to https://stackoverflow.com/a/1962244 (we could use `$key[0] === '_'` but we need another check to verify the string isn't empty, so it might not be faster)

jvillafanez (commented):
last touches and we're good to go

apps/files_sharing/lib/Controller/PersonalSettingsController.php
karakayasemi: adjusted

apps/files_sharing/lib/Controller/PersonalSettingsController.php
karakayasemi: I changed the function to validate key-value pair. IMHO, it is improving readability and we can add specific controls to this function when validating future key-values.

apps/files_sharing/lib/Controller/PersonalSettingsController.php
karakayasemi: I don't think that's right for our case. 
- `strpos` should iterate over all string to determine a non-existing character's position. Its worst-case time-complexity should be the length of string O(N). 
-  substr complexity should be the length of substring O(k), which is 1 for our case O(1).  I think substr should work faster here. 
Anyway, IMHO time performance is not a big deal for our function. Nothing changes in readability, I left it the same way.

jvillafanez (approved):

apps/files_sharing/lib/Controller/PersonalSettingsController.php
```diff
@@ -0,0 +1,104 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+class PersonalSettingsController extends Controller {
+	const USERCONFIGS = [
+		'auto_accept_share'
```
PVince81: a bit over-engineering for just a single key...

PVince81 (approved):
:+1: I'll accept this for now so we don't invest any more time

PVince81:
@karakayasemi please backport

phil-davis:
Backport `stable10` #34647

PR #32213, opened 2018-08-01
[wip] cleanup makefile

patrickjahns:
## Description
Housekeeping for the Makefile + Dependencies to be great again

## Motivation and Context
We want the Makefile to be the entrypoint for developers 

First change is ab out requiring composer per default, and not fetching it for builts/locally 

TODO:
- [ ] introduce composer-bin and migrate (CI) tooling into composer-bin namespaces

## How Has This Been Tested?
🤖 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34630, opened 2019-02-27
Makefile composer

phil-davis:
## Description
- modify `Makefile` so it expects and finds a pre-installed `composer` on the system, rather than downloading a copy.
- modify `.drone.yml` so it no longer tries to self-install `composer`

This makes composer-related stuff work like it does in the app repos.

This is a resolved version of PR #32213 and moves forward from that.

## Related Issue
https://github.com/owncloud/QA/issues/582

## Motivation and Context
Have a consistent CI and development environment in core as in apps.

## How Has This Been Tested?
Local 
```
make clean
make
make test-php-style
make test-php
```
to see that these still work locally.

## Types of changes
- [ ] Technical debt
- [x] Tests only (no source changes) (and CI scripting)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -61,7 +61,7 @@ pipeline:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
-      - ./tests/drone/composer-install.sh
+      - make install-composer-deps
     when:
       event: [push, pull_request]
 
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,9 @@ KARMA=$(NODE_PREFIX)/node_modules/.bin/karma
 JSDOC=$(NODE_PREFIX)/node_modules/.bin/jsdoc
 PHPUNIT="$(shell pwd)/lib/composer/phpunit/phpunit/phpunit"
 COMPOSER_BIN := $(shell command -v composer 2> /dev/null)
+ifndef COMPOSER_BIN
+    $(error composer is not available on your system, please install composer)
+endif
 
 # bin file definitions
 PHP_CS_FIXER=php -d zend.enable_gc=0 vendor-bin/owncloud-codestyle/vendor/bin/php-cs-fixer
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,6 @@ update-composer:
 
 .PHONY: clean-composer-deps
 clean-composer-deps:
-	rm -f $(COMPOSER_BIN)
 	rm -Rf lib/composer
 	rm -Rf vendor-bin/**/vendor vendor-bin/**/composer.lock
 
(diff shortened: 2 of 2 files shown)


.drone.yml
```diff
@@ -61,7 +61,7 @@ pipeline:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
     commands:
-      - ./tests/drone/composer-install.sh
+      - make install-composer-deps
```
PVince81: somehow I had the feeling that all drone scripts were avoiding the Makefile for some reason... not sure if it's good to add it back

Makefile
```diff
@@ -38,7 +41,7 @@ YARN := $(shell command -v yarn 2> /dev/null)
 KARMA=$(NODE_PREFIX)/node_modules/.bin/karma
 JSDOC=$(NODE_PREFIX)/node_modules/.bin/jsdoc
 PHPUNIT="$(shell pwd)/lib/composer/phpunit/phpunit/phpunit"
-COMPOSER_BIN=build/composer.phar
+COMPOSER_BIN := $(shell command -v composer 2> /dev/null)
```
PVince81: what if there is no composer ? when to show a warning ?

note that in some build paths like JS tests no composer is required for Drone, so if a warning is shown it must only apply whenever composer is really needed

.drone.yml
phil-davis: In recent times we have been using the `make` targets directly in the drone scripts - e.g. `make test-acceptance-api` etc. This has happened in both core and apps.

Makefile
phil-davis: I can add the "standard" code block that has been added in app make files:
```
ifndef COMPOSER_BIN
    $(error composer is not available on your system, please install composer)
endif
```
We never made it smart in the app make files! It is likely emitting this warning in some drone build jobs that do not have `composer` in their docker, and also do not need it.

I will think about how to make it smarter.

phil-davis: I looked at drone. It has a `composer:` pipeline step that uses `image: owncloudci/php` which has `composer` available and works.
Javascript runs `test-javascript.sh` and does not go via `Makefile` so it does not notice if `composer` is there or not. In any case it uses the same docker image that has `composer` in it.

`phan` `phpstan` `owncloud-coding-standard` use `make` commands. They use `php` and `composer` anyway and have that docker images also.

So everything relevant in drone has `composer` already installed.

I have added the message block in my post above ^
That will only appear when a developer first does some `make` command and they do not have `composer` installed on their system. They should install it ;)

phil-davis:
@PVince81 @individual-it @patrickjahns ready for review again

PVince81 (approved):
fine :+1:

PVince81:
@phil-davis please backport

phil-davis:
Backport `stable10` #34637

PR #34659, opened 2019-03-02
allow admins to enable medial search on group and user

### LINKED ISSUE #33883: [FR] Global user and group medial search boolean option
we have `accounts.enable_medial_search` but we should move to one option respected globally for users, and one for groups, making the results more consistent. 

Eg: custom groups backend
Eg: local groups backend
Eg: local user backend
Eg: ldap user backend

karakayasemi:
## Description
allow admin to enable medial search on group and user

## Related Issue
Fixes #33883 and https://github.com/owncloud/enterprise/issues/2311

## Motivation and Context
we have accounts.enable_medial_search but we should move to one option respected globally for users, and one for groups, making the results more consistent.

## How Has This Been Tested?
- Create a group with "test" gid,
- add `'groups.enable_medial_search' => true` line to config
- Sharing search dialog should list "test" group when you write "es"
- change config as `'groups.enable_medial_search' => false` 
-  Sharing search dialog should not list "test" group when you write "es"

## Types of changes

- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Group/Database.php
+++ b/lib/private/Group/Database.php
@@ -50,13 +50,21 @@ class Database extends \OC\Group\Backend {
 	/** @var \OCP\IDBConnection */
 	private $dbConn;
 
+	/** @var \OCP\IConfig */
+	private $config;
+
 	/**
 	 * \OC\Group\Database constructor.
 	 *
 	 * @param \OCP\IDBConnection|null $dbConn
+	 * @param \OCP\IConfig|null $config
 	 */
-	public function __construct(\OCP\IDBConnection $dbConn = null) {
+	public function __construct(
+		\OCP\IDBConnection $dbConn = null,
+		\OCP\IConfig $config = null
+	) {
 		$this->dbConn = $dbConn;
+		$this->config = $config;
 	}
 
 	/**
@@ -65,6 +73,9 @@ class Database extends \OC\Group\Backend {
 	private function fixDI() {
 		if ($this->dbConn === null) {
 			$this->dbConn = \OC::$server->getDatabaseConnection();
+		}
+		if ($this->config === null) {
+			$this->config = \OC::$server->getConfig();
 		}
 	}
 
@@ -231,10 +242,18 @@ class Database extends \OC\Group\Backend {
 	 * Returns a list with all groups
 	 */
 	public function getGroups($search = '', $limit = null, $offset = null) {
+		$this->fixDI();
+
 		$parameters = [];
 		$searchLike = '';
 		if ($search !== '') {
-			$parameters[] = '%' . $search . '%';
+			$search = $this->dbConn->escapeLikeParameter($search);
+			$allowMedialSearches = $this->config->getSystemValue("groups.enable_medial_search", true);
+			if ($allowMedialSearches) {
+				$parameters[] = '%' . $search . '%';
+			} else {
+				$parameters[] = $search . '%';
+			}
 			$searchLike = ' WHERE LOWER(`gid`) LIKE LOWER(?)';
 		}
 
@@ -284,6 +303,8 @@ class Database extends \OC\Group\Backend {
 	 * @return array an array of user ids
 	 */
 	public function usersInGroup($gid, $search = '', $limit = null, $offset = null) {
+		$this->fixDI();
+
 		$parameters = [$gid];
 		$searchLike = '';
 		if ($search !== '') {
@@ -310,6 +331,8 @@ class Database extends \OC\Group\Backend {
 	 * @throws \OC\DatabaseException
 	 */
 	public function countUsersInGroup($gid, $search = '') {
+		$this->fixDI();
+
 		$parameters = [$gid];
 		$searchLike = '';
 		if ($search !== '') {
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -285,6 +285,13 @@ $CONFIG = array(
 'accounts.enable_medial_search' => true,
 
 /**
+ * Allow medial search on the group id. Allows finding 'test' when searching for 'es'.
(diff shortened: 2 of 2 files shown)


lib/private/Group/Database.php
```diff
@@ -313,7 +325,13 @@ public function countUsersInGroup($gid, $search = '') {
 		$parameters = [$gid];
 		$searchLike = '';
 		if ($search !== '') {
-			$parameters[] = '%' . $this->dbConn->escapeLikeParameter($search) . '%';
+			$search = \OC::$server->getDatabaseConnection()->escapeLikeParameter($search);
+			$allowMedialSearches = \OC::$server->getConfig()->getSystemValue("accounts.enable_medial_search", true);
```
jvillafanez: Try to inject the services

jvillafanez:
If the option will be used only for DB groups, it should be clarified in the sample file.

lib/private/Group/Database.php
karakayasemi: I injected services in this class and applied `fixDI` hack on changed functions.  But, `User/Database`  class a bit tricky, I don't prefer to touch this class for dependency injection in the scope of this pull request.

karakayasemi:
> If the option will be used only for DB groups, it should be clarified in the sample file.

A sentence was added for this purpose to sample.config.php.

jvillafanez:
> A sentence was added for this purpose to sample.config.php.

If not sure if it's clear enough...
We have several backends (DB, LDAP, shibboleth, ...). That flag will only apply to the DB backend, and not any other.

lib/private/Group/Database.php
```diff
@@ -284,10 +303,18 @@ public function groupExists($gid) {
...
 	 */
 	public function usersInGroup($gid, $search = '', $limit = null, $offset = null) {
+		$this->fixDI();
+
 		$parameters = [$gid];
 		$searchLike = '';
 		if ($search !== '') {
-			$parameters[] = '%' . $this->dbConn->escapeLikeParameter($search) . '%';
+			$search = $this->dbConn->escapeLikeParameter($search);
+			$allowMedialSearches = $this->config->getSystemValue("accounts.enable_medial_search", true);
```
jvillafanez: I'd rather not use this flag here. The flag is intended to be used for accounts, which means "**synced** users regardless of the procedence".
If needed, I'd prefer to use another different flag just to be used here for this case.

karakayasemi: flag removed.

karakayasemi:
> If not sure if it's clear enough...
> We have several backends (DB, LDAP, shibboleth, ...). That flag will only apply to the DB backend, and not any other.

@jvillafanez, as far as I understand from following sentences of this comment https://github.com/owncloud/core/issues/33883#issuecomment-468623653, it should apply to only DB backend: 
> The key thing should be that within ownCloud - all account searches are only hitting the database so we have control over the expectations for the search time.

If you have any suggestion, we can improve the explanation line of `config.sample.php`

jvillafanez:
The "accounts.enable_medial_search" key has to apply only to account searches, so using it against an specific backend is wrong.

I'm not sure if using medial search for users inside the DB backend (not in accounts) has any use. We'll need to sync the users into the account table, and the user manager checks against the account table, so a "userbackend.db.enable_medial_search" key affecting just the DB backend has a very limited usage.
For groups is different because we're still checking against the backend.

---

```
* Allow medial search on group properties like gid and other search terms.
* Allows finding 'test' when searching for 'es'.
* It is used in group database queries and may slow down group search.
```
What about
```
* Allow medial search on the group id. Allows finding 'test' when searching for 'es'.
* This is only used in the DB group backend (local groups), and won't be used against LDAP
* Shibboleth or any other group backend.
```
It's more clear that it will apply on the group id (there is no other usage right now), and also that it will apply only to the DB backend.

karakayasemi:
@jvillafanez  I changed the `sample.config.php` as you suggest.

lib/private/User/Database.php
```diff
@@ -184,8 +184,14 @@ public function getDisplayNames($search = '', $limit = null, $offset = null) {
 		$searchLike = '';
 		if ($search !== '') {
 			$search = \OC::$server->getDatabaseConnection()->escapeLikeParameter($search);
-			$parameters[] = '%' . $search . '%';
-			$parameters[] = '%' . $search . '%';
+			$allowMedialSearches = \OC::$server->getConfig()->getSystemValue("accounts.enable_medial_search", true);
```
jvillafanez: I'm still very concern about using this key here. I'd prefer to use a new one just to be used for this case instead of reusing another key which was created for another purpose.

lib/private/Group/Database.php
PVince81: @karakayasemi you can simply add a "TODO: inject" then

lib/private/User/Database.php
PVince81: we shouldn't touch / reuse the old keys and only work with new ones as discussed here: https://github.com/owncloud/core/issues/33883#issuecomment-466057804

karakayasemi: I just realized that this function is only used in the account sync operation. I believe, this time I understand @jvillafanez's concern.
@PVince81, IMHO, there is no much benefit on adding a key like `userbackend.db.enable_medial_search` as @jvillafanez said. I removed `User/Database` changes from PR.

jvillafanez (approved):

karakayasemi:
@ownclouders rebase

ownclouders:

ownclouders:

PR #34660, opened 2019-03-04
[PoC] Provide common make files

phil-davis:
See discussion below. This PR is being left here for now, until we get back to designing a better solution.

## Description
Provide various `*.mk` files in core that can be included by app `Makefile` so that chunks of `make` code do not have to be repeated in every app.

`check-composer.mk` - the common block of code that gives the warning if composer does not exist. This is included by other includes that care about composer.
`check-npm.mk` - the common block of code that gives the warning if npm does not exist. This is included by other includes that care about npm.
`clean.mk` - common `clean` and `clean_deps` targets. Other `clean` targets are added by other include files.
`dist.mk` - the "standard" way of making the tarball, and of doing a `clean dist`
`help.mk` - the "standard" way that help output is done these days in apps
`test-acceptance.mk` - the make targets for acceptance tests and their dependency rules
`test-js.mk` - the JS testing make targets and their dependency rules
`test-php.mk` - the PHP coding standard, analysis and unit test make targets and their dependency rules
`test-all.mk` - a convenience that includes all 3 `test-*.mk` files - so "standard" apps can just include this to get "the works" (now I am not so sure this is useful - not many apps have JS tests, some of those do or do not have acceptance tests,... so there is not really a "common" set of PHP+JS+acceptance tests)
`vendor-bin.mk` - the dependency rules for getting the `vendor` and `vendor-bin` stuff happening and cleaned. This is included by other includes that need it.

## Related Issue
https://github.com/owncloud/QA/issues/604

## How Has This Been Tested?
Local runs of JS unit, PHP unit and acceptance tests from various app repos.
PoC drone CI runs from activity, password_policy and... apps.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/build/rules/clean.mk
+++ b/build/rules/clean.mk
@@ -1,0 +1,16 @@
+##
+## Common clean rules
+##--------------------------------------
+
+ifndef CLEAN_HAS_BEEN_INCLUDED
+	CLEAN_HAS_BEEN_INCLUDED=true
+
+.PHONY: clean-deps
+clean-deps: ## Clean all dependencies
+clean-deps: $(clean_deps_rules)
+
+.PHONY: clean
+clean: ## Clean all dependencies, build and dist
+clean: clean-deps $(clean_dist_rules) $(clean_build_rules)
+
+endif
--- a/build/rules/dist.mk
+++ b/build/rules/dist.mk
@@ -28,7 +28,16 @@ endif
 
 # If not specified by the caller, then select some common doc files to put in the tarball.
 ifndef doc_files
-	doc_files=CHANGELOG.md CONTRIBUTING.md LICENSE README.md
+	doc_files=CHANGELOG.md README.md
+ifneq ("$(wildcard LICENSE)","")
+	doc_files+=LICENSE
+endif
+ifneq ("$(wildcard LICENSE.md)","")
+	doc_files+=LICENSE.md
+endif
+ifneq ("$(wildcard CONTRIBUTING.md)","")
+	doc_files+=CONTRIBUTING.md
+endif
 endif
 
 # If not specified by the caller, then select a "standard" set of dirs to put in the tarball.
--- a/build/rules/dist.mk
+++ b/build/rules/dist.mk
@@ -3,7 +3,11 @@
 ##--------------------------------------
 
 # signing
-occ=$(CURDIR)/../../occ
+PATH_TO_OCC=$(CURDIR)/occ
+ifeq ("$(wildcard $(PATH_TO_OCC))","")
+	PATH_TO_OCC=$(CURDIR)/../../occ
+endif
+occ=$(PATH_TO_OCC)
 private_key=$(HOME)/.owncloud/certificates/$(app_name).key
 certificate=$(HOME)/.owncloud/certificates/$(app_name).crt
 sign=$(occ) integrity:sign-app --privateKey="$(private_key)" --certificate="$(certificate)"
--- a/build/rules/dist.mk
+++ b/build/rules/dist.mk
@@ -1,0 +1,77 @@
+##
+## Distribution tarball
+##--------------------------------------
+
+# signing
+occ=$(CURDIR)/../../occ
+private_key=$(HOME)/.owncloud/certificates/$(app_name).key
+certificate=$(HOME)/.owncloud/certificates/$(app_name).crt
+sign=$(occ) integrity:sign-app --privateKey="$(private_key)" --certificate="$(certificate)"
+sign_skip_msg="Skipping signing, either no key and certificate found in $(private_key) and $(certificate) or occ can not be found at $(occ)"
+ifneq (,$(wildcard $(private_key)))
+ifneq (,$(wildcard $(certificate)))
+ifneq (,$(wildcard $(occ)))
+	CAN_SIGN=true
+endif
+endif
+endif
+
+# If the app_name has not been specified by the caller,
+# then use the name of the directory that the main "make" is being run from.
+ifndef app_name
+	app_name=$(notdir $(CURDIR))
+endif
+
+# If not specified by the caller, then select some common doc files to put in the tarball.
+ifndef doc_files
+	doc_files=README.md CHANGELOG.md
+endif
+
+# If not specified by the caller, then select a "standard" set of dirs to put in the tarball.
+ifndef src_dirs
+	src_dirs=appinfo css img js l10n lib templates
+endif
+
+all_src=$(src_dirs) $(extra_dirs) $(doc_files) $(extra_files)
+
+# Put the tarball and any other artifacts in a build dir by default
+ifndef build_dir
+	build_dir=$(CURDIR)/build
+endif
+
+# Put the tarball in a dist sub-dir of build_dir by default
+ifndef dist_dir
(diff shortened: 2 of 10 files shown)


build/rules/dist.mk
```diff
@@ -0,0 +1,81 @@
...
+
+# If the app_name has not been specified by the caller,
+# then use the name of the directory that the main "make" is being run from.
+ifndef app_name
+	app_name=$(notdir $(CURDIR))
+endif
+
+# If not specified by the caller, then select some common doc files to put in the tarball.
+ifndef doc_files
+	doc_files=README.md CHANGELOG.md
```
phil-davis: Should `CONTRIBUTING.md` be in this `doc_files` list. Most app repos have one, and it is easy enough to add one. For private repos it can say something different, because the private repo won't be publicly visible to contribute to :)

phil-davis:
Questions:

- [ ] what to put in the default `doc_files` list?
- [ ] what to put in the default `src_files` list?
- [ ] what to do about `app:check-code` ? put it in a common `test-app-check-code.mk` in core? or just leave that block in the `Makefile` of apps that happen to have it? or remove it everywhere? or?

build/rules/dist.mk
individual-it: I would vote to add `CONTRIBUTING.md` and `LICENSE`

build/rules/clean.mk
```diff
@@ -0,0 +1,16 @@
+##
+## Common clean rules
+##--------------------------------------
+
+ifndef CLEAN_HAS_BEEN_INCLUDED
+	CLEAN_HAS_BEEN_INCLUDED=true
+
+.PHONY: clean-deps
+clean-deps: ## Clean all dependencies
```
PVince81: I think the clean rules need to be part of each module

the dist rules module has clean-dist, the npm deps has clean-npm-deps, etc

you can use a generic variable to which every module can add their clean rules to the main "clean" target.
see https://github.com/owncloud/customgroups/blob/master/Makefile#L25 and https://github.com/owncloud/customgroups/blob/master/rules/deps.mk#L16

individual-it:
don't think we need `test-all.mk`

PVince81:
there is one drawback with this: app builds will now depend on the core git checkout that was used.

which means that if we fix makefile rules in core we might need to do another release of apps.

we'll need to settle on a single OC version when buildling apps then, we could say it always needs to be from the git checkout of the latest stable release.

note that we had an internal POC from @patrickjahns where we build apps using a docker based on the OC-10.0.10 release. this approach would not work any more because the release tarball does not contain the makefile rules.

might need a different approach then, like putting the makefile rules in a common repo like the coding standards one ?

build/rules/vendor-bin.mk
```diff
@@ -0,0 +1,25 @@
...
+composer.lock: composer.json
+	@echo composer.lock is not up to date.
+
+vendor: composer.lock
+	composer install --no-dev
+
+vendor/bamarni/composer-bin-plugin: composer.lock
+	composer install
+
+clean_deps_rules+=clean-vendor clean-vendor-bin
```
PVince81: ah, you aready have this

PVince81:
ideas for later: a rules file with checkers.

as this will be used by apps, we can check that many things are already properly in place in the dist tarball, like info.xml, changelog.md exists, etc

build/rules/dist.mk
```diff
@@ -0,0 +1,81 @@
...
+
+# If the app_name has not been specified by the caller,
+# then use the name of the directory that the main "make" is being run from.
+ifndef app_name
+	app_name=$(notdir $(CURDIR))
+endif
+
+# If not specified by the caller, then select some common doc files to put in the tarball.
+ifndef doc_files
+	doc_files=CHANGELOG.md CONTRIBUTING.md LICENSE README.md
```
phil-davis: Some app repos have `LICENSE.md` others have just `LICENSE`

Do we set a standard? Or do I write some more fancy code that sucks in whichever file exists?

build/rules/dist.mk
phil-davis: Now it adds `CONTRIBUTING.md` `LICENSE` and/or `LICENSE.md` only if they exist.
That makes it flexible and we do not have to force a decision here to always require these in every repo.

build/rules/dist.mk
phil-davis: I made it flexible so it adds whichever file exists.

phil-davis:
> which means that if we fix makefile rules in core we might need to do another release of apps.

> we'll need to settle on a single OC version when buildling apps then, we could say it always needs to be from the git checkout of the latest stable release.

If the problem caused an app tarball to be released with missing/wrong stuff, then fir sure the app has to be released again. But that is already the case if there is a problem in the local `Makefile`

The default "recommended" action when making an app release tarball would be to have the end of the current core release branch checked out, e.g. core `stable10`. That way you get the included `*.mk` as they were at the last core release, plus any recent fixes that will end up in the next release tag. But also in most cases core `master` will have exactly the same `*.mk` as core `stable10` so it will not make a real difference.

phil-davis:
@PVince81 IMO this now matches the "standard" `Makefile` code in the apps, and has flexibility to deal with finding various files. It can reduce the `Makefile` code duplication across the apps. 

If we find we need more flexibility, then it is easy to add `ifdef` etc conditional statements so that it continues to work for apps that use it, but can also handle some more unusual thing we find in "an app I haven't thought of yet".

Or if the app really is "weird", then it does not have to include all the `*.mk` files, and can continue to "do its thing its own way" until we really sort it out. I don't believe there are any apps like that.

If we some day want to migrate to putting this in a separate repo, or use some completely different build tool, then that will be possible and should be easier because we have this common set of knowledge about what build and test services are commonly used by apps.

Please review again.

PVince81:
> If we some day want to migrate to putting this in a separate repo, or use some completely different build tool, then that will be possible and should be easier because we have this common set of knowledge about what build and test services are commonly used by apps.

Would be better to decide this right away. We're going to need to touch every app already for this change to make sense. If later on we introduce a common location, we'd need to touch all apps yet again.

phil-davis:
OK, so how do we decide?

patrickjahns:
Please do not make apps depend on core makefiles. 

If we are looking for a central part for makefiles, please put them in a dedicated repository to have clear separation of concerns. 
Core repository should be about ownCloud core and not about a build/makefile infrastructure for apps

phil-davis:
I removed the "To review" label. 
This PR is now a "demonstration" of the makefile targets that could be put somewhere common, and a way that those could be split into logical parts in `*.mk` files.
We need to research a good solution for where and how to store and retrieve this common build infrastructure - issue https://github.com/owncloud/QA/issues/604
And then perhaps be able to efficiently copy-paste some structure and code from here into the selected separate repo/tool/whatever.

PVince81 (changes requested):
:-1: for now (to prevent accidental merge)

phil-davis:
Closing see comment https://github.com/owncloud/richdocuments/pull/253#issuecomment-513429624
There will be a different approach "some day"

PR #34687, opened 2019-03-05
Renaming and adding new options for public link

sharidas:
1) Renaming View/Download/Upload to View/Download/Edit.
2) Adding new option View/Download/Upload.
So if user wants to give provide all permissions,
View/Download/Edit option could be opted. If the user
wants to have an option where no edit and delete to be
granted for link shared, the View/Download/Upload could
be used.

## Description

This PR renames and adds an option in the public link share dialog. The `View/Download/Upload` is now renamed to `View/Download/Edit`. So if the user wants to create a link which would let others to create, edit, upload and delete in the public link then `View/Download/Edit` should be selected.
The new option `View/Download/Upload` has lesser permission. When a public link with this option is created, the user could create folder(s), no delete option is available, no edit possible. By no edit option, I mean no re-write of the existing files. Say for example if there is a file `a.txt` already available in the public link, and user tries to uplaod a file `a.txt`, the end result would be like this: the newly uploaded file would become `a (2).txt`. The original file `a.txt` would not be overwritten.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

- Renaming `View/Download/Upload` to `View/Download/Edit` in the public link share creation dialog.
- Created `View/Download/Upload` with minimal permission. Removed edit and delete option.

## How Has This Been Tested?

- [x] Here is the screenshot 
granularPermission

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -829,7 +829,11 @@ OC.Uploader.prototype = _.extend({
 				return true;
 			}
 			var fileInfo = fileList.findFile(file.name);
-			if (fileInfo) {
+			var sharePermission = $("#sharePermission").val();
+			if (sharePermission !== undefined) {
+				sharePermission &= (OC.PERMISSION_READ | OC.PERMISSION_UPDATE | OC.PERMISSION_CREATE | OC.PERMISSION_DELETE);
+			}
+			if (fileInfo && (sharePermission !== (OC.PERMISSION_READ | OC.PERMISSION_UPDATE | OC.PERMISSION_CREATE))) {
 				conflicts.push([
 					// original
 					_.extend(fileInfo, {
@@ -839,13 +843,24 @@ OC.Uploader.prototype = _.extend({
 					upload
 				]);
 				return false;
+			} else {
+				/**
+				 * The public link with view-download-upload option. Here we would
+				 * not overwrite the files. Instead create new file by autorenaming,
+				 * if the file already exists.
+				 *
+				 */
+				if (fileInfo) {
+					callbacks.onAutoRenameForPublicLink(upload);
+				}
 			}
 			return true;
 		});
+
 		if (conflicts.length) {
 			// wait for template loading
-			OC.dialogs.fileexists(null, null, null, this).done(function() {
-				_.each(conflicts, function(conflictData) {
+			OC.dialogs.fileexists(null, null, null, this).done(function () {
+				_.each(conflicts, function (conflictData) {
 					OC.dialogs.fileexists(conflictData[1], conflictData[0], conflictData[1].getFile(), this);
 				});
 			});
@@ -1096,6 +1111,9 @@ OC.Uploader.prototype = _.extend({
 
 							onNoConflicts: function (selection) {
 								self.submitUploads(selection.uploads);
+							},
+							onAutoRenameForPublicLink: function(upload) {
+								self.onAutorename(upload);
 							},
 							onSkipConflicts: function (selection) {
 								//TODO mark conflicting files as toskip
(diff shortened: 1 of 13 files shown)


sharidas:
Here is the snapshot of how this feature looks and behaves, kindly have a look at the screencast: 
share

apps/files/js/file-upload.js
```diff
@@ -839,13 +843,24 @@ OC.Uploader.prototype = _.extend({
...
 				return false;
+			} else {
+				/**
+				 * The public link with view-download-upload option. Here we would
+				 * not overwrite the files. Instead create new file by autorenaming,
+				 * if the file already exists.
+				 *
+				 */
+				if (fileInfo) {
+					callbacks.onAutoRenameForPublicLink(upload);
```
PVince81: what if we have update permissions but not create ones ? in this case autorename cannot work but overwrite would.

I wonder if we should still show a dialog at least for the latter.

to me it sounds like this part would rather belong to another PR, because the conflict dialog handling is not specific to public link but also applies to local shares and I'm not 100% clear about what the expected behavior should be here

apps/files/js/file-upload.js
```diff
@@ -1097,6 +1112,9 @@ OC.Uploader.prototype = _.extend({
 							onNoConflicts: function (selection) {
 								self.submitUploads(selection.uploads);
 							},
+							onAutoRenameForPublicLink: function(upload) {
```
PVince81: not specific to public link. please try the same permissions setup for a local share

tests/acceptance/features/lib/FilesPageElement/SharingDialogElement/EditPublicLinkPopup.php
```diff
@@ -50,6 +50,7 @@ class EditPublicLinkPopup extends OwncloudPage {
 	private $permissionLabelXpath = [
 		'read' => ".//label[contains(@for, 'sharingDialogAllowPublicRead')]",
 		'read-write' => ".//label[contains(@for, 'sharingDialogAllowPublicReadWrite')]",
+		'upload-wirte-without-modify' =>  ".//label[contains(@for, 'sharingDialogAllowpublicUploadWrite')]",
```
PVince81: s/wirte/write/

PVince81 (commented):
Looks fine in general.
Please check behavior with local shares. If conflict dialog has the same issue then we should rather move the conflict dialog fix to a separate PR / topic.

apps/files/js/file-upload.js
sharidas: Ok, then if I understand the comment properly, I would remove autorename changes from this PR. That would make the PR simple. Which means:
- Conflict dialog would behave as it is. 
   - That is the user will have the option to overwrite the existing file or keep the original file or add a new file with auto-rename.
Let me know if this sounds ok?

PVince81: yes. leave the conflict dialog changes for another PR as it will likely need more discussion regarding the expected behavior

tests/acceptance/features/lib/FilesPageElement/SharingDialogElement/EditPublicLinkPopup.php
sharidas: Done.

apps/files/js/file-upload.js
sharidas: Done

PVince81 (approved):
if you retested this and still works then it's good to go :+1:

sharidas:
@PVince81 Yes I have tested this and I have uploaded the screenshare in the PR template.

PVince81:
@sharidas please backport

sharidas:
Backport PR https://github.com/owncloud/core/pull/34983

PR #34690, opened 2019-03-05
Add webdav trash bin endpoint

DeepDiver1975:
## Description
We want a WebDAV API as discussed in #15646 

This is how it looks like:
- remote.php/dav/trash-bin is the root folder for all trashbins - each user will find their own trahsbin inside
- remote.php/dav/trash-bin/admin is the users root for trashbins
- remote.php/dav/trash-bin/items holds all deleted items - items are listed by file id
- remote.php/dav/trash-bin/restore is a folder which is used to restore trashed items by moving them in there. This folder itself will not hold any items

## Related Issue
- refs #15646 

## How Has This Been Tested?
- :hand: 
- tests will follow

Screenshot from 2019-06-21 13-25-17
Screenshot from 2019-06-21 13-25-28
Screenshot from 2019-06-21 13-25-48
Screenshot from 2019-06-21 13-26-48

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/Connector/Sabre/CopyEtagHeaderPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/CopyEtagHeaderPlugin.php
@@ -22,6 +22,7 @@
 
 namespace OCA\DAV\Connector\Sabre;
 
+use Sabre\DAV\Exception\NotFound;
 use Sabre\HTTP\RequestInterface;
 use Sabre\HTTP\ResponseInterface;
 
@@ -71,11 +72,15 @@ class CopyEtagHeaderPlugin extends \Sabre\DAV\ServerPlugin {
 	 * @return void
 	 */
 	public function afterMove($source, $destination) {
-		$node = $this->server->tree->getNodeForPath($destination);
-		if ($node instanceof File) {
-			$eTag = $node->getETag();
-			$this->server->httpResponse->setHeader('OC-ETag', $eTag);
-			$this->server->httpResponse->setHeader('ETag', $eTag);
+		try {
+			$node = $this->server->tree->getNodeForPath($destination);
+			if ($node instanceof File) {
+				$eTag = $node->getETag();
+				$this->server->httpResponse->setHeader('OC-ETag', $eTag);
+				$this->server->httpResponse->setHeader('ETag', $eTag);
+			}
+		} catch (NotFound $ex) {
+			// nothing to do then ....
 		}
 	}
 }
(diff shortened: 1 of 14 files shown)


DeepDiver1975:
>     * remote.php/dav/trash-bin/restore is a folder which is used to restore trashed items by moving them in there. This folder itself will not hold any items

I'm not too happy with this .... we will not get the etag back in the move header - which is required by the clients. Maybe we need to add special logic here.

apps/dav/lib/Connector/Sabre/CopyEtagHeaderPlugin.php
```diff
@@ -71,11 +72,15 @@ public function afterMethod(RequestInterface $request, ResponseInterface $respon
...
-			$this->server->httpResponse->setHeader('ETag', $eTag);
+		try {
+			$node = $this->server->tree->getNodeForPath($destination);
+			if ($node instanceof File) {
+				$eTag = $node->getETag();
+				$this->server->httpResponse->setHeader('OC-ETag', $eTag);
+				$this->server->httpResponse->setHeader('ETag', $eTag);
+			}
+		} catch (NotFound $ex) {
+			// nothing to do then ....
```
DeepDiver1975: destination will be /trash-bin/$user/restore/$original which is not know because the restore folder will not list the content.

Maybe we do a 
```
mv trash-bin/$user/item/$fileId files/$user/$originalLocation

michaelstingl:
@davigonz @hosy @felix-schwarz Could this work for a mobile implementation? Any feedback? /cc @guruz @ckamm @ogoffart

DeepDiver1975:
As per discussion with @butonic we will go for the second approach which will give us more flexibility in terms of the restore location and naming.

DeepDiver1975:
> @davigonz @hosy @felix-schwarz Could this work for a mobile implementation? Any feedback? /cc @guruz @ckamm @ogoffart

please follow up in #35636

guruz:
I haven't looked at the code code but this sounds dangerous, what if I have a user called `restore` or `items`

> remote.php/dav/trash-bin/admin is the users root for trashbins

Assuming user is called 'admin'

> remote.php/dav/trash-bin/items holds all deleted items - items are listed by file id
> remote.php/dav/trash-bin/restore is a folder which is used to restore trashed items by moving them in there. This folder itself will not hold any items

guruz:
OK, seems like this is adressed, the description of https://github.com/owncloud/core/pull/35636 looks without the `items` and `restore`

PR #34692, opened 2019-03-06
[test] Do not use selenium debug on drone

phil-davis:
Results:

- taking off `-debug` from the selenium containers does not reduce the amount of logging output, but firefox seems to break more. So there is no point doing that.
- reducing the selenium log level to warning does reduce the output from, e.g. 700 lines to <200 lines. That seems worth doing.
- changing `apachectl` from `-e debug` to `-e error` has no significant effect on the size of log output. That is because the volume that is output is from the "access log" output, logging every GET, POST...

So the remaining challenge is to be able to not have the Apache access log output.


phil-davis:
experiments finished

PR #34753, opened 2019-03-12
Put back files:scan --groups option for backward-compatibility

phil-davis:
## Description
PR #31719 changed the `occ files:scan` command so that you specify groups like `--group=foo --group=bar` rather than `--groups=foo,bar` - which let's the user specify group names that contain a comma `--group=Finance,Sydney --group=Admin,Singapore`
 
That change was not backward-compatible and so would have to wait for a major version release.

But we can keep the previous behavior also, so that for groups without a comma in their name the user can use `--groups` just like they have always been able to.

That makes the net change a bug fix that adds a new feature/functionality.

So this is the code to do that.

## Related Issue
#31578
#31835
PR #31719 

Also related to comparing `master` with `stable10` and getting them closer together.

## How Has This Been Tested?
Local run of unit tests, and trying the `occ files:scan` command.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue) (recovers backward-compatibility)
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/742

## Open tasks:

### DIFF
--- a/apps/files/lib/Command/Scan.php
+++ b/apps/files/lib/Command/Scan.php
@@ -106,9 +106,15 @@ class Scan extends Base {
 			)
 			->addOption(
 				'group',
-				'g',
+				'',
 				InputOption::VALUE_IS_ARRAY|InputOption::VALUE_REQUIRED,
 				'Scan user(s) under the group(s). This option can be used as --group=foo --group=bar to scan groups foo and bar'
+			)
+			->addOption(
+				'groups',
+				'g',
+				InputArgument::OPTIONAL,
+				'Scan user(s) under the group(s). This option can be used as --groups=foo,bar to scan groups foo and bar'
 			)
 			->addOption(
 				'quiet',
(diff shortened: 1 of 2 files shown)


phil-davis:
If this goes ahead, then a backport can be done of the last 2 commits from #31719 plus the commit here.

mmattel:
I guess that this needs a documentation update to cover both variants, or? ...

phil-davis:
> I guess that this needs a documentation update to cover both variants, or? ...

Yes, if this gets accepted then the docs need to be adjusted to show that the "old" `--groups` command is the same and the new `--group` command is now also available.

PVince81 (approved):
:+1:

phil-davis:
Backport `stable10` #34754

PR #34764, opened 2019-03-14
dont allow user to override admin sharing settings when its disabled

### LINKED ISSUE #34708: user has to disable and enable share autoaccept setting to override disabled admin setting
### Steps to reproduce
1. create a new user
1. as admin disable "Automatically accept new incoming local user shares" 
1. check the user-based setting for auto-accepting shares

### Expected behaviour
setting should be disabled (the user never touched them)

### Actual behaviour
setting is enabled
but shares are not auto-accepted
to really enable them user has to disable and re-enable the setting

### LINKED ISSUE #34763: Automatically accept remote shares from trusted servers always enabled in sharing admin panel
### Steps to reproduce
1. Go Sharing Admin Panel, Federated Cloud Sharing section
2. Disable Automatically accept remote shares from trusted servers	
3. Refresh Admin page

### Expected behaviour
Automatically accept remote shares from trusted servers checkbox should be disabled. 

### Actual behaviour
it is always enabled.

`oc_appconfig` table is storing the correct value. The problem is only in the admin panel interface.

karakayasemi:
## Description
Implementation of https://github.com/owncloud/enterprise/issues/3128#issuecomment-471185144
- Hide sharing personal panel options from users if they are disabled by admin as a security concern.
- If they are enabled, allow users to make their own choice.
- Fix small UI issues on federatedfilesharing admin panel #34763
- Fix inconcistency between user and admin selection #34708

If all the options are disabled by admin, instead of showing empty page, I added `Nothing to configure`  sentence for each section.

## To-do:
- [x] Fix acceptance tests

## Related Issue
Fixes https://github.com/owncloud/enterprise/issues/3128 
Closes #34763
Closes #34708

## How Has This Been Tested?
- Disable auto_accept options from admin panel
- User should see `Nothing to configure` text in sharing personal panel
- Enable auto_accept options from admin panel
- User should allowed to make their own choice

## Types of changes

- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/federatedfilesharing/lib/Panels/AdminPanel.php
+++ b/apps/federatedfilesharing/lib/Panels/AdminPanel.php
@@ -57,7 +57,7 @@ class AdminPanel implements ISettings {
 		$tmpl->assign('incomingServer2serverShareEnabled', $this->shareProvider->isIncomingServer2serverShareEnabled());
 		$tmpl->assign(
 			'autoAcceptTrusted',
-			$this->config->getAppValue('federatedfilesharing', 'auto_accept_trusted', 'no')
+			$this->config->getAppValue('federatedfilesharing', 'auto_accept_trusted', 'no') === 'yes'
 		);
 		return $tmpl;
 	}
--- a/apps/federatedfilesharing/lib/Panels/SharingPersonalPanel.php
+++ b/apps/federatedfilesharing/lib/Panels/SharingPersonalPanel.php
@@ -46,17 +46,29 @@ class SharingPersonalPanel implements ISettings {
 	 * @return TemplateResponse | Template
 	 */
 	public function getPanel() {
+		$tmpl = new Template('federatedfilesharing', 'settings-personal-sharing');
+		$showEmptyTemplate = true;
+		$globalAutoAcceptShareEnabled  = $this->config->getAppValue(
+			'federatedfilesharing',
+			'auto_accept_trusted',
+			'no'
+		);
 		$autoAcceptShareEnabled = $this->config->getUserValue(
 			$this->userSession->getUser()->getUID(),
 			'federatedfilesharing',
 			'auto_accept_share_trusted',
-			'yes'
+			$globalAutoAcceptShareEnabled
 		);
-		$tmpl = new Template('federatedfilesharing', 'settings-personal-sharing');
-		$tmpl->assign(
-			'userAutoAcceptShareTrustedEnabled',
-			$autoAcceptShareEnabled
-		);
+		if ($globalAutoAcceptShareEnabled === 'yes') {
+			$showEmptyTemplate = false;
+			$tmpl->assign(
+				'userAutoAcceptShareTrustedEnabled',
+				$autoAcceptShareEnabled
+			);
+		}
+		if ($showEmptyTemplate) {
+			return new Template('federatedfilesharing', 'settings-personal-sharing-empty');
+		}
 		return $tmpl;
 	}
 
--- a/apps/federatedfilesharing/tests/Panels/SharingPersonalPanelTest.php
+++ b/apps/federatedfilesharing/tests/Panels/SharingPersonalPanelTest.php
@@ -56,7 +56,20 @@ class SharingPersonalPanelTest extends \Test\TestCase {
 		$this->assertEquals(40, $this->sharingPersonalPanel->getPriority());
 	}
 
-	public function testGetPanel() {
+	public function globalFederatedSharingConfigProvider() {
+		return [
+			[['auto_accept_trusted' => 'yes'], '<form class="section" id="federatedfilesharing_settings">'],
+			[['auto_accept_trusted' => 'no'], '<p>Nothing to configure.</p>'],
+		];
+	}
+
+	/**
+	 * @dataProvider globalFederatedSharingConfigProvider
+	 *
+	 * @param array $globalConfigs
+	 * @param string $expectedString
+	 */
+	public function testGetPanel($globalConfigs, $expectedString) {
 		$mockUser = $this->getMockBuilder(IUser::class)
 			->disableOriginalConstructor()
 			->getMock();
@@ -67,8 +80,12 @@ class SharingPersonalPanelTest extends \Test\TestCase {
 		$this->userSession->expects($this->any())
 			->method('getUser')
 			->willReturn($mockUser);
-
+		$this->config->expects($this->once())
+			->method('getAppValue')
+			->with('federatedfilesharing', 'auto_accept_trusted', 'no')
+			->willReturn($globalConfigs['auto_accept_trusted']);
+		
 		$templateHtml = $this->sharingPersonalPanel->getPanel()->fetchPage();
-		$this->assertContains('<form class="section" id="federatedfilesharing_settings">', $templateHtml);
+		$this->assertContains($expectedString, $templateHtml);
 	}
 }
--- a/tests/acceptance/features/bootstrap/WebUIPersonalSharingSettingsContext.php
+++ b/tests/acceptance/features/bootstrap/WebUIPersonalSharingSettingsContext.php
@@ -80,4 +80,26 @@ class WebUIPersonalSharingSettingsContext extends RawMinkContext implements Cont
 			$this->getSession(), $action
 		);
 	}
+
+	/**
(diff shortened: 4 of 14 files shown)


apps/federatedfilesharing/lib/Panels/AdminPanel.php
```diff
@@ -57,7 +57,7 @@ public function getPanel() {
 		$tmpl->assign('incomingServer2serverShareEnabled', $this->shareProvider->isIncomingServer2serverShareEnabled());
 		$tmpl->assign(
 			'autoAcceptTrusted',
-			$this->config->getAppValue('federatedfilesharing', 'auto_accept_trusted', 'no')
+			$this->config->getAppValue('federatedfilesharing', 'auto_accept_trusted', 'no') === 'yes'
```
karakayasemi: fix for #34763

apps/federatedfilesharing/lib/Panels/SharingPersonalPanel.php
```diff
@@ -46,16 +46,29 @@ public function __construct(IConfig $config, IUserSession $userSession) {
...
+			$tmpl->assign(
+				'userAutoAcceptShareTrustedEnabled',
+				$autoAcceptShareEnabled
+			);
+		}
 		$tmpl->assign(
-			'userAutoAcceptShareTrustedEnabled',
-			$autoAcceptShareEnabled
+			'showSection',
+			$showSection
```
jvillafanez: The name seems a bit too generic. Maybe "showFederatedFileSharingSection" shows better the purpose of the variable.

apps/federatedfilesharing/templates/settings-personal-sharing.php
```diff
@@ -21,16 +21,27 @@
 
 /** @var array $_ */
 /** @var \OCP\IL10N $l */
+
 script('federatedfilesharing', 'settings-personal-sharing');
 ?>
 
-<form class="section" id="federatedfilesharing_settings">
+<div class="section" id="federatedfilesharing_settings">
```
jvillafanez: I guess this change is intended, but not sure why.

jvillafanez:
for the template, how about having 2 different templates, one for the "nothing to configure" and another one for the normal content? I'm not sure it this approach can help with the unittests and simplify the code.

In addition, you're following the same approach for the federated sharing and "normal" sharing, is this what we want? (just checking)

apps/federatedfilesharing/templates/settings-personal-sharing.php
karakayasemi: Not important actually. For empty config case, I thought `div` is better than empty `form`.

karakayasemi:
> 
> for the template, how about having 2 different templates, one for the "nothing to configure" and another one for the normal content? I'm not sure it this approach can help with the unittests and simplify the code.

I think it makes sense. Using two different template can reduce template and panel complexity.

>
> In addition, you're following the same approach for the federated sharing and "normal" sharing, is this what we want? (just checking)

According to the agreement on this comment https://github.com/owncloud/enterprise/issues/3128#issuecomment-471185144, it is desired same functionality for bot federated and local sharing.

apps/federatedfilesharing/templates/settings-personal-sharing.php
jvillafanez: ok, just double-check there are no problems with the css styles.

PVince81:
some sharing tests failed, @karakayasemi please have a look.

if you need assistance with adjusting or extending UI tests, feel free to ask @individual-it or @phil-davis

karakayasemi:
@individual-it I tried to adapt acceptance tests to the new behavior, removed some of the issue tags. 
About acceptance testing I read the documentation, it is pretty explanatory by the way. But, I have no prior experience in it. Please review, acceptance tests part. Thanks.

tests/acceptance/features/bootstrap/WebUIPersonalSharingSettingsContext.php
```diff
@@ -80,4 +80,22 @@ public function switchAutoAcceptingFederatedShares($action) {
...
 		);
 	}
+
+	/**
+	 * @Then User-based auto accepting checkbox should not be displayed on the personal sharing settings page in the webUI
+	 *
+	 * @return void
+	 */
+	public function autoAcceptingCheckboxShouldNotBeDisplayedOnThePersonalSharingSettingsPageInTheWebui() {
+		PHPUnit_Framework_Assert::assertFalse($this->personalSharingSettingsPage->isAutoAcceptLocalSharesCheckboxDisplayed());
```
individual-it: please make the line shorter, that will make it easier to review
```suggestion
		PHPUnit_Framework_Assert::assertFalse(
		    $this->personalSharingSettingsPage->isAutoAcceptLocalSharesCheckboxDisplayed()
		);
```

individual-it:
the tests look very good to me, just don't like the long lines

very good, achievement unlocked :tada:

individual-it:
:+1: for the acceptance tests part

karakayasemi:
@jvillafanez I adjusted the code to your last review. Templates and panel simplified with creating two templates. Please review, one more time. Thanks.

apps/federatedfilesharing/templates/settings-personal-sharing.php
```diff
@@ -21,16 +21,21 @@
 
 /** @var array $_ */
 /** @var \OCP\IL10N $l */
+
 script('federatedfilesharing', 'settings-personal-sharing');
 ?>
 
 <form class="section" id="federatedfilesharing_settings">
 	<h2 class="app-name"><?php p($l->t('Federated Cloud Sharing'));?></h2>
+	<?php if (isset($_['userAutoAcceptShareTrustedEnabled'])) {
```
jvillafanez: For templates, I'd prefer to use "template syntax", something like:
```
<?php if (isset($_['userAutoAcceptShareTrustedEnabled'])): ?>
    <input ..... />
    <label></label>
<?php endif; ?>
```
it should be easier to follow big conditional blocks.

In addition, check readability:
```
<?php if ($_['userAutoAcceptShareTrustedEnabled'] === 'yes'): ?>
    <input type="checkbox" name="auto_accept_share_trusted" id="userAutoAcceptShareTrustedInput" class="checkbox" value="1" checked="checked" />
<?php else: ?>
    <input type="checkbox" name="auto_accept_share_trusted" id="userAutoAcceptShareTrustedInput" class="checkbox" value="1" />
<?php endif; ?>
```
These changes should make the template easier to read.

apps/federatedfilesharing/tests/Panels/SharingPersonalPanelTest.php
```diff
@@ -67,8 +80,16 @@ public function testGetPanel() {
...
 			->willReturn($mockUser);
-
+		$this->config->expects($this->once())
+			->method('getAppValue')
+			->with('federatedfilesharing', 'auto_accept_trusted', 'no')
+			->willReturn($globalConfigs['auto_accept_trusted']);
+		
 		$templateHtml = $this->sharingPersonalPanel->getPanel()->fetchPage();
-		$this->assertContains('<form class="section" id="federatedfilesharing_settings">', $templateHtml);
+		if ($showEmptyTemplate) {
```
jvillafanez: There is no need to change something (up to you), but I'd prefer not to have conditions in the tests. This is easy to follow, but there could be complex conditions and it might possible that some part of the test never executes (not this particular case though)
Alternatives could be:
* split into 2 different tests
* include the expected output in the data provider and check it in the test.

apps/federatedfilesharing/templates/settings-personal-sharing.php
karakayasemi: I was not aware of the `endif` control structures syntax. Thanks for the detailed explanation, I believe I got the idea.

apps/federatedfilesharing/tests/Panels/SharingPersonalPanelTest.php
karakayasemi: Tests are simplified with including expected output in the data provider.

karakayasemi:
@jvillafanez the code adjusted to your last suggestions, please review one more time. Thanks.

jvillafanez (approved):

PR #33233, opened 2018-10-19
Require dg/composer-cleaner

DeepDiver1975:
## Description
By adding https://github.com/dg/composer-cleaner we can reduce the size of the release packages.
Let's give it a try.

## Related Issue
- @butonic opened an issue somewhere ....

## How Has This Been Tested?
- :robot: 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)
- [ ] find something similar for npm/yarn

PR #34784, opened 2019-03-18
[stable10] add composer cleaner

phil-davis:
Backport #33233

Note: 3rd commit "Don't clean inside behat" has been backported for clarity (that all commits from the PR have been backported).
But since the original `master` PR #33233 `behat` has moved to `vendor-bin` and so the 3rd commit is no longer relevant. It is effectively reversed by the 4th commit.

### DIFF
--- a/build/package.json
+++ b/build/package.json
@@ -55,6 +55,6 @@
   },
   "scripts": {
     "clean-modules": "modclean -r --patterns=\"default:safe,owncloud:basic\"",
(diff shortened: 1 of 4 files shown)


PVince81 (approved):
:+1:

phil-davis:
Note: this also did a recently-released:
```
Updating pear/pear-core-minimal (v1.10.7 => v1.10.9)
```
because I did `composer update`

That is a good version bump anyway - PR #34795 does it in `master`

PR #31733, opened 2018-06-12
[stable10] unit test files_primary_s3 on core

patrickjahns:
## Description
port of https://github.com/owncloud/core/pull/31635/commits/4950f87bbbc83a479c66a7027e1ab47699ed6b44 

better approach then what is done in master already - would forward port the changes then again

## Motivation and Context
Test files_primary_s3 during PRs so changes to not necessarily break objectstorage integration 

## How Has This Been Tested?
🤖 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #34788, opened 2019-03-18
[master] unit test files_primary_s3 on core - from stable10 31733

phil-davis:
Forward port parts of PR #31733 to `master`

In order to get the logic of `.drone.yml` more similar in both `master` and `stable10`

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -144,6 +144,8 @@ pipeline:
   install-server:
     image: owncloudci/php:${PHP_VERSION}
     pull: true
+    environment:
+      - DB_TYPE=${DB_TYPE}
     commands:
       - ./tests/drone/install-server.sh
       - php occ a:l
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -731,21 +735,6 @@ matrix:
       INSTALL_SERVER: true
       INSTALL_TESTING_APP: true
 
-    #- PHP_VERSION: 7.2
```
phil-davis: These commented-out matrix entries are not in `stable10` at all. Si this is a chance to remove them.
They have a bit out-of-date format anyway, and it is easy to add them anyway, without needing a "template" like this.

.drone.yml
```diff
@@ -759,47 +748,60 @@ matrix:
       INSTALL_SERVER: true
       INSTALL_TESTING_APP: true
 
-  # test on objectstore
```
phil-davis: The diff looks confusing - actually "test on objectstore" matrix entry gets deleted, and a new matrix entry is added down at the end of the diff.

individual-it (approved):

PR #34798, opened 2019-03-18
Fix Webdav error page, include CSP and message

PVince81:
## Description
Add CSP to Webdav error page.
Include exception message.

## Related Issue
- Discovered with https://github.com/owncloud/core/pull/33992

## How Has This Been Tested?

1. Enable firewall app
1. Upload a file "test.txt"
1. Apply a new system tag "forbidden"
1. Create a firewall rule to block based on this tag
1. Try downloading the file

Before this fix: error page is missing styles, browser console full of errors.
After this fix: page renders properly, no more errors.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/Files/BrowserErrorPagePlugin.php
+++ b/apps/dav/lib/Files/BrowserErrorPagePlugin.php
@@ -80,19 +80,22 @@ class BrowserErrorPagePlugin extends ServerPlugin {
 		}
 		$this->server->httpResponse->addHeaders($headers);
 		$this->server->httpResponse->setStatus($httpCode);
-		$body = $this->generateBody();
+		$body = $this->generateBody($ex);
 		$this->server->httpResponse->setBody($body);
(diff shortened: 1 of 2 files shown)


mrow4a:
Do you want to also cherry-pick this here https://github.com/owncloud/core/commit/ddfe043f49cd8ec5bacbe329cc3e2f7f770986b5 ?

IljaN (approved):
Can't think of any possible attack scenario as stack-trace is the only thing which is dynamic on this page. Would feel better if we could somehow make it work without unsafe-inline ("defense in depth").

There is a risk if attacker is able to modify the DOM of the error page by using a vulnerability (probably js) in the previous page.

PVince81:
stable10: https://github.com/owncloud/core/pull/34817

PR #34821, opened 2019-03-19
[stable10] Update changelog post sprint 23

PVince81:
Many changes are related to either PHP 5.6 deprecation (especially the dependency bumps) and to aligning stable10 code with core through many backports.


phil-davis (approved):
Looks a reasonable choice of what is "real" from among the "rubbish backports" of minor code diffs.

Issue #34797, opened 2019-03-18
Lost password JS applies to wrong field


PVince81:
### Steps

1. Check out git master
1. Open browser console
1. Open login page
1. Enter user name
1. Start typing password

### Expected
No JS errors.

### Actual
![image](https://user-images.githubusercontent.com/277525/54546831-cfb94300-49a4-11e9-8c6e-9cc3c6aa1135.png)

### Version
git master f0de0d3f572d518326b2b70d4f7cc7de1c216007

Not happening in OC 10.1.0.

Need to check stable10 in case we backported this bug...

@sharidas @phil-davis

PVince81:
also happens on stable10

sharidas:
- [x] Happens in stable10
- [x] Not found in oC 10.1.0

sharidas:
Created PR: 
- [x] master branch https://github.com/owncloud/core/pull/34816

PR #34834, opened 2019-03-19
[stable10] Backport of Check the retypepassword exist in the dom befo…

sharidas:
## Description

The issue here is, the id `retypepassword` is not available during the login. Its only available when the lostpassword.js is loaded from https://github.com/owncloud/core/blob/master/core/templates/lostpassword/resetpassword.php#L36. This issue was found in the console log and not in the UI. So in the lostpassword.js before checking if the id `retypepassword` exist, the objects `.val().length` was fetched which was throwing error in the web dev console. So in this PR, first we check if the id exist, if so then check the `.val().length` equal to zero, to remove the warning class and then empty the message text. ( copied from PR https://github.com/owncloud/core/pull/34816)

## Related Issue

- Fixes https://github.com/owncloud/core/issues/34797

## Motivation and Context

No js errors should be seen in the dev console. Try to check the length of val(), once we verify if the id exist in the DOM.

## How Has This Been Tested?

- Attached screencast here, tested on stable10 branch: 
stable10_lostpasswd

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/js/lostpassword.js
+++ b/core/js/lostpassword.js
@@ -157,7 +157,9 @@ $(document).ready(function () {
 		 The warning message should be shown only during password mismatch.
 		 Else it should not.
 		 */
-		if (($('#password').val().length >= 0) && ($('#retypepassword').val().length === 0)) {
+		if (($('#password').val().length >= 0)
+			&& ($('#retypepassword').length)
+			&& ($('#retypepassword').val().length === 0)) {
 			$('#message').removeClass('warning');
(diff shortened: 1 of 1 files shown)


sharidas:
Original PR : https://github.com/owncloud/core/pull/34816

PVince81 (approved):
:+1:

PR #34847, opened 2019-03-20
decode only the first level of JSON from log-files

individual-it:
## Description
treat everything below the first level of JSON as string
if its not a string after decode, encode it back to make it a string

## Related Issue
needed for https://github.com/owncloud/admin_audit/issues/79

## Motivation and Context
when having a log entry that is some object/array itself e.g. `"auditUsers":["admin", "user"]` it's hard to reflect that in the flat behat table.
so easier just to compare against the string
## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Logging.php
+++ b/tests/acceptance/features/bootstrap/Logging.php
@@ -79,6 +79,9 @@ trait Logging {
 						"could not find attribute: '$attribute' in log entry: '{$logLines[$lineNo]}'"
 					);
 					$message = "log entry:\n{$logLines[$lineNo]}\n";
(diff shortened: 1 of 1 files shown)


phil-davis (approved):

phil-davis:
YAGNI principle here - this works fine for key-value pairs where the value is an array with a "reasonably small" number of string values. 

If we find example in future where there are arrays with various key-value that contain other arrays... then we can design some additional scheme for specifying particular nested keys down to a value.

individual-it:
backport in  #34849

PR #34850, opened 2019-03-20
group shares should respect to users auto accept preference

### LINKED ISSUE #34705: user-based autoaccepting is off but shares are auto accepted when shared with group
### Steps to reproduce
1. create user1 and put the user into group1
1. as user1 disable "Automatically accept new incoming local user shares"
1. share a folder with group1

### Expected behaviour
user1 should not automatically receive the share

### Actual behaviour
user1 received the share

### Client configuration
**Browser:**

**Operating system:**

```
{"reqId":"MNZ2NNCqWxgYDbg9wRaL","level":3,"time":"2019-03-07T10:52:20+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"PHP","method":"GET","url":"\/owncloud-core\/ocs\/v1.php\/apps\/files_sharing\/api\/v1\/sharees?format=json&search=g1&perPage=200&itemType=folder","message":"Invalid argument supplied for foreach() at \/home\/artur\/www\/owncloud-core\/apps\/files_sharing\/lib\/SharingBlacklist.php#96"}
{"reqId":"EXA9df4mDerSVuIDeiod","level":3,"time":"2019-03-07T10:52:27+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"PHP","method":"POST","url":"\/owncloud-core\/ocs\/v2.php\/apps\/files_sharing\/api\/v1\/shares?format=json","message":"Invalid argument supplied for foreach() at \/home\/artur\/www\/owncloud-core\/apps\/files_sharing\/lib\/SharingBlacklist.php#96"}
```

karakayasemi:
## Description
If auto accept enabled by admin and if it is a group share, create sub-share for auto accept disabled users in the pending state to respect users auto accept preference.

## Related Issue
Fixes #34705 

## Motivation and Context
Removing bug.
## How Has This Been Tested?
- Enable auto-accept share in the admin sharing panel
- Create `testuser` and `testgroup`.
- add `testuser` to `testgroup`
- login with test-user and disable auto-accept share from the personal panel
- share a file to `testgroup` with admin user
- the share should be created in pending status for `testuser`

## Types of changes

- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -379,12 +379,12 @@ class Share20OcsController extends OCSController {
 
 		$shareWith = $this->request->getParam('shareWith', null);
 
-		$autoAccept = false;
-		$globalAutoAcceptValue = $this->config->getAppValue('core', 'shareapi_auto_accept_share', 'yes');
-		if ($globalAutoAcceptValue === 'yes') {
-			$autoAccept = $this->config->getUserValue($shareWith, 'files_sharing', 'auto_accept_share', $globalAutoAcceptValue) === 'yes';
-		}
+		$globalAutoAccept = $this->config->getAppValue('core', 'shareapi_auto_accept_share', 'yes') === 'yes';
 		if ($shareType === Share::SHARE_TYPE_USER) {
+			$userAutoAccept = false;
+			if ($globalAutoAccept) {
+				$userAutoAccept = $this->config->getUserValue($shareWith, 'files_sharing', 'auto_accept_share', 'yes') === 'yes';
+			}
 			// Valid user is required to share
 			if ($shareWith === null || !$this->userManager->userExists($shareWith)) {
 				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
@@ -392,7 +392,7 @@ class Share20OcsController extends OCSController {
 			}
 			$share->setSharedWith($shareWith);
 			$share->setPermissions($permissions);
-			if ($autoAccept) {
+			if ($userAutoAccept) {
 				$share->setState(Share::STATE_ACCEPTED);
 			} else {
 				$share->setState(Share::STATE_PENDING);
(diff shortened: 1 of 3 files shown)


jvillafanez:
Code looks fine. I have some questions:

If the admin has set autoaccept and the group has one member with explicit autoaccept disabled, the user will have his share in a pending state. However, what happens the other way around? What if the admin doesn't have set the autoaccept, but the user has it enabled? I expect the user to autoaccept the share (or maybe this situation isn't possible, I don't remember)

For the code, I think we should start considering splitting things. The `createShare` method has more than 200 lines of code. Just having a quick look:
* Recommended maximum number of lines of code is 50. Even if we consider just logic lines or sentencies, I'm pretty sure it will be above 100 lines. This method is becoming very big.
* Recommended maximum complexity for a method is 10. I might be counting wrong but the complexity is around 40-50. The method is becoming more complex.

As long as the logic works fine, for me it's ok to merge this PR (double-check the above case first) and figure out a refactor plan at some point.

karakayasemi:
> What if the admin doesn't have set the autoaccept, but the user has it enabled?

@jvillafanez like you said if it is disabled by admin, user can not even see this option. User only can configure when it is enabled. We have also acceptance tests for this case.

>For the code, I think we should start considering splitting things. The `createShare` method has more than 200 lines of code. Just having a quick look:

Actually, most of the sharing code base has complex and some parts of it are including legacy things. We should definitely consider to refactor it. However, this PR is only a bug fix. I prefer to make it out of this PR's scope.

jvillafanez:
> @jvillafanez like you said if it is disabled by admin, user can not even see this option. User only can configure when it is enabled. We have also acceptance tests for this case.

So if it's disabled, we ignore whatever option the user could have been set (?).
I mean, the admin might have enabled the autoaccept, and the user might explicitly set the autoaccept (the preference is written for that user). Then the admin decided to disable the autoaccept. So technically, this scenario is still possible.

If that's what we want and there are test covering that scenario, then it's ok. I think we should somehow document this in order to make the behaviour as clear as possible, although I'm not sure where.

karakayasemi:
I guess you are describing the following scenario:
https://github.com/owncloud/core/blob/master/tests/acceptance/features/webUISharingAcceptShares/acceptShares.feature#L281
When admin disables auto-accept, user's auto-accept preference is not valid. A user can only disable auto-accept share for himself when it is enabled by admin. Also, he/she can re-activate it.

jvillafanez (approved):

PVince81:
@karakayasemi please backport

PR #34854, opened 2019-03-20
Don't show reset link on failed LDAP login

micbar:
## Description
Do not create a reset link for failed LDAP Logins

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3220

## Motivation and Context
Some Instances have DB Users and LDAP Users in parallel. They want a working reset link for DB Users and don't show a reset link for LDAP users because they cannot be changed in OC. The current behaviour is misleading and could be considered as a bug

## How Has This Been Tested?
Manually

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -141,6 +141,13 @@ class LoginController extends Controller {
 		if (!$parameters['resetPasswordLink']) {
 			if ($user !== null && $user !== '') {
 				$userObj = $this->userManager->get($user);
+				if ($userObj === null) {
+					$users = $this->userManager->find($user);
+					if (\count($users) === 1) {
+						$userId = \key($users);
+						$userObj = $this->userManager->get($userId);
+					}
+				}
 				if ($userObj instanceof IUser) {
 					$parameters['canResetPassword'] = $userObj->canChangePassword();
 				}


jvillafanez:
Check if setting the "canResetPassword" as false by default in https://github.com/owncloud/core/pull/34854/files#diff-5ab999f57c47dddcd9de7539846aeee3L139 works.

micbar:
@PVince81 
Could you provide feedback on this approach?
I could finish this too before code freeze

PVince81 (changes requested):
:-1: no, this would also match partial user names. needs a different approach

micbar:
@PVince81 ok

We would need an mechanism to find users by
- Displayname
- UserID
- E-Mail
- all allowed LDAP Attributes from the Login Filter eg. CN, samaccountname, Distinguished name and so on

PVince81:
@micbar the part I don't understand is that the PR title says it's removing something, the reset link, but the code seems to want to match more, which doesn't seem to fit with "removing".

as far as I understand, the match by email was there mostly for the local user backend, so we could also add a backend class name check maybe

PVince81:
in general we advise people with LDAP setups to disable password reset by link completely by setting this value to "disabled": https://github.com/owncloud/core/blob/stable10/config/config.sample.php#L277

if that's not acceptable, this means that the use case in question is a setup with LDAP users mixed with local ones, maybe guests ?

micbar:
> @micbar the part I don't understand is that the PR title says it's removing something, the reset link, but the code seems to want to match more, which doesn't seem to fit with "removing".

@PVince81 The reset link does not show up, when the user backend returns false on the canChangePassword() method. If the correct UserBackend does not get determined, it will create the link anyway. This is a bug IMO

micbar:
> if that's not acceptable, this means that the use case in question is a setup with LDAP users mixed with local ones, maybe guests ?

The have a mixed seutp and want to keep the reset link for the DB users.

PVince81:
wouldn't that be a security issue ?
this would let an attacker know what users exist in the system by finding the ones where the reset link isn't there.
in general I thought we'd let the user request a link anyway but in some cases we just don't send it out.

and maybe there for LDAP users, if they do click the link, the page should tell them that they need to contact admin for changing password or something. I wonder if that wouldn't be the better flow

micbar:
Originating Issue was closed.

PR #34857, opened 2019-03-21
[do not merge] Demo incorrect phpunit status (PHPUnit6)

phil-davis:
## Description
See issue for the detailed description of this problem.

## Related Issue
Issue #34858 

## Motivation and Context
CI should reliably report red-fail when a test fails.

## How Has This Been Tested?
CI


phil-davis:
https://phpunit.de/supported-versions.html

We are currently using PHPunit 5.7.27 - official support finished 2018-02-02 and it ony claims to support PHP 5.6, PHP 7.0, PHP 7.1 - so there are no guarantees about what happens with PHP 7.2 and 7.3

PHPunit 6 - official support ended 2019-02-01 - it supports PHP PHP 7.0, PHP 7.1, PHP 7.2 - so that is what we could use

PHPunit 7 is for PHP 7.1, PHP 7.2, PHP 7.3 and so we can only use that when we drop PHP 7.0

phil-davis:
Keeping this here for further investigation this week, to try and understand exactly what unit test step is causing grief for the PHPUnit exit status.

PR #34852, opened 2019-03-20
Set proper value when upgrading app patch version

PVince81:
## Related Issue
Fixes https://github.com/owncloud/core/issues/34851

## How Has This Been Tested?
Manual test with steps
- [ ] TODO: add unit test if possible

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34859, opened 2019-03-21
[stable10] Set proper value when upgrading app patch version

PVince81:
Backport of https://github.com/owncloud/core/pull/34852 to stable10


IljaN (approved):
LGTM

PR #34871, opened 2019-03-22
[master] Add explicit PHPUnit asserts to unit tests flagged as 'risky' by PHPUnit6

phil-davis:
## Description
PHPUnit6 reports as "risky" tests that ran the test case but no `assert` was ever done.

Those are suspicious, because they might never fail.

Often the tests are checking that "something works and does not throw an exception". In that case the test will fail if an exception is thrown (for whatever reason - bug...) In those cases, put an assert of some sort about what the method call in the test is expected to return. If it is a method call that returns something, e.g. returns an int, then check that the return value is an int. If it is a method that does not return anything ("return void") then use `assertNull` to check that the return value "is not".

And fix other tests that turned out to be "snake oil" on examination (see inline comments).

It seems that the reporting of these sort of tests as "risky" is worth having.

## Related Issue
Part of the rabbit-hole being followed from issue #34858 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/federatedfilesharing/lib/Middleware/OcmMiddleware.php
+++ b/apps/federatedfilesharing/lib/Middleware/OcmMiddleware.php
@@ -93,7 +93,7 @@ class OcmMiddleware {
 	 *
 	 * @param string[] $params
 	 *
-	 * @return bool
+	 * @return void
 	 *
 	 * @throws BadRequestException
 	 */
--- a/apps/files/tests/Command/ScanTest.php
+++ b/apps/files/tests/Command/ScanTest.php
@@ -239,12 +239,20 @@ class ScanTest extends TestCase {
 		if (\count($groups) === 2) {
 			$this->assertContains('Starting scan for user 1 out of 10 (user1)', $output);
 			$this->assertContains('Starting scan for user 1 out of 10 (user11)', $output);
-		}
-		if (\count($groups) === 3) {
+		} elseif (\count($groups) === 3) {
 			$this->assertContains('Starting scan for user 1 out of 10 (user1)', $output);
 			$this->assertContains('Starting scan for user 1 out of 10 (user11)', $output);
 			$this->assertContains('Starting scan for user 1 out of 10 (user21)', $output);
 			$this->assertContains('Starting scan for user 10 out of 10 (user30)', $output);
+		} elseif (\count($groups) === 4) {
+			$this->assertContains('Starting scan for user 1 out of 10 (user1)', $output);
+			$this->assertContains('Starting scan for user 1 out of 20 (user11)', $output);
+			$this->assertContains('Starting scan for user 11 out of 20 (user21)', $output);
+			$this->assertContains('Starting scan for user 10 out of 10 (user40)', $output);
+		} else {
+			$this->fail(
+				"testMultipleGroups supports testing with 2,3, or 4 groups but the input has $groups groups"
+			);
 		}
 	}
 
--- a/tests/lib/Events/EventEmitterTraitTest.php
+++ b/tests/lib/Events/EventEmitterTraitTest.php
@@ -97,6 +97,10 @@ class EventEmitterTraitTest extends TestCase {
 			$this->assertInstanceOf(GenericEvent::class, $calledAfterEvent[1]);
 			$this->assertArrayHasKey('item', $calledAfterEvent[1]);
 			$this->assertEquals('testing', $calledAfterEvent[1]->getArgument('item'));
+		} else {
+			$this->assertEquals($calledAfterEvent[0], "$className.after$eventName");
+			$this->assertArrayHasKey('item', $calledAfterEvent[1]);
+			$this->assertEquals('testing', $calledAfterEvent[1]->getArgument('item'));
 		}
 	}
 }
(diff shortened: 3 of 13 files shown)


apps/federatedfilesharing/lib/Middleware/OcmMiddleware.php
```diff
@@ -93,7 +93,7 @@ public function __construct(
 	 *
 	 * @param string[] $params
 	 *
-	 * @return bool
+	 * @return void
```
phil-davis: While adjusting `OcmMiddelwareTest` I saw that the PHPdoc return specified here was wrong.
This method either returns nothing (`void`) ir throws an exception.

apps/files/tests/Command/ScanTest.php
```diff
@@ -239,12 +239,20 @@ public function testMultipleGroups($input) {
...
 			$this->assertContains('Starting scan for user 1 out of 10 (user1)', $output);
 			$this->assertContains('Starting scan for user 1 out of 10 (user11)', $output);
-		}
-		if (\count($groups) === 3) {
+		} else if (\count($groups) === 3) {
 			$this->assertContains('Starting scan for user 1 out of 10 (user1)', $output);
 			$this->assertContains('Starting scan for user 1 out of 10 (user11)', $output);
 			$this->assertContains('Starting scan for user 1 out of 10 (user21)', $output);
 			$this->assertContains('Starting scan for user 10 out of 10 (user30)', $output);
+		} else if (\count($groups) === 4) {
```
phil-davis: This test was doing no `assertContains` checks for the data set:
```
[['--groups' => 'group1,group2', '--group' => ['group2','group3']], '']
```
and so it did not fail, but also was snake oil.

apps/files/tests/Command/ScanTest.php
```diff
@@ -239,12 +239,20 @@ public function testMultipleGroups($input) {
...
+			$this->assertContains('Starting scan for user 1 out of 10 (user1)', $output);
+			$this->assertContains('Starting scan for user 1 out of 20 (user11)', $output);
+			$this->assertContains('Starting scan for user 11 out of 20 (user21)', $output);
+			$this->assertContains('Starting scan for user 10 out of 10 (user40)', $output);
+		} else {
+			$this->fail(
```
phil-davis: Make sure that if someone adds test cases with different numbers of groups in the command, that the test will fail until they add appropriate asserts.

tests/lib/Events/EventEmitterTraitTest.php
```diff
@@ -97,6 +97,10 @@ public function testEmittingCallWithAdditionalArgument($className, $eventName, $
 			$this->assertInstanceOf(GenericEvent::class, $calledAfterEvent[1]);
 			$this->assertArrayHasKey('item', $calledAfterEvent[1]);
 			$this->assertEquals('testing', $calledAfterEvent[1]->getArgument('item'));
+		} else {
```
phil-davis: Previously no checks were happening for the test cases that had an empty `after` specified. So those test cases were "snake oil". Actually it turns out that the tests are OK - checking with these asserts passes.

tests/lib/Group/Backend.php
```diff
@@ -155,7 +155,13 @@ public function testSearchUsers() {
 	public function testAddDouble() {
 		$group = $this->getGroupName();
 
-		$this->backend->createGroup($group);
-		$this->backend->createGroup($group);
+		$this->assertTrue(
+			$this->backend->createGroup($group),
+			"there was a problem creating $group"
+		);
+		$this->assertFalse(
```
phil-davis: The first time calling `createGroup()` returns true (it did it)
The second time returns false - the group already exists.
We can check this stuff, then we will know if that behavior changes.

phil-davis:
Backport `stable10` #34876

PVince81:
@phil-davis can we configure this in PHPUnit so it runs like we want to instead of skipping tests ?

phil-davis:
> @phil-davis can we configure this in PHPUnit so it runs like we want to instead of skipping tests ?

@PVince81 I don't understand what you mean? All the tests are actually run by `phpunit`, it is just that some of them contain no `assertSomething()` calls, and so `phpunit` (after running them) reports that they are "risky" because they "cannot fail" in a typical way.

For safety, I put an `assertSomething()` call in each offending test. That makes sure that the test always asserts something, and thus could fail if the assertion is not true at run-time.

Of course all tests can fail if something that they call throws an exception, but `phpunit` does not consider only having that to be an acceptable thing.

PVince81:
@phil-davis I was a bit out of context.

I'm surprised that PHPUnit 5 doesn't report such tests as risky already ? Or did we have an option in phpunit.xml to disable this ?

PVince81 (approved):
:+1: fine by me

phil-davis:
> @phil-davis I was a bit out of context.
> 
> I'm surprised that PHPUnit 5 doesn't report such tests as risky already ? Or did we have an option in phpunit.xml to disable this ?

PHPUnit6 seems to report the risky stuff. The default for `beStrictAboutTestsThatDoNotTestAnything` was false in PHPUnit5, now the default is true. So we suddenly start seeing these reported in PHPUnit6.

PR #34872, opened 2019-03-22
Make sure testvalidateExpirationDateEnforceToFarIntoFuture unit test will fail if no exception is thrown

phil-davis:
## Description
Add a `fail` in the test path that should not happen. The test is expected to throw an exception, and the `catch` block get executed. The `fail` at the end of the `try` block ensures that if the expected exception is not thrown, then the test will fail.

Note: the test does go down the correct `catch` block at present, so the test itself does not need changing. I thought that some time in the past the exception was not being thrown, but maybe I was wrong.

## Related Issue
#34874 

## Motivation and Context
Remove the possibility of snake-oil tests.

## How Has This Been Tested?
- local unit test run and CI
- change the expected status from `404` to `403` and see that the test fails, thus confirming that the code in the `catch` block is really being executed.
- adjust the test case data so that no exception is thrown. See that the new `fail()` message happens and the test fails.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/lib/Share20/ManagerTest.php
+++ b/tests/lib/Share20/ManagerTest.php
@@ -975,6 +975,7 @@ class ManagerTest extends \Test\TestCase {
 
 		try {
 			$this->invokePrivate($this->manager, 'validateExpirationDate', [$share]);
+			$this->fail("expected GenericShareException but no exception was thrown");
 		} catch (\OCP\Share\Exceptions\GenericShareException $e) {
 			$this->assertEquals('Cannot set expiration date more than 3 days in the future', $e->getMessage());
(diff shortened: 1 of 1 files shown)


micbar:
@phil-davis Do you need some help on this?

phil-davis:
@micbar I will look tomorrow. 
I got side-tracked last week with "random" problems in webUI testing due to chrome 75 release.

phil-davis:
Backport `stable10` #35611

individual-it (approved):

PR #34878, opened 2019-03-22
[10.1.1] Set proper value when upgrading app patch version

PVince81:
Adjusted fix from https://github.com/owncloud/core/pull/34859 for the code on top of v10.1.0

:warning: this is not the exact same fix as the variables were different on v10.1.0 vs stable10.


phil-davis (approved):
Good to notice it now!

PR #34897, opened 2019-03-26
[10.1.1] CI use compatible app versions

PVince81:
This change would only be for the release branch...


PVince81:
some strange failure related to an extra newline: https://drone.owncloud.com/owncloud/core/16478/544

@individual-it

PR #34898, opened 2019-03-26
Add the apps-external directory during setup only

### LINKED ISSUE #29839: [Feature Request] make the apps-external folder default used in new installations

mmattel:
## Description
- This PR adds the `apps-external` directory only during installation but not as before as part of the tar.
- Relates to changes made in master #30889 and stable10 #34656, both merged
Removing the changes made in Makefile,
Removing the physically created apps-external directory
Adopting the code in Setup.php

## Related Issue

- Fixes https://github.com/owncloud/core/issues/29839#issuecomment-476576489
([Feature Request] make the apps-external folder default used in new installations)

## Motivation and Context

Prevents re-adding that directory on every update.

## How Has This Been Tested?

- make test-php-unit TEST_DATABASE=mysql TEST_PHP_SUITE=tests/lib/SetupTest.php --> ok

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -285,7 +285,22 @@ class Setup {
 		) {
 			$error[] = $l->t("Can't create or write into the data directory %s", [$dataDir]);
 		}
-		
+
+		// create the apps-external directory only during installation
+		$appsExternalDir = \OC::$SERVERROOT.'/apps-external';
+		if (!\file_exists($appsExternalDir)) {
+			@\mkdir($appsExternalDir);
+		}
+
+		// validate the apps-external directory
+		if (
+			(!\is_dir($appsExternalDir) and !\mkdir($appsExternalDir)) or
+			!\is_writable($appsExternalDir)
+		) {
+			$htmlAppsExternalDir = \htmlspecialchars_decode($appsExternalDir);
+			$error[] = $l->t("Can't create or write into the apps-external directory %s", $htmlAppsExternalDir);
+		}
+
 		if (\count($error) != 0) {
 			return $error;
 		}
@@ -383,10 +398,10 @@ class Setup {
 				$config->setSystemValue('logtimezone', \date_default_timezone_get());
 			}
 
-			// adding the apps-external directory by default using apps_path
+			// add the apps-external directory to config using apps_path
+			// add the key only if it does not exist (protect against overwriting)
 			$apps2Key = \OC::$server->getSystemConfig()->getValue('apps_paths', false);
 
-			// add the key only if it does not exist (protect against overwriting)
 			if ($apps2Key === false) {
 				$defaultAppsPaths = [
 					'apps_paths' => [
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ core_vendor=core/vendor
 
 core_doc_files=AUTHORS COPYING README.md CHANGELOG.md
 core_src_files=$(wildcard *.php) index.html db_structure.xml .htaccess .user.ini robots.txt
-core_src_dirs=apps apps-external core l10n lib occ ocs ocs-provider ocm-provider resources settings
+core_src_dirs=apps core l10n lib occ ocs ocs-provider ocm-provider resources settings
 core_test_dirs=tests
 core_all_src=$(core_src_files) $(core_src_dirs) $(core_doc_files)
 core_config_files=config/config.sample.php config/config.apps.sample.php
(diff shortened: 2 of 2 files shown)


PVince81:
- [x] TEST: no op when folder exists and no write permissions
- [ ] TEST: folder creates when not existing :no_entry_sign: FAIL
- [ ] TEST: error when no permission when attempting creation
- [ ] TEST: created folder has write permissions

PVince81:
tested with occ maintenance:install ^

lib/private/Setup.php
```diff
@@ -285,7 +285,22 @@ public function install($options) {
 		) {
 			$error[] = $l->t("Can't create or write into the data directory %s", [$dataDir]);
 		}
-		
+
+		// create the apps-external directory only during installation
+		$appsExternalDir = $this->config->getSystemValue('datadirectory', \OC::$SERVERROOT.'/apps-external');
```
PVince81: key value is "datadirectory" here ??

lib/private/Setup.php
```diff
@@ -285,7 +285,22 @@ public function install($options) {
...
+		if (!\file_exists($appsExternalDir)) {
+			@\mkdir($dataDir);
```
PVince81: and here you create the data dir instead of the $appsExternalDir, seems all is a bit mixed up?

mmattel:
```
sudo -uwww-data ./occ maintenance:install --database "mysql" --database-name "apptest" \
--database-host "localhost" --database-user "xyz" --database-pass "abc" \
--admin-user "admin" --admin-pass "def" 
ownCloud was successfully installed
```

PVince81:
forgot to say that in my setup the setup goes through but the folder "apps-external" doesn't get created, because the `file_exists` check is done on the data folder, not on "apps-external" and returns true, so it doesn't attempt to create it.

mmattel:
I fixed the issue, tested it, squashed and pushed, see my test from above.

PVince81:
ah ok, didn't see the update. thanks

lib/private/Setup.php
```diff
@@ -285,7 +285,22 @@ public function install($options) {
 		) {
 			$error[] = $l->t("Can't create or write into the data directory %s", [$dataDir]);
 		}
-		
+
+		// create the apps-external directory only during installation
+		$appsExternalDir = \OC::$SERVERROOT.'/apps-external';
+		if (!\file_exists($appsExternalDir)) {
```
PVince81: seems we already do `is_dir` for existence check and `mkdir` in the block below, so this one here feel redundant ?

mmattel: I just did the same as we do for the data directory to stay inline.
I can of course drop the if statement and just do @mkdir. Just le me know.

PVince81: ah ok, fine then. it only happens once at setup time, so not a big issue

PVince81 (approved):
:+1:

PVince81:
@mmattel please backport

phil-davis:
Backport `stable10` #34902

Issue #25470, opened 2016-07-13
Email invitation for events should come from the user's email address, not server's [$20]


fracture-point:
### Steps to reproduce

Using OC 9.0.3 + Calendar 1.2.2, invite someone to calendar event.  
The outbound email "From" address shows as the server from, configured in 
Admin-->Email Server-->From Address.

The email should come from the user's configured email address, not the server.  This is especially confusing when the OC instance is host for a number of unrelated domains.
### Expected behaviour

Tell us what should happen
### Actual behaviour

Tell us what happens instead
### Server configuration

Operating system:
Ubuntu 14.04
Web server:
Apache 2.4.7
Database:
MariaDB 10.0.26
PHP version:
php 5.5.9
ownCloud version: (see ownCloud admin page)
9.0.3
Contacts version: (see ownCloud apps page)
1.3.1.0 (calendar = 1.2.2)
Updated from an older ownCloud or fresh install:
Updated from 8.2.2
Signing status (ownCloud 9.0 and above):

No errors have been found.

List of activated apps:
Enabled:

activity: 2.2.1
calendar: 1.2.2
comments: 0.2
contacts: 1.3.1.0
dav: 0.1.6
federatedfilesharing: 0.1.0
federation: 0.0.4
files: 1.4.4
files_pdfviewer: 0.8.1
files_sharing: 0.9.1
files_texteditor: 2.1
files_trashbin: 0.8.0
files_versions: 1.2.0
files_videoplayer: 0.9.8
gallery: 14.5.0
notifications: 0.2.3
provisioning_api: 0.4.1
systemtags: 0.2
tasks: 0.8.1
updatenotification: 0.1.0 Disabled:
bookmarks
encryption
external
files_antivirus
files_external
firstrunwizard
news
notes
templateeditor
user_external
user_ldap

**The content of config/config.php:**

**Are you using external storage, if yes which one:** local/smb/sftp/...
No
**Are you using encryption:** yes/no
No
**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...
No
### Client configuration

**Browser:**
Chrome / Safari
**Operating system:**
OS X El Cap
**CalDAV-clients:**
#### Web server error log

N/A
#### ownCloud log (data/owncloud.log)

N/A
#### Browser log

N/A

DeepDiver1975:
@georgehrke interested in having a look? THX

CliffMitchell:
[...]

PR #34899, opened 2019-03-26
Use creator of invitation as From field in invitation mails

valentinstn:
## Motivation and Context

Invitations to events are send by people and not the technical email address. Invitees expect to get an invitation from another person, not a service.

## How Has This Been Tested?

Manually tested in our environment and by other users here: https://github.com/owncloud/core/issues/25470#issuecomment-460729618

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
+++ b/apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
@@ -122,6 +122,7 @@ class IMipPlugin extends SabreIMipPlugin {
 		$message = $this->mailer->createMessage();
 
 		$message->setReplyTo([$sender => $senderName])
+			->setFrom([$sender => $senderName])
 			->setTo([$recipient => $recipientName])
 			->setSubject($subject)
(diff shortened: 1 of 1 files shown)


DeepDiver1975 (approved):

PVince81:
@m55c55 can you backport this to the stable10 branch ? just send a PR with the same change to stable10

m55c55:
@PVince81 Done

phil-davis:
Backport `stable10` is PR #34901

magicWF (commented):
This correction causes problems when sending mails when the OC-Server tries to send mails with foreign senders (domains for which the OC MTA is not responsible). In the worst case, my server should send a mail to Apple or Google that actually comes from them. In order to reduce the amount of SPAM, the easiest method is not to accept mail from your own domain if it comes from other servers. This is done using techniques like DMARC (RFC5322), SPF, DKIM.

Therefore, this patch should not be published in this way!

PR #34906, opened 2019-03-27
reduce query size by properly reusing the query builder

butonic:
:disappointed: ... :crossed_fingers: this works

for reviewing ignore whitespace changes, that makes a lot clearer what I did.

### DIFF
--- a/apps/federatedfilesharing/lib/FederatedShareProvider.php
+++ b/apps/federatedfilesharing/lib/FederatedShareProvider.php
@@ -577,44 +577,45 @@ class FederatedShareProvider implements IShareProvider {
 	public function getAllSharesBy($userId, $shareTypes, $nodeIDs, $reshares) {
 		$shares = [];
 
-		$nodeIdsChunks = \array_chunk($nodeIDs, 100);
-		foreach ($nodeIdsChunks as $nodeIdsChunk) {
-			$qb = $this->dbConnection->getQueryBuilder();
-			$qb->select('*')
-				->from($this->shareTable);
+		$qb = $this->dbConnection->getQueryBuilder();
+		$qb->select('*')
+			->from($this->shareTable);
 
-			// In federated sharing currently we have only one share_type_remote
-			$qb->andWhere($qb->expr()->eq('share_type', $qb->createNamedParameter(self::SHARE_TYPE_REMOTE)));
+		// In federated sharing currently we have only one share_type_remote
+		$qb->andWhere($qb->expr()->eq('share_type', $qb->createNamedParameter(self::SHARE_TYPE_REMOTE)));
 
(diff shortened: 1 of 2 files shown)


butonic:
tests seem to be running into unrelated problems

butonic:
failing tests unrelated.

phil-davis:
I restarted the drone build, so you will (hopefully) get a clean run of green CI today.

jvillafanez:
Changes make sense 👍 
As far as I know, the limit for the "in" clause is 1000 due to Oracle: not sure if we want to rise the chunk size a bit more.

butonic:
Other db platforms have different limits for the IN clause. 900 should be safe, even if someone accidentally adds other elements to the in clause ... on the other hand ... that should explode as early as possible. Again, define a constant for every platform with the max allowed elements in the in clause, use it instead of hard coded values. In a different PR.

PVince81:
the limit is not only on the "IN" clause as far as I reember. at least with sqlite it was the total count of parameters, including the ones specified in "IN", so depending how many params were used in the WHERE clause the IN limit goes down.

VicDeo:
yes, this limit is applied to the total amount of placeholders in the query

butonic:
@PVince81 that would be a different limit. It depents on many factors:
- does doctrine build the query (then no parameters are used) or does the driver natively send prepared statements (oracle does as you can see in the query plans: the parameters are still placeholders)
- is there a limit on the number of parameters per query: https://stackoverflow.com/a/6582902 and the [jooq docs](https://www.jooq.org/doc/3.12/manual/sql-building/dsl-context/custom-settings/settings-inline-threshold/): 
  - Oracle : 32767
  - PostgreSQL : 32767
  - SQLite : 999

butonic:
I think we have a :+1: from @jvillafanez and tests failing for unrelated reasons. can we merge this or are we waiting for a response from @voroyam?

voroyam:
A response to what?

PVince81:
I've restarted the build

PVince81:
@butonic you can already raise the backport PR

phil-davis:
I rebased this. I don't know why it keeps failing CI today, when other PRs are getting through.
I made a cherry-pick of this to my own test PR #34916 and it looks like that is passing.
Maybe a rebase will do magic here?

mrow4a (approved):
wow.. good find!

mrow4a:
The fix is much correct, as this was clearly overlooked that query could become massive, but I am actually quite curious what happens in Oracle that just parsing takes forever.

butonic:
@mrow4a AFAICT the hard parsing step in oracle also check access permissions on columns. See https://blog.pythian.com/cursor-pin-s-wait-on-x-in-the-top-5-wait-events/

phil-davis:
Backport `stable10` #34915

PR #34916, opened 2019-03-28
reduce query size by properly reusing the query builder

phil-davis:
See if CI passes for this commit.


phil-davis:
passes - so why is there so much hassle with CI for the original PR?

PR #32062, opened 2018-07-16
Tech dept/remove deprecated owncloud 8.0.0 calls

DeepDiver1975:
## Description
remove deprecated owncloud 8.0.0 calls

## How Has This Been Tested?
- unit tests

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #34927, opened 2019-03-29
[stable10] remove deprecated owncloud 8.0.0 calls

phil-davis:
Backport #32062

Only `ownbrander` still has these `OCP\Config` uses. See comment https://github.com/owncloud/core/pull/32062#issuecomment-477939500 in master PR.

These were deprecated a long way back, so we should be OK to release the removal in ownCloud 10.* - 2 major versions later. ?


PVince81:
@jvillafanez FYI

PVince81 (approved):
:+1:

ho4ho:
Why broke apps in 10.2: https://github.com/Yetangitu/owncloud-apps/issues/119 ?

> The next minor version (which could be 10.2.0) will bring new features. _**All apps will continue working as before. API functionality is only expanded, not reduced.**_

https://owncloud.org/news/switching-owncloud-to-semantic-versioning-semver/

ltning:
This breaks Rainloop, which we have dozens of users depending on. This kind of breakage is unacceptable. Rainloop was installed from the Market app, and I expect very clear and up-front warnings if _existing_apps_ that are endorsed in the market are broken as a result of a stable/required update.

phil-davis:
The `master` branch of rainloop-webmail https://github.com/RainLoop/rainloop-webmail/blob/master/build/owncloud/rainloop-app/appinfo/info.xml has:
```
<owncloud min-version="6" max-version="9.2" />
```
So as far as I can see the app has never yet been working with any ownCloud 10.* version.
But the marketplace says "tested for ownCloud 9.1.0 - 10.0.0". I must be missing something, because `max-version` in the source repo would prevent it being installed on ownCloud 10.0.0.

Oh, when I look in the downloaded `rainloop-5.1.0.tar.gz` there is an `info.xml` with:
```
<owncloud min-version="9.1" max-version="10.0"/>
```

That must have come from somewhere else. So it will install on ownCloud 10.0.* but will not install on 10.1.* or 10.2.* anyway.

There are quite a lot of recent commits in the app's repo, but no recent releases in the marketplace https://marketplace.owncloud.com/apps/rainloop

The changelog for 10.2.0 https://github.com/owncloud/core/blob/stable10/CHANGELOG.md does mention this PR. But yes, ideally the deprecated things would have been released in 10.0.0 where people would go looking for that sort of change

What to do now?

Maybe Rainloop can make a release that supports ownCloud `10.2` ?
Or?

@DeepDiver1975 @micbar

PR #33360, opened 2018-10-30
No max version check for apps when using release channel git or daily

DeepDiver1975:
## Description
When installing owncloud from git or the daily release tar ball the max version requirement is no longer checked when installing/enabling apps

This will help with the development and release of apps in general

@patrickjahns as discussed long time back - we did go for the release channel and no env variable.

## Related Issue
- fixes #31518 

## How Has This Been Tested?
- enable an app with max version lower then 11.0
- see if being enabled

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #34931, opened 2019-03-29
[stable10] No max version check for apps when using release channel git or daily

phil-davis:
Backport #33360

Let's see if this works.

### DIFF
--- a/lib/private/App/Platform.php
+++ b/lib/private/App/Platform.php
@@ -44,29 +44,29 @@ class Platform {
 	/**
 	 * @return string
 	 */
-	public function getPhpVersion() {
+	public function getPhpVersion(): string {
 		return \phpversion();
 	}
 
 	/**
 	 * @return int
 	 */
-	public function getIntSize() {
+	public function getIntSize(): int {
 		return PHP_INT_SIZE;
 	}
 
 	/**
 	 * @return string
 	 */
-	public function getOcVersion() {
+	public function getOcVersion(): string {
 		$v = \OCP\Util::getVersion();
-		return \join('.', $v);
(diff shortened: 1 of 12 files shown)


lib/private/App/Platform.php
```diff
@@ -78,22 +78,25 @@ public function getDatabase() {
...
 	 * @return bool
 	 */
-	public function isCommandKnown($command) {
+	public function isCommandKnown($command): bool {
 		$path = \OC_Helper::findBinaryPath($command);
 		return ($path !== null);
 	}
 
-	public function getLibraryVersion($name) {
+	public function getLibraryVersion($name): ?string {
```
phil-davis: Note: this is PHP7.1

tests/lib/App/ManagerTest.php
```diff
@@ -99,22 +102,23 @@ protected function setUp() {
...
 			->willReturn($this->cache);
+		$this->platform = $this->createMock(Platform::class);
 		$this->manager = new AppManager($this->userSession, $this->appConfig,
 			$this->groupManager, $this->cacheFactory, $this->eventDispatcher,
-			$this->config);
+			$this->config, $this->platform);
 	}
 
-	protected function expectClearCache() {
+	protected function expectClearCache(): void {
```
phil-davis: Return `void` is PHP 7.1
So this PR will be best to wait until after "drop PHP 7.0"

phil-davis:
Leaving this until after PHP 7.0 is gone.
Then drop the 2nd commit, and it should be good.

phil-davis:
See #35976 for a new branch and PR against "new" `master`

PR #32404, opened 2018-08-21
[stable10] use the core exception logger functionality

patrickjahns:
## Description
by using the `writeLog` method - we avoid `logException` which can provide the exception to 3rdparty exception logger integrations (sentry)

## Motivation and Context
Collect cronjob errors in 3rdpart tools like sentry 

## How Has This Been Tested?
Manually tested while working on sentry app and figuring out how to receive errors from cron in sentry

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

Backport not applicable, code in master has changed

PR #31914, opened 2018-06-26
Have a real controller handling cron requests

DeepDiver1975:
## Description
OC_Json is deprecated - we need to start cleaning up the mess.
This is one on many steps

## Related Issue
#ajax

## Motivation and Context
Clean code is better code

## How Has This Been Tested?
- manual

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #34932, opened 2019-03-29
[stable10] Have a real controller handling cron requests

phil-davis:
Backport #31914 

The last commit removes some PHP7.1-only `: void` return declarations so that the unit tests pass here in `stable10` with PHP 7.0.

Note: since the PR in master, `cron.php` was modified by PR #32404 - that PR changed some exception logging https://github.com/owncloud/core/pull/32404/files

~~I did not see where to apply a similar "fix" to this new code with `CronController`. I think there is nothing needed. But someone please confirm that while reviewing.~~ - see comment below, this is OK.

### DIFF
--- a/core/Application.php
+++ b/core/Application.php
@@ -32,6 +32,7 @@ use OC\AppFramework\Utility\SimpleContainer;
 use OC\AppFramework\Utility\TimeFactory;
 use OC\Core\Controller\AvatarController;
 use OC\Core\Controller\CloudController;
+use OC\Core\Controller\CronController;
 use OC\Core\Controller\LoginController;
 use OC\Core\Controller\LostController;
 use OC\Core\Controller\TokenController;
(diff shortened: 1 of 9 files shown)


PVince81:
the logger change was needed only because cron.php was what I call a "naked PHP file".

now that the logic is streamlined within the app framework, exceptions should already be properly caught there, so no need to add this change

@phil-davis ready for review then ?

phil-davis:
Rebased to get a new run of CI and investigate why some unit tests are failing.

phil-davis:
@DeepDiver1975 @PVince81 I'm not sure why I left this here so long. It works fine, just needed that last commit to remove the PHP7.1+ `: void` code so that unit tests pass on PHP  7.0

Ready for review.

DeepDiver1975 (approved):

PR #34951, opened 2019-04-02
[stable10] View-only dav plugin using IShare attributes

PVince81:
Backport of https://github.com/owncloud/core/pull/33994/ to stable10

There were minor conflicts in unit test files.

PR #34965, opened 2019-04-04
Type::JSON not supported in current Doctrine

PVince81:
Doctrine/dbal 2.5 doesn't have json, so using text instead for now to avoid migration troubles.

Had to apply this on stable10 here https://github.com/owncloud/core/pull/34951


phil-davis (approved):

phil-davis:
Oh for the day we can drop PHP 7.0 and get all the modern dependencies.

PR #34992, opened 2019-04-08
Improve error handling for incoming shares poll command

### LINKED ISSUE #34981: Improve error handling for incoming shares poll command
Steps to reproduce are not clear yet but might be guessable.

When running `occ incoming-shares:poll` there are multiple issues appearing:

1) Ghost users in oc_share_external where the user id doesn't exist on the current instance any more. This throws an unhandled exception. We should skip such entries. They'll likely get cleant up by some background jobs. (if not we need to raise as it would mean we're missing a cleanup job for these)

2) Unauthorized issues, possibly when the remote isn't accessible any more. Maybe the token expired or the remote share was deleted but somehow the local instance did not get notified. Or it's one of the "add to ownCloud" shares. `Sabre\HTTP\ClientHttpException: Unauthorized`. Need to skip these issues.

3) "Unable to parse malformed url: https:///ocm-provider/". Skip invalid URLs.

4) Other potential error cases ?

Maybe we should log those though with the id of the oc_share_external entry so an admin could decide to clean these up ?

VicDeo:
## Description
Output more details when the share is skipped due to exception

## Related Issue
- Fixes #34981

## How Has This Been Tested?
1. Created a federated share,
2. Accepted it
3. Changed token on the owner side to be invalid.
4. executed `occ incoming-shares:poll` on the recipient side

```
> php occ incoming-shares:poll
Skipping external share with id "1225" from remote "http://localhost/~deo/stable10". Reason: "Sabre\HTTP\ClientHttpException: Unauthorized"
```
## Types of changes

- [x] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/federatedfilesharing/lib/Command/PollIncomingShares.php
+++ b/apps/federatedfilesharing/lib/Command/PollIncomingShares.php
@@ -90,7 +90,12 @@ class PollIncomingShares extends Command {
 					$storage = $mount->getStorage();
 					$this->refreshStorageRoot($storage);
 				} catch (\Exception $e) {
-					$output->writeln($e->getMessage());
+					$entryId = $this->getExternalShareId($data['user'], $mount->getMountPoint());
+					$remote = $storage->getRemote();
+					$reason = $e->getMessage();
+					$output->writeln(
+						"Skipping external share with id \"$entryId\" from remote \"$remote\". Reason: \"$reason\""
+					);
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


PVince81:
is the share automatically deleted through polling when it becomes invalid ? what if you run the command again, does the skipped entry still appear or not ?

apps/federatedfilesharing/lib/Command/PollIncomingShares.php
```diff
@@ -124,4 +129,25 @@ protected function getCursor() {
...
+			->where(
+				$qb->expr()->eq('user',
+					$qb->expr()->literal($userId)
+				))
+		->andWhere(
+			$qb->expr()->eq('mountpoint',
+				$qb->expr()->literal($relativeMountPoint)
+			));
+		$result = $qb->execute();
+		$externalShare = $result->fetch();
```
PVince81: didn't we have `fetchOne()` ?

PVince81 (approved):
:+1:

PVince81:
@VicDeo feel free to merge and backport after clarifying

apps/federatedfilesharing/lib/Command/PollIncomingShares.php
VicDeo: we do, in `lib/private/legacy/db/statementwrapper.php`

VicDeo:
@PVince81 >is the share automatically deleted through polling when it becomes invalid ? what if you run the command again, does the skipped entry still appear or not ?

I tried polling only for external shares with invalid token. In this case skipped entries are still appear on the next poll.

apps/federatedfilesharing/lib/Command/PollIncomingShares.php
VicDeo: But as you see  this is a legacy class and Doctrine native method `fetch` does exactly the same. (`fetch` fetches the next row while `fetchAll` fetches them all)

VicDeo:
Stable10: https://github.com/owncloud/core/pull/34993

PR #34993, opened 2019-04-09
[stable10] Improve error handling for incoming shares poll command

### LINKED ISSUE #34981: Improve error handling for incoming shares poll command

VicDeo:
Backport of https://github.com/owncloud/core/pull/34992

## Description
Output more details when the share is skipped due to exception

## Related Issue
- Fixes #34981

## How Has This Been Tested?
1. Created a federated share,
2. Accepted it
3. Changed token on the owner side to be invalid.
4. executed `occ incoming-shares:poll` on the recipient side

```
> php occ incoming-shares:poll
Skipping external share with id "1225" from remote "http://localhost/~deo/stable10". Reason: "Sabre\HTTP\ClientHttpException: Unauthorized"
```
## Types of changes

- [x] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/lib/Command/PollIncomingShares.php
+++ b/apps/federatedfilesharing/lib/Command/PollIncomingShares.php
@@ -90,7 +90,12 @@ class PollIncomingShares extends Command {
 					$storage = $mount->getStorage();
 					$this->refreshStorageRoot($storage);
 				} catch (\Exception $e) {
-					$output->writeln($e->getMessage());
+					$entryId = $this->getExternalShareId($data['user'], $mount->getMountPoint());
+					$remote = $storage->getRemote();
+					$reason = $e->getMessage();
+					$output->writeln(
+						"Skipping external share with id \"$entryId\" from remote \"$remote\". Reason: \"$reason\""
+					);
 				}
 			}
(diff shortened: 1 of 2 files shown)


PVince81 (approved):
:+1:

PR #34999, opened 2019-04-09
Add Makefile for syncing translations with Transifex

### LINKED ISSUE #164: Migrate old translation-sync-core Jenkins job
Since the old Jenkins instance is dead we need to migrate the translation sync to Drone or our internal Jenkins instance.

All relevant Job descriptions are stored on Maggie, if you want to work on this task just contact @tboerger to get access to the files.

I suggest creating make targets within the related repositories that work similar to https://github.com/owncloud/user_management/pull/68/files.

dschmidt:
## Description

This adds a Makefile to sync translations from transifex.

## Related Issue

- Fixes https://github.com/owncloud/administration/issues/164

## Motivation and Context
It's a port of the former translation-sync-core job of jenkins.owncloud.org which was shut down. This is going to be used by https://github.com/owncloud/translation-sync. C.f. https://github.com/owncloud/translation-sync/pull/17
 
I put it in a separate Makefile so I don't need to have composer available and don't have to mess too much with the main Makefile wiith the chance of breaking something.

## How Has This Been Tested?

- I ran the translation-sync job in drone and it succeeded. 
- I will run the job again, so an actual resulting commit will be added to this PR

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

- [ ] If you agree with this change, I should probably move the cpan call to here: https://github.com/owncloud-ci/transifex/blob/e0cf277/Dockerfile#L15


l10n/Makefile
```diff
@@ -0,0 +1,25 @@
...
+	rm -rf l10n
+
+.PHONY: l10n-read
+l10n-read:
+	cpan Locale::PO
+	perl l10n.pl read
+
+.PHONY: l10n-write
+l10n-write:
+	cpan Locale::PO
```
DeepDiver1975: why that? never seen before ....

dschmidt: It is used by l10n.pl: https://github.com/owncloud/core/blob/master/l10n/l10n.pl#L3

As I said I would move this to the transifex image instead if you're happy with the change otherwise.

DeepDiver1975: which image are you using? there is already a transifex image and all necessary tools/packages should already be in place ....

dschmidt: Yes, exactly. I'm talking about that image, it's just missing that perl module. 
So basically I would move this cpan invocation to here: https://github.com/owncloud-ci/transifex/blob/master/Dockerfile#L16

dschmidt: Hehe, maybe I was testing with a different package before. This one seems to actually have the module installed. So I'm just removing the cpan line and I'd call this ready to merge as soon as a build with that succeeds in drone.

dschmidt:
Good to merge from my POV.

dschmidt:
*Now* it is good to merge from my POV.

tboerger:
The translation sync itself should not be part of that PR, that should be handled by Drone.

dschmidt:
> The translation sync itself should not be part of that PR, that should be handled by Drone.

In case you are talking about the commit with translation changes: Yes, that commit was done by Drone ... it's just and example that it works, also there's no difference if we add it now or if I remove it and then drone readds it on the next run....

guruz:
@michaelstingl FYI

If I understand https://github.com/owncloud/client/issues/7063#issuecomment-483585829 correctly we need this merged here too, else we can't get translations for 2.6

dschmidt:
Well, I can remove it from the other PR too, so we can progress.. but I would obviously like to avoid that if we can also just merge this ... If anything needs to be done or we're waiting on anything, please let me know.

l10n/Makefile
tboerger: this line is still there?

tboerger:
I would personally prefer make targets within the root folder, that would enable a unification of the process accross all projects instead of requiring to change folders on so0me projects and some others not.

dschmidt:
That's messy. 

Those are separate things with separate tools needed. Instead of the atomisation of tasks this is artificially clobbering tasks together in makefiles. Yes we could switch transifex docker image to python2 but that doesn't sound like a particularly smart way forward and asks for immediate porting action of our tools at some point (when we need an update that doesn't work on python2 anymore).

In this repo this would even mean messing with completely unrelated global checks in the root Makefile. 
To me it sounds like a bad design decision, so I would ask you to do it yourself if you want it like that nevertheless.
Just be aware that we can't do it in the root Makefile for the desktop client without breaking cmake in source builds with Makefile generator.

patrickjahns:
@dschmidt 
What's the current progress here?

tboerger (approved):
Let's merge it as it can't change the current Drone or CI behavior. Let's get it rolling again!

micbar:
Seems that merging it breaks master, because of failing unit tests.
Created revert pr https://github.com/owncloud/core/pull/35323

tboerger:
Instead of reverting it we must fix the unit tests if translation update kills the unit tests. The translation sync is executed every night automatically.

micbar:
@tboerger you have to fix this unit test

https://github.com/owncloud/core/blob/master/tests/lib/L10N/FactoryTest.php#L305

This test does not expect so many languages present.
Something seems to have changed with this makefile.

I would say: 
1) revert it
2) fix the test
3) Merge it again

If you keep it and try to find someone to fix it, you are blocking many people in the meantime.

PR #35010, opened 2019-04-12
Filter unavailable users and remove unavailable shares  on polling

VicDeo:
## Description
Protect polling from exploding on unavailable users

## Motivation and Context
`An unhandled exception has been thrown:
TypeError: Argument 1 passed to OCA\Files_Sharing\External\MountProvider::getMountsForUser() must implement interface OCP\IUser, null given,` when UserManager knows nothing about a userId

## How Has This Been Tested?
1. Create and accept a federated share
2.  Execute `update oc_share_external set user="Iwouldliketobreakthepolling" limit 1;`

### Expected
Polling goes through all entries with no termination

### Actual
Polling  terminates with `TypeError: Argument 1 passed to OCA\Files_Sharing\External\MountProvider::getMountsForUser() must implement interface OCP\IUser, null given`

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/External/Manager.php
+++ b/apps/files_sharing/lib/External/Manager.php
@@ -30,6 +30,7 @@
 namespace OCA\Files_Sharing\External;
 
 use OC\Files\Filesystem;
+use OC\User\NoUserException;
 use OCP\Files;
 use OCP\Notification\IManager;
 use OCP\Share\Events\AcceptShare;
@@ -257,7 +258,7 @@ class Manager {
 	 * @return string
 	 */
 	protected function stripPath($path) {
-		$prefix = '/' . $this->uid . '/files';
+		$prefix = "/{$this->uid}/files";
 		return \rtrim(\substr($path, \strlen($prefix)), '/');
 	}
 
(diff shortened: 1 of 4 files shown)


apps/files_sharing/lib/External/Manager.php
```diff
@@ -308,7 +309,24 @@ public function setMountPoint($source, $target) {
 		return $result;
 	}
 
+	/**
+	 * @param string|null $uid
+	 */
+	public function setUid($uid) {
```
PVince81: was very confused when seeing this, then noticing that it would just update the already injected `$uid`

can you adjust the PHPDocs to clarify what this is.
is this the currently logged in user ? or do we have one external manager instance per user ?

ideal would be to use IUserSession instead. if you think it's too risky to change, add some "FIXME" to hint at this

PVince81 (changes requested):

apps/files_sharing/lib/External/Manager.php
VicDeo: @PVince81 we polling in CLI and there is not user in session.
While removing the share manager tries to to strip `/uid/files` and fails to do it correctly. See 
https://user-images.githubusercontent.com/991300/56204970-3dfb2f00-6051-11e9-8c7d-737ee65f588c.png

VicDeo: @PVince81 Added PHPDoc and fixme

PVince81 (approved):
:+1:

PVince81:
@VicDeo please backport to release-10.2.0 branch

VicDeo:
release-10.2.0: https://github.com/owncloud/core/pull/35073

phil-davis:
This was backported into `release-10.2.0`  which was then merged into `stable10` - so it is part of `stable10` release series.

PR #29404, opened 2017-10-31
Implementation of delta-sync support on server-side.

ahmedammar:
## Description
This commit adds the required server-side support for delta-sync.

The basic approach is to store zsync metadata files in a folder called `files_zsync/` which mirrors the structure and layout of the `files/` folder but appends `.zsync` to the metadata files. These files can be requested by the client via a new route `dav/files/$user/$path?zsync`. They can also be deleted using the same route. This is implemented using a new `ServerPlugin` called `ZsyncPlugin`.

Filesystem hooks are used to mirror any `move/copy/delete` operation on the base file onto the metadata file.

The upload path is implemented by modifying the `ChunkingPlugin`, the chunk files are now assumed to be named as the offsets into the original file.  Special handling is done when a chunk named `.zsync` is found, including copying the contents to the `files_zsync/` folder. This is to ensure that both the metadata and the actual file are put in place atomically, as part of the final `MOVE` request. The implemenation adds a new class `AssemblyStreamZsync` which extends `AssemblyStream` with additional support to fill in the data between chunk offsets from a `backingFile`.

A new `zsync` capability is added to the files app, which can be checked by the client to know if delta-sync is supported or not.

## Related Issue

owncloud/core#16162.

## How Has This Been Tested?

Mostly tested by using the owncloud command line and macOS client applications. With some randomly generated files using `dd` and various modifications to those files:
```
dd if=/dev/urandom of=1g.rand.img bs=$[1024*1024] count=1024

cp 1g.rand.img 1g.rand.add300m.img
dd if=/dev/urandom of=1g.rand.add300m.img bs=$[1024*1024] count=300 oseek=1024

cp 1g.rand.img 1g.rand.mod200m.img
dd if=/dev/urandom of=1g.rand.mod200m.img bs=$[1024*1024] count=100 oseek=123 conv=notrunc 
dd if=/dev/urandom of=1g.rand.mod200m.img bs=$[1024*1024] count=100 oseek=532 conv=notrunc

cp 1g.rand.img 1g.rand.mov200m.img
dd if=1g.rand.img of=1g.rand.mov200m.img bs=$[1024*1024] count=200 iseek=118 oseek=418 conv=notrunc

cp 1g.rand.img 1g.rand.del200m.img
dd if=1g.rand.img of=1g.rand.del200m.img bs=$[1024*1024] count=622
dd if=1g.rand.img of=1g.rand.del200m.img bs=$[1024*1024] iseek=822 oseek=622
```

## Types of changes

[...]

PR #35011, opened 2019-04-12
[stable10] This commit adds the required server-side support for delta-sync.

PVince81:
Backport of https://github.com/owncloud/core/pull/29404 to stable10.

Should we add a config switch to disable the feature ?

### DIFF
--- a/apps/dav/lib/Files/ZsyncPlugin.php
+++ b/apps/dav/lib/Files/ZsyncPlugin.php
@@ -1,0 +1,147 @@
+<?php
+/*
+ * Copyright (C) by Ahmed Ammar <ahmed.a.ammar@gmail.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+ * documentation files (the "Software"), to deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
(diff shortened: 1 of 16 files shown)


apps/dav/lib/Files/ZsyncPlugin.php
```diff
@@ -0,0 +1,147 @@
...
+
+	// namespace
+	const ZSYNC_PROPERTYNAME = '{http://owncloud.org/ns}zsync';
+
+	/** @var OC\Files\View */
+	private $view;
+
+	public function __construct(View $view) {
+		$this->view = $view;
+		$this->view->mkdir('files_zsync');
```
mrow4a: @PVince81 @DeepDiver1975 we have to find another way. This is invasive operation, and should be lazy.

mrow4a (changes requested):
Lets fix that mkdir issue. Lets maybe disable it by default, (informing this is experimental feature?)

guruz:
It's already behind a [x] switch in the client, so I'd say you can keep it ON in the server since clients control if they want to use it?

apps/dav/lib/Files/ZsyncPlugin.php
guruz: Note that this is a backport, not the original PR.

mrow4a:
@guruz this is exactly the problem... This `mkdir` is execute even you don't use delta sync

guruz:
Just curious about this one :-)
As 2.6.0 will ship with delta sync

PR #35004, opened 2019-04-11
[stable10] vfsstream tokenizer dependency bumps

phil-davis:
## Description
```
composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 3 updates, 0 removals
  - Updating mikey179/vfsstream (v1.6.5 => v1.6.6): Loading from cache
  - Updating theseer/tokenizer (1.1.0 => 1.1.2): Loading from cache
```
Similar to `master` PR #35003 but `myclabs/DeepCopy` needs PHP 7.1, so that bump does not yet come to `stable10`

https://github.com/bovigo/vfsStream/compare/v1.6.5...v1.6.6 - related to build changes and `composer.json` ...

https://github.com/theseer/tokenizer/releases/tag/1.1.2 - bugfix for passing an empty file

## Motivation and Context
See if these are OK to get into `10.2`

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #35012, opened 2019-04-12
[release-10.2.0] vfsstream tokenizer dependency bumps

phil-davis:
Backport #35004 into the 10.2 release

We could have these dependencies up-to-date in the 10.2 release.
(I had hoped that #35004 would have got merged before branching `release-10.2.0` but the PR did not get noticed in time)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3909,16 +3909,16 @@
         },
         {
             "name": "mikey179/vfsStream",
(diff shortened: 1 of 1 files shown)


PVince81:
please close if not critical

phil-davis:
OK. The fixes in these version bumps do not impact ownCloud.

PR #35018, opened 2019-04-15
merge config.app.sample and config.sample again

voroyam:
merge config.app.sample and config.sample again.

The split was a bad decision as it does more harm than good. Support can't waste time by searching 2 documents, it's not clear no a not professional developer which parameters are in what document.

## Description
merge config.app.sample and config.sample again.

The split was a bad decision as it does more harm than good. Support can't waste time by searching 2 documents, it's not clear no a not professional developer which parameters are in what document.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1405,4 +1405,78 @@ $CONFIG = array(
  * Async dav extensions can be enabled or disabled.
  */
 'dav.enable.async' => false,
+
+/**
+ * App: Activity
+ * 
+ * Possible values: `activity_expire_days` days
+ */
+ 
+/**
+ * Retention for activities of the activity app
+ */
+
+'activity_expire_days' => 365,
+
+/**
+ * App: LDAP
+ * 
+ * Possible values: `ldapIgnoreNamingRules` 'doSet' or false
+ * 
+ * Possible values: `user_ldap.enable_medial_search` true or false
+ */
+ 
+/**
+ * Configuring the LDAP app
+ */
+
+'ldapIgnoreNamingRules' => false,
+'user_ldap.enable_medial_search' => false,
+
+/**
+ * App: Market
+ * 
+ * Possible values: `appstoreurl` URL
+ */
+
+/**
+ * Configuring the download URL for apps
+ */
+
+'appstoreurl' => 'https://marketplace.owncloud.com',
+
+/**
+ * App: Firstrunwizard
+ * 
+ * Possible values: `customclient_desktop` URL
+ * 
+ * Possible values: `customclient_android` URL
+ * 
+ * Possible values: `customclient_ios` URL
+ */
+
+/**
+ * Configuring the download links for ownCloud clients, 
+ * as seen in the first-run wizard and on Personal pages
+ */
+ 
+'customclient_desktop' =>
+	'https://owncloud.org/install/#install-clients',
+'customclient_android' =>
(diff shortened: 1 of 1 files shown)


voroyam:
@settermjd @PVince81 @mmattel can you review this so we can merge this?

settermjd:
If it's genuinely causing a significant time drain, them I'm happy to merge it. However, there were justifiable reasons for the separation.

voroyam:
> If it's genuinely causing a significant time drain, them I'm happy to merge it. However, there were justifiable reasons for the separation.

it causes time drain.

Reasons by who? What reasons exactly?

We don't know if our users really read the docs, most of them probably don't judging from the messages at our forum. But we know that the support team reads the docs and many of our customers. That's where the focus should be on imo. If we manage docs based of the opinion of some one else, the results will be different.

I feel like I already explained myself numerous times in many tickets. PLZ just merge it.

settermjd (commented):
It's not a case of just completing this merge to merge the two files. Other work, such as updating the converter script also has to be completed. If this PR is merged as is, then it will break existing functionality.

voroyam:
So... splitting it was easy, merging it is hard?

cdamken:
IMHO I agree with @voroyam. I would like in one file too. 

The first part should be from the core and the last part would be from additional apps.

pako81:
I do not completely agree with @voroyam and @cdamken 

In my opinion, it makes sense to keep them separated because this also aligns to our official doc, where we split since some time already apps config parameters from the core ones:
 https://doc.owncloud.com/server/10.2/admin_manual/configuration/server/config_apps_sample_php_parameters.html

BUT as I can see it, the list for app keys is not comprehensive - that is, it does not include all possible app config.php parameters (and so does our doc).

My 2 cents: keep them separated, but migrate all apps-related parameters from config.sample.php to config.apps.sample.php as well as align our documentation

pako81:
This also goes in the direction of keeping core small and move as much as possible to apps, that we are trying to achieve, isn't it?

voroyam:
> In my opinion, it makes sense to keep them separated **because this also aligns to our official doc**, where we split since some time already apps config parameters from the core ones:

This exact PR's purpose is to revert that split. The document you are referring to is generated here. It was split some time ago  for better readability I suppose but to the cost of search-ability.

If you don't know exactly what the command/key is and you start searching in the page, then notice that it's in the other page, then notice that the command is spelled differently, go to the firs page again and so on = Work

Having all the commands / keys in one place = 1 search = Less Work

Also keep in mind that just because an experienced ownCloud Support Engineer like yourself, @pako81, does not need to keep it simple since you know more or less where to search what, not everybody does.
I am not that experienced so I rely heavily on the structure, and we can't assume that new members of the support / success team are very experienced as well.  

Lastly, if our search would be great - I would close this PR and forget about it. But our search is sub-optimal to say it nicely.

pako81:
Understood. So the current doc structure is a consequence of the split of the config.php file. Yes, in this case then I do agree with your comments and probably the best idea is to merge them, as it used to be before.

voroyam:
no reviews after 3 months. closing. not a priority apparently.

PR #35039, opened 2019-04-17
[stable10] Backport of Fixing encrypted version test

sharidas:
Its better to verify if the command output contains specific
strings, instead of grabbing the whole command output
and comparing with string equal.

## Description

Adjust the unit test for fix-encrypted-version command so that it should not rely on the skeleton files. It should only rely on the output that we need to look into for the files which requires fixing in the tests. In general this PR makes the unit tests good, so that it tries to verify the results that it is supposed to and not rely on what are the files provided by the skeleton.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Improve the unit tests to grab the output of the fix-encrypted-version command and verify the output of command for the files which were found to be problematic. Lets not bother about the files which does not have problem.

## How Has This Been Tested?

-  Ran unit test locally.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/encryption/tests/Command/FixEncryptedVersionTest.php
+++ b/apps/encryption/tests/Command/FixEncryptedVersionTest.php
@@ -147,14 +147,13 @@ class FixEncryptedVersionTest extends TestCase {
 		]);
 
 		$output = $this->commandTester->getDisplay();
-		$this->assertEquals("Verifying the content of file /test_enc_version_affected_user1/files/hello.txt
+
+		$this->assertContains("Verifying the content of file /test_enc_version_affected_user1/files/hello.txt
 Attempting to fix the path: /test_enc_version_affected_user1/files/hello.txt
 Increment the encrypted version to 1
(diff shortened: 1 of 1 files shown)


sharidas:
Original PR https://github.com/owncloud/encryption/pull/120

phil-davis:
@sharidas the diffs here are much shorter. Can you make a comment about that?
Then future readers (and me) will understand why the diffs are not the same as in the `encryption` app.

sharidas:
The diff is much shorter compared to https://github.com/owncloud/encryption/pull/120 ( alias original PR )
- In the original PR, the skeleton had image files, odt file and pdf file in the skeleton. This is due to the daily master tar file ( or the tar file ) used in the CI has the skeleton with the image files, odt and pdf file in the skeleton. Where as the stable10 CI doesn't have these files in the skeleton folder. So in stable10 the diff is shorter.

phil-davis (approved):

PR #35045, opened 2019-04-17
[stable10] Wrong translation file referenced for accept & decline share

jmdeboer-surfsara:
## Description
The strings for 'Accept Share' and 'Decline Share' are not translated.

These string live in apps/files_sharing/js/app.js
The translation for the strings is in apps/files_sharing/l10n/

However, in the code it says:
displayName: t('files', 'Decline Share')

So the string is searched for in the wrong file. Changed it to
displayName: t('files_sharing', 'Decline Share')

## Motivation and Context
See above

## How Has This Been Tested?
Tested on local installation

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

PR #35063, opened 2019-04-18
[10.2] Wrong translation file referenced for accept & decline share

phil-davis:
## Description
Backport into 10.2 release:
PR #35045 Wrong translation file referenced for accept & decline share
PR #35043 Get translations from correct app
PR #35049 Use consistent app name in writeLog() calls

## How Has This Been Tested?
Observing some translations in the webUI


PVince81 (approved):
:+1:

PR #35083, opened 2019-04-23
add activity for self unshare action

### LINKED ISSUE #615: Activity not reporting unshare actions done by the receiver of a share
In 10.0.7 unshare actions done by the receiver of a share are now properly logged. However, it looks like that no entry is generated in the `oc_activity` table for such actions.

Since the admin_audit app correctly logs this means the hook is working correctly. Activity might just need to adjust listening to the hook. 

@PVince81 as discussed.

karakayasemi:
## Description
This PR publishes an activity entry by listening to self unshare events. 
As far as I see, current sharing activities are published in this class by the activity app: https://github.com/owncloud/activity/blob/master/lib/FilesHooks.php. However, IMHO, holding them in files_sharing app is better. So, I added this event listener on files_sharing app. Am I missing something here?

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/2541
- Fixes https://github.com/owncloud/activity/issues/615

## Motivation and Context
whenever a recipient "unshares from self" they should see an activity in their activity stream that says they did so. This would help them to find out why something disappeared from their file tree in case they don't remember doing it. 

## How Has This Been Tested?
Unit tests and manually with the following steps:
- Activate activity app
- Create two users (`user1`, `user2`)
- Share a file from `user1` to `user2`
- Login with `user2` and unshare the file from `user2`
- `user2` should see this action in the activity stream

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/Activity.php
+++ b/apps/files_sharing/lib/Activity.php
@@ -59,6 +59,7 @@ class Activity implements IExtension {
 	const SUBJECT_RESHARED_USER_BY = 'reshared_user_by';
 	const SUBJECT_UNSHARED_USER_SELF = 'unshared_user_self';
 	const SUBJECT_UNSHARED_USER_BY = 'unshared_user_by';
+	const SUBJECT_UNSHARED_FROM_SELF = 'unshared_from_self';
 
 	const SUBJECT_SHARED_GROUP_SELF = 'shared_group_self';
 	const SUBJECT_RESHARED_GROUP_BY = 'reshared_group_by';
@@ -218,6 +219,8 @@ class Activity implements IExtension {
 				return (string) $l->t('You removed the share of %2$s for %1$s', $params);
 			case self::SUBJECT_UNSHARED_USER_BY:
 				return (string) $l->t('%2$s removed the share of %3$s for %1$s', $params);
+			case self::SUBJECT_UNSHARED_FROM_SELF:
+				return (string) $l->t('You unshared %1$s shared by %2$s from self', $params);
 
 			case self::SUBJECT_SHARED_GROUP_SELF:
 				return (string) $l->t('You shared %1$s with group %2$s', $params);
@@ -272,6 +275,8 @@ class Activity implements IExtension {
 				return (string) $l->t('Removed share for %2$s', $params);
 			case self::SUBJECT_UNSHARED_USER_BY:
 				return (string) $l->t('%2$s removed share for %3$s', $params);
+			case self::SUBJECT_UNSHARED_FROM_SELF:
+				return (string) $l->t('Unshared %1$s from self', $params);
 
 			case self::SUBJECT_SHARED_GROUP_SELF:
 				return (string) $l->t('Shared with group %2$s', $params);
@@ -366,7 +371,8 @@ class Activity implements IExtension {
 						1 => 'username',
 						2 => 'username',
 					];
-
+				case self::SUBJECT_UNSHARED_FROM_SELF:
+					return [0 => 'file', 1 => 'username'];
 				case self::SUBJECT_SHARED_GROUP_SELF:
 				case self::SUBJECT_UNSHARED_GROUP_SELF:
 					return [
--- a/apps/files_sharing/lib/AppInfo/Application.php
+++ b/apps/files_sharing/lib/AppInfo/Application.php
@@ -150,9 +150,6 @@ class Application extends App {
 			);
 		});
 
-		/*
-		 * Register trashbin service
-		 */
 		$container->registerService('Hooks', function ($c) {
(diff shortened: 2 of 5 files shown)


jvillafanez:
> However, IMHO, holding them in files_sharing app is better. So, I added this event listener on files_sharing app

I'd prefer to keep the same events in the same place. If we start spreading the location it will be a pain to find where the events are being triggered or are being listened. I'm not saying that putting the event listener in the files_sharing app is a bad idea, but we should move the rest of the events there too.

Regarding the specific event, I guess it's triggered by either core or the files_sharing app. The one interested in getting that event is the activity app. If it's triggered by core, why would the files_sharing app need to get that event to send it to the activity app? Do other similar events follow the same flow?

karakayasemi:
> I'd prefer to keep the same events in the same place. If we start spreading the location it will be a pain to find where the events are being triggered or are being listened. I'm not saying that putting the event listener in the files_sharing app is a bad idea, but we should move the rest of the events there too.

As far as I see, currently only `files app` and `files_sharing app` activities are published by `activity app`. Moreover, both `files_sharing app` and `files app` have already `Activity` class which extending `OCP\Activity\IExtension` to supply the rest of the related things for an activity such as translations, icons, etc. The other apps that have activities are storing all the activity related things on their own repo like; 
- https://github.com/owncloud/core/tree/master/apps/comments/lib/Activity
- https://github.com/owncloud/core/tree/master/apps/systemtags/lib/Activity
IMHO, this should be the same for `files` and `files_sharing`. In this way, we can have a more consistent structure on all activity places.

> Regarding the specific event, I guess it's triggered by either core or the files_sharing app. The one interested in getting that event is the activity app. If it's triggered by core, why would the files_sharing app need to get that event to send it to the activity app? Do other similar events follow the same flow?

As far as I see, this flow is common in other apps. I guess, in this way, we can put all the activity related things of an app on the same place. Also, these events are not dispatched for only publishing activity.

karakayasemi:
@jvillafanez How should I continue with this PR? If we decide something, I can change the PR in that way. Your further thoughts can help me here. Thanks.

jvillafanez:
Ok, I've checked the event is triggered by core, not by the files_sharing app.

There is a thing that we should check. Does the files_sharing app provides a "self unshare" action in a controller, or something similar?

The idea I have is that the files_sharing app has a controller with a "unshareFromSelf" method or something similar, which call the core's method. For me, it makes more sense to trigger the activity event at that point instead of waiting for symfony's event. This is probably what happens with the rest of the activity events triggered by the files_sharing app.

The other idea is that core is the one triggering the activity event, but I guess this will be more work, so we can discard this.

If there are no better options, we'll probably have to use this solution.

karakayasemi:
@jvillafanez `files_sharing` app has not any method about self-unshare. Unshare request is same with the WebDAV delete operation, just `OCA\Files_Sharing\SharedStorage` storage provider class handles the delete operation on storage level. So, unfortunately, the files_sharing app has not any controller method for it. Core calls files_sharing app methods for needs. As far as I see, the other files_share events that Activity app listens are also triggered by the core with OC\Share20\Manager like `fromself.unshare` event.

My question here is why Activity app is publishing files_sharing activities? We are publishing events from Activity app, however, the other dependencies (translations, icons, etc.) of these activities are provided by the files_sharing app. IMHO, they should be in the same place. 

In the current situation, whenever we want to add a new activity for the files_sharing app, we should first listen to related event from Activity app and publish the activity in there. After that, we should add translations with another PR in core. The release process of core and Activity app is different, I could not understand why we prefer this like that. IMHO, we should move the other files_sharing activities to `files_sharing` repo like the other apps.

jvillafanez (approved):
I guess there is no better option at the moment.

karakayasemi:
@jvillafanez what do you think about to move other activities from activity app to files_sharing? Should I raise a ticket for it?

jvillafanez:
Changes in the architecture are quite difficult to acomplish. You can open a ticket explaining your plans with some easy usage example, remarking what your solution will improve. Obviously, drawbacks or things that we might lose comparing against the current solution need to be shown.
Note that we'll also need a migration plan in order to reach your solution from the current one.

This is some work to do to open a discussion about it. It's up to you to decide what to do. Note that there is no real guarantee that the plans will go forward even if everyone agrees with those plans.

I haven't dig into the architecture around the activities, so I can't say why it's implemented this way. Maybe I'd reach the same point we have at the moment, or maybe not.

karakayasemi:
backport PR for `stable10` in here: https://github.com/owncloud/core/pull/35193

PR #35094, opened 2019-04-24
Trashbin memory

jvillafanez:
## Description
Reduce memory usage in the TrashExpire background job

## Related Issue
https://github.com/owncloud/enterprise/issues/2925

## Motivation and Context
There have been problems related to excesive memory usage, probably caused by an abnormally big trashbin.
Note that problems with excesive group caching has also been detected, but the fix isn't implemented here yet. In addition, the `Trashbin::getLocations` calls should also be changed.

## How Has This Been Tested?
Manually tested with 3 users and a bunch of files inside a couple of folders. The code has been slightly changed to get the peak memory usage in several places.

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_trashbin/lib/BackgroundJob/ExpireTrash.php
+++ b/apps/files_trashbin/lib/BackgroundJob/ExpireTrash.php
@@ -87,7 +87,7 @@ class ExpireTrash extends \OC\BackgroundJob\TimedJob {
 				Trashbin::deleteFileIfExpired($this->expiration, $trashFile, $uid);
 			}
 		});
-		
+
 		\OC_Util::tearDownFS();
 	}
 
--- a/apps/files_trashbin/lib/BackgroundJob/ExpireTrash.php
+++ b/apps/files_trashbin/lib/BackgroundJob/ExpireTrash.php
@@ -81,8 +81,11 @@ class ExpireTrash extends \OC\BackgroundJob\TimedJob {
 			if (!$this->setupFS($uid)) {
 				return;
 			}
-			$dirContent = Helper::getTrashFiles('/', $uid, 'mtime');
-			Trashbin::deleteExpiredFiles($dirContent, $uid);
+
+			$trashFilesGenerator = Helper::getTrashFilesGenerator('/', $uid, 'mtime');
+			foreach ($trashFilesGenerator as $trashFile) {
+				Trashbin::deleteFileIfExpired($this->expiration, $trashFile, $uid);
+			}
 		});
 		
 		\OC_Util::tearDownFS();
--- a/apps/files_trashbin/lib/Helper.php
+++ b/apps/files_trashbin/lib/Helper.php
@@ -130,7 +130,7 @@ class Helper {
 	 * @param string $user
 	 * @param string $sortAttribute attribute to sort on (default to "name")
 	 * @param bool $sortDescending true for descending sort, false otherwise
-	 * @return Generator a generator to traverse the trash files, getting \OCP\Files\FileInfo objects
+	 * @return \Generator a generator to traverse the trash files, getting \OCP\Files\FileInfo objects
 	 */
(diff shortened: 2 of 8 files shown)


micbar:
@jvillafanez @pako81 Is this code still needed? I see in the original ticket that you need a new ticket?
What about this one?
If valid, it would need more test coverage.

pako81:
@micbar it is still under investigation. We are trying to verify if 10.2.0, which should already come with some improvements in terms of memory usage, is already solving the issue. If not, it may be required to apply this code and check the outcome.
@jvillafanez correct me if I am wrong

jvillafanez:
It's on my radar. I got side-tracked with other more urgent things.
The code is still needed because it fixes a couple of problems regarding high memory usage, but it's under investigation if we need more changes or not.

jvillafanez:
I don't think I can add more tests here:
* `deleteFileIfExpired` in "Trashbin.php" relies on the `self::delete` function, which access to the DB and creates 2 new View objects. In addition, it does a lot of things. Not only the dependencies can't be mocked, but also we'll need access both to the DB and the FS. Unittests are impossible to make here without a heavy refactor.
* `getTrashFilesGenerator` in "Helper.php". It was copied from the `getTrashFiles` and adjusted to use a generator. Most of the copied code can't be unittested. There is a new View object that can't be mocked, and the `Trashbin::getLocations` calls depends on access to the DB, so we can't mock that call.
* `run` in the "ExpireTrash.php" depends a couple of static code. Those calls aren't isolated and depend on several components, so we can't add unittests there.

The rest of the changes should be already covered by previous tests.

jvillafanez:
Not sure how to proceed here. There are several problems with the tests due to a confusing architecture around the Files/Cache.

Options:
* Remove the change around Files/Cache/Cache (no more generator). We'll keep the rest of the changes. This will increase the memory usage, but it should fix CI and it also keep the changes at minimum.
* Spread the `getFolderContentsGenerator` and `getFolderContentsByIdGenerator` to all the Cache subclasses. I need to investigate if this is possible or not, mainly because of how generators work and the expected behaviour of the cache wrappers.
* Try to fix the CacheWrapper and subclasses by NOT inheriting from the Cache class. This probably will break things and we'll have additional things to fix.
* Try to implement a proper decorator pattern. The problem is that we probably won't be able to move from the current code to there without breaking a lot of things.

jvillafanez:
I think I'll close this PR and create a new one fixing the memory leaks found.

Although using a generator in the trashbin was a good idea to reduce memory usage, it requires changes in the Cache implementation which either needs to be in the interface or we need to downcast to a private implementation.

For consistency, if we downcast to a private implementation (which is wrong because the expected implementation is in core, not in the trashbin app), we'll also need to handle the rest of the implementation somehow, so we'll still need to rely on the interface. Sound better to rely on the interface in the first place.

If we decide to create a new method in the interface, all the implementations needs to provide something similar. This is starting to get out of scope. The additional problem is that the wrappers also need to implement the same method. This method have to return a generator which wraps the returned generator.
This is either reshaping the architecture or adding new functionality, but not fixing a bug.

jvillafanez:
Simplified version of the expected solution is in https://github.com/owncloud/core/pull/35612 
It doesn't have the changes for the trashbin, but the other 2 problems are also fixed there.

jvillafanez:
@micbar we should probably close this PR in favour of https://github.com/owncloud/core/pull/35612 . As said above, I don't think it's possible to fix the tests without major changes.

micbar:
@jvillafanez ok for me , please proceed as proposed :-)

jvillafanez:
Closing in favour of https://github.com/owncloud/core/pull/35612

phil-davis:
Closed, so backport not needed

Issue #35093, opened 2019-04-24
header set twice in 4xx responses


individual-it:
### Steps to reproduce
I'm testing the backport of https://github.com/owncloud/core/pull/31231 for 10.2.0RC1
and it looks to me that the change made it worse, but also I might have a setup issue

1. enable apache `headers` module
1.  in apache config add 
    ```
    <IfModule mod_headers.c>
    Header set X-Content-Type-Options "bloedsin"
    </IfModule>
    ```
1. make sure the header is set for oC unrelated requests
1. update htaccess `sudo -u www-data ./occ  maintenance:update:htaccess` (maybe not needed)
1. force a 404 or 401 response e.g `curl -v http://localhost/owncloud-10.2.0/rc1_a/remote.php/dav/files/admin/file.txt`

### Expected behaviour
`X-Content-Type-Options` header set once

### Actual behaviour
in 10.2.0RC1 the header is set twice
```
< HTTP/1.1 401 Unauthorized
< Date: Wed, 24 Apr 2019 12:04:12 GMT
< Server: Apache/2.4.39 (Ubuntu)
< X-Content-Type-Options: nosniff
< X-XSS-Protection: 1; mode=block
< X-Robots-Tag: none
< X-Frame-Options: SAMEORIGIN
< X-Download-Options: noopen
< X-Permitted-Cross-Domain-Policies: none
< Set-Cookie: oc1fqx9gxt25=87i4aopfu3m27a2pcfnrfpt50l; path=/owncloud-10.2.0/rc1_a; HttpOnly
< Expires: Thu, 19 Nov 1981 08:52:00 GMT
< Cache-Control: no-store, no-cache, must-revalidate
< Pragma: no-cache
< Set-Cookie: oc_sessionPassphrase=MPQ1wpbNO8gu312ik3Sr2WPO6fWNo3DXYA83MKgdfMEA3LWxDrs6uRCiDeesfCwp1YFMpeaRU4LW0vst4nh9hbNZtuvdfZZEO8BhrJMiOxTeFnWa%2B3ydxp1rVn1gX8PU; path=/owncloud-10.2.0/rc1_a; HttpOnly
< Content-Security-Policy: default-src 'none';
< WWW-Authenticate: Basic realm="ownCloud", charset="UTF-8"
< X-Content-Type-Options: bloedsin
< Content-Length: 448
< Content-Type: application/xml; charset=utf-8
```

in 10.1.1 it is set once
```
[...]

PR #35099, opened 2019-04-25
Avoid headers set twice in some situations

benrubson:
Hi,

This PR follows #31231, and solves behavior discussed in #35093.
We are then sure headers are set only one time, whatever the table, `onsuccess` or `always`, Apache configuration uses to set its headers.

Thank you 👍

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -15,22 +15,30 @@
 
   <IfModule mod_env.c>
     # Add security and privacy related headers
+    Header      unset X-Content-Type-Options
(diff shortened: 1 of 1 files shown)


patrickjahns:
I'll have a look - if we merge this we need to port it over to stable10 and 10.2.0

PVince81:
@benrubson please backport this to the "release-10.2.0" branch

benrubson:
@PVince81 see #35118 👍

PR #35073, opened 2019-04-22
[10.2.0] Filter unavailable users and remove unavailable shares on polling

VicDeo:
Backport of https://github.com/owncloud/core/pull/35010

## Description
Protect polling from exploding on unavailable users

## Motivation and Context
`An unhandled exception has been thrown:
TypeError: Argument 1 passed to OCA\Files_Sharing\External\MountProvider::getMountsForUser() must implement interface OCP\IUser, null given,` when UserManager knows nothing about a userId

## How Has This Been Tested?
1. Create and accept a federated share
2.  Execute `update oc_share_external set user="Iwouldliketobreakthepolling" limit 1;`

### Expected
Polling goes through all entries with no termination

### Actual
Polling  terminates with `TypeError: Argument 1 passed to OCA\Files_Sharing\External\MountProvider::getMountsForUser() must implement interface OCP\IUser, null given`

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #35106, opened 2019-04-26
[stable10] Filter unavailable users and remove unavailable shares on polling

phil-davis:
Backport #35010

Also note, has been backported to release-10.2.0 in PR #35073


phil-davis:

PR #35097, opened 2019-04-25
[10.2.0] Respect 'writable' appdir flag on update

VicDeo:
Backport of https://github.com/owncloud/core/pull/35072
## Description
Check if the current appdir is marked as writable before updating the app

## Related Issue
- Fixes #35031

## Motivation and Context
Fix App is updated inside the directory that is marked as not writable.

## How Has This Been Tested?
see #35031

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #35072, opened 2019-04-22
Respect 'writable' appdir flag on update

VicDeo:
## Related Issue
- Fixes #35031

## Motivation and Context
Fix App is updated inside the directory that is marked as not writable.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35107, opened 2019-04-26
[stable10] Respect 'writable' appdir flag on update

phil-davis:
Backport #35072

Note: also being backported to release-10.2.0 in PR #35097


phil-davis:

PR #35104, opened 2019-04-26
[10.2.0] add new capability privateLinksDetailsParam

karakayasemi:
Backport of #35101
## Description
add capability entry for this feature: owncloud/core#32152

## Related Issue
- Fixes https://github.com/owncloud/docs/issues/265

## Motivation and Context
Show clients whether this feature #32152 is supported or not.

## How Has This Been Tested?
Unit tests added, also tested via ocs rest api for capabilities.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #35101, opened 2019-04-26
add new capability privateLinksDetailsParam

karakayasemi:
## Description
add capability entry for this feature: owncloud/core#32152

## Related Issue
- Fixes https://github.com/owncloud/docs/issues/265

## Motivation and Context
Show clients whether this feature #32152 is supported or not.

## How Has This Been Tested?
Unit tests added, also tested via ocs rest api for capabilities.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35108, opened 2019-04-26
[stable10] add new capability privateLinksDetailsParam

phil-davis:
Backport #35101

Note: has also been backported to release-10.2.0 in PR #35104


phil-davis:

PR #35082, opened 2019-04-23
[10.2.0] Backport of Remove share permission check from the file up…

sharidas:
…load

Remove share permission check from the file upload
js file.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Remove `sharePermission` checks from the file-upload.js. We are no where using it and there is a problem in the code due to this. When we know its a file, we just have to check the fileinfo ( in the if condition ). The problem this PR tries to address is, when files are uploaded, the file information should be shown. Instead of this, the directory information was shown.

This also fixes the cancel button problem in the conflict dialog. The cancel button clicked once cancels the conflict dialog.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35056
- Fixes https://github.com/owncloud/core/issues/35057

## Motivation and Context

Remove sharePermission checks from the file-upload.js file. This solves 2 problems:
1. The file which has conflict will show the file information in the conflict dialog popup and not the directory info.
2. Cancel button of the conflict dialog pop up will work with one click, instead of multiple click issue mentioned at #35057.

## How Has This Been Tested?

- Tested as per https://github.com/owncloud/core/pull/35060#issue-271583209
- Verified that when file is uploaded from UI it shows file size and not folder size.
- Verified that `Cancel` button of conflict dialog works correctly.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35060, opened 2019-04-18
Remove share permission check from the file upload

sharidas:
Remove share permission check from the file upload
js file.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Remove `sharePermission` checks from the file-upload.js file. We are no where using it and there is a problem in the code due to this. When we know its a file, we just have to check the fileinfo ( in the if condition ). The problem this PR tries to address is, when files are uploaded, the file information should be shown. Instead of this, the directory information was shown. 

This also fixes the cancel button problem in the conflict dialog. The cancel button clicked once cancels the conflict dialog.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35056
- Fixes https://github.com/owncloud/core/issues/35057

## Motivation and Context

Remove `sharePermission` checks from the file-upload.js file. This solves 2 problems:
1. The file which has conflict will show the file information in the conflict dialog popup and not the directory info.
2. Cancel button of the conflict dialog pop up will work with one click, instead of multiple click issue mentioned at https://github.com/owncloud/core/issues/35057.

## How Has This Been Tested?

-  Kindly see the screen share 
granularPermission

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35109, opened 2019-04-26
[stable10] Backport of Remove share permission check from the file upload

phil-davis:
Backport #35060 

Note: has also been backported to release-10.2.0 in PR #35082


phil-davis:
This happened in #35082 to `release-10.2.0`

PR #35059, opened 2019-04-18
[10.2.0] Fix missing progress bar in files drop view

PVince81:
backport of https://github.com/owncloud/core/pull/35032 for 10.2

PR #35032, opened 2019-04-16
Fix missing progress bar in files drop view

PVince81:
## Description
The uploader is expecting the progress bar element to be in the DOM when
created. This fix makes sure to append the PublicUploadView's element to
the DOM before rendering it.

## Related Issue
https://github.com/owncloud/enterprise/issues/3281

Regression introduced through https://github.com/owncloud/core/pull/32170 where the progress bar stops working.

## How Has This Been Tested?
Manual test.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] backport to release-10.2.0

PR #35110, opened 2019-04-26
[stable10] Fix missing progress bar in files drop view

phil-davis:
Backport #35032 

Note: has also been backported to release-10.2.0 in PR #35059


phil-davis:

PR #35116, opened 2019-04-29
Always provide return status for functions

sharidas:
Provide return for functions. This would help
to know the status of methods called.

## Description

Functions should return values which would help to process whether the call to a method was successful or not. For instance in a scenario, where a file is uploaded in an objectstorage the filecache is updated but there is no clue if the file upload to the objectstorage is successful or not. Lets say if the file upload to objectorage fails and the system considers it as successful, then user will end up in a situation where:
- there is an entry in the filecache for the file
- there is no entry in the object storage
This PR addresses the issue.
Note: This PR is under development.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

The methods/function should return the status of file upload to object storage. Say if the status is 200 return true. Else return false. This would also help to process the status of the method called for objectorage upload.

## How Has This Been Tested?

- At the end of `writeObject()` return false and then run curl command :
```
 ✘ sujith@sujith-ownCloud  /tmp  curl -u admin:admin -T file2.txt 'http://localhost/testing/remote.php/dav/files/admin/'
<!DOCTYPE html>
<html class="ng-csp" data-placeholder-focus="false" lang="en" >
        <head data-requesttoken="">
                <meta charset="utf-8">
                <title>
                ownCloud                </title>
                <meta http-equiv="X-UA-Compatible" content="IE=edge">
                <meta name="referrer" content="never">
                <meta name="viewport" content="width=device-width, minimum-scale=1.0, maximum-scale=1.0">
                                        <meta name="apple-itunes-app" content="app-id=543672169">
                                <meta name="theme-color" content="#1d2d44">
                <link rel="icon" href="/testing/core/img/favicon.ico">
                <link rel="apple-touch-icon-precomposed" href="/testing/core/img/favicon-touch.png">
                <link rel="mask-icon" sizes="any" href="/testing/core/img/favicon-mask.svg" color="#1d2d44">
                                        <link rel="stylesheet" href="/testing/core/vendor/select2/select2.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/styles.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/inputs.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/header.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/icons.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/fonts.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/apps.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/global.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/fixes.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/multiselect.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/mobile.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/vendor/jquery-ui/themes/base/jquery-ui.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/jquery-ui-fixes.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/tooltip.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/apps/files_versions/css/versions.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/jquery.ocdialog.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                        <link rel="stylesheet" href="/testing/core/css/share.css?v=5cf6c400779c8af51639613f8d1fef7f">
                                                                        <script src="/testing/core/vendor/jquery/dist/jquery.min.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/jquery-ui/jquery-ui.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/underscore/underscore.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/moment/min/moment-with-locales.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/handlebars/handlebars.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/blueimp-md5/js/md5.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/bootstrap/js/tooltip.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/backbone/backbone.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/es6-promise/es6-promise.auto.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/davclient.js/lib/client.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/clipboard/dist/clipboard.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/bowser/src/bowser.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/jquery.ocdialog.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/oc-dialogs.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/js.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/l10n.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/octemplate.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/eventsource.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/config.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/search/js/search.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/oc-requesttoken.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/apps.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/mimetype.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/mimetypelist.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/snapjs/dist/latest/snap.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/vendor/select2/select2.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/oc-backbone.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/oc-backbone-webdav.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/placeholder.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/jquery.avatar.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/files/fileinfo.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/files/client.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/shareconfigmodel.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/shareitemmodel.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialogresharerinfoview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialoglinkshareview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialoglinksocialview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialogmailview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialogexpirationview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialogshareelistview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialogview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/share.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharemodel.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharescollection.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        <script src="/testing/core/js/sharedialoglinklistview.js?v=5cf6c400779c8af51639613f8d1fef7f"></script>
                                        </head>
        <body id="body-login">
                <noscript>
        <div id="nojavascript">
                <div>
                        This application requires JavaScript for correct operation. Please <a href="https://www.enable-javascript.com/" target="_blank" rel="noreferrer">enable JavaScript</a> and reload the page.          </div>
        </div>
</noscript>
                <div class="wrapper">
                        <div class="v-align">
                                                                        <header role="banner">
                                                <div id="header">
                                                        <div class="logo">
                                                                <h1 class="hidden-visually">
                                                                        ownCloud                                                                </h1>
                                                        </div>
                                                        <div id="logo-claim" style="display:none;"></div>
                                                </div>
                                        </header>
                                                                <span class="error error-wide">
        <h2><strong>Internal Server Error</strong></h2>
                <p>The server encountered an internal error and was unable to complete your request.</p>
                <p>Please contact the server administrator if this error reappears multiple times and include the technical details below in your report.</p>
                <p>More details can be found in the <a target="_blank" rel="noreferrer" href="https://doc.owncloud.org/server/11.0/go.php?to=admin-logfiles">server log</a>.</p>
        <br>

        <h2><strong>Technical details</strong></h2>
        <ul>
                <li>Remote Address: 127.0.0.1</li>
                <li>Request ID: SDDwScgu2JguOwzLeNmq</li>
                        </ul>

        </span>
                                <div class="push"></div><!-- for sticky footer -->
                        </div>
                </div>
                <footer role="contentinfo">
                        <p class="info">
                                <a href="https://owncloud.org" target="_blank" rel="noreferrer">ownCloud</a> &ndash; A safe home for all your data                      </p>
                </footer>
        </body>
</html>
 sujith@sujith-ownCloud  /tmp 
```
- Remove the changes made in the above step and re-run the curl command:
```
sujith@sujith-ownCloud  /tmp  curl -u admin:admin -T file2.txt 'http://localhost/testing/remote.php/dav/files/admin/'
 sujith@sujith-ownCloud  /tmp 
```

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Required https://github.com/owncloud/files_primary_s3/pull/205
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -391,11 +391,16 @@ class ObjectStoreStorage extends Common {
 		$fileId = $this->getCache()->put($path, $stat);
 		try {
 			//upload to object storage
-			$this->objectStore->writeObject($this->getURN($fileId), \fopen($tmpFile, 'r'));
+			$status = $this->objectStore->writeObject($this->getURN($fileId), \fopen($tmpFile, 'r'));
 		} catch (\Exception $ex) {
 			$this->getCache()->remove($path);
 			\OCP\Util::writeLog('objectstore', 'Could not create object: ' . $ex->getMessage(), \OCP\Util::ERROR);
 			throw $ex; // make this bubble up
+		}
+		if ($status !== true) {
+			$this->getCache()->remove($path);
+			\OCP\Util::writeLog('objectstore', 'Could not create object: kindly contact your administrator.', \OCP\Util::ERROR);
+			return false;
 		}
 	}
 
--- a/lib/public/Files/ObjectStore/IObjectStore.php
+++ b/lib/public/Files/ObjectStore/IObjectStore.php
@@ -47,6 +47,7 @@ interface IObjectStore {
 	 * @param string $urn the unified resource name used to identify the object
 	 * @param resource $stream stream with the data to write
 	 * @throws \Exception when something goes wrong, message will be logged
+	 * @return null|bool
 	 * @since 7.0.0
 	 */
 	public function writeObject($urn, $stream);
--- a/tests/lib/Files/ObjectStore/ObjectStoreTest.php
+++ b/tests/lib/Files/ObjectStore/ObjectStoreTest.php
@@ -76,14 +76,34 @@ class ObjectStoreTest extends TestCase {
 		$this->assertFalse($this->objectStore->file_exists('test'));
 	}
 
-	public function testGetAndPutContents() {
-		$this->impl->expects($this->once())->method('writeObject');
+	public function providesGetAndPutContents() {
+		return [
+			['123456', true],
+			['123456', false]
+		];
+	}
+
+	/**
+	 * @dataProvider providesGetAndPutContents
+	 * @param string $streamContent
+	 * @param bool $expectedResult
+	 */
+	public function testGetAndPutContents($streamContent, $expectedResult) {
+		$this->impl->expects($this->once())->method('writeObject')->willReturn($expectedResult);
 		$this->assertEquals(5, $this->objectStore->file_put_contents('test.txt', 'lorem'));
 
-		$stream = $this->createStreamFor('123456');
+		$stream = $this->createStreamFor($streamContent);
 
-		$this->impl->expects($this->once())->method('readObject')->willReturn($stream);
-		$this->assertEquals('123456', $this->objectStore->file_get_contents('test.txt'));
+		if ($expectedResult === true) {
+			$this->impl->expects($this->once())->method('readObject')->willReturn($stream);
+		}
+
+		if ($expectedResult === true) {
+			$result = '123456';
+		} else {
+			$result = false;
+		}
+		$this->assertEquals($result, $this->objectStore->file_get_contents('test.txt'));
 	}
 
 	public function testGetScanner() {


sharidas:
- [ ] Tests need to be adjusted. But there is a dependency with https://github.com/owncloud/files_primary_s3/pull/205 due to which the CI is failing for objectstore.

VicDeo (changes requested):
wouldn't it log `undefined variable $status` in case of Exception? I guess you need to put `$status = false;` into the catch block to prevent it.

sharidas:
> wouldn't it log `undefined variable $status` in case of Exception? I guess you need to put `$status = false;` into the catch block to prevent it.

I had verified the code while testing it with objectstorage ( i.e with, files_primary_s3 ),in case if exception is raised from `writeObject` method then the `if ($status !== true) {` would not be executed because in the catch block there is a line `throw $ex;`. So my idea here is `writeObject()` when called can either throw exception or return true ( on success ) or return false ( on failure ).

Before this PR, the scenario was `writeObject()` didn't returned any boolean value for the status. It only handled the exception. Here is the PR where I made modifications for `writeObject()` method: https://github.com/owncloud/files_primary_s3/pull/205.
@VicDeo Let me know if this makes sense?

VicDeo:
@sharidas I was wrong as the Exception is rethrown in the catch block.

VicDeo (approved):
Looks goods, let's merge https://github.com/owncloud/files_primary_s3/pull/205 first and restart the build

jvillafanez:
I disagree. The code assumes that an exception will be thrown if there is an error, so if you want to check if the function has been successful or not, you use the exception mechanism.

The problem here is either the implementation is ignoring the expectations defined by the interface or the "client" code isn't properly handling all the scenarios.
In the first case, the implementation should throw an exception in case it couldn't write the contents.
In the second case, the exception thrown should be properly handled.

The only change that we might consider to do around these classes is that the interface should throw an specific exception. However, I don't know if it could have unforseen consequences at this point, so I think we can skip this.

sharidas:
> I disagree. The code assumes that an exception will be thrown if there is an error, so if you want to check if the function has been successful or not, you use the exception mechanism.

The code relied on exception thrown if there is an error - Yes.
But the problem is in one of the instance at the customer, no exception was thrown and the upload of file was shown successful to the client using which file was uploaded. Which means there was an entry in the filecache and the objectstorage did not had any file uploaded. Which also means the user would see a file in the UI but actually there is no file in the storage. So my idea here is to get return for the upload status of file(s) to the storage. If the file is not uploaded to the object storage and no exception happens during the upload then just return `false`. And this returned value from `writeObject()` method is evaluated in the `writeBack()`  method.

> The only change that we might consider to do around these classes is that the interface should throw an specific exception. However, I don't know if it could have unforseen consequences at this point, so I think we can skip this.

Not sure if its a good idea to skip this :(

jvillafanez:
The objectStore implementation (files_primary_s3?) knows whether the upload has been successful or not, otherwise the solution wouldn't matter. If the implementation knows that the file hasn't been properly uploaded, it's the implementation the one that should throw an exception indicating the error. It should throw an exception instead of returning false.
This behaviour is covered by the interface, so we shouldn't need to touch the interface.

The problem, as you've said, is that the implementation doesn't report the error in the upload. However, the "client" code should know how to handle the error because the interface says that it will throw an exception if the upload fails.

My point here is that the component misbehaving is the implementation, not the interface, so it's the implementation the one that should fix the problem. There is no need to change the interface when the behaviour with upload errors is already defined.

> Not sure if its a good idea to skip this :(

Without further checking more code, I can't say for sure if we can or not change the interface and say that instead of a plain `Exception` the interface will throw an `ObjectStoreException` or something similar. There is also the problem of apps such as files_primary_s3 that could rely on throwing any random exception. This is why I prefer to keep the interface as it is without any additional change.

sharidas:
I am fine if everyone here agrees to have a generic `ObjectStoreException` which would be thrown by files_primary_s3-> `writeObject()` if the file is not uploaded based on the status of the file uploaded. In that case we would have to revert the commit https://github.com/owncloud/files_primary_s3/commit/f4ac0978befca762d7b3450dbc25e5247854f92d

@PVince81 @VicDeo let me know your opinion please

Note: Personally not liking the idea to just rely on exception :(

jvillafanez:
> Personally not liking the idea to just rely on exception :(

Exceptions are a better choice if it's possible to have different failures in the operation. The `writeObject()` function can fail because there is no connection, because the storage doesn't have enough space, because the file has been deleted, and there could be more reasons. At least, you can throw an exception explaining the reason of the failure, specially if the error comes from an external system, which is something you can't do with just "true" or "false".

Returning false might be convenient for some scenarios, but I don't think this is one of them, specially when the interface already says to throw an exception in case of error.

Having both will be confusing for both the client code and the implementation, because the client code will be forced to handle both situations, and the implementation won't know when an exception should be thrown or when it should return false.

PVince81:
historically speaking, we had many places where we used boolean and it turned out to be a bad thing as it wasn't possible to distinguish the reason for errors, for example whether a "file_exists" returning false means the file does not exist or maybe the storage is not available.

so if possible, using exceptions is a better approach

DeepDiver1975 (changes requested):
I'm all in with @jvillafanez - the api are originally designed works on the assumption that in case of any error an exception will be thrown.

This change is not needed at all.

That single issue at the customer site needs to be solved in the object store implementation - files_primary_s3 in this case.

sharidas:
@DeepDiver1975 Yup. Hence I have decided to get merge https://github.com/owncloud/files_primary_s3/pull/210 ( i.e, revert https://github.com/owncloud/files_primary_s3/pull/205 )

And move ahead with exception.

sharidas:
Closing this PR because I have opened another PR which is aligned with the exception solution here https://github.com/owncloud/core/pull/35144

PR #35095, opened 2019-04-25
[10.2.0] share attribute required feature

PVince81:
Backport of https://github.com/owncloud/core/pull/35016 to release-10.2.0

PR #35124, opened 2019-04-29
Refactor share attribute js tests

mrow4a:
@PVince81 required for https://github.com/owncloud/core/pull/35095

### DIFF
--- a/core/js/tests/specs/sharedialogshareelistview.js
+++ b/core/js/tests/specs/sharedialogshareelistview.js
@@ -152,7 +152,7 @@ describe('OC.Share.ShareDialogShareeListView', function () {
 			var $li = listView.$('li').eq(0);
 			var input = $li.find("input[name='test-attribute']");
 			expect(input.is(':checked')).toEqual(true);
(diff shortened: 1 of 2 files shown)


core/js/tests/specs/sharedialogshareelistview.js
```diff
@@ -36,7 +36,7 @@ describe('OC.Share.ShareDialogShareeListView', function () {
 		$('#testArea').append('<input id="mailNotificationEnabled" name="mailNotificationEnabled" type="hidden" value="yes">');
 
 		fileInfoModel = new OCA.Files.FileInfoModel({
-			id: '123',
+			id: 123,
```
PVince81: AFAIK the fileid is a string on master.

since you decided to revert this I guess we don't have fileid as string on stable10.

is this change absolutely necessary to make share attribute work ? if not please remove

mrow4a: Ok, this all is misaligned.. I will just adjust the stable10 separately with the int/string.

PVince81 (approved):
:+1:

PR #35131, opened 2019-04-30
[stable10] Allow configuring thumbnails path

PVince81:
Similar to "cache_path" but for thumbnails.
This was quickly copied from the CacheMountProvider

Why ? Because I tried to set this up on my server with symlinks and it says "following symlinks not allowed".

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -777,6 +777,17 @@ $CONFIG = array(
 'enable_previews' => true,
 
 /**
+ * Location of the thumbnails folder, defaults to `data/$user/thumbnails` where
+ * `$user` is the current user. When specified, the format will change to
+ * `$previews_path/$user` where `$previews_path` is the configured previews base directory
+ * and `$user` will be substituted with the user id automatically.
+ *
+ * For example if `previews_path` is `/var/cache/owncloud/thumbnails` then for a logged in
+ * user `user1` the thumbnail path will be `/var/cache/owncloud/thumbnails/user1`.
+ */
+'previews_path' => '',
+
+/**
  * The maximum width, in pixels, of a preview. A value of `null` means there
  * is no limit.
  */
--- a/lib/private/Files/Mount/PreviewsMountProvider.php
+++ b/lib/private/Files/Mount/PreviewsMountProvider.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * @author Vincent Petry <pvince81@owncloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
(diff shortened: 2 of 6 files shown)


PVince81:
One reason I want a separate thumbnails folder is because I consider this like a cache and don't want this in backups. Also having it separate makes it easier to clean up.

PVince81:
Tested on my own server by patching on top of 10.2.0 RC1, works fine.

My thumbnails are stored in "/var/cache/thumbnails/$user" now.

config/config.sample.php
```diff
@@ -776,6 +776,14 @@
  */
 'enable_previews' => true,
 
+/**
+ * Location of the thumbnails folder, defaults to `data/$user/thumbnails` where
+ * `$user` is the current user. When specified, the format will change to
+ * `$previews_path/$user` where `$previews_path` is the configured previews directory
+ * and `$user` is the user.
```
mmattel: I would add, that `$user` will be set/substituted by ownCloud, only the thumbs base path must be set

PVince81:
- [ ] update doc
- [ ] add unit tests for both the cache mount provider and the new one

PVince81:
@mmattel I've adjusted the doc, please check.

I've now written unit tests for both the previews mount provider and cache mount provider.
Apparently the "$loader" wasn't properly passed in.

@DeepDiver1975 please review

DeepDiver1975 (approved):
Just one minor phpdoc issue - :+1:

lib/private/Files/Mount/PreviewsMountProvider.php
```diff
@@ -0,0 +1,69 @@
...
+ * Mount provider for custom preview storages
+ */
+class PreviewsMountProvider implements IMountProvider {
+	/**
+	 * @var IConfig
+	 */
+	private $config;
+
+	/**
+	 * ObjectStoreHomeMountProvider constructor.
```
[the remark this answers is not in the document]
PVince81: fixed now in both cache and preview providers

config/config.sample.php
```diff
@@ -776,6 +776,17 @@
...
 'enable_previews' => true,
 
+/**
+ * Location of the thumbnails folder, defaults to `data/$user/thumbnails` where
+ * `$user` is the current user. When specified, the format will change to
+ * `$previews_path/$user` where `$previews_path` is the configured previews base directory
+ * and `$user` will be substituted with the user id automatically.
+ *
+ * For example if "previews_path" is "/var/cache/owncloud/thumbnails" then for a logged in
+ * user "user1" the thumbnail path will be "/var/cache/owncloud/thumbnails/user1".
```
mmattel: Can you pls use back-ticks instead quotes
As you correctly do in the block above...

PVince81:
all adjusted, including CI issues

mmattel (approved):

PVince81:
master: https://github.com/owncloud/core/pull/35196

Issue #35086, opened 2019-04-24
File upload on files_drop does not abort  if tried to cancel after waiting for few seconds


skshetry:
Regarding #34097:

Canceling upload in files_drop did not work correctly if it is canceled after waiting for a few sec. But, this does work if it is canceled immediately (which didn't work on `v10.1.1`).  The uploading message does disappear after the file is uploaded.

See test video here: https://cloud.owncloud.com/index.php/s/1psr6GzChn3u83L

_Originally posted by @skshetry in https://github.com/owncloud/core/issues/35054#issuecomment-486070762_

### Steps to reproduce
1. Share a public link for files_drop.
2.  Upload a file (upload a large file so you have a larger window to cancel upload).
3. Wait for a few seconds (but not too much that the time you cancel, it's already uploaded).
4. Cancel upload

### Expected behavior
`Uploading...` message disappears and the file uploading is canceled. 

### Actual behavior
`Uploading...` doesn't disappear. The file gets uploaded and the message(`Uploading...`) disappears.

### Server configuration
**Operating system**: Ubuntu 18.04.2

**Web server:** Apache/2.4.29 (Ubuntu)

**Database:** mysql  Ver 14.14 Distrib 5.7.25, for Linux (x86_64)

**PHP version:** PHP 7.2.17-1+ubuntu18.04.1+deb.sury.org+3 (cli) (built: Apr 10 2019 10:50:57) ( NTS )

**ownCloud version:** (see ownCloud admin page) ownCloud 10.2.0 RC1 (stable) 

**Updated from an older ownCloud or fresh install:** Fresh install

**Where did you install ownCloud from:** Tarball

**Are you using external storage, if yes which one:** local/sftp

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** No

### Client configuration
**Browser:** Firefox 66.0.3 (64-bit)/ Chrome 73

**Operating system:** Ubuntu 18.04.2

skshetry:
@PVince81, looks like the bug was when we try to cancel upload after already canceling upload once in the first try. 
So, if you:
1. Create a files_drop link.
2. Open the share link.
3. Try to upload a file and then cancel it.
4. Try to again re-upload and cancel it.

Then, the file upload does not cancel. It works if we reload the tab and retry.

PVince81:
ok thanks, so it's a different bug

PVince81:
@sharidas can you debug this ?

sharidas:
[...]

PR #35130, opened 2019-04-30
Aborted uploads are not cleared properly

sharidas:
File uploads should also be checked if they
are aborted or not. This would help to check
if the file is aborted or not. If aborted then
it would be cleared from the list.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

When user uploads file and then aborts the upload ( the abort is done for the first time ), the global value of `this._uploads` is not updated properly. If the file is aborted, then as of now the state of `upload` is `pending`. And hence `this._uploads` will always have currently aborted upload. So when the user tries to upload a file next time ( say second time ), then the value in the `this_upload` mismatches. Because it has previous value. And the error could also be seen in the console. 

In this change set I have introduced an attribute `aborted` for the `upload`. When the upload is aborted, `aborted` attribute of `upload` is set to true. And hence while clearing the uploads, it checks whether the current aborted upload should be added to remaining uploads or not.

In simple terms its an issue of not clearing uploads list.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35086

## Motivation and Context

The aborted upload(s) should be cleared from the list.

## How Has This Been Tested?

- Here is how I have tested in my instance 
upload-cancel
- I have also tested without aborting/cancelling the upload. And it worked. The files were available in the UI. Was able to download them.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35134, opened 2019-05-02
[10.2.0] Backport of Aborted uploads are not cleared properly

sharidas:
File uploads should also be checked if they
are aborted or not. This would help to check
if the file is aborted or not. If aborted then
it would be cleared from the list.

## Description

When user uploads file and then aborts the upload ( the abort is done for the first time ), the global value of `this._uploads` is not updated properly. If the file is aborted, then as of now the state of `upload` is `pending`. And hence `this._uploads` will always have currently aborted upload. So when the user tries to upload a file next time ( say second time ), then the value in the `this_upload` mismatches. Because it has previous value. And the error could also be seen in the console. 

In this change set I have introduced an attribute `aborted` for the `upload`. When the upload is aborted, `aborted` attribute of `upload` is set to true. And hence while clearing the uploads, it checks whether the current aborted upload should be added to remaining uploads or not.

In simple terms its an issue of not clearing uploads list.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35086

## Motivation and Context

The aborted upload(s) should be cleared from the list.

## How Has This Been Tested?

- Tested similarly mentioned at https://github.com/owncloud/core/pull/35130#issue-274743297

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -682,6 +682,7 @@ OC.Uploader.prototype = _.extend({
 		this.log('canceling uploads');
 		jQuery.each(this._uploads, function(i, upload) {
 			upload.abort();
+			upload.aborted = true;
 		});
 		this.clear();
 	},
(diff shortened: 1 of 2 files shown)


sharidas:
Original PR: https://github.com/owncloud/core/pull/35130

PVince81 (approved):
:+1:

PR #35144, opened 2019-05-03
Add new exceptions for objectstore

sharidas:
Add new exceptions for objectstore.

## Description

Introducing new exception(s) to objectstorage. The idea of having new exception is because the interface `IObjectStore` relies on exception.
So we have an exception heirarchy as shown below:
ObjectStoreWriteException

The idea is any operation specific exceptions ( like `readObject`, `writeObject` and `deleteObject` ) will use exceptions like `ObjectStoreWriteException` ( for `writeObject` ) or `ObjectStoreReadException` ( for `readObject` ) or `ObjectStoreDeleteException` ( for `deleteObject` )  which inherits from `ObjectStoreOperationException`. The `ObjectStoreOperationException` inherits from `ObjectStoreException`.

This PR focuses only on the 3 exceptions shared above( i.e, ObjectSoreException, ObjectStoreOperationException and ObjectStoreWriteException ). So in future lets say we need to add exception for `readObject` then we could create new exception like `ObjectStoreReadException` which would inherit from `ObjectStoreOperationException` and so on.
But lets say if we need to have an exception for connection, then we have an option to create, say `ObjectStoreConnectionException` which can be inherited from `ObjectStoreException`

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Introduces new exception for objectstorage.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Requires https://github.com/owncloud/files_primary_s3/pull/212
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/public/Files/ObjectStore/ObjectStoreWriteException.php
+++ b/lib/public/Files/ObjectStore/ObjectStoreWriteException.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCP\Files\ObjectStore;
+
+/**
+ * Exception when file upload to object store fails
+ * Class ObjectStoreWriteException
+ *
+ * @package OCP\Files\ObjectStore
+ * @since 10.3.0
+ */
+class ObjectStoreWriteException extends ObjectStoreOperationException {
+	/** @var int  */
+	private $statusCode;
+
+	/**
+	 * ObjectStoreWriteException constructor.
+	 *
+	 * @param string $message
+	 * @param int $code
+	 * @param \Exception|null $previous
+	 * @since 10.3.0
+	 */
+	public function __construct($message = "", $code = 0, \Exception $previous = null) {
+		parent::__construct($message, $code, $previous);
+	}
+}
--- a/lib/public/Files/ObjectStore/ObjectStoreException.php
+++ b/lib/public/Files/ObjectStore/ObjectStoreException.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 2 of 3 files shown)


jvillafanez:
The `getStatusCode()` should be in the `ObjectStoreOperationException`. I expect the read and delete operation to have also a status code we can check. The `ObjectStoreWriteException` (and others) will set the status code using the parent constructor, and they'll use the inherited method (no need to redefine it)

Another question: do we need to distinguish between the exception code and the status code? Is there any potential problem if we set the exception code as the status code? Just trying to simplify the number of things we need to get to throw the exception.

lib/public/Files/ObjectStore/ObjectStoreWriteException.php
```diff
@@ -0,0 +1,58 @@
...
+	 *
+	 * @param string $message
+	 * @param int $statusCode
+	 * @param int $code
+	 * @param \Exception|null $previous
+	 * @since 10.2.0
+	 */
+	public function __construct($message = "", $statusCode = 0, $code = 0, \Exception $previous = null) {
+		$this->statusCode = $statusCode;
+		$message = $message . " The status code from object storage is $statusCode.";
```
jvillafanez: I don't think we need to rewrite the message here. There is no expectation of subclassing this particular exception, so whatever throws this exception should write the appropiate message by itself.
If you still want to show the status code automatically in the message, You can try to use `$message = "($statusCode) $message";` (just prepending the status code to the message), which is more compact and it doesn't feel so invasive.

sharidas:
##### Regarding readObject statusCode

`statusCode` for read is not available. The only exception it would throw is from `init()` call in the `readObject()` ( in files_primary_s3 ). The reason is that it expects return from fopen. Hence even if it fails all we would get a `false`.

##### Regarding deleteObject statusCode

The return from `$this->connection->deleteObject()` does have a result with `@metadata`. Here is the list from `@metadata`:
```
'statusCode' => 204,
  'effectiveUri' => 'http://localhost:8000/owncloud/urn%3Aoid%3A35',
  'headers' => 
  array (
    'x-amz-delete-marker' => 'true',
    'x-amz-version-id' => '3938343432373839363432303134393939393939524730303120203231',
    'server' => 'S3 Server',
    'x-amz-id-2' => '38e32c73ae43b4071577',
    'x-amz-request-id' => '38e32c73ae43b4071577',
    'date' => 'Tue, 07 May 2019 06:25:58 GMT',
    'connection' => 'close',
  ),
  'transferStats' => 
  array (
    'http' => 
    array (
      0 => 
      array (
      ),
    ),
  ),
)
```

jvillafanez:
There are tests failing.

phil-davis:
Unit test fail https://drone.owncloud.com/owncloud/core/17672/127 is due to the current issue #35330 - there is a unit test that is uccrently always failing in master

jvillafanez (approved):

lib/public/Files/ObjectStore/ObjectStoreWriteException.php
```diff
@@ -0,0 +1,46 @@
...
+	/** @var int  */
+	private $statusCode;
+
+	/**
+	 * ObjectStoreWriteException constructor.
+	 *
+	 * @param string $message
+	 * @param int $code
+	 * @param \Exception|null $previous
+	 * @since 10.2.0
```
DeepDiver1975: 10.2 is Out. This will Go Info 10.3

DeepDiver1975 (changes requested):

lib/public/Files/ObjectStore/ObjectStoreWriteException.php
sharidas: Updated all new files created in this PR with `10.3.0`

sharidas:
@DeepDiver1975 I have updated the PR as requested. Kindly review the change set.

DeepDiver1975 (approved):

sharidas:
Backport PR to stable10 : https://github.com/owncloud/core/pull/35389

PR #35156, opened 2019-05-06
Extend SharingApi test context to test uppercase to lowercase conversion

micbar:
## Description
Extend Acceptance Test Suite

## Related Issue
- Needed for https://github.com/owncloud/guests/pull/325

## Motivation and Context
We need to test the stability of the SharingAPI even if we feed it with uppercase/lowercase differences to the correctn UID

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -875,6 +875,28 @@ trait Sharing {
 	}
 
 	/**
+	 * @Given /^user "([^"]*)" has shared (?:file|folder|entry) "([^"]*)" with user with uppercase letters "([^"]*)"(?: with permissions (.*))?$/
+	 *
+	 * @param string $user1
+	 * @param string $filepath
+	 * @param string $user2
+	 * @param int $permissions
+	 *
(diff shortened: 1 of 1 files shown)


individual-it:
in owncloud/guests#325 you don't use that new step and I don't understand why creating a new function and not using `userHasSharedFileWithUserUsingTheSharingApi()` ?

micbar:
@individual-it
1)  I do not use it in the guest pr because the new method is not in daily-stable-10-qa.
I intend to use it.

2) The other method throws an error if you use it with an uppercase UserID and the shareAPI returns the UserID converted to lowercase. We need a method which accepts the conversion of the UserID.

individual-it:
so we could use `\strtolower` every time. meaning put it into `userHasSharedFileWithUserUsingTheSharingApi()` as it will not make a difference for the other tests and we can have one function less

individual-it:
closing in favour of https://github.com/owncloud/core/pull/35342

PR #35184, opened 2019-05-09
reorganization of createTags.feature into createTags, removeTag and deleteTags

haribhandari07:
## Description

tests/acceptance/features/webUITags/createTags.feature has scenarios that test creating, removing and deleting tags.
I reorganized them into three different feature files:

1. createTags.feature
2. removeTags.feature
3. removeTags.feature

## Related Issue

- Fixes <issue_link>
#35182
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35186, opened 2019-05-09
[stable10] reorganization of createTags.feature into createTags, removeTag and deleteTags

haribhandari07:
backport for #35184


phil-davis:
@HariBhandari07 there are other commits here also. What happened?

phil-davis (approved):
Note: not so many scenarios needed to be removed from `createTags.feature` in `stable10` because the newer scenarios had not yet been backported from `master`

PR #35194, opened 2019-05-09
Fix permission and update behaviour for public link

### LINKED ISSUE #35192: Changes required for public link Download View Upload as it does not work as expected
After discussion with PM here is the summary:
- Remove the Update permission for the `Download / View / Upload` link.
- User should be able to create file or folder with the link.
- User cannot update or modify the file with the same name. Say for example if there is a file `a.txt` in the public link with permission `Download / View / Upload` link, then user cannot upload file with same name.
- User cannot create folder with same name in the public link. That is if there is a folder with name `foo` in the public link. Then user cannot create another folder with the same name `foo`.
- When the user clicks the 3 dots in the public link only download option should be available.
- There should not be any delete reference in the UI. That is from the public link user should not be able to delete a file or folder.
- There should not be any reference to rename in UI ( in the current UI rename option is available ).

@pmaier1

sharidas:
Remove update permission for Download/View/Upload
option for public link. And hence the behaviour of
public links created using this option has changed.
1. User cannot delete the files/folders inside this public link.
2. User cannot update the files which are already present
in the public link
3. User cannot rename files.

## Description

This PR addresses the changes below:
When user selects the `Download / View / Upload` option to create public link, the update permission for the link is no longer available. Which means the share permission will be `5` ( i.e, READ | CREATE ). And due to this change, the public link UI behaves as follows:
When a public link share is created for a folder with permission `Download / View / Upload` below are the changes observed:
- Cannot delete the files.
- Folders can be renamed. The folder rename occurs because of calculation PERMISSION_SHARE | PERMISSION_CREATE | PERMISSION_READ. And this results in `21`. Now https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/Node.php#L344 adds `CK` to the permissions. Which brings the `Rename` to the folder. But I have observed that if a folder when shared to an oC user ( within oC instance ), change the edit permissions of the shared folder to create and delete. The folder inside the shared folder gets `Rename`, `Download` and `Delete`. Whereas the files inside the shared folder get `Download` and `Delete`. Now change the permission of shared folder to `create` and `change`. The folder inside the shared folder gets `Rename` and `Download` option. Where as the files inside the shared folder gets `Download` and `Delete`. 
Let me know considering this behaviour if the folder rename visible in the UI looks ok or not?
- Files and Folders can be downloaded.
- Folders with same name cannot be created.
- Files with same name when uploaded throws notification message in the UI.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35192

## Motivation and Context

Updated the permission for the `Download / View / Upload` option to create public link. Accordingly minor changes to update the UI.

## How Has This Been Tested?

- [x] Create public link with `Download / View / Upload` and verified the share permissions in the share table is `5`. Here is the screenshot:
```
mysql> select id, share_name, file_target, permissions, token from oc_share;
+----+-------------+-------------------+-------------+-----------------+
| id | share_name  | file_target       | permissions | token           |
+----+-------------+-------------------+-------------+-----------------+
|  1 | Public link | /publicLinkUpload |           5 | DiMnszGUqdxXdRJ |
+----+-------------+-------------------+-------------+-----------------+
1 row in set (0.00 sec)

mysql>
```
- [x] Tested by uploading a file, say `a.txt` to the newly created public link
- [x] Tested uploading same file name results in notification as shown 
notification
- [x] Tested and no conflict dialog window appears when same file is uplaoded.
- [x] Tested creating a folder in the new public link
- [x] Tested cannot create same folder name which already exists in the UI 
samefolder
- [x] Tested rename not shown in the files listed in the public link
- [x] Tested and shows rename for the folder in the public link
- [x] Tested delete is not shown in the 3 dots for folder and files. Also when all the files are selected no delete is shown in the UI.
- [x] Tested download is shown when the 3 dots are clicked for the folder and for file. Also when all files are selected download button is shown in the UI.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -831,6 +831,11 @@ OC.Uploader.prototype = _.extend({
 			}
 			var fileInfo = fileList.findFile(file.name);
 			if (fileInfo) {
+				var sharePermission = parseInt($("#sharePermission").val());
+				if (sharePermission === (OC.PERMISSION_READ | OC.PERMISSION_CREATE)) {
+					OC.Notification.show(t('files', 'The file {file} already exists', {file: fileInfo.name}), {type: 'error'});
+					return false;
+				}
 				conflicts.push([
 					// original
 					_.extend(fileInfo, {
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -461,7 +461,7 @@ class Share20OcsController extends OCSController {
 				);
 			} elseif ($permissions === \OCP\Constants::PERMISSION_CREATE ||
 				$permissions === (\OCP\Constants::PERMISSION_READ | \OCP\Constants::PERMISSION_CREATE | \OCP\Constants::PERMISSION_UPDATE | \OCP\Constants::PERMISSION_DELETE) ||
-				$permissions === (\OCP\Constants::PERMISSION_READ | \OCP\Constants::PERMISSION_UPDATE | \OCP\Constants::PERMISSION_CREATE)) {
+				$permissions === (\OCP\Constants::PERMISSION_READ | \OCP\Constants::PERMISSION_CREATE)) {
 				$share->setPermissions($permissions);
 			} else {
 				// because when "publicUpload" is passed usually no permissions are set,
@@ -776,6 +776,7 @@ class Share20OcsController extends OCSController {
 			if ($newPermissions !== null &&
 				$newPermissions !== Constants::PERMISSION_READ &&
 				$newPermissions !== Constants::PERMISSION_CREATE &&
+				$newPermissions !== (Constants::PERMISSION_READ | Constants::PERMISSION_CREATE) &&
 				// legacy
 				$newPermissions !== (Constants::PERMISSION_READ | Constants::PERMISSION_CREATE | Constants::PERMISSION_UPDATE) &&
 				// correct
--- a/core/js/sharedialoglinkshareview.js
+++ b/core/js/sharedialoglinkshareview.js
@@ -265,8 +265,8 @@
 
 				publicUploadWriteLabel       : t('core', 'Download / View / Upload'),
 				publicUploadWriteDescription : t('core', 'Recipients can view, download and upload contents.'),
-				publicUploadWriteValue       : OC.PERMISSION_READ | OC.PERMISSION_UPDATE | OC.PERMISSION_CREATE,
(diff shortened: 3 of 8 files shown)


PVince81:
> Let me know considering this behaviour if the folder rename visible in the UI looks ok or not?

what happens when local shares set with "read + create", is folder renaming inside still possible ?
it seems you only mentionned "create+delete" and "create+change" but those don't match the public link one which has "read + create", so it's difficult to compare.

When writing these old tests I didn't realize that files and folders inside a share would get different actual permissions: https://github.com/owncloud/docs/issues/361
According to said doc, "read+create" would not grant any rename ability.

What happens if you actually rename the folder, does it go through ? I wonder if it's just a visual bug...

@sharidas

individual-it:
@dpakach please test this and adjust the tests you are writing accordingly

sharidas:
@PVince81 

When local share created with `create` + `read`, then the rename option is available for the folder inside the share. But when the recipient user of the share tries to rename the folder it shows up 
localshare_cannot_rename

For the public link share created with `Download / View / Uplaode` the rename does appear for the folder inside the public link share folder. But when user tries to rename the folder the same error notification appears 
public_link_no_rename

PVince81:
@sharidas ok, so this tells me that the permission is enforced correctly on the server side but somehow the UI decides to display the icon.

can you investigate ? (check with older versions) I don't remember this being an issue a year ago or earlier. depending what we find and when we can decide to either fix it right away or consider a "known issue"

sharidas:
> can you investigate ? (check with older versions) I don't remember this being an issue a year ago or earlier. depending what we find and when we can decide to either fix it right away or consider a "known issue"

@PVince81 Yes looking into it.

PVince81:
if CI passes from my POV this is good to go.

we can fix the UI thing separately if we find a straightforward solution (which I doubt, knowing the codebase...)

PVince81 (approved):
:+1:

PVince81:
@sharidas please clarify whether you had additional changes in mind.

if not, please backport to release-10.2.0 to save some time (CI)

sharidas:
Issue is found in below versions:
- [x] Tested on 10.0.0  and issue is found
- [x] Tested on 9.1.8 and issue is found

PVince81:
@sharidas ok, let's raise a ticket and address separately

sharidas:
Backport to release branch https://github.com/owncloud/core/pull/35197

sharidas:
> @sharidas ok, let's raise a ticket and address separately

Created a ticket https://github.com/owncloud/core/issues/35198

Issue #35195, opened 2019-05-10
add acceptance tests for notify-by-email feature in acceptance tests


dpakach:
### Add acceptance test for notify by email feature in the webUI.

When the users are allowed to send notification for the shares, they can click the notify-by-share button in the sharing dialog in the UI to send email notification to the share notifications. Currently there is no test for this feature. 
Add following tests for this feature.
- [x] webUI #35215

paurakhsharma:
Added tests in #35215

PR #35215, opened 2019-05-14
Add webUI tests for share mail notification

paurakhsharma:
## Description
Add webUi acceptance test for mail notification on internal share.

## Related Issue
#35195

## How Has This Been Tested?
🤖

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebUISharingContext.php
+++ b/tests/acceptance/features/bootstrap/WebUISharingContext.php
@@ -675,7 +675,7 @@ class WebUISharingContext extends RawMinkContext implements Context {
 		$fileRows = $this->sharedWithYouPage->findAllFileRowsByName(
 			$share, $this->getSession()
 		);
-		
+
 		$found = false;
 		foreach ($fileRows as $fileRow) {
 			if ($offeredBy === $fileRow->getSharer()) {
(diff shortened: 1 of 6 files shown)


skshetry (commented):
:+1: 

EDIT: Just a suggestion. Everything looks good to me.

tests/acceptance/features/bootstrap/WebUISharingContext.php
```diff
@@ -776,6 +776,40 @@ public function removesPublicLinkAtCertainPosition($number, $entryName) {
...
+	 */
+	public function theUserShouldNotBeAbleToSendTheShareNotificationByEmailUsingTheWebui() {
+		$errorMessage = "";
+		PHPUnit\Framework\Assert::assertNotNull(
+			$this->sharingDialog, "Sharing Dialog is not open"
+		);
+		try {
+			$this->sharingDialog->sendShareNotificationByEmail($this->getSession());
+		} catch (Exception $e) {
+			$errorMessage = $e->getMessage();
```
skshetry: Would be better to have the following exception inside catch

tests/acceptance/features/bootstrap/WebUISharingContext.php
```diff
@@ -776,6 +776,40 @@ public function removesPublicLinkAtCertainPosition($number, $entryName) {
...
+		}
+		PHPUnit\Framework\Assert::assertContains(
```
skshetry: Just throwing exception should work here, if above suggestion is accepted.

tests/acceptance/features/bootstrap/WebUISharingContext.php
paurakhsharma: If we move this inside catch block the code will still pass even if it was able to click the "notify by email" button which is not expected in this case.

tests/acceptance/features/bootstrap/WebUISharingContext.php
paurakhsharma: If we move this inside catch block and add new throw Exception here. It would work but it would be unnecessary in my opinion.

individual-it (changes requested):
what if a user was added to the group after the notification was send?
what if one user in the group has an email address send and an other not?

tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature
```diff
@@ -130,3 +130,77 @@ Feature: Sharing files and folders with internal groups
...
+  Scenario: user should not get an email notification if the user is added to the group after the mail notification was sent
+    Given parameter "shareapi_allow_mail_notification" of app "core" has been set to "yes"
+    And user "user0" has been created with default attributes
+    And user "user3" has logged in using the webUI
+    And user "user3" has shared file "lorem.txt" with group "grp1"
+    And the user has opened the share dialog for file "lorem.txt"
+    When the user sends the share notification by email using the webUI
+    Then a notification should be displayed on the webUI with the text "Email notification was sent!"
+    When the administrator adds user "user0" to group "grp1" using the provisioning API
+    Then the email address "user1@example.org" should not have received an email
```
individual-it: why `user1@example.org` not receive the email? what about `user0`?
Can the sharer resend the notification to the newly added users?

individual-it (changes requested):

tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature
paurakhsharma: > why `user1@example.org` not receive the email? what about `user0`?

It's a mistake. That should be `user1@example.org`

> Can the sharer resend the notification to the newly added users?

No, the sharer can not resend the notification after a new user is added to the group.

individual-it (approved):

paurakhsharma:
Backport on: #35235

paurakhsharma:
release-10.2.0: #35240

PR #35220, opened 2019-05-14
Treat send attempt as unsuccessful when there is a message.

VicDeo:
## Description
`result.ocs.meta.status` is 200 when the email was not sent.

## Related Issue
- Fixes #35218 

## Motivation and Context
Frontend says that email is  successfully sent even when it has been not sent.

## How Has This Been Tested?
See #35218 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35221, opened 2019-05-14
[10.2.0] Treat send attempt as unsuccessful when there is a message.

### LINKED ISSUE #35218: Sharer gets "Email notification was sent!" message even when the email was not sent
# Steps to reproduce
``` behat
Scenario 1: user should get an error message when trying to send notification by email to a user who has not setup their email
    Given parameter "shareapi_allow_mail_notification" of app "core" has been set to "yes"
    And user "user1" has logged in using the webUI
    And user "user1" has shared file "lorem.txt" with user "user2"
    And the user has opened the share dialog for file "lorem.txt"
    When the user sends the share notification by email using the webUI
    Then a notification should be displayed on the webUI with the text "Email notification was not sent!"
```

- Note: In the first scenario the email of the receiver is not set.

``` behat
Scenario 2: user should get an error message when trying to send notification when email server setting is invalid
    Given parameter "shareapi_allow_mail_notification" of app "core" has been set to "yes"
    And user "user1" has logged in using the webUI
    And user "user1" has shared file "lorem.txt" with user "user2"
    And the user has opened the share dialog for file "lorem.txt"
    When the user sends the share notification by email using the webUI
    Then a notification should be displayed on the webUI with the text "Email notification was not sent!"
```
- Note: In the second scenario set the receiver's email but change your Email Server setting in Admin General settings page to invalid one.

# Actual behaviour
User should get an error message saying "Email notification was sent!"

# Expected behaviour
User should get an error message saying "Email notification was not sent!"

c.c @individual-it @phil-davis @PVince81 

# Response in DevTools
`{"ocs":{"meta":{"status":"ok","statuscode":200,"message":"Couldn't send mail to following recipient(s): arkouser ","totalitems":"","itemsperpage":""},"data":[]}}`

VicDeo:
Backport of https://github.com/owncloud/core/pull/35220

## Description
`result.ocs.meta.status` is 200 when the email was not sent.

## Related Issue
- Fixes #35218 

## Motivation and Context
Frontend says that email is  successfully sent even when it has been not sent.

## How Has This Been Tested?
See #35218 

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -963,13 +963,17 @@ class Share20OcsController extends OCSController {
 			Share::setSendMailStatus($itemType, $itemSource, $shareType, $recipient, true);
 		}
 
-		$message = empty($result)
-			? null
-			: $this->l->t(
+		if (empty($result)) {
+			$message = null;
+			$data = ['status' => 'success'];
+		} else {
+			$message = $this->l->t(
 				"Couldn't send mail to following recipient(s): %s ",
 				\implode(', ', $result)
 			);
-		return new Result([], 200, $message);
+			$data = ['status' => 'error'];
+		}
+		return new Result($data, 200, $message);
 	}
 
 	/**
@@ -988,7 +992,7 @@ class Share20OcsController extends OCSController {
 	public function notifyRecipientsDisabled($itemSource, $itemType, $shareType, $recipient) {
 		// FIXME: migrate to a new share API
 		Share::setSendMailStatus($itemType, $itemSource, $shareType, $recipient, true);
-		return new Result();
+		return new Result(['status' => 'success']);
 	}
 
 	/**
--- a/core/js/sharedialogshareelistview.js
+++ b/core/js/sharedialogshareelistview.js
@@ -369,7 +369,7 @@
 			$loading.removeClass('hidden');
 
 			this.model.sendNotificationForShare(shareType, shareWith, true).then(function(result) {
-				if (result.ocs.meta.status === 'ok') {
+				if (result.ocs.data.status === 'success') {
 					OC.Notification.showTemporary(t('core', 'Email notification was sent!'));
 					$target.remove();
 				} else {
--- a/core/js/tests/specs/sharedialogshareelistview.js
+++ b/core/js/tests/specs/sharedialogshareelistview.js
@@ -221,7 +221,7 @@ describe('OC.Share.ShareDialogShareeListView', function () {
 			expect(notificationStub.called).toEqual(true);
 			notificationStub.restore();
 
-			deferred.resolve({ ocs: { meta: {status: 'ok' }}});
+			deferred.resolve({ ocs: { data : { status: 'success'},  meta: {message: null }}});
 			expect(notifStub.calledOnce).toEqual(true);
 			notifStub.restore();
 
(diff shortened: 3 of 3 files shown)


micbar:
@VicDeo Please do not merge until the decisions about the release are made.

paurakhsharma:
I have tested this and it works fine on the frontend.
But network response tab have two status
`{"ocs":{"meta":{"status":"ok","statuscode":200,"message":"Couldn't send mail to following recipient(s): merouser3 ","totalitems":"","itemsperpage":""},"data":{"status":"error"}}}
`
status "ok" and status "error".

Is this the expected behavior?

patrickjahns:
@paurakhsharma 
What was the behavior with 10.1 ? can you please double check

paurakhsharma:
> @paurakhsharma
> What was the behavior with 10.1 ? can you please double check

I double checked and this is the network response in 10.1 `{"data":{"message":"Couldn't send mail to following recipient(s): arkouser "},"status":"error"}`

PVince81:
@VicDeo would be good to also update the frontend unit tests. I believe we have some for the email part.

PVince81:
backend unit tests as well... but here I believe this is not coverable...

please always add statements and reasons if you decide to skip unit tests

PVince81:
> > @paurakhsharma
> > What was the behavior with 10.1 ? can you please double check
> 
> I double checked and this is the network response in 10.1 `{"data":{"message":"Couldn't send mail to following recipient(s): arkouser "},"status":"error"}`

@VicDeo so this changes the response format ? I thought this PR was an attempt to bring it back to what it was before

VicDeo:
@PVince81 to bring it back I need to kill the controller and send the response from `core/ajax/share.php`.
Frontend tests are adjusted already.

not coverable as I can't mock `Share::setSendMailStatus` properly

VicDeo:
10.1 sends a response  with `\OCP\JSON::success` or `\OCP\JSON::error` from **core/ajax/share.php**

https://github.com/owncloud/core/pull/33180 changes it send a response from `OCA\Files_Sharing\Controller\Share20OcsController` class via OCS

that's why the responses differ

PVince81:
@VicDeo would you qualify this as API change ? or would the new format be solidified as a public API if it's OCS ?

PVince81:
the old API as far as I remember was not OCS so considered private

PVince81:
@michaelstingl I also assume no client was using this private API in ajax/share.php

michaelstingl:
> @michaelstingl I also assume no client was using this private API in ajax/share.php

Is this the mail for public links? Never made it to the clients because of the missing proper API.

VicDeo:
@PVince81 it didn't look like a public API at all as it needed logged in user and a valid CSRF token:
https://github.com/owncloud/core/blob/master/core/ajax/share.php#L40-L41

VicDeo:
```
{
  "ocs":
  {
      "meta": 
      {
           "status":"ok",
           "statuscode":200,
           "message":"Couldn't send mail to following recipient(s): merouser3 ",
           "totalitems":"",
           "itemsperpage":""
      },
     "data":
     {
       "status":"error"
     }
   }
} 
```
`meta.status` is  HTTP-alike OCS status. **ok**  here is just a text test representation of a statuscode **200**
while `data.status` is exactly the same we had before (success/error)

PVince81 (approved):
:+1:

PR #35283, opened 2019-05-21
Accessing sharetab directly for a long list of files

### LINKED ISSUE #35200: browsing directly to details does not work for files far down the list
### Steps to reproduce
1. have a long list of file e.g. from https://github.com/owncloud/testing/tree/master/data/webUISkeleton
2. try to navigate directly to the sharing tab of last file e.g. `localhost/owncloud-core/index.php/apps/files/?dir=/&scrollto=zzzz-must-be-last-file-in-folder.txt&details=shareTabView`

### Expected behaviour
Sharing tab of the file should open

### Actual behaviour
image

### Server configuration
**Operating system**:
Ubuntu 18.04
**Web server:**
Apache2
**Database:**
SQLite
**PHP version:**
7.1
11.0.0
10.1.1
fresh

seems not to be a regression
bayernbox has a similar issue https://github.com/owncloud/enterprise/issues/3310#issuecomment-490865700

CC @patrickjahns

sharidas:
Accessing sharetab directly for a long list of
files caused problems. The pages need to be fetched
appropriately to get the data. This changeset fixes
the problem.

## Description

Lets say we have 30 files listed in the the UI. The maximum page size we have is 18. And hence when user tries to access the 19th file from the UI using `http://localhost/testing3/index.php/apps/files/?dir=/test&&scrollto=file19.txt&details=shareTabView` ( where `file19.txt` ) is the 19th file in the list, then the sharetab would be shown without any details. The reason for this is its the 19th element and hence the filelist doesn't have the `tr` added. So to solve this issue, the index of the file from the `this.files` is taken. And this would help to calculate the number of attempts required to fetch the page so that we get the data.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35200

## Motivation and Context

We need to fetch the pages to get the data populated, so that when they are accessed for the share tab field, the tab does not show empty.

## How Has This Been Tested?

-  Create a folder test
- Upload 150 files
- [x] Access 15th elements sharetab field
- [x] Access 110th file's share tab field
- [x] Access 150th files share tab field
- [x] Access 190th files share tab field
- [x] Access 210th files share tab field shows 
beyondarray

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files/js/filelist.js
+++ b/apps/files/js/filelist.js
@@ -2539,6 +2539,18 @@
 			}
 			if (!_.isUndefined(detailTabId)) {
 				var filename = file[file.length - 1];
+				//Double check if the area that you are scrolling is beyond the page limit?
+				var pageSize = this.pageSize();
+				var index = _.findIndex(this.files, function (obj) {
+					return obj.name === filename;
+				});
+				if (index >= pageSize) {
+					var numberOfMorePagesToScroll = Math.floor(index / pageSize);
+					while (numberOfMorePagesToScroll > 0) {
+						this._nextPage();
+						numberOfMorePagesToScroll--;
+					}
+				}
 				this.showDetailsView(filename, detailTabId);
 			}
 			this.highlightFiles(file, function($tr) {
--- a/tests/acceptance/features/webUIFiles/browseDirectlyToDetailsTab.feature
+++ b/tests/acceptance/features/webUIFiles/browseDirectlyToDetailsTab.feature
@@ -23,12 +23,11 @@ Feature: browse directly to details tab
 
   #merge this tests into previous scenario when bug is fixed
   @smokeTest
-  @issue-35200
   Scenario Outline: Browse directly to the sharing details of a file
     When the user tries to browse directly to display the "sharing" details of file "<file>" in folder "<folder>"
     #Then the thumbnail should be visible in the details panel
     Then the "sharing" details panel should be visible
-    And the share-with field should not be visible in the details panel
+    And the share-with field should be visible in the details panel
     #And the share-with field should be visible in the details panel
     Examples:
       | file                                 | folder        | comment   |
@@ -48,7 +47,6 @@ Feature: browse directly to details tab
       | lorem.txt         | simple-folder | a file somewhere in between |
 
   #merge this tests into previous scenario when bug is fixed
-  @issue-35200
   @comments-app-required
   Scenario Outline: Browse directly to the comments details of a file
     When the user tries to browse directly to display the "comments" details of file "<file>" in folder "<folder>"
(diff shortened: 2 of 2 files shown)


jvillafanez:
Some questions:
* Is there any element loaded before the scrollTo method is called? We don't really want to make unnecessary calls.
* How does this affect to "normal" scrolling? You can have 500 files inside a folder, you enter the folder normally and start scrolling down to see the last element. I expect the minimum number of calls to the `_nextPage()` method needed to do so. I want to avoid having to call 3 times that method to get the 3rd page, and then 4 times to get the next one, and then 5 times to get the 5th page.

There are minor style things we can adjust in the code, but I'd prefer to check those couple of things first in case there are changes needed.

sharidas:
> * Is there any element loaded before the scrollTo method is called? We don't really want to make unnecessary calls.

The `scrollTo` method is called from `setFiles` method which calls `_nextPage` method once. Thats the one which populates the data where the call reaches `updateDetailsView` method. Inside `updateDetailsView` method the call `this.findFileEl(fileName);` would return nothing. And hence call to `this.getModelForFile($tr);` would return null making the `model`, in the code path null.

> * How does this affect to "normal" scrolling? You can have 500 files inside a folder, you enter the folder normally and start scrolling down to see the last element. I expect the minimum number of calls to the `_nextPage()` method needed to do so. I want to avoid having to call 3 times that method to get the 3rd page, and then 4 times to get the next one, and then 5 times to get the 5th page.
> 

This does not affect normal scrolling. Because the code path is a little different, but eventually it calls `nextPage` method subsequntly while scrolling. So the containers scroll is binded to `onScroll` method ( in the same filelist.js file ). And this method calls nextPage method while user scrolls down. So lets say there are 100 files in a folder. The nextPage method from `onScroll` would be called 6 times.

Let me know if this addresses your query?

apps/files/js/filelist.js
```diff
@@ -2539,6 +2539,16 @@
 			}
 			if (!_.isUndefined(detailTabId)) {
 				var filename = file[file.length - 1];
+				//Double check if the area that you are scrolling is beyond the page limit?
+				var pageSize = this.pageSize();
+				var index = _.findIndex(this.files, function (obj) { return obj.name === filename});
```
jvillafanez:
```
var index = _.findIndex(this.files, function (obj) {
    return obj.name === filename;
});
```
This is more common.

apps/files/js/filelist.js
```diff
@@ -2539,6 +2539,16 @@
...
+				if (index > pageSize) {
+					for (var numberOfMorePagesToScroll = Math.floor(index / pageSize);
```
jvillafanez:
```
var numberOfMorePagesToScroll = Math.floor(index / pageSize);
while (numberOfMorePagesToScroll > 0) {
  this._nextPage();
  numberOfMorePagesToScroll--;
}
```
this feels a bit more readable.

In addition, "for" loops usually go upwards (from 0 to X) not downwards (from X to 0). The "while" loop also seems for readable in this case.

apps/files/js/filelist.js
sharidas: Updated the change.

jvillafanez (approved):

PR #35204, opened 2019-05-12
API acceptance tests for public link share permissions

dpakach:
## Description

Add api acceptance tests for public link share feature permissions.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Some of the public link permission features were changed in #35194. To test those features.

## How Has This Been Tested?

:robot: 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35295, opened 2019-05-22
get rid of "the share fields of the last share should include" step

individual-it:
## Description
1) saving the response of the share data after update  was added in #35204 but causes problems when an invalid request was done but the original `lastShareData` need to be accessed e.g in `\Sharing::getMimeTypeOfLastSharedFile()`
2) completely get rid of `Then the share fields of the last share should include`.
If we done the share creation request directly before the check we should use we use `Then the fields of the last response should include`.
If there are some steps in between the share creation and the check, we anyway should query the current state of the share by `When user "..." gets the info of the last share using the sharing API`, otherwise we only check the saved data and not the current state of the share

## Motivation and Context
fixing CI in various apps that were broken by #35204

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/updateShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/updateShare.feature
@@ -315,7 +315,8 @@ Feature: sharing
     And user "user0" has shared folder "/folder1" with user "user1" with permissions 31
     And user "user1" has shared folder "/folder1/folder2" with user "user2" with permissions 31
     When user "user1" moves folder "/folder1/folder2" to "/moved-out/folder2" using the WebDAV API
-    Then the share fields of the last share should include
+    And user "user1" gets the info of the last share using the sharing API
+    Then the fields of the last response should include
       | id                | A_NUMBER             |
       | item_type         | folder               |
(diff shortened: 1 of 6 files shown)


tests/acceptance/features/apiShareManagementBasic/updateShare.feature
```diff
@@ -315,7 +315,8 @@ Feature: sharing
     And user "user0" has shared folder "/folder1" with user "user1" with permissions 31
     And user "user1" has shared folder "/folder1/folder2" with user "user2" with permissions 31
     When user "user1" moves folder "/folder1/folder2" to "/moved-out/folder2" using the WebDAV API
-    Then the share fields of the last share should include
```
individual-it: this actually used to be :snake: :oil_drum: as `the share fields of the last share should include` would not check the current status but the status saved when creating the share

dpakach (approved):
Looks Good. Lets keep decreasing our code base :smile: !

individual-it:
backport in #35298

PR #35342, opened 2019-05-28
Add remove user to groups not case sensitive

phil-davis:
## Description
Add API and CLI acceptance tests to:
- add and delete users using different case of the username (UID)
- add and delete groups using different case of the group name, checking that distinct groups are being managed
- add and delete users in groups, using different case of user and group name, making sure that distinct groups are managed correctly, and the user names behave in the "opposite" way (they are not case-sensitive)
- when remembering the users that have been created during a test scenario, always remember the lowercase name, and index into the remembered array with the lowercase version - so that the acceptance test code happily understands that when `user0` has been created then `USER0` is deleted, that those were the same user.

## Related Issue
Part of issue #35354 

## Motivation and Context
User names (IDs) are not case-sensitive, so `user0`, `User0` and `USER0` are all the same user.
Group names are case-sensitive, so `group1`, `Group1` and `GROUP1` are 3 different groups that can have different members.

We should "document" this behaviour in acceptance tests so that we know if it accidentally changes.

## How Has This Been Tested?
Local runs of the acceptance test scenarios plus CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiProvisioning-v1/deleteUser.feature
+++ b/tests/acceptance/features/apiProvisioning-v1/deleteUser.feature
@@ -15,6 +15,13 @@ Feature: delete users
     And the HTTP status code should be "200"
     And user "brand-new-user" should not exist
 
+  Scenario: Delete a user, and specify the user name in different case
+    Given user "brand-new-user" has been created with default attributes
+    When the administrator deletes user "Brand-New-User" using the provisioning API
+    Then the OCS status code should be "100"
+    And the HTTP status code should be "200"
+    And user "brand-new-user" should not exist
+
   @smokeTest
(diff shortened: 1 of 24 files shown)


individual-it:
@HariBhandari07 please have a look if that makes sense to you

HariBhandari07:
looks good to me

paurakhsharma (approved):
Looks good to me :+1:

phil-davis:
Backport `stable10` #35343

individual-it (approved):

PR #35175, opened 2019-05-08
feature test for public link share expiration date deletion

haribhandari07:
## Description

Includes two tests for deletion of expiration date when the expiration date is not enforced and enforced.
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35368, opened 2019-05-30
[Stable10] feature test for public link share expiration date deletion

haribhandari07:
Backport for #35175


HariBhandari07:
This is a mistake. I merged into master instead of stable10.

PR #35403, opened 2019-06-03
Add batch user creation step for speeding up the requests

dpakach:
## Description

Add new step for user creation using batch requests to speed up the ci

## Related Issue

- Related issue https://github.com/owncloud/QA/issues/621

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/TestHelpers/HttpRequestHelper.php
+++ b/tests/TestHelpers/HttpRequestHelper.php
@@ -29,6 +29,8 @@ use GuzzleHttp\Message\ResponseInterface;
 use SimpleXMLElement;
 use Sabre\Xml\LibXMLException;
 use Sabre\Xml\Reader;
+use GuzzleHttp\BatchResults;
+use GuzzleHttp\Pool;
 
 /**
  * Helper for HTTP requests
@@ -68,6 +70,90 @@ class HttpRequestHelper {
 		if ($client === null) {
 			$client = new Client();
 		}
+		$request = self::createRequest(
+			$url,
+			$method,
+			$user,
+			$password,
+			$headers,
+			$body,
+			$config,
+			$cookies,
+			$stream,
+			$timeout,
+			$client
+		);
+
+		try {
+			$response = $client->send($request);
+		} catch (BadResponseException $ex) {
+			$response = $ex->getResponse();
+
+			//if the response was null for some reason do not return it but re-throw
+			if ($response === null) {
+				throw $ex;
+			}
+		}
+		return $response;
+	}
+
+	/**
+	 * Send the requests to the server in parallel.
+	 * This function takes an array of requests and an optional client.
+	 * It will send all the requests to the server using the Pool object in guzzle.
+	 *
+	 * @param array $requests
+	 * @param Client|null $client
+	 *
+	 * @return BatchResults
+	 */
+	public static function sendBatchRequest(
+		$requests,
+		$client = null
+	) {
+		if (!isset($client)) {
+			$client = new Client();
+		}
+		$results = Pool::batch($client, $requests);
+		return $results;
+	}
+
+	/**
+	 * Create an http request based on given parameters.
+	 * This creates an RequestInterface object before sending it to the server.
+	 * This also enables to create multiple requests in advance so that we can send them to the server at once in parallel.
+	 *
+	 * @param string $url
+	 * @param string $method
+	 * @param string $user
+	 * @param string $password
+	 * @param array $headers ['X-MyHeader' => 'value']
+	 * @param mixed $body
+	 * @param array $config
+	 * @param CookieJar $cookies
+	 * @param bool $stream Set to true to stream a response rather
+	 *                     than download it all up-front.
+	 * @param int $timeout
+	 * @param Client|null $client
+	 *
+	 * @return RequestInterface
+	 */
+	public static function createRequest(
+		$url,
+		$method = 'GET',
+		$user = null,
+		$password = null,
+		$headers = null,
+		$body = null,
+		$config = null,
+		$cookies = null,
+		$stream = false,
+		$timeout = 0,
+		$client = null
+	) {
+		if ($client === null) {
+			$client = new Client();
+		}
 
 		$options = [];
 		if ($user !== null) {
@@ -95,18 +181,7 @@ class HttpRequestHelper {
 				}
 			}
 		}
-
-		try {
-			$response = $client->send($request);
-		} catch (BadResponseException $ex) {
-			$response = $ex->getResponse();
-			
-			//if the response was null for some reason do not return it but re-throw
-			if ($response === null) {
-				throw $ex;
-			}
-		}
-		return $response;
+		return $request;
 	}
 
 	/**
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -25,6 +25,7 @@ use TestHelpers\OcsApiHelper;
 use TestHelpers\SetupHelper;
 use TestHelpers\UserHelper;
 use TestHelpers\HttpRequestHelper;
+use GuzzleHttp\Client;
 
 require __DIR__ . '/../../../../lib/composer/autoload.php';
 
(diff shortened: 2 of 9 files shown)


skshetry:
Can you make this run by default? Would be good to find what difference does this make.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -375,6 +376,118 @@ public function theseUsersHaveBeenCreated($setDefaultAttributes, $doNotInitializ
...
+	 *
+	 * @return void
+	 * @throws \Exception
+	 */
+	public function theseUsersHaveBeenCreatedUsingBatchAction($setDefaultAttributes, $doNotInitialize, TableNode $table) {
+		$table = $table->getColumnsHash();
+		$setDefaultAttributes = $setDefaultAttributes !== null;
+		$initialize = $doNotInitialize !== null;
+
+		// We add all the requests bodies in an array.
```
phil-davis:
```suggestion
		// We add all the request bodies in an array.
```

tests/TestHelpers/HttpRequestHelper.php
```diff
@@ -68,6 +70,90 @@ public static function sendRequest(
...
+			if ($response === null) {
+				throw $ex;
+			}
+		}
+		return $response;
+	}
+
+	/**
+	 * Send the requests to the server in parallel.
+	 * This function takes an array of requests and an option an client.
```
phil-davis: Do you mean:
```suggestion
	 * This function takes an array of requests and an optional client.
```

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -375,6 +376,118 @@ public function theseUsersHaveBeenCreated($setDefaultAttributes, $doNotInitializ
...
+		// These values cannot be set while creating the user, so we have to edit the newly created user to set these values.
+		$users = [];
+		$editData = [];
+		foreach ($bodies as $user) {
+			\array_push($users, $user['userid']);
+			$this->addUserToCreatedUsersList($user['userid'], $user['password']);
+			if (isset($user['displayName'])) {
+				\array_push($editData, ['user' => $user['userid'], 'key' => 'displayname', 'value' => $user['displayName']]);
+			}
+			if (isset($user['email'])) {
```
phil-davis: This sets the email in a separate request. IMO that is the way it has to be. The API does not allow setting both `displayName` and `email` in the same API call.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1193,6 +1306,44 @@ public function initializeUser($user, $password) {
...
+	 *
+	 * @return void
+	 * @throws \Exception
+	 */
+	public function initializeUserBatch($users) {
+		$url = "/cloud/users/%s";
+		$requests = [];
+		$client = new Client();
+		foreach ($users as $user) {
+			// create new request for each user but not send it yet.
```
phil-davis:
```suggestion
			// create a new request for each user but do not send it yet.
```

tests/acceptance/features/bootstrap/Provisioning.php
skshetry: API does not seem to support that, but, in theory, we can parallel-ly send requests, and as soon as a user gets created(i.e. we get a response among the batch of others requests), we could just throw other 2 requests for setting email and display name at the server. But, I don't think, it's good to invest time for that. The bottleneck currently is initializing users only.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -375,6 +376,118 @@ public function theseUsersHaveBeenCreated($setDefaultAttributes, $doNotInitializ
...
+		}
+
+		$results = HttpRequestHelper::sendBatchRequest($requests, $client);
+		// Retrieve all failures.
+		foreach ($results->getFailures() as $e) {
+			throw $e;
+		}
+
+		// Create requests for setting displayname and email for the newly created users.
+		// These values cannot be set while creating the user, so we have to edit the newly created user to set these values.
```
paurakhsharma: What if the user is not yet created in the server and the user edit request is sent to edit the information
of that user?

Is `user edit request` sent only after the all user creation is completed in the server and resolved?

dpakach: Accordint to guzzle documentation 
```
Sometimes you just want to send a few requests concurrently and then process the results all at 
once after they’ve been sent. Guzzle provides a convenience function GuzzleHttp\Pool::batch()
```
Which means the Pool::batch() will just send requests concurrently and wait for all of them to complete before processing them. Since we send user creation requests in first batch it should wait for all the user creation requests to complete before proceeding to the editing step.

tests/acceptance/features/bootstrap/Provisioning.php
dpakach: yes but we are already setting the displayname and email of the user in parallel. 
for example if there are 3 users then to set displayname and email for those 3 users we will need 6 requests. We are sending all of those 6 requests in parallel.

tests/acceptance/features/bootstrap/Provisioning.php
skshetry: It'd be too much detail, but the `Pool::batch()`, I think, only returns an array of promise (technically, a `BatchResults`). Only, when you go for checking contents/status of the response, it'll block till the response comes.

individual-it (approved):
please squash before merging

phil-davis (approved):

phil-davis:
There are various acceptance test CI fails, so I guess something in this code is not quite right.

individual-it (approved):

phil-davis (approved):

tests/TestHelpers/HttpRequestHelper.php
```diff
@@ -68,6 +70,90 @@ public static function sendRequest(
...
+	 * @param array $requests
+	 * @param Client|null $client
+	 *
+	 * @return BatchResults
+	 */
+	public static function sendBatchRequest(
+		$requests,
+		$client = null
+	) {
+		if (!isset($client)) {
```
skshetry: Only null check should be enough. `isset` checks if the variable is declared or not in addition to null-check. 

```suggestion
		if ($client !== null) {
```

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -331,47 +332,113 @@ public function theseUsersHaveBeenCreatedWithoutSkeletonFiles(TableNode $table)
...
+		}
+		// Edit the users in parallel to make the process faster.
+		if (\count($editData) > 0) {
+			$results = UserHelper::editUserBatch(
+				$this->getBaseUrl(),
+				$editData,
+				$this->getAdminUsername(),
+				$this->getAdminPassword()
+			);
+			foreach ($results->getFailures() as $e) {
```
skshetry: Should we also parallelize the `editUserBatch` and `initializeUserBatch`? I mean, we could, have this loop outside as well and gain a few speedups.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1193,6 +1260,44 @@ public function initializeUser($user, $password) {
...
+					$this->getPasswordForUser($user),
+					$method = 'GET',
+					\sprintf($url, $user),
+					[],
+					$client
+				)
+			);
+		}
+		// Send all the requests in parallel.
+		$response = HttpRequestHelper::sendBatchRequest($requests, $client);
```
skshetry: This is not really a response. Could get confusing.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1193,6 +1260,44 @@ public function initializeUser($user, $password) {
...
+		// throw an exception if any request fails.
+		foreach ($response->getFailures() as $e) {
```
skshetry: This blocks, right? If it does, we have, for example, `editUserBatch` which does not block at all. But, this blocks itself. 

Can we somehow make it clearer that a given function will block or not? Maybe, by changing function name? What do you think?

dpakach:
Backport on #35522

phil-davis:
@dpakach tomorrow please talk with @skshetry about his recent comments. Then if there is good refactoring to do, make another PR for that.

dpakach:
> @dpakach tomorrow please talk with @skshetry about his recent comments. Then if there is good refactoring to do, make another PR for that.

@phil-davis I have already started refactoring. It's almost over. I couldnt do PR today because it was late.

PR #35420, opened 2019-06-04
[DO NOT MERGE] Smb windows temp

phil-davis:
See if the `fsweb` `smb_windows` is actually available at `172.16.115.3`


phil-davis:
This is fixed - see https://github.com/owncloud/core/pull/35426#issuecomment-508426936

PR #35084, opened 2019-04-23
Convert shareWith to lower case

sharidas:
Convert shareWith to lower case when share is created.
This would help to make it consistent across oC.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Convert `shareWith` to lower case. The share API does not convert it to lower case, and as a result, it would cause inconsistency: if the API is used by passing `shareWith` with a mix of upper and lower case. The share would be inaccessible. This was found with guest user.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Convert `shareWith` to lower case in the Share20OcsController::createShare(). If its not converted the share might be inaccessible to the user. This was found with guest user.

## How Has This Been Tested?

- Create guest user
- Set the guest user password by accessing the link from email
- The guest share is accessible from the webUI by logging in as the guest user.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

Issue #35382, opened 2019-05-31
File is not shared with user whose username contains capital letter


HariBhandari07:
### Steps to reproduce
1. Create a user with upper case username (eg. User1)
2. Share a file(eg somefile.txt) with User1 using webui
3. Relogin as User1
4. The shared file (somefile.txt) is not listed in the files page and shared with you page

### Expected behaviour
file should be shared with user with uppercase username
### Actual behaviour
file is not shared with user with uppercase username

### Server configuration
**Operating system**:
Ubuntu 18.04.2 LTS
**Web server:**
Apache
**Database:**
SQLite
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
11.0.0.0
**Updated from an older ownCloud or fresh install:**
fresh install
**Where did you install ownCloud from:**
git

**List of activated apps:**

```
  - comments: 0.3.0
  - dav: 0.4.0
  - encryption: 1.3.1
  - federatedfilesharing: 0.5.0
  - federation: 0.1.0
  - files: 1.5.2
  - files_external: 0.7.1
  - files_sharing: 0.12.0
  - files_texteditor: 2.3.0
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - guests: 0.8.1
  - market: 0.4.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - testing: 0.1.0
  - updatenotification: 0.2.1
  - user_management: 0.0.1

```

#### ownCloud log (data/owncloud.log)
```
{"reqId":"IZmafUYSKehiSWdPKnZr","level":3,"time":"2019-05-31T04:53:57+00:00","remoteAddr":"::1","user":"admin","app":"no app in context","method":"POST","url":"\/core\/index.php\/apps\/user_management\/users","message":"User backend OC\\User\\Database is returning home: \/home\/haribhandari\/www\/core\/data\/User1 for user: User1 which differs from existing value: "}
{"reqId":"FHbiNrBQKTnBrPVjtGaf","level":2,"time":"2019-05-31T04:54:03+00:00","remoteAddr":"::1","user":"admin","app":"no app in context","method":"GET","url":"\/core\/remote.php\/dav\/avatars\/admin\/31.jpeg","message":""}
```

phil-davis:
The same issue happens whatever the display name of the user. e.g. create `User1` with display-name `John Smith`
```
php occ user:add --display-name "John Smith" User1
```

[...]

PR #35448, opened 2019-06-06
Revert "Convert shareWith to lower case"

micbar:
Reverts owncloud/core#35084

Revert of `stable10` can be found in PR #35449


sharidas (approved):

phil-davis:
Also needs adjustment of related acceptance test(s), because they no longer pass.
I will look...

phil-davis:
I pushed 2 acceptance test commits:
- "API share_with user response is now case sensitive" - after reverting this code, when somebody uses the sharing API to share with another user and they specify the user name (UID) in mixed or upper case, then the `share_with` field in the success response has the user name in the mixed or upper case.
- "Issue 35382 is resolved - webUI sharing to uppercase username works" - after reverting this code, issue #35382 is fixed, so the test scenario is adjusted to expect the correct successful behaviour.

PR #35454, opened 2019-06-07
Fixes orthographic mistake in error message

tabesin:
This fixes a simple orthographic mistake in an error message affecting German speaking users.


patrickjahns:
@DeepDiver1975 
Would this work - or do we need to push this to transifex somehow?

DeepDiver1975:
I fixed this on transifex - THX @tabesin

PR #35397, opened 2019-05-31
[stable10] use a multi line approach to display share autocomplete

DeepDiver1975:
## Related Issue
- Fixes #35098 

## How Has This Been Tested?
- :hand: 

### Federated or Guest
Screenshot from 2019-05-31 17-26-25

### Group
Screenshot from 2019-05-31 17-26-00

### With additional user information
Screenshot from 2019-05-31 17-25-46

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35486, opened 2019-06-11
use a multi line approach to display share autocomplete

karakayasemi:
## Description
Port of #35397 
Enhance sharing dialog UI to prevent overflow.

## Related Issue
- Fixes #35098 

## How Has This Been Tested?
- :hand: 

### Federated or Guest
Screenshot from 2019-05-31 17-26-25

### Group
Screenshot from 2019-05-31 17-26-00

### With additional user information
Screenshot from 2019-05-31 17-25-46

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #35487, opened 2019-06-11
[master] use a multi line approach to display share autocomplete

phil-davis:
Forward port of PR #35397


phil-davis:
See #35486 
Two of us did the forward port at the same time.

PR #35362, opened 2019-05-29
Log exceptions from exceptionHandler method

sharidas:
Log exceptions from console.php

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Log exceptions from console.php. As of now the exceptions are thrown in the console and then the script is exited. Before the script exits, it would be better to log the exception. This change set achieves the goal.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Log the exception that happens in the `exceptionHandler()` in console.php

## How Has This Been Tested?

- I updated the changes as shown below
   - Before line `require_once 'public/Constants.php';` in lib/base.php add `hello`
   - Now run `./occ a:l`, you should see exception:
```
Cannot load Xdebug - it was already loaded
System is broken.
ParseError: syntax error, unexpected 'require_once' (T_REQUIRE_ONCE) in /home/sujith/test/owncloud3/lib/base.php:61
Stack trace:
#0 /home/sujith/test/owncloud3/occ(11): require_once()
#1 {main}%
```
   - Now remove `hello` and add `throw new \Exception('foo');` and re-run the command `./occ a:l`, you should see message like this in the console.
```
System is broken.
Exception: foo in /home/sujith/test/owncloud3/lib/base.php:60
Stack trace:
#0 /home/sujith/test/owncloud3/console.php(65): require_once()
#1 /home/sujith/test/owncloud3/occ(11): require_once('/home/sujith/te...')
#2 {main}%
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35543, opened 2019-06-14
[stable10] Backport of Log exceptions from exceptionHandler method

sharidas:
## Description

Log the exceptions thrown by occ command and also echo the exceptions into the console. This solves the problem when the command is executed as cron job and due to some reason if exception happens in the command, the data is logged in to the owncloud logger. So its helpful for the admin to look for the exception due to which the command was exited. Also we set the `catchExceptions` of symfony's console application to `false`. This helps to catch the exception in owncloud's console.php. 

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Log the exception which is caused by the occ commands.

## How Has This Been Tested?

- The test at https://github.com/owncloud/core/pull/35362#issuecomment-500849944 applies here too.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/console.php
+++ b/console.php
@@ -58,54 +58,68 @@ if (\strtoupper(\substr(PHP_OS, 0, 3)) === 'WIN') {
 function exceptionHandler($exception) {
 	echo "An unhandled exception has been thrown:" . PHP_EOL;
 	echo $exception;
+	\OC::$server->getLogger()->logException($exception);
 	exit(1);
 }
 try {
 	require_once __DIR__ . '/lib/base.php';
-
-	// set to run indefinitely if needed
-	\set_time_limit(0);
-
-	if (!OC::$CLI) {
-		echo "This script can be run from the command line only" . PHP_EOL;
-		exit(1);
-	}
-
-	\set_exception_handler('exceptionHandler');
-
-	if (!\function_exists('posix_getuid')) {
-		echo "The posix extensions are required - see http://php.net/manual/en/book.posix.php" . PHP_EOL;
-		exit(1);
-	}
-	$user = \posix_getpwuid(\posix_getuid());
-	$configUser = \posix_getpwuid(\fileowner(OC::$configDir . 'config.php'));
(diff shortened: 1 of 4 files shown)


sharidas:
Original PR: https://github.com/owncloud/core/pull/35362

micbar (approved):

micbar:
@sharidas @phil-davis 
We need to revert this. It does not meet the expected behaviuor.

Expected behaviour was: Don't change the occ error messages, but log exception in owncloud.log

This doesn't seem possible.

@phil-davis Do we need to consider something related to tests?

phil-davis:
From what I see, the command-line behaviour change that is "strange" is that the error messages come in `stdout`  rather than `stderr`. For a real person who is just typing commands and looking at the output, there is not really a difference. But if people have scripts that do `occ` commands and parse `stderr` for error message text, then this will break a script like that.

So it would be nice if we could get the `stderr` output back.

sharidas:
@phil-davis I have created revert PR https://github.com/owncloud/core/pull/35873. This should bring back the change which you have mentioned at https://github.com/owncloud/core/pull/35543#issuecomment-512689624

phil-davis:
> @phil-davis I have created revert PR #35873. This should bring back the change which you have mentioned at [#35543 (comment)](https://github.com/owncloud/core/pull/35543#issuecomment-512689624)

That is the easy way to get back to the old behaviour. But is there a way to get that old `stderr`  behaviour and also get the exception logging of this PR?

sharidas:
> That is the easy way to get back to the old behaviour. But is there a way to get that old `stderr` behaviour and also get the exception logging of this PR?

I doubt we can acheive both. Because the old stderr exceptions are handled by symfony and then I couldn't find a way where symfony throws back the exception to oC.

micbar:
@phil-davis We come to the conclusion, that the logging needs to be done in the Command Classes themselves.

PR #35544, opened 2019-06-14
[stable10] Return "password fields" only if public-link password is set

IljaN:
## Description
Show "share_with" and "share_with_displayname" fields for public-links only if a share-password
is set.

## Related Issue

- Fixes #35541
- Caused by https://github.com/owncloud/core/pull/34691 

## Motivation and Context
The iOS client uses the presence of these fields to determine if a password is required. Despite their names they are historically used to store pw-hashes for public-links.

Due to a recent security fix (https://github.com/owncloud/core/commit/31b47dc973ddf48bb9d5a9fd7ff30db89e33c164) which replaced the hashes with ```***redacted***```, share_with and share_with_displayname was never null thus for the client it looked like all links have passwords.

## How Has This Been Tested?
Adjusted acceptance tests

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport https://github.com/owncloud/core/pull/35545

Issue #35541, opened 2019-06-14
Share20OcsController no longer provides info whether public link share is password-protected


felix-schwarz:
### Steps to reproduce
1. Retrieve list of shares 

### Expected behaviour
For public link shares with a password, the record included `share_with` and `share_with_displayname` with values, looking like this:
```xml
  <element>
   <id>29</id>
   <share_type>3</share_type>
   <uid_owner>admin</uid_owner>
   <displayname_owner>admin</displayname_owner>
   <permissions>1</permissions>
…
   <share_with>some value</share_with>
   <share_with_displayname>some value</share_with_displayname>
…
   <name>Public link</name>
  </element>
```

For public link shares without password, the `share_with` and `share_with_displayname` records were not included:
```xml
  <element>
   <id>29</id>
   <share_type>3</share_type>
   <uid_owner>admin</uid_owner>
   <displayname_owner>admin</displayname_owner>
   <permissions>1</permissions>
…
   <name>Public link</name>
  </element>
```

### Actual behaviour
The `share_with` and `share_with_displayname` tags are now always included for **all** public links, regardless of whether they have a password set:

```xml
  <element>
   <id>29</id>
   <share_type>3</share_type>
   <uid_owner>admin</uid_owner>
   <displayname_owner>admin</displayname_owner>
   <permissions>1</permissions>
…
   <share_with>***redacted***</share_with>
   <share_with_displayname>***redacted***</share_with_displayname>
…
   <name>Public link</name>
  </element>
```

### Possible solutions
I believe [this code segment](https://github.com/owncloud/core/blob/13ef7f969126fc1ba3a5c368babf4560646a7445/apps/files_sharing/lib/Controller/Share20OcsController.php#L204) to be responsible for this:
```php
		} elseif ($share->getShareType() === Share::SHARE_TYPE_LINK) {
			$result['share_with'] = '***redacted***';
			$result['share_with_displayname'] = '***redacted***';
			$result['name'] = $share->getName();
```

If it only returned the two tags `***redacted***` if the share has a password set, it'd match past behaviour. Speculatively, the code would look something like this then:
```php
		} elseif ($share->getShareType() === Share::SHARE_TYPE_LINK) {
			if ($share->getSharedWith()) {
				$result['share_with'] = '***redacted***';
				$result['share_with_displayname'] = '***redacted***';
			}
			$result['name'] = $share->getName();
```

### Impact
[...]

PR #35545, opened 2019-06-14
[master] Return "password fields" only if public-link password is set.

IljaN:
Issue:  #35541
Forward port of: https://github.com/owncloud/core/pull/35544


patrickjahns (approved):
👍

PR #35549, opened 2019-06-17
Fix method to create users without skeleton files in acceptance tests

dpakach:
## Description

Fix the method to create users without skeleton files in acceptance tests to use parallel requests.

## Related Issue

- Related Issue https://github.com/owncloud/QA/issues/621

## Motivation and Context

This method used loop to call user creation request which was slow compared to parallel requests.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -317,9 +317,7 @@ trait Provisioning {
 	public function theseUsersHaveBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles(TableNode $table) {
 		$path = $this->popSkeletonDirectoryConfig();
 		try {
-			foreach ($table as $row) {
-				$this->userHasBeenCreatedWithDefaultAttributes($row['username']);
-			}
+			$this->theseUsersHaveBeenCreated("default attributes and", "", $table);
 		} finally {
 			// restore skeletondirectory even if user creation failed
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -317,9 +317,7 @@ public function userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles($
 	public function theseUsersHaveBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles(TableNode $table) {
 		$path = $this->popSkeletonDirectoryConfig();
 		try {
-			foreach ($table as $row) {
-				$this->userHasBeenCreatedWithDefaultAttributes($row['username']);
-			}
+			$this->theseUsersHaveBeenCreated("default attributes and", "", $table);
```
phil-davis: Note: passing any non-blank string in the first parameter is enough to cause default attributes to be applied.
But passing the exact literal text from the step-definition regex is safe.
If someone changes the regex text some day, then this is a dependency here that also should be updated.

phil-davis (approved):
Looks fine.
Ideally there would be some parameter interface to some underlying method for `theseUsersHaveBeenCreated` that uses booleans. Then we would not have to rely on passing "special" strings.

dpakach:
Backport for #35552

PR #34960, opened 2019-04-03
Fix public link share, default expiration behavior

karakayasemi:
## Description
- "default link expiration" sub-entry location problem fixed. Now, it is under "Set expiration date".

- Currently, when the expiration date enabled and not enforced,  the default expiration date is set even though the user deleted it on share creation. With this PR, if the user removes the expiration date on share creation, no expiration date will be set as discussed in here: https://github.com/owncloud/enterprise/issues/3204#issuecomment-478657949

## Related Issue
https://github.com/owncloud/enterprise/issues/3204

## Motivation and Context
Fixing bugs.

## How Has This Been Tested?
Unit tests are adjusted. Also, manually tested with following steps:
- Activate, "Set expiration date" from admin sharing settings. The expiration date sub-entry should be listed under "Set expiration date" input.

To test share creation behavior;
- Activate, "Set expiration date" but not enforce.
- Click share a file via public link and remove default expiration date from its input field. The share should be created without an expiration date.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35564, opened 2019-06-17
set default expiration date if expiration date is enforced and it is new share

### LINKED ISSUE #35550: Public link share creation no longer automatically sets expiration date
Probably relates to https://github.com/owncloud/core/pull/34971

Previously on creation of a public link share, if "Set default expiration date" was enabled and no expire data parameter was set on the share creation API request, the server would automatically set the default expiration date on share creation.

Now, it does not do that anymore and instead replies "Expiration date is enforced".

This breaks how the desktop client used the API. In particular it will make it impossible to create any public link shares on desktop clients before 2.6.0 if "Enforce expiration date" is also enabled.

See https://github.com/owncloud/client/issues/7245

Possibly the link share creation API could distinguish between
- requesting "no expiry" (to resolve the problem PR #34971 was addressing)
- requesting "expire at that date"
- not having the "expireDate" parameter at all (restoring the previous behavior for this case)

karakayasemi:
## Description
Automatic expiration date setting behavior changed with #34971. It caused regression on client sides. This PR fixes this regression by keeping #34971's functionality. 

If max expiration date is enforced set default expiration date for new share. @ckamm please check the behavior is same as expected or not. 

@micbar @pmaier1  we should consider this PR for v10.2.1.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/35550

## Motivation and Context
Automatic expiration date setting behavior changed with #34971. It caused regression on client sides. This PR fixes this regression by keeping #34971's functionality. 

## How Has This Been Tested?.
To test issue #35550 
- Activate, "Set expiration date" and enforce max expiration date.
- Try to create new public link from a desktop client
- It should create link and fill expiration date with default.

To test issue #34960
- Activate, "Set expiration date" but not enforce.
- Click share a file via public link and remove default expiration date from its input field. The share should be created without an expiration date.

## Types of changes
- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -331,6 +331,13 @@ class Manager implements IManager {
 
 		// If we enforce the expiration date check that is does not exceed
 		if ($this->shareApiLinkDefaultExpireDateEnforced()) {
+			// If expiredate is empty and it is a new share, set a default one if there is a default
+			if ($this->isNewShare($share) && $expirationDate === null && $this->shareApiLinkDefaultExpireDate()) {
+				$expirationDate = new \DateTime();
+				$expirationDate->setTime(0, 0, 0);
+				$expirationDate->add(new \DateInterval('P'.$this->shareApiLinkDefaultExpireDays().'D'));
+			}
+
 			if ($expirationDate === null) {
 				throw new \InvalidArgumentException('Expiration date is enforced');
 			}
@@ -1548,5 +1555,19 @@ class Manager implements IManager {
 		}
 
 		return \md5(\json_encode($perms->toArray()));
+	}
+
+	/**
+	 * @param IShare $share
+	 * @return boolean
+	 */
+	private function isNewShare(IShare $share) {
+		$fullId = null;
+		try {
+			$fullId = $share->getFullId();
+		} catch (\UnexpectedValueException $e) {
+			// This is a new share
+		}
+		return ($fullId === null);
 	}
 }
--- a/tests/lib/Share20/ManagerTest.php
+++ b/tests/lib/Share20/ManagerTest.php
@@ -67,7 +67,7 @@ class ManagerTest extends \Test\TestCase {
 	protected $manager;
 	/** @var ILogger | \PHPUnit\Framework\MockObject\MockObject */
 	protected $logger;
-	/** @var IConfig */
+	/** @var IConfig | \PHPUnit\Framework\MockObject\MockObject */
 	protected $config;
 	/** @var ISecureRandom */
 	protected $secureRandom;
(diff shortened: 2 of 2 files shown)


micbar:
@karakayasemi Great!
Please force push, random CI failure

jvillafanez:
It feels better if this is added in the `createShare` method unless this is also somehow applied when a user updates the share.

karakayasemi:
> 
> 
> It feels better if this is added in the `createShare` method unless this is also somehow applied when a user updates the share.

Both `createShare` and `updateShare` method are using `validateExpirationDate` method. 
If expirationDate null, for a new share `validateExpirationDate` method sets default date with this PR (expected behavior for clients), but for an update case, it throws `InvalidArgumentException`.

ckamm:
@karakayasemi Excellent, I'll test it now.

ckamm:
Tests:
1) Default + enforced expiration date
- Web: Expiration date is pre-filled. Removing it leads to error. (good)
- Desktop 2.6: Expiration date is pre-filled. Removing is impossible. (good)
- Desktop 2.5: Creating share works, has expiration date. Removing is impossible. (good)

2) Default, but not enforced expiration date
- Web: Expiration date is pre-filled. Removing it before creation leads to share **with** expiration. (bug)
- Desktop 2.6: Expiration date is **not** pre-filled (client side bug). Removing it before creation leads to share **with** expiration. (bug)
- Desktop 2.5: Creating share works, has expiration date. Can be removed. (good)

What's the API for "create share without expiration" vs "create share with default expiration"?

ckamm (changes requested):
Test issues, see comment

karakayasemi:
@ckamm thanks for detailed explanation. I moved the code-block which set expiration date inside `if($this->shareApiLinkDefaultExpireDateEnforced())` condition.  
I tested scenarios with webUI and client 2.5.4, everything seems okay this time. Please you also test one more time. Thanks.

> What's the API for "create share without expiration" vs "create share with default expiration"?

They both uses same OCS API and following same code flow, no difference between them.

lib/private/Share20/Manager.php
```diff
@@ -331,6 +331,20 @@ protected function validateExpirationDate(\OCP\Share\IShare $share) {
 
 		// If we enforce the expiration date check that is does not exceed
 		if ($this->shareApiLinkDefaultExpireDateEnforced()) {
+			// If expiredate is empty set a default one if there is a default
+			$fullId = null;
+			try {
```
jvillafanez: It might be better if this check is moved to a private function `isNewShare` or similar.

karakayasemi: @jvillafanez moved to a new function. Thanks.

jvillafanez:
code looks fine, so if it behaves fine I think it's good to merge

patrickjahns:
> I'm pretty sure one can only get everything right by keeping the old "null means default" behavior intact and making the new "request share without expiration" have different API.

Thanks for highlighting this - was this change introduced in 10.2.0 or before?

ckamm (approved):

patrickjahns (approved):

phil-davis:
Backports
`stable10` #35592 
`release-10.2.1` #35593

PR #35580, opened 2019-06-18
Add cli tests for app upgrade of major and minor release

paurakhsharma:
## Description
Test `major` and `minor` app release upgrade. 

This tests can only be run in a local test system because when there needs to be a minor or major
upgrade of the app `./occ upgrade` needs to be executed which in turn disables `testing app`
and switches to `maintenance` mode and hence remote test calls cannot be executed.

## Related Issue
https://github.com/owncloud/QA/issues/613

## Motivation and Context
- Make CI faster in scenarios where there is no need for user files.

## How Has This Been Tested?
🤖

## Types of changes

- [ ] Technical debt
- [x] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


paurakhsharma:
Some problem in `drone.yml` will fix it tomorrow.

.drone.yml
```diff
@@ -1083,6 +1094,18 @@ matrix:
       OWNCLOUD_LOG: true
       INSTALL_TESTING_APP: true
 
+    # CLI AppManagement Acceptance tests
+      - PHP_VERSION: 7.1
```
phil-davis: It probably does not like the extra indent here

paurakhsharma: Yeah, this might be the cause. Thank you :)

paurakhsharma:
I am not sure about using `local-test-acceptance-cli` method. Please let me know your thought.

.drone.yml
```diff
@@ -402,6 +402,17 @@ pipeline:
       matrix:
         TEST_SUITE: cli
 
+  local-cli-acceptance-tests:
+    image: owncloudci/php:${PHP_VERSION}
+    pull: true
+    environment:
+      - TEST_SERVER_URL=${SERVER_PROTOCOL}://server-${SERVER_PROTOCOL}
+    commands:
+      - make local-test-acceptance-cli
```
phil-davis:
```suggestion
      - make test-acceptance-local-cli
```
All the acceptance test make targets should start with `test-acceptance` so that they are consistent and you can auto-complete nicely when typing.

IMO `local-cli` is an OK name. Others might have other ideas?

tests/acceptance/features/bootstrap/OccContext.php
```diff
@@ -114,6 +114,19 @@ public function invokingTheCommandWithEnvVariable(
 		);
 	}
 
+	/**
+	 * @When the administrator runs upgrade routines on local server using the occ command
+	 *
+	 * @return void
+	 */
+	public function theAdministratorRunsUpgradeRoutinesOnLocalServerUsingTheOccCommand() {
+		\system("sudo -u www-data ./occ upgrade", $status);
```
skshetry: Use of `sudo` is bad, but don't know any alternatives. But, we previously used to call local occ command. How did we do it previously, @phil-davis ?

.drone.yml
skshetry: I'd opt for using `TESTING_EXTERNAL` env variable instead of creating targets from makefile, and set that env variable and use tag (eg: `@local-cli`) and skip-tag (`skipWhenTestingExternal` or something similar).

tests/acceptance/features/bootstrap/OccContext.php
skshetry: This is how we used to do that: https://github.com/owncloud/core/blob/dac4bf080022d009254ec31a3c6fc8930ab1a0f2/tests/acceptance/features/bootstrap/CommandLine.php#L41-L72

.drone.yml
paurakhsharma: > I'd opt for using TESTING_EXTERNAL env variable instead of creating targets from makefile

For that we have to remove `--remote` flag from test start script in our `Makefile`.

And I think think that is a good way to do this.
What do you think @phil-davis?

tests/acceptance/features/bootstrap/OccContext.php
paurakhsharma: Yes we want both `php-dev` server users and `apache` users to be able to use this.

paurakhsharma:
weird that only two scenarios are passing in drone. Locally all scenarios are passing. 
Looks like last 2 scenarios are failing because there is some problem running `occ` command
in the `AfterScenario` of the second scenario.

.drone.yml
```diff
@@ -394,6 +394,7 @@ pipeline:
     environment:
       - TEST_SERVER_URL=${SERVER_PROTOCOL}://server-${SERVER_PROTOCOL}
       - MAILHOG_HOST=email
+      - TESTING_REMOTE_SYSTEM=${TESTING_REMOTE_SYSTEM:=true}
```
individual-it: I don't think we should need that,  `run-sh` still can do all its actions as on a remote system, should not matter because at that time the testing app is active and functional

individual-it (changes requested):
looks generally good, I believe we could do it even simpler

.drone.yml
paurakhsharma: Yes, that's right I had added this for the previous solution. 
This is not required now.

paurakhsharma:
@skshetry @phil-davis  please review again.

phil-davis (approved):
Looks good now.
Should we also have tests that put the app in combinations of `apps` and `apps-external` ?
(those could come in a next PR)

individual-it (approved):

paurakhsharma:
no code cov 😢 @DeepDiver1975 please merge this.

phil-davis:
No CLA either ???

phil-davis:
I rebased. Now the CLA has passed - must have been some GitHub problem the other day.
Let's see if we get drone and codecov results.

paurakhsharma:
Backport on: #35729

PR #35585, opened 2019-06-18
Do not allow to set higher permissions on a public link share for a r…

DeepDiver1975:
…esource which was shared with limited permissions

## Description
As the acceptance scenario shows a public link is not allowed to have more permissions then the original share.

## Related Issue
- refs owncloud/enterprise#3364

## How Has This Been Tested?
- running acceptance tests

## Types of changes
- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -825,6 +825,14 @@ class Share20OcsController extends OCSController {
 			}
 
 			if ($newPermissions !== null) {
+				$shareHome = $this->rootFolder->getUserFolder($share->getSharedBy());
+				$nodes = $shareHome->getById($share->getNode()->getId());
+				foreach ($nodes as $node) {
+					if (($node->getPermissions() | $newPermissions) !== $node->getPermissions()) {
+						return new Result(null, 404, 'Cannot increase permission of ' . $share->getTarget());
+					}
+				}
+
 				$share->setPermissions($newPermissions);
 				$permissions = $newPermissions;
 			}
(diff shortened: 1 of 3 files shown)


apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -825,6 +825,14 @@ public function updateShare($id) {
 			}
 
 			if ($newPermissions !== null) {
+				$shareHome = $this->rootFolder->getUserFolder($share->getSharedBy());
+				$nodes = $shareHome->getById($share->getNode()->getId());
+				foreach ($nodes as $node) {
+					if (($node->getPermissions() | $newPermissions) !== $node->getPermissions()) {
+						return new Result(null, 404, 'Cannot increase permission of ' . $share->getTarget());
```
IljaN: Can we use 403 instead 404 here?

phil-davis: Lots of the scenarios in `reShare.feature` test sharing requests that are not allowed. The code currently returns 404 for all those. I have no idea why. IMO "403 - forbidden" seems the more appropriate code. But if it is changed here, then it should be changed for all the other permissions check cases, and that is a bigger change which could break clients that have specifically coded to know about 404.

DeepDiver1975: 403 looks more reasonable - yes

In order to stay consistent with all other error cases of the same kind I prefer to stick with 404

patrickjahns: Agreed - let's stay consistent with 404 for now.
We can create a separate PR with api changes for upcoming releases that move theses error scenarios into a 403 if desired

IljaN (dismissed):

phil-davis (commented):
Adjusted acceptance test is good, it passes and all other CI passes. So it must be good ;)
Someone else can review-approve the technical implementation code.

IljaN (approved):

phil-davis:
Backports:
`stable10` #35599 
`release-10.2.1` #35600

PR #35594, opened 2019-06-19
Test regression with public link not created when default expiration date set and enforced

### LINKED ISSUE #35550: Public link share creation no longer automatically sets expiration date

skshetry:
## Description

This adds a regression test that public link was not getting created when default expiration date was set and enforced. This also tests the functionality of #34960.

## Related Issue

- Fixes #35550 

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -975,3 +975,20 @@ Feature: sharing
       | 2              | Case-Sensitive-Group | CASE-SENSITIVE-GROUP | case-sensitive-group | 200             | 404              |
       | 2              | CASE-SENSITIVE-GROUP | case-sensitive-group | Case-Sensitive-Group | 200             | 404              |
 
+  @issue-35550
+  Scenario Outline: Create a public link with default expiration date set and max expiration date enforced
+    Given using OCS API version "<ocs_api_version>"
+    And parameter "shareapi_default_expire_date" of app "core" has been set to "yes"
+    And parameter "shareapi_enforce_expire_date" of app "core" has been set to "yes"
+    When user "user0" creates a public link share using the sharing API with settings
+      | path | textfile0.txt |
+    Then the OCS status code should be "<ocs_status_code>"
+    And the OCS status message should be "Expiration date is enforced"
+    And the HTTP status code should be "<http_status_code>"
+#    And the last public shared file should be able to be downloaded without a password
+    Examples:
+      | ocs_api_version | ocs_status_code | http_status_code |
+      | 1               | 403             | 200              |
+#      | 1               | 100             | 100              |
+      | 2               | 403             | 403              |
+#      | 2               | 200             | 200              |
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -975,7 +975,7 @@ Feature: sharing
       | 2              | Case-Sensitive-Group | CASE-SENSITIVE-GROUP | case-sensitive-group | 200             | 404              |
       | 2              | CASE-SENSITIVE-GROUP | case-sensitive-group | Case-Sensitive-Group | 200             | 404              |
 
-  @issue-35550
+  @public_link_share-feature-required
   Scenario Outline: Create a public link with default expiration date set and max expiration date enforced
     Given using OCS API version "<ocs_api_version>"
     And parameter "shareapi_default_expire_date" of app "core" has been set to "yes"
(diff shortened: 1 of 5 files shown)


skshetry:
I am currently testing this without the fix in #35564. I'll just remove comments when the tests fail.

micbar:
@skshetry Fix got merged

tests/acceptance/features/apiShareManagementBasic/createShare.feature
```diff
@@ -845,4 +845,22 @@ Feature: sharing
     Examples:
       | ocs_api_version |
       | 1               |
-      | 2               |
\ No newline at end of file
+      | 2               |
+
+  @issue-35550
```
dpakach:
```suggestion
  @issue-35550 @public_link_share-feature-required
```

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1145,6 +1170,7 @@ public function userDeletesLastShareUsingTheSharingApi($user) {
 	 */
 	public function theUserGetsInfoOfLastShareUsingTheSharingApi() {
 		$this->userGetsInfoOfLastShareUsingTheSharingApi($this->currentUser);
+		$this->setLastShareData($this->getResponseXml());
```
skshetry: I don't know how previous tests were passing without this.

tests/acceptance/features/bootstrap/WebUISharingContext.php
```diff
@@ -513,6 +517,8 @@ public function theUserShouldSeeAnErrorMessageOnThePublicLinkShareDialogSaying(
 	 * @return void
 	 */
 	public function theUserShouldSeeAnErrorMessageOnThPublicLinkPopupSaying($message) {
+		// update public-sharing popup, as it might not have been set previously.
+		$this->publicSharingPopup = $this->publicShareTab->updateSharingPopup($this->getSession());
```
skshetry: There's no way to get the popup. It's not the best way, but I went with the easiest way.

karakayasemi (approved):

tests/acceptance/features/bootstrap/WebUISharingContext.php
```diff
@@ -750,6 +756,53 @@ public function thePublicAccessesPublicLinkWithPasswordUsingTheWebui($password)
 		$this->webUIGeneralContext->setCurrentPageObject($this->publicLinkFilesPage);
 	}
 
+	/**
+	 * @When /^user "([^"]*)" gets the info of the last public link using the sharing API$/
```
paurakhsharma: I think this should go inside `Sharing.php`

paurakhsharma (commented):
Just a small comment, other looks good for me.

HariBhandari07:
looks good

tests/acceptance/features/bootstrap/WebUISharingContext.php
individual-it: why not using `userGetsInfoOfLastShareUsingTheSharingApi()`?

individual-it (changes requested):

tests/acceptance/features/bootstrap/WebUISharingContext.php
skshetry: I feel this is *good-enough*, because if I put this in `Sharing.php`:
1. The similar step will exist.
2. Other developers might use this step even though, they should be using other step (due to suggestion from autocompletion).
3. Might need to use xpath on most element (due to responseXml differences).
4. This is only needed for `webUI` stuffs. Keeping on `webUI` prevents step2 from happening and provide isolation.

skshetry: Creating public links does not set `lastShareData` which `userGetsInfoOfLastShareUsingTheSharingApi` uses. So, I have to manually set.

individual-it: OK, you don't have the id of the share here, get it and then use `userGetsInfoOfLastShareUsingTheSharingApi`

individual-it (changes requested):
also fix https://github.com/owncloud/core/commit/aded6395c9f687042ffef1d3d362e98350e6e8d6#diff-7a6686160e3f99113ae6c2c807e63a58R801

I think that is what you are stumbling about.
in `isFieldInResponse` we mostly use data from the last response `$this->getResponseXml()` but for the expiration date we get the data when the share was created `$this->getServerLastShareTime()` this is not correct if the share was created but an other response like
`And the user gets the info of the last share using the sharing API` was done afterwards

individual-it (changes requested):

individual-it (approved):

phil-davis (approved):

phil-davis:
@skshetry please make the `stable10` backport.
Then I will look at what acceptance test changes are pending `release-10.2.1` backport and make a PR for those. That will run all these scenarios against `release-10.2.1`

skshetry:
Backport for `stable10`: #35703

PR #35595, opened 2019-06-19
Fix invalid token on pw reset

micbar:
## Description
The Password reset link is not working if a user used his e-mail address as a log in attribute.

## Expected Behaviour
1) Log in a user with his e-mail address and wrong password
2) Click on "Reset Password" Link
3) Click on Reset Link in E-Mail
4) See Reset Form

## What happens instead
Screenshot_2019-06-19 ownCloud Enterprise Edition

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3316

## Motivation and Context
Registration in Customer Portal is broken

## How Has This Been Tested?
1) Log in a user with his e-mail address and wrong password
2) Click on "Reset Password" Link
3) Click on reset Link in E-Mail
4) Reset the pw

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests modified

## Open tasks:

### DIFF
--- a/core/Controller/LostController.php
+++ b/core/Controller/LostController.php
@@ -36,6 +36,7 @@ use \OCP\IURLGenerator;
 use \OCP\IRequest;
 use \OCP\IL10N;
 use \OCP\IConfig;
+use OCP\IUser;
 use OCP\IUserManager;
 use OCP\Mail\IMailer;
 use OCP\Security\ISecureRandom;
@@ -203,6 +204,22 @@ class LostController extends Controller {
 	 * @return array
 	 */
 	public function email($user) {
+		try {
+			$userObject = $this->userManager->get($user);
+			if ($userObject === null) {
+				$users = $this->userManager->getByEmail($user);
+				// we only allow login by email if unique
+				if (\count($users) === 1) {
+					$user = $users[0]->getUID();
+				}
+			}
+			if ($userObject instanceof IUser) {
+				$user = $userObject->getUID();
+			}
+		} catch (\Exception $e) {
+			// Do not disclose any information during User lookup
+			return $this->success();
+		}
 		// FIXME: use HTTP error codes
 		try {
 			list($link, $token) = $this->generateTokenAndLink($user);
--- a/tests/Core/Controller/LostControllerTest.php
+++ b/tests/Core/Controller/LostControllerTest.php
@@ -85,6 +85,11 @@ class LostControllerTest extends TestCase {
 			->method('getDisplayName')
 			->willReturn('Existing User Name');
 
+		$this->existingUser
+			->expects($this->any())
+			->method('getUID')
+			->willReturn('ExistingUser');
+
 		$this->config = $this->getMockBuilder('\OCP\IConfig')
 			->disableOriginalConstructor()->getMock();
 		$this->l10n = $this->getMockBuilder('\OCP\IL10N')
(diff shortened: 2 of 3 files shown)


micbar:
@phil-davis Do we have acceptance tests for password reset with a real e-mail scenario?
Could we add this case?

micbar:
@IljaN Could you make a security review?

core/Controller/LostController.php
```diff
@@ -203,6 +204,22 @@ private function success() {
 	 * @return array
 	 */
 	public function email($user) {
+		try {
+			$userObject = $this->userManager->get($user);
+			if ($userObject === null) {
+				$users = $this->userManager->getByEmail($user);
```
patrickjahns: what if `$this->userManager->getByEmail($user);` will return `[]` ; ?

patrickjahns: can you cover this by a test case - as far as i could see the lookup will not throw an exception when no users have been found per mail

micbar: sure, good point.

patrickjahns: see https://github.com/owncloud/core/blob/master/lib/private/User/Manager.php#L465-L473

micbar: It will log something
```{"reqId":"vsYXKrfLFdU29JYrpI81","level":3,"time":"2019-06-19T10:15:47+00:00","remoteAddr":"10.254.254.254","user":"--","app":"core","method":"POST","url":"\/index.php\/lostpassword\/email","message":"Could not send reset email because User does not exist. User: terst\u00a5lsdsd@.com"}``

**Comment by phil-davis** [comment]
> @phil-davis Do we have acceptance tests for password reset with a real e-mail scenario?
> Could we add this case?

See issue #32889
And test scenario in `tests/acceptance/features/webUILogin/resetPasswordUsingEmail.feature` line 29.

You should be able to adjust the scenario, and if your fix is good then the scenario will pass.

File: core/Controller/LostController.php (hunk 1 above)
Comment (micbar) [reply core/Controller/LostController.php]: Added the test

File: tests/Core/Controller/LostControllerTest.php [hunk 2]
```diff
@@ -48,25 +47,25 @@ class LostControllerTest extends TestCase {
 	private $lostController;
 	/** @var IUser */
 	private $existingUser;
-	/** @var IURLGenerator | PHPUnit\Framework\MockObject\MockObject */
+	/** @var IURLGenerator | \PHPUnit_Framework_MockObject_MockObject */
```
phil-davis: Are these changes needed?
IMO `PHPUnit\Framework\MockObject\MockObject` is the newer way in PHPunit6 onwwards

micbar: reverted

micbar:
@phil-davis @patrickjahns 
We agree that the LostController is currently spaghetti code.
If we want to improve it, we would need to refactor the occ command too.

@phil-davis I adapted the acceptance scenario, let us see.

phil-davis:
@micbar there was a test step that had slightly out-of-date step text. I adjusted it (see 2nd commit) and ran the scenario locally - it passes.

patrickjahns (approved):
LGTM - it fixes the issue - refactoring for a cleaner approach needs to be scheduled later

phil-davis:
Backports:
`stable10` #35597 
`release-10.2.1` #35607

PR #35598, opened 2019-06-19
Return false if the rename fails except for connect exceptions

jvillafanez:
## Description
A failure during the rename operation due to insufficient permissions was causing the storage to be unavailable. In this case, the operation should return false instead of throwing a StorageNotAvailableException

## Related Issue
https://github.com/owncloud/phoenix/issues/1326

## Motivation and Context
It should fail normally instead of blocking the storage

## How Has This Been Tested?
This has been checked with the "old" UI, not with phoenix, although it's expected a similar behaviour.

1. Ensure the file can be edited by the user, but the parent folder is only readable
2. Try to rename a file.

**Note**: it seems the UI behaves badly under these circumstances. There is an error shown, but the "renamed" file disappears. The storage is available, and you can go outside and then inside the folder to make the file reappear again. The "move" http request ends up with a 404 error code.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -276,10 +276,15 @@ class SMB extends \OCP\Files\Storage\StorageAdapter {
 			} else {
 				$result = false;
 			}
+		} catch (ConnectException $e) {
+			$ex = new StorageNotAvailableException($e->getMessage(), $e->getCode(), $e);
+			$this->leave(__FUNCTION__, $ex);
+			throw $ex;
 		} catch (Exception $e) {
 			$this->swallow(__FUNCTION__, $e);
 			// Icewind\SMB\Exception\Exception, not a plain exception
 			if ($e->getCode() === 22) {
+				// some servers seem to return an error code 22 instead of the expected AlreadyExistException
 				if ($this->unlink($target)) {
 					$result = $this->share->rename($this->root . $source, $this->root . $target);
 					if ($result) {
@@ -289,13 +294,8 @@ class SMB extends \OCP\Files\Storage\StorageAdapter {
 				} else {
 					$result = false;
 				}
-			} elseif ($e->getCode() === 16) {
-				$this->swallow(__FUNCTION__, $e);
-				$result = false;
 			} else {
-				$ex = new StorageNotAvailableException($e->getMessage(), $e->getCode(), $e);
-				$this->leave(__FUNCTION__, $ex);
-				throw $ex;
+				$result = false;
 			}
 		} catch (\Exception $e) {
 			$this->swallow(__FUNCTION__, $e);
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -307,9 +307,7 @@ class SMB extends \OCP\Files\Storage\StorageAdapter {
 	private function removeFromCache($path) {
 		$path = \trim($path, '/');
 		// TODO The CappedCache does not really clear by prefix. It just clears all.
-		//$this->dirCache->clear($path);
 		$this->statCache->clear($path);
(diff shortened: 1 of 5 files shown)


apps/files_external/lib/Lib/Storage/SMB.php
```diff
@@ -289,13 +294,8 @@ public function rename($source, $target) {
 				} else {
 					$result = false;
 				}
-			} elseif ($e->getCode() === 16) {
-				$this->swallow(__FUNCTION__, $e);
-				$result = false;
 			} else {
-				$ex = new StorageNotAvailableException($e->getMessage(), $e->getCode(), $e);
```
DeepDiver1975: This code logic is there in other operations as well. Please double check those locations as well. THX

DeepDiver1975 (changes requested):

mmattel:
@jvillafanez question:
would your change also work in following situation:
client: macos, sync client installed
backend mount to sync to: smb
create a file on macos containing an invalid character for smb like mypicture_2019:06:19.png
when syncing, it tells that storage is not available instead rejecting the sync

jvillafanez:
It's very likely. In any case, the error isn't caused by the storage not being available but due to invalid chars in the file name.
I'm not fully sure how the client will behave because the changes are currently limited to the storage connector and there are several layers on top of it. This is still under investigation.

jvillafanez:
Rename UI problem should be fixed with https://github.com/owncloud/core/pull/35598/commits/1799af879aa8eda8c26da3921814eeff0fb52ba9 (checked in the "old" UI, not in phoenix). Throwing a Forbidden exception might not be nice, but I haven't seen a better alternative.

I've also notice that we've been writing information in the cache even though the operation failed. Maybe that has been the cause of ghost entries. They should be fixed with https://github.com/owncloud/core/pull/35598/commits/57807013f30047009a964fef1decc66aeab485cb (pending to be verified, but changes make sense to me)

mmattel:
> Throwing a Forbidden exception might not be nice, but I haven't seen a better alternative.

Maybe http 422 (Unprocessable Entity) ?
(The response status code indicates that the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions.)

jvillafanez:
I think we should throw any of the exceptions in https://github.com/sabre-io/dav/tree/master/lib/DAV/Exception which should be the ones that are expected to be handled by the sabre library (which will be the one handling the request anyway). It's unlikely that any other exception outside of that list will be properly handled.
I'm not familiar with the code around dav, so maybe I'm wrong. In any case I'd wait until @DeepDiver1975 confirms

mmattel:
FYI, I tried to test on my local Ubuntu 18.04 dev machine with a client installed how your PR works with the client. To the shame of oC, the Ubuntu client is either far out of date (2.4.x and unusable due to bugs) or not manually installable due to broken dependencies. I reported a client issue in central... :face_palm:

jvillafanez:
I don't think I can improve coverage here. Most of the changes related to the storage can't be unittested, and testing the changes in the view, if possible, is too complex.

DeepDiver1975 (approved):

DeepDiver1975:
> Maybe http 422 (Unprocessable Entity) ?

no - see https://tools.ietf.org/html/rfc4918#section-11.2

422 has a clear definition which does not match to this case.

403 looks reasonable and is consistent with the way we use this exception in other places

jvillafanez:
Backport in https://github.com/owncloud/core/pull/35683

HariBhandari07:
@jvillafanez I are doing the manual testing of this pr. How can I achieve step 1 ```Ensure the file can be edited by the user, but the parent folder is only readable``` of ```How has this been tested?``` . 

When a parent folder is shared with read-only permission the receiver cannot edit the child files.

PR #35447, opened 2019-06-06
The recipient of internal share should not increase permission

sharidas:
The recipient of internal share, should not be able
to increase the permission.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

When an internal/regular share is created by say `user1` to `user2`, the `user2` should not be allowed to increase the permissions granted by `user1`. The UI prevents it. But by using ocs api, its possible to increase the permision. This pr addresses the issue.

## Related Issue

- Fixes 

## Motivation and Context

The user who is not the owner of the share shouldn't be allowed to increase the permission of the share.

## How Has This Been Tested?

- Create 3 users `user1`, `user2` and `user3`
- As `user1` create folder `testFolder/subFolder`
- As `user1` share `testFolder` to `user2` with permission `can share` and uncheck `can edit` ( i.e, readonly )
- Now as `user2` try to reshare `testFolder/subFolder` to `user3`
- Execute the command 
```
 sujith@sujith-ownCloud  ~  curl -u user2:1 'http://localhost/testing/ocs/v2.php/apps/files_sharing/api/v1/shares/share_id' --request PUT --data 'permissions=21'                           
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>404</statuscode>
  <message>Cannot change permission of /subFolder</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data/>
</ocs>
 sujith@sujith-ownCloud  ~  curl -u user2:1 'http://localhost/testing/ocs/v2.php/apps/files_sharing/api/v1/shares/share_id' --request PUT --data 'permissions=31'                           
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>404</statuscode>
  <message>Cannot change permission of /subFolder</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data/>
</ocs>
 sujith@sujith-ownCloud  ~ 
```
- Similarly when a file is shared with `can share` and readonly permission by `user1` to `user2`, and `user2` reshares to `user3`, when the command is executed for the share of `user2`:
```
sujith@sujith-ownCloud  ~  curl -u user2:1 'http://localhost/testing/ocs/v2.php/apps/files_sharing/api/v1/shares/5' --request PUT --data 'permissions=21'
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
[...]

PR #35600, opened 2019-06-19
[release-10.2.1] Do not allow to set higher permissions on a public link share for a resource which was shared with limited permissions

DeepDiver1975:
…share-update-test

Do not allow to set higher permissions on a public link share for a r…

backport #35585

PR #35614, opened 2019-06-20
if user has enough permission, allow to increase permission

### LINKED ISSUE #35528: Permission of reshare cannot be increased
### Steps to reproduce
1. `user0` shares a folder `TMP` with `user1` with permissions 23 (can-share+read-update-create)
2. `user1` shares `TMP` with `user2` with permissions 17 (can-share+read)
3. `user1` tries to update the permissions of the share to 23

### Expected behaviour
Step 3 works, and the share now has permissions 23

### Actual behaviour
Step 3 returns a 404

### Server configuration
**Operating system**:
Ubuntu 18.04

**Web server:**
drone CI Apache

**Database:**
MySQL

**PHP version:**
7.*

current master from git

A Gherkin acceptance test scenario that demonstrates the issue:
```
  Scenario Outline: Update of reshare can increase permissions to the maximum allowed
    Given using OCS API version "<ocs_api_version>"
    And user "user2" has been created with default attributes and without skeleton files
    And user "user0" has created folder "/TMP"
    And user "user0" has created a share with settings
      | path        | /TMP  |
      | shareType   | 0     |
      | shareWith   | user1 |
      | permissions | 23    |
    And user "user1" has created a share with settings
      | path        | /TMP  |
      | shareType   | 0     |
      | shareWith   | user2 |
      | permissions | 17    |
    When user "user1" updates the last share using the sharing API with
      | permissions | 23 |
    Then the OCS status code should be "404"
    And the HTTP status code should be "<http_status_code>"
    #Then the OCS status code should be "<ocs_status_code>"
    #And the HTTP status code should be "200"
    Examples:
      | ocs_api_version | http_status_code |
      | 1               | 200              |
      | 2               | 404              |
      #| ocs_api_version | ocs_status_code |
      #| 1               | 100             |
      #| 2               | 200             |
```

karakayasemi:
## Description
After https://github.com/owncloud/core/pull/35447 merged, it has become that only node owners can increase permissions on a share. It has lead to this regression: https://github.com/owncloud/core/issues/35528

This PR checks Session User's permission on the share node. If user has enough permission, it let session user to increase permission of the share.

## Related Issue
- Fixes #35528 

## Motivation and Context
Solving bug.

## How Has This Been Tested?
Followed the scenario described in the issue. Also, we have acceptance tests that are covering bug scenario.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -828,7 +828,7 @@ class Share20OcsController extends OCSController {
 				$shareHome = $this->rootFolder->getUserFolder($share->getSharedBy());
 				$nodes = $shareHome->getById($share->getNode()->getId());
 				foreach ($nodes as $node) {
-					if (($node->getPermissions() | $newPermissions) !== $node->getPermissions()) {
+					if (!$this->strictSubsetOf($node->getPermissions(), $newPermissions)) {
 						return new Result(null, 404, 'Cannot increase permission of ' . $share->getTarget());
 					}
 				}
@@ -860,11 +860,11 @@ class Share20OcsController extends OCSController {
 				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
 				return new Result(null, 400, $this->l->t('Wrong or no update parameter given'));
 			} else {
-				if (($share->getPermissions() < $permissions) &&
-					!$this->canIncreasePermission($share)) {
+				$permissions = (int)$permissions;
+				if (!$this->strictSubsetOf($share->getPermissions(), $permissions) &&
+					!$this->canIncreasePermission($share, $permissions)) {
 					return new Result(null, 404, 'Cannot increase permission of ' . $share->getTarget());
 				}
-				$permissions = (int)$permissions;
 				$share->setPermissions($permissions);
 			}
 		}
@@ -880,7 +880,7 @@ class Share20OcsController extends OCSController {
 					$maxPermissions |= $incomingShare->getPermissions();
 				}
 
-				if ($share->getPermissions() & ~$maxPermissions) {
+				if (!$this->strictSubsetOf($maxPermissions, $share->getPermissions())) {
 					$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
 					return new Result(null, 404, $this->l->t('Cannot increase permissions'));
 				}
@@ -906,28 +906,30 @@ class Share20OcsController extends OCSController {
 	}
 
 	/**
-	 * If the user is not the owner of the share, then lets not allow to increase
-	 * the permission of the share.
+	 * If the user is not the owner of the share and has not enough permission on Node,
+	 * then lets not allow to increase the permission of the share.
 	 *
 	 * @param IShare $share
+	 * @param int $newPermissions
 	 * @return bool , true if permission is allowed to increase, else false. Also false if no user session available
 	 */
-	private function canIncreasePermission(IShare $share) {
+	private function canIncreasePermission(IShare $share, $newPermissions) {
 		$userObj = $this->userSession->getUser();
 		if ($userObj === null) {
 			return false;
 		}
 
-		$shareHome = $this->rootFolder->getUserFolder($share->getSharedBy());
+		$shareHome = $this->rootFolder->getUserFolder($userObj->getUID());
 		$nodes = $shareHome->getById($share->getNode()->getId());
 		foreach ($nodes as $node) {
+			$allowedPermissions = $node->getPermissions();
 			$nodeOwner = $node->getOwner()->getUID();
-			if ($nodeOwner !== $userObj->getUID()) {
-				return false;
+			if ($nodeOwner === $userObj->getUID() || $this->strictSubsetOf($allowedPermissions, $newPermissions)) {
+				return true;
 			}
 		}
 
-		return true;
+		return false;
 	}
 
 	/**
@@ -1262,5 +1264,16 @@ class Share20OcsController extends OCSController {
 		$share->setAttributes($newShareAttributes);
 
 		return $share;
+	}
+
+	/**
+	 * Check $newPermissions bit is a subset of $allowedPermissions
+	 *
+	 * @param int $allowedPermissions
+	 * @param int $newPermissions
+	 * @return boolean ,true if $allowedPermissions bit super set of $newPermissions bit, else false
+	 */
+	private function strictSubsetOf($allowedPermissions, $newPermissions) {
+		return (($allowedPermissions | $newPermissions) === $allowedPermissions);
 	}
 }
--- a/apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
+++ b/apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
@@ -2291,8 +2291,6 @@ class Share20OcsControllerTest extends TestCase {
 	public function testUpdateShareCannotIncreasePermissions() {
 		$ocs = $this->mockFormatShare();
 
-		$date = new \DateTime('2000-01-01');
-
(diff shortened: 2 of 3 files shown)


sharidas (approved):
LGTM :+1:

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -921,8 +922,9 @@ private function canIncreasePermission(IShare $share) {
 		$shareHome = $this->rootFolder->getUserFolder($share->getSharedBy());
 		$nodes = $shareHome->getById($share->getNode()->getId());
 		foreach ($nodes as $node) {
+			$allowedPermissions = $node->getPermissions();
 			$nodeOwner = $node->getOwner()->getUID();
-			if ($nodeOwner !== $userObj->getUID()) {
+			if ($nodeOwner !== $userObj->getUID() && ($allowedPermissions < $permissions)) {
```
DeepDiver1975: Our permissions have no hierarchy in their values.
Sharing(16) > Write(8) -> does this imply that if a user has share permissions he has automatically write permissions?

DeepDiver1975 (changes requested):
Please have a look at my comment above - I'm not happy ....

patrickjahns:
@karakayasemi 
please add unit tests to cover your change and outline intended behavior

apps/files_sharing/lib/Controller/Share20OcsController.php
karakayasemi: @DeepDiver1975 Thank you for pointing out this. Now, bit-wise operations used in comparison.

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -906,13 +906,14 @@ public function updateShare($id) {
 	}
 
 	/**
-	 * If the user is not the owner of the share, then lets not allow to increase
-	 * the permission of the share.
+	 * If the user is not the owner of the share and has not enough permission on Node,
+	 * then lets not allow to increase the permission of the share.
 	 *
 	 * @param IShare $share
+	 * @param int $permissions
```
jvillafanez: This might need extra information. I don't understand where these permissions come from, nor what these permissions are used for.
As far as I know, you're allowed to increase the permissions only if you're the owner, so you have enough information from the user session and the share.

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -921,8 +922,9 @@ private function canIncreasePermission(IShare $share) {
 		$shareHome = $this->rootFolder->getUserFolder($share->getSharedBy());
 		$nodes = $shareHome->getById($share->getNode()->getId());
 		foreach ($nodes as $node) {
+			$allowedPermissions = $node->getPermissions();
 			$nodeOwner = $node->getOwner()->getUID();
-			if ($nodeOwner !== $userObj->getUID()) {
+			if ($nodeOwner !== $userObj->getUID() && ($permissions & ~$allowedPermissions)) {
```
DeepDiver1975: Just to be sure - this is identical to https://github.com/owncloud/core/pull/35600/files#diff-faddd9984dfaf09a2b85d8dc4867207aR831

Who ever introduced bitfuxxing to owncloud:  :gun:

DeepDiver1975 (changes requested):

apps/files_sharing/lib/Controller/Share20OcsController.php
jvillafanez: yes. Better include at least a comment around so we know what this code is trying to do. It would be better to have function to handle these operations, but it's out of scope.

karakayasemi: I moved this logic to `hasEnoughPermissions` method and I singularized this check handling by using this method at least this class wide.

apps/files_sharing/lib/Controller/Share20OcsController.php
karakayasemi: I changed variable name as `newPermissions`.

> As far as I know, you're allowed to increase the permissions only if you're the owner, so you have enough information from the user session and the share.

To fix regression, I have add additional condition to this method and update its phpdoc. Now, if the share node is mounted to Session User's home folder and Session User has enough permission to increase permission, The user can increase permission even he is not owner of the node.

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -1263,4 +1265,15 @@ private function setShareAttributes(IShare $share, $formattedShareAttributes) {
...
 	}
+
+	/**
+	 * Check $newPermissions bit is a subset of $allowedPermissions
+	 *
+	 * @param int $allowedPermissions
+	 * @param int $newPermissions
+	 * @return boolean ,true if $allowedPermissions bit super set of $newPermissions bit, else false
+	 */
+	private function hasEnoughPermissions($allowedPermissions, $newPermissions) {
```
jvillafanez: `strictSubsetOf` might be a better name. `hasEnoughPermission` might sound like if $allowedPermissions is "read" and $newPermissions is "read+write" it should return true because "read+write" has enough permissions if "read" is allowed

karakayasemi: changed as `strictSubsetOf`

apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
```diff
@@ -3804,4 +3823,31 @@ public function testAcceptRejectShareOperationError($method, $target, $targetExi
...
+				$this->ocs,
+				'strictSubsetOf',
+				[$allowedPermissions, $newPermissions]
+			)
+		);
+	}
+
+	public function strictSubsetOfDataProvider() {
+		return [
+			[\bindec('11111'), \bindec('0111'), true],
```
DeepDiver1975: :+1:

DeepDiver1975 (approved):

apps/files_sharing/lib/Controller/Share20OcsController.php
jvillafanez: The problem I have is that I read "*If the user is not the owner of the share*" and I think "*if the method is called "canIncreasePermission" this is fine*" but then I read "*has not enough permission on Node,*" and I don't know what to think. How we can know if the user has or not enough permissions?
There is also a parameter with newPermissions which, just reading the documentation, I don't understand what is used for. Are those the permissions we're expecting to be set later?

My main point here is that, if I already have problems to understand these changes (just a part of it, not all), I'm pretty sure next year I'll have no idea what this code do, and this will be something risky to touch because it isn't clear what the code should be doing (not what the code really does).

Don't be afraid of writing an explanation in the code as long as it helps to understand what the code is doing.

micbar:
@karakayasemi new tests merged into release branch. Could you rebase?

phil-davis:
Note: acceptance tests in `master` `stable10` and `release-10.2.1` are all the same. So for getting the first version of this PR reviewed, `master` base should be fine.
Then we also make the backport PRs to `stable10` and `release-10.2.1` and confirm CI is good there also.

micbar:
@phil-davis @karakayasemi Sorry for my misleading comment.
Anyway, I would give CI a last chance to finish correctly.

patrickjahns:
@ownclouders rebase

IljaN (approved):

phil-davis (approved):

karakayasemi:
Codecov came for backport PRs. I deleted old codecov report comments. I will try my chance with this way, started a new build.

phil-davis:
Backports
`stable10` #35634 
`release-10.2.1` #35633

PR #35618, opened 2019-06-21
Improve isFieldInResponse to handle more combinations

phil-davis:
## Description
Enhance `isFieldInResponse()` acceptance test method so that it:
- knows if the caller is expecting success, and if the field does not exist or has an unexpected value, or... then emit some debug information to help the developer know what the actual value was etc.
- add the special value `ANY_VALUE` so we can test for a field existing (or not existing) and we do not care what the value is.

## Motivation and Context
If I write:
```
    And the fields of the last response should not include
      | share_with | anne |
```
Then if the `share_with` field exists, but has some other value like `bob`, then the test passes.
I want to be able to to test that a field does not exist at all.

If the field is not as expected, then as well as a fail message, I would like to know what value it was set to, or if the field did not exist. That can really help when trying to work out why a test is failing.

## How Has This Been Tested?
Local runs of `createShare.feature` scenarios with different combinations of fields and values in steps for:
```
And the fields of the last response should include
And the fields of the last response should not include
```
to verify that when something is different from expected, that the test does fail, and emits a somewhat-useful message.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -638,12 +638,43 @@ trait Sharing {
 
 	/**
 	 * @param string $field
+	 * @param string $value
 	 * @param string $contentExpected
+	 * @param bool $expectSuccess if true then the caller expects that the field
+	 *                            has the expected content
+	 *                            emit debugging information if the field is not as expected
+	 *
+	 * @return bool
+	 */
+	public function doesFieldValueMatchExpectedContent(
+		$field, $value, $contentExpected, $expectSuccess = true
+	) {
+		if (($contentExpected === "ANY_VALUE")
+			|| (($contentExpected === "A_TOKEN") && (\strlen($value) === 15))
+			|| (($contentExpected === "A_NUMBER") && \is_numeric($value))
+			|| (($contentExpected === "AN_URL") && $this->isAPublicLinkUrl($value))
+			|| (($field === 'remote') && (\rtrim($value, "/") === $contentExpected))
+			|| ($contentExpected === $value)
+		) {
+			if (!$expectSuccess) {
+				echo $field . " is unexpectedly set with value '" . $value . "'\n";
+			}
+			return true;
+		}
+		return false;
+	}
+
+	/**
+	 * @param string $field
+	 * @param mixed $contentExpected
+	 * @param bool $expectSuccess if true then the caller expects that the field
+	 *                            is in the response with the expected content
+	 *                            so emit debugging information if the field is not correct
 	 * @param SimpleXMLElement $data
 	 *
 	 * @return bool
 	 */
-	public function isFieldInResponse($field, $contentExpected, $data = null) {
+	public function isFieldInResponse($field, $contentExpected, $expectSuccess = true, $data = null) {
 		if ($data === null) {
 			$data = $this->getResponseXml()->data[0];
 		}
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -639,11 +639,14 @@ public function createShare(
...
 	 * @param string $contentExpected
+	 * @param bool $expectSuccess if true then the caller expects that the field
+	 *                            is in the response with the expected content
+	 *                            so emit debugging information if the field is not correct
 	 * @param SimpleXMLElement $data
 	 *
 	 * @return bool
 	 */
-	public function isFieldInResponse($field, $contentExpected, $data = null) {
+	public function isFieldInResponse($field, $contentExpected, $expectSuccess = true, $data = null) {
```
phil-davis: Note: nothing currently calls this passing `$data` parameter. So we can push it out to parameter 4 without breaking anything.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -659,35 +662,58 @@ public function isFieldInResponse($field, $contentExpected, $data = null) {
...
+			$fieldValue = "";
 			foreach ($data as $element) {
 				if (!isset($element->$field)) {
 					continue;
 				}
-				if ($contentExpected == "A_TOKEN") {
-					return (\strlen((string)$element->$field) == 15);
+				$fieldIsSet = true;
+				$fieldValue = (string)$element->$field;
+				if ($contentExpected == "ANY_VALUE") {
```
individual-it:
```suggestion
				if ($contentExpected === "ANY_VALUE") {
```

individual-it (changes requested):

tests/acceptance/features/bootstrap/Sharing.php
phil-davis: @individual-it I did another refactoring, also putting in plenty of `===`
Please review the whole thing again.

tests/acceptance/features/bootstrap/Sharing.php
skshetry: That's really not the `$expectSuccess`, but provides debug information. And, as it's purpose is to provide debug info, it would be good to push it to the edge, I think.

individual-it (approved):

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -636,14 +636,48 @@ public function createShare(
...
+	 * @param bool $expectSuccess if true then the caller expects that the field
+	 *                            has the expected content
+	 *                            emit debugging information if the field is not as expected
+	 *
+	 * @return bool
+	 */
+	public function doesFieldValueMatchExpectedContent(
+		$field, $value, $contentExpected, $expectSuccess = true
+	) {
+		if (($contentExpected === "ANY_VALUE")
```
skshetry: It was quite difficult to understand previously. Next thing to do would be to not pass `$contentExpected` as string (except, for if we are checking for value). We could encapsulate these on something like `Token` class as static variables.

individual-it: or using our `%something%` inline substitutions

phil-davis: Yes, someday this will need the next refactoring.
e.g. if/when somebody writes a test with usernames `A_TOKEN` and `A_NUMBER` and they want to test that `share_with` field has the actual value `A_NUMBER` :)

tests/acceptance/features/bootstrap/Sharing.php
phil-davis: It does actually tell the function if the caller is expecting a `true` or `false` return.
If the caller is expecting `true` but the function is about to return `false` then emit some debug info to give clues about why the expected content was not found.
If the caller is expecting `false` but the function is about to return `true` then emit some debug info to give clues about what (unexpected) content was found.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -658,38 +689,44 @@ public function isFieldInResponse($field, $contentExpected, $data = null) {
 					)
 				) . " 00:00:00";
 		}
+
+		$contentExpected = (string) $contentExpected;
```
phil-davis: We let the caller pass "mixed" in `$contentExpected` (just because they can anyway).
The responses really only have string values (even if they happen to look like numbers...). So we cast to string here, then can do `===` comparisons in comfort.

phil-davis:
Backport `stable10` #35622

PR #35636, opened 2019-06-24
Feature/webdav trash

DeepDiver1975:
## Description
We want a WebDAV API as discussed in #15646 

This is how it looks like:
- remote.php/dav/trash-bin is the root folder for all trashbins - each user will find their own trahsbin inside
- remote.php/dav/trash-bin/$user is the users root for trashbins - it holds all trahsed items of the user listed by file id
- to restore a trash item use a MOVE operation. In the final implementation the destination can be anything and anywhere: MOVE remote.php/dav/trash-bin/$user/$fileId remote.php/dav/$user/path/to/restore.txt
- to permanently delete an item from the trashbin perform a DELETE remote.php/dav/trash-bin/$user/$fileId
- to delete all items from the trashbin perform a DELETE remote.php/dav/trash-bin/$user/

### Listing trash items
```sh
$ curl 'http://localhost:8080/remote.php/dav/trash-bin/admin//' -X PROPFIND -H 'authorization: Basic ......' -H 'OCS-APIREQUEST: true' -H 'Content-Type: application/xml; charset=UTF-8' -H 'Depth: 1' --data-binary $'<?xml version="1.0"?>\n<d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">\n  <d:prop>\n    <oc:trashbin-original-filename />\n    <oc:trashbin-original-location />\n    <oc:trashbin-delete-timestamp />\n    <d:getcontentlength />\n    <d:resourcetype />\n  </d:prop>\n</d:propfind>' --compressed | xmllint --format -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1817    0  1533  100   284   8612   1595 --:--:-- --:--:-- --:--:-- 10265
```
```xml
<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:oc="http://owncloud.org/ns">
  <d:response>
    <d:href>/remote.php/dav/trash-bin/admin/</d:href>
    <d:propstat>
      <d:prop>
        <d:resourcetype>
          <d:collection/>
        </d:resourcetype>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
    <d:propstat>
      <d:prop>
        <oc:trashbin-original-filename/>
        <oc:trashbin-original-location/>
        <oc:trashbin-delete-timestamp/>
        <d:getcontentlength/>
      </d:prop>
      <d:status>HTTP/1.1 404 Not Found</d:status>
    </d:propstat>
  </d:response>
  <d:response>
    <d:href>/remote.php/dav/trash-bin/admin/2147488234</d:href>
    <d:propstat>
      <d:prop>
        <oc:trashbin-original-filename>Neue Textdatei.txt</oc:trashbin-original-filename>
        <oc:trashbin-original-location>./Neue Textdatei.txt</oc:trashbin-original-location>
        <oc:trashbin-delete-timestamp>1561454917</oc:trashbin-delete-timestamp>
        <d:getcontentlength>17</d:getcontentlength>
        <d:resourcetype/>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
  <d:response>
    <d:href>/remote.php/dav/trash-bin/admin/2147488238/</d:href>
    <d:propstat>
      <d:prop>
        <oc:trashbin-original-filename>Neuer Ordner</oc:trashbin-original-filename>
        <oc:trashbin-original-location>./Neuer Ordner</oc:trashbin-original-location>
        <oc:trashbin-delete-timestamp>1561454917</oc:trashbin-delete-timestamp>
        <d:resourcetype>
          <d:collection/>
        </d:resourcetype>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
    <d:propstat>
      <d:prop>
        <d:getcontentlength/>
      </d:prop>
      <d:status>HTTP/1.1 404 Not Found</d:status>
    </d:propstat>
  </d:response>
</d:multistatus>
```

### Restoring a file
```sh
$ curl 'http://localhost:8080/remote.php/dav/trash-bin/admin/2147488234' -X MOVE -H 'authorization: Basic YWRtaW46YWRtaW4=' -H 'OCS-APIREQUEST: true' -H 'Overwrite: F' -H 'Destination: http://localhost:8080/remote.php/dav/files/admin/test.txt' -H 'X-Request-ID: fcff1fa3-60d3-4b6b-b7bd-06598aeb8b1c' -v 
* Expire in 0 ms for 6 (transfer 0x55816f5d8dd0)
* Expire in 1 ms for 1 (transfer 0x55816f5d8dd0)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* Expire in 1 ms for 1 (transfer 0x55816f5d8dd0)
* Expire in 1 ms for 1 (transfer 0x55816f5d8dd0)
*   Trying ::1...
* TCP_NODELAY set
* Expire in 149998 ms for 3 (transfer 0x55816f5d8dd0)
* Expire in 200 ms for 4 (transfer 0x55816f5d8dd0)
* Connected to localhost (::1) port 8080 (#0)
> MOVE /remote.php/dav/trash-bin/admin/2147488234 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.64.0
> Accept: */*
> authorization: Basic ......
> OCS-APIREQUEST: true
> Overwrite: F
> Destination: http://localhost:8080/remote.php/dav/files/admin/test.txt
> X-Request-ID: fcff1fa3-60d3-4b6b-b7bd-06598aeb8b1c
> 
  0     0    0     0    0     0      0      0 --:--:--  0:00:42 --:--:--     0< HTTP/1.1 201 Created
< Host: localhost:8080
< Date: Tue, 25 Jun 2019 09:39:11 GMT
< Connection: close
< X-Powered-By: PHP/7.2.19-1+0~20190531112545.22+buster~1.gbp75765b
< Pragma: no-cache
< Set-Cookie: oc_sessionPassphrase=xxxx
< X-Frame-Options: SAMEORIGIN
< X-Robots-Tag: none
< X-Download-Options: noopen
< Set-Cookie: xxxx
< Set-Cookie: xxxx
< OC-ETag: "11713caa12b7f721046b55a9d707f86c"
< ETag: "11713caa12b7f721046b55a9d707f86c"
< OC-FileId: 2147488234ocrhbj031xi4
< Content-Length: 0
< 
  0     0    0     0    0     0      0      0 --:--:--  0:00:42 --:--:--     0
* Closing connection 0
```

## Related Issue
- refs #15646 

## How Has This Been Tested?
- :hand: 
- tests will follow

Screenshot from 2019-06-21 13-25-17
Screenshot from 2019-06-21 13-25-28
Screenshot from 2019-06-21 13-25-48
Screenshot from 2019-06-21 13-26-48

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/apiTrashbin/trashbinRestore.feature
+++ b/tests/acceptance/features/apiTrashbin/trashbinRestore.feature
@@ -19,7 +19,6 @@ Feature: Restore deleted files/folders
     And user "user0" has shared folder "/shared" with user "user1"
     And user "user1" has moved file "/shared" to "/renamed_shared"
     And user "user1" has deleted file "/renamed_shared/shared_file.txt"
-    And user "user1" has logged in to a web-style session
     When user "user1" restores the file with original path "/renamed_shared/shared_file.txt" using the trashbin API
     Then as "user1" the file with original path "/renamed_shared/shared_file.txt" should not exist in trash
     And user "user1" should see the following elements
@@ -36,7 +35,6 @@ Feature: Restore deleted files/folders
     And user "user0" has been created with default attributes and skeleton files
     And user "user0" has deleted file "/textfile0.txt"
     And as "user0" file "/textfile0.txt" should exist in trash
-    And user "user0" has logged in to a web-style session
     When user "user0" restores the folder with original path "/textfile0.txt" using the trashbin API
     Then as "user0" the folder with original path "/textfile0.txt" should not exist in trash
     And user "user0" should see the following elements
@@ -59,26 +57,9 @@ Feature: Restore deleted files/folders
     And user "user0" has created folder "/new-folder"
     And user "user0" has moved file "/textfile0.txt" to "/new-folder/new-file.txt"
     And user "user0" has deleted file "/new-folder/new-file.txt"
-    And user "user0" has logged in to a web-style session
     When user "user0" restores the file with original path "/new-folder/new-file.txt" using the trashbin API
     Then as "user0" the file with original path "/new-folder/new-file.txt" should not exist in trash
     And as "user0" file "/new-folder/new-file.txt" should exist
-    Examples:
-      | dav-path |
-      | old      |
-      | new      |
-
-  Scenario Outline: A file deleted from a folder is restored to root if the original folder does not exist
-    Given using <dav-path> DAV path
-    And user "user0" has been created with default attributes and skeleton files
-    And user "user0" has created folder "/new-folder"
-    And user "user0" has moved file "/textfile0.txt" to "/new-folder/new-file.txt"
-    And user "user0" has deleted file "/new-folder/new-file.txt"
-    And user "user0" has deleted folder "/new-folder"
-    And user "user0" has logged in to a web-style session
-    When user "user0" restores the file with original path "/new-folder/new-file.txt" using the trashbin API
-    Then as "user0" the file with original path "/new-folder/new-file.txt" should not exist in trash
-    And as "user0" file "/new-file.txt" should exist
     Examples:
       | dav-path |
       | old      |
@@ -91,7 +72,6 @@ Feature: Restore deleted files/folders
     And user "user0" has moved file "/textfile0.txt" to "/new-folder/new-file.txt"
     And user "user0" has deleted file "/new-folder/new-file.txt"
     And user "user0" has deleted folder "/new-folder"
-    And user "user0" has logged in to a web-style session
     When user "user0" restores the folder with original path "/new-folder" using the trashbin API
     And user "user0" restores the file with original path "/new-folder/new-file.txt" using the trashbin API
     Then as "user0" the file with original path "/new-folder/new-file.txt" should not exist in trash
@@ -108,7 +88,6 @@ Feature: Restore deleted files/folders
     And user "user0" has moved file "/textfile0.txt" to "/new-folder/new-file.txt"
     And user "user0" has deleted file "/new-folder/new-file.txt"
     And user "user0" has deleted folder "/new-folder"
-    And user "user0" has logged in to a web-style session
     When user "user0" creates folder "/new-folder" using the WebDAV API
     And user "user0" restores the file with original path "/new-folder/new-file.txt" using the trashbin API
     Then as "user0" the file with original path "/new-folder/new-file.txt" should not exist in trash
@@ -129,7 +108,6 @@ Feature: Restore deleted files/folders
     And user "user0" has moved file "/textfile0.txt" to "/local_storage/tmp/textfile0.txt"
     And user "user0" has deleted file "/local_storage/tmp/textfile0.txt"
     And as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" should exist in trash
-    And user "user0" has logged in to a web-style session
     When user "user0" restores the folder with original path "/local_storage/tmp/textfile0.txt" using the trashbin API
     Then as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" should not exist in trash
     And user "user0" should see the following elements
@@ -153,7 +131,6 @@ Feature: Restore deleted files/folders
     And user "user0" has uploaded chunk file "1" of "1" with "AA" to "/local_storage/tmp/textfile0.txt"
     And user "user0" has deleted file "/local_storage/tmp/textfile0.txt"
     And as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" should exist in trash
-    And user "user0" has logged in to a web-style session
     When user "user0" restores the folder with original path "/local_storage/tmp/textfile0.txt" using the trashbin API
     Then as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" should not exist in trash
     And the downloaded content when downloading file "/local_storage/tmp/textfile0.txt" for user "user0" with range "bytes=0-1" should be "AA"
@@ -171,8 +148,6 @@ Feature: Restore deleted files/folders
       | 1 | AA |
     And user "user0" has deleted file "/local_storage/tmp/textfile0.txt"
     And as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" should exist in trash
-    And user "user0" has logged in to a web-style session
     When user "user0" restores the folder with original path "/local_storage/tmp/textfile0.txt" using the trashbin API
     Then as "user0" the folder with original path "/local_storage/tmp/textfile0.txt" should not exist in trash
     And the downloaded content when downloading file "/local_storage/tmp/textfile0.txt" for user "user0" with range "bytes=0-1" should be "AA"
-    
--- a/apps/dav/lib/Connector/Sabre/CopyEtagHeaderPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/CopyEtagHeaderPlugin.php
@@ -22,6 +22,7 @@
 
 namespace OCA\DAV\Connector\Sabre;
 
+use Sabre\DAV\Exception\NotFound;
 use Sabre\HTTP\RequestInterface;
 use Sabre\HTTP\ResponseInterface;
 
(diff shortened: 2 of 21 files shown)


felix-schwarz:
> * remote.php/dav/trash-bin/admin is the users root for trashbins - it holds all trahsed items listed by file id

What would a `PROPFIND` on that location look like then? Which metadata would be available?

Would the `remote.php/dav/trash-bin/admin` root allow for change detection via `ETag` (just like the regular root folder does)?

> * to restore a trash item use a MOVE operation. In the final implementation the destination can be anything and anywhere: MOVE remote.php/dav/trash-bin/$user/$fileId remote.php/dav/$user/path/to/restore.txt

How can a client implement functionality identical to the `Restore` button in the current web UI? 

The previous PR mentioned a special folder to `MOVE` to to achieve this. Is this still planned?

If not (or alternatively), providing the original location as part of the response to a `PROPFIND` on `remote.php/dav/trash-bin/admin` would have the benefit that clients could show users where an item will be restored to (and possibly navigate there following a restore).

@DeepDiver1975 @davigonz @hosy @guruz @ckamm @ogoffart

DeepDiver1975:
@felix-schwarz THX for the feedback - I'll update the description asap.

One thing is clear: there is no etag propagation on the trashbin. I agree this would be nice but it is not possible.
What is also missing on the trash-bin endpoint is the capability to load page-wise (like the REPORT on the files endpoint)

davigonz:
> to restore a trash item use a MOVE operation. In the final implementation the destination can be anything and anywhere: MOVE remote.php/dav/trash-bin/$user/$fileId remote.php/dav/$user/path/to/restore.txt

Yes, this solution is much better than the previous PR one, as long as you mean moving the restored file to its previous location when you say "the destination can be anything and anywhere". There should not be a new location for a restored file, but the original one.

> remote.php/dav/trash-bin/admin is the users root for trashbins - it holds all trashed items listed by file id

What is the purpose of this API? I do not see a usecase in which admin users would like to restore normal user files, unless restore option can be disabled for normal users. It's like if we had `remote.php/dav/files/admin` to see all the user files, it sounds a bit weird to me. 

> If not (or alternatively), providing the original location as part of the response to a PROPFIND on remote.php/dav/trash-bin/admin would have the benefit that clients could show users where an item will be restored to (and possibly navigate there following a restore).

Not sure if this would be needed, a restored file should be moved to its previous location, not anywhere and that location could keep inmutable in local database `path` field.

DeepDiver1975:
> Yes, this solution is much better than the previous PR one, as long as you mean moving the restored file to its previous location when you say "the destination can be anything and anywhere". There should not be a new location for a restored file, but the original one.

The client has to grab the original location from the webdav property (see the screenshots - which I will update asap). But in addition the api allows to restore the file to any other location - e.g. if there is already a file with the same name living in the original location.



> What is the purpose of this API? I do not see a usecase in which admin users would like to restore normal user files, unless restore option can be disabled for normal users. It's like if we had `remote.php/dav/files/admin` to see all the user files, it sounds a bit weird to me.

Maybe a bit missleading descrpiption ...
remote.php/dav/trash-bin/admin/ holds the trash items of admin
remote.php/dav/trash-bin/tom/ holds the trash items of tom

DeepDiver1975:
(example curl commands for listing trash items and for restoring have been added to the description)

davigonz:
Thanks for the explanation @DeepDiver1975 

> The client has to grab the original location from the webdav property (see the screenshots - which I will update asap). But in addition the api allows to restore the file to any other location - e.g. if there is already a file with the same name living in the original location.

I see your point, what I'm used to see when there's already a file with the same name living in the original location is:

1. File is automatically renamed to something like `file(restored).txt`
2. A dialog with the options `Keep both` and `Replace` is prompted.

Both scenarios end up with the file restored in the same location. The other approach is more flexible for sure but in my opinion doesn't fully fit with the definition of restoring a deleted file. Besides, it would require more effort in the clients since the users would need to choose the new location for the file and if there's a conflict with an existing file there, ask them to keep both or replace.

DeepDiver1975:
Changing the Destination in server Side violates the move semantics.
Furthermore i consider more flexibility a Bonus.

butonic (approved):
The added flexibility is a bonus:
- clients can show the origin of a deleted file (@deepdiver do we have a path in addition to the parent fileid? the path to the original location might have changed) at the cost of doing the lookup every time they want to restore a file with a MOVE
- the restore target path determination logic moves to the client, which is more more complex to handle, but relieves the server from a lot of special purpose code and simplifies the API because we use a standard MOVE to restore the file, no magic source or target node

DeepDiver1975:
>     * (@deepdiver do we have a path in addition to the parent fileid? the path to the original location might have changed)

trash items have 2 webdav properties for these cases: 
- {http://owncloud.org/ns}trashbin-original-filename - which is the original file name
- {http://owncloud.org/ns}trashbin-original-location - which is the originial location

Looking at them they seem to have some redundency. {http://owncloud.org/ns}trashbin-original-location might be enough

davigonz:
> Changing the Destination in server Side violates the move semantics.

Agreed, what I meant is that using the MOVE operation is fine but it does not necessarily mean that the restore target path needs to be chosen by the final user.

I was thinking in this flow:

1. Admin user selects the deleted files option:
    - `PROPFIND` to `http://localhost:8080/remote.php/dav/trash-bin/admin/` will get all the deleted files for the admin user. Let's save them into database and update the list in the view.
2. Admin user wants to restore `picture.png`, located in root directory:
    - `MOVE` 'http://localhost:8080/remote.php/dav/trash-bin/admin/12345678' to `remote.php/dav/admin/picture.png`. The restore target path is always what was retrieved via `<oc:trashbin-original-location>` and we already have this information in the local database (was saved in the step 1). 
3. If the file already exists in the restore target path, let the user decides what to do with it => Keep both, replace or nothing.

> clients can show the origin of a deleted file (@deepdiver do we have a path in addition to the parent fileid? the path to the original location might have changed) at the cost of doing the lookup every time they want to restore a file with a MOVE

I think that showing the origin of a deleted file is possible even without the flexibility of letting the user chose the final destination of the file.

> the restore target path determination logic moves to the client, which is more more complex to handle, but relieves the server from a lot of special purpose code and simplifies the API because we use a standard MOVE to restore the file, no magic source or target node

In case you want to proceed with the MOVE operation to anywhere, yes, it makes total sense to handle that logic in the client.

Just sharing my thoughts, proceed with the solution you consider better 😉

butonic:
@davigonz yap, for now the clients should just try to restore to the original folder (id based). If that fails for whatever reason the client can either follow the old implementation and append the `(restored)`, although that might run into filename lenght issues, or present the user with a file picker dialog ...

In any case, I think this actually makes it easier to control what is happening. The less magic the better.

tests/acceptance/features/apiTrashbin/trashbinRestore.feature
```diff
@@ -68,22 +68,6 @@ Feature: Restore deleted files/folders
       | old      |
       | new      |
 
-  Scenario Outline: A file deleted from a folder is restored to root if the original folder does not exist
```
individual-it: why did you had to remove that scenario? Is it not applicable anymore?

individual-it (commented):
with the webdav API we should not need `And user "user0" has logged in to a web-style session` in the tests

tests/acceptance/features/apiTrashbin/trashbinRestore.feature
DeepDiver1975: Yes - the behavior of restore changed with the new API. The target location is fully controlled by the clients. There is no need in verifying that under certain situations the file is restore into a different location.

felix-schwarz:
Would it be possible to compute a hash over the largest `oc:trashbin-delete-timestamp` value and the total number of items (both files and collections) in the trash (f.ex. `sha1("1561454917"."2")` for the example above) - and then return it as `ETag` for the trash root folder?

Then a `PROPFIND` with `Depth: 0` could be used to check for changes.

@DeepDiver1975 @davigonz @hosy @guruz @ckamm @ogoffart

DeepDiver1975:
> with the webdav API we should not need `And user "user0" has logged in to a web-style session` in the tests

@individual-it I assume this can be done all over the place in all feature files where the trashbin api is used in steps

DeepDiver1975:
> Would it be possible to compute a hash over the largest `oc:trashbin-delete-timestamp` value and the total number of items (both files and collections) in the trash (f.ex. `sha1("1561454917"."2")` for the example above) - and then return it as `ETag` for the trash root folder?
> 
> Then a `PROPFIND` with `Depth: 0` could be used to check for changes.
> 
> @DeepDiver1975 @davigonz @hosy @guruz @ckamm @ogoffart

@felix-schwarz let me see what I can do ....

butonic (approved):
same date format as for getlastmodified.

individual-it (approved):

phil-davis:
Backport `stable10` #35716

PR #35641, opened 2019-06-24
Use storages to copy the skeleton files

sharidas:
Introduce a way to copy the skeleton files.
Here we copy the files using the storage,
instead of old way as in the util file.

## Description

Lets say when user created a rule in the firewall: user who belongs to a group & user device is a browser, then first time login users are affected. They are not allowed to login. If the user is already logged in, then they cannot create new files/folders nor read files or download is allowed.

This change would allow the user to login who are logging in for the first time. A small change has been made to the way skeletons are copied to the user folder. This helps to resolve the inconsistency of some users are able to login and some not because of firewall rule mentioned above.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

If the user is prohibitted to login by a firewall rule, then login should not be blocked, even if its a first time login ( because that's when the skeleton files are copied and necessary folder are created ).

## How Has This Been Tested?

- Create a user, `user1` and assign to a group, `group1`
- Create a User Group rule with group name `group1`
- Create a User Device with `Others (Browsers etc )`
- Now try to login as `user`.
- The user can see the skeleton file after login. But user cannot modify or create new files, because of the rule, which is ok.
- Tested this with encryption (masterkey and user-keys encryption), files_primary_s3 app, as this change modifies the way skeleton is copied.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Requires https://github.com/owncloud/firewall/pull/580

## Open tasks:

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -444,7 +444,7 @@ class Session implements IUserSession, Emitter {
 
 			try {
 				// copy skeleton
-				\OC_Util::copySkeleton($user, $userFolder);
+				\OC_Util::copySkeletonUsingStorage($userFolder);
 			} catch (NotPermittedException $ex) {
 				// possible if files directory is in an readonly jail
 				$this->logger->warning(
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -57,6 +57,7 @@
  */
 
 use OCP\Authentication\Exceptions\AccountCheckException;
+use OCP\Files\Folder;
 use OCP\Files\NoReadAccessException;
 use OCP\Files\Storage\IStorage;
 use OCP\IUser;
@@ -355,11 +356,45 @@ class OC_Util {
 	}
 
 	/**
+	 * Copies the skeleton to user/files using the storage
+	 *
+	 * @param Folder $userFolder
+	 */
+	public static function copySkeletonUsingStorage(Folder $userFolder) {
+		//Set local storage
+		$skeletonDir = \OC::$server->getConfig()->getSystemValue('skeletondirectory', \OC::$SERVERROOT . '/core/skeleton');
+		$storageArgs = [
+			'datadir' => $skeletonDir
+		];
+		$sourceStorage = new OC\Files\Storage\Local($storageArgs);
+		$sourceDirHandler = $sourceStorage->opendir('');
+		while (($file = \readdir($sourceDirHandler)) !== false) {
+			if (!OC\Files\Filesystem::isIgnoredDir($file)) {
+				try {
+					$userFolder->getStorage()->copyFromStorage($sourceStorage, $file, "files/$file");
+				} catch (\OCP\Files\NotFoundException $ex) {
+					\OC::$server->getLogger()->logException($ex);
+				}
+			}
+		}
+
+		try {
+			$userFolder->getStorage()->getScanner()->scan('', OC\Files\Cache\Scanner::SCAN_RECURSIVE);
+		} catch (\OCP\Files\StorageNotAvailableException $ex) {
+			\OC::$server->getLogger()->logException($ex);
+		} catch (\OCP\Files\NotFoundException $ex) {
+			\OC::$server->getLogger()->logException($ex);
+		}
+	}
+
+	/**
 	 * copies the skeleton to the users /files
 	 *
 	 * @param String $userId
 	 * @param \OCP\Files\Folder $userDirectory
 	 * @throws \OC\HintException
+	 *
+	 * @deprecated Use \OC_Util::copySkeletonUsingStorage($userFolder) instead.
 	 */
 	public static function copySkeleton($userId, \OCP\Files\Folder $userDirectory) {
 		$skeletonDirectory = \OC::$server->getConfig()->getSystemValue('skeletondirectory', \OC::$SERVERROOT . '/core/skeleton');
--- a/tests/lib/User/SessionTest.php
+++ b/tests/lib/User/SessionTest.php
@@ -24,6 +24,10 @@ use OC\User\Session;
 use OCP\App\IServiceLoader;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Authentication\IAuthModule;
+use OCP\Files\Cache\IScanner;
+use OCP\Files\Folder;
+use OCP\Files\IRootFolder;
+use OCP\Files\Storage\IStorage;
 use OCP\IConfig;
 use OCP\ILogger;
 use OCP\IRequest;
@@ -74,6 +78,11 @@ class SessionTest extends TestCase {
 		$this->serviceLoader = $this->createMock(IServiceLoader::class);
 		$this->userSyncService = $this->createMock(\OC\User\SyncService::class);
 		$this->eventDispatcher = new EventDispatcher();
+	}
+
+	protected function tearDown() {
+		$this->restoreService('RootFolder');
+		parent::tearDown();
 	}
 
 	public function testGetUser() {
@@ -192,6 +201,52 @@ class SessionTest extends TestCase {
 			$this->tokenProvider, $this->config, $this->logger, $this->serviceLoader,
 			$this->userSyncService, $this->eventDispatcher);
 		$userSession->setUser($user);
+	}
+
+	public function testPrepareUserLoginWithFirstTimeLogin() {
+		$eventDispatcher = $this->createMock(EventDispatcher::class);
+		$rootFolder = $this->createMock(IRootFolder::class);
+		$useFolder = $this->createMock(Folder::class);
+		$scanner = $this->createMock(IScanner::class);
+		$scanner->method('scan')
+			->willReturn([]);
+		$storage = $this->createMock(IStorage::class);
+		$storage->method('copyFromStorage')
+			->willReturn(true);
+		$storage->method('getScanner')
+			->willReturn($scanner);
+		$useFolder->method('getStorage')
+			->willReturn($storage);
+		$rootFolder->method('getUserFolder')
+			->willReturn($useFolder);
+		$this->overwriteService('RootFolder', $rootFolder);
+
+		$manager = $this->createMock(IUserManager::class);
+		$session = $this->createMock(ISession::class);
+		$user = $this->createMock(IUser::class);
+		$user->method('getUID')
+			->willReturn('foo');
+
+		$session->method('get')
+			->with('user_id')
+			->willReturn('foo');
+
+		$manager->method('get')
+			->with('foo')
+			->willReturn($user);
+
+		$userClassEvent = new GenericEvent($user);
+		$userFirstEvent = new GenericEvent($user);
+
+		$eventDispatcher->method('dispatch')
+			->withConsecutive(
+				[$this->equalTo(IUser::class . '::firstLogin'), $this->equalTo($userClassEvent)],
+				[$this->equalTo('user.firstlogin'), $this->equalTo($userFirstEvent)]
+			);
+		$userSession = new Session($manager, $session, $this->timeFactory,
+			$this->tokenProvider, $this->config, $this->logger, $this->serviceLoader,
+			$this->userSyncService, $eventDispatcher);
+		$this->assertNull($userSession->prepareUserLogin(true));
 	}
 
 	public function testLoginValidPasswordEnabled() {
--- a/tests/lib/UtilTest.php
+++ b/tests/lib/UtilTest.php
@@ -436,6 +436,7 @@ class UtilTest extends \Test\TestCase {
 
 		\OC_Util::$scripts = [];
 		\OC_Util::$styles = [];
+		\OC::$server->getConfig()->deleteSystemValue('skeletondirectory');
 	}
 
 	public function testAddScript() {


jvillafanez:
I'm going to ping @pmaier1 to clarify what is the expected behaviour here. For me it isn't clear if the rules should only apply when the users access to files, or what components the firewall is expected to affect.

If the firewall is expected to affect just the accesses to the files, I expect the users to be able to login normally, although the skeleton files might be missing.

sharidas:
- [ ] Fix the unit test errors caused.

pako81:
In my opinion, with this PR the behavior will be not consistent with how the FF currently works: allow users to log in but prevent any files operation.

pako81:
discussed with @sharidas - we agreed that we should try to stay consistent with the current behavior of the FF app, therefore allowing log in for the first-time-login users. Since the issue here seems to be related to find a way to recognize skeleton files/folders and allow them to be copied into the user's data dir, I guess we do have two options now:

1.  allow log in for first-time-login users but do not copy any skeleton files/folders. First-time-login users will be presented with an empty page

2. try to find a way for the FF app to allow copy of skeleton files/folders for rirst-time-login users, which may be complex, according to @sharidas code analysis.

@pmaier1 we need a PM decision here, please.

jvillafanez:
We can try the following:
1. Create a `Local` storage pointing to where the skeleton files are.
2. In the utility function we're using to copy the skeleton files, use the `copyFromStorage` function. I expect this function to copy the files recursively.
3. At this point, the firewall wrapper should be able to check if the operation is copying the skeleton files or not.

Key point is to investigate if it's possible to use raw storages in that utility function. We could create a new function instead of replacing the current one, but we still have the problem of getting the storage from somewhere.

sharidas:
- While looking into the https://github.com/owncloud/core/pull/35641#issuecomment-505810096, I am currently investigating the change set required when encryption is enabled for core.

sharidas:
- [ ] Add unit tests

lib/private/Files/Storage/Local.php
```diff
@@ -467,4 +471,35 @@ public function moveFromStorage(\OCP\Files\Storage $sourceStorage, $sourceIntern
...
+
+	/**
+	 * @param \OCP\Files\Storage $sourceStorage
+	 * @param null $dirToAppend null when nothing to append to home dir else the directory name
+	 * @throws ForbiddenException
+	 * @throws \OCP\Files\StorageNotAvailableException
+	 * @throws \OCP\Lock\LockedException
+	 * @throws \OC\ServerNotAvailableException
+	 */
+	public function copySekeletonFromStorage(\OCP\Files\Storage $sourceStorage, $dirToAppend = null) {
```
jvillafanez: No need to create new methods.
Note that you would need to spread the changes to the storage interface, to all the implementations, and to all the wrappers. This is out of scope

lib/private/legacy/util.php
```diff
@@ -354,32 +354,6 @@ public static function getUserQuota($userId) {
...
 	}
 
-	/**
-	 * copies the skeleton to the users /files
-	 *
-	 * @param String $userId
-	 * @param \OCP\Files\Folder $userDirectory
-	 * @throws \OC\HintException
-	 */
-	public static function copySkeleton($userId, \OCP\Files\Folder $userDirectory) {
```
jvillafanez: Do not remove the method. Just add a deprecation notice and point to the alternative method for now.

jvillafanez (changes requested):

lib/private/Files/Storage/Local.php
sharidas: I tried with a new approach by not having new methods in the storage ( both `Local` and `Encryption`). I was bit doubtful if this approach would work. Because the source storage I have used is `Local` for the skeleton. Here is my analysis when we don't use new methods in the storage and when we use:

#### Without new methods in storage 
When we use this approach, any call to create folder or file will land up in https://github.com/owncloud/firewall/blob/master/lib/storagewrapper.php. Even a call to `copyFromStorage` will be prohibitted. Also note that while testing I have enabled firewall rule for this user.

#### With new methods in storage
When we use this approach, we have a new method in the storage. The advantage here is we are bypassing the check https://github.com/owncloud/firewall/blob/master/lib/storagewrapper.php#L57. Because the storage is different now. That way we can get things working as expected.

I also tried with a different approach, take the `datadirectory` and append `$userFolder->getPath()` to get the absolute path. And then make targetSource as `Local` where the files of skeleton will be copied. That will not work with encryption. The files were not encrypted, though they are copied to the user folder.

sharidas: Done

lib/private/legacy/util.php
sharidas: Reverted the change and the method is not removed.

jvillafanez (changes requested):

lib/private/User/Session.php
```diff
@@ -466,6 +479,24 @@ public function prepareUserLogin($firstTimeLogin = false) {
...
+	private function copySkeleton(\OCP\Files\Storage $sourceStorage, Folder $userFolder) {
+		$sourceDir = \explode('::', $sourceStorage->getId())[1];
+		$sourceDirHandler = \opendir($sourceDir);
+		while (($file = \readdir($sourceDirHandler)) !== false) {
+			if (!OC\Files\Filesystem::isIgnoredDir($file)) {
+				$userFolder->getStorage()->copyFromStorage($sourceStorage, $file, "files/$file");
+			}
+		}
+
+		$userFolder->getStorage()->getScanner()->scan('', OC\Files\Cache\Scanner::SCAN_RECURSIVE);
```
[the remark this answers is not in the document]
sharidas: Yes this scan is required.

lib/private/User/Session.php
```diff
@@ -466,6 +479,24 @@ public function prepareUserLogin($firstTimeLogin = false) {
 		}
 	}
 
+	/**
+	 * @param \OCP\Files\Storage $sourceStorage
+	 * @param Folder $userFolder
+	 * @throws NotFoundException
+	 * @throws StorageNotAvailableException
+	 */
+	private function copySkeleton(\OCP\Files\Storage $sourceStorage, Folder $userFolder) {
```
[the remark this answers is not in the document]
sharidas: Moved to legacy/utils.php file.

lib/private/User/Session.php
```diff
@@ -466,6 +479,24 @@ public function prepareUserLogin($firstTimeLogin = false) {
...
+		$sourceDir = \explode('::', $sourceStorage->getId())[1];
```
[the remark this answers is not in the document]
sharidas: Fixed.

lib/private/User/Session.php
```diff
@@ -458,6 +467,10 @@ public function prepareUserLogin($firstTimeLogin = false) {
 			} catch (HintException $hintEx) {
 				// only if Skeleton no existing Dir
 				$this->logger->error($hintEx->getMessage());
+			} catch (StorageNotAvailableException $ex) {
```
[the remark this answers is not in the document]
sharidas: Done

lib/private/User/Session.php
```diff
@@ -444,7 +444,13 @@ public function prepareUserLogin($firstTimeLogin = false) {
 
 			try {
 				// copy skeleton
-				\OC_Util::copySkeleton($user, $userFolder);
+				//Set local storage
+				$skeletonDir = $this->config->getSystemValue('skeletondirectory', \OC::$SERVERROOT . '/core/skeleton');
+				$storageArgs = [
+					'datadir' => $skeletonDir
+				];
+				$localStorage = new OC\Files\Storage\Local($storageArgs);
```
jvillafanez: I'm not sure where the storage creation should be. As things are right now, you can copy any storage which isn't related to the skeleton directory with this `copySkeletonUsingStorage` function, so it doesn't seem a good name.
Overall, I think it's better to move the the storage creation inside the method. The method can receive a string parameter that will be interpreted as the skeleton location.

jvillafanez:
I think we should decide between having an easy to use method interface, but without easy testing, or having a testable method.
Main problem is that it's a static method, and whatever method that uses that static method won't be testable.

Taking into account that, in order to make the code testable, we'll have to inject a storage factory and get a mount point somewhere (in order to generate the storage using the factory), probably the first option is a better choice.

lib/private/legacy/util.php
```diff
@@ -354,12 +355,42 @@ public static function getUserQuota($userId) {
 		return OC_Helper::computerFileSize($userQuota);
 	}
 
+	/**
+	 * Copies the skeleton to user/files using the storage
+	 *
+	 * @param \OCP\Files\Storage $sourceStorage
+	 * @param Folder $userFolder
+	 */
+	public static function copySkeletonUsingStorage(\OCP\Files\Storage $sourceStorage, Folder $userFolder) {
```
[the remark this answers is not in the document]
jvillafanez: check the options below because we might opt for an easier method signature.

lib/private/legacy/util.php
```diff
@@ -354,12 +355,42 @@ public static function getUserQuota($userId) {
...
+		$sourceDir = $sourceStorage->getSourcePath('');
+		$sourceDirHandler = \opendir($sourceDir);
```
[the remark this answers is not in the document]
sharidas: used ownCloud's storage related function for `opendir`.

lib/private/legacy/util.php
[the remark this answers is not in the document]
sharidas: Added optional parameter logger. The logger is now passed from `\OC\User\Session::prepareUserLogin` where it is called from. Now the method signature is compacted to `\OC_Util::copySkeletonUsingStorage($userFolder)`.

sharidas:
> Taking into account that, in order to make the code testable, we'll have to inject a storage factory and get a mount point somewhere (in order to generate the storage using the factory), probably the first option is a better choice.

The method is now compacted. Let me know if this looks ok.

lib/private/User/Session.php
sharidas: Storage creation is now moved inside method `copySkeletonUsingStorage`.

lib/private/legacy/util.php
```diff
@@ -354,12 +355,51 @@ public static function getUserQuota($userId) {
 		return OC_Helper::computerFileSize($userQuota);
 	}
 
+	/**
+	 * Copies the skeleton to user/files using the storage
+	 *
+	 * @param \OCP\Files\Storage $sourceStorage
+	 * @param Folder $userFolder
+	 */
+	public static function copySkeletonUsingStorage(Folder $userFolder, \OCP\ILogger $logger = null) {
```
jvillafanez: Taking into account that we won't be able to unittest this method properly, I think we can remove the logger parameter and let the function get it from the `\OC::$server` directly.

lib/private/legacy/util.php
```diff
@@ -354,12 +355,51 @@ public static function getUserQuota($userId) {
...
+			'datadir' => $skeletonDir
+		];
+		$sourceStorage = new OC\Files\Storage\Local($storageArgs);
+		$sourceDirHandler = $sourceStorage->opendir('');
+		while (($file = \readdir($sourceDirHandler)) !== false) {
+			if (!OC\Files\Filesystem::isIgnoredDir($file)) {
+				try {
+					$userFolder->getStorage()->copyFromStorage($sourceStorage, $file, "files/$file");
+				} catch (\OCP\Files\NotFoundException $ex) {
+					\OC::$server->getLogger()->logException($ex);
```
jvillafanez: Use the same logger instance in the whole function

lib/private/legacy/util.php
```diff
@@ -354,12 +355,51 @@ public static function getUserQuota($userId) {
...
+	}
+
 	/**
 	 * copies the skeleton to the users /files
 	 *
 	 * @param String $userId
 	 * @param \OCP\Files\Folder $userDirectory
 	 * @throws \OC\HintException
+	 *
+	 * Note: Another method to acheive the same behaviour is using copySkeletonUsingStorage
```
jvillafanez: I think we should use the `@deprecated` anotation and point to the new method, otherwise people won't likely use the new method

lib/private/legacy/util.php
sharidas: Done.

lib/private/legacy/util.php
sharidas: Done.

lib/private/legacy/util.php
sharidas: Updated with `@deprecated` I was bit confused to use `@deprecated` or not.

tests/lib/UtilTest.php
```diff
@@ -436,6 +436,7 @@ protected function tearDown() {
 
 		\OC_Util::$scripts = [];
 		\OC_Util::$styles = [];
+		\OC::$server->getConfig()->deleteSystemValue('skeletondirectory');
```
jvillafanez: Is this needed? We shouldn't touch anything outside of the tests.

tests/lib/User/SessionTest.php
```diff
@@ -194,6 +203,52 @@ public function testSetUser() {
...
+		$session->method('get')
+			->with('user_id')
+			->willReturn('foo');
+
+		$manager->method('get')
+			->with('foo')
+			->willReturn($user);
+
+		$this->eventDispatcher->addListener('user.firstlogin', function (GenericEvent $event) use ($user) {
+			$this->assertEquals($event->getSubject()->getUID(), $user->getUID());
```
jvillafanez: I don't really like this... what happens if the event isn't triggered? The test should fail.
It seems a better choice to mock the event dispatcher and assert that it dispatches both events with the expected information.

tests/lib/User/SessionTest.php
```diff
@@ -194,6 +203,52 @@ public function testSetUser() {
...
+			->willReturn($useFolder);
+		$this->overwriteService('RootFolder', $rootFolder);
+
+		$this->config->method('getSystemValue')
+			->willReturn('/foo/bar');
+		$manager = $this->createMock(IUserManager::class);
+		$session = $this->createMock(ISession::class);
+		$user = $this->createMock(IUser::class);
+
+		$this->config->method('getSystemValue')
```
jvillafanez: I don't think you can mock the same method twice. I don't know where this is needed.

tests/lib/UtilTest.php
sharidas: Yes this is required. Its not failing because of any change that we made in the test. Instead, its failing because other tests try to access our new code and it finds the stale changes in the config.php. So I thought its better to clean this up.

tests/lib/User/SessionTest.php
sharidas: Removed.

tests/lib/User/SessionTest.php
sharidas: Yah, event dispatcher is now mocked.

tests/lib/UtilTest.php
jvillafanez: But I don't see any change in the configuration in this set of changes. The code must work with any possible skeleton directory, so, at least, this change shouldn't be needed for us.
I expect that the master branch has the tests passing. If this is the case, again this change isn't needed; if not, the problem should be fixed in another PR.
If there are other tests failing unrelated to this change set, maybe the problem is either on those tests (doing things that they shouldn't do) or on our changes (as said, code must work with any skeleton directory)

phil-davis:
https://drone.owncloud.com/owncloud/core/19160/293
```
--- Failed scenarios:

    /drone/src/tests/acceptance/features/apiMain/checksums.feature:288
    /drone/src/tests/acceptance/features/apiMain/checksums.feature:289

88 scenarios (86 passed, 2 failed)
589 steps (585 passed, 2 failed, 2 skipped)
```

The scenario is:
```
  Scenario Outline: Uploading a file with invalid SHA1 checksum overwriting an existing file
    Given using <dav_version> DAV path
    When user "user0" uploads file "filesForUpload/textfile.txt" to "/textfile0.txt" with checksum "SHA1:f005ba11f005ba11f005ba11f005ba11f005ba11" using the WebDAV API
    Then as user "user0" the webdav checksum of "/textfile0.txt" via propfind should match "SHA1:0c1d334e686d1039c9ead0dbc047f02dbf696be8 MD5:d991cd854c53729d066c6ed5e34bcda3 ADLER32:8685092b"
    And the content of file "/textfile0.txt" for user "user0" should be "ownCloud test text file 0" plus end-of-line
    Examples:
      | dav_version |
      | old         |
      | new         |
```

In the old way of creating the skeleton files, a checksum was calculated and stored for each file. So `textfile0.txt` already has a checksum stored when it is copied from the skeleton.
In the new way that this PR does it, there seems to be no checksum calculated and stored while creating skeleton files.
The upload-overwrite purposely fails (because a wrong checksum is given). 
The next step does a propfind on /textfile0.txt to find out what is its checksum and verify that it is the expected value. But it has no checksum yet, so the propfind returns an empty checksum.

If I make the scenario start like:
```
  Scenario Outline: Uploading a file with invalid SHA1 checksum overwriting an existing file
    Given using <dav_version> DAV path
    And user "user0" downloads file "/textfile0.txt" using the WebDAV API
```
then it passes, because the download triggers code that calculates and stores the checksum in the backend.

Do we want to preserve the existing behavior, that for a new user the just-copied-in-place skeleton files all have a checksum stored?

sharidas:
With the new way of copying skeletons using storages, the codeflow doesn't reach checksum wrappers. And hence the checksums are not calculated. But with the same idea we can acheive this by modifying the storage methods to know its a skeleton copy that is going to happen and subsequently modifying the calls to `fopen` ( not the read mode ) and so on. But I am wondering is it really worth the effort for this PR? 

Note:When we use methods like `file_put_contents` of the user storage ( or `userFolder` as in the PR ), the checksums are updated. Unfortunately for the method `copyFromStorage` that's not the case. 

Feedback here would be nice.

micbar:
@DeepDiver1975 Could you take a look?

DeepDiver1975:
> Note:When we use methods like `file_put_contents` of the user storage ( or `userFolder` as in the PR ), the checksums are updated. Unfortunately for the method `copyFromStorage` that's not the case.

The checksum wrapper is not implementing ```copyFromStorage``` - could that be the reason?

DeepDiver1975:
> ## Motivation and Context
> 
> If the user is prohibitted to login by a firewall rule, then lets not prohibit the login. Let the rule apply to the user, but login should not be prevented.

I question that this is the correct approach ....

sharidas:
> 
> I question that this is the correct approach ....

You are right, this is not the correct approach. This was an approach taken first time ( when the PR was created), then changed. But forgot to update the `Motivation context`. Sorry for that :(

sharidas:
@DeepDiver1975 is it ok to have skeleton files copied without having checksum updated in the table?

> The checksum wrapper is not implementing copyFromStorage - could that be the reason?

Yes, the checksum wrapper does not implement copyFromStorage. Should we focus on implementing the same in the checksum wrapper? I was wondering if skeleton files which are copied without checksum updated, do they cause any problem with the clients like desktop or mobile for example. If nothing is impacted as such, then the existing solution in the PR, would help us, right? Or am I completely wrong?

sharidas:
Closing this issue because inorder to fix the checksum for skeleton files, we would have to touch the checksum wrapper. And it seems too much to implement for the gain. Hence closing this issue.

Issue #35658, opened 2019-06-25
drone CleanupRemoteStoragesTest intermittent fail


phil-davis:
We are getting an intermittent fail in drone PHPunit tests:
e.g.
https://drone.owncloud.com/owncloud/core/18820/209 `stable10` with oracle DB

```
There was 1 failure:

1) OCA\Files_Sharing\Tests\Command\CleanupRemoteStoragesTest::testCleanup
Expectation failed for method name is equal to 'writeln' when invoked at sequence index 1
Parameter 0 for invocation Symfony\Component\Console\Output\OutputInterface::writeln('6 remote share(s) exist', 0) does not match expected value.
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'5 remote share(s) exist'
+'6 remote share(s) exist'

/drone/src/apps/files_sharing/lib/Command/CleanupRemoteStorages.php:67
/drone/src/apps/files_sharing/tests/Command/CleanupRemoteStoragesTest.php:183
```

I can run the unit test locally without error.

phil-davis:
Other examples:
https://drone.owncloud.com/owncloud/core/18817/212 `master` MariaDB
https://drone.owncloud.com/owncloud/core/18813/137 `master` mysql
https://drone.owncloud.com/owncloud/core/18810/162 `stable10` postgreSQL

and so on.

There seems to be no particular branch or database that causes the problem.

Some timing issue with that unit test case?

@micbar it is rather annoying for developers to get a CI pass. Who could look at this?

phil-davis:
Demonstration in PR #35666 

Example of the problem:
https://drone.owncloud.com/owncloud/core/18860/127
```
............................................................. 9150 / 9678 ( 94%)
.............................Farray(1) {
  [9]=>
  string(40) "shared::439de3b339c48d6c88064ac93a3615c4"
}
array(1) {
  [9]=>
  string(31) "shared::token6@http://localhost"
}
............................... 9211 / 9678 ( 95%)
............................................................. 9272 / 9678 ( 95%)
............................................................. 9333 / 9678 ( 96%)
............................................................. 9394 / 9678 ( 97%)
............................................................. 9455 / 9678 ( 97%)
............................................................. 9516 / 9678 ( 98%)
............................................................. 9577 / 9678 ( 98%)
............................................................. 9638 / 9678 ( 99%)
........................................                      9678 / 9678 (100%)

Time: 18.87 minutes, Memory: 4088.00MB

[...]

PR #35665, opened 2019-06-26
Load PHPunit app tests in alphabetical order

phil-davis:
## Description
When finding the unit tests for the built-in core apps, use `scandir()` so that the relevant folders and files are discovered and loaded in a predictable order - in this case it will be alphabetical order.

At least this will mean that the unit test runs will more reliably either pass or fail. There will not be the potential for "random" failures due to the order of test case execution.

## Related Issue
#35658 

## Motivation and Context
We are getting intermittent unit test fails in CI. Actually those fails are because `apps/files_sharing/tests/External/ManagerTest.php` `testAddShareAccepted` is not cleaning up the share that it creates.

If that happens to run before `apps/files_sharing/tests/Commands/CleanupRemoteStoragesTest` then `CleanupRemoteStoragesTest` fails because its environment starts out with an extra unexepected entry in `share_external` table.

`apps/files_sharing/tests/External/ManagerTest.php` can be fixed in a separate PR.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/apps.php
+++ b/tests/apps.php
@@ -13,15 +13,14 @@ function loadDirectory($path) {
 	if (\strcasecmp(\basename($path), 'ui') === 0) {
 		return;
 	}
-	if ($dh = \opendir($path)) {
-		while ($name = \readdir($dh)) {
-			if ($name[0] !== '.') {
-				$file = $path . '/' . $name;
-				if (\is_dir($file)) {
-					loadDirectory($file);
-				} elseif (\substr($name, -4, 4) === '.php') {
-					require_once $file;
-				}
+	$dirEntries = \scandir($path);
+	foreach ($dirEntries as $name) {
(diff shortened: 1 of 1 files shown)


phil-davis:
CI has got some of the out-of-memory errors that we have also been getting the last few days.
https://drone.owncloud.com/owncloud/core/18859/175
https://drone.owncloud.com/owncloud/core/18859/279

I will restart and see if we can get lucky with those.

phil-davis:
drone is happy - great!
codecov status did ot come - no problem, somebody can override and merge this if it is approved.

phil-davis:
Backport `stable10` #35669

PR #35695, opened 2019-06-28
Trigger CI

phil-davis:
because there seems to be some problem with `phan` on PHP 7.2 (and 7.3)


phil-davis:
phan is now disabled for PHP 7.2 and 7.3

PR #35716, opened 2019-07-01
[stable10] Webdav TrashBin API

DeepDiver1975:
backport #35636

### DIFF
--- a/apps/dav/lib/TrashBin/TrashBinPlugin.php
+++ b/apps/dav/lib/TrashBin/TrashBinPlugin.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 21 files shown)


individual-it (approved):

apps/dav/lib/TrashBin/TrashBinPlugin.php
```diff
@@ -0,0 +1,66 @@
...
+
+		$propFind->handle(self::TRASHBIN_ORIGINAL_LOCATION, static function () use ($node) {
+			return $node->getOriginalLocation();
+		});
+
+		$propFind->handle(self::TRASHBIN_DELETE_TIMESTAMP, static function () use ($node) {
+			return $node->getDeleteTimestamp();
+		});
+
+		$propFind->handle(self::TRASHBIN_DELETE_DATETIME, static function () use ($node) {
```
butonic: I think we can leave out datetime. the delete timestamp by definition (= unix timestamp) is in the GMT timezone ... or at least should be. and everything I know can digest unix timestamps. we can add it, but should recommend using the timestamp.

butonic (approved):

apps/dav/lib/TrashBin/TrashBinPlugin.php
DeepDiver1975: unix timestamps will limit us to the interval of unix timestamps 01.01.1970-19.1.2038(int32 based)
Using a real date time format is not limited to this. 

furthermore the timezone is fixed/not part of the transmission - using a real date time format can support this (even if not used as of today)

DeepDiver1975: and as discussed: http://apiux.com/2013/03/20/5-laws-api-dates-and-times/ advised to use ISO-8601 - we use http://tools.ietf.org/html/rfc7231#section-7.1.1.1 to stay consistent with the getlastmodified property

butonic: that makes sense ... nevertheless, unix timestamps are supposed to be GMT / UTC but that is jist screaming for errors ;-)

PR #20439, opened 2015-11-10
Take submount etag into account for folder etags

icewind1991:
This removes the need to do cross-storage etag propagation

PR #35718, opened 2019-07-01
Fix share permissions combination tests

phil-davis:
## Description
See what breaks here - hopefully there are unit tests and other acceptance tests that show why just commenting out this code is a bad idea.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -1419,7 +1419,7 @@ class View {
 			}
 
 			if ($mount instanceof MoveableMount && $internalPath === '') {
-				$data['permissions'] |= \OCP\Constants::PERMISSION_DELETE;
+				//$data['permissions'] |= \OCP\Constants::PERMISSION_DELETE;
 			}
 
 			$owner = $this->getUserObjectForOwner($storage->getOwner($internalPath));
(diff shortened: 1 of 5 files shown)


lib/private/Files/View.php
```diff
@@ -1419,7 +1419,7 @@ public function getFileInfo($path, $includeMountPoints = true) {
 			}
 
 			if ($mount instanceof MoveableMount && $internalPath === '') {
-				$data['permissions'] |= \OCP\Constants::PERMISSION_DELETE;
+				//$data['permissions'] |= \OCP\Constants::PERMISSION_DELETE;
```
phil-davis: This was added in https://github.com/owncloud/core/pull/20439/files#diff-d9747e68e80ce288dcea911cb8607546R1250 some years ago.

phil-davis:
The issue here is about unsharing.

When anyone receives a share, they have the ability to "delete" the received file or folder. If they do that, then they are just deleting their received share of the item. The original item stays in existence for the owner.

For file shares, this does not cause a problem. A file cannot be shared with "real" delete access anyway. The "implied" delete access is "unshare" access.

For folder shares, the "implied" delete access for the shared folder itself is good - that helps the code to let the share receiver unshare the folder. But for items inside the folder, "implied" delete access should not happen, which I think is OK (I will check that with some scenarios). And "implied" delete access should not be allowed to propagate in a reshare to become an "actual" delete access.

phil-davis:
Closing, this was testing/PoC

PR #35724, opened 2019-07-02
[stable10] php73 check tests

phil-davis:
Run many of the drone unit and acceptance test jobs with PHP 7.3 to double-check that they pass.

drone had to be adjusted so that the federated server, which runs the `stable10` QA tarball - would run PHP 7.2. Because that QA tarball is the current `stable110` that only officially supports PHP 7.2. Actually that is a good thing to test. We know that a PHP 7.3 system with "newer" code can talk OK to a PHP 7.2 system.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -642,7 +642,7 @@ matrix:
 
   # frontend
     - TEST_SUITE: javascript
-      PHP_VERSION: 7.1
+      PHP_VERSION: 7.3
       COVERAGE: true
 
   # owncloud-coding-standard
@@ -658,7 +658,7 @@ matrix:
       PHP_VERSION: 7.1
 
   # Litmus
-    - PHP_VERSION: 7.1
+    - PHP_VERSION: 7.3
       USE_SERVER: true
       SERVER_PROTOCOL: https
       TEST_SUITE: litmus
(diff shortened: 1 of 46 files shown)


phil-davis:
See note https://github.com/owncloud/core/pull/34559#issuecomment-507499102 about the fails reported when trying to generate codecov with PHP 7.3

phil-davis:
2nd run of CI gets the same sort of errors trying to produce codecov. e.g.
https://drone.owncloud.com/owncloud/core/19099/134
```
OK, but incomplete, skipped, or risky tests!
Tests: 10810, Assertions: 56509, Skipped: 205.

Generating code coverage report in Clover XML format ... done
[PHP Fatal error:  Uncaught ErrorException: Unknown: Cannot destroy the zip context: Can't remove file: No such file or directory in Unknown:0
Stack trace:
#0 [internal function]: OCA\Files\Command\Scan->exceptionErrorHandler(2, 'Unknown: Cannot...', 'Unknown', 0, NULL)
#1 {main}
  thrown in Unknown on line 0] 
```
See comment https://github.com/owncloud/core/pull/34559#issuecomment-507499102

phil-davis:
This PR has done its job demonstrating tests with PHP 7.3

PR #32203, opened 2018-07-31
Remove $CLASSPATH - this is no longer needed. Apps are using proper n…

DeepDiver1975:
…amespaces today.

## Description
$CLASSPATH as  a very old concept to perform auto loading in ownCloud.

As of today all apps use a proper namespace e.g. OCA\Calendar\ to load classes living in the calendar app.

Furthermore composer is more and more used and has a far better autoloaded then ownCloud itself.

## Issues

closes https://github.com/owncloud/core/issues/12573

## Motivation and Context
Deprecate/remove old concepts 

## How Has This Been Tested?
- manually 
- unit tests

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35755, opened 2019-07-03
[stable10] Remove $CLASSPATH and move autoloader to namespace

phil-davis:
Backport 
#32203 Remove $CLASSPATH - this is no longer needed. Apps are using proper namespaces today.
#32204 move autoloader to namespace

These resolve the remaining real differences between `composer.json` in `master` and `stable0`

After this PR, the diff is just:
```
diff --git a/composer.json b/composer.json
index 66190d3ba3..1948f7af74 100644
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
         "optimize-autoloader": true,
         "classmap-authoritative": false,
         "platform": {
-            "php": "7.0.8"
+            "php": "7.1"
         }
     },
     "autoload" : {
@@ -30,7 +30,7 @@
         "roave/security-advisories": "dev-master"
     },
     "require": {
-        "php": ">=7.0.8",
+        "php": ">=7.1",
         "doctrine/dbal": "^2.5",
         "phpseclib/phpseclib": "^2.0",
         "rackspace/php-opencloud": "v1.9.2",
```
which is only that `stable10` still supports PHP 7.0

Motivation: get rid of more diffs between `master` and `stable10`


phil-davis:
Note: this is on top of PR #35752 because that PR is also adjusting similar files in `stable10` to match `master`

DeepDiver1975 (approved):

PR #35774, opened 2019-07-07
[stable10] Introduce phpstan

phil-davis:
## Description
1) Backport of various commits from PR #31810 
2) Then fix other stuff in `stable10` that `phpstan` reports.

Also fixes #35773 

## Motivation and Context
When trying some things for "drop PHP  7.0" and for PHPunit7 I noticed some problems and ended up in some of the same code that was corrected/adjusted by `master` PR #31810 

Issue 

## How Has This Been Tested?
Local run of:
```
make test-php-phpstan
```

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #33209, opened 2018-10-17
Remove not longer valid update script - pre-8.1 upgrade  ....

DeepDiver1975:
## Description
This update script is no longer required. It was to upgrade a pre-8.1 owncloud

## How Has This Been Tested?
:robot: 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

PR #35781, opened 2019-07-08
[stable10] Remove not longer valid update script - pre-8.1 upgrade

phil-davis:
Backport #33209

I noticed this when backporting "Introduce PHPstan" #35774 - in that PR I had to exclude this file from analysis because:
```
php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan analyse --memory-limit=2G --configuration=./phpstan.neon --level=0 apps core settings lib/private lib/public ocs ocs-provider
 1392/1392 

 ------ -------------------------------------------------------------------------------------------------------------------------------- 
  Line   apps/files/appinfo/update.php                                                                                                   
 ------ -------------------------------------------------------------------------------------------------------------------------------- 
  52     Function owncloud_reset_encrypted_flag not found while trying to analyse it - autoloading is probably not configured properly.  
  72     Function owncloud_reset_encrypted_flag not found.                                                                               
  79     Function owncloud_reset_encrypted_flag not found.                                                                               
 ------ -------------------------------------------------------------------------------------------------------------------------------- 
```

It is "ordinary" PHP, not a class.

Rather than trying to sort all that out, delete it.
Deleted code is debugged code.


phil-davis:
@DeepDiver1975 does it seem OK to delete this for `10.3` ?
(rather than in a major version 11)

DeepDiver1975 (approved):

PR #34936, opened 2019-03-30
[stable10] future drop php 7.0

phil-davis:
## Description
This draft PR is a place to put things that are waiting to go into `stable10` but that require PHP7.1+ and so cannot be merged until PHP7.0 support can be dropped.

- #28043 Increase version to PHP7.1 (the remaining commits from that PR that relate to PHP7.1 and thus have never been backported yet)
- Set min PHP version to 7.1 (1)
- PHP 7.1+ is required (2)
- "[Workaround] .... until sebastianbergmann/phpunit-mock-objects#402 is merged and released" https://github.com/owncloud/core/pull/28043/commits/407a70d90dd2692fab8f79cbc28014cff37a1995 (I think this will be needed as long as we are on phpunit6, because that `phpunit-mock-objects` change never got into a release with dependencies for phpunit6) (3)
- (other commits from #28043 are already back in `stable10`)
- #32065 2nd commit "Remove our own sqlite autoincrement hack" (needed with `doctrine/dbal` version bump that comes when composer PHP 7.1 happens) (4)
- Adjust drone matrix to match what is currently being done in master (actual commits from master would be from 2 years of travis and drone times!) (5)
- Disable phpstan in drone until we can sort out code fix backporting (see #35774 that is waiting for review) (6)
- Do not test postgres:10.3 - needs more backports (I will look for what is needed for this) (7)
- Fix ExceptionPlugin.php logException like in master (this is fixed with "Introduce phpstan") (8)
That stops the warning:
```
[PHP Warning:  Declaration of OCA\DAV\Tests\unit\Connector\Sabre\RequestTest\ExceptionPlugin::logException(Exception $ex) should be compatible with OCA\DAV\Connector\Sabre\ExceptionLoggerPlugin::logException(Throwable $ex) in /drone/src/apps/dav/tests/unit/Connector/Sabre/RequestTest/ExceptionPlugin.php on line 26]
```

## Motivation and Context
Have a place to consolidate backports that are "blocked" waiting for PHP7.0 to be dropped.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
[...]

PR #33255, opened 2018-10-22
Add mariadb 10.3 + mysql 8 to Drone, remove 10.2 workarounds

PVince81:
Add mariadb 10.3 to the unit test matrix.

For https://github.com/owncloud/core/issues/29483 but doesn't cover mb4 AFAIK

PR #35785, opened 2019-07-09
[stable10] drone mariadb 10.3

phil-davis:
## Description
On top of #34936 [stable10] future drop php 7.0

Backport #33255 Add mariadb 10.3 + mysql 8 to Drone, remove 10.2 workarounds

And see if we get CI passing

## How Has This Been Tested?
CI


.drone.yml
```diff
@@ -737,6 +737,13 @@ matrix:
       INSTALL_SERVER: true
       INSTALL_TESTING_APP: true
 
+    - PHP_VERSION: 7.1
```
phil-davis: note: this is already in an earlier commit from "future drop php 7.0" - so at the moment we get these mariadb:10.3 unit tests run twice. No harm in that!

.drone.yml
```diff
@@ -688,65 +692,94 @@ matrix:
...
 
     - PHP_VERSION: 7.1
-      DB_TYPE: sqlite
+      DB_TYPE: mariadb
+      # mb4 support by default
+      MARIADB_IMAGE: mariadb:10.3
       TEST_SUITE: phpunit
       COVERAGE: true
       INSTALL_SERVER: true
-      INSTALL_TESTING_APP: true
```
phil-davis: this line should not have been deleted.
That caused the fails in https://drone.owncloud.com/owncloud/core/19318/172

phil-davis:
It looks like this can work. See #35786 for a version with commits cherry-picked in a better order etc.

PR #35822, opened 2019-07-15
Bugfix/not load groups if not necessary

DeepDiver1975:
_please note this targets stable10 already - I will not create PRs for master anymore due to your plans to kill master branch_

## Description
Even if no read only groups are defined (~99,9% of all instance) all groups for the current user will be loaded.
In case of LDAP this means loading the groups from there and this has a major performance impact.

## How Has This Been Tested?
- :hand:

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -243,15 +243,17 @@ class OC_Util {
 			if (!\is_array($readOnlyGroups)) {
 				$readOnlyGroups = [];
 			}
+			$readOnlyGroupMemberships = [];
+			if ($readOnlyGroups) {
+				$userGroups = \array_keys(
+					\OC::$server->getGroupManager()->getUserIdGroups($user)
+				);
 
-			$userGroups = \array_keys(
-				\OC::$server->getGroupManager()->getUserIdGroups($user)
-			);
-
-			$readOnlyGroupMemberships = \array_intersect(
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
smb_windows tests are failing -> ignore

lib/private/legacy/util.php
```diff
@@ -243,15 +243,17 @@ public static function setupFS($user = '') {
...
-			$userGroups = \array_keys(
-				\OC::$server->getGroupManager()->getUserIdGroups($user)
-			);
-
-			$readOnlyGroupMemberships = \array_intersect(
-				$readOnlyGroups,
-				$userGroups
-			);
+			$readOnlyGroupMemberships = [];
+			if ($readOnlyGroups) {
```
jvillafanez: it might be worthy to include a comment to remind that an empty array is considered "false" and, as such, we won't hit the group manager (typical PHP things)

jvillafanez (approved):

phil-davis:
> smb_windows tests are failing -> ignore

It got the "hung restore step in drone"  problem. Not even an smb_winddows issue.

I restated the drone job. That should get green CI.

PR #35825, opened 2019-07-15
Use human readable permission strings

### LINKED ISSUE #34176: do not use magic numbers for permissions in sharing steps
after https://github.com/owncloud/core/pull/34175 we can use human readable permissions in sharing steps, so we should refactor existing steps to make them more readable

skshetry:
Use human readable strings instead of numbers in permission. Part of issue #34176 

Still feature files need to be changed wherever we use tablenode to set permissions (though, some has been updated to check if it really works there).

Another thing to note, is, if we should also make `Then the fields of the last response should include` step aware of the permissions, or just check number in here to confirm?

Fixes https://github.com/owncloud/core/issues/34176


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -91,6 +91,39 @@ public function getServerLastShareTime() {
 		return (int) $this->lastShareData->data->stime;
 	}
 
+	/**
+	 * Split given permissions string each separated with + into array of strings
```
phil-davis:
```suggestion
	 * Split given permissions string each separated with "," into array of strings
```

phil-davis:
> Another thing to note, is, if we should also make `Then the fields of the last response should include` step aware of the permissions, or just check number in here to confirm?

IMO it will be good to refactor that also.

Similar other "then" steps are:
```
Then the fields of the last response should not include
Then the response when user "([^"]*)" gets the info of the last share should include
Then /^user "([^"]*)" should not be able to share (?:file|folder|entry) "([^"]*)" with (user|group) "([^"]*)"(?: with permissions (.*))? using the sharing API
Then /^user "([^"]*)" should be able to share (?:file|folder|entry) "([^"]*)" with (user|group) "([^"]*)"(?: with permissions (.*))? using the sharing API
```

phil-davis (approved):

paurakhsharma:
Backport on: #35841

skshetry:
Previously, we could use  `When user :arg1 shares file 'lorem.txt' with user 'user0' with permissions read using the sharing API` kind of step (notice the permission 'read' is without the quotes). 

This really should not be possible, and would be better to be inside the quotes. Please refactor those steps. 

```
--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When user :arg1 shares file :arg2 with user :arg3 with permissions read using the sharing API
```

In `getWebDAVSharePermissions.feature`, we get
```
Property "ocs:share-permissions" found with value "19", expected "#^3$#" or "#^3$#"
```

Probably, the permissions array (that we have parsed is passed to ocs:share-permissions` not the permissionsSum.


And, some problem with `updateShare.feature` which i didn't understand: 
```
        OCS status code is not the expected value
        --- Expected
        +++ Actual
        @@ @@
        -'100'
        +'400'
        OCS status code is not the expected value
        --- Expected
        +++ Actual
        @@ @@
        -'200'
        +'400'

```

phil-davis:
Restarted drone - master CI should be fixed now.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -91,6 +91,43 @@ public function getServerLastShareTime() {
 		return (int) $this->lastShareData->data->stime;
 	}
 
+	/**
+	 * Split given permissions string each separated with "," into array of strings
+	 * ToDo: we should be able to refactor this inside SharingHelper::getPermissionSum
```
skshetry: This was discussed on the team. We decided to handle parsing at the entry points, rather than the bottom most layer, i.e. `createShare` or `getPermissionSum`. This PR does not handle that, though (would have been better, but would take a lot of refactoring to do it right).

Again, `getPermissionSum` takes a lot of weight already  (it accepts `int`, `string`, `string[]`, `int[]` and even `null`). Adding to that, parsing/exploding strings or even array of strings did not seem right to me. It's better to parse data at the top most layer, and keep on passing parsed value around. And, we could always create helper functions that can generate permissionSum from array of strings and similarly, `string[]` from given `int`.

PR #35826, opened 2019-07-15
Add webui Acceptance test for trusted server in admin settings

dpakach:
## Description

Add webUI Acceptance test for trusted server in admin settings.

## Related Issue

- Related issue https://github.com/owncloud/QA/issues/22

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/WebUIAdminSharingSettingsContext.php
+++ b/tests/acceptance/features/bootstrap/WebUIAdminSharingSettingsContext.php
@@ -42,6 +42,12 @@ class WebUIAdminSharingSettingsContext extends RawMinkContext implements Context
 	private $webUIGeneralContext;
 
 	/**
+	 *
+	 * @var FeatureContext
+	 */
+	private $featureContext;
+
+	/**
 	 * WebUIAdminSharingSettingsContext constructor.
 	 *
 	 * @param AdminSharingSettingsPage $adminSharingSettingsPage
@@ -62,6 +68,7 @@ class WebUIAdminSharingSettingsContext extends RawMinkContext implements Context
 		$this->webUIGeneralContext->adminLogsInUsingTheWebUI();
 		$this->adminSharingSettingsPage->open();
 		$this->adminSharingSettingsPage->waitTillPageIsLoaded($this->getSession());
+		$this->webUIGeneralContext->setCurrentPageObject($this->adminSharingSettingsPage);
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


haribhandari07 (commented):
Looks good

tests/acceptance/features/webUISharingExternal/adminSharingSettings.feature
```diff
@@ -0,0 +1,28 @@
+@webUI @insulated @disablePreviews @TestAlsoOnExternalUserBackend
+Feature: Manage Trusted servers from the webUI
+  As a administrator
```
phil-davis:
```suggestion
  As an administrator
```

tests/acceptance/features/bootstrap/WebUIAdminSharingSettingsContext.php
```diff
@@ -271,6 +278,46 @@ public function theAdministratorEnablesAddServerAutomatically($action) {
...
+	 */
+	public function theAdministratorDeletesAsATrustedServerUsingTheWebui($url) {
+		$this->adminSharingSettingsPage->deleteTrustedServer(
+			$this->getSession(),
+			$this->featureContext->substituteInLineCodes($url)
+		);
+	}
+
+	/**
+	 * @Then a error message for trusted server should contain :text
```
phil-davis: Maybe:
```suggestion
	 * @Then a trusted server error message should be displayed on the webUI with the text :text
```

which is similar to `WebUIPersonalGeneralSettingsContext` 
```
Then a password error message should be displayed on the webUI with the text :wrongPasswordMessageText
```

tests/acceptance/features/lib/AdminSharingSettingsPage.php
```diff
@@ -374,6 +380,58 @@ public function toggleAutoAddServer(Session $session, $action) {
...
+	/**
+	 * Add trusted server
+	 *
+	 * @param Session $session
+	 * @param string $url
+	 *
+	 * @return AdminSharingSettingsPage
+	 */
+	public function addTrustedServer(Session $session, $url) {
+		$btn = $this->find("xpath", $this->addTrustedServerBtnXpath);
```
individual-it: need to have checks is find returns `null`
otherwise we will have ugly error messages

individual-it (changes requested):
:+1: just more checks if `find` returns `null`

individual-it (approved):

tests/acceptance/features/lib/AdminSharingSettingsPage.php
```diff
@@ -374,6 +380,78 @@ public function toggleAutoAddServer(Session $session, $action) {
...
+			"xpath",
+			\sprintf($this->deleteTrustedServerBtnXpath, $url)
+		);
+		$this->assertElementNotNull(
+			$btn,
+			__METHOD__ . "Could not find the button to delete trusted server."
+		);
+		$btn->click();
+		$this->waitForAjaxCallsToStartAndFinish($session);
+		return $this;
```
skshetry: Why does this return the whole page object? Used anywhere?

skshetry (dismissed):

tests/acceptance/features/lib/AdminSharingSettingsPage.php
dpakach: Not used currently but I dont think it is a good practice and also doesn't breaks anything.

dpakach:
Backport on #35842

tests/acceptance/features/lib/AdminSharingSettingsPage.php
individual-it: methods should return pageObjects https://github.com/SeleniumHQ/selenium/wiki/PageObjects

skshetry:
```
Because we're encouraging the developer of a test to try and think about the services that they're interacting with rather than the implementation, PageObjects should seldom expose the underlying WebDriver instance. To facilitate this, methods on the PageObject should return other PageObjects. This means that we can effectively model the user's journey through our application.
```

We don't chain the methods in our test code. We just use it to interact. Unless we do that, I think, we should stick to the usual standards (if there is) to return void.

phil-davis:
weird Javascript drone job issue. I restarted the build.

phil-davis:
Restarted drone - master CI should be fixed now.

PR #35836, opened 2019-07-16
Share item model attributes javascript api v2

mrow4a:
This PR:
- depreciates share attributes javascript api v1 which relied on "in-core registration of attributes" via `ShareItemModel.registerShareAttribute`
- new js attr api v2 allows apps to extend `ShareItemModel.updateShare` and `ShareItemModel.addShare` function to pass the custom 'attributes'. This allows for much more customizations to checkbozes instead of relying on core api.
- target 10.3? 

Tested matrix of v1/v2 with:
- richdocuments https://github.com/owncloud/richdocuments/pull/264
- onlyoffice 2.3.1
- onlyoffice https://github.com/ONLYOFFICE/onlyoffice-owncloud/pull/275
- richdocuments https://github.com/owncloud/richdocuments/pull/264 together with onlyoffice 2.3.1 on docx 
- richdocuments https://github.com/owncloud/richdocuments/pull/264 together with onlyoffice https://github.com/ONLYOFFICE/onlyoffice-owncloud/pull/275 on docx 

Screenshot at Jul 16 09-06-08

### DIFF
--- a/core/js/shareitemmodel.js
+++ b/core/js/shareitemmodel.js
@@ -65,6 +65,7 @@
 	 */
 
 	/**
+	 * ShareAttributesApi v1 registered attributes
 	 * @typedef {object} OC.Share.Types.RegisteredShareAttribute
 	 * @property {string} key
 	 * @property {bool}   default
@@ -75,6 +76,7 @@
 	 * @property {number[]} incompatiblePermissions
 	 * @property {number[]} requiredPermissions
 	 * @property {OC.Share.Types.ShareAttribute[]} incompatibleAttributes
+	 * @deprecated shareAttributesApi v2 requires apps to extend ShareItemModel with attributes
 	 */
 
 	/**
@@ -114,6 +116,7 @@
 
 		/**
 		 * @type {OC.Share.Types.RegisteredShareAttribute[]} registered available share permissions for this file/folder share
+		 * @deprecated shareAttributesApi v2 requires apps to extend ShareItemModel with attributes
 		 */
 		_registeredAttributes: null,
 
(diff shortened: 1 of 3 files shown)


micbar:
We are blocked, needs review from @PVince81

mrow4a:
@pmaier1 @PVince81 @DeepDiver1975 updated top post with tested matrix of v1/v2 with:

mrow4a:
@DeepDiver1975 @pmaier1 all tests green

core/js/tests/specs/shareitemmodelSpec.js
```diff
@@ -832,11 +930,41 @@ describe('OC.Share.ShareItemModel', function() {
 
 				// test
 				expect(
-					testShareWithAttributes(permissionsToSet, attributesToRegister, sharePropertiesToUpdate)
+					testUpdateShareWithAttributes(permissionsToSet, attributesToRegister, sharePropertiesToUpdate)
 				).toEqual(attributesToExpect);
 			});
 
-			it('uses/hides attributes with permission filters of registered attributes', function() {
+			it('updates attribute with new enabled value correctly when boyh shareAttributesApi v1 and v2 are used', function() {
```
PVince81: "boyh" => "both"

core/js/shareitemmodel.js
```diff
@@ -161,51 +164,12 @@
...
-					_.map(filteredRegisteredAttributes, function(checkAttr) {
-						if (incompatibleAttribute.scope === checkAttr.scope &&
-							incompatibleAttribute.key === checkAttr.key &&
-							incompatibleAttribute.enabled === checkAttr.default) {
-							isCompatible = false;
-						}
-					});
-				});
+			// Get default permissions
+			var permissions = properties.permissions || OC.PERMISSION_ALL;
```
PVince81: optional: would be cool to rename "properties" to "attributes" as seeing both used for the same thing is a bit confusing (in jquery for example attributes and properties are not 100% the same thing, so people familiar with those might think the same here)

PVince81 (commented):
Looks fine overall.

One thing I'm not very clear about is that some comment says that the API v2 requires overriding the ShareItemModel.addShare/updateShare, but I don't see any overriding done in the unit tests. What I see however, is that the model has an "attributes" JS property that can be set.

I guess the overriding by apps is only needed in case special handling is needed ? Can you clarify ?

mrow4a:
@PVince81 yes, this is used when some special handling is needed:
- if app has custom checkboxes not relying on other permissions (edit, share etc) app could just send a request to api server directly
- if app checkbox changes with permissions (edit, share etc), `addShare/updateShare` is wrapped to intercept that change.

PVince81:
> If this is possible, I can try to write a test that wrapps the updateShare call. Should I?

the test might not be that useful as the code path has no effect on the rest of the logic, considering that the overriding is mostly there for rendering purposes ?

if you think there's more to it, please go ahead and add tests. (and fix the typo)

then we're good to go :-)

mrow4a:
@PVince81 added tests for wrapping - check for `can be wrapped` in js tests.

PVince81 (approved):
:+1:

LinneyS:
@mrow4a 
Sorry, but I don’t understand which branch should I use now to test the new logic?

mrow4a:
@LinneyS please for now use share_item_model_attr_api_v2, in the future owncloud master (which is current target for next oc10 release), but this fix is required there https://github.com/owncloud/core/pull/36086 .

mrow4a:
@LinneyS I would recommend also not to rush with implementing new logic in your app, we are still searching for best recommended "styling". I would be happy also to submit some PR to your working branch for share attributes, or in talk.owncloud.com we could have a chat and make sure we are all aligned.

Issue #35777, opened 2019-07-08
Change branching model for core ...


DeepDiver1975:
## Background

We use [OneFlow](https://www.endoflineblog.com/oneflow-a-git-branching-model-and-workflow) successfully for our apps for quite some time.
It is about time to use OneFlow for core as well!

## Steps to take
- [x] move current master branch to [archive/master](https://github.com/owncloud/core/tree/archive/master) - just for the sake of not loosing anything
- [x] move stable10 to master - this will be our main branch to release any future ownCloud 10.x.x
- [x] review all drone and remaining travis configurations and test with daily-master-qa only - done by @phil-davis 2019-08-01
- [x] transifex will operate on master
- [x] update nightly scripts to build only nightly for `master` ( which is then stable10 ) - https://github.com/owncloud/administration-internal/pull/149
- [x] build only containers for nightly master
- [x] smashbox && update-testing only for master
- [x] app acceptance tests no longer against core `stable10` - just against "new" core `master` and latest core release (see long list of linked PRs below)
- [x] keep user_management in core for now - app can be split again later - if necessary
- [ ] remove encryption app - https://github.com/owncloud/core/pull/35949

## Benefits
- no waste of time with backporting master -> stable10
- no more translation backports
- far less jobs on drone

patrickjahns:
Some aspect to add:
- [ ] update nightly scripts to build only nightly for `master` ( which is then stable10 )
- [ ] build only containers for nightly master
- [ ] smashbox && update-testing only for master

phil-davis:
"move current master branch to master-old - just for the sake of not loosing anything"
good - because I keep finding "goodies" in the current core `master` that are useful to have and still should be back-ported (e.g. after "drop PHP 7.0")

phil-davis:
Apps that were separated from `core` in the "old core master", but never got into a release like that:
- [ ] `user_management`
- [ ] `encryption`

what to do about those app repos?

DeepDiver1975:
for the sake of keep it simple and be safe: just keep the apps in core and archive these two repos - in a later stage we can still continue to split them off .... objections?

phil-davis:
How to remove deprecated code/interfaces/stuff in the future?
e.g. https://github.com/owncloud/core/pull/32051#issuecomment-509196336 
#32051 is in `master` only awaiting "`v11`"

DeepDiver1975:
v11 will not happen - so no removing of deprecated constructs

patrickjahns:
[...]

PR #35844, opened 2019-07-16
overwrite appconfig with config/apps.json if it exists

butonic:
cc @xoxys, @tboerger, @micbar, @felixboehm this is a PoC for reading the config from `config/apps.json` if it exists and skip the database. It would allow configuring apps with ansible by writing json to that file.

The current appconfig can be exported with `occ config:list apps`. Don't do a `occ config:list apps > config/apps.json`, that will create an empty file before running the script, producing a ton of errors. rm the `config/apps.json` if you want to change anything. changes will land in the database, so this might need some tweaking for things that are not really config options ...

AFAICT we should read the appconfig from the file and only fetch missing entries from the db if necessary.


butonic:
some appconfig options actually reflect state, eg the backgroundjob `lastjob` ... hm maybe it is enough to just overwrite the db result. that would also allow us to make some oracle problems go away with regards to the length of the configvalue column ... hmmm nice

butonic:
I changed to only overwrite the result from the db. that allows fixing the app config values with ansible and jets apps still store their state in the db. values from the apps.json will however always take precedence. I am using [array_replace_recursive](https://www.php.net/manual/de/function.array-replace-recursive.php)

DeepDiver1975:

PR #35848, opened 2019-07-17
avoid querying accounts when search string is too short

phil-davis:
## Description
If the given search string is too short ( not `isSearchable()`) then do not bother calling the `accountMapper` methods to get the matching accounts. That should save some wasted processing if/when short search strings are sent.

## Motivation and Context
Do not do extra processing.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:
- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35849, opened 2019-07-17
[stable10] avoid querying accounts when search string is too short

phil-davis:
Backport #35848


phil-davis:
Closing while `stable10` -> `master` change is happening.
This can easily be found again and progressed later.

PR #35862, opened 2019-07-17
Session fixes

jvillafanez:
## Description
Better session handling when the session file is removed

## Related Issue
https://github.com/owncloud/enterprise/issues/2509

#### login + logout
1. start a new browser and go to the login page
2. login with a user. Check the session id is different
3. logout. Check the session id has changed again
4. login with a different user. Check the session id has changed one more time.

There is no noticeable difference comparing against what was there before

#### session removed while browsing (web UI)
1. login with a user.
2. remove the session from the server
3. navigate to a different page, such as the settings one.

The session id is kept, although there is no information stored (the file with the associated data has been removed). The user is kicked out and forced to reauthenticate.
New session id is created after reauthentication. This is fine and will likely help if the user changes.

#### session removed while browsing using android client (checked with oAuth2)
1. login with a user
2. pull down the root folder to refresh the contents
3. remove the session from the server
4. pull down again

There is no aparent change, and there is no duplicated auth token in the logs

#### session expired while browsing using android client (checked with oAuth2)
1. setup the server to expire sessions after 30 secs
2. login with the android client
3. wait 30 secs so the session expires in the server
4. navigate through the android client

No aparent change. No duplicated auth token.
Note that both desktop client and web UI keep pinging the server each 30 secs (?) so the session is unlikely to expire by itself

Note that there are still cases to be tested

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


jvillafanez:
@micbar this is ready to review.

jvillafanez:
@owncloud/qa some acceptance tests (https://drone.owncloud.com/owncloud/core/19528/314) are failing because they can't authenticate in the "/index.php/app/files" directly. I guess the offending change is https://github.com/owncloud/core/pull/35862/files#diff-0bb5cca5abd6cda96b669c31856a3f59L240

question: do those tests make sense? I mean that endpoint is mainly for the webUI, so using basic auth there or any other auth... I think you're expected to have a logged-in session already at that point in order to access.

I can revert that change, but it's there because when the session expires while browsing I expect the browser to go back to the login page in order to force the user to reauthenticate. The other option is to find an alternative to fix the test in the code (without touching the test, of course)

individual-it:
If the expectations is now that at this point you need a browser session we just need to change the expected results of the tests.
There is still one case where we receive `200` `When the user requests "/index.php/apps/files" with "GET" using the browser session`. If that is the only case when that should happen, just change the other test outcomes.

Only there would be the need of an other positive token Auth tests, to make sure it really works with an app-password - `tests/acceptance/features/apiAuth/tokenAuth.feature:21`

jvillafanez:
Aparently the tests are fixed now with the latest change, so no need to change the acceptance tests for now.

micbar:
@jvillafanez Seems that you are pretty happy with your code now.
I would like to have a PR against stable10. We could try to get a OC 10.2.1 patched and test it in a real scenario.

jvillafanez:
Backport in https://github.com/owncloud/core/pull/35888

There are still some things to investigate about session handling (I think we should somehow get rid of the isDavAuthenticated checks) and also cookie handling (it's spread in several classes), but taking into account the amount of time it has taken to have a fix which, so far, doesn't break anything, I think this should be merged first before continuing.

DeepDiver1975:
@jvillafanez I'm closing this in favor of #35888

PR #35873, opened 2019-07-18
Revert "Log exceptions from exceptionHandler method"

sharidas:
Reverts owncloud/core#35362

### DIFF
--- a/console.php
+++ b/console.php
@@ -58,68 +58,54 @@ if (\stripos(PHP_OS, 'WIN') === 0) {
 function exceptionHandler($exception) {
 	echo 'An unhandled exception has been thrown:' . PHP_EOL;
 	echo $exception;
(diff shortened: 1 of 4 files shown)


sharidas:
Why do we have to revert the change:

- The exceptions are redirected to the console, instead of owncloud.log. The idea behind the issue was to redirect the error to the log file and not to the console.
- From the admin's perspective this is not good. Because admin's would like to see the clean error message, the way symfony throws into the console.
- One of the obstacle here is the colored error messages to stderr come from symfony. So either we pass on the data to the symfony and have the colored error messages on the screen or we may have to have our own way of coloring the messages. Which imho is not worth.

micbar (approved):

sharidas:
Stable10 revert : https://github.com/owncloud/core/pull/35876

PR #35882, opened 2019-07-19
Include URL after substituteInLineCodes in acceptance test error messages

phil-davis:
## Description
If there are "inline codes" for the URL mentioned in an acceptance test step (e.g. `%remote_server%`) and there is an error, then when reporting the error include both the code an its substituted value in the error message.

## Related Issue
Investigating acceptance test fails when running with user_ldap.

## Motivation and Context
When one of the trusted server acceptance tests fails, we get an error like:
```
    And url "%remote_server%" should be a trusted server                                                                       # FeatureContext::urlShouldBeATrustedServer()
      Given url %remote_server% is not a trusted server
```

It would be nice to know what the value of `%remote_server%` really was at the time of the error.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/AppConfiguration.php
+++ b/tests/acceptance/features/bootstrap/AppConfiguration.php
@@ -22,6 +22,7 @@
  *
  */
 
+use PHPUnit\Framework\Assert;
 use TestHelpers\AppConfigHelper;
 use TestHelpers\HttpRequestHelper;
 use TestHelpers\OcsApiHelper;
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Backport `stable10` #35883

PR #35884, opened 2019-07-19
Refactor share permissions handling

### LINKED ISSUE #35922: Sharing a file with just create and/or delete permission should fail
### Steps to reproduce
Use the sharing API. Attempt to share a single file with another user, but request just create (4) and/or delete (8) permissions.

### Expected behaviour
create and delete permissions are not valid/relevant for single file shares, so they get masked out. The resulting effective requested share permission is 0 (no permissions).

The share should not be created. A "40*" response should be returned.

### Actual behaviour
When create is requested, the share is created with `0` permissions. It is a useless share because the receiver cannot see the file in any way.

When delete is requested, the share is created with read `1` permissions. The receiver can read the file - that is weird, I wonder what code causes that to happen.

See PR #35921 for the example test scenarios that demonstrate this.

### Server configuration
Current core on GitHub.

karakayasemi:
## Description
This PR moves some of the validation on Share20Controlller to the share manager instead and changes permission calculation on re-shares.

Currently, permission handling different for `update` and `create` share operations, also two different classes(`Share20OcsController.php` and `Manager`) are responsible from this permission checks. This situation causes issues such as a scenario that works well in one operation does not work in another. (https://github.com/owncloud/enterprise/issues/3299 only affecting `update` share and the another is only affecting `create` share).

In a clean approach, Share manager should take care of all sorts of validations and logics instead of controller. I unified these logics under Share Manager and applied this unified checks for both `update` and `create` operations.

Also, with this PR re-share permission will be calculated based on share permission of re-share mount point node's.

IMHO, most of the permission exceptions should return `403`. However, to keep acceptance test and behavior changes minimal, I did not touch return codes and I kept them as it is.

I removed many unit tests. Some of them were duplicate and redundant. I will add more unit tests after reviews.

**_Share API, no longer set read permission as default._ Clients should include this permission to their requests. We need to test this behavior on clients to ensure there is no regression.**

## Related Issue
fixes https://github.com/owncloud/enterprise/issues/3299
fixes https://github.com/owncloud/enterprise/issues/3404
fixes #35922 

## Motivation and Context
To get rid of from bugs. To have a cleaner code. Next time we will know where we should look at on a share permission issue.

## How Has This Been Tested?
Acceptance tests, Also it resolves described scenarios in these tickets: 
- https://github.com/owncloud/enterprise/issues/3299 
- https://github.com/owncloud/enterprise/issues/3404
- #35922 

## Types of changes

- [x] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -208,7 +208,7 @@ class Manager implements IManager {
 	 * @throws \InvalidArgumentException
 	 * @throws GenericShareException
 	 */
-	protected function generalCreateChecks(\OCP\Share\IShare $share) {
+	protected function generalChecks(\OCP\Share\IShare $share) {
 		if ($share->getShareType() === \OCP\Share::SHARE_TYPE_USER) {
 			// We expect a valid user as sharedWith for user shares
 			if (!$this->userManager->userExists($share->getSharedWith())) {
@@ -270,29 +270,56 @@ class Manager implements IManager {
 			throw new GenericShareException($message_t, $message_t, 404);
 		}
 
+		$this->validatePermissions($share);
+	}
+
+	/**
+	 * Validate if the permission allowed
+	 *
+	 * @param IShare $share The share to validate its permission
+	 * @throws GenericShareException
+	 * @throws \InvalidArgumentException
+	 */
+	protected function validatePermissions(IShare $share) {
 		// Permissions should be set
 		if ($share->getPermissions() === null) {
 			throw new \InvalidArgumentException('A share requires permissions');
 		}
 
+		if ($share->getPermissions() < 0 || $share->getPermissions() > \OCP\Constants::PERMISSION_ALL) {
+			$message_t = $this->l->t('invalid permissionss');
+			throw new GenericShareException($message_t, $message_t, 404);
+		}
+
+		if ($share->getPermissions() === 0) {
+			$message_t = $this->l->t('Cannot remove all permissions');
+			throw new GenericShareException($message_t, $message_t, 400);
+		}
+
+		$shareNode = $share->getNode();
+		/** Use share node permission as default $maxPermissions*/
+		$maxPermissions = $shareNode->getPermissions();
 		/*
 		 * Quick fix for #23536
 		 * Non moveable mount points do not have update and delete permissions
 		 * while we 'most likely' do have that on the storage.
 		 */
-		$permissions = $share->getNode()->getPermissions();
-		$mount = $share->getNode()->getMountPoint();
-		if (!($mount instanceof MoveableMount)) {
-			$permissions |= \OCP\Constants::PERMISSION_DELETE | \OCP\Constants::PERMISSION_UPDATE;
+		if (!($shareNode->getMountPoint() instanceof MoveableMount)) {
+			$maxPermissions |= \OCP\Constants::PERMISSION_DELETE | \OCP\Constants::PERMISSION_UPDATE;
+		}
+
+		/** If it is re-share, calculate $maxPermissions based on all incoming share permissions */
+		if ($shareNode->getOwner()->getUID() !== $share->getSharedBy()) {
+			$maxPermissions = $this->calculateReshareNodePermissions($share);
 		}
 
 		// Check that we do not share with more permissions than we have
-		if ($share->getPermissions() & ~$permissions) {
+		if (!$this->strictSubsetOf($maxPermissions, $share->getPermissions())) {
 			$message_t = $this->l->t('Cannot increase permissions of %s', [$share->getNode()->getPath()]);
 			throw new GenericShareException($message_t, $message_t, 404);
 		}
 
-		if ($share->getNode() instanceof \OCP\Files\File) {
+		if ($shareNode instanceof \OCP\Files\File) {
 			if ($share->getPermissions() & \OCP\Constants::PERMISSION_DELETE) {
 				$message_t = $this->l->t('Files can\'t be shared with delete permissions');
 				throw new GenericShareException($message_t);
@@ -302,6 +329,48 @@ class Manager implements IManager {
 				throw new GenericShareException($message_t);
 			}
 		}
+	}
+
+	/**
+	 * It calculates reshare permissions based on all share mount points
+	 * that mounted to UserFolder of sharer user
+	 *
+	 * @param \OCP\Share\IShare $share The share to validate its permission
+	 * @return int
+	 */
+	protected function calculateReshareNodePermissions(IShare $share) {
+		$maxPermissions = 0;
+		$incomingShares = [];
+		$shareTypes = [
+			\OCP\Share::SHARE_TYPE_USER,
+			\OCP\Share::SHARE_TYPE_GROUP
+		];
+		$userFolder = $this->rootFolder->getUserFolder($share->getSharedBy());
+		/**
+		 * The node can be shared multiple times, get all share nodes
+		 */
+		$incomingShareNodes = $userFolder->getById($share->getNode()->getId());
+		foreach ($incomingShareNodes as $incomingShareNode) {
+			/**
+			 * find mountpoint of the share node,
+			 * check incoming share permissions based on mountpoint node
+			 */
+			$mount = $incomingShareNode->getMountPoint();
+			$shareMountPointNodes = $userFolder->getById($mount->getStorageRootId());
+			foreach ($shareMountPointNodes as $shareMountPointNode) {
+				$incomingShares = \array_merge($incomingShares, $this->getAllSharedWith(
+					$share->getSharedBy(),
+					$shareTypes,
+					$shareMountPointNode
+				));
+			}
+		}
+
+		foreach ($incomingShares as $incomingShare) {
+			$maxPermissions |= $incomingShare->getPermissions();
+		}
+
+		return $maxPermissions;
 	}
 
 	/**
@@ -551,13 +620,14 @@ class Manager implements IManager {
 	 * @param \OCP\Share\IShare $share
 	 * @return Share The share object
 	 * @throws \Exception
+	 * @throws GenericShareException
 	 *
 	 * TODO: handle link share permissions or check them
 	 */
 	public function createShare(\OCP\Share\IShare $share) {
 		$this->canShare($share);
 
-		$this->generalCreateChecks($share);
+		$this->generalChecks($share);
 
 		// Verify if there are any issues with the path
 		$this->pathCreateChecks($share->getNode());
@@ -789,6 +859,7 @@ class Manager implements IManager {
 	 * @param \OCP\Share\IShare $share
 	 * @return \OCP\Share\IShare The share object
 	 * @throws \InvalidArgumentException
+	 * @throws GenericShareException
 	 */
 	public function updateShare(\OCP\Share\IShare $share) {
 		$expirationDateUpdated = false;
@@ -818,7 +889,7 @@ class Manager implements IManager {
 			throw new \InvalidArgumentException('Can\'t share with the share owner');
 		}
 
-		$this->generalCreateChecks($share);
+		$this->generalChecks($share);
 
 		if ($share->getShareType() === \OCP\Share::SHARE_TYPE_USER) {
 			$this->userCreateChecks($share);
@@ -1568,5 +1639,16 @@ class Manager implements IManager {
 			// This is a new share
 		}
 		return ($fullId === null);
+	}
+
+	/**
+	 * Check $newPermissions bit is a subset of $allowedPermissions
+	 *
+	 * @param int $allowedPermissions
+	 * @param int $newPermissions
+	 * @return boolean ,true if $allowedPermissions bit super set of $newPermissions bit, else false
+	 */
+	private function strictSubsetOf($allowedPermissions, $newPermissions) {
+		return (($allowedPermissions | $newPermissions) === $allowedPermissions);
 	}
 }
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -286,6 +286,17 @@ class Manager implements IManager {
 			throw new \InvalidArgumentException('A share requires permissions');
 		}
 
+		$shareNode = $share->getNode();
+		if ($shareNode instanceof \OCP\Files\File) {
+			// Single file shares should never have delete or create permissions
+			$share->setPermissions($share->getPermissions() & ~\OCP\Constants::PERMISSION_DELETE);
+			$share->setPermissions($share->getPermissions() & ~\OCP\Constants::PERMISSION_CREATE);
+		}
+
+		/**
+		 * TODO: ideally, getPermissions should always return valid permission
+		 *       and this check should be done in Share object's setPermission method
+		 */
 		if ($share->getPermissions() < 0 || $share->getPermissions() > \OCP\Constants::PERMISSION_ALL) {
 			$message_t = $this->l->t('invalid permissionss');
 			throw new GenericShareException($message_t, $message_t, 404);
@@ -296,7 +307,6 @@ class Manager implements IManager {
 			throw new GenericShareException($message_t, $message_t, 400);
 		}
 
-		$shareNode = $share->getNode();
 		/** Use share node permission as default $maxPermissions*/
 		$maxPermissions = $shareNode->getPermissions();
 		/*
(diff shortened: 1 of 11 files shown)


jvillafanez:
I haven't seen the whole code in depth, but on a general perspective I think it's better to move the permission calculation to a different class, so the share manager can rely on it. This way, we can unittest this new class completely, increase the code coverage, and also simplify the code in the share manager.

Side note, I don't like protected methods unless there is a reason for it.

karakayasemi:
When I looking to failing unit tests, I discovered that, when a user shares a file with `create` permission, `Share20Controller` modifies the share and removes this `create` permission. If the share had only `create` permission, this share is created with 0 permission and this should not happen.

In this PR's implementation, this bug is also seems resolved. One of the failing test scenarios was using this wrong behavior. I believe, I corrected the test scenario with this commit:  c4f5dc523c10f5332110b578b0e3e9d2de05faf5. @phil-davis , @individual-it please correct me, if I am wrong about the intended behaviour on this acceptance test.

phil-davis:
Correct, that scenario tried to create a share of a file and give `create` permission. A file share cannot have `create` permission, because a single file is already, by definition "created". The  back-end subtracts the `create` permission from the requested permissions when there is a share of a file.

So that scenario should never have worked, because the `Given` step should not have created the share.

I will make some tests, in another PR, that test attempting to create a share of a file like that. That will confirm the current (probably wrong) behaviour. Then the work in this PR can correct whatever is wrong.

karakayasemi:
@jvillafanez I used protected-private methods to give descriptive names for chunks of code and shorten method lines. The class has also similar protected validate methods. If we move permission calculations to another class, we should move the rest of the validation methods too. However, I'm not sure this effort worths.

Recently, we faced lots of issues about share permission calculations. Some of them are still waiting to be resolved. As far as I see, some of the features about permissions are working just in luck and they may have undiscovered bugs. The main focus of this PR is unifying share permission checks and fixing these bugs as I described above.

We covered lots of the issues with acceptance tests and the CI is now green. Please, help by reviewing in-depth to do not miss anything.

lib/private/Share20/Manager.php
```diff
@@ -271,29 +271,56 @@ protected function generalCreateChecks(\OCP\Share\IShare $share) {
...
-		$mount = $share->getNode()->getMountPoint();
-		if (!($mount instanceof MoveableMount)) {
-			$permissions |= \OCP\Constants::PERMISSION_DELETE | \OCP\Constants::PERMISSION_UPDATE;
+		if (!($shareNode->getMountPoint() instanceof MoveableMount)) {
+			$maxPermissions |= \OCP\Constants::PERMISSION_DELETE | \OCP\Constants::PERMISSION_UPDATE;
+		}
+
+		/** If it is re-share, calculate $maxPermissions based on all incoming share permissions */
+		if ($shareNode->getOwner()->getUID() !== $share->getSharedBy()) {
+			$maxPermissions = $this->calculateReshareNodePermissions($share);
```
jvillafanez: This will override previous calculated `maxPermissions` just a few lines before.

lib/private/Share20/Manager.php
```diff
@@ -305,6 +332,54 @@ protected function generalCreateChecks(\OCP\Share\IShare $share) {
...
+	 * that mounted to UserFolder of sharer user
+	 *
+	 * @param \OCP\Share\IShare $share The share to validate its permission
+	 * @return int
+	 */
+	protected function calculateReshareNodePermissions(IShare $share) {
+		/*
+		 * if it is an incoming federated share, use node permission
+		 */
+		if ($share->getNode()->getStorage()->instanceOfStorage('OCA\Files_Sharing\External\Storage')) {
```
jvillafanez: any better alternative? core shouldn't check for specific apps.

lib/private/Share20/Manager.php
```diff
@@ -305,6 +332,54 @@ protected function generalCreateChecks(\OCP\Share\IShare $share) {
...
+		];
+		$userFolder = $this->rootFolder->getUserFolder($share->getSharedBy());
+		/**
+		 * The node can be shared multiple times, get all share nodes
+		 */
+		$incomingShareNodes = $userFolder->getById($share->getNode()->getId());
+		foreach ($incomingShareNodes as $incomingShareNode) {
+			/**
+			 * find mountpoint of the share node,
+			 * check incoming share permissions based on mountpoint node
```
jvillafanez: I'm a bit lost here... could you explain a use case?

User1 has a share in "/t1", which points to User2's "/my/shared/folder". This shared folder is in an external storage mounted on User2's "/my" folder.
In this case, I think you're trying to get the permissions for the User2's "/my" folder... however, I think we need to check the permissions for the "/my/shared/folder". I don't see any reason to check "/my" folder: if the permissions for both "/my" and "/my/shared/folder" are the same, we don't want to spend time finding the storage root, and if the permissions are different, the permissions for the "/my/shared/folder" should always take precedence.

lib/private/Share20/Manager.php
```diff
@@ -271,29 +271,56 @@ protected function generalCreateChecks(\OCP\Share\IShare $share) {
...
 		if ($share->getPermissions() === null) {
 			throw new \InvalidArgumentException('A share requires permissions');
 		}
 
+		if ($share->getPermissions() < 0 || $share->getPermissions() > \OCP\Constants::PERMISSION_ALL) {
+			$message_t = $this->l->t('invalid permissionss');
+			throw new GenericShareException($message_t, $message_t, 404);
+		}
+
+		if ($share->getPermissions() === 0) {
```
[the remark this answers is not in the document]
karakayasemi: Even a small change, can affect the behavior of API consumers. I tried to do not change any response as possible as I can. This check was in `ShareController`, I moved it here.

lib/private/Share20/Manager.php
karakayasemi: Yes, it is intended. Re-share permission calculation is different than normal share. Instead of using `else` statement, I assigned a default value and overrode it if it is a re-share.

lib/private/Share20/Manager.php
karakayasemi: We need to determine permission for remote share when re-sharing it as a local share. Ideally, I would like to get remote sharing permissions with the way that I use below for the other type of shares. However, `FederatedShareProvider`'s `getAllSharedWith` method is always returning an empty response for a local user. I could not find any better way to determine incoming remote share permission of the user and took it from re-sharer User `View`.

I would not like to use and depend on 'OCA' namespace classes under an `OC` namespace class. However, refactoring `FederatedShareProvider` is very costly and the same check is already used in this class. Using this as it is the easiest way.

lib/private/Share20/Manager.php
karakayasemi: > 
> User1 has a share in "/t1", which points to User2's "/my/shared/folder". This shared folder is in an external storage mounted on User2's "/my" folder.
> In this case, I think you're trying to get the permissions for the User2's "/my" folder...

Every share has own mount point. In your scenario we will check at the mount point on User2's `/my/shared/folder/t2`, will not User2's `/my` folder.

One of our current problems, `View` is always adding additional `delete` permission to share nodes to allow recipient user to unshare this share from his home folder. So, we can not trust the permissions coming from `Node` view.

Now let's continue to your first scenario. Let's say, user1 shared `/t2` folder with "share+read" permission to `user2` and there is a file on `/t2/resharedfile`. When user2 try to share `/t2/resharedfile`, to avoid from additional delete bit we should look at the its mountpoint's (`\t2`) share permission for user2.

lib/private/Share20/Manager.php
```diff
@@ -271,29 +271,56 @@ protected function generalCreateChecks(\OCP\Share\IShare $share) {
...
+	 * @throws GenericShareException
+	 * @throws \InvalidArgumentException
+	 */
+	protected function validatePermissions(IShare $share) {
 		// Permissions should be set
 		if ($share->getPermissions() === null) {
 			throw new \InvalidArgumentException('A share requires permissions');
 		}
 
+		if ($share->getPermissions() < 0 || $share->getPermissions() > \OCP\Constants::PERMISSION_ALL) {
```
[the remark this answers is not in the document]
karakayasemi: I don't think there's a better alternative. When we compare permissions, we are already using implementation details.

phil-davis:
PR #35921 has been merged. That adds more share permissions scenarios.
@karakayasemi you can rebase to latest master, then adjust the scenarios added in that PR so that they demonstrate the expected correct behaviour and then make sure they pass.

karakayasemi:
I rebased the PR to test against newly added acceptance tests. Looks like some of them failed. I will have a look what's wrong there.

lib/private/Share20/Manager.php
jvillafanez: I guess the extra computation won't be big enough to have any noticeable impact... since this is intended I'm fine to keep it.

lib/private/Share20/Manager.php
jvillafanez: Could we assume first that the share is local, and if not fallback to this?

lib/private/Share20/Manager.php
jvillafanez: ok, got it.

lib/private/Share20/Manager.php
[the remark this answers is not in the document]
jvillafanez: what about `if ($share->getPermissions() & \OCP\Contants::PERMISSION_ALL)`? If that works, we could move it to a function for easier understanding.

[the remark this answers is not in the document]
karakayasemi: I guess, your condition returns the value of a number in mod 32. It can not look for invalid bits. We can somehow represent this condition with bit operations. Need to think a way to checking whether a number is in a particular range by using bit operators.  But I doubt it is necessary.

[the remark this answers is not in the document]
jvillafanez: The best option would be for the share to ensure it always has valid permissions, so the `getPermissions` always return a valid value. I think this is something we should do at some point.
Meanwhile, you can choose any method as long as you include a "TODO".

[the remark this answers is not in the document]
karakayasemi: I added a `TODO` after your redirections. Thanks.

lib/private/Share20/Manager.php
karakayasemi: I would like to keep it like that if you don't mind. Because, if we refactor `FederatedShareProvider` to return shares when `getAllSharedWith` called, we can remove this check completely. 
If I move this to the at the end of the method, since we will still make share mount point searches, complexity will increase unnecessarily. Also, we will add an one more condition ` ($maxPermission === 0)` for fallback. So, I can not see a significant benefit.

jvillafanez (approved):

DeepDiver1975:
@karakayasemi Just rebased this PR because of https://github.com/owncloud/core/issues/35777

karakayasemi:
Finally, the CI is green again. @phil-davis your review, especially about acceptance tests, would help us to not miss anything in here. There are some changes related to removing default read permission.

tests/acceptance/features/apiComments/createComments.feature
```diff
@@ -53,16 +53,16 @@ Feature: Comments
     And the user should have the following comments on file "/myFileToComment.txt"
       | user1 | Comment from sharee |
 
-  Scenario: sharee comments on upload-only shared file
-    Given the user has uploaded file "filesForUpload/textfile.txt" to "/myFileToComment.txt"
+  Scenario: sharee comments on upload-only shared folder
+    Given the user has created folder "/FOLDER_TO_SHARE"
     And the user has created a share with settings
```
phil-davis: Note: the reason for changing this is that the old scenario created a share of a file with only `create` permission.
That is not a valid thing to do anyway, and so the `When` step was  failing anyway, because the share may not have existed at all.
So using a folder in the scenario is better/valid.

tests/acceptance/features/apiShareManagementBasic/createShare.feature
```diff
@@ -53,18 +53,15 @@ Feature: sharing
...
+      # Ask for read, share (17), create and delete. You get share plus read
+      | 1               | 29                    | 17                  | 100             |
+      | 2               | 29                    | 17                  | 200             |
       # Ask for read, update, create, delete. You get read plus update.
       | 1               | 15                    | 3                   | 100             |
       | 2               | 15                    | 3                   | 200             |
-      # Ask for create and delete. You get just read.
-      | 1               | 12                    | 1                   | 100             |
-      | 2               | 12                    | 1                   | 200             |
       # Ask for just update. You get read plus update.
```
phil-davis:
```suggestion
      # Ask for just update. You get exactly update (you do not get read or anything else)
```

tests/acceptance/features/apiShareManagementBasic/createShare.feature
```diff
@@ -983,15 +933,14 @@ Feature: sharing
...
+    When user "user3" shares file "randomfile.txt" with user "user1" with permissions "read,update" using the sharing API
     And user "user1" gets the info of the last share using the sharing API
     Then the fields of the last response should include
       | uid_owner   | user3              |
       | share_with  | user1              |
       | file_target | /randomfile (2).txt|
       | item_type   | file               |
-      | permissions | read,update        |
-    # Here the last response contains permissions = 3 which is equivalent to permissons: read(1) + update(2)
+      | permissions | read,update             |
```
phil-davis: line up the table columns (while you are changing other stuff)

tests/acceptance/features/apiShareManagementBasic/createShare.feature
```diff
@@ -1010,22 +959,22 @@ Feature: sharing
...
-    When user "user3" shares folder "zzzfolder" with user "user1" with permissions "share" using the sharing API
+    When user "user3" shares folder "zzzfolder" with user "user1" with permissions "read,share" using the sharing API
     And user "user1" gets the info of the last share using the sharing API
     Then the fields of the last response should include
       | uid_owner   | user3          |
       | share_with  | user1          |
       | file_target | /zzzfolder (2) |
       | item_type   | folder         |
-      | permissions | share,read     |
+      | permissions | read,share          |
```
phil-davis: line up the table columns (while you are changing other stuff)

phil-davis (commented):
Also in `apps/files_sharing/tests/ShareTest.php` `testFileSharePermissions()` the comment says:
```
shared files should never have delete permissions
```
but now all the `dataProviderTestFileSharePermissions()` cases are `true`.

I think that is because `$this->share()` now does not throw exceptions like it did before. Maybe the requested delete permission gets masked out and the share is created OK, just without delete permission?

And there could be test cases for `PERMISSION_CREATE` also, because that is a special permission that does not apply to sharing a single file.

phil-davis:
Congratulations on adjusting all those scenarios, refactoring lots of stuff and also getting green CI! A good example of TDD.

tests/acceptance/features/apiShareManagementBasic/createShare.feature
karakayasemi: Thank you, now it is aligned.

tests/acceptance/features/apiShareManagementBasic/createShare.feature
karakayasemi: Thank you, now it is aligned.

tests/acceptance/features/apiShareManagementBasic/createShare.feature
karakayasemi: Changed as suggested.

karakayasemi:
> 
> Also in `apps/files_sharing/tests/ShareTest.php` `testFileSharePermissions()` the comment says:
> 
> ```
> shared files should never have delete permissions
> ```
> 
> but now all the `dataProviderTestFileSharePermissions()` cases are `true`.
> 
> I think that is because `$this->share()` now does not throw exceptions like it did before. Maybe the requested delete permission gets masked out and the share is created OK, just without delete permission?
> 

@phil-davis You are right about your prediction. In this test, shares are created correctly with the current desired behavior. Even so, I changed `testFileSharePermissions` implementation to ensure file shares never have create and delete permissions.

DeepDiver1975 (approved):

phil-davis (approved):

phil-davis:
No codecov came. I rebased to get a fresh run of CI, and maybe get codecov numbers?
IMO, it would be nice to know if the code+test changes have the same or better coverage.

karakayasemi:
@phil-davis I guess, we need https://github.com/owncloud/core/pull/35973 for fully green CI. I will rebase this when it merged.

karakayasemi:
Tried two times Codecov is not willing to come. We saw fully green CI before, I guess we can merge this with the power of one of the repo god.

PR #30421, opened 2018-02-08
No need to handle authentication that early aka in here at all

DeepDiver1975:
## Description
For an unauthenticated request there is no need to try to login implicitly in base.php

Login is done in the login controller.

## Related Issue
https://github.com/owncloud/core/pull/30398#issuecomment-363830803

## How Has This Been Tested?
- [x] manually browser
  - various apps load and seem to function properly: activity, files, calendar, contacts
- [x] desktop client with oauth
  - sync folder
  - auto completion of sharee
  - share

- [ ] ldap
- [ ] shib
- [x] mobile clients
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #35888, opened 2019-07-22
Session fixes

jvillafanez:
## Description
backport of https://github.com/owncloud/core/pull/35862 to stable10

## Related Issue
https://github.com/owncloud/enterprise/issues/2509

## How Has This Been Tested?

checked in the original PR

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
@@ -237,11 +237,6 @@ class SecurityMiddleware extends Middleware {
 	}
 
 	private function isLoggedIn() {
-		static $loginCalled = false;
-		if (!$loginCalled && !$this->session->isLoggedIn()) {
-			\OC::handleLogin($this->request);
-			$loginCalled = true;
-		}
 		return $this->session->isLoggedIn();
 	}
 }
--- a/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
@@ -126,7 +126,7 @@ class SecurityMiddleware extends Middleware {
 		// security checks
 		$isPublicPage = $this->reflector->hasAnnotation('PublicPage');
 		if (!$isPublicPage) {
-			if (!$this->isLoggedIn()) {
+			if (!$this->session->isLoggedIn() && !\OC::handleLogin($this->request)) {
 				throw new NotLoggedInException();
 			}
 
@@ -234,9 +234,5 @@ class SecurityMiddleware extends Middleware {
 		}
 
 		throw $exception;
-	}
-
-	private function isLoggedIn() {
-		return $this->session->isLoggedIn();
 	}
 }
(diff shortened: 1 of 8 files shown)


foobarable:
This caused seriosuly more memory consumption for our redis-server.

jvillafanez:
There is a `session_lifetime` key in the config.php that was used but never worked until now (https://github.com/owncloud/core/blob/8680f9c3a45142030011a0927e5df522004c5712/lib/base.php#L465).
The default value is set as 24 hours (didn't change that). This means that the session keys in redis will have a ttl of 24 hours.

You can adjust the value in the config.php file. Something between 15-30 minutes (600-900 seconds) should be fine. Note that activity in the sesion will refresh the ttl, so you need to be idle that amount of time for the session to expire. In addition, the web UI pings for notifications periodically, which prevents the session expiration.

Note that this also depend on how the clients (desktop, mobile) works.

jvillafanez:
Changed the default value to 20 minutes, which seems more reasonable.

DeepDiver1975:
@jvillafanez I'm rebasing this to get latest ci feedback ....

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -235,13 +235,4 @@ public function afterException($controller, $methodName, \Exception $exception)
 
 		throw $exception;
 	}
-
-	private function isLoggedIn() {
-		static $loginCalled = false;
```
PVince81: please confirm that we can safely remove this

PVince81:
@jvillafanez What happens with the old php-redis version with this PR merged ? Can you confirm that the behavior will not cause any troubles ?

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
jvillafanez: Looks like https://github.com/owncloud/core/pull/30421 is the one introducing the change, but I don't see any information about the one-time call.

jvillafanez:
> What happens with the old php-redis version with this PR merged ?

There is no change in the behaviour, so race conditions might still happen. The "critical" stuff is handled by redis, and the PR doesn't introduce any change targetting redis in particular. Most of the changes are minor touches aiming to simplify the code flow, but as said, the behaviour is the same. Anything that was happening before will keep on happening, except for the scenario where the session information disappears.

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
PVince81: @jvillafanez are you able to guess / deduct the reason why this could have been added ?

jvillafanez: I don't know. Possible reasons could be:
* problem in the unittest causing a user to log in twice or more (basically, the user didn't logged out after the test)
* bug that caused multiple calls to the method, but now fixed / patched / no reproducible
* wrong troubleshooting of a race condition caused by multiple requests made at the same time

There could be more, but I can't say what is the right one.

PVince81: Ok, thanks for the research

PVince81 (approved):
:+1: was reported working.

Extensive QA needed for all auth methods before the release. (cc @davitol)

PR #35906, opened 2019-07-23
throw exception in TrashbinContext restoreElement if element not found

phil-davis:
## Description
In `TrashbinContext` `restoreElement()` if the required element is not even found in the trashbin then the method silently exits without making any API  call. If that happens, then there will not be any updated API response data. So later test steps will be examining some old previous response when checking HTTP status etc. That will be misleading and could allow tests to pass when really there is something wrong.

- throw an exception if nothing was found to restore.

`sendUndeleteRequest() ` was not saving the response, so checks of HTTP status were snake oil.

- correctly set response in TrashbinContext sendUndeleteRequest and check HTTP status.

## Motivation and Context
Make tests more robust.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


individual-it (approved):

phil-davis:
Backport `stable10` #35907

PR #35931, opened 2019-07-26
Tell the browser not to autofill those password fields

jvillafanez:
## Description
Both firefox and chrome can autofill some username and password fields in forms without user interaction. This is a problem specially in the users page where the password field is hidden but the browser autocompletes it.

Note that neither safari nor edge have this behaviour. They require user interaction in order to fill the password field. This problem doesn't happen in those browsers.

## Related Issue
no ticket opened as far as I know.

## Motivation and Context
The admin was setting a password without any knowledge.

## How Has This Been Tested?
Checked with firefox, chrome, safari and edge:
1. login as admin.
2. when prompted, save the password in the browser
3. go to the users page
4. at this point the new username and password fields should be empty (this needs to be even if the password field isn't visible)
5. add a new user using the username + email fields
6. in a new browser, check the email sent and reset the password for that user
7. note that the password field for the reset form are also empty.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

**This should be moved to the user_management app**

### DIFF
--- a/settings/js/users/users.js
+++ b/settings/js/users/users.js
@@ -734,7 +734,7 @@ $(document).ready(function () {
 		var $td = $(this).closest('td');
 		var $tr = $(this).closest('tr');
 		var uid = UserList.getUID($td);
-		var $input = $('<input type="password">');
+		var $input = $('<input type="password" autocomplete="new-password" autocapitalize="off" autocorrect="off">');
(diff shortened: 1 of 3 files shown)


jvillafanez:
These are minor html changes, so nothing to unittest

DeepDiver1975 (approved):

PR #35938, opened 2019-07-29
make notificationsShouldBeDisplayedOnTheWebUIWithTheText more flexible

phil-davis:
## Description
`notificationsShouldBeDisplayedOnTheWebUIWithTheText` already allows there to be extra notifications in the list after the expected notifications.

Actually, it is quite reasonable that some "async" notification could come during the test setup phase, and be listed before the expected notification(s). And less likely, but possible, some "async" notification could come during the test `When` steps and be in the middle of the notifications list.

Also, there is not really a requirement that the notifications are displayed in some particular order.

The code is modified so that the expected notifications are looked for anywhere in the list of actual notifications. If the expected notifications are all found, then the test passes. It does not matter if there are extra notifications anywhere in the list.

## Related Issue
https://github.com/owncloud/password_policy/issues/239

## Motivation and Context
There can be other notifications that come on the webUI while tests are running. They can come before, during or after the `When`  test steps. So they could be in any order in the list of notifications.

## How Has This Been Tested?
Local testing with temporary scenarios that expect more or less notification messages.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/tests/acceptance/features/bootstrap/WebUIGeneralContext.php
+++ b/tests/acceptance/features/bootstrap/WebUIGeneralContext.php
@@ -344,23 +344,25 @@ class WebUIGeneralContext extends RawMinkContext implements Context {
 			"expected at least $numExpectedNotifications notifications but only found $numActualNotifications"
 		);
 
-		$notificationCounter = 0;
 		foreach ($expectedNotifications as $expectedNotification) {
 			$expectedNotificationText = $expectedNotification[0];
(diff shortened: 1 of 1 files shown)


individual-it (approved):

phil-davis:
Backport `stable10` #35941

PR #35947, opened 2019-07-30
[stable10] Adding id for auth message

phil-davis:
Backport #35121


phil-davis:
@DeepDiver1975 already did this

PR #35955, opened 2019-07-31
Trigger CI

phil-davis:
to see how CI is doing


phil-davis:
https://drone.owncloud.com/owncloud/core/19853/639
```
Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: TLS handshake timeout
```
These drone agent "random" errors are a PITA. It makes it very difficult to get a clean run of CI.

PR #35923, opened 2019-07-25
test more details of API response when creating a share

individual-it:
## Description
just checking a couple more API fields when creating a share

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35965, opened 2019-07-31
Do not require a fixed value for storage_id in createShare.feature

phil-davis:
## Description
PR #35923 added test steps that check that share create responses contain a `storage_id` field. Those tests specified the exact value of that field - `home::user0`.

When `files_primary_s3` runs those test scenarios with different storage backends, then `storage_id` has different values and the tests fail.

- do not specify a particular value for `storage_id`, we just want to know that it exists in the response.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -23,7 +23,7 @@ Feature: sharing
       | displayname_owner          | User Zero         |
       | item_type                  | file              |
       | mimetype                   | text/plain        |
-      | storage_id                 | home::user0       |
+      | storage_id                 | ANY_VALUE         |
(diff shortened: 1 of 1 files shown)


phil-davis:
Example fail in https://drone.owncloud.com/owncloud/files_primary_s3/1039/670
```
    Examples:
      | ocs_api_version | ocs_status_code |
        │ storage_id has unexpected value 'object::user:user0'
        │ 
        storage_id doesn't have value 'home::user0'
        │ storage_id has unexpected value 'object::user:user0'
        │ 
        storage_id doesn't have value 'home::user0'
```

individual-it (approved):

PR #35977, opened 2019-08-03
obey user min search length configuration on remote shares

karakayasemi:
## Description
Currently, sharing API is not respecting `user.min_search_length` configuration when returning remote users. This PR aims to fix this behavior. 
The API should return exact matches even `search_min_length` is not enough. For that reason, the PR only changes behavior on the non-exact match scenario.

## Related Issue
https://github.com/owncloud/enterprise/issues/3438

## Motivation and Context
Fixing bug.

## How Has This Been Tested?
Unit tests and manually with the following steps:
- https://github.com/owncloud/core/blob/master/config/config.sample.php#L302 set to "2"
- after searching `u@co`, share dialog should include `u@co`
- https://github.com/owncloud/core/blob/master/config/config.sample.php#L302 set to "5"
- after searching `u@co`, share dialog should not include `u@co`

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/ShareesController.php
+++ b/apps/files_sharing/lib/Controller/ShareesController.php
@@ -41,6 +41,7 @@ use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Share;
 use OCA\Files_Sharing\SharingBlacklist;
+use OCP\Util\UserSearch;
 
 class ShareesController extends OCSController {
 
@@ -70,6 +71,9 @@ class ShareesController extends OCSController {
 
 	/** @var \OCP\Share\IManager */
 	protected $shareManager;
+
+	/** @var UserSearch*/
+	protected $userSearch;
 
 	/** @var bool */
 	protected $shareWithGroupOnly = false;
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: I have asked for the drone JavaScript cache problem to be fixed. CI is broken because of that.

jvillafanez:
I think we should clarify first what is the expected behaviour.

If you're looking for users in the remote server, I expect the remote server to return users based on whatever is configured there. If the remote server has a value of `user.min_search_length = 5`, I expect the remote server to return users after filling the first 5 chars. After that, the search is done, so the local server shouldn't need to apply anything specific to the search.

Note that there are at least a couple of cases to check:
* local = 5, remote = 2
* local = 2, remote = 5

I guess this is for the first case where the local server has a greater restriction. However, since the search is done in the remote server, I think only the remote server's configuration should apply. If the local server wants to apply additional restrictions, I think it's better to use a different config key.

--------

Alternatively, I think it's better if the local server doesn't make the request to the remote server in the first place. Basically, if the local configuration has `user.min_search_length = 5`, the local server won't make a request to the remote one until those chars are filled.
It isn't the local server the one filtering the results, but asking for a minimum search length to the remote server. I think this is a better approach without adding new configuration keys.

Both cases described above are handled the same way without any modification.

karakayasemi:
@jvillafanez As far I see from the code, there is no request for the remote server. We are only searching in the local address book. If we have an exact match, we are returning the result whatever search length. It is intended behavior and I did not change this behavior.

If there is no exact match, currently, if there are half-matched results in contacts, the API returns these results. Also, if the search string matches with the remote user format, the API adds the string itself to returned results. I just added a length control for this non-exact match scenario.

DeepDiver1975 (approved):

PR #35756, opened 2019-07-03
Properly clean up `configkey` values from oc_appconfig table when disabling encryption

pako81:
## Description 
Currently, we have the problem that we do not correctly clean up the `configkey` values in the oc_appconfig table when disabling encryption. 

Specifically the `useMasterKey` key for the encryption app, which stays set to `1`. This seems to lead to the `encryption_enabled` key for core to remain set to `no` so when trying to reenable the encryption app and running `encryption:select-encryption-type masterkey`, this will fail.

The same has been observed for user-keys encryption. This PR fixes this behavior by properly cleaning up the `configkey` values and making re-enabling of encryption possible.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3410

## How Has This Been Tested?

- Manually by first activating encryption by running: `occ app:enable encryption`, `occ encryption:enable`, `occ encryption:select-encryption-type masterkey -y` and `encryption:encrypt-all`

- Then disable it with `occ app:disable encryption`, `occ encryption:disable` 

- Re-enabling encryption once again and check that when running `occ encryption:select-encryption-type masterkey`, no error is triggered.

Tested the same with user-keys encryption.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

## Open tasks:
- [x] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #35980, opened 2019-08-05
Properly clean up `configkey` values from oc_appconfig table when disabling encryption

pako81:
Follow-up of https://github.com/owncloud/core/pull/35756 because of changes in the branching model.

## Description
Currently, we have the problem that we do not correctly clean up the configkey values in the oc_appconfig table when disabling encryption.

Specifically the useMasterKey key for the encryption app, which stays set to 1. This seems to lead to the encryption_enabled key for core to remain set to no so when trying to reenable the encryption app and running encryption:select-encryption-type masterkey, this will fail.

The same has been observed for user-keys encryption. This PR fixes this behavior by properly cleaning up the configkey values and making re-enabling of encryption possible.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3410

## How Has This Been Tested?

- Manually by first activating encryption by running: `occ app:enable encryption`, `occ encryption:enable`, `occ encryption:select-encryption-type masterkey -y` and `encryption:encrypt-all`

- Then disable it with `occ app:disable encryption`, `occ encryption:disable` 

- Re-enabling encryption once again and check that when running `occ encryption:select-encryption-type masterkey`, no error is triggered.

Tested the same with user-keys encryption.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/35980
+++ b/changelog/unreleased/35980
@@ -1,0 +1,9 @@
+Enhancement: Cleanup encryption config values on disabling encryption
+
+occ encryption:disable command was changed to delete some encryption-specific
+config key-value pairs that made reenabling encryption not possible.
+A safety check was added to prevent disabling encryption until all files are
+decrypted. The occ encryption:disable command exits with an error code and message
+if the system still has any encrypted files.
+
+https://github.com/owncloud/core/pull/35980
--- a/core/Command/Encryption/Disable.php
+++ b/core/Command/Encryption/Disable.php
@@ -22,19 +22,24 @@
 namespace OC\Core\Command\Encryption;
 
 use OCP\IConfig;
+use OCP\IDBConnection;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class Disable extends Command {
+	/** @var IDBConnection */
+	protected $db;
 	/** @var IConfig */
 	protected $config;
 
 	/**
+	 * @param IDBConnection $db
 	 * @param IConfig $config
 	 */
-	public function __construct(IConfig $config) {
+	public function __construct(IDBConnection $db, IConfig $config) {
 		parent::__construct();
+		$this->db = $db;
 		$this->config = $config;
 	}
 
@@ -46,8 +51,21 @@ class Disable extends Command {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select($qb->expr()->literal('1'))
+			->from('filecache', 'fc')
+			->where($qb->expr()->gte('fc.encrypted', $qb->expr()->literal('1')))
+			->setMaxResults(1);
+		$results = $qb->execute();
+		$hasEncryptedFiles = (bool) $results->fetchColumn(0);
+		$results->closeCursor();
+		if ($hasEncryptedFiles !== false) {
+			$output->writeln('<info>The system still have encrypted files. Please decrypt them all before disabling encryption.</info>');
+			return 1;
+		}
+
 		/**
-		 * Delete both useMasterKey and userSepecificKey
+		 * Delete both useMasterKey and userSpecificKey
 		 */
 		$this->config->deleteAppValue('encryption', 'useMasterKey');
 		$this->config->deleteAppValue('encryption', 'userSpecificKey');
@@ -56,10 +74,10 @@ class Disable extends Command {
 
 		if ($this->config->getAppValue('core', 'encryption_enabled', 'no') !== 'yes') {
 			$output->writeln('Encryption is already disabled');
-			return 0;
+		} else {
+			$this->config->setAppValue('core', 'encryption_enabled', 'no');
+			$output->writeln('<info>Encryption disabled</info>');
 		}
-
-		$this->config->setAppValue('core', 'encryption_enabled', 'no');
-		$output->writeln('<info>Encryption disabled</info>');
+		return 0;
 	}
 }
--- a/core/Command/Encryption/Disable.php
+++ b/core/Command/Encryption/Disable.php
@@ -46,11 +46,20 @@ class Disable extends Command {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
+		/**
+		 * Delete both useMasterKey and userSepecificKey
+		 */
+		$this->config->deleteAppValue('encryption', 'useMasterKey');
+		$this->config->deleteAppValue('encryption', 'userSpecificKey');
+
+		$output->writeln('<info>Cleaned up config</info>');
+
 		if ($this->config->getAppValue('core', 'encryption_enabled', 'no') !== 'yes') {
 			$output->writeln('Encryption is already disabled');
-		} else {
-			$this->config->setAppValue('core', 'encryption_enabled', 'no');
-			$output->writeln('<info>Encryption disabled</info>');
+			return 0;
 		}
+
+		$this->config->setAppValue('core', 'encryption_enabled', 'no');
+		$output->writeln('<info>Encryption disabled</info>');
 	}
 }
--- a/core/register_command.php
+++ b/core/register_command.php
@@ -98,7 +98,7 @@ if (\OC::$server->getConfig()->getSystemValue('installed', false)) {
 	$application->add(new OC\Core\Command\Db\Migrations\GenerateCommand(\OC::$server->getDatabaseConnection()));
 	$application->add(new OC\Core\Command\Db\Migrations\ExecuteCommand(\OC::$server->getDatabaseConnection()));
 
-	$application->add(new OC\Core\Command\Encryption\Disable(\OC::$server->getConfig()));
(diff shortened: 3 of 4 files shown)


core/Command/Encryption/Disable.php
```diff
@@ -46,11 +46,21 @@ protected function configure() {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
+		$masterKeyEnabled = $this->config->getAppValue('encryption', 'useMasterKey', '');
+		$userKeyEnabled = $this->config->getAppValue('encryption', 'userSpecificKey', '');
+		if ($masterKeyEnabled !== '') {
+			$this->config->deleteAppValue('encryption', 'useMasterKey');
```
jvillafanez: any reason not to delete both master and user keys directly at the same time?

sharidas: At a point of time we can have only one type of encryption, either master or user-keys. So I thought to delete the one which is present at that time.

sharidas: Removed the checks. Updated the code to delete both `useMasterKey` and `userSpecificKey`.

jvillafanez:
I'm wondering if we should decrypt the files. What could happen if there are encrypted files and we disable the encryption? What happen to those files if we enable encryption (either master key or user key) afterwards?

I guess that we can't decrypt the files ouselves before disabling encryption, but if there is any risk we should at least warn the admin.

sharidas:
Lets say if admin disables encryption before decrypting the file(s), the admin always have the chance to recover the files, if he/she knows what type of encryption was used.
For example:
- Admin enables user-key encryption
- Encrypts the fs with user-key encryption
- Disables encryption 

At this point the files which are encrypted cannot be decrypted. Because the encryption is disabled. Now to recover the encrypted files, admin has to enable encryption and select the type of encryption which was used. In this example user-key encryption. And login to the respective user(s), the encrypted files will be decrypted if user tries to download them.

If the admin attemtps for masterkey then previously encrypted files will not be decrypted.

All I see is the admin should be aware of previously used encryption here.

jvillafanez:
Maybe we should call to PM to decide what to do, mainly because the admin will reenable encryption with the wrong type (or we should assume so).
* Write a wall of text in blood (otherwise the admin won't read it) explaining the risks if there are still encrypted files in the system
* Ensure that it's possible to disable encryption only if there is no encrypted files.
* Store somewhere, the previous encryption type in order to ensure the wrong encryption type won't be enabled afterwards. Makes me wonder if we really want to delete the keys stored in the configuration.

pako81:
My 2 cents: do not allow disabling encryption if we notice there are still encrypted files in FS. This way we are 100% on the safe side.

And this can easily be achieved by querying the filecache for files having the `encrypted` flag set to > 1.

If we do not want to consider this option, the alternative should be to allow disable but throw a warning when trying to disable encryption.

micbar:
@pmaier1 @pako81 @cdamken Could we make a decision here?

phil-davis:
@micbar there is no response here to your ping on Oct 21.
Is this something that will move forward? For 10.4 or?

pako81:
@cdamken @pmaier1 may we have your thoughts on this, please? My idea on this at https://github.com/owncloud/core/pull/35980#issuecomment-529343220

pako81:
This is not making any progress.

mmattel:
> do not allow disabling encryption if we notice there are still encrypted files in FS. This way we are 100% on the safe side

I would add that and finalize this PR. Then this can be reviewed and merged.

phil-davis:
@pako81 what is the status of this PR? Moving anywhere?
CC: @micbar

pako81:
@phil-davis this needs a PM decision and already pinged several times https://github.com/owncloud/core/pull/35980#issuecomment-560297253

pmaier1:
> @cdamken @pmaier1 may we have your thoughts on this, please? My idea on this at #35980 (comment)

I agree with @pako81, makes sense. First decrypt, then disable.

micbar:
@pako81 can DEV take over? @VicDeo Can bring this to a mergeable state?

VicDeo:
@pako81 does the current state match your proposal in https://github.com/owncloud/core/pull/35980#issuecomment-529343220 ?

pako81:
@VicDeo yes, looks fine from my POV - thx

core/Command/Encryption/Disable.php
```diff
@@ -46,11 +51,32 @@ protected function configure() {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select($qb->createFunction('COUNT(*)'))
+			->from('filecache', 'fc')
+			->where($qb->expr()->eq('fc.encrypted', $qb->expr()->literal('1')));
```
pako81, line 57: there could be the (very unlikely) case where all files have the encrypted flag different from `1` since they have been updated multiple times. Should we not just check for `!=0`?

jvillafanez: I think we should double-check this. Just for the name, it could also be a flag to indicate if the file is encrypted or not, without any meaning about the number of times the file has been encrypted.

What about something like `select 1 from oc_filecache where encrypted >= 1 limit 1`? This should perform faster because we just get the first entry without asking for any column. We also could benefit more than the other queries if the "encrypted" column is indexed (no need to add an index just for this).

jvillafanez: We might need to change the approach for this. I'm worried that the performance might be too low taking into account we'll have to deal with millions of rows and that the DB will need to perform a full scan. Even a simple query such as `select 1 from oc_filecache where encrypted >= 1 limit 1` could take a while.

I don't know if we have an instance big enough to have some estimates.

VicDeo: yes, this is a number that is used as a boolean. 
It's ok from my side to change it into `select 1 from oc_filecache where encrypted >= 1 limit 1`
This operation is not something that one should do even once per day. So it would be enough IMO.

jvillafanez:
```
MariaDB [owncloud]> select sql_no_cache count(*) from oc_filecache where encrypted >= 1;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.232 sec)

MariaDB [owncloud]> select sql_no_cache 1 from oc_filecache where encrypted >= 1 limit 1;
Empty set (0.228 sec)

MariaDB [owncloud]> select sql_no_cache count(*) from oc_filecache where encrypted >= 1;
+----------+
| count(*) |
+----------+
|        0 |
+----------+
1 row in set (0.222 sec)

MariaDB [owncloud]> select sql_no_cache 1 from oc_filecache where encrypted >= 1 limit 1;
Empty set (0.216 sec)

MariaDB [owncloud]> select sql_no_cache count(*) from oc_filecache;
+----------+
| count(*) |
+----------+
|   101270 |
+----------+
1 row in set (0.059 sec)
```
It should be around 2s / million of records, at least in my environment (no encrypted data there). It should perform better in real servers.

pako81: If it makes sense, I could ask one Customer having a medium-size installation and master-key encryption enabled to run that query so that we get an estimate for real servers.

jvillafanez: I don't think it's needed. The customer will have several encrypted files, so the row scan will stop with the first match. This could be the first row found.
We're interested in the worst case, which is having all the files decrypted. This is also required for the command to proceed disabling the encryption, so we'll eventually hit this point.

VicDeo: The query has been updated

jvillafanez:
Just to double-check, did you test running the command without encryption? I think `fetchColumn` will return `false` if there are no more rows, so I think it's better to check against this, otherwise you'll likely be checking `false > 0` which might work but feels strange.

VicDeo:
> you'll likely be checking `false > 0` which might work but feels strange.

true. I explicitly casted the result to boolean. Is it better now?

jvillafanez (approved):

changelog/unreleased/35980
```diff
@@ -0,0 +1,8 @@
+Enhancement: Cleanup encryption config values on disabling encryption
+
+occ encryption:disable command was changed to delete some encryption-specific
+config key-value pairs that made reenabling encryption not possible.
+Also a safety check was added to prevent disabling encrypion until all files are
+decrypted. And finally occ encryption:disable command was granted with an exit code.
```
phil-davis, line 6:
```suggestion
A safety check was added to prevent disabling encryption until all files are
decrypted. The occ encryption:disable command exits with an error code and message
if the system still has any encrypted files.
```

phil-davis: Some suggested words and a typo fix for the changelog.

VicDeo: thanks, updated

phil-davis (approved):

VicDeo:
No idea about the failures:
>apiFederation1-10.4.1-mariadb10.2-php7.4: Error response from daemon: No such image: owncloudci/core:latest

https://hub.docker.com/r/owncloudci/core/tags
connectivity issue?

phil-davis:
I restarted drone - it's having a bad day.

PR #35998, opened 2019-08-08
Restrict deletion of share_folder by user

sharidas:
Restrict deletion of share_folder by user

## Description

The idea behind this PR is to restrict user to delete the folder `share_folder` set in the config.php. In this change, I have tried to restrict the deletion from the View and from the dav app (i.e, access via webdav ).

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Restrict user to delete folder `share_folder` set in the config.php.

## How Has This Been Tested?

- Add `SharedFolder` to the config as `share_folder => 'SharedFolder'`
- Tested by deleting the folder from the UI results in the notification as shown below: 
sharenoitification
- The above test covers testing of webdav request to delete.
- Created a script to access the View of a user which has `SharedFolder` and then called the `rmdir` method of View. Received the return as false.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -100,7 +100,11 @@ class View {
 	/** @var \OC\User\Manager  */
 	private $userManager;
 
+	/** @var \Symfony\Component\EventDispatcher\EventDispatcherInterface */
 	private $eventDispatcher;
+
+	/** @var \OCP\IConfig */
+	private $config;
 
 	private static $ignorePartFile = false;
 
@@ -121,6 +125,7 @@ class View {
 		$this->lockingEnabled = !($this->lockingProvider instanceof \OC\Lock\NoopLockingProvider);
 		$this->userManager = \OC::$server->getUserManager();
 		$this->eventDispatcher = \OC::$server->getEventDispatcher();
+		$this->config = \OC::$server->getConfig();
 	}
 
 	public function getAbsolutePath($path = '/') {
@@ -347,6 +352,11 @@ class View {
 	 */
 	public function rmdir($path) {
 		return $this->emittingCall(function () use (&$path) {
+			if (($path !== '') &&
+				(\strpos($this->config->getSystemValue('share_folder', '/'), $path) !== false)) {
+				Util::writeLog("core", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
+				return false;
+			}
 			$absolutePath = $this->getAbsolutePath($path);
 			$mount = Filesystem::getMountManager()->find($absolutePath);
 			if ($mount->getInternalPath($absolutePath) === '') {
--- a/tests/lib/Files/ViewTest.php
+++ b/tests/lib/Files/ViewTest.php
@@ -7,6 +7,7 @@
 
 namespace Test\Files;
 
+use OC\AllConfig;
 use OC\Cache\CappedMemoryCache;
 use OC\Files\Cache\Watcher;
 use OC\Files\Filesystem;
@@ -80,6 +81,9 @@ class ViewTest extends TestCase {
 	/** @var \OC\Files\Storage\Storage */
 	private $tempStorage;
 
+	/** @var \OC\AllConfig */
+	private $config;
+
 	protected function setUp() {
 		parent::setUp();
 		\OC_Hook::clear();
@@ -104,9 +108,12 @@ class ViewTest extends TestCase {
 		Filesystem::clearMounts();
 
 		$this->tempStorage = null;
+
+		$this->config = $this->createMock(AllConfig::class);
 	}
 
 	protected function tearDown() {
+		$this->restoreService('AllConfig');
 		\OC_User::setUserId($this->user);
 		foreach ($this->storages as $storage) {
 			$cache = $storage->getCache();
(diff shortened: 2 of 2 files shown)


sharidas:
Queries I have here is :
- Do we need to DI for config in the Directory.php ?

jvillafanez:
I don't like this solution. Neither the directory nor the dav app doesn't need to know that the folder will be used to hold the shares. For them, it's just another folder without any special restriction.

If the files_sharing app is the one creating the folder, it should be the files_sharing app the one preventing the deletion. The app already has a storage wrapper available, so we can reuse that storage wrapper and prevent the folder deletion there.
This approach has several advantages:
* If all the sharing is disabled at some point, that folder will become empty (no shares are expected) and could be removed as long as the sharing app is disabled because the wrapper won't be registered any longer.
* If the files_sharing app decide to change the expected folder name to avoid collisions, this can be easily handled dinamically within the files_sharing app (including the storage wrapper)
* Config key handling is managed only by the one that creates / use the config key, not by any other app.

sharidas:
Looking at the code, the share_folder storage points to home storage and not the ShareStorage. And hence deletion call made from the trashbin storage causes the code flow reach Local storage. Debugged the storages available during the rename from the home folder to the trashbin, unfortunately sharedstorage is not there.

sharidas:
Looks like from the codeflow, the folder created using `share_folder` cannot be added to the sharedstorage, as it is not a share. The share is within the folder. Hence deletion of `share_folder` will not traverse through the sharedstorage. We can trim down the solution again to View only and remove the dav app change in the PR. 
Other than that I not able to find an apt solution to make it totally depend on the files_sharing app.
Any suggestion or pointers here would be appreciated.

jvillafanez:
Is it possible to use a mount point instead of creating a folder? it might be worthy to have a look and check if it's possible or not.

sharidas:
> Is it possible to use a mount point instead of creating a folder? it might be worthy to have a look and check if it's possible or not.

It looks like say if we use mount point for `share_folder`, then there is a chance that our code might fail at https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/SharedStorage.php#L103. Because `share_folder` is not a share.

jvillafanez:
Looks like we've run out of ideas.... it seems we'll have to go with this solution.

Do we need the changes in the dav app? I think the changes in the `View` should be enough. In addition, we should include a log somewhere in order to know that the folder deletion is being prevented on purpose.

sharidas:
> Do we need the changes in the dav app?

Yes you are right, we don't need the dav change. The View change should be enough. I have adjusted the PR for view only. Let me know if this change looks ok.

lib/private/Files/View.php
```diff
@@ -347,6 +352,11 @@ protected function renameUpdate(Storage $sourceStorage, Storage $targetStorage,
 	 */
 	public function rmdir($path) {
 		return $this->emittingCall(function () use (&$path) {
+			if (($path !== '') &&
+				(\strpos($this->config->getSystemValue('share_folder', '/'), $path) !== false)) {
+				Util::writeLog("files_sharing", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
```
jvillafanez: I think "core" as app instead of "files_sharing" makes more sense. The log is triggered by core and the code around is part of core.

sharidas: Updated the PR by using `core` instead of `files_sharing`.

tests/lib/Files/ViewTest.php
```diff
@@ -2670,4 +2670,22 @@ public function testFopenFail() {
...
+
+	/**
+	 * @dataProvider providesShareFolder
+	 */
+	public function testDeleteShareFolder($shareFolder, $deleteFolder) {
+		\OC::$server->getConfig()->setSystemValue('share_folder', $shareFolder);
+		$view = new View('/' . $this->user . '/files');
+		$view->mkdir(\OC::$server->getConfig()->getSystemValue('share_folder', '/'));
+		$this->assertFalse($view->rmdir($deleteFolder));
+		\OC::$server->getConfig()->deleteSystemValue('share_folder');
```
jvillafanez: Any better alternative to this cleanup? There are a couple of things I don't like:
* This will change the configuration without restoring the previous value. It will delete whatever value there was before.
* If the test fails for whatever reason, a different value for the "share_folder" will be kept. This might cause problems in other "not-so-unit" tests.

sharidas: I have updated the PR by using `overwriteService` in the test method. And using `restoreService` in the `tearDown`. The reason for using `overwriteService` in the method and not in the `setUp` is: there are methods which have call to config and that were causing issues. So I think its better to have it only for one method where we need it.

jvillafanez (approved):

HariBhandari07:
@sharidas I added ```'share_folder' => 'SharedFolder'``` in config.php but I can delete SharedFolder using the webUI.

phil-davis:
When I delete `SharedFolder` then all the received shares that are in `SharedFolder` go to the declined state. I can see them in "Shared with you". When I accept them again, then `SharedFolder` comes back and has the share(s) in it again.

What is the requirement here? (the behaviour above seems not so bad - you do not lose any data and can easily accept yyour shares again if you accidentallly delete them)

Before I receive any shares, there is no `SharedFolder` - that seems reasonable (it only gets created when the first share comes)

phil-davis:
Now I discovered that you have to put the leading `/` in the config:
`'share_folder' => '/SharedFolder'`
then it prevents deletion.

@sharidas that is not mentioned in the original post describing the manual testing?

PR #20609, opened 2015-11-19
Do not authenticate over ajax Webdav

PVince81:
This makes sure that whenever a Webdav call is done through Ajax, if the
session has expired, it will not send back a challenge but a simple 401
response. Without this fix, the default code would send back a challenge
and trigger the browser's basic auth dialog.

This will be needed when we use Webdav more often in the web UI, as in https://github.com/owncloud/core/pull/16902

To test, check out the branch from https://github.com/owncloud/core/pull/16902, kill the session and then attempt to navigate into a folder (it will use PROPFIND).
Before this fix: basic auth dialog shows from the browser.
After this fix: no browser dialog, redirects to login page.

@LukasReschke @DeepDiver1975 @nickvergessen @rullzer please review

PR #36003, opened 2019-08-08
Don't send WWW-Authenticate headers with schema Basic for ajax requests

DeepDiver1975:
## Description
This is basically a follow up of https://github.com/owncloud/core/pull/20609

As soon as an unauthenticated requests hits the server we are responding with a 401 and a WWW-Authenticate header with the auth schema 'Basic'.
Some browsers (like :nauseated_face: IE) are reacting to this with a password auth dialog which is disliked for ajax requests where the response is handled in the javascript code base.

A common way to identify an ajax request is to send X-Requested-With header to the server.

This was all basically implemented with https://github.com/owncloud/core/pull/20609

The owncloud-sdk (and as a result phoenix or and other browser based client) is using Bearer auth headers only and no session. Because of that the handling of X-Requested-With header has to be moved.

## Related Issue
https://github.com/owncloud/owncloud-sdk/pull/313

## How Has This Been Tested?
- [x] run sdk unit tests as of https://github.com/owncloud/owncloud-sdk/pull/313
- [ ] retest old web frontend with different browsers

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Auth.php
+++ b/apps/dav/lib/Connector/Sabre/Auth.php
@@ -153,8 +153,7 @@ class Auth extends AbstractBasic {
 	 */
 	public function check(RequestInterface $request, ResponseInterface $response) {
 		try {
-			$result = $this->auth($request, $response);
-			return $result;
+			return $this->auth($request, $response);
 		} catch (LoginException $e) {
 			throw new NotAuthenticated($e->getMessage(), $e->getCode(), $e);
 		} catch (NotAuthenticated $e) {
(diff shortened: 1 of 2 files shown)


apps/dav/lib/Connector/Sabre/Auth.php
```diff
@@ -237,13 +236,6 @@ private function auth(RequestInterface $request, ResponseInterface $response) {
 			}
 		}
 
-		if (!$this->userSession->isLoggedIn() && \in_array('XMLHttpRequest', \explode(',', $request->getHeader('X-Requested-With')))) {
```
jvillafanez: we should decide what to do if the user isn't logged in at this point. Note that below, there is a `$this->userSession->getUser()` which might return null.

DeepDiver1975: at this point if isLoggedIn() returns false really means: we have no session - this condition was correct for the webui but no longer for the future.

the missing null check will be added

phil-davis:
This still has the scality drone job in it. Needs a rebase to pick up yesterday's drone changes.

jvillafanez (approved):

phil-davis (approved):

PR #35932, opened 2019-07-26
Feature/public links

DeepDiver1975:
## Description
- Allow access to public links via webdav
- redirect public links to phoenix

## Related Issue
- Fixes #23269
- refs https://github.com/owncloud/phoenix/issues/270

## How Has This Been Tested?
- :construction: 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

PR #36007, opened 2019-08-09
Document phoenix.baseUrl in config.sample.php

DeepDiver1975:
## Description
The new config parameter phoenix.baseUrl is now documented in config.sample.php

## Related Issue
- refs #35932 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -442,6 +442,15 @@ $CONFIG = array(
 'overwrite.cli.url' => '',
 
 /**
+ * If phoenix.baseUrl is set, public and private links will be redirected to this
+ * url. Phoenix will handle these links accordingly.
+ *
+ * As an example, in case 'phoenix.baseUrl' is set to 'http://phoenix,example.com',
+ * the shared link 'http://ocx.example.com/index.php/s/THoQjwYYMJvXMdW' will be redirected
+ * by ownCloud to 'http://phoenix.example.com/index.html#//s/THoQjwYYMJvXMdW'.
+ */
+'phoenix.baseUrl' => '',
+/**
  * To have clean URLs without `/index.php` this parameter needs to be configured.
  *
  * This parameter will be written as `RewriteBase` on update and installation of


config/config.sample.php
```diff
@@ -441,6 +441,11 @@
  */
 'overwrite.cli.url' => '',
 
+/**
+ * If phoenix.baseUrl is set public and private links will be redirected to this
+ * url. Phoenix will handle these links accordingly.
```
mmattel: But only if the Phoenix App is loaded and enabled...
(Additional, add a comma for better readability: `is set public` --> `is set, public`)
Maybe to tell why this is necessary ? A brief reason would help.

DeepDiver1975: > But only if the Phoenix App is loaded and enabled...

Phoenix can operate independent from owncloud - there is no need in having it installed as an app

mmattel: confusing, as there is an app on the marketplace...

DeepDiver1975: To be precise: 
- phoenix can be installed as an owncloud app
- phoenix can be installed on any web space (basically when installed as an app nothing else then serving the web assets is done by owncloud/apache)

config/config.sample.php
```diff
@@ -441,6 +441,15 @@
  */
 'overwrite.cli.url' => '',
 
+/**
+ * If phoenix.baseUrl is set, public and private links will be redirected to this
+ * url. Phoenix will handle these links accordingly.
+ *
+ * As an example, in case 'phoenix.baseUrl' is set to 'http://phoenix,example.com,
```
settermjd: The URL needs a trailing single quote.

settermjd (changes requested):
Other than the missing, trailing, single quote, it looks good to me.

PR #36021, opened 2019-08-13
Don't list files on file drop folders + rename existing files when up…

DeepDiver1975:
…loading to file drop folders

## Description
- File drop folders (only share permission CREATE) shall not list folder contents.
- in case an uploaded file already exists the auto renaming will rename the file

## How Has This Been Tested?
- set phoenix.baseUrl in config.php
- create a file drop only share in oc10
- open public link in browser
- see the file drop screen
- upload files

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
+++ b/apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
@@ -27,6 +27,8 @@ use Sabre\DAV\PropFind;
 use Sabre\DAV\Server;
 use Sabre\DAV\ServerPlugin;
 use Sabre\DAV\Xml\Property\GetLastModified;
+use Sabre\HTTP\RequestInterface;
+use Sabre\HTTP\ResponseInterface;
 
 /**
  * Class PublicFilesPlugin - additional PROPFIND properties for public shared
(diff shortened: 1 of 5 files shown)


apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -48,6 +50,41 @@ public function initialize(Server $server) {
...
+
+		// only in share roots of file drop folders auto renaming will be applied
+		if (!$parent instanceof PublicSharedRootNode) {
+			return;
+		}
+		if (!$parent->isFileDropFolder()) {
+			return;
+		}
+
+		// node already exists at target path - the path will be rewritten
```
jvillafanez: For me is better to fail the request if the server won't answer anything. If I make a request uploading a file to /folder/file.txt , I expect the file to be there or at least the server to answer where the file is as part of the response in order to avoid asking the server again where the file is.
Not sure if we can change that behaviour at this point.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -48,6 +50,41 @@ public function initialize(Server $server) {
...
+		$newPath = $this->newFileName($path);
+		$request->setUrl($request->getBaseUrl() . $newPath);
+	}
+
+	private function newFileName($path) {
```
jvillafanez: Any chance to move this function to a helper class in core? we could use a callback or an interface to supply the "fileExists" check for the function.

apps/dav/lib/Files/PublicFiles/PublicSharedRootNode.php
```diff
@@ -57,6 +57,10 @@ public function __construct(IShare $share) {
 	 * @return INode[]
 	 */
 	public function getChildren() {
+		if ($this->isPropfind() && $this->isFileDropFolder()) {
```
jvillafanez: Maybe add a comment in the code to confirm this is intended? It will also help to know the use case that introduced this change, in order to prevent accidental removal.

apps/dav/lib/Files/PublicFiles/PublicSharedRootNode.php
```diff
@@ -151,4 +155,15 @@ public function getPermissions() {
 	protected function checkPermissions($permissions) {
 		return ($this->share->getPermissions() & $permissions) === $permissions;
 	}
+
+	private function isPropfind() {
+		return \OC::$server->getRequest()->getMethod() === 'PROPFIND';
```
jvillafanez: Any chance to inject this?

apps/dav/lib/Files/PublicFiles/SharedFolder.php
```diff
@@ -56,6 +56,10 @@ public function __construct(Folder $folder, IShare $share) {
 	 * @inheritdoc
 	 */
 	public function getChildren() {
+		if ($this->share->getPermissions() === Constants::PERMISSION_CREATE) {
```
jvillafanez: I think this really needs a comment. I don't know why this is needed, and without proper context this seems a bug.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
DeepDiver1975: Especially for the anonymous upload scenario the client and the user have no idea which files already do exists.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
DeepDiver1975: Too much hassle for 5 lines of code from my pov :man_shrugging:

jvillafanez (approved):

PR #36038, opened 2019-08-15
Fix invalid foreach in SharingBlacklist.php

### LINKED ISSUE #36037: Invalid argument supplied for foreach SharingBlacklist.php
### Steps to reproduce
1. run a test scenario like:
```
make test-acceptance-webui BEHAT_FEATURE=tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature:239
```

### Expected behaviour
No PHP messages in the log

### Actual behaviour
The log contains messages like:
```
{"reqId":"KJT4oRmZiVahfJHPQYgb","level":3,"time":"2019-08-15T05:08:31+00:00","remoteAddr":"172.17.0.3","user":"some-user","app":"PHP","method":"GET","url":"\/ocs\/v1.php\/apps\/files_sharing\/api\/v1\/sharees?format=json&search=case-sensitive-group&perPage=200&itemType=folder","message":"Invalid argument supplied for foreach() at \/home\/phil\/git\/owncloud\/core\/apps\/files_sharing\/lib\/SharingBlacklist.php#96"}
```

### Server configuration
Current git core master.

phil-davis:
## Description
The "string" value that it is set to is supposed to be valid JSON representing an array of group ids.

`blacklisted_receiver_groups` can be set "manually" by the administrator using an `occ` command, or by some automated test or... So it can end up being an empty string, or some string that is not valid JSON, or some string that represents a multi-level nested array-of-arrays, or some other weird structure.

1) If the JSON decode does not produce an array, then return an empty list. (JSON decode might have failed, or returned a plain string/int etc)

2) Check the elements of the decoded array. Keep elements that are "string-like" (string, int and float) and cast them to string. Discard other elements (e.g. an array, or boolean). This allows for group names like "0", "1", "1.23", "1.23e-4" that can look like both numbers and strings, while discarding other crud.

## Related Issue
- Fixes #36037 

## How Has This Been Tested?
Local run of acceptance test and observe that the log message no longer happens.
```
make test-acceptance-webui BEHAT_FEATURE=tests/acceptance/features/webUISharingInternalGroups/shareWithGroups.feature:239
```

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/SharingBlacklist.php
+++ b/apps/files_sharing/lib/SharingBlacklist.php
@@ -75,7 +75,7 @@ class SharingBlacklist {
 	 * @return string[] the list of group ids
 	 */
 	public function getBlacklistedReceiverGroups() {
-		return \json_decode($this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]'), true);
+		return \array_keys($this->fetchBlacklistedReceiverGroupIds());
 	}
 
 	private function initCache() {
@@ -88,14 +88,27 @@ class SharingBlacklist {
 		}
 	}
 
+	/**
+	 * @return bool[] an array with group ids as the keys, the values do not matter
+	 *                If the blacklisted receiver groups cannot be parsed as valid JSON,
+	 *                then an empty list is returned.
+	 *                Any parsed group ids that are not string, int or float are dropped from the list.
+	 */
 	private function fetchBlacklistedReceiverGroupIds() {
 		$configuredBlacklist = $this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]');
 		$decodedGroups = \json_decode($configuredBlacklist, true);
-		// expected a plain array here
+
+		if (!\is_array($decodedGroups)) {
+			$decodedGroups = [];
+		}
+
 		$groupSet = [];
 		foreach ($decodedGroups as $group) {
-			$groupSet[$group] = true;
+			if (\is_string($group) || \is_int($group) || \is_float($group)) {
+				$groupSet[(string) $group] = true;
+			}
 		}
+
 		return $groupSet;
 	}
 }
--- a/apps/files_sharing/tests/SharingBlacklistTest.php
+++ b/apps/files_sharing/tests/SharingBlacklistTest.php
@@ -67,6 +67,34 @@ class SharingBlacklistTest extends \Test\TestCase {
 		$this->assertEquals($ids, $this->sharingBlacklist->getBlacklistedReceiverGroups());
 	}
 
+	public function getBlacklistedReceiverGroupsProvider() {
+		return [
+			["", []],
+			["[\"invalid JSON missing right square bracket\"", []],
+			["[]", []],
+			["{\"abc\":\"group1\"}", ["group1"]],
+			["[[\"group1\"]]", []],
+			["[0]", ["0"]],
+			["[1]", ["1"]],
+			["[1.23]", ["1.23"]],
+			["[1.23e-4]", ["1.23e-4"]],
+			["[\"group1\"]", ["group1"]],
+			["[\"group1\", \"group2\", \"group3\"]", ["group1", "group2", "group3"]],
+		];
+	}
+
+	/**
+	 * Test cases for when the admin directly sets blacklisted_receiver_groups,
+	 * e.g. with an occ command, and might not set valid JSON, or might set valid
+	 * JSON but not in the expected structure.
+	 *
+	 * @dataProvider getBlacklistedReceiverGroupsProvider
+	 */
+	public function testGetBlacklistedReceiverGroups($ids, $result) {
+		$this->config->method('getAppValue')->willReturn($ids);
+		$this->assertEquals($result, $this->sharingBlacklist->getBlacklistedReceiverGroups());
+	}
+
 	private function getGroupMock($id, $displayname) {
 		$groupMock = $this->getMockBuilder(IGroup::class)
 			->disableOriginalConstructor()


apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -89,8 +93,7 @@ private function initCache() {
 	}
 
 	private function fetchBlacklistedReceiverGroupIds() {
-		$configuredBlacklist = $this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]');
-		$decodedGroups = \json_decode($configuredBlacklist, true);
+		$decodedGroups = $this->getBlacklistedReceiverGroups();
```
phil-davis: I refactored here, because this code was duplicating what `getBlacklistedReceiverGroups` was already supposed to be doing.

jvillafanez: I'd prefer to keep it like before. The reason is that this might mess up with the unittest code coverage.
The problem I see is that the public method `getBlacklistedReceiverGroups` will be called by a private method. This means that some code in the public method might not be tested while being marked as covered because of the private method's call.

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -75,7 +75,11 @@ public function setBlacklistedReceiverGroups(array $ids) {
 	 * @return string[] the list of group ids
 	 */
 	public function getBlacklistedReceiverGroups() {
-		return \json_decode($this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]'), true);
+		$configuredBlacklist = $this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]');
+		if ($configuredBlacklist === "") {
```
jvillafanez: I'd prefer to fail the method, either returning false or throwing an exception, if the decoded json isn't an array. As things are with this code, you also need to check if the app value returns a string, a number or whatever random value gets there, not just an empty string.

The code assumed that the `setBlacklistedReceiverGroups` would be use, so it has guarantee that an array will be stored.

apps/files_sharing/lib/SharingBlacklist.php
phil-davis: Or do I make a private `getBlacklistedReceiverGroupsXyz()` and call it from both `getBlacklistedReceiverGroups` and `fetchBlacklistedReceiverGroupIds` ?

jvillafanez: Works for me. The original code was only one line, so the effort wasn't worthy.

apps/files_sharing/lib/SharingBlacklist.php
phil-davis: Hmm - so somehow on my system I have ended up with the "empty list" being stored as the empty string, rather than the "empty JSON array".
I will see if I can make that happen for real using the webUI - if so, then that needs to be fixed anyway.

jvillafanez:
```
core/tests/acceptance/features/bootstrap/WebUIAdminSharingSettingsContext.php:
  353  		$this->featureContext = $environment->getContext('FeatureContext');
  354  		SetupHelper::runOcc(
  355: 			['config:app:set files_sharing blacklisted_receiver_groups --value=']
  356  		);
  357  	}
```
I guess that's the cause.

Taking into account that the admin can mess it up via occ, I think it's a good idea to harden the check.

PVince81:
@phil-davis any update ?

phil-davis:
> @phil-davis any update ?

I thought it was going to be an easy/obvious fix - silly me. Now I need to spend some time "hardening" against this setting being potentially any invalid JSON or weird JSON.

phil-davis:
@jvillafanez @PVince81 I switched around the logic so that the code is in the private method `fetchBlacklistedReceiverGroupIds` rather than previously, where I had the private method calling the public method.

I added a more general check using `json_last_error` and report the problem as an error log. But I go on by setting the group ids empty.

Please review.

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -88,10 +88,27 @@ private function initCache() {
 		}
 	}
 
-	private function fetchBlacklistedReceiverGroupIds() {
+	/**
+	 * @param bool $returnAnArrayWithGroupsAsValues if true then the group ids are the values in the array
```
jvillafanez: Text needs to be fixed

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -88,10 +88,27 @@ private function initCache() {
...
+	 */
+	private function fetchBlacklistedReceiverGroupIds($returnAnArrayWithGroupsAsValues = false) {
 		$configuredBlacklist = $this->config->getAppValue('files_sharing', 'blacklisted_receiver_groups', '[]');
 		$decodedGroups = \json_decode($configuredBlacklist, true);
-		// expected a plain array here
+
+		if (!\is_array($decodedGroups)) {
+			$decodedGroups = [];
+			$jsonError = \json_last_error();
+			\OC::$server->getLogger()->error('SharingBlacklist::getBlacklistedReceiverGroups - can\'t decode blacklisted_receiver_groups: ' . \print_r($configuredBlacklist, true) . ' - json error code: ' . $jsonError, ['app' => 'files_sharing']);
```
jvillafanez: just ignore the error message. This needs to be injected in order to include unittests, which is too much work for little benefit. Adding a comment in the PHPDocs saying that an empty list will be returned if we can't parse the json is enough.
Note that there is no need for anyone to know that we're storing the information as a json string.

jvillafanez:
Note that this needs unittests to cover this new case.

phil-davis:
@jvillafanez or anybody - ready for review again.
I went crazy and covered whatever crappy things I thought someone might be able to set this to.

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -88,12 +88,37 @@ private function initCache() {
...
-	private function fetchBlacklistedReceiverGroupIds() {
+	/**
+	 * @param bool $returnAnArrayWithGroupsAsValues if true then the group ids are the values in the array
+	 *                                              if false then the group ids are the keys in the array
+	 * @return string[] the list of group ids
+	 *                  If the blacklisted receiver groups cannot be parsed as valid JSON,
+	 *                  then an empty list is returned.
+	 *                  Any parsed group ids that are not string, int or float are dropped from the list.
+	 */
+	private function fetchBlacklistedReceiverGroupIds($returnAnArrayWithGroupsAsValues = false) {
```
jvillafanez: return the group set as originally planned, and use the `array_keys` function in the caller which needs to return the group set as list. This function shouldn't return different data

apps/files_sharing/tests/SharingBlacklistTest.php
```diff
@@ -63,8 +71,12 @@ public function testSetGetBlacklistedReceiverGroups($ids) {
 				return (isset($keyValues[$key])) ? $keyValues[$key] : $default;
 			}));
 
-		$this->sharingBlacklist->setBlacklistedReceiverGroups($ids);
-		$this->assertEquals($ids, $this->sharingBlacklist->getBlacklistedReceiverGroups());
+		if (\is_array($ids)) {
+			$this->sharingBlacklist->setBlacklistedReceiverGroups($ids);
+		} else {
+			$this->config->setAppValue('files_sharing', 'blacklisted_receiver_groups', $ids);
```
jvillafanez: I think this is a different test. Although the call is mocked, I don't really like to have "direct" calls to objects other than the one being tested.
You can test the same with something like:
```
$this->config->method('getAppValue')->willReturn($ids);
$this->assertEquals($result, $this->sharingBlacklist->getBlacklistedReceiverGroups());
```
This seems a more standard approach.

apps/files_sharing/tests/SharingBlacklistTest.php
```diff
@@ -40,7 +40,7 @@ public function setUp() {
 		$this->sharingBlacklist = new SharingBlacklist($this->config);
 	}
 
-	public function setGetBlacklistedReceiverGroupsProvider() {
+	public function setBlacklistedReceiverGroupsProvider() {
```
phil-davis: This existing set of unit test cases is really about testing that `setBlacklistedReceiverGroups`  works.
"by accident" it uses `getBlacklistedReceiverGroups` to check the result.

apps/files_sharing/tests/SharingBlacklistTest.php
```diff
@@ -67,6 +67,30 @@ public function testSetGetBlacklistedReceiverGroups($ids) {
 		$this->assertEquals($ids, $this->sharingBlacklist->getBlacklistedReceiverGroups());
 	}
 
+	public function getBlacklistedReceiverGroupsProvider() {
```
phil-davis: These test cases let us "manually" insert various strings as the blacklisted receiver groups config setting.
That happens in real life if the sysadmin "manually" sets the config entry with, for example, an `occ` command.

They check that for various "weird" strings of valid or invalid JSON the code does something reasonable.

phil-davis:
@jvillafanez I refactored stuff - please review again.

apps/files_sharing/lib/SharingBlacklist.php
```diff
@@ -88,12 +88,30 @@ private function initCache() {
...
 		foreach ($decodedGroups as $group) {
+			if (\is_string($group) || \is_int($group) || \is_float($group)) {
+				$validDecodedGroups[] = (string) $group;
+			}
+		}
+
+		// switch around so that the GroupIds are the keys of an array
+		$groupSet = [];
+		foreach ($validDecodedGroups as $group) {
 			$groupSet[$group] = true;
```
jvillafanez: you can generate the group set directly during the first iteration of the groups.

apps/files_sharing/tests/SharingBlacklistTest.php
jvillafanez: Technically the test checks that you get what you set. It's intentional. I agree it isn't an unittest, but I don't think it's worthy to check that a oneline function call one specific method. I think the previous name make this more evident (or not)

apps/files_sharing/tests/SharingBlacklistTest.php
jvillafanez: If you think this is worthy to know, I think it's better to write a code comment. Personally, I'd rather not to do archeology to know why the test was created or any other important thing

jvillafanez (commented):
one last thing and we're done

apps/files_sharing/tests/SharingBlacklistTest.php
phil-davis: Description added to PHPdoc of method

apps/files_sharing/tests/SharingBlacklistTest.php
phil-davis: I put the original name back

apps/files_sharing/lib/SharingBlacklist.php
phil-davis: yes, good idea.

phil-davis:
@jvillafanez ready for review again

jvillafanez (approved):

sharidas:
@phil-davis I am merging this change.

HanaGemela:
Cannot share to blacklisted group with group names like "0", "1", "1.23", "1.23e-4" on the desktop sync client 2.6.0 RC1, macOS 10.14.6 and testpilot 2.5.4

PR #36032, opened 2019-08-14
Check status on given steps in Comments and WebDav acceptance test code

phil-davis:
## Description
1) For acceptance steps that have both a `When` and a `Given` form, we want to be (reasonably) confident that `Given` steps have actually worked. Make separate methods for `Given` steps that call the original `When` step and then check that it claims to have worked - usually just check that a good HTTP status was returned. This helps test scenarios to fail early if a `Given` step goes wrong. Without doing this, it becomes confusing when seeing later failures in the scenario, if you are not aware that actually something in the `Given` setup steps went wrong. This is done for `Comments` and `WebDav` steps in this PR.

2) Fixup scenarios in `quota.feature` that were not really doing quite what it looked like. These were noticed because they started failing after the checks added at (1).

3) Remove an unneeded step from `shareWithGroups.feature` that was noticed because it started failing after the checks added at (1).

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #36039, opened 2019-08-15
Make separate 'given' steps  in AppManagementContext Auth BasicStructure

phil-davis:
## Description
Separate the `Given` steps and check the status of API calls that are done, so that we fail early when `Given` setup goes wrong.

This is "more of the same" like PR #36032 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Auth.php
+++ b/tests/acceptance/features/bootstrap/Auth.php
@@ -19,6 +19,7 @@
  *
  */
 
+use PHPUnit\Framework\Assert;
 use TestHelpers\HttpRequestHelper;
 use Behat\Gherkin\Node\TableNode;
 
@@ -86,7 +87,6 @@ trait Auth {
 
 	/**
 	 * @When a user requests :url with :method and no authentication
-	 * @Given a user has requested :url with :method and no authentication
 	 *
 	 * @param string $url
 	 * @param string $method
@@ -95,6 +95,19 @@ trait Auth {
 	 */
 	public function userRequestsURLWith($url, $method) {
 		$this->sendRequest($url, $method);
+	}
+
+	/**
+	 * @Given a user has requested :url with :method and no authentication
+	 *
+	 * @param string $url
+	 * @param string $method
+	 *
+	 * @return void
+	 */
+	public function userHasRequestedURLWith($url, $method) {
+		$this->sendRequest($url, $method);
+		$this->theHTTPStatusCodeShouldBeSuccess();
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/bootstrap/Auth.php
```diff
@@ -331,9 +344,10 @@ public function aNewClientTokenHasBeenGenerated($user) {
 		);
 		$headers = ['Content-Type' => 'application/json'];
 		$url = $this->getBaseUrl() . '/token/generate';
-		$resp = HttpRequestHelper::post($url, null, null, $headers, $body);
+		$this->response = HttpRequestHelper::post($url, null, null, $headers, $body);
+		$this->theHTTPStatusCodeShouldBe("200");
```
phil-davis: This check is done for both `When` and `Given` because actually if the `post` went wrong in either case, then there is not point trying to extract a token from in the response.

tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -994,6 +1021,40 @@ public function theHTTPStatusCodeShouldBeOr($statusCode1, $statusCode2) {
...
 
+	/**
+	 * @Then /^the HTTP status code should be between "(\d+)" and "(\d+)"$/
+	 *
+	 * @param int $minStatusCode
+	 * @param int $maxStatusCode
+	 *
+	 * @return void
+	 */
+	public function theHTTPStatusCodeShouldBeBetween(
```
phil-davis: This seemed handy to have. We already had a similar step in `WebDav.php`
```
Then /^the HTTP status code of all upload responses should be between "(\d+)" and "(\d+)"$/
```
so I did similar here for a single response.

tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -994,6 +1021,40 @@ public function theHTTPStatusCodeShouldBeOr($statusCode1, $statusCode2) {
...
+			$message
+		);
+	}
+
+	/**
+	 * @Then the HTTP status code should be success
+	 *
+	 * @return void
+	 */
+	public function theHTTPStatusCodeShouldBeSuccess() {
```
phil-davis: And this seemed the way to check "general" API calls where we do not know exactly what the status should be, but we do know that in all `Given` steps the status should be "success" `2xx`

tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -994,6 +1021,40 @@ public function theHTTPStatusCodeShouldBeOr($statusCode1, $statusCode2) {
...
+		$this->theHTTPStatusCodeShouldBeBetween("200", "299");
```
individual-it:
```suggestion
		$this->theHTTPStatusCodeShouldBeBetween(200, 299);
```

individual-it (changes requested):

phil-davis:
@individual-it please review again - see 2nd commit.

individual-it (approved):

Issue #35983, opened 2019-08-06
Share Notifications API


DeepDiver1975:
## Background
In Phoenix shared links shall be send via email. Therfore an api is missing.

There is a notify route already:

https://github.com/owncloud/core/blob/5ee2ac506181a52192889b511fe0f2be71bb3780/apps/files_sharing/appinfo/routes.php#L91

But it only handles emails to internal users (which is no longer needed?) @pmaier1 

oc10 uses the old ajax endpoint for sending emails

https://github.com/owncloud/core/blob/5ee2ac506181a52192889b511fe0f2be71bb3780/core/ajax/share.php#L81

## Actions
- Move logic from share.php over to a new ocs controller which handles email notifications (share20ocscontroller is already pretty hugh)
- decide if we still need the internal email notification mechanism
- add unit tests
- add acceptance tests
- add sdk api -> https://github.com/owncloud/owncloud-sdk/
- adjust oc10 web frontend to use the new api
- kill code in share.php

## Stretch Goal
- analyse if the fetch operation can be replaced with a call to the sharee api
https://github.com/owncloud/core/blob/5ee2ac506181a52192889b511fe0f2be71bb3780/core/ajax/share.php#L232

pmaier1:
> But it only handles emails to internal users (which is no longer needed?) @pmaier1

Emails to internal users does not make much sense, yes, as you should rather use user shares for those.

DeepDiver1975:
shall we kill the server side code for internal email delivery? if unused code shall be killed @pmaier1

pmaier1:
> shall we kill the server side code for internal email delivery? if unused code shall be killed @pmaier1

I don't know whether this is needed somewhere. Would need more background / grasp the whole picture.

DeepDiver1975:
The question is: shall we kill this feature
![Screenshot from 2019-08-14 16-31-26](https://user-images.githubusercontent.com/1005065/63029534-08583c80-beb1-11e9-984f-8daa186a7a70.png)
![Screenshot from 2019-08-14 16-31-06](https://user-images.githubusercontent.com/1005065/63029535-08583c80-beb1-11e9-9c63-dcd5645d14d3.png)

pmaier1:
Ahh, I didn't get that one. I thought this issue is about public links and their "send via mail" functionality?!

The above one is still needed for the following reason:
[...]

PR #36063, opened 2019-08-19
add ocs api for public link share email notification

karakayasemi:
## Description
This PR adds an OCS API for public link e-mail notifications. old ajax endpoints removed and oc10 adjusted to use the new API.

I aimed to keep the same functionality as oc10. In addition, the current code is simplified by removing redundant post parameters:
- The old endpoint is taking many pieces of information related to the share of link as POST parameter like expiration date, filename, itemType, etc. It is also trying to validate them on the backend. However, we can already extract these parameters by reaching related share object on the backend. So, I considered them redundant and, removed.
- I could not find any usage of `cc` and `bcc` fields in the codebase. So, I removed these options regards to YAGNI principle.
- When `toSelf` parameter is posted, the backend side is only adding user's e-mail to the recipient list. I moved this logic to the client-side and removed `toSelf` from POST parameters.

### UPDATE
- Internal email notification mechanism also moved to the new controller.

## Usage
* **URL**
/ocs/v1.php/apps/files_sharing/api/v1/notification/notifypubliclinkwithemail

* **Method:**
  `POST`
  
*  **POST Params**
   **Required:**
   `recipients=[string[]]`
   `link=[string]`
   `personalNote=[string]`
```
* **Success Response:**
{"ocs":{"meta":{"status":"ok","statuscode":100,"message":"OK","totalitems":"","itemsperpage":""},"data":[]}}
 
* **Fail Response:**
{"ocs":{"meta":{"status":"error","statuscode":400,"message":"Couldn't send mail to following recipient(s): test@email.com","totalitems":"","itemsperpage":""},"data":[]}}
```
## Related Issue
- Part of #35983

## Motivation and Context
In Phoenix, shared links shall be sent via email. We need an API for that.

## How Has This Been Tested?
- Unit tests are added.
- Manually by sending public link e-mails from the web interface.
- Manually with API call.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


apps/files_sharing/appinfo/routes.php
```diff
@@ -96,6 +96,11 @@
 			'url' => '/api/v1/notification/marksent',
 			'verb' => 'POST'
 		],
+		[
+			'name' => 'Notification#notifyPublicLinkRecipientsWithEmail',
+			'url' => '/api/v1/notification/notifypubliclinkwithemail',
```
DeepDiver1975: maybe kebab style looks better? notify-public-link-by-email
Or pascal notifyPublicLinkByEmail

apps/files_sharing/lib/Controller/NotificationController.php
```diff
@@ -0,0 +1,119 @@
...
+use OCP\IL10N;
+use OCP\ILogger;
+use OCP\IRequest;
+use OCP\IURLGenerator;
+use OCP\IUserSession;
+use OCP\Mail\IMailer;
+use OCP\Share\IManager;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+/**
```
DeepDiver1975: please move to the beginning of the file

apps/files_sharing/lib/Controller/NotificationController.php
```diff
@@ -0,0 +1,119 @@
...
+	 *
+	 * @param string $link
+	 * @param string $recipients
+	 * @param string $personalNote
+	 * @return Result
+	 */
+	public function notifyPublicLinkRecipientsWithEmail($link, $recipients, $personalNote) {
+		$message = null;
+		$code = 100;
+		$mailNotification = new MailNotifications(
```
DeepDiver1975: maybe simplify this be injecting class MailNotifications?

apps/files_sharing/lib/Controller/NotificationController.php
```diff
@@ -0,0 +1,119 @@
...
+		$this->urlGenerator = $urlGenerator;
+		$this->eventDispatcher = $eventDispatcher;
+		$this->defaults = $defaults;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @param string $link
+	 * @param string $recipients
```
DeepDiver1975: string[] ??

apps/files_sharing/lib/Controller/NotificationController.php
```diff
@@ -0,0 +1,119 @@
...
+		$this->userSession = $userSession;
+		$this->urlGenerator = $urlGenerator;
+		$this->eventDispatcher = $eventDispatcher;
+		$this->defaults = $defaults;
+	}
+
+	/**
+	 * @NoAdminRequired
+	 *
+	 * @param string $link
```
DeepDiver1975: ~This way this API could be used for sending any data to anybody. We might want to change the api to reference a share id and also check if the current use is allowed to access the share.~

apps/files_sharing/lib/Controller/NotificationController.php
DeepDiver1975: also add type hint array for this argument

DeepDiver1975 (changes requested):

DeepDiver1975:
- [ ] The email subject did change. Before this change: 'Admin shared foo.txt with you' After the change: 'Admin shares public link with you' - link name va file name
- [ ] ^ same with the email body
- [ ] unrelated to this change but found while testing: the subject is html escaped - which is bad
Screenshot from 2019-08-19 16-50-31

apps/files_sharing/lib/Controller/NotificationController.php
karakayasemi: If we refactor this class, it will be unit testable. Since Internal email notification part of the code is also using this class. I can move internal email notification to the new controller after refactoring.

apps/files_sharing/lib/Controller/NotificationController.php
karakayasemi: its type is string. if it has multiple emails, they are separated with a comma. `MailNotifications` class is accepting input in this way. If you prefer in array type, I can refactor it also.

DeepDiver1975: Please at least add a comment then - THX a lot

DeepDiver1975:
- [ ] acceptance tests need to be adjusted I guess - see https://drone.owncloud.com/owncloud/core/20273/76/14

apps/files_sharing/lib/Controller/NotificationController.php
karakayasemi: resolved

apps/files_sharing/appinfo/routes.php
karakayasemi: now it is kebab style.

apps/files_sharing/lib/Controller/NotificationController.php
karakayasemi: After refactoring, now it takes `string[]`. Doc updated.

karakayasemi:
@DeepDiver1975 Reason behind the e-mail subject and, body change is related to my mistake. I mistakenly used `$share->getName()` as file name. Now, I corrected it. Html escape issue is also related to this mistake. Since it's no longer needed, I've reverted your last commit changes.

Acceptance test fail is also related to this mistake. I hope this time, it will pass. Please, review the PR one more time. Thanks.

apps/files_sharing/lib/Controller/NotificationController.php
karakayasemi: `MailNotifications` class is now injectable. Internal sharing email notification mechanism also moved to the new controller. PR description updated to according to this.

DeepDiver1975:
> Html escape issue is also related to this mistake. Since it's no longer needed, I've reverted your last commit changes.

If the file name holds e.g. äöü - the subject is not escaped? This issue also existed before this PR - please double check. THX

apps/files_sharing/lib/Controller/NotificationController.php
```diff
@@ -0,0 +1,169 @@
...
+		}
+		// don't send a mail to the user who shared the file
+		$recipientList = \array_filter($recipientList, function ($user) {
+			/** @var IUser $user */
+			return $user->getUID() !== $this->userSession->getUser()->getUID();
+		});
+
+		$userFolder = $this->rootFolder->getUserFolder($sender->getUID());
+		$nodes = $userFolder->getById($itemSource, true);
+		$node = $nodes[0] ?? null;
```
jvillafanez: `$node = $nodes[0]` ? I don't think we need more than that...
In addition, what happens with the rest of the nodes? Consider to add a comment explainig.

karakayasemi: @jvillafanez in the beginning, this pr aimed only to add ocs api for public link email notification. However, I had to refactor `MailNotifications` class to make it injectable and testable. 
The only other consumer of this class was this method.  Since `Share20OcsController` already pretty big, that's why I did not want to inject new dependency to it. To keep the change minimum, I just moved this method from `Share20OcsController` to the new Notification Controller and, adapted the code to use injected `MailNotifications` class. So, I did not interest with the logic behind this method. I guess this node will be used for checking both sender and recipient have this node  :confused:.

jvillafanez: we'll have to check it at some point. It's ok for now.

jvillafanez (approved):

settermjd:
@karakayasemi, I've not been able to use the API successfully, as yet. The only response I get is the following:

```xml
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>403</statuscode>
  <message>Public link mail notification is not allowed</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data/>
</ocs>
```

That said, the following two configuration options are in effect:

```php
'shareapi_allow_public_notification' => 'yes',
'dav.enable.tech_preview' => true,
```

Any suggestions?

karakayasemi:
image

Please enable the option that I highlighted in the image from admin sharing settings.

settermjd:
Thanks for the clarification. I had thought that setting the config value would be sufficient.

settermjd:
It seems I don't have the same version as you do. 

Screenshot_2019-10-04 Settings - ownCloud

settermjd:
I'm running 10.3.0 alpha2 (cloned from master).

karakayasemi:
There is another option appearing after allowing users to share via link.

PR #36079, opened 2019-08-22
IE doesn't redirect to login page when user is not logged in

VicDeo:
## Description
send html response when the client says any content-type  is acceptable

## Motivation and Context
0. IE 11 is set as a default browser
1. ownCloud User on Windows clicks on ownCloud private permalink (included in e.g. email) 

# Expected behavior
 IE opens and login page is presented

# Actual behavior
IE opens and user receives download dialog to download a file of 43 bytes (which is the uid of the user)
User needs to change URL to point to ../login.php and authenticate there

## How Has This Been Tested?
Not tested yet

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


PVince81 (approved):
:+1:

PR #36086, opened 2019-08-24
Fix bug with attribute disabled being saved as null

mrow4a:
This commit (https://github.com/owncloud/core/commit/2ef190a4fce0d0ab6173da1304ffaa35e95ede95) broke the logic of share attributes in case where share attribute "enabled" was set to false. 

This fixes it by checking for null instead.


PVince81 (approved):
:+1:

phil-davis:
Fixes #36090 which I noticed (richdocuments tests for extra attributes were failing)

PR #36105, opened 2019-08-27
Add password enforcement for Download/View/Edit option and add capability

sharidas:
Add password enforcement for Download/View/Edit
option for the public share.
Also arranged the order of the permission, options
that come when user creates the public share.

## Description

This PR comprises of 2 commits. The first commit focuses on the change done in re-arranging the display of permissions displayed when user creates a new public link, that is the dialog window. Or say the user edits the dialog window ( both windows are same ).
Here is the detail about the first commit:
Updated the order of how permissions of public link appears when user creates a new share or edits the existing share. The current order according to this PR is:
1. `Download / View`
2. `Download / View / Upload`
3. `Download / View / Edit`
4. `Upload only ( File Drop )`

This change updates the description of permission `Download / View / Edit` to `Recipients can view, download, edit, delete and upload contents.`.

This change also includes password enforcement for all the permission. Because of the introduction of new permission to public link, the password enforcement change was not done. This change brings that too.

Detail about second commit:
The second commit uses the roles api introduced in the server. So in this commit, I used core/js/config.php to fetch the details about roles provided by core. And this value is passed across the configmodel which is used by the respective frontend. I gave a thought about using ajax request to get the roles details. But then this data has to be saved and we should not do query again and again as the data wouldn't change in the server that frequently. So I thought the good place to have the roles details in core/js/config.php. 
And then I have introduced functions like `populateData` in the sharedialoglinkshareview.js. The idea is to populate the role data received from the server and pass it to the handlebar template.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Re-arrange the order in which the permission has to be displayed in the public link.
Adjust the description text for `Download / View / Edit`.
And make sure password enforcement is available for all the permissions of public link share.

## How Has This Been Tested?

- Tested the password enforcement by enabling the checkbox in the sharing page ( of admins settings page )
     - Verified that public links created for `Download / View` with password set ( because its checked in the Sharing page for testing), can be opened with the same password provided.
     - Verified that public links created for `Download / View / Upload` with password set ( because its checked in the Sharing page for testing), can be opened with the same password provided.
     - Verified that public links created for `Download / View / Edit` with password set ( because its checked in the Sharing page for testing), can be opened with the same password provided.
     - Verified that public links created for `Upload only` with password set ( because its checked in the Sharing page for testing), can be opened with the same password provided.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/js/config.php
+++ b/core/js/config.php
@@ -69,6 +69,7 @@ if ($defaultExpireDateEnabled) {
 $enforceLinkPasswordReadOnly = $config->getAppValue('core', 'shareapi_enforce_links_password_read_only', 'no') === 'yes';
 $enforceLinkPasswordReadWrite = $config->getAppValue('core', 'shareapi_enforce_links_password_read_write', 'no') === 'yes';
 $enforceLinkPasswordWriteOnly = $config->getAppValue('core', 'shareapi_enforce_links_password_write_only', 'no') === 'yes';
+$enforceLinkPasswordReadWriteDelete = $config->getAppValue('core', 'shareapi_enforce_links_password_read_write_delete', 'no') === 'yes';
 $outgoingServer2serverShareEnabled = $config->getAppValue('files_sharing', 'outgoing_server2server_share_enabled', 'yes') === 'yes';
 
 $countOfDataLocation = 0;
@@ -171,6 +172,7 @@ $array = [
 				'defaultExpireDateEnforced' => $enforceDefaultExpireDate,
 				'enforceLinkPasswordReadOnly' => $enforceLinkPasswordReadOnly,
 				'enforceLinkPasswordReadWrite' => $enforceLinkPasswordReadWrite,
+				'enforceLinkPasswordReadWriteDelete' => $enforceLinkPasswordReadWriteDelete,
 				'enforceLinkPasswordWriteOnly' => $enforceLinkPasswordWriteOnly,
 				'sharingDisabledForUser' => \OCP\Util::isSharingDisabledForUser(),
 				'resharingAllowed' => \OCP\Share::isResharingAllowed(),
--- a/core/js/config.php
+++ b/core/js/config.php
@@ -72,6 +72,10 @@ $enforceLinkPasswordWriteOnly = $config->getAppValue('core', 'shareapi_enforce_l
 $enforceLinkPasswordReadWriteDelete = $config->getAppValue('core', 'shareapi_enforce_links_password_read_write_delete', 'no') === 'yes';
 $outgoingServer2serverShareEnabled = $config->getAppValue('files_sharing', 'outgoing_server2server_share_enabled', 'yes') === 'yes';
 
+$container = \OC::$server->getAppContainer('files_sharing');
+$rolesController = $container->query('OC\Core\Controller\RolesController');
+$roles = $rolesController->getRoles();
+
 $countOfDataLocation = 0;
 
 $value = $config->getAppValue('core', 'shareapi_enable_link_password_by_default', 'no');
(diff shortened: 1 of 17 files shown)


PVince81:
> Add password enforcement for Download/View/Edit option

"capability" term is missing in PR title

At first I believed you were adding enforcement...

sharidas:
Added `capability` term to the PR title.

core/js/config.php
```diff
@@ -69,8 +69,13 @@
 $enforceLinkPasswordReadOnly = $config->getAppValue('core', 'shareapi_enforce_links_password_read_only', 'no') === 'yes';
 $enforceLinkPasswordReadWrite = $config->getAppValue('core', 'shareapi_enforce_links_password_read_write', 'no') === 'yes';
 $enforceLinkPasswordWriteOnly = $config->getAppValue('core', 'shareapi_enforce_links_password_write_only', 'no') === 'yes';
+$enforceLinkPasswordReadWriteDelete = $config->getAppValue('core', 'shareapi_enforce_links_password_read_write_delete', 'no') === 'yes';
 $outgoingServer2serverShareEnabled = $config->getAppValue('files_sharing', 'outgoing_server2server_share_enabled', 'yes') === 'yes';
 
+$container = \OC::$server->getAppContainer('files_sharing');
```
jvillafanez: we should find a better way...
* core doesn't need to know there is a "files_sharing" app
* "files_sharing" app can be disabled
* I don't have any idea why the "OC\Core\Controller\RolesController" is inside the "files_sharing" app container.

In addition, this solution doesn't seem extensible enough in case we have plans for external apps to add custom roles.

core/js/tests/specs/sharedialoglinklistviewSpec.js
```diff
@@ -29,17 +29,31 @@ describe('OC.Share.ShareDialogLinkListView', function() {
...
 	var showPopupStub;
 	var publicLinkStub;
+	var roles;
 
 	beforeEach(function() {
+		roles = {
+			"ocs":{
+				"data": {
+					"data":[
+						{"id":"core.viewer","displayName":"Download \/ View","context":{"publicLinks":{"displayDescription":"Recipients can view or download contents.","order":10,"resourceTypes":["*"],"permissions":{"ownCloud":{"read":true}}}}},
```
jvillafanez: Reformat these lines because they can't be read easily

core/js/sharedialoglinkshareview.js
```diff
@@ -298,6 +267,69 @@
...
+
+				fileNameLabel              : t('core', 'Filename'),
+				passwordLabel              : t('core', 'Password'),
+
+				publicUploadPossible       : this._isPublicUploadPossible(),
+
+				isMailEnabled: showEmailField
+			};
+
+			for (var i = 0; i < _.size(data); i++) {
```
jvillafanez: try to move this to a different function. In addition, the "if" blocks seems pretty similar, so maybe we can refactor that part too.

core/js/tests/specs/sharedialoglinkshareviewSpec.js
```diff
@@ -26,12 +26,22 @@ describe('OC.Share.ShareDialogLinkShareView', function() {
...
 	var view;
+	var roles;
 
 	var PASSWORD_PLACEHOLDER_STARS = '**********';
 	var PASSWORD_PLACEHOLDER_MESSAGE = 'Choose a password';
 
 	beforeEach(function() {
+		roles = {
+					"data":[
+						{"id":"core.viewer","displayName":"Download \/ View","context":{"publicLinks":{"displayDescription":"Recipients can view or download contents.","order":10,"resourceTypes":["*"],"permissions":{"ownCloud":{"read":true}}}}},
```
jvillafanez: same here. The indentation is wrong also.

PVince81:
deferred as this goes further than planned, need to think this through

settermjd:
> deferred as this goes further than planned, need to think this through

@PVince81, based on the above comment, is this PR effectively on hold?

PVince81:
@sharidas please link to the new PR

DeepDiver1975:
old aka dead -> close

PR #36110, opened 2019-08-28
if a log-entry does not contain an attribute mark it at not-matching

individual-it:
## Description
if the log-entry would not have that searched attribute at all it would not be marked as not-matching because of the `isset`

## Motivation and Context
found while writing tests for admin_audit

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


phil-davis (approved):
As I understand the previous code, it could sometimes wrongly think that a log entry matched. So it would give "false fails".

PR #36144, opened 2019-09-03
compare shareowner and session user when detecting re-share

### LINKED ISSUE #36107: [10.3.0 alpha] Owner cannot edit group's permissions
Works with 2.6.0-daily20190820 (build 12341), macOS 10.14.6, server 10.2
Broken on 2.6.0-daily20190820 (build 12341), macOS 10.14.6, server 10.3.0 alpha
Broken on 2.5.4, Windows 10, server 10.3.0 alpha

Steps to recreate:
1) User1 shares a file to User2 with only Share permission
2) User2 shares a file to group friends (User1 and User2 are both members of this group) without any permissions
3) User1 adds Edit permissions to group friends

Actual result: The entry gets greyed out and cannot be edited again
Expected result: User1 can again immediately remove the Edit permissions

Screenshot 2019-08-23 at 16 18 04

karakayasemi:
## Description
Shareowner should have full permission on re-shares. For detecting re-share, this PR compares the session user UID with the shareowner UID as described in this comment: https://github.com/owncloud/core/issues/36107#issuecomment-525170914. 

## Related Issue
- Fixes #36107 

## Motivation and Context
Fighting with bugs.

## How Has This Been Tested?
Manually with steps in #36107. Also, unit tests are adjusted.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -886,7 +886,8 @@ class Server extends ServerContainer implements IServerContainer, IServiceLoader
 				$c->getLazyRootFolder(),
 				$c->getEventDispatcher(),
 				new View('/'),
-				$c->getDatabaseConnection()
+				$c->getDatabaseConnection(),
+				$c->getUserSession()
 			);
 
 			return $manager;
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -40,6 +40,7 @@ use OCP\IGroupManager;
 use OCP\IL10N;
 use OCP\ILogger;
 use OCP\IUserManager;
+use OCP\IUserSession;
 use OCP\Security\IHasher;
 use OCP\Security\ISecureRandom;
 use OCP\Share\Exceptions\GenericShareException;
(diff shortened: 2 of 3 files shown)


karakayasemi:
Injecting `UserSession` to `ShareManager` looks like not a good idea. I faced many issues when applying this approach. Currently, several integration tests are failing and also I can not use scalar type declaration in `ShareManager` constructor, because UserSession can be null in some cases. IMHO, touching lots of tricky places like DI Container is too risky for this issue.

In addition, as stated in here https://github.com/owncloud/core/issues/36107#issuecomment-527816932
>
>2. a higher-level file/share owner can edit the permissions of a lower-level reshare to be greater than the permissions given in the original share - even though the original share still has the lower permissions. (contentious - IMO they should have to increase the permissions of the original share first, then they can increase the permissions of the reshare by the other user)

aimed behavior is contentious. The current behavior is more close to @pdavis's opinion. We should continue to discuss requirements under https://github.com/owncloud/core/issues/36107. I am closing this PR because of the mentioned problems.

PR #35972, opened 2019-08-02
Move user creation to a separate service

sharidas:
Move user creation to a separate service.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

## Description

Move the user creation to the service class.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Add the create user to a service class. It is moved from user_management apps UsersController. This will help us in integrating the change with the `user:add` command support creating users with username and email address. It will also help for provisioning API.

## How Has This Been Tested?

- Verified that from webUI users can be created using users page, with and without email address
- Verified that from webUI users can be created using users page and while creation the user could be added to the group. Tested with and without email address.
- Verified in the settings page of the logged in user and in the users page of admin that the group associated with the user is reflected.
- Verified that using `user:add` command the users are created with email address and/or with password. For example:
```
sujith@sujith-ownCloud  ~/test/owncloud   move-createuser-toservice-v2 ●  ./occ user:add --display-name "Test User5" --email "***********" --group "group1" --group "group2" user5
Cannot load Xdebug - it was already loaded
The user "user5" was created successfully
Display name set to "Test User5"
Email address set to "************"
User user5 added to group group1
User user5 added to group group2
 sujith@sujith-ownCloud  ~/test/owncloud   move-createuser-toservice-v2 ●  
```
- In the above example `group1` and `group2` were created and the `user5` is added to the group
- An email is sent to to the user to set the password, if the email is configured in the sever.
- Verified that user can set the pasword.
- Verified using provisioningapi the users are created with and without email address, for example:
```
sujith@sujith-ownCloud  ~/test/owncloud   move-createuser-toservice-v2 ●  curl -X POST http://admin:admin@localhost/testing/ocs/v1.php/cloud/users -d userid="test6" -d email='********' -d groups\[\]="group1" -d groups\[\]="group2"                           
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
 sujith@sujith-ownCloud  ~/test/owncloud   move-createuser-toservice-v2 ● 
```
- From the above example verified that password could be reset from the email received.
- Another example using password:
```
[...]

Issue #36020, opened 2019-08-13
Words for "Unable to create user due to exception"


phil-davis:
PR #35972 added the text `Unable to create user due to exception:` to the front of error messages when creating a user.

1) There is a PR in the `user_management` app https://github.com/owncloud/user_management/pull/196 that makes the core changes happen to the `user_management` UI. That has not been merged yet, because the `user_management` app is "in limbo" right now. In the past, "normally", that would have been merged. Then the core `master` PR would be backported to `stable10` and the `user_management` app changes would also be "backported" into core `stable10`. Now that we do not have `stable10`, IMO the process should be that the changes proposed in https://github.com/owncloud/user_management/pull/196 should be applied to core `master`.

2) If we do (1), then the error messages shown to the user when creating a new user look like:
```
Error creating user: Unable to create user due to exception: The password contains too few lowercase letters. At least one lowercase letter is required.
```
That has excess text. The `user_management`  text `Error creating user:` could be removed.

But the words "due to exception" IMO are not very nice to present to a user (even though they are the admin user). Maybe the words used in `core` can be just `Unable to create user: `
Then the message that comes in OCS status and to real users on the UI would be:
```
Unable to create user: The password contains too few lowercase letters. At least one lowercase letter is required.
```

3) The text `Unable to create user due to exception` is not getting translated. So if the admin user has set their language to something other than English, the message they get has `Unable to create user due to exception` in English and `The password contains too few lowercase letters. At least one lowercase letter is required` in their own language.

phil-davis:
@micbar @DeepDiver1975 @sharidas 
What to do? Maybe nothing?

pmaier1:
Thanks for the catch @phil-davis :+1: . I agree with 2)

PVince81:
@phil-davis closed through https://github.com/owncloud/core/pull/36022 or further adjustments needed ?

phil-davis:
What to do about??

3. `Unable to create user` is not translated - that will result in, for example, password policy messages that are partly in English and partly in the selected local language. Maybe this is really a bigger question about translation of exception message text (which sometimes ends up in a log file or other  "technical" place, and sometimes is used to generate a message to a human user)

4. Do we want to have a similar format message for when a user modify cannot be done.

PVince81:
@sharidas any reason why the texts have changed after the refactoring, like with https://github.com/owncloud/core/issues/36019 ? any chance to align them and make them look like what they were before ?

Any further changes and translations would need to be done separately later.

[...]

Issue #36019, opened 2019-08-13
Behaviour changes due to "Move user creation to a separate service"


phil-davis:
After PR #35972 was merged to core master, password_policy tests started to fail due to some small changes in error reporting behaviour. See https://github.com/owncloud/password_policy/pull/241 for the detailed test scenario changes needed.

1) When adding a user via the Provisioning API, if there is a `PolicyException` (e.g. the password does not meet some password policy requirement) then the OCS status message has the text `Unable to create user due to exception: ` prepended to it - e.g. 
```
Unable to create user due to exception: The password contains too few lowercase letters. At least 3 lowercase letters are required.
```
Previously the OCS status text was just:
```
The password contains too few lowercase letters. At least 3 lowercase letters are required.
```

Note: when changing a user's password via the Provisioning API the messages ahve not changed - the extra bit about the exception does not appear.

2) When adding a user using the `occ` command, then the text messages are unchanged. But they are coming on `stdout` rather than `stderr`.

Note: when changing a user password the error messages are still coming on `stderr`

What to do about each of these behaviour changes, and the inconsistencies between "add user" and "change password"?

phil-davis:
@sharidas @micbar @DeepDiver1975 ?

PVince81:
Sounds like a regression.

@sharidas please make the status text look like it was before

micbar:
@sharidas please add the previous exception https://github.com/owncloud/core/pull/35972/files#diff-a1f4d1e489770bca5e42e3d6572ba89bR128

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are closing it as outdated. If this is still relevant on a current ownCloud release, please **reopen** (or open a fresh issue) with up-to-date details — thank you!

PR #36145, opened 2019-09-03
[WIP] Move user creation to a separate service

sharidas:
Move user creation to a separate service.

## Description

Move the user creation to the service class.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Add the create user to a service class. It is moved from user_management apps UsersController. This will help us in integrating the change with the user:add command support creating users with username and email address. It will also help for provisioning API.

## How Has This Been Tested?

- This PR is same as https://github.com/owncloud/core/pull/35972. Meaning all the tests are applicable to https://github.com/owncloud/core/pull/35972 are applicable here too.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:


sharidas:
This PR is marked as deffered. The changes are same as https://github.com/owncloud/core/pull/35972.

karakayasemi:
With `Do not change the old behavior of user creation` commit I tried to resolve the problems in that issue https://github.com/owncloud/enterprise/issues/2831 .

- Additional string at the exception response of provisioning API removed.
- Old text messages of `occ user:add` command now coming from `stdErr` again. 

@phil-davis is there anything that I miss related to https://github.com/owncloud/core/issues/36019 and https://github.com/owncloud/core/issues/36020?

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,202 @@
...
+				} else {
+					$this->logger->info('Added userid ' . $user->getUID() . ' to group ' . $groupName, ['app' => 'ocs_api']);
+				}
+			}
+		}
+		return $failedToAdd;
+	}
+
+	/**
+	 * Check if the user exist
```
phil-davis:
```suggestion
	 * Check if the user exists
```

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,202 @@
...
+	 *
+	 * This function is for convinience. It helps to use this service to check if user exist,
```
phil-davis:
```suggestion
	 * This function is for convenience. It helps to use this service to check if user exist,
```

phil-davis (commented):
This looks good. There are acceptance tests in cliProvisioning suite for the `occ` `user` and `group` commands. They check the success and the command output text... Those are still passing, so the `occ` command interface has not been affected.

phil-davis:
1) Is this just the first stage refactoring?
Or 
2) does it also fix any issues with the `occ` `user` commands, e.g. creating users with just an email address?

If (2) then we should add some acceptance tests to cover the new/improved behavior.

karakayasemi:
> 
>     1. Is this just the first stage refactoring?
>        Or
> 
>     2. does it also fix any issues with the `occ` `user` commands, e.g. creating users with just an email address?
> 
@phil-davis I just worked on #36019 and #36020. In addition, rebased the PR and made it compatible with the new PHPUnit version. @sharidas has more knowledge than me about the rest of the things.

sharidas:
@phil-davis The change which @karakayasemi had made is with respect to the wording issues that was caused because of the commit https://github.com/owncloud/core/pull/36145/commits/128b83a5e38aaccc760666a3871c8639db2d1cbc. Basically now with this change, the UI and cli would be aligned with user creation.

phil-davis:
unit tests fail with
```
[PHP Fatal error:  Declaration of Tests\Core\Controller\UserControllerTest::setUp() must be compatible with Test\TestCase::setUp(): void in /drone/src/tests/Core/Controller/UserControllerTest.php on line 477]
```
needs some more refactoring...

core/Controller/UserController.php
```diff
@@ -76,4 +117,166 @@ public function getDisplayNames($users) {
...
+		}
+
+		try {
+			$this->userSendMailService->sendNotificationMail($user);
+		} catch (EmailSendFailedException $e) {
+			$this->logger->logException($e, ['app' => 'user_management']);
+			return new JSONResponse(
+				[
+					'status' => 'error',
+					'message' => $this->l10n->t('Failed to send email. Please contact your administrator.'),
```
jvillafanez: Taking into account the password is set and we aren't reverting it here, I'm not sure if the message is a bit harsh. It can be interpreted as the whole operation to have failed, including setting the password. I'm not sure what are the consequences of leaving a password set.

core/js/setpassword.js
```diff
@@ -0,0 +1,83 @@
...
+					OC.User.SetPassword._resetDone(result);
+				}).fail(function (result) {
+					OC.User.SetPassword._onSetPasswordFail(result);
+				});
+			} else {
+				//Password mismatch happened
+				passwordObj.val('');
+				retypePasswordObj.val('');
+				passwordObj.parent().addClass('shake');
+				$('#message').addClass('warning');
```
jvillafanez:
```
$('#message').addClass('warning')
  .text('Passwords do not match')
  .show()
```
might be slightly better because we don't recreate / find the `$('#message')` object each time.

core/js/setpassword.js
```diff
@@ -0,0 +1,83 @@
...
+				$('#message').text('Passwords do not match');
+				$('#message').show();
+				passwordObj.focus();
+			}
+		},
+
+		_onSetPasswordFail: function(result) {
+			var responseObj = JSON.parse(result.responseText);
```
jvillafanez: we might need to change the variable name. The `*Obj` variables above are jquery objects, so we should have a consistent naming to minimize errors.

core/js/setpassword.js
```diff
@@ -0,0 +1,83 @@
...
+			}
+
+			errorObject.text(errorMessage);
+			errorObject.show();
+			$('#submit').prop('disabled', true);
+		},
+
+		_resetDone : function(result){
+			if (result && result.status === 'success') {
+				var getRootPath = OC.getRootPath();
```
jvillafanez: we should change the variable name. `getRootPath` can easily be confused with a method name instead of a variable. Note that functions can also return other functions.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,202 @@
...
+	 *
+	 * @param array $arguments
+	 * @return IUser
+	 * @throws CannotCreateUserException
+	 * @throws InvalidEmailException
+	 * @throws UserAlreadyExistsException
+	 */
+	public function createUser($arguments) {
+		$password = $email = '';
+		if (\array_key_exists('username', $arguments)) {
```
jvillafanez: I'm not sure if we're accepting "null" as a valid value, if not, `isset` is faster.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,202 @@
...
+		if (\array_key_exists('email', $arguments)) {
+			$email = $arguments['email'];
+		}
+
+		if ($email !== '' && !$this->userSendMailService->validateEmailAddress($email)) {
+			throw new InvalidEmailException("Invalid mail address");
+		}
+
+		if ($this->userManager->userExists($username)) {
+			throw new UserAlreadyExistsException('A user with that name already exists.');
```
jvillafanez: It might be better to move this check up to prioritize this exception over the invalid email.

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,202 @@
...
+	 *
+	 * @param IUser $user
+	 * @param array $groups list of group names, example ['group1', 'group2']
+	 * @param bool $checkInGroup
+	 * @return array Returns an array of groups which failed to add user
+	 */
+	public function addUserToGroups(IUser $user, array $groups, $checkInGroup = true) {
+		$failedToAdd = [];
+
+		if (\is_array($groups) && \count($groups) > 0) {
```
jvillafanez: `is_array` shouldn't be needed.

lib/private/User/Service/UserSendMailService.php
```diff
@@ -0,0 +1,190 @@
...
+	 */
+	public function checkPasswordSetToken($token, IUser $user) {
+		$splittedToken = \explode(':', $this->config->getUserValue($user->getUID(), 'owncloud', 'lostpassword', null));
+		if (\count($splittedToken) !== 2) {
+			$this->config->deleteUserValue($user->getUID(), 'owncloud', 'lostpassword');
+			throw new InvalidUserTokenException('The token provided is invalid.');
+		}
+
+		//The value 43200 = 60*60*12 = 1/2 day
+		if ($splittedToken[0] < ($this->timeFactory->getTime() - (int)$this->config->getAppValue('core', 'token_expire_time', '43200')) ||
```
jvillafanez: maybe we can split this long line somehow.

lib/private/User/Service/UserSendMailService.php
```diff
@@ -0,0 +1,190 @@
...
+	 * @param IUser $user
+	 * @return bool returns true when an email is sent else false
+	 * @throws EmailSendFailedException
+	 */
+	public function sendNotificationMail(IUser $user) {
+		$email = $user->getEMailAddress();
+		$fromMailAddress = Util::getDefaultEmailAddress('no-reply');
+
+		if ($email !== '') {
+			try {
```
jvillafanez: I'd prefer to keep the try block as small as possible, specially if it's only the last sentence the one that could throw the exception we want to catch.

core/Controller/UserController.php
```diff
@@ -76,4 +117,166 @@ public function getDisplayNames($users) {
...
+	 * @NoSubadminRequired
+	 *
+	 * @param string $userId
+	 * @return TemplateResponse
+	 */
+	public function resendToken($userId) {
+		$user = $this->userManager->get($userId);
+
+		if ($user === null) {
+			$this->logger->error('User: ' . $userId . ' does not exist', ['app' => 'core']);
```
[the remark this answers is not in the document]
karakayasemi: Changed as suggested.

core/js/setpassword.js
karakayasemi: Changed as suggested in whole file.

core/js/setpassword.js
karakayasemi: the variable name changed as `response`.

core/js/setpassword.js
karakayasemi: the variable name changed as `rootPath`.

lib/private/User/Service/CreateUserService.php
karakayasemi: changed as suggested.

lib/private/User/Service/CreateUserService.php
karakayasemi: this check moved to next line of the username assignment.

lib/private/User/Service/CreateUserService.php
karakayasemi: `is_array` removed.

lib/private/User/Service/UserSendMailService.php
karakayasemi: According to method signatures, the last line was the only exception thrower. The unnecessary part moved to outside of the try-catch block.

core/Controller/UserController.php
```diff
@@ -76,4 +117,166 @@ public function getDisplayNames($users) {
...
+					[
+						'status' => 'error',
+						'message' => $this->l10n->t('Failed to set password. Please contact your administrator.', [$userId]),
+						'type' => 'passwordsetfailed'
+					], Http::STATUS_FORBIDDEN
+				);
+			}
+
+			\OC_Hook::emit('\OC\Core\LostPassword\Controller\LostController', 'post_passwordReset', ['uid' => $userId, 'password' => $password]);
+			@\OC_User::unsetMagicInCookie();
```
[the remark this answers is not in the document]
karakayasemi: I investigated the usage of the same method. Probably this line is taken from `LostController`. I have no other clue about error suppression.
This method originally belongs to `\OC\User\Session` and other usages of this method have no error suppression. I injected `\OC\User\Session` and removed the static method call. Also, I removed error suppression in both `LostController` and `UserController`.

core/Controller/UserController.php
```diff
@@ -76,4 +117,166 @@ public function getDisplayNames($users) {
...
+				return new JSONResponse(
+					[
+						'status' => 'error',
+						'message' => $this->l10n->t('Failed to set password. Please contact your administrator.', [$userId]),
+						'type' => 'passwordsetfailed'
+					], Http::STATUS_FORBIDDEN
+				);
+			}
+
+			\OC_Hook::emit('\OC\Core\LostPassword\Controller\LostController', 'post_passwordReset', ['uid' => $userId, 'password' => $password]);
```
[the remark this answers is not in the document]
karakayasemi: I think, this line is also taken from `LostController`. Actually, hooks are quite outdated, I would change all the usages of this hook with `EventDispatcher` pattern. However, I still don't think we can change this hook due to the possible usage of it.
I searched the listener for this hook in ownCloud GitHub organization-wide, could not find any. However, there may be a community app using this hook.

core/Command/User/Add.php
```diff
@@ -92,34 +92,40 @@ protected function configure() {
...
-			$email = null;
+		if (!$email) {
+			$email = '';
 		}
 
-		if ($input->getOption('password-from-env')) {
+		$password = '';
+		if ($passwordFromEnv) {
 			$password = \getenv('OC_PASS');
 			if (!$password) {
```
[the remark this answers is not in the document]
karakayasemi: I guess the current way is better. `\getenv` has multiple return types like false when an error. Only checking with `$password === '0'` may miss something.

apps/provisioning_api/lib/Users.php
```diff
@@ -140,7 +150,7 @@ public function addUser() {
 			return new Result(null, 102, 'User already exists');
 		}
 
-		if (\is_array($groups)) {
+		if (\is_array($groups) && (\count($groups) > 0)) {
```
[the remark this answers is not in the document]
karakayasemi: `is_array` seems valid to me and I could not find any meaningful early way to ensure `$groups` is an array.

lib/private/User/Service/UserSendMailService.php
karakayasemi: the line shortened.

core/Command/User/Add.php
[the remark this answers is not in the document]
jvillafanez: what I meant is if we have covered the case where the password is "0".

core/Controller/UserController.php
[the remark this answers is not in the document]
jvillafanez: Include a comment explaining why that method is there. If there is no comment I'd be very tempted to either remove or change the method.

The alternative could be to move some part of the code (including the event) to the lostController, and include a call here, but I don't think it's a good idea to have dependencies between controllers.

I'm opened to other solutions.

apps/provisioning_api/lib/Users.php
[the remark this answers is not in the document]
jvillafanez: I'd rather ensure that we have a list before this point. Up to you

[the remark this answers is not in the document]
karakayasemi: I moved this condition block to just after the admin authorization check. I could not quite understand what you desire, but I hope it is now better. Please, resolve conservation if it satisfies your criteria.

core/Command/User/Add.php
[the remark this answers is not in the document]
karakayasemi: Now, I covered all the possible types of `\getenv` result including $password === '0'` case.

core/Command/User/Add.php
```diff
@@ -92,34 +92,40 @@ protected function configure() {
...
 	}
 
+	/**
+	 * @param InputInterface $input
+	 * @param OutputInterface $output
+	 * @return int
+	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
+		$stdErr = $output;
+		if ($output instanceof ConsoleOutputInterface) {
```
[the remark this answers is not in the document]
karakayasemi: As far as I understand, if command ends with an exception, the exception message is written to `stdErr`. Otherwise, all messages are written to `stdOut`. `OutputInterface` has not any method to write `stdErr`.

Actually, I made the related changes to bring up old behavior. Another solution specific to our case can be, not catching the exceptions and, adjusting the exception messages to the old one.

I could not imagine a generic way to solve the output channel problem.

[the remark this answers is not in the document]
jvillafanez: > Another solution specific to our case can be, not catching the exceptions and, adjusting the exception messages to the old one.

It sounds a better choice to me, at least for now. I think it's more aligned with what we have now in other commands.

I don't think we need to adjust the exceptions messages. This should be informative to the user, but it the output isn't intended to be scriptable in any way, so changing the message should be irrelevant. We might need to adjust the error code though.

core/Controller/UserController.php
```diff
@@ -76,4 +119,166 @@ public function getDisplayNames($users) {
...
+					[
+						'link' => $this->urlGenerator->linkToRouteAbsolute('core.user.resendToken', ['userId' => $userId])
+					], 'guest'
+				);
+			}
+			$this->logger->logException($e, ['app' => 'core']);
+			return new TemplateResponse(
+				'core', 'error',
+				[
+					"errors" => [["error" => $this->l10n->t($e->getMessage())]]
```
karakayasemi: The exception should handle translations. Also, I am not sure, how Transifex script will determine the translation source string in this way.

core/Controller/UserController.php
```diff
@@ -76,4 +119,166 @@ public function getDisplayNames($users) {
...
+				'guest'
+			);
+		}
+
+		if ($user->getEMailAddress() === null) {
+			$this->logger->error('Email address not set for: ' . $userId, ['app' => 'core']);
+			return new TemplateResponse(
+				'core', 'error',
+				[
+					"errors" => [["error" => $this->l10n->t('Failed to create activation link. Please contact your administrator.', [$userId])]]
```
karakayasemi: As far as I know, the second parameter of the `l10n->t` is necessary when there is a format specifier (`%s`, `%d`) in translation string, in this way it is redundant. We need to add the format specifier of `$userID` to the string.

core/Controller/UserController.php
```diff
@@ -76,4 +119,166 @@ public function getDisplayNames($users) {
...
+	 */
+	public function setPassword($token, $userId, $password) {
+		$user = $this->userManager->get($userId);
+
+		if ($user === null) {
+			$this->logger->error('User: ' . $userId . ' does not exist.', ['app' => 'core']);
+			return new JSONResponse(
+				[
+					'status' => 'error',
+					'message' => $this->l10n->t('Failed to set password. Please contact the administrator.', [$userId]),
```
karakayasemi: Same format specifier problem.

lib/private/User/Service/UserSendMailService.php
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	/**
+	 * @param string $userId
+	 * @param string $email
+	 * @throws \OCP\PreConditionNotMetException
+	 */
+	public function generateTokenAndSendMail($userId, $email) {
+		$fromMailAddress = Util::getDefaultEmailAddress('no-reply');
+		$token = $this->secureRandom->generate(21,
```
karakayasemi: `ISecureRandom->generate` method has only 2 parameters, but 4 parameters given in here. I guess the last 3 parameters should be one concenated string.

core/Controller/UserController.php
karakayasemi: I changed error message as "Password set successfully but failed to send email. Please contact your administrator for details" in here.

core/Controller/UserController.php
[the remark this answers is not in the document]
karakayasemi: I checked the documentation, we do not mention from this hook any of the developer manual version. Even in the oldest version that I could found (oc8). Instead of this one, we are suggesting to use `\OC\User::postSetPassword`. Also, as I said, there is not a single usage of it in GitHub organization-wide.
So, to not maintain anymore this redundant, inconsistent, unwanted, static function call, I removed this hook from both Lost and User controllers. Since we do not mention anywhere, I do not think it is even necessary, but we may mention this change in the release notes.

core/Controller/UserController.php
karakayasemi: Now, the message will be translated at exception instance creation.

core/Controller/UserController.php
karakayasemi: The string changed as 'Failed to set password for %s. Please contact your administrator.'

core/Controller/UserController.php
karakayasemi: The string changed as 'Failed to set password for %s. Please contact your administrator.'

lib/private/User/Service/UserSendMailService.php
karakayasemi: Now, the second parameter is the concatenation of the previous last 3 parameters.

core/Command/User/Add.php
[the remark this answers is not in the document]
karakayasemi: We have no longer splitting output in here. I changed the command in a way that not catching the exceptions. In this way, exception messages will be written to stdErr.

karakayasemi:
@jvillafanez I made some changes for aligning the code to your previous review. Also, I found some problems with translations and resolved them.
@sharidas @jvillafanez please help me by reviewing the PR one more time. Thanks.

lib/private/User/Service/UserSendMailService.php
```diff
@@ -0,0 +1,190 @@
...
+		$this->config->setUserValue($userId, 'owncloud',
+			'lostpassword', $this->timeFactory->getTime() . ':' . $token);
+
+		// data for the mail template
+		$mailData = [
+			'username' => $userId,
+			'url' => $this->urlGenerator->linkToRouteAbsolute('core.user.setPasswordForm', ['userId' => $userId, 'token' => $token])
+		];
+
+		$mail = new TemplateResponse('core', 'new_user/email-html', $mailData, 'blank');
```
jvillafanez: I think it's more common to use `Template` than `TemplateResponse`. Any reason to use `TemplateResponse`?

lib/private/User/Service/UserSendMailService.php
```diff
@@ -0,0 +1,190 @@
...
+	/**
+	 * Send notification email to user
+	 *
+	 * @param IUser $user
+	 * @return bool returns true when an email is sent else false
+	 * @throws EmailSendFailedException
+	 */
+	public function sendNotificationMail(IUser $user) {
+		$email = $user->getEMailAddress();
+		$fromMailAddress = Util::getDefaultEmailAddress('no-reply');
```
jvillafanez: we can move it inside the condition if the variable won't be used outside of it.

lib/private/User/Service/UserSendMailService.php
```diff
@@ -0,0 +1,190 @@
...
+			$message->setHtmlBody($msg);
+			$message->setPlainBody($msgAlt);
+			$message->setFrom([$fromMailAddress => $this->defaults->getName()]);
+			try {
+				$this->mailer->send($message);
+				return true;
+			} catch (\Exception $exception) {
+				throw new EmailSendFailedException($this->l10n->t("Email could not be sent."));
+			}
+		}
```
jvillafanez: any action to take if no email is set? should we log something?
Maybe we could add a comment if we don't want to take any action to remark doing nothing is intentional.

core/Controller/LostController.php
```diff
@@ -252,8 +250,7 @@ public function setPassword($token, $userId, $password, $proceed) {
 				throw new \Exception();
 			}
 
-			\OC_Hook::emit('\OC\Core\LostPassword\Controller\LostController', 'post_passwordReset', ['uid' => $userId, 'password' => $password]);
```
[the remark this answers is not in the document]
karakayasemi: Yes, I think this way is better, explained here: https://github.com/owncloud/core/pull/36145#discussion_r359540844 . I would like to learn your opinion also.

core/Controller/LostController.php
```diff
@@ -252,8 +250,7 @@ public function setPassword($token, $userId, $password, $proceed) {
...
-			@\OC_User::unsetMagicInCookie();
+			$this->userSession->unsetMagicInCookie();
```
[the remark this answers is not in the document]
karakayasemi: Explained in more detail in here: https://github.com/owncloud/core/pull/36145#discussion_r357330149

No error log appears in manual tests with the described scenarios in there: https://github.com/owncloud/core/pull/35972#issue-303645242

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,203 @@
...
+	 * Add user to the groups
+	 *
+	 * The groups argument could not be empty and it is a list of group names. For example:
+	 *  $groups = ['group1', 'group2']
+	 *
+	 * This function returns the list of group(s) which failed to add user to the group(s)
+	 *
+	 * @param IUser $user
+	 * @param array $groups list of group names, example ['group1', 'group2']
+	 * @param bool $checkInGroup
```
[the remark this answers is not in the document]
karakayasemi: I do not even know, why this parameter is necessary. There is no usage of it in the code. Based on the YAGNI principle, I would like to remove it.

lib/private/User/Service/UserSendMailService.php
karakayasemi: done

lib/private/User/Service/CreateUserService.php
```diff
@@ -0,0 +1,203 @@
...
+	 * @return array Returns an array of groups which failed to add user
+	 */
+	public function addUserToGroups(IUser $user, array $groups, $checkInGroup = true) {
+		$failedToAdd = [];
+
+		if (\count($groups) > 0) {
+			foreach ($groups as $groupName) {
+				$groupObject = $this->groupManager->get($groupName);
+
+				if (empty($groupObject)) {
```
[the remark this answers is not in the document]
karakayasemi: done

lib/private/User/Service/CreateUserService.php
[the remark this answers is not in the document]
sharidas: Yes I think you can remove it.

lib/private/User/Service/UserSendMailService.php
sharidas: One of the reason I recall for the usage of `TemplateResponse` is to pass the value in `$mailData`.

lib/private/User/Service/UserSendMailService.php
sharidas: Added a comment.

lib/private/User/Service/CreateUserService.php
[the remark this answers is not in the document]
karakayasemi: I removed this parameter. Now, we are adding the user to the desired group, afterward checking the user is in the group. If not, adding the failing group to the return array ($failedToAdd). 

If the user is already in the group, we are still assuming the add operation successful.

DeepDiver1975:
old aka dead -> close

PR #36159, opened 2019-09-04
shareowner can edit re-share permissions without any restriction

### LINKED ISSUE #36107: [10.3.0 alpha] Owner cannot edit group's permissions

karakayasemi:
## Description
Shareowner should have full permission on re-shares.

To exclude ShareOwner from the re-share permission check, I need to know the session user is shareowner or not in `ShareManager` class. In #36144, I injected `IUserSession` to `ShareManager`, but it led to several problems. 

According to my observations, `IRootFolder` `getOwner` method returns the user who made the request. If no user exists in the session it returns `false`. So, I used `IRootFolder`'s `getOwner` method to determine the user who made the request.  @jvillafanez, @VicDeo Do you see any side effect of this approach?

## Related Issue
- Fixes #36107 

## Motivation and Context
Fighting with bugs.

## How Has This Been Tested?
Manually with steps in #36107. 

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -321,8 +321,14 @@ class Manager implements IManager {
 			$maxPermissions |= \OCP\Constants::PERMISSION_DELETE | \OCP\Constants::PERMISSION_UPDATE;
 		}
 
-		/** If it is re-share, calculate $maxPermissions based on all incoming share permissions */
-		if ($shareNode->getOwner()->getUID() !== $share->getSharedBy()) {
+		$nodeOwner = $shareNode->getOwner()->getUID();
+		/**
+		 * If it is re-share, calculate $maxPermissions based on all incoming share permissions
+		 * TODO: $this->rootFolder->getOwner() was used for determining Session User in here.
+		 *       To fix this confusion, properly inject UserSession.
+		 *       See: https://github.com/owncloud/core/issues/36107#issuecomment-529977195
+		 */
+		if ($nodeOwner !== $share->getSharedBy() && $nodeOwner !== $this->rootFolder->getOwner()->getUID()) {
 			$maxPermissions = $this->calculateReshareNodePermissions($share);
 		}
 
@@ -1280,13 +1286,13 @@ class Manager implements IManager {
 	 */
 	public function getAllSharedWith($userId, $shareTypes, $node = null) {
 		$shares = [];
-		
+
 		// Aggregate all required $shareTypes by mapping provider to supported shareTypes
 		$providerIdMap = $this->shareTypeToProviderMap($shareTypes);
 		foreach ($providerIdMap as $providerId => $shareTypeArray) {
 			// Get provider from cache
 			$provider = $this->factory->getProvider($providerId);
-			
+
 			// Obtain all shares for all the supported provider types
 			$queriedShares = $provider->getAllSharedWith($userId, $node);
 			$shares = \array_merge($shares, $queriedShares);
@@ -1294,7 +1300,7 @@ class Manager implements IManager {
 
 		return $shares;
 	}
-	
+
 	/**
 	 * @inheritdoc
 	 */
--- a/tests/acceptance/features/apiShareReshare/reShare.feature
+++ b/tests/acceptance/features/apiShareReshare/reShare.feature
@@ -224,6 +224,7 @@ Feature: sharing
       | permissions | share,read |
     Then the OCS status code should be "<ocs_status_code>"
     And the HTTP status code should be "200"
+    And user "user2" should not be able to upload file "filesForUpload/textfile.txt" to "/TMP/textfile.txt"
     Examples:
       | ocs_api_version | ocs_status_code |
       | 1               | 100             |
@@ -239,6 +240,7 @@ Feature: sharing
       | permissions | share,create,update,read |
     Then the OCS status code should be "<ocs_status_code>"
     And the HTTP status code should be "200"
+    And user "user2" should be able to upload file "filesForUpload/textfile.txt" to "/TMP/textfile.txt"
     Examples:
       | ocs_api_version | ocs_status_code |
       | 1               | 100             |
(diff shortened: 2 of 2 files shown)


jvillafanez:
What if that additional restriction is placed up in the stack? I mean, core maybe doesn't need to check who is calling that function as long as the share makes sense; the one that should check is either the controller or the app making that call.
Probably not a good idea because we can't control who is going to make the call (3rdparty apps can take advantage of this to bypass some restrictions).

In any case, I think the problem is what to do if the session doesn't have a user at that point. If this is required, maybe we should abort the request.

I haven't checked where the `rootFolder->getOwner()` call leads to, but I guess it will still rely on session information, otherwise I don't know how it can return the correct user.

PVince81:
Please investigate whether `Root->getOwner()` is really based on the session or whether it's the storage owner. As far as I remember it should rather be the storage owner as it's the owner of the node, not the owner of the share.

In terms of storage owner, there are scenarios where there is no owner, for example for a system-wide external storage. In the latter case, I remember that there is fallback where it would indeed use the session user. But only as a fallback.

micbar:
@PVince81 @jvillafanez 
I investigated the $rootFolder->getOwner code path.

You are correct, it can be both.
We can get `false` as a response.

micbar:
Further down Common Storage uses `\OC_User::getUser();` as a fallback which does
```
/**
	 * get the user id of the user currently logged in.
	 *
	 * @return string|bool uid or false
	 */
	public static function getUser() {
		$uid = \OC::$server->getSession() ? \OC::$server->getSession()->get('user_id') : null;
		if ($uid !== null && self::$incognitoMode === false) {
			return $uid;
		} else {
			return false;
		}
	}
```

karakayasemi:
> 
> What if that additional restriction is placed up in the stack? I mean, core maybe doesn't need to check who is calling that function as long as the share makes sense; the one that should check is either the controller or the app making that call.

If the original share owner makes a request, we will calculate `$maxPermission` with his own permissions instead of sharer user's permissions. Actually, what we are trying to do is make an exclusion for the original share owner and let him edit reshare without the sharer user's permission limits. 

>
> In any case, I think the problem is what to do if the session doesn't have a user at that point. If this is required, maybe we should abort the request.

If there is no user in session, permissions will be calculated with re-sharer user's permission. So, permission control is even more strict in this case. `occ files:transfer-ownership` command is using `updateShare` function to transfer shared files ownership and there is no session user in this case. When I tried this scenario, `$this->rootFolder->getOwner()->getUID()` returned false and since `false` not equal to Sharer User's UID, `$maxPermission` calculated  with `calculateReshareNodePermissions` method.

jvillafanez:
Just giving some thoughts:
* if there is a user in the session:
    * if it matches the share owner, use the share owner to calculate the permissions
    * if it matches the resharer, use the resharer to calculate the permissions
* if there is no user in the session: abort.

The problem is that core is tying the user in the user session with the user that will perform the action. Maybe the solution is core to be more flexible in this regard, technically allowing any user to perform any operation, and letting other apps to handle proper access or providing specific services to access properly to those resources.

A simple service such as (incompleted dummy code)
```
function updateShare($user, IShare $share) {
  if ($user === $share->getOwner() || $user === $share->getResharer()) {
    updateShare($share);
  }
}
```
should fix the problem as long as the actual update doesn't depend on the session information. The question is if we can reach that point from the current code.

Alternatively, we can try to inject the target user in the session.

micbar (changes requested):
@karakayasemi Please implement the code changes, that we discussed.

karakayasemi:
As we discussed, I implemented passing user information via an optional parameter approach here: https://github.com/owncloud/core/pull/36166.  Also, I did not want to quickly eliminate this PR, and I've investigated possible scenarios of it.

I tracked `IRootFolder` in the DIContainer, as a result, I reached it is a `Root` object: 
https://github.com/owncloud/core/blob/ae9ee7ab15070b1f9a4d5ea6511f63e7f452a908/lib/private/Server.php#L239

When we call `getOwner` function from `IRootFolder`, it equals to this call (`Node` is a parent object of `Root`):
https://github.com/owncloud/core/blob/ae9ee7ab15070b1f9a4d5ea6511f63e7f452a908/lib/private/Files/Node/Node.php#L347

And then, I reached that `View`'s `getFileInfo` method is determining the owner: 
https://github.com/owncloud/core/blob/ae9ee7ab15070b1f9a4d5ea6511f63e7f452a908/lib/private/Files/Node/Node.php#L93

Finally, this method is setting owner user in `View`: 
https://github.com/owncloud/core/blob/ae9ee7ab15070b1f9a4d5ea6511f63e7f452a908/lib/private/Files/View.php#L1352
If storage owner founds, this method returns the storage owner, else, it returns `RemoteUser`. This means we will never call `getUID` method from a null. I had concerns that `getOwner` might return null, looks like it is not possible.

When `$this->rootFolder->getOwner()->getUID()` returns something, even it returns false or any other non-owner user, this won't pass being equal to ShareOwner condition. So, if it is a re-share, `ShareManager` will use the re-sharer permissions to calculate the permissions as it's supposed to be. What I am trying to say is, I could not find any problematic situation for this PR too. I would like to keep open this PR until #36166 merged.

PR #36161, opened 2019-09-05
obey to config in share mail notifications APIs

### LINKED ISSUE #36156: api for public link share email notification does not obey `Allow users to send mail notification for shared files`
### Steps to reproduce
1. disable `Allow users to send mail notification for shared files`
2. send an email notification by using `ocs/v1.php/apps/files_sharing/api/v1/notification/notify-public-link-by-email` (CSRF needs to be disabled when using basic auth - that is a different issue)

### Expected behaviour
Email should not be send

### Actual behaviour
Email is send

CC @karakayasemi

karakayasemi:
## Description
We have two different admin options (public link and internal mail) for enabling/disabling share mail notifications, but APIs don't obey to this config. This pr fixes this behavior.

## Related Issue
- Fixes #36156 

## Motivation and Context
Fighting with bugs.

## How Has This Been Tested?
Manually with Curl and unit test.
## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/NotificationController.php
+++ b/apps/files_sharing/lib/Controller/NotificationController.php
@@ -32,6 +32,7 @@ use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\Share;
+use OCP\Share\Exceptions\GenericShareException;
 
 class NotificationController extends OCSController {
 	/** @var MailNotifications */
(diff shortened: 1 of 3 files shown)


apps/files_sharing/tests/Controller/NotificationControllerTest.php
```diff
@@ -60,6 +64,10 @@ protected function setUp() {
 		$this->groupManager = $this->createMock(IGroupManager::class);
 		$this->rootFolder = $this->createMock(IRootFolder::class);
 		$this->l = $this->createMock(IL10N::class);
+		$this->config = $this->createMock(IConfig::class);
+		$this->config
```
jvillafanez: better setup the mock in each tests

karakayasemi:
@jvillafanez I applied suggested changes. Please review one more time. Thanks.

jvillafanez (approved):

jvillafanez:
I kind of have to complain about files_sharing using configuration for core. If it's a setting for core it should be core the one managing that setting, files_sharing shouldn't care about it.
A better solution would probably be to use some kind of service managed by core to send those email, so the restrictions are place there, but this will require a lot of time.

We can go with this solution, but I think we should include a fixme.

karakayasemi:
@jvillafanez instead of adding a fixme, I moved the controls to `core` as you suggested. Please review one more time. Thank you.

PR #36168, opened 2019-09-06
Allow share_folder to be specified without leading slash

### LINKED ISSUE #36167: Protection against deletion of share_folder is not effective if it does not have a leading slash
PR #35998 added code to prevent a user from deleting the folder define by `share_folder` in the config.

### Steps to reproduce
1. in `config.php`  set `share_folder => 'SharedFolder'`
2. share a folder `stuff` from user1  to user2
3. login as user2
4. you see a folder called `SharedFolder`  and that has `stuff` inside it - good
5. delete `SharedFolder`

### Expected behaviour
There should be some error message and `SharedFolder` is not deleted.

### Actual behaviour
`SharedFolder` gets deleted.

### Server configuration
Current core master

Note: if you set `share_folder => '/SharedFolder'` (with leading slash) then the system behaves as expected.

But other behaviour of `share_folder` config setting works fine without the leading slash. So there could be installations "out there" that have it defined without the leading slash.

phil-davis:
## Description
`normalizePath` of `share_folder` value. That ensures that it has a leading slash (and other good stuff). 

Note: other uses of `share_folder` already pass it through `normalizePath` - that is why they work when `share_folder` does not have a leading slash.

## Related Issue
- Fixes #36167 

## How Has This Been Tested?
Unit tests and manual verification by setting `share_folder` to just `SharedFolder` and then receiving a share and trying to delete that folder.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/lib/Files/ViewTest.php
+++ b/tests/lib/Files/ViewTest.php
@@ -2681,7 +2681,10 @@ class ViewTest extends TestCase {
 	public function providesShareFolder() {
 		return [
 			['/MyTestShareFolder', '/MyTestShareFolder'],
+			['MyTestShareFolder', '/MyTestShareFolder'],
 			['/MyTestShareFolder/Share/Foo', '/MyTestShareFolder'],
+			['MyTestShareFolder/Share/Foo', '/MyTestShareFolder'],
+			['/MyTestShareFolder/Share/Foo', '/MyTestShareFolder/Share'],
 		];
 	}
 
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -352,10 +352,14 @@ class View {
 	 */
 	public function rmdir($path) {
 		return $this->emittingCall(function () use (&$path) {
-			if (($path !== '') &&
-				(\strpos($this->config->getSystemValue('share_folder', '/'), $path) !== false)) {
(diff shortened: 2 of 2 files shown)


tests/lib/Files/ViewTest.php
```diff
@@ -2681,7 +2681,10 @@ public function testFopenFail() {
 	public function providesShareFolder() {
 		return [
 			['/MyTestShareFolder', '/MyTestShareFolder'],
+			['MyTestShareFolder', '/MyTestShareFolder'],
 			['/MyTestShareFolder/Share/Foo', '/MyTestShareFolder'],
+			['MyTestShareFolder/Share/Foo', '/MyTestShareFolder'],
+			['/MyTestShareFolder/Share/Foo', '/MyTestShareFolder/Share'],
```
phil-davis: I added some test cases without the leading slash, and also this one that checks that you cannot delete a folder in the middle of the `share_folder` path.

phil-davis:
See PR #36170 for later code, and that does not bother using `normalizePath`

phil-davis:
PR #36170 was merged. So no need for this PR.

PR #36170, opened 2019-09-06
Improve share_folder checks in rmdir

### LINKED ISSUE #36167: Protection against deletion of share_folder is not effective if it does not have a leading slash

### LINKED ISSUE #36169: Cannot delete other folders that match part of share_folder
PR #35998 add code to prevent a user from deleting the folder define by `share_folder` in the config.

### Steps to reproduce
1. in `config.php`  set `share_folder => '/Company/Files/Shared'`
2. share a folder `stuff` from user1  to user2
3. login as user2
4. you see a folder called `Company/Files/Shared` and that has `stuff` inside it - good
5. create top-level folders called `Com` `File` `Files` `Share` and `Shared`
6. create folder `Shared` inside `Files`
7. try to delete any of those folders that you just created

### Expected behaviour
Each folder can be deleted.

### Actual behaviour
There is an error message and `Error deleting...`

### Server configuration
Current core master

phil-davis:
## Description
The folder being deleted is only part of the path to `share_folder` if its path matches at the start of `share_folder` and has the complete matching folder name (i.e. when `share_folder` is `/abc/def` then it is OK to `rmdir` `/abc/de`)

Make it so with trim and putting slashes and strpos.

The code here also fixes what is in PR #36168 

## Related Issue
- Fixes #36167 
- Fixes #36169 

## How Has This Been Tested?
Local unit test runs, and trying combinations of folder deletes in the webUI.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -352,10 +352,13 @@ class View {
 	 */
 	public function rmdir($path) {
 		return $this->emittingCall(function () use (&$path) {
-			if (($path !== '') &&
-				(\strpos($this->config->getSystemValue('share_folder', '/'), $path) !== false)) {
-				Util::writeLog("core", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
-				return false;
+			if ($path !== '') {
+				$shareFolder = \trim($this->config->getSystemValue('share_folder', '/'), '/');
+				$trimmedPath = \trim($path, '/');
+				if ((\strpos("$shareFolder/", "$trimmedPath/") === 0)) {
+					Util::writeLog("core", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
+					return false;
+				}
 			}
 			$absolutePath = $this->getAbsolutePath($path);
 			$mount = Filesystem::getMountManager()->find($absolutePath);
--- a/tests/lib/Files/ViewTest.php
+++ b/tests/lib/Files/ViewTest.php
@@ -2681,7 +2681,10 @@ class ViewTest extends TestCase {
 	public function providesShareFolder() {
 		return [
 			['/MyTestShareFolder', '/MyTestShareFolder'],
+			['MyTestShareFolder', '/MyTestShareFolder'],
 			['/MyTestShareFolder/Share/Foo', '/MyTestShareFolder'],
+			['MyTestShareFolder/Share/Foo', 'MyTestShareFolder'],
+			['/MyTestShareFolder/Share/Foo', '/MyTestShareFolder/Share'],
 		];
 	}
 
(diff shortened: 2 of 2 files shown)


lib/private/Files/View.php
```diff
@@ -352,10 +352,14 @@ protected function renameUpdate(Storage $sourceStorage, Storage $targetStorage,
...
 		return $this->emittingCall(function () use (&$path) {
-			if (($path !== '') &&
-				(\strpos($this->config->getSystemValue('share_folder', '/'), $path) !== false)) {
-				Util::writeLog("core", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
-				return false;
+			if ($path !== '') {
+				$shareFolder = Filesystem::normalizePath(
+					$this->config->getSystemValue('share_folder', '/')
+				);
+				if ((\strpos("$shareFolder/", '/' . \trim($path, '/'). '/') === 0)) {
```
jvillafanez:
```
$shareFolder = \trim($this->config->getSystemValue('share_folder', '/'), '/');
$trimmedPath = \trim($path, '/');
if ($shareFolder !== '' && $shareFolder === $path) {
  ...
}
```
This seems easier to follow

phil-davis: Yes, sure. I had the "normalize" code already in the first PR.

phil-davis: And your code example also has to be enhanced to deal with parent folders.
e.g. if `share_folder` is `/abc/def/Shares` then you are not allowed to delete any of:
```
/abc
/abc/def
```
But you can delete folders like:
```
/ab
/abc/de
```

phil-davis:
@jvillafanez ready for review again

sharidas (commented):
LGTM

lib/private/Files/View.php
```diff
@@ -352,10 +352,13 @@ protected function renameUpdate(Storage $sourceStorage, Storage $targetStorage,
...
 	public function rmdir($path) {
 		return $this->emittingCall(function () use (&$path) {
-			if (($path !== '') &&
-				(\strpos($this->config->getSystemValue('share_folder', '/'), $path) !== false)) {
-				Util::writeLog("core", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
-				return false;
+			if ($path !== '') {
+				$shareFolder = \trim($this->config->getSystemValue('share_folder', '/'), '/');
+				$trimmedPath = \trim($path, '/');
+				if ((\strpos("$shareFolder/", "$trimmedPath/") === 0)) {
```
jvillafanez: it seems the condition we want is `$shareFolder === $trimmedPath`, anything I'm missing?

phil-davis: If `share_folder` is `/Company/Stuff/ReceivedShares` and path is `/Company` or `/Company/Stuff` then those parent folders of `ReceivedShares` are also not allowed to be deleted.

See the unit test data provider test cases.

jvillafanez (approved):

micbar:
## Steps to reproduce
- Set Server setting for Default shared folder in `config.php` to `/Shared`
- User 1 shares files and folders with User 2
- User 2 should see the incoming shares under the `/Shared` Folder 
- User 2 tries to delete the `/Shared` Folder

HanaGemela:
Works as expected with the desktop sync client 2.6.0 RC1, macOS 10.14.6. User cannot delete the folder. But the client doesn't restore the deleted folder https://github.com/owncloud/client/issues/7496

OS-WS:
Hi, is this the fix for [CVE-2020-36251](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-36251)?

thanks in advance!

micbar:
@C0rby @IljaN Can you provide an answer?

PR #36208, opened 2019-09-22
Do not calculate reshare permission, use supershare permissions

mrow4a:
There is no need to calculate reshare permission, these are already calculated by ShareProviders e.g. https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/MountProvider.php#L71-L105

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -330,10 +330,24 @@ class Manager implements IManager {
 			$maxPermissions |= \OCP\Constants::PERMISSION_DELETE | \OCP\Constants::PERMISSION_UPDATE;
 		}
 
-		/** If it is re-share, calculate $maxPermissions based on all incoming share permissions */
+		/** If share node is also share (reshare), get $maxPermissions based on supershare permissions */
 		if ($this->userSession !== null && $this->userSession->getUser() !== null &&
 			$share->getShareOwner() !== $this->userSession->getUser()->getUID()) {
-			$maxPermissions = $this->calculateReshareNodePermissions($share);
+			// if it is an incoming federated share, use node permission directly
+			if ($shareNode->getStorage()->instanceOfStorage('OCA\Files_Sharing\External\Storage')) {
+				return $shareNode->getPermissions();
+			}
+
+			// check if share node file is also a share
+			$userFolder = $this->rootFolder->getUserFolder($share->getSharedBy());
+			$shareFileNode = $userFolder->getById($share->getNode()->getId(), true)[0];
+			$shareFileStorage = $shareFileNode->getStorage();
+			if ($shareFileStorage->instanceOfStorage('OCA\Files_Sharing\SharedStorage')) {
+				/** @var \OCA\Files_Sharing\SharedStorage $shareFileStorage */
+				'@phan-var \OCA\Files_Sharing\SharedStorage $shareFileStorage';
+				$parentShare = $shareFileStorage->getShare();
+				$maxPermissions = $parentShare->getPermissions();
+			}
 		}
 
 		// Check that we do not share with more permissions than we have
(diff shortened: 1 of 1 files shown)


PVince81:
is this an optimization or a bug fix ?

phil-davis:
Acceptance test fails https://drone.owncloud.com/owncloud/core/20663/45/12
```
  Scenario Outline: User is not allowed to reshare folder and add delete permission bit (8)                                  # /drone/src/tests/acceptance/features/apiShareReshare/reShare.feature:191
    Given using OCS API version "<ocs_api_version>"                                                                          # FeatureContext::usingOcsApiVersion()
    And user "user2" has been created with default attributes and without skeleton files                                     # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "user0" has shared folder "/PARENT" with user "user1" with permissions <received_permissions>                   # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    When user "user1" shares folder "/PARENT" with user "user2" with permissions <reshare_permissions> using the sharing API # FeatureContext::userSharesFileWithUserUsingTheSharingApi()
    Then the OCS status code should be "404"                                                                                 # OCSContext::theOCSStatusCodeShouldBe()
    And the HTTP status code should be "<http_status_code>"                                                                  # FeatureContext::theHTTPStatusCodeShouldBe()
    And as "user2" folder "/PARENT" should not exist                                                                         # FeatureContext::asFileOrFolderShouldNotExist()
    But as "user1" folder "/PARENT" should exist                                                                             # FeatureContext::asFileOrFolderShouldExist()

    Examples:
      | ocs_api_version | http_status_code | received_permissions | reshare_permissions |
      | 1               | 200              | 17                   | 25                  |
        OCS status code is not the expected value
        --- Expected
        +++ Actual
        @@ @@
        -'404'
        +'100'
...
```
With this code, it looks like a user can do a reshare and add delete permission, when they do not have delete permission.

mrow4a:
@PVince81 it is optimisation - the max permission is already computed as part of supershare. I need to look however why test fails which is worrying (SuperShare mechanism is the one responsible for displaying the files in the UX @phil-davis )

phil-davis:
PR #36210 has an adjusted test scenario, which demonstrates that with this code a reshare can add delete permission when it should not be allowed.

karakayasemi (commented):
We can not rely on super share node permission in here, because `View` is adding delete bit to every mount, as we see in this issue https://github.com/owncloud/enterprise/issues/3404. The solution that we found for this problem is calculating re-share permission from incoming shares (Super shares). I guess that's why acceptance tests failed.

Another my concern in this PR is, a re-share may have two super shares  (For example one of them is group share, and the other one is normal-share ). I am not sure, how we deduplicate these super shares and determine the re-share node permission. However, in the current code, we are determining re-share permission by summing up all incoming shares permission. Also, 
according to the interface description of `$userFolder->getById()` and $shareManager->`getAllSharedWith()`, both returns arrays. I used two `foreach` to make sure I don't miss anything. Just relying on only the first node coming from `$userFolder->getById()` can create a problem.

mrow4a:
@karakayasemi not really, supershare is exactly one that calculates cumulative permission of all incoming shares (user and group for the same file). What you do in calculateResharePermission is calculates cumulative permission of all incoming shares which generates a supershare permission.

micbar:
@mrow4a Was there a bug or a ticket? What is this supposed to achieve?

mrow4a:
@micbar i work on strict checks for attributes in sharing. It is equivalent of strict permission checks but for attributes. Thus my concerns of how we do it for permissions. It is optimization and work in progress.

micbar:
@mrow4a Ok, sounds good.
Let us skip this in context of 10.3
We can do this in the next patch release.

mrow4a:
@micbar this is not required in 10.3 (as advanced sharing attributes enforcement is also not).

karakayasemi (approved):
CI is green. As @mrow4a said, the re-share permission already calculated as super-share permission with a similar method in https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/MountProvider.php#L175. LGTM 👍

micbar:
@mrow4a please do not merge yet.

mrow4a:
Closing this as it will be part of https://github.com/owncloud/core/pull/36265

PR #36216, opened 2019-09-24
[API] user-sync OCS API

DeepDiver1975:
## Description
External user provisioning system need an http api to trigger user-sync for a specific user.

### Authorization
The http API can only be executed by a user with admin priviledges. Suggestion is to create a technical user who is in the admin group

### Route
```sh
 curl -X POST http://localhost:8080/ocs/v2.php/cloud/user-sync/admin -v -u admin

```

### Response
- 200 - if sync was executed
- 404 - given userId is unknown
- 409 - mutliple users have been found for the given user id - not unique user criteria

## Related Issue
- refs https://github.com/owncloud/enterprise/issues/3521

## How Has This Been Tested?
- user curl command as above with an admin user

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/core/routes.php
+++ b/core/routes.php
@@ -57,6 +57,7 @@ $application->registerRoutes($this, [
 		['root' => '/cloud', 'name' => 'Cloud#getCapabilities', 'url' => '/capabilities', 'verb' => 'GET'],
 		['root' => '/cloud', 'name' => 'Cloud#getCurrentUser', 'url' => '/user', 'verb' => 'GET'],
 		['root' => '/cloud', 'name' => 'Roles#getRoles', 'url' => '/roles', 'verb' => 'GET'],
+		['root' => '/cloud', 'name' => 'UserSync#syncUser', 'url' => '/user-sync/{userId}', 'verb' => 'POST'],
 	]
 ]);
 
--- a/core/Application.php
+++ b/core/Application.php
@@ -39,6 +39,9 @@ use OC\Core\Controller\RolesController;
 use OC\Core\Controller\TokenController;
 use OC\Core\Controller\TwoFactorChallengeController;
 use OC\Core\Controller\UserController;
+use OC\Core\Controller\UserSyncController;
+use OC\User\AccountMapper;
+use OC\User\SyncService;
 use OC_Defaults;
 use OCP\AppFramework\App;
 use OCP\BackgroundJob\IJobList;
@@ -151,6 +154,19 @@ class Application extends App {
 				$c->query('Request'),
 				$c->query('L10N'),
 				$serverContainer->getEventDispatcher()
+			);
+		});
+		$container->registerService('UserSyncController', static function (SimpleContainer $c) {
+			$syncService = new SyncService(
+				$c->query(IConfig::class),
+				$c->query(ILogger::class),
+				$c->query(AccountMapper::class)
+			);
+			return new UserSyncController(
+				$c->query('AppName'),
+				$c->query('Request'),
+				$syncService,
+				$c->query('UserManager')
 			);
 		});
 		$container->registerService('CronController', static function (SimpleContainer $c) {
--- a/core/Command/User/SyncBackend.php
+++ b/core/Command/User/SyncBackend.php
@@ -260,8 +260,7 @@ class SyncBackend extends Command {
 		$dummy = new Account(); // to prevent null pointer when writing messages
 		if (\count($users) === 1) {
 			// Run the sync using the internal username if mapped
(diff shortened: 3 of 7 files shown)


PVince81:
Doesn't "occ user:sync" also prompt the admin about what to do with some users ? (like "delete" vs "disabled", ...) maybe the API also needs to accept a parameter to set the behavior in case a user is gone ?

I also assume that the sync is fast and there are no concerns about time outs ? Or if it times out, nothing bad will happen like having the database / user data in a in-between state ?

DeepDiver1975:
> Doesn't "occ user:sync" also prompt the admin about what to do with some users ? (like "delete" vs "disabled", ...) maybe the API also needs to accept a parameter to set the behavior in case a user is gone ?

Out of scope for our explicit case. For a general case: if some external user provisioning system is deprovisioning a user - they better disable or delete the user anyway...




> I also assume that the sync is fast and there are no concerns about time outs ? Or if it times out, nothing bad will happen like having the database / user data in a in-between state ?

All API calls run with no php time out - nothing more we can do from my understanding.

PVince81:
@DeepDiver1975 please add a PHP Doc comment then in the controller method that will explain what the default behavior is, whether we're disabling users or removing them. Or are we doing nothing at all ?

Looking here https://github.com/owncloud/core/blob/master/core/Command/User/SyncBackend.php#L108 I see there is also another case called "re-enable", likely also out of scope.

The reason for documenting is that API users might wonder what the difference is with "occ user:sync" they might be used to.

core/routes.php
```diff
@@ -57,6 +57,7 @@
 		['root' => '/cloud', 'name' => 'Cloud#getCapabilities', 'url' => '/capabilities', 'verb' => 'GET'],
 		['root' => '/cloud', 'name' => 'Cloud#getCurrentUser', 'url' => '/user', 'verb' => 'GET'],
 		['root' => '/cloud', 'name' => 'Roles#getRoles', 'url' => '/roles', 'verb' => 'GET'],
+		['root' => '/cloud', 'name' => 'UserSync#syncUser', 'url' => '/user-sync/{userId}', 'verb' => 'POST'],
```
PVince81: I wonder if we should put this on the provisioning API endpoint instead ?

phil-davis: ocis will not have the current Provisioning API, because it will require other "external" ways to provision users/groups. But it will need the ability to trigger a user sync. So I guess this user sync API endpoint will be needed, but not any of the existing Provisioning API.

Does that influence the question about putting this in the Provisioning API?

PVince81: not sure if OCIS will have the user:sync feature either as it will have a different way of managing users

in any case, I think the proposed route is alright

DeepDiver1975: provisioning api sounds reasonable as well .... let me think about this ...

DeepDiver1975: The UserSyncService is a private class in core - there is no public API to allow the provisioning API to access this class. :gun:

PVince81 (approved):
:+1: then

phil-davis:
Since this is a feature addition, it will trigger a minor version update to `10.4.0`

After `10.3` is released, we want to follow a new process for doing more frequent patch releases, and coordinating the timing of when feature branches are merged to `master` (that cause a minor version bump, and perhaps more QA effort than needed for patch releases that are "bugfixes only").

And we should have some API acceptance tests added.

So can we hold-off on merging this?

@HanaGemela have you discussed the proposed new process with developers?

micbar:
@phil-davis You are correct, this was merged too early.

The new process ist not implemented yet, so this has kind of fallen between the cracks...

phil-davis:
I created an issue to make acceptance tests. If we have those done and merged, then probably they will be able to automatically cover the QA needs for this. So this will cause the next release to be numbered `10.4.0` (which is no problem, it's what we do with semver), but should not cause any big QA delays.
Needed client testing would be no more than the usual smoke tests, which verify that the existence of a new and unused API endpoint has no bad effect.

dpakach:
@DeepDiver1975  Does this works for LDAP? I tried to make acceptance tests for this (issue - #36236). But I could not get this to work with LDAP. Also the occ command seems to have a problem (issue - https://github.com/owncloud/user_ldap/issues/456).

DeepDiver1975:
>  Does this works for LDAP?

this is the prime use case - and it works in prod at a customer ;-)



>  Also the occ command seems to have a problem (issue - [owncloud/user_ldap#456](https://github.com/owncloud/user_ldap/issues/456)).

I'm suspecting config/setup issues ....

PR #36221, opened 2019-09-26
cron.php calls occ system:cron as fallback

DeepDiver1975:
## Description
cron.php calls the occ system:cron command as fallback

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -40,4 +40,8 @@ if (!\OC::$CLI) {
 }
 
 echo 'Please use ./occ system:cron' . PHP_EOL;
-exit(1);
+$return = \system('./occ system:cron');
+// in case of an error while cron execution we exit with error code as well
+if ($return === false) {
+	exit(1);
+}


cron.php
```diff
@@ -40,4 +40,8 @@
 }
 
 echo 'Please use ./occ system:cron' . PHP_EOL;
-exit(1);
+$return = system('./occ system:cron');
```
PVince81: this implies that PHP is allowed to call system functions

is there maybe another way where we instantiate it through Symfony like we do for unit tests ?

DeepDiver1975: > this implies that PHP is allowed to call system functions

at this level where php scripts are called from crontab we can assume that we can call system functions.

Adding more complexity here by using symfony is an overkill from my perspective

PVince81: okay, so assuming that the php.ini for the cli is more permissive

PVince81 (approved):
:+1:

PVince81:
@micbar

cron.php
micbar: Coding standard is complaining

micbar (changes requested):
Could you push the changes?

micbar (approved):

PR #36222, opened 2019-09-26
only collect shares inside the given path in transfer-ownership command

karakayasemi:
## Description
`files:transfer-ownership` command is not respecting to the given path when collecting shares for transfer. With this PR it will only collect shares inside the given path. 

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3520

## Motivation and Context
Even `path` argument given, `files:transfer-ownership` command is trying to transfer all shares of `sourceUser`. This situation causes random errors like mentioned issue.  `files:transfer-ownership` command should only transfer the shares inside of the given folder. 

## How Has This Been Tested?
Unit tests and ✋ Manually with the following steps:  
- Create at least two user (`user1`, `user2`)
- `user1` creates a folder and fills it with random files.
- `user1` creates a file outside of the created folder and shares this file with `user2`.
- Transfer the created folder with this command;
`./occ files:transfer-ownership --path="folder-path"  user1 user2`

### Expected behavior
- The folder should be transferred without any error messages.

### Current behavior
- There are some errors in console output like `Share with id x points at deleted file, skipping`, because the command is trying to transfer all the shares.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -30,7 +30,6 @@ use OC\Files\View;
 use OC\Share20\ProviderFactory;
 use OCP\Files\FileInfo;
 use OCP\Files\Mount\IMountManager;
-use OCP\ILogger;
 use OCP\IUserManager;
 use OCP\Share\IManager;
 use OCP\Share\IShare;
@@ -54,9 +53,6 @@ class TransferOwnership extends Command {
 
 	/** @var Manager  */
 	private $encryptionManager;
-
-	/** @var ILogger  */
-	private $logger;
 
 	/** @var ProviderFactory  */
 	private $shareProviderFactory;
@@ -85,12 +81,11 @@ class TransferOwnership extends Command {
 	/** @var string */
 	private $finalTarget;
 
-	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ILogger $logger, ProviderFactory $shareProviderFactory) {
+	public function __construct(IUserManager $userManager, IManager $shareManager, IMountManager $mountManager, Manager $encryptionManager, ProviderFactory $shareProviderFactory) {
 		$this->userManager = $userManager;
 		$this->shareManager = $shareManager;
 		$this->mountManager = $mountManager;
 		$this->encryptionManager = $encryptionManager;
-		$this->logger = $logger;
 		$this->shareProviderFactory = $shareProviderFactory;
 		parent::__construct();
 	}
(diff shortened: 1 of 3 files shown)


micbar:
@karakayasemi could you increase coverage?

apps/files/lib/Command/TransferOwnership.php
```diff
@@ -257,11 +256,23 @@ private function collectUsersShares(OutputInterface $output) {
...
+				if (\strlen($this->inputPath)>1) {
+					/**
+					 * filter only shares within the source folder
+					 */
+					$sharePage = \array_filter($sharePage, function (IShare $share) {
+						/**
+						 * strpos result should be 1 for the shares inside source folder,
+						 * because `/` character trimmed with ltrim in inputPath
+						 */
+						return \strpos($share->getNode()->getPath(), $this->inputPath) === 1;
```
phil-davis: Without testing myself, I have a suspicion that if:
- the user has 2 folders `sub` and `subfolder`
- both folders contain shares
- someone tries to transfer for `sub`

This will also match `subfolder`, so shares from both will get transferred.

phil-davis:
> @karakayasemi could you increase coverage?

There is also `tests/acceptance/features/cliMain/transfer-ownership.feature` - hopefully it should be easy to make or adjust acceptance test scenarios to cover this.

apps/files/lib/Command/TransferOwnership.php
micbar: @karakayasemi please investigate.

karakayasemi: @phil-davis Most probably, you are right about your suspicions. I will investigate and fix it. Thank you!

karakayasemi:
> 
> 
> @karakayasemi could you increase coverage?

The command has not any unit test class. But, I can add tests by using Symfony's `CommandTester` class.

apps/files/lib/Command/TransferOwnership.php
phil-davis: We had to do similar `strpos` path-parsing logic recently in https://github.com/owncloud/core/pull/36170/files - it might give some clues, or you can think of a cleaner way.

micbar:
> > @karakayasemi could you increase coverage?
> 
> The command has not any unit test class. But, I can add tests by using Symfony's `CommandTester` class.

Yes, there are good examples in other apps.

micbar (changes requested):
Can you add a unit test where you transfer a subfolder and have another folder with shares which should not be transferred?
Like: testTransferSpecificFolderSkipOthers()

micbar:
@karakayasemi Could you rebase?
After rebase you will get the new changelog generator templates.
You have the honour to place one of the the first changelog files into [changelog](https://github.com/owncloud/core/changelog/). Just create a folder `unreleased` there if it does not exist and put a file there with your PR as a filename.
Write the changelog into that file according to https://github.com/owncloud/core/blob/master/changelog/TEMPLATE

micbar (approved):

PR #36225, opened 2019-09-30
Take accounts.enable_medial_search into account while searching for remote users in share autocomplete

VicDeo:
## Description
Add `matchMode` search option for carddav backend. Make it `any` by default and switch to `start` when accounts medial search is off
 
## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3495

## Motivation and Context
Use medial search only if it is enabled when searching for users synced via federation app

## How Has This Been Tested?
1. Setup 2 ownclouds and add them as trusted to each other at Settings -> Admin -> Sharing -> Federation
2. Execute cron on both servers e.g. with `occ system:cron`
3. Execute `occ federation:sync-addressbooks`
4. Disable account medial search on both
5 Try share autocomplete using  the middle chars of remote users usernames 

#### Expected 
No remote users in autocomplete as medial search is off
 
#### Actual 
Remote users with the search patterns in the middle of username are still proposed 

medial

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Open tasks:

### DIFF
--- a/apps/dav/lib/CardDAV/AddressBookImpl.php
+++ b/apps/dav/lib/CardDAV/AddressBookImpl.php
@@ -91,7 +91,7 @@ class AddressBookImpl implements IAddressBook {
 	 * @since 5.0.0
 	 */
 	public function search($pattern, $searchProperties, $options, $limit = null, $offset = null) {
-		$results = $this->backend->search($this->getKey(), $pattern, $searchProperties, $limit, $offset);
+		$results = $this->backend->searchEx($this->getKey(), $pattern, $searchProperties, $options, $limit, $offset);
 
 		$vCards = [];
 		foreach ($results as $result) {
--- a/apps/dav/lib/CardDAV/CardDavBackend.php
+++ b/apps/dav/lib/CardDAV/CardDavBackend.php
@@ -812,14 +812,52 @@ class CardDavBackend implements BackendInterface, SyncSupport {
 	 * @return array an array of contacts which are arrays of key-value-pairs
 	 */
 	public function search($addressBookId, $pattern, $searchProperties, $limit = 100, $offset = 0) {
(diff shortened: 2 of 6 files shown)


karakayasemi (commented):
The code looks okay and it is working like expected. But, I have some doubts about the architecture. 

Since `accounts.enable_medial_search` is a system value, I feel like it should be forced in core implementation. In my POV, it should not be an option. The other usages of the same system value are also in this way. It is not selectable by API consumers like this one https://github.com/owncloud/core/blob/bd3df85448af0e32bf199d6772b5b5eaf8c02091/lib/private/User/AccountMapper.php#L156. Also, the following issue description is stating that this config is intended for Backend implementations https://github.com/owncloud/core/issues/33883#issue-390688876.

In this implementation, we expect every class which uses `ContactManager` or `CardDAV` backend search to be aware of this system config and send the correct option to obey the configured setting. Also when I look at `ContactManager` interface or its implementation, I can not see which are these search options.

I may have missed something, but I would keep it simple and just add a check in the CardDAV search for this config.

VicDeo:
@karakayasemi there is an internal conflict there.
I don't think dav  contacts search should depend on `accounts.enable_medial_search`
Contacts are NOT accounts, user backend, etc. 

Such kind of dependency might have side effects while working with address books.

micbar (approved):
LGTM

micbar:
@VicDeo Could you rebase and push?

micbar:
After rebase you will get the new changelog generator templates.
You have the honour to place the first changelog file into [changelog](https://github.com/owncloud/core/changelog/). Just create a folder `unreleased` there and put a file there with your PR as a filename.
Write the changelog into that file according to https://github.com/owncloud/core/blob/master/changelog/TEMPLATE

PR #36254, opened 2019-10-08
[For 10.4] Add dir/path sharing infos

lordelix:
## Description

Filelist shows share indicators on files and folders, that reside inside a shared folder / shared folders.
The sidebar share-tab reveals a detailed view of the share-tree including share-recipients and the respective folders the life in.

## Related Issue

https://github.com/owncloud/enterprise/issues/3453 (Discussion)
https://github.com/owncloud/enterprise/issues/3556 (Vis. Concept)

## How Has This Been Tested?

Manual testing in
- Google Chome
- Mozilla Firefox
- MSIE 11

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item, see [TEMPLATE](../changelog/README.md)


apps/files/js/filelist.js
```diff
@@ -1723,6 +1723,38 @@
 			return OC.linkToRemoteBase('dav') + '/files/' + uid + encodedPath;
 		},
 
+		getDirShareInfo: function(dir) {
```
PVince81: missing JS Doc, what does this function do and return ?

apps/files/js/filelist.js
```diff
@@ -1723,6 +1723,38 @@
...
+						name : dir.name,
+						shareTypes : dir.shareTypes
+					})
+				}).fail(function(error) {
+					reject(error)
+				})
+			})
+		},
+
+		getPathShareInfo: function(path) {
```
PVince81: JSDoc missing

apps/files/js/filelist.js
```diff
@@ -1723,6 +1723,38 @@
...
+			var crumbs     = [];
+			var pathToHere = '';
+			var parts      = (path === '/') ? ['/'] : path.split('/')
```
PVince81: what if path is an empty string ? it happens sometimes...

PVince81 (commented):
Looks fine so far, except for the missing docs.

Now to see how this will be used

apps/files/css/files.css
```diff
@@ -512,23 +512,69 @@ table td.filename .uploadtext {
...
 	bottom: 4px;
 	right: 0;
 	z-index: 10;
 }
 
-html.ie8 #fileList tr td.filename > .selectCheckBox {
-	filter: alpha(opacity=0);
-	opacity: 0;
+#fileList tr:not(:hover):not(.selected) td.filename .thumbnail.sharetree-item:after {
+	position: absolute;
```
PVince81: indent

apps/files/css/files.css
```diff
@@ -512,23 +512,69 @@ table td.filename .uploadtext {
...
+    display: block;
+    content: '';
+    bottom: 0px;
+    right: -5px;
+    z-index: 10;
+    background-image: url(../img/shareinfo.svg);
```
PVince81: any chance to use an existing image / icon class ? else themes will need to override this one additionally

apps/files/css/files.css
```diff
@@ -512,23 +512,69 @@ table td.filename .uploadtext {
...
+}
+
+.shareTree-item-avatar {
+	border-radius: 50%;
 	float: left;
-	top: 0;
-	margin: 32px 0 4px 32px;
-	/* bigger clickable area doesn’t work in FF width:2.8em; height:2.4em;*/
+	margin-right: 10px;
+	transform: translateY(5px);
```
PVince81: why not use margin instead of transform ?

apps/files/css/files.css
```diff
@@ -512,23 +512,69 @@ table td.filename .uploadtext {
...
+	display: block;
+}
+
+.shareTree-item-path {
+	display: block;
+}
+
+.shareTree-item-path:before,
+.shareTree-item-path:after {
+	content: '"';
```
PVince81: adding content with CSS feels hacky, unless there is no better way

apps/files/css/files.css
```diff
@@ -983,3 +1029,54 @@ html.ie8 #controls .button.new {
 	position: inherit;
 	right: inherit;
 }
+
+/*
+ * Directory sharing indocator
```
PVince81: s/indocator/indicator/

apps/files/css/files.css
```diff
@@ -983,3 +1029,54 @@ html.ie8 #controls .button.new {
...
+ *
+ * #fff4cc - Background yellow
+ * #d5c999 - Border dirty yellow
+ * #927800 - Text dark dirty yellow
```
PVince81: assuming these are colors we already used elsewhere ?
also need to remember theming

apps/files/css/files.css
```diff
@@ -983,3 +1029,54 @@ html.ie8 #controls .button.new {
...
+ */
+
+#controls #shareinfo_indicator {
+	float: right;
```
PVince81: indent...

apps/files/js/filelist.js
```diff
@@ -356,6 +363,34 @@
 			OC.Plugins.attach('OCA.Files.FileList', this);
 		},
 
+		dirShareInfoIndicator: function () {
+			var $el = $('#shareinfo_indicator')
+
+			if ( !$el.length ) {
+
+				$el        = $('<div>', { id : 'shareinfo_indicator' });
+				$textbox   = $('<span>', { class : "text", text : t('core', 'shared') })
```
PVince81: missing `var` for these variables. jslint should have complained ?

apps/files/js/filelist.js
```diff
@@ -356,6 +363,34 @@
...
+			}
+
+			this.getPathShareInfo(this.getCurrentDirectory()).then(function(path) {
+				var shares = path.filter( function(dir) {
+					return dir.shareTypes.length > 0
+				})
+
+				// Stop if there are no shares present
+				if (shares.length === 0) {
+					$el.fadeOut('fast')
```
PVince81: this seems to contract "fadeIn" below.

is the initial state visible or hidden ?
if visible, you can only fade out, not fade in
if hidden, you can only fade in, not fade out

apps/files/js/filelist.js
```diff
@@ -1048,6 +1083,11 @@
 			$(window).scrollTop(0);
 
 			this.$fileList.trigger(jQuery.Event('updated'));
+			
+			this._setShareTree().then(function() {
```
PVince81: let's call those `_updateShareXXX()` because "set" feels like a regular setter which is usually not asynchronous

apps/files/js/filelist.js
```diff
@@ -1723,6 +1770,135 @@
...
+						name : dir.name,
+						path : dir.path
+					}
+
+					if (dir.shareTypes === undefined) {
+						shareInfo.shares = []
+						resolve(shareInfo)
+					}
+					else {
+						$.get( OC.linkToOCS('apps/files_sharing/api/v1', 2) + 'shares?' + OC.buildQueryString({format: 'json', path: (dir.path + '/' + dir.name)}), function(e) {
```
PVince81: so this is for getting shares for children and not parents ? please add comments for both calls just to clarify which shares (incoming/outgoing/...) we are fetching to avoid confusion

apps/files/js/filelist.js
```diff
@@ -1723,6 +1770,135 @@
...
+			})
+		},
+
+		_setShareTreeIcons: function() {
+			if (!this._shareTree.length)
+				return
+
+			// Add share-tree icon to files and folders
+			// each per <tr> in the table
+			$('#fileList tr td.filename .thumbnail:not(.sharetree-item)').addClass('sharetree-item')
```
PVince81: use `this.$el` or similar, please avoid using global access.

also consider that such global access will not work for the "shared with you" and other file lists.
did you test this PR with those ? we should not fetch share info there

apps/files/js/filelist.js
```diff
@@ -1723,6 +1770,135 @@
...
+			var $shareTabView  = $('#shareTabView .dialogContainer');
+			var $shareTreeView = $('<div>', { class : 'shareTreeView' , html : '<ul></ul>'});
+
+			$shareTabView.ready( function() {
+
+				if (!self._shareTree.length)
+					return
+
+				// Add items to the shareview in the sidebar ... if it's open
+				if (!$('#app-sidebar').hasClass('disappear')) {
```
PVince81: I suggest to extend core/js/apps.js to add a method to check visibility

apps/files/js/filelist.js
```diff
@@ -1723,6 +1770,135 @@
...
+				return false
+			}
+
+			var client     = this.filesClient
+			var options    = {
+				properties : ['{' + OC.Files.Client.NS_OWNCLOUD + '}share-types']
+			}
+
+			return new Promise( function(resolve, reject) {
+				client.getFileInfo(dir, options).done(function(s, dir) {
```
PVince81: when navigating from the top to bottom, are you reloading all shares for all the path sections ?

for example if I'm in "/first" we have already loaded all share information for "/first"
then I navigate into "/first/second": when loading the list we should then only load the share information for "second" as we already have the one for "/first", we can cache it.

PVince81 (changes requested):
See comments.

apps/files/js/filelist.js
felixheidecke: I was about to lay out the problem(s) I had with none existing folder IDs yet it just came into my mind that I use each path `/first`  ... `/first/second` as the keys.

I'll get to it right away 😸

felixheidecke: @PVince81 https://github.com/owncloud/core/pull/36254/commits/18ef04c780a4f28cddb6ec7fb4666c80b85f640e#diff-dafefd9519924bec4efa9f5af4af8e68R1789

felixheidecke:
@PVince81 there is no need to merge it yet, but if the code is fine, we can supply a patch to the customer.

apps/files/js/filelist.js
felixheidecke: this (and the above) will be addressed, if the code comes back.

apps/files/css/files.css
```diff
@@ -983,3 +1026,60 @@ html.ie8 #controls .button.new {
...
+ *
+ * #fff4cc - Background yellow
+ * #d5c999 - Border dirty yellow
+ * #927800 - Text dark dirty yellow
+ */
+
+ /*
+ ---------------------------------------------------
+ Keep the code, since final descision about
+ sharing indicator on the top right is still pending
```
PVince81: the code is kept anyway in git history ?

apps/files/js/filelist.js
```diff
@@ -356,6 +363,42 @@
 			OC.Plugins.attach('OCA.Files.FileList', this);
 		},
 
+		/*
+
+		---------------------------------------------------
```
PVince81: git history

apps/files/js/filelist.js
```diff
@@ -1723,6 +1778,164 @@
...
+		},
+
+		_setShareTreeIcons: function() {
+
+			if (!this.partOfSharePath())
+				return
+
+			// Add share-tree icon to files and folders
+			// each per <tr> in the table
+			this.$el.find('#fileList tr td.filename .thumbnail:not(.sharetree-item)').addClass('sharetree-item')
```
PVince81: don't match by id, use this.$fileList then, it should already contain the "fileList" element

PVince81: by the way, this won't be enough: what happens if you scroll down and the icon is supposed to be on the second page of results ? please have a test with 100+ items

if it doesn't work you'll likely need to tap into the "fileActionsReady" event which is fired after new elements are appended, and then add the icon in question. see https://github.com/owncloud/core/blob/master/apps/files_sharing/js/share.js#L128

apps/files/js/filelist.js
```diff
@@ -1723,6 +1778,164 @@
...
+							if (share.share_type === 0) {
+								let $name   = $('<strong>', { class : 'shareTree-item-name', text : share.share_with_displayname })
+								let $avatar = $('<div>',    { class : 'shareTree-item-avatar' })
+
+								$('<li class="shareTree-item">').append( $avatar, $name, $path).appendTo($shareTreeView.find('> ul'))
+								$avatar.avatar(share.share_with, 32)
+							}
+
+							// link shares
+							else if (share.share_type === 3) {
```
PVince81: `OC.Share.SHARE_TYPE_LINK`. same for the user group share

PVince81 (commented):
Looks fine so far except for the concern about scrolling to the next page where items are rendered dynamically. Once this is solved this could be tested as patch.

Also, before we merge this we must add JS unit tests.

felixheidecke:
I think we are good for delivering the patch to test. If all is fine, I can get to adding unit tests.

apps/files/js/filelist.js
```diff
@@ -1723,6 +1741,152 @@
...
+			var client     = this.filesClient;
+			var options    = {
+				properties : ['{' + OC.Files.Client.NS_OWNCLOUD + '}share-types']
+			}
+
+			return new Promise( function(resolve, reject) {
+				client.getFileInfo(dir, options).done(function(s, dir) {
+
+					var glue = (dir.path === '/') ? '' : '/'
+					var path = dir.path + glue + dir.name
```
PVince81: please note that there are utility functions for joining paths: `OC.joinPaths` that handle slashes

PVince81 (commented):
Apart from the cache purge logic the code looks fine now.

felixheidecke:
@PVince81 I couldn't keep myself from writing a little event system called OCE, because it made a couple of things much easier, especially catching the initial render of the linklist as you will see.

https://github.com/owncloud/core/pull/36254/files#diff-ffd2df04c44ef2411d4fab6e43174ed2

If we could set an env variable from DEV to PROD and vice versa, we could keep the logging here.

@pmaier1 User/Group and link shares are now separated.

core/js/events.js
```diff
@@ -0,0 +1,116 @@
...
+ * This file is licensed under the Affero General Public License version 3
+ * or later.
+ *
+ * See the COPYING-README file.
+ *
+ */
+
+/** @namespace */
+
+var OCE = {
```
PVince81: I disagree with adding yet another event system.

We already have jquery, you can use this on the body element if you need global events. There are places where we already do this, for example in FileActions.

Or if you need something cleaner but local to classes, you can extend the class with Backbone's event system.

apps/files/js/filelist.js
```diff
@@ -1459,20 +1459,27 @@
...
+					resolve()
+					return
 				}
-			});
+				self._setCurrentDir(targetDir, changeUrl, fileId);
+				// discard finished uploads list, we'll get it through a regular reload
+				self._uploads = {};
+				self.reload().then(function(success){
+					if (!success) {
+						self.changeDirectory(currentDir, true);
```
PVince81: no risk of infinite loop here ?

apps/files/js/filelist.js
```diff
@@ -1723,6 +1741,152 @@
...
+			if (dir === '/') {
+				this._shareTreeCache = {}
+				return Promise.resolve()
+			}
+
+			return this.getPathShareInfo(dir).then(() => {
+
+				// Purge deeper children
+				_.each(self._shareTreeCache, function(path, key) {
+					if (key > dir) {
```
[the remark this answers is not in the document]
PVince81: seems this is still unsolved ?

apps/files/js/filelist.js
```diff
@@ -360,6 +360,15 @@
 				}
 			}
 
+			// Setup event listeners
+			OCE.on('linkShareViewRendered', setShareTreeLinkView = function() {
```
PVince81: I dislike that these events are global. They should be local to the sharing sidebar panel and the FileList instance has access to the sidebar.

Normally the FileList doesn't know anything about sharing and this "knowledge" is injected by the share.js plugins by extension. Please check whether you can extend this function from inside the share.js plugin instead, the plugin already knows about the sidebar panel so the sidebar panel can trigger events on itself. (it already triggers other events (backbone style) like "sharesUpdated" or something, so use the same approach)

core/js/events.js
felixheidecke: 😢

apps/files/js/filelist.js
felixheidecke: The only thing I did was wrapping it in a Promise so I could act after the reload was done.
This seems to have worked for the last 5 years ;-)

micbar:
@felixheidecke Can you adress the input from @PVince81 ?
And I would be glad about some JS unit tests

felixheidecke:
@micbar @PVince81 just added the changes as discussed. If the current implementation behaves as desired, I'll add tests.

felixheidecke:
I added three tests for sharing, yet I can't figure out why this one here calles the fakeServer twice 🤔 

```
PhantomJS 2.1.1 (Linux 0.0.0) OCA.Sharing.PublicApp tests File list Uses public webdav endpoint FAILED
        Expected 2 to equal 1.
        apps/files_sharing/tests/js/publicAppSpec.js:97:46
```

phil-davis:
Added QA-team and acceptance test labels.
@skshetry please delegate someone to make some UI acceptance tests for this.
Needs to just check that the share-indicator icon does (or not) appear when viewing a file, folder, sub-folder, file in folder that is shared.

phil-davis (changes requested):
Needs some acceptance tests.
I am adding this "request changes" review so that the PR his held until the acceptance tests are done.
QA-team will push a commit with the acceptance tests

skshetry:
Assigning @kiranparajuli589, for writing the test.

phil-davis:
See #36534 for a rebased/squashed version of this.

individual-it:
closing in favour of #36534

PR #36256, opened 2019-10-08
Fix scan for Guest users

pako81:
## Description

This PR fixes the files:scan (and the correspondent background job `OCA\Files\BackgroundJob\ScanFiles`) for Guest users.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/3561

## Motivation and Context

Currently, we are spamming the owncloud.log with `Forbidden` exceptions every time the background job or files:scan is processing Guest users.

## How Has This Been Tested?

Manually by running `files:scan` for Guest users and observing the scan correctly processing those users and by observing that the `OCA\Files\BackgroundJob\ScanFiles` background job does not trigger any `Forbideen` exception anymore.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item

### DIFF
--- a/lib/private/Files/Utils/Scanner.php
+++ b/lib/private/Files/Utils/Scanner.php
@@ -188,6 +188,16 @@ class Scanner extends PublicEmitter {
 		if ($storage->instanceOfStorage('\OC\Files\Storage\Home') and
 			(!$storage->isCreatable('') or !$storage->isCreatable('files'))
 		) {
+			$isGuest = \OC::$server->getConfig()->getUserValue(
+					$this->user,
+				'owncloud',
+				'isGuest',
+				false
+				);
+			if ($isGuest) {
+				return true;
+			}
+
 			if ($storage->file_exists('') or $storage->getCache()->inCache('')) {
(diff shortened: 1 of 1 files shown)


jvillafanez:
The comment just above those lines:
> // if the home storage isn't writable then the scanner is run as the wrong user

Why the guest user can't write in the storage? Is it the right condition? Should we change the condition to "isn't readable" instead?

The other question I have is: is it possible to have a read-only home storage? Why are we throwing an exception if the home storage is read-only (or so it seems looking at the conditions)

I'd change the condition to check if the storage is readable. I expect guests users to be able to read at least the root folder of his own storage, even if they can't write in it.
https://github.com/owncloud/core/blob/master/lib/private/legacy/util.php#L228

phil-davis:
status of this? scheduling for review/adjustment? or?
@micbar

pako81:
@phil-davis alternative approach https://github.com/owncloud/core/pull/36288 which has been already merged. This can be closed - thx

PR #36258, opened 2019-10-08
Fix personal setting for apps which are not whitelisted

sharidas:
When apps are not whitelisted for guest user
they should not be visible in the personal settings.

## Description

When apps are not enabled for guest users, they should not be listed in personal settings. In this change a new method is introduced in the user class `getExtendedAttributes()`. The purpose of this method is to grab the attributes related to the user for the apps. So when this method is invoked an event is emitted which is listened by the apps, and populate the event array with the attributes in the form of key => value. So when logged in as guest user the guest app would populate the event array with `isGuest` and `whiteListedApps`. Which could be used by core or any other apps. This feature is used SettingsManagers `loadPanel` to filter out the apps which are not whitelisted.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

Do not show the apps which are not whitelisted for the guest user.

## How Has This Been Tested?

- Enable customgroups. Make sure its not listed as whitelisted apps in guest configuration.
- As guest user navigate to the settings page.
- The settings page should not show customgroups. And it does not.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item, see [TEMPLATE](../changelog/README.md)

### DIFF
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -45,6 +45,7 @@ use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IUser;
 use OCP\IUserSession;
+use OCP\User\NotPermittedActionException;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 class AppManager implements IAppManager {
@@ -160,9 +161,9 @@ class AppManager implements IAppManager {
 	 */
 	public function getEnabledAppsForUser(IUser $user = null) {
 		$apps = $this->getInstalledAppsValues();
-		$appsForUser = \array_filter($apps, function ($enabled) use ($user) {
-			return $this->checkAppForUser($enabled, $user);
-		});
+		$appsForUser = \array_filter($apps, function ($enabled, $appName) use ($user) {
+			return $this->checkAppForUser($enabled, $appName, $user);
+		}, ARRAY_FILTER_USE_BOTH);
 		return \array_keys($appsForUser);
 	}
 
@@ -182,7 +183,7 @@ class AppManager implements IAppManager {
 		}
 		$installedApps = $this->getInstalledAppsValues();
 		if (isset($installedApps[$appId])) {
-			return $this->checkAppForUser($installedApps[$appId], $user);
+			return $this->checkAppForUser($installedApps[$appId], $appId, $user);
 		} else {
 			return false;
 		}
@@ -191,9 +192,28 @@ class AppManager implements IAppManager {
 	/**
 	 * @param string $enabled
 	 * @param IUser $user
+	 * @param string $appName
 	 * @return bool
+	 * @throws NotPermittedActionException
 	 */
-	private function checkAppForUser($enabled, $user) {
+	private function checkAppForUser($enabled, $appName, $user) {
+		if ($user !== null) {
+			$userAppAttributes = $user->getExtendedAttributes();
+			/**
+			 * Guests will only have access to some whitelisted apps
+			 * - If the "whitelistedAppsForGuests" isn't found, the user is not considered guest and the user will have access to all the enabled apps.
+			 * - If the "whitelistedAppsForGuests" list is empty, the guest won't be able to use any app
+			 * - If the "whitelistedAppsForGuests" has some apps, only those apps will be available for that guest
+			 *
+			 */
+			if (isset($userAppAttributes['whitelistedAppsForGuests'])) {
+				$whiteListedAppsForGuest = $userAppAttributes['whitelistedAppsForGuests'];
+				if (\is_array($whiteListedAppsForGuest) && !\in_array($appName, $whiteListedAppsForGuest)) {
+					return false;
+				}
+			}
+		}
+
 		if ($enabled === 'yes') {
 			return true;
 		} elseif ($user === null) {
--- a/lib/private/User/User.php
+++ b/lib/private/User/User.php
@@ -11,6 +11,7 @@
  * @author Tom Needham <tom@owncloud.com>
  * @author Victor Dubiniuk <dubiniuk@owncloud.com>
  * @author Vincent Petry <pvince81@owncloud.com>
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
  *
  * @copyright Copyright (c) 2018, ownCloud GmbH
  * @license AGPL-3.0
@@ -37,13 +38,16 @@ use OC_Helper;
 use OCP\Events\EventEmitterTrait;
 use OCP\IAvatarManager;
 use OCP\IImage;
+use OCP\ILogger;
 use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IConfig;
 use OCP\IUserBackend;
 use OCP\IUserSession;
+use OCP\User\NotPermittedActionException;
 use OCP\PreConditionNotMetException;
 use OCP\User\IChangePasswordBackend;
+use OCP\User\UserExtendedAttributesEvent;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
@@ -77,6 +81,18 @@ class User implements IUser {
 	/** @var Session  */
 	private $userSession;
 
+	/** @var null|array */
+	private $userExtendedAttributes = null;
+
+	/**
+	 * This flag is true by default. This flag when set to false
+	 * would not allow the set operations in this class. Or in other
+	 * words the update to the accounts table will be restricted through this flag.
+	 * This flag is only modified inside getExtendedAttributes method.
+	 * @var bool
+	 */
+	private $allowUserAccountUpdate = true;
+
 	/**
 	 * User constructor.
 	 *
@@ -88,6 +104,7 @@ class User implements IUser {
 	 * @param EventDispatcher|null $eventDispatcher
 	 * @param \OC\Group\Manager|null $groupManager
 	 * @param Session|null $userSession
+	 * @param ILogger|null $logger
 	 */
 	public function __construct(Account $account, AccountMapper $mapper, $emitter = null, IConfig $config = null,
 								$urlGenerator = null, EventDispatcher $eventDispatcher = null,
@@ -96,6 +113,9 @@ class User implements IUser {
 		$this->account = $account;
 		$this->mapper = $mapper;
 		$this->emitter = $emitter;
+		if ($eventDispatcher === null) {
+			$eventDispatcher = \OC::$server->getEventDispatcher();
+		}
 		$this->eventDispatcher = $eventDispatcher;
 		if ($config === null) {
 			$config = \OC::$server->getConfig();
@@ -140,8 +160,13 @@ class User implements IUser {
 	 * TODO move username to account table
 	 *
 	 * @param string $userName
+	 * @throws NotPermittedActionException
 	 */
 	public function setUserName($userName) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		$currentUserName = $this->getUserName();
 		if ($userName !== $currentUserName) {
 			$uid = $this->getUID();
@@ -171,8 +196,13 @@ class User implements IUser {
 	 *
 	 * @param string $displayName
 	 * @return bool
+	 * @throws NotPermittedActionException
 	 */
 	public function setDisplayName($displayName) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		if (!$this->canChangeDisplayName()) {
 			return false;
 		}
@@ -199,9 +229,14 @@ class User implements IUser {
 	 *
 	 * @param string|null $mailAddress
 	 * @return void
+	 * @throws NotPermittedActionException
 	 * @since 9.0.0
 	 */
 	public function setEMailAddress($mailAddress) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		$mailAddress = \trim($mailAddress);
 		if ($mailAddress === $this->account->getEmail()) {
 			return;
@@ -289,8 +324,13 @@ class User implements IUser {
 	 * @param string $recoveryPassword for the encryption app to reset encryption keys
 	 * @return bool
 	 * @throws \InvalidArgumentException
+	 * @throws NotPermittedActionException
 	 */
 	public function setPassword($password, $recoveryPassword = null) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		if (\OCP\Util::isEmptyString($password)) {
 			throw new \InvalidArgumentException('Password cannot be empty');
 		}
@@ -410,11 +450,16 @@ class User implements IUser {
 	}
 
 	/**
-	 * set the enabled status for the user
+	 * Set the enabled status for the user
 	 *
 	 * @param bool $enabled
+	 * @throws NotPermittedActionException
 	 */
 	public function setEnabled($enabled) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		if ($enabled === true) {
 			$this->account->setState(Account::STATE_ENABLED);
 		} else {
@@ -452,13 +497,19 @@ class User implements IUser {
 	}
 
 	/**
-	 * set the users' quota
+	 * Set the users' quota
 	 *
 	 * @param string $quota
 	 * @return void
+	 * @throws NotPermittedActionException
+	 * @throws PreConditionNotMetException
 	 * @since 9.0.0
 	 */
 	public function setQuota($quota) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		if ($quota !== 'none' and $quota !== 'default') {
 			$quota = OC_Helper::computerFileSize($quota);
 			$quota = OC_Helper::humanFileSize($quota);
@@ -538,9 +589,14 @@ class User implements IUser {
 
 	/**
 	 * @param string[] $terms
+	 * @throws NotPermittedActionException
 	 * @since 10.0.1
 	 */
 	public function setSearchTerms(array $terms) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+
 		// Check length of terms, cut if too long
 		$terms = \array_map(function ($term) {
 			return \substr($term, 0, 191);
@@ -554,5 +610,66 @@ class User implements IUser {
 	 */
 	public function getAccountId() {
 		return $this->account->getId();
+	}
+
+	/**
+	 * get the attributes of user for apps
+	 * This method sends event which is listened by the apps. The apps would add attributes which
+	 * are specific to this user. Say for example a user might have access to a blog site, in such
+	 * case the app which is responsible for this control could listen to this event and
+	 * add an attribute say:
+	 * "blogSite" => "https://foo/bar"
+	 * Apps add attributes and their value in the form of key => value. The userExtendedAttributes
+	 * does not care which app added the attributes. It only considers about the
+	 * attributes.
+	 * The argument clearCache is used to clear userExtendedAttributes. If there are
+	 * external apps involved or under any circumstance we know there will be delay
+	 * in response from the app, then its safe to use clearCache as false.
+	 * New event is triggered under the following conditions:
+	 * - if the userExtendedAttributes is null or empty array ( even if clearCache is set to false, in this condition, event will be triggered )
+	 * - if clearCache is set to true
+	 * The flag allowUserAccountUpdate is set to true by default. This flag is set to false before event is emitted.
+	 * This flag is checked on all the set operations, in this class to make sure no user account
+	 * table update is allowed when the extended attributes are provided by the apps. Once the
+	 * event listeners have done their task, the flag is set back to true.
+	 * The exception is thrown when the listener tries call this method again. This is to
+	 * prevent infinite loop. Also the exceptions thrown during any operation by the listeners
+	 * are allowed to go up. The exceptions are not caught in this method.
+	 *
+	 *
+	 * @param bool $clearCache, set to true if user attributes should be created every time, else false is set to reuse the userExtendedAttributes cache.
+	 * @return array
+	 * @throws NotPermittedActionException
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes($clearCache = false) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+		/**
+		 * using empty because userExtendedAttributes could either be null or empty array
+		 * or an array of attributes
+		 */
+		if ($clearCache || !isset($this->userExtendedAttributes) || [] === $this->userExtendedAttributes) {
+			$userExtendedAttributesEvent = new UserExtendedAttributesEvent($this);
+
+			/**
+			 * Restrict the user accounts table update for the set operations as the apps
+			 * listening to the event below will be providing extended attributes of this
+			 * user.
+			 */
+			$this->allowUserAccountUpdate = false;
+			try {
+				$this->eventDispatcher->dispatch(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, $userExtendedAttributesEvent);
+			} finally {
+				//Reset the flag to true so that account table can now be allowed to update.
+				$this->allowUserAccountUpdate = true;
+			}
+
+			//Just overwrite the userExtendedAttributes if clearCache is true
+			$this->userExtendedAttributes = $userExtendedAttributesEvent->getAttributes();
+		}
+
+		return $this->userExtendedAttributes;
 	}
 }
--- a/lib/public/IUser.php
+++ b/lib/public/IUser.php
@@ -25,6 +25,8 @@
 
 namespace OCP;
 
+use OCP\User\NotPermittedActionException;
+
 /**
  * Interface IUser
  *
@@ -59,6 +61,7 @@ interface IUser {
 	 * set the user name
 	 *
 	 * @param string $userName
+	 * @throws NotPermittedActionException
 	 * @since 10.0.10
 	 */
 	public function setUserName($userName);
@@ -76,6 +79,7 @@ interface IUser {
 	 *
 	 * @param string $displayName
 	 * @return bool
+	 * @throws NotPermittedActionException
 	 * @since 8.0.0
 	 */
 	public function setDisplayName($displayName);
@@ -109,6 +113,7 @@ interface IUser {
 	 * @param string $password
 	 * @param string $recoveryPassword for the encryption app to reset encryption keys
 	 * @return bool
+	 * @throws NotPermittedActionException
 	 * @since 8.0.0
 	 */
 	public function setPassword($password, $recoveryPassword = null);
@@ -162,9 +167,10 @@ interface IUser {
 	public function isEnabled();
 
 	/**
-	 * set the enabled status for the user
+	 * Set the enabled status for the user
 	 *
 	 * @param bool $enabled
+	 * @throws NotPermittedActionException
 	 * @since 8.0.0
 	 */
 	public function setEnabled($enabled);
@@ -199,6 +205,7 @@ interface IUser {
 	 *
 	 * @param string|null $mailAddress
 	 * @return void
+	 * @throws NotPermittedActionException
 	 * @since 9.0.0
 	 */
 	public function setEMailAddress($mailAddress);
@@ -214,10 +221,11 @@ interface IUser {
 	public function getQuota();
 
 	/**
-	 * set the users' quota
+	 * Set the users' quota
 	 *
 	 * @param string $quota
 	 * @return void
+	 * @throws NotPermittedActionException
 	 * @since 9.0.0
 	 */
 	public function setQuota($quota);
@@ -227,6 +235,7 @@ interface IUser {
 	 *
 	 * @param array $terms
 	 * @return void
+	 * @throws NotPermittedActionException
 	 * @since 10.0.1
 	 */
 	public function setSearchTerms(array $terms);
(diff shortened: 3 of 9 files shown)


lib/private/User/User.php
```diff
@@ -555,4 +555,21 @@ public function setSearchTerms(array $terms) {
...
+	 *
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes() {
+		if ($this->eventDispatcher === null) {
+			$this->eventDispatcher = \OC::$server->getEventDispatcher();
+		}
+
+		$event = new GenericEvent('appAttributesForUser', []);
```
jvillafanez: Use a custom event so we can have full control on how we want to store the information

lib/private/App/AppManager.php
```diff
@@ -160,9 +160,20 @@ public function getInstalledApps() {
 	 */
 	public function getEnabledAppsForUser(IUser $user = null) {
 		$apps = $this->getInstalledAppsValues();
-		$appsForUser = \array_filter($apps, function ($enabled) use ($user) {
-			return $this->checkAppForUser($enabled, $user);
-		});
+		$whiteListedAppsForGuest = [];
```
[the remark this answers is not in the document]
sharidas: Moved to the `checkAppForUser` method.

lib/private/App/AppManager.php
```diff
@@ -182,7 +193,15 @@ public function isEnabledForUser($appId, $user = null) {
 		}
 		$installedApps = $this->getInstalledAppsValues();
 		if (isset($installedApps[$appId])) {
-			return $this->checkAppForUser($installedApps[$appId], $user);
+			$whiteListedApps = [];
```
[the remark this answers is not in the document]
sharidas: Moved to `checkAppForUser` method.

lib/private/App/AppManager.php
```diff
@@ -191,9 +210,14 @@ public function isEnabledForUser($appId, $user = null) {
 	/**
 	 * @param string $enabled
 	 * @param IUser $user
+	 * @param string|null $appName
+	 * @param array $whiteListedAppsForGuest
 	 * @return bool
 	 */
-	private function checkAppForUser($enabled, $user) {
+	private function checkAppForUser($enabled, $user, $appName = null, $whiteListedAppsForGuest = []) {
```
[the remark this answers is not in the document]
sharidas: Removed the whitelist and kept the appName.

lib/private/User/User.php
```diff
@@ -555,4 +556,23 @@ public function setSearchTerms(array $terms) {
...
+
+	/**
+	 * get the attributes of user for apps
+	 *
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes() {
+		if ($this->eventDispatcher === null) {
+			$this->eventDispatcher = \OC::$server->getEventDispatcher();
```
jvillafanez: I don't like this. This should be done in the constructor, not here. Nobody will expect that the server's event dispatcher will be assigned in a "get" method.
If you don't want to throw an exception, returning "false" could be fine. (whatever you decide, make sure is documented in the interface)

sharidas: Correct. We can safely remove this. I was bit worried when I saw this line in the file https://github.com/owncloud/core/blob/master/lib/private/User/User.php#L425. I have tested it by removing it in my instance and it works. So in the next version, it won't be there.

jvillafanez: The event dispatcher can be null (for a reason I don't understand). This is something we need to handle, unless we force the event dispatcher to be not null in the constructor.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,73 @@
...
+		return $this->attributes;
+	}
+
+	/**
+	 * Set the attributes
+	 *
+	 * @param string $appName
+	 * @param array $data
+	 */
+	public function setAttributes($app, $data) {
```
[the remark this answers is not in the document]
sharidas: > How are we going to deal with multiple apps? I think we could have a "first in wins" policy: if the attribute doesn't exists, write it and return false; it it already exists, ignore it and return false

There is another possibile flaw I see here, which is say for example if we use the policy to only add new attributes which don't exist and leave if the attribute already exist. Then it could also happen for example the user updated the app attribute and we might miss that, because we don't update. We only add the new attributes and skip if the attribute already exists. imho, shouldn't we consider to add/update it anyway? Or did I misread your feedback, please let me know your opinion.

lib/private/App/AppManager.php
```diff
@@ -191,9 +191,23 @@ public function isEnabledForUser($appId, $user = null) {
...
 	 * @return bool
 	 */
-	private function checkAppForUser($enabled, $user) {
+	private function checkAppForUser($enabled, $appName, $user) {
+		if ($user !== null) {
+			$userAppAttributes = $user->getExtendedAttributes();
+			/**
+			 * Check if the user is guest user, if so get the apps which it is allowed for.
+			 */
+			if (\array_key_exists('guests', $userAppAttributes)) {
```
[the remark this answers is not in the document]
sharidas: - If `guests` and `whitelistedApps` don't exist, then do the normal checks which already existed before this change.
- If the `whitelistedApps` is empty / null / string, then continue with the checks which already existed before this change.
- Expectation is whitelistedApps should be an array we would check if the `appName` is there in any of the elements in the array. Expectation here is when the code will access `$userAttributes['guests']['whitelistedApps']` we should get a list of strings. 
- When `whitelistedApps` is an empty array then proceed ahead with further checks which were already present before this change.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,73 @@
...
+	private $attributes;
+
+	/**
+	 * UserExtendedAttributesEvent constructor.
+	 *
+	 * @param string $event
+	 * @param array $attributes
+	 * @since 10.3.1
+	 */
+	public function __construct($event, $attributes) {
```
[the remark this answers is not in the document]
sharidas: Removed these parameters.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,82 @@
...
+ * @since 10.3.1
+ */
+class UserExtendedAttributesEvent extends Event {
+	const USER_EXTENDED_ATTRIBUTES = 'UserExtendedAttributesEvent';
+
+	/** @var array */
+	private $attributes = [];
+
+	/**
+	 * Get the extended attributes of user for an app
```
jvillafanez: To clarify, we're fetching user attributes, and we're asking for user attributes. I don't see why we need to know anything about apps here. As far as I know, the app is irrelevant.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,82 @@
...
+	 * If the appName key is not found in the attributes array then
+	 * an empty array is returned, else the result is returned.
+	 * It is also easy to fetch the attributes based on the app.
+	 *
+	 * @param string $appName
+	 * @return mixed
+	 */
+	public function getAttribute($appName) {
```
jvillafanez: As far as I know, this isn't needed. Keep things as easy as possible. Adding more methods will increase the confusion.
We only need one method for the listeners to add new user attributes, and another one for the `User` class to get those new attributes, nothing more.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,82 @@
...
+	 * Here the attributes of the app with key and value are set.
+	 * The reason to use appName as one of the keys is its easier to know if the
+	 * attributes are set for the app. If we don't use the appName then the the list
+	 * would be a scattered with attributes and it won't look nice.
+	 *
+	 * @param string $appName
+	 * @param string $attrubuteKey
+	 * @param mixed $attributeValue
+	 */
+	public function setAttributes($appName, $attrubuteKey, $attributeValue) {
```
jvillafanez: As said, return true or false to give the apps a chance to know if the attribute has been set or not.
(btw, there is a typo in the "attrubuteKey")

jvillafanez:
One additional thing: the event is "owned" or "controlled" by the `User` class. This means that it's the `User` class the one that should decide the format of the event.

The `User` class doesn't need to know what app has filled the event, and it doesn't have any need for the app name. In addition, the `getExtendedAttributes` method is expected to return user attributes in a "attr => value" fashion. Returning app names isn't part of the specification.
The expectation is that the method will return something like:
```
"isGuest" => true,
"loginName" => "Joe_Joe",
"dateOfBirth" => "3/3/1954",
"blogSite" => "https://server/umpa/lumpa"
.....
```
Noone cares which app added the "blogSite" key.

lib/private/App/AppManager.php
```diff
@@ -191,9 +191,27 @@ public function isEnabledForUser($appId, $user = null) {
...
 	 * @param IUser $user
+	 * @param string $appName
 	 * @return bool
 	 */
-	private function checkAppForUser($enabled, $user) {
+	private function checkAppForUser($enabled, $appName, $user) {
+		if ($user !== null) {
+			$userAppAttributes = $user->getExtendedAttributes();
+			/**
+			 * Check if the user is guest user, if so get the apps which it is allowed for.
```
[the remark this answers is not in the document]
sharidas: Added it into the doc.

lib/private/App/AppManager.php
```diff
@@ -191,9 +191,27 @@ public function isEnabledForUser($appId, $user = null) {
...
+			 * Check if guests key and whiteListedKey is available. If so fetch
+			 * the white listed apps to see if the appName is allowed for the user.
+			 * Else proceed with further checks in the method.
+			 */
+			if (\array_key_exists('guests', $userAppAttributes) &&
```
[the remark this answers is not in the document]
sharidas: Removed the usage of  app name in the code. And updated to use `isset` instead of `array_key_exists`.

lib/private/User/User.php
```diff
@@ -113,6 +123,10 @@ public function __construct(Account $account, AccountMapper $mapper, $emitter =
 		if ($this->userSession === null) {
 			$this->userSession = \OC::$server->getUserSession();
 		}
+		if ($userExtendedAttributesEvent === null) {
```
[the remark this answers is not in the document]
sharidas: Removed the event from the constructor. So new events are triggered each time, unless clear cache is set to false.

lib/public/IUser.php
```diff
@@ -238,4 +238,13 @@ public function setSearchTerms(array $terms);
...
 	public function getSearchTerms();
+
+	/**
+	 * Get attributes of user for apps
+	 * The attributes are of the form key => value map
+	 *
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes();
```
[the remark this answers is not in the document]
sharidas: Included a `clearCache` parameter which is optional. By default the `clearCache` is assigned to `true`. Meaning if new event is triggered and updates the `userAttributes`. If we set the `clearCache` to false we would just return the existing content of `userAttributes`.

lib/private/User/User.php
```diff
@@ -555,4 +569,16 @@ public function setSearchTerms(array $terms) {
 	public function getAccountId() {
 		return $this->account->getId();
 	}
+
+	/**
+	 * get the attributes of user for apps
```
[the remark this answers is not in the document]
sharidas: I have added documenation for this method. Let me know if this looks ok

lib/public/User/UserExtendedAttributesEvent.php
sharidas: Yes. I have removed the usage of apps or appnames from this code.

lib/public/User/UserExtendedAttributesEvent.php
sharidas: The `setAttributes` method now returns true if it has set the attribute, else it returns false.
Fixed the typo.

lib/private/User/User.php
```diff
@@ -555,4 +562,38 @@ public function setSearchTerms(array $terms) {
...
+	 * @param bool $clearCache, set to true if user attributes should be created every time, else false is set to reuse the userAttributes cache.
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes($clearCache = true) {
+		if (!$clearCache) {
+			return $this->userAttributes;
+		}
+
+		$userExtendedAttributesEvent = new UserExtendedAttributesEvent();
```
jvillafanez: I think we need to send this user object in the event. I mean, the guests app should verify this user is a guest before adding the whitelistedApps; ldap would need something to join the ldap information with this user info...

We can send this object inside the event. This is a more flexible option taking into account that we'll need to send at least the uid and any app can fetch this same user via the user manager.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @return array
+	 */
+	public function getAttributes() {
+		return $this->attributes;
+	}
+
+	/**
+	 * Set the attributes
+	 * Here the attributes of the app with key and value are set.
+	 * The reason to use appName as one of the keys is its easier to know if the
```
jvillafanez: no appName. Adjust the docs.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,70 @@
...
+	 * attributes are set for the app. If we don't use the appName then the the list
+	 * would be a scattered with attributes and it won't look nice.
+	 *
+	 * @param string $attributeKey
+	 * @param mixed $attributeValue
+	 * @return bool true if the attribute is set by the method false if the attribute is already present
+	 */
+	public function setAttributes($attributeKey, $attributeValue) {
+		if (isset($this->attributes[$attributeKey])) {
```
jvillafanez: we need to handle the "null value" case.

sharidas: Done

lib/public/User/UserExtendedAttributesEvent.php
sharidas: Verified and removed the appName in the file.

lib/private/User/User.php
sharidas: Updated the code. Now user object is passed through the event.

lib/private/User/User.php
```diff
@@ -555,4 +562,38 @@ public function setSearchTerms(array $terms) {
...
+	 * is triggered. If clearCache is set to true then new event is triggered and
+	 * hence userAttributes are updated accordingly.
+	 *
+	 * @param bool $clearCache, set to true if user attributes should be created every time, else false is set to reuse the userAttributes cache.
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes($clearCache = true) {
+		if (!$clearCache) {
+			return $this->userAttributes;
```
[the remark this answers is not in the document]
sharidas: So with the updated code, the event is triggered if:
- `clearCache` is `true` or
- `userExtendedAttributes` is either `null` or empty array ( in this case even if the `clearCache` is `false`, the event will be triggered.

lib/private/User/User.php
```diff
@@ -555,4 +562,38 @@ public function setSearchTerms(array $terms) {
...
+	 * in response from the app, then its safe to use clearCache as false. When
+	 * clearCache is false, the userAttributes array is used as it is. And no event
+	 * is triggered. If clearCache is set to true then new event is triggered and
+	 * hence userAttributes are updated accordingly.
+	 *
+	 * @param bool $clearCache, set to true if user attributes should be created every time, else false is set to reuse the userAttributes cache.
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes($clearCache = true) {
```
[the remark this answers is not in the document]
sharidas: Updated the `clearCache` to have default value `false`.

lib/private/User/User.php
```diff
@@ -77,6 +78,9 @@ class User implements IUser {
 	/** @var Session  */
 	private $userSession;
 
+	/** @var array */
+	private $userAttributes = [];
```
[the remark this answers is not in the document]
sharidas: Updated to `userExtendedAttributes`

lib/private/App/AppManager.php
```diff
@@ -191,9 +191,27 @@ public function isEnabledForUser($appId, $user = null) {
...
+		if ($user !== null) {
+			$userAppAttributes = $user->getExtendedAttributes();
+			/**
+			 * Guests will only have access to some whitelisted apps
+			 * - If the list is missing (no "whitelistedApps" key found), the guest will have access to all the apps as any normal user. The same will happen if the "whitelistedApps" isn't an array.
+			 * - If the list is empty, the guest won't be able to use any app
+			 * - If the list has some apps, only those apps will be available for that guest
+			 *
+			 */
+			if (isset($userAppAttributes['whiteListedApps'])) {
```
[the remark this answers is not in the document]
sharidas: updated to `whitelistedAppsForGuests`.

jvillafanez:
I think we'll need to protect some methods to be called from the apps. It could be risky to let the apps modify things, and calling again the "getExtendedAttributes" could trigger an infinite loop.

The idea I have for this is to change all these "problematic" methods to check for a flag and throw an exception if the flag is set. This flag will be private, set as active before sending the event, and disable the flag after the sending the event.
I think this is an easy way to allow or deny what methods can be called inside the listeners.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,86 @@
...
+	/**
+	 * Set the attributes
+	 * Here the attributes of the app with key and value are set.
+	 *
+	 * @param string $attributeKey
+	 * @param mixed $attributeValue
+	 * @return bool true if the attribute is set by the method false if the attribute is already present
+	 */
+	public function setAttributes($attributeKey, $attributeValue) {
+		if (isset($this->attributes[$attributeKey]) && ($this->attributes[$attributeKey] !== null)) {
```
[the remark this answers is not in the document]
sharidas: Updated code to use `isset` so that we can check if the index is there or not and if index is there and not null.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,86 @@
...
+
+	/** @var IUser */
+	private $user;
+
+	public function __construct(IUser $user) {
+		$this->user = $user;
+	}
+
+	/**
+	 * get the user who has triggered the event
```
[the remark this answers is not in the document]
sharidas: Updated the doc

lib/private/User/User.php
```diff
@@ -555,4 +562,42 @@ public function setSearchTerms(array $terms) {
...
+	 * @param bool $clearCache, set to true if user attributes should be created every time, else false is set to reuse the userExtendedAttributes cache.
+	 * @return array
+	 * @since 10.3.1
+	 */
+	public function getExtendedAttributes($clearCache = false) {
+		/**
+		 * using empty because userExtendedAttributes could either be null or empty array
+		 * or an array of attributes
+		 */
+		if ($clearCache || empty($this->userExtendedAttributes)) {
```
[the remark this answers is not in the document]
sharidas: Done. Instead of `empty`, check for null is added.

sharidas:
Well passing user object can be dangerous in one way, the apps can change user attributes through the user object. So lets say if the user id is passed, which is also dangerous because the app can access the same user object via userManager. I would like to grab inputs here, if we have to protect the set methods in the user object which can alter the user attributes? Another worrying factor here as @jvillafanez suggested in https://github.com/owncloud/core/pull/36258#issuecomment-541027084 if the app call `getExtendedAttributes` might cause an infinite loop?

I did an experiment using `clone` of user object, but unfortunately that does not help here. Because its a shallow copy and say if we try to change the display name, the change does reflect in the accounts table.

@micbar @PVince81 @VicDeo Let me know your thoughts regarding the same as this input will help us to move ahead with the pr.

lib/private/User/User.php
```diff
@@ -140,8 +158,13 @@ public function getUserName() {
 	 * TODO move username to account table
 	 *
 	 * @param string $userName
+	 * @throws NotPermitUserAccountException
 	 */
 	public function setUserName($userName) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermitUserAccountException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
```
jvillafanez: Note that this needs a change in the interface

lib/public/NotPermitUserAccountException.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace OCP;
+
+/**
+ * Class NotPermitUserAccountException
+ *
+ * @package OCP
+ * @since 10.3.1
+ */
+class NotPermitUserAccountException extends \Exception {
```
jvillafanez: Needs a better naming.
If this is intended to be used only for this specific case, `NotPermittedUserAction` seems a better choice. You can also put it inside the public/User folder.
If this could be used in a more generic way, `NotPermittedAction` could also fit.

lib/private/User/User.php
```diff
@@ -555,4 +609,55 @@ public function setSearchTerms(array $terms) {
...
+		if ($clearCache || $this->userExtendedAttributes === null) {
+			$userExtendedAttributesEvent = new UserExtendedAttributesEvent($this);
+
+			/**
+			 * Restrict the user accounts table update for the set operations as the apps
+			 * listening to the event below will be providing extended attributes of this
+			 * user.
+			 */
+			$this->allowUserAccountUpdate = false;
+			$this->eventDispatcher->dispatch(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, $userExtendedAttributesEvent);
```
jvillafanez: we need to handle exceptions triggered from inside the listeners, specially because the flag could remain active.

tests/lib/User/UserExtendedAttributesEventTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+		$this->user = $this->createMock(IUser::class);
+		$this->userExtendedAttributesEvent = new UserExtendedAttributesEvent($this->user);
+	}
+
+	/**
+	 * This test is to get the user whose extended attributes are to be fetched
+	 */
+	public function testGetUser() {
+		$this->assertEquals($this->userExtendedAttributesEvent->getUser(), $this->user);
```
jvillafanez: `assertSame` might be a better choice because we want to ensure it's the same instance, not a copy or a difference object with the same attributes.

tests/lib/User/UserExtendedAttributesEventTest.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	/**
+	 * @dataProvider providesSetAttributes
+	 * @param string $attributeKey
+	 * @param mixed $attributeVal
+	 * @param bool $expectedResult
+	 */
+	public function testSetAttributes($attributeKey, $attributeVal, $expectedResult) {
+		if ($attributeKey === 'one' && $expectedResult === false) {
```
jvillafanez: I don't understand this "if"

tests/lib/User/UserExtendedAttributesEventTest.php
```diff
@@ -0,0 +1,78 @@
...
+			$this->userExtendedAttributesEvent->setAttributes('one', [1, 2, 3]);
+		}
+		$this->assertEquals($this->userExtendedAttributesEvent->setAttributes($attributeKey, $attributeVal), $expectedResult);
+	}
+
+	public function testGetAttributes() {
```
jvillafanez: I think "testSetGetAttributes" is more aligned with what the test is doing.

tests/lib/User/UserTest.php
```diff
@@ -544,4 +545,25 @@ public function setTermsData() {
...
+	 * @param bool $clearCache
+	 */
+	public function testGetExtendedAttributes($clearCache) {
+		$userExtendedAttributesEvent = new UserExtendedAttributesEvent($this->user);
+
+		$this->eventDispatcher->expects($this->once())
+			->method('dispatch')
+			->with(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, $userExtendedAttributesEvent);
+
+		$this->assertNull($this->user->getExtendedAttributes($clearCache));
```
jvillafanez:
```
 * @param bool $clearCache, set to true if user attributes should be created every time, else false is set to reuse the userExtendedAttributes cache.
	 * @return array
	 * @since 10.3.1
	 */
	public function getExtendedAttributes($clearCache = false);
```
The function always returns an array. This test is invalid.

jvillafanez:
BTW, we'll need tests for the changes made in the `User` class

tests/lib/User/UserTest.php
sharidas: Done

tests/lib/User/UserExtendedAttributesEventTest.php
sharidas: Updated.

tests/lib/User/UserExtendedAttributesEventTest.php
sharidas: I have added the comment for this before this condition. So what I am trying to test here is:
- add attribute with key `one`.
- Again try to set the attribute with key `one`
- `setAttributes` should return false.

lib/private/User/User.php
sharidas: I have updated the code to handle the exception thrown. Also I have updated the AppManager where `getExtendedAttributes` is called. I have added try/catch block. The catch block is kept empty. So that the code flow will happen as if its a regular user and not a guest user.

lib/public/NotPermitUserAccountException.php
sharidas: Done

lib/private/User/User.php
sharidas: Updated the interface.

sharidas:
> BTW, we'll need tests for the changes made in the `User` class

I have updated the tests for the exceptions thrown for the `User` class.

tests/lib/User/UserTest.php
```diff
@@ -544,4 +546,103 @@ public function setTermsData() {
...
+	 */
+	public function testGetExtendedAttributesForException() {
+		$userExtendedAttributesEvent = new UserExtendedAttributesEvent($this->user);
+
+		$this->eventDispatcher->expects($this->once())
+			->method('dispatch')
+			->with(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, $userExtendedAttributesEvent)
+			->willThrowException(new NotPermittedActionException());
+
+		$this->assertEquals($this->user->getExtendedAttributes(), []);
```
jvillafanez: This makes things confusing: either you assert the call throws an exception, or you assert that the method returns an empty array.

lib/public/User/UserExtendedAttributesEvent.php
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Class UserExtendedAttributesEvent
+ *
+ * @package OCP\User
+ * @since 10.3.1
+ */
+class UserExtendedAttributesEvent extends Event {
+	const USER_EXTENDED_ATTRIBUTES = 'UserExtendedAttributesEvent';
+
+	/** @var array|null */
```
jvillafanez: adjust doc here

lib/private/User/User.php
```diff
@@ -555,4 +618,62 @@ public function setSearchTerms(array $terms) {
...
+			/**
+			 * Restrict the user accounts table update for the set operations as the apps
+			 * listening to the event below will be providing extended attributes of this
+			 * user.
+			 */
+			$this->allowUserAccountUpdate = false;
+			try {
+				$this->eventDispatcher->dispatch(UserExtendedAttributesEvent::USER_EXTENDED_ATTRIBUTES, $userExtendedAttributesEvent);
+			} catch (NotPermittedActionException $ex) {
+				$this->logger->logException($ex);
```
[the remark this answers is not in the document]
sharidas: Removed the DI for logger. Also removed the catch block. Reduced to try/finally block.

tests/lib/User/UserTest.php
sharidas: That was a mistake. Fixed it.

lib/public/User/UserExtendedAttributesEvent.php
sharidas: Done

sharidas:
- The `getExtendedAttributes` is now protected with the flag. So that the listeners cannot call this method, which might cause infinite loop.
- Updated the unit test for the flag to test
   - normal execution. That is after the normal execution the flag is reset to `true`.
   - With exception `NotPermittedActionException`. After this exception is thrown, verified that the value of flag is set to `true`.
   - Tested with generic exception `Exception`. After this exception is thrown, verified that the value of the flag is set to `true`.

- With the current approach the event does not reach the other listeners if the initial/first listener throws the exception. So for instance I added a test code in the comments app ( as this app is found first in the list ), and tried to set the display name. I observed that code did not reached `guests` listener. 
    - One solution I see is to catch the exception `NotPermittedActionException` in the `getExtendedAttributes()` and then just log it inside the catch block. I think for this method we can catch the `NotPermittedActionException` exception. But if any other exception comes then lets allow the exception to bubble up? Or another way is we can ask the listeners to catch the exception? Even that way it should work imho. This second solution is like completely relying on apps. Personally I am in favour of logging the exception of `NotPermittedActionException` in the `User` class and then move ahead. And for any other exceptions let the listener handle it.

tests/lib/User/UserExtendedAttributesEventTest.php
jvillafanez: this should be removed now that we have the other test

jvillafanez:
> With the current approach the event does not reach the other listeners if the initial/first listener throws the exception. So for instance I added a test code in the comments app ( as this app is found first in the list ), and tried to set the display name. I observed that code did not reached guests listener.

Looks like it isn't possible to force the event to reach all the listeners. We'd need a custom eventDispatcher (which is out of scope at the moment).
So, for the `getExtendedAttributes` method, I think the best solution is to let the exceptions (even the `NotPermittedActionException`) bubble up out of the method because it can't handle those. We'll have to rely on the listeners not to leak any exception.

tests/lib/User/UserExtendedAttributesEventTest.php
sharidas: Removed this test.

lib/private/User/User.php
```diff
@@ -555,4 +611,65 @@ public function setSearchTerms(array $terms) {
...
+	 */
+	public function getExtendedAttributes($clearCache = false) {
+		if (!$this->allowUserAccountUpdate) {
+			throw new NotPermittedActionException("Operation cannot be allowed as other apps are fetching extended attributes of this user.");
+		}
+		/**
+		 * using empty because userExtendedAttributes could either be null or empty array
+		 * or an array of attributes
+		 */
+		if ($clearCache || !isset($this->userExtendedAttributes) || [] === $this->userExtendedAttributes) {
```
jvillafanez: We don't need to retrigger the event if the `userExtendedAttributes` is an empty array.
You triggered the event, no one listened and you get an empty array. This is a valid result. There is no need to trigger the event again to get another empty array.

jvillafanez:
Hopefully last change. Please make sure the comments are updated and match the code behaviour.

PVince81:
I don't feel comfortable with this approach, it seems it doesn't fully suit.

Looking at the guests app we are setting an attribute "whitelistedAppsForGuests" which sounds like a generic setting, not something specific to a user. So having this in a user specific attribute feels wrong.

Thinking further: if the user attribute is something that defines what is accessible for a user, then maybe it fits more into a "roles" concept or "permissions" concept, and in this case the attribute would need to be called "whitelistedApps", as it applies only for this one user.

Maybe we should rather introduce the concept of "app access" into core in a more API-like way instead of just attributes. We already have the concept of "enabled apps for users" but here it's a softer one as the app needs to be loaded and its elements need to be discarded.

In a broader view, an app should have a way to find out whether the current user is supposed to see its controls or not, so having an API for "app access" or "app permissions" might be more suitable.

@DeepDiver1975 @micbar any thoughts on this ? does this fit somehow into the discussions related to roles concept ?

jvillafanez:
The problem we've noticed at the last moment is that the all the apps are required to be loaded before calling the `getExtendedAttributes` method.
For this particular case, the guests app is required to be loaded so we can know what other apps are whitelisted. In the worst case the guests app is loaded the last, we might wrongly assume that the rest of the apps can be loaded and be accessed normally by that user.

I think the idea is worth to keep because it has other uses such as properly expose the login name of the user, expose LDAP attributes for LDAP users, or other information for shibboleth or openIdConnect.

> Looking at the guests app we are setting an attribute "whitelistedAppsForGuests" which sounds like a generic setting, not something specific to a user. So having this in a user specific attribute feels wrong

The original idea was to use 2 keys: the "isGuest" in order to identify guests users, and the  "whitelistedApps" to include the list. "whitelistedAppsForGuests" was a simple way to merge both keys since the intention was to use that key for both identify a guest user and get the list of apps.

In any case, I'm starting to think this approach won't work for guests. Guests are added in the DB backend and they aren't inside a "guests" backend part of the guests app. When the guests app is disabled, guests users will still be available, but we won't be able to identify them as a guest with this approach because the app will be disabled.

PVince81:
didn't we have an event (hook) for the moment after which all apps have finished loading ?

sharidas:
> didn't we have an event (hook) for the moment after which all apps have finished loading ?

Searched the code base in core, and found this https://github.com/owncloud/core/blob/master/lib/private/legacy/app.php#L162
And this would return false https://github.com/owncloud/core/blob/master/lib/private/legacy/hook.php#L91

Other than this I couldn't find further references.

mmattel:
@phil-davis @sharidas is there any progress?

mmattel:
@jvillafanez maybe you could take a look on to bring this forward...

jvillafanez:
Looks good to me. Maybe @DeepDiver1975 can have a second look for the approach.
We might need to rebase the branch and retest.

phil-davis:
> We might need to rebase the branch and retest.

Rebased in #40257 - we can review and comment there, and if needed and good it can easily be merged from there.

phil-davis:
PR #40257 is passing now. I had to just fixup some code-style and unit test things. So someone can comment there if this is wanted, and if the code is OK.

phil-davis:
This code has been merged in PR #40257

PR #36265, opened 2019-10-10
Validate reshare permissions and attributes based on supershare

mrow4a:
Goal: do not allow privilege escalations with share attributes e.g. in Secure View resharing scenario.

There is no need to calculate reshare permission and attributes, these are already calculated by ShareProviders e.g. https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/MountProvider.php#L71-L105

- Obsoletes https://github.com/owncloud/core/pull/36208
- This improves security in the scope of e.g. Secure View (as reshare check for attributes is done)

Attributes in this logic, in reshare scenario **can only be disabled** (1->0 in bitmap), it is **not possible however to add or remove additional attributes** (add/remove bits in bitmap) **nor enabling the disabled attribute** (0->1 in bitmap).

- [x] add more unit tests coverage in ManagerTest

### DIFF
--- a/changelog/unreleased/36265
+++ b/changelog/unreleased/36265
@@ -1,0 +1,6 @@
+Change: Validate reshare permissions and attributes based on supershare
+
+This PR uniformed a way reshare permissions and attributes are enforced in a
+reshare scenario - strict subset checking has been used in both cases.
+
+https://github.com/owncloud/core/pull/36265
--- a/changelog/unreleased/36265
+++ b/changelog/unreleased/36265
@@ -1,6 +1,6 @@
 Change: Validate reshare permissions and attributes based on supershare
 
-This PR uniformed a way reshare permissions and attributes are enforced in a
-reshare scenario - strict subset checking has been used in both cases.
+This change provides a uniform way that reshare permissions and attributes are
+internally checked and enforced. There is no change to external behaviour.
 
 https://github.com/owncloud/core/pull/36265
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -306,7 +306,7 @@ class Manager implements IManager {
 			$share->setPermissions($share->getPermissions() & ~\OCP\Constants::PERMISSION_CREATE);
 		}
 
-		/**
+		/*
 		 * TODO: ideally, getPermissions should always return valid permission
 		 *       and this check should be done in Share object's setPermission method
 		 */
@@ -320,8 +320,9 @@ class Manager implements IManager {
 			throw new GenericShareException($message_t, $message_t, 400);
 		}
 
-		/** Use share node permission as default $maxPermissions*/
+		/* Use share node permission as default $maxPermissions */
 		$maxPermissions = $shareNode->getPermissions();
+
 		/*
 		 * Quick fix for #23536
 		 * Non moveable mount points do not have update and delete permissions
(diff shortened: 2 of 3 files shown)


lib/private/Share20/Manager.php
```diff
@@ -1693,7 +1675,43 @@ private function isNewShare(IShare $share) {
...
+	 * @return boolean ,true if $allowedAttributes enabled is super set of $newAttributes enabled, else false
+	 */
+	private function strictSubsetOfAttributes($allowedAttributes, $newAttributes) {
+		// if both are empty, it is strict subset
+		if ((!$allowedAttributes || empty($allowedAttributes->toArray()))
+			&& (!$newAttributes || empty($newAttributes->toArray()))) {
+			return true;
+		}
+
+		// removing completely attribute or enabling the disabled attribute is not allowed
```
karakayasemi: As far as I understand, the count of the allowed attributes and new attributes should be equal. Also, new attributes should not enable any non-existing or disabled attribute.  
Can we simplify these two `foreach` blocks with one count check and one `foreach` block like I described above? Or am I missing something?

karakayasemi (commented):
👍 LGTM. Just I wonder if we could improve the readability of attribute comparison a bit.

lib/private/Share20/Manager.php
mrow4a: @karakayasemi It is not that easy - I have even unit test for it https://github.com/owncloud/core/pull/36265/files#diff-a7c98b6bb73fd4ada066d52c96f48467R3688-R3695 :)

karakayasemi: What about this one, unit tests are passing in my local for it:
````
private function strictSubsetOfAttributes($allowedAttributes, $newAttributes) {
	// if both are empty, it is strict subset
	if ((!$allowedAttributes || empty($allowedAttributes->toArray()))
		&& (!$newAttributes || empty($newAttributes->toArray()))) {
		return true;
	}

	if (\count($allowedAttributes->toArray()) !== \count($newAttributes->toArray())) {
		return false;
	}

	// enabling a disabled attribute or completely removing an attribute is not allowed
	foreach ($newAttributes->toArray() as $newAttribute) {
		$allowedEnabled = $allowedAttributes->getAttribute($newAttribute['scope'], $newAttribute['key']);
		if (($newAttribute['enabled'] == true && $allowedEnabled !== true) || $allowedEnabled === null) {
			return false;
		}
	}

	return true;
}
````

Comment (mrow4a) [reply lib/private/Share20/Manager.php]: so you check if newly set attributes are equal in number to old and, and if they do, you ensure that all are matching or that they were not disabled. Makes sense, I will change and add a comment !

Comment (mrow4a) [reply lib/private/Share20/Manager.php]: @karakayasemi I extended the testcase and added one more OR condition to cover cornercases.

**Comment by karakayasemi** [comment]
@mrow4a we need a changelog entry before merge, then I can approve the changes.

**Comment by mrow4a** [comment]
@karakayasemi @PVince81 I improved the readability of attribute comparison as discussed, and added the changelog.

File: changelog/unreleased/36265 [hunk 2]
```diff
@@ -0,0 +1,6 @@
+Change: Validate reshare permissions and attributes based on supershare
+
+This PR uniforms a way reshare permissions and attributes are enforced in a
```
karakayasemi: Please, use past tense in this paragraph as suggested in the changelog template.
https://github.com/owncloud/core/blob/master/changelog/TEMPLATE#L6

phil-davis:
@micbar this has a changelog entry.
Is it planned for release?

changelog/unreleased/36265
mrow4a: Thanks, I will adjust

karakayasemi (approved):

mrow4a:
@micbar should I rebase this PR to master or this is on halt? This adds a security layer to secure-view, not allowing to have privilege escalations with reshare.

micbar:
@mrow4a Yes, please rebase.

micbar:
@phil-davis @HanaGemela @jnweiger This is a 10.4 candidate.

phil-davis:
We have quite a good set of acceptance test scenarios for sharing permission combinations. Given that CI is currently green here, then it is "a good thing" from my server QA point-of-view.
Let's see after @mrow4a resolves the conflict and rebases.

phil-davis:
I adjusted the changelog text to reflect the fact that this is essentially an internal refactor, it is not changing external behaviour.

When CI goes green again it can be merged.

phil-davis:
@mrow4a see rich documents PR https://github.com/owncloud/richdocuments/pull/310

This core PR did have the effect of preventing Secure View resharing, which seems to have been the goal.

Please have a look at that PR and confirm that the acceptance test scenario behaviour changes are what you are expecting.

PR #36261, opened 2019-10-09
Changelog for Symfony 3.4.32

phil-davis:
## Description
Add a changelog entry for the Symfony 3.4.32 dependency bumps done by dependabot yesterday.

Symfony release notes: https://symfony.com/blog/symfony-3-4-32-released

Related PRs:
https://github.com/owncloud/core/pull/36244
https://github.com/owncloud/core/pull/36245
https://github.com/owncloud/core/pull/36246
https://github.com/owncloud/core/pull/36247
https://github.com/owncloud/core/pull/36248

## How Has This Been Tested?
N/A

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](changelog/TEMPLATE)

PR #36267, opened 2019-10-11
Update changelog post sprint 52

PVince81:
@micbar FYI


phil-davis (commented):
@micbar is changing to https://github.com/owncloud/core/blob/master/changelog/README.md
and there are already some entries in https://github.com/owncloud/core/tree/master/changelog/unreleased
and PR #36261 ...

PVince81:
ah, I didn't know the process changed. Then I'll leave this up to @micbar to update the changelog when he's back next week.

as far as I understand, developers should create those new files with each PRs but if they don't then the release manager needs to do this after the fact manually

PVince81:
maybe we need a script or an extension for gen_changelog.sh that automatically creates such files then...

phil-davis:
And currently people have been naming the file in `changelog/unreleased` using the PR number. It is not possible to do that in the first submission of a PR, because locally you do not know the PR number until after you push the branch and submit the PR. It would be easier to use the issue number (assuming there is an issue)

phil-davis:
@PVince81 out-of-date, close?

PR #36281, opened 2019-10-15
Avoid unnecessary "Avatar not found" logs by initializing ViewOnlyPlugin for only files

karakayasemi:
## Description
https://github.com/owncloud/core/blob/b9ff4c93e051c94adfb301545098ae627e52ef76/apps/dav/lib/DAV/ViewOnlyPlugin.php#L86
This line is throwing NotFound exception when a node not found and, this action is logged as a warning. By registering ViewOnlyPlugin for only files problem resolved. 

Also, minor bugs in `Server.php` that were previously overlooked were fixed.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3332

## Motivation and Context
The server clients are asking user avatars with DAV requests. When an avatar not found, the server producing lots of warning logs because of the ViewOnlyPlugin. By registering ViewOnlyPlugin for only files problem resolved. 

## How Has This Been Tested?
Manually with following scenario;
- Create a user but not upload any avatar.
- Login with the desktop client, no warning log should appear in owncloud.log.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Server.php
+++ b/apps/dav/lib/Server.php
@@ -137,7 +137,6 @@ class Server {
 			$this->server->addPlugin(new PublicFilesPlugin());
 			$authPlugin->addBackend(new PublicSharingAuth($this->server, OC::$server->getShareManager()));
 			$this->server->addPlugin(new PublicLinkEventsPlugin(\OC::$server->getEventDispatcher()));
-			$this->server->addPlugin(new PublicFilesPlugin());
 		}
 		$authPlugin->addBackend(new PublicAuth());
 		$this->server->addPlugin($authPlugin);
@@ -218,10 +217,11 @@ class Server {
 			OC::$server->getLazyRootFolder()
 		));
 
-		// Allow view-only plugin for webdav requests
-		$this->server->addPlugin(new ViewOnlyPlugin(
-			OC::$server->getLogger()
-		));
+		if ($this->isRequestForSubtree(['files', 'trash-bin', 'public-files'])) {
+			$this->server->addPlugin(new ViewOnlyPlugin(
+				OC::$server->getLogger()
+			));
+		}
 
 		if (BrowserErrorPagePlugin::isBrowserRequest($request)) {
 			$this->server->addPlugin(new BrowserErrorPagePlugin());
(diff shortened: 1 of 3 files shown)


jvillafanez:
I'm slightly against reducing the log level: for me, using a debug level usually means that you expect the code to go through that point; it's used for tracing purposes.
I guess the code expects that people ask for an existing node, so if people ask for a non-existing node, this is something to report (warning level looks good). So why are there avatar requests ending up here? People not having an avatar seems a common enough case to consider it.

I think this is the easiest solution, but I also think we should understand why avatar requests are ending up here. I'm pretty sure this isn't something considered.

VicDeo:
@jvillafanez because `ViewOnlyPlugin` is registered for any file and it uses exactly `checkViewOnly` method to detect whether the access should be blocked for any GET request.

It needs a storage to detect the permissions, and storage is taken from a Node, thus the first action in `checkViewOnly` method  is `getNodeForPath` so it fails right away for non-existing nodes.

https://github.com/owncloud/core/blob/c69e2c1e3f1b90208059922c35d50c1bc560eb17/apps/dav/lib/DAV/ViewOnlyPlugin.php#L71-L116


As for me the approach is valid unless you could offer a better solution.

VicDeo:
@karakayasemi CI failed due to timeout, please rebase

karakayasemi:
Like @VicDeo explain the plugin throws NotFound exception from its first line and I could not find any other solution.

Also, as I stated in the description, the other 
DAV plugins like `CorePlugin` are also facing with this exception for non-existing paths. `ViewOnlyPlugin` aims to work before `CorePlugin`. 
https://github.com/owncloud/core/blob/3108bb496929ab06f52c93fe54e8bdd895c3233f/apps/dav/lib/DAV/ViewOnlyPlugin.php#L66-L68

After `ViewOnlyPlugin`, `CorePlugin` runs and it does not even catch the exception. When NotFound exception threw, it handled at a higher level and the server returns NotFound response by breaking request.

The other solution could be not catching the exception like `CorePlugin`. In this way, the server returns a 404 response without any log.

jvillafanez:
> @jvillafanez because ViewOnlyPlugin has a quite high prio, it is registered for any file

Any chance to attach it only inside the "files" tree? I'm not sure if it's possible...

jvillafanez (approved):
Assuming we aren't missing anything

micbar (approved):
👍

PR #36290, opened 2019-10-18
[For 10.4] Drop php 7.0 bump phpunit7

phil-davis:
## Description
This PR is a place to put things that are waiting to go into `stable10` but that require PHP7.1+ and so cannot be merged until PHP7.0 support can be dropped. (from PR #36289 )
Plus moving from `phpunit6` to `phpunit7`

1) From PR #33255 "Add mariadb 10.3 + mysql 8 to Drone, remove 10.2 workarounds' (5 commits)
2) Drop PHP 7.0

- #28043 Increase version to PHP7.1 (the remaining commits from that PR that relate to PHP7.1 and thus have never been applied yet)
- Set min PHP version to 7.1
- PHP 7.1+ is required
- "[Workaround] .... until sebastianbergmann/phpunit-mock-objects#402 is merged and released" https://github.com/owncloud/core/pull/28043/commits/407a70d90dd2692fab8f79cbc28014cff37a1995 (I think this will be needed as long as we are on phpunit6, because that `phpunit-mock-objects` change never got into a release with dependencies for phpunit6)
- (other commits from #28043 are already back in `stable10` which is nowadays `master`)
- #32065 2nd commit "Remove our own sqlite autoincrement hack" (needed with `doctrine/dbal` version bump that comes when composer PHP 7.1 happens)
- Adjust drone for PHP7.1 like master (remaining drone adjustments)
- Use is_iterable for PHP7.1 (to match more modern test code from archive/master)
- Update to PHP 7.1 in travis - in case we use it to run UI  tests on other browsers
- Put back empty line in db_structure.xml that was removed by PR 34783 (resolves a minor diff between old stable10 and old master)
- Remove special PHP7.0 checks from Makefile
3) support PostgreSQL 10.3

- Make sure to have at least doctrine/dbal 2.8 (see issue #33187 )
- Dismiss OCPostgreSqlPlatform as issues it addresses were fixed in doctrine/dbal since 2.6.4 (picked from #31059 )

4) put back some test code that was for PHP 7.1-only (ref PR: #35958 )

5) add commits to update `phpunit` to V7.5.17

## Related Issue
#29483 MariaDB
#33187 PostgreSQL 10

## Motivation and Context
Gather all the changes required for "drop PHP7.0" and related support for newer MariaDB and PostgreSQL versions... that were in the "old" master, and have them ready for "new" master.

PR #36291, opened 2019-10-18
Adjust phpunit withConsecutive calls

phil-davis:
## Description
`phpunit7` reports test fails like:
```
There was 1 error:

1) Test\Share20\ManagerTest::testDeleteChildren
PHPUnit\Framework\InvalidParameterGroupException: Parameter group #0 must be an array or Traversable, got object

/home/phil/git/owncloud/core/tests/lib/Share20/ManagerTest.php:565
```

Some unit test calls to `withConsecutive` are passing just a list of ordinary strings. But in the case where there is just a single parameter to send on each call, that single parameter should still be an array that has just 1 entry, the single parameter. `phpunit6` does not complain about this, but `phpunit7` will.

## Motivation and Context
May as well apply little unit test code fixes like this to the existing `phpunit6` test code.

## How Has This Been Tested?
CI

## Types of changes
- [x] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/tests/API/ShareesTest.php
+++ b/apps/files_sharing/tests/API/ShareesTest.php
@@ -607,8 +607,8 @@ class ShareesTest extends TestCase {
 				$this->groupManager->expects($this->exactly(2))
 					->method('getUserGroupIds')
 					->withConsecutive(
-						$user,
-						$singleUser
+						[$user],
+						[$singleUser]
 					)
 					->willReturn($groupResponse);
 			} else {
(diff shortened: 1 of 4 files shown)


jvillafanez (approved):
I guess the CI will complain if we've missed something

phil-davis:
> I guess the CI will complain if we've missed something

In `phpunit6` it won't complain, but it does in `phpunit7` - I had to do these changes in https://github.com/owncloud/core/pull/36290/commits/bce1713c6db23385b33b458843df752f305ac4ec to keep `phpunit7` happy.

PR #36324, opened 2019-10-23
[For 10.4] User group share expiration

lordelix:
## Description

Excerpt from the feature file:

* As a user I want to set an expiration date for collaborators in order to stay in control of access permissions by giving time-based access instead of granting access forever or until I manually remove the collaborator.
* As a user I want to automatically limit my responsibility for a guest account. Guests are bound to being a collaborator. If they are not a collaborator of any resource, they can't log in anymore.
* As an admin I want to set a default expiration date without enforcement to give a recommendation to users and help them maintain control over their resources. Still, I don't want to force or annoy users as some may have reasons to collaborate on resources for a longer or unrestricted amount of time, e.g., in projects.
* As an admin I have to follow corporate policies which dictate that data exchange between employees or also with externals may only take place for a certain maximum amount of time to prevent data leaks by human mistake or negligence.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/2001

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Screenshots:
Anmerkung 2019-11-11 133034

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


lib/public/Share/IShareProvider.php
```diff
@@ -31,6 +31,25 @@
...
+	 * here because those rules will be checked at API level, not here.
+	 * If a share provider doesn't support any of these capabilities, it should
+	 * either store the value as null or don't store it, and return a null value
+	 * as value when the share is gotten.
+	 */
+	/**
+	 * Store when the share expires. An expired share needs to be automatically
+	 * deleted. If a share won't expire, a "null" value will be used
+	 */
+	const CAPABILITY_STORE_EXPIRATION = 'share expiration';
```
PVince81: use camel case maybe as this is not a text to display ?

settings/js/admin.js
```diff
@@ -73,6 +73,29 @@ $(document).ready(function(){
 		$("#setDefaultExpireDate").toggleClass('hidden', !this.checked);
 	});
 
+
+	$('#shareapiExpireAfterNDaysUserShare').change(function() {
+		var value = $(this).val();
+		if (isNaN(value) || (parseInt(value) <= 0) || parseInt(value).toString() !== value) {
```
PVince81: always use `parseInt(value, 10)` to avoid octal parsing in case there is a leading zero, see https://stackoverflow.com/questions/850341/how-do-i-work-around-javascripts-parseint-octal-behavior

PVince81 (commented):
Had a quick look and it looks fine so far. Might need a deeper review to check corner cases etc.

The part I didn't is where does the expiration field gets added to the UI ?

felixheidecke:
> The part I didn't is where does the expiration field gets added to the UI ?

I opened @jvillafanez branch so we can keep this PR open for conversation. UI Code is in development.

felixheidecke:
@jvillafanez 

- [x] "Set default expiration date for user shares" doesn't set a date
- [ ] "Enforce as maximum expiration date"

**Doesn't set the date** 👎 

- [x] "Set default expiration date for user shares" doesn't set a date
- [x] "Enforce as maximum expiration date"

**Set's the date** 👍

jvillafanez:
> :heavy_check_mark: "Set default expiration date for user shares"
> :x: "Enforce as maximum expiration date"

It should match the behaviour of link expiration in previous ownCloud versions. I think that the default expiration date is just a hint for the frontend, but it won't be set unless explicitly requested by sending that date in the "create" request.
In the case the expiration is enforced, we set the default as maximum because we don't want to set it as "undefined" / "null", which is what happen if the expiration isn't enforced.

That should explain the behaviour. I agree this is kind of weird, but as said it's the same behaviour the link expiration had before, so I think it's better to keep the same behaviour for now.

felixheidecke:
@jvillafanez in order to update/set the expiration date, the API needs to accept `expiration` as a parameter.

Currently, I get "Wrong or no update parameter given" if I'm trying to PUT `expiration: 2019-11-14T00:00:00+01:00` --> `/ocs/v2.php/apps/files_sharing/api/v1/shares/123?format=json`

PVince81:
as discussed, need to fix parameter presence validation in "updateShare" to properly check for "expireDate" additionally to "permissions"

jvillafanez:
> Currently, I get "Wrong or no update parameter given" if I'm trying to PUT expiration: 2019-11-14T00:00:00+01:00 --> /ocs/v2.php/apps/files_sharing/api/v1/shares/123?format=json

@felixheidecke  did you check with `2019-11-13 00:00:00`? You probably just need the date because you can fill the "00:00:00" easily before sending the info.
It should be compatible with whatever is being used for the link shares.

jvillafanez:
```
curl -X POST -u admin:Password --data 'shareType=0&shareWith=user1&permissions=19&path=/pic3.jpg&expireDate=2019-11-15 10:00:00' 'http://10.0.2.27:6080/ocs/v2.php/apps/files_sharing/api/v1/shares?format=json'
```
that works, although the time part is ignored. `expireDate=2019-11-15` is enough

jvillafanez:
> as discussed, need to fix parameter presence validation in "updateShare" to properly check for "expireDate" additionally to "permissions"

I fail to see where the problem is. Backend changes are mostly moving things around, so the same validation we had for links is being applied to both users and groups. Any use case?

felixheidecke:
I found a little problem with the defaultExpireationDate approach: @PVince81 @jvillafanez 

With links, we have two different states. We are either the add-modal with new links, or the edit-modal with existing links.

**While adding a link**, the default expiration date is displayed in the form where the user would be able to remove it if it's not wanted.

**While editing a link** the expiration date field is empty, as the value is null.

**With user/group shares**, there is only edit. So there is no place to display the recommended default expiration date.

I currently have no solution for this :-/

PVince81:
@felixheidecke unrelated note: I had this branch checked out and noticed that we are not using the same date picker element. Did you intend to switch to the same picker element later on ? (I did not see a TODO checkbox for it, that's why I'm asking)

jvillafanez:
Removing some checks around https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/Controller/Share20OcsController.php#L755 and https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/Controller/Share20OcsController.php#L844 should do the trick I think

felixheidecke:
> @PVince81  Did you intend to switch to the same picker element later on?

In my current (unpushed) code, I do use the OC default datepicker. It'll be there soon ;-)

felixheidecke:
@jvillafanez works great, thanks 👍 

I talked with @pmaier1 and we do have a solution for this problem right here --> https://github.com/owncloud/core/pull/36324#issuecomment-551485101

We need to change the behavior for the user/group `defaultExpirationDate` a bit.
As we do not have a confirmation dialog für user/group shares, we do net to set the `expirationDate` in advance. The user then has the remove the unwanted expiration date instead of setting it.

thus solving the problem :-D

@jvillafanez can you do that little change? Just ping me if you have any questions.

jvillafanez:
> We need to change the behavior for the user/group defaultExpirationDate a bit.
As we do not have a confirmation dialog für user/group shares, we need to set the expirationDate in advance. The user then has the option to remove the unwanted expiration date instead of setting it.

There are some capabilities in place for that, although I think they aren't put in the config model. I'm not sure why there are capabilities copied into the config model, so I didn't touch that part.
(https://github.com/owncloud/core/pull/36324/files#diff-ce11fa52acdfc132e8854ee17e90c873R114-R126)

apps/files_sharing/css/sharetabview.css
```diff
@@ -76,6 +76,22 @@
...
+}
+
+#shareWithList li .removeExpiration {
+	text-indent: -9999px;
+	border: 0 none;
+	transform: translate(calc(-100% - 2px), -1px);
+	background-color: #F8F8F8;
+	background-position: center;
+	background-repeat: no-repeat;
+	background-image: url('../../../core/img/actions/close.svg');
```
PVince81: don't we have already a reusable CSS class for this icon ?

core/js/tests/specs/sharedialogviewSpec.js
```diff
@@ -970,4 +970,26 @@ describe('OC.Share.ShareDialogView', function() {
...
+		beforeEach(function() {
+			configModel.set({
+				isDefaultExpireDateUserEnabled: true
+			});
+			shareModel.set({
+				shares: [{
+					id: 100,
+					item_source: 123,
+					permissions: 31,
+					share_type: OC.Share.SHARE_TYPE_USER,
```
PVince81: add a test for the group enforced ?

also, does it make sense to also test the other possible flags ?

lib/public/Share/IShareProvider.php
PVince81: ping

settings/js/admin.js
PVince81: ping

core/js/tests/specs/sharedialogviewSpec.js
felixheidecke: Group enforced happens on the server side only.

I added some tests for USER and GROUPS in here: https://github.com/owncloud/core/pull/36324/files?file-filters%5B%5D=.js#diff-eb4e94032220ff30f56e13ac40197070R1144

felixheidecke:
@jvillafanez there are some PHP unit test failing :-|
Unfortunately I can't solve these.

apps/files_sharing/css/sharetabview.css
felixheidecke: I didn't find any. Yet, we could use this class in other occurences ... don't you think? I could make it a more general class.

felixheidecke:
@PVince81 @jvillafanez I did the requested changes as well as gave some feedback to some points.
The php related CI errors need to be adressed.

PVince81 (approved):
:+1: fine by me if the tests pass now

jvillafanez:
I think we should retest everything. I've done some changes in the code for the tests, so maybe I've reverted some behaviour. Mainly for https://github.com/owncloud/core/pull/36324/commits/b2e40c76a4ee376a27e3d15daeb090eb2ad60d98#diff-faddd9984dfaf09a2b85d8dc4867207aR754

skshetry (changes requested):
This is missing tests. Assigning myself for it.

core/js/sharedialogshareelistview.js
```diff
@@ -384,6 +402,41 @@
...
+
+			$el.datepicker({
+				minDate: "+1d",
+				dateFormat : 'dd-mm-yy',
+				onSelect : function() {
+					self._onExpirationChange(el)
+				}
+			});
+
+			if (this.configModel.isDefaultExpireDateUserEnforced())
```
phil-davis: Do group shares also come through here? Does there need to be more logic for the "group" case also?

core/js/shareconfigmodel.js
```diff
@@ -92,7 +98,56 @@
...
+		 * @returns {boolean}
+		 */
+		isDefaultExpireDateUserEnabled: function() {
+			return this.get('isDefaultExpireDateUserEnabled')
+		},
+
+		/**
+		 * @returns {boolean}
+		 */
+		isDefaultExpireDateUserEnforced: function() {
```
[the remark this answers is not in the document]
skshetry: yes, the expiration field does not appear for group share, if user share expiration is disabled. cc @phil-davis @felixheidecke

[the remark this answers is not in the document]
phil-davis: @micbar this is the problem mentioned in standup.
IMO, since this is "only" JS, it is probably just in the UI (the UI does not correctly understand that the group expire date should be available to enter). I suspect that API requests will work OK.

phil-davis:
To all: this PR has 17 commits and is quite out-of-date from master. There are multiple people working on acceptance tests and there will be bugs to fix also. But rebasing and squashing this will cause risk from people maybe re-pushing and old local branch that they have...

PR #36531 has the 17 commits from here rebased and squashed on top of master - that branch is called `feature/user_group_share_expiration`

When doing fixes or test code, please first checkout `feature/user_group_share_expiration` then checkout a new branch from there and make the changes in that new branch. Then we can carefully cherry-pick good new things into `feature/user_group_share_expiration` as they are ready.

micbar:
@felixheidecke Please investigate.

felixheidecke:
Fixed via https://github.com/owncloud/core/pull/36535

individual-it:
closing as its superseded by #36531

Issue #36266, opened 2019-10-11
TransferOwnershipTest fails with files_primary_s3 scality multibucket


phil-davis:
### Steps to reproduce
1. run core acceptance tests with files_primary_s3 enabled with scality multibucket config from files_primary_s3 `tests/drone/scality.multibucket.config.php`

### Expected behaviour
Tests pass

### Actual behaviour
A new test is failing:
```
There was 1 failure:

1) OCA\Files\Tests\Command\TransferOwnershipTest::testTransferAllFiles
Failed asserting that actual size 3 matches expected size 0.

/var/www/owncloud/server/apps/files/tests/Command/TransferOwnershipTest.php:180
```

e.g. https://drone.owncloud.com/owncloud/files_primary_s3/1219/11/7

### Server configuration
CI with current core master and files_primary_s3 master

phil-davis:
The new test was added in PR #36222 by @karakayasemi 
That PR was only merged yesterday, so last night was the first night that the test was run in the context of files_primary_s3 object storage.
The test passes for single-bucket scality, but fails in the multi-bucket case. I suspect that in the multi-bucket case the 2 users are on different storages, so maybe an ownership transfer like that does not work (the files would have to be moved between storages?).
Because this is a new test added to core, I don't know if it would have passed or not previously.

phil-davis:
I can get the test fail locally from `files_primary_s3` with:
```
drone exec --pipeline phpunit-scality-multibucket-php7.2-sqlite > unit-output.txt
```
and look in `unit-output.txt`
The  fail is in line 2176 of the output:
```
[phpunit-tests:129] ...SSSSSSSSSSSSSSSSSSSF....................................  7257 / 10000 ( 72%)
```
[unit-output.txt](https://github.com/owncloud/core/files/3715601/unit-output.txt)

Maybe the scality log output before that point has something useful?

davitol:
Feature seems to work for me with Scality single-bucket:

**Background** Using 10.0.3.RC1 and files_primary_s3 and Scality

1 - Create a ```user```
2 - Log in with ```user``` and create some file/folders
3 - Via CLI run ```occ files:transfer-ownership user1 admin```

Result: Files are transferred without problems to admin user

No logs are shown in owncloud.log

phil-davis:
Yes, the "Scality with single bucket" CI passes.
It is multi-bucket that [can] fail.

jvillafanez:
confirmed problem regarding shares:
```
[...]

PR #36326, opened 2019-10-23
Objectstore transfer

jvillafanez:
## Description
Fix `files:transfer-ownership` command for multibucket objectstores (checked with files_primary_s3 app against a scality server, with multibucket configuration)

**NOTE** Only files are moved between buckets if needed. Thumbnails and versions remain in the old bucket. The owncloud's filecache also keeps the thumbnails and versions in the same storages (to be investigated further)

## Related Issue
https://github.com/owncloud/enterprise/issues/3520

## Motivation and Context
transfer-ownership was causing problems with the shares because the fileid of the transferred files was changing. This PR aims to prevent the fileid change so the shares are still bound to valid ids (the files are transferred anyway)

## How Has This Been Tested?
1. With user1, create "file1.txt", "f1/file2.txt" and "f1/f2/file3.txt" (with the required folders)
2. user1 shares "f1/f2" and "f1/file2.txt" with the admin user
3. run `occ files:transfer-ownership user1 user2` (user2 must exists and be valid)

The transfer happens. The admin can still access to the previously shared files. Shares are visible for user2 (user2 owns the shares)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/tests/Command/TransferOwnershipTest.php
+++ b/apps/files/tests/Command/TransferOwnershipTest.php
@@ -163,6 +163,15 @@ class TransferOwnershipTest extends TestCase {
 			->setShareType(Share::SHARE_TYPE_USER)
 			->setPermissions(19);
 		$this->shareManager->createShare($share);
+
+		$subFolder = $userFolder->get('transfer/sub_folder');
+		$share = $this->shareManager->newShare();
+		$share->setNode($subFolder)
+			->setSharedBy('source-user')
+			->setSharedWith('share-receiver')
+			->setShareType(Share::SHARE_TYPE_USER)
+			->setPermissions(19);
+		$this->shareManager->createShare($share);
 	}
 
 	public function testTransferAllFiles() {
@@ -178,13 +187,13 @@ class TransferOwnershipTest extends TestCase {
 		$sourceShares = $this->shareManager->getSharesBy($this->sourceUser->getUID(), Share::SHARE_TYPE_USER);
 		$targetShares = $this->shareManager->getSharesBy($this->targetUser->getUID(), Share::SHARE_TYPE_USER);
 		$this->assertCount(0, $sourceShares);
-		$this->assertCount(3, $targetShares);
+		$this->assertCount(4, $targetShares);
 	}
 
 	public function folderPathProvider() {
 		return [
-			['transfer', 1, 2],
-			['transfer/sub_folder', 2, 1]
+			['transfer', 1, 3],
+			['transfer/sub_folder', 2, 2]
 		];
 	}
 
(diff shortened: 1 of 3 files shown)


jvillafanez:
To be double-checked:
* Versions are lost. The copy between buckets only copies the file (object), but it doesn't copy the versions along with the file. It's unclear if the old object will be completely removed in the old bucket, or if it will remain accessible if it has versions.
* Thumbnails remain in the old bucket and in the filecache. This will use space
* Trashbin hasn't been tested yet. The behaviour is unknown at the moment

jvillafanez:
We'll have to decide what to do with the versions. They'll likely need special handling because they aren't moved or copied by default, and you'll need to copy the versions one by one (and delete them also one by one if you want to move them).
I haven't seen anything in the s3 api to perform this operations in the server side, so this will kill the performance because we'll need to connect several times to the s3 server per file.

jvillafanez:
I'm not sure if we can add unittest here. There are a couple of problems:
* the constructor calls some methods of the object.
    * We could try to mock the object without calling the constructor, but there is a risk that any function we test relies on those methods to have been called.
    * Calling the constructor implies calling those methods, but they could fail if the dependencies aren't fully mocked. This means we can't mock correctly the class.
* The `moveFromStorage` relies on the parent's `copyFromStorage` method (at least for the scenario we want to test). That method has calls to `fclose` that can't be mocked.

We might need to pull off the new unittests.

jvillafanez:
Removed the unittests from the list of changes due to problems with the code. We'll rely on the integration tests to verify this works properly using multibucket.
In addition, added an additional test case to check with folder shares (whose fileid must also remain unchanged)

jvillafanez:
I don't think I can raise the code coverage. Unittests problems are described in https://github.com/owncloud/core/pull/36326#issuecomment-546349340 and the acceptance tests already tests what is expected.

phil-davis:
CI passes in `files_primary_s3` when run against core `objectstore_transfer` branch. Great stuff.
So this is "a good thing" from a QA/CI point of view.

micbar:
> We'll have to decide what to do with the versions. They'll likely need special handling because they aren't moved or copied by default, and you'll need to copy the versions one by one (and delete them also one by one if you want to move them).
> I haven't seen anything in the s3 api to perform this operations in the server side, so this will kill the performance because we'll need to connect several times to the s3 server per file.

@pmaier @cdamken We need to skip versions here. ok for you?

phil-davis:
Feedback/review/approval needed here. Getting this sorted will unblock #36266 and let backlogged `files_primary_s3` PRs be rebased/merged.
@cdamken @pmaier1 - ^^

pmaier1:
> @pmaier @cdamken We need to skip versions here. ok for you?

As the docs say (https://doc.owncloud.com/server/admin_manual/configuration/server/occ_command.html#the-filestransfer-ownership-command), the command never transfers file versions. So this should be fine for objectstore as well.

cdamken:
If it fixed the problem from https://github.com/owncloud/enterprise/issues/3520 I'm ok with it ;)

phil-davis:
Sadly a different test fails in the multi-bucket environment: https://github.com/owncloud/core/issues/36266#issuecomment-553944879

Issue #36322, opened 2019-10-23
CI is failing


phil-davis:
### Steps to reproduce
1. Run drone CI

### Expected behaviour
Pass

### Actual behaviour
1) phpunit `LargeFileHelper` fails e.g. https://drone.owncloud.com/owncloud/core/20935/13/9
```
There were 2 failures:

1) Test\LargeFileHelperGetFileSizeTest::testGetFileSizeViaCurl with data set #0 ('/drone/src/tests/data/lorem.txt', 446)
Failed asserting that null is identical to 446.

/drone/src/tests/lib/LargeFileHelperGetFileSizeTest.php:53

2) Test\LargeFileHelperGetFileSizeTest::testGetFileSizeViaCurl with data set #1 ('/drone/src/tests/data/strängé...2).txt', 446)
Failed asserting that null is identical to 446.

/drone/src/tests/lib/LargeFileHelperGetFileSizeTest.php:53
```
on various PHP 7.0 and 7.1 unit test runs.

And fails on 7.2 https://drone.owncloud.com/owncloud/core/20935/21/9 and 7.3 https://drone.owncloud.com/owncloud/core/20935/23/9 but the drone job is green!?!?!?

2) samba and SMB-windows unit tests fail
https://drone.owncloud.com/owncloud/core/20935/26/10
```
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

Runtime:       PHPDBG 7.1.32-1+ubuntu18.04.1+deb.sury.org+1
Configuration: /drone/src/tests/phpunit-autotest-external.xml

mkdir failed on directory /var/cache/samba: Permission denied
mkdir failed on directory /var/cache/samba: Permission denied
mkdir failed on directory /var/cache/samba: Permission denied
Fmkdir failed on directory /var/cache/samba: Permission denied
...
There were 85 failures:

1) OCA\Files_External\Tests\Storage\SmbTest::testRenameWithSpaces
Failed asserting that false is true.

/drone/src/apps/files_external/tests/Storage/SmbTest.php:53
...
```

### Server configuration
core master

phil-davis:
PR #36318 demonstrates the fails.

PR #36319 fixes the `LargeFileHelper` fails.

phil-davis:
PR #36423 fixes the samba tests

PR #36328, opened 2019-10-24
[WIP] Enable samba tests

phil-davis:
For issue #36322 

- [x] enable samba tests again
- [x] cut down the drone CI test suite (so we get answers quicker)
- [ ] try to fix the failures


phil-davis:
See #36422 #36423

It seems that `'domain'=>'WORKGROUP'` is needed in the samba config now.

So all this trying to find the `sudo apt install` things to install was never going to succeed. Just `mkdir /var/cache/samba` was enough in the first place.

Issue #29759, opened 2017-12-05
[Feature Request] Check if file names restrictions apply for a given mount (filesystem)


mmattel:
**Problem:**
If you copy/move files from one storage type to another, it might be possible that file name restrictions apply. This can happen when you copy/move files originating in a *unix derivate to a smb mount. Because the mount is a folder, you can not see the underlaying filesystem and its restrictions. The same when you create a new file via the browser with restricted patterns. Currently you just get a deny or a strange not related error message/behaviour but without a explaining reason.

**Example1:**
``mount_unix:`` ``LPT1.txt`` copy to ``mount_smb:``
``LPT1.txt`` is a valid filename in *unix but not in Windows.

**Example2:**
``mount_windows:`` ``a&b.txt`` copy to ``mount_unix:``
``a&b.txt`` is a valid filename in Windows but you will get problems in *unix as it is not escaped.

**Impact:**
Because you do not know the reason for not beeing able to copy/move/create, and assume the the user has not the deep knowledge about the different restrictions of filesystems file namings, this can become very annoying and confusing.

**Proposal:**
Add an identifyer to the mount settings where you can preset different types of file name restrictions, like Windows, *unix ect (extendable if a finer granularity is needed)
Having this identifyer, a file create/copy/move action passes the file name restriction part and tells the calling function respectively user about if it found a restriction match.
On the http response type, a possibility would be to use code 422 (Unprocessable Entity) see https://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm
The identifyer should be part of the webdav propery so the client can use this info.
With the proper http code, core and apps can differntiate responses like reject, rename, ect and include a reason why.

The client side is a different issue and is addressed in https://github.com/owncloud/client/issues/6202

**what-characters-are-forbidden-in-windows-and-linux-directory-names**
https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

1.) The forbidden printable ASCII characters are:
- Linux/Unix: ``/ (forward slash)`` 
- Windows: ``< (less than)`` ``> (greater than)`` ``: (colon - sometimes works, but is actually NTFS Alternate Data Streams)`` ``" (double quote)`` ``/ (forward slash)`` ``\ (backslash)`` ``| (vertical bar or pipe)`` ``? (question mark)`` ``* (asterisk)`` 

2.) Non-printable characters. If your data comes from a source that would permit non-printable characters then there is more to check for.
[...]

PR #36331, opened 2019-10-24
Allow regex in blacklisted dirs and files

tdaget:
## Description
I need to support regex in blacklist functionality for files and dirs. Need to set blacklist on the server (want to enforce corporate IT policies).
`excluded_directories` and `blacklisted_files` variables already exist in the parameter file but only work on exact folder name and filename.

In lib/private/Files/Filesystem.php : envolve `isFileBlacklisted` function in order to evaluate regex match for each entry declared in $excluded_directories and $blacklisted_files config parameters respectively for folder name and filename.
In config/config/sample.php : For $excluded_directories and $blacklisted_files variables, add chars to be regex compatible.

Regex syntax can be tested with [https://www.phpliveregex.com](https://www.phpliveregex.com)

## Motivation and Context
This change was initiated as in our structure several people sync there local Outlook pst file (despite IT recommendations). This generate a lot (a very lot) of file's change on local disk so made a lot of network traffic between Owncloud desktop client and server.

## How Has This Been Tested?
Currently, this patch is deployed on our test platform without any drawbacks. Community comments are welcome ...
Note: Regex match is CPU consuming, hope this will not raise major problem on a production platform. Stress tests must be performed.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added


lib/private/Files/Filesystem.php
```diff
@@ -655,20 +655,26 @@ public static function isForbiddenFileOrDir($FileOrDir, $ed = []) {
...
-		$blacklist = \array_map('trim', $blacklist);
-		$blacklist = \array_map('strtolower', $blacklist);
-		$match = \array_intersect($path_parts, $blacklist);
-		if ($match) {
-			return true;
-		}
+                if ($excluded) {
+                        $excluded = \array_map('trim', $excluded);
+                        $excluded = \array_map('strtolower', $excluded);
+                        foreach($path_parts as $path_part) {
```
mmattel: * You could do a check here (and below) if `$excluded` contains a to be defined "forbidden" character (out of a list, for hints see: https://github.com/owncloud/core/issues/29759 plus usual regex characters, blacklist or whitelist what ever better fits) which would indicate a present regex and then branch between the old and your new regex querey. This would greatly improve performance as it continues using the old way with `array_intersect` and uses `preg_match` in the loop only if necessary.
* The query characters used for indicating a regex must be described in the config.sample file above.

* Add an example like:
```
hugo   // does not contain a regex character
will match the first item of the path component only:
hugo
hugo.test
testhugo

(?:hugo)  // contains regex
will match all items of the path component:
hugo
hugo.test
testhugo
```

config/config.sample.php
```diff
@@ -1219,8 +1221,8 @@
  */
 'excluded_directories' =>
 	array (
-		'.snapshot',
-		'~snapshot',
+		'^\.snapshot$',
```
mmattel: * I propose that you leave the original examples as written and add additional but different more fancy once to the list (blacklisted and excluded). Helps to make things more clear that you can use the old and the new (regex) way.
* In addition you must add a descriptive text that regex is allowed.
Add a `WARNING: ` message about the possible performance impact like:
```
WARNING: COMPLEX REGEX MAY SIGNIFICANT IMPACT YOUR SCANNING PERFORMANCE
```
The term warning + the colon + blank afterwards must be exacly written as in the example above because this has a doc rendering background.

lib/private/Files/Filesystem.php
```diff
@@ -655,20 +655,26 @@ public static function isForbiddenFileOrDir($FileOrDir, $ed = []) {
...
+                                foreach($excluded as $blackitem) {
+                                        if(preg_match('/'.$blackitem.'/i', $path_part)) {
```
mmattel: Please use: `preg_match` --> `\preg_match` (and below)

See: https://doc.owncloud.com/server/10.3/developer_manual/core/code-standard.html

mmattel:
Your `Fixes` list from above is not correct as sync excludes are not addressed by this PR.
These are a different topic. Pls remove them

lib/private/Files/Filesystem.php
tdaget: Thanks for the doc's reference, correction is done.

config/config.sample.php
tdaget: Can't leave previous examples as they must be regex compliant (per example the dot in '.snapshot'). But more information and example might be welcome.
Warning added. Thanks

lib/private/Files/Filesystem.php
tdaget: Advise 2 and 3: done.
Advise 1: don't know how to implement them ...

tdaget:
> Your `Fixes` list from above is not correct as sync excludes are not addressed by this PR.
> These are a different topic. Pls remove them

Done

lib/private/Files/Filesystem.php
```diff
@@ -658,16 +658,30 @@ public static function isForbiddenFileOrDir($FileOrDir, $ed = []) {
 		if ($excluded) {
 			$excluded = \array_map('trim', $excluded);
 			$excluded = \array_map('strtolower', $excluded);
-			$match = \array_intersect($path_parts, $excluded);
-			if ($match) {
-				return true;
+			foreach($excluded as $blackitem) {
+				if(@\preg_match($blackitem, null) === false) {
+					\OC::$server->getLogger()->error('Malformed exclude regex: '.$blackitem.' - Check excluded_directories variable in config file.', ['app' => __CLASS__]);
```
mmattel: 1. Great that there is an error message now.
The message itelf only shows that something bad happened but gives no clue what it was.
It would be good to **add** a better hint about the cause. Please see the example below:
Based on:
https://www.php.net/manual/de/function.preg-last-error.php and https://stackoverflow.com/questions/4440626/how-can-i-validate-regex/12941133#12941133
I allowed myself to add/rename some stuff...
```
function is_preg_error()
{
    $errors = array(
        PREG_NO_ERROR               => 'Code 0 : No errors',
        PREG_INTERNAL_ERROR         => 'Code 1 : Internal PCRE error, most likely a syntax error',
        PREG_BACKTRACK_LIMIT_ERROR  => 'Code 2 : Backtrack limit was exhausted',
        PREG_RECURSION_LIMIT_ERROR  => 'Code 3 : Recursion limit was exhausted',
        PREG_BAD_UTF8_ERROR         => 'Code 4 : The offset did not correspond to the begin of a valid UTF-8 code point',
        PREG_BAD_UTF8_OFFSET_ERROR  => 'Code 5 : Malformed UTF-8 data',
        PREG_JIT_STACKLIMIT_ERROR   => 'Code 6 : Just-in-time compiler stack limit reached',
    );

    return $errors[\preg_last_error()];
}
```

lib/private/Files/Filesystem.php
```diff
@@ -658,16 +658,30 @@ public static function isForbiddenFileOrDir($FileOrDir, $ed = []) {
...
+				} else {
+					foreach($path_parts as $path_part) {
+						if(\preg_match('/'.$blackitem.'/i', $path_part)) {
+							return true;
```
mmattel: You may want to add a check if regex processing was ok:
`\preg_last_error() !== PREG_NO_ERROR`
and log an error in case using the function I described above.

mmattel:
1. There are still coding style errors, pls see https://doc.owncloud.com/server/10.3/developer_manual/core/code-standard.html
and the build failing details from drone above

2. I am still not convinced hard switching from `array_intersect` done in one shot to `preg_match` double looping by default. I highly recommend as written in my comments above to at least do some checking for common regex characters and decide based on the result the way of checking. I know that there is no 100% identifyer, but I refrenced a hint in my comment. Things are easier for Windows backends as they have more excluding characters in filenames. But this is a matter of a proper description. This was the reason for keeping the old examples and adding new regex once.

lib/private/Files/Filesystem.php
tdaget: Done

lib/private/Files/Filesystem.php
tdaget: Your code is added.
I put it directly in Filesystem.php class, don't know where to put it in a utility class ...

tdaget:
@mmattel 
I agree, switching from `array_intersect` to `preg_match` might induce high CPU consumption.
I think we can consider following solutions:
1- create a new `excluded_directories` and a new `blacklisted_files` dedicated to regex match.
    I am scared of configuration difficulties and misunderstanding by users, I prefer to avoid this situation
2- allow a mix of strict filename and regex in these configuration variables, for each test is this is a string, if yes search for a direct match, if not try regex match. Double work ... And I don't know how to identify a simple string and invalid regex
3- consider that strict filename matching (previous way of working) is cover by regex use with an acceptable CPU cost
    The way I'm currently considering. Plus it permits to check regex validity systematically.

You propose to search for common regex characters, maybe not so reliable as these characters (for many of them and not least) can be used in filenames.

But all advice is welcome :-/

tdaget:
An idea to speed up the regex check: no need to check all `blacklisted_files` to all part of 'path to file' array, we need to match only the filename part.
Question (as I'm not an expert PHP developer): in the path_parts array variable, is the last item is ALWAYS (the always is very important here) the filename itself? In that case, we can update code with -drop foreach line 709, and replace `if (\preg_match('/'.$blackitem.'/i', $path_part)) {` by `if (\preg_match('/'.$blackitem.'/i', end($path_parts))) {` line 710.

mmattel:
Hope you are fine with all my comments... :smile:
1. You need to rebase as there was a change in the drone setup that needs to be populated into your PR so drone can process properly and turns into green 
2. `foreach ($excluded as $blackitem)` and `foreach ($blacklist as $blackitem)` :
only an optical thing, but to avoid confusion, please use a more neutral name for `$blackitem`
3. The code for checking `\preg_last_error()` should also run after each `foreach ($path_parts`.
This because you can run in all the other errors that syntax which is checked before. Recommending that the check and the error is put into a function which is called with an (optional) error text as you have written it already. This would reduce number of lines and makes the code more readble.
4. I get your point regarding regex / non regex confusion. To keep the positives of both variants, what about the idea to implement two new config parameters like `blacklisted_files_regex` and `excluded_directories_regex`? You then can run the code based on which one is set. Yes, you can set both but it gives admins the ability deciding which way is best for them. If both are set, you must run both and combine the result with `or`. Please a good descriptive text if you go for that approach.
5. Thanks for looking at `integrity.excluded.files`, but pls turn that back, as there is another PR already in review which takes care on that (and more).
6. Regarding `blacklisted_files`:
-Reminder, this code **must** always run, but not necessarily a empty regex version... (see idea above)
-The last element of `$path_parts` can either be the directoryname or the filename if a filename exists. You can improve the code as you thought and do the same for the current way if you want. Thanks for spotting that.

tdaget (commented):
Typo corrected in tests/data/app/invalid-info.xml
Add regular match for blacklisted folders and blacklisted files

apps/federatedfilesharing/templates/settings-personal-general.php
```diff
@@ -44,7 +44,7 @@ class='js-gs-share social-gnu'>
 				Facebook
 			</button>
 			<button class="social-googleplus pop-up"
-				data-url='https://plus.google.com/share?url=<?php p(\urlencode($_['reference'])); ?>'/>
+				data-url='https://plus.google.com/share?url=<?php p(\urlencode($_['reference'])); ?>'>
```
mmattel: Great for finding bugs / creating improvements.
As this one has nothing to do with regex, it must be put into a own PR.

config/config.sample.php
```diff
@@ -1201,32 +1201,73 @@
...
+ */
+'blacklisted_files_regex' => array (
+//	'.*\.pst$',		// disallow *.pst (MS Outlook) file
+//	'.*dummy.*',	// disallow file with name contaning the string 'dummy'
+//	'^sample.*',	// disallow file with name beginning by 'sample'
+),
+
+
+/**
+ * Blacklist specific directory names and disallow scanning, creating and renaming
```
mmattel: `Blacklist specific directory` --> `Exclude specific directory`

config/config.sample.php
```diff
@@ -1201,32 +1201,73 @@
...
  */
-'excluded_directories' =>
-	array (
+'excluded_directories' => array (
+	'.snapshot',	// exclude '.snapshot' folder
+	'~snapshot',	// exclude '~snapshot' folder
+),
+
+/**
+ * Blacklist specific directory names based on regex expression. Scanning, creating and renaming
```
mmattel: `Blacklist specific directory` --> `Exclude specific directory`

config/config.sample.php
```diff
@@ -1201,32 +1201,73 @@
...
+ * WARNING: COMPLEX REGEX MAY SIGNIFICANT IMPACT YOUR SCANNING PERFORMANCE
+ * WARNING: String case is systematicaly lowered
+ */
+'excluded_directories_regex' => array (
+//	'.*backup.*',	// exclude folder with name containing the string 'backup'
+//	'Thomas.*',		// exclude folder with name beginning by 'Thomas'
+),
+
+/**
+ * Exclude specific directory names and disallow scanning, creating and renaming
```
mmattel: This block seems to be a leftover and forgotten to be deleted...

config/config.sample.php
```diff
@@ -1201,32 +1201,73 @@
...
-	),
+),
+
+
 /**
  * Exclude files from the integrity checker command
  */
 'integrity.excluded.files' =>
-	array (
+	array(
```
mmattel: As already stated, pls no change here as we already have an open PR which covers this issue (plus other stuff).

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,91 @@ public static function isIgnoredDir($dir) {
 	}
 
 	/**
-	* Check if the directory path / file name contains a Blacklisted or Excluded name
```
mmattel: Pls keep this explanation text from the content point of view as it describes what is meant with each kind.
It is totally ok to move it down to line 688.

settings/templates/help.php
```diff
@@ -45,11 +45,12 @@
 	<?php
 	} ?>
 
-	<li>
-		<a href="https://owncloud.com/subscriptions/" target="_blank" rel="noreferrer">
-			<?php p($l->t('Commercial support')); ?> ↗
-		</a>
-	</li>
+		<li>
```
mmattel: As stated above, this has nothing to do with regex.
Pls create an own PR for fixing other issues than regex.

tests/data/app/invalid-info.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<info
+<info>
```
mmattel: As stated above, this has nothing to do with regex.
Pls create an own PR for fixing other issues than regex.

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,91 @@ public static function isIgnoredDir($dir) {
...
+								.self::is_preg_error(),
+								['app' => __CLASS__]);
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Manage regex blacklist against Folder of File.
```
mmattel: Remove the word `blacklist`

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,91 @@ public static function isIgnoredDir($dir) {
...
+	 * @param blacklist Blacklist regex array as defined in Config file
+	 * @param path Folder of File complete path contained in an array
+	 * @return boolean folder of file forbudden status
+	 */
+	private static function blacklistRegexAgainstFolderOrFile($blacklist, $path) {
```
mmattel: Please use a neutral name instead of `$blacklist`

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,91 @@ public static function isIgnoredDir($dir) {
...
+		}
+
+		return true;
+	}
+
+	/**
+	 * Manage regex blacklist against Folder of File.
+	 * @param blacklist Blacklist regex array as defined in Config file
+	 * @param path Folder of File complete path contained in an array
+	 * @return boolean folder of file forbudden status
```
mmattel: `folder of file forbudden status` --> `folder of file regex status`

mmattel:
We are going into the right direction :smile:
I see that you are a big fan of the term blacklist...
Using it as a common name everywhere is a bad idea. The original naming was made by design.
- Have function and variable names based on the usecase which further reflects config.sample.
(means, use blacklist... where you run code for blacklist and excluded... where you run code for excluded)
- `blacklistRegexAgainstFolderOrFile`--> `checkRegexAgainstFolderOrFile`
- `regex_validy_check($regexToBeChecked)` --> `regex_validy_check($regexToBeChecked, $message="")`
Add a second parameter and handover the message via the caller (like `Check excluded_directories variable in config file`). This message differs either from blacklist or excluded. Rename the first part of the message `Exclude regex error:` --> `Regex error:`
- Based on the example `blacklist_folders` and `blacklist_folders_regex` (same for excluded), as stated above, you **must** run them if set and combine the results with `or` for a final return.

config/config.sample.php
tdaget: Corrected

config/config.sample.php
tdaget: Corrected

config/config.sample.php
tdaget: You're alright
Corrected

config/config.sample.php
tdaget: Can't see this point in my code ... 'll check after correcting all other points

lib/private/Files/Filesystem.php
tdaget: Word dropped

lib/private/Files/Filesystem.php
tdaget: Corrected

lib/private/Files/Filesystem.php
tdaget: Changed

tdaget:
> We are going into the right direction 
> I see that you are a big fan of the term blacklist...
> Using it as a common name everywhere is a bad idea. The original naming was made by design.
> 
> * Have function and variable names based on the usecase which further reflects config.sample.
>   (means, use blacklist... where you run code for blacklist and excluded... where you run code for excluded)
> * `blacklistRegexAgainstFolderOrFile`--> `checkRegexAgainstFolderOrFile`
Changed

> * `regex_validy_check($regexToBeChecked)` --> `regex_validy_check($regexToBeChecked, $message="")`
>   Add a second parameter and handover the message via the caller (like `Check excluded_directories variable in config file`). This message differs either from blacklist or excluded. Rename the first part of the message `Exclude regex error:` --> `Regex error:`
Good idea
Corrected

> * Based on the example `blacklist_folders` and `blacklist_folders_regex` (same for excluded), as stated above, you **must** run them if set and combine the results with `or` for a final return.
Sorry, Can't understand this point.
As I modified the isForbiddenFileOrDir method the sequence of tests (exclude then exclude_regex then blacklist then blacklist_regex) do by construction an implicit OR ==> first test make a 'return true' and kill all other tests. And each test is done only if directive is set in config file.
In my understanding this permit to skip all non-mandatory work.
Why do you want to combine tests ? I'll make code less readable ...

> * Please rebase to make CI happy and green
I'm working on it, step after step, but lessons are hard to remember ...

apps/federatedfilesharing/templates/settings-personal-general.php
tdaget: Reverted

settings/templates/help.php
tdaget: Reverted

tests/data/app/invalid-info.xml
tdaget: Reverted

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,92 @@ public static function isIgnoredDir($dir) {
...
+				}
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	* Check if the directory / file name is blacklisted in config.php.
+	* $ed and the query with can be redesigned if filesystem.php will get
```
mmattel: Please use at least parts the original description (written in one of my comments already)
I have made a suggestion and added some words to blacklist to extend the usecase description
```
	/**
	* Check if the directory path / file name contains a Blacklisted or Excluded name
	* config.php parameter arrays can contain file names to be blacklisted or directory names to be excluded
	* Blacklist ... files that may harm the owncloud environment like a foreign .htaccess file or other unwanted files
	* Excluded  ... directories that are excluded from beeing further processed, like snapshot directories
	* $ed and the query with can be redesigned if filesystem.php will get a constructor where it is then possible 
	* to define the excluded directory names for unit tests
```

lib/private/Files/Filesystem.php
```diff
@@ -655,20 +719,32 @@ public static function isForbiddenFileOrDir($FileOrDir, $ed = []) {
...
+			}
+		}
+		if ($blacklist_files) {
+			if (\array_intersect($blacklist_files, \end($path_parts))) {
+				return true;
+			}
 		}
+		if ($blacklist_files_regex) {
+			if (self::checkRegexAgainstFolderOrFile($blacklist_files_regex, \end($path_parts),
+													"Check blacklisted_files_regex variable in config file")) {
```
mmattel: The intend is a bit too big :smile:

config/config.sample.php
```diff
@@ -1217,23 +1234,38 @@
...
-		'~snapshot',
-	),
+'excluded_directories' => array (
+	'.snapshot',	// exclude '.snapshot' folder
+	'~snapshot',	// exclude '~snapshot' folder
+),
+
+/**
+ * Exclude specific directory names based on regex expression. Scanning, creating and renaming
+ * is disabled when folder match one of the regex.
```
mmattel: Just put this into two lines for a better readbility:
```
 * Exclude specific directory names based on regex expression.
 * Scanning, creating and renaming is disabled when folder matches one of the regex.
```

config/config.sample.php
```diff
@@ -1217,23 +1234,38 @@
...
+ * Use of excluded_directories variable have precedence (evaluated before) on excluded_directories_regex variable.
```
mmattel: `Use of...` --> `INFO: Use of ...`
The term info or warnig has a special meaning here.
It will be rendered in the documentation in a special optical way.

config/config.sample.php
mmattel: pls re-check. still there...

mmattel:
Very good! only a view things to take care on :smile:

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,92 @@ public static function isIgnoredDir($dir) {
...
+	 * @return boolean folder of file regex status
+	 */
+	private static function checkRegexAgainstFolderOrFile($regexList, $path, $message="") {
+		foreach ($regexList as $item) {                               // foreach given regex
+			\preg_match($item, null);                                 // regex validy check
+			if (self::regexValidyCheck($item, $message)) {            // check if regex error occur
+
+				foreach ($path as $path_part) {                       // foreach folder or file in given path
+					if (\preg_match('/'.$item.'/i', $path_part)       // regex match item
+							&& self::regexValidyCheck($item)) {       // check if regex error occur
```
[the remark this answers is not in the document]
tdaget: Good idea, added.

lib/private/Files/Filesystem.php
tdaget: Sorry, dropped this block by mistake
Corrected

lib/private/Files/Filesystem.php
```diff
@@ -655,20 +719,32 @@ public static function isForbiddenFileOrDir($FileOrDir, $ed = []) {
...
 			}
 		}
-		$blacklist = \array_map('trim', $blacklist);
-		$blacklist = \array_map('strtolower', $blacklist);
-		$match = \array_intersect($path_parts, $blacklist);
-		if ($match) {
-			return true;
+		if ($exclude_folders_regex) {
+			if (self::checkRegexAgainstFolderOrFile($exclude_folders_regex, $path_parts,
+													"Check excluded_directories_regex variable in config file")) {
```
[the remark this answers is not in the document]
tdaget: Corrected

lib/private/Files/Filesystem.php
tdaget: Corrected

config/config.sample.php
tdaget: Corrected

config/config.sample.php
tdaget: Added

config/config.sample.php
```diff
@@ -1201,12 +1201,29 @@
...
+ */
+'blacklisted_files' => array (
+	'.htaccess',	// disallow .htaccess file
+),
+
 /**
- * Blacklist a specific file or files and disallow the upload of files
- * with this name. `.htaccess` is blocked by default.
+ * Blacklist filenames based on regex expression.
+ * Use of blacklisted_files variable have precedence (evaluated before) on blacklisted_files_regex variable.
```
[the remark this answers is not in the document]
tdaget: Added

config/config.sample.php
```diff
@@ -1201,12 +1201,29 @@
...
+ * WARNING: USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
+ */
+'blacklisted_files' => array (
+	'.htaccess',	// disallow .htaccess file
+),
+
 /**
- * Blacklist a specific file or files and disallow the upload of files
- * with this name. `.htaccess` is blocked by default.
+ * Blacklist filenames based on regex expression.
```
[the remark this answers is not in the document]
tdaget: Corrected

mmattel:
- `integrity.excluded.files` seems still not reverted, maybe just overseen.
- in config.sample, in the new regex examples, no need to put the array elements as remarks (ok for the description). 
- maybe a good idea to squash the commits now

lib/private/Files/Filesystem.php
```diff
@@ -625,28 +625,102 @@ public static function isIgnoredDir($dir) {
...
+	}
+
+	/**
+	 * Regex validy check.
+	 * @param string $regexToBeChecked regex to be checked
+	 * @param string $message message given in log in case of invalid regex
+	 * @return boolean
+	 */
+	private static function regexValidyCheck($regexToBeChecked, $message="") {
+		if (self::is_preg_error() !== PREG_NO_ERROR) {
```
mmattel: Shouldnt this be `\preg_last_error()` as `is_preg_error` returns a string ?

tdaget:
> * in config.sample, in the new regex examples, no need to put the array elements as remarks (ok for the description).

No need? Maybe but as sysadmin I know examples are always welcome, especially with regex

If you want absolutely to remove these lines of example maybe we can put them in comment block?

lib/private/Files/Filesystem.php
tdaget: Right

tdaget:
> * maybe a good idea to squash the commits now

Need some help for this (don't understand the "squash commits" ??)

mmattel:
I meant: `//    '.*\.pst$',     // di...` —> `    '.*\.pst$',     // di...`
It is removing the first two `//`

mmattel:
Example squashing the last three commits:
(https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git)
```
git reset --soft HEAD~3 &&
git commit
git push -f
````
At the moment you have 32 commits... (just see the number on top).
If you squash all 32 commits and push them, you see finally only one.
Note: the number of commits to squash must not exceed the existing number of commits of this PR. You do squashing when the work is close to final / or final already. This makes checking easier.

mmattel:
on top, pls set the checkmark that you have added unit tests

tdaget:
> on top, pls set the checkmark that you have added unit tests

Done

tdaget:
> I meant: `// '.*\.pst$', // di...` —> ` '.*\.pst$', // di...`
> It is removing the first two `//`

Remove the two `//` mean activating this regex by default (as everybody bases his config.php on the sample file). I'm afraid of confusion and questions for a lot of people.

At most I can consider doing this change if the `*.pst` regex as Outlook files raises a lot of undesirable network traffic and Owncloud activities.

mmattel:
> activating this regex by default

No worry about that one. The name says **sample** and there is a description in the documentation not to just copying stuff but using it wisely... And only a limited set config options are used during installation. Anything else must be used by actively adding/configuring it to the running config file.

mmattel:
I just tried to help a bit and wanted to clone your branch but have seen that you did not used a branch.
Now this is somehow a bad thing because you will hardly come into sync with master and its ongoing changes without having a branch. But this is needed because of changes running CI. My proposal therefore would be the following:
- Create a branch from actual master and make/copy all the changes you did in this PR.
- Push it and make a new PR from that push.
- Close this PR

Big benefit: Clean, short and open for changes in master (easy rebaseable) and CI will turn green
If you cant do that, I am willing to help, but can do this earliest late evening today. Just let me know.

Great work!

mmattel:
Closing in favour of https://github.com/owncloud/core/pull/36360.
I have improved your changes. Tested and runs like a charm :smile:
@tdaget please do any changes if necessary in #36360

tdaget:
> Closing in favour of #36360.
> I have improved your changes. Tested and runs like a charm 
> @tdaget please do any changes if necessary in #36360

Thanks for doing this.
I must admit that I was a little lost in the management of branches and PRs.
I'm still not comfortable with git / github ...

mmattel:
You could use smartgit (https://www.syntevo.com/smartgit/download/) available for various OS.
Makes life a lot easier when you start using git

tdaget:
Thanks for our advise.
I start to use github with the official 'github desktop' under linux
not so bad once we understand the basic principles ...

Thanks for your help in this PR
I've close all my branch/PR created in a desperate discover of tools. I created another branch named `Typo-corrections` in order to work on all small typo found during this PR.

PR #36337, opened 2019-10-25
Disallow sharing and deleting of share_folder and it's parents

### LINKED ISSUE #36241: share_folder can be shared
Server: 10.3.0RC1

Steps to recreate:
1) Add `share_folder` into `config.php`
2) User1 shares a file with User2
3) User2 deletes the file from their account
4) User2 shares `share_folder` 

Actual result: `share_folder` can be shared. Any new files which are added to that folder as newly received shares are not shared to the collaborators. But new files can be added via public links. This is confusing

Expected result: Sharing of `share_folder` is not allowed

### LINKED ISSUE #36252: Deleted folder is not restored when deletion is not allowed
Server: 10.3.0 RC1
Client: 2.6.0 RC1 

Steps to recreate:
1) Add the below line into `config.php` on the server (From now on the incoming shares will be shown in that folder)
`'share_folder' => '/ReceivedShares'`
2) User1 shares a resource to User2
3) User2 deletes `ReceivedShares` folder on the client

Actual result: 403 Forbidden reply from the server but the folder has been deleted on the client
Expected result: Folder has been restored on the client 

Screenshot 2019-10-01 at 14 34 23

Similar logic as in https://github.com/owncloud/client/issues/7490 When server forbids something, the action on the client should be reverted 

Workaround available in 2.5 client (not working in 2.6 anymore)
1) Unselect that folder in selective sync list
2) Apply 
3) Select it again
4) Apply

karakayasemi:
## Description
Remove share and delete permission for `share_folder` . Since View's root is not persistent, we can not rely on the relative path of share folder when restricting share_folder actions. Because of this, the share folder is deletable via desktop client.
I changed the path comparison to absolute path, and remove  "share and delete" permission of share_folder.

## Related Issue
- Fixes #36241
- Fixes #36252

## Motivation and Context
The server shouldn't allow sharing and deleting the entire `share_folder`

## How Has This Been Tested?
Scenario 1:
1. Add share_folder into config.php
2. User1 shares a file with User2
3. User2 deletes the file from their account
4. User2 shares share_folder
5. User2 tries to delete share_folder via desktop client

### Actual result
- share_folder can be shared. Any new files which are added to that folder as newly received shares are not shared with the collaborators. But new files can be added via public links.
- share_folder deletable via desktop client

### Expected result
Sharing and deleting of share_folder should not be allowed in anyway.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -8,7 +8,7 @@
  * @author Jesús Macias <jmacias@solidgear.es>
  * @author Joas Schilling <coding@schilljs.com>
  * @author Jörn Friedrich Dreyer <jfd@butonic.de>
- * @author karakayasemi <karakayasemi@itu.edu.tr>
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
  * @author Klaas Freitag <freitag@owncloud.com>
  * @author Lukas Reschke <lukas@statuscode.ch>
  * @author Luke Policinski <lpolicinski@gmail.com>
@@ -353,9 +353,7 @@ class View {
 	public function rmdir($path) {
 		return $this->emittingCall(function () use (&$path) {
 			if ($path !== '') {
-				$shareFolder = \trim($this->config->getSystemValue('share_folder', '/'), '/');
-				$trimmedPath = \trim($path, '/');
-				if ((\strpos("$shareFolder/", "$trimmedPath/") === 0)) {
+				if ($this->isShareFolderOrShareFolderParent($path)) {
 					Util::writeLog("core", "The folder $path could not be deleted as it is configured as share_folder in config.", Util::WARN);
 					return false;
 				}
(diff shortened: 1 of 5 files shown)


VicDeo:
Looks familiar somehow... :roll_eyes:

lib/private/Files/View.php
```diff
@@ -380,6 +378,24 @@ public function rmdir($path) {
...
+	 * @param $path - path relative to the files folder
+	 *
+	 * @return bool
+	 */
+	protected function isShareFolderOrShareFolderParent($path) {
+		$shareFolder = \trim($this->config->getSystemValue('share_folder', '/'), '/');
+		if ($shareFolder === '') {
+			return false;
+		}
+		$user = $user = \OC_User::getUser();
```
phil-davis:
```suggestion
		$user = \OC_User::getUser();
```

lib/private/Files/View.php
```diff
@@ -380,6 +378,24 @@ public function rmdir($path) {
...
+		$shareFolderAbsolutePath = "/$user/files/$shareFolder";
+		$trimmedAbsolutePath = $this->getAbsolutePath(\trim($path, '/'));
+		return $shareFolderAbsolutePath === $trimmedAbsolutePath || \strpos($shareFolderAbsolutePath, "$trimmedAbsolutePath/") === 0;
```
phil-davis:
```suggestion
		return $shareFolderAbsolutePath === $trimmedAbsolutePath || \strpos("$shareFolderAbsolutePath/", "$trimmedAbsolutePath/") === 0;
```
without this, I think that if `share_folder` is `Shares` and someone also creates their own folder called `SharesMyLocalExtraFolder` then they will be prevented from deleting `SharesMyLocalExtraFolder`

phil-davis: Now I see that this does not matter. The first `===` test handles the case when the action is on the actual share folder. So the 2nd test is only needed when the `$trimmedAbsolutePath` is a parent of the share folder. And in that case it does not matter if `$shareFolderAbsolutePath` has a trailing `/` or not.
I tried a few combinations, and it works OK (and CI is passing)

VicDeo: @phil-davis trailing slash matters. Otherwise it won't be possible to delete 
`/sharedfiles`  when  the share_folder is set to `/shared`
We have unit tests for that btw :)
https://github.com/owncloud/core/blob/master/tests/lib/Files/ViewTest.php#L2726

VicDeo (approved):
Looks good

phil-davis (approved):

micbar:
@cdamken Fixed.

PR #36338, opened 2019-10-25
Allow core to whitelist apps for a user

sharidas:
This changeset allows core to whitelist apps for
a user. This way core gets controll over the apps
which a user can access or not. But this data needs
to be provided either by the app or the caller of
specific method in the IRegulateAppAcces to set.

## Description

In this PR, I have tried to grab the information about access of apps for a user. The idea here is to have the data available for core. So lets say for example the SettingsManager resides in core. And SettingsManager cannot know what are the apps whitelisted for a user, unless the data is shared with core. So any app(s) which requires to share this data to core, should use this API. So that core has this information.

The data is saved in the array where user id is the key and the apps are the list. So we get the details per user.

## Related Issue

- Fixes <issue_link>

## Motivation and Context

The whitelisted apps details should be available to core. 

## How Has This Been Tested?

- Enable guests and customgroups app
- Create a guest user
- Now login as guest user and navigate to personal settings page
- customgroups ( by default ) will not appear in the list.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Migrations/Version20191030124642.php
+++ b/core/Migrations/Version20191030124642.php
@@ -1,0 +1,89 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Types\Type;
+use OCP\Migration\ISchemaMigration;
+
+/**
+ * Create a new table appaccess.
+ * This table will have 4 columns
+ * 1. id -> primary key -> with autoincrement
+ * 2. accessor -> which will accept userid and groupid -> cannot be null
+ * 3. accessor-identifier -> which will be either 0 or 1 -> 0 for user and 1 for group
+ * 4. apps -> strings of appid which are separated by commas eg: files, shares etc
+ * The accessor-identifier is the one which will determine if accessor is a user
+ * or a group.
+ */
+ class Version20191030124642 implements ISchemaMigration {
+ 	public function changeSchema(Schema $schema, array $options) {
+ 		$prefix = $options['tablePrefix'];
+
+ 		if (!$schema->hasTable("${prefix}appaccess")) {
+ 			$appAccessTable = $schema->createTable("${prefix}appaccess");
+
+ 			$appAccessTable->addColumn(
+				'id',
+				Type::BIGINT,
+				[
+					'notnull' => true,
+					'autoincrement' => 1,
+					'unsigned' => true
+				]
+			);
+
+ 			$appAccessTable->addColumn(
+				'accessor',
+				Type::STRING,
+				[
+					'length' => 64,
+					'default' => '',
+					'notnull' => true,
+
+				]
+			);
+
+ 			$appAccessTable->addColumn(
+				'accessoridentifier',
+				Type::SMALLINT,
+				[
+					'notnull' => true,
+					'default' => 0,
+					'unsigned' => true,
+				]
+			);
+
+ 			$appAccessTable->addColumn(
+				'apps',
+				Type::TEXT,
+				[
+					'notnull' => false,
+				]
+			);
+
+ 			$appAccessTable->setPrimaryKey(['id']);
+
+ 			$appAccessTable->addIndex(['accessor', 'accessoridentifier'], 'usergroupwith_whitelistedapps');
+ 		}
+ 	}
+ }
--- a/lib/private/App/AppAccessService.php
+++ b/lib/private/App/AppAccessService.php
@@ -1,0 +1,328 @@
+<?php
+/**
+ * @author Sujith Haridasan <sharidasan@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\App;
+
+use Doctrine\DBAL\DBALException;
+use OCP\App\IAppAccessService;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\IDBConnection;
+use OCP\IGroup;
+use OCP\IGroupManager;
+use OCP\ILogger;
+use OCP\IUser;
+
+class AppAccessService implements IAppAccessService {
+	/** @var IDBConnection */
+	private $connection;
+
+	/** @var IGroupManager */
+	private $groupManager;
+
+	/** @var ILogger */
+	private $logger;
+
+	/**
+	 * AppAccessService constructor.
+	 * In this constructor we set the connection and groupManager to null. Else this
+	 * would throw error during oc installation. Once installation is done, we get
+	 * the real value fetched from the Server.
+	 *
+	 * @param ILogger $logger
+	 * @param IDBConnection $connection
+	 * @param IGroupManager $groupManager
+	 * @since 10.3.1
+	 */
+	public function __construct(ILogger $logger, IDBConnection $connection, IGroupManager $groupManager) {
+		$this->logger = $logger;
+		$this->connection = $connection;
+		$this->groupManager = $groupManager;
+	}
+
+	/**
+	 * Set whitelisted apps for a group
+	 * In this method we set the whitelisted apps for a group. The appaccess table
+	 * will be set with list of apps. The apps are list of strings. If the apps
+	 * are successfully set the method returns true. Else false is returned.
+	 * Before executing this function kindly check if apps exist for the group
+	 * using one of the get methods in this class.
+	 * We use upsert in this method so that the apps would be inserted if there
+	 * is no entry in the table, else it would update it.
+	 *
+	 * @param IGroup $group
+	 * @param array $apps
+	 * @return bool true if whitelisted apps are set else false is returned
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForGroup(IGroup $group, $apps) {
+		try {
+			$this->connection->upsert(
+				'*PREFIX*appaccess',
+				[
+					'accessor' => $group->getGID(),
+					'accessoridentifier' => 1,
+					'apps' => \implode(',', $apps),
+				],
+				[
+					'accessor',
+					'accessoridentifier',
+				]
+			);
+		} catch (DBALException $ex) {
+			$this->logger->logException($ex);
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
+	 * Set the whitelisted apps for user.
+	 * This method helps apps to store the information related to apps which can
+	 * be accessed by user. The advantage of using this method is it helps core
+	 * to grab access to the apps which are accessible to user. This method returns
+	 * true if the apps list is set successfully. Else false is returned. Remember
+	 * this method can overwrite any existing data. Say for example an app decided
+	 * to update the whitelists accessible for user, it would be done.
+	 * We use upsert in this method so that the apps would be inserted if there
+	 * is no entry in the table, else it would update it.
+	 *
+	 * @param IUser $user
+	 * @param string[] $apps
+	 * @return bool
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForUser(IUser $user, $apps) {
+		try {
+			$this->connection->upsert(
+				'*PREFIX*appaccess',
+				[
+					'accessor' => $user->getUID(),
+					'accessoridentifier' => 0,
+					'apps' => \implode(',', $apps),
+				],
+				[
+					'accessor',
+					'accessoridentifier',
+				]
+			);
+		} catch (DBALException $ex) {
+			$this->logger->logException($ex);
+			return false;
+		}
+		return true;
+	}
+
+	/**
+	 * Get the whitelisted apps for the user
+	 * This is a special method which computes the apps which a user has access to.
+	 * Meaning it computes the whitelisted apps of user and the whitelisted apps
+	 * of the group(s), in which the user is a member of. And then merges the result.
+	 * The final result would have the total apps accessible for the user.
+	 * If the user is not found in the table, this method would return false.
+	 * The significance of a false in return from this method means: the user has
+	 * access to the enabled apps.
+	 * If an array of whitelisted apps is returned then it means: the user has
+	 * access to only those apps. The returned list is of the type key-value pair.
+	 * Where Key is the name of the app and Value is also the name of the app.
+	 * This method can also return an empty list. This means the user does not
+	 * have access to any apps. Unless core provides certain apps which should
+	 * be available to all users. For example files app is one such app.
+	 *
+	 * @param IUser $user
+	 * @return bool|mixed, array when whitelisted apps are available. The array can be empty. If there is no entry in table false is returned.
+	 * @since 10.3.1
+	 *
+	 */
+	public function getComputedWhitelistedAppsForUser(IUser $user) {
+		/**
+		 * Check if the user is available in different groups. If so grab the apps
+		 * user has access to these groups.
+		 */
+		$groupObjects = $this->groupManager->getUserGroups($user);
+		$groupIds = [];
+		foreach ($groupObjects as $groupObject) {
+			$groupIds[] = $groupObject->getGID();
+		}
+
+		/**
+		 * Sql Query is
+		 * SELECT apps FROM appaccess
+		 * WHERE (
+		 * (accessor = 'user') AND (accessoridentifier = '0')) OR
+		 * ((accessor IN (listofgroups)) AND (accessoridentifier = '1'))
+		 */
+		$whitelistedApps = [];
+		$userHasAppsWhitelisted = false;
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select('apps')
+			->from('appaccess')
+			->where($qb->expr()->orX(
+				$qb->expr()->andX(
+					$qb->expr()->eq('accessor', $qb->expr()->literal($user->getUID())),
+					$qb->expr()->eq('accessoridentifier', $qb->expr()->literal('0'))
+				),
+				$qb->expr()->andX(
+					$qb->expr()->in('accessor', $qb->createPositionalParameter($groupIds, IQueryBuilder::PARAM_STR_ARRAY)),
+					$qb->expr()->eq('accessoridentifier', $qb->expr()->literal('1'))
+				)
+			));
+		$statement = $qb->execute();
+		while ($row = $statement->fetch()) {
+			$userHasAppsWhitelisted = true;
+			foreach (\explode(',', $row['apps']) as $app) {
+				if ($app !== '') {
+					$whitelistedApps[$app] = $app;
+				}
+			}
+		}
+
+		$statement->closeCursor();
+
+		if ($userHasAppsWhitelisted) {
+			return $whitelistedApps;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Get whitelisted apps for user excluding the results of groups
+	 * This method fetches the whitelisted apps for the user. It does not considers
+	 * if the user is available in other groups and the apps whitelisted in other groups.
+	 *
+	 * @param IUser $user
+	 * @return bool|mixed
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppsForUser(IUser $user) {
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select(['apps'])
+			->from('appaccess')
+			->where($qb->expr()->eq('accessor', $qb->expr()->literal($user->getUID())))
+			->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal(0)));
+		$statement = $qb->execute();
+
+		/**
+		 * We expect only one row for the user. Hence we go ahead with fetch, instead.
+		 */
+		$result = $statement->fetch();
+		$statement->closeCursor();
+
+		if ($result && \array_key_exists('apps', $result)) {
+			if ($result['apps'] === '' || $result['apps'] === null) {
+				return  [];
+			}
+			return \explode(',', $result['apps']);
+		}
+
+		return false;
+	}
+
+	/**
+	 * Get the whitelisted apps for group
+	 * If the whitelisted apps are set for a group then it fetches from the array
+	 * appsAvailableForGroups and returns the array. Else it returns false.
+	 *
+	 * @param IGroup $group
+	 * @return bool|array get the array if whitelisted apps exist for it else false.
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppsForGroup(IGroup $group) {
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select(['apps'])
+			->from('appaccess')
+			->where($qb->expr()->eq('accessor', $qb->expr()->literal($group->getGID())))
+			->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal('1')));
+		$statement = $qb->execute();
+
+		/**
+		 * We expect single row for a group. Hence we use fetch() method to extract
+		 * the result.
+		 */
+		$result = $statement->fetch();
+		$statement->closeCursor();
+
+		if ($result && \array_key_exists('apps', $result)) {
+			if ($result['apps'] === '' || $result['apps'] === null) {
+				return [];
+			}
+			return \explode(',', $result['apps']);
+		}
+
+		return false;
+	}
+
+	/**
+	 * Delete association of whitelisted apps and user
+	 * This method will grant full app access to the user as after this operation
+	 * there wont't be any whitelisted apps for the user.
+	 * There is another condition which would apply here is though the row with
+	 * user and the apps is deleted from the appaccess table:
+	 * - If the user is part of a group, and if that group has whitelisted apps,
+	 *   then the user still obeys the whitelisted apps of a group ( which it is a
+	 *   member of)
+	 * The argument used is string (uid) because its post deletion of the user. Hence
+	 * its risky to use user object at this phase.
+	 *
+	 * @param string $uid
+	 * @return bool true if operation is success, else false
+	 * @since 10.3.1
+	 */
+	public function wipeWhitelistedAppsForUser($uid) {
+		try {
+			$qb = $this->connection->getQueryBuilder();
+			$qb->delete('appaccess')
+				->where($qb->expr()->eq('accessor', $qb->expr()->literal($uid)))
+				->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal('0')));
+			$qb->execute();
+			return true;
+		} catch (\Exception $ex) {
+			$this->logger->logException($ex);
+			return false;
+		}
+	}
+
+	/**
+	 * Delete association of whitelisted apps and group
+	 * This method will grant full app access to the group as after this operation
+	 * there wont be any whitelisted apps for the group.
+	 * The argument used is string because its post deletion of the group. Hence
+	 * its risky to use group object at this phase.
+	 *
+	 * @param string $gid
+	 * @return bool true if operation is success, else false
+	 * @since 10.3.1
+	 */
+	public function wipeWhitelistedAppsForGroup($gid) {
+		try {
+			$qb = $this->connection->getQueryBuilder();
+			$qb->delete('appaccess')
+				->where($qb->expr()->eq('accessor', $qb->expr()->literal($gid)))
+				->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal('1')));
+			$qb->execute();
+			return true;
+		} catch (\Exception $ex) {
+			$this->logger->logException($ex);
+			return false;
+		}
+	}
+}
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -35,6 +35,7 @@ use OC\Memcache\ArrayCache;
 use OC\Memcache\NullCache;
 use OC\Installer;
 use OCP\App\AppNotFoundException;
+use OCP\App\IAppAccessService;
 use OCP\App\IAppManager;
 use OCP\App\AppManagerException;
 use OCP\App\ManagerEvent;
@@ -46,6 +47,7 @@ use OCP\IGroupManager;
 use OCP\IUser;
 use OCP\IUserSession;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 class AppManager implements IAppManager {
 
@@ -82,6 +84,8 @@ class AppManager implements IAppManager {
 	private $dispatcher;
 	/** @var IConfig */
 	private $config;
+	/** @var AppAccessService */
+	private $appAccessService;
 
 	/**
 	 * Apps as 'appId' => [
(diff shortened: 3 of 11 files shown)


lib/private/App/AppManager.php
```diff
@@ -160,9 +160,13 @@ public function getInstalledApps() {
 	 */
 	public function getEnabledAppsForUser(IUser $user = null) {
 		$apps = $this->getInstalledAppsValues();
-		$appsForUser = \array_filter($apps, function ($enabled) use ($user) {
-			return $this->checkAppForUser($enabled, $user);
-		});
+		$regulatedAppForUser = false;
+		if ($user !== null) {
+			$regulatedAppForUser = \OC::$server->getRegulateAppAccessService()->getWhitelistedAppsForUser($user);
```
jvillafanez: This should be injected. In addition, taking into account that the feature is deeply attached to the appManager, it might be interesting not to make the service "publicly available" but access to it through the appManager. I mean, something like `$appManager->getAccessService()`.

lib/private/Server.php
```diff
@@ -907,6 +908,10 @@ public function __construct($webRoot, \OC\Config $config) {
 		$this->registerService(IServiceLoader::class, function () {
 			return $this;
 		});
+
+		$this->registerService('RegulateAppAccessService', function (Server $c) {
```
jvillafanez: As said above, I'd rather not to expose this service directly but through the appManager.

lib/private/App/RegulateAppAccess.php
```diff
@@ -0,0 +1,89 @@
...
+	 * true if the apps list is set successfully. Else false is returned. Remember
+	 * this method can overwrite any existing data. Say for example an app decided
+	 * to update the whitelists accessible for user, it would be done.
+	 *
+	 * @param IUser $user
+	 * @param array $apps
+	 * @return bool
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForUser(IUser $user, $apps) {
```
jvillafanez: Taking into account that the main goal is to use this for the guests apps, how is the guests app expected to use this?
Note that the list of apps can change anytime.

lib/private/App/AppManager.php
```diff
@@ -82,6 +82,8 @@ class AppManager implements IAppManager {
 	private $dispatcher;
 	/** @var IConfig */
 	private $config;
+	/** @var RegulateAppAccess */
+	private $regulateAppAccess;
```
jvillafanez: this needs a more meaningful name.

jvillafanez:
Note that we need persistence in the service.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,175 @@
...
+	 */
+	public function getWhitelistedAppsForUser(IUser $user) {
+		if ($user === null) {
+			return false;
+		}
+		/**
+		 * Check if the user is available in different groups. If so grab the apps
+		 * user has access to these groups.
+		 */
+		$groupIds = $this->groupManager->getUserGroups($user);
```
jvillafanez: double-check the performance with LDAP and a user in a lot of groups. This needs to be blazing fast because it will affect to all requests.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,175 @@
...
+		$groupWhitelistedApps = [];
+		if (\is_array($groupIds) && !empty($groupIds)) {
+			foreach ($groupIds as $groupId) {
+				if (isset($this->appsAvailableForGroups[$groupId->getGID()])) {
+					$groupWhitelistedApps = \array_merge($groupWhitelistedApps, $this->appsAvailableForGroups[$groupId->getGID()]);
```
jvillafanez: we'll have to check the performance on this one too.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,175 @@
...
+	/**
+	 * Get the whitelisted apps for group
+	 * If the whitelisted apps are set for a group then it fetches from the array
+	 * appsAvailableForGroups and returns the array. Else it returns false.
+	 *
+	 * @param string $gid
+	 * @return bool|array get the array if whitelisted apps exist for it else false.
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppsForGroup($gid) {
```
jvillafanez: I'd use an `IGroup` for consistency with the other call.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,196 @@
...
+	 * In this method we set the whitelisted apps for a group. The appsAvailableForGroups
+	 * array with key as group id will be set with list of apps. The apps are list
+	 * of strings. If the apps are successfully set the method returns true. Else
+	 * false is returned.
+	 *
+	 * @param string $group
+	 * @param array $apps
+	 * @return bool true if whitelisted apps are set else false is returned
+	 */
+	public function setWhitelistedAppsForGroup(string $group, $apps = []) {
```
jvillafanez: Better use `IGroup` as type hint. This way we can ignore the `!empty($group)` check.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,196 @@
...
+	 * true if the apps list is set successfully. Else false is returned. Remember
+	 * this method can overwrite any existing data. Say for example an app decided
+	 * to update the whitelists accessible for user, it would be done.
+	 *
+	 * @param string $uid
+	 * @param string[] $apps
+	 * @return bool
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForUser($uid, $apps = []) {
```
jvillafanez: Better use `IUser` as type hint to remove the check below

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,196 @@
...
+	 * a false can be expected.
+	 *
+	 * Precisely it will check in 2 arrays appsAvailableForGroups and/or appsAvailableForUsers
+	 *
+	 * @param IUser $user
+	 * @return bool|mixed
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppsForUser(IUser $user) {
+		if ($user === null) {
```
jvillafanez: I don't think this is possible to happen.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,196 @@
...
+	 * Get whitelisted apps for user excluding the results of groups
+	 * This method fetches the whitelisted apps for the user. It does not considers
+	 * if the user is available in other groups and the apps whitelisted in other groups.
+	 *
+	 * @param IUser $user
+	 * @return bool|mixed
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppForUserExcludingGroups(IUser $user) {
+		if ($user !== null) {
```
jvillafanez: `$user === null` shouldn't happen

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,196 @@
...
+	 * Get the whitelisted apps for group
+	 * If the whitelisted apps are set for a group then it fetches from the array
+	 * appsAvailableForGroups and returns the array. Else it returns false.
+	 *
+	 * @param IGroup $group
+	 * @return bool|array get the array if whitelisted apps exist for it else false.
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppsForGroup(IGroup $group) {
+		if ($group === null) {
```
jvillafanez: This shouldn't happen

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,267 @@
...
+	 * This method would update the apps allowed for the group in the table.
+	 *
+	 * @param IGroup $group
+	 * @param array $apps
+	 * @return bool, true if the apps are updated in the table, else false.
+	 * @since 10.3.1
+	 */
+	public function updateWhitelistedAppsForGroup(IGroup $group, $apps = []) {
+		$qb = $this->connection->getQueryBuilder();
+		$qb->update('appaccess')
```
jvillafanez: Check for "upsert". I'm not sure about the method signature, but the expectation is that it updates the row if exists and insert a new one if it doesn't exists. We can also remove this `updateWhitelistedAppsForGroup` method then.

db_structure.xml
```diff
@@ -1646,4 +1646,56 @@
 
 	</table>
 
+	<table>
```
jvillafanez: I think it's better to use migrations.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,267 @@
...
+	 * this method can overwrite any existing data. Say for example an app decided
+	 * to update the whitelists accessible for user, it would be done.
+	 *
+	 * @param IUser $user
+	 * @param string[] $apps
+	 * @return bool
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForUser(IUser $user, $apps = []) {
+		if ($user !== null) {
```
jvillafanez: user can't be null

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,267 @@
...
+
+	/** @var ILogger */
+	private $logger;
+
+	/**
+	 * AppAccessService constructor.
+	 *
+	 * @param ILogger $logger
+	 */
+	public function __construct(ILogger $logger, IDBConnection $connection = null, IGroupManager $groupManager = null) {
```
jvillafanez: both connection and groupManager can be null, but there is no check for null. Either it's impossible that the connection is null and the "null" can be removed, or this will crash in some scenario.
Note that nullable dependencies are kind of bad, so if it's needed include a comment explaining the need.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,267 @@
...
+		$groupGids[] = $user->getUID();
+		foreach ($groupIds as $groupId) {
+			$groupGids[] = $groupId->getGID();
+		}
+		$qb = $this->connection->getQueryBuilder();
+		if (\is_array($groupIds) && !empty($groupIds)) {
+			$qb->select('apps')
+				->from('appaccess')
+				->where($qb->expr()->in('accessor', $qb->createPositionalParameter($groupGids, IQueryBuilder::PARAM_STR_ARRAY)));
+			foreach ($qb->execute()->fetchAll() as $value) {
```
jvillafanez: `fetchAll` is usually a bad practice because you don't know the amount of information that can be loaded into memory. Either ensure the amount of information is reasonably small or document what is the expected maximum limit.

db_structure.xml
```diff
@@ -1646,4 +1646,56 @@
...
+
+			<field>
+				<name>apps</name>
+				<type>clob</type>
+				<notnull>true</notnull>
+			</field>
+
+			<index>
+				<name>appaccess_accessor</name>
+				<unique>true</unique>
```
jvillafanez: this isn't right. Having an "admin" user and another "admin" group is common

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,267 @@
...
+	public function getWhitelistedAppsForGroup(IGroup $group) {
+		if ($group === null) {
+			return false;
+		}
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select(['apps'])
+			->from('appaccess')
+			->where($qb->expr()->eq('accessor', $qb->expr()->literal($group->getGID())))
+			->andWhere($qb->expr()->eq('accessor-identifier', $qb->expr()->literal('1')));
+		$result = $qb->execute()->fetch();
```
jvillafanez: always close DB cursor when you're done in order to release memory.

jvillafanez:
We might need some in-memory caching. I've checked with 10k LDAP users and 1k LDAP groups, with only one whitelisted app for one group. The user who belongs to all the groups has a delay of 0.5 secs on average per request (sometimes even some seconds).
Note that this is over a local network connection. Performance might be worse if the LDAP server is outside. On the other hand, it shouldn't be too bad under normal circumstances (a user isn't expected to be member of more than 20 groups).

There is also some weird things to investigate: the third request to the `getEnabledAppsForUser` method is the one that hits the 0.5 secs, the first and second are faster, but it seems they bypass this new access restrictions somehow (3rd call in my case is https://github.com/owncloud/core/blob/master/ocs/v1.php#L58)

sharidas:
Before I was updating the latest changeset, there was a thought in my mind, should I remove the caching ( the one we had in our previous versions ). Because if we keep caching, then we can reduce the hits to db.

Do you also recommend to cache `$this->groupManager->getUserGroups($user);` ( i.e, caching of users in groups)

core/Migrations/Version20191030124642.php
```diff
@@ -0,0 +1,88 @@
...
+ 			$appAccessTable->addColumn(
+				'apps',
+				Type::BLOB,
+				[
+					'notnull' => true,
+					'default' => '',
+				]
+			);
+
+ 			$appAccessTable->setPrimaryKey(['id']);
```
jvillafanez: It's unlikely we'll need indexes in this table (it's only planned for the guests group, and I don't think we'll need much more information), but I think we should include them as a general good practice

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,283 @@
...
+	private $groupManager;
+
+	/** @var ILogger */
+	private $logger;
+
+	/** @var array  */
+	private $userAccessAppList = [];
+
+	/** @var array */
+	private $groupAccessAppList = [];
```
jvillafanez: Remove the caching for now. I think it's better to ensure the code works properly, so the easier to follow the better.
We can add caching later on top of this class if it's really needed.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,283 @@
...
+			return $this->userAccessAppList[$user->getUID()];
+		}
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select(['apps'])
+			->from('appaccess')
+			->where($qb->expr()->eq('accessor', $qb->expr()->literal($user->getUID())))
+			->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal(0)));
+		$statement = $qb->execute();
+		$result = $$statement->fetch();
```
jvillafanez: `$$statement` doesn't look good

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,283 @@
...
+				$foundInCache = true;
+			}
+		}
+
+		// Not found in the cache hence check the db.
+		if (!$foundInCache) {
+			$qb = $this->connection->getQueryBuilder();
+			$qb->select('apps')
+				->from('appaccess')
+				->where($qb->expr()->in('accessor', $qb->createPositionalParameter($userGroupIds, IQueryBuilder::PARAM_STR_ARRAY)));
```
jvillafanez: this is buggy.
* "admin" user, member of "admin" group -> all apps whitelisted
* "admin" group -> "app1" and "app2" whitelisted
* "adminSubstitute" user, member of "admin" group -> "app3" and "app4" whitelisted

The "adminSustitute" should have only access to apps 1-4, but he will have access to all the apps with that query because it will get the whitelist from the "admin" user too instead of just the "admin" group.

lib/private/App/DummyAppAccessService.php
```diff
@@ -0,0 +1,161 @@
...
+
+	/**
+	 * DummyAppAccessService constructor.
+	 *
+	 * @param ILogger $logger
+	 * @param IDBConnection|null $connection
+	 * @param IGroupManager|null $groupManager
+	 * @since 10.3.1
+	 */
+	public function __construct(ILogger $logger, IDBConnection $connection = null, IGroupManager $groupManager = null) {
```
jvillafanez: This isn't needed. Plain constructor is enough
```
public function __construct() {
}
```

sharidas: Done

lib/private/App/AppAccessService.php
sharidas: Done

lib/private/App/AppAccessService.php
sharidas: Removed.

lib/private/App/AppAccessService.php
sharidas: Done.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,262 @@
...
+					$qb->expr()->eq('accessoridentifier', $qb->expr()->literal('0'))
+				),
+				$qb->expr()->andX(
+					$qb->expr()->in('accessor', $qb->createPositionalParameter($groupIds, IQueryBuilder::PARAM_STR_ARRAY)),
+					$qb->expr()->eq('accessoridentifier', $qb->expr()->literal('1'))
+				)
+			));
+		$statement = $qb->execute();
+		while ($row = $statement->fetch()) {
+			$whitelistedApps = \array_unique(\array_merge($whitelistedApps, \explode(',', $row['apps'])));
```
jvillafanez:
```
foreach (\explode(',', $row['apps']) as $app) {
  $whitelistedApps[$app] = $app;
}
```
That's likely faster. You don't need to make the array unique per row.
You can also return that kind of array (as long as it's documented) to avoid calling `array_values` to generate a new array.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,262 @@
...
+					$qb->expr()->eq('accessor', $qb->expr()->literal($user->getUID())),
+					$qb->expr()->eq('accessoridentifier', $qb->expr()->literal('0'))
+				),
+				$qb->expr()->andX(
+					$qb->expr()->in('accessor', $qb->createPositionalParameter($groupIds, IQueryBuilder::PARAM_STR_ARRAY)),
+					$qb->expr()->eq('accessoridentifier', $qb->expr()->literal('1'))
+				)
+			));
+		$statement = $qb->execute();
+		while ($row = $statement->fetch()) {
```
jvillafanez: double-check this. Most of the code examples uses`while (($row = $statement->fetch()) !== false)`.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,262 @@
...
+			$whitelistedApps = \array_unique(\array_merge($whitelistedApps, \explode(',', $row['apps'])));
+		}
+		$statement->closeCursor();
+
+		if (!empty($whitelistedApps)) {
```
jvillafanez: Empty whitelist should be a possible result.
There are a couple of scenarios to check:
* If the user (or group) doesn't have a DB entry, we should return false.
* If the user (or group) has an entry with an empty list, we should return an empty list

core/Migrations/Version20191030124642.php
```diff
@@ -0,0 +1,90 @@
...
+				[
+					'notnull' => true,
+					'default' => 0,
+					'unsigned' => true,
+				]
+			);
+
+ 			$appAccessTable->addColumn(
+				'apps',
+				Type::BLOB,
```
jvillafanez: Question: BLOB or TEXT? It seems to me that BLOB is inappropriate for this....
Alternatively, we could store the list into multiple rows, although this will require more code changes:
```
user1 | 0 | app1
user1 | 0 | app2
user1 | 0 | app3
....
```

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,262 @@
...
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppForUserExcludingGroups(IUser $user) {
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select(['apps'])
+			->from('appaccess')
+			->where($qb->expr()->eq('accessor', $qb->expr()->literal($user->getUID())))
+			->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal(0)));
+		$statement = $qb->execute();
+		$result = $statement->fetch();
```
jvillafanez: Include a comment saying that we expect only one row to be present for that user.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,262 @@
...
+	 * @since 10.3.1
+	 */
+	public function getWhitelistedAppsForGroup(IGroup $group) {
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select(['apps'])
+			->from('appaccess')
+			->where($qb->expr()->eq('accessor', $qb->expr()->literal($group->getGID())))
+			->andWhere($qb->expr()->eq('accessoridentifier', $qb->expr()->literal('1')));
+		$statement = $qb->execute();
+		$result = $statement->fetch();
```
jvillafanez: Same as above

lib/public/App/IAppAccessService.php
```diff
@@ -0,0 +1,118 @@
...
+	 * This method helps apps to store the information related to apps which can
+	 * be accessed by a collection of users. The advantage of this method is it
+	 * helps to allocate the apps array to the the collection of users. Internally
+	 * this method calls setWhitelistedAppsForUser method by iterating the users.
+	 *
+	 * @param IUser[] $users
+	 * @param string[] $apps
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForUsers(IUser $users, $apps = []);
```
jvillafanez: We can remove this method. I don't think it adds any value. The caller can loop through the users and call the `setWhitelistedAppsForUser` method

lib/private/App/AppAccessService.php
sharidas: Most of the code in core is following `if/while` `($row = $statment->fetch())`. Hence I followed this.

core/Migrations/Version20191030124642.php
```diff
@@ -0,0 +1,90 @@
...
+				[
+					'notnull' => true,
+					'default' => '',
+				]
+			);
+
+ 			$appAccessTable->setPrimaryKey(['id']);
+
+ 			$appAccessTable->addIndex(['accessor'], 'usergroupwith_whitelistedapps');
```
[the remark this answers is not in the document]
sharidas: Done. Updated with `$appAccessTable->addIndex(['accessor', 'accessoridentifier'], 'usergroupwith_whitelistedapps');`

lib/private/App/AppAccessService.php
sharidas: Done. Updated it.

core/Migrations/Version20191030124642.php
sharidas: Updated with TEXT.

lib/private/App/AppAccessService.php
sharidas: Added comment.

lib/public/App/IAppAccessService.php
sharidas: Removed this method.

lib/private/App/AppAccessService.php
sharidas: Done

lib/private/App/AppAccessService.php
sharidas: So now if the list is empty returend by user or group we return empty list.
if user or group does not have DB entry, user has access to all the apps.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,277 @@
...
+	 */
+	public function getWhitelistedAppsForUser(IUser $user) {
+		/**
+		 * Check if the user is available in different groups. If so grab the apps
+		 * user has access to these groups.
+		 */
+		$groupObjects = $this->groupManager->getUserGroups($user);
+		$whitelistedApps = [];
+		$grantAllAppsAccessForUser = true;
+		$userIds[] = $user->getUID();
```
jvillafanez: This isn't used

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,277 @@
...
+		$groupIds = [];
+		foreach ($groupObjects as $groupObject) {
+			$groupIds[] = $groupObject->getGID();
+		}
+
+		/**
+		 * First check if the user is available in the table.
```
jvillafanez: This isn't needed. There are 3 outcomes with the next query (the one getting the list of apps):
* No results -> this means all apps are allowed
* Results combined to an empty list -> this means the user won't have any app whitelisted
* Results combined to a list -> the user will have only those apps whitelisted.

lib/public/App/IAppAccessService.php
```diff
@@ -0,0 +1,106 @@
...
+	 * are successfully set the method returns true. Else false is returned.
+	 * Before executing this function kindly check if apps exist for the group
+	 * using one of the get methods in this class.
+	 *
+	 * @param IGroup $group
+	 * @param array $apps
+	 * @return bool true if whitelisted apps are set else false is returned
+	 * @since 10.3.1
+	 */
+	public function setWhitelistedAppsForGroup(IGroup $group, $apps = []);
```
jvillafanez: Keep the method interface consistent. The `setWhitelistedAppsForUser` call requires an array of apps, but it's optional here.

jvillafanez:
It seems we missed something to remove the DB entries.

lib/public/App/IAppAccessService.php
sharidas: I will update `setWhitelistedAppsForUser` to have optional. That way it becomes consistent.

lib/private/App/AppAccessService.php
sharidas: Removed.

lib/private/App/AppAccessService.php
sharidas: Removed and updated the code.

sharidas:
> It seems we missed something to remove the DB entries.

I have added 2 new methods:
- `grantAllAppsAccessForUser` -> to delete the user entry in the table and
- `grantAllAppsAccessForGroup` -> to delete the group entry in the table.

lib/private/App/AppAccessService.php
jvillafanez: empty list should be a valid result

lib/private/App/AppManager.php
```diff
@@ -768,4 +785,18 @@ public function getAppVersionByPath($path) {
...
 	}
+
+	/**
+	 * Get the regulated app access for each user
+	 *
+	 * @return AppAccessService
+	 * @since 10.3.1
+	 */
+	public function getAppsAccessServiceForUser() {
+		$this->dispatcher->addListener('group.postDelete', function (GenericEvent $event) {
```
jvillafanez: this is a really bad place to add hook. Each call will add a new hook.
In addition, try to keep all the hooks as close as possible (although it isn't a hook, you're doing things when the user is deleted), otherwise it will be difficult to find things.

sharidas: Moved this change to Server. Both the hook and event are now in the same place. Also its a small change of code. I have also removed the change made to User.php, where I called the wipe method for user in the delete method.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,310 @@
...
+		if (isset($result['apps'])) {
+			return \explode(',', $result['apps']);
+		}
+
+		return false;
+	}
+
+	/**
+	 * Delete association of whitelisted apps and user
+	 * This method will grant full app access to the user as after this operation
```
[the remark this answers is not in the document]
sharidas: Done.

sharidas:
Updated the changes requested for rename of methods mentioned here https://github.com/owncloud/core/pull/36338#issuecomment-551442545

sharidas:
Tested the sections:
- Created a user `adminSubst` and added this user to `admin` group
- Add empty list as whitelisted apps for this user ( just create a small script to call the API in this pr).
- Now try to navigate to the settings page
- The sharing page and encryption page are visible in the admin panel
   - The encryption page, though might look ugly. But as of now it looks ok. Because the ui is fetched from the core( the template ) and not from the encryption app.
    - The sharing page also fetches the template for the ui from the core.

Also verified with encryption ( **masterkey** ):
- The `admin` user ( who has access to the encryption module ) can upload/download file. The file is encrypted in oC.
- The `adminSubst` can upload/download file. The file is not encrypted ( expected behaviour) , as encryption app is not whitelisted for the user.

tests/lib/App/AppAccessServiceTest.php
```diff
@@ -0,0 +1,165 @@
...
+		$this->appAccessService->setWhitelistedAppsForUser($this->user1, []);
+		$result = $this->appAccessService->getComputedWhitelistedAppsForUser($this->user1);
+		$this->assertEquals(['' => ''], $result);
+	}
+
+	/**
+	 * This test proves that if user is not there in the table, then the user
+	 * has access to all the apps.
+	 */
+	public function testGetWhitelistedAppsForUserNotAddedToTable() {
```
jvillafanez: Having a `getWhitelistedAppsForUser` method I'd expect this test to test that method, not the "computed" one. Adjust the name of the test to reflect what you're testing.

tests/lib/App/AppAccessServiceTest.php
```diff
@@ -0,0 +1,165 @@
...
+		$this->group2->addUser($this->user2);
+
+		$this->appAccessService->setWhitelistedAppsForGroup($this->group1, ['app1', 'app2']);
+		$this->appAccessService->setWhitelistedAppsForGroup($this->group2, ['app1', 'app3', 'app4', 'app5']);
+
+		$this->appAccessService->setWhitelistedAppsForUser($this->user1, ['appUser1', 'appUser10']);
+		$this->appAccessService->setWhitelistedAppsForUser($this->user2, ['appUser2', 'appUser20']);
+
+		$this->assertEquals(
+			['app1' => 'app1', 'app2' => 'app2', 'app3' => 'app3', 'app4' => 'app4',
```
jvillafanez:
```
$expectedAppList = [
  'app1' => 'app1',
  'app2' => 'app2',
  ....
];
$this->assertEquals($expectedAppList, $this->appAccessService->.....);
```
This seems more readable

tests/lib/App/DummyAppAccessServiceTest.php
```diff
@@ -0,0 +1,72 @@
...
+		$groupObject = $this->createMock(IGroup::class);
+		$this->assertFalse($this->dummyAppAccessService->setWhitelistedAppsForGroup($groupObject));
+	}
+
+	public function testSetWhitelistedAppsForUser() {
+		$user = $this->createMock(IUser::class);
+		$this->assertFalse($this->dummyAppAccessService->setWhitelistedAppsForUser($user, []));
+	}
+
+	public function testGetWhitelistedAppsForUser() {
```
jvillafanez: Adjust the name of the test

tests/lib/App/ManagerTest.php
```diff
@@ -405,10 +414,27 @@ public function testGetAppsForUser() {
 		], $this->manager->getEnabledAppsForUser($user));
 	}
 
+	public function testGetWhitelistedAppsForUser() {
```
jvillafanez: change the name to `testGetEnabledAppsForUserWithWhitelist` or similar to make more clear what you're testing.

lib/private/App/AppManager.php
```diff
@@ -160,9 +166,13 @@ public function getInstalledApps() {
 	 */
 	public function getEnabledAppsForUser(IUser $user = null) {
 		$apps = $this->getInstalledAppsValues();
-		$appsForUser = \array_filter($apps, function ($enabled) use ($user) {
-			return $this->checkAppForUser($enabled, $user);
-		});
+		$regulatedAppForUser = false;
+		if ($user !== null) {
+			$regulatedAppForUser = $this->getAppsAccessServiceForUser()->getWhitelistedAppsForUser($user);
```
jvillafanez: :exclamation: This should be using the "computed" version, otherwise groups won't be considered.

lib/private/App/AppAccessService.php
```diff
@@ -0,0 +1,329 @@
...
+	 * there wont't be any whitelisted apps for the user.
+	 * There is another condition which would apply here is though the row with
+	 * user and the apps is deleted from the appaccess table:
+	 * - If the user is part of a group, and if that group has whitelisted apps,
+	 *   then the user still obeys the whitelisted apps of a group ( which it is a
+	 *   member of)
+	 *   member of)
```
jvillafanez: duplicated line

tests/lib/App/AppAccessServiceTest.php
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	public function testSetAndGetWhitelistedAppsForGroups() {
+		//Set the whitelisted apps for the group1
+		$this->appAccessService->setWhitelistedAppsForGroup($this->group1, ['foo', 'bar']);
+		//Set the whitelisted apps for the group2
+		$this->appAccessService->setWhitelistedAppsForGroup($this->group2, ['hello', 'world']);
+
+		$result = $this->appAccessService->getWhitelistedAppsForGroup($this->group1);
+		$this->assertEquals(['foo', 'bar'], $result, 'Ignore the order of the apps', 0.0, 10, true);
```
jvillafanez: Don't use undocumented parameters.

jvillafanez (commented):
Last details for me

lib/private/App/AppAccessService.php
sharidas: Fixed.

tests/lib/App/AppAccessServiceTest.php
sharidas: Removed this usage. Found another way to tackle the problem.

tests/lib/App/AppAccessServiceTest.php
```diff
@@ -0,0 +1,285 @@
...
+		$this->assertEquals(
+			[], \array_diff_assoc([
+			'app1' => 'app1',
+			'app2' => 'app2',
+			'app3' => 'app3',
+			'app4' => 'app4',
+			'app5' => 'app5',
+			'appUser1' => 'appUser1',
+			'appUser10' => 'appUser10'
+		], $this->appAccessService->getComputedWhitelistedAppsForUser($this->user1)));
```
jvillafanez: This seems badly formatted, and could be confusing when reading it. There are other cases above and below.

sharidas:
@DeepDiver1975 I would like your review on this PR. @jvillafanez had reviewed this PR. I would like your opinion, as its a new API change to core. Thanks.

PR #36354, opened 2019-10-31
Acceptance tests: fix create share function to use correct permission on uploadwriteonly

dpakach:
## Description

In acceptance tests the create share function added read permission in addition to create when we specify `uploadwriteonly` as permission string. this PR removes the read option and only keeps read as it acts as upload only public links.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareOperations/uploadToShare.feature
+++ b/tests/acceptance/features/apiShareOperations/uploadToShare.feature
@@ -440,7 +440,7 @@ Feature: sharing
     Then the HTTP status code should be "403"
 
   @smokeTest @public_link_share-feature-required
-  Scenario: Uploading to a public upload-read-write and no edit and no overwrite share
+  Scenario: Uploading to a public upload-write and no edit and no overwrite share
     Given as user "user0"
     And the user has created a public link share with settings
       | path        | FOLDER          |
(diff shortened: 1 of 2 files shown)


phil-davis (commented):
Also see https://github.com/owncloud/core/blob/master/tests/acceptance/features/bootstrap/Sharing.php#L175
```
5 = uploadwriteonly
```
should be 4?

phil-davis:
In https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareOperations/uploadToShare.feature#L443 we have some uplod-write-only public link test scenarios.

Can we add checks in those to make sure that, after uploading something, "the public" cannot delete or download it?

skshetry (commented):
Just need to resolve comments from Phil. Otherwise, looks good to me. 

The problem with the following function is, it is documenting something that it should not care about and is creating duplication. 

https://github.com/owncloud/core/blob/814259de6d6d40c840136a438de783506f6e56d4/tests/acceptance/features/bootstrap/Sharing.php#L171-L179

Just removing it and `@see ` reference would be great. cc @phil-davis

dpakach:
> In https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareOperations/uploadToShare.feature#L443 we have some uplod-write-only public link test scenarios.
> 
> Can we add checks in those to make sure that, after uploading something, "the public" cannot delete or download it?

@phil-davis  I will add those tests in another PR.

phil-davis (approved):

PR #36361, opened 2019-11-03
Typos

tdaget:
## Description
Minor typos corrections

## Motivation and Context
Softwares are always better without bugs or typos ...

## How Has This Been Tested?
no tests

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #36370, opened 2019-11-04
Formatting and other small stuff

mmattel:
This PR is based on the work of @tdaget at PR #36361(closed)
The referenced PR is now closed because of rebasing issues and CI will never get green there.

## Description

Fix some small formatting issues and minor corrections

## Related Issue

No related issue

## Motivation and Context

Improve code and readability

## How Has This Been Tested?

Run local

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: No changes needed

### DIFF
--- a/settings/templates/help.php
+++ b/settings/templates/help.php
@@ -1,55 +1,52 @@
 <div id="app-navigation">
 	<ul>
-	<?php if ($_['admin']) {
-	?>
-		<li>
-			<a class="<?php p($_['style1']); ?>" target="_blank"
-				href="https://doc.owncloud.com/server/user_manual/index.html">
-				<?php p($l->t('User documentation')); ?> ↗
-			</a>
-		</li>
-		<li>
-			<a class="<?php p($_['style2']); ?>" target="_blank"
-				href="https://doc.owncloud.com/server/admin_manual/index.html">
-				<?php p($l->t('Administrator documentation')); ?> ↗
-			</a>
-		</li>
-		<li>
-			<a class="<?php p($_['style1']); ?>" target="_blank"
-			   href="https://doc.owncloud.com/server/developer_manual/index.html">
-				<?php p($l->t('Developer documentation')); ?> ↗
-			</a>
-		</li>
-	<?php
-} ?>
+		<?php if ($_['admin']) { ?>
+			<li>
+				<a class="<?php p($_['style1']); ?>" target="_blank"
+					href="https://doc.owncloud.com/server/user_manual/index.html">
+					<?php p($l->t('User documentation')); ?>
+				</a>
+			</li>
+			<li>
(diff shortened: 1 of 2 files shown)


tdaget:
You can add a typo correction in tests/data/app/invalid-info.xml
line 2: add `>` char at the end of `<info>``

Thanks

mmattel:
> tests/data/app/invalid-info.xml

I think that `<info` is correctly wrong to show that the content is invalid.
But I can´t decide on that in detail - therefor not corrected.

settings/templates/help.php
```diff
@@ -1,55 +1,52 @@
...
 		<li>
 			<a href="https://owncloud.org/support" target="_blank" rel="noreferrer">
-				<?php p($l->t('Online documentation')); ?> ↗
+				<?php p($l->t('Online documentation')); ?>
 			</a>
 		</li>
 		<li>
 			<a href="https://central.owncloud.org" target="_blank" rel="noreferrer">
-				<?php p($l->t('Forum')); ?> ↗
+				<?php p($l->t('Forum')); ?>
```
phil-davis: The  ↗ character has been removed in various places, like this line.
Is there a reason for that?

phil-davis (commented):
And I raised issue #36381 because it seems that `settings/templates/help.php` is not accessible any more. So someone needs to decide what the requirement is for it.

settings/templates/help.php
mmattel: Yes, `↗` is not part of any parsable html code, most likely a leftover from an editor.
Removing it is cleaning up stuff.

mmattel:
@phil-davis is codecov a necessity in this case and if, how to make that green ?

settings/templates/help.php
phil-davis: The  ↗ character does actually display on the UI. I guess it was put there "back in the old days"  to indicate that this is a link and the user will be taken somewhere when they click it.
Probably should be indicated in a different way "these days".
"somebody" will need to decide exactly what should be done.

phil-davis:
> @phil-davis is codecov a necessity in this case and if, how to make that green ?

The files edited here are really `.html` files that have "old fashioned" embedded PHP. But they have `.php` file extension, so `phpunit` includes them in coverage calculations.

I guess somebody will override the codecov result for a PR  like this.

settings/templates/help.php
phil-davis: This needs someone with knowledge of project history to say why those "little arrow" characters were chosen to use here. And then to say what should be done nowadays for that sort of thing.

DeepDiver1975:
old aka dead -> close

PR #36384, opened 2019-11-06
Allow sharing with guest users when sharing is limited by user groups

VicDeo:
## Description
Allow sharing with guest users when `Restrict users to only share with users in their groups` option is enabled

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3578

## Motivation and Context
Guests app should not be affected by the ordinary sharing group restriction

## How Has This Been Tested?
1. Enable the guests app
2.  Check `Restrict users to only share with users in their groups` on the settings page
3. Share any item with `someuser@example.org` (guest)

### Expected 
User is created, a share is added for this user. No errors

### Actual
User is created, but the share is NOT added for this user. Sharing error popup in the web UI

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Helper/UserTypeHelper.php
+++ b/lib/private/Helper/UserTypeHelper.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * @author Viktar Dubiniuk <dubiniuk@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Helper;
+
+use OCP\App\IAppManager;
+use OCP\IConfig;
+
+class UserTypeHelper {
+	/** @var IAppManager */
+	private $appManager;
+
+	/** @var IConfig */
+	private $config;
+
+	public function __construct(IAppManager $appManager = null, IConfig $config = null) {
+		$this->appManager = $appManager;
+		if ($appManager === null) {
+			$this->appManager = \OC::$server->getAppManager();
+		}
+		$this->config = $config;
+		if ($config === null) {
+			$this->config = \OC::$server->getConfig();
+		}
+	}
+
+	/**
+	 * Checks whether given uid belongs to the user created by the guests app
+	 *
+	 * @param string $uid
+	 * @return bool
+	 */
+	public function isGuestUser($uid) {
+		$guestsAppEnabled = $this->appManager->isEnabledForUser('guests');
+		if ($guestsAppEnabled) {
+			return (bool) $this->config->getUserValue(
+				$uid, 'owncloud', 'isGuest', false
+			);
+		}
+		return false;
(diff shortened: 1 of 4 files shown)


jvillafanez:
We might need to consider to create a `UserTypeDetection` class or similar at least to move this code to a place. It isn't the first place we try to detect if a user is a guest, and it won't be the last.

micbar:
looks promising.

lib/private/Share20/Manager.php
```diff
@@ -471,6 +472,13 @@ protected function userCreateChecks(\OCP\Share\IShare $share) {
 					$this->groupManager->getUserGroupIds($sharedBy),
 					$this->groupManager->getUserGroupIds($sharedWith)
 			);
+
+			// We still should be able to share with guest user even when it's not in the same group
+			$userTypeHelper = new UserTypeHelper();
+			if ($userTypeHelper->isGuestUser($sharedWith->getUID())) {
+				return;
```
karakayasemi: Returning from here misses the rest of the userCreateChecks like already shared path check. I guess, we can change the upper if condition to something like the below condition instead of returning: 
`if ($this->shareWithGroupMembersOnly() && !$userTypeHelper->isGuestUser($sharedWith->getUID()))`

karakayasemi (commented):
I am not familiar with the Guest app concept, so I am not sure of the concept. In terms of code, rest of the things are good.

lib/private/Share20/Manager.php
```diff
@@ -471,7 +472,10 @@ protected function userCreateChecks(\OCP\Share\IShare $share) {
 					$this->groupManager->getUserGroupIds($sharedBy),
 					$this->groupManager->getUserGroupIds($sharedWith)
 			);
-			if (empty($groups)) {
+
+			// We still should be able to share with guest user even when it's not in the same group
+			$userTypeHelper = new UserTypeHelper();
+			if (empty($groups) && !$userTypeHelper->isGuestUser($sharedWith->getUID())) {
```
karakayasemi: I know, it is not a big deal about performance, but if the user is a guest, no need to make any calculation about group intersection. Because of that, moving this check to one upper level is better in terms of performance.

karakayasemi (changes requested):
I know, it is not a big deal about performance, but if the user is a guest, no need to make any calculation about group intersection. Because of that, moving this check to one upper level is better in terms of performance.

micbar (approved):

karakayasemi (approved):

lib/private/Helper/UserTypeHelper.php
```diff
@@ -0,0 +1,60 @@
...
+use OCP\IConfig;
+
+class UserTypeHelper {
+	/** @var IAppManager */
+	private $appManager;
+
+	/** @var IConfig */
+	private $config;
+
+	public function __construct(IAppManager $appManager = null, IConfig $config = null) {
```
jvillafanez: include a comment explaining why the appManager and config can be null.

lib/private/Helper/UserTypeHelper.php
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * Checks whether given uid belongs to the user created by the guests app
+	 *
+	 * @param string $uid
+	 * @return bool
+	 */
+	public function isGuestUser($uid) {
+		$guestsAppEnabled = $this->appManager->isEnabledForUser('guests');
+		if ($guestsAppEnabled) {
```
jvillafanez: should the method return true for the guest user even if the app is disabled? I find it weird that a user is reported as guest, but then, after disabling the app, the user isn't a guest any longer.

lib/private/Share20/Manager.php
```diff
@@ -462,15 +463,19 @@ protected function validateExpirationDate(\OCP\Share\IShare $share) {
 	 * @throws \Exception
 	 */
 	protected function userCreateChecks(\OCP\Share\IShare $share) {
+		$userTypeHelper = new UserTypeHelper();
```
jvillafanez: this should be injected

PR #36399, opened 2019-11-11
Fix for app:enable in case of not valid/not existing enterprise key

### LINKED ISSUE #36351: occ app:enable should exit with proper error message if enterprise key is invalid
### Steps to reproduce
1. Disable an enterprise app
2. Set an invalid Enterprise Key
3. Try to enable an enterprise app

### Expected behaviour
The `occ app:enable example-ee-app` command should exit with a proper error message and notify the user that the app will not be enabled due to an invalid license key.

### Actual behaviour
The `occ app:enable example-ee-app` command exited with a success message while the app is still disabled. Only if you look into the owncloud.log you will see an error message about the invalid license key.

### Server configuration
**Operating system**: Ubuntu 18.04

**Web server:** Apache

**Database:** MySQL

**PHP version:** 7.2

**ownCloud version:** 10.3

pako81:
## Description

This PR fixes the current behaviour of the `app:enable` occ command in case of not valid/not existing enterprise key.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/36351

## Motivation and Context

Currently, when running the `app:enable` occ command for enabling an enterprise app and the enterprise key is not valid/not existing, the command does not return any warning/error message. Still, the enterprise app stays disabled. Only when looking into the owncloud.log file one will see an error message about the invalid license key.

## How Has This Been Tested?

Manually, by: - disabling an enterprise app, - setting an invalid enteprise key, - trying to re-enable that enterprise app. Now, the `$appId cannot be enabled because of invalid enteprise key` message is shown.

Implemented for both cases in case groups are/are not given as input to the command.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Changelog item

### DIFF
--- a/core/Command/App/Enable.php
+++ b/core/Command/App/Enable.php
@@ -24,6 +24,7 @@
 
 namespace OC\Core\Command\App;
 
+use OCP\IConfig;
 use OCP\App\IAppManager;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
@@ -36,12 +37,17 @@ class Enable extends Command {
 	/** @var IAppManager */
 	protected $manager;
 
+	/** @var IConfig */
+	protected $config;
+
 	/**
 	 * @param IAppManager $manager
+	 * @param IConfig $config
 	 */
-	public function __construct(IAppManager $manager) {
+	public function __construct(IAppManager $manager, IConfig $config) {
 		parent::__construct();
 		$this->manager = $manager;
+		$this->config = $config;
 	}
 
 	protected function configure() {
@@ -68,6 +74,14 @@ class Enable extends Command {
 		if (!\OC_App::getAppPath($appId)) {
 			$output->writeln($appId . ' not found');
 			return 1;
+		}
+
+		if (\OC_App::isEnabled('enterprise_key')) {
+			$key = new \OCA\Enterprise_Key\EnterpriseKey(false, $this->config);
+			if ($key && !$key->isValid()) {
+				$output->writeln($appId . ' cannot be enabled because of invalid enterprise key');
+				return 1;
+			}
 		}
 
 		$groups = $input->getOption('groups');
--- a/core/Command/App/Enable.php
+++ b/core/Command/App/Enable.php
@@ -77,7 +77,9 @@ class Enable extends Command {
 		}
 
 		if (\OC_App::isEnabled('enterprise_key')) {
+			/* @phan-suppress-next-line PhanUndeclaredClassMethod */
 			$key = new \OCA\Enterprise_Key\EnterpriseKey(false, $this->config);
+			/* @phan-suppress-next-line PhanUndeclaredClassMethod */
 			if ($key && !$key->isValid()) {
 				$output->writeln($appId . ' cannot be enabled because of invalid enterprise key');
 				return 1;
--- a/changelog/unreleased/36399
+++ b/changelog/unreleased/36399
@@ -1,0 +1,10 @@
+Bugfix: Inform the admin if they enable an enterprise app without valid key
+
+Previously no message was displayed but the app was not enabled.
+
+Now, when the admin tries to enable an enterprise app when the enterprise key
+is invalid, the message "cannot be enabled because of invalid enterprise key"
+is displayed.
+
+https://github.com/owncloud/core/issues/36351
+https://github.com/owncloud/core/pull/36399
--- a/core/register_command.php
+++ b/core/register_command.php
@@ -59,7 +59,7 @@ $application->add(new \OC\Core\Command\Integrity\CheckCore(
 
 if (\OC::$server->getConfig()->getSystemValue('installed', false)) {
 	$application->add(new OC\Core\Command\App\Disable(\OC::$server->getAppManager()));
-	$application->add(new OC\Core\Command\App\Enable(\OC::$server->getAppManager()));
+	$application->add(new OC\Core\Command\App\Enable(\OC::$server->getAppManager(), \OC::$server->getConfig()));
 	$application->add(new OC\Core\Command\App\GetPath());
 	$application->add(new OC\Core\Command\App\ListApps(\OC::$server->getAppManager()));
 	
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -34,6 +34,9 @@ parameters:
     - '#Undefined variable: \$vendor#'
     - '#Undefined variable: \$baseuri#'
     - '#Instantiated class OC_Theme not found.#'
+    -
+      message: '#Instantiated class OCA\\Enterprise_Key\\EnterpriseKey not found.#'
+      path: core/Command/App/Enable.php
     # errors below are to be addressed by own pull requests - non trivial changes required
     - '#OCA\\DAV\\Connector\\Sabre\\ObjectTree::__construct\(\) does not call parent constructor from Sabre\\DAV\\Tree.#'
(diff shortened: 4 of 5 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/core/21265/3/6
```
enabling apps
+ echo 'enabling apps'
+ php occ app:enable files_sharing
An unhandled exception has been thrown:
Error: Class 'OCA\Enterprise_Key\EnterpriseKey' not found in /drone/src/core/Command/App/Enable.php:74
```
It still needs to work for the community edition, when there is no `enterprise_key` app.

core/Command/App/Enable.php
```diff
@@ -71,10 +71,19 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		}
 
 		$groups = $input->getOption('groups');
+		$key = new \OCA\Enterprise_Key\EnterpriseKey(false, \OC::$server->getConfig());
 		if (empty($groups)) {
```
micbar: Please inject config in constructor.

micbar (dismissed):

core/Command/App/Enable.php
pako81: will do

core/Command/App/Enable.php
```diff
@@ -71,10 +77,21 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		}
 
 		$groups = $input->getOption('groups');
-		if (empty($groups)) {
+		if (\OC_App::isEnabled('enterprise_key')) {
+			$key = new \OCA\Enterprise_Key\EnterpriseKey(false, \OC::$server->getConfig());
+		}
```
micbar: use the global property here: $this->config

pako81:
better now?

core/Command/App/Enable.php
```diff
@@ -71,10 +77,21 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		}
 
 		$groups = $input->getOption('groups');
+		if (\OC_App::isEnabled('enterprise_key')) {
+			$key = new \OCA\Enterprise_Key\EnterpriseKey(false, $this->config);
+		}
 		if (empty($groups)) {
+			if ($key && !$key->isValid()) {
+				$output->writeln($appId . ' cannot be enabled because of invalid enteprise key');
```
phil-davis:
```suggestion
				$output->writeln($appId . ' cannot be enabled because of invalid enterprise key');
```

core/Command/App/Enable.php
```diff
@@ -71,10 +77,21 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+				return 1;
+			}
 			\OC_App::enable($appId);
 			$output->writeln($appId . ' enabled');
 		} else {
+			if ($key && !$key->isValid()) {
+				$output->writeln($appId . ' cannot be enabled because of invalid enteprise key');
```
phil-davis:
```suggestion
				$output->writeln($appId . ' cannot be enabled because of invalid enterprise key');
```

core/Command/App/Enable.php
```diff
@@ -71,10 +77,21 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		}
 
 		$groups = $input->getOption('groups');
+		if (\OC_App::isEnabled('enterprise_key')) {
+			$key = new \OCA\Enterprise_Key\EnterpriseKey(false, $this->config);
```
karakayasemi: To prevent creating duplicate code block, you can validate the key directly after `$key` assignment. Also, in this way, you do not need to check whether the key exists.

pako81: The duplicate code block is actually needed depending if the `$groups` option has been given or not as input to the occ command, and it was already there before this PR. Or did I misunderstand the suggestion?

core/Command/App/Enable.php
```diff
@@ -71,10 +77,21 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+		}
 		if (empty($groups)) {
+			if ($key && !$key->isValid()) {
```
phil-davis: He means that this `if` test can be moved to before `if (empty($groups))` and then it is only needed once.

core/Command/App/Enable.php
karakayasemi: As the last step, just run `make test-php-style-fix` for code style adjustment.

core/Command/App/Enable.php
```diff
@@ -70,6 +76,14 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			return 1;
 		}
 
+		if (\OC_App::isEnabled('enterprise_key')) {
+		$key = new \OCA\Enterprise_Key\EnterpriseKey(false, $this->config);
```
phil-davis: Indent is strange

core/Command/App/Enable.php
```diff
@@ -70,6 +76,14 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+		        if ($key && !$key->isValid()) {
+			$output->writeln($appId . ' cannot be enabled because of invalid enterprise key');
+			return 1;
```
phil-davis: And these lines need indent inside the `if` block

phil-davis (approved):

phil-davis:
@micbar is this progressing?
(could have an acceptance test added if it is planned for release...)

phil-davis:
@pako81 @micbar I  rebased and fixed the code-style fail.
Let's see if the "real" CI passes...

core/Command/App/Enable.php
micbar: The enterprise key app is not enabled in CI. Therefore it fails.
@phil-davis

phil-davis: Looking at it - when can likely give the code analysis tool(s) a clue to skip that line

phil-davis:
And adjusted the unit test so it knows about the new param to the constructor.
Maybe this time CI will pass?

micbar:
> And adjusted the unit test so it knows about the new param to the constructor.
> Maybe this time CI will pass?

PHpstan fails because the code of Enterprise Key is not available.

phil-davis:
> > And adjusted the unit test so it knows about the new param to the constructor.
> > Maybe this time CI will pass?
> 
> PHpstan fails because the code of Enterprise Key is not available.

I "fixed" that by making it ignore that - phpstan CI is passing.

phil-davis:
I added a changelog. I think this is ready for review.

There are existing `cliProvisioning/enableApp.feature` tests that cover enabling an ordinary app - so that ensures that this code change did not break anything for ordinary apps.

We are not easily able to get ourselves an enterprise app and the enterprise_key app etc during core CI. So this change will need manual testing during the release QA process.

karakayasemi (approved):

phil-davis (approved):

core/Command/App/Enable.php
```diff
@@ -70,6 +76,16 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			return 1;
 		}
 
+		if (\OC_App::isEnabled('enterprise_key')) {
```
phil-davis: The issue we found now is that if you have the enterprise app enabled, but have an invalid enterprise key, and you are just trying to enable a  non-enterprise app, then the app cannot be enabled. e.g. enabling the `testing` app does not work.

If there is some way to check if the app being enabled is an Enterprise app, then this could be fixed by:
```
if (\OC_App::isEnterpriseApp($appId) && \OC_App::isEnabled('enterprise_key')) {
```

Issue #36064, opened 2019-08-20
new public webdav API does not obey LOCKS


individual-it:
### Steps to reproduce
1. lock a folder `curl -X LOCK -u admin:admin  http://localhost/owncloud-core/remote.php/webdav/public-folder -d "<?xml version='1.0' encoding='UTF-8'?><d:lockinfo xmlns:d='DAV:'> <d:lockscope><d:shared/></d:lockscope></d:lockinfo>"`
2. share this folder via public link
3. upload new files or overwrite existing files in that folder via new public webdav API `curl -X PUT http://localhost/owncloud-core/remote.php/dav/public-files/0M2kxLVbYuIOSPZ/12.txt -d "222"`

### Expected behaviour
upload does not work

### Actual behaviour
files can be uploaded and overwritten

CC @DeepDiver1975

DeepDiver1975:
Needs analysis .... will have a look. THX a lot

phil-davis:
Note: #38922 might fix some or all of this.

Issue #34394, opened 2019-02-06
even locking is forbidden on public webdav supportedlock property is still returned


individual-it:
### Steps to reproduce
1. share a folder publicly with read/write access
2. access a file that public share with LibreOffice "Open Remote" function

### Expected behaviour
LibreOffice should open the file and should be able to save it back

### Actual behaviour
LibreOffice shows an error message
![image](https://user-images.githubusercontent.com/2425577/52320437-6fbf9c00-29f7-11e9-99a4-c265fb783b48.png)
and only opens the file read-only

Also new files cannot be saved by LibreOffice 
![image](https://user-images.githubusercontent.com/2425577/52320569-2a4f9e80-29f8-11e9-9fbd-e937cdb22998.png)

The reason is that LibreOffice thinks that Locking is possible but LOCK requests get a 403 response
LibreOffice assumes that locks are possible because it made a `supportedlock` discovery and got a positive response:

```
PROPFIND /owncloud-10.1.0/owncloud/public.php/webdav/Documents2/Example.odt HTTP/1.1
<?xml version="1.0" encoding="utf-8"?>
<propfind xmlns="DAV:"><prop>
<resourcetype xmlns="DAV:"/>
<IsReadOnly xmlns="http://ucb.openoffice.org/dav/props/"/>
<getcontenttype xmlns="DAV:"/>
<supportedlock xmlns="DAV:"/>
</prop></propfind>


<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:oc="http://owncloud.org/ns"><d:response><d:href>/owncloud-10.1.0/owncloud/public.php/webdav/Documents2/Example.odt</d:href><d:propstat><d:prop><d:resourcetype/><d:getcontenttype>application/vnd.oasis.opendocument.text</d:getcontenttype><d:supportedlock><d:lockentry><d:lockscope><d:exclusive/></d:lockscope><d:locktype><d:write/></d:locktype></d:lockentry><d:lockentry><d:lockscope><d:shared/></d:lockscope><d:locktype><d:write/></d:locktype></d:lockentry></d:supportedlock></d:prop><d:status>HTTP/1.1 200 OK</d:status></d:propstat><d:propstat><d:prop><x1:IsReadOnly xmlns:x1="http://ucb.openoffice.org/dav/props/"/></d:prop><d:status>HTTP/1.1 404 Not Found</d:status></d:propstat></d:response></d:multistatus>
[...]

PR #36402, opened 2019-11-11
public.php and public-files endpoints won't report locking support

jvillafanez:
## Description
Remove support for persistent locking in the public endpoints. Requested in https://github.com/owncloud/core/issues/34394#issuecomment-460924172
Note that locking wasn't supported in public endpoints as seen in https://github.com/owncloud/core/blob/master/apps/dav/lib/Files/FileLocksBackend.php#L166-L168 . This PR will remove the lockdiscovery from the propfind response

## Related Issue
https://github.com/owncloud/core/issues/34394

## Motivation and Context
Libreoffice was showing an error trying to lock the file through the public endpoint. Now it won't try to lock the file.

## How Has This Been Tested?
Manually tested using the steps defined in https://github.com/owncloud/core/issues/34394

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/PublicDavLocksPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/PublicDavLocksPlugin.php
@@ -1,0 +1,72 @@
+<?php
+/**
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\Connector\Sabre;
+
+use Sabre\DAV\Locks\Backend\BackendInterface;
+use Sabre\DAV\PropFind;
+use Sabre\DAV\INode;
+use Sabre\DAV\Exception\MethodNotAllowed;
+use Sabre\DAV\Exception\Locked;
+use Sabre\HTTP\RequestInterface;
+use Sabre\HTTP\ResponseInterface;
+use Sabre\DAV\Xml\Property\LockDiscovery;
+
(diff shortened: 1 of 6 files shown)


jvillafanez:
Need to check if this is the approach we want or not. The PR fixes the issue mentioned, but it break quite a bunch of lock tests, which would need to be removed or adjusted.
There are other tests failing that don't seem to be related with the changes. We need to double-check those.

jvillafanez:
> Need to check if this is the approach we want or not. The PR fixes the issue mentioned, but it break quite a bunch of lock tests, which would need to be removed or adjusted.
There are other tests failing that don't seem to be related with the changes. We need to double-check those.

This is fixed. Failing test caused by a infrastructure problem.

> Note that this could be dangerous if the file is locked and someone edits the file through the public link at the same time.

Still happening. The file is edited through the public link even though the file is locked.

jvillafanez:
Copying from the commit:

For public links:
* LOCK and UNLOCK methods won't be reported as allowed
* lock information (d:lockdiscovery) and support (d:supportedlock) will
be shown in the propfind if requested (lock support is needed for
libreoffice to try to lock the file)
* trying to lock the file through public link will either throw a Locked
exception if there is a lock in place, or MethodNotAllowed if there are
no locks (the Locked exception is needed for libreoffice to show a popup
warning the user that the file is locked)

This means that the file can't be locked through the public link, but both the "Locked" exception trying to unlock and also the information in the propfind should be enough information to know if the file is locked or not.
If the file is locked, libreoffice will show a popup warning the user that the file is locked and he can open the file as read-only or a copy (this seems quite nice because the user knows there is someone editing the file - no user info is shown though). If the file isn't locked, the user can edit the file normally, but the file won't be locked.

There is still some polishing to do in the PR, but I think this is the behaviour we want.

jvillafanez:
coverage upload errors.... the tests are fine

phil-davis:
@micbar this is in the current server sprint.
Are we aiming to merge it for 10.4?
There are a few small and good changes to acceptance tests here - just needs a quick review to think if there are extra scenarios worth adding.

micbar (approved):

phil-davis:
Note that this PR does not fix #36064 - if a folder/file has been locked by the owner, and a public link share of it has been made, then the public can still upload files etc into the locked folder...
See the scenarios in `tests/acceptance/features/apiWebdavLocks/publicLink.feature`

So this PR is stage1, fixing up a bit of locking support stuff. But actually it will be good to also "some day soon" address #36064 so that, e.g. public uploads into a locked folder will get rejected.

phil-davis:
> There are a few small and good changes to acceptance tests here - just needs a quick review to think if there are extra scenarios worth adding.

This looks OK to merge.

Issue #36072, opened 2019-08-21
When the DB server goes away, HTTP Status code stays 200 and oC shows exception in browser


enbrnz:
### Steps to reproduce
1. Install ownCloud
2. Stop MySQL or MariaDB server
3. Visit ownCloud URL

### Expected behaviour
This should show a 500 Server Error

### Actual behaviour
Webserver replies with HTTP Status code 200 and the following exception is shown to end users:
```
Exception occurred while logging exception: Failed to connect to the database: An exception occured in driver: SQLSTATE[HY000] [2002] No such file or directory
#0 /var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(429): OC\DB\Connection->connect()
#1 /var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(389): Doctrine\DBAL\Connection->getDatabasePlatformVersion()
#2 /var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(328): Doctrine\DBAL\Connection->detectDatabasePlatform()
#3 /var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(623): Doctrine\DBAL\Connection->getDatabasePlatform()
#4 /var/www/owncloud/lib/private/DB/Connection.php(144): Doctrine\DBAL\Connection->setTransactionIsolation(2)
#5 /var/www/owncloud/lib/composer/doctrine/dbal/lib/Doctrine/DBAL/DriverManager.php(172): OC\DB\Connection->__construct(Array, Object(Doctrine\DBAL\Driver\PDOMySql\Driver), Object(Doctrine\DBAL\Configuration), Object(Doctrine\Common\EventManager))
#6 /var/www/owncloud/lib/private/DB/ConnectionFactory.php(145): Doctrine\DBAL\DriverManager::getConnection(Array, Object(Doctrine\DBAL\Configuration), Object(Doctrine\Common\EventManager))
#7 /var/www/owncloud/lib/private/Server.php(529): OC\DB\ConnectionFactory->getConnection('mysql', Array)
#8 /var/www/owncloud/lib/composer/pimple/pimple/src/Pimple/Container.php(113): OC\Server->OC\{closure}(Object(OC\Server))
[...]

PR #36413, opened 2019-11-12
[For 10.4] Set 599 HTTP code on error

jvillafanez:
## Description
web page shows a 200 status instead of a 500 (or similar) in case of error

## Related Issue
https://github.com/owncloud/core/issues/36072

## How Has This Been Tested?
Load any ownCloud web page without having connection to the DB (ownCloud was installed but the DB connection was cut later)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/36413
+++ b/changelog/unreleased/36413
@@ -1,0 +1,11 @@
+Bugfix: Set 599 HTTP code on error
+
+Previously, a hard crash, such as DB being down, was being reported with a stacktrace
+and a 200 HTTP code. In order to homogenize the behaviour with all the endpoints, we've
+changed the behaviour to return a 599 HTTP code and an empty content.
+
+In addition, we've included a new option in the config.php, "crashdirectory", which
+defaults to the "datadirectory" set, where the crash logs will be created. Note that
+normal errors will still be reported normally and will log into the owncloud.log file.
+
+https://github.com/owncloud/core/pull/36413
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -83,6 +83,17 @@ $CONFIG = array(
 'datadirectory' => '/var/www/owncloud/data',
 
 /**
+ * Define the directory where the crash logs will be stored.
+ * By default, this will be the same as the one configured as "datadirectory".
+ * The directory MUST EXIST and be WRITABLE by the web server.
+ * Note that crashes are extremely rare (although they can come in burst due to
+ * multiple requests), so the default location is usually fine.
+ * Also note that the log can contain sensitive information, but it should be useful
+ * to pinpoint where is the problem.
+ */
+'crashdirectory' => '/var/www/owncloud/data',
+
+/**
  * Current version number of your ownCloud installation
  * This is set up during installation and update, so you shouldn't need to change it.
  */
--- a/console.php
+++ b/console.php
@@ -56,9 +56,19 @@ if (\stripos(PHP_OS, 'WIN') === 0) {
 }
 
 function exceptionHandler($exception) {
-	echo 'An unhandled exception has been thrown:' . PHP_EOL;
-	echo $exception;
-	exit(1);
+	try {
+		// try to log the exception
+		\OC::$server->getLogger()->logException($ex, ['app' => 'index']);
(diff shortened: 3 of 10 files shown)


jvillafanez:
Summing up the changes:
* Shutdown log is removed due to unreliability. There is no guarantee the ownCloud's logger will be available during the shutdown (seen when there is no connection to the DB)
* /favicon.ico route will be rewritten to /core/img/favicon.ico. This happens when there is no connection to the DB. The browser doesn't load any page due to the server crash, and make a request to "/favicon.ico" to get the favicon. The change will make the favicon load properly under these circunstances.
* New logger for crashes and improved error handling on the endpoints.

-----------

Notes on the error handling:

The request processing goes as follows:
1. Try to process the request normally. If there isn't any problem, this will generate a normal response.
2. In case the request isn't handled properly and throws an exception that isn't handled, the endpoint should try to log that exception using ownCloud's logging facility and show a proper error message to the user.

    This will likely depend on the endpoint, but generally, the expected HTTP status should be 500 and an error message should be presented to the user. ownCloud already provides some templates that can be used, such as `OC_Template::printExceptionErrorPage`. In this case, we can assume that ownCloud is working properly, and it's an app the one that failed to process the request.

    Note that this exception handling is expected to be light, mostly log the error and show a nice message to the user
3. In case we can't handle the exception, this means that ownCloud is broken and we can't report the error properly to the user. This has been seen if the connection to the DB is stopped, but there could be some additional scenarios.

    In this case we assume that ownCloud is unreliable (we've decided to just log and show an error in step 2, but we couldn't even do that). In order to provide a similar behaviour in all the endpoints, we'll perform the following actions:

    * For web environments (not CLI):
       1. send a "599 Broken" HTTP status to the client. This is intended to help to distinguish between this situation and a 500, which could mean that an app failed the processing but ownCloud still works.
       2. log the exceptions in the "crash-Y-m-d.log" ("crash-2019-11-13.log", for example) file.
    * For CLI:
       1. log the exceptions in the "crash-Y-m-d.log" file (same as above)
       2. log the original exception in the console.

This new crash log file will be generated in the configured data directory, unless the `crashdirectory` is present in the config.php file. If it's present, that directory must exists and must be writable by the web server.

ho4ho:
Really good idea by using 599 for status code?

> 599 Network Connect Timeout Error
>
> This status code is not specified in any RFCs, but is used by some HTTP proxies to signal a network connect timeout behind the proxy to a client in front of the proxy.
>
> https://httpstatuses.com/599
> https://github.com/rmaake1/httpstatuses/issues/22

jvillafanez:
I checked the code list from mozilla, and the 599 didn't appear... it might not be the best code.
I'm opened to suggestions, but I'd like to avoid 500 and 503 in order to distinguish this particular case from other "normal" errors.

sharidas (commented):
Haven't tested. But code wise looks ok to me.

micbar (changes requested):
please add a changelog item

micbar (approved):

phil-davis:
@micbar should this be added to 10.4.0 project?
It might be difficult to add acceptance tests, because we would have to induce a server error to make code paths happen. Existing CI being green is good!

micbar:
@phil-davis 10.4 is ok.
IMO this could be done via Changelog testing.

phil-davis:
Someone needs to decide what to do about the codecov result.

VicDeo:
@micbar  
ErrorHandler: only removed lines, adding tests won't increase coverage (and the method is static too)
Setup.php: 1 line was added into a static method with no proper DI
base.php: static method once again technically it is possible to cover `crashLog` method but it would be not a classic unit-test.


the rest is too low-level for unit tests
console.php 
index.php 
public.php 
remote.php 
status.php

micbar (approved):
Accepted the PR, Needs a manual Release test during QA phase.
@davitol Please post an example of the crashlog contents.

phil-davis (approved):
LGTM - as in the discussion, automated tests are not realistic.
Someone with privs needs to override the codecov result and merge.

PR #36422, opened 2019-11-13
[WIP] Enable samba tests 20191113

phil-davis:
Try to get samba "unit" tests passing again

PR #36424, opened 2019-11-13
[WIP] Enable samba tests without 'domain WORKGROUP'

phil-davis:
This is on top of #36422 and is for test confirmation only.

It removes the `'domain'=>'WORKGROUP'` from the samba configs. This should make the tests fail, confirming that `'domain'=>'WORKGROUP'` really is essential.


phil-davis:
"good" - https://drone.owncloud.com/owncloud/core/21341/3/10
```
There were 85 failures:

1) OCA\Files_External\Tests\Storage\SmbTest::testRenameWithSpaces
Failed to create a root dir

/drone/src/apps/files_external/tests/Storage/SmbTest.php:52
```

So `domain WORKGROUP` is essential.

PR #36431, opened 2019-11-17
fix user search problem happening after user deletion

karakayasemi:
## Description
In the web-UI user management page, When user make a search, if the search result has only one entry and the listed user is deleted from the search result page, the search does not work anymore until refreshing the page.

Reason for the problem, Js search function uses a hidden <tr> column to clone new rows from it. In our scenario, after user deletion, this row also being deleted. As a result, search functionality can not add a new row to the user list.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3585

## Motivation and Context
Fixing bugs.

## How Has This Been Tested?
Manually with the following steps:
- Open Menu Users
- Searching for a user (search result should be match with only one entry)
- Delete a user
- Make a new search for another user did not work, no result for the search.

This problem now fixed.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/36431
+++ b/changelog/unreleased/36431
@@ -1,0 +1,5 @@
+Bugfix: Fix user search problem happening after user deletion
+
+After a user search in user management web-UI, if the search result has a single user entry and afterward the user was deleted from the interface, the search was no longer work until refreshing the page. This bug has been fixed.
+
+https://github.com/owncloud/core/pull/36431
--- a/settings/js/users/users.js
+++ b/settings/js/users/users.js
@@ -177,10 +177,7 @@ var UserList = {
 		 */
 		$tr.appendTo($userList);
 		if(UserList.isEmpty === true) {
-			//when the list was emptied, one row was left, necessary to keep
-			//add working and the layout unbroken. We need to remove this item
 			$tr.show();
-			$userListBody.find('tr:first').remove();
 			UserList.isEmpty = false;
 			UserList.checkUsersToLoad();
 		}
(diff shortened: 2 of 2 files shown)


jvillafanez:
I'm worried about the deleted comment... could you check / investigate it?

karakayasemi:
@jvillafanez I agree about your concern too, I spent lots of time to debug and understand the code. But I could not find the purpose of removing `<'tr:first'>`. As line 180 says, one row should always be left to keep working the functionality of adding. The source of the problem of the related issue is also this remove line. Because when we delete the last row, new search results can not add new items.

I will investigate if this issue is a regression. It may help.

jvillafanez:
Just trying to give some ideas:
* Add + remove + add + remove + add?
* Small / mobile screens

You could also try with "git blame" to know when the comment was added.

karakayasemi:
@jvillafanez I investigated the code with more details. The deleted comment belongs to the following commit https://github.com/owncloud/core/commit/478393e15ad3d6c884368f0b1a922eda6f9ffe0e which looks like merged 6 years ago with the re-write of the user management page. Also, I tried the issue with older oc versions, the issue is not a regression, probably its scenario is reproducible from the beginning. 

I made several tests with consecutive adding-removing-searching operations on the PR's branch and could not see any unexpected behavior. In terms of code, if there is a hidden `<tr>` element for cloning new rows, I see no reason for removing it in add new row method.  🤷‍♂

jvillafanez (approved):

jvillafanez:
One last thing: double-check the generated html code make sense, without any wrong element added (you can use the browser's developer tools)

karakayasemi:
The HTML code looks fine. Also, no error log in the browser console. @enbrnz before merging this one, I would like to get your approval that the problem has been solved and that there are no visible problems.
I will also add a changelog item.

enbrnz:
Hi, sorry for the delay. So I created a patch from this PR and applied it to a docker-compose setup. However it doesn't seem to have fixed the issue, as far as I can tell the behavior hasn't changed.

PR #36433, opened 2019-11-18
[DO NOT MERGE] adjust timestamps in VersioningTest

phil-davis:
## Description
`VersioningTest` is failing intermittently in `files_primary_s3` - see comment https://github.com/owncloud/core/issues/36266#issuecomment-554883019

I seems to fail "about half the time". The tests create "fake" versions at the current Unix timestamp and for 14 days ago. That creates a "version" "right now". If the test goes on to create a new file, or do "something interesting" then it could try to create another version of the file, which might sometimes happen in the same "unix second" and sometimes in the next "unix second". We know that `files_versions` does not support multiple versions in the same second.

In another test case, the timestamp of the latest version was being set to 60 seconds ago. We may as well do the same in all test cases (we could just make in 1 second ago).

This PR adjusts the time of the latest file version to be 60 seconds ago.

I am not sure why these test cases all seem to pass reliably in core, but 1 of them fails in `files_primary_s3`. So the `files_primary_s3` cause might be deeper than the fix here.

## Related Issue
- #36266 

## Motivation and Context
Make tests reliable.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_versions/tests/VersioningTest.php
+++ b/apps/files_versions/tests/VersioningTest.php
@@ -160,8 +160,10 @@ class VersioningTest extends TestCase {
 		$v1 = $versionsFolder2 . '/test.txt.v' . $t1;
 		$v2 = $versionsFolder2 . '/test.txt.v' . $t2;
 
-		$this->rootView->file_put_contents($v1, 'version1');
 		$this->rootView->file_put_contents($v2, 'version2');
+		// do not write both files in the same second
+		\sleep(1);
+		$this->rootView->file_put_contents($v1, 'version1');
 
 		// move file into the shared folder as recipient
 		\OC\Files\Filesystem::rename('/test.txt', '/folder1/test.txt');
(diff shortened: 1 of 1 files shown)


phil-davis:
@jvillafanez I will run the `files_primary_s3` CI against this branch to see if it helps.

In any case, IMO this change is "a good thing" to do.

jvillafanez:
* For local FS you can create a "f.v000001" without problems. The correct data is inserted both in the DB and FS.
* For S3 with versioning, the versions are completely handled by the S3 storage. I don't think the S3 storage has any problem saving several versions within the same second (the order might be different depending on which version comes first, but it shouldn't be critical)
* For S3 without versioning, the problem is that the DB might contain different information than what is stored in S3. The DB might reject the version because is made in the same second, but S3 might not.

As much as I hate it, the alternative (if this PR doesn't work consistently) could be to add a delay of 1 second in the tests.
Completely changing the versioning model to another one more robust doesn't seem possible at this late stage.

phil-davis:
I am not getting any success in files_primary_s3 with this core change.
So there is some different reason that the test fails (about half the time)

PR #36436, opened 2019-11-18
Add API acceptance tests for moving in and out of a shared folder

phil-davis:
## Description
Add API acceptance test scenarios to cover moving a file and/or folder into and out of a shared folder, both as the sharee and as the sharer.

## Related Issue
- related to #36266 

## Motivation and Context
The current scenarios in `moveFile.feature` cover "ordinary" moving, and attempts to move into a share when the sharee does not have enough permission. But I don't see scenarios anywhere for "ordinary" moves in and out of shares when the sharee does have permission.

I looked for these because there are unit tests failing in `files_primary_s3` related to this sort of thing (when the sharer and sharee end up with different S3 storage buckets). So it would be useful to have these test scenarios in core, and they can get run from `files_primary_s3`

## How Has This Been Tested?
Local run of new test scenarios

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavMove/moveFile.feature
+++ b/tests/acceptance/features/apiWebdavMove/moveFile.feature
@@ -68,7 +68,97 @@ Feature: move (rename) file
       | old         |
       | new         |
 
-  Scenario Outline: Moving a file to a folder with no permissions
+  Scenario Outline: Moving a file into a shared folder as the sharee and as the sharer
+    Given using <dav_version> DAV path
+    And user "user1" has been created with default attributes and without skeleton files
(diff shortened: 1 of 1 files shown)


phil-davis:
@individual-it I couldn't find this in the existing API acceptance tests. The `apiShare*` suites are doing all sorts of upload/download with shares with different permissions. reshares... But they do not seem to do file/folder moves in and out of the shares.

individual-it (approved):

PR #36458, opened 2019-11-22
Test with disabling public upload

### LINKED ISSUE #36442: When 'Allow public uploads' is disabled cannot create share with permission 5 but creating share with permission 7 works (fallback to permission 1)
### Steps to reproduce
1. disable "Allow public uploads"
2. create a public link with permission 5 ` curl 'http://172.17.0.1/owncloud-core/ocs/v2.php/apps/files_sharing/api/v1/shares' -X POST -u admin:admin --data 'permissions=5&name=Public+link&shareType=3&path=/ssss`
3. do the same with permission 7

### Expected behaviour
both cases should behave the same

### Actual behaviour
with permission 5: 
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>403</statuscode>
  <message>Public upload not allowed</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data/>
</ocs>
```

with permission 7 the link is created but the permission defaults to 1

skshetry:
## Description

This PR tests creating public links with public upload disallowed.

## Related Issue

- Fixes #36442

## How Has This Been Tested?

- locally and CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -425,6 +425,50 @@ Feature: sharing
       | 1               | 100             |
       | 2               | 200             |
 
+  @issue-36442 @public_link_share-feature-required
+  Scenario Outline: Creating a public link share with read+create permissions defaults to read permissions when public upload disabled globally
+    Given using OCS API version "<ocs_api_version>"
+    And parameter "shareapi_allow_public_upload" of app "core" has been set to "no"
+    And user "user0" has created folder "/afolder"
+    When user "user0" creates a public link share using the sharing API with settings
+      | path        | /afolder    |
+      | permissions | read,create |
+    # And the fields of the last response should include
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
LGTM

There is a question about which scenario behaves "correctly" and which behaves "badly". I asked about that in the issue https://github.com/owncloud/core/issues/36442#issuecomment-557405829

The answer does not matter so much. If the answer is the opposite to what is suggested in this PR, then the developer will just have to make different adjustments to the 2 scenarios.

skshetry:
@phil-davis, yes, the expected behavior is not clear. But, I'd expect the permission to be`read` only, given that, for example, if we create a public link with permission 31, we only receive read. So, that case should hold here, as well imo.

skshetry:
Sorry, i messed up. Fixing

PR #36460, opened 2019-11-22
Revert "Objectstore transfer"

phil-davis:
Reverts owncloud/core#36326

This let's us back-up a step of code and try again at the initial problem of issue #36266 

See PR #36462 for a possible thing to do next in working-around the test fail on multi-bucket object storage.

### DIFF
--- a/apps/files/tests/Command/TransferOwnershipTest.php
+++ b/apps/files/tests/Command/TransferOwnershipTest.php
@@ -163,15 +163,6 @@ class TransferOwnershipTest extends TestCase {
 			->setShareType(Share::SHARE_TYPE_USER)
 			->setPermissions(19);
 		$this->shareManager->createShare($share);
-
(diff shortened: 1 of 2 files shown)


phil-davis:
The problem with reverting this is that it fixed a different test fail.
https://drone.owncloud.com/owncloud/files_primary_s3/1296/3/7
```
There was 1 failure:

1) OCA\Files\Tests\Command\TransferOwnershipTest::testTransferAllFiles
Failed asserting that actual size 3 matches expected size 0.

/var/www/owncloud/server/apps/files/tests/Command/TransferOwnershipTest.php:180
```

So if we do this revert then we get back the original problem that was reported in https://github.com/owncloud/core/issues/36266#issue-505618712

@micbar @jvillafanez @karakayasemi

micbar (approved):

PR #36462, opened 2019-11-22
Skip testTransferAllFiles when on ObjectStoreStorage

phil-davis:
## Description
This is on top of PR #36460

In addition to reverting "Objectstore transfer", also add code to skip the failing `testTransferAllFiles` test when on `ObjectStoreStorage`.

This skips the test case on any `ObjectStoreStorage` - it would be even better if there is a way to detect multi-bucket `ObjectStoreStorage` and only skip on that.

## Related Issue
- issue #36266 

## How Has This Been Tested?
CI - I will check with files_primary_s3

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/tests/Command/TransferOwnershipTest.php
+++ b/apps/files/tests/Command/TransferOwnershipTest.php
@@ -22,9 +22,12 @@
 namespace OCA\Files\Tests\Command;
 
 use OC\Encryption\Manager;
+use OC\Files\ObjectStore\ObjectStoreStorage;
+use OC\Files\View;
 use OC\Share20\ProviderFactory;
 use OCA\Files\Command\TransferOwnership;
 use OCP\Files\Mount\IMountManager;
+use OCP\Files\Storage;
 use OCP\IUser;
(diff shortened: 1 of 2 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/files_primary_s3/1297/3/7
The "problem" test gets skipped in the ObjectStorage environment. That allows `files_primary_s3` CI  to pass.

However, the test is being skipped for all ObjectStorage - both single and multi-bucket.

It would be even better to be able to detect when it is in a multi-bucket environment and only skip the test then (i.e. let it run for the single bucket environment)

Is there good code that could find that out at run-time?
Or do we have to do something like define some env var that we can check for to "guess" the environment? 
(and set that env var in the drone scripts before the appropriate test runs)

micbar (approved):

karakayasemi (approved):
IMHO, it's better to continue with this one instead of only reverting old commit.

PR #36476, opened 2019-11-26
Bump readme.md file to 10.3 version

davitol:
Bump readme.md file content links to 10.3 version


phil-davis (approved):
Does README.md end up in the released tarball?
If not, then we do not really need to make this change in the release branch (but it does not hurt)

davitol:
> Does README.md end up in the released tarball?

Yes, it is included

micbar (approved):

PR #36483, opened 2019-11-27
Change the Changelog  Template

micbar:
# Description
Improved the Changelog Template
 - Use the long type 
 - Indent the list items in the "Details" correctly.


.htaccess
```diff
@@ -92,3 +92,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
+
+ErrorDocument 403 /core/templates/403.php
+ErrorDocument 404 /core/templates/404.php
```
phil-davis: These lines need to be removed.

jnweiger (approved):
Two minor issues:
* There is a lonely closing parenthesis at the end of the details section in the generated CHANGELOG.md -- I don't see anything in the template that would cause it.

* The diffs listing at the end is not shown in HTML. It looks correct to me in CHANGELOG.md

Approved nevertheless.

Issue #29483, opened 2017-11-07
MariaDB 10.3 compatibility in regard to mb4


PVince81:
@VicDeo mentioned that MariaDB 10.3 will get rid of the innodb table format switch.

The question is whether this will make the DB default to always use mb4 or not.

We'll need to:
1) find a alternative for the missing config switch
2) if MariaDB 10.3 defaults to mb4 or to the other table format, check whether we need to provide a migration for all the data.

Let's do the research first before deciding on what to do.

@DeepDiver1975 @VicDeo

VicDeo:
Meanwhile MariaDB 10.3.3 with **Beta** status was released on 23 Dec 2017.
IMO this is something for 10.1

PVince81:
Agreed, moving to planned.

Do you have a rough idea about how much work this will be ? (estimate)

ownclouders:
Hey, this issue has been closed because the label `status/STALE` is set and there were no updates for 7 days. Feel free to reopen this issue if you deem it appropriate.

(This is an automated comment from [GitMate.io](https://gitmate.io).)

VicDeo:
We're getting closer....
https://mariadb.com/kb/en/library/release-notes-mariadb-103-series/
MariaDB 10.3.5 
**Status: RC** | Release Date: 26 Feb 2018

PVince81:
moving to planned

with a bit of luck no changes are needed on the code and only testing needed?

VicDeo:
I wouldn't count on bit of gluck here... We had unit tests with 10.3 when it was in alpha and they failed.

To revive the essence of  the issue:
`innodb_file_format` option is used to detect utfmb4 here 
https://github.com/owncloud/core/blob/92b658a60e6c28ef9dd7594384180e0d73e96660/lib/private/DB/MySqlTools.php#L35-L36

According to https://mariadb.com/kb/en/library/xtradbinnodb-server-system-variables/#innodb_file_format 
it's always barracuda since MariaDB 10.0.2.2
```
Deprecated: MariaDB 10.2
Removed: MariaDB 10.3.1
```

VicDeo:
so, for MariaDB 10.3 it's enough to bypass checking this option.
But there could be more surprises of a different kind.

PVince81:
something to look into for OC 11 while we upgrade libs and everything

PVince81:
@VicDeo any action points ? if yes, would be good to estimate and we can put this in the next sprint.

[...]

PR #36493, opened 2019-11-29
[Test] with mariadb:10.3

phil-davis:
Run various unit and acceptance tests with MariaDb 10.3 to verify they pass.

Part of issue #29483


phil-davis:
Looks good. The test pipelines are all using mariadb:10.3 and they pass.

PR #36496, opened 2019-11-29
Use ICacheIterator to retrieve ICacheEntry child entries

mrow4a:
Background and motivation: 
- in case of UI or user facing requests it is better to optimize for response time (which usually is expected to last few seconds), thus higher memory footprint and more optimized access to the database
- in case of background jobs like versions and trashbin, one needs to be careful with memory footprint, as these php jobs in specific cases might take many minutes. this can cause memory allocation to explode in a corner cases. 

One approach to above problem is to use optimized access for UI and short requests (get all contents at once and cache as in `getDirectoryContents->array`) where memory is released quickly and access patterns are usually not exotic, while for long-running and data wrangling jobs retrieve filecache entries in a memory-constrained way (`getChildren->Iterator`) ). 

Related to issue https://github.com/owncloud/enterprise/issues/2925

**Results for const memory iteration - max 512 records per batch:**
getChildren memory before : used 22237592 \/ allocated 23068672 
getChildren(/test) returned 9601 entries took 0.17817783355713 memory after : used 22760288 / allocated 25165824 B

**Result for putting all entries in memory - 9600 records in 1 chunk:**
getFolderContents memory before : used 22237744 / allocated 23068672 B
getFolderContents(/test) returned 9601 entries took 0.10629200935364 memory after : used 33988656 / allocated 44040192 B

Thus, `getChildren(/test)` used 3MB of memory for iterating over 9600 filecache entries in the folder and took 0.18s, while `getFolderContents(/test)` used 21 MB of memory for iterating over 9600 filecache entries in the folder and took 0.11s.

### DIFF
--- a/lib/private/Files/Cache/CacheChildrenIterator.php
+++ b/lib/private/Files/Cache/CacheChildrenIterator.php
@@ -1,0 +1,224 @@
+<?php
+/**
+ * @author Piotr Mrowczynski <piotr@owncloud.com>
+ *
+ * @copyright Copyright (c) 2019, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Files\Cache;
+
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\Files\Cache\ICacheEntry;
+use OCP\Files\Cache\ICacheIterator;
+use OCP\Files\IMimeTypeLoader;
+use OCP\IDBConnection;
+use OCP\ILogger;
+
+/**
+ * This is implementation of ICacheIterator for const memory retrieval of
+ * ICacheEntries for children of file with given $fileId (children is defined as
+ * having parent field with the $fileId.
+ *
+ * @since 10.3.0
+ */
+class CacheChildrenIterator implements ICacheIterator {
+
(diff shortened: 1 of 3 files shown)


lib/private/Files/Cache/CacheChildrenIterator.php
```diff
@@ -0,0 +1,233 @@
...
+	 * @var int
+	 */
+	private $fileId;
+
+	/**
+	 * @param int $fileId the file id of the folder
+	 * @param int $cacheCapacity cache capacity, disabled if -1
+	 */
+	public function __construct($fileId, $cacheCapacity = 128) {
+		$this->mimetypeLoader = \OC::$server->getMimeTypeLoader();
```
DeepDiver1975: Inject please

DeepDiver1975 (changes requested):
No generator? Am i looking at the wrong PR?

mrow4a:
> No generator? Am i looking at the wrong PR?

@DeepDiver1975 @jvillafanez there are few reasons why not to use generator after my investigation:
- if you run query with `LIMIT ~ OFFSET 0`, if you use generator to do `fetch()`, first call will return to the driver WHOLE result anyways - https://www.php.net/manual/en/pdostatement.fetch.php (cursor orientation and scrollable cursor behaviour). To prevent allocating whole RESULT space in memory, we would need to use Scrolling Cursor enabled, which is not available e.g. for MySQL as this cursor type would allow scroll also backwards. You will also need to be able to manipulate PDO configuration, which is not possible in current implementation of `Database Connection` class.
- because of no Scrolling Cursor, the only way to achieve constant memory is by fetching data in batches of const memory using `LIMIT 512 OFFSET 0 where fileid > [last-seen-fileid]`. This is why I decided on implementing `\Iterator` interface, which in the background fetches data in batches  when triggered on certain Iterator calls to `iterator->next()`.  
- `\Iterator` forces developer to implement properly interfaces responsible for iterating over single record of the data (batching in memory), while generator you will need to do hacks to retrieve queries after 512 records.
- `\Iterator` class in php is inheriting `\Traversable`, can be iterated with foreach, like a normal array (cannot however be used with index access, count and other ArrayAccess interfaces as this is exactly what we want to prevent).

DeepDiver1975:
I see - thanks for clarification! 

I'd still suggest to add all this to a new method and not change the current impl of Cache::getFolderContentsById - just as a risk prevention action.

mrow4a:
Closing, would require too high refactor, and only few places could benefit

PR #36497, opened 2019-11-29
#36490 changelog extend

ho4ho:
#36490 include code done by https://github.com/sabre-io/http/pull/119, think worth mentioned excplicite:

> Directly reading a large file (let’s say 10GiB) on that server will get a read speed around 540-550 MiB/s. However the download speed with webdav is only about 240-260 MiB/s.

and

> After applying this PR, I got full download speed about just like reading that file locally (even faster than fpassthru).

### DIFF
--- a/changelog/unreleased/36490
+++ b/changelog/unreleased/36490
@@ -1,5 +1,5 @@
 Change: Update sabre/http (5.0.2 => 5.0.5)
 
-Includes functionality to ignificantly improve file download speed by enabling mmap based stream_copy_to_stream.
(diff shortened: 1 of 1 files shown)


changelog/unreleased/36490
```diff
@@ -1,3 +1,5 @@
 Change: Update sabre/http (5.0.2 => 5.0.5)
 
+Includes functionality to ignificantly improve file download speed by enabling mmap based stream_copy_to_stream.
```
phil-davis:
```suggestion
Includes functionality to significantly improve file download speed by enabling mmap based stream_copy_to_stream.
```

Issue #33187, opened 2018-10-15
Postgres 10 support


PVince81:
From https://github.com/owncloud/core/issues/29997.

- [ ] update to dbal v2.8 on master
- [ ] setup Drone target for Postgres 10
- [ ] fix all found issues, if any

PVince81:
we already have the drone target https://github.com/owncloud/core/blob/master/.drone.yml#L622, closing

gramakri:
@PVince81 Will Postgres 10 be supported in oc 10.1.0 or just oc 11? When can we expect oc 11 (I don't see a beta even).

phil-davis:
Note: https://github.com/doctrine/dbal/releases/tag/v2.6.0
"PHP 7.1.0 is the minimum supported PHP version"

core `stable10` is dropping PHP 5.6 support, but for now keeps support for PHP 7.0. So `composer` in `stable10` finds `doctrine/dbal` `v2.5.13` - it can't yet use `doctrine/dbal` `v2.6.0` or above.

In particular, we can't get to `doctrine/dbal` `v2.8.0` in `stable10` just yet :(

phil-davis:
Note: PR #31059 was also an essential part of Postgres 10 support.

phil-davis:
Now that core `master` is the old core `stable10` we no longer have the PostgreSQL 10 support code sitting in core `master`. It is now waiting in PR #35962

PostgreSQL 10 needs minimum PHP 7.1. So getting this into a release is blocked until we can drop PHP  7.0.

Reopening the issue so it gains visibility again.

phil-davis:
Code for this is in the commits waiting in #36289

phil-davis:
PR #36498 runs the full test suite with postgres:10.3 as the database and passes.

phil-davis:
PR #36290 was merged. This is now in master for release in 10.4

PR #36498, opened 2019-11-30
Test postgres:10.3

phil-davis:
to see that CI passes now we have dropped PHP 7.0

ref issue #33187


phil-davis:
Tests all pass - good.

Issue #32670, opened 2018-09-11
Cannot store complex custom Webdav property values + wrong values + Oracle failure


PVince81:
### Steps

1. Create a file "proppatch.xml" with the contents `<?xml version="1.0" encoding="utf-8" ?><propertyupdate xmlns='DAV:'><set><prop><t:valnspace xmlns:t='http://example.com/neon/litmus/'><foo xmlns='http://bar'/></t:valnspace></prop></set></propertyupdate>` which sets a complex value
1. Create a folder "test" in the OC instance
1. `curl -k -D - -u admin:admin -X PROPPATCH -H "Content-Type: text/xml" --data-binary "@proppatch.xml" "http://localhost/owncloud/remote.php/webdav/test"`
1. Create a file "propfind.xml" with the contents `<?xml version="1.0"?><a:propfind xmlns:a="DAV:" xmlns:oc="http://owncloud.org/ns"><a:prop><t:valnspace xmlns:t='http://example.com/neon/litmus/'/>        </a:prop></a:propfind>`
1. `curl -k -D - -u admin:admin -X PROPFIND -H "Content-Type: text/xml" --data-binary "@propfind-oracle.xml" "http://localhost/owncloud/remote.php/webdav/test"`
1. Check the "oc_properties" database table

### Expected result
oc_properties table must contain `<foo xmlns='http://bar'/>` or something alike.
the PROPFIND response should contain `<foo xmlns='http://bar'/>`.

#### With MySQL
oc_properties table contain the string "Object" as value.
the PROPFIND response contains `<x1:valnspace xmlns:x1="http://example.com/neon/litmus/">Object</x1:valnspace>`

#### With Oracle
PROPPATCH fails:
```

[...]

PR #36500, opened 2019-11-30
Php7.4 20191130

phil-davis:
Try running tests with PHP 7.4


phil-davis:
This has lots of `phpunit` messages that will need to be looked into. So it is not a simple 10-minute job for this week!

phil-davis:
litmus and webDAV problem with object __toString conversion exception - see https://github.com/owncloud/core/issues/32670#issuecomment-565080211

phil-davis:
see #36670

PR #36506, opened 2019-12-02
Make settings page work in Phoenix

PVince81:
## Description
Adjust CSP rules to allow iframing.
Receive "phoenix=true" URL parameter to signal that we're coming from
Phoenix.
Remove page header when in Phoenix mode.
Use different URLs for navigation sections.

## Related Issue
For https://github.com/owncloud/phoenix/issues/2598

## Motivation and Context
Make settings page accessible without leaving Phoenix frontend

## How Has This Been Tested?
Manual test, try switching sections, etc

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/css/styles.css
+++ b/core/css/styles.css
@@ -170,6 +170,9 @@ a.two-factor-cancel {
 #content .hascontrols {
 	margin-top: 45px;
 }
+.phoenix #content-wrapper {
+	padding-top: 0px;
+}
 #content-wrapper {
 	position: absolute;
 	height: 100%;
--- a/core/templates/layout.user.php
+++ b/core/templates/layout.user.php
@@ -35,9 +35,17 @@
 			<script src="<?php print_unescaped($jsfile); ?>"></script>
 		<?php endforeach; ?>
 		<?php print_unescaped($_['headers']); ?>
+		<?php if (isset($_['isPhoenix']) && $_['isPhoenix']): ?>
+			<base target="_parent" />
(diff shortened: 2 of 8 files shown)


PVince81:
The links in the navigation have been modified in this PR to point back to Phoenix, if the phoenix param was passed.

I was thinking that maybe we rather need a Router decorator that rewrites OC URLs to Phoenix, this way we're able to better catch those routes. Route rewriting will also need to work on the JS side...

PVince81:
- [ ] find a way to redirect to Phoenix when opening settings page URL directly ? (like from the desktop client)

DeepDiver1975 (changes requested):
To be honest i dislike the approach to circumvent csp by supplying a request Parameter.
It actually breaks the idea of csp.
With this we allow any Client to bypass iframing....

PVince81:
> To be honest i dislike the approach to circumvent csp by supplying a request Parameter.
> It actually breaks the idea of csp.
> With this we allow any Client to bypass iframing....

I considered that approach and decided against it already.
The CSP rule condition is only based on whether Phoenix exists in the config.php, so we can use that URL and add it to the CSP whitelist

PVince81:
Abandoned, see https://github.com/owncloud/phoenix/issues/2598#issuecomment-563932184

PR #36511, opened 2019-12-03
Behat bootstrap.php does not need core lib/composer/autoload.php

phil-davis:
## Description
The `bootstrap.php` in the acceptance tests folders is including the core `lib/composer/autoload.php` but actually the acceptance tests should be able to run independently of the system-under-test. They should not need to suck in some core composer dependency autoloader stuff.

In apps it would be good to minimize the dependencies that we have to install when setting up the test-runner. The apps need to find Gherkin step definitions/code from core, so they do need to use the core Behat `bootstrap.php`. But if that also includes `lib/composer/autoload.php` then we also have to do a full `composer install` of all the core stuff (when we only want the minimum needed ffor the test-runner)

I ran across this while doing some user_ldap tests and trying to get them to start quicker.

## How Has This Been Tested?
Local run of some acceptance tests - they still pass

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/bootstrap.php
+++ b/tests/acceptance/features/bootstrap/bootstrap.php
@@ -20,8 +20,6 @@
  *
  */
 
-require __DIR__ . '/../../../../lib/composer/autoload.php';
-
 $classLoader = new \Composer\Autoload\ClassLoader();
 $classLoader->addPsr4("Page\\", __DIR__ . "/../lib", true);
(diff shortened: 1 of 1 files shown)


phil-davis:
@individual-it do you have any idea why we used to need to include autoloader.php?
Or maybe it is a piece of history from the long-distant past.

individual-it (approved):

PR #36513, opened 2019-12-03
Remove dependency of acceptance tests on core lib/composer/autoload.php

phil-davis:
## Description
- Revert #36511 - I didn't do the full/proper job there.

-  Delete all `require __DIR__ . '/../../../../lib/composer/autoload.php';` from acceptance test code

- Add to `vendor-bin/behat/composer.json` the dependencies that the Behat acceptance test code was relying on from core `lib/composer` :
- `sabre/xml` `^2.1` - the acceptance test code makes use of this library when parsing XML in responses
- `guzzlehttp/guzzle` `^5.3` - the acceptance test code (as well as core code) still has guzzle v5-style code
- `phpunit/phpunit` `^7.5` - the acceptance test code makes use of `phpunit` `Assert`. That is an integrated part of `phpunit`, so we need to depend on the whole `phpunit`

- acceptance test code was finding out the share state constants directly from backend code `\OCP\Share::STATE_PENDING` etc. That is not a good thing anyway. Move the acceptance test knowledge about those numbers into `SharingHelper` because that is external behaviour and actually we want to know if the values of the numbers to be used in API calls are changed in the backend.

## Issue
#36514

## Motivation and Context
- have the acceptance test code less dependent on the core code
- have the possibility (e.g. in app repos) to have just the required Behat dependencies installed in the test runner, independently of the system-under-test (at the moment we get forced to do a `composer install` of the whole of core in the test-runner directory, which should not be necessary)

## How Has This Been Tested?
Local test run plus CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/SharingHelper.php
+++ b/tests/TestHelpers/SharingHelper.php
@@ -45,6 +45,13 @@ class SharingHelper {
 			'federated' => 6,
 	];
 
+	const SHARE_STATES = [
+			'accepted' => 0,
+			'pending' => 1,
+			'rejected' => 2,
+			'declined' => 2, // declined is a synonym for rejected
+	];
+
 	/**
 	 *
 	 * @param string $baseUrl baseURL of the ownCloud installation without /ocs.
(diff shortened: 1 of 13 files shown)


phil-davis:
Note: do not merge yet - apps with acceptance tests also have to have the same extra dependencies add to their `vendor-bin/behat/composer.json`

So this will have to be co-ordinated.

phil-davis:
@individual-it this should be OK to review-approve-merge now.

individual-it (approved):

PR #36523, opened 2019-12-04
Fix warning when resetting user's password with masterkey encryption

pako81:
## Description

Suppress warning when resetting user's password over the Users management page with masterkey encryption enabled.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/3647

## Motivation and Context

This PR fixes the following warning displayed when the admin is about to reset user's password over the Users management page with masterkey encryption in place:

70141012-5d090880-1696-11ea-90ad-476b4b1d648b

The warning itself does not prevent resetting user's password but it is scary and makes no sense when masterkey encryption is in use.

## How Has This Been Tested?

Manually by enabling masterkey encryption and as admin hovering over the password field for one random user: the warning is not displayed anymore.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Changelog item

### DIFF
--- a/settings/Controller/UsersController.php
+++ b/settings/Controller/UsersController.php
@@ -84,8 +84,6 @@ class UsersController extends Controller {
 	private $fromMailAddress;
 	/** @var IURLGenerator */
 	private $urlGenerator;
-	/** @var bool contains the state of the encryption app */
-	private $isEncryptionAppEnabled;
 	/** @var bool contains the state of the admin recovery setting */
 	private $isRestoreEnabled = false;
 	/** @var IAvatarManager */
@@ -153,8 +151,7 @@ class UsersController extends Controller {
 		$this->eventDispatcher = $eventDispatcher;
 
 		// check for encryption state - TODO see formatUserForIndex
-		$this->isEncryptionAppEnabled = $appManager->isEnabledForUser('encryption');
-		if ($this->isEncryptionAppEnabled) {
+		if ($appManager->isEnabledForUser('encryption')) {
 			// putting this directly in empty is possible in PHP 5.5+
 			$result = $config->getAppValue('encryption', 'recoveryAdminEnabled', 0);
 			$this->isRestoreEnabled = !empty($result);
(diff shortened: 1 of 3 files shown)


phil-davis:
Unit test expectations also need adjusting: https://drone.owncloud.com/owncloud/core/21714/13/9

settings/Controller/UsersController.php
```diff
@@ -186,6 +186,9 @@ private function formatUserForIndex(IUser $user, array $userGroups = null) {
 					// user also has recovery mode enabled
 					$restorePossible = true;
 				}
+			} else {
```
sharidas: Lots of if's and else would bring down the readability. How about changing this code to:
```
if ($this->isRestoreEnabled) {
	$recoveryMode = $this->config->getUserValue($user->getUID(), 'encryption', 'recoveryEnabled', '0');
	// method call inside empty is possible with PHP 5.5+
	$recoveryModeEnabled = !empty($recoveryMode);
	if ($recoveryModeEnabled) {
		// user also has recovery mode enabled
		$restorePossible = true;
	}
} else {
	$restorePossible = true;
}
```

pako81: @sharidas thanks for the suggestion! I think we can indeed get rid of the `if (this->isEncryptionAppEnabled)`check since `if ($this->isRestoreEnabled)` is already performing the same. And will help us to get a better code readability.

Got also rid of any reference to `isEncryptionAppEnabled ` from code.

sharidas:
There was a semi colon added in the proposesd change, due to which the tests were failing. I have removed the semi colon and now the tests should pass.

sharidas (commented):
Now the code looks good to me. I have not tested this. If it works as expected, then Ok from my side.

phil-davis:
I rebased just now to make sure that CI is still good for this.

@micbar is this to be released?
Needs someone to review/approve.
I added to 10.4 project so it gets noticed - otherwise these PRs seem to just sit and nobody looks at them.

sharidas:
#### Tests done 
- With masterkey encryption enabled
   - Create a user in the users page and try to reset the password for the new user. The message related to data recovery is not shown.

- With user keys encryption
   - Create a user in the users page and try to enable the recovery password for the user.
       - Now try to reset the password. There is no message displayed related to data recovery.
   - Create another user in the users page and do not enable the recovery password for the user
       - Now try to reset the password. There is a message that appears. This is expected behaviour.

- Tested without encryption
    - All users created in the users page can reset their password gracefully and the data recovery message does not pops up. This behaves as expected.

sharidas (approved):
LGTM :+1:

pako81:
thanks @phil-davis @sharidas @micbar

PR #36526, opened 2019-12-05
Add sample OIDC configuration to the sample app config file

settermjd:
## Description

Add sample OIDC (OpenID Connect) configuration to the sample app config file

## Related Issue

- Relates to https://github.com/owncloud/docs/issues/2100

## How Has This Been Tested?

- 

## Types of changes

- [ ] Technical debt
- [X] Documentation

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/config/config.apps.sample.php
+++ b/config/config.apps.sample.php
@@ -33,7 +33,7 @@ $CONFIG = [
  *
  * Possible values: `activity_expire_days` days.
  */
- 
+
 /**
  * Define the retention for activities of the activity app
  */
@@ -47,7 +47,7 @@ $CONFIG = [
  *
  * Possible values: `user_ldap.enable_medial_search` `true` or `false`.
  */
- 
+
 /**
  * Define parameters for the LDAP app
  */
@@ -82,7 +82,7 @@ $CONFIG = [
  * Configuring the download links for ownCloud clients,
  * as seen in the first-run wizard and on Personal pages
  */
- 
+
 'customclient_desktop' =>
 	'https://owncloud.org/install/#install-clients',
 'customclient_android' =>
@@ -101,5 +101,64 @@ $CONFIG = [
  */
 
 'collabora_group' => '',
+
+/**
+ * OpenID Connect(OIDC) Configuration
+ *
+ * App: openidconnect
+ */
+
+/**
+ * Configure OpenID Connect
+ *
+ * The `provider-url`, `client-id` and `client-secret` variables are to be
+ * taken from the OpenID Connect Provider's setup. The `loginButtonName`
+ * variable can be freely chosen, depending on the installation.
+ *
+ * NOTE: The provider-params configuration array only needs to be used if the
+ * OpenID Connect Provider does NOT support service discovery.
+ *
+ * autoRedirectOnLoginPage::
+ * If true, the login page will automatically be redirected to the OpenID
+ * Connect Provider, as when the button is pressed. The default is `false`.
+ *
+ * mode::
+ * This is the attribute in the owncloud accounts table to search for users.
+ * The default value is `email`. An alternative value: `userid`.
+ *
+ * search-attribute::
+ * This is the claim from the OpenID Connect user information which shall be
+ * used for searching in the accounts table. The default value is `email`. For
+ * more information about the claim, see
+ * https://openid.net/specs/openid-connect-core-1_0.html#Claims.
+ *
+ * use-token-introspection-endpoint::
+ * There are tokens which are not JSON WebToken(JWT) and information like the
+ * expiry cannot be read from the token itself. In these cases, the OpenID
+ * Connect Provider needs to call on the token introspection endpoint to get
+ * this information. The default value is `false`. See
+ * https://tools.ietf.org/html/rfc7662 for more information on token
+ * introspection.
+ */
+'openid-connect' => [
+	'autoRedirectOnLoginPage' => false,
+	'client-id' => '',
+	'client-secret' => '',
+	'loginButtonName' => 'OpenId Connect',
+	'mode' => 'userid',
+	// Only required if the OpenID Connect Provider does not support service discovery
+	'provider-params' => [
+		'authorization_endpoint' => '',
+		'end_session_endpoint' => '',
+		'jwks_uri' => '',
+		'registration_endpoint' => '',
+		'token_endpoint' => '',
+		'token_endpoint_auth_methods_supported' => '',
+		'userinfo_endpoint' => ''
+	],
+	'provider-url' => '',
+	'search-attribute' => 'sub',
+	'use-token-introspection-endpoint' => true,
+  ],
 
 ];


config/config.apps.sample.php
```diff
@@ -102,4 +102,33 @@
...
+    'provider-params' => array(
+        'authorization_endpoint' => '...',
+        'end_session_endpoint' => '...',
+        'jwks_uri' => '...',
+        'registration_endpoint' => '...',
+        'token_endpoint' => '...',
+        'token_endpoint_auth_methods_supported' => '...',
+        'userinfo_endpoint' => '...'
+    ),
+    'provider-url' => '',
```
DeepDiver1975: following are missing (they are optional aka come with reasonable defaults) - never the less
```
'autoRedirectOnLoginPage' => false,
  'mode' => 'userid',
      'search-attribute' => 'sub',
      'use-token-introspection-endpoint' => true
```

- autoRedirectOnLoginPage (default is false) - if true the login page will automatically be redirected to the idp (just like the button is pressed)
- use-token-introspection-endpoint (default is false) - there are tokens which are not JWT and information like the expiry cannot be read from the token it self. In these cases the IdP needs to be called on the token introspection endpoint to get this information
- mode (default is 'email' - alternative value: 'userid') - this is the attribute in the owncloud accounts table to search for users
- search-attribute (default: 'email') - this is the claim from the OpenID user info which shall be used for searching in the accounts table

DeepDiver1975 (changes requested):

config/config.apps.sample.php
settermjd: Thank you for the further information.

settermjd: The four parameters have been added to the documentation.

phil-davis (commented):
Language looks OK. I am not qualified to know if the detailed content is correct/complete.

mmattel:
@DeepDiver1975 :wave: mind to re-review so we could merge ?
Because of other changes in config.sample.php, a config-to docs run needs to be done and I want to do it in one shot...

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
 
 'collabora_group' => '',
 
+/**
+ * OIDC Configuration
+ */
+
+/**
+ * Configure OIDC OpenID Configuration
```
DeepDiver1975: duplication OIDC is OpenID Connect - and it is "OpenID Connect" not OpenID - this is something totally different

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
 
 'collabora_group' => '',
 
+/**
+ * OIDC Configuration
```
DeepDiver1975: following https://en.wikipedia.org/wiki/OpenID_Connect

We might want to write "OpenID Connect(OIDC)"

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ */
+
+/**
+ * Configure OIDC OpenID Configuration
+ *
+ * The `provider-url`, `client-id` and `client-secret` variables are to be
+ * taken from the OpenID Provider's setup. The `loginButtonName` variable can
```
DeepDiver1975: OpenID Connect

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ * be freely chosen, depending on the installation.
+ *
+ * NOTE: The provider-params configuration array only needs to be used if the
+ * OpenID Provider does NOT support service discovery.
```
DeepDiver1975: add Connect

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ *
+ * autoRedirectOnLoginPage::
+ * If true, the login page will automatically be redirected to the IdP, as when
```
DeepDiver1975: IdP -> OpenID Connect Provider - for the sake for clarity

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ * the button is pressed. The default is `false`.
+ *
+ * mode::
+ * This is the attribute in the owncloud accounts table to search for users.
+ * The default value is `email`. An alternative value: `userid`.
+ *
+ * search-attribute::
+ * This is the claim from the OpenID user information which shall be used for
```
DeepDiver1975: add Connect

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ * searching in the accounts table. The default value is `email`.
+ *
+ * use-token-introspection-endpoint::
+ * There are tokens which are not JWT and information like the expiry cannot be
```
DeepDiver1975: JWT -> JSON WebToken(JWT)

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ * read from the token itself. In these cases, the IdP needs to call on the
```
DeepDiver1975: IdP -> OpenID Connect Provider

config/config.apps.sample.php
```diff
@@ -102,4 +102,57 @@
...
+ * token introspection endpoint to get this information. The default value is
+ * `false`.
+ */
+'openid-connect' => array(
+    'autoRedirectOnLoginPage' => false,
+    'client-id' => '',
+    'client-secret' => '',
+    'loginButtonName' => 'OpenId Connect',
+    'mode' => 'userid',
+    // Only required if the OpenID Provider does not support service discovery
```
DeepDiver1975: add Connect

config/config.apps.sample.php
DeepDiver1975: ref for claims - https://openid.net/specs/openid-connect-core-1_0.html#Claims

config/config.apps.sample.php
DeepDiver1975: ref token introspection -> https://tools.ietf.org/html/rfc7662

DeepDiver1975 (changes requested):

phil-davis:
@settermjd I don't see where changes have been pushed yet

DeepDiver1975 (changes requested):
please respect given comments - THX

mmattel:
@settermjd 
Because these config values belong to a particular app which are not part of core or part of apps bundled with core in the tar file, you need to write which app is needed for these config values.
```
/**
 * OIDC Configuration  (Which App??)
 */
```
Please see the examples in the file. If there is no app needed, you have to move this to the standard config.sample file into the correct section.

mmattel:
@settermjd ping, any update?

DeepDiver1975 (dismissed):
Looks good now - THX

- [ ] squash
- [ ] make drone happy (code style is wrong)

config/config.apps.sample.php
```diff
@@ -101,5 +101,62 @@
...
+ * this information. The default value is `false`. See
+ * https://tools.ietf.org/html/rfc7662 for more information on token
+ * introspection.
+ */
+'openid-connect' => array(
+    'autoRedirectOnLoginPage' => false,
+    'client-id' => '',
+    'client-secret' => '',
+    'loginButtonName' => 'OpenId Connect',
+    'mode' => 'userid',
```
mmattel: @deepdiver are there other modes than `userid` ? or is this a single fixed value.
In case of other possible values, which are they and when to chose?

config/config.apps.sample.php
```diff
@@ -101,5 +101,62 @@
...
+        'authorization_endpoint' => '...',
+        'end_session_endpoint' => '...',
+        'jwks_uri' => '...',
+        'registration_endpoint' => '...',
+        'token_endpoint' => '...',
+        'token_endpoint_auth_methods_supported' => '...',
+        'userinfo_endpoint' => '...'
+    ),
+    'provider-url' => '',
+    'search-attribute' => 'sub',
```
mmattel: @deepdiver are there other modes than sub? or is this a single fixed value.
In case of other possible values, which are they and when to chose?

config/config.apps.sample.php
```diff
@@ -101,5 +101,62 @@
...
+    'use-token-introspection-endpoint' => true,
```
mmattel: @DeepDiver1975 what happens if false?
In case of other possible values, which are they and when to chose?

config/config.apps.sample.php
```diff
@@ -101,5 +101,62 @@
...
+ */
+'openid-connect' => array(
+    'autoRedirectOnLoginPage' => false,
+    'client-id' => '',
+    'client-secret' => '',
+    'loginButtonName' => 'OpenId Connect',
+    'mode' => 'userid',
+    // Only required if the OpenID Connect Provider does not support service discovery
+    'provider-params' => array(
+        'authorization_endpoint' => '...',
```
mmattel: @DeepDiver1975 why do we write `'...'` where above we only have  `''` does not look consitent

DeepDiver1975: ask @settermjd not me ...

mmattel: ah, ok. so no prerequisite.

@settermjd do you have a reason for?
I think using dots are not that good

settermjd:
> Looks good now - THX
> 
>     * [ ]  squash
> 
>     * [ ]  make drone happy (code style is wrong)

shall do.

config/config.apps.sample.php
settermjd: It might have been a carry-over from the notes that I was given to use. I'll edit the reference shortly.

settermjd: Done.

micbar:
@settermjd please use the shorthand array format `[]` instead of `array()`

phil-davis:
@settermjd ping
This likely only needs you to make minor fixes and squash.

phil-davis (approved):

mmattel (approved):

PR #36528, opened 2019-12-05
Update defaults.php

### LINKED ISSUE #36512: Change doc version string from calculated to „latest“
### Steps to reproduce
1. The doc version string currently looks like: `https://doc.owncloud.org/server/10.3` 
2. where `10.3` gets calculated

Background: in docs, the latest release is always set to `latest` which makes any release cycle not dependent of a version number hardcoded in core. When core changes the version number, docs updates internally latest with the release number of core.

To avoid inconsistencies or missing matches, the core reference to docs should contain `latest`.

### Expected behaviour
`https://doc.owncloud.org/server/latest`

### Actual behaviour
`https://doc.owncloud.org/server/10.3`

### Server configuration
**ownCloud version:** 10.3

The location of the change is at:
`/lib/private/legacy/defaults.php`, line 72
`$this->defaultDocVersion = $version[0] . '.' . $version[1]; // used to generate doc links` 
A change like `$version[0] . '.' . $version[1]` —> `latest` should be made there.

@micbar, @settermjd

mmattel:
## Description
This PR changes the defaultDocVersion string from a calculated number to `latest`
`https://doc.owncloud.org/server/10.3` -->
`https://doc.owncloud.org/server/latest`

## Related Issue
- Fixes https://github.com/owncloud/core/issues/36512

## Motivation and Context
Documentation uses `latest` as part of the URL pointing always to the latest release number given by published core. Release numbers are a internal one time change in docs per published core release.

## How Has This Been Tested?
Manually tested

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/legacy/defaults.php
+++ b/lib/private/legacy/defaults.php
@@ -58,7 +58,6 @@ class OC_Defaults {
 	public function __construct() {
 		$this->l = \OC::$server->getL10N('lib');
 		$this->config = \OC::$server->getConfig();
-		$version = \OCP\Util::getVersion();
 
 		$this->defaultEntity = 'ownCloud'; /* e.g. company name, used for footers and copyright notices */
 		$this->defaultName = 'ownCloud'; /* short name, used when referring to the software */
@@ -69,7 +68,7 @@ class OC_Defaults {
 		$this->defaultiTunesAppId = '1359583808';
 		$this->defaultAndroidClientUrl = 'https://play.google.com/store/apps/details?id=com.owncloud.android';
 		$this->defaultDocBaseUrl = 'https://doc.owncloud.org';
-		$this->defaultDocVersion = $version[0] . '.' . $version[1]; // used to generate doc links
+		$this->defaultDocVersion = 'latest'; // used to generate doc links
 		$this->defaultSlogan = $this->l->t('A safe home for all your data');
 		$this->defaultLogoClaim = '';
 		$this->defaultMailHeaderColor = '#1d2d44'; /* header color of mail notifications */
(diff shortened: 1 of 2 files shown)


mmattel:
Needs fixng of `tests/acceptance/features/webUIAdminSettings/helpAndTips.feature`

phil-davis:
I pushed a commit. There was some fancy logic to work out what the link should be with the embedded version number. Now that is a lot simpler - see the diffs.

phil-davis (dismissed):
LGTM

micbar (changes requested):
@settermjd had objections. Please explain.

VicDeo:
@micbar let me explain the consequences...
This patch replace current OC version with the word `latest` in the documentation link.
Imagine we release OC 10.4 with this change and then we release  OC 10.5, 11.0, 100.500.

if it is expected that  10.4, 10.5, 11.0, 100.500 have **the same** link to docs  `https://doc.owncloud.org/server/latest/_topic_` regardless the fact that we drop something sometimes (like APC cache support in 10.3) - feel free to merge

otherwise let it be separate docs for 10.4, 10.5, 11.0, 100.500 (like we have it now)

phil-davis:
Now I understand this. When you run 10.0 you should be linked to the 10.0 docs. When you run 10.1 your should be linked to the 10.1 docs... - which is what happens without this change.

The "assumption" is then that the docs for each version are actually published for "a decent amount of time" after that version stops being supported.

PR #36531, opened 2019-12-06
[For 10.4] Include expiration option for user and group shares

phil-davis:
## Description
This is a re-submit of PR #36324 with the commits squashed and rebased to the current master (that has "drop PHP 7.0" and phpunit7).

Please work on top of this in a separate branch for making acceptance tests or fixing bugs. Then we need to cherry-pick tests and bugs fixes into here in a controlled manner so that we do not write over each other's work.

Excerpt from the feature file:

* As a user I want to set an expiration date for collaborators in order to stay in control of access permissions by giving time-based access instead of granting access forever or until I manually remove the collaborator.
* As a user I want to automatically limit my responsibility for a guest account. Guests are bound to being a collaborator. If they are not a collaborator of any resource, they can't log in anymore.
* As an admin I want to set a default expiration date without enforcement to give a recommendation to users and help them maintain control over their resources. Still, I don't want to force or annoy users as some may have reasons to collaborate on resources for a longer or unrestricted amount of time, e.g., in projects.
* As an admin I have to follow corporate policies which dictate that data exchange between employees or also with externals may only take place for a certain maximum amount of time to prevent data leaks by human mistake or negligence.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/2001

## How Has This Been Tested?
CI acceptance tests added for:
- [x] new capabilities added in capabilities API
- [x] setting/clearing the new user and group expiration settings in the admin sharing settings UI page
- [x] sharing with users and groups with/without expiration date on the UI
- [x] using the sharing API to directly share/reshare/change share with default or specified expiry dates

## Screenshots:
Anmerkung 2019-11-11 133034

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added


phil-davis:
Fix for the "group expiration only" problem have been merged from PR #36535 
Next is to get the remaining acceptance tests sorted out.

individual-it:
Without the administrator setting a "default" expiration date, the user cannot set any expiration date at all. This is different to public link shares, where the user can always set an expiration date, whether a default is set of not.
For my logic there should be 3 different options:
1. enable expiration dates
2. set default expiration date
3. enforce default expiration date as maximum expiration date

currently 1&2 are combined

maybe a renaming of the option would also do

jvillafanez:
> Without the administrator setting a "default" expiration date, the user cannot set any expiration date at all. This is different to public link shares, where the user can always set an expiration date, whether a default is set of not.

Backend behaviour should be the same for link, user and group shares. It seems a frontend problem to me.

What I understand is:
* "enable expiration dates" will show or hide the date picker
* if there is as default expiration date set, the date picker should choose / show that option if possible, but it must allow the user to choose any date. If there isn't a default expiration date set, probably the easiest option is not to do anything, just show the date picker without any particular date selected.
* if the default is enforced, the date picker shouldn't allow to choose dates further than the one set (if possible), but the user should be able to choose any date between the current date and the default date

There is a corner case: no default expiration set, but enforced. As far as I know, this shouldn't happen. Worst case should be to enter either "0" days or "-1" days, but other than immediate expiration, it shouldn't be a problem (we might need to double-check this)

phil-davis:
I have squashed and rebased this code in PR #36573 (I did not want to lose anything here)

We can do final reviews and discussion of outstanding issues there.

PR #36532, opened 2019-12-06
CI updates for 10.3.2 branch

phil-davis:
The `release-10.3.2` branch now has old CI that tries to run some CI drone pipelines with PHP 7.0.
That is a problem when we try to merge it back to master, because master has dropped PHP 7.0 support.

This PR cherry-picks the drone CI-related changes from master, so that CI here can be up-to-day - and hopefully even pass!

### DIFF
--- a/.drone.starlark
+++ b/.drone.starlark
@@ -30,7 +30,6 @@ config = {
 		},
 		'reducedDatabases' : {
 			'phpVersions': [
-				'7.0',
 				'7.2',
 				'7.3',
 			],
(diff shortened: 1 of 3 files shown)


phil-davis:
Not so good - this PR is running pipelines for the newly-supported mariaDB and postgreSQL databases versions. And of course those do not pass in the `release-10.3.2` branch.

IMO we are in a catch-22 here - I don't want to push extra "temporary" changes (that remove those pipelines) onto this PR, because then they will end up on the `release-10.3.2` branch and will get merged back to `master` and we will lose the pipelines from master.

micbar:
Not needed.

PR #36499, opened 2019-11-30
Phpunit8 20191130

phil-davis:
Revive PR #34961 Declare testCase methods void to match modern phpunit declaration
And do stuff that we can do now in `phpunit7` and PHP 7.1 to be up-to-date for `phpunit8` and PHP 7.2`
https://phpunit.de/announcements/phpunit-8.html
https://thephp.cc/news/2019/02/help-my-tests-stopped-working

See if we can run `phpunit` tests with `phpunit` V8.

1) the setup* and teardown* methods in `phpunit8` have return type `void`, so they have to be declared like that where we inherit-override them.

2) `expectException` annotations are being deprecated in favour of `$this->expectException()` so change them.
https://thephp.cc/news/2016/02/questioning-phpunit-best-practices
php-cs-fixer fixes this automagically by adding this to the config:
```
			'php_unit_no_expectation_annotation' => true,
			'php_unit_expectation' => true,
```
That could be added to `owncloud-codestyle` so that we find and "fix" any of these in people's PRs.

3) `assertInternalType` is being deprecated in favour of `assertIsArray()` `assertIsInt()` etc.
https://github.com/sebastianbergmann/phpunit/issues/3368
php-cs-fixer fixes this automagically by adding this to the config:
```
			'php_unit_dedicate_assert_internal_type' => true,
```
That could be added to `owncloud-codestyle` so that we find and "fix" any of these in people's PRs.

PR #36539, opened 2019-12-07
Phpunit8 20191207

phil-davis:
Now that core has lots of deprecated unit test stuff fixed, re-run CI with phpunit8 and PHP 7.2 to see what happens. (this supersedes PR #36499 )


phil-davis:
See PR #37168 for a later branch of this.

PR #35105, opened 2019-04-26
Include a sync limiter for LDAP

jvillafanez:
## Description
Include a sync limiter and disable LDAP users automatically if the number of users is above the limit

## Related Issue
https://github.com/owncloud/enterprise/issues/3245

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

## Open tasks:

- [ ] Backport (if applicable set "backport-request" label and remove when the backport was done)

## TODO:
* [x] notification of disabled users
* [x] status / warning during `user:sync`
* [x] make "notification sent" key backend-agnostic
* [x] distinguish between no limit set for the backend and limit disabled for the backend
* [x] enabling user prevention in `user:enable` if applies (otherwise we might disable a different user in the next sync round)
* [x] handle migration scenario
* [x] unittests
* [ ] include "since" tag in the `OCP\IUserManager`
* [ ] (nice to have) enable auto_disabled users until fill the limit

PR #36563, opened 2019-12-10
Limit sync2

jvillafanez:
## Description
Include a sync limiter and disable LDAP users automatically if the number of users is above the limit
Cherry-picked and updated from https://github.com/owncloud/core/pull/35105

## Related Issue
https://github.com/owncloud/enterprise/issues/3245

## How Has This Been Tested?
Requires testing with the latest ownCloud

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

## TODO:
- [ ] Update "since" tag
- [ ] Check migrations (might need to change the filename)


jvillafanez:
I don't think I can rise coverage easily.
`AccountMapper` changes would require to modify the DB data, but restoring the previous state won't be clean taking into account how the tests are planned at the moment.
`SyncBackend` requires mocking 2-level dependencies (or more) due to the `SyncService` being created and not injected. In addition, the code is quite complex around the notifications, so we should consider to move that piece out to a different class.

jvillafanez:
:exclamation: `/settings/users/<uid>/enabled` endpoint allows enabling the users over the limit. This is a bug that should be fixed. The limit will be restored in the next login / sync, but it's very likely that a different user is disabled to force the limit.

DeepDiver1975:

PR #36565, opened 2019-12-11
Make sure to filter filecache subfolders in sql query instead of memory

mrow4a:
Related issue: https://github.com/owncloud/enterprise/issues/2925 
Goal: avoid having peaks in memory for large directory contents

While deleting directories with a lot of files inside using`view->unlink`, we do not need to put in memory unnecessarily records. Filter for directories when only subdirs needed.

Example:
- we have folder with 50k files and 100 subfolders
- some parent function already loaded all 50100 records in memory
- now we are going to run `unlink` on that folder contents
- previously, all 50100 records were in memory in `unlink`
- now, only 100 records are in memory `unlink`
- resulting peak memory is 50200 instead of 100200

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -36,6 +36,7 @@
 namespace OC\Files\Cache;
 
 use Doctrine\DBAL\Platforms\OraclePlatform;
+use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\Cache\ICache;
 use OCP\Files\Cache\ICacheEntry;
 use \OCP\Files\IMimeTypeLoader;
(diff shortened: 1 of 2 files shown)


mrow4a:
I will probably add 1 more unit test

jvillafanez:
Code looks good but we need to fix the tests.

mrow4a:
yes I noticed that my test that I wrote failed because of some wrapping logic

tests/lib/Files/Cache/Wrapper/CacheJailTest.php
```diff
@@ -80,4 +80,9 @@ public function testGetIncomplete() {
 		//not supported
 		$this->assertTrue(true);
 	}
+
+	public function testFolderContentsFilter() {
```
jvillafanez: I don't really like this...
If the problem is that the parent test is calling a private method, can we adjust that test to call a public one ensuring that the call goes through the code we want? If needed, add 2 different tests, one per case.

mrow4a: @jvillafanez I actually did as some other examples above in that test case. I am not sure if this is relevant to build testcase here.

jvillafanez: My worry is that, if we ever need to change the implementation in the `CacheJail` class, we might assume that there are tests, and due to the test passing we could think the code works even though it might be broken.

I think it's easier to call a public method in the `CacheTest`, and then copy & paste the test to adjust the expectations, if needed, in the rest of the implementations

mrow4a: @jvillafanez ok now I remember why I commented with `not supported as parent CacheTest calls private method of Cache class`. Because if I try to call this private function I get exceptions on connection being null. 

Please note that this function under test is private! It is called from public function `remove`, that currently has no unit tests.

mrow4a:
@micbar release note for this PR in https://github.com/owncloud/core/pull/36602

tests/lib/Files/Cache/Wrapper/CacheJailTest.php
jvillafanez: What if you test the public functions that call that private function instead?

I'm not sure if it's a better alternative that we mark the test as skipped or incomplete instead.

mrow4a: Ok I will test public function remove, but it means also we cannot test filtering mechanism itself (coverage will however be happy)

mrow4a: @jvillafanez done

micbar:
@mrow4a code style, double empty line

jvillafanez (approved):

PR #22700, opened 2016-02-28
For now only allow 1 link share per file/folder

rullzer:
Fixes #22692
This is a temp fix util we get #22327

We just block it in the OCS Share API since we have 1 endpoint now anyway.

PR #36566, opened 2019-12-11
Remove '1 link share per file/folder' old test

phil-davis:
## Description
Delete an old test:
`Scenario: Only allow 1 link share per file/folder`

That was introduced by PR #22700 - which temporarily restricted ownCloud to only create 1 public link for any resource. That restriction is not true any more, but the test was snake oil and was still passing "by accident".

## Motivation and Context
PHP 7.4 reported a notice for the code that did 
`$this->savedShareId = $this->lastShareData['data']['id']`
because `lastShareData` is an XML Object, not an array.

In earlier PHP versions, `savedShareId` got silently set to the empty string.

The `Then` step compared a new `lastShareData['data']['id']` empty string to the saved empty string, and they matched. So the test passed!

Note: If the test had still been relevant, then the access of the share Id should have been done like:
```
$this->savedShareId = $this->lastShareData->data->id;
```
Doing that made the test fail (which is what we really expect).
There are already tests in `apiShareManagement/mutlilinksharing.feature` that check that that multiple public link shares of the same resource can be created.

## How Has This Been Tested?
Delete code is tested code.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -678,17 +678,6 @@ Feature: sharing
       | 1               | 403             | 200              |
       | 2               | 403             | 403              |
 
-  @public_link_share-feature-required
-  Scenario: Only allow 1 link share per file/folder
-    Given using OCS API version "1"
-    And as user "user0"
-    And the user has created a public link share with settings
-      | path | welcome.txt |
-    And the last share id has been remembered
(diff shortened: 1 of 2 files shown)


individual-it (approved):
:+1: deleted code, is debugged code!

individual-it:
oracle install-server failed, restarted CI

PR #36576, opened 2019-12-11
[for 10.4] Receive multiple users from backend for user sync

sharidas:
Receive multiple users from the backend for user sync
instead of checking for single user. From the received
users check if the user to be synced is available in
the list.

## Description

Problem: When admin tries to sync a user say for ldap backend, there can be cases where names like `mhof`, `mhoff`, `mhoffg` exist in the ldap. And when admin tries to sync `mhof`, the admin gets Multiple user exists exception.

What this PR proposes:
Grab all the users available from the user backend, which the admin had searched for. And then try to match the users with the user to be searched for. If there are multiple users with same user name available then throw the exception. If only one user is available then sync it, else as usual show message saying user does not exist in the backend.

## Related Issue

- Fixes enterprise ticket 3447

## Motivation and Context

Do not check for the count of users. Instead check if the users received from the backend contains the exact match or not. 

## How Has This Been Tested?

- Created LDAP setup
- Created LDAP users `mhof`, `mhoff`, `mhoffg`
- Now as oC admin tried to sync the user `mhof`
```
sujith@sujith-ownCloud  ~/test/owncloud   fix-sync-user ●  ./occ user:sync "OCA\User_LDAP\User_Proxy" -u 'mhof'          
If unknown users are found, what do you want to do with their accounts? (removing the account will also remove its data)
  [0] disable
  [1] remove
  [2] ask later
 > 0
Syncing mhof ...

 sujith@sujith-ownCloud  ~/test/owncloud   fix-sync-user ● 
```
- As admin user try to sync `mhoffg123` which does not exist:
```
sujith@sujith-ownCloud  ~/test/owncloud   fix-sync-user ●  ./occ user:sync "OCA\User_LDAP\User_Proxy" -u 'mhoffg123'
If unknown users are found, what do you want to do with their accounts? (removing the account will also remove its data)
  [0] disable
  [1] remove
  [2] ask later
 > 0
Syncing mhoffg123 ...
Disabling accounts:
mhoffg123, ,  (no longer exists in the backend)

 sujith@sujith-ownCloud  ~/test/owncloud   fix-sync-user ● 
```

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Command/User/SyncBackend.php
+++ b/core/Command/User/SyncBackend.php
@@ -252,15 +252,23 @@ class SyncBackend extends Command {
 		$missingAccountsAction
 	) {
 		$output->writeln("Syncing $uid ...");
-		$users = $backend->getUsers($uid, 2);
-		if (\count($users) > 1) {
-			throw new \LengthException("Multiple users returned from backend for: $uid. Cancelling sync.");
+		$userUids = $backend->getUsers('', null);
+		$userToSync = null;
+		foreach ($userUids as $userUid) {
+			if ($userUid === $uid) {
+				if ($userToSync === null) {
+					$userToSync = $userUid;
+				} else {
+					throw new \LengthException("Multiple users returned from backend for: $uid. Cancelling sync.");
+				}
+			}
 		}
 
 		$dummy = new Account(); // to prevent null pointer when writing messages
-		if (\count($users) === 1) {
+
+		if ($userToSync !== null) {
 			// Run the sync using the internal username if mapped
-			$syncService->run($backend, new \ArrayIterator([$users[0]]), function () {
+			$syncService->run($backend, new \ArrayIterator([$userToSync]), function () {
 			});
 		} else {
 			// Not found
--- a/changelog/unreleased/36576
+++ b/changelog/unreleased/36576
@@ -1,0 +1,11 @@
+Bugfix: Receive multiple users for user sync command
+
+Recieve multiple users for user sync command. Previously
+when multiple users were returned, an exception was thrown
+and the command was aborted. In this fix we allow multiple
+users to be returned, and we check for the uid provided by the
+admin matches with the returned users. And if we find matches
+of more than one users with same uid, then we throw the exception
+that was thrown previously. The messages are kept intact.
+
+https://github.com/owncloud/core/pull/36576
--- a/tests/lib/Command/User/SyncBackendTest.php
+++ b/tests/lib/Command/User/SyncBackendTest.php
@@ -276,7 +276,9 @@ class SyncBackendTest extends TestCase {
 		$outputInterface = $this->createMock(OutputInterface::class);
 		$syncService = $this->createMock(SyncService::class);
 
-		$this->dummyBackend->method('getUsers')->willReturn(['existing-uid', 'should-explode']);
+		$this->dummyBackend->method('getUsers')->willReturn(['existing-uid', 'existing-uid']);
+		$this->dummyBackend->method('getDisplayName')
+			->willReturn('existing-uid');
 
 		$missingAccountsAction = 'disable';
 		$syncService->expects($this->never())->method('run');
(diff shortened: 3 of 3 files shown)


core/Command/User/SyncBackend.php
```diff
@@ -252,15 +252,25 @@ private function syncSingleUser(
 		$missingAccountsAction
 	) {
 		$output->writeln("Syncing $uid ...");
-		$users = $backend->getUsers($uid, 2);
-		if (\count($users) > 1) {
-			throw new \LengthException("Multiple users returned from backend for: $uid. Cancelling sync.");
+		$users = $backend->getUsers($uid);
```
jvillafanez: double-check you'll get all the users. I'm not sure if there is an implicit limit

core/Command/User/SyncBackend.php
```diff
@@ -252,15 +252,25 @@ private function syncSingleUser(
...
+		$userFound = false;
+		$userToSync = null;
+		foreach ($users as $user) {
+			if ($backend->getDisplayName($user) === $uid) {
```
jvillafanez: I think we're syncing using the uid, not the displayname. Make sure we keep the same behaviour

core/Command/User/SyncBackend.php
```diff
@@ -252,15 +252,25 @@ private function syncSingleUser(
...
+				if (!$userFound) {
+					$userFound = true;
```
jvillafanez: this `$userFound` shouldn't be needed. After the loop you can have:
* `$userToSync = null` -> user not found
* `$userToSync = $user` -> user found
* exception thrown -> multiple users found

Basically, during the loop, if `$userToSync` is null you assign the user (if matches), and if there is already a user assigned you throw the exception.

core/Command/User/SyncBackend.php
sharidas: As of now the backend does not provide uid. Hence I had to use the one which is available.

core/Command/User/SyncBackend.php
sharidas: If no value is passed the default value of the limit is set to null here.

core/Command/User/SyncBackend.php
jvillafanez: The `$backend->getUsers()` gives the uids of the users found. You can check with those uids.

sharidas: Unfortunately that won't work. Because say if there are 15 users in LDAP. And if we use `$backend->getUsers()` instead of `$backend->getUsers($uid)` then in the result we see around 10 uids. Which may or may not match with the uid we have passed. In fact I was hit with a situation while testing: the uid is available in LDAP, but the `getUsers()` returned the list which didn't had that uid.

jvillafanez: to clarify: we were checking and looking for user using the **owncloud's uid**. This must be kept in order not to change the behaviour.
It's irrelevant if the ownCloud's uid is linked to the LDAP's entryuuid or the LDAP's uid or whatever. The admin is expected to enter the ownCloud's uid, whatever it is.

sharidas: Updated the code. Now it would fetch all the users and will check for the uid passed by the admin.

jvillafanez (approved):
Code looks good.

phil-davis:
I added issue #36597 to make CLI acceptance tests.
We do not seem to have any acceptance tests for the `user:sync` command - so we won't hold up merging this PR.

phil-davis:
@sharidas needs a changelog entry

phil-davis:
@sharidas 
1) Is there an issue to link to in the top post?
2) needs a changelog item

sharidas:
@phil-davis
- I have updated the enterprise ticket in the pr template
- Also the changelog was updated changelog/unreleased/36576 in this PR.

phil-davis:
Looks good - I will merge.

PR #36577, opened 2019-12-11
check for isset before accessing in toXML()

phil-davis:
## Description
When running PHP 7.4 I noticed the server emit:
```
[Wed Dec 11 19:37:16 2019] Trying to access array offset on value of type int at /home/phil/git/owncloud/core/lib/private/legacy/api.php#463
```
I happened to be running an acceptance test:
```
make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/apiWebdavProperties/setFileProperties.feature:12
```
But I suspect this can happen with anything that needs to call `toXML()`

Check that `$k[0]` is set before accessing `$k[0]` (PHP is happy with `isset($k[0])` when `$k` itself is an int...)

## How Has This Been Tested?
Run the acceptance test again with PHP 7.4 and observe that the server does not emit the message.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/legacy/api.php
+++ b/lib/private/legacy/api.php
@@ -460,7 +460,7 @@ class OC_API {
 	 */
 	private static function toXML($array, $writer) {
 		foreach ($array as $k => $v) {
-			if ($k[0] === '@') {
+			if (isset($k[0]) && ($k[0] === '@')) {
 				if (\is_array($v)) {
 					foreach ($v as $name => $value) {
 						$writer->writeAttribute($name, $value);


phil-davis:
Hmmm - https://drone.owncloud.com/owncloud/core/21932/37/12
```
  Scenario: admin gets app info                                                                                 # /drone/src/tests/acceptance/features/apiProvisioning-v1/getAppInfo.feature:11
    When the administrator gets the info of app "files"                                                         # FeatureContext::theAdministratorGetsTheInfoOfApp()
    Then the OCS status code should be "100"                                                                    # OCSContext::theOCSStatusCodeShouldBe()
    And the HTTP status code should be "200"                                                                    # FeatureContext::theHTTPStatusCodeShouldBe()
    And the XML "data" "id" value should be "files"                                                             # FeatureContext::theXMLKey1Key2ValueShouldBe()
    And the XML "data" "name" value should be "Files"                                                           # FeatureContext::theXMLKey1Key2ValueShouldBe()
    And the XML "data" "types" "element" value should be "filesystem"                                           # FeatureContext::theXMLKey1Key2Key3ValueShouldBe()
    And the XML "data" "dependencies" "owncloud" "min-version" attribute value should be a valid version string # FeatureContext::theXMLKey1Key2AttributeValueShouldBe()
      attribute min-version value  is not a valid version string
    And the XML "data" "dependencies" "owncloud" "max-version" attribute value should be a valid version string # FeatureContext::theXMLKey1Key2AttributeValueShouldBe()
```
The change has some side-effect - need to look in the morning.

individual-it (approved):

PR #36581, opened 2019-12-12
Fix PHP Syntax error in config.sample.php

DeepDiver1975:
Quite obvious ...


mmattel (approved):

phil-davis (approved):
We should lint this file in CI

PR #36586, opened 2019-12-13
php unit test changes to address issues that php7.4 will report

phil-davis:
I am finding things in the unit tests that PHP7.4 complains about (PR #36500 )
Also put them here to see that the fixes/changes still pass with an older version(s) of PHP 7.1 7.2 7.3

1) Fix 'Function ReflectionType::__toString() is deprecated' - in `ControllerMethodReflector.php` we will not be able to "just cast to string", use the `getName()` method.
2) Fix 'stripos(): Non-string needles will be interpreted as strings in the future - `$pattern` might look like something other than a string. Cast it to string because we really want it to be a string.
3) Fix 'Trying to access array offset on value of type bool' - the file called "hack" that was being put in the cache is sometimes cleaned out of the cache already by the test and the attempt to `instance->remove('hack')` complained because actually hack does not always exits. Only remove it if it still exists.
4) Pass real core.key core.crt in `SignAppTest` and `SignCoreTest` to fix 'Trying to access array offset on value of type bool' - the mocking was not actually reading in the test key and crt files. Somehow the test was passing with the privateKey being literally `privateKey` and the certificate being the literal string `certificate`
5) And similar stuff in other files.

### DIFF
--- a/lib/private/AppFramework/Utility/ControllerMethodReflector.php
+++ b/lib/private/AppFramework/Utility/ControllerMethodReflector.php
@@ -64,7 +64,7 @@ class ControllerMethodReflector implements IControllerMethodReflector {
 			if (\method_exists($param, 'getType')) {
 				$type = $param->getType();
 				if ($type !== null) {
-					$this->types[$param->getName()] = (string) $type;
+					$this->types[$param->getName()] = $type->getName();
 				}
 			}
 
--- a/lib/private/AppFramework/Utility/ControllerMethodReflector.php
+++ b/lib/private/AppFramework/Utility/ControllerMethodReflector.php
@@ -63,6 +63,7 @@ class ControllerMethodReflector implements IControllerMethodReflector {
 			// over phpdoc annotations
 			if (\method_exists($param, 'getType')) {
 				$type = $param->getType();
+				'@phan-var \ReflectionNamedType $type';
 				if ($type !== null) {
 					$this->types[$param->getName()] = $type->getName();
 				}
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -111,7 +111,7 @@ class Cache implements ICache {
 	 * get the stored metadata of a file or folder
 	 *
 	 * @param string | int $file either the path of a file or folder or the file id for a file or folder
-	 * @return ICacheEntry|false the cache entry as array of false if the file is not found in the cache
+	 * @return ICacheEntry|false the cache entry as array or false if the file is not found in the cache
 	 */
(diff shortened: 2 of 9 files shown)


phil-davis:
@DeepDiver1975 or whoever cares to give review/feedback on this - these are things that are reported by `phpunit` when running PHP 7.4, e.g. see PR #36670 

Does the way I have adjusted things in this PR make sense?

If so, then we could merge this, and then "someone" can be tasked to adjust the rest of the issues "someday".

phil-davis:
@DeepDiver1975 @micbar or anybody - is this the right way to adjust these unit tests?

If so, then this is an example for when we really get working on PHP 7.4 support.

It could be merged now - this all works with PHP 7.1 7.2 7.3

phil-davis:
Rebased just now to get new `.drone.star` stuff.

phil-davis:
@jvillafanez here is some changes for supporting PHP 7.4. This seems to be the sort of stuff we need to do. PR #37004 runs CI with PHP 7.4 to see what it reports (there are plenty more things to adjust)

phil-davis:
@micbar @DeepDiver1975 @jvillafanez @VicDeo or anyone.
This can be reviewed to see if there is agreement about what I have done here.
It all works fine with current PHP 7.1 7.2 7.3 core master. So it could be merged.
And then when somebody allocates resources in the future for PHP 7.4 support, it will be an example of changes, for the next developer to work through whatever other unit test warnings... need to be fixed.

jvillafanez:
The cleanup in the cache should be moved to the tear down function, but taking into account the previous code, I'm fine with these changes. :+1:

phil-davis:
> The cleanup in the cache should be moved to the tear down function, but taking into account the previous code, I'm fine with these changes. +1

@jvillafanez did you mean that I can leave the code as I have written it, or that you want me to refactor code into the teardown function?

jvillafanez:
Up to you. Technically, this PR is to fix php 7.4 issues, so the tear down problem is out of the scope. You can open a new PR to fix that problem, or if it's easier for you (and the amount of changes are controlled) you can fix it here.

phil-davis:
The code like:
```
$this->cache->remove($dir1);
```
Is followed by existing asserts like:
```
$this->assertFalse($this->cache->inCache($dir1));
```
So the cache remove code is no "just" cleanup at the moment.

That could all be sorted out (e.g. why are those `assertFalse` there? Do they really do some useful check related to the test case?) But IMO that is for some other PR.

@micbar @DeepDiver1975 @jvillafanez @VicDeo or anyone. Please review.

jvillafanez (approved):

PR #36573, opened 2019-12-11
[For 10.4] Include expiration option for user and group shares

phil-davis:
## Description
This is a re-submit of PR #36324 and then PR #36531 with the commits squashed and rebased to the current master (that has "drop PHP 7.0" and phpunit7).

Please work on top of this in a separate branch for making acceptance tests or fixing bugs. Then we need to cherry-pick tests and bugs fixes into here in a controlled manner so that we do not write over each other's work.

Excerpt from the feature file:

* As a user I want to set an expiration date for collaborators in order to stay in control of access permissions by giving time-based access instead of granting access forever or until I manually remove the collaborator.
* As a user I want to automatically limit my responsibility for a guest account. Guests are bound to being a collaborator. If they are not a collaborator of any resource, they can't log in anymore.
* As an admin I want to set a default expiration date without enforcement to give a recommendation to users and help them maintain control over their resources. Still, I don't want to force or annoy users as some may have reasons to collaborate on resources for a longer or unrestricted amount of time, e.g., in projects.
* As an admin I have to follow corporate policies which dictate that data exchange between employees or also with externals may only take place for a certain maximum amount of time to prevent data leaks by human mistake or negligence.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/2001
- Fixes #36545 [Acceptance tests related]

## How Has This Been Tested?
CI acceptance tests added for:
- [x] new capabilities added in capabilities API
- [x] setting/clearing the new user and group expiration settings in the admin sharing settings UI page
- [x] sharing with users and groups with/without expiration date on the UI
- [x] using the sharing API to directly share/reshare/change share with default or specified expiry dates

## Screenshots:
Anmerkung 2019-11-11 133034

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

Issue #36569, opened 2019-12-11
default expire date config parameters can be set to 0 days


phil-davis:
### Steps to reproduce
1. set `shareapi_expire_after_n_days_user_share` or `shareapi_expire_after_n_days_group_share` to 0
2. send a sharing API  request to create a share

### Expected behaviour
The parameter is not allowed to be zero.

### Actual behaviour
The share is created, and expires today

### Server configuration
Current core with feature branch `feature/user_group_share_expiration` PR #36531

phil-davis:
Another thought is that the value `0` could be re-used to mean that there is no default expire date.
In that case, when sharing, the UI would display the expiry date field and leave it empty - so that users can pick an expiry date if they want, but will get no expiry date "by default"

phil-davis:
After discussion there is no need to prohibit zero days - if the admin wants new shares to expire "today" by default then so be it. It will work, and shares created today will expire around midnight. If an admin does not want that behaviour, then they do not have to set this parameter to 0.

phil-davis:
Currently a share expires at the very beginning of the expiry date - e.g. if I set to expire on 2019-12-20 then when I get to the office on 2019-12-20 the share has already expired and is gone.

That results in strange behaviour if a user creates a share and chooses today as the expiry date. Similarly if the admin sets the default expire days to 0. The share gets created with expiry date today. As soon as the share receiver accesses the share, the backend detects that the share has already expired and deletes the share. The share receiver never sees the share.

Also, IMO, it would make sense that if I specify 2019-12-20 as the share expiry date, then the share will keep working right up to the end of 2019-12-20.

See PR #36609 for how the code and tests could be adjusted.

@pmaier1 @micbar @HanaGemela what do you think?

phil-davis:
https://github.com/owncloud/product/blob/master/modules/ROOT/examples/oc10/collaborators/expiration_date_for_collaborators.feature#L35 specifies

`WHEN the specified day has just ended`

So that matches the behaviour that I am suggesting above and in the PR.

phil-davis:
The code to make the share expire at the end of the specified day was commit https://github.com/owncloud/core/pull/36573/commits/8a1996c788e9944cf5f6731a351d92fcbfadcc12 in #36573 

After that code was done, it became sensible to allow "today" as the expiry date for a share. That was implemented in the sharing API (with some issues related to timezone offsets), but the webUI date-picker did not allow choosing "today".

[...]

PR #36609, opened 2019-12-18
Expire share after end of day

phil-davis:
## Description
With the current code in #36573 the share expires at the very start of the day specified. e.g. set the expiration date to "2019-12-20" and when you get to the office on 2019-12-20 the share is already expired and gone.

You can set the expiry date to today (and set the default expiry days to 0, and enforce that - any of those). If you then create a share that expires today, then the share gets created but effectively expires at the beginning of today. As soon as the share receiver goes to access the share, the backend code detects that the share is already expired and it deletes the share. That is all a bit silly.

This PR changes the meaning of the expiry so that "expire date 2019-12-20" means that the share expires at the end of "2019-12-20" - i.e. the moment that the date ticks over to "2019-12-21" then the share will get deleted.

That allows shares created with expire date "today" to work in a reasonable fashion.

And also maybe users will be happier that a share expiring on some future date will still work until the end of that day.

Edit: see https://github.com/owncloud/product/blob/master/modules/ROOT/examples/oc10/collaborators/expiration_date_for_collaborators.feature#L35 which specifies `WHEN the specified day has just ended` - that is the behaviour that this PR is implementing.

## Related Issue
- #36569 

## Motivation and Context
Adjust weird behaviour.

## How Has This Been Tested?
Local test runs of the acceptance test and unit tests that were changed.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -169,6 +169,17 @@ class Manager implements IManager {
 	}
 
 	/**
+	 * Decide if a share has expired
+	 *
+	 * @param \OCP\Share\IShare $share
+	 * @return bool
+	 */
+	private function shareHasExpired($share) {
+		$expirationDate = $share->getExpirationDate();
+		return ($expirationDate !== null) && ($expirationDate < new \DateTime("today"));
+	}
+
+	/**
 	 * Verify if a password meets all requirements
 	 *
 	 * @param string $password
@@ -1204,14 +1215,13 @@ class Manager implements IManager {
 
 		$providerIdMap = $this->shareTypeToProviderMap($shareTypes);
 
-		$today = new \DateTime();
 		foreach ($providerIdMap as $providerId => $shareTypeArray) {
 			// Get provider from cache
 			$provider = $this->factory->getProvider($providerId);
 
 			$queriedShares = $provider->getAllSharesBy($userId, $shareTypeArray, $nodeIDs, $reshares);
 			foreach ($queriedShares as $queriedShare) {
-				if ($queriedShare->getExpirationDate() !== null && $queriedShare->getExpirationDate() <= $today) {
+				if ($this->shareHasExpired($queriedShare)) {
 					try {
 						$this->deleteShare($queriedShare);
 					} catch (NotFoundException $e) {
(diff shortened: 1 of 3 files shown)


phil-davis:
As well as code review, we need to decide if this is the required behaviour.

phil-davis:
Note: JS tests are failing because they are currently failing in the underlying PR

lib/private/Share20/Manager.php
```diff
@@ -1361,15 +1369,16 @@ public function getAllSharedWith($userId, $shareTypes, $node = null) {
 		// Aggregate all required $shareTypes by mapping provider to supported shareTypes
 		$providerIdMap = $this->shareTypeToProviderMap($shareTypes);
 
-		$today = new \DateTime();
+		$today = new \DateTime("today");
```
jvillafanez: Not needed

phil-davis: removed, thanks for spotting

jvillafanez:
I think the problem is that we're setting the expiration date as `2020-1-12 00:00:00`, and we're probably overwriting the time (sending `2020-1-10 11:11:11` would be converted to `2020-1-10 00:00:00`).

I'd prefer to set the full datetime properly in the DB or set `23:59:59` as time if none is provided (`2020-1-11` would be converted to `2020-1-11 23:59:59`). This is probably more consistent with the current code, but it's probably more work.

In any case, I don't have a strong opinion on this if we decide just to focus on the specific date.

phil-davis:
Yes, I am not sure if we are setting the time in the database to 00:00:00 or to the current time.

In the acceptance tests we create a share, e.g. at 2019-12-16:09:15:30:00, with expiry date sent as "2019-12-16". The database might record 2019-12-16:09:15:30:00 or 2019-12-16:00:00:00:00 as the expiry date.

Then 1 second later at 2019-12-16:09:15:30:01 we check the receiver's view of the share. Either way the share is expired.

jvillafanez:
I'll ping @micbar to decide what to do. From my point of view, this is mergeable.

I'd recommend to merge this PR with the whole feature, and not block the release. We can open an enhancement to support specifying the time so you could set `2020-1-3 12:30:00` as expiration, to be prioritized after the release.

phil-davis:
CI passed (apart from the known JS test problem)
The underlying branch was rebased and the JS test problem was fixed. I squashed the 4 commits that were here into 1 commit and rebased it on top of the feature branch, so this should all be passing now.

phil-davis:
This has been cherry-picked into PR #36573

PR #36612, opened 2019-12-18
Add limit and offset for callForAllUsers

mrow4a:
I think this was somehow lost during the refactor - https://github.com/owncloud/core/blob/f10d105ea2f2d0822693313ddcd6f9b2eaac6ebe/lib/private/User/Manager.php#L396-L419

Due to the problem with PHP PDO, even using `fetch`, query returns and caches result at the driver. This can be problematic with instances that have over 10k+ users

### DIFF
--- a/lib/private/User/AccountMapper.php
+++ b/lib/private/User/AccountMapper.php
@@ -215,26 +215,43 @@ class AccountMapper extends Mapper {
 	}
 
 	public function callForAllUsers($callback, $search, $onlySeen) {
-		$qb = $this->db->getQueryBuilder();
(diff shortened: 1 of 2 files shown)


mrow4a:
Sorry, I am closing it, and add alternative - >`callForSeenUsers` that allows defining offset and limit in https://github.com/owncloud/core/pull/36602

mrow4a:
Obsoleted by https://github.com/owncloud/core/pull/36602/commits/b41ddc60b5775d41e70e4ea91d02da4a0a3dfd08 in https://github.com/owncloud/core/pull/36602

PR #36613, opened 2019-12-18
Allow "+" character in usernames

pako81:
## Description

Allow `+` character when creating users.

## Related Issue

- Relates to https://github.com/owncloud/guests/pull/384
- Fixes https://github.com/owncloud/enterprise/issues/3712

## Motivation and Context

This PR allows the `+` character in the username when creating new users. This is also required when sharing with guest users having i.e. the following format `name+2@test.com`.

## How Has This Been Tested?

Manually by creating new users having the `+` sign in their usernames over the Users management page as well by using the `user:add` occ command.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Changelog item

### DIFF
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -332,10 +332,10 @@ class Manager extends PublicEmitter implements IUserManager {
 			$l = \OC::$server->getL10N('lib');
 
 			// Check the name for bad characters
-			// Allowed are: "a-z", "A-Z", "0-9" and "_.@-'"
-			if (\preg_match('/[^a-zA-Z0-9 _\.@\-\']/', $uid)) {
+			// Allowed are: "a-z", "A-Z", "0-9" and "+_.@-'"
+			if (\preg_match('/[^a-zA-Z0-9 \+_\.@\-\']/', $uid)) {
 				throw new \Exception($l->t('Only the following characters are allowed in a username:'
-					. ' "a-z", "A-Z", "0-9", and "_.@-\'"'));
+					. ' "a-z", "A-Z", "0-9", and "+_.@-\'"'));
 			}
 			// No empty username
 			if (\trim($uid) == '') {
--- a/changelog/unreleased/36613
+++ b/changelog/unreleased/36613
@@ -1,0 +1,5 @@
+Enhancement: Allow plus sign in username
+
+The plus sign is now allowed in a username, e.g. John+Smith
+
+https://github.com/owncloud/core/pull/36613
--- a/core/Command/User/Add.php
+++ b/core/Command/User/Add.php
@@ -64,7 +64,7 @@ class Add extends Command {
 			->addArgument(
 				'uid',
 				InputArgument::REQUIRED,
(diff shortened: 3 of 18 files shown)


lib/private/User/Manager.php
```diff
@@ -332,10 +332,10 @@ public function createUser($uid, $password) {
 			$l = \OC::$server->getL10N('lib');
 
 			// Check the name for bad characters
-			// Allowed are: "a-z", "A-Z", "0-9" and "_.@-'"
-			if (\preg_match('/[^a-zA-Z0-9 _\.@\-\']/', $uid)) {
+			// Allowed are: "a-z", "A-Z", "0-9" and "+_.@-'"
+			if (\preg_match('/[^a-zA-Z0-9 \+_\.@\-\']/', $uid)) {
```
phil-davis: Note: the escaping of `\+` here is not strictly required (just like the escaping of `\.` is actually not strictly required) but it does not hurt. In other contexts in a regex the plus sign does need to be escaped, so escaping it here just helps people not to have to think about that.

phil-davis:
@pako81 I adjusted acceptance tests to include some cases with a user with `+` in the username. And also added a changelog entry.

IMO  this works. @micbar this could be merged for 10.4 - is that OK?

pako81:
thanks a lot @phil-davis :)

pako81:
@phil-davis I was now thinking about the user provisioning APIs. In there, the `+` sign is still not correctly escaped and rather replaced with a space character.

phil-davis:
@pako81 I added provisioning API tests. They are passing locally for me and I expect that they will pass in drone CI. 
Probably the tests are providing the `+` in the URL as ASCII 43, `%2B` (I think that is true, without actually digging deep into the test code)
And that is how any client using the Provisioning API would need to specify a username containing a `+` - and the backend code will receive the `%2B` turned back into a `+` and it "just works"

pako81:
@phil-davis correct, I was testing the wrong way. The following seems to work correctly.

`curl -X POST http://admin:password@example.com/ocs/v1.php/cloud/users -d userid="name%2B1@test.com" -d password="test"`

phil-davis:
There have been other major features merged just now. So I rebased this PR to make sure that the CI is all green on top of those changes now in master.

phil-davis:
@micbar @HanaGemela I have added acceptance test coverage and it passes (adding users with `+` in ther username via the webUI, cli and provisioning API and doing stuff with them...) So this looks like it works fine.

I some developer can review the code and approve, and you think this is a change/enhancement that we should have, then feel free to merge it for 10.4

sharidas (commented):
LGTM :+1:

karakayasemi (changes requested):
Everything seems good. As a last, this should be adjusted also.
https://github.com/owncloud/core/blob/98922f11f87cb4818212d530a59c51d57fb4f124/core/Command/User/Add.php#L67

pako81:
thanks @karakayasemi :) missed that

karakayasemi (approved):

Issue #36518, opened 2019-12-04
Review acceptance test steps that process tables


phil-davis:
Steps that have the form: `Then stuff in the following table should be so (or not so)` are easy to accidentally pass when the table data and/or header row is missing. Or when the header of columns is misspelt or...

e.g. see #36517

If there are effectively no table rows correctly written, then the step code can loop around zero times, checking nothing. So it has no chance to fail.

- [x] review these type of step definitions
- [x] add code so they will fail if the table does not exist/is wrongly written
- [x] fix any broken/missing tables in steps of scenarios

dpakach:
@phil-davis I think we covered all. can we close this now?

phil-davis:
Yes, CI was good on Saturday and Sunday night, so any broken steps in app repos have also been fixed.

PR #36642, opened 2019-12-24
Acceptance tests: Add check for table rows and columns

dpakach:
## Description

Add check for table rows and columns on features from following suites
- apiAuth
- apiAuthOcs
- apiAuthWebDav
- apiCapabilities
- apiComments
- apiFavorites

## Related Issue

- Related #36518

## How Has This Been Tested?

:robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiAuthWebDav/webDavDELETEAuth.feature
+++ b/tests/acceptance/features/apiAuthWebDav/webDavDELETEAuth.feature
@@ -62,12 +62,12 @@ Feature: delete file/folder
     And a new browser session for "user0" has been started
     And the user has generated a new app password named "my-client"
     When the user requests these endpoints with "DELETE" using the generated app password then the status codes should be as listed
-      | endpoint                                      | http-code | body          |
-      | /remote.php/webdav/textfile0.txt              | 401       | doesnotmatter |
-      | /remote.php/dav/files/user0/textfile0.txt     | 401       | doesnotmatter |
-      | /remote.php/webdav/PARENT                     | 401       | doesnotmatter |
-      | /remote.php/dav/files/user0/PARENT            | 401       | doesnotmatter |
-      | /remote.php/dav/files/user0/PARENT/parent.txt | 401       | doesnotmatter |
+      | endpoint                                      | http-code |
+      | /remote.php/webdav/textfile0.txt              | 401       |
+      | /remote.php/dav/files/user0/textfile0.txt     | 401       |
+      | /remote.php/webdav/PARENT                     | 401       |
+      | /remote.php/dav/files/user0/PARENT            | 401       |
+      | /remote.php/dav/files/user0/PARENT/parent.txt | 401       |
 
   Scenario: send DELETE requests to webDav endpoints using app password token as password
     Given token auth has been enforced
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -2562,11 +2562,15 @@ trait BasicStructure {
 	 * @param array $requiredHeader
 	 * @param array $allowedHeader
 	 *
-	 * @return mixed[]
+	 * @return void
+	 * @throws Exception
 	 */
 	public function verifyTableNodeColumns($table, array $requiredHeader = [], array $allowedHeader = []) {
 		if (!($table instanceof TableNode)) {
 			throw new Exception("TableNode expected but got " . \gettype($table));
+		}
+		if (\count($table->getHash()) < 1) {
+			throw new Exception("Table should have at least one row.");
 		}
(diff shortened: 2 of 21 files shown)


tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -2586,4 +2587,56 @@ public function verifyTableNodeColumns($table, array $requiredHeader = [], array
...
+			foreach ($tableHeaders as $element) {
+				if (!\in_array($element, $allowedRows)) {
+					throw new Exception("Row with name '$element' is not allowed in table but found");
+				}
+			}
+		}
+	}
+
+	/**
+	 * Verify that the tableNode contains expected headers
```
phil-davis: function description needs updating

tests/acceptance/features/apiAuthWebDav/webDavDELETEAuth.feature
```diff
@@ -62,12 +62,12 @@ Feature: delete file/folder
     And a new browser session for "user0" has been started
     And the user has generated a new app password named "my-client"
     When the user requests these endpoints with "DELETE" using the generated app password then the status codes should be as listed
-      | endpoint                                      | http-code | body          |
-      | /remote.php/webdav/textfile0.txt              | 401       | doesnotmatter |
-      | /remote.php/dav/files/user0/textfile0.txt     | 401       | doesnotmatter |
-      | /remote.php/webdav/PARENT                     | 401       | doesnotmatter |
-      | /remote.php/dav/files/user0/PARENT            | 401       | doesnotmatter |
-      | /remote.php/dav/files/user0/PARENT/parent.txt | 401       | doesnotmatter |
+      | endpoint                                      | http-code |
```
HariBhandari07: Is there any reason for removing the body column?

tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
```diff
@@ -76,6 +76,7 @@ public function userGetsPropertiesOfFolder(
 		$user, $path, $propertiesTable
 	) {
 		$properties = null;
+		$this->featureContext->verifyTableNodeColumns($propertiesTable);
 		if ($propertiesTable instanceof TableNode) {
```
phil-davis: IMO this if can be removed - we know it is a TableNode

tests/acceptance/features/apiAuthWebDav/webDavDELETEAuth.feature
dpakach: Yes, We don't send the body while making the request in the code and it just doesn't matters because we are expecting it to fail and it gives 401.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -372,6 +372,7 @@ public function theseUsersHaveBeenCreatedWithoutSkeletonFiles(TableNode $table)
 	 * @throws \Exception
 	 */
 	public function theseUsersHaveBeenCreated($setDefaultAttributes, $doNotInitialize, TableNode $table) {
+		$this->verifyTableNodeColumns($table, ['username']);
```
phil-davis: Usually the table should have at least one entry. We should check that, otherwise we can easily accidentally have steps that do nothing.

Maybe `verifyTableNodeColumns` can check that there is at least 1 row of data in the table?
Then this will get checked for all column tables.

tests/acceptance/features/apiAuthWebDav/webDavDELETEAuth.feature
HariBhandari07: There may be many scenarios with this redundant body column. I think we should create a separate issue and handle it over there.

tests/acceptance/features/bootstrap/Provisioning.php
dpakach: Yeah, The username column should be always be present in the table. But there are other columns (eg. displayname, email, password) which are optional. That is already being checked by the function which is causing the tests to fail. I forgot adding those optional columns in this function.

tests/acceptance/features/apiAuthWebDav/webDavDELETEAuth.feature
dpakach: They are used in the other steps. The problem was with just this step. And in fact putting body in the request is alright but here this was just not used anywhere.

tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
dpakach: I think we can check the number of columns in the table.

phil-davis (approved):

PR #36646, opened 2019-12-26
refactor thenTheHTTPStatusCodeShouldBe

phil-davis:
## Description
The acceptance test step function `theHTTPStatusCodeShouldBe` had become used in its normal role as a Gherkin step definition, but also could be called with an array of valid status codes. That meant that the PHPdoc for the `statusCode` parameter was "overloaded" with stuff that is not generated by the Gherkin step text.

This PR refactors so that `theHTTPStatusCodeShouldBe` is a separate convenience function which takes the same parameters it currently does (so all callers will still work). The Gherkin step is now a function called `thenTheHTTPStatusCodeShouldBe`. That takes just a single status code from the Gherkin step text.

This is a similar style of refactoring to what we are doing with `Given` steps... I noticed this when doing some other code where I was calling `theHTTPStatusCodeShouldBe`

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/BasicStructure.php
+++ b/tests/acceptance/features/bootstrap/BasicStructure.php
@@ -981,28 +981,45 @@ trait BasicStructure {
 	}
 
 	/**
-	 * @Then /^the HTTP status code should be "([^"]*)"$/
+	 * Check that the status code in the saved response is the expected status
+	 * code, or one of the expected status codes.
 	 *
-	 * @param int|int[]|string|string[] $statusCode
+	 * @param int|int[]|string|string[] $expectedStatusCode
 	 * @param string $message
 	 *
 	 * @return void
 	 */
-	public function theHTTPStatusCodeShouldBe($statusCode, $message = "") {
-		if ($message === "") {
-			$message = "HTTP status code is not the expected value";
-		}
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/BasicStructure.php
```diff
@@ -981,30 +981,48 @@ public function isAPublicLinkUrl($url) {
...
 
+	/**
+	 * @Then /^the HTTP status code should be "([^"]*)"$/
+	 *
+	 * @param int|string $statusCode
+	 * @param string $message
+	 *
+	 * @return void
+	 */
+	public function thenTheHTTPStatusCodeShouldBe($statusCode, $message = "") {
```
individual-it: if that function is only called from Gherkin, why do we need the message parameter?

individual-it (dismissed):

tests/acceptance/features/bootstrap/BasicStructure.php
phil-davis: true - I cut-pasted to quickly without thinking enough.
fixed

individual-it (approved):

PR #36673, opened 2020-01-01
Acceptance tests: Add more fields in verify table calls used in other apps

dpakach:
## Description

Acceptance tests: Add more fields in verify table calls used in other apps

## Related Issue

- Related https://github.com/owncloud/notifications/issues/300, https://github.com/owncloud/richdocuments/issues/312

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebUINotificationsContext.php
+++ b/tests/acceptance/features/bootstrap/WebUINotificationsContext.php
@@ -75,7 +75,7 @@ class WebUINotificationsContext extends RawMinkContext implements Context {
 			\count($notifications),
 			"expected $number notifications, found " . \count($notifications)
 		);
-		$this->featureContext->verifyTableNodeColumns($expectedNotifications, ['title'], ['content']);
+		$this->featureContext->verifyTableNodeColumns($expectedNotifications, ['title'], ['content', 'link', 'message']);
 		foreach ($expectedNotifications as $expectedNotification) {
 			$found = false;
 			foreach ($notifications as $notification) {
--- a/tests/acceptance/features/bootstrap/WebUISharingContext.php
+++ b/tests/acceptance/features/bootstrap/WebUISharingContext.php
@@ -756,7 +756,11 @@ class WebUISharingContext extends RawMinkContext implements Context {
 	public function theUserSetsTheSharingPermissionsOfForOnTheWebUI(
 		$userOrGroup, $userName, $fileName, TableNode $permissionsTable
 	) {
-		$this->featureContext->verifyTableNodeRows($permissionsTable, [], ['share', 'edit', 'create', 'change', 'delete']);
+		$this->featureContext->verifyTableNodeRows(
+			$permissionsTable,
+			[],
+			['share', 'edit', 'create', 'change', 'delete', 'edit', 'secure-view', 'print']
+		);
 		// The capturing groups of the regex include the quotes at each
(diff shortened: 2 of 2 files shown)


tests/acceptance/features/bootstrap/WebUINotificationsContext.php
```diff
@@ -75,7 +75,7 @@ public function assertNotificationsOnWebUI(
 			\count($notifications),
 			"expected $number notifications, found " . \count($notifications)
 		);
-		$this->featureContext->verifyTableNodeColumns($expectedNotifications, ['title'], ['content']);
+		$this->featureContext->verifyTableNodeColumns($expectedNotifications, ['title'], ['content', 'link', 'message']);
```
phil-davis: how is `content` different to `message` ?

should we choose one of these and use it consistently?

tests/acceptance/features/bootstrap/WebUISharingContext.php
```diff
@@ -756,7 +756,11 @@ public function theUserTypesInTheShareWithField($input) {
 	public function theUserSetsTheSharingPermissionsOfForOnTheWebUI(
 		$userOrGroup, $userName, $fileName, TableNode $permissionsTable
 	) {
-		$this->featureContext->verifyTableNodeRows($permissionsTable, [], ['share', 'edit', 'create', 'change', 'delete']);
+		$this->featureContext->verifyTableNodeRows(
+			$permissionsTable,
+			[],
+			['share', 'edit', 'create', 'change', 'delete', 'edit', 'secure-view', 'print']
```
phil-davis: This is OK for now. But it does mean that core test code knows about the particular extra permission names in richdocuments. I  wonder how we can "pass in" the extra permission names that are allowed?

tests/acceptance/features/bootstrap/WebUINotificationsContext.php
phil-davis: @dpakach ping

tests/acceptance/features/bootstrap/WebUISharingContext.php
skshetry: I thought, the changes were regarding the missing headers case (and, that too where those headers are expected). This function does not really set the contract for what the headers should be, and therefore is not the concern of this function here. If certain arrays of data are required, those need to be verified by other functions (`setSharingPermissions` in this case). This is only making it more tightly coupled.

phil-davis: for when there are rows of settings, we also made it possible to define what "names" (strings) are allowed in the left column of a row. This helps detect accidental typos in a table of share attributes to be set, or a table with a list of permission names.

Maybe an app that wants to use this kind of step can provide some method that informs about the extra valid values?

skshetry: I see the valid use case of this in `postConditions`/`assertions`, but, not really on other places.
We do need to harden assertions functions, but, here I don't really see any gains (at least compared to the trouble).

Even here, if some attributes were **required**, I can see it's use, but this specific function does not.

phil-davis (approved):
Some CI is failing, and this is an OK first solution. Let's merge it to get CI green again in the apps.
I will make an issue to review this - #36684

PR #36691, opened 2020-01-03
Refactor runOcc calls to save occ command results

phil-davis:
## Description
Some code was directly calling `SetupHelper::runOcc` - those calls do not save the exit status, stdout and stderr from the occ command.

When that happens in a "real" step, then you cannot sensibly do `Then the command should have been successful` as the next sttep in the scenario.

Fix those to call `featureContext->runOcc` which saves the occ command result.

Directly calling `SetupHelper::runOcc` is OK in before/after scenario methods, in `Given` steps (checks for success are embedded in given steps), and `Then` steps (we do not want to overwrite the "most recent" occ command result if we have to do an occ command in a `Then` step to check something)

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -2302,35 +2302,48 @@ trait Sharing {
 	 */
 	public function administratorAddsGroupToExcludeFromSharingList($group) {
 		//get current groups
-		$occResult = SetupHelper::runOcc(
+		$occExitCode = $this->runOcc(
 			['config:app:get files_sharing blacklisted_receiver_groups']
 		);
-		//if the setting was never set before stdOut will be empty and return code will be 1
-		if (\trim($occResult['stdOut']) === "") {
-			$occResult['stdOut'] = "[]";
+		$occStdOut = $this->getStdOutOfOccCommand();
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -2302,35 +2302,48 @@ public function assertThatNoSharesAreSharedWithUser($user) {
...
 	public function administratorAddsGroupToExcludeFromSharingList($group) {
 		//get current groups
-		$occResult = SetupHelper::runOcc(
+		$occExitCode = $this->runOcc(
 			['config:app:get files_sharing blacklisted_receiver_groups']
 		);
+		$occStdOut = $this->getStdOutOfOccCommand();
+		$occStdErr = $this->getStdErrOfOccCommand();
+
+		if ($occExitCode !== "0") {
```
individual-it: should fail runOcc returns an int

individual-it (changes requested):

individual-it (approved):

PR #36698, opened 2020-01-03
Adjust read-only local storage mount scenario

phil-davis:
## Description
The scenario above this was refactored a bit in recent days so it enables external storage via the API first and then the admin browses to the admin storage settings page before doing the actions to create a local storage.

This scenario "administrator creates a read-only local storage mount" was added for the read-only storage feature. It missed the refactoring. Maybe the now-out-of-order steps will work anyway, but it will be better to be adjusted.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
@micbar it will be useful if this little scenario correction also ends up in `release-10.4.0` branch.
But maybe CI passes without this - I am not 100% sure if the scenario will be reliable without this adjustment.

individual-it (approved):

PR #36699, opened 2020-01-03
Revert "Adjusted combined given-when in WebUIAdminStorageSettings con…

phil-davis:
to remove accidentally duplicated `Given the administrator has enabled the external storage` step.

And then apply the commit from PR #36698 to fix the scenario correctly, like being done in master.


phil-davis:
Not needed any more. The reverted commit has been removed from `release-10.4.0` branch.

PR #36748, opened 2020-01-13
[Tests-only]: Remove all steps with relogs in to server

### LINKED ISSUE #36607: Refactor UI acceptance tests to not rely on the remote federated server UI
Some acceptance test scenarios login to the remote federated server and "test" the UI there. Actually the federated server can be running `daily-master-qa` or `latest` (not the code of the current PR under test). So the test is not really useful, and also the test might fail if the current PR has UI features/changes that change the UI. The test code in the current PR will not be the right code to drive the "old" UI on the federated server.

Refactor the test scenarios so that we never need to login to the remote federated server webUI. Usually the refactoring will be one of:

- turn the scenario around so that the UI actions are done on the local server, and setup (Given) or result checks (Then) are done on the remote server
- split the scenario into 2 scenarios (when it is logging in to both the local and remote server UI), so that whatever 2 (local and remote) things are being checked are done separately

dpakach:
## Description

Remove all the steps that use `re-logs in to server *`
since we are not running webui federation tests without logging in to remote server, these steps are not needed anymore.

## Related Issue

- Fixes #36607

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis (approved):
This is good if CI passes. We really should not need to use any of these steps any more.

dpakach:
> This is good if CI passes. We really should not need to use any of these steps any more.

@phil-davis Do you think there are other apps that run their own federation tests directly(not the core tests). These steps might be used there and it would be good to know about those in advance before merging this.

phil-davis:
I have all the apps with acceptance tests installed on my system. I checked with PHPstorm and it does not find any use of these steps in apps.
Merging.

Issue #36770, opened 2020-01-16
[Tests-only] Manage Context and traits in acceptance Tests


dpakach:
### Contexts and traits in acceptance tests
Currently we have Several traits such as `Provisioning`, `Webdav` etc. These contains steps which are used in almost all of the tests suites. These are available through the featureContext in tests. 
The featureContext uses BasicStructure which in turn uses all other traits such as `Provisioning`, `Webdav` etc. So Basic Structure is just a unnecessary interface in between which we can remove.
To improve the contexts and traits in acceptance tests, we need to do
- [x] Get Rid of BasicStructure
- [x] Refactor all traits into contexts except `Provisioning`, `Webdav` and `Sharing` as these are the essential features of owncloud and traits such as `Logging` and `Ip` specific features only used by some tests.
- [x] Fix `behat.yml` in core and all other apps to use newly created contexts.

dpakach:
Done so closing.
cc @phil-davis

PR #36771, opened 2020-01-16
[Tests-only] Refactor Traits and context files in acceptance tests

dpakach:
## Description

Change `Auth`, `AppConfiguration`, `Ip` and `Logging` traits in acceptance tests into contexts.

## Related Issue

- Related Issue https://github.com/owncloud/core/issues/36770

## Motivation and Context

For making the code better by using context classes rather than traits. The main advantage of traits is reusing same code in different classes whereas in our case we just dump all the traits in featureContext so it doesn't makes much sense to keep using traits for these contexts. See #36770 for more details.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


tests/acceptance/features/bootstrap/AuthContext.php
```diff
@@ -83,19 +90,26 @@ public function getAppToken() {
 	 * get the app token that was last generated
 	 * app acceptance tests that have their own step code may need to use this
 	 *
-	 * @return string app token
+	 * @return array app token
```
phil-davis:
```suggestion
	 * @return array app tokens
```
(if you need to edit again anyway, then may as well make this little change)

phil-davis (approved):

PR #36775, opened 2020-01-16
Change the wording in the additional tab to not confuse users

ChrisEdS:
## Description
This PR changes two strings that caused confusion in the additional tab.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/2500

## Motivation and Context
Confusion should be avoided.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #36776, opened 2020-01-17
Change Error to Other on settingsPage

phil-davis:
## Description
If we change `Error` to `Other` (which is already a translated string) then we avoid the translation problem. IMO "Other" is an OK title for this "Additional" settings page when it is empty.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/2500

## Motivation and Context
Confusion should be avoided.

## How Has This Been Tested?
Manual

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/36776
+++ b/changelog/unreleased/36776
@@ -1,0 +1,6 @@
+Change: Adjust wording displayed for empty additional settings panel
+
+The wording displayed when the admin personal settings panel is empty has been
+adjusted so that it no longer looks like there is an error.
+
+https://github.com/owncloud/core/pull/36776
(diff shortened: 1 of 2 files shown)


pmaier1:
Isn't this the same as https://github.com/owncloud/core/pull/36775 but with other wording?

phil-davis:
> Isn't this the same as #36775 but with other wording?

yes, the "advantage" is that the word "Other" is already used elsewhere and exists in translations. So there is no need to wait for people to do translations on Transifex.

pmaier1 (approved):
Thx

PR #36793, opened 2020-01-21
[Tests-only] Add test scenarios for updating a share when a user and group have the same name

phil-davis:
## Description
Add acceptance test scenarios that share with a user and group that have the same name, and then update sharing permissions of one of the shares.

I had to fix `isUserOrGroupInSharedData()` - it needs to know if it is supposed to be matching a user or group share.

Note: we already have test scenarios that create shares with a user and group having the same name. https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareCreateSpecial/createShareGroupAndUserWithSameName.feature

## Issue
#36813 

## Motivation and Context
There is an issue on the webUI with changing sharing permissions when a user and group have the same name. First demonstrate that the API is OK.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareUpdate/updateShareGroupAndUserWithSameName.feature
+++ b/tests/acceptance/features/apiShareUpdate/updateShareGroupAndUserWithSameName.feature
@@ -1,0 +1,50 @@
+@api @TestAlsoOnExternalUserBackend @files_sharing-app-required
+Feature: updating shares to users and groups that have the same name
+
+  Background:
+    Given using old DAV path
+    And these users have been created with default attributes and without skeleton files:
+      | username |
+      | user0    |
+      | user1    |
+      | user2    |
+    And group "user1" has been created
+    And user "user2" has been added to group "user1"
+    And user "user0" has created folder "/TMP"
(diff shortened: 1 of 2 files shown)


dpakach (commented):
LGTM :+1:

tests/acceptance/features/apiShareUpdate/updateShareGroupAndUserWithSameName.feature
```diff
@@ -0,0 +1,50 @@
+@api @TestAlsoOnExternalUserBackend @files_sharing-app-required
+Feature: updating shares to users and groups that have the same name
+
+  Background:
+    Given using old DAV path
```
individual-it: no need

individual-it (approved):

tests/acceptance/features/apiShareUpdate/updateShareGroupAndUserWithSameName.feature
phil-davis: I will remove these in another PR

PR #36796, opened 2020-01-21
Improve wording of minimum supported comment

jnweiger:
The description for the 'minimum.supported.desktop.version' parameter was misleading. The term 'supported' had multiple meanings here. This PR tries to make the language clearer. 

Can we change the name of the parameter to 'minimum.allowed.desktop.version' too? I fear this needs migration steps and has other pitfalls.

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -550,7 +550,7 @@ $CONFIG = [
  *     Note: files may not be removed if space is not required.
  * * `D, auto`
  *     keeps files and folders in the trash bin for D+ days, delete anytime if
- *     space needed (note: files may not be deleted if space is not needed)
+ *     space needed (Note: files may not be deleted if space is not needed)
  * * `auto, D`
  *     delete all files in the trash bin that are older than D days
  *     automatically, delete other files anytime if space needed
@@ -593,8 +593,8 @@ $CONFIG = [
  *
  * * `auto`
  *     default setting. Automatically expire versions according to expire
- *     rules. Please refer to :doc:`../configuration/files/file_versioning` for
- *     more information.
+ *     rules. Please refer to https://doc.owncloud.org/server/latestadmin_manual/configuration/files/file_versioning.html
+ *    for more information.
  * * `D, auto`
  *     keep versions at least for D days, apply expire rules to all versions
  *     that are older than D days
@@ -728,7 +728,7 @@ $CONFIG = [
  *                     this condition is met
  *  - `logfile`:       The log message invoked by the specified apps get redirected to
  *		               this logfile, this condition is met
- *		               Note: Not applicable when using syslog.
+ *		               Note: Not applicable when using syslog
  *
  * Defaults to an empty array
  */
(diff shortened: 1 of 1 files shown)


config/config.sample.php
```diff
@@ -1354,15 +1354,16 @@
...
- * client may not function as expected, and could lead to permanent data loss for
- * clients or other unexpected results.
+ * by the server.  
+ *
+ * As shipped, the value hee is the oldest oldest desktop client that technically is
+ * compatible with the server. The version number seen here does not imply official
+ * support or test coverage on behalf of owncloud.
+ * 
+ * NOTE: When allowing older and officially unsupported versions of the ownCloud desktop
+ * client may not function as expected, and could lead to unexpected behaviour including data loss.
```
phil-davis:
```suggestion
 * client, the client may not function as expected, and could lead to unexpected behaviour including data loss.
```

phil-davis (commented):
And do:
`make test-php-style-fix`
to correct the formatting (CI was failing on code style)

phil-davis (commented):
words LGTM

phil-davis:
@micbar what is the process for changelog with this sort of thing?
Should one be made, and what category?

phil-davis:
https://drone.owncloud.com/owncloud/core/22811/2/6
code style fails - looks like there might be white-space somewhere it is not allowed.

mmattel:
Please do a change in the text.

```
 * Define the minimum ownCloud desktop client version that is allowed to sync with
 * this server instance. ...
```
**Following background:**
The very first line of a config value description is used for building the table of contents in docs.
You can imagine that using this current proposal has a quite bad result in docs.

**Proposal:**
```
 * Define the minimum ownCloud desktop client version
 * This config value defines the minimum ownCloud desktop client version that is allowed to sync with
 * this server instance. ...
```

phil-davis:
@jnweiger ping
Are you progressing this?

mmattel:
I fixed the stuff and cleaned up some things additionally.
These were mainly links in rst style which do not work any more (we use antora) and broken links as the content has been removed in docs.

phil-davis (commented):
words LGTM

@micbar what is the process for changelog with this sort of thing?

config/config.sample.php
```diff
@@ -925,10 +925,6 @@
  *  - OC\Preview\TIFF
  *  - OC\Preview\Font
  *
- * .. note:: Troubleshooting steps for the MS Word previews are available
- *    at the :doc:`../configuration/files/collaborative_documents_configuration`
- *    section of the Administrators Manual.
- *
```
phil-davis: @mmattel did you intend to delete these lines? If so, say why.
It is not obvious to me ;0

phil-davis:
@micbar do we need a changelog for this PR?
It is just changing descriptive words in `config.sample.php`

config/config.sample.php
mmattel: As briefly written in my comment above,
* The links written were in .rst style.
These link styles (.rst format using :doc: or .adoc which is using xref:) should not be used here as this file is primary used as sample from core and secondary in docs. Therefore you need to write in case real http links.
* The deleted lines contained a link to a non existing documentation. I looked in the current docs but also in the previous (archived) documentation and did not found any usable reference.
* You may have seen, that I corrected broken links to workable once, using http links

Note: imho, a changelog may not be necessary as this only contains text fixes.

phil-davis (approved):

PR #36814, opened 2020-01-23
Initial draft for license management

jvillafanez:
## Description
Move enterprise_key features inside core. This will require some changes in the EE apps.
List of things brought by this PR:
* New grace period when an EE app is enabled.
  * You can enable any EE app without having the enterprise_key app installed, nor having a valid license key. By doing this, a grace period will start allowing to use the EE app without a license key for 24 hours, then the EE app will be disabled.
  * The grace period is common for all EE apps and it will be started only once. Once the grace period ends, no new grace period will be started (in case you want to enable a different EE app in the future).
  * A popup will be shown only for the admins to warn them about the grace period. Regular users won't see the popup. This popup will be shown in all the pages until the grace period ends or a valid license key is installed. You can use the popup to install the license key.
* License key status in the admin's setting page.
  * The status of the license (even missing) will be shown in the admin's general section of the settings page. Invalid or expired licenses will show a warning message
  * A new textbox has been added to allow setting a new license from the settings page
    * The current license won't be shown.
    * There is no validation for the license. It will be set as it is
    * The page will be reloaded after setting the license.
* A yellow notification will be shown if the license is invalid, expired or about to expire (with a 60 days of margin). No notification will be shown if the license is missing or is valid
  * The notification will be shown only for admins
* A license warning message will be displayed in the login page under some conditions:
  * If the license is missing (community edition), no license message will be displayed
  * If it's a demo license, the license message will be displayed in the login page
  * If it's a normal license, the license message will be displayed only if the license is invalid or has expired. If the license is valid or about to expire, no message will be shown in the login page.
* ownCloud's edition string (shown in the status.php) will change based on the presence of a license
 
## Related Issue

https://github.com/owncloud/enterprise/issues/3775

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


lib/private/License/LicenseManager.php
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	public function isAppUnderTrialPeriod(string $appid) {
+		$trialMark = $this->config->getAppValue('core', "trials_{$appid}", null);
+		if ($trialMark === null) {
+			$callerFile = \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1)[0]['file'];
+			if (\strpos($callerFile, $this->appManager->getAppPath($appid)) === 0) {
+				// ensure it's the app itself the one starting the trial.
+				// neither core nor other app should be able to start a trial on behalf of another app
+				$this->config->setAppValue('core', "trials_{$appid}", $this->timeFactory->getTime());
```
DeepDiver1975: try using setAppValue('trials', $appid, $time)

This way you can get all trials using the config->getAppKeys

DeepDiver1975 (changes requested):

lib/private/License/LicenseFetcher.php
```diff
@@ -0,0 +1,51 @@
...
+use OC\License\BasicLicense;
+use OCP\IConfig;
+
+/**
+ * Fetch the licenses being used.
+ * The class has the capacity to fetch the licenses from multiple places (only config for now) as
+ * well as different licenses (in case that we need to handle multiple kind of licenses at some point)
+ * and also per-app licenses (not planned for now)
+ */
+class LicenseFetcher {
```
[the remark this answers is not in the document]
jvillafanez: While I agree this seems too much for what is currently implemented, we'd need to take into account what are the plans we have for this feature.

Unless we won't ever implement additional things over this feature (as sketched in the PHP comment) I'd prefer to keep this class, otherwise I'm pretty sure the `LicenseManager` will end up bloated with more things because we won't have time to do things properly.

With what is currently under implementation, the main benefit is that we can inject this for unittests. In addition, small things usually implies easier tests.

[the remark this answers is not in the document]
DeepDiver1975: I'd split up classes later as needed - but I keep this up to you. Go ahead

lib/private/License/LicenseManager.php
```diff
@@ -0,0 +1,93 @@
...
+			}
+		} else {
+			return ILicenseManager::LICENSE_STATE_INVALID;
+		}
+	}
+
+	public function isAppUnderTrialPeriod(string $appid) {
+		$trialMark = $this->config->getAppValue('core', "trials_{$appid}", null);
+		if ($trialMark === null) {
+			$callerFile = \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 1)[0]['file'];
```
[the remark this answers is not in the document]
jvillafanez: I haven't found better alternatives to ensure the trial only starts if it's called from the app. I'd rather not rely on the "programmer's well behaviour".

If we use something like `startTrialPeriod($app)` we can't guarantee that the function will be called only by the target app: either core or any rogue app could start a trial over "myApp"

So far, the function is intended to be called only once per app, so as long as it doesn't crash it should be fine I guess. It might need some additional stress test to ensure it doesn't break, but hopefully, memory-wise it should be fine because we're ignoring the arguments (which is what it usually break the function) and we only get the first element

lib/private/License/LicenseManager.php
jvillafanez: I'll use "core-trials" then. It's even less likely there will be a collision with a valid app name.

lib/private/License/LicenseManager.php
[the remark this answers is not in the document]
DeepDiver1975: Well - who is the app developer of apps which can start a trial: it is us at ownCloud GmbH

There is not much damage caused I guess.

lib/private/License/LicenseManager.php
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	public function getGracePeriod(bool $includeExtras = false) {
+		$gracePeriod = $this->config->getAppValue('core', 'grace_period', null);
+		if ($gracePeriod === null || !$this->isNowUnderGracePeriod($gracePeriod)) {
+			return null;
+		}
+
+		if ($includeExtras) {
+			$apps = [];
```
jvillafanez: This "includeExtras" needs to be moved to a private method.

lib/private/License/LicenseManager.php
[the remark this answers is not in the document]
jvillafanez: Taking into account there will be only one "global" grace period, I think we can remove this. If an app wrongly checks the license of a different app, it shouldn't be a big deal

pmaier1:
## Suggestion for wording

**Upgrade to ownCloud Enterprise**

You have enabled one or more ownCloud Enterprise apps but your installation does not have a valid license yet.
A grace period of 24 hours has started to allow you to get going right away. Once the grace period ends, all Enterprise apps will become disabled unless you supply a valid license key.

To try ownCloud Enterprise, just [start a 30-day trial on the ownCloud Marketplace](link) and enter the provided license key below.

jvillafanez:
Screenshot from 2020-02-19 09-08-17

@pmaier1 I guess we want to include a link in "start a 30-day trial on the ownCloud Marketplace" but it isn't available yet? If so, I think we should block this until we have the link

pmaier1:
> @pmaier1 I guess we want to include a link in "start a 30-day trial on the ownCloud Marketplace" but it isn't available yet? If so, I think we should block this until we have the link

Exactly. We'll need to wait for the Marketplace rework to include the URL of the landing page there.

lib/private/License/LicenseManager.php
```diff
@@ -0,0 +1,221 @@
...
+		}
+
+		$licenseWithState = $this->getLicenseWithState($appid);
+		$licenseObj = $licenseWithState[0];
+		$licenseState = $licenseWithState[1];
+		if (!$this->isNowUnderGracePeriod($gracePeriod)) {
+			// we aren't under a grace period, so the app must be disabled if the license isn't valid
+			if ($licenseState !== ILicenseManager::LICENSE_STATE_VALID) {
+				$this->appManager->disableApp($appid);
+				return false;
```
micbar: Can we log this event?

lib/private/License/LicenseManager.php
```diff
@@ -0,0 +1,221 @@
...
+		} else {
+			if ($licenseState !== ILicenseManager::LICENSE_STATE_VALID) {
+				// we're under a grace period but the license for the app isn't valid
+				// mark the app to know there is at least one app using the grace period.
+				// we'll still return true in this case.
+				$licenseKey = '';
+				if ($licenseObj) {
+					$licenseKey = $licenseObj->getLicenseString();
+				}
+				$this->config->setAppValue('core-license-complains', $appid, $licenseKey);
```
micbar: logging?

micbar:
Tested the PR, looks good to me. IMO we need some logging. See my comments.

Next question: How would the market app play together with this change.

lib/private/License/LicenseManager.php
jvillafanez: I guess the warning message in the log is enough. No need to distinguish different license states.

lib/private/License/LicenseManager.php
jvillafanez: Using a debug log because the "complain" will be set on every request, so the log will spam too much otherwise.
We could use a "get + set" approach to prevent the spam, although not sure how efficient it will be because the cause will be executed on every request (during the grace period). It might be fine if we rely on the caching mechanism of the `config`.

jvillafanez:
Logs added.

micbar:
@pmaier1 We now need the link to the landing page.

micbar:
@jvillafanez Could you check that we also support keys in the `oc_appconfig` Table

micbar:
@enbrnz @xoxys Can you give us an opinion about changing the `config.php` from OC side after deployment?

Or is that not the case here, because you would not do a evaluation deployment with ansible?
If you do a demo deployment, you would anyway provide a demo key? Correct?

enbrnz:
> Or is that not the case here, because you would not do a evaluation deployment with ansible?
> If you do a demo deployment, you would anyway provide a demo key? Correct?

Yes, I don't think we have to worry about this in regards to ansible. When you do a deployment, people have most likely already decided whether they need CE or EE. And even if it would set a license key in the `config.php` and remove it again later, I don't think it would be a problem for ansible-deployments.

xoxys:
> don't think it would be a problem for ansible-deployments.
Shouldn't be a problem. Ansible (and I guess also docker) are using a custom-config.php

jvillafanez:
> @jvillafanez Could you check that we also support keys in the oc_appconfig Table

Fixed with the latest commit. If the config.php is marked as read-only, the license will be stored in the DB. In addition, the DB will also be checked if the license isn't found in the config.php file (license in the config.php takes precedence)

micbar:
@jvillafanez Does the PR have feature parity enterprise key app for the key expiration?

Please check again the messages and the UI notifications.

micbar:
Url for landing page should be generic.

jvillafanez:
Feature parity missing pieces:
* [ ] yellow notifications about current license (invalid, about to expire etc)
* [ ] license status in the settings page

jvillafanez:
Some clarifications regarding feature-parity with the enterprise_key app:
* Messages shown in the settings panel and in the yellow notification will be the same. As such, some messages have been slightly changed to make it clear it comes from the license part of core, specially with the yellow notification.
* There is no distinction regarding the behaviour between a normal and demo key, although the expected messages might be different.
* Yellow notification message will appear in all the pages if the license is invalid, expired or about to expire.
  * An "about to expire" license is a valid license which will expire in 60 days or less (currently hardcoded)
* Demo licenses are expected to have a 30-days valid period. This means that they'll be in a constant "about to expire" state and the yellow notification will always appear.
* Yellow notification will only be shown to admins.

phil-davis:
Changelog still needs to be added.

core/js/license-state-notification.js
```diff
@@ -0,0 +1,23 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ */
+
+$(document).ready(function() {
+	$.get(OC.generateUrl('/license/licenseMessage'), {app: 'core'})
+		.done(function(data) {
+			OC.Notification.showHtml(data.translated_message.join('<br/>'), {type: 'error'});
```
DeepDiver1975, line 21: what about escaping? more a general thought ...

lib/public/License/ILicenseManager.php
```diff
@@ -0,0 +1,122 @@
...
+
+	/**
+	 * @since 10.5.0
+	 * Get the license state for $appid. This function will return one of the LICENSE_STATE_*
+	 * constants.
+	 *
+	 * @param string $appid the id of the app to get the license for
+	 * @return int one of the LICENSE_STATE_* constants
+	 */
+	public function getLicenseStateFor(string $appid);
```
DeepDiver1975, line 75: we should define php return types where ever possible

core/js/license-state-notification.js
jvillafanez: Some of the translated messages contain a mailto link which would break if we escape the message. Since the mailto link also has a prefilled subject (which might be used for sales), I'm not sure how important is to keep the same message as before or if we can change / adjust it

jvillafanez:
todo things:
* [ ] demo license message in the login page
* [ ] revert back text for the demo messages
* [ ] recheck return types

I'll add the changelog item after having everything in place, as the last step.

micbar:
@jvillafanez Question:

Does the change affect the behavior of the status.php page?

jvillafanez:
> Does the change affect the behavior of the status.php page?

Fixed with the latest changes

I've only found the "enterprise_key" string there and in the "new" `licenseFetcher` (to keep compatibility if the license is present in the DB)

micbar:
@jvillafanez This pr might need a rebase to fix the unit tests.

jvillafanez:
https://github.com/owncloud/core/pull/36814/commits/3b02e9d07e2f948d7159a103e434a7a346dabb2b#diff-ec8c121d705ff47eacbfdd23f36a9f6cR476

The "community vs enterprise" edition string might need some additional hardening... I've followed the previous code, but now I think if the license is invalid or expired, we should change the edition string to community

micbar:
> The "community vs enterprise" edition string might need some additional hardening... I've followed the previous code, but now I think if the license is invalid or expired, we should change the edition string to community

It might help to understand, that we have 2 different use cases:
- 1) Expired demo key
- 2) Expired Production key

We always had the understanding that on a production system we do not "disable" any features. That includes the status.php IMO

For a demo key, this would be a desired feature.

@pmaier1 Please confirm

jvillafanez:
If I didn't miss anything, the following table shows the current state with all the cases. It might need some changes.
:red_circle: => message shouldn't appear
:green_circle: => message should appear

|                                                | login message  | grace popup    | notification   | settings warning | status.php |
|------------------------------------------------|:--------------:|:--------------:|:--------------:|------------------|------------|
| grace active + license missing                 | :red_circle:   | :green_circle: | :red_circle:   | normal           | community  |
| grace active + license invalid                 | :green_circle: | :green_circle: | :green_circle: | warning          | enterprise |
| grace active + license valid (demo)            | :green_circle: | :red_circle:   | :green_circle: | normal           | enterprise |
| grace active + license valid (prod)            | :red_circle:   | :red_circle:   | :red_circle:   | normal           | enterprise |
| grace active + license almost expired (demo)   | :green_circle: | :red_circle:   | :green_circle: | normal           | enterprise |
| grace active + license almost expired (prod)   | :red_circle:   | :red_circle:   | :green_circle: | normal           | enterprise |
| grace active + license expired (demo)          | :green_circle: | :green_circle: | :green_circle: | warning          | enterprise |
| grace active + license expired (prod)          | :green_circle: | :green_circle: | :green_circle: | warning          | enterprise |
|                                                |                |                |                |                  |            |
| grace inactive + license missing               | :red_circle:   | :red_circle:   | :red_circle:   | normal           | community  |
| grace inactive + license invalid               | :green_circle: | :red_circle:   | :green_circle: | warning          | enterprise |
| grace inactive + license valid (demo)          | :green_circle: | :red_circle:   | :green_circle: | normal           | enterprise |
| grace inactive + license valid (prod)          | :red_circle:   | :red_circle:   | :red_circle:   | normal           | enterprise |
| grace inactive + license almost expired (demo) | :green_circle: | :red_circle:   | :green_circle: | normal           | enterprise |
| grace inactive + license almost expired (prod) | :red_circle:   | :red_circle:   | :green_circle: | normal           | enterprise |
| grace inactive + license expired (demo)        | :green_circle: | :red_circle:   | :green_circle: | warning          | enterprise |
| grace inactive + license expired (prod)        | :green_circle: | :red_circle:   | :green_circle: | warning          | enterprise |

micbar:
@jvillafanez Would it be possible to have a call-to-action button "Get your demo key". for the link to the landing page?

jvillafanez:
> @jvillafanez Would it be possible to have a call-to-action button "Get your demo key". for the link to the landing page?

The link is already clickable. The only problem is that with the default style the link could be unnoticed.

micbar:
@jvillafanez The link is clickable yes.

PM requested that we use a button in a new line.

micbar:
Screenshot_2020-05-19 Files - ownCloud

@hodyroff @pmaier1 

OK.

LGTM 👍

micbar (approved):

DeepDiver1975 (approved):

jvillafanez:
This is ready to merge

jvillafanez:
hopefully last touches:
* [ ] open " get a demo key" link in a new tab +  change target link to "owncloud.com/try-enterprise"

phil-davis:
@jvillafanez if you rebase and force-push then you should get codecov results/status back. That problem has been fixxed today in latest master.

jvillafanez:
> 2. Config (can come later, but nice to have if small thing)
> if there is no config, use default url: https://owncloud.com/try-enterprise/
> if there is a configured url, use that.

I'll open a new ticket for that in order to merge this PR today.

PR #36829, opened 2020-01-27
Allow expiration date to be today

### LINKED ISSUE #36569: default expire date config parameters can be set to 0 days

phil-davis:
## Description
The expiration date validation code had some issues when the server was not in UTC. 

The proposed expiration date was in a `DateTime` object that had the time-zone of the server - e.g. it represented `2020-01-27:00:00:00+01:00`. 

The date-time was being compared with "current day" - a new default `DateTime` object that had the time zeroed, but that object was in UTC time - e.g. `2020-01-27:00:00:00+00:00` - but in the local timezone that is actually `2020-01-27:01:00+01:00`

And so the proposed expiration date was 1 hour before the "current day" and the request would return "Expiration date is in the past".

- adjust the webUI JS date-picker so it allows "today"  to be selected
- add API acceptance tests that set share expiration to "today" for various combinations of default/enforced expiration

## Related Issue
- Fixes #36569 

## How Has This Been Tested?
CI an locally using the date-picker in webUI to choose "today"

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/js/sharedialoglinkexpirationview.js
+++ b/core/js/sharedialoglinkexpirationview.js
@@ -115,8 +115,6 @@
 			// what if there is another date picker on that page?
 			var minDate = new Date();
 			var maxDate = null;
-			// min date should always be the next day
-			minDate.setDate(minDate.getDate()+1);
 
 			if(isExpirationSet) {
 				if(isExpirationEnforced) {
--- a/core/js/sharedialogshareelistview.js
+++ b/core/js/sharedialogshareelistview.js
@@ -500,7 +500,7 @@
 			var $el = $(el);
 
 			$el.datepicker({
-				minDate: "+1d",
+				minDate: "+0d",
 				dateFormat : 'dd-mm-yy',
 				onSelect : function() {
 					self._onExpirationChange(el);
--- a/core/js/tests/specs/sharedialogexpirationviewSpec.js
+++ b/core/js/tests/specs/sharedialogexpirationviewSpec.js
@@ -122,7 +122,7 @@ describe('OC.Share.ShareDialogLinkExpirationView', function() {
 			beforeEach(function() {
 				// pick a fake date
 				clock = sinon.useFakeTimers(new Date(2014, 0, 20, 14, 0, 0).getTime());
-				expectedMinDate = new Date(2014, 0, 21, 14, 0, 0);
+				expectedMinDate = new Date(2014, 0, 20, 14, 0, 0);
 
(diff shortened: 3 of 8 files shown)


core/js/sharedialoglinkexpirationview.js
```diff
@@ -116,7 +116,7 @@
 			var minDate = new Date();
 			var maxDate = null;
 			// min date should always be the next day
-			minDate.setDate(minDate.getDate()+1);
+			minDate.setDate(minDate.getDate());
```
individual-it: what does this thing even do? get the date and set it to it self?

phil-davis:
@micbar this is adjusting (fixing) the sharing API so it reliably accepts an expiration date of "today" for a user/group share. And adjusts the webUI date picker so that it is possible in the webUI to actually choose "today".

Please select an appropriate person to review the JS and PHP code changes.

@individual-it please review the acceptance test changes.

individual-it (dismissed):

core/js/sharedialoglinkexpirationview.js
phil-davis: Maybe it now does nothing. I claim zero JS knowledge! I just removed the `+1`
And tthe comment about is now rubbish - I better at least adjust/remove that!

phil-davis: removed this line and the comment above

micbar:
@phil-davis How would this affect the clients, if they don't set a value because they know nothing about expiration date?

phil-davis:
> @phil-davis How would this affect the clients, if they don't set a value because they know nothing about expiration date?

This PR only adjusts the existing code for user/group share expiry that is already in master for 10.4 There should be no effect on clients by any of this expiry stuff. If the default expiry is not set, then clients can optionally supply an expiry date (that would be for the clients to add in future), but when not mentioning a expiry date then no expiry date will be set. If a default expiry date is enabled (and also if enforced...) then the expiry date happens by default without the client needing to do anything.

individual-it (commented):
tests look OK

micbar (approved):

PR #36750, opened 2020-01-13
[Tests-only] Add functions to create/delete/manage ldap users and groups

kiranparajuli589:
## Description
Previously, user creation were skipped and pre-created users were used when running ldap tests.
Function to create user in ldap server are added.

## How Has This Been Tested?
- CI
- manually

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #36840, opened 2020-01-28
[Tests-only] have a single after scenario to clean users and groups

individual-it:
## Description
use a single after scenario to do the cleanup of users and groups

restore parameters need to happen before user/group deletion
deleting users in the DB does not need to happen when running on LDAP

follow up for #36750

## Motivation and Context
make sure actions are done in the correct order and in the correct cases

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis (approved):

dpakach (approved):
LGTM :+1:

PR #36854, opened 2020-01-30
[Tests-only] Fix running fed tests with ldap backend

dpakach:
## Description

Dont create same multiple times when running federation test with ldap.
Skip autocompletion tests that fail when using LDAP backend due to medial search.

## How Has This Been Tested?

- Locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/webUISharingAutocompletion/shareAutocompletion.feature
+++ b/tests/acceptance/features/webUISharingAutocompletion/shareAutocompletion.feature
@@ -29,6 +29,7 @@ Feature: Autocompletion of share-with names
       | finance3      |
       | users-finance |
       | other         |
+    And the administrator has added system config key "user_ldap.enable_medial_search" with value "true" and type "boolean"
 
   @smokeTest
   Scenario: autocompletion of regular existing users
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/webUISharingAutocompletion/shareAutocompletion.feature
```diff
@@ -238,6 +239,7 @@ Feature: Autocompletion of share-with names
     When the user types "de" in the share-with-field
     Then only user "John Finn Smith" should be listed in the autocomplete list on the webUI
 
+  @skipOnLDAP
```
phil-davis: does this fail because we do not have a working way to reliably sync groups from LDAP?
or?
Is there an issue that we can put an issue tag here?

phil-davis: OK, this test sets the medial search config for local groups. So it is not expected to work for LDAP groups (which have a different config setting)

phil-davis (approved):

PR #36855, opened 2020-01-30
Php7.4 20200130

phil-davis:
Try running tests with PHP 7.4


phil-davis:
Rebased so it uses the new `.drone.star`

phil-davis:
Note:
litmus and webDAV problem with object __toString conversion exception - see https://github.com/owncloud/core/issues/32670#issuecomment-565080211

PR #36872, opened 2020-02-03
Fix translation for the personal sharing text

sharidas:
Fix translation for personal sharing text.

## Description

Add tranlsation for the text in Persnoal->Sharing page.

## Related Issue

- Fixes 36869

## Motivation and Context

Add translation for the text in Personal->Sharing page.

## How Has This Been Tested?

- This requires transifex to run for testing.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
+++ b/apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
@@ -23,6 +23,7 @@ namespace OCA\Files_Sharing\Panels\Personal;
 
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\IConfig;
+use OCP\IL10N;
 use OCP\IUserSession;
 use OCP\Settings\ISettings;
 use OCP\Template;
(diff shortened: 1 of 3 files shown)


apps/files_sharing/lib/Panels/Personal/PersonalPanel.php
```diff
@@ -63,6 +68,7 @@ public function getPanel() {
 		$tmpl = new Template('files_sharing', 'settings-personal');
 		$enabledConfigs = [];
 		foreach (self::USER_CONFIGS as $config) {
+			$this->l10n->t($config['label']);
```
DeepDiver1975: This is just plain wrong! gettext can only extract strings which are within the t() function. Using variables does not work.

DeepDiver1975 (changes requested):

sharidas:
Ok. thanks for the clarification. Unfortunately this PR is closed which I cannot open. Before assigning the value (text) to the key `label`, inside the `getPanel` we can use the t() function with the text. Instead of using variable.

PR #36896, opened 2020-02-06
[Tests-Only] bump test tool dependencies

phil-davis:
## Description
- mention the current minor versions of tools in `vendor-bin` because actually we do use these versions, and might now be depending on them (this will not actually change the versions that get used at test-time, those latest minor versions will be being installed now anyway)
- bump symfony to 4.4 (we did that in the main core `composer.json` the other day)

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/vendor-bin/behat/composer.json
+++ b/vendor-bin/behat/composer.json
@@ -1,14 +1,14 @@
 {
     "require": {
-        "behat/behat": "^3.0",
-        "behat/mink-extension": "^2.2",
+        "behat/behat": "^3.6",
+        "behat/mink-extension": "^2.3",
         "behat/mink-goutte-driver": "^1.2",
         "behat/mink-selenium2-driver": "dev-master",
         "jarnaiz/behat-junit-formatter": "^1.3",
         "rdx/behat-variables": "^1.2",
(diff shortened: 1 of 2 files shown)


vendor-bin/behat/composer.json
```diff
@@ -1,14 +1,14 @@
...
+        "behat/mink-extension": "^2.3",
         "behat/mink-goutte-driver": "^1.2",
         "behat/mink-selenium2-driver": "dev-master",
         "jarnaiz/behat-junit-formatter": "^1.3",
         "rdx/behat-variables": "^1.2",
-        "sensiolabs/behat-page-object-extension": "^2.0",
-        "symfony/translation": "^3.4",
-        "sabre/xml": "^2.1",
+        "sensiolabs/behat-page-object-extension": "^2.3",
+        "symfony/translation": "^4.4",
```
phil-davis: We have moved to Symfony4 in the "real" core. So we might as well do it here.
This is the change that could actually make a difference.

dpakach (approved):

PR #36921, opened 2020-02-10
Consider file to be encrypted only if it has encrypted property set

VicDeo:
## Description
Consider a file to be encrypted only if it has encrypted property set to true

## Related Issue
https://github.com/owncloud/enterprise/issues/1739#issuecomment-393223048

## Motivation and Context
It's not possible to download file that is encrypted externally

## How Has This Been Tested?
0. Setup encryption on a separate instance
1. upload any encrypted file to another ownCloud instance where the encryption is not active
2. Try to download this file
### Expected
File is downloaded

### Actual
Web UI shows an error:
```
Forbidden
Encryption not ready: Module with id: OC_DEFAULT_MODULE does not exist.
```

## Types of changes

- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -392,8 +392,11 @@ class Encryption extends Wrapper {
 				// specified in the file header - otherwise we need to fail hard to
 				// prevent data loss on client side
 				if (!empty($encryptionModuleId)) {
-					$targetIsEncrypted = true;
-					$encryptionModule = $this->encryptionManager->getEncryptionModule($encryptionModuleId);
+					$data = $this->storage->getCache()->get($path);
+					if (isset($data['encrypted']) && $data['encrypted'] === true) {
+						$targetIsEncrypted = true;
+						$encryptionModule = $this->encryptionManager->getEncryptionModule($encryptionModuleId);
+					}
 				}
 
 				if ($this->file_exists($path)) {
--- a/changelog/unreleased/36921
+++ b/changelog/unreleased/36921
@@ -1,0 +1,7 @@
+Bugfix: It's not possible to download externally encrypted files
+
+Downloading was failing with the message "Encryption not ready: Module
+with id: OC_DEFAULT_MODULE does not exist." if the file was encrypted
+with another ownCloud instance.
+
+https://github.com/owncloud/core/pull/36921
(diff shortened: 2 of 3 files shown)


lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -392,8 +392,13 @@ public function fopen($path, $mode) {
 				// specified in the file header - otherwise we need to fail hard to
 				// prevent data loss on client side
 				if (!empty($encryptionModuleId)) {
-					$targetIsEncrypted = true;
-					$encryptionModule = $this->encryptionManager->getEncryptionModule($encryptionModuleId);
+					$data = $this->storage->getCache()->get($path);
+					if ($data) {
```
jvillafanez:
```
$data = $this->storage->getCache()->get($path);
if (isset($data['encrypted']) && $data['encrypted'] === true) {
  $encryptionModule = $this->encryptionManager->getEncryptionModule($encryptionModuleId);
}
```
It seems a bit better to me. I'm not sure if we need the `$targetIsEncrypted` but we can assign it inside the condition if needed.

We might also need a comment with a little explanation, because accessing the storage cache could be expensive. I don't think we can avoid it at the moment, but maybe someone could have a better idea in the future.

VicDeo: @jvillafanez 
> accessing the storage cache could be expensive

True. But it happens only if we have found an encrypted header in the line 383
As for replacing `$targetIsEncrypted` with  `$encryptionModule` - this var is matches it's content perfectly. 
In addition it was not added by me and is used below this point several times.

jvillafanez: > As for replacing `$targetIsEncrypted` with `$encryptionModule` - this var is matches it's content perfectly.

I didn't want to replace it. I just didn't see where the `$targetIsEncrypted` was being used
```
$data = $this->storage->getCache()->get($path);
if (isset($data['encrypted']) && $data['encrypted'] === true) {
  $targetIsEncrypted = true;
  $encryptionModule = $this->encryptionManager->getEncryptionModule($encryptionModuleId);
}
```
I can live with the above

VicDeo: thanks for the clarification. updated.

jvillafanez (approved):

PR #36947, opened 2020-02-12
Revert "Fix for app:enable in case of not valid/not existing enterpri…

micbar:
## Description
Revert https://github.com/owncloud/core/pull/36399

## Reason
- CI could not enable apps from the CLI
```su-exec www-data php /var/www/owncloud/occ app:enable -n testing
testing cannot be enabled because of invalid enterprise key```
- The enterprise check for a license key is disabled in CI
- This check on the cli command breaks the current behavior

## Further Reasoning
- In the future, this function will be not needed anymore


phil-davis:
@micbar can we write some words about why this needs to be reverted - then if someone looks at it again for a future release they will have some idea what the problem is.

micbar:
@phil-davis

PR #36948, opened 2020-02-13
[Tests-only] Do not use occ context from behat.yml in PublicWebDavContext

dpakach:
## Description

Use 
```php
$this->occContext = new OccContext();
$environment->registerContext($this->occContext);
```

instead of 
```php
$environment->getContext('OccContext');
```
in `PublicWebDavContext.php`

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/PublicWebDavContext.php
+++ b/tests/acceptance/features/bootstrap/PublicWebDavContext.php
@@ -956,6 +956,7 @@ class PublicWebDavContext implements Context {
 		$environment = $scope->getEnvironment();
 		// Get all the contexts you need in this context
 		$this->featureContext = $environment->getContext('FeatureContext');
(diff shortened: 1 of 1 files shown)


phil-davis (commented):
This is a change in `PublicWebDavContext`
I don't understand why this change is needed/good.

individual-it (approved):
This makes the Context being registered even when not mentioned in the behat.yml

dpakach:
Looks like this way we have to call all the before and after scenario manually from the contexts. Just putting the contexts in `behat.yml` and using the old method is relatively easier.

Issue #36942, opened 2020-02-12
Support PHP 7.4 in acceptance test infrastructure


phil-davis:
The Behat acceptance test infrastructure runs separately to the system-under-test. Nowadays it gets used for running acceptance tests against other backends as well as the regular ownCloud10.

Update the dependencies for the Behat acceptance test infrastructure so it will support PHP 7.4

See issue #36509 for the other task - to support PHP 7.4 for the ownCloud10 server.

Things to check when doing this:
- [x] tests work fine in drone against ownCloud10
- [x] tests run fine locally when using just a PHP dev server and running the test infrastructure out of the same clone of the repo
- [x] tests run fine locally when the git-clone-under-test is running with an Apache web server user `www-data`

phil-davis:
Note: https://drone.owncloud.com/owncloud/core/23222/4/4
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - sensiolabs/behat-page-object-extension v2.3.0 requires php >=7.1.3,<7.4 -> your PHP version (7.4.1) does not satisfy that requirement.
    - sensiolabs/behat-page-object-extension v2.3.1 conflicts with guzzlehttp/guzzle[5.3.0].
    - sensiolabs/behat-page-object-extension v2.3.1 conflicts with guzzlehttp/guzzle[5.3.1].
    - sensiolabs/behat-page-object-extension v2.3.1 conflicts with guzzlehttp/guzzle[5.3.2].
    - sensiolabs/behat-page-object-extension v2.3.1 conflicts with guzzlehttp/guzzle[5.3.3].
    - sensiolabs/behat-page-object-extension v2.3.1 conflicts with guzzlehttp/guzzle[5.3.4].
    - Installation request for sensiolabs/behat-page-object-extension ^2.3 -> satisfiable by sensiolabs/behat-page-object-extension[v2.3.0, v2.3.1].
    - Installation request for guzzlehttp/guzzle ^5.3 -> satisfiable by guzzlehttp/guzzle[5.3.0, 5.3.1, 5.3.2, 5.3.3, 5.3.4].

make: *** [vendor-bin/behat/vendor] Error 2
Makefile:402: recipe for target 'vendor-bin/behat/vendor' failed
```
This is the primary reason for this issue.

https://github.com/sensiolabs/BehatPageObjectExtension/pull/128/files brought support for PHP 7.4

[...]

PR #36959, opened 2020-02-17
[Tests-Only] Set shareapi_exclude_groups_list to the empty string at the start of acceptance tests

phil-davis:
## Description
At the start of each acceptance test scenario we set various sharing settings to some default values. Most of those are true or false settings which end up getting the value `yes` or `no` sent through to the testing app which puts those in the core settings - all good.

`shareapi_exclude_groups_list` is a formatted string of group names that are excluded from sharing. We have been setting it to `null`. With Guzzle5 that has been getting through in the request body to the testing app as `value` the empty string, which is actually what we want.

With Guzzle6 the body gets constructed differently items like `value` that are specified `null` just do not end up at all in the request.

IMO this is an unseen bug in the current test code, so fix it now.

## Related Issue
#36942 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -2606,7 +2606,7 @@ trait Sharing {
 				'capabilitiesParameter' => 'exclude_groups_from_sharing_list',
 				'testingApp' => 'core',
 				'testingParameter' => 'shareapi_exclude_groups_list',
-				'testingState' => null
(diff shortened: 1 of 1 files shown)


individual-it:
yarn install issue in CI, restarted build

individual-it (approved):

PR #32537, opened 2018-08-31
Update Guzzle to 6.3

phil-davis:
For ref: https://github.com/guzzle/guzzle/blob/master/UPGRADING.md

PR #37008, opened 2020-02-24
[WIP] Guzzle6 for PHP 74

phil-davis:
- back-end code for Guzzle6 taken as-is from PR #32537 
- on top of code for PHP 7.4

See what happens here - will need adjusting, I am sure.


phil-davis:
See #37011

Issue #32357, opened 2018-08-16
Acceptance Tests for case-sensitive file names


phil-davis:
Add acceptance test scenarios to API and webUI for things like:
- [ ] renaming (moving) a file to be the same name, with just case different
- [ ] moving a file (changing the name) "over the top" of an existing file that just has different case
- [ ] moving a file (into another folder) "over the top" of an existing file that just has different case
- [ ] deleting a file, when 2 files exist with different case
- [ ] copying a file, when 2 files exist with different case
- [ ] downloading a file, when 2 files exist with different case
- [ ] do similar with folders
- [ ] share to a sharee that has an existing file/folder with different case
- [ ] mixup file and folder, e.g. file called MyStuff and folder called mystuff

ownclouders:
[GitMate.io](https://gitmate.io) thinks the contributor most likely able to help you is @PVince81.

Possibly related issues are https://github.com/owncloud/core/issues/31579 (test), https://github.com/owncloud/core/issues/22492 (test), https://github.com/owncloud/core/issues/22485 (test), https://github.com/owncloud/core/issues/1314 (Case sensitive user name folders causing files to disappear), and https://github.com/owncloud/core/pull/31516 (Acceptance test coding standard changes).

phil-davis:
Needs someone to look at what useful test scenarios remain to be added.

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

This issue was opened before 2020 and has had no activity since then, so we are closing it as outdated. If this is still relevant on a current ownCloud release, please **reopen** (or open a fresh issue) with up-to-date details — thank you!

PR #37011, opened 2020-02-25
[WIP] Guzzle6 for core

phil-davis:
## Description
- [x] get adjusted code from old PR #32357 

## Motivation and Context
Guzzle5 does not have official support on PHP 7.4 and probably never will have. It is likely that Guzzle5 will work on PHP  7.4 but will emit various warning/notice messages about deprecated code.

So it could be "easier/better" to move to Guzzle6 in ownCloud10.

Note1: if we do this then apps that use Guzzzle5 will also need updating to Guzzle6 and coordinated release of core and app versions etc.

Note2: it seems that we do not have to do this for PHP 7.4

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
See PR #38266

PR #37022, opened 2020-02-26
Show pending remote shares at the 'Shared with you' tab

VicDeo:
## Description
- [x] Remove deprecated static API class usage in favor of OcsController
- [x] Display pending remote shares at the `Shared with you` tab
- [x] Adapt accept/decline process to the remote shares

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3793

## Motivation and Context
Shared with should include pending remote shares

## How Has This Been Tested?
1. A create several  federated shares
2. As a recipient browse to the `shared with you` tab and accept/decline them

shared with you

## Types of changes
- [x] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/js/sharedfilelist.js
+++ b/apps/files_sharing/js/sharedfilelist.js
@@ -244,7 +244,7 @@
 
 			if (!!this._sharedWithUser) {
 				var remoteShares = $.ajax({
-					url: OC.linkToOCS('apps/files_sharing/api/v1') + 'remote_shares',
+					url: OC.linkToOCS('apps/files_sharing/api/v1') + 'remote_shares/all',
 					/* jshint camelcase: false */
 					data: {
 						format: 'json',
@@ -339,7 +339,7 @@
 				.map(function(share) {
 					var file = {
 						shareOwner: share.owner + '@' + share.remote.replace(/.*?:\/\//g, ""),
-						shareState: share.accepted ? OC.Share.STATE_ACCEPTED : OC.Share.STATE_PENDING,
+						shareState: !!parseInt(share.accepted) ? OC.Share.STATE_ACCEPTED : OC.Share.STATE_PENDING,
 						name: OC.basename(share.mountpoint),
 						mtime: share.mtime * 1000,
 						mimetype: share.mimetype,
--- a/apps/files_sharing/js/sharedfilelist.js
+++ b/apps/files_sharing/js/sharedfilelist.js
@@ -339,7 +339,7 @@
 				.map(function(share) {
 					var file = {
 						shareOwner: share.owner + '@' + share.remote.replace(/.*?:\/\//g, ""),
-						shareState: !!parseInt(share.accepted) ? OC.Share.STATE_ACCEPTED : OC.Share.STATE_PENDING,
+						shareState: !!parseInt(share.accepted, 10) ? OC.Share.STATE_ACCEPTED : OC.Share.STATE_PENDING,
 						name: OC.basename(share.mountpoint),
 						mtime: share.mtime * 1000,
 						mimetype: share.mimetype,
--- a/apps/files_sharing/lib/AppInfo/Application.php
+++ b/apps/files_sharing/lib/AppInfo/Application.php
@@ -28,9 +28,12 @@
 namespace OCA\Files_Sharing\AppInfo;
 
 use OC\AppFramework\Utility\SimpleContainer;
+use OC\Files\Filesystem;
+use OCA\Files_Sharing\Controller\RemoteOcsController;
 use OCA\Files_Sharing\Controller\Share20OcsController;
 use OCA\Files_Sharing\Controllers\ExternalSharesController;
 use OCA\Files_Sharing\Controllers\ShareController;
+use OCA\Files_Sharing\External\Manager;
 use OCA\Files_Sharing\Middleware\SharingCheckMiddleware;
 use OCA\Files_Sharing\MountProvider;
 use OCA\Files_Sharing\Notifier;
@@ -94,6 +97,24 @@ class Application extends App {
 				$c->query(NotificationPublisher::class),
 				$server->getEventDispatcher(),
 				$c->query(SharingBlacklist::class)
+			);
+		});
+
+		$container->registerService('RemoteOcsController', function (SimpleContainer $c) use ($server) {
+			$user = $server->getUserSession()->getUser();
+			$uid = $user ? $user->getUID() : null;
+			return new RemoteOcsController(
+				$c->query('AppName'),
+				$server->getRequest(),
+				new Manager(
+					$server->getDatabaseConnection(),
+					Filesystem::getMountManager(),
+					Filesystem::getLoader(),
+					$server->getNotificationManager(),
+					$server->getEventDispatcher(),
+					$uid
+				),
+				$uid
 			);
 		});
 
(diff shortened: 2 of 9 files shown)


apps/files_sharing/lib/AppInfo/Application.php
```diff
@@ -97,6 +100,24 @@ public function __construct(array $urlParams = []) {
...
+				$server->getRequest(),
+				new Manager(
+					$server->getDatabaseConnection(),
+					Filesystem::getMountManager(),
+					Filesystem::getLoader(),
+					$server->getNotificationManager(),
+					$server->getEventDispatcher(),
+					$uid
+				),
+				$uid
```
jvillafanez: I'd prefer to inject the userSession object in order to follow what has been done in other places, but taking into account that the "Manager" being injected uses the uid, it might be problematic: if the userSession changes, the manager would still use the previous user id.
In addition, I'm not sure how it will behave under this "change the user session" scenario with the Filesystem objects being injected.

Worst case, we could end up with a static controller, but being a controller it might not be a big problem

apps/files_sharing/lib/Controller/RemoteOcsController.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @NoAdminRequired
+	 * @param int $id
+	 *
+	 * @return Result
+	 */
+	public function acceptShare($id) {
+		if ($this->externalManager->acceptShare((int) $id)) {
+			$share = $this->externalManager->getShare($id);
+			return new Result(
+				[[
```
jvillafanez: This might need a code comment because it seems a bit weird. I guess you want to return a list of "state + file_target" objects, with only one element in this case. Reformatting this a bit, might also help, as `[[` and `]]` might be misinterpreted.

apps/files_sharing/lib/Controller/RemoteOcsController.php
```diff
@@ -0,0 +1,227 @@
...
+					'state' => Share::STATE_ACCEPTED,
+					'file_target' => $share['mountpoint']
+				]]
+			);
+		}
+
+		// Make sure the user has no notification for something that does not exist anymore.
+		$this->externalManager->processNotification((int) $id);
```
jvillafanez: At the moment, it's a bit unclear if the `processNotification` will be called (or should be called) if the share is accepted.
If the function is automatically called after the share is accepted, include a code comment here.

apps/files_sharing/lib/Controller/RemoteOcsController.php
```diff
@@ -0,0 +1,227 @@
...
+	 *
+	 * @return Result
+	 */
+	public function declineShare($id) {
+		if ($this->externalManager->declineShare((int) $id)) {
+			return new Result();
+		}
+
+		// Make sure the user has no notification for something that does not exist anymore.
+		$this->externalManager->processNotification((int) $id);
```
jvillafanez: Same here.

apps/files_sharing/lib/Controller/RemoteOcsController.php
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	/**
+	 * @NoCSRFRequired
+	 * @NoAdminRequired
+	 *
+	 * @return Result
+	 */
+	public function getAllShares() {
+		return $this->getShares(true);
```
jvillafanez: Technically, I'd prefer to have a private method such as `getSharesPrivate($includePending=false)` in order to make both functions `getShares` and `getAllShares` to explicitly return a Result.

apps/files_sharing/lib/Controller/RemoteOcsController.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @return Result
+	 */
+	public function unshare($id) {
+		$shareInfo = $this->externalManager->getShare($id);
+
+		if ($shareInfo === false) {
+			return new Result(null, 404, 'Share does not exist');
+		}
+
+		$mountPoint = '/' . $this->uid . '/files' . $shareInfo['mountpoint'];
```
jvillafanez: String interpolation is preferred if it's easy to do:
```
$mountPoint = "/{$this->uid}/files{$shareInfo['mountpoint']}";
```

apps/files_sharing/lib/Controller/RemoteOcsController.php
```diff
@@ -0,0 +1,227 @@
...
+		return $share;
+	}
+
+	/**
+	 * @param string $mountpoint
+	 * @return false|\OC\Files\FileInfo
+	 * @throws \Exception
+	 */
+	protected function getFileInfo($mountpoint) {
+		$view = new View('/' . $this->uid . '/files/');
```
jvillafanez: Any better alternative? Maybe we can inject an `IRootFolder` and get the info from there? This might be a problem for testing.

VicDeo: Not a problem for testing as this method was added exactly to make testing possible.
Moreover, RemoteOcsController has coverage 95.23% is based on the deleted `apps/files_sharing/lib/API/Remote.php` that had 0% coverage.

apps/files_sharing/lib/AppInfo/Application.php
VicDeo: tried that, not possible - everything breaks down.

apps/files_sharing/lib/Controller/RemoteOcsController.php
VicDeo: Make sure the user has no notification for **something that does not exist anymore.**

VicDeo: when the share is accepted we don't care about notification.

apps/files_sharing/lib/Controller/RemoteOcsController.php
VicDeo: same as above. this is a sanity measure 'for something that does not exist anymore'

apps/files_sharing/js/sharedfilelist.js
```diff
@@ -339,7 +339,7 @@
 				.map(function(share) {
 					var file = {
 						shareOwner: share.owner + '@' + share.remote.replace(/.*?:\/\//g, ""),
-						shareState: share.accepted ? OC.Share.STATE_ACCEPTED : OC.Share.STATE_PENDING,
+						shareState: !!parseInt(share.accepted) ? OC.Share.STATE_ACCEPTED : OC.Share.STATE_PENDING,
```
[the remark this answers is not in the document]
VicDeo: done

apps/files_sharing/lib/Controller/RemoteOcsController.php
VicDeo: reformatted, comment added

apps/files_sharing/lib/Controller/RemoteOcsController.php
VicDeo: done

apps/files_sharing/lib/Controller/RemoteOcsController.php
VicDeo: Not essential IMO

jvillafanez (approved):

PR #37025, opened 2020-02-27
[Tests-Only] Run acceptance tests with Oracle

phil-davis:
To see what passes or not.


phil-davis:
Issues #37026 and #37027 raised.
Everything else looks OK.

Issue #37028, opened 2020-02-27
database name in .drone.star should be passed through to tests/drone/install-server.sh


phil-davis:
e.g. see PR #37021 

If I change the database name returned by `.drone.star` `getDbDatabase()` then CI does not work. The database name is also hard-coded in `tests/drone/install-server.sh`

- [x] Pass the database name `DB_NAME` into `tests/drone/install-server.sh` (e.g. by setting env var...)
- [x] Pass any other parameters also (e.g. `DB_USERNAME` `DB_PASSWORD`)

phil-davis:
Fixed by #37031

PR #37032, opened 2020-02-27
[Tests-Only] Test pass settings to install server 2

phil-davis:
Issue #37028 

This PR verifies that the DbUsername, DbPassword and DbDatabase can be changed in `.drone.star` and that those values get passed successfully to `install-server.sh` and that the install works and CI continues to pass.


phil-davis:
works, good.

PR #37038, opened 2020-02-27
Initialize the user before transfer-ownership

jvillafanez:
## Description
Initialize the user's FS if he hasn't logged in yet before transferring files to him.

## Related Issue

https://github.com/owncloud/enterprise/issues/3819

## Motivation and Context
Transfer was failing silently. The command output didn't show any error but the transfer failed

## How Has This Been Tested?

2. Create 2 users, user1 and user2 (local users)
3. Upload some files to user1
4. Without having logged in once with user2, execute `occ files:transfer-ownership user1 user2`
5. Files are transferred correctly to the user2

Checked also with master-key encryption **but not user-key encryption**

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37038
+++ b/changelog/unreleased/37038
@@ -1,0 +1,9 @@
+Bugfix: Initialize the user before the transfer command
+
+Trying to transfer the ownership of files to a user who hadn't logged in was causing problems
+because the FS of such user wasn't initialized and it wasn't possible to move the files there.
+The command appeared to work, but the files weren't moved.
+
+Now such user has the FS initialized so the transfer can be completed normally.
+
+https://github.com/owncloud/core/pull/37038
--- a/apps/files/lib/Command/TransferOwnership.php
+++ b/apps/files/lib/Command/TransferOwnership.php
@@ -121,12 +121,6 @@ class TransferOwnership extends Command {
 				null,
 				InputOption::VALUE_REQUIRED,
 				'selectively provide the path to transfer. For example --path="folder_name"'
-			)
-			->addOption(
-				'force-user-init',
-				null,
-				InputOption::VALUE_NONE,
-				'force the initialization of the destination user if he hasn\'t logged in yet'
 			);
 	}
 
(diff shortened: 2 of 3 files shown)


jvillafanez:
@phil-davis could you check the failing test? It seems to be caused by https://github.com/owncloud/core/blob/master/tests/lib/UtilTest.php#L388 and https://github.com/owncloud/core/blob/master/tests/lib/UtilTest.php#L411

I think the `$config->deleteSystemValue('skeletondirectory');` isn't being called due to the expected exception being thrown in those tests. This causes the system value not being properly reset, which affects this PR's tests.

phil-davis:
@jvillafanez I pushed a commit that should fix that. Let's see how CI goes!

I also put the fix into a separate PR #37041 because it would be good to fix that UtilTest straight away, rather than waiting until this PR gets finished.

I wonder how many other places in the unit tests have "cleanup" code in a test case after an expected exception? Cleanup code should all go in the `tearDown()` method.

phil-davis:
@jvillafanez If you pull this branch locally, rebase and force push, it should get rid of that 2nd commit - the `UtilTest` fixes are now merged in `master`

jvillafanez:
I'll add the changelog entry after the review.

I have my doubts about including and using the `--force-user-init`. The only use case I see for the option is that you want to verify the destination user has logged in at least once. However, I think that normally you just want to transfer to that user.
I mean, we can remove the option and initialize the user without user confirmation.

micbar:
@carlos @pako81 Can you confirm?

pako81:
Having the `--force-user-init` input option would be maybe more transparent for the admin, to give him/her the information that the destination user has not log in yet.

On the other hand, final admin's aim is to just transfer files so, yes, I do agree with @jvillafanez that we could probably get rid of it.

changelog/unreleased/37038
```diff
@@ -0,0 +1,9 @@
+Bugfix: Initialize the user before the transfer-ownership command if the user has never logged in
+
+Trying to transfer the ownership of files to a user who hadn't logged in was causing problems
+because the FS of such user wasn't initialized and it wasn't possible to move the files there.
+The command appeared to work, but the files weren't moved.
+
+Now such user has the FS initialized to the transfer can be completed normally.
```
phil-davis:
```suggestion
Now such user has the FS initialized so the transfer can be completed normally.
```

micbar:
@jvillafanez Please update the title and the top post with the new approach.

Rest looks fine IMO

jvillafanez:
Done

micbar (approved):

PR #37041, opened 2020-02-27
[Tests-Only] Unset skeletondirectory after UtilTest exception

phil-davis:
## Description
See https://github.com/owncloud/core/pull/37038#issuecomment-592005411

`tests/lib/UtilTest.php` has some tests that purposely create a skeleton directory with no read access. They expect an exception when trying to create a user with skeleton. At the end of the test they (try) to delete the skeleton directory setting (cleanup the mess they purposely made).

But the expected exception is thrown before that point, so the "cleanup" code never runs. This leaves a bad skeleton directory setting that might cause later tests to fail.

Move the cleanup code into the `tearDown()` method so that it gets run after the expected exception has happened.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [x] Unit tests added (changed)
- [ ] Acceptance tests added

### DIFF
--- a/tests/lib/UtilTest.php
+++ b/tests/lib/UtilTest.php
@@ -16,6 +16,8 @@ use OCP\App\IAppManager;
  * @group DB
  */
 class UtilTest extends \Test\TestCase {
+	public $skeletonDirectoryWasSet = false;
+
 	public function testGetVersion() {
 		$version = \OCP\Util::getVersion();
 		$this->assertIsArray($version);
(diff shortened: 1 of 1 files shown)


phil-davis:
How many more of these are there? "someone" could inspect every unit test that hase `expectException` and make sure that it does not have any "cleanup" code after the point where the exception should happen.

jvillafanez (approved):
Maybe it's better if we get and store the current skeleton value in the setup method and restore or delete it (if it wasn't setup) in the tear down.
Another option could be to always delete the system value regardless of being set or not (not optimal, but more expected)

In any case, taking into account the tests, this is fine, so up to you.

phil-davis:
Yes, it would be even better if we get the current value first (if set), remember it, and then put it back or delete as appropriate. But that is refactoring for another day :)

PR #37043, opened 2020-02-27
Allow all IAuthModules to check the auth headers

### LINKED ISSUE #37059: Allow oauth2 and openidconnect to coexist
When enabling the oauth2 app and the openidconnect app at the same time the openidconnect app does not get a chance to authenticate the bearer token header when authenticating oidc auth tokens, because the oauth2 app will throw an exception while the [Session](https://github.com/owncloud/core/blob/18831fa81167a9a876d0bb0d06c5106460714743/lib/private/User/Session.php#L1152-L1179) checks the auth headers.

The IAuthModules ([TokenAuthModule](https://github.com/owncloud/core/blob/2de709ee929ff8ffd480019c82e09929134ad41d/lib/private/User/TokenAuthModule.php#L64), [OAuth2\AuthModule](https://github.com/owncloud/oauth2/blob/master/lib/AuthModule.php#L47), [OpenIdConnectAuthModule](https://github.com/owncloud/openidconnect/blob/master/lib/OpenIdConnectAuthModule.php#L84) and [BasicAuthModule](https://github.com/owncloud/core/blob/2de709ee929ff8ffd480019c82e09929134ad41d/lib/private/User/BasicAuthModule.php#L124)) are built in a way to return null when they cannot authenticate a user. But they also throw exceptions, when something goes wrong.

An exception from the OAuth2 AuthModuleThis prevents the OpenIDConnectAuth Module the verify the Bearer token.

Also looking at the code flow it seems the Session relies on the BasicAuthModule to throw an exception in case it could not verify a user ... like a catch all. Because if no backend throws an exception the user will not be logged out... Which seems to be wrong?

Shouldnt the Session ignore all exceptions as long as one backend can authenticate the user (the Token / cookie / php session being one of them)? Or at least leave the user logged in as long as all auth modules agree on which user is logged in ...

butonic:
## Description
While the UserSession in ownCloud 10 is currently used to test all available IAuthModule implementations, it immediately logs out the user when an exception occurs. However, existing owncloud 10 instances use the oauth2 app to create Bearer tokens for mobile and desktop clients.

This PR allows the OAuth2 app and the openidconnect app to coexist.

## Issue
Fixes https://github.com/owncloud/core/issues/37059

## Motivation and Context
Trying to make oauth2 and openid connect work in parallel for a smooth transition from owcloud 10 to ocis.

## How Has This Been Tested?
locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -1151,26 +1151,30 @@ class Session implements IUserSession, Emitter {
 
 	public function verifyAuthHeaders($request) {
 		$shallLogout = false;
-		try {
-			$lastUser = null;
-			foreach ($this->getAuthModules(true) as $module) {
+		$lastUser = null;
+		foreach ($this->getAuthModules(true) as $module) {
+			try {
 				$user = $module->auth($request);
-				if ($user !== null) {
-					if ($this->isLoggedIn() && $this->getUser()->getUID() !== $user->getUID()) {
-						$shallLogout = true;
-						break;
-					}
-					if ($lastUser !== null && $user->getUID() !== $lastUser->getUID()) {
-						$shallLogout = true;
-						break;
-					}
-					$lastUser = $user;
-				}
+			} catch (Exception $ex) {
+				// convert to no user so all IAuthModule implementations can check
+				// TODO log exception
+				$user = null;
 			}
-		} catch (Exception $ex) {
-			$shallLogout = true;
+			if ($user !== null) {
+				// check if the logged in user is the same as the one found by the auth module
+				if ($this->isLoggedIn() && $this->getUser()->getUID() !== $user->getUID()) {
+					$shallLogout = true;
+					break;
+				}
+				// check if the auth modules all agree on the same user
+				if ($lastUser !== null && $user->getUID() !== $lastUser->getUID()) {
+					$shallLogout = true;
+					break;
+				}
+				$lastUser = $user;
+			}
 		}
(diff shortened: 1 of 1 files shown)


mmattel:
Note, at the moment we have in docs (https://github.com/owncloud/docs/pull/2110), but only as internal remark the statement `` //== OAuth2 and OIDC are mutually exclusive``. If this PR gets merged and be part of 10.4, we should document the possible coexistance.

@DeepDiver1975 @settermjd fyi

DeepDiver1975:
As the failing unit tests show this heavily changes the behavior of the system.

Implemented design work s as follows:
All auth headers are verified and as soon as one is invalid the whole request is rejected.
This is necessary in a scenario where e.g. an invalid bearer token is sent together with a valid session cookie. This one invalid bearer token is enough to reject the request.

This is why modules return a user or null or throw an exception.
user -> valid auth header
null -> unknown auth header
exception -> invalid auth header

In order to allow multiple modules to handle the same auth schema we would need to bundle modules by auth schema and check that at least one of them can work with the header and declare it as valid.

To be honest I consider this a critical change which I'd advise not to take into consideration.

butonic:
The only conflicting apps are oauth2 and openidconnect.

Closing in favor of a non core change in the oauth app: https://github.com/owncloud/oauth2/pull/255

PR #37053, opened 2020-02-29
fix remaining time estimation for public link uploads

### LINKED ISSUE #25053: Estimate properly or not show ETA for uploaded files using the web UI.
### Steps to reproduce:
1. Upload a file with a size of several megabytes.
### Expected behaviour:

Time left written in the progress bar decreases per second.
### Actual behaviour

Time left is variable, it can increase a lot, decrease, increase again.

I recommend to remove this text from progress bar or improve the estimation.

cc: @luckydonald

karakayasemi:
## Description
Currently, remaining time estimation is relies on `new Date().getMilliseconds();` difference between the two `fileupload.fileuploadprogressall` events. When time second change there is no meaning in this milliseconds difference. We are using Jquery-file-upload package in file-upload and the package is already providing bitrate for upload. With this PR the code will use this bitrate. 

In addition, to provide a smooth experience, the code is using a buffer mechanism. But, in the first bufferSize(20) events, since the buffer is empty, time estimation is wrong. The remaining time is rapidly increasing at the beginning. This problem resolved by dividing bufferTotal to the filled buffer size. Also, this buffer logic explained by adding a comment line.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/25053
- Fixes https://github.com/owncloud/enterprise/issues/3827

## Motivation and Context
Fixing bugs.

## How Has This Been Tested?
Test 1:
- Upload a 5 GB file via a public link
- Look at the estimated upload time (Previously it was wrong on Chrome)

Test 2:
- Upload a 5 GB file as user
- Look at the estimated upload time

I repeated tests with IE, Opera, Chrome and Firefox browsers. All of them look okay.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -1269,12 +1269,11 @@ OC.Uploader.prototype = _.extend({
 
 			if (this._supportAjaxUploadWithProgress()) {
 				//remaining time
-				var lastUpdate = new Date().getMilliseconds();
-				var lastSize = 0;
 				var bufferSize = 20;
 				var buffer = [];
 				var bufferIndex = 0;
 				var bufferTotal = 0;
+				var filledBufferSize = 0;
 				for(var i = 0; i < bufferSize;i++){
 					buffer[i] = 0;
 				}
@@ -1308,19 +1307,20 @@ OC.Uploader.prototype = _.extend({
 				fileupload.on('fileuploadprogressall', function(e, data) {
 					self.log('progress handle fileuploadprogressall', e, data);
 					var progress = (data.loaded / data.total) * 100;
-					var thisUpdate = new Date().getMilliseconds();
-					var diffUpdate = (thisUpdate - lastUpdate)/1000; // eg. 2s
-					lastUpdate = thisUpdate;
-					var diffSize = data.loaded - lastSize;
-					lastSize = data.loaded;
-					diffSize = diffSize / diffUpdate; // apply timing factor, eg. 1mb/2s = 0.5mb/s
-					var remainingSeconds = ((data.total - data.loaded) / diffSize);
+					var remainingBits = (data.total - data.loaded) * 8;
+					var remainingSeconds = remainingBits / data.bitrate;
+
+					//Take the average remaining seconds of the last bufferSize events
+					//to prevent fluctuation and provide a smooth experience
 					if (isFinite(remainingSeconds) && remainingSeconds >= 0) {
 						bufferTotal = bufferTotal - (buffer[bufferIndex]) + remainingSeconds;
-						buffer[bufferIndex] = remainingSeconds; //buffer to make it smoother
+						buffer[bufferIndex] = remainingSeconds;
 						bufferIndex = (bufferIndex + 1) % bufferSize;
+						if (filledBufferSize < bufferSize) {
+							filledBufferSize++;
+						}
 					}
-					var smoothRemainingSeconds = (bufferTotal / bufferSize); //seconds
+					var smoothRemainingSeconds = (bufferTotal / filledBufferSize);
 					var h = moment.duration(smoothRemainingSeconds, "seconds").humanize();
 					self.$uploadprogressbar.attr('data-loaded', data.loaded);
 					self.$uploadprogressbar.attr('data-total', data.total);
--- a/changelog/unreleased/37053
+++ b/changelog/unreleased/37053
@@ -1,0 +1,7 @@
+Bugfix: Fix public link upload remaining time estimation
+
+Public link upload wrong remaining time estimation problem has been resolved.
+Also, the remaining time calculation logic has been changed for smoother performance.
+
+https://github.com/owncloud/core/pull/37053
+https://github.com/owncloud/core/issues/25053


apps/files/js/file-upload.js
```diff
@@ -1308,19 +1307,19 @@ OC.Uploader.prototype = _.extend({
...
 					self.log('progress handle fileuploadprogressall', e, data);
 					var progress = (data.loaded / data.total) * 100;
-					var thisUpdate = new Date().getMilliseconds();
-					var diffUpdate = (thisUpdate - lastUpdate)/1000; // eg. 2s
-					lastUpdate = thisUpdate;
-					var diffSize = data.loaded - lastSize;
-					lastSize = data.loaded;
-					diffSize = diffSize / diffUpdate; // apply timing factor, eg. 1mb/2s = 0.5mb/s
-					var remainingSeconds = ((data.total - data.loaded) / diffSize);
+					var remainingSeconds = ((data.total - data.loaded) * 8) / data.bitrate;
```
jvillafanez: That `* 8` might need a quick comment. Maybe using a "remainingBits" (not bytes) variable could help to explain that `* 8`

jvillafanez:
Couple of questions:

> In addition, to provide a smooth experience, the code is using a buffer mechanism.

Do we need this? I think this "sugarcoating" not only it isn't accurate but also makes the code more complex.
In addition, it might be easier to understand to use an average of the previous "known" bitrates in order to calculate the remaining time.

The other question is if we're handling a possible "0" bitrate, mainly to not crash something.

apps/files/js/file-upload.js
karakayasemi: remainingBits now has own variable.

karakayasemi:
> 
> Couple of questions:
> 
> > In addition, to provide a smooth experience, the code is using a buffer mechanism.
> 
> Do we need this? I think this "sugarcoating" not only it isn't accurate but also makes the code more complex.
> In addition, it might be easier to understand to use an average of the previous "known" bitrates in order to calculate the remaining time.

Yes, I agree on the complexity part. However, now I tried calculating with average bitrates, I even implemented with different weighted averages to reduce the effect of a single bitrate, the buffer implementation seems slightly more performant. In addition, I found the commit of this buffer algorithm with git blame. The committer claims that he followed the same logic with the desktop client. IMHO, it can stay.

> 
> The other question is if we're handling a possible "0" bitrate, mainly to not crash something.

Yes, we have controls for 0 or negative bitrate case. The code ignores these bitrates.

jvillafanez (approved):

karakayasemi:
@cdamken, @voroyam If you confirm the issue resolved, we can merge this one. Thanks.

phil-davis:
When I have a local ownCloud and upload a ~2GB ISO file locally it actually takes about 20 seconds.

Before this change the progress bar would say "a few seconds" and then after a few seconds it would change to saying "2 minutes" "3 minutes"... and then near the end go back to estimating "a few seconds".

After this change the progress bar says "a few seconds"  all the time (and on some runs might say "1 minute" early in the upload if the upload is going a bit slow).

On a 10G file it says "2 minutes" pretty consistently, sometimes estimating 3 or 4 minutes (I think the file in system in the background slows down every now and then as it writes out the uploading file). Without the change, it seems to always say "1 minute" or "a few seconds" - i.e. it was too optimistic.

For me. this change is "a good thing".

It does need to be tried in an environment with "normal" internet speeds and a file size that will take 10 minutes or more to upload, to get a real idea of how "steady" the time estimate is.

karakayasemi:
>
> It does need to be tried in an environment with "normal" internet speeds and a file size that will take 10 minutes or more to upload, to get a real idea of how "steady" the time estimate is.

I made some tests by limiting the internet speed of my virtual machine. As far as I see, all test results are more stable than before.

jvillafanez:
@karakayasemi check if https://github.com/owncloud/core/pull/36814/files#diff-87d6d7ffd7622e59f65a6e6be8de6a56R53-R59 could be useful here. It's mainly to adjust the thresholds in order to show more accurate info instead of "a few seconds"

karakayasemi:
> 
> 
> @karakayasemi check if https://github.com/owncloud/core/pull/36814/files#diff-87d6d7ffd7622e59f65a6e6be8de6a56R53-R59 could be useful here. It's mainly to adjust the thresholds in order to show more accurate info instead of "a few seconds"

@jvillafanez thank you for the suggestion. I guess increasing the precision needs a pm decision. Let's merge it as it is by only fixing its errors.

davitol:
@karakayasemi Do we refer in this ticket to this time estimation?:

uploadBar


If so, it looks like estimation moves randomly

karakayasemi:
@davitol Can you compare the new code performance with the old one? Please, make sure 

If the first several calculated bit rates are slower in your instance it can led this behavior. We are also using an smoothing algorithm in here that uses last 20 consecutive packages average for calculating remaining time. The first 20 packages should be calculated quickly, but any fluctuated package can affect the average significantly at the beginning of upload.

micbar:
@karakayasemi Can you clarify if we need another fix? If yes, we would need it on the `release-10.4.1` branch

karakayasemi:
I do not think, we need a new fix. Since bitrate fluctation is very common problem, the time estimation is very tricky. However, the new implementation has better performance in my tests. Also @phil-davis stated similar comment on the above https://github.com/owncloud/core/pull/37053#issuecomment-594474566

phil-davis:
When I tested this in a local git clone and uploaded a few GB just on my `localhost` I found that the estimation was more stable than previously. There is no perfect solution for this - software cannot predict the future (e.g. if the network suddenly slows to a trickle then obviously the existing estimate is now wrong). IMO it is an improvement. Nothing is broken. If someone wants to improve it more, then great - propose a PR to master... But the code in 10.4.1 is fine to release.

jvillafanez:
I'm more inclined to provide a percentage based on the mount of data transferred because it should be more foreseeable and more accurate, but that's a different thing to be considered in the future.
The main problem with these kind of estimations is that we don't know, from a user's perspective, if that's a bug in the code, or a bitrate fluctuation, or a different thing. We could provide a random amount of time and blame the bitrate fluctuation.

I'm not saying to change anything (it's out of scope anyway), but I think it's something to take into account for future developments.

davitol:
> @davitol Can you compare the new code performance with the old one?

Yes, i already did when i tested it and I did not see relevant differences. But if you made some tests as you said and found it more stable, and also went fine for @phil-davis I'm ok to keep it like it is now for this version and let's see if @jvillafanez will be worthy for the future. Thank you all for the feedback.

Issue #35660, opened 2019-06-25
Cannot upload a file in a webDAV External Storage


davitol:
### Steps to reproduce

1)  As admin enable external storage

2) Create a mount point with webDAV option 

3) Check the gauge is green and the files already hosted are available

4) Try to upload a file into the eternal storage

### Expected behavior

The file is uploaded

### Actual behavior

It is not possible to upload a file. A log is spotted in ownCloud  log.

```
[...]

PR #37062, opened 2020-03-03
allow storages to explicitly opt-out of uploading .part files

### LINKED ISSUE #25826: Make it possible for storage implementations to disable part files
A storage implementation must be able to decide whether parts files are needed or not `needsPartFile`.

Some storages like GDrive and Dropbox do not really need it because uploading a file there doesn't result in partial file content. The atomicity is guaranteed by the provider.
Unlike SMB or FTP.

See https://central.owncloud.org/t/issue-with-versioning-on-google-drive/1683/9 for context

curiousercreative:
## Description
Allow for storage classes to decide whether to use .part files. Default to using part files to preserve previous behavior.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/25826

## Motivation and Context

Google Drive external storage (and others?) should not create a duplicate file with the same name every time a file is changed.

## How Has This Been Tested?

- no serious testing, just UAT that ownCloud is no longer generating duplicate Google Drive files

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


lib/private/Files/Storage/Wrapper/Wrapper.php
```diff
@@ -51,6 +51,13 @@ public function __construct($parameters) {
 		$this->storage = $parameters['storage'];
 	}
 
+	/**
+	 * @return boolean
+	 */
+	public function needsPartFile() {
+		return true;
```
DeepDiver1975: should call the wrapped storage - right?

lib/public/Files/Storage/IStorage.php
```diff
@@ -46,6 +46,12 @@
  */
 interface IStorage {
 
+	/**
+	 * needsPartFile - allows a storage to opt-out of .part upload and rename
+	 * @return boolean
+	 */
```
DeepDiver1975: needs a since annotation - due to api change 10.5 might be reasonable

DeepDiver1975 (dismissed):
Generally speaking this looks good - THX
I left a few minor comments

curiousercreative:
@DeepDiver1975 prior revision requests should be resolved and this is ready for another review

curiousercreative:
@DeepDiver1975 @jvillafanez bump for this PR. Would be great to either:
1. get this merged before more conflicting/redundant work is done
1. discuss a more sustainable solution and assign the work

jvillafanez:
Changes look good to me. Maybe @micbar can move this forward.

mmattel:
Adding @micbar and @phil-davis to reviewers.
Would solve a long lasting problem for gdrive.
Would be great to get into 10.5

phil-davis:
@micbar I am not really qualified to approve this. The code looks reasonable to me. Can you decide what happens next?

jvillafanez:
Technically the problem is solved with https://github.com/owncloud/core/pull/37265/commits/b14e8e46bfd6aef7ace5cb2f3821275251665111 , but this is a better option and we should use something like this.

mmattel:
What about reverting b14e8e4 and use this PR instead?
As you stated, it is a better solution...

jvillafanez:
Merging this PR will overwrite that piece, so there is no need to revert anything

karakayasemi:
The PR can also help to resolve https://github.com/owncloud/core/issues/35660 . Disabling part file usage in webdav external storage should be enough with this pr.

micbar:
@curiousercreative This lookg good to me.

I would like to have more technical info in the changelog item, as this not only affects gdrive but external storages in general

micbar:
Is it possible to cover that by a unit test?

curiousercreative:
I've not written any unit tests for PHP code let alone whatever test runner and/or framework ownCloud is using (PHP Unit?). I did some digging and see two potential directions for adding the test:

1. [apps/dav/tests/unit/Connector/Sabre/FileTest.php](https://github.com/owncloud/core/blob/master/apps/dav/tests/unit/Connector/Sabre/FileTest.php)
1. or a new test that, similar to [this](https://github.com/owncloud/core/blob/master/apps/dav/tests/unit/Connector/Sabre/RequestTest/PartFileInRootUploadTest.php) extends [apps/dav/tests/unit/Connector/Sabre/RequestTest/UploadTest.php](https://github.com/owncloud/core/blob/master/apps/dav/tests/unit/Connector/Sabre/RequestTest/UploadTest.php)

If adding to the first file, I was thinking to add a test function named `testNeedsPartFileMatchesStorage`. Are we wanting to test anything other than our connector relaying the underlying storages preference for `needsPartFile`?  More ambitious would be something closer to a functional test, likely the second idea above and try to do the following (but I'm not familiar enough with this codebase, PHP, etc):
1. create an arbitrary Storage backend (akin to the Google Drive external Storage)
1. override `needsPartFile` method and have that return `true`
1. start a file upload
1. assert that file exists during upload with a part file name
1. duplicate test for `needsPartFile` returning `false` and assert that file exists during upload with part file name

For further discussion, what is the history of [this system setting](https://github.com/owncloud/core/blob/master/config/config.sample.php#L1389)? Is it in any way affected by this pull request (obsoleted perhaps)?

curiousercreative:
@micbar updated changelog. LMK if you're still missing any particular detail. See above comment re: unit tests.

mmattel:
@phil-davis can you restart drone, stuck on connecting to github...

mmattel:
Any update?
Squashing commits would restart the checking process, then we see also the result of codecov/patch.

mmattel:
@karakayasemi could you support for tests?

karakayasemi:
@curiousercreative Thank you for contribution! I cherry-picked your commit and completed rest of the things in https://github.com/owncloud/core/pull/37755. Now, it is merged. Closing this one.

PR #37067, opened 2020-03-03
Check the file in the primary storage and allow remove it from the

jvillafanez:
cache if it's missing

## Description
Files in the primary storage could be deleted outside of ownCloud, leaving information in the ownCloud's file cache. This command intends to check if the target file can be read from the backend storage, and it not, it allows you to remove the information cached.

Main goal is to clean the cache for objectstores (objectstore and files_primary_S3 apps) as primary storage, but the behaviour has been homogenized, so it can be used for local storage as long as the same conditions apply.

Note that removing files directly from the primary storage isn't supported and shouldn't happen. As such, the cases where you need to run this command should be extremely weird. This is why we only provide a command to check for only one file instead of scanning for the whole ownCloud's FS

## Related Issue
https://github.com/owncloud/enterprise/issues/3801

## Motivation and Context
Provide a tool to clean up stored information in the ownCloud's filecache in case a file disappears from the primary storage

## How Has This Been Tested?
Check both with local storage and objectstore app (files_primary_S3 not tested but expected to behave the same as with objectstore)
1. Upload some files into the primary storage through ownCloud
2. Remove one of those files directly from the storage (bypassing ownCloud)
3. Run the command

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/lib/Command/CheckCache.php
+++ b/apps/files/lib/Command/CheckCache.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * @author Juan Pablo Villafañez Ramos <jvillafanez@solidgeargroup.com>
+ * @author Juan Pablo Villafáñez Ramos <jvillafanez@solidgeargroup.com>
  *
  * @copyright Copyright (c) 2020, ownCloud GmbH
  * @license AGPL-3.0
--- a/apps/files/lib/Command/CheckCache.php
+++ b/apps/files/lib/Command/CheckCache.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * @author Juan Pablo Villafañez Ramos <jvillafanez@solidgeargroup.com>
+ *
+ * @copyright Copyright (c) 2020, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


apps/files/lib/Command/CheckCache.php
```diff
@@ -0,0 +1,103 @@
...
+			$output->writeln("<comment>Ignoring $targetFile because it is shared or not inside the primary storage</comment>");
+			return 0;
+		}
+
+		if (!($node instanceof File)) {
+			$output->writeln("<error>$targetFile isn't a file</error>");
+			return self::ERROR_NOT_A_FILE;
+		}
+
+		$stream = @$node->fopen('rb');  // errors are expected if the file is missing
```
jvillafanez: Need to check if we want to suppress the errors or not. I assume the admin knows what file is giving problems as well as the reason behind those problems, so there is no reason to log the same thing again.
Note that, taking into account the current implementation of the objectstore, we can't know the reason why the file can't be opened. We're assuming here that the file is missing, but it might happen that there is no connection or wrong permissions; in this case, we rely on the admin to use or not this command knowing that the file (and/or the filecache infor) might be deleted.

micbar:
@jvillafanez LGTM from a code POV

@cdamken @pako81 @enbrnz Could support / consulting give some feedback on this feature please?

micbar (approved):

mmattel:
You "forgot" to create a doc issue 👎 
These things could be easily missed and forgotten...
I added one for you, see link.

PR #37069, opened 2020-03-04
[Tests-Only] Add acceptance test for translated response messages

jasson99:
## Description
This PR adds acceptance test for the translated error messages when the Accept-Language header is set to some preferred language.

## Related Issue
# https://github.com/owncloud/QA/issues/608

## How Has This Been Tested?
CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiTranslation/translation.feature
+++ b/tests/acceptance/features/apiTranslation/translation.feature
@@ -1,0 +1,30 @@
+@api
+Feature: translate messages in api response to preferred language
+  As a user
+  I want response messages to be translated in preferred language
+  So that I can see and understand the response messages in my language
+
+  Scenario Outline: user tries to get non existing share and uses some preferred language
+    Given user "user0" has been created with default attributes and without skeleton files
+    And these users have been created with default attributes and skeleton files:
+      | username |
+      | user1    |
+      | user2    |
+    And using <dav_version> DAV path
+    And user "user1" has shared file "textfile0.txt" with user "user2"
+    When user "user0" gets the info of the last share in language "de-DE" using the sharing API
+    Then the OCS status code should be "404"
+    And the OCS status message should be "Fehlerhafte Freigabe-ID, Freigabe existiert nicht"
+    When user "user0" gets the info of the last share in language "zh-CN" using the sharing API
+    Then the OCS status code should be "404"
+    And the OCS status message should be "错误的共享 ID，共享不存在"
+    When user "user0" gets the info of the last share in language "fr-FR" using the sharing API
+    Then the OCS status code should be "404"
+    And the OCS status message should be "Mauvais ID de partage, le partage n'existe pas"
+    When user "user0" gets the info of the last share in language "es-ES" using the sharing API
+    Then the OCS status code should be "404"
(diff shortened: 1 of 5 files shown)


tests/acceptance/features/apiTranslation/translation.feature
```diff
@@ -0,0 +1,22 @@
...
+
+  Scenario Outline: user tries to get non existing share and uses language "deutsch"
+    Given user "user0" has been created with default attributes and without skeleton files
+    And these users have been created with default attributes and skeleton files:
+      | username |
+      | user1    |
+      | user2    |
+    And using <dav_version> DAV path
+    And user "user1" has shared file "textfile0.txt" with user "user2"
+    When user "user0" gets the info of the last share in language "deutsch" using the sharing API
```
phil-davis:
```suggestion
    When user "user0" gets the info of the last share in language "de-DE" using the sharing API
```
I am surprised that this works by sending "deutsch". I thought that you have to send a recognised language code.

https://en.wikipedia.org/wiki/Language_localisation has some examples of the language codes.

tests/acceptance/features/apiTranslation/translation.feature
```diff
@@ -0,0 +1,22 @@
+@api
+Feature: translate messages in api response to preferred language
+  As a user
+  I want response messages to be translated in preferred language
+  So that I can see and understand the response messages in my language
+
+  Scenario Outline: user tries to get non existing share and uses language "deutsch"
```
dpakach: check with other languages as well. At least couple to be sure.

dpakach (dismissed):
LGTM :+1:

tests/acceptance/features/apiTranslation/translation.feature
jasson99: fixed

tests/acceptance/features/apiTranslation/translation.feature
jasson99: fixed

phil-davis:
@dpakach please review again. This looks OK - I guess the risk is the a translator of one of these languages decides to change the translation, and then CI will suddenly break. But it will be easy to fix if it happens.

Then the commits can be squashed before merging.

tests/acceptance/features/apiTranslation/translation.feature
dpakach: Use Scenario outline rather than repeating it every time.

dpakach (commented):
Use outline to test for different languages rather than repeating it every time.

tests/acceptance/features/apiTranslation/translation.feature
jasson99: fixed

dpakach (commented):
LGTM :+1:

phil-davis (approved):
For this sort of scenario it is OK to have repeated When/Then - otherwise it takes so much time if we have 8 rows in the examples table...

phil-davis:
drone had timeouts pulling some docker images. I restarted it.

PR #37103, opened 2020-03-10
Add force option to delete user home folder

sharidas:
Add a force option to the user delete command
in order to delete the users home folder if
exists.

## Description

Add force option for the `user:delete` command, to delete the home folder of the user if available.

## Related Issue

https://github.com/owncloud/enterprise/issues/2969

## Motivation and Context
Its been noticed that when users upgrade from 9.1.5 to 10 of core, the accounts table does not have the user details and neither the home entry added to it. Now if the backend had deleted a user before upgrade then oC admin will not be able to delete the oc user mapped to the ldap backend. Because the entry is missing in the accounts table. Under such condition it would be handy to delete the home folder of those users (i.e, for missing backend users).

## How Has This Been Tested?

- Install oC 9.1.5
- Create LDAP users
- Now enable user_ldap app
- Configure it with the oC as admin
- Make sure the LDAP users are synced. You can see the result in the Users page.
- Now try to login to few users who are mapped with the LDAP.
- Delete the users in the LDAP.
- Upgrade to oC 10.4.0
- Now try to run `user:delete` with uid and --force, where uid is the user unavailable in the LDAP or say the deleted one.
- The home folder ( i.e, the remains ) should be deleted.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/lib/User/ManagerTest.php
+++ b/tests/lib/User/ManagerTest.php
@@ -14,6 +14,7 @@ use OC\User\Backend;
 use OC\User\Database;
 use OC\User\Manager;
 use OC\User\SyncService;
+use OC\User\DeletedUser;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\IConfig;
@@ -149,6 +150,29 @@ class ManagerTest extends TestCase {
 	public function testGetAccountNotExists() {
 		$this->accountMapper->expects($this->once())->method('getByUid')->with('foo')->willThrowException(new DoesNotExistException(''));
 		$this->assertNull($this->manager->get('foo'));
+	}
+
+	public function testGetAccountNotExistsButForceReturn() {
+		$this->accountMapper->expects($this->once())
+			->method('getByUid')
+			->with('foo')
+			->willThrowException(new DoesNotExistException(''));
+
+		$user = $this->manager->get('foo', true);
+		$this->assertInstanceOf(DeletedUser::class, $user);
+		$this->assertEquals('foo', $user->getUID());
+	}
+
+	public function testGetAccountNotExistsButForceReturnAfterCache() {
+		$this->accountMapper->expects($this->exactly(2))  // FIXME: second call shouldn't hit the mapper
+			->method('getByUid')
+			->with('foo')
+			->willThrowException(new DoesNotExistException(''));
+		$this->assertNull($this->manager->get('foo'));
+
+		$user = $this->manager->get('foo', true);
+		$this->assertInstanceOf(DeletedUser::class, $user);
+		$this->assertEquals('foo', $user->getUID());
 	}
 
 	public function testGetDuplicateAccountNotExists() {
--- a/apps/dav/lib/CardDAV/SyncService.php
+++ b/apps/dav/lib/CardDAV/SyncService.php
@@ -247,6 +247,37 @@ class SyncService {
 	}
 
 	/**
+	 * Delete the card owned by the user if the userId is unique
+	 * Returns:
+	 *   - true if the card is deleted (the user was unique)
+	 *   - false if there are 2 or more cards found with the same uid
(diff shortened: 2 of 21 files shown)


jvillafanez:
I don't think the approach is good.
* The `--force` option is meant to bypass the safety checks there could be to prevent the command from running. Removing the home folder is a consequence of removing the user, but not the goal of the `--force` option
* The usual approach is more streamlined: you have actions A, B and C, and B (or A) could fail, preventing C from running. In this case, the `--force` option would need to ensure that A, B and C actions runs always, even though it could be unsafe

From our case, there is only one action, which is the `$user->delete()`, with a safety check, which is that the user must exists.
The `user->delete()` should have several actions which should involve deleting the home directory, deleting the shares, deleting user preferences, etc, These actions should be replicated in order to ensure the user is properly deleted, as if the `$user->delete()` has been called. If there is some event being triggered, the same event should be triggered in this case.

micbar:
@jvillafanez give it a new try with a fake user

jvillafanez:
Apart from the unit tests, there are a couple of problems to be fixed:
* user ldap mappings won't be cleaned. We require a "postDelete" listener in the user_ldap app to do the job. Note that the app will likely need to remove that mapping regardless of where the uid comes from (we can't rely on any information about the user backend)
* cardDav sync table will still hold some user information. It tries to get the user object from the user manager, but the user is missing (https://github.com/owncloud/core/blob/master/apps/dav/lib/HookManager.php#L95). We might need to include a "delete user info regardless of the backend if it's unique" method in order to try to attempt removing that information

jvillafanez:
It seems we have a huge mess with the hooks.

Current solution:
`DeletedUser` requires the `User\Manager` implementation, which extends the `PublicEmitter` class for whatever reason. This allows us to trigger the same events the `User` implementation is triggering, plus we ensure the listeners are the expected ones (we aren't notifying to a different set of listeners by accident)
However, this breaks the tests because the mocked class is `IUserManager`, without any relation to the `User\Manager`


#### Alternative 1
Include a `getPublicEmitter()` function in the `IUserManager` interface. This would effectively allow us to require a `PublicEmitter` in the `DeletedUser` class. The `Delete` command would get that `PublicEmitter` from the interface and inject it in the `DeletedUser` class.
Ideally, there is no big change in the `User\Manager` implementation: the method could just return the same instance (assuming it implements the same interface)

The problem is that anyone could trigger events anywhere (which, in fact, is what we're doing). This is potentially dangerous because any app could trigger an "user deleted" event even though such event never happened,

#### Alternative 2
Change the `IUserManager::get(...)` interface to include an option to return a `DeletedUser` instance if the user is missing. The default behaviour won't change.


I'll try this second approach

tests/lib/User/ManagerTest.php
```diff
@@ -151,6 +152,29 @@ public function testGetAccountNotExists() {
...
+			->with('foo')
+			->willThrowException(new DoesNotExistException(''));
+
+		$user = $this->manager->get('foo', true);
+		$this->assertInstanceOf(DeletedUser::class, $user);
+		$this->assertEquals('foo', $user->getUID());
+	}
+
+	public function testGetAccountNotExistsButForceReturnAfterCache() {
+		$this->accountMapper->expects($this->exactly(2))  // FIXME: second call shouldn't hit the mapper
```
jvillafanez, line 167: We need to decide what to do with this. The `AccountMapper` should be called only once because the missing user should be cached (and stored as null) accordingly to what the `UserManager` does, but the cache implementation ignore these missing users when asked if they're in the cache (`hasKey` method returns false even if there is a key with a null value). This means that a second hit to the `AccountMapper` will happen as shown by this test.
Either there is a bug in the `CappedMemoryCache` class, or the `UserManager`'s behaviour is wrong. It isn't clear if the missing users should be cached or not by the `UserManager`.

micbar (approved):

Issue #37092, opened 2020-03-06
Missing DAV share types for reshares when viewed as share owner


PVince81:
### Steps

1. Setup Phoenix with the test share scenarios: https://github.com/owncloud/product-internal/blob/master/tools/create_all_possible_share_scenarios.py
1. Login as "Alice"
1. Enter the folder "alice2bob2charlie+link (IUP, OUP, OLP)
1. Examine the indicators
1. Open the share panel for each of the four entries successively

### Expected result
The two entries which have an direct outgoing share must have it shown in the indicator directly (blue instead of grey)

### Actual result
All indicators are grey until the share panel is open, then the relevant ones will become blue.

### Version
Phoenix master 0dec7af44bb3023c696908c43b91fdcb1c2ae2c7

Somehow blocks https://github.com/owncloud/phoenix/pull/3135 where an indicator test failed, so could be related

Related story: https://github.com/owncloud/product-internal/issues/36

PVince81:
Yep, also happening in OC 10, steps:

1. Setup OC 10.4.0 with the test share scenarios: https://github.com/owncloud/product-internal/blob/master/tools/create_all_possible_share_scenarios.py
1. Login as "Alice"
1. Enter the folder "alice2bob2charlie+link (IUP, OUP, OLP)
1. Examine the indicators
1. Open the share panel for each of the four entries successively and click "Sharing" tab

As soon as the Sharing tab is opened, the indicator in line now says "Shared".

It seems this is a very old undiscovered bug...

PVince81:
@micbar FYI

I don't think we can work this around easily in Phoenix without making even more expensive calls

jvillafanez:
https://github.com/owncloud/core/pull/37107 patch over OC master code should fix the problem. Checked with OC 10, but not with phoenix

jvillafanez:
PR merged, closing

PR #36865, opened 2020-02-01
show reshare public links to the original share owner

karakayasemi:
## Description
With this PR, reshare public links will be visible for the original share-owner. 
There will be no difference in the UI between owner-created public links and reshare public links. However, if we want to show reshares to the owner properly, we need a solution for all clients (desktop, mobile, web). The desktop client already shows this information, no need to hide them in here.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3757

## How Has This Been Tested?
- create a reshare public link
- it should be visible for original share owner

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37107, opened 2020-03-11
Include reshares in the webdav response when asking for share types

jvillafanez:
## Description
Consider reshares when asking for the share types via webdav

## Related Issue
https://github.com/owncloud/core/issues/37092

## How Has This Been Tested?
Tested with master code (35f7a3f45005230e414d9b4784f2776001a94da6). Requires https://github.com/owncloud/core/pull/36865 to prevent buggy behaviour in OC 10
Patch won't work in a 10.4.0 installation due to the above requirement, which was merged into master recently

Tested following the same steps as described in https://github.com/owncloud/core/issues/37092

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37107
+++ b/changelog/unreleased/37107
@@ -1,0 +1,6 @@
+Change: Include reshares in the webdav response when asking for share types
+
+Previously, only shares initiated by the user were being returned when asked for
+the shares. Now, reshares are also included in the response
+
+https://github.com/owncloud/core/pull/37107
--- a/apps/dav/lib/Connector/Sabre/SharesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/SharesPlugin.php
@@ -140,7 +140,8 @@ class SharesPlugin extends \Sabre\DAV\ServerPlugin {
 		$allShares = $this->shareManager->getAllSharesBy(
 			$this->userId,
 			$requestedShareTypes,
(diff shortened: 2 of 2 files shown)


PVince81:
I've just tested this PR with https://github.com/owncloud/core/issues/37092 and it works fine.

I suspect this will have some performance impact, not sure if linear ? Is the sharing code just doing an additional SQL query or is it more ?

jvillafanez:
The code ends up in https://github.com/owncloud/core/blob/master/lib/private/Share20/DefaultShareProvider.php#L503

There are 3-4 calls to that due to different share types, but not due to checking for reshares. Checking for the reshares is expected to have a linear impact for the amount of shares we need to send. It will also depend on how the DB will handle the query (use of indexes, mainly)

I don't expect the performance to degrade too much.

micbar (approved):

changelog/unreleased/37107
```diff
@@ -0,0 +1,6 @@
+Change: Include reshares in the webdav response when asking for share types
+
+Previously, only shares initiated by the user was being returned when asked for
+the shares. Now, reshares are also included in the response
+
```
micbar: 'were' plural

PR #37100, opened 2020-03-10
Drop support for PHP 7.1

phil-davis:
## Description
Drop support for PHP 7.1

This might be helpful before adding support for PHP 7.4 (it might mean less issues with dependencies...) People could work on top of this for supporting PHP 7.4

composer finds it can update the following PHP dependencies:
```
composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 4 updates, 0 removals
  - Updating doctrine/dbal (v2.9.3 => v2.10.1): Loading from cache
  - Updating doctrine/lexer (1.0.2 => 1.2.0): Loading from cache
  - Updating phpdocumentor/type-resolver (1.0.1 => 1.1.0): Loading from cache
  - Updating phpdocumentor/reflection-docblock (4.3.4 => 5.1.0): Loading from cache
```

`master` is now on-the-way to 10.5.0. If we are happy to drop PHP 7.1 support in 10.5.0 (and add PHP 7.4 support...) then we could merge this now.

## Related Issue
- Fixes #36510

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37109, opened 2020-03-11
[Tests-Only] [Do not merge] Run PHP unit tests with PHP 7.2

phil-davis:
This PR is to check if unit tests run OK with PHP 7.2
(to compare with #37100 )


phil-davis:
Works with PHP 7.2.
See PR #37119 for the real code to merge to master.

PR #37111, opened 2020-03-12
Trigger CI

phil-davis:
To check the results of webUI tests.

They fail with `Fatal error: Call to a member function open() on null (Behat\Testwork\Call\Exception\FatalThrowableError)` - see issue #37113 for details.


phil-davis:
See PR linked to the issue. This is fixed.

PR #37122, opened 2020-03-13
Handle case of ZIP Archive with no files

### LINKED ISSUE #37121: OC\Archive\ZIP getFiles() when archive has been emptied
### Steps to reproduce
1. have a unit test that starts with an empty Zip archive, adds a file to it then removes the file
2. call `getFiles()`

### Expected behaviour
`getFiles()` returns an empty array

### Actual behaviour
`getFiles()` returns:
```
array(1) {
  [0] =>
  bool(false)
}
```

### Server configuration
current core GitHub master

phil-davis:
## Description
The call directly to `$this->zip->numFiles` "internal"  variable returns `1` when a ZipArchive is actually empty. That is internals of the PHP `ZipArchive` class. When we loop through the `1` file, we get `false` returned by `$this->zip->getNameIndex(0)`, which is the documented return value when there is no file at the index. That is good. Adjust our code to check for that `false` return value.

Note: the PHP `ZipArchive` class has a `count()` method since PHP 7.2. That might be good to use, but we still support PHP 7.1, so not yet.

This code seems to only really get run from `lib/private/Installer.php`. So IMO there will not be real-life use of this in a way that ends up with an empty ZipArchive. So this fix is really just tidying up the class so it consistently returns what it claims.

## Related Issue
- Fixes #37121 

## How Has This Been Tested?
CI unit tests

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Archive/ZIP.php
+++ b/lib/private/Archive/ZIP.php
@@ -128,7 +128,10 @@ class ZIP extends Archive {
 		$fileCount=$this->zip->numFiles;
 		$files= [];
 		for ($i=0;$i<$fileCount;$i++) {
-			$files[]=$this->zip->getNameIndex($i);
+			$name = $this->zip->getNameIndex($i);
+			if ($name !== false) {
+				$files[] = $name;
+			}
 		}
 		return $files;
 	}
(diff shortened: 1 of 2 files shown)


phil-davis:
@micbar for "fixes" like this that are "theoretical" (I can't identify a real-life use case that has bad behaviour now and will be fixed), what should I do about a changelog?

VicDeo (approved):
Makes sense

PR #37135, opened 2020-03-17
Add share type to the verifyExpirationDate hook

VicDeo:
## Description
add type of the share to the verifyExpirationDate hook

## Related Issue
- Part of the fix for https://github.com/owncloud/password_policy/issues/287

## Motivation and Context
it is not possible to distinguish public links from other share types 

## How Has This Been Tested?
1. Set public link expiration policies in the password policy app according to https://github.com/owncloud/password_policy/issues/287

2. Create a user or group share

### expected
share is created

### actual
share creation is blocked due to public link policies

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37135
+++ b/changelog/unreleased/37135
@@ -1,0 +1,10 @@
+Bugfix: Add share type to the verifyExpirationDate hook
+
+The verifyExpirationDate hook notifies the password_policy app about proposed
+expiration dates of shares. The share type was not being passed in the hook.
+This meant that the password_policy app incorrectly processed user and group
+share expiration dates. See the linked issue for details.
+The problem has been corrected.
+
+https://github.com/owncloud/password_policy/issues/287
+https://github.com/owncloud/core/pull/37135
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -469,6 +469,7 @@ class Manager implements IManager {
 			'accepted' => &$accepted,
 			'message' => &$message,
 			'passwordSet' => $share->getPassword() !== null,
+			'shareType' => $share->getShareType(),
 		]);
 
 		if (!$accepted) {


phil-davis:
This looks like "a good thing" (tm). Can a changelog be added? Then we could merge it when CI passes.

In the morning I can make some acceptance tests in password_policy that demonstrate the bug and then that the fix works.

phil-davis:
@VicDeo I added a changelog.

What about unit test changes/additions?

For example, https://github.com/owncloud/core/blob/master/tests/lib/Share20/ManagerTest.php#L1130 could have:
```
$share->setShareType(\OCP\Share::SHARE_TYPE_LINK);
```
and the expects for listener can be:
```
		$hookListner->expects($this->once())->method('listener')->with($this->callback(function ($data) use ($expected) {
			return $data['expirationDate'] == $expected && $data['passwordSet'] === false && $data['shareType'] == \OCP\Share::SHARE_TYPE_LINK;
		}));
```

The existing tests (which are really for public links) could be adjusted so they explicitly setShareType.

And add some tests to check what happens to the hook when share type is `SHARE_TYPE_USER` and `SHARE_TYPE_GROUP` ?

Whatever you think.

For me it would be nice to get this PR merged ASAP. It makes drone CI testing from password_policy app much easier.

phil-davis:
Tested locally/manually along with https://github.com/owncloud/password_policy/pull/293 and works.

VicDeo:
@phil-davis this oneliner is not worthy an unit test for me. 
The coverage is already at 100% and removing this line will break acceptance tests anyway.

phil-davis (approved):
This has been discussed and tested - approving.

PR #37058, opened 2020-03-02
Fixes #36920 - custom properties are now properly set on REPORT with …

DeepDiver1975:
## Related Issue
- Fixes  #36920

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37145, opened 2020-03-20
[Tests-Only] Removed issue tag from fixed issue

kiranparajuli589:
## Description
After #37058, few cleanup is needed
Issue tag from the scenario is removed!


phil-davis:
@kiranparajuli589 we are not needing to clean these up any more. Actually the tags can be left behind, because for the future they give people a pointer to the issue in GitHub which was the reason for making the scenario. That can be useful.

@individual-it - is that what we discussed?

kiranparajuli589:
> @kiranparajuli589 we are not needing to clean these up any more. Actually the tags can be left behind, because for the future they give people a pointer to the issue in GitHub which was the reason for making the scenario. That can be useful.
> 
> @individual-it - is that what we discussed?

Oh OK! I agree with the concept, that is actually nice!

PR #37152, opened 2020-03-21
Add index on addressbookid

mrow4a:
Fixes owncloud/enterprise#3625. The work should add index for `'addressbookid' 'name' 'value'` that allows to improve scan performance of search addressbook query when e.g. end of term wildcard (term%) is used.

Please note combinations that allow to fully utilize index according to https://github.com/owncloud/core/pull/36225, and comment explaining why index cannot be utilized in these cases https://github.com/owncloud/enterprise/issues/3625#issuecomment-573861541:
- `accounts.enable_medial_search=true` - `%value%` - index ignored
- `accounts.enable_medial_search=false` - `value%` - index used

However, query with this index should be faster in general too regardles of wildcard on `value`, due to index on `'addressbookid', 'name'`, that will prefilter search

Targeting 10.5 (@micbar )

### DIFF
--- a/apps/dav/appinfo/Migrations/Version20200114181454.php
+++ b/apps/dav/appinfo/Migrations/Version20200114181454.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * @author Jörn Friedrich Dreyer <jfd@butonic.de>
+ *
+ * @copyright Copyright (c) 2020, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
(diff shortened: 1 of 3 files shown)


mrow4a:
added some unit tests just to be sure

jvillafanez (approved):
Code looks good

mrow4a:
@micbar shall we merge as part of 10.5?

phil-davis:
Note: this is (correctly) after the release-10.4.1 branch point - good.
"some time soon" we should update version.php to say 10.5.0 - but doing that while 10.4.1 branch is still in progress will cause a merge conflict for that branch. So it might be easiest to wait for now.

mrow4a:
@pmaier1 migration complexity -> we need to build index on `oc_cards_properties` for all rows in `'addressbookid', 'name', 'value'` . This can be quite long operation on tables with a lot of entries.

PR #37153, opened 2020-03-21
Respect sharing.federation.allowHttpFallback config option

mrow4a:
This MR in general help local testing for federated sharing :
- catches exception that allow fallback from https->http 
- add some more unit tests testing `tryHttpPostToShareEndpoint`

Reasons:
- without this MR, exception is shown to the user:
`cURL error 60: SSL certificate problem: self signed certificate`
- now, better exception is shown as request gets properly handled:
`server cannot be reached`
- now, with the config `sharing.federation.allowHttpFallback` set, allows to create federated share.

Do we need some documentation for this config (developer docs) ?
```
occ config:system:set sharing.federation.allowHttpFallback --value true --type boolean
```

### DIFF
--- a/apps/federatedfilesharing/lib/Notifications.php
+++ b/apps/federatedfilesharing/lib/Notifications.php
@@ -330,15 +330,15 @@ class Notifications {
 		$try = 0;
 
 		while ($result['success'] === false && $try < 2) {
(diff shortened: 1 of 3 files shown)


VicDeo (approved):
Looks good

PR #37163, opened 2020-03-25
[Tests-Only] Added method listTrashbinFileInServerRoot 

swoichha:
## Description
Added method `listTrashbinFileInServerRoot` and made some changes in method `theFileWithContentShouldExistInTheServerRoot` in FeatureContext.php

## Related Issue
- https://github.com/owncloud/data_exporter/pull/149

## How Has This Been Tested?
- locally 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -2146,11 +2146,10 @@ class FeatureContext extends BehatVariablesContext {
 	 *
 	 * @param string $path
 	 * @param string $content
-	 * @param bool $trash
 	 *
-	 * @return string
+	 * @return void
 	 */
-	public function theFileWithContentShouldExistInTheServerRoot($path, $content, $trash = false) {
+	public function theFileWithContentShouldExistInTheServerRoot($path, $content) {
 		$this->readFileInServerRoot($path);
 		Assert::assertSame(
 			200,
@@ -2160,10 +2159,6 @@ class FeatureContext extends BehatVariablesContext {
 		$fileContent = HttpRequestHelper::getResponseXml($this->getResponse());
 		$fileContent = (string) $fileContent->data->element->contentUrlEncoded;
 		$fileContent = \urldecode($fileContent);
-
-		if ($trash) {
-			return $fileContent;
-		}
 
 		Assert::assertSame(
 			$content,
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -2123,26 +2123,47 @@ public function readFileInServerRoot($path) {
...
 			200,
 			$this->getResponse()->getStatusCode(),
 			"Failed to read the file {$path}"
 		);
-
 		$fileContent = HttpRequestHelper::getResponseXml($this->getResponse());
 		$fileContent = (string) $fileContent->data->element->contentUrlEncoded;
 		$fileContent = \urldecode($fileContent);
 
+		if($trash){
```
phil-davis:
```suggestion
		if ($trash){
```
https://drone.owncloud.com/owncloud/core/24035/2/6
code style is failing

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -2123,26 +2123,47 @@ public function readFileInServerRoot($path) {
...
 			"Failed to read the file {$path}"
 		);
-
 		$fileContent = HttpRequestHelper::getResponseXml($this->getResponse());
 		$fileContent = (string) $fileContent->data->element->contentUrlEncoded;
 		$fileContent = \urldecode($fileContent);
 
+		if ($trash) {
+			return $fileContent;
+		}
```
phil-davis: why is this before `assertSame` ?
If this function is called with `$trash = true` then the assert should still pass???

phil-davis: And actually, it could always return `$fileContent` - I don't think that will cause any trouble for Behat/Gherkin when it calls the step.

tests/acceptance/features/bootstrap/FeatureContext.php
swoichha: It should be fixed by now

tests/acceptance/features/bootstrap/FeatureContext.php
swoichha: For `$trash = true` this function should not call assert as we check the content for file in trashbin using method `assertFilePhysicallyExistInLastExportWithContent` in `DataExporterContext.php`. So if-condition is before assert.

dpakach: I dont think this function needs any change. We directly give the path of the trashbin file so it should work the way it is originally.

swoichha: I'll change it to original and try again.

phil-davis (approved):

PR #37177, opened 2020-03-27
Version bump and changelog update

HanaGemela:


jnweiger (changes requested):
Nah, the new changlog folder is missing.

phil-davis:
Somehow the changelogs got deleted, not renamed into a new folder.
We need a well-known process that works :)

micbar:
@phil-davis check handbook.owncloud.com should be documented.

@HanaGemela If you move files in a git repo, you need to add the new locations
I always do `git add .`

phil-davis:
Found it https://handbook.owncloud.com/changelog_entries.html
I thought there was some scripting magic, but not so. Just a matter of creating `changelog/10.4.1_2020-03-30` and moving the files to it, and making sure that they are added to the commit.

PR #37179

PR #37197, opened 2020-04-01
Update 37135

HanaGemela:
To make changelog to display the core issue instead of password policy

### DIFF
--- a/changelog/10.4.1_2020-03-30/37135
+++ b/changelog/10.4.1_2020-03-30/37135
@@ -6,5 +6,5 @@ This meant that the password_policy app incorrectly processed user and group
 share expiration dates. See the linked issue for details.
 The problem has been corrected.
 
-https://github.com/owncloud/password_policy/issues/287
 https://github.com/owncloud/core/pull/37135
+https://github.com/owncloud/password_policy/issues/287


changelog/10.4.1_2020-03-30/37135
```diff
@@ -6,5 +6,5 @@ This meant that the password_policy app incorrectly processed user and group
 share expiration dates. See the linked issue for details.
 The problem has been corrected.
 
-https://github.com/owncloud/password_policy/issues/287
 https://github.com/owncloud/core/pull/37135
+https://github.com/owncloud/password_policy/issues/287
```
phil-davis, line 10: @micbar `CHANGLEOG.md`  at line 21 in `release-10.4.1` branch already has the `password_policy/issues/287` reference. How will that get regenerated from this `37135` changelog file?

jnweiger: The CHANGELOG.md gets generated from the changelog/10.4.1/* entries on the fly. We do that on master automatically, but on the release branch we cheat: The release script checks out the branch, refreshes CHANGELOG.md, builds the tars, but does not commit it back to git. That means, the CHANGELOG.md we see in the branch, is old and harmless. This trick avoids merge conflicts later.

phil-davis (approved):

jnweiger (approved):
lgtm

changelog/10.4.1_2020-03-30/37135
micbar: > @micbar `CHANGLEOG.md` at line 21 in `release-10.4.1` branch already has the `password_policy/issues/287` reference. How will that get regenerated from this `37135` changelog file?

The CHANGELOG.md is autogenerated. It will be gererated only on master branch and d inside the tarball during build process. The source of truth is always the `changelog` folder

You can anytime get a generated CHANGELOG.md with a docker command.

In OCIS, this step is part of the makefile.

We could implement that in core repo.

micbar: See https://github.com/owncloud/core/pull/37198

PR #37200, opened 2020-04-01
Add logging of general exception to error log in federated share polling

mrow4a:
It can happen that polling throws an exception that a bit ambigous as `Skipping external share with id "id" from remote
"https://test.test. Reason: ""`. For debugging in such cases it might be better to log such exceptions to better understand the cause.


mmattel:
Pls run: `make test-php-style-fix` to fix coding style error(s)

mrow4a:
included in https://github.com/owncloud/core/pull/37391

PR #37202, opened 2020-04-02
[Tests-Only] Use newadmin for invalid auth tests

phil-davis:
## Description
If we use the built-in admin account for tests that purposely send an invalid password to API endpoints then there is the risk that the built-in admin account might get locked out. If the built-in admin account gets locked out during a test run then later `AfterScenario` and `BeforeScenario` actions will fail just because a previous test scenario "messed up" the built-in admin account.

For example, this can happen when running core acceptance tests while the `brute_force_protection` app is enabled.

Change these sort of scenarios so that they create a new user that is in the `admin` group, and use that user. That user can get locked-out in a scenario, and it will not break the built-in admin account.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiAuth/cors.feature
+++ b/tests/acceptance/features/apiAuth/cors.feature
@@ -185,7 +185,9 @@ Feature: CORS headers
   Scenario Outline: CORS headers should be returned when invalid password is used (admin only endpoints)
     Given using OCS API version "<ocs_api_version>"
     And the administrator has added "https://aphno.badal" to the list of personal CORS domains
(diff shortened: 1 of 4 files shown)


phil-davis:
I noticed this problem when running acceptance tests against a customized themed installation that has `brute_force_protection` enabled.

individual-it (approved):

Issue #37170, opened 2020-03-26
Browser reload does not always trigger a sync if new file is found created via external write


mmattel:
### Steps to reproduce
1. Using a oC Google Drive mount and add a file via GD directly (not via oC)
2. Pressing force-reload on the browser accessing oC, the new file(s) appear
3. Sync does not get triggered
4. Renaming the file via browser accessing oC, the file gets renamed and synced.

### Expected behaviour
Files should get synced when force-reload identifies new files on the backend.

### Actual behaviour
At least one file needs a renaming, then all new files get synced

### Note
When doing the same with eg. SMB, files added directly, force reload on the browser also triggers the sync process. Seems that the trigger process for sync is not homogen

### Server configuration
**Operating system**: Ubuntu 18.04

**Web server:** nginx

**Database:** MariaDB

**PHP version:** 7.3

**ownCloud version:** (10.4)

**Updated from an older ownCloud or fresh install:** updated

**Where did you install ownCloud from:** tar

**Signing status (ownCloud 9.0 and above):**
all green

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "ocbdscdhwdo8",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "xxxx"
        ],
        "datadirectory": "xxxx\/data",
        "overwrite.cli.url": "https:\/\/xxxx",
        "dbtype": "mysql",
        "version": "10.4.0.4",
        "dbname": "xxxx",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
        "mysql.utf8mb4": true,
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "UTC",
        "installed": true,
        "filelocking.enabled": "true",
        "memcache.local": "\\OC\\Memcache\\Redis",
        "memcache.locking": "\\OC\\Memcache\\Redis",
        "memcache.distributed": "\\OC\\Memcache\\Redis",
        "redis": {
            "host": "\/var\/run\/redis\/redis-server.sock",
            "port": 0,
            "timeout": 0,
            "dbindex": 0
        },
        "filesystem_check_changes": 1,
        "excluded_directories": [
            ".snapshot",
            "~snapshot"
        ],
        "upgrade.automatic-app-update": true,
        "quota_include_external_storage": false,
[...]

PR #37218, opened 2020-04-07
Include option to allow scanner to reuse etags only for files

jvillafanez:
This option will be used by the watcher, in case it detects a change in
a folder, so the etag of the folder will always change (not just
depending on mtime) but the contents might retain its current etag.

## Description
Changes in gdrive weren't being propagated properly because the mtime of the folder doesn't change when a file is added / removed / modified there.
The etag of the folder didn't change, so the desktop client loses track of where to look for changes once it reaches the modified folder. This means that the new files inside the folder don't appear in the desktop (files aren't removed neither)

Note that the mtime of the folder in ownCloud still be the same as the one reported by gdrive

## Related Issue
https://github.com/owncloud/client/issues/7837#issuecomment-609791541

## Motivation and Context
gdrive integration don't behave like other external storages from a desktop client's perspective otherwise.

## How Has This Been Tested?
Manually checked:
1. Connect ownCloud with a gdrive account
2. Let desktop client sync normally the first time
3. Add a new file directly in a folder inside the gdrive account
4. Make a propfind to the target folder (you can navigate the ownCloud's web UI to the target folder instead) to detect the change
5. Sync client should pick up the change

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37218
+++ b/changelog/unreleased/37218
@@ -1,0 +1,17 @@
+Bugfix: Ensure ETag changes if a change is detected in a folder
+
+Previously, if a change was detected in a folder, the ETag of the folder only
+changed if the folder's mtime changed. The ETag propagation to the root folder
+was working fine.
+If the folder's mtime didn't change, the ETag of the folder didn't change neither.
+This behaviour was causing problems in the desktop client because it was looking for
+a change, but it lost track once the client reached the modified folder because the
+ETag was the same.
+This was detected in the GDrive storage integration. Other storage works without problems.
+Basically, the desktop client wasn't able to download newly-added files in GDrive
+because it was unable to find where those files were.
+
+The changes fix the problem mentioned above, so the GDrive storage integration keeps
+the same behaviour as other external storages
+
+https://github.com/owncloud/core/pull/37218
--- a/lib/private/Files/Cache/Scanner.php
+++ b/lib/private/Files/Cache/Scanner.php
@@ -188,13 +188,17 @@ class Scanner extends BasicEmitter implements IScanner {
 						$fileId = $cacheData['fileid'];
 						$data['fileid'] = $fileId;
 						// only reuse data if the file hasn't explicitly changed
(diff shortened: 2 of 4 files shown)


jvillafanez:
@pmaier1 I'll need confirmation whether this PR will be in 10.4.1 or 10.5.0.
If this will be available in 10.4.1, we'll need to close this PR and open a new one targeting 10.4.1

mmattel:
This could be related: https://github.com/owncloud/core/issues/37170 (Browser reload does not always trigger a sync if new file is found created via external write)

jvillafanez:
https://github.com/owncloud/core/issues/37170 should be fixed with this PR. The target use case is the same.

micbar:
@jvillafanez @mmattel I have doubts for 10.4.1.

It touches a central part of core (etag propagation)
One could argument, that this change requires a minor version.

I would like to avoid the testing effort now. 10.4.1 should be out  1-2 days. This would delay other important fixes.

IMO waiting for a fix of Gdrive external storage compared to a very urgent bugfix of password_policy is an easy decision.

@pmaier1 @HanaGemela open for your input.

HanaGemela:
I wouldn't delay 10.4.1

pmaier1:
10.4.1 is closed and should not be delayed by this. THX!

DeepDiver1975 (approved):

mrow4a:
@jvillafanez I think we might have similar issue with fed shares. 

Try to verify this scenario:
- pause sync client
- add the files in gdrive in test1/test1.1/test1.1.1/test1.1.1.1
- navigate in UI to test1, and to test1.1
- unpause to sync in sync client -> it might happen changes won't be detected

mmattel:
@mrow4a it would be a good idea to create a new issue so we can track it.
This comment gets definitely lost...

mrow4a:
@mmattel ignore it for now, external storages are working a bit differently than shared external storages. if we get similar issue, maybe this would be good hint for someone.

PR #37219, opened 2020-04-07
Stop writing data to the output buffer in case the connection is no l…

DeepDiver1975:
## Description
Bugfix: Stop writing data to the output buffer when the connection is not alive

Publicly shared video playback is sending a range http request to get the video
content. In cases where the user is seeking to different positions of the video
will result in a pretty high server load because all the video content is sent
to the browser. Without detecting the connection state on server side all data
is put to the output buffer.
With this change the server processes will stop sending data as soon as the
connection is detected as non-active.

https://github.com/owncloud/core/pull/37219

## How Has This Been Tested?
- :hand: 

## Types of changes
- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -475,6 +475,7 @@ class View {
 			$size = $this->filesize($path);
 			while (!\feof($handle)) {
 				echo \fread($handle, $chunkSize);
+				$this->checkConnectionStatus();
 				\flush();
 			}
 			return $size;
@@ -504,10 +505,10 @@ class View {
 						$len = $chunkSize;
 					}
 					echo \fread($handle, $len);
+					$this->checkConnectionStatus();
 					\flush();
 				}
-				$size = \ftell($handle) - $from;
-				return $size;
+				return \ftell($handle) - $from;
 			}
 
 			throw new \OCP\Files\UnseekableException('fseek error');
@@ -2267,5 +2268,12 @@ class View {
 	 */
 	public static function setIgnorePartFile($isIgnored) {
 		self::$ignorePartFile = $isIgnored;
+	}
+
+	private function checkConnectionStatus(): void {
+		$connectionStatus = \connection_status();
+		if ($connectionStatus !== 0) {
+			throw new \RuntimeException("Connection lost. Status: $connectionStatus");
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


lib/private/Files/View.php
```diff
@@ -2268,4 +2269,12 @@ private function createParentDirectories($filePath) {
...
 		self::$ignorePartFile = $isIgnored;
 	}
+
+	private function checkConnectionStatus(): void {
+		$connectionStatus = \connection_status();
+		if ($connectionStatus !== 0) {
+			throw new \RuntimeException("Connection lost. Status: $connectionStatus");
+		}
+	}
+
```
phil-davis, line 2279: https://drone.owncloud.com/owncloud/core/24179/2/6
code style does not like this @DeepDiver1975

lib/private/legacy/files.php
```diff
@@ -192,9 +192,6 @@ public static function get($dir, $files, $params = null) {
 			$l = \OC::$server->getL10N('lib');
 			/* @phan-suppress-next-line PhanUndeclaredMethod */
 			$hint = \method_exists($ex, 'getHint') ? $ex->getHint() : '';
-			if ($event->hasArgument('message')) {
```
VicDeo, line 195: intentionally removed?

DeepDiver1975: Yes. There is no $event. Code goes ☀️

VicDeo (approved):

lib/private/Files/View.php
```diff
@@ -475,6 +475,7 @@ public function readfile($path) {
 			$size = $this->filesize($path);
 			while (!\feof($handle)) {
 				echo \fread($handle, $chunkSize);
+				$this->checkConnectionStatus();
```
phil-davis, line 478: Can we get unit test coverage of `readfile` and `readfilePart` ?

`tests/lib/Files/ViewTest.php` does not seem to have "nice" test cases that clearly test each public method of this class.

phil-davis: ^ will be something for some day - getting better unit tests for old code like this.

DeepDiver1975: testing the connections status in unit tests is hard to do - one needs to mock this runtime function and then all you can test is that the status is !==0 and an exception is thrown ...

PR #37238, opened 2020-04-11
Replace jeremeamia/superclosure with opis/closure

phil-davis:
## Description

`jeremeamia/superclosure` library is no longer maintained and is only officially tested up to PHP 7.2. 
https://github.com/jeremeamia/super_closure/blob/master/README.md

Replace it with the recommended https://github.com/opis/closure library.

```
composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 1 install, 0 updates, 3 removals
  - Removing symfony/polyfill-util (v1.15.0)
  - Removing symfony/polyfill-php56 (v1.15.0)
  - Removing jeremeamia/superclosure (2.4.0)
  - Installing opis/closure (3.5.1): Loading from cache
```

This happens to also get rid of some `symfony/polyfill` dependencies.
And means we are using a dependency that is officially tested on PHP 7.3 and 7.4

## How Has This Been Tested?
CI and local run of unit tests:
```
make test-php-unit TEST_PHP_SUITE=tests/lib/Command/AsyncBusTest.php
```

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Command/ClosureJob.php
+++ b/lib/private/Command/ClosureJob.php
@@ -22,16 +22,17 @@
 namespace OC\Command;
 
 use OC\BackgroundJob\QueuedJob;
-use SuperClosure\Serializer;
 
 class ClosureJob extends QueuedJob {
 	protected function run($serializedCallable) {
-		$serializer = new Serializer();
-		$callable = $serializer->unserialize($serializedCallable);
-		if (\is_callable($callable)) {
-			$callable();
-		} else {
-			throw new \InvalidArgumentException('Invalid serialized callable');
+		$serializedClosure = \unserialize($serializedCallable);
+		if (\method_exists($serializedClosure, 'getClosure')) {
+			$callable = $serializedClosure->getClosure();
+			if (\is_callable($callable)) {
+				$callable();
+				return;
+			}
 		}
+		throw new \InvalidArgumentException('Invalid serialized callable');
 	}
 }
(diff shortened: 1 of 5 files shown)


lib/private/Command/ClosureJob.php
```diff
@@ -22,12 +22,11 @@
 namespace OC\Command;
 
 use OC\BackgroundJob\QueuedJob;
-use SuperClosure\Serializer;
 
 class ClosureJob extends QueuedJob {
 	protected function run($serializedCallable) {
-		$serializer = new Serializer();
-		$callable = $serializer->unserialize($serializedCallable);
+		$serializedClosure = \unserialize($serializedCallable);
```
VicDeo, line 28: compare result with `false` to avoid `getClosure()` on `false` below?

VicDeo (changes requested):

lib/private/Command/ClosureJob.php
phil-davis: I used `method_exists` because it is also nice to know that the unserialized "thing" is an object that has the required method. I checked that if `$serializedClosure` is not even an object (e.g. is `false` or "abc" or...) then `method_exists` happily returns false.

VicDeo (approved):

Issue #36803, opened 2020-01-22
Cannot share top-level of read-only storage


phil-davis:
### Steps to reproduce
1. As an admin, in settings-admin-storage create some local storage that makes some folder on the local server available - e.g. `local-folder`
2. set "Available for" to some user "Anne" (or group that Anne is a member of)
2. select the "enable sharing" and "Set read-only" options
3. as user "Anne" try to share `local-folder` with "Bob"

### Expected behaviour
The share should be created with just the allowed "read" permissions

### Actual behaviour
Notification is displayed:
```
Cannot set the requested share permissions for local-folder
```
and the share is not created.

There is no way to get to the sharing permissions checkboxes and turn off "edit".

(note: it works fine when sharing a sub-folder or file inside `local-folder`. In that case, it "knows" to make the share read-only)

### Server configuration
Current core master

The "Set read-only" option for storage was added by PR https://github.com/owncloud/core/pull/36397 and issue https://github.com/owncloud/core/issues/29873

phil-davis:
A similar case is when Anne has received a share from someone and has "can share" permission and the received share is read-only.

When Anne tries to reshare the received share to Bob, the UI happily understands that the received share is read-only, and so it makes the reshare read-only also. Anne does not even see the not-allowed permission checkboxes for "can edit"...

I guess that in the read-only `local-folder` case, the UI is not aware that `local-folder` is read-only.

davitol:
AFAIK It has been always the behaviour implemented and we decided that 'was fine' in this way (cc @pmaier1 just maybe for more feedback)

phil-davis:
This is a bit different new scenario. Local storage can now be explicitly set to read-only (regardless of the underlying storage file/folder permissions). So now "ordinary installs" can set local-storage to read-only. When they do that, there is this webUI problem (which maybe is something we "live with" or we fix)

micbar:
@pmaier1 IMO we can leave that now

phil-davis:
Moved this to QA/CI to make sure there is some acceptance test that demonstrates the behavior.

- [ ] make webUI test scenario to demonstrate the behavior
- [ ] make API test scenario to demonstrate the behavior

Tag the scenarios with this issue number.

Then someone can decide if the behavior needs to be changed.

pmaier1:
> This is a bit different new scenario. Local storage can now be explicitly set to read-only (regardless of the underlying storage file/folder permissions).

[...]

PR #37262, opened 2020-04-15
[Tests-Only] Add acceptance tests to share top-level of read-only storage

jasson99:
## Description
Add acceptance tests to share top-level of read-only storage

## Related Issue
#36803 

## How Has This Been Tested?
CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/OccContext.php
+++ b/tests/acceptance/features/bootstrap/OccContext.php
@@ -191,6 +191,31 @@ class OccContext implements Context {
 
 	/**
 	 * @param string $mountPoint
+	 * @param boolean $setting
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function setExtStorageSharingUsingTheOccCommand($mountPoint, $setting = true) {
+		$command = "files_external:option";
+
+		$mountId = $this->featureContext->getStorageId($mountPoint);
+
+		$key = "enable_sharing";
+
+		if ($setting) {
+			$value = "true";
+		} else {
+			$value = "false";
+		}
+
+		$this->invokingTheCommand(
+			"$command $mountId $key $value"
+		);
+	}
+
+	/**
+	 * @param string $mountPoint
 	 * @param string $setting "never" (switch it off) otherwise "Once every direct access"
 	 *
 	 * @return void
@@ -932,6 +957,19 @@ class OccContext implements Context {
 	 */
 	public function theAdminHasSetTheExtStorageToReadOnly($mountPoint) {
 		$this->setExtStorageReadOnlyUsingTheOccCommand($mountPoint);
+		$this->theCommandShouldHaveBeenSuccessful();
+	}
+
+	/**
+	 * @Given the administrator has set the external storage :mountPoint to sharing
+	 *
+	 * @param string $mountPoint
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function theAdminHasSetTheExtStorageToSharing($mountPoint) {
+		$this->setExtStorageSharingUsingTheOccCommand($mountPoint);
 		$this->theCommandShouldHaveBeenSuccessful();
 	}
 
--- a/tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
+++ b/tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
@@ -1,0 +1,40 @@
+@cli @skipOnLDAP @local_storage
+Feature: create local storage and enable read-only and sharing from the command line
+  As an admin
+  I want to create read-only local storage and enable sharing from the command line
+  So that local folders on my server can be made visible but read-only to users of ownCloud
+
+  Background:
+    Given these users have been created with default attributes and without skeleton files:
+      | username |
+      | user0    |
+      | user1    |
+    And the administrator has created the local storage mount "local_storage1"
+    And the administrator has uploaded file with content "this is a file in local storage" to "/local_storage1/file-in-local-storage.txt"
+    And the administrator has added user "user0" as the applicable user for the last local storage mount
+    And the administrator has set the external storage "local_storage1" to read-only
+    And the administrator has set the external storage "local_storage1" to sharing
+
+  Scenario: applicable user is not able to share with all permissions the top-level of read-only storage
+    When user "user0" shares folder "local_storage1" with user "user1" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "404"
+    And the OCS status message should be "Cannot set the requested share permissions for local_storage1"
+
+  Scenario: applicable user is able to share with read permissions the top-level of read-only storage
+    When user "user0" shares folder "local_storage1" with user "user1" with permissions "read" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" folder "local_storage1" should exist
+
+  Scenario: applicable user is able to share with read permissions the file inside the top-level of read-only storage
+    When user "user0" shares file "/local_storage1/file-in-local-storage.txt" with user "user1" with permissions "read" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" file "file-in-local-storage.txt" should exist
+
+  Scenario: applicable user is able to share with default permissions the file inside the top-level of read-only storage
+    When user "user0" shares file "/local_storage1/file-in-local-storage.txt" with user "user1" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" file "file-in-local-storage.txt" should exist
--- a/tests/acceptance/features/webUIAdminSettings/adminStorageSettings.feature
+++ b/tests/acceptance/features/webUIAdminSettings/adminStorageSettings.feature
@@ -163,3 +163,22 @@ Feature: admin storage settings
     Then the last created local storage mount should be listed on the webUI
     And the user re-logs in as "user1" using the webUI
     And folder "local_storage1" should be listed on the webUI
+
+  @issue-36803
+  Scenario: applicable user is not able to share top-level of read-only storage
+    Given these users have been created with default attributes and without skeleton files:
+      | username |
+      | user0    |
+      | user1    |
+    And the administrator has enabled the external storage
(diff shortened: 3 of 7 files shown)


tests/acceptance/features/webUIAdminSettings/adminStorageSettings.feature
```diff
@@ -163,3 +163,20 @@ Feature: admin storage settings
...
+    And the administrator has enabled the external storage
+    And the administrator has browsed to the admin storage settings page
+    And the administrator has created the local storage mount "local_storage1" from the admin storage settings page
+    And the administrator has added user "user0" as the applicable user for the last local storage mount from the admin storage settings page
+    And the administrator has enabled read-only for the last created local storage mount using the webUI
+    And the administrator has enabled sharing for the last created local storage mount using the webUI
+    And the user has re-logged in as "user0" using the webUI
+    When the user shares folder "local_storage1" with user "User One" using the webUI
+    Then notifications should be displayed on the webUI with the text
+      | Cannot set the requested share permissions for local_storage1 |
```
Talank, line 182: This is the actual behaviour. What is the Expected behaviour? Something like `The share should be created with just the allowed "read" permissions` as written in the tagged issue. 
I think you should write that step in comment !!

jasson99: I went through the discussion in the issue and thought that the expected behavior is not clear, even though the issue mentions the expected behavior to be "share should be created with just the allowed "read" permissions". I am not sure, therefore, whether to write the expected behavior in comment or not.

jasson99: @phil-davis what do you think?

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
```diff
@@ -0,0 +1,21 @@
+@cli @skipOnLDAP @local_storage
+Feature: create local storage and enable read-only and sharing from the command line
```
phil-davis, line 2: Also:
- add a scenario where the local storage has a folder inside it, and the applicable user shares the folder - that sharing should work.
- add a feature for createLocalStorageAndShare that has the default read-write and has sharing enabled. Put 2 scenarios in that feature - sharing the top-level and sharing a sub-folder. Both of those scenarios should work.

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
```diff
@@ -0,0 +1,21 @@
...
+      | user0    |
+      | user1    |
+    And the administrator has created the local storage mount "local_storage1"
+    And the administrator has added user "user0" as the applicable user for the last local storage mount
+    And the administrator has set the external storage "local_storage1" to read-only
+    And the administrator has set the external storage "local_storage1" to sharing
+    When user "user0" shares folder "local_storage1" with user "user1" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "404"
+    And the OCS status message should be "You are not allowed to share /user0/files/local_storage1"
```
phil-davis, line 20: Actually this is probably correct behavior. The user can share only with read-only permissions, and the `When` step will be asking for all permissions?

So also add a scenario where user "user0" shares folder "local_storage1" with only read permission.
That should work. Maybe it will work with the API (because IMO the bug/issue is only in the UI).

If the API scenarios all work as expected, then the issue tag can be removed from those.

tests/acceptance/features/webUIAdminSettings/adminStorageSettings.feature
phil-davis: It is a bit hard to know exactly what `Then` steps will be "expected" when the issue is fixed, probably something like:
`And as "user1" folder "local_storage1" should exist`
So you could write that as a comment line, and as real line put:
`And as "user1" folder "local_storage1" should not exist`

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
jasson99: fixed

tests/acceptance/features/webUIAdminSettings/adminStorageSettings.feature
jasson99: fixed

tests/acceptance/features/bootstrap/OccContext.php
```diff
@@ -189,6 +189,31 @@ public function setExtStorageReadOnlyUsingTheOccCommand($mountPoint, $setting =
...
+		$key = "enable_sharing";
+
+		if ($setting) {
+			$value = true;
+		} else {
+			$value = false;
+		}
+
+		$this->invokingTheCommand(
+			"$command $mountId $key $value"
```
phil-davis, line 213: When `$value` is `true` this will insert "1" at the end of the command.
When `$value` is `false` this will put the empty string at the end of the command. IMO it does not put "0"

It seems a bit obscure. Why not set `$value` to the strings that should be put in the command?

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
```diff
@@ -0,0 +1,42 @@
...
+    And the OCS status code should be "404"
+    And the OCS status message should be "Cannot set the requested share permissions for local_storage1"
+
+  Scenario: applicable user is able to share with read permissions the top-level of read-only storage
+    When user "user0" shares folder "local_storage1" with user "user1" with permissions "read" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" folder "local_storage1" should exist
+
+  Scenario: applicable user is able to share the file inside the top-level of read-only storage
```
phil-davis, line 30:
```suggestion
  Scenario: applicable user is able to share with read permissions the file inside the top-level of read-only storage
```

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
```diff
@@ -0,0 +1,42 @@
...
+    When user "user0" shares file "/local_storage1/file-in-local-storage.txt" with user "user1" with permissions "read" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" file "file-in-local-storage.txt" should exist
+
+  Scenario: applicable user is not able to share the top-level of read-only storage
+    When user "user0" shares file "/local_storage1/file-in-local-storage.txt" with user "user1" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" file "file-in-local-storage.txt" should exist
```
phil-davis, line 40: I am surprised that this passes. But maybe when the share is created without specifying any permissions, then the server automatically sets read-only permissions.

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
```diff
@@ -0,0 +1,42 @@
...
+    And the OCS status code should be "100"
+    And as "user1" folder "local_storage1" should exist
+
+  Scenario: applicable user is able to share the file inside the top-level of read-only storage
+    When user "user0" shares file "/local_storage1/file-in-local-storage.txt" with user "user1" with permissions "read" using the sharing API
+    Then the HTTP status code should be "200"
+    And the OCS status code should be "100"
+    And as "user1" file "file-in-local-storage.txt" should exist
+
+  Scenario: applicable user is not able to share the top-level of read-only storage
```
phil-davis, line 36:
```suggestion
  Scenario: applicable user is able to share with default permissions the top-level of read-only storage
```
Maybe this is the correct scenario description?

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
phil-davis: Then you could add a scenario that explicitly tries to share the file with all permissions, and that should fail.

tests/acceptance/features/bootstrap/OccContext.php
jasson99: when `$value` is true, doesn't it put `true` at the end of command?..And `false` when `$value` is false.?

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
jasson99: fixed

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
jasson99: I believe that the share is created with all permissions when nothing is specified..That's what makes the difference in scenario:18 and scenario:24. And I also tried with all permissions declared explicitly and it passes. so I thought of not writing two different scenarios for the same thing, with default permissions, and with all permissions specified. What do you think?

tests/acceptance/features/cliLocalStorage/createLocalStorageReadOnlyAndShare.feature
jasson99: fixed

tests/acceptance/features/bootstrap/OccContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/OccContext.php
```diff
@@ -189,6 +189,31 @@ public function setExtStorageReadOnlyUsingTheOccCommand($mountPoint, $setting =
...
+		$key = "enable_sharing";
+
+		if ($setting) {
+			$value = "true";
+		} else {
+			$value = "false";
+		}
+
+		$this->invokingTheCommand(
+			"$command $mountId $key $value"
```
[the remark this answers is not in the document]
jasson99: I think it is working this way as well

phil-davis (approved):

phil-davis:
I rebased and squashed the commits. This is ready-to-merge when CI passes.

PR #37263, opened 2020-04-15
[Tests-Only] Bump phan 2.7

phil-davis:
## Description
- bump `phan` to 2.7 and exclude various `routes.php` files that do not analyze nicely
- Remove extra setLatestChildDateTime in prepareCommentForDatabaseWrite
- Teach phan about the objects in settings/Controller/UsersController.php (there were already various hints to phan about `groupManager` and it also then gets confused about `userManager`)

## Related Issue
- Fixes https://github.com/owncloud/QA/issues/644

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Comments/Comment.php
+++ b/lib/private/Comments/Comment.php
@@ -302,7 +302,7 @@ class Comment implements IComment {
 	/**
 	 * sets the date of the most recent child
 	 *
-	 * @param \DateTime $dateTime
+	 * @param \DateTime|null $dateTime
 	 * @return IComment
 	 * @since 9.0.0
 	 */
--- a/lib/private/Comments/Manager.php
+++ b/lib/private/Comments/Manager.php
@@ -113,7 +113,7 @@ class Manager implements ICommentsManager {
 		if ($comment->getId() === '') {
 			$comment->setChildrenCount(0);
 			$comment->setLatestChildDateTime(new \DateTime('0000-00-00 00:00:00', new \DateTimeZone('UTC')));
-			$comment->setLatestChildDateTime(null);
+			$comment->setLatestChildDateTime();
 		}
 
 		if ($comment->getCreationDateTime() === null) {
--- a/lib/public/Comments/IComment.php
+++ b/lib/public/Comments/IComment.php
@@ -203,11 +203,11 @@ interface IComment {
 	/**
 	 * sets the date of the most recent child
 	 *
-	 * @param \DateTime $dateTime
+	 * @param \DateTime|null $dateTime
 	 * @return IComment
 	 * @since 9.0.0
 	 */
-	public function setLatestChildDateTime(\DateTime $dateTime);
+	public function setLatestChildDateTime(\DateTime $dateTime = null);
 
 	/**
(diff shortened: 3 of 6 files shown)


lib/private/Comments/Manager.php
```diff
@@ -113,7 +113,6 @@ protected function prepareCommentForDatabaseWrite(IComment $comment) {
 		if ($comment->getId() === '') {
 			$comment->setChildrenCount(0);
 			$comment->setLatestChildDateTime(new \DateTime('0000-00-00 00:00:00', new \DateTimeZone('UTC')));
-			$comment->setLatestChildDateTime(null);
```
phil-davis, line 116: @DeepDiver1975 or anyone - any idea why `setLatestChildDateTime` was called here a 2nd time, and with `null` ?
`phan`  complained about it because `setLatestChildDateTime` is supposed to always be passed a `DateTime`

individual-it (approved):

lib/private/Comments/Manager.php
```diff
@@ -113,7 +113,6 @@ protected function prepareCommentForDatabaseWrite(IComment $comment) {
 		if ($comment->getId() === '') {
 			$comment->setChildrenCount(0);
 			$comment->setLatestChildDateTime(new \DateTime('0000-00-00 00:00:00', new \DateTimeZone('UTC')));
```
phil-davis, line 115: This causes unit test fails:
```
SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: '-0001-11-30 00:00:00' for column `owncloud`.`oc_comments`.`latest_child_timestamp` at row 1
```
https://drone.owncloud.com/owncloud/core/24277/15/10

So actually we need the following line `$comment->setLatestChildDateTime(null);`
And we need to fixup the PHPdoc for `setLatestChildDateTime` so it officially allows the parameter to be `null`

lib/private/Comments/Comment.php
```diff
@@ -302,7 +302,7 @@ public function getLatestChildDateTime() {
 	/**
 	 * sets the date of the most recent child
 	 *
-	 * @param \DateTime $dateTime
+	 * @param \DateTime|null $dateTime
```
phil-davis, line 305: The parameter already defaults to null, so put that in the PHPdoc

lib/private/Comments/Manager.php
phil-davis: So whay is this line here at all?

lib/public/Comments/IComment.php
```diff
@@ -203,11 +203,11 @@ public function getLatestChildDateTime();
 	/**
 	 * sets the date of the most recent child
 	 *
-	 * @param \DateTime $dateTime
+	 * @param \DateTime|null $dateTime
```
phil-davis, line 206: The implementation of this allows `null`, so officially allow `null` in the interface definition.

Issue #36509, opened 2019-12-03
Support PHP 7.4


phil-davis:
PHP 7.4 was released last week. It is now available in `owncloud-ci/php:7.4`

Enable support for PHP 7.4 in core.

- [x] run CI with 7.4 (see WIP PR #37280 )
- [x] investigate and fix the fails that are reported
- [x] think about other QA needed to be confident that core is ready for PHP 7.4 - regular release smoke tests should be enough
- [x] sort out PHP 7.4 support for apps https://github.com/owncloud/QA/issues/645 https://github.com/owncloud/QA/issues/647

phil-davis:
PR #36500 gets various errors/warnings in `phpunit` and in some other test suites. So this is not a completely trivial job. Needs to be scheduled.

phil-davis:
@micbar I am seeing developers mention PHP 7.4 more often - I think some people's Mac's etc have auto-updated to PHP 7.4... and they are trying to work with codebase...

What is the priority for getting this to happen (e.g. immediately that we get 10.4 release done?)

jvillafanez:
* [ ] Need to update google API library for the `implode` order deprecation (https://github.com/owncloud/core/tree/master/apps/files_external/3rdparty/google-api-php-client)

Helios07:
This month' release of Fedora 32 will bring php7.4. Do you see the php7.4 support in 10.5 so that the upgrade will not leave a Fedora based owncloud server unusable?

phil-davis:
- [ ] might need Guzzle6 (some code in #37011 - see my notes there)

phil-davis:
For reference, these are PRs this week that adjusted unit tests and related to get PHP unit tests passing for PHP 7.4:
#37279 
#37281 
#37288 
#37290 
#37292 
#37293 
#37297 
#37298 
#37300

micbar:
@phil-davis Can we close this?

phil-davis:
@jvillafanez you made a couple of checklust items above. Are they still relevant?
If so, should we raise individual issues?

jvillafanez:
There is a PR with the gdrive lib update that should be reviewed and merged.
The other check is just a small detail. We can ignore it for now and make the change on demand.

phil-davis:
PR  #37265 "Gdrive lib update"

phil-davis:
PR #37265 has been merged.

PR #37265, opened 2020-04-15
Gdrive lib update

jvillafanez:
## Description
* Update google library to version 2.4.1
* There are some update of the library code for icewind's 3.1.2 (code in the repo seems older than the last composer update. Maybe the code wasn't updated. No real update in the composer's dependency)
* In addition, google's new version uses guzzle, so try to enforce guzzle version to 5.3.1 (matching core's version). Using guzzle 6 might reduce the amount of dependencies brought by google, but could be problematic for core.

First commit is the library update, both google and icewind.
The second commit contains the actual changes to adapt to the API changes.
The third commit is just readability improvements.

Note that change detection has been changed because the API no longer provides a id per change, so using the google's change API was a problem. It probably isn't suited for the use case implemented in the code.

## Related Issue
https://github.com/owncloud/core/issues/36509#issuecomment-597061791

## Motivation and Context
Update google's library version to be compatible with latest php 7.4 deprecations.

## How Has This Been Tested?
Tested manually basic file operations through ownCloud.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### Requires testing for gdrive and SMB connectors due to library updates


jvillafanez:
:exclamation: Please double-check "[gdrive_lib_update bcbe4f753b] Fix google docs downloads" is there :exclamation:

jvillafanez:
There is a known issue when a file is re-uploaded (just upload the same file twice or more): the file is duplicated in the gdrive with the same name, which will be problematic (we don't handle multiple files with the same name)

It seems the problem is with the ownCloud's .part file. Disabling the .part files in gdrive in https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/File.php#L637 would solve the problem because the file will be uploaded directly to the previous file, overwriting it.
We need to decide how to proceed with this issue. Adding the gdrive storage there is the simplest solution, but a bad approach.

mmattel:
I am adding @micbar so he knows about this PR and may add his thoughts

micbar:
@jvillafanez what is the question?

jvillafanez:
The question is if we want to spend time refactoring https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/File.php#L637 in order to provide a good solution, or if we include directly the storage name there to prevent using .part files in google drive.

cdamken:
IMHO .part files should be always in the temp folder and then upload to the external storage once is finished (and even checked by the antivirus if enabled)

jvillafanez:
Decided to add the google storage to the core's ".part file exclusion list" for now in order to progress with the PR.
Opened https://github.com/owncloud/core/issues/37313 to discuss a possible refactor.

ho4ho:
.part file related by #37062, #25826 ?

phil-davis:
This is also related to https://github.com/owncloud/core/issues/36509#issuecomment-623394556 for PHP 7.4 support.

phil-davis:
I added this to the 10.5.0 project. It needs review and moving forward.

phil-davis:
I rebased to get this up-to-date with other stuff that has happened in core master in the last week.

@jvillafanez @micbar what happens next?

This needs to move forward - it is becming a blocker if we need it for 10.5.0

jvillafanez:
Code-wise this is ready. It still need review.
Ignoring the first commit which contains both library update, the rest of the commits should be easy to understand.

micbar:
@jvillafanez I requested a review. Hope that helps ;-)

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -84,11 +84,33 @@ public function getId() {
 	}
 
 	private function getDefaultFieldsForFile() {
-		return 'id,name,mimeType,parents,capabilities/canEdit,size,viewedByMeTime,createdTime,modifiedTime';
+		static $fileFields = null;  // keep static variable to avoid recreating the string
```
VicDeo: IMO static is kinda functional thing. class constant suits better here

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -84,11 +84,33 @@ public function getId() {
...
+				'createdTime',
+				'modifiedTime',
+			]);
+		}
+		return $fileFields;
 	}
 
 	private function getDefaultFieldsForFolderScan() {
-		return 'incompleteSearch,nextPageToken,files(id,name,mimeType,parents,capabilities/canEdit,size,viewedByMeTime,createdTime,modifiedTime)';
+		static $folderFields = null;  // keep static variable to avoid recreating the string
```
VicDeo: the same as above. could be a class property.

VicDeo (changes requested):

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -663,59 +657,38 @@ public function hasUpdated($path, $time) {
 		if ($this->is_file($path)) {
 			return parent::hasUpdated($path, $time);
 		} else {
```
karakayasemi: If upper `if statement` ends with a `return` like here, no need to open an `else` block. I am aware of cyclomatic complexity is so high in the `Google.php` class wide, but, optionally we may reduce this complexity at least the places that we touch in this pr. It increases readability by reducing number of multiple nested conditonal blocks.

karakayasemi (commented):
Reviewing deeply is not easy. The changeset looks okay, basic file operations and authorization are working.

apps/files_external/lib/Lib/Storage/Google.php
jvillafanez: I'm not sure what it the best option. The problem with the class constant is that is easy that the field list becomes unreadable if a string is used. If we use an array, we'll have to build the string every time.
Another alternative could be to use a class attribute, but for me it would be polluting the class with information that isn't part of the class. In addition, the attribute could be changed in any part of the class.

I think this is a way the keep balance between readability and performance. The string is built only once when needed, and using an array makes it easier to read. In addition, all this complexity is isolated inside the method, without polluting the class.

apps/dav/lib/Connector/Sabre/Node.php
```diff
@@ -222,7 +222,12 @@ public function setETag($etag) {
 	 * @return integer
 	 */
 	public function getSize() {
-		return $this->info->getSize();
+		$size = $this->info->getSize();
+		if ($size < 0) {
+			return null;
```
VicDeo, line 227: this invalidates the current PHPDoc for `getSize`

VicDeo (changes requested):

VicDeo (approved):

karakayasemi (approved):

phil-davis:
`codecov/project` fails because of the change in size of all the "crud" here.... - I will override that.

Issue #37267, opened 2020-04-16
Create a user "files_encryption"


cs35-owncloud:
Hello everyone,

first don't do this on an ownCloud production !
I've noticed that the apps **files_external**  and **files_encryption** create a directory in the **datadirectory**. It's then possible to destroy them by creating a user with the same name.

### Steps to reproduce
1. Install ownCloud
2. Enable master key encryption
3. Create the user "files_encryption"
4. Delete the user "files_encryption"

### Expected behaviour
I'm thinking you shouldn't be able to create that user, there should be a blacklist of usernames. Or maybe not put any useful thing in the **datadirectory**.

### Actual behaviour
The content of **datadirectory**/files_encryption will be destroyed, your encryption keys too. The same happens for files_external if you create a user with that name. It seems the keys get re-generated but are not the same as previously (I checked the md5sum) and previously uploaded files can be browsed. But newly uploaded files are failing the preview. I'm a bit puzzled since everything seems to be working even if the keys have been destroyed and then changed.

### Server configuration
**Operating system**:

**Web server:** NGinx 1.16.1

**Database:** MySQL 5.7

**PHP version:** 7.3.16

**ownCloud version:** 10.3.2

**Updated from an older ownCloud or fresh install:** fresh install

**Where did you install ownCloud from:** tarball

**Signing status (ownCloud 9.0 and above):** all good

**The content of config/config.php:**

```
# occ config:list system
{
    "system": {
        "datadirectory": "\/var\/owncloud\/data",
        "version.hide": true,
        "logtimezone": "UTC",
        "installed": true,
        "upgrade.disable-web": true,
        "upgrade.automatic-app-update": false,
        "updatechecker": false,
        "updater.server.url": "127.0.0.1",
        "maintenance": false,
        "user.search_min_length": 2,
        "loglevel": 2,
        "log_type": "syslog",
        "singleuser": false,
        "filelocking.enabled": true,
        "memcache.local": "\\OC\\Memcache\\Redis",
        "redis": {
            "host": "localhost",
            "port": 6379,
            "timeout": 0
        },
        "memcache.locking": "\\OC\\Memcache\\Redis",
        "apps_paths": [
            {
                "path": "\/var\/owncloud\/apps_orig",
                "url": "\/apps_orig",
                "writable": false
            },
            {
                "path": "\/var\/owncloud\/apps",
                "url": "\/apps",
                "writable": false
            },
            {
                "path": "\/var\/owncloud\/theme",
[...]

Issue #32547, opened 2018-09-03
Username 'meta' login problem


phil-davis:
### Steps to reproduce
1. Create a user named ``meta``
2. Try to login on the webUI as user ``meta``

### Expected behaviour
Normal login, skeleton files displayed.

### Actual behaviour
"Internal server error"

Note: I added some ``error_log()`` calls to the various ``newFolder()`` methods to see which one is being called and the value of ``$path``
```
[Mon Sep  3 08:24:35 2018] PDD: AbstractFolder newFolder /files
[...]

Issue #30438, opened 2018-02-09
After delete the user "avatars", the avatars for all users are deleted


prsnbrg:
### Steps to reproduce
1. Add the user "avatars" and log in with this user
2. Upload an avatar with another user
3. Delete the user "avatars"

### Expected behaviour
Tell us what should happen
Delete all personal files from "avatar" user.

### Actual behaviour
Tell us what happens instead
Delete the completly avatar directory with all avatars from other users.

### Server configuration
**Operating system**:

**Web server:**

**Database:**

**PHP version:**

**ownCloud version:** (see ownCloud admin page)
10.0.6.1
**Updated from an older ownCloud or fresh install:**
Update from 10.0.4 
**Where did you install ownCloud from:**
Web Installer and the update over the Shell

phil-davis:
I will bet that doing this for any other "special" folder name will create similar havoc.
e.g. I have a ``files_external`` folder in my ``data`` folder, along with my ``admin`` ``user`` ``phil`` etc folders. I bet if I create a user called ``files_external`` there will be problems.

The namespace of the folder where the files for each user are stored should not be polluted with "special" folder names. But maybe it is too late to sort that out.

Ban those "special" names as user names?

prsnbrg:
I have test this for the user `files_external` and i have the same problem, as the avatar user.

I think the best solution is a ban for the names `avatar` and `files_external`

phil-davis:
And what other "special" folder names "appear" in the ``data`` dir?

hodyroff:
Ban does not help if we get the user in an external way aka LDAP?!

phil-davis:
True
The ownCloud ``data`` directory typically holds all the "user data files" (i.e. everything that is not the code, and not a record in the database). So it needs to have folders for each type of "namespaced" files to be stored, and so if there are files that are not to be stored in the individual user's folder tree, then there needs to be separate folders for that stuff - e.g. ``avatars`` and ``files_external`` have been mentioned here.

So the per-user folder(s) need to go down a level:
```
data/avatars
data/files_external
data/other-files-that-have-to-go-somewhere
data/users
data/users/admin
data/users/phil
data/users/user1
```
a change like that would be a significant migration! But might not take long to run because it just moves the folder for each user down a level. (and has to be careful about the process if there is already a user called ``users`` !)

butonic:
IMO users files should reside in data/files/<uid>
[...]

PR #37268, opened 2020-04-16
Do not allow various special usernames (UIDs)

phil-davis:
## Description
If UIDs are created with these "special" values then "bad things happen" because there are folders in the `data` directory, or end-points that become ambiguous or...
```
avatars
files_external
files_encryption
meta
```

Prevent them from being created as local users.

(will also need some action for user_ldap to sort out what to do if a UID like this exists in LDAP)

## Related Issue
#30438 and #32547 and #37267 

## How Has This Been Tested?
CI and manually try to create these users in the webUI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/32547
+++ b/changelog/unreleased/32547
@@ -1,0 +1,8 @@
+Change: Disallow various special usernames
+
+Special names "avatars", "files_encryption", "files_external" and "meta" are
+used for other purposes in ownCloud and are not valid usernames (UIDs).
+Creating a user with any of these names is now disallowed.
+
+https://github.com/owncloud/core/issues/32547
+https://github.com/owncloud/core/pull/37268
--- a/lib/private/User/Manager.php
+++ b/lib/private/User/Manager.php
@@ -366,6 +366,17 @@ class Manager extends PublicEmitter implements IUserManager {
 				throw new \Exception($l->t('The username can not be longer than 64 characters'));
 			}
 
+			$invalidUids = [
+				'avatars',
+				'meta',
+				'files_external',
+				'files_encryption'
+			];
+
+			if (\in_array(\strtolower($uid), $invalidUids)) {
+				throw new \Exception($l->t("The special username $uid is not allowed"));
+			}
+
 			// No empty password
 			if (\trim($password) == '') {
 				throw new \Exception($l->t('A valid password must be provided'));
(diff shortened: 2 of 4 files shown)


phil-davis:
Comments please from anybody. 
I am proposing that, for the time being, we disallow these special usernames. (Some day we could re-engineer the universe so that this unfortunate  stuff does not happen)

Are there others that should be added to the list?

micbar:
good improvement.
I tested this in the past, was surprised that thes names were allowed.

cs35-owncloud:
Hey,

how about *updater-data* I think it's a directory created by the updater.

phil-davis:
> Hey,
> 
> how about _updater-data_ I think it's a directory created by the updater.

Anyone know about this? I  didn't find the string mentioned in the core repo.

Can anyone think of other special strings that should be prohibited?

cs35-owncloud:
It's related to this [doc](https://doc.owncloud.org/server/10.4/admin_manual/maintenance/update.html) 

```
Creates an updater_backup directory under your ownCloud data directory
```

I don't know why I found *updater-data* in my case on some ownClouds, but *updater_backup* seems official, I guess.

PVince81:
@VicDeo might know about the updater

jvillafanez:
Is it worthy to provide some kind of "UserUidValidator" to be injected there, instead of using a fixed list of names? We could use a chain of implementations, such as a hardcoded list or a list configurable via config.php file.
Basically, we inject a "ValidatorChain", which would contain one or more validators. If the username passes through all the validators, the name is valid and could be used.

This could be more flexible in case we miss a name, or if the admin wants to forbid a particular name for whatever reason.

micbar:
@jvillafanez Yes, please go in that direction.

phil-davis:
@jvillafanez please take over this PR. (Or make a new one)
I raised this initial code just to get things moving.

VicDeo:
@phil-davis 
>Anyone know about this? I didn't find the string mentioned in the core repo.

Updater is located in a separate repo  and the path to it's data is  `updater-data` indeed
https://github.com/owncloud/updater/blob/master/src/Utils/Locator.php#L183

jvillafanez:
@micbar (and anyone who wants to say anything), I think we need to decide what happen to the accounts that have been created before making the uid invalid.
Use case could be similar to:
1. Sync one backend
2. Notice that one of the users has  a potentially dangerous uid
3. Invalidate that uid

For now, the code doesn't really do anything to that account, but the user won't be able to login (there is a crash I have to check, caused by the invalid uid).
We might consider to disable the account, although the deletion would need to be performed by the admin at some point (better let the admin to take the decision).
I haven't checked in depth, but I guess all the interactions with the "dangerous" account are still possible.

The current code won't sync accounts (looking particularly at LDAP, but applicable to any backend) with potentially dangerous uids.

jvillafanez:
Reverted the changes to the original ones due to overengineering

phil-davis:
@micbar @DeepDiver1975 @PVince81 @jvillafanez @VicDeo or anyone. Please review and advise if this is OK as-is or if there is more to do or another approach needed.

VicDeo (commented):
In general I'm fine with this approach.
As for me, making `$invalidUids` a protected class member is preferable.

jvillafanez (approved):
assuming we aren't missing any name

micbar (approved):

PR #37272, opened 2020-04-16
Use OcsController and routes instead of API::register and static methods

### LINKED ISSUE #34664: ocs config endpoint does not return CORS headers
### Steps to reproduce
1. set `http://owncloud.org` CORS domain
2. `curl -u admin:admin http://localhost/owncloud-core/ocs/v1.php/config -H "Origin: http://owncloud.org" -v`

### Expected behaviour
CORS headers returned

### Actual behaviour
no CORS headers but 
```
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Permitted-Cross-Domain-Policies: none
```

VicDeo:
## Description
- [x] Legacy classes removed
- [x] a separate file for ocs routes removed
- [x] ocs routes are registered in core using `$application->registerRoutes` instead of `API::register`
- [x] these routes use non-static controller
- [x] use DI for this controller
- [x] proper unit tests

## Related Issue
Fixes #34664
Partially covers 
https://github.com/owncloud/core/issues/12454

## How Has This Been Tested?
with curl
```
* set a key
* test: curl http://login:passwd@oc/core/ocs/v1.php/privatedata/setattribute/testy/123  --data "value=foobar"
	 
* read keys
* test: curl http://login:passwd@oc/core/ocs/v1.php/privatedata/getattribute
          curl http://login:passwd@oc/core/ocs/v1.php/privatedata/getattribute/testy
          curl http://login:passwd@oc/core/ocs/v1.php/privatedata/getattribute/testy/123

* delete a key
* test: curl http://login:passwd@oc/core/ocs/v1.php/privatedata/deleteattribute/testy/123 --data "post=1"
```

#### Case 2:
```
php occ  user:setting admin core domains --value '["http:\/\/abcd.com"]'
curl -u admin:admin http://localhost/owncloud/ocs/v1.php/config -H "Origin: http://abcd.com" -v
```
should contain
`Access-Control-Allow-Origin: http://abcd.com`

## Types of changes
- [x] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added


jvillafanez:
It's difficult for me to justify having sql statements in a controller. I'd rather split the responsibilities in order to have one class focused on storing the info and the controller to build the response. Having an easy interface with the DB will also help to test the controller very intensively.

VicDeo:
IMO Working with DB in a controller that is declared as controller explicitly is less critical than using classes in **lib/private/OCS/** as controllers.

micbar:
@VicDeo This seems stalled.

Can we merge this for 10.6?

VicDeo:
@micbar From my POV: it works and fixes the issue.

@phil-davis what is the right way to proceed with `tests/acceptance/features/apiAuth/corsOc10Issue34664.feature` here. Should I just delete it?

DeepDiver1975 (dismissed):
As long as the CSRF/OCS_API_REQUEST issue is not yet sorted you I'd not move more OCS apis over to the ocs controllers ....

VicDeo:
Feel free to close then...

phil-davis:
> what is the right way to proceed with tests/acceptance/features/apiAuth/corsOc10Issue34664.feature here. Should I just delete it?

@VicDeo for tests, if you fix an issue then delete the bug-demo scenario. And find the "real expected behavior" scenario in the "real" feature file. In this case it should be in `cors.feature`. Remove the `skipOnOcV10` tag from it. If the bug is fixed then the "real expected behavior" scenario should pass. Adjust the scenario if needed - we wrote the "real expected behavior" scenarios as a best-effort at what the behaviour should be, but of course the final implementation of a fix might have design changes that affect the expected flow of a test scenario.

(I realise that this PR is on-hold anyway, but the tests could be sorted out anyway...)

VicDeo:
@phil-davis I updated `cors.feature` a while ago and the build was successful those days. 
But I observe a new file after this rebase that's why I was asking
https://github.com/owncloud/core/pull/37272/files#diff-0ea7c9f325c45ed41d432d94ee0b6d23

micbar:
@C0rby @IljaN Can you review and help merging together with the CORS Header Pr?

VicDeo:
Tests are passing. Waiting for  CORS Header Pr

phil-davis:
@micbar @VicDeo what is the status of this?

Which is the "CORS Header PR"?

mmattel:
Any update?

C0rby:
> @micbar @VicDeo what is the status of this?
> 
> Which is the "CORS Header PR"?

@micbar meant the CSRF Header PR which was this: https://github.com/owncloud/core/pull/38019
The PR is merged so this PR here could be merged too from my POV.

The only thing I would check is if the annotation `@NoCSRFRequired` is necessary here and if not, remove the annotations.

phil-davis (commented):
Acceptance test changes look good. Hopefully CI will agree.
@DeepDiver1975 or some dev, please review the code again.

jvillafanez:
In addition to Corby's comment, I think we should include a comment somewhere stating that the code in the controller has been moved from a different place.

Honestly, if I hadn't any context, I'd say the code is quite bad to the point of make me wonder why this code exists. A comment in the controller would help to understand that we've just moved the code (barely touching it) and we opted to keep the backwards-compatibility. This hint should be enough to check the history if someone is interested.

VicDeo:
@C0rby 
> The only thing I would check is if the annotation @NoCSRFRequired is necessary here and if not, remove the annotations.

This will fail existing acceptance tests as none of these endpoints  checked CSRF token in past

VicDeo:
@C0rby @DeepDiver1975 I've  got your concerns after the checking API auth flow.
What if I check HTTP_OCS_APIREQUEST header presence directly in the controller?

C0rby:
I just saw in the description that there is a CORS header issue. So is there still work to do? @DeepDiver1975 

> @C0rby @DeepDiver1975 I've got your concerns after the checking API auth flow.
> What if I check HTTP_OCS_APIREQUEST header presence directly in the controller actions to keep the previous behavior?

The `HTTP_OCS_APIREQUEST` header shouldn't be checked anymore.
It was removed sometime ago and isn't sent by all clients. https://github.com/owncloud/core/pull/38019 is providing a CSRF check which doesn't require the header `HTTP_OCS_APIREQUEST` header.

VicDeo:
@C0rby 
> It was removed sometime ago and isn't sent by all clients

not completely ;)
it is still in use https://github.com/owncloud/core/blob/master/lib/private/legacy/api.php#L378

Here I'm migrating multiple legacy API endpoints into a single controller (and remove dependence  on prehistoric OC_API class)
The only intention for me was to fix  https://github.com/owncloud/core/issues/34664 - legacy API does NOT send CORS with the **response**. Now it does send CORS headers

jvillafanez:
Anything left to do here? I'm ok approving the current changes

VicDeo:
These endpoints don't need HTTP_OCS_APIREQUEST header anymore but require CSRF token instead.
On the one hand I consider it to be a breaking change on the other hand I wonder does someone still use this API.

I guess some acceptance tests will fail.

C0rby:
The CSRF token is only checked on requests using cookie authorization.

VicDeo:
Let's merge it finally

jvillafanez (approved):

PR #37277, opened 2020-04-19
Adjust expire trash background job interval

mrow4a:
Adjusting to the same interval as [ScanFiles](https://github.com/owncloud/core/blob/v10.4.1/apps/files/lib/BackgroundJob/ScanFiles.php). Due to https://github.com/owncloud/core/pull/36602 job runs in chunks, so interval needs adjustment too.


jvillafanez (approved):
I assume 10 minutes is enough for the job to finish

PR #37287, opened 2020-04-21
[Tests-Only] Updated phpDoc

kiranparajuli589:
## Description
In https://drone.owncloud.com/owncloud/files_external_dropbox/568/2/3, phan returned error as:
```bash
lib/Backend/Dropbox.php:46 PhanTypeMismatchArgument Argument 1 ($custom) is ['files_external_dropbox','dropbox'] of type array{0:'files_external_dropbox',1:'dropbox'} but \OCP\Files\External\Backend\Backend::addCustomJs() takes string defined at ../../lib/private/Files/External/FrontendDefinitionTrait.php:97
```
- That's why phpDoc is updated for method `addCustomJs`

## Related Issue
- Part of https://github.com/owncloud/QA/issues/643

## Motivation and Context
https://github.com/owncloud/files_external_dropbox/pull/83

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37289, opened 2020-04-21
[Tests-Only] Updated phpDoc for get/set customJs

kiranparajuli589:
## Description
Fix phpDoc of customJs getter setter

## Related Issue

- https://github.com/owncloud/QA/issues/643

## Motivation and Context
https://github.com/owncloud/core/pull/37287#issuecomment-616995150

## How Has This Been Tested?

- :robot: 
## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/External/FrontendDefinitionTrait.php
+++ b/lib/private/Files/External/FrontendDefinitionTrait.php
@@ -84,7 +84,7 @@ trait FrontendDefinitionTrait {
 	}
 
 	/**
-	 * @return string[]
+	 * @return string[]|array
 	 */
 	public function getCustomJs() {
 		return $this->customJs;
@@ -100,7 +100,7 @@ trait FrontendDefinitionTrait {
 	}
 
 	/**
-	 * @param string $custom
+	 * @param string|array $custom
 	 * @return self
 	 * @deprecated 9.1.0, use addCustomJs() instead
 	 */


kiranparajuli589:
@phil-davis `setCustomJs()` is `depricated` and `unsed` too. (instead `addCustomJs()`) is used!
Is it necessary to change `phpDoc` of an unsed method??

lib/private/Files/External/FrontendDefinitionTrait.php
```diff
@@ -84,7 +84,7 @@ public function addParameter(DefinitionParameter $parameter) {
 	}
 
 	/**
-	 * @return string[]
+	 * @return string[]|array
```
phil-davis, line 87: I wonder what is the "correct" way to specify this?
Maybe:
```suggestion
	 * @return string[]|string[][]
```
IMO it can be an "array of strings" or "an array of arrays of strings"

But maybe what is here is enough.

phil-davis (approved):

phil-davis:
> @phil-davis `setCustomJs()` is `depracated` and `unused` too. (instead `addCustomJs()`) is used!
> Is it necessary to change `phpDoc` of an unused method??

We don't really need to make an epic project out of fixing these - but we may as well fix what we notice when `phan` or `phstan` notices it. The changes in this PR look reasonable.

lib/private/Files/External/FrontendDefinitionTrait.php
kiranparajuli589: both are same! IMO array is more readable. Whay do you think?

PR #37310, opened 2020-04-26
[Tests-Only] Use assertContainsEquals to compare status codes that could be in string or int form

phil-davis:
## Description
In phpunit9 and the associated `Assert` class, `assertContains` is going to do a strict comparison, including types (e.g. int or string). When comparing HTTP or OCS status codes we do not care if the actual and expected codes arrive as int or string - we want a loose comparison so that we do not have to constantly be fussy about exactly what is being passed around.

Make these checks use `assertContainsEquals` which is a loose comparison.

In phpunit9 `assertEquals` on 2 arrays will start doing an exact comparison (the array order has to be the same as well as the key values and labels), and the extra parameters that used to be available are being removed. `assertEqualsCanonicalizing` is the new method to use for when we just care about the keys and values (and the order does not matter), which is normally the case. So those are fixed.

In phpunit9 `assertArraySubset` is removed. Replace the uses of it in acceptance test code with code to specifically check the expected array keys.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -1221,7 +1221,7 @@ class FeatureContext extends BehatVariablesContext {
 				$message = "HTTP status code $actualStatusCode is not one of the expected values";
 			}
 
-			Assert::assertContains(
+			Assert::assertContainsEquals(
 				$actualStatusCode, $expectedStatusCode,
 				$message
 			);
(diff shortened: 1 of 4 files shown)


individual-it (approved):

individual-it:
please write also an explanation of the `assertEqualsCanonicalizing` changes

PR #37322, opened 2020-04-29
this->l holds wrong app and translations cannot be found

### LINKED ISSUE #37321: Subject of mail when sharing files via link does not get translated to german
### Steps to reproduce
1. Set default_language to de_DE in the config.php.
2. Click on the sharing button of a file or folder.
3. Create a public link.
4. Share it with someone over mail with the designated button.

### Expected behaviour
The subject "X has shared Y with you" should be in german (default_language).

### Actual behaviour
The subject "X has shared Y with you" is not getting translated to german. 

This is because in **lib/private/Share/MailNotifications.php:257**, the string that should get translated is "%s shared »%s« with you" - which seems correct. 

However, in _apps/files_sharing/l10n/de_DE.json_ there is only the string "%2$s shared %1$s with you" (https://github.com/owncloud/core/blob/master/apps/files_sharing/l10n/de_DE.json#L64) - but not the correct one:
(https://github.com/owncloud/core/blob/master/lib/private/Share/MailNotifications.php#L257). 

To fix this, maybe add the string "%s shared »%s« with you" to the de_DE.json file?

DeepDiver1975:
## Description
Sharing notification email subject is not translated...
... because the l10n object from the share app notifications controller is injected and no translations are found.

Other usage locations of $this->l need to be verified .... follow up PR ...

## Related Issue
- Fixes #37321 

## How Has This Been Tested?
- :hand: 
- follow steps in #37321 

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37321
+++ b/changelog/unreleased/37321
@@ -1,0 +1,4 @@
+Bugfix: Translate public link sharing email subject
+
+https://github.com/owncloud/core/issues/37321
+https://github.com/owncloud/core/pull/37322
--- a/lib/private/Share/MailNotifications.php
+++ b/lib/private/Share/MailNotifications.php
@@ -244,11 +244,7 @@ class MailNotifications {
 		} catch (ShareNotFound $e) {
 			return $recipients;
 		}
-		if ($notificationLang !== null) {
-			$l10n = \OC::$server->getL10N('lib', $notificationLang);
-		} else {
-			$l10n = $this->l;
-		}
+		$l10n = \OC::$server->getL10N('lib', $notificationLang);
 		$expirationDate = null;
 		if ($share->getExpirationDate()) {
 			$expirationDate = $share->getExpirationDate()->getTimestamp();


changelog/unreleased/37321
```diff
@@ -0,0 +1,4 @@
+Bugfix: Translate public link sharing email subject
+
+https://github.com/owncloud/core/pull/37321
```
ho4ho, line 3:
```suggestion
https://github.com/owncloud/core/issues/37321
```

DeepDiver1975: THX

jvillafanez:
I'm not sure where is the problem... each app should take care of its own translations, and I really doubt they need to look for translations in other apps.

Injecting one `l10n` and then using a different one gotten from the server seems wrong. From my point of view, you either inject the right `l10n` instance, or make sure the translation is available in the injected instance.

VicDeo:
@jvillafanez injected locale is always preferred current user locale but some notifications should be sent using the lang of the recipient.

jvillafanez:
You can solve that problem either by injecting a "multiL10n" (something like `$ml->t('Translate this', 'en_BR')`) to switch the language on-the-fly, or by injecting a custom factory to get the translations for a specific language (`$l10n = $factory->getForLanguage('en_BR');  $l10n->t('Translate this');`)

We'd need to adjust the architecture and create the missing classes. It's possible, although I don't know if we want to invest time on this.

DeepDiver1975:
The problem is related to the fact that this method is called from a different app and the injected l10n is not for 'lib' but afaik files_sharing ....

DeepDiver1975:
call location: https://github.com/owncloud/core/blob/9b395f00bf14001ade592f9813c0d3fbb186221b/apps/files_sharing/lib/Controller/NotificationController.php#L86
 injected here: https://github.com/owncloud/core/blob/9b395f00bf14001ade592f9813c0d3fbb186221b/apps/files_sharing/lib/Controller/NotificationController.php#L54 

@jvillafanez

jvillafanez:
> The problem is related to the fact that this method is called from a different app and the injected l10n is not for 'lib' but afaik files_sharing ....

Does this mean that any injected `IL10N` instance in core is compromised because it's likely the app's instance being injected instead of the core's instance? If so, I think we need to investigate this issue carefully and propose a better solution.

I'll approve this as hotfix, but as said, we need a deep investigation on this.

jvillafanez (approved):

DeepDiver1975:
> Does this mean that any injected `IL10N` instance in core is compromised because it's likely the app's instance being injected instead of the core's instance? If so, I think we need to investigate this issue carefully and propose a better solution.

To some extend yes - but we did not experience any other issue in this area .... so we are just lucky ;-)

PR #37325, opened 2020-04-30
Unskip issue-ocis-reva-62

### LINKED ISSUE #62: PROPFIND on the root of the /webdav endpoints returns ERROR 500
```
curl -X PROPFIND http://localhost:9140/remote.php/webdav -u user1:1234 -v
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 9140 (#0)
* Server auth using Basic with user 'user1'
> PROPFIND /remote.php/webdav HTTP/1.1
> Host: localhost:9140
> Authorization: Basic dXNlcjE6MTIzNA==
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 500 Internal Server Error

PVince81:
Doing a PROPFIND on the Webdav root now returns the correct value.

Fixes https://github.com/owncloud/ocis-reva/issues/62

### DIFF
--- a/tests/acceptance/features/apiAuth/webDavAuth.feature
+++ b/tests/acceptance/features/apiAuth/webDavAuth.feature
@@ -9,7 +9,7 @@ Feature: auth
     When a user requests "/remote.php/webdav" with "PROPFIND" and no authentication
     Then the HTTP status code should be "401"
 
-  @smokeTest @issue-ocis-reva-62
+  @smokeTest
   Scenario: using WebDAV with basic auth
     When user "user0" requests "/remote.php/webdav" with "PROPFIND" using basic auth
     Then the HTTP status code should be "207"


phil-davis (approved):

individual-it:
that change did not unskip anything. The test did also run before, the change in this PR only removed the tag. We usually keep that for future reference

phil-davis:
I didn't look closely enough. I guess we could revert this?
If the test was already running in ocis-reva then why did @PVince81 feel the need to adjust something in core? It should have already been running and (hopefully) passing?

PVince81:
Right, I didn't see that there is no "@skipOnOcis" there but the original ticket was still open. I was tired and confused last week. Sorry.

Should I revert ?

individual-it:
@PVince81 lets just leave it, probably not worth burning CI time

Issue #33596, opened 2018-11-20
Add acceptance tests for more variety of user Ids


phil-davis:
Current acceptance tests user user IDs like ``user0``, ``user1``,... There are some tests that verify that user IDs with non-alpha-numeric characters can be created (or not created). But we do not have much verification that user IDs containing those non-alpha-numeric characters actually work correctly in a range of situations.

- [ ] run some or all acceptance tests with "interesting" user IDs
- [ ] run some or all acceptance tests with user IDs that look like an email address, and with the user's email address set to the same string, and to a different string

We have already "abstracted out" the "standard passwords" used in the acceptance tests - so that the passwords used by the tests can be passed in from environment variables. That allows the acceptance test suite to be run with a different set of interesting/varied passwords if desired.

We could do the same with the user IDs. Then we can easily run the acceptance test suite with user IDs like ``123.Me.test_user@test-organization.org`` (which uses lots of the valid special characters for a user ID).

Note: the webUI message when specifying an invalid user ID says:
```
Error creating user: Only the following characters are allowed in a username: "a-z", "A-Z", "0-9", and "_.@-'"
```
That is a clue to what should be valid and tested.

Note: since core 10.4 `+` is also allowed in a username.

phil-davis:
Related issue that triggered me to write this acceptance test issue: #33594

ownclouders:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/owncloud/core/issues/22485 (test), https://github.com/owncloud/core/issues/22492 (test), https://github.com/owncloud/core/issues/31579 (test), https://github.com/owncloud/core/pull/32347 (Add getters for private vars of acceptance test traits), and https://github.com/owncloud/core/pull/33506 (Adding more federation acceptance tests).

phil-davis:
- [ ] sort out if a space character should be allowed in a user ID
- [ ] add acceptance tests to verify that a space character is allowed and works (or not)

https://github.com/owncloud/core/issues/33594#issuecomment-440308620

@PVince81 @DeepDiver1975 @patrickjahns any idea if a user ID like ``John Smith`` should be valid?

phil-davis:
Added to upcoming sprint just for the purpose of adding basic validation tests for whatever should be the allowed and disallowed user ID strings, when attempts are made to create such user IDs with each of
- [ ] occ command
- [ ] Provisioning API
- [ ] webUI user management page

[...]

PR #37332, opened 2020-05-04
[Tests-Only] Acceptance tests for user with unusual names provided as env variable

jasson99:
## Description
Acceptance tests for user with unusual names provided as env variable

## Related Issue
https://github.com/owncloud/core/issues/33596

## How Has This Been Tested?
CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -73,6 +73,17 @@ class FeatureContext extends BehatVariablesContext {
 	private $originalAdminPassword = '';
 
 	/**
+	 * An array of values of replacement values of user attributes.
+	 * These are only referenced when creating a user. After that, the
+	 * run-time values are maintained and referenced in the $createUsers array.
+	 *
+	 * Key is the username, value is an array of user attributes
+	 *
+	 * @var array|null
+	 */
+	private $userReplacements = null;
+
+	/**
 	 * @var string
 	 */
 	private $regularUserPassword = '';
@@ -381,6 +392,46 @@ class FeatureContext extends BehatVariablesContext {
 	 */
 	public function isTestingWithLdap() {
 		return (\getenv("TEST_EXTERNAL_USER_BACKENDS") === "true");
+	}
+
+	/**
+	 * @return bool
+	 */
+	public function isTestingReplacingUsernames() {
+		return (\getenv('REPLACE_USERNAMES') === "true");
+	}
+
+	/**
+	 * @return array|null
+	 */
+	public function usersToBeReplaced() {
+		if (($this->userReplacements === null) && $this->isTestingReplacingUsernames()) {
+			$this->userReplacements = \json_decode(
+				\file_get_contents("./tests/acceptance/usernames.json"),
+				true
+			);
+			// Loop through the user replacements, and make entries for the lower
+			// and upper case forms. This allows for steps that specifically
+			// want to test that user names like "alice", "Alice" and "ALICE" all work.
+			// Such steps will make useful replacements for each form.
+			foreach ($this->userReplacements as $key => $value) {
+				$lowerKey = \strtolower($key);
+				if ($lowerKey !== $key) {
+					$this->userReplacements[$lowerKey] = $value;
+					$this->userReplacements[$lowerKey]['username'] = \strtolower(
+						$this->userReplacements[$lowerKey]['username']
+					);
+				}
+				$upperKey = \strtoupper($key);
+				if ($upperKey !== $key) {
+					$this->userReplacements[$upperKey] = $value;
+					$this->userReplacements[$upperKey]['username'] = \strtoupper(
+						$this->userReplacements[$upperKey]['username']
+					);
+				}
+			}
+		}
+		return $this->userReplacements;
 	}
 
 	/**
@@ -1113,6 +1164,7 @@ class FeatureContext extends BehatVariablesContext {
 	 * @return void
 	 */
 	public function userSendsHTTPMethodToUrl($user, $verb, $url) {
+		$user = $this->getActualUsername($user);
 		$this->sendingToWithDirectUrl($user, $verb, $url, null);
 	}
 
@@ -1435,6 +1487,7 @@ class FeatureContext extends BehatVariablesContext {
 	 * @return void
 	 */
 	public function userHasLoggedInToAWebStyleSessionUsingTheAPI($user) {
+		$user = $this->getActualUsername($user);
 		$loginUrl = $this->getBaseUrl() . '/login';
 		// Request a new session and extract CSRF token
 
@@ -1468,14 +1521,15 @@ class FeatureContext extends BehatVariablesContext {
 	}
 
 	/**
-	 * @When the client sends a :method to :url with requesttoken
+	 * @When the client sends a :method to :url of user :user with requesttoken
 	 *
 	 * @param string $method
 	 * @param string $url
+	 * @param string $user
 	 *
 	 * @return void
 	 */
-	public function sendingAToWithRequesttoken($method, $url) {
+	public function sendingAToWithRequesttoken($method, $url, $user) {
 		$headers = $this->guzzleClientHeaders;
 
 		$config = null;
@@ -1489,7 +1543,9 @@ class FeatureContext extends BehatVariablesContext {
 
 		$headers['requesttoken'] = $this->requestToken;
 
+		$user = \strtolower($this->getActualUsername($user));
 		$url = $this->getBaseUrl() . $url;
+		$url = $this->substituteInLineCodes($url, $user);
 		$this->response = HttpRequestHelper::sendRequest(
 			$url, $method, null, null, $headers, null, $config, $this->cookieJar
 		);
@@ -1509,14 +1565,15 @@ class FeatureContext extends BehatVariablesContext {
 	}
 
 	/**
-	 * @When the client sends a :method to :url without requesttoken
+	 * @When the client sends a :method to :url of user :user without requesttoken
 	 *
 	 * @param string $method
 	 * @param string $url
+	 * @param string $user
 	 *
 	 * @return void
 	 */
-	public function sendingAToWithoutRequesttoken($method, $url) {
+	public function sendingAToWithoutRequesttoken($method, $url, $user) {
 		$config = null;
 		if ($this->sourceIpAddress !== null) {
 			$config = [
@@ -1526,7 +1583,9 @@ class FeatureContext extends BehatVariablesContext {
 			];
 		}
 
+		$user = \strtolower($this->getActualUsername($user));
 		$url = $this->getBaseUrl() . $url;
+		$url = $this->substituteInLineCodes($url, $user);
 		$this->response = HttpRequestHelper::sendRequest(
 			$url, $method, null, null, $this->guzzleClientHeaders,
 			null, $config, $this->cookieJar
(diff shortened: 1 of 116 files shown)


tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,10 +1785,20 @@ public function getEmailAddressForUser($userName) {
...
+			} else {
+				return $functionalUsername;
+			}
 		} else {
-			return $functionalUsername;
+			if ($functionalUsername === "%admin%") {
+				return (string)$this->getAdminUsername();
+			} else {
+				return $functionalUsername;
+			}
```
phil-davis, line 1801: IMO we will need to keep this as a separate `if` block. 
If somebody is actually already using env variable `ADMIN_USERNAME` to override the admin usernamethen we need to keep doing that.

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,10 +1785,20 @@ public function getEmailAddressForUser($userName) {
 	 * @return string
 	 */
 	public function getActualUsername($functionalUsername) {
-		if ($functionalUsername === "%admin%") {
-			return (string) $this->getAdminUsername();
+		$usernameFromEnv = \getenv('Username');
```
phil-davis, line 1788: Naming is hard - the name of the env variable used as the "switch on" here needs to be longer and more unique so that it does not accidentally match something already in some environment that this is run in some day.

phil-davis (commented):
And we need to think about what to do with the current override of passwords code - do we allow that to be in the JSON file? So that there are not 2 completely separate ways that we override usernames and override passwords?

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: I changed the naming. But I am not sure if this is any better. Can you suggest one?

individual-it (commented):
what about a solution where that JSON data can also be in an env. variable?
then we don't need to write the file to prepare the test

jasson99:
@individual-it Keeping JSON data in a file as it now seems good to me rather than passing it as env variable every time we run the tests using that. Is passing as env variable a better option??

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,6 +1785,15 @@ public function getEmailAddressForUser($userName) {
 	 * @return string
 	 */
 	public function getActualUsername($functionalUsername) {
+		$usernameFromEnv = \getenv('REPLACE_WITH_UNUSUAL_USERNAMES');
+		if ($usernameFromEnv == true) {
+			$usernames = \json_decode(\file_get_contents("./usernames.json"), true);
+			if (isset($usernames[$functionalUsername])) {
+				return $usernames[$functionalUsername];
+			} else {
+				return $functionalUsername;
```
phil-davis, line 1794: If there is no replacement defined, then we do not want to return yet.
`$functionalUsername` might be `%admin%` and we want to check for that in the `if` below.
So, IMO, these 2 lines can be removed.

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,6 +1785,15 @@ public function getEmailAddressForUser($userName) {
 	 * @return string
 	 */
 	public function getActualUsername($functionalUsername) {
+		$usernameFromEnv = \getenv('REPLACE_WITH_UNUSUAL_USERNAMES');
```
phil-davis, line 1788:
```suggestion
		$usernameFromEnv = \getenv('REPLACE_USERNAMES');
```
Currently this is a boolean environment variable that "switches on" using replacement usernames. They might or might not be "unusual"

usernames.json
```diff
@@ -0,0 +1,8 @@
+{
```
phil-davis, line 1: We could have an example file like this configured. That will show people what they should edit to make a specific set of replacements. But it needs to be in `tests/acceptance` so that it does not confuse people.

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,6 +1785,15 @@ public function getEmailAddressForUser($userName) {
...
+		if ($usernameFromEnv == true) {
+			$usernames = \json_decode(\file_get_contents("./usernames.json"), true);
```
phil-davis, line 1790: This could be done once in the constructor of FeatureContext, and the replacement array would only need to be read in from file 1 time at the very start of a test and stored in a property of FeatureContext.

Then `getActualUsername` will just need the `isset` code below to check if there is a replacement defined. It will not need to read the JSON file every time it is called.

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

usernames.json
jasson99: fixed

.drone.star
```diff
@@ -1473,6 +1473,19 @@ def acceptance():
 												'%smake %s' % (suExecCommand, makeParameter)
 											]
 										}),
+										({
+										     'name': 'acceptance-tests',
+										     'image': 'owncloudci/php:%s' % phpVersion,
+										     'pull': 'always',
+										     'environment': {
+										                    'REPLACE_USERNAME': True,
```
phil-davis:
```suggestion
										                    'REPLACE_USERNAMES': True,
```

.drone.star
```diff
@@ -1466,7 +1466,10 @@ def acceptance():
 											'name': 'acceptance-tests',
 											'image': 'owncloudci/php:%s' % phpVersion,
 											'pull': 'always',
-											'environment': environment,
+											'environment': {
+											              environment,
+											              'REPLACE_USERNAME': True,
```
phil-davis, line 1471:
```suggestion
											              'REPLACE_USERNAMES': True,
```

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: I have made changes to it. Is it okay now?

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -373,6 +373,15 @@ public function isTestingWithLdap() {
 		return (\getenv("TEST_EXTERNAL_USER_BACKENDS") === "true");
 	}
 
+	/**
+	 * @return array|null
+	 */
+	public function usersToBeReplaced() {
+		if (\getenv('REPLACE_USERNAMES') !== false) {
+			return \json_decode(\file_get_contents("./tests/acceptance/usernames.json"), true);
+		}
```
individual-it, line 382:
```suggestion
		}
	return null;
```

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,6 +1795,12 @@ public function getEmailAddressForUser($userName) {
 	 * @return string
 	 */
 	public function getActualUsername($functionalUsername) {
+		$usernames = $this->usersToBeReplaced();
+		if (isset($usernames)) {
+			if (isset($usernames[$functionalUsername])) {
```
individual-it, line 1800: can be one IF statement I think

individual-it (dismissed):
generally looks good to me

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/TagsContext.php
```diff
@@ -1508,6 +1512,8 @@ public function fileHasNoTagsForUser($fileName, $adminOrUser = null, $user = nul
 	 * @return void
 	 */
 	private function untag($untaggingUser, $tagName, $fileName, $fileOwner) {
+		$untaggingUser = $this->featureContext->getActualUsername($untaggingUser);
+		$fileOwner = $this->featureContext->getActualUsername($fileOwner);
```
phil-davis, line 1516: These lines can be deleted. They are already below.

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1785,6 +1796,10 @@ public function getEmailAddressForUser($userName) {
 	 * @return string
 	 */
 	public function getActualUsername($functionalUsername) {
+		$usernames = $this->usersToBeReplaced();
+		if (isset($usernames, $usernames[$functionalUsername])) {
+			return $usernames[$functionalUsername];
+		}
```
phil-davis, line 1802:
```suggestion
		if (isset($usernames)) {
			if (isset($usernames[$functionalUsername])) {
				return $usernames[$functionalUsername];
			}
			$normalizedUsername = $this->normalizeUsername($functionalUsername);
			if (isset($usernames[$normalizedUsername])) {
				return $usernames[$normalizedUsername];
			}
		}
```
Some test scenarios check that different mixes of uppercase and lowercase in the username work - e.g. `user0` `User0` and `USER0` should be able to mixed up in API calls etc.

If somebody specifies an override for each of these, then we can replace each of them individually (e.g. `user0` =" `somebody` `User0` => `SomeBody` `USER0` => `SOMEBODY`). But if they do not want to do that, then we can automatically check and use the replacement for the normalized value of the username.

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1797,13 +1797,19 @@ public function getEmailAddressForUser($userName) {
...
+			}
+			$normalizedUsername = $this->normalizeUsername($functionalUsername);
+			if (isset($usernames[$normalizedUsername])) {
+				return $usernames[$normalizedUsername];
+			}
+			if ($functionalUsername === "%admin%") {
+				return (string) $this->getAdminUsername();
+			} else {
+				return $functionalUsername;
+			}
```
phil-davis, line 1812: This needs to stay outside of `if (isset($usernames))` - it needs to happen when there are no `usersToBeReplaced`

tests/acceptance/features/bootstrap/TagsContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

tests/acceptance/features/bootstrap/FeatureContext.php
jasson99: fixed

phil-davis:
@jasson99 something bad happened when rebasing - you now have 12 thousand files changed.

phil-davis:
@jasson99 @kiranparajuli589 I merged  various changes to feature files just now - PR #37377 - so you will need to rebase and resolve conflicts.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1523,7 +1523,7 @@ public function adminChangesTheDisplayNameOfUserUsingKey(
 				. $result->getStatusCode() . " " . $result->getBody()
 			);
 		}
-		$this->createdUsers[$user]["displayname"] = $displayname;
+		$this->createdUsers[$user]["displayname"] = $displayName;
```
phil-davis, line 1526: There is now function `rememberUserDisplayName` 
That should be used, because it checks for special cases, e.g. if the display name of `admin` is changed, that is not one of the `createdUsers` and the code does a bit different stuff.

And have a look at all the things that call `adminChangesTheDisplayNameOfUserUsingKey` - many of them also call `rememberUserDisplayName`, and maybe all those calls can be removed.

phil-davis:
- cherry-picked commits from #37466 (those are extra review changes)
- removed the env var definition of `REPLACE_USERNAMES` in `.drone.star` (we do not want to commit that)
- rebased and squashed it all

ToDo:
- verify that CI of #37466 passes (that is with `REPLACE_USERNAMES` defined)
- verify that CI passes here (without `REPLACE_USERNAMES` defined)
- verify that CI passes in OCIS and ocis-reva

phil-davis (approved):

phil-davis:
Passing - the fails are known because fsweb is currently down.
I will override and merge.

PR #37336, opened 2020-05-04
Fix sharing with uids in numerical format

### LINKED ISSUE #37324: Can't share with user that only has numbers as name
### Steps to reproduce
1. Create a user with a name like "12345" (just numbers)
2. Try to share a file with the user by searching their name and clicking on it

### Expected behaviour
The file should get shared with the user.

### Actual behaviour

The "popup" comes back again and the file won't get shared. A 500 error comes back from /ocs/v2.php/apps/files_sharing/api/v1/shares?format=json. However, there are no informations in the owncloud.log about this. 

### Server configuration
**Operating system**: CentOS 6.10

**Web server:** Apache

**Database:** MYSQL

**PHP version:** 7.2

**ownCloud version:** 10.4.1

**Updated from an older ownCloud or fresh install:** updated from older ownCloud

**Where did you install ownCloud from:** https://download.owncloud.org/community/owncloud-10.4.1.tar.bz2

**Signing status (ownCloud 9.0 and above):**

```
No errors have been found.
```

https://hastebin.com/givitefefo.json

**List of activated apps:**

```
Enabled:
  - activity: 2.5.3
  - comments: 0.3.0
  - configreport: 0.2.0
  - dav: 0.5.0
  - external: 1.4.0
  - federatedfilesharing: 0.5.0
  - federation: 0.1.0
  - files: 1.5.2
  - files_antivirus: 0.15.1
  - files_external: 0.7.1
  - files_mediaviewer: 1.0.2
  - files_pdfviewer: 0.11.1
  - files_sharing: 0.12.0
  - files_texteditor: 2.3.0
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - firstrunwizard: 1.2.0
  - gallery: 16.1.1
  - impersonate: 0.5.0
  - market: 0.5.0
  - notifications: 0.5.0
  - ##redacted##: 1.0.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - templateeditor: 0.4.0
  - updatenotification: 0.2.1
Disabled:
  - encryption
  - user_external
```

**Are you using external storage, if yes which one:** No

**Are you using encryption:** No

**Are you using an external user-backend, if yes which one:** No

karakayasemi:
## Description
Php does not allow numeric string as array key and converts these strings to integer. Share autocompletion api returns numerical uids as integer because of the desribed behavior of php. 

When autcomplete share api response directly used in ownCloud clients without type checking, it leds to problem when sharing with numerical uids. This PR fixes api response.

## Related Issue
- Fixes #37324

## How Has This Been Tested?
- Acceptance tests added.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/ShareesController.php
+++ b/apps/files_sharing/lib/Controller/ShareesController.php
@@ -189,7 +189,12 @@ class ShareesController extends OCSController {
 
 		$foundUserById = false;
 		$lowerSearch = \strtolower($search);
-		foreach ($users as $uid => $user) {
+		foreach ($users as $user) {
+			/**
+			 * Php parses numeric UID strings as integer in array key,
+			 * because of that, we need to learn uid from User object
+			 */
+			$uid = $user->getUID();
 			/* @var $user IUser */
 			$entry = [
 				'label' => $user->getDisplayName(),
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -404,7 +404,7 @@ class Share20OcsController extends OCSController {
 				$userAutoAccept = $this->config->getUserValue($shareWith, 'files_sharing', 'auto_accept_share', 'yes') === 'yes';
 			}
 			// Valid user is required to share
-			if ($shareWith === null || !$this->userManager->userExists($shareWith)) {
+			if (!\is_string($shareWith) || !$this->userManager->userExists($shareWith)) {
 				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
 				return new Result(null, 404, $this->l->t('Please specify a valid user'));
 			}
@@ -422,7 +422,7 @@ class Share20OcsController extends OCSController {
 			}
 
 			// Valid group is required to share
-			if ($shareWith === null || !$this->groupManager->groupExists($shareWith)) {
+			if (!\is_string($shareWith) || !$this->groupManager->groupExists($shareWith)) {
 				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
 				return new Result(null, 404, $this->l->t('Please specify a valid group'));
 			}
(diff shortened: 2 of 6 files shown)


apps/files_sharing/lib/Controller/ShareesController.php
```diff
@@ -195,7 +195,7 @@ protected function getUsers($search) {
 				'label' => $user->getDisplayName(),
 				'value' => [
 					'shareType' => Share::SHARE_TYPE_USER,
-					'shareWith' => $uid,
```
phil-davis, line 198: Note: at line 182 there is:
```
$users[$user->getUID()] = $user;
```
And `getUID()` does return a string. But when that string is numeric, e.g. "123456", then stupid PHP sets `$users[123456]`  rather than `$users["123456"]

And so the value of `$uid` here is the integer `123456` rather than the string "123456"

I hate this sort of stupidity in PHP.

phil-davis:
This fixes the webUI by always sending it string UID - good, the webUI client works.
But if some other client still makes a mess of string UIDs and sends back an API request to `shareWith` an integer then it still breaks:
```
curl --user jim:jim http://172.17.0.1:8080/ocs/v2.php/apps/files_sharing/api/v1/shares?format=json -d '{"shareType": 0, "shareWith": 123, "permissions": 31, "path": "/faaa"}' -X POST -H "Content-Type: application/json"
{"ocs":{"meta":{"status":"failure","statuscode":500,"message":"","totalitems":"","itemsperpage":""},"data":[]}}
```
IMO when also want to, when we receive an incoming API request, either:
1) "stringify" whatever is in `shareWith` and continue to process the share request, or;
2) If `shareWith` is not a string, then catch the exception and respond with some 4nn error code (that will let the API client know that their request is not valid, without giving them a 500)

apps/files_sharing/lib/Controller/ShareesController.php
jvillafanez: @karakayasemi include a comment in the code explaining the situation, otherwise I don't think I'll be the only one tempted to revert the change at some point.

> Note: at line 182 there is:

that should be fine, but I think we should include a comment to clarify the behaviour, mainly to make people aware of PHP doing weird things and forcing us to do weirder things.

karakayasemi:
@phil-davis according to `IShare` interface description, it is not throwing any exception. Since `Share20OcsController.php` has already request parameter validation, I extended these validations with is_string check instead of catching exception.

apps/files_sharing/lib/Controller/ShareesController.php
karakayasemi: I included a comment that explains the situation to the first commit.

phil-davis (approved):
LGTM

phil-davis:
I added skip tags to the changed acceptance test scenario, because that test scenario will not pass on old ownCloud releases.

PR #37343, opened 2020-05-05
Catch setSharedWith InvalidArgumentException

phil-davis:
## Description
POC on top of PR #37336 

This PR demonstrates how we could catch an `InvalidArgumentException` thrown by `setSharedWith` if someone does an API call like:

```
curl --user jim:jim http://172.17.0.1:8080/ocs/v2.php/apps/files_sharing/api/v1/shares?format=json -d '{"shareType": 0, "shareWith": 256, "permissions": 31, "path": "/faaa"}' -X POST -H "Content-Type: application/json"
{"ocs":{"meta":{"status":"failure","statuscode":403,"message":"sharedWith is integer but must be a string","totalitems":"","itemsperpage":""},"data":[]}}
```

So they get a 403 and error message instead of a 500 with no message.

## Related Issue
#37324 

## How Has This Been Tested?
Needs unit tests.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -408,7 +408,12 @@ class Share20OcsController extends OCSController {
 				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
 				return new Result(null, 404, $this->l->t('Please specify a valid user'));
 			}
-			$share->setSharedWith($shareWith);
+			try {
+				$share->setSharedWith($shareWith);
+			} catch (\InvalidArgumentException $e) {
+				$share->getNode()->unlock(ILockingProvider::LOCK_SHARED);
+				return new Result(null, 403, $e->getMessage());
+			}
 			$share->setPermissions($permissions);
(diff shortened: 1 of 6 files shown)


phil-davis:
@karakayasemi should we do something like this?
(in case someone makes API calls like this, so they get 403 instead of 500)

karakayasemi:
> 
> 
> @karakayasemi should we do something like this?
> (in case someone makes API calls like this, so they get 403 instead of 500)

Yes, this looks good. I don't know, why these exceptions are ignored before. 

We can continue with this PR. Need to add an explanatory comment to `ShareesController.php` change as @jvillafanez suggest and correcting changelog, or I can cherry-pick the commit and include it to https://github.com/owncloud/core/pull/37336

phil-davis:
@karakayasemi Please cherry-pick into your PR and add some unit tests etc.

phil-davis:
closing - this code will go into the parent PR

PR #37346, opened 2020-05-05
[Tests-Only] Example of using realistic usernames

phil-davis:
This is an example of using realistic usernames in acceptance test scenarios.

Instead of "user0" "user1" "user2" "user3" "user4", it uses "Alice" "Ben" "Carol" "David" "Emily".

The objective is to make test scenarios easier to read and comprehend.

I would like feedback about if using realistic usernames helps. The example feature file is https://github.com/owncloud/core/blob/ba1377ebe4de5e8cbb38c09a6e4f55db23ee9520/tests/acceptance/features/apiShareManagementBasic/createShare.feature

And would use realistic group names also - e.g. instead of `grp1` `grp2`... use `Finance` `Engineering`...

(if we implement a change like this is another thing - the bulk search-replace is easy, then finding the little things that were missed might be more painful)

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -2,31 +2,31 @@
 Feature: sharing
 
   Background:
-    Given user "user0" has been created with default attributes and without skeleton files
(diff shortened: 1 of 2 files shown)


mrow4a:
I think this is good idea.

phil-davis:
Implemented in #37397

PR #37351, opened 2020-05-06
[Tests-Only] Fix testCreateShareGroupNoValidShareWith issue 37350

### LINKED ISSUE #37350: testCreateShareGroupNoValidShareWith not doing what it claims
https://github.com/owncloud/core/blob/master/apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
```
	public function testCreateShareGroupNoValidShareWith() {
```

It looks like it should be testing a case of sharing with a group when that group does not exist. But the expected result is:
```
		$expected = new Result(null, 404, 'Please specify a valid user');
```
user???

And it passes, so the code being executed is actually going down the `SHARE_TYPE_USER` path.

IMO the expected result should be:
```
		$expected = new Result(null, 404, 'Please specify a valid group');
```

I noticed this while reviewing #37336

phil-davis:
## Description
Adjust the unit test so that it actually tests the group sharing case.

## Related Issue
- Fixes #37350 

## How Has This Been Tested?
Unit test run

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
+++ b/apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
@@ -888,8 +888,8 @@ class Share20OcsControllerTest extends TestCase {
 				->will($this->returnValueMap([
 						['path', null, 'valid-path'],
 						['permissions', null, \OCP\Constants::PERMISSION_ALL],
-						['shareType', $this->any(), Share::SHARE_TYPE_GROUP],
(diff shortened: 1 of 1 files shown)


karakayasemi (approved):
I have discovered similar behavior when writing unit test for remote share for same class. But, I could not catch this one.

PR #37359, opened 2020-05-07
[Tests-only] skip webdav post request tests on ocis

C0rby:
Skip one acceptance test scenario to make ocis-reva tests pass.
Specifically we skipped the webdavPost request test.
Opened an issue in ocis-reva https://github.com/owncloud/ocis-reva/issues/179


PVince81 (approved):
:+1:

PVince81:
@C0rby please add "[Tests-only]" in the ticket header when touching only the tests
also would be nice to post a link to the ocis-reva ticket you raised, makes it easier to find later on

PVince81:
merging this as the change doesn't affect OC 10 anyway

PR #37360, opened 2020-05-07
[Tests-Only] Refactor unit test calls

phil-davis:
## Description
In phpunit9 some things are changing. The new way of doing these things is already possible in phpunit8, so do it now:

- `assertContains` stops looking at keys, it just looks in values. Use `assertArrayHasKey` to look for a key existing.
- `contains` becomes `containsEquals` or `containsIdentical` - use the new ones.
- `assertContains` adds `assertContainsEquals` or `assertContainsIdentical`

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/lib/Share20/DefaultShareProviderTest.php
+++ b/tests/lib/Share20/DefaultShareProviderTest.php
@@ -1521,7 +1521,7 @@ class DefaultShareProviderTest extends TestCase {
 				$shareNode = $share->getNode();
 				$this->assertSame($node1, $shareNode);
 			} else {
-				$this->assertContains($share->getId(), $idGroups);
+				$this->assertContains((int) $share->getId(), $idGroups);
 				$this->assertStringStartsWith('group', $share->getSharedWith());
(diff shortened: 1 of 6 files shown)


tests/lib/Share20/DefaultShareProviderTest.php
```diff
@@ -1521,7 +1521,7 @@ public function testGetAllSharedGroupChunking($storageStringId, $fileName1, $fil
 				$shareNode = $share->getNode();
 				$this->assertSame($node1, $shareNode);
 			} else {
-				$this->assertContains($share->getId(), $idGroups);
+				$this->assertContains((int) $share->getId(), $idGroups);
```
phil-davis, line 1524: For this one, `getId` returns a string that is numeric, and the array has the int values in it. In phpunit9 it is more strict and does not find "123" in an array that has int `123`

individual-it (approved):

PR #37386, opened 2020-05-13
[Tests-only] Adjust ocis-reva-94 for versions

PVince81:
The versions are now properly deleted in OCIS after removing a file.

Ref https://github.com/owncloud/ocis-reva/issues/94

Unblocks https://github.com/owncloud/ocis-reva/pull/192


PVince81:
note: I ran the test locally and it passed fine with the code from https://github.com/owncloud/ocis-reva/pull/192

individual-it (approved):

PR #37394, opened 2020-05-15
Allow reading params of PUT with empty body more than once

VicDeo:
## Description
Allows read URL params of PUT request with an empty body more than once.
Missing or 0  Content-Length means that the request has an empty body and we should NOT throw  LogicException while reading it.

## How Has This Been Tested?
1. Create an OCS route with urlParam like this:
```
[
	'root' => '/cloud',
	'name' => 'Users#enableUser',
	'url' => '/users/{userId}/enable',
	'verb' => 'PUT'
],
```
2. Create a controller
```
use OCP\AppFramework\OCSController;

class UsersController extends OCSController {
/**
	 * @NoCSRFRequired
	 *
	 * @param string $userId
	 * @return Result
	 */
	public function enableUser($userId) {
		return $this->setEnabled($userId, true);
	}
```

2. Use it
```
curl -X PUT http://admin:admin@localhost/OC/ocs/v1.php/cloud/users/fatima/enable
```

### Expected
`$userId` is 'fatima' in the controller

### Actual
`$userId` is `null`

The log file contains
```
{"reqId":"08vb9aIDIHfQVAQp5ZB1","level":3,"time":"2020-05-15T09:19:24+00:00","remoteAddr":"127.0.0.1","user":"admin","app":"PHP","method":"PUT","url":"\/OC\/ocs\/v1.php\/cloud\/users\/user1\/disable","message":"LogicException: \"put\" can only be accessed once if not application\/x-www-form-urlencoded or application\/json. at 
 \/lib\/private\/AppFramework\/Http\/Request.php#395"}
```

OCS routes always get `format` param from URL and adding one more URL param  causes a logic exception  

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37394
+++ b/changelog/unreleased/37394
@@ -1,0 +1,7 @@
+Bugfix: Allow unlimited access to PUT body if content length is 0
+
+It was not possible to read more than one URL param of the PUT request with the empty body.
+This change checks Content-Length and do not throw the exception on empty request body if 
+Content-Length states that the empty body had been sent. 
+
+https://github.com/owncloud/core/pull/37394
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -388,6 +388,8 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 	protected function getContent() {
 		// If the content can't be parsed into an array then return a stream resource.
 		if ($this->method === 'PUT'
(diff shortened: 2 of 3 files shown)


jvillafanez (approved):
Code changes look fine, although I'm not sure about possible side effects.

PR #37398, opened 2020-05-17
Use BackendUserIterator with uid filter for user:sync --uid

mrow4a:
Commit https://github.com/owncloud/core/commit/dd8440595146b577e598d981610530cf1cb8c630 introduced a problem that when searching for a user in LDAP backend that does not fit in LDAP Paging Size made `user:sync --uid` break (essentially not finding it and attempting to remove

How tested?
- [x] unit tests
- [x] insert 1000 users, including `1_aaliyah_abernathy`, `aaliyah_abernathy` and `aaliyah_abernathy_1` **at the beginning**, searched for `aaliyah_abernathy` - without this fix it works, with this fix it also it works
- [x] insert 1000 users, including `1_abagail_zieme`, `abagail_zieme` and `abagail_zieme_1` **at the end**, searched for `abagail_zieme` - without this fix it wont work, with this fix it works (fixes https://github.com/owncloud/enterprise/issues/3981)

### DIFF
--- a/changelog/unreleased/37398
+++ b/changelog/unreleased/37398
@@ -1,0 +1,8 @@
+Bugfix: Adjust user:sync --uid to use user backend iterator
+
+It fixes the behavior for user:sync --uid that attempts to retrieve all
+user backend users without limit at offset, that is not supported by LDAP backend.
+Instead, proper iterator and search query has been used
+
+https://github.com/owncloud/core/pull/37398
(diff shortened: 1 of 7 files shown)


jvillafanez (approved):

phil-davis:
No codecov arrived. I also had this happen yesterday. Someone can decide if they want to override to merge, or if this should be rebased to try again.

Issue #37424, opened 2020-05-21
Cannot clear email address from provisioning API or occ command


phil-davis:
### Steps to reproduce
1. create a user account, e.g. `phil`
2. set an email address, e.g. with the provisioning API:
```
$ curl -X PUT http://phil:phil@172.17.0.1:8080/ocs/v2.php/cloud/users/phil -d key="email" -d value="test@example.org"
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>200</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
$ php occ user:list -a email
  - admin: 
  - phil: test@example.org

```
3. try to remove the email address:
```
curl -X PUT http://phil:phil@172.17.0.1:8080/ocs/v2.php/cloud/users/phil -d key="email" -d value=""
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>400</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

Similarly with occ commands:
```
$ php occ user:modify phil email "someone@example.org"
The email address of phil updated to someone@example.org
$ php occ user:list -a email
  - admin: 
  - phil: someone@example.org
$ php occ user:modify phil email ""
The value cannot be empty
```

### Expected behaviour
There should be some way to clear the email address using the Provisioning API  and an occ command.

### Actual behaviour
Errors as above when trying to specify the empty string as the email address.

### Server configuration
Current git core master on Ubuntu 18.04

Note:
1) In the webUI settings, personal, general you can clear and save an empty email address. That calls some other private endpoint.
2) For displayname it works - I can set displayname with e Provisioning API or the occ command, and I can reset it to the empty string.

phil-davis:
Note: old PR #27671 was related to similar stuff and got closed.
Maybe there is a way to allow clearing email addresses through the Provisioning API and occ command - because the webUI certainly does it successfully.

phil-davis:
PR #37425 has a solution for the `occ user:modify` command. That seems to work fine setting email or display name to the empty string.

phil-davis:
Fixed in #37425 (occ user:modify command) and #37427 (Provisioning API)

PR #37425, opened 2020-05-21
Allow setting an empty email address or display name with user:modify

phil-davis:
## Description
`occ user:modify <username> email ''`
`occ user:modify <username> displayname ''`

would report that the value cannot be empty.

Actually if the email address or display name gets set somehow and needs to be cleared, then there is no real reason why it should not be cleared (back to the "default" on user creation, which is no special display name and no email address).

The code is adjusted so that the empty string is allowed in these cases.

## Related Issue
- part of #37424 

## How Has This Been Tested?
Local `occ user:modify` commands.

CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37424-1
+++ b/changelog/unreleased/37424-1
@@ -1,0 +1,15 @@
+Bugfix: Allow clearing a user email address or display name
+
+The occ user:modify command would not allow the email or display name of a user
+to be cleared. This problem has been fixed.
+
+The email of a user can be cleared with:
+occ user:modify <username> email ''
+
+The display name of a user can be cleared with:
+occ user:modify <username> displayname ''
+
+And the effective display name reverts to the username.
+
+https://github.com/owncloud/core/issues/37424
+https://github.com/owncloud/core/pull/37425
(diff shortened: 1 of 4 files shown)


jvillafanez:
Technically, the value should be null in the DB instead of an empty string. However, setting a null value seems to be problematic and might require deeper changes.

Not sure if we're fine with this solution as a partial workaround. The code changes look good though.

phil-davis:
When the email address is cleared in the webUI settings-personal-general it sends a PUT to an endpoint like:

`http://172.17.0.1:8080/index.php/settings/users/phil/mailAddress`

with form data `mailAddress` set to the empty string.

That currently works. So if we do a similar thing here, we can't be any worse off.

phil-davis:
100% of patch changes are covered. Because of deleting some code, `codecov/project` average is slightly lower.

jvillafanez (approved):

PR #34158, opened 2019-01-16
add before-after share link auth events

karakayasemi:
## Description
Adds before and after events for public share link auths.

## Related Issue
The first step for:
#33542, https://github.com/owncloud/brute_force_protection/issues/57

## Motivation and Context
To be able to detect and interrupt public link share auth attempts.

## How Has This Been Tested?
Unit test and manually.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 

I am not sure about the location of the events. As far as I see, adding events in `linkShareAuth` function is enough. But, if you see any missing place, please let me know. 

When this pr merged, I will update here: https://github.com/owncloud/docs/issues/311#issuecomment-419377043

PR #37430, opened 2020-05-21
share manager should emit share.link auth events to catch all accesses

karakayasemi:
## Description
Currently, WebDAV access for the password-protected public links is not emitting any link share auth events. These events are only emitted when ShareController used.

This PR moves link share auth events emitting logic to share manager to cover all access cases.
## Related Issue
- Discovered in here: https://github.com/owncloud/brute_force_protection/pull/123#issuecomment-630781876
- Partly related https://github.com/owncloud/admin_audit/issues/229

## How Has This Been Tested?
- The problem described in here fixed with this pr: https://github.com/owncloud/brute_force_protection/pull/123#issuecomment-630781876

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Changelog item

### DIFF
--- a/apps/files_sharing/lib/Controllers/ShareController.php
+++ b/apps/files_sharing/lib/Controllers/ShareController.php
@@ -182,13 +182,10 @@ class ShareController extends Controller {
 	 * @return bool
 	 */
 	private function linkShareAuth(\OCP\Share\IShare $share, $password = null) {
-		$beforeEvent = new GenericEvent(null, ['shareObject' => $share]);
-		$this->eventDispatcher->dispatch('share.beforelinkauth', $beforeEvent);
 		if ($password !== null) {
 			if ($this->shareManager->checkPassword($share, $password)) {
 				$this->session->set('public_link_authenticated', (string)$share->getId());
 			} else {
-				$this->emitAccessShareHook($share, 403, 'Wrong password');
 				return false;
 			}
 		} else {
@@ -198,8 +195,6 @@ class ShareController extends Controller {
 				return false;
 			}
 		}
-		$afterEvent = new GenericEvent(null, ['shareObject' => $share]);
-		$this->eventDispatcher->dispatch('share.afterlinkauth', $afterEvent);
 		return true;
 	}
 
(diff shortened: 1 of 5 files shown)


phil-davis:
No codecov result had come. The branch was off a quite-old version of master.
I rebased and force-pushed just now. The codecov reporting problem was fixed recently in master, so this should get a codecov result from CI this time.

jvillafanez:
Question: how about triggering a new event? I'm concerned this PR might break things because the events are triggered in a different place, so it could happen that the event was triggered before an error happened, but this might not happen any longer.
In addition, having an error code of 403, probably referencing a http error code, doesn't make sense in the share manager. I know it's to keep compatibility, but this one problem of moving events after they're been established.

karakayasemi:
> 
> 
> Question: how about triggering a new event? I'm concerned this PR might break things because the events are triggered in a different place, so it could happen that the event was triggered before an error happened, but this might not happen any longer.

@jvillafanez before and after share link auth events was already newly added with this pr https://github.com/owncloud/core/pull/34158, for the purpose of using it in brute force protection app. So, by assuming there is no other usage of it in the organization-wide (not even brute force protection app yet), I can say there is no need to afraid to move them to a different class.

> In addition, having an error code of 403, probably referencing a http error code, doesn't make sense in the share manager. I know it's to keep compatibility, but this one problem of moving events after they're been established.


I assume this event is intended to emit on every access attempt of a share. However, currently, WebDAV accesses miss these events. I identify this situation as a bug. I agree with you on about the error code issue but could not find a better solution. I kept the event format the same for compatibility reasons as you said.

Since `ShareController` also using the Share Manager, the previously emitted events will still be emitted as identical with before in this PR's solution, so, I expect It should not break any current usage. In addition, it will fix the bug by emitting event signals in DAV accesses also.

Another solution would be adding the same events to dav app, but I think, unifying emitting logic in the share manager is a better solution.

jvillafanez:
I still think it's a better idea to deprecate the old events (leaving them for a while) and add new events to fix this, then the app can switch to the new events once they're in place. This way we can adjust the event name to be more intuitive (using this the "checkPassword" function doesn't mean that a link is being accessed, but a link, in this particular case, has its password checked), and adjust what the event is sending to be more meaningful.

karakayasemi:
I will open a new pr without touching existing events. Closing this for now.

PR #37436, opened 2020-05-22
Allow the get demo key link to be configurable

jvillafanez:
## Description
Allow to configure a link to get the demo key from the grace period popup

## Related Issue
https://github.com/owncloud/core/issues/37420

## How Has This Been Tested?
Manually tested as well as with unit tests

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1540,4 +1540,13 @@ $CONFIG = [
  */
 'dav.enable.async' => false,
 
+/**
+ * Link to get a demo key during active grace period
+ * The admin will be directed to that web page if he clicks in the "get a demo key"
+ * link in the grace period popup. It's expected that the web page contains instructions
+ * on how to get a valid demo key to be used in the ownCloud server.
+ * If this key isn't present, ownCloud's default will be used
+ */
+'grace_period.demo_key.link' => 'https://owncloud.com/try-enterprise/',
+
 ];
--- a/core/js/license-trial-notification.js
+++ b/core/js/license-trial-notification.js
@@ -51,6 +51,12 @@ OC.License = {
 				if (!data || data.apps < 1) {
 					return;
 				}
+
+				var link = 'https://owncloud.com/try-enterprise/';
+				if (data.demoKeyLink) {
+					link = data.demoKeyLink;
+				}
+
 				// adjust moment's time thresholds for better accuracy:
 				// it will show "24 hours" instead of "a day" and "89 minutes" instead of "2 hours"
 				// we'll have to restore the defaults afterwards.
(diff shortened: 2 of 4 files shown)


config/config.sample.php
```diff
@@ -1540,4 +1540,12 @@
  */
 'dav.enable.async' => false,
 
+/**
+ * The link the admin should use to get a demo key while the grace period is active.
```
mmattel, line 1544: The first line of the comment block is used to create a table of contents text. No ending dots please.
The text used here is quite long. Proposal:
`Link to get a demo key during active grace period`
You can put the proposal above your description because the description itself is ok.
See other examples in config sample.

Q: what happens if the link is not present or `<empty>` or contains a typo?
I tried your link and it gives me:
`The page you are looking for might have been removed, had its name changed, or is temporarily unavailable.` :cry:
Please use a valid link or type like `'https://example.com/try-enterprise/'`
Else you put a negative touch on the owncloud brand.

mmattel:
This contains a `config.sample` change!
You MUST file a documetation issue as the changes get reflected in docs and there needs some work to be done when merged here in core!

config/config.sample.php
jvillafanez: The default link will be available soon :tm: , before 10.5.0 is officially out. That was the one ownCloud requested.

> what happens if the link is not present or <empty> or contains a typo?

If not present (which should happen on the majority of the installations), the default one will be used. If present, it's the admin's (or the one who deployed ownCloud) responsibility to provide a valid link

micbar (approved):

config/config.sample.php
mmattel: Thanks for clarification and fixing the headline.
Pls, when changes are implemented pointing to a currently non existant page, add a note that the page link is intended to be so and the page is being created before release 😄

phil-davis:
What should happen about a changelog here?
Maybe add this PR in the references of the changelog of #36814 ?
https://github.com/owncloud/core/blob/master/changelog/unreleased/36814

phil-davis:
@micbar @jnweiger @HanaGemela I guess this is supposed to get into 10.5, but nobody pressed the merge button yet???

davitol:
@jvillafanez I have test this PR and it looks like it does not work for me

## Test steps

1. In config.php file set parameter: `'grace_period.demo_key.link' => 'https://f1.com',`
2. In grace period popup, click  on 'Get your demo key' buttom

## Expected result

A tab with https://f1.com is opened

## Current behavior

A tab with https://owncloud.com/try-enterprise/  is opened

@jvillafanez JFYI

davitol:
@jvillafanez Tested with latest code and works for me

PR #37437, opened 2020-05-22
[Tests-only] Use valid Depth values in tests

PVince81:
Added validation to prevent using invalid depths and causing failure
against OCIS servers.
Fixes occurrences of invalid depth values and used infinity instead.

Context: https://github.com/owncloud/ocis-reva/pull/221#issuecomment-632592440

### DIFF
--- a/tests/TestHelpers/WebDavHelper.php
+++ b/tests/TestHelpers/WebDavHelper.php
@@ -132,6 +132,10 @@ class WebDavHelper {
 			}
 			$propertyBody .= "<$namespacePrefix:$property/>";
 		}
+		$folderDepth = (string) $folderDepth;
+		if ($folderDepth !== '0' && $folderDepth !== '1' && $folderDepth !== 'infinity') {
+			throw new InvalidArgumentException('Invalid depth value ' . $folderDepth);
+		}
 		$headers = ['Depth' => $folderDepth];
 		$body = "<?xml version=\"1.0\"?>
 				<d:propfind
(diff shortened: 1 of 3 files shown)


PVince81:
hah... `Invalid depth value systemtags (InvalidArgumentException)`
shifted args

tests/TestHelpers/WebDavHelper.php
```diff
@@ -132,6 +132,10 @@ public static function propfind(
 			}
 			$propertyBody .= "<$namespacePrefix:$property/>";
 		}
+		$folderDepth = "" . $folderDepth;
```
phil-davis, line 135: I guess this is equivalent to code like:
```suggestion
		$folderDepth = (string) $folderDepth;
```

PVince81: right, that the JS mindset I used here...

PVince81:
fixed the tags depth arg

tests/acceptance/features/bootstrap/TagsContext.php
```diff
@@ -1142,7 +1142,7 @@ private function requestTagsForFile($user, $fileName, $sharingUser = null) {
 			$this->featureContext->getBaseUrl(),
 			$user,
 			$this->featureContext->getPasswordForUser($user),
-			$fullPath, $properties, 'systemtags',
+			$fullPath, $properties, 0, 'systemtags',
```
phil-davis, line 1145: Well spotted. It looks like this has been wrong for a long time. By good luck the remaining parameters being shuffled down 1 place did not have much real difference (at least in the oC10 case)

phil-davis (approved):

PVince81:
for some reason depth 0 did not work for the tags test, so I've set it to 1 now (it passes locally).

PR #37438, opened 2020-05-22
emit events for before-after-failed link password checks

karakayasemi:
## Description
This PR adds 3 new events (before-fail-after) for link password validations. 

## Related Issue
- Discovered in here: https://github.com/owncloud/brute_force_protection/pull/123#issuecomment-630781876
- Partly related https://github.com/owncloud/admin_audit/issues/229
- Continuation of https://github.com/owncloud/core/pull/37430

## How Has This Been Tested?
- The problem described in here fixed with this pr: https://github.com/owncloud/brute_force_protection/pull/123#issuecomment-630781876

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Changelog item

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -1474,6 +1474,10 @@ class Manager implements IManager {
 
 	/**
 	 * Verify the password of a public share
+	 * Dispatches following events:
+	 * 'share.beforepasswordcheck' is dispatched before every password verification
+	 * 'share.afterpasswordcheck' is dispatched after successful password verifications
+	 * 'share.failedpasswordcheck' is dispatched after unsuccessful password verifications
 	 *
 	 * @param \OCP\Share\IShare $share
 	 * @param string $password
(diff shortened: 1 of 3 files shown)


mmattel:
Does this need a doc change/addon in the developer section ?
If so, pls file a issue in docs.

karakayasemi:
@mmattel the Symfony events never documented, there is an open issue about it but it seems inactive for a long time. See https://github.com/owncloud/docs/issues/311

I will update the mentioned documentation issue with these events after the merge anyway. Thank you for reminding me.

lib/private/Share20/Manager.php
```diff
@@ -1488,12 +1488,16 @@ public function checkPassword(\OCP\Share\IShare $share, $password) {
 		if ($password === null || $share->getPassword() === null) {
 			return false;
 		}
-
+		$beforeEvent = new GenericEvent(null, ['shareObject' => $share]);
+		$this->eventDispatcher->dispatch('share.beforelinkpasswordcheck', $beforeEvent);
```
jvillafanez, line 1492: I'd remove the "link" reference in the name: "share.beforepasswordcheck". It's unlikely that we remove the "only for links" restriction, but the method name doesn't refer to links explicitly.

tests/lib/Share20/ManagerTest.php
```diff
@@ -3151,17 +3151,50 @@ public function testCheckPasswordInvalidPassword() {
 
 		$this->hasher->method('verify')->with('invalidpassword', 'password', '')->willReturn(false);
 
+		$calledBeforeEvent = [];
+		$this->eventDispatcher->addListener('share.beforelinkpasswordcheck',
```
jvillafanez, line 3155: Just for the record (this was probably there already and changing it will take time), it seems better to use an interface and ensure that the "dispatch" method is called with the right parameters; no need to rely on an actual implementation.
As said, no need to make this adjustment this now.

jvillafanez:
Should we include a PHPDoc to clarify when the events will be thrown? It would help with the expectations in case something changes in the future.

lib/private/Share20/Manager.php
karakayasemi: changed as suggested.

jvillafanez (approved):

phil-davis:
Looks good. @karakayasemi I guess you will adjust brute_force_protection so that it will receive and respond to these?

karakayasemi:
> 
> 
> Looks good. @karakayasemi I guess you will adjust brute_force_protection so that it will receive and respond to these?

Yes, I expect brute_force_protection will be aware of all share password validations now.

PR #37449, opened 2020-05-26
here be beta1

jnweiger:
tag and version.php update


HanaGemela (approved):

phil-davis:
This branch does not seem to be off current master, so it misses things that were merged today.
@jnweiger Was that intended?

PR #37454, opened 2020-05-27
Fix logging when logger has no writeExtra method

### LINKED ISSUE #37453: Error when logging with a logger that has no writeExtra method
### Steps to reproduce
1. use some logger that does not implement a `writeExtra` method
2. have events happen that provide extra fields that could be logged

### Expected behaviour
The extra fields are ignored, and the basic log message/data is logged.

### Actual behaviour
call_user_func() expects parameter 1 to be a valid callback, class 'OC\Log\Syslog' does not have a method 'writeExtra' at /var/www/html/owncloud/lib/private/Log.php#398

### Server configuration
Current GitHub master, or 10.4.1 (or any "recent" release, I expect)

phil-davis:
## Description
- Correct the call to `\is_callable` - the first parameter should be an array whose entries are the object concerned and the name of the hoped-for method in the object.
- Add a unit test

## Related Issue
- Fixes #37453 

## How Has This Been Tested?
`LoggerWithoutWriteExtraTest.php` is a cut-down version of `LoggerTest.php` that does not have a `writeExtra` method, and just has the relevant test case.

Without the fix in ` lib/private/Log.php` the test fails:
```
PHPUnit 8.5.5 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.4.6 with Xdebug 2.9.5
Configuration: /home/phil/git/owncloud/core/tests/phpunit-autotest.xml

E                                                                   1 / 1 (100%)

Time: 110 ms, Memory: 24.00 MB

There was 1 error:

1) Test\LoggerWithoutWriteExtraTest::testExtraFields
call_user_func() expects parameter 1 to be a valid callback, class 'Test\LoggerWithoutWriteExtraTest' does not have a method 'writeExtra'

/home/phil/git/owncloud/core/lib/private/Log.php:398
/home/phil/git/owncloud/core/lib/private/Log.php:247
/home/phil/git/owncloud/core/tests/lib/LoggerWithoutWriteExtraTest.php:59

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
Makefile:189: recipe for target 'test-php-unit' failed
make: *** [test-php-unit] Error 2
```

That reproduces the error reported in the issue.

With the fix, the test passes.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37453
+++ b/changelog/unreleased/37453
@@ -1,0 +1,10 @@
+Bugfix: Logging of extra fields when logger does not have a writeExtra method
+
+If a logger in use does not have a writeExtra method then an error message would
+be generated when a log entry with extra data happens.
+
+This problem has been corrected. In this case the basic log information will be
+written without the extra data.
+
+https://github.com/owncloud/core/issues/37453
+https://github.com/owncloud/core/pull/37454
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -394,7 +394,7 @@ class Log implements ILogger {
 		if ($level >= $minLevel) {
 			$logger = $this->logger;
 			// check if logger supports extra fields
-			if (!empty($extraFields) && \is_callable($logger, 'writeExtra')) {
+			if (!empty($extraFields) && \is_callable([$logger, 'writeExtra'])) {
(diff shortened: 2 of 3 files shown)


micbar (approved):
Thanks a lot!

PR #37455, opened 2020-05-27
config.php contents won't be overwritten by other files

jvillafanez:
## Description
Ensure config.php takes priority over other config files

## Related Issue
https://github.com/owncloud/enterprise/issues/3775#issuecomment-633996000

## Motivation and Context
If there are multiple configuration files, changing the config.php file doesn't guarantee the change will take effect.

## How Has This Been Tested?
Checked manually. The contents of the config.php takes priority over other config files that could be in the config directory.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37455
+++ b/changelog/unreleased/37455
@@ -1,0 +1,11 @@
+Change: Prioritize config.php file over other configurations
+
+Previously, in case there were several configuration files, all those files
+were merged into one. If there were duplicated configuration keys, the values
+were overwritten.
+
+Now, the files are merged the same way, but if there are duplicated configuration
+keys, they won't be overwritten. This means that the config.php contents will
+always be present.
+
+https://github.com/owncloud/core/pull/37455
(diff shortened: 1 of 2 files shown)


micbar (changes requested):
Please evaluate with PM and deployments before moving on.

mmattel:
Q:
if (1) `config.php`  and (2) `my-other-config.php` has the same key but different value pairs, which will then take precedence. Does reading first implies first key wins on multiple occurences or do we still have last key overwrites pre-last.

In any case, this PR is doc relevant, pls file an issue and reference

jvillafanez:
Currently, "last key wins". This PR change it to "first key wins". In practice, this PR ensures that the keys in the config.php don't get overwritten by other configuration files.
This is still under evaluation and it isn't clear if this will be merged.

felixboehm:
rejected. we have to keep the load order of config files.
1. config.php
2. other.config.php

This mechanism is used to force certain configs by admins.
The issue is, that oc writes to the config, which is bad when config files are expected to not change.

micbar:
I fully agree with that!

Overriding config.php with more specific values is a needed feature.

PR #37471, opened 2020-06-02
[Tests-Only] Add PHPdoc for possible null return values of IRequest methods

phil-davis:
## Description
Some methods in `IRequest` can return `null` if the related information does not exist in the request. But those are not documented in the PHP doc. That is causing `phpstan` in some apps to complain about code that calls these methods and then, for example, compares the return value to `null`.

Add the possibility of `null` return value to the PHP doc.

I put `[Tests-Only]` on this PR because it really only effects code-analysis tools, not actual PHP production run-time. So IMO a changelog is not needed.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -288,7 +288,7 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 	 * This method returns null if the header did not exist.
 	 *
 	 * @param string $name
-	 * @return string
+	 * @return string|null
 	 */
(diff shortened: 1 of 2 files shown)


jvillafanez (approved):

phil-davis:
2 drone CI pipelines fail because fsweb.test.owncloud.com is down.
I will override the result when it comes.

Issue #35484, opened 2019-06-11
sharing via API and changing the cases in the username does not work correctly


individual-it:
### Steps to reproduce
1. create a user called `user1`
2. share a file via API with `USER1`
3. as `user1` check if you received the file

similar to https://github.com/owncloud/enterprise/issues/3241
related to https://github.com/owncloud/core/pull/35084 (was reverted because of  #35382)

tests for this case were :snake: :oil_drum: 

### Expected behaviour
sender should see the file as shared
receive should receive the file

### Actual behaviour
API replies to the sender that everything is fine
sender sees the file as shared
receiver cannot access the file

### Server configuration
**Operating system**:
Ubuntu 18.04
**Web server:**
Apache2
**Database:**
SQlite
**PHP version:**
7.1
**ownCloud version:** (see ownCloud admin page)
11.0 (prealpha)
**Updated from an older ownCloud or fresh install:**
fresh
**Where did you install ownCloud from:**
git
#### curl
```
curl -u admin:admin http://localhost/ocs/v1.php/apps/files_sharing/api/v1/shares -d "path=%2Florem.txt&shareType=0&shareWith=USER1"
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data>
  <id>7</id>
  <share_type>0</share_type>
  <uid_owner>admin</uid_owner>
  <displayname_owner>admin</displayname_owner>
  <permissions>19</permissions>
  <stime>1560254506</stime>
  <parent/>
  <expiration/>
  <token/>
  <uid_file_owner>admin</uid_file_owner>
  <displayname_file_owner>admin</displayname_file_owner>
  <path>/lorem.txt</path>
  <item_type>file</item_type>
  <mimetype>text/plain</mimetype>
  <storage_id>home::admin</storage_id>
  <storage>1</storage>
  <item_source>11</item_source>
  <file_source>11</file_source>
  <file_parent>5</file_parent>
  <file_target>/lorem.txt</file_target>
  <share_with>USER1</share_with>
  <share_with_displayname>user1</share_with_displayname>
  <share_with_additional_info/>
  <mail_send>0</mail_send>
  <attributes/>
 </data>
</ocs>

```

individual-it:
not a regression, same issue in 10.1.1

micbar:
[...]

PR #37475, opened 2020-06-02
[Tests-Only] Test with some unusual usernames

phil-davis:
Trying out this test functionality. This should run the whole test suite with some different usernames, display names etc.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1304,7 +1304,9 @@ def acceptance():
 		'useHttps': True,
 		'extraSetup': [],
 		'extraServices': [],
-		'extraEnvironment': {},
+		'extraEnvironment': {
+			'REPLACE_USERNAMES': True,
+		},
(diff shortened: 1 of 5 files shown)


dpakach:
@phil-davis I have added a commit that should fix the issues with most failing tests except for 
```
tests/acceptance/features/apiShareManagementBasic/createShare.feature:324
```
Since this scenario is actually a bug demonstration, it still fails
issue in https://github.com/owncloud/core/issues/35484

phil-davis:
> Since this scenario is actually a bug demonstration, it still fails

@dpakach I added a special step to cover that scenario, so it understands under which conditions the file should be seen or not seen.

phil-davis:
Code changes from here have been cherry-picked into #37502 to be reviewed and merged.

phil-davis:
Passing acceptance tests - great.

PR #37481, opened 2020-06-03
[Tests-only] Improve error message when XML is expected

PVince81:
Whenever we expect an XML response, throw a more clear exception.
In case the body is malformed, embed it in the exception message to get
it in the test output.

### DIFF
--- a/tests/TestHelpers/HttpRequestHelper.php
+++ b/tests/TestHelpers/HttpRequestHelper.php
@@ -334,7 +334,16 @@ class HttpRequestHelper {
 	public static function getResponseXml($response) {
 		// rewind just to make sure we can re-parse it in case it was parsed already...
 		$response->getBody()->rewind();
-		return new SimpleXMLElement($response->getBody()->getContents());
+		$contents = $response->getBody()->getContents();
+		try {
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
```diff
@@ -190,11 +190,11 @@ Feature: create a public link share
       | 1               | 100             |
       | 2               | 200             |
 
-  Scenario Outline: Getting the share information of passwordless public-links hides credential placeholders
```
phil-davis, line 193: Changes to this file need to be removed.

PVince81: Indeed. Sorry for the mix up.

Adjusted.

phil-davis (approved):

phil-davis:
drone CI will fail for some unit test pipelines because fsweb.test.owncloud.com is down.
https://drone.owncloud.com/owncloud/core/25193/39/9

I will override the result after I see that all the acceptance test pipelines have passed.

PR #37502, opened 2020-06-09
[Tests-Only] Adjust tests so they run with numeric usernames

### LINKED ISSUE #37493: Adjust tests that check mixed-case usernames
Some acceptance tests check results when a user is, for example, created as `Alice` and then accessed as `alice` or `ALICE`.

When we do username replacement, and have a replacement username that is the same when "upper" or "lower" (e.g. usernames that are just numbers, dots) then such scenarios do not work nicely, and some fail. e.g. see example PR #37475 

Investigate if we can make those more flexible so that the test scenarios understand when the uppercase/lowercase of the username are the same, and have the appropriate expectations in `Then` steps.

phil-davis:
## Description
Adjust the acceptance tests so that they can run OK with usernames that do not have any alphabetic characters (e.g. "000" "1.1" etc)

## Related Issue
- Fixes #37493 

## Motivation and Context
Have flexible acceptance tests.

## How Has This Been Tested?
CI

See demonstration PR #37475 that runs the acceptance tests with usernames like "000" "1.1"

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareManagementBasic/createShare.feature
+++ b/tests/acceptance/features/apiShareManagementBasic/createShare.feature
@@ -335,7 +335,7 @@ Feature: sharing
     #And user "brian" should see the following elements
     #  | /randomfile.txt |
     #And the content of file "randomfile.txt" for user "brian" should be "Random data"
-    And user "brian" should not see the following elements
+    And user "brian" should not see the following elements if the upper and lower case username are different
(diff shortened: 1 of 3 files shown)


dpakach (approved):
LGTM :+1:

PR #37506, opened 2020-06-09
[Tests-Only] Test with some unusual usernames

phil-davis:
Trying out this test functionality. This runs the whole test suite with some different usernames, display names etc. It should be passing now.


phil-davis:
Good - passes. Now to try some different usernames...

phil-davis:
See #38111 for the next test like this.

PR #37507, opened 2020-06-09
[Tests-Only] Consolidate OCIS skeleton creation

phil-davis:
## Description
When using a step like this on OCIS:
```
Given these users have been created with default attributes and skeleton files:
      | username |
      | Alice    |
      | Brian    |
```
the skeleton files do not get created.

But when using a step like this, the skeleton files do get created:
```
```

That is because the special code to copy the skeleton files in-place when running on OCIS is only in `createUser`, but it needs to also be in the "batch user creation" function `usersHaveBeenCreated`

Make it so.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -817,11 +817,14 @@ trait Provisioning {
 	 *
 	 * @param boolean $initialize
 	 * @param array $usersAttributes
+	 * @param boolean $skeleton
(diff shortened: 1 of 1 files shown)


individual-it (approved):

individual-it:
@jasson99 please try if it works

PR #37509, opened 2020-06-10
[Tests-Only] check for the existence of group in core db when testing with ldap

### LINKED ISSUE #576: Acceptance tests failing when checking if group exists
e.g. https://drone.owncloud.com/owncloud/user_ldap/2593/28/11

```
  Scenario Outline: admin creates a database group when ldap is enabled                                      # /var/www/owncloud/testrunner/apps/user_ldap/tests/acceptance/features/apiProvisioningLDAP/groups.feature:12
    Given using OCS API version "<ocs-api-version>"                                                          # FeatureContext::usingOcsApiVersion()
    When the administrator sends a group creation request for group "simplegroup" using the provisioning API # FeatureContext::adminSendsGroupCreationRequestUsingTheProvisioningApi()
    Then the OCS status code should be "<ocs-status-code>"                                                   # OCSContext::theOCSStatusCodeShouldBe()
    And the HTTP status code should be "<http-status-code>"                                                  # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And group "simplegroup" should exist                                                                     # FeatureContext::groupShouldExist()

    Examples:
      | ocs-api-version | ocs-status-code | http-status-code |
      | 1               | 100             | 200              |
        Group 'simplegroup' should exist but does not exist
      | 2               | 200             | 200              |
        Group 'simplegroup' should exist but does not exist
```

It seems that the step `groupExists()` was changed in core PR https://github.com/owncloud/core/pull/37494 

I think it neeeds to be more flexible so it knows that when `isTestingWithLdap()` it also needs to understand if the group is expected to exist in LDAP or in the core database backend.

jasson99:
## Description
check for the existence of group in core db when testing with ldap

## Related Issue
- Fixes https://github.com/owncloud/user_ldap/issues/576

## How Has This Been Tested?
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -3239,9 +3239,9 @@ trait Provisioning {
 	 * @throws Exception
 	 */
 	public function groupExists($group) {
-		$baseDN = $this->getLdapBaseDN();
-		$newDN = 'cn=' . $group . ',ou=' . $this->ou . ',' . $baseDN;
 		if ($this->isTestingWithLdap() && OcisHelper::isTestingOnOcis()) {
+			$baseDN = $this->getLdapBaseDN();
+			$newDN = 'cn=' . $group . ',ou=' . $this->ou . ',' . $baseDN;
 			if ($this->ldap->getEntry($newDN) !== null) {
 				return true;
 			}
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -3241,7 +3241,7 @@ trait Provisioning {
 	public function groupExists($group) {
 		$baseDN = $this->getLdapBaseDN();
 		$newDN = 'cn=' . $group . ',ou=' . $this->ou . ',' . $baseDN;
-		if ($this->isTestingWithLdap()) {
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -3241,7 +3241,7 @@ public function userTriesToDeleteGroupUsingTheProvisioningApi($user, $group) {
 	public function groupExists($group) {
 		$baseDN = $this->getLdapBaseDN();
 		$newDN = 'cn=' . $group . ',ou=' . $this->ou . ',' . $baseDN;
-		if ($this->isTestingWithLdap()) {
+		if ($this->isTestingWithLdap() && OcisHelper::isTestingOnOcis()) {
```
phil-davis, line 3244: while we are here, we could move more lines inside the `if` block. It looks strange to call `getLdapBaseDN()` when there is no LDAP. `$baseDN` and `$newDN` are only used inside the `if` block.

jasson99: fixed

phil-davis (approved):

PR #37511, opened 2020-06-10
Store the personal certificates in the user's home directory

jvillafanez:
## Description
Store personal SSL certificates inside the user's home folder.
Note that there are some additional cases (commented in the code) to be checked, which might need a bigger refactor

## Related Issue
https://github.com/owncloud/enterprise/issues/3961

## How Has This Been Tested?
Manually verified with additional changes in the code to open an easy entrypoint. Note that there isn't an official way to provide personal certificates in core, and there could be 3rd party apps using this support.
The certificates end up in the right location

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Security/CertificateManager.php
+++ b/lib/private/Security/CertificateManager.php
@@ -28,8 +28,10 @@
 namespace OC\Security;
 
 use OC\Files\Filesystem;
+use OC\Files\View;
 use OCP\ICertificateManager;
 use OCP\IConfig;
+use OCP\IUserManager;
 
 /**
  * Manage trusted certificates for users
@@ -50,15 +52,19 @@ class CertificateManager implements ICertificateManager {
 	 */
 	protected $config;
 
+	/** @var IUserManager */
+	protected $userManager;
+
 	/**
 	 * @param string $uid
 	 * @param \OC\Files\View $view relative to data/
 	 * @param IConfig $config
 	 */
-	public function __construct($uid, \OC\Files\View $view, IConfig $config) {
+	public function __construct($uid, View $view, IConfig $config, IUserManager $userManager) {
 		$this->uid = $uid;
 		$this->view = $view;
 		$this->config = $config;
+		$this->userManager = $userManager;
 	}
 
 	/**
@@ -203,7 +209,7 @@ class CertificateManager implements ICertificateManager {
 		}
 		if ($this->needsRebundling($uid)) {
 			if ($uid === null) {
-				$manager = new CertificateManager(null, $this->view, $this->config);
+				$manager = new CertificateManager(null, $this->view, $this->config, $this->userManager);
 				$manager->createCertificateBundle();
 			} else {
 				$this->createCertificateBundle();
@@ -220,7 +226,30 @@ class CertificateManager implements ICertificateManager {
 		if ($uid === '') {
 			$uid = $this->uid;
 		}
-		$path = $uid === null ? '/files_external/' : '/' . $uid . '/files_external/';
+
+		if ($uid !== null) {
+			$userObj = $this->userManager->get($uid);
+			if ($userObj !== null) {
+				$homeDir = $userObj->getHome();
+				$dataDir = $this->config->getSystemValue('datadirectory', \OC::$SERVERROOT . '/data');
+				if (\strncmp($homeDir, $dataDir, \strlen($dataDir)) === 0) {
+					$relativeHomeDir = \trim(\substr($homeDir, \strlen($dataDir)), '/');
+					$path = "/{$relativeHomeDir}/files_external/";
+				} else {
+					// FIXME: What to do here? Returned path is expected to start with "/"
+					// so returning an absolute path won't work because they'll be used
+					// as relative path from the server root
+					// Fallback to previous behaviour for now
+					$path = "/{$uid}/files_external/";
+				}
+			} else {
+				// FIXME: This might be wrong, need to change the interface to throw exception otherwise
+				// fallback to the previous behaviour for now
+				$path = "/{$uid}/files_external/";
+			}
+		} else {
+			$path = '/files_external/';
+		}
 
 		return $path;
 	}
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -553,7 +553,7 @@ class Server extends ServerContainer implements IServerContainer, IServiceLoader
 			$uid = $user ? $user : null;
 			return new ClientService(
 				$c->getConfig(),
-				new \OC\Security\CertificateManager($uid, new View(), $c->getConfig())
+				new \OC\Security\CertificateManager($uid, new View(), $c->getConfig(), $c->getUserManager())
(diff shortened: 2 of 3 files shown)


mmattel:
The label 3-to review was set but no reviewer added...

phil-davis:
A drone agent seemed to hang and timeout after 3 hours: https://drone.owncloud.com/owncloud/core/25332/116/15

I restarted the drone CI just now.

mmattel:
Q: does this PR need a docs change?

jvillafanez:
I don't think this needs any change in the docs. It isn't a new feature.

mmattel:
With this new PR, certificates are stored in the users `files_external` directory.
What are you doing with existing user certificates? Is there any mitration possible/necessary?

Coming back to docs, if we delete a user with the current setup, will the certificates get orphaned? If so, we should write for 10.5 that deleting a user also deletes the certificates. (ref: https://doc.owncloud.com/server/10.5/admin_manual/configuration/server/import_ssl_cert.html)
If certificates do not get orphaned, what is the way of removing the certificates exactly for that deleted user. There is also no mentioning in docs about that?

jvillafanez:
> What are you doing with existing user certificates? Is there any migration possible/necessary?

That's something to be decided. I don't know how many installations are / will be affected, or how big those installations are. The easiest workaround would be for the users to reupload the certificates.
Note that personal certificates doesn't seem to be "officially" supported by ownCloud for a while, not sure if intentionally or not. I haven't seen a way to upload personal certificates from the web UI.

> if we delete a user with the current setup, will the certificates get orphaned?

Certificates are stored within the user's data directory, so they're expected to be removed when the user's account is removed.
In case of the certificates stored in the wrong user's directory, the user delete command has been improved with a `--force` option recently to remove the user even if it doesn't exists. This should help to deal with this case by removing the wrong account (not tested, but should work)

mmattel:
@jvillafanez 
For better understanding this change, you may add a certificate with the existing functionality and print the path and then do the same with this PR. So we can see the difference more easily.

jvillafanez:
The case where this was detected was:
* A user from ldap ("abcde") with a home directory of "homes/001-abcde" (for example).
* The user_ldap app configured to use the home directory attribute

The expected base path for the user's content in this case is `<data-directory>/homes/001-abcde/`. Certificates for that user should be stored in `<data-directory>/homes/001-abcde/files_external`, however, without this fix, the certificates end up in `<data-directory>/abcde/files_external`.

Note that deleting the "abcde" user in this scenario is expected to delete the `<data-directory/homes/001-abcde` folder as long as the user is present with that directory as home.
Also note that there could be a valid "abcde" local account with `<data-directory>/abcde` folder as home

There could be additional twisted scenarios to be taken into account, and it's unclear how many installations are using this feature a part from the one that reported this problem.

pako81:
> That's something to be decided. I don't know how many installations are / will be affected, or how big those installations are. The easiest workaround would be for the users to reupload the certificates.

I don't think there is a way for users to upload such certificates. Those are automatically created. So maybe the way to go in case of such migration scenarios is having the admin moving over certs for existing users to the right location? For new users, certs should already land in the correct place thanks to this PR.

My gut feeling is that this affecting a very small percentage of installations out there.

micbar:
@VicDeo can we get a code review?

VicDeo:
@micbar  @jvillafanez  I'm not aware of any ability for **user** to store a certificate...
and `occ security:certificates:import` does **NOT** work per user.

https://github.com/owncloud/enterprise/issues/3961 seems to be wrongly qualified as a bug

micbar:
> @micbar @jvillafanez I'm not aware of any ability for **user** to store a certificate...
> and `occ security:certificates:import` does **NOT** work per user but per instance
> 
> [owncloud/enterprise#3961](https://github.com/owncloud/enterprise/issues/3961) seems to be wrongly qualified as a bug

That is entirely true.

VicDeo:
@micbar wait... I found how to reproduce it, but i need to setup LDAP to check if the fix is working

phil-davis:
And this needs a changelog entry.
So please add one and rebase... when trying out the next iteration.

micbar:
ok, so this is valid - fine.

micbar:
Dont rush, we will not include in 10.5.0 as clarified with @pako81

VicDeo:
@micbar @jvillafanez @pako81 
here what I did:
oc@master, 
user_ldap@master
1. Settings -> admin -> storage 
Allow users to mount external storage  is checked
2. Connect ou with the single user  to ldap, set **deparmentNumber** for this user to 
'/tmp/african'
3. set `User Home Folder Naming Rule` to **attr:departmentNumber**
4. `php occ user:sync -m remove "OCA\User_LDAP\User_Proxy"` 
5. login as this user
6. tried to add Webdav storage for a valid domain at personal -> storage
7. 
```
s -l data/
total 844
drwxr-x--- 4 deo www   4096 Jun 23 12:55 admin
-rw-r--r-- 1 deo www    145 Sep 26  2019 htaccesstest.txt
-rw-r--r-- 1 deo www      0 Jun 23 13:58 index.html
-rw-r--r-- 1 deo www   4096 Jul 30  2019 owncloud.db
-rw-r----- 1 deo www 842168 Jun 23 15:49 owncloud.log
-rw-r--r-- 1 deo www    447 Apr 29 15:56 update.log
```

8. 
```
> ls -l /tmp/african/
total 20
drwxr-x--- 3 deo www 4096 Jun 23 15:39 archive
drwxr-xr-x 2 deo www 4096 Jun 23 15:39 cache
drwxr-xr-x 2 deo www 4096 Jun 23 15:39 files
drwxr-xr-x 2 deo www 4096 Jun 23 15:40 files_external
drwxr-xr-x 3 deo www 4096 Jun 23 15:39 thumbnails
```

so, `files_external` was created at the correct location for master.
It would be nice to have the more details on the env where the bug is reproducible.

phil-davis:
@jvillafanez @micbar is this a thing that is moving ahead for 10.6 ?

jvillafanez:
Without clear steps to reproduce we can't verify properly if this PR solves the problem.

micbar:
needs to be tested with the customer first @cdamken

mmattel:
Ping, any news on that?

phil-davis:
Why did this PR suddenly get a codecov report posted to it?
There was no rebase, and the link to the drone run goes to a drone build that was 6 months ago!

phil-davis:
I rebased - this needs fresh CI anyway.

micbar (approved):
LGTM

phil-davis:
CI was red, so I rebased again. Let's see what happens :)

phil-davis:
https://drone.owncloud.com/owncloud/core/28954/9/8
There are unit test fails. @jvillafanez ?

DeepDiver1975:

PR #37518, opened 2020-06-11
[Tests-Only] Fix PHP doc of Response::render

phil-davis:
## Description
`OCP\AppFramework\Http\Response::render()` can (and usually does) return a string.
Fixup the core PHP doc.

Otherwise `phpstan` reports stuff like:
```
phil@phil-OptiPlex-3050:~/git/owncloud/core/apps-external/configreport$ make test-php-phpstan 
php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan analyse --memory-limit=4G --configuration=./phpstan.neon --no-progress --level=5 appinfo lib
 ------ -------------------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Http/ReportResponse.php                                                                                                           
 ------ -------------------------------------------------------------------------------------------------------------------------------------- 
  54     Return type (string) of method OCA\ConfigReport\Http\ReportResponse::render() should be compatible with return type (null) of method  
         OCP\AppFramework\Http\Response::render()                                                                                              
 ------ -------------------------------------------------------------------------------------------------------------------------------------- 

 [ERROR] Found 1 error                                                                                                  

Makefile:139: recipe for target 'test-php-phpstan' failed
make: *** [test-php-phpstan] Error 1
```

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/public/AppFramework/Http/Response.php
+++ b/lib/public/AppFramework/Http/Response.php
@@ -214,8 +214,11 @@ class Response {
 	}
 
 	/**
-	 * By default renders no output
-	 * @return null
+	 * Return a string to be used in the body of an HTTP response.
+	 * By default returns null and so renders no output.
+	 * Can be overridden to provide the desired response body content.
+	 *
+	 * @return string|null
 	 * @since 6.0.0
 	 */
 	public function render() {


lib/public/AppFramework/Http/Response.php
```diff
@@ -215,7 +215,7 @@ public function getHeaders() {
 
 	/**
 	 * By default renders no output
```
jvillafanez, line 217: Needs clarification on the non-default behaviour. I think the method should return a string that is expected to be used in the body of the HTTP response.
Without this clarification, I could think that it's fine to `echo "blablabla"`.

We might need to include a small note to say that this method could be overwritten in other more specialized responses. I mean, I consider, as part of a good design, that anyone should easily know what methods are expected to be overwritten (usually abstract methods, but not in this case) and have clear expectations about how to do it.
Just a small note should be enough, which also complements the reason why the method could return a string even though it's always returning null.

phil-davis: Added description.

jvillafanez (approved):

PR #37555, opened 2020-06-18
use relative path in shared_by_email activity

micbar:
## Description
Do not use the absolute Webdav Path for the `shared_by_email` activity.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4045

## Motivation and Context
Do not display unecessary information.

## How Has This Been Tested?
- manually

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Tasks
- [x] Needs Changelog item
- [x] Needs manual testing

### DIFF
--- a/lib/private/Share/MailNotifications.php
+++ b/lib/private/Share/MailNotifications.php
@@ -30,7 +30,9 @@
 
 namespace OC\Share;
 
+use OCP\Files\IRootFolder;
 use OCP\Files\Node;
+use OCP\Files\NotFoundException;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IURLGenerator;
@@ -71,6 +73,8 @@ class MailNotifications {
 	private $eventDispatcher;
 	/** @var \OCP\Activity\IManager */
 	private $activityManager;
+	/** @var IRootFolder */
+	private $rootFolder;
 
 	/**
 	 * @param IManager $shareManager
@@ -92,7 +96,8 @@ class MailNotifications {
 		Defaults $defaults,
 		IURLGenerator $urlGenerator,
 		EventDispatcher $eventDispatcher,
-		\OCP\Activity\IManager $activityManager
+		\OCP\Activity\IManager $activityManager,
+		IRootFolder $rootFolder
 	) {
 		$this->shareManager = $shareManager;
 		$this->l = $l10n;
@@ -103,6 +108,7 @@ class MailNotifications {
 		$this->urlGenerator = $urlGenerator;
 		$this->eventDispatcher = $eventDispatcher;
 		$this->activityManager = $activityManager;
+		$this->rootFolder = $rootFolder;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


lib/private/Share/MailNotifications.php
```diff
@@ -272,7 +279,8 @@ public function sendLinkShareMail($sender, $recipients, $link, $personalNote = n
 		$failedRecipients =  $this->sendLinkShareMailFromBody($recipients, $subject, $htmlBody, $textBody, $from, $replyTo);
 		if (empty($failedRecipients)) {
 			$event = $this->activityManager->generateEvent();
-			$path = $shareNode->getPath();
+			$userFolder = $this->rootFolder->getUserFolder($sender->getUID());
+			$path = $userFolder->getRelativePath($shareNode->getPath());
```
karakayasemi, line 283: For reshare case, since the node path is outside of the re-sharer folder, `$path` will be null.

karakayasemi (changes requested):

karakayasemi:
Units tests also need to be adjusted. @micbar if you prefer, I can take over this one.

lib/private/Share/MailNotifications.php
micbar: good catch.

micbar: what could we do? We are not allowed to have a userid here

karakayasemi:
@micbar The code now gets the actual file node of the sender. In this way, I believe the relative path calculated correctly. I manually tested the code with both re-share and the normal case, everyhing worked without problem. Also, unit tests adjusted.

karakayasemi (approved):

karakayasemi:
Changelog added.

micbar (commented):
Do not merge yet. We need to decide if we release it woth 10.5.0

lib/private/Share/MailNotifications.php
```diff
@@ -272,7 +279,11 @@ public function sendLinkShareMail($sender, $recipients, $link, $personalNote = n
 		$failedRecipients =  $this->sendLinkShareMailFromBody($recipients, $subject, $htmlBody, $textBody, $from, $replyTo);
 		if (empty($failedRecipients)) {
 			$event = $this->activityManager->generateEvent();
-			$path = $shareNode->getPath();
+			// it can be a re-share, get actual share file node
+			$userFolder = $this->rootFolder->getUserFolder($sender->getUID());
+			$shareFileNodes = $userFolder->getById($shareNode->getId(), true);
```
micbar, line 284: Why not `$share->getNodeId()`?

karakayasemi: No difference. `$shareNode` already defined in above, no drawback to using it also. There is similar actual share node detection logic in the share manager. Copied it from there.
https://github.com/owncloud/core/blob/f0b2901b0fe912675e9528ee64c2e39d6e36346a/lib/private/Share20/Manager.php#L358-L361

PR #37567, opened 2020-06-22
[Tests-Only] Refactor using the OCS API version defined externally

phil-davis:
## Description
We have a bit of difficulty with acceptance test scenarios that run for multiple versions of some API. e.g. OCS API, or the different webdav/dav endpoint alternatives, or the different public link endpoints. We have to break up existing "Scenario Outline" when the behavior is different (often broken) against core/ocis-reva/ocis.

The code here demonstrates how we could have single scenarios that can be run with the run-time value of OCS API version set by an environment variable. That saves having the constantly-repeated scenario outline examples table. (Or in the apiProvisioning case it merges many scenarios that got duplicated in 2 feature files).

When the scenario outcomes are consistently different for, e.g. OCS API V1 and V2, then we write steps like:
```
    Then the OCS status code should be "100" for API version 1 or "200" for API version 2
```
that define the test expectations in each situation.

- When running from oC10 apps like user_ldap, we can still do the `numberOfParts` thing in `.drone.star` to automatically group all the API suites into some number of pipelines. And we run the whole of that twice, for OCS API v1 and v2. Some suites/features/scenarios maybe do not doing any "OCS API" requests. Those would end up getting run twice. If there are not many like that then that does not really matter.

- need to see if there are many (or any) scenarios that must ONLY  run on a particular OCS API version (e.g. are there features that only exist in OCS API v2, and not in OCS API v1). For those we need some solution (tag...) to skip them on the OCS API version that does not apply.

Now I need to think about the combinations of skip tags that will be needed when the scenario is known to fail, e.g. on OCIS+OCS API v2 but passes with OCIS+OCS API v2.

## Related Issue
https://github.com/owncloud/ocis-reva/issues/282

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
Closing - we are not going this way for now.

PR #37569, opened 2020-06-22
Add new system config to enforce strict login check for password in user backend

mrow4a:
When in LDAP config user login attributes are setuped disallowing login by email, verify settings as below work correctly:

However, as of at least OC9.X, when logging in with LDAP with loginname and password, OC always retries to get a user, and try login agains LDAP also with email. For some customers this should not be allowed.

This PR:
- [x] adds new system config `/occ config:system:set --type boolean --value true strict_login_enforced` that disallows to retry with email
- [x] add unit tests

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -237,6 +237,14 @@ $CONFIG = [
 'token_auth_enforced' => false,
 
 /**
+ * Enforce strict login check with user backend
+ * If enabled, strict login check for password in user backend will be enforced,
+ * meaning only the login name typed by the user would  be validated. With this
+ * configuration enabled, e.g. an additional check for email will not be performed.
+ */
+'strict_login_enforced' => false,
+
+/**
  * Define additional login buttons on the logon screen
  * Provides the ability to create additional login buttons on the logon screen, for e.g., SSO integration
  *  'login.alternatives' => [
--- a/changelog/unreleased/37569
+++ b/changelog/unreleased/37569
@@ -1,0 +1,9 @@
+Security: Add new system config to enforce strict login check with user backend
+
+Adds new system config to enforce strict login check for password
+in user backend, meaning only login name typed by user would be validated.
+With this configuration enabled, e.g. additional check for email will
+not be performed.
+
+https://github.com/owncloud/core/pull/37569
+https://github.com/owncloud/user_ldap/pull/581
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -219,7 +219,7 @@ class LoginController extends Controller {
 		$originalUser = $user;
 		// TODO: Add all the insane error handling
 		$loginResult = $this->userSession->login($user, $password);
-		if ($loginResult !== true) {
+		if ($loginResult !== true && $this->config->getSystemValue('strict_login_enforced', false) !== true) {
 			$users = $this->userManager->getByEmail($user);
 			// we only allow login by email if unique
 			if (\count($users) === 1) {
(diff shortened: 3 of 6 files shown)


micbar (approved):
LGTM

phil-davis:
https://drone.owncloud.com/owncloud/core/25508/3/2
file changelog/unreleased/37569: entry type "Feature" is invalid, valid types: map[Bugfix:2 Change:3 Enhancement:4 Security:1]

micbar:
@C0rby Can you review and check the prio.
We can get this into 10.5.0 if it is really urgent.

phil-davis:
https://drone.owncloud.com/owncloud/core/25509/3/2
file changelog/unreleased/37569: title is too long

Still not getting past the changelog parsing.

micbar:
title is limited to 80 chars

config/config.sample.php
```diff
@@ -236,6 +236,13 @@
  */
 'token_auth_enforced' => false,
 
+/**
+ * Enforce strict login for password authentication that will enforce
```
mmattel, line 240: Please see the other config entries regarding the title line.
The first sentence must end at the same line and is used as item in the table of contents.
You can put more description below that first line.

mmattel:
Doc relevant, pls file a doc issue, because we need to do a config-to-docs run.
Else this config item will NOT be shown in docs!

pako81:
What about the use case where only the log in via email address for LDAP users should be prevented, while local users should still be able to normally log in via email address?

micbar:
@pako81 good point.

@mmattel could a config.sample.php doc not be sufficient? This is a very special use case.

phil-davis:
IMO @mmattel means just that - when this is merged we just need to do the "config-to-docs run" which will update the `config.sample.php` that is published in the docs. Nothing more will need to be written.

mrow4a:
> What about the use case where only the log in via email address for LDAP users should be prevented (by unchecking the LDAP Login Attribute), while local users should still be able to normally log in with email?

@pako81 @micbar I will check it, but again: strict login is strict login, if Local Users Backend has support only for UID that internally is not email, and LDAP allows only UID,MYSPECIALATTRIBUTE, we should only login on these ones..

mrow4a:
@pako81 @micbar I can confirm that local user backend `OC\User\Database` currently supports searching for users only on their UIDs. This support can be of course extended to search also on other attributes as in LDAP, and this would be the cleanest approach. 

So as mentioned, whoever enables `occ config:system:set --type boolean --value true strict_login_enforced` has to understand that ALL user backends will check passwords only againdsy user types logins. 

Personal Opinion: `strict_login_enforced` should be default in OC, but probably because of legacies it would cause mess..

config/config.sample.php
mrow4a: done

config/config.sample.php
```diff
@@ -236,6 +236,15 @@
  */
 'token_auth_enforced' => false,
 
+/**
+ * Add new system config to enforce strict login check with user backend
```
mmattel, line 240: Attn, proposing the following:

```
/** 
 * Enforce strict login check with user backend
 * If enabled, strict login check for password in user backend will be enforced, meaning only
 * login name typed by user would  be validated. With this configuration enabled, e.g. additional
 * check for email will not be performed.
 */
```

1) The first line must not have a text starting with "Add new config...". This text would appear in the table of contents.
2) The second line contained an error. There was a hard linebreak and the third line did not start with ` * `. This would lead in an parsing error when running config-to-docs...

config/config.sample.php
mmattel: pls see my comment on the changes below

config/config.sample.php
phil-davis: done

phil-davis (approved):

config/config.sample.php
micbar: input adressed

PR #37573, opened 2020-06-22
Bump version in version.php

jnweiger:
Let's move ahead:
RC1 shall provide an owncloud core to test apps against. It is not intended to become the release.

As soon as we have some newer commits to the release branch, or a few more apps built as RC1 we will build an RC2 with all what we have.


micbar:
@jnweiger Nice, but we need to merge back the release branch before we go to rc1.
But we can do that after the version bump.

micbar (approved):

PR #37591, opened 2020-06-24
Make login page consistent with strict_login_enforced config

mrow4a:
When `occ config:system:set --type boolean --value true strict_login_enforced` is set, we should also display different login page (as we no longer allow only username or emails, but "login" in general only in this case). 

Without `strict_login_enforced` (default)
Zrzut ekranu 2020-06-24 o 21 06 43

With `strict_login_enforced` 
Zrzut ekranu 2020-06-24 o 21 06 08

High prio, and needs translations as this is very user facing @micbar

### DIFF
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -198,6 +198,8 @@ class LoginController extends Controller {
 			}
 		}
 
+		$parameters['strictLoginEnforced'] = $this->config->getSystemValue('strict_login_enforced', false);
+
 		return new TemplateResponse(
 			$this->appName, 'login', $parameters, 'guest'
 		);
--- a/core/templates/login.php
+++ b/core/templates/login.php
@@ -48,11 +48,11 @@ script('core', [
 	?> shake<?php
 } ?>">
 			<input type="text" name="user" id="user"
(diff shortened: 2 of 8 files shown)


micbar:
:+1: Great catch!

micbar:
Can you rebase that and change the base to `release-10.5.0` branch?

mrow4a:
@phil-davis could you help finishing this ?

phil-davis:
I have fixed the unit tests and added acceptance tests that check the behavior when trying to login with username or email address when `strict_login_enforced` is set...

IMO this should be merged to `master` so that the translation string can get to Transifex and we can do the translations and get them back into core.

I raised PR #37592 to cherry-pick this across to `release-10.5.0`

phil-davis:
IMO no changelog is needed, there is already a changelog for the `strict_login_enforced` and this is just a small fix/enhancement of it.

individual-it (commented):
tests look good to me

phil-davis (approved):
The matching PR #37592 to `release-10.5.0` branch has been approved. And this PR has had review of the tests. So we are all good.

PR #37600, opened 2020-06-26
[10.5.0] Update translations

phil-davis:
## Description
"backport" the Transifex changes in `master` from the last 2 days into the `release-10.5.0` branch.

This gets the translations up-to-date. In particular some translations for "Manual File Locking" and "Login"

Note: no changelog needed - we do not do changelogs for commits that came across from Transifex

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


core/l10n/bs.json
```diff
@@ -152,6 +152,7 @@
     "Search" : "Potraži",
     "Server side authentication failed!" : "Autentikacija na strani servera nije uspjela!",
     "Please contact your administrator." : "Molim kontaktirajte svog administratora.",
+    "Login" : "Prijava",
```
DeepDiver1975, line 155: There is something wrong.
We have login an log in.
This should be consistent.

DeepDiver1975 (changes requested):

DeepDiver1975:
Strange that these commits are done by me .... Should be owncloud account

core/l10n/bs.json
phil-davis: There are existing sentences like:
`core/templates/login.php`

```
You are trying to access a private link. Please log in first.
```
And
```
			<input type="submit" id="submit" class="login primary icon-confirm" title="<?php p($l->t('Log in')); ?>" value="" disabled="disabled"/>
```
which is the hover text of the "right-arrow" icon that you press to "log in" or "login".

These are existing things.

IMO the translations are already in core `master`. So we need to get them into `release-10.5.0` anyway.

And as a separate issue, we need to decide if it should be "login" or "log in" in each place.

phil-davis:
> Strange that these commits are done by me .... Should be owncloud account

The Transifex commit has been generated with your name for "a long time"  (tm)

core/l10n/bs.json
phil-davis: I raised issue #37603

DeepDiver1975:
> The Transifex commit has been generated with your name for "a long time" (tm)

no - there has been a change recently - see e.g. https://github.com/owncloud/core/commits/632544afb539dbd66906522efd8f5fd978f29b2b/core/l10n/af_ZA.js

core/l10n/bs.json
jnweiger: The verb in `Please log in first` and the noun `Login` should be spelled exactly like seen above. One with a whitespace, the other without.

DeepDiver1975: > You are trying to access a private link. Please log in first.

this is not the string we see in the translations files. There must we a single string 'Log in' ...

phil-davis:
@DeepDiver1975 see https://github.com/owncloud/core/issues/37603#issuecomment-650053201

The "new" "Login" form seems to be preferred to that other string "Log in", so actually I will probably make a PR to change that "Log in" to "Login". That means we do need the string "Login" translated.

So, IMO, this translation PR  to `release-10.5.0` can go ahead. (And more translation changes will come every day anyway)

Please review again.

core/l10n/bs.json
phil-davis: "Log in" is hover-text on the "submit" button, which is the right-arrow icon on the login page.
"Login" is the placeholder text in the "username" field on the login page when `strict_login_enforced` is set.

DeepDiver1975 (approved):

PR #37608, opened 2020-06-29
[Tests-Only] update tests related to deleting public shares

C0rby:
## Description
Update tests related to deleting public links
Maybe the tags of the feature must be changed to make the tests run with the ocis tests. I don't know which should be set.
PR causing the test changes: https://github.com/cs3org/reva/pull/899
## Types of changes

- [ ] Technical debt

### DIFF
--- a/tests/acceptance/features/apiSharePublicLink1/deletePublicLinkShare.feature
+++ b/tests/acceptance/features/apiSharePublicLink1/deletePublicLinkShare.feature
@@ -1,4 +1,4 @@
-@api @skipOnOcis
+@api
 Feature: delete a public link share
 
   Background:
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -2369,7 +2369,11 @@ public function getPublicShareIDByName($user, $path, $name) {
 		$dataResponded = $this->getShares($user, $path);
 		foreach ($dataResponded as $elementResponded) {
 			if ((string) $elementResponded->name[0] === $name) {
-				return (int) $elementResponded->id[0];
+				if (\TestHelpers\OcisHelper::isTestingOnOcis()) {
+					return (string) $elementResponded->id[0];
```
phil-davis, line 2373: @C0rby see PR #37609 - IMO we can always treat this as string. No need for the special `isTestingOnOcis()` check.

I made the separate PR to see if all the tests are happy with it in core. If we merge that PR, then the diffs here will not be needed.

C0rby: Nice! I wasn't sure if the cast to int was optional so I left it in but yeah after the other PR finishes we can remove this here.

phil-davis:
@C0rby PR #37609 has been merged. You can rebase here and remove the not-needed diffs.

phil-davis (approved):
PR https://github.com/cs3org/reva/pull/899 was merged into reva last night. And the latest reva was updated in ocis-reva and the latest ocis-reva was updated in ocis. So this fix is implemented everywhere.

PR #37609, opened 2020-06-29
[Tests-Only] Treat public link share id as string in acceptance tests

phil-davis:
## Description
The acceptance tests have code to find the public link share id that has a given name. In core that share id usually looks like a number (int). But actually it is just a string that identifies the public link share.

In other implementations (e.g. OCIS) it might be alphanumeric. So treat it as a general string.

PR #37608 brought this up. IMO rather than have different code if running-on-OCIS, we can always treat this as a string.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -2363,13 +2363,13 @@ trait Sharing {
 	 * @param string $path to share
 	 * @param string $name of share
 	 *
-	 * @return int|null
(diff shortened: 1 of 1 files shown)


C0rby (approved):
Cool! :+1:

individual-it (approved):

PR #37615, opened 2020-06-30
Introduce files-multiget plugin to allow zip archive download of sele…

DeepDiver1975:
…cted files

## Description
A webdav report is introduced to support a multi get request - similar to the carddav multi get request.

For the Webdav API endpoint (to be implemented by either OC 10 or the download service) for both bulk download and multiple resource download:

We could use a multiget REPORT call like https://tools.ietf.org/html/rfc6352#page-31 but only deliver zip file, as requested through an "Accept" header.
The REPORT XML request body would contain a list of URLs / resources to retrieve, relative to the request path:

REPORT Request example to http://host/owncloud/remote.php/dav/files/someuser/

Headers:
```
Accept: application/zip
```

Request body:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<oc:files-multiget
    xmlns:a="DAV:"
    xmlns:oc="http://owncloud.org/ns">
    <DAV:href>first_resource</DAV:href>
    <DAV:href>second_resource</DAV:href>
    <DAV:href>some_directory</DAV:href>
</oc:multiget>
```

Response:
- content-disposition header with a file name that is either "download.zip" (or tar if tar was requested), or in case only a single item was selected, the name of said item with ".zip" appended.
- a zip file with all the above packaged in

In case of error (ex: firewall), a 403 status code would be returned and the operation is aborted.
If one of the resources does not exist, a 404 status code would be returned and the operation is aborted.

This can be used for single folder download also.

We can use the same format also for the **public dav enpoint**

## Related Issue
https://github.com/owncloud/enterprise/issues/3916

## How Has This Been Tested?
- :hand: 

## Types of changes
- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Files/MultiGetPlugin.php
+++ b/apps/dav/lib/Files/MultiGetPlugin.php
@@ -1,0 +1,151 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2020, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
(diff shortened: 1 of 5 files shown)


mmattel:
@DeepDiver1975 
Shall I add the changelog description text to the PR description?
This helps to better read what this PR is about without clicking a lot and finally reading the changelog :smile:

DeepDiver1975:
> @DeepDiver1975
> Shall I add the changelog description text to the PR description?
> This helps to better read what this PR is about without clicking a lot and finally reading the changelog smile

hold your horses - this is a draft ... only opened to get ci running.

mmattel:
ok
running_horse

PR #37619, opened 2020-07-01
[Tests-only] Sharing tests use string as ids

PVince81:
Adjust sharing test expectations to expect ids of shares and files as
string instead of int.

This aligns with the requirements from OCIS.

Partly for https://github.com/owncloud/ocis-reva/issues/252 as it blocks some tests

### DIFF
--- a/tests/acceptance/features/apiShareOperations/gettingShares.feature
+++ b/tests/acceptance/features/apiShareOperations/gettingShares.feature
@@ -102,20 +102,20 @@ Feature: sharing
     Then the OCS status code should be "<ocs_status_code>"
     And the HTTP status code should be "200"
     And the fields of the last response to user "Alice" sharing with user "Brian" should include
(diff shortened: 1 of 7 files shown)


tests/acceptance/features/apiShareOperations/gettingShares.feature
```diff
@@ -102,20 +102,20 @@ Feature: sharing
     Then the OCS status code should be "<ocs_status_code>"
     And the HTTP status code should be "200"
     And the fields of the last response to user "Alice" sharing with user "Brian" should include
-      | id                     | A_NUMBER           |
+      | id                     | A_STRING			|
```
phil-davis, line 105: column width looks odd. But not worth running the whole of CI again just for that!

phil-davis (approved):

PVince81:
cool, didn't break anything :tada:

PR #37634, opened 2020-07-02
Support signed urls

DeepDiver1975:
## Description
- OCS API /cloud/user/signing-key to get the user's key for url signing
- capability to state that signed urls are supported
- allow signed url access for webdav endpoints

## Related Issue
- Refs https://github.com/owncloud/enterprise/issues/3912

## How Has This Been Tested?
- https://github.com/owncloud/owncloud-sdk/pull/504

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Auth.php
+++ b/apps/dav/lib/Connector/Sabre/Auth.php
@@ -29,15 +29,16 @@
  */
 namespace OCA\DAV\Connector\Sabre;
 
-use Exception;
 use OC\AppFramework\Http\Request;
 use OC\Authentication\Exceptions\PasswordLoginForbiddenException;
 use OC\Authentication\TwoFactorAuth\Manager;
 use OC\Authentication\AccountModule\Manager as AccountModuleManager;
+use OC\Security\SignedUrl\Verifier;
 use OC\User\LoginException;
 use OC\User\Session;
 use OCA\DAV\Connector\Sabre\Exception\PasswordLoginForbidden;
 use OCP\Authentication\Exceptions\AccountCheckException;
+use OCP\IConfig;
 use OCP\IRequest;
 use OCP\ISession;
 use Sabre\DAV\Auth\Backend\AbstractBasic;
@@ -47,7 +48,7 @@ use Sabre\HTTP\RequestInterface;
 use Sabre\HTTP\ResponseInterface;
 
 class Auth extends AbstractBasic {
-	const DAV_AUTHENTICATED = 'AUTHENTICATED_TO_DAV_BACKEND';
+	public const DAV_AUTHENTICATED = 'AUTHENTICATED_TO_DAV_BACKEND';
 
 	/** @var ISession */
 	private $session;
@@ -55,12 +56,14 @@ class Auth extends AbstractBasic {
 	private $userSession;
 	/** @var IRequest */
 	private $request;
-	/** @var string */
-	private $currentUser;
 	/** @var Manager */
 	private $twoFactorManager;
 	/** @var AccountModuleManager */
 	private $accountModuleManager;
+	/**
+	 * @var IConfig
+	 */
+	private $config;
 
 	/**
 	 * @param ISession $session
@@ -68,6 +71,7 @@ class Auth extends AbstractBasic {
 	 * @param IRequest $request
 	 * @param Manager $twoFactorManager
 	 * @param AccountModuleManager $accountModuleManager
+	 * @param IConfig $config
 	 * @param string $principalPrefix
 	 */
 	public function __construct(ISession $session,
@@ -75,6 +79,7 @@ class Auth extends AbstractBasic {
 								IRequest $request,
 								Manager $twoFactorManager,
 								AccountModuleManager $accountModuleManager,
+								IConfig $config,
 								$principalPrefix = 'principals/users/') {
 		$this->session = $session;
 		$this->userSession = $userSession;
@@ -86,6 +91,7 @@ class Auth extends AbstractBasic {
 		// setup realm
 		$defaults = new \OC_Defaults();
 		$this->realm = $defaults->getName();
+		$this->config = $config;
 	}
 
 	/**
@@ -158,7 +164,7 @@ class Auth extends AbstractBasic {
 			throw new NotAuthenticated($e->getMessage(), $e->getCode(), $e);
 		} catch (NotAuthenticated $e) {
 			throw $e;
-		} catch (Exception $e) {
+		} catch (\Exception $e) {
 			$class = \get_class($e);
 			$msg = $e->getMessage();
 			throw new ServiceUnavailable("$class: $msg");
@@ -205,6 +211,7 @@ class Auth extends AbstractBasic {
 	 * @return array
 	 * @throws NotAuthenticated
 	 * @throws ServiceUnavailable
+	 * @throws LoginException
 	 */
 	private function auth(RequestInterface $request, ResponseInterface $response) {
 		$forcedLogout = false;
@@ -230,7 +237,6 @@ class Auth extends AbstractBasic {
 				$this->checkAccountModule($user);
 				$uid = $user->getUID();
 				\OC_Util::setupFS($uid);
-				$this->currentUser = $uid;
 				$this->session->close();
 				return [true, $this->principalPrefix . $uid];
 			}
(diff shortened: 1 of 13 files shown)


apps/dav/lib/Connector/Sabre/Auth.php
```diff
@@ -230,7 +236,7 @@ private function auth(RequestInterface $request, ResponseInterface $response) {
 				$this->checkAccountModule($user);
 				$uid = $user->getUID();
 				\OC_Util::setupFS($uid);
-				$this->currentUser = $uid;
+				$currentUser = $uid;
```
butonic, line 239: I don't see `$currentUser` being used anywhere (neither was there a `$this->currentUser` to begin with). Remove completely?

core/Controller/CloudController.php
```diff
@@ -71,4 +71,25 @@ public function getCurrentUser() {
...
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 * @CORS
+	 *
+	 * @return array
+	 * @throws \OCP\PreConditionNotMetException
+	 */
+	public function getSigningKey(): array {
+		$userId = \OC_User::getUser();
+		$signingKey = \OC::$server->getConfig()->getUserValue($userId, 'core', 'signing-key', null);
```
butonic, line 85: I haven't looked into oc10 for a while, but will this work when acessing public link pages? No user is logged in -> urls cannot be signed -> thumbnails cannot be fetched on public pages?

DeepDiver1975: Creating a signed url requires a user. So yes .... This will not work for public pages.
But not needed anyway. There are mechanisms in place to build up a public page ....

DeepDiver1975: I mean we can create presigned urls within the server when rendering public pages. Because we know the user as well as the key. On pure client side this will not be possible ...

But this is okay from my understanding ....

PVince81: public pages usually don't have a bearer token, so accessing the URL directly would work

however, public links with password would need said share_token+password to be embedded in the download URL, could be problematic

DeepDiver1975: When accessing a public link the browser will ask for the password and verify this with the backend.
After that a signed url for downloading the file can be created based on the information we have from the share ...

Anyway - this needs a closer look once we get there .... iterative approach ... let's make this happen first.

apps/dav/lib/Files/FileLocksBackend.php
```diff
@@ -192,6 +192,7 @@ public function lock($uri, Locks\LockInfo $lockInfo) {
 
 		// in case the timeout has not been accepted, adjust in lock info
 		$lockInfo->timeout = $lock->getTimeout();
+		$lockInfo->owner = $lock->getOwner();
```
PVince81, line 195: PR mix up ?

core/Controller/CloudController.php
```diff
@@ -71,4 +71,25 @@ public function getCurrentUser() {
...
+		if ($signingKey === null) {
+			$signingKey = \OC::$server->getSecureRandom()->generate(64);
+			\OC::$server->getConfig()->setUserValue($userId, 'core', 'signing-key', $signingKey, null);
```
PVince81, line 88: maybe use the `$preCondition` field to make sure no value exists there to avoid race conditions ?
not sure if "null" works here as precondition in the `setUserValue` method to signify "empty".

lib/private/Security/SignedUrl/Verifier.php
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	public function isSignedRequest(): bool {
+		$params = $this->getQueryParameters();
+		return isset($params['OC-Signature']);
+	}
+
+	public function signedRequestIsValid(): bool {
+		$params = $this->getQueryParameters();
+		$urlSignature = $params['OC-Signature'];
```
PVince81, line 55: in the spec you said some of these are required, so need to check with `isset()` and throw a bad request in case they aren't

lib/private/Security/SignedUrl/Verifier.php
```diff
@@ -0,0 +1,98 @@
...
+		unset($params['OC-Signature']);
+
+		$qp = \http_build_query($params);
+		$url =  \Sabre\Uri\parse($this->getAbsoluteUrl());
+		$url['query'] = $qp;
+		$url = \Sabre\Uri\build($url);
+
+		$signingKey = $this->config->getUserValue($urlCredential, 'core', 'signing-key');
+
+		$hash = \hash_pbkdf2("sha512", $url, $signingKey, 10000, 64, false);
```
PVince81, line 70: seems our `Crypto` class was not up to the task

lib/private/Security/SignedUrl/Verifier.php
```diff
@@ -0,0 +1,98 @@
...
+		return !($date < $this->now);
+	}
+
+	private function getQueryParameters(): array {
+		return $this->request->getQueryParameters();
+	}
+
+	public function getUserId(): string {
+		$params = $this->getQueryParameters();
+		return $params['OC-Credential'];
```
PVince81, line 88: will :boom: when not set ? add `isset` ?

tests/lib/Security/SignedUrl/VerifierTest.php
```diff
@@ -0,0 +1,56 @@
...
+		if ($isSignedUrl) {
+			$this->assertTrue($v->isSignedRequest());
+			$this->assertEquals('alice', $v->getUserId());
+			$this->assertEquals($isUrlValid, $v->signedRequestIsValid());
+		} else {
+			$this->assertFalse($v->isSignedRequest());
+		}
+	}
+
+	public function providesUrls() {
```
PVince81, line 48: add cases where some of the required attrs are missing

lib/private/Security/SignedUrl/Verifier.php
```diff
@@ -0,0 +1,98 @@
...
+
+		$hash = \hash_pbkdf2("sha512", $url, $signingKey, 10000, 64, false);
+		if ($hash !== $urlSignature) {
+			return false;
+		}
+		if (\strtoupper($this->getMethod()) !== \strtoupper($urlVerb)) {
+			return false;
+		}
+		$date = new \DateTime($urlDate);
+		$date->add(new \DateInterval("PT${urlExpires}S"));
```
PVince81, line 78: a bit obscure, but ok, someone posted in https://www.php.net/manual/en/class.dateinterval.php#102930 about that. maybe add a PHP comment to avoid head-scratching ?

lib/private/Security/SignedUrl/Verifier.php
```diff
@@ -0,0 +1,98 @@
...
+		return !($date < $this->now);
+	}
+
+	private function getQueryParameters(): array {
+		return $this->request->getQueryParameters();
+	}
+
+	public function getUserId(): string {
```
PVince81, line 86: rename to `getUrlCredential` for consistency ? everywhere you used it you always stored it in a var `getUrlCredential`. or alternatively rename those vars to `$userId` ?

PVince81 (approved):
Some minor picky stuff.

Otherwise this looks good from the code perspective, I haven't tested :+1:

lib/private/Security/SignedUrl/Verifier.php
DeepDiver1975: I honestly did not have a look at our Crypto class :see_no_evil:

core/Controller/CloudController.php
DeepDiver1975: $preConsition is set to null - I'll double check if this works ....

DeepDiver1975:
codecov is nuts .... most likly because of the rebase .....

C0rby (approved):
:+1: Looks good to me.

core/Command/Security/CreateSignKey.php
```diff
@@ -0,0 +1,89 @@
...
+		parent::__construct();
+		$this->userManager = $userManager;
+		$this->config = $config;
+		$this->secureRandom = $secureRandom;
+	}
+
+	protected function configure() {
+		$this
+			->setName('security:sign-key:create')
+			->setDescription('Create and recreate a users sign key for signed urls')
```
PVince81, line 59: "Create or recreate a users signing key"

"sign key" doesn't sound good, I believe it should be "signing key" as it's used for signing

core/Command/Security/CreateSignKey.php
```diff
@@ -0,0 +1,89 @@
...
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$uid = $input->getArgument('user');
+		$user = $this->userManager->get($uid);
+		if ($user === null) {
+			$output->write('User unknown');
+			return 1;
+		}
+		$signingKey = $this->config->getUserValue($uid, 'core', 'signing-key', null);
+		if ($signingKey !== null) {
+			$output->writeln('This user already has a sign key. Recreating the key will invalidate all existing signed urls.');
```
PVince81, line 76: "signing key"

lib/private/Security/SignedUrl/Verifier.php
PVince81: minor thing, never mind

lib/private/Security/SignedUrl/Verifier.php
PVince81: added above, marking as resolved

PVince81 (approved):
All fine, just some minor text change :+1:

micbar:
@DeepDiver1975 Why was this merged without a changelog?

PR #37636, opened 2020-07-02
[Tests-only] Added crude debug output for requests

PVince81:
### Todos

- [ ] better looking output
- [ ] also include username/password as they are not visible in the headers
- [ ] add env variable switch for when debugging must be enabled
- [ ] some other code paths might be skipping the current "sendRequest"

@phil-davis feel free to take over, this is what I used locally to debug many tests.


phil-davis:
See #37897

PR #37638, opened 2020-07-03
[Tests-Only] Disable grace_period popup by default in acceptance testing

phil-davis:
## Description
The grace period popup appears in the webUI of an admin when an enterpise app is enabled and there is not yet any enterprise key. It interferes with normal running of UI acceptance tests that are run with enterprise apps.

We have disabled it in app testing scripts for various enterprise apps. Yesterday it became annoying again - https://github.com/owncloud-docker/server/pull/190 - when building and testing the server docker image for 10.5.0RC3 the UI tests broke because of this. The grace period popup was disabled in the scripts for that docker build process.

Rather than have to do this popup disable in "random" places were the test infrastructure is used in an "enterprise" environment, disable the popup by default in all acceptance test runs. If some day we add tests for the popup, then it can easily be enabled in that particular test scenario(s),

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
Note:

1) I will port this to 10.5.0 branch also
2) this will allow us to remove the `grace_period.demo_key.show_popup` special code from the server docker scripts for 10.5.0RC4
3)  and we can remove the `grace_period.demo_key.show_popup` special code from the app testing scripts of various enterprise apps.

individual-it (approved):

PR #37643, opened 2020-07-03
When locking a resource and no owner is given via the http request th…

DeepDiver1975:
## Description
When locking a resource and no owner is given via the http request the current users displayname is set as lock owner

## How Has This Been Tested?
- :hand: 

Screenshot from 2020-07-03 10-02-05

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Files/FileLocksBackend.php
+++ b/apps/dav/lib/Files/FileLocksBackend.php
@@ -23,6 +23,7 @@ namespace OCA\DAV\Files;
 
 use OCA\DAV\Connector\Sabre\Node;
 use OCP\Files\Storage\IPersistentLockingStorage;
+use OCP\IUserSession;
 use OCP\Lock\Persistent\ILock;
 use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\Locks;
@@ -42,12 +43,17 @@ class FileLocksBackend implements BackendInterface {
 	private $timeFactory;
 	/** @var bool */
 	private $isPublicEndpoint;
+	/**
+	 * @var IUserSession
+	 */
+	private $userSession;
 
-	public function __construct($tree, $useV1, $timeFactory, $isPublicEndpoint = false) {
+	public function __construct($tree, $useV1, $timeFactory, IUserSession $userSession, $isPublicEndpoint = false) {
 		$this->tree = $tree;
 		$this->useV1 = $useV1;
 		$this->timeFactory = $timeFactory;
 		$this->isPublicEndpoint = $isPublicEndpoint;
+		$this->userSession = $userSession;
 	}
 
 	/**
@@ -178,6 +184,10 @@ class FileLocksBackend implements BackendInterface {
 		$storage = $node->getFileInfo()->getStorage();
 		if (!$storage->instanceOfStorage(IPersistentLockingStorage::class)) {
 			return false;
+		}
+		if ($lockInfo->owner === null) {
+			$user = $this->userSession->getUser();
+			$lockInfo->owner = $user->getDisplayName();
 		}
 
 		/** @var IPersistentLockingStorage $storage */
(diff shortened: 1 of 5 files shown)


phil-davis:
I  made a "backport" to 10.5.0 in PR #37645 so it will be ready-to-merge (assuming @jvillafanez reviews and approves)

jvillafanez:
Is it standard that the lock is owned by whoever is requesting it if no owner is set? If not, we likely need to include a comment to clarify that it's our own policy (and not a bug)

jvillafanez:
There are several options and I think it's important to clarify somehow the policy we'll take:
* If no owner is set -> lock can be unlock by anyone having the lock token. Lock owner remains unknown
* If no owner is set -> lock cannot be unlocked (only the lock owner should be able to unlock it using the lock token), but it will expire. Lock owner remains unknown.
* If no owner is set -> use the current user as owner (this PR)
* If no owner is set -> fail the operation and require an owner

For the cases where the lock owner is unknown, it would be the frontend responsibility to provide a valid owner.

I think those are valid policies that could be applied, so it's important to clarify what is the one we want to use to prevent unwanted changes in the future.

DeepDiver1975:
To clarify this a bit: as per WebDav specs this lock owner is an informative value given by requesting webdav client. This information has no technical meaning. It is meant for the end user to have some information to get in touch with the lock owner to chat in real life about unlocking and so on. 
Reasonable values can be the full name, email, phone number as so on.

In addition to this we have a real technical owner in the db which we use to find out who is allowed to unlock - but this is unrelated to this lock owner this PR is talking about.

apps/dav/lib/Files/FileLocksBackend.php
```diff
@@ -179,6 +185,10 @@ public function lock($uri, Locks\LockInfo $lockInfo) {
 		if (!$storage->instanceOfStorage(IPersistentLockingStorage::class)) {
 			return false;
 		}
+		if ($lockInfo->owner === null) {
```
jvillafanez, line 188: Maybe we should include a comment here to ensure and clarify this is the behaviour we want.
As said, there are other possibilities, and we could think that assigning the current user as owner if there is no explicit owner is wrong at some point. A comment would help to clarify this is intended.

jvillafanez (approved):

phil-davis:
Some webUI test scenarios failing: https://drone.owncloud.com/owncloud/core/25736/129/15
e.g.
```
  Scenario: setting a lock shows the display name of a user in the locking details (user has set email address)
    Given these users have been created with skeleton files:
      | username               | displayname   | email       |
      | user-with-display-name | My fancy name | mail@oc.org |
    Given user "user-with-display-name" has locked folder "simple-folder" setting following properties
      | lockscope | shared |
    And user "user-with-display-name" has locked file "data.zip" setting following properties
      | lockscope | exclusive |
    When the user re-logs in with username "user-with-display-name" and password "%regular%" using the webUI
    And folder "simple-folder" should be marked as locked by user "My fancy name (mail@oc.org)" in the locks tab of the details panel on the webUI
    And file "data.zip" should be marked as locked by user "My fancy name (mail@oc.org)" in the locks tab of the details panel on the webUI
```

It used to somehow show the email address (if any) in the lock owner, as well as the user name/display name.

Actually maybe that behavior was not so desirable anyway? Semi-public information disclosure of the lock owner's email address?

So I could just adjust the test scenarios to remove the email address from the expect result.

phil-davis:
The previous behavior came from https://github.com/owncloud/core/blob/master/lib/private/Lock/Persistent/LockManager.php#L75
```
		$owner = isset($lockInfo['owner']) ? $lockInfo['owner'] : null;
		if ($owner === null && $this->userSession->isLoggedIn()) {
			$user = $this->userSession->getUser();
			if ($user !== null) {
				$owner = $user->getDisplayName();
				if ($user->getEMailAddress() !== null) {
					$owner .= " ({$user->getEMailAddress()})";
				}
			}
		}
```

If the user was considered properly `isLoggedIn` then the displayname (and email if exists) was set as the lock owner text.

That logic is still there. This PR effectively overrides it because it never passes a null lock owner text.

DeepDiver1975:
Needs a second debug session ....

DeepDiver1975:
PR coming ....

DeepDiver1975:
replaced by https://github.com/owncloud/core/pull/37647

PR #37645, opened 2020-07-03
[10.5.0] When locking a resource and no owner is given via the http request th…

phil-davis:
…e current users displayname is set as lock owner

Backport #37643 to `release-10.5.0`


phil-davis:
#37643 is now closed and the approach is being looked at again. So this pre-emptive backport was too hopeful! Closing.

PR #37647, opened 2020-07-03
The response of LOCK shall hold the updated owner set by the LockManager

DeepDiver1975:
##  Description
In case the owner is updated by the server within the  LOCK operation the updated owner value needs to be sent back to the client

## How Has This Been Tested?
- :hand: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Files/FileLocksBackend.php
+++ b/apps/dav/lib/Files/FileLocksBackend.php
@@ -192,6 +192,7 @@ class FileLocksBackend implements BackendInterface {
 
 		// in case the timeout has not been accepted, adjust in lock info
 		$lockInfo->timeout = $lock->getTimeout();
+		$lockInfo->owner = $lock->getOwner();
 
 		return !empty($lock);
 	}
--- a/changelog/unreleased/37460
+++ b/changelog/unreleased/37460
@@ -1,3 +1,4 @@
 Change: Add file action to lock a file
 
 https://github.com/owncloud/core/pull/37460
+https://github.com/owncloud/core/pull/37647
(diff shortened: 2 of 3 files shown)


lib/private/Lock/Persistent/LockManager.php
```diff
@@ -80,6 +80,7 @@ public function lock($storageId, $internalPath, $fileId, array $lockInfo) {
 					$owner .= " ({$user->getEMailAddress()})";
 				}
 			}
+			$lockInfo['owner'] = $owner;
```
phil-davis, line 83: I don't understand how this helps/changes anything. `$lockInfo['owner']` is not referenced in any later code of this function. And `$lockInfo` array is not inserted into anything, passed to anything or returned. And `$lockInfo` was nit passed by reference.

DeepDiver1975: The array is passed in by reference as well as the lock info object

phil-davis: I don't see any `&` in the parameter passing. But maybe I am out-of-date with my thinking about how PHP behaves by default when passing arrays.

DeepDiver1975: LOL - there is code missing - I see your point now :see_no_evil:

DeepDiver1975: ... and the wrong file .... Friday ....

phil-davis (approved):
LGTM now

PR #37656, opened 2020-07-06
[tests-only] test the API response of a share-update response

individual-it:
## Description
test response fields when changing shares

## Related Issue
tests for https://github.com/owncloud/ocis-reva/issues/332

## Motivation and Context
make sure all values are reported and correct

## How Has This Been Tested?
https://github.com/owncloud/ocis-reva/pull/353

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature
+++ b/tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature
@@ -5,6 +5,87 @@ Feature: update a public link share
     Given using OCS API version "1"
     And user "Alice" has been created with default attributes and skeleton files
 
+  @skipOnOcis @issue-ocis-reva-243 @issue-ocis-reva-349 @issue-37653
+  #after fixing all the issues merge this scenario with the one below
+  Scenario Outline: API responds with a full set of parameters when owner changes the expireDate of a public share
+    Given using OCS API version "<ocs_api_version>"
+    When user "Alice" creates a public link share using the sharing API with settings
+      | path | FOLDER |
+    And user "Alice" updates the last share using the sharing API with
+      | expireDate | +3 days |
+    Then the OCS status code should be "<ocs_status_code>"
+    And the OCS status message should be ""
+    And the HTTP status code should be "200"
+    And the fields of the last response to user "Alice" should include
+      | id                         | A_STRING             |
+      | share_type                 | public_link          |
+      | uid_owner                  | %username%           |
+      | displayname_owner          | %displayname%        |
+      | permissions                | read                 |
+      | stime                      | A_NUMBER             |
+      | parent                     |                      |
+      | expiration                 | A_STRING             |
+      | token                      | A_STRING             |
+      | uid_file_owner             | %username%           |
+      | displayname_file_owner     | %displayname%        |
+      | additional_info_owner      |                      |
+      | additional_info_file_owner |                      |
+      | item_type                  | folder               |
+      | item_source                | A_STRING             |
+      | path                       | /FOLDER              |
+      | mimetype                   | httpd/unix-directory |
+      | storage_id                 | A_STRING             |
+      | storage                    | A_NUMBER             |
+      | file_source                | A_STRING             |
+      | file_target                | /FOLDER              |
+      | mail_send                  | 0                    |
+      | name                       |                      |
+    Examples:
+      | ocs_api_version | ocs_status_code |
+      | 1               | 100             |
+      | 2               | 200             |
+
+  @skipOnOcV10 @issue-ocis-reva-243 @issue-ocis-reva-349 @issue-37653
+  #after fixing all the issues merge this scenario with the one above
+  Scenario Outline: API responds with a full set of parameters when owner changes the expireDate of a public share
+    Given using OCS API version "<ocs_api_version>"
+    When user "Alice" creates a public link share using the sharing API with settings
+      | path | FOLDER |
+    And user "Alice" updates the last share using the sharing API with
+      | expireDate | +3 days |
+    Then the OCS status code should be "<ocs_status_code>"
+    And the OCS status message should be "OK"
+    And the HTTP status code should be "200"
+    And the fields of the last response to user "Alice" should include
+      | id                         | A_STRING             |
+      | share_type                 | public_link          |
+      | uid_owner                  | %username%           |
+      | displayname_owner          | %displayname%        |
+      | permissions                | read                 |
+      | stime                      | A_NUMBER             |
+      | parent                     |                      |
+      | expiration                 | A_STRING             |
+      | token                      | A_STRING             |
+      | uid_file_owner             | %username%           |
+      | displayname_file_owner     | %displayname%        |
+      | additional_info_owner      |                      |
+      | additional_info_file_owner |                      |
+      | state                      | 0                    |
+      | item_type                  | folder               |
+      | item_source                | A_STRING             |
+      | path                       | /FOLDER              |
+      | mimetype                   | httpd/unix-directory |
+      | storage_id                 | A_STRING             |
+      | storage                    | A_NUMBER             |
+      | file_source                | A_STRING             |
+      | file_target                | /FOLDER              |
+      | mail_send                  | 0                    |
+      | name                       |                      |
+    Examples:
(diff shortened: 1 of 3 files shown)


individual-it:
@PVince81 these tests show the differences in the responses
1. should we add tests to check the responses when other parameters get changed
2. some fields are not shown in oC10. Which empty fields are mandatory and need to be checked?  e.g. "name" is not reported in oC10 when changing a user share, but it also does not mean anything for user-shares, is it a mandatory field?

PVince81:
> should we add tests to check the responses when other parameters get changed

would be good to have at least one test that changes all params and checks that the values got set, yes

> Which empty fields are mandatory and need to be checked

not sure, need to compared with OC 10.

the "name" field only exists for public shares and I'd expect it to always be returned with public shares

individual-it:
> would be good to have at least one test that changes all params and checks that the values got set, yes

we have those further down in the file, this is about the immediate response after changing the parameter

> the "name" field only exists for public shares and I'd expect it to always be returned with public shares

ocis reports it also for user-shares, is that a bug in OCIS in oc10, or fine in both cases?

That would also be a use-case for the new tags as described in https://github.com/owncloud/ocis-reva/issues/329#issuecomment-654044911

PVince81:
> ocis reports it also for user-shares

then it's a bug there

tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature
```diff
@@ -5,6 +5,89 @@ Feature: update a public link share
...
+      | additional_info_owner      |                      |
+      | additional_info_file_owner |                      |
+      | item_type                  | folder               |
+      | item_source                | A_STRING             |
+      | path                       | /FOLDER              |
+      | mimetype                   | httpd/unix-directory |
+      | storage_id                 | A_STRING             |
+      | storage                    | A_NUMBER             |
+      | file_source                | A_STRING             |
+      | file_parent                | A_STRING             |
```
PVince81, line 40: we remove file_parent from everywhere, see https://github.com/owncloud/ocis-reva/issues/326

tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature
```diff
@@ -5,6 +5,89 @@ Feature: update a public link share
...
+      | uid_owner                  | %username%           |
+      | displayname_owner          | %displayname%        |
+      | permissions                | read                 |
+      | stime                      | A_NUMBER             |
+      | parent                     |                      |
+      | expiration                 | A_STRING             |
+      | token                      | A_STRING             |
+      | uid_file_owner             | %username%           |
+      | displayname_file_owner     | %displayname%        |
+      | additional_info_owner      |                      |
```
PVince81, line 31: see https://github.com/owncloud/ocis-reva/issues/349 for OCIS

tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature
```diff
@@ -5,6 +5,89 @@ Feature: update a public link share
...
+      | item_source                | A_STRING             |
+      | path                       | /FOLDER              |
+      | mimetype                   | httpd/unix-directory |
+      | storage_id                 | A_STRING             |
+      | storage                    | A_NUMBER             |
+      | file_source                | A_STRING             |
+      | file_parent                | A_STRING             |
+      | file_target                | /FOLDER              |
+      | mail_send                  | 0                    |
+      | name                       |                      |
```
PVince81, line 43: this should not be there, add a check to enforce its absence ?

tests/acceptance/features/apiSharePublicLink2/updatePublicLinkShare.feature
```diff
@@ -5,6 +5,89 @@ Feature: update a public link share
...
+      | additional_info_file_owner |                      |
+      | item_type                  | folder               |
+      | item_source                | A_STRING             |
+      | path                       | /FOLDER              |
+      | mimetype                   | httpd/unix-directory |
+      | storage_id                 | A_STRING             |
+      | storage                    | A_NUMBER             |
+      | file_source                | A_STRING             |
+      | file_parent                | A_STRING             |
+      | file_target                | /FOLDER              |
```
PVince81, line 41: we remove file_parent from everywhere, see https://github.com/owncloud/ocis-reva/issues/326

tests/acceptance/features/apiShareUpdate/updateShare.feature
```diff
@@ -239,6 +239,98 @@ Feature: sharing
...
+    Then the fields of the last response to user "Alice" sharing with user "Brian" should include
+      | id                         | A_STRING             |
+      | share_type                 | user                 |
+      | uid_owner                  | %username%           |
+      | displayname_owner          | %displayname%        |
+      | permissions                | all                  |
+      | stime                      | A_NUMBER             |
+      | parent                     |                      |
+      | expiration                 |                      |
+      | token                      |                      |
```
PVince81, line 263: 2token" and "name" should not be there, add a check to enforce its absence ?

tests/acceptance/features/apiShareUpdate/updateShare.feature
```diff
@@ -239,6 +239,98 @@ Feature: sharing
...
+      | storage_id                 | home::Alice          |
+      | storage                    | A_STRING             |
+      | file_source                | A_STRING             |
+      | file_parent                | A_NUMBER             |
+      | file_target                | /Alice-folder        |
+      | share_with                 | %username%           |
+      | share_with_displayname     | %displayname%        |
+      | share_with_additional_info |                      |
+      | mail_send                  | 0                    |
+      | attributes                 |                      |
```
PVince81, line 281: let's keep enforcing its existence.
will come later with OCIS https://github.com/owncloud/ocis-reva/issues/350 but I think OCIS should already return it empty

individual-it:
@PVince81 I've adjusted the tests according to your suggestions, mostly added more tags

tests/acceptance/features/apiShareUpdate/updateShare.feature
```diff
@@ -239,6 +239,99 @@ Feature: sharing
...
+    And the fields of the last response should not include
+      | name |  |
+    Examples:
+      | ocs_api_version | ocs_status_code |
+      | 1               | 100             |
+      | 2               | 200             |
+
+  @skipOnOcV10 @issue-ocis-reva-243 @issue-ocis-reva-349 @issue-ocis-reva-350 @issue-ocis-reva-352 @issue-37653
```
PVince81, line 288: add the tag "to be implemented in ocis" ?

individual-it: is https://github.com/owncloud/ocis-reva/issues/329 finalised?

PVince81: yeah, it looks fine

individual-it: @PVince81 I've added the new tags  `@toFixOnOCIS @toFixOnOcV10` (both because there is also an issue in oc10)
please review again

tests/acceptance/features/apiShareUpdate/updateShare.feature
PVince81: I'd consider an empty value not the same as "completely absent from the XML / JSON structure".

Did you intend to leave this empty or missed this bit ? I'm also ok adding this in a separate PR if this is more complicated.

Maybe the meta value could be "IS_ABSENT"

PVince81 (commented):
See one comment above

tests/acceptance/features/apiShareUpdate/updateShare.feature
individual-it: this demonstrates the current behaviour
after fixing it we can add check to ensure its absents with
`And the fields of the last response should not include`
see https://github.com/owncloud/core/pull/37656/files#diff-19e0cca58fe712dccb26dc265fc820b7R281

I've added a commented line with the token to hint that we want to use that line after fixing the issue

PVince81 (approved):
:+1:

PR #37696, opened 2020-07-16
Update 37438

HanaGemela:
To make clear that also brute_force_protection app is needed


jnweiger (approved):
LGTM

karakayasemi:
The new events are generic events listenable from any of the apps. Brute force protection app will use them, but I am not sure we need to specificially mention about brute force protection app in related changelog item.

jnweiger (approved):
LGTM

PR #37706, opened 2020-07-19
[Tests-Only] Fix replacement usernames

phil-davis:
## Description
When I do an acceptance test run with replacement usernames I get various test fails because some recently-added scenarios are not abstracting the username in the required places. So, for example, "Alice" is replaced with username "000" in PR #37506 but some `Then` steps fail because they expect hard-coded "Alice" when they should be expecting `%username%` which will be substituted at run-time for the actual replaced username.

- Fix these so that they work properly when replacing the username.
- Add some log output at the beginning of a test run when `REPLACE_USERNAMES` is defined. Show the username and user attribute replacements so that we know what was happening when we later lok at the log output.

## How Has This Been Tested?
CI in PR #37506 passes

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiFederation1/federated.feature
+++ b/tests/acceptance/features/apiFederation1/federated.feature
@@ -65,14 +65,14 @@ Feature: federated
     When user "Brian" gets the list of pending federated cloud shares using the sharing API
     Then the OCS status code should be "<ocs-status>"
     And the HTTP status code should be "200"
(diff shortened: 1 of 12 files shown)


dpakach (approved):
LGTM :+1:

PR #37711, opened 2020-07-20
[10.5.0] Ignore enterprise_key app when checking for apps to default_enable

phil-davis:
## Description
If someone upgrades to 10.5.0 and then, after putting the 10.5.0 code in place, they copy back their non-core apps from the previous installation, and that includes an recent but old version of the `enterprise_key` app, then the `enterprise_key` app will not have any core `max_version` set, and it will have `default_enable` in its `info.xml`. So the `enterprise_key` app will get auto-enabled.

That is unfortunate in 10.5.0!

This PR teaches the server code about the special-case of the `enterprise_key` app.

1) it will now be ignored when calculating the "shipped apps" (which include apps that have `default_enable`)
2) the `occ app:enable` command will also refuse to enable it (e.g. if someone puts an old version in their `apps` dir and then manually tries to `occ app:enable enterprise_key`)

## How Has This Been Tested?
```
$ php occ a:e enterprise_key

In app.php line 390:
                                                                                 
  App enterprise_key can't be enabled. It is not used from core 10.5.0 onwards.  
                                                                                 

app:enable [-g|--groups GROUPS] [--] <app-id>
```

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/10.5.0_2020-07-09/36814
+++ b/changelog/10.5.0_2020-07-09/36814
@@ -12,3 +12,4 @@ key from there, and it will take into account whether the config.php is read-onl
 You can also enter a license key from the grace period popup.
 
 https://github.com/owncloud/core/pull/36814
+https://github.com/owncloud/core/pull/37711
--- a/lib/private/Installer.php
+++ b/lib/private/Installer.php
@@ -447,6 +447,12 @@ class Installer {
 			if ($dir = \opendir($app_dir['path'])) {
 				$nodes = \scandir($app_dir['path']);
 				foreach ($nodes as $filename) {
+					// Since core 10.5.0, enterprise_key app is no longer used
+					// Be sure not to accidentally enable it if it has been
+					// left behind in an apps dir.
+					if ($filename === 'enterprise_key') {
+						continue;
+					}
 					if (\substr($filename, 0, 1) != '.' and \is_dir($app_dir['path']."/$filename")) {
 						if (\file_exists($app_dir['path']."/$filename/appinfo/info.xml")) {
 							if (!Installer::isInstalled($filename)) {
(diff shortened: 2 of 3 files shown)


phil-davis:
These seem to be the 2 places to protect against accidentally enabling `enterprise_key` in core from 10.5.0 onward. It should stop the `enterprise_key` app from getting accidentally enabled.

Any better ideas are welcome. core needs to have special knowledge about this "old" app to avoid it. So I'm not sure how to make the "nicer"/"more generic"

phil-davis:
What is the chance for unit tests in this code?
Somebody give me clues please.

jnweiger (changes requested):
We have an alternative solution to mitigate the situation addressed here.
https://github.com/owncloud/enterprise_key/pull/132
@phil-davis please review there and check, if that completeley this PR in part, or fully.

phil-davis:
@jnweiger the "empty" enterprise_key app will be fine, as long as the site actually has that new version of enterprise_key app. If it is distributed in the "app bundle" with 10.5.0 tarballs, then admins will get it.

But what happens if an admin installs some old version of the enterprise_key app like 0.2.0 which does not have max_version 10.4 and enables it, and ignores that it not the latest? Do we want the server to protect against installation of old versions of apps that should be "gone"?

jnweiger:
Valid point. So the dummy app does not really replace this. It amends it nicely.
I propose to build RC6 including this and the dummy app.

jnweiger (approved):
Thanks for the clarifications. Approved!

phil-davis:
Merged. IMO RC6 can be built when you have the "dummy" enterprise_key app ready.

PR #37715, opened 2020-07-21
Change cache prefix only with server upgrades but not installing apps

jvillafanez:
Upgrading the OC server will clear all the caches and change the cache
prefix. This commit will add the "clear" step after a successful upgrade
to avoid leaving unused data. Changing the cache prefix was already
being done in this case.

Installing a new app won't make the cache prefix to change. This fixes a
problem installing an app using openID auth. The openID app uses a
distributed cache whose prefix was changing after installing a new app
through the market app, causing the user to logout due to wrong or
missing info (coming from the cache)

Apps are now responsible to clear their caches after the app is
upgraded, if applicable. The apps can also decide not to clean the cache

## Related Issue
https://github.com/owncloud/appliance/issues/80
https://github.com/owncloud/enterprise/issues/4089

## How Has This Been Tested?
Linked issue hasn't been tested yet. However, I've verified the expected behaviour:
* Installing an new app through the market won't change the cache prefix. This should fix the linked issue
* Upgrading core will clear all the caches. This is new behaviour, but anyway the old caches weren't used any longer due to the change in the prefix, so it shouldn't cause new problems. Cache prefix will still be changed.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37715
+++ b/changelog/unreleased/37715
@@ -1,0 +1,16 @@
+Bugfix: Fix problem with the market app installing an app using OpenIDConnect
+
+The OpenIDConnect app uses an in-memory cache to store an OpenID session
+in order to avoid hitting the OpenID provider too much.
+After an app was installed, the prefix used to store information in the cache was
+changing. This was causing problems because the OpenIDConnect app thought the
+OpenID session was no longer valid and, as a consequence, it was logging out the
+user. In practice, installing an new app with the market app having logged in via
+OpenIDConnect would cause the user to logout.
+
+Now, the cache prefix only changes after an ownCloud upgrade. Installing an app
+won't cause the cache prefix to change, so this will fix the problem.
+The OpenIDConnect app will still find the stored session information after the
+new apps are installed.
+
+https://github.com/owncloud/core/pull/37715
--- a/core/Command/Upgrade.php
+++ b/core/Command/Upgrade.php
@@ -33,6 +33,7 @@ use OC\Console\TimestampFormatter;
 use OC\Updater;
 use OCP\IConfig;
 use OCP\ILogger;
+use \OCP\ICacheFactory;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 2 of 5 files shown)


mmattel:
@jvillafanez you have added `3-To Review` but added no reviewer...
It took the suggestion from git and added @phil-davis  (I hope you do not mind :smile: )

mmattel:
You need to add a changelog, CI will complain about this

jvillafanez:
Pinging @voroyam to deploy the fix in the test instance to ensure the PR fixes the problem there.

> @jvillafanez you have added 3-To Review but added no reviewer...

You're too fast :smile: I was expecting broken tests in the CI, but it doesn't seem so.

micbar (approved):

voroyam:
patch installed. Works.

voroyam:
https://cloud.damken.com/index.php/s/HxOG2bTCijVKFjJ

voroyam:
Great work @jvillafanez

VicDeo:
Beware that `occ upgrade` is not the only way to upgrade ownCloud. Web upgrade will cause a dirty cache.

jvillafanez:
> Beware that occ upgrade is not the only way to upgrade ownCloud. Web upgrade will cause a dirty cache.

Agreed. It shouldn't be a problem because the cache will be swapped, but we should clean the old cache both not to leave remnants around and to keep consistency with what the console does.

jvillafanez:
https://github.com/owncloud/core/pull/37715/commits/a822cd4140273006afabfc7de8235cbe1631d630 works with the web UI

There is also the `OccController` but I think it calls the occ command, so it shouldn't be required to touch it (occ command already clears the cache in this PR). Other than that, I think there isn't any additional entrypoint for the upgrade.

VicDeo:
> There is also the OccController but I think it calls the occ command,

true. There is no need to worry about `OccController` - it just a web wrapper for a limited set of CLI commands

VicDeo (approved):
:see_no_evil:

jvillafanez:
rebased

phil-davis:
The code here was reviewed and manual tests done back in July. IMO there is nothing in the rebase or in other recent core changes that should stop this from working. I would be happy for this to be merged. QA will sort out some manual verification again anyway during QA  for 10.6.0

PR #37716, opened 2020-07-21
[Tests-Only] Expected failures demo

phil-davis:
## Description
This is a demonstrator for defining a list of scenarios that are expected to fail.

- 1st commit purposely breaks some scenarios in `apiComments` and `apiFavorites` suites, and lists those in ` tests/acceptance/expected-failures.txt`
- 2nd commit defines `EXPECTED_FAILURES_FILE` in drone so that the failures will be checked to make sure exactly the expected failures have - no more and no less
- 3rd commit provides the code in ` tests/acceptance/run.sh` to do the checking

The file that lists the expected failures should list just the suite name followed by the feature and line number of each scenario (or scenario outline example) that is expected to fail.

Lines starting with `#` are comments and are ignored.

For example:
```
apiComments/comments.feature:9
# write a comment
apiComments/comments.feature:38
apiFavorites/favorites.feature:28
apiFavorites/favorites.feature:29
```

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37717, opened 2020-07-21
[Tests-Only] Add EXPECTED_FAILURES_FILE capability to acceptance tests run.sh

phil-davis:
## Description
Add a feature to the acceptance tests `run.sh` so that the user can define environment variable `EXPECTED_FAILURES_FILE` to point to a file that lists the expected failures.

This is useful when running the acceptance test suite against a new implementation like OCIS or reva where all the tests do not yet pass. 

The `EXPECTED_FAILURES_FILE` should list just the suite name followed by the feature and line number of each scenario (or scenario outline example) that is expected to fail.

Lines starting with `#` are comments and are ignored.

For example:
```
apiComments/comments.feature:9
# write a comment
apiComments/comments.feature:38
apiFavorites/favorites.feature:28
apiFavorites/favorites.feature:29
```

The 2nd commit fixes an existing latent bug in `run.sh` - `TEST_LOG_FILE` was not being created early enough, and when a pipeline run all API tests (usually with some tag selection that reduces the total number of test scenarios selected) then `TEST_LOG_FILE` would be empty. This effects some potential existing error processing/reporting when something has gone wrong, and it effected the processing of actual vs expected failures, because the actual failures were not captured in `TEST_LOG_FILE`

The 3rd commit makes sure that we do not bother to rerun scenarios that failed and were expected to fail.

See PR #37716 for a demonstration of this as it could be used in core. (Note: there is not actually likely to be a use-case for it in core)

Example PR https://github.com/owncloud/ocis-reva/pull/394 shows how it can be used there.

## Related Issues
https://github.com/owncloud/ocis-reva/issues/282 Split old public API webdav tests from new public webdav tests 
we probably will not need to do anything in core - we can just list the known fails in `EXPECTED_FAILURES_FILE`

https://github.com/owncloud/ocis-reva/issues/329 Find untagged tests that need adjusting for OCIS (and the comment at https://github.com/owncloud/ocis-reva/issues/329#issuecomment-654044911 )
we can control all the skip-tagging that currently causes the real day-to-day pain by using `EXPECTED_FAILURES_FILE` and local acceptance tests in OCIS repos that demonstrate different behavior

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -1083,6 +1083,8 @@ export IPV4_URL
 export IPV6_URL
 export FILES_FOR_UPLOAD="${SCRIPT_PATH}/filesForUpload/"
 
+TEST_LOG_FILE=$(mktemp)
+
 if [ ${#BEHAT_SUITES[@]} -eq 0 ] && [ -z "${BEHAT_FEATURE}" ]
 then
 	SUITE_FEATURE_TEXT="all ${TEST_TYPE_TEXT}"
@@ -1103,8 +1105,6 @@ else
 		SUITE_FEATURE_TEXT="${SUITE_FEATURE_TEXT} ${BEHAT_FEATURE_FILE}"
 	fi
 fi
-
-TEST_LOG_FILE=$(mktemp)
 
 for i in "${!BEHAT_SUITES[@]}"
 	do
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -393,7 +393,7 @@ function run_behat_tests() {
 	then
 		echo "Checking expected failures"
 		PASSED=true
-		SOME_SCENARIO_RERUN=false
+		FAILED_SCENARIO_FOUND=false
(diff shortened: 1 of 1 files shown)


individual-it:
when using the rerunning of failed test feature, I think it will also rerun the expected to fail feature

phil-davis:
> when using the rerunning of failed test feature, I think it will also rerun the expected to fail feature

I added code so that, when rerunning scenarios, if the scenario is expected to fail then it is not rerun.

phil-davis:
Stupid drone-docker https://drone.owncloud.com/owncloud/core/26054/46/11
"Error: No such image: docker.io/plugins/codecov:latest"

I will restart CI (or will need to change or rebase this anyway before approval and merge)

individual-it (approved):

PR #37721, opened 2020-07-22
[Tests-Only]Modify functions to check the presence of files and folders

jasson99:
## Description
This PR modifies the function used to assert the presence or absence of files/folders such that it strictly checks for the files and folders as provided in the step.

## Motivation and Context
We had functions to assert that a file/folder exists or doesnot exist. The tests passed for the presence or absence of a file with the same name when the step actually wanted to assert that for a folder, and vice versa. However, sometimes strict assertions are needed.

## How Has This Been Tested?
- https://github.com/owncloud/ocis/pull/419
- https://github.com/owncloud/ocis-reva/pull/396

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/WebDavHelper.php
+++ b/tests/TestHelpers/WebDavHelper.php
@@ -298,7 +298,8 @@ class WebDavHelper {
 	 * @param string $type
 	 * @param int $davPathVersionToUse
 	 *
-	 * @return SimpleXMLElement
+	 * @return SimpleXMLElement|ResponseInterface
+	 * @throws Exception
 	 */
 	public static function listFolder(
 		$baseUrl,
@@ -312,17 +313,25 @@ class WebDavHelper {
 	) {
 		if (!$properties) {
 			$properties = [
-				'getetag'
+				'getetag', 'resourcetype'
 			];
 		}
 		$response = self::propfind(
 			$baseUrl, $user, $password, $path, $properties,
 			$folderDepth, $type, $davPathVersionToUse
 		);
-		$responseXmlObject = HttpRequestHelper::getResponseXml($response);
-		$responseXmlObject->registerXPathNamespace(
-			'ocs', 'http://open-collaboration-services.org/ns'
-		);
+		try {
+			$responseXmlObject = HttpRequestHelper::getResponseXml($response);
+			$responseXmlObject->registerXPathNamespace(
+				'ocs', 'http://open-collaboration-services.org/ns'
+			);
+		} catch (\Exception $e) {
+			if (OcisHelper::isTestingOnOcis()) {
+				return $response;
+			} else {
+				throw new \Exception($e);
+			}
+		}
 		return $responseXmlObject;
 	}
 
--- a/tests/acceptance/features/apiMain/external-storage.feature
+++ b/tests/acceptance/features/apiMain/external-storage.feature
@@ -46,6 +46,7 @@ Feature: external-storage
     And as "Alice" file "/local_storage/foo2/textfile0.txt" should not exist
     And as "Brian" file "/local.txt" should exist
 
+  @issue-37723
   Scenario: Download a file that exists in filecache but not storage fails with 404
     Given user "Alice" has been created with default attributes and skeleton files
     And user "Alice" has created folder "/local_storage/foo3"
@@ -53,7 +54,8 @@ Feature: external-storage
     And file "foo3/textfile0.txt" has been deleted from local storage on the server
     When user "Alice" downloads file "local_storage/foo3/textfile0.txt" using the WebDAV API
     Then the HTTP status code should be "404"
-    And as "Alice" file "local_storage/foo3/textfile0.txt" should not exist
+    And as "Alice" file "local_storage/foo3/textfile0.txt" should exist
+#   And as "Alice" file "local_storage/foo3/textfile0.txt" should not exist
 
   Scenario: Upload a file to external storage while quota is set on home storage
     Given user "Alice" has been created with default attributes and skeleton files
--- a/tests/acceptance/features/bootstrap/TransferOwnershipContext.php
+++ b/tests/acceptance/features/bootstrap/TransferOwnershipContext.php
@@ -190,7 +190,7 @@ class TransferOwnershipContext implements Context {
 	 * @param string $entry
 	 * @param string $path
 	 *
-	 * @return ResponseInterface
+	 * @return void
 	 * @throws \Exception
 	 */
 	public function asFileOrFolderShouldNotExist($user, $entry, $path) {
@@ -202,7 +202,7 @@ class TransferOwnershipContext implements Context {
 			$user, $entry, $this->getLastTransferPath()
 		);
 		$path = $this->getLastTransferPath() . $path;
-		return $this->featureContext->asFileOrFolderShouldNotExist(
+		$this->featureContext->asFileOrFolderShouldNotExist(
 			$user, $entry, $path
 		);
 	}
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -27,6 +27,7 @@ use Guzzle\Http\Exception\BadResponseException;
 use PHPUnit\Framework\Assert;
 use Psr\Http\Message\ResponseInterface;
 use TestHelpers\OcsApiHelper;
+use TestHelpers\OcisHelper;
 use TestHelpers\SetupHelper;
 use TestHelpers\UploadHelper;
 use TestHelpers\WebDavHelper;
@@ -535,9 +536,10 @@ trait WebDav {
 	}
 
 	/**
-	 * @When /^user "([^"]*)" moves (?:file|folder|entry) "([^"]*)"\s?(asynchronously|) to these (?:filenames|foldernames|entries) using the webDAV API then the results should be as listed$/
+	 * @When /^user "([^"]*)" moves (file|folder|entry) "([^"]*)"\s?(asynchronously|) to these (?:filenames|foldernames|entries) using the webDAV API then the results should be as listed$/
 	 *
 	 * @param string $user
+	 * @param string $entry
 	 * @param string $fileSource
 	 * @param string $type "asynchronously" or empty
 	 * @param TableNode $table
@@ -546,6 +548,7 @@ trait WebDav {
 	 */
 	public function userMovesEntriesUsingTheAPI(
 		$user,
+		$entry,
 		$fileSource,
 		$type,
 		TableNode $table
@@ -570,7 +573,7 @@ trait WebDav {
 				"HTTP status code is not the expected value while trying to move " . $targetName
 			);
 			if ($row['exists'] === "yes") {
-				$this->asFileOrFolderShouldExist($user, "entry", $targetName);
+				$this->asFileOrFolderShouldExist($user, $entry, $targetName);
 				// The move was successful.
 				// Move the file/folder back so the source file/folder exists for the next move
 				$this->userMovesFileUsingTheAPI(
@@ -580,7 +583,7 @@ trait WebDav {
 					$fileSource
 				);
 			} else {
-				$this->asFileOrFolderShouldNotExist($user, "entry", $targetName);
+				$this->asFileOrFolderShouldNotExist($user, $entry, $targetName);
 			}
 		}
 	}
@@ -1302,28 +1305,38 @@ trait WebDav {
 	 * @param string $path
 	 * @param string $type
 	 *
-	 * @return ResponseInterface
+	 * @return void
 	 * @throws \Exception
 	 */
 	public function asFileOrFolderShouldNotExist(
-		$user, $entry, $path, $type = "files"
+		$user, $entry="file", $path= null, $type = "files"
 	) {
 		$user = $this->getActualUsername($user);
-		$path = $this->substituteInLineCodes($path, $user);
-		$response = WebDavHelper::makeDavRequest(
-			$this->getBaseUrl(), $this->getActualUsername($user),
-			$this->getPasswordForUser($user), 'GET', $path,
-			[], null, 2, $type
+		$path = $this->substituteInLineCodes($path);
+		$this->responseXmlObject = $this->listFolder(
+			$user, $path, 0, null, $type
 		);
-
-		if ($response->getStatusCode() < 401 || $response->getStatusCode() > 404) {
-			throw new \Exception(
-				"$entry '$path' expected to not exist " .
-				"(status code {$response->getStatusCode()}, expected 401 - 404)"
-			);
+		if (\get_class($this->responseXmlObject) === "GuzzleHttp\Psr7\Response") {
+			$response = $this->responseXmlObject;
+			if ($response->getStatusCode() < 401 || $response->getStatusCode() > 404) {
+				throw new \Exception(
+					"$entry '$path' expected to not exist " .
+					"(status code {$response->getStatusCode()}, expected 401 -404)"
+				);
+			}
+		} else {
+			if ($this->isEtagValid()) {
+				$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
+				if (\count($isCollection) === 0) {
+					$actualResourceType = "file";
+				} else {
+					$actualResourceType = "folder";
+				}
+				Assert::fail(
+					"$entry '$path' should not exist. But it does exist and is a $actualResourceType"
+				);
+			}
 		}
-
-		return $response;
 	}
 
 	/**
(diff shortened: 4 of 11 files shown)


tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1302,28 +1302,29 @@ public function jobStatusValuesShouldMatchRegEx($user, $table) {
...
-			);
+		if($this->isEtagValid()) {
+			if ($entry === "folder") {
+				Assert::assertEquals(count($isCollection), 0, "Unexpectedly, folder `$path` exists");
+			} else {
+				Assert::assertEquals(count($isCollection), 1, "Unexpectedly, file `$path` exists");
+			}
+		}
+		else {
+			return;
```
dpakach, line 1326: I don't think you need to check all this if the file/folder is not expected to exist.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1345,10 +1346,16 @@ public function asFileOrFolderShouldExist(
...
 		);
+		$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
 		Assert::assertTrue(
 			$this->isEtagValid(),
 			"$entry '$path' expected to exist but not found"
 		);
+		if ($entry === "folder") {
+			Assert::assertEquals(count($isCollection), 1, "Unexpectedly, `$path` is not a folder");
+		} else {
+			Assert::isEmpty(count($isCollection), 0, "Unexpectedly, `$path` is not a file");
```
dpakach, line 1357:
```suggestion
			Assert::assertEquals(count($isCollection), 0, "Unexpectedly, `$path` is not a file");
```

phil-davis:
@jasson99 needs code style fixed so we can see the actual test run:
https://drone.owncloud.com/owncloud/core/26047/2/6

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
jasson99: but it should pass when we are checking for the absence of a folder but a file with the same name is found

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1302,28 +1302,28 @@ public function jobStatusValuesShouldMatchRegEx($user, $table) {
...
-				"(status code {$response->getStatusCode()}, expected 401 - 404)"
-			);
+		if ($this->isEtagValid()) {
+			if ($entry === "folder") {
+				Assert::assertEquals(\count($isCollection), 0, "Unexpectedly, folder `$path` exists");
+			} else {
+				Assert::assertEquals(\count($isCollection), 1, "Unexpectedly, file `$path` exists");
+			}
+		} else {
+			return;
```
phil-davis, line 1325: These lines could be removed. The function will exit anyway right after here.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1302,28 +1302,28 @@ public function jobStatusValuesShouldMatchRegEx($user, $table) {
...
-		$path = $this->substituteInLineCodes($path, $user);
-		$response = WebDavHelper::makeDavRequest(
-			$this->getBaseUrl(), $this->getActualUsername($user),
-			$this->getPasswordForUser($user), 'GET', $path,
-			[], null, 2, $type
+		$path = $this->substituteInLineCodes($path);
+		$this->responseXmlObject = $this->listFolder(
+			$user, $path, 0, null, $type
 		);
+		$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
```
phil-davis, line 1316: This line can go inside `if ($this->isEtagValid()) {}`
And then there will be less chance that it will fail completely somehow.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1302,28 +1302,28 @@ public function jobStatusValuesShouldMatchRegEx($user, $table) {
...
-			throw new \Exception(
-				"$entry '$path' expected to not exist " .
-				"(status code {$response->getStatusCode()}, expected 401 - 404)"
-			);
+		if ($this->isEtagValid()) {
+			if ($entry === "folder") {
+				Assert::assertEquals(\count($isCollection), 0, "Unexpectedly, folder `$path` exists");
+			} else {
+				Assert::assertEquals(\count($isCollection), 1, "Unexpectedly, file `$path` exists");
+			}
```
phil-davis, line 1323:
```suggestion
			$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
			if (\count($isCollection) === 0) {
				$actualResourceType = "file";
			} else {
				$actualResourceType = "folder";
			}
			Assert::assertFail(
				"$entry '$path' should not exist. But it does exist and is a $actualResourceType";
			);
```

whenever we get inside `isEtagValid()` then we have to fail. Because if the Etag is valid then the file or folder exists.

Code something like this might work to report what type of resource was expected to not exist, and what type of resource did actually exist.

tests/acceptance/features/bootstrap/WebDav.php
individual-it, line 1316: should be moved into the `if ($this->isEtagValid()) {` block as its only needed in that case

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1345,10 +1345,16 @@ public function asFileOrFolderShouldExist(
 		$this->responseXmlObject = $this->listFolder(
 			$user, $path, 0, null, $type
 		);
+		$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
```
individual-it, line 1348: can be moved after the assert in the next line

individual-it (changes requested):

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/bootstrap/TransferOwnershipContext.php
```diff
@@ -190,7 +190,7 @@ public function downloadedContentWhenDownloadingForUserWithRangeShouldBe(
 	 * @param string $entry
 	 * @param string $path
 	 *
-	 * @return ResponseInterface
+	 * @return void
```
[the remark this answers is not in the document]
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1349,6 +1350,12 @@ public function asFileOrFolderShouldExist(
 			$this->isEtagValid(),
 			"$entry '$path' expected to exist but not found"
 		);
+		$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
+		if ($entry === "folder") {
+			Assert::assertEquals(\count($isCollection), 1, "Unexpectedly, `$path` is not a folder");
+		} else {
```
phil-davis, line 1356:
```suggestion
		} elseif ($entry === "file") {
```

We should only make this assert when the step specifically said `file "abc" should exist'

If the feature file has a step `entry "abc" should exist` then we only care that the entry exists. It can be a file or a folder.

phil-davis:
https://drone.owncloud.com/owncloud/core/26053/55/13
```
  Scenario: Download a file that exists in filecache but not storage fails with 404          # /drone/src/tests/acceptance/features/apiMain/external-storage.feature:49
    Given user "Alice" has been created with default attributes and skeleton files           # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Alice" has created folder "/local_storage/foo3"                                # FeatureContext::userHasCreatedFolder()
    And user "Alice" has moved file "/textfile0.txt" to "/local_storage/foo3/textfile0.txt"  # FeatureContext::userHasMovedFile()
    And file "foo3/textfile0.txt" has been deleted from local storage on the server          # FeatureContext::fileHasBeenDeletedInLocalStorage()
    When user "Alice" downloads file "local_storage/foo3/textfile0.txt" using the WebDAV API # FeatureContext::userDownloadsFileUsingTheAPI()
    Then the HTTP status code should be "404"                                                # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And as "Alice" file "local_storage/foo3/textfile0.txt" should not exist                  # FeatureContext::asFileOrFolderShouldNotExist()
      file 'local_storage/foo3/textfile0.txt' should not exist. But it does exist and is a file
```

The new code for `And as "Alice" file "local_storage/foo3/textfile0.txt" should not exist` is probably finding the file in the file cache.

That last step could be changed to "should exist"

An raise an issue to discuss what should happen. Because I think that  when a client attempts a download of a file, the file is not found, and the client gets a 404, then the server should remove knowledge about the file from the file cache.

At the moment it seems to leave the file entry in the file cache, and someone has to do a scan of the storage to clear the file cache entry.

phil-davis:
https://drone.owncloud.com/owncloud/core/26053/63/13
```
  Scenario Outline: user tries to share a folder with another user when the sharing api has been disabled  # /drone/src/tests/acceptance/features/apiShareManagement/disableSharing.feature:25
    Given using OCS API version "<ocs_api_version>"                                                        # FeatureContext::usingOcsApiVersion()
    When parameter "shareapi_enabled" of app "core" has been set to "no"                                   # AppConfigurationContext::serverParameterHasBeenSetTo()
    Then user "Alice" should not be able to share folder "/FOLDER" with user "Brian" using the sharing API # FeatureContext::userTriesToShareFileUsingTheSharingApi()
    And the OCS status code should be "404"                                                                # OCSContext::theOCSStatusCodeShouldBe()
    And the HTTP status code should be "<http_status_code>"                                                # FeatureContext::thenTheHTTPStatusCodeShouldBe()

    Examples:
      | ocs_api_version | http_status_code |
      | 1               | 200              |
        Unexpectedly, `/FOLDER` is not a file
        Failed asserting that 0 matches expected 1.
      | 2               | 404              |
        Unexpectedly, `/FOLDER` is not a file
        Failed asserting that 0 matches expected 1.
```

I guess that `Then user "Alice" should not be able to share folder "/FOLDER" with user "Brian" using the sharing API` is first checking that the folder exists. Probably the code will need to pass "folder" through to the "should exist" code.

There are other scenarios that  fail in a similar way. Please check the CI results.

This change is "a good thing" - it is finding places where file/folder might be mixed up in test scenarios.

tests/acceptance/features/bootstrap/WebDav.php
jasson99: fixed

tests/acceptance/features/apiMain/external-storage.feature
```diff
@@ -53,7 +53,7 @@ Feature: external-storage
     And file "foo3/textfile0.txt" has been deleted from local storage on the server
     When user "Alice" downloads file "local_storage/foo3/textfile0.txt" using the WebDAV API
     Then the HTTP status code should be "404"
-    And as "Alice" file "local_storage/foo3/textfile0.txt" should not exist
+    And as "Alice" file "local_storage/foo3/textfile0.txt" should exist
```
individual-it, line 56: why should that file now exist?

individual-it (dismissed):
please make sure these changes also work correctly in ocis, ocis-reva and cs3org/reva

tests/acceptance/features/apiMain/external-storage.feature
phil-davis: The code for "should not exist" has changed the way that it detects if the file "exists".

Previously it did a `GET` of the file, which was effectively the same as `When user "Alice" downloads file...` - because the file actual does not exist in the underlying local storage file system, the `GET` returns a 404.

But now, "should not exist" uses a similar request to "should exist" and is just getting the Etag etc properties of the file. That is found in the file cache (database) and gets returned. For that "properties-only" request the server does not need to read the actual file content on disk. So the server never notices that the file content is gone. So the file "does exist".

Actually IMO this makes the test "better" (more readable), after reading the description:
`Scenario: Download a file that exists in filecache but not storage fails with 404`

The file "exists" as far as the ownCloud database is concerned, but cannot be downloaded.

The behavior on OCIS will need to be sorted out, because OCIS is probably better - there will not be the distinction between "the database file cache" and "the real filesystem". So this might become an example of bad behavior in oC10 and correct behavior in OCIS

tests/TestHelpers/WebDavHelper.php
```diff
@@ -298,7 +298,7 @@ public static function proppatchWithMultipleProps(
 	 * @param string $type
 	 * @param int $davPathVersionToUse
 	 *
-	 * @return SimpleXMLElement
+	 * @return SimpleXMLElement|ResponseInterface
```
phil-davis, line 301: This `listFolder` is called by WebDav.php `listFolder` and TrashbinContext.php `listTrashbinFolder`
And those are called by various things. But they seem to be all "positive" checks on a resource that should exist. So maybe they will be OK?

tests/TestHelpers/WebDavHelper.php
```diff
@@ -319,6 +319,9 @@ public static function listFolder(
 			$baseUrl, $user, $password, $path, $properties,
 			$folderDepth, $type, $davPathVersionToUse
 		);
+		if (OcisHelper::isTestingOnOcis()) {
+			return $response;
+		}
 		$responseXmlObject = HttpRequestHelper::getResponseXml($response);
```
phil-davis, line 325: Maybe we should only directly return the response if `isTestingOnOcis` and HttpRequestHelper::getResponseXml fails?

`HttpRequestHelper::getResponseXml()` throws an exception, so we could catch that and if `OcisHelper::isTestingOnOcis()` then `return $response` and otherwise re-throw the exception.

That will allow us to process the `$responseXmlObject` in cases where OCIS does return it.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1315,17 +1316,26 @@ public function asFileOrFolderShouldNotExist(
...
 			$user, $path, 0, null, $type
 		);
-
-		if ($this->isEtagValid()) {
-			$isCollection = $this->getResponseXmlObject()->xpath("//d:prop/d:resourcetype/d:collection");
-			if (\count($isCollection) === 0) {
-				$actualResourceType = "file";
-			} else {
-				$actualResourceType = "folder";
+		if (OcisHelper::isTestingOnOcis()) {
```
phil-davis, line 1319: We could check here if `$this->responseXmlObject` is a `ResponseInterface` 

If we are testing on OCIS but do get a proper XML object back, then it will be good to run the "normal"  code in the `else` block.

jasson99: But listFolder function returns response only when testingOnOcis.

phil-davis: You can modify `listFolder` so that it always tries to get the XML. But if that throws an exception, then check if testing on OCIS, and only then return the whole response.

That way we minimize the change in behavior of the code. If an OCIS call does return some valid XML in the response then we will check it in the same way as we do on oC10. As these sort of issues are fixed on OCIS, the test code will automatically check the XML of OCIS  responses.

jasson99: fixed

tests/TestHelpers/WebDavHelper.php
jasson99: fixed

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1302,28 +1305,38 @@ public function jobStatusValuesShouldMatchRegEx($user, $table) {
...
+		$this->responseXmlObject = $this->listFolder(
+			$user, $path, 0, null, $type
 		);
-
-		if ($response->getStatusCode() < 401 || $response->getStatusCode() > 404) {
-			throw new \Exception(
-				"$entry '$path' expected to not exist " .
-				"(status code {$response->getStatusCode()}, expected 401 - 404)"
-			);
+		if (\gettype($this->responseXmlObject === "ResponseInterface")) {
```
phil-davis, line 1319:
```suggestion
		if (\gettype($this->responseXmlObject) === "ResponseInterface") {
```
IMO the brackets need fixing

jasson99: fixed

phil-davis (commented):
Looking good so far - let's see what CI says in each repo.

phil-davis:
I requested review from myself again so that I see this in my list. @jasson99 please ping when you have got this passing in ocis-reva (or when you get really stuck!)

phil-davis (approved):

PR #37729, opened 2020-07-26
Delete the correct expired share and handle user/group shares on ExpiredSharesJob

karakayasemi:
## Description
Delete the correct expired share by using share id on `ExpireSharesJob`. `ExpireSharesJob` now also handles user and group shares. 

I created the pr based on @pako81 in https://github.com/owncloud/enterprise/issues/4112#issuecomment-662842454.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4112

## Motivation and Context
`ExpiredSharesJobs` currently uses `\OCP\Share::unshare` method for deleting expired shares.
If multiple links exist for the same node and, non-expired share created before expired share, expiredSharesJob currently deletes non-expired share.

This problem resolved with deleting correct expired shares by using share id.

Also, `ExpireSharesJob` now handles user and group shares. If we prefer, I can separate this part of the change to another pr.

## How Has This Been Tested?
The class has an integration test. Also, tried manually.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/lib/FederatedShareProvider.php
+++ b/apps/federatedfilesharing/lib/FederatedShareProvider.php
@@ -878,7 +878,7 @@ class FederatedShareProvider implements IShareProvider {
 	 */
 	private function createShareObject($data) {
 		$share = new Share($this->rootFolder, $this->userManager);
-		$share->setId((int)$data['id'])
+		$share->setId($data['id'])
 			->setShareType((int)$data['share_type'])
 			->setPermissions((int)$data['permissions'])
 			->setTarget($data['file_target'])
--- a/apps/files_sharing/lib/ExpireSharesJob.php
+++ b/apps/files_sharing/lib/ExpireSharesJob.php
@@ -1,6 +1,7 @@
 <?php
 /**
  * @author Roeland Jago Douma <rullzer@owncloud.com>
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
  *
  * @copyright Copyright (c) 2018, ownCloud GmbH
  * @license AGPL-3.0
@@ -22,6 +23,10 @@
 namespace OCA\Files_Sharing;
 
 use OC\BackgroundJob\TimedJob;
+use OC\Share20\DefaultShareProvider;
+use OCP\IDBConnection;
+use OCP\Share\Exceptions\ShareNotFound;
+use OCP\Share\IManager;
 
 /**
  * Delete all shares that are expired
@@ -29,11 +34,37 @@ use OC\BackgroundJob\TimedJob;
 class ExpireSharesJob extends TimedJob {
 
 	/**
-	 * sets the correct interval for this timed job
+	 * @var IManager $shareManager
 	 */
-	public function __construct() {
+	private $shareManager;
+
+	/**
+	 * @var IDBConnection $connection
+	 */
+	private $connection;
+
+	/**
+	 * @var DefaultShareProvider $defaultShareProvider
+	 */
+	private $defaultShareProvider;
+
+	/**
+	 * sets the correct interval for this timed job
+	 *
+	 * @param IManager $shareManager
+	 * @param IDBConnection $connection
+	 * @param DefaultShareProvider $defaultShareProvider
+	 */
+	public function __construct(
+		IManager $shareManager,
+		IDBConnection $connection,
+		DefaultShareProvider $defaultShareProvider
+	) {
 		// Run once a day
 		$this->setInterval(24 * 60 * 60);
+		$this->shareManager = $shareManager;
+		$this->connection = $connection;
+		$this->defaultShareProvider = $defaultShareProvider;
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -68,7 +83,12 @@ public function run($argument) {
 
 		$shares = $qb->execute();
 		while ($share = $shares->fetch()) {
-			\OCP\Share::unshare($share['item_type'], $share['file_source'], \OCP\Share::SHARE_TYPE_LINK, null, $share['uid_owner']);
+			try {
+				$share = $this->shareManager->getShareById('ocinternal:' . $share['id']);
+				$this->shareManager->deleteShare($share);
```
karakayasemi, line 88: Since `getShareById` already checks expiration date and remove expired shares, code never reaches this point. That's why project test coverage decreased. Actually, this line is safe to remove, but I am not sure about the readability of code.

jvillafanez: You could comment this line and also include a comment explaining why this isn't needed.

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -68,7 +83,12 @@ public function run($argument) {
 
 		$shares = $qb->execute();
 		while ($share = $shares->fetch()) {
-			\OCP\Share::unshare($share['item_type'], $share['file_source'], \OCP\Share::SHARE_TYPE_LINK, null, $share['uid_owner']);
+			try {
+				$share = $this->shareManager->getShareById('ocinternal:' . $share['id']);
```
jvillafanez, line 87: Any alternative to avoid using "ocinternal:"? It shouldn't be known from the outside.
If it isn't possible, we should include a comment so we could find and adjust this piece of code in case the identifier changes.

micbar:
@karakayasemi Can you add a changelog plse?

karakayasemi:
Changelog has been added.

apps/files_sharing/lib/ExpireSharesJob.php
karakayasemi: Ideally, we need to inject `\OCP\Share\IProviderFactory` and get the correct share provider for share_type from it. However, IProviderFactory can not auto-injectable and manually creating it a bit tricky. See https://github.com/owncloud/core/blob/6038c8e3f0c24f504e809959680b453345b09b93/lib/private/Server.php#L877-L880

IMHO, for now, using the default share provider is the best option even it has `\OC` namespace.

jvillafanez:
Code looks good :+1:

karakayasemi:
share id is coming as integer in a pgsql database, but, it is coming as string in all the other supported databases. Because of that, `returnValueMap` did not work in Postgres and Postgres tests failed.

The failing tests opened a can of worms about share id inconsistency. Share id is supposed to be a string according to its interface description. However, `IShareProvider` expects share id as integer, in fact, its implementations also cast share id to integer when they setting id to any `IShare` instance. In contrast, `setId` method of `Share` implementation casts incoming integer argument to string. 
https://github.com/owncloud/core/blob/378d38ab844e048bc88ca7cc365592017f0c7227/lib/private/Share20/Share.php#L95-L97

So, I corrected `IShareProvider`'s `getShareById` method description and removed the unnecessary casting operation from `IShareProvider` implementations.

jvillafanez:
Could you double-check the type of the DB column in postgresql? At least in mysql, I see the "id" as int column.

I guess this works due to how we're using it, not due to it's properly implemented:
* Doctrine is likely fetching an integer, which is stored as string in the object because of the type checking done in the `setId` method of the share.
* We aren't actively setting any id to any share. We're only set the id when we create the object from the DB.
* There are checks in place to prevent changing the id after being set.

My worry is the interaction being the share object and the DB. If the object has a string but the DB requires an integer, it's either Doctrine or the actual DB the one making an automatic type conversion.

apps/federatedfilesharing/lib/FederatedShareProvider.php
```diff
@@ -878,7 +878,7 @@ private function getRawShare($id) {
 	 */
 	private function createShareObject($data) {
 		$share = new Share($this->rootFolder, $this->userManager);
-		$share->setId((int)$data['id'])
```
karakayasemi, line 881: `setId` expects string. Casting the parameter to integer does not make sense in any case here.

lib/public/Share/IShareProvider.php
```diff
@@ -135,11 +135,11 @@ public function getSharesBy($userId, $shareType, $node, $reshares, $limit, $offs
 	 * @since 10.0.0
 	 */
 	public function getAllSharesBy($userId, $shareTypes, $nodeIDs, $reshares);
-	
+
 	/**
 	 * Get share by id
 	 *
-	 * @param int $id
+	 * @param string $id
```
karakayasemi, line 142: I can revert this change and cast incoming share id to an integer when I call this method in `ExpireSharesJob` to ensure consistency between databases.

karakayasemi:
Share id is integer in Postgres like Mysql also. But, when we query this column, DBAL converts integer id's to string on all the supported databases except Postgres, probably due to possible overflow reasons. I checked one more time.  Since `IShare` getId method always returns string, `Phpunit`'s `returnValueMap` method configuration did not work for Postgres.

jvillafanez:
I think this only happens here. I mean, we're using the ids from other tables in other places and it works fine I guess.
I'm fine with the fix, but include a comment explaining the situation, specially mentioning mysql and postgresql, so we can remember this problem in the future and prevent breaking the same thing again. In addition, it would make easier to know what we need to check if we touch something around this.

karakayasemi:
This id string/integer inconcistency more likely happen in other tables but doctrine/dbal handling these cases without problem. I guess, there is no other test case that configures a stub method with id parameter comes from database. That's why we did not notice before.

@jvillafanez, thank you for review. I added a comment that explain why we used type casting.

jvillafanez (approved):

karakayasemi:
Project coverage was not enough. Added a test to increase coverage.

PR #37749, opened 2020-07-29
[tests-only] use usersHaveBeenCreated & delete UserHelper::createUser

individual-it:
## Description
`usersHaveBeenCreated()` is also able to create users through the provisioning API, so let's use it everywhere and delete `UserHelper::createUser()`

## Motivation and Context
deleted code, is debugged code
for https://github.com/owncloud/ocis/pull/409 I also need to upload the skeleton files when not using the LDAP server, that would result in code duplication currently

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/UserHelper.php
+++ b/tests/TestHelpers/UserHelper.php
@@ -32,52 +32,6 @@ use Psr\Http\Message\ResponseInterface;
  */
 class UserHelper {
 	/**
-	 * @param string $baseUrl
-	 * @param string $user
-	 * @param string $password
-	 * @param string $adminUser
-	 * @param string $adminPassword
-	 * @param string $displayName
-	 * @param string $email
-	 *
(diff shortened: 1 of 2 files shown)


phil-davis (approved):
That should work fine. It will put the single user into an array with 1 entry and then do all the HttpRequestHelper::sendBatchRequest stuff to make a "batch" of 1 request.

phil-davis:
https://drone.owncloud.com/owncloud/core/26143/29/10 - some phpunit pipelines seem to have hung. They have run all the tests but drone is not finishing the pipeline step.

Acceptance tests look OK. Some pipelines there are running slow. So maybe some drone agents are struggling.

I will cancel and restart.

PR #37755, opened 2020-07-30
allow storages to explicitly opt-out of uploading .part files

### LINKED ISSUE #25826: Make it possible for storage implementations to disable part files

karakayasemi:
Continuation of https://github.com/owncloud/core/pull/37062
- Commits have been rebased and squashed
- Some unit tests have been added (Hard to write meaningful tests, I just asserted the boolean value according to related Storage class.)

## Description
Allow for storage classes to decide whether to use .part files. Default to using part files to preserve previous behavior.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/25826

## Motivation and Context
Google Drive external storage (and others?) should not create a duplicate file with the same name every time a file is changed.

## How Has This Been Tested?
- no serious testing, just UAT that ownCloud is no longer generating duplicate Google Drive files

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/public/Files/Storage/IStorage.php
+++ b/lib/public/Files/Storage/IStorage.php
@@ -47,6 +47,13 @@ use OCP\Files\StorageNotAvailableException;
 interface IStorage {
 
 	/**
+	 * usePartFile - allows a storage to opt-out of .part upload and rename
+	 * @return boolean
+	 * @since 10.5.1
+	 */
+	public function usePartFile();
+
+	/**
 	 * Get the identifier for the storage,
 	 * the returned id should be the same for every storage object that is created with the same parameters
 	 * and two storage objects with the same id should refer to two storages that display the same files.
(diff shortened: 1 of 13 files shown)


mmattel:
Only a minor question, wouldnt it be better to name it `usePartFile` just to make it more clear what it is about?

lib/public/Files/Storage/IStorage.php
```diff
@@ -46,6 +46,13 @@
  */
 interface IStorage {
 
+	/**
+	 * usePartFile - allows a storage to opt-out of .part upload and rename
+	 * @return boolean
+	 * @since 10.5.0
```
VicDeo, line 52: 10.5.1 ?

VicDeo (approved):
Looks good

PR #37756, opened 2020-07-31
[tests-only] Delete users after test ocis ocs

individual-it:
## Description
for owncloud/ocis#409 we need to create users using the provisioning API also when running tests with OCIS, so we better delete the users at the end of the scenario

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -2365,9 +2365,8 @@ trait Provisioning {
 	public function deleteUser($user) {
 		if (OcisHelper::isTestingOnOcis()) {
 			OcisHelper::deleteRevaUserData($user);
-		} else {
-			$this->deleteTheUserUsingTheProvisioningApi($user);
 		}
+		$this->deleteTheUserUsingTheProvisioningApi($user);
 
 		$this->userShouldNotExist($user);
 	}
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -4114,13 +4113,12 @@ public function afterScenario() {
...
 		if ($this->isTestingWithLdap()) {
 			$this->deleteLdapUsersAndGroups();
-		}
-
-		if (OcisHelper::isTestingOnOcis()) {
-			OcisHelper::deleteRevaUserData("");
 		} else {
 			$this->cleanupDatabaseUsers();
 			$this->cleanupDatabaseGroups();
+		}
```
phil-davis, line 4119: What effect will this have when tests run for the oC10 `user_ldap` app?

Before this change, these functions were being called when running tests for the oC10 `user_ldap` app. Now they will not run. It might effect tests that create local users and groups that are not in LDAP.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -2365,9 +2365,8 @@ public function createUser(
 	public function deleteUser($user) {
 		if (OcisHelper::isTestingOnOcis()) {
 			OcisHelper::deleteRevaUserData($user);
-		} else {
-			$this->deleteTheUserUsingTheProvisioningApi($user);
 		}
+		$this->deleteTheUserUsingTheProvisioningApi($user);
```
phil-davis, line 2369: What will happen in `cs3org/reva` ? I guess that will not have the provisioning API?
Maybe need to use `OcisHelper::isTestingOnReva()` to detect that?

individual-it: its called from https://github.com/owncloud/core/blob/4bc7efe5cb59b0af92986b6b9e2269038b2b4a11/tests/acceptance/features/bootstrap/Provisioning.php#L4150 so only when trying to delete DB users, that should not affect LDAP runs

tests/acceptance/features/bootstrap/Provisioning.php
individual-it: true, I guess running them always will not hurt, just use some time. I've changed it

phil-davis (approved):

PR #37783, opened 2020-08-07
Fix exit codes of security:certificates commands

### LINKED ISSUE #35364: removal of non-existing security certificates by cli doesn't produce error and exit code is still zero
### Steps to reproduce
1. Remove a security certificate that doesn't exist(eg someCertificate.crt)  by following cli  command 
``` sudo -u www-data ./occ security:certificates:remove someCertificate.crt```
2. Now check the exit code as:
``` echo $?```
This returns 0 (zero) and doesn't produce any error

### Expected behaviour
This should give some error like: 
 ``` certificate not found```

### Actual behaviour
Produces no error and exit code is 0 (zero)

### Server configuration
**Operating system**:
Ubuntu 18.04.2 LTS

### PR DESCRIPTION
**Pull request by phil-davis**
## Description
I was getting acceptance test fails when running `cliMain/securityCertificates.feature` against ownCloud10 `latest` (currently 10.5.0 tarball). The `security:certificates:remove` and `security:certificates:import` commands are not exiting with proper error codes when something is wrong, so that makes it impossible for the acceptance test scenarios to react appropriately in steps like `Then the command should have been successful` - that step never fails because the command status is always 0 = success.

The easiest thing to do is first to fix the bug that was already reported in the related issue.

1) When importing a certificate and the certificate file is not found, exit with status 1.

2) When removing a certificate and the certificate does not exist, exit with status 1.

3) Adjust the acceptance tests to have cases for these.

4) Adjust `.drone.star` - it has `security:certificates:import` commands for when tests are using HTTPS and certificates. Those commands were also being done in pipelines where the certificates did not exist and were not needed. The commands were never successful. Now that the commands exit with an error status, the pipelines were failing during the setup. The starlark code has been adjusted so that it only does `security:certificates:import` in the correct cases.

Note 1: I did not find any existing unit tests for commands in ` core/Command/Security`. The commands seem to have been introduced in PR #21370 but that PR had no automated tests.

Note 2: I added some simple unit tests

## Related Issue
- Fixes #35364

## How Has This Been Tested?
Locally running commands:
```
$ php occ security:certificates:remove xyz.crt
certificate not found
$ echo $?
1
$ php occ security:certificates:import xyz.crt
certificate not found
$ echo $?
1
```

and scenarios in acceptance tests.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -750,7 +750,8 @@ def litmus():
 
 	default = {
 		'phpVersions': ['7.2', '7.3', '7.4'],
-		'logLevel': '2'
+		'logLevel': '2',
+		'useHttps': True,
 	}
 
 	if 'defaults' in config:
@@ -800,7 +801,7 @@ def litmus():
 					composerInstall(phpVersion) +
 					vendorbinInstall(phpVersion) +
 					yarnInstall(phpVersion) +
-					installServer(phpVersion, db, params['logLevel']) +
+					installServer(phpVersion, db, params['logLevel'], params['useHttps']) +
 					setupLocalStorage(phpVersion) +
 					fixPermissions(phpVersion, False) +
 					createShare(phpVersion) +
@@ -890,7 +891,7 @@ def litmus():
 					],
 				'services':
 					databaseService(db) +
-					owncloudService(phpVersion, 'server', '/drone/src', True),
+					owncloudService(phpVersion, 'server', '/drone/src', params['useHttps']),
 				'depends_on': [],
 				'trigger': {
 					'ref': [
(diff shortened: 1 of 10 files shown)


phil-davis:
Also see issue #37784 which was the trigger for fixing this.

individual-it (commented):
tests look good to me

phil-davis:
codecov/project https://codecov.io/gh/owncloud/core/pull/37783/diff

The uncovered code is in the `core/Command/Security` code, which did not have any unit tests anyway. I have covered it with acceptance test scenarios.

karakayasemi (approved):

PR #37785, opened 2020-08-07
[Tests-Only] Do not exit acceptance test if a test suite has a failure

### LINKED ISSUE #37776: Acceptance tests exit early if a suite has a failure
When running multiple test suites in a single run, if there is a test fail in one of the suites, the test runs exits and does not run the remaining test suites.

That means that developers do not get a full list of all test failures in a single run. It is nicer to know all the test failures when you start trying to fix something.

See https://github.com/owncloud/ocis-reva/pull/425 for a demo of the current behavior.

kiranparajuli589:
## Description
With this PR:
:heavy_check_mark: unexpected failures are watched and logged at the end of the test run.
:heavy_check_mark: unexpected success are also watched and logged at the end of the test run.
:heavy_check_mark: if a certain test-suite has a failure then remember that unexpected scenario behavior and continue the test.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/37776

## How Has This Been Tested?

- https://github.com/owncloud/ocis-reva/pull/425

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -414,6 +414,14 @@ function run_behat_tests() {
 		fi
 	fi
 
+	FAILED_SCENARIO_PATHS_COLORED=`awk '/Failed scenarios:/',0 ${TEST_LOG_FILE} | grep feature`
+	# There will be some ANSI escape codes for color in the FEATURE_COLORED var.
+	# Strip them out so we can pass just the ordinary feature details to Behat.
+	# Thanks to https://en.wikipedia.org/wiki/Tee_(command) and
+	# https://stackoverflow.com/questions/23416278/how-to-strip-ansi-escape-sequences-from-a-variable
+	# for ideas.
+	FAILED_SCENARIO_PATHS=$(echo "${FAILED_SCENARIO_PATHS_COLORED}" | sed "s/\x1b[^m]*m//g")
+
 	if [ -n "${EXPECTED_FAILURES_FILE}" ]
 	then
 		if [ -n "${BEHAT_SUITE_TO_RUN}" ]
@@ -425,13 +433,6 @@ function run_behat_tests() {
 
 		PASSED=true
 		FAILED_SCENARIO_FOUND=false
-		FAILED_SCENARIO_PATHS_COLORED=`awk '/Failed scenarios:/',0 ${TEST_LOG_FILE} | grep feature`
-		# There will be some ANSI escape codes for color in the FEATURE_COLORED var.
-		# Strip them out so we can pass just the ordinary feature details to Behat.
-		# Thanks to https://en.wikipedia.org/wiki/Tee_(command) and
-		# https://stackoverflow.com/questions/23416278/how-to-strip-ansi-escape-sequences-from-a-variable
-		# for ideas.
-		FAILED_SCENARIO_PATHS=$(echo "${FAILED_SCENARIO_PATHS_COLORED}" | sed "s/\x1b[^m]*m//g")
 
 		# Check that every failed scenario is in the list of expected failures
 		for FAILED_SCENARIO_PATH in ${FAILED_SCENARIO_PATHS}
@@ -490,6 +491,15 @@ function run_behat_tests() {
 		else
 			echo "Failure - actual and expected failures did not match"
 		fi
+	else
+		for FAILED_SCENARIO_PATH in ${FAILED_SCENARIO_PATHS}
+		do
+			SUITE_PATH=$(dirname "${FAILED_SCENARIO_PATH}")
+			SUITE=$(basename "${SUITE_PATH}")
+			SCENARIO=$(basename "${FAILED_SCENARIO_PATH}")
+			SUITE_SCENARIO="${SUITE}/${SCENARIO}"
+			UNEXPECTED_FAILED_SCENARIOS+=("${SUITE_SCENARIO}")
+		done
 	fi
 
 	# With webUI tests, we try running failed tests again.
@@ -1173,7 +1183,7 @@ done
 
 TOTAL_SCENARIOS=$((SCENARIOS_THAT_PASSED + SCENARIOS_THAT_FAILED))
 
-echo -e "\nrunsh: Total ${TOTAL_SCENARIOS} scenarios (${SCENARIOS_THAT_PASSED} passed, ${SCENARIOS_THAT_FAILED} failed)"
+echo "runsh: Total ${TOTAL_SCENARIOS} scenarios (${SCENARIOS_THAT_PASSED} passed, ${SCENARIOS_THAT_FAILED} failed)"
 
 teardown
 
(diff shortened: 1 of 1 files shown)


tests/acceptance/run.sh
```diff
@@ -1167,7 +1170,10 @@ for i in "${!BEHAT_SUITES[@]}"
 
 	if [ "${PASSED}" = false ]
 	then
-		break
+	  UNEXPECTED_FAILURE=true
+		UNEXPECTED_FAILED_SCENARIO_PATHS+=("${FAILED_SCENARIO_PATHS}")
+  else
+    UNEXPECTED_FAILURE=false
```
phil-davis, line 1176: If the next suite passes, then `UNEXPECTED_FAILURE` will get set back to false.
This variable needs to be set to false before looping through the suites.

tests/acceptance/run.sh
```diff
@@ -1187,5 +1193,7 @@ fi
 if [ -n "${EXPECTED_FAILURES_FILE}" ]
 then
 	echo "runsh: Exit code after checking expected failures: ${FINAL_EXIT_STATUS}"
+	echo "runsh: Total failed scenario throughout the test run"
```
phil-davis, line 1196:
```suggestion
	echo "runsh: Total failed scenarios throughout the test run:"
```

tests/acceptance/run.sh
```diff
@@ -1151,6 +1151,9 @@ else
 	fi
 fi
 
+declare -a UNEXPECTED_FAILED_SCENARIO_PATHS
```
phil-davis, line 1154: This could go higher up. Above the `if` at line 1133, so it is declared before `run_behat_tests` at line 1136.
And after `run_behat_tests` at line 1136 we will need to have logic about `if [ "${PASSED}" = false ]` then `UNEXPECTED_FAILURE=true` etc. So that `UNEXPECTED_FAILURE` has the correct value down in line 1186

tests/acceptance/run.sh
```diff
@@ -1134,6 +1137,13 @@ if [ ${#BEHAT_SUITES[@]} -eq 0 ] && [ -z "${BEHAT_FEATURE}" ]
 then
 	SUITE_FEATURE_TEXT="all ${TEST_TYPE_TEXT}"
 	run_behat_tests
+
+	if [ "${PASSED}" = false ]
+	then
+		UNEXPECTED_FAILURE=ture
```
phil-davis, line 1143:
```suggestion
		UNEXPECTED_FAILURE=true
```

tests/acceptance/run.sh
```diff
@@ -1187,5 +1199,7 @@ fi
 if [ -n "${EXPECTED_FAILURES_FILE}" ]
 then
 	echo "runsh: Exit code after checking expected failures: ${FINAL_EXIT_STATUS}"
+	echo "runsh: Total failed scenarios throughout the test run:"
+	printf "%s\n" "${UNEXPECTED_FAILED_SCENARIO_PATHS[@]}"
```
phil-davis, line 1203: There might be no unexpected failures. This could echo and printf could go inside an `if`  statement, maybe something like:
```suggestion
if [ "${UNEXPECTED_FAILURE}" = true ]
then
	echo "runsh: Total failed scenarios throughout the test run:"
	printf "%s\n" "${UNEXPECTED_FAILED_SCENARIO_PATHS[@]}"
else
	echo "runsh: There were no unexpected failures."
fi
```

tests/acceptance/run.sh
```diff
@@ -1164,11 +1174,6 @@ for i in "${!BEHAT_SUITES[@]}"
 				fi
 				run_behat_tests
 			done
-
-	if [ "${PASSED}" = false ]
-	then
-		break
-	fi
```
phil-davis, line 1171: We need to remember if any suite did not pass, so we can use that information at the end. Something like:
```
if [ "${PASSED}" = false ]
then
  UNEXPECTED_FAILURE=true
fi
```

Otherwise, if the last suite passes, then `PASSED` is true and we have no way to know that for one of the suites, `PASSED` was false.

tests/acceptance/run.sh
```diff
@@ -1177,7 +1175,14 @@ echo "runsh: Total ${TOTAL_SCENARIOS} scenarios (${SCENARIOS_THAT_PASSED} passed
 
 teardown
 
-if [ "${PASSED}" = true ]
+if [ "${PASSED}" = false ]
+then
+	UNEXPECTED_FAILURE=true
+else
+	UNEXPECTED_FAILURE=false
+fi
```
phil-davis, line 1183: We should already know the value of `UNEXPECTED_FAILURE` from code that happened just after each call to `run_behat_tests`. We can't set it here, because `PASSED` is only the result of the last suite that was run.

phil-davis: And `UNEXPECTED_FAILURE=false` will need to be initialized somewhere early in the script.

tests/acceptance/run.sh
```diff
@@ -442,6 +444,7 @@ function run_behat_tests() {
 				then
 					echo "Error: Scenario ${SUITE_SCENARIO} failed but was not expected to fail."
 					PASSED=false
+					UNEXPECTED_FAILED_SCENARIOS+=("${SUITE_SCENARIO}")
```
phil-davis, line 447: Also needs something around line 481 where it checks the opposite thing - scenarios that were expected to fail to but did not fail.

phil-davis (commented):
It looks good for when `EXPECTED_FAILURES_FILE` is defined.

But when `EXPECTED_FAILURES_FILE` is not defined, then `UNEXPECTED_FAILED_SCENARIOS` will never get anything in it. And so an ordinary test run with failing scenarios is going to pass CI.

One way to fix that is that when `EXPECTED_FAILURES_FILE` is not defined, then check for Failed scenarios and add all those to `UNEXPECTED_FAILED_SCENARIOS` - an `else` at line 493 should do it.

```
		FAILED_SCENARIO_PATHS_COLORED=`awk '/Failed scenarios:/',0 ${TEST_LOG_FILE} | grep feature`
		# There will be some ANSI escape codes for color in the FEATURE_COLORED var.
		# Strip them out so we can pass just the ordinary feature details to Behat.
		# Thanks to https://en.wikipedia.org/wiki/Tee_(command) and
		# https://stackoverflow.com/questions/23416278/how-to-strip-ansi-escape-sequences-from-a-variable
		# for ideas.
		FAILED_SCENARIO_PATHS=$(echo "${FAILED_SCENARIO_PATHS_COLORED}" | sed "s/\x1b[^m]*m//g")
```
can come out of `if [ -n "${EXPECTED_FAILURES_FILE}" ]` because that `FAILED_SCENARIO_PATHS` array will be needed in the `if` and the `else`

phil-davis (approved):
LGTM now

phil-davis:
PR #37806 has a demonstration of induced ordinary failures in a core run. That looks good - we know that CI will correctly fail in core if there is a scenario failing.

PR #37796, opened 2020-08-10
[Tests-only] adjust tests for EOS storage

dpakach:
## Description
1. delete `@skipOnOcis-EOS-Storage` tags because we have now the expected failures list
2. introduce `STORAGE_DRIVER` setting, this setting is used to generate the correct delete command to delete user data
3. hack to create user storage on EOS storage `createEOSStorageHome()` after deleting the user and recreating it we need an extra request to create the user home storage on EOS
4. when creating a user use guid and uid numbers (needed for EOS)
5. when testing on OCIS delete all shares of all users before deleting the user - this is needed because shares are not deleted automatically when user is deleted

## How Has This Been Tested?
https://github.com/owncloud/ocis/pull/704

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/OcisHelper.php
+++ b/tests/TestHelpers/OcisHelper.php
@@ -30,7 +30,6 @@ namespace TestHelpers;
  * @package TestHelpers
  */
 class OcisHelper {
-
 	/**
 	 * @return bool
 	 */
@@ -46,10 +45,32 @@ class OcisHelper {
 	}
 
 	/**
-	 * @return bool
+	 * @return bool|string false if no command given or the command as string
 	 */
 	public static function getDeleteUserDataCommand() {
-		return (\getenv("DELETE_USER_DATA_CMD"));
+		$cmd = \getenv("DELETE_USER_DATA_CMD");
+		if (\trim($cmd) === "") {
+			return false;
+		}
+		return $cmd;
+	}
+
+	/**
+	 * @return string
+	 * @throws \Exception
+	 */
+	public static function getStorageDriver() {
+		$storageDriver = (\getenv("STORAGE_DRIVER"));
+		if ($storageDriver === false) {
+			return "OWNCLOUD";
+		}
+		if ($storageDriver !== "OCIS" && $storageDriver !== "EOS" && $storageDriver !== "OWNCLOUD") {
+			throw new \Exception(
+				"Invalid storage driver. " .
+				"STORAGE_DRIVER must be OCIS|EOS|OWNCLOUD"
+			);
+		}
+		return $storageDriver;
 	}
 
 	/**
@@ -59,12 +80,18 @@ class OcisHelper {
 	 */
 	public static function deleteRevaUserData($user = "") {
 		$deleteCmd = self::getDeleteUserDataCommand();
-		if ($deleteCmd !== false) {
-			$deleteCmd = \sprintf($deleteCmd, $user);
-			\exec($deleteCmd);
-		} else {
+		if ($deleteCmd === false) {
 			self::recurseRmdir(self::getOcisRevaDataRoot() . $user);
+			return;
 		}
+		if (self::getStorageDriver() === "EOS") {
+			$deleteCmd = \str_replace(
+				"%s", $user[0] . '/' . $user, $deleteCmd
+			);
+		} else {
+			$deleteCmd = \sprintf($deleteCmd, $user);
+		}
+		\exec($deleteCmd);
 	}
 
 	/**
@@ -218,5 +245,23 @@ class OcisHelper {
 			return \rmdir($dir);
 		}
 		return true;
+	}
+
+	/**
+	 * On Eos storage backend when the user data is cleared after test run
+	 * Running another test immediately fails. So Send this request to create user home directory
+	 *
+	 * @param string $baseUrl
+	 * @param string $user
+	 * @param string $password
+	 *
+	 * @return void
+	 */
+	public static function createEOSStorageHome($baseUrl, $user, $password) {
+		HttpRequestHelper::get(
+			$baseUrl . "/ocs/v2.php/apps/notifications/api/v1/notifications",
+			$user,
+			$password
+		);
 	}
 }
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -50,6 +50,13 @@ trait Provisioning {
 	private $ou = "TestGroups";
 
 	/**
+	 * gidnumber 3000 is gid of default group "users"
+	 *
+	 * @var int
+	 */
+	private $baseEosGidNumber = 3000;
+
+	/**
 	 * list of users that were created on the remote server during test runs
 	 * key is the lowercase username, value is an array of user attributes
 	 *
@@ -85,6 +92,14 @@ trait Provisioning {
 		"quota definition", "quota free", "quota user", "quota total", "quota relative"
 	];
 
+	/**
+	 * @param int $offset
+	 *
+	 * @return int
+	 */
+	private function getNewEosUidNumber($offset = 0) {
+		return 40040 + \count($this->getCreatedUsers()) + $offset;
+	}
 	/**
 	 * Check if this is the admin group. That group is always a local group in
 	 * ownCloud10, even if other groups come from LDAP.
@@ -880,7 +895,7 @@ trait Provisioning {
 		if ($skeletonStrategy === 'copy') {
 			foreach ($usersAttributes as $userAttributes) {
 				$user = $userAttributes['userid'];
-				$dataDir = $revaRoot . "data/$user/files";
+				$dataDir = $revaRoot . "$user/files";
 				if (!\file_exists($dataDir)) {
 					\mkdir($dataDir, 0777, true);
 				}
@@ -918,7 +933,7 @@ trait Provisioning {
 			$useLdap = false;
 		}
 
-		foreach ($usersAttributes as $userAttributes) {
+		foreach ($usersAttributes as $i => $userAttributes) {
 			if ($useLdap) {
 				$this->createLdapUser($userAttributes);
 			} else {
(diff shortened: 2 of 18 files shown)


tests/TestHelpers/OcisHelper.php
```diff
@@ -62,7 +62,8 @@ public static function getDeleteUserDataCommand() {
 	public static function deleteRevaUserData($user = "") {
 		$deleteCmd = self::getDeleteUserDataCommand();
 		if ($deleteCmd !== false) {
-			$deleteCmd = \sprintf($deleteCmd, $user);
+			$deleteCmd = \sprintf($deleteCmd, $user[0] . '/' . $user);
```
butonic, line 65: this emulates the user layout `{{substr 0 1 .Id.OpaqueId}}/{{.Id.OpaqueId}}` of reva and is incompatible with oc10 ... so we need an isOCIS switch?

individual-it:
I've rebased on current master and will take this over, because I'm currently working on EOS tests

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -49,6 +49,13 @@ trait Provisioning {
 	 */
 	private $ou = "TestGroups";
 
+	/**
+	 * gidnumber 30000 is gid of default group "users"
```
phil-davis, line 53:
```suggestion
	 * gidnumber 3000 is gid of default group "users"
```

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -880,7 +895,7 @@ public function manuallyAddSkeletonFiles($usersAttributes) {
 		if ($skeletonStrategy === 'copy') {
 			foreach ($usersAttributes as $userAttributes) {
 				$user = $userAttributes['userid'];
-				$dataDir = $revaRoot . "data/$user/files";
+				$dataDir = $revaRoot . "$user/files";
```
phil-davis, line 898: This looks like a general fix for all storages. `copy` strategy is being used in both `owncloud/ocis` and `cs3org/reva`

We better make sure that the fix does not break anything.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -932,11 +947,12 @@ public function usersHaveBeenCreated(
 							$userAttributes,
 							__METHOD__ . " userAttributes array does not have key 'userid'"
 						);
-
-						$attributesToCreateUser['email'] = $userAttributes['userid'] . '@owncloud.org';
+						$attributesToCreateUser['email'] = $userAttributes['username'] ?? $userAttributes['userid'] . '@owncloud.org';
```
phil-davis, line 950:
```suggestion
						$attributesToCreateUser['email'] = ($userAttributes['username'] ?? $userAttributes['userid']) . '@owncloud.org';
```

If the email address is not specified, but username is specified, then do you want to append `@owncloud.org` to construct an email address, or is `username` expected to already be a valid email address like `someone@example.org` ?

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -932,11 +947,12 @@ public function usersHaveBeenCreated(
...
 					} else {
 						$attributesToCreateUser['email'] = $userAttributes['email'];
 					}
+					$attributesToCreateUser['gidnumber'] = $this->baseEosGidNumber;
+					$attributesToCreateUser['uidnumber'] = $this->getNewEosUidNumber($i);
```
phil-davis, line 955: If the step that creates users has been used twice ( `usersHaveBeenCreated()` gets called twice in a scenario) then the numbers will get duplicated.

For example, `createUser()` builds an array of users to create with just 1 entry, and calls `usersHaveBeenCreated()` - so that can easily happen multiple times in a scenario.

e.g. writing:
```
And user "Brian" has been created with default attributes and skeleton files
```
might be a problem.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1249,6 +1272,8 @@ public function adminSendsUserCreationRequestWithFollowingAttributesUsingTheProv
 			}
 			$userAttributes["username"] = $username;
 			$userAttributes["email"] = $email;
+			$userAttributes['uidnumber'] = $this->getNewEosUidNumber();
```
phil-davis, line 1275: If there were already some `Given` steps that created some other user, then this will duplicate the `uidnumber`

And even more confusing - there might be some test scenarios where `Alice` is created in the `Given` section, and this `When` step is purposely used to test what happens when we try to create `Alice` again - in that case we probably do want the same `uidnumber` - and then we will want to be able to also wrote test scenarios where we purposely try to create `Alice` a 2nd time, but with a different `uidnumber` !

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1283,7 +1308,16 @@ public function adminSendsUserCreationRequestUsingTheProvisioningApi($user, $pas
...
 		if (OcisHelper::isTestingOnOcis()) {
 			$email = $user . '@owncloud.org';
-			$bodyTable = new TableNode([['userid', $user], ['password', $password], ['username', $user], ['email', $email]]);
+			$bodyTable = new TableNode(
+				[
+					['userid', $user],
+					['password', $password],
+					['username', $user],
+					['email', $email],
+					['uidnumber', $this->getNewEosUidNumber()],
```
phil-davis, line 1317: Similar potential problem here with ending up with duplicate `uidnumber`

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1323,7 +1358,17 @@ public function theAdministratorCreatesUserPasswordGroupUsingTheProvisioningApi(
...
 			$email = $user . '@owncloud.org';
-			$bodyTable = new TableNode([['userid', $user], ['password', $password], ['username', $user], ['email', $email], ['groups[]', $group]]);
+			$bodyTable = new TableNode(
+				[
+					['userid', $user],
+					['password', $password],
+					['username', $user],
+					['email', $email],
+					['groups[]', $group],
+					['uidnumber', $this->getNewEosUidNumber()],
```
phil-davis, line 1368: Similar potential problem here with ending up with duplicate `uidnumber`

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -4313,19 +4358,20 @@ public function theListOfGroupsReturnedByTheApiShouldBeEmpty() {
...
-			if ($this->someUsersHaveBeenCreated()) {
-				OcisHelper::deleteRevaUserData("");
+		if (OcisHelper::isTestingOnOcis() && $this->someUsersHaveBeenCreated()) {
+			foreach ($this->getCreatedUsers() as $user) {
+				$this->deleteAllSharesForUser($user["actualUsername"]);
+				OcisHelper::deleteRevaUserData($user["actualUsername"]);
 			}
 		} else {
 			$this->resetAdminUserAttributes();
 		}
```
phil-davis, line 4369: If we are testing on OCIS and no users have been created, then this logic will cause us to call `resetAdminUserAttributes` - but I think we can only call `resetAdminUserAttributes` on oC10.

Should the logic be put back to have the nested if statements like in the deleted lines?

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -2757,6 +2757,41 @@ public function thePublicAccessesThePreviewOfTheSharedFileUsingTheSharingApi($pa
...
 	}
 
+	/**
+	 * @param $user
+	 *
+	 * @throws Exception
+	 *
+	 * @return void
+	 */
+	public function deleteAllSharesForUser($user) {
```
phil-davis, line 2767: This is a good idea - helps to not leave rubbish share data after a scenario.

tests/acceptance/features/bootstrap/Provisioning.php
individual-it: tried in https://github.com/owncloud/ocis/pull/704

tests/acceptance/features/bootstrap/Provisioning.php
phil-davis: Oh, my bad. `getNewEosUidNumber` calculaton includes counting the users that have already been created.

tests/acceptance/features/bootstrap/Provisioning.php
phil-davis: Oh, my bad. `getNewEosUidNumber` calculaton includes counting the users that have already been created.

tests/acceptance/features/bootstrap/Provisioning.php
phil-davis: Oh, my bad. `getNewEosUidNumber` calculaton includes counting the users that have already been created.

tests/acceptance/features/bootstrap/Provisioning.php
phil-davis: Oh, my bad. `getNewEosUidNumber` calculaton includes counting the users that have already been created.

tests/acceptance/features/bootstrap/Provisioning.php
individual-it: fixed in e98a5d3fe7f5695063fafc3f3dfc719fdd1a80c1

tests/acceptance/features/bootstrap/Provisioning.php
individual-it: as we are checking that userid exists, we can just rely on it, see 701a4ee35b86eb22fe3a302fc40c98843b142255

tests/acceptance/features/bootstrap/Provisioning.php
individual-it: fixed in eab8fbf5ad60de96d93030d35422300a7f7e131d

phil-davis (approved):

phil-davis:
@individual-it squash?

phil-davis (approved):

PR #37797, opened 2020-08-10
[tests-only] export TESTING_REMOTE_SYSTEM variable

individual-it:
## Description
make the variable from the Makefile a environment variable, so it will be passed to `run.sh`

## Motivation and Context
setting the variable to default `true` for `run.sh` never really worked (outside of CI) and the script hang at https://github.com/owncloud/core/blob/d2cc4ec5d88a82f94a5eaaeda78d273e2cd03a4c/tests/acceptance/run.sh#L878. No idea why it worked in CI.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 #
 #  curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
 #  sudo apt-get install -y nodejs build-essential
-#  
+#
 # (installation from distro packages is not recommended, often old versions)
 #
 #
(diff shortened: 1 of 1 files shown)


refs (approved):
works fine 👍 I had this locally but wasn't sure it was the proper fix :D

phil-davis (approved):
Now this works in a local shell. It is strange that it seems to have been working when used in CI.

PR #37811, opened 2020-08-13
[Tests-Only] Check for http status code 204

haribhandari07:
## Description

This PR adds check for http status code = 204. This PR is needed because of PR https://github.com/owncloud/password_policy/pull/314 . Here, `admin expires password of a user and then the user tries to share a file` scenario fails as there is no response when the user tries to create a share.
## Related Issue

 Is needed for PR https://github.com/owncloud/password_policy/pull/314

## How Has This Been Tested?

- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -913,11 +913,16 @@ trait Sharing {
 			$this->ocsApiVersion,
 			$this->sharingApiVersion
 		);
-		$this->lastShareData = $this->getResponseXml();
-		if ($shareType === 'public_link' && isset($this->lastShareData->data)) {
-			$linkName = (string) $this->lastShareData->data[0]->name;
-			$linkUrl = (string) $this->lastShareData->data[0]->url;
-			$this->addToListOfCreatedPublicLinks($linkName, $linkUrl);
+		// In case of HTTP status code 204, there is no content in response payload body.
+		if ($this->response->getStatusCode() === 204) {
+			$this->lastShareData = null;
+		} else {
+			$this->lastShareData = $this->getResponseXml();
+			if ($shareType === 'public_link' && isset($this->lastShareData->data)) {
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -913,6 +913,10 @@ public function createShare(
 			$this->ocsApiVersion,
 			$this->sharingApiVersion
 		);
+		// In case of HTTP status code 204, there is no content in response payload body.
+		if ($this->response->getStatusCode() === 204) {
+			return;
+		}
```
HariBhandari07, line 919: @phil-davis  have a look.

phil-davis: We can set `$this->lastShareData` and to `null` in this case. That will make sure that later `Then` steps do not get tricked by old stale share data from a previous `createShare`

And also set `$this->localLastShareTime = \microtime(true);` because this was the time of an attempt to create a share.

If `204` happens, then we are not really sure if the share did get created, or not. It will be the responsibility of later steps to notice the HTTP status and to check if the sharee can actually see content in the share etc.

Something like:
```
		// In case of HTTP status code 204, there is no content in response payload body.
		if ($this->response->getStatusCode() === 204) {
			$this->lastShareData = null;
		} else {
			$this->lastShareData = $this->getResponseXml();
			if ($shareType === 'public_link' && isset($this->lastShareData->data)) {
				$linkName = (string) $this->lastShareData->data[0]->name;
				$linkUrl = (string) $this->lastShareData->data[0]->url;
				$this->addToListOfCreatedPublicLinks($linkName, $linkUrl);
			}
		}
		$this->localLastShareTime = \microtime(true);
```

phil-davis (approved):

PR #37813, opened 2020-08-13
Include new options for the license manager

jvillafanez:
These options are intended to be used in one of the following cases:

If the application can be used freely but it has some parts that require
a license the app should check the license with "startGracePeriod =>
false" and "disableApp => false". This way the app will mostly work with
the open parts without starting the grace period nor having a valid
license.

If the app is intended to be used with a license but has some minor
parts that can be used without it, the app should check the license with
"startGracePeriod => true" and "disableApp => false". The grace period
will start (if possible) allowing the full usage of the app for a
limited time, but once the grace period ends the app won't be disabled,
leaving only the free parts running.

Note that it's the app's responsibility to handle the license state
switch by themselves. The grace period could end anytime, as well as a
new license could be used

## Description
New options aiming for apps with some parts under a license. These changes are aimed for the next OC release. Some apps will require adjustments in order to use these new changes.

## Related Issue
https://github.com/owncloud/enterprise/issues/4143

## Motivation and Context
Apps with parts under a license (such as richdocuments) aren't taken into account properly. Either they need to be fully licensed or not.
These options aims to fix this problem, so the app can have parts that can be used freely and other parts under a license.

## How Has This Been Tested?
Not tested yet

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/License/LicenseManager.php
+++ b/lib/private/License/LicenseManager.php
@@ -228,14 +228,30 @@ class LicenseManager implements ILicenseManager {
 	 * This method will register a "complain" if we're under the grace period but the app
 	 * doesn't have a valid license, This way it's easier to keep track of what apps
 	 * have invalid license during the grace period
+	 *
+	 * Options:
+	 * - 'startGracePeriod' => true/false (default true): whether a grace period should
+	 *   start (if possible) automatically. Note that if a grace period is ongoing or
+	 *   has expired, this option won't have any effect. Set the option to false if you
+	 *   don't want to start the grace period.
+	 * - 'disableApp' => true/false (default true): whether the app will be disabled if the
+	 *   grace period has finished and there is no valid license. Setting this
+	 *   option to false is intended to be used by apps which are mostly open but
+	 *   have some parts that require a license: you don't want to disable the whole
+	 *   app if there is no valid license.
 	 */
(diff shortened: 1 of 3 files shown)


mrow4a:
Looks good. But to me this is nonbreaking feature -> 10.6

lib/private/License/LicenseManager.php
```diff
@@ -228,14 +228,30 @@ public function getLicenseMessageFor(string $appid, string $language = null): ar
...
+	 *
+	 * Options:
+	 * - 'startGracePeriod' => true/false (default true): whether a grace period should
+	 *   start (if possible) automatically. Note that if a grace period is ongoing or
+	 *   has expired, this option won't have any effect. Set the option to false if you
+	 *   don't want to start the grace period.
+	 * - 'disableApp' => true/false (default true): whether the app will be disabled if the
+	 *   grace period has finished and there is no valid license. Setting this
+	 *   option to false is intended to be used by apps which are mostly open but
+	 *   have some parts that require a license: you don't want to disable to whole
```
phil-davis, line 240:
```suggestion
	 *   have some parts that require a license: you don't want to disable the whole
```

micbar (approved):
Nice! I like the approach.

PR #37824, opened 2020-08-18
[Tests-Only] Fix intermittent fail in public link removal using webUI

### LINKED ISSUE #37817: Test for removing a public link on the webUI fails intermittently
For example:

https://drone.owncloud.com/owncloud/files_primary_s3/1837/67/19

```
  Scenario: user removes the public link of a file                                                                        # /var/www/owncloud/testrunner/tests/acceptance/features/webUISharingPublic2/shareByPublicLink.feature:151
    Given user "Alice" has uploaded file "filesForUpload/lorem.txt" to "/lorem.txt"                                       # FeatureContext::userHasUploadedAFileTo()
    And user "Alice" has created a public link share of file "/lorem.txt"                                                 # FeatureContext::userHasCreatedAPublicLinkShareOf()
    And user "Alice" has logged in using the webUI                                                                        # WebUILoginContext::theUserHasLoggedInUsingTheWebUI()
    When the user removes the public link of file "lorem.txt" using the webUI                                             # WebUISharingContext::theUserRemovesThePublicLinkOfFileUsingTheWebui()
      Fatal error: Call to a member function accept() on boolean (Behat\Testwork\Call\Exception\FatalThrowableError)
    Then the public should see an error message "File not found" while accessing last created public link using the webUI # WebUISharingContext::thePublicShouldSeeAnErrorMessageWhileAccessingLastCreatedPublicLinkUsingTheWebui()
SCENARIO RESULT: (fail)
```

Needs investigation.

haribhandari07:
## Description

Public Link removal using webUI have been failing intermittently in various apps (files_primary_s3, encryption). This PR fixes this intermittent failing issue.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/37817

## How Has This Been Tested?

- https://github.com/owncloud/files_primary_s3/pull/377

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
+++ b/tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
@@ -30,6 +30,7 @@ use Page\OwncloudPage;
 use SensioLabs\Behat\PageObjectExtension\PageObject\Exception\ElementNotFoundException;
 use Page\OwncloudPageElement\OCDialog;
 use WebDriver\Exception\StaleElementReference;
+use PHPUnit\Framework\Assert;
 
 /**
  * The Sharing Dialog
@@ -706,7 +707,20 @@ class SharingDialog extends OwncloudPage {
 	 */
 	public function removePublicLink(Session $session, $number = 1) {
 		$this->clickRemoveBtn($session, $number);
-		$ocDialog = $this->getLastOcDialog($session);
+		// In some cases, it takes some time for the oc dialog to appear.
+		// Here, if oc dialog is not present we are waiting for 0.1 second
+		// and then checking it's visibility on the webUI
+		for ($i = 0; $i < 20; $i++) {
+			$ocDialog = $this->getLastOcDialog($session);
+			if ($ocDialog) {
+				break;
+			}
+			\usleep(100000);
+		}
+		Assert::assertNotFalse(
+			$ocDialog,
+			'oc dialog was expected to be present but not found'
+		);
 		$ocDialog->accept($session);
 		$this->waitForAjaxCallsToStartAndFinish($session);
 	}


tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
```diff
@@ -707,6 +707,7 @@ public function openPublicShareTab(Session $session) {
 	public function removePublicLink(Session $session, $number = 1) {
 		$this->clickRemoveBtn($session, $number);
 		$ocDialog = $this->getLastOcDialog($session);
```
phil-davis, line 709: IMO the problem is that `$this->getLastOcDialog($session)` returns `false` 
So after waiting a bit, and `ocDialogXpath` becomes visible, still `$ocDialog` will be false and line 711 will fail:
`false->ccept($session);`

Somewhere in `getLastOcDialog` we need to make it try harder (wait longer) to actually find a dialog.

tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
```diff
@@ -757,6 +757,11 @@ private function clickRemoveBtn(Session $session, $number = 1) {
 	 */
 	private function getLastOcDialog(Session $session) {
 		$ocDialogs = $this->getOcDialogs();
+		if (!$ocDialogs) {
+			$ocDialogs = $this->waitFor(
```
phil-davis, line 761: When I run this code I get:
```
  Background:                                                                              # /home/phil/git/owncloud/core/tests/acceptance/features/webUISharingPublic2/shareByPublicLink.feature:11
    Given user "Alice" has been created with default attributes and without skeleton files # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()

  Scenario: user removes the public link of a file                                                                        # /home/phil/git/owncloud/core/tests/acceptance/features/webUISharingPublic2/shareByPublicLink.feature:151
    Given user "Alice" has uploaded file "filesForUpload/lorem.txt" to "/lorem.txt"                                       # FeatureContext::userHasUploadedAFileTo()
    And user "Alice" has created a public link share of file "/lorem.txt"                                                 # FeatureContext::userHasCreatedAPublicLinkShareOf()
    And user "Alice" has logged in using the webUI                                                                        # WebUILoginContext::theUserHasLoggedInUsingTheWebUI()
    When the user removes the public link of file "lorem.txt" using the webUI                                             # WebUISharingContext::theUserRemovesThePublicLinkOfFileUsingTheWebui()
      Given callback is not a valid callable (InvalidArgumentException)
    Then the public should see an error message "File not found" while accessing last created public link using the webUI # WebUISharingContext::thePublicShouldSeeAnErrorMessageWhileAccessingLastCreatedPublicLinkUsingTheWebui()
SCENARIO RESULT: (fail)

--- Failed scenarios:

    /home/phil/git/owncloud/core/tests/acceptance/features/webUISharingPublic2/shareByPublicLink.feature:151

1 scenario (1 failed)
6 steps (4 passed, 1 failed, 1 skipped)
0m30.38s (26.46Mb)
```

This `waitFor` method does not work like this. It might be easier to just write a short loop that does
```
usleep(100000);
$ocDialogs = $this->getOcDialogs();
```
until !$ocDialogs or the loop count gets "high" (maybe 20 loops of 0.1 second)

HariBhandari07: Screenshot from 2020-08-20 15-44-56

What??? How is it passing in my machine?

phil-davis: I have to modify the code so that it actually runs it:
```
	private function getLastOcDialog(Session $session) {
		//$ocDialogs = $this->getOcDialogs();
		$ocDialogs = $this->waitFor(2000, $this->getOcDialogs());
		return \end($ocDialogs);
	}
```

Normally it does not go into the `if (!$ocDialogs) {` and so the `waitFor` does not actually get called. So you do not know if it really works.

tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
```diff
@@ -707,6 +708,17 @@ public function openPublicShareTab(Session $session) {
 	public function removePublicLink(Session $session, $number = 1) {
 		$this->clickRemoveBtn($session, $number);
 		$ocDialog = $this->getLastOcDialog($session);
+		if (!$ocDialog) {
+			for ($i =0; $i < 20; $i++) {
+				usleep(100000);
+				$ocDialog = $this->getLastOcDialog($session);
+				if ($ocDialog) break;
+			}
```
HariBhandari07, line 716: @phil-davis  here instead of looping for 20 times, we can wait for 2 seconds and check for the ocDialog. What do you think?

tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
```diff
@@ -707,6 +708,17 @@ public function openPublicShareTab(Session $session) {
...
+		}
+		Assert::assertNotFalse(
+			$ocDialog,
+			'oc dialog was expected to be present but not found'
+		);
```
HariBhandari07, line 721: Also,  I think it's good if we check if `$ocDialog` is falsey(null, empty..) as we need `$ocDialog` to be present so that it can be clicked in `accept()` function.

tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
```diff
@@ -707,6 +708,19 @@ public function openPublicShareTab(Session $session) {
...
 		$this->clickRemoveBtn($session, $number);
 		$ocDialog = $this->getLastOcDialog($session);
+		if (!$ocDialog) {
+			for ($i = 0; $i < 20; $i++) {
+				\usleep(100000);
+				$ocDialog = $this->getLastOcDialog($session);
+				if ($ocDialog) {
+					break;
+				}
+			}
```
phil-davis, line 718: This loop code be the main thing here. There is no need to do the first `$ocDialog = $this->getLastOcDialog($session);` at line 710 separately. And so no need for the `if` enclosing this loop at line 711.

The `usleep` can then move after the if-break, so it only happens after each time `$ocDialog`  is checked.

phil-davis (commented):
Also we could just loop around a couple of times with `sleep(1)` in each loop - because we do not really care so much to exactly minimize the time here. But the `usleep` is also fine.

phil-davis (approved):

tests/acceptance/features/lib/FilesPageElement/SharingDialog.php
phil-davis: We kept the loop like it is. It makes not much difference if we poll every 0.1 second or less frequently.

HariBhandari07:
@phil-davis  our codecov fixer, please have a look :grin:

PR #37827, opened 2020-08-18
Add ability to explicitly set an license implementation

DeepDiver1975:
## Description
In some specific cases the default license implementation shall be replaced by a different implementation.

To make sure the license is replaced as early as possible (before any other app being loaded - with exception of session apps) a new app type 'license' has been introduced.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/3991

## How Has This Been Tested?
- :hand: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37827
+++ b/changelog/unreleased/37827
@@ -1,0 +1,5 @@
+Change: Add ILicenseManager::setLicense
+
+The default license implementation can now be replaced.
+
+https://github.com/owncloud/core/pull/37827
--- a/lib/base.php
+++ b/lib/base.php
@@ -582,10 +582,11 @@ class OC {
 		\stream_wrapper_register('quota', 'OC\Files\Stream\Quota');
 
 		\OC::$server->getEventLogger()->start('init_session', 'Initialize session');
-		OC_App::loadApps(['session', 'theme']);
+		OC_App::loadApps(['session']);
 		if (!self::$CLI) {
 			self::initSession();
 		}
+		OC_App::loadApps(['license', 'theme']);
 
 		\OC::$server->getEventLogger()->end('init_session');
 
--- a/lib/private/License/ILicense.php
+++ b/lib/private/License/ILicense.php
@@ -20,8 +20,8 @@
 namespace OC\License;
 
 interface ILicense {
-	const LICENSE_TYPE_NORMAL = 0;
-	const LICENSE_TYPE_DEMO = 1;
+	public const LICENSE_TYPE_NORMAL = 0;
+	public const LICENSE_TYPE_DEMO = 1;
 
 	/**
 	 * get the raw license string, such as "owncloud_28731_df987_234sf"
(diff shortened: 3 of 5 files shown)


jvillafanez:
What about creating a `getQnapLicense` method in the `LicenseFetcher`?
The `getLicenseWithState` method of the `LicenseManager` can prioritize the Qnap license by getting it from the `LicenseFetcher` (same as in the PR but getting it from the `LicenseFetcher` instead of letting the `LicenseManager` store it.

For the "setLicense", if the license string can be distinguished clearly, you can use the `setLicenseString` method in the `LicenseManager`: just detect the type of the license and set it with the `LicenseFetcher` using a new `setQnapLicense` method.

In order to store the license, if it will be stored in-memory, you can use the `LicenseFetcher` for now.

This way, fetching and storing the license is still `LicenseFetcher`'s responsibility, and devs can still use the `LicenseManager` transparently.

DeepDiver1975:
> What about creating a `getQnapLicense` method in the `LicenseFetcher`?

and with the next license we add getHyperpriceLicense method?

regarding the rest: there is no need in storing a license key anywhere for my use case - all I need is a different ILicense implementation.

jvillafanez:
I don't think having multiple different licenses was fully planned, that's why the path to take for the implementation isn't clear.
In case of a third implementation, we'll need to consider a refactor, which should be limited to the `LicenseFetcher` with some adjustments to the `LicenseManager` to adjust to the new code. In any case, the current public interface shouldn't change despite the refactor.

DeepDiver1975:
> I don't think having multiple different licenses was fully planned, that's why the path to take for the implementation isn't clear.
> In case of a third implementation, we'll need to consider a refactor, which should be limited to the `LicenseFetcher` with some adjustments to the `LicenseManager` to adjust to the new code. In any case, the current public interface shouldn't change despite the refactor.

feel free to propose a different approach - all I want is a way to set a different ILicense impl at runtime ;-)

jvillafanez:
After checking other tickets, what we want is to load a different ILicense implementation from an app. This wasn't planned originally.
What was partially planned was to have multiple implementations controlled by core (inside core, not in apps). This would fit with the sketched solution in https://github.com/owncloud/core/pull/37827#issuecomment-675551222

If we need a fix soon, I guess we'll have to use this PR, but we should implement proper support for this scenario (which will likely break compatibility with this solution)

jvillafanez:
Do we have a deadline for this? I'm looking for ideas to implement the license manager 2.0 in order to include these new requirements, but it will take some time to get there. We can merge this, but we might need to break things in the future

DeepDiver1975:
> Do we have a deadline for this?

10.6

wkloucek:
superseded by #37867

PR #37834, opened 2020-08-22
setupFS for share user before accepting a federated share

### LINKED ISSUE #778: Federated Shares can't accepted if music app is enabled
Since ownCloud 10.4.1 we run into a bug that federated shares cannot be accepted, when the Music player is enabled.
It seems that the user session is broken, so that the user session element is no longer correctly passed.

How to reproduce
1. Create a new federated share from an another owncloud to an owncloud 10.4.1 with music player enabled.
2. Try  to accept the share via the notification pop-up or via "Shared-with-you" option in the menu at the left side.

### LINKED ISSUE #37719: Calling getRootFolder() within any app.php prevents the accepting of federated shares
### Steps to reproduce
1. Have an application installed which has a function call like `$app->getContainer()->getServer()->getRootFolder()` within their `app.php`. The [Music](https://github.com/owncloud/music) app is such, but you can also inject this line of code to the bootstrapping code of another app, and get similar results.
2. Make a federated share from another instance of ownCloud to the first instance (the one with the said app).
3. In the "Shared with you" view, attempt to accept the federated share.

### Expected behaviour
The share can be accepted.

### Actual behaviour
The share is not accepted, but stays in the "Pending" state.

The browser console shows an error
```
Uncaught TypeError: response is null
    responseCallback app.js:236
    jQuery 5
app.js:236:16
```
The owncloud.log has an error:
```
{"reqId":"x8pO5zDRVmmDQPz9tSy7","level":3,"time":"2020-07-20T20:17:32+00:00","remoteAddr":"xxx.xxx.xxx.xxx","user":"xxxx","app":"PHP","method":"POST","url":"\/owncloud_10.4\/ocs\/v1.php\/apps\/files_sharing\/api\/v1\/remote_shares\/pending\/7?format=json","message":"Error: Call to a member function file_exists() on null at \/var\/www\/html\/owncloud_10.4\/apps\/files_sharing\/lib\/Helper.php#285"}
```

This was originally reported at https://github.com/owncloud/music/issues/778. There, it was reported that the bug has been present since ownCloud 10.4.1. Personally, I could reproduce it on 10.4.1.3 but not on 10.2.0.

### Server configuration
**Operating system**: Raspbian GNU/Linux 8.0 (jessie)

**Web server:** nginx/1.6.2

**Database:** sqlite 3.8.7.1

**PHP version:** PHP 7.1.33

**ownCloud version:** 10.4.1.3

**Updated from an older ownCloud or fresh install:** fresh

**Where did you install ownCloud from:** https://download.owncloud.org/community/owncloud-10.4.1.zip (maybe)

**Signing status (ownCloud 9.0 and above):** https://pastebin.com/r9deh0eq (this indicates the modification in the encryption app which I made to reproduce the bug without the Music app)

**The content of config/config.php:** https://pastebin.com/tsCMMT5Y

**List of activated apps:**

```
Enabled:
  - comments: 0.3.0
  - configreport: 0.2.0
  - dav: 0.5.0
  - encryption: 1.4.0
  - federatedfilesharing: 0.5.0
  - federation: 0.1.0
  - files: 1.5.2
  - files_external: 0.7.1
  - files_mediaviewer: 1.0.2
  - files_sharing: 0.12.0
  - files_trashbin: 0.9.1
  - files_versions: 1.3.0
  - firstrunwizard: 1.2.0
  - market: 0.5.0
  - notifications: 0.5.0
  - provisioning_api: 0.5.0
  - systemtags: 0.3.0
  - updatenotification: 0.2.1
Disabled:
  - audioplayer
  - external
  - user_external
```

**Are you using external storage, if yes which one:** no

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** no

### Client configuration
**Browser:** Firefox 78.0.2 (64-bit)

**Operating system:** Windows 10

#### Web server error log
```
(no new lines in the log when the error happens)
```

```
(the related log line already shown above)
```

#### Browser log
```
(the related log line already shown above)
```

karakayasemi:
## Description
Filesystem may not set already for shared user in some cases when accepting a federated share. This PR resolves this problem. Since `setupFS` function does not allows second initialization, this pr change will not affect performance. 

## Related Issue
- Fixes https://github.com/owncloud/core/issues/37719
- Fixes https://github.com/owncloud/music/issues/778
- Fixes https://github.com/owncloud/enterprise/issues/4136

## Motivation and Context
Resolving bugs.

## How Has This Been Tested?
Manually with the sconario describe in related issues.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/External/Manager.php
+++ b/apps/files_sharing/lib/External/Manager.php
@@ -191,6 +191,7 @@ class Manager {
 		$share = $this->getShare($id);
 
 		if ($share) {
+			\OC_Util::setupFS($share['user']);
 			$shareFolder = Helper::getShareFolder();
 			$mountPoint = Files::buildNotExistingFileName($shareFolder, $share['name']);
 			$mountPoint = Filesystem::normalizePath($mountPoint);
--- a/changelog/unreleased/37834
+++ b/changelog/unreleased/37834
@@ -1,0 +1,8 @@
+Bugfix: Fix federated share accepting problem which occurs with some apps enabled
+
+Filesystem may not set already for shared user in some cases when accepting a federated share.
+This situation broke accept federated share api. This problem has been resolved.
+
+https://github.com/owncloud/core/issues/37719
+https://github.com/owncloud/music/issues/778
+https://github.com/owncloud/core/pull/37834


jvillafanez:
> Since setupFS function does not allows second initialization

That's a safety net that we have but we shouldn't use or abuse. If you call the `setupFS` function twice, you have a problem to solve. In fact, I think we should have a debug log (or even I warning) to detect this problem easily.

According to my investigation, the problem isn't that the `setupFS` isn't called but the app calls the FS too early. The FS isn't initialized yet when the 3rd-party app is being loaded.
I think we should check the app loading order. It makes more sense to me that core calls the `setupFS` after all the filesystem apps are loaded. In this case, the music app should be loaded after the filesystem apps (I don't see a reason why the music app should be tagged as filesystem app).
Obviously, the same problem could happen with apps that must be loaded before the FS (session, authentication... apps), but that's kind of expected.

paulijar:
@jvillafanez 
> I don't see a reason why the music app should be tagged as filesystem app

Do you mean this definition in the info.xml:
```
<types>
	<filesystem/>
</types>
```
?

If yes, then that's actually pretty important for any app which listens the filesystem hooks, *and depends on them for correct functionality*. If this tag is not given, the application is allowed to be "Enable only for specific groups" by the admin. When this is done, the application cannot react to the file system hooks from all users, and the music database may end up to inconsistent state, containing references to files which do not exist.

cdamken:
@karakayasemi @paulijar @jvillafanez  @micbar Applying the patch works, I tested manually.

Which are the next steps?

jvillafanez:
I'm not fully happy with the solution, but I don't want to block this fix. If everyone agrees, I'm fine with merging the PR.

VicDeo:
@karakayasemi please rebase

VicDeo (approved):

PR #37836, opened 2020-08-24
[Tests-Only]  Support skipOnFedOcV tags in acceptance tests 

phil-davis:
## Description
Currently on top of #37835 

Sometimes a fix is known to be not compatible with older federated server versions. In that case we need to be able to skip a scenario when running against an older federated server.

When running tests with a federated server, tags are automatically added for the federated server under test, in the form:

`@skipOnFedOcV10.5`
`@skipOnFedOcV10.5.0`

So these tags can be added to scenarios as necessary.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
Implemented in #37837

PR #37839, opened 2020-08-24
[Tests-Only] Assert created user

### LINKED ISSUE #37828: theseUsersHaveBeenCreatedUsingTheOccCommand is missing a check
https://github.com/owncloud/core/blob/master/tests/acceptance/features/bootstrap/OccUsersGroupsContext.php#L129

`theseUsersHaveBeenCreatedUsingTheOccCommand()` ends with:

```
\in_array($tableUsername, $usernameArray, true);
```

which does nothing - the true/false value is not used.

Fix it so that it actually asserts something useful.

swoichha:
## Description
Assert created users with the table of users required to be created.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/37828#

## How Has This Been Tested?
- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/OccUsersGroupsContext.php
+++ b/tests/acceptance/features/bootstrap/OccUsersGroupsContext.php
@@ -24,6 +24,7 @@ use Behat\Behat\Context\Context;
 use Behat\Behat\Hook\Scope\BeforeScenarioScope;
 use Behat\Gherkin\Node\TableNode;
 use PHPUnit\Framework\Assert;
+use TestHelpers\OcsApiHelper;
 
 require_once 'bootstrap.php';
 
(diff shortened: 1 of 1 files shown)


dpakach (commented):
LGTM :+1:

tests/acceptance/features/bootstrap/OccUsersGroupsContext.php
```diff
@@ -130,13 +130,13 @@ public function theseUsersHaveBeenCreatedUsingTheOccCommand(TableNode $table) {
...
 		$createdUsersList = $this->featureContext->getCreatedUsers();
-		$tableUsername = [];
-		$displayName = [];
 		$usernameArray = \array_keys($createdUsersList);
 		foreach ($table as $row) {
-			\array_push($tableUsername, $row['username']);
+			Assert::assertTrue(
+				\in_array($row['username'], $usernameArray),
+				"User: " . $row['username'] . " was not created"
+			);
```
dpakach, line 138: Ok, I think there is something weird going on here. We want to check that the user we just created is actually in the server, but we only check it against the `$username` array which is just a local array stored in the test-runner, so the user is obviously gonna be there even if the user creation failed.
I think you would want to get the users list from the server and check there.

phil-davis: The code here was already strange. 

You could just loop through each user and assert that `featureContext->userShouldExist` is `true`

dpakach (commented):
LGTM :+1:

phil-davis (approved):

PR #37841, opened 2020-08-24
[Tests-Only] Used unexpected array length to determine passed or failed behat status

kiranparajuli589:
## Description
Used unexpected array length to determine passed or failed BEHAT status

## Related Issue
- Found while working for  #37810, So it is a Follow-up PR

## Motivation and Context
- do not crowd run.sh with variables set and reset

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37842, opened 2020-08-24
[Tests-Only] Purposely make some acceptance tests fail

phil-davis:
to check that CI will fail correctly after PR #37841 

3 acceptance test suites should fail.


phil-davis:
Fails as expected.

PR #37852, opened 2020-08-26
[Tests-Only] Reporting expected failures when running a single feature

### LINKED ISSUE #37815: Reporting of unexpected failures when running just a single feature
If a developer runs just a single feature, e.g. putting
```
make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/apiComments/editComments.feature
```

and `EXPECTED_FAILURES_FILE` is defined,

then `tests/acceptance/run.sh` does understand to check just the expected failures for the `apiComments` suite - that is a good start.

But it will also report about expected failures in other feature files like `apiComments/createComments.feature`

1) Make it smarter so that when running just a single feature, it only processes expected failure for the feature.
2) Make it even smarter so that when running just a single scenario, it only processes expected failure for that exact scenario line number (if that is not "too hard" to do)

swoichha:
## Description
Reporting log when running just a single feature or single scenario that are listed in `expected-failure.txt`. Previously we used to get the log about scenarios that were actually running and also for scenarios that were not running but were listed in `expected-failure`.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/37815

## How Has This Been Tested?
- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -1269,8 +1269,25 @@ else
 fi
 if [ "${UNEXPECTED_SUCCESS}" = true ]
 then
+  ACTUAL_UNEXPECTED_PASS=()
+  # if running a single feature or a single scenario
+  if [[ -n "${BEHAT_FEATURE}" ]]
+  then
+    for unexpected_passed_value in "${UNEXPECTED_PASSED_SCENARIOS[@]}"
+    do
+      # check only for the running feature
+      if [[ $BEHAT_FEATURE == *"${unexpected_passed_value}" ]]
+      then
+        ACTUAL_UNEXPECTED_PASS+="${unexpected_passed_value}"
+      fi
+    done
+  else
+    ACTUAL_UNEXPECTED_PASS="${UNEXPECTED_PASSED_SCENARIOS[@]}"
+  fi
+
   tput setaf 3; echo "runsh: Total unexpected passed scenarios throughout the test run:"
-  tput setaf 1; printf "%s\n" "${UNEXPECTED_PASSED_SCENARIOS[@]}"
+  tput setaf 1; printf "%s\n" "${ACTUAL_UNEXPECTED_PASS[@]}"
+
 else
   tput setaf 2; echo "runsh: There were no unexpected success."
 fi


tests/acceptance/run.sh
```diff
@@ -1228,8 +1228,17 @@ else
...
+  for unexpected_passed_value in "${UNEXPECTED_PASSED_SCENARIOS[@]}"
+  do
+    if [[ $unexpected_passed_value != *"${BEHAT_FEATURE}"* ]]
+    then
+     tput setaf 3; echo "runsh: There were no unexpected passed scenarios throughout the test run."
+    else
+      tput setaf 3; echo "runsh: Total unexpected passed scenarios throughout the test run:"
+      tput setaf 1; printf "%s\n" "${UNEXPECTED_PASSED_SCENARIOS[@]}"
+    fi
+  done
```
kiranparajuli589, line 1240: IMO Logging inside the for loop will be confusing.
```suggestion
    ACTUAL_UNEXPECTED_PASS=()
    # if running a single scenario
    if [[ -n "${BEHAT_FEATURE}" && "${BEHAT_FEATURE}" == *":"* ]]
        then
            for unexpected_passed_value in "${UNEXPECTED_PASSED_SCENARIOS[@]}"
            do
                # check only for the running feature
                if [[ $unexpected_passed_value != *"${BEHAT_FEATURE}"* ]]
                then
                    ACTUAL_UNEXPECTED_PASS+="${unexpected_passed_value}"
                fi
            done
    else
        ACTUAL_UNEXPECTED_PASS="${UNEXPECTED_PASSED_SCENARIOS[@]}"
    fi
    tput setaf 3; echo "runsh: Total unexpected passed scenarios throughout the test run:"
    tput setaf 1; printf "%s\n" "${ACTUAL_UNEXPECTED_PASS[@]}"
```

swoichha: Yes, that would be confusing. But now it has been fixed.

tests/acceptance/run.sh
```diff
@@ -1229,8 +1229,25 @@ else
...
 if [ "${UNEXPECTED_SUCCESS}" = true ]
 then
+  ACTUAL_UNEXPECTED_PASS=()
+  # if running a single scenario
+  if [[ -n "${BEHAT_FEATURE}" && "${BEHAT_FEATURE}" == *":"* ]]
+  then
+    for unexpected_passed_value in "${UNEXPECTED_PASSED_SCENARIOS[@]}"
+    do
+      # check only for the running feature
+      if [[ $unexpected_passed_value != *"${BEHAT_FEATURE}"* ]]
```
phil-davis, line 1239: Should this be the other way around? We want to add the to ACTUAL_UNEXPECTED_PASS if it is the BEHAT_FEATURE.

Also,
- I think BEHAT_FEATURE is set to something like `tests/acceptance/features/apiComments/comments.feature:42` but the "unexpected" array will have just `apiComments/comments.feature:42` in it. See code that calculates `SUITE_SCENARIO` from the end of a path.
- it needs to not match entries like `apiComments/comments.feature:421` - the last `*` here is probably not needed.

It might be easier to use the bash regex-match operator `=~` - then you can match strings exactly at the start or end. See examples in `run.sh` that use `=~`

Or actually I think that after extracting the ``apiComments/comments.feature:42` part from BEHAT_FEATURE then you can check for an exact match - no need for "glob *" or regex.

swoichha: `BEHAT_FEATURE` is set to `apiComments/comments.feature:42` as well. So, I think we can directly check for an exact match using`==` by removing the first and last `*` from `*"${BEHAT_FEATURE}"* ]]`

tests/acceptance/run.sh
```diff
@@ -1229,8 +1229,25 @@ else
 fi
 if [ "${UNEXPECTED_SUCCESS}" = true ]
 then
+  ACTUAL_UNEXPECTED_PASS=()
+  # if running a single scenario
+  if [[ -n "${BEHAT_FEATURE}" && "${BEHAT_FEATURE}" == *":"* ]]
```
kiranparajuli589, line 1234:
```suggestion
  # if running a single feature
  if [[ -n "${BEHAT_FEATURE}" && "${BEHAT_FEATURE}" != *":"* ]]
```

kiranparajuli589:
lgtm :+1:

kiranparajuli589 (commented):
lgtm

phil-davis (approved):

PR #37856, opened 2020-08-27
If the client disconnects while downloading, use debug level instead

jvillafanez:
## Description
If a client disconnects while downloading a shared file or folder, use a debug log instead of logging the whole exception.

## Related Issue
https://github.com/owncloud/enterprise/issues/4162

## Motivation and Context
There seems to be frequent client disconnections in some environment which cause these exceptions to be logged. This could be normal if the client decide to cancel downloads (especially for big files)

## How Has This Been Tested?

1. Share a file via public link
2. Using a slow network connection, download the file via the public link, and cancel it while the file is still downloading.

Instead of an exception to be logged, a debug log level should appear in the logs

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37856
+++ b/changelog/unreleased/37856
@@ -1,0 +1,9 @@
+Change: Use a debug log level if a share download is aborted
+
+If a client was downloading a file through a public link share
+and he decided to disconnect and abort the download, ownCloud was
+logging that exception.
+Now ownCloud will log a message with a debug level instead of logging
+the exception in order to reduce the noise.
+
+https://github.com/owncloud/core/pull/37856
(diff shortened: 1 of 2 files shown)


jvillafanez:
I don't think we need a changelog entry for this.

micbar:
> I don't think we need a changelog entry for this.

why? I would like that written down

jvillafanez:
Done.

VicDeo (approved):
`connection_aborted() === 1` check suits here better, but it's rather a matter of taste

phil-davis:
Rebased to get "modern" CI. Can be merged when CI is green.

PR #37857, opened 2020-08-27
[Tests-Only] Add u to username

phil-davis:
This is just a demonstrator for https://github.com/owncloud/product/issues/187 to show that it is possible to have `userid` and `username` different in `owncloud/ocis`

OCIS PR https://github.com/owncloud/ocis/pull/498 demonstrates that this passes.

If we actually do this some day, the code will be different - this is just a quick change to show that it works.


phil-davis:
Leaving the branch here so it can be used from other repos if needed.

PR #37858, opened 2020-08-27
private/Files/Type/Detection: fix isWrapped cond

quasilyte:
## Description

This is a minor bug fix.

Due to the `and` and `=` precedence, instead of checking for both
`://` and `file://` we only check for `://`:
```php
$path = 'foo://';
$isWrapped = (\strpos($path, '://') !== false) and (\substr($path, 0, 7) === 'file://');
var_dump($isWrapped); // => true, but want false
```
If we change `and` to `&&` the result becomes `false` as it should.

## Related Issue

I don't know.

## Motivation and Context

The code behaves incorrectly and/or it's misleading.

## How Has This Been Tested?

If the current tests were passing, it probably was not tested before as well.

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Type/Detection.php
+++ b/lib/private/Files/Type/Detection.php
@@ -231,7 +231,7 @@ class Detection implements IMimeTypeDetector {
 				return empty($mimeType) ? 'application/octet-stream' : $mimeType;
 			}
 		}
-		$isWrapped = (\strpos($path, '://') !== false) and (\substr($path, 0, 7) === 'file://');
+		$isWrapped = (\strpos($path, '://') !== false) && (\substr($path, 0, 7) === 'file://');
 		if (!$isWrapped and $mimeType === 'application/octet-stream' && \function_exists("mime_content_type")) {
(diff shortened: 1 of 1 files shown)


phil-davis:
```
$x = 2;
$y = 3;
$t = ($x == 2) and ($y == 4);
var_dump($t);
bool(true)
```
It doesn't give the answer that a mere human might expect!

https://github.com/owncloud/core/blob/master/tests/lib/Files/Type/DetectionTest.php has the unit tests for this.

@quasilyte can you make an extra test case there that would fail on current master and will pass with your fix?

quasilyte:
I'm testing the static analyzer while reporting issues that were found to the php-corpus repositories upstreams in forms of PRs. If it requires more than a simple edit, I'm afraid I'm not motivated enough.

I have no project knowledge, nor I know whether this code path is actually ever executed.
I'll let anyone else submit a proper patch. :)

PR #37867, opened 2020-09-01
Add ability to explicitly set an license implementation via system co…

DeepDiver1975:
…nfig

## Description
An alternative license implementation can be loaded via the system config 'license-class'. 

## Related Issue
-    Fixes owncloud/enterprise#3991

## How Has This Been Tested?
- :hand: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/License/LicenseFetcher.php
+++ b/lib/private/License/LicenseFetcher.php
@@ -19,7 +19,7 @@
  */
 namespace OC\License;
 
-use OC\License\BasicLicense;
+use OC\HintException;
 use OCP\IConfig;
 use OCP\AppFramework\Utility\ITimeFactory;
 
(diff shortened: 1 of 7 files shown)


lib/private/License/LicenseFetcher.php
```diff
@@ -48,9 +49,18 @@ public function __construct(IConfig $config, ITimeFactory $timeFactory) {
...
+	 * @throws HintException
 	 */
 	public function getOwncloudLicense(): ?ILicense {
+		$licenseClass = $this->config->getSystemValue('license-class', null);
+		if ($licenseClass) {
+			if (!\class_exists($licenseClass)) {
+				throw new HintException("Unknown license $licenseClass");
+			}
+			return new $licenseClass();
+		}
```
jvillafanez, line 63: What about using the license class as a wrapper of the license string? Does it make sense?
Basically, we could have `BasicLicense($licenseKey)` as well as `MyLicense($licenseKey)` being the `$licenseKey` the same license string in both cases. Obviously, it would be unlikely that both objects would evaluate the same license as valid (although it could happen).
This means that the new license string will still be stored the same way as the current license.

```
$licenseClass = $this->config->getSystemValue('license-class', BasicLicense::class);
if (!\class_exists($licenseClass)) {
  throw new HintException("Unknown license $licenseClass");
}
.....
  $license = new $licenseClass($licenseKey);  // instead of BasicLicense($licenseKey)
.....
```

This should also work with the textbox used to set the license in the settings page regardless of the actual license "wrapper"

DeepDiver1975: :shrug:

DeepDiver1975: generally speaking I'd avoid to force ctor args in such situations .....

jvillafanez: I don't see a clean solution other than going back to square one and provide a registration option via events so the apps can register their own provider.
The main problem I see is that core must create the object when the object should be provided somehow by the app to core. I think this approach is bound to have problems because we can't force the constructor to have a particular signature. If we want to go this way, I'd rather have consistency as much as possible.

Maybe I'm biased with my idea, but I think it's weird that a license object, which is expected to be immutable, doesn't have a initialization method using the license string as base.
I think another option could be to adjust the code so the `BasicLicense` doesn't use the license string as constructor parameter in order to keep consistency with the license provided by the apps, but I'm not sure how many changes are required to pull this off.

jvillafanez (approved):

PR #37874, opened 2020-09-02
[Tests-Only] Fix add users in provisioning api when testing on ocis

jasson99:
## Description
This PR fixes add users such that when run on ocis, users are created with email.

## Related Issue
- https://github.com/owncloud/ocis/issues/512#issuecomment-685282719

## How Has This Been Tested?

core CI and https://github.com/owncloud/ocis/pull/517

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -815,6 +815,60 @@ trait Provisioning {
 	}
 
 	/**
+	 * Manually add skeleton files for a single user on OCIS and reva systems
+	 *
+	 * @param string $user
+	 * @param string $password
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function manuallyAddSkeletonFilesForUser($user, $password) {
+		$settings = [];
+		$setting["userid"] = $user;
+		$setting["password"] = $password;
+		\array_push($settings, $setting);
+		$this->manuallyAddSkeletonFiles($settings);
+	}
+
+	/**
+	 * Manually add skeleton files on OCIS and reva systems
+	 *
+	 * @param array $usersAttributes
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function manuallyAddSkeletonFiles($usersAttributes) {
+		$skeletonDir = \getenv("SKELETON_DIR");
+		$revaRoot = \getenv("OCIS_REVA_DATA_ROOT");
+		if (!$skeletonDir) {
+			throw new Exception('Missing SKELETON_DIR environment variable, cannot copy skeleton files for OCIS');
+		}
+		if (!$revaRoot && OcisHelper::getDeleteUserDataCommand() !== false) {
+			foreach ($usersAttributes as $userAttributes) {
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1177,14 +1177,39 @@ public function adminSendsUserCreationRequestWithFollowingAttributesUsingTheProv
...
+			$setting["userid"] = $user;
+			$setting["displayName"] = $username;
+			$setting["password"] = $password;
+			$setting["email"] = $email;
+			\array_push($settings, $setting);
+			$this->usersHaveBeenCreated(
+				true,
+				$settings,
+				'api'
+			);
```
phil-davis, line 1209: Why do we need to do this?
`userSendsHTTPMethodToOcsApiEndpointWithBody` already created the user???

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -1177,14 +1177,39 @@ public function adminSendsUserCreationRequestWithFollowingAttributesUsingTheProv
...
 		$user = $this->getActualUsername($user);
 		$password = $this->getActualPassword($password);
-		$bodyTable = new TableNode([['userid', $user], ['password', $password]]);
+		if (OcisHelper::isTestingOnOcis()) {
+			$username = $user;
+				Assert::assertNotNull(
+					$username,
+					__METHOD__ . $username . " is null"
+				);
+				$email = $username . '@owncloud.org';
```
phil-davis, line 1186: Indent is different.

jasson99: fixed

tests/acceptance/features/bootstrap/Provisioning.php
phil-davis: It gets the skeleton for the user.

But that will also send a user create request again. (Which might not do any damage, but is not optimal.)

It would be good to separate the code in `usersHaveBeenCreated` `if ($skeleton && OcisHelper::isTestingOnOcis()) {}` into a separate function (call it something like `manuallyAddSkeletonFiles` ). Then it will be easy to call `manuallyAddSkeletonFiles` here, instead of `usersHaveBeenCreated`

jasson99: fixed

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -814,6 +814,43 @@ public function resetOldLdapConfig() {
...
 
+	/**
+	 * This function will allow us to manually add skeleton files
+	 *
+	 * @param array $usersAttributes
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function manuallyAddSkeletonFiles($usersAttributes) {
```
phil-davis, line 825: This should be able to be called from `usersHaveBeenCreated` also, to avoid this code being all duplicated.

phil-davis (approved):

PR #37875, opened 2020-09-02
redesign ugly, yellow notification center

hurradieweltgehtunter:
While working on some changes for oc.online I redesigned the (ugly) yellow notification center.
Before:
grafik

After:
grafik

- Moved the notification bar to the bottom of the screen to not block the centered top logo
- Stacking messages is not a problem, lines are separating messages
- Closing button is also re-designed and long messages do not overlap the button and wrap nicely.

As discussed with @pmaier1 @PVince81 and @micbar I created this PR to adopt this change for OC in general.

### DIFF
--- a/core/js/js.js
+++ b/core/js/js.js
@@ -1134,13 +1134,15 @@ OC.Notification = {
 	 *
 	 * @param {string} html Message to display
 	 * @param {Object} [options] options
-	 * @param {string] [options.type] notification type
+	 * @param {string} [options.type] notification type
+	 * @param {string} [options.showCloseButton] show close button
 	 * @param {int} [options.timeout=0] timeout value, defaults to 0 (permanent)
 	 * @return {jQuery} jQuery element for notification row
 	 */
 	showHtml: function (html, options) {
 		options = options || {};
 		_.defaults(options, {
+			showCloseButton: true,
 			timeout: 0
 		});
 
@@ -1153,7 +1155,7 @@ OC.Notification = {
 		if (options.type) {
 			$row.addClass('type-' + options.type);
 		}
-		if (options.type === 'error') {
+		if (options.showCloseButton === true) {
 			// add a close button
 			var $closeButton = $('<a class="action close icon-close" href="#"></a>');
 			$closeButton.attr('alt', t('core', 'Dismiss'));
@@ -1184,6 +1186,7 @@ OC.Notification = {
 	 * @param {string} text Message to display
 	 * @param {Object} [options] options
 	 * @param {string} [options.type] notification type
+	 * @param {string} [options.showCloseButton] show close button
 	 * @param {int} [options.timeout=0] timeout value, defaults to 0 (permanent)
 	 * @return {jQuery} jQuery element for notification row
 	 */
--- a/core/js/license-state-notification.js
+++ b/core/js/license-state-notification.js
@@ -18,6 +18,6 @@
 $(document).ready(function() {
 	$.get(OC.generateUrl('/license/licenseMessage'), {app: 'core'})
 		.done(function(data) {
-			OC.Notification.showHtml(data.translated_message.join('<br/>'), {type: 'error'});
+			OC.Notification.showHtml(data.translated_message.join('<br/>'), {showCloseButton: false});
 		});
 });
--- a/changelog/unreleased/37875
+++ b/changelog/unreleased/37875
@@ -1,0 +1,9 @@
+Enhancement: Re-design UI yellow notification bar
+
+The yellow notification center has been redesigned:
+- not blocking the top centered logo anymore
+- gets more attention using icons and red color
+- looks more modern
+- shows a close button by default so users can dismiss most notifications
+
+https://github.com/owncloud/core/pull/37875
--- a/core/css/styles.css
+++ b/core/css/styles.css
@@ -639,25 +639,36 @@ td.avatar {
 
 
 #notification-container {
-	position: absolute;
-	top: 0;
+	position: fixed;
+	bottom: 0;
+	top: auto;
 	width: 100%;
 	text-align: center;
+	z-index: 8000;
 }
 #notification {
 	margin: 0 auto;
 	max-width: 60%;
-	z-index: 8000;
-	background-color: #fc4;
+	background-color: #fef4f6;
+	background-image: url('../img/actions/exclamation-triangle-solid.svg');
+	background-position: 20px 50%;
+	background-repeat: no-repeat;
+	background-size: 42px;
+	color: #f0506e;
 	border: 0;
-	padding: 1px 8px;
+	padding: 10px 40px 0px 75px;
 	display: none;
 	position: relative;
 	top: 0;
-	border-bottom-left-radius: 3px;
-	border-bottom-right-radius: 3px;
+	border-radius: 3px;
 	-ms-filter: "progid:DXImageTransform.Microsoft.Alpha(Opacity=90)";
 	opacity: .9;
+	max-height: none;
+}
+
+#notification-container #notification * {
+	color: #f0506e !important;
+	font-weight: bold;
 }
 #notification span {
 	cursor: pointer;
@@ -666,20 +677,33 @@ td.avatar {
 #notification {
 	overflow-x: hidden;
 	overflow-y: auto;
-	max-height: 100px;
+}
+#notification-container #notification a:hover {
+	text-decoration: underline;
 }
 #notification .row {
 	position: relative;
+	border-bottom: 1px solid #fdc7d1;
+	padding-bottom: 5px;
+	margin-bottom: 5px;
+	text-align: left;
 }
+#notification .row:last-child {
+	border: none;
+}
+
 #notification .row .close {
 	display: inline-block;
 	vertical-align: middle;
 	position: absolute;
 	right: 0;
-	top: 0;
+	top: calc(50% - 7px);
 }
 #notification .row.closeable {
 	padding-right: 20px;
+}
+#notification .row .close.icon-close {
+	background-image: url('../img/actions/close-red.svg');
 }
 
 tr .action:not(.permanent),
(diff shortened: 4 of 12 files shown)


mmattel:
Without having tested it, this must work very well with https://github.com/owncloud/direct_menu 😃

phil-davis:
CI fails - see #37877 - need to look at how a user can clear the notification so that they can see file(s) behind it.

hurradieweltgehtunter:
> CI fails - see #37877 - need to look at how a user can clear the notification so that they can see file(s) behind it.

Uh that's unfortunate. I didn't touch the JS logic, I just did css changes. Some notifications do not have a close button in general (only type: error).
IMHO the notifications do not block the access of files since the user can just scroll down. 

There could be a solution: Change all notifications to `type:error` so the closing icon appears.
See comment https://github.com/owncloud/core/issues/37877#issuecomment-686314819

hurradieweltgehtunter:
I think these are two different topics. My changes only affect the appearance. Clearing notifications is changing the behaviour -> JS work
Question is, if the notifications on the bottom are really that bad for the user (= failing test) or if the test needs to be adjusted.
IMO the new look does does not overlay the whole file (max-width: 60%) so it's still clickable. For non-clearable notifications it is quite important to resolve the hint. Thus, permanently showing it is ok.
Clearable notifications are not blocking anything too.

Todo: Check all notifications if they should be clearable or not and change them accordingly -> different scope to this PR
What do you think?

phil-davis:
Yes, the problem with the tests is not strictly a blocker for this PR. A real user can scroll the files list so that the file they want to see is visible above the notification, and then open the sharing panel or whatever they want to do.

The selenium test driver "automagically" scrolls the file list so that the required file  is in the "displayed" part of the file list. e.g. it scrolls down so that `lorem.txt` is "on  the screen" - but only just on the bottom of the "screen". It ends up where the notification overlays the bottom part of the file list.

We can either:
1) sort out the tests so they don't rely on files that are a long way down the list (or some other test workaround) or;
2) make this notice dismissable, and adjust the test so that after checking that the notice is there, it clicks the "close-dismiss" button to get rid of the notification.

Actually (2) will provide better UI anyway (give the user a way to acknowledge and dismiss the notification), and the test scenario will also flow smoothly in a similar way to a typical user.

hurradieweltgehtunter:
Agreed. But then we need to change the notifications' JS logic. Either
- Change the defaults notification behaviour to have all types of notifications include the close button
or
- Change all notifications in OC to `type: error`

So far I didn't find the declaration of the notification function. Need to inspect which one is more suitable ( I think the cirst one)

hurradieweltgehtunter:
Latest change switched notifications to show close button as default setting. 
Added option `hideCloseButton` to hide it explicitly (useful for some notifications).

Questions is: Were the notifications until now intented to be not closeable?

hurradieweltgehtunter:
Made some important notifiactions not dismissible (Integrity Check, Trial License). 
Somehow the JS Tests fail even more. Someone should check this out, I'm not aware how therse tests work and can be fixed.

phil-davis:
Needs a rebase because https://drone.owncloud.com/owncloud/core/26733/64/13 
```
[Behat\Testwork\Suite\Exception\SuiteNotFoundException]                            
  `apiShareManagementBasic` suite is not found or has not been properly registered.
```
The API test suite name changed yesterday. This branch needs and up-to-date "view" of the changed `.drone.star`

phil-davis:
I pushed webUI acceptance test code to dismiss the notification in the locking tests before going on to check items on the files page. I will let CI run then we can rebase...

phil-davis:
> Made some important notifiactions not dismissible (Integrity Check, Trial License).
> Somehow the JS Tests fail even more. Someone should check this out, I'm not aware how therse tests work and can be fixed.

I adjusted a JS test expectation so that it knows to expect the close button to exist.

phil-davis:
Rebased. CI should all pass now.

@micbar ready for technical review and review of the new UI from a user perspective. Please allocate the appropriate people.

If all is approved, then commits could be squashed... before merge.

core/js/js.js
```diff
@@ -1135,6 +1135,7 @@ OC.Notification = {
 	 * @param {string} html Message to display
 	 * @param {Object} [options] options
 	 * @param {string] [options.type] notification type
+	 * @param {string] [options.hideCloseButton] hide close button
```
DeepDiver1975, line 1138: negative logic is beyond best practice

Please use 'showCloseButton' instead of 'hideCloseButton'

DeepDiver1975 (changes requested):
And some sqashing of commits would be good habit from my pov .....

core/js/js.js
phil-davis, line 1138:
```suggestion
	 * @param {string} [options.type] notification type
	 * @param {string} [options.hideCloseButton] hide close button
```
Bracket-pairing needs fixing.
(and also comment above by @DeepDiver1975

hurradieweltgehtunter: But it is the desired behaviour. Close button is always visible until this option is set to true.
Using `showCloseButton` would mean to change every single implementation of `showNotification`

micbar: Use positive default?

DeepDiver1975: I'm not asking to change the behavior but to change the program and api logic to make it more intuitive.

Having showCloseButton with default true is the same as hideCloseButton with default false and vise-versa.

hurradieweltgehtunter: Yeah wasn't thinking right. Switched to positive logic.

hurradieweltgehtunter: done

hurradieweltgehtunter:
Squashed commits

core/js/js.js
```diff
@@ -1153,7 +1155,7 @@ OC.Notification = {
 		if (options.type) {
 			$row.addClass('type-' + options.type);
 		}
-		if (options.type === 'error') {
+		if (options.showCloseButton !== true) {
```
phil-davis, line 1158:
```suggestion
		if (options.showCloseButton === true) {
```
Or do I not understand?

core/js/js.js
```diff
@@ -1184,6 +1186,7 @@ OC.Notification = {
 	 * @param {string} text Message to display
 	 * @param {Object} [options] options
 	 * @param {string} [options.type] notification type
+	 * @param {string} [options.hideCloseButton] hide close button
```
phil-davis, line 1189: This looks out-of-date.

core/js/js.js
hurradieweltgehtunter: you are right. done

core/js/js.js
hurradieweltgehtunter: done

core/js/js.js
```diff
@@ -1134,13 +1134,15 @@ OC.Notification = {
...
-	 * @param {string] [options.type] notification type
+	 * @param {string} [options.type] notification type
+	 * @param {string} [options.showCloseButton] show close button
 	 * @param {int} [options.timeout=0] timeout value, defaults to 0 (permanent)
 	 * @return {jQuery} jQuery element for notification row
 	 */
 	showHtml: function (html, options) {
 		options = options || {};
 		_.defaults(options, {
+			showCloseButton: false,
```
phil-davis, line 1145: I was fixing the JS tests, and switched  this from default true to default false. That preserves whatever was the previous behavior of existing notifications.

But maybe we do want to give notifications a close button by default???

core/js/js.js
```diff
@@ -1153,7 +1155,7 @@ OC.Notification = {
 		if (options.type) {
 			$row.addClass('type-' + options.type);
 		}
-		if (options.type === 'error') {
+		if ((options.showCloseButton === true) || (options.type === 'error')) {
```
phil-davis, line 1158: And I put back the `(options.type === 'error')` so that whatever notifications are currently type "error" will continue to show a close button.

core/js/license-state-notification.js
```diff
@@ -18,6 +18,6 @@
 $(document).ready(function() {
 	$.get(OC.generateUrl('/license/licenseMessage'), {app: 'core'})
 		.done(function(data) {
-			OC.Notification.showHtml(data.translated_message.join('<br/>'), {type: 'error'});
+			OC.Notification.showHtml(data.translated_message.join('<br/>'), {showCloseButton: false});
```
phil-davis, line 21: This is a behavior change. Prior to this, the license notification would have had a close button, now it will not have a close button. Is that change intended?

phil-davis:
I rebased and fixed the JS tests. While doing that, I made `showCloseButton` default to `false` to preserve the existing behavior more closely. See my comments above.

Or should we make `showCloseButton` be true by default, and make sure to find all the notifications that should not have a close button, and make them explicitly set `showCloseButton` to false?

core/js/license-state-notification.js
hurradieweltgehtunter: License notifications and CSFR check notifications should not be closeable, yes

hurradieweltgehtunter:
`showCloseButton = true` should be the default

phil-davis:
I will change the default back to `showCloseButton = true` and adjust tests...

phil-davis:
Done - `showCloseButton = true` is the default. Hopefully CI passes!

I assigned myself so that I easily notice this PR. If the net code diff is OK now, I can squash the commits and merge.

@DeepDiver1975 @micbar please review again.

DeepDiver1975 (approved):

phil-davis (approved):
squashed - this should be OK to merge when CI passes

hurradieweltgehtunter:
🍾  thank you for your help!

mmattel:
@hurradieweltgehtunter 
can you please file a docs issue and link it.
this merged pr is docs relevant 😃

phil-davis:
docs PR https://github.com/owncloud/docs/issues/2793

PR #37881, opened 2020-09-03
[Tests-Only] Refactor createShareToShareFolder.feature to use Shares folder

phil-davis:
## Description
This is a way to get sharing API test scenarios that will be valid in both oC10 and OCIS. OCIS always puts received shares into a folder called `Shares` for the share receiver. On OCIS, auto-accept shares is always disabled.

1) keep the existing oC10 sharing scenarios that deliver shares into the root `/` folder of the share receiver. Tag them `notToImplementOnOCIS`. That maintains the test coverage of the default sharing on oC10 and the scenarios will not get run on OCIS.
2) make new scenarios in new feature files `*ToSharesFolder.feature` - for example `createShareToSharesFolder.feature`. In those scenarios, set the folder for received shares to "Shares".
`Given the administrator has set the default folder for received shares to "Shares"`
And adjust the test steps and expectations so that the share receiver expects the shares in `/Shares` folder.
Put no specific tag on these test scenarios - they should pass on both oC10 and OCIS.
That tests a full range of oC10 behavior when the default folder for received shares is configured, and also tests sharing behavior in OCIS.

3) in the new scenarios, disable auto-accept shares. On OCIS this will do nothing (auto-accept shares is already disabled). On oC10 this will provide the same expected behavior as on OCIS.
`Given auto-accept shares has been disabled`
And adjust the scenarios so that they have steps to accept a share before the receiver tries to access it. e.g.
```
    When user "Brian" accepts the share "/textfile0.txt" offered by user "Alice" using the sharing API
    And the content of file "/Shares/textfile0.txt" for user "Brian" should be "ownCloud test text file 0"
```

So the pattern of these new feature files will have a `Background` that has at least these lines:
```
  Background:
    Given the administrator has set the default folder for received shares to "Shares"
    And auto-accept shares has been disabled
```

And such features should work the same way on oC10 and OCIS - so the scenarios can be run on both.

4) Put the new feature files into new suites so that each suite still has a reasonable run-time. e.g. in this PR the existing `apiShareManagementBasic` suite becomes `apiShareToRootManagementBasic` and `apiShareToSharesManagementBasic`

## Related Issue
https://github.com/owncloud/ocis/issues/518

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


individual-it:
`Given the administrator has set the default folder for received shares to "Shares"` - we need to understand that this step will also not do anything on OCIS and if s.o. sets it to s.th. else than `Shares` the tests in ocic will fail.
Do we need to document that?

phil-davis:
> `Given the administrator has set the default folder for received shares to "Shares"` - we need to understand that this step will also not do anything on OCIS and if s.o. sets it to s.th. else than `Shares` the tests in ocic will fail.
> Do we need to document that?

For a start, I could change the words to:
```
Given the default folder for received shares has been set to "Shares"
```

That is more the standard grammar that we use. Not mentioning administrator makes it a bit less like an "action" step.

We could also say:
```
Given the default folder for received shares is "Shares"
```
That would not imply that any action has to be done - on some system-under-test it might need a setting to be set/changed, on another system-under-test it might be "is" already by default.

individual-it (approved):

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1923,11 +1943,11 @@ public function checkingLastShareIDIsIncluded() {
 	}
 
 	/**
-	 * @Then /^the last share_id should not be included in the response/
+	 * @Then /^the last share id should not be included in the response/
```
phil-davis, line 1946: Note: I took the opportunity to change the step text slightly.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1966,15 +1986,15 @@ public function theResponseShouldNotContainAnyShareIds() {
 	}
 
 	/**
-	 * @Then user :user should not see share_id of last share
+	 * @Then user :user should not see the share id of the last share
```
phil-davis, line 1989: Note: I took the opportunity to change the step text slightly.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -2462,7 +2482,7 @@ public function userHasDeletedPublicLinkShareNamedUsingTheSharingApi(
 	}
 
 	/**
-	 * @When /^user "([^"]*)" (declines|accepts) the share "([^"]*)" offered by user "([^"]*)" using the sharing API$/
+	 * @When /^user "([^"]*)" (declines|accepts) share "([^"]*)" offered by user "([^"]*)" using the sharing API$/
```
phil-davis, line 2485: Note: I took the opportunity to change the step text slightly.

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -2504,7 +2524,7 @@ public function userReactsToShareOfferedBy($user, $action, $share, $offeredBy) {
 	}
 
 	/**
-	 * @Given /^user "([^"]*)" has (declined|accepted) the share "([^"]*)" offered by user "([^"]*)"$/
+	 * @Given /^user "([^"]*)" has (declined|accepted) share "([^"]*)" offered by user "([^"]*)"$/
```
phil-davis, line 2527: Note: I took the opportunity to change the step text slightly.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -494,7 +494,7 @@ public function slowdownDavRequests($method, $seconds) {
 	public function destinationHeaderValue($user, $fileDestination) {
 		$fullUrl = $this->getBaseUrl() . '/' .
 			WebDavHelper::getDavPath($user, $this->getDavPathVersion());
-		return $fullUrl . '/' . \ltrim($fileDestination, '/');
+		return \rtrim($fullUrl, '/') . '/' . \ltrim($fileDestination, '/');
```
phil-davis, line 497: When developing this PR, I noticed that some requests had a double-slash here. If `WebDavHelper::getDavPath` already ends with a slash `/` then we do not want to end up with 2 slashes. (the server was responding OK, but it will be nicer to test the ordinary case when the request URL has just a single slash between each part)

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1373,7 +1373,7 @@ public function asFileOrFolderShouldExist(
 		);
 		Assert::assertTrue(
 			$this->isEtagValid(),
-			"$entry '$path' expected to exist but not found"
+			"$entry '$path' expected to exist for user $user but not found"
```
phil-davis, line 1376: I enhanced this message. In scenario outlines with multiple users, it was not clear which user was missing the expected file or folder.

Issue #37884, opened 2020-09-04
cloud/apps endpoint does not report always-enabled-apps in list of all apps


phil-davis:
### Steps to reproduce
1. `curl -k -u admin:admin -X GET http://172.17.0.1:8080/ocs/v1.php/cloud/apps?filter=enabled`
Some reasonable list is returned:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data>
  <apps>
   <element>files</element>
   <element>comments</element>
   <element>dav</element>
   <element>federatedfilesharing</element>
   <element>federation</element>
   <element>files_external</element>
   <element>files_sharing</element>
   <element>files_trashbin</element>
   <element>files_versions</element>
   <element>provisioning_api</element>
   <element>systemtags</element>
   <element>systemtags_management</element>
   <element>testing</element>
   <element>updatenotification</element>
  </apps>
 </data>
</ocs>
```

2. `curl -k -u admin:admin -X GET http://172.17.0.1:8080/ocs/v1.php/cloud/apps?filter=disabled`
Another reasonable list is returned:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data>
  <apps>
   <element>files_classifier</element>
   <element>configreport</element>
...
   <element>files_mediaviewer</element>
  </apps>
 </data>
</ocs>
```
3. `curl -k -u admin:admin -X GET http://172.17.0.1:8080/ocs/v1.php/cloud/apps`
Query for all apps:
```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data>
  <apps>
   <element>files_classifier</element>
   <element>configreport</element>
...
   <element>files_mediaviewer</element>
  </apps>
 </data>
</ocs>

```

### Expected behaviour
All the apps listed at (1) and (2) are in the combined list at (3)

### Actual behaviour
The "always-enabled-apps" are missing from the list. The missing apps are the ones listed in https://github.com/owncloud/core/blob/master/core/shipped.json

```
  "alwaysEnabled": [
    "files",
    "files_external",
    "dav",
    "federatedfilesharing"
  ]
```

### Server configuration
Current core master.

phil-davis:
This seems a bit odd. 

[...]

PR #37885, opened 2020-09-04
[Tests-Only] Add test scenarios for get all and disabled apps with API

phil-davis:
## Description
Add tests to cover parts of the `cloud/apps` API that do not have test cases:
- get a list of all apps
- get a list of all disabled apps

## Related Issue
- helps to cover current behavior related to #37884 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiProvisioning-v1/getApps.feature
+++ b/tests/acceptance/features/apiProvisioning-v1/getApps.feature
@@ -38,3 +38,58 @@ Feature: get apps
       | files_sharing        |
       | updatenotification   |
       | files_external       |
+
+  @comments-app-required
+  Scenario: admin gets enabled apps when some app is disabled
+    Given app "comments" has been disabled
+    When the administrator gets all enabled apps using the provisioning API
+    Then the OCS status code should be "100"
+    And the HTTP status code should be "200"
+    And the apps returned by the API should include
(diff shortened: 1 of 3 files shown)


tests/acceptance/features/apiProvisioning-v1/getApps.feature
```diff
@@ -38,3 +38,27 @@ Feature: get apps
...
+    And the HTTP status code should be "200"
+    And the apps returned by the API should include
+      | comments             |
+    #  | dav                  |
+    #  | federatedfilesharing |
+      | federation           |
+    #  | files                |
+      | files_sharing        |
+      | updatenotification   |
+    #  | files_external       |
```
phil-davis, line 64: I didn't tag this scenario with the issue, because it is not yet decided if the issue is actually a bug, or the intended behavior.

individual-it (changes requested):
I would suggest to add the issue tag anyway, even if its not an issue we have the reference for later, just in case s.o. wonders one day why that behaviour is like described in the scenario

individual-it (approved):

PR #37888, opened 2020-09-06
[Tests-Only] Fix timing problem in testScanShareInvalidLastScan

phil-davis:
## Description
https://drone.owncloud.com/owncloud/core/26677/35/9
```
There was 1 failure:

1) OCA\Files_Sharing\Tests\External\ScanExternalSharesJobTest::testScanShareInvalidLastScan
Failed asserting that true matches expected false.

/drone/src/apps/files_sharing/tests/External/ScanExternalSharesJobTest.php:281
```

If the system time ticks over the 1-second boundary just after the test has remembered the time in to return in the mock of `getLastLogin` and the time that is put in `lastscan`, then `$now` in `shouldScan` becomes 1 second later. This happens to make both reasons to return false be false, and `shouldScan` unexpectedly returns true.

Putting a \sleep(1); just before:
`$this->invokePrivate($scanShares, 'shouldScan', [$share, $lastLoginThreshold, $lastScanThreshold]);`
Makes the test fail consistently.

Adjust `$lastScanThreshold` so that an accidental 1-second change in the system time will not make the test fail.

## How Has This Been Tested?
Local unit test runs

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/tests/External/ScanExternalSharesJobTest.php
+++ b/apps/files_sharing/tests/External/ScanExternalSharesJobTest.php
@@ -271,11 +271,11 @@ class ScanExternalSharesJobTest extends TestCase {
 			'token'	=> 'test',
 			'password' => 'test',
 			'mountpoint' => 'test',
-			'lastscan' => \time(),
+			'lastscan' => \time() - 10,
 			'owner'	=> 'test'
 		];
 		$lastLoginThreshold = '1';
-		$lastScanThreshold = '2';
+		$lastScanThreshold = '20';
 		$result = $this->invokePrivate($scanShares, 'shouldScan', [$share, $lastLoginThreshold, $lastScanThreshold]);
 
 		$this->assertEquals(false, $result);
--- a/apps/files_sharing/tests/External/ScanExternalSharesJobTest.php
+++ b/apps/files_sharing/tests/External/ScanExternalSharesJobTest.php
@@ -275,7 +275,7 @@ class ScanExternalSharesJobTest extends TestCase {
 			'owner'	=> 'test'
 		];
 		$lastLoginThreshold = '1';
-		$lastScanThreshold = '1';
+		$lastScanThreshold = '2';
 		$result = $this->invokePrivate($scanShares, 'shouldScan', [$share, $lastLoginThreshold, $lastScanThreshold]);
 
 		$this->assertEquals(false, $result);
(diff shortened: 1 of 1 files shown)


apps/files_sharing/tests/External/ScanExternalSharesJobTest.php
```diff
@@ -275,7 +275,7 @@ public function testScanShareInvalidLastScan() {
 			'owner'	=> 'test'
 		];
 		$lastLoginThreshold = '1';
-		$lastScanThreshold = '1';
+		$lastScanThreshold = '2';
```
phil-davis, line 278: Without this change, adding `\sleep(1)` here makes the test fail.
An unlucky tick of the system clock between setting `'lastscan' => \time(),` at line 274, and getting into the the code of `shouldScan` causes the test to fail.

mrow4a: perfectly fine, most important is that it is larger :) could be even larger, as `\time()+2` could still see issue.

Plus would be good to add -> `add 10 seconds to time() in order to force shouldScan return false`

mrow4a (changes requested):

apps/files_sharing/tests/External/ScanExternalSharesJobTest.php
phil-davis: Or I could make a more realistic "ordinary" case - set `$lastScanThreshold = '20'`  and `'lastscan' => \time() - 10,`

A share that has been scanned 10 seconds ago, and the last scan threshold is 20 seconds, should not be chosen for scanning.

The existing test here was checking exactly the edge case of the last scan time being just 1 second different from the threshold. But actually for this feature we don't care about that sort of edge case. e.g. if the threshold is set to 900 seconds (15 minutes) then if a share was last scanned 899, 900, or 901 seconds ago we really do not care if the one at exactly 900 seconds is chosen for scanning or not. The requirement does not need to specify and test that.

phil-davis: I did what I wrote above ^

mrow4a:
@phil-davis shalll we do the same with login ?

phil-davis:
> @phil-davis shall we do the same with login ?

`testScanShareInvalidLastLogin` does not have the same timing issue. It sets the user to have last logged in 2 seconds ago, then sets the last login threshold to 1 second, then calls `shouldScan`. If there is 1 or more seconds delay before `shouldScan` runs then it makes no difference to the test result. The user will look like they last logged in 3, 4, 5... seconds ago, the threshold is 1 second, so they will not be considered a "recent" user.

Do you think I should adjust those numbers anyway to be more "realistic"?

phil-davis:
@mrow4a or anyone, please review

mrow4a (approved):

PR #37902, opened 2020-09-11
[Tests-Only] Make PROPFIND response href test checks more flexible

### LINKED ISSUE #471: make steps that check the PROPFIND response less specific
OCIS storage tests like `apiOcisSpecific/apiWebdavProperties2-getFileProperties.feature:65` fail because the order of responses are different that expected

`And the value of the item "//d:response[2]/d:href" in the response to user "Alice" should match "/remote\.php\/<expected_href>\/file1.txt/"`
if the response has that file1.txt in the first item the test would fail, but the response actually is fine

phil-davis:
## Description
Make a more generic test step for checking for entries of files in a PROPFIND of a folder.
The step is more flexible so that the files do not have to be in a specific sorted order - they can be in any order in the list of files in the response.

## Related Issue
- Fixes https://github.com/owncloud/ocis-reva/issues/471

## How Has This Been Tested?
CI and local test runs

PR https://github.com/owncloud/ocis-reva/pull/474 to demonstrate this passes in `owncloud/ocis-reva`

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
+++ b/tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
@@ -618,6 +618,45 @@ class WebDavPropertiesContext implements Context {
 	}
 
 	/**
+	 * @Then there should be an entry with href matching :pattern in the response to user :user
+	 *
+	 * @param string $pattern
+	 * @param string $user
+	 *
+	 * @return void
+	 * @throws \Exception
+	 */
+	public function assertEntryWithHrefMatchingRegExpInResponseToUser($pattern, $user) {
+		$resXml = $this->featureContext->getResponseXmlObject();
+		if ($resXml === null) {
+			$resXml = HttpRequestHelper::getResponseXml(
+				$this->featureContext->getResponse(),
+				__METHOD__
+			);
+		}
+
+		$user = $this->featureContext->getActualUsername($user);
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
```diff
@@ -617,6 +617,45 @@ public function assertValueOfItemInResponseRegExp($xpath, $pattern) {
...
+			);
+		}
+
+		$user = $this->featureContext->getActualUsername($user);
+		$pattern = $this->featureContext->substituteInLineCodes(
+			$pattern, $user, ['preg_quote' => ['/']]
+		);
+
+		$index = 0;
+		while (true) {
```
PVince81, line 644: I suppose you could have used a "do...while" here

PVince81 (approved):
:+1:

tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
phil-davis: True, I didn't revisit the loop at the end of coding. As it turned out the test+break is at the end of the loop, so do-while would have worked. For now I will leave it as-is - we want to get this in and working.

PR #37913, opened 2020-09-15
[Tests-Only] Report HTTP status when file content is wrong

phil-davis:
## Description
https://github.com/cs3org/reva/pull/1170#issuecomment-692734081

If the downloaded content of a file is wrong (often if it is empty) and the test step fails, then report the HTTP status of the download response (as well as the difference between the actual and expected content). That can be helpful to whoever is trying to work out what was the problem.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


PVince81 (approved):
:+1: I missed that a lot

PR #37917, opened 2020-09-18
[Tests-Only] Do not overwrite user information when testing duplicate user creation

phil-davis:
## Description
Some test scenarios purposely try to create a user twice, to check that the 2nd user create request fails. They do this with different case of `userid` (username), because you should not be able to create both `brand-new-user` and `BRAND-NEW-USER`

The test code remembers the users that have been created, so it can delete those users in the `afterScenario` cleanup of the test. But some test steps are remembering "again" the 2nd version of the `userid` and overwriting the first value. That means that, for example, the actual user on the system-under-test is `brand-new-user` but the `afterScenario` tries to delete `BRAND-NEW-USER`. Actually that will work fine on a system-under-test that has a "correct" Provisioning API "delete" method. But on a system-under-test that is in development, a small problem in the Provisioning API can cause "bonus" test failures that impact on later scenarios.

In `afterScenario` we should try as best we can to use the "lowest-risk" API calls to do the cleanup actions that reset the system back to a known state.

In this PR the code is more careful that if the user create was not successful then to not overwrite any memory it already has about the existence of that user. That will ensure that deleting users in the `afterScenario` is done using the exact same `userid` that was used to create the user.

This should help with https://github.com/owncloud/ocis/pull/587 - preventing the "duplicate user creation" tests from messing up later test scenarios.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/FeatureContext.php
+++ b/tests/acceptance/features/bootstrap/FeatureContext.php
@@ -1414,6 +1414,15 @@ class FeatureContext extends BehatVariablesContext {
 	}
 
 	/**
+	 *
+	 * @return bool
+	 */
+	public function theHTTPStatusCodeWasSuccess() {
+		$statusCode = $this->response->getStatusCode();
+		return (($statusCode >= 200) && ($statusCode <= 299));
+	}
+
+	/**
 	 * Check the text in an HTTP reason phrase
 	 *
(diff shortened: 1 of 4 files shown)


kulmann (approved):
LGTM 🚀

PR #37919, opened 2020-09-18
[Tests-Only] Refactor reshare2 test suite to use root and share folder

dpakach:
## Description

Split `apiReshare2` test suite to 2 different test suites, one that uses the share folder and one in which shares are received in the root.

## Related Issue

- related https://github.com/owncloud/ocis/issues/518

## Motivation and Context

Receiving shares in the root folder is not available in ocis so this needs to be done so we can run the sharing tests in ocis.

## How Has This Been Tested?

:robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis (approved):

phil-davis:
@dpakach while reviewing, I update the test expectations in some test steps. For the "negative" tests, when the reshare is expected to have failed, then we need to be careful about the `Then file does not exist` steps. There are billions of files that should not exist!

Maybe we should check multiple possibilities that might exist - e.g. have 2 steps and check that `/textfile0.txt` does not exist and that `/Shares/textfile0.txt` does not exist?

PR #37943, opened 2020-09-24
[Tests-Only] Have separate scenarios for public link share of root folder on oC10 and OCIS

phil-davis:
## Description
On ownCloud10 a user cannot make a public link share of their root folder. That behaviour stays the same on ownCloud10, but will be different on OCIS. So tag the ownCloud10 scenario with `notToImplementOnOCIS`

On OCIS a user will be allowed to make a public link share of their root folder. Write a new scenario for the expected OCIS behaviour. Tag it `skipOnOcV10` because that is not the expected behaviour on oC10.

## Related Issue
Implements API tests for https://github.com/owncloud/ocis-reva/issues/283

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
+++ b/tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
@@ -608,8 +608,8 @@ Feature: create a public link share
       | 1               | 100             |
       | 2               | 200             |
 
-  @issue-ocis-reva-283
-  Scenario Outline: Do not allow public sharing of the root
+  @issue-ocis-reva-283 @notToImplementOnOCIS
+  Scenario Outline: Do not allow public sharing of the root on ownCloud10
     Given using OCS API version "<ocs_api_version>"
     When user "Alice" creates a public link share using the sharing API with settings
       | path | / |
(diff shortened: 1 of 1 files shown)


phil-davis:
For discussion: what approach should we take when there is an agreed difference in behaviour between implementations of the OCS APIs? In this case the implementations are ownCloud10 and OCIS. There might be an agreed difference between OCIS and `cs3org/reva` for something. There might be some other implementation in the future...

At the moment, the API tests here in `owncloud/core` are being used as the "defined standard" to make pass in order to be a "good implementation" of the OCS and WebDav APIs.

Approaches we could take when the "defined standard" has been agreed to be different for different implementations include:

1) do what I have done here - add the different scenarios to the "common" test suite with tags to indicate which implementations do or do not implement the scenario.

2) keep just the oC10 behaviour in `owncloud/core`, put the scenarios for the different behaviour into the repos with the other implementation (e.g. this new scenario could be put inot `owncloud/ocis`)

individual-it (approved):
I would vote for the 1. approach like done in this PR. Then we know that all the "good" tests are in this repo

PR #37860, opened 2020-08-30
Command to troubleshoot invalid shares generated by transfer ownership

mrow4a:
Adds new commands that detect and fix invalid shares generated by transfer ownership. 

- [x] add logging of invalid shares
- [x] add option to fix invalid shares
- [x] decide whether we want to release this as command (as is), or as update repair steps, or just as patch for specific customers -> and add unit tests if first 2 options (@micbar @pmaier1 )
- [ ] add docs - see issue https://github.com/owncloud/docs/issues/2794
- [x] add unit and integration tests

**1. Invalid initiator**

This has been triggered by past bug, that has been fixed in https://github.com/owncloud/core/pull/37791

```
$ occ files:troubleshoot-transfer-ownership invalid-initiator

SEARCHING FOR RESHARES THAT HAVE INVALID UID_INITIATOR(RESHARER) - RESHARER WHICH DOES NOT HAVE HIS SHARE MOUNTED ANY MORE AND THUS SHARE IS CORRUPTED
Invalid reshare initiator found for user test4 in shares: 
id=2,file_source=8,uid_owner=test1,uid_initiator=test4,share_with=test3
FOUND 1 INVALID INITIATOR RESHARES
```
  

**2. Invalid owner_uid (detached shares from their file storages, these would not show in UI to the user unless adjusted)**

This could have been created for three reasons I can think of:
- mismatching initiator (above case 1) somehow lead to this
- fact that `Share\IManager->transferShares` is not wrapped with transaction, and thus any termination in the middle could have resulted in mismatch between `share. uid_owner` and `storages.id` 
- past bug that is yet not identified or has been already fixed

```
$ occ files:troubleshoot-transfer-ownership invalid-owner

SEARCHING FOR SHARES THAT HAVE INVALID UID_OWNER - UID_OWNER VALUES THAT HAVE NOT BEEN UPDATED TO MATCH REQUIRED TRANSFERRED STORAGE
Invalid shares:
id=1,file_source=8,storage=home::test1,uid_owner=test4,uid_initiator=test4,share_with=test2
FOUND 1 INVALID SHARE OWNERS
```

Related: https://github.com/owncloud/enterprise/issues/4121

PR #37950, opened 2020-09-27
allow scope by user in troubleshoot transfer

mrow4a:
Extension of https://github.com/owncloud/core/pull/37860 with additional option for scoping for user

- [x] add docs issue https://github.com/owncloud/docs/issues/2794

```
occ files:troubleshoot-transfer-ownership all  --uid <userid>
```

Related: https://github.com/owncloud/enterprise/issues/4121

### DIFF
--- a/apps/files/lib/Command/TroubleshootTransferOwnership.php
+++ b/apps/files/lib/Command/TroubleshootTransferOwnership.php
@@ -70,12 +70,20 @@ class TroubleshootTransferOwnership extends Command {
 				'f',
 				InputOption::VALUE_NONE,
 				'perform auto-fix for found problems'
+			)
+			->addOption(
+				'uid',
+				'u',
+				InputArgument::OPTIONAL,
+				'scope for particular user',
+				null
 			);
 	}
 
 	public function execute(InputInterface $input, OutputInterface $output) {
 		$type = $input->getArgument('type');
 		$fix = $input->getOption('fix');
+		$scopeUid = $input->getOption('uid');
 
 		$allowedOps = \explode("|", $this->allowedOps);
 		if (!\in_array($type, $allowedOps)) {
@@ -86,10 +94,10 @@ class TroubleshootTransferOwnership extends Command {
 		}
 
 		if ($type == 'all' || $type == 'invalid-initiator') {
-			$this->findInvalidReshareInitiator($input, $output, $fix);
+			$this->findInvalidReshareInitiator($input, $output, $fix, $scopeUid);
 		}
 		if ($type == 'all' || $type == 'invalid-owner') {
-			$this->findInvalidShareOwner($input, $output, $fix);
+			$this->findInvalidShareOwner($input, $output, $fix, $scopeUid);
 		}
 
 		return 0;
@@ -103,8 +111,9 @@ class TroubleshootTransferOwnership extends Command {
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
 	 * @param boolean $fix
+	 * @param $scopeUid
 	 */
-	protected function findInvalidShareOwner(InputInterface $input, OutputInterface $output, $fix) {
+	protected function findInvalidShareOwner(InputInterface $input, OutputInterface $output, $fix, $scopeUid) {
 		$output->writeln([
 			"<info>==========================</>",
 			"<info>Searching for shares that have invalid uid_owner, meaning share uid_owner that does not match associated file storage id.</>",
@@ -115,8 +124,8 @@ class TroubleshootTransferOwnership extends Command {
 		$invalidSharesCount = 0;
 
 		$shareStorages = \array_merge(
-			$this->getAllInvalidShareStorages('home::'),
-			$this->getAllInvalidShareStorages('object::user:')
+			$this->getAllInvalidShareStorages('home::', $scopeUid),
+			$this->getAllInvalidShareStorages('object::user:', $scopeUid)
 		);
 
 		$tableRows = [];
@@ -156,7 +165,7 @@ class TroubleshootTransferOwnership extends Command {
 				}
 
 				if ($shareToFix['share_with'] !== null && $userIdFromStorage == $shareToFix['share_with']) {
-					$this->deleteCorruptedShare($shareToFix);
+					$this->deleteCorruptedShare($shareToFix['share_id'], (int) $shareToFix["share_type"]);
 				} else {
 					$this->adjustShareOwner($shareToFix['share_id'], $userIdFromStorage);
 				}
@@ -177,8 +186,9 @@ class TroubleshootTransferOwnership extends Command {
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
 	 * @param boolean $fix
+	 * @param $scopeUid
 	 */
-	protected function findInvalidReshareInitiator(InputInterface $input, OutputInterface $output, $fix) {
+	protected function findInvalidReshareInitiator(InputInterface $input, OutputInterface $output, $fix, $scopeUid) {
 		$output->writeln([
 			"<info>==========================</>",
 			"<info>Searching for reshares that have invalid uid_initiator(resharer), meaning resharer which does not have the received share mounted anymore (that he reshared with other user).</>",
@@ -188,7 +198,12 @@ class TroubleshootTransferOwnership extends Command {
 
 		$runExceptions = 0;
 		$invalidReshareInitiatorCount = 0;
-		$resharers = $this->getAllResharers();
+
+		if ($scopeUid == null) {
+			$resharers = $this->getAllResharers();
+		} else {
+			$resharers = [['uid_initiator' => $scopeUid]];
+		}
 
 		$resharesToFix = [];
 		foreach ($resharers as $resharer) {
@@ -201,6 +216,7 @@ class TroubleshootTransferOwnership extends Command {
 			]);
 			$tableRows = [];
 			try {
+				$this->setupFS($resharerUid);
 				$userFolder = $this->getUserFolder($resharerUid);
 
 				// extract all reshares for this user and check if they have mount node
@@ -218,6 +234,7 @@ class TroubleshootTransferOwnership extends Command {
 					$table->render();
 					$output->writeln("");
 				}
+				$this->tearDownFS();
 			} catch (\Exception $e) {
 				$runExceptions = $runExceptions + 1;
 				$table->setRows($tableRows);
@@ -225,9 +242,9 @@ class TroubleshootTransferOwnership extends Command {
 				$output->writeln("<info>Encountered error for user {$resharer['uid_initiator']}, command might need to be retried: </info>");
 				$output->writeln("<error>{$e->getMessage()}: </error>");
 				$output->writeln("<error>{$e->getTraceAsString()}: </error>");
-				$output->writeln("<info>Waiting 1 seconds after error before continuing with next user..</info>");
+				$output->writeln("<info>Waiting 5 seconds after error before continuing with next user..</info>");
 				$output->writeln("");
-				\sleep(1);
+				\sleep(5);
 			}
 		}
 
@@ -287,7 +304,7 @@ class TroubleshootTransferOwnership extends Command {
 		return $reshares;
 	}
 
-	protected function getAllInvalidShareStorages($storageType) {
+	protected function getAllInvalidShareStorages($storageType, $scopeUid) {
 		$query = $this->connection->getQueryBuilder();
 
 		if ($this->connection->getDatabasePlatform() instanceof MySqlPlatform) {
@@ -319,6 +336,10 @@ class TroubleshootTransferOwnership extends Command {
 				$concatFunction,
 				'st.id'
 			));
+
+		if ($scopeUid !== null) {
+			$query->andWhere($query->expr()->eq('s.uid_owner', $query->createNamedParameter($scopeUid)));
+		}
 
 		$cursor = $query->execute();
 		$shareStorages = $cursor->fetchAll();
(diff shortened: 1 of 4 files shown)


mmattel:
doc relevant! pls file a issue/pr in docs when approved/merged.

mrow4a:
@mmattel there is already https://github.com/owncloud/docs/issues/2794 . Will do after merging this one

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -188,7 +198,12 @@ protected function findInvalidReshareInitiator(InputInterface $input, OutputInte
 
 		$runExceptions = 0;
 		$invalidReshareInitiatorCount = 0;
-		$resharers = $this->getAllResharers();
+
+		if ($scopeUid == null) {
+			$resharers = $this->getAllResharers();
+		} else {
+			$resharers = [['uid_initiator' => $scopeUid]];
```
jvillafanez, line 205: This might need a code comment. I guess we expect the `resharers` to be a list of resharers, which in this particular case we only have one resharer with a fixed `uid_initiator` key because we don't need anything else.
Just looking a the changes it might be confusing because you could expect `['uid_initiator' => .....]` instead.

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -218,16 +234,17 @@ protected function findInvalidReshareInitiator(InputInterface $input, OutputInte
 					$table->render();
 					$output->writeln("");
 				}
+				$this->tearDownFS();
```
jvillafanez, line 237: This should be in a "finally" block

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -218,16 +234,17 @@ protected function findInvalidReshareInitiator(InputInterface $input, OutputInte
...
 				$table->setRows($tableRows);
 				$table->render();
 				$output->writeln("<info>Encountered error for user {$resharer['uid_initiator']}, command might need to be retried: </info>");
 				$output->writeln("<error>{$e->getMessage()}: </error>");
 				$output->writeln("<error>{$e->getTraceAsString()}: </error>");
-				$output->writeln("<info>Waiting 1 seconds after error before continuing with next user..</info>");
+				$output->writeln("<info>Waiting 5 seconds after error before continuing with next user..</info>");
 				$output->writeln("");
-				\sleep(1);
+				\sleep(5);
```
jvillafanez, line 247: This definitely requires a comment explaining why it is needed.
Sleep is bad and we don't want the user to wait. If I don't see any code comment explaining why the sleep is needed, I'll probably remove it at some point.

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -188,7 +198,12 @@ protected function findInvalidReshareInitiator(InputInterface $input, OutputInte
 
 		$runExceptions = 0;
 		$invalidReshareInitiatorCount = 0;
-		$resharers = $this->getAllResharers();
+
+		if ($scopeUid == null) {
```
jvillafanez, line 202: `===` by default or include a comment explaining this is intentional

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -156,7 +165,7 @@ protected function findInvalidShareOwner(InputInterface $input, OutputInterface
 				}
 
 				if ($shareToFix['share_with'] !== null && $userIdFromStorage == $shareToFix['share_with']) {
-					$this->deleteCorruptedShare($shareToFix);
+					$this->deleteCorruptedShare($shareToFix['share_id'], (int) $shareToFix["share_type"]);
```
jvillafanez, line 168: The typecasting might need a comment. I expect the "share_type" to be an int already, so there shouldn't be any need to typecast.

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -361,13 +382,21 @@ protected function adjustShareOwner($shareId, $newUidOwner) {
...
+		$shareProvider = $this->shareProviderFactory->getProviderForType($shareType);
+		$share = $shareProvider->getShareById($shareId);
 		$shareProvider->delete($share);
 	}
 
 	protected function getUserFolder($uid) {
 		return \OC::$server->getUserFolder($uid);
 	}
+
+	protected function tearDownFS() {
```
jvillafanez, line 395: I see several things wrong with this approach:
* Creating a method to just call a different method feels weird when you could just call the second method directly instead. I don't see any gain.
* Creating a protected method in a class that isn't expected to be extended doesn't look good. From my point of view, you use a protected method because either you are implementing a method from the parent class, or you expect a child class to implement the method.
* For testing, if the idea is to mock these method, I think the approach is wrong too: it's difficult to justify test the class when you have to mock parts of the class. You mock the dependencies; the methods are part of the class under testing.

If the idea of these methods is to test the class, it's better to move them to a Utils / Helper class and inject an instance of that class here, as a dependency, so you can mock it in a standard way.

mrow4a: @jvillafanez I know, the problems these are already utility classes. I will see if I can somehow make them being easily mocked instead of using protected.  

And yes, protected I used in whole this code to be able to easily mock -> I wanted to have minimal testing on this command not to spend too much effort on covering it all.

apps/files/lib/Command/TroubleshootTransferOwnership.php
mrow4a: will refactor and comment, makes sense

apps/files/lib/Command/TroubleshootTransferOwnership.php
mrow4a: @jvillafanez I tried to properly mock it and I have to admin I failed. There are two ways to solve this problem:
- create dedicated class for troubleshoot-transfer-ownership that would hide all DB and static util classes that is very hard to mock otherwise
- use DB directly

Both of the options create quite time consuming complexity. I suggest on just adding notes why protected has been used. Static util classes and command classes are very specific pattern in owncloud, that are complicated to test without above mentioned approaches.

phil-davis:
https://drone.owncloud.com/owncloud/core/27166/97/18
Needs a rebase to current master - the API acceptance test suites have been split in the last few days.

apps/files/lib/Command/TroubleshootTransferOwnership.php
mrow4a: One more note - I totally agree that with normal class using protected is antipattern. But Command/Job is a bit specific, and in perfect scenario command implementing class should inject another “logic” class and just call its methods (like e.g. getAllResharers). Thus here is more decision on clean vs effort

mrow4a:
Note: this command was used mostly to solve one-time issue for one of customers. We decided to publish it to broader community (but I think it requires good balance between value and effort).

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -156,9 +167,16 @@ protected function findInvalidShareOwner(InputInterface $input, OutputInterface
 				}
 
 				if ($shareToFix['share_with'] !== null && $userIdFromStorage == $shareToFix['share_with']) {
-					$this->deleteCorruptedShare($shareToFix);
+					try {
+						// delete corrupted share
+						// cast share_type to int just to be sure as we extracted it directly with PDO
+						$this->deleteCorruptedShare($shareToFix['share_id'], (int) $shareToFix["share_type"]);
```
jvillafanez, line 173: should we show some message here?

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -156,9 +167,16 @@ protected function findInvalidShareOwner(InputInterface $input, OutputInterface
...
+					} catch (ProviderException|ShareNotFound $e) {
+						$output->writeln("<error>Internal error while attempting to remove share {$shareToFix['share_id']}: {$e->getMessage()}</>");
+					}
 				} else {
-					$this->adjustShareOwner($shareToFix['share_id'], $userIdFromStorage);
+					// cast share type to int just to be sure, as we extracted it directly with PDO
+					$this->adjustShareOwner($shareToFix['share_id'], (int) $shareToFix["share_type"], $userIdFromStorage);
```
jvillafanez, line 179: same as above, any message we should show?

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -177,8 +195,9 @@ protected function findInvalidShareOwner(InputInterface $input, OutputInterface
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
 	 * @param boolean $fix
+	 * @param $scopeUid
```
jvillafanez, line 198: This needs to be adjusted, specially stating that "null" is accepted and what is the expected behaviour if "null" is used

jvillafanez:
```
	// the wrapping loop is performing very resource intensive tasks of mounting
	// oc filesystem for each of the users, that could cause timeouts,
	// sleep for 5 seconds not to overload
```
I'm still not fully convinced with this explanation... if we don't want to overload, we should probably apply the same mechanism after a successful loop, not just when an error happens. In addition, it should probably the admin the one to decide how much time he wants to wait; the server might be down for maintenance, so the admin won't likely want to wait.

mrow4a:
@jvillafanez I adjusted to review, good points.

apps/files/lib/Command/TroubleshootTransferOwnership.php
```diff
@@ -156,13 +168,20 @@ protected function findInvalidShareOwner(InputInterface $input, OutputInterface
 				}
 
 				if ($shareToFix['share_with'] !== null && $userIdFromStorage == $shareToFix['share_with']) {
-					$this->deleteCorruptedShare($shareToFix);
+					try {
+						// delete corrupted share
+						// cast share_type to int just to be sure as we extracted it directly with PDO
+						$this->deleteCorruptedShare($shareToFix['share_id'], (int) $shareToFix["share_type"]);
+						$output->writeln("<info>Deleted corrupted share with id={$shareToFix['share_id']}</>");
```
jvillafanez, line 175: Should it end with `</info>`?

In addition, I don't know if we can identify the share in some other way. Even if you say that the share 34 has been deleted, I won't know which share you're referring to, and the share has been already deleted so I can't check in the DB. It's better than nothing though.

jvillafanez:
I think that is the last change from my side. Need to check the CI error, but otherwise it looks good.

phil-davis:
https://drone.owncloud.com/owncloud/core/27203/114/14
```
  Scenario: troubleshoot transfer ownerships for all with share and reshare                                 # /drone/src/tests/acceptance/features/cliMain/transfer-ownership.feature:313
    Given user "Alice" has been created with default attributes and skeleton files                          # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Brian" has been created with default attributes and skeleton files                            # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Carol" has been created with default attributes and skeleton files                            # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "David" has been created with default attributes and skeleton files                            # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Alice" has created folder "/testByAlice"                                                      # FeatureContext::userHasCreatedFolder()
    And user "Alice" has shared folder "/testByAlice" with user "Brian" with permissions "all"              # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Brian" has shared folder "/testByAlice" with user "Carol" with permissions "all"              # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Brian" has created folder "/testByBrian"                                                      # FeatureContext::userHasCreatedFolder()
    And user "Brian" has shared folder "/testByBrian" with user "Carol" with permissions "all"              # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And the administrator transfers ownership from "Brian" to "David" using the occ command                 # TransferOwnershipContext::theAdministratorTransfersOwnershipFromToUsingTheOccCommand()
    When the administrator troubleshoots transfer of ownerships for "all" using the occ command             # TransferOwnershipContext::theAdministratorTroubleshootsTransferOwnershipUsingTheOccCommand()
    Then the command should have been successful                                                            # OccContext::theCommandShouldHaveBeenSuccessful()
    And the command output should contain the text "Searching for reshares that have invalid uid_initiator" # OccContext::theCommandOutputContainsTheText()
    And the command output should contain the text "Found 0 invalid initiator reshares"                     # OccContext::theCommandOutputContainsTheText()
    And the command output should contain the text "Repaired 0 invalid initiator reshares"                  # OccContext::theCommandOutputContainsTheText()
      The command output did not contain the expected text on stdout 'Repaired 0 invalid initiator reshares'
      The command output on stdout was:
      ==========================
      Searching for reshares that have invalid uid_initiator(resharer), meaning resharer which does not have the received share mounted anymore (that he reshared with other user).
      ==========================
      
      Found 0 invalid initiator reshares
      
      ==========================
      Searching for shares that have invalid uid_owner, meaning share uid_owner that does not match associated file storage id.
      ==========================
      
      Found 0 invalid share owners
      
      
      Failed asserting that 0 is equal to 1 or is greater than 1.
```

I guess that you removed the output "Repaired 0 invalid initiator reshares" becuase it is not relevant when there 0 shares have been found. The step that checks this just needs to be removed from the scenario.

apps/files/lib/Command/TroubleshootTransferOwnership.php
mrow4a: @jvillafanez all the shares are printed with very high detail to stdout in a step above (actual troubleshooting)

micbar:
@mrow4a Can we merge this very soon? We are planning a 10.6 release branch this week.

jvillafanez (approved):

mmattel:
@mrow4a as it is merged now, pls check: https://github.com/owncloud/core/pull/37950#issuecomment-700169361 to fix docs

PR #37951, opened 2020-09-27
with user sync command error properly exit and log

mrow4a:
When running `./occ user:sync` make sure to exit on error with non-zero, plus add log to console informing on error

- [x] add tests
- [x] add changelog

Related: https://github.com/owncloud/enterprise/issues/4218

### DIFF
--- a/core/Command/User/SyncBackend.php
+++ b/core/Command/User/SyncBackend.php
@@ -122,7 +122,7 @@ class SyncBackend extends Command {
 	/**
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
-	 * @return int|null
+	 * @return int
 	 */
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		if ($input->getOption('list')) {
@@ -169,19 +169,27 @@ class SyncBackend extends Command {
 		$uid = $input->getOption('uid');
 
 		if ($uid) {
-			$this->syncSingleUser($input, $output, $syncService, $backend, $uid, $missingAccountsAction);
+			$syncSuccess = $this->syncSingleUser($input, $output, $syncService, $backend, $uid, $missingAccountsAction);
 		} else {
-			$this->syncMultipleUsers($input, $output, $syncService, $backend, $missingAccountsAction);
+			$syncSuccess = $this->syncMultipleUsers($input, $output, $syncService, $backend, $missingAccountsAction);
+		}
+
+		if (!$syncSuccess) {
+			return 1;
 		}
 		return 0;
 	}
 
 	/**
+	 * Sync multiple users. Returns true if all users
+	 * were successfully synced, and returns false if at least 1 user had error
+	 *
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
 	 * @param SyncService $syncService
 	 * @param UserInterface $backend
 	 * @param string $missingAccountsAction
+	 * @return bool
 	 */
 	private function syncMultipleUsers(
 		InputInterface $input,
@@ -217,31 +225,48 @@ class SyncBackend extends Command {
 			$iterator = new BackendUsersIterator($backend);
 		}
 
-		$p = new ProgressBar($output);
+		$progress = new ProgressBar($output);
 		$max = null;
 		if ($backend->implementsActions(\OC_User_Backend::COUNT_USERS) && $input->getOption('showCount')) {
 			/* @phan-suppress-next-line PhanUndeclaredMethod */
 			$max = $backend->countUsers();
 		}
-		$p->start($max);
+		$progress->start($max);
 
-		$syncService->run($backend, $iterator, function () use ($p) {
-			$p->advance();
+		$syncErrors = [];
+		$syncService->run($backend, $iterator, function ($uid, $syncError) use ($progress, &$syncErrors) {
+			if ($syncError !== null) {
+				// NOTE: we cannot output here exception as this would break the progress bar
+				$syncErrors[] = "Sync error for {$uid}: {$syncError->getMessage()}";
+			}
+			$progress->advance();
 		});
+		$progress->finish();
+		$output->writeln('');
 
-		$p->finish();
+		foreach ($syncErrors as $syncError) {
+			$output->writeln("<error>$syncError</error>");
+		}
+
 		$output->writeln('');
 		$output->writeln('');
+
+		$syncErrorCount = \count($syncErrors);
+		$output->writeln("Sync of users finished, encountered $syncErrorCount errors.");
+		return $syncErrorCount === 0;
 	}
 
 	/**
+	 * Sync single user. Returns true if user
+	 * was successfully synced, and returns false if there was an error
+	 *
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
 	 * @param SyncService $syncService
 	 * @param UserInterface $backend
 	 * @param string $uid
 	 * @param string $missingAccountsAction
-	 * @throws \LengthException
+	 * @return bool
 	 */
 	private function syncSingleUser(
 		InputInterface $input,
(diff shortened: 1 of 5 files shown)


phil-davis:
Needs a rebase - https://drone.owncloud.com/owncloud/core/27168/97/18
The API acceptance test suites have been split up a few days ago. Somehow the force-push 8 hours ago was not to a current master.

core/Command/User/SyncBackend.php
```diff
@@ -169,11 +169,10 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		$uid = $input->getOption('uid');
 
 		if ($uid) {
-			$this->syncSingleUser($input, $output, $syncService, $backend, $uid, $missingAccountsAction);
+			return $this->syncSingleUser($input, $output, $syncService, $backend, $uid, $missingAccountsAction);
```
jvillafanez, line 172: These functions should be documented to return an error code

core/Command/User/SyncBackend.php
```diff
@@ -218,20 +217,32 @@ private function syncMultipleUsers(
 		}
 
 		$p = new ProgressBar($output);
+		$ec = 0;
```
jvillafanez, line 220: Better use meaningful variable names, as unique as possible.
In addition, I'd move this just above the `$syncService->run(...)` command because the value is closely attached to what the function is doing. You can think that the value is initialized within the function, and in won't be used before it.

core/Command/User/SyncBackend.php
```diff
@@ -218,20 +217,32 @@ private function syncMultipleUsers(
...
 		$max = null;
 		if ($backend->implementsActions(\OC_User_Backend::COUNT_USERS) && $input->getOption('showCount')) {
 			/* @phan-suppress-next-line PhanUndeclaredMethod */
 			$max = $backend->countUsers();
 		}
 		$p->start($max);
 
-		$syncService->run($backend, $iterator, function () use ($p) {
+		$syncService->run($backend, $iterator, function ($uid, $e) use ($p, &$ec) {
```
jvillafanez, line 228: This isn't easy to read due to the variable names.

core/Command/User/SyncBackend.php
```diff
@@ -218,20 +217,32 @@ private function syncMultipleUsers(
...
+			}
 			$p->advance();
 		});
 
 		$p->finish();
+
 		$output->writeln('');
 		$output->writeln('');
+
+		if ($ec > 0) {
```
jvillafanez, line 240: As an alternative which might have more sense: this function could return the number of sync errors (the `$ec` var) instead of an error code. It would be the `execute` method the one that would show the "Encountered sync errors for {$ec} users" message and return the corresponding error code.

lib/private/User/SyncService.php
```diff
@@ -125,21 +125,22 @@ private function checkIfAccountReappeared(Account $a, array &$removed, array &$r
...
 				$backendClass = \get_class($backend);
-				$this->logger->error("Error syncing user with uid: $uid and backend: $backendClass");
-				$this->logger->logException($e);
+				$this->logger->logException($e, ['message' => "Error syncing user with uid: $uid and backend: $backendClass"]);
 			}
 
 			// call the callback
 			if (\is_callable($callback)) {
-				$callback($uid);
+				$callback($uid, $currentException);
```
jvillafanez, line 143: What about using 2 additional optional callbacks?
`->run($backend, $userIds, $genericCallback = null, $successCallback = null, $errorCallback = null)`

The success callback would be call only after a successful user sync, and the error callback only if there is an error. We can keep the generic callback for compatibility.

```
try {
    sync($uid);
    $successCallback($uid);
} catch (\Exception $ex) {
    $errorCallback($uid, $ex);
}
$genericCallback($uid);
```

jvillafanez: If you prefer to keep just one callback function, I think it's cleaner to call it in different places:
```
try {
    sync($uid);
    $callback($uid, null);
} catch (\Exception $e) {
    $callback($uid, $e);
}
```
Note that the expectations for the callback must be clearly documented (which they aren't at the moment). It isn't documented that the uid of the user will be send as the first parameter in the callback, and that (now) an exception will be set as second parameter or null

mrow4a: @jvillafanez I did want to keep the compatibility, this is why I did it like that. But what I did not think of was genericCallback :) will have a look into it.

mrow4a:
@jvillafanez could you have a look after review fixes?

core/Command/User/SyncBackend.php
```diff
@@ -217,31 +225,48 @@ private function syncMultipleUsers(
...
+			$output->writeln("<error>$syncError</error>");
+		}
 
-		$p->finish();
 		$output->writeln('');
 		$output->writeln('');
+
+		$syncErrorCount = \count($syncErrors);
+		$output->writeln("Sync of users finished, encountered $syncErrorCount errors.");
+		return $syncErrorCount == 0;
```
jvillafanez, line 256: `===` by default

core/Command/User/SyncBackend.php
```diff
@@ -217,31 +225,48 @@ private function syncMultipleUsers(
...
-		$p->start($max);
+		$progress->start($max);
 
-		$syncService->run($backend, $iterator, function () use ($p) {
-			$p->advance();
+		$syncErrors = [];
+		$syncService->run($backend, $iterator, function ($uid, $syncError) use ($progress, &$syncErrors) {
+			if ($syncError !== null) {
+				// NOTE: we cannot output here exception as this would break the progress bar
+				$syncErrors[] = "Sync error for {$uid}: {$syncError->getMessage()}. Check error log for more details";
```
jvillafanez, line 240: That "Check error log" message will be annoying if there are 3 or more errors. We probably need to move it to a different place or remove it.

core/Command/User/SyncBackend.php
```diff
@@ -260,17 +285,30 @@ private function syncSingleUser(
...
-					throw new \LengthException("Multiple users returned from backend for: $uid. Cancelling sync.");
+					$output->writeln("<error>Multiple users returned from backend for: $uid. Cancelling sync.</error>");
+					return false;
 				}
 			}
 		}
 
 		$dummy = new Account(); // to prevent null pointer when writing messages
 
+		$syncSuccess = true;
```
jvillafanez, line 296: It seems better to move it inside the condition if the variable won't be used outside

core/Command/User/SyncBackend.php
```diff
@@ -260,17 +285,30 @@ private function syncSingleUser(
...
 		if ($userToSync !== null) {
 			// Run the sync using the internal username if mapped
 			$output->writeln("Syncing $uid ...");
-			$syncService->run($backend, new \ArrayIterator([$userToSync]));
+			$syncService->run($backend, new \ArrayIterator([$userToSync]), function ($uid, $syncError) use ($output, &$syncSuccess) {
+				if ($syncError !== null) {
+					$syncSuccess = false;
+					$output->writeln("<error>Encountered sync error for {$uid}: {$syncError->getMessage()}. Check error log for more details</error>");
```
jvillafanez, line 303: Same here with the "Check error log"

jvillafanez (approved):
Just a couple of little optional details, but approving anyway.

micbar:
should go in 10.6

core/Command/User/SyncBackend.php
mrow4a: makes sense

core/Command/User/SyncBackend.php
mrow4a: makes sense

core/Command/User/SyncBackend.php
mrow4a: overlooked, thanks

jvillafanez (approved):

karakayasemi (approved):

PR #37956, opened 2020-09-28
Retry moving mountpoint sane amount of times

VicDeo:
# Description
- Catch and handle integrity constraint violations and return `false` as expected
- Make 10 attempts to move the mountpoint

## Related Issue
- https://github.com/owncloud/enterprise/issues/4094

## Motivation and Context
a newly generated mounpoint path may exist but hasn't been mount yet. An attempt to move to this path will cause breaking constraint on index `sh_external_mp` for `oc_share_external` table

## How Has This Been Tested?
no exact reproduction steps provided :no_good:

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1388,6 +1388,18 @@ $CONFIG = [
 'filesystem_check_changes' => 0,
 
 /**
+ * Define unsuccessful mountpoint rename attempts
+ * This config value avoids infinite loops for seldom cases where a file renaming
+ * conflict between different share backends could occur.
+ * The value defines how many unsuccessful mountpoint rename attempts are allowed.
+ * e.g. target mountpoint name could be claimed as unused by the filesystem but
+ * renaming to this target name will fail due to some other reasons like database
+ * constraints.
+ * Change this value only under supervision of ownCloud support.
+ */
+'filesystem.max_mountpoint_move_attempts' => 10,
+
+/**
  * Define where part files are located
  * By default ownCloud will store the part files created during upload in the
  * same storage as the upload target. Setting this to false will store the part
--- a/apps/files_sharing/lib/External/Manager.php
+++ b/apps/files_sharing/lib/External/Manager.php
@@ -29,6 +29,7 @@
 
 namespace OCA\Files_Sharing\External;
 
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use OC\Files\Filesystem;
 use OC\User\NoUserException;
 use OCA\Files_Sharing\Helper;
@@ -315,7 +316,11 @@ class Manager {
 			WHERE `mountpoint_hash` = ?
 			AND `user` = ?
 		');
-		$result = (bool)$query->execute([$target, $targetHash, $sourceHash, $this->uid]);
+		try {
+			$result = (bool)$query->execute([$target, $targetHash, $sourceHash, $this->uid]);
+		} catch (UniqueConstraintViolationException $e) {
+			$result = false;
+		}
 
 		return $result;
 	}
--- a/apps/files_sharing/lib/External/Mount.php
+++ b/apps/files_sharing/lib/External/Mount.php
@@ -54,8 +54,9 @@ class Mount extends MountPoint implements MoveableMount {
 	 */
 	public function moveMount($target) {
 		$result = $this->manager->setMountPoint($this->mountPoint, $target);
-		$this->setMountPoint($target);
-
+		if ($result === true) {
+			$this->setMountPoint($target);
+		}
 		return $result;
 	}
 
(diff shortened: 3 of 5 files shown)


jvillafanez:
Is it possible to get the list of the mount points from the DB somehow? Otherwise, I agree we should make the amount of times configurable.

VicDeo:
@jvillafanez we know nothing about DB in the context of **lib/private/Files/Config/MountProviderCollection.php**
what we have there is an array of providers.

phil-davis:
https://drone.owncloud.com/owncloud/core/27320/139/15
```
--- Failed scenarios:

    /drone/src/tests/acceptance/features/webUISharingInternalGroups1/shareWithGroups.feature:133
    /drone/src/tests/acceptance/features/webUISharingInternalGroups1/shareWithGroups.feature:150
    /drone/src/tests/acceptance/features/webUISharingInternalGroups1/shareWithGroups.feature:170
    /drone/src/tests/acceptance/features/webUISharingInternalGroups1/shareWithGroups.feature:183

32 scenarios (28 passed, 4 failed)
471 steps (458 passed, 4 failed, 9 skipped)
```

This sort of problem was reported:
```
    Then a notification should be displayed on the webUI with the text "Email notification was sent!"              # WebUIGeneralContext::aNotificationShouldBeDisplayedOnTheWebUIWithTheText()
      WebUIGeneralContext::aNotificationShouldBeDisplayedOnTheWebUIWithTheText A notification was expected to be displayed on the webUI with the text 'Email notification was sent!', but got '' instead
      --- Expected
      +++ Actual
      @@ @@
      -'Email notification was sent!'
      +''
```

I don't see how the code in this PR could break that. I will restart drone...

phil-davis:
And this time it passed.

micbar:
@VicDeo Can we make the amount of retries configurable?

mmattel:
This is docs relevant (config-to-docs run) because of a change in `config/config.sample.php`.
I propose to give the description in `config/config.sample.php` a bit more text for better understanding what it is about.

VicDeo:
@mmattel frankly speaking I just made it configurable as requested. The only explanation I can add: "if you don't know what is it about then you most likely have no need to adjust it"

mmattel:
> if you don't know what is it about then you most likely have no need to adjust it

I completely agree. In case it gets merged, 3 months later, nobody will remember what it was about. Means, you do not only add a improved description for endusers, but also for ownCloud itself to prevent that.

VicDeo:
@mmattel  I don't know how to explain it properly

0 There could be a collision of the l mount point name for local and remote shares. E.g. when both are named `file`
1. In this case we move the mount point using the following rule - `file (1)`, `file (2)`, etc. The name is generated by checking whether `file (n)` exists **on the file system**.
2. if the storage goes offline for the remote share it could happen that `file (n)` does NOT exist **on the file system** but it's still not possible to move the mount point as we already have it in the database. So the move operation returns `false`
3. When we have `false` after the move  operation we consider `file (n)` name is unavailable and retry the move operation with `file (n+1)`. 
4. 3 is repeated exactly `filesystem.max_mountpoint_move_attempts`  times to avoid an infinite loop.

mmattel:
**What about:**
This config value avoids infinite loops for seldom cases where a file renaming conflict could occur.
The value defines how many unsuccessful rename attempts are allowed, if the mountpoint to a file is taken but the initiated rename failed. Change this value only under supervision of ownCloud support.

Pls check the text if valid and consistent and adopt if needed.

VicDeo:
@mmattel what about 
```
/**
 * This config value avoids infinite loops for seldom cases where a file renaming
 * conflict between different share backends could occur.
 * The value defines how many unsuccessful mountpoint rename attempts are allowed.
 * E.g. target mountpoint name could be claimed as unused by the filesystem but
 * renaming to this target name will fail due to some other reasons like database 
 * constraints.
 * Change this value only under supervision of ownCloud support.
 */
```
Is it clear enough?

mmattel:
This is great 👍 
One thing I would do: even when starting at the beginning of the sentence, I would write `E.g.`-->`e.g.`
Else it looks at least to my eyes awful :smile:

config/config.sample.php
```diff
@@ -1387,6 +1387,17 @@
  */
 'filesystem_check_changes' => 0,
 
+/**
+ * This config value avoids infinite loops for seldom cases where a file renaming
```
mmattel, line 1391:
```suggestion
 * Define unsuccessful mountpoint rename attempts
 * This config value avoids infinite loops for seldom cases where a file renaming
```
@VicDeo There was a "headline" missing. This headline will be automatically used as item text in the table of contents on top of the document created by the config-to-docs run. I forgot to mention this when making the first text proposal. This commit suggestion fixes it. If you agree, you can securly commit it.

phil-davis:
I don't see a changelog entry yet?

phil-davis (commented):
Looks OK to me now.
@jvillafanez review?

jvillafanez (approved):

PR #37962, opened 2020-09-30
[Tests-Only] Test on postgres:9.4

phil-davis:


phil-davis:
https://drone.owncloud.com/owncloud/core/27379/84/13
Some bug-demo scenarios for issue #23151 are not passing. PGSQL sometimes behaves differently for the order of things returned from the database, so that could explain it. Those are only bug-demo scenarios anyway, so we do not require that the bug exhibit the same behaviour on every database.

phil-davis:
#38058 should help - I will try again after that.

phil-davis:
#38060 is the next demo PR of testing with postgres:9.4

PR #37982, opened 2020-10-06
[Tests-Only] Send coverage to SonarCloud

phil-davis:
## Description
Analyse and upload to SonarCloud.

1) At the end of each JavaScript and PHPunit pipeline that collects coverage, sync the coverage report to an S3 cache
2) After all coverage-related pipelines are finished, sync the coverage reports from the S3 cache to a local pipeline `sonar-analysis`
3) Setup a "server" in that analysis pipeline so that the source code is there to analyse.
4) Run the Sonar analysis and let it upload to SonarCloud

SonarCloud analysis in drone is at https://drone.owncloud.com/owncloud/core/27145/34/9
The report is at https://sonarcloud.io/dashboard?id=owncloud_core&pullRequest=37982

The report shows 55.1% coverage. I was hoping that it would show about 64% like the old covedcov reports, e.g. https://github.com/owncloud/core/pull/37956#issuecomment-700255057

The analysis emits messages like:
```
INFO: Analyzing PHPUnit coverage report: results/clover-phpunit-php7.2-mariadb10.2.xml
WARN: Could not resolve 5098 file paths in clover-phpunit-php7.2-mariadb10.2.xml, first unresolved path: /drone/src/config/config.php
```

I have not been able to work out why the `WARN` is there. The source files are in `/drone/src` - I asked in chat https://community.sonarsource.com/t/how-to-get-total-coverage-from-multiple-unit-test-runs/31719/11 but no response.

Note: This is a cleaned-up, rebased and squashed version of PR #37946 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #37984, opened 2020-10-06
[Tests-Only] Make the excludes longer in sonar-project 

phil-davis:
On top of #37982 

See if this helps the coverage calculation.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -264,47 +264,70 @@ config = {
 }
 
 def main(ctx):
-	initial = initialPipelines()
+	initial = initialPipelines(ctx)
 
 	before = beforePipelines(ctx)
 	dependsOn(initial, before)
 
+	coverageTests = coveragePipelines(ctx)
+	if (coverageTests == False):
+		print('Errors detected in coveragePipelines. Review messages above.')
+		return []
+
+	dependsOn(before, coverageTests)
+
 	stages = stagePipelines(ctx)
 	if (stages == False):
-		print('Errors detected. Review messages above.')
+		print('Errors detected in stagePipelines. Review messages above.')
 		return []
 
 	dependsOn(before, stages)
 
-	after = afterPipelines()
-	dependsOn(stages, after)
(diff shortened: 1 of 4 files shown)


phil-davis:
Only gets 55.2% - that hardly made any difference from 55.1%

micbar:
> Only gets 55.2% - that hardly made any difference from 55.1%

Thanks for trying out.

So either
1) Codecov is wrong :-)
2) We are not getting every report.

phil-davis:
IMO there are some coverage-specific include/exclude parameters. I will look for the documentation on those and see if it can help.

micbar:
I cannot spot in the sonar-analysis log where it finds the `lcov.info` from the JS tests.

phil-davis:
`sonar.php.coverage.reportPaths` has `lcov.info` in it. @dpakach got the coverage to increase from 50.7% to 55.1% by adding that. So it did have an effect.

I suspect that more files are being included in the "total amount of code to cover" and that brings down the percentage coverage. But I don't actually know. It would be nice to see a detailed report online showing each file - it should be obvious from that if there are groups if files that are listed as 0% coverage.

phil-davis:
When I go to https://sonarcloud.io/code?id=owncloud_core&pullRequest=37984 and select the "Code" tab, it is empty.
Maybe we have to merge something first, and then it will have some "baseline coverage" in master and will start showing useful coverage diffs or?

micbar:
Yes. let us merge and see.

phil-davis:
"Yes. let us merge and see." - to get an initial set of `master` coverage information where we will be able to (hopefully)  browse around the results and see the detail of the coverage that SonarCloud is calculating.

I also found https://community.sonarsource.com/t/sonarqube-and-code-coverage/4725 - that has information about how the "coverage" metric is calculated by SonarCloud. It includes calculations of conditions that have evaluated to both true and false, as well as "basic line coverage". So that might also have some influence on the numbers.

After merging, we will see...

individual-it (approved):

PR #37993, opened 2020-10-07
don't treat Google Docs special for file size

### LINKED ISSUE #37977: "server reported no size" error for Google Drive docs in desktop client sync
### Steps to reproduce
1. Connect to Google Drive storage
2. Add Google Doc or spreadsheet into Google Drive connected storage
3. In desktop client, login to account with connected Google Drive

### Expected behaviour
Google Doc should sync without error

### Actual behaviour
Client reports sync error for every Google Doc file "server reported no size"

### Server configuration
**Operating system**: Ubuntu 18.04

**Web server:** Apache 2.4

**Database:** MySQL 5.7

**PHP version:** 7.2

**ownCloud version:** 10.5

**Updated from an older ownCloud or fresh install:** updated from 10.3.2 > 10.4.1 > 10.5 using updater found in admin web app.

**Signing status (ownCloud 9.0 and above):** `No errors have been found.`

**Are you using external storage, if yes which one:** Google Drive

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** No

### Client configuration
**Browser:** Safari 14.0

**Operating system:** macOS 10.14.6

curiousercreative:
## Description
Don't treat Google Docs (in Google Drive connected external storage) differently from other files with respect to file size. Under the hood, the Google Drive SDK will return undefined which likely converts to `0` in ownCloud server response.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/37977

## Motivation and Context
Bugfix, Google Docs were not syncing with desktop client and spamming error notifications.

## How Has This Been Tested?
- test environment:
  - on my production server (OC 15.0, Ubuntu 18.04, Apache 2.4, PHP 7.2, MySQL 5.7)
  - latest 15.0 docker image
- test case 1:
  1. connect to Google Drive storage
  1. add Google Doc
  1. connect client to server with Google Drive storage
  1. Expect Google Doc to sync

## Types of changes

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/lib/Lib/Storage/Google.php
+++ b/apps/files_external/lib/Lib/Storage/Google.php
@@ -369,13 +369,7 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 			if ($this->filetype($path) === 'dir') {
 				$stat['size'] = 0;
 			} else {
-				// Check if this is a Google Doc
-				if ($this->isGoogleDocFile($file)) {
-					// Return unknown file size
-					$stat['size'] = \OCP\Files\FileInfo::SPACE_UNKNOWN;
-				} else {
-					$stat['size'] = $file->getSize();
-				}
+				$stat['size'] = $file->getSize();
 			}
 			$stat['atime'] = \strtotime($file->getViewedByMeTime());
 			$stat['mtime'] = \strtotime($file->getModifiedTime());
--- a/changelog/unreleased/37993
+++ b/changelog/unreleased/37993
@@ -1,0 +1,6 @@
+Bugfix: Fix for Google Docs not syncing with error "server reported no size"
+
+Users with Google Drive connected external storage were previously subjected to a "server reported no size" error in desktop sync client for every Google Doc that attempted to sync. Additionally, the Google Doc would not be downloaded.
+
+https://github.com/owncloud/core/issues/37997
+https://github.com/owncloud/core/pull/37993


jvillafanez:
The problem was caused by https://github.com/owncloud/core/pull/37265/commits/9728492618550b62863fe5d57d837296a1bbea2f
That commit fixes a problem with downloading google docs files. The size were treated as 0, so the webdav interface sent a 0 byte file instead of the actual exported file from google. This PR will likely revert to the previous behaviour where you can't download google docs files.

Note that we're downloading an exported version of the file from google, so the file size is unknown.

curiousercreative:
@jvillafanez thanks for reviewing and explaining. There's still a gap in my understanding of what's happening here because with my patch applied, the desktop client is downloading the exported version of these files as expected, same as in OC 10.3.2. With my patch, our Google external storage would be returning `NULL` for a size instead of `-2` it seems.

jvillafanez:
Maybe the desktop client kind of ignores the size of the file during the download. Web browsers might behave differently and download only what has been set in the content-size header of the download (which would probably be 0 bytes).
Please double-check with different browsers using the web UI.

curiousercreative:
@jvillafanez ah, I hadn't thought to check in web UI. Checking now, it actually downloads a complete file in Safari 14.0, Chrome 80 and Firefox 81, all from macOS 10.14.6.

jvillafanez (approved):

jvillafanez:
Verified it fixes the problem and the documents can be downloaded

PR #37996, opened 2020-10-08
Adjust some sonarcloud items

phil-davis:
## Description
Fix some things reported by SonarCloud, so I can see what an "improved" report looks like.
Unfortunately the SonarCloud UI does not show issues that will be resolved by the code in the PR - you have to "hope for the best" and merge the PR in order to see if you fixed the issues. https://community.sonarsource.com/t/view-all-issues-in-a-branch-not-only-new-ones/12172/11

1) Adjust use of properties parameter in CommentNode.php - the parameter was ignored and immediately overwritten.
2) Do not re-use input parameter as local variable in sendFileIdHeader - renamed the local variable. But the parameter `$node` is not used (and never was used)
3) Comment out unreachable code in OcmController.php - what to do about this unreachable code? Was it a good example of what might be done in future? Or should it be deleted?
4) Do not re-use input parameter as local variable in manipulateStorageConfig - renamed the local variable. But the parameter `$user` is not used (and never was used)
5) Fixup operator precedence in ShareController showShare - remove a `!!` double-not (that was there to force `bool`). Fix the operator precedence by adding brackets so that the `bool` type is calculated directly. Note: the previous code did actually end up with correct answers, so there was no actual bug - just lucky :)
6) Standardize exception handling in Log/Manage.php validateTimezone - SonarCloud did not like having a method that just instantiates an object and does nothing with it. This method was doing that just for the side-effect that an exception would be thrown if the timezone is invalid. Re-organize the code so that it explicitly throws `InvalidArgumentException` when the timezone is invalid (similar to what happens for other validation in the same file)
7) Do not re-use input parameter as local variable in Output.php setCookie - rename the variable. The input parameter `$path` was never used, and is still not used.
8) Remove double !! usage in Comments/Manager.php save - and explain in a comment what is going on casting `getParentId()` to `bool`
9)  Simplify redundant sub-exxpressions in formatDateRelativeDay - The same expression was used twice in a conditional. I have no idea what the 2nd one was supposed to be. Also, I don't see where the method `formatDateRelativeDay` is called anyway.
10) Remove useless returns from CacheJail.php moveFromCache - The methods `move` and `moveFromCache` do not return a value anyway, so "returning their return value" is not useful. And anyway, `moveFromCache` is not documented to return anything.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/comments/lib/Dav/CommentNode.php
+++ b/apps/comments/lib/Dav/CommentNode.php
@@ -218,11 +218,13 @@ class CommentNode implements \Sabre\DAV\INode, \Sabre\DAV\IProperties {
 	 * conforming to the list of requested properties.
 	 * The Server class will filter out the extra.
 	 *
-	 * @param array $properties
+	 * @param array|null $properties requested properties or null for all
 	 * @return array
 	 */
 	public function getProperties($properties) {
-		$properties = \array_keys($this->properties);
+		if (($properties === null) || ($properties === [])) {
+			$properties = \array_keys($this->properties);
+		}
 
 		$result = [];
 		foreach ($properties as $property) {
--- a/apps/dav/lib/Connector/Sabre/FilesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesPlugin.php
@@ -428,9 +428,9 @@ class FilesPlugin extends ServerPlugin {
 		if (!$this->server->tree->nodeExists($filePath)) {
 			return;
 		}
-		$node = $this->server->tree->getNodeForPath($filePath);
-		if ($node instanceof \OCA\DAV\Connector\Sabre\Node) {
-			$fileId = $node->getFileId();
+		$nodeForPath = $this->server->tree->getNodeForPath($filePath);
+		if ($nodeForPath instanceof \OCA\DAV\Connector\Sabre\Node) {
+			$fileId = $nodeForPath->getFileId();
 			if ($fileId !== null) {
 				$this->server->httpResponse->setHeader('OC-FileId', $fileId);
 			}
--- a/apps/federatedfilesharing/lib/Controller/OcmController.php
+++ b/apps/federatedfilesharing/lib/Controller/OcmController.php
@@ -380,13 +380,6 @@ class OcmController extends Controller {
 						['message' => "Notification of type {$notificationType} is not supported"],
 						Http::STATUS_NOT_IMPLEMENTED
 					);
-
-					// owner or sender unshared a resource
-					$share = $this->ocmMiddleware->getValidShare(
-						$providerId, $notification['sharedSecret']
-					);
-					$this->fedShareManager->undoReshare($share);
-					break;
 				default:
 					return new JSONResponse(
 						['message' => "Notification of type {$notificationType} is not supported"],
--- a/apps/files_sharing/lib/Controllers/ShareController.php
+++ b/apps/files_sharing/lib/Controllers/ShareController.php
@@ -364,7 +364,7 @@ class ShareController extends Controller {
 			}
 		}
 
-		$shareTmpl['canDownload'] = !!($share->getPermissions() & \OCP\Constants::PERMISSION_READ > 0);
+		$shareTmpl['canDownload'] = ($share->getPermissions() & \OCP\Constants::PERMISSION_READ) > 0;
 		$shareTmpl['sharePermission'] = $share->getPermissions();
 		$shareTmpl['downloadURL'] = $this->urlGenerator->linkToRouteAbsolute('files_sharing.sharecontroller.downloadShare', ['token' => $token]);
 		$shareTmpl['shareUrl'] = $this->urlGenerator->linkToRouteAbsolute('files_sharing.sharecontroller.showShare', ['token' => $token]);
--- a/lib/private/DateTimeFormatter.php
+++ b/lib/private/DateTimeFormatter.php
@@ -124,7 +124,7 @@ class DateTimeFormatter implements \OCP\IDateTimeFormatter {
 	 * @return string Formatted relative date string
 	 */
 	public function formatDateRelativeDay($timestamp, $format = 'long', \DateTimeZone $timeZone = null, \OCP\IL10N $l = null) {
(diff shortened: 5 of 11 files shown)


apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -428,9 +428,9 @@ public function sendFileIdHeader($filePath, \Sabre\DAV\INode $node = null) {
 		if (!$this->server->tree->nodeExists($filePath)) {
 			return;
 		}
-		$node = $this->server->tree->getNodeForPath($filePath);
```
phil-davis, line 431: Note: input parameter `$node` was never used, and is not used. What to do about that?

phil-davis:
@micbar I guess we might as well fix this stuff and see what happens to the SonarCloud reports in master.

Please get someone to review this.

IMO none of these changes actually change real behavior. So there is no real bug that is fixed. It is effectively a re-factoring. What do we do about a changelog for this sort of thing?

micbar:
@jvillafanez Can you take a look at the code changes?

apps/comments/lib/Dav/CommentNode.php
```diff
@@ -218,11 +218,13 @@ public function propPatch(PropPatch $propPatch) {
 	 * conforming to the list of requested properties.
 	 * The Server class will filter out the extra.
 	 *
-	 * @param array $properties
+	 * @param array|null $properties requested properties or null for all
 	 * @return array
 	 */
 	public function getProperties($properties) {
-		$properties = \array_keys($this->properties);
+		if (($properties === null) || ($properties === [])) {
```
jvillafanez, line 225: As I understand the PHPDoc, if I request no properties, I should get an empty result, not a list with all the properties.
Either adjust the doc to clarify this behaviour, or adjust the code.

apps/dav/lib/Connector/Sabre/FilesPlugin.php
jvillafanez: This function is used as a callback for a SabreDav event. I'm haven't checked what is the expected callback signature, but I'd match whatever it is. Clarification in the PHPDoc would help.
I'm assuming that it's part of the expected callback signature, so I'd keep it.

apps/federatedfilesharing/lib/Controller/OcmController.php
```diff
@@ -381,12 +381,13 @@ public function processNotification($notificationType,
...
 					);
 
+					// This code was unreachable and has been commented out:
 					// owner or sender unshared a resource
-					$share = $this->ocmMiddleware->getValidShare(
-						$providerId, $notification['sharedSecret']
-					);
-					$this->fedShareManager->undoReshare($share);
-					break;
+					//$share = $this->ocmMiddleware->getValidShare(
```
jvillafanez, line 386: Maybe @VicDeo can clarify if this code was misplaced and should be executed before the return statement. Otherwise, I don't think we want to keep this code around.

lib/private/DateTimeFormatter.php
```diff
@@ -124,7 +124,7 @@ public function formatDate($timestamp, $format = 'long', \DateTimeZone $timeZone
 	 * @return string Formatted relative date string
 	 */
 	public function formatDateRelativeDay($timestamp, $format = 'long', \DateTimeZone $timeZone = null, \OCP\IL10N $l = null) {
-		if (\substr($format, -1) !== '*' && \substr($format, -1) !== '*') {
```
jvillafanez, line 127: I wonder how this did ever happen....

phil-davis: It has been there from the beginning - I guess it was missed in code review many years ago, and actually nothing is broken - there is just bonus work for the CPU  to do.

apps/comments/lib/Dav/CommentNode.php
phil-davis: In the PHPdoc above (which is also the same in the parent) it already says:
`If the array is empty, it means 'all properties' were requested.`

apps/dav/lib/Connector/Sabre/FilesPlugin.php
phil-davis: Yes, it is in the callback signature. For whatever reason, the `$node` parameter is never used in  this method. But I suppose we have to keep it in the method declaration so that the signature matches the expected.

apps/files_sharing/lib/Controllers/ShareController.php
```diff
@@ -364,7 +364,7 @@ public function showShare($token, $path = '') {
 			}
 		}
 
-		$shareTmpl['canDownload'] = !!($share->getPermissions() & \OCP\Constants::PERMISSION_READ > 0);
+		$shareTmpl['canDownload'] = (($share->getPermissions() & \OCP\Constants::PERMISSION_READ) > 0);
```
VicDeo, line 367: just
`$shareTmpl['canDownload'] = ($share->getPermissions() & \OCP\Constants::PERMISSION_READ) > 0;`
is enough

phil-davis: true - I just like having extra `()` around the whole of a "complex" expression so that it is super-clear that the whole expression on the right is evaluated and then the result is assigned to the variable on the left.

Because actually the assignment operator `=` appears in the operator precedence list https://www.php.net/manual/en/language.operators.precedence.php - and maybe in some language it might have higher precedence than `>` or `&` :)

I will remove the brackets.

apps/federatedfilesharing/lib/Controller/OcmController.php
phil-davis: I deleted the code, rather than leaving it commented-out.

phil-davis:
Rebased.
@jvillafanez @VicDeo I have addressed the comments. Please review again.

jvillafanez (approved):

phil-davis:
SonarCloud "bugs" went down from 153 to 144 after this was merged: https://sonarcloud.io/dashboard?id=owncloud_core
So the analysis on merge does do something - good.

PR #37997, opened 2020-10-08
[Tests-Only] Add a silly command to demonstrate SonarCloud results

phil-davis:
Do not merge!
I just want to see the SonarCloud analysis of this...

### DIFF
--- a/core/Command/User/Something.php
+++ b/core/Command/User/Something.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


core/Command/User/Something.php
```diff
@@ -0,0 +1,68 @@
...
+			$output->writeln('<error>User does not exist</error>');
+			return 1;
+		}
+
+		$this->silly($output, "first", "second");
+		$output->writeln('<info>Maybe something happened</info>');
+		return 0;
+	}
+	private function silly(OutputInterface $output, $param1, $param2) {
+		$param2 = "hardcoded string";
```
phil-davis, line 65: I expect SonarCloud to complain about overriding the parameter here.
And to report reduced code coverage, because I added this file and it has no tests - so the codecov of the new code should be 0% and the project codecov should be a little bit less %

But SonarCloud passes???

phil-davis: I setup a "Quality Gate" for `owncloud/core` in SonarCloud that should fail if there is >0 new bugs, or <55% coverage of new code. This PR should fail both of those checks.
I restarted drone - let's see...

phil-davis: It still says that it passed, and no mention of uncovered code or the extra "bug"

phil-davis:
It does not report or fail the way that I hoped.

PR #38005, opened 2020-10-13
[tests-only] new env. variable to set how files/folders are provisioned for tests

### LINKED ISSUE #223: adjust creating skeleton folder and user data deletion for API tests when using OCIS-storage
when using OCIS storage and creating a user, we cannot just dump files into the folder to emulate the skeleton folder as we do with the OC storage but we need to upload the items via WebDav.

to delete data of the user after deleting the user our current delete strategy does not work, (https://github.com/owncloud/core/blob/828f07802fa3b977fcdbf315f9671d5442288886/tests/TestHelpers/OcisHelper.php#L68) we have to use a delete command

To accomplish both that we have currently to set 
```
'TEST_OCIS':'true'
'OCIS_REVA_DATA_ROOT':''
'DELETE_USER_DATA_CMD':'rm -rf /srv/app/tmp/ocis/root/nodes/root/*'
```
see https://github.com/owncloud/ocis-reva/pull/466/files#diff-74dc60589e0610c07e491b31f8693b2bR112-R129

it would be better to simplify that, maybe have an other setting like `REVA_STORAGE_OC_DRIVER` and according to that `deleteRevaUserData` will delete the correct folder and choose to upload the skeleton via WebDAV

individual-it:
## Description
add a new switch to be able to set the strategy how to provision the skeleton folder when tests run on OCIS.

## Related Issue
fixes https://github.com/owncloud/product/issues/223

## Motivation and Context
currently the provisioning strategy depends on if a delete command is given or not, that is confusing.

## How Has This Been Tested?
https://github.com/owncloud/ocis/pull/691

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -842,10 +842,26 @@ trait Provisioning {
 	public function manuallyAddSkeletonFiles($usersAttributes) {
 		$skeletonDir = \getenv("SKELETON_DIR");
 		$revaRoot = \getenv("OCIS_REVA_DATA_ROOT");
+		$skeletonStrategy = \getenv("OCIS_SKELETON_STRATEGY");
+		if (!$skeletonStrategy) {
+			$skeletonStrategy = 'upload'; //slower, but safer, so make it the default
+		}
+		if ($skeletonStrategy !== 'upload' && $skeletonStrategy !== 'copy') {
+			throw new Exception(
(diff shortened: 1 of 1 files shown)


phil-davis (approved):

phil-davis:
@individual-it where are we documenting all these env vars that control test setup?

PR #38016, opened 2020-10-17
check public link already authenticated before checking password of link

### LINKED ISSUE #138: (1.1.0) Password Protected Public Link Subfolder issue
There is an issue when moving inside a subfolder and back when accessing a password protected public link.

1. create a folder and a subfolder
2. share the folder as a public link and add a password
3. access the public link, enter password, move to the subfolder and back up

expected
you can move from subfolder to the top folder

actual result
you get an error that the directory is unavailable. If you try again - you get a warning that you tried it too many times.

a downgrade to the previous version (1.0.1) fixes the issue for now.

The actual error is in here https://github.com/owncloud/brute_force_protection/blob/v1.1.0/lib/Throttle.php#L103-L116

as found by our great @pako81

karakayasemi:
## Description
ShareManager was checking password of already authenticated public links.
This situation has been led to wrong "share.failedpasswordcheck" events emitting in already authenticated links and led to the related issue on brute force protection app. This problem has been resolved by first checking link already authenticated.

@micbar I would like to add this PR to 10.6 if it is possible.

## Related Issue
- Fixes https://github.com/owncloud/brute_force_protection/issues/138

## How Has This Been Tested?
- https://github.com/owncloud/brute_force_protection/issues/138 is resolved after the fix.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/PublicAuth.php
+++ b/apps/dav/lib/Connector/PublicAuth.php
@@ -95,10 +95,10 @@ class PublicAuth extends AbstractBasic {
 		// check if the share is password protected
 		if ($share->getPassword() !== null) {
 			if ($share->getShareType() === \OCP\Share::SHARE_TYPE_LINK) {
-				if ($this->shareManager->checkPassword($share, $password)) {
-					return true;
-				} elseif ($this->session->exists('public_link_authenticated')
+				if ($this->session->exists('public_link_authenticated')
 					&& $this->session->get('public_link_authenticated') === (string)$share->getId()) {
+					return true;
+				} elseif ($this->shareManager->checkPassword($share, $password)) {
 					return true;
 				} else {
 					if (\in_array('XMLHttpRequest', \explode(',', $this->request->getHeader('X-Requested-With')))) {
--- a/changelog/unreleased/38016
+++ b/changelog/unreleased/38016
@@ -1,0 +1,8 @@
+Bugfix: Do not emit "share.failedpasswordcheck" events for authenticated links
+
+ShareManager was checking password of already authenticated public links.
+This situation led to wrong "share.failedpasswordcheck" events emitting in already authenticated links.
+This problem has been resolved by first checking link already authenticated.
+
+https://github.com/owncloud/brute_force_protection/issues/138
+https://github.com/owncloud/core/pull/38016


phil-davis:
I restarted drone - in the previous run there were lots of pipelines that had 10-minute timeouts waiting for stuff.

mrow4a (changes requested):
Before this PR we always checked for password first (even if public_link_authenticated is set), or was that just a bug? . Do you think this can have some security implications? Could you comment on this in the code (if relevant)?

jvillafanez (approved):

jvillafanez:
It might be a better idea to let the `shareManager->checkPassword()` function handle the "public_link_authenticated" session variable.
Basically, you don't do anything with the "public_link_authenticated" variable outside the function, but call always to the `checkPassword` function. It will be the `checkPassword` function the one that checks (and sets) the session variable. It could also provide a better control of when the "share.failedpasswordcheck" event is triggered.

phil-davis:
I restarted ddrone CI  - there were some silly startup fails of PHP unit test pipelines.

micbar:
@karakayasemi Can we merge this soon to get into 10.6?

phil-davis:
@micbar - @mrow4a gave a "change requested" review comment. That needs to be cleared (assuming the comment has been addressed...)

karakayasemi:
> Before this PR we always checked for password first (even if public_link_authenticated is set), or was that just a bug? . Do you think this can have some security implications? Could you comment on this in the code (if relevant)?

@mrow4a if `checkPassword` fails, the code looks `public_link_authenticated` anyway. So, in my opinion there is no risk in terms of security.

> It might be a better idea to let the shareManager->checkPassword() function handle the "public_link_authenticated" session variable.

@jvillafanez I thought same first but there are other usages of `checkPassword` method, since it will change the inside logic of `checkPassword`. So, I did not want to touch `checkPassword` method to keep the change minimal.

> @karakayasemi Can we merge this soon to get into 10.6?

@micbar IMHO, PR is ready to merge after CI problems resolved. There were some unrelated CI fails, I hope this time we will see green CI.

phil-davis:
fsweb.test.owncloud.com seems to be down (or not working nicely). All core CI pipelines that test against it are failing, in all PRs currently running.

mrow4a (approved):
@karakayasemi @jvillafanez I did some research, and indeed this change is bugfix. Also, the public link code is spaghetti a little that begs for refactoring https://github.com/owncloud/core/blob/v10.5.0/apps/files_sharing/lib/Controllers/ShareController.php . Also used in public apps https://github.com/owncloud/richdocuments/blob/740bd280caf35c7e8b6e8e96f05320acd9e689b1/lib/Storage.php#L136-L141 . Considering how it is currently implemented, legit solution.

phil-davis:
Rebased to get fresh CI.

PR #38017, opened 2020-10-19
[tests-only] Remove PHP unit test pipelines against fsweb.test.owncloud.com

phil-davis:
## Description
Remove the drone test pipelines that run the PHP unit tests with fsweb.test.owncloud.com as the storage.
There are already similar pipelines that run the PHP unit tests with `samba` storage in a docker container.

This removes the reliance on that "fsweb" server being up at all times for CI.

A question to consider is, how well does the behaviour of the `samba` docker match "real windows"? i.e. will running on `samba` find any issues that happen on "real windows".

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


DeepDiver1975:
Windows smb and samba have different behavior. This is why we test with samba and windows

phil-davis:
fsweb is working fine again now. I guess we will not be removing these test pipelines now.

PR #38018, opened 2020-10-19
Use random_bytes instead of self-rolled random function

### LINKED ISSUE #214: SecureRandom::generate should use bytes instead of ASCII characters
**Reported by @lynn-stephenson**

[`OC\Security\SecureRandom::generate`](https://github.com/owncloud/core/blob/master/lib/private/Security/SecureRandom.php#L72) will by default use `a-Z0-9+/` as the set of characters, which has 64 possibilities.

Most notably [`OC\Security\Crypto::encrypt`](https://github.com/owncloud/core/blob/master/lib/private/Security/Crypto.php#L87) will use a generated [IV](https://github.com/owncloud/core/blob/master/lib/private/Security/Crypto.php#L93) with a length of 16 bytes, with the default character set provided by the generation function.

If you recall there are 64 possible bytes with the default character set. This IV is constrained to the 64 possibilities, such that the set of elements is 64^16. There are *256 possible bytes* (2^8 = 256) in actuality. 64^16 = 256^12. This means this randomly generated IV is equivalent to a 12-byte/96-bit IV, not the expected 16-byte/128-bit IV.

I strongly recommend against using [`OC\Security\SecureRandom::generate`](https://github.com/owncloud/core/blob/master/lib/private/Security/SecureRandom.php#L72) in any parts of the project, at least with it's current design.

## Solution

Use *raw bytes* (not UTF8 bytes, multi-bytes, or weird internal byte look ups) when working in cryptography related areas. Use PHP's built-in [`random_bytes`](https://www.php.net/manual/en/function.random-bytes.php) function to generate cryptographic keys and IVs.

## Impact

This exponentially decreases the *expected* strengths of cryptographic keys and IVs alike. This is applicable *everywhere* this function is used, and could have unintended side effects as shown above.

Note: It impacts the supported version of OwnCloud.

IljaN:
## Description
Replace Secure::generate with random_bytes to strengthen the IV.
This creates a new format for the ecrypted-string with a v2| prefix. The modified encryption
code encrypts the session on disk

## Related Issue
- Fixes https://github.com/owncloud/encryption/issues/214

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Security/Crypto.php
+++ b/lib/private/Security/Crypto.php
@@ -90,13 +90,13 @@ class Crypto implements ICrypto {
 		}
 		$this->cipher->setPassword($password);
 
-		$iv = $this->random->generate($this->ivLength);
+		$iv = \random_bytes($this->ivLength);
 		$this->cipher->setIV($iv);
 
 		$ciphertext = \bin2hex($this->cipher->encrypt($plaintext));
 		$hmac = \bin2hex($this->calculateHMAC($ciphertext.$iv, $password));
 
-		return $ciphertext.'|'.$iv.'|'.$hmac;
+		return 'v2|' . $ciphertext.'|'. \bin2hex($iv).'|'.$hmac;
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


IljaN:
CI Fail. The issue is that the random values are outside of the UTF8 range and thus can't be written to the database. This is probably the reason why the weak "random-text" method was used to generate the IV. Or maybe the code predates random_bytes.

The correct implementation would be to use base64Encode before writing the cipher-material to disk or in to the database.

C0rby:
How about this?
```php
$iv = \random_bytes($this->ivLength);
$this->cipher->setIV($iv);
$iv = \bin2hex($iv);
...
return $ciphertext.'|'.$iv.'|'.$hmac;
```

And when one needs the iv again it's just:
`$iv = \hex2bin($ivhex);`

lynn-stephenson:
@C0rby That's the correct way of doing it. You can encode the output to ASCII hexadecimal before storing the data, or base64ing it. Your choice.

C0rby (approved):
Looks good to me! :+1:

DeepDiver1975 (changes requested):
Please add unit tests

DeepDiver1975 (changes requested):
V1 test ?

IljaN:
@DeepDiver1975 V1 is already tested here: https://github.com/owncloud/core/blob/816bc5b9ee901bcfa1003a283fffa6fa936024e2/tests/lib/Security/CryptoTest.php#L49

DeepDiver1975 (approved):

Issue #37891, opened 2020-09-07
Refactor oC10 bug demonstration scenarios


phil-davis:
We currently have acceptance test scenarios that demonstrate existing bugs in oC10. We do not want other implementations to try to make those scenarios pass. It will be much better if other implementations make the correct behavior happen, and the bugs are also fixed in oC10. To do that:

- find the acceptance test scenarios that demonstrate oC10 bugs, they should be labelled issue-nnnnn
- write a new scenario that demonstrates the correct expected behavior. Tag that scenario `skipOnOcV10` because it will not pass on oC10. Tag it `issue-nnnnn` so that we can know which issue it relates to in future.
- move the bug-demonstration scenario into its own feature file - e.g. `moveReceivedShareOc10Issue30325.feature`
- tag that scenario `notToImplementOnOCIS` because it only demonstrates an oC10 bug. We don't want it on OCIS.
- tag that scenario `issue-nnnn` so it can be found easily.

e.g. `moveReceivedShare.feature` has:
```
  @issue-30325
  Scenario: receiver tries to rename a received share with share, read permissions
    Given user "Alice" has created folder "folderToShare"
    And user "Alice" has uploaded file with content "thisIsAFileInsideTheSharedFolder" to "/folderToShare/fileInside"
    And user "Alice" has shared folder "folderToShare" with user "Brian" with permissions "share,read"
    When user "Brian" moves folder "folderToShare" to "myFolder" using the WebDAV API
    Then the HTTP status code should be "403"
#    Then the HTTP status code should be "201"
    And as "Brian" folder "myFolder" should not exist
#    And as "Brian" folder "myFolder" should exist
    But as "Alice" folder "myFolder" should not exist
#    When user "Brian" moves file "/myFolder/fileInside" to "/myFolder/renamedFile" using the WebDAV API
    When user "Brian" moves file "/folderToShare/fileInside" to "/folderToShare/renamedFile" using the WebDAV API
    Then the HTTP status code should be "403"
    And as "Brian" file "/folderToShare/renamedFile" should not exist
    But as "Brian" file "/folderToShare/fileInside" should exist
```

The good scenario might be:
```
  @skipOnOcV10 @issue-30325
  Scenario: receiver tries to rename a received share with share, read permissions
    Given user "Alice" has created folder "folderToShare"
    And user "Alice" has uploaded file with content "thisIsAFileInsideTheSharedFolder" to "/folderToShare/fileInside"
[...]

PR #38025, opened 2020-10-21
[Tests-Only] Oc10 bug demonstration scenarios added separately for webUI tests

jasson99:
## Description
Refactored some webUI test scenarios demonstrating the actual behaviors.

Scenarios demonstrating bugs in Oc10 are now in new feature files and tagged `notToImplementOnOCIS`  because they only demonstrate an oC10 bug that needs to be fixed. These feature files have been named with respect to the issue that each one demonstrates.

The main scenarios now demonstrate the expected behavior, which we want to happen on both Oc10 and OCIS. They are tagged `skipOnOcV10` for now, because they would currently fail on Oc10.

## Related Issue
- Part of https://github.com/owncloud/core/issues/37891

## How Has This Been Tested?
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/webUIRenameFolders/renameFoldersOc10Issue30325.feature
+++ b/tests/acceptance/features/webUIRenameFolders/renameFoldersOc10Issue30325.feature
@@ -1,0 +1,22 @@
+@webUI @insulated @disablePreviews
+Feature: rename folders
+  As a user
+  I want to rename folders
+  So that I can organise my data structure
+
+  Background:
+    Given user "Alice" has been created with default attributes and without skeleton files
+
+  @issue-30325 @notToImplementOnOCIS
+  Scenario: Rename a folder which is received as a share (without change permission)
+    Given user "Brian" has been created with default attributes and without skeleton files
+    And user "Brian" has created folder "RandomFolder"
+    And user "Brian" has uploaded file with content "thisIsFileInsideFolder" to "/RandomFolder/newFile"
+    And user "Brian" has shared folder "RandomFolder" with user "Alice" with permissions "read, share"
+    And user "Alice" has logged in using the webUI
+    When the user renames folder "RandomFolder" to "renamedFolder" using the webUI
+#   Then folder "renamedFolder" should be listed on the webUI
+    Then a notification should be displayed on the webUI with the text 'Could not rename "RandomFolder"'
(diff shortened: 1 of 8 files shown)


tests/acceptance/features/webUIRenameFolders/renameFoldersOc10Issue30325.feature
```diff
@@ -0,0 +1,22 @@
+@webUI @insulated @disablePreviews
+Feature: rename folders
+  As a user
+  I want to rename folders
+  So that I can organise my data structure
+
+  Background:
+    Given user "Alice" has been created with default attributes and without skeleton files
```
Talank, line 8: You do not need the background when there is only one scenario in the feature file.

tests/acceptance/features/webUISharingExternal2/acceptDeclineFederatedSharesOc10Issue34742.feature
```diff
@@ -0,0 +1,32 @@
+@webUI @federation-app-required @insulated @disablePreviews @files_sharing-app-required
+Feature: Federation Sharing - sharing with users on other cloud storages
+  As a user
+  I want to share files with any users on other cloud storages
+  So that other users have access to these files
+
+  Background:
```
Talank, line 7: Same here. You do not need background.

Talank (dismissed):

tests/acceptance/features/webUISharingExternal2/acceptDeclineFederatedSharesOc10Issue34742.feature
phil-davis: Keeping the background, like in the original feature file, allows the scenario to also be "the same" (similar) to the "good" one in the original feature file. That will make it easier for developers to see what is the real difference between the bug-demo scenario and the "good: scenario.

phil-davis (approved):

tests/acceptance/features/webUIRenameFolders/renameFoldersOc10Issue30325.feature
phil-davis: See comment below.

PR #38054, opened 2020-10-29
Add the getSharesByPath share method to the interface

jvillafanez:
This method was being used in the files_sharing app even though it was
an implementation detail. The IShareProvider interface requires that
method to be implemented in the providers already, and the current share
manager already has the method implemented. This means that this will
just expose that method through the interface
In addition, the pagination wasn't being used and it wasn't implemented
in the share providers, so it has been removed

## Description
Make the `getSharesByPath` method available through the interface

## Related Issue
No opened issue.
`getSharesByPath` is used in the files_sharing app, as shown in this PR, and will be used in https://github.com/owncloud/windows_network_drive/pull/310

## Motivation and Context
Using methods not exposed by the interface while using the interface (and not the actual implementation) is a bad practice.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/SharedMount.php
+++ b/apps/files_sharing/lib/SharedMount.php
@@ -202,7 +202,6 @@ class SharedMount extends MountPoint implements MoveableMount {
 		$shareManager = \OC::$server->getShareManager();
 		// FIXME: make it stop earlier in '/$userId/files'
 		while ($targetNode !== null && $targetNode->getPath() !== '/') {
-			'@phan-var \OC\Share20\Manager $shareManager';
 			$shares = $shareManager->getSharesByPath($targetNode);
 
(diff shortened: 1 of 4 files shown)


karakayasemi (approved):

karakayasemi (changes requested):
The change make sense to me. I double checked usage of deleted parameters. Just needs changelog entry before merge.

phil-davis:
@jvillafanez needs a rebase due to a composer major version 2.0 change that happened in CI yesterday.

karakayasemi (approved):

PR #38055, opened 2020-10-29
avoid retrieving user root iteratively in share controller

mrow4a:
Fixes performance problem caused by code that was not refactored for long time.

Prerequisite:
- Create 100 user shares from test1 to test2 root folder
- Listing folders mounted for user test2-> 2s (complexity O(m), m->complexity of propfind, m is small overhead)

Attempt to reduce complexity but reusing available properties from "propfind kind of operation":

- Opening shared with you panel before fix -> 15s complexity O(m+n), n>>m -> O(n)
- Opening shared with you panel after attempt to fix, but due to need to return properties of mount like storage, storage_id it is not possible -> 2s complexity O(m)
- Opening shared with you panel after a fix that just reduced impact of n -> 7s complexity O(m+n), n>m -> O(n)

Finally, I settled on being required on reducing to 7s from 15s, but still not perfect as we need to respond with many mount properties for /shares  .. paging would solve the problem best.

Zrzut ekranu 2020-10-29 o 21 41 11

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -86,6 +86,9 @@ class Share20OcsController extends OCSController {
 	 */
 	private $additionalInfoField;
 
+	/** @var Folder[] */
+	private $currentUserFolder;
+
 	public function __construct(
 		$appName,
 		IRequest $request,
@@ -133,6 +136,21 @@ class Share20OcsController extends OCSController {
 	}
 
 	/**
+	 * Returns root folder of the current user
+	 *
+	 * @return Folder
+	 */
+	private function getCurrentUserFolder() {
+		// cache only one key, but be sure to check current user session id in case
+		// current user folder changes
+		$userSessionId = $this->userSession->getUser()->getUID();
+		if (!isset($this->currentUserFolder[$userSessionId])) {
+			$this->currentUserFolder = [$userSessionId => $this->rootFolder->getUserFolder($userSessionId)];
+		}
+		return $this->currentUserFolder[$userSessionId];
+	}
+
+	/**
 	 * Convert an IShare to an array for OCS output
 	 *
 	 * @param IShare $share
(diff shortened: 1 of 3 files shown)


phil-davis:
"something happened" - there are unit test fails, and acceptance test fails, for lots of sharing scenarios.

mrow4a:
@phil-davis seems pending shares broke, I suspected this might happen (thus draft). Unit tests failures are expected due to logic change.

This PR is legacy code fight, so might need strong QA after.

phil-davis:
> This PR is legacy code fight, so might need strong QA after.

CI has a good range of sharing and resharing test scenarios. So if you get CI green, it will be 99% good-to-go - we will just need to look at the code changes and imagine if there are weird corner-cases to double-check.

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -172,21 +186,19 @@ protected function formatShare(IShare $share, $received = false) {
...
 				$userFolder = $this->rootFolder->getUserFolder($share->getShareOwner());
 			}
 		} else {
-			$userFolder = $this->rootFolder->getUserFolder($this->userSession->getUser()->getUID());
+			$userFolder = $this->getCurrentUserFolder();
 		}
 
-		$nodes = $userFolder->getById($share->getNodeId(), true);
-		$node = $nodes[0] ?? null;
+		$node = $share->getNode();
```
mrow4a, line 198: this is the change that broke current acceptance shares for pending shares. I think I know what happens, and will add extensive comments

mrow4a:
@jvillafanez can you have a look? I tried to equalize the performance to PROPFIND, but it is not possible as OCS requires to return a set of fields that makes it not possible.

Proper solution would be paging..

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -132,6 +135,18 @@ private function getAdditionalUserInfo(IUser $user) {
 		return null;
 	}
 
+	/**
+	 * Returns root folder of the current user
+	 *
+	 * @return Folder
+	 */
+	private function getCurrentUserFolder() {
+		if ($this->currentUserFolder === null) {
```
jvillafanez, line 144:
```
if (!isset($this->currentUserFolder[$userSessionId])) {
  $this->currentUserFolder = [$userSessionId => $this->rootFolder->getUserFolder(....)];  // cache only one item
}
return $this->currentUserFolder[$userSessionId];
```
It might be better solution because it would allow us to change and update the user folder if the user in the session changes. I'm thinking about the impersonation app, although I don't know if it could give some trouble because it shouldn't switch users in the middle of some operation.

mrow4a: I think this is very good idea anyways, just to be on a save side

jvillafanez: @mrow4a can you take care or should I take over?

mrow4a: I will take care.

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -172,34 +190,33 @@ protected function formatShare(IShare $share, $received = false) {
...
-			$userFolder = $this->rootFolder->getUserFolder($this->userSession->getUser()->getUID());
+			$userFolder = $this->getCurrentUserFolder();
 		}
 
-		$nodes = $userFolder->getById($share->getNodeId(), true);
-		$node = $nodes[0] ?? null;
-		if ($node === null) {
+		// we need to retrieve a share mountpoint node for the userfolder,
+		// we cannot use share->getNode() as it retrieves the original node and
+		// not a user folder mount. Share node will be needed to retrive e.g.
```
phil-davis, line 204:
```suggestion
		// not a user folder mount. Share node will be needed to retrieve e.g.
```

changelog/unreleased/38055
```diff
@@ -0,0 +1,8 @@
+Bugfix: Avoid retrieving user root iteratively in share controller
+
+There was performance problem that with many shares, "share tab" was slow
```
phil-davis, line 3:
```suggestion
There was a performance problem that with many shares, the "share tab" was slow
```

jvillafanez (approved):
Code looks fine. Need to fix typos

mrow4a:
please squash the PR when merging

phil-davis (approved):

phil-davis:
I will squash now...

PR #38057, opened 2020-10-30
Harden against key-relation

### LINKED ISSUE #215: Use PHP's built-in hash_hkdf function
**Reported by @lynn-stephenson**

[`OC\Security\Crypto::calculateHMAC`](https://github.com/owncloud/core/blob/master/lib/private/Security/Crypto.php#L68) computes the MAC key like so:

```php
$password = \hash('sha512', $password . 'a');
```

This is vulnerable to [related key attacks](https://en.wikipedia.org/wiki/Related-key_attack). The relationship between the encryption key and the MAC key is not well isolated. Note that if the MAC key is obtained, it is not feasible to obtain the encryption key with the MAC key due to the one-way function.

## Solution

Use PHP's built-in [`hash_hkdf`](https://www.php.net/manual/en/function.hash-hkdf) to split a single key into multiple keys. I highly recommend you use SHA512 with this function. Obtain the first 256 bits and use that as the encryption key, and use the last 256 bits for the MAC key. This complicates the relationship between the encryption and MAC key providing both a different key and prevent them from being easily relatable to each other.

## Impact

Should an adversary obtain the encryption key (which is not entirely impractical these days), by any means it is trivial to also compute the MAC key, resulting in loss of not only confidentially, but also integrity, and authenticity.

--- 
For further clarification, when I'm referring to a *single key*, I talking about OwnCloud's `$this->config->getSystemValue('secret')`.

IljaN:
## Related Issue

- Fixes https://github.com/owncloud/encryption/issues/215 (Use PHP's built-in hash_hkdf function)

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Security/Crypto.php
+++ b/lib/private/Security/Crypto.php
@@ -88,13 +88,19 @@ class Crypto implements ICrypto {
 		if ($password === '') {
 			$password = $this->config->getSystemValue('secret');
 		}
+
+		// Split single key in to multiple keys, use one for mac and one for key
+		// to harden against related-key attacks
+		// https://github.com/owncloud/encryption/issues/215
+		$derived = \hash_hkdf('sha512', $password, 0);
+		list($password, $hmacKey) = \str_split($derived, 32);
 		$this->cipher->setPassword($password);
 
 		$iv = \random_bytes($this->ivLength);
 		$this->cipher->setIV($iv);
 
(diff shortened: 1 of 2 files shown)


C0rby (approved):
Looks good! :+1:

PR #38059, opened 2020-10-30
[Tests-Only] delete users using provisioning api

C0rby:
Implemented a (probably dirty) way to delete the users after a scenario has run. IMO that is required to make the tests in https://github.com/owncloud/ocis/pull/755 pass cause I haven't found any place were the provisioning api was used to delete the user.

### DIFF
--- a/tests/TestHelpers/OcisHelper.php
+++ b/tests/TestHelpers/OcisHelper.php
@@ -78,18 +78,16 @@ class OcisHelper {
 	 *
 	 * @return void
 	 */
-	public static function deleteRevaUserData($user = "") {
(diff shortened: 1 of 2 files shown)


phil-davis:
https://github.com/owncloud/core/blob/master/tests/acceptance/features/bootstrap/Provisioning.php `cleanupDatabaseUsers()` is called from `afterScenario()` - that runs at the end of every scenario.

For each user, that calls `deleteUser` which calls `deleteTheUserUsingTheProvisioningApi`

I expect that that will be called when testing on OCIS.

C0rby:
Ah yes you are right... Then there is no need for this change.
But I still need to figure out why the delete on the provisioning API wasn't called when I ran the tests locally...

PR #38062, opened 2020-11-01
Add Oracle to web wizard

mrow4a:
When adding Oracle to available dbs in owncloud, it was overlooked to add it also to web wizard (at least from what I inspected in Blame).

This PR adds support for Oracle in web wizard

Steps: 
- setup docker-compose for oracle, good example is https://github.com/owncloud/core/blob/master/.drone.star#L1687-L1696, but needs to be applied to docker-compose
- install Oracle driver on OC server, good example is https://github.com/owncloud-ci/php/tree/master/latest/overlay-amd64/etc/php/7.2
- check oracle installed (php -m | grep oci)
- install owncloud with oracle
Zrzut ekranu 2020-11-1 o 20 35 07
Zrzut ekranu 2020-11-1 o 20 35 35
Zrzut ekranu 2020-11-1 o 20 35 54

### DIFF
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -150,7 +150,7 @@ class Setup {
 			$configuredDatabases = \array_keys($availableDatabases);
 		} else {
 			$configuredDatabases = $this->config->getSystemValue('supportedDatabases',
-				['sqlite', 'mysql', 'pgsql']);
+				['sqlite', 'mysql', 'pgsql', 'oci']);
 		}
 		if (!\is_array($configuredDatabases)) {
 			throw new Exception('Supported databases are not properly configured.');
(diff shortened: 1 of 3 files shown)


lib/private/Setup.php
```diff
@@ -150,7 +150,7 @@ public function getSupportedDatabases($allowAllDatabases = false) {
 			$configuredDatabases = \array_keys($availableDatabases);
 		} else {
 			$configuredDatabases = $this->config->getSystemValue('supportedDatabases',
```
DeepDiver1975, line 152: We did add this config switch to enable oracle ..... see https://github.com/owncloud/enterprise/issues/4018#issuecomment-720153383

DeepDiver1975 (changes requested):

micbar:
Decision: works as designed, closing.
Thx @DeepDiver1975

PR #38077, opened 2020-11-04
[tests-only] Adjust logic for testing on OCIS or REVA

phil-davis:
## Description
Before this PR, when testing on REVA you have to define bothe environment variables `TEST_OCIS` and `TEST_REVA` - that is not such an obvious thing.

1) adjust the test logic so that if just `TEST_REVA` is defined, then all the appropriate test setup and teardown happens.

2) change the environment variable `TEST_EXTERNAL_USER_BACKENDS` to `TEST_WITH_LDAP` to better reflect what it does. The test suite uses this environment variable to know to create/delete users and groups on LDAP... (rather than with the provisioning API)

Note: `TEST_OCIS` is only documented in `owncloud/ocis` repo. And the use of that does not change. 
`TEST_REVA` and `TEST_EXTERNAL_USER_BACKENDS` are only documented in `cs3org/reva` repo. That will be changed in the PR that bumps the core commit.
None of these are mentioned in `owncloud/docs` so there is nothing to change there! 

I will make a PR in `user_ldap` repo to change `TEST_EXTERNAL_USER_BACKENDS` after this PR is merged.

After this PR, we want to adjust the logic in a separate PR so that user data and shares are not cleaned up "under-the-hood" when running on OCIS. There is an OCIS PR in development that will cleanup user files, shares and other metadata correctly when the user is deleted using the provisioning API.

## How Has This Been Tested?
CI

PR https://github.com/cs3org/reva/pull/1295 demonstrates that this works.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1445,7 +1445,7 @@ def acceptance(ctx):
 									environment['MAILHOG_HOST'] = 'email'
 
 								if params['ldapNeeded']:
-									environment['TEST_EXTERNAL_USER_BACKENDS'] = True
+									environment['TEST_WITH_LDAP'] = True
 
 								if params['testingRemoteSystem']:
 									environment['TESTING_REMOTE_SYSTEM'] = True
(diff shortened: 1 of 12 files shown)


jasson99 (commented):
Looks good to me

jasson99 (approved):

PR #38086, opened 2020-11-06

phil-davis:


phil-davis:
Somehow drone gets https://drone.owncloud.com/owncloud/core/27499 "yaml: line 1: mapping values are not allowed in this context" when there is the 13-character string "postgres:10.3" and does not start.

Change it to "postgres:9.4" and drone happily processes it.

I have seen this kind of weird stuff before, but never managed to track down the problem.

phil-davis:
Out-of-date - see PR #38401

PR #38099, opened 2020-11-10
Coerce occ config:app:set boolish value to 0 or 1

### LINKED ISSUE #37697: config:app:set has no "type" parameter
`occ config:system:set` has `--type` parameter and you can set a setting to boolean true or false.

`occ config:app:set` has no `--type` parameter. Do something like:

```
php occ config:app:set files enable_lock_file_action --value true
```

Actually the setting is string "true". Code that gets the setting ends up understanding that it is boolean `true` because it is non-zero.

```
php occ config:app:set files enable_lock_file_action --value false
```

The setting is string "false". Code that gets the setting ends up understanding that it is boolean `true` because it is non-zero. But users get confused, it looks false but the functionality is still enabled.

```
php occ config:app:set files enable_lock_file_action --value 0
```
Setting to `0` really switched off the setting.

```
occ config:app:delete files enable_lock_file_action
```
Deleting the setting allows it to go back to its default of `false`.

It would be less confusing if we could do:
```
php occ config:app:set files enable_lock_file_action --type boolean --value false
```

If the backend database field is only storing a string, then `--type boolean --value false` could store "0" and `--type boolean --value true` could store "1".

AlexAndBear:
## Description
Setting an app value via the occ config:app:set command to false may not result to a negation evaluation due the value will be stored in the database as 'false' type string. 
As this is a non-zero value it may evaluate to true.
To come over this problem we will coerce the value: 
true => 1
false => 0

## Related Issue

- Fixes <https://github.com/owncloud/core/issues/37697>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/37697
+++ b/changelog/unreleased/37697
@@ -1,0 +1,10 @@
+Enhancement: Fix expiring a wrong share entry problem
+
+Setting an app value via the occ config:app:set command to false may not result to a negation evaluation due the value will be stored in the database as 'false' type string.
+As this is a non-zero value it may evaluate to true.
+To come over this problem we will coerce the value:
+true => 1
+false => 0
+
+https://github.com/owncloud/core/issues/37697
+https://github.com/owncloud/core/pull/38099
--- a/core/Command/Config/App/SetConfig.php
+++ b/core/Command/Config/App/SetConfig.php
@@ -80,10 +80,27 @@ class SetConfig extends Base {
 			return 1;
 		}
 
-		$configValue = $input->getOption('value');
+		$configValue = $this->coerceValue($input->getOption('value'));
 		$this->config->setAppValue($appName, $configName, $configValue);
 
 		$output->writeln('<info>Config value ' . $configName . ' for app ' . $appName . ' set to ' . $configValue . '</info>');
 		return 0;
+	}
+
+	/**
+	 * @param $value string
+	 *
+	 * @return int|string
+	 */
+	protected function coerceValue(string $value) {
+		if ($value === 'false') {
+			return 0;
+		}
+
+		if ($value === 'true') {
+			return 1;
+		}
+
+		return $value;
 	}
 }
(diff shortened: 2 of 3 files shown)


DeepDiver1975 (changes requested):
unit test? THX

HieronymusPerro:
> unit test? THX

Added a acceptance test instead

jvillafanez:
I don't think this is a good idea.

There are a couple of things to take into account:
* The DB will store any value as string because that's how the DB column is defined.
* The user will always input a string

I guess the main problem here is the lack of information. This cli command is part of core and it very agnostic about how the value will be used.
The key point is that it depends on the app how that value will be interpreted, and as such, it should be documented by the app using the value.

For example, an app could store "useFeature" with "true" or "false", so the app can tell you to use `occ config:app:set app useFeature --value true`. Now, the app could have something like:
```
$useFeature = getFromDB('useFeature');
if ($useFeature === 'true') {
.....
}
```
so yeah, the app is explicitly checking for the "true" string because it knows that the function will return a string, and it expect it to be "true" the same it could expect it to be "theAppMustUseThis" string.
The app could also use the `filter_var` function for a more flexible validation so "1" or "on" could also be accepted as valid, but again, it depends on the app.

My main point is that the interpretation of the value depend on each specific app using the value, and core shouldn't interfere.

Note that this is different than the case with `occ config:system:set` where it will be translated to actual php code, which requires explicit typing.

jvillafanez:
I agree that having "true" / "false" stored and used as strings is confusing and should be avoided, but it's probably a better idea to promote the usage of "yes" / "no" instead and check the value with `filter_var`.

HieronymusPerro:
I mostly agree @micbar must decide

mmattel:
@JanAckermann any news on that?

JanAckermann:
Not yet @mmattel, I need to speak about this with @micbar 

More thoughts on this are welcome as well

changelog/unreleased/37697
```diff
@@ -0,0 +1,10 @@
+Enhancement: Fix expiring a wrong share entry problem
```
karakayasemi, line 1: This looks like a copy-paste mistake.

JanAckermann:
@JammingBen time to say good bye to our real first PR :cry:

PR #38104, opened 2020-11-11
when adding public link to owncloud, check for ssl for both with and without redirects

mrow4a:
This solves the issue, that when adding public link to owncloud, when checking SSL certificate and servers does not follow redirects, we should also try with "allow_redirects=>false` so redirects are not allowed.

### DIFF
--- a/apps/files_sharing/ajax/external.php
+++ b/apps/files_sharing/ajax/external.php
@@ -73,14 +73,51 @@ $externalManager = new \OCA\Files_Sharing\External\Manager(
 
 // check for ssl cert
 if (\substr($remote, 0, 5) === 'https') {
+	// check allowing redirects
 	try {
 		\OC::$server->getHTTPClientService()->newClient()->get($remote, [
 			'timeout' => 10,
 			'connect_timeout' => 10,
+			'allow_redirects' => [
+				'max'             => 10,
+				'strict'          => false,
+				'referer'         => false,
+				'protocols'       => ['https'],
+				'track_redirects' => false
+			]
 		])->getBody();
+		$sslCertOk = true;
+	} catch (\GuzzleHttp\Exception\TooManyRedirectsException $e) {
+		// we got too many redirects as we allowed redirects but server cannot handle it
+		$sslCertOk = false;
 	} catch (\Exception $e) {
+		\OCP\Util::writeLog(
+			'files_sharing',
+			'Check for SSL Certificate failed: ' . \get_class($e) . ': ' . $e->getMessage(),
+			\OCP\Util::DEBUG
+		);
 		\OCP\JSON::error(['data' => ['message' => $l->t('Invalid or untrusted SSL certificate')]]);
 		exit;
+	}
+
+	// ssl cert check with redirects did not succeed, try disallowing redirects
+	if (!$sslCertOk) {
+		try {
+			\OC::$server->getHTTPClientService()->newClient()->get($remote, [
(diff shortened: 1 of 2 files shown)


changelog/unreleased/38104
```diff
@@ -0,0 +1,9 @@
+Bugfix: adding public link - check for ssl for both with and without redirects
+
+When adding public link to owncloud, checking of SSL certificate is performed.
+However when remote servers cannot handle redirects ToManyRedirects
```
phil-davis, line 4:
```suggestion
However when remote servers cannot handle redirects TooManyRedirects
```
Very minor change for next time you need to rebase or change something.
Having this correctly typed will help in future if people search for TooManyRedirects

apps/files_sharing/ajax/external.php
```diff
@@ -73,15 +73,52 @@
...
+				'protocols'       => ['https'],
+				'track_redirects' => false
+			]
 		])->getBody();
+		$sslCertOk = true;
+	} catch (\GuzzleHttp\Exception\TooManyRedirectsException $e) {
+		// we got too many redirects as we allowed redirects but server cannot handle it
+		$sslCertOk = false;
 	} catch (\Exception $e) {
+		\OCP\Util::writeLog(
```
mrow4a, line 94: @jvillafanez @HieronymusPerro please note here that if there is other issue then TooManyRedirectsException , I do not pass it here.

apps/files_sharing/ajax/external.php
```diff
@@ -73,15 +73,52 @@
...
+				'max'             => 10,
+				'strict'          => false,
+				'referer'         => false,
+				'protocols'       => ['https'],
+				'track_redirects' => false
+			]
 		])->getBody();
+		$sslCertOk = true;
+	} catch (\GuzzleHttp\Exception\TooManyRedirectsException $e) {
+		// we got too many redirects as we allowed redirects but server cannot handle it
```
mrow4a, line 91: @jvillafanez @HieronymusPerro I only allow retry when there is clear response from client with TooManyRedirectsException

jvillafanez:
I don't see the point of this change. Either you follow or don't follow.
If you get too many redirects accessing to my.server because it redirects you to my.server/a and then to my.server/b and so on, by disabling the redirections you'll get just to my.server which will return a (likely) 302 HTTP code, which will be fine to guzzle. By hitting the redirection limit, you already know that the all the previous servers are fine (or they redirect you fine), so there is no need to recheck again the first server.

Basically, I see the following options:
* Follow the redirections until limit. /a -> /b -> /c .... all of those are expected to return a 3xx status. If we get a "too many redirects" exception, we don't know what will be the status of the target url, which could be a 2xx, 4xx or 5xx
* Don't follow redirections. We'll get a 3xx code on the first link, which will be interpreted as ok by guzzle. Note that the last server might be broken and we'll assume it's ok.

The problem is that there is no guarantee that all the redirections will stay on the same host. www.my.server might redirect to my.server, which then could redirect to my.server.de. Each of these hosts might have different certificates.
This is why not following the redirects doesn't seem a good idea: the first server might be fine, but the last might have a wrong / expired certificate.

mrow4a:
@jvillafanez makes sense, started to agree that we need to force the redirect. But I think setting high limit for redirects does not make sense. We can probably set it to 25, but over this it really does not make sense.

davitol:
Superseeded by https://github.com/owncloud/core/pull/38107

PR #38111, opened 2020-11-13
[tests-only] Test with some unusual usernames

phil-davis:
Run CI with "unusual" user names and attributes.


phil-davis:
Passing, good. The needed changes are being merged in PR #38112

PR #38117, opened 2020-11-17
[tests-only] Add test for adding a public link shared file to your ownCloud

### LINKED ISSUE #38109: Tests for adding public link to your ownCloud
https://github.com/owncloud/core/blob/master/tests/acceptance/features/webUISharingPublic2/shareByPublicLink.feature has some scenarios that check that the "add to your ownCloud" button is displayed.

Create new scenarios that actually click the button, add the public link, and check that the user can see the file(s).

Probably this will need to be done in federated sharing test suites. The remote federated system can use the sharing API to create the public link. Then the scenario can browse to the public link with the webUI, click the "add"  button, and add to the local system. Then check that the files are available on the local system.

phil-davis:
## Description
There was already a test scenario that adds a public-link-shared folder to your ownCloud.
1) add to that scenario so that it checks the the folder really is read-only (cannot upload to it)
2) add a scenario for adding a public-link-shared fike to your ownCloud.

## Related Issue
- Fixes #38109 

## How Has This Been Tested?
Local run of the test scenarios.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -1395,6 +1395,26 @@ trait WebDav {
 	}
 
 	/**
+	 *
+	 * @param string $user
+	 * @param string $entry
+	 * @param string $path
+	 * @param string $type
+	 *
+	 * @return bool
+	 */
+	public function fileOrFolderExists(
+		$user, $entry, $path, $type = "files"
+	) {
+		try {
+			$this->asFileOrFolderShouldExist($user, $entry, $path, $type);
+			return true;
+		} catch (Exception $e) {
+			return false;
+		}
+	}
+
+	/**
 	 * @Then /^as "([^"]*)" exactly one of these (files|folders|entries) should exist$/
 	 *
 	 * @param string $user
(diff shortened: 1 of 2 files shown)


individual-it (approved):

phil-davis:
https://drone.owncloud.com/owncloud/core/27606/145/15
```
    But user "Brian" should not be able to upload file "filesForUpload/textfile.txt" to "/file-shared-by-public-link.txt"                         # FeatureContext::theUserShouldNotBeAbleToUploadFileTo()
      file '/file-shared-by-public-link.txt' should not exist. But it does exist and is a file
```
I thought that I had that passing locally yesterday. But I see that the step is a bit too fussy - Brian cannot upload (overwrite) that file, but the file does exist after the upload, because it already existed. I will look.

dpakach (commented):
LGTM :+1:  looks like coding standard has failed

phil-davis:
>  looks like coding standard has failed

I fixed that at the same moment that you reviewed.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -1940,8 +1960,23 @@ public function userShouldBeAbleToUploadFileTo($user, $source, $destination) {
 	 * @return void
 	 */
 	public function theUserShouldNotBeAbleToUploadFileTo($user, $source, $destination) {
+		$fileAlreadyExists = $this->fileOrFolderExists($user, "file", $destination);
+		if ($fileAlreadyExists) {
+			$this->downloadFileAsUserUsingPassword($user, $destination);
+			$initialContent = (string) $this->response->getBody();
+		}
 		$this->userUploadsAFileTo($user, $source, $destination);
```
individual-it, line 1968: what about checking the response code here?

individual-it (changes requested):

tests/acceptance/features/bootstrap/WebDav.php
phil-davis: yes, can do. But in any case, if the response code is not 2xx then we still need to check the file content - the server might return 4xx but still store the file.

If the response is 2xx (upload succeeded) then I guess we can fail the step - even if the server somehow did not store the file, then we would like to know that it looked like the file was uploaded OK.

phil-davis: I checked for 403 - that is the status that should happen for all attempted file uploads done by this step. 403 is returned both if it is a new file, or if it overwrites an existing file. This is a step that always provides the expected-to-be-correct password - so if some other 4xx is returned (e.g. 401) then actually we would like to know about that. So in that case the step will fail.

individual-it (approved):

Issue #38124, opened 2020-11-19
Use icon from theme when sharing a llink in public post


T0mWz:
When a public link is shared on a Social Media or other location, in some cases the ownCloud favicon-fb.png is shown.  This icon is located in core/img. But when a custom theme is used, then this icon should be taken from these theme instead of core. 

### Steps to reproduce
1. Create public link of a folder
2. Share that public link on a social media

### Expected behaviour
Your own icon should be displayed there.

### Actual behaviour
ownCloud favicon-fb icon from core is used

T0mWz:
Example; ![Screenshot 2020-11-19 at 17 06 08](https://user-images.githubusercontent.com/2477797/99691505-96927980-2a89-11eb-8a39-a7dc25cdf881.png)

T0mWz:
https://github.com/owncloud/core/compare/master...T0mWz:patch-4

T0mWz:
https://github.com/owncloud/core/pull/38125

T0mWz:
Oh, huh? Which version of ownCloud are you using?
Have cleared all my cache, but the social icon keeps coming from core instead of the theme. 
Have tested with 10.4.1, but will test 10.5 too. Will let you know. Thanks for helping!

micbar:
@T0mWz still neded?

T0mWz:
Have upgrade to 10.6.0 yesterday. But icon still not appears, really strange..

micbar:
maybe releated to https://github.com/owncloud/core/pull/38246

T0mWz:
Hi @micbar , @jvillafanez 
Sorry for my late reply. But with https://github.com/owncloud/core/pull/38246, my fix isn't needed anymore.
But had you already applied this patch to your environment, @jvillafanez ?

jvillafanez:
Probably I tested it with 10.5.0 version, which I don't think it has the bug. The bug mentioned in https://github.com/owncloud/core/pull/38246 was introduced after 10.5.0 and is probably present in 10.6.0, so the patch should fix that.

T0mWz:
Had the issue too with 10.5.0. But never mind, it's fixed with this patch.

PR #38125, opened 2020-11-19
Use Favicon FB icon from theme

T0mWz:
Use favicon-fb icon from theme instead of core location.


T0mWz:
See aso bug report; https://github.com/owncloud/core/issues/38124

PR #38135, opened 2020-11-23
Updating the .htaccess

cdamken:
Removing the php5 that is not needed anymore
Updating some preferences the helps for the pubic shares

## Related Issue
Some but I don't remember

## Motivation and Context
I just remember we needed this

## How Has This Been Tested?
After 7 years using oC I have seen this configuration works fine ;)

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -44,12 +44,10 @@
 </IfModule>
 
 <IfModule mod_php7.c>
-    php_value upload_max_filesize 16G
-    php_value post_max_size 16G
     php_value memory_limit 1G
     php_value mbstring.func_overload 0
     php_value default_charset 'UTF-8'
-    php_value output_buffering 4096
+    php_value output_buffering 0
     <IfModule mod_env.c>
       SetEnv htaccessWorking true
     </IfModule>
--- a/.htaccess
+++ b/.htaccess
@@ -43,26 +43,13 @@
   </FilesMatch>
 </IfModule>
 
-<IfModule mod_php5.c>
-    php_value always_populate_raw_post_data -1
-    php_value upload_max_filesize 513M
-    php_value post_max_size 513M
-    php_value memory_limit 512M
-    php_value mbstring.func_overload 0
-    php_value default_charset 'UTF-8'
-    php_value output_buffering 0
-    <IfModule mod_env.c>
-      SetEnv htaccessWorking true
-    </IfModule>
-</IfModule>
-
 <IfModule mod_php7.c>
-    php_value upload_max_filesize 513M
-    php_value post_max_size 513M
-    php_value memory_limit 512M
+    php_value upload_max_filesize 16G
+    php_value post_max_size 16G
+    php_value memory_limit 1G
     php_value mbstring.func_overload 0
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -43,26 +43,13 @@
...
-    php_value upload_max_filesize 513M
-    php_value post_max_size 513M
-    php_value memory_limit 512M
+    php_value upload_max_filesize 16G
+    php_value post_max_size 16G
+    php_value memory_limit 1G
     php_value mbstring.func_overload 0
     php_value default_charset 'UTF-8'
-    php_value output_buffering 0
+    php_value output_buffering 4096
```
DeepDiver1975, line 52: buffer should be 0 if I recall correctly ......

.htaccess
```diff
@@ -43,26 +43,13 @@
...
-      SetEnv htaccessWorking true
-    </IfModule>
-</IfModule>
-
 <IfModule mod_php7.c>
-    php_value upload_max_filesize 513M
-    php_value post_max_size 513M
-    php_value memory_limit 512M
+    php_value upload_max_filesize 16G
+    php_value post_max_size 16G
```
DeepDiver1975, line 48: not relevant anymore - we don't do any POST anymore to upload files ..... anyway

.htaccess
cdamken: I took it from:https://www.php.net/manual/en/outcontrol.configuration.php#ini.output-buffering
output_buffering bool/int
You can enable output buffering for all files by setting this directive to 'On'. If you wish to limit the size of the buffer to a certain size - you can use a maximum number of bytes instead of 'On', as a value for this directive (e.g., output_buffering=4096). This directive is always Off in PHP-CLI.

.htaccess
cdamken: I took it from: https://doc.owncloud.com/server/10.5/admin_manual/configuration/files/big_file_upload_configuration.html#configuring-your-web-server

cdamken:
@DeepDiver1975 Thanks for reviewing! ;)

.htaccess
DeepDiver1975: documentaiton always lies ;-)

.htaccess
DeepDiver1975: never the less - we should disable output buffering

.htaccess
mmattel: My thumbsup is for @DeepDiver1975 comment!

JanAckermann:
Suggestions accepted, please re-review ;)

mmattel:
I dont have apache but nginx, please let me know what results with `^/apps/gallery/s/(.*)$ /s/$1 [L,R=301]`

DeepDiver1975:
> I dont have apache but nginx, please let me know what results with `^/apps/gallery/s/(.*)$ /s/$1 [L,R=301]`

gallery app is dead since years - should not be in there

DeepDiver1975:
old aka dead -> close

PR #38136, opened 2020-11-23
Update Users.ini

cdamken:
This changes are needed too when you change the htaccess


phil-davis:
@cdamken is this still something that is needed or useful?

PR #38138, opened 2020-11-24
[Tests-Only] edit users only when running with oc10 backend

### LINKED ISSUE #38134: check if editing a created user in the API tests is needed when testing OCIS
after creating a user for the tests, the test-code also edits the user to set email & displayname.
https://github.com/owncloud/core/blob/b5b29f317727a81b48d6c25e7664532b8e94c3f3/tests/acceptance/features/bootstrap/Provisioning.php#L970

but when testing on ocis we set the email already in the first request https://github.com/owncloud/core/blob/b5b29f317727a81b48d6c25e7664532b8e94c3f3/tests/acceptance/features/bootstrap/Provisioning.php#L937
so if its possible also to set the displayname in the first request, we can skip the edit of the user when testing ocis

dpakach:
## Description

For acceptance tests when we create users, we send separate requests to set user's email and displayname after the user is created. This is because in oc10 we cannot create user with email, displayname set. but this is possible in ocis
So we can remove sending separate requests when running tests in ocis.
CI in ocis passes https://drone.owncloud.com/owncloud/ocis/1683

## Related Issue

- Fixes https://github.com/owncloud/core/issues/38134

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -924,6 +924,7 @@ trait Provisioning {
 			} else {
 				$attributesToCreateUser['userid'] = $userAttributes['userid'];
 				$attributesToCreateUser['password'] = $userAttributes['password'];
+				$attributesToCreateUser['displayname'] = $userAttributes['displayName'];
 				if (OcisHelper::isTestingOnOcisOrReva()) {
 					$attributesToCreateUser['username'] = $userAttributes['userid'];
 					if ($userAttributes['email'] === null) {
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -924,6 +924,7 @@ public function usersHaveBeenCreated(
 			} else {
 				$attributesToCreateUser['userid'] = $userAttributes['userid'];
 				$attributesToCreateUser['password'] = $userAttributes['password'];
+				$attributesToCreateUser['displayname'] = $userAttributes['displayName'];
```
phil-davis, line 927: Everything that calls `usersHaveBeenCreated` is passing in `$userAttributes['displayName']` - so it is always set, but possibly sometimes to `null`. I guess that is OK. The request to the "add user" endpoint might go out asking for `displayname` = `null` - probable it will be requesting the empty string, or guzzle will ignore that item in the body and not send it at all. On oC10 it does not matter - it ignores a `displayname` in the "add user" request. On OCIS we might find out what happens if the request explicitly asks for an empty `displayname`

phil-davis (approved):

PR #38145, opened 2020-11-26
[tests-only] reword test step to upload a file with a mtime

individual-it:
## Description
this step does not take any content, but only creates an empty file

## Motivation and Context
naming is hard, but we still should try to improve it

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -2306,7 +2306,7 @@ trait WebDav {
 	}
 
 	/**
-	 * @When user :user uploads file to :destination with mtime :mtime using the WebDAV API
+	 * @When user :user uploads a file to :destination with mtime :mtime using the WebDAV API
 	 * @Given user :user has uploaded a file to :destination with mtime :mtime using the WebDAV API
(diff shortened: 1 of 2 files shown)


individual-it:
maybe it would be better to change the step to actually upload a file

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -2306,7 +2306,7 @@ public function userUploadsAFileWithContentToUsingTus(
 	}
 
 	/**
-	 * @When user :user uploads file to :destination with mtime :mtime using the WebDAV API
+	 * @When user :user uploads a file to :destination with mtime :mtime using the WebDAV API
```
phil-davis, line 2309:
```
@When user :user uploads some data to file :destination with mtime :mtime using the WebDAV API
```
@individual-it what about words like this?

individual-it: there is no data at all, not even "some data" ;-) see #38146

PR #38148, opened 2020-11-27
[tests-only] tests for TUS uploads with mtime

individual-it:
## Description
test if the TUS uploads work correctly when the mtime is send in the `Upload-Metadata` header

## Related Issue
part of https://github.com/owncloud/product/issues/153

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavUploadTUS/uploadFileMtime.feature
+++ b/tests/acceptance/features/apiWebdavUploadTUS/uploadFileMtime.feature
@@ -1,0 +1,50 @@
+@api @skipOnOcV10
+Feature: upload file
+  As a user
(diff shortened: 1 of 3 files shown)


phil-davis (approved):
LGTM

Do the TUS upload requests return a useful HTTP status?

If so, then in the next PR of TUS tests we could also add a "Then the HTTP status should be" step after each "When uploads...using the TUS protocol" step. Because a client will look at that status and want it to be reliable.

individual-it:
@phil-davis its a multi stage process, we could check the last HTTP code, but the library we are using is hiding that from us :-(

PR #38151, opened 2020-11-27
Alternative SMB external storage implementation where all users of on…

DeepDiver1975:
…e share use the same storage id. User specific access is filtered in a cache wrapper by stating the files on SMB

## Description
- all users share the same storage id for the smb share (loggin via user credentials)
- fileids are constant over users
- fileid based concepts work (tagging, comments, direct link, file locking, wopi)

## How Has This Been Tested?
- :hand: 

### Setup of SMB version 2
Screenshot from 2020-11-30 09-54-55

### Same file id is used and as result comments and tags are show properly on both users:
Screenshot from 2020-11-24 12-12-41

### File ID is the same - see last query parameter in url
Screenshot from 2020-11-24 12-10-00

### Tow users with different permissions see folder content accordingly:
Screenshot from 2020-11-24 12-08-56

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/lib/Lib/Backend/SMB2.php
+++ b/apps/files_external/lib/Lib/Backend/SMB2.php
@@ -1,9 +1,8 @@
 <?php
 /**
- * @author Robin McCorkell <robin@mccorkell.me.uk>
- * @author Vincent Petry <pvince81@owncloud.com>
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
  *
- * @copyright Copyright (c) 2018, ownCloud GmbH
+ * @copyright Copyright (c) 2020, ownCloud GmbH
  * @license AGPL-3.0
  *
  * This code is free software: you can redistribute it and/or modify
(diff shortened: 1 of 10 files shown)


lib/public/Files/External/Auth/AuthMechanism.php
```diff
@@ -117,7 +117,8 @@ public function jsonSerialize() {
 	public function validateStorage(IStorageConfig $storage) {
 		// does the backend actually support this scheme
 		$supportedSchemes = $storage->getBackend()->getAuthSchemes();
-		if (!isset($supportedSchemes[$this->getScheme()])) {
+		if (!isset($supportedSchemes[$this->getScheme()]) &&
+			!isset($supportedSchemes[$this->getIdentifier()])) {
```
jvillafanez, line 121: Not sure if this is a good idea. Having an auth mechanism id instead of a scheme id will be confusing.
It might be better to keep using the scheme and allow username+password even though it doesn't make sense for this scenario. Note that WND will add new auth mechanism that might be usable in this context too.

apps/files_external/lib/Lib/Backend/SMB2.php
```diff
@@ -0,0 +1,65 @@
...
+use OCP\IUser;
+
+class SMB2 extends ExternalBackend {
+	use LegacyDependencyCheckPolyfill;
+
+	public function __construct(IL10N $l) {
+		$this
+			->setIdentifier('smb2')
+			->setStorageClass('\OCA\Files_External\Lib\Storage\SMB2')
+			->setText($l->t('SMB / CIFS (version 2)'))
```
jvillafanez, line 40: This might need some rewording:
* Is it for SMB2 protocol? Can I use this with a SMB1 connection?
* Does this new code deprecated the previous SMB mounts? (version 2 suggests improvements over the old SMB connection code)

As far as I know, previous SMB code is still valid and might perform better in some scenarios, so it seems this implementation can live side by side with the old one.

jvillafanez:
Note sure if we need to include a cache for the stat calls. Other than the code, the code looks fine

lib/public/Files/External/Auth/AuthMechanism.php
DeepDiver1975: I honestly want to limit this storage to sessioncredentials (and in wnd whatever will work there) anything else will just result in bullshit.
But yes - i am not too happy with this either - maybe we better introduce an explicit array of auth identifiers in addition ....

apps/files_external/lib/Lib/Backend/SMB2.php
DeepDiver1975: Yes - this is where product management is called to do some naming bingo ;-)

The 'old' implementation is still valid to be used ... no need fro deprication

DeepDiver1975:
> Note sure if we need to include a cache for the stat calls.

We will find out in some real life testing ....

hodyroff:
I would name this: "shared file ID" in German: "file ID Mehrfachnutzung" maybe @pmaier has a better idea ...

pmaier1:
Regarding wording, I actually like @DeepDiver1975's suggestion in another channel:
- SMB Personal (unique file IDs)
- SMB Collaborative (shared file IDs)

DeepDiver1975:
converting back to draft because of detected side effects while testing ....

jvillafanez (approved):

PR #38152, opened 2020-11-27
Allow removing shares which are invalid due to missing source fileid

jvillafanez:
## Related Issue
Related to https://github.com/owncloud/windows_network_drive/pull/310

## Motivation and Context
clean up data that won't be accessible

## How Has This Been Tested?
Checked along with https://github.com/owncloud/windows_network_drive/pull/310

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38152
+++ b/changelog/unreleased/38152
@@ -1,0 +1,7 @@
+Change: API changes to remove shares pointing to missing files
+
+If a file was completely deleted without unsharing first, the share
+would still exist in the DB even though it wouldn't be shown to the users.
+This change prepares a way to remove those shares.
+
+https://github.com/owncloud/core/pull/38152
--- a/apps/federatedfilesharing/lib/FederatedShareProvider.php
+++ b/apps/federatedfilesharing/lib/FederatedShareProvider.php
@@ -847,6 +847,44 @@ class FederatedShareProvider implements IShareProvider {
 	}
 
 	/**
+	 * @inheritDoc
+	 */
+	public function getSharesWithInvalidFileid(int $limit) {
+		// mimic the implementation for the default share provider in core
+		$validShareTypes = [
+			\OCP\Share::SHARE_TYPE_REMOTE,
+		];
+		// other share types aren't handled by this provider
+
+		$qb = $this->dbConnection->getQueryBuilder();
+
+		$qb = $qb->select('s.*')
+			->from('share', 's')
+			->leftJoin('s', 'filecache', 'f', $qb->expr()->eq('s.file_source', 'f.fileid'))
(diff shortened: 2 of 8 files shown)


jvillafanez:
Code-wise this is ready, but we might need to adjust the "since" tags. I'm assuming this will be for 10.7

mmattel:
@jvillafanez any progress on that?
Could you add a changelog pls?

jvillafanez:
Since this PR has API changes, I'm waiting for the official 10.7 alpha. Master is still with 10.6

changelog/unreleased/38152
```diff
@@ -0,0 +1,7 @@
+Change: API changes to remove shares pointing to missing files
+
+If a file was completely deleted without unsharing first, the share
+would still exist in the DB even though it wouldn't be shown to the users.
+This change opens a way to remove those shares.
```
mmattel, line 5: The text reads like a user/admin has the capability to do that. As I can´t see a occ command for doing so, this seems not to be the case. Therefore I propose to rewrite the text a bit to make it more clear how this is initiated.

jvillafanez: For now, the windows_network_drive app is the only app that will use this feature (in a future release), so I'd rather not mention it here.
Other than that, this feature isn't used at the moment. Including an occ command could be useful, but it needs to be discussed.

mmattel: Maybe: ‘This change opens‘ —> ‘This change prepares‘

jvillafanez:
This is currently frozen due to API changes. Waiting for OC 10.7 in order to resume the work in this area. There is a 10.6.1 version planned I think.
@micbar FYI

phil-davis:
> There is a 10.6.1 version planned I think.

Note: there is no 10.6.1 planned. Next version should be 10.7.0

VicDeo:
@jvillafanez looks good in general. 
Is it possible not to duplicate the code for the core federatedfilesharing app? As I see the only difference here is a set of share types.

jvillafanez:
> Is it possible not to duplicate the code for the core federatedfilesharing app? As I see the only difference here is a set of share types.

The problem is that the actual action depends on each provider. The manager just forwards the request to all the providers. In this context, we could expect that each provider has its own way of storing the information, and its own way to delete it. The fact that both providers use the same DB table could be considered as a bad coincidence.

I agree we should avoid that duplication, but I think we'd also need to refactor all the related operations and provide a common way for the providers to store the information. For now, as said, it seems the storing part is completely delegated to the providers.

VicDeo (approved):

AlexAndBear (approved):

PR #38155, opened 2020-11-29
[tests-only] Run acceptance tests with mysql:5.5

phil-davis:
Issue #38156


phil-davis:
The test scenarios that fail are all related to group, file or folder names that have emojis in them. mysql:5.5 is not handling those "extended" Unicode characters (at least not by default). The current officially supported databases are listed at https://doc.owncloud.org/server/10.5/admin_manual/installation/system_requirements.html - mySQL8+ is supported. So mySQL 5.5 is not officially supported anyway.

PR #38161, opened 2020-11-29
Allow subfolders for Google Drive mounts

mmattel:
## Description
This adds the ability to add a Google Drive mount from a GD subfolder.
Similar when you eg mount form SMB.

## Related Issue
No issue related.
Based on the work of: https://github.com/owncloud/files_external_gdrive/pull/9 (added support for changing root folder in google drive)

## Motivation and Context
One can now mount from a GD subfolder. This has big benefits:
- only a fraction of your GD is mounted and seen
- it is now possibe to use the `$user` variable to have a mount based on a subfolder structure like:
`sub_folder/$user` !
- when using encryption, only the subfolder gets encrypted

## How Has This Been Tested?
Manually having:
- subfolder = empty (like you have it without this PR)
- subfolder = name
- subfolder = name/$user

In case name or name/$user does not exist in GD, the mount does not show up without error.

image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

Adding a doc ticket when merged

### DIFF
--- a/apps/files_external/lib/Lib/Storage/Google.php
+++ b/apps/files_external/lib/Lib/Storage/Google.php
@@ -14,6 +14,7 @@
  * @author Robin McCorkell <robin@mccorkell.me.uk>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
  * @author Vincent Petry <pvince81@owncloud.com>
+ * @author Martin Mattel <github@diemattels.at>
  *
  * @copyright Copyright (c) 2017, ownCloud GmbH
  * @license AGPL-3.0
@@ -77,6 +78,7 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 			$this->service = new \Google_Service_Drive($this->client);
 			$token = \json_decode($params['token'], true);
 			$this->id = 'google::'.\substr($params['client_id'], 0, 30).$token['created'];
+			$this->root = isset($params['root']) ? $params['root'] : '';
 		} else {
 			throw new \Exception('Creating Google storage failed');
 		}
@@ -90,7 +92,11 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 	}
 
 	public function getId() {
-		return $this->id;
+		if ($this->root === '') {
+			return $this->id;
+		} else {
+			return "{$this->id}/{$this->root}";
+		}
 	}
 
 	private function getDefaultFieldsForFile() {
@@ -122,12 +128,37 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 	}
 
 	/**
-	 * Get the Google_Service_Drive_DriveFile object for the specified path.
+	 * Transform an external path to one originating from the virtual root.
+	 * @param $path the path relative to the virtual root directory
+	 * @return a path starting at the real root of Google Drive
+	 */
+	private function getAbsolutePath($path) {
+		if ($path === '.') {
+			$path = '';
+		}
+		$path = "{$this->root}/{$path}";
+		$path = \trim($path, '/');
+		return $path;
+	}
+
+	/**
+	 * Get elements of the relative path given.
+	 * The path is relative, in case a subfolder is used.
 	 * Returns false on failure.
 	 * @param string $path
 	 * @return \Google_Service_Drive_DriveFile|false
 	 */
 	private function getDriveFile($path) {
+		return $this->getInternalDriveFile($this->getAbsolutePath($path));
+	}
+
+	/**
+	 * Get the elements of the absolute path in case a subfolder is used.
+	 * Returns false on failure.
+	 * @param string $path
+	 * @return \Google_Service_Drive_DriveFile|false
+	 */
+	private function getInternalDriveFile($path) {
 		// Remove leading and trailing slashes
 		$path = \trim($path, '/');
 		if ($path === '.') {
@@ -144,7 +175,7 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 		} else {
 			// Google Drive SDK does not have methods for retrieving files by path
 			// Instead we must find the id of the parent folder of the file
-			$parentId = $this->getDriveFile('')->getId();
+			$parentId = $this->getInternalDriveFile('')->getId();
 			$folderNames = \explode('/', $path);
 			$path = '';
 			// Loop through each folder of this path to get to the file
@@ -179,7 +210,7 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 						$pos = \strrpos($path, '.');
 						if ($pos !== false) {
 							$pathWithoutExt = \substr($path, 0, $pos);
-							$file = $this->getDriveFile($pathWithoutExt);
+							$file = $this->getInternalDriveFile($pathWithoutExt);
 							if ($file && $this->isGoogleDocFile($file)) {
 								// Switch cached Google_Service_Drive_DriveFile to the correct index
 								unset($this->driveFiles[$pathWithoutExt]);
@@ -204,6 +235,15 @@ class Google extends \OCP\Files\Storage\StorageAdapter {
 	 * @param \Google_Service_Drive_DriveFile|false $file
 	 */
 	private function setDriveFile($path, $file) {
+		return $this->setDriveFileHelper($this->getAbsolutePath($path), $file);
+	}
+
+	/**
+	 * Set the Google_Service_Drive_DriveFile object in the cache
+	 * @param string $path
+	 * @param \Google_Service_Drive_DriveFile|false $file
+	 */
+	private function setDriveFileHelper($path, $file) {
 		$path = \trim($path, '/');
 		$this->driveFiles[$path] = $file;
 		if ($file === false) {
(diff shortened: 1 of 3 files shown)


apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -121,13 +123,37 @@ private function getDefaultFieldsForFolderScan() {
 		return $this->defaultFieldsForFolderScan;
 	}
 
+	/**
+	* Transform an external path to one originating from the virtual root.
+	* @param $path the path relative to the virtual root directory
+	* @return a path starting at the real root of Google Dirve
```
phil-davis, line 129:
```suggestion
	* @return a path starting at the real root of Google Drive
```

phil-davis (commented):
Just a typo - I will leave the technical code review to someone else.

apps/files_external/lib/Lib/Storage/Google.php
mmattel: fixed

mmattel:
@phil-davis thanks for finding the typo, this is fixed. Q: I have added a changelog, but the bot does not recognize it. Anything to do here?

phil-davis:
> @phil-davis thanks for finding the typo, this is fixed. Q: I have added a changelog, but the bot does not recognize it. Anything to do here?

The bot is too stupid to delete its comment. I deleted the changelog comment.

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -121,13 +123,37 @@ private function getDefaultFieldsForFolderScan() {
...
+		return $this->getDriveFileHelper($this->getAbsolutePath($path));
+	}
+
+	/**
+	 * Get the Google_Service_Drive_DriveFile object for the specified path.
+	 * Returns false on failure.
+	 * @param string $path
+	 * @return \Google_Service_Drive_DriveFile|false
+	 */
+	private function getDriveFileHelper($path) {
```
jvillafanez, line 156: I think this needs a different name.
For me, a helper is an **object** which helps to do things. As object, it has several methods in order to perform the tasks.

My usage expectation with this function is something as below.
```
$helper = $this->getDriveFileHelper();
$absolutePath = $helper->getAbsolutePath($root, $path);
```
which doesn't match the current implementation.

Maybe `getInternalDriveFile` might fit better.

Note that the method description is the PHPDoc should be different because the method does a different thing. Otherwise I could use `a($path)` and `b($path)`

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -90,7 +92,7 @@ public function usePartFile() {
 	}
 
 	public function getId() {
-		return $this->id;
+		return $this->id.'/'.$this->root;
```
jvillafanez, line 95: We might need to adjust this for the upgrade.
Before the upgrade -> "id001rad". After the upgrade -> "id001rad/".
Note that the ending "/" will cause ownCloud to create a new storage because the id is different.

```
if ($this->root === '') {
  return "{$this->id}/{$this->root}";
} else {
  return $this->id;
}
```
seems a better choice

jvillafanez:
> One can now mount from a GD subfolder. This has big benefits:
> * only a fraction of your GD is mounted and seen
> * it is now possibe to use the $user variable to have a mount based on a subfolder structure like: sub_folder/$user !
> * when using encryption, only the subfolder gets encrypted

For me, the only selling point is the first one. The other 2 are side effects, which might not be bad.

Anyway, pinging @pmaier1 for him to be aware of this change.

apps/files_external/lib/Lib/Storage/Google.php
mmattel: You are right. This function is called by the function above which was the original function. Therefore I kept the PHPDoc as it was. I changed in the upcoming commit both PHPDoc statements.

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -121,13 +123,37 @@ private function getDefaultFieldsForFolderScan() {
...
+	/**
+	* Transform an external path to one originating from the virtual root.
+	* @param $path the path relative to the virtual root directory
+	* @return a path starting at the real root of Google Drive
+	*/
+	private function getAbsolutePath($path) {
+		if ($path === '.') {
+			$path = '';
+		}
+		$path = $this->root !== '' ? $this->root.'/'.$path : $path;
```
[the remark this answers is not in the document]
mmattel: fixed

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -474,6 +509,7 @@ public function rename($path1, $path2) {
 	}
 
 	public function fopen($path, $mode) {
+		//$path = $this->getAbsolutePath($path);
```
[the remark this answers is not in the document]
mmattel: fixed

apps/files_external/lib/Lib/Storage/Google.php
mmattel: fixed

apps/files_external/lib/Lib/Storage/Google.php
mmattel: fixed

mmattel:
@jvillafanez I fixed all your comments, retested and squashed the commits before pushing. Thanks for taking a look over!

mmattel:
Note regarding selling points:
One driver for me picking this up was, I do not want to use the whole GD but only a fraction of it AND this fraction should be encrypted. The fraction part is used as a backup folder for data I mirror via the client so it cant get lost. This needs to be encrypted as it is on a foreign infrastructure. One having sensitive data can now use this functionality and not waste the whole GD drive for it. Hoping that this can go into 10.6.

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -90,7 +92,11 @@ public function usePartFile() {
 	}
 
 	public function getId() {
-		return $this->id;
+		if ($this->root === '') {
```
jvillafanez, line 95: The condition should be the opposite.

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -122,12 +128,37 @@ private function getDefaultFieldsForFolderScan() {
 	}
 
 	/**
-	 * Get the Google_Service_Drive_DriveFile object for the specified path.
+	* Transform an external path to one originating from the virtual root.
```
jvillafanez, line 131: It seems this needs an additional space to align the asterisks.

apps/files_external/lib/Lib/Storage/Google.php
```diff
@@ -122,12 +128,37 @@ private function getDefaultFieldsForFolderScan() {
...
 	 * Returns false on failure.
 	 * @param string $path
 	 * @return \Google_Service_Drive_DriveFile|false
 	 */
 	private function getDriveFile($path) {
+		return $this->getInternalDriveFile($this->getAbsolutePath($path));
+	}
+
+	/**
+	 * Get the elements of the absolut path in case a subfolder is used.
```
jvillafanez, line 156: Typo: "absolute"

jvillafanez (commented):
I wrote the condition wrongly and you copied wrongly :)
The other 2 are just minor changes taking advantage that you need to change the condition

jvillafanez:
We're starting the release process so I doubt we'll add more features for 10.6 unless they're strictly needed. I'm not the one making the decision, so things might change.

apps/files_external/lib/Lib/Storage/Google.php
mmattel: fixed 👍 :man_facepalming:

apps/files_external/lib/Lib/Storage/Google.php
mmattel: fixed

apps/files_external/lib/Lib/Storage/Google.php
mmattel: fixed

phil-davis:
Note: there is currently a problem with drone not actually starting pipelines. Hopefully will be resolved soon.

mmattel:
@pmaier1 I would be very happy if you would decide that this PR goes into 10.6

jvillafanez:
No more code changes from my side :+1: . Waiting for CI before approving. Anyway, we still need confirmation whether this will be on 10.6 or not.

phil-davis:
I added @micbar to the reviewers so that he notices this tomorrow and also can give an opinion.

micbar:
I am ooto until thursday. Restarted the build.

micbar:
Reasonable Change IMO. But 10.6 is already leaving. There is ongoing QA on 10.6.0RC1.

mmattel:
> There is ongoing QA on 10.6.0RC1

Is there anything I could do to support a mind change ? :smirk:
(Who not asks will not get the chance for a different outcome :smile:)

mmattel:
@jvillafanez 
> No more code changes from my side

mind to approve?

jvillafanez (approved):

PR #38181, opened 2020-12-02
Do not redirect if browser asks for a .properties file

jvillafanez:
## Description
files_pdfviewer app asks for a locale.properties file in order to handle the translations. In some setups, this request gets redirected to the files view, so the app can't get the translations.
This change makes possible to get the file so the files_pdfviewer can translate the pdf viewer content according to the language set by the user.

## Related Issue
No issue related. Found in https://github.com/owncloud/files_pdfviewer/pull/228#issuecomment-736578183

## Motivation and Context
The whole UI should be shown in the same language.

## How Has This Been Tested?
1. Install OC via docker setup (used official docker setup for 10.5)
2. Install files_pdfviewer app
3. As user1, change the ownCloud language settings for the user to anything other than English
4. As user1, upload a pdf file
5. As user1, access to that pdf file. The pdf viewer should open the file. The UI related to the pdf viewer should be translated

Screenshot from 2020-12-02 10-50-15

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38181
+++ b/changelog/unreleased/38181
@@ -1,0 +1,10 @@
+Bugfix: Don't redirect if the browser ask for a .properties file
+
+In order to provide translations, the files_pdfviewer app requested a
+.properties file. This request failed because the server redirected it to
+the default page (the files view), so the app couldn't get the translations
+
+This redirection doesn't happen any longer, and the app can translate the
+UI elements now.
+
+https://github.com/owncloud/core/pull/38181
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -489,7 +489,7 @@ class Setup {
 			$content .= "\n  RewriteRule ^favicon.ico$ core/img/favicon.ico [L]";
 			$content .= "\n  RewriteRule ^core/js/oc.js$ index.php [PT,E=PATH_INFO:$1]";
 			$content .= "\n  RewriteRule ^core/preview.png$ index.php [PT,E=PATH_INFO:$1]";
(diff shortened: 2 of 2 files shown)


jvillafanez:
Couple of things to consider:
* This should probably be included in 10.6
* The change will likely affect only to new installations. Updated installations might need to manually adjust this, or we might need to provide a migration / repair step

mmattel:
> Updated installations might need to manually adjust this, or we might need to provide a migration / repair step

I would not recommend a manual driven adoption for existing installations. How ill you propagate this properly so it will be implemented? Just giving a note in the changelog and hope is imho insufficient...
(But thanks for raising that point.)

jvillafanez:
The touched function is called during the upgrade, and there is also an occ command to update the .htaccess manually if needed.
QA might still need to verify it, but I don't think we need to touch additional things.

VicDeo (approved):
Makes sense

mmattel:
You should add a changelog to satisfy the bot :smile:
This is maybe a good place to put a note to use the occ command for existing installations

jvillafanez:
I haven't seen a complain about this, so I don't think it's critical enough to make noise.
Existing installation can either update to pick this change (when it's released), or apply the patch. In the second case, they should know what they're doing anyway and what are they applying, otherwise they should wait until a new version with the fix is released.

mmattel:
All green, merging?

PR #38190, opened 2020-12-03
Catch StorageNotAvailableException while formatting the share

jvillafanez:
The getById method called while formatting could go down to the storage
if somehow the node needs to be updated. If so, the storage where the
node is could not be available, causing the exception to pop up.
Such exception causes the list of shares to be broken since it wasn't
being caught.

## Description
Catch a possible StorageNotAvailableException that could be thrown while formatting the shares.
This specially affects to the share listing, which could be broken and the UI might not show any local share if such exception happens.

So far, this has been detected only while listing the shares, and the steps to reproduce aren't fully clear. There are some other locations for the `formatShare` method being called, although it's expected to behave better since other similar exception that could happen such as `NotFoundException` aren't being handled in those places.

## Related Issue
Related to https://github.com/owncloud/enterprise/issues/4250

## Motivation and Context
Could lead to an empty list of shares

## How Has This Been Tested?
Not tested yet

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/Share20OcsController.php
+++ b/apps/files_sharing/lib/Controller/Share20OcsController.php
@@ -31,6 +31,7 @@ use OCP\Files\File;
 use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
+use OCP\Files\StorageNotAvailableException;
 use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IL10N;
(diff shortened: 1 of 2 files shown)


cdamken:
Works with Damken-Cloud!

karakayasemi (dismissed):

karakayasemi:
We need changelog entry before merge.

jvillafanez:
I assume we don't need to rush so this will be in 10.7. I haven't been able to reproduce the scenario locally, so maybe the conditions to reach the issue cannot happen in the recent versions.

jvillafanez:
Added changelog item

karakayasemi (approved):

PR #38196, opened 2020-12-07
Add missing event emitters for Symfony EventDispatcher

### LINKED ISSUE #38195: Missing event emitters for Symfony EventDispatcher
Currently we are missing event emitters for *postUserCreate* and *postSetPassword*. We need those when working with the Symfony EventDispatcher. This issue occured as we switched to the Symfony component in https://github.com/owncloud/encryption/pull/25

### Expected behaviour
The events should fire and the Symfony EventDispatcher should listen/receive to them.

### Actual behaviour
The Symfony event dispatcher does not receive the event calls.

JammingBen:
## Description
Right now we are missing emitters for the events postUserCreate and postSetPassword. This change adds those calls as we need them when working with the Symfony EventDispatcher.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/38195

## Motivation and Context
This issue came up as we changed the event dispatchers in the Encryption app: https://github.com/owncloud/encryption/pull/25/

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38195
+++ b/changelog/unreleased/38195
@@ -1,0 +1,7 @@
+Change: Add missing event emitters for Symfony EventDispatcher
+
+Prior to this fix we were missing emitters for the events postUserCreate and
+postSetPassword. This change adds those calls as we need them when working
+with the Symfony EventDispatcher.
+
+https://github.com/owncloud/core/issues/38195
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -339,6 +339,9 @@ class Server extends ServerContainer implements IServerContainer, IServiceLoader
 			$userSession->listen('\OC\User', 'postCreateUser', function ($user, $password) {
 				/** @var $user \OC\User\User */
 				\OC_Hook::emit('OC_User', 'post_createUser', ['uid' => $user->getUID(), 'password' => $password]);
+				$this->emittingCall(function () {
+					return true;
+				}, ['before' => [], 'after' => ['uid' => $user->getUID(), 'password' => $password]], 'user', 'create');
 			});
 			$userSession->listen('\OC\User', 'preDelete', function ($user) {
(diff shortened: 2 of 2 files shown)


mmattel:
@JammingBen having the label `3-to review` needs reviewers :smile:
I added some I think the can take a look at

mmattel:
@jvillafanez @micbar 
Could you take a look and approve if ok?
This is needed to continue a PR in encryption as described on top.

jvillafanez:
I'd prefer to reconsider this and look for a different solution.

The `emittingCall` was (badly, IMHO) designed to wrap a function inside a "before" and "after" events. This is similar to:
```
callBeforeEvents($params);
function();
callAfterEvents($params);
```
Even if the "before events" don't trigger due to the empty parameters for the call, your function doesn't do anything other than returning true. For me, it's adding more confusion to an already-confusing method.
I'd expect your function to actually create the user or set the user's password. Right now, you're twisting the expected usage.

Is it possible for the encryption to get an instance of the user's session and listen directly there?

If there isn't any other quick alternative I guess we could accept this for now. However, we should rethink how we can deal with this.

mmattel:
> If there isn't any other quick alternative I guess we could accept this for now. However, we should rethink how we can deal with this.

In case this is currently the only way, it might be good to document this in the changes made (both here and in the corresponding encryption PR)

JammingBen:
Closing this PR as I found a way to use the existing event emitters in https://github.com/owncloud/encryption/pull/25.

* https://github.com/owncloud/core/blob/master/lib/private/User/User.php#L323 for *user.aftersetpassword*
* https://github.com/owncloud/core/blob/master/lib/private/User/Manager.php#L414 for *user.aftercreate*

PR #38247, opened 2020-12-23
[Tests-Only] add tests for downloading file with range request

swoichha:
## Description
This PR adds tests for downloading file with range request

## Related Issue
- https://github.com/owncloud/ocis/issues/1026

## How Has This Been Tested?
- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavOperations/downloadFile.feature
+++ b/tests/acceptance/features/apiWebdavOperations/downloadFile.feature
@@ -129,3 +129,71 @@ Feature: download file
       | new         | "sample,1.txt" |
       | new         | ",,,.txt"      |
       | new         | ",,,.,"        |
+
+  Scenario Outline: download a file with single part ranges
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=0-51" using the WebDAV API
+    Then the HTTP status code should be "206"
+    And the following headers should be set
+      | header         | value         |
+      | Content-Length | 52            |
+      | Content-Range  | bytes 0-51/52 |
+    And the downloaded content should be "Welcome this is just an example file for developers."
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
+
+  Scenario Outline: download a file with multipart ranges
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=0-6, 40-51" using the WebDAV API
+    Then the HTTP status code should be "206" or "200"
+    And if the HTTP status code was "206" then the following headers should match these regular expressions
+      | Content-Length | /\d+/                                               |
+      | Content-Type   | /^multipart\/byteranges; boundary=[a-zA-Z0-9_.-]*$/ |
+    And if the HTTP status code was "206" then the downloaded content for multipart byterange should be:
+      """
+      Content-type: text/plain;charset=UTF-8
+      Content-range: bytes 0-6/52
+
+      Welcome
+
+      Content-type: text/plain;charset=UTF-8
+      Content-range: bytes 40-51/52
+
+      developers.
+      """
+    But if the HTTP status code was "200" then the downloaded content should be "Welcome this is just an example file for developers."
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
+
+  Scenario Outline: download a file with last byte range out of bounds
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=0-55" using the WebDAV API
+    Then the HTTP status code should be "206"
+    And the downloaded content should be "Welcome this is just an example file for developers."
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
+
+  Scenario Outline: download a range at the end of a file
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=-11" using the WebDAV API
+    Then the HTTP status code should be "206"
+    And the downloaded content should be "developers."
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
+
+  Scenario Outline: download a file with range out of bounds
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=55-60" using the WebDAV API
+    Then the HTTP status code should be "416"
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -883,11 +883,45 @@ trait WebDav {
 		// But if the content is wrong (e.g. empty) then it is useful to
 		// report the HTTP status to give some clue what might be the problem.
 		$actualStatus = $this->response->getStatusCode();
+		$pattern = ["/--\w*/", "/\s*/m"];
+		$actualContent = \preg_replace($pattern, "", $actualContent);
+		$content = \preg_replace("/\s*/m", '', $content);
 		Assert::assertEquals(
 			$content,
 			$actualContent,
 			"The downloaded content was expected to be '$content', but actually is '$actualContent'. HTTP status was $actualStatus"
 		);
+	}
+
+	/**
+	 * @Then /^if the HTTP status code was "([^"]*)" then the downloaded content for multipart byterange should be:$/
+	 *
+	 * @param int|string $statusCode
+	 * @param PyStringNode $content
+	 *
+	 * @return void
+	 *
+	 */
+	public function theDownloadedContentForMultipartByterangeShouldBe($statusCode, PyStringNode $content) {
+		$actualStatusCode = $this->response->getStatusCode();
+		if ($actualStatusCode === $statusCode) {
+			$this->downloadedContentShouldBe($content->getRaw());
+		}
+	}
+
+	/**
+	 * @Then /^if the HTTP status code was "([^"]*)" then the downloaded content should be "([^"]*)"$/
+	 *
+	 * @param int|string $statusCode
+	 * @param string $content
+	 *
+	 * @return void
+	 */
+	public function checkStatusCodeForDownloadedContentShouldBe($statusCode, $content) {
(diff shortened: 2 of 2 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/core/28080/2/7 fails coding standard. Please run `make test-php-style` before pushing PRs.

tests/acceptance/features/apiWebdavOperations/downloadFile.feature
```diff
@@ -129,3 +129,61 @@ Feature: download file
...
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=0-55" using the WebDAV API
+    Then the HTTP status code should be "206"
+    And the downloaded content should be "Welcome this is just an example file for developers."
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
+
+  Scenario Outline: download a file with invalid range
```
phil-davis, line 181:
```suggestion
  Scenario Outline: download a range at the end of a file
```

tests/acceptance/features/apiWebdavOperations/downloadFile.feature
```diff
@@ -129,3 +129,61 @@ Feature: download file
...
+      | header         | value         |
+      | Content-Length | 52            |
+      | Content-Range  | bytes 0-51/52 |
+    And the downloaded content should be "Welcome this is just an example file for developers."
+    Examples:
+      | dav_version |
+      | old         |
+      | new         |
+
+  Scenario Outline: download a file with multipart ranges
```
phil-davis, line 147: Note: this is failing in CI. It gets a `200` status instead of `206`.
https://tools.ietf.org/html/rfc7233#section-4 says that a server can choose to combine a multi-part range request, and return a combined response that includes some data "in the middle" if the server "thinks" that is more efficient/easier... And can even choose to return the whole file.
In this case it looks like the server returns the whole file.
It might be different in CI from in your local setup - if so, then that will be a bit if a pain for tests.

phil-davis (commented):
I wonder if ownCloud10 will ever return a 416 "Range not satisfiable" - https://tools.ietf.org/html/rfc7233#section-4.4

It would be good to have a test scenario that asks for a range that is completely outside the file, and see what happens.

tests/acceptance/features/apiWebdavOperations/downloadFile.feature
```diff
@@ -129,3 +129,61 @@ Feature: download file
...
+    Given using <dav_version> DAV path
+    When user "Alice" downloads file "/welcome.txt" with range "bytes=0-6, 40-51" using the WebDAV API
+    Then the HTTP status code should be "206"
```
phil-davis, line 150:
```suggestion
    Then the HTTP status code should be "206" or "200"
```

The "206" Then step could be changed like this in all the scenarios, because an implementation can choose to return 206 or 200.

tests/acceptance/features/apiWebdavOperations/downloadFile.feature
```diff
@@ -129,3 +129,61 @@ Feature: download file
...
+    And the following headers should match these regular expressions
```
phil-davis, line 151:
```suggestion
    And if the HTTP status code was "206" then the following headers should match these regular expressions
```

We could write "if statement" "Then" steps.

tests/acceptance/features/apiWebdavOperations/downloadFile.feature
```diff
@@ -129,3 +129,61 @@ Feature: download file
...
+      | Content-Length | /\d+/                                               |
+      | Content-Type   | /^multipart\/byteranges; boundary=[a-zA-Z0-9_.-]*$/ |
+    And the downloaded content for multipart byterange should be:
```
phil-davis, line 154:
```suggestion
    And if the HTTP status code was "206" then the downloaded content for multipart byterange should be:
    ...
    And if the HTTP status code was "200" then the downloaded content should be "Welcome this is just an example file for developers."
```

And do something like this for checking the content.

phil-davis (commented):
Maybe we have to have "if" in the gherkin step text, because the "expected" result of a range request can be different things depending on the system-under-test.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -883,13 +883,49 @@ public function downloadedContentShouldBe($content) {
...
+	 * @return null
+	 *
+	 */
+	public function theDownloadedContentForMultipartByterangeShouldBe($statusCode, PyStringNode $content) {
+		$actualStatusCode = $this->response->getStatusCode();
+		if ($actualStatusCode === $statusCode) {
+			$this->downloadedContentShouldBe($content->getRaw());
+		}
+		return null;
+	}
```
HariBhandari07, line 911: This code will pass even when status code is not `206` as it will return null. So, IMO it's better to assert the status code and then check for the download.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -883,13 +883,49 @@ public function downloadedContentShouldBe($content) {
...
+	 *
+	 * @return null
+	 */
+	public function checkStatusCodeForDownloadedContentShouldBe($statusCode, $content) {
+		$actualStatusCode = $this->response->getStatusCode();
+		if ($actualStatusCode === $statusCode) {
+			$this->downloadedContentShouldBe($content);
+		}
+		return null;
+	}
```
HariBhandari07, line 927: same as above

tests/acceptance/features/bootstrap/WebDav.php
phil-davis: The Gherkin will already check that the status code is one of 200 or 206 in a previous step.
The code here is what we want - we only want to do this particular `downloadedContentShouldBe` check if the request had the given status code.

If the request had some other status code, then this step should do nothing - because in that case we do not expect the content to have multi-part...

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -883,13 +883,49 @@ public function downloadedContentShouldBe($content) {
...
+	 * @param int|string $statusCode
+	 * @param PyStringNode $content
+	 *
+	 * @return null
+	 *
+	 */
+	public function theDownloadedContentForMultipartByterangeShouldBe($statusCode, PyStringNode $content) {
+		$actualStatusCode = $this->response->getStatusCode();
+		if ($actualStatusCode === $statusCode) {
+			$this->downloadedContentShouldBe($content->getRaw());
```
phil-davis, line 908: Note: `$content->getRaw()` will be a whole `PyString` that has the details of the various multi-part sections. That is written in the actual Gherkin in the feature file. So that is why there is no special code needed here to analyse the details of the multi-part response.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -883,13 +883,49 @@ public function downloadedContentShouldBe($content) {
...
+		}
+		return null;
```
phil-davis, line 910: IMO we do not need this - the code of a step can happily finish without returning anything.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -883,13 +883,49 @@ public function downloadedContentShouldBe($content) {
...
+	 * @param string $content
+	 *
+	 * @return null
+	 */
+	public function checkStatusCodeForDownloadedContentShouldBe($statusCode, $content) {
+		$actualStatusCode = $this->response->getStatusCode();
+		if ($actualStatusCode === $statusCode) {
+			$this->downloadedContentShouldBe($content);
+		}
+		return null;
```
phil-davis, line 926: IMO we do not need this - the code of a step can happily finish without returning anything.

tests/acceptance/features/bootstrap/WebDav.php
phil-davis: This is fine - if the status code was not the code mentioned in the step, then we want this step to do nothing.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -3288,6 +3324,23 @@ public function headersShouldMatchRegularExpressions(TableNode $table) {
...
+	 *
+	 * @return null
+	 * @throws \Exception
+	 */
+	public function statusCodeShouldMatchTheseRegularExpressions($statusCode, TableNode $table) {
+		$actualStatusCode = $this->response->getStatusCode();
+		if ($actualStatusCode === $statusCode) {
+			$this->headersShouldMatchRegularExpressions($table);
+		}
+		return null;
```
phil-davis, line 3341: IMO we do not need this - the code of a step can happily finish without returning anything.

tests/acceptance/features/bootstrap/WebDav.php
HariBhandari07: I didn't read the scenario steps :wink:  It's fine.

tests/acceptance/features/apiWebdavOperations/downloadFile.feature
```diff
@@ -129,3 +129,71 @@ Feature: download file
...
+      Content-range: bytes 0-6/52
+
+      Welcome
+
+      Content-type: text/plain;charset=UTF-8
+      Content-range: bytes 40-51/52
+
+      developers.
+      """
+    And if the HTTP status code was "200" then the downloaded content should be "Welcome this is just an example file for developers."
```
phil-davis, line 166:
```suggestion
    But if the HTTP status code was "200" then the downloaded content should be "Welcome this is just an example file for developers."
```

Maybe "But" here will help people to understand the scenario?

tests/acceptance/features/bootstrap/WebDav.php
swoichha: There are two status code cases in which we want to check the downloaded content. So, if we use assert to check the status code `206` and the actual status code is different from the expected code e.g `200` then in such case whole test will fail and it won't go to the next step where it will again check the status code  for `200`. The downloaded content for each status code is different so we are  checking it this way using if in Then step of feature file.

phil-davis (approved):

Issue #10831, opened 2014-09-03
Use binary files for encryption instead of base64 encoded ones


PVince81:
Currently encrypted files produce a 33% size increase, because a 6126 bytes block produces a 8192 bytes encrypted block.
Also the encrypted file is stored in an ASCII format which seems to be base64.

We should look into switching to a binary format, if possible, to hopefully reduce the size of encrypted files.

The new header that was added for AES-256 can be reused to specify a new flag that tells whether the encrypted file is in binary format or the old ascii format.

@schiesbn what do you think ?

PVince81:
Could be implemented as a separate encryption module.

Hunter-Github:
Raised at [IT Security StackExchange](https://security.stackexchange.com/questions/117068/why-would-an-encrypted-file-be-35-larger-than-an-unencrypted-one).

tyzoid:
It seems odd to be imposing a premium on space where none would otherwise exist. This also has other minor effects, such as on the OwnCloud Linux Client showing the raw value during upload, then jumps up to the "correct" value of usage once the sync finishes. From the perspective of both an administrator and user, it can be annoying: the quotas that are assigned count the base-64 encoded data on disk, rather than the data the user actually wants to store.

PVince81:
The quota as implemented in ownCloud takes care of the unencrypted file size, not the encrypted one.
But if you have FS-level quotas then those would apply on the encoded data, yes.

Is the raw upload value bug still there ?

I think this issue also applies to master key encryption.

cc @pmaier1

tyzoid:
@PVince81 When I wrote my comment, there was no disk quota in use. The client was counting encrypted, base64 data as 'used'.

karakayasemi:
Fixed with https://github.com/owncloud/encryption/pull/224

PR #38249, opened 2020-12-27
determine unencrypted block size after begin operation

karakayasemi:
## Description
Unencrypted block size of encrypted file can change with different encoding types. Determine unencrypted block size after begin operation. In this way, EncryptionModule can decide block size after reading the header of the file.

## Related Issue
https://github.com/owncloud/encryption/issues/210
https://github.com/owncloud/core/issues/10831

## How Has This Been Tested?
https://github.com/owncloud/encryption/pull/224 is able to support both base64 and binary encoding after this change.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38249
+++ b/changelog/unreleased/38249
@@ -1,0 +1,7 @@
+Bugfix: Determine unencrypted block size after begin operation
+
+Unencrypted block size of encrypted file can change with different encoding types.
+Unencrypted block size determination has been moved after begin operation in stream opening.
+In this way, EncryptionModule can decide block size after reading the header of the file.
+
+https://github.com/owncloud/core/pull/38249
--- a/lib/private/Files/Stream/Encryption.php
+++ b/lib/private/Files/Stream/Encryption.php
@@ -234,7 +234,6 @@ class Encryption extends Wrapper {
 		$this->position = 0;
 		$this->cache = '';
 		$this->writeFlag = false;
-		$this->unencryptedBlockSize = $this->encryptionModule->getUnencryptedBlockSize($this->signed);
 
 		if (
 			$mode === 'w'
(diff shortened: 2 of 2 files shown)


micbar:
Sounds reasonable. Can you explain a bit more the "why?"

karakayasemi:
@micbar Unencrypted block sizes of old and new encoding will be different after https://github.com/owncloud/encryption/pull/224. We are planning to add encoding type to header to support old and new encoding formats together. 

If we determine unencrypted block size after begin operation, EncryptionModule can decide block size by reading the header of the file.

mmattel:
@micbar could you continue to review so this can be approved and merged?

micbar:
@karakayasemi @mmattel This change should be covered by tests.

Can we add some unit tests?

If not, try to think about a senseful integration test.

phil-davis:
The encryption app runs all the core API and webUI acceptance tests with encryption enabled in the backend. After the PR is merged, the nightly encryption app CI will be running those.

I made PR The encryption app runs all the core API and webUI acceptance tests with encryption  which is running the tests using this core branch. That will demonstrate that this change does not break things.

If there is also a way to have core unit tests also, then that will be good.

phil-davis:
PR https://github.com/owncloud/encryption/pull/228 passed, so I this change has not broken anything obvious :)

mmattel:
@micbar as @phil-davis comments, does this confident you enough :crossed_fingers: ?

phil-davis (commented):
From an end-to-end testing point of view, I am happy that there is not any regression in existing behaviour. Needs an approval from a dev also.

jvillafanez (approved):

karakayasemi:
@micbar since only the line of an independent function call has changed, I could not find easy unit test scenario. The encryption app acceptance tests scenarios have proved that the change does not break things. So, I have merged the PR.

PR #38252, opened 2020-12-28
[tests-only] Fix: new trashbinFilesFolders.feature test failing in oC10 app testing

saw-jan:
Co-authored-by: SwikritiT <swikriti808@gmail.com>

## Description
`trashbinFilesFolders.feature:285` test was failing sometime due to the time difference between `mtime` in response and `mtime` saved in `lastUploadDeleteTime`

## Related Issue
https://github.com/owncloud/QA/issues/650

## How Has This Been Tested?

- local environment

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/TrashbinContext.php
+++ b/tests/acceptance/features/bootstrap/TrashbinContext.php
@@ -260,16 +260,23 @@ class TrashbinContext implements Context {
 		foreach ($elementRows as $expectedElement) {
 			$found = false;
 			$expectedPath = $expectedElement['path'];
-			$expectedMtime = $expectedElement['mtime'] === "deleted_mtime" ? $this->featureContext->getLastUploadDeleteTime() : $expectedElement['mtime'];
+			$expectedMtime = $this->featureContext->getLastUploadDeleteTime();
+			$responseMtime = '';
 			
 			if (isset($expectedElement['mtime'])) {
 				foreach ($files as $file) {
-					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/") && \ltrim($expectedMtime, "/") === \ltrim($file['mtime'], "/")) {
-						$found = true;
-						break;
+					$responseMtime = $file['mtime'];
+
+					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/")) {
+						$mtime_difference = \abs((int)\trim($expectedMtime) - (int)\trim($file['mtime']));
+
+						if ($mtime_difference <= 2) {
+							$found = true;
+							break;
+						}
 					}
 				}
-				Assert::assertTrue($found, "$expectedPath with mtime $expectedMtime expected to be listed in response but not found");
+				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime $responseMtime but found $expectedMtime");
 			} else {
 				foreach ($files as $file) {
 					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/")) {
--- a/tests/acceptance/features/bootstrap/TrashbinContext.php
+++ b/tests/acceptance/features/bootstrap/TrashbinContext.php
@@ -275,7 +275,7 @@ class TrashbinContext implements Context {
 						}
 					}
 				}
-				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime '$responseMtime' but found '$expectedMtime'");
+				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime '$expectedMtime' but found '$responseMtime'");
 			} else {
 				foreach ($files as $file) {
 					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/")) {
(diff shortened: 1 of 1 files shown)


phil-davis:
@saw-jan https://drone.owncloud.com/owncloud/core/28105/2/7
This fails code-style checks.
Can you fix that, and hopefully CI will pass...

tests/acceptance/features/bootstrap/TrashbinContext.php
```diff
@@ -260,16 +260,23 @@ public function theLastWebdavResponseShouldContainFollowingElements(TableNode $e
...
 			
 			if (isset($expectedElement['mtime'])) {
 				foreach ($files as $file) {
-					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/") && \ltrim($expectedMtime, "/") === \ltrim($file['mtime'], "/")) {
-						$found = true;
-						break;
+					$responseMtime = $file['mtime'];
+
+					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/")) {
+						$mtime_difference = \abs((int)\trim($expectedMtime) - (int)\trim($file['mtime']));
```
phil-davis, line 271:
```suggestion
						$mtime_difference = \abs((int)\trim($expectedMtime) - (int)\trim($responseMtime));
```

tests/acceptance/features/bootstrap/TrashbinContext.php
```diff
@@ -260,16 +260,23 @@ public function theLastWebdavResponseShouldContainFollowingElements(TableNode $e
...
-			$expectedMtime = $expectedElement['mtime'] === "deleted_mtime" ? $this->featureContext->getLastUploadDeleteTime() : $expectedElement['mtime'];
+			$expectedMtime = $this->featureContext->getLastUploadDeleteTime();
+			$responseMtime = '';
 			
 			if (isset($expectedElement['mtime'])) {
 				foreach ($files as $file) {
-					if (\ltrim($expectedPath, "/") === \ltrim($file['original-location'], "/") && \ltrim($expectedMtime, "/") === \ltrim($file['mtime'], "/")) {
-						$found = true;
-						break;
+					$responseMtime = $file['mtime'];
```
phil-davis, line 268: This line can be moved inside the `if` statement below. That way we only set it when we actually find the file that we are looking for. (Otherwise we might end the loop and will have overwritten `$responseMtime` with the value for some other file, and the failure message will not be so useful)

tests/acceptance/features/bootstrap/TrashbinContext.php
```diff
@@ -260,16 +260,23 @@ public function theLastWebdavResponseShouldContainFollowingElements(TableNode $e
...
+						$mtime_difference = \abs((int)\trim($expectedMtime) - (int)\trim($file['mtime']));
+
+						if ($mtime_difference <= 2) {
+							$found = true;
+							break;
+						}
 					}
 				}
-				Assert::assertTrue($found, "$expectedPath with mtime $expectedMtime expected to be listed in response but not found");
+				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime $responseMtime but found $expectedMtime");
```
phil-davis, line 279:
```suggestion
				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime '$expectedMtime' but found '$responseMtime'");
```

I put single-quotes around the values, because `$responseMtime` might be an empty string, and the single-quotes will make that a bit easier to guess in the error message.

tests/acceptance/features/bootstrap/TrashbinContext.php
```diff
@@ -260,16 +260,22 @@ public function theLastWebdavResponseShouldContainFollowingElements(TableNode $e
...
+						$mtime_difference = \abs((int)\trim($expectedMtime) - (int)\trim($responseMtime));
+
+						if ($mtime_difference <= 2) {
+							$found = true;
+							break;
+						}
 					}
 				}
-				Assert::assertTrue($found, "$expectedPath with mtime $expectedMtime expected to be listed in response but not found");
+				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime '$responseMtime' but found '$expectedMtime'");
```
phil-davis, line 278:
```suggestion
				Assert::assertTrue($found, "$expectedPath expected to be listed in response with mtime '$expectedMtime' but found '$responseMtime'");
```

phil-davis (approved):

phil-davis:
Note: this is a test that checks the time of some event as recorded by the server and compares it with some expected value. The challenges with all tests like that includde:

1) the system time on the system-under-test might be different than on the system running the tests (by just a few seconds, or they might be in different timezones!)
2) the test runner and/or the system-under-test might be slow, so when the test runner remembers the "current time" and when the system-under-test stores the "current time" that the operation/event was done, those 2 time might be different
3) whichever order the test runner and the system-under-test record the "current time" they might be just either side of a 1-second boundary. And so, in the case of time-stamps with 1-second granularity, there is always the potential for a 1-second difference.

The fix here addresses point (3) and helps a lot with point (2).

We could write more code. For example "remember" the "current time" both before and after doing the upload or delete. That might get us 2 different seconds. Then we could expect the server to record one of these times, or a time in-between. But that seems a lot of messing-about when it could still go wrong if the time on the system-under-test is different anyway.

PR #38266, opened 2021-01-04
[WIP] Guzzle6 for core

phil-davis:
## Description
- [x] get adjusted code from old PR #37011 (that itself came from PR #32357 )
- [x] Update guzzle and other dependencies in apps/files_external/3rdparty like https://github.com/owncloud/core/pull/37011/commits/6225510ab8f63c9c3e70d2af635e6bed878ef7fb

## Motivation and Context
Guzzle5 does not have official support on PHP 7.4 and probably never will have. Guzzle5 works on PHP 7.4
For future PHP versions (e.g. PHP 8.0) we might be forced to do this Guzzle update.

Note1: if we do this then apps that use Guzzzle5 will also need updating to Guzzle6 and coordinated release of core and app versions etc.

Note2: guzzle is actually up to release https://github.com/guzzle/guzzle/releases/tag/7.2.0 - so if we bump guzzle we might want to bump it all the way to major version 7.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
Rebased.

phil-davis:
See PR #38706

PR #38269, opened 2021-01-07
[tests-only] Drop laminas-ldap from behat acceptance test dependencies

phil-davis:
## Description
IMO we do not use this in core CI. But it is probably used when the core tests are run from `user_ldap` and also if the core API tests are run from `owncloud/ocis` with an LDAP as the back-end auth behind OIDC...

So I am not so sure that deleting this is a good idea!

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
No need to do this now. A version has been release with PHP 8.0 support.

PR #38280, opened 2021-01-10
Fix subadmin group validation for addToGroup function

### LINKED ISSUE #38279: Subadmins cannot add users to groups they manage
### Background
I have read the background on this topic, through the Github repo issues and PRs and this seems to be a point of conjecture about how subadmins and groups work.  At a high-level, we want to be able to have subadmins create and manage groups in a dynamic fashion (creating groups, adding and removing users to these groups by the subadmin, or deleting groups).  In our system, users do not have any files in their own personal space, and we do not support impersonation for security reasons.  I am not sure how custom groups work internally but cannot see them appear as normal groups via the **owncloud-sdk**.

There appear to be business rules in place so that users created by a subadmin, implicitly get added to the group(s) the creator is a subadmin of.  They can be added and removed to groups (at least in the OC10 UI) freely, but cannot be detached from every group the subadmin is a subadmin of.

Further reading here:
https://github.com/owncloud/core/issues/31016
https://github.com/owncloud/core/issues/12349
https://github.com/owncloud/core/pull/31337

### Steps to reproduce
1.  From a default OC 10 instance, login as **admin** and a user (x), add a group (y) and make user (x) a subadmin of group (y).
2.  In the OC10 PHP UI, login as user (x) create a new user (z) then add user (z) as a member of group (y) - **successful**.
3.  Attempt the same as user (x) calling the **owncloud-sdk** method eg `client.users.addToGroup(z, y)` which fails with a 401 Unauthorised.

This is inconsistent behaviour.  Subadmins are currently (consistently) allowed to remove users from their groups, but not add them.  This means they effectively cannot 'manage' the group, which is the purpose of being a subadmin of a group.  If a subadmin accidently removed a user from a group, they could not re-add.

### Expected behaviour
Correct behaviour should be to allow subadministrators to add users to groups they are a subadmin of.  AARNet needs this feature to properly implement group management policies for delegation of group management.

### Actual behaviour
Fails with a 401 Unauthorised.

## Related Issues
This relates to an old issue and PR that was merged here: https://github.com/owncloud/core/pull/25788 however it seems @IljaN made a commit on 15 May 2018 that broke this feature (see https://github.com/owncloud/core/commit/792728ad9ce59f48009677129048ab81bcfc92cd#diff-5a8fe7b8beda1a977aa1886d1b2aa92f91ab686864c00f58fe9eed947044a054) and this has left the system in an inconsistent state.

_Subadmins need to be able to add users to groups they are subadmins of._

### Server configuration
**ownCloud version:** ownCloud 10.5.0

Fresh install

Docker image

No other settings have any relationship to this raised issue, the issue is in the code for the Provisioning API.

rpocklin:
## Description
This change aligns the provisioning API (and by extension, the **owncloud-sdk**) to allow subadministrators of groups (not just admins) to add users to groups they are a subadministrator of.  This is already permitted via the OC10 UI (which uses the PHP API directly).  The method `removeFromGroup` in this same file already had this check, but it seems the `addToGroup` method was changed and became inconsistent.  This PR makes it consistent across all 3 approaches.

I have re-opened the issue again here:
- Fixes https://github.com/owncloud/core/issues/38279

## Motivation and Context
As detailed above.

## How Has This Been Tested?
Locally it has been tested and works as expected.

There are related acceptances here
https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiProvisioningGroups-v1/addToGroup.feature#L132-L143

https://github.com/rpocklin/core/blob/master/tests/acceptance/features/apiProvisioningGroups-v2/removeFromGroup.feature#L103-L114

## Types of changes
- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/provisioning_api/lib/Users.php
+++ b/apps/provisioning_api/lib/Users.php
@@ -484,7 +484,7 @@ class Users {
 			return new Result(null, 102);
 		}
 
-		if (!$this->groupManager->isAdmin($user->getUID())) {
+		if (!$this->canUserManageGroup($user, $group)) {
 			return new Result(null, 104);
 		}
 
--- a/changelog/unreleased/38280
+++ b/changelog/unreleased/38280
@@ -1,0 +1,3 @@
+Change: Allow subadministrators to add users to groups they manage
+
+https://github.com/owncloud/core/pull/38280


rpocklin:
Rebased to current master.

mmattel:
@micbar @phil-davis no reviewer was added before. Pls feel free to add one 😃

apps/provisioning_api/lib/Users.php
```diff
@@ -484,7 +484,7 @@ public function addToGroup($parameters) {
 			return new Result(null, 102);
 		}
 
-		if (!$this->groupManager->isAdmin($user->getUID())) {
```
mrow4a, line 487: I think isAdmin is still needed, but put it below canUserManageGroup please.

mrow4a (changes requested):

apps/provisioning_api/lib/Users.php
rpocklin: @mrow4a I think that check is encapsulated inside the `canUserManageGroup` method: https://github.com/owncloud/core/blob/089357601f8634710324e84eeea0e1a750682c68/apps/provisioning_api/lib/Users.php#L453-L464

mrow4a (changes requested):
currently does not satisfy requirements from security standpoint -> https://github.com/owncloud/core/issues/38279#issuecomment-760714329

rpocklin:
Closing in favour of toggleGroups approach, see https://github.com/owncloud/core/issues/38279 - will raise an alternate PR for this approach in the future.

PR #38288, opened 2021-01-12
Improve UI success/error while adding storages

AlexAndBear:
## Description
* While adding or changing an external storage, a success/error notification will be shown
* Added cursor->pointer to the status item, due a click results in a recheck

## Related Issue

- Fixes  https://github.com/owncloud/enterprise/issues/4302

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38288
+++ b/changelog/unreleased/38288
@@ -1,6 +1,6 @@
 Enhancement: Improve the UX in the external storage settings page
 
 When a user adds or edits an external storage, the user will see a notification if the storage has been added or an error occured.
-When a user hovers over the status indicator, the user will see a pointer cursor, this clearifies that a click will result in config (re-)check.
+When a user hovers over the status indicator, the user will see a pointer cursor, this clarifies that a click will result in a config (re-)check.
 
 https://github.com/owncloud/core/pull/38288
--- a/changelog/unreleased/38288
+++ b/changelog/unreleased/38288
@@ -1,0 +1,6 @@
+Enhancement: Improve the UX in the external storage settings page
+
+When a user adds or edits an external storage, the user will see a notification if the storage has been added or an error occured.
+When a user hovers over the status indicator, the user will see a pointer cursor, this clearifies that a click will result in config (re-)check.
+
+https://github.com/owncloud/core/pull/38288
--- a/apps/files_external/css/settings.css
+++ b/apps/files_external/css/settings.css
@@ -5,6 +5,7 @@
 }
 
 #externalStorage td.status > span {
+	cursor: pointer;
 	display: inline-block;
 	height: 16px;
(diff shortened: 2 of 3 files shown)


JammingBen (approved):
Looks good to me

changelog/unreleased/38288
```diff
@@ -0,0 +1,6 @@
+Enhancement: Improve the UX in the external storage settings page
+
+When a user adds or edits an external storage, the user will see a notification if the storage has been added or an error occured.
+When a user hovers over the status indicator, the user will see a pointer cursor, this clearifies that a click will result in config (re-)check.
```
phil-davis, line 4:
```suggestion
When a user hovers over the status indicator, the user will see a pointer cursor, this clarifies that a click will result in a config (re-)check.
```

phil-davis (commented):
A typo in the changelog text. Sad to have to run all that CI again just for this.

phil-davis (approved):

JanAckermann:
Can be merged after tech demo

jvillafanez:
I see a potential problem. Let's assume the following setup:
1. SMB / WND with login credentials (shouldn't matter whether they're saved in session or in DB)
2. Both ownCloud and the storage are connected to the windows' AD. This means that only the AD users will be able to access to the external storage. In particular, the admin won't be able to access

Normally, the connectivity check is done using the admin account. If the admin cannot access to the external storage, we assume there is a problem with the setup. However, in the setup above, the problem is that there is no admin in the AD. Using a valid AD account instead of the ownCloud's admin would show a green light instead.

I'm a bit worried that the error message will make more noise even though the setup itself could be fine. A softer message could be better, such as "The current ownCloud's user (in case there are multiple admins) can't access to the external storage. In some scenarios, this could be ok, but please, recheck the configuration"

JanAckermann:
@jvillafanez Edgy case but a good objection, maybe we could show this 'soften message' only if there is an error with the SMB/WND configuration?

jvillafanez:
Technically, it can happen with any storage. I guess the setup above is kind of expected for SMB/WND, so this flaw is more visible there.
For the case of SFTP, for example, the same could also happen, although it's likely easier to create an account for the admin inside the SFTP storage

JanAckermann:
@pmaier1 @cdamken @hodyroff @micbar 
Please take @jvillafanez comment into account for the tech demo meeting

JanAckermann:
@jvillafanez as discussed with PM we will still show this message but enter a hint in the docs

mmattel:
>  as discussed with PM we will still show this message but enter a hint in the docs

@JanAckermann Please create a docs issue

PR #38298, opened 2021-01-14
Prevent group assignment in Web UI if not supported

### LINKED ISSUE #431: Error on adding "guest_app" group to user; No Feedback in UI
### Steps to reproduce
A)
1. Open user administration (/index.php/settings/users)
2. Create new user, select group "guest_app"
3. Click "Create"

OR

B)
1. Add group "guest_app" in Groups column on a existing user

### Expected behaviour
A) New created user is expected to be created as guest user (member of group "guest_app")
B) Existing user is added to the group "guest_app" (= converted to guest user account)

OR

In case of an error show an error message.

### Actual behaviour
In both cases the user is not added to the group "guest_app". The AJAX call to togglegroups.php when adding the group "guest_app" to a existing user gets a `{status: error}` JSON response.
In both cases (A) and B)) the possibility to add the user to the group"guest_app" suggests that this is possible. In both cases no error message is shown. The admin is let to believe to have created a guest user account.

#### Special case ownCloud.online:
Guest accounts are free of charge, normal user accounts initate a payment process. Customers are wondering why they get charged when adding free guest users via the user administration.

### Server configuration
Latest official OC Docker release (oC 10.5)
Guest app v0.9

JammingBen:
## Description
This enhancement checks if users can be assigned to (or removed from) groups via Web UI. All group backends which do not support this functionality will be disabled in corresponding menus.

## Related Issue
- Fixes https://github.com/owncloud/guests/issues/431

image

## Types of changes

- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Group/Manager.php
+++ b/lib/private/Group/Manager.php
@@ -502,6 +502,6 @@ class Manager extends PublicEmitter implements IGroupManager {
 	 * @return \OCP\GroupInterface[]
 	 */
 	public function getBackends() {
-		return \array_values($this->backends);
+		return $this->backends;
 	}
 }
--- a/lib/private/Group/Manager.php
+++ b/lib/private/Group/Manager.php
@@ -496,4 +496,12 @@ class Manager extends PublicEmitter implements IGroupManager {
 		}
 		return false;
 	}
+
+	/**
+	 * Get the active backends
+	 * @return \OCP\GroupInterface[]
+	 */
+	public function getBackends() {
+		return \array_values($this->backends);
+	}
 }
--- a/settings/Controller/GroupsController.php
+++ b/settings/Controller/GroupsController.php
@@ -28,6 +28,7 @@ use OC\AppFramework\Http;
 use OC\Group\MetaData;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\DataResponse;
+use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IL10N;
 use OCP\IRequest;
@@ -45,6 +46,8 @@ class GroupsController extends Controller {
 	private $userSession;
 	/** @var bool */
 	private $isAdmin;
+	/** @var IConfig */
+	private $config;
 
 	/**
 	 * @param string $appName
@@ -53,18 +56,21 @@ class GroupsController extends Controller {
 	 * @param IUserSession $userSession
 	 * @param bool $isAdmin
 	 * @param IL10N $l10n
+	 * @param IConfig $config
 	 */
 	public function __construct($appName,
 								IRequest $request,
 								IGroupManager $groupManager,
 								IUserSession $userSession,
 								$isAdmin,
-								IL10N $l10n) {
+								IL10N $l10n,
+								IConfig $config) {
 		parent::__construct($appName, $request);
 		$this->groupManager = $groupManager;
 		$this->userSession = $userSession;
 		$this->isAdmin = $isAdmin;
 		$this->l10n = $l10n;
+		$this->config = $config;
 	}
 
 	/**
@@ -154,6 +160,24 @@ class GroupsController extends Controller {
 				],
 			],
 			Http::STATUS_FORBIDDEN
+		);
+	}
+
+	/**
+	 * @return DataResponse
+	 */
+	public function getGuestGroupName() {
+		$guestGroupName = '';
+		if (\class_exists('\OCA\Guests\GroupBackend', false) && \OC_App::isEnabled('guests')) {
+			/* @phan-suppress-next-line PhanUndeclaredClassConstant */
+			$guestGroupName = $this->config->getAppValue('guests', 'group', \OCA\Guests\GroupBackend::DEFAULT_NAME);
+		}
+
+		return new DataResponse(
+			[
+				'data' => ['name' => $guestGroupName],
+				Http::STATUS_OK
+			]
 		);
 	}
 }
--- a/settings/Controller/GroupsController.php
+++ b/settings/Controller/GroupsController.php
@@ -160,23 +160,19 @@ class GroupsController extends Controller {
 	/**
 	 * Get available groups for assigning and removing via WebUI.
 	 *
-	 * @param array $groups
 	 * @return DataResponse
 	 */
-	public function getAvailableGroups($groups) {
+	public function getAvailableGroups() {
 		$assignableGroups = [];
 		$removableGroups = [];
-		foreach ($groups as $gID) {
-			$group = $this->groupManager->get($gID);
-			if ($group) {
-				$backend = $group->getBackend();
-				$supportedActions = $backend->getSupportedActions();
-				if ($backend::ADD_TO_GROUP & $supportedActions) {
-					$assignableGroups[] = $gID;
-				}
-				if ($backend::REMOVE_FROM_GROUP & $supportedActions) {
-					$removableGroups[] = $gID;
-				}
+
+		foreach ($this->groupManager->getBackends() as $backend) {
+			$groups = $backend->getGroups();
+			if ($backend->implementsActions($backend::ADD_TO_GROUP)) {
+				\array_push($assignableGroups, ...$groups);
+			}
+			if ($backend->implementsActions($backend::REMOVE_FROM_GROUP)) {
+				\array_push($removableGroups, ...$groups);
 			}
 		}
 
--- a/settings/Controller/GroupsController.php
+++ b/settings/Controller/GroupsController.php
@@ -162,7 +162,7 @@ class GroupsController extends Controller {
 	 *
 	 * @return DataResponse
 	 */
-	public function getAvailableGroups() {
+	public function getAssignableAndRemovableGroups() {
 		$assignableGroups = [];
 		$removableGroups = [];
 
--- a/settings/Controller/GroupsController.php
+++ b/settings/Controller/GroupsController.php
@@ -28,7 +28,6 @@ use OC\AppFramework\Http;
 use OC\Group\MetaData;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\DataResponse;
-use OCP\IConfig;
 use OCP\IGroupManager;
 use OCP\IL10N;
 use OCP\IRequest;
@@ -46,8 +45,6 @@ class GroupsController extends Controller {
 	private $userSession;
 	/** @var bool */
 	private $isAdmin;
-	/** @var IConfig */
-	private $config;
 
 	/**
 	 * @param string $appName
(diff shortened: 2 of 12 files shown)


settings/js/users/users.js
```diff
@@ -642,7 +648,16 @@ var UserList = {
 				// can't become subadmin of "admin" group
 				return;
 			}
-			$groupsSelect.append($('<option value="' + escapeHTML(group) + '">' + escapeHTML(group) + '</option>'));
+			// hide guest group for non-guest-users
+			if (isGuest !== true && group === 'guest_app' && !isSubadminSelect) {
```
jvillafanez, line 652: I think the name of the guest group can be changed

settings/Controller/UsersController.php
```diff
@@ -207,6 +207,10 @@ private function formatUserForIndex(IUser $user, array $userGroups = null) {
 			}
 		}
 
+		$isGuest = (bool) $this->config->getUserValue(
+			$user->getUID(), 'owncloud', 'isGuest', false
```
jvillafanez, line 211: Getting the UID isn't cheap. Better reduce the usage by storing the uid in a local variable and reuse it

settings/js/users/users.js
phil-davis: https://github.com/owncloud/guests/blob/master/lib/GroupBackend.php has code to allow different name.

I guess core code will have to check first if `guests` app is enabled, and if so, then call into https://github.com/owncloud/guests/blob/master/lib/AppInfo/Application.php `getGroupName()` to find out the name.

JammingBen: Ohh good point, I did not find this in the UI, but you are right. Another way would be via config:

`$this->config->getAppValue('guests', 'group', \OCA\Guests\GroupBackend::DEFAULT_NAME)`

I'll dig into that tomorrow :+1:

JammingBen: Ok so I went for an AJAX call to get the guest group name. Not 100% sure if that's the "OC-way" of doing things here, it seems to work though.

jvillafanez:
https://github.com/owncloud/core/pull/38298/files#diff-b229d762bcebc1e3804ce9733dc70e9dfbf5a1890eca0825dafd5e10b2ecafccR173 introduces a dependency from ownCloud to the app. Such dependency shouldn't happen.

We might need to include a new feature to solve this problem.

The group backend interface could expose some options to be implemented. One of them could be "allow adding only specific users". This option not only could force the guest group to allow only guest users, but also force ldap groups to have only ldap users (this isn't restricted at the moment).
The main goal would be for core to know what options are available and how core can interact with those options.

Although providing a good migration path to implement this feature will likely be a challenge (user_ldap and customgroups apps will be affected, maybe more), I think this will be a more consistent solution.

JammingBen:
@jvillafanez 
Thank's for your input regarding this! I looked into it and found the method `getSupportedActions()` for group backends, which already has this functionality, kind of. This way, the controller would check all supported actions for a specific group backend. Then the UI knows what actions to display and what not.

From my point of view, this issue is less about "Prevent user abc from being assigned to group xyz" and more about "Prevent group yxz from being assigned to any user via Web UI (and vice versa)". Plus we wouldn't need to change the interfaces.

JammingBen:
@jvillafanez See https://github.com/owncloud/core/pull/38298/files#diff-b229d762bcebc1e3804ce9733dc70e9dfbf5a1890eca0825dafd5e10b2ecafccR166 for the above mentioned implementation. What do you think?

jvillafanez:
It might fit if we assume that the operations to add and remove guest users are hidden. Basically, only the guest app is allowed to add or remove users from the guest group. I don't like too much relying on private stuff, but the solution could fit for this case.
My only worry is that I'm not fully sure how the guests are added to the group. For consistency, it should be the guest app the one adding the users via this private / internal call because core isn't allowed according to what the backend is reporting.

For now, I think it worth to try

settings/Controller/GroupsController.php
```diff
@@ -164,18 +158,34 @@ public function destroy($id) {
 	}
 
 	/**
+	 * Get available groups for assigning and removing via WebUI.
+	 *
+	 * @param array $groups
```
jvillafanez, line 163: Need more details here. I'm not sure what is the purpose of the parameter.
Just by reading the description, it seems more efficient to get all the group backends, check what of those can add or remove users, and then get a list of the groups in the backend.

If the param refers to the list of groups we want to check for, both the description and the method name should be changed to reflect that we only check those groups.
In addition, I expect a relatively low number of groups to check per call.

JammingBen: > Just by reading the description, it seems more efficient to get all the group backends, check what of those can add or remove users, and then get a list of the groups in the backend.

I agree that would be more efficient. Unfortunately I did not find a way to get all group backends as there is no `getBackends()` method like the `UserManager` has. Do you think it's worth implementing? Would of course mean changing the interface of the `GroupManager`...

JammingBen:
As discussed I implemented a `getBackends` method for the `GroupManager`. Should be far more efficient now.

Technically speaking we now have a breaking change as I had to extend the `IGroupManager` interface. Although I did not find any OC apps implementing it... Let's wait for the pipeline first.

lib/private/Group/Manager.php
```diff
@@ -496,4 +496,12 @@ public function inGroup($uid, $gid) {
 		}
 		return false;
 	}
+
+	/**
+	 * Get the active backends
+	 * @return \OCP\GroupInterface[]
+	 */
+	public function getBackends() {
+		return \array_values($this->backends);
```
jvillafanez, line 505: The `array_values` will create a new array, consuming time and space. I don't expect any impact but it doesn't add anything. Returning the array directly is better in this case.

settings/Controller/GroupsController.php
```diff
@@ -156,4 +156,34 @@ public function destroy($id) {
 			Http::STATUS_FORBIDDEN
 		);
 	}
+
+	/**
+	 * Get available groups for assigning and removing via WebUI.
+	 *
+	 * @return DataResponse
+	 */
+	public function getAvailableGroups() {
```
jvillafanez, line 165: I'm not sure if there is a better naming for this method.... I'd expect a "classic" group list, not a double list of groups which are likely in both lists, and some of them are missing

settings/js/users/users.js
```diff
@@ -638,43 +640,70 @@ var UserList = {
...
 		function createItem(group) {
-			if (isSubadminSelect && group === 'admin') {
-				// can't become subadmin of "admin" group
+			if (isSubadminSelect) {
+				if (group === 'admin') {
+					// can't become subadmin of "admin" group
+					return;
+				}
+
+				$groupsSelect.append($('<option value="' + escapeHTML(group) + '">' + escapeHTML(group) + '</option>'));
```
jvillafanez, line 649: Maybe I'm missing something, but it seems the subadmin can bypass the restrictions. I don't think it's a good idea that the subadmin can have more power than the admin.

settings/js/users/users.js
```diff
@@ -638,43 +640,70 @@ var UserList = {
...
+				return;
+			}
+
+			var groupIsChecked = checked.indexOf(group) !== -1;
```
jvillafanez, line 653: I expect a Set object (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) to perform noticeable better.
We might need to traverse the 3 lists (checked, assignableGroups and removableGroups) completely, which seems expensive considering more than 1000 groups.

phil-davis:
> Technically speaking we now have a breaking change as I had to extend the IGroupManager interface. Although I did not find any OC apps implementing it... Let's wait for the pipeline first.

I can see lots of apps that reference it, for example https://github.com/owncloud/activity/blob/master/lib/FilesHooks.php#L82

But the change is an extension of the interface, and the object seems to be injected via the constructor. So the extended object should get passed in fine, and the apps should not complain that there is a bonus method.

After this is merged, the nightly CI of every app will run all the app unit tests etc, so we will know the next day if there is any app to change.

JammingBen:
> I can see lots of apps that reference it, for example https://github.com/owncloud/activity/blob/master/lib/FilesHooks.php#L82
>
> But the change is an extension of the interface, and the object seems to be injected via the constructor. So the extended object should get passed in fine, and the apps should not complain that there is a bonus method.

Right, but that's just an injection of the Group Manager which should be fine. Implementing the interface in another class (like the GroupManager here https://github.com/owncloud/core/blob/master/lib/private/Group/Manager.php#L61) would be a problem as it would also require the newly added method `getBackends`.

settings/js/users/users.js
JammingBen: As far as I can see, this is only used for the second group menu "Group Admin for" (=subadmin). Whereas a user cannot be assigned to the guest-group manually, the user can be assigned as its group admin. So we need to bypass the restriction here.

jvillafanez: Ok, got it. I'm not sure if we can make it more clear... I'm pretty sure I'll ask myself the same question again the next time I go through this code.

settings/js/users/users.js
```diff
@@ -638,43 +640,70 @@ var UserList = {
...
+				if (group === 'admin') {
+					// can't become subadmin of "admin" group
+					return;
+				}
+
+				$groupsSelect.append($('<option value="' + escapeHTML(group) + '">' + escapeHTML(group) + '</option>'));
+				return;
+			}
+
+			var groupIsChecked = new Set(checked).has(group) === true;
```
jvillafanez, line 653: Please double-check if we still support IE because it doesn't support this way of initialization.
In addition, the `=== true` isn't needed.

jvillafanez: Also, the idea I had is that the `checked` variable (the `assignableGroups` and `removableGroups` too) were a set already. I didn't want to keep traversing the whole list several times, but this solution will still do it in order to create the set.

JammingBen: Ooops yeah, you're totally right. Should create the sets before that loop.

Still, IE is an issue as OC supports IE11. Since I'm not a JS expert, I did some research on it. The proper way to convert an array to a set seems to be to loop through the array and call `set.add(el)`... not sure about this. I think it requires some testing which way is more efficient at the end.

jvillafanez: You're getting a list with the assignable and removable groups in https://github.com/owncloud/core/pull/38298/files#diff-89fda7202f65d0e3ad5878d84f67bb8d2f49da91cbe9efb1eebebf0228dc0e4cR672-R673 . There is where you should convert that list into a set. You'll have to traverse the list once to add the items to the set, but that's probably not different of what the "good" constructor would do it for you.

After that, that variable can be used as a set here, which should perform better.

settings/js/users/users.js
JammingBen: I at least added a comment for now

settings/js/users/users.js
JammingBen: Okay, I did some tests with ~200 groups. It's indeed faster that way. Not by a significant amount, bit still.

jvillafanez:
Ok, nothing more to add from me.
I think there is planned a 10.6.1 release, which means this will have to wait to be merged.

phil-davis:
> I think there is planned a 10.6.1 release, which means this will have to wait to be merged.

There is lots of other stuff already merged in master that will already cause a minor version bump to 10.7.0 - so IMO stuff like this can be merged when working. @micbar please confirm.

AlexAndBear (approved):

jnweiger:
Confirmed in 10.7.0RC1
* guest user cannot be removed from group guest_app, :heavy_check_mark: 
* admin user cannot be added to user guest_app. :heavy_check_mark:

PR #27345, opened 2017-03-09
Adding integration tests to checksums

SergioBertolinSG:
Requires https://github.com/owncloud/core/pull/27097

PR #27097, opened 2017-02-07
Server side checksum verification #26655

IljaN:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #22199, opened 2016-02-08
Make checksum propfind future proof

rullzer:
The checksum PR from https://github.com/owncloud/core/pull/21997 returns a propfind like:

``` xml
<oc:checksum>TYPE:CHECKSUM</oc:checksum>
```

Which if we will ever support multiple cheksums results in:

``` xml
<oc:checksum>TYPE1:CHECKSUM1,TYPE2:CHECKSUM2</oc:checksum>
```

Which is not very webdavvy.

This PR moves that to:

``` xml
<oc:checksums>
  <oc:checksum>TYPE1:CHECKSUM1</oc:checksum>
  <oc:checksum>TYPE2:CHECKSUM2</oc:checksum>
</oc:checksum>
```

Which is a bit more future proof.

Testing is basically the same as: https://github.com/owncloud/core/pull/21997#issuecomment-178602078 with the exception that your propfind file should contain:

``` xml
<?xml version="1.0"?>
<a:propfind xmlns:a="DAV:">
<a:prop xmlns:oc="http://owncloud.org/ns">
    <oc:checksums/>
</a:prop>
</a:propfind>
```

CC: @PVince81 @dragotin @MorrisJobke @nickvergessen

PR #21997, opened 2016-01-28
Checksums

rullzer:
For https://github.com/owncloud/core/issues/18716

First step at a checksums in file transfers.

~~The current modular approach allows for implementations of different checksum providers. The example app will just store them in the database and return them if requested. However other apps could be writen that actually validate the data etc.~~
- Extra column added to the filecache table.
- Updated methods to handle extra column
- dav stuff to get and return the checksum.

PR #38304, opened 2021-01-15
fix rendering of multiple checksums

butonic:
Hey OC10, long time no see...

I am implementing checksum support in ocis and wondered why oc10 is not rendiring multiple checksums as descriped in the original pr: https://github.com/owncloud/core/pull/22199
```xml
<oc:checksums>
  <oc:checksum>TYPE1:CHECKSUM1</oc:checksum>
  <oc:checksum>TYPE2:CHECKSUM2</oc:checksum>
</oc:checksum>
```

The root cause is that at the time there was only a single checksum and it was expected to be comma seperated as can be read in https://github.com/owncloud/core/pull/21997#issuecomment-181230285

When [multiple checksums got implemented](https://github.com/owncloud/core/commit/b9852612698b047b29371191831ef73f846bf88a) they were seperated by `' '` in the database. leading to this output, which was however untested:
```xml
<oc:checksums>
  <oc:checksum>TYPE1:CHECKSUM1 TYPE2:CHECKSUM2 TYPE3:CHECKSUM3</oc:checksum>
</oc:checksum>
```

😭 

https://github.com/owncloud/core/commit/b9852612698b047b29371191831ef73f846bf88a#diff-c76e89d5d2c586430906ac3c09dab6fb41a261ddc475febbcc6c79fd85872faeR25 would add tests but with multiple checksums in one row. This is part of https://github.com/owncloud/core/pull/27097 

but webdav only returned one checksum: https://github.com/owncloud/core/pull/27097/commits/05e11ce17a23937f38593508f7f6f8feeef89213

still digging for reasons where this went unnoticed ...

the current line in the tests was introduced in https://github.com/owncloud/core/commit/014530b29c19f4cf78f502ec07a7197eb0935170 which was part of https://github.com/owncloud/core/pull/27345 

This was the change that exposed all hashes via webdav: https://github.com/owncloud/core/pull/27345/files#diff-1d6ed4804160da213bc58248990a6936622ce53100a23535d207dba153f7e759

The PR also cemented the wrong rendering.

@dragotin @micbar @IljaN what do we do about this?

### DIFF
--- a/apps/dav/lib/Connector/Sabre/ChecksumList.php
+++ b/apps/dav/lib/Connector/Sabre/ChecksumList.php
@@ -39,7 +39,7 @@ class ChecksumList implements XmlSerializable {
 	 * @param string $checksum
 	 */
 	public function __construct($checksum) {
-		$this->checksums = \explode(',', $checksum);
+		$this->checksums = \explode(' ', $checksum);
 	}
 
 	/**


butonic:
@felix-schwarz It seems that the ios-sdk overwrites the item checksums with the ones from the last oc:checksum property in Line 196 of

https://github.com/owncloud/ios-sdk/blob/538ef354ccc460bc3cba66f1844313ba71f1bcb8/ownCloudSDK/Item/OCItem%2BOCXMLObjectCreation.m#L177-L199

butonic:
@michaelstingl the android client does not use checksums, yet: https://github.com/owncloud/android/issues/1840 right?

butonic:
@TheOneRing The desktop client parses a byte array `QByteArray findBestChecksum(const QByteArray &checksums)`:
https://github.com/owncloud/client/blob/3ef9835f8bed0ca36e67dd79abed10acba63f77d/src/common/checksums.cpp#L145-L160

It tries to grab the best hash until the next whitespace or the end of the line: https://github.com/owncloud/client/blob/3ef9835f8bed0ca36e67dd79abed10acba63f77d/src/common/checksums.cpp#L157

This should work when parsing the `OC-Checksum` header in https://github.com/owncloud/client/blob/ba2339a9865d9fd1e9400a673900f34c3c254d65/src/libsync/propagatedownload.cpp#L745

But when parsing the propfind response the code checks the `cheksums` (plural s) property and passes it in as a whole: https://github.com/owncloud/client/blob/8ecd87fed6a81eab46229d83a45aab481f781780/src/libsync/discoveryphase.cpp#L409-L410

Wouldn't that pass in all of `<oc:checksum>TYPE1:CHECKSUM1 TYPE2:CHECKSUM2 TYPE3:CHECKSUM3</oc:checksum>`?
So when that string contains only one hash, eg:   `<oc:checksum>sha1:h3xc0d3</oc:checksum>` `findBestChecksum` would return `sha1:h3xc0d3</oc:checksum>` ... this seems to be a bug, right?

michaelstingl:
> @michaelstingl the android client does not use checksums, yet: [owncloud/android#1840](https://github.com/owncloud/android/issues/1840) right?

yes, exactly

butonic:
To summarize:
- android does not use checksums
- ios does but would only keep the last `<oc:checksum>` element value of the `<oc:checksums>` tag.
- the desktop client would pick up the closing tag of `</oc:checksum>` as part of the checksum

I fear we need to stay bug compatible with old clients.

felix-schwarz:
> @felix-schwarz It seems that the ios-sdk overwrites the item checksums with the ones from the last oc:checksum property in Line 196 of
> 
> https://github.com/owncloud/ios-sdk/blob/538ef354ccc460bc3cba66f1844313ba71f1bcb8/ownCloudSDK/Item/OCItem%2BOCXMLObjectCreation.m#L177-L199
> 
> so ... fixing this would make ios clients not pick up the correct checksum, right?

The SDK won't overwrite previously parsed checksums, but re-use them and append new ones:

https://github.com/owncloud/ios-sdk/blob/538ef354ccc460bc3cba66f1844313ba71f1bcb8/ownCloudSDK/Item/OCItem%2BOCXMLObjectCreation.m#L184

```
NSMutableArray<OCChecksum *> *checksums = (item.checksums!=nil) ? // Does this item already have (a) checksum(s)?
	[[NSMutableArray alloc] initWithArray:item.checksums] :  // Yes -> Create a new array, but pre-populate with existing checksums
	[[NSMutableArray alloc] initWithCapacity:checksumStrings.count]; // No -> Create a new array and pre-allocate enough space in advance, so no reallocation is needed as items are added
…
// Parse checksums
{
	…
	// add parsed checksums to the local checksums array
	[checksums addObject:checksum];
}
…
item.checksums = checksums; // store the checksums into the item
```

dragotin:
Let's stay bug compatible. Do not merge this.

PR #38320, opened 2021-01-20
[tests-only] Fixed regex pattern used in download content matching

### LINKED ISSUE #38300: Test step for downloaded content is not strict enough
Example: this test scenario passes:
```
make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/apiProvisioning-v1/disableUser.feature:179
```

Remove the spaces from the expected content in the last step:
```
  Scenario: getting shares shared by disabled user
    Given the administrator has set the default folder for received shares to "Shares"
    And auto-accept shares has been disabled
    And user "Alice" has been created with default attributes and skeleton files
    And user "Brian" has been created with default attributes and without skeleton files
    And user "Alice" has shared file "/textfile0.txt" with user "Brian"
    And user "Brian" has accepted share "/textfile0.txt" offered by user "Alice"
    When the administrator disables user "Alice" using the provisioning API
    Then as "Brian" file "/Shares/textfile0.txt" should exist
    And the content of file "/Shares/textfile0.txt" for user "Brian" should be "ownCloudtesttextfile0"
```

The test still passes. But it should fail, because the actual content is not what the test expects.

https://github.com/owncloud/core/pull/38247/files#diff-730f2be358a1fc6a85ee5100526fb1bbcfc003547649bf9133ef13f4674a0dadR886 added some regex logic to the content comparison. But that seems to result in white-space differences being ignored.

Investigate. Maybe there are some test steps where we want to ignore some differences (maybe end-of-line characters, or space at the beginning or end or?) - but mostly we should be looking for an exact match.

Talank:
## Related Issue
- Fixes #38300 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -900,16 +900,15 @@ trait WebDav {
 	public function theDownloadedContentForMultipartByterangeShouldBe($statusCode, PyStringNode $content) {
 		$actualStatusCode = $this->response->getStatusCode();
 		if ($actualStatusCode === $statusCode) {
-			$actualContent = (string) $this->response->setBody();
+			$actualContent = (string) $this->response->getBody();
 			$pattern = ["/--\w*/", "/\s*/m"];
 			$actualContent = \preg_replace($pattern, "", $actualContent);
 			$content = \preg_replace("/\s*/m", '', $content->getRaw());
 			Assert::assertEquals(
 				$content,
 				$actualContent,
-				"The downloaded content was expected to be '$content', but actually is '$actualContent'. HTTP status was $actualStatus"
+				"The downloaded content was expected to be '$content', but actually is '$actualContent'. HTTP status was $actualStatusCode"
 			);
-			$this->downloadedContentShouldBe($content);
 		}
 	}
 
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -881,9 +881,8 @@ trait WebDav {
 		// But if the content is wrong (e.g. empty) then it is useful to
 		// report the HTTP status to give some clue what might be the problem.
 		$actualStatus = $this->response->getStatusCode();
-		$pattern = ["/--\w*/", "/\s*/m"];
+		$pattern = ["/\n\s*$(?!\n)/"];
 		$actualContent = \preg_replace($pattern, "", $actualContent);
-		$content = \preg_replace("/\s*/m", '', $content);
 		Assert::assertEquals(
 			$content,
 			$actualContent,
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -881,6 +881,7 @@ trait WebDav {
 		// But if the content is wrong (e.g. empty) then it is useful to
 		// report the HTTP status to give some clue what might be the problem.
 		$actualStatus = $this->response->getStatusCode();
+		$actualContent = \preg_replace("/\n/", "", $actualContent);
 		Assert::assertEquals(
 			$content,
 			$actualContent,
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -881,7 +881,6 @@ trait WebDav {
 		// But if the content is wrong (e.g. empty) then it is useful to
 		// report the HTTP status to give some clue what might be the problem.
 		$actualStatus = $this->response->getStatusCode();
-		$actualContent = \preg_replace("/\n/", "", $actualContent);
 		Assert::assertEquals(
 			$content,
 			$actualContent,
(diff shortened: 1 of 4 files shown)


phil-davis:
See my comment https://github.com/owncloud/core/issues/38300#issuecomment-764394701

IMO it will be easier and clearer code if we remove the 3 lines of special `preg_replace` stuff from `downloadedContentShouldBe` and make a separate step that does a not-exact "special" match of the content. Maybe the "special" code is only needed for the new step "the downloaded content for multipart byterange should be" `theDownloadedContentForMultipartByterangeShouldBe` and so the special `preg_replace` code can be put in there.

saw-jan:
@Talank also if it is to remove extra space before and after the text content,  `trim()` will do the job well.

Talank:
> @Talank also if it is to remove extra space before and after the text content, `trim()` will do the job well.

it's more about preserving the spaces between words, The trimming is done perfectly by the special `preg_replace` stuff

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -903,7 +900,16 @@ public function downloadedContentShouldBe($content) {
 	public function theDownloadedContentForMultipartByterangeShouldBe($statusCode, PyStringNode $content) {
 		$actualStatusCode = $this->response->getStatusCode();
 		if ($actualStatusCode === $statusCode) {
-			$this->downloadedContentShouldBe($content->getRaw());
+			$actualContent = (string) $this->response->setBody();
```
phil-davis, line 903:
```suggestion
			$actualContent = (string) $this->response->getBody();
```

This "looks wrong"

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -903,7 +900,16 @@ public function downloadedContentShouldBe($content) {
...
+			$pattern = ["/--\w*/", "/\s*/m"];
+			$actualContent = \preg_replace($pattern, "", $actualContent);
+			$content = \preg_replace("/\s*/m", '', $content->getRaw());
+			Assert::assertEquals(
+				$content,
+				$actualContent,
+				"The downloaded content was expected to be '$content', but actually is '$actualContent'. HTTP status was $actualStatus"
```
phil-davis, line 910:
```suggestion
				"The downloaded content was expected to be '$content', but actually is '$actualContent'. HTTP status was $actualStatusCode"
```

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -903,7 +900,16 @@ public function downloadedContentShouldBe($content) {
...
+			);
+			$this->downloadedContentShouldBe($content);
```
phil-davis, line 912: Why do we call this? I thought that the content was already compared in the `Assert::assertEquals` call above.

phil-davis (dismissed):

phil-davis:
https://drone.owncloud.com/owncloud/core/28384/57/13
Some scenarios are failing, for example:
```
  Scenario: getting shares shared by disabled user                                                         # /drone/src/tests/acceptance/features/apiProvisioning-v1/disableUser.feature:179
    Given the administrator has set the default folder for received shares to "Shares"                     # OccContext::theAdministratorHasSetTheDefaultFolderForReceivedSharesTo()
    And auto-accept shares has been disabled                                                               # FeatureContext::autoAcceptSharesHasBeenDisabled()
    And user "Alice" has been created with default attributes and skeleton files                           # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And user "Brian" has been created with default attributes and without skeleton files                   # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "Alice" has shared file "/textfile0.txt" with user "Brian"                                    # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Brian" has accepted share "/textfile0.txt" offered by user "Alice"                           # FeatureContext::userHasReactedToShareOfferedBy()
    When the administrator disables user "Alice" using the provisioning API                                # FeatureContext::adminDisablesUserUsingTheProvisioningApi()
    Then as "Brian" file "/Shares/textfile0.txt" should exist                                              # FeatureContext::asFileOrFolderShouldExist()
    And the content of file "/Shares/textfile0.txt" for user "Brian" should be "ownCloud test text file 0" # FeatureContext::contentOfFileForUserShouldBe()
      The downloaded content was expected to be 'ownCloud test text file 0', but actually is 'ownCloud test text file 0
      '. HTTP status was 200
      --- Expected
      +++ Actual
      @@ @@
      -'ownCloud test text file 0'
      +'ownCloud test text file 0\n
      +'
```

There is a step `contentOfFileForUserShouldBePlusEndOfLine` that should work for those scenarios - `textfile0.txt` has an end-of-line character.

Also the similar fails in https://drone.owncloud.com/owncloud/core/28384/58/13

Talank:
@phil-davis I think the idea of completely removing the `preg_replace()` from `downloadedContentShouldBe` is not that accurate. I will add a `preg_replace()` to filter the `\n`

phil-davis (changes requested):
We have a step already to specifically check for the end-of-line character. We want the tests to check the exact file content (we do not want the test to still pass if the downloaded content has an extra `\n` or is missing the `n`.

We can just use the step.

phil-davis (approved):

PR #38329, opened 2021-01-22
[tests-only] improve user-creation error

individual-it:
## Description
improve the error message when a user could not be created in the test run
1. a space was missing
2. when multiple users are created in a table its not clear which one failed

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis (approved):
This is a good improvement. The existing code looks like it will end up throwing the exception for the last user that had a problem. An other `new Exception` will be overwritten by the last one.

We could build a longer-and-longer exception message in the loop, then create the exception at the end with a long message that tells about every user that failed to be created.

PR #38364, opened 2021-02-01
[tests-only] Add replaceUsernames option to .drone.star

### LINKED ISSUE #33596: Add acceptance tests for more variety of user Ids

phil-davis:
## Description
To make it easier for developers and QA to run the acceptance tests in CI with different user names and attributes.

Documented in PR https://github.com/owncloud/docs/pull/3091

2nd commit - Refactor provisioning tests to allow REPLACE_USERNAMES - a lot of the provisioning test code did not abstract-out the usernames used in the test steps. That was noticed in the demo PR #38365 which should pass after this commit is added.

## Related Issue
- Fixes #33596

## How Has This Been Tested?
CI - see demo PR #38365 

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1335,6 +1335,7 @@ def acceptance(ctx):
 		'scalityS3': False,
 		'testingRemoteSystem': True,
 		'useHttps': True,
+		'replaceUsernames': False,
 		'extraSetup': [],
 		'extraServices': [],
 		'extraEnvironment': {},
@@ -1432,6 +1433,7 @@ def acceptance(ctx):
 								environment['TEST_SERVER_URL'] = '%s://%s' % (protocol, serverUnderTest)
 
 								environment['BEHAT_FILTER_TAGS'] = params['filterTags']
+								environment['REPLACE_USERNAMES'] = params['replaceUsernames']
 
 								if (params['runAllSuites'] == False):
 									environment['BEHAT_SUITE'] = suite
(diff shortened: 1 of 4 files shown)


phil-davis:
PR #38365 has passed. That demonstrates that this code is working.

individual-it (approved):

PR #38374, opened 2021-02-03
[Tests-Only]make escaping/encoding case insensitive

### LINKED ISSUE #38367: Make tests of escaped file and folder names more flexible.
See https://github.com/cs3org/reva/pull/1425

There are tests that check the `d:href` string that is returned in `PROPFIND` or other API requests. If the file path contains "special characters" then those have been escaped. For example, a space character is escaped as `%20`, a question mark is escaped as `%3F` or `%3f`

https://tools.ietf.org/html/rfc3986#section-2.1 specifies that this escaping/encoding is not case-sensitive. `%3F` and `%3f` are both valid representations of `?`.

Find the places in the core API test suite where the tests are currently "too fussy". Make those tests case-insensitive.

This will help when testing against other implementations like OCIS.

swoichha:
## Description
This PR makes tests of escaped file and folder names with special characters more flexible.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/38367

## How Has This Been Tested?
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavProperties2/getFileProperties.feature
+++ b/tests/acceptance/features/apiWebdavProperties2/getFileProperties.feature
@@ -33,15 +33,15 @@ Feature: get file properties
     Then the properties response should contain an etag
     And the value of the item "//d:response/d:href" in the response to user "Alice" should match "/remote\.php\/<expected_href>/"
     Examples:
-      | dav_version | file_name     | expected_href                               |
-      | old         | /C++ file.cpp | webdav\/C%2b%2b%20file\.cpp                 |
-      | old         | /file #2.txt  | webdav\/file%20%232\.txt                    |
-      | old         | /file ?2.txt  | webdav\/file%20%3f2\.txt                    |
-      | old         | /file &2.txt  | webdav\/file%20%262\.txt                    |
-      | new         | /C++ file.cpp | dav\/files\/%username%\/C%2b%2b%20file\.cpp |
-      | new         | /file #2.txt  | dav\/files\/%username%\/file%20%232\.txt    |
-      | new         | /file ?2.txt  | dav\/files\/%username%\/file%20%3f2\.txt    |
-      | new         | /file &2.txt  | dav\/files\/%username%\/file%20%262\.txt    |
+      | dav_version | file_name     | expected_href                                     |
+      | old         | /C++ file.cpp | webdav\/C%2[bB]%2[bB]%20file\.cpp                 |
+      | old         | /file #2.txt  | webdav\/file%20%232\.txt                          |
+      | old         | /file ?2.txt  | webdav\/file%20%3[fF]2\.txt                       |
+      | old         | /file &2.txt  | webdav\/file%20%262\.txt                          |
+      | new         | /C++ file.cpp | dav\/files\/%username%\/C%2[bB]%2[bB]%20file\.cpp |
+      | new         | /file #2.txt  | dav\/files\/%username%\/file%20%232\.txt          |
+      | new         | /file ?2.txt  | dav\/files\/%username%\/file%20%3[fF]2\.txt       |
+      | new         | /file &2.txt  | dav\/files\/%username%\/file%20%262\.txt          |
 
   @issue-ocis-reva-214
   Scenario Outline: Do a PROPFIND of various folder names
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/apiMain/checksums.feature
```diff
@@ -331,7 +331,6 @@ Feature: checksums
     Then the HTTP status code should be "400"
     And the content of file "/textfile0.txt" for user "Brian" should be "ownCloud test text file 0" plus end-of-line
 
-  @issue-ocis-reva-214
```
dpakach, line 334: do not delete this line this changes line number of other tests

tests/acceptance/features/bootstrap/WebDavPropertiesContext.php
```diff
@@ -679,10 +679,8 @@ public function assertValueOfItemInResponseToUserRegExp($xpath, $user, $pattern)
 			isset($xmlPart[0]), "Cannot find item with xpath \"$xpath\""
 		);
 		$user = $this->featureContext->getActualUsername($user);
-		$value = $xmlPart[0]->__toString();
-		$pattern = $this->featureContext->substituteInLineCodes(
-			$pattern, $user, ['preg_quote' => ['/']]
-		);
+		$value = \strtolower($xmlPart[0]->__toString());
+		$pattern = \strtolower($this->featureContext->substituteInLineCodes($pattern, $user, ['preg_quote' => ['/']]));
```
dpakach, line 683: I think we only want to do this for some properties such as `d:href`, you can do this only for such properties otherwise do case sensitive comparisioin

dpakach (commented):
LGTM :+1:

individual-it (commented):
I'm not sure if I'm happy with that, because now we are silently changing the regex and what the codes checks is not what the feature files says.
What about changing the regular expression in the places where its an issue?

swoichha:
> I'm not sure if I'm happy with that, because now we are silently changing the regex and what the codes checks is not what the feature files says.
> What about changing the regular expression in the places where its an issue?

@individual-it It was a problem for https://github.com/owncloud/core/blob/b5eadc5c9fcb1c2e5b1f697633c32fea36e7e5f8/tests/acceptance/features/apiWebdavProperties2/getFileProperties.feature#L39
So, this only changes the regex if we are checking response for `/d:href` like `//d:response/d:href` in this case. But for all the other cases it does not change the regex into lowercase.

individual-it:
@swoichha what about changing the regex itself to something like `webdav\/file%20%3[fF]2\.txt` or adding the case-insensitive modifier `i` in line 34  https://github.com/owncloud/core/blob/b5eadc5c9fcb1c2e5b1f697633c32fea36e7e5f8/tests/acceptance/features/apiWebdavProperties2/getFileProperties.feature#L34
`/webdav\/file%20%3f2\.txt/i`

swoichha:
> @swoichha what about changing the regex itself to something like `webdav\/file%20%3[fF]2\.txt` or adding the case-insensitive modifier `i` in line 34
> 
> https://github.com/owncloud/core/blob/b5eadc5c9fcb1c2e5b1f697633c32fea36e7e5f8/tests/acceptance/features/apiWebdavProperties2/getFileProperties.feature#L34
> 
> 
> `/webdav\/file%20%3f2\.txt/i`

@individual-it Yes, we can do that and it will work but in future we must have to remember doing  the same for encoding case-insensitive characters.

individual-it:
the quoted RFC is for all URIs https://tools.ietf.org/html/rfc3986#section-2.1 so we would have to make the substitution cleverer to match all percent encoding but then we have to be careful not to change if we want to check real `%` in any item-name

I think the original solution would also match `wEBdav\/FiLe%20%3f2\.tXT`

individual-it:
after verbal discussion with @swoichha 

`/i` will also match too much because if the response comes back with the filename with changed cases the test would not catch the bug if we make the regex completely case insensitive
checking the decoded filename is problematic because that would not catch the original bug reported in https://github.com/owncloud/ocis/issues/1296
meaning:
1. we have to check the raw value
2. we have to be explicit what the step does and it should not manipulate the regex or values underneath

so the only real solution is to refactor all regex to be in the form of `file%20%3[fF]2`

individual-it (approved):

PR #38376, opened 2021-02-03
pre-signed download urls for password protected public links

C0rby:
## Description
When a propfind request is done for a password protected public link share and the property `'{http://owncloud.org/ns}downloadURL'` is requested, then a signed url is created and sent to the client as the value of that property.
The downloadURL property was added some time ago but haven't been used since because the demand disappeared. So I think using this property should be fine. 

## Motivation and Context

To support clients which don't use cookies I implemented pre-signed urls
for password protected public links. The share password is used as the
signing key and the signed url is then added to the propfind response in
the field `downloadURL` which was added before but never used. This
change allows owncloud Web to implement a more efficient download
mechanism for password protected link shares.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
+++ b/apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
@@ -147,6 +147,57 @@ class PublicFilesPlugin extends ServerPlugin {
 			$propFind->handle(FilesPlugin::SIZE_PROPERTYNAME, static function () use ($node) {
 				return $node->getNode()->getSize();
 			});
+			if ($node->getNode()->getType() === FileInfo::TYPE_FILE) {
+				$server = $this->server;
+				$propFind->handle(FilesPlugin::DOWNLOADURL_PROPERTYNAME, static function () use ($node, $server) {
+					$share = $node->getShare();
+					$shareNode = $share->getNode();
+					// We want to get the relative path of the shared file.
+					// If the shared resource is a folder e.g.
+					// - <shared folder>/
+					//   - subfolder/
+					//     - meme.jpg
+					//   - somefile.txt
+					// And we want the path of 'meme.jpg' we expect the resource
+					// path to be '/subfolder/meme.jpg'.
+					// If the shared resource is a file we can just take the
+					// name of that file prefixed with a slash like '/cool.gif'
+					if ($shareNode->getType() === FileInfo::TYPE_FOLDER) {
+						$shareRoot = $shareNode->getPath();
+						$sharedResourcePath = \substr($node->getNode()->getPath(), \strlen($shareRoot));
+					} else {
+						$sharedResourcePath = '/' . $shareNode->getName();
+					}
+
+					$path = $server->getBaseUri() . $server->getRequestUri();
+					// Let's assume we have this share
+					// - <shared folder>/
+					//   - subfolder/
+					//     - meme.jpg
+					//   - somefile.txt
+					// If the PROPFIND request is done against
+					// 'remote.php/dav/public-files/{token}/subfolder/meme.jpg'
+					// then we don't need to append the file name for response as
+					// it is already in the path.
+					// Otherwise if the PROPFIND is done against
+					// 'remote.php/dav/public-files/{token}/subfolder/'
+					// then we need to append the file name to the path.
+					if (\substr($path, -\strlen($sharedResourcePath)) !== $sharedResourcePath) {
+						$path .= '/' . $node->getNode()->getName();
+					}
+
+					if ($share->getPassword() === null) {
+						return $path;
+					}
+
+					$validUntil = new \DateTime();
+					$validUntil->add(new \DateInterval("PT30M")); // valid for 30 minutes
+					$key = \hash_hkdf('sha256', $share->getPassword());
+
+					$s = new PublicShareSigner($share->getToken(), $sharedResourcePath, $validUntil, $key);
+					return $path . '?signature=' . $s->getSignature() . '&expires=' . \urlencode($validUntil->format(\DateTime::ATOM));
+				});
+			}
 		}
 	}
 
--- a/apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php
+++ b/apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php
@@ -101,6 +101,25 @@ class PublicSharingAuth extends AbstractBasic {
 			return [true, 'principals/system/public'];
 		}
 
+		// Clients which don't use cookie based session authentication and want
+		// to use anchor tags `<a href...` to download password protected files
+		// can't add the basic authentication header.
+		// They can use pre-signed urls instead.
+		$query = $request->getQueryParameters();
+		if (isset($query['signature'], $query['expires'])) {
+			$sig = $query['signature'];
+			$validUntil = \DateTime::createFromFormat(\DateTime::ATOM, $query['expires']);
+			$now = new \DateTime();
+			if ($now < $validUntil) {
+				$key = \hash_hkdf('sha256', $this->share->getPassword());
+				$resource_path = \explode($this->share->getToken(), $request->getPath())[1];
+				$s = new PublicShareSigner($this->share->getToken(), $resource_path, $validUntil, $key);
+				if (\hash_equals($s->getSignature(), $sig)) {
+					return [true, 'principals/system/public'];
+				}
+			}
+		}
+
 		try {
 			return parent::check($request, $response);
 		} catch (LoginException $e) {
--- a/apps/dav/lib/Files/PublicFiles/PublicShareSigner.php
+++ b/apps/dav/lib/Files/PublicFiles/PublicShareSigner.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * @author David Christofas <dchristofas@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\DAV\Files\PublicFiles;
+
+class PublicShareSigner {
+	private $token;
+	private $fileName;
+	private $validUntil;
+	private $signingKey;
+
+	public function __construct(String $token, String $fileName, \DateTime $validUntil, String $signingKey) {
+		$this->token = $token;
+		$this->fileName = $fileName;
+		$this->validUntil = $validUntil->format(\DateTime::ATOM);
+		$this->signingKey = $signingKey;
+	}
+
+	public function getSignature() {
+		return \hash_hmac('sha512/256', \implode('|', [$this->token, $this->fileName, $this->validUntil]), $this->signingKey, false);
+	}
+}
--- a/apps/dav/tests/unit/Files/PublicFiles/PublicShareSignerTest.php
+++ b/apps/dav/tests/unit/Files/PublicFiles/PublicShareSignerTest.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * @author David Christofas <dchristofas@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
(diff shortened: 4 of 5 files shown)


micbar:
@C0rby Can you add a changelog please?

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -148,6 +148,25 @@ public function propFind(PropFind $propFind, INode $node) {
...
+				$path .= '/' . $nodeName;
+			}
+
+			$share = $node->getShare();
+			if ($share->getPassword() === null) {
+				return $path;
+			}
+
+			$validUntil = new \DateTime();
+			$validUntil->add(new \DateInterval("PT1H")); // valid for 1 hour
```
jvillafanez, line 165: A couple of questions here:
* Is it worthy to make it configurable? I'm not sure how this is expected to work with unreliable connections, but I guess for the normal use case you'd request the pre-signed download just before downloading the file. 15-30 minutes should be plenty of time to make the pre-signed download request.
  In case of unstable network connection or when the download cuts off in the middle, I guess you want to retry the download. If the network is slow, 1 hour might not be enough to retry the download without requesting a new pre-signed url.
* In case the network is slow and the file is too big, it's very likely that the file takes more than 1 hour to download. I guess this isn't a problem for stable networks (ownCloud won't actively cut the connection after 1 hour), but if the network is cut after that hour, what is the expected action plan? How do we expect the user to download the rest of the file?

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -148,6 +148,25 @@ public function propFind(PropFind $propFind, INode $node) {
...
+			$key = \hash_hkdf('sha256', $share->getPassword());
+			$s = new PublicShareSignature($share->getToken(), $node->getName(), $validUntil, $key);
+			return  $path . '?signature=' . $s->get() . '&expires=' . \urlencode($validUntil->format(\DateTime::ATOM));
```
jvillafanez, line 168: being picky here, there is double space between `return` and `$path`, plus string interpolation is preferred instead of concatenation. (just in case you touch this file)

apps/dav/lib/Files/PublicFiles/PublicShareSignature.php
```diff
@@ -0,0 +1,45 @@
...
+		$this->fileName = $fileName;
+		$this->validUntil = $validUntil->format(\DateTime::ATOM);
+		$this->signingKey = $signingKey;
+	}
+
+	public function get() {
+		return \hash_hmac('sha512/256', \implode('|', [$this->token, $this->fileName, $this->validUntil]), $this->signingKey, false);
+	}
+
+	public function verify($signature) {
```
jvillafanez, line 41: Doesn't seem to add much value... I mean, you could just call the "get" method and compare it outside. Any future plan to make this verification more complex? That could justify this method.

apps/dav/lib/Files/PublicFiles/PublicShareSignature.php
```diff
@@ -0,0 +1,45 @@
...
+	private $signingKey;
+
+	public function __construct(String $token, String $fileName, \DateTime $validUntil, String $signingKey) {
+		$this->token = $token;
+		$this->fileName = $fileName;
+		$this->validUntil = $validUntil->format(\DateTime::ATOM);
+		$this->signingKey = $signingKey;
+	}
+
+	public function get() {
```
jvillafanez, line 37: In general, I prefer to avoid too generic method names. It will be more difficult to find where this method is being used. `getSignature` or `getShareSignature` will be easier to find

apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php
```diff
@@ -101,6 +101,24 @@ public function check(RequestInterface $request, ResponseInterface $response) {
...
 		}
 
+		// Clients which don't use cookie based session authentication and want
+		// to use anchor tags `<a href...` to download password protected files
+		// can't add the basic authentication header.
+		// They can use pre-signed urls instead.
+		$query = $request->getQueryParameters();
+		if (isset($query['signature'], $query['expires'])) {
+			$sig = $query['signature'];
+			$validUntil = \DateTime::createFromFormat(\DateTime::ATOM, $query['expires']);
```
jvillafanez, line 111: Couple of questions here:
* From the user's perspective, having to send the "expires" parameter doesn't look good. Any alternative not to require the "expires" parameter?
  If it isn't possible, concatenating both "signature" and "expires" parameters could be an easy solution. This way, the server can send a "code" parameter to the client (just to make the idea more clear) with "signature" + "expires", and then the client can send the same "code" as if it was a password. The server can then split that "code" parameter into the "signature" and "expires".
* What happens if the "expires" parameter is random / invalid?. If this is handled, include a comment stating what will happen.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
C0rby: > 15-30 minutes should be plenty of time to make the pre-signed download request.

Yeah, actually you are right. I'm fine with lowering the expiration time.
> Is it worthy to make it configurable?

I think for this special case it doesn't need to be configurable since as you already said one would get the signed url and then immediately use it. So 15-30 minutes should be enough.
Once the download started the signed url isn't 'needed' anymore unless the connection drops.

> In case the network is slow and the file is too big, it's very likely that the file takes more than 1 hour to download. I guess this isn't a problem for stable networks (ownCloud won't actively cut the connection after 1 hour), but if the network is cut after that hour, what is the expected action plan? How do we expect the user to download the rest of the file?

Yes, right now the user would have to reload the public share page (or the frontend is doing that without the user noticing by issuing a new propfind request) and then initiate the download again. This would probably not reuse the already downloaded data from the browser but that needs to be checked.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
C0rby: Nice catch! :+1: Will fix

apps/dav/lib/Files/PublicFiles/PublicShareSignature.php
C0rby: Yes you are right. The idea behind it was just to have a nice API.

```php
$signature = new PublicShareSignature($share->getToken(), $node->getName(), $validUntil, $key);
if ($signature->verify($urlSignature)) {
...
```
vs.
```php
$signature = new PublicShareSignature($share->getToken(), $node->getName(), $validUntil, $key);
if ($signature->get() == $urlSignature) {
...
```
It doesn't add much value and won't get any more complex in the future so I'm good with deleting the method and comparing the hash outside.

apps/dav/lib/Files/PublicFiles/PublicShareSignature.php
C0rby: I agree with it potentially being hard to find.
Though if we change the name from `get` to `getSignature` then I would change the class as well.
Currently the usage looks like:
```php
$signature = new PublicShareSignature($share->getToken(), $node->getName(), $validUntil, $key);
$signature->get();
```
By changing the method to `getSignature` it would look wrong.
```php
$signature = new PublicShareSignature($share->getToken(), $node->getName(), $validUntil, $key);
$signature->getSignature();
```
So how do you find this:
```php
$signer = new PublicShareSigner($share->getToken(), $node->getName(), $validUntil, $key);
$signer->createSignature();
```

apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php
C0rby: > What happens if the "expires" parameter is random / invalid?. If this is handled, include a comment stating what will happen.

Right, I should probably add exception handling to the date parsing.

> From the user's perspective, having to send the "expires" parameter doesn't look good. 

Why is that?


I guess we could concatenate the fields and maybe base64 encode it. Then it isn't obvious to most people.
Also this would make it easier for clients since they don't need to url encode the expires parameter.

apps/dav/lib/Files/PublicFiles/PublicShareSignature.php
jvillafanez: Ok, just remember that `===` is always preferred. If you use `==`, include a comment explaining the reason in order to prevent changing it later.

apps/dav/lib/Files/PublicFiles/PublicShareSignature.php
jvillafanez: "create" doesn't necessarily imply returning a value. You could create the signature, keep it inside the class, and return "true" or "false" to indicate if the signature was created successfully.
Anyway, I don't want to overcomplicate this, so I'm fine if you want to keep the "createSignature" name as long as it's properly documented and explained in the PHPDoc.

> By changing the method to getSignature it would look wrong.

I partially agree. It looks wrong because that's the only method the object has. However, it doesn't look bad if you have a signature object with multiple methods, being one of them a method to return the raw or stringified signature.

apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php
jvillafanez: From the user's perspective, the "signature" parameter acts as a temporary password to access to the link. This is opaque for the user.
The "expires" parameter, however, seems to be for informational purposes: you need to request the download before that time. In this sense, why would the user need to send back the "expires" parameter? Just sending the signature (as a password) should be enough.

As said, that's for the user's perspective. I know the "expires" parameter is required. By concatenating both parameters the server can say "if you want to download this file, include this signature".

Anyway, I'm not going to push for this change taking into account that the expected flow is that the user copy and paste the download url as it is sent by the server.

C0rby:
I converted the PR to a draft because I noticed a bug when sharing a folder.
The signature check is not working correctly because when checking the signature we use the shared resource name (In case of a folder, the folder name) not the filename.

C0rby:
So folder shares work now.

apps/dav/lib/Files/PublicFiles/PublicSharingAuth.php
C0rby: I left it as it is because the user is not even expected to copy that URL but just click a link in the web ui.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -147,6 +147,35 @@ public function propFind(PropFind $propFind, INode $node) {
...
+				$share = $node->getShare();
+				if ($share->getPassword() === null) {
+					return $path;
+				}
+
+				$validUntil = new \DateTime();
+				$validUntil->add(new \DateInterval("PT30M")); // valid for 30 minutes
+				$key = \hash_hkdf('sha256', $share->getPassword());
+
+				if ($share->getNode()->getType() === FileInfo::TYPE_FOLDER) {
```
jvillafanez, line 170: got confused with the initial `if ($node->getNode()->getType() === FileInfo::TYPE_FOLDER) {` since the code is almost the same. Maybe a comment could highlight the difference, or maybe use `$shareNode = $share->getNode()`

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -147,6 +147,35 @@ public function propFind(PropFind $propFind, INode $node) {
...
+					$shareRoot = $share->getNode()->getPath();
+					$shared_resource_path = \substr($node->getNode()->getPath(), \strlen($shareRoot));
+				} else {
+					$shared_resource_path = '/' . $share->getNode()->getName();
```
jvillafanez, line 174: Use camelcase names for consistency with the rest of the code.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -147,6 +147,35 @@ public function propFind(PropFind $propFind, INode $node) {
...
+				}
+				$s = new PublicShareSigner($share->getToken(), $shared_resource_path, $validUntil, $key);
+				return $path . '?signature=' . $s->getSignature() . '&expires=' . \urlencode($validUntil->format(\DateTime::ATOM));
```
jvillafanez, line 177: I assume the date format is consistent with the rest of the code. If not, there are 2 options: either change it for consistency or include a comment explaining.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -147,6 +147,35 @@ public function propFind(PropFind $propFind, INode $node) {
 			$propFind->handle(FilesPlugin::SIZE_PROPERTYNAME, static function () use ($node) {
 				return $node->getNode()->getSize();
 			});
+			$server = $this->server;
+			$propFind->handle(FilesPlugin::DOWNLOADURL_PROPERTYNAME, static function () use ($node, $server) {
+				if ($node->getNode()->getType() === FileInfo::TYPE_FOLDER) {
+					return "";
```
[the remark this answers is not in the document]
C0rby: Yes, for folders it should be missing.
It's fixed now.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
C0rby: Well, there is no 'one' date format. In the code I found usages of `ATOM`, `ISO8601`, `RFC2822`.
I chose `ATOM` because of this: https://www.php.net/manual/en/class.datetimeinterface.php#datetime.constants.iso8601

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -147,6 +147,35 @@ public function propFind(PropFind $propFind, INode $node) {
...
+				}
+				$path = $server->getBaseUri() . $server->getRequestUri();
+				$nodeName = $node->getNode()->getName();
+				if (\substr($path, -\strlen($nodeName)) !== $nodeName) {
```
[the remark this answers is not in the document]
C0rby: I rewrote the check and added comments to clarify the reasons.

jvillafanez (approved):

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
jvillafanez: Yes, we should have only one, but it seems we don't... at least we can change this one later without problems.

PR #38384, opened 2021-02-09
[Tests-Only] Added api test for getting the version number after moving a file

Talank:
## Description
Added api test for getting the version number after moving a file

## Related Issue
- Demonstrates https://github.com/owncloud/ocis/issues/1238

## How Has This Been Tested?
- https://github.com/owncloud/ocis/pull/1613

## Types of changes

- [ ] Technical debt
- [z] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiVersions/fileVersions.feature
+++ b/tests/acceptance/features/apiVersions/fileVersions.feature
@@ -403,3 +403,15 @@ Feature: dav-versions
     When user "Alice" restores version index "1" of file "/renamedfile.txt" using the WebDAV API
     Then the HTTP status code should be "201"
     And the content of file "/renamedfile.txt" for user "Alice" should be "old content"
+    
+  @issue-ocis-1238
+  Scenario: The version number is wrong after moving a file
+    Given user "Alice" has created folder "testFolder"
+    And user "Alice" has uploaded file with content "uploaded content" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 1" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 2" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 3" to "textfile0.txt"
+    When user "Alice" moves file "textfile0.txt" to "/testFolder/textfile0.txt" using the WebDAV API
+    And user "Alice" tries to get versions of file "/testFolder/textfile0.txt" from "Alice"
+    Then the HTTP status code should be "207"
+    And the number of versions should be "3"
--- a/tests/acceptance/features/apiVersions/fileVersions.feature
+++ b/tests/acceptance/features/apiVersions/fileVersions.feature
@@ -405,13 +405,18 @@ Feature: dav-versions
     And the content of file "/renamedfile.txt" for user "Alice" should be "old content"
     
   @issue-ocis-1238
-  Scenario: The version number is wrong after moving a file
+  Scenario: User can access version number after moving a file
     Given user "Alice" has created folder "testFolder"
     And user "Alice" has uploaded file with content "uploaded content" to "textfile0.txt"
     And user "Alice" has uploaded file with content "version 1" to "textfile0.txt"
     And user "Alice" has uploaded file with content "version 2" to "textfile0.txt"
     And user "Alice" has uploaded file with content "version 3" to "textfile0.txt"
     When user "Alice" moves file "textfile0.txt" to "/testFolder/textfile0.txt" using the WebDAV API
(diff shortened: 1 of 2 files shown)


tests/acceptance/features/apiVersions/fileVersions.feature
```diff
@@ -394,4 +394,17 @@ Feature: dav-versions
...
+  Scenario: The version number is reset after moving a file
+    Given user "Alice" has created folder "newFolder1"
+    And user "Alice" has uploaded file with content "uploaded content" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 1" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 2" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 3" to "textfile0.txt"
+    When user "Alice" moves file "textfile0.txt" to "/newFolder1/textfile0.txt" using the WebDAV API
+    And user "Alice" tries to get versions of file "/newFolder1/textfile0.txt" from "Alice"
+    Then the HTTP status code should be "207"
+    And the number of versions should be "0"
```
saw-jan, line 409: Should we not skip on `oCV10` and make it pass on it (uncommenting the comment), and add this test in expected_failures later in oCIS.

Also, I get versions `4` instead of `0`while running on oCIS

tests/acceptance/features/apiVersions/fileVersions.feature
```diff
@@ -394,4 +394,17 @@ Feature: dav-versions
...
+#    And the number of versions should be "3"
```
saw-jan, line 410: Should we not skip on `oCV10` and make it pass on it (uncommenting the comment), and add this test in expected_failures later in oCIS.

Also, I get versions `4` instead of `0`while running on oCIS

Talank: Yes, the version is 4, it should have been 3. Thanks.

swoichha: I agree with @saw-jan. You haven't remove the tag `skipOnOcV10` yet @Talank .

swoichha (commented):
I agree with @saw-jan. You haven't remove the tag skipOnOcV10 yet @Talank .

tests/acceptance/features/apiVersions/fileVersions.feature
```diff
@@ -394,4 +394,16 @@ Feature: dav-versions
...
+
+  @issue-ocis-1238
+  Scenario: The version number is wrong after moving a file
+    Given user "Alice" has created folder "testFolder"
+    And user "Alice" has uploaded file with content "uploaded content" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 1" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 2" to "textfile0.txt"
+    And user "Alice" has uploaded file with content "version 3" to "textfile0.txt"
+    When user "Alice" moves file "textfile0.txt" to "/testFolder/textfile0.txt" using the WebDAV API
+    And user "Alice" tries to get versions of file "/testFolder/textfile0.txt" from "Alice"
```
phil-davis, line 407: This step is a bit odd. It is useful for when a user tries to get the versions of a file belonging to another user, and they do not have privilege to do it.

We should make an "ordinary" step that does:
```
When user "Alice" gets the number of versions of file "/testFolder/textfile0.txt"
```

and for the existing scenarios that use the "tries to" step, if they actually succeed then make a step without the words "tries to":
```
When user "Brian" gets the number of versions of file "textfile0.txt" from "Alice"
```

And make various scenarios that check when there are 0 versions (just the original file) and more versions, and when Brian does not have access to the file versions.

I am surprised that we do not already have these type of scenarios. But I can't find them!

phil-davis: Make scenarios that pass on oC10. Then if they fail on OCIS we can add them to expected-failures and discuss what is the correct version count (oC10 or OCIS)

Talank: @phil-davis we already have a scenario where a user tries to get the version of a file that is not accessible to that user. The test is in the scenario `Receiver tries to get file versions of unshared file from the sharer`.
However, I did not find the scenario which checks when the version number is 0. So, I added one scenario for that.

dpakach (commented):
LGTM :+1:

phil-davis (approved):

Issue #37085, opened 2020-03-05
Growing number of ajax requests when clicking "Shared with X" sections


PVince81:
### Steps

1. Open the files app
1. Open the network console
1. Switch repeatedly between "All files" and on of the "Shared with X" sections

### Expected result
No repeated requests when opening the "Shared with X" section

### Actual result
Every time the section is opened again, an additional request is made.
The first time 1, the second time 2, the third time 3 requests, etc.
For "Shared with you" there's a pair of request. That pair gets repeated multiple times when returning.

### Version
ownCloud 10.4.1 pre-alpha git master 433ba3ee41cedb7387424f108534bfd62ae6eb28

Likely minor issue as this is not a proper use case.

karakayasemi:
Fixed with #38385

PR #38385, opened 2021-02-09
Trigger events once

jvillafanez:
## Description
Prevent the registration of the same event listener several times, which causes the same action (from the duplicated listeners) to be performed multiple times.

Primary problem is with the "urlChanged" event, which causes an ajax request to be thrown. This means that, without this fix, the ajax requests performed were increasing along with the increasing event listeners being registered for that event.
The other events don't seem to be too critical since the attached action is pretty light.

## Related Issue
Helps with https://github.com/owncloud/enterprise/issues/4374

## Motivation and Context
No need to register the same event listener twice

## How Has This Been Tested?
Mostly tested going back and forth between the "all files" and "shared with you" sections in the files view, as well as checking there is no apparent problem with the scrolling in those sections.
Event registration has been checked with the firefox dev tools, ensuring there is no increasing event associated with the target web element. There could be multiple listeners for the same event, but it's expected to be for different purposes (not duplicated)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38385
+++ b/changelog/unreleased/38385
@@ -1,0 +1,10 @@
+Bugfix: Prevent multiple calls by not registering the same listener twice
+
+Going back and forth among the file sections ("all files", "shared with you",
+etc) was making some event listeners to be registered twice or more times.
+This was causing the same ajax request to be called several times causing
+unnecessary load in the server.
+
+Now, these additional requests won't happen
+
+https://github.com/owncloud/core/pull/38385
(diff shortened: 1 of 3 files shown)


JammingBen (approved):
Didn't know that namespacing jQuery-events is a thing, cool stuff :+1:

karakayasemi:
Looks good. Fixes https://github.com/owncloud/core/issues/37085

changelog/unreleased/38385
```diff
@@ -0,0 +1,10 @@
+Bugfix: Prevent duplicated calls by not registering the same listener twice
```
JanAckermann, line 1: duplicated -> multiple

JanAckermann: Otherwise, good working solution :+1:

AlexAndBear (approved):

Issue #38382, opened 2021-02-08
Groups not created in drone tests


ishank011:
I'm working on adding group sharing to reva, for that we need to be able to retrieve groups from the ldap server using the [userprovider service](https://github.com/cs3org/reva/blob/master/pkg/user/manager/ldap/ldap.go#L347-L384) and the [corresponding config](https://github.com/cs3org/reva/blob/master/tests/oc-integration-tests/drone/ldap-users.toml#L37). In a recent PR, the groups are added when authenticating as well: https://github.com/cs3org/reva/pull/1456. But the logs show that no groups are actually retrieved

`2021-02-05 13:25:36.001 INF drone/src/internal/grpc/services/authprovider/authprovider.go:114 > user id:<idp:"http://localhost:18000" opaque_id:"Alice" > username:"Alice" mail:"alice@example.org" display_name:"Alice Hansen" opaque:<map:<key:"gid" value:<decoder:"plain" value:"5000" > > map:<key:"uid" value:<decoder:"plain" value:"30000" > > >  authenticated pid=15 pkg=rgrpc traceid=4789c1400ce382bb155dc43b7f9a2048`

and they aren't added either https://drone.cernbox.cern.ch/cs3org/reva/498/10/2. Since in the [acceptance tests](https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementBasicToShares/createShareToSharesFolder.feature#L165), we try sharing with groups such as 'grp1', we need to add these for the CI to pass. @phil-davis @butonic

ishank011:
@phil-davis can you look at this please? We wanted to merge the group sharing PR so it'll be good to run the tests before that

phil-davis:
https://drone.cernbox.cern.ch/cs3org/reva/553/9/6
```
And group "grp1" has been created 
LDAP group exists: cn=grp1,ou=TestGroups,dc=owncloud,dc=com
```

That is logged by https://github.com/owncloud/core/pull/38392/files#diff-53ecaebb4552a8c12cb891ca32d9879923ec2d0226afeda7e559b9f631a1300cR3662

And that calls laminas-ldap `getEntry()` https://github.com/laminas/laminas-ldap/blob/2.12.x/src/Ldap.php#L1286 which claims that the group entry exists in the TestGroups ou.

ishank011:
[...]

PR #38393, opened 2021-02-11
[tests-only] Do not run occ group:list in OCIS CI

phil-davis:
## Description
When running CI with `cs3org/reva` we use LDAP. But the existing code tries to run `occ group:list` to sync the groups after creating an LDAP group. That is only useful on oC10 - the `occ` command does not exist on OCIS-reva.

This was noticed while investigating https://github.com/owncloud/core/issues/38382

We might also need to do some "magic" in OCIS-reva to make sure that the new group "really is known" to OCIS-reva. But that will come later if we need it.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


dpakach (approved):
lgtm

individual-it (approved):

PR #38396, opened 2021-02-12
Add the owner to public link shares

JammingBen:
## Description
Add the owner to public link shares because we cannot retrieve this information otherwise (in contrast to what the comment in https://github.com/owncloud/core/blob/master/core/js/sharemodel.js#L72 says).

## Related Issue
- needed for https://github.com/owncloud/files_spaces/issues/51

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38396
+++ b/changelog/unreleased/38396
@@ -1,0 +1,7 @@
+Bugfix: Add the owner to public link shares
+
+Add the owner to public link shares because we cannot retrieve this
+information otherwise.
+
+https://github.com/owncloud/core/pull/38396
+https://github.com/owncloud/files_spaces/issues/51
--- a/core/js/sharemodel.js
+++ b/core/js/sharemodel.js
@@ -70,8 +70,6 @@
 			delete data.url;
 
 			// these can be read from the parent object or fileinfo
-			delete data.displayname_file_owner;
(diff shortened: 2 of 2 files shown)


jvillafanez:
I think we should include either the uid (both owner and file owner) or uid + displayname. For me, the uid is primary information while the displayname is secondary, so having just the displayname seems lacking.
Also note that the displayname isn't unique, so there could be duplicates.

JammingBen:
> I think we should include either the uid (both owner and file owner) or uid + displayname. For me, the uid is primary information while the displayname is secondary, so having just the displayname seems lacking.
Also note that the displayname isn't unique, so there could be duplicates.

You mean additionally, or instead of the displayname? Because for my case, I really need the displayname. Or is there an easy way to get the displayname via uid in frontend?

jvillafanez:
Additionally. If you need the displayname, that's totally fine.
You can use the uid to distinguish between 2 users with the same displayname.

JammingBen:
Alright, makes sense, thanks! Removed those two lines as well.

jvillafanez (approved):

PR #38400, opened 2021-02-14
polyfill samesite cookie attribute

fschade:
## Description
some browsers do not support samesite cookie attribute, this pr introduces a cookie helper to polyfill those requirements. The samesite attribute is needed to define where the cookie can be used. This is a major requirement when using openid connect where the idp is deployed to a different domain or similar scenarios.  The cookie helper supports 2 types of polyfills.

1.) DETECT: tries to detect if given user_agent supports samesite attribute
2.) FALLBACK: creates a second cookie "cookie-name-legacy" wich has no samesite attribute (recommended by google)

the detect polyfill is required because the session_cookie is not able to have 2 cookies.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4191

## Motivation and Context
- https://auth0.com/blog/browser-behavior-changes-what-developers-need-to-know/
- https://blog.chromium.org/2019/10/developers-get-ready-for-new.html
- https://devblogs.microsoft.com/aspnet/upcoming-samesite-cookie-changes-in-asp-net-and-asp-net-core/
- https://developer.mozilla.org/de/docs/Web/HTTP/Headers/Set-Cookie/SameSite

## How Has This Been Tested?
- local installation 
- openidconnect on azure ad 

## Types of changes

## Checklist:

- [x] test with all supported browsers
- [x] test with supported idps
- [x] test with supported platforms
- [ ] changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)
- [ ] unit tests

### DIFF
--- a/lib/private/Http/CookieHelper.php
+++ b/lib/private/Http/CookieHelper.php
@@ -23,12 +23,11 @@ namespace OC\Http;
 
 use OCP\IRequest;
 
-
 /**
  * Class CookieHelper
  *
  * This Class polyfills cookie to work around samesite restrictions on unsupported browsers.
- * It supports POLYFILL_DETECT mode which tries to detect if the samesite attribute ist supported.
+ * It supports POLYFILL_DETECT mode which tries to detect if the samesite attribute is supported.
  * It supports POLYFILL_FALLBACK mode which creates a second cookie without samesite attribute.
  *
  * @package OC\Http
@@ -48,9 +47,9 @@ class CookieHelper {
 
 	public static function setCookie(string $name, string $value = '', array $options = [], string $polyfill = self::POLYFILL_FALLBACK): void {
 		// check if given cookie options are allowed
-		foreach (array_keys($options) as $option) {
-			if (!in_array($option, array(self::PATH, self::DOMAIN, self::EXPIRES, self::SECURE, self::HTTPONLY, self::SAMESITE))) {
-				trigger_error(
+		foreach (\array_keys($options) as $option) {
+			if (!\in_array($option, [self::PATH, self::DOMAIN, self::EXPIRES, self::SECURE, self::HTTPONLY, self::SAMESITE])) {
+				\trigger_error(
 					'unknown cookie option : ' . $option,
 					E_WARNING
 				);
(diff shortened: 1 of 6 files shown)


lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,195 @@
...
+namespace OC\Http;
+
+use OCP\IRequest;
+
+
+/**
+ * Class CookieHelper
+ *
+ * This Class polyfills cookie to work around samesite restrictions on unsupported browsers.
+ * It supports POLYFILL_DETECT mode which tries to detect if the samesite attribute ist supported.
```
phil-davis, line 31:
```suggestion
 * It supports POLYFILL_DETECT mode which tries to detect if the samesite attribute is supported.
```

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,195 @@
...
+			if (strtolower($sameSite) === strtolower($v)) {
+				array_push($header, sprintf('SameSite=%s', $v));
+				break;
+			}
+		}
+
+		header(implode('; ', $header), false);
+	}
+
+	// if polyfill is set to true it returns the cookie legacy name
```
phil-davis, line 127:
```suggestion
	// return the cookie legacy name
```

I don't see an "if" statement in this function.

phil-davis (commented):
Minor comments only. And there are some code-style fails https://drone.owncloud.com/owncloud/core/28612/2/7

fschade:
> Minor comments only. And there are some code-style fails https://drone.owncloud.com/owncloud/core/28612/2/7

thanks phil, done

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,192 @@
+<?php declare(strict_types=1);
+/**
+ * @author Florian Schade <f.schade@icloud.com>
+ *
+ * @copyright Copyright (c) 2018, ownCloud GmbH
```
DeepDiver1975, line 5: 2021

phil-davis:
Lots of people have been requested for review, so I removed "WIP" from the title.

phil-davis (commented):
Needs changelog?

IljaN (approved):

micbar (changes requested):
CookieHelper needs full unit test coverage

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,192 @@
...
+	public const PATH = 'path';
+	public const DOMAIN = 'domain';
+	public const EXPIRES = 'expires';
+	public const SECURE = 'secure';
+	public const HTTPONLY = 'httponly';
+	public const SAMESITE = 'samesite';
+	public const POLYFILL_DETECT = 'detect';
+	public const POLYFILL_FALLBACK = 'fallback';
+
+	public static function setCookie(string $name, string $value = '', array $options = [], string $polyfill = self::POLYFILL_FALLBACK): void {
```
DeepDiver1975, line 48: no static functions - you cannot unit test this that way

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,192 @@
...
+				\trigger_error(
+					'unknown cookie option : ' . $option,
+					E_WARNING
+				);
+			}
+		}
+
+		// check if given polyfill mode is supported
+		if (!\in_array($polyfill, [self::POLYFILL_DETECT, self::POLYFILL_FALLBACK])) {
+			\trigger_error(
```
DeepDiver1975, line 61: no trigger error - throw exceptions

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		foreach ([self::SAMESITE_NONE, self::SAMESITE_LAX, self::SAMESITE_STRICT] as $v) {
+			if (\strtolower($sameSite) === \strtolower($v)) {
+				\array_push($header, \sprintf('SameSite=%s', $v));
+				break;
+			}
+		}
+
+		\header(\implode('; ', $header), false);
```
DeepDiver1975, line 121: you cannot unit test the header call

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,192 @@
...
+		if (\array_key_exists(self::legacyName($name), $request->cookies)) {
+			return $request->cookies[self::legacyName($name)];
+		}
+
+		return null;
+	}
+
+	public static function canSameSite($user_agent = ''): bool {
+		if ($user_agent === '') {
+			$user_agent = $_SERVER['HTTP_USER_AGENT'];
```
DeepDiver1975, line 143: user request instance

lib/private/Http/CookieHelper.php
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		if (\array_key_exists(self::legacyName($name), $request->cookies)) {
+			return $request->cookies[self::legacyName($name)];
+		}
+
+		return null;
+	}
+
+	public static function canSameSite($user_agent = ''): bool {
```
DeepDiver1975, line 141: string declaration missing

DeepDiver1975 (changes requested):

lib/private/Http/CookieHelper.php
micbar: move into wrapper method

lib/private/Http/CookieHelper.php
IljaN: Request instance via \OC::$server->getRequest is not initialized at this point. Also not in the calling code. Any ideas?

micbar: good catch.

DeepDiver1975: classic cyclic dependency ....

get rid of it by removing the default value from user_agent.

caller has to properly fill this parameter

this way you get a testable CookieHelper - caller afaik is anyhow not unit tested ....

IljaN:
Superseded by https://github.com/owncloud/core/pull/38457

Issue #38156, opened 2020-11-29
Run acceptance tests with various databases


phil-davis:
Run acceptance tests with different databases to verify that all is OK.

- [x] mariadb:10.2 - this is the regular database used in CI
- [x] mariadb:10.3 #38157 pass
- [x] mariadb:10.4 #38158 pass
- [x] mariadb:10.5 #38159 pass
- [x] mysql:5.5 #38155 see https://github.com/owncloud/core/pull/38155#issuecomment-735387782 5.5 is not officially supported anyway
- [x] mysql:5.7 #38154 pass (but not officially supported anyway)
- [x] mysql:8.0 #38084 pass
- [x] postgres:9.4 #38153 passes with a couple of `skipOnDbPgsql`
- [x] postgres:10.3 #38086 
- [x] oracle #38082 passes with the `skipOnDbOracle` tags added in #38160 and #38166

phil-davis:
I am re-running these to make sure it is all good:
- [x] postgres:9.4 #38402 
- [ ]

phil-davis:
All good now. It is nice to know that the test suite runs with each database. I raised any issues related to specific databases.

PR #38401, opened 2021-02-14

phil-davis:
Issue #38156


phil-davis:
https://drone.owncloud.com/owncloud/core/28618
`yaml: line 1: mapping values are not allowed in this context`

There is some crazy problem with drone with this particular starlark. I will try to work-around it...

phil-davis:
The problem is due to the length of the generated pipeline names - closing here.

PR #38405, opened 2021-02-14
[tests-only] Run acceptance tests with mysql:5.5

phil-davis:
Issue #38156


phil-davis:
https://drone.owncloud.com/owncloud/core/28622/54/13 - fails with emojis

PR #38430, opened 2021-02-19
fix(files_versions): when resolving the storage of a file the full pa…

DeepDiver1975:
…th has to be taken into account

## Description
files_versions has been integrated with the new storage based versioning IVersionedStorage
Until today this was only tested with objectstore versioning which is hooked up as primary storage.

When trying to access a versioned storage which is mounted as non-root this logic did not work out.

## Motivation and Context
Allow to use versioned storage in mounted storages.

## How Has This Been Tested?
- under development in files_spaces

## Types of changes
- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -78,7 +78,7 @@ class Storage {
 		//until the end one version per week
 		6 => ['intervalEndsAfter' => -1,      'step' => 604800],
 	];
-	
+
 	/** @var \OCA\Files_Versions\AppInfo\Application */
 	private static $application;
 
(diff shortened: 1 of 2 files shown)


phil-davis:
@DeepDiver1975 changelog needed?

DeepDiver1975:
> @DeepDiver1975 changelog needed?

yes - and I want to add a test for this .... Monday ;-)

DeepDiver1975:
>  and I want to add a test for this

not that easy .... bloody codebase ....

apps/files_versions/lib/Storage.php
```diff
@@ -161,8 +161,9 @@ public static function store($filename) {
 			}
 
 			list($uid, $filename) = self::getUidAndFilename($filename);
+			$fullFileName = "/$uid/files$filename";
```
jvillafanez, line 164: Couple of tiny details here, which could be solved with a quick code comment:
* It seems there is a "/" missing: `/$uid/files/$filename`. I guess the filename starts with "/" so it isn't needed.
* $filename could be misunderstood as just the filename, not the path. I assume it contains the whole path, such as `/folder1/folder2/filename`

jvillafanez (approved):
Just tiny details, but approving anyway.

apps/files_versions/lib/Storage.php
DeepDiver1975: Push permission granted ... feel free to add a comment ... THX

VicDeo, line 164: isn't `$filename` a full file name before the reassignment in the previous line?

DeepDiver1975: in a simplistic scenario: yes
but getUidAndFilename is also taking care of resolving shares to the owner

AlexAndBear (approved):

PR #38437, opened 2021-02-23
[tests-only] use oC10 tarball installation as system under test in drone

saw-jan:
this PR only adds the functions that can download tarball and setup server

issue https://github.com/owncloud/QA/issues/437


dpakach (commented):
LGTM :+1:

swoichha (commented):
lgtm

individual-it (approved):

PR #38440, opened 2021-02-24
Add value userType to sharees api 

### LINKED ISSUE #38422: [QA] guest user email is shown as normal user, when used for the second time
Seen with owncloud server 10.6.0 and guests app 0.9.0 

* Enter a yet unknown email address in the share dialog.
* A choice appears for "federated" or "guest". OK

image

* Click the guest. ownCloud creates a user entry for the guest. 
* Can be verified in Admin Users or e.g. metrics overview:
image

* Enter the same email address again for another share. The choice now is "user" and "Federated". BAD
image

Expected behaviour: guest users are always flagged as "guest".

AlexAndBear:
## Related Issue

- Fixes https://github.com/owncloud/core/issues/38422

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/public/User.php
+++ b/lib/public/User.php
@@ -44,7 +44,7 @@ namespace OCP;
  * about the currently logged in user and the permissions for example
  * @since 5.0.0
  */
-class User {
+class User extends \OCP\User\Constants {
 	/**
 	 * Get the user id of the user currently logged in.
 	 * @return string uid or false
--- a/apps/files_sharing/lib/Controller/ShareesController.php
+++ b/apps/files_sharing/lib/Controller/ShareesController.php
@@ -26,6 +26,7 @@
 
 namespace OCA\Files_Sharing\Controller;
 
+use OC\Helper\UserTypeHelper;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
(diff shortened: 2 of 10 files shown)


mrow4a (commented):
can api return „userType” instead? is this possible? this would make it more generic

JanAckermann:
> can api return „userType” instead? is this possible? this would make it more generic

Of course, we can do this, are there more types than 'user' and 'guest'?

mrow4a:
> > can api return „userType” instead? is this possible? this would make it more generic
> 
> Of course, we can do this, are there more types than 'user' and 'guest'?

needs to be investigated, best if returns some ID / CONST defined in core.

JanAckermann:
@mrow4a ok updated in that favor, might have quick look? if you give me green light, I will update the tests

lib/public/User.php
```diff
@@ -44,7 +44,7 @@
  * about the currently logged in user and the permissions for example
  * @since 5.0.0
  */
-class User {
+class User extends \OC\User\Constants{
```
mrow4a: public/User using private/User ? Constanst should be public, and not private implementation

mrow4a (changes requested):
generally makes sense, one small detail to be fixed

lib/public/User.php
JanAckermann: I took the share implementation as boilerplate

JanAckermann:
@mrow4a Tests passed, approve?

mrow4a (approved):

PR #38445, opened 2021-02-25
[Test-only] Adds issue tag to an existing API test

SwikritiT:
## Description
 After renaming a file inside of a directory , eTag of the directory should change but that doesn't happen in OCIS. This PR adds issue tag to an existing API test that demonstrates this bug.
 
## Related Issue

- https://github.com/owncloud/ocis/issues/1590

## How Has This Been Tested?

- locally

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavEtagPropagation1/moveFileFolder.feature
+++ b/tests/acceptance/features/apiWebdavEtagPropagation1/moveFileFolder.feature
@@ -4,7 +4,7 @@ Feature: propagation of etags when moving files or folders
   Background:
     Given user "Alice" has been created with default attributes and without skeleton files
 
-
+  @issue-ocis-1590
   Scenario Outline: renaming a file inside a folder changes its etag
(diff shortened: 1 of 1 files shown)


phil-davis:
@dpakach @individual-it what is going to be the way we do this sort of thing now?

I cannot find the scenario in the expected-failures files in `owncloud/ocis` - so it seems to pass in CI - what is going on?

I was expecting that the scenario(s) would be in the expected-failures files, and that we would put the link to the issue in the expected-failures files.

individual-it:
its true, the tests pass in CI https://drone.owncloud.com/owncloud/ocis/3692/38/7 but did not work for @SwikritiT locally see #1590
I guess there is a setup issue in the local ocis run

alialqh:
A

alialqh:
A

PR #38438, opened 2021-02-23
[tests-only] name the used skeleton dir explicitly in the feature files

individual-it:
## Description
name the used skeleton explicitly in the feature description, don't rely on guessing anymore

## Related Issue

- Fixes owncloud/QA/issues/652

## Motivation and Context
working towards owncloud/QA/issues/628
generally speaking the person reading the feature file should understand what happens and should not need to know what "magic" happens in the background

## How Has This Been Tested?
- :robot: in this repo and on ocis: https://github.com/owncloud/ocis/pull/1703

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38446, opened 2021-02-25
[tests-only] Remove skeletondirectory

phil-davis:
## Description
- remove the code in acceptance tests `run.sh` that automagically chooses "api" or "webui" skeleton
- adjust tests that were depending on some skeleton files existing "by default"

## Related Issue
- Fixes https://github.com/owncloud/QA/issues/656

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/run.sh
+++ b/tests/acceptance/run.sh
@@ -978,22 +978,8 @@ do
 done
 
 #set the skeleton folder
-if [ -z "${SKELETON_DIR}" ]
+if [ -n "${SKELETON_DIR}" ]
 then
-	# calculate the correct skeleton folder
-	if [ "${RUNNING_API_TESTS}" = true ] || [ "${RUNNING_CLI_TESTS}" = true ]
-	then
-		# CLI tests use the apiSkeleton so that API-based "then" steps can be used
-		# to check the state of users after CLI commands
-		export SRC_SKELETON_DIR="apiSkeleton"
-	else
-		export SRC_SKELETON_DIR="webUISkeleton"
-	fi
-	for URL in ${TESTING_APP_URL} ${TESTING_APP_FED_URL}
-	do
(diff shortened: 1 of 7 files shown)


tests/acceptance/run.sh
```diff
@@ -978,22 +978,8 @@ do
...
-		export SRC_SKELETON_DIR="webUISkeleton"
-	fi
-	for URL in ${TESTING_APP_URL} ${TESTING_APP_FED_URL}
-	do
-		curl -k -s -u ${ADMIN_AUTH} ${URL}testingskeletondirectory -d "directory=${SRC_SKELETON_DIR}" > /dev/null
-	done
-else
 	for URL in ${OCC_URL} ${OCC_FED_URL}
 	do
 		remote_occ ${ADMIN_AUTH} ${URL} "config:system:set skeletondirectory --value=${SKELETON_DIR}"
```
phil-davis, line 985: Currently a developer can specify `SKELETON_DIR` before running `run.sh` and this logic will set it. But actually I think that the tests will now always set the skeleton to small, large or none in each scenario where it matters. So the developer will just be "injecting" a skeleton that has effect only in some "random" scenarios that actually to not mention the skeleton at all and do not reference any skeleton files... anyway.

Should I remove this code?

individual-it: It still needed for OCIS. The tests take the skeleton env. variable and the specific steps only change the last part of it if needed see https://github.com/owncloud/core/pull/38438/files#diff-53ecaebb4552a8c12cb891ca32d9879923ec2d0226afeda7e559b9f631a1300cR5016-R5019

so what really should change is that the setting for ocis is not `SKELETON_DIR` anymore but `SKELETON_ROOT_DIR` and then the steps that need it would have to add to that directory

but that would be an breaking change and needs to be communicated and documented well

phil-davis: Good - we can plan that separately.

individual-it (approved):

PR #38474, opened 2021-03-07
Do not remove federated share on CURLE errors, provide improved notification in UI and exception handling

mrow4a:
fixes:
- https://github.com/owncloud/enterprise/issues/4311 
-  implements "TODO" from https://github.com/owncloud/core/blob/v10.6.0/apps/files_sharing/lib/External/Storage.php#L238

problem:
currently (and without some sophisitcated implementation in the future) there is no clean way of deciding of state of federated share. Currently PROPFIND is send, and if 404 not found is returned, then availability of system by probing e.g. status.php endpoint is performed. If status check succeed, federated share is removed. However, customer with large federation setup reported that shares get falsely removed under certain conditions on the server (e.g. some maintenance).

usecase:
- there are two servers, fed1 and fed2. fed1 gets unavailable, user in fed2 removes federated share that he shared with fed1. fed1 never receives share removal. now fed1 becomes available, provide proper notification and log exception in the logs
- due to some maintenance work, server returns CURLE_COULDNT_CONNECT or CURLE_OPERATION_TIMEDOUT or NotFound. This indicated to fed share storage provider that it should "test remote". However, now for some reason all "status checks like status.php" succeed, and we should let the user in fed1 decide what to do with share, instead of blindly removing

to be done:
- [x] provide most basic notification
- [x] investigate further notifications
- [x] tests

### DIFF
--- a/apps/files_sharing/lib/External/Storage.php
+++ b/apps/files_sharing/lib/External/Storage.php
@@ -49,6 +49,8 @@ class Storage extends DAV implements ISharedStorage {
 	private $memcacheFactory;
 	/** @var \OCP\Http\Client\IClientService */
 	private $httpClient;
+	/** @var \OCP\ILogger */
+	private $logger;
 	/** @var \OCP\ICertificateManager */
 	private $certificateManager;
 	/** @var bool */
@@ -62,6 +64,7 @@ class Storage extends DAV implements ISharedStorage {
 	public function __construct($options) {
 		$this->memcacheFactory = \OC::$server->getMemCacheFactory();
 		$this->httpClient = \OC::$server->getHTTPClientService();
+		$this->logger = \OC::$server->getLogger();
 		$this->manager = $options['manager'];
 		$this->certificateManager = $options['certificateManager'];
 		$this->remote = $options['remote'];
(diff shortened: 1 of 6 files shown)


apps/files_sharing/lib/External/Storage.php
```diff
@@ -220,12 +223,17 @@ public function test() {
 	public function checkStorageAvailability() {
 		// see if we can find out why the share is unavailable
 		try {
-			if (! $this->propfind('')) {
-				// a 404 can either mean that the share no longer exists or there is no ownCloud on the remote
+			if (!$this->propfind('', $strictNotFoundCheck=true)) {
```
VicDeo, line 226: @mrow4a PHP 8.0?

apps/files_sharing/lib/External/Storage.php
```diff
@@ -220,12 +223,17 @@ public function test() {
...
+				// not found returned, can either mean that the share no longer exists
+				// or there is no ownCloud on the remote (proxy could return 404 for that path)
 				if ($this->testRemote()) {
 					// valid ownCloud instance means that the public share no longer exists
 					// since this is permanent (re-sharing the file will create a new token)
 					// we remove the invalid storage
+					$this->logger->error(
+						'Storage for external share {shareId} returns not found error and 
+						likely has been removed with failed notification to this instance receipent',
```
jvillafanez, line 235: Double-check this is properly formatted in the logs. There could be tabs showing.

jvillafanez:
I'm more inclined to remove the auto-delete feature and let the users decide. As long as the error is properly notified to the users and they can remove the failing remote, it shouldn't be a problem.
There could be additional problems such as a temporary network problem, problems with your ISP, server temporary down... so auto-removing the remote share could be a wrong choice if the admins can recover the connection later.

mrow4a:
@jvillafanez I also was inclined in this direction until I saw how sync clients is handling this. Not so good -> with auto-delete on auth-error [1] it just removes the share, and uploads its local files to the newly created local folder as replacement, however without auto-delete it goes in some very noisy internal server error loop. on not-found error [2] it tests remote , and only if that fails fed share is removed.

This is why I made sure to do "strict not-found check" to exclude all network related problems. I am however myself puzzled whether not to stop testing remote and in that case (when not found is returned) to just throw error to client and let user decide what to do with that. @VicDeo @jvillafanez  thoughs? 

[1] auth-error is rather clear signal share got removed, it is that this fed share token can no longer be found.
[2] not-found error to me it is not really clear signal as cannot really think of case when this is thrown

jvillafanez:
> however without auto-delete it goes in some very noisy internal server error loop

I guess that's something we should investigate. Maybe the clients should implement an exponential backoff for the automatic retries, and reset the backoff on manual interaction.

I'm ok with the solution, but it seems we have a technical debt we'll have to clean at some point.

jvillafanez (approved):

AlexAndBear (approved):

JanAckermann:
@mrow4a LGTM, merge?

PR #38481, opened 2021-03-09
Add exceptions messages whenever a node cannot be moved

JammingBen:
## Related Issue
- Helps with https://github.com/owncloud/enterprise/issues/4427

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -431,7 +431,7 @@ class Directory extends Node implements ICollection, IQuota, IMoveTarget {
 
 		# check the destination path, for source see below
 		if (Filesystem::isForbiddenFileOrDir($destinationPath)) {
-			throw new SabreForbidden();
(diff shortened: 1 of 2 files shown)


jvillafanez:
I wanted to add the message within the exception.
```
{"Exception":"Sabre\DAV\Exception\Forbidden","Message":"","Code":0,"Trace":"......
```
You can consider that there is already a log in place since we're logging the exception somewhere, but the problem is that the message of the exception is empty, as shown above. In this particular case, this empty message is what makes difficult to trace where the exception is being triggered because there are multiple places where this could happen.

Something like `throw new SabreForbidden('The destination directory cannot be created');`, for example, should be enough to fill that empty message in the exception

JammingBen:
Ahh sorry, misunderstood this, please have a look again.

jvillafanez (approved):

PR #38483, opened 2021-03-09
Disable sharing on files_external if auth mechanism is not compatible

AlexAndBear:
## Description
If we set up external storage and choose the auth mechanism 'Log-in credentials, save in session' sharing is not possible, to show this to the user, the checkbox will be disabled and a hint due a tooltip will be shown

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4444

image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/js/settings.js
+++ b/apps/files_external/js/settings.js
@@ -26,8 +26,8 @@ var MOUNT_OPTIONS_DROPDOWN_TEMPLATE =
 	'		<input id="mountOptionsPreviews" name="previews" type="checkbox" value="true" checked="checked"/>' +
 	'		<label for="mountOptionsPreviews">{{t "files_external" "Enable previews"}}</label>' +
 	'	</div>' +
-	'	<div class="optionRow">' +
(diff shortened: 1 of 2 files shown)


apps/files_external/js/settings.js
```diff
@@ -499,8 +500,12 @@ MountOptionsDropdown.prototype = {
 			MountOptionsDropdown._template = template;
 		}
 
+		var sharingDisabled = ['password::sessioncredentials'].includes(storageConfig.authMechanism);
```
jvillafanez, line 503: According to https://developer.mozilla.org/es/docs/Web/JavaScript/Reference/Global_Objects/Array/includes , `includes` isn't supported by IE. I don't think it's critical, but we could check if there is an easy alternative.

JanAckermann: Yes, we don't need this at all, it's just easy to extend the array

jvillafanez (approved):

PR #38486, opened 2021-03-09
On OC-RequestAppPassword header and login exception return www-authenticate header

mrow4a:
fixes https://github.com/owncloud/enterprise/issues/4196

- [x] test with different authentication methods installed and use app password here for OCS API
- [x] add tests

### 1. Try accessing API in web browser, it will fail
Zrzut ekranu 2021-03-9 o 22 24 23

### 2. Create APP Password in ownCloud personal settings
Zrzut ekranu 2021-03-9 o 22 26 39

### 3. Add header `OC-RequestAppPassword` to indicate we want to request interactive app password authentication
Zrzut ekranu 2021-03-9 o 22 24 35

### 4. Authenticate using created app password and enjoy

Zrzut ekranu 2021-03-9 o 22 24 47
Zrzut ekranu 2021-03-9 o 22 25 23


DeepDiver1975:
Plz remove X-

This prefix is depreciated

jvillafanez:
Code looks good. However, I think we miss some additional information which should be documented somehow:
* Expected interaction with other authentication methods such as saml, oauth, etc.
  * Is it possible to authenticate using an oAuth token instead of username+password?
* Is it required the usage of an app password?
* Is it possible to send the authentication directly without letting the browser pop up the window?

This is mainly to know the limitations, if any, so we can decide if we need some additional fix.

mrow4a:
@jvillafanez this mechanism is only limited to app passwords , and it is not planned to use it otherwise (up to my understanding). Thus name of the header.

This needs to be properly documented.

micbar (approved):
LGTM, thanks for the acceptance tests!

mrow4a:
@micbar merge?

Issue #30313, opened 2018-01-30
Configurable minimum characters before autocomplete user searches


tomneedham:
Motivation:

For large user lists searching for the first two characters can be very slow to return.

PVince81:
custom groups also has such autocomplete...

Here's a possible approach:

1) Create a new oc_appconfig setting that sets the minimum chars for any user or group related autocomplete
2) Config value is exposed over capabilities API
3) Add setting in UI somewhere or keep it internal but documented ?
4) Frontend code must read capability value (accessing through `OC.getCapabilities()`) and adjust autocomplete threshold accordingly (needs change in sharing code and customgroups)
5) Backend code `UserManager` and `GroupManager` refuse to work if the search pattern given to any `find*` and `search*` methods is shorter than the minimum. This saves us for modifying backend for any other endpoints.

One concern with 5 is that some code paths might call the search functions with empty string. Maybe we should still allow empty string. But then need to check that the autocomplete backend code doesn't allow for empty string either (it already shouldn't).

tomneedham:
> call the search functions with empty string

Hmm why would they do that... to list all users. We have other methods for that.

PVince81:
I don't know... legacy stuff...

PVince81:
as discussed in chat, the approach above is what we want to do

PVince81:
- [ ] also apply the limit to ajax/share.php's "getShareWithEmail" that is used by the link share email field (see https://github.com/owncloud/core/issues/30670 for steps)

VicDeo:
https://github.com/owncloud/core/pull/30798 and https://github.com/owncloud/core/pull/30994 were merged -> close.

PR #38489, opened 2021-03-10
fix: require search param to not be empty

C0rby:
## Description

Before listing users or groups as share recipients we check if the search parameter is empty and if it is empty we return an empty list. This prevents a users to circumvent the 'user.search_min_length' restriction.

The better fix would be to change the `isSearchable` method in lib/public/Util/UserSearch.php. This would require many changes though since this method is being used at the Manager level and some users of the managers expect the old behavior. Ideally the `isSearchable` call would be moved to the controller / API level.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4456

## How Has This Been Tested?

- locally:
- via curl against the sharees api
- in the browser via webui

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controller/ShareesController.php
+++ b/apps/files_sharing/lib/Controller/ShareesController.php
@@ -165,6 +165,11 @@ class ShareesController extends OCSController {
 	protected function getUsers($search) {
 		$this->result['users'] = $this->result['exact']['users'] = $users = [];
 
+		if (\strlen(\trim($search)) === 0 && $this->userSearch->getSearchMinLength() > 0) {
+			$this->result['users'] = [];
+			return;
+		}
+
 		$userGroups = [];
 		if ($this->shareWithGroupOnly || $this->shareeEnumerationGroupMembers) {
 			// Search in all the groups this user is part of
@@ -295,6 +300,11 @@ class ShareesController extends OCSController {
 	 */
 	protected function getGroups($search) {
 		$this->result['groups'] = $this->result['exact']['groups'] = [];
+
+		if (\strlen(\trim($search)) === 0 && $this->userSearch->getSearchMinLength() > 0) {
+			$this->result['groups'] = [];
+			return;
+		}
 
 		$groups = $this->groupManager->search($search, $this->limit, $this->offset, 'sharing');
 		$groupIds = \array_map(function (IGroup $group) {
--- a/apps/files_sharing/tests/API/ShareesTest.php
+++ b/apps/files_sharing/tests/API/ShareesTest.php
@@ -526,6 +526,53 @@ class ShareesTest extends TestCase {
 				true,
 				'email'
 			],
+			// Test empty searchTerm should not return users.
+			['', false, true, [], [], [], [], false, false, false, false, 'id', 1],
+			['', true, true, [], [], [], [], false, false, false, false, 'id', 1],
+			[
+				'',
+				false,
+				true,
+				[],
+				[
+					$this->getUserMock('test0', 'Test'),
+					$this->getUserMock('test1', 'Test One'),
+					$this->getUserMock('test2', 'Test Two'),
+				],
+				[
+					[
+						'label' => 'Test',
+						'value' => [
+							'shareType' => Share::SHARE_TYPE_USER,
+							'shareWith' => 'test0',
+							'userType' => User::USER_TYPE_USER,
+						]
+					],
+					[
+						'label' => 'Test One',
+						'value' => [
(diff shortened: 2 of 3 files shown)


phil-davis:
@C0rby I guess that there should be some unit test cases added to cover the "early returns" here.

Should we add add a couple of API acceptance test cases also?

C0rby:
> @C0rby I guess that there should be some unit test cases added to cover the "early returns" here.

Yes, you're right.

> Should we add add a couple of API acceptance test cases also?

Wouldn't hurt, I guess.

VicDeo:
Changes are looking good, who is expected to add unit tests here? :)

IljaN:
@phil-davis Acceptance test outline:

- [ ] If query param `search` is empty or not present no users should be returned unless 'user.search_min_length' = 0 in config.php
- [ ] Users should only be returned if length of ``search`` >=  'user.search_min_length'

C0rby:
Also only if `\trim(search) !== ''`.
So `user.search_min_length = 0` and `$search = ' '` won't return anything.

phil-davis:
I have put this in WQA/CI/Test Automation project and pinged the guys about it. They will have a look in the morning.

C0rby:
@IljaN, I fixed the unit tests but now the smoke tests are failing...

dpakach:
@C0rby @IljaN I've added some acceptance tests for this in https://github.com/owncloud/core/pull/38491, please review that

micbar (approved):
LGTM

micbar:
@dpakach please also merge the new Tests PR

dpakach:
@micbar @C0rby The tests are failing because of issue
When `users.search_min_length` is set to `0` and get sharees with empty search, you get normal users under `data>exact>users`
Drone build here - https://drone.owncloud.com/owncloud/core/29125/65/13 

Is this normal behavior?

Also confirmed manually
```
curl http://localhost/ocs/v1.php/apps/files_sharing/api/v1/sharees\?itemType\=file -u admin:admin
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message>OK</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data>
  <exact>
   <users>
    <element>
     <label>tu1</label>
     <value>
      <shareType>0</shareType>
      <shareWith>tu1</shareWith>
      <userType>0</userType>
     </value>
    </element>
   </users>
   <groups/>
   <remotes/>
  </exact>
  <users>
   <element>
    <label>admin</label>
    <value>
     <shareType>0</shareType>
     <shareWith>admin</shareWith>
     <userType>0</userType>
    </value>
   </element>
  </users>
  <groups>
   <element>
    <label>admin</label>
    <value>
     <shareType>1</shareType>
     <shareWith>admin</shareWith>
    </value>
   </element>
  </groups>
  <remotes/>
 </data>
</ocs>

C0rby:
@dpakach, we haven't changed anything about the response format/behavior so I guess it's intended.

Akidxx:
Hi, Regarding this issue (empty search parameter) the customer wrote me ;

Hi Akila,

See also
https://github.com/owncloud/core/issues/30313

Here it does say "One concern with 5 is that some code paths might call
the search functions with empty string. Maybe we should still allow
empty string."


Hello Akila,

There is something we are wondering.

As I have explained, I have made a small change to the code, to ensure
the API does not return any results when the search string is empty.

Now, in the original github ticket, there is a mention of possible
unexpected/unwanted behaviour when making this change.

On https://github.com/owncloud/core/issues/30313 it says:
"One concern with 5 is that some code paths might call the search
functions with empty string. Maybe we should still allow empty string."

We have not found anything to be not working with our change, but we are
wondering if you have some automated tests in place that might find it
if there is something broken after all.

It would be a matter of adding 6 lines of php to a single file and
running the tests to see if anything fails. Do you think that would be a
possibility?


Any idea regarding what she's asking about? Or we just need to tell her that the issue will be fixed in the ownCloud 10.7.0 version ?

Thanks!

C0rby:
@Akidxx,

The customer doesn't need to worry. Her change is not breaking anything.
What the comment
> "One concern with 5 is that some code paths might call the search
functions with empty string. Maybe we should still allow empty string."

is referring to is not the API layer (which the customer changed) but some underlying functions. If we would implement a minimal length there then it would cause some problems.

Yes, the fix should be included in ownCloud 10.7.0.

Issue #38500, opened 2021-03-12
API tests for case-sensitive tags


phil-davis:
PR #38498 adjusted the webUI so that tags are properly displayed case-sensitive - tags `test` and `Test` are different and should display as separate tags in the UI.
![Screenshot_2021-03-12 Files - ownCloud](https://user-images.githubusercontent.com/1535615/110891784-d1f7d300-831a-11eb-91e1-e0311b1cbcae.png)

Tags with different case should also be separate in the API.

Add some API test scenarios that verify that tags with different case really are processed as different.

individual-it:
@SwikritiT @phil-davis is this done?

phil-davis:
PR #38503 added some webUI test scenarios for tags that are only different in their case. So this is covered for the display on the webUI. And #38512 covers direct use of the API.

Done - closing.

PR #38512, opened 2021-03-15
[tests-only] Adds API tests to create tag in lowercase and uppercase

SwikritiT:
## Description
This PR adds test scenarios to create tags with names in lowercase and uppercase and proves that they're processed differently.

## Related Issue

- https://github.com/owncloud/core/issues/38500

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiTags/createTags.feature
+++ b/tests/acceptance/features/apiTags/createTags.feature
@@ -104,3 +104,27 @@ Feature: Creation of tags
     Given the administrator has created a "static" tag with name "StaticTag"
     When the administrator creates a "static" tag with name "StaticTag" using the WebDAV API
     Then the HTTP status code should be "409"
+
+  Scenario: Creating tags in lowercase and uppercase should work
+    When the administrator creates a "normal" tag with name "UppercaseTag" using the WebDAV API
+    Then the HTTP status code should be "201"
+    When the administrator creates a "normal" tag with name "lowercasetag" using the WebDAV API
+    Then the HTTP status code should be "201"
+    And the following tags should exist for the administrator
+      | name         | type   |
+      | UppercaseTag | normal |
+      | lowercasetag | normal |
+
+  Scenario: Creating tags with the same name in lowercase and uppercase should work
+    When the administrator creates a "normal" tag with name "testtag" using the WebDAV API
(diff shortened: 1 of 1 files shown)


Talank (commented):
LGTM

tests/acceptance/features/apiTags/createTags.feature
```diff
@@ -104,3 +104,26 @@ Feature: Creation of tags
...
+    When the administrator creates a "normal" tag with name "<tag_name>" using the WebDAV API
+    Then the HTTP status code should be "201"
+    And the following tags should exist for the administrator
+      | name       | type   |
+      | <tag_name> | normal |
+    Examples:
+      | tag_name |
+      | testtag  |
+      | Testtag  |
+      | TESTTAG  |
```
dpakach, line 129: I think a better scenario would be creating all these tags on single scenario not in outline

tests/acceptance/features/apiTags/createTags.feature
```diff
@@ -104,3 +104,24 @@ Feature: Creation of tags
...
+    And the following tags should exist for the administrator
+      | name         | type   |
+      | UppercaseTag | normal |
+      | lowercasetag | normal |
+
+  Scenario: Creating tags with the same name in lowercase and uppercase should work
+    When the administrator creates a "normal" tag with name "testtag" using the WebDAV API
+    And the administrator creates a "normal" tag with name "Testtag" using the WebDAV API
+    And the administrator creates a "normal" tag with name "TESTTAG" using the WebDAV API
+    Then the HTTP status code should be "201"
```
jasson99, line 121: Isn't this step asserting the status code of the third `when` step only?

jasson99: Either we need to check the status code and assert that the tag is created after every when stop, or should not check at all. I think it would be good to check after every when step to assert that the tags are actually created with lower and upper case.

jasson99 (commented):
LGTM

individual-it (approved):

PR #38532, opened 2021-03-16
fix downloadURL of public links

C0rby:
## Description
Make the downloadURL always relative to the owncloud base path.

## Related Issue

- https://github.com/owncloud/web/pull/4689#pullrequestreview-613291079

## Motivation and Context

If ownCloud is setup in a subdirectory like `https://example.com/owncloud/` then we don't wan't to include the subdirectory in the relative downloadURL.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
+++ b/apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
@@ -21,6 +21,7 @@
 
 namespace OCA\DAV\Files\PublicFiles;
 
+use OC;
 use OC\Files\Node\Folder;
 use OCA\DAV\Connector\Sabre\FilesPlugin;
 use OCP\Files\FileInfo;
@@ -168,8 +169,7 @@ class PublicFilesPlugin extends ServerPlugin {
 					} else {
 						$sharedResourcePath = '/' . $shareNode->getName();
 					}
-
-					$path = $server->getBaseUri() . $server->getRequestUri();
+					$path = \OC::$server->getURLGenerator()->getAbsoluteURL('/remote.php/dav/' . $server->getRequestUri());
 					// Let's assume we have this share
 					// - <shared folder>/
 					//   - subfolder/
--- a/changelog/10.7.0_2021-03-15/38376
+++ b/changelog/10.7.0_2021-03-15/38376
@@ -3,3 +3,4 @@ Enhancement: Implement pre-signed download urls for public links
 Added pre-signed download urls for password protected public links to support clients which don't use cookies.
 
 https://github.com/owncloud/core/pull/38376
+https://github.com/owncloud/core/pull/38532


kulmann (approved):
👍

micbar (approved):

jnweiger:
@C0rby when and how can the double prefix occur? I don't understand the context of https://github.com/owncloud/web/pull/4689#pullrequestreview-613291079

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
```diff
@@ -169,7 +169,7 @@ public function propFind(PropFind $propFind, INode $node) {
 						$sharedResourcePath = '/' . $shareNode->getName();
 					}
 
-					$path = $server->getBaseUri() . $server->getRequestUri();
+					$path = '/remote.php/dav/' . $server->getRequestUri();
```
jnweiger, line 172: This results in a partial URL starting with a /.
So it is not relative, but rooted. 
I'd normally object that, as it looks like it should break, when installing owncloud in a subdirectory.

But as you introduce that exactly to fix something with subdirectories, I assume you know what you are doing.

kulmann: Good spot. In fact the leading slash gets removed again in ownCloud Web. So it'd be fine to have a relative URL here.

jnweiger (changes requested):
I lack enough insights to review the fix itself.
It affects the behavior of the propFind() method. Not sure how widely used that method is. 

Please provide a clear example how to make a PROPFIND call that would benefit from that.
Is this a problem with new code, or is the same problem also present in released versions?

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
C0rby: > This results in a partial URL starting with a /.
So it is not relative, but rooted.

Yes, you're right.

> I'd normally object that, as it looks like it should break, when installing owncloud in a subdirectory.

It's the client responsibility to prepend the base path to this. Including the subdirectory if owncloud was intalled in one.
But as you commented here: https://github.com/owncloud/web/pull/4689#issuecomment-801107583
> Why do we go for flexibility? I'd argue that a full URL is more robust, as no errors can be made during URL-assembly.

I was thinking that maybe in a loadbalanced setup having the full URL could break things. After thinking about it again I don't think so anymore.
The instances would have to configure the `overwrite.cli.url` anyways right? So I could just use that as the base url and it should point to the load balancer/public address.

C0rby:
> I lack enough insights to review the fix itself.
> It affects the behavior of the propFind() method. Not sure how widely used that method is.
> 
> Please provide a clear example how to make a PROPFIND call that would benefit from that.
> Is this a problem with new code, or is the same problem also present in released versions?

It's not a problem in released versions. This PR fixes a problem which was introduced with a new feature for 10.7 #38376 

To test this you can send this PROPFIND request:
```
curl -s -u public:<share_password>  -k -X PROPFIND -d '<?xml version="1.0"?>
  <d:propfind  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns">
    <d:prop>
      <oc:permissions />
      <oc:favorite />
      <oc:fileid />
      <oc:owner-id />
      <oc:owner-display-name />
      <oc:share-types />
      <oc:privatelink />
      <d:getcontentlength />
      <oc:size />
      <d:getlastmodified />
      <d:getetag />
      <d:resourcetype />
      <oc:downloadURL />
    </d:prop>
  </d:propfind>' https://localhost:9200/remote.php/dav/public-files/rUenhLAqWpfHPXJ | xmllint --format -
  ```
  
  You have to do propfind requests for public shares with and without passwords.

apps/dav/lib/Files/PublicFiles/PublicFilesPlugin.php
C0rby: > Also note that legacy "/remote.php/webdav/" endpoint is still supported I think. It doesn't seem a good idea to mix both endpoints. I don't know if some webdav clients could throw errors if you change the endpoint; at least I would be suspicious

I see your point but will a generic webdav client even use this owncloud proprietary attribute?
Or am I wrong and this isn't a proprietary attribute?

Anyways this feature is especially implemented to solve a problem of webclients or more specifically of ownCloud web.
Our other clients don't even need this.

jnweiger:
Tested the patch in https://188.34.186.241/owncloud
Looks good. Thanks for the details!

jnweiger (approved):
Tested the patch in https://188.34.186.241/owncloud (Classic UI only!)
Looks good. Thanks!

Or maybe not: Could that be related? https://github.com/owncloud/core/issues/38543

C0rby:
@micbar @jnweiger, regarding putting the full URL into `downloadURL`. Can I assume that the config option `overwrite.cli.url` will always be set? If not I'd leave it at the rooted url.

Or is there any other reliable way to get the owncloud base path like `https://owncloud.example.com/` or `https://example.com/owncloud/`.

C0rby:
I found a way. And now the downloadURL is a full URL not a rooted/relative URL. So the handling for the frontend should be easier now. :)

Issue #36550, opened 2019-12-09
Behaviour for expiry date of user and group shares


phil-davis:
We have some questions about the required behaviour of the expiry date option for user and group shares (and the same principles should probably apply for public link shares)

1)  When the "expiry date" options are not enabled at all, and the sharing API is used to create or modify a share and specifies an expiry date, what should happen?
a) respond with a 4nn status code and do not do the action
~b) respond with a 2nn status code, create or modify the share using the other parameters provided, and just ignore the expiry date in the request~
~c) respond with a 2nn status code, create or modify the share including setting an expiry date on the share - but the expiry date will be ineffective (because later processing will never check the expiry date~
~d) or ???~
Note: public links seem to do (c) at the moment - so we can consider if their behaviour should be adjusted to (a)
See issue #36568 

2) When the "expiry date" options are enabled, and there is a default expiry date, and a sharing API request comes without explicit mention of an expiry date then:
~a) the default expiry date should be calculated and set, and the response will include an expiry date attribute (that was not in the request)~
b) do not set an expiry date (unless it is enforced)

3) When the "expiry date" options have been enabled for a while, and there are share in the system with expiry dates. Then the admin disables the expiry date options:
a) any expiry dates on existing shares continue to be enforced (but are not displayed in the UI, or in API requests for share info)
b) any expiry dates on existing shares continue to be displayed and enforced
~c) expiry dates on existing shares "hang around" in the database but are not displayed or enforced any more (they "hang around" so that if the admin clicks the "expiry date" option on again in 10 minutes, then the existing expiry will not all be lost)~
~d) expiry dates on existing shares are all removed permanently (so they do not come back when the "expiry date" option is enabled again~

4) When admin reduces the enforced maximum expiry date:
a) existing shares with expiry dates beyond the new maximum stay as they are
~b) existing shares with expiry dates beyond the new maximum get their expiry date reduced~

5) If the admin sets the default expiry date to 0 days (and they could enforce that also):
~a) it means shares created today will expire at the end of the day~
b) 0 is not allowed - see issue #36569 
~c) ???~

phil-davis:
@pmaier1 @micbar or anyone.

We are producing the sharing API acceptance tests and would like to be able to decide which test results are the required behaviour, and which are "not quite correct"

pmaier1:
The expiration date feature can be turned on/off? I didn't know.

[...]

PR #38540, opened 2021-03-17
Fix expiration date

micbar:
## Description

Shares will expire now at the end of the day. This was a bit undeterministic before, depending on the background job execution time. Now it will not expire during the day but at the first possible moment in time after the end of the day. The exact expiration time will still depend on the scheduling of the background job.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4454
- Fixes https://github.com/owncloud/enterprise/issues/4324

## Motivation and Context

The expiration of shares was a bit undeterministic before

## How Has This Been Tested?

- manually
- Ci

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/tests/ApiTest.php
+++ b/apps/files_sharing/tests/ApiTest.php
@@ -1153,10 +1153,10 @@ class ApiTest extends TestCase {
 		$config->setAppValue('core', 'shareapi_enforce_expire_date', 'yes');
 
 		$dateWithinRange = new \DateTime();
-		$dateWithinRange->setTime(0, 0, 0);
+		$dateWithinRange->setTime(23, 59, 0);
 		$dateWithinRange->add(new \DateInterval('P5D'));
 		$dateOutOfRange = new \DateTime();
-		$dateOutOfRange->setTime(0, 0, 0);
+		$dateOutOfRange->setTime(23, 59, 0);
 		$dateOutOfRange->add(new \DateInterval('P8D'));
 
 		// update expire date to a valid value
@@ -1533,7 +1533,7 @@ class ApiTest extends TestCase {
 		$this->assertEquals($url, $data['url']);
 
 		$share = $this->shareManager->getShareById('ocinternal:'.$data['id']);
-		$date->setTime(0, 0, 0);
+		$date->setTime(23, 59, 0);
 		$this->assertEquals($date, $share->getExpirationDate());
 
 		$this->shareManager->deleteShare($share);
--- a/changelog/unreleased/38540
+++ b/changelog/unreleased/38540
@@ -1,0 +1,6 @@
+Bugfix: Expire shares not before the end of the day
+
+Shares will expire now at the end of the day. This was a bit undeterministic before, depending on the background job execution time. Now it will not expire during the day but at the first possible moment in time after the end of the day. The exact expiration time will still depend on the scheduling of the background job.
+
+
+https://github.com/owncloud/core/pull/38540
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -430,11 +430,11 @@ class Manager implements IManager {
 
 		if ($expirationDate !== null) {
 			// Set the expiration date to just the date at "zero" time in the day
-			$expirationDate->setTime(0, 0, 0, 0);
+			$expirationDate->setTime(23, 59, 0, 0);
 
 			// Get the current date in the same timezone, and at "zero" time in the day
 			$date = new \DateTime('now', new DateTimeZone($expirationDate->getTimezone()->getName()));
-			$date->setTime(0, 0, 0, 0);
+			$date->setTime(23, 59, 0, 0);
 
 			if ($date > $expirationDate) {
 				$message = $this->l->t('Expiration date is in the past');
@@ -473,7 +473,7 @@ class Manager implements IManager {
 			// If expiredate is empty and it is a new share, set a default one if there is a default
 			if ($this->isNewShare($share) && $expirationDate === null && $thereIsDefault) {
 				$expirationDate = new \DateTime();
-				$expirationDate->setTime(0, 0, 0);
+				$expirationDate->setTime(23, 59, 0);
 				$expirationDate->add(new \DateInterval('P'.$defaultDays.'D'));
 			}
 
@@ -482,7 +482,7 @@ class Manager implements IManager {
 			}
 
 			$date = new \DateTime();
-			$date->setTime(0, 0, 0);
+			$date->setTime(23, 59, 0);
 			$date->add(new \DateInterval('P' . $defaultDays . 'D'));
 			if ($date < $expirationDate) {
 				$message = $this->l->t('Cannot set expiration date more than %s days in the future', [$defaultDays]);
--- a/tests/lib/Share20/ManagerTest.php
+++ b/tests/lib/Share20/ManagerTest.php
@@ -1029,7 +1029,7 @@ class ManagerTest extends \Test\TestCase {
 	 */
 	public function testValidateExpirationDateToday() {
 		$today = new \DateTime();
-		$today->setTime(0, 0, 0);
+		$today->setTime(23, 59, 0);
 		$share = $this->manager->newShare();
 		$share->setExpirationDate($today);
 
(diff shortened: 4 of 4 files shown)


mmattel:
Even draft, docs relecant because of a change in config.sample
Add a proper text (see examples in configs above) and describe possible values
Pls file an issue when close to merge

micbar:
> Even draft, docs relecant because of a change in config.sample
> Add a proper text (see examples in configs above) and describe possible values
> Pls file an issue when close to merge

I removed the config change here, Will do a separate PR

karakayasemi (approved):
LGTM!

AlexAndBear (approved):

micbar:
hmm, I am a bit hesitating to merge now.

1) We never noticed the wrong behavior before
2) How do we know that this is fixed now? The unit tests alone are not proving it

I know that this is very hard to test, but could we do anythig here? @phil-davis 

Maybe introduce a way to manipulate the expiration date during the acceptance tests run, to "simulate" the passing time?

phil-davis:
I assigned myself to issue https://github.com/owncloud/enterprise/issues/4324 to look into the testing.

micbar:
@phil-davis What can i do next here in the PR? rebase on a different branch? Add some tests?

phil-davis:
I found where the code checks if `shareHasExpired` - see comments at https://github.com/owncloud/core/issues/36550#issuecomment-847666569

Note: I found the PR #36573 "Include expiration option for user and group shares" - that has some good comment history abut this stuff. And the code that makes shares expire at the end of the day is https://github.com/owncloud/core/blob/master/lib/private/Share20/Manager.php#L185
```
	private static function shareHasExpired($share) {
		$expirationDate = $share->getExpirationDate();
		return ($expirationDate !== null) && ($expirationDate < new \DateTime("today"));
	}
```

The logic is to get the DateTime of "today". That is 00:00 at the beginning of today. If the share expired before that time (that is, at any time on earlier day/s) then it really has expired. A share that expires any time today (even at 00:00 today) will still be valid.

This is called by lots of methods in https://github.com/owncloud/core/blob/master/lib/private/Share20/Manager.php whenever shares are looked up. If the share has expired, then it gets deleted "on the fly". So the share becomes unavailable immediately at 00:00 on the day after the day that it is to expire. You do not have to wait for the background job to run.

Q: So why do we have a background job?
A: Maybe it is useful to delete expired shares that nobody tries to access (=database cleanup), otherwise they stay in the database forever waiting for someone to attempt to access them.

Q: Is the background job already using the logic from `shareHasExpired`?
If yes, then this should already be working fine.
If no, then the background job might be deleting "expired" shares too early. It should be fixed to use the logic from `shareHasExpired`.

The other issue is the "cosmetic" response like `"expiration":"2021-03-05 00:00:00"` - actually we want clients that receive this response to understand that the share expires at the end of that day. What is the best way to respond:
1) `"expiration":"2021-03-05 23:59:59"` or
2) `"expiration":"2021-03-06 00:00:00"` (the very beginning of the next day)
3) `"expiration":"2021-03-05"` (leave out the time)

Actually (2) is more accurate. The share is still available at `2021-03-05 23:59:59.5`, past the DateTime (1).
(3) is maybe also a thing to do. Do not specify a time?

If we do (2) then UI clients that display the "share expiration date" will need to think to subtract 1 day when displaying to the user. From the regular user's point-of-view this share expires at the end of 2021-03-05, `2021-03-05 23:59:59.999999999..."

phil-davis:
See #38775 - IMO that fixes the real problem, that the `ExpireSharesJob` is using the wrong logic.

Then we can separately consider what to do about the `"expiration":"2021-03-05 00:00:00"` time "zero" in the response.

micbar:
@phil-davis I approved your PR #38775

micbar:
> Even draft, docs relecant because of a change in config.sample
> Add a proper text (see examples in configs above) and describe possible values
> Pls file an issue when close to merge

We removed the change in the config. Not necessary.

karakayasemi (approved):
Looks good to me!

phil-davis:
I was a little surprised that the acceptance tests still pass. So  I tried running one and logging the request-responses:
`make test-acceptance-api BEHAT_FEATURE=tests/acceptance/features/apiShareCreateSpecialToRoot1/createShareExpirationDate.feature:49`

```
### AUTH: Brian:1234
### REQUEST: GET http://172.17.0.1:8080/ocs/v1.php/apps/files_sharing/api/v1/shares/3
Headers:
Host: 172.17.0.1:8080
OCS-APIREQUEST: 1
Content-Type: application/x-www-form-urlencoded
Body:
string(0) ""
```

```
### RESPONSE
Status: 200
...
Body:
string(1163) "<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data>
  <element>
   <id>3</id>
   <share_type>0</share_type>
   <uid_owner>Alice</uid_owner>
   <displayname_owner>Alice Hansen</displayname_owner>
   <permissions>17</permissions>
   <stime>1623066610</stime>
   <parent/>
   <expiration>2021-06-22 00:00:00</expiration>
   <token/>
   <uid_file_owner>Alice</uid_file_owner>
   <displayname_file_owner>Alice Hansen</displayname_file_owner>
   <additional_info_owner/>
   <additional_info_file_owner/>
   <path>/FOLDER</path>
   <mimetype>httpd/unix-directory</mimetype>
   <storage_id>shared::/FOLDER</storage_id>
   <storage>31</storage>
   <item_type>folder</item_type>
   <item_source>2147483976</item_source>
   <file_source>2147483976</file_source>
   <file_parent>2147483974</file_parent>
   <file_target>/FOLDER</file_target>
   <share_with>Brian</share_with>
   <share_with_displayname>Brian Murphy</share_with_displayname>
   <share_with_additional_info/>
   <mail_send>0</mail_send>
   <attributes/>
  </element>
 </data>
</ocs>
"

### END RESPONSE
```

phil-davis:
The test creates a share with expiration date 15 days in the future. The response has:
`<expiration>2021-06-22 00:00:00</expiration>`

But with this PR I was expecting to see:
`<expiration>2021-06-22 23:59:00</expiration>`

phil-davis:
@micbar IMO the code changes here are just internal to the back-end processing. They don't result in any externally-observabke differences. The date-time comparisons work the same when comparing 00:00 on the relevant days or 23:59 on the relevant days.

PR #38813 changes the reported expiration time to 23:59. Please have a look at that - maybe that is something to do for issue https://github.com/owncloud/enterprise/issues/4454 ?

phil-davis:
@micbar the expiration date in share responses now specifies the expiration time 23:59:59 - PR https://github.com/owncloud/core/pull/38813

IMO  this PR is no longer needed. Close?

micbar:
Closing

jnweiger:
> @micbar the expiration date in share responses now specifies the expiration time 23:59:59 - PR #38813
> 
> IMO this PR is no longer needed. Close?

This should be obsoleted by #38813 , OK
But in #38856 we reverted the fix again?

@phil-davis @micbar are we sure the issue is actually fixed?
In manual QA @gabi18 could not reproduce any misbehaviour though, ... leaving me a bit puzzled :-)

phil-davis:
Actually, there was not a real problem. When a share receiver accesses a share, and if the share has an expiration date, then the expiration date is checked on-the-fly by the back-end server code. If the date is past the expiration date, then the back-end server code immediately deletes the share, and returns a "not found" status.

Some time during the day, the "cron" job runs. That finds any shares that have passed their expiration date and deletes those shares. Each day, that will find shares that expired yesterday, and have not been accessed yet today, and delete them. If the cron job never ran, then expired shares that never get accessed again would sit around in the database (but do no real harm). So the cron job is a bit of a cleanup tool, but not 100% essential.

The cron job was deleting shares too early - it would delete shares that expire "today". That was fixed in PR #38775 - so that is the actual fix for the inconsistent expiry "time" of shares.

The `00:00:00` vs `23:59:59` thing is purely cosmetic. All the back-end code "ignores" the time, and only expires shares AFTER the date has finished. It is unfortunate that the share response has that `00:00:00` specified, because that makes people/clients think that the share expires at the start of the day. We wanted to change that, but the desktop client (and maybe other clients?) has that `00:00:00` hard-coded in it - so we could not change the server response.

PR #38542, opened 2021-03-17
when resharing with group that user belongs permissions are most permissive already in mount provider

mrow4a:
fixes https://github.com/owncloud/enterprise/issues/4382

This PR solved the issue where user is resharing with group that himself belongs to, and we should not check for more permissive permissions/attributes in supershare as these are most permissive already .

- [x] manual tests
- [ ] add unit and integration tests


mrow4a:
In this PR I tried to fix the behaviour and adjust logic of permission handling/mounting. in https://github.com/owncloud/core/pull/38581 I fixed root cause of the issue.

PR #38550, opened 2021-03-19
[Tests-Only] removed phpstan skip

kiranparajuli589:
## Description
With this PR:
- redundant phpstan ignore is removed from `legacy/utils.php`

Fixes error seen in recent drone CI
https://drone.owncloud.com/owncloud/core/29311

## Related Issue

- Fixes https://github.com/owncloud/core/issues/38551

## How Has This Been Tested?

- :robot:

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38556, opened 2021-03-20
[tests-only] Trigger CI for release 10.7.0

phil-davis:

PR #38557, opened 2021-03-20
[10.7] [tests-only] removed phpstan skip

phil-davis:
Backport #38550 to `release-10.7.0`

A new release of `phpstan` https://github.com/phpstan/phpstan/releases/tag/0.12.82 is improved and no longer reports this code as a problem.

We removed the `phpstan-ignore-next-line` annotation in `master`. We need to remove it in `release-10.7.0` also, so that CI will pass here.


phil-davis:
#38556 demonstrates the problem in the `release-10.7.0` branch.

individual-it (approved):

PR #38561, opened 2021-03-22
[tests-only] Trigger CI

phil-davis:
`make test-php-style` seems to be failing. Get a CI result...


phil-davis:
code style fails https://drone.owncloud.com/owncloud/core/29339/2/7

PR #38591, opened 2021-03-28
if source and target owners are different, check quota on move

karakayasemi:
## Description
If source and target owners are different, check free space in dav quota plugin

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4427

## Motivation and Context
Fixing bugs

## How Has This Been Tested?
Manually tested below scenarios
- Create two users `user1` and `user2`
- Set quota for `user1` to unlimited, `user2` to 0
- Share a folder from `user1` to `user2`
- Upload a file from `user1` to shared folder, it should be successful
- `user1` moves the uploaded file to home folder, it should be successful
- Upload a file from `user1` to shared folder again, it should be successful
- When `user2` moves file to home folder, it should not be succesful and `user2` should show `Not enough free space` notification

Tried same scenario for chunked files also.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
@@ -80,27 +80,38 @@ class QuotaPlugin extends \Sabre\DAV\ServerPlugin {
 	/**
 	 * Check if we're moving a Futurefile in which case we need to check
 	 * the quota on the target destination.
+	 * If target and source storages are different, quota plugin should check
+	 * free space in target storage.
 	 *
 	 * @param string $source source path
 	 * @param string $destination destination path
 	 */
 	public function handleBeforeMove($source, $destination) {
 		$sourceNode = $this->server->tree->getNodeForPath($source);
-		if (!$sourceNode instanceof FutureFile) {
-			return;
-		}
 
 		// get target node for proper path conversion
 		if ($this->server->tree->nodeExists($destination)) {
-			$destinationNode = $this->server->tree->getNodeForPath($destination);
-			'@phan-var Node $destinationNode';
-			$path = $destinationNode->getPath();
+			$targetNode = $this->server->tree->getNodeForPath($destination);
 		} else {
-			$parentNode = $this->server->tree->getNodeForPath(\dirname($destination));
-			'@phan-var Node $parentNode';
-			$path = $parentNode->getPath();
+			$dirname = \dirname($destination);
+			$dirname = $dirname === '.' ? '/' : $dirname;
+			$targetNode = $this->server->tree->getNodeForPath($dirname);
 		}
 
+		// Check quota for only FutureFile or
+		// target storage different than source storage
+		if (!$sourceNode instanceof FutureFile) {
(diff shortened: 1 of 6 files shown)


phil-davis:
I added issue #38592 to get some end-to-end test scenarios added. Those can be done in a later PR, if this PR is merged quickly.

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
```diff
@@ -80,27 +80,35 @@ public function initialize(\Sabre\DAV\Server $server) {
...
+			$targetNode = $this->server->tree->getNodeForPath($dirname);
 		}
 
+		// Check quota for only FutureFile or
+		// target Node owner different than source Node owner
+		if (!$sourceNode instanceof FutureFile) {
+			if (!$sourceNode instanceof Node || !$targetNode instanceof Node) {
+				return;
+			}
+			if ($sourceNode->getOwner()->getUID() === $targetNode->getOwner()->getUID()) {
```
jvillafanez, line 107: Although unlikely, it might be possible that 2 different storages have the same owner.

VicDeo: true. This case needs to be validated. e.g. move from local to external storage

karakayasemi: We are handling these situations in storage wrappers, but I will try to cover in here also.

phil-davis:
Note: I made some review comments about the tests in #38599 - when those are sorted out, the commit from that PR can be cherry-picked into here...

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
karakayasemi: The described problem is also valid for external storages. Copying from external storage to different storage when encryption active and target storage leads to broken copy on target storage if target has not enough quota. 

I have covered this scenario in QuotaPlugin as you suggested.

jvillafanez: I'd prefer to just check if the storage ids are different, as done lines below. I don't think we need to check the owners.

karakayasemi: > 
> 
> I'd prefer to just check if the storage ids are different, as done lines below. I don't think we need to check the owners.

I agree. I have changed the code as you suggested.

phil-davis (commented):
Tests look good now. dev review needed.

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
```diff
@@ -23,6 +23,7 @@
  *
  */
 namespace OCA\DAV\Connector\Sabre;
+use OC\Files\Storage\Home;
```
jvillafanez, line 26: not used

jvillafanez (approved):
just a tiny detail

PR #38605, opened 2021-03-31
Automations in activity stream

AlexAndBear:
## Description
This adjustment provides a way to handle the activity stream for actions that were triggered by an automation (like the workflow app e.g.). As the workflow app currently only handles `delete` and `assign_tag`, this is the only action we take care of for now.

The changes go together with implementations in the workflow and the activity app, although they do not rely on them.

* https://github.com/owncloud/workflow/pull/255
* https://github.com/owncloud/activity/pull/942

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4222

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


lib/public/Activity/DataProvider.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCP\Activity;
+
+/**
+ * Class Constants
```
JanAckermann, line 26:
```suggestion
 * Class DataProvider
```

lib/public/Activity/DataProvider.php
```diff
@@ -0,0 +1,63 @@
...
+namespace OCP\Activity;
+
+/**
+ * Class DataProvider
+ *
+ * @package OCP
+ * @since 10.8.0
+ */
+
+class DataProvider {
```
jvillafanez, line 32: This seems too suspicious to me.
* No reference inside core repo. This makes it a candidate to be removed as dead code. In addition, since there is no reference about where this is used, we'll have to check all the apps.
* So far, this seems to be just a global variable.

If there is no other approach to do what you want, make sure the intentions are crystal clear. Right now, just looking at this class, I don't know what you're trying to do with this, and I'm more inclined to remove this class.

apps/systemtags/lib/Activity/Extension.php
```diff
@@ -41,6 +41,7 @@ class Extension implements IExtension {
 	const DELETE_TAG = 'delete_tag';
 
 	const ASSIGN_TAG = 'assign_tag';
+	const ASSIGN_TAG_BY_AUTOMATION = 'assign_tag_by_automation';
```
jvillafanez, line 44: I guess this is an implicit dependency with the workflow app. Did you check if it's possible to move it completely to the workflow app?

jvillafanez:
It seems that whether the activity comes from a user or from something automated should be bound to the event itself. The `$event->setAuthor(...)` could be an idea, and it seems more consistent.

I guess the problem of using `$event->setAuthor(null)` is with oracle storing an empty string in the DB, which could cause problems. I'm not sure if we can store something like "user:myuser" or "auto:cli" in the DB so we can distinguish these cases.
We can have some constants so we can do `$event->setAuthor(IEvent::AUTHOR_CLI)` and `$event->getAuthor() === IEvent::AUTHOR_CLI` so it's a bit easier to deal with this from the outside.
I'm not sure if it will be easy to keep compatibility, but I think it's worth a shot.

lib/public/Activity/DataProvider.php
JanAckermann: @jvillafanez ping

JanAckermann:
@jvillafanez This was our idea at first, but the event ist triggered inside for example in the node->delete function, we could assign the delete function a metaObject paramerer, which will be passed trough the event, BUT this needs a change to the Interface and all classes which implements it (and this means required changes in a Lot of the Apps)
Any Ideas Welcome!

jvillafanez:
The problem with the tags seems to be in https://github.com/owncloud/core/blob/master/apps/systemtags/lib/Activity/Listener.php#L88-L93 . Assuming that the author is the user in the session is wrong. We'll likely need to change the `ManagerEvent` and get the author from there. We can assume the author is the user in the session only as a fallback.
I haven't traced all the code, so there could be more issues to go around.

FS actions seem to be more complex since they're using the old hooks and not the symfony's events, so we can't enrich the events with additional information.

----

An alternative to explore could be around https://github.com/owncloud/core/blob/eebd754771a99bc44820eb67a6d57cea238e6b86/lib/private/Activity/Manager.php#L178-L182 .
We could add a `setDefaultAuthor(....)` in the activity manager. Whenever someone wants to publish an activity without an explicit user, we can get the default author and override the event.
The idea is to set the default author to the session user by default in order to keep compatibility. Now, for our case, we could change this default author to a different one, perform the action, and then restore the old default author. We'll have to check if this works or not.

We'll still need something like https://github.com/owncloud/core/pull/38605#issuecomment-811044120 mainly to prevent collisions with valid usernames.

JanAckermann:
@jvillafanez we save now the automated user in the database which works fine, the thing is we did not understand how you want to achieve saving the user not by the session (we still go by the global var, just changed the location), can you create a PR and show us, what you mean exactly?

lib/private/Activity/Manager.php
```diff
@@ -175,6 +178,13 @@ public function publish(IEvent $event) {
 			throw new \BadMethodCallException('Subject not set', 13);
 		}
 
+		if (self::$defaultAuthor) {
```
jvillafanez, line 181: This should be checked if there is no author explicitly set in the event. If the event already has an explicit author, we shouldn't override it.

jvillafanez:
> @jvillafanez we save now the automated user in the database which works fine, the thing is we did not understand how you want to achieve saving the user not by the session (we still go by the global var, just changed the location), can you create a PR and show us, what you mean exactly?

We're overriding the author in the `publish` method if the event doesn't provide an author. Instead of using the global variable, we just set the default author in the activity manager, perform the action, and then restore the previous default author. The idea is roughly the same of using the global variable, but without using a global variable.
If the default author is null, we can still override the author with the user in the session, with the same rules. This will keep the code compatible with what we have now.

JanAckermann:
@jvillafanez the automated tagging provides the session with the current user, so we still have a user in the session

JammingBen:
@jvillafanez Ok we implemented your suggestions, please have a look again. We renamed `defaultAuthor` to `overwriteAuthor` because the term "default" is kinda misleading here IMO. It's more that we overwrite an existing author with this implementation. Also, we now set the actor in apps/systemtags/lib/Activity/Listener.php. This way we do not need to handle the subject separately.

jvillafanez:
I'm a bit confused with the expected usage. I expected the `setOverwriteAuthor` to overwrite the author inside the `publish` method so the apps can get the author normally. If it will be handled separately, as a secondary author or something like that, then we aren't overwriting it; it doesn't seem a good name.
I'm ok with the solution, but this needs to be clarified.

apps/systemtags/lib/Activity/Listener.php
```diff
@@ -172,12 +173,17 @@ public function mapperEvent(MapperEvent $event) {
...
 
 		$actor = $this->session->getUser();
+
 		if ($actor instanceof IUser) {
 			$actor = $actor->getUID();
 		} else {
 			$actor = '';
 		}
 
+		if ($this->activityManager->getOverwriteAuthor()) {
```
jvillafanez, line 183: no need to call the `getOverwriteAuthor` method twice.

apps/systemtags/lib/Activity/Extension.php
```diff
@@ -216,6 +223,20 @@ protected function actorIsCurrentUser($user) {
...
+	/**
+	 * Check if the author is automation user
+	 *
+	 * @param string $user Parameter e.g. `<user display-name="admin">admin</user>`
+	 * @return bool
+	 */
+	protected function actorIsAutomation($user) {
+		try {
+			return \strip_tags($user) === Event::AUTOMATION_AUTHOR;
+		} catch (\UnexpectedValueException $e) {
```
jvillafanez, line 235: Is this exception ever thrown? Documentation doesn't say anything

apps/systemtags/lib/Activity/Extension.php
```diff
@@ -216,6 +223,20 @@ protected function actorIsCurrentUser($user) {
...
 
+	/**
+	 * Check if the author is automation user
+	 *
+	 * @param string $user Parameter e.g. `<user display-name="admin">admin</user>`
+	 * @return bool
+	 */
+	protected function actorIsAutomation($user) {
+		try {
+			return \strip_tags($user) === Event::AUTOMATION_AUTHOR;
```
jvillafanez, line 234: The constant is defined in the public `IEvent` interface, so we should use that. Note that, unless there is no other option, apps should only access to public classes / interfaces.

lib/public/Activity/IManager.php
```diff
@@ -223,4 +223,20 @@ public function setCurrentUserId($currentUserId);
...
+	 */
+	public function getOverwriteAuthor();
+
+	/**
+	 * Set the author which will be used to override an existing author.
+	 *
+	 * @param string
+	 * @since 10.8.0
+	 */
+	public function setOverwriteAuthor($overwriteAuthor);
```
jvillafanez, line 241: The PHPDoc should reference to the `IEvent` constants since that's the main use of this method. It could also be useful to clarify if any random string could be used, or if the string should match with a valid existing user id.
Similar information should also be documented in the `getOverwriteAuthor` method to clarify the possible values, specially if some of them should have special meaning, and if the apps should check by themselves the existence of the user (if they want)

JammingBen:
We could call `setOverwriteAuthor()` inside the `publish` method. But we need to adjust the subject params as well, because the author usually sits there, too. The problem here is: subject params is an array and the position of the author varies based on the subject. So we would need to add some more logic to identify what subject we're using and at what position the author param is. Hence we decided to handle such things a little earlier in `Listener.php`.

I get your point with the naming though. Now we could set `overwriteAuthor` without actually overwriting it, because it needs to be handled separately - that's confusing. Do you have a proposal for a better name? I mean, we could also set the author in the `publish` method again. But we still need to handle the subject params in `Listener.php`.

jvillafanez:
> Do you have a proposal for a better name?

`setSecondaryAuthor` or `setAgentAuthor` could fit. Anyway, it must be properly documented in the code.

jvillafanez:
Since we still have the same problem with the cli, it would be interesting to have some plans on how it's expected to be implemented. I'm not saying that it has to be implemented now, but I think it should fit within the current solution, so having some words on this would help to implement the solution for the cli when the time comes.

JammingBen:
@jvillafanez I renamed the methods and tried to explain the usage in the interface docs.

apps/files/lib/Activity.php
```diff
@@ -422,4 +426,18 @@ public function getQueryForFilter($filter) {
...
+	/**
+	 * Check if the author is automation user
+	 *
+	 * @param string $user Parameter e.g. `<user display-name="admin">admin</user>`
+	 * @return bool
+	 */
+	protected function actorIsAutomation($user) {
+		try {
+			return \strip_tags($user) === IEvent::AUTOMATION_AUTHOR;
+		} catch (\UnexpectedValueException $e) {
```
jvillafanez, line 439: This isn't thrown here.

apps/systemtags/lib/Activity/Listener.php
```diff
@@ -22,6 +22,7 @@
 
 namespace OCA\SystemTags\Activity;
 
+use OC\Activity\Event;
```
jvillafanez, line 25: Use public `IEvent` instead

jvillafanez (commented):
Couple of things and I think we're on this side

jvillafanez:
After checking the PR for the workflow app, I've remember why I wanted the bind the "agent" author to the event: I expect a usage such as:
```
$activityManager->setAgentAuthor(IEvent::AUTOMATION_AUTHOR);
doSomething();
$activityManager->setAgentAuthor(null);
```
This way, the scope is clear and limited to the `doSomething` method.

The problem I see with the current approach is that we lose control over when we can reset the agent. Other actions that could happen after that `doSomething` might be misinterpreted as being automated because we didn't reset the agent. In addition, I don't think it's a good idea to rely on the activity app to reset the agent.

JammingBen:
I honestly don't see a way to keep `$activityManager->setAgentAuthor(IEvent::AUTOMATION_AUTHOR);` and `$activityManager->setAgentAuthor(null);` together. Let's look at the automated delete:

We have to set `$agentAuthor` to `IEvent::AUTOMATION_AUTHOR` within our workflow app: https://github.com/owncloud/workflow/blob/87066fdef98a94afd7a624d7c0d51f68c37fc5ee/lib/Retention/TagBasedRetention.php#L295. This is the beginning of our "trace". There is no other place to do that besides passing a bunch of parameters to `$node->delete()`. Setting the agent author back to null immediately afterwards also wont work as it would be null by the time the event is being handled.

Binding the agent author to the event also does not work because we don't have an event yet at this point.

jvillafanez:
> Binding the agent author to the event also does not work because we don't have an event yet at this point.

When a new event is published, the activity manager should override the author of the event with the agent. Basically, with the `setAgentAuthor` you tell the activity manager to set that author for any event that will be published, so you can restore the previous agent just after the `$node->delete()`. Basically, the workflow app just hints the activity manager, but it's the activity manager the one binding the agent with the event.
Did we reject this idea? There was some code previously, but I think we just shifted the approach.

JammingBen:
Hmm yeah, restoring the agent directly after `$node->delete()` works. Strange, I'm pretty sure we ran into other issues when we tried this first 🤔  Although I'm slowly getting confused about all the changes and versions we had...

Please have a look again. This PR did not change, but those for workflow and activity did. Now we've properly wrapped the calls this way:

```
$activityManager->setAgentAuthor(IEvent::AUTOMATION_AUTHOR);
doSomething();
$activityManager->setAgentAuthor(null);
```

Reseting the agent author in the activity app was removed therefore.

jvillafanez:
We might need to include a stack or similar to support nested contexts. Expected usage could be:
```
$activityManager->setAgentAuthor('cli');
doSomething();
$activityManager->setAgentAuthor('automation');
doAnotherThing();
$activityManager->restoreAgentAuthor();
doAthirdThing();
$activityManager->restoreAgentAuthor();
```
Both `doSomething` and `doAThirdThing` method should be authored by the cli, while the `doAnotherThing` should be authored by the automation. Note that the calls will likely be in different places.

A valid example could be a command in the workflow app. We'll likely have to add the cli as author at some point

JammingBen:
@jvillafanez Could you get us a hint on how to achieve that?

jvillafanez:
You can have an array and use it as a stack. That should be easy to implement if we don't want to create a class just for this:
```
private $agentStack = [];

public function setAgentAuthor($author) {
  $this->agentStack[] = $author;
}

public function getAgentAuthor() {
  return $this->agentStack[\count($this->agentStack) - 1];
}

public function restoreAgentAuthor() {
  return \array_pop($this->agentStack);
}
```
I'm pretty sure there are some things to take care, such as getting or restoring an agent when the stack is empty, but that should be most of it. This should be transparent to the apps.
You can also document the functions properly to make the code as easy as possible. For example, the `restoreAgentAuthor` could return the last agent set (after resetting the agent) or null if there is no agent set (because `array_pop` will return null if the array is empty)

JammingBen:
Alrighty, I applied your suggestion. Also updated the docs in the interface regarding its usage.

JanAckermann:
@jvillafanez can you have a Look again about the whole implementation, we should really come to an end here.

lib/private/Activity/Manager.php
```diff
@@ -529,4 +533,25 @@ protected function getUserFromToken() {
 		// Token found login as that user
 		return \array_shift($users);
 	}
+
+	/**
+	 * {@inheritdoc}
+	 */
+	public function getAgentAuthor() {
+		return $this->agentStack[\count($this->agentStack) - 1] ?? null;
```
jvillafanez, line 541: Does it show a php warning in the logs if the stack is empty? I expect an undefined index... maybe I'm wrong, or it might depend on the php version

lib/public/Activity/IManager.php
```diff
@@ -223,4 +223,46 @@ public function setCurrentUserId($currentUserId);
...
+	 * used via restoreAgentAuthor().
+	 *
+	 * @param string
+	 * @since 10.8.0
+	 */
+	public function setAgentAuthor($agentAuthor);
+
+	/**
+	 * Restore the latest agent user and return it. Returns null if the
+	 * $agentStack array is empty.
```
jvillafanez, line 262: it returns a string, but we need to clarify if it returns the "previous old" (the one you could get before calling this method), or the current agent in the top of the stack.
As it's implemented now, it returns the agent you're poping. This is fine, but we should document this.

jvillafanez:
Assuming it works, I guess those are the last changes here.

lib/private/Activity/Manager.php
JammingBen: It does not show a warning for me. `??` seems to be fully supported since PHP 7.

jvillafanez: ok, let's leave it like that.

jvillafanez:
Do we want to merge this PR now, or do we freeze it until the rest of the pieces are ready? I don't think we'll need to make changes here, but you never know

jvillafanez:
BTW, we should include some unit tests for these changes. The ones for the agent stuff should be pretty easy.

JanAckermann:
Rest of the pieces? This PR needs to merged to get CI Green for activity and Workflow. As well another issue ist waiting, which relies on this

jvillafanez:
> Rest of the pieces? This PR needs to merged to get CI Green for activity and Workflow. As well another issue ist waiting, which relies on this

The PR for the workflow app should use the `restoreAgentAuthor` instead of setting a null agent. That's why I'm asking.

I'm ok with merging this.

jvillafanez (approved):

JammingBen:
Added a simple unit test for setting, getting and restoring an agent author and also adjusted the workflow app.

PR #38631, opened 2021-04-09
Translation in the activity stream for shares which expired automatically

JammingBen:
## Description
Add a proper translation in the activity stream for shares which expired automatically. Previous to this, the expiry was authored by a user, which is technically not true.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4455

image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/ExpireSharesJob.php
+++ b/apps/files_sharing/lib/ExpireSharesJob.php
@@ -70,7 +70,7 @@ class ExpireSharesJob extends TimedJob {
 		ActivityIManager $activityManager
 	) {
 		// Run once a day
-		$this->setInterval(5);
+		$this->setInterval(24 * 60 * 60);
 		$this->shareManager = $shareManager;
 		$this->connection = $connection;
 		$this->defaultShareProvider = $defaultShareProvider;
--- a/apps/files_sharing/lib/ExpireSharesJob.php
+++ b/apps/files_sharing/lib/ExpireSharesJob.php
@@ -28,6 +28,7 @@ use OCP\Activity\IEvent;
 use OCP\IDBConnection;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Share\IManager;
+use OCP\Activity\IManager as ActivityIManager;
 
 /**
  * Delete all shares that are expired
@@ -50,7 +51,7 @@ class ExpireSharesJob extends TimedJob {
 	private $defaultShareProvider;
 
 	/**
-	 * @var \OCP\Activity\IManager $activityManager
+	 * @var ActivityIManager $activityManager
 	 */
 	private $activityManager;
 
@@ -60,16 +61,16 @@ class ExpireSharesJob extends TimedJob {
 	 * @param IManager $shareManager
 	 * @param IDBConnection $connection
 	 * @param DefaultShareProvider $defaultShareProvider
-	 * @param \OCP\Activity\IManager $activityManager
+	 * @param ActivityIManager $activityManager
 	 */
 	public function __construct(
 		IManager $shareManager,
 		IDBConnection $connection,
 		DefaultShareProvider $defaultShareProvider,
-		\OCP\Activity\IManager $activityManager
+		ActivityIManager $activityManager
 	) {
 		// Run once a day
-		$this->setInterval(24 * 60 * 60);
+		$this->setInterval(5);
 		$this->shareManager = $shareManager;
 		$this->connection = $connection;
 		$this->defaultShareProvider = $defaultShareProvider;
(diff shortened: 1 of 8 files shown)


AlexAndBear (approved):

jvillafanez:
I'm not sure if this is right. Technically, the agent is like a virtual user that performs the action on behalf of another user.

Having a "share_expire_author" doesn't fit here. What if there are more actions triggered during the share expiration? They could have the "share_expire_author" for a delete file action, which seems weird. Also note that whoever is listening for the delete file event won't likely take into account a "share_expire_author" agent, or won't receive any special treatment.

I'd prefer to reuse the "automation_author" in this case and try to figure out if the share has expired in a different way. A "cli_author" seems to generic for this case, but we could use a different author.
The question is who performs the share expiration **and any other related action** that could happen during that time.

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -48,23 +49,31 @@ class ExpireSharesJob extends TimedJob {
...
 	 * @param IDBConnection $connection
 	 * @param DefaultShareProvider $defaultShareProvider
+	 * @param \OCP\Activity\IManager $activityManager
 	 */
 	public function __construct(
 		IManager $shareManager,
 		IDBConnection $connection,
-		DefaultShareProvider $defaultShareProvider
+		DefaultShareProvider $defaultShareProvider,
+		\OCP\Activity\IManager $activityManager
```
jvillafanez, line 69: I'm not sure if there is any advantage in using this instead of `use OCP\Activity\IManager as ActivityManager` along with the rest of the "use" statements. I'd prefer the later since all the "use" statements are close to each other.

lib/private/Share20/Manager.php
```diff
@@ -123,6 +127,7 @@ public function __construct(
 			EventDispatcher $eventDispatcher,
 			View $view,
 			IDBConnection $connection,
+			\OCP\Activity\IManager $activityManager,
```
jvillafanez, line 130: same here

JammingBen:
@jvillafanez 

> I'd prefer to reuse the "automation_author" in this case and try to figure out if the share has expired in a different way.

What do you think of expanding the event parameters here by `shareExpired`: https://github.com/owncloud/core/blob/4d11dd4524ca25cf7f92a5002fb9be758248238f/lib/private/Share20/Manager.php#L1098. This way we could use this information in the activity app and add it as param. The author could be `automation_author` then.

jvillafanez:
That's fine

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -60,16 +61,16 @@ class ExpireSharesJob extends TimedJob {
...
 	public function __construct(
 		IManager $shareManager,
 		IDBConnection $connection,
 		DefaultShareProvider $defaultShareProvider,
-		\OCP\Activity\IManager $activityManager
+		ActivityIManager $activityManager
 	) {
 		// Run once a day
-		$this->setInterval(24 * 60 * 60);
+		$this->setInterval(5);
```
JanAckermann, line 73: WHOOOOPS

AlexAndBear (changes requested):

apps/files_sharing/lib/ExpireSharesJob.php
JammingBen: Dammit!

changelog/unreleased/38631
```diff
@@ -0,0 +1,7 @@
+Enhancement: Expired shares in activity stream
+
+Add a proper translation in the activity stream for shares which expired automatically.
```
JanAckermann, line 3:
```suggestion
Add a proper message in the activity stream for shares which expired automatically.
```

lib/private/Share20/Manager.php
```diff
@@ -175,7 +182,7 @@ private function splitFullId($id) {
 	 * @param \OCP\Share\IShare $share
 	 * @return bool
 	 */
-	private function shareHasExpired($share) {
+	private static function shareHasExpired($share) {
```
JanAckermann, line 185: Why this needs to be a static ?

AlexAndBear (changes requested):

lib/private/Share20/Manager.php
JammingBen: Because we need to call it in `static function formatUnshareHookParams()`. So either make that one non-static, or make `shareHasExpired()` static. I went for the latter because it has no connection to the class instance however.

JammingBen:
@jvillafanez I implemented it the way you suggested. Some changes are required in the activity app for this, see https://github.com/owncloud/activity/pull/944. We basically now use the automation author and the `shareExpired` param to indicate if a share expired.

AlexAndBear (approved):

jvillafanez (approved):

PR #38636, opened 2021-04-13
[tests-only] Use tiny skeleton in acceptance tests

### LINKED ISSUE #38607: Acceptance tests with no skeleton have welcome.txt
See PR #38606 for an idea of what was unexpected.

In core drone CI test scenarios pass that create a user "without skeleton files" and then make use of "welcome.txt".

Actually there should be no skeleton files at all. When drone CI runs for an oC10 app, such test scenarios fail. There is no "welcome.txt".

Investigate what is happening, why "without skeleton files" is not completely effective in some cases. And fix the code. (We want to have "without skeleton files" behave as expected in core, so that we immediately find any accidental uses of files like `welcome.txt`)

phil-davis:
## Description
PR https://github.com/owncloud/testing/pull/171 added `tinySkeleton` to the skeletons available in the testing app. That has just a `.gitignore` file.

Use `tinySkeleton` for users that are created "without skeleton files".

Note: "without skeleton files" is not really a thing in oC10. You always have some folder that is the skeleton directory. So this change makes "without skeleton files" have a tiny known skeleton, rather than getting whatever might happen in the system-under-test. See the discussion in the issue for more details.

Note 2: I discovered an existing latent problem in the case where a user is created but not initialized, and they require a skeleton that is not the existing one that is in effect. In that case, for example, the skeleton is switched from `tiny` to `small`, then the user is created (without initializing), then the skeleton is switched back to tiny, then the scenario proceeds and the user does an action at some point. At that point the user gets initialized, but with the tiny skeleton. The 2nd commit corrects the problem - do not revert the skeleton setting if the user was not initialized. (We rely on AfterScenario and/or the code in `run.sh` to put the skeleton back to the original value when the test(s) are finished)

## Related Issue
- Fixes #38607 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -1104,7 +1104,13 @@ trait Provisioning {
 		try {
 			$this->createTheseUsers($setDefaultAttributes, $initialize, true, $table);
 		} finally {
-			$this->setSkeletonDir($originalSkeletonPath);
+			// The effective skeleton directory is the one when the user is initialized
+			// If we did not initialize the user on creation, then we need to leave
+			// the skeleton directory in effect so that it applies when some action
+			// happens later in the scenario that causes the user to be initialized.
+			if ($initialize) {
+				$this->setSkeletonDir($originalSkeletonPath);
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


individual-it (approved):

kiranparajuli589 (commented):
lgtm 👍🏼

PR #38637, opened 2021-04-13
[tests-only] Verify tiny skeleton

phil-davis:
Verify that tinySkeleton contains `.gitignore` and is in use.


phil-davis:
Looks good - `.gitignore` exists when the user has tiny skeleton.

PR #38643, opened 2021-04-16
Added excludeAction param

AlexAndBear:
## Description
Adding excludeActions parameter to fileActions registerDefaultActions

## Motivation and Context
Imagine you write your own file list for a specific app, you want you to use all the default actions, but not the 'Rename' action.
Therefore you needed to define all the actions again in the app. This can now be overcome by passing ['Rename'] as excludeActions parameter. 
 

## How Has This Been Tested?

Manual UI Testing

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/fileactions.js
+++ b/apps/files/js/fileactions.js
@@ -615,47 +615,85 @@
 
 		/**
 		 * Register the actions that are used by default for the files app.
+		 * @param {['Download', 'Rename', 'Delete']|null} [excludeActions] exclude actions
 		 */
-		registerDefaultActions: function() {
-			this.registerAction({
-				name: 'Download',
-				displayName: t('files', 'Download'),
-				order: -20,
-				mime: 'all',
-				permissions: OC.PERMISSION_READ,
-				iconClass: 'icon-download',
-				actionHandler: function (filename, context) {
-					var dir = context.dir || context.fileList.getCurrentDirectory();
-					var isDir = context.$file.attr('data-type') === 'dir';
-					var url = context.fileList.getDownloadUrl(filename, dir, isDir);
+		registerDefaultActions: function (excludeActions) {
+			excludeActions = excludeActions || [];
 
-					var downloadFileaction = $(context.$file).find('.fileactions .action-download');
+			var self = this;
+			var defaultActions = [
+				{
+					name: 'Download',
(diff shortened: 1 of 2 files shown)


JanAckermann:
@phil-davis any clue why the js tests are failing ?

apps/files/js/fileactions.js
```diff
@@ -615,47 +615,85 @@
 
 		/**
 		 * Register the actions that are used by default for the files app.
+		 * @param {['Download', 'Rename', 'Delete']|null} [excludeAction] exclude actions
```
JammingBen, line 618: `excludeActions` (plural) seems more suitable.

apps/files/js/fileactions.js
```diff
@@ -615,47 +615,85 @@
...
-				displayName: t('files', 'Rename'),
-				mime: 'all',
-				order: -30,
-				permissions: OC.PERMISSION_UPDATE,
-				iconClass: 'icon-rename',
-				actionHandler: function (filename, context) {
-					context.fileList.rename(filename);
+
+			defaultActions.forEach(function (action) {
+				if (!excludeAction.includes(action.name)) {
```
JammingBen, line 695: I believe we cannot use `includes()`, that's also probably why the tests fail. Better use `indexOf`, although it's ugly.

JammingBen (changes requested):

JammingBen:
> @phil-davis any clue why the js tests are failing ?

As mentioned in the review, I assume it's due to `includes()`. At least we can give it a try 🙂

JammingBen (commented):
Nice! Add a changelog, then it's good to go from my side 🚀

JammingBen (approved):

JanAckermann:
@jvillafanez cool with it?

jvillafanez:
My only "complain" is that the `defaultActions` variable seems too big and it makes the function difficult to read. It's a bit easy to lose focus on what the function is doing. Taking into account that we don't plan to add more default actions there, it could be fine since we probably don't want to keep a copy of that variable somewhere else.
Anyway, I'm approving since I don't have better options at the moment.

jvillafanez (approved):

PR #38649, opened 2021-04-20
Improve move performance by doing only one query if possible

jvillafanez:
## Description
Make only one query for the move operation if possible. Except for sqlite, which will keep the previous behaviour, the rest of the DBs will use 1-2 queries.

## Related Issue
No linked issue

## Motivation and Context
This will improve performance for the affected DBs

## How Has This Been Tested?
Tested manually with postgresql 9.2, oracle 11 XE, oracle 18 XE and MariaDB

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/38649
+++ b/changelog/unreleased/38649
@@ -1,0 +1,10 @@
+Change: Improve performance for the MOVE operation
+
+Previously, in order to move a folder, we needed to get all the content
+and replace the path of each one one by one.
+Now we replace the path of all the content in one DB query, which reduces
+the network traffic and allows better optimization in the DB.
+This change affects MySQL, MariaDB, Postgresql, and Oracle. Sqlite is excluded
+from this optimization and will use the previous behaviour.
+
+https://github.com/owncloud/core/pull/38649
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -596,16 +596,13 @@ class Cache implements ICache {
 		$versionString = $this->connection->getDatabaseVersionString();
 		$versionArray = \explode('.', $versionString);
 
-		$sourcePath .= '/';
-		$targetPath .= '/';
-
 		$sql = null;
 		$sqlParams = [
 			'sourceStorageId' => $sourceStorageId,
-			'sourcePath' => $sourcePath,
+			'sourcePath' => "$sourcePath/",
 			'targetStorageId' => $targetStorageId,
-			'targetPath' => $targetPath,
-			'sourcePathLike' => $this->connection->escapeLikeParameter($sourcePath) . '%'
+			'targetPath' => "$targetPath/",
+			'sourcePathLike' => $this->connection->escapeLikeParameter("$sourcePath/") . '%'
 		];
 		switch ($platformName) {
 			case 'oracle':
@@ -648,7 +645,6 @@ class Cache implements ICache {
 		$result = $this->connection->executeQuery($sql, [$sourceStorageId, $this->connection->escapeLikeParameter($sourcePath) . '/%']);
 		$childEntries = $result->fetchAll();
 		$sourceLength = \strlen($sourcePath);
-		$this->connection->beginTransaction();
 		$query = $this->connection->prepare('UPDATE `*PREFIX*filecache` SET `storage` = ?, `path` = ?, `path_hash` = ? WHERE `fileid` = ?');
 
 		foreach ($childEntries as $child) {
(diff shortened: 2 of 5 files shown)


lib/private/Files/Cache/Cache.php
```diff
@@ -600,6 +591,72 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
+		// MariaDB should be included as mysql
+		// if there is an (optimized) sql query with the parameters, run it
+		if (isset($sql, $sqlParams)) {
+			$this->connection->executeQuery($sql, $sqlParams);
+			return;
+		}
+
+		// for other DBs (sqlite), we keep the old behaviour -> get the list and update one by one
+		$sql = 'SELECT `path`, `fileid` FROM `*PREFIX*filecache` WHERE `storage` = ? AND `path` LIKE ?';
+		$result = $this->connection->executeQuery($sql, [$sourceStorageId, $this->connection->escapeLikeParameter($sourcePath) . '/%']);
```
jvillafanez, line 648: `$sourcePath` already contains a trailing '/' (not previously). sqlite tests could be failing due to this.

jvillafanez:
There is a connectivity issue between ownCloud and oracle 18c XE, but the query itself works.

mrow4a:
@jvillafanez should be for review or still draft?

jvillafanez:
The connectivity issue with oracle 18 XE can be fixed separately (if it isn't a problem on my side), so yes, this is ready to review.

changelog/unreleased/38649
```diff
@@ -0,0 +1,10 @@
+Change: Improve performance for the MOVE operation
+
+Previously, in order to move a folder, we needed to get all the content
+and replace the path of each one one by one.
+Now we replace the path of all the content in one DB query, which reduces
+the network traffic and allows better optimization in the DB.
+This change affects MySQL, MariaDB, Postgresql, and Oracle. Sqlite is excluded
+of this optimization is will use the previous behaviour.
```
phil-davis, line 8:
```suggestion
from this optimization and will use the previous behaviour.
```

lib/public/IDBConnection.php
```diff
@@ -244,6 +244,19 @@ public function quote($input, $type = IQueryBuilder::PARAM_STR);
 	 */
 	public function getDatabasePlatform();
 
+	/**
+	 * Get the database version as string
+	 *
+	 * This method depends on the database drive and might not return accurate information.
```
phil-davis, line 250:
```suggestion
	 * This method depends on the database driver and might not return accurate information.
```

lib/public/IDBConnection.php
```diff
@@ -244,6 +244,19 @@ public function quote($input, $type = IQueryBuilder::PARAM_STR);
...
+	 * An explicit server version could have been configured in the database driver, or the
+	 * driver might not be able to determine the version
+	 *
+	 * @return string|null the database version or null if the database driver can't
+	 * determine the version
+	 * @since 10.7.1
```
phil-davis, line 256: What do we do about guessing the version that this will be released in?
Probably there will not be any 10.7.1 - it will be 10.8.0.
Maybe this does not matter.

jvillafanez: 10.8.0 seems is a better choice. Need to change that.

lib/private/Files/Cache/Cache.php
```diff
@@ -600,6 +591,68 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
+
+		// MariaDB should be included as mysql
+		// if there is an (optimized) sql query with the parameters, run it
+		if (isset($sql, $sqlParams)) {
+			$this->connection->executeQuery($sql, $sqlParams);
+			return;
+		}
+
+		// for other DBs (sqlite), we keep the old behaviour -> get the list and update one by one
+		$sql = 'SELECT `path`, `fileid` FROM `*PREFIX*filecache` WHERE `storage` = ? AND `path` LIKE ?';
```
mrow4a, line 644: isnt it cleaner to put in switch case default ? and for other just have duplicate $this->connection->executeQuery ? It is pretty confusing to read.

jvillafanez: I can't say for sure. For sqlite, we have multiple queries to do, it isn't just one query. The implementation is quite different. I didn't want to touch the old implementation because it "works". Note that it isn't a "run queries A, B and C", but a "run query A, and then, based on the results, run query B multiple times"
In addition, one of the reasons to implement it this way is that we can easily comment out a possible faulty query (specially for oracle) so we can switch to the old behaviour.

Maybe splitting into 2 different method might help, but we lose the easy switching between old and new implementation.

AlexAndBear (approved):

jvillafanez:
Bugfix in https://github.com/owncloud/core/pull/38783

PR #38654, opened 2021-04-26
Change implementation to prevent triggering a file scan

jvillafanez:
The previous "$targetNode->getParent()" call was causing a big scan on
big installations, scanning more than 76000 files, when the mount point
was renamed

## Description
Change the implementation of the "isTargetAllowed" method to prevent a file scan

## Related Issue
https://github.com/owncloud/enterprise/issues/4266

## Motivation and Context
Renaming a shared mount won't trigger a file scan any longer

## How Has This Been Tested?
1. user1 shares a folder with user2
2. user2 renames the shared folder

A file scan isn't triggered.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38656, opened 2021-04-27
Allow easier rename of shared mount

jvillafanez:
## Description
Superseeds https://github.com/owncloud/core/pull/38654

If we're renaming the shared mount, allow it as long as the target doesn't exists. No need to check for shares in this case.
For moving the shared mount around, we still use the previous check (which could cause a file scan)

## Related Issue
https://github.com/owncloud/enterprise/issues/4266

## How Has This Been Tested?
Manually checked.
1. user1 shares folder with user2
2. user2 renames shared file

There is no scan triggered by the rename.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/SharedMount.php
+++ b/apps/files_sharing/lib/SharedMount.php
@@ -185,9 +185,18 @@ class SharedMount extends MountPoint implements MoveableMount {
 	 * @return bool true if allowed, false otherwise
 	 */
 	public function isTargetAllowed($target) {
+		$currentMount = $this->getMountPoint();
+		// note: $currentMount has a trailing slash. It doesn't matter for the check below
+		$isRename = \dirname($currentMount) === \dirname($target);
+
 		list($targetStorage, $targetInternalPath) = \OC\Files\Filesystem::resolvePath($target);
 		// note: cannot use the view because the target is already locked
 		$fileId = (int)$targetStorage->getCache()->getId($targetInternalPath);
+		if ($isRename) {
+			// if it's a rename, it's allowed if the target doesn't exists
+			return $fileId === -1;
+		}
+
 		if ($fileId === -1) {
 			// target might not exist, need to check parent instead
 			$fileId = (int)$targetStorage->getCache()->getId(\dirname($targetInternalPath));
--- a/changelog/unreleased/38656
+++ b/changelog/unreleased/38656
@@ -1,0 +1,9 @@
+Change: Optimize share rename
+
+Renaming a received share could cause a file scan to be triggered. This could
+potentially be a performance problem if the file scan took a while.
+
+Now, renaming a received share won't trigger that file scan, so the
+performance will be faster.
+
+https://github.com/owncloud/core/pull/38656


micbar (changes requested):
Needs to go into 10.8.0

Do not merge yet.

micbar:
@phil-davis Can we check if the scenario from the Support Case is covered in CI?

phil-davis:
> @phil-davis Can we check if the scenario from the Support Case is covered in CI?

@micbar 

https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementToRoot/moveReceivedShare.feature has scenarios where the share receiver renames a received shared file and/or folder.

https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementToShares/moveReceivedShare.feature has similar scenarios for the case when share folder is set to `Shares` so that the original share is received in the `Shares` folder, then the sharee moves/renames it elsewhere.

micbar:
@phil-davis thanks
The problem is, that the rename triggered a file scan for the whole user account.

Could we be aware of that in the CI?

`AND a file_scan for <User> has not been triggered`

phil-davis:
> Could we be aware of that in the CI?

What evidence would there be after a file scan has happened?

If there is some way to set-up a user's file system so that there is some other "externally visible inconsistency" that will be fixed by a file scan, then if that inconsistency "magically" fixes itself after the rename, we can deduce that a file scan happened in the background.

butonic (approved):
Yep, makes sense to return early when we are just renaming! Nice find @jvillafanez!

butonic:
@jvillafanez maybe https://github.com/owncloud/core/pull/38656/files#diff-d323977d6ebb4748a02fcd21bed187528232608fbe9e4c82faed2605f89c6079R212 can help with further corner cases.

AFAICT if you try to move a shared file into a subfolder ... we should still see the same lag ... while renaming should be fast with this PR.

IljaN (approved):

jvillafanez:
Having "/foo/folder1" and "/foo/folder2" as a received folders from a different user it might be possible to rename "/foo/folder1" to "/foo/folder2" causing an overlap. However, this scenario isn't easily reproducible because the UI doesn't allow it ("/foo/folder2" already exists), and from the dav interface the target folder ("/foo/folder2") is removed before the rename.
We might need to refine the code a bit.

Moving the mount might still trigger the scan, but since we need to know if we're moving to a share... https://github.com/owncloud/core/pull/38654 also avoids the file scan but we need to get all the shares, so we can hit a different performance problem.

jvillafanez:
Checked with custom script: renaming to another local share is also prevented. The fileid we get is the one matching the source.
1. user1 shares "/folder1/sh0" and "/folder1/sh1"
2. user2 renames "/sh0" to "/sh1" (both received shares from user1)

The fileid we get is the one from user1's "/folder1/sh1" folder, so the fileid exists and the move is prevented. No need to change the code above.

For the remote shares, it seems the code goes through a different path so we might need to investigate separately.

VicDeo (approved):
Looks good, makes sense

JammingBen (approved):
LGTM 👍

JanAckermann:
@micbar since we don't plan a patch version like 10.7.1 we could merge this ?

micbar (approved):

micbar:
yes, go on

Issue #38657, opened 2021-04-27
[QA] unexpected test failure 


SwikritiT:
api tests :

https://github.com/owncloud/core/blob/1347fbd46fcd8f1f8ff6d98f1607c7064dc2a0f6/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature#L201

https://github.com/owncloud/core/blob/1347fbd46fcd8f1f8ff6d98f1607c7064dc2a0f6/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature#L202

are failing in drone and needs further investigation 

https://drone.owncloud.com/owncloud/ocis/4720/36/7

phil-davis:
```
  Scenario Outline: Listing other user's trashbin is prohibited for newly recreated user with same name     # /srv/app/testrunner/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature:179
    Given using <dav-path> DAV path                                                                         # FeatureContext::usingOldOrNewDavPath()
    And user "testtrashbin102" has been created with default attributes and without skeleton files          # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "testtrashbin102" has uploaded file "filesForUpload/textfile.txt" to "/textfile0.txt"          # FeatureContext::userHasUploadedAFileTo()
    And user "testtrashbin102" has uploaded file "filesForUpload/textfile.txt" to "/textfile1.txt"          # FeatureContext::userHasUploadedAFileTo()
    And user "Brian" has been created with default attributes and without skeleton files                    # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "testtrashbin102" has deleted file "/textfile0.txt"                                            # FeatureContext::userHasDeletedFile()
    And user "testtrashbin102" has deleted file "/textfile2.txt"                                            # FeatureContext::userHasDeletedFile()
    And the administrator deletes user "testtrashbin102" using the provisioning API                         # FeatureContext::theAdminDeletesUserUsingTheProvisioningApi()
    And these users have been created with default attributes and small skeleton files but not initialized: # FeatureContext::theseUsersHaveBeenCreated()
      | username        |
      | testtrashbin102 |
    And user "testtrashbin102" has deleted file "/textfile3.txt"                                            # FeatureContext::userHasDeletedFile()
[...]

PR #38541, opened 2021-03-17
[Tests-Only] used skeleton files more wisely on api tags, translation and trashbin test features

kiranparajuli589:
## Description
With this PR:
- skeleton file/folders are used more wisely on the following features suite
  - apiTags
  - apiTranslation
  - apiTrashbin

## Related Issue

- Part of https://github.com/owncloud/QA/issues/658

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38658, opened 2021-04-27
[tests-only] Adjust trashbinFilesFolders test setups

phil-davis:
## Description
PR #38541 adjusted the use of skeleton files. The files created in the Given steps need to be adjusted in some scenarios that use `textfile2`

## Related Issue
- related to #38657 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature
+++ b/tests/acceptance/features/apiTrashbin/trashbinFilesFolders.feature
@@ -160,7 +160,7 @@ Feature: files and folders exist in the trashbin after being deleted
     Given using <dav-path> DAV path
     And user "testtrashbin101" has been created with default attributes and without skeleton files
     And user "testtrashbin101" has uploaded file "filesForUpload/textfile.txt" to "/textfile0.txt"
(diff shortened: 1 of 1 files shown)


individual-it (changes requested):
shouldn't `    And user "testtrashbin101" has deleted file "/textfile2.txt"` have failed before this change?

phil-davis:
> shouldn't ` And user "testtrashbin101" has deleted file "/textfile2.txt"` have failed before this change?

The file did not exist, so maybe the given step passes because actually the file does not exist at the end of the step. The code in WebDav.php `userHasDeletedFile` does:
```
		$this->theHTTPStatusCodeShouldBe(
			["204", "404"],
			"HTTP status code was not 204 or 404 while trying to $deleteText $fileOrFolder '$entry' for user '$user'"
		);
```

So it passes because the file does not exist (and never existed).

Maybe we should change that in a separate PR, so that we know when a `Given` step tries to delete a file that is already not existing?

individual-it (approved):

PR #38660, opened 2021-04-28
[tests-only] Be more strict when checking Given user has deleted file

phil-davis:
## Description
If we use the `Given user "xyz" has deleted file "abc"` step then we really expect that there is an existing file to delete. This PR adjusts the code for the step so that only success (204) is allowed. A 404 "not found" will now cause the scenario to fail, so we will know if someone writes a step like this that has the wrong file name or is unnecessary.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/cliBackground/backgroundQueue.feature
+++ b/tests/acceptance/features/cliBackground/backgroundQueue.feature
@@ -25,6 +25,6 @@ Feature: get status, delete and execute jobs in background queue
   Scenario: delete one of the job in background queue
     Given user "Alice" has been created with default attributes and small skeleton files
     And user "Alice" has deleted file "/textfile0.txt"
-    When the administrator deletes last background job "OC\Command\CommandJob" using the occ command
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/cliBackground/backgroundQueue.feature
```diff
@@ -25,6 +25,6 @@ Feature: get status, delete and execute jobs in background queue
   Scenario: delete one of the job in background queue
     Given user "Alice" has been created with default attributes and small skeleton files
     And user "Alice" has deleted file "/textfile0.txt"
-    When the administrator deletes last background job "OC\Command\CommandJob" using the occ command
+    When the administrator deletes the last background job "OC\Command\CommandJob" using the occ command
```
phil-davis, line 28: I noticed this step while looking at the scenario. So I adjusted the English grammar.

individual-it (approved):

PR #38683, opened 2021-05-03
[tests-only] Trigger CI

phil-davis:
The `owncloudci/php` images were just updated to the latest PHP patch releases.
Confirm that CI passes.


phil-davis:
working - good

PR #38682, opened 2021-05-03
objectids for tags and comment are strings

jvillafanez:
## Description
The object ids changed in the PR are stored as strings in the DB, but ownCloud tries to store them as int.

## Related Issue
No issue opened
https://central.owncloud.org/t/php-7-4-18-breaking-owncloud-10-6-wih-postgresql-9-6-on-centos-7/31933

## Motivation and Context
Fixed bug in php caused a problem on our side.

## How Has This Been Tested?
Manually checked with postgresql 9.2 and php 7.4.18
**NOT** tested with other DBs yet.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38685, opened 2021-05-03
[tests-only] Test with postgres:9.4

phil-davis:
to see if the problem fixed in #38682 causes any tests to fail in CI

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1099,7 +1099,7 @@ def phpTests(ctx, testType):
 			'mysql:5.5',
 			'mysql:5.7',
 			'mysql:8.0',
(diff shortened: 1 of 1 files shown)


phil-davis:
Everything passes with pgSQL 9.4

phil-davis:
With postgres:9.6 
https://drone.owncloud.com/owncloud/core/29759/151/15
`webUIWebdavLocks/unlock.feature:205` fails. But that is already tagged `@skipOnDbPgsql` - we do not expect that to pass anyway, for different reasons:
"Pgsql and Oracle database can have different sorting, so allow this scenario to be skipped on those"

PR #38689, opened 2021-05-04
Throw generic exception to overcome, senstitive exception data exposure

AlexAndBear:
## Description
Sensible information could be exposed when downloading a share via public link. We now throw a generic exception that overwrites the original exception message. Also fixed an error which caused such behavior when appending a null byte to the download URL.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4536

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controllers/ShareController.php
+++ b/apps/files_sharing/lib/Controllers/ShareController.php
@@ -400,6 +400,7 @@ class ShareController extends Controller {
 	 * @param string $path
 	 * @param string $downloadStartSecret
 	 * @return NotFoundResponse|RedirectResponse|void
+	 * @throws \Exception
 	 */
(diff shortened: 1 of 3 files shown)


C0rby (approved):
This seems to do the trick.
Looks good to me. :+1:

JanAckermann:
@C0rby wouldn't it be better to wrap the whole controller function in a try-catch block and dedicate the NotFoundException (will throw NotFoundExcpetion) and every other Exception will throw the generic \Exception ?

C0rby:
I have no hard opinion on that. It seems that the generic Exception ends up showing a "Not Found" error in the frontend.

JanAckermann:
Then we'll leave it as it is, as it has minium impact.

PR #38693, opened 2021-05-05
Regenerate session for anonymous user after succesfully auth

AlexAndBear:
## Description
Regenerate session after entering the password of a password protected public share for anonymous users 

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4535

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Controllers/ShareController.php
+++ b/apps/files_sharing/lib/Controllers/ShareController.php
@@ -50,6 +50,7 @@ use OCP\IRequest;
 use OCP\ISession;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
+use OCP\IUserSession;
 use OCP\Share;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Template;
@@ -81,6 +82,8 @@ class ShareController extends Controller {
 	protected $previewManager;
 	/** @var IRootFolder */
 	protected $rootFolder;
+	/** @var IUserSession */
+	protected $userSession;
 	/** @var EventDispatcher */
 	protected $eventDispatcher;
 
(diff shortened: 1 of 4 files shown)


apps/files_sharing/lib/Controllers/ShareController.php
```diff
@@ -164,6 +170,9 @@ public function authenticate($token, $password = '') {
 		$authenticate = $this->linkShareAuth($share, $password);
 
 		if ($authenticate === true) {
+			if (!$this->userSession->isLoggedIn()) {
```
mrow4a, line 173: this is to detect public link?

mrow4a (commented):
what if I was in normal folder, and then accessed public link? would I need to authenticate again?

apps/files_sharing/lib/Controllers/ShareController.php
JanAckermann: We Check If you was logged in before, so I guess nope

JanAckermann: No, we don't create a new session for a already logged in user

JanAckermann:
> what if I was in normal folder, and then accessed public link? would I need to authenticate again?

We Check If you was logged in before, so I guess nope

JanAckermann:
> what if I was in normal folder, and then accessed public link? would I need to authenticate again?

No, we only reset the cookie/session for a non-authenticated user, if the cookie was injected before to the authenticated user via XSS, the potential attacker would already have access.

apps/files_sharing/lib/Controllers/ShareController.php
JanAckermann: No, we don't create a new session for a already logged in user, see comment below

C0rby (approved):
Looks good to me! :+1:

PR #38699, opened 2021-05-05
Fix potential problems with the cache due to removing wrong data

jvillafanez:
The ICache interface (implemented by the CappedMemoryCache) states that
the clear method must remove the items starting with the prefix passed
as parameter. However, the implementation removes the whole cache.
The current usage seems to work because of this flaw. However, assuming
that "bug" is fixed, the implementation will likely have problems due to
clearing (or not) wrong data.
This commit aims to prevent problems when the CappedMemoryCache adjusts
its code according to the interface.

## Description
As said in the commit, the `CappedMemoryCache` removes all the items in the cache instead of cleaning it by prefix. The SMB storage might have been relying on this flaw indirectly and trying to "clear" wrong folders from the cache. Since the implementation removes all the cache, it didn't matter what folder the SMB wanted to remove from its cache. However, this will become a problem if the `CappedMemoryCache` is fixed.

In addition, some variable have been renamed to be more clear about what are the cache keys we're using and to try to prevent mistakes using other similar variables.

## Related Issue
No opened issue

## Motivation and Context
Spotted while working on other issue. It's better not to rely on a wrong implementation.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -222,8 +222,9 @@ class SMB extends StorageAdapter {
 			$path = $this->buildPath($path);
 			$result = [];
 			$children = $this->share->dir($path);
+			$trimmedPath = \rtrim($path, '/');
 			foreach ($children as $fileInfo) {
-				$fullPath = "{$path}/{$fileInfo->getName()}";
+				$fullPath = "{$trimmedPath}/{$fileInfo->getName()}";
 				if (isset($this->statCache[$fullPath])) {
 					// reference in the cache might have its fileinfo's mode
 					// already resolved, so use that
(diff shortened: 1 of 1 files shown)


apps/files_external/lib/Lib/Storage/SMB.php
```diff
@@ -332,10 +335,10 @@ public function rename($source, $target) {
 	}
 
 	private function removeFromCache($path) {
-		$path = \trim($path, '/');
 		// TODO The CappedCache does not really clear by prefix. It just clears all.
 		'@phan-var \OC\Cache\CappedMemoryCache $this->statCache';
-		$this->statCache->clear($path);
+		$this->statCache->clear("$path/");
```
VicDeo, line 340: @jvillafanez is this method supposed to work for directories only? 
I see from code path that it is used on rename. What happens when the file is renamed?

jvillafanez: If $path is a file, we won't have the same name as a directory.
We can assume that there won't be "path/to/file.txt/and/more" if "path/to/file.txt" is a file, so it won't do anything. Basically, the `clear` will take care of the contents of the directory (if it's a file, nothing happens), and then the `unset` call below will delete the actual file or directory.

Side note: currently the `CappedMemoryCache` wipes out all the cache with the `clear` method, although it should remove the data by prefix. These changes work with the assumption that the cache will be deleted by prefix (which is how it should be implemented.)

VicDeo (approved):

PR #38701, opened 2021-05-06
[tests-only] Add ability to skip pipelines in CI

saw-jan:
## Description
This PR adds the ability to skip certain CI pipelines (`PhpUnitTests`, `JavascriptTest` and `acceptanceTest`)

## Related Issue
- part of  https://github.com/owncloud/QA/issues/670

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38702, opened 2021-05-06
[tests-only] skip some pipelines

phil-davis:
To check that the 'skip' setting added in PR #38701 is working.


phil-davis:
The expected reduced set of unit test and acceptance test pipelines have started - good.

PR #38704, opened 2021-05-06
Trashbin skip list 

AlexAndBear:
## Description
With this change new config parameters has been introduced.
Admins can now decide, based on file extensions, directory names and size,
if a resource should not be observed by the trashbin and deleted immediately.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_trashbin/lib/Storage.php
+++ b/apps/files_trashbin/lib/Storage.php
@@ -46,9 +46,13 @@ class Storage extends Wrapper {
 	/** @var  IUserManager */
 	private $userManager;
 
-	public function __construct($parameters, IUserManager $userManager = null) {
+	/** @var  TrashbinSkipChecker */
+	private $trashbinSkipChecker;
+
+	public function __construct($parameters, IUserManager $userManager = null, TrashbinSkipChecker $trashbinSkipChecker = null) {
 		$this->mountPoint = $parameters['mountPoint'];
 		$this->userManager = $userManager;
+		$this->trashbinSkipChecker = $trashbinSkipChecker;
 		parent::__construct($parameters);
 	}
 
@@ -178,6 +182,16 @@ class Storage extends Wrapper {
 		$normalized = Filesystem::normalizePath($this->mountPoint . '/' . $path, true, false, true);
 		$result = true;
 		$view = Filesystem::getView();
+
+		if ($view instanceof View) {
+			$relativePath = $view->getRelativePath($normalized);
+
+			// Skip trashbin based on config
+			if ($relativePath && $this->trashbinSkipChecker->shouldSkipPath($view, $relativePath) === true) {
+				return \call_user_func_array([$this->storage, $method], [$path]);
+			}
+		}
+
 		if (!isset($this->deletedFiles[$normalized]) && $view instanceof View) {
 			$this->deletedFiles[$normalized] = $normalized;
 			if ($filesPath = $view->getRelativePath($normalized)) {
@@ -228,7 +242,11 @@ class Storage extends Wrapper {
 		\OC\Files\Filesystem::addStorageWrapper('oc_trashbin', function ($mountPoint, $storage) {
 			return new \OCA\Files_Trashbin\Storage(
 				['storage' => $storage, 'mountPoint' => $mountPoint],
-				\OC::$server->getUserManager()
+				\OC::$server->getUserManager(),
+				new TrashbinSkipChecker(
+					\OC::$server->getLogger(),
+					\OC::$server->getConfig()
+				)
 			);
 		}, 1);
 	}
--- a/apps/files_trashbin/lib/TrashbinSkipChecker.php
+++ b/apps/files_trashbin/lib/TrashbinSkipChecker.php
@@ -1,0 +1,147 @@
+<?php
+/**
+ * @author Jan Ackermann <jackermann@owncloud.com>
+ * @author Jannik Stehle <jstehle@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Trashbin;
+
+use OC\Files\View;
+use OCP\IConfig;
+use OCP\ILogger;
+use OCP\Util;
+
+/**
+ * Class TrashbinSkipChecker
+ *
+ * TrashbinSkipChecker won't be initialized trough dependency injection
+ *
+ * @package OCA\Files_Trashbin
+ */
+class TrashbinSkipChecker {
+
+	/**
+	 * @var ILogger
+	 */
+	private $logger;
+
+	/**
+	 * @var  IConfig
+	 */
+	private $config;
+
+	/**
+	 * @param ILogger $logger
+	 * @param IConfig $config
+	 */
+	public function __construct(
+		ILogger $logger,
+		IConfig $config
+	) {
+		$this->logger = $logger;
+		$this->config = $config;
+	}
+
+	/**
+	 * Check if a given path needs to be skipped by the trashbin
+	 *
+	 * @param View $view resource view
+	 * @param string $relativePath relative path to the file or folder
+	 * @return bool
+	 */
+	public function shouldSkipPath(View $view, string $relativePath): bool {
+		$isFolder = $view->getMimeType($relativePath) === 'httpd/unix-directory';
+		$relativePath = \ltrim($relativePath, '/');
+
+		if ($this->checkDirectoryName($relativePath, $isFolder) === true) {
+			return true;
+		}
+
+		if (!$isFolder && $this->checkFileExtension($relativePath) === true) {
+			return true;
+		}
+
+		if ($this->checkFileSize($view, $relativePath) === true) {
+			return true;
+		}
+
+		return false;
+	}
+
+	protected function checkDirectoryName(string $relativePath, bool $isFolder): bool {
+		$directoriesToSkip = $this->config->getSystemValue('trashbin_skip_directories', []);
+
+		if ($isFolder) {
+			// add trailing slash to ensure correct behavior with folders.
+			// otherwise a config like "tes" would apply to a folder "test".
+			$relativePath = \rtrim($relativePath, '/');
+			$relativePath .= '/';
+		}
+
+		foreach ($directoriesToSkip as $skipItem) {
+			$skipItem = \trim($skipItem);
+			$skipItem .= '/';
+
+			if (\substr($relativePath, 0, \strlen($skipItem)) === $skipItem) {
+				$this->logger->info("Path $relativePath skipped the trashbin and has been deleted immediately due to the trashbin_skip_directories config $skipItem");
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	protected function checkFileExtension(string $relativePath): bool {
+		$fileExtensionsToSkip = $this->config->getSystemValue('trashbin_skip_extensions', []);
+
+		foreach ($fileExtensionsToSkip as $skipItem) {
+			$skipItem = \trim($skipItem);
+			$skipItem = '.'.$skipItem;
+
+			if (\substr($relativePath, -\strlen($skipItem)) === $skipItem) {
+				$this->logger->info("Path $relativePath skipped the trashbin and has been deleted immediately due to the trashbin_skip_extensions config $skipItem");
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	protected function checkFileSize(View $view, string $relativePath): bool {
+		$skipSizeThreshold = $this->config->getSystemValue('trashbin_skip_size_threshold', false);
+
+		if ($skipSizeThreshold === false) {
+			return false;
+		}
+
+		$computerSkipSizeThreshold = Util::computerFileSize($skipSizeThreshold);
+
+		if ($computerSkipSizeThreshold === false || $computerSkipSizeThreshold < 0) {
+			$this->logger->warning("Config trashbin_skip_size_threshold value $skipSizeThreshold is invalid");
+			return false;
+		}
+
+		if ($view->filesize($relativePath) >= $computerSkipSizeThreshold) {
+			$this->logger->info("Path $relativePath skipped the trashbin and has been deleted immediately due to the trashbin_skip_size_threshold config $skipSizeThreshold");
+			return true;
+		}
+
+		return false;
+	}
+}
--- a/apps/files_trashbin/tests/TrashbinSkipCheckerTest.php
+++ b/apps/files_trashbin/tests/TrashbinSkipCheckerTest.php
@@ -1,0 +1,94 @@
+<?php
+/**
+ * @author Jan Ackermann <jackermann@owncloud.com>
+ * @author Jannik Stehle <jstehle@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program. If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_Trashbin\Tests;
+
+use OC\Files\View;
+use OCA\Files_Trashbin\TrashbinSkipChecker;
+use OCP\IConfig;
+use OCP\ILogger;
+
+/**
+ * Class SkipTrashbinManagerTest
+ *
+ * @group DB
+ */
+class TrashbinSkipCheckerTest extends TestCase {
+
+	/** @var  ILogger */
+	protected $logger;
+	/** @var  IConfig */
+	protected $config;
+	/** @var  TrashbinSkipChecker */
+	protected $trashbinSkipChecker;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->logger = $this->getMockBuilder(ILogger::class)
+			->disableOriginalConstructor()->getMock();
+		$this->config = $this->getMockBuilder(IConfig::class)
+			->disableOriginalConstructor()->getMock();
+
+		$this->trashbinSkipChecker = new TrashbinSkipChecker($this->logger, $this->config);
+	}
+
+	public function testSkipDueToDirectoryName() {
+		$this->config->expects($this->any())
+			->method('getSystemValue')
+			->willReturn(['somefolder'], [], null);
+		$view = $this->createMock(View::class);
+
+		$view->expects($this->once())
+			->method('getMimeType')->willReturn('application/pdf');
+
+		$this->assertTrue($this->trashbinSkipChecker->shouldSkipPath($view, 'somefolder/somefile.pdf'));
+	}
+
+	public function testSkipDueToFileExtension() {
+		$this->config->expects($this->any())
+			->method('getSystemValue')
+			->willReturn([], ['pdf'], null);
+		$view = $this->createMock(View::class);
+
+		$view->expects($this->once())
+			->method('getMimeType')->willReturn('application/pdf');
+
+		$this->assertTrue($this->trashbinSkipChecker->shouldSkipPath($view, 'somefolder/somefile.pdf'));
+	}
+
+	public function testSkipDueToFileSize() {
+		$view = $this->createMock(View::class);
+
+		$this->config->expects($this->any())
+			->method('getSystemValue')
+			->willReturn([], [], '5B');
+
+		$view->expects($this->once())
+			->method('getMimeType')->willReturn('application/pdf');
+
+		$view->expects($this->once())
+			->method('filesize')->willReturn(6);
+
+		$this->assertTrue($this->trashbinSkipChecker->shouldSkipPath($view, 'somefolder/somefile.pdf'));
+	}
+}
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -579,6 +579,44 @@ $CONFIG = [
 'trashbin_purge_limit' => 50,
 
 /**
+ * Define trashbin directory skip list
+ * Define a list of directories that will skip the trashbin and therefore be deleted immediately.
+ * Only defined directories and only in the root of a mount will skip the trashbin.
+ * Consider not to use reserved directory names when using snapshot capable storage systems.
+ * The setting expects folder names with or without trailing slash.
+ * All the content of such directories including their subdirectories will also skip the trashbin.
+ *
+*/
+'trashbin_skip_directories' => [
+	'temp',
+],
+
+/**
+ * Define trashbin file extension skip list
+ * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately.
+ * Extension names are valid for all mount points, take care when selecting the names.
+ *
+ * Values must not have a leading ".", otherwise corresponding files wont be detected.
+ *
+ */
+'trashbin_skip_extensions' => [
+	'iso',
+	'mkv',
+],
+
+/**
+ * Define trashbin threshold size
+ * Define a threshold for files to skip the trashbin and delete immediately
+ * Once the size of a resource is greater than or equal the given value, the trashbin will be skipped.
+ * File sizes are valid for all mount points, take care when defining the threshold.
+ *
+ * All positive numbers and zero is allowed. Append one of the following options directly and without space:
+ * B, K, KB, MB, M, GB, G, TB, T, PB, P
+ *
+ */
+'trashbin_skip_size_threshold' => "1GB",
+
+/**
  * File versions
  *
  * These parameters control the Versions app.
--- a/changelog/unreleased/38704
+++ b/changelog/unreleased/38704
@@ -1,0 +1,7 @@
+Enhancement: Add trashbin skip list to config
+
+With this change new config parameters has been introduced.
+Admins can now decide, based on file extensions, directory names and size,
+if a resource should not be observed by the trashbin and deleted immediately.
+
+https://github.com/owncloud/core/pull/38704
--- a/lib/private/legacy/helper.php
+++ b/lib/private/legacy/helper.php
@@ -132,7 +132,7 @@ class OC_Helper {
 	/**
 	 * Make a computer file size
 	 * @param string $str file size in human readable format
-	 * @return float a file size in bytes
+	 * @return float|false a file size in bytes
 	 *
 	 * Makes 2kB to 2048.
 	 *
--- a/lib/public/Util.php
+++ b/lib/public/Util.php
@@ -453,7 +453,7 @@ class Util {
 	/**
 	 * Make a computer file size (2 kB to 2048)
 	 * @param string $str file size in a fancy format
-	 * @return int a file size in bytes
+	 * @return float|false a file size in bytes
 	 *
 	 * Inspired by: http://www.php.net/manual/en/function.filesize.php#92418
 	 * @since 4.0.0


JanAckermann:
@jvillafanez even if this is files_trashbin I would love to save the settings in the config.php because it's hard to set arrays via occ config:app:set agree?

jvillafanez:
Just a couple of things so far:
* I'd move the checks to its own class. This way it's easier to add unit tests for this and it also keeps the code isolated.
* I'd try to avoid using regular expressions unless needed.
  * So far, the examples provided in the config.php are for extensions and files inside folders (basically, paths ending or starting with some chars). This can be done faster without regular expressions.
  * Regular expressions are error-prone. My concern here is that the admin could enter a "wrong" regular expression, such as `.*.png` which could affect to files the admin doesn't want to, such as `a_file.apng`. This simply cannot be detected.
  * Case insensitiveness seems a bad idea if it wasn't asked for explicitly. The config.php file doesn't say anything about this.
* Taking into account that we don't expect too many skips, we probably should include some logs at info level (warning might be too much) for tracing purposes.

jvillafanez:
I still think it's better to move the checks to a dedicated class. We'll need to add unit tests, and it should be easy to add them in a dedicated class. It should also be easier to include an initialization phase to clean the config input and make the actual action easier to read.
In addition, it might be better to split into multiple smaller functions in order to give meaningful names to the functions to make the code easier to follow.

Also note that there could be filenames with "*" in the path (I don't think we're doing anything special), so there could be a "*abc" folder in the root.

jvillafanez:
Note that we should include an option to skip the trashbin by filesize, which will probably be the main use case for this feature. Taking this into account, maybe having "trashbin.skip.extensions", "trashbin.skip.directories" and "trashbin.skip.sizes" as config options might be a better choice

JanAckermann:
@jvillafanez without the cool new wildcard logic ?

jvillafanez:
I think so. The code seems complex at first sight, probably because the PHP string functions aren't easy to handle, and it also has some edge cases we need to take care. I don't mind having the wildcard logic if it easy to follow though.
My concern is that a 3rd person might take 1-2 days trying to put a fix there successfully because every time he touches something he breaks another thing. If you're confident your code is fool-proof, go ahead.

JanAckermann:
@jvillafanez can you have a second look, this is not quite final, tests need to be added.

.htaccess
```diff
@@ -94,3 +94,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
+
+ErrorDocument 403 //core/templates/403.php
+ErrorDocument 404 //core/templates/404.php
```
phil-davis: This is an accident - need to remove

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+ */
+
+namespace OCA\Files_Trashbin;
+
+use OC\Files\Filesystem;
+use OCP\IConfig;
+use OCP\ILogger;
+use OCP\Util;
+
+class SkipTrashbinManager {
```
jvillafanez, line 30: I don't think "manager" is the right term here. Sounds like it should have CRUD operations somehow, which isn't the case. Maybe "TrahbinSkipper" or "TrashbinSkipChecker" is more meaningful as classname

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Check if path needs to be skipped
+	 *
+	 * @param string $path path
+	 * @param string $mountPoint mount point
+	 * @return bool
+	 */
+	public function shouldSkipPath(string $path, string $mountPoint): bool {
```
jvillafanez, line 57: It seems it should be asking for the full path. I don't think we need to split mount point + path.
Also "path" is quite generic: full path including uid, such as "/user1/files/folder/to/file.txt"? or FS path such as "/folder/to/file.txt"?  with or without leading "/"? You can describe the expectations in the PHPDoc so it's easier to know how we can use the function properly.

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+
+		if ($this->checkSkipBySize($path, $mountPoint) === true) {
+			return true;
+		}
+
+		return false;
+	}
+
+	protected function checkSkipByDirectoryName($path, $mountPoint) {
+		$skipByDirectoriesConfigList = $this->config->getSystemValue('trashbin_skip_by_directories', []);
```
jvillafanez, line 74: Should we use "trashbin.skip.directories" for the config key? Although there is no special meaning, the namespacing gives the similar properties a sense of togetherness.

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+
+		$view = Filesystem::getView();
+		$normalizedPath = Filesystem::normalizePath($mountPoint . '/' . $path, true, false, true);
+		$path = \ltrim($view->getRelativePath($normalizedPath), '/');
+
+		foreach ($skipByDirectoriesConfigList as $skipItem) {
+			$skipItem = \trim($skipItem);
+
+			if (\substr($path, 0, \strlen($skipItem)) === $skipItem) {
+				$this->logger->info("Path $path record exists in config value trashbin_skip_by_directories and will be deleted immediately");
```
jvillafanez, line 84: Should we include in the log the rule that made the skip happen? It might give interesting information.

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+
+		return false;
+	}
+
+	protected function checkSkipByFileExtension($path) {
+		$skipByDirectoriesExtensionsConfigList = $this->config->getSystemValue('trashbin_skip_by_extensions', []);
+
+		$skipByDirectoriesListIncludesExtension = \in_array(\pathinfo($path, PATHINFO_EXTENSION), $skipByDirectoriesExtensionsConfigList);
```
jvillafanez, line 95: I prefer to use strict checking in the `in_array` function (adding `true` as 3rd param).

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+		if ($skipByDirectoriesListIncludesExtension) {
+			$this->logger->info("Path $path record exists in config value trashbin_skip_by_extensions and will be deleted immediately");
+			return true;
+		}
+
+		return false;
+	}
+
+	protected function checkSkipBySize($path, $mountPoint) {
+		$skipByDirectoriesExtensionsConfigList = $this->config->getSystemValue('trashbin_skip_by_size', false);
```
jvillafanez, line 106: `skipByDirectoriesExtensionsConfigList` is a wrong variable name to be used here.

apps/files_trashbin/lib/SkipTrashbinManager.php
```diff
@@ -0,0 +1,128 @@
...
+		$view = Filesystem::getView();
+		$normalizedPath = Filesystem::normalizePath($mountPoint . '/' . $path, true, false, true);
+		$filesize = $view->filesize($view->getRelativePath($normalizedPath));
+		$configComputerFilesize = Util::computerFileSize($skipByDirectoriesExtensionsConfigList);
+
+		if ($configComputerFilesize === false) {
+			return false;
+		}
+
+		if ($filesize > $configComputerFilesize) {
```
jvillafanez, line 121: I guess we don't want to consider skipping files with less than a specific size. It might have some usage if we consider files less than 100 bytes as useless or dummies which don't need to go to the trashbin. It's a weird use case, so I'm not going to push in this direction, but you can check with PM if there is any interest.

Anyway, I think this check should be `>=`. An exact filesize match shouldn't go to the trashbin I think.

apps/files_trashbin/lib/Storage.php
```diff
@@ -226,9 +232,13 @@ public function retainKeys($filename, $owner, $ownerPath, $timestamp, $sourceSto
...
 	public static function setupStorage() {
 		\OC\Files\Filesystem::addStorageWrapper('oc_trashbin', function ($mountPoint, $storage) {
+			$app = new App('files_trashbin');
+			$container = $app->getContainer();
+
 			return new \OCA\Files_Trashbin\Storage(
 				['storage' => $storage, 'mountPoint' => $mountPoint],
-				\OC::$server->getUserManager()
+				\OC::$server->getUserManager(),
+				$container->query('OCA\Files_Trashbin\SkipTrashbinManager')
```
jvillafanez, line 241: `SkipTrashbinManager::class` is usually preferred

config/config.sample.php
```diff
@@ -578,6 +578,35 @@
...
+	'big_files',
+],
+
+/**
+ * Define the trashbin skip by size value
+ * This setting defines which files or folders should not be moved to the trashbin and deleted immediately based on greater than size.
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
+ */
+'trashbin_skip_by_size' => "1GB",
```
jvillafanez, line 608: It might be useful to know what units are available, and the specific format, if any.
Is it "1 GB" (with the space) allowed? What about "01GB"?

jvillafanez:
For the config.sample file, I prefer to use the default values if any. If it isn't possible (the defaults in this case will likely be empty lists, which doesn't add any value), some sane defaults is a better choice, so it's easy to copy and paste them. You might also want to clarify whether there is a default value in use or not (mainly to clarify that 1GB files will still reach the trashbin by default, for example). You can add additional examples in the comments to clarify things.
In this regard, you could include:
* for the extensions, the .exe files or .mkv might be more common choices.
* for the directories, I think our use case is the .snapshot folder. It's something we might need, so better to have it around
* for the filesize, 1GB seems reasonable. Depending on the usage, you might want to increase it to 5GB or 10GB

apps/files_trashbin/lib/SkipTrashbinManager.php
JammingBen: Some trashbin-config-settings already exist, they all start with `trashbin_`. So I'd go with the same naming convention here to keep things consistent.

I removed the `_by_` though as it was kind of misleading.

JammingBen:
@jvillafanez I updated the PR with some general improvements and your suggestions. Also added tests for the new class. I'm not a 100% sure about the ´config.sample´ yet. At least it should now be more clear how to use these config params.

Thanks for your help here btw 👍

apps/files_trashbin/lib/Storage.php
```diff
@@ -226,9 +242,13 @@ public function retainKeys($filename, $owner, $ownerPath, $timestamp, $sourceSto
 	 */
 	public static function setupStorage() {
 		\OC\Files\Filesystem::addStorageWrapper('oc_trashbin', function ($mountPoint, $storage) {
+			$app = new App('files_trashbin');
```
jvillafanez, line 245: I'm very wary about creating a new app on our own. The app could register itself in some place and could cause problems with duplicated calls among other things.
For this case, I think it's better to create a new instance of the `TrashbinSkipChecker` directly. The dependencies can be picked up from the `\OC::$server` instance (it's already there for the userManager anyway)

You might also want to include a note in the `TrashbinSkipChecker` to warn that there are instances created without DI (in case that we add a new dependency and the code here doesn't yell)

apps/files_trashbin/lib/TrashbinSkipChecker.php
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	protected function checkFileSize(string $path): bool {
+		$skipSizeThreshold = $this->config->getSystemValue('trashbin_skip_size_threshold', false);
+
+		if ($skipSizeThreshold === false) {
+			return false;
+		}
+
+		$filesize = Filesystem::getView()->filesize($path);
```
jvillafanez, line 121: Taking into account that the trashbin storage already uses a `View`, the `shouldSkipPath` could need the view as parameter. Something like `shouldSkipPath($view, $relativatePath)`. This has some advantages:
* No dependency with the `Filesystem`. Using a mock view in the tests is easier and reduce coupling.
* You don't need to ask whether the path is a folder or not. The view can take care of that. Method invocation is easier
* It's more natural to think that the relativePath required is based on the view

jvillafanez:
Warning messages seem harsh. I understand that we might want that message for the "skip directories" configuration as a deterrent, but the ones for the extension and file size... I expect the skip by file size to be widely used.

JanAckermann:
TBD: 
- [ ] Inform @mmattel to run config-to-docs-command for 10.8

mmattel:
@JanAckermann @jvillafanez 
Could you pls make in config.php each first line a brief description line as you can see in the other examples above or below.
The first line of a new config parameter is always taken as line for the table of contents. The lines below can be anything you like. See also the doc on the web how it renders.
I would like to avoid post merge corrections if possible 😃

config/config.sample.php
```diff
@@ -578,6 +578,39 @@
  */
 'trashbin_purge_limit' => 50,
 
+/**
+ * Define a list of folders that will skip the trashbin and therefore be deleted immediately
```
mmattel, line 582:
```suggestion
  * Define trashbin skip folder list
  * Define a list of folders that will skip the trashbin and therefore be deleted immediately
```

config/config.sample.php
```diff
@@ -578,6 +578,39 @@
...
+ * inside such folders will also skip the trashbin.
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
+ */
+'trashbin_skip_directories' => [
+	'.snapshots',
+],
+
+/**
+ * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately
```
mmattel, line 593:
```suggestion
  * Define trashbin file skip list
  * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately
```

config/config.sample.php
```diff
@@ -578,6 +578,39 @@
...
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
+ */
+'trashbin_skip_extensions' => [
+	'exe',
+	'mkv',
+],
+
+/**
+ * Define a threshold for files to skip the trashbin and delete immediately
```
mmattel, line 604:
```suggestion
  * Define trashbin threshold size
  * Define a threshold for files to skip the trashbin and delete immediately
```

config/config.sample.php
```diff
@@ -578,6 +578,42 @@
  */
 'trashbin_purge_limit' => 50,
 
+/**
+  * Define trashbin folder skip list
+  * Define a list of folders that will skip the trashbin and therefore be deleted immediately
```
mmattel, line 583:
```suggestion
 * Define trashbin folder skip list
 * Define a list of folders that will skip the trashbin and therefore be deleted immediately
```

config/config.sample.php
```diff
@@ -578,6 +578,42 @@
...
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
+ */
+'trashbin_skip_directories' => [
+	'.snapshots',
+],
+
+/**
+  * Define trashbin file extension skip list
+  * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately
```
mmattel, line 595:
```suggestion
 * Define trashbin file extension skip list
 * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately
```

config/config.sample.php
```diff
@@ -578,6 +578,42 @@
...
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
+ */
+'trashbin_skip_extensions' => [
+	'exe',
+	'mkv',
+],
+
+/**
+  * Define trashbin threshold size
+  * Define a threshold for files to skip the trashbin and delete immediately
```
mmattel, line 607:
```suggestion
 * Define trashbin threshold size
 * Define a threshold for files to skip the trashbin and delete immediately
```

mmattel:
@JanAckermann sorry, my fault, there was a leading blank too much on each changed line, one is correct, two not. Fixed

mmattel (approved):
OK from config.php POV
Code not reviewed

mmattel:
@phil-davis info: drone is inaccessible which results in an unsuccessful check

JanAckermann:
@mmattel Thank you, not yet ready some minor code improvements needs to be done

JanAckermann:
@jvillafanez do you mean the warning messages in the config sample ?
```
 * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
 ```

Mabye @mmattel can suggest what we should shout here, because data loss could be possible...

**Comment by JanAckermann** [comment]
@jvillafanez changed in favour of your suggestions

**Comment by JanAckermann** [comment]
- [ ] Improve config sample @mmattel

File: config/config.sample.php [hunk 19]
```diff
@@ -578,6 +578,42 @@
  */
 'trashbin_purge_limit' => 50,
 
+/**
+ * Define trashbin folder skip list
+ * Define a list of folders that will skip the trashbin and therefore be deleted immediately
+ * The setting expects folder names with or without trailing slash. All the content
+ * inside such folders will also skip the trashbin.
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
```
mmattel, line 587:
```suggestion
 * Define a list of folders that will skip the trashbin and therefore be deleted immediately.
 * Folder names are valid for all mount points, take care selecting the name.
 * Defined folder names can be in any location of the path and skip this folder from the trashbin.
 * All content of defined folders and their subfolders are affected.
 * Consider not using reserved folder names when using a storage system that provides snapshots.
 * It is not necessary to set a trailing slash, the folder name is enough.
 *
```

mmattel, line 587:
```suggestion
 * Define trashbin directory skip list
 * Define a list of directories that will skip the trashbin and therefore be deleted immediately.
 * Directory names are valid for all mount points, take care selecting the name.
 * Defined directory names can be in any location of the path and skip this directory from the trashbin.
 * All content of defined directories and their subdirectories s are affected.
 * Consider not using reserved directory names when using a storage system that provides snapshots.
 * It is not necessary to set a trailing slash, the directory name is enough.
 *
```

config/config.sample.php
```diff
@@ -578,6 +578,42 @@
...
+'trashbin_skip_directories' => [
+	'temp',
+],
+
+/**
+ * Define trashbin file extension skip list
+ * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately
+ * Values must not have a leading ".", otherwise corresponding files wont be detected.
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
```
mmattel, line 598:
```suggestion
 * Define trashbin file extension skip list
 * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately
 * Extension names are valid for all mount points, take care when selecting the names.
 * Values must not have a leading ".", otherwise corresponding files wont be detected.
 *
```

config/config.sample.php
```diff
@@ -578,6 +578,42 @@
...
+
+/**
+ * Define trashbin threshold size
+ * Define a threshold for files to skip the trashbin and delete immediately
+ * Once the size of a resource is greater than or equal the given value, the trashbin will be skipped.
+ *
+ * All positive numbers and zero are allowed. Append one of the following options directly and without space:
+ * B, K, KB, MB, M, GB, G, TB, T, PB, P
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
```
mmattel, line 613:
```suggestion
 * Define trashbin threshold size
 * Define a threshold for files to skip the trashbin and delete immediately
 * Once the size of a resource is greater than or equal the given value, the trashbin will be skipped.
 * File sizes are valid for all mount points, take care when defining the threshold.
 *
 * All positive numbers and zero is allowed. Append one of the following options directly and without space:
 * B, K, KB, MB, M, GB, G, TB, T, PB, P
 *
```
> and zero is allowed

means, you can not have a zero length file = no file at all (but directories only) ... hmmm ... does this make sense?

.htaccess
```diff
@@ -94,3 +94,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
jvillafanez, line 97: I don't think the changes in this file are needed.

config/config.sample.php
jvillafanez: @JanAckermann could you double-check with ".tar.gz" or ".tar.bz2" files? I'm not sure how it will be handled. I guess the expectation is to use those extensions, so if it isn't possible we should document it.

jvillafanez (changes requested):
the .htaccess file shouldn't be modified.

apps/files_trashbin/lib/TrashbinSkipChecker.php
```diff
@@ -0,0 +1,142 @@
...
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	protected function checkFileExtension(string $relativePath): bool {
+		$fileExtensionsToSkip = $this->config->getSystemValue('trashbin_skip_extensions', []);
+		$skipFile = \in_array(\pathinfo($relativePath, PATHINFO_EXTENSION), $fileExtensionsToSkip, true);
```
mmattel, line 112: Referring to comments from @jvillafanez 
https://github.com/owncloud/core/pull/38704#discussion_r633266666
https://github.com/owncloud/core/pull/38704#discussion_r633266666

This will definitely not catch `.tar.gz`
Also noted in https://www.php.net/manual/en/function.pathinfo.php

You need to extract the `basename` from `$relativePath` and use `strstr — Find the first occurrence of a string` on the basename to catch extensions starting with a dot at the first occurrence in the basename. Then you catch `.mkv` and `.tar.gz`

I would NOT propose to exclude an extension like `.tar.gz` but adopting the code to make it work.

JanAckermann: @mmattel @jvillafanez I am not quite sure if we can really catch this, dot separation in file names are not always extensions.
If we follow the mentioned example, a file named '2021.07.01.jpg' would give us the file extension: 07.01.jpg.

This would mean, that our config wouldn't have any effect here

```'trashbin_skip_extensions' => [
	'jpg'
]```

File: config/config.sample.php (hunk 19 above)
Comment (JanAckermann) [reply config/config.sample.php:587]: > Defined directory names can be in any location of the path and will be skipped from the trashbin.
This is not quite true, we do an exact match here inside the files path

File: apps/files_trashbin/lib/TrashbinSkipChecker.php (hunk 23 above)
Comment (jvillafanez) [reply apps/files_trashbin/lib/TrashbinSkipChecker.php:112]: What about something like "endsWith", with a leading "."?
By putting "jpg" we'd match files which end with ".jpg", and by putting "tar.bz2" we'd match files ending with ".tar.bz2". We could also match something like ".01.jpg" or ".myfile.gz", which we could consider as extensions although they aren't. It should be acceptable.
We'd need to take care about ".jpg" file (exact match), but it should be easy.

Comment (mmattel) [reply apps/files_trashbin/lib/TrashbinSkipChecker.php:112]: You are right,
`2021.07.01.tar.gz.` --> `07.01.tar.gz.`
But this would be eliminated when using:
`str_ends_with ( string $haystack , string $needle )`
or?

File: config/config.sample.php (hunk 19 above)
Comment (mmattel) [reply config/config.sample.php:587]: I fixed the suggestion and changed the text accordingly

File: apps/files_trashbin/lib/TrashbinSkipChecker.php (hunk 23 above)
Comment (jvillafanez) [reply apps/files_trashbin/lib/TrashbinSkipChecker.php]: For our case, it would be removed, yes. However, anyone could create an app which uses "07.01.tar.gz" as extension. Anyway, the setting will be set by the admin, and he won't set that kind of extension.
Although not technically correct, if it helps to simplify the code, I think the solution is good enough.

File: config/config.sample.php [hunk 24]
```diff
@@ -578,6 +578,43 @@
...
+	'mkv',
+],
+
+/**
+ * Define trashbin threshold size
+ * Define a threshold for files to skip the trashbin and delete immediately
+ * Once the size of a resource is greater than or equal the given value, the trashbin will be skipped.
+ * File sizes are valid for all mount points, take care when defining the threshold.
+ *
+ * All positive numbers and zero is allowed. Append one of the following options directly and without space:
```
jvillafanez, line 612: Did you check with negative values?
I think the code still works and it will cause the trashbin to skip all the files. We probably should clarify what happens because the admin could think that a negative value will disable the feature, which means that all files should go to the trashbin and won't be removed directly.
As long as it's documented, I'm ok with removing the files directly (no code changes)

apps/files_trashbin/tests/TrashbinSkipCheckerTest.php
```diff
@@ -0,0 +1,85 @@
...
+			->method('getSystemValue')
+			->willReturn([], ['pdf'], null);
+		$view = Filesystem::getView();
+
+		$this->assertTrue($this->trashbinSkipChecker->shouldSkipPath($view, 'somefolder/somefile.pdf'));
+	}
+
+	public function testSkipDueToFileSize() {
+		$folder = 'somefolder';
+		Filesystem::mkdir($folder);
```
jvillafanez, line 75: Better use a mock view not to rely on static calls.

config/config.sample.php
JanAckermann: To be honest, I would love to have a validator here, what do you think?

jvillafanez:
I think those will be the last changes for me.

apps/files_trashbin/tests/TrashbinSkipCheckerTest.php
JanAckermann: > Also check other calls to the Filesystem class in the tests.
Which ones?

config/config.sample.php
jvillafanez: Up to you.

apps/files_trashbin/tests/TrashbinSkipCheckerTest.php
```diff
@@ -0,0 +1,97 @@
...
+			->method('getMimeType')->willReturn('application/pdf');
+
+		$this->assertTrue($this->trashbinSkipChecker->shouldSkipPath($view, 'somefolder/somefile.pdf'));
+	}
+
+	public function testSkipDueToFileSize() {
+		$folder = 'somefolder';
+		$view = $this->createMock(View::class);
+		$view->mkdir($folder);
+		$view->file_put_contents($folder . '/somefile.pdf', 'file1');
```
jvillafanez, line 83: neither the `mkdir` nor the `file_put_contents` methods should be needed now.

JanAckermann: what do we use instead?

config/config.sample.php
JanAckermann: @jvillafanez would love to comeup with symfony/validator here (and for the future)... Any objections ?

apps/files_trashbin/tests/TrashbinSkipCheckerTest.php
jvillafanez: the method to be tested will use the mock view, so just configure the mock as you need. In fact, those methods shouldn't be doing anything.

config/config.sample.php
jvillafanez: It seems too complex... you're already using the `computerFileSize` method, so it should be returning a negative value if the size is negative. You can just check for that.

config/config.sample.php
```diff
@@ -578,6 +578,43 @@
...
+ */
+'trashbin_skip_directories' => [
+	'temp',
+],
+
+/**
+ * Define trashbin file extension skip list
+ * Define a list of file extensions to determine files that will skip the trashbin and therefore be deleted immediately.
+ * Extension names are valid for all mount points, take care when selecting the names.
+any subdirectores. Take care when selecting the names.
```
mmattel, line 597:
```suggestion
 * Extension names are valid for all mount points, take care when selecting the names.
```

mmattel, line 597:
```suggestion
```

config/config.sample.php
```diff
@@ -578,6 +578,43 @@
  */
 'trashbin_purge_limit' => 50,
 
+/**
+ * Define trashbin folder skip list
+ * Define a list of folders that will skip the trashbin and therefore be deleted immediately
+ * The setting expects folder names with or without trailing slash. All the content
+ * inside such folders will also skip the trashbin.
+ *
+ * WARNING: THIS SETTING IS SYSTEM WIDE AND COULD CAUSE DATA LOSS. USE THIS ONLY IF YOU KNOW WHAT YOU ARE DOING.
```
mmattel, line 587:
```suggestion
 * Define trashbin directory skip list
 * Define a list of directories that will skip the trashbin and therefore be deleted immediately.
 * Only defined directories and only in the root of a mount will skip for the trashbin. 
 * The setting expects folder names with or without trailing slash.
 * All the content inside such directories will also skip the trashbin.
 *
```

mmattel:
For some strange reasons, already committed text suggestions for config.php got corrupted.
For one smaller corruption, I changed the bad line and committed it directly.
The other reconstructed suggestion is multi line and I would like to get another eye on @JanAckermann

mmattel:
@jvillafanez 
> the .htaccess file shouldn't be modified.

This has been fixed by @JanAckermann, no changes in the file.
Pls re-review/approve.

mmattel (approved):
OK from the config.php POV
Code not tested

apps/files_trashbin/lib/TrashbinSkipChecker.php
```diff
@@ -0,0 +1,147 @@
...
+
+		if ($skipSizeThreshold === false) {
+			return false;
+		}
+
+		$computerSkipSizeThreshold = Util::computerFileSize($skipSizeThreshold);
+
+		if ($computerSkipSizeThreshold === false || $computerSkipSizeThreshold < 0) {
+			$this->logger->warning("Config trashbin_skip_size_threshold value $skipSizeThreshold is invalid");
+			return  false;
```
jvillafanez, line 137: Double space

jvillafanez (approved):
Tiny detail, maybe for the next time.

PR #38706, opened 2021-05-07
[WIP] Guzzle6 for core

phil-davis:
## Description
- [x] get adjusted code from old PR #38266 (that itself came from PRs #37011 #32357 )
- [x] Update guzzle and other dependencies in apps/files_external/3rdparty 

This is a recreation of PR #38266 

## Motivation and Context
Guzzle5 does not have official support on PHP 7.4 and probably never will have. Guzzle5 works on PHP 7.4
For future PHP versions (e.g. PHP 8.0) we might be forced to do this Guzzle update.

Note1: if we do this then apps that use Guzzle5 will also need updating to Guzzle6 and coordinated release of core and app versions etc.

Note2: guzzle is actually up to release https://github.com/guzzle/guzzle/releases/tag/7.2.0 - so if we bump guzzle we might want to bump it all the way to major version 7.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
Federated acceptance tests are failing. 

sqlite+scality unit tests with messages like `TypeError: Argument 3 passed to GuzzleHttp\Client::request() must be of the type array, object given, called in /drone/src/lib/composer/guzzlehttp/guzzle/src/Client.php on line 95`

Needs investigating when someone wants to move forward with this.

phil-davis:
This is here as a "record" of past code for updating Guzzle from v5 to v6. It should provide a good starting point if someone does this some day.

PR #38712, opened 2021-05-07
improve-public-share-interface

AlexAndBear:
## Description
This adjustment provides a new way to add a public share to ownCloud.
A new button will be displayed where you can easily add the share to the current
server without entering the string, this requires less user input.
Next to this button, a dropdown menu will be displayed,
where you can see the current server or change the server.
Hitting change a server will now display a prompt with a detailed explanation
and more space to enter another server address.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4393

### Tested on
- [x] Firefox 
- [x] Chrome
- [x] IE11 

### Before:
image

image

### After:
image

image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/js/public.js
+++ b/apps/files_sharing/js/public.js
@@ -149,8 +149,7 @@ OCA.Sharing.PublicApp = {
 			img.attr('src', OC.Util.replaceSVGIcon(mimetypeIcon));
 			img.attr('width', 128);
 			img.appendTo('#imgframe');
-		}
-		else if (previewSupported === 'true') {
+		} else if (previewSupported === 'true') {
 			$('#imgframe > video').attr('poster', OC.filePath('files_sharing', 'ajax', 'publicpreview.php') + '?' + OC.buildQueryString(params));
 		}
 
@@ -162,7 +161,7 @@ OCA.Sharing.PublicApp = {
 
 				var filesPart = '';
 				if (_.isArray(filename)) {
-					_.each(filename, function(name) {
+					_.each(filename, function (name) {
 						filesPart += '&files[]=' + encodeURIComponent(name);
 					});
 				} else {
@@ -171,7 +170,7 @@ OCA.Sharing.PublicApp = {
 				return base + filesPart;
 			};
 
-			this.fileList.getUploadUrl = function(fileName, dir) {
+			this.fileList.getUploadUrl = function (fileName, dir) {
 				if (_.isUndefined(dir)) {
 					dir = this.getCurrentDirectory();
 				}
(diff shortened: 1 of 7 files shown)


JanAckermann:
TODO:
* Create Docs issue if needed
* Remove old code
* Fix remote without http/https

JanAckermann:
@JammingBen pre-review is welcome :rabbit:

apps/files_sharing/js/public.js
```diff
@@ -313,47 +405,51 @@ OCA.Sharing.PublicApp = {
...
-			else {
-				$('#save-button-confirm')
-				.removeClass("icon-confirm")
-				.addClass("icon-loading-small");
+		var that = this;
 
+		var toggleLoading = function (){
+			var loading = that.$saveToOcButton.find('#save-to-oc-button-loading').length;
+
+			if(!loading){
```
JammingBen, line 413: It should be `if(loading) {}` here, shouldn't it?

JanAckermann: no, why?

JammingBen: Because the next line indicates loading: `that.$saveToOcButton.html(t('files_sharing', 'Loading...') + '');`. So I expect `loading` to be true. Maybe I'm just confused about the variable naming 😄  It's meant to be something like `isLoaded` then?

JanAckermann: if it is not loading we add a loading indicator, I don't get it :D

JanAckermann:
@phil-davis Patrick told me that, you are a quite good text reviewer, can you look at the texts for the oc.dialogs.drop maybe :) ?

VicDeo (changes requested):
Cool improvement but it needs to respect the setups where OC is located in the subfolder of the server webroot:
1. Install OC into `localhost/owncloud`
2. Create a public folder share
3. Open it as a guest user
4. Click 'Add to localhost'

### Expected
login screen

### Actual 
HTTP 404

as it requests 
http://localhost/index.php/apps/files#  instead of
http://localhost/owncloud/index.php/apps/files#

JanAckermann:
@VicDeo good catch, thx!

.htaccess
```diff
@@ -94,3 +94,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
+
+ErrorDocument 403 //core/templates/403.php
```
VicDeo, line 99: @JanAckermann accidental commit?

JanAckermann: Yep :/

VicDeo (approved):

PR #38719, opened 2021-05-10
use filter_var instead of (bool)

AlexAndBear:
## Description
String values like "true" and "false" were always converted to true
when creating a tag via API. We now use filter_var() to fix this
behavior.

## Related Issue

- Related https://github.com/owncloud/files_backup/issues/37

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/SystemTag/SystemTagPlugin.php
+++ b/apps/dav/lib/SystemTag/SystemTagPlugin.php
@@ -177,15 +177,15 @@ class SystemTagPlugin extends \Sabre\DAV\ServerPlugin {
 		$userEditable = false;
 
 		if (isset($data['userVisible'])) {
-			$userVisible = filter_var($data['userVisible'], FILTER_VALIDATE_BOOLEAN);
+			$userVisible = \filter_var($data['userVisible'], FILTER_VALIDATE_BOOLEAN);
 		}
 
 		if (isset($data['userAssignable'])) {
-			$userAssignable = filter_var($data['userAssignable'], FILTER_VALIDATE_BOOLEAN);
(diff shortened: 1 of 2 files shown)


phil-davis:
I guess this means that there are not currently any API acceptance tests that create tags with different settings `userVisible` etc.
Does it need `systemtags_management` app to enable the endpoint features? Or are all the parts of the "create tag" API provided in core?

JammingBen:
> I guess this means that there are not currently any API acceptance tests that create tags with different settings userVisible etc.

At least not with the string values `"true"` and `"false"` it seems.

> Does it need systemtags_management app to enable the endpoint features? Or are all the parts of the "create tag" API provided in core?

You can still use the API with the `systemtags` app disabled.

phil-davis:
https://github.com/owncloud/core/blob/master/tests/TestHelpers/TagsHelper.php#L258
`validateTypeOfTag` sets the `userVisible` etc to either the string "1" or "0". Those work "as expected" when casting to boolean. 

Is the API supposed to support callers sending "true"/"false"? Or are we just being nice here and handling that?

JanAckermann:
>Is the API supposed to support callers sending "true"/"false"? Or are we just being nice here and handling that?
According to the docs, it is supposed to accept these values:
https://doc.owncloud.com/server/developer_manual/webdav_api/tags.html#create-tags

Tests can even be truthful because we do a lot with the incoming values:
https://github.com/owncloud/core/blob/master/lib/private/SystemTag/SystemTagManager.php#L212

jvillafanez (approved):

PR #38720, opened 2021-05-10
[tests-only] Feature to debug specific test suites in CI

saw-jan:
## Description
adds feature to be able to run specific test suites in CI

## Related Issue
part of https://github.com/owncloud/QA/issues/670

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1354,7 +1354,9 @@ def acceptance(ctx):
 		'includeKeyInMatrixName': False,
 		'runAllSuites': False,
 		'numberOfParts': 1,
-		'skip': False
+		'skip': False,
+		'debugSuites': [],
+		'skipExceptParts': []
 	}
 
 	if 'defaults' in config:
@@ -1369,6 +1371,14 @@ def acceptance(ctx):
 				suites[suite] = suite
 		else:
 			suites = matrix['suites']
+
+		if 'debugSuites' in matrix and len(matrix['debugSuites']) != 0:
+			if type(matrix['debugSuites']) == "list":
+				suites = {}
+				for suite in matrix['debugSuites']:
+					suites[suite] = suite
+			else:
+				suites = matrix['debugSuites']
 
 		for suite, alternateSuiteName in suites.items():
 			isWebUI = suite.startswith('webUI')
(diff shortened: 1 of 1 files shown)


.drone.star
```diff
@@ -129,6 +133,8 @@ config = {
 				'apiWebdavUpload1',
 				'apiWebdavUpload2',
 			],
+			'skip': False,
+			'debugSuites': []
```
kiranparajuli589, line 137: imo default values needn't be used in `config`

saw-jan: Yeah, I place those so that anyone would know where to skip and debug suites

phil-davis: We do want the skip false in the config section - that makes it easy for people to change to skip true. I guess we will learn how to insert the other special debugSuites and skipExceptParts.

.drone.star
```diff
@@ -1370,6 +1372,11 @@ def acceptance(ctx):
 		else:
 			suites = matrix['suites']
 
+		if 'debugSuites' in matrix and len(matrix['debugSuites']) != 0:
+			suites = {}
+			for suite in matrix['debugSuites']:
+				suites[suite] = suite
```
phil-davis, line 1378:
```suggestion
			if type(matrix['suites']) == "list":
				suites = {}
				for suite in matrix['debugSuites']:
					suites[suite] = suite
			else:
				suites = matrix['debugSuites']
```

There are 2 ways to specify a list of suites:
- as a dict key-value list - the keys are the suite names, and the values are the shortened names to use in the pipeline name
- as a list (ordinary array) - the items are the suite names, that are also used directly in the pipeline names

IMO the code here needs to handle both.

.drone.star
```diff
@@ -1370,6 +1372,14 @@ def acceptance(ctx):
 		else:
 			suites = matrix['suites']
 
+		if 'debugSuites' in matrix and len(matrix['debugSuites']) != 0:
+			if type(matrix['suites']) == "list":
```
phil-davis, line 1376:
```suggestion
			if type(matrix['debugSuites']) == "list":
```

I guess we had better not assume that `suites` and `debugSuites` will always have the same type.

phil-davis (approved):

PR #38721, opened 2021-05-10
[tests-only] Test creating tags with true/false string settings

phil-davis:
PR #38719 adjusted the server code so that the attributes of tags (`userVisible` `userAssignable` `userEditable`) will process "boolean" strings like "true" and "false" to the appropriate boolean `true` or `false` value in incoming API requests. 

We have some client(s) that can send strings "true"/"false" instead of "1"/"0".

This PR adjusts the test code so that it sends strings "true"/"false" by default, and optionally "1"/"0" for the attributes of tags when creating tags.

~We could have "duplicated" test scenarios that try creating tags both ways? But that seems a bit overkill?~

The "basic" tag creation tests have scenarios that check the previous "0"/"1" behaviour as well as the newer "true"/"false" behaviour.

### DIFF
--- a/tests/TestHelpers/TagsHelper.php
+++ b/tests/TestHelpers/TagsHelper.php
@@ -178,9 +178,9 @@ class TagsHelper extends \PHPUnit\Framework\Assert {
 	 * @param string $user
 	 * @param string $password
 	 * @param string $name
-	 * @param string $userVisible "true" or "false"
-	 * @param string $userAssignable "true" or "false"
-	 * @param string $userEditable "true" or "false"
+	 * @param string $userVisible "true", "1" or "false", "0"
+	 * @param string $userAssignable "true", "1" or "false", "0"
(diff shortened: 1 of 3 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/core/29892/82/13 apiTags tests fail - the ones that expected a tag to have some attribute set to false. That is good to know.

individual-it (commented):
If both is valid I feel both should be tested
firstly of course  #38719 should have unit tests to check both possibilities
but to make sure ocis will react the same way one day, we need somewhere tests that try to send various values and make sure they are converted to the correct tag

phil-davis:
I will adjust so that both are done.

phil-davis:
> I will adjust so that both are done.

@individual-it done - please review.

individual-it (approved):

Issue #35741, opened 2019-07-02
fix path encoding problems in tests steps


individual-it:
some steps like like `Then user "user1" should see the following elements` need the path to be encoded in the step definition e.g. `%20` for a space
```
    Then user "user1" should see the following elements
      | /PARENT%20(2)/           |
      | /PARENT%20(2)/parent.txt |
```

Other steps don't need that e.g. `Then as "user1" folder "/merge-test-inside-twogroups-perms (2)" should not exist`

This is particular bad when checking for something that does not exist
```
    Then user "user1" should see the following elements
      | /PARENT (2)/           |
```
would always pass but be :snake: :oil_drum: 

We need to make it uniform

saw-jan:
Done in PR https://github.com/owncloud/core/pull/38734
Closing the issue for now

PR #38734, opened 2021-05-14
[tests-only] Fix path encoding in tests steps

saw-jan:
## Related Issue
- https://github.com/owncloud/core/issues/35741

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -1766,6 +1766,7 @@ trait WebDav {
 			// Allow the table of expected elements to have entries that do
 			// not have to specify the "implied" leading slash, or have multiple
 			// leading slashes, to make scenario outlines more flexible
+			$expectedElement = $this->encodePath($expectedElement);
 			$expectedElement = "/" . \ltrim($expectedElement, "/");
 			$webdavPath = "/" . $this->getFullDavFilesPath($user) . $expectedElement;
 			$element = $responseXmlObject->xpath(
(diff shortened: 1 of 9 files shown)


tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -3556,7 +3557,11 @@ private function deleteUpload($user, $id, $headers) {
 	 */
 	public function encodePath($path) {
 		// slashes need to stay
-		return \str_replace('%2F', '/', \rawurlencode($path));
+		$encodedPath = \str_replace('%2F', '/', \rawurlencode($path));
+		// do not encode '(' and ')'
+		$encodedPath = \str_replace('%28', '(', $encodedPath);
+		$encodedPath = \str_replace('%29', ')', $encodedPath);
```
dpakach, line 3563: why is this needed? dont we need to encode these characters in the URL?

saw-jan: Yeah, `()` should not be encoded but `rawurlencode` encodes it.

dpakach (commented):
lgtm 👍🏼

kiranparajuli589:
lgtm 👍🏼

tests/acceptance/features/bootstrap/WebDav.php
phil-davis: yes, https://www.php.net/manual/en/function.rawurlencode.php does more than what we need here. It does not have options to just encode a subset of special characters. At least the special exceptions are in a common method here `encodePath` so that everything else can call it consistently.

phil-davis (approved):

PR #38742, opened 2021-05-18
Add row for guest users to command occ user:report

AlexAndBear:
## Description
Enhancement: Command occ user:report shows additional row for guests

With this improvement, a new row will be rendered for guest user count on the
occ user:report command.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/3467

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Command/User/Report.php
+++ b/core/Command/User/Report.php
@@ -25,6 +25,8 @@
 
 namespace OC\Core\Command\User;
 
+use OC\Helper\UserTypeHelper;
+use OCP\IUser;
 use OCP\IUserManager;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
@@ -35,11 +37,16 @@ class Report extends Command {
 	/** @var IUserManager */
 	protected $userManager;
 
+	/** @var UserTypeHelper */
+	protected $userTypeHelper;
+
 	/**
 	 * @param IUserManager $userManager
+	 * @param UserTypeHelper $userTypeHelper
 	 */
-	public function __construct(IUserManager $userManager) {
+	public function __construct(IUserManager $userManager, UserTypeHelper $userTypeHelper) {
 		$this->userManager = $userManager;
+		$this->userTypeHelper = $userTypeHelper;
 		parent::__construct();
 	}
 
@@ -53,6 +60,8 @@ class Report extends Command {
 		$table = new Table($output);
 		$table->setHeaders(['User Report', '']);
 		$userCountArray = $this->countUsers();
+		$guests = 0;
+
 		if (!empty($userCountArray)) {
 			$total = 0;
 			$rows = [];
@@ -61,6 +70,14 @@ class Report extends Command {
 				$rows[] = [$classname, $users];
 			}
 
+			$this->userManager->callForAllUsers(function (IUser $user) use (&$guests) {
+				if ($this->userTypeHelper->isGuestUser($user->getUID()) === true) {
+					$guests++;
+				}
+			});
+
+			$rows[] = [' '];
+			$rows[] = ['guest users', $guests];
 			$rows[] = [' '];
 			$rows[] = ['total users', $total];
 		} else {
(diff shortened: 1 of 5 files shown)


JanAckermann:
@jvillafanez not quite sure if we need to introduce \OC::$server->getUserTypeHelper

jvillafanez:
Double-check the following scenario:
1. Using the "guests" app, create some guests
2. Check `occ user:report` for guests
3. Disable the "guests" app
4. Recheck `occ user:report`

I think the guests will be counted as normal users in step 4. It doesn't seem right.

Honestly, I don't know how much knowledge core has about guests, nor what is the exact relationship with the guests app. It seems that we're adding responsibilities into core that wasn't designed for.
We might need to redesign this part to clarify responsibilities and relationships among components.

JanAckermann:
@jvillafanez We can easily overcome this issue by adjusting the UserTypeHelper::isGuestUser, it pre-checks if the app is enabled and if not, it returns false.

I introduced this helper and feel confident with removing the check

JanAckermann:
@jvillafanez changed check to find out if a user is guest user, now your provided example give correct results

core/Command/User/Report.php
```diff
@@ -25,6 +25,9 @@
 
 namespace OC\Core\Command\User;
 
+use Cassandra\Type\UserType;
```
jvillafanez, line 28: Do we need this?

core/Command/User/Report.php
```diff
@@ -61,6 +71,14 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 				$rows[] = [$classname, $users];
 			}
 
+			$this->userManager->callForAllUsers(function (IUser $user) use (&$guests) {
+				if ($this->userTypeHelper->isGuestUser($user->getUID()) === true) {
+					$guests ++;
```
jvillafanez, line 76: `$guests++` (without space) is more common.

lib/private/Helper/UserTypeHelper.php
```diff
@@ -21,42 +21,31 @@
...
 	/**
-	 * Checks whether given uid belongs to the user created by the guests app
+	 * Checks whether given uid belongs to the user created for example by
+	 * guests app
 	 *
 	 * @param string $uid
 	 * @return bool
 	 */
 	public function isGuestUser($uid) {
-		$guestsAppEnabled = $this->appManager->isEnabledForUser('guests');
```
jvillafanez, line 53: @VicDeo any expected side effect for this removal? I don't remember if there was a reason for this check. Removing it looks fine to me.

lib/private/Helper/UserTypeHelper.php
```diff
@@ -21,42 +21,31 @@
...
-		if ($guestsAppEnabled) {
-			return (bool) $this->config->getUserValue(
-				$uid, 'owncloud', 'isGuest', false
-			);
-		}
-		return false;
+		return (bool) $this->config->getUserValue(
+				$uid, 'owncloud', 'isGuest', false);
```
jvillafanez, line 48: Format looks weird here

jvillafanez:
I'd wait a bit for confirmation on https://github.com/owncloud/core/pull/38742#discussion_r634316259 . The rest is mostly style things easy to fix.

mmattel:
doc relevant, pls file an issue in docs close to merging.
pls add an example output which eases our documentation effort 😄

JanAckermann:
@mmattel can you link the relevant paragraph?

JanAckermann:
@VicDeo your opinion is highly required :rocket:

lib/private/Helper/UserTypeHelper.php
VicDeo: @jvillafanez well... my initial logic was like this: if the `guests` app is disabled the instance has no guests at all. Because  the guests capability is provided by the respective app

jvillafanez: Yes, that was my idea too. Technically, it makes sense.
We should properly implement something in core to provide different user types

jvillafanez (approved):
Good to go for the short term

PR #38746, opened 2021-05-19
[WIP] Phpunit 9

phil-davis:
Run the unit tests with phpunit 9 to see what works (or not)

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -25,12 +25,11 @@ config = {
 	'phpunit': {
 		'allDatabases' : {
 			'phpVersions': [
-				'7.2',
+				'7.3',
 			]
 		},
 		'reducedDatabases' : {
 			'phpVersions': [
-				'7.3',
 				'7.4',
 			],
 			'databases': [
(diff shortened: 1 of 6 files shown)


phil-davis:
I raised issue https://github.com/sebastianbergmann/phpunit/issues/4671 to find out what the problem is with "Schema for PHPUnit 9.2 is not available"

phil-davis:
https://drone.owncloud.com/owncloud/core/30189/8/8
```
There were 307 warnings:

1) Test\AppFramework\Db\MapperTest::testFindQuery
The at() matcher has been deprecated. It will be removed in PHPUnit 10. Please refactor your test to not rely on the order in which methods are invoked.

2) Test\AppFramework\Db\MapperTest::testFindEntity
The at() matcher has been deprecated. It will be removed in PHPUnit 10. Please refactor your test to not rely on the order in which methods are invoked.
```

Most of them are about the `at()` matcher. That needs a big refactoring that will have to be done some time in the future when we really do move to phpunit 9, but not today!

PR #38748, opened 2021-05-19
Improve mobile UI for files list 

AlexAndBear:
## Description
This improvement solves the problem that files and folders with a longer name
than 10 chars will cut off to 7 chars by hiding non-essential meta data.
This is unwieldy to navigate through files and folders.
As well the uploading indicator text on a folders table row will now be
displayed at the beginning to avoid UI glitches.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

### Before
before

### After 
after

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/css/mobile.css
+++ b/apps/files/css/mobile.css
@@ -19,11 +19,6 @@ table.multiselect thead {
 	padding-left: 0;
 }
 
-/* restrict length of displayed filename to prevent overflow */
-table td.filename .nametext {
-	width: 130px;
-}
-
 #fileList a.action.action-menu img {
 	padding-left: 0;
 }
@@ -46,7 +41,7 @@ table td.filename .nametext {
 
 /* ellipsis on file names */
 table td.filename .nametext .innernametext {
-	max-width: 80%;
+	max-width: 90%;
 }
 
 /* proper notification area for multi line messages */
@@ -68,6 +63,15 @@ table td.filename .nametext .innernametext {
 }
 #uploadprogressbar .mobile {
 	display: block !important;
+}
+
+/* hide mobile table content*/
+table #headerSize, #headerDate, .filesize, .modified, .date {
+	display: none;
+}
+
+#fileList a.action.action-menu.permanent {
+	padding-right: 14px;
 }
 
 }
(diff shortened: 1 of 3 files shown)


JanAckermann:
@pmaier1 cool with it?

apps/files/css/mobile.css
```diff
@@ -70,4 +66,14 @@ table td.filename .nametext .innernametext {
 	display: block !important;
 }
 
+/* hide mobile table content*/
+table #headerSize, #headerDate, .filesize, .modified, .date {
+		display: none;
```
ChrisEdS, line 71: Code is indented incorrectly?

ChrisEdS (approved):
Technically approved.

phil-davis:
Thinking out loud - I wonder what we should do about testing that the webUI "behaves nicely" when being displayed on a smaller device/window? It would be possible to run some CI with a smaller-sized browser window, but we can also just leave it to manual smoke-test to be done during release QA.

JanAckermann:
@phil-davis Do we have similar tests already, how much effort do you think is required for a full coverage?

phil-davis:
> @phil-davis Do we have similar tests already, how much effort do you think is required for a full coverage?

In web we set an env var, for example, `"SCREEN_RESOLUTION": "375x812"` to test with a mobile-phone-sized screen.

Here in the "traditional" oC10 UI we can do similar. Code like this resizes the window:
`$this->getSession()->resizeWindow(375, 812, 'current');`

We can add some code to do that if `SCREEN_RESOLUTION` is set. And then run whatever subset of test scenarios we like in CI with whatever `SCREEN_RESOLUTION` settings.

The "problem" is that we keep adding more CI pipelines! So maybe we just do it for a set of "smokeTest" scenarios.

JanAckermann:
@pmaier1 waiting for your go :+1:

pmaier1:
> @pmaier1 waiting for your go

Fine for me :+1:

PR #38751, opened 2021-05-19
implement fallback response type for ocm discover

karakayasemi:
## Description
Some ocm-provider implementations are slightly different than us. This pr implements fallback for different ocm-provider API responses.

## Related Issue
https://github.com/owncloud/enterprise/issues/4522#issuecomment-841884476

## Motivation and Context
Prevent unnecesary error logs.

## How Has This Been Tested?
Tested federated sharing functionality manually with different ocm providers.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/lib/DiscoveryManager.php
+++ b/apps/federatedfilesharing/lib/DiscoveryManager.php
@@ -141,22 +141,26 @@ class DiscoveryManager {
 		$decodedService = $this->makeRequest($remote . '/ocm-provider/');
 		if (!empty($decodedService) && $decodedService['enabled'] === true) {
 			$discoveredServices['ocm'] = $decodedService['endPoint'];
-			$shareTypes = $decodedService['shareTypes'];
-			foreach ($shareTypes as $type) {
+			$types = [];
+			if (isset($decodedService['shareTypes'])) {
+				$types = $decodedService['shareTypes'];
+			} elseif (isset($decodedService['resourceTypes'])) {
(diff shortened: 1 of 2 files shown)


apps/federatedfilesharing/lib/DiscoveryManager.php
```diff
@@ -141,22 +141,30 @@ private function ocmDiscover($remote) {
...
-					$discoveredServices['webdav'] = $type['protocols']['webdav'];
+			if (isset($decodedService['shareTypes'])) {
+				$shareTypes = $decodedService['shareTypes'];
+				foreach ($shareTypes as $type) {
+					if ($type['name'] == 'file') {
+						$discoveredServices['webdav'] = $type['protocols']['webdav'];
+					}
+				}
+			} elseif (isset($decodedService['resourceTypes'])) {
+				$resourceTypes = $decodedService['resourceTypes'];
```
phil-davis, line 152:
```suggestion
				$shareTypes = $decodedService['resourceTypes'];
```
And then the 2 copies of `foreach` can become a single loop after the if/elseif/else. (To protect, it would need `$shareTypes = []` in an `else` so that the loop always has some iterable in `$shareTypes`

karakayasemi: @phil-davis changed with sth similar. Thanks.

jvillafanez (approved):

PR #38765, opened 2021-05-21
[tests-only] reduce Apache output log for tests

individual-it:
## Description
dont't use apache debug mode to serve core

## Motivation and Context
this logs are usually not so important and that should save a couple of thousand line of logs in every pipeline

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


individual-it:
OK, that does not make any real difference because the access log is not triggered by `debug` but by `CustomLog /dev/stdout combined` in the apache log

PR #38769, opened 2021-05-21
Add excludeFromApplicationSelectMenuFlag

AlexAndBear:
## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/fileactions.js
+++ b/apps/files/js/fileactions.js
@@ -153,7 +153,8 @@
 				iconClass: action.iconClass,
 				permissions: action.permissions,
 				type: action.type || FileActions.TYPE_DROPDOWN,
-				altText: action.altText || ''
+				altText: action.altText || '',
+				excludeFromApplicationSelectMenu: action.excludeFromApplicationSelectMenu || false
 			};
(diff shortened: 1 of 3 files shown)


JanAckermann:
@jvillafanez Ideally the fileAction should know nothing about the ApplicationSelectMenu so I am searching for a more generic name, any idea ?

jvillafanez:
`isHiddenAction` maybe? I'm not sure if that will be too generic... anyway, it will likely require some additional comment to know how it's expected to be used.

Another option could be to use some kind of callback so the caller can send some context for the app to decide if the action should be excluded or not, something like:
```
excludeFrom(context) {
  return context.name === 'applicationMenu';
}
.......
context = {'name': 'applicationMenu'};
if (myaction.excludeFrom && myaction.excludeFrom(context)) {
  .....
}
```
This way it seems easier to reuse the idea in other places.

PR #38774, opened 2021-05-25
Add action to resend the invitation email for a user

JammingBen:
## Description
Resend invitation email
Implemented an action to resend the invitation email for a user that hasnever been logged in yet.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4577

image

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket raised: <link> @mmattel required ? 

### DIFF
--- a/tests/acceptance/features/webUIAddUsers/addUsers.feature
+++ b/tests/acceptance/features/webUIAddUsers/addUsers.feature
@@ -91,6 +91,27 @@ Feature: add users
       | guiusr1  | simple user-name      |
       | a@-+_.'b | complicated user-name |
 
+  @smokeTest @skipOnLDAP @skipOnOcV10.6 @skipOnOcV10.7
+  Scenario Outline: user sets his own password after being created with an Email address only and invitation link resend
+    When the administrator creates a user with the name "<username>" and the email "guiusr1@owncloud" without a password using the webUI
+    And the administrator resends invitation email for user with the name "<username>" using the webUI
+    And the administrator logs out of the webUI
+    And the user follows the password set link received by "guiusr1@owncloud" in Email number 2 using the webUI
+    Then the user should see an error message saying "The token provided is invalid."
+    And the user follows the password set link received by "guiusr1@owncloud" in Email number 1 using the webUI
+    And the user sets the password to "%regular%" and confirms with the same password using the webUI
+    Then the user should be redirected to the login page
+    And the email address "guiusr1@owncloud" should have received an email with the body containing
+      """
+      Password changed successfully
+      """
+    When the user logs in with username "<username>" and password "%regular%" using the webUI
+    Then the user should be redirected to a webUI page with the title "Files - %productname%"
+    Examples:
+      | username | comment               |
+      | guiusr1  | simple user-name      |
+      | a@-+_.'b | complicated user-name |
+
   @skipOnOcV10.3
   Scenario Outline: user sets his own password but retypes it wrongly after being created with an Email address only
     When the administrator creates a user with the name "<username>" and the email "guiusr1@owncloud" without a password using the webUI
--- a/changelog/unreleased/38774
+++ b/changelog/unreleased/38774
@@ -1,0 +1,7 @@
+Enhancement: Resend invitation email
+
+Implemented an action to resend the invitation email for a user that has
+never been logged in yet.
+
+https://github.com/owncloud/core/pull/38774
+https://github.com/owncloud/enterprise/issues/4577
--- a/settings/css/settings.css
+++ b/settings/css/settings.css
@@ -747,3 +747,7 @@ li > a.icon-settings {
 div.app-settings .section {
 	margin-bottom: 0;
 }
+
+.action.resendInvitationEmail img {
+	vertical-align: text-bottom;
+}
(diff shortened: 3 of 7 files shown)


JanAckermann:
@phil-davis @JammingBen acceptance tests needed ?

phil-davis:
> @phil-davis @JammingBen acceptance tests needed ?

We have existing test scenarios like: https://github.com/owncloud/enterprise/issues/4454
```
  Scenario Outline: user sets his own password after being created with an Email address only
    When the administrator creates a user with the name "<username>" and the email "guiusr1@owncloud" without a password using the webUI
    And the administrator logs out of the webUI
    And the user follows the password set link received by "guiusr1@owncloud" using the webUI
    And the user sets the password to "%regular%" and confirms with the same password using the webUI
    Then the user should be redirected to the login page
    And the email address "guiusr1@owncloud" should have received an email with the body containing
      """
      Password changed successfully
      """
    When the user logs in with username "<username>" and password "%regular%" using the webUI
    Then the user should be redirected to a webUI page with the title "Files - %productname%"
    Examples:
      | username | comment               |
      | guiusr1  | simple user-name      |
      | a@-+_.'b | complicated user-name |
```

We could add a scenario where the user "forgets" and the email needs to be sent again.

It will need some additions to the page object for the users page, so that the test can find where to click to re-send the email. Who should do that? (I can get someone to look at it tomorrow, if you like)

JammingBen:
> It will need some additions to the page object for the users page, so that the test can find where to click to re-send the email. Who should do that? (I can get someone to look at it tomorrow, if you like)

We take care 👍

ChrisEdS:
Will this action only be showed when the user was invited via email initially?

JanAckermann:
> Will this action only be showed when the user was invited via email initially?
Nope

tests/acceptance/features/webUIAddUsers/addUsers.feature
```diff
@@ -91,6 +91,25 @@ Feature: add users
       | guiusr1  | simple user-name      |
       | a@-+_.'b | complicated user-name |
 
+  @smokeTest @skipOnLDAP @skipOnOcV10.3
```
JanAckermann, line 94: @phil-davis should we change the @skipOnOCV part?

phil-davis, line 94:
```suggestion
  @smokeTest @skipOnLDAP @skipOnOcV10.6 @skipOnOcV10.7
```

Yes. This test will only pass on 10.8 (current master). So skip on recent releases that we care to test against (some oC10 apps test against latest and one earlier release)

tests/acceptance/features/webUIAddUsers/addUsers.feature
```diff
@@ -91,6 +91,25 @@ Feature: add users
...
+  Scenario Outline: user sets his own password after being created with an Email address only and invitation link resend
+    When the administrator creates a user with the name "<username>" and the email "guiusr1@owncloud" without a password using the webUI
+    And the administrator resends invitation email for user with the name "<username>" using the webUI
+    And the administrator logs out of the webUI
+    And the user follows the password set link received by "guiusr1@owncloud" using the webUI
```
phil-davis, line 99: There should now be 2 emails waiting. Probably this scenario is going to pass even if the "resend click" does not work.
Maybe we can add a step that specifically follows the second link:
```suggestion
    And the user follows the second password set link received by "guiusr1@owncloud" using the webUI
```
The code for that step can specifically check that there are 2 emails, and then use the 2nd email.

JanAckermann:
@phil-davis now we also cover, that the first link have an expired/invalid token. Thanks for the hint!

phil-davis (approved):
LGTM

jvillafanez:
Maybe a designer should check the icon. It isn't crystal clear to me that you'll send an email. The color is also different from the trashbin icon.

jvillafanez (approved):
Code looks good

JanAckermann:
@jvillafanez  there is a gray icon available, will have a look. In general NC has the same Icon, I am OK with it.

AlexAndBear (changes requested):
- [x] Add title attribute for tooltip
- [x] Make icon look similar to delete

AlexAndBear (approved):

mmattel:
Maybe stupid but why not in addition a occ command set where you can trigger to resend an invite to either a named or all never logged in like:
```
occ resend-invite:list
occ resend-invite:send --all
occ resend-invite:send <name>
```

JammingBen:
> Maybe stupid but why not in addition a occ command set where you can trigger to resend an invite to either a named or all never logged in like:

I personally find it more intuitive via WebUI. The admin creates users via UI, therefore it makes sense to place this action there, too. These commands could be a nice addition though.

JanAckermann:
@mmattel can you point out a real-world use-case for the command?

phil-davis:
Note: IMO `occ user:add` does not currently have the ability to specify just the email address, and have an invitation email sent to the new user. It seems to always want a password. So maybe there are no admins currently using the command line to create users in this way.

PR #38775, opened 2021-05-25
Fix ExpireSharesJob so that it only expires shares that expired yesterday or earlier

phil-davis:
## Description
`ExpireSharesJob` is currently deleting shares that have an expiration date in the past of "now". For example, if the job runs at 2021-05-25:11:30 then a share that has an expiration date set to "2021-05-25" will be "expired" and get removed. But actually the "meaning" of expiring on "2021-05-25" is understood to be that the share works all of that day, right up to and including "2021-05-25:23:59:59"

https://github.com/owncloud/core/blob/master/lib/private/Share20/Manager.php#L185 has the real-time logic of this:
```
	private static function shareHasExpired($share) {
		$expirationDate = $share->getExpirationDate();
		return ($expirationDate !== null) && ($expirationDate < new \DateTime("today"));
	}
```

That compares the expiration date of the share to the "zero" hour of "today", and only considers the share expired if the expiration date is strictly less than the beginning of "today". So a share that has an expiration date "today" will not actually expire until the next day starts.

This PR adjusts `ExpireSharesJob` so that it uses the same logic.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4324

## How Has This Been Tested?
Unit test.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/ExpireSharesJob.php
+++ b/apps/files_sharing/lib/ExpireSharesJob.php
@@ -84,8 +84,8 @@ class ExpireSharesJob extends TimedJob {
 	 */
 	public function run($argument) {
 		//Current time
-		$now = new \DateTime();
-		$now = $now->format('Y-m-d H:i:s');
+		$today = new \DateTime("today");
+		$today = $today->format('Y-m-d H:i:s');
 
 		/*
 		 * Expire file shares only (for now)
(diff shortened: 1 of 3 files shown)


micbar (approved):
Thanks! LGTM

PR #38783, opened 2021-05-27
Switch order to prevent double-replacement in the query

jvillafanez:
## Description
The path used in the replacement for the path_hash column was the "new" one already replaced in the previous "set" clause. This was causing problems since the replacement could also happen in that new path, and could generate a wrong md5.
Switching the order makes sure the path used in the replace for both the path and path_hash columns will be the same

## Related Issue
https://github.com/owncloud/encryption/issues/276

## How Has This Been Tested?
Manually tested using `occ files:transfer-ownership`
1. user1 uploads a file
2. user1 shares the file with user3
3. run `occ files:transfer-ownershp user1 user2`

Previously, due to the wrong hash, a duplicate of the transfer file was generated in the DB, which was visible in the user2 account. This duplicate was causing side effects
This PR fixes that. The behaviour is the expected one.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -609,15 +609,15 @@ class Cache implements ICache {
 				if (\intval($versionArray[0]) < 12) {
 					$sql = 'UPDATE `*PREFIX*filecache`
 						SET `storage` = :targetStorageId,
-							`path` = REPLACE(`path`, :sourcePath, :targetPath),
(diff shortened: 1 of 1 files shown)


IljaN (approved):
Failing tests pass :heavy_check_mark:

jvillafanez:
Linking with https://github.com/owncloud/core/pull/38649 since it's a bug fix for that PR. No need for changelog because target PR hasn't been released yet and this one is a follow up of the other.

PR #38793, opened 2021-05-31
Optional skip expiration validation for update share

kiranparajuli589:
## Description:
- to expire a share explicitly, `expiration` field of a share should not be validated to be a future date
- optional skip feature to bypass expiration field validation

## Motivation
- https://github.com/owncloud/core/issues/36550

### DIFF
--- a/lib/public/Share/IManager.php
+++ b/lib/public/Share/IManager.php
@@ -53,12 +53,13 @@ interface IManager {
 	 * The share can't be removed this way (permission 0): use deleteShare
 	 *
 	 * @param IShare $share
+	 * @param Boolean $skipExpirationCheck
 	 * @return IShare The share object
 	 * @throws \InvalidArgumentException If $share is a link share or the $recipient does not match
 	 * @throws GenericShareException If $share requirements do not match
 	 * @since 9.0.0
 	 */
-	public function updateShare(IShare $share);
+	public function updateShare(IShare $share, $skipExpirationCheck = false);
 
 	/**
 	 * Delete a share
@@ -105,7 +106,7 @@ interface IManager {
 	 * @since 10.0.0
 	 */
 	public function getAllSharesBy($userId, $shareTypes, $nodeIDs, $reshares = false);
-	
+
 	/**
 	 * Get shares shared by (initiated) by the provided user.
 	 *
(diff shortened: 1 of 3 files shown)


phil-davis:
@micbar we want to be able to have a reasonable test that shares expire at the end of the expiry date. But it is rather difficult to wait until the next day to get the test result!

We propose to test by this style of steps:
1) create share that expires today (using the normal API for creating a share)
2) check that the share is working (Brian can download a file that Alice shared with him)
3) change the expiry of the share to yesterday
4) check that Brian cannot see the share any more
5) check that the share is gone (the share does not even exist for Alice)

The challenge is at step 3. The core code currently prevents specifying and expiry date in the past. We would like to provide an end-point in the testing app that allows setting an expiry date in the past, to be used by the test code at step (3). That testing app code should use the existing `updateShare` code to make the update, rather than directly accessing the database row, and rather than the testing app copying the `updateShare` logic.

This PR provides a way to call `updateShare` ignoring the expiration date validation. The production code will never call it like that. There are already tests like https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareCreateSpecialToRoot1/createShareExpirationDate.feature which confirm that the production API endpoint does validate the expiration date.

There is PR https://github.com/owncloud/testing/pull/173 for the testing which would use this core change.

What do you think?

lib/public/Share/IManager.php
```diff
@@ -53,12 +53,13 @@ public function createShare(IShare $share);
...
 	 *
 	 * @param IShare $share
+	 * @param Boolean $skipExpirationCheck
 	 * @return IShare The share object
 	 * @throws \InvalidArgumentException If $share is a link share or the $recipient does not match
 	 * @throws GenericShareException If $share requirements do not match
 	 * @since 9.0.0
 	 */
-	public function updateShare(IShare $share);
+	public function updateShare(IShare $share, $skipExpirationCheck = false);
```
phil-davis, line 62:
```suggestion
	public function updateShare(IShare $share, $skipExpirationValidation = false);
```

phil-davis (approved):
This works and the production behavior is the same. Merging so that this can get into the daily-master-qa tarball tonight.

PR #38709, opened 2021-05-07
Adjust permissions for the root folder of the SMB mount

jvillafanez:
## Description
Fix permissions for the root folder of the mount point

## Related Issue
https://github.com/owncloud/enterprise/issues/4546

## How Has This Been Tested?
Manually tested. Renaming or deleting the mount point will get a 403 error code

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38794, opened 2021-05-31
Skip initial permission check on setName method for shared mounts

### LINKED ISSUE #30325: Rename option is visible, even though user only has share permissions
### Steps to reproduce

1. Share a folder (or file) with a group with only "can share" permissions set
2. Log in as the share receiver
3. Click the ellipsis next to the file or folder and the `Rename` option is visible
4. Change the name in the popup text field and click enter
5. The operation will fail, the name will revert, and a UI failure notice will appear

rename issue

### Expected behaviour

If a user doesn't have edit permissions, and if attempting to rename a file or folder results in the action not succeeding and the UI message `Could not rename "sample_folder"` being rendered, then it doesn't make sense to give the user the ability to attempt to rename the item in the first place.

### Actual behaviour

If a user doesn't have edit permissions to a file or folder, hide the Rename option in the UI.

### Server configuration

**Operating system**: Ubuntu 16.04

**Web server:** Apache 2

**Database:** MariaDB

**PHP version:** 5.63

**ownCloud version:** 10.0.4 (stable)

**Updated from an older ownCloud or fresh install:** fresh install

**Where did you install ownCloud from:** git clone (a06cbff)

JammingBen:
## Description
Because renaming a mount should always be possible (see https://github.com/owncloud/enterprise/issues/4582#issuecomment-851328502).

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4582
- fixes https://github.com/owncloud/core/issues/30325

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


JammingBen:
It seems that [the check](https://github.com/owncloud/core/pull/38794/files#diff-7ed74baefb94f2f93809c92d12948f3b4f276dafca8a57c621def5668f670e9aL127) can be removed safely. This makes it possible to always rename a share regardless of its permissions. Not a 100% sure though, any opinions or thoughts @mrow4a ?

Another thing to mention: we still have the same issue with mounted storages. The UI shows the "rename"-action for a storage, but renaming results in an error. @jvillafanez You've dealt with a similar issue lately if I remember correctly? Something related to WND?

jvillafanez:
Yes. https://github.com/owncloud/core/pull/38709 is just for SMB. WND has similar changes too. I don't know if we want to make similar changes for all the storages.

jvillafanez:
Could we check instead if the node refers to a shared mount? Something like "if node is not a shared mount and not writable, then throw forbidden exception" It seems more reasonable than just removing the permission check.

JammingBen:
@jvillafanez 

> I don't know if we want to make similar changes for all the storages.

Depends on the solution I would say. According to Patrick, a mount point should always be allowed to be renamed. Your fix is heading in a different direction if I understood correctly, I assume because it was an easy fix for the problem? So that would be something for the future.

> Could we check instead if the node refers to a shared mount? Something like "if node is not a shared mount and not writable, then throw forbidden exception" It seems more reasonable than just removing the permission check.

Good point, I updated the PR.

@phil-davis How do we handle the acceptance test in https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementToRoot/moveReceivedShareOc10Issue30325.feature? Do we adjust it, or better remove entirely?

phil-davis:
> How do we handle the acceptance test in https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementToRoot/moveReceivedShareOc10Issue30325.feature? Do we adjust it, or better remove entirely?

Remove it, and search for `issue-30325` tags. Remove the `skipOnOcV10` tags from the "good" scenarios. Those should now pass.

The same sort of thing is here: https://github.com/owncloud/core/blob/master/tests/acceptance/features/webUIRenameFolders/renameFoldersOc10Issue30325.feature

tests/acceptance/features/apiShareManagementToRoot/moveReceivedShare.feature
```diff
@@ -74,7 +74,6 @@ Feature: sharing
     And as "Alice" file "/folderToShare/renamedFile" should exist
     But as "Alice" file "/folderToShare/fileInside" should not exist
 
-  @issue-30325 @skipOnOcV10
```
phil-davis, line 77: If you need to make more changes anyway, then please leave an empty line here. Nowadays we are trying to leave 2 lines between scenarios, so that scenario tags can come and go without all the line numbers of scenarios changing. That minimizes changes to expected-failures files in ocis and reva :)

tests/acceptance/features/webUIRenameFolders/renameFolders.feature
```diff
@@ -121,7 +121,6 @@ Feature: rename folders
     When the user renames folder "a-folder" to "a.part" using the webUI
     Then near folder "a-folder" a tooltip with the text '"a.part" has a forbidden file type/extension.' should be displayed on the webUI
 
-  @issue-30325 @skipOnOcV10
```
phil-davis, line 124: If you need to make more changes anyway, then please leave an empty line here.

changelog/unreleased/38794
```diff
@@ -0,0 +1,7 @@
+Bugfix: Always allow renaming shared mounts
+
+We now skip the initial permission check when renaming a shared mount as
+this should always be possible.
+
+https://github.com/owncloud/core/pull/38794
+https://github.com/owncloud/enterprise/issues/4582
```
phil-davis, line 7: Should we also link to issue #30325 - that is fixed by this PR

phil-davis (commented):
Code and test changes LGTM

apps/dav/lib/Connector/Sabre/Node.php
```diff
@@ -123,9 +124,8 @@ public function getPath() {
 	 * @throws InvalidPath
 	 */
 	public function setName($name) {
-
-		// rename is only allowed if the update privilege is granted
-		if (!$this->info->isUpdateable()) {
+		$mountPoint = $this->info->getMountPoint();
+		if (!($mountPoint instanceof SharedMount) && !$this->info->isUpdateable()) {
```
jvillafanez, line 128: I'd prefer to have a comment to explain the exception we're doing for the SharedMount

jvillafanez (approved):
Little detail. Code looks good anyway.

phil-davis (approved):

PR #38795, opened 2021-05-31
Improve external storages settings ui: Show dedicated messages for failed to load mount points / Autofocus available for while adding external mount points 

AlexAndBear:
## Description
Enhancement: Improve admin external storage settings UI

Before this PR no error notification was shown while an external mount point configuration was not able to load.
This was only indicated with a red square with a long list of external mount points, this was not handy.
Therefore an error notification will be shown with the dedicated external mount point which fails to load.

As well improved the add external mount point functionality with opening the available for select while adding an external mount point.
The select will have now a 'select all' item.
These changes prevents the admin to expose the mount point
unwanted to all users immediately.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4585

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/js/settings.js
+++ b/apps/files_external/js/settings.js
@@ -191,6 +191,19 @@ function addSelect2 ($elements, userListLimit) {
 			}
 		});
 	});
+
+	var $selectAll = $elements.closest('tr').find('.select2-drop').
+	prepend(
+		'<ul class="select2-controls">' +
+		'	<li class="selectAllApplicableUsers">' +
+		'		<div class="avatardiv"><img width="32" height="32" src="'+OC.imagePath('core','places/contacts-dark')+'"></div>' +
+		'		<span>'+t('files_external', 'All users')+'</span>' +
+		'	</li>' +
+		'</ul>');
+
+	$selectAll.on('click', function (){
+		$($elements).val(null).trigger('change');
+	});
 }
 
 /**
@@ -769,7 +782,7 @@ MountConfigListView.prototype = _.extend({
 		$tr = this.newStorage(storageConfig, onCompletion);
 		onCompletion.resolve();
 
-		$tr.find('td.configuration').children().not('[type=hidden]').first().focus();
+		$tr.find('td.applicable').children().first().select2('open');
 		this.saveStorageConfig($tr);
 	},
 
@@ -1271,6 +1284,7 @@ MountConfigListView.prototype = _.extend({
 	recheckStorageConfig: function($tr) {
 		var self = this;
 		var storage = this.getStorageConfig($tr);
+
 		if (!storage.validate()) {
 			return false;
 		}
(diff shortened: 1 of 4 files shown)


jvillafanez:
Double-check with multiple mount points. I think that the initial page load triggers an update status for all the mounts, so if there are several wrong mount points the notifications could be piled up.

JanAckermann:
@jvillafanez 
> Double-check with multiple mount points. I think that the initial page load triggers an update status for all the mounts, so if there are several wrong mount points the notifications could be piled up.

That's intended, customer have about 70 mount points with a large list and wants to be informed about all failing mount points without scrolling trough the list

jvillafanez:
My worry is that, if there is a connection problem and all the storages are disconnected, the admin will have 70 notifications in the web UI, and I'm not sure if the UI is ready to handle such amount. It could be fine for 3-5 errors but I think the UI will be bloated if there are more.
I don't remember a scroll in the notification area, so I suspect the area will keep expanding.

JanAckermann:
@jvillafanez 
Don't worry. 

I tried it, and it looks good to me:
image

apps/files_external/js/settings.js
```diff
@@ -1279,13 +1281,39 @@ MountConfigListView.prototype = _.extend({
...
 		});
 	},
 
+	/**
+	 * Recheck storage availability
+	 *
+	 * @param {jQuery} $tr storage row
+	 * @param {number} status storage status
+	 */
+	onRecheckStorageConfig: function($tr, status){
```
jvillafanez, line 1300: I think the "on" prefix is usually used when the function is directly bound to an event. In this case, it's the anonymous function the one bound, and this function is called by that anonymous function.
Changing the filename to "recheckStorageConfig" is enough to me.

JanAckermann: Nope the function caller is recheckStorageConfig and onRecheckStorageConfig is a  callback,
so we can call it recheckStorageConfigCallBack aswell. 
I don't have strong feelings about it, its on you.

.htaccess
```diff
@@ -94,3 +94,35 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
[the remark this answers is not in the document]
JanAckermann: whoops

apps/files_external/js/settings.js
jvillafanez: Ok, there is already a "recheckStorageConfig" method... leave it as it is.

IljaN (approved):

phil-davis (commented):
Test changes LGTM and pass.

PR #38804, opened 2021-06-03
New cache state

jvillafanez:
## Description
Add a new state in the cache to know the folders that have been shallow scanned to avoid re-scanning them

## Related Issue
https://github.com/owncloud/enterprise/issues/4594

## Motivation and Context
A partial scan implies that a folder might have been scanned on its first level but not deeper. Except for the size, the rest of the information can be reused and there is no need to scan the folder again (if there is no change detected).
This PR improves this scenario and prevents hitting the backend storage if there is no change

## How Has This Been Tested?
Checked with an slow external storage
1. Create folders "/ext/A/B/C/D" in the backend
2. Ensure "/ext/A" contains plenty of files and folders
3. Enter in "/ext/A". A scan of the backend should happen
4. Reload the page

Page will load faster because we won't hit the external storage.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -38,6 +38,7 @@ namespace OC\Files\Cache;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\Cache\ICache;
+use OCP\Files\Cache\IScanner;
 use OCP\Files\Cache\ICacheEntry;
 use \OCP\Files\IMimeTypeLoader;
 use OCP\IDBConnection;
@@ -681,10 +682,11 @@ class Cache implements ICache {
 	 * - Cache::PARTIAL: File is not stored in the cache but some incomplete data is known
 	 * - Cache::SHALLOW: The folder and it's direct children are in the cache but not all sub folders are fully scanned
 	 * - Cache::COMPLETE: The file or folder, with all it's children) are fully scanned
+	 * - Cache::NOT_SCANNED: Only the folder is in the cache. The contents are unknown, so the folder needs a scan
 	 *
 	 * @param string $file
 	 *
-	 * @return int Cache::NOT_FOUND, Cache::PARTIAL, Cache::SHALLOW or Cache::COMPLETE
+	 * @return int Cache::NOT_FOUND, Cache::PARTIAL, Cache::SHALLOW, Cache::COMPLETE or Cache::NOT_SCANNED
 	 */
 	public function getStatus($file) {
 		// normalize file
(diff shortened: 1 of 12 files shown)


lib/private/Files/Cache/Cache.php
```diff
@@ -885,8 +898,8 @@ public function getAll() {
 	 */
 	public function getIncomplete() {
 		$query = $this->connection->prepare('SELECT `path` FROM `*PREFIX*filecache`'
-			. ' WHERE `storage` = ? AND `size` = -1 ORDER BY `fileid` DESC', 1);
-		$query->execute([$this->getNumericStorageId()]);
+			. ' WHERE `storage` = ? AND `size` = ? ORDER BY `fileid` DESC', 1);
+		$query->execute([$this->getNumericStorageId(), IScanner::SIZE_NEEDS_SCAN]);
```
jvillafanez, line 902: Do we need to check for the new "-2" state? I don't think so.

lib/private/Files/View.php
```diff
@@ -1412,6 +1413,8 @@ private function getCacheEntry($storage, $internalPath, $relativePath) {
 				$storage->getPropagator()->propagateChange($internalPath, \time());
 				$data = $cache->get($internalPath);
 				$this->unlockFile($relativePath, ILockingProvider::LOCK_SHARED);
+			} elseif ($data['size'] === IScanner::SIZE_SHALLOW_SCANNED) {
+				$cache->correctFolderSize($internalPath, $data);
```
jvillafanez, line 1417: Lock shouldn't be needed because we don't hit the storage... should we propagate the change using the propagator?

lib/private/Files/Cache/Cache.php
JammingBen: I believe so, too. Including the -2 here would result in those folders being scanned in `backgroundScan()`, which you want to avoid, right?

The method name and docs are kind of misleading now though. We should at least clarify the docs IMO. `find a folder in the cache which has not been fully scanned` -> this is not true anymore as shallow scanned folders are also not fully scanned either.

JammingBen (commented):
Added input regarding the `getIncomplete()` method. Other than that it looks good to me, but changelog is missing.

lib/private/Files/Cache/Cache.php
jvillafanez: The method is used by the `occ files:scan --unscanned` command. The command relies on finding unscanned folders, which works better with the separation of states.

JammingBen: I agree, but don't you think that the docs are kinda misleading now? When reading `find a folder in the cache which has not been fully scanned` I would assume it searches for `SIZE_NEEDS_SCAN` or `SIZE_SHALLOW_SCANNED` because in both cases the folder has been not fully scanned.

jvillafanez:
This is ready.
The only open question is https://github.com/owncloud/core/pull/38804#discussion_r644805533 but I don't think it's critical. Size should be properly update in the DB in every directory, and client should perform a propfind and update the size of the requested folders. No need to force a change in the etag to force change detection just for the size of a folder.

JammingBen (approved):

ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUIAddUsers-chrome-mariadb10.2-php7.4</strong> failed. The build is cancelled...</summary>

phil-davis:
https://drone.owncloud.com/owncloud/core/30591/125/15
Fatal error: Call to undefined method GuzzleHttp\Exception\ConnectException::getResponse() (Behat\Testwork\Call\Exception\FatalThrowableError)
Some problem with mailhog - I will restart CI.

PR #38806, opened 2021-06-04
Do not show change permission as duplicate for can edit in case of files

mrow4a:
fixes https://github.com/owncloud/enterprise/issues/4587

Background:

The can edit checkbox is in fact role. For files it is filesEditor, for folder it is folderEditor. The delete, create, change are like permissions belonging to corresponding roles.

For filesEditor allowed values are change.
For folderEditor allowed are delete, create, change.

- [ ] instead of checking file/folder do it more generic and check if current permission checkbox is the only one available out of all coming from "can edit" role.


micbar:
Will not be developed further.

PR #38813, opened 2021-06-07
Report 23:59:59 in share expiration

phil-davis:
## Description
Share responses when creating a share, modifying or getting share information report the expiration date like:
`<expiration>2021-06-22 00:00:00</expiration>`

They always report expiration time `00:00:00`. But we have a convention in the back-end code that all shares work until the end of the expiration day, and back-end code that checks expiration dates handles that logic - so actually that share works until `2021-06-22 23:59:59.9999999999...`

This PR changes the reported time of expiration to `23:59:59` - that might be less confusing for clients that care about looking at the expiration time.

ToDo:
- [ ] add a changelog after we decide exactly what to do here

## Related Issue
- https://github.com/owncloud/enterprise/issues/4454

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/3680


phil-davis:
@micbar what is the requirement for external behavior for reporting the expiration date-time of a share?

phil-davis:
@TheOneRing one of the questions is: do the clients care if the expiration of a share has a date and then time of `23:59:00` rather than `00:00:00`?

And the other question could be: What benefit might there be to the expiration in a share response having time `23:59:00`?

phil-davis:
Yes, IMO all the current clients only set/display share expiration down to the day. They have no display of the expiration time within a day. So, if a PR like this goes ahead, we just need to double-check in QA that each client (desktop, iOS, Android) copes OK  with the expiration time set to `23:59:00` (i.e. no observable difference to now). The traditional webUI is fine - CI passes here.

micbar:
I did another deep thinking here.

IMO it is totally fine if the api doesn't report the time, only the day.

But in the best practise for API stability, changing this is not possible without a new API version.

It is bad, that `<expiration>2021-06-22 00:00:00</expiration>` implicitly means `<expiration>2021-06-23 00:00:00</expiration>`

If we are strict here, we should not change it regardless of the fact that it is just wrong.

micbar:
@michaelstingl @dragotin @butonic This is so "broken"

The bad thing about this is, we need to keep the APIs for oCIS "bug" compatible.

phil-davis:
> IMO it is totally fine if the api doesn't report the time, only the day.

PR #38817 demonstrates not reporting the time at all. Nothing breaks in the webUI, API tests pass with the expected difference. It would be easy enough to check that the other clients behave OK when the time is left out. So that is a possible thing to do.

And it is possible to just do nothing :)

michaelstingl:
> It would be easy enough to check that the other clients behave OK when the time is left out.

@jesmrec @gabi18 can we validate this in the next desktop and mobile releases?

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1015,14 +1015,21 @@ public function isFieldInResponse($field, $contentExpected, $expectSuccess = tru
 		$dateFieldsArrayToConvert = ['expiration', 'original_date', 'new_date'];
 		//do not try to convert empty date
 		if ((string) \in_array($field, \array_merge($dateFieldsArrayToConvert)) && !empty($contentExpected)) {
+			if ($field === 'expiration') {
+				// shares expire at the end of the expiry day
+				// responses indicate this by including a time of 23:59:00
+				$timeOffset = "23:59:00";
```
JanAckermann, line 1021: Maybe we call it time instead of timeOffset ?

AlexAndBear (approved):

tests/acceptance/features/bootstrap/Sharing.php
JanAckermann: and maybe we should make use of 23:59:59 which represents EOD a bit more

phil-davis: Agree. 23:59:59 is as close as we can go to end-of-day.
First, someone needs to decide if we do this at all, or we remove the time completely - demonstrated in #38817

JanAckermann: Ok, I'll move it to needs info then, so it will get recognized again ;)

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -222,7 +222,7 @@ protected function formatShare(IShare $share, $received = false) {
 
 		$expiration = $share->getExpirationDate();
 		if ($expiration !== null) {
-			$result['expiration'] = $expiration->format('Y-m-d 00:00:00');
+			$result['expiration'] = $expiration->format('Y-m-d 23:59:00');
```
JanAckermann, line 225:
```suggestion
			$result['expiration'] = $expiration->format('Y-m-d 23:59:59');
```

apps/files_sharing/tests/ApiTest.php
```diff
@@ -1526,7 +1526,7 @@ public function testCreatePublicLinkExpireDateValid() {
 
 		$data = $result->getData();
 		$this->assertIsString($data['token']);
-		$this->assertEquals($date->format('Y-m-d') . ' 00:00:00', $data['expiration']);
+		$this->assertEquals($date->format('Y-m-d') . ' 23:59:00', $data['expiration']);
```
JanAckermann, line 1529:
```suggestion
		$this->assertEquals($date->format('Y-m-d') . ' 23:59:59', $data['expiration']);
```

apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
```diff
@@ -505,7 +505,7 @@ public function dataGetShare() {
 			'file_target' => 'target',
 			'file_parent' => 3,
 			'token' => 'token',
-			'expiration' => '2000-01-02 00:00:00',
+			'expiration' => '2000-01-02 23:59:00',
```
JanAckermann, line 508:
```suggestion
			'expiration' => '2000-01-02 23:59:59',
```

apps/files_sharing/tests/Controller/Share20OcsControllerTest.php
```diff
@@ -2564,7 +2564,7 @@ public function dataFormatShare() {
 				'attributes' => null,
 				'stime' => 946684862,
 				'parent' => null,
-				'expiration' => '2001-01-02 00:00:00',
+				'expiration' => '2001-01-02 23:59:00',
```
JanAckermann, line 2567:
```suggestion
				'expiration' => '2001-01-02 23:59:59',
```

tests/acceptance/features/bootstrap/Sharing.php
JanAckermann, line 1021:
```suggestion
				$timeOffset = "23:59:59";
```

micbar (approved):

AlexAndBear (changes requested):

JanAckermann:
Changelog needs to be provided, as it says in the todos :)

tests/acceptance/features/bootstrap/Sharing.php
phil-davis: expiration time is now reported as 23:59:59

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1015,14 +1015,21 @@ public function isFieldInResponse($field, $contentExpected, $expectSuccess = tru
 		$dateFieldsArrayToConvert = ['expiration', 'original_date', 'new_date'];
 		//do not try to convert empty date
 		if ((string) \in_array($field, \array_merge($dateFieldsArrayToConvert)) && !empty($contentExpected)) {
+			if ($field === 'expiration') {
+				// shares expire at the end of the expiry day
+				// responses indicate this by including a time of 23:59:00
```
phil-davis, line 1020:
```suggestion
				// responses indicate this by including a time of 23:59:59
```

AlexAndBear (approved):
LGTM, changelog missing, then :rocket:

phil-davis:
Documentation updated in PR https://github.com/owncloud/docs/pull/3681

PR #38819, opened 2021-06-08
Check a local attribute for the SMB log switch instead

jvillafanez:
## Description
Use a local attribute to decide whether to log something in the SMB connector or not
It also improves the log handling in the subclasses because the inherited methods will use the same flag

## Related Issue
Related to https://github.com/owncloud/enterprise/issues/4594

## Motivation and Context
Just deciding to log nothing in the connector takes some time. It might be a few microseconds, but the sheer amount of calls that big installations can make might rise this number a lot.
For a particular setup with a folder with 1000 files and another folder inside, there are 73240 calls to the log method. Although each call could take around 10-15 microseconds per call on average, the total amount takes close to 1 second in order to log nothing.

This PR make the logging calls to take nanoseconds if the logging is disabled. For the same amount of data, the waste should be around 10 miliseconds instead of 1 second.

Side note: the request takes around 35 seconds, so it isn't a big gain.

## How Has This Been Tested?
Manually tested. Checked with a propfind against the folder with 1000 files using curl, as well as navigating to the folder using the browser. In both cases, a file scan was forced to be triggered with the propfind.

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -54,6 +54,9 @@ use OCP\Files\StorageNotAvailableException;
 use OCP\Util;
 
 class SMB extends StorageAdapter {
+	/** @var bool */
+	protected $logActive;
+
 	/**
 	 * @var IServer
 	 */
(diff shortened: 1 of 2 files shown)


AlexAndBear (approved):
Sweet, can you also provide a changelog item? 
Small (before, after) benchmark numbers would be put the ice on the cake according to the changelog item.

jvillafanez:
Before patch
Screenshot from 2021-06-10 09-34-24

After patch
Screenshot from 2021-06-10 09-35-43

That's the data I have. So I guess, without the patch, the method takes 0.63% of the time, and with the patch it takes 0.07%.
I don't really want to put numbers because there are always additional factors to take into account, such as network traffic, server load, etc, that could affect the result.

The data refers to the WND app, but the SMB app should have similar results.

Changelog items incoming.

PR #38697, opened 2021-05-05
Drop support for PHP 7.2

phil-davis:
## Description
Now that `doctrine/dbal` has been successfully bumped, try dropping PHP 7.2 support and see what else fails.

composer found the following that it could update:
```
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Package operations: 0 installs, 7 updates, 0 removals
  - Upgrading laminas/laminas-zendframework-bridge (1.1.1 => 1.3.0): Extracting archive
  - Upgrading laminas/laminas-stdlib (3.2.1 => 3.5.0): Extracting archive
  - Upgrading laminas/laminas-validator (2.13.5 => 2.14.5): Extracting archive
  - Upgrading laminas/laminas-servicemanager (3.5.2 => 3.7.0): Extracting archive
  - Upgrading laminas/laminas-filter (2.9.4 => 2.11.1): Extracting archive
  - Upgrading laminas/laminas-inputfilter (2.10.1 => 2.12.0): Extracting archive
  - Upgrading phpunit/php-token-stream (3.1.3 => 4.0.4): Extracting archive
```

## Related Issue
Fixes #39134 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation PR https://github.com/owncloud/docs/pull/3992
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38822, opened 2021-06-09
Update to doctrine DBAL v3

JammingBen:
## Description
This is needed to run PostgreSQL with a persistent connection.

## Related Issue

- Needed for https://github.com/owncloud/enterprise/issues/4613

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -25,12 +25,11 @@ config = {
 	'phpunit': {
 		'allDatabases' : {
 			'phpVersions': [
-				'7.2',
+				'7.3',
 			]
 		},
 		'reducedDatabases' : {
 			'phpVersions': [
-				'7.3',
 				'7.4',
 			],
 			'databases': [
@@ -40,7 +39,7 @@ config = {
 		},
 		'external-samba-windows' : {
 			'phpVersions': [
-				'7.2',
+				'7.3',
 				'7.4',
 			],
 			'databases': [
@@ -60,7 +59,7 @@ config = {
 		},
 		'external-other' : {
 			'phpVersions': [
-				'7.2',
+				'7.3',
 				'7.4',
 			],
 			'databases': [
@@ -356,7 +355,7 @@ def dependencies(ctx):
 		return pipelines
 
 	default = {
-		'phpVersions': ['7.2'],
+		'phpVersions': ['7.3'],
 	}
 
 	if 'defaults' in config:
@@ -679,7 +678,7 @@ def phan():
 		return pipelines
 
 	default = {
-		'phpVersions': ['7.2', '7.3', '7.4'],
+		'phpVersions': ['7.3', '7.4'],
 		'logLevel': '2',
 	}
 
@@ -753,7 +752,7 @@ def litmus():
 		return pipelines
 
 	default = {
-		'phpVersions': ['7.2', '7.3', '7.4'],
+		'phpVersions': ['7.3', '7.4'],
 		'logLevel': '2',
 		'useHttps': True,
 	}
@@ -915,7 +914,7 @@ def dav():
 		return pipelines
 
 	default = {
-		'phpVersions': ['7.2', '7.3', '7.4'],
+		'phpVersions': ['7.3', '7.4'],
 		'logLevel': '2'
 	}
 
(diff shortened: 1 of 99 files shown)


JammingBen:
@micbar As discussed I've opened this draft PR for the doctrine DBAL update. As expected this is a *huge* one.

This PR includes some minor changes already. But we basically need to adjust every database interaction because methods like `fetch()` or `closeCursor()` are not available anymore. Maybe we could introduce some kind of compatibility layer by implementing own DB classes which "translate" our old methods to new ones. I'll check 👍

phil-davis:
Link to upgrade notes: https://github.com/doctrine/dbal/blob/3.1.x/UPGRADE.md
https://github.com/doctrine/dbal/blob/3.1.x/composer.json requires PHP minimum of 7.3, which I see is done in this PR.
PR #38697 has the minimal core changes needed to drop PHP 7.2 (without increasing major versions of dependencies) - look there for "interesting stuff"  that you will need to do for that.

The dropping PHP 7.2 has been advised in release notes of at least 10.6 and 10.7. So it is not a surprise, when we do it. But if it is going to happen for 10.8, then someone needs to decide that ASAP.

JammingBen:
Okay, so I managed to get OC running with the update. The main things that changed for us:

* Minimum PHP version is 7.3 (was 7.2 previously)
* Type constants are now located in `Doctrine\DBAL\Types\Types` (was `Doctrine\DBAL\Types\Type` previously)
* `closeCursor()` is now `free()`
* `\Doctrine\DBAL\DBALException` is now `\Doctrine\DBAL\Exception`
* The `Statement` interface no longer extends `ResultStatement`. That means we cannot fetch from a statement directly like we often do.
* ...and some more "minor" changes

Note that this PR does not provide options for backwards-compatibility. That means while core might run fine, most (or all) of our apps probably wont. I guess we could go for some implementations to ensure backwards-compatibility, but we have to dig deeper.

To the main reason for going on this adventure: I tried running with PostgreSQL and a persistent connection, it seems(!) to run. Meaning, I don't get any errors like before. But I cannot tell for sure if the connection is now persistent 🤷  

@micbar what do you suggest?

micbar:
>     * Minimum PHP version is 7.3 (was 7.2 previously)
Not a problem, we wanted to remove php7.2 support anyway. 
>     * Type constants are now located in `Doctrine\DBAL\Types\Types` (was `Doctrine\DBAL\Types\Type` previously)
will be used across a lot of apps
>     * `closeCursor()` is now `free()`
same
>     * `\Doctrine\DBAL\DBALException` is now `\Doctrine\DBAL\Exception`
same
>     * The `Statement` interface no longer extends `ResultStatement`. That means we cannot fetch from a statement directly like we often do.
same
>     * ...and some more "minor" changes
same
> 
> Note that this PR does not provide options for backwards-compatibility. That means while core might run fine, most (or all) of our apps probably wont. I guess we could go for some implementations to ensure backwards-compatibility, but we have to dig deeper.

Main Effort: identify all changes in all apps that are needed. Would need a big release effort.
> 
> To the main reason for going on this adventure: I tried running with PostgreSQL and a persistent connection, it seems(!) to run. Meaning, I don't get any errors like before. But I cannot tell for sure if the connection is now persistent 🤷

That is good news, but we would need to be sure. we need support help and maybe benchmarking.
> @micbar what do you suggest?

1) Use this branch as a experiment to understand really if that solves our DBConnection requirements.
2) if yes: Estimate the changes in the apps.

JammingBen:
One thing I can already tell regarding performance: Getting the connection is noticeably faster in persistent mode while using a SSL connection.

A simple time test on my local setup:

* Without persistent: 0.04s
* With persistent: 0.002

https://github.com/owncloud/core/blob/eebd754771a99bc44820eb67a6d57cea238e6b86/lib/private/DB/ConnectionFactory.php#L137

JammingBen:
Closing for now as we won't do this anytime soon. We can reopen one day if necessary.

PR #38824, opened 2021-06-09
Update coding-standard to major version 3 

phil-davis:
## Description
1) Update coding-standard to major version 3. That gets php-cs-fixer major version 3 also.

2) Apply code changes suggested by php-cs-fixer v3. IMO it now applies PSR12 by default, and other breaking changes to code style stuff.

I have not added a changelog, because this is a code-style-only change. There is not observable behavior change at any level (unit test, acceptance test, or production). Do we write a changelog anyway for this sort of thing?

ToDo: sort out if we want to suppress any of these new "standard" rules, to get closer to the previous rules applied by v2.

## Related Issue
- part of https://github.com/owncloud/coding-standard/issues/22

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


apps/comments/lib/Activity/Extension.php
```diff
@@ -302,11 +302,11 @@ public function getQueryForFilter($filter) {
 	 * @return string
 	 */
 	protected function convertParameterToComment($parameter) {
-		if (\preg_match('/^\<parameter\>(\d*)\<\/parameter\>$/', $parameter, $matches)) {
+		if (preg_match('/^\<parameter\>(\d*)\<\/parameter\>$/', $parameter, $matches)) {
```
phil-davis, line 305: It removes the `\` from some function calls because of how https://cs.symfony.com/doc/rules/function_notation/native_function_invocation.html behaves now.

Not all functions benefit from having the `\` in front of them - see discussion around https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048#issuecomment-328810072

We can stop these from being forcibly removed by changing coding-standard rules to:
```
'native_function_invocation' => ['strict' => false],
```

apps/comments/lib/Activity/Listener.php
```diff
@@ -52,11 +52,13 @@ class Listener {
 	 * @param IMountProviderCollection $mountCollection
 	 * @param IRootFolder $rootFolder
 	 */
-	public function __construct(IManager $activityManager,
-								IUserSession $session,
-								IAppManager $appManager,
-								IMountProviderCollection $mountCollection,
-								IRootFolder $rootFolder) {
+	public function __construct(
+		IManager $activityManager,
```
phil-davis, line 56: Now it puts the parameters all on separate lines. The change here looks like "a good thing"

apps/comments/lib/Activity/Extension.php
```diff
@@ -36,10 +36,10 @@
  * @package OCA\Comments\Activity
  */
 class Extension implements IExtension {
-	const APP_NAME = 'comments';
+	public const APP_NAME = 'comments';
```
phil-davis, line 39: class constants are public by default. This change makes that explicit. This seems to be "a good thing".

We could look through to find class constants that are only ever referenced inside the class, and make them private?

phil-davis:
@micbar @JammingBen @JanAckermann @jvillafanez and anyone else.

The new major version of php-cs-fixer checks code style using PSR12 as the default "standard". 

See the 3 things that I commented on above. They seem "reasonable", and all CI passes, so these code-style changes are not breaking anything. But there are >10k lines of diff.

Opinions please.

apps/comments/lib/Activity/Extension.php
jvillafanez: I'd keep the leading `\` for now. Even if not all functions benefit from having the leading `\`, we're talking about micro-optimizations which usually have extremely little impact in the code.
I'm not sure if we can somehow optionally check this change for specific parts of the code where we might be interested in maximum performance, but in general I'd skip this.

jvillafanez:
I'm not sure if we can apply the changes one by one in different PR. This way, reviewing the changes will be easier. We might also take advantage that sonarcloud seems to be yelling to include some minor improvements around.
I'd expect that by skipping the first change (about the leading `\`) we can reduce the number of changes drastically.

Anyway, I'd vote to include these changes after 10.8.

apps/comments/lib/Activity/Extension.php
phil-davis: I removed this change - PR #38828 - that makes less lines of diff.

phil-davis:
Closing in favor of #38828

PR #38821, opened 2021-06-09
Improve login screen

phil-davis:
## Description
This is a squashed and rebased version of PR #38506 

It might be easy to merge the code from here, rather than the other PR, which has 50 commits.

I left the changelog the same, so that it links to the original PR. That has all the discussion, manual test results, etc.

## How Has This Been Tested?
CI, and see posts in #38506 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38833, opened 2021-06-11
[tests-only] Break CI

phil-davis:
This purposely breaks a test in webUILogin suite to see if the earlyFail behavior is working correctly. - that worked, there is a single comment posted below.

The 2nd commit also breaks a scenario in webUIComments - we got just a post about webUIComments. webUILogin was cancelled before it got to post a fail - that is what should happen.

So this seems to work as it should. I don't know what happen in PR #38821 when it got spammed with test failure comments.

### DIFF
--- a/tests/acceptance/features/webUIComments/comments.feature
+++ b/tests/acceptance/features/webUIComments/comments.feature
@@ -19,7 +19,7 @@ Feature: Add, delete and edit comments in files and folders
     And the user comments with content "<comment>" using the webUI
     Then the comment "<comment>" should be listed in the comments tab in details dialog
     When the user deletes the comment "<comment>" using the webUI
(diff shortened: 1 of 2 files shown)


ownclouders:
<details><summary>:boom: Acceptance tests <strong>webUILogin-chrome-mariadb10.2-php7.4</strong> failed. The build is cancelled...</summary>

phil-davis:
@JanAckermann FYI the earlyFail works here as I expect it to work.

I don't know what happened to cause the comment spam in PR #38821

JanAckermann:
@phil-davis I squashed all commits, Florian Made a change and pushed, which was a merge than. All commits popped up again, maybe that's why ?

PR #38850, opened 2021-06-15
ci: replace unmaintained alpine step image

xoxys:
CI only

Would be good to back port to 10.7 and 10.6 as well.

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -543,7 +543,7 @@ def changelog(ctx):
 				},
 				{
 					'name': 'diff',
-					'image': 'owncloud/alpine:latest',
+					'image': 'owncloudci/alpine:latest',
 					'pull': 'always',
 					'commands': [
 						'git diff',


micbar (approved):

phil-davis (approved):

phil-davis:
> @phil-davis can you take care of the backports?

@xoxys do you mean making this change to the tip of the `release-10.6.0` and `release-10.7.0` branches? (also=release tags)

I am not sure why we need to do that. IMO nobody needs to run CI from those old branches any more.

xoxys:
If we don't do code back ports to the release branches and the ci is not getting triggered :+1:

phil-davis:
We only "wake up" an old release branch if there is some really critical bugfix that we want to apply, and put out a patch release. For example, to make a `release-10.7.1` when `master`  has already moved on with lots of new stuff. In that case we might backport the critical PR and go through a patch release process. Now that 10.8 code freeze is "really close", there will not be any 10.7.1.

PR #38867, opened 2021-06-19
[tests-only] test issue 35974

phil-davis:
Related issue: #35974

PR #38868, opened 2021-06-19
[tests-only] Allow for no unit tests with coverage in CI

phil-davis:
## Description
If there are no unit tests with coverage, then the existing `.drone.star` still schedules a sonar-analysis pipeline, which fails. I noticed this in PR #38867 when doing cut-down CI to demonstrate an acceptance test issue.

The starlark logic to decide whether a sonar-analysis pipeline is needed or not was implemented in the activity app last week - https://github.com/owncloud/activity/pull/959 but was not applied to core.

This PR applies the same logic to core.

99.9% of the time this is not an issue in core - unit tests are nearly always run in CI. But we may as well have the logic in `.drone.star` here.

## How Has This Been Tested?
CI of #38867 

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -300,6 +300,13 @@ def main(ctx):
 
 	dependsOn(before, coverageTests)
 
+	nonCoverageTests = nonCoveragePipelines(ctx)
+	if (nonCoverageTests == False):
+		print('Errors detected in nonCoveragePipelines. Review messages above.')
+		return []
+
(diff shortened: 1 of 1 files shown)


dpakach (approved):
LGTM 👍🏼

Issue #35974, opened 2019-08-02
Restore file from trashbin to a different place overwriting a file


phil-davis:
### Steps to reproduce
```
  Scenario Outline: restoring a file to an already existing path overrides the file
    Given using <dav-path> DAV path
    And user "user0" has been created with default attributes and skeleton files
    And user "user0" has uploaded file with content "file to delete" to "/textfile0.txt"
    And user "user0" has uploaded file with content "PARENT textfile0 content" to "/PARENT/textfile0.txt"
    And user "user0" has deleted file "/textfile0.txt"
    When user "user0" restores the file with original path "/textfile0.txt" to "/PARENT/textfile0.txt" using the trashbin API
    Then the HTTP status code should be "204"
    # Sometimes "/PARENT/textfile0.txt" is found in the trashbin. Should it? Or not?
    # That seems to be what happens when the restore-overwrite happens properly,
    # The original /PARENT/textfile0.txt seems to be "deleted" and so goes to the trashbin
    #And as "user0" the file with original path "/PARENT/textfile0.txt" should not exist in trash
    And as "user0" file "/PARENT/textfile0.txt" should exist
    # sometimes the restore from trashbin does overwrite the existing file, but sometimes it does not. That is also surprising.
    # the current observed behavior is that if the original /PARENT/textfile0.txt ended up in the trashbin,
    # then the new /PARENT/textfile0.txt has the "file to delete" content.
    # otherwise /PARENT/textfile0.txt has its old content
    And the content of file "/PARENT/textfile0.txt" for user "user0" if the file is also in the trashbin should be "file to delete" otherwise "PARENT textfile0 content"
    #And the content of file "/PARENT/textfile0.txt" for user "user0" should be "file to delete"
    Examples:
      | dav-path |
      | old      |
      | new      |
```

### Expected behaviour
Maybe:
1) the restored file should just overwrite the existing file, and the existing file is gone (or is saved as an old version if `files_versions` is enabled), or;
2) The restored file has `(2)` on the end of its name, and the existing file stays, or;
3) ???

### Actual behaviour
Sometimes the restore from trashbin overwrites the existing file. When that happens, the existing file is found in the trashbin. I guess that the code-path deletes the existing file (triggering an event that causes it to be saved into the trashbin). And the restored file has the correctly-restored content.

But also sometimes the existing file is not found in the trashbin, and has not been overwritten (it still has the old content). But the API call to restore from trashbin still returns `204` success.

### Server configuration
[...]

PR #38869, opened 2021-06-19
[tests-only] Cleanup good scenario for trashbinRestore issue-35974

phil-davis:
## Description
This scenario is supposed to describe the desired good behavior when restoring a file to an already existing path. The existing file should be over-written with the content of the restored file.

There happens to be another side-effect on oC10 - the original file (sometimes) gets put into the trashbin. That side-effect has not been decided to be "required behavior". Maybe the original file should end up in the trashbin, maybe it should be available as an old version of the file, or maybe it should just be thrown away.

This PR removes the check "file should not exist in the trashbin". That makes the scenario suitable as a test for implementations, for example OCIS.

See more notes and discussion in #38867 and https://github.com/cs3org/reva/pull/1795

## Related Issue
- #35974

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiTrashbin/trashbinRestore.feature
+++ b/tests/acceptance/features/apiTrashbin/trashbinRestore.feature
@@ -101,15 +101,7 @@ Feature: Restore deleted files/folders
     And user "Alice" has deleted file <delete-path>
     When user "Alice" restores the file with original path <delete-path> to <upload-path> using the trashbin API
     Then the HTTP status code should be "204"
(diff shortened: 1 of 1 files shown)


phil-davis:
@dpakach it will be good to get this merged into core, and to get the core commit id updated in reva and ocis. Then they will have a better "good" scenario to work with as they sort out fixing trashbin issues.

dpakach (approved):

PR #38880, opened 2021-06-23
[full-ci] Trigger file scan after accepting a federated share

### LINKED ISSUE #970: [QA] Activity when receiving a federated share is missing in sidebar and `Activities by others`
### Steps to reproduce
1. Receive and accept a share from a federated OC system
2. Check the sidebar and you activity stream

### Expected behaviour
The activity should be listed in the file sidebar and in the activity stream in `Activities by others`

### Actual behaviour
The activity is missing in the file sidebar as well as in `Activities by others`

JammingBen:
## Description
This is necessary as we need the `fileId` to pass it to the `remoteshare.accepted` event. The activity app can then hook in to this event and update the activity by setting the correct `fileId` (see https://github.com/owncloud/activity/pull/972). The `shareId` has been added to the published activity which is used to retrieve the activity later when it is updated.

Also added a short translation for the `SUBJECT_REMOTE_SHARE_RECEIVED` activity.

## Related Issue

- Fixes https://github.com/owncloud/activity/issues/970

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/External/Manager.php
+++ b/apps/files_sharing/lib/External/Manager.php
@@ -208,13 +208,46 @@ class Manager {
 				WHERE `id` = ? AND `user` = ?');
 			$acceptShare->execute([1, $mountPoint, $hash, $id, $this->uid]);
 
+			$options = [
+				'remote'	=> $share['remote'],
+				'token'		=> $share['share_token'],
+				'mountpoint'	=> $mountPoint,
+				'owner'		=> $share['owner']
+			];
+
+			// We need to scan the new file/folder here to get its fileId
+			// which will be passed to the event for further processing.
+			$mount = $this->getMount($options);
+			$storage = $mount->getStorage();
+			$fileId = null;
+
+			if ($storage) {
+				$scanner = $storage->getScanner('');
+
+				// No need to scan all the folder contents as we only care about the root share
+				$file = $scanner->scanFile('');
+
+				if (isset($file['fileid'])) {
+					$fileId = $file['fileid'];
+				}
+			}
+
 			$this->eventDispatcher->dispatch(
 				AcceptShare::class,
 				new AcceptShare($share)
 			);
 
-			$event = new GenericEvent(null, ['sharedItem' => $share['name'], 'shareAcceptedFrom' => $share['owner'],
-				'remoteUrl' => $share['remote']]);
+			$event = new GenericEvent(
+				null,
+				[
+					'sharedItem' => $share['name'],
+					'shareAcceptedFrom' => $share['owner'],
+					'remoteUrl' => $share['remote'],
+					'fileId' => $fileId, // can be null in case the file was not scanned properly
+					'shareId' => $id,
+					'shareRecipient' => $this->uid,
+				]
+			);
 			$this->eventDispatcher->dispatch('remoteshare.accepted', $event);
 			\OC_Hook::emit('OCP\Share', 'federated_share_added', ['server' => $share['remote']]);
 
--- a/apps/federatedfilesharing/lib/FedShareManager.php
+++ b/apps/federatedfilesharing/lib/FedShareManager.php
@@ -159,7 +159,7 @@ class FedShareManager {
 		$this->publishActivity(
 			$shareWith,
 			Activity::SUBJECT_REMOTE_SHARE_RECEIVED,
-			[$ownerAddress->getCloudId(), \trim($name, '/')],
+			[$ownerAddress->getCloudId(), \trim($name, '/'), ['shareId' => $shareId]],
 			'files',
 			'',
 			'',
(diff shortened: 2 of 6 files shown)


ownclouders:
<details><summary>:boom: Acceptance tests <strong>cliExternalStorage-latest-mariadb10.2-php7.4</strong> failed. The build is cancelled...</summary>

apps/files_sharing/lib/External/Manager.php
```diff
@@ -208,15 +208,43 @@ public function acceptShare($id) {
...
+				'owner'		=> $share['owner']
+			];
+
+			$mount = $this->getMount($options);
+			$storage = $mount->getStorage();
+			$fileId = null;
+
+			if ($storage) {
+				$scanner = $storage->getScanner('');
+				$file = $scanner->scanFile('');
```
jvillafanez, line 224: We might need some comments here. I was confused because it seems you'll be scanning the same file regardless of what's being accepted.
I guess the storage will be different per accepted share, as well as the scanner, which would point to a different file depending on the share, so the "root" which will be scanned will be different.

In addition, it might be a good idea to include a comment explaining why you need a `scanFile` and not a `scan`. In case the share points to a folder, I think you'll only scan the folder itself without going deeper. I'm not sure if that's what we want or not. Note that a swallow scan (not just a simple `scanFile`) can be very expensive.

phil-davis:
This is a bit behind `master` (9 commits behind) - I noticed because the message about "Acceptance tests failed" is in an old format.

ownclouders:
<details><summary>:boom: Acceptance tests <strong>cliExternalStorage-latest-mariadb10.2-php7.4</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>cliExternalStorage-10.6.0-mariadb10.2-php7.4</strong> failed. The build is cancelled...</summary>

ownclouders:
<details><summary>:boom: Acceptance tests <strong>cliExternalStorage-git-mariadb10.2-php7.4</strong> failed. The build is cancelled...</summary>

JammingBen:
Rebased 👍

apps/files_sharing/lib/External/Manager.php
```diff
@@ -208,15 +208,43 @@ public function acceptShare($id) {
...
 
-			$event = new GenericEvent(null, ['sharedItem' => $share['name'], 'shareAcceptedFrom' => $share['owner'],
-				'remoteUrl' => $share['remote']]);
+			$event = new GenericEvent(
+				null,
+				[
+					'sharedItem' => $share['name'],
+					'shareAcceptedFrom' => $share['owner'],
+					'remoteUrl' => $share['remote'],
+					'fileId' => $fileId,
```
jvillafanez, line 242: the fileId could be null, so it's a good idea to include a comment to clarify it's intended or it's handled.

JanAckermann:
@phil-davis @jvillafanez 
After accepting a share we trigger the file scan for the home folder of the user, to get the file cache entry for the share, this is why the ETag is changing and results in failing tests, IHMO we should to the opposite and test if the ETag will change, agree?

jvillafanez:
I think it makes sense that we trigger an etag change because, from the user's perspective, it's a new file / folder that appears in the tree.
You can also check what happens when the `share_folder` configuration is present in the config.php file.

ownclouders:

https://drone.owncloud.com/owncloud/core/30778/124/1

phil-davis:
> @phil-davis @jvillafanez
> After accepting an external share we trigger the file scan for the home folder of the user, to get the file cache entry for the share, this is why the ETag is changing and results in failing tests, IHMO we should to the opposite and test if the ETag will change, agree?

Agree.

I edited the PR title to say `[full-ci]` - that will prevent the "fail early" CI behavior. So next time you trigger CI, every pipeline will run to completion and you get to see all the failing test scenarios.

phil-davis:
I restarted CI just now.

apps/files_sharing/lib/External/Manager.php
JammingBen: Added some comments, do you think it's more clear now?

jvillafanez:
".htaccess" and "Makefile" files shouldn't be touched.

JammingBen:
> I think it makes sense that we trigger an etag change because, from the user's perspective, it's a new file / folder that appears in the tree.

Actually this is not happening with the file scan. But it's not related to our change.

To test this, I shared a folder with a local user. After logging in with that user the new folder was inserted in the filecache table. But the etag of the home folder did not change. So it seems to behave like this in general.

jvillafanez (approved):

PR #38996, opened 2021-07-14
Do not allow empty system or app config key

phil-davis:
## Description
Currently this sort of thing works:
```
$ php occ config:app:set "" ""
Config value  for app  set to 
$ php occ config:list ""
{
    "apps": {
        "": {
            "": null
        }
    }
}
```

The app name and config key should not be able to be the empty string. There is never a need to have an app named "" or a key "". This had somehow happened in one developers install, and caused issues when trying to get the config settings.

This PR disallows using an app name or config key that is the empty string. That will avoid accidentally making such settings.

I did not change the corresponding `config:delete` commands - if someone does somehow have one of these "empty string" set, they need to be able to delete it.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Command/Config/App/SetConfig.php
+++ b/core/Command/Config/App/SetConfig.php
@@ -75,6 +75,16 @@ class SetConfig extends Base {
 		$appName = $input->getArgument('app');
 		$configName = $input->getArgument('name');
 
+		if ($appName === '') {
+			$output->writeln('<comment>App name must not be empty.</comment>');
+			return 1;
+		}
+
+		if ($configName === '') {
+			$output->writeln('<comment>Config name must not be empty.</comment>');
+			return 1;
+		}
+
 		if (!\in_array($configName, $this->config->getAppKeys($appName)) && $input->hasParameterOption('--update-only')) {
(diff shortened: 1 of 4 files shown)


phil-davis:
Note: we could probably also not allow a n app name or config key that is all white-space. But that is not so much of an issue - it is just "silly and useless".

core/Command/Config/App/SetConfig.php
```diff
@@ -75,6 +75,16 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		$appName = $input->getArgument('app');
 		$configName = $input->getArgument('name');
 
+		if ($appName === '') {
+			$output->writeln('<comment>App name must not be empty.</comment>');
```
jvillafanez, line 79: `<error>` fits better I think.

jvillafanez (approved):

AlexAndBear (approved):

core/Command/Config/App/SetConfig.php
phil-davis: Agree - I was cut-pasting and not thinking! Fix in the next commit.

JammingBen (approved):

PR #39010, opened 2021-07-15
[tests-only] Do not gather PHPunit coverage with Oracle Db

phil-davis:
## Description
The install of Oracle and then running the PHP unit tests with Oracle as the database takes about 40 minutes in CI.
For other databases, the install plus PHP unit tests pipeline takes about 15 to 20 minutes.
The SonarCloud analysis can only happen after all the coverage reports are available, so currently it takes more than 40 minutes for a developer to find out that their PR failed SonarCloud code coverage analysis.

This PR stops gathering PHPunit coverage for the Oracle database pipeline. That will allow SonarCloud analysis to run and post feedback to the PR on GitHub in about 20 minutes.

Doing this means that we won't have coverage measures for any Oracle-specific code paths. But I expect that there are not many of those. Most database-specific code should be hidden in the DBAL (DataBase Abstraction Layer) dependency, and we are not aiming to test our coverage of that.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToRoot1-git-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31251/107/1

phil-davis:
This was discussed in some chat today. Do we want to lose PHPunit+Oracle coverage stats, in return for getting the SonarCloud result a lot more quickly?

Opinions welcome.

jvillafanez:
In theory, the coverage should be pretty similar, and it should be the same if we don't add specific code for each DB, so I'm ok with the change. It might still be interesting in some PRs where we have code specific for Oracle, so maybe some kind of switch might be a better choice.

phil-davis:
> It might still be interesting in some PRs where we have code specific for Oracle, so maybe some kind of switch might be a better choice.

We could have some special string in the PR title that switches on "full coverage". But I suspect that a special string for that will not be used much, and will be forgotten.

The coverage can be done by making a demonstration PR that puts Oracle back into the unit tests with coverage section in `.drone.star` config section. Maybe that will be OK?

JammingBen:
Since the coverage requirements from SonarCloud actually matter, I think this would be a good thing. So 👍  from my side.

jvillafanez:
> The coverage can be done by making a demonstration PR that puts Oracle back into the unit tests with coverage section in .drone.star config section. Maybe that will be OK?

That's ok for me. As long as it's as easy as adding or removing a comment, that's no problem.

phil-davis:
I rebased just now. Feel free to keep discussing if we want to do this.

AlexAndBear (approved):
♥️

Issue #39034, opened 2021-07-20
open_basedir errors after upgrade to PHP 7.4.21 (from 7.4.20), BUT...


criwayc:
### Steps to reproduce
1. Upgrade or switch to PHP 7.4.21 from any other (7.2, 7.3) or older PHP version (<=7.4.20)
2. Open Owncloud login page in webbrowser
3. Check owncloud.log

### Expected behaviour
no error/warning messages in owncloud.log

### Actual behaviour
logs this error:
 ```
{"reqId":"7IbUb9x1ecNegVjEk5cV","level":3,"time":"2021-07-20T12:07:42+02:00","remoteAddr":"194.246.xxx.xxx","user":"--","app":"PHP","method":"GET","url":"\/index.php\/core\/js\/oc.js?v=0a6bc82e3b6d497ce779c3020f959136","message":"file_exists(): open_basedir restriction in effect. File(\/appinfo\/routes.php) is not within the allowed path(s): (\/webhome\/myowncloud.net\/:\/usr\/share\/php\/:\/usr\/bin\/:\/tmp\/:\/dev\/urandom:\/zpool1\/myowncloud.net) at \/webhome\/myowncloud.net\/lib\/private\/Route\/Router.php#136"
 ```

### Server configuration
**Operating System:** Debian 10.10, Debian 9.13

**Web server:** Apache2, Nginx

**Database:** MariaDB 10.1, MariaDB 10.5

**PHP version:** PHP 7.4.21 (libapache2-mod-php7.4 & php7.4-fpm)

**ownCloud version:** 10.7.0.4

**Updated from an older ownCloud or fresh install:** fresh install

**Where did you install ownCloud from:** https://download.owncloud.org/community/owncloud-10.7.0.tar.bz2

**Signing status (ownCloud 9.0 and above):**
```
No errors have been found.
```

**The content of config/config.php:**
```
{
    "system": {
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "myowncloud.net",
            "www.myowncloud.net"
        ],
        "datadirectory": "\/zpool1\/myowncloud.net",
        "overwrite.cli.url": "https:\/\/myowncloud.net",
        "dbtype": "mysql",
        "version": "10.7.0.4",
        "dbname": "myowncloudnet",
        "dbhost": "localhost",
        "dbtableprefix": "oc_",
[...]

PR #39035, opened 2021-07-20
Fix file_exists call to avoid open_basedir log message

phil-davis:
## Description
PHP 7.4.21 checks open_basedir settings more exactly. Calls to file_exists can
emit log messages like "file_exists(): open_basedir restriction in effect" that
were not emitted by PHP 7.4.20. 

This change fixes an incorrect file_exists check. The open_basedir message will
no longer be emitted in this case.

The PHP 7.4.21 changelog https://www.php.net/ChangeLog-7.php has a fix https://bugs.php.net/bug.php?id=76359
The code of that is https://github.com/php/php-src/pull/7024/files

But I don't see exactly how that change is causing open_basedir to start complaining now.

## Related Issue
- potential fix for #39034 

OP has confirmed that fix works for them: https://github.com/owncloud/core/issues/39034#issuecomment-883359624

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/lib/Controller/ViewController.php
+++ b/apps/files/lib/Controller/ViewController.php
@@ -114,7 +114,7 @@ class ViewController extends Controller {
 		$content = '';
 		$appPath = \OC_App::getAppPath($appName);
 		$scriptPath = $appPath . '/' . $scriptName;
-		if (\file_exists($scriptPath)) {
+		if ($appPath !== false && \file_exists($scriptPath)) {
 			// TODO: sanitize path / script name ?
 			\ob_start();
 			include $scriptPath;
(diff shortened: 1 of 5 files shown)


JammingBen (approved):
Looks reasonable to me as `\OC_App::getAppPath($app) . '/appinfo/routes.php'` never evaluated to `false`.

jvillafanez:
Well, concatenating `false` with a string is something we should avoid anyway.
```
$routingFiles = [];
$appPath = \OC_App::getAppPath($app);
if ($appPath !== false) {
  if (\file_exists("{$appPath}/appinfo/routes.php")) {
    $routingFiles = [$app => $file];
  }
}
```
It seems better.

phil-davis:
> It seems better.

Yes, regardless of the "strabge different behavior" of PHP 7.4.21 that caused this to be noticed, we should fix it anyway. The existing code was silly/wrong.

phil-davis:
I have looked through all the calls to `OC_App::getAppPath`. Most of them already have checks for return value of `false`. I added similar checks where they were missing.

PR #39031, opened 2021-07-20
[tests-only] [full-ci] Send X-Request-Id string in test API requests

phil-davis:
## Description
See #39036 for the squashed version of this PR.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39036, opened 2021-07-20
[tests-only] [full-ci] Send X-Request-Id string in test API requests

phil-davis:
## Description

If SEND_SCENARIO_LINE_REFERENCES environment variable is defined then the acceptance
tests will send a string in header X-Request-ID with each API request. The string
identifies the test suite, feature, scenario line number and step line number in the
scenario. The format is like:

apiComments/editComments.feature:26-28

That indicates the apiComments test suite, editComments feature, the scenario at line 26
and the test step at line 28.

A system-under-test can choose to log that information in any way it wants. For example,
if there is an error in the request (authentication, file not found etc) then the
system-under-test could include the X-Request-ID value in its error log. Doing that can
help someone who is investigating the test results. They can easily match up the error
log entries with the test scenario and step line that caused the error log.

Note: the diffs are big because, somewhere in the calling path of every step, the code has to get the "scenario step string" that needs to be sent in the `X-Request-Id` header. And that string has to be passed in the calls to the functions in the TestHelpers static classes, so that they can keep passing it to HttpRequestHelper::sendRequest. Unfortunately HttpRequestHelper::sendRequest has no knowledge about/access to the `BeforeScenarioScope` object. So it can't generate the string itself.

The interesting code is in FeatureContext https://github.com/owncloud/core/pull/39036/files#diff-4a7e898feb4593b73a10915c04aee63bb8f3b8c7689872fcc716515d3b7b5125
and HttpRequestHelper https://github.com/owncloud/core/pull/39036/files#diff-413f2318db251792e6959a8b2ae9e6affa244ea2a4e219a18693e0280d0af5e1

## Motivation and Context
When inspecting OCIS and reva logs, specially in drone CI, it is difficult to find the place in the server logs that matches with the test scenario and test step that is reporting a failure. This test suite feature should help.

## How Has This Been Tested?
CI - see PR #39031 which documents the development of this code and CI runs that demonstrate that useful `X-Request-Id` strings are being set.

OCIS PR https://drone.owncloud.com/owncloud/ocis/6066 passes against this branch.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -176,6 +177,29 @@ class FeatureContext extends BehatVariablesContext {
...
+	 * The suite name, feature name and scenario line number.
+	 * Example: apiComments-createComments-s24
+	 *
+	 * @var string
+	 */
+	private $scenarioString = '';
+
+	/**
+	 * A full unique reference to the step that is currently executing.
+	 * Example: apiComments-createComments-s24-l28
```
individual-it, line 190: why not using the same format that is used in Gherkin `apiComments/createComments.feature:24-128`?

phil-davis: Can do. I didn't realize that there is a format like that. It will be easy to change.

individual-it: I think that would be good because if would reduce the turns the :brain: needs to take

phil-davis: Done.

individual-it (approved):

phil-davis:
I will wait to merge this until after `release-10.8.0` has been merged back. The changes are big, and I don't want to conflict with the merge-back.

phil-davis:
release-10.8.0 branch has been merged. I just rebased this PR. It is OK to merge when CI passes.

PR #39044, opened 2021-07-21
Detect LDAP connection errors during `occ user:sync`

AlexAndBear:
## Description
Needs https://github.com/owncloud/user_ldap/pull/669 to work properly.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4642

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Command/User/SyncBackend.php
+++ b/core/Command/User/SyncBackend.php
@@ -219,7 +219,6 @@ class SyncBackend extends Command {
 
 		$output->writeln('');
 		$backendClass = \get_class($backend);
-
 		if ($input->getOption('seenOnly')) {
 			$output->writeln("Updating seen accounts from $backendClass ...");
 			$iterator = new SeenUsersIterator($this->accountMapper, $backendClass);
--- a/core/Command/User/SyncBackend.php
+++ b/core/Command/User/SyncBackend.php
@@ -219,12 +219,13 @@ class SyncBackend extends Command {
 
 		$output->writeln('');
 		$backendClass = \get_class($backend);
+
 		if ($input->getOption('seenOnly')) {
 			$output->writeln("Updating seen accounts from $backendClass ...");
 			$iterator = new SeenUsersIterator($this->accountMapper, $backendClass);
 		} else {
 			$output->writeln("Inserting new and updating all known users from $backendClass ...");
-			$iterator = new BackendUsersIterator($backend);
+			$iterator = new BackendUsersIterator($backend, '', true);
 		}
 
 		$progress = new ProgressBar($output);
--- a/lib/private/User/Sync/BackendUsersIterator.php
+++ b/lib/private/User/Sync/BackendUsersIterator.php
@@ -44,11 +44,11 @@ class BackendUsersIterator extends UsersIterator {
 	/** @var string search for the uid string in backend */
 	private $search;
 
-	/** @throws \Exception for the uid string in backend */
-	private $testConnection;
+	/** @var boolean check if backend encountered connections errors */
+	private $checkForConnectionErrors;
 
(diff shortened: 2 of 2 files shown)


jvillafanez:
I'm not sure how useful this solution will be. For the LDAP case, you have to search over all the results in order to count the users, which means that you're doing the same requests twice including pagination.

If we consider the network is unstable, counting could go wrong too. I'm not sure how it's handled, but unless the error is noticeable, the counting could say there are 500 users but you could sync 750 because there was an error while requesting the second page of the results for the count.

In addition, just an error message might not be enough. I've been told that the recommendation is to sync the users via (linux) cron job each 15-30 minutes, which means the command is expected to run automatically without user intervention, so it's unlikely that there will be anyone noticing that the command is showing an error message.

core/Command/User/SyncBackend.php
```diff
@@ -228,24 +228,50 @@ private function syncMultipleUsers(
...
-			/* @phan-suppress-next-line PhanUndeclaredMethod */
-			$max = $backend->countUsers();
+
+		$toProcessUsersCount = null;
+
+		if ($backend->implementsActions(\OC_User_Backend::COUNT_USERS)) {
+			if ($input->getOption('seenOnly')) {
+				$toProcessUsersCount = 0;
+
+				$this->userManager->callForSeenUsers(function (IUser $user) use ($backend, &$toProcessUsersCount) {
```
JammingBen, line 238: @JanAckermann Maybe we can skip this ugly loop by calling `$toProcessUsersCount = $this->accountMapper->getUserCountPerBackend(true);`. I seems to return an array with all the available backends including their users.

JammingBen:
> In addition, just an error message might not be enough. I've been told that the recommendation is to sync the users via (linux) cron job each 15-30 minutes, which means the command is expected to run automatically without user intervention, so it's unlikely that there will be anyone noticing that the command is showing an error message.

We already log the error, I think that's all we can do for now. I wouldn't go for active notifications like emails in case of a cron run, especially if the cron runs every 15 minutes.

JanAckermann:
@jvillafanez 

>I'm not sure how useful this solution will be. For the LDAP case, you have to search over all the results in order to count the users, which means that you're doing the same requests twice including pagination.

Well, this is not good at all, I am not an expert in LDAP, I thought it would be like mysql `SELECT COUNT(*) FROM users`....
If it is not, well it breaks the whole solution.

>In addition, just an error message might not be enough. I've been told that the recommendation is to sync the users via (linux) cron job each 15-30 minutes, which means the command is expected to run automatically without user intervention, so it's unlikely that there will be anyone noticing that the command is showing an error message.

Errors are in logs, the issue is about running the command manually. 

So, I don't know how to solve it if we are not able to throw an Exception here

lib/private/User/Sync/BackendUsersIterator.php
```diff
@@ -63,7 +72,13 @@ public function next() {
 		if ($this->hasMoreData && $this->dataPos >= $this->endPos) {
 			$this->page++;
 			$offset = $this->page * self::LIMIT;
+
 			$this->data = $this->backend->getUsers($this->search, self::LIMIT, $offset);
+
+			if($this->checkForConnectionErrors && method_exists($this->backend,'testConnection')){
```
JammingBen, line 78:
```suggestion
			if($this->checkForConnectionErrors && method_exists($this->backend, 'checkForConnectionErrors')){
```

JammingBen:
This is a "won't fix" for now: https://github.com/owncloud/enterprise/issues/4642#issuecomment-986586722

PR #39015, opened 2021-07-16
Log number of mounts when moving encryption keys fails 

IljaN:
## Description
Due to wrong configuration or bugs it is possible that more than one mount is returned. In this case we should log the mount-count for easier debugging.

## Motivation and Context
Ease the debugging process in production-systems.

## How Has This Been Tested?
More info in log-line no additional tests needed.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #38583, opened 2021-03-25
[full-ci] Enable streaming for propfind

DeepDiver1975:
## Description
Propfind requests will now be streamed to reduce memory usage with large responses.
Additionally, the new config `dav.propfind.depth_infinity` has been added to tell clients whether `depth=infinity` is allowed for propfind requests. It defaults to true. 

https://github.com/owncloud/core/pull/36336 reloaded
Fixes https://github.com/owncloud/core/issues/14531

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket raised:  https://github.com/owncloud/docs/issues/4127
- [x] Changelog item

PR #39058, opened 2021-07-26
Allow copying file keys when storage has multiple mount points

butonic:
cc @IljaN @cdamken 
In the old logic might break when a group share has been mounted by multiple users and files are repeatedly moved in and out of that group share while encryption is enabled.

- [ ] Needs a Unit test that reproduces this bug
    - enable encryption
    - create group share, accept it for multiple users
    - might depend on order of acceptance (which is the order read from db)
    - has to do with additional user filesystem initializations in https://github.com/owncloud/core/blob/78fcc9e0a92329ecbb2aca750210db817dfa6263/lib/private/Encryption/Keys/Storage.php#L386, called by https://github.com/owncloud/core/blob/78fcc9e0a92329ecbb2aca750210db817dfa6263/lib/private/Encryption/Keys/Storage.php#L344, called by https://github.com/owncloud/core/blob/78fcc9e0a92329ecbb2aca750210db817dfa6263/lib/private/Encryption/Keys/Storage.php#L319-L320 (which initializes source & target), called by https://github.com/owncloud/core/blob/d329971010ec6443d1e924ed989d0ec2ec27d69d/lib/private/Files/Storage/Wrapper/Encryption.php#L783

### DIFF
--- a/changelog/unreleased/39058
+++ b/changelog/unreleased/39058
@@ -1,0 +1,3 @@
+Bugfix: Allow copying file keys when storage has multiple mount points
+
+https://github.com/owncloud/core/pull/39058
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -776,7 +776,8 @@ class Encryption extends Wrapper {
 		// first copy the keys that we reuse the existing file key on the target location
 		// and don't create a new one which would break versions for example.
 		$mount = $this->mountManager->findByStorageId($sourceStorage->getId());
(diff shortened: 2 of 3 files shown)


IljaN:
@JanAckermann Could your team please take care so this lands in 10.9?

IljaN:
Related: https://github.com/owncloud/core/pull/39015

JanAckermann:
@IljaN will take care but https://github.com/owncloud/core/pull/38583 which needs to land in 10.9 also, has a higher prio

IljaN (approved):

PR #39085, opened 2021-08-02
[tests-only] Try phpunit9

phil-davis:


phil-davis:
`The at() matcher has been deprecated. It will be removed in PHPUnit 10. Please refactor your test to not rely on the order in which methods are invoked.`
297 PHP unit test fails https://drone.owncloud.com/owncloud/core/31713/9/8

There will be a bit of work to do to refactor all of that - not today!

phil-davis:
I adjusted a couple of things in #39087
There is not much more that can easily be done - the `at()` replacement can be done with phpunit8 but will be a big project!
Closing

PR #39088, opened 2021-08-02
Use md5 hash for chunk name

AlexAndBear:
## Description
Before this PR, while uploading a file, chunks were created with filename
consist of original filename plus chunk suffix.
This could lead to a too long filename error.
With this PR the chunk filenames consist of md5 hash and chunk suffix, which
decreases the filename and prevents a too long filename error.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4692
## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -305,7 +305,8 @@ class File extends Node implements IFile, IFileNode {
 	private function getPartFileBasePath($path) {
 		$partFileInStorage = \OC::$server->getConfig()->getSystemValue('part_file_in_storage', true);
 		if ($partFileInStorage) {
-			return $path;
+			$pathParts = \pathinfo($path);
+			return Filesystem::normalizePath($pathParts['dirname'].'/'.\md5($pathParts['basename']));
 		} else {
 			return \md5($path); // will place it in the root of the view with a unique name
 		}
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -305,8 +305,7 @@ class File extends Node implements IFile, IFileNode {
 	private function getPartFileBasePath($path) {
 		$partFileInStorage = \OC::$server->getConfig()->getSystemValue('part_file_in_storage', true);
 		if ($partFileInStorage) {
-			$pathParts = \pathinfo($path);
-			return Filesystem::normalizePath($pathParts['dirname'].'/'.\md5($pathParts['basename']));
+			return Filesystem::hashFileName($path);
 		} else {
 			return \md5($path); // will place it in the root of the view with a unique name
 		}
--- a/changelog/unreleased/39088
+++ b/changelog/unreleased/39088
@@ -1,0 +1,10 @@
+Bugfix: Hash chunk(v2) filename
+
+Before this PR, while uploading a file, chunks(v2) were created with filename
+consist of original filename plus chunk suffix.
+This could lead to a too long filename error.
+With this PR the chunk filenames consist of md5 hash and chunk suffix, which
+decreases the filename and prevents a too long filename error.
+
+https://github.com/owncloud/core/pull/39088
+https://github.com/owncloud/enterprise/issues/4692
--- a/lib/private/Files/Filesystem.php
+++ b/lib/private/Files/Filesystem.php
@@ -1091,7 +1091,14 @@ class Filesystem {
 	}
 
 	/**
-	 * Hash the file name in a given path while preserving the rest of it
+	 * Hash the file name in a given path while preserving the rest of it.
+	 * This method is used in combination with file chunking to prevent the
+	 * file name from exceeding its limit.
+	 *
+	 * Examples:
+	 * /files/test.txt -> /files/dd18bf3a8e0a2a3e53e2661c7fb53534
+	 * files/test.txt -> files/dd18bf3a8e0a2a3e53e2661c7fb53534
+	 * /test.txt -> /dd18bf3a8e0a2a3e53e2661c7fb53534
 	 *
 	 * @param string $path
 	 * @return string
--- a/lib/private/Files/Filesystem.php
+++ b/lib/private/Files/Filesystem.php
@@ -1089,4 +1089,22 @@ class Filesystem {
 	public static function getETag($path) {
 		return self::$defaultInstance->getETag($path);
 	}
+
+	/**
+	 * Hash the file name in a given path while preserving the rest of it
+	 *
+	 * @param string $path
+	 * @return string
+	 */
+	public static function hashFileName($path) {
+		$pathParts = \pathinfo($path);
+		$hashedFileName = \md5($pathParts['basename']);
+
+		if (isset($pathParts['dirname']) && $pathParts['dirname'] !== '.') {
+			$dirname = \rtrim($pathParts['dirname'], '/');
+			$hashedFileName = $dirname . '/' . $hashedFileName;
+		}
+
+		return $hashedFileName;
+	}
 }
--- a/lib/private/Files/Stream/Checksum.php
+++ b/lib/private/Files/Stream/Checksum.php
@@ -213,6 +213,18 @@ class Checksum extends Wrapper {
 			return self::$checksums[$path];
 		}
 
+		// check the md5 in case only the filename was hashed
+		// this was introduced in 10.9
+		$pathParts = \pathinfo($path);
+
+		$hashedNameToCheck = \md5($pathParts['basename']);
+		if (isset($pathParts['dirname']) && $pathParts['dirname'] !== '.') {
+			$hashedNameToCheck = $pathParts['dirname'] . '/' . $hashedNameToCheck;
+		}
+		if (isset(self::$checksums[$hashedNameToCheck])) {
+			return self::$checksums[$hashedNameToCheck];
+		}
+
 		// check the md5($path) in case "part_file_in_storage" is set to false
 		// see apps/dav/lib/Connector/Sabre/File.php getPartFileBasePath  (around line 305)
 		// strip initial dir, usually "files" from "files/dir1/dir2"
--- a/lib/private/Files/Stream/Checksum.php
+++ b/lib/private/Files/Stream/Checksum.php
@@ -24,6 +24,7 @@ namespace OC\Files\Stream;
 
 use Icewind\Streams\Wrapper;
 use OC\Cache\CappedMemoryCache;
+use OC\Files\Filesystem;
 
 /**
  * Computes the checksum of the wrapped stream. The checksum can be retrieved with
@@ -213,14 +214,9 @@ class Checksum extends Wrapper {
 			return self::$checksums[$path];
 		}
 
-		// check the md5 in case only the filename was hashed
+		// check the hash in case only the filename was hashed
 		// this was introduced in 10.9
-		$pathParts = \pathinfo($path);
-
-		$hashedNameToCheck = \md5($pathParts['basename']);
-		if (isset($pathParts['dirname']) && $pathParts['dirname'] !== '.') {
-			$hashedNameToCheck = $pathParts['dirname'] . '/' . $hashedNameToCheck;
-		}
+		$hashedNameToCheck = Filesystem::hashFileName($path);
 		if (isset(self::$checksums[$hashedNameToCheck])) {
 			return self::$checksums[$hashedNameToCheck];
 		}
--- a/tests/lib/Files/FilesystemTest.php
+++ b/tests/lib/Files/FilesystemTest.php
@@ -562,4 +562,23 @@ class FilesystemTest extends TestCase {
 		\OC::$server->getMountProviderCollection()->registerProvider($mountProvider);
 		$this->assertEquals('/foo/bar/', Filesystem::getMountPoint('/foo/bar'));
 	}
+
+	public function getHashedFileNamePaths() {
+		return [
+			['/test/test.txt', '/test/dd18bf3a8e0a2a3e53e2661c7fb53534'],
+			['test/test.txt', 'test/dd18bf3a8e0a2a3e53e2661c7fb53534'],
+			['/test.txt', '/dd18bf3a8e0a2a3e53e2661c7fb53534'],
+			['test.txt', 'dd18bf3a8e0a2a3e53e2661c7fb53534'],
+		];
+	}
+
+	/**
+	 * @dataProvider getHashedFileNamePaths
+	 *
+	 * @param string $path
+	 * @param string $expected
+	 */
+	public function testHashFileName($path, $expected) {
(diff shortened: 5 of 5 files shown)


ownclouders:

https://drone.owncloud.com/owncloud/core/31719/173/1

.htaccess
```diff
@@ -94,3 +94,35 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
phil-davis, line 97: Accidental commit

ownclouders:

https://drone.owncloud.com/owncloud/core/31723/59/1

phil-davis:
In master I can do an "old" chunked upload of:
`नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-012345678.txt`

But an "old" chunked upload of:
`नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-0123456789.txt`
fails.

So that is an example of the current point of failure.

Then I switch to this branch, and the longer file name still fails. I see a message like:
`fopen(/home/phil/git/owncloud/core/data/Alice/cache/नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-नेपालि-file-नाम-0123456789.txt-chunking-1298-0.haWImtSPYnIig60i.part): failed to open stream: File name too long at /home/phil/git/owncloud/core/lib/private/Files/Storage/Local.php#298`

That looks strange - all the UTF filename is there, plus there is `.haWImtSPYnIig60i` before `.part`

I was expecting that the full file path might be something like:
`fopen(/home/phil/git/owncloud/core/data/Alice/cache/haWImtSPYnIig60i.txt-chunking-1298-0.part)`

And also, now that I see that the full path includes the real path to the user "home" and the username, that if there is a reasonably short maximum path length on the OS+filesystem in use, then, for example, a long username cannot store a super-long filename.

JammingBen:
We're still missing some parts of the implementation. That's probably why the hashing does not work as expected yet.

> And also, now that I see that the full path includes the real path to the user "home" and the username, that if there is a reasonably short maximum path length on the OS+filesystem in use, then, for example, a long username cannot store a super-long filename.

The filename-length-check should be done based on the filename only, so the username does not affect the result. At least that's what we observed - do you have other information?

Additionally, there is the config setting `part_file_in_storage`. When set to `false`, the chunked parts wont be saved in the user directory.

phil-davis:
> The filename-length-check should be done based on the filename only, so the username does not affect the result. At least that's what we observed - do you have other information?

AFAIK the maximum path length in most Unix file-systems is "long". Much longer than the filename maximum. So ignore my comments about path length - people will only come across that if they have made really long and/or many levels of directories.

JanAckermann:
@phil-davis on your system webDav chunking is enabled, in our system not :)
Any idea how you enabled that?

phil-davis:
> @phil-davis on your system webDav chunking is enabled, in our system not :)
> Any idea how you enabled that?

I am running API acceptance tests locally, not using any client. The tests just "hit" the chunking endpoints, they don't check any capabilities or other evidence about "chunking enabled". Actually, I don't see how chunking can be purposely disabled.

JanAckermann:
@phil-davis the example with the `-chunking-` part in filename seems to be legacy code (old chunking mechanism) and is not used by WebUI or modern desktop clients, I think we should not touch it.

ownclouders:
:boom: Acceptance tests pipeline <strong>cliLocalStorage-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31736/121/1

ownclouders:
:boom: Acceptance tests pipeline <strong>webUISharingExt2-10.7.0-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31740/165/1

ownclouders:
:boom: Acceptance tests pipeline <strong>cliExternalStorage-latest-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31740/127/1

phil-davis:
There were some pipelines that failed in the "setup"  steps. I restarted CI.

IljaN:
Code looks good :+1: 

We should ask some people from the client-teams to take a look because there could be some assumptions made (accidentally) about the chunk naming in client-code. @michaelstingl

michaelstingl:
> We should ask some people from the client-teams to take a look because there could be some assumptions made (accidentally) about the chunk naming in client-code. @michaelstingl

did you compare `mitmproxy` flows from desktop client with this PR?

IljaN:
The http flows should be unchanged except for the hashed values. I guess trying chunked upload with the latest clients should do the trick.

JanAckermann:
@IljaN agreed, http request should not change here, tested successfully with desktop client for linux (2.5.1)

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareManagementBasicToRoot-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/71/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareUpdateToRoot-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/85/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareUpdateToShares-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/86/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToShares1-git-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/114/1

ownclouders:
:boom: Acceptance tests pipeline <strong>cliLocalStorage-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/121/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareCreateSpecialToRoot1-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/64/1

ownclouders:

https://drone.owncloud.com/owncloud/core/31747/60/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavPreviews-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/101/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToRoot2-10.7.0-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/113/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavUpload2-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/105/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToRoot2-git-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/111/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareReshareToShares1-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/80/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiSharingNotificationsToRoot-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/106/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavLocks-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/94/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavEtagPropagation1-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/92/1

ownclouders:
:boom: Acceptance tests pipeline <strong>cliMain-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/122/1

ownclouders:

https://drone.owncloud.com/owncloud/core/31747/88/1

ownclouders:
:boom: Acceptance tests pipeline <strong>cliExternalStorage-latest-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/127/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavLocksUnlock-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/97/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareReshareToShares3-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/84/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToShares1-latest-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/115/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToShares2-10.7.0-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/119/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31747/91/1

IljaN (approved):

jvillafanez:
Some questions:
* What happens if the code changes during an upload? The use case could be:
  1. Start uploading a big file. Upload 5 out of 20 chunks.
  2. The server upgrades to 10.9 (or when this patch is applied)
  3. Keep uploading the next chunks.

  As long as the server doesn't explode, I think it's fine. If the server has to signal someone to the client to restart the upload, that's ok. I expect the server to recover automatically at some point.
* Should we move the code to a function somewhere? It should be easier to adjust the code in the future if we need. We might want not to hash the name so it's easier to identify what file had problems if the chunks are kept for too long, or we could use a different algorithm.

Code changes looks good, assuming we aren't missing anything.

mrow4a (changes requested):
make it configurable by flag, otherwise we explode on upgrade in minor version.

jvillafanez:
@mrow4a could you elaborate a bit more?

For the use case described in https://github.com/owncloud/core/pull/39088#issuecomment-893283657 , I expect the server throw an error due to missing chunks (the first ones might not be found due to the code change), or maybe waiting until a timeout. In any case, I expect the client to retry the upload and the chunks should be eventually removed (I think we have a cron job for cleaning the chunks)
Note that, anyway, the use case is unlikely to happen because the admins usually announce a downtime for the upgrade so the users aren't expecting to be uploading files just before the upgrade.
If we use a flag, we're more likely to bump into this scenario because a downtime isn't expected to happen just to change the value.

mrow4a:
@jvillafanez it is just risk management. We would change an algorithm for chunking. I would make it optional and applicable only to these oc instances that have problems with too long filenames (which are mostly the ones with non-latin characters). When going to OC11 we could make it default mechanism.

jvillafanez:
As far as I know, we're only changing the name of the part file, anything else remains the same. I don't think we're storing permanently the name anywhere, the name is transient data I think.
Taking this into account, unless I'm missing something, the only potential problem is the one explained above, which should fix itself after a while. It seems a low risk.

The problem I see with the option is that I don't really see any reason to keep the current behaviour. The alternatives we have are: keep the current behaviour (which is problematic) or use md5 in the filename. We're already discarding the current behaviour, so the only option left is to use md5. Unless we include some other useful option, using a flag to select from only one valid option doesn't seem good.

phil-davis:
This is an internal implementation "detail" changing. AFAICT the external API remains the same. (the existing test scenarios all work unchanged) So the only issue is the "migration" of uploads that are in progress at the time of the server upgrade. (as discussed above)

In-progress uploads will fail when they get to the point when they try to assemble all the chunks. So the client will get an error and restart the upload. IMO that is acceptable.

Any "orphaned" chunks get cleaned up by `occ dav:cleanup-chunks`

JammingBen:
> What happens if the code changes during an upload? 

When updating OC, we advice our users to enable the maintenance mode. I don't see a scenario where the server crashes due to the change, but I'll try and see what happens.

> Should we move the code to a function somewhere? It should be easier to adjust the code in the future if we need. We might want not to hash the name so it's easier to identify what file had problems if the chunks are kept for too long, or we could use a different algorithm.

Good idea. Just create a private method in `apps/dav/lib/Connector/Sabre/File.php`? Or do you think about a more "generic" place such as a helper?

> As far as I know, we're only changing the name of the part file, anything else remains the same. I don't think we're storing permanently the name anywhere, the name is transient data I think.

Exactly. Therefore I don't see a good reason to make this configurable via flag.

mrow4a (approved):
assuming risks are accepted and all integration tests passing

jvillafanez:
> Good idea. Just create a private method in apps/dav/lib/Connector/Sabre/File.php? Or do you think about a more "generic" place such as a helper?

It needs to be used in the "Files/Stream/Checksum.php", so it should be somewhere in core. A helper might make sense; we could include a generic way to check for part files there too (maybe not in this PR, but at some point)

JammingBen:
@jvillafanez Please check again. I added the method in `Filesystem`, as it currently holds plenty helper methods. Any objections?

ownclouders:

https://drone.owncloud.com/owncloud/core/31767/131/1

jvillafanez:
No objections from my side.
Just include some additional information in the PHPDoc, I think it's important to mention that it's expected to be used with the file chunking to prevent file names over the limit (in case we want to adjust the implementation in the future), and some examples for the expected output (to show if there will be a leading slash or not, what part of the file name will be hashed).
We should also include some unit tests.

PR #39091, opened 2021-08-03
feat: encryption keys location can be specified by the storage implementation

DeepDiver1975:
## Description
This allows storage implementations to specify the location for encryption keys
https://github.com/owncloud/files_spaces needs this because the space does not belong to a user.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4687

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Encryption/Util.php
+++ b/lib/private/Encryption/Util.php
@@ -73,6 +73,9 @@ class Util {
 	/** @var array paths excluded from encryption */
 	protected $excludedPaths;
 
+	/** @var array paths excluded full paths from encryption */
+	protected $excludedFullPaths;
+
 	/** @var \OC\Group\Manager $manager */
 	protected $groupManager;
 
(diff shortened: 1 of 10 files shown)


cdamken:
Is there a test from the keys that are already in the users to be moved in the spaces?

DeepDiver1975:
> Is there a process to move the keys that are already on the users to be moved in the spaces?

no - and I doubt that there is a general applicable solution.

keys are stored in the individual user's encryption key location.
I'd perform a manual analysis of the explicit system and move key files as needed ...

lib/private/Encryption/Util.php
```diff
@@ -104,6 +107,9 @@ public function __construct(
 		$this->excludedPaths[] = 'avatars';
 		$this->excludedPaths[] = 'avatar.png';
 		$this->excludedPaths[] = 'avatar.jpg';
+
+		// TODO: find a sophisticated solution
```
DeepDiver1975, line 111: Yeah - that's fishy ....

JammingBen: I'm afraid someone else has to come up with a more sophisticated solution, I don't see a general approach without mentioning the spaces app or providing a public interface to hook up to. This goes not only for this line here - we have multiple references to the spaces app, as you probably already have seen :(

PR #39092, opened 2021-08-03
[tests-only] [full-ci] Refactor previews tests so they can run on oC10 or OCIS

phil-davis:
## Description
There are various `previews.feature` tests that are not being skipped on OCIS. And in the OCIS  repo there are versions of some of these tests that are "bug demos". These days we have the expected-failures files. So we can run the "real" tests from core in OCIS CI  and add them to expected-failures if they do not pass.

This PR:

1) Removes the skip tags from all the previews tests, so that they will all run in all CI
2) Provides two different test scenarios for sizing of previews of files downloaded through the webdav API - the behavior of oC10 is that the exact requested size is returned (even if the request is something "weird" like 1 pixel by 32 pixels). OCIS behaves differently (and more reasonably, depending on how you define "reasonable" - it always returns a width-height that preserves the aspect-ratio of the original file). That is discussed in https://github.com/owncloud/ocis/issues/2069 - and we are accepting that the 2 implementations can have this difference in behavior.

## Related Issue
https://github.com/owncloud/ocis/issues/2313

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavPreviews/previews.feature
+++ b/tests/acceptance/features/apiWebdavPreviews/previews.feature
@@ -4,21 +4,6 @@ Feature: previews of files downloaded through the webdav API
   Background:
     Given user "Alice" has been created with default attributes and without skeleton files
 
-  @skipOnOcis @issue-ocis-2069
-  Scenario Outline: download different sizes of previews of file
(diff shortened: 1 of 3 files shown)


SwikritiT (approved):
LGTM

PR #39101, opened 2021-08-10
config_apps_sample changes for WND 2.1.0

mmattel:
## Description
config_apps_sample changes for WND 2.1.0

## Related Issue
As part of https://github.com/owncloud/windows_network_drive/pull/366 (Additional check for detecting directories with libsmbclient 4.11+)

## Motivation and Context
Update config

## How Has This Been Tested?
no tests necessary

## Types of changes

- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

config-to-docs run will be part of a PR in docs, PR will be linked when ready

### DIFF
--- a/config/config.apps.sample.php
+++ b/config/config.apps.sample.php
@@ -428,11 +428,11 @@ $CONFIG = [
  *
  * Possible keys: `wnd.groupmembership.checkUserFirst` BOOL
  *
- * *Note* With WND 2.1.0, key `wnd.storage.testForHiddenMount` is obsolete and has been removed completely
+ * *Note* With WND 2.1.0, key `wnd.storage.testForHiddenMount` is obsolete and has been removed completely.
  */
 
 /**
- * Mandatory listener reconnect to the database
+ * Mandatory Listener Reconnect to the Database
  * The listener will reconnect to the DB after given seconds. This will
  * prevent the listener to crash if the connection to the DB is closed after
  * being idle for a long time.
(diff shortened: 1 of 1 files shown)


config/config.apps.sample.php
```diff
@@ -443,14 +443,25 @@
...
- * Check for visible target mount folders when connecting
- * Ensure that the connectivity check verifies the mount point is visible.
- * This means the target folder is NOT hidden.
- * Setting this option to false can speed up the connectivity check by skipping
- * this step. It will be the admin's responsibility to ensure the mount
- * point is visible. This setting will affect all the WND mount points.
+ * The way file attributes for folders and files will be handled
+ * There are 3 possible values: "none", "stat" and "getxattr":
+ *
+ * - "stat". This is the default in case the option is missing or has an invalid value
```
phil-davis, line 449:
```suggestion
 * - "stat". This is the default if the option is missing or has an invalid value
```

phil-davis (commented):
What do we do about this sort of change that depends on an app release version, and not on the oC10 core release?
Sites that are still using the existing WND app version will still be able to use `wnd.storage.testForHiddenMount` but it will not be in the docs.

mmattel:
> What do we do about this sort of change that depends on an app release version...

This is a special case here. The source PR is https://github.com/owncloud/windows_network_drive/pull/366 which will be merged hopefully soon and will be part of the wnd 2.1.0 release. That PR obsoletes and removes the `wnd.storage.testForHiddenMount` + the code for it. Means, when the new WND version is released, the key is dead.
But you are completely right, that the situation is not optimal. For core this PR will go into 10.9 so we have a gap between now and the release of 10.9. For docs the situation is more complicated as it will appear immediately for 10.8, removing the key when merged- in case what I strongly believe it will be backported. But I also believe that customers will go quite fast to WND 2.1.0...

mmattel:
@phil-davis I have added a text note that the key is obsolete with wnd2.1.0 and is removed

phil-davis (approved):

mmattel:
Put the PR to draft to protect against accidentally merging.
We will get a notice when owncloud/windows_network_drive#366 (Additional check for detecting directories with libsmbclient 4.11+) is merged.

PR #39105, opened 2021-08-11
Include a flag to mark if the loginName2userName has been processed

jvillafanez:
## Description
Add a flag when asking for the username to indicate that someone has already processed the request and there is no need for others to check further.

Requires:
* https://github.com/owncloud/user_ldap/pull/675

## Related Issue

- Fixes <issue_link>

## Motivation and Context
When the user_ldap app is enabled, logging in with the admin user triggered an LDAP request. This shouldn't happen when the admin's account is empty and there is no relation with any ldap user.
If the LDAP server is busy, it will slow down the login process even for non-ldap users.

## How Has This Been Tested?
Having the user_ldap app enabled (and patched), log in with the admin user. With both patches, there is no bind request send to the LDAP server when logged in with the admin user.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/lib/Address.php
+++ b/apps/federatedfilesharing/lib/Address.php
@@ -155,10 +155,11 @@ class Address {
 	protected function translateUid($uid) {
 		// FIXME this should be a method in the user management instead
 		// Move to a helper instead of C&P meanwhile?
+		$handled = false;
 		\OCP\Util::emitHook(
 			'\OCA\Files_Sharing\API\Server2Server',
 			'preLoginNameUsedAsUserName',
-			['uid' => &$uid]
+			['uid' => &$uid, 'hasBeenHandled' => &$handled]
 		);
 		return $uid;
 	}
--- a/apps/federatedfilesharing/lib/Address.php
+++ b/apps/federatedfilesharing/lib/Address.php
@@ -156,6 +156,8 @@ class Address {
 		// FIXME this should be a method in the user management instead
 		// Move to a helper instead of C&P meanwhile?
 		$handled = false;
+		// the $handled var will be sent as reference so the listeners can use it as a flag
+		// in order to know if the event has been processed already or not.
 		\OCP\Util::emitHook(
 			'\OCA\Files_Sharing\API\Server2Server',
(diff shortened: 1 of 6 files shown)


jvillafanez:
This is pending to confirm that it fixes the original issue. Code should be ready to review, and only the changelog is missing.

AlexAndBear (changes requested):
I don't like how the value is passed by reference through multiple files, maybe we should use a class to handle this?

jvillafanez:
I don't think we have a better choice. The result comes from an event, so we shouldn't change the approach. We have multiple listeners and all of them have to respond, so that flag will indicate that we have a valid result and the next listeners shouldn't need to process anything.

A better approach would be to use an interface with a registration process so core could have a better control on who and how the calls are done. The problem is that this is a breaking change which will need changes in all the related apps.

Note that the goal of the fix is that if someone has already returned a valid value for the event, the rest of the listeners shouldn't need to bother to process the event because we already have a result. Without changing the approach of using an event (which, as said, is questionable), using a flag in the event is the easiest solution.

AlexAndBear (approved):
Understood, maybe you could add some comments, that those vars are called by reference, so we make sure,no bugs will be implemented later...

JanAckermann:
Restarted CI

ownclouders:

https://drone.owncloud.com/owncloud/core/32303/124/1

phil-davis:
Note: Issue #39213 PR #39219 will skip it. That will be merged when CI finishes. After that you can rebase, and that annoying fail will be gone.

JammingBen (approved):

jnweiger:
Confirmed fixed in 10.9.0 beta1

When user admin types his correct password, no ldap requests are done. 
When user admin mistypes his password, ldap is still queried. That is as expected.

jvillafanez:
Weird... https://github.com/owncloud/user_ldap/pull/675 should be needed and it hasn't been merged yet... :thinking: 
I'd expect no changes comparing the behaviour with previous versions.

Issue #38549, opened 2021-03-18
[QA] occ upgrade does nothing, although web-ui says 'Update needed'


jnweiger:
Seen in 10.7.0 RC1

I have no clear reproducer for this, but I ran by accdent into this situation multiple times:

The web ui asks for update.
![image](https://user-images.githubusercontent.com/1108546/111711431-f4727a80-884b-11eb-97d1-1ed0bf3d44bc.png)

The occ command does not work: BAD
```
occ upgrade   
ownCloud is already latest version
```

Attempted workaround:
* in config.php switch `upgrade.disable-web' => false,`
* a new error occurs:
`It looks like your instance may host many files and/or users. To ensure a smooth upgrade process, please use the command line updater (occ upgrade)`

This is a deadlock situation.

Evil workaround: 
* hack into "lib/base.php" and force `$tooBig = false;` after the ldap and user evaluation code and before the error printing.
* now the web interface allows to run the upgrade procedure. Login is possible again.

Suggested improvement:
 Add a --force option to `occ upgrade` to try as hard as the web version would do.

micbar:
> I have no clear reproducer for this, but I ran by accdent into this situation multiple times:

we need that context, otherwise we cannot qualify

jnweiger:
> > I have no clear reproducer for this, but I ran by accdent into this situation multiple times:
> 
> we need that context, otherwise we cannot qualify

One instance, where this happened repeatedly yesterday and today is still online.
All I can do for now, is leave that instance alone, and give investigators ssh access.

cortho:
copied from https://github.com/owncloud/core/issues/37361

Still having this issue with a recent ownCloud 10.7

> > The expected behavior is to have two app versions in two folders, when updating an app that is installed in the read-only app folder?
> 
> yes, `'writable' => false,` means `oC has no permissions to manage files and directories in this app folder` so deletion of the previous app version is not allowed.

While I agree with this behavior I think the ownCloud app loading mechanism is broken. I did not dig into the code but shouldn't ownCloud browse all available apps and then use the most recent version, regardless of their locations?

Since it is the default ownCloud behavior to come with a read only apps directory and a writable apps-external directory it must not be delegated to the end user managing apps versions.

I consider this a nasty bug which leads to lots of confusion, especially for beginners. See some postings from central as reference:
* https://central.owncloud.org/t/automatic-updating-is-disabled/32869/
* https://central.owncloud.org/t/update-needed-and-nobody/32982
[...]

PR #39108, opened 2021-08-12
Update appinfo cache only if the app version is newer

jvillafanez:
## Description
Cache app info for the appid only if the version is newer.

## Related Issue
https://github.com/owncloud/core/issues/38549

## Motivation and Context
If there are multiple versions of the same app in different locations, we're expected to use the one with the latest version.
Note that the problem happens having a local cache active.

## How Has This Been Tested?
1. Having at least 2 different app locations, install the same app (not the latest version) in all of them
2. In the config.php file, set the first app location as writable
3. Update the app via `occ market:upgrade <app>`

There is no problem with the upgrade. The installed version in the DB is the latest one (downloaded from the market). Reloading the page doesn't trigger the "update needed" page.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/39108
+++ b/changelog/unreleased/39108
@@ -1,0 +1,15 @@
+Bugfix: Update appinfo cache only if the app version is newer
+
+Previously, in case there were multiple copies of the same app with different
+versions, the information being cached was the latest one found based on the
+locations defined in the config.php file, which might not be the one from the
+latest app version. This might be a problem in some scenarios specially checking
+the version of the app. Note that the code used was the one from the latest app
+version found.
+
+Now, the information cached is always from the latest version found. In the
+weird case that both versions are the same, the information from the first one
+will be kept. This shouldn't be a problem because the information is expected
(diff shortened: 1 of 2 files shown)


jvillafanez:
There are no clear steps to reproduce the original problem, so it's difficult to foresee if this PR will fix the issue.
The provided steps in this PR are convoluted and unlikely to happen for common users. This means that it might not fix the problem affecting those users.
This PR fixes the issue caused by following the provided steps, so it's still worthy to merge.

JammingBen (dismissed):
Plus changelog, then LGTM. Hopefully it fixes the issue. But it's a good thing either way.

VicDeo (approved):
Ready to go

jnweiger:
cannot reproduce the issue now, but the changes look indeed very promising. Thanks for the hard digging!

PR #39126, opened 2021-08-20
Feature/version metadata

pushnov-i:
The file version author is now being saved in an appropriate timestamp-prefixed JSON file and displayed in the version list on PROPFIND request. Rename, copy and deletion are take into account.

Enable with config-flag: 'file_storage.save_version_author' => true, in config.php

### DIFF
--- a/apps/dav/lib/Meta/MetaFile.php
+++ b/apps/dav/lib/Meta/MetaFile.php
@@ -131,7 +131,7 @@ class MetaFile extends File implements ICopySource, IFileNode, IProvidesAddition
 	/**
 	 * @return string
 	 */
-	public function getVersionAuthor() {
+	public function getVersionAuthor() : string {
 		if ($this->file instanceof IProvidesVersionAuthor) {
 			return $this->file->getCreatedBy() ?: $this->file->getEditedBy();
 		}
@@ -141,12 +141,12 @@ class MetaFile extends File implements ICopySource, IFileNode, IProvidesAddition
 	/**
 	 * @return string
 	 */
-	public function getVersionAuthorName() {
+	public function getVersionAuthorName() : string {
 		if ($this->file instanceof IProvidesVersionAuthor) {
 			$uid =  $this->file->getCreatedBy() ?: $this->file->getEditedBy();
 			$manager = \OC::$server->getUserManager();
 			$user = $manager->get($uid);
-			return $user !== null ? $user->getDisplayName() : "";
+			return $user !== null ? $user->getDisplayName() : '';
 		}
 		return '';
 	}
--- a/apps/dav/lib/Meta/MetaFile.php
+++ b/apps/dav/lib/Meta/MetaFile.php
@@ -133,7 +133,7 @@ class MetaFile extends File implements ICopySource, IFileNode, IProvidesAddition
 	 */
 	public function getVersionAuthor() : string {
 		if ($this->file instanceof IProvidesVersionAuthor) {
-			return $this->file->getCreatedBy() ?: $this->file->getEditedBy();
+			return $this->file->getEditedBy();
 		}
 		return '';
 	}
@@ -143,7 +143,7 @@ class MetaFile extends File implements ICopySource, IFileNode, IProvidesAddition
 	 */
 	public function getVersionAuthorName() : string {
 		if ($this->file instanceof IProvidesVersionAuthor) {
-			$uid =  $this->file->getCreatedBy() ?: $this->file->getEditedBy();
+			$uid =  $this->file->getEditedBy();
 			$manager = \OC::$server->getUserManager();
 			$user = $manager->get($uid);
 			return $user !== null ? $user->getDisplayName() : '';
--- a/apps/dav/lib/Meta/MetaFile.php
+++ b/apps/dav/lib/Meta/MetaFile.php
@@ -25,6 +25,7 @@ use OC\Files\Meta\MetaFileVersionNode;
 use OCA\DAV\Files\ICopySource;
 use OCA\DAV\Files\IProvidesAdditionalHeaders;
 use OCA\DAV\Files\IFileNode;
+use OCP\Files\IProvidesVersionAuthor;
 use OCP\Files\Node;
 use Sabre\DAV\File;
 
@@ -125,5 +126,28 @@ class MetaFile extends File implements ICopySource, IFileNode, IProvidesAddition
 	 */
 	public function getNode() {
 		return $this->file;
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getVersionAuthor() {
+		if ($this->file instanceof IProvidesVersionAuthor) {
+			return $this->file->getCreatedBy() ?: $this->file->getEditedBy();
+		}
+		return '';
+	}
+
+	/**
+	 * @return string
+	 */
+	public function getVersionAuthorName() {
+		if ($this->file instanceof IProvidesVersionAuthor) {
+			$uid =  $this->file->getCreatedBy() ?: $this->file->getEditedBy();
+			$manager = \OC::$server->getUserManager();
+			$user = $manager->get($uid);
+			return $user !== null ? $user->getDisplayName() : "";
+		}
+		return '';
 	}
 }
--- a/apps/dav/lib/Meta/MetaPlugin.php
+++ b/apps/dav/lib/Meta/MetaPlugin.php
@@ -32,6 +32,7 @@ class MetaPlugin extends ServerPlugin {
 	// namespace
 	public const NS_OWNCLOUD = 'http://owncloud.org/ns';
 	public const PATH_FOR_FILEID_PROPERTYNAME = '{http://owncloud.org/ns}meta-path-for-user';
+	const VERSION_EDITED_BY_PROPERTYNAME = '{http://owncloud.org/ns}metaversioneditedby';
 
 	/**
 	 * Reference to main server object
@@ -96,6 +97,11 @@ class MetaPlugin extends ServerPlugin {
 				}
 				$file = \current($files);
 				return $baseFolder->getRelativePath($file->getPath());
+			});
+		} elseif ($node instanceof MetaFile){
+			$propFind->handle(self::VERSION_EDITED_BY_PROPERTYNAME, function () use ($node) {
+				// FIXME: get from storage of the $node required data
+				return "test1@owncloud.com";
 			});
 		}
 	}
--- a/apps/dav/lib/Meta/MetaPlugin.php
+++ b/apps/dav/lib/Meta/MetaPlugin.php
@@ -32,8 +32,9 @@ class MetaPlugin extends ServerPlugin {
 	// namespace
 	public const NS_OWNCLOUD = 'http://owncloud.org/ns';
 	public const PATH_FOR_FILEID_PROPERTYNAME = '{http://owncloud.org/ns}meta-path-for-user';
-	const VERSION_EDITED_BY_PROPERTYNAME = '{http://owncloud.org/ns}metaversioneditedby';
 
+	public const VERSION_EDITED_BY_PROPERTYNAME = '{http://owncloud.org/ns}meta-version-edited-by';
+	public const VERSION_EDITED_BY_PROPERTYNAME_NAME = '{http://owncloud.org/ns}meta-version-edited-by-name';
 	/**
 	 * Reference to main server object
 	 *
@@ -98,10 +99,12 @@ class MetaPlugin extends ServerPlugin {
 				$file = \current($files);
 				return $baseFolder->getRelativePath($file->getPath());
 			});
-		} elseif ($node instanceof MetaFile){
+		} elseif ($node instanceof MetaFile) {
 			$propFind->handle(self::VERSION_EDITED_BY_PROPERTYNAME, function () use ($node) {
-				// FIXME: get from storage of the $node required data
-				return "test1@owncloud.com";
+				return $node->getVersionAuthor();
+			});
+			$propFind->handle(self::VERSION_EDITED_BY_PROPERTYNAME_NAME, function () use ($node) {
+				return $node->getVersionAuthorName();
 			});
 		}
 	}
--- a/apps/files_trashbin/lib/Trashbin.php
+++ b/apps/files_trashbin/lib/Trashbin.php
@@ -400,19 +400,35 @@ class Trashbin {
 			$rootView = new View('/');
 
 			if ($rootView->is_dir($owner . '/files_versions/' . $ownerPath)) {
+				$metadataFileExists = $rootView->file_exists($owner . '/files_versions/' . $ownerPath . '.json');
+
 				if ($owner !== $user || $forceCopy) {
 					self::copy_recursive($owner . '/files_versions/' . $ownerPath, $owner . '/files_trashbin/versions/' . \basename($ownerPath) . '.d' . $timestamp, $rootView);
+					if ($metadataFileExists) {
+						self::copy_recursive($owner . '/files_versions/' . $ownerPath . '.json', $owner . '/files_trashbin/versions/' . \basename($ownerPath) . '.json' . '.d' . $timestamp, $rootView);
+					}
 				}
 				if (!$forceCopy) {
 					self::move($rootView, $owner . '/files_versions/' . $ownerPath, $user . '/files_trashbin/versions/' . $filename . '.d' . $timestamp);
+					if ($metadataFileExists) {
+						self::move($rootView, $owner . '/files_versions/' . $ownerPath . '.json', $user . '/files_trashbin/versions/' . $filename . '.json' . '.d' . $timestamp);
+					}
 				}
 			} elseif ($versions = \OCA\Files_Versions\Storage::getVersions($owner, $ownerPath)) {
 				foreach ($versions as $v) {
+					$metaVersionExists = $rootView->file_exists($owner . '/files_versions' . $v['path'] . '.v' . $v['version'] . '.json');
+
 					if ($owner !== $user || $forceCopy) {
 						self::copy($rootView, $owner . '/files_versions' . $v['path'] . '.v' . $v['version'], $owner . '/files_trashbin/versions/' . $v['name'] . '.v' . $v['version'] . '.d' . $timestamp);
+						if ($metaVersionExists) {
+							self::move($rootView, $owner . '/files_versions' . $v['path'] . '.v' . $v['version'] . '.json', $owner . '/files_trashbin/versions/' . $filename . '.v' . $v['version'] . '.json' . '.d' . $timestamp);
+						}
 					}
 					if (!$forceCopy) {
 						self::move($rootView, $owner . '/files_versions' . $v['path'] . '.v' . $v['version'], $user . '/files_trashbin/versions/' . $filename . '.v' . $v['version'] . '.d' . $timestamp);
+						if ($metaVersionExists) {
+							self::move($rootView, $owner . '/files_versions' . $v['path'] . '.v' . $v['version'] . '.json', $user . '/files_trashbin/versions/' . $filename . '.v' . $v['version'] . '.json' . '.d' . $timestamp);
+						}
 					}
 				}
 			}
--- a/apps/files_versions/js/versioncollection.js
+++ b/apps/files_versions/js/versioncollection.js
@@ -11,11 +11,26 @@
 /* global moment */
 
 (function() {
+
+	_.extend(OC.Files.Client, {
+		PROPERTY_FILEID:	'{' + OC.Files.Client.NS_OWNCLOUD + '}id',
+		PROPERTY_VERSION_EDITED_BY:	'{' + OC.Files.Client.NS_OWNCLOUD + '}metaversioneditedby',
+	});
+
 	/**
 	 * @memberof OCA.Versions
 	 */
 	var VersionCollection = OC.Backbone.Collection.extend({
 		sync: OC.Backbone.davSync,
+
+		davProperties: {
+			'metaversioneditedby':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY,
+			'id':	OC.Files.Client.PROPERTY_FILEID,
+			'getlastmodified': OC.Files.Client.PROPERTY_GETLASTMODIFIED,
+			'getcontentlength': OC.Files.Client.PROPERTY_GETCONTENTLENGTH,
+			'resourcetype': OC.Files.Client.PROPERTY_RESOURCETYPE,
+			'getcontenttype': OC.Files.Client.PROPERTY_GETCONTENTTYPE,
+		},
 
 		model: OCA.Versions.VersionModel,
 
@@ -46,9 +61,10 @@
 					id: revision,
 					name: revision,
 					fullPath: fullPath,
-					timestamp: moment(new Date(version['{DAV:}getlastmodified'])).format('X'),
-					size: version['{DAV:}getcontentlength'],
-					mimetype: version['{DAV:}getcontenttype'],
+					timestamp: moment(new Date(version.getlastmodified)).format('X'),
+					size: version.getcontentlength,
+					mimetype: version.getcontenttype,
+					editedBy: version.metaversioneditedby,
 					fileId: fileId
 				};
 			});
--- a/apps/files_versions/js/versioncollection.js
+++ b/apps/files_versions/js/versioncollection.js
@@ -14,7 +14,8 @@
 
 	_.extend(OC.Files.Client, {
 		PROPERTY_FILEID:	'{' + OC.Files.Client.NS_OWNCLOUD + '}id',
-		PROPERTY_VERSION_EDITED_BY:	'{' + OC.Files.Client.NS_OWNCLOUD + '}metaversioneditedby',
+		PROPERTY_VERSION_EDITED_BY:	'{' + OC.Files.Client.NS_OWNCLOUD + '}meta-version-edited-by',
+    	PROPERTY_VERSION_EDITED_BY_NAME:	'{' + OC.Files.Client.NS_OWNCLOUD + '}meta-version-edited-by-name',
 	});
 
 	/**
@@ -24,7 +25,8 @@
 		sync: OC.Backbone.davSync,
 
 		davProperties: {
-			'metaversioneditedby':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY,
+			'meta-version-edited-by':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY,
+      'meta-version-edited-by-name':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY_NAME,
 			'id':	OC.Files.Client.PROPERTY_FILEID,
 			'getlastmodified': OC.Files.Client.PROPERTY_GETLASTMODIFIED,
 			'getcontentlength': OC.Files.Client.PROPERTY_GETCONTENTLENGTH,
@@ -62,9 +64,11 @@
 					name: revision,
 					fullPath: fullPath,
 					timestamp: moment(new Date(version.getlastmodified)).format('X'),
+					relativeTimestamp:  moment(new Date(version.getlastmodified)).fromNow(),
 					size: version.getcontentlength,
 					mimetype: version.getcontenttype,
-					editedBy: version.metaversioneditedby,
+					editedBy: version['meta-version-edited-by'],
+          			editedByName: version['meta-version-edited-by-name'],
 					fileId: fileId
 				};
 			});
--- a/apps/files_versions/js/versioncollection.js
+++ b/apps/files_versions/js/versioncollection.js
@@ -26,7 +26,7 @@
 
 		davProperties: {
 			'meta-version-edited-by':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY,
-      'meta-version-edited-by-name':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY_NAME,
+      		'meta-version-edited-by-name':	OC.Files.Client.PROPERTY_VERSION_EDITED_BY_NAME,
 			'id':	OC.Files.Client.PROPERTY_FILEID,
 			'getlastmodified': OC.Files.Client.PROPERTY_GETLASTMODIFIED,
 			'getcontentlength': OC.Files.Client.PROPERTY_GETCONTENTLENGTH,
--- a/apps/files_versions/js/versionstabview.js
+++ b/apps/files_versions/js/versionstabview.js
@@ -39,6 +39,7 @@
 		'{{#hasDetails}}' +
 		'<div class="version-details">' +
 		'<span class="size has-tooltip" title="{{altSize}}">{{humanReadableSize}}</span>' +
+		'<span class="size has-tooltip" title="{{editedBy}}">{{editedBy}}</span>' +
 		'</div>' +
 		'{{/hasDetails}}' +
 		'</div>' +
@@ -213,7 +214,8 @@
 				revertIconUrl: OC.imagePath('core', 'actions/history'),
 				previewUrl: getPreviewUrl(version),
 				revertLabel: t('files_versions', 'Restore'),
-				canRevert: (this.collection.getFileInfo().get('permissions') & OC.PERMISSION_UPDATE) !== 0
+				canRevert: (this.collection.getFileInfo().get('permissions') & OC.PERMISSION_UPDATE) !== 0,
+				editedBy: version.has('editedBy'),
 			}, version.attributes);
 		},
 
--- a/apps/files_versions/js/versionstabview.js
+++ b/apps/files_versions/js/versionstabview.js
@@ -216,7 +216,7 @@
 				revertLabel: t('files_versions', 'Restore'),
 				canRevert: (this.collection.getFileInfo().get('permissions') & OC.PERMISSION_UPDATE) !== 0,
 				editedBy: version.has('editedBy'),
-        editedByName: version.has('editedByName')
+        		editedByName: version.has('editedByName')
 			}, version.attributes);
 		},
 
--- a/apps/files_versions/js/versionstabview.js
+++ b/apps/files_versions/js/versionstabview.js
@@ -39,7 +39,7 @@
 		'{{#hasDetails}}' +
 		'<div class="version-details">' +
 		'<span class="size has-tooltip" title="{{altSize}}">{{humanReadableSize}}</span>' +
-		'<span class="size has-tooltip" title="{{editedBy}}">{{editedBy}}</span>' +
+		'<span title="{{editedBy}}">{{editedByName}}</span>' +
 		'</div>' +
 		'{{/hasDetails}}' +
 		'</div>' +
@@ -216,6 +216,7 @@
 				revertLabel: t('files_versions', 'Restore'),
 				canRevert: (this.collection.getFileInfo().get('permissions') & OC.PERMISSION_UPDATE) !== 0,
 				editedBy: version.has('editedBy'),
+        editedByName: version.has('editedByName')
 			}, version.attributes);
 		},
 
--- a/apps/files_versions/lib/Hooks.php
+++ b/apps/files_versions/lib/Hooks.php
@@ -36,6 +36,7 @@ class Hooks {
 	public static function connectHooks() {
 		// Listen to write signals
 		\OCP\Util::connectHook('OC_Filesystem', 'write', 'OCA\Files_Versions\Hooks', 'write_hook');
+
 		// Listen to delete and rename signals
 		\OCP\Util::connectHook('OC_Filesystem', 'post_delete', 'OCA\Files_Versions\Hooks', 'remove_hook');
 		\OCP\Util::connectHook('OC_Filesystem', 'delete', 'OCA\Files_Versions\Hooks', 'pre_remove_hook');
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -208,8 +208,7 @@ class Storage {
 				if ($config->getSystemValue('file_storage.save_version_author', false) === true) {
 					$user = \OC::$server->getUserSession()->getUser();
 					if ($user !== null && !$users_view->file_exists($versionFileName . '.json')) {
-						$versions = self::getVersions($uid, $filename);
-						$metaDataKey =  \sizeof($versions) === 1 ? Constants::CREATED_BY_USER_METADATA : MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME;
+						$metaDataKey =  MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME;
 						$metadata = [$metaDataKey => $user->getUID()];
 						$metadataJsonObject = \json_encode($metadata);
 						$users_view->file_put_contents($versionFileName . '.json', $metadataJsonObject);
@@ -366,8 +365,7 @@ class Storage {
 
 			if ($metaDataEnabled === true) {
 				$metaTargetPath = $version . '.json';
-				$versions = self::getVersions($uid, $filename);
-				$metaDataKey = \sizeof($versions) == 1 ? Constants::CREATED_BY_USER_METADATA : MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME;
+				$metaDataKey = MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME;
 				$metadataJsonObject = \json_encode([$metaDataKey => $uid]);
 				$users_view->file_put_contents($metaTargetPath, $metadataJsonObject);
 			}
@@ -506,9 +504,7 @@ class Storage {
 						if ($view->file_exists($jsonMetadataFile)) {
 							$metaDataFileContents = $view->file_get_contents($jsonMetadataFile);
 							if ($decoded = \json_decode($metaDataFileContents, true)) {
-								if (isset($decoded[Constants::CREATED_BY_USER_METADATA])) {
-									$versions[$key]['created_by'] = $decoded[Constants::CREATED_BY_USER_METADATA];
-								} elseif ($decoded[MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME] !== null) {
+								if (isset($decoded[MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME])) {
 									$versions[$key]['edited_by'] = $decoded[MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME];
 								}
 							}
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -238,6 +238,10 @@ class Storage {
 	 */
 	protected static function deleteVersion($view, $path) {
 		$view->unlink($path);
+		/**
+		 * @var \OC\Files\Storage\Storage $storage
+		 * @var string $internalPath
+		 */
 		list($storage, $internalPath) = $view->resolvePath($path);
 		$cache = $storage->getCache($internalPath);
 		$cache->remove($internalPath);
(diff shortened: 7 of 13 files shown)


IljaN:
- You can fix codestyle by running: ```make test-php-style-fix```
- You can run stan and phan (static-analyzers) via: ``` make test-php-phpstan``` and ``` make test-php-phan ```

JammingBen (commented):
Added some comments 👍 

Some indentations seem to be off, `make test-php-style-fix` should fix this.

apps/dav/lib/Meta/MetaFile.php
```diff
@@ -126,4 +126,8 @@ public function getContentDispositionFileName() {
 	public function getNode() {
 		return $this->file;
 	}
+
+	public function getUsername() {
```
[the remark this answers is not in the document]
pushnov-i: done https://github.com/owncloud/core/pull/39126/commits/94729f8101878b2cdd9d470d75575541cb75b932

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -402,17 +402,21 @@ private static function retainVersions($filename, $owner, $ownerPath, $timestamp
 			if ($rootView->is_dir($owner . '/files_versions/' . $ownerPath)) {
 				if ($owner !== $user || $forceCopy) {
 					self::copy_recursive($owner . '/files_versions/' . $ownerPath, $owner . '/files_trashbin/versions/' . \basename($ownerPath) . '.d' . $timestamp, $rootView);
+					self::copy_recursive($owner . '/files_versions/' . $ownerPath . '.json', $owner . '/files_trashbin/versions/' . \basename($ownerPath) . 'json' . '.d' . $timestamp, $rootView);
```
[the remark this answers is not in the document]
pushnov-i: good catch, fixed

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -402,17 +402,21 @@ private static function retainVersions($filename, $owner, $ownerPath, $timestamp
...
 				}
 				if (!$forceCopy) {
 					self::move($rootView, $owner . '/files_versions/' . $ownerPath, $user . '/files_trashbin/versions/' . $filename . '.d' . $timestamp);
+					self::move($rootView, $owner . '/files_versions/' . $ownerPath . 'json', $user . '/files_trashbin/versions/' . $filename . 'json' . '.d' . $timestamp);
```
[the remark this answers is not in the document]
pushnov-i: fixed

apps/files_versions/lib/Storage.php
```diff
@@ -66,22 +67,26 @@ class Storage {
...
+		6 => ['intervalEndsAfter' => -1, 'step' => 604800],
 	];
 
 	/** @var \OCA\Files_Versions\AppInfo\Application */
 	private static $application;
 
+	/**
+	 * @var IUserSession $userSession
+	 */
+
```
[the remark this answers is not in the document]
pushnov-i: fixed

apps/dav/lib/Meta/MetaFile.php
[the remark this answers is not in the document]
pushnov-i: thanks, fixed that

> * You can fix codestyle by running: `make test-php-style-fix`
> * You can run stan and phan (static-analyzers) via: ` make test-php-phpstan` and `make test-php-phan`

code style is fixed by test-php-style-fix

apps/files_trashbin/lib/Trashbin.php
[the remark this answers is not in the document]
prsnbrg: I think, there is still a missing dot bevore `json`.

ownclouders:
:boom: Acceptance tests pipeline <strong>webUISharingExt1-latest-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/31929/162/1

apps/files_trashbin/lib/Trashbin.php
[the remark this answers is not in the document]
pushnov-i: fixed

JammingBen:
I'm not sure if we should add `getUsername()` to the overall file interface. This method is (only?) needed in the `MetaFileVersionNode` class. Look for example how this it done for `getContentDispositionFileName()` [1].

You could implement a dedicated interface for the `getUsername()` method, similar to `IProvidesAdditionalHeaders`. This has then to be implemented by your `MetaFileVersionNode` class [2]. Your `getUsername()` method in `MetaFile` would look something like:

```
public function getUsername() {
	if ($this->file instanceof \OCP\Files\IVersionAuthor) {
		return $this->file->getUsername();
	}
	return '';
}
```

[1] https://github.com/owncloud/core/blob/feature/version-metadata/apps/dav/lib/Meta/MetaFile.php#L116
[2] https://github.com/owncloud/core/blob/feature/version-metadata/lib/private/Files/Meta/MetaFileVersionNode.php#L42

About the failing JS test: As the UI now shows the author in the versions tab, you have to adjust the expected result in the tests: https://github.com/owncloud/core/blob/feature/version-metadata/apps/files_versions/tests/js/versionstabviewSpec.js#L71

On a side note: I tried it on my local instance, but I only see a `{` where the author should be:

image

Feel free to ask if you need any help with that 🙂

mrow4a:
@hodyroff @pushnov-i what about making it optional in admin settings ? Just not to release new feature and reduce blast radius in case of bugs? It touches core functionality of sync/share. 

I am bit worried what happens when one of that version “copy” commands fails in the middle. How consistent is the state? 

What about expiring old versions ? Is this also handled (if relevant)? 

What about acceptance tests here? 

Unit tests covering major classes ?

Xxxxx1111 (approved):

lib/public/Files/IProvidesVersionAuthor.php
```diff
@@ -0,0 +1,38 @@
...
+ * This interface provides version author retrieval for file version
+ *
+ * @package OCP\Files
+ * @since 10.0.9
+ */
+interface IProvidesVersionAuthor {
+	/**
+	 * Returns the author's username
+	 * @return string
+	 * @since 10.0.9
```
phil-davis, line 35:
```suggestion
	 * @since 10.9.0
```

lib/public/Files/IProvidesVersionAuthor.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace OCP\Files;
+
+/**
+ * Interface IProvidesVersionAuthor
+ * This interface provides version author retrieval for file version
+ *
+ * @package OCP\Files
+ * @since 10.0.9
```
phil-davis, line 29:
```suggestion
 * @since 10.9.0
```

pushnov-i:
@phil-davis thank you, corrected the changes in the version for the interface.
@JammingBen I implemented everything via interface and adjusted the tests, the unit tests and js tests pass now new quick action feature prevent the drone from getting a successful build

@mrow4a not sure if this is related to the expiration, only `retainVersions` and `getVersion` methods from storage are affected, tests have been adjusted. regarding the optional feature, how it should appear on the frontend?

IljaN:
There are 6 version-related unit-tests failing https://drone.owncloud.com/owncloud/core/31947/12/9 
Click "Show all Lines" and scroll down to line 249. You can ignore the skipped tests.

pushnov-i:
@IljaN thank you, I fixed one bug and adjusted the unit tests. it seems all is good now

apps/files_trashbin/tests/StorageTest.php
```diff
@@ -352,7 +352,7 @@ public function testDeleteVersionsOfFolder() {
 
 		// check if versions are in trashbin
 		$results = $this->rootView->getDirectoryContent($this->user . '/files_trashbin/versions/' . $name . '/');
-		$this->assertCount(1, $results);
+		$this->assertCount(2, $results);
```
IljaN, line 355: Could you check additionally for an explicit expected result value in the second element of $result here? Only if you don't need to jump trough too much hoops.

IljaN, line 355: Could you check additionally for an explicit expected result value in the second element of $result here? Only if you don't need to jump trough too much hoops.

IljaN: Also in all following assertions.

apps/dav/lib/Meta/MetaPlugin.php
```diff
@@ -97,6 +99,10 @@ public function handleGetProperties(PropFind $propFind, INode $node) {
 				$file = \current($files);
 				return $baseFolder->getRelativePath($file->getPath());
 			});
+		} elseif ($node instanceof MetaFile) {
+			$propFind->handle(self::VERSION_EDITED_BY_PROPERTYNAME, function () use ($node) {
+				return $node->getUsername() ?? "Not available";
```
IljaN, line 104: Will "Not available" be visible in the WebUI? If yes this string should be translated. You might need to inject https://github.com/owncloud/core/blob/3bff5c2aca12499972f285e0cf2ea467a473c6b5/lib/public/IL10N.php#L44 in to your component alternatively if this is an issue you can globally get the service from the registry like so ```		\OC::$server->getL10N('core')->t('Not Available');```

IljaN:
@pushnov-i Looks good now, good work! :man_dancing:  Just a few small nitpicks left.

apps/files_trashbin/tests/StorageTest.php
pushnov-i: @IljaN I did a check yesterday with xdebug and just checked once again now - for this specific case the trashbin dir contains two files first one is the file version with the timestamp and the second is the JSON file with metadata:

Снимок экрана 2021-08-24 в 12 13 46

IljaN: So if the second file was introduced trough your change we should maybe explicitly check for it.

IljaN: So if the second file was introduced trough your change we should maybe explicitly check for it.

apps/dav/lib/Meta/MetaPlugin.php
pushnov-i: @IljaN, you are right, and this placeholder is redundant since if no metadata is available, the empty string will be shown instead of the author's username. I will take translation wrapper into account though

IljaN: :+1:

apps/files_trashbin/tests/StorageTest.php
pushnov-i: Do you imply a completely separate test for the trash bin, or just an additional test if the second file is a JSON, has a valid name, and contains metadata within `testDeleteVersionsOfFolder`?

IljaN: >  just an additional test if the second file is a JSON, has a valid name, and contains metadata within testDeleteVersionsOfFolder?

This ^

IljaN: But the more the better ;)

mrow4a:
@pushnov-i is there a unit/integration test covering backwards compatibility? How code/ui is behaving if there is no metadata-file ?

IljaN:
> @mrow4a  wrote:
> Tested with encryption?

This is also a point we should take in to account as the feature will be used first in an ownCloud instnace with encryption.

phil-davis:
@pushnov-i the CLA assistant still says that you have not "signed" the CLA. Please press the button to do that.

IljaN:
@mrow4a I just checked with @phil-davis. If we merge this all acceptance-tests will be run with encryption (master-key and user-key) as nightly-job. Results will be posted by drone in #builds in chat.

apps/files_trashbin/tests/StorageTest.php
```diff
@@ -672,7 +672,7 @@ public function testSingleStorageDeleteFileFail() {
 		$this->assertTrue($storage->file_exists('files/test.txt'));
 		$this->assertTrue($cache->inCache('files/test.txt'));
 
-		// file should not be in the trashbin
+		// file should not be in the  trashbin
```
phil-davis, line 675: random space

phil-davis:
I don't see any changelog. Did I miss it?

apps/files_trashbin/tests/StorageTest.php
pushnov-i: fixed

pushnov-i:
@phil-davis added changelog, fixed space.
@IljaN I did not check the encryption since this requires installing an additional plugin, but if only the content of the files will be encrypted, this feature should not affect the workflow as long as the structure of the files within storage won't be changed. I did not manage to add additional tests yet... @mrow4a, if a JSON file is missing, no author will be shown on the UI side. On the code's side, within the storage, you can see that there are conditional checks everywhere, so if the JSON file is present, it will be considered, otherwise ignored.

phil-davis:
@pushnov-i please click the link to sign the CLA at https://github.com/owncloud/core/pull/39126#issuecomment-902547524

Does it not work for you? Or?

pushnov-i:
> @pushnov-i please click the link to sign the CLA at [#39126 (comment)](https://github.com/owncloud/core/pull/39126#issuecomment-902547524)
> 
> Does it not work for you? Or?

@phil-davis CLA requires me to change the author of one merge commit since it does not recognize it in github, working on it

pushnov-i:
> @pushnov-i please click the link to sign the CLA at [#39126 (comment)](https://github.com/owncloud/core/pull/39126#issuecomment-902547524)
> 
> Does it not work for you? Or?

done

apps/dav/lib/Meta/MetaFile.php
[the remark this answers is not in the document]
DeepDiver1975: getUserName() if one asks me ....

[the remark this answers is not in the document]
DeepDiver1975: instead of phpdoc we could use real php return values ...

apps/dav/lib/Meta/MetaFile.php
```diff
@@ -126,4 +126,14 @@ public function getContentDispositionFileName() {
 	public function getNode() {
 		return $this->file;
 	}
+
+	/**
+	 * @return string
+	 */
+	public function getUsername() {
+		if ($this->file instanceof \OCP\Files\IProvidesVersionAuthor) {
```
DeepDiver1975, line 134: import IProvidesVersionAuthor

apps/dav/lib/Meta/MetaPlugin.php
```diff
@@ -33,6 +33,8 @@ class MetaPlugin extends ServerPlugin {
 	public const NS_OWNCLOUD = 'http://owncloud.org/ns';
 	public const PATH_FOR_FILEID_PROPERTYNAME = '{http://owncloud.org/ns}meta-path-for-user';
 
+	public const VERSION_EDITED_BY_PROPERTYNAME = '{http://owncloud.org/ns}metaversioneditedby';
```
DeepDiver1975, line 36: meta-version-edited-by

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -400,19 +400,37 @@ private static function retainVersions($filename, $owner, $ownerPath, $timestamp
...
 				}
 				if (!$forceCopy) {
 					self::move($rootView, $owner . '/files_versions/' . $ownerPath, $user . '/files_trashbin/versions/' . $filename . '.d' . $timestamp);
+					if ($metadataFileExists) {
+						self::move($rootView, $metadataFilename, $user . '/files_trashbin/versions/' . $filename . '.json' . '.d' . $timestamp);
+					}
 				}
 			} elseif ($versions = \OCA\Files_Versions\Storage::getVersions($owner, $ownerPath)) {
 				foreach ($versions as $v) {
+					$metaVersionFile = $owner . '/files_versions' . $v['path'] . '.v' . $v['version'] . '.json';
```
DeepDiver1975, line 420: I always find string interpolation more readable ....

apps/files_versions/js/versioncollection.js
```diff
@@ -11,12 +11,27 @@
 /* global moment */
 
 (function() {
+
+	_.extend(OC.Files.Client, {
+		PROPERTY_FILEID:	'{' + OC.Files.Client.NS_OWNCLOUD + '}id',
+		PROPERTY_VERSION_EDITED_BY:	'{' + OC.Files.Client.NS_OWNCLOUD + '}metaversioneditedby',
```
DeepDiver1975, line 17: meta-version-edited-by

lib/private/Share/Constants.php
```diff
@@ -61,4 +61,6 @@ class Constants {
 	protected static $backends = [];
 	protected static $backendTypes = [];
 	protected static $isResharingAllowed;
+
+	public const CREATED_BY_USER_METADATA = "{http://owncloud.org/ns}createdbyuser";
```
DeepDiver1975, line 65: created-by-user

lib/public/Files/IProvidesVersionAuthor.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/**
+ * @author Illia Pushnov <illia.pushnov@gmail.com>
+ *
+ * @copyright Copyright (c) 2017, ownCloud GmbH
```
DeepDiver1975, line 5: 2021 - please check all other file headers - THX

DeepDiver1975 (changes requested):
I am missing explicit uni tests (later acceptance tests as well) where the user information is stored and retrieved. THX

mrow4a:
same as @DeepDiver1975 hard to judge if this works without manual testdrive if there are no functional unit/acceptance tests. 

e.g. what is displayed in versions tab when it has only 1 version (new file). What about shares and reshares?

apps/files_versions/lib/Storage.php
```diff
@@ -192,19 +193,29 @@ public static function store($filename) {
...
-				$users_view->getFileInfo("files_versions/$filename.v$mtime");
+				$users_view->getFileInfo($versionFileName);
 				// update checksum of the version
-				$users_view->putFileInfo("files_versions/$filename.v$mtime", [
+				$users_view->putFileInfo($versionFileName, [
 					'checksum' => $sourceFileInfo->getChecksum(),
 				]);
+
+				$user = \OC::$server->getUserSession()->getUser();
+				if ($user !== null) {
```
mrow4a, line 207: @pushnov-i regarding making this feature configurable, this is good example how it could be done if you decide on doing so -> https://github.com/owncloud/core/pull/39130

mrow4a: to disable this feature, looking at existing code logic, it is enough just not to write the metadata.

lib/private/Files/Meta/MetaFileVersionNode.php
```diff
@@ -54,6 +55,11 @@ class MetaFileVersionNode extends AbstractFile implements IPreviewNode, IProvide
 	/** @var array */
 	private $versionInfo;
 
+	/**
+	 * @var string
+   **/
```
mrow4a, line 60: typo

phil-davis:
I have tried this out manually:
- have some versions existing before checking out this branch
- view an existing file with versions, the list in the UI does not e. the usernames
- make some edits by different users - the new versions do display the usernames
- create a new file and edit it - all the versions have usernames

I didn't look closely at the usernames. So I need to check again a scenario like:
- Alice creates file.txt
- Brian edits file.txt
- Carol edits file.txt

There should be 2 old versions. The oldest version was created by Alice. The next version was created by Brian. Carol is the user who created the current file. I just have a suspicion that the code may not have a way to know that - the oldest version is created when Brian does his edit, the code would have to know somewhere that Alice is the original creator...

Also need to check what happens when one of the usernames in the version history no longer exists. Hopefully the code displays the usernames from the JSON file without anything bad happening.

What do we want to do about display names? If username `alice` has set their display name to "Alice Hansen" then do we show the display name?

pmaier1:
> display full name instead of a username

Please do it like it's done in the sharing dialog (e.g., autocomplete or share recipient list). There is an option to add the user id or the email address to the display name. This should be respected.

pmaier1:
Please let me know when this is ready for a demo. I'll invite then.

mmattel:
Docs relevant because of a change in `config/config.sample.php`
Please file a docs issue
@EParzefall

ownclouders:

https://drone.owncloud.com/owncloud/core/32300/124/1

phil-davis:
Note: we are having trouble with an intermittent webUI test. Issue #39213 

PR #39219 will kip it. That will be merged when CI finishes. After you can rebase, and that annoying fail will be gone.

phil-davis:
https://github.com/owncloud/core/tree/feature/version-metadata
"This branch is 49 commits ahead, 97 commits behind master."

There has been a lot happening in core master recently. This branch is running quite "old" CI now. It would be good to rebase some time. (And to squash/cleanup the 49 commits)

DeepDiver1975 (changes requested):

IljaN:
Checked this branch out for testing, Do I need to enable something?
noversion

JammingBen:
> Yes, it's false by default as it takes extra space for each file version and should be disabled by default as team has mentioned before

Okay 👍  In this case I would rename the config to something like `file_storage.save_version_author` to make it more clear. Versions in general are always saved.

On a side note: I noticed that some recent changed were overwritten during the merge process, e.g. in `config.sample.php`. Let's not forget to check that before merging eventually!

pushnov-i:
> > Yes, it's false by default as it takes extra space for each file version and should be disabled by default as team has mentioned before
> 
> Okay 👍 In this case I would rename the config to something like `file_storage.save_version_author` to make it more clear. Versions in general are always saved.
> 
> On a side note: I noticed that some recent changed were overwritten during the merge process, e.g. in `config.sample.php`. Let's not forget to check that before merging eventually!

done

pushnov-i:
> We did a short review:
> 
> 1.) The json files in the storage should store the uid and use that to lookup the display name. With the current implementation if the display name is changed afterwards the list becomes inconsistent.
> 
> In the screenshot bellow the display name of user1 was changed to "Hans Müller" versions_list
> 
> 2.) If you hover over the display name in the versions list the uid should be displayed in the tooltip in case there are two persons with the same display-name.
> 
> 3.) The expire versions job should also clean the corresponding json files:
> 
> https://github.com/owncloud/core/blob/ad8eb74bd6322019cce933b3f7a99857c7f06f22/apps/files_versions/lib/BackgroundJob/ExpireVersions.php#L32
> 
> We discovered this because the job ran during our review 🍀
> 
> 4.) @JammingBen mentioned that we should check if the rebase was done correctly. From what I see everything should be fine though.
> 
> 5.) We had an issue where we saw the current version also being in the list, but could not reproduce it reliably. Maybe related to 3.
> 
> /cc @pmaier1

Fixed UID save, display name, tooltip, tested expired version deletion, and rebased the branch on master. Regarding 5) as far as I understand the initial version of the file should be saved as an initial version as well? If so, everything seems to be working as expected.

phil-davis:
code style checks fail: https://drone.owncloud.com/owncloud/core/33484/3/6
Please fix those, and then the rest of CI will run and we can see what tests pass.

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33487/85/1

tests/acceptance/features/apiVersions/fileVersions.feature
```diff
@@ -469,3 +469,37 @@ Feature: dav-versions
...
+      Given the administrator has enabled file version storage feature
+      And user "Brian" has been created with default attributes and without skeleton files
+      And user "Carol" has been created with default attributes and without skeleton files
+      And user "Alice" creates folder "/test" using the WebDAV API
+      And user "Alice" shares folder "/test" with user "Brian" with permissions "all" using the sharing API
+      And user "Alice" shares folder "/test" with user "Carol" with permissions "all" using the sharing API
+      When user "Alice" has uploaded file with content "uploaded content" to "/test/textfile0.txt"
+      When user "Brian" has uploaded file with content "uploaded content new" to "/test/textfile0.txt"
+      When user "Carol" has uploaded file with content "uploaded content new again" to "/test/textfile0.txt"
+      When user "Alice" gets the number of versions of file "/test/textfile0.txt" with edited by prop
```
phil-davis, line 483:
```suggestion
      When user "Alice" gets the number of versions of file "/test/textfile0.txt"
```

There is no step definition that has "with edited by prop" in the text. Is that an accident?
https://drone.owncloud.com/owncloud/core/33487/85/13
```
--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When user :arg1 gets the number of versions of file :arg2 with edited by prop
     */
    public function userGetsTheNumberOfVersionsOfFileWithEditedByProp($arg1, $arg2)
    {
        throw new PendingException();
    }
```

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33499/85/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33500/85/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33512/85/1

IljaN:
Rebased with master

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33570/85/1

IljaN:
image

pushnov-i:
> Debugging: image

yes, WIP

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33579/85/1

.htaccess
```diff
@@ -94,3 +94,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
phil-davis, line 97: Accidentally committed.

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33603/85/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiVersions-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33617/85/1

IljaN:
Squashed history a bit

IljaN:
All tests should pass now

IljaN:
https://github.com/owncloud/core/pull/39126/commits/ff13cf3235042bd86fb8fe2a0fa414c1924d203a was done because I am not sure how to mock the config here: https://github.com/owncloud/core/pull/39126/files#diff-15bc6a3d0922d2990f622f71c6f18a14ee37052dcf2ca8e65ed03cb128c01920R208

Maybe relegate this to acceptance tests?

JammingBen:
> was done because I am not sure how to mock the config here:

> [ff13cf3](https://github.com/owncloud/core/commit/ff13cf3235042bd86fb8fe2a0fa414c1924d203a) was done because I am not sure how to mock the config here: https://github.com/owncloud/core/pull/39126/files#diff-15bc6a3d0922d2990f622f71c6f18a14ee37052dcf2ca8e65ed03cb128c01920R208

You could either inject the config to the class (don't know if that's easily possible here) and mock it, or you have to set the config value during your tests (and reset if afterwards). The latter would kinda "break" the scheme of a unit test IMO, but I believe it is done like that quite often.

IljaN:
@JammingBen This is already done here: https://github.com/owncloud/core/blob/34ca8fc9adfa4a3a7be49efa0fbbf19641192f8d/apps/files_versions/tests/VersioningTest.php#L150

But not sure how to do this with the static storage class. In all other classes it is non-static and the overriding works. But not in this case.

apps/files_versions/lib/Hooks.php
```diff
@@ -36,6 +36,9 @@ class Hooks {
 	public static function connectHooks() {
 		// Listen to write signals
 		\OCP\Util::connectHook('OC_Filesystem', 'write', 'OCA\Files_Versions\Hooks', 'write_hook');
+		if (\OC::$server->getConfig()->getSystemValue('file_storage.save_version_author', false) === true) {
+			\OCP\Util::connectHook('OC_Filesystem', 'post_write', 'OCA\Files_Versions\Hooks', 'post_write_hook');
+		}
```
JammingBen, line 41: This seems outdated? I can't find the `post_write_hook` method in this class 🤔

apps/files_versions/js/versioncollection.js
```diff
@@ -11,12 +11,29 @@
 /* global moment */
 
 (function() {
+
+	_.extend(OC.Files.Client, {
+		PROPERTY_FILEID:	'{' + OC.Files.Client.NS_OWNCLOUD + '}id',
+		PROPERTY_VERSION_EDITED_BY:	'{' + OC.Files.Client.NS_OWNCLOUD + '}meta-version-edited-by',
+    PROPERTY_VERSION_EDITED_BY_NAME:	'{' + OC.Files.Client.NS_OWNCLOUD + '}meta-version-edited-by-name',
```
JammingBen, line 18: Some indentations are a bit off in the js-files, like here for example

apps/files_versions/lib/Storage.php
```diff
@@ -192,13 +194,27 @@ public static function store($filename) {
...
+				$users_view->putFileInfo($versionFileName, [
 					'checksum' => $sourceFileInfo->getChecksum(),
 				]);
+
+				$config = \OC::$server->getConfig();
+				if ($config->getSystemValue('file_storage.save_version_author', false) === true) {
+					$user = \OC::$server->getUserSession()->getUser();
+					if ($user !== null && !$users_view->file_exists($versionFileName . '.json')) {
+						$versions = self::getVersions($uid, $filename);
+						$metaDataKey =  \sizeof($versions) == 1 ? Constants::CREATED_BY_USER_METADATA : MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME;
```
JammingBen, line 212:
```suggestion
						$metaDataKey =  \sizeof($versions) === 1 ? Constants::CREATED_BY_USER_METADATA : MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME;
```

apps/files_versions/lib/Storage.php
```diff
@@ -356,11 +389,18 @@ public static function restoreVersion($uid, $filename, $fileToRestore, $revision
 		if (self::copyFileContents($users_view, $fileToRestore, 'files' . $filename)) {
 			$users_view->touch("/files$filename", $revision);
 			Storage::scheduleExpire($uid, $filename);
+
+			$users_view->unlink($fileToRestore . '.json');
```
JammingBen, line 393: Do we need to check here if the file exists or if `file_storage.save_version_author` is enabled?

JammingBen (commented):
Cool! 🚀  Some minor annotations.

apps/files_versions/lib/Hooks.php
IljaN: :+1:

JammingBen:
> @JammingBen This is already done here:
> 
> https://github.com/owncloud/core/blob/34ca8fc9adfa4a3a7be49efa0fbbf19641192f8d/apps/files_versions/tests/VersioningTest.php#L150
> 
> But not sure how to do this with the static storage class. In all other classes it is non-static and the overriding works. But not in this case.

Okay but then `enabledMetadataFeature()` has to be called I think. Like done in `testRestoreSameStorage()`:

```
if ($metaDataEnabled) {
	$this->enabledMetadataFeature();
}
```

Or maybe I'm just not seeing it. Is it working for `testRestoreSameStorage()`?

IljaN:
Does not work, even if enableMetadataFeature is called:
```
There were 2 failures:

1) OCA\Files_Versions\Tests\VersioningTest::testRestoreSameStorage with data set "metaEnabled" (true)
A version metadata-file must be created for the file before restoration

/home/ilja/code/core/apps/files_versions/tests/VersioningTest.php:1054
/home/ilja/code/core/apps/files_versions/tests/VersioningTest.php:912

2) OCA\Files_Versions\Tests\VersioningTest::testRestoreCrossStorage with data set "metaEnabled" (true)
A version metadata-file must be created for the file before restoration

/home/ilja/code/core/apps/files_versions/tests/VersioningTest.php:1054
/home/ilja/code/core/apps/files_versions/tests/VersioningTest.php:926
```

Also tried adding: 

```
if ($metaDataEnabled) {
			\OC::$server->getConfig()->setSystemValue('file_storage.save_version_author', true);
}
```
before https://github.com/owncloud/core/blob/34ca8fc9adfa4a3a7be49efa0fbbf19641192f8d/apps/files_versions/tests/VersioningTest.php#L1000

IljaN:
Any hints on how to enable the feature correctly in testing? I am out of ideas, but without unit-tests we can't merge.

JammingBen:
Ohhh wait! https://github.com/owncloud/core/blob/feature/version-metadata/apps/files_versions/lib/Storage.php#L354 -> `setSystemValue`?? Shouldn't it be `getSystemValue`? That's why it's never evaluating to `true`.

IljaN:
This didn't help, the issue was the same, the bug with set was introduced in a later refactoring

IljaN:
@JammingBen Thanks for the hint. If CI-Passes now I would squash and then push again.

apps/dav/lib/Meta/MetaFile.php
```diff
@@ -126,4 +127,27 @@ public function getContentDispositionFileName() {
 	public function getNode() {
 		return $this->file;
 	}
+
+	/**
+	 * @return string
+	 */
+	public function getVersionAuthor() {
+		if ($this->file instanceof IProvidesVersionAuthor) {
+			return $this->file->getCreatedBy() ?: $this->file->getEditedBy();
```
DeepDiver1975, line 136: I did not yet have a deeper look .... but shouldn't this be inverted?
The version author is the one who edited the file last - if there is none -> use the creator.

Is it possible that neither of them is set?

apps/dav/lib/Meta/MetaFile.php
```diff
@@ -126,4 +127,27 @@ public function getContentDispositionFileName() {
 	public function getNode() {
 		return $this->file;
 	}
+
+	/**
+	 * @return string
+	 */
+	public function getVersionAuthor() {
```
DeepDiver1975, line 134: adding return type?

apps/files_versions/js/versionstabview.js
```diff
@@ -213,7 +214,9 @@
 				revertIconUrl: OC.imagePath('core', 'actions/history'),
 				previewUrl: getPreviewUrl(version),
 				revertLabel: t('files_versions', 'Restore'),
-				canRevert: (this.collection.getFileInfo().get('permissions') & OC.PERMISSION_UPDATE) !== 0
+				canRevert: (this.collection.getFileInfo().get('permissions') & OC.PERMISSION_UPDATE) !== 0,
+				editedBy: version.has('editedBy'),
+        editedByName: version.has('editedByName')
```
DeepDiver1975, line 219: indent

apps/files_versions/lib/Storage.php
```diff
@@ -222,10 +238,6 @@ public static function markDeletedFile($path) {
 	 */
 	protected static function deleteVersion($view, $path) {
 		$view->unlink($path);
-		/**
-		 * @var \OC\Files\Storage\Storage $storage
```
DeepDiver1975, line 226: why was that removed?

lib/private/Files/Meta/MetaFileVersionNode.php
```diff
@@ -55,6 +56,16 @@ class MetaFileVersionNode extends AbstractFile implements IPreviewNode, IProvide
 	/** @var array */
 	private $versionInfo;
 
+	/**
+	 * @var string
+	 **/
+	private $editedBy = "";
```
DeepDiver1975, line 62: these two props are not really needed.

lib/private/Files/Meta/MetaFileVersionNode.php
```diff
@@ -77,8 +88,30 @@ public function __construct(
...
+
+		if (isset($version['created_by'])) {
+			$this->createdBy = $version['created_by'];
+		}
 	}
 
+	/**
+	 * @return string
+	 */
+	public function getEditedBy() {
```
DeepDiver1975, line 104: php return type

lib/public/Files/IProvidesVersionAuthor.php
```diff
@@ -0,0 +1,45 @@
...
+ * @package OCP\Files
+ * @since 10.9.0
+ */
+interface IProvidesVersionAuthor {
+	/**
+	 * Returns the version author's username
+	 * @return string
+	 * @since 10.9.0
+	 */
+	public function getEditedBy();
```
DeepDiver1975, line 37: php return type

lib/public/Files/IProvidesVersionAuthor.php
```diff
@@ -0,0 +1,45 @@
...
+
+	/**
+	 * Returns the file author's username
```
DeepDiver1975, line 40: desc of both methods should reflect the different usage.

DeepDiver1975 (changes requested):

lib/private/Files/Meta/MetaFileVersionNode.php
IljaN: They are used by the getters, which are needed to satisfy the interface

DeepDiver1975: sure - but all the logic can be implemented purly in the getters. No need to use props and the related logic in the ctor.

IljaN:
@DeepDiver1975 All changes commited, checking https://github.com/owncloud/core/pull/39126/files#r750037657 separately.

apps/dav/lib/Meta/MetaFile.php
DeepDiver1975: If I understood the logic correctly either of the two is set - strange behavior for my understanding.

If there is always only one available - just implement one method.
No reason to overcomplicate things.

IljaN: Thinking about your first comment, there might be extreme-edge-case (metadata files got lost due to error) I think in this case non of the both are set.
----

- We have a metadata-file collection for each file (one metdata-file corresponding to each version file)
- Chronologically the first metadata-file (first version) contains the "created-by" field while all subsequent metadata files contain the "edited-by" field.

DeepDiver1975: Looking deeper in the code:
- there is always either getCreatedBy or getEditedBy filled
- on dav api level - there is only one property
-> use only one property in the  file node api level as well

DeepDiver1975 (dismissed):

apps/dav/lib/Meta/MetaFile.php
IljaN: But semantically created_by != edited_by not sure if we should use one property for both

IljaN: Hmhh I think you are right, the initial idea was to have the initial version to have the "created-by" key, but in reviews we decided against it since it is inconsistent with the "normal" version implementation where the very first version of a file does not get a version created. So in this case we can remove the created-by key alltogether? 

Also this is currently wrong as the second edit to a file creates a first version which has the "created-by" key in the metadata.

phil-davis:
Note: This branch is 4 commits ahead, 71 commits behind master. 
https://github.com/owncloud/core/tree/feature/version-metadata

When doing the next changes/refactoring/... I suggest rebase to get the very latest CI result.

phil-davis:
- user uu1 creates folder "nf"
- user "uu1" shares folder "nf" with user "uu2"
- user uu1 creates file "atf.txt" in folder "nf"
- no versions are listed for "atf.txt" - correct
- user uu2 edits file "atf.txt" in folder "nf"
- user uu2 show the version history
- one version is shown - correct
- the version "owner" is displayed as "uu2" - not correct, that was the original file, which was created by uu1

version-user-is-off-by-one

That could be confusing for users who want choose a version to restore based on "who edited it"

phil-davis:
Here is an example with a file shared to a group of 4 users uu1 uu2 uu3 uu4. The version history shows uu2 uu3 uu4, but actually the current file is by uu4 and the 3 older versions are by uu1 uu2 uu3
edits-by-group-members

pmaier1:
> The requirements need to be defined.

The requirement is to see the user who **created** a version (or: the user who uploaded/saved this state of a file) as part of the version metadata. Basically this enhancement shall allow to answer the questions "Check the last version that Peter created" or "Let's roll back to the state before Peter started working on it". 

Here's the original input https://github.com/owncloud/enterprise/issues/4518.

IljaN:
This requirement was removed in a subsequent call => Creator can be checked via activity app. Will clarify with @pmaier1 now. 

If we need to show the creator than this would mean that we will need to show the very first version of a file in the versions view which we don't do yet.

phil-davis:
We don't have to show the "creator" of the "current version" of a file (the "version" that is not shown in the version history, because it is not yet history!) But technically, IMO, we will have to remember it somewhere so that it is available if and when that data becomes an "old version".

At the moment, if 
1) Alice creates a file
2) Brian edits it
3) Carol edits it

At step (2) it is Brian who is authenticated. I don't think "the system" has any way to reliably know that the original file was created by Alice.

At step (3) it is Carol who is authenticated. I don't think "the system" has any way to reliably know that the file before the edit was uploaded by Brian.

IljaN:
Maybe we can create a specially-named metadata-file which is not returned by the webdav-api but is replaced as soon as a new version is available.

IljaN:
> Maybe we can create a specially-named metadata-file which is not returned by the webdav-api but is internally replaced as soon as a new version is available.

@pmaier1 Will implement something like this

phil-davis:
I raised an issue to make some acceptance tests for this. We can do that in parallel in a branch on top of this.

Note my comment https://github.com/owncloud/core/issues/39504#issuecomment-972533285

Currently for this feature there is no way for end-users to find out who is the author of the data in the current file. If I was user "Alice" looking at a file, and seeing the author of each old version in the version history, an thinking about restoring some version, then probably I am doing it because the current data in the file is "wrong/corrupt" in some way. So I will immediately want to know "who is the author of the current data in the file?"

As Alice I can already find out other meta-data about the current file - modified date, tags, comments. Should users be able to find out the "latest author"?

IljaN:
As of now we don't write versions for empty files (filesize=0): So we also shouldn't write metadata for emtpy files. (?)

phil-davis:
> Possible edge-case: As of now we don't write versions for empty files (filesize=0): So we also shouldn't write metadata for emtpy files. (?)

And a combination edge-case. If we add some way for users to find out the "author" of the current file, then if there is version history of non-empty versions of the file, and "someone" has uploaded an empty file as the current file, then I think that user Alice will want to know who uploaded the "nothing" to the file.

AlexAndBear (approved):
Approving for now - Still in Beta Phase - Will be ready before RC

IljaN:
Continued here: https://github.com/owncloud/core/pull/39516

jnweiger:
## tested in 10.9.0 beta1

* works fine with and without encryption.
* delete restore should be improved -> 39571

phil-davis:
Note: I am quite sure that the version authors are off-by-one in 10.9.0 beta1. That is being fixed in #39516

PR #39131, opened 2021-08-25
Implement meta data for part files

JammingBen:
## Description
https://github.com/owncloud/core/pull/39088 introduced the hashing of file names on chunked uploads. However, we experienced that some apps need the file path including the "real" (unhashed) file name. The `files_classifier` e.g. determines the file extension during a chunked upload based on the name.

Therefore this PR implements a way to retrieve meta data for part files. This way, the real path or file name can be retrieved.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/File.php
+++ b/apps/dav/lib/Connector/Sabre/File.php
@@ -170,7 +170,7 @@ class File extends Node implements IFile, IFileNode {
 
 		if ($usePartFile) {
 			// mark file as partial while uploading (ignored by the scanner)
-			$partFilePath = $this->getPartFileBasePath($this->path) . '.ocTransferId' . \rand() . '.part';
+			$partFilePath = Filesystem::createPartFilePath($this->path);
 		} else {
 			// upload file directly as the final path
 			$partFilePath = $this->path;
@@ -300,15 +300,6 @@ class File extends Node implements IFile, IFileNode {
 			\OC::$server->getEventDispatcher()->dispatch('file.afterupdate', $afterEvent);
 		}
 		return '"' . $this->info->getEtag() . '"';
-	}
-
-	private function getPartFileBasePath($path) {
-		$partFileInStorage = \OC::$server->getConfig()->getSystemValue('part_file_in_storage', true);
-		if ($partFileInStorage) {
-			return Filesystem::hashFileName($path);
-		} else {
-			return \md5($path); // will place it in the root of the view with a unique name
-		}
 	}
 
 	/**
@@ -530,7 +521,7 @@ class File extends Node implements IFile, IFileNode {
 
 				if ($usePartFile) {
 					// we first assembly the target file as a part file
-					$partFile = $this->getPartFileBasePath($path . '/' . $info['name']) . '.ocTransferId' . $info['transferid'] . '.part';
+					$partFile = Filesystem::createPartFilePath($path . '/' . $info['name'], $info['transferid']);
 					/** @var \OC\Files\Storage\Storage $targetStorage */
 					list($partStorage, $partInternalPath) = $this->fileView->resolvePath($partFile);
 
--- a/lib/private/Files/Filesystem.php
+++ b/lib/private/Files/Filesystem.php
@@ -211,6 +211,9 @@ class Filesystem {
 	/** @var bool */
 	private static $logWarningWhenAddingStorageWrapper = true;
 
+	/** @var array */
+	private static $partFileMetaData = [];
+
 	/**
 	 * @param bool $shouldLog
 	 * @return bool previous value
@@ -1113,5 +1116,58 @@ class Filesystem {
 		}
 
 		return $hashedFileName;
+	}
+
+	/**
+	 * Create the path to a part file. It consists of the hashed file path
+	 * or name (depending on part_file_in_storage) and a transferId. E.g.
+	 * /files/098f6bcd4621d373cade4e832627b4f6.ocTransferId2111130212.part
+	 *
+	 * This method also saves some meta data for the duration of the request.
+	 * This data can be retrieved by passing the hashed file name
+	 * to getPartFileMetaData().
+	 *
+	 * @param string $path path to the original file
+	 * @param string|null $transferId generates a random string if not given
+	 * @return string hashed file path or name
+	 */
+	public static function createPartFilePath($path, $transferId = null) {
+		if (!$transferId) {
+			$transferId = \rand();
+		}
+
+		// if part_file_in_storage is false, we will hash the whole file path
+		// as the part file is then always stored in the user's root directory.
+		$partFileInStorage = \OC::$server->getConfig()->getSystemValue('part_file_in_storage', true);
+		if ($partFileInStorage) {
+			$hashedPath = self::hashFileName($path);
+		} else {
+			$hashedPath = \md5($path);
+		}
+
+		$partFilePath = "$hashedPath.ocTransferId$transferId.part";
+		$partFileName = \basename($partFilePath);
+		self::$partFileMetaData[$partFileName] = [
+			'path' => $partFilePath,
+			'originalRoot' => self::getRoot(),
+			'originalPath' => $path,
+			'originalName' => \basename($path),
+		];
+
+		return $partFilePath;
+	}
+
+	/**
+	 * Retrieve the meta data of a part file by its hashed part file name. This way
+	 * you can also retrieve the original file path and name.
+	 *
+	 * This presumes that createPartFilePath() was called before during this
+	 * request cycle.
+	 *
+	 * @param string $partFileName e.g. 098f6bcd4621d373cade4e832627b4f6.ocTransferId2111130212.part
+	 * @return array|null the found meta data as array or null if none found
+	 */
+	public static function getPartFileMetaData($partFileName) {
+		return self::$partFileMetaData[$partFileName] ?? null;
 	}
 }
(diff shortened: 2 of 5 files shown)


JammingBen:
Although I find the approach quite O.K., I'm not a 100% percent sure about it. @IljaN @mrow4a As you also reviewed the original PR, any thoughts in it?

@phil-davis This would help to fix all these failing pipelines. But of course we would need to implement proper fixes in all affected apps. Which would also mean that these apps have to be released together with OC 10.9 and are not compatible with older OC versions anymore.

JammingBen:
@jvillafanez Maybe you have an idea about it? Still looking for the best approach to solve this.

jvillafanez:
If we only need to keep the extension, we can hash only the filename and append the extension. "file.txt" could be converted to "abc123.txt" (`hash('file') + '.txt'` or maybe `hash('file.txt') + '.txt'`). It's likely irrelevant if we have to hash the filename with or without the extension.

If we need the original filename, I think it's better to plan it like it's metadata. You have the hashed filename, which has associated metadata, including the original filename (with the ".transfer" id), the target filename, the extension, and other things we could add.
The question in this case is whether we need to persist this information somewhere and if it will apply only to chunked uploads

mrow4a:
I still hold my opinion as in https://github.com/owncloud/core/pull/39088#issuecomment-893310198 , and make new chunking algorithm incompatible with whatever is incompatible as part of docs.

mrow4a:
I also think it makes sense to consider filename/extension etc like metadata. not sure how high development would be required for this.

JammingBen:
I agree with the meta data approach. So yeah, where to store such information?

> The question in this case is whether we need to persist this information somewhere and if it will apply only to chunked uploads

I think the information is needed for the duration of the upload/request. I don't see a reason to keep this information afterwards as the chunked parts don't exist anymore. Thus the cache may be a good place maybe?

> If we only need to keep the extension, we can hash only the filename and append the extension. "file.txt" could be converted to "abc123.txt" (hash('file') + '.txt' or maybe hash('file.txt') + '.txt'). It's likely irrelevant if we have to hash the filename with or without the extension.

That's just an example. The encryption app for example needs the original file name to retrieve the encrypted version.

jvillafanez:
> I think the information is needed for the duration of the upload/request. I don't see a reason to keep this information afterwards as the chunked parts don't exist anymore. Thus the cache may be a good place?

If the Filesystem class is the one hashing, then it should be the one holding the metadata. I don't expect we need to hold a lot of information from a lot of files, so we can keep the information in the memory with a class variable.

JammingBen:
@jvillafanez Please have a look. `Filesystem` now provides a way to store and retrieve meta data of part files. `getPartFilePath()` automatically stores meta data for the given part file. It also handles the `transferId`-stuff now.

ownclouders:

https://drone.owncloud.com/owncloud/core/32060/174/1

apps/dav/lib/Connector/Sabre/File.php
```diff
@@ -302,13 +302,21 @@ public function put($data) {
...
 		} else {
-			return \md5($path); // will place it in the root of the view with a unique name
+			$hashedPath = \md5($path); // will place it in the root of the view with a unique name
 		}
+
+		$partFilePath = "$hashedPath.ocTransferId$transferId.part";
+		$root = $this->fileView->getRoot();
+		$absolutePartFilePath = $root . '/' . \ltrim($partFilePath, '/');
+		$absoluteFilePath = $root.$path;
+		Filesystem::addPartFileMetaData($absolutePartFilePath, $absoluteFilePath);
```
jvillafanez, line 317: What about something a bit more isolated?
```
// Filesystem.php
static function generatePartFileMetadata($path, $root) {
  $hashedPath = self::hashFileName($path);
  $transferId = rand();
  $partFilePath = "$hashedPath.ocTransferId$transferId.part";
  $partFileName = \basename($hashedPath);
  self::$partFileMetaData[$partFileName] = [
    'absolutePath' => $partFilePath,
    'originalPath' => $path,
    'originalName' => \basename($path),
  ];
  return $hashedPath;
}
```
Not sure if we need the root or not. Maybe it should be part of the path already, no need to ask for it.

The expected usage in the Sabre/File.php should be something like:
```
$hashedPath = Filesystem::generatePartFileMetadata($path, $this->fileView->getRoot());
```

I'm not fully sure, but I guess we want to force the metadata to be generated, so we probably should either integrate the `Filesystem::hashFileName` function inside the `generatePartFileMetadata` method or make it private.
Basically, we want to ensure that any hashed path has metadata associated, so we don't want to use 2 different functions.

JammingBen: It's now handled in one method in `Filesystem.php`, although I kept the name `getPartFilePath()`. The caller is still mainly interested in getting the part file path IMO.

I removed the root concatenation and added the root path as value to the meta data. I'm not sure but we may need this info in other places.

lib/private/Files/Filesystem.php
```diff
@@ -1114,4 +1117,45 @@ public static function hashFileName($path) {
 
 		return $hashedFileName;
 	}
+
+	/**
+	 * Return the hashed part file path of a file and store its
```
jvillafanez, line 1122: This might need a better explanation about what's the "hashed part file path of a file".
We should also link with the `getPartFileMetaData` method so it's more clear how we can retrieve the stored metadata, and we'll need to clarify the information is bound to the request and won't be persisted

lib/private/Files/Filesystem.php
```diff
@@ -1114,4 +1117,45 @@ public static function hashFileName($path) {
...
+	public static function getPartFilePath($path, $transferId = null) {
+		if (!$transferId) {
+			$transferId = \rand();
+		}
+
+		$partFileInStorage = \OC::$server->getConfig()->getSystemValue('part_file_in_storage', true);
+		if ($partFileInStorage) {
+			$hashedPath = self::hashFileName($path);
+		} else {
+			$hashedPath = \md5($path); // will place it in the root of the view with a unique name
```
jvillafanez, line 1137: any chance to reuse the `hashFileName` method somehow? It's unlikely that we'll change the hash algorithm from md5 to another one, but it seems better if we don't need to change it in 2 different places

lib/private/Files/Filesystem.php
```diff
@@ -1114,4 +1117,45 @@ public static function hashFileName($path) {
...
+		}
+
+		$partFilePath = "$hashedPath.ocTransferId$transferId.part";
+		$partFileName = \basename($partFilePath);
+		self::$partFileMetaData[$partFileName] = [
```
mrow4a, line 1142: getPartFilePath for $path? and getPartFileMetadata for $partfilepath ? I know in here we know the function call content, but normally you would be exposed only with interface description

lib/private/Files/Filesystem.php
```diff
@@ -1114,4 +1117,45 @@ public static function hashFileName($path) {
...
+		return $partFilePath;
+	}
+
+	/**
+	 * Retrieve the meta data of a file part.
+	 *
+	 * @param string $partFileName
+	 * @return array|null
+	 */
+	public static function getPartFileMetaData($partFileName) {
```
mrow4a, line 1158: should be part of doc for this function that before using this function one has to call getPartFilePath to populate partfile metadata for some path?

lib/private/Files/Filesystem.php
```diff
@@ -1114,4 +1117,45 @@ public static function hashFileName($path) {
...
+			'originalRoot' => self::getRoot(),
+			'originalPath' => $path,
+			'originalName' => \basename($path),
+		];
+
+		return $partFilePath;
+	}
+
+	/**
+	 * Retrieve the meta data of a file part.
```
jvillafanez, line 1153: Need clarification. It needs the basename of the hashed part generated by the `getPartFilePath` method, not any ".part" file

jvillafanez (commented):
Looks good. It just need some additional phpDoc info.

jvillafanez:
In addition, for the methods returning different types, please specify when each type is expected to be returned.
A `@return array|null` could imply that null is returned on success and an array containing multiple error messages is returned if something goes wrong. I know it's a dumb example, but it could happen.

lib/private/Files/Filesystem.php
JammingBen: `getPartFileMetadata()` should be called via the name only, because we never now for sure how the path looks like. Could be internal, could be with root, without root etc. Specifying this on the docs should be enough, don't you think?

mrow4a:
@JammingBen I still find it confusing a little bit the interdependency between function calls. I would really name that `getPartFilePath` to reflect what it actually does, thus `createPartFileId()` , and then one could recover metadata using`getPartFileMetadata($partFileId)`  . It still feels a bit dirty and we beg ourselves for some race-condition or misunderstanding of the interface.

JammingBen:
@mrow4a @jvillafanez I tried to clarify the docs and names, please check again.

Reusing `hashFileName()` to hash the file path is a bit ugly, because it's supposed to hash the file name by preserving the rest of the path. We could pass the algorithm via parameter though.

jvillafanez (approved):

ownclouders:
:boom: Acceptance tests pipeline <strong>apiProxySmoke-8-4-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32070/176/1

mrow4a:
@DeepDiver1975 could you also check this out ? I am a little concerned about this change regarding interdependency of two functions. Have a look at unit test

JammingBen:
Closing this in favor of an overall less destructive approach: https://github.com/owncloud/core/pull/39168

PR #39136, opened 2021-08-26
feat: new command to move a user's home folder - occ user:move-home

DeepDiver1975:
## Description
```
➭ php7.4 ./occ help user:move
Description:
  Move a user's home folder to a new location.

Usage:
  user:move-home [<user_id> [<new_location>]]
```

## How Has This Been Tested?
- create a user user_1
- login
- create some files, version, thumbnails - maybe with encryption
- occ u:mo user_1 /tmp
- make sure all files are copied over to /tmp/user_1
- delete original user home (data/user_1)
- login again and check if all files are still accessible

Screenshot from 2021-08-27 15-40-32

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/3971

### DIFF
--- a/core/Command/User/MoveHome.php
+++ b/core/Command/User/MoveHome.php
@@ -1,0 +1,137 @@
+<?php
+/**
+ * @author Thomas Müller <thomas.mueller@tmit.eu>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Core\Command\User;
+
+use InvalidArgumentException;
+use OC\User\AccountMapper;
+use OCP\IUser;
+use OCP\IUserManager;
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class MoveHome extends Command {
+
+	/**
+	 * @var IUserManager
+	 */
+	private $userManager;
+
+	public function __construct(
+		IUserManager $userManager
+	) {
+		parent::__construct();
+		$this->userManager = $userManager;
+	}
+
+	protected function configure() {
+		$this->setName('user:move-home')
+			->setDescription('Move a user\'s home folder to a new location.')
+			->addArgument('user_id', InputArgument::REQUIRED, 'Id of the user whose home folder is to be moved. ')
+			->addArgument('new_location', InputArgument::REQUIRED, 'Absolute path to the parent folder of the new location of the home folder.')
+			;
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		$user = $this->getUser($input);
+		$userId = $user->getUID();
+		$oldHome = $user->getHome();
+
+		$newLocation = $this->getNewLocationForUser($input, $user);
+		$output->writeln("Move $userId from $oldHome to $newLocation");
+
+		# disable user
+		$output->writeln("Disabling user ....");
+		$user->setEnabled(false);
+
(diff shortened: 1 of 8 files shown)


phil-davis:
https://drone.owncloud.com/owncloud/core/32044/1/9

needs a reabse - the CI now uses `owncloudci/nodejs:14`

mmattel:
@DeepDiver1975 to document properly, I need some info for understanding it:

We have in `config.php` the key `datadirectory` which points default lets say to `/var/www/owncloud/data`. Creating users will create for each user a subfolder with his user_id containing the default structure and user created content.

Now, when using the command above to migrate a users home, some questions raise:

1. Is it true that you can move the complete user folder from `data/` to somewhere else, using any possible target path which is not necessarily is in the owncloud directory (with proper rights for the webserver user of course).
2. In case (1) is true, this would mean that you have two datadirectory paths. One for the remaining users and one for the moved ones. Is this correct? If not, please explain
3. You can define a particular user in the command. This means that you have to rerun this command for every user you want to do so. Is there a plan to have additional options to do this for multiple users?
4. What is the usecase for, when does an admin need this command to run?
5. Are there any limitation necessary to know? Like what happens if there are big files to move where you maybe run into php/webserver timeouts or other stuff.

Sorry for that many questions, but they are necessary for a proper documentation.

DeepDiver1975:
> * Is it true that you can move the complete user folder from `data/` to somewhere else, using any possible target path which is not necessarily in the owncloud directory (with proper rights for the webserver user of course).
> 

yes

> * In case (1) is true, this would mean that you have two datadirectory paths. One for the remaining users and one (or many!) for the moved ones. Is this correct? If not, please explain.
> 

not only two data dirs - as many as you want ;-)

> * You can define a particular user in the command. This means that you have to rerun this command for every user you want to do so. Is there a plan to have additional options to do this for multiple users?
> 

nothing planned at the moment. doing multiple users at once can be too error prone

> * Is it correct that in the target path, the user folder _must not_ exist.
> 

yes - the parent must exist. the user folder is rsynced over to the new location

> * What is the usecase for, when does an admin need this command to run?
> 

think of an nfs based installation where different nfs shares exist and due to 'load balancing of storage usage' you want one user to have their files on a different share.


> * Are there any limitation necessary to know? Like what happens if there are big files to move where you maybe run into php/webserver timeouts or other stuff.

it is an occ command - no php timeouts play into this. heavy lifting of files is done using rsync
one limitation is that this obviously only works with posix storage and obviously not on object stores

mmattel:
Great answers, many thanks !

> not only two data dirs - as many as you want ;-)

I was not aware of this capability, how/where do you define multiple data dirs (in config.php), where can I read more? As far I read our documentation, `datadirectory` is a string and not an array...

core/Command/User/MoveHome.php
```diff
@@ -0,0 +1,137 @@
...
+		$user->setEnabled(true);
+
+		# notify admin about necessary LDAP change
+		$output->writeln('');
+		$output->writeln("Files for user $userId have been moved to $newLocation");
+		$output->writeln("In case you are using LDAP Home Folder please update the property in your LDAP directory for this user");
+		$output->writeln('');
+
+		$output->writeln('');
+		$output->writeln("The old home folder can now be deleted (rm -rf $oldHome)");
```
mmattel, line 87: Text proposal, because logged in admins also should not have root previledges by default:
```suggestion
		$output->writeln("The old home folder can now be deleted (sudo rm -rf $oldHome)");
```

mmattel:
Q: would it be possible to show the progress of rsync in the output
with the rsync option `--progress`?
maybe as an occ option like `-p`

This would help the admin to see how far the command has progressed already

DeepDiver1975:
> I was not aware of this capability, how/where do you define multiple data dirs (in config.php), where can I read more? As far I read our documentation, `datadirectory` is a string and not an array...

mainly via the ldap home folder app or some custom code - this is only necessary in very specific large scale setups.

DeepDiver1975:
> Q: would it be possible to show the progress of rsync in the output
> with the rsync option `--progress`?
> maybe as an occ option like `-p`
> 
> This would help the admin to see how far the command has progressed already

Not easily right away .... one would need to do some research ...

core/Command/User/MoveHome.php
DeepDiver1975: sudo -u www-data on ubuntu/debian
sudo -u www on suse and related
sudo -u apache on redhat (if I remember correctly)

what to do in cases where sudo is not available ....


too many options - an admin knows how to execute this command on their systems

mmattel: Misunderstanding, without `-uwww`, I did not mention this...
We have Ubuntu as default in our docs and ALWAYS use sudo in cases where higher priviledges are used.

mmattel:
See related issue: https://github.com/owncloud/enterprise/issues/4637

AlexAndBear (approved):
Code-wise LGTM, also works quite well on my system

PR #39143, opened 2021-08-27
default pollinterval 30000 msec

jnweiger:
The desktop client uses a default poll interval of 30000 msec whenever the capabilities announce something less than 5000.
To get in sync with this behaviour, I suggest to also change the value here to 30000.

A value of 60 also leads to confusion, as users assume the unit is seconds, which it is not.

Suggestion for future: Hardcode the unit in the name of the variable, like. e.g. `pollinterval_msec`

Related doc issue: https://github.com/owncloud/docs/issues/3970

Doc issue transferred to https://github.com/owncloud/docs-server/issues/109

### DIFF
--- a/changelog/unreleased/39143
+++ b/changelog/unreleased/39143
@@ -1,0 +1,6 @@
+Change: Update the default poll-interval in capabilities
+
+The default pollinterval advertised in capabilities has been set to
+30000 milliseconds. Previously it was 60 milliseconds.
+
+https://github.com/owncloud/core/pull/39143
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1492,6 +1492,16 @@ $CONFIG = [
 'minimum.supported.desktop.version' => '2.3.3',
 
 /**
+ * Define the suggested poll interval for clients
+ * Specifies how often clients should poll the server for changes.
+ * The value is in milliseconds. The value is not enforced.
+ * Clients may use this value to decide how frequently to check the server for
+ * changes.
+ *
+ */
+'pollinterval' => 30000,
+
+/**
  * Define whether to include external storage in quota calculation
  * EXPERIMENTAL: option whether to include external storage in quota
  * calculation, defaults to false.
--- a/lib/private/OCS/CoreCapabilities.php
+++ b/lib/private/OCS/CoreCapabilities.php
@@ -51,7 +51,8 @@ class CoreCapabilities implements ICapability {
 	public function getCapabilities() {
 		return [
 			'core' => [
-				'pollinterval' => $this->config->getSystemValue('pollinterval', 60),
+				// pollinterval is an integer number of milliseconds
+				'pollinterval' => $this->config->getSystemValue('pollinterval', 30000),
 				'webdav-root' => $this->config->getSystemValue('webdav-root', 'remote.php/webdav'),
 				'status' => Util::getStatusInfo(true),
 				'support-url-signing' => true,
--- a/tests/acceptance/features/apiCapabilities/capabilities.feature
+++ b/tests/acceptance/features/apiCapabilities/capabilities.feature
@@ -41,7 +41,7 @@ Feature: capabilities
     And the HTTP status code should be "200"
     And the capabilities should contain
       | capability    | path_to_element                           | value             |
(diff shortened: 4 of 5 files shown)


butonic (approved):

butonic:
yes, please change the default to align with the 30sec / 30000msec used by clients

phil-davis (approved):

changelog/unreleased/39143
```diff
@@ -0,0 +1,6 @@
+Change: Update the default poll-interval in capabilities
+
+The default pollinterval advertised in capabilities has been set to
+30000 milliseconds. Previously it was 60 milliseconds.
```
mmattel, line 4: wasnt this 60 seconds??

phil-davis: > wasnt this 60 seconds??

It was "60" of an unspecified unit-of-measure. Existing clients that use it have code that treats it as milliseconds (desktop-sync-client and iOS IMO are the known clients that make use of it)

It is not enforced in any way by the server, so there is no server code that processes the value in a check, so there is no way to know for sure in the server what the intended unit-of-measure ever was.

What to write here? "Previously it was 60 unspecified units"?

butonic: well, it was `60` ... no dimension given ... could have been shoes ... I only found the discussion in https://github.com/owncloud/client/pull/8777

@dragotin @TheOneRing the desktop cient assumes milliseconds?
@felix-schwarz ios assumes what?
@michaelstingl what about android?

TheOneRing: The client side handling was implemented last year.
In order to keep the client behavior compatible it was discussed to only handle the value if it has the form of `60!` to define an explicitly set value.
In a following discussion we discussed that using milliseconds would feel more correct.
The clients would still only change their behavior if the value was explicitly touched.
With out the explicit requirement the clients would have changed the polling interval without any apparent reason.

mmattel:
Note following comment in client#8777

> Note: To use 60 seconds, the capability needs to contain the string "60!" for the reason that 60 was the default in the capas, but using that would change the behaviour of all clients which needs to be avoided.

Is this still valid and does this has an impact to documentation?

TheOneRing:
> Note following comment in [owncloud/client#8777](https://github.com/owncloud/client/pull/8777)
> 
> > Note: To use 60 seconds, the capability needs to contain the string "60!" for the reason that 60 was the default in the capas, but using that would change the behaviour of all clients which needs to be avoided.
> 
> Is this still valid and does this has an impact to documentation?

See https://github.com/owncloud/core/pull/39143#discussion_r811743976

changelog/unreleased/39143
phil-davis: I raised client PR https://github.com/owncloud/client/pull/9465 to (hopefully) clarify some of the comments there.

phil-davis:
I raised client PR https://github.com/owncloud/client/pull/9465 to adjust some client comments etc.

IMO this core PR is ready to merge - the general agreement is that the unit-of-measure for `pollinterval` will be advertised/documented as milliseconds, and all clients that take any notice of it will implement/do implement it that way.

Can we merge?

changelog/unreleased/39143
phil-davis: https://github.com/owncloud/client/pull/9465 has been merged.
Android client team have confirmed that they do not use `pollinterval` - so no impact on that.
Desktop client ignores any value under 5000 advertised by the server, and keep its internal default of 30000. So old servers that are advertising 60 are effectively ignored.

phil-davis:
Merging - if anyone objects we can modify something, but IMO this is "a good thing".

changelog/unreleased/39143
felix-schwarz: @butonic Sorry for being late to the party. For documentation: 
- the iOS SDK uses a `10` seconds poll interval by default
- the default value of `60` is ignored
- for values other than `60` coming from the server:
	- the number is interpreted as milliseconds
	- the number is ignored if it is below `5` seconds. In that case, the client will log a warning.

jnweiger:
Confirmed fixed in 10.10.0 rc2

PR #39148, opened 2021-08-30
Remove tag edit/trash actions from file list

### LINKED ISSUE #39147: Edit/Delete tag actions are highly misleading
### Steps to reproduce
1. Create a tag
2. Go to file list
3. Open file details (right sidebar)
4. Click on "Tags"
5. Click in input field
6. Click on Trash bin icon right to a tag

### Expected behaviour
The UI suggests that the trash bin icon removes the tag from the file

### Actual behaviour
The tag gets deleted completely possibly having huge consequences, for example on retention policies.

### Server configuration
**ownCloud version:** 10.8

grafik

hurradieweltgehtunter:
## Description
This PR's removes the edit/delete actions from the tag selection list. While those actions are only visible to the admin it is still a very misleading way since users think they remove a tag from a file but instead delete it completely. This can have bad consequences, for example breaking retention policies.

## Related Issue
- Fixes #39147 

Before:
grafik

After:
grafik

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/39148
+++ b/changelog/unreleased/39148
@@ -1,0 +1,9 @@
+Change: Remove misleading tag edit/trash actions
+
+This PR's removes the edit/delete actions from the tag selection list.
+While those actions are only visible to the admin it is still a very misleading 
+way since users think they remove a tag from a file but instead delete it completely. 
+This can have bad consequences, for example breaking retention policies.
+
+https://github.com/owncloud/core/issues/39147
+https://github.com/owncloud/core/pull/39148
--- a/core/js/systemtags/systemtagsinputfield.js
+++ b/core/js/systemtags/systemtagsinputfield.js
@@ -22,12 +22,6 @@
 		'{{else}}' +
 		'    <span class="label">{{name}}</span>' +
 		'{{/if}}' +
-		'{{#allowActions}}' +
-		'    <span class="systemtags-actions">' +
-		'        <a href="#" class="rename icon icon-rename" title="{{renameTooltip}}"></a>' +
-		'        <a href="#" class="delete icon icon-delete" title="{{deleteTooltip}}"></a>' +
-		'    </span>' +
-		'{{/allowActions}}' +
 		'</span>';
 
 	var SELECTION_TEMPLATE =


ownclouders:

https://drone.owncloud.com/owncloud/core/32021/58/1

ownclouders:

https://drone.owncloud.com/owncloud/core/32021/60/1

ownclouders:
:boom: Acceptance tests pipeline <strong>webUITags-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32023/154/1

JanAckermann:
I highly understand the thoughts behind it, clicking on the trash item might be understood as removing the tag from the item, but removes the whole tag (and for all tagged items). 

On the other hand, we don't have any default management system for tags so far, leaving the admin user with no choice to delete or rename a tag which would be a big downside. 

I would recommend rendering a prompt dialog while clicking on the trash icon with the message, that this might remove the tag and also unassign the tag for all items -> Confirm / Decline.

Codewise, we have `OC.dialogs.confirm` for this. 

@pmaier1 would do you think ?

pmaier1:
I absolutely agree that the usability here is bad and that the trash bin can be misleading. Generally speaking, the tagging feature design has a number of shortcomings we know about.

> While those actions are only visible to the admin it is still a very misleading way since users think they remove a tag from a file but instead delete it completely.

This is not true. The feature is available for all users. All users can see the (regular) tags that are assigned to files/folders they have access to and can unassign/edit/delete them. There are special types of tags introduced by the `systemtags_management` app that have more restrictions (those should be used e.g., for automations).

If we just remove that feature as you suggest, only admins would be able to remove tags using `systemtags_management` (enterprise feature!) and they would not even know how those tags are used. Without providing an alternative, this can't be accepted.

> I would recommend rendering a prompt dialog while clicking on the trash icon with the message, that this might remove the tag and also unassign the tag for all items -> Confirm / Decline.

Thanks. That sounds like a good solution to me :+1:

JanAckermann:
=> Decision add a prompt for deleting, will take care

Issue #39038, opened 2021-07-21
Add type declarations in acceptance test code


phil-davis:
Bugs in test code like https://github.com/owncloud/notifications/pull/346 would be caught very quickly if all function calls had type declarations explicitly for each parameter, and a return type defined.

Enhance function declarations to be like:
```
	public function substituteInLineCodes(
		string $value,
		string $user = null,
		array $functions = [],
		array $additionalSubstitutions = []
	): string {
```

And then fix any test code that fails.

SwikritiT:
This issue is continued here - https://github.com/owncloud/core/issues/39243

phil-davis:
Other parts will be:
- add type declarations in lib folder (of core)
- add type declarations in oC10 app acceptance test code

sakshamgurung:
`lib` part of this issue is continued here https://github.com/owncloud/core/issues/39333

PR #39154, opened 2021-08-31
[tests-only] [full-ci] adds function param and return type declaration

sakshamgurung:
## Description

- Adds function parameters and return type declaration for existing context files
- Added for:
- [x] CalDavContext.php
- [x] CardDavContext.php
- [x] WebUIGeneralContext.php
- [x] WebUIFilesContext.php
## Related Issue

- Working on issue: https://github.com/owncloud/core/issues/39038
- Complete lists: https://github.com/owncloud/core/issues/39243

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebUIFilesContext.php
+++ b/tests/acceptance/features/bootstrap/WebUIFilesContext.php
@@ -26,6 +26,9 @@ use Behat\Gherkin\Node\TableNode;
 use Behat\MinkExtension\Context\RawMinkContext;
 use Page\FavoritesPage;
 use Page\FilesPage;
+use Page\FilesPageBasic;
+use Page\FilesPageElement\DetailsDialog;
+use Page\FilesPageElement\FileRow;
 use Page\OwncloudPage;
 use Page\SharedByLinkPage;
 use Page\SharedWithOthersPage;
@@ -196,7 +199,7 @@ class WebUIFilesContext extends RawMinkContext implements Context {
 	 *
 	 * @return OwncloudPage
 	 */
-	private function getCurrentPageObject() {
+	private function getCurrentPageObject():?OwncloudPage {
 		$pageObject = $this->webUIGeneralContext->getCurrentPageObject();
 		if ($pageObject === null) {
 			$pageObject = $this->filesPage;
@@ -209,7 +212,7 @@ class WebUIFilesContext extends RawMinkContext implements Context {
 	 *
 	 * @return string
 	 */
-	private function getCurrentFolderFilePath() {
+	private function getCurrentFolderFilePath():string {
 		return \rtrim($this->currentFolder, '/') . '/' . $this->currentFile;
 	}
 
@@ -219,7 +222,7 @@ class WebUIFilesContext extends RawMinkContext implements Context {
 	 *
 	 * @return void
 	 */
-	public function resetFilesContext() {
+	public function resetFilesContext():void {
 		$this->currentFolder = "";
 		$this->currentFile = "";
 		$this->deletedElementsTable = null;
@@ -231,9 +234,9 @@ class WebUIFilesContext extends RawMinkContext implements Context {
 	 * @Given the user has browsed to the files page
 	 *
 	 * @return void
-	 * @throws \Exception
+	 * @throws Exception
 	 */
-	public function theUserBrowsesToTheFilesPage() {
+	public function theUserBrowsesToTheFilesPage():void {
 		$this->theUserBrowsesToThePage($this->filesPage);
 	}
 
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/bootstrap/WebUIGeneralContext.php
```diff
@@ -167,15 +167,15 @@ public function __construct(
...
 	}
 
 	/**
 	 *
 	 * @return OwncloudPage
 	 */
-	public function getCurrentPageObject() {
+	public function getCurrentPageObject():OwncloudPage {
 		return $this->currentPageObject;
 	}
```
sakshamgurung, line 181: Giving return type ```string``` in ```public function getCurrentServer():string {...}``` and running test that uses this function gives error: ```Return value of WebUIGeneralContext::getCurrentServer() must be of the type string, null returned```

sakshamgurung:
Some common error after declaring param and return type of function in WebUIGeneralContext.php
- ```When the user reloads the current page of the webUI```
```Type error: Return value of WebUIGeneralContext::getCurrentPageObject() must be an instance of Page\OwncloudPage, null returned (Behat\Testwork\Call\Exception\FatalThrowableError)```
- ```Given the setting "([^"]*)" in the section "([^"]*)" has been (disabled/enabled)```
```Type error: Return value of WebUIGeneralContext::getSavedCapabilitiesXml() must be of the type string, array returned (Behat\Testwork\Call\Exception\FatalThrowableError)```

I will try to solve it, but if anyone has an idea then reply below.

**Comment by kiranparajuli589** [comment]
is it possible to set return type like (string|null) 🤔 ?? that may solve your problem

**Comment by saw-jan** [comment]
> is it possible to set return type like (string|null)  ?? that may solve your problem

yeah, I think it is possible with PHP 8

**Comment by sakshamgurung** [comment]
> > is it possible to set return type like (string|null)  ?? that may solve your problem
> 
> yeah, I think it is possible with PHP 8

It is not possible below PHP 8

**Comment by phil-davis** [comment]
> Are we using PHP 8 in drone and project?

oC10 only supports up to PHP 7.4. The Behat test runner will work with PHP 8.0, so we could run the acceptance test step with PHP 8.0 and the server-under-test with PHP 7.4. But that will be a real pain for testing locally - you would have to sort out how to run the oC10 server with PHP 7.4 and switch to using PHP 8.0 in the terminal window where you run Behat.

For now, it is best to use only PHP 7.4 features. Usually we are happy to run everything locally on PHP 7.4. PHP 7.2 is about to be dropped, and if we want to test the oC10 server on PHP 7.3, then we can run it in CI.

File: tests/acceptance/features/bootstrap/WebUIGeneralContext.php [hunk 2]
```diff
@@ -642,9 +642,12 @@ public function theUserReloadsTheCurrentPageOfTheWebUI() {
 	/**
 	 * returns the saved capabilities as XML
 	 *
-	 * @return string
+	 * @return array
 	 */
-	public function getSavedCapabilitiesXml() {
+	public function getSavedCapabilitiesXml():array {
+		echo "Hello there: ";
+		print_r($this->savedCapabilitiesXml);
```
sakshamgurung, line 649: I will remove this in next push

tests/acceptance/features/bootstrap/WebUIFilesContext.php
```diff
@@ -2269,9 +2274,9 @@ public function theUserOpensTheFileActionMenuOfFileFolderOnTheWebui($name) {
 	 *
 	 * @param string $fileName Name of the file/Folder
 	 *
-	 * @return ResponseInterface
+	 * @return void
```
sakshamgurung, line 2277: There was no return from this function. So, changed it to `void`. If it is suppose to return any thing please inform.

phil-davis:
https://drone.owncloud.com/owncloud/core/32045/3/6
Fails code style.

tests/acceptance/features/bootstrap/WebUIFilesContext.php
phil-davis: That is correct - there is no return value. The response is saved in the context object(s).

phil-davis (approved):
LGTM if CI passes.

phil-davis:
https://drone.owncloud.com/owncloud/core/32050/154/15
```
Type error: Argument 1 passed to WebUIFilesContext::theUserBrowsesDirectlyToDetailsTabOfFileInFolder() must be of the type string, null given, called in /drone/src/tests/acceptance/features/bootstrap/WebUIFilesContext.php on line 361 (Behat\Testwork\Call\Exception\FatalThrowableError)
```

PR #39155, opened 2021-08-31
Keep pagination active with empty search queries

JammingBen:
## Description
Before this fix, an empty search string would pre-render all rows in the file list, ignoring the pagination. This fix ensures that the file list is paginated correctly in combination with an empty search query.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4615

## How Has This Been Tested?
- Upload a folder with 1000 files
- Enter a string in the search box
- Wait until results are loaded (doesn't matter if there are any) and remove the string again
- Pagination should now still work

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/filelist.js
+++ b/apps/files/js/filelist.js
@@ -2896,7 +2896,10 @@
 			var $trs = this.$fileList.find('tr');
 			do {
 				_.each($trs, filterRows);
-				if (visibleCount < total) {
+
+				// prerender the file list until all items are found.
+				// skip this with an empty filter to keep the pagination.
+				if (visibleCount < total && filter !== '') {
 					$trs = this._nextPage(false);
 				}
 			} while (visibleCount < total && $trs.length > 0);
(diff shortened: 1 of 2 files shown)


apps/files/js/filelist.js
```diff
@@ -2896,7 +2896,10 @@
 			var $trs = this.$fileList.find('tr');
 			do {
 				_.each($trs, filterRows);
-				if (visibleCount < total) {
+
+				// prerender the file list until all items are found.
+				// skip this with an empty filter to keep the pagination.
+				if (visibleCount < total && filter !== '') {
```
JanAckermann, line 2902: I didn't try yet, but what is with blank space, maybe we can make use of trim as well?

JammingBen: I'm not sure about this, because you can actually search for a blank space. It will result in all files including such. I don't want to change the behavior here.

AlexAndBear (approved):
Much fluffy :1st_place_medal:

PR #39162, opened 2021-09-02
Fix file conflict dialog when dragging a file into a folder

### LINKED ISSUE #39133: Drag-and-drop a file into a folder while the same file exists in parent shows file conflict
### Steps to reproduce
1. Upload file "test.txt" to your root dir
2. Create an empty folder "folder" in your root dir
3. Drag-and-drop "test.txt" from your computer's explorer/finder onto "folder", resulting in a new upload

### Expected behaviour
The file should be uploaded without any issues.

### Actual behaviour
The UI shows the "file-conflict" dialog saying the file already exists; which is not true as "folder" is empty.

JammingBen:
## Description
When dragging a file into a folder, while another file with an identical name exists in the parent directory, the UI falsely showed a conflict dialog alert.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/39133

## How Has This Been Tested?

- Upload file "test.txt" to your root dir
- Create an empty folder "folder" in your root dir
- Drag-and-drop "test.txt" from your computer's explorer/finder onto "folder", resulting in a new upload
- -> The file should be uploaded without any issues.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -822,8 +822,9 @@ OC.Uploader.prototype = _.extend({
 		// only keep non-conflicting uploads
 		selection.uploads = _.filter(selection.uploads, function(upload) {
 			var file = upload.getFile();
-			if (file.relativePath || upload.getTargetFolder() !== '/') {
-				// can't check in subfolder contents
+			var targetUploadFolder = upload.getTargetFolder();
+			if (file.relativePath || (targetUploadFolder && targetUploadFolder !== '/')) {
+				// can't check in subfolder contents, let backend handle this
 				return true;
 			}
 			if (!fileList) {
@@ -997,6 +998,7 @@ OC.Uploader.prototype = _.extend({
 				 */
 				add: function(e, data) {
 					self.log('add', e, data);
+					console.log(data.targetDir);
 					var that = $(this), freeSpace;
 
 					var upload = new OC.FileUpload(self, data);
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -818,10 +818,11 @@ OC.Uploader.prototype = _.extend({
 	checkExistingFiles: function (selection, callbacks) {
 		var fileList = this.fileList;
 		var conflicts = [];
+
 		// only keep non-conflicting uploads
 		selection.uploads = _.filter(selection.uploads, function(upload) {
 			var file = upload.getFile();
-			if (file.relativePath) {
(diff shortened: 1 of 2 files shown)


phil-davis:
Works now for the case:
- Alice has folder "top" and file "test.txt" in the root of the file system
- Alice drags-and-drops another file called "test.txt" from her local system on top of folder "top"
- folder "top" now has that "test.txt" in it - good.

The next thing is more tricky (and probably was tricky already before this PR):
- Alice edits the local "test.txt"
- Alice drags-and-drops the changes "test.txt" from her local system on top of folder "top"
- Alice is shown a dialog titled "One file conflict" and asked "Which files do you want to keep?" - good
- But the dialog mentions folder "top" as the "already existing file" and shows a "text file" icon for the "new file", but there is no mention of "test.txt" - not so good, see screenshot:

Drag-and-drop-into-folder

- if Alice selects the new file and "Continue" then `top/test.txt` does get updated with the new data - good.
- if Alice selects the "already existing file" (the folder "top") and "Continue" then no changes happen - good.

So the real behavior is OK. It is just that the "One file conflict" dialog should mention `top/test.txt` as being the existing file.

Is there something that can easily be done about that?

phil-davis:
Note: the "will have a number added to its name" thing works - if I select both "files" and "Continue" then I end up with both `test.txt` and `test (2).txt` inside folder `top` - good

JammingBen:
Much thanks for the detailled testing! 👍 

> But the dialog mentions folder "top" as the "already existing file" and shows a "text file" icon for the "new file", but there is no mention of "test.txt" - not so good, see screenshot:

Yes, we also observed this. Unfortunately there is no quick fix for this, because the frontend would need to scan the folder content. Also what happens if you drag a folder into the folder? It would require quite some work and thought either way. Therefore, and because it's "only" an UI issue, we decided to leave this behavior as it is right now.

phil-davis:
> Therefore, and because it's "only" an UI issue, we decided to leave this behavior as it is right now.

That's fine with me. Sorting all that out is a separate issue. The changes here do not make that any different than it already is.

phil-davis (approved):
LGTM and works.
Note: we can't automated-test this, because Behat+selenium has no access to do an "external" select, drag, drop of a file from a "file explorer" running outside the browser window.

jnweiger:
Confirmed in 10.9.0-beta1

PR #39168, opened 2021-09-03
Return proper error if part file name is too long

JammingBen:
## Description
This is an alternative, less destructive approach than https://github.com/owncloud/core/pull/39088.

With this implementation, we keep the filename limit for part file names and just return a proper error. That means, if you upload a file with 250 chars (filesystem limit is usually 255), then you still get an "filename too long" error because of the part file name added by ownCloud.

This PR also reverts https://github.com/owncloud/core/pull/39088 therefore.

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/4692

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/3994


JammingBen:
@mrow4a @DeepDiver1975 As discussed, here is the alternative approach for our part file name issue. Please have a look.

JammingBen:
Please note that by setting `part_file_in_storage` to `false`, part files will still get hashed (see https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/File.php#L310)! That means this setting is basically broken because, once again, several apps cannot work with a hashed part file name yet. https://github.com/owncloud/core/pull/39131 would solve this problem though.

DeepDiver1975:
I see no change from 255 to 250 in this pr.
Am I missing anything? Thx

JammingBen:
Changing this won't affect anything, because the part file is being written to the filesystem before the length check of the original file. Thus we just added a length check for the part file itself.

changelog/unreleased/39088
```diff
@@ -1,10 +0,0 @@
-Bugfix: Hash chunk(v2) filename
```
mrow4a, line 1: I think we still need changelog, don't we?

mrow4a (commented):
add quick unit test

DeepDiver1975 (approved):

ownclouders:

https://drone.owncloud.com/owncloud/core/32162/53/1

mrow4a (approved):

config/config.sample.php
```diff
@@ -1498,6 +1498,9 @@
  * same storage as the upload target. Setting this to false will store the part
  * files in the root of the users folder which might be required to work with certain
  * external storage setups that have limited rename capabilities.
+ *
+ * Note that this setting causes issues with the following apps: Encryption,
```
mmattel, line 1502: It is not clear if the issue is when the setting is default (true) or when changing it to false.

phil-davis:
https://drone.owncloud.com/owncloud/core/32163
```
There was 1 failure:

1) OCA\DAV\Tests\unit\Connector\Sabre\FileTest::testPutTooLongPartFileName
Failed asserting that exception of type "OCA\DAV\Connector\Sabre\Exception\FileNameTooLong" is thrown.
```

https://drone.owncloud.com/owncloud/core/32163/32/10
```
There was 1 failure:

1) OCA\DAV\Tests\unit\Connector\Sabre\FileTest::testPutTooLongPartFileName
Failed asserting that exception of type "OCA\DAV\Connector\Sabre\Exception\FileNameTooLong" is thrown.
```

These are both errors when running with `scality` storage. My guess is that `scality` storage allows longer file names, and so the exception is not thrown in that case.

I am not sure how you can get the file system max filename length from within PHP. If you could do that, then the test could generate a filename a bit longer than the maximum allowed by the current storage. Or maybe tag this test case to skip on scality (however that is done in the unit tests).

mmattel:
> I am not sure how you can get the file system max filename length from within PHP

Maybe this can help a bit...
https://doc.owncloud.com/server/admin_manual/troubleshooting/path_filename_length.html

JammingBen:
This is not the issue I think as the character limit is hard coded: https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Common.php#L507. It rather seems that the scality storage does not use part files at all, therefore no exception. Let's see if it succeeds now.

phil-davis:
All the failing CI in oC10 apps is passing now. This core change was "a good thing".

jnweiger:
### Tested with owncloud core 10.9.0 beta1

228 characters is the limit on my https://oc1090beta1-ldap-0160-rc2-20211126.jw-qa.owncloud.works testsystem. A file with 229 chars triggers a "filename too long" error. OKayish.

PR #39171, opened 2021-09-03
Allow checkouts done without .git

jnweiger:
Also documented accepted inputs to the sed monster and expected. 
Feel free to convert that into a unit test :-)


phil-davis:
@jnweiger is this something that is still wanted/needed?

DeepDiver1975:
old aka dead -> close

PR #33665, opened 2018-11-26
Feature/yarn stable10

DeepDiver1975:
## Description
Finally use yarn for front end dependencies in stable10 for 10.1

In order to fix files which are moved into the release tar ball a modclean pattern plugin has been developed and is applied

Versions have been used as per current stable 10 - as far as possible ..

## Related Issue
- Fixes #33651 

## Motivation and Context
bower is dead - only package.json is monitored by sec tools like dependebot and snyk

## How Has This Been Tested?
- :hand: 

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Technical debt
- [ ] Tests

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link>

PR #39172, opened 2021-09-03
Fix Makefile for qa-master

jnweiger:
The Makefile target `$(dist_dir)/qa/owncloud`
differs from target `$(dist_dir)/owncloud`
where it should not. 

Only the latter runs yarn clean-modules. This seems needed in both cases -- at least it helps me to avoid a 'too many levels of symbolic link' error
caused by a cyclic symlink core/vendor/moment/meteor/moment.js when building the daily qa master tar.

Not sure if that only covers symptoms of something else, please review.


phil-davis (approved):
LGTM - looks like this was accidentally missed in PR #33665

PR #39175, opened 2021-09-06
[tests-only] Use headless Firefox for js tests instead of PhantomJS

JammingBen:
## Description
Replace the deprecated PhantomJS with a headless Firefox browser to run JS karma tests. This also should improve performance for these tests.

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1162,10 +1162,10 @@ def javascript(ctx, withCoverage):
                         "from_secret": "cache_s3_endpoint",
                     },
                     "bucket": "cache",
-                    "source": "tests/output/coverage/PhantomJS 2.1.1 (Linux 0.0.0)/lcov.info",
+                    "source": "tests/output/coverage/Firefox 91.0.0 (Ubuntu 0.0.0)/lcov.info",
                     "target": "%s/%s/coverage" % (ctx.repo.slug, ctx.build.commit + "-${DRONE_BUILD_NUMBER}"),
                     "path_style": True,
-                    "strip_prefix": "tests/output/coverage/PhantomJS 2.1.1 (Linux 0.0.0)",
+                    "strip_prefix": "tests/output/coverage/Firefox 91.0.0 (Ubuntu 0.0.0)",
                     "access_key": {
                         "from_secret": "cache_s3_access_key",
                     },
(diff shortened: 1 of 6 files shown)


JammingBen:
@phil-davis @micbar Any objections using Chromium as headless browser for running JS tests? We often run these and the performance with PhantomJS is not the best, especially on Mac M1. PhantomJS is discontinued anyways. It would require that Chromium is installed on the `owncloudci` docker image though.

If we want to keep the docker image we could at least make it configurable so devs can use it.

micbar:
speaks nothing against it.

Can you add?

JammingBen:
@micbar Sure, but I couldn't find the `owncloudci` in our docker repos. Could you please send me a link where to find it?

phil-davis:
@JammingBen the GitHub organisation has a dash in it - https://github.com/owncloud-ci/ - then each repo name matches the docker name.

JammingBen:
MIchael already sent me the link, thanks 👍 

However, I had another thought on this. Maybe it's not ideal to force Chrome or Firefox as dependency. What do you think about leaving PhantomJS as default but give developers the option to provide another browser via `make test-js TEST_KARMA_BROWSER=Fifrefox` @phil-davis ?

phil-davis:
@JammingBen Chrome usually works well (as god as or better than other browsers) with UI test frameworks. I don't see a problem with switching to Chrome for all regular CI. Adding it to the docker image should be fine - it just means that we should update the Chrome version in docker "every so often".

Firefox could also be added, if you like - but not super-critical.

JammingBen:
I tried locally with the nodejs image, installing Chrome is a bit ugly because of the underlying Ubuntu image. It didn't get it to run at a first glance, Firefox was no problem though (by just adding `firefox` to the list of dependencies that are installed via `apt`).

JammingBen:
I did some more tests, still think that Firefox is the easiest solution. Installing Chromium in Ubuntu requires `snap` nowadays, which then also needs to be installed and enabled.

Since I cannot push to the repo I forked it and created a PR from there: https://github.com/owncloud-ci/nodejs/pull/47

phil-davis:
> Since I cannot push to the repo I forked it and created a PR from there: owncloud-ci/nodejs#47

Approved and merged.

https://drone.owncloud.com/owncloud-ci/nodejs/144

Merge CI has finished. The docker images should now have Firefox.

phil-davis (approved):

phil-davis:
@micbar can you override the SonarCloud problem and merge please.

PR #39194, opened 2021-09-09
Unify API responses when setting permissions for public links

### LINKED ISSUE #36442: When 'Allow public uploads' is disabled cannot create share with permission 5 but creating share with permission 7 works (fallback to permission 1)

### LINKED ISSUE #36443: when "Allow public uploads" is disabled API responds with different messages, depending on the permissions
### Steps to reproduce
1. disable "Allow public uploads"
2. try to create public shares with different permissions
3. try to update existing shares with different permissions

### Expected behaviour
same message in every case

### Actual behaviour
create share with permission 5: message: `Public upload not allowed` statusCode: 403
create share with permission 7: share is created with permission 1 see #36442
create share with permission 4: message `Public upload disabled by the administrator` statusCode: 403
update existing public share setting permission to 5: message: `Public upload not allowed` statusCode: 400
update existing public share setting permission to 4,7,15: message: `Public upload disabled by the administrator` statusCode: 403

This issue results in different error messages in Phoenix, depending on the selected permission
image

image

JammingBen:
## Description
Setting (and changing) the permissions of public links via the OCS API will now return proper and unified API responses. Adding create permissions while public uploading is disabled globally will always return a 403 response.

## Before & after changes

The following table shows the changes before and after this PR. The rows highlighted in bold are the scenarios which now return a different response.

`public uploads allowed`

Action | response before | response after
------ | ------------- | -----------
Create link for folder with permission 1 | 200 | 200
Create link for folder with permission 4 | 200 | 200
Create link for folder with permission 5 | 200 | 200
Create link for folder with permission 7 | 200 | 200
Create link for folder with permission 15 | 200 | 200
Create link for folder with permission 31 | 200 | 200
Create link for file with permission 1 | 200 | 200
Create link for file with permission 4 | 404 | 404
Create link for file with permission 5 | 200 (will default to permission 1) | 200 (will default to permission 1)
Create link for file with permission 7 | 200 (will default to permission 1) | 200 (will default to permission 1)
Create link for file with permission 15 | 200 (will default to permission 3) | 200 (will default to permission 3)
Create link for file with permission 31 | 200 (will default to permission 1) | 200 (will default to permission 1)
Update link for folder with permission 1 | 200 | 200
Update link for folder with permission 4 | 200 | 200
Update link for folder with permission 5 | 200 | 200
Update link for folder with permission 7 | 200 | 200
Update link for folder with permission 15 | 200 | 200
Update link for folder with permission 31 | 400 | 400
Update link for file with permission 1 | 200 | 200
Update link for file with permission 4 | 400 | 400
**Update link for file with permission 5** | **200 (will default to permission 1)** | **400**
Update link for file with permission 7 | 400 | 400
Update link for file with permission 15 | 400 | 400
Update link for file with permission 31 | 400 | 400

`public uploads not allowed`

Action | response before | response after
------ | ------------- | -----------
Create link for folder with permission 1 | 200 | 200
Create link for folder with permission 4 | 403 | 403
Create link for folder with permission 5 | 403 | 403
**Create link for folder with permission 7** | **200 (will default to permission 1)** | **403**
Create link for folder with permission 15 | 403 | 403
**Create link for folder with permission 31** | **200  (will default to permission 1)**  | **403**
Create link for file with permission 1 | 200 | 200
Create link for file with permission 4 | 403 | 403
**Create link for file with permission 5** | **200 (will default to permission 1)** | **403**
**Create link for file with permission 7** | **200 (will default to permission 1)** | **403**
Create link for file with permission 15 | 403 | 403
**Create link for file with permission 31** | **200 (will default to permission 1)** | **403**
Update link for folder with permission 1 | 200 | 200
Update link for folder with permission 4 | 403 | 403
**Update link for folder with permission 5** | **400** | **403**
Update link for folder with permission 7 | 403 | 403
Update link for folder with permission 15 | 403 | 403
Update link for folder with permission 31 | 400 | 400
Update link for file with permission 1 | 200 | 200
Update link for file with permission 4 | 403 | 403
**Update link for file with permission 5** | **200 (will default to permission 1)** | **403**
Update link for file with permission 7 | 403 | 403
Update link for file with permission 15 | 403 | 403
Update link for file with permission 31 | 400 | 400

## Related Issue
- Fixes https://github.com/owncloud/core/issues/36442
- Fixes https://github.com/owncloud/core/issues/36443

## How Has This Been Tested?
See reproduction steps in the above mentioned tickets.

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


ownclouders:
:boom: Acceptance tests pipeline <strong>apiSharePublicLink2-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32228/71/1

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -385,10 +385,12 @@ public function createShare() {
 		}
 
 		$shareType = (int)$this->request->getParam('shareType', '-1');
+		$defaultPermissionsUsed = false;
 
 		// Parse permissions (if available)
 		$permissions = $this->getPermissionsFromRequest();
 		if ($permissions === null) {
+			$defaultPermissionsUsed = true;
```
jvillafanez, line 393: I'm not sure if this is accurate. There are changes in the permissions below, so if we assume that the `Constants::PERMISSION_ALL` is the default and then the permission changes (for example, in the external storage below), then the permission won't be the default any longer.

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -490,6 +494,13 @@ public function createShare() {
 				}
 			}
 
+			// don't allow "create"-permission if public upload is not allowed.
+			// we only need this check if we're not dealing with the default permissions.
+			if (!$defaultPermissionsUsed && !$publicUploadAllowed && $permissions & Constants::PERMISSION_CREATE) {
```
jvillafanez, line 499: Maybe move the `$permissions & Constants::PERMISSION_CREATE` up into a variable? I don't think that having a `&` is common inside conditions, so it could be misread or cause some confusion with `&&`

apps/files_sharing/lib/Controller/Share20OcsController.php
```diff
@@ -873,9 +884,9 @@ public function updateShare($id) {
 				}
 			}
 
-			// create-only (upload-only)
+			// create (upload)
 			if (
-				$newPermissions === Constants::PERMISSION_CREATE
+				$newPermissions & Constants::PERMISSION_CREATE
```
jvillafanez, line 889: The `===` might be intentional. I think the public uploads only allows uploading files there without knowing what's inside.

JammingBen: That's right, but either way, you can only set a `create` permission if 1. public uploading is allowed and 2. the resource is a folder, not a file.

apps/files_sharing/lib/Controller/Share20OcsController.php
JammingBen: `$defaultPermissionsUsed` is more meant to be in the sense of "no permissions given by the user via request". I renamed the variable to make that more clear.

jvillafanez:
Code-wise it looks fine, but I'm still a bit unsure with the change of the behavior for the public uploads.
Maybe either QA or PM could verify if the behavior is right and there isn't any side effect.

tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
```diff
@@ -293,64 +293,40 @@ Feature: create a public link share
...
-      | id          | A_STRING    |
-      | share_type  | public_link |
-      | permissions | read        |
-    And the OCS status code should be "<ocs_status_code>"
-    And the HTTP status code should be "<http_status_code>"
-    And the public upload to the last publicly shared folder using the <webdav_api_version> public WebDAV API should fail with HTTP status code "403"
+    Then the OCS status code should be "<ocs_status_code>"
 
-    @notToImplementOnOCIS @issue-ocis-2079 @issue-36442 @skipOnOcV10
+    @notToImplementOnOCIS @issue-ocis-2079
```
phil-davis, line 306:
```suggestion

```

These tags can be removed. The adjusted test scenario should pass on oCIS (and should be fixed on oCIS if it does not pass)

tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
```diff
@@ -293,64 +293,40 @@ Feature: create a public link share
...
-      | id          | A_STRING    |
-      | share_type  | public_link |
-      | permissions | read        |
-    And the OCS status code should be "<ocs_status_code>"
-    And the HTTP status code should be "<http_status_code>"
-    And the public upload to the last publicly shared folder using the <webdav_api_version> public WebDAV API should fail with HTTP status code "403"
+    Then the OCS status code should be "<ocs_status_code>"
 
-    @notToImplementOnOCIS @issue-ocis-2079
+    @notToImplementOnOCIS @issue-ocis-2079 @issue-ocis-reva-41
```
phil-davis, line 322:
```suggestion

```

These tags can be removed. The adjusted test scenario should pass on oCIS (and should be fixed on oCIS if it does not pass)

tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
```diff
@@ -359,91 +335,31 @@ Feature: create a public link share
...
```
phil-davis, line 340:
```suggestion

```

These tags can be removed. The adjusted test scenario should pass on oCIS (and should be fixed on oCIS if it does not pass)

tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
```diff
@@ -359,91 +335,31 @@ Feature: create a public link share
...
       | permissions | read,update,create |
     Then the OCS status code should be "<ocs_status_code>"
-    And the HTTP status code should be "200"
-    And the fields of the last response to user "Alice" should include
-      | id          | A_STRING    |
-      | share_type  | public_link |
-      | permissions | read        |
-    And the public upload to the last publicly shared folder using the <webdav_api_version> public WebDAV API should fail with HTTP status code "403"
 
     @notToImplementOnOCIS @issue-ocis-2079
```
phil-davis, line 357:
```suggestion

```

These tags can be removed. The adjusted test scenario should pass on oCIS (and should be fixed on oCIS if it does not pass)

JammingBen:
I added a before & after-table in the first comment of this PR. Although some return codes still seem kinda random, it's much more streamlined with this "fix".

@pmaier1 @micbar When public share uploads are not allowed via config, I would expect an API request with create permissions to return a forbidden response. That goes for the create permissions alone, but also for combinations like read+create+update. Is my assumption correct?

JammingBen:
Now that 10.9 is out, I'd like to catch up on this again. @pmaier1 What do you think about it? To summarize it:

The API responses when creating/editing public links are kinda random at the moment. Like for example when public uploading is forbidden, creating a link with the permissions `create` and `read` will throw a `403`, whereas creating a link with permissions `create`, `read` and `update` will succeed, but silently converted to `read` only. The table in the top post gives detailed information about what has changed.

Assuming the clients are fine with it (which I will re-test), I think it would be good to streamline that.

phil-davis:
It will be nice to get this more consistent. I expect (hope) that some API test combinations will fail and need to be adjusted to match the changed responses.

JammingBen:
> I expect (hope) that some API test combinations will fail and need to be adjusted to match the changed responses.

I adjusted some acceptance test scenarios to work with the new response code, pipeline seems to be fine with it now.

phil-davis (approved):

jvillafanez:
@mmattel in case there are docs to be updated.

jnweiger:
Confirmed fixed in 10.10.0 RC2

When Sharing settings are without "Allow public uploads"
* All combinations containing 4 (aka PERMISSION_CREATE) cause the same message "Public upload disabled by the administrator"
* I wonder what the meaning of 2 (aka PERMISSION_UPDATE) should be, when we don't allow public uploads....
  (inplace modification is *not* counted as upload?)

PR #39226, opened 2021-09-14
Move delete button in file list next to download button

AlexAndBear:
## Related Issue

- Fixes #39221 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39227, opened 2021-09-14
Provide mobile delete button

### LINKED ISSUE #39221: Unable to batch delete files and folders in mobile view. Test fails.
### Steps to reproduce
1. Go to `files` view
2. Resize the browser to mobile size (eg: 375 width by 812 height)
3. Batch select multiple files and folder
4. Only batch download is available no delete option

### Expected behaviour
After selecting multiple files and folder, delete option should be displayed in mobile view of webUI

### Actual behaviour
Only batch download option is shown in mobile view of webUI
Screenshot from 2021-09-13 17-21-22

### Server configuration
**Operating system**:
Ubuntu 20.04.3 LTS
**Web server:**
Apache
**Database:**
mysql  Ver 8.0.26-0ubuntu0.20.04.2 for Linux on x86_64 ((Ubuntu))
**PHP version:**
PHP 7.4.3
Latest master branch from github repo
Fresh install
Github

AlexAndBear:
## Description
Bugfix: Show missing delete button in file list on mobile view

## Related Issue

- Fixes  https://github.com/owncloud/core/issues/39221

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

### Before
image

### After
image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_trashbin/templates/index.php
+++ b/apps/files_trashbin/templates/index.php
@@ -34,8 +34,8 @@
 							<span><?php p($l->t('Restore'))?></span>
 						</a>
 							<a href="" class="delete-selected mobile">
-							<span><?php p($l->t('Delete'))?></span>
 							<span class="icon icon-delete"></span>
+							<span><?php p($l->t('Delete'))?></span>
 						</a>
 					</span>
 				</div>
@@ -44,8 +44,8 @@
 				<a id="modified" class="columntitle" data-sort="mtime"><span><?php p($l->t('Deleted')); ?></span><span class="sort-indicator"></span></a>
 				<span class="selectedActions">
 					<a href="" class="delete-selected">
-						<span><?php p($l->t('Delete'))?></span>
 						<span class="icon icon-delete"></span>
+						<span><?php p($l->t('Delete'))?></span>
 					</a>
 				</span>
 			</th>
--- a/apps/files_trashbin/templates/index.php
+++ b/apps/files_trashbin/templates/index.php
@@ -33,6 +33,10 @@
 							<span class="icon icon-history"></span>
 							<span><?php p($l->t('Restore'))?></span>
 						</a>
+							<a href="" class="delete-selected mobile">
+							<span><?php p($l->t('Delete'))?></span>
+							<span class="icon icon-delete"></span>
+						</a>
 					</span>
 				</div>
 			</th>
--- a/apps/files/css/files.css
+++ b/apps/files/css/files.css
@@ -696,6 +696,11 @@ a.action > img {
 	padding-right: 15px;
 }
 
+.selectedActions a.delete-selected.mobile {
+	display: none;
+	padding-right: 5px;
+}
+
 .selectedActions a.hidden {
 	display: none;
 }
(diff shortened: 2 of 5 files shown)


JanAckermann:
@phil-davis with the PR #39226 I moved the delete button next to the download button, @jvillafanez demur, that due to the empty space on the right side it doesn't look quite good, even if the useability would be better (All actions in one place).

I tried to move the actions quite next to the select box, but there is some weird magic in the code that expunges inner lying `a` tags to the parent element and some more stuff... I really don't want to touch that.

So this solution will now add a hidden delete button next to the download button, which will be visible in mobile view.

phil-davis:
@JanAckermann I noticed the same problem when trying to delete from the trashbin in the webUI. Select "Deleted files" then choose some files. The "Restore" selected icon is visible, but the "Delete" selected icon is not visible.

That could also be fixed?

JanAckermann:
@phil-davis If you agree with the overall strategy of adding an additional button for mobile (testing), yes I will fix it in trashbin as well.

phil-davis:
> Feels a bit hacky, but did not found a better solution

It works! I don't know what else to say. @jvillafanez might have an opinion.

JanAckermann:
Added to trashbin file list

apps/files_trashbin/templates/index.php
```diff
@@ -33,6 +33,10 @@
 							<span class="icon icon-history"></span>
 							<span><?php p($l->t('Restore'))?></span>
 						</a>
+							<a href="" class="delete-selected mobile">
+							<span><?php p($l->t('Delete'))?></span>
+							<span class="icon icon-delete"></span>
```
phil-davis, line 38: In ordinary files view the bin icon is to the left of the "Delete" text:
Files-delete-selected-mobile

In the Trashbin view the bin icon is to the right of the "Delete" text:
Trashbin-delete-selected-mobile

IMO these 2 lines need to be swapped.

phil-davis (approved):
works for me

jvillafanez (approved):
not fan of hacky solutions, but I don't think we have a better option without big changes in the UI

jvillafanez:
Screenshot from 2021-09-14 18-26-25

I'm not sure if it's worthy to adjust a bit the left margin of the header, it's looks ok on my mobile phone.

JanAckermann:
@jvillafanez we can try to make the margins smaller, but as usual this quite depends on the phone and the language.
So we don't solve the root of the problem.
Just that it works in the combination with your device...

OC is not designed to work on every mobile device.

Moreover we could try to hide the text and just show buttons.

I already put a lot of effort in make it more mobile friendly ☺️

JanAckermann:
@jvillafanez we can try to make the margins smaller, but as usual this quite depends on the phone and the language.
So we don't solve the root of the problem.
Just that it works in the combination with your device...

OC is not designed to work on every mobile device.

Moreover we could try to hide the text and just show buttons.

I already put a lot of effort in make it more mobile friendly ☺️

PR #39238, opened 2021-09-15
Set expire end of day in UI

### LINKED ISSUE #39234: Link expiration date in webUI Shared by Link is one day too short
### Steps to reproduce
1. make a public link that expires today (that means the end of today)
2.make a public link that expires tomorrow (that means the end of tomorrrow)
3. browse to "Shared by link"

### Expected behaviour
The link at point (1) should expire "in 10 hours" (not "14 hours ago")
The link at point (2) should expire in 1 day and 10 hours (whatever text the system generates for that - it might just say "1 day")

### Actual behaviour
shared-by-link-expiration-date

### Server configuration
Current core master.

This happens because the API returns just the expiration date with a "zero" time `00:00:00`. Because of backward-compatibility issues we could not fix that at the API level. So clients (like the webUI here) need to understand that the link is valid right up to and including 23:59:59 on the expiry day.

Please also have a look around the webUI code to see where else the expiration date is rendered in a way includes the time in the calculation. Any such places are likely to have a similar off-by-one-day error.

AlexAndBear:
## Description
Bugix: Show the correct expiring date in Shared by link files list

Before this PR the "Expiration date" column did not respect that shares 
expiration will be end of date and showing wrong values.
This has been fixed with this PR

## Related Issue

- Fixes https://github.com/owncloud/core/issues/39234

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/js/sharedfilelist.js
+++ b/apps/files_sharing/js/sharedfilelist.js
@@ -93,7 +93,7 @@
 			if (this._linksOnly) {
 				var expirationTimestamp = 0;
 				if(fileData.shares && fileData.shares[0].expiration !== null) {
-					expirationTimestamp = moment(fileData.shares[0].expiration).valueOf();
+					expirationTimestamp = moment(fileData.shares[0].expiration).endOf('day').valueOf();
 				}
 				$tr.attr('data-expiration', expirationTimestamp);
 
--- a/apps/files_sharing/tests/js/sharedfilelistSpec.js
+++ b/apps/files_sharing/tests/js/sharedfilelistSpec.js
@@ -614,7 +614,9 @@ describe('OCA.Sharing.FileList tests', function() {
 			fileList.reload();
 
 			var expirationDateInADay = moment()
-				.add(1, 'days').format('YYYY-MM-DD HH:mm:ss');
+				.set({"hour": 0, "minute": 0, "second": 0})
+				.add(1, 'days')
+				.format('YYYY-MM-DD HH:mm:ss');
 
 			/* jshint camelcase: false */
 			ocsResponse = {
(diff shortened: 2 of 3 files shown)


JanAckermann:
@phil-davis this would be a UI based solution, I can remeber that we wanted to change it in the backend, why did we discard it?

ownclouders:
:boom: Acceptance tests pipeline <strong>webUISharingPublic1-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32402/144/1

phil-davis:
> @phil-davis this would be a UI-based solution, I remember that we wanted to change it in the backend, why did we discard it?

The desktop client has a hardcoded matching of the time:
https://github.com/owncloud/client/blob/master/src/gui/sharemanager.cpp#L229
```
    if (data.value("expiration").isString()) {
        _expireDate = QDate::fromString(data.value("expiration").toString(), "yyyy-MM-dd 00:00:00");
    } else {
        _expireDate = value.toDate();
    }
```

So all desktop clients "in the wild" expect the time to always be "00:00:00". If the server changes that, it would be a major version change and would take years for every client everywhere to update.

phil-davis:
https://drone.owncloud.com/owncloud/core/32411/9/4
```
	Expected 'in 2 days' to equal 'in a day'.
	@apps/files_sharing/tests/js/sharedfilelistSpec.js:727:50
```

There is a JS unit test to adjust.

phil-davis (approved):
works - just needs the JS test expectation adjusted.

JanAckermann:
@phil-davis If adjust it to `in 2 days` my tests will fail, otherwise my tests run currently trough.
Expecting some timezone issues or so

phil-davis:
> Expecting some timezone issues or so

True. That might be painful. Have to sort out how that library decides how to do rounding and generate text "in 2 days" "in 3 days" relative to the current time.

phil-davis (approved):

PR #39198, opened 2021-09-09
Set files_list to empty array if we don't have any selected files

AlexAndBear:
## Description
Bugfix: Set file name to download.zip while downloading the whole user directory

Before this PR, selecting all files and folders in the webUI and clicking the
download button, resulted in a file downloaded with the name ".zip".
With this PR the file name is now "download.zip".

## Related Issue

- Fixes https://github.com/owncloud/core/issues/39187

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39239, opened 2021-09-15
Handle case if no files was supplied

AlexAndBear:
## Description
https://github.com/owncloud/core/pull/39198 introduced a bug, where the root dir folder download -> download.zip was empty.
Fixed with this PR

## Related Issue

https://central.owncloud.org/t/files-are-downloaded-as-a-file-without-extension/33798/13

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/legacy/files.php
+++ b/lib/private/legacy/files.php
@@ -111,6 +111,11 @@ class OC_Files {
 		}
 
 		try {
+			/*
+			 * $listOfFiles might be an empty array,
+			 * for example if the client requests to download a whole folder
+			 * and does not supply individual files.
+			 */
 			if (\count($listOfFiles) === 0) {
 				$filename = $dir;
 				$getType = self::ZIP_DIR;
--- a/lib/private/legacy/files.php
+++ b/lib/private/legacy/files.php
@@ -111,7 +111,11 @@ class OC_Files {
 		}
 
 		try {
-			if (\count($listOfFiles) === 1) {
+			if(\count($listOfFiles) === 0){
+				$filename = $dir;
+				$getType = self::ZIP_DIR;
+			}
+			else if (\count($listOfFiles) === 1) {
 				$filename = "{$dir}/{$listOfFiles[0]}";
 				if (!$view->is_dir($filename)) {
 					self::getSingleFile($view, $dir, $listOfFiles[0], $params === null ? [] : $params);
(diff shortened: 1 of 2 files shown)


lib/private/legacy/files.php
```diff
@@ -111,7 +111,11 @@ public static function get($dir, $files, $params = null) {
 		}
 
 		try {
-			if (\count($listOfFiles) === 1) {
+			if(\count($listOfFiles) === 0){
```
phil-davis:
```suggestion
			if (\count($listOfFiles) === 0){
```

phil-davis (commented):
Fails coding standard. But the fix works - I select all files/folders in the root of a user and `download/zip` actually has the files/folders in it.

phil-davis (approved):

lib/private/legacy/files.php
```diff
@@ -111,7 +111,10 @@ public static function get($dir, $files, $params = null) {
 		}
 
 		try {
-			if (\count($listOfFiles) === 1) {
+			if (\count($listOfFiles) === 0) {
```
JammingBen, line 114: Maybe add a comment on why `$listOfFile` could be empty and why we're handling this?

JammingBen (commented):
Code itself LGMT

lib/private/legacy/files.php
```diff
@@ -111,7 +111,10 @@ public static function get($dir, $files, $params = null) {
 		}
```
phil-davis, line 111: The indent displayed by GitHub is strange. This `}` ends the else of the if statement. But it is 1 tab less.

If you are touching this PR again, then maybe make a separate re-formatting commit too.

lib/private/legacy/files.php
```diff
@@ -101,7 +101,6 @@ private static function sendHeaders($filename, $name, array $rangeArray) {
 	 */
 	public static function get($dir, $files, $params = null) {
 		$view = \OC\Files\Filesystem::getView();
-		$getType = self::FILE;
```
JanAckermann, line 104: Note, that this was even not used before this PR

lib/private/legacy/files.php
JanAckermann: this is kind of weird

phil-davis: Not important.

PR #39241, opened 2021-09-15
[full-ci] Hide share owner file path for share receiver

### LINKED ISSUE #38027: full path of share at the sharer side is disclosed to share receiver when autoaccept is off
### Steps to reproduce
1. disable `Automatically accept new incoming local user shares`
2. create this folder structure: `PARENT/CHILD/child.txt`
3. share `child.txt` to an other user
4. as share receiver query received shares `curl -uadmin:admin 'http://localhost/owncloud-core/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json&shared_with_me=true&state=all' | jq` 

### Expected behaviour
receiver should not see where the sharer has stored the file, only `child.txt` should be listed as `path`

### Actual behaviour
the full path of the file at the sharer side is shown `"path": "/PARENT/CHILD/child.txt"`

JammingBen:
## Description
This fixes an issue where a share receiver could see the share owner's full file path before accepting the share.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/38027

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


JammingBen:
@phil-davis This change requires adjustments in some acceptance test scenarios. It affects those scenarios:

```
And user "Alice" has shared file "PARENT/parent.txt" with user "Brian"
And user "Brian" has accepted share "/PARENT/parent.txt" offered by user "Alice"
```

Which have to look like this now:

```
And user "Alice" has shared file "PARENT/parent.txt" with user "Brian"
And user "Brian" has accepted share "/parent.txt" offered by user "Alice"
```

Because the share receiver does not know anything about the share owners file path anymore. What do you think, let's take the effort?

On a side note: Actually, it should be `And user "Brian" has accepted share "/Shares/parent.txt" offered by user "Alice"` if the share folder is set to `Shares` (which seems to be the case for most acceptance tests). However, I adjusted the `userReactsToShareOfferedBy()` method to adjust this "on the fly".

ownclouders:
:boom: Acceptance tests pipeline <strong>apiProxySmoke-8-5-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32421/169/1

phil-davis:
> Because the share receiver does not know anything about the share owners file path anymore. What do you think, let's take the effort?

Yes, that is correct. We need to adjust the expectation of what Brian sees as the share being offered. The various test scenarios had been written with the "wrong" expectation because otherwise there would have been a whole lot of failures, and it was preferable that the scenarios continued on and demonstrated their later steps.

I will edit the PR title to say `full-ci` so that the next time you push a change, all the pipelines will run to completion and we can find these.

JammingBen:
The following 2 scenarios are still a bit difficult to solve:

`tests/acceptance/features/apiWebdavLocks3/independentLocksShareToShares.feature:33`

2 folders with the same name but different locations get shared with Brian:

```
And user "Alice" has shared folder "locked/toShare" with user "Brian"
And user "Alice" has shared folder "notlocked/toShare" with user "Brian"
```

Brian wants to accept both:

```
And user "Brian" has accepted share "/toShare" offered by user "Alice"
And user "Brian" has accepted share "/toShare" offered by user "Alice"
```

But the second attempt will accept the first share again, therefore the second share never gets accepted. I tried to filter out already accepted/declined shares in `userReactsToShareOfferedBy()`. But we have other tests which rely on accepting the same share twice. So that's not an option.

`tests/acceptance/features/apiShareManagementToShares/mergeShare.feature:89`

I don't unterstand this test. Alice shares `merge-test-outside-groups-renamebeforesecondshare` with Brian. Brian moves this file out of the Shares-folder and renames it to `merge-test-outside-groups-renamebeforesecondshare-renamed`. Now Alice shares `merge-test-outside-groups-renamebeforesecondshare-renamed` with Brian.

1. Why does Alice have `merge-test-outside-groups-renamebeforesecondshare-renamed`? I don't see this file being created for her.
2. Brian has only one received share in `userReactsToShareOfferedBy()`. IMO that's wrong. He should have two: `merge-test-outside-groups-renamebeforesecondshare` and `merge-test-outside-groups-renamebeforesecondshare-renamed`.

@phil-davis Any ideas here?

phil-davis:
I pushed a commit to fix independentLocksShareToShares.feature:33
I made a new step definition for that:
```
@Given /^user "([^"]*)" has accepted the (?:first|next|) pending share "([^"]*)" offered by user "([^"]*)"$/
```
And made the code only find the (first|next) share that is in the pending state. So in the scenario we can accept the "first pending share" and "the next pending share"...
In this step, "first" and "next" are just alternative words to use when convenient - the step does the same thing each time, it selects the currently-first matching share that is in the pending state.

phil-davis:
I fixed up the `apiShareManagementToShares/mergeShare.feature` scenarios. There were some existing steps that did not work. For example:
```
And user "Alice" shares folder "/merge-test-outside-groups-renamebeforesecondshare-renamed" with user "Brian" using the sharing API
```

actually Alice does not have a folder with that name. So the share did not even happen, and there was no `Then` step to check that the HTTP status was OK or anything. So actually no 2nd share happened. And the `Then` steps had been written in a way that was based only on the 1st share, and which pass anyway.

I fixed it up. The purpose of the scenario (and the one following it) is to demonstrate what happens when a user receives and accepts a share in one (as a user or as a group member), and moves/renames the received share to something else.

Then the user receives the same shared data again (as a group member or as a user) and accepts the share.

The user receives the data via 2 different "ways". But the server merges those "received views of the data" and presents the user with only a single copy.

In the case of these scenarios, the 1st moved/renamed folder "disappears" and the user sees the folder back where it came as the 2nd share that they accepted.

JammingBen:
@phil-davis Great, thank you! Pipeline passed, looks good.

Something to keep in mind: All these changed tests won't succeed on OC <= 10.8. Add skip tags for those versions? Would be a lot of skipped tests though... Maybe we can add different "value-tables" for older OC versions? However, then we would need a label like `skipNewerThanOc10.8`.

phil-davis:
> Add skip tags for those versions?

Yes, we will have to have skip tags for the "old" oC10 versions.

It's not easy to design a `skipOnAllVersionsGreaterThan` general tagging system that `tests/acceptance/run.sh` can "easily" process, because the pipeline has to generate all the possible old version tag combinations to pass in to the Behat command - `skipOnAllVersionsGreaterThan10.0.0` `skipOnAllVersionsGreaterThan10.0` `skipOnAllVersionsGreaterThan10.0.1` `skipOnAllVersionsGreaterThan10.0.2` `skipOnAllVersionsGreaterThan10.0.3` ... (up to what patch version might there be a tag?) `skipOnAllVersionsGreaterThan10.1` `skipOnAllVersionsGreaterThan10.2` `skipOnAllVersionsGreaterThan10.2.1` (which minor versions had multiple patch releases)...

We could have a special "hard-coded" tag just for this case, and CI against 10.9 and later could specifically skip it.

Is it worth it?

JammingBen:
> We could have a special "hard-coded" tag just for this case, and CI against 10.9 and later could specifically skip it.

Sounds like the "lest-effort-most-value" solution to me. Basically create such tags for versions where it is needed, such like in this case. Maybe we can also document this. Is there some kind of documentation for the tags? Otherwise a developer might suggest that this tag exists for all OC versions...

phil-davis:
> Maybe we can also document this. Is there some kind of documentation for the tags? Otherwise a developer might suggest that this tag exists for all OC versions...

https://doc.owncloud.com/server/next/developer_manual/testing/acceptance-tests.html#skip-tests-on-a-particular-version-of-owncloud

I can add something there, when we have an example solution here.

JammingBen:
Phew, that was quite some work... Let's see what the current pipeline says.

I'm not sure how OCIS behaves against these tests. It's possible that the `@skipOnOcis`-tags have to be switched in the newly introduced example-tables.

phil-davis:
> I'm not sure how OCIS behaves against these tests. It's possible that the @skipOnOcis-tags have to be switched in the newly introduced example-tables.

We want OCIS to run the new test scenarios, and then deal with making OCIS behavior compatible with that. I will review...

phil-davis (approved):
LGTM

phil-davis:
IMO this is "a good thing". I  will merge and then bump the core-commit-id in and see how much pain there is in expected-failures!

phil-davis:
@JammingBen this PR changes some core API behavior. 
FYI:
1) I had to adjust a couple of test scenarios in https://github.com/owncloud/admin_audit/pull/386
2) Issue https://github.com/owncloud/ocis/issues/2540 describes various differences between oC10 and oCIS and between various accept and decline actions.

I haven't looked in detail at (2) yet. But we need to see what needs adjusting in which of oC10 and oCIS (next week)

butonic:
@phil-davis @JammingBen IMNSHO the sharer should never see where a recipient mounted a share. It discloses the path parents that exists in the recipients trees to the sharer. AFAIK that information is neither used by the web ui nor the clients.

I would drop the `file_path` property when listing shares as the sharer. The `path` property is used everywhere.

jnweiger:
Confirmed fixed in 10.9.0 beta1

PR #39242, opened 2021-09-15
[full-ci] add percent to usernames

butonic:
we ran into this when trying a user that had an email as the username. ocis-web would encode the `@` as `%40` and the chi router would not match the route. See https://github.com/go-chi/chi/pull/659

### DIFF
--- a/tests/acceptance/features/apiProvisioning-v1/editUser.feature
+++ b/tests/acceptance/features/apiProvisioning-v1/editUser.feature
@@ -29,6 +29,11 @@ Feature: edit users
       | a@-+_.b  | a.b@example.com     |
       | a space  | a.space@example.com |
 
+    @skipOnOcV10
+    Examples:
+      | username | email           |
+      | a%b      | a.b@example.com |
+
   @smokeTest
   Scenario: the administrator can edit a user display (the API allows editing the "display name" by using the key word "display")
     Given user "brand-new-user" has been created with default attributes and without skeleton files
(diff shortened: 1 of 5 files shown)


phil-davis (commented):
There should be matching scenarios in of the apiProvisioning-v1/v2 suites.
One change has been made in apiProvisioning-v1 and 3 in apiProvisioning-v2.
Maybe there could/should be 4 changes in each?

ownclouders:

https://drone.owncloud.com/owncloud/core/32426/53/1

phil-davis:
The `%` sign is not allowed in a username in oC10:
```
{"reqId":"54sWVWbrPSx9rqEg7rqp","level":3,"time":"2021-09-16T02:14:36+00:00","remoteAddr":"192.168.1.64","user":"admin","app":"ocs_api","method":"POST","url":"\/ocs\/v2.php\/cloud\/users","message":"Failed addUser attempt with exception: Only the following characters are allowed in a username: \"a-z\", \"A-Z\", \"0-9\", and \"+_.@-'\""}
```

So we know that the test cases in the first commit will not pass.

I put the test cases for `%` in a username into separate examples tables that are skipped ownCloud 10. That will let the CI pass here.
But I am not sure what the actual requirement is. Are actual literal `%` signs allowed in "username" strings in oCIS? Or do we want to test something different?

phil-davis:
There was an unrelated fail - see issue #39244

DeepDiver1975:

PR #39248, opened 2021-09-16
[tests-only] check caldav carddav CI

phil-davis:
`python2` was added back to https://hub.docker.com/r/owncloudci/php just now.

This PR checks if the caldav-cardav CI works again.


phil-davis:
works

PR #39258, opened 2021-09-20
Allow the occ command to list users and groups matching a single character

### LINKED ISSUE #31117: occ user and group list cannot filter on short names
### Steps to reproduce
1. Create users ``a11`` ``a22`` ``a33`` and some other users not having an ``a`` in their UID
2. ``./occ user:list a1``
3. ``./occ user:list a``

### Expected behaviour
At step 2, user ``a11`` is listed.
At step 3, users ``a11`` ``a22`` ``a33`` are listed.

### Actual behaviour
No output.

### Server configuration
current core master or stable10

Similar issue with trying to filter the groups list.

phil-davis:
## Description
The `occ user:list` and `occ group:list` commands were being limited by the value of `user.search_min_length`. A command like `php occ user:list a` would return an empty list, even though there were usernames with the letter "a".

If `user.search_min_length` is set to a higher value (e.g. 4) then `php occ user:list abc` would return and empty list, instead of the users with "abc" in the username.

This behavior can be annoying at the command line. It is not obvious to the command-line user why no users are found.

There is no particular reason to limit the command-line matching of users. The command-line user can execute `php occ user:list` without parameters and get a full list of users anyway. The behavior was an "accident"  from the implementation of `user.search_min_length`.

This PR stops `user:list` and `group:list` from being limites by the setting of `user.search_min_length`

## Related Issue
- Fixes #31117 

## How Has This Been Tested?
CI - see new test cases

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/OccUsersGroupsContext.php
+++ b/tests/acceptance/features/bootstrap/OccUsersGroupsContext.php
@@ -501,6 +501,20 @@ class OccUsersGroupsContext implements Context {
 	}
 
 	/**
+	 * @When the administrator gets the groups containing :substring in the group name in JSON format using the occ command
+	 *
+	 * @param string $subString
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function theAdministratorGetsTheGroupsContainingInJsonUsingTheOccCommand(string $subString) {
+		$this->occContext->invokingTheCommand(
+			"group:list $subString --output=json"
+		);
+	}
+
+	/**
 	 * @When the administrator removes user :username from group :group using the occ command
 	 *
 	 * @param string $username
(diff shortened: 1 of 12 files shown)


tests/acceptance/features/bootstrap/OccUsersGroupsContext.php
```diff
@@ -578,7 +592,7 @@ public function theLanguageOfUserReturnedByTheOccCommandShouldBe($username, $lan
 	 * @throws Exception
 	 */
 	public function theUsersReturnedByTheOccCommandShouldBe(TableNode $useridTable) {
-		$this->featureContext->verifyTableNodeColumns($useridTable, ['uid', 'display name']);
+		$this->featureContext->verifyTableNodeColumns($useridTable, ['uid']);
```
phil-davis, line 595: Note: I noticed that the "display name" table column was not actually used in the test code. It got overwritten by a call to 
`$this->featureContext->getDisplayNameForUser($row['uid'])`

So I have removed it from the test step code and from the test steps where it is used.

JammingBen (approved):
LGTM

AlexAndBear (approved):

PR #39271, opened 2021-09-22
Allow txt previews for chinese, japanese, korean, arabic, devanagari langs

### LINKED ISSUE #16229: Upload txt file, File has Chinese, Garbled
When I upload txt file that contains Chinese when, after the online preview garbled, download files is garbled.

File contents:  传说中文会出现乱码
server：centos6.4
client：win7 chinese

AlexAndBear:
## Description
Enhancement: Previews for text files including non-latin characters

Generate and display previews for text files that include non-latin
characters like Chinese, Japanese, Korean and Arabic.

## Related Issue

- Fixes #16229 

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

### Before
image

### After
image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


JanAckermann:
@phil-davis here is a funny one ;)

@hodyroff FYI

JanAckermann:
Note: Cyrillic chars are already included in OpenSans

lib/private/Preview/TXT.php
```diff
@@ -94,4 +92,81 @@ public function getThumbnail(File $file, $maxX, $maxY, $scalingUp) {
...
+
+		if ($this->isKorean($text)) {
+			return $fontFileDir.'NotoSansCJKkr/NotoSansMonoCJKkr-Regular.otf';
+		}
+
+		if ($this->isArabic($text)) {
+			return $fontFileDir.'NotoSansArabic/NotoSansArabic-Regular.ttf';
+		}
+
+		return $fontFileDir.'OpenSans-Regular.ttf';
```
jvillafanez, line 130: I'd vote for using a NotoSans font instead in order to keep consistency.

IljaN (approved):

phil-davis:
नेपालि (Nepali) Devanagari script (same as Hindi) does not work. I guess that script/font used across North India and Nepal is not supported by this.

The Chinese works for me when I test it.

lib/private/Preview/TXT.php
```diff
@@ -94,4 +92,81 @@ public function getThumbnail(File $file, $maxX, $maxY, $scalingUp) {
...
+	}
+
+	/**
+	 * Detect whether string contains arabic letters
+	 *
+	 * @param string $text
+	 * @return bool
+	 */
+	public function isArabic($subject) {
+		return (preg_match('/[\x{0600}-\x{06FF}]/u', $subject) > 0);
```
jvillafanez, line 170: This is different check than the others. If it's intentional, a comment helps to understand that it isn't a bug.

In addition, I think it's a good idea to document the source of those codes. Without any additional info I don't know if the checks are correct, and looking if they're the right ones might take a lot of time.

jvillafanez:
I'm a bit worried that we have to compile up to 4 different regular expression and match them against a text of 3000 bytes (at least it isn't the whole file). If there isn't a better solution to distinguish the languages I guess we'll have to live with it.

lib/private/Preview/TXT.php
JanAckermann: Ok

lib/private/Preview/TXT.php
JanAckermann: What do you mean with source?

JanAckermann:
> I'm a bit worried that we have to compile up to 4 different regular expression and match them against a text of 3000 bytes (at least it isn't the whole file). If there isn't a better solution to distinguish the languages I guess we'll have to live with it.

There is none.

phil-davis (approved):

phil-davis:
I can push some links for those character ranges, and add Devanagari support - Hindi+Nepali and other North Indian languages. 

"What is the number of people who can read Roman script in India in comparison to Devanagari script?"
"Rough estimates are about 800–900 million people can read the Devanagari script and about 150–200 million can read Latin/Roman script. The reason for this is apart from Hindu numerous other Indian languages also use this script."

So that covers another >10% of world population.

phil-davis:
> I can push some links for those character ranges, and add Devanagari support - Hindi+Nepali and other North Indian languages.

Done. I can read the Nepali (Devanagari) script in previews of text files.

JanAckermann:
@phil-davis nice thx 🙌 thinking about adding heabraeic too...

lib/private/Preview/TXT.php
jvillafanez: > What do you mean with source?

I meant the web links used for reference. I see you've added them already.

jvillafanez: I see there is a `/\p{Han}/` for chinese... should we include similar properties for the rest of the languages? I was checking at https://www.php.net/manual/en/regexp.reference.unicode.php and we could use a bunch of them.

JanAckermann: Of course, feel free to commit :)

JanAckermann:
@jvillafanez we could add an pregmatch a-z,0-9 in the first place, which could improve the performance in first place

JanAckermann:
Hold back! Let's see if we can merge those fonts and go away from all this pregmatch stuff!

IljaN:
Would it be possible to merge the fonts some how so no Lang detection is needed?

kulmann:
> Hold back! Let's see if we can merge those fonts and go away from all this pregmatch stuff!

You could convert the PR to a draft, so that it doesn't get merged accidentally. ;-)

jvillafanez:
One question that just happened: what if we have a string with multiple languages in it? For example "Choose you font: latin, عرب, 中国人 , नेपाली " . We might choose a font that won't display some of the characters.

JanAckermann:
@jvillafanez Mentioned in the code as a comment, it might fail

phil-davis:
> One question that just happened: what if we have a string with multiple languages in it?

IMO the preview only contains some of the "early" text in the document. Maybe only a maximum of the first line? So instead of passing the whole content to `getFontFile`, we can just pass the first line (or just the first 20, 40? bytes, as long as that still forms a valid set of Unicode chars).

And `getFontFile` could become `getFontFiles` and return an array of fonts? Or return a merged font-set if any of the languages are detected?

Or we could just always provide all the fonts, or a single merged font-set? And not bother with any of the processing.

JanAckermann:
> single merged font-set?
Tried to merge those fonts its simply exceeding the max size of 65,535 glyphs so far

phil-davis:
> Well imagettftext only accept one particular font, what do we do with the font array?

That's a challenge!

In reality, there are not going to be many text documents that start with, for example, a few Japanese characters and then switch to Hindi in the very first line! I guess we pass in the first line and process that for a language match. Whichever is detected first by the code wins.

JanAckermann:
Here I found a comment, according why merging all the fonts causes trouble https://github.com/fonttools/fonttools/issues/743#issuecomment-262323134

Updated the PR, so we use NotoSans as the default font as well cutting the 10 first letters for detection

ownclouders:

https://drone.owncloud.com/owncloud/core/32619/166/1

phil-davis:
https://drone.owncloud.com/owncloud/core/32619/166/14
Silly drone agent pipeline startup problem - I restarted.

JanAckermann:
@phil-davis @jvillafanez 

I can merge NotoSans, NotoSansArabic, and NotoSansDevanagari, but I can't add a merged CJK.
So I wonder If merging is really an option as we might add more and more fonts and the developer needs to merge them until we hit the max size limit.

I am ok with the current implementation

jvillafanez:
> @jvillafanez we could add an pregmatch a-z,0-9 in the first place, which could improve the performance in first place

The idea I had was to process the string as a stream and analyze each utf8 char to gather some statistics about how many chars of specific unicode scripts are in the strings. For example, the result could be that the target string contains 312 balinese chars and 10 latin chars, or 217 han chars. This way we could decide what is the font that could fit better.

Anyway, this will take some work, so I'd prefer to keep it as an optimization option to be dealt in a different PR.

JanAckermann:
@jvillafanez good idea, one thing to keep in mind:

if we have a text like `aaa传` CJK font is able to render it since it has latin chars included, if we load the regular font this won't work.

jvillafanez:
> if we have a text like aaa传 CJK font is able to render it since it has latin chars included, if we load the regular font this won't work.
On the other hand arabic noto does not contain latin letts

Yes, that's something for the text preview to decide. If the text contains 1 chinese char and 3 latin chars, it's fine to choose the chinese font even though there are more latin chars.
Something similar for the arabic and latin. The text preview might choose to prioritize arabic, or choose the one with the greatest number of chars.

JanAckermann:
IHMO to deliver a really good solution, it is necessary, that we split every character and create an individual image and combine it after the processing, so we would be able to render:
`aa传说中文ضضضضضضضضضض😍`

PR #39282, opened 2021-09-23
[tests-only] [full-ci] Run tests with MariaDb 10.6

phil-davis:
see if there is even a docker image for MariaDb 10.6

Issue #39283

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1546,7 +1546,7 @@ def acceptance(ctx):
                                 if federatedDbName not in ["mariadb", "mysql"]:
                                     # Do not try to run 2 sets of Oracle, Postgres etc databases
                                     # When testing with these, let the federated server use mariadb
-                                    federatedDb = "mariadb:10.6"
+                                    federatedDb = "mariadb:10.5"
 
                                 if isWebUI or isAPI or isCLI:
                                     browserString = "" if browser == "" else "-" + browser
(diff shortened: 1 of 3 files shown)


IljaN:
Not sure where this is coming from: https://drone.owncloud.com/owncloud/core/32640/21/5

IljaN:
I guess @JanAckermann was correct, it looks to me as if doctrine is creating all tables with default compressed

IljaN:
Found another instance of "compressed": https://github.com/owncloud/core/pull/39282/commits/ec9474920181ad12afeba19653f2b72f0b21bb6d

IljaN:
Are there more configs which set row_format?

IljaN:
So unit tests pass with my commits on 10.6 now, but some acceptance tests are pulling the 10.7 tarball. What now @phil-davis ?

phil-davis:
> So unit tests pass with my commits on 10.6 now, but some acceptance tests are pulling the 10.7 tarball. What now @phil-davis ?

https://drone.owncloud.com/owncloud/core/32643/103/11 is using the "latest" tarball to install a federated server = 10.8.0

Actually we don't care what database the federated server is using - we just care that it is running the latest release, so that we test that the newer code in core master still talks correctly with a "latest" federated server.

I will change that back to MariaDb 10.5 for the federated server.

phil-davis:
That's good - all tests passed. For new installations it seems like this will be the only code changes needed.

We will have to investigate what issues there might be, if any, for existing sites that are on oC 10.8.0 or earlier, upgrade to a future 10.9.0 and then update MariaDb to 10.6.

phil-davis:
PR #39286 implements this and has been merged to master.

Issue #39283, opened 2021-09-23
MariaDb 10.6 stable


phil-davis:
has been released on 2021-07-06 
https://downloads.mariadb.org/mariadb/+releases/

Test with it and see if it works "out of the box" or not.

phil-davis:
First problem: https://drone.owncloud.com/owncloud/core/32635/50/8

+ php occ maintenance:install -vvv --database=mysql --database-name=owncloud --database-table-prefix=oc_ --admin-user=admin --admin-pass=admin --data-dir=/drone/src/data --database-host=mariadb --database-user=owncloud --database-pass=owncloud
Error while trying to create admin user: An exception occurred while executing 'CREATE TABLE `oc_appconfig` (`appid` VARCHAR(32) DEFAULT '' NOT NULL, `configkey` VARCHAR(64) DEFAULT '' NOT NULL, `configvalue` LONGTEXT DEFAULT NULL, INDEX appconfig_config_key_index (`configkey`), INDEX appconfig_appid_key (`appid`), PRIMARY KEY(`appid`, `configkey`)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_bin` ENGINE = InnoDB ROW_FORMAT = compressed':

SQLSTATE[HY000]: General error: 4047 InnoDB refuses to write tables with ROW_FORMAT=COMPRESSED or KEY_BLOCK_SIZE.

IljaN:
https://mariadb.com/kb/en/innodb-compressed-row-format/

IljaN:
If I understand correctly we should remove "COMPRESSED" from our migrations so that new systems are  10.6+ ready  in the future

IljaN:
Possible Migration for existing System:
https://github.com/nextcloud/docker/issues/1492#issuecomment-881216028

phil-davis:
I made a docs PR to specify the exact list of MariaDb 10.* versions that are currently supported. We can adjust that again after we know the result of this issue.

IljaN:
Do we need to adjust docker (compose) files somewhere? Or are those pinned to a specific version

phil-davis:
> Do we need to adjust docker (compose) files somewhere? Or are those pinned to a specific version?

https://drone.owncloud.com/owncloud/core/32635/21/2
`10.6: Pulling from library/mariadb`

MariaDb pulls from a docker image that is already provided, not by ownCloud. I guess that we need to look into how to control the settings in that image?

[...]

PR #39286, opened 2021-09-25
MariaDb changes for MariaDb 10.6

phil-davis:
## Description
These are the code changes cherry-picked from PR #39282 

They are code changes that make CI pass against MariaDb 10.6

This PR will confirm if these code changes are OK for the oldest supported MariaDb version 10.2.

## Related Issue
- part of #39283 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/4115

### DIFF
--- a/lib/private/DB/ConnectionFactory.php
+++ b/lib/private/DB/ConnectionFactory.php
@@ -218,7 +218,6 @@ class ConnectionFactory {
 			$connectionParams['defaultTableOptions'] = [
 				'collate' => 'utf8mb4_bin',
 				'charset' => 'utf8mb4',
-				'row_format' => 'compressed',
 				'tablePrefix' => $connectionParams['tablePrefix']
 			];
 		}
--- a/lib/private/Repair/Collation.php
+++ b/lib/private/Repair/Collation.php
@@ -66,11 +66,6 @@ class Collation implements IRepairStep {
 		$tables = $this->getAllNonUTF8BinTables($this->connection);
 		foreach ($tables as $table) {
 			$output->info("Change collation for $table ...");
-			if ($characterSet === 'utf8mb4') {
-				// need to set row compression first
-				$query = $this->connection->prepare('ALTER TABLE `' . $table . '` ROW_FORMAT=COMPRESSED;');
-				$query->execute();
-			}
 			$query = $this->connection->prepare('ALTER TABLE `' . $table . '` CONVERT TO CHARACTER SET ' . $characterSet . ' COLLATE ' . $characterSet . '_bin;');
 			$query->execute();
 		}
--- a/changelog/unreleased/39286
+++ b/changelog/unreleased/39286
@@ -1,0 +1,10 @@
+Enhancement: Add support for MariaDB 10.6
+
+With this change support for MariaDB 10.6 has been added.
+If it is a fresh ownCloud installation, MariaDB 10.6 can be used right away.
+For upgrading from an older version the command 'occ db:restore-default-row-format'
+has been introduced to remove the deprecated row_format 'compressed' from the
+ownCloud database tables and set it to the default value.
+
+https://github.com/owncloud/core/pull/39286
+https://github.com/owncloud/core/issues/39283
--- a/core/Command/Db/RestoreDefaultRowFormat.php
+++ b/core/Command/Db/RestoreDefaultRowFormat.php
@@ -1,0 +1,95 @@
+<?php
+/**
+ * @author Jan Ackermann <jackermann@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\Core\Command\Db;
+
+use Doctrine\DBAL\Platforms\MySqlPlatform;
+use OCP\IConfig;
+use OCP\IDBConnection;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class RestoreDefaultRowFormat extends Command {
+	/** @var IConfig */
+	private $config;
+
+	/** @var IDBConnection */
+	private $connection;
+
+	/**
+	 * @param IConfig $config
+	 * @param IDBConnection $connection
+	 */
+	public function __construct(IConfig $config, IDBConnection $connection) {
+		$this->config = $config;
+		$this->connection = $connection;
+		parent::__construct();
+	}
+
+	protected function configure() {
+		$this
+			->setName('db:restore-default-row-format')
+			->setDescription('Restore default row format of MySQL/MariaDB tables.');
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output) {
+		if (!$this->connection->getDatabasePlatform() instanceof MySqlPlatform) {
+			$output->writeln("<error>This command is only valid for MySQL/MariaDB databases.</error>");
+			return 1;
+		}
+
+		$dbPrefix = $this->config->getSystemValue("dbtableprefix");
+		$dbName = $this->config->getSystemValue("dbname");
+
+		$defaultRowFormatStatement = $this->connection->executeQuery("SELECT @@GLOBAL.innodb_default_row_format as 'default'");
+		$defaultFormat = $defaultRowFormatStatement->fetchColumn();
+
+		/**
+		 * Fetch tables with the row_format not matching default.
+		 */
+		$tableSelectStatement = $this->connection->executeQuery(
+			"SELECT DISTINCT(TABLE_NAME) AS `name`, LOWER(ROW_FORMAT) as `format`" .
+			"	FROM INFORMATION_SCHEMA . TABLES" .
+			"	WHERE TABLE_SCHEMA = ?" .
+			"	AND TABLE_NAME LIKE ?" .
+			"	AND ROW_FORMAT != ?",
+			[$dbName, $dbPrefix.'%', $defaultFormat]
+		);
+		$tables = $tableSelectStatement->fetchAll();
+		$tablesCount = \count($tables);
+
+		$output->writeln("<info>Found $tablesCount tables to convert</info>");
+
+		// Update row_format to DEFAULT
+		foreach ($tables as $table) {
+			$tableName = $table["name"];
+			$tableRowFormat = $table["format"];
+			$output->writeln("<info>Converting table '$tableName' with row format '$tableRowFormat' to default value '$defaultFormat'</info>");
+
+			$alterTableQuery = $this->connection->prepare('ALTER TABLE `' . $tableName . '` ROW_FORMAT=DEFAULT;');
+			$alterTableQuery->execute();
+		}
+
+		$output->writeln("<info>Conversion done</info>");
+		return 0;
+	}
+}
--- a/core/register_command.php
+++ b/core/register_command.php
@@ -98,6 +98,7 @@ if (\OC::$server->getConfig()->getSystemValue('installed', false)) {
 		)
 	);
 	$application->add(new OC\Core\Command\Db\ConvertMysqlToMB4(\OC::$server->getConfig(), \OC::$server->getDatabaseConnection(), \OC::$server->getURLGenerator()));
+	$application->add(new OC\Core\Command\Db\RestoreDefaultRowFormat(\OC::$server->getConfig(), \OC::$server->getDatabaseConnection()));
 	$application->add(new OC\Core\Command\Db\Migrations\StatusCommand(\OC::$server->getDatabaseConnection()));
 	$application->add(new OC\Core\Command\Db\Migrations\MigrateCommand(\OC::$server->getDatabaseConnection()));
 	$application->add(new OC\Core\Command\Db\Migrations\GenerateCommand(\OC::$server->getDatabaseConnection()));
--- a/tests/Core/Command/Db/RestoreDefaultRowFormatTest.php
+++ b/tests/Core/Command/Db/RestoreDefaultRowFormatTest.php
@@ -1,0 +1,104 @@
+<?php
+/**
+ * @author Jan Ackermann <jackermann@owncloud.com>
+ * @author Jannik Stehle <jstehle@owncloud.com>
+ *
+ * @copyright Copyright (c) 2021, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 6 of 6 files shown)


phil-davis:
@IljaN these are the changes that you made to get CI passing with MariaDb 10.6. The CI passes here with MariaDb 10.2, so the changes are backward-compatible on a newly-installed system.

1) can you please write a changelog for this?

For everyone:
2) what should we do about checking upgrades of existing oC10 systems that use MariaDb 10.2/3/4/5 ?
3) what else can we think of that might be a problem?

JanAckermann:
Did we already tried to connect from a former compressed database with the new connection params?

JanAckermann:
Do we know how long it might take to migrate for a user with a big database ?

lib/private/Repair/Collation.php
```diff
@@ -68,7 +68,7 @@ public function run(IOutput $output) {
 			$output->info("Change collation for $table ...");
 			if ($characterSet === 'utf8mb4') {
 				// need to set row compression first
-				$query = $this->connection->prepare('ALTER TABLE `' . $table . '` ROW_FORMAT=COMPRESSED;');
+				$query = $this->connection->prepare('ALTER TABLE `' . $table .'`;');
```
IljaN, line 71: Not sure if this is the right thing to do as this is in a repair command. So we might need to adjust the command so that it works with both

IljaN:
> Do we know how long it might take to migrate for a user with a big database ?
I couldn't find any info on this. So I guess we need to test. If it's only a metadata DDL change for MariaDB then it might not take long (depends). By the looks of it (just guessing) MariaDB used to store Columns (and indices?) in a compressed form and deprecated this now. Two scenarios I can think of:

- We only change the DDL and Maria is smart enough to convert it's internal structure on the fly / on use
- Database needs to be taken offline and converted/migrated where all data on disk is rewritten

lib/private/Repair/Collation.php
mrow4a: I think it is ok to remove it here, but we just need to make sure it is required for ALTER statement below

mrow4a:
In this PR migration step for ALTERING ROW_FORMAT for tables is missing.

IljaN:
Is this is something which should be done during a normal ownCloud (minor) upgrade or does it need planing by the admin?  There is also the option to set the row type globally so we might be able to remove it from table definitions?

https://github.com/nextcloud/server/issues/25436#issuecomment-873545194

mrow4a:
@IljaN ok, maybe we then provide Repair command only for these that fail to upgrade ? For all new instances we just do not use ROW_FORMAT ? Plus we make sure when MariaDB upgrade is attempted we throw proper error with what to do?

JanAckermann:
@IljaN @mrow4a I would rather love to have a migration command (if the migrate process is quite short), 
ROW_FORMAT DYNAMIC also has the advantage of a faster database.

From a user perspective, figuring out the problem, reading the docs, run the repair command is pretty laborious,
what do you think?

core/Migrations/Version20211005081736.php
```diff
@@ -0,0 +1,41 @@
+<?php
+namespace OC\Migrations;
+
+use OCP\IDBConnection;
+use OCP\Migration\ISqlMigration;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
```
phil-davis, line 8: Comment needs updating

core/Migrations/Version20211005081736.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+class Version20211005081736 implements ISqlMigration {
+
+	public function sql(IDBConnection $connection) {
+		$platform  = $connection->getDatabasePlatform();
+		$platformName = $platform->getName();
+
+		// Only execute this migration on mysql platform
+		if($platformName !== 'mysql'){
```
phil-davis, line 17:
```suggestion
		if ($platformName !== 'mysql'){
```

core/Migrations/Version20211005081736.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Auto-generated migration step: Please modify to your needs!
+ */
+class Version20211005081736 implements ISqlMigration {
+
+	public function sql(IDBConnection $connection) {
+		$platform  = $connection->getDatabasePlatform();
+		$platformName = $platform->getName();
+
+		// Only execute this migration on mysql platform
```
phil-davis, line 16:
```suggestion
		// Only execute this migration on mysql platform (MySQL and MariaDb)
```

Maybe note here that this will apply to both MySQLcand MariaDb implementations of the "mysql" platform.

core/Migrations/Version20181220085457.php
```diff
@@ -31,10 +31,8 @@
 class Version20181220085457 implements ISchemaMigration {
 	public function changeSchema(Schema $schema, array $options) {
 		$prefix = $options['tablePrefix'];
-
```
phil-davis, line 34: Do we need to touch this file? Will that cause any problem or side-effect?

.htaccess
```diff
@@ -94,3 +94,7 @@ Options -Indexes
 <IfModule pagespeed_module>
   ModPagespeed Off
 </IfModule>
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
phil-davis, line 97: Accidental commit

jvillafanez:
I'd vote for removing the row format from our code. From my point of view, we shouldn't need to care about it. We could suggest to use a specific format because of performance reasons, but I don't think we should enforce it. The problem we have now is that we enforce to use a specific format that is now deprecated.

Ideally, that should be fixed by a DBA, completely isolated from ownCloud: switch off ownCloud, convert the DB (on your own, with whatever tools the DB provides), switch on ownCloud.

JanAckermann:
> I'd vote for removing the row format from our code


I agree, I will check if we can set the row_format=DEFAULT in the migration which is DYNAMIC in MariaDB

lib/private/DB/ConnectionFactory.php
```diff
@@ -218,7 +218,7 @@ public function createConnectionParams() {
 			$connectionParams['defaultTableOptions'] = [
 				'collate' => 'utf8mb4_bin',
 				'charset' => 'utf8mb4',
-				'row_format' => 'compressed',
+				'row_format' => 'dynamic',
```
JanAckermann, line 221: As @jvillafanez commented, we should remove this part here

AlexAndBear (changes requested):

phil-davis:
We could bump the oC10 version from `10.8.1 prealpha` to `10.9.0 prealpha` in this PR. That should trigger the migration to auto-run when any developer/QA person updates. That will allow someone to test by installing MariaDb 10.5 with oC 10.8.0 checked out, then checkout this branch (or head of master when this is merged) and `php occ:upgrade` and it should "just work".

JanAckermann:
@phil-davis good idea, still wondering if we should migrate or repair (which should be triggered), my favor is migrate

phil-davis:
> @phil-davis good idea, still wondering if we should migrate or repair (which should be triggered), my favor is migrate

I am happy whichever is decided - auto-migrate, or something for an admin to do explicitly.

jvillafanez:
I'm not sure ownCloud 10.x can work normally without having that row format in the DB, at least until the next core upgrade.

Assuming OC 10.x can work without problems (at least until the next core upgrade), I'd remove the migration.

If the admin keeps using mariaDB 10.5, the only changes we have is that we no longer require the compressed row format, so the repair step won't do anything.

If the admin wants to use mariaDB 10.6, he can change the row format at any time; there is no need to wait for the upgrade (assuming OC doesn't give any problem). The only thing he needs to be careful about is to update to 10.9 for the next upgrade. The upgrade itself on our side won't do anything special.
The only problematic scenario is that the admin, having mariaDB 10.6, updates to OC 10.8 or earlier. In that case, I expect the migration to take some time because I expect ownCloud to change the DB row format back to compressed (due to the repair steps). Anyway, I don't think we can do anything about this other than document that mariaDB 10.6 is "officially supported" from 10.9 onward.

For an update from mariaDB 10.5 + OC 10.8 to mariaDB 10.6 + OC 10.9, I think the best path is:
1. Change row format in mariaDB 10.5
2. Update mariaDB to 10.6
3. Update OC to 10.9
ownCloud should be able to run after each step without problem. In theory, you should be able to stay for a while in step 1, and the same for step 2.

JanAckermann:
> If the admin keeps using mariaDB 10.5, the only changes we have is that we no longer require the compressed row format, so the repair step won't do anything.

This is not quite correct, I will give the admin the possibility to change MariaDB to 10.6 easily.
**Why do we want to enforce our users to search for a problem, which we then need to explain and document, so the user needs to run several queries?**
We can solve all this overhead for us and the admin, by a simple migration

phil-davis:
If we have a migration, then it is currently going to run on every installation where `$platformName == 'mysql'`

I guess that will be every installation that uses "real" MySQL and that uses any supported MariaDb 10.2, 10.3 10.4 or 10.5.

Do installations with "real" MySQL need to do this change at all?
(Is a similar deprecation coming for MySQL?)

IljaN:
> Do installations with "real" MySQL need to do this change at all?
(Is a similar deprecation coming for MySQL?)

MySQL is not affected.

JanAckermann:
@phil-davis according to @jvillafanez  and also this comment https://github.com/nextcloud/server/issues/25436#issuecomment-932979889
I agree that row_format should be set by the DBA if needed and not by ownCloud, so I like to 
pull right away

jvillafanez:
> Why do we want to enforce our users to search for a problem, which we then need to explain and document, so the user needs to run several queries and commands?

* In theory, any ownCloud 10.x version is compatible with mariaDB 10.6 ("compatible" doesn't mean "officially supported"). The problem happens during the upgrade, when ownCloud enforces the compressed row format.
* Admins can update the DB separately from ownCloud. They don't need to wait until ownCloud 10.9 is released. It's probably safer to update one component at a time and check that everything keeps working afterwards.
* I expect the row format conversion to take a lot of time depending on the amount of data in the DB. If we use a migration, all that time will be consumed during the ownCloud upgrade. This is something to avoid if possible, and I think this time is possible.

The problem I have with the migration is that we have to accurately detect which mariaDB (**and not mysql**) version we're using, and I don't think we have such information. If the detection is wrong we end up converting the row format and wasting a lot of time when it isn't needed.

In general, I think that updating the DB separately is safer. The DB update also implies the row format conversion.
If the DB is already updated an ready to be used, ownCloud doesn't need to bother, it just needs not to enforce anything. This is why I don't think we need a migration in this case.

JanAckermann:
Discussed with @pmaier1 :
* remove migration
* introduce db:convert-row-format (without param, will set it to default)
* Official support for MariaDB 10.6 only >= ownCloud 10.9
* Good documentation Martin Mattel I will keep you updated

@jvillafanez We should not recommend switching to MariaDB 10.6 if the ownCloud instance is below 10.9 because the DBA would need to tackle every table which might be brought in after an app installation

jvillafanez:
So I guess the plan is:
1. Update to OC 10.9
2. Run `occ db:convert-row-format`
3. Update mariaDB to 10.6

On the one hand, I'm not happy having to deal with the conversion ourselves. On the other hand, the upgrade path seems reasonable, and probably safer from an admin point of view (less things he should worry about).

Note that the command MUST show progress somehow (a "converting row format for table whatever" message might be enough), and the command must also check if the row format is already the right one (either it was converted beforehand, or the command has been run multiple time).

> @jvillafanez We should not recommend switching to MariaDB 10.6 if the ownCloud instance is below 10.9 because the DBA would need to tackle every table which might be brought in after an app installation

I'm not sure because, as far as I know, apps usually don't fiddle with the DB, only just to create the tables they need. New tables should have been created with the default row format, I think.
Anyway, we should always recommend to go for official support, and official support for mariaDB 10.6 is with ownCloud 10.9

JanAckermann:
> New tables should have been created with the default row format, I think.
I suspect, that the connectionParams already eenforce the Compressed format, but we should check that.


>And the command must also check if the row format is already the right one

In the migration we already check if there are tables with row_format compressed, I think we can still rely on it, in this special case

phil-davis:
@JanAckermann this PR branch is quite a way behind master now. When you need to do more changes, or feel like squashing, you can rebase and force-push.

JanAckermann:
@phil-davis Thx, done

JanAckermann:
@jvillafanez 

> I'm not sure because, as far as I know, apps usually don't fiddle with the DB, only just to create the tables they need. New tables should have been created with the default row format, I think.
Anyway, we should always recommend to go for official support, and official support for mariaDB 10.6 is with ownCloud 10.9

I have double-checked that:
* Install ownCloud instance (master)
* Converting all table row_formats to default -> results in *dynamic*
* Activate files_spaces -> those tables are *compressed*

This is the issuer:
https://github.com/owncloud/core/blob/11c85238feabba1a7147bd7c0517e129b49edd57/lib/private/DB/ConnectionFactory.php#L221

image

AlexAndBear (approved):

JammingBen (approved):

core/Command/Db/ConvertRowFormat.php
```diff
@@ -0,0 +1,121 @@
...
+			"	AND TABLE_NAME LIKE ?" .
+			"	AND ROW_FORMAT != ?",
+			[$dbName, $dbPrefix.'%', $formatInputArgument]
+		);
+		$tables = $tableSelectStatement->fetchAll();
+		$tablesCount = \count($tables);
+
+		$output->writeln("<info>Found $tablesCount tables to convert</info>");
+
+		// Update row_format to DYNAMIC
```
phil-davis, line 108:
```suggestion
		// Update row_format to DEFAULT
```

phil-davis: I will commit this little comment fix and then squash...

core/Command/Db/ConvertRowFormat.php
```diff
@@ -0,0 +1,121 @@
...
+
+class ConvertRowFormat extends Command {
+	/** @var IConfig */
+	private $config;
+
+	/** @var IDBConnection */
+	private $connection;
+
+	/** @var array */
+	private $allowedFormats = ['default'];
```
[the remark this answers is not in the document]
JanAckermann: No, this has been decided by @pmaier1, not for now

core/Command/Db/ConvertRowFormat.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	protected function configure() {
+		$this
+			->setName('db:convert-row-format')
+			->setDescription('Convert row format of MySQL/MariaDB tables.')
+			->addArgument(
+				'format',
+				InputArgument::OPTIONAL,
+				'Desired row format',
```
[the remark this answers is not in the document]
JanAckermann: See comment above

jvillafanez:
It seems to be a mismatch between the code and the expectations. The code seems to convert the table row format to the default one, however, the command's help suggests that it's possible to convert to other row formats, but this isn't the case.

core/Command/Db/ConvertRowFormat.php
```diff
@@ -0,0 +1,121 @@
...
+
+		$output->writeln("<info>Found $tablesCount tables to convert</info>");
+
+		// Update row_format to DEFAULT
+		foreach ($tables as $table) {
+			$tableName = $table["name"];
+			$tableRowFormat = $table["format"];
+			$output->writeln("<info>Converting table '$tableName' with row format '$tableRowFormat' to '$formatInputArgument'</info>");
+
+			$alterTableQuery = $this->connection->prepare('ALTER TABLE `' . $tableName . '` ROW_FORMAT=DEFAULT;');
```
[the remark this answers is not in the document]
JanAckermann: >ROW_FORMAT=DEFAULT
this should be the input

core/Command/Db/ConvertRowFormat.php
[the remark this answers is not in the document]
phil-davis: Do we even want to have this argument?
At the moment, we only want the admin to be able to set the table-row-format to default, so we can have no argument - the command can "just do it".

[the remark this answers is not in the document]
jvillafanez: I agree with @phil-davis . This will leave a bad impression with the current code. The admin can check what options MariaDB provides and check them one by one just to see no option is allowed.

I'm ok with having plans for the future, but in this case it seems better to have just a "TODO" comment to include an option to choose the row format; just a comment without any code.

phil-davis (commented):
LGTM - I am the PR "owner" so I can't approve.

jvillafanez (approved):

PR #39292, opened 2021-09-27
Don't show mail in lock owner data if mail adress is empty string

### LINKED ISSUE #37878: Lock owner has extra brackets at end
### Steps to reproduce
1. As Alice, set your full name to "This is my full name"
2. As Alice, do not set an email address
3. Lock a file using the webUI manual locking
4. View the lock details

### Expected behaviour
Lock information has:
`This is my full name has locked this resource via /IM-6553-D13_1pdf`

(i.e. without the empty `()`)

### Actual behaviour
Lock information has:
`This is my full name () has locked this resource via /IM-6553-D13_1pdf`

### Server configuration
Current core master.

AlexAndBear:
## Description
Bugfix: Don't show empty brackets in lock owner information if mail not set

## Related Issue

- Fixes https://github.com/owncloud/core/issues/37878

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Lock/Persistent/LockManager.php
+++ b/lib/private/Lock/Persistent/LockManager.php
@@ -78,7 +78,7 @@ class LockManager {
 			$user = $this->userSession->getUser();
 			if ($user !== null) {
 				$owner = $user->getDisplayName();
-				if ($user->getEMailAddress() !== null) {
+				if (!empty($user->getEMailAddress())) {
 					$owner .= " ({$user->getEMailAddress()})";
 				}
 			}
--- a/lib/private/Lock/Persistent/LockManager.php
+++ b/lib/private/Lock/Persistent/LockManager.php
@@ -78,8 +78,9 @@ class LockManager {
 			$user = $this->userSession->getUser();
 			if ($user !== null) {
 				$owner = $user->getDisplayName();
-				if (!empty($user->getEMailAddress())) {
-					$owner .= " ({$user->getEMailAddress()})";
+				$userEmail = $user->getEMailAddress();
+				if ($userEmail !== null && $userEmail !== '') {
+					$owner .= " ({$userEmail})";
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


JanAckermann:
On the other hand we can just remove the extra () field as @DeepDiver1975 suggested in the recent issue

lib/private/Lock/Persistent/LockManager.php
```diff
@@ -78,7 +78,7 @@ public function lock($storageId, $internalPath, $fileId, array $lockInfo) {
 			$user = $this->userSession->getUser();
 			if ($user !== null) {
 				$owner = $user->getDisplayName();
-				if ($user->getEMailAddress() !== null) {
+				if (!empty($user->getEMailAddress())) {
```
phil-davis, line 81: I don't like these. In this case the email address should never be the string "0" or "0.0" so I guess there is no real email address that will evaluate to `empty()` `true`.

JanAckermann: Well @DeepDiver1975 recommended getting rid of the extra data, as this data won't be updated if the user changes the email. I am also not quite sure who is able to see the email and it might not be GDPR compliant.

JanAckermann: > so I guess there is no real email address that will evaluate to empty() true
Empty string evaluates to true, so I don't understand the objection

>In this case the email address should never be the string "0" or "0.0" 
And that is correct, isn't it?

phil-davis: In this case `empty()` works, because the value being compared can never be `0`, `0.0`, "0" or "0.0"!

JanAckermann: updated the condition

JammingBen (approved):

phil-davis (approved):
works.

Note: to get this problem in the first place, an email address has to have been set for/by a new user. Then the user deleted the email address (sets it to the empty string, effectively). If a user never ever had an email address then this problem did not happen - the user's email address was effectively `null`.

DeepDiver1975 (changes requested):
Test coverage?

DeepDiver1975 (approved):

PR #39296, opened 2021-09-27
Optimized txt preview

jvillafanez:
## Description
Optimize the way of choosing a font for the text preview. The changes with the previous implementation are:
* Go back to check the first 3000 bytes. It should give a better overview of the text
* Analyze the whole string to gather data, mostly the number of chars that belong to specific scripts. This is done in only one pass regardless of the number of scripts we want to check.
* Each char is analyzed separately in order to group them with the right script.
  * A binary search is performed in order to find the right script for the char because we have to check multiple ranges. Indexing seems impractical due to the sheer amount of chars we have to deal, specially for chinese chars.
  * Only specific scripts are checked, mostly for those we have or need specific fonts.
* Adding an additional font shouldn't cost too much performance due to the binary search.

As far as I know, both the previous and this implementation only works with files encoded in utf8. Note that the file encoding might change during the upload (checked by uploading an utf-16 encoded file via web browser to ownCloud, file was utf-8 encoded in the ownCloud FS).
I don't think it will be possible to correctly detect the encoding of the file without limiting the number of options we have. So far, in order both to generate the preview image and to choose the font, we assume the text is utf8-encoded even though we don't know for sure. 

## Related Issue
Follow up for https://github.com/owncloud/core/pull/39271

## Motivation and Context
Improve the implementation

## How Has This Been Tested?
Checked by uploading multiple text files containing chinese, japanese and korean text respectively. The preview shows the character correctly

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


phil-davis:
> Go back to check the first 3000 bytes

What part of the text actually appears in the preview? Just the first "n characters"? Or does it depend on the size of preview requested? (Maybe a big preview has more of the text in it) Is there text from multiple lines? Or just the first line?

If previews only contain text from near the beginning of the file, then it does not matter what different scripts appear later in the file - they will not effect the preview. We only need to look in the text file as far as the characters that are used in the preview.

jvillafanez:
> What part of the text actually appears in the preview? Just the first "n characters"? Or does it depend on the size of preview requested? (Maybe a big preview has more of the text in it) Is there text from multiple lines? Or just the first line?

There is text from multiple lines.
As far as I see in the code, those 3000 bytes are split into different lines, and each line is drawn into the image even if the line is bigger than the actual width. There is no fixed limit of lines to draw, but it's limited to the vertical size of the image.
Just checking the bytes seems a bit random, and as it is at the moment, I don't think we can't decide beforehand how many bytes we should read.

We might want to change the algorithm in order to make it more predictable, but I think it's better to tackle it in a different PR.

JanAckermann:
This looks quite cool to me.
But we should really test how the performance is, I would suspect it might be slower than the previous solution.

Note: Having different langs like Arabian and Chinese will still result in a corrupted preview.

jvillafanez:
test code:
```
<?php
require_once 'lib/base.php';

$newPrev = new \OC\Preview\TXT();

$user = 'admin';
\OC_User::setUserId($user);
$userObject = \OC::$server->getUserManager()->get($user);
if ($userObject !== null) {
  $userObject->updateLastLoginTimestamp();
}
\OC_Util::setupFS($user);
if (\OC_User::userExists($user)) {
  $uFolder = \OC::$server->getUserFolder($user);
}

$test = $uFolder->get('test.mul.txt');

$t1 = microtime(true);
$image = $newPrev->getThumbnail($test, 64, 64, false);
$t2 = microtime(true);

echo $t2 - $t1;
echo "\n";
```
------

test file
[test.mul.txt](https://github.com/owncloud/core/files/7244286/test.mul.txt)

Preg-based implementation (showing arab chars):
```
0.059641122817993
0.059026956558228
0.055532932281494
0.060762166976929
```

New implementation (showing nepali chars)
```
0.061712980270386
0.056956052780151
0.057214975357056
0.058809995651245
```

------

second test file
[test.chi2.txt](https://github.com/owncloud/core/files/7244488/test.chi2.txt)

Preg-based implementation
```
0.037962913513184
0.042697906494141
0.038923025131226
0.038248062133789
0.046655178070068
```

New implementation
```
0.040887117385864
0.038702011108398
0.044713973999023
0.040306091308594
0.039341926574707
```

----

It's a bit slower with the "chinese-only" file, but taking into account that chinese is prioritize in the preg-based implementation and that the new implementation checks 3000 bytes instead of ~40 bytes.

I'll try to tweak a bit some things, but it seems the bottleneck is the time generating the image taking into account that a 6k files takes less time to generate the thumbnail than a 2.5k file in both scenarios.

jvillafanez:
New implementation, reading 1024 bytes instead of 3000, with the chinese test file:
```
0.040230989456177
0.041236877441406
0.037501096725464
0.038942813873291
0.035165071487427
```
The performance of the new implementation is similar to the previous one in its best case.

Just for the record, I was checking if we could change the `preg_split` call in the text preview, but I'm rejecting the idea for now. We have enough information to know where the new lines are (the detail information can show the bytes and the unicode code point of each char of the stream), but parsing the information and splitting the lines outside of the utf8analyzer will take time. We could also try to split the lines inside the utf8analyzer, but it's difficult to fit the behavior in only one function (we don't want to have to go through the same stream again).
My expectation is that we might get similar performance of the `preg_split` function, but the complexity we need to add doesn't justify the change.

For performance, it seems a better choice to focus on the actual image generation.

No plans to change more code in this PR, so moving to add unit tests

jvillafanez:
More speed improvements with the latest commit (not planned but too good to let it go):
* with the "test.mul.txt" file
  * previous best: 0.055532932281494
  * current worst: 0.038245916366577
* with the "test.chi2.txt":
  * previous best: 0.035165071487427  (with 1024 bytes)
  * current worst: 0.02238392829895  (with 2024 bytes)

JanAckermann:
Note: CI is failing because of https://github.com/owncloud/core/pull/39290 (suspect)

phil-davis:
CI fixed by PR #39317 
This needs a rebase.

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavPreviews-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32849/94/1

jvillafanez:
It seems the tests are checking if the file contents are exactly the same as a predefined one... I think it's too strict and there are some problems with the approach:
* The error message doesn't say anything useful. Yes, the content is different, but I don't know if it's properly rendered, or the image contains some additional metadata (which might be invisible in the preview), or we're showing a different character.
* Any change will modify the file content and will fail the test. This include, a partial character that could now appear at the edge of the image, some changes in the lines positioning (less space between the lines, for example), and other minor changes that doesn't really affect to the preview but will change the contents of the file.
* The difference could be just a pixel. It will be difficult to spot even having both images side by side.

phil-davis:
We thought about trying to "recognize" if the text "looks like" what we expect. That kind of logic would be very difficult to implement. So we went with keeping a known-correct image. If the rendering changes a bit, then a new version of the known-good PNG file can be committed. The "string"  from the API preview response just needs to be written to a file, name the file with `.png` and then view it. It should be "easy" enough for the developer to visually open the old and new PNG files and confirm the the characters look similar - you don't need to actually be able to read the script.

Other ideas are welcome.

jvillafanez:
We could set some metadata in the image, but it's probably a bad idea since it will be implemented as a post-processing action (basically, read and rewrite the image). We'd need to hack our way through.
It might have some uses, but if it's just for checking tests.... it might be too much work for very little reward.
It's technically possible for png files, but I haven't check for other formats.

AlexAndBear (approved):
Overall LGTM.
I would like if you to add a comment to class TXT about the limitations of having different language symbols like Arabian and Japanese, will still result in a faulty image

phil-davis (approved):

jvillafanez:
Squashed. It shouldn't have lost too much context

PR #39210, opened 2021-09-10
Add permission check when restoring from trashbin

JammingBen:
## Description
This fixes an issue where a user was able to restore a file or folder into a read-only shared folder.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/35900

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39305, opened 2021-09-29
[tests-only] skip more sharing scenarios on old oC10

phil-davis:
## Description
These were missed in PR #39303 yesterday.

In the 2nd commit I have added a step definition that lets us say:
```
    When the administrator tries to add user "db-user" to group "grp1" using the webUI
```

That step looks for the element(s) on the UI to add the user to the group, but does not throw an exception when any of them are not found. This is the sort of thing that we do these days to indicate that a `When` step is "expected" to not really work. The `Then` steps after it should/will check that the thing did not happen:
```
    Then user "Alice" should exist
    But user "Alice" should not belong to group "grp1"
```

This will allow us to refactor the scenarios in user_ldap tests that are failing:
```
webUIProvisioning/groups.feature:70
webUIProvisioning/groups.feature:85
```

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiTrashbin/trashbinSharingToShares.feature
+++ b/tests/acceptance/features/apiTrashbin/trashbinSharingToShares.feature
@@ -1,4 +1,4 @@
-@api @files_trashbin-app-required @files_sharing-app-required
+@api @files_trashbin-app-required @files_sharing-app-required @skipOnOcV10.6 @skipOnOcV10.7 @skipOnOcV10.8.0
 Feature: using trashbin together with sharing
 
   Background:
@@ -63,7 +63,7 @@ Feature: using trashbin together with sharing
       | old      |
       | new      |
 
-  @skipOnOcV10.6 @skipOnOcV10.7 @skipOnOcV10.8.0
+
   Scenario Outline: restoring a file to a read-only folder
     Given using <dav-path> DAV path
     And user "Brian" has been created with default attributes and without skeleton files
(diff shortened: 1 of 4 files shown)


tests/acceptance/features/apiTrashbin/trashbinSharingToShares.feature
```diff
@@ -63,7 +63,7 @@ Feature: using trashbin together with sharing
       | old      |
       | new      |
 
-  @skipOnOcV10.6 @skipOnOcV10.7 @skipOnOcV10.8.0
```
JammingBen, line 66: Are you sure that these should be removed? I believe this scenario only works because of one of the recent PRs, therefore not in OC <= 10.8.0.

JammingBen: I mean this PR: https://github.com/owncloud/core/pull/39210 and the followup with the skip-tags: https://github.com/owncloud/core/pull/39215

phil-davis: I put the skip tags on the whole feature file at line 1, so they apply to every scenario.

phil-davis:
@JammingBen or anyone. Please review again. After this is merged, I can make a PR  to `user_ldap` to use this and get CI green.

individual-it (approved):

PR #39306, opened 2021-09-29
Sanitize data that  gets sent to server on user create via webUI

### LINKED ISSUE #32619: Switch between email and password in user creation form issue
### Steps

1. Setup OC
1. Go to the users page
1. Type "user1"
1. In the email field type "test", believing that this was the password field. Leave the value in the field.
1. Click on the cog icon at the bottom left
1. Click "Set password for user", this toggles the email field to be a password field
1. Type in "test" as password
1. Click "create"

### Expected result
User created.

### Actual result
Error message about invalid email address.
This is because the email field is just hidden and still has its value.

Solution is to either:
1. clear the field automatically when toggling
or
2. ignore the hidden field when gathering form values

### Version
ownCloud 10.0.1 RC1

AlexAndBear:
## Description
Bugfix: Sanitize data send to the server while creating users via webUI

Before this change toggle between 'Set password for new users' option,
may preserve and send unwanted password or email information.
This has been fixed, the webUI will not send email data to the server
if the option 'Set password for new users' is active,
vice versa password won't be sent if the option is disabled.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/32619

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/settings/js/users/users.js
+++ b/settings/js/users/users.js
@@ -959,13 +959,15 @@ $(document).ready(function () {
 		var username = $('#newusername').val();
 		var password = $('#newuserpassword').val();
 		var email = $('#newemail').val();
+		var setPassword = $('#CheckBoxPasswordOnUserCreate').is(':checked');
+
 		if ($.trim(username) === '') {
 			OC.Notification.showTemporary(t('settings', 'Error creating user: {message}', {
 				message: t('settings', 'A valid username must be provided')
 			}));
 			return false;
 		}
-		if ($('#CheckBoxPasswordOnUserCreate').is(':checked')) {
+		if (setPassword === true) {
 			if ($.trim(password) === '') {
 				OC.Notification.showTemporary(t('settings', 'Error creating user: {message}', {
 					message: t('settings', 'A valid password must be provided')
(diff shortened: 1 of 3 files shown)


phil-davis (approved):
Test by:
- uncheck "Set password for new users"
- enter username "t123" and email "test" (thinking it is the password field)
- click Create
- see the message about "invalid email address"
- realize your error, now check "Set password for new users"
- enter "test" in the password field
- click Create

On master I get the message again about "invalid email address"

On this branch the user is created and has the initial password "test".

Works

phil-davis:
IMO this is a bit of an edge case. No need to add an automated test scenario for this sequence. (If we have a scenario for every different possible order of clicking settings on/off and entering data in different orders, then the test matrix will explode and we will be the cause of global warming)

settings/js/users/users.js
```diff
@@ -981,42 +981,54 @@ $(document).ready(function () {
 			}
 		}
 
+
 		var promise = $.Deferred().resolve().promise();
 
 		promise.then(function() {
```
jvillafanez, line 987: Either I'm missing something or I don't get this promise... The promise is already resolved, so the code will be executed immediately, and the promise isn't used elsewhere, so why don't execute the inner function directly?

settings/js/users/users.js
```diff
@@ -981,42 +981,54 @@ $(document).ready(function () {
...
-			var groups = $('#newuser .groups').data('groups') || [];
-			$.post(
-				OC.generateUrl('/settings/users/users'),
-				{
+			OC.AppConfig.getValue('core', 'umgmt_set_password', 'false', function (setPassword) {
```
jvillafanez, line 988: I'm not sure if we want to cache the value to prevent calling the server if we don't need it. It might become cumbersome, so probably it's better to keep it this way.

jvillafanez (commented):
Just a couple of minor things, but overall looks good

settings/js/users/users.js
JanAckermann: I suggest keeping it that way, we could also use the jquery object to check if the box is set, but this is just one request if you create a user, so I don't see any issues with that.

JanAckermann:
Note, added a slightly UI improvement, + add group has now cursor:pointer

JanAckermann:
Before the recent commit we had 0 coverage Sonar Cloud passed, now he have 0 again and it fails?

jvillafanez (approved):

JanAckermann:
@micbar can you merge please ?

phil-davis (approved):

phil-davis:
@micbar ping - this is ready for merge. It just needs override of the coverage.

PR #39310, opened 2021-09-30
Trim starting and trailing spaces while creating/uploading/renaming files and folders in the webUI

### LINKED ISSUE #35017: spaces at the end and the beginning of a filename are trimmed away for display
### Steps to reproduce
1. have a files called `   my file.txt    ` (spaces at the beginning and the end of the name)

### Expected behaviour
spaces are displayed in the UI

### Actual behaviour
spaces are trimmed 

image

`data-file` in the `tr` has the correct name but the content of `<span class="nametext">` has the trimmed name

so we can end up with same looking file names in the folder
image

CC @phil-davis we have tests that do renames to names with trailing and leading spaces, they all pass nicely because we only check the `data-file` attribute of `tr` not the human readable name

AlexAndBear:
## Description
Enhancement: Trim spaces while creating, uploading or renaming entities in webUI

With this change starting or trailing spaces will be removed while:
* Creating files and folders in the webUI
* Uploading files using the 'Upload' button in the webUI
* Uploading files and folders using drag & drop in the webUI
* Renaming files and folders in the webUI

This also solves problems when the sync client can't sync files and folders
with trailing spaces based on the underlying file system.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/4772
- Fixes https://github.com/owncloud/core/issues/35017

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -111,7 +111,38 @@ OC.FileUpload.prototype = {
 		if (this._newName) {
 			return this._newName;
 		}
-		return this.getFile().name;
+
+		var fileName = this.getFile().name;
+		return this.sanitizeFileName(fileName);
+	},
+
+	/**
+	 * Return the sanitized file name.
+	 *
+	 * @return {String} file name
+	 */
+	sanitizeFileName: function(fileName) {
+		return fileName.trim();
+	},
+
+	/**
+	 * Return the sanitized path.
+	 *
+	 * @return {String} path
+	 */
+	sanitizePath: function(path) {
+		if(!path){
+			return path;
+		}
+
+		var pathSegments = path.split('/');
+		var sanitizedPathSegments = [];
+
+		for (var i = 0; i < pathSegments.length; i++) {
+			sanitizedPathSegments.push(pathSegments[i].trim());
+		}
+
+		return sanitizedPathSegments.join('/');
 	},
 
 	getLastModified: function() {
@@ -140,7 +171,9 @@ OC.FileUpload.prototype = {
 	 * @return {String} full path
 	 */
 	getFullPath: function() {
-		return OC.joinPaths(this._targetFolder, this.getFile().relativePath || '');
+		var relativePath = this.getFile().relativePath;
+		var sanitizedRelativePath =  this.sanitizePath(relativePath);
+		return OC.joinPaths(this._targetFolder, sanitizedRelativePath || '');
 	},
 
 	/**
@@ -186,7 +219,7 @@ OC.FileUpload.prototype = {
 	 * Multiple calls will increment the appended number.
 	 */
 	autoRename: function() {
-		var name = this.getFile().name;
+		var name = this.sanitizeFileName(this.getFile().name);
 		if (!this._renameAttempt) {
 			this._renameAttempt = 1;
 		}
@@ -218,7 +251,7 @@ OC.FileUpload.prototype = {
 		 * Files being handled separately
 		 */
 		if (file.isDirectory){
-			return this.uploader.ensureFolderExists(OC.joinPaths(this._targetFolder, file.fullPath));
+			return this.uploader.ensureFolderExists(OC.joinPaths(this._targetFolder, this.sanitizePath(file.fullPath)));
 		}
 
 		// it was a folder upload, so make sure the parent directory exists already
@@ -824,6 +857,7 @@ OC.Uploader.prototype = _.extend({
 	 * @param {function} callbacks.onCancel
 	 */
 	checkExistingFiles: function (selection, callbacks) {
+		var self = this;
 		var fileList = this.fileList;
 		var conflicts = [];
 		// only keep non-conflicting uploads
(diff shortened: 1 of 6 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>webUIFfSmoke-3-3-firefox-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32777/161/1

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIProxySmoke-3-3-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32780/164/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiOnSqlite-sqlite-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32785/173/1

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIRenameFolders-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32786/134/1

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIRenameFiles-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32788/133/1

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIRenameFiles-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32789/133/1

apps/files/js/file-upload.js
```diff
@@ -881,6 +915,15 @@ OC.Uploader.prototype = _.extend({
 		callbacks.onNoConflicts(selection);
 	},
 
+	/**
+	 * Return the sanitized file name.
+	 *
+	 * @return {String} file name
+	 */
+	sanitizeFileName: function(fileName) {
```
jvillafanez, line 923: I'm ok with this function since we could add additional checks at some point, but it should also be used in all the places where we're trimming the filename ("filelist.js", "newfilemenu.js"...). This might not be the best place to put this function.

Note that this is duplicated (~ line 124)

phil-davis (commented):
This works for me. I tried edge-cases like:
- start with folders and sub-folders that have leading/trailing space in the name
- try to upload files or folders into the root, or into any folder or sub-folder - they get trimmed of leading/trailing space
- rename existing files/folders that have leading/trailing space, the space is trimmed off
- try to rename an existing file folder to add leading/trailing space, the space is trimmed off
- drag-and-drop to upload stuff - leading/trailing space is trimmed, including from any files/folders in the whole tree that is uploaded

apps/files/js/file-upload.js
JanAckermann: Duplicated because these are different objects aka classes in one File:-1: 
I don't want to put this function in utils.js because it loses context and does not have any purpose outside of the file-upload.js

So for now, I think it is best to leave it where it is and move it, if it will solve more purposes.

JanAckermann:
@phil-davis  nice, thx! 
Do you have any edge cases in mind if it comes to replacing items aka conflict ?

apps/files/js/file-upload.js
JanAckermann: It is duplicated because the file holds more than one class/objects :-1: 

I don't feel like moving it to utils.js because it will lose context, therefore we also need to shift sanitizePath to utils.js.
The thing is: you can still upload files via API with starting/trailing slashes. Using sanitizePath or sanitizeFileName systemWide might be able to produce bugs.
So I vote to keep it there. 

 Also: there are plans to trim spaces in all inputs anyway, so don't want to use the sanitizeFileName function for trimming inputs ("filelist.js", "newfilemenu.js"...).

phil-davis:
> Do you have any edge cases in mind if it comes to replacing items aka conflict?

It will be a bit odd if there is already a file or folder named like `something ` (with a space at the end) and then you upload a file/folder called `something` (without the space). You will have both resources there. That will be an edge-case that does not automagically get rid of a space. But that is already existing behavior.

Other things that I have tried bring up the conflict dialog and ask what to do, and happily create names like `something (2)` etc.

Because any stripping of spaces gets applied from the web UI to the server, those will been seen as new changes and synced to other clients, which seems good and should work.

JanAckermann:
@phil-davis THX

>That will be an edge-case that does not automagically get rid of a space. But that is already existing behavior.
Quite correct, but this should not be handled via UI IHMO, if we need it, we should write a migration

phil-davis:
> Quite correct, but this should not be handled via UI IHMO, if we need it, we should write a migration

IMO we don't need to bother having an automatic for any existing files with leading/trailing spaces.

It might be useful to have a command somewhere that scans user storage for files with leading/trailing spaces and trims them. An admin can choose to do that if they like, which will cleanup this all-at-once.

apps/files/js/file-upload.js
jvillafanez: Ok. Fair enough

jvillafanez (approved):

PR #39112, opened 2021-08-16
[Tests-Only] Accept all shares matching the path when accepting shares

aduffeck:
## Description

This pr changes the `When /^user "([^"]*)" (declines|accepts) share "([^"]*)" offered by user "([^"]*)" using the sharing API` testing step to accept all shares matching the path instead of just the first one.
This is required for test cases like [this one](https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiShareManagementToShares/mergeShare.feature#L24) where both a group and a user share for the same resource need to be considered.

## Motivation and Context

It allows for properly test the existing test cases.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [X] Tests only (no source changes)

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39338, opened 2021-10-08
[tests-only] Accept both the user and the group share offered by Alice

aduffeck:
## Description
This PR changes a few tests to explicitly accept all shares for the same resource. That is required for the tests to pass in the sharesstorageprovider branch.

## Types of changes
- [X] Tests only (no source changes)

### DIFF
--- a/tests/acceptance/features/apiShareManagementToShares/mergeShare.feature
+++ b/tests/acceptance/features/apiShareManagementToShares/mergeShare.feature
@@ -25,7 +25,8 @@ Feature: sharing
     Given user "Alice" has created folder "/merge-test-outside-perms"
     When user "Alice" shares folder "/merge-test-outside-perms" with group "grp1" with permissions "read" using the sharing API
     And user "Alice" shares folder "/merge-test-outside-perms" with user "Brian" with permissions "all" using the sharing API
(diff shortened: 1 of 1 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareManagementToShares-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/32956/64/1

aduffeck:
@phil-davis the server accepts/rejects the shares by id, it doesn't automagically accept all shares to the same resource if one of them is accepted, for example. But once multiple shares for the same resources have been accepted the permissions will be merged.
After the first accept request was sent the according share will no longer be part of the pending shares list so the next one in line can be picked up. So in the end we need two requests to be sent and in https://github.com/owncloud/core/pull/39112 I tried to make the `accepts share` step do them both at once.
@individual-it rightfully claimed it was not the best way of doing it so I'm now using the `has accepted the first pending` and `has accepted the next pending` steps which make it clearer what is going on in my opinion.

phil-davis:
@aduffeck the acceptance test scenarios and code used in ocis drone CI were all copied to the ocis repo some time ago.
So this PR to core is no longer relevant.
Close?

DeepDiver1975:

PR #39343, opened 2021-10-11
Fix type declaration in docstring for the setParameter method

### LINKED ISSUE #39341: IQueryBuilder::setParameter @param annotation is incorrect (?)
https://github.com/owncloud/core/blob/12e821c9fc611bd403a7ea249fe4ed0eae70c1ff/lib/public/DB/QueryBuilder/IQueryBuilder.php#L147-L165

AFAICT the type annotation for the third param `$type` is incorrect here.
The annotation mandates to use `string` or `null`. The documentation says one is supposed to use `One of the IQueryBuilder::PARAM_* constants.`  Those are all `int`s.
Now with code like `->setParameter('ids', $accountIds, IQueryBuilder::PARAM_STR_ARRAY);` phan is unhappy because `IQueryBuilder::PARAM_STR_ARRAY` is not a string.

```
SomeFile.php:94 PhanTypeMismatchArgument Argument 3 ($type) is IQueryBuilder::PARAM_STR_ARRAY of type 102 but \OCP\DB\QueryBuilder\IQueryBuilder::setParameter() takes null|string defined at ../../lib/public/DB/QueryBuilder/IQueryBuilder.php:163
```

Apparently the documentation/signature is copied from doctrine upstream - so not sure what to do here.

After all: Why is this only causing issues in my app while the same code seems to work everywhere else without phan going crazy over it?

JammingBen:
## Description
...to match with the underlying doctrine docstring.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/39341

## Types of changes

- [x] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/DB/QueryBuilder/QueryBuilder.php
+++ b/lib/private/DB/QueryBuilder/QueryBuilder.php
@@ -25,6 +25,7 @@ namespace OC\DB\QueryBuilder;
 
 use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\DBAL\Types\Type;
 use OC\DB\OracleConnection;
 use OC\DB\QueryBuilder\ExpressionBuilder\ExpressionBuilder;
 use OC\DB\QueryBuilder\ExpressionBuilder\MySqlExpressionBuilder;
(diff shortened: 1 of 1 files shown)


JammingBen:
No need for a changelog item I think.

phil-davis (approved):

dschmidt:
Nice - thanks!

PR #39393, opened 2021-10-22
[full-ci][tests-only] Add tests for db-convert occ command

### LINKED ISSUE #39190: add cli tests for `occ` db commands
From https://github.com/owncloud/core/issues/33051

add cli tests for `db` commands (might have to setup a ci pipeline to start with multiple dbs and convert them back and forth)

Ref:
[Databse conversion command docs](https://doc.owncloud.com/server/10.8/admin_manual/configuration/server/occ_command.html#database-conversion)

saw-jan:
## Related Issue
- Fixes #39190

## How Has This Been Tested?
- test environment:

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added


saw-jan:
Currently blocked due to unknown error in afterScenario hooks

kiranparajuli589:
I've initialized a state variable in featureContext to check if running for DB conversion. The variable state is set to true after successful DB conversion. but for the before scenario we don't know yet if the scenario is for DB conversion.

although multiple `after` hooks can be implemented even for `specific` tags, but the same is not true for a `before`. It can only be one and doesn't filter with tags.

Should we use some test environment variable, so that we can test some edge cases in this DB conversion feature?

cc @saw-jan @individual-it @phil-davis

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -5467,6 +5467,17 @@ public function restoreAppEnabledDisabledState():void {
 		}
 	}
 
+	/**
+	 * @AfterScenario @dbConversion
+	 *
+	 * @return void
+	 */
+	public function unsetDbConversionState():void {
```
phil-davis, line 5475: why do we need this at all?
the context is thrown away at the end of every scenario anyway.

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -5426,7 +5426,7 @@ public function cleanupDatabaseGroups():void {
 	 * @throws Exception
 	 */
 	public function rememberAppEnabledDisabledState():void {
-		if (!OcisHelper::isTestingOnOcisOrReva()) {
+		if (!OcisHelper::isTestingOnOcisOrReva() && !$this->isRunningForDbConversion()) {
```
phil-davis, line 5429: why can't we let this run before a Db conversion scenario?

tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -5447,7 +5447,7 @@ public function rememberAppEnabledDisabledState():void {
 	 * @throws Exception
 	 */
 	public function restoreAppEnabledDisabledState():void {
-		if (!OcisHelper::isTestingOnOcisOrReva()) {
+		if (!OcisHelper::isTestingOnOcisOrReva() && !$this->isRunningForDbConversion()) {
```
phil-davis, line 5450: What happens when this code is run after a Db conversion scenario?
Is something not working in the test environment?

tests/acceptance/features/cliDbConversion/dbConversion.feature
```diff
@@ -0,0 +1,14 @@
...
+  So that I can change from one database type to another
+
+
+   Scenario Outline: convert different database types
+     When the administrator changes the database type to "<dbtype>"
+     Then the command should have been successful
+     And the system config should have dbtype set as "<dbtype>"
+     Examples:
+       | dbtype   |
+       | mysql    |
```
phil-davis, line 14: This will be the easiest conversion - IMO the pipeline starts with MariaDb installed, and MySQL is very similar.

It is going to be very difficult to "revert" the converted system after running one scenario. Do we just choose one conversion to test?

Or a chain of conversions? (The first example does MariaDb to MySQL, the next is MySQL to PostgreSQL - whatever is supported)

tests/acceptance/features/cliDbConversion/dbConversion.feature
```diff
@@ -0,0 +1,14 @@
...
+Feature: Change from one database to another
+  As an admin
+  I want to be able to convert database from one type to another
+  So that I can change from one database type to another
+
+
+   Scenario Outline: convert different database types
+     When the administrator changes the database type to "<dbtype>"
+     Then the command should have been successful
+     And the system config should have dbtype set as "<dbtype>"
```
phil-davis, line 11: Is the test environment in a working state at this point?

If so, then we can put some `Given` steps ("Alice" "Brian"... has been created, some group has been created, some file/folder and share created.) and `Then` steps can check that the users/groups still exist, Brian can download the shared file...

tests/acceptance/features/cliDbConversion/dbConversion.feature
kiranparajuli589: According to documentation [see here](https://doc.owncloud.com/server/10.8/admin_manual/configuration/database/db_conversion.html#introduction)
> If you have an existing ownCloud installation which uses SQLite, and you want to convert to a better performing database, such as MySQL, MariaDB or PostgreSQL, you can use the ownCloud command line tool: occ.

conversion is possible from
```
             |---- psql
sqlite    ---|---- mysql
             |---- mariadb
```

So, IMO chaining is not possible.

We need to have separate sites for each conversion if we want tests for each type of `db conversion`s.

So do we write tests for other conversions?

tests/acceptance/features/bootstrap/Provisioning.php
kiranparajuli589: removed

tests/acceptance/features/bootstrap/Provisioning.php
kiranparajuli589: these was behaving conflicting when there were edge cases scenarios in the feature
removed now

tests/acceptance/features/bootstrap/Provisioning.php
kiranparajuli589: https://drone.owncloud.com/owncloud/core/33952/121/14

I think the `app:list` command fails here. I tried to run `app:list` after conversion but works fine locally.

tests/acceptance/features/cliDbConversion/dbConversion.feature
phil-davis: No, it is enough to test sqlite to mysql.
(Others would have to be run in separate CI pipelines, and, IMO, we really just want to demonstrate that the command works and does convert at least the basic things like users, files, shares.

phil-davis:
https://drone.owncloud.com/owncloud/core/33955/4/14
`Notice: Array to string conversion in /drone/src/tests/TestHelpers/SetupHelper.php line 920`

I have a fix for that in https://github.com/owncloud/core/pull/39514/commits/bf96d4a10eafd27e79ea79d60b244b32bdc1a980

You can cherry-pick that commit into this branch. Then you should get the real Exception message logged.

kiranparajuli589:
https://drone.owncloud.com/owncloud/core/33957/4/14

```
      ╳  Response is not valid XML after executing 'occ app:list --output json --no-ansi'. Is the testing app installed and enabled?
```

after a successful dB conversion, the `testing` app is disabled, maybe there is a reset.
but actually this was not true for me while trying locally.

cc @phil-davis

phil-davis:
That is in an AfterScenario. For this special test, we only want to run a single scenario. So we can skip any AfterScenario actions. Basically, we throw-away the system-under-test after all the Scenario `Then` steps have finished.
(If someone runs this db-convert scenario locally, it is going to expect that their existing install starts with sqlite and will convert them to mysql, assuming they have a mysql database ready etc, so it is not the sort of test that anyone can easily run locally anyhow, so cleanup after the test is not important)

So add whatever code and/or tags are needed to avoid running stuff in AfterScenario that causes fails.

kiranparajuli589:
> That is in an AfterScenario. For this special test, we only want to run a single scenario. So we can skip any AfterScenario actions. Basically, we throw-away the system-under-test after all the Scenario `Then` steps have finished. (If someone runs this db-convert scenario locally, it is going to expect that their existing install starts with sqlite and will convert them to mysql, assuming they have a mysql database ready etc, so it is not the sort of test that anyone can easily run locally anyhow, so cleanup after the test is not important)
> 
> So add whatever code and/or tags are needed to avoid running stuff in AfterScenario that causes fails.

@phil-davis suggested changes have been implemented. Please review again.

phil-davis (approved):
LGTM. I will cherry-pick and squash this onto the "pending acceptance tests" PR #39514

phil-davis:
cherry-picked to https://github.com/owncloud/core/pull/39514/commits/5f98837bd95d75660fc9fbc9a8802f7db8793a89

PR #39416, opened 2021-10-25
Enable stale action, disable stale bot

JammingBen:
## Description
The stale bot we're using currently is not maintained and has an issue with removing the stale label (see https://github.com/owncloud/core/issues/39399). Therefor we enable the the [GitHub stale action](https://github.com/actions/stale) to handle stale issues.

Note that the GitHub schedule plan is a bit wonky. Scheduled tasks often run minutes after the scheduled time, sometimes they don't run at all (see https://upptime.js.org/blog/2021/01/22/github-actions-schedule-not-working/). We need to see how this behaves for us.

If the schedule makes to many problems, we could either try the Drone plugin (https://github.com/drone-plugins/github-actions) or call the action externally via a scheduled API.

## Related Issue
- Fixes https://github.com/owncloud/core/issues/39399


phil-davis (commented):
Looks fine to me. @micbar are you OK with changing to the GitHub actions stale bot?

xoxys (approved):

PR #39424, opened 2021-10-26
[tests-only] do not upload .gitignore file

micbar:
# Description

- accidentally we copy the .gitignore file from the skeleton to the home folder while "creating a user without skeleton"


v-scharf (approved):
I checked. Now when running api tests, the .gitignore file and other extra files are not created in the `tree -a /Users/scharfviktor/.ocis/storage/users` directory.

PR #39429, opened 2021-10-27
[tests-only] [full-ci] Return early from manuallyAddSkeletonFiles when using emptySkeleton

phil-davis:
## Description
We don't need to run all the logic in `manuallyAddSkeletonFiles` when using `emptySkeleton`.
Avoid trying to create a directory called `emptySkeleton` and then trying to upload the zero files in it.
Do the initialize user logic if using the empty skeleton. On oCIS we previously ended up "initializing" a new user because we always at least uploaded some file in the skeleton (tiny skeleton had `.gitignore` in it). So we preserve that "initialize" behavior for now by doing the `initializeUsers` logic.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -868,6 +868,10 @@ trait Provisioning {
 	 * @throws Exception
 	 */
 	public function manuallyAddSkeletonFiles(array $usersAttributes):void {
+		if ($this->isEmptySkeleton()) {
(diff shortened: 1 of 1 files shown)


phil-davis:
This works OK in https://github.com/owncloud/ocis/pull/2691

individual-it (approved):

PR #39435, opened 2021-10-29
[tests-only] [full-ci] Enable declare(strict_types=1) for FeatureContext

phil-davis:
## Description
We have declared function parameter and return types where we can. Now enable `strict_types=1` and that gets much more checking of the actual types that are passed around.

1) fixed some code so that it really uses and returns the correct types
2) removed explicit casts and checks from places that no longer require it (types are already checked in parameters etc and the IDE knew when these things are now redundant)
3) removed `{}` from around variables in strings like `"some text {$variable} etc"` to stop the IDE  from always telling me that the `{}` are not needed.
4) other minor refactoring

## Related Issue
Part of https://github.com/owncloud/QA/issues/694

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/SetupHelper.php
+++ b/tests/TestHelpers/SetupHelper.php
@@ -338,7 +338,9 @@ class SetupHelper extends \PHPUnit\Framework\Assert {
 			$adminPassword,
 			$xRequestId
 		);
-		return $sysInfo->server_root;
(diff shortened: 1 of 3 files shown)


saw-jan (approved):
LGTM

PR #39444, opened 2021-11-01
[tests-only] [full-ci] waitForDavRequestsToFinish before deleting users

### LINKED ISSUE #501: test fail apiWebdavMove1/moveFileToBlacklistedNameAsync.feature:13
For example, failed in nightly CI:
https://drone.owncloud.com/owncloud/files_primary_s3/2568/93/17
```
  Background:                                                                                      # /var/www/owncloud/testrunner/tests/acceptance/features/apiWebdavMove1/moveFileToBlacklistedNameAsync.feature:7
    Given using new DAV path                                                                       # FeatureContext::usingOldOrNewDavPath()
    And user "Alice" has been created with default attributes and without skeleton files           # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "Alice" has uploaded file with content "ownCloud test text file 0" to "textfile0.txt" # FeatureContext::userHasUploadedAFileWithContentTo()
      HTTP status code was not 201 or 204 while trying to upload file 'textfile0.txt' for user 'Alice'
      Failed asserting that an array contains 423.
    And the administrator has enabled async operations                                             # FeatureContext::triggerAsyncUpload()

  Scenario: rename a file to a filename that is banned by default                                     # /var/www/owncloud/testrunner/tests/acceptance/features/apiWebdavMove1/moveFileToBlacklistedNameAsync.feature:13
    When user "Alice" moves file "/textfile0.txt" asynchronously to "/.htaccess" using the WebDAV API # FeatureContext::userMovesFileUsingTheAPI()
    Then the HTTP status code should be "403"                                                         # FeatureContext::thenTheHTTPStatusCodeShouldBe()
    And user "Alice" should see the following elements                                                # FeatureContext::userShouldSeeTheElements()
      | /textfile0.txt |
```

The background `Given` step seems to often fail with HTTP status 423 when uploading a file for Alice, and only in the 1st scenario of the feature. In later scenarios the file uploads fine.

423 is "Locked" - so maybe some leftover problem from the previous scenario, or?

phil-davis:
## Description
1) If slowdownDavRequests has been used in a test scenario then there
could be DAV requests still "purposely slowly" processing on the
server-under-test. Deleting a user while the user still has such
a request still processing can cause problems, and can effect the
next test scenario. So, before deleting users in AfterScenario
first wait the number of seconds that DAV requests have been slowed
down.

2) Refactored the variable names. "old" and "new" get confused with the "old" and "new" DAV endpoints. Actually the variables refer to previous and current values of scenario settings.

3) Call getOCSPath with string version number - in `substituteInLineCodes` the function `getOCSPath` was being called with an `int` version "number" but actually that should be a string "1" or "2". This was noticed in admin_audit CI: https://drone.owncloud.com/owncloud/admin_audit/2182/26/9
```
    Then the log file should contain at least one entry matching each of these lines: # LoggingContext::logFileShouldContainEntriesMatching()
      | user  | method | url                                                          | message                                                                                               | action     | created | settingName           | settingValue | oldValue | CLI   | actor                             |
      | admin | POST   | /%ocs_path_v1%/apps/testing/api/v1/app/core/shareapi_enabled | user <admin>, displayName <admin> updated appconfig of "core" "shareapi_enabled" from "yes" with "no" | config_set | false   | core.shareapi_enabled | no           | yes      | false | user <admin>, displayName <admin> |
      Type error: Argument 1 passed to FeatureContext::getOCSPath() must be of the type string, int given, called in /var/www/owncloud/testrunner/tests/acceptance/features/bootstrap/FeatureContext.php on line 2980 (Behat\Testwork\Call\Exception\FatalThrowableError)
```

4) Fix type of ldapPort to always be int. Fixes https://drone.owncloud.com/owncloud/user_ldap/3570/27/11
```
Type error: Return value of FeatureContext::getLdapPort() must be of the type int, string returned
```

## Related Issue
- Fixes https://github.com/owncloud/files_primary_s3/issues/501

## How Has This Been Tested?
https://github.com/owncloud/files_primary_s3/pull/503 runs the failing test sequence using the code in this core branch.
Before changing any code, it was failing on most runs. Now it passes.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -72,14 +72,14 @@ trait WebDav {
 	 */
 	private $customDavPath = null;
 
-	private $oldAsyncSetting = null;
+	private $previousAsyncSetting = null;
 
-	private $oldDavSlowdownSetting = null;
+	private $previousDavSlowdownSetting = null;
 
 	/**
 	 * @var int
 	 */
-	private $newDavSlowdownSettingSeconds = 0;
+	private $currentDavSlowdownSettingSeconds = 0;
 
 	/**
 	 * response content parsed from XML to an array
@@ -480,12 +480,12 @@ trait WebDav {
 		} else {
 			$value = 'false';
 		}
-		if ($this->oldAsyncSetting === null) {
-			$oldAsyncSetting = SetupHelper::runOcc(
+		if ($this->previousAsyncSetting === null) {
+			$previousAsyncSetting = SetupHelper::runOcc(
 				['config:system:get', 'dav.enable.async'],
 				$this->getStepLineRef()
 			)['stdOut'];
-			$this->oldAsyncSetting = \trim($oldAsyncSetting);
+			$this->previousAsyncSetting = \trim($previousAsyncSetting);
 		}
 		$this->runOcc(
 			[
@@ -520,12 +520,12 @@ trait WebDav {
 	 * @throws Exception
 	 */
 	public function slowdownDavRequests(string $method, int $seconds):void {
-		if ($this->oldDavSlowdownSetting === null) {
-			$oldDavSlowdownSetting = SetupHelper::runOcc(
+		if ($this->previousDavSlowdownSetting === null) {
+			$previousDavSlowdownSetting = SetupHelper::runOcc(
 				['config:system:get', 'dav.slowdown'],
 				$this->getStepLineRef()
 			)['stdOut'];
-			$this->oldDavSlowdownSetting = \trim($oldDavSlowdownSetting);
+			$this->previousDavSlowdownSetting = \trim($previousDavSlowdownSetting);
 		}
 		OcsApiHelper::sendRequest(
 			$this->getBaseUrl(),
@@ -535,7 +535,7 @@ trait WebDav {
 			"/apps/testing/api/v1/davslowdown/$method/$seconds",
 			$this->getStepLineRef()
 		);
-		$this->newDavSlowdownSettingSeconds = $seconds;
+		$this->currentDavSlowdownSettingSeconds = $seconds;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


phil-davis:
I have rerun CI a few times for demonstration PR https://github.com/owncloud/files_primary_s3/pull/503 and it passes every time.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -76,6 +76,11 @@ trait WebDav {
 
 	private $oldDavSlowdownSetting = null;
 
+	/**
+	 * @var int
+	 */
+	private $newDavSlowdownSettingSeconds = 0;
```
individual-it, line 82:
```suggestion
	private $oldDavSlowdownSettingSeconds = 0;
```
I believe the name refers to the old version of the DAV endpoint

individual-it (changes requested):
should that variable be set back to 0 somewhere?

tests/acceptance/features/bootstrap/WebDav.php
phil-davis: I don't know if `oldDavSlowdownSetting` (line 77) is meaning that it is the `previousDavSlowdownSetting` that was in effect before the scenario started, or if it is a setting that only applies to the "old" DAV endpoint. The actual config key is `dav.slowdown`, which does not mention which of the DAV endpoints it refers to.

"Naming is hard"

individual-it: I was wrong, its called `old` because it's used in `resetOldSettingsAfterScenario` to reset settings after the scenario

individual-it (approved):

phil-davis:
> should that variable be set back to 0 somewhere?

The whole context object is recreated at the start of each scenario, so there is no need to reset class variables in AfterScenario.

PR #39458, opened 2021-11-06
Bump phan PHP analysis v4 to v5

phil-davis:
## Description
`phan` v5 was released a while ago: https://github.com/phan/phan/releases

Update from v4 to v5.

It reported:
```
php -d zend.enable_gc=0 vendor-bin/phan/vendor/bin/phan --config-file .phan/config.php --require-config-exists -p
   analyze ████████████████████████████████████████████████████████████ 100.0% 2180MB/2236MB
lib/private/DB/MigrationService.php:307 PhanTypeNonVarPassByRef Only variables can be passed by reference at argument 1 of \end(array|object &$array)
lib/private/DB/MigrationService.php:339 PhanTypeNonVarPassByRef Only variables can be passed by reference at argument 1 of \end(array|object &$array)
make: *** [Makefile:233: test-php-phan] Error 1
```

And that is correct - there are calls to `\end()` and those should be passed an actual local variable that is an array. I corrected the code.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/DB/MigrationService.php
+++ b/lib/private/DB/MigrationService.php
@@ -304,7 +304,9 @@ class MigrationService {
 			case 'latest':
 				$this->ensureMigrationsAreLoaded();
 
-				return @\end($this->getAvailableVersions());
+				// A variable must be used because \end() must pass the array by reference
+				$availableVersionsArray = $this->getAvailableVersions();
+				return @\end($availableVersionsArray);
 		}
 		return '0';
 	}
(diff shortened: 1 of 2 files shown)


phil-davis:
See the changes to ` lib/private/DB/MigrationService.php`

Will those fix some existing (maybe latent) problem with DB migrations?

If so, then we should identify the problem and I will add a changelog.

If not, then we probably do not need a changleog - this is just an analysis tool update.

jvillafanez (approved):

PR #39463, opened 2021-11-08
[tests-only] Add more error checking in SetupHelper runOcc

phil-davis:
## Description
SetupHelper `runOcc` is used by acceptance test code that sets various settings when the acceptance tests start (among other things). If the system-under-test is not completely set up, not available, responding with some unexpected response etc, then the call to `simplexml_load_string` to format the expected XML of the response might fail and it is currently difficult for the person running the test to understand what might be the problem.

With this PR, if `simplexml_load_string` has a problem then an exception is thrown that reports the attempted `occ` command and the contents of the response. That will help the person running the test to read the response - and the response might have a good clue about what setup is missing from the system-under-test.

## How Has This Been Tested?
Local acceptance test run.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/SetupHelper.php
+++ b/tests/TestHelpers/SetupHelper.php
@@ -890,56 +890,66 @@ class SetupHelper extends \PHPUnit\Framework\Assert {
 		}
 
 		$return = [];
-		$resultXml = \simplexml_load_string($result->getBody()->getContents());
+		$contents = $result->getBody()->getContents();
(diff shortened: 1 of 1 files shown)


kiranparajuli589 (approved):
lgtm 👍🏼

PR #39469, opened 2021-11-09
Karma 6.3.8

phil-davis:
drone is not starting in #39468 - see if it starts here


phil-davis:
That makes drone start in both PRs - they have the same commit at the tip of the branches.

PR #39473, opened 2021-11-09
[tests-only] check correct folder

butonic:
shouldn't this always be `PARENT`?

### DIFF
--- a/tests/acceptance/features/apiAuthWebDav/webDavPUTAuth.feature
+++ b/tests/acceptance/features/apiAuthWebDav/webDavPUTAuth.feature
@@ -41,10 +41,10 @@ Feature: get file info using PUT
     When user "Brian" requests these endpoints with "PUT" including body "doesnotmatter" about user "Alice"
       | endpoint                                       |
       | /remote.php/dav/files/%username%/textfile1.txt |
-      | /remote.php/dav/files/%username%/PARENT       |
+      | /remote.php/dav/files/%username%/PARENT        |
     Then the HTTP status code of responses on all endpoints should be "403"
     When user "Brian" requests these endpoints with "PUT" including body "doesnotmatter" about user "Alice"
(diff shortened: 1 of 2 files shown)


phil-davis:
> shouldn't this always be `PARENT`?

Yes!

https://github.com/owncloud/core/pull/36601/files added those scenarios, and the problem was not noticed.

There are a few more `PARENTS` to fix.

There could be some test scenarios for a different/unauthenticated user trying to access a resource of another user that does not even exist - but they should be separate scenarios with clearly different non-existent resource names.

phil-davis:
I will push another commit to fix the other `PARENTS`

phil-davis (approved):

tests/acceptance/features/apiAuthWebDav/webDavPUTAuth.feature
```diff
@@ -41,11 +41,11 @@ Feature: get file info using PUT
...
       | /remote.php/dav/files/%username%/textfile1.txt |
-      | /remote.php/dav/files/%username%/PARENTS       |
+      | /remote.php/dav/files/%username%/PARENT        |
     Then the HTTP status code of responses on all endpoints should be "403"
     When user "Brian" requests these endpoints with "PUT" including body "doesnotmatter" about user "Alice"
-      | endpoint                                            |
-      | /remote.php/dav/files/%username%/PARENTS/parent.txt |
+      | endpoint                                           |
+      | /remote.php/dav/files/%username%/PARENT/parent.txt |
     Then the HTTP status code of responses on all endpoints should be "409"
```
phil-davis, line 49: I wonder why this gets a 409 (conflict) for the file inside a folder, but 403 (forbidden) when accessing a file or folder in the top-level root?

This passes on oC10, but is it what webDAV is supposed to do?

PR #39482, opened 2021-11-11
Load l10n js files from correct app folder

AlexAndBear:
## Description
Bugfix: Load l10n js files from the correct app folder

With this PR, we ensure that the translations will be loaded from the correct
app folder path.
For example, if you have two versions of an app, where one version is in the
apps folder and the other one in the apps-external folder.
The change also ensures that theme translations will be loaded additionally
instead of replacing the whole app translations.
This has the advantage that the user can cherry-pick single words to translate.

## Reproduction steps
1. Download the ownCloud 10.8 server (files_mediaviewer is located in the apps folder)
2. Download the files_mediaviewer 1.0.5 (RC release)
3. Extract files_mediaviewer into the apps-external folder (this is the location, where it typically lands, if you install the new version via market app)
4. Open ownCloud in the web browser
5. Open the network tab
6. files_medaviewer js files will be loaded from **apps-external**/files_mediaviewer/*.js (correct)
7. files_medaviewer language js files will be loaded from **apps**/files_mediaviewer/*.js (**incorrect**)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Template/JSResourceLocator.php
+++ b/lib/private/Template/JSResourceLocator.php
@@ -40,6 +40,9 @@ class JSResourceLocator extends ResourceLocator {
 		}
 
 		if (\strpos($script, '/l10n/') !== false) {
+			$app = \substr($fullScript, 0, \strpos($fullScript, '/'));
+			$appFolderLocation = \explode('/', $this->appManager->getAppWebPath($app))[1] ?? 'apps';
+
 			// For language files we try to load them all, so themes can overwrite
 			// single l10n strings without having to translate all of them.
 			$found = 0;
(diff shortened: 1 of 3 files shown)


phil-davis:
I confirmed with the announcementcenter app. That has `js/script.js` which, among other things, has `t('announcementcenter', 'Delete')`

I cloned the latest git of announcement center into both `apps` and `apps-external` and enabled the app. In `apps-external` I increased the version of the app in `apps-external/announccementcenter/appinfo/info.xml` (to make sure that the system should be using the version in `apps-external`)

Then change the translation of "Delete" in `apps-external/announcementcenter/l10n/de.js`

1) Checkout core master:
Login as admin and create an announcement. Hover over the announcement, a "Delete" button is displayed with the old text (from the app in `apps`)

2) Checkout this core load-l10n-js-files-from-actual-app-folder:
Login as admin and create an announcement. Hover over the announcement, a "Delete" button is displayed with the new text (from the app in `apps-external`)

The code in this PR branch fixes the problem.

This will fix this situation for various apps that have this kind of JS code.

phil-davis (approved):
Works for me

JanAckermann:
We might need to think about what happens if a theme is set, usually people don't have an apps external folder in the theme, so we might only resolve the actual app path if no theme is present

JanAckermann:
@phil-davis Improved the PR. before theme translations replaced the whole app translations, this was bad behaviour as we need to have all the strings in the theme translations file. 
Now the theme translations will be loaded additionally

lib/private/Template/JSResourceLocator.php
```diff
@@ -40,14 +40,31 @@ public function doFind($script) {
 		}
 
 		if (\strpos($script, '/l10n/') !== false) {
+			$app = \substr($fullScript, 0, \strpos($fullScript, '/'));
```
jvillafanez, line 43: Any idea if something like `./myscript.js` or `../js/myscript.js` is properly handled? It might be handled elsewhere...

jvillafanez (approved):
Code looks fine. I'm assuming there is no problem handling the paths.

PR #39489, opened 2021-11-15
[tests-only][full-ci]Fix type of functions

SwikritiT:
## Description
   ```
Then as "Alice" the file with original path "/PARENT/textfile0.txt" should not exist in the trashbin                    # TrashbinContext::elementIsNotInTrashCheckingOriginalPath()
      Type error: ltrim() expects parameter 1 to be string, null given (Behat\Testwork\Call\Exception\FatalThrowableError)
```
Errors like this were reported while bumping the commit id. So this PR fixes functions so that the tests pass.

## How Has This Been Tested?

- CI 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/TUSContext.php
+++ b/tests/acceptance/features/bootstrap/TUSContext.php
@@ -146,13 +146,14 @@ class TUSContext implements Context {
 	 * @param string $checksum
 	 *
 	 * @return void
-	 * @throws ReflectionException
-	 * @throws GuzzleException
 	 * @throws ConnectionException
+	 * @throws GuzzleException
+	 * @throws JsonException
+	 * @throws ReflectionException
 	 * @throws TusException
 	 */
 	public function userUploadsUsingTusAFileTo(
-		string $user,
+		?string $user,
 		string $source,
 		string $destination,
 		array $uploadMetadata = [],
@@ -198,7 +199,7 @@ class TUSContext implements Context {
 		} elseif ($noOfChunks === 1 && $bytes === null) {
 			$client->file($sourceFile, $destination)->upload();
 		} else {
-			$bytesPerChunk = \ceil(\filesize($sourceFile) / $noOfChunks);
+			$bytesPerChunk = (int)\ceil(\filesize($sourceFile) / $noOfChunks);
 			for ($i = 0; $i < $noOfChunks; $i++) {
 				$client->upload($bytesPerChunk);
 			}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by SwikritiT** [comment]
The build is passing https://drone.owncloud.com/owncloud/ocis/7806. So after the merge of this PR we can update the commit id of ocis and reva as well and they'll pass CI.

File: tests/acceptance/features/bootstrap/TUSContext.php [hunk 1]
```diff
@@ -141,18 +141,19 @@ public function sendsAChunkToTUSLocationWithOffsetAndData(string $user, string $
 	 *                              `Upload-Metadata` header.
 	 *                              see https://tus.io/protocols/resumable-upload.html#upload-metadata
 	 *                              Don't Base64 encode the value.
-	 * @param int $noOfChunks
+	 * @param float $noOfChunks
```
phil-davis, line 144: This is strange??? It is still `int` in the function declaration.

tests/acceptance/features/bootstrap/TUSContext.php
```diff
@@ -200,7 +201,7 @@ public function userUploadsUsingTusAFileTo(
 		} else {
 			$bytesPerChunk = \ceil(\filesize($sourceFile) / $noOfChunks);
```
phil-davis, line 202: Note for my own benefit!
https://www.php.net/manual/en/function.ceil.php
"num rounded up to the next highest integer. The return value of ceil() is still of type float as the value range of float is usually bigger than that of int."

So we really do need to cast to `int` when we want to use it as an `int`

tests/acceptance/features/bootstrap/TUSContext.php
```diff
@@ -229,30 +230,32 @@ public function userUploadsAFileWithContentToUsingTus(
 				$destination
 			);
 		} catch (Exception $e) {
-			Assert::assertStringContainsString('TusPhp\Exception\FileException: Unable to create resource', $e);
+			Assert::assertStringContainsString('TusPhp\Exception\FileException: Unable to create resource', (string)$e);
```
phil-davis, line 233: https://www.php.net/manual/en/exception.tostring.php
That outputs text like:
exception 'Exception' with message 'Some error message' in ...

The exception message is in there, so this works. We could also compare to just the $e->message() - but the code is also OK as it is.

phil-davis (approved):

PR #39494, opened 2021-11-16
Add login.token_auth_enforced parameter to disallow password based logins

### LINKED ISSUE #39484: Disabling password auth all together
Hey,

I've enabled OIDC Authentication in my private ownCloud yesterday, now I'm trying to make sure password based login is not possible anymore.

I've found [token_auth_enforced](https://doc.owncloud.com/server/next/admin_manual/configuration/server/config_sample_php_parameters.html#enforce-token-only-authentication-for-apps-and-clients-connecting-to-owncloud) config parameter which disables password based access to webdav and other apps but not for the regular browser based login.
 
Then there is the  [autoRedirectOnLoginPage](https://doc.owncloud.com/server/next/admin_manual/configuration/server/config_apps_sample_php_parameters.html#app-openid-connect-oidc) setting for the oidc-app which redirects the user to my IdP automatically. As this prevents the csrf `requesttoken` to be delivered (and hopefully it's not generated in the first place) there is probably no way to forge a custom request to login without the form, right?

So can I consider it impossible to login to my ownCloud just with password now or is there any additional hardening I can/must do? 

I considered hacking a `return false;` into the Database-UserBackend here at the top: https://github.com/owncloud/core/blob/2eead619a0e1bab2ca21bc7bbc4ef77f9d82c42d/lib/private/User/Database.php#L217-L235
or better here at the bottom:
https://github.com/owncloud/core/blob/2eead619a0e1bab2ca21bc7bbc4ef77f9d82c42d/lib/private/User/Session.php#L326-L341

If guarded by a config parameter, could that be interesting to upstream?
...as a sort of double bottom if the redirect did not work for whatever reason and the form was still rendered.. additional benefit: we could also hide the form on the login page and still show additional logins - that means we could have multiple non-password logins and don't need to rely on a redirect in an app.

Alternatively is there a way to disable certain authentication backends or can they only be explicitly [enabled](https://doc.owncloud.com/server/next/admin_manual/configuration/server/config_sample_php_parameters.html#code-sample-32)

What do you think? I'd be willing to send a PR, if you consider it valuable. Suggestions for config parameter name would be more than welcome :-D 

As Green Day put it: Am I just paranoid or am I just stoned?

dschmidt:
… completely.

## Description
This prevents the login formular from rendering and sessions from being created even for handcrafted requests (even if csrf protection is disabled).

## Related Issue
- Fixes https://github.com/owncloud/core/issues/39484

## Motivation and Context
As described in the original ticket, I'm paranoid and want to disable password based logins after enabling two factor authentication on my personal ownCloud instance.

What do you think? Especially of the parameter name? 
`token_auth_enforced` is used for everything but the browser login, `login.alternatives` is used for adding the buttons - so that namespace and the analogous parameter name seemed to kinda make sense to me, but I have no strong opinion about this.

## How Has This Been Tested?
In my regular development setup :)

Screenshot_20211116_220342

With the customer portal it obviously doesn't make sense, because you still need to login after registering - but I the button is analogous to the OIDC (e.g.) login button.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -337,6 +337,11 @@ class Session implements IUserSession, Emitter {
 		if ($this->validateToken($password, $uid)) {
 			return $this->loginWithToken($password);
 		}
+		if ($this->isLoginTokenAuthEnforced()) {
+			$this->logger->warning("Login failed because token auth is enforced in browser: '$uid'", ['app' => 'core']);
+			$this->emitFailedLogin($uid);
+			throw new PasswordLoginForbiddenException();
+		}
 		return $this->loginWithPassword($uid, $password);
 	}
 
@@ -398,6 +403,10 @@ class Session implements IUserSession, Emitter {
 
 	private function isTokenAuthEnforced() {
 		return $this->config->getSystemValue('token_auth_enforced', false);
+	}
+
+	private function isLoginTokenAuthEnforced() {
+		return $this->config->getSystemValue('login.token_auth_enforced', false);
 	}
 
 	protected function isTwoFactorEnforced($username) {
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -169,6 +169,7 @@ class LoginController extends Controller {
 			$altLogins = \array_merge($altLogins, $altLogins2);
 		}
 		$parameters['alt_login'] = $altLogins;
+		$parameters['login.token_auth_enforced'] = $this->config->getSystemValue('login.token_auth_enforced', false);
 		$parameters['rememberLoginAllowed'] = OC_Util::rememberLoginAllowed();
 		$parameters['rememberLoginState'] = !empty($remember_login) ? $remember_login : 0;
 
--- a/core/templates/login.php
+++ b/core/templates/login.php
@@ -10,6 +10,7 @@ script('core', [
 ?>
 
 <!--[if IE 8]><style>input[type="checkbox"]{padding:0;}</style><![endif]-->
+<?php if (empty($_['login.token_auth_enforced'])) { ?>
 <form method="post" name="login" autocapitalize="none">
 <?php if (!empty($_['accessLink'])) {
 	?>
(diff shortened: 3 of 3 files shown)


lib/private/User/Session.php
```diff
@@ -337,6 +337,11 @@ public function login($uid, $password) {
 		if ($this->validateToken($password, $uid)) {
 			return $this->loginWithToken($password);
 		}
+		if ($this->isBrowserTokenAuthEnforced()) {
```
phil-davis, line 340:
```suggestion
		if ($this->isLoginTokenAuthEnforced()) {
```

phil-davis:
https://github.com/owncloud/core/tree/work/login.token_auth_enforced
`This branch is 1 commit ahead, 1074 commits behind master. `
Please start from a current core master branch.

lib/private/User/Session.php
dschmidt: Oops, yeah, renamed it last minute.

dschmidt:
Rebased

mmattel:
This is docs relevant, please file a docs issue when close to merge.

dschmidt:
Ping?

phil-davis:
https://drone.owncloud.com/owncloud/core/34822/10/7
```
There were 13 failures:

1) Tests\Core\Controller\LoginControllerTest::testResponseForNotLoggedinUser
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
         'alt_login' => Array ()
         'rememberLoginAllowed' => false
         'rememberLoginState' => 0
-        'strictLoginEnforced' => false
+        'strictLoginEnforced' => null
+        'login.token_auth_enforced' => null
     )
     'renderAs' => 'guest'
     'appName' => 'core'
...
```

mmattel:
@DeepDiver1975 maybe you can have a look on this

DeepDiver1975:

Issue #39533, opened 2021-11-26
Creating a group with space at end does not work


phil-davis:
### Steps to reproduce
```
  Background:
    Given using OCS API version "2"

  @issue-39533 @skipOnOcis
  Scenario: admin creates a group that has white space at the end of the name
    When the administrator sends a group creation request for group "white-space-at-end " using the provisioning API
    Then the OCS status code should be "200"
    And the HTTP status code should be "200"
    # Note: it seems that white space at the end of a group name gets stripped off
    # Groups "white-space-at-end " and "white-space-at-end" seem to be effectively the same
    And group "white-space-at-end " should exist
    And group "white-space-at-end" should exist
```

### Expected behaviour
I am not sure, but it should be consistent.
Maybe white-space at the end of a group name should not be allowed, and the initial group creation request should fail?
Maybe it should be allowed, and in that case the groups "abc" and "abc " should be 2 different groups.

### Actual behaviour
When a group is created with white-space on the end of the name, the request seems to work.
But when querying the group it looks like 2 groups exist - both the group with and without the white-space at the end.

### Server configuration
current core master

phil-davis:
Note1: space at the start of a group name is currently allowed, and the 2 groups "abc" and " abc" get created and work as 2 different groups.

Note2: the current implementation of the Provisioning API on oCIS seems to allow space at either or both ends of a grup name, and treats all such groups as distinct/different.

phil-davis:
PR #39540 prohibits space at the start or end of a group name.

individual-it:
@phil-davis anything left to do here or close the issue?

phil-davis:
white space at the beginning or end of a group name is now not allowed. The code for that was reviewed in PR #39540 and ended-up in #39612 which has now been merged to master.

PR #39530, opened 2021-11-25
[tests-only] Add tests to demonstrate group names with leading and trailing spaces

phil-davis:
## Description
This adds extra acceptance test scenarios that demonstrate the current behavior when adding groups with leading or trailing spaces in the name, or the group name " " (a single space).

Needs input about if the current behavior or some different behavior is required.

## Related Issue
https://github.com/owncloud/enterprise/issues/4890

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39514, opened 2021-11-22
[tests-only] Acceptance test changes pending for master

phil-davis:
## Description
Currently we are in the core 10.9.0 release process. There is a `release-10.9.0` branch and there will be some behavior and test changes that will be done in that branch during the process. So we don't want to make unnecessary test core changes in core master for now. That might cause conflicts with the release branch.

We will add commits here for test code changes that will be useful to run for `owncloud/ocis` and `cs3org/reva`. After 10.9.0 release is stable we can decide when to merge this test branch to core master.

1) Improve checks in tests for disabling a user - related to issues https://github.com/owncloud/ocis/issues/1420 and #39512 - this provides extra checks in test scenarios that will help the oCIS test suite to fail for the right reason on these test scenarios.

2) Reduce use of "download range" in acceptance tests - part of #39513

3) Fix SetupHelper runOcc to use strings only in exception messages

4) Add tests to demonstrate group names with leading and trailing spaces

5) Adjust group tests for issue #39533

6) Refactor getGroups tests to not depend on the starting groups - this removes the need to list the pre-existing groups in a system-under-test.

7) Fix missing HTTP status code in createPublicLinkShare test scenarios

8) Use when step for the test scenario with only Then. (small refactoring of `webUIAdminSettings/adminGeneralSettings.feature` )

9) Fixed order dependent flaky test detected in files_classifier app (cherry-picked from PR #39548 )

10) add tests for db-convert occ command (cherry-picked and squashed from PR #39393 )

11) Do not allow white space at the start or end of a group name (cherry-picked and squashed from PR #39540 ) Having this here will allow the API tests to run in oCIS, and group name validation rules can be corrected there as needed.

12)  Add bug demonstration test for sending PUT request to other user's webdav endpoint

13) refactor test helpers and contexts for parallel deployment tests (code from #39603 )

14) skipOnEncryption dbConversion.feature as it needs special CI setup - skips the dbConversion scenario, because that is a very special scenario that requires starting with sqlite installed.

15) added tests for files_external occ command - cherry-picked from PR #39511 

16-17) Add spaces as a new webdav url type - merged from #39599 

18) Add extra checks to getPersonalSpaceIdForUser  - to better report when a space can't be found

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
[...]

PR #39540, opened 2021-11-30
Do not allow white space at the start or end of a group name

phil-davis:
## Description
Do not allow a group to be created with a name that has white space:
- at the beginning, or
- at the end, or
- has only white space

This is on top of PR #39530 which demonstrates the current behavior for groups with white space at the beginning or end of the group name.

The Provisioning API acts as if "finance" and "finance " are the same group, because of:
https://stackoverflow.com/questions/17876478/why-the-sql-server-ignore-the-empty-space-at-the-end-automatically
https://support.microsoft.com/en-us/topic/inf-how-sql-server-compares-strings-with-trailing-spaces-b62b1a2d-27d3-4260-216d-a605719003b0

1) If I create group "finance" first, then try to create group "finance ", the groupExists checks find that "finance " exists already, and refuses to create it.

2) If I create group "finance " first, then try to create group "finance", the groupExists checks find that "finance" exists already, and refuses to create it.

Item (1) seems a reasonable "accident" of the code - it prevents creating an extra group name with a space on the end.

Item (2) is a bit odd - if we accidentally create group "finance " then we are stuck with "finance " in the database. And when we do an API request like:
```
curl -X GET http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/groups
```

We get elements with that space at the end of the name like:
```
<element>finance </element>
```

And to query the group directly we have to:
```
curl -X GET http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/groups/finance%20
```

Also: https://jira.atlassian.com/browse/CWD-4290
That claims that pgsql will behave differently than mySQL/MariaDb, and likely happily add separate table rows for "finance" and "finance " groups, so they will be treated as different groups in an oC10 installation using pgsql.

So that is another reason to refuse to create groups (or any string data) with trailing spaces - different databases, and implementations not using an SQL database, might behave differently.

Therefore this PR prevents creation of a group with a name that ends in white space, including a group name that is only white space.

It seems silly to allow group names that start with white space, like " finance", that would be strangely inconsistent with banning white space at the end of the name. So this PR also prevents white space at the start of a group name.

As well as unit test cases, I have added API acceptance test scenarios, because we want oCIS to behave the same. IMO oCIS currently allows spaces at the beginning and end, and considers all combinations like " finance", "finance" and "finance " to be separate valid groups.

## Related Issue
https://github.com/owncloud/enterprise/issues/4890

#39533 - sorts out the requirements/rules for that issue.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/provisioning_api/lib/Groups.php
+++ b/apps/provisioning_api/lib/Groups.php
@@ -136,6 +136,10 @@ class Groups {
 			\OCP\Util::writeLog('provisioning_api', 'Group name not supplied', \OCP\Util::ERROR);
 			return new OC_OCS_Result(null, 101, 'Invalid group name');
 		}
+		if (\trim($groupId) !== $groupId) {
+			\OCP\Util::writeLog('provisioning_api', 'Group name must not start or end with white space', \OCP\Util::ERROR);
+			return new OC_OCS_Result(null, 101, 'Invalid group name');
+		}
 		// Check if it exists
 		if ($this->groupManager->groupExists($groupId)) {
 			return new OC_OCS_Result(null, 102);
--- a/apps/provisioning_api/tests/GroupsTest.php
+++ b/apps/provisioning_api/tests/GroupsTest.php
@@ -333,11 +333,26 @@ class GroupsTest extends \Test\TestCase {
 		$this->assertEquals([], $result->getData());
 	}
 
-	public function testAddGroupEmptyGroup() {
+	public function dataAddGroupWithInvalidName() {
+		return [
+			[''],
+			[' '],
+			[' white-space-at-start'],
+			['white-space-at-end '],
+			[' white-space-at-both-ends '],
+		];
+	}
+
+	/**
+	 * @dataProvider dataAddGroupWithInvalidName
+	 *
+	 * @param string $groupName
+	 */
+	public function testAddGroupWithInvalidName($groupName) {
 		$this->request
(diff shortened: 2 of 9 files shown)


phil-davis:
I have set this to "Draft" because we do not want to merge stuff to master right now. But I would appreciate a review of this. There is currently edge-case differences between oC10 (with MySQL/MariaDb) and oCIS, particularly for the case of groups "finance" and "finance "

oC10 effectively treats those names as "the same thing". oCIS treats them as separate groups. (And I think oC10 with pgsql treats them as separate groups)

This PR defines the behavior for this crud. So please review and give your feedback.

If we agree with what I have done here, then I can cherry-pick this into #39514 (where I am accumulating stuff that is waiting to go in `master`, and is being used in oCIS CI) And oCIS can be refactored to match the behavior in these API acceptance test scenarios.

JammingBen (approved):
LGTM, I think it's a good thing as well.

jnweiger:
LGTM too. Let's make group name rules consistent with user name rules, where exavtly that is already forbidden
image

AlexAndBear (approved):

phil-davis:
Merged to master in PR #39612

PR #39554, opened 2021-12-04
Fix so that a null appconfig value is treated as an empty string

### LINKED ISSUE #39553: occ config:app:set stores an empty string that cannot be overwritten when called without --value
Seen in 10.9.0 beta1 

```
# occ config:app:delete foo bar
# occ config:app:set foo bar
Config value bar for app foo set to 
# occ config:app:get foo bar

# occ config:app:set foo bar --value='hammer'
Config value bar for app foo set to hammer
# occ config:app:get foo bar

#
```

The second `set` command reports that it did set the value to `hammer`, but in reality it failed.

Expected behaviour:
* Invoking config:app:set without --value should print a syntax error.
* Correct commands that fail should print an error, not a success message.

Workaround: `occ config:app:delete foo bar`

phil-davis:
## Description
If an app config key has been set to value NULL then make sure to allow it to be updated to some other value (empty string, or some string...)

This fixes the SQL logic that updates values in the `appconfig` table. The database has already been able to store the value NULL. This PR does not change that ability. The PR makes changing the value to and from NULL work correctly.

https://stackoverflow.com/questions/16186674/mysql-syntax-not-evaluating-not-equal-to-in-presence-of-null has some information about how the database handles this sort of thing.

## Related Issue
- Fixes #39553 

## How Has This Been Tested?
```
$ php occ config:list someapp
{
    "apps": {
        "someapp": []
    }
}
$ php occ config:app:set someapp somename
Config value somename for app someapp set to 
$ php occ config:list someapp
{
    "apps": {
        "someapp": {
            "somename": null
        }
    }
}
$ php occ config:app:set someapp somename --value "somevalue"
Config value somename for app someapp set to somevalue
$ php occ config:list someapp
{
    "apps": {
        "someapp": {
            "somename": "somevalue"
        }
    }
}
```

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -185,9 +185,16 @@ class AppConfig implements IAppConfig {
 			 * http://docs.oracle.com/cd/E11882_01/server.112/e26088/conditions002.htm#i1033286
 			 * > Large objects (LOBs) are not supported in comparison conditions.
 			 */
-			if (!($this->conn instanceof \OC\DB\OracleConnection)) {
-				// Only update the value when it is not the same
-				$sql->andWhere($sql->expr()->neq('configvalue', $sql->createParameter('configvalue')))
+			if (!($this->conn instanceof \OC\DB\OracleConnection) && ($value !== null)) {
+				// Only update the value if:
+				// - it is not the same as the value currently in the database, or
+				// - the value in the database is currently NULL
+				// Note: the SQL comparison operator "<>" neq() does not think that "string" is not equal to NULL.
+				//       so we need to also explicitly update if the current row value is null.
+				$or = $sql->expr()->orX();
+				$or->add($sql->expr()->neq('configvalue', $sql->createParameter('configvalue')));
+				$or->add($sql->expr()->isNull('configvalue'));
+				$sql->andWhere($or)
 					->setParameter('configvalue', $value);
 			}
 
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -150,6 +150,10 @@ class AppConfig implements IAppConfig {
 	 * @return bool True if the value was inserted or updated, false if the value was the same
 	 */
 	public function setValue($app, $key, $value) {
+		if ($value === null) {
+			$value = '';
+		}
+
 		return $this->emittingCall(function (&$afterArray) use (&$app, &$key, &$value) {
 			if (!$this->hasKey($app, $key)) {
 				$inserted = (bool) $this->conn->insertIfNotExist('*PREFIX*appconfig', [
@@ -185,7 +189,7 @@ class AppConfig implements IAppConfig {
 			 * http://docs.oracle.com/cd/E11882_01/server.112/e26088/conditions002.htm#i1033286
 			 * > Large objects (LOBs) are not supported in comparison conditions.
 			 */
-			if (!($this->conn instanceof \OC\DB\OracleConnection) && ($value !== null)) {
+			if (!($this->conn instanceof \OC\DB\OracleConnection)) {
 				// Only update the value if:
 				// - it is not the same as the value currently in the database, or
 				// - the value in the database is currently NULL
@@ -321,7 +325,7 @@ class AppConfig implements IAppConfig {
 			) {
 				$row['configvalue'] = '0.0.1';
 			}
-			$this->cache[$row['appid']][$row['configkey']] = $row['configvalue'];
+			$this->cache[$row['appid']][$row['configkey']] = ($row['configvalue'] === null) ? '' : $row['configvalue'];
 		}
 		$result->closeCursor();
 
--- a/tests/lib/AppConfigTest.php
+++ b/tests/lib/AppConfigTest.php
@@ -241,30 +241,55 @@ class AppConfigTest extends TestCase {
 		$this->assertConfigKey('testapp', 'installed_version', '1.33.7');
 	}
 
-	public function dataSetValueUpdateEmptyValues() {
+	public function dataSetValue() {
 		return [
 			[null, ''],
-			[null, 'non-empty'],
-			['', null],
+			['', ''],
+			['non-empty', 'non-empty'],
+		];
+	}
+
+	/**
+	 * @dataProvider dataSetValue
+	 * @param string|null $inputValue
+	 * @param string|null $returnedValue
+	 */
+	public function testSetValue($inputValue, $returnedValue) {
+		$config = new \OC\AppConfig(\OC::$server->getDatabaseConnection());
+
+		$config->setValue('someapp', 'somekey', $inputValue);
+		$this->assertSame(
+			$returnedValue,
+			$config->getValue('someapp', 'somekey', $returnedValue)
+		);
+	}
+
+	public function dataSetValueUpdateValues() {
+		return [
 			['', 'non-empty'],
 			['non-empty', ''],
-			['non-empty', null],
 			['non-empty', 'different-non-empty-string'],
 		];
 	}
 
 	/**
-	 * @dataProvider dataSetValueUpdateEmptyValues
+	 * @dataProvider dataSetValueUpdateValues
 	 * @param string|null $firstValue
 	 * @param string|null $updatedValue
 	 */
-	public function testSetValueUpdateEmptyValues($firstValue, $updatedValue) {
+	public function testSetValueUpdateValues($firstValue, $updatedValue) {
 		$config = new \OC\AppConfig(\OC::$server->getDatabaseConnection());
 
 		$config->setValue('someapp', 'somekey', $firstValue);
-		$this->assertConfigKey('someapp', 'somekey', $firstValue);
+		$this->assertSame(
+			$firstValue,
+			$config->getValue('someapp', 'somekey', $firstValue)
+		);
 		$config->setValue('someapp', 'somekey', $updatedValue);
-		$this->assertConfigKey('someapp', 'somekey', $updatedValue);
+		$this->assertSame(
+			$updatedValue,
+			$config->getValue('someapp', 'somekey', $updatedValue)
+		);
 	}
 
 	public function testSetValueInsert() {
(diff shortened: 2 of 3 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareOperationsToRoot1-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33894/67/1

ownclouders:

https://drone.owncloud.com/owncloud/core/33894/82/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiWebdavLocksUnlock-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33894/91/1

ownclouders:

https://drone.owncloud.com/owncloud/core/33894/51/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiAuthOcs-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33894/48/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareOperationsToShares1-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33894/69/1

ownclouders:
:boom: Acceptance tests pipeline <strong>apiAuth-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33894/47/1

phil-davis:
Note: 
```
	/**
	 * Sets
	 * 
	 *  a value. If the key did not exist before it will be created.
	 *
	 * @param string $app app
	 * @param string $key key
	 * @param string|float|int $value value
	 * @return bool True if the value was inserted or updated, false if the value was the same
	 */
	public function setValue($app, $key, $value) {
```

The PHPdoc says that value should be string, float or int. But actually the value `null` also works fine and gets stored in the database. That happens when someone does:
`occ config:app:set someapp somekey`

When they do not mention a value, it gets set to `null`.

I have not changed that existing behavior. Maybe some app sets a key with value `null` for some purpose?

Should we actually prevent the value ever being `null`?

lib/private/AppConfig.php
```diff
@@ -185,9 +185,17 @@ public function setValue($app, $key, $value) {
...
-			if (!($this->conn instanceof \OC\DB\OracleConnection)) {
-				// Only update the value when it is not the same
-				$sql->andWhere($sql->expr()->neq('configvalue', $sql->createParameter('configvalue')))
+			if (!($this->conn instanceof \OC\DB\OracleConnection) && ($value !== null)) {
+				// Only update the value if:
+				// - it is not the same as the value currently in the database, or
+				// - the value in the database is currently NULL
+				// Note: the SQL comparison operator "<>" neq() does not think that "string" is not equal to NULL.
+				//       so we need to also explicitly update if the current row value is null.
+				$orBuilder = $this->conn->getQueryBuilder();
```
JammingBen, line 194: Is it necessary to fetch another query builder instance here? I think you could just use `$sql` again.

tests/lib/AppConfigTest.php
```diff
@@ -239,9 +239,31 @@ function (GenericEvent $event) use (&$calledAfterSetValue) {
...
+			['', 'non-empty'],
+			['non-empty', ''],
+			['non-empty', null],
+			['non-empty', 'different-non-empty-string'],
+		];
+	}
+
+	/**
+	 * @dataProvider dataSetValueUpdateEmptyValues
+	 * @param mixed $value
```
JammingBen, line 258:
```suggestion
	 * @param string|null $firstValue
	 * @param string|null $updatedValue
```

JammingBen (commented):
Just some minor nitpicks 👍

lib/private/AppConfig.php
phil-davis: I will check...

phil-davis: Correct - I  removed `$orBuilder`

tests/lib/AppConfigTest.php
phil-davis: Fixed

tests/lib/AppConfigTest.php
```diff
@@ -456,6 +479,6 @@ protected function assertConfigKey($app, $key, $expected) {
 		$actual = $query->fetch();
 		$query->closeCursor();
 
-		$this->assertEquals($expected, $actual['configvalue']);
+		$this->assertSame($expected, $actual['configvalue']);
```
phil-davis, line 482: Note: `assertSame` is more strict - it asserts that both the type and value are the same.
Without this change, the unit test case `[null, ''],` above would still pass, even without the code fix.
Now if I remove the code fix then both of the first two test cases fail, which is what the behavior was prior to this PR.

JammingBen (approved):

phil-davis:
https://drone.owncloud.com/owncloud/core/33914/36/8
```
There were 4 failures:

1) Test\AppConfigTest::testSetValueUpdateEmptyValues with data set #0 (null, '')
Failed asserting that null is identical to ''.

/drone/src/tests/lib/AppConfigTest.php:482
/drone/src/tests/lib/AppConfigTest.php:267

2) Test\AppConfigTest::testSetValueUpdateEmptyValues with data set #2 ('', null)
Failed asserting that null is identical to ''.

/drone/src/tests/lib/AppConfigTest.php:482
/drone/src/tests/lib/AppConfigTest.php:265

3) Test\AppConfigTest::testSetValueUpdateEmptyValues with data set #3 ('', 'non-empty')
Failed asserting that null is identical to ''.

/drone/src/tests/lib/AppConfigTest.php:482
/drone/src/tests/lib/AppConfigTest.php:265

4) Test\AppConfigTest::testSetValueUpdateEmptyValues with data set #4 ('non-empty', '')
Failed asserting that null is identical to ''.

/drone/src/tests/lib/AppConfigTest.php:482
/drone/src/tests/lib/AppConfigTest.php:267
```

I will think about what happens with Oracle...

tests/lib/AppConfigTest.php
phil-davis: Now this brings out the fact that Oracle will not store an empty string in a string field of a database table. Oracle always stores `null` in this case. So changing the value from `null` to the empty string (or vice-versa) will result in no change in Oracle. And changing from 'somestring' to the empty string will result in the value being null.

https://stackoverflow.com/questions/13278773/null-vs-empty-string-in-oracle

phil-davis:
@JammingBen see my comment https://github.com/owncloud/core/pull/39554/files#r762919447

Oracle does not see any difference between `null` and the empty string - if we pass it an empty string to store, then it actually stores `null`. So there seems no hope in Oracle that users could purposely choose to set an app config key to value `null` or value empty-string and that they be different things that could result in different app behavior.

What to do?

Maybe the interface to AppConfig->setValue should refuse to accept `null` as a value (the current PHPdoc does not say that `null` is OK). (Or treat `null` just like the empty string).

And AppConfig->getValue can consistently return the empty string when it finds either the empty string or `null` in the database table. (Effectively eliminating `null`)

I can't think why any app would need to specially distinguish between the empty string and `null`

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIMobileSize-3-1-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33921/167/1

JammingBen:
Hmm, I am torn on this one.

Probably there are no apps that need to distinguish between `null` and an empty string, but I can't say that for sure. Also, I'm not the biggest fan of setting a value and then silently changing it to something else. I understand that Oracle does that anyways, but it's only an "Oracle thing" for now.

I agree that we don't have a clean solution right now, however, my main concern is that we accidentally break something with this change.

phil-davis:
Changed to draft. I will also make a PR that keeps the `null` stuff as much as possible. The unit tests of that will demonstrate that a different result happens on Oracle, and that Oracle will not support the `null` setting.

phil-davis:
> Hmm, I am torn on this one.

Alternative is PR #39574

jnweiger:
> Also, I'm not the biggest fan of setting a value and then silently changing it to something else
I agree with that in general.
But we are not changing random bits here, we are defining an equivalence that was previously undefined and that appears helpful.

I'd suggest that both
`occ config:app:set foo bar`
and
`occ config:app:set foo bar --value=""`
shall be defined as equivalents.
(Or even make the first form a deprecated syntax)

I believe this PR here implements my view of the world.

lib/private/AppConfig.php
```diff
@@ -150,6 +150,10 @@ public function hasKey($app, $key) {
 	 * @return bool True if the value was inserted or updated, false if the value was the same
 	 */
 	public function setValue($app, $key, $value) {
+		if ($value === null) {
+			$value = '';
```
jnweiger, line 154: Is this rewriting really needed?
As far as I can see, we, when we retrieve and compare values we'll treat null and '' the same anyway.

jnweiger (approved):
This PR fixes my issue and
this PR makes ownCloud behave the same, weather the underlying db is Oracle or not. GOOD.

(The other PR only fixes my issue, but does not provide an abstraction for the Oracle case.)

lib/private/AppConfig.php
JammingBen: I think so, too. Unit tests succeed as well when removing this block. Do you agree @phil-davis ?

phil-davis: Strictly speaking, this if block is not needed. The only advantage that I thought of is that it means that the database entries will be consistent. In non-Oracle the database will always store the empty string, and the class interface will always return the empty string. For someone debugging something in future, it might avoid them wondering why the database has a mix of `null` and empty string.

(Oracle will always store `null`, but that's a different thing.

Should I remove this code?

jnweiger: I am fine keeping it in. I was more after that explanation. Thanks!

PR #39574, opened 2021-12-08
Process null app config value

### LINKED ISSUE #39553: occ config:app:set stores an empty string that cannot be overwritten when called without --value

phil-davis:
## Description
An app config value that is set to null can now be updated to a not-null value, for example, with the occ config:app:set command.

Note: when Oracle stores an app config value of the empty string '' it actually stores `null` and returns `null`. Trying to update the value from `null` to the empty string in Oracle results in no-change. I expanded the unit tests so that they demonstrate this difference.

This PR is an alternative to #39554 - which changes the interface to set/get config values so that `null` is not an available setting, and any current uses of `null` become the empty string.

"someone"  (tm) needs to decide which PR to approve.

## Related Issue
- Fixes #39553 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/39554
+++ b/changelog/unreleased/39554
@@ -1,0 +1,6 @@
+Bugfix: An app config value of null could not be updated
+
+An app config value that is set to null can now be updated to a not-null value,
+for example, with the occ config:app:set command.
+
+https://github.com/owncloud/core/pull/39554
--- a/changelog/unreleased/39574
+++ b/changelog/unreleased/39574
@@ -3,4 +3,5 @@ Bugfix: An app config value of null could not be updated
 An app config value that is set to null can now be updated to a not-null value,
 for example, with the occ config:app:set command.
 
-https://github.com/owncloud/core/pull/39554
+https://github.com/owncloud/core/pull/39574
+https://github.com/owncloud/core/issues/39553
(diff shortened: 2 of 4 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>apiOnSqlite-sqlite-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/33953/178/1

mmattel:
Is this docs relevant?

phil-davis:
> Is this docs relevant?

It fixes an existing bug - when an app config value was set to `null` (usually accidentally) then it could not be updated with the `occ config:app:set` command. So for that, there is no need to change documentation - the system will now work as it should.

Investigation here has uncovered the fact that an app config value cannot be set to the empty string on Oracle (Oracle turns that into `null`) I have no idea if that should be documented! I can't think of any actual app where a config value needs to differentiate between `null` and "".

phil-davis:
The solution in #39554 was chosen and merged.

Issue #39502, opened 2021-11-17
additional occ user commands necessary


mmattel:
Post merging https://github.com/owncloud/core/pull/39136 (feat: new command to move a user's home folder - occ user:move-home)

* A users home can now be moved with the `occ user:move` command
* You can report for a particular user which home he has:
```
sudo -uwww-data ./occ user:list hugo -a home
  - hugo: /var/www/owncloud/data/hugo
```
* There are commands missing:
1.) list which homes are present like:
`user:homes` --> `/var/www/owncloud/data/` (only unique entries)
2.) list which users belong to a particular home like:
`user:homes -u (or --users)` `/var/www/owncloud/data/` --> `<line by line list of users>`

**Reason**
After moving a users home, you have no knowledge which directories you have to maintain and which users have their home in an particular directory. From an admin pov a nightmare, except you take a physically look into directories and do a search for - if you know the directory...

IMHO both commands are not a big deal as the data is already present and just needs to be queried only and printed.

@pmaier1 as discussed

pmaier1:
@JanAckermann @JammingBen Let's check whether this is a quick thing, as time permits. I think it would add value to the user:move command as it makes it traceable where user homes have been moved.

JammingBen:
If we really want to have two different kinds of output in one command (home paths <> users), I'd prefer to force the user to always specify. Otherwise it seems confusing to me. @pmaier1 What do you think?

* `user:homes -l (--list)` List all user home directories
* `user:homes -u (--users) path` List all users that have their home within a given directory

mmattel:
Q: how do you know the different locations you have (user independent) without a query?

JammingBen:
The backend basically needs to look up all users including their home directories and then generate a list with unique items. Is that answering your question? I'm not sure what you mean by "query" in this context.

mmattel:
After having an audio call with @JammingBen, things are clarified.

JammingBen:
Implemented via https://github.com/owncloud/core/pull/39579

jnweiger:
Confirmed fixed in 10.9.0 RC1


For me, who *wants* to see the two different outputs together, this is my favorite invocation:

```
$ occ user:home:list-dirs | sed -e 's/ - //' | xargs -t -n 1 occ user:home:list-users
occ user:home:list-users /mnt/data/files 
  - admin
  - user02
occ user:home:list-users /tmp 
  - user01
```

mmattel:
[...]

PR #39579, opened 2021-12-09
Add user:home:list-dirs and user:home:list-users commands

JammingBen:
## Description
Implement two new users commands:

* `occ user:home:list-dirs` List all homes which are currently used by users
* `occ user:home:list-users <path>` List all users who have their home in a given path

## Related Issue

- Fixes https://github.com/owncloud/core/issues/39502

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/4492 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

Issue #39278, opened 2021-09-23
User:Report does not show the correct amount if s3 is primary storage


JanAckermann:
### Steps to reproduce
1. Create an ownCloud 10.8 instance with s3 as primary storage
2. Login and upload some files
3. Run `occ user:report`

### Expected behaviour
User directory count should be 1

### Actual behaviour
User directory count should is 0

JanAckermann:
FYI @phil-davis @jvillafanez

DeepDiver1975:
🤖 Automated issue triage by Claude Code review agent.

**Confirmed bug.** The root cause is in the `user:report` command's directory-counting logic, which relies on a local-filesystem call that does not work with S3 primary object storage.

In `core/Command/User/Report.php`, `countUserDirectories()` counts a user directory only when `\is_dir()` returns true for the user's home path:

```php
private function countUserDirectories() {
    $userDirCount = 0;
    $this->userManager->callForSeenUsers(function (IUser $user) use (&$userDirCount) {
        $homePath = $user->getHome();
        if (\is_dir($homePath)) {   // <-- only true for a real local directory
            $userDirCount++;
        }
    });
    return $userDirCount;
}
```

When S3 is the primary object store, the user's "home" is not a real directory on the local filesystem — there is no on-disk path for `\is_dir()` to stat — so `\is_dir($homePath)` returns `false` for every user. As a result the **"user directories"** row is reported as `0` instead of the actual count, which exactly matches the reported behaviour (expected `1`, actual `0`).

This was already partially acknowledged in commit `b80e33ef` ("Add warnings for commands that are incompatible with object storage", 2022-02-08), which added the runtime notice:

> We detected that the instance is running on a S3 primary object storage, user directories count might not be accurate

That commit only added the warning — it did **not** fix the counting itself, so the inaccurate `0` is still produced on current master.

**Fix direction:** for object-storage setups, the count must not depend on `\is_dir()`. Options include using the storage/filesystem abstraction to check for the user's home (e.g. via the user's mount/`Filesystem` view `is_dir`/`file_exists` rather than the raw PHP `\is_dir()` on a path that has no local representation), or counting seen users with an initialized home through the object store API. The local-`\is_dir()` path can be kept for non-object-storage backends.

(No fix PR is being opened from here — `owncloud/core` is excluded from automated fix fleets; this is triage analysis only.)

dj4oC:
## Implementation prompt

[...]

PR #39583, opened 2021-12-13
[10.9.0] occ user:home:list-users --all command

phil-davis:
## Description
Docs PR https://github.com/owncloud/docs/pull/4499 has a command that puts together the `occ user:home:list-dirs` and `occ user:home:list-users` commands to generate a list of users for each different home path in the system.

```
occ user:home:list-dirs | \
  sed -e 's/  - //' | tr '\n' '\0' | xargs -0 -t -n 1 \
  occ user:home:list-users
occ user:home:list-users /mnt/newhome_1 
  - lisa
occ user:home:list-users /var/www/owncloud/data 
  - admin
  - user01
```

That looks like "magic" and actually we can quite easily do that in the existing `occ user:home:list-users` command.

Instead of only being able to match a single path, this PR enhances the command so that it has a `--all` option. If `--all` is specified, then the command lists every home path, with the users that are in each home path.

An alternative would be to add an option to `occ user:home:list-dirs` to list all the users in each of the home paths.

## Related Issue

#39502 

## How Has This Been Tested?
CI unit tests and local run of occ command:
```
$ php occ user:home:list-users --all
  - /home/phil/git/owncloud/core/data:
    - admin
    - uu1
    - uu2
    - uu3
    - uu4
    - uu5
``` 

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/10.9.0_2021-12-09/39579
+++ b/changelog/10.9.0_2021-12-09/39579
@@ -4,6 +4,8 @@ Added two new users commands:
 
 * `occ user:home:list-dirs` List all homes which are currently used by users
 * `occ user:home:list-users <path>` List all users who have their home in a given path
+* `occ user:home:list-users --all` List all users for every home path
 
 https://github.com/owncloud/core/pull/39579
+https://github.com/owncloud/core/pull/39583
 https://github.com/owncloud/core/issues/39502
--- a/core/Command/User/HomeListUsers.php
+++ b/core/Command/User/HomeListUsers.php
@@ -23,20 +23,29 @@ namespace OC\Core\Command\User;
 
 use OC\Core\Command\Base;
 use OCP\IDBConnection;
+use OCP\IUserManager;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class HomeListUsers extends Base {
 	/** @var IDBConnection */
 	protected $connection;
 
+	/** @var \OCP\IUserManager */
+	protected $userManager;
+
 	/**
 	 * @param IDBConnection $connection
 	 */
-	public function __construct(IDBConnection $connection) {
+	public function __construct(
+		IDBConnection $connection,
+		IUserManager $userManager
+	) {
 		parent::__construct();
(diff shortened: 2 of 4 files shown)


pmaier1:
:+1: thanks!

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIUpload-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/34001/150/1

jvillafanez:
Did you tested with files_primary_S3? I think we'll have to check how it works in case we have to add some additional information. For example, I'm not sure how it will work with the "path" option.

phil-davis:
> Did you test with files_primary_S3? I think we'll have to check how it works in case we have to add some additional information. For example, I'm not sure how it will work with the "path" option.

That will be the same as was already merged in `10.9.0-RC1` PR #39579 - @JammingBen did you check what happens to this command when used on a system with S3 storage?

JammingBen:
No I didn't, let me check. But I suspect that it won't work :( Kinda related maybe: https://github.com/owncloud/core/issues/39278

mmattel:
The 'user:home' commands are related to 'user:move' which is limited to POSIX storage (which is documented). In the open docs PR describing the 'user:home' commands, there is currently no mentioning with regards to POSIX - which I have to add to make it clear. Nontheless it would be interesting to see results when used with files_primary_s3...

mmattel:
> I assume there are a lot of commands where this should be added though, not just these new ones.

Walking a path starts with the first step 😃

JammingBen (approved):
LGTM.

I've created https://github.com/owncloud/core/issues/39590 to keep track of the warning for incompatible commands.

mmattel:
Linking to the docs issue fixing the occ commands:
https://github.com/owncloud/docs/issues/4492 (New occ commands extending user:move these are user:home:list-dir and user:home:list-users <path>)

mmattel (approved):
LGTM, note that I just added the option to the docs PR, only waiting that this PR gets merged.

jnweiger:
Confirmed fixed in 10.9.0 RC2 (posix filesystem case)

(I did not expect that my quirky comment would make it into a feature. Wow!)

PR #39600, opened 2021-12-15
Add default app setting on a user basis

JammingBen:
## Description
This change allows users to to have their own default app defined. The setting is stored in the user preference table and uses its existing API. It will overwrite the system-wide `defaultapp` setting if set.

## Related Issue

- Related https://github.com/owncloud/enterprise/issues/4694

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs/issues/4508

### DIFF
--- a/changelog/unreleased/39600
+++ b/changelog/unreleased/39600
@@ -1,0 +1,3 @@
+Enhancement: Add default app setting on a user basis
+
+https://github.com/owncloud/core/pull/39600
--- a/lib/private/legacy/util.php
+++ b/lib/private/legacy/util.php
@@ -1175,7 +1175,17 @@ class OC_Util {
 				$location = $urlGenerator->getAbsoluteURL($defaultPage);
 			} else {
 				$appId = 'files';
-				$defaultApps = \explode(',', \OC::$server->getConfig()->getSystemValue('defaultapp', 'files'));
+				$uid = \OC_User::getUser();
+				$config = \OC::$server->getConfig();
+				$defaultApps = \explode(',', $config->getSystemValue('defaultapp', 'files'));
+
+				if ($uid) {
+					$userDefaultApp = $config->getUserValue($uid, 'core', 'defaultapp', null);
+					if ($userDefaultApp) {
+						\array_unshift($defaultApps, $userDefaultApp);
+					}
+				}
+
 				// find the first app that is enabled for the current user
 				foreach ($defaultApps as $defaultApp) {
 					$defaultApp = OC_App::cleanAppId(\strip_tags($defaultApp));
--- a/tests/lib/UtilTest.php
+++ b/tests/lib/UtilTest.php
@@ -362,6 +362,25 @@ class UtilTest extends \Test\TestCase {
 		$this->assertSame('http://localhost'.\OC::$WEBROOT.'/apps/files/', OC_Util::getDefaultPageUrl());
 	}
 
+	public function testGetDefaultPageUrlWithUserConfig() {
+		$uid = $this->getUniqueID();
+		$userDefaultApp = 'user_default_app';
+		\OC_User::setUserId($uid);
+		\OC::$server->getConfig()->setUserValue($uid, 'core', 'defaultapp', $userDefaultApp);
+
+		$appManager = $this->createMock(IAppManager::class);
+		$appManager->expects($this->any())
+			->method('isEnabledForUser')
+			->willReturn(true);
+		Dummy_OC_Util::$appManager = $appManager;
+
+		\putenv('front_controller_active=true');
+		$_REQUEST['redirect_url'] = 'myRedirectUrl.com@foo.com:a';
+		$this->assertSame('http://localhost'.\OC::$WEBROOT.'/apps/' . $userDefaultApp . '/', Dummy_OC_Util::getDefaultPageUrl());
+		\OC::$server->getConfig()->deleteUserValue($uid, 'core', 'defaultapp');
+		\OC_User::setUserId(null);
+	}
+
 	/**
 	 * Test needUpgrade() when the core version is increased
 	 */


phil-davis:
Tested by:
- enable activity and notes apps
- set `'defaultapp' => 'activity'` in config.php (system-wide default app)
- `php occ user:setting brian core defaultapp --value=notes
- login as brian
- the Notes app is displayed - good
- login as alice
- the Activity app is displayed - good

- remove `defaultapp` from config.php
- login as brian
- the Notes app is displayed - good
- login as alice
- the Files app is displayed - good

Works

phil-davis (approved):

jvillafanez:
Taking into account that it won't be in 10.9, I think the users should be able to set the app from the web ui, in the personal settings.
Unless I'm missing something, it seems lacking in terms of UX. I mean, it's a per-user setting but the user can't change it in any way.

JanAckermann:
@jvillafanez 
No, this is not the plan right now. There will be a setting in ownCloud/web where you can toggle the new design off and on.

mmattel:
Reading this I think it is not the right approach (while implementation is relatively easy). Having a setting on a per user basis makes management quite complicated. Imho it would be better to do this at a group level. 
This makes management from an admin pov much easier. Just imagine to find users who have a particular setting set or move users from one setting to another. Because it is now not in 10.9, it offers a re-thinking about the best approach.

Finally, it is docs relevant...

As a rule of thumb, when touching occ, it **IS** docs relevant, no matter what it is - file a docs issue and we will take a look on.

JanAckermann:
@mmattel 
Please decouple your mind from 
Business logic, API and UI.
Don't be to harsh in your judgements.


Anyways this is not a admin setting, it is a user setting, done by the user (UI changes incoming).

It might be docs relevant, but for this setting, we usually don't want anyone to use the occ command.
Therefore a API and UI implementation is planned.

mmattel:
Sorry if my words sounded harsh, not intended to be so... 😢 
I just wanted to highlight, if based on an occ command, a user setting approach will be more challenging for an admin to administrate then a group based approach. If it is planned to have that on a user-UI base, it is a different story.
> ...for this setting, we usually don't want anyone to use the occ command...

This is a very dangerous path

JanAckermann:
@mmattel Got you, but this is not a setting the admin want to take care of! 

Me as a individual want to decide if I want to use the new design aka owncloud/web.

>This is a very dangerous path

See it from the correct perspective, it would be possible for the admin, but why should they do that.

phil-davis:
> This is a very dangerous path if you create/extend an occ command and then have in mind not using it.

@mmattel the `occ user:setting` command already exists. That just provides an existing way for an admin to set various user-specific preferences if they want/need to do that at any time.

JammingBen:
It's pretty much said already, to sum it up:

This PR just lays the groundwork for having this feature. It doesn't include UI (yet) as we currently plan to use this for the web app only. The UI for this will then be in the web app directly.

It's not intended for the admin to set default apps for other users. But because the setting sits in the user preferences, existing APIs such as `occ user:setting` can be used modify it. The doc issue I created might be misleading though, I'll adjust it 👍

cdamken:
we need to add the option in the UI, it will be impossible for the admins to set and remove the option for 300K users 

btw, which are all other opinions?

> `occ user:setting --value web admin core defaultapp`

only web or remove the setting to go back?

JanAckermann:
Please read the description of the related issue!
Again: this is s USER setting!

cdamken:
I read and that's why I asked the second part about the options

The first part means that I can't use this feature so far ;)

JanAckermann:
There will be a UI implementation, for the USER to switch between the new and classic design (UI), therefore we need this code.

 >I read and that's why I asked the second part about the options
It's not planned that you set options on your own.

>The first part means that I can't use this feature so far ;)
No, you shouldn't

mmattel:
@JanAckermann we have all green - merging? (or are there more changes necessary?)

phil-davis:
I rebased to get up-to-date CI. "someone" can decide what is to happen with this.

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToRoot2-git-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/36076/108

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFederationToShares1-10.8.0-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/36076/113

jvillafanez (approved):

phil-davis:
The API for a user to set their email address (for example) is:
```
curl -X PUT http://aaaa:aaaa@172.17.0.1:8080/ocs/v1.php/cloud/users/aaaa \
    -d key="email" \
    -d value="aaaa@example.com"
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

But if I try:
```
curl -X PUT http://aaaa:aaaa@172.17.0.1:8080/ocs/v1.php/cloud/users/aaaa \
>     -d key="defaultapp" \
>     -d value="activity"
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>997</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

After merging this PR, how can a client app call an API endpoint to set `defaultapp` ?

Without an API  end-point that works, there won't be a way for clients (web, whatever other client) to actual implement this option.

phil-davis:
> Without an API end-point that works, there won't be a way for clients (web, whatever other client) to actual implement this option.

To me, it seems pointless to add this in a release, if there is no way to apply the setting other than with a "back end" occ command. I am hoping that someone can say how to use some API endpoint to set the default app for a user. But I don't know who "someone" is ;)

But it does no harm, so could be merged and then follow-up if an API endpoint needs to be adjusted.

jvillafanez:
> After merging this PR, how can a client app call an API endpoint to set defaultapp ?
>
> Without an API end-point that works, there won't be a way for clients (web, whatever other client) to actual implement this option.

I don't think it's planned for now. The idea was to use it with the new web interface so the admin can change the interface for specific users manually.
Basically, the admin decides which users will have the new web interface by default so they can test it before making it available to all the users.

We should follow up this ticket at some point, but I'm ok with merging this.

JanAckermann:
>I don't think it's planned for now. The idea was to use it with the new web interface so the admin can change the interface for specific users manually.
Basically, the admin decides which users will have the new web interface by default so they can test it before making it available to all the users.

That's not quite true, we planed to have an UI setting and every user can decide by their selfs

mmattel:
I think this is the key message which is relevant for the new webUI,
therefore this should go into the upcoming release:
`Basically, the admin decides which users will have the new web interface by default so they can test it before making it available to all the users.`
I will prepare a comment in the webui admin documentation.

JanAckermann:
We are mixing up different things here. 

>The idea was to use it with the new web interface so the admin can change the interface for specific users manually.
Basically, the admin decides which users will have the new web interface by default so they can test it before making it available to all the user

I don't know where this info is gathered, but it is basically wrong.

Please also check:
https://github.com/owncloud/web/pull/6173/files

pmaier1:
AFAIR, the idea was to save the state the user had the last time. So each time the user switches from Classic to Web, that would be persisted and vice versa.

PR #39607, opened 2021-12-21
add expiration extension to tus option requests

wkloucek:
## Description
The testsuite does currently not expect the tus expiration extension on option requests, but REVA is returning it already. This PR updates the testsuite, so that we can remove these tests from the expected failures list.

The current behaviour can be seen here: https://drone.cernbox.cern.ch/cs3org/reva/4593/26/7

Tus spec about expiration extension on option requests: https://tus.io/protocols/resumable-upload.html#options

## Related Issue

- Removes some expected failures from https://github.com/owncloud/ocis/issues/1755 (even though they are not really related to the issue)

## Motivation and Context
Slim down the expected failures in REVA / oCIS


phil-davis (commented):
See #39615 for a rebase to master. LGTM

phil-davis:
@wkloucek please review #39615 and that can be merged to master. Then I will include it in commit-id bumps that are in progress.

PR #39644, opened 2022-01-05
Remove BOM from text preview content

### LINKED ISSUE #39645: Thumb nail generator fails with text files with BOM
Seen with 10.9.1RC1 (but probably unrelated to the specific release)

* Create a text file via web interface.
* edit the text file with collabora office
* confirm the edits in text editor. the bom is visible as a red dot. OK
image

* the thumbnail shows only empty squares:
image

* The file is a plain ascii text file except for th efist two bytes, which are the well defined form of the BOM, as e.g. defined in 
image

AlexAndBear:
## Description
Bugfix: Text file previews have faulty content if a byte order mark is present

While having a valid byte order mark in a text file, the preview was corrupt.
With this change, the BOM will be removed and the preview will be rendered
correctly.

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Preview/TXT.php
+++ b/lib/private/Preview/TXT.php
@@ -54,6 +54,9 @@ class TXT implements IProvider2 {
 		$content = \stream_get_contents($stream, 2048);
 		\fclose($stream);
 
+		// Remove byte order mark
+		$content = \str_replace("\xEF\xBB\xBF", '', $content);
+
 		//don't create previews of empty text files
 		if (\trim($content) === '') {
 			return false;
--- a/changelog/unreleased/39644
+++ b/changelog/unreleased/39644
@@ -1,0 +1,7 @@
+Bugfix: Text file previews have faulty content if a byte order mark is present
+
+While having a valid byte order mark in a text file, the preview was corrupt.
+With this change, the BOM will be removed and the preview will be rendered
+correctly.
+
+https://github.com/owncloud/core/pull/39644
--- a/changelog/unreleased/39644
+++ b/changelog/unreleased/39644
@@ -5,3 +5,4 @@ With this change, the BOM will be removed and the preview will be rendered
 correctly.
 
 https://github.com/owncloud/core/pull/39644
+https://github.com/owncloud/core/issues/39645


lib/private/Preview/TXT.php
```diff
@@ -54,6 +54,9 @@ public function getThumbnail(File $file, $maxX, $maxY, $scalingUp) {
 		$content = \stream_get_contents($stream, 2048);
 		\fclose($stream);
 
+		// Remove byte order mark
+		$content = \str_replace("\xEF\xBB\xBF", '', $content);
```
DeepDiver1975, line 58: BOM is only relevant in the beginning of the file - so no need to do a replace over the full content.

replacing only the first few bytes is enough

DeepDiver1975: https://gist.github.com/chrisguitarguy/6096271

DeepDiver1975 (changes requested):
in addition to the comment: adding a test would be great ;-) THX

lib/private/Preview/TXT.php
JanAckermann: @DeepDiver1975 
I agree in a perfect world you should have only one BOM in a file, but it might be possible that a user created a combined file using the unix cat command and have accidentally two or more BOMs in a file.

Should we ignore those user errors?

jnweiger: The BOM is only a BOM when at the beginning of a file. But U+FEFF may legally appear anywhere inside a text file. And it is still valid text. Unicode.org says "For historical reasons, the character U+FEFF used for the byte order mark is named `zero width no-break space`" -- and requests proper handling (as a zero-width-nbsp, when found elsewhere) for backwards compatibility.

So that is not a user error :see_no_evil: -- that is part of the unicode horrors.

jvillafanez:
There are a couple of issues here:
* The text analyzer detects the BOM as an arabic character (last range in https://github.com/owncloud/core/blob/0b3cbdd5b6675a27d35cd181665e92ac72142ef1/lib/private/Utf8Analyzer.php#L95 seems to be wrong, I need to recheck)
* Since there is one arabic character, the text is considered arabic, not latin. The problem seems to be in https://github.com/owncloud/core/blob/0b3cbdd5b6675a27d35cd181665e92ac72142ef1/lib/private/Preview/TXT.php#L137-L144 , where latin is only used if there isn't any other match. Basically, although the text contains X latin chars, the text is considered as arabic due to the wrongly detected BOM.

I think these are the 2 issues we should fix instead of removing the BOM.

jvillafanez:
Alternative fix in https://github.com/owncloud/core/pull/39669

PR #39652, opened 2022-01-07
[tests-only] [full-ci] Test with mariadb:10.3

phil-davis:
1) test with current master


phil-davis:
https://drone.owncloud.com/owncloud/core/34238/119/10
cliEncryption
```
No encryption module is loaded
+ php occ encryption:select-encryption-type masterkey --yes

  Command "encryption:select-encryption-type" is not defined.  
                                                               
  Did you mean one of these?                                   
      encryption:change-key-storage-root                       
      encryption:decrypt-all                                   
      encryption:disable                                       
      encryption:enable                                        
      encryption:encrypt-all                                   
      encryption:list-modules                                  
      encryption:set-default-module                            
      encryption:show-key-storage-root                         
      encryption:status                                        
```

Looks like some other action needs to happen before selecting master key.

PR #39653, opened 2022-01-07
[tests-only] [full-ci] Test with mariadb:10.4

phil-davis:
1) test with current master


phil-davis:
https://drone.owncloud.com/owncloud/core/34188/130/17
`webUIFiles/scrollMenuIntoView.feature:13`

`element click intercepted: Element <a class="action action-menu permanent" href="#" data-action="menu" data-original-title="" title="">...</a> is not clickable at point (792, 663). Other element would receive the click: <li>...</li>`

Looks like an unrelated fail. I will restart.

PR #39655, opened 2022-01-07
[tests-only] [full-ci] Test with mariadb:10.6

phil-davis:
1) test with current master
2) do not test against core 10.8 - that does not support mariadb 10.6
3) test with 10.9.1RC1 tarball


phil-davis:
good - passes against core master

PR #39656, opened 2022-01-07
[tests-only] [full-ci] Test with mysql:5.5

phil-davis:


phil-davis:
https://doc.owncloud.com/server/10.9/admin_manual/installation/system_requirements.html#server

mysql 5.5 is no longer officially supported.

Known things fail, like emojis in comments (Unicode strings):
https://drone.owncloud.com/owncloud/core/34191/51/13

Comment with actorId = 'Alice' and message = '😀    🤖' not found

Closing - we don't need to analyze all the results here.

PR #39657, opened 2022-01-07
[tests-only] [full-ci] Test with mysql:5.7

phil-davis:
1) test with current master


phil-davis:
great - passes.

phil-davis:
Note: https://doc.owncloud.com/server/10.9/admin_manual/installation/system_requirements.html#server

MySQL 5.7 is no longer in the supported databases list. But it is nice to know that the acceptance test suite still passes.

PR #39658, opened 2022-01-07
[tests-only] [full-ci] test with mysql:8.0

phil-davis:
1) test with current master


phil-davis:
great - passes against current master.

Issue #39662, opened 2022-01-07
Editing a shared file sometimes fails to save the author with the versions


jnweiger:
Seen in 10.9.1RC1 (without enabling encryption)

* user admin creates a text file in his Documents folder and does a few edits. (I used a file called eicar.com.txt -- stressing the virus scanner a bit)
* versions get saved.
* user admin shares the Documents folder with user user01
* We add `'file_storage.save_version_author' => true,` to config.php
* user01 and admin alternatingly edit the file. (admin uses both files_texteditor and a desktop client)
* The activities look like this:
![image](https://user-images.githubusercontent.com/1108546/148603795-3c74ae2d-188c-4b79-bfa5-2f165bef972a.png)

----

* The versions log looks like this:
![image](https://user-images.githubusercontent.com/1108546/148603983-187a0d02-c520-490f-893e-684e738b11eb.png)

At ca 6 min in the past, save_versions_author was set to true.
Thus it is correct that there is no author information on older versions.
But newer versions never mention user01. His edits appear anonymous, while the edits of user admin are correctly attributed.
The corresponding json files are missing:

```
 find /var/www/owncloud/data -name eicar\*
/var/www/owncloud/data/admin/files/Documents/eicar.com.txt
/var/www/owncloud/data/admin/files_versions/Documents (2)/eicar.com.txt.current.json
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641586072
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641586025
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641586205.json
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641586205
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641585092
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641585734
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641586072.json
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641585349
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641585736
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.v1641586132.json
/var/www/owncloud/data/admin/files_versions/Documents/eicar.com.txt.current.json
[...]

PR #39639, opened 2022-01-04
[tests-only] Tests 10.9.1

phil-davis:
## Description
This PR has some tests that were recently added to core master, and provide some extra test coverage for things that should work in 10.9.1. It will demonstrate that the tests do pass for 10.9.1.

We don't have to merge this to `release-10.9.1` but we could if we like.

## Related Issue
https://github.com/owncloud/enterprise/issues/4933

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [x] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39670, opened 2022-01-10
Fix getting the file owner for share recipients

JammingBen:
## Description

Fixed a bug where a wrong file owner was retrieved when saving version authors. This scenario happened for share recipients if they had a file with the same name as the shared one.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/39662

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_versions/lib/MetaStorage.php
+++ b/apps/files_versions/lib/MetaStorage.php
@@ -103,12 +103,14 @@ class MetaStorage {
 			return false;
 		}
 
-		$userPath = "/files_versions$currentFileName" . MetaStorage::CURRENT_FILE_EXT;
-		$owner = Filesystem::getOwner($currentFileName);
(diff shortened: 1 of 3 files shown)


IljaN (approved):
LGTM :+1:

phil-davis (approved):
See PR #39675 - that has 2 new scenarios that demonstrate this issue, and they pass with this code.

I can't just cherry-pick those in here, because they need various test code from master that is not in the `release-10.9.1` branch. If we want, we could merge #39639 to `release-10.9.1` to get all that test code into the release branch. But it might be best not to do that - we might accidentally cause some merge problem back to master...

When we merge back things from `release-10.9.1` to master (however we do that in a controlled way), we can also apply the test scenarios to master.

PR #39262, opened 2021-09-21
 Redesign group list in the user settings view

AlexAndBear:
## Description
Enhancement: Redesign group list in the user settings view

Before this PR, the group list settings were confusing as the user needs to
click on a group first before the trash bin icon appears to delete a group.
As well the user count was nondescript.
With this PR, a redesign takes care of those issues and also fixes the problem
that while deleting a group in the group list,
it was still available in the user group list.
Furthermore, after deleting a group, the user will be redirected to the
'Everyone' group summary view.
Also fixes an issue where the 'active'-class on a newly created group
was duplicated.

## Related Issue

- Fixes https://github.com/owncloud/core/issues/36273
- Fixes https://github.com/owncloud/core/issues/29057

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

### Before
image

### After
image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39721, opened 2022-01-26
Remove the user count if the number isn't positive.

jvillafanez:
This restores the previous behavior. It also remove the "everyone"'s
counter if it doesn't have any data, which could happen with LDAP, in
order to match what happens with the LDAP groups

## Description
Restore the previous behavior where the user counter didn't appear if the counter was 0 or negative (which happens with LDAP)

Restored from https://github.com/owncloud/core/pull/39262

In addition, the "everyone" counter is removed if it doesn't have any text to prevent showing the counter without any data. There is a padding that makes a blue background to show up, as show in the related issue.

## Related Issue

https://github.com/owncloud/enterprise/issues/4957

## How Has This Been Tested?

Manually checked the user list with an LDAP connected

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/settings/js/users/groups.js
+++ b/settings/js/users/groups.js
@@ -344,6 +344,10 @@ $(document).ready( function () {
 		GroupList.getEveryoneCount();
 	}
 
+	if ($('#everyonecount').text().trim() === '') {
+		// remove the count to match the rest of the groups
+		$('#everyonecount').remove();
+	}
 	// Display or hide of Create Group List Element
 	$('#newgroup-form').hide();
 	$('#newgroup-init').on('click', function (e) {
--- a/settings/js/users/groups.js
+++ b/settings/js/users/groups.js
@@ -45,12 +45,13 @@ var GroupList;
 		},
 
 		setUserCount: function (groupLiElement, usercount) {
+			var $groupLiElement = $(groupLiElement);
 			if (this.$sortGroupBy !== 1) {
 				// If we don't sort by group count we don't display them either
+				$groupLiElement.find('span.usercount.tag').remove();
 				return;
 			}
 
-			var $groupLiElement = $(groupLiElement);
 			if (usercount === undefined || usercount === 0 || usercount < 0) {
 				usercount = 0;
 				$groupLiElement.data('usercount', 0);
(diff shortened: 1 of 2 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>webUIManageUsersGrps-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/34431/133/1

phil-davis:
https://drone.owncloud.com/owncloud/core/34431/133/17
`Then the user count of group "grp1" should display 0 users on the webUI`

`Page\UsersPage::getUserCountOfGroup xpath //li[@data-gid='grp1']//span[@class='usercount tag'] could not find user count for group grp1 (SensioLabs\Behat\PageObjectExtension\PageObject\Exception\ElementNotFoundException)`

Those tests will need to be refactored. Maybe a step like this is needed:
`Then the user count of group "grp1" should not be displayed on the webUI`

Do you want me to do that, or you? @jvillafanez

jvillafanez:
@phil-davis I'll delegate that on you because you'll do it faster for sure. You can use this branch if it's easier.

phil-davis:
When the user management page is first displayed, the counts are correct. Groups with no members do not have a count at all, groups with members display their member count - good.

But the display of member does not switch on/off in real-time:

Case 1:
- have a group with 1 member only
- remove that member from the group
- the member count is shown as "0" (it should disappear)

Case 2:
- have a group with no members
- add a member to the group
- the member count is not shown (it should show "1")

@jvillafanez I am going home for dinner now, so feel free to fix that real-time behavior and push. I will adjust tests later this evening or tomorrow morning.

jvillafanez:
Case 1 seems ok to me, but case 2 might be bad.

I think I'll adjust the behavior a bit. I'm quite convinced that, previously, the user count only show if the group list is ordered by user count (this is set in the server before the page is loaded). If the list isn't ordered or is ordered by group name, the user count doesn't appear.
When LDAP is involved, I think the list is sorted by group name.

jvillafanez:
This should be running more stable now I think.

As said, the user count will be shown if the group list is sorted by the user count. This seems to be the default behavior. The user count won't be removed if it reaches 0.

If the group list is sorted differently, which happens if LDAP is involved, the user count won't be shown in any case.

Linking with https://github.com/owncloud/core/issues/39722 since it might be difficult to verify with that bug around. To be fixed separately.

jvillafanez:
I'll need some opinions on the flicker effect that happens now having an LDAP server connected.

From my point of view, it can be fixed, but we need a change of approach which requires additional changes, so not sure if we want to invest time on this right now.

Right now, the server prefills the data and the javascript code removes it if needed, so, until that javascript code runs, the user count is visible.
The alternative approach is that the server doesn't prefill the data, and it's the javascript code the one filling it on demand.

jvillafanez:
Fix for https://github.com/owncloud/core/issues/39722 also included because is was easy.

AlexAndBear (approved):

PR #39746, opened 2022-02-03
[tests-only] [full-ci] Refactor and fix dav-path-spaces test code

phil-davis:
## Description
When running on reva, the personal space id has to be looked up by more tricky ways. That code is coming. But at the moment some test scenarios are failing for new and old dav path because code in TusContext calls `getPersonalSpaceIdForUser` even though the scenario is not actually using the "spaces dav path".

CI has been green in oCIS because the call to ऽgetPersonalSpaceIdForUser` does work on oCIS. The call is wasted, the space id is not actually used, but at least the code runs.

On `cs3org/reva` the space id is not found, because the `/graph/v1.0/me/drives` endpoint does not exist. And those scenarios fail when they should not fail.

https://drone.cernbox.cern.ch/cs3org/reva/5429/15/6
`Notice: Trying to get property 'value' of non-object in /drone/src/tmp/testrunner/tests/TestHelpers/WebDavHelper.php line 401`

This PR does:
1) refactor test code so that the "davPathVersion" magic numbers 1,2,3 are held in constants `DAV_VERSION_OLD` `DAV_VERSION_NEW` `DAV_VERSION_SPACES`. That should make code easier to read, rather than having 1,2,3 mentioned in multiple places.

2) modify `FeatureContext` `getPersonalSpaceIdForUser` so that it only actually tries to get the space id if the scenario is using `DAV_VERSION_SPACES`. If not, then it returns `null`. That will allow it to be called more flexibly.

## How Has This Been Tested?
CI

oCIS PR https://github.com/owncloud/ocis/pull/3105 passes against this branch, so the changes do not break that.

cs3org/reva PR https://github.com/cs3org/reva/pull/2497 passes against this branch, with just sensible line number changes in the expected-failures files.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/WebDavHelper.php
+++ b/tests/TestHelpers/WebDavHelper.php
@@ -37,6 +37,9 @@ use DateTime;
  *
  */
 class WebDavHelper {
+	public const DAV_VERSION_OLD = 1;
+	public const DAV_VERSION_NEW = 2;
+	public const DAV_VERSION_SPACES = 3;
 
 	/**
 	 * @var array of users with their different spaces ids
@@ -124,7 +127,7 @@ class WebDavHelper {
 		?string $xRequestId = '',
 		?string $folderDepth = '0',
 		?string $type = "files",
-		?int $davPathVersionToUse = 2
+		?int $davPathVersionToUse = self::DAV_VERSION_NEW
 	):ResponseInterface {
 		$propertyBody = "";
 		$extraNamespaces = "";
@@ -201,7 +204,7 @@ class WebDavHelper {
 		?string $propertyValue,
 		?string $xRequestId = '',
 		?string $namespaceString = "oc='http://owncloud.org/ns'",
-		?int $davPathVersionToUse = 2,
+		?int $davPathVersionToUse = self::DAV_VERSION_NEW,
 		?string $type="files"
 	):ResponseInterface {
 		$matches = [];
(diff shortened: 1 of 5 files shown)


tests/TestHelpers/WebDavHelper.php
```diff
@@ -558,13 +561,18 @@ public static function getDavPath(
 		if ($type === "customgroups") {
 			return "remote.php/dav/";
 		}
-		if ($davPathVersionToUse === 3) {
+		if ($davPathVersionToUse === self::DAV_VERSION_SPACES) {
+			if (($spaceId === null) || (\strlen($spaceId) === 0)) {
+				throw new InvalidArgumentException(
+					__METHOD__ . " A spaceId must be passed when using DAV path version 3 (spaces)"
+				);
+			}
```
kiranparajuli589, line 569: :100:

tests/acceptance/features/bootstrap/ChecksumContext.php
```diff
@@ -163,11 +163,6 @@ public function userRequestsTheChecksumOfViaPropfind(string $user, string $path)
 				<oc:checksums />
 			  </d:prop>
 			</d:propfind>';
-		$url = $this->featureContext->getBaseUrl() . '/' .
-			WebDavHelper::getDavPath(
-				$user,
-				$this->featureContext->getDavPathVersion()
-			) . $path;
```
kiranparajuli589, line 170: `$url` is never used afterward. :+1:

kiranparajuli589 (approved):
looks good

tests/TestHelpers/WebDavHelper.php
phil-davis: This helped me find bugs!

PR #39769, opened 2022-02-08
[tests-only] Run tests with release-10.9.1

phil-davis:
This is a test PR to see if the test code plus oC10 server code in release-10.9.1 do actually work nicely together.


phil-davis:
Works - that's a good start.

Issue #39664, opened 2022-01-08
Restoring an encrypted file as a share recipient breaks its versions


jnweiger:
Seen with 10.9.1RC1 with encryption 1.5.1 (initially disabled)

* admin creates a file shared.txt in his Documents folder
* admin shares the folder with user Alice
* both edit the file several times
* Alice deletes the file. It appears in the trashcan of both users, with all versions info each.

enable masterkey encryption. logout/login all users.

* Alice restores the file from the trashcan.
* Alice inspects the versions: The topmost listed version has a broken thumbnail and cannot be downloaded.
* Alice restores this version. The file is now permanently lost. no other versions can now be restored.

![image](https://user-images.githubusercontent.com/1108546/148629871-87dcb7d5-efc9-4710-90c3-a01eee6f678c.png)

The share owner admin also sees this error, when he tries to edit the file.
Removing the share and re-sharing does not help.

JanAckermann:
Reproduced - error is existing

**But this is also existing on 10.8 and 10.9, not a regression**

JammingBen:
Note that this error is not occurring due to enabling encryption in between the steps. The issue is also happening when enabling encryption at the beginning.

jvillafanez:
It seems the file itself is fine and "could" be downloaded. The problem is the signature verification fails somehow. Bypassing the signature verification (not recommended) works. Running the `occ encryption:fix-encrypted-version` fixes the issue.

I'll try to figure out why the signature verification fails.

jvillafanez:
Unless someone has a better idea, I think we should create some tooling to check the encrypted files:
* Comparing decrypted content of any file in ownCloud, including trashbin, versions, external storages, etc.
* Comparing filekeys of any file
* Signature verification (without modification)

Right now, it seems that something goes wrong when we try to restore the file from the trashbin, but we can't ensure that the signature broke when they landed in the trashbin, so the problem could happen before restoring.

mrow4a:
maybe stupid but maybe when restoring a file code from `occ encryption:fix-encrypted-version` should be executed?

jvillafanez:
For me it isn't clear that the signature gets broken when the file is restored. It might happen when the file is deleted and moved to the trashbin, so then you restore the file "correctly" with the broken signature. The problem is that I'm not sure how we can verify when the signature get broken.

> maybe stupid, but maybe when restoring a file, code from occ encryption:fix-encrypted-version should be executed ?

[...]

PR #39822, opened 2022-02-22
Fix issue restoring versions from the trashbin after sharing

jvillafanez:
A folder contains a file with multiple versions. The folder is shared
with another user. That user deletes the shared file (with the
versions). Then, that user restores the file. At that point, the
versions were broken due to a bad signature.

## Description
According to docs, versions are copied as such, so the signature in the file doesn't change. Since the signature doesn't change, the "encrypted" column shouldn't change neither.
Even if we're moving versions, we won't increase the "encrypted" value for the version in order to not disrupt the signature in the file.

## Related Issue
https://github.com/owncloud/core/issues/39664

## Motivation and Context
Versions of the file were broken after following the steps described in the issue

## How Has This Been Tested?
Manually tested following the steps

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -722,10 +722,13 @@ class Encryption extends Wrapper {
 
 		// in case of a rename we need to manipulate the source cache because
 		// this information will be kept for the new target
-		if ($isRename) {
+		if ($isRename && !($this->isVersion($sourceInternalPath) || $this->isVersion($targetInternalPath))) {
 			/*
 			 * Rename is a process of creating a new file. Here we try to use the
 			 * incremented version of source file, for the destination file.
+			 * Note that if we're renaming or moving a version, the version is copied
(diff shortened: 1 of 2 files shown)


lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -722,7 +722,7 @@ private function updateEncryptedVersion(Storage $sourceStorage, $sourceInternalP
 
 		// in case of a rename we need to manipulate the source cache because
 		// this information will be kept for the new target
-		if ($isRename) {
+		if ($isRename && !($this->isVersion($sourceInternalPath) || $this->isVersion($targetInternalPath))) {
 			/*
```
JanAckermann, line 726: Please add a comment here

AlexAndBear (changes requested):

AlexAndBear (approved):

PR #39841, opened 2022-03-01
Avoid sending passwords in the API

jvillafanez:
## Description
Passwords could be seen in several requests made by ownCloud

## Related Issue
https://github.com/owncloud/enterprise/issues/5036

## How Has This Been Tested?
Manually checked with some external storages and some operations on them.
* Create a new storage
* Update a storage config without changing the password
* Reload the web page (password should be replaced)
* Check with additional WND authentication methods
  * WND Collaborative
  * User-entered authentication
* Personal mounts

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/lib/Controller/GlobalStoragesController.php
+++ b/apps/files_external/lib/Controller/GlobalStoragesController.php
@@ -118,6 +118,9 @@ class GlobalStoragesController extends StoragesController {
 
 		$this->updateStorageStatus($newStorage);
 
+		// replace the password to prevent leaking it
+		$this->replacePasswords($newStorage);
+
 		return new DataResponse(
 			$newStorage,
 			Http::STATUS_CREATED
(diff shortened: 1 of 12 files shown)


C0rby (approved):
Looks good to me. But we should also tackle the other things like private key or OAuth2 secret.

phil-davis:
There is no changelog.

Tests? There could be unit tests to check that the special password is produced in output, and that it is correctly recognised when it arrives as input.

And/or I can get someone to add a couple of API acceptance tests that check GET and POST of this stuff.

C0rby (changes requested):
Yeah, @phil-davis is right. At least a unit test and a changelog would be good.

C0rby (approved):

AlexAndBear (changes requested):
LGTM so far.
But we need to add tests, that the actual password won't be changed if you provided the REDACTED_PASSWORD via a patch.

If this would ever happen, it might be a crucial bug, so let's better be safe here.

AlexAndBear (approved):

phil-davis (approved):

PR #39844, opened 2022-03-01
emit correct legacy hook

butonic:
shouldn't this be update instead of add?

just asking...

### DIFF
--- a/changelog/unreleased/39844
+++ b/changelog/unreleased/39844
@@ -1,0 +1,6 @@
+Bugfix: Trigger the right event when the filecache is updated
+
+When the filecache was being updated, an "addToCache" event was triggered.
+The event has been changed to an "updateCache" event
+
+https://github.com/owncloud/core/pull/39844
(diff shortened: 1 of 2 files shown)


phil-davis:
Someone needs to review this, and if the change is correct:
- write something about what possible bug(s) this might fix
- write a changelog

jvillafanez:
Rebased and changelog added.

I don't think it will fix anything. The code is used in https://github.com/owncloud/core/blob/817f54f53d64249166feaa0eb9231c50ca1ed7e0/lib/private/Files/Utils/Scanner.php#L153-L161 and https://github.com/owncloud/core/blob/817f54f53d64249166feaa0eb9231c50ca1ed7e0/lib/private/Files/Utils/Scanner.php#L230-L238 , and the code path is the same with and without the PR.

I'm not sure if there is any 3rd-party app listening to those events, but it should be more consistent now.

jvillafanez (approved):

phil-davis (approved):

phil-davis:
Assigned myself so that I notice this and merge it when CI passes.

PR #28165, opened 2017-06-20
[WIP] no trip to the db is better than a fast trip to the db

butonic:
## Description
No trip to the db is better than a fist trip to the db.

## How to reproduce
1. install diagnostics app
2. create 3 users, A and B
3. make user A share 10 files (probably with different mimetypes) and folders with user B
4. log in as user b and go to 'shared with me'

## What should happen
A sane amount of queries to the db. At least methods like getPathByID, getID and get should only query the db once unless there was a change to the filecache

## What actually happens:
```
{"type":"SUMMARY","reqId":"WUjk9X8AAQEAAFH53iYAAAAJ","time":"2017-06-20T09:03:49+00:00","remoteAddr":"127.0.0.1","user":"switch2","method":"GET","url":"\/oc\/core\/ocs\/v1.php\/apps\/files_sharing\/api\/v1\/shares?format=json&shared_with_me=true&include_tags=true","diagnostics":{"totalSQLQueries":504,"totalSQLDurationmsec":36.458969116211,"totalSQLParams":1000,"totalEvents":22,"totalEventsDurationmsec":291.75233840942}}
```
<details>
  <summary>Click to expand 504 queries</summary>
<pre>
✗ grep WUjk9X8AAQEAAFH53iYAAAAJ diagnostic.log | sort
{"type":"QUERY","reqId":"WUjk9X8AAQEAAFH53iYAAAAJ","diagnostics":{"sqlStatement":"SELECT appid, configkey, configvalue FROM oc_preferences WHERE userid = ?","sqlParams":"array (   0 => 'switch2', )","sqlQueryDurationmsec":0.12421607971191,"sqlTimestamp":1497949429.2164}}
{"type":"QUERY","reqId":"WUjk9X8AAQEAAFH53iYAAAAJ","diagnostics":{"sqlStatement":"SELECT category, categoryid, objid FROM oc_vcategory_to_object r, oc_vcategory WHERE categoryid = id AND uid = ? AND r.type = ? AND objid IN (?)","sqlParams":"array (   0 => 'switch2',   1 => 'files',   2 =>    array (     0 => 306,     1 => 328,     2 => 326,     3 => 324,     4 => 322,     5 => 321,     6 => 325,     7 => 323,     8 => 327,     9 => 343,   ), )","sqlQueryDurationmsec":0.2591609954834,"sqlTimestamp":1497949429.463}}
[...]

PR #39847, opened 2022-03-02
introduce a static fileid->(storage, path) cache

butonic:
This PR significantly reduces the number of queries for `Cache::pathById()` lookups. This is less intrusive than a [full cache implementation](https://github.com/owncloud/core/pull/28165) but it can at least cache and invalidate the id -> path lookup per request.

Before:
```json
{
  "type": "SUMMARY",
  "reqId": "Yh-dLiX1gfglyR5qoetJZwAAAIw",
  "time": "2022-03-02T21:10:07+00:00",
  "remoteAddr": "172.20.0.11",
  "user": "jfd",
  "method": "GET",
  "url": "/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json&include_tags=true",
  "diagnostics": {
    "totalSQLQueries": 2872,
    "totalSQLDurationmsec": 357.25855827331543,
    "totalSQLParams": 3430,
    "totalEvents": 32,
    "totalEventsDurationmsec": 445.83773612976074
  }
}
```

After:
```json
{
  "type": "SUMMARY",
  "reqId": "Yh-dsiX1gfglyR5qoetJfQAAAIY",
  "time": "2022-03-02T21:12:19+00:00",
  "remoteAddr": "172.20.0.11",
  "user": "jfd",
  "method": "GET",
  "url": "/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json&include_tags=true",
  "diagnostics": {
    "totalSQLQueries": 637,
    "totalSQLDurationmsec": 135.25128364562988,
    "totalSQLParams": 1195,
    "totalEvents": 32,
    "totalEventsDurationmsec": 207.22270011901855
  }
}
```

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -83,7 +83,7 @@ class Cache implements ICache {
 	protected $connection;
 
 	/**
-	 * we use a static cache for the path lookup by id because to filecache already knows about all storages
+	 * we use a static cache for the path lookup by id because the filecache already knows about all storages
 	 * and a single trip to the db is sufficient to answer subsequent calls.
 	 * Each entry will contain a storage and a path, such as:
 	 *
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -324,6 +324,7 @@ class Cache implements ICache {
 	 * @param array $data [$key => $value] the metadata to update, only the fields provided in the array will be updated, non-provided values will remain unchanged
 	 */
 	public function update($id, array $data) {
+		unset(self::$path_cache[(int)$id]);
 		if (isset($data['path'])) {
 			// normalize path
 			$data['path'] = $this->normalize($data['path']);
@@ -507,6 +508,7 @@ class Cache implements ICache {
 	public function remove($file) {
 		$entry = $this->get($file);
 		if ($entry !== false) {
+			unset(self::$path_cache[(int)$entry['fileid']]);
 			$sql = 'DELETE FROM `*PREFIX*filecache` WHERE `fileid` = ?';
 			$this->connection->executeQuery($sql, [$entry['fileid']]);
 			if ($entry['mimetype'] === 'httpd/unix-directory') {
@@ -532,6 +534,7 @@ class Cache implements ICache {
 	 * @throws \OC\DatabaseException
 	 */
 	private function removeChildren($entry) {
+		self::$path_cache = [];
 		$subFolders = $this->getSubFolders($entry);
 		foreach ($subFolders as $folder) {
 			$this->removeChildren($folder);
@@ -570,6 +573,7 @@ class Cache implements ICache {
 	 * @throws \Exception if the given storages have an invalid id
 	 */
 	public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
+		self::$path_cache = [];
 		if ($sourceCache instanceof Cache) {
 			// normalize source and target
 			$sourcePath = $this->normalize($sourcePath);
@@ -672,6 +676,7 @@ class Cache implements ICache {
 	 * remove all entries for files that are stored on the storage from the cache
 	 */
 	public function clear() {
+		self::$path_cache = [];
 		Storage::remove($this->storageId);
 	}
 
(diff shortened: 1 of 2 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>cliMain-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/34803/116/1

ownclouders:
:boom: Acceptance tests pipeline <strong>cliProvisioning-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/34803/117/1

micbar:
I think this has some value. Thanks @butonic 

I would appreciate your POV @jvillafanez @mrow4a @DeepDiver1975

mrow4a:
Problem with cache of filecache in OC is always a question: is this class always used for interfacing with filecache or in some places we do direct calls omitting this class. If yes than we should be safe from any desync.

butonic:
@mrow4a true, which is why this pr is limited to the getPathByID lookup. We are neither caching etags or full filecache rows nor are we caching them across requests. The path lookup is only cached per request and needs to be invalidated when a delete or rename operation happens.

For a share mount/unmount the result actually does not need to be invalidated because the path is relative to the storage root, which remains the same.

mrow4a:
ah true, then yes this would work.

butonic:
https://github.com/owncloud/core/pull/28166/files has explored the places that ar bypassing the cache.

I don't see places that affect this cache:
- Jobs an Repair steps don't affect this static cache
- when storages are deleted, the code can no longer instantiate a storage cache with the corret numeric storage id, so the static cache does not need to be cleared
- Test, while green, don't affect runtime

lib/private/Files/Cache/Cache.php
```diff
@@ -922,24 +927,42 @@ public function getIncomplete() {
 		}
 	}
 
+	/**
+	 * we use a static cache for the path lookup by id because to filecache already knows about all storages
+	 * and a single trip to the db is sufficient to answer subsequent calls.
+	 *  @var CE[] $path_cache
+	 */
+	private static $path_cache = [];
```
jvillafanez, line 935: Better use a CappedMemoryCache (https://github.com/owncloud/core/blob/master/lib/private/Cache/CappedMemoryCache.php), so we have limited entries and don't overload the memory.

lib/private/Files/Cache/Cache.php
```diff
@@ -979,3 +1002,14 @@ public function normalize($path) {
 		return \trim(\OC_Util::normalizeUnicode($path), '/');
 	}
 }
+
+class CE {
```
jvillafanez, line 1006: I'd use just an array and document the keys if needed.

lib/private/Files/Cache/Cache.php
```diff
@@ -324,6 +324,7 @@ public function insert($file, array $data) {
 	 * @param array $data [$key => $value] the metadata to update, only the fields provided in the array will be updated, non-provided values will remain unchanged
 	 */
 	public function update($id, array $data) {
+		unset(self::$path_cache[(int)$id]);
```
jvillafanez, line 327: should we try to update the cache instead of removing the element?

lib/private/Files/Cache/Cache.php
```diff
@@ -507,6 +508,7 @@ public function inCache($file) {
 	public function remove($file) {
 		$entry = $this->get($file);
 		if ($entry !== false) {
+			unset(self::$path_cache[(int)$entry['fileid']]);
```
jvillafanez, line 511: I think we should remove the cache after the operation. In case the operation fails, the cache will still have the value so we'll avoid a hit.

mrow4a: over optimisation in my opinion :>

lib/private/Files/Cache/Cache.php
mrow4a: over optimisation in my opinion :>

butonic:
The patch is in production as is. @jvillafanez feel free to take over the PR and polish it.

jvillafanez:
This is ready to review again.

lib/private/Files/Cache/Cache.php
```diff
@@ -81,10 +82,26 @@ class Cache implements ICache {
 	 */
 	protected $connection;
 
+	/**
+	 * we use a static cache for the path lookup by id because to filecache already knows about all storages
```
phil-davis, line 86:
```suggestion
	 * we use a static cache for the path lookup by id because the filecache already knows about all storages
```

phil-davis (commented):
Maybe not worth fixing one word in a comment, unless there are other reasons to change something or rebase...

phil-davis:
I guess this is ready for final review?

"This branch is 3 commits ahead, 102 commits behind master."
@jvillafanez IMO it is worth rebasing to get up-to-date CI.

And then request people again for review. (It looks OK to me)

lib/private/Files/Cache/Cache.php
```diff
@@ -81,10 +82,26 @@ class Cache implements ICache {
...
+	 * ]
+	 */
+	private static $path_cache = null;
+
 	/**
 	 * @param \OC\Files\Storage\Storage|string $storage
 	 */
 	public function __construct($storage) {
+		if (self::$path_cache === null) {
+			self::$path_cache = new CappedMemoryCache();
```
mrow4a, line 102: is 512 too much / too less ?

mrow4a (approved):
default for CappedMemoryCache not clear, and lots of (int) X casting. But overall looks good and logic approved earlier.

lib/private/Files/Cache/Cache.php
jvillafanez: It depends on usage, but I'm not sure how the cache is being used.
* If the access is Ax3, Bx3, Cx3 .... then it's ok. Cache might be smaller.
* If it's A, B, C ... Z and then A, B, C ... Z again, it might not work. It will rely on the cache being big enough to fit all the data, and 512 might be too small. There is a different implementation in WND, and the cache size is configurable, so we might need to move the implementation here if this is the case.
* If it's completely random, it doesn't matter too much since we can't predict which element will be the next one.

PR #39872, opened 2022-03-12
[WIP] Fix shareAccessByID tests

phil-davis:
## Description
1) run acceptance tests with pgsql 10.20 to demonstrate the problem
2) throw ShareNotFound when share id is not an integer to fix the problem

## Related Issue
Issue #39868 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #39873, opened 2022-03-12
throw ShareNotFound when share id is not an integer

### LINKED ISSUE #39868: shareAccessByID tests fail on pgsql 10.3
PR #39865 
https://drone.owncloud.com/owncloud/core/34881/70/13

```
apiShareOperationsToShares2/shareAccessByID.feature:50
apiShareOperationsToShares2/shareAccessByID.feature:51
apiShareOperationsToShares2/shareAccessByID.feature:52
apiShareOperationsToShares2/shareAccessByID.feature:53
```
These scenarios fail against pgsql 10.3 - see PR
And they fail with pgsql 10.20 - https://github.com/owncloud/core/pull/39865#issuecomment-1064932484
But they pass in regular CI with mariaDb.

IMO that used to pass, so there might be a regression.

These are scenarios that send an invalid share-id to try and get share information. The examples with a non-numeric invalid share-id get a 500 response rather than a 404. "something happens" (tm) with processing that with pgsql database.

phil-davis:
## Description
Throw ShareNotFound early when share id is not an integer, rather than attempting to query the database.

Both share-providers only work with integer share-ids. I added the check to both of them, rather than to some higher-level common code, because it is possible that another future share provider could implement share-ids with a different format.

## Related Issue
- Fixes #39868 

## How Has This Been Tested?
CI - see PR #39872 - that runs acceptance tests with pgsql and demonstrates the fails before the code fix, and that it passes after the code fix.

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/tests/FederatedShareProviderTest.php
+++ b/apps/federatedfilesharing/tests/FederatedShareProviderTest.php
@@ -130,6 +130,26 @@ class FederatedShareProviderTest extends \Test\TestCase {
 		parent::tearDown();
 	}
 
+	public function getShareByIdNotExistProvider() {
+		return [
+			[1],
+			[0],
+			[-1],
+			[42.42],
+			["text"],
+			["123text"],
+		];
+	}
+
+	/**
+	 * @dataProvider getShareByIdNotExistProvider
+	 */
+	public function testGetShareByIdNotExist($shareId) {
+		$this->expectException(\OCP\Share\Exceptions\ShareNotFound::class);
+
+		$this->provider->getShareById($shareId);
+	}
+
 	public function testCreate() {
 		$share = $this->shareManager->newShare();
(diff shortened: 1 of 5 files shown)


apps/federatedfilesharing/tests/FederatedShareProviderTest.php
```diff
@@ -130,6 +130,26 @@ public function tearDown(): void {
 		parent::tearDown();
 	}
 
+	public function getShareByIdNotExistProvider() {
+		return [
+			[1],
```
jvillafanez, line 135: we might need a short comment to explain why this is case is expected to fail.

jvillafanez (approved):
Just a tiny optional comment

PR #39874, opened 2022-03-14
[tests-only] [full-ci] Run acceptance tests with pgsql 10.20

phil-davis:
To demonstrate that they now pass.

Issue #39881


phil-davis:
passes - good

PR #39875, opened 2022-03-14
[tests-only] [full-ci] Run acceptance tests with pgsql 9.4

phil-davis:
to check that they all pass

Issue #39881


phil-davis:
passes - good

PR #39876, opened 2022-03-14
[tests-only] [full-ci] Run acceptance tests with MariaDb 10.6

phil-davis:
to check the result

Issue #39881


phil-davis:
passes - good.

PR #39877, opened 2022-03-14
[tests-only] [full-ci] Run acceptance tests with MySql 8.0

phil-davis:
to check the results.

Issue #39881


phil-davis:
passes - good

PR #39908, opened 2022-03-22
Fix sharing with user-key encryption

jvillafanez:
## Description
Having user-key encryption enabled, shares being received by the user might not be properly decrypted.

The cause was a wrong path and a length-base check. Basically a `//files/path/to/file` was used (notice the double `//`) and the check was using a `strlen('/files')`; the resulting path was `s/path/to/file` instead of the expected `/path/to/file`.
When we checked for the shares of the target path (`s/path/to/file`), we didn't find it, so we returned an access list containing only the owner of the file instead of the owner and the sharees. In the end, the key file for the sharee wasn't present in the right place.

## Related Issue
https://github.com/owncloud/encryption/issues/329

## How Has This Been Tested?
Manually tested
1. Setup up user-key encryption
2. user1 shares a file with user2
3. user2 tries to download the file.

With the patch, user2 can preview the file in the file list and he can also download the file without problems

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Encryption/Util.php
+++ b/lib/private/Encryption/Util.php
@@ -235,7 +235,11 @@ class Util {
 		// strip the first 2 directories (expected to be "/<user>/files/path/to/files")
 		// so mountPoint is expected to be "/files/path/to..."
 		$mountPoint = \implode('/', \array_slice($parts, 2));
-		$originalPath = "/{$mountPoint}/{$internalPath}";
+		if ($mountPoint !== '') {
+			$originalPath = "/{$mountPoint}/{$internalPath}";
+		} else {
+			$originalPath = "/{$internalPath}";
+		}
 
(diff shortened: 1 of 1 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>apiProvisioningGroups-v2-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/35034/57

phil-davis (approved):
Looks sensible anyway, to remove the accidental `//` that was being produced if mountPoint was the empty string. Let's merge this, and we will get a new result from nightly CI. If there are other issues, we can followup with another PR(s).

PR #39910, opened 2022-03-23
[tests-only] Make deleteLdapUsersAndGroups backward-compatible for oC10 user_ldap testing

### LINKED ISSUE #729: Acceptance tests failing
Last night all nightly acceptance test pipelines failed with a problem in `BeforeScenario # FeatureContext::before()`

For example https://drone.owncloud.com/owncloud/user_ldap/3872/24/12

`Empty password not allowed - see allowEmptyPassword option. (Laminas\Ldap\Exception\LdapException)`

core PR https://github.com/owncloud/core/pull/39893 was merged yesterday. That made changes to LDAP things in acceptance tests - so the code changes there should be examined.

phil-davis:
## Description
The new code in #39893 tries to delete each user and group one-by-one from the LDAP server. It was not done that way previously - we used to just delete the whole user and group "ou".

Put "if" checks around the new code so that we do not do it when testing oC10. That gets the oC10 tests passing.
(We could investigate in more detail about exactly why deleting one-by-one does not work, but for now just put back the original behavior and get CI passing)

## Related Issue
- Fixes https://github.com/owncloud/user_ldap/issues/729

## How Has This Been Tested?
user_ldap PR https://github.com/owncloud/user_ldap/pull/731 passes.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -853,19 +853,25 @@ trait Provisioning {
 	 * @throws Exception
 	 */
 	public function deleteLdapUsersAndGroups():void {
+		// pdd
+		$isOcisOrReva = OcisHelper::isTestingOnOcisOrReva();
 		foreach ($this->ldapCreatedUsers as $user) {
-			$this->ldap->delete(
-				"uid=" . ldap_escape($user, "", LDAP_ESCAPE_DN) . ",ou=" . $this->ldapUsersOU . "," . $this->ldapBaseDN,
-			);
+			if ($isOcisOrReva) {
+				$this->ldap->delete(
+					"uid=" . ldap_escape($user, "", LDAP_ESCAPE_DN) . ",ou=" . $this->ldapUsersOU . "," . $this->ldapBaseDN,
+				);
+			}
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -853,19 +853,25 @@ public function setLdapSetting(string $configId, string $configKey, string $conf
...
-			);
+			if ($isOcisOrReva) {
+				$this->ldap->delete(
+					"cn=" . ldap_escape($group, "", LDAP_ESCAPE_DN) . ",ou=" . $this->ldapGroupsOU . "," . $this->ldapBaseDN,
+				);
+			}
 			$this->rememberThatGroupIsNotExpectedToExist($group);
 		}
-		if (!$this->skipImportLdif) {
+		if (!$isOcisOrReva || !$this->skipImportLdif) {
```
phil-davis, line 874: Note: we want to always execute this "if" block when we are doing regular oC10 testing (i.e. when not testing oCIS or reva) because in oC10 testing we delete the whole user and group ous and create them again for each test scenario.

individual-it (approved):

PR #39914, opened 2022-03-23
[tests-only] Smaller fixes for LDAP based tests on oCIS

rhafer:
## Description

The OUs for group where mixed up. This went unnoticed since the reva configuration was using the top of the LDAP tree as the search base for users and groups.

This also includes a small change for creating LDAP user when testing against oCIS. Current the default configuration for oCIS requires the mail attribute to be present on the user object. We make sure to add it. (Similar to what is done for the non-LDAP case)

## Motivation and Context
Being able to run the LDAP based test-suite against ocis

### DIFF
--- a/tests/acceptance/features/bootstrap/Provisioning.php
+++ b/tests/acceptance/features/bootstrap/Provisioning.php
@@ -734,6 +734,8 @@ trait Provisioning {
 		}
 		if (isset($setting["email"])) {
 			$entry['mail'] = $setting["email"];
+		} elseif (OcisHelper::isTestingOnOcis()) {
+			$entry['mail'] = $userId . '@owncloud.com';
 		}
 		$entry['gidNumber'] = 5000;
 		$entry['uidNumber'] = $uidNumber;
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Provisioning.php
```diff
@@ -734,6 +734,8 @@ public function createLdapUser(array $setting):void {
 		}
 		if (isset($setting["email"])) {
 			$entry['mail'] = $setting["email"];
+		} elseif (OcisHelper::isTestingOnOcis()) {
```
phil-davis, line 737: Do you want this to happen also when testing on reva?
If so, then there is `isTestingOnOcisOrReva()`

phil-davis (approved):

tests/acceptance/features/bootstrap/Provisioning.php
rhafer: ocis is enough here. It's really one of the ocis services that needs the mail attribute currently.

PR #39935, opened 2022-03-31
Sftp key handling

jvillafanez:
## Description
Modify public/private key behavior for SFTP storage in order to avoid exposing the private key.

The keys will always be generated in the server. The user won't be able to provide his own keys.
The public key will be handled as any "public" parameter, so no encryption or encoding will be applied to this parameter.
For the private key, an opaque token will be sent instead of the actual key. For the user's perspective, this token needs to be sent as private key. The actual private key will be kept in ownCloud and it will be encrypted.

Since the parameters for the storage are different, a migration is provided to convert the old parameters to the new format.

## Related Issue
https://github.com/owncloud/enterprise/issues/5036

## Motivation and Context
Private key shouldn't be exposed in any way.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/appinfo/info.xml
+++ b/apps/files_external/appinfo/info.xml
@@ -13,7 +13,7 @@
 		<admin>admin-external-storage</admin>
 	</documentation>
 	<rememberlogin>false</rememberlogin>
-	<version>0.8.0</version>
+	<version>0.9.0</version>
 	<types>
 		<filesystem/>
 	</types>
--- a/apps/files_external/lib/Lib/RSAStore.php
+++ b/apps/files_external/lib/Lib/RSAStore.php
@@ -57,7 +57,7 @@ class RSAStore {
 	 * instance was there.
 	 * This shouldn't be needed outside of unit tests
 	 */
-	public static function setGlobalInstance(RSAStore $rsaStore) {
+	public static function setGlobalInstance(?RSAStore $rsaStore) {
 		self::$rsaStore = $rsaStore;
 	}
 
--- a/apps/files_external/lib/Lib/RSAStore.php
+++ b/apps/files_external/lib/Lib/RSAStore.php
@@ -1,0 +1,109 @@
+<?php
+/**
+ * @author Juan Pablo Villafáñez Ramos <jvillafanez@owncloud.com>
+ *
+ * @copyright Copyright (c) 2022, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Files_External\Lib;
+
+use OCP\Security\ICredentialsManager;
+use OCP\IConfig;
+use phpseclib3\Crypt\RSA;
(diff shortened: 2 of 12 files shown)


ownclouders:

https://drone.owncloud.com/owncloud/core/35148/82

jvillafanez:
@C0rby could you check if the approach is good enough from a secure point of view?

C0rby:
> @C0rby could you check if the approach is good enough from a secure point of view?

Security wise these changes look good.

There are some other things I want to change though but not in this pull request. Let's do that afterwards.
What I just saw was that the parameters for generating the RSA keypair are not up to todays standards.

tests/lib/Files/External/StorageConfigTest.php
```diff
@@ -55,7 +55,7 @@ public function testJsonSerialization() {
 		$storageConfig->setMountPoint('test');
 		$storageConfig->setBackend($backend);
 		$storageConfig->setAuthMechanism($authMech);
-		$storageConfig->setBackendOptions(['user' => 'test', 'password' => 'password123', 'secure' => true, 'key' => "12345\r\n"]);
+		$storageConfig->setBackendOptions(['user' => 'test', 'password' => 'password123', 'secure' => true, 'key' => "12345"]);
```
mrow4a, line 58: can you elaborate?

apps/files_external/lib/Lib/Storage/SFTP.php
```diff
@@ -92,8 +93,9 @@ public function __construct($params) {
 		}
 		$this->user = $params['user'];
 
-		if (isset($params['public_key_auth'])) {
-			$this->auth = $params['public_key_auth'];
+		if (isset($params['private_key'])) {
```
mrow4a, line 96: this in fact is supposed to be a token and not private_key right?

mrow4a: or this is for backwards compatibility ?

apps/files_external/lib/Lib/RSAStore.php
```diff
@@ -0,0 +1,109 @@
...
+				\OC::$server->getConfig()
+			);
+		}
+		return self::$rsaStore;
+	}
+
+	/**
+	 * Set a new RSAStore instance as a global instance overwriting whatever
+	 * instance was there.
+	 * This shouldn't be needed outside of unit tests
```
mrow4a, line 58: param missing

apps/files_external/appinfo/info.xml
```diff
@@ -13,7 +13,7 @@
 		<admin>admin-external-storage</admin>
 	</documentation>
 	<rememberlogin>false</rememberlogin>
-	<version>0.8.0</version>
+	<version>0.9.0</version>
```
mrow4a, line 16: how migration looks like ? or it will be invisible for the user ? otherwise we need 1.0

tests/lib/Files/External/StorageConfigTest.php
jvillafanez: The key, as any regular parameter, won't be modified in any way. If the key contains additional chars such as `\r` and / or `\n`, those will also be saved and used.
Previously, I think the key was either trimmed, or the new lines were removed.

apps/files_external/lib/Lib/Storage/SFTP.php
jvillafanez: Yes, it will be the token to get the private key. I didn't want to make additional changes because we'd need to adjust more code.
Maybe we need a comment here to hint it's the token what we have there.

apps/files_external/appinfo/info.xml
jvillafanez: It should be transparent to the user. The private key isn't visible (I think it's in a hidden parameter), and the public key should be able to be copied (I don't know if it was possible that the public key was returned with a base64 encoding sometimes).
For the "running" storages, there shouldn't be any change, so the user won't need to touch anything.

mrow4a (approved):

PR #39966, opened 2022-04-08
Allow reuploading a folder after being renamed

jvillafanez:
## Description
Drop an empty folder to the web UI, then, after renaming, you couldn't upload the same empty file. This is fixed in this PR.

## Related Issue
https://github.com/owncloud/enterprise/issues/5025

## Motivation and Context
There is no reason to prevent the upload because the folder doesn't exist any longer. In addition, there wasn't any error message shown to the user. In fact, the expected MKCOL request (after the rename) wasn't happening.

## How Has This Been Tested?
Followed steps described in the issue

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -653,6 +653,12 @@ OC.Uploader.prototype = _.extend({
 			promise = deferred.promise();
 			this._knownDirs[fullPath] = promise;
 
+			// once the promise ends, we'll remove it from the this._knownDirs
+			var that = this;
+			promise.always(function() {
+				delete that._knownDirs[fullPath];
+			});
+
 			// make sure all parents already exist
 			var parentPath = OC.dirname(fullPath);
 			var parentPromise = this._knownDirs[parentPath];
--- a/apps/files/js/file-upload.js
+++ b/apps/files/js/file-upload.js
@@ -654,9 +654,8 @@ OC.Uploader.prototype = _.extend({
 			this._knownDirs[fullPath] = promise;
 
 			// once the promise ends, we'll remove it from the this._knownDirs
-			var that = this;
 			promise.always(function() {
-				delete that._knownDirs[fullPath];
+				delete self._knownDirs[fullPath];
 			});
 
 			// make sure all parents already exist
--- a/changelog/unreleased/39966
+++ b/changelog/unreleased/39966
@@ -1,0 +1,8 @@
+Bugfix: Allow re-uploading the same folder after being renamed
+
+Previously, you couldn't upload a folder, rename it in the web UI and then
+re-upload the same folder.
+
+This behavior is fixed, so you can now re-upload the folder after renaming it
+
+https://github.com/owncloud/core/pull/39966


apps/files/js/file-upload.js
```diff
@@ -653,6 +653,12 @@ OC.Uploader.prototype = _.extend({
 			promise = deferred.promise();
 			this._knownDirs[fullPath] = promise;
 
+			// once the promise ends, we'll remove it from the this._knownDirs
+			var knownDirs = this._knownDirs;
```
JanAckermann, line 657: Would rather go with that=this in the upper scope  otherwise it might be a little hard to read

AlexAndBear (changes requested):
Please also add a change log item

JanAckermann:
We might also want to add acceptance tests for that case

JanAckermann:
We might also want to add acceptance tests for that case

phil-davis:
> We might also want to add acceptance tests for that case

drag-and-drop from an external file explorer window into the browser does not work in the Behat acceptance test framework. We don't have access to open and navigate and drag-drop from outside the browser tabs.

For "internal" drag-and-drop of a file/folder into another folder in the webUI files view, we do have test code that can do it. And for the file upload dialogue we have code that calls an `attachFile` method that allows us to upload without having to navigate in a file explorer. I tried using `attachFile` to "attach" a folder, but that gets an error - which I expected, because the "Upload file" functionality really does only allow uploading a single file.

`apps/files/js/jquery.fileupload.js` has `_getDroppedFiles` which "unpacks" the items that come in `dataTransfer` from the `onDrop` event. To do an automated test, we would have to find some way to insert some data in `dataTransfer` and cause `onDrop` to fire. That will take "some time" to investigate if it is even possible to do from the test framework.

Anyway, I tested manually and the problem is solved by the code in this PR.

phil-davis:
Note: I also noticed and raised issue #39985 while doing manual testing. If it's an easy fix near code that is already being looked at here, then maybe fix it. Otherwise it can be looked at separately "some time".

apps/files/js/file-upload.js
```diff
@@ -653,6 +653,12 @@ OC.Uploader.prototype = _.extend({
 			promise = deferred.promise();
 			this._knownDirs[fullPath] = promise;
 
+			// once the promise ends, we'll remove it from the this._knownDirs
+			var that = this;
```
mrow4a, line 657: standard in other parts of the code is `var self = this;`

jvillafanez: @JanAckermann do you agree? I'm also for `var self = this`

JanAckermann: If this is standard, let's just go for it, we shouldn't waste too much time on it ;)

apps/files/js/file-upload.js
```diff
@@ -653,6 +653,11 @@ OC.Uploader.prototype = _.extend({
 			promise = deferred.promise();
 			this._knownDirs[fullPath] = promise;
 
+			// once the promise ends, we'll remove it from the this._knownDirs
+			promise.always(function() {
+				delete self._knownDirs[fullPath];
```
jvillafanez, line 658: `self` is already defined some lines above

AlexAndBear (approved):

PR #39995, opened 2022-04-21
[tests-only] Test apps code 39992

phil-davis:
In this case source code of an app has changed, so all tests (unit and acceptance) should run.


phil-davis:
acceptance tests did not run - see #39996 for better code.

PR #40000, opened 2022-04-21
Do not crash the command when exception happens on delete

jvillafanez:
## Description
Exception could happen if there is any issue when we're deleting the uploads. If this happens, the command would crash and prevent further cleaning from happening

## Related Issue
https://github.com/owncloud/enterprise/issues/5125

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Command/CleanupChunks.php
+++ b/apps/dav/lib/Command/CleanupChunks.php
@@ -117,8 +117,8 @@ class CleanupChunks extends Command {
 
 			$p->finish();
 			$output->writeln('');
-			if ($failures) {
-				$output->writeln("<error>following folders gave problems with the deletion:</error>");
+			if (\count($failures) > 0) {
+				$output->writeln("<error>there were problems deleting the following folders:</error>");
 				foreach ($failures as $key => $value) {
 					$output->writeln("<error>{$key} -> {$value}</error>");
 				}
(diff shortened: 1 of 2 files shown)


mmattel:
@jvillafanez mind to add a changelog entry. We could approve and merge afterwards.

phil-davis:
And rebase please - the branch is now a long way behind.

phil-davis (commented):
Looks reasonable to me. I am struggling to think how to really test this, because it handles a situation that does not normally/easily happem.

Dev review please.

jvillafanez:
> Looks reasonable to me. I am struggling to think how to really test this, because it handles a situation that does not normally/easily happen.

While probably not the best scenario, you can try to force an error by changing the permissions.
1. Upload a couple of chunks of a big file
2. Change the permissions of the uploaded chunks directly in the FS so they can't be deleted (not sure if the permission change should be applied to the chunk folder too or not)
3. Run the cleanup-chuk command

I'd expect an error to happen because it isn't possible to delete the chunks.

AlexAndBear (approved):

PR #40008, opened 2022-04-24
[full-ci] Encryption added to password config command line.

ahherrera:
## Description

Password encryption is added when running the config:app:set command. Added security to the config:app:get command by hiding the actual value and returning a default value.

## Related Issue

- Fixes <[Elastic Search - Issue 281](https://github.com/owncloud/search_elastic/issues/281)>

## How Has This Been Tested?

Run commands:
- occ config:app:set search_elastic server_password --value "newpassword"
- occ config:app:get search_elastic server_password

Captura de pantalla 2022-04-24 165452

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Command/Config/App/GetConfig.php
+++ b/core/Command/Config/App/GetConfig.php
@@ -32,12 +32,15 @@ class GetConfig extends Base {
 	/** * @var IConfig */
 	protected $config;
 
+	private $encryptKey;
+
 	/**
 	 * @param IConfig $config
 	 */
 	public function __construct(IConfig $config) {
 		parent::__construct();
 		$this->config = $config;
+		$this->encryptKey = ['search_elastic' => 'server_password'];
 	}
 
 	protected function configure() {
@@ -87,6 +90,9 @@ class GetConfig extends Base {
 			$configValue = $this->config->getAppValue($appName, $configName);
 		}
 
+		if (isset($this->encryptKey[$appName]) && $this->encryptKey[$appName] === $configName) {
+			$configValue = IConfig::SENSITIVE_VALUE;
+		}
 		$this->writeMixedInOutputFormat($input, $output, $configValue);
 		return 0;
 	}
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -46,12 +46,15 @@ class AppConfig implements IAppConfig {
 
 	private $configLoaded;
 
+	private $encryptKey;
+
 	/**
 	 * @param IDBConnection $conn
 	 */
 	public function __construct(IDBConnection $conn) {
 		$this->conn = $conn;
 		$this->configLoaded = false;
+		$this->encryptKey = ['search_elastic' => 'server_password'];
 	}
 
 	/**
@@ -152,6 +155,9 @@ class AppConfig implements IAppConfig {
 	public function setValue($app, $key, $value) {
 		if ($value === null) {
 			$value = '';
+		}
+		if (isset($this->encryptKey[$app]) && $this->encryptKey[$app] === $key) {
+			$value = \OC::$server->getCrypto()->encrypt($value);
 		}
 
 		return $this->emittingCall(function (&$afterArray) use (&$app, &$key, &$value) {


ownclouders:
:boom: Acceptance tests pipeline <strong>webUISettingsMenu-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/35406/141

phil-davis:
The code works, good:
```
$ php occ config:app:set testing somepassword --value=chars
$ php occ config:app:get testing somepassword
***REMOVED SENSITIVE VALUE***
```

But:
```
$ php occ config:list
...
        "testing": {
            "enabled": "yes",
            "installed_version": "0.1.0",
            "somepassword": "v2|49bea6fb3ac553bc8546c7364178c7bd|b737befd63ebccb82b8d46ea8626ca49|0a5fa92a1e21188612d34fd0ec452a4e67bcd3fd30ac77446e9307d9af47ac815c345adfb45d3115bf439e3506dc489b3ba090bfbebd73b5c7288b0dafb9dc80",
            "something": "astring",
            "types": "dav,authentication"
        },
...
```

Maybe `config:list` should also have this "REMOVED SENSITIVE VALUE" substitution?

And does there need to be a `--private` option to `config:app:get` (similar to `config:list`) that allows the command-line user to choose to display the sensitive value(s)?

phil-davis:
This changes the core behavior so that app settings with a name that contains the string "password" will now get encrypted.

If any of these currently exist in apps (or in "app core") then there will have to be thought about migration. Fir example:
```
        "password_policy": {
            "enabled": "no",
            "installed_version": "2.1.2",
            "spv_def_special_chars_checked": "",
            "spv_def_special_chars_value": "#!",
            "spv_expiration_nopassword_checked": "",
            "spv_expiration_nopassword_value": "7",
            "spv_expiration_password_checked": "",
            "spv_expiration_password_value": "7",
            "spv_lowercase_checked": "on",
            "spv_lowercase_value": "3",
            "spv_min_chars_checked": "",
            "spv_min_chars_value": "8",
            "spv_numbers_checked": "",
            "spv_numbers_value": "1",
            "spv_password_history_checked": "",
            "spv_password_history_value": "3",
            "spv_special_chars_checked": "",
            "spv_special_chars_value": "1",
            "spv_uppercase_checked": "",
            "spv_uppercase_value": "1",
            "spv_user_password_expiration_checked": "",
            "spv_user_password_expiration_notification_checked": "",
            "spv_user_password_expiration_notification_value": "2592000",
            "spv_user_password_expiration_value": "90",
            "spv_user_password_force_change_on_first_login_checked": "",
            "types": ""
        },
```

```
        "core": {
            "default_encryption_module": "OC_DEFAULT_MODULE",
            "enable_external_storage": "yes",
...
            "shareapi_enabled": "yes",
            "shareapi_enforce_expire_date": "no",
            "shareapi_enforce_expire_date_group_share": "no",
            "shareapi_enforce_expire_date_remote_share": "no",
            "shareapi_enforce_expire_date_user_share": "no",
            "shareapi_enforce_links_password_read_only": "v2|a773ffa5ce456c49ff62f07cf2d9f862|71991ce0a8ee4e0e03d1ec381ac1dc4b|7f8218be82c0ff389001911766570e311d15f4c6a49a0047516dd9175f72cc0fc4181a9a9087be8354ed2b2e36c23f0b27930e2fe0e85ba06c947a027562bf6b",
            "shareapi_enforce_links_password_read_write": "v2|0b3ca024bad4e88e54213cff936a860e|fc1a2e1f312af788242ee2ea52d6ea3b|5278b68762140c03abfe4d624d682ff00dd0cbce7f195233d33792a6be78b69d9108caa0237efd467efc91f4ab3207667ea07a48c955ed5c5367c7f69af40f3c",
            "shareapi_enforce_links_password_write_only": "v2|377839757b72da80596c0dc692c4d764|1a9b680df27d58346938b8a3f5c75ce4|43f1b4d6b899cbcbe604fa419a612a4e417eee510745caa232399100955a4603b4902d77cc9edb25c73c02500f54ab02273cc72aba8c9bdd0fbe5d1c1f537caf",
            "shareapi_exclude_groups": "no",
...
            "umgmt_set_password": "v2|8a1e1dd8dea62f753294072ff9c221b5|50fddd37f5ad99545815bbb88672327f|280b642e5762e1d3b2c8a272f7e62a68d44317f7f8fb1089d34385ede2f61af6afbfb28a1098f0d468f9b9aa79b0322dbb29b3325a07df403f78fd9c8a3b0b75",
            "umgmt_show_backend": "true",
            "umgmt_show_email": "true",
            "umgmt_show_last_login": "true",
            "umgmt_show_password": "false",
            "umgmt_show_quota": "true",
            "vendor": "owncloud"
        },
```

I have already run some acceptance tests with the code in this PR, and they touched some of those `app core` settings and the values have been encrypted. Those settings do not actually store sensitive values - they store boolean yes-no settings.

Code that fetches the value of any of these settings would need to look and see if the value looks encrypted, and if so, then decrypt it.

What to do?

phil-davis:
I added full-ci to the PR title and forced a restart of CI. That will let us see all the test fails. An example is:
https://drone.owncloud.com/owncloud/core/35407/141/18
The test tries to set the UI setting "Set password for new users".
That sends a request to set app "core" key "umgmt_set_password", value "true".
That now gets encrypted. And so when the value is requested by later code, it gets some encrypted string, which idoes not match "true". So the settings stays "false". That setting cannot be enabled.

phil-davis:
https://drone.owncloud.com/owncloud/core/35408/10/7
PHP unit test fails:
```
There was 1 failure:
1) OCA\Files_Sharing\Tests\ApiTest::testEnforceLinkPassword
Failed asserting that true is false.
/drone/src/apps/files_sharing/tests/ApiTest.php:311
```

https://drone.owncloud.com/owncloud/core/35408/50/13
```
apiCapabilities/capabilities.feature:455
apiCapabilities/capabilities.feature:483
apiCapabilities/capabilities.feature:511
```
The value for some fields that have "password" in the key name does not work any more.

https://drone.owncloud.com/owncloud/core/35408/127/17
```
webUIAddUsers/addUsers.feature:34
```
Admin cannot add users with password in the UI any more.

https://drone.owncloud.com/owncloud/core/35408/128/17
```
webUIAdminSettings/adminSharingSettings.feature:33
webUIAdminSettings/adminSharingSettings.feature:38
webUIAdminSettings/adminSharingSettings.feature:43
webUIAdminSettings/adminSharingSettings.feature:48
```
Various settings that "password" in the name do not work any more.

https://drone.owncloud.com/owncloud/core/35408/141/17
```
webUISettingsMenu/settingsMenu.feature:87
```
Sub-admin cannot create users with password.

```
webUISharingPublic1/createPublicLinkShares.feature:321
webUISharingPublic1/permissionsShareByPublicLink.feature:38
webUISharingPublic1/permissionsShareByPublicLink.feature:46
webUISharingPublic1/permissionsShareByPublicLink.feature:55```
```
More sharing settings that do not work because the options have "password" in the name.

jnweiger (changes requested):
While it is in general a good idea to store passwords encrypted, this is currently specific to search_elastic's 'server_password'. I know only one other case, where passwords are stored encrpyted: user_ldap, but with a simpler hashing.

The general substring match for a key that contains `password` is too broad, multiple examples already found by Phil above. It breaks e.g.

```
occ config:app:set core umgmt_set_password --value true
occ config:app:set user_ldap s01ldap_agent_password --value b3duY2xdvWQ=
```

For the case of config:app:get, I don't think we should hide the result. 
In case of search_elastic, the output is an ecrypted string that looks like
`v2|d668abbf3f8611269e0a0634888b1ff2|1691b0bc66...`
Imo, it is perfectly fine, that users see this result in this case.

Other cases should also continue to work. E.g.
```
occ config:app:get core umgmt_set_password
occ config:app:get user_ldap s01ldap_agent_password
```

I'd suggest to go for a more defensive approach, like suggested as expected behaviour in https://github.com/owncloud/search_elastic/issues/281#issue-1209654352

ahherrera:
@jnweiger @phil-davis I have added an array to place in it the keys to be encrypted so as not to encrypt everything with a password.

phil-davis:
@JanAckermann @JammingBen and any oC10 devs. The current core code here has specific knowledge about the search_elastic app. If/when tghis sort of thing changes in search_elastic it will need a matching core change. How do we get around that? Is there some way that an app could "listen" to an "event" that is the setting or getting of a config:app key, and "notify" core that the key is "special" (in this case has a sensitive value to not be displayed as clear-text, or a key that we want core to store with some encryption (or the app can transform the core-supplied value and return to core the actual value that it wants stored)...

jnweiger:
@ahherrera Offloading encryption handling to core may or may not be feasable. We'll see.

In any case, I strongly suggest that each app should self-protect against malformed (or non-decryptable) values. We don't want to enforce 10.9.2 or later for the current elastic search release. In your case, please see, if you can catch the exception in https://github.com/owncloud/search_elastic/issues/281#issue-1209654352 - that blocks the release there.

IljaN:
There is no hook for reading/writing configuration. As @phil-davis mentioned we shouldn't have specific knowledge about this app in core.

The windows_network_drive app for example also handles external password but encrypts/decrypts them by using core encryption apis. Could this approach be also used here?

ahherrera:
@IljaN @jnweiger @phil-davis 
I understand that the responsibility of encrypting the data is of the apps, but the functionality in the command console does not delegate this responsibility to the apps. It would be nice if you could validate that what I did to encrypt is the desired and if not please explain me what is the desired behavior.
THX!!

IljaN:
Sorry my previous comment was a bit  unclear:
- When the user fills in the password in the UI it should be encrypted and stored
- When we talk to elastic we decrypt the password on the fly

What do you think?

IljaN:
Found an undocumented hook which can be used to listen to the config change from search_elasitc:
https://github.com/owncloud/admin_audit/blob/5b9d8999957a157bfe1c91a61d0d753f27fed1b2/appinfo/app.php#L43

IljaN:
Continued in search_elastic repo

ahherrera:
> Continued in search_elastic repo

@IljaN This development cannot be moved to elastic will be because all the methods of the functionality are in the core. Here we have two options:
1- Leave temporarily the solution proposed by me and create a ticket to improve it in the future.
2- Define how we want the development to be in terms of core and other apps responsibilities. This option may take a little more time because it would be necessary to make important changes in the current developments related to saving and obtaining the sensitive and non-sensitive configurations.

jvillafanez:
What about a different approach? Let's assume that everything in the appconfig table, and consequently anything that could be set via `config:app:set` will be visible. If we need to hide secrets or passwords, the app should choose a different place to store them, being the credentials table (through the credentials manager) a reasonable choice.

For this particular case, the password from the elastic search will be stored encrypted in the credentials table (the credentials manager does this for you). This means that we don't need to touch core because all the data in the configuration is expected to be visible.
If we need to add the password from the command line, the app can provide its own command.

If we want to show that there is a password set for elastic search, we can add a different configuration key, such as `occ config:app:get search_elastic has_server_password` that could be set / removed when the password is added / removed. The only problem is that it could be modified with the `occ config:app:set`, so it could be inconsistent. Not sure how important is knowing this information.

PR #40016, opened 2022-04-25
Default for propfind depth infinity should be false as in the past

mrow4a:
Change in https://github.com/owncloud/core/pull/38583/files#diff-efe547293d3c10c67f8014fd439977f1c8f1073dcb65a1a83dc33e7d266dec70R48 release for oc10.9 set dav.propfind.depth_infinity' => true as default and caused some performance issues on some setups. This PR brings back the default as in the past, as discussed in https://github.com/owncloud/core/pull/38583#issuecomment-937842124. 

- [x] review unit tests for potential adjustments

fixes https://github.com/owncloud/enterprise/issues/5154

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1717,11 +1717,12 @@ $CONFIG = [
 /**
  * Allow propfind depth infinity
  *
- * With this setting that defaults to true, propfind requests will now be streamed to reduce memory usage
- * with large responses. It tells the clients whether `depth=infinity` is allowed for propfind requests.
+ * It tells the clients whether `depth=infinity` is allowed for propfind requests.
+ * Streamed infinite depth propfind requests can reduce memory usage
+ * with large responses.
  * For details see: https://datatracker.ietf.org/doc/html/rfc4918#section-10.2
  */
-'dav.propfind.depth_infinity' => true,
+'dav.propfind.depth_infinity' => false,
 
 /**
  * Show the grace period popup
--- a/apps/dav/lib/Capabilities.php
+++ b/apps/dav/lib/Capabilities.php
@@ -45,7 +45,7 @@ class Capabilities implements ICapability {
 					'search-files',
 				],
 				'propfind' => [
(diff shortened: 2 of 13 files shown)


ownclouders:

https://drone.owncloud.com/owncloud/core/35437/51

mmattel:
Hooking myself in to get notified when merged sowe can start a `config-to-docs` run.
@EParzefall FYI

phil-davis:
Note: I am working on the acceptance tests in PR #40018 so that I don't make a mess of this PR.

I will first get them passing, then sort out adding a few scenarios to test:
a) what happens when a client does send a depth infinity request, but depth infinity is disabled
b) enabling depth infinity and a few test cases to be confident that it generally works

config/config.sample.php
```diff
@@ -1696,11 +1696,11 @@
 /**
  * Allow propfind depth infinity
  *
- * With this setting that defaults to true, propfind requests will now be streamed to reduce memory usage
+ * With this setting enabled propfind requests can be streamed to reduce memory usage
```
jvillafanez, line 1699: I don't think this is true. The streaming capabilities aren't linked with this option, or at least they shouldn't be.
You could request a propfind with infinity depth without streaming. It will likely crash due to high memory usage, specially in large environments, but it could be fine if there is few data to send.
You could also request a "standard" propfind with streaming in order to send the data as soon as possible.

The recommendation is to set this option to true only if streaming is available.

mrow4a: adjusted

mrow4a:
@phil-davis I think we should first have this change and then add acceptance on this basis.

phil-davis:
> @phil-davis I think we should first have this change and then add acceptance on this basis.

agree - we just need to do the minimal change to the acceptance tests so that the current tests pass. I will keep working on that in #40018

jvillafanez (approved):

phil-davis:
I rebased and pushed the test changes from #40021 

IMO CI should pass and we can merge.

PR #40018, opened 2022-04-26
[full-ci] [tests-only] Adjust tests for infinite depth changes

phil-davis:
On top of PR #40016 

WIP - need to sort out acceptance tests that use the new default (infinite depth off), and with it on.


phil-davis:
see #40021

PR #40031, opened 2022-04-29
Allow partial initialization of the FS

jvillafanez:
## Description
Using LDAP, setting up the whole FS causes a connection to the LDAP server because we need to mount the shares of any group the user belongs to. This is confusing with some commands that don't expect a connection to the LDAP server since the users are fetched from the DB.

In order to solve this, the commands will setup the FS partially. This means that no mounts will be added other than the home one. This is expected to affect access to shares and external storages. However, the touched commands should need to access to those locations.

Note that, by default, the whole FS will be initialized. For web access, this is done pretty early (in "lib/base.php"), meaning that this feature is mostly limited to CLI.

The main goal of this feature is to prevent hitting the LDAP server. In case the LDAP server isn't accessible, the modified commands will still run instead of crashing. As side effect, there is a minor performance improvement due to the initialization skip (a few ms per FS setup)

## Related Issue
https://github.com/owncloud/enterprise/issues/5081

## Motivation and Context
Some commands crash due to hitting the LDAP server, although they shouldn't hit the LDAP server in the first place.

## How Has This Been Tested?
1. Setup LDAP connection
2. Set `'trashbin_retention_obligation' => '0, 0',` in the config.php file to remove the files from the trashbin without waiting.
3. Add some files in a LDAP account
4. Remove some files from the LDAP account
5. Stop the LDAP server
5. Run the `occ trashbin:expire` command

The command runs without crashing.

The rest of the touched commands and brackground jobs should behave the same way, although they haven't tested yet.

Note that the `occ trashbin:cleanup` command might still crash because it gets the users from the backend directly, not through the account.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Command/CleanupChunks.php
+++ b/apps/dav/lib/Command/CleanupChunks.php
@@ -60,8 +60,8 @@ class CleanupChunks extends Command {
 				'local',
 				'l',
 				InputOption::VALUE_NONE,
-				'only delete chunks that exist on the local filesystem, 
-				this applies to setups with multiple servers connected to the same database and 
+				'only delete chunks that exist on the local filesystem,
+				this applies to setups with multiple servers connected to the same database and
 				chunk folder is not shared among'
 			);
 	}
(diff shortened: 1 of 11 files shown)


jvillafanez:
We might need to include a list with the affected apps in the release notes when this PR is released

DeepDiver1975:
besides the comment above this looks reasonable

DeepDiver1975 (approved):

PR #40006, opened 2022-04-22
[tests-only] Run PHP tests with 7.4-ubuntu20.04

phil-davis:
## Description
WIP - replaces PR #39460 

## Related Issue
#38348 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

Issue #38348, opened 2021-01-27
Unit tests failing on Ubuntu 20.04 with PHP 7.4


phil-davis:
Yesterday the `owncloudci/php:7.4` docker images was updated to use Ubuntu 20.04 as the base image (was previously Ubuntu 18.04). See PR https://github.com/owncloud-ci/php/pull/128

Now some unit tests are failing. See demonstration in PR #38346 

Maybe Ubuntu 20.04 is missing some things that were installed by default in Ubuntu 18.04, or...?

Investigate and fix (or revert the docker image change for now)

phil-davis:
An example in the nightly CI - files_primary_s3 running the core unit tests with PHP 7.4
https://drone.owncloud.com/owncloud/files_primary_s3/2077/8/7
```
There were 8 failures:

1) Test\LargeFileHelperGetFileSizeTest::testGetFileSizeViaCurl with data set #0 ('/var/www/owncloud/server/test...em.txt', 446)
Failed asserting that null is identical to 446.

/var/www/owncloud/server/tests/lib/LargeFileHelperGetFileSizeTest.php:53

2) Test\LargeFileHelperGetFileSizeTest::testGetFileSizeViaCurl with data set #1 ('/var/www/owncloud/server/test...2).txt', 446)
Failed asserting that null is identical to 446.

/var/www/owncloud/server/tests/lib/LargeFileHelperGetFileSizeTest.php:53

3) Test\Preview\BitmapTest::testGetThumbnail with data set #0 (-68, -77)
Failed asserting that 0.77 matches expected 1.41.

/var/www/owncloud/server/tests/lib/Preview/Provider.php:155
/var/www/owncloud/server/tests/lib/Preview/Provider.php:104

4) Test\Preview\BitmapTest::testGetThumbnail with data set #1 (48, 24)
Failed asserting that 0.77 matches expected 1.41.

/var/www/owncloud/server/tests/lib/Preview/Provider.php:155
/var/www/owncloud/server/tests/lib/Preview/Provider.php:104

5) Test\Preview\BitmapTest::testGetThumbnail with data set #2 (-95, 73)
Failed asserting that 0.77 matches expected 1.41.

/var/www/owncloud/server/tests/lib/Preview/Provider.php:155
/var/www/owncloud/server/tests/lib/Preview/Provider.php:104

6) Test\Preview\BitmapTest::testGetThumbnail with data set #3 (49, -34)
Failed asserting that 0.77 matches expected 1.41.

[...]

PR #40059, opened 2022-05-09
[tests-only] Make TemplatesTest more flexible

phil-davis:
## Description
Unit tests in CI currently run with the owncloud-ci/php image based on Ubuntu 18.04.

On Ubuntu 20.04 some unit tests fail. See the related issue. They are failing in CI and locally on my Ubuntu 20.04 system.

This PR adjusts the 2 failing TemplatesTest tests. For some reason, on Ubuntu 20.04 (with whatever settings we have in CI and on my local system), the HTML output of the templates is "nicely formatted" with new-lines and indenting. That makes the test cases fail, because the test cases currently expect a particular exact formatting.

PHP unit test fails: https://drone.owncloud.com/owncloud/core/35632/9/8
This PR addresses items (3) and (4).
```
There were 4 failures:

1) Test\LargeFileHelperGetFileSizeTest::testGetFileSizeViaCurl with data set #0 ('/drone/src/tests/data/lorem.txt', 446)
Failed asserting that null is identical to 446.

/drone/src/tests/lib/LargeFileHelperGetFileSizeTest.php:53

2) Test\LargeFileHelperGetFileSizeTest::testGetFileSizeViaCurl with data set #1 ('/drone/src/tests/data/strängé...2).txt', 446)
Failed asserting that null is identical to 446.

/drone/src/tests/lib/LargeFileHelperGetFileSizeTest.php:53

3) Tests\Core\Templates\TemplatesTest::test403
--- Expected
+++ Actual
@@ @@
 '<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">\n
-<html><body><ul><li class="error">\n
-		Access forbidden<br><p class="hint"></p>\n
-</li></ul></body></html>\n
+<html><body><ul>\n
+<li class="error">\n
+		Access forbidden<br>\n
+		<p class="hint"></p>\n
+	</li>\n
+</ul></body></html>\n
 '

/drone/src/tests/lib/TestCase.php:510
/drone/src/tests/lib/TestCase.php:488
/drone/src/tests/Core/Templates/TemplatesTest.php:9

4) Tests\Core\Templates\TemplatesTest::test404
--- Expected
+++ Actual
@@ @@
 '<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">\n
-<html><body><ul><li class="error">\n
-			File not found<br><p class="hint">The specified document has not been found on the server.</p>\n
-<p class="hint"><a href="/owncloud/">You can click here to return to ownCloud.</a></p>\n
-		</li></ul></body></html>\n
+<html><body><ul>\n
+<li class="error">\n
+			File not found<br>\n
+			<p class="hint">The specified document has not been found on the server.</p>\n
+			<p class="hint"><a href="/owncloud/">You can click here to return to ownCloud.</a></p>\n
+		</li>\n
+	</ul></body></html>\n
 '

/drone/src/tests/lib/TestCase.php:510
/drone/src/tests/lib/TestCase.php:488
/drone/src/tests/Core/Templates/TemplatesTest.php:16
--
```

The changed test code "normalizes" the expected and actual HTML before comparing. That way the test will not be so brittle - it will work even when the formatting of the HTML is a bit different.

## Related Issue
#38348 

## How Has This Been Tested?
CI in this PR (demonstrates that the test still passes on the Ubuntu 18.04 CI image)
Local unit test run on my Ubuntu 20.04 laptop - passes
CI of PR #40006 (demonstrates that these 2 tests pass on the Ubuntu 20.04 CI image)

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/templates/403.php
+++ b/core/templates/403.php
@@ -11,7 +11,7 @@ if (!isset($_)) {//also provide standalone error page
 ?>
 <ul>
 	<li class='error'>
-		<?php p($l->t('Access forbidden')); ?><br>
+		<?php p($l->t('Access forbidden')); ?><br/>
 		<p class='hint'><?php if (isset($_['file'])) {
 	p($_['file']);
 }?></p>
--- a/core/templates/404.php
+++ b/core/templates/404.php
@@ -17,7 +17,7 @@ if (!isset($_)) {//also provide standalone error page
 <?php else: ?>
 	<ul>
 		<li class="error">
-			<?php p($l->t('File not found')); ?><br>
+			<?php p($l->t('File not found')); ?><br/>
 			<p class="hint"><?php p($l->t('The specified document has not been found on the server.')); ?></p>
 			<p class="hint"><a href="<?php p(\OC::$server->getURLGenerator()->linkTo('', 'index.php')) ?>"><?php p($l->t('You can click here to return to %s.', [$theme->getName()])); ?></a></p>
 		</li>
--- a/tests/Core/Templates/TemplatesTest.php
+++ b/tests/Core/Templates/TemplatesTest.php
@@ -5,14 +5,14 @@ namespace Tests\Core\Templates;
 class TemplatesTest extends \Test\TestCase {
 	public function test403() {
 		$template = \OC::$SERVERROOT . '/core/templates/403.php';
-		$expectedHtml = "<ul><li class='error'>\n\t\tAccess forbidden<br><p class='hint'></p></li></ul>";
+		$expectedHtml = "<ul><li class='error'>Access forbidden<br><p class='hint'></p></li></ul>";
 		$this->assertTemplate($expectedHtml, $template);
 	}
 
 	public function test404() {
 		$template = \OC::$SERVERROOT . '/core/templates/404.php';
 		$href = \OC::$server->getURLGenerator()->linkTo('', 'index.php');
-		$expectedHtml = "<ul><li class='error'>\n\t\t\tFile not found<br><p class='hint'>The specified document has not been found on the server.</p>\n<p class='hint'><a href='$href'>You can click here to return to ownCloud.</a></p>\n\t\t</li></ul>";
+		$expectedHtml = "<ul><li class=\"error\">File not found<br><p class=\"hint\">The specified document has not been found on the server.</p><p class=\"hint\"><a href=\"$href\">You can click here to return to ownCloud.</a></p></li></ul>";
 		$this->assertTemplate($expectedHtml, $template);
 	}
 }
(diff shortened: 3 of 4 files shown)


phil-davis:
@jvillafanez @JanAckermann @JammingBen can someone review this please. IMO it should be an easy review - no change to real code, just making the unit test more flexible.

jvillafanez:
Did you try checking with https://phpunit.readthedocs.io/en/latest/assertions.html#assertxmlstringequalsxmlstring ? I'm not sure if it will work, but it's worthy to try.

phil-davis:
> Did you try checking with https://phpunit.readthedocs.io/en/latest/assertions.html#assertxmlstringequalsxmlstring ? I'm not sure if it will work, but it's worthy to try.

I tried just now. `assertXmlStringEqualsXmlString` complains about the `<br>` element in `403.php` and `404.php`. If I make those `<br/>` then it parses the strings as XML OK. So I fixed those 2 files.

But then it complained about a lot of white-space differences. There is space that is "printable" (actually would be part of the text that could be rendered). In real life that white-space makes no visible difference when a browser renders it as HTML. But s XML, the PHP assert sees it as different.

To make it pass, I needed use my `normalizeHTML` function anyway. (or had do to that sort of white-space cleanup anyway)

I pushed a commit - @jvillafanez please review again.

jvillafanez:
> But then it complained about a lot of white-space differences. There is space that is "printable" (actually would be part of the text that could be rendered). In real life that white-space makes no visible difference when a browser renders it as HTML. But s XML, the PHP assert sees it as different.

I think that should be ok. I mean, if the text is different because there are white-spaces, then it different. We could considered it as a malformed document that should be fixed.

I'm having doubts with the `normalizedHTML` function because I don't think it covers all the cases, but if you're ok with it let's merge it.

jvillafanez (approved):

phil-davis:
> I'm having doubts with the normalizedHTML function because I don't think it covers all the cases, but if you're ok with it let's merge it.

It is not intended to be a "general normalizer". It just does the "obvious" things that might be text white-space new-line differences which we know/decide that we don't care about for the purposes of the comparison in this test. If there was a need, then some full normalizer could be written (or find one on GitHub). But actually, we just want to allow some output variation to cope with the white-space differences in rendering that have been observed on different Ubuntu versions - presumably due to some HTML rendering utility version difference somewhere. And the difference is not so important, so we don't want to spend more time to exactly find out which version of which utility is causing this white-space difference.

Merging - this is a tests-only change, so we don't need to 'freeze" that, waiting for release-10.10.0

PR #40087, opened 2022-05-19
Add config lang

AlexAndBear:
## Description
Requirement for:
https://github.com/owncloud/web/issues/6243

## Related Issue

- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/provisioning_api/lib/Users.php
+++ b/apps/provisioning_api/lib/Users.php
@@ -225,6 +225,7 @@ class Users {
 		$data['home'] = $targetUserObject->getHome();
 		$data['two_factor_auth_enabled'] = $this->twoFactorAuthManager->isTwoFactorAuthenticated($targetUserObject) ? 'true' : 'false';
 		$data['last_login'] = $targetUserObject->getLastLogin();
+		$data['language'] =  $this->config->getUserValue($targetUserObject->getUID(), 'core', 'lang', $this->config->getSystemValue('default_language', 'en'));
 
 		return new Result($data);
 	}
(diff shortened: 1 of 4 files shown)


jvillafanez:
The idea looks ok to me.

Do we have any plans for adding more information in the "config" key? Otherwise, it seems better to use the "language" key directly.
In addition, should we include the default configured language if the user hasn't configured anything? or maybe include the default in a different key?

phil-davis:
> In addition, should we include the default configured language if the user hasn't configured anything? or maybe include the default in a different key?

The code could return the "effective" language for the user. If the user has explicitly set a language, then return that. If the user has not set a language then return the system default language (which is the language that the user is going to be shown when they log in). There must already be a function somewhere for `getEffectiveLanguage`

phil-davis:
> get default system lang if not user specified

works

ownclouders:
:boom: Acceptance tests pipeline <strong>apiSharingNotificationsToRoot-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/35794/103

kulmann (approved):
😍

kulmann:
> The code could return the "effective" language for the user. If the user has explicitly set a language, then return that. If the user has not set a language then return the system default language (which is the language that the user is going to be shown when they log in). There must already be a function somewhere for `getEffectiveLanguage`

Fully agree, clients don't/shouldn't know or care where the language key comes from. They only need a value. The chain of fallbacks as proposed and implemented is perfect. 👍  (user choice -> system default -> hardcoded "en")

Issue #40105, opened 2022-05-27
phpstan new release 1.7.2 fails CI


phil-davis:
```
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Installing phpstan/phpstan (1.7.2): Extracting archive
Generating autoload files
1 package you are using is looking for funding.
Use the `composer fund` command to find out more!
php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan analyse --memory-limit=2G --configuration=./phpstan.neon --level=0 apps core settings lib/private lib/public ocs ocs-provider
 1466/1466

 -- ------------------------------------------------------------------------------------------------------------- 
     Error                                                                                                        
 -- ------------------------------------------------------------------------------------------------------------- 
     Child process error (exit code 255): PHP Fatal error:  Cannot declare class Laminas\Validator\EmailAddress,  
     because the name is already in use in                                                                        
     /home/phil/git/owncloud/core/lib/composer/laminas/laminas-validator/src/EmailAddress.php on line 31          
                                                                                                                  
     Child process error (exit code 255): PHP Fatal error:  Cannot declare class Laminas\InputFilter\Factory,     
     because the name is already in use in                                                                        
     /home/phil/git/owncloud/core/lib/composer/laminas/laminas-inputfilter/src/Factory.php on line 12             
                                                                                                                  
 -- ------------------------------------------------------------------------------------------------------------- 


                                                                                                                     
 [ERROR] Found 2 errors                                                                                  
```

If I pin to phpstan 1.7.1 then it passes.

phil-davis:
I don't see what really changed in https://github.com/phpstan/phpstan/releases/tag/1.7.2

But I also don't find anywhere in oC10 core codebase that uses these laminator classes in an obviously wrong way.

phil-davis:
See phpstan issue https://github.com/phpstan/phpstan/issues/7328

We can go to a newer phpstan release after that issue has been closed.

PR #40108, opened 2022-05-27
[PoC-only] Phpstan 1.7.2

phil-davis:
I seem to be able to get phpstan 1.7.2 to pass if I move the PHP dependencies out of `lib/composer` and into `vendor`

We do not actually want to do this. We would need to track down all other references to `lib/composer` and make sure not to break anything.

This PR is a demonstration that the problem with phpstan 1.7.2 is somehow related to having:
```
"vendor-dir": "lib/composer",
```
in `composer.json`

Related issue #40105 

Note: this PR is on top of #39368 because I want to start from that more up-to-date point of PHP dependencies.


phil-davis:
https://drone.owncloud.com/owncloud/core/35884/6/8 - `phpstan` passed

https://drone.owncloud.com/owncloud/core/35884/8/7 - phan is failing. There would be a bunch more things to do if we actually did move the dependencies to `vendor`. I am not spending time looking at that.

phil-davis:
Fixed in #40112

Issue #40093, opened 2022-05-23
Flaky acceptance tests because suite expects an ordered list of shares


aduffeck:
The acceptance test suite expects the list of shares returned by the ocs API to be ordered, which is currently not guaranteed (see https://github.com/owncloud/core/blob/master/tests/acceptance/features/bootstrap/Sharing.php#L2037).

The cs3 share manager for examples uses uuids as share ids and sorts the returned list by those (I believe) which causes flaky tests as seen in https://drone.owncloud.com/owncloud/ocis/12005.

aduffeck:
**Note:** The list of received shares is also not guaranteed to be sorted by creation date, just like the list of owned shares. I'm not sure if that needs to be adjusted in the test suite as well.

phil-davis:
Thinking... - might need to refactor a bit to remember "the last share" at an earlier point in the test code.

phil-davis:
Example PR with this problem: https://github.com/owncloud/ocis/pull/3697

phil-davis:
This should be improved (or fixed) by https://github.com/owncloud/core/pull/40114

The refactored API test code is running in:
oCIS https://github.com/owncloud/ocis/pull/3908
reva edge https://github.com/cs3org/reva/pull/2915
reva master https://github.com/cs3org/reva/pull/2914

I will deal with the 2 new test fails in reva (issue https://github.com/cs3org/reva/issues/2916 ) separately. I don't think that they impact the test code in any significant way.

Please ping me if any problems are found, and close when you think this is working well.

PR #40114, opened 2022-06-01
[tests-only] [full-ci] More flexible tests for share ids

phil-davis:
## Description
The core acceptance test steps often refer to "the last created share". Most of the time, the existing code would do an API request to get a list of shares for the user, then choose the returned element that had the largest share id (and I think was usually last in the list). This logic assumed that share id is an ever-increasing value, and/or that the returned response was in some ffirst-to-last order. Those assumptions/heuristics are not valid in all implementations. An implementation (like oCIS or reva) can have share id as some alpha-numeric string that is not "increasing". The list of shares in a response does not have to be in any date-time first-to-last order.

This PR removes that logic from the acceptance test suite.

The new code remembers:
- the response to the last share by user, which contains the share id.
- the share id of that last share by user (this allows finding the last share that was shared by a particular user in the scenario)
- the user name of the user that last created a share (this helps for finding the very last share that was created in a scenario)
- the response to the last creation of a public link share (not by user)
- the share id of the last created public link share

Test steps that actually reference a public link share have been refactored to be like:
```
Given the administrator has expired the last created public link share using the testing API
When user "Alice" tries to update the last public link share using the sharing API with
```
They mention "the last created public link share". Corresponding test step code implementation makes use of the remember last public link share id.

We also have to remember shares that were created in webUI test scenarios.

For public link shares, the webUI has the full link to the new public link share. That has the public-link-token on the end. After creating a public link in the webUI, the test code now uses the API  to get the shares of the user, and finds the share that has the just-created token. It remembers that share data and share id as the "last created public link share".

For shares to users and groups, the webUI has no easy "hook" to the share id of the just-created share. But the test code knows where the user has navigated to in the webUI, and the name of the resource that was just shared. So the test code uses the API to get the shares of the user, and finds the share that has the expected path. It remembers that share data and share id as the "last created share" for the user.

This way, and `Then` steps later in either an API or a webUI test scenario can mention the "last created share" or "last created public link share", and the remembered data will already be up-to-date in the test-runner.

There are some edge cases handled. For example, if a share action is tried, but not expected to work (should give a 4xx response, or the webUI should give some error notification...), and so the share actually will not exist, that is passed through the test-runner code so that it does not try to get and remember the "last created share" in that case.

## Related Issue
#40093 

## How Has This Been Tested?
CI passed here in core
oCIS https://github.com/owncloud/ocis/pull/3906 - passed
reva https://github.com/cs3org/reva/pull/2912 - passed except see comment https://github.com/cs3org/reva/pull/2912#issuecomment-1143564584 - I can look into that later.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiShareCreateSpecialToRoot1/createShareExpirationDate.feature
+++ b/tests/acceptance/features/apiShareCreateSpecialToRoot1/createShareExpirationDate.feature
@@ -831,10 +831,9 @@ Feature: a default expiration date can be specified for shares with users or gro
       | path       | /textfile0.txt  |
       | shareWith  | brand-new-group |
       | expireDate | +15 days        |
-    And the administrator has expired the last created share using the testing API
+    And the administrator has expired the last created public link share using the testing API
     When the public accesses the preview of file "textfile0.txt" from the last shared public link using the sharing API
     Then the HTTP status code should be "404"
-    And user "Alice" should not see the share id of the last share
     And as "Alice" file "/textfile0.txt" should exist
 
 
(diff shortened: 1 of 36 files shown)


micbar (approved):

phil-davis:
> passing also in ocis. Should we merge?

yes, I think so. I will see the reva result in a few minutes, and if it passes or has only known expected-failures differences, then I will merge this...

PR #40125, opened 2022-06-08
[tests-only] [PoC] Latest dependencies 20220608

phil-davis:
This has code from #39368 and #40122 merged together, with the latest `composer update` done.
It checks that everything works fine with PHP 7.4 and all the latest dependencies.

PR #40127, opened 2022-06-08
[tests-only] [full-ci] Behat test runner php8

phil-davis:
Run the behat acceptance test pipeline step with PHP 8.1 to see if it runs OK.

On top of PR #40125


phil-davis:
https://drone.owncloud.com/owncloud/core/36004/112/14
```
cliEncryption/encryption.feature:27
```
That fails because the scenario does `decrypt-all`. To do that `occ` command it needs to turn on `singleuser` mode. If we did that via the testing app, then the testing app (and all other API endpoints) get disabled, and the test cannot turn off `singleuser` mode. So the test does the `occ` commands directly as `\system()` calls. The test-runner has PHP 8.1, and we know the oC10 does not support that yet. So the `occ` commands fail. This is a situation where the test-runner must use a PHP version that is supported by the system-under-test. I can't think of a "simple/reasonable" way to get around that.

phil-davis:
https://drone.owncloud.com/owncloud/core/36004/111/13
```
  Scenario: Update of minor version of an app                                         # /drone/src/tests/acceptance/features/cliManageApps/appUpgrade.feature:10
    Give app "updatetest" with version "2.0.0" has been put in dir "apps"            # AppManagementContext::appHasBeenPutInDir()
    And app "updatetest" has been enabled                                             # FeatureContext::appHasBeenDisabled()
    And app "updatetest" has been disabled                                            # FeatureContext::appHasBeenDisabled()
    When the administrator puts app "updatetest" with version "2.1.0" in dir "apps"   # AppManagementContext::putAppInDir()
    And the administrator enables app "updatetest"                                    # FeatureContext::adminEnablesOrDisablesApp()
    And the administrator runs upgrade routines on local server using the occ command # OccContext::theAdministratorRunsUpgradeRoutinesOnLocalServerUsingTheOccCommand()
      │ This version of ownCloud is not compatible with PHP 8.0
      │ You are currently running PHP 8.1.6.
      │ This version of ownCloud is not compatible with PHP 8.0
      │ You are currently running PHP 8.1.6.
      │ 
    Then the installed version of "updatetest" should be "2.1.0"                      # AppManagementContext::assertInstalledVersionOfAppIs()
      Return code not found after executing 'occ config:app:get updatetest installed_version --no-ansi'. Is the testing app installed and enabled?
      <?xml version="1.0"?>
      <ocs>
       <meta>
        <status>failure</status>
        <statuscode>503</statuscode>
        <message>Service unavailable</message>
       </meta>
       <data/>
      </ocs>
       (Exception)
Failed asserting that 503 matches expected '200'.
```

This is another situation where a local `occ` command has to be done. After creating and enabling the new app, the scenario needs to do `occ upgrade`. At this point the testing app (and other oC10 apps and API endpoints) are not responding. So `occ upgrade` has to be done directly, and not through the testing app.

Everything else passes, so that is good.

phil-davis:
The one needed fix is applied to core master code by PR #40128

PR #40155, opened 2022-06-21
Add batch action to allow sharing with multiple users at once

JammingBen:
## Description
This feature allows sharing resources with multiple users at once via the following format: *user1, user2, user3*

If you have the `guest` app installed and active, https://github.com/owncloud/guests/pull/506 is required to properly use this feature. Unfortunately, there is no way to overcome this as the `guest` app extends the sharing functionality in a very bad way by re-implementing it. 

## Related Issue

- Works towards https://github.com/owncloud/enterprise/issues/2865

## Screenshot

image

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/40155
+++ b/changelog/unreleased/40155
@@ -1,0 +1,7 @@
+Enhancement: Allow sharing with multiple users at once
+
+It is now possible to share resources with multiple users at once
+via the following format: user1;user2;user3.
+
+https://github.com/owncloud/core/pull/40155
+https://github.com/owncloud/enterprise/issues/2865
--- a/changelog/unreleased/40155
+++ b/changelog/unreleased/40155
@@ -1,7 +1,7 @@
 Enhancement: Allow sharing with multiple users at once
 
 It is now possible to share resources with multiple users at once
-via the following format: user1;user2;user3.
+via the following format: user1, user2, user3.
 
 https://github.com/owncloud/core/pull/40155
 https://github.com/owncloud/enterprise/issues/2865
--- a/core/js/sharedialogview.js
+++ b/core/js/sharedialogview.js
@@ -277,16 +277,25 @@
 						if (isBatch) {
 							return view._getUsersForBatchAction(trimmedSearch).then(function (res) {
 								if (res.found.length) {
-									suggestions.push({ batch: res.found, failedBatch: res.notFound, label: trimmedSearch, value: {} });
+									var labelArray = [];
+									for (i = 0; i < res.found.length; i++) {
+										labelArray.push(res.found[i].shareWith)
+									}
+
+									suggestions.push({ batch: res.found, failedBatch: res.notFound, label: labelArray.join(', '), value: {} });
 								}
 
 								$loading.addClass('hidden');
 								$loading.removeClass('inlineblock');
 								if (suggestions.length) {
+									$('.shareWithField').removeClass('error')
+										.tooltip('hide')
+										.autocomplete("option", "autoFocus", true);
 									return response(suggestions, result);
 								}
 
 								view._displayError(t('core', 'No users or groups found'));
+								return response(undefined, result);
 							})
 						}
 
--- a/core/js/sharedialogview.js
+++ b/core/js/sharedialogview.js
@@ -96,7 +96,7 @@
 		shareeListView: undefined,
 
 		/** @type {string} **/
-		batchActionSeparator: ', ',
+		batchActionSeparator: ',',
 
 		events: {
 			'input .shareWithField': 'onShareWithFieldChanged',
@@ -604,6 +604,8 @@
 			var users = Array.from(new Set(search.split(this.batchActionSeparator)));
 
 			for (var i = 0; i < users.length; i++) {
+				if (!users[i]) continue;
+				var user = users[i].trim();
 				promises.push(
 					$.ajax({
 						url: OC.linkToOCS('apps/files_sharing/api/v1') + 'sharees',
@@ -611,11 +613,11 @@
 						dataType: 'json',
 						data: {
 							format: 'json',
-							search: users[i],
+							search: user,
 							perPage: 200,
 							itemType: view.model.get('itemType')
 						},
-						context: { user: users[i] }
+						context: { user: user }
 					})
 					.done(function (result) {
 						if (result.ocs.data.exact.users.length) {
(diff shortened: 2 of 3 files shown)


ownclouders:

https://drone.owncloud.com/owncloud/core/36102/128

mmattel:
Docs relevant if merged. Pls file a docs-server issue when ready.

changelog/unreleased/40155
```diff
@@ -0,0 +1,7 @@
+Enhancement: Allow sharing with multiple users at once
+
+It is now possible to share resources with multiple users at once
+via the following format: user1;user2;user3.
```
phil-davis, line 4:
```suggestion
via the following format: user1, user2, user3.
```

The semi-colon separator does not work for me.

core/js/sharedialogview.js
```diff
@@ -95,6 +95,9 @@
 		/** @type {object} **/
 		shareeListView: undefined,
 
+		/** @type {string} **/
+		batchActionSeparator: ', ',
```
phil-davis, line 99:
```suggestion
		batchActionSeparator: ',',
```

This could be just a comma, and the code could normalize the input by removing any white-space (thus allowing users to put white-space or not)

phil-davis (commented):
If I paste in a string like "qwerty, bbbb, ccc" and only user "bbbb" exists, it still suggests to add all 3 "users". Then I get a message about not being able to add "qwerty" and "ccc".

Could it validate the list up-front? And let the user know that some of the users are not found?

I tried some combinations using the user's display name in the list. That works. You just have to be careful to provide the exact display name.

phil-davis:
Note: this feature (currently) never shares with groups. I tried with users "bbbb" and "cccc", and groups "abc" and "cccc".

When I try to share with "abc, cccc" it shares with user "cccc" and tells me "Could not be shared with the following users: abc"

I guess that the feature is only for sharing with multiple users. It will be trickier to include groups that match, because there can be a user and group with the same name. So some system would have to be designed to help the user when such and ambiguity happened.

JammingBen:
> the "No users or groups found for" popup should go away when I start typing more stuff (when I start typing "cccc"

Yes, that's a general problem of the input field. But I might as well tackle it with this PR 👍 

> why do we insist on comma followed by space as the separator? The UI could allow any combination of spaces and commas, and that would make it easier for users to enter "whatever they like within reason".

Hmm I need to think about that one. Let's say you have a group called `aaa, bbb`. You couldn't share with this group if we remove the white spaces internally, so we'd need to search for both cases... Let me check again how it behaves, but if it makes things more complicated, I would leave it as it is.

> Note: this feature (currently) never shares with groups. I tried with users "bbbb" and "cccc", and groups "abc" and "cccc".

Correct, the batch action only invites users. However, when having a group called `aaa, bbb` and a user `aaa` (and/or `bbb`), it should display 2 options for the user: share with user(s) or with the group.

> If I paste in a string like "qwerty, bbbb, ccc" and only user "bbbb" exists, it still suggests to add all 3 "users". Then I get a message about not being able to add "qwerty" and "ccc".
>
> Could it validate the list up-front? And let the user know that some of the users are not found?

Good idea, I think that's relatively easy to implement as we already have the failed users.

phil-davis:
> Correct, the batch action only invites users. However, when having a group called aaa, bbb and a user aaa (and/or bbb), it should display 2 options for the user: share with user(s) or with the group.

That works:
Screenshot from 2022-06-28 18-56-55

I hadn't thought about that. So, the code should always send off to see if there is a match for the full literal string entered - it might be a group. And if there are comma(s) in the string (with or without white-space around the comma(s)) then ignore the white-space and suggest adding multiple users.

phil-davis:
The user and group suggestions are both blue. Single user suggestions are green. Is that the way the colors are intended to be?

I guess multiple users is a "group of users", so that's why it is blue?

phil-davis:
Looking good - I have tried various combinations of user and group names and having comma-separator with and without a space after the comma. Works.

This situation is a bit odd:
1. users aaaa bbbb cccc exist
2. login as aaaa and create a folder
3. try to share the folder with bbbb,unknown,cccc
4. the suggestion comes up to share with multiple users bbbb,cccc - good "unknown" does not exist, so it is not in the suggested list of matches.
5. click to share with those suggested users
6. shares appear with bbbb and cccc - good
7. A notification appears with "Could not be shared with the following users: unknown" - not so good.

A request to share with "unknown" should not be sent. The JS already knows that "unknown" has no match.

phil-davis:
There is an "i" information popup to the right of the share-with text box that has stuff about "Share with people on other ownClouds using the syntax..."

Maybe some instructions could be added there about the syntax to use to share with multiple users? It would be good to have some clue given in the UI that this feature exists.

phil-davis:
Note: I added issue #40178 to add some webUI automated test coverage. There is no big rush for that, if it is ready before this PR is merged then we can add a commit here. If this PR is already merged, then we add the test coverage in a separate PR  to master.

mmattel:
Ping, CI is green, approvable? mergable?

JammingBen:
> @JammingBen I made comment https://github.com/owncloud/core/pull/40155#issuecomment-1170132354
> 
> Are you progressing this? Maybe just sort out that little thing, and we can merge.

I'm still on vacation until next Monday, I can work that stuff out then. Although I'm not sure about the expected behavior, IMO it's fine to display that the resource could not be shared with `unknown`. Why shouldn't it be displayed?

phil-davis:
I added @jvillafanez and @JanAckermann as reviewers.
My comments are only minor, and we need to move forward. Please review, and this can be merged.

jvillafanez:
While unlikely, it might be possible that a group contain the separator, for example a group `music, pictures and arts`. Such group should appear somehow.
Maybe it's easy to include an additional search without considering the separator in order to find that group.

phil-davis:
> While unlikely, it might be possible that a group contain the separator, for example a group `music, pictures and arts`. Such group should appear somehow. Maybe it's easy to include an additional search without considering the separator in order to find that group.

See https://github.com/owncloud/core/pull/40155#issuecomment-1168711596

If a group matches the comma-separated string then it is shown in the drop-down list as a possible thing to share with.

jvillafanez (approved):

PR #40165, opened 2022-06-23
[full-ci] Make add group error responses consistent

### LINKED ISSUE #40164: Error responses to add group API requests are inconsistent
### Steps to reproduce
The webUI uses `/index.php/settings/users/groups` endpoint to "do its thing", rather than the Provisioning API https://doc.owncloud.com/server/next/developer_manual/core/apis/provisioning-api.html#add-group-example

1. On the webUI, try to add a group with a space at the end of the name. Inspect the response in the browser.
2. Create a group with the Provisioning API (and do not refresh the webUI)
```
$ curl -X POST http://admin:admin@172.17.0.1:8080/ocs/v1.php/cloud/groups -d groupid="somegroup"
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data/>
</ocs>
```

3. On the webUI, try to add a group with that name `somegroup`. Inspect the response in the browser.

### Expected behaviour
Both error responses have the same format.

### Actual behaviour
Response 1:
```
{"status":"error","data":{"message":"Unable to add group."}}
```

Response 2:
```
{"message":"Group already exists."}
```

One response has the message inside "data", the other has the message not inside "data".
This makes it difficult for the webUI to correct parse the error response and show it to the user.

### Server configuration
Current core master

phil-davis:
## Description
See the issue for details of what was the problem. This PR makes the error responses to "create group" consistent - the error message text is not enclosed in a `'data'` structure. So the webUI can now find the message consistently.

## Related Issue
- Fixes #40164 

## How Has This Been Tested?
Local run of test scenarios.

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/settings/Controller/GroupsController.php
+++ b/settings/Controller/GroupsController.php
@@ -104,6 +104,7 @@ class GroupsController extends Controller {
 		if ($this->groupManager->groupExists($id)) {
 			return new DataResponse(
 				[
+					'status' => 'error',
 					'message' => (string)$this->l10n->t('Group already exists.')
 				],
 				Http::STATUS_CONFLICT
@@ -121,9 +122,7 @@ class GroupsController extends Controller {
 		return new DataResponse(
 			[
 				'status' => 'error',
-				'data' => [
-					'message' => (string)$this->l10n->t('Unable to add group.')
-				]
+				'message' => (string)$this->l10n->t('Unable to add group.')
 			],
 			Http::STATUS_FORBIDDEN
 		);
--- a/tests/acceptance/features/webUIManageUsersGroups/addGroup.feature
+++ b/tests/acceptance/features/webUIManageUsersGroups/addGroup.feature
@@ -80,3 +80,8 @@ Feature: Add group
   Scenario: Cannot add group with white-space in the name
     When the administrator adds group "whitespace " using the webUI
     Then a notification should be displayed on the webUI with the text "Error creating group: Unable to add group."
+
+  Scenario: Add group with the same name as an existing group
+    Given group "grp1" has been created
+    When the administrator adds group "grp1" using the webUI
+    Then a notification should be displayed on the webUI with the text "Error creating group: Group already exists."
--- a/changelog/unreleased/40165
+++ b/changelog/unreleased/40165
@@ -1,0 +1,7 @@
+Bugfix: Error responses to add group API requests are inconsistent
+
+Some response error messages were contained in a `data` structure. Others were not.
+They have now been made consistent. They are not contained in a `data` structure.
+
+https://github.com/owncloud/core/issues/40164
+https://github.com/owncloud/core/pull/40165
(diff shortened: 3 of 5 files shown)


settings/Controller/GroupsController.php
```diff
@@ -104,6 +104,7 @@ public function create($id) {
 		if ($this->groupManager->groupExists($id)) {
 			return new DataResponse(
 				[
+					'status' => 'error',
```
phil-davis, line 107: Other error responses in `GroupsController.php` have this `'status' => 'error'` in the response. So I added it here to make it consistent.

But I notice that the similar code in UsersController.php `create()` method does not have `'status'` at all.

What to do? Should I remove `'status'` everywhere? Or add it everywhere?

settings/Controller/GroupsController.php
```diff
@@ -121,9 +122,7 @@ public function create($id) {
 		return new DataResponse(
 			[
 				'status' => 'error',
-				'data' => [
-					'message' => (string)$this->l10n->t('Unable to add group.')
-				]
+				'message' => (string)$this->l10n->t('Unable to add group.')
```
phil-davis, line 125: No other messages are inside `'data'`, so I took `'data'` out of here. That makes all the message structures consistent.

tests/acceptance/features/webUIManageUsersGroups/addGroup.feature
```diff
@@ -80,3 +80,8 @@ Feature: Add group
   Scenario: Cannot add group with white-space in the name
     When the administrator adds group "whitespace " using the webUI
     Then a notification should be displayed on the webUI with the text "Error creating group: Unable to add group."
+
+  Scenario: Add group with the same name as an existing group
```
phil-davis, line 84: This adds proper test coverage for this error condition in the webUI, so that we don't get regressions again in future.

settings/Controller/GroupsController.php
jvillafanez: If this is the only place missing the status information, I'd add it here even if it isn't being used now. Adding new information should be easier because we don't need to check additional code that might break.
If there are only one or two places where the status is being used, we might remove it, but we'll have to ensure there is no code using the status info.

phil-davis: @jvillafanez IMO this is good like I have it. As you say, adding extra meta-data in a response is usually easy and less harmful.

I will leave `UsersController.php` untouched - we know that any apps that process those error responses cannot be depending on the `'status'` because it currently does not exist!

phil-davis:
@jvillafanez @IljaN @JanAckermann @JammingBen can someone give a review please - this should also fix a nightly test that fails in user_ldap. So I would like to merge so that I get a new (and hopefully passing) result tonight.

jvillafanez (approved):

PR #40174, opened 2022-06-28
[tests-only] [full-ci] Add the ability to find an object in the project space

v-scharf:
Previously, the search result was only in the personal space
-I added the ability to find objects in project space and shares jail

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -5277,20 +5277,6 @@ trait WebDav {
 			$trimmedEntryNameToSearch = \trim($entryNameToSearch, "/");
 		}
 
-		// topWebDavPath should be something like /remote.php/webdav/ or
-		// /remote.php/dav/files/alice/
-		$topWebDavPath = "/" . $this->getFullDavFilesPath($user) . "/";
-		if (OcisHelper::isTestingOnOcis() && $method === "REPORT") {
-			$spaceId = WebDavHelper::getPersonalSpaceIdForUser(
-				$this->getBaseUrl(),
-				$user,
-				$this->getPasswordForUser($user),
-				$this->getStepLineRef()
-			);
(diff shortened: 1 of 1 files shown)


SagarGi (commented):
LGTM :+1:

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -5307,11 +5296,18 @@ public function findEntryFromPropfindResponse(
...
 			foreach ($multistatusResults as $multistatusResult) {
 				$entryPath = $multistatusResult['value'][0]['value'];
-				$entryName = \str_replace($topWebDavPath, "", $entryPath);
-				$entryName = \rawurldecode($entryName);
-				$entryName = \trim($entryName, "/");
-				if ($trimmedEntryNameToSearch === $entryName) {
-					return $multistatusResult;
+				if (OcisHelper::isTestingOnOcis() && $method === "REPORT") {
+					$entryName = \rawurldecode($entryPath);
+					if (str_ends_with($entryName, $entryNameToSearch)) {
```
phil-davis, line 5301: https://www.php.net/manual/en/function.str-ends-with.php is a new function in PHP 8.0

CI passes in core because this does not get executed when testing on oC10 core.

It will pass in CI in oCIS and reva as long as we make sure to use PHP 8 in the CI. That will be simple to do.

But many developers will have PHP 7.4 installed on their local system because they are doing development of oC10 core (as well as oCIS). So if they want to run a "REPORT" acceptance test scenario against oCIS, they will have to switch to PHP 8. It might be easier to avoid that for now, and write some extra code that does the same as `str_ends_with`

phil-davis: Now I realise that https://github.com/symfony/polyfill-php80 has `str_ends_with()` and that polyfill dependency is included in core. So this will work fine.

phil-davis (approved):

PR #40176, opened 2022-06-29
Restore remember me feature

jvillafanez:
## Description
Restore the "remember me" feature. This will allow auto-login if the session expires and the user has marked the "stay logged in" checkbox in the login page.

Note that this feature relies on all the enabled apps to allow this feature. Some apps such as "files_external" or "encryption" can disable this feature if they're enabled.

## Related Issue
https://github.com/owncloud/core/issues/40170

## How Has This Been Tested?
Currently tested basic scenario:
1. Set a low `session_lifetime` in the config.php (to test it easier)
2. Ensure that the "stay logged in" checkbox in visible in the login page. You might need to disable some apps (files_external, for example).
3. Log in with any user having marked the "stay logged in" checkbox.
4. Close the tab without login out.
5. Wait until the session expires.
6. Access to ownCloud again

At step 6, you're logged in directly.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Notes:
* The feature is expected to work only with the login form. The initial remember-me token is set after a successful login from the login page.
  * Shibboleth, oAuth2 and other authentication methods shouldn't be directly impacted. There are some assumptions here:
    * Those apps should disable the "remember me" feature from their side.
    * An initial remember-me token shouldn't be set from their side (at least it isn't expected). If somehow there is a remember-me token already stored (the user might have access previously through the login form), it's possible that the user is authenticated with such token.
  * Basic authentication from any other place (such as webdav) won't generate a new remember-me token
* After the remember-me token is used, a new one will be generated. The behavior is similar to a one-time token
* Explicitly login out will remove the remember-me token, forcing the user to login again
* Remember-me token is expected to be used only once. After login in with it, regular cookie-based authentication is expected to be used.
* Tokens are hashed in the DB

### DIFF
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -104,7 +104,7 @@ class LoginController extends Controller {
 	public function logout() {
 		$loginToken = $this->request->getCookie('oc_token');
 		if ($loginToken !== null) {
-			$this->config->deleteUserValue($this->userSession->getUser()->getUID(), 'login_token', $loginToken);
+			$this->userSession->clearRememberMeTokensForLoggedInUser();
 		}
 		$this->userSession->logout();
 
@@ -144,7 +144,7 @@ class LoginController extends Controller {
 		}
 
 		// check if user logged in already and has active session
-		if ($this->userSession->isLoggedIn()) {
+		if ($this->userSession->isLoggedIn() || $this->userSession->tryRememberMeLogin($this->request)) {
 			// most likely user manually entered this page, redirect to default url
 			return new RedirectResponse($this->getDefaultUrl());
 		}
@@ -259,11 +259,13 @@ class LoginController extends Controller {
 	 * @param string $password
 	 * @param string $redirect_url
 	 * @param string $timezone
+	 * @param string $remember_login "1" implies we should remember the login; not present (or null)
+	 * implies we shouldn't need to do anything
 	 * @return RedirectResponse
 	 * @throws \OCP\PreConditionNotMetException
 	 * @throws \OC\User\LoginException
 	 */
-	public function tryLogin($user, $password, $redirect_url, $timezone = null) {
+	public function tryLogin($user, $password, $redirect_url, $timezone = null, $remember_login = null) {
 		$originalUser = $user;
 		// TODO: Add all the insane error handling
 		$loginResult = $this->userSession->login($user, $password);
@@ -295,6 +297,9 @@ class LoginController extends Controller {
 		// TODO: remove password checks from above and let the user session handle failures
 		// requires https://github.com/owncloud/core/pull/24616
 		$this->userSession->createSessionToken($this->request, $userObject->getUID(), $user, $password);
+		if ($remember_login) {
+			$this->userSession->setNewRememberMeTokenForLoggedInUser();
+		}
 
 		// User has successfully logged in, now remove the password reset link, when it is available
 		$this->config->deleteUserValue($userObject->getUID(), 'owncloud', 'lostpassword');
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -103,9 +103,7 @@ class LoginController extends Controller {
 	 */
 	public function logout() {
 		$loginToken = $this->request->getCookie('oc_token');
-		if ($loginToken !== null) {
-			$this->userSession->clearRememberMeTokensForLoggedInUser();
-		}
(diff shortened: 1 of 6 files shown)


ho4ho:
Could be related by #26027

phil-davis:
https://drone.owncloud.com/owncloud/core/36192/10/7
some unit tests are failing

jvillafanez:
I think all the authentication apps should have the `<rememberlogin>false</rememberlogin>` in their info.xml file.

With shibboleth, it might be problematic if the user is allowed to login normally and marks the "stay logged in" checkbox. This might be problematic if the shibboleth session expires but you're allowed to login due to the "remember me" feature. Since this is something we can't guarantee, I think it's better to prevent it using the `<rememberlogin>false</rememberlogin>` in the info.xml.

Oauth2 might be an exception here. As far as I know, oauth2 is used mostly for desktop and mobile clients where web access and cookies shouldn't be needed. For the desktop client, it uses the web browser, which could have a "remember me" token. While you have to explicitly authorize the user, you might be trying to authorize the wrong user (which could also happen without the "remember me" feature if you have a session active).
I guess it will depend on the expected usage. If you're expected to have access only to one account, it shouldn't be too problematic because you'll be authorizing the same user anyway, you're just auto-login via the "remember me" token. However, if you have multiple accounts it might be confusing to know which account you are. In addition, I remember some problems with the user switch in the authorization page in some scenarios, and taking into account that the "remember me" is a long life token (expected to last several days) it might become problematic.

jvillafanez:
Tested, working:
1. Ensure the "stay logged in" checkbox is available (you might need to disable files_external app and maybe others)
2. With any user other than the admin, mark the "stay logged in" checkbox and login
3. Close the tab (do not logout explicitly)
4. Wait until the session expires
5. With a different browser and the admin user, enable the files_external app
6. Try to access ownCloud with the same browser used for the user in step 2

You won't access directly to ownCloud and you'll need to login again. The "remember me" feature stopped working when the files_external app got enabled. The cookies used for the "remember me" feature might get deleted when the user tries to access. This isn't important and no change is expected for this.
The explanation is the following: you're trying to login via "remember me", but the feature is disabled, so the login fails. Since the login fails it's assumed something went wrong with the token and it's removed. The token is still present in the DB as valid token, but the cookie is removed from the user's browser due to the failed login.

There is a related scenario that might present a different behavior depending on the user action:
1. Log in with the "stay logged in" active
2. Enable the files_external app (or any other that disables the "remember me" feature)
3. Disable the app enabled in step 2

If the user logs in between steps 2 and 3, the "remember me" cookies will be removed, but this won't happen if he doesn't log in. This means that a different user might be able to login with the "remember me" after step 3 (assuming the token lives long enough).
For now, there is no intention to change this behavior because it isn't critical enough. The user would need to login again.

jvillafanez:
Tested with multiple browsers:
1. Using one browser, login with a user having marked the "stay logged in" checkbox
2. Using a different browser, login with **the same** user having marked the "stay logged in" checkbox
3. Close the tab of one browser and let the session expire
4. Logout the user from the other browser
5. Try to login with the browser from step 3

The expectation is that you'll be auto-login in step 5. The logout will only clear the token associated to the account from the browser1 but the token coming from browser2 will still be valid.
Note any obsolete tokens for the user will also be removed on logout. This will clear any possible token leftover that could exist. It also implies that if no logout happens (the user keeps logging in via "remember me") tokens might accumulate.

jvillafanez:
No issues found using ldap users. The feature works normally

jvillafanez:
Checked with user_shibboleth and it seems to work. However, it might work due to the specific setup.
Without being an expert, it's difficult to predict the behavior and interactions, so there might still be problems. It might be better for the shibboleth app to disable the feature in order to be on the safe side.

In addition, taking into account that the expectation is that all the users go through the shibboleth login, the "remember me" in the login page doesn't make too much sense. Maybe the admin might need a way to bypass shibboleth by using a different url to access ownCloud, but in that case I don't think the admin would want to use the "remember me".

phil-davis:
I added some people for review.

@pmaier1 is this targeted/wanted for 10.11?

JammingBen (approved):
LGTM

phil-davis:
@jvillafanez needs a rebase to resolve the conflict in `core/Controller/LoginController.php`

PR #40182, opened 2022-07-06
Enhanced clustered-instance mode documentation

d7oc:
I just learned today, that some documentation cannot be changed in the https://github.com/owncloud/docs-server repo but rather needs to be changed here in core. So this PR is basically meant to move the changed in https://github.com/owncloud/docs-server/pull/506 to the right place.

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -751,11 +751,15 @@ $CONFIG = [
 
 /**
  * Define ownCloud operation modes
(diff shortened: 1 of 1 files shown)


mmattel:
CI complained about php style which means that you need to run in the shell `make test-php-style-fix` to get it fixed. I have done that plus added some commas for better reading. If CI responds green, I will approve and merge and start the docs part of the change.

mmattel (approved):

PR #40202, opened 2022-07-13
Php cs fixer 3.9.3

phil-davis:
testing php-cs-fixer 3.9.3 only


phil-davis:
see #40211 for php-cs-fixer 3.9.4

Issue #39881, opened 2022-03-14
Run acceptance tests with different databases


phil-davis:
We want to run the acceptance tests with the common supported databases to verify that all is well:

mysql 8.0 #39877 - passed

mariadb 10.2 - that is the default already run in regular CI

mariadb 10.3 #39882 - passed

mariadb 10.4 #39883 - passed

mariadb 10.5 #39884 - passed

mariadb 10.6 #39876 - passed

pgsql 9.4 #39875 - passed

pgsql 10.20 #39874 - passed

oracle #39885 - all scenarios that fail have related issues (see comments in the PR)

kiranparajuli589:
running acceptance tests with this big set of databases in every PR will certainly take much time. @phil-davis is it that crucial to run acceptance tests with all DBs in every build(which targets running acceptance test)? otherwise, how about we run acceptance tests only for `full-ci` or some special text in the pr-title or in nightly builds.

phil-davis:
This issue was just to run CI one-time for each database, to verify that they all currently pass. Sorry for the misunderstanding.

Oracle has some fails - I need to check those and see if there are already issues for those.

phil-davis:
The Oracle fails already have issues raised and are tagged in the scenarios. Everything else passes - good.

phil-davis:
Note: I tried newer MariaDb today:
Mariadb 10.7 #40206 
Mariadb 10.8 #40207

PR #40207, opened 2022-07-14
[tests-only] [full-ci]Test MariaDb 10.8

phil-davis:
Testing only - do not merge. This is to see if all tests pass when using Mariadb 10.8

Similar to what was done for Mariadb 10.3 10.4 10.5 10.6 in issue #39881


phil-davis:
https://drone.owncloud.com/owncloud/core/36314/160/9
Error response from daemon: No such image: owncloudci/nodejs:14

stupid drone - I will restart.

PR #26750, opened 2016-11-30
Adding group display name support in group backends

PVince81:
A step forward for https://github.com/owncloud/core/issues/7918

#### TODOs

- [x] Adjust ShareesController to use the display name
    - [x] requires adjusting the search methods to also return display names
    - [x] adjust search to also search in display name
- [x] Unit tests

@jvillafanez @DeepDiver1975

PR #40229, opened 2022-07-22
Use the displayname of the groups

jvillafanez:
## Description
Use the displayname name of the group instead of the group id in multiple places. Currently checked in the user's list and in the personal settings.

Note that for regular (local) groups, the id and the displayname is the same, so there shouldn't be any noticeable changes.

The LDAP app will rely on this for further functionality (changes for LDAP in progress)

## Related Issue
https://github.com/owncloud/enterprise/issues/5071

## Motivation and Context
The group id must be immutable, as it is the user id. The group displayname should be shown instead, as the displayname can be changed at any point.

## How Has This Been Tested?
Checked along with a specific patch for the user_ldap app.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/40229
+++ b/changelog/unreleased/40229
@@ -1,0 +1,15 @@
+Bugfix: Use group's displayname in the user's profile and user list
+
+Previously, the group id was being used in both the user's profile and
+the user list. This hasn't been important because the local groups have
+matching group id and displayname, the same for ldap groups.
+
+Due to recent changes with the ldap app (version 0.17.0), the group
+id and the displayname could be different, and they'll be different
+by default in the ldap app.
+
+In both the user's profile and the user list, the group id was being used
+instead of the displayname. This is fixed now, and the displayname will
+be used as intended.
+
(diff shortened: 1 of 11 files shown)


phil-davis:
Just a note for reference: basic code to allow groups to have a display name was added in #26750
In practice, in core, the display name has always been the same as the Group Id.

phil-davis (commented):
Looks OK to me, and all tests pass. So we do not have any obvious mixup of group id and group name for core. And this is more progress supporting the ability to have the display name of a group different from the group id.

Needs review by some developer(s).

phil-davis:
What happens about a changelog entry?
In a way, this is a no-external-change refactoring for core. So does it require a changelog?

jvillafanez:
I'll add a bugfix entry taking into account that this is mainly for ldap.

phil-davis (approved):
Changelog looks good. Let's merge this.

PR #40233, opened 2022-07-25
[tests-only] Update README.md

phil-davis:
This is just to check if drone CI is being triggered.


phil-davis:
drone is working in core

PR #40234, opened 2022-07-25
[tests-only] check for test mentioned in unexpected pattern inside the expected failures file

kiranparajuli589:
Signed-off-by: Kiran Parajuli <kiranparajuli589@gmail.com>

## Description
The expected failures linter already notifies for empty link occurrences like:
- `[apiMain/apiSuite]()`
- `[apiMain/apiSuite]`

With this PR:
- check for occurrences like `- apiMain/checksum.feature` and suggest an expected format
- pretty console logs (error, info, success)
- Line numbers in the error message

## Related Issue

- Part of https://github.com/owncloud/QA/issues/716

## How Has This Been Tested?

- :robot:

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/lint-expected-failures.sh
+++ b/tests/acceptance/lint-expected-failures.sh
@@ -1,23 +1,35 @@
 #!/usr/bin/env bash
 
+log_error() {
+    echo -e "\e[31m$1\e[0m"
+}
+
+log_info() {
+    echo -e "\e[34m$1\e[0m"
+}
+
+log_success() {
+    echo -e "\e[32m$1\e[0m"
+}
+
 if [ -n "${EXPECTED_FAILURES_FILE}" ]
 then
 	if [ -f "${EXPECTED_FAILURES_FILE}" ]
 	then
-		echo "Checking expected failures in ${EXPECTED_FAILURES_FILE}"
+		log_info "Checking expected failures in ${EXPECTED_FAILURES_FILE}"
 	else
-		echo "Expected failures file ${EXPECTED_FAILURES_FILE} not found"
-		echo "Check the setting of EXPECTED_FAILURES_FILE environment variable"
+		log_error "Expected failures file ${EXPECTED_FAILURES_FILE} not found"
+		log_error "Check the setting of EXPECTED_FAILURES_FILE environment variable"
 		exit 1
 	fi
 	FINAL_EXIT_STATUS=0
 	# If the last line of the expected-failures file ends without a newline character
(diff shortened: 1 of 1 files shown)


tests/acceptance/lint-expected-failures.sh
```diff
@@ -56,14 +71,20 @@ then
 			# Else ignore the line.
 			if [[ "${INPUT_LINE}" =~ \[([a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+\.feature:[0-9]+)] ]]; then
 				SUITE_SCENARIO_LINE="${BASH_REMATCH[1]}"
+			elif [[ "${INPUT_LINE}" =~ ^[[:space:]]*-[[:space:]][a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+\.feature:[0-9]+[[:space:]]*$ ]]; then
```
phil-davis, line 74: I don't understand how this ever matches. Any line that might match this, will have already matched the `if` above, and o the `elseif` will never happen.

kiranparajuli589: if we have items like `- someSuite/somefeature.feature:520` in the file, this `elif` block will be executed. Otherwise, such lines are bypassed. do we care if such a line appears?

phil-davis: OK - then that is "a good thing" to find. A line like this fails the linter:
```
- apiSuite/download.feature:1234
```

But without the `-` it does not fail the linter. And without a scenario line-number it does not fail the linter. I guess that we want to be able to write general notes in this file, as well as exact references to scenario line-numbers, so it is a bit difficult to know exactly what is "an invalid line".

Let's keep this code.

phil-davis:
PR title is "Check for test mentioned without any connected issue in the the expected failures linter"

The code in this PR does useful stuff, but I don't see how it does what the PR title says.

tests/acceptance/lint-expected-failures.sh
kiranparajuli589: I can write documentation for this lint feature. But where to put it? Inside JankariTech/Technical or owncloud docs?

phil-davis (approved):

phil-davis:
@kiranparajuli589 please organise someone (or you) to bump the commit-id in oCIS and reva so that any existing problems in the expected-failures files get found.

PR #40240, opened 2022-07-26
Init encryption module before calculating unencrypted block-size

IljaN:
## Description
Fixes a bug where the un-encrypted blocksize was calculated before initializing the encryption-module. This yields the wrong size after a file-scan. The issue is only triggered if encryption.use_legacy_encoding is set to true.

## Related Issue

Related test PR: https://github.com/owncloud/encryption/pull/360#issue-1318035322

## Motivation and Context
When restoring encrypted files from backup and doing a ``occ file:scan --repair`` and ``encryption:fix-encrypted-version`` the file was still not readable correctly because the size in oc_filecache is wrong. 

The wrong calculation was introduced via https://github.com/owncloud/core/pull/38249 where the size calculation depends on the encoding mode (binary vs base64). The code assumes that the begin method is called before the size is calculated which is not  the case for the call-site which this PR fixes.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -572,6 +572,8 @@ class Encryption extends Wrapper {
 			return 0;
 		}
 
+		// initialize encryption module to get the correct unencrypted block-size
+		$encryptionModule->begin($this->getFullPath($path), $this->uid, 'r', $header, [], null);
 		$signed = (isset($header['signed']) && $header['signed'] === 'true') ? true : false;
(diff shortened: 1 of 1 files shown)


micbar:
Nice catch!

butonic (approved):

phil-davis:
@IljaN @butonic what to do about a changelog in core?
What can be written about this from the PoV of core?

IljaN:
@phil-davis I'll take care, merged a bit too early. Sorry.

PR #40259, opened 2022-08-01
Add .gitattributes, exclude tests from archive

iamvar:
## Description
Add .gitattributes, exclude tests from archive

## Motivation and Context
less release size

## Types of changes
- [x] Technical debt

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,8 @@
+/.gitattributes export-ignore
+/.github export-ignore
+/.gitignore export-ignore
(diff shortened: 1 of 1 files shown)


phil-davis:
The official release tarballs are built from `Makefile` and other scripts that explicitly select and exclude files that are needed or not. So the official tarballs will not respect `.gitattributes` anyway.

Do we want to add a `.gitattributes` ?

@micbar or anyone - opinions please.

PR #40264, opened 2022-08-03
[full-ci] feat: allow editing of public link shared single files

DeepDiver1975:
## Description
Add option to share a single file via public link with write permissions

## Related Issue
https://github.com/owncloud/enterprise/issues/5274

## How Has This Been Tested?
- acceptance tests in CI
- manual creation of public editable file link, and checking with the API

Note: there is no web UI to make use of this directly - when browsing to the public link, the standard page for viewing and downloading is shown. The purpose of this enhancement is to allow other clients to be able to edit single files in public links.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs-server/issues/572
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #40269, opened 2022-08-05
[full-ci] Test feat editable public link file 20220805

phil-davis:
Test code and adjustments for #40264 

This is rebased to current master so as to get some recent enhancements to acceptance test code that I did in the last few days.

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -322,7 +322,7 @@ class Manager implements IManager {
 		if ($shareNode instanceof \OCP\Files\File) {
 			// Single file shares should never have delete or create permissions
 			$share->setPermissions($share->getPermissions() & ~\OCP\Constants::PERMISSION_DELETE);
-			$share->setPermissions($share->getPermissions() & ~\OCP\Constants::PERMISSION_CREATE);
+#			$share->setPermissions($share->getPermissions() & ~\OCP\Constants::PERMISSION_CREATE);
 		}
 
 		/*
@@ -341,6 +341,9 @@ class Manager implements IManager {
 
 		/* Use share node permission as default $maxPermissions */
 		$maxPermissions = $shareNode->getPermissions();
+		if ($shareNode instanceof \OCP\Files\File) {
+			$maxPermissions |= \OCP\Constants::PERMISSION_CREATE;
+		}
 
 		/* By default, there are no required attributes to be set on a file */
 		$requiredAttributes = $this->newShare()->newAttributes();
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -341,9 +341,6 @@ class Manager implements IManager {
 
 		/* Use share node permission as default $maxPermissions */
 		$maxPermissions = $shareNode->getPermissions();
-		if ($shareNode instanceof \OCP\Files\File) {
-			$maxPermissions |= \OCP\Constants::PERMISSION_CREATE;
-		}
 
 		/* By default, there are no required attributes to be set on a file */
 		$requiredAttributes = $this->newShare()->newAttributes();
@@ -403,6 +400,10 @@ class Manager implements IManager {
 					$maxPermissions |= $shareNode->getPermissions();
 				}
 			}
+		}
+
+		if ($shareNode instanceof \OCP\Files\File) {
+			$maxPermissions |= \OCP\Constants::PERMISSION_CREATE;
 		}
 
 		/**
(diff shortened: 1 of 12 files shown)


tests/acceptance/features/apiSharePublicLink1/createPublicLinkShare.feature
```diff
@@ -106,28 +106,28 @@ Feature: create a public link share
...
 
-  Scenario Outline: Trying to create a new public link share of a file with edit permissions only grants read access using the public WebDAV API
+  Scenario Outline: Create a new public link share of a file with edit permissions
     Given using OCS API version "<ocs_api_version>"
     And user "Alice" has uploaded file with content "Random data" to "/randomfile.txt"
     When user "Alice" creates a public link share using the sharing API with settings
-      | path        | randomfile.txt |
-      | permissions | all            |
+      | path        | randomfile.txt            |
+      | permissions | read,update,create,delete |
```
phil-davis, line 114: @DeepDiver1975 In order to get "read,update" (permissions 3) below, I have to specify "read,update,create,delete" (permissions 15) in the request. That seems a bit odd.

But once I have the public link share with permissions 3 (which is what comes in this test scenario), then the test scneario can download ffrom and upload to the single file - so that is " a good thing"

tests/acceptance/features/bootstrap/PublicWebDavContext.php
```diff
@@ -1224,6 +1224,56 @@ public function publiclyUploadingShouldWork(string $publicWebDAVAPIVersion):void
...
+	 *
+	 * @return void
+	 * @throws Exception
+	 */
+	public function publiclyUploadingToPublicLinkSharedFileShouldWork(string $publicWebDAVAPIVersion):void {
+		$content = "test $publicWebDAVAPIVersion";
+
+		if ($publicWebDAVAPIVersion === "new") {
+			$techPreviewHadToBeEnabled = $this->occContext->enableDAVTechPreview();
+			$path = $this->featureContext->getLastPublicSharePath();
```
phil-davis, line 1240: Note: for the "new" public webDAV API the filename (=path) has to be provided after the token, making a URL like:
`GET http://172.17.0.1:8080/remote.php/dav/public-files/UUoHGttLWK8p1UM/randomfile.txt`

tests/acceptance/features/bootstrap/PublicWebDavContext.php
```diff
@@ -1224,6 +1224,56 @@ public function publiclyUploadingShouldWork(string $publicWebDAVAPIVersion):void
...
+		} else {
+			$techPreviewHadToBeEnabled = false;
+			$path = "";
```
phil-davis, line 1243: But for the "old" public WebDAV API the filename is not mentioned:
```
AUTH: SgdcsbofcZ3jFGk:
REQUEST: PUT http://172.17.0.1:8080/public.php/webdav
```

The token is provided as the authentication, and you simply PUT to `public.php/webdav` and the server stuffs the data in the body of the PUT into the file associated with the single-file public link share.

tests/acceptance/features/bootstrap/PublicWebDavContext.php
```diff
@@ -1447,6 +1497,10 @@ public function publicUploadContent(
 			\array_map('rawurlencode', \explode('/', $filename))
 		);
 		$url .= \ltrim($filename, '/');
+		// Trim any "/" from the end. For example, if we are putting content to a
+		// single file that has been shared with a link, then the URL should end
+		// with the link token and no "/" at the end.
+		$url = \rtrim($url, "/");
```
phil-davis, line 1503: This is a double-check that we don't end up with an unwanted (unneeded) "/" on the end of things like `public.php/webdav`

lib/private/Share20/Manager.php
```diff
@@ -341,9 +341,6 @@ protected function validatePermissions(IShare $share) {
 
 		/* Use share node permission as default $maxPermissions */
 		$maxPermissions = $shareNode->getPermissions();
-		if ($shareNode instanceof \OCP\Files\File) {
-			$maxPermissions |= \OCP\Constants::PERMISSION_CREATE;
-		}
```
phil-davis: @DeepDiver1975 I moved this code block, because otherwise it caused problems with some ordinary sharing of a single file, and there were test scenarios failing.
`$maxPermissions` gets set to `0` down at line 374, which was overriding what this code block did.

phil-davis:
Note: the existing acceptance test code has different ways of remembering the details of public links created by an API request and with the webUI.

API requests provide lots of detail in the response, so the tests can save all of it, and thus get back whatever identifying information they need in later `Then` steps.

But when creating with the webUI, we don't have easy access to the API request-response that the webUI makes. So we save what data we can gather from the UI itself.

Much of the test code has 2 paths for this stuff - test steps that work with the saved API response data, and other test steps that work with the saved webUI data. I patched things up a bit so that some functions are smart enough to look in both places for the information they need. Actually it would be good to refactor this test code so that it works more cleanly, but that is a job for another day. I don't want to add a big refactoring to this PR! I will raise a separate issue for refactoring.

phil-davis:
This has been cherry-picked into #40264

PR #40274, opened 2022-08-09
[tests-only][full-ci] Remove space related tests with /shares for favorite

amrita-shrestha:
## Description
The `/Shares` related for the `spaces` WebDAV is removed from core since this is to be implemented and run on ocis.

## Related Issue
https://github.com/owncloud/ocis/issues/4154#issuecomment-1206026045

## How Has This Been Tested?

- Locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiFavorites/favorites.feature
+++ b/tests/acceptance/features/apiFavorites/favorites.feature
@@ -146,7 +146,7 @@ Feature: favorite
       | dav_version |
       | spaces      |
 
-  @files_sharing-app-required @issue-ocis-2968
+  @files_sharing-app-required @notToImplementOnOCIS
   Scenario Outline: moving a favorite file out of a share keeps favorite state
     Given using <dav_version> DAV path
     And user "Brian" has been created with default attributes and without skeleton files
@@ -161,11 +161,6 @@ Feature: favorite
       | dav_version |
       | old         |
       | new         |
-
-    @skipOnOcV10 @personalSpace
-    Examples:
-      | dav_version |
-      | spaces      |
 
   @issue-33840 @skipOnOcV10
   Scenario Outline: Get favorited elements and limit count of entries
--- a/tests/acceptance/features/apiFavorites/favoritesSharingToShares.feature
+++ b/tests/acceptance/features/apiFavorites/favoritesSharingToShares.feature
@@ -8,7 +8,7 @@ Feature: favorite
     And user "Alice" has created folder "/PARENT"
     And user "Alice" has uploaded file with content "some data" to "/PARENT/parent.txt"
 
-  @issue-ocis-2968
+
   Scenario Outline: favorite a file inside of a received share
     Given using <dav_version> DAV path
     And user "Brian" has been created with default attributes and without skeleton files
@@ -22,12 +22,7 @@ Feature: favorite
       | old         |
       | new         |
 
-    @skipOnOcV10 @personalSpace
-    Examples:
-      | dav_version |
-      | spaces      |
 
-  @issue-ocis-2968
   Scenario Outline: favorite a folder inside of a received share
     Given using <dav_version> DAV path
     And user "Brian" has been created with default attributes and without skeleton files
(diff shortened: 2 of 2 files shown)


tests/acceptance/features/apiFavorites/favorites.feature
```diff
@@ -162,11 +162,6 @@ Feature: favorite
       | old         |
       | new         |
 
-    @skipOnOcV10 @personalSpace
-    Examples:
-      | dav_version |
-      | spaces      |
-
```
SagarGi, line 169: i am confused the above scenarios does not have any `/Shares` step and also user has not accepted any shares or it is auto accepted by default. @amrita-shrestha ?

SagarGi (commented):
other LGTM :+1:

tests/acceptance/features/apiFavorites/favorites.feature
kiranparajuli589: but yet, this scenario will not be valid for the spaces DAV version. the `when` step will fail if we run like this because the personal space will be looked into for the resources to move. so I think the move is necessary for this scenario too.

kiranparajuli589: plus the scenario should be skipped for `ocis`

tests/acceptance/features/apiFavorites/favorites.feature
```diff
@@ -146,7 +146,7 @@ Feature: favorite
       | dav_version |
       | spaces      |
 
-  @files_sharing-app-required @issue-ocis-2968
+  @files_sharing-app-required @skipOnOcis
```
SwikritiT, line 149:
```suggestion
  @files_sharing-app-required @notToImplementOnOCIS
```
Can we use this tag for this test since it's a test dealing with root share?

amrita-shrestha: Doesn't both tags do same thing?

phil-davis: > Don't both tags do the same thing?

Yes- it is just naming. If a test scenario is testing a feature that is "never" going to be implemented on oCIS, then `notToImplementOnOCIS` makes that clear. `skipOnOcis` can be used for "other reasons". Probably we should have the reason mentioned every time that we use `skipOnOcis`. We should discuss what we have been doing so far with these tag names, and what should be the convention for the future.

tests/acceptance/features/apiFavorites/favorites.feature
```diff
@@ -146,7 +146,7 @@ Feature: favorite
       | dav_version |
       | spaces      |
 
-  @files_sharing-app-required @issue-ocis-2968
+  @files_sharing-app-required @notToImplementOnOCIS
   Scenario Outline: moving a favorite file out of a share keeps favorite state
```
kiranparajuli589, line 150: the move operations out of share is only not possible with `spaces` DAV. but, IMO, this is possible in oCIS while using `old` or `new` DAV versions. So, just removing the spaces example would be enough here imo

SwikritiT: @kiranparajuli589 I think the scenario is for `root` sharing and that shouldn't be possible at ocis at all. So I think that tag is needed there.

tests/acceptance/features/apiFavorites/favorites.feature
```diff
@@ -146,7 +146,7 @@ Feature: favorite
       | dav_version |
       | spaces      |
 
-  @files_sharing-app-required @issue-ocis-2968
+  @files_sharing-app-required
```
SwikritiT, line 149:
```suggestion
  @files_sharing-app-required @notToImplementOnOCIS
```
@amrita-shrestha I think this is still needed here as it's for root sharing for shares https://github.com/owncloud/core/blob/b973c38384421e15ea8eb902c6f7d8e255855bdb/tests/acceptance/features/apiFavorites/favoritesSharingToShares.feature#L55 here's another scenario that should work in case of OCIS

tests/acceptance/features/apiFavorites/favorites.feature
kiranparajuli589: just checked the move operation using old and new DAV versions. Found that, it is not possible to move files out of the share. tag is necessary, change looks reasonable :+1: .

kiranparajuli589 (commented):
lgtm :+1:

saw-jan (commented):
LGTM

SagarGi (commented):
lgtm :+1:

phil-davis (approved):

Issue #40119, opened 2022-06-03
Case sensitive usernames when logging in with an app password via webdav


T0mWz:
This issue is previously reported (here https://github.com/owncloud/core/issues/29708), but we are now encountering it ourselves too. The way a webdav password is generated and how it is then accepted is not always consistent. 

### Steps to reproduce
1. Create user with a uppercase character
2. Generate a App passwords (Security tab by your personal settings) with a completely lower case username and a webdav token with the normal username which contains one of more uppercase characters
3. try to login with both

### Expected behaviour
Expected behaviour should be that both accounts should work and the casing of username doesn't makes a difference.

### Actual behaviour
Depends how the webdav token is generated, how the application response;
```
tomw@M0421014 ~ % curl -u 'u3342109@exchange-example.edu:app-token-is-here' https://miskatonic.data.surfsara.nl/remote.php/nonshib-webdav
This is the WebDAV interface. It can only be accessed by WebDAV clients such as the ownCloud desktop sync client.%
tomw@M0421014 ~ % curl -u 'U3342109@exchange-example.edu:app-token-is-here' https://miskatonic.data.surfsara.nl/remote.php/nonshib-webdav
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:exception>Sabre\DAV\Exception\NotAuthenticated</s:exception>
  <s:message>Username or password was incorrect, Username or password was incorrect, No 'Authorization: Bearer' header found. Either the client didn't send one, or the server is mis-configured</s:message>
</d:error>
```

```
tomw@M0421014 ~ % curl -u 'U3342109@exchange-example.edu:app-token-is-here' https://miskatonic.data.surfsara.nl/remote.php/nonshib-webdav
This is the WebDAV interface. It can only be accessed by WebDAV clients such as the ownCloud desktop sync client.%
tomw@M0421014 ~ % curl -u 'u3342109@exchange-example.edu:app-token-is-here' https://miskatonic.data.surfsara.nl/remote.php/nonshib-webdav
This is the WebDAV interface. It can only be accessed by WebDAV clients such as the ownCloud desktop sync client.%
```

### Server configuration
**Operating system**:
CentOS Linux release 7.9.2009 (Core)

**Web server:**
Apache/2.4.6 (CentOS)

**Database:**
MariaDB 10.3

**PHP version:**
PHP 7.4.29

**ownCloud version:** 
[...]

Issue #29708, opened 2017-11-29
Case sensitive usernames when logging in with an app password via webdav


mdusher:
### Steps to reproduce
1. Create an app password
2. Try and login with your username in a different case as to what is stored in ownCloud

### Expected behaviour
The login should be case insensitive for the login.

### Actual behaviour
The login is rejected because it does not match what is stored in ownCloud

### Server configuration
**Operating system**: Redhat 7

**Web server:** Apache 2.2.15

**Database:** MariaDB 10.0.27 with Galera 25.3.18

**PHP version:** 7.0.23

**ownCloud version:** 10.0.3

**Updated from an older ownCloud or fresh install:** Updated from 8.2.11

**Where did you install ownCloud from:** https://owncloud.org/install/#edition

**Signing status (ownCloud 9.0 and above):** 
```
Integrity checker has been disabled. Integrity cannot be verified.
```

**The content of config/config.php:**

```
{
    "system": {
        "instanceid": "5230042dc1897",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": {
            "0": "cloudstor.aarnet.edu.au"
        },
        "datadirectory": "\/cloudstor\/data\/owncloud\/data",
        "version": "10.0.3.3",
        "dbtype": "mysql",
        "dbname": "owncloud",
        "dbhost": "127.0.0.1:3306",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "dbtableprefix": "",
        "installed": true,
        "operation.mode": "clustered-instance",
        "default_language": "en_GB",
        "defaultapp": "files",
        "knowledgebaseenabled": true,
        "enable_avatars": false,
        "allow_user_to_change_display_name": false,
        "session_lifetime": 86400,
        "session_keepalive": true,
        "token_auth_enforced": false,
        "mail_domain": "aarnet.edu.au",
        "mail_from_address": "cloudstor-noreply",
        "mail_smtpmode": "php",
        "overwriteprotocol": "https",
        "overwrite.cli.url": "https:\/\/cloudstor.aarnet.edu.au\/plus",
        "htaccess.RewriteBase": "\/plus",
        "trashbin_retention_obligation": "30, auto, auto",
        "appcodechecker": false,
        "updatechecker": false,
        "has_internet_connection": true,
        "check_for_working_webdav": false,
[...]

PR #40280, opened 2022-08-10
Fix case sensitivity for app passwords/tokens

pako81:
## Description

Fix case sensitivity for app passwords/tokens

## Related Issue

- https://github.com/owncloud/core/issues/40119
- https://github.com/owncloud/core/issues/29708

## Motivation and Context

Currently, app passwords/tokens are case sensitive. That is, login will be prevented if the entered username has different case as the app password/token's username (which comes from the user's session).

## How Has This Been Tested?

1. Login as user `User1` and create an app password/token --> username will be automatically set to `User1` and saved as the `login_name` in the `oc_authtoken` table. 
2. Try to authenticate by using i.e. the ownCloud desktop client with the newly generated app password/token by using `user1` as username (note the different casing):

Before this fix --> login failed with exception https://github.com/owncloud/core/blob/master/lib/private/User/Session.php#L927

After this fix --> login is correctly allowed, which is consistent with the "normal" login via username/password where usernames are case-insensitive.
 
## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ ownCloud admins and users.
 Summary
 -------
 
+* Bugfix - Fix case sensitivity for app passwords/tokens - [#40280](https://github.com/owncloud/core/pull/40280)
 * Bugfix - Trigger the right event when the filecache is updated: [#39844](https://github.com/owncloud/core/pull/39844)
 * Bugfix - Replace userid with username in login form: [#39870](https://github.com/owncloud/core/pull/39870)
 * Bugfix - Quota can be exceeded by user: [#40140](https://github.com/owncloud/core/issues/40140)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ ownCloud admins and users.
 Summary
 -------
 
+* Bugfix - Fix case sensitivity for app passwords/tokens - [#40280](https://github.com/owncloud/core/pull/40280)
```
phil-davis, line 11: We don't directly edit this file. See PR #40281 for a suggested changelog entry. I also made a test scenario there.

pako81:
Closing this in favour of https://github.com/owncloud/core/pull/40281

Issue #40286, opened 2022-08-12
Versions are not deleted when the file is deleted from the trashbin


jvillafanez:
### Steps to reproduce
1. Create "test1" folder
2. Upload "file.txt" inside the "test1" folder
3. Upload multiple versions of "file.txt"
4. Delete "test1" folder
5. In the "deleted files" section of the files page, enter the "test1" folder and permanently delete the "file.txt" file

### Expected behaviour
All versions of the "file.txt" file should have been deleted when the file was permanently deleted from the trashbin

### Actual behaviour
Versions are still around taking up space. In particular, versions are still in "[user-folder]/files_trashbin/versions" and also in the DB. Versions are likely inaccessible from the web UI because the file itself was removed.

### Server configuration
**Operating system**:

**Web server:** apache 2.4

**Database:** mysql

**PHP version:** 7.4

**ownCloud version:** 10.11 pre-alpha

**Updated from an older ownCloud or fresh install:** fresh install

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results into https://gist.github.com/ and puth the link here.
```

**The content of config/config.php:**

```
Log in to the web-UI with an administrator account and click on
'admin' -> 'Generate Config Report' -> 'Download ownCloud config report'
This report includes the config.php settings, the list of activated apps
and other details in a well sanitized form.

or 

If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

*ATTENTION:* Do not post your config.php file in public as is. Please use one of the above
methods whenever possible. Both, the generated reports from the web-ui and from occ config:list
consistently remove sensitive data. You still may want to review the report before sending.
If done manually then it is critical for your own privacy to dilligently
remove *all* host names, passwords, usernames, salts and other credentials before posting.
You should assume that attackers find such information and will use them against your systems.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder.
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/...

#### LDAP configuration (delete this part if not used)

```
With access to your command line run e.g.:
sudo -u www-data php occ ldap:show-config
[...]

PR #40291, opened 2022-08-16
Restore and delete operations in trashbin will affect versions

jvillafanez:
## Description
Versions of files were left in the trashbin even though the associated file was either restored or deleted. This causes a performance loss for large scenarios because we have to deal with too many versions while restoring or deleting files from the trashbin.

## Related Issue
https://github.com/owncloud/core/issues/40286

## Motivation and Context
Versions keep piling up without any chance to clean them up.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_trashbin/lib/Trashbin.php
+++ b/apps/files_trashbin/lib/Trashbin.php
@@ -513,14 +513,21 @@ class Trashbin {
 	 * @throws LockedException
 	 * @throws StorageNotAvailableException
 	 */
-	public static function restore($file, $filename, $timestamp, $targetLocation = null) {
+	public static function restore($filename, $timestamp, $targetLocation = null) {
 		$user = User::getUser();
 		$view = new View('/' . $user);
+		$nameOfFile = \basename($filename);
+		$dirOfFile = \dirname($filename);
+
+		$filenameWithTimestamp = $filename;
+		if ($dirOfFile === '/' || $dirOfFile === '.') {
+			$filenameWithTimestamp .= ".d{$timestamp}";
+		}
 
 		if ($targetLocation === null) {
 			$location = '';
 			if ($timestamp) {
-				$location = self::getLocation($user, $filename, $timestamp);
+				$location = self::getLocation($user, $nameOfFile, $timestamp);
 				if ($location === false) {
 					\OCP\Util::writeLog('files_trashbin', 'Original location of file ' . $filename .
 						' not found in database, hence restoring into user\'s root instead', \OCP\Util::DEBUG);
@@ -536,11 +543,11 @@ class Trashbin {
 			}
 
 			// we need a  extension in case a file/dir with the same name already exists
-			$uniqueFilename = self::getUniqueFilename($location, $filename, $view);
+			$uniqueFilename = self::getUniqueFilename($location, $nameOfFile, $view);
 			$targetLocation = $location . '/' . $uniqueFilename;
 		}
 
-		$source = Filesystem::normalizePath('files_trashbin/files/' . $file);
+		$source = Filesystem::normalizePath('files_trashbin/files/' . $filenameWithTimestamp);
 		$target = Filesystem::normalizePath('files/' . $targetLocation);
 		if (!$view->file_exists($source)) {
 			return false;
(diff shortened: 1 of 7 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>apiProxySmoke-8-6-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/36579/185

jvillafanez:
I don't think I can add unit tests to cover these cases. The files_trashbin app is handling the restoration and deletion of the versions (which is wrong, at least the way it's being done), and core doesn't seem to provide a way to access the versions because they seem to be completely handled by the files_versions app.
If I want to restore a file with multiple versions, I have to know where those versions will end up. If the file is deleted along with the versions, I'd need a way to verify that those versions aren't present anywhere.

@phil-davis is it possible to include some acceptance tests to cover these scenarios? Maybe we have better luck using API calls.

phil-davis:
> is it possible to include some acceptance tests to cover these scenarios?

We can try - but I suspect that after restoring or deleting a file from the trashbin, there will be no way to externally access the versions that might be left behind in the trashbin. First we need to try on a current core system to access such "lost" trashed versions with the API. If we can do that, then we can make a test scenario that accesses those, then reverse the checks of those steps (so that they expect the access attempts to fail), then confirm that the test fails on current master, and passes with the code in this PR.

phil-davis:
A test was added in PR https://github.com/owncloud/core/pull/40322 that confirms that versions are restored after a file is deleted and then restored from the trashbin. That is "a good thing" and already works without this PR.

We did not find a way to have an end-to-end automated test that verifies that, when a file is deleted from the trashbin, the versions are also deleted from storage and from the database. The existence of versions of a trashbin file seems to be hidden from access via the external API - a user cannot see the versions that might be saved in the trashbin along with a file in the trashbin. So end-to-end tests cannot verify that versions exist in the trashbin, and cannot then verify that the versions are "really" gone when the file is deleted from the trashbin.

This is a common challege for any end-to-end test that just accesses an API. On any server that stores persistent data, the data can be verified to exist by accessing it from the API, and checking that the response has the expected data. Then use the API to delete the data, and verify that a GET via the API now returns a "not found". But the server could have only removed the data reference from whatever "database" it uses to know what data items exist. The data might still exist and be taking up space in the persistent storage - there is no way for any "ordinary" end-to-test to know that.

phil-davis:
IMO this code is good-to-go. It will need manual verification during the release testing.

When do we want this released? @pmaier1 @jnweiger

phil-davis:
@jvillafanez how to existing "stranded" version files get cleaned up from the trashbin after someone upgrades to this code?

jvillafanez:
> @jvillafanez how to existing "stranded" version files get cleaned up from the trashbin after someone upgrades to this code?

I think that those stranded versions come from files within trashed folders. If I remember correctly, deleting or restoring that folder (the one that shows in the root of the trashbin) should fix the issue.

phil-davis:
@jvillafanez lots of code has been merged-back to master with the release of 10.11.0 yesterday.
Please rebase so that we get up-to-date CI. And request people to review. IMO this can move forward now.

pako81:
Can we get this reviewed and eventually merged, please? Thx.

pako81:
can we move this forward?

mrow4a (approved):
I just hope these function were not "forcefully" used in other apps, as with this change they would brake. just make sure to do quick regex. but change itself looks very legit

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -502,25 +502,47 @@ private static function copy(View $view, $source, $target) {
...
+	 * will be. For example, $targetLocation = "folder1/restored.txt" will
+	 * put the restored file ("file.txt.d123456") there. This is basically a rename.
+	 * In particular, restoring "bigFolder.d12345" to the target location
+	 * "largeContent/toobig" WON'T create "largeContent/toobig/bigFolder".
+	 * If null is provided, we'll try to put the file in it's previous location
+	 * if it still exists, otherwise the file will be restored in the root folder
+	 * (with an unique name if needed) as fallback. Note that the previous location
+	 * is only known for the top level directory in the root's trashbin, such as
+	 * "file1.txt.d12345" or "topFolder.d98765", but not "folder.d12345/file.txt"
+	 * or "folder.d98765/folder2"
```
mrow4a, line 519: this behaviour was also previous behaviour, or?

jvillafanez: it was the previous behaviour, I just clarified the expectations for the parameters since they weren't clear

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -513,14 +513,21 @@ private static function copy(View $view, $source, $target) {
 	 * @throws LockedException
 	 * @throws StorageNotAvailableException
 	 */
-	public static function restore($file, $filename, $timestamp, $targetLocation = null) {
+	public static function restore($filename, $timestamp, $targetLocation = null) {
```
phil-davis: Note: this is an interface change. See issue https://github.com/owncloud/ransomware_protection/issues/221 and PR https://github.com/owncloud/ransomware_protection/pull/306

We need to adjust the call that is in the ransomware_protection app.

PR #40307, opened 2022-08-22
[tests-only]Remove unnecessary test from core

SwikritiT:
## Description
The test is tagged with `@notToImplementOnOCIS` and it's skipped in `oc10` as well since it's a test related to `TUS` protocol.

## Motivation and Context
According to the discussion here in https://github.com/owncloud/ocis/issues/1141 seems like this test is not relevant so I'm removing it 

## How Has This Been Tested?

- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/apiWebdavUploadTUS/uploadFile.feature
+++ b/tests/acceptance/features/apiWebdavUploadTUS/uploadFile.feature
@@ -216,19 +216,3 @@ Feature: upload file
     | spaces      | "filewithLF-and-CR\r\n" | ZmlsZXdpdGhMRi1hbmQtQ1INCgo= |
     | spaces      | "folder/file"           | Zm9sZGVyL2ZpbGU=             |
     | spaces      | "my\\file"              | bXkMaWxl                     |
-
-  @notToImplementOnOCIS @issue-ocis-1141
-  Scenario Outline: upload a file using the resource URL of another user
-    Given using <dav_version> DAV path
-    And user "Brian" has been created with default attributes and without skeleton files
-    And user "Alice" has created a new TUS resource on the WebDAV API with these headers:
-      | Upload-Length   | 10                        |
-      #    dGV4dEZpbGUudHh0 is the base64 encode of textFile.txt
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/apiWebdavUploadTUS/uploadFile.feature
```diff
@@ -216,19 +216,3 @@ Feature: upload file
...
-      | Upload-Length   | 10                        |
-      #    dGV4dEZpbGUudHh0 is the base64 encode of textFile.txt
-      | Upload-Metadata | filename dGV4dEZpbGUudHh0 |
-    When user "Brian" sends a chunk to the last created TUS Location with offset "0" and data "12345" using the WebDAV API
-    Then the HTTP status code should be "403"
-    And as "Alice" file "/textFile.txt" should not exist
-    Examples:
-      | dav_version |
-      | old         |
-      | new         |
```
SwikritiT, line 234: This is a test related to `TUS` protocol and if it's `notToImplement` on ocis I think it can be removed as `oc10` doesn't support TUS. 

@phil-davis or do we want to keep it incase `tus` is implemented in `oc10`? Will it be implemented here?

phil-davis: Let's remove it.
@SwikritiT we need to have some TUS "security" tests like this in the local API tests in oCIS. There just need to be a few scenarios (maybe only 1 needed?) that check that users cannot interfere with other user's TUS uploads (they get 403 or some other sensible 4xxx status if they try) Please make an issue in oCIS for such test(s) so that we do not forget.

phil-davis (approved):

tests/acceptance/features/apiWebdavUploadTUS/uploadFile.feature
phil-davis: Now I read the linked discussion - actually we will not enforce user authentication on that TUS "temporary" endpoint. It has a "token" that can be used by whoever knows the token.

PR #40315, opened 2022-08-25
Adjust OC_VersionString to more closely match the git tag

phil-davis:
## Description
The git tag is `v10.11.0-beta.1` which matches the pattern at item 9 of https://semver.org/ - good.

IMO we can keep the value in `version.php` in this format also. It is still "human-readable" and also it will get parsed correctly by any automated code that looks at it.

## How Has This Been Tested?
N/A

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/version.php
+++ b/version.php
@@ -27,8 +27,8 @@
 // when updating major/minor version number.
 $OC_Version = [10, 11, 0, 3];
 
-// The human readable string
-$OC_VersionString = '10.11.0 beta 1';
+// The human-readable string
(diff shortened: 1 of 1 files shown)


jnweiger (approved):
thnaks.
I had spelled the vesions with whitespace so far, to make it less technical and more suited for human consumption.
It only affects prereleases, so no objectsions to use semver string here too.

PR #40328, opened 2022-09-02
[tests-only][full-ci]Refactor code to reuse code in `ocis`

SagarGi:
### Description
This PR adds some code adjustment (in core) so that the code can be reused in ocis (If any new implementation is done). Since as of now we are duplicating the same code in core as well as in ocis.

We have a `favourite.feature` feature in ocis where we have been duplicating code so for reference i have tried to reduce the code in `ocis` for the feature so that it uses code (resource) already in core (for reducing code duplication). Follow this PR https://github.com/owncloud/ocis/pull/4492

## Related Issue
No related Issue So far since this is a sample PR (After approval) a separate issue is made for refactoring and code reuse from core for ocis implementation.

## How Has This Been Tested?

- Locally
- CI

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/TestHelpers/WebDavHelper.php
+++ b/tests/TestHelpers/WebDavHelper.php
@@ -40,7 +40,7 @@ class WebDavHelper {
 	public const DAV_VERSION_OLD = 1;
 	public const DAV_VERSION_NEW = 2;
 	public const DAV_VERSION_SPACES = 3;
-	public static $SPACE_ID_FROM_OCIS = null;
+	public static $SPACE_ID_FROM_OCIS = '';
 
 	/**
 	 * @var array of users with their different spaces ids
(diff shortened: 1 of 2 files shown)


SagarGi:
NOTE: We can do like this instead of creating extra helper function to reuse then code from ocis. If feasible then we can go ahead and refactor where code duplication has been done.

SwikritiT (commented):
LGTM.

tests/TestHelpers/WebDavHelper.php
```diff
@@ -40,6 +40,7 @@ class WebDavHelper {
 	public const DAV_VERSION_OLD = 1;
 	public const DAV_VERSION_NEW = 2;
 	public const DAV_VERSION_SPACES = 3;
+	public static $SPACE_ID_FROM_OCIS = '';
```
phil-davis, line 43: Because this is public, this means that we can set the value of `$SPACE_ID_FROM_OCIS` directly from the test code whenever we want - usually when the step is doing something for a Project Space.

phil-davis (approved):
Looks OK - let's merge this and we can see what the refactoring of oCIS local API tests looks like. We can adjust again if we need to...

tests/TestHelpers/WebDavHelper.php
SagarGi: @phil-davis yes an example is here https://github.com/owncloud/ocis/pull/4492 in this PR

PR #40334, opened 2022-09-05
Move view versions author setting (config.sample.php)

mmattel:
This is just a small change and moved the config setting `file_storage.save_version_author` to the correct location of Files Versions. No content change, move only.

@jnweiger would be great to get this into 10.11 (if possible)

Referencing: [Add recent core changes](https://github.com/owncloud/docs-server/pull/600#top)

I will do an config-to-docs run post merging.

@pmaier1 fyi

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -705,6 +705,13 @@ $CONFIG = [
 'versions_retention_obligation' => 'auto',
 
 /**
+ * Save and display the author of each version of uploaded and edited files.
+ *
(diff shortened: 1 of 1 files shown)


phil-davis (approved):
LGTM - I will make a PR to `release-10.11.0`
I don't see a problem with including this among other things that are to be put into the release.

phil-davis:
Cherry-picked to `relese-10.11.0` in PR #40340

PR #40333, opened 2022-09-05
[full-ci] [10.11.0] Allow editing of public link shared single files 10.11

phil-davis:
## Description
This is the same as #40317 but targeted at `release-10.11.0` branch.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #40317, opened 2022-08-25
Allow editing of public link shared single files

phil-davis:
## Description
This feature was initially in PR #40264 which was merged to master. It was reverted in PR #40312 because it is not yet for release.
PR #40283 also made some minor adjustments to skip tags on new/changed acceptance tests.

This PR is a cherry-pick of the commits from #40264 and #40283 - I will leave the PR in draft mode until after 10.11.0 release.

## Related Issue
https://github.com/owncloud/enterprise/issues/5274

## How Has This Been Tested?
CI

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #40342, opened 2022-09-06
[full-ci] [10.11.0] Allow editing of public link shared single files 10.11

phil-davis:
Replacement PR for #40333


phil-davis:
@jnweiger @mrow4a I have adjusted the acceptance tests and JS tests - I only had to change things that were due to the name changes of some DOM elements, so that was good. I think that CI will pass now.

Please review, and merge whenever you like.

CC @pmaier1

jnweiger (approved):
Looks very similar to the earlier #40333 and #40317 -- OK.

PR #40324, opened 2022-08-31
[tests-only] [full-ci] Try tests with PHP 8.0

phil-davis:
a) CI passed for PHP 7.4 with Ubuntu 20.04
b) I have just pushed the commits to run tests with PHP 8.0

https://php.watch/versions/8.0

PR #40354, opened 2022-09-12
PHP 8.0 with DBAL 3

phil-davis:
On top of #40324 

- change `doctrine/dbal` from major version 2 to 3. https://github.com/doctrine/dbal/releases/tag/3.4.4


phil-davis:
Note: https://www.doctrine-project.org/2020/11/17/dbal-3.0.0.html

"We intend to support DBAL 2.12 a while longer, including support for the upcoming PHP 8.0, so that there is no rush for you to upgrade to DBAL 3."

"Deprecated methods in DBAL 2.12 are tagged with the @deprecated doc-block and static analysis tools such as Psalm, PHPStan and Phan can already help you detect using or calling this deprecated code in your application."

So there should not be a need to do this major version upgrade until later. PHP 8 support does not depend on it.

PR #40045, opened 2022-05-04
feat: previews are physically deleted from disk and db in an async ma…

DeepDiver1975:
…nner

## ToDos
- [x] background job for automation
- [x] occ command for cron on bigger installations
- [x] define occ command printout
- [x] add tests for cleanup procedure
- [x] implement cleanup logic for objectstore - requires different sql queries
- [x] invalidate previews if a file is upldated
- [x] kills gallery because of https://github.com/owncloud/gallery/blob/0d980ea3fd7858f27e9096c9f2c0c71e1fb160ed/preview/preview.php#L105 :vomiting_face: :vomiting_face: :vomiting_face: :vomiting_face: :vomiting_face: 

## Description
All preview hooks are gone now. Previews are cleaned up on disk and db in a background job (alternative occ command)

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/5164

### Scenario I
- disable trashbin
- upload file which can have previews (e.g. image)
- wait for the preview to be generated
- note the fileid (e.g.query filecache)
- look up the thumbnails for this file (data/$user/thumbnails/$fileid)
- delete the file
- make sure the thumbnail is still there
- run occ p:c --all
- make sure thumbnail got removed from disk and failcache

### Scenario II
- same as Scenario II - but with trashbin enabled
- file needs to be deleted from trashbin

### Scenario III
- create new test file
- enter some text via texteditor
- see the text on the preview
- edit the file again aka change the contents
- make sure the preview holds the new file contents

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #40361, opened 2022-09-12
fix: previews:cleanup with chunk size argument

DeepDiver1975:
## Description
Fix previews:cleanup with chunk size argument

## Related Issue
- https://github.com/owncloud/core/pull/40045#issuecomment-1243721634

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/core/Command/Previews/Cleanup.php
+++ b/core/Command/Previews/Cleanup.php
@@ -58,7 +58,7 @@ class Cleanup extends Base {
 		$chunk_size = $input->getArgument('chunk_size');
 
 		$pc = new PreviewCleanup($this->connection);
-		$count = $pc->process($all, $chunk_size, static function ($userId, $name, $action) use ($output) {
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
no change log entry - belongs to 40045

jnweiger (changes requested):
Casting a string "chunk_size=1234" to int does not do the trick. It yields 0.
Martin uses it wrong.

Insert coffee.

mmattel:
> Martin uses it wrong.

This is correct, but casting to int reduces to possible errors when using non integer...

phil-davis (dismissed):
Should we add more validation?
If `chunk_size` `is_numeric` and greater than 0 then it is probably OK and can be cast to `int` and used. Otherwise spit out an error message to the user and `exit(1)`

phil-davis:
And do we target this to `release-10.11.0` branch?

PR #40362, opened 2022-09-12
Validate previews:cleanup chunk_size

phil-davis:
## Description
Validate the `chunk_size` that is input by the user and give and error message if it is not a positive number.

## Related Issue
https://github.com/owncloud/core/pull/40045#issuecomment-1243721634

## How Has This Been Tested?
```
$ php occ previews:cleanup abc
chunk_size must be a positive integer.
$ php occ previews:cleanup 123
0 orphaned previews deleted
$ php occ previews:cleanup 0
chunk_size must be a positive integer.
```

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

Changelog not needed - this is a fix/enhancement to an unreleased feature.

### DIFF
--- a/core/Command/Previews/Cleanup.php
+++ b/core/Command/Previews/Cleanup.php
@@ -56,6 +56,17 @@ class Cleanup extends Base {
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$all = $input->hasOption('all');
 		$chunk_size = $input->getArgument('chunk_size');
+		$chunk_size_valid = false;
+		if (is_numeric($chunk_size)) {
+			$chunk_size = (int) $chunk_size;
+			if ($chunk_size > 0) {
+				$chunk_size_valid = true;
+			}
+		}
+		if (!$chunk_size_valid) {
+			$output->writeln('<error>chunk_size must be a positive integer.</error>');
+			return 1;
+		}
 
 		$pc = new PreviewCleanup($this->connection);
 		$count = $pc->process($all, $chunk_size, static function ($userId, $name, $action) use ($output) {


mmattel (approved):

DeepDiver1975:
Instead of implementing this in each and every command a method like this should be implemented in the base class so that we have consistent behavior in each occ command which uses this.

copy'n'paste is not what is meant with code reuse ;-)

mmattel:
@phil-davis  Thanks for taking care 👍

phil-davis:
> copy'n'paste is not what is meant with code reuse

I agree - who is going to make a set of common/flexible methods in the base class for doing various validations, and then refactor all the existing commands to use the common validation methods, and then get review-approval... ?

I don't see that happening for 10.11

jnweiger:
Consistency is hard. And also UX design is hard.

image

A small and easy step this is better than nothing. Thanks for the effort!

DeepDiver1975:
Based on iterative development approach this could be a simple start.
Move method to base class and use it in here.
Done.

phil-davis:
> Consistency is hard. And also UX design is hard.

But chunk_size is an "argument" not an "option", so the command-line user has to put it as "just an argument" - you do not mention the string "chunk_size". Commands could have some "guessing" code that tries to take a guess at what the user wants - but then if it guesses wrong there can be trouble.

jnweiger:
Naah, don't take my cheeky comment too serious. 
There is no need to iron out all possible UX misunderstandings here. It is an admin command.
The difference between syntax `[--] [chunksize]` and syntax `[--chunksize=N]` may be too subtle to switch the mental model from "its an option as usual" to "oh, no, this time its an argument" -- then the admin learns something the hard way. :shrug: 

Parameters, options, arguments, variables ... yada yada..

PR #40368, opened 2022-09-13
transifex url update

jnweiger:
The old https://www.transifex.com/projects/p/owncloud/ just goes to the generic entry page these days.

Pick something that stays specific to owncloud.


phil-davis (approved):
LGTM

PR #40377, opened 2022-09-20
[tests-only][full-ci]Added script to report closed issues in expected failure file

SagarGi:
### Description
This PR Adds bash script to report or find the issue that are closed but are still in the expected to failure file (In any Owncloud repository).

### Related issue
https://github.com/owncloud/QA/issues/716

### DIFF
--- a/tests/acceptance/lint-closed-issue.sh
+++ b/tests/acceptance/lint-closed-issue.sh
@@ -1,21 +1,21 @@
 #!/usr/bin/env bash
 
-# This is a script to check (lint) if there are any closed issue in the expected to failure file.
-# This script does not run in the drone pipeline but can be manually run often to find out if there are still closed issues in expected to fail.
+# This is a script to check (lint) if there are any closed issues in an expected failures file.
+# This script does not run in the drone pipeline but can be manually run often to find out if there are still closed issues in an expected failures file.
 # Pass GITHUB_ACCESS_TOKEN=(Your Access Token) for unlimited API request to GITHUB API
 # To run this script pass EXPECTED_FAILURES_FILE= (Path to expected to failure file)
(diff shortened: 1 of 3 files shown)


phil-davis:
Note: this script is not being used in CI (of core, or oCIS or reva...) because it would cause unexpected CI fails if someone closes an issue that is in an expected-failures file, and that is outside the real control of the CI of PRs. We don't want developers to get their PRs "broken" by "external" things like that.

We can use this script manually when we want to. In the future, if it seems worthwhile, we could put it in nightly CI - but first lets use it manually.

tests/acceptance/lint-closed-issue.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+# This is a script to check (lint) if there are any closed issue in the expected to failure file.
+# This script do not run in drone pipeline but can be manually run often to find out if there are still closed issue in expected to failure.
```
SwikritiT, line 4:
```suggestion
# This is a script to check (lint) if there are any closed issues in the expected failure file.
# This script does not run in the drone pipeline but can be manually run often to find out if there are still closed issues in expected to fail.
```

phil-davis:
I adjust the indent to always use tabs. And added a check for invalid issue numbers. I will review and test again later today.

phil-davis (approved):

phil-davis:
I added a few more checks for unusual things:
1) if the `( )` pair has space inside the brackets (as well as maybe a link). We want to consistently not have the space, and be able to reasonably reliably parse what should be a link.
2) detect when the issue "state" returns "null" or nothing at all. That happens for non-existent links, or some link to some other random place on the internet that does not return the JSON that we are hoping for.

The objective is to also try and find supposed "links to issues" that are "just plain wrong".

phil-davis:
This will be used to check expected-failures files: https://github.com/owncloud/ocis/issues/4635

PR #40378, opened 2022-09-20
Don't change the share target if it isn't available

jvillafanez:
## Description
The share folder we get could be different than the one configured if somehow the target folder isn't available. This might be bad since the shares will be placed in a different place than the one expected, and might be even more confusing if the storage becomes unavailable and the share are magically moved to a different place.

We'll always get the configured share folder now. If the share folder is missing and can't be created, an error will be logged

## Related Issue
https://github.com/owncloud/enterprise/issues/5383

## Motivation and Context
The case where the target folder can't be created, likely due to permissions, seems very rare. In this case, the admin should adjust the configured share folder to a different usable one. Meanwhile, the case where the whole FS becomes unavailable might be more common. If this happens, this PR will keep the shares in the same location after the FS comes back

## How Has This Been Tested?
Tested withe the steps provided in the linked issue. It still needs additional tests for possible side effects

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/lib/Helper.php
+++ b/apps/files_sharing/lib/Helper.php
@@ -296,12 +296,13 @@ class Helper {
 			}
 
 			$dirsToCreate = \array_reverse($dirsToCreate);
-			$shareFolder = '/';
 			foreach ($dirsToCreate as $dirToCreate) {
 				if ($view->mkdir($dirToCreate) === false) {
 					break;
 				}
-				$shareFolder = $dirToCreate;
+			}
+			if ($shareFolder !== $dirToCreate) {
+				\OCP\Util::writeLog('files_sharing', "Couldn't create missing share folder $shareFolder. Please choose a different share folder", \OCP\Util::ERROR);
 			}
 		}
 
--- a/apps/files_sharing/lib/Helper.php
+++ b/apps/files_sharing/lib/Helper.php
@@ -285,6 +285,10 @@ class Helper {
 		$shareFolder = Filesystem::normalizePath($shareFolder);
 
 		if (!$view->file_exists($shareFolder)) {
+			// if the share folder doesn't exists, create the folder in order
+			// to place the shares inside it. All the parent folders need to
+			// be created as well. By creating these folders, the shares will
+			// be accessible
 			$currentDir = $shareFolder;
 			$dirsToCreate = [$shareFolder];
 			while (($currentDir = \dirname($currentDir)) !== '/') {
@@ -302,6 +306,7 @@ class Helper {
 				}
 			}
 			if ($shareFolder !== $dirToCreate) {
+				// share folder might fail to be created
 				\OCP\Util::writeLog('files_sharing', "Couldn't create missing share folder $shareFolder. Please choose a different share folder", \OCP\Util::ERROR);
 			}
 		}
--- a/apps/files_sharing/tests/HelperTest.php
+++ b/apps/files_sharing/tests/HelperTest.php
@@ -32,21 +32,21 @@ use OC\Files\View;
  * @group DB
  */
 class HelperTest extends TestCase {
+	public function tearDown(): void {
+		// tests will set this config option. It needs to be removed
+		\OC::$server->getConfig()->deleteSystemValue('share_folder');
+		parent::tearDown();
+	}
 
 	/**
 	 * test set and get share folder
 	 */
 	public function testSetGetShareFolder() {
-		$this->assertSame('/', \OCA\Files_Sharing\Helper::getShareFolder());
-
 		\OCA\Files_Sharing\Helper::setShareFolder('/Shared/Folder');
 
 		$sharedFolder = \OCA\Files_Sharing\Helper::getShareFolder();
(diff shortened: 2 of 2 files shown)


pako81:
Can we get this reviewed and eventually merged, please? Thx.

mrow4a:
@pako81 I will have a look today

pako81:
@mrow4a review this? thx

apps/files_sharing/lib/Helper.php
```diff
@@ -296,12 +296,13 @@ public static function getShareFolder($view = null) {
...
 			$dirsToCreate = \array_reverse($dirsToCreate);
-			$shareFolder = '/';
 			foreach ($dirsToCreate as $dirToCreate) {
 				if ($view->mkdir($dirToCreate) === false) {
 					break;
 				}
-				$shareFolder = $dirToCreate;
+			}
+			if ($shareFolder !== $dirToCreate) {
+				\OCP\Util::writeLog('files_sharing', "Couldn't create missing share folder $shareFolder. Please choose a different share folder", \OCP\Util::ERROR);
```
mrow4a, line 305: in what case this error could happen ?

mrow4a (commented):
@jvillafanez this PR is quite hard to understand, as there is no unit test or acceptance test covering the case. Is it possible to add 1/2, or this is too complicated? 

Also, what I do not fully understand, is why we had this behaviour to move to shared folder in a first place?

apps/files_sharing/lib/Helper.php
jvillafanez: Usually folder permissions. If "/folder1/folder2/share" is configured, we have to create those folders, but there could be problems. There is no real guarantee that those folders will be correctly created.

mrow4a: just not sure, why we need to create these folders?

jvillafanez:
The tests are there, just adjusted to make them pass. I don't think there isn't any new case to cover.

> Also, what I do not fully understand, is why we had this behaviour to move to shared folder in a first place?

That's probably a wrong move. I assume that, if the folder becomes inaccessible, then the shares would also become inaccessible. That could be a reason to move them to the root folder, so at least you could access to the shares. However, the cause of the folder becoming inaccessible is what seems to be wrong.
The shares will likely point to another user's folder, so those ones must be accessible for the initial user to access those.
Basically, user1's FS is inaccessible while the rest of the users can access to the FS. I don't think this is a valid use case because the only way I can think of to reproduce that issue is by manipulating the permissions in the FS, which is something that shouldn't happen.

The current assumption in this PR is that the whole FS is down. This could happen with NFS having a temporary disconnection. If this is the case, moving the share target won't fix anything because the share are expected to also be inaccessible. It will also cause issues because none would expect the shares to be moved.

mrow4a:
@jvillafanez ok, I see. Not that I underrate federated shares, but what would happen to federated shares? Would my assumption be correct, that if the share points to federated server, if host FS is unavailable then we wont make federated share available even though most likely it is available ?

jvillafanez:
Yes, very likely.
The additional problem of changing the share target is that the change is permanent. The shares will be moved to the root, but when the problem is fixed, new shares will be created in the original location. This means that we'll have shares in 2 different places even though the user didn't actively move any share.

mrow4a:
> The additional problem of changing the share target is that the change is permanent. The shares will be moved to the root, but when the problem is fixed, new shares will be created in the original location. This means that we'll have shares in 2 different places even though the user didn't actively move any share.

@jvillafanez but this is what your PR tries to solve, correct?

jvillafanez:
Yes. With this PR, the share target won't change. The shares won't be moved to the root folder, so they might be temporary inaccessible until the problem with the FS is fixed.

mrow4a:
@jvillafanez ok, if you address conversations to resolve, then I approve

apps/files_sharing/lib/Helper.php
mrow4a: maybe a comment to document. for me it is not clear what this block of code is for.

jvillafanez:
commented

apps/files_sharing/lib/Helper.php
mrow4a: @jvillafanez I still do not get why we need to create folders. Are not shares virtual mounts, and not actual posix folders?

jvillafanez: the shares yes, but you need to place them in a folder. Usually, that folder is the root folder, so you don't need to do anything, but if not, you need to ensure the folder exists.

For example, if you have configured the share folder as "/share/folder", you need to create that folder so that if another user shares "file001.txt" with you, that share is placed in "/share/folder/file001.txt"
You'd have "/share" as regular folder, "/share/folder" as regular folder and "/share/folder/file001.txt" as virtual mount pointing to the location of the real "file001.txt" file in the other user's account.
For the case of folders, the behavior is the same: they'll be placed inside the "/share/folder" folder as virtual mounts.

mrow4a (approved):

PR #40389, opened 2022-09-28
allow specifying available space for objectstorages

mrow4a:
## Description

Feature: Allow specifying available space for objectstorages

Before this change, objectstorages were reporting only infinite storage space. This could have caused problems in other apps that rely on this storage method, e.g. metrics app that monitors available space

-  [x] Doc Update https://github.com/owncloud/docs-server/pull/658

## Related Issue
- Fixes https://github.com/owncloud/enterprise/issues/5384
- Related https://github.com/owncloud/enterprise/issues/5006

## How Has This Been Tested?
- manually in the developer instance
- unit/acceptance tests

Add configurations for files primary s3 being ceph, specifying `availableStorage`

```
<?php
$CONFIG = [
	'objectstore' => [
		'class' => 'OCA\Files_Primary_S3\S3Storage',
		'arguments' => [
			// replace with your bucket
			'bucket' => 'owncloud',
			// uncomment to indicate to owncloud available storage size in bytes (in this example 1TB)
			'availableStorage' => 1099511627776,
			// uncomment to enable server side encryption
			//'serversideencryption' => 'AES256',
			'options' => [
				// version and region are required
				'version' => '2006-03-01',
				'region'  => 'us-central-1',
				'credentials' => [
					// replace key and secret with your credentials
					'key' => 'owncloud123456',
					'secret' => 'secret123456',
				],
				'use_path_style_endpoint' => true,
				'endpoint' => 'http://ceph:8009/',
			],
		],
	],
];
```

Start ceph additionaly with owncloud instance
```
  ceph:
    image: ceph/daemon:master-dba849b-pacific-centos-8-aarch64
    command: demo
    restart: always
    environment:
      - NETWORK_AUTO_DETECT=4
      - RGW_NAME=ceph
      - CEPH_DEMO_UID=owncloud
      - CEPH_DEMO_ACCESS_KEY=owncloud123456
      - CEPH_DEMO_SECRET_KEY=secret123456
      - CEPH_DEMO_BUCKET=owncloud
    ports:
      - ${OWNCLOUD_S3_PORT}:8080
      - ${OWNCLOUD_S3_REST_PORT}:5000
```

## Types of changes

## Checklist:
- [x] Base code changes
- [x] Unit tests added
- [x] Acceptance tests added

## Example

Example use-case: metrics app

Now metric is correctly reported in the app
Screenshot 2022-09-27 at 23 37 12

### DIFF
--- a/changelog/unreleased/40389
+++ b/changelog/unreleased/40389
@@ -1,0 +1,6 @@
+Change: Allow specifying available space for objectstorages
+
+Before this change, objectstorages were reporting only infinite storage space. This could have caused problems in other apps that rely on this storage method, e.g. metrics app that monitors available space
+
+https://github.com/owncloud/core/pull/40192
+https://github.com/owncloud/enterprise/issues/5384
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -431,6 +431,16 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 	}
 
 	/**
+	 * overwrite this method if objectstorage supports getting total bucket objects size
+	 *
+	 * @return int
+	 */
+	protected function getTotalUsedStorage() {
+		$rootInfo = Filesystem::getFileInfo('/', false);
+		return $rootInfo->getSize();
+	}
+
+	/**
 	 * get the free space in the object storage as indicated by the objectstore config
 	 *
 	 * NOTE: getting total free space for objectstorage is not possible,
@@ -442,8 +452,7 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common {
 	 */
 	public function free_space($path) {
 		if (isset($this->availableStorage)) {
-			$rootInfo = Filesystem::getFileInfo('/', false);
-			$used = $rootInfo->getSize();
+			$used = $this->getTotalUsedStorage();
 			if ($used < 0) {
 				$used = 0;
 			}
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -27,6 +27,8 @@ namespace OC\Files\ObjectStore;
 
 use Icewind\Streams\IteratorDirectory;
 use OC\Files\Cache\CacheEntry;
+use OC\Files\Filesystem;
+use OCP\Files\FileInfo;
 use OCP\Files\NotFoundException;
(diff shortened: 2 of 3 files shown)


jvillafanez:
Maybe "availableStorageHint" makes more sense taking into account we aren't enforcing that limit. It seems more clear that the storage could have more or even less space available.

You can also check https://github.com/owncloud/core/blob/master/lib/private/legacy/helper.php#L141 so the admin can use "1GB" instead of the actual number of bytes.

Other than that, it looks good.

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -68,6 +74,10 @@ public function __construct($params) {
 		if (isset($params['objectPrefix'])) {
 			$this->objectPrefix = $params['objectPrefix'];
 		}
+		if (isset($params['availablestorage']) && \is_int($params['availablestorage'])) {
```
DeepDiver1975, line 77: camel case please

tests/lib/Files/ObjectStore/ObjectStoreTest.php
```diff
@@ -2,11 +2,7 @@
 /**
  * @author Thomas Müller <thomas.mueller@tmit.eu>
  *
-<<<<<<< HEAD
```
DeepDiver1975, line 5: :see_no_evil:

DeepDiver1975 (approved):

tests/lib/Files/ObjectStore/ObjectStoreTest.php
phil-davis: This crud was already in this file in core master!!! It's good to remove it.

mrow4a:
@DeepDiver1975 needed to adjust the class so that I could add test coverage (due to static functions with Filesystem::X)

phil-davis (approved):

changelog/unreleased/40389
```diff
@@ -0,0 +1,6 @@
+Change: Allow specifying available space for objectstorages
+
+Before this change, objectstorages were reporting only infinite storage space. This could have caused problems in other apps that rely on this storage method, e.g. metrics app that monitors available space
```
DeepDiver1975, line 3: wasn't there a 80 chars line limit?

mrow4a: @DeepDiver1975 done

DeepDiver1975 (approved):

jnweiger:
Confirmed in 10.12.0-rc.2 with scality s3.

PR #40409, opened 2022-10-07
[full-ci] [tests-only] Fix su exec ci

phil-davis:
## Description
1) `su-exec` is no longer provided in the owncloud-ci docker images, so do not use it, use `setpriv` (similar to https://github.com/owncloud-docker/base/pull/231 )
2) `setpriv` is only in Ubuntu 20.04, so use that for running tests. (cherry-picked from #40385 - we were thinking to do this anyway, and now we have to)
3) adjust the construction of pipeline names to avoid some names getting longer than 50 characters

This is needed to get CI green again. CI broke when `owncloudci/php` Docker was rebuilt recently, and base images no longer had `su-exec` in them.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -1378,7 +1378,7 @@ def phpTests(ctx, testType, withCoverage):
 
         for phpVersion in params["phpVersions"]:
             if testType == "phpunit":
-                command = "su-exec www-data bash tests/drone/test-phpunit.sh"
+                command = "setpriv --reuid=www-data --regid=www-data --init-groups bash tests/drone/test-phpunit.sh"
             else:
                 command = "unknown tbd"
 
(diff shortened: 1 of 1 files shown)


phil-davis:
Note: this PR explicitly uses the `owncloudci/php:7.4-ubuntu20.04` docker image. 

`owncloudci/php:7.4` uses Ubuntu 18.04 - next week I will think about if we should change that so it uses Ubuntu 20.04. But for now, we want to get CI green.

xoxys (approved):

Issue #40404, opened 2022-10-04
external storage's "applicable" column shows uids instead of displaynames for ldap groups


jvillafanez:
### Steps to reproduce
1. install user_ldap
2. setup user_ldap with barely the defaults. In particular, group displayname should be "cn" and groupname should be the default (empty)
3. Ensure we have multilpe groups available in the connected ldap
4. Enable external storage and setup a mount point

### Expected behaviour
Displayname for groups are shown in the "applicable" column

### Actual behaviour
Group uids are shown instead

![Screenshot from 2022-10-04 10-49-33](https://user-images.githubusercontent.com/1477829/193778995-78679737-79a2-4ed0-b68d-b6f9b462d4d5.png)

This will likely happen on fresh 10.11 installs. For updates, the migrations should set the groupname to "cn" automatically to keep the old behavior, so it shouldn't happen.
Note that setting the groupname to "cn" is discouraged because the groupname is expected to be unchangeable.

### Server configuration
**Operating system**:

**Web server:** apache

**Database:** mysql

**PHP version:** 7.4

**ownCloud version:** 10.11

**Updated from an older ownCloud or fresh install:** fresh install

**Where did you install ownCloud from:**

**Signing status (ownCloud 9.0 and above):**

```
Login as admin user into your ownCloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results into https://gist.github.com/ and puth the link here.
```

**The content of config/config.php:**

```
Log in to the web-UI with an administrator account and click on
'admin' -> 'Generate Config Report' -> 'Download ownCloud config report'
This report includes the config.php settings, the list of activated apps
and other details in a well sanitized form.

or 

If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your ownCloud installation folder

*ATTENTION:* Do not post your config.php file in public as is. Please use one of the above
methods whenever possible. Both, the generated reports from the web-ui and from occ config:list
consistently remove sensitive data. You still may want to review the report before sending.
If done manually then it is critical for your own privacy to dilligently
remove *all* host names, passwords, usernames, salts and other credentials before posting.
You should assume that attackers find such information and will use them against your systems.
```

**List of activated apps:**

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your ownCloud installation folder.
```

**Are you using external storage, if yes which one:** local/smb/sftp/...

**Are you using encryption:** yes/no

[...]

PR #40412, opened 2022-10-10
Files_external mount config will show group displaynames

jvillafanez:
## Description
Group ids were being shown in the "available for" selection of the files_external mount point configuration. This happens due to recent changes in the user_ldap app and OC 10.11

This PR shows the displayname instead, making eaiser for the admin to select the right group

## Related Issue
https://github.com/owncloud/core/issues/40404

## Motivation and Context
Since recently, the user_ldap app uses the objectuid (or similar attributes) as group id. Previously, the app used the "cn" attribute.
Anyway, using the displayname for visualizing the list makes more sense.

## How Has This Been Tested?
1. Setup an OC 10.11 and user_ldap 0.17.0 with default configuration. Make sure there are ldap groups available
2. Go to the settings -> admin storages, and click and the "available for" select.

The list includes groups showing their configured displayname (the "cn" attribute for ldap groups, but default)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_external/ajax/applicable.php
+++ b/apps/files_external/ajax/applicable.php
@@ -41,7 +41,7 @@ if (isset($_GET['offset'])) {
 
 $groups = [];
 foreach (\OC::$server->getGroupManager()->search($pattern, $limit, $offset) as $group) {
-	$groups[$group->getGID()] = $group->getGID();
+	$groups[$group->getGID()] = $group->getDisplayName();
 }
 
 $users = [];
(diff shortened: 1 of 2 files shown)


jvillafanez:
I thought we have a new user_ldap release with https://github.com/owncloud/user_ldap/pull/748 , but it seems it didn't happen yet. I probably was using an unreleased version.

mrow4a (approved):
I just wonder why this was GID in the past.. but as there is no comment in the code, I find this change legit.

jvillafanez:
I think the group displayname wasn't available from the beginning. It was added later.
For local groups, the gid and the displayname is the same, and for ldap, the cn attribute was used as gid, which is very visible. I guess nobody noticed because nobody was using some internal name like the entryuuid as gid.

PR #40433, opened 2022-10-18
Copy encryption keys first and then rename the files

jvillafanez:
## Description
Instead of renaming the file and then move the keys, we'll copy the keys first so we have a chance of reverting the changes if the file fails to be renamed. We either remove the original keys if rename succeeds or the copied keys if it fails.
Worst case, if exception happens and cleanup can't be performed, the file should still be accessible because there are valid keys still in place

## Related Issue
https://github.com/owncloud/enterprise/issues/5344

## Motivation and Context
This will harden the code and prevent the file from becoming inaccessible if something goes wrong.

## How Has This Been Tested?
Manually tested with the steps provided in the related issue.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/changelog/unreleased/40433
+++ b/changelog/unreleased/40433
@@ -1,0 +1,17 @@
+Change: Copy the encryption keys first and then rename the files
+
+Having encryption enabled, when a file was renamed, first the actual
+file was renamed, and then the encryption keys were moved to the new
+location. If something went wrong, it was possible that the keys weren't
+moved. This caused the file to become inaccessible because we couldn't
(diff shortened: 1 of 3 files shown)


pako81:
@jvillafanez grab some reviewers?

IljaN (approved):
LGTM :+1:  @DeepDiver1975 Could you also please have a look as you were recently working with key stuff.

IljaN:
Please add changelog

PR #40456, opened 2022-10-28
fix: on rejected shares no activities are reported to this user any more ... agin

DeepDiver1975:
Reverts owncloud/core#40454


DeepDiver1975:
:shrug:

PR #40473, opened 2022-11-04
Add a genrenic config sample description for S3 Primary

mmattel:
## Description
Referencing https://github.com/owncloud/docs-server/pull/704, I have seen that we do not have a minimalistic S3 config description in our config sample. This is fixed now. Note that there is no link to the documentation by purpose but a text finding the referenced document.

## Related Issue

- Fixes <issue_link>

## Motivation and Context
Completeness of docs

## How Has This Been Tested?
- text only

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Documentation ticket raised: https://github.com/owncloud/docs-server/pull/705

### DIFF
--- a/config/config.apps.sample.php
+++ b/config/config.apps.sample.php
@@ -449,6 +449,25 @@ $CONFIG = [
 'collabora_group' => '',
 
 /**
+ * App: S3 Primary Object Storage
+ *
+ * Possible keys: `objectstore` ARRAY
+ */
+
+/**
+ * Configure the access parameters for a particular S3 provider.
+ * The detailed configuration of that array depends on the S3 provider.
+ * This example can therefore only show the general construct.
+ * See the "S3 Compatible Object Storage as Primary Storage Location" documentation for more details.
+ */
+'objectstore' => [
+	'class' => 'Implementation\Of\OCP\Files\ObjectStore\IObjectStore',
+	'arguments' => [
+		// ...
+	],
+],
+
+/**
  * App: Windows Network Drive (WND)
  *
  * Note: This app is for Enterprise customers only.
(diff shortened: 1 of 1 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>apiAuth-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37207/35

config/config.apps.sample.php
```diff
@@ -448,6 +448,25 @@
...
+ */
+
+/**
+ * Configure the access parameters for a particular S3 provider.
+ * The detailed configuration of that array depends on the S3 provider.
+ * This example can therefore only show the general construct.
+ * See the "S3 Compatible Object Storage as Primary Storage Location" documentation for more details.
+ */
+'objectstore' => [
+	'class' => 'Implementation\Of\OCP\Files\ObjectStore\IObjectStore',
```
jnweiger, line 464:
```suggestion
	'class' => 'OCA\Files_Primary_S3\S3Storage',

```

How about a real world string there? -- I see it as a fixed string. Is there a reason to consider it as a variable? The only other implementation I know of is objectstore -- which is deprecated.

jnweiger (changes requested):

config/config.apps.sample.php
mmattel: Your point is valid but his is because the original documentation about S3 primary tells this and then further refines it to the implemented string. What we do here is to describe the constructor on a general base and then reference to the s3 documentation for the details.
As for now this is the most easiest way to follow what we already have. Honestly, I do not care if this is the only implementation or not, but if that needs a change, we have to do this also on the S3 primary documentation causing more work...
Just let me know

jnweiger (approved):
yes.

mmattel:
Thanks for the change, looks good, pls also do that change in docs-server 705 (that MUST be 1:1)

phil-davis (approved):

PR #40477, opened 2022-11-07
Add per-user storage view

pako81:
## Description
Add per-user storage view

## Related Issue
- https://github.com/owncloud/enterprise/issues/5468

## Motivation and Context
Make possible to visualise used storage on per-user basis

## How Has This Been Tested?

- Access ~/settings/users URL. A new column `Storage Used` will display the amount of used storage for every user. This option can be also opted in/out under `Settings` in order to align with all other available users' options.

## Types of changes

- [ ] Technical debt

## Checklist:

- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [X] Changelog item

### DIFF
--- a/settings/Controller/UsersController.php
+++ b/settings/Controller/UsersController.php
@@ -234,7 +234,7 @@ class UsersController extends Controller {
 			$storage = \OC_Helper::getStorageInfo('/');
 			$storageUsed = \sprint("%.1f MB", $storage['used'] / 1024 / 1024);
 		} catch (\Exception $e) {
-                        $this->log->error("Can't compute used storage for user " . $user->getUID() . ": " . $e->getMessage(), ['app' => 'settings']);
+			$this->log->error("Can't compute used storage for user " . $user->getUID() . ": " . $e->getMessage(), ['app' => 'settings']);
 			$storageUsed = "0 MB";
 		}
 
--- a/settings/Controller/UsersController.php
+++ b/settings/Controller/UsersController.php
@@ -232,8 +232,9 @@ class UsersController extends Controller {
 			\OC_Util::tearDownFS();
 			\OC_Util::setupFS($user->getUID());
 			$storage = \OC_Helper::getStorageInfo('/');
-			$storageUsed = sprintf("%d MB", $storage['used'] / 1024 / 1024);
-		} catch (\Exception $ex) {
+			$storageUsed = \sprint("%.1f MB", $storage['used'] / 1024 / 1024);
+		} catch (\Exception $e) {
+                        $this->log->error("Can't compute used storage for user " . $user->getUID() . ": " . $e->getMessage(), ['app' => 'settings']);
 			$storageUsed = "0 MB";
 		}
 
@@ -245,7 +246,7 @@ class UsersController extends Controller {
 			'isEnabled' => $user->isEnabled(),
 			'quota' => $user->getQuota(),
 			'storageLocation' => $user->getHome(),
-						'storageUsed' => $storageUsed,
+			'storageUsed' => $storageUsed,
 			'lastLogin' => $user->getLastLogin() * 1000,
 			'backend' => $user->getBackendClassName(),
 			'email' => $displayName,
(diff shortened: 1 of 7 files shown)


settings/Controller/UsersController.php
```diff
@@ -227,6 +227,16 @@ private function formatUserForIndex(IUser $user, array $userGroups = null) {
 			];
 		}
 
+		// storageUsed
+		try {
+			\OC_Util::tearDownFS();
+			\OC_Util::setupFS($user->getUID());
+			$storage = \OC_Helper::getStorageInfo('/');
+			$storageUsed = sprintf("%d MB", $storage['used'] / 1024 / 1024);
+		} catch (\Exception $ex) {
```
jvillafanez, line 236: This will require a log entry, specially because the 0 MB storage should be rare. We don't expect users without any file, so a 0 MB storage would means that something has happened, and in this case we must know what happened.
Also note that we're catching all the exceptions. We should catch only specific exceptions that we know it could happen and we know how to handle them. If we have to catch any exception, it's important to know what's going on. We might think we're catching an storageNotAvailable exception, but in fact, we might hiding a different exception, so troubleshooting will be harder.

settings/Controller/UsersController.php
```diff
@@ -227,6 +227,16 @@ private function formatUserForIndex(IUser $user, array $userGroups = null) {
 			];
 		}
 
+		// storageUsed
+		try {
+			\OC_Util::tearDownFS();
+			\OC_Util::setupFS($user->getUID());
+			$storage = \OC_Helper::getStorageInfo('/');
+			$storageUsed = sprintf("%d MB", $storage['used'] / 1024 / 1024);
```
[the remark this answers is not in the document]
pako81: I have rounded it to 1 decimal since the personal setting page is doing the same `You are using 4.1 MB of 10 MB (41.35 %)`, so we should align?

settings/Controller/UsersController.php
pako81: added log entry, not sure however if this is enough based on your comment above.

jvillafanez:
We probably need to restore the FS of the initial user after the `formatUserForIndex` loop unless there is no risk of leaving the FS of a random user there.
In addition, I think it should be documented in the method. The method doesn't have any hint about changing the FS of the user, so it might be unexpected.
Other than that, the code looks good

pako81:
@jvillafanez done. Not sure this is enough.

settings/Controller/UsersController.php
```diff
@@ -227,6 +227,18 @@ private function formatUserForIndex(IUser $user, array $userGroups = null) {
...
+		try {
+			\OC_Util::tearDownFS();
+			\OC_Util::setupFS($user->getUID());
+			$storage = \OC_Helper::getStorageInfo('/');
+			$storageUsed = \sprintf("%.1f MB", $storage['used'] / 1024 / 1024);
+		} catch (\Exception $e) {
+			$this->log->error("Can't compute used storage for user " . $user->getUID() . ": " . $e->getMessage(), ['app' => 'settings']);
+			$storageUsed = "0 MB";
+		} finally {
+			\OC_Util::tearDownFS();
```
jvillafanez, line 239: This seems the wrong place. Taking into account that this is a private function (can't be called from outside), the method is called at the end (chance of breaking things due to using the wrong user's FS is fairly low) and it's in a controller (there shouldn't be any code needed to access the FS after this), we could just ignore this and include some comments to warn that this method will setup a different FS.

jvillafanez:
Is it worthy to create a helper method somewhere in order to get the size of the root folder of the user's home without setting up a different FS? It seems less risky and there should be less overhead.
We're likely need to deal with buckets of 200 users, and I'm not sure how well this solution will perform taking into account that we need to setup each of the user's FS.

pako81:
@jvillafanez we can try, however I can see that i.e. the metrics app is doing the same https://github.com/owncloud/metrics/blob/master/lib/Metrics/QuotaMetrics.php#L62

jvillafanez:
Ok, then I think we just add some comments to warn that we're changing the FS user and we aren't reverting it back, at least for now. That should be good enough to finish this PR.

jvillafanez (approved):

PR #40483, opened 2022-11-09
Encryption copy keys first3

jvillafanez:
## Description
Follow up on https://github.com/owncloud/core/pull/40433 to solve https://github.com/owncloud/enterprise/issues/5344

The PR will copy the encryption keys instead of renaming them, so in case of error we can revert the action and we keep the encryption keys in the right location so the target file can be decrypted.

It seems the old PR was causing problems due to the failures caused by empty directories, which don't have encryption keys associated. The copyKeys method was reporting a failure because the expected directory with the keys doesn't exist. This has been adjusted: both copyKeys and renameKeys methods will return true if the directory doesn't exists.

## Related Issue
https://github.com/owncloud/enterprise/issues/5344

## Motivation and Context
Tests from other apps were failing due to the changes made in the old PR (which was reverted). This PR should cover the old failing scenarios.

## How Has This Been Tested?
Checked some of the previously failing scenarios.
1. With encryption active (after re-login to create the keys), create an empty directory
2. Delete the directory

With the old PR, a log about the encryption keys not being copied was showing. There was a forbidden exception happening. This is not happening with this PR.

1. With encryption active, create a directory and upload a file there
2. Run `occ encryption:recreate-master-key` command
3. From the desktop client, logout, login and upload a file

With the old PR, there were errors happening and uploading wasn't possible. This is also fixed with this PR.

1. With encryption active, create a directory
2. Share the directory with user2
3. With user2, upload a file and create a directory inside the shared directory.
4. With user2, delete the file and the folder

There were errors happening in the acceptance tests (these steps are based on the failing test). It works fine with the PR.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -268,28 +268,47 @@ class Encryption extends Wrapper {
 	 * @return bool
 	 */
 	public function rename($path1, $path2) {
-		$result = $this->storage->rename($path1, $path2);
-
-		if ($result &&
+		$renameOk = false;
+		$copyKeysOk = true;  // assume keys are copied, in case we deal with versions
+		if ($this->isVersion($path2) === false && $this->encryptionManager->isEnabled()) {
 			// versions always use the keys from the original file, so we can skip
 			// this step for versions
-			$this->isVersion($path2) === false &&
-			$this->encryptionManager->isEnabled()) {
 			$source = $this->getFullPath($path1);
 			if (!$this->util->isExcluded($source)) {
 				$target = $this->getFullPath($path2);
 				if (isset($this->unencryptedSize[$source])) {
 					$this->unencryptedSize[$target] = $this->unencryptedSize[$source];
 				}
-				$this->keyStorage->renameKeys($source, $target);
-				$module = $this->getEncryptionModule($path2);
-				if ($module) {
-					$module->update($target, $this->uid, []);
+
+				$copyKeysOk = $this->keyStorage->copyKeys($source, $target);
+				if ($copyKeysOk) {
+					$module = $this->getEncryptionModule($path2);
+					if ($module) {
+						$module->update($target, $this->uid, []);
+					}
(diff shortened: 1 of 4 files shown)


jvillafanez:
@phil-davis any chance to test this PR with all the apps in order to prevent having to revert it if something goes wrong?

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -268,28 +268,62 @@ public function unlink($path) {
...
+		if (!$isVersion && !$keysExcluded) {
 			// versions always use the keys from the original file, so we can skip
 			// this step for versions
-			$this->isVersion($path2) === false &&
-			$this->encryptionManager->isEnabled()) {
-			$source = $this->getFullPath($path1);
-			if (!$this->util->isExcluded($source)) {
-				$target = $this->getFullPath($path2);
-				if (isset($this->unencryptedSize[$source])) {
-					$this->unencryptedSize[$target] = $this->unencryptedSize[$source];
```
MeCorey, line 282: - [ ]

MeCorey (changes requested):

jvillafanez:
ping

IljaN (commented):
No need to check for exceptions as the files are always in a readable/acessible state between both operations. LGTM :+1: 

Restarting CI to check why it fails.

ownclouders:
:boom: Acceptance tests pipeline <strong>webUILogin-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37415/125

IljaN:
CI seems to non-deterministically fail on different steps on each re-start :thinking:

pako81:
The only test failing is related to `fsweb.test.owncloud.com` which is a known issue with this host, so it can be ignored.

IljaN (approved):

PR #40500, opened 2022-11-17
[tests-only][do-not-merge]Add debug when search match not found

amrita-shrestha:
## Description
 
This PR is not for fixing. It's just for debugging purposes for the future, it adds code to print the search request responses when a match is not found.

### Related issue 
- https://github.com/owncloud/search_elastic/issues/302

### DIFF
--- a/tests/acceptance/features/bootstrap/WebDav.php
+++ b/tests/acceptance/features/bootstrap/WebDav.php
@@ -5375,6 +5375,10 @@ trait WebDav {
 		if ($entryNameToSearch === null) {
 			return $results;
 		}
+
+		//print all response of search request if match not found
+		print_r($multistatusResults);
+
 		return false;
 	}
 


tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -5375,6 +5375,10 @@ public function findEntryFromPropfindResponse(
 		if ($entryNameToSearch === null) {
 			return $results;
 		}
+
+		//print all response of search request if match not found
+		print_r($multistatusResults);
+
```
saw-jan, line 5381: :thinking:  why do we need this? Anyone who is debugging can do it locally if that's needed.

tests/acceptance/features/bootstrap/WebDav.php
```diff
@@ -5375,6 +5375,10 @@ public function findEntryFromPropfindResponse(
 		if ($entryNameToSearch === null) {
 			return $results;
 		}
+
+		//print all response of search request if match not found
+		print_r($multistatusResults);
```
kiranparajuli589, line 5380: If this is needed for CI. how about just doing an assertion step here. sth like:
```suggestion
		Assert::assertFail("Expected to be something :) but got something else. \n Mulistatus results is here:" . print_r(multistatusResults, true))
```

tests/acceptance/features/bootstrap/WebDav.php
amrita-shrestha: intermittent failures like - https://github.com/owncloud/search_elastic/issues/302 are hard to reproduce and sometimes it looks impossible to reproduce locally in such cases we close such an issue if it didn't reoccur in one month time.
I think it will be better if we can print the response to the search request when a test fails.

tests/acceptance/features/bootstrap/WebDav.php
amrita-shrestha: the printed response contains a response to the search response.It saves our time and gives some idea why tests are failing

tests/acceptance/features/bootstrap/WebDav.php
saw-jan: hmmm, if I were to vote, I would vote against it.

amrita-shrestha:
currently closing this PR because using print_r for debugging is not a good practice.

PR #40502, opened 2022-11-17
improve wording with untrusted domain

### LINKED ISSUE #35800: misleading javascript error, when client violates trusted domain setting
Server version 10.2.1RC2 and Win10 desktop client 2.6.0daily20190702.

The server host changed its ip address, the trusted domains setting in config.php was not updated. Now the client tries to connect to the new ip address and is greeted with a 
wrong error message about javascript:
image

The javascript error should not appear. The client is not a web browser.

(The wording of the actual trusted domains error is also misleading. Saying "from" may lead to the conclusion that the client ip address is at fault, which it is not.)

jnweiger:
This avoids a misunderstanding and thus fixes https://github.com/owncloud/core/issues/35800

### DIFF
--- a/core/templates/untrustedDomain.php
+++ b/core/templates/untrustedDomain.php
@@ -2,7 +2,7 @@
 
 <ul class="error-wide">
 	<li class='error'>
-		<?php p($l->t('You are accessing the server from an untrusted domain.')); ?><br>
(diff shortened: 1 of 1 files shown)


pako81 (approved):

pako81:
@jnweiger thanks, it makes indeed sense.

Issue #40457, opened 2022-10-28
PHP Unit test failing for sqlite windows


grgprarup:
Drone build: https://drone.owncloud.com/owncloud/core/37121/19/6, https://drone.owncloud.com/owncloud/core/37123/23/6, https://drone.owncloud.com/owncloud/core/37124/23/7, https://drone.owncloud.com/owncloud/core/37128/23/7
Tests are failing for pipeline `phpunit-php7.4-sqlite-windows`
```
+ phpunit_cmd='php ../lib/composer/bin/phpunit'
+ [[ true == \t\r\u\e ]]
+ phpunit_cmd='phpdbg -d memory_limit=6G -rr ../lib/composer/bin/phpunit'
+ [[ -n windows ]]
+ set_up_external_storage
+ php occ app:enable files_external
files_external enabled
+ php occ config:app:set core enable_external_storage --value=yes
Config value enable_external_storage for app core set to yes
+ case "${FILES_EXTERNAL_TYPE}" in
+ wait-for-it -t 600 fsweb.test.owncloud.com:445
services aren't ready in 10m0s
```

grgprarup:
@phil-davis Do you have any idea about this failure?

saw-jan:
Looks like the external server is down. I don't think we can do anything other than the discussion on what to do in such cases.

Some queries that may help to decide:
- Will `fsweb.test.owncloud.com:445` be available or is it permanently down?
- Any ideas on what can be done whenever the external service is down?

phil-davis:
PR #40426 is waiting for a decision. The fsweb server seems to have gone down again.
I suggest that we merge the PR to get CI green. And then someone can decide what to do to get a CI environment going again.

jnweiger:
@GeraldLeikam this is the one that is waiting for your new AD server

phil-davis:
@GeraldLeikam (cc @jnweiger ) what happened about getting a new server for this?

saw-jan:
The windows pipeline has been removed: https://github.com/owncloud/core/pull/40426

DeepDiver1975:
Once the new server is there we will reactivate the pipeline.

PR #40503, opened 2022-11-21
Allow to temporarily ignore invalid federated shares in tightly federated setup

mrow4a:
fixes https://github.com/owncloud/enterprise/issues/5427

Currently, if federated share is invalid or api endpoint returns not found, availability check would validate whether this is a problem with a server and if checks complete that given share is removed. However, in some cases these checks might not be enough (e.g. complex migrations in tightly federated setups), and in that case invalidation behaviour can be disabled using below app setting:

```
occ config:app:set files_sharing enable_cleanup_invalid_external_shares --value no
```

Then instead of removal below warning is displayed

Screenshot 2022-12-16 at 11 49 16

This PR also improves handling and logging to better debug the situation

### DIFF
--- a/apps/files_sharing/lib/External/Storage.php
+++ b/apps/files_sharing/lib/External/Storage.php
@@ -226,10 +226,10 @@ class Storage extends DAV implements ISharedStorage {
 	 */
 	public function checkStorageAvailability() {
 		// WARNING: Disabling this setting is recommended only in tightly integrated federated setups as temporarly setting,
-		//  should NOT be used in decentralized federation setup. It could cause bad user experience when dealing
-		//  with deleted external shares
+		// should NOT be used in decentralized federation setup. It could cause bad user experience when dealing
+		// with deleted external shares
 		// NOTE: This allows administrator to disable cleanup of invalid shares so only manual removal by user or admin is possible,
-		//  this is particularly useful when dealing with complex migrations that could cause unexpected server responses and instabilities.
+		// this is particularly useful when dealing with complex migrations that could cause unexpected server responses and instabilities.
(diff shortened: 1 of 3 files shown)


pako81:
@mrow4a should we move this forward?

apps/files_sharing/lib/External/Storage.php
```diff
@@ -221,42 +225,89 @@ public function test() {
...
 		}
+
+		if (!$checkShareStorage) {
+			// propfind returns false if not found returned,
+			// can either mean that the share no longer exists or there is no ownCloud on
+			// the remote (proxy could return 404 for that path), check remote if accessible
+			if ($this->testRemote()) {
+				// valid ownCloud instance means that the external share no longer exists
+				// since this is permanent (re-sharing the file will create a new token)
+
```
pako81, line 282: Missing the last part of the comment here `// we remove the invalid storage`?

pako81 (approved):

pako81:
Apart from a small remark about comments, LGTM 👍

apps/files_sharing/lib/External/Storage.php
mrow4a: adjusted

mrow4a:
@pako81 I cannot merge this due to https://github.com/owncloud/core/issues/40457

mrow4a:
@pako81 sonar coverage is hard to do with this class, would need some refactor. would need force-merge still

mmattel:
`occ config:app:set files_sharing enable_cleanup_invalid_external_shares --value no`

Needs documentation in our admin docs, pls file an issue in the `docs-ocis` repo

PR #40504, opened 2022-11-22
[tests-only] Match step to check group in the share response

saw-jan:
## Description
Match the these steps:
- `Then group "grp2" should be included in the response`
- `Then group "grp1" should not be included in the response`

## Related Issue
- Required for https://github.com/owncloud/ocis/pull/5076

## How Has This Been Tested?
- test environment: local

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/tests/acceptance/features/bootstrap/Sharing.php
+++ b/tests/acceptance/features/bootstrap/Sharing.php
@@ -1431,15 +1431,18 @@ trait Sharing {
 	}
 
 	/**
-	 * @Then /^user "([^"]*)" should be included in the response$/
+	 * @Then /^(user|group) "([^"]*)" should be included in the response$/
 	 *
+	 * @param string $type
 	 * @param string $user
 	 *
 	 * @return void
 	 * @throws Exception
 	 */
-	public function checkSharedUserInResponse(string $user):void {
-		$user = $this->getActualUsername($user);
+	public function checkSharedUserOrGroupInResponse(string $type, string $user):void {
+		if ($type === 'user') {
+			$user = $this->getActualUsername($user);
+		}
 		Assert::assertTrue(
 			$this->isFieldInResponse('share_with', "$user"),
 			"'share_with' value '$user' was not found in response"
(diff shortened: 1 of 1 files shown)


tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1431,15 +1431,18 @@ public function checkSharedFileAsPathNotInResponse(string $filename):void {
...
 	 *
+	 * @param string $type
 	 * @param string $user
 	 *
 	 * @return void
 	 * @throws Exception
 	 */
-	public function checkSharedUserInResponse(string $user):void {
-		$user = $this->getActualUsername($user);
+	public function checkSharedUserInResponse(string $type, string $user):void {
```
kiranparajuli589, line 1442:
```suggestion
	public function checkSharedUserOrGroupInResponse(string $type, string $userOrGroup):void {
```

tests/acceptance/features/bootstrap/Sharing.php
```diff
@@ -1448,16 +1451,17 @@ public function checkSharedUserInResponse(string $user):void {
...
+	 * @Then /^group "([^"]*)" should not be included in the response$/
 	 *
-	 * @param string $user
+	 * @param string $userOrGroup
 	 *
 	 * @return void
 	 * @throws Exception
 	 */
-	public function checkSharedUserNotInResponse(string $user):void {
+	public function checkSharedUserNotInResponse(string $userOrGroup):void {
```
kiranparajuli589, line 1461:
```suggestion
	public function checkSharedUserOrGroupNotInResponse(string $userOrGroup):void {
```

kiranparajuli589 (commented):
lgtm :+1:

phil-davis (approved):
Note: if there are tests that have both a user and a group with the same name, then this `Then` step might not be specific enough. In that case we might expect 2 entries in the response, one for user "xyz" and one for group "xyz". That will need a way to explicitly check for a combination of share_type and share_with.

PR #40505, opened 2022-11-22
Prevent creation of empty files/folders for accounts having no available quota

pako81:
## Description
Prevent creation of empty files/folders for accounts having no available quota

## Related Issue
- https://github.com/owncloud/enterprise/issues/5478

## Motivation and Context
Until now it was possible for users having 0 quota or who already reached the limit of their assigned quota to still create empty files/folders, which generates confusion. The PR fixes this behaviour.

## How Has This Been Tested?
Manually:

Test 1. :

- as admin user `admin` assign quota `0 B` to user `user1`
- login as user `user1` over WebUI and try to create empty files or folders: an error `Could not create..` is returned

Test 2. :

- as `user1` who has a valid quota assigned create a folder `for guests` and share it with full permissions with guest user `guest@owncloud.com`
-  as admin user `admin` assign now quota `0 B` to user `user1`
- login as guest user `guest@owncloud.com`, enter the folder `for guests` and try to create empty files or (sub)folders: an error `Could not create..` is returned

Test 3. : 

- Same steps as Test 1. and 2. but this time with `user1` having i.e. 10MB quota assigned and already using 100% of the available space

Test 4. :

- as admin user `admin` assign quota `0 B` to user `user1`
- connect `user1` via desktop client to his personal storage: sync client returns a 403 Forbidden to `MKCOL` and `PUT` operations when trying to respectively create empty folders and files.

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -155,6 +155,11 @@ class Directory extends Node implements ICollection, IQuota, IMoveTarget {
 				throw new SabreForbidden();
 			}
 
+			list($used, $free) = $this->getQuotaInfo();
+			if ($free == 0) {
+				throw new SabreForbidden();
+			}
+
 			$this->fileView->verifyPath($this->path, $name);
 
 			$path = $this->fileView->getAbsolutePath($this->path) . '/' . $name;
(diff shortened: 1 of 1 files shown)


pako81:
@jvillafanez do you think the approach here is correct?

jvillafanez:
The approach could be fine if we want to allow to create folders under special circumstances. I mean, the 0 quota only applies to access through the webdav interface, but there could be other type of accesses that could bypass that check by using the view object directly. There is probably no good use case for this, but it's an option.

If we need to ensure that no folder is added under any circumstance (having 0 quota), we'll have to move the check elsewhere.

jvillafanez:
Not sure why I left the review as pending... anyway I'm assuming the approach is ok (regarding my previous post)

pako81:
Re-opened on https://github.com/owncloud/core/pull/40567.

PR #40513, opened 2022-11-23
Set checksum if not seeked

jvillafanez:
## Description
Range downloads might set up a wrong checksum if there isn't a checksum set. This causes problems later because a full download will send the wrong checksum and clients might flag the download even though the file was downloaded ok.

Note that range downloads might still get the checksum of the whole file, not just the downloaded piece.

In addition, appending content won't calculate a checksum.

There are some additional cases that might still need to be covered

## Related Issue
https://github.com/owncloud/enterprise/issues/5457

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [x] Unit tests added
- [ ] Acceptance tests added

### Notes
* On average, range download don't generate a checksum because not the whole file is read. However, it really depends on how PHP reads the stream internally. PHP usually read blocks of 8KB and keep them in a buffer even though only a few bytes are requested and returned. This means that, for small files (less than 8KB), there is a chance that the whole file is buffered, so in this case a checksum could be stored even though, technically, not the whole file is returned.
* It's expected that, regardless of the outcome, the checksum stored in the DB is always correct. This is important taking into account the behavior of the previous point.
* While the expected actions on the data stream are expected to be easy coming from the outside (either read the whole o a part of the file, or write contents) additional actions might happen, such as overwrite a part of the stream while leaving the rest intact. In this case, a conservative approach is taken, so if there are doubts on whether we've read the whole file or the checksum could be compromised (by moving the stream pointer to a previous position, for example), no checksum will be stored because it's very likely it will be wrong.
* Partial data won't be saved among requests. This could be relevant for apps that download the file by chunks. The app could request the first megabyte, then the second, then the third... In this case, even though the file ends up being downloaded, no checksum will be stored.

### DIFF
--- a/changelog/unreleased/40513
+++ b/changelog/unreleased/40513
@@ -1,0 +1,15 @@
+Bugfix: Store checksums only if the whole stream has been read
+
+Previously, range downloads (or downloads requesting a specific
+byte range) would store a checksum, if needed, based only on the
+requested range. This causes problems because the checksum is
+expected to be for the whole file.
+
+Now, those range downloads won't store a checksum because only a
+part of the file has been read, so the checksum would be incomplete.
+
+Some additional cases have been taken into account, mostly based
+on actions that could happen on the data stream, but they shouldn't
+happen normally.
+
+https://github.com/owncloud/core/pull/40513
--- a/lib/private/Files/Stream/Checksum.php
+++ b/lib/private/Files/Stream/Checksum.php
@@ -77,11 +77,17 @@ class Checksum extends Wrapper {
 			]
 		]);
 
+		// need to check the underlying stream's mode
+		// The `wrapSource` will use 'r+' by default, so the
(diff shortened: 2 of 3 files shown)


ownclouders:

https://drone.owncloud.com/owncloud/core/37334/168

jvillafanez:
@phil-davis is it possible to add some acceptance tests for this scenario? Basically, if the download has a request range, no checksum should be generated. If there is a checksum already, it won't be changed.
A relatively big file (1-5MB) should be enough. A small file might get hit by the first point in the notes, so a checksum might be generated even though a couple of bytes have been requested.

IljaN (approved):
Nice find! LGTM :100:

jnweiger:
Merging. We have only the one expected fsweb error in the ci, that is unrelated to this.

PR #40520, opened 2022-11-30
Refactor login to converge in the same function

jvillafanez:
## Description
Refactor login process so it ends going through the new `loginInOwnCloud` method.

The new method will be perform the following:
* It will trigger the login events (pre and post login)
* Perform common checks to all auth mechanisms, such as checking if the logging user is disabled
* Set the user, and prepare it
* Update login timestamp

Specific checks based on the auth mechanism (checking if the password is correct, or checking the expiration of the token) will be performed outside.

Additional changes brought in the PR:
* As said, the login will be performed by the new `loginInOwnCloud` method. This has caused some changes for some authentication methods about when the login events are triggered. This means that for some auth methods, the events will be triggered later and after some operations / checks have been performed. This might cause issues and will require verification.
* For the `loginWithToken`, the token will be validated before trying to login with it. Previously, we logged in and then validated the token, so this might cause changes in the behavior
* The `createSessionToken` will be called after a successful login. There were cases were this was called before login and others after login, so it's unclear when the session token should be created. This might need to be adjusted.
* `preRememberedLogin` and `postRememberedLogin` events have been removed. These were additional events triggered by the "remember me" login which was broken until recently, so it's very unlikely they were used. Currently, the "remember me" login won't trigger any login event, but it might need to trigger the regular ones (prelogin and postlogin). They're currently being skipped.

Known issues:
* openidconnect's `LoginFlowController` will call the `loginUser` method without a module class. This will report the login type as empty instead of the expected openidconnect class name. There is nothing we can do without changing the app. This is harmless for this PR, but it will affect further functionality.
* Since user_shibboleth depends on an apache module and ownCloud will get the info from apache, the "apache" login type will be reported for the user_shibboleth app.

## Related Issue
Preparations for https://github.com/owncloud/enterprise/issues/5295

## Motivation and Context
This is a preparation to bring new functionality. We need a centralized point where all login paths can converge.

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -242,7 +242,8 @@ class Session implements IUserSession, Emitter {
 	 * - For browsers, the session token validity is checked
 	 */
 	public function validateSession() {
-		if (!$this->getUser()) {
+		$sessionUser = $this->getUser();
+		if (!$sessionUser) {
 			return;
 		}
 
@@ -258,6 +259,11 @@ class Session implements IUserSession, Emitter {
 			}
 		} else {
 			$token = $appPassword;
+		}
+
+		if (!$sessionUser->isEnabled()) {
+			$this->tokenProvider->invalidateToken($token);
+			$this->logout();
 		}
 
 		if (!$this->validateToken($token)) {
@@ -567,11 +573,19 @@ class Session implements IUserSession, Emitter {
 			return false;
 		}
 
-		$loginOk = $this->loginInOwnCloud('token', $user, $password);
+		try {
+			$loginOk = $this->loginInOwnCloud('token', $user, $password);
 
-		// set the app password
-		if ($loginOk) {
-			$this->session->set('app_password', $token);
+			// set the app password
+			if ($loginOk) {
+				$this->session->set('app_password', $token);
+			} else {
+				$this->tokenProvider->invalidateToken($token);
+			}
+		} catch (LoginException $e) {
+			// need to invalidate the token
+			$this->tokenProvider->invalidateToken($token);
+			throw $e;
 		}
 
 		return $loginOk;
@@ -755,20 +769,6 @@ class Session implements IUserSession, Emitter {
 		} catch (PasswordlessTokenException $ex) {
 			// Token has no password
 
-			if ($this->activeUser !== null && !$this->activeUser->isEnabled()) {
-				$this->logger->debug(
-					'user {uid}, {email}, {displayName} was disabled',
-					[
-						'app' => __METHOD__,
-						'uid' => $this->activeUser->getUID(),
-						'email' => $this->activeUser->getEMailAddress(),
-						'displayName' => $this->activeUser->getDisplayName(),
-					]
-				);
-				$this->tokenProvider->invalidateToken($token);
-				return false;
-			}
-
 			$dbToken->setLastCheck($now);
 			$this->tokenProvider->updateToken($dbToken);
 			return true;
@@ -781,28 +781,14 @@ class Session implements IUserSession, Emitter {
 			return true;
 		}
 
-		if ($this->manager->checkPassword($dbToken->getLoginName(), $pwd) === false
-			|| ($this->activeUser !== null && !$this->activeUser->isEnabled())) {
-			// FIXME: protect debug statement this way to avoid regressions
-			if ($this->activeUser !== null) {
-				$this->logger->debug(
-					'user uid {uid}, email {email}, displayName {displayName} was disabled or password changed',
-					[
-						'app' => __METHOD__,
-						'uid' => $this->activeUser->getUID(),
-						'email' => $this->activeUser->getEMailAddress(),
-						'displayName' => $this->activeUser->getDisplayName(),
-					]
-				);
-			} else {
-				$this->logger->debug(
-					'user with login name {loginName} was disabled or password changed (no activeUser)',
-					[
-						'app' => __METHOD__,
-						'loginName' => $dbToken->getLoginName(),
-					]
-				);
-			}
+		if ($this->manager->checkPassword($dbToken->getLoginName(), $pwd) === false) {
+			$this->logger->debug(
+				'password changed for user with login name {loginName}',
+				[
+					'app' => __METHOD__,
+					'loginName' => $dbToken->getLoginName(),
+				]
+			);
 
 			$this->tokenProvider->invalidateToken($token);
 			// Password has changed or user was disabled -> log user out
@@ -958,6 +944,10 @@ class Session implements IUserSession, Emitter {
 	 * @throws LoginException if an app canceled the login process or the user is not enabled
 	 */
 	public function loginUser(IUser $user = null, $password = null, $authModuleClass = null) {
+		if ($user === null) {
+			$this->emitFailedLogin(null);
+			return false;
+		}
 		// openidconnect calls the loginUser method. It might not have an $authModuleClass
 		return $this->loginInOwnCloud($authModuleClass, $user, $password);
 	}
@@ -1047,6 +1037,8 @@ class Session implements IUserSession, Emitter {
 			// injecting l10n does not work - there is a circular dependency between session and \OCP\L10N\IFactory
 			$message = \OC::$server->getL10N('lib')->t('Login canceled by app');
 			throw new LoginException($message);
+		} else {
+			$this->logger->info("login {$user->getUID()} cancelled. User disabled", ['app' => __METHOD__]);
 		}
 
 		// injecting l10n does not work - there is a circular dependency between session and \OCP\L10N\IFactory
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -78,8 +78,6 @@ use Symfony\Component\EventDispatcher\GenericEvent;
  * - preLogin(string $user, string $password)
  * - postLogin(\OC\User\User $user, string $password)
  * - failedLogin(string $user)
- * - preRememberedLogin(string $uid)
- * - postRememberedLogin(\OC\User\User $user)
  * - logout()
  * - postLogout()
  *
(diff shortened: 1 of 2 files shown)


jvillafanez:
Need to check the following, having oauth2 and openidconnect apps active. It seems openidconnect triggers even though it's an oauth2 token.
```
{"reqId":"bbc1f402-53e7-469c-8f18-4a80ffa455a4","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"OpenID","method":"GET","url":"\/ocs\/v1.php\/cloud\/capabilities?format=json","message":"OpenIdConnectAuthModule::authToken Bearer SKWHXCJ4fxeByCoglUK0AlQ2tkeGZoHGmEO57XRgS3cbo7MlEroVgXdvxxvs9Irf"}
{"reqId":"bbc1f402-53e7-469c-8f18-4a80ffa455a4","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"PHP","method":"GET","url":"\/ocs\/v1.php\/cloud\/capabilities?format=json","message":"Undefined offset: 1 at \/var\/www\/owncloud\/apps\/openidconnect\/vendor\/jumbojett\/openid-connect-php\/src\/OpenIDConnectClient.php#1319"}
{"reqId":"bbc1f402-53e7-469c-8f18-4a80ffa455a4","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"no app in context","method":"GET","url":"\/ocs\/v1.php\/cloud\/capabilities?format=json","message":"Token (as per introspection) is inactive: {\"active\":false}"}
{"reqId":"bbc1f402-53e7-469c-8f18-4a80ffa455a4","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"OpenID","method":"GET","url":"\/ocs\/v1.php\/cloud\/capabilities?format=json","message":"Exception: {\"Exception\":\"Jumbojett\\\\OpenIDConnectClientException\",\"Message\":\"Token (as per introspection) is inactive\",\"Code\":0,\"Trace\":\"#0 \\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/OpenIdConnectAuthModule.php(160): OCA\\\\OpenIdConnect\\\\Client->verifyToken()\\n#1 \\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/OpenIdConnectAuthModule.php(116): OCA\\\\OpenIdConnect\\\\OpenIdConnectAuthModule->verifyToken()\\n#2 \\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/OpenIdConnectAuthModule.php(95): OCA\\\\OpenIdConnect\\\\OpenIdConnectAuthModule->authToken()\\n#3 \\\/var\\\/www\\\/owncloud\\\/lib\\\/private\\\/User\\\/Session.php(1235): OCA\\\\OpenIdConnect\\\\OpenIdConnectAuthModule->auth()\\n#4 \\\/var\\\/www\\\/owncloud\\\/ocs\\\/v1.php(80): OC\\\\User\\\\Session->verifyAuthHeaders()\\n#5 {main}\",\"File\":\"\\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/Client.php\",\"Line\":211}"}
{"reqId":"75972b02-db89-47c9-a075-f08cb32b030b","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"OpenID","method":"GET","url":"\/ocs\/v1.php\/cloud\/user?format=json","message":"OpenIdConnectAuthModule::authToken Bearer SKWHXCJ4fxeByCoglUK0AlQ2tkeGZoHGmEO57XRgS3cbo7MlEroVgXdvxxvs9Irf"}
{"reqId":"75972b02-db89-47c9-a075-f08cb32b030b","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"PHP","method":"GET","url":"\/ocs\/v1.php\/cloud\/user?format=json","message":"Undefined offset: 1 at \/var\/www\/owncloud\/apps\/openidconnect\/vendor\/jumbojett\/openid-connect-php\/src\/OpenIDConnectClient.php#1319"}
{"reqId":"75972b02-db89-47c9-a075-f08cb32b030b","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"no app in context","method":"GET","url":"\/ocs\/v1.php\/cloud\/user?format=json","message":"Token (as per introspection) is inactive: {\"active\":false}"}
{"reqId":"75972b02-db89-47c9-a075-f08cb32b030b","level":3,"time":"2022-11-30T09:41:46+00:00","remoteAddr":"10.0.2.27","user":"admin","app":"OpenID","method":"GET","url":"\/ocs\/v1.php\/cloud\/user?format=json","message":"Exception: {\"Exception\":\"Jumbojett\\\\OpenIDConnectClientException\",\"Message\":\"Token (as per introspection) is inactive\",\"Code\":0,\"Trace\":\"#0 \\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/OpenIdConnectAuthModule.php(160): OCA\\\\OpenIdConnect\\\\Client->verifyToken()\\n#1 \\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/OpenIdConnectAuthModule.php(116): OCA\\\\OpenIdConnect\\\\OpenIdConnectAuthModule->verifyToken()\\n#2 \\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/OpenIdConnectAuthModule.php(95): OCA\\\\OpenIdConnect\\\\OpenIdConnectAuthModule->authToken()\\n#3 \\\/var\\\/www\\\/owncloud\\\/lib\\\/private\\\/User\\\/Session.php(1235): OCA\\\\OpenIdConnect\\\\OpenIdConnectAuthModule->auth()\\n#4 \\\/var\\\/www\\\/owncloud\\\/ocs\\\/v1.php(80): OC\\\\User\\\\Session->verifyAuthHeaders()\\n#5 {main}\",\"File\":\"\\\/var\\\/www\\\/owncloud\\\/apps\\\/openidconnect\\\/lib\\\/Client.php\",\"Line\":211}"}
```

DeepDiver1975:
> Need to check the following, having oauth2 and openidconnect apps active. It seems openidconnect triggers even though it's an oauth2 token.

Both are Bearer tokens - both modules are triggered. Generally speaking this is/was fine as these two modules learned to live together ....

jvillafanez:
Are both modules expected to be enabled at the same time? If so, we should fix those errors because they happen periodically with the desktop client. Otherwise, we could treat it as incompatible apps so only one of them should be enabled.

jvillafanez:
Things tested so far:
* regular login page -> reports "password" login
* basic auth (for webdav, for example) -> "password" login
* app password -> reports "token" login (also in the login page)
* user_shibboleth -> reports "apache" login
* oauth2 -> reports "OCA\OAuth2\AuthModule" login
* openidconnect from the login page -> reports empty login (known issue)
* openidconnect (using an access token) -> reports "OCA\OpenIdConnect\OpenIdConnectAuthModule" login

Not tested:
* kerberos -> expects "OCA\Kerberos\AuthModule"
* user_external -> expects "password" (no specific auth module found, likely using the regular user + password code path, but checking against different services)

jvillafanez:
Some additional changes in the code:
* the `validateSession` method will explicitly check if the user is disabled. This was happening inside the `validateToken` method, but we're already checking it in the new `loginInOwnCloud` method. Any login attempt should go through the `loginInOwnCloud` method, so we'll check if the user is disabled or not. However, the `validateSession` won't go through the login process, so we need an explicit check there.
* Additional changes to the token login:
  * Token will be invalidated if the login failed, or if a LoginException happens (not sure if these cases were covered previously)
  * Checking if the user is disabled has been removed from the `validateToken` method.
* The `loginUser` requires a valid user, if no user is provided, a failed login will be triggered. (This is a fix in this PR. Code in master had this check in place)
* New info log added to notify the user is disabled. This has been added because there is an info log to show the uid and the login type used. If the user is disabled, that log will show and it might be misleading; the admin could think the disabled user has logged in successfully. This new log will clarify that such user is disabled.

Other than that, the unit tests showed some code duplication in the token login flow (expected only one call in some methods, but currently calling them twice). No plans to fix it for the moment in order to limit the changes, but we should look it up at some point.

lib/private/User/Session.php
```diff
@@ -530,38 +534,12 @@ public function tryBasicAuthLogin(IRequest $request) {
 	 * compatibility.
 	 */
 	private function loginWithPassword($login, $password) {
-		$beforeEvent = new GenericEvent(null, ['loginType' => 'password', 'login' => $login, 'uid' => $login, '_uid' => 'deprecated: please use \'login\', the real uid is not yet known', 'password' => $password]);
```
DeepDiver1975, line 533: This changes the order of events. Could have some side effects ...

DeepDiver1975: ... applies to all refactored methods

DeepDiver1975 (changes requested):

lib/private/User/Session.php
jvillafanez: We can try to use an "afterPreLogin" callback in the `loginInOwnCloud` method in order to move the checks. It will add additional complexity though, and it looks bad to have to use callbacks in the same class.
I don't think we have better options. The code paths are quite different and I had trouble trying to funnel the code into one method.

jvillafanez:
I think we'll have to take the risk of changing the behavior....

Currently the `loginInOwnCloud` method does:
```
trigger beforeLogin event
if the user is enabled
    set session user
    trigger afterLogin event
    prepare user login (initialize FS)
```

In order to keep the previous behavior the method should do
```
trigger beforeLogin event
run custom code (for apache login)
get user instance either via userManager's `get` or `checkPassword` (need a flag)
run custom code (for cookie login)
if the user is enabled
    set session user
    trigger afterLogin event
    prepare user login (initialize FS)
```
So we need a flag because the `loginInOwnCloud` method will need to use `checkPassword` when it comes from the loginWithPassword (likely due to LDAP), and we need 2 callbacks to inject code before and after getting the user.
Having to use a different method to get the user looks bad, but having to use callbacks to inject code in a private method seems to much spaghetti. Anyway, I don't think there are other options in order to keep the previous behavior.

-------

Regarding the change in the behavior, I think it sums up to when the events are triggered.

Previously, a pre-login event could be triggered followed by a failed-login event. The post-login event wouldn't trigger if a failed-login event happened. A wrong password attempt would trigger that event sequence.

With the current code, the login events are triggered quite late. It would be rare that something goes wrong after the pre-login event.
For the particular case of a wrong password, the current code would trigger only the failed-login event, but it doesn't trigger the pre-login event. The same behavior happens in many other situations.

In case of disabled users, some previous code triggered a failed-login event while other not. For now, the PR doesn't trigger the event, but we might need to change this.

I think this could be a good chance to homogenize the behavior with the events:
* If a failed-login event is triggered, there shouldn't be a pre-login event. This seems reasonable to me.
* If a pre-login event happens, all the required checks (at least the ones from core) should have passed
* If the user is disabled, a failed-login event should happen, but not a pre-login event (this is something we should adjust in the PR)
* A post-login event should happen after the user is set in the session and the FS is initialized (it seems reasonable that the FS initialization is part of the login process)

DeepDiver1975:
The only code using the preLogin event is in one code location and it blocks login of one specfic user.
Running this code in a later point in time has no further impact.

Let's move on :+1:

DeepDiver1975 (approved):

IljaN:
CI Fails

DeepDiver1975:
> CI Fails

windows smb -> known
webdav -> timeout -> just trigger it again

IljaN:
Already re-triggered once, will retry again

phil-davis:
I rebased to trigger fresh CI - we can see if anything unexpected fails.

phil-davis:
I rebased again - CI should be green now.

phil-davis:
CI is green. I will merge this. Overnight we will get CI results of all nightly oC10 apps. That will help to know if there are any app-related side-effects.

PR #40531, opened 2022-12-11
[full-ci] Implement persistent major version workflow (v2)

mrow4a:
## Description

_Remark: For reviews, please keep in mind that this code is many years old with lots of legacies, I tried as much as I can not to refactor the code too much in order to avoid regressions and fit this change into OC10 release according to semver._

This PR introduces the following enhancements:

- Change no. 1: Restore operation logic changed. Now restore is creating new current version of the file from one of past noncurrent versions of the file. Current version also receives incremented mtime for the file, and author of the files is the user that restored the file. The old noncurrent version is no longer removed upon restore and current version no longer receives mtime of the version.
- Change no. 2: The current version of the file is now shown in the Versions Tab, highlighted with "gray" background
 
Screenshot 2023-01-04 at 15 32 58

- Change no. 3: Versions now persist additional extended metadata on versioning tags, that allow easier identification of the versions. Each update increases minor version for the file

Screenshot 2023-01-04 at 15 34 50

- Change no. 4: Current version of the file now can be published, which increases major version tag. 
Screenshot 2023-01-04 at 15 37 46

- Change no. 5: Each new edit of the file would create noncurrent versions. The ones tagged with major version due to publishing, will be persisted long term and it wont be subject to any retention policies
Screenshot 2023-01-04 at 15 33 33

## Related Issue
- Replaces https://github.com/owncloud/core/pull/40319
- Implements https://github.com/owncloud/enterprise/issues/5286

## How Has This Been Tested?
- manually in the developer instance
  - [x] test with normal fs for file and folder
  - [x] test with normal fs for shared file file and folder
  - [x] test with trashbin for file and folder
  - [x] test with objectstore (it should be properly disabled)
  - [x] test with encryption (metafiles are plaintext, files/versions encrypted)
- unit/acceptance tests

Enable the feature and test generating new versions e.g. in shared folder on a document with TextEditor/OnlyOffice/Collabora
```
occ config:system:set file_storage.save_version_metadata --value true --type boolean
```

## Types of changes

## Checklist:
- [x] Base code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [x] Docs update https://github.com/owncloud/docs-server/pull/840 

## Example

- User `test1` creates a file `shared/nest/test1.txt`, version tab shows `current` label on the latest version
- User `test2` updates a file `shared/nest/test1.txt` two times, that creates two noncurrent versions
- User `test2` publishes latest version `shared/nest/test1.txt` , and thus would prevent expiry in the future. File is now marked with `persistent` label
- Two more edits happen that increment the version tag

Screenshot 2023-01-04 at 15 09 17


ownclouders:
:boom: Acceptance tests pipeline <strong>webUITags-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37402/143

ownclouders:

https://drone.owncloud.com/owncloud/core/37511/41

ownclouders:
:boom: Acceptance tests pipeline <strong>apiAuthWebDav-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37511/37

phil-davis:
https://drone.owncloud.com/owncloud/core/37534/9/7
```
There were 2 failures:

1) OCA\Files_Versions\Tests\VersioningTest::testExpire with data set "metaEnabled" (true)
Failed asserting that 3 matches expected 4.

/drone/src/apps/files_versions/tests/VersioningTest.php:354

2) OCA\Files_Versions\Tests\VersioningTest::testIsPublishedVersion with data set "metaEnabled" (true)

/drone/src/apps/files_versions/tests/VersioningTest.php:409
```

phil-davis:
https://drone.owncloud.com/owncloud/core/37534/143/18
```
webUITags/editTags.feature:15
webUITags/editTags.feature:28
```

```
    And the user edits the tag with name "random" and sets its name to "random-big" using the webUI # WebUITagsContext::theUserEditsTheTagWithNameAndSetsItsNameToUsingTheWebui()
      Page\FilesPageElement\DetailsDialog::renameTagxpath: //input[@id='view12-rename-input'] could not find tag edit input (SensioLabs\Behat\PageObjectExtension\PageObject\Exception\ElementNotFoundException)
```

phil-davis:
By default, CI gets killed as soon as one pipeline fails.
I suggest that you add `[full-ci]` to the PR title. Then all pipelines will run to completion and you can see all the things that are failing.

mrow4a:
@phil-davis would you mind having a look into https://drone.owncloud.com/owncloud/core/37551/142/17 ? I quite cannot scratch my head what is the problem, my change has nothing to do with Tags Tab, it touches only Versions Tab. Also backend logic nowhere touches tags logic. Would be quite grateful 🙏🏻

mrow4a:
@phil-davis started adjusting other acceptance tests due to the fact that PROPFIND now also returns current version and noncurrent version, plus restore logic changed. 

And... found out that in one case there is a bug in my implementation (`enable file versioning and check the history of changes from multiple users after renaming file by sharer`) ... without these tests would never have thought of that corner case, good job!

tests/acceptance/features/webUIFiles/versions.feature
```diff
@@ -140,14 +141,17 @@ Feature: Versions of a file
...
+    Then the authors of the current and noncurrent versions of file "randomfile.txt" should be:
       | index | author |
-      | 1     | Alice  |
-      | 2     | Brian  |
-      | 3     | Alice  |
+      | 1     | Carol  |
+      | 2     | Alice  |
+      | 3     | Brian  |
+      | 4     | Alice  |
     When the user restores the file to last version using the webUI
```
mrow4a, line 150: @phil-davis this case works perfectly when manually navigating, but somehow the restore here is wrongly catched as part of the test. I get error. Would appreciate any help here

```
    When the user restores the file to last version using the webUI                                         # WebUIFilesContext::theUserRestoresTheFileToLastVersionUsingTheWebui()
      element click intercepted: Element <a href="#" class="revertVersion" title="Restore">...</a> is not clickable at point (1016, 344). Other element would receive the click: <img src="/core/img/actions/checkmark.svg">
        (Session info: chrome=74.0.3729.169)
        (Driver info: chromedriver=74.0.3729.6 (255758eccf3d244491b8a1317aa76e1ce10d57e9-refs/branch-heads/3729@{#29}),platform=Linux 5.4.0-135-generic x86_64) (WebDriver\Exception\ElementClickIntercepted)
```

tests/acceptance/features/lib/FilesPageElement/DetailsDialog.php
```diff
@@ -67,7 +67,7 @@ class DetailsDialog extends OwncloudPage {
 	private $tagsDropDownResultXpath = "//div[contains(@class, 'systemtags-select2-dropdown')]" .
 	"//ul[@class='select2-results']" .
 	"//span[@class='label']";
-	private $tagEditInputXpath = "//input[@id='view12-rename-input']";
```
mrow4a, line 70: @phil-davis this was the problem with acceptance test for tags. I think this view13 should not be hardcoded. Lets not change this as part of this PR, should be separate I think

mrow4a:
@DeepDiver1975 @phil-davis ready for review, adjusted all tests (except one that has test code issue)

lib/public/Files/Storage/IVersionedStorage.php
```diff
@@ -29,26 +29,35 @@
...
 	/**
-	 * List all versions for the given file
+	 * Get metadata for current version of the file (versions root)
 	 *
 	 * @param string $internalPath
-	 * @return array
+	 * @return array metadata or null if not supported
+	 * @since 10.12.0
+	 */
+	public function getCurrentVersion($internalPath);
```
DeepDiver1975, line 38: not perfect method naming.

`getCurrentVersion` implies that the current version is returned.

What about merging this into `getVersion` where `$versionId` is null?

mrow4a: @DeepDiver1975 ye ok... what maybe can be done, with getVersions() - which is used in public interface in DAV and FS, I could add optional argument `includeCurrent/includeRoot/..?` , which will make it also backwards compatible.. thoughts?

apps/dav/lib/Meta/MetaPlugin.php
```diff
@@ -34,6 +36,8 @@ class MetaPlugin extends ServerPlugin {
 
 	public const VERSION_EDITED_BY_PROPERTYNAME = '{http://owncloud.org/ns}meta-version-edited-by';
 	public const VERSION_EDITED_BY_PROPERTYNAME_NAME = '{http://owncloud.org/ns}meta-version-edited-by-name';
```
DeepDiver1975, line 38: should be named `VERSION_EDITED_BY_NAME_PROPERTYNAME_NAME`

apps/dav/lib/Meta/MetaPlugin.php
```diff
@@ -98,12 +102,36 @@ public function handleGetProperties(PropFind $propFind, INode $node) {
...
-				return $node->getVersionAuthor();
+				return $node->getVersionEditedBy();
 			});
 			$propFind->handle(self::VERSION_EDITED_BY_PROPERTYNAME_NAME, function () use ($node) {
-				return $node->getVersionAuthorName();
+				$versionEditedBy = $node->getVersionEditedBy();
+				if (!$versionEditedBy) {
+					return '';
+				}
+				$manager = \OC::$server->getUserManager(); // FIXME: not so good
```
DeepDiver1975, line 129: this indeed sucks - some eager loading would be great - but out of scope

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -413,35 +414,48 @@ private static function retainVersions($filename, $owner, $ownerPath, $timestamp
 					$src = $owner . '/files_versions/' . $ownerPath;
 					$dst = $owner . '/files_trashbin/versions/' . \basename($ownerPath) . '.d' . $timestamp;
 					self::copy_recursive($src, $dst, $rootView);
-					if ($metaEnabled) {
```
DeepDiver1975, line 416: why was that removed?

apps/files_versions/lib/AppInfo/Application.php
```diff
@@ -24,18 +24,30 @@
...
 class Application extends App {
 	public function __construct(array $urlParams = []) {
 		parent::__construct('files_versions', $urlParams);
 
 		$container = $this->getContainer();
 
+		/**
+		 * Controllers
+		 */
+		$container->registerService('VersionController', function ($c) {
```
DeepDiver1975, line 44: should not be necessary - container should resolve the class by itself

apps/files_versions/lib/Controller/VersionController.php
```diff
@@ -0,0 +1,63 @@
...
+ * Class VersionController
+ *
+ * @package OCA\Files\Controller
+ */
+class VersionController extends Controller {
+	/**
+	 * @param string $appName
+	 * @param IRequest $request
+	 */
+	public function __construct($appName, IRequest $request) {
```
DeepDiver1975, line 42: not needed -> remove ctor

apps/files_versions/lib/Controller/VersionController.php
```diff
@@ -0,0 +1,63 @@
...
+	 * @NoAdminRequired
+	 *
+	 * @param string $path
+	 * @return JSONResponse
+	 */
+	public function publishVersion($path) {
+		try {
+			Storage::publishCurrentVersion($path);
+		} catch (\Exception $e) {
+			return new JSONResponse([], Http::STATUS_INTERNAL_SERVER_ERROR);
```
DeepDiver1975, line 58: log the exception

DeepDiver1975 (changes requested):

apps/files_trashbin/lib/Trashbin.php
mrow4a: the restore/retain from trashbin was wrong here (it was missing `.current.json` file in some scenarious). I fixed this one and also remove need for this operation, by implementing this handling properly in `apps/files_versions/lib/Storage.php`

apps/files_versions/lib/AppInfo/Application.php
mrow4a: done

apps/files_versions/lib/Controller/VersionController.php
mrow4a: good catch

tests/acceptance/features/webUIFiles/versions.feature
mrow4a: @phil-davis would be grateful if you could have a look on how to fix this.

apps/files_trashbin/lib/Trashbin.php
mrow4a: in fact, the block below probably could also have been hidden behind the storage, but did not want to refactor too much and did not find time to investigate how to do it properly.

mmattel:
`occ config:system:set file_storage.save_version_metadata --value true --type boolean`

Needs documentation in our admin docs, pls file an issue in the docs-ocis repo

mrow4a:
> `occ config:system:set file_storage.save_version_metadata --value true --type boolean`
> 
> Needs documentation in our admin docs, pls file an issue in the docs-ocis repo

It is not docs-ocis repo. In PR description I already pointed which doc needs update (chexkbox)

mmattel:
> It is not docs-ocis repo

this was just a moring-typo, already corrected 😄

apps/files_versions/lib/AppInfo/Application.php
```diff
@@ -24,6 +24,7 @@
 namespace OCA\Files_Versions\AppInfo;
 
 use OC\AllConfig;
+use OCA\Files_Versions\Controller\VersionController;
```
DeepDiver1975, line 27: not needed

DeepDiver1975 (changes requested):
- needs rebase again :see_no_evil: 
- what about s3 implementation of this?

DeepDiver1975:
rebasing ....

DeepDiver1975:
@JammingBen looks like we had the same intention at the same time :see_no_evil:

DeepDiver1975:
> * what about s3 implementation of this?

answered by looking into the tests - this is all not implemented on s3

DeepDiver1975:
@mmattel for docu: this feature (tagging major versions) is not implemented on s3.
this pr also changes the implementation of verisons ordering - this is anyhow only non-s3 relevant

tests/acceptance/features/webUIFiles/versions.feature
```diff
@@ -140,14 +141,53 @@ Feature: Versions of a file
...
       | index | author |
       | 1     | Carol  |
-      | 2     | Brian  |
+      | 2     | Carol  |
+      | 3     | Alice  |
+      | 4     | Brian  |
+      | 5     | Alice  |
+
+  @skipOnStorage:ceph @files_primary_s3-issue-67
+  Scenario: sharer can see the versions' respective author after version restore
```
mrow4a, line 160: @phil-davis so I wrote new scenario here, to prove scenario works, and that somehow in test **sharee can see the versions' respective author after version restore** the step `And the user restores the file to last version using the webUI Then the authors of the current and noncurrent versions of file "randomfile.txt" should be:` is buggy somehow

mrow4a:
@phil-davis I am also a bit fighting some other tests, as they hit race-conditions. The versions are based on mtime, which is in seconds. You could easily write a test that could have 2 different results based on timing (this is my current assumption)

tests/acceptance/features/bootstrap/FilesVersionsContext.php
```diff
@@ -154,6 +154,9 @@ public function userRestoresVersionIndexOfFile(string $user, int $versionIndex,
 			WebDavHelper::getDavPath($user, 2) . \trim($path, "/");
 		$fullUrl = $this->featureContext->getBaseUrlWithoutPath() .
 			$xmlPart[$versionIndex];
+			
+		\usleep(VERSION_MTIME_WAIT_TIMEOUT_MICROSEC); // make sure new version gets generated
```
mrow4a, line 158: @phil-davis ok I think I solved it with above

tests/acceptance/features/webUIFiles/versions.feature
mrow4a: @phil-davis ok... now this one fails, and other not.. it is like tossing a coin..

pako81:
@mrow4a we still have a failing test here https://drone.owncloud.com/owncloud/core/37761/124/17. Mind to take care, thx.

mrow4a:
> @mrow4a we still have a failing test here https://drone.owncloud.com/owncloud/core/37761/124/17. Mind to take care, thx.

@pako81 I need assistance from @phil-davis or @individual-it here, do not know how to continue there as it seems problem with a test itself and not a code.

phil-davis:
Now I can reproduce it - if I make the browser width less, then the "tick" icon that should be displayed for the current version, actually overlays the next version.
Screenshot from 2023-02-02 17-44-03

phil-davis:
And it only seems to happen while the time of the current version says "a few seconds ago - current"

If I wait a while, and it says "a minute ago - current" then it works OK. The test is shorter.

mrow4a:
@phil-davis cannot the selenium resolution be adjusted somehow?

phil-davis:
> @phil-davis cannot the selenium resolution be adjusted somehow?

I can manually reproduce the problem easily. So adjusting the screen size in the test will just hide the problem.

Can something be done in the UI code to make the "tick" stay against the correct "current" version?

mrow4a:
> Can something be done in the UI code to make the "tick" stay against the correct "current" version?

@phil-davis ok I will check, but if you use oc10.11 you might notice the problem was there actually already. But instead the restore buttons overlapped. Currentl OC10 design with frontend is not really resolution friendly.. I think we should define resolution where OC10 works 100% stable. Other resolutions can have a lot of functionalities malfunctioning .. check yourself :>

mrow4a:
@phil-davis ok.. I think I know where the problem is. It was not there before. It is because of the new "status" bar. I am weak in frontend, but will try to figure this out as probably would be hard to get capacity for this from someone fluent in it..

mrow4a:
@phil-davis I fixed the overlapping checkboxes. But something is still off there. Any clue what is going on now?

jvillafanez:
```
  Scenario: sharee can see the versions' respective author after version restore                            # /drone/src/tests/acceptance/features/webUIFiles/versions.feature:131
    Given the administrator has enabled the file version storage feature                                    # FeatureContext::theAdministratorHasEnabledTheFileVersionStorage()
    And user "Brian" has been created with default attributes and without skeleton files                    # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "Carol" has been created with default attributes and without skeleton files                    # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And user "Alice" has uploaded file with content "some content" to "/randomfile.txt"                     # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Alice" has shared file "randomfile.txt" with user "Brian"                                     # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Alice" has shared file "randomfile.txt" with user "Carol"                                     # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Brian" has uploaded file with content "Brian lorem content" to "/randomfile.txt"              # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Alice" has uploaded file with content "Alice lorem content" to "/randomfile.txt"              # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Carol" has uploaded file with content "Carol lorem content" to "/randomfile.txt"              # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Carol" has logged in using the webUI                                                          # WebUILoginContext::theUserHasLoggedInUsingTheWebUI()
    And the user has browsed to the files page                                                              # WebUIFilesContext::theUserBrowsesToTheFilesPage()
    When the user browses directly to display the "versions" details of file "randomfile.txt" in folder "/" # WebUIFilesContext::theUserBrowsesDirectlyToDetailsTabOfFileInFolder()
    Then the authors of the current and noncurrent versions of file "randomfile.txt" should be:             # WebUIFilesContext::theAuthorsOfCurrentAndNoncurrentVersionsOfFileShouldBe()
      | index | author |
      | 1     | Carol  |
      | 2     | Alice  |
      | 3     | Brian  |
      | 4     | Alice  |
INFORMATION: timed out waiting for ajax calls to start
    When the user restores the file to last version using the webUI                                         # WebUIFilesContext::theUserRestoresTheFileToLastVersionUsingTheWebui()
      │ INFORMATION: timed out waiting for ajax calls to start
    Then the authors of the current and noncurrent versions of file "randomfile.txt" should be:             # WebUIFilesContext::theAuthorsOfCurrentAndNoncurrentVersionsOfFileShouldBe()
      | index | author |
      | 1     | Carol  |
      | 2     | Carol  |
      | 3     | Alice  |
      | 4     | Brian  |
      | 5     | Alice  |
      WebUIFilesContext::theAuthorsOfCurrentAndNoncurrentVersionsOfFileShouldBe Expected author 'Carol King' at index '2' but got 'Alice Hansen'
      --- Expected
      +++ Actual
      @@ @@
      -'Carol King'
      +'Alice Hansen'
```
It seems a wrong expectation in the test. I've done the steps manually, and there is no restore action for the "current" (latest) version of the document. This means that it's impossible to restore Carol's version since it's the latest one, so a sequence of Carol - Carol - Alice - Brian - Alice can't happen.
It can also explain the timeout waiting in the test, because the test can't find the restore button.

Restoring the 2nd, 3rd, or 4th version (anything except the current version) also has a different behavior. As Carol, restoring Brian's version leads to a sequence Carol 0.5 - Alice 0.3 - Brian 0.2 - Alice 0.1 . I don't know if it's intended or it's a bug (I probably expect Carol 0.5 - Carol 0.4 - Alice 0.3 - Brian 0.2  Alice 0.1, and even Brian 0.5 instead of Carol 0.5)

mrow4a:
@jvillafanez it is incorrect, I modified the `tests/acceptance/features/lib/FilesPageElement/DetailsDialog.php` to take second entry as "last version", which is not a current one.

jvillafanez:
> as per expectation -> it is also incorrect, now restore is creating new version with the author being the one restoring. behaviour changed, as described in the changelog and description

This seems to be the problem. After restoring, the expected 0.4 version disappears. I only see versions 0.5 (current), 0.3, 0.2 and 0.1, with Carol, Alice, Brian and Alice as their respective authors.

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -413,35 +414,50 @@ private static function retainVersions($filename, $owner, $ownerPath, $timestamp
...
+					if ($owner !== $user || $forceCopy) {
+						$src = '/files_versions/' . $ownerPath . MetaStorage::CURRENT_FILE_PREFIX . MetaStorage::VERSION_FILE_EXT;
+						$dst = '/files_trashbin/versions/' . \basename($ownerPath) . MetaStorage::CURRENT_FILE_PREFIX . '.d' . $timestamp . MetaStorage::VERSION_FILE_EXT ;
+						$metaStorage->renameOrCopy('copy', $src, $owner, $dst, $owner);
+					}
+					if (!$forceCopy) {
+						$src = '/files_versions/' . $ownerPath . MetaStorage::CURRENT_FILE_PREFIX . MetaStorage::VERSION_FILE_EXT;
+						$dst = '/files_trashbin/versions/' . $filename . MetaStorage::CURRENT_FILE_PREFIX . '.d' . $timestamp . MetaStorage::VERSION_FILE_EXT;
+						$metaStorage->renameOrCopy('rename', $src, $owner, $dst, $user);
+					}
```
jvillafanez, line 437: @mrow4a could you check this? It seems wrong.

If `$owner` is different than `$user`, we copy the metadata. But then, if we aren't forcing a copy, we rename the metadata. This seems to match what I'm seeing and it could be the cause of the test failure.

mrow4a: @jvillafanez this is a trashbin function. Should not be related to trashbin. The acceptance tests that fails is working in the UI when testing manual.

mrow4a: also please note, that logic that I implemented is the same as was in the past. This code is not addition, it is replace+add

mrow4a:
> This seems to be the problem. After restoring, the expected 0.4 version disappears. I only see versions 0.5 (current), 0.3, 0.2 and 0.1, with Carol, Alice, Brian and Alice as their respective authors.

hmmm.... ok need to check this.. because maybe some retention policy triggered that is related to the fact that versions got created in close "mtime" to each other 🤔

jvillafanez:
I didn't change any setting other than the one to enable this feature. This means, no retention policy configured (other than the default one) and the quota for the users are unlimited. Moreover, there should be enough free space in the disk.
With these conditions, I don't think the retention policy should trigger and remove versions.

mrow4a:
@jvillafanez this is not true, read the doc again. The default retention is AUTO. This means after you reach 5 version retention is triggered for max versions per timeframe

mrow4a:
> The versions are adjusted along this pattern every time a new version is created and the background job was executed.

https://doc.owncloud.com/server/next/admin_manual/configuration/files/file_versioning.html#default-versions-delete-patterns

@jvillafanez

jvillafanez:
Assuming the retention policy triggers after restoring the file and it will never remove that file, we could try to disable the retention policy in the tests. I mean, if a previous version is removed from the list, that would still be ok.
The only reason to disable the retention is policy is to have a more predictable behavior.

I haven't checked, but I assume we have a test to ensure that persistent versions aren't removed by the retention policy. That would be the related use case we need to ensure.

jvillafanez:
I think I manually followed the test steps described in https://github.com/owncloud/core/pull/40531#issuecomment-1418967273 . The "When the user restores the file to last version using the webUI" was the confusing one, I think I restored Brian's version in that step.

mrow4a:
> I think I manually followed the test steps described in [#40531 (comment)](https://github.com/owncloud/core/pull/40531#issuecomment-1418967273) . The "When the user restores the file to last version using the webUI" was the confusing one, I think I restored Brian's version in that step.

you sure you refreshed browser caches so new javascript kicks in?

jvillafanez:
Yes, still happens with the latest code of this branch. Note that I'm doing the steps manually in the web UI. I'm not running the tests.

phil-davis:
Screenshot from 2023-02-07 17-07-13

I run the test scenario locally, putting a 30-second sleep in before DetailsDialog `restoreCurrentFileToLastVersion` so that the test "hangs" 30 seconds before it tries to find the top (latest) version revert button. That lets me look and take a screen shot.

The "tick" image that should be next to the current version, is actually over the top of the revert button of the latest version. The test code (and a real user) cannot effectively click the revert button.

phil-davis:
Screenshot from 2023-02-07 17-17-54

If I change versionstabview.js line 350 to
```
				currentVersionLabel: t('files_versions', 'c')
```
(use just the shorter string "c" rather than the full word "current")
then the "tick" is rendered in its proper place, and the test passes.

We really just need to do whatever to stop the "tick" from being rendered in the wrong place.

mrow4a:
@phil-davis what is the resolution in acceptance test?

phil-davis:
> @phil-davis what is the resolution in acceptance test?

Looking in the chrome console while the test is sleeping, it tells me about 1050*978 pixels for the browser tab.

I see where `tests/acceptance/run.sh` can try to set `screenResolution` - but I think that was for choosing an actual OS screen size back in the days when we ran UI tests with SauceLabs. I tried changing those numbers, and it does not look like it has any effect.

phil-davis:
```
MOBILE_RESOLUTION='1200x800' make test-acceptance-webui BEHAT_FEATURE=tests/acceptance/features/webUIFiles/versions.feature:131
```

I found where we made the option to specify the browser window size. There is code in WebUIGeneralContext.php `setUpScenario()` which checks for a definition of `MOBILE_RESOLUTION` and sets the browser size to that. Normally we use it for running some test scenarios on a smaller size.

If I set the size like in the example command above, then the test passes. If it is set to a small width, then the test fails.

We could make a test step that resizes the browser window to a larger size, and use that in this scenario for now. (And make a separate issue to investigate and fix the problem that happens when the browser width is less)

jvillafanez:
This patch might work.
```
diff --git a/apps/files_versions/css/versions.css b/apps/files_versions/css/versions.css
index 2fe7954e67..570bedbee4 100644
--- a/apps/files_versions/css/versions.css
+++ b/apps/files_versions/css/versions.css
@@ -9,6 +9,12 @@
        float: left;
        border-bottom: 1px solid rgba(100,100,100,.1);
 }
+
+.versionsTabView li > div {
+       display: flex;
+       height: 100%;
+}
+
 .versionsTabView li:last-child {
        border-bottom: none;
 }
@@ -33,14 +39,13 @@
 }
 
 .versionsTabView .preview-container {
-       display: inline-block;
+       display: flex;
+       flex-direction: column;
        vertical-align: top;
        text-align: center;
-}
-
-.versionsTabView .preview-container {
        padding: 5px;
 }
+
 .versionsTabView img {
        cursor: pointer;
        padding-right: 4px;
@@ -56,10 +61,19 @@
 }
 
 .versionsTabView .version-container {
-       display: inline-block;
+       display: inline-flex;
+       flex-direction: column;
+       flex: auto;
+       overflow: auto;
        margin-top: 5px;
 }
 
+.versionsTabView .version-container .version-headline {
+       overflow: hidden;
+       text-overflow: ellipsis;
+       white-space: nowrap;
+}
+
 .versionsTabView .versiondate {
        min-width: 100px;
        vertical-align: super;
@@ -84,18 +98,17 @@
 }
 
 .versionsTabView .action-container {
-       display: inline;
+       display: flex;
+       flex: 0 0 auto;
 }
 
 .versionsTabView .revertVersion {
        cursor: pointer;
-       float: right;
        margin-right: -10px;
 }
 
 .versionsTabView .publishVersion {
        cursor: pointer;
-       float: right;
        margin-right: -10px;
 }
 .versionsTabView .current-version .version-headline {
```
At least, the button isn't hidden in small screens, so it should be clickable in the tests. Looks good to me.
Someone with more CSS expertise could double-check in case there is something wrong with it.

mrow4a:
@DeepDiver1975 @jvillafanez @phil-davis finally CI is green.

I decided not to change the existing logic with CSS to avoid risk related to regressions. Instead, I only show the new current/persistent label if screensize is >1200px in width. Commit is https://github.com/owncloud/core/pull/40531/commits/5cc9921b1682055bd20030af11dfdd23a1e38b96 , and FIXME tag added. Long term, it would be good to fix it for smaller screens too, and thus also to allow acceptance tests on these labels. But in this PR it will be too big/risky.

apps/files_versions/lib/MetaStorage.php
```diff
@@ -223,23 +196,38 @@ public function deleteCurrent(View $versionsView, string $filename) {
...
+		$restoreVersionTag = $restoreMetaData['restored_from_tag'] ?? '';
+		if (!$restoreVersionTag) {
+			$restoreVersionTag = $restoreMetaData['version_tag'] ?? '';
 		}
+
+		// increment version tag for current meta
+		$oldCurrentVersionTag = $currentMetaData['version_tag'] ?? '';
+		$newCurrentVersionTag = $this->incrementVersionTag($oldCurrentVersionTag, true);
+			
+		// create new currentMetaFile and remove restoreMetaFile
```
jvillafanez, line 224: I don't see we're removing the restoreMetaFile :)
We might need to add in the PHPDoc that the restored metadata isn't removed in order to clarify the behavior

apps/files_versions/lib/MetaStorage.php
```diff
@@ -303,6 +291,36 @@ public function isObjectStoreEnabled(): bool {
...
+	/**
+	 * Get the incremented version tag for a new version, which is
+	 * latest version +0.1 or new major version.
+	 *
+	 * @param string $oldVersionTag
+	 * @param bool $minor if true increases minor version, otherwise major
+	 * @return string
+	 */
+	private function incrementVersionTag(string $oldVersionTag, bool $minor) : string {
+		if (!$oldVersionTag) {
```
jvillafanez, line 303: `$oldVersionTag === ''` seems more natural

apps/files_versions/lib/Storage.php
```diff
@@ -353,6 +390,13 @@ public static function renameOrCopy($sourcePath, $targetPath, $operation) {
 			// create missing dirs if necessary
 			self::getFileHelper()->createMissingDirectories(new View("/$targetOwner"), $targetPath);
 
+			if (self::metaEnabled()) {
+				// Also move/copy the current version
+				$src = '/files_versions/' . $sourcePath . MetaStorage::CURRENT_FILE_PREFIX . MetaStorage::VERSION_FILE_EXT;
+				$dst = '/files_versions/' . $targetPath . MetaStorage::CURRENT_FILE_PREFIX . MetaStorage::VERSION_FILE_EXT;
+				self::$metaData->renameOrCopy($operation, $src, $sourceOwner, $dst, $targetOwner);
+			}
```
jvillafanez, line 398: The code might need to be moved.
If you have files with metadata and any of the parent folders are moved or copied, the metadata of the files inside that folder need to be moved.

lib/private/Files/Meta/MetaVersionCollection.php
```diff
@@ -51,6 +55,56 @@ public function __construct(IRootFolder $root, \OCP\Files\Node $node) {
...
+
+	public function getPath() {
+		return "/meta/{$this->getId()}/v";
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function getEditedBy() : string {
+		$storage = $this->node->getStorage();
```
jvillafanez, line 70: we should be accessing the `$this->versionInfo` cache first.
```
if ($this->versionInfo) {
  return $this->versionInfo['edited_by'] ?? '';
}

$storage = $this->node->getStorage();
........
```

jvillafanez:
Taking into account that the user can't remove versions, I think we should have some condition to remove published versions. We might also need to check the performance of the expiration if there are quite a bunch of published versions.

We could remove published versions if there are only published versions, and starting from the oldest one. It seems in line with the current retention policy.

mrow4a:
@jvillafanez I will try to address the inputs, but some of them would need to discard. 

This change is large, and to avoid regressions I would prefer not to refactor too much or change existing logic. This code is many years old.

jvillafanez:
Yes, I'm ok with that. We'll have to be aware of possible issues, and be clear about what is left to do.

apps/files_versions/js/versionstabview.js
```diff
@@ -150,26 +206,62 @@
...
+
+					self.collection.setFileInfo(fileInfoModel);
+					self.collection.reset([], {silent: true});
+					self.collection.fetch();
+
+					self.$el.find('.versions').removeClass('hidden');
+				},
+				error: function() {
+					self.$el.find('.versions').removeClass('hidden');
+					self._toggleLoading(false);
```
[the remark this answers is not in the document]
mrow4a: In fact I copied it from `versionModel.revert({` which is "related" button, just different function. I do not dare to change this as it is not documented why needed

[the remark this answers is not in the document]
jvillafanez: ok, then maybe write a TODO / FIXME to investigate later.

apps/files_versions/lib/MetaStorage.php
```diff
@@ -83,115 +82,89 @@ public function __construct(string $dataDir, FileHelper $fileHelper) {
...
+		$userView = $this->fileHelper->getUserView($uid);
+		$this->fileHelper->createMissingDirectories($userView, $currentFileName);
 
-		list($owner, $fileName) = Storage::getUidAndFilename($currentFileName);
+		// get metadata for old current file (if exists)
+		$oldMetadata = $this->readMetaFile($absPathOnDisk);
 
-		$userPath = "/files_versions$fileName" . MetaStorage::CURRENT_FILE_EXT;
+		// generate Version Edited By property
+		$newVersionEditedBy = \OC_User::getUser();
```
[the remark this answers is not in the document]
mrow4a: even if it is, the editor will be false here, and not displayed.

apps/files_versions/lib/MetaStorage.php
```diff
@@ -83,115 +82,89 @@ public function __construct(string $dataDir, FileHelper $fileHelper) {
...
-	 * @return string|null
-	 */
-	protected function getVersionAuthorByPath(string $path) : ?string {
-		$json = \file_get_contents($path);
-		if ($decoded = \json_decode($json, true)) {
-			if (isset($decoded[MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME])) {
-				return $decoded[MetaPlugin::VERSION_EDITED_BY_PROPERTYNAME];
-			}
+		if (\file_exists($currentMetaFile)) {
+			@\copy($currentMetaFile, $targetMetaFile);
```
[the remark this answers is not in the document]
mrow4a: I think this is something for follow up work. It would make sense to change it literally everywhere in this class

apps/files_versions/lib/MetaStorage.php
```diff
@@ -223,23 +196,38 @@ public function deleteCurrent(View $versionsView, string $filename) {
...
+		$currentMetaData = $this->readMetaFile($currentMetaFile);
+		$restoreMetaData = $this->readMetaFile($restoreMetaFile);
 
-		if (\file_exists($src)) {
-			\rename($src, $dst);
-		} elseif (\file_exists($dst)) {
-			// Remove current author file in case there is no author to restore
-			\unlink($dst);
+		// restored version is technically a new version generated by a user
+		$newVersionEditedBy = \OC_User::getUser();
```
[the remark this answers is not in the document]
mrow4a: same as before answered

apps/files_versions/js/versionstabview.js
```diff
@@ -198,9 +302,10 @@
 			}
 		},
 
-		_formatItem: function(version) {
+		_formatVersion: function(version) {
 			var timestamp = version.get('timestamp') * 1000;
 			var size = version.has('size') ? version.get('size') : 0;
+			var isMajorVersion = version.has('versionTag') && version.get('versionTag').indexOf('.0', version.get('versionTag').length - '.0'.length) !== -1;
```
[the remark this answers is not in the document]
mrow4a: should be in the documentation and extended. ref https://doc.owncloud.com/server/next/admin_manual/configuration/files/file_versioning.html#conscious-persistent-major-versions

apps/files_versions/js/versionstabview.js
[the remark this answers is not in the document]
mrow4a: Ye I mean.. this code I guess is just to endure loading is disabled. Just in case some logic interferes. I would keep it just as safety (legacy) net

config/config.sample.php
```diff
@@ -715,11 +715,17 @@
...
+'file_storage.save_version_author' => false,
+
+/**
+ * Save additional metadata (author, version tag, etc.) of each version of uploaded and edited files.
  *
+ * WARNING: This feature CANNOT be temporarily disabled once enabled. Consequitive enabling would require repair job that erases all extended versions metadata.
  * WARNING: This does not work for S3 storage backends.
  */
-'file_storage.save_version_author' => false,
+'file_storage.save_version_metadata' => false,
```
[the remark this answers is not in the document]
mrow4a: done

apps/files_versions/js/versionstabview.js
```diff
@@ -150,26 +206,62 @@
...
 		},
 
+		_onClickPublishVersion: function(ev) {
+			var self = this;
+			var fileInfoModel = this.versionsRoot.getFileInfo();
+			ev.preventDefault();
+
+			this.versionsRoot.publish({
+				success: function() {
+					// reset and re-fetch the updated collection
```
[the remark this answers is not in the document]
mrow4a: commented

apps/files_versions/lib/MetaStorage.php
```diff
@@ -83,115 +82,89 @@ public function __construct(string $dataDir, FileHelper $fileHelper) {
...
+	 * @return array metadata
+	 * @throws \Exception
 	 */
-	public function moveCurrentToVersion(string $currentFileName, FileInfo $versionFile, string $uid) {
-		$currentMetaFile = $this->pathToAbsDiskPath($uid, "/files_versions/$currentFileName" . self::CURRENT_FILE_EXT);
-		$targetMetaFile = $this->dataDir . $versionFile->getPath() . self::VERSION_FILE_EXT;
-
-		if (\file_exists($currentMetaFile)) {
-			@\rename($currentMetaFile, $targetMetaFile);
+	public function getVersion(FileInfo $versionFile) : array {
```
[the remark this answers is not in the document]
mrow4a: done

apps/files_versions/lib/MetaStorage.php
```diff
@@ -223,23 +196,38 @@ public function deleteCurrent(View $versionsView, string $filename) {
...
 
-		if (\file_exists($src)) {
-			\rename($src, $dst);
-		} elseif (\file_exists($dst)) {
-			// Remove current author file in case there is no author to restore
-			\unlink($dst);
+		// restored version is technically a new version generated by a user
+		$newVersionEditedBy = \OC_User::getUser();
+
+		// preserve original restoration tag if set
```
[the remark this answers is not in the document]
mrow4a: removed

apps/files_versions/lib/MetaStorage.php
mrow4a: done

apps/files_versions/lib/MetaStorage.php
mrow4a: done

apps/files_versions/lib/Storage.php
mrow4a: commented in the code

apps/files_versions/js/versionstabview.js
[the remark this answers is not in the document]
jvillafanez: I'm not sure if that we'll be enough. I mean, we might let the user to set the versioning himself in the future, and we might assume that versions such as "1.0.1" will work. It won't be just allowing the user to set the versioning, but there is more code to be done.
I'm not saying that we have to code something for it right now, but I'd like to have something so we are reminded about this if we ever need to implement that use case. A well placed comment in the code is enough for me.

apps/files_versions/lib/MetaStorage.php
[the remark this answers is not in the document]
jvillafanez: I'd remove the "@" for now so it's the same for all the methods. For the actual error handling, I think it's ok for a follow up (we'd need a ticket :) )

jvillafanez (approved):
Might need a squash, but code looks ok.

mmattel:
@mrow4a pls add the change that you proposed in docs in config_sample_php_parameters here in core. I will then do a `config-to-docs` run post merging this PR which transports the changes to docs properly.

DeepDiver1975 (approved):

jnweiger:
Confirmed fixed in 10.12.0-rc.3
- Persistent versions are retained with auto, other versions disapear as per policy.
- Persistent versions older than 2 days are retained with 'auto, 2', when `occ versions:expire` is run. 
- Persistent versions are deleted by 'occ versions:cleanup' - which is expected and documented behaviour.
- Works fine also with encryption.

Issue #39796, opened 2022-02-16
Refactor API tests that drill down into XML or JSON structures


phil-davis:
There are test scenarios like:
```
  Scenario: getting default capabilities with admin user with new values
    When the administrator retrieves the capabilities using the capabilities API
    Then the capabilities should contain
      | capability    | path_to_element                                          | value             |
      | files_sharing | user@@@expire_date@@@enabled                             | EMPTY             |
      | files_sharing | group@@@expire_date@@@enabled                            | EMPTY             |
      | files_sharing | providers_capabilities@@@ocinternal@@@user@@@element[0]  | shareExpiration   |
      | files_sharing | providers_capabilities@@@ocinternal@@@group@@@element[0] | shareExpiration   |
      | files_sharing | providers_capabilities@@@ocinternal@@@link@@@element[0]  | shareExpiration   |
      | files_sharing | providers_capabilities@@@ocinternal@@@link@@@element[1]  | passwordProtected |
```
https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiCapabilities/capabilities.feature

The `path_to_element` has dependency on referencing explicit numeric array indices `[0]` `[1]` etc. And the "@@@" separator is not very nice to read.

1) refactor the code so that it can look for and find the data anywhere in the numeric-indexed arrays

2) implement a nicer way of referring to the nested array structure than having those "@@@" in the scenario steps.

Related issue in oCIS: https://github.com/owncloud/ocis/issues/3179

The solution should be similar in both repos, so that scenarios will read consistently.

SagarGi:
TODO:

- [ ] apiCapabilities https://github.com/owncloud/core/pull/40619

similar to the above PR. Search `@@@` in tests of core repo and refactor it as above PR.

saw-jan:
>similar to the above PR. Search `@@@` in tests of core repo and refactor it as above PR.

I see that only this [adminSharingSettings.feature](https://github.com/owncloud/core/blob/master/tests/acceptance/features/webUIAdminSettings/adminSharingSettings.feature) feature has `@@@` and that is used to check only single configuration value that is set using webUI.

IMO, we don't have to implement json assertion steps there because they are webUI tests and we are only confirming the webUI action using some api requests (that can be requests to production endpoints or testing endpoints). I see no value in implementing response json-assertion in those scenarios. So I can only suggest two things:
1. Leave them as it is
2. Rename then steps so that they don't use `@@@`

[...]

PR #40562, opened 2023-01-04
[tests-only] Refactor drilling of the JSON responses for improved assertions

kiranparajuli589:
## Description
Added steps to compare (exact match or contains in the actual response) xml API responses to be asserted as JSON.

Why choosing JSON over XML:
- Sharing API supports JSONs
- Easier for assertions

How to get JSON from XML:
- we can set the query param `format=json` while doing the request.
- If all the endpoint does not support JSON format, this site could be helpful for the conversion: https://codebeautify.org/xmltojson

## Related Issue

- #39796 

## Motivation and Context
Better assertions for deep values

## How Has This Been Tested?

- :robot: 

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [x] Acceptance tests added


tests/acceptance/features/bootstrap/CapabilitiesContext.php
```diff
@@ -39,6 +40,392 @@ class CapabilitiesContext implements Context {
...
+	 * Converts a XML response into an array
+	 *
+	 * @param SimpleXMLElement $xml
+	 *
+	 * @return array
+	 */
+	public function simpleXMLElementToArray(SimpleXMLElement $xml): array {
+		$json = json_encode($xml);
+		return (array)json_decode($json, true);
+	}
```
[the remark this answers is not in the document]
saw-jan: But other contexts can access it through step def. So, you decide

tests/acceptance/features/apiCapabilities/capabilities.feature
```diff
@@ -37,57 +37,94 @@ Feature: capabilities
...
+          },
+          "group": {
+            "expire_date": {
+              "enabled": {}
+            }
+          },
+          "providers_capabilities": {
+            "ocinternal": {
+              "user": {
+                "element": "shareExpiration"
```
saw-jan, line 111: Is key `element` the property in the response or just the special one?
If it is some special key then we can name it something like this in order to separate it from actual response properties. What do you think?
```suggestion
                "__element__": "shareExpiration"
```

tests/acceptance/features/apiCapabilities/capabilities.feature
```diff
@@ -37,57 +37,94 @@ Feature: capabilities
...
+              },
+              "group": {
+                "element": "shareExpiration"
+              },
+              "link": {
+                "element": [
+                  "shareExpiration",
+                  "passwordProtected"
+                ]
```
saw-jan, line 120: accepting both cases? or just can do like this. But just a suggestion, the choice is yours
```suggestion
              "group": {
                "elements": ["shareExpiration"]
              },
              "link": {
                "elements": [
                  "shareExpiration",
                  "passwordProtected"
                ]
```

tests/acceptance/features/apiCapabilities/capabilities.feature
saw-jan: or maybe this can be a server thing. My suggestion will be void if it is so

tests/acceptance/features/apiCapabilities/capabilities.feature
saw-jan: or maybe this can be a server thing. My suggestion will be void if it is so

tests/acceptance/features/bootstrap/CapabilitiesContext.php
```diff
@@ -39,6 +40,392 @@ class CapabilitiesContext implements Context {
...
+				"\n- Expected regex: '" . $needle . "'\n" .
+				"\n+ Actual array: '" . \json_encode($actualArray) . "'"
+			);
+		} else {
+			$expectedSubstitutedValue = \is_string($needle)
+				? $this->featureContext->substituteInLineCodes($needle)
+				: (string) $needle;
+
+			foreach ($actualArray as $actualValue) {
+				if ((string)$expectedSubstitutedValue === (string) $actualValue) {
```
saw-jan, line 212:
```suggestion
				if ((string)$expectedSubstitutedValue === (string)$actualValue) {
```

tests/acceptance/features/apiCapabilities/capabilities.feature
kiranparajuli589: `element` is not manually provided but will be present in the response xml. this is used show list. for eg:
```xml
<somekey>
  <element>one</element>
  <element>two</element>
</somekey>
```
if the xml is like this, then the PHP encoded json form will be like: `{"somekey": {"element": ["one", "two"]}`. if the parsed output was like:
```json
`{"somekey": ["one", "two"]}`
```
the assertion would have been lot easier. Most of the online, xml->json converter will give the output with the key element and values as array _(will be easier to make the pystring to dump in the the feature step)_.
the new comparision step will support both type of input pattern. If we just copy paste the converted xml to feature file, that will be valid for the assertion and if we construct the pystring manually, with list instead of `element => list`, that will be valid for the assertion too. It's because:
inside the assertion implementation, the php->json_encode will always give o/p with `element`. So while asserting array, the input and outputs are modified to the parallel form first and compared after that.

kiranparajuli589: Screenshot from 2023-01-09 14-33-31

saw-jan: This will be an extra effort. I thought that `element` was provided for special matching. Since it comes by default from the response, your implementation is good to go.

tests/acceptance/features/apiCapabilities/capabilities.feature
saw-jan: it comes from server response so not needed

tests/acceptance/features/apiCapabilities/capabilities.feature
saw-jan: Since that comes from server response, why bother tweaking the response.

tests/acceptance/features/apiCapabilities/capabilities.feature
```diff
@@ -37,57 +37,94 @@ Feature: capabilities
...
-      | files_sharing | share_with_group_members_only             | EMPTY             |
-      | files_sharing | share_with_membership_groups_only         | EMPTY             |
-      | files_sharing | auto_accept_share                         | 1                 |
-      | files_sharing | user_enumeration@@@enabled                | 1                 |
-      | files_sharing | user_enumeration@@@group_members_only     | EMPTY             |
-      | files_sharing | user@@@send_mail                          | EMPTY             |
-      | files         | bigfilechunking                           | 1                 |
-      | files         | privateLinks                              | 1                 |
-      | files         | privateLinksDetailsParam                  | 1                 |
+    Then the data of the response data should contain
```
grgprarup, line 40:
```suggestion
    Then the data of the response should contain
    OR
    Then the response data should contain
```
Can't we omit redundant term data?

tests/acceptance/features/bootstrap/CapabilitiesContext.php
```diff
@@ -39,6 +40,392 @@ class CapabilitiesContext implements Context {
...
+	 * @return array array of values of the provided array
+	 */
+	public function getArrayValues(array $arr): array {
+		$values = [];
+		if (\array_keys($arr) === ["element"]) {
+			if (\is_array($arr["element"])) {
+				$values = $arr["element"];
+			} else {
+				$values = [$arr["element"]];
+			}
```
grgprarup, line 131:
```suggestion
			$values = (\is_array($arr["element"])) ? $arr["element"] : [$arr["element"]];
```
Prefer to use ternary op.

tests/acceptance/features/bootstrap/CapabilitiesContext.php
```diff
@@ -39,6 +40,392 @@ class CapabilitiesContext implements Context {
...
+			// if the types of expected and actual values are different, then the assertion fails right away
+			$this->assertTypeEquals($value, $actual[$key], $key);
+
+			if (\is_array($value)) {
+				if ($this->isMultiDimensionalArray($value)) {
+					// if the value is an array, we need to check it recursively
+					$this->assertArrayEquals($value, $actual[$key]);
+				} else {
+					$this->assert1DArrayEquals($value, $actual[$key], $key);
+				}
```
grgprarup, line 388:
```suggestion
                                // if the value is an array, we need to check it recursively
				($this->isMultiDimensionalArray($value)) ? $this->assertArrayEquals($value, $actual[$key]) : $this->assert1DArrayEquals($value, $actual[$key], $key);
```

kiranparajuli589:
@saw-jan @grgprarup I've addressed the reviews. Please have a look again.

saw-jan (commented):
LGTM :+1:

How about naming?
`AArray.php` -> `ArrayUtils.php`
`AssertArrayContext.php` -> `ResponseContext.php`
or any other names.

phil-davis (commented):
This looks good to me.
We need to double-check that if the response and the expected structures are different, that the test step fails, and gives some sensible idea of what the difference is. I can play with that locally to verify.

phil-davis:
There is quite a long list of scenarios (in capabilities and anywhere else that has the old `@@@` thing used)
After we agree on the new way of doing it, do we:
a) merge this PR and then make another PR that adjusts more scenarios (and removes the "old way with `@@@`)
or
b) add all the scenario adjustments and old code removal to this PR.
>

individual-it:
this is a lot of code to maintain, what about using https://github.com/martin-helmich/phpunit-json-assert/ ?

The step might then look like this:
```
Then the JSON data of the response should match
"""
{
  "type": "object",
  "required": [
    "capabilities"
  ],
  "properties": {
    "capabilities": {
      "type": "object",
      "required": [
        "core",
        "files",
        "files_sharing"
      ],
      "properties": {
        "core": {
          "type": "object",
          "required": [
            "pollinterval",
            "webdav-root",
            "status"
          ],
          "properties": {
            "pollinterval": {
              "type": "string",
              "pattern": "^30000$"
            },
            "webdav-root": {
              "type": "string",
              "pattern": "^remote\\.php\\/webdav$"
            },
            "status": {
              "type": "object"
            }
          }
        }
      }
    }
  }
}
"""

ScharfViktor:
> this is a lot of code to maintain, what about using https://github.com/martin-helmich/phpunit-json-assert/ ?
it looks really cool. we can check the structure and the response body. I would like it to be in ocis

SagarGi:
> this is a lot of code to maintain, what about using https://github.com/martin-helmich/phpunit-json-assert/ ?
> 
> The step might then look like this:
> 
> ```
> Then the JSON data of the response should match
> """
> {
>   "type": "object",
>   "required": [
>     "capabilities"
>   ],
>   "properties": {
>     "capabilities": {
>       "type": "object",
>       "required": [
>         "core",
>         "files",
>         "files_sharing"
>       ],
>       "properties": {
>         "core": {
>           "type": "object",
>           "required": [
>             "pollinterval",
>             "webdav-root",
>             "status"
>           ],
>           "properties": {
>             "pollinterval": {
>               "type": "string",
>               "pattern": "^30000$"
>             },
>             "webdav-root": {
>               "type": "string",
>               "pattern": "^remote\\.php\\/webdav$"
>             },
>             "status": {
>               "type": "object"
>             }
>           }
>         }
>       }
>     }
>   }
> }
> """
> ```

TODO:

SagarGi:
i will be over taking this issue.

SagarGi:
Since we have found an alternative to it with this PR https://github.com/owncloud/core/pull/40619. This PR can be closed.

PR #40567, opened 2023-01-09
[full-ci] Prevent creation of empty files/folders for accounts having no available quota

pako81:
## Description
Prevent creation of empty files/folders for accounts having no available quota

## Related Issue
- https://github.com/owncloud/enterprise/issues/5478

## Motivation and Context
Until now it was possible for users having 0 quota or who already reached the limit of their assigned quota to still create empty files/folders, which generates confusion. The PR fixes this behaviour.

## How Has This Been Tested?
Manually:

Test 1. :

- as admin user `admin` assign quota `0 B` to user `user1`
- login as user `user1` over WebUI and try to create empty files or folders: an error `Could not create..` is returned

Test 2. :

- as `user1` who has a valid quota assigned create a folder `for guests` and share it with full permissions with guest user `guest@owncloud.com`
-  as admin user `admin` assign now quota `0 B` to user `user1`
- login as guest user `guest@owncloud.com`, enter the folder `for guests` and try to create empty files or (sub)folders: an error `Could not create..` is returned

Test 3. : 

- Same steps as Test 1. and 2. but this time with `user1` having i.e. 10MB quota assigned and already using 100% of the available space

Test 4. :

- as admin user `admin` assign quota `0 B` to user `user1`
- connect `user1` via desktop client to his personal storage: sync client returns a 403 Forbidden to `MKCOL` and `PUT` operations when trying to respectively create empty folders and files.

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Changelog item


pako81:
Actually using `==` as per https://github.com/owncloud/core/pull/40505#pullrequestreview-1208745332 does not seem to make this work correctly.

ownclouders:
:boom: Acceptance tests pipeline <strong>apiFilesExt-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37568/89

jvillafanez:
> Actually using === as per https://github.com/owncloud/core/pull/40505#pullrequestreview-1208745332 does not seem to make this work correctly.

Double-check the type because the `free_space` method in https://github.com/owncloud/core/blob/c4f8cd14fb2761cca93fba53199a7595cdb80e70/lib/private/legacy/helper.php#L613 could return false.
It's probably better to handle the `false` case there, before the calculations.

pako81:
From what I can see, this is rather because https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Local.php#L314 returns `float` (verified by using `gettype` on https://github.com/owncloud/core/blob/master/lib/private/legacy/helper.php#L613).

So casting it to `int` seems to make this work correctly when using `===`. Not sure this is the right approach however.

jvillafanez:
The interface says it should return either an int or false, so if it's returning a float then, technically, it's a bug in the implementation.
This means that the local storage is responsible of converting the `disk_free_space` result to int to be compliant with the interface.

I don't know why the `disk_free_space` returns a float and not an int, but we might need to take care of the conversion to int. Probably something like:
```
$space = @\disk_free_space($sourcePath);
if ($space === false || $space === null) {
  return \OCP\Files\FileInfo::SPACE_UNKNOWN;
}
$space = \intval($space);
if ($space === PHP_INT_MAX) {
  return \OCP\Files\FileInfo::SPACE_UNKNOWN;
}
return $space;
```

The alternative would be to change the interface so it returns a float instead of an int, but we'll have to check that using a float is properly supported in all the places where this method is used. Seems a lot of work.

pako81:
I was actually wrong, the float (double) type seems rather to come from `$this->quota` in https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Wrapper/Quota.php#L96. I've therefore added the conversion to int there.

jvillafanez:
The problem might not be really in the wrapper but something inside. When the wrapper is created, the quota is gotten from https://github.com/owncloud/core/blob/e4c8e50a55b30c17ffc0688d6e308bf2d06a3829/lib/private/legacy/util.php#L375 , and the `computerFileSize` could be a float (the returning type hint of the `computerFileSize` method is also wrong because it can return float, int and false).

Note that the problem with the `disk_free_space` function is still present.

At this point, it might be easier to use `==` instead and write a comment pointing to all these issues we're finding. I don't think we're using strings or other data types around there that could mess up with the behavior.

jvillafanez (approved):

jvillafanez:
Need to check what happens with the tests

IljaN:
I wonder if this could have potential side-effects/edge-cases:
- Creating user home after first login
- Creating skeleton files

IljaN:
> However, this seems indeed to break the skeleton files creation as folders' structure gets correctly created but with empty files inside, which looks bad.

This would be okay as this is also the current behaviour. I was fearing that the exception might interrupt the process, but I guess it's fine if it is implemented in the DAV-Layer.

jvillafanez:
I think the "block" only happens when accessing through webdav because the check are placed in our dav connector. Access through other channels (view, nodes, storage, etc) would still allow to create folders. However, the file quota applies at stream level.

Basically, when the skeleton files are copied, the folders are created because we aren't accessing through webdav, but the files can't be written due to the quota being applied in the stream.

I assume this is an acceptable behavior. Note that apps could still create folders by themselves in the FS following the same mechanism that core is using.

IljaN (approved):

pako81:
Changed it to return `Sabre\DAV\Exception\InsufficientStorage` to be consistent with clients behaviour.

pako81:
@jvillafanez do you maybe have an idea why https://github.com/owncloud/core/blob/master/tests/acceptance/features/apiMain/external-storage.feature#L89 is currently failing? AFAICT uploading to external storages when no quota is set should still work, so not sure why the test expects now a 507 instead of a 201.

jvillafanez:
I think it's because of the new check you've set for the file.
I assume that it worked before because the quota was controlled in the stream, and for external storages maybe the quota wrapper isn't set. With your new check, you might be uploading a 5-bytes file having a quota of 1 byte, and your check applies to all storages, including external ones.

phil-davis:
This branch is [10 commits ahead](https://github.com/pako81/core/compare/owncloud:core:master...prevent_empty_folders), [100 commits behind](https://github.com/pako81/core/compare/prevent_empty_folders...owncloud:core:master) owncloud:master.

https://github.com/pako81/core/tree/prevent_empty_folders

This is now a long way behind master - it would be good to rebase and get fresh CI results. I put `full-ci` in the PR title so that a full set of results will come when you rebase.

phil-davis:
https://drone.owncloud.com/owncloud/core/37767/41/13
```
  Scenario: user cannot create file on shared folder by a user with zero quota                                                   # /drone/src/tests/acceptance/features/apiMain/quota.feature:225
    Given the administrator has set the default folder for received shares to "Shares"                                           # OccContext::theAdministratorHasSetTheDefaultFolderForReceivedSharesTo()
    And auto-accept shares has been disabled                                                                                     # FeatureContext::autoAcceptSharesHasBeenDisabled()
    And user "Brian" has been created with default attributes and without skeleton files                                         # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And the quota of user "Brian" has been set to "0 B"                                                                          # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And the quota of user "Alice" has been set to "10 MB"                                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And user "Brian" has created folder "shareFolder"                                                                            # FeatureContext::userHasCreatedFolder()
      HTTP status code was not 201 or 204 while trying to create folder 'shareFolder' for user 'Brian'
      Failed asserting that an array contains 507.
```

https://drone.owncloud.com/owncloud/core/37767/90/13
```
  Scenario: Upload a file to external storage while quota is set on home storage                                                                               # /drone/src/tests/acceptance/features/apiMain/external-storage.feature:89
    Given user "Alice" has been created with default attributes and small skeleton files                                                                       # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And the quota of user "Alice" has been set to "1 B"                                                                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
    When user "Alice" uploads file "filesForUpload/textfile.txt" to filenames based on "/local_storage/testquota.txt" with all mechanisms using the WebDAV API # FeatureContext::userUploadsAFileToWithAllMechanisms()
    Then the HTTP status code of all upload responses should be "201"                                                                                          # FeatureContext::theHTTPStatusCodeOfAllUploadResponsesShouldBe()
      Response did not return expected status code
      Failed asserting that 507 matches expected 201.
```

I will adjust those tests so that they expect the new behavior...

phil-davis (approved):

phil-davis:
https://drone.owncloud.com/owncloud/core/37769/90/13
```
  Scenario: Cannot upload a file to external storage while quota is too low                                                                                    # /drone/src/tests/acceptance/features/apiMain/external-storage.feature:89
    Given user "Alice" has been created with default attributes and small skeleton files                                                                       # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And the quota of user "Alice" has been set to "1 B"                                                                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
    When user "Alice" uploads file "filesForUpload/textfile.txt" to filenames based on "/local_storage/testquota.txt" with all mechanisms using the WebDAV API # FeatureContext::userUploadsAFileTothAllMechanisms()
    Then the HTTP status code of all upload responses should be "507"                                                                                          # FeatureContext::theHTTPStatusCodeOfAllUploadResponsesShouldBe()
      Response did not return expected status code
      Failed asserting that 201 matches expected 507.
```

I need to look again at exactly what happens when a user with not enough quota uploads a file to a destination that is in an "external storage" (rather than in their own storage, or into a received share that is someone else's storage)

pako81:
> I need to look again at exactly what happens when a user with not enough quota uploads a file to a destination that is in an "external storage" (rather than in their own storage, or into a received share that is someone else's storage)

Upload should succeed as "external storages" are not counted against user's quota.

phil-davis:
The failing step actually does 4 uploads.
```
Destination: /local_storage/testquota.txt-olddav-regular HTTP status: 201
Destination: /local_storage/testquota.txt-olddav-oldchunking HTTP status: 201
Destination: /local_storage/testquota.txt-newdav-regular HTTP status: 201
Destination: /local_storage/testquota.txt-newdav-newchunking HTTP status: 507
```

The one using new dav and new chunking now returns 507 - the server decides that there is not enough quota (even though Alice is uploading to an external storage, not her own personal storage).

The other combinations of ways of uploading are still returning 201. At the end of the scenario, 3 files exist in the storage. The last file does not exist, because it was rejected with a 507.

Something in the code of this PR is effecting the "new dav, new chunking" way of uploading, but not the other ways.

phil-davis:
The last commit is "Adjust test for upload to external storage while quota is too low"

Actually, IMO, that change is not the behavior that we really expect or want. We want the original behavior that the scenario expected - uploads to external storage work even if the user does not have enough quota themselves. So that commit can be removed.

pako81:
> Actually, IMO, that change is not the behavior that we really expect or want. We want the original behavior that the scenario expected - uploads to external storage work even if the user does not have enough quota themselves. So that commit can be removed.

agree 👍

phil-davis:
I removed that last commit, and force-pushed.

I don't see why this special little unexpected behavior change should happen - the code has nothing special for the different old/new dav and old/new chunking upload methods.

Can a dev have a look please, maybe there are other code paths that need to be looked at?

jvillafanez:
The new code will prevent uploads regardless of the storage if the quota is 0 and the request comes from webdav.
As far as I know, the old code prevents the uploads via a stream wrapper, and such wrapper isn't "active" in external storages.

pako81:
> 
The new code will prevent uploads regardless of the storage if the quota is 0 and the request comes from webdav.
As far as I know, the old code prevents the uploads via a stream wrapper, and such wrapper isn't "active" in external storages.

Maybe we should think about adding a check in case the storage we want to create the file/dir into is an external storage and allow it.

pako81:
Like moving the check after: https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/Directory.php#L161:

```
list($targetStorage, $targetInternalPath) = \OC\Files\Filesystem::resolvePath($path);

// We are using == instead of === as the computerFileSize method which is used to get the quota may return a float type. Note that the same has been observed for the disk_free_space function in local storage
list($used, $free) = $this->getQuotaInfo();
if ($free == 0 && $targetStorage->instanceOfStorage('\OCP\Files\IHomeStorage') === true)) {
	throw new SabreInsufficientStorage('Creation of empty files is forbidden in case of no available quota');
}
```
Or something along the lines.

jvillafanez:
I think this is a "choose your poison" situation:
* Moving the code to https://github.com/owncloud/core/blob/c4f8cd14fb2761cca93fba53199a7595cdb80e70/apps/dav/lib/Connector/Sabre/QuotaPlugin.php might make more sense, but there is no "beforeCreateCollection" event that is triggered by sabre dav, so there is no place where we can put the code.
* Put the limitation in the stream would affect the skeleton creation in more things. It might be ok, but there won't be a proper error message thrown, just a "something went wrong" message in the best case.

Maybe you could try to remove those lines from the `createFile` method because the quota should be already taken care of in other places.
I don't really like checking the type of the storage, but I don't have better and easier solutions, so we could do that only for the `createDirectory` method because you should still be able to create directories in the external directory.

pako81:
> Maybe you could try to remove those lines from the createFile method because the quota should be already taken care of in other places.

You mean to completely remove the quota check from the `createFile` method or just the additional proposed check for the storage type? If completely removed, creation of empty files for 0 quota users will again be possible.

Also note that with the current PR in place, creation of empty files and directories in external storages (like WND) for 0 quota users is still possible. So it is not prevented.

jvillafanez:
You can remove the code from the `createFile` method because it should be taken care of already. For the `createDirectory`, you need to keep the check for the home storage so the quota is only applied there.

pako81:
Done, however I would expect following tests to fail now:

- Scenario: user cannot create an empty file in a shared folder of a user with no remaining quota
- Scenario: share receiver with 0 quota can copy empty file from shared folder to home folder
- Scenario: User with zero quota cannot upload an empty file
- Scenario: User with no remaining quota cannot upload an empty file

pako81:
@phil-davis should we adjust the failing tests?

phil-davis:
https://drone.owncloud.com/owncloud/core/37774/90/13
```
apiMain/external-storage.feature:89
  Scenario: Upload a file to external storage while quota is set on home storage                                                                               # /drone/src/tests/acceptance/features/apiMain/external-storage.feature:89
    Given user "Alice" has been created with default attributes and small skeleton files                                                                       # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
    And the quota of user "Alice" has been set to "1 B"                                                                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
    When user "Alice" uploads file "filesForUpload/textfile.txt" to filenames based on "/local_storage/testquota.txt" with all mechanisms using the WebDAV API # FeatureContext::userUploadsAFileToWithAllMechanisms()
    Then the HTTP status code of all upload responses should be "201"                                                                                          # FeatureContext::theHTTPStatusCodeOfAllUploadResponsesShouldBe()
      Response did not return expected status code
      Failed asserting that 507 matches expected 201.
```

This one gets at least some `507`.

phil-davis:
https://drone.owncloud.com/owncloud/core/37774/41/13
```
apiMain/quota.feature:234
apiMain/quota.feature:324
apiMain/quota.feature:332
```

```
  Scenario: user cannot create an empty file in a shared folder of a user with no remaining quota                # /drone/src/tests/acceptance/features/apiMain/quota.feature:234
    Given the administrator has set the default folder for received shares to "Shares"                           # OccContext::theAdministratorHasSetTheDefaultFolderForReceivedSharesTo()
    And auto-accept shares has been disabled                                                                     # FeatureContext::autoAcceptSharesHasBeenDisabled()
    And user "Brian" has been created with default attributes and without skeleton files                         # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And the quota of user "Brian" has been set to "10 B"                                                         # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And the quota of user "Alice" has been set to "10 MB"                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And user "Brian" has created folder "shareFolder"                                                            # FeatureContext::userHasCreatedFolder()
    And user "Brian" has shared file "/shareFolder" with user "Alice"                                            # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Brian" has uploaded file with content "ten bytes." to "useUpQuota.txt"                             # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Alice" has accepted share "/shareFolder" offered by user "Brian"                                   # FeatureContext::userHasReactedToShareOfferedBy()
    When user "Alice" uploads file with content "" to "/Shares/shareFolder/newTextFile.txt" using the WebDAV API # FeatureContext::userUploadsAFileWithContentTo()
    Then the HTTP status code should be "507"                                                                    # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 201 is not the expected value 507
      Failed asserting that 201 matches expected '507'.
```

```
  Scenario: User with zero quota cannot upload an empty file                               # /drone/src/tests/acceptance/features/apiMain/quota.feature:324
    Given the quota of user "Alice" has been set to "0 B"                                  # FeatureContext::theQuotaOfUserHasBeenSetTo()
    When user "Alice" uploads file with content "" to "testquota.txt" using the WebDAV API # FeatureContext::userUploadsAFileWithContentTo()
    Then the HTTP status code should be "507"                                              # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 201 is not the expected value 507
      Failed asserting that 201 matches expected '507'.
```

```
  Scenario: User with no remaining quota cannot upload an empty file                       # /drone/src/tests/acceptance/features/apiMain/quota.feature:332
    Given the quota of user "Alice" has been set to "10 B"                                 # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And user "Alice" has uploaded file with content "ten bytes." to "useUpQuota.txt"       # FeatureContext::userHasUploadedAFileWithContentTo()
    When user "Alice" uploads file with content "" to "testquota.txt" using the WebDAV API # FeatureContext::userUploadsAFileWithContentTo()
    Then the HTTP status code should be "507"                                              # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 201 is not the expected value 507
      Failed asserting that 201 matches expected '507'.
```

phil-davis:
1) It looks like the current code successfully prevents a folder being created in a place that has no quota remaining.

2) A user with no quota can still create empty files.

If we just want to do (1), then I can adjust the tests so that they do not expect (2).

If we want to also do (2), then the code needs to somehow be adjusted to prevent uploads of empty files when quota runs out.

pako81:
> 1) It looks like the current code successfully prevents a folder being created in a place that has no quota remaining.
> 
> 2) A user with no quota can still create empty files.
> 
> If we just want to do (1), then I can adjust the tests so that they do not expect (2).
> 
> If we want to also do (2), then the code needs to somehow be adjusted to prevent uploads of empty files when quota runs out.

@jvillafanez  based on https://github.com/owncloud/core/pull/40567#issuecomment-1406386534, wasn't suggested to remove the code from the `createFile` method? This, however, would still allow creation of empty files for users with no quota.

jvillafanez:
Quota for files is being handled in the quota dav plugin (https://github.com/owncloud/core/blob/master/apps/dav/lib/Connector/Sabre/QuotaPlugin.php) and in the quota stream (https://github.com/owncloud/core/blob/master/lib/private/Files/Stream/Quota.php), so code changes should be there unless there isn't any other option.

The problem is that the folder creation isn't handled in any of those classes, so we had to put the solution in the `Directory` class.

> @jvillafanez based on https://github.com/owncloud/core/pull/40567#issuecomment-1406386534, wasn't suggested to remove the code from the createFile method?

That was to fix the problem in the PR that the quota was being applied to external storages, which is something we don't want.

pako81:
@jvillafanez I have added the check for empty files in the quota dav plugin. Not sure it is required to have it in the quota stream as well.

phil-davis:
https://drone.owncloud.com/owncloud/core/37816/10/7
```
There were 24 failures:
1) OCA\DAV\Tests\unit\Connector\Sabre\QuotaPluginTest::testCheckQuotaChunkedOk with data set #0 (1024, 0, array('1024'))
Expectation failed for method name is "free_space" when invoked zero or more times
Parameter 0 for invocation OC\Files\View::free_space('sub/test.txt-chunking-12345-3-1') does not match expected value.
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'sub/test.txt'
+'sub/test.txt-chunking-12345-3-.1'

/drone/src/apps/dav/lib/Connector/Sabre/QuotaPlugin.php:249
/drone/src/apps/dav/lib/Connector/Sabre/QuotaPlugin.php:191
/drone/src/apps/dav/tests/unit/Connector/Sabre/QuotaPluginTest.php:202
...
```

phil-davis:
https://drone.owncloud.com/owncloud/core/37816/41/13
```
Scenario: share receiver with 0 quota can copy empty file from shared folder to home folder                  # /drone/src/tests/acceptance/features/apiMain/quota.feature:309
    Given the administrator has set the default folder for received shares to "Shares"                         # OccContext::theAdministratorHasSetTheDefaultFolderForReceivedSharesTo()
    And auto-accept shares has been disabled                                                                   # FeatureContext::autoAcceptSharesHasBeenDisabled()
    And user "Brian" has been created with default attributes and without skeleton files                       # FeatureContext::userHasBeenCreatedWithDefaultAttributesAndWithoutSkeletonFiles()
    And the quota of user "Brian" has been set to "0 B"                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And the quota of user "Alice" has been set to "10 MB"                                                      # FeatureContext::theQuotaOfUserHasBeenSetTo()
    And user "Alice" has created folder "shareFolder"                                                          # FeatureContext::userHasCreatedFolder()
    And user "Alice" has uploaded file with content "" to "/shareFolder/testquota.txt"                         # FeatureContext::userHasUploadedAFileWithContentTo()
    And user "Alice" has shared folder "/shareFolder" with user "Brian"                                        # FeatureContext::userHasSharedFileWithUserUsingTheSharingApi()
    And user "Brian" has accepted share "/shareFolder" offered by user "Alice"                                 # FeatureContext::userHasReactedToShareOfferedBy()
    When user "Brian" copies file "/Shares/shareFolder/testquota.txt" to "/testquota.txt" using the WebDAV API # FeatureContext::userCopiesFileUsingTheAPI()
    Then the HTTP status code should be "201"                                                                  # FeatureContext::thenTheHTTPStatusCodeShouldBe()
      HTTP status code 507 is not the expected value 201
      Failed asserting that 507 matches expected '201'.
```
I will adjust this test, it now also does "a good thing"

phil-davis:
The quota API acceptance tests are passing - good.

https://drone.owncloud.com/owncloud/core/37818/90/13
We still have `apiMain/external-storage.feature:89` - the different upload mechanisms (new and old dav, with and without chunking) give different results. When the user has no quota remaining on their home storage, some upload methods to the external storage allow the upload, and some reject it.

phil-davis:
I am looking at the unit test fails. It looks like the code-changes need to take into account the real target filename when chunking is being used.

phil-davis:
Now the only thing failing is `apiMain/external-storage.feature:89`

The resulting HTTP status codes are:
201 /local_storage/testquota.txt-olddav-regular
201 /local_storage/testquota.txt-olddav-oldchunking
201 /local_storage/testquota.txt-newdav-regular
507 /local_storage/testquota.txt-newdav-newchunking

Somehow we are now blocking upload when using "new" DAV and chunking.

But in the other 3 cases, we continue to let the upload go to the local storage, even when Alice has no personal quota remaining. That has always been the behavior.

What code-paths are happening for these 4 cases? Why do we get a different result for case 4?

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
```diff
@@ -188,28 +188,27 @@ public function checkQuota($path, $length = null, $extraSpace = 0) {
 		if ($length === null) {
 			$length = $this->getLength();
 		}
-		if ($length) {
```
jvillafanez, line 191: I'd keep this check. The length could be null if there is no "length" header available.

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
```diff
@@ -188,28 +188,27 @@ public function checkQuota($path, $length = null, $extraSpace = 0) {
...
+			$chunkHandler = $this->getFileChunking($info);
+			// subtract the already uploaded size to see whether
+			// there is still enough space for the remaining chunks
+			$length -= $chunkHandler->getCurrentSize();
+			// use target file name for free space check in case of shared files
+			$path = \rtrim($parentPath, '/') . '/' . $info['name'];
+		}
+		$freeSpace = $this->getFreeSpace($path);
+		$availableSpace = $freeSpace + $extraSpace;
+		if ($freeSpace !== FileInfo::SPACE_UNKNOWN && $freeSpace !== FileInfo::SPACE_UNLIMITED && (($length > $availableSpace) || ($availableSpace == 0))) {
```
jvillafanez, line 207: I'd add a comment in line 206 to say that false + int will return an int, so the `$availableSpace` will be an integer. This way, we can check `$availableSpace === 0` (with triple "=") as usual.
Otherwise, we should have a comment for the `==`

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
phil-davis: I wanted to execute the code even when length is 0.
I guess that I can check if length is not null.

phil-davis: done

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
phil-davis: done

phil-davis:
> What code-paths are happening for these 4 cases? Why do we get a different result for case 4?

Actually I think that we want case 4 to also allow the upload, and return 201. (the same as it currently does in core)

Any changes to `apiMain/external-storage.feature:89` should be reverted, and the original scenario should be made to pass.

IljaN:
> Any changes to apiMain/external-storage.feature:89 should be reverted, and the original scenario should be made to pass.

Yes I agree, because quota excludes ext-storages

pako81:
Maybe this is intended as a local path mounted as external storage? In any case, I also think this test should allow the upload and return a 201. Why is it still expecting a 507?

IljaN:
Yes probably the external mount is named "local_storage". Debugging...

pako81:
Indeed with this code in place uploading into an external storage mounted via the "Local" option fails for an user with 0 quota when uploading files > 10MB. Related to the chunking algorithm.

pako81:
Seems related to the creation of the `web-file-upload` folder which is being called by https://github.com/owncloud/core/blob/v10.11.0/apps/dav/lib/Upload/UploadHome.php#L48

phil-davis:
> Indeed with this code in place uploading into an external storage mounted via the "Local" option fails for an user with 0 quota when uploading files > 10MB. Related to the chunking algorithm.

yes, somehow the code changes in this PR are "accidentally" doing a quota check in the case of a "new" chunking upload to the "local/external" storage. If someone can see the code path that ends up running through code touched by this PR, then they could design some sort of check to add to an "if" in the code of this PR. For example, be able to know if the final destination of the file is "local/external" storage, and if so, then do not bother checking quota.

IljaN:
Last commit d21dc8e8 is work in progress to test for acceptance, however it probably will break because clients name the folder differently. Also there is maybe a more buletproof way to detect that we are trying to create something outside of "$user/files"

tests/acceptance/features/apiMain/external-storage.feature
```diff
@@ -85,7 +85,7 @@ Feature: external-storage
     # in the folder. The file cache will be cleared after the local storage is scanned.
     And as "Alice" file "local_storage/foo3/textfile0.txt" should exist
 
-
+  @skipOnOcV10.10 @skipOnOcV10.11
```
phil-davis, line 88: Note: this scenario is back to being unchanged. So I will remove the skip tags. It should pass on all of 10.10 10.11 10.12

phil-davis: done

pako81:
I am still unsure about the approach. @jvillafanez maybe there is a different, better way to check this. I.e. by additionally checking if instance of `UploadHome` --> https://github.com/owncloud/core/blob/master/apps/dav/lib/Upload/UploadHome.php#L31

jvillafanez:
Do we have an explanation for the problem with the uploads folder? As far as I see in https://github.com/owncloud/core/blob/c4f8cd14fb2761cca93fba53199a7595cdb80e70/apps/dav/lib/Upload/ChunkLocationProvider.php#L62-L75 the uploads folder should be considered as a external storage of type "local", so it should be excluded from the quota calculation.
What could be happening is that such mount point isn't mounted under some determined circumstances, so the folder isn't created in the "local" external storage but in the home storage, and so the quota is applied.

Looking at the code, it seems that if the `dav.chunk_base_dir` isn't set, the mount point isn't configured, so it should go through the home storage....

Summing up, could you check if the code works without the chunk patch if the `dav.chunk_base_dir` is set? We might need to return a properly configured mount point in https://github.com/owncloud/core/blob/c4f8cd14fb2761cca93fba53199a7595cdb80e70/apps/dav/lib/Upload/ChunkLocationProvider.php#L62-L75 with the right datadir if the `dav.chunk_base_dir` isn't set

pako81:
@jvillafanez thanks for the pointer, this seems indeed the issue. Without the chunks patch and with `dav.chunk_base_dir` properly set in config.php, uploads into external storages are possible when quota is 0.

pako81:
@jvillafanez so maybe something like this 

```
 if ($chunkBaseDir === '') {
            $datadir = $this->config->getSystemValue('datadirectory');
            return [
                    new MountPoint(Local::class, '/' . $user->getUID() . '/uploads', ['datadir' => $datadir, $loader])
            ];
}

**Comment by jvillafanez** [comment]
Yes, something like that. Maybe adjusting the `$cacheDir` variable so we only just one return statement, but anyway, that's the idea.

I don't know if there is any additional thing to take care of. I mean, assuming that the data directory is in a specific machine, it makes sense to have a quota there so the system doesn't run out of space. The upload folder might be considered as an external storage because it could be also placed in a secondary machine by setting up an NFS folder in the ownCloud host so the chunks don't take space that should be used by ownCloud.
Basically, I assume that if ownCloud's host has 30GB of free space, it's ok to setup quotas such as 15GB for 10 users knowing that the host won't have enough space. Furthermore, even if the quotas are properly setup, (3GB for 10 users), the host might still not have enough space.

**Comment by pako81** [comment]
@jvillafanez would be fine like this? Apparently, we now have some acceptance tests related to the new chunking algorithm which are failing.

**Comment by phil-davis** [comment]
https://drone.owncloud.com/owncloud/core/37947/41/13
```
apiMain/checksums.feature:192
apiMain/checksums.feature:208
apiMain/checksums.feature:226
apiMain/checksums.feature:355
apiMain/checksums.feature:366
```

```
  Scenario: Upload new DAV chunked file using async MOVE where checksum matches                                                                                                           # /drone/src/tests/acceptance/features/apiMain/checksums.feature:192
    Given using new DAV path                                                                                                                                                              # FeatureContext::usingOldOrNewDavPath()
    And the administrator has enabled async operations                                                                                                                                    # FeatureContext::triggerAsyncUpload()
    And user "Alice" has created a new chunking upload with id "chunking-42"                                                                                                              # FeatureContext::userHasCreatedANewChunkingUploadWithId()
      The HTTP status code 405 is not between 200 and 299
      Failed asserting that 405 is equal to 299 or is less than 299.
```

The various failing test scenarios are when Alice starts a new chunking upload. 

Similar for https://drone.owncloud.com/owncloud/core/37947/88/13 and https://drone.owncloud.com/owncloud/core/37947/89/13

It looks like the code changes are now rejecting some chunked uploads with 405 "method not allowed"

**Comment by IljaN** [comment]
Debugged this a bit w @pako81. Somehow the server thinks that the chunking directory does already exist. But the folder seems not to be created in the storage.

**Comment by jvillafanez** [comment]
You'll probably need to ensure the mounted folder exists before returning the mount, the same way it's done with a custom `dav.chunk_base_dir`

**Comment by pako81** [comment]
We are back to the case where only the `Upload a file to external storage while quota is set on home storage (/drone/src/tests/acceptance/features/apiMain/external-storage.feature:89)` acceptance test is failing --> https://drone.owncloud.com/owncloud/core/37995/90/13

@jvillafanez we still have the issue that something seems to be not correct when returning the datadir mountpoint in ChunkLocationProvider.php in case `dav.chunk_base_dir` is not set.

File: apps/dav/lib/Upload/ChunkLocationProvider.php [hunk 4]
```diff
@@ -62,15 +62,24 @@ public function __construct(IConfig $config) {
...
-			return [];
+			// this is needed as otherwise the returned mount point will be in the home storage rather
+			// than in the "local" external storage and this would cause troubles when applying quota.
+			$dataDir = $this->config->getSystemValue('datadirectory');
+			$cacheDir = \rtrim($dataDir, '/') . '/' . $user->getUID() . '/uploads';
+			if (!\file_exists($cacheDir)) {
+				\mkdir($cacheDir, 0770, true);
+			}
+			return [
+				new MountPoint(Local::class, $cacheDir, ['datadir' => $dataDir, $loader])
```
jvillafanez, line 73:
```
new MountPoint(Local::class, $cacheDir, ['datadir' => $cacheDir, $loader])
```
It should be `'datadir' => $cacheDir` instead of `'datadir' => $dataDir`. I don't know if this is causing the error in the tests

pako81: Yes, initially tried with `'datadir' => $cacheDir` as well - no difference. The test still fails locally for me with:

```
Scenario: Upload a file to external storage while quota is set on home storage                                                          # /home/core/tests/acceptance/features/apiMain/external-storage.feature:89
Given user "Alice" has been created with default attributes and small skeleton files                                                                       # FeatureContext::userHasBeenCreatedWithDefaultAttributes()
And the quota of user "Alice" has been set to "1 B"                                                                                                        # FeatureContext::theQuotaOfUserHasBeenSetTo()
When user "Alice" uploads file "filesForUpload/textfile.txt" to filenames based on "/local_storage/testquota.txt" with all mechanisms using the WebDAV API # FeatureContext::userUploadsAFileToWithAllMechanisms()
Then the HTTP status code of all upload responses should be "201"                                                                                          # FeatureContext::theHTTPStatusCodeOfAllUploadResponsesShouldBe()
Response did not return expected status code
Failed asserting that 507 matches expected 201.
And as "Alice" the files uploaded to "/local_storage/testquota.txt" with all mechanisms should exist                                                       # FeatureContext::filesUploadedToWithAllMechanismsShouldExist()
```

pako81: Which is because we are still getting through https://github.com/owncloud/core/blob/4f60253e39fa798743cfe087ce8185978591b4d1/apps/dav/lib/Connector/Sabre/Directory.php#L225 so apparently the target storage is still seen of type Home.

The other, simplified, approach which had all tests passing was https://github.com/owncloud/core/pull/40567/commits/b8e34585548faf82adf07e7314a4afa58d6c6ed7, where ChunkLocationProvider.php was not touched.

pako81: Isn't this because the class Home is extending the Local one (https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Home.php#L32) so basically there is no way to differentiate in this case and we always end up at not passing this check https://github.com/owncloud/core/blob/4f60253e39fa798743cfe087ce8185978591b4d1/apps/dav/lib/Connector/Sabre/Directory.php#L224

jvillafanez: I don't think it's like that.  A local storage isn't a home storage.

It might be that we're pointing exactly at the root of the mount, so maybe the target storage is the home storage with a relative path of `/user/files/uploads` instead of what we're expecting (local storage with relative path of `''` - basically the root of the mount)

pako81:
@jvillafanez @phil-davis should we go ahead and merge this?

jvillafanez (approved):
I assume it works, haven't tested it

pako81:
I have again tested manually all the steps from the top post with the new code in place and did not observe any unexpected behaviour.

individual-it:
Tests with desktop client 3.2.0 on Windows 10:

1. create folder for a user that has 0 quota:
    1. create a user with 0 quota
    2. join the desktop client
    3. create a folder on the desktop
    => message that it was not possible to sync :+1: 
image
    => red cross in the desktop client settings UI and taskbar :+1: 
image
image
    => Windows Explorer does not indicate a problem :-1: 
image
2. increase quota of a user that had no quota before
    1. use the user of first test and change the quota to 100KB
    2. force-sync the desktop client
    => new folder is created on the server :+1: 
3. overfill quota
    1. use the user from test no 2
    2. copy a couple of files that individually fit into the quota but in sum don't
    => some files are uploaded, others are not with the message that there is not sufficient quota :+1:
4. create a folder when quota is almost full
    1. use the user from test no 3
    2. create a folder while the quota is nearly full (see test no 2)
    => folder is created and synced to the server :+1: 
5. reduce quota, so that its over 100% full and try to create a folder
    1. use the user from test no 4
    2. as admin change the quota of the user to 90kb
    3. try to create a folder
    => same result as test no 1
6. upload to a share without quota
    1. create a new user without quota
    2. from this user share a folder to the user from test 5, who has limited quota and already exceeded it
    3. as share receiver upload a file to the share
    4. as share receiver create a folder in the share
    => both works and is synced to the server :+1: 
7. create a folder inside of a share that belongs to a user who has exceeded its quota
    1. use both users of test no 6
    2. limit the sharer, so that its quota is exceeded
    3. as share receiver try to create a folder
    => same result at test no 1 :+1:

ho4ho:
Looked like could cause problem by 32bit system:
- #40695
- #40703

Issue #31546, opened 2018-05-28
[Concept] Introduce (Pluggable) Authentication Modules


butonic:
I took a step back and had a look at [what PAM does](https://www.linux.com/news/understanding-pam) to provide a flexible authentication process and what we can learn from that.

How does PAM structure the login process?
1. _auth management_ checks auth factors like password and TOTP
2. _account management_ might deny access based on non auth related constraints, eg only during working times, checks account expiry or disabled state
3. _session management_ (called on open & close) setup session, eg auth token, user home, enforce quota

There is also _password management_, but it is only relevant when the auth token is actually changed (in our case password complexity rules are checked)

All of these steps are implemented in ownCloud. But most of the time are hardcoded and messy.

I propose adding additional modules in the spirit of PAM modules and subsequently refactoring the existing code.

# (Pluggable) Authentication Modules
OwnCloud should give apps a way to interact with the different authentication steps.
1. `IAuthModule`s are responsible for authentication and granting group membership
2. `IAccountModule`s might deny access based on non auth related constraints
3. `ISessionModule`s are used to prepare and cleanup the user session
4. `IFactorModule`s handle token updates eg password complexity

## IAuthModule
_check auth factors like password and TOTP_
We started implementing these, but only to verify an existing session. We should reuse them during login as well.
- the first facter needs te identify the user (= determine his uid) so 2FA can look up the challenges configured for the user
- [ ] refactor basic auth as `IAuthModule`
- [ ] refactor shibbolet auth as `IAuthModule`
- [ ] refactor 2FA as `IAuthModule`? for now leave as step between `IAuthModule` checks and `IAccountModule` checks?
- FR: grant roles

## IAccountModule
_might deny access based on non auth related constraints, eg only during working times_
- [x] add an `AccountModuleMiddleware` that only grants access to controllers implementing `@IAccountModuleController`? (or annotated with @AccountModule)
  - PoC in https://github.com/owncloud/core/pull/31547
- [ ] refactor account is enabled check as `IAccountModule`
- [ ] refactor account autoprovisioning and metadata sync as `IAccountModule`
- FR: enforce password update, see https://github.com/owncloud/password_policy/pull/15
- FR: enforce TOTP setup
- FR: enforce Avatars
- FR: enforce email
- FR: introduce account expiry
- FR: check roles granted in an `IAuthModule`

## ISessionModule
_prepare and cleanup the user session (called on open & close)_ 
- [ ] refactor create auth token for session as `ISessionModule`
[...]

PR #40574, opened 2023-01-12
Add support for login policies

jvillafanez:
A policy has been implemented so admins can allow or reject groups of users to access via specific login mechanisms

## Description
Add support for login policies. An initial login policy has been implemented so admins can allow or reject groups to access ownCloud via specific login mechanism.
For example, only the groups "guests" and "admin" could access via username + password, while the rest of users must access through other mechanisms such as openidconnect.

Login policies will emit a "failed login" event if the user isn't allowed to login.

Some additional changes have been made in the token access (app password) so it isn't invalidated under some circumstances (login policies were causing the invalidation of the token if the policy rejected the user.)

## Related Issue
https://github.com/owncloud/enterprise/issues/5295

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## TODO:
- [x] Unit tests.
- [x] Changes in config.sample.php

## Usage:

The `GroupLoginPolicy` is already pre-registered, so there is nothing to do in order to register this policy in the `LoginPolicyManager`

The `LoginPolicyManager` will use the `loginPolicy.order` key in the config.php file to activate and use the policies in the list in the specified order.
```
'loginPolicy.order' => ['OC\Authentication\LoginPolicies\GroupLoginPolicy'],
```
If no login policy is activated in the `loginPolicy.order` list, ownCloud will work normally. Only registered policies can be activated.

For now, only the `OC\Authentication\LoginPolicies\GroupLoginPolicy` is available.

## GroupLoginPolicy configuration.
The configuration of the `GroupLoginPolicy` is done through the config.php file.
```
loginChecker.groupLoginChecker.forbidMap => [
  'password' => [
    'allowOnly' => ['group1, group2'],
    'reject' => ['group3'],
  ],
  'token' => [
    'reject' => ['admin'],
  ],
  '' => [
    'allowOnly' => ['group2'],
  ],
]
```

Each key of the forbidMap is a known login type (**to be listed every known login type**), and the value is a map containing a list of groups that are allow to use that login type (rejecting everyone else), and a list of groups that are rejected from using that login type.

In the above example, users belonging to the admin group won't be able to access via token (app password), while the rest of the users can.
Only users from group1 and group2 are allowed to access through username + password, and users from group3 will be rejected. In case a users is member of an allowOnly group and a reject group, rejection will take priority. This means that even if user1 is member of group1 he won't be able to access if he's also member of group3.

In some weird cases the login type is unknown (there was an issue with openidconnect which should be solved in the master code of the branch). In this case, the login type might be the empty string `''`.

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -143,7 +143,7 @@ $CONFIG = [
 /**
  * Define the ownCloud database user
  * This must be unique across ownCloud instances using the same SQL database.
- * This is setup during installation, so you shouldn't need to change it.
+ * This is set up during installation, so you shouldn't need to change it.
  */
 'dbuser' => '',
 
@@ -224,7 +224,7 @@ $CONFIG = [
  * The web UI might send a "heartbeat" based on the activity happening
  * in order to extend the session lifetime and keeping it from timing out
  * prematurely. If there is no activity happening and the lifetime is
- * reached, you'll have to login again.
+ * reached, you'll have to log in again.
  * The default is 20 minutes, expressed in seconds.
  */
 'session_lifetime' => 60 * 20,
@@ -417,7 +417,7 @@ $CONFIG = [
 
 /**
  * Define the IP address of your mail server host
- * Depends on `mail_smtpmode`. May contain multiple hosts separated by a semi-colon.
+ * Depends on `mail_smtpmode`. May contain multiple hosts separated by a semicolon.
  * If you need to specify the port number, append it to the IP address separated by
  * a colon, like this: `127.0.0.1:24`.
  */
@@ -518,7 +518,7 @@ $CONFIG = [
  * As an example, alerts shown in the browser to upgrade an app are triggered by
  * a cron background process and therefore uses the url of this key, even if the user
  * has logged on via a different domain defined in key `trusted_domains`. When the
- * user clicks an alert like this, they will be redirected to that URL and must logon again.
+ * user clicks an alert like this, they will be redirected to that URL and must log on again.
  */
 'overwrite.cli.url' => '',
 
@@ -702,7 +702,7 @@ $CONFIG = [
  *     rules. Please refer to https://doc.owncloud.com/server/latest/admin_manual/configuration/files/file_versioning.html
  *    for more information.
  * * `D, auto`
- *     keep versions at least for D days, apply expire rules to all versions
+ *     keep versions at least for D days, apply expiry rules to all versions
  *     that are older than D days
  * * `auto, D`
  *     delete all versions that are older than D days automatically, delete
@@ -898,7 +898,7 @@ $CONFIG = [
 /**
  * Alternate Code Locations
  *
- * Some of the ownCloud code may be stored in alternate locations.
+ * Some ownCloud code may be stored in alternate locations.
  */
 
 /**
@@ -1165,7 +1165,7 @@ $CONFIG = [
 	'host' => 'localhost', // can also be a unix domain socket: '/tmp/redis.sock'
 	'port' => 6379,
 	'timeout' => 0.0,
-	'password' => '', // Optional, if not defined no password will be used.
+	'password' => '', // Optional, if not defined, no password will be used.
 	'dbindex' => 0,   // Optional, if undefined SELECT will not run and will use Redis Server's default DB Index. Out of the box, every Redis instance supports 16 databases so `<dbIndex>` has to be set between 0 and 15.
 	 // Optional config option
 	 // In order to use connection_parameters php-redis extension >= 5.3.0 is required
@@ -1196,14 +1196,14 @@ $CONFIG = [
  *  - \RedisCluster::FAILOVER_DISTRIBUTE - randomly distribute read commands across primary and replica nodes
  */
 'redis.cluster' => [
-	'seeds' => [ // provide some/all of the cluster servers to bootstrap discovery, port required
+	'seeds' => [ // provide some/all the cluster servers to bootstrap discovery, port required
 	  'localhost:7000',
 	  'localhost:7001'
 	],
 	'timeout' => 0.0,
 	'read_timeout' => 0.0,
 	'failover_mode' => \RedisCluster::FAILOVER_DISTRIBUTE,
-	'password' => '', // Optional, if not defined no password will be used.
+	'password' => '', // Optional, if not defined, no password will be used.
 	 // Optional config option
 	 // In order to use connection_parameters php-redis extension >= 5.3.0 is required
 	 // In order to use SSL/TLS redis server >= 6.0 is required
@@ -1491,7 +1491,7 @@ $CONFIG = [
  * The list of apps that are allowed and must not have a signature.json file present.
  * Besides ownCloud apps, this is particularly useful when creating ownCloud themes,
  * because themes are treated as apps. The app is identified with it´s app-id.
- * The app-id can be identified by the foldername of the app in your apps directory.
+ * The app-id can be identified by the folder name of the app in your apps directory.
  * The following example allows app-1 and theme-2 to have no signature.json file.
  */
 'integrity.ignore.missing.app.signature' => [
@@ -1675,7 +1675,7 @@ $CONFIG = [
 'upgrade.disable-web' => false,
 
 /**
- * Define whether or not to enable automatic update of market apps
+ * Define whether to enable automatic update of market apps
  * Set to `false` to disable.
  */
 'upgrade.automatic-app-update' => true,
@@ -1788,7 +1788,7 @@ $CONFIG = [
 
 /**
  * Configuration of the Group Login Policy
- * Povide configuration for the
+ * Provide configuration for the
  * 'OC\Authentication\LoginPolicies\GroupLoginPolicy' policy.
  *
  * The configuration will be something like:
@@ -1808,11 +1808,11 @@ $CONFIG = [
  * ],
  *
  * Each login type can have a list of groups that will be the ones
- * only allowed to login using that login type, and also a list of
+ * only allowed to log in using that login type, and also a list of
  * groups that will be rejected from using that login type.
- * Note that this applies to users beloging to thos groups. If a user
+ * Note that this applies to users belonging to those groups. If a user
  * is member of an "allowOnly" group and also of a "reject" group,
- * the "reject" will take priority, so the user won't be able to login
+ * the "reject" will take priority, so the user won't be able to log in
  * using that login type.
  *
  * List of known login types:
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1760,4 +1760,72 @@ $CONFIG = [
  */
 'grace_period.demo_key.link' => 'https://owncloud.com/try-enterprise/',
 
+/**
+ * The order of the login policies that will be checked, if any.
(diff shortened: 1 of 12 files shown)


butonic:
The Idea with https://github.com/owncloud/core/issues/31546 was to use an IAuthModule or a similar PAM concept for thngs like this. But if you think it does not fit ... go ahead ...

jvillafanez:
At least for this particular use case, I don't think we can use https://github.com/owncloud/core/issues/31546 . We need to know how the user is being logged in or how he has logged in, and we don't have that information there. This would require changes in the interface.

In addition, I'm not exactly sure about the architecture. It seems the AuthModule depends on the AccountModule. This would mean that each AuthModule would depend and use the AccountModuleManager, which I'm not sure if it happens already (for openidconnect, kerberos and others).
There are also cases where the login doesn't go through an auth module (regular username + password, for example). It would be weird if you can't access through webdav but you can through web UI using the same username and password.

lib/private/Authentication/LoginPolicies/GroupLoginPolicy.php
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	public function checkPolicy(string $loginType, IUser $user): bool {
+		$policyConfig = $this->config->getSystemValue('loginPolicy.groupLoginPolicy.forbidMap', []);
+		if (!\is_array($policyConfig)) {
+			// if not properly configured, allow access
+			return true;
+		}
+
+		if (isset($policyConfig[$loginType])) {
```
DeepDiver1975, line 71: usage of `??` can reduce line number .....

DeepDiver1975:
I guess the idea is also to allow apps to add login policies via calling the manager in app.php.
Was that tested as well? THX

lib/private/Authentication/LoginPolicies/GroupLoginPolicy.php
jvillafanez: Done.

For the "reject" and "allowOnly", I think it's clearer this way because we won't check anything if the keys aren't present.

jvillafanez:
> I guess the idea is also to allow apps to add login policies via calling the manager in app.php.
Was that tested as well? THX

That's the idea. I haven't tested it yet though. I'll probably check it by adding the expected code in a random app instead of creating one from scratch.

mmattel:
@jvillafanez pls do not forget to add a `config.sample.php` entry like done in `GroupLoginPolicy.php`.
When merged, docs need to do a `config-to-docs` run to get the changes from config sample added.

jvillafanez:
For testing, I've used the brute_force_protection app as base. Added the "lib/ProtPolicy.php" and the patch:
```
<?php

namespace OCA\BruteForceProtection;

use OCP\IUser;
use OCP\IRequest;
use OCP\Authentication\LoginPolicies\ILoginPolicy;
use OC\User\LoginException;

class ProtPolicy implements ILoginPolicy {
	private $request;

	public function __construct(IRequest $request) {
		$this->request = $request;
	}

	public function checkPolicy(string $loginType, IUser $user): bool {
		if ($user->getUID() === 'admin' && $this->request->getServerProtocol() === 'http') {
			throw new LoginException('admin access requires https');
		}
		return true;
	}
}
```
patch:
```
diff --git a/appinfo/app.php b/appinfo/app.php
index 0212f5f..1bc1170 100644
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -27,3 +27,4 @@ use OCA\BruteForceProtection\Hooks;
 
 $app = new Application();
 $app->getContainer()->query(Hooks::class)->register();
+$app->registerLoginPolicy();
diff --git a/lib/AppInfo/Application.php b/lib/AppInfo/Application.php
index 07987c9..4217aef 100644
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -24,9 +24,16 @@
 namespace OCA\BruteForceProtection\AppInfo;
 
 use \OCP\AppFramework\App;
+use OCA\BruteForceProtection\ProtPolicy;
 
 class Application extends App {
        public function __construct(array $urlParams=[]) {
                parent::__construct('brute_force_protection', $urlParams);
        }
+
+       public function registerLoginPolicy() {
+               $server = $this->getContainer()->getServer();
+               $loginPolicyManager = $server->getLoginPolicyManager();
+               $loginPolicyManager->registerPolicy($this->getContainer()->query(ProtPolicy::class));
+       }
 }
```

Note that the brute_force_protection app is of the type "prelogin", so the app is expected to load quite early.
We require the app to be loaded before the user has logged in, so I'm not fully sure if the app requires any special type. I guess the "authentication" type could also make sense.

The basic idea is that the `LoginPolicy` is registered in the `LoginPolicyManager` when the app is loaded. This should be quite light because it's just creating the instance of the policy and register it. Other than that, it depends on what the actual policy does.
The configuration of the policy depends entirely on the app itself. In this dummy case, it doesn't require configuration, but more complex apps might provide a web UI to configure the policy, or do it through the config.php app.
The only thing to take into account is that, the policy needs to be registered and activated. To activate the policy, it needs to be in the `loginPolicy.order` list (config.php file). For this dummy test, it should be:
```
  'loginPolicy.order' => [
    'OC\Authentication\LoginPolicies\GroupLoginPolicy',
    'OCA\BruteForceProtection\ProtPolicy',
  ],
```

I think this is enough as a PoC that 3rd party apps can also provide custom login policies with this mechanism. I don't think we want to create a complete app just for testing this feature, at least without specific use cases.

ownclouders:

https://drone.owncloud.com/owncloud/core/37643/116

config/config.sample.php
```diff
@@ -1760,4 +1760,72 @@
  */
 'grace_period.demo_key.link' => 'https://owncloud.com/try-enterprise/',
 
+/**
+ * The order of the login policies that will be checked, if any.
+ * Policies must be registered in order to use / activate them. This is usually
```
mmattel, line 1765:
```suggestion
/**
 * Order of login policies
 * The order of the login policies that will be checked, if any.
 * Policies must be registered in order to use / activate them. This is usually
```
The first line is the headline printed in the TOC on top. It always should be short and without an ending period.

config/config.sample.php
```diff
@@ -1760,4 +1760,72 @@
...
+ * ],
+ *
+ * The configuration of the policies depends on the policy itself, so they could
+ * be configured in multiple and different ways. We won't cover them here.
+ */
+'loginPolicy.order' => [],
+
+/**
+ * Povide configuration for the
+ * 'OC\Authentication\LoginPolicies\GroupLoginPolicy' policy.
```
mmattel, line 1790:
```suggestion
/**
 * Configuration of login policies
 * Povide configuration for the
 * 'OC\Authentication\LoginPolicies\GroupLoginPolicy' policy.
```

jvillafanez:
This is ready to review. I've checked the original issue and it works fine with this PR (tested with openidconnect and guests app to ensure it works).
We might need to adjust the wording of the error messages.

@mmattel I guess this will need to end up in the docs eventually. There should be enough information in this PR, but feel free to ask any question.

DeepDiver1975:
> The only thing to take into account is that, the policy needs to be registered and activated. To activate the policy, it needs to be in the `loginPolicy.order` list (config.php file). For this dummy test, it should be:
> 
> ```
>   'loginPolicy.order' => [
>     'OC\Authentication\LoginPolicies\GroupLoginPolicy',
>     'OCA\BruteForceProtection\ProtPolicy',
>   ],
> ```
> 
> I think this is enough as a PoC that 3rd party apps can also provide custom login policies with this mechanism. I don't think we want to create a complete app just for testing this feature, at least without specific use cases.

That's a bit odd from my pov that this has to be setup in two places.....

jvillafanez:
The registration part is for the devs. If we want to give them a chance to implement custom policies, someone will need to create the objects, and for core to create them will require a specific signature in the constructors, which will limit the options. By using a registration process, it's expected that the app will provide the objects already created, so it will just need to register them.
This registration process is expected to be done by the apps in the app.php file, so once the app is enabled the policy will always be registered. Note that for the policies provided by core, they're expected to be registered as soon as the policy manager instance is created.

Now, the admin needs a way to control the registered policies. Since core policies will always be registered, the admin needs a way to activate or deactivate them. In addition, he also has control over the policy order, so maybe he wants a custom policy to be checked before any of the core ones.

Maybe the confusion comes from using the class names there... we could require a `getPolicyName()` function in the interface, but we'd need to check for possible name collisions which would be a pain to deal with. Using the full class name should be easy enough to ensure there are no name collisions, and there is no additional requirement.

Note that an app could provide and register multiple login policies. There could be an app that groups a bunch of policies and registers 10 policies; in this case it's very unlikely that the admin will use all of them, but maybe 2 or 3.

DeepDiver1975 (approved):

config/config.sample.php
```diff
@@ -1760,4 +1760,74 @@
...
+ *
+ * The names of the policies must be documented if they come from an app.
+ * Core provide the following list of policies (only one for now):
+ * - 'OC\Authentication\LoginPolicies\GroupLoginPolicy'
+ *
+ * In order to use / activate the policy, include the name in the policy
+ * order below, such as:
+ * 'loginPolicy.order' => ['OC\Authentication\LoginPolicies\GroupLoginPolicy'],
+ *
+ * Multiple policies could be use as long as they're registered (the
```
pako81, line 1777: `used`

pako81:
@jvillafanez merge it?

jvillafanez:
no problems from my side

config/config.sample.php
phil-davis: fixed

phil-davis:
I fixed the last-mentioned grammar thing in config.sample.php
And rebased, because the branch was now a long way behind master.
IMO this is good to merge when CI finishes.

lib/private/Authentication/LoginPolicies/GroupLoginPolicy.php
```diff
@@ -0,0 +1,104 @@
...
+ * A specific group could be allowed only via token. Another group could
+ * be denied to access via username + password.
+ *
+ * Configuration is done via config.php file
+ * ```
+ * loginPolicy.groupLoginPolicy.forbidMap => [
+ *  'password' => [
+ *    'allowOnly' => ['group1, group2'],
+ *    'reject' => ['group3'],
+ *  ],
```
phil-davis, line 40: I don't understand the use case of having `allowOnly` and `reject` mentioned for the same login type.

If I am user `phil` and not in any of `group1`, `group2` or `group3` then what happens when I send a request with my ordinary password?
I am not in `allowOnly`, so I guess that the request will be rejected.
So then, what is the point of `reject` of `group3`?
Everyone who is not in `group1` or `group2` will be rejected anyway.

jvillafanez: There could be a user which is member of group1 AND group3. In this case, the "reject" policy takes precedence, so even if the user is member of the group1, he will be rejected because he's member of the group3.

phil-davis: Thanks, that makes sense.

PR #40584, opened 2023-01-17
tweak rewrite conditions in htaccess

C0rby:
This change hardens the rewrite rules to match the exact paths we want and not any subpaths e.g. `/somefolder/status.php`. 

Thanks to Terry Franklin, Matt Harris, Hayden Barker and Colin Smith (aka yoloClin) from Radiant Security (https://radiant.security) for reporting this.


ownclouders:
:boom: Acceptance tests pipeline <strong>apiProvisioningGroups-v2-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37669/45

xoxys (approved):

C0rby:
@IljaN, could you maybe tell me why the test is failing?
I'm not sure if this is because of my change.

DeepDiver1975:
https://github.com/owncloud/core/blob/32b69847223ef45f4cc4516e99864d407e63f0e5/tests/acceptance/run.sh#L764

lib/private/Setup.php
```diff
@@ -498,18 +498,18 @@ public static function updateHtaccess() {
...
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/public.php";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/cron.php";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/core/ajax/update.php";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/status.php";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/ocs/v1.php";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/ocs/v2.php";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/updater/";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/ocs-provider/";
-			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !/ocm-provider/";
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/core/img/favicon.ico$";
```
DeepDiver1975, line 501: `$` has to be escaped

DeepDiver1975 (dismissed):

lib/private/Setup.php
C0rby: I'm not so sure about that. This line already has a `$` which is not escaped and that was working before.
https://github.com/owncloud/core/blob/24fd78db1debb4b0f1523aefc94f9bb4f3b19e74/lib/private/Setup.php#L500

DeepDiver1975: true .... 

well run the occ command yourself and debug it :see_no_evil:

lib/private/Setup.php
```diff
@@ -498,18 +498,18 @@ public static function updateHtaccess() {
...
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/robots.txt$";
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/remote.php";
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/public.php";
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/cron.php";
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/core/ajax/update.php";
+			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/status.php$";
```
ho4ho, line 507:
```suggestion
			$content .= "\n  RewriteCond %{REQUEST_FILENAME} !^/status\\.php$";
```

Also needing escaping to not machting by `status-php`?

mdusher:
It may be worth looking at swapping `%{REQUEST_FILENAME}` out for `%{REQUEST_URI}` as according to the mod_rewrite documentation (https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html) there is some cases where `%{REQUEST_FILENAME}` can resolve to the full local path of the script/file (eg. /var/www/owncloud/status.php instead `/status.php`) which would break those `RewriteCond` rules.

C0rby:
> It may be worth looking at swapping `%{REQUEST_FILENAME}` out for `%{REQUEST_URI}` as according to the mod_rewrite documentation (https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html) there is some cases where `%{REQUEST_FILENAME}` can resolve to the full local path of the script/file (eg. /var/www/owncloud/status.php instead `/status.php`) which would break those `RewriteCond` rules.

That is actually why the CI failed.... Thanks for the hint. :+1:

phil-davis:
@C0rby is a changelog needed for this?

C0rby:
> @C0rby is a changelog needed for this?

I added one. :+1:

IljaN (approved):

ho4ho:
Looked like issues cause by this by users have now this by upgradation by ownCloud 10.12.0 :-/

> Your web server is not yet set up properly to allow file synchronization because the WebDAV interface seems to be broken.

https://central.owncloud.org/t/owncloud-stoped-working-after-update/41513

ho4ho:
Problem could be cause by new adding `^` not match by e.g. `/owncloud/remote.php` if use sub-dir install?

C0rby:
> Problem could be cause by new adding `^` not match by e.g. `/owncloud/remote.php` if use sub-dir install?

I did test sub-dir installations and that worked at that time but maybe I've missed something.

ho4ho:
Workaround document by https://github.com/owncloud/docs/pull/4823

PR #40586, opened 2023-01-19
[Not Merge] Workaround to fix the orientation of some jpg images

jvillafanez:
The orientation of the image is read from the exif data of the image. The data was being read correctly in some cases, but not in all of them. The reason of the failure is still unclear because reading the same data from the same file using the same exif_read_data function works differently depending on the stream being used.

## Description
Somehow there are errors loading the exif data from some images. Since such data isn't loaded, we can't fix the orientation of the image if needed. This workaround allows loading the data in those cases.

## Related Issue
https://github.com/owncloud/enterprise/issues/5500

## How Has This Been Tested?
Tested using the file below. After uploading through the web UI, the image should be shown correctly instead of being rotated

portrait

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

## Notes
Taking into account that we need to write again the content of the file and it could be several MBs, this could hinder the performance. Unless a better solution is found, we might need to consider using a config flag or something similar to enable / disable this fix.

### DIFF
--- a/lib/private/legacy/image.php
+++ b/lib/private/legacy/image.php
@@ -494,7 +494,19 @@ class OC_Image implements \OCP\IImage {
 	public function loadFromFileHandle($handle) {
 		$contents = \stream_get_contents($handle);
 		if ($this->loadFromData($contents)) {
-			$this->loadExifData($handle);
+			// The loadExifData function has problems using the $handle resource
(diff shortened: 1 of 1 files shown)


jvillafanez:
This PR will be superseded by https://github.com/owncloud/core/pull/40600

PR #40577, opened 2023-01-13
Make remote sharee search configurable

michielbdejong:
## Description
When the 'sciencemesh' app is installed, use it instead of the federatedfilesharing app to
find sharee matches for OCM sharing.

## Related Issue
- Fixes #40576

## Motivation and Context
This is needed for OC-10 instances that want to join https://sciencemesh.io

## How Has This Been Tested?
- test environment:
  - GitPod from https://github.com/cs3org/ocm-test-suite
  - run ./gitpod-init.sh
  - run ./scripts/clean.sh
  - run ./scripts/orro-testing.sh
  - go the Firefox browser-in-a-browser that is exposed to port 5800
  - in there, navigate to https://oc1.docker (einstein / relativity)
  - create a local user 'john'
  - using the ScienceMesh app, generate an invite
  - accept the invite on oc2.docker (marie / radioactivity)
  - confirm that marie is now a ScienceMesh contact of einstein on oc1.docker
- test case 1: sharing via ScienceMesh
  - as einstein on oc1.docker, share a file and start typing "m..." as the sharee
  - as marie on oc2.docker, accept the share
  - check the revaoc1.docker logs and see that the sharing happened via oc1 -> revaoc1 -> revaoc2 -> oc2
- test case 2: sharing via local
  - as einstein on oc1.docker, share a file and start typing "jo..." as the sharee
  - it should offer to share with John as a local user

## Screenshots (if appropriate):
Screenshot 2023-01-13 at 13 53 37

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:

- [x] Code changes
- [ ] ~~Unit tests added~~
- [ ] ~~Acceptance tests added~~
- [ ] ~~Documentation ticket raised~~
- [x] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

PR #40587, opened 2023-01-19
Make remote sharee search configurable - adjust

phil-davis:
This is the code of #40577 plus the last commit that adjusts the code of `getRemote` and the associated test.

https://github.com/owncloud/core/pull/40587/commits/431099872ab7ff090d5e6d0fc3c256a3318eacfc

We can call `getSystemValue` without having to specify a default - that all works fine. We just need to add to the mocks in `returnValueMap` in the related unit test.


michielbdejong:
Merged into https://github.com/owncloud/core/pull/40577

PR #40588, opened 2023-01-20
[full-ci] Add account creation time

pako81:
## Description
Add account creation time

## Related Issue
- https://github.com/owncloud/enterprise/issues/2298

## Motivation and Context
Add account creation time in the `oc_accounts` table so that this info can be displayed over the Users page. Also extended the users provisioning API and the `user:list` occ command to display this info.

## How Has This Been Tested?

Manually:

**Test 1 :**

- create an user over the WebUI and check that the `Creation Time` column correctly displays this info

image

**Test 2 :**

- create an LDAP user and run the `user:sync` command and check that the `Creation Time` column correctly displays this info

**Test 3 :** 

-Share with a guest user (create the user) and check that the `Creation Time` column correctly displays this info

**Test 4 :**

- Use the `user:add` occ command to create a new user and check that the `Creation Time` column correctly displays this info

**Test 5 :**

- Create the user by mean of the users provisioning API and check that this info is correctly displayed:

```
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message/>
 </meta>
 <data>
  <enabled>true</enabled>
  <quota>
   <definition>default</definition>
  </quota>
  <email/>
  <displayname>user14</displayname>
  <home>/var/www/html/owncloud10110/data/user14</home>
  <two_factor_auth_enabled>false</two_factor_auth_enabled>
  <last_login>0</last_login>
  <creation_time>1674161957</creation_time>
  <language>en</language>
 </data>
</ocs>
```

**Test 6 :**

- Use the `user:list` occ command to show user's attributes and check that this info is correctly displayed:

```
- user14:
    - uid: user14
    - displayName: user14
    - email
    - quota: default
    - enabled: true
    - lastLogin: 0
    - creationTime: 1674161957
    - home: /var/www/html/owncloud10110/data/user14
    - backend: Database
    - cloudId: user14@xxx.xxx.xx.xx/owncloud
    - searchTerms:
```

## Types of changes
- [X] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [x] Changelog item


lib/public/IUser.php
```diff
@@ -100,6 +100,22 @@ public function getLastLogin();
...
+	  * @return int
+	  * @since 8.0.0
+	  */
+	public function getCreationTime();
+ 
+	/**
+	  * set the timestamp of the user's creation time
+	  *
+	  * @return int
+	  * @since 8.0.0
```
DeepDiver1975, line 115: 10.12.0

DeepDiver1975 (changes requested):
indent looks off in some places

ownclouders:
:boom: Acceptance tests pipeline <strong>webUIFfSmoke-3-1-firefox-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37716/158

DeepDiver1975:
Screenshot from 2023-01-24 11-50-14

phil-davis:
On the UI, for existing users, the UI shows Creation Time "never". Maybe it would be better to have a special-case text in that case (when Creation Time" is Unix timestamp 0), and say "unknown".

DeepDiver1975:
> On the UI, for existing users, the UI shows Creation Time "never". Maybe it would be better to have a special-case text in that case (when Creation Time" is Unix timestamp 0), and say "unknown".

@pako81

phil-davis:
Locally, I had to make sure that the migration ran, so that the accounts table change happened. I pushed a commit to do the smallest version bump in this PR. I also put `full-ci` in the PR title, so that I can see all the test results.

Locally the reported failing scenario runs for me - I will dig more after CI finishes.

pako81:
> On the UI, for existing users, the UI shows Creation Time "never". Maybe it would be better to have a special-case text in that case (when Creation Time" is Unix timestamp 0), and say "unknown".

Should be addressed now.

phil-davis:
> Should be addressed now.

works for me.

phil-davis:
https://drone.owncloud.com/owncloud/core/37752/158/17
```
webUIAddUsers/addUsers.feature:115
webUIAddUsers/addUsers.feature:116
```

```
Failed step: And the administrator resends the invitation email for user "guiusr1" using the webUI
        (1366.300048828125, 175.31666564941406) is out of bounds of viewport width (1360) and height (899)
```

phil-davis:
https://drone.owncloud.com/owncloud/core/37752/119/17
The scenario passes in the normal pipeline running chrome.
But fails in the pipeline that runs all the smoke-tests on Firefox.

In tests/acceptance/run.sh there is special code that tries to set the browser window big enough:
```
	if [ "${BROWSER}" == "firefox" ]
	then
		# Set the screen resolution so that hopefully draggable elements will be visible
		# FF gives problems if the destination element is not visible
		EXTRA_CAPABILITIES='"screenResolution":"1920x1080",'

		# This selenium version works for Firefox after V47
		# We no longer need to support testing of Firefox V47 or earlier
		EXTRA_CAPABILITIES='"seleniumVersion":"3.4.0",'${EXTRA_CAPABILITIES}
	fi
```

The problem is that the UI testing framework (Behat+Mink+selenium) does not successfully auto-scroll elements into view during test execution. When a UI is wide, it is easy to have elements that are off the right-hand side of the displayed area, and so are not clickable. The User Management page is wide, (lots of columns) and gets wider with this change. The failing scenario trys to click the "Resend invitation email" button, and that is way off on the right-hand edge. I am running Firefox, and I can do it manually, but I have to scroll the window all the way to the right.

IMO it is not worth trying to run this scenario on Firefox and not worth trying to adjust the screen size etc. I will skip it on Firefox.

phil-davis:
I added a couple of acceptance tests that check that creation time is returned in the provisioning API, as is visible on the webUI.

They pass locally for me. I expect that CI will pass.

phil-davis (commented):
LGTM, works locally.
Needs a developer review. The code looks OK to me, but it would be good to have another review.

lib/private/User/DeletedUser.php
```diff
@@ -102,6 +102,14 @@ public function updateLastLoginTimestamp() {
 		return true;
 	}
 
+	/**
+		  * Reset user creation time (return 0)
```
jvillafanez, line 106: Indentation is wrong here

lib/private/User/User.php
```diff
@@ -271,6 +271,25 @@ public function updateLastLoginTimestamp() {
...
+	public function getCreationTime() {
+		return (int)$this->account->getCreationTime();
+	}
+ 
+	/**
+	 * set the timestamp of the user's creation time
+	 *
+	 * @return void
+	*/
+	public function setCreationTime() {
```
jvillafanez, line 288: Should this method be implemented? I mean, I could get an user instance and update the creation time, but I don't think we should allow that.

lib/public/IUser.php
```diff
@@ -100,6 +100,22 @@ public function getLastLogin();
...
+	 */
+	public function getCreationTime();
+ 
+	/**
+	 * set the timestamp of the user's creation time
+	 *
+	 * @return void
+	 * @since 10.12.0
+	 */
+	public function setCreationTime();
```
jvillafanez, line 117: As said, I don't think this method should be public.
The creation time is handled by the account class, and the user has a dependency on the account in order to get the creation time, but it doesn't imply the user should be allowed to set the creation time.

DeepDiver1975: Mechanisms like the user sync command needs this? 

But I guess you are right - this field should be set internally on creation of an account.

I'll take care.

THX @jvillafanez

DeepDiver1975 (approved):

jvillafanez (approved):

phil-davis:
Note: unit tests for https://github.com/owncloud/configreport/pull/185 had to be adjusted to include mention of this new account data item.

jnweiger:
Confirmed in 10.12.0-rc.2

Issue #39630, opened 2021-12-31
Update Symfony from 4.4.* to 5.4.*


phil-davis:
https://symfony.com/releases/5.4

Symfony LTS major version 5.4 was release in Nov 2021. It supports PHP 7.2.5 and up - https://symfony.com/doc/5.4/setup.html#technical-requirements - that matches easily with oC10 master which currently supports PHP ~7.3 and~ 7.4.

The major version has 3 years bugfix support to Nov 2024 and another 1 year security support to Nov 2025.

Note: there are also apps that mention Symfony in their `composer.json`, so those should also be updated in sync.

Symfony 4.4 has bugfix support to Nov 2022 and security support to Nov 2023. So there is not a rush to do this issue.
https://symfony.com/releases/4.4

phil-davis:
https://symfony.com/releases

Symfony 4.4 is at the end of regular support. There is another 12 months of security-fixes-only.

https://symfony.com/blog/symfony-4-4-49-released
"WARNING: 4.4.49 is the last version for the Symfony 4.4 branch. If some of your projects are still using this version, consider upgrading as soon as possible."

phil-davis:
Apps to check if they need a release:
- [ ] activity

phil-davis:
Note: I think that the code is all done. The pending tasks are all the app releases and core 10.13.0

PR #40592, opened 2023-01-23
Add a default Event to dispatch events with just a name

phil-davis:
## Description
In Symfony 5, `eventDispatcher->dispatch` must be passed an Event as the first parameter\, followed by the event name.
In Symfony 4, an event can be dispatched with just a name (as the first parameter) or can have an Event and name (the Symfony 5 call interface)

This PR adjusts the remaining eventDispatcher->dispatch calls so that thet all pass an event as the first parameter.

## Related Issue

#39630 

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files/lib/Controller/ViewController.php
+++ b/apps/files/lib/Controller/ViewController.php
@@ -43,6 +43,7 @@ use OCP\IUserSession;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use OCP\AppFramework\Http;
 use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Contracts\EventDispatcher\Event;
 
 /**
  * Class ViewController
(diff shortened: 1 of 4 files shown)


DeepDiver1975:
So updating the dispatch to Symfony 5 will break all apps which dispatch events ..... well so be it ....

DeepDiver1975 (approved):

phil-davis:
> So updating the dispatch to Symfony 5 will break all apps which dispatch events ..... well so be it ....

Yes, the order of parameters to `dispatch` changes in Symfony5. In Symfony4 they made the code accept the parameters in both the new and the old orders.

If/when we actually go to Symfony5, we first should get the `dispatch` calls adjusted in all apps, and get those apps released. They will then work with Symfony4 (and 5).

"Some time later" we can bump core to Symfony5.

I am trying to find the obvious things that need adjusting in core now, and they can be in 10.12. That is a small step forward.

There is security support for Symfony4 to Nov 2023, so we have time to gradually sort this out.

PR #40600, opened 2023-01-24
Adjust stream chunk size to load big exif metadata

jvillafanez:
## Description
There were problems loading the exif metadata of some images if the metadata was too large. This was causing problems loading the images because the right orientation was in the metadata that wasn't loaded.

## Related Issue
https://github.com/owncloud/enterprise/issues/5500

## Motivation and Context
Users expect the image to be loaded with the right orientation, like most of the apps.

## How Has This Been Tested?
Just upload the image below:
portrait

Tested with and without encryption. I'm not aware of any additional stream wrapper that could affect the behavior (quota isn't applied to read streams)

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

Supersedes https://github.com/owncloud/core/pull/40586


lib/private/legacy/image.php
```diff
@@ -494,12 +542,61 @@ public function load($imageRef) {
...
+			$this->adjustStreamChunkSize($handle);
 			$this->loadExifData($handle);
 			return $this->resource;
 		}
 		return false;
 	}
 
+	/**
+	 * Adjust the size of the chunks in the stream. This is required for
+	 * the "exif_red_data" native method to read a whole chunk of metadata
```
phil-davis, line 554:
```suggestion
	 * the "exif_read_data" native method to read a whole chunk of metadata
```

phil-davis (commented):
changelog needed

phil-davis (commented):
Looks reasonable to me. Needs a dev to review.

lib/private/legacy/image.php
```diff
@@ -494,12 +542,61 @@ public function load($imageRef) {
...
+	 * have access to all the wrapped streams and adjust the chunk size of all
+	 * of them. A "getSource" method must be present in the custom wrapper
+	 * providing access to the underlying stream, otherwise a warning will be
+	 * logged because it could cause problems.
+	 */
+	private function adjustStreamChunkSize($handle) {
+		$stream = $handle;
+		$metadata = \stream_get_meta_data($stream);
+		while ($metadata['stream_type'] === 'user-space') {
+			\stream_set_chunk_size($stream, 64 * 1024);
```
DeepDiver1975, line 575: error handling?

DeepDiver1975 (changes requested):
My usual question: unit test? :see_no_evil:

lib/private/legacy/image.php
DeepDiver1975: ok - not really - it only fails if the chunk size is negative or bigger then php max int ....https://www.php.net/manual/en/function.stream-set-chunk-size.php

jvillafanez:
I don't think unit tests are possible in the strict sense. We need a custom stream wrapper to trigger the behavior, so we either create a custom wrapper just for the tests or use ownCloud's node or view object to get the stream.

Even if we include integration tests as unit tests, it will take some time (estimate ~2 days) and it's unlikely to cover all the cases (need to find TIFF images that trigger this behavior).

mrow4a:
> I don't think unit tests are possible in the strict sense. We need a custom stream wrapper to trigger the behavior, so we either create a custom wrapper just for the tests or use ownCloud's node or view object to get the stream.
> 
> Even if we include integration tests as unit tests, it will take some time (estimate ~2 days) and it's unlikely to cover all the cases (need to find TIFF images that trigger this behavior).

I think it would be good to cover it with some integration test. Same story as with https://github.com/owncloud/files_classifier/tree/master/tests/acceptance/data and https://github.com/owncloud/files_classifier/blob/master/tests/unit/DocumentPropertiesTest.php#L37-L67

pako81:
@jvillafanez @mrow4a do we want to have tests for this one?

jvillafanez:
If it's required, I'd vote for open a tech debt ticket to include the tests because it will take time. Note that, as said, we'd need to create a custom wrapper to trigger the behavior.

jvillafanez:
I've added a test that should cover the main use case, but there are things still not covered:
* TIFF images aren't loaded with the gd extension. This means that the code will be left untested for .tiff files. Currently, the exif data is loaded only if the image is loaded, so we'd need to change the code to load the exif data regardless of the images, but the data is pointless if we can't load the image.
* We're using the `close` stream wrapper because it seems the easiest one we can reuse. However, this will trigger a warning due to the missing `getSource` method. This is still ok for the tests because there isn't any additional custom wrapper in the stream whose chunk size needs to be adjusted. It shouldn't cause problems.

jvillafanez:
@DeepDiver1975 is this good enough?

tests/lib/ImageTest.php
```diff
@@ -338,4 +338,26 @@ public function testConvert($mimeType) {
...
+	public function testExifDataBigStream($targetFile) {
+		$img = new \OC_Image();
+
+		\OC\Files\Stream\Close::registerCallback($targetFile, function () {
+		});
+		$stream = \fopen("close://{$targetFile}", 'rb');
+
+		$img->loadFromFileHandle($stream);
+		\fclose($stream);
+		$this->assertSame(6, $img->getOrientation());  // orientation = right, top
```
DeepDiver1975, line 361: all assert functions of phpunit are static if I am not totally miss taken ....

Should be self::assertSame ...

DeepDiver1975 (changes requested):

tests/lib/ImageTest.php
phil-davis: `$this->assertSomething()` is done throughout the existing code.
But yes, the asserts are like:
```
    public static function assertSame($expected, $actual, string $message = ''): void
```

I wonder why we have done `$this->` rather than `self::` for all these years?

ownclouders:
:boom: Acceptance tests pipeline <strong>apiShareManagementToShares-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37892/52

ownclouders:
:boom: Acceptance tests pipeline <strong>apiProvisioningGroups-v1-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/37892/44

phil-davis:
I restarted CI - some pipelines failed when initially fetching docker images.

tests/lib/ImageTest.php
DeepDiver1975: Maybe because earlier versions of phpunit have been using `this->` ... no idea ..

I tend to fix such things when touching any code base ..... my idea is even throwing this as warnings into my face :shrug:

DeepDiver1975 (approved):

PR #40612, opened 2023-01-31
Fix share into share move scenario

pako81:
### Scenario:

Given three users Alice, Bob and Charlie.

Alice shares a folder X with Bob.
Bob shares a file Y with Charlie.
Bob shares a file Z via a public link.

Bob moves Y and Z into the shared folder X.

What is expected:

Alice becomes owner of the files Y and Z.
The share for Charlie on Y and the public link on Z become visible to both Alice and Bob.

What happens:

Alice becomes owner of the files Y and Z.
The share for Charlie on Y becomes visible to Alice and Bob.
The public link on Z is only visible to Bob, not visible to Alice. Owner is not updated to Alice.

## Related Issue
- https://github.com/owncloud/enterprise/issues/5565

## Motivation and Context
Public links are lost upon moving share into another share as the share owner is not correctly set.

## How Has This Been Tested?

Manually by reproducing the scenario above.

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item

@IljaN FYI.

### DIFF
--- a/apps/files_sharing/lib/Updater.php
+++ b/apps/files_sharing/lib/Updater.php
@@ -57,6 +57,7 @@ class Updater {
 
 		$shares = $shareManager->getSharesBy($userFolder->getOwner()->getUID(), \OCP\Share::SHARE_TYPE_USER, $src, false, -1);
 		$shares = \array_merge($shares, $shareManager->getSharesBy($userFolder->getOwner()->getUID(), \OCP\Share::SHARE_TYPE_GROUP, $src, false, -1));
+                $shares = \array_merge($shares, $shareManager->getSharesBy($userFolder->getOwner()->getUID(), \OCP\Share::SHARE_TYPE_LINK, $src, false, -1));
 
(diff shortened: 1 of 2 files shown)


pako81:
Note this is only fixing Scenario I as of now.  For Scenario II the issue seems to be around https://github.com/owncloud/core/blob/master/apps/files_sharing/lib/Updater.php#L54 as the path returned resolves to the root folder so that shares located inside subfolders are not found and, therefore, not correctly moved.

pako81:
In case we want to have this partial fix in 10.12.0.

IljaN (approved):

pako81:
@jvillafanez @phil-davis are you ok with merging this?

jvillafanez (approved):

phil-davis:
What to do about tests? There is some change (improvement) in behavior, so there should be a test that would have failed previously and now passes.

PR #40617, opened 2023-02-03
Fix two-factor enable over API

pako81:
## Description

Fix reactivation of two-factor auth when using provisioning API.

## Motivation and Context
Using the provisioning API for reenabling two-factor auth for an user is currently broken.

1. Running:

```
curl -X PUT http://admin:admin@example.com/ocs/v1.php/cloud/users/user1 -d key="two_factor_auth_enabled" -d value="false"
```

correctly disable two-factor auth for that user. The configkey `two_factor_auth_disabled` is correctly set to the configvalue `1` in `oc_preferences` and user can normally login with username/password.

2. However, running:

 ```
curl -X PUT http://admin:admin@example.com/ocs/v1.php/cloud/users/user1 -d key="two_factor_auth_enabled" -d value="true"
```

does not correctly reenable two-factor auth for the user. The configkey `two_factor_auth_disabled` is not removed from the `oc_preferences` table and two-factor is still not enforced.

## How Has This Been Tested?

Manually: by following the steps above.

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item

### DIFF
--- a/apps/provisioning_api/lib/Users.php
+++ b/apps/provisioning_api/lib/Users.php
@@ -320,7 +320,7 @@ class Users {
 				}
 				break;
 			case 'two_factor_auth_enabled':
-				if ($parameters['_put']['value'] === true) {
+				if ($parameters['_put']['value'] === 'true') {
 					$this->twoFactorAuthManager->enableTwoFactorAuthentication($targetUser);
 				} else {
 					$this->twoFactorAuthManager->disableTwoFactorAuthentication($targetUser);
--- a/apps/provisioning_api/tests/UsersTest.php
+++ b/apps/provisioning_api/tests/UsersTest.php
@@ -1154,7 +1154,7 @@ class UsersTest extends OriginalTest {
 		$this->setupIsUserAccessibleMock($loggedInUser, $targetUser, false);
 
 		$expected = new Result(null, 100);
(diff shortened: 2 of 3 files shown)


DeepDiver1975 (changes requested):
change log is missing - will add it

DeepDiver1975 (approved):

phil-davis (approved):

jnweiger:
Confirmed fixed in 10.12.0-rc.2
`-d key="two_factor_auth_enabled" -d value="true"` works correctly, however
`-d key="two_factor_auth_enabled" -d value="yes"` or
`-d key="two_factor_auth_enabled" -d value="1"` dont. the 'two_factor_auth_disabled' configkey' remains at 1.

pako81:
AFAICT `two_factor_auth_enabled` is currently the only boolean value when editing the user https://github.com/owncloud/core/blob/master/apps/provisioning_api/lib/Users.php#L295-L339. Do we want to also cover the case `yes/no` and `1/0` ?

DeepDiver1975:
Boolean values are true and false.
No need to support anything else

PR #40619, opened 2023-02-06
[full-ci][tests-only]Try `JSON` response drilling with library

SagarGi:
### Description
This PR is the json drill down of the response data using `helmich/phpunit-json-assert` library which is alternative process of this PR https://github.com/owncloud/core/pull/40562

For this Drill down of the `datatable`, i have only formatted the when request to `json` and checked the response with json schema (expected). 
This PR covers for suite:
`apiCapabilities`

***Notes:***
- The only thing i doubt about using this library is if the json check or drilling is to long then the response to match value is also too  long that might decrease the readability of the feature file which is actually is the key point to a feature file.

But also has many benefits over this PR https://github.com/owncloud/core/pull/40562.

### Related Issue:
https://github.com/owncloud/core/issues/39796


SagarGi:
***Notes:***
- The only thing i doubt about using this library is if the json check or drilling is to long then the response to match value is also too  long that might decrease the readability of the feature file which is actually is the key point to a feature file.

But also has many benefits over this PR https://github.com/owncloud/core/pull/40562.

tests/acceptance/features/bootstrap/FeatureContext.php
```diff
@@ -1640,6 +1641,41 @@ public function theHTTPStatusCodeShouldBe($expectedStatusCode, ?string $message
...
+	 * @return void
+	 *
+	 * @throws Exception
+	 */
+	public function theDataOfTheResponseShouldMatch(
+		PyStringNode $schemaString
+	): void {
+		$responseXml = $this->getResponseXml()->data;
+		$jsonResponse = json_encode($responseXml);
+		$arrayResponse = json_decode($jsonResponse);
```
individual-it, line 1672: that does not feel right to have the XML response checked with JSON.
I would suggest to check the JSON responses with this system. e.g. the capabilities can be retrieved as JSON by adding `?format=json` to the URL

tests/acceptance/features/apiCapabilities/jsonDrillExample.feature
```diff
@@ -0,0 +1,445 @@
...
+#      | files_sharing | public@@@social_share                 | 1                 |
+#      | files_sharing | resharing                             | 1                 |
+#      | files_sharing | federation@@@outgoing                 | 1                 |
+#      | files_sharing | federation@@@incoming                 | 1                 |
+#      | files_sharing | group_sharing                         | 1                 |
+#      | files_sharing | share_with_group_members_only         | EMPTY             |
+#      | files_sharing | user_enumeration@@@enabled            | 1                 |
+#      | files_sharing | user_enumeration@@@group_members_only | EMPTY             |
+#      | files         | bigfilechunking                       | 1                 |
+    And the data of the response should match
```
individual-it, line 31:
```suggestion
    And the JSON data of the response should match
```

individual-it (changes requested):

SagarGi:
Some findings to be able to implement the above comment https://github.com/owncloud/core/pull/40619#discussion_r1098208051 .

1. Using `?format=json` at each endpoint of an API, we might need to refactor test suites according to it like i have done in this PR.
2. Also some API responses does not always doesn't give `json` so we will have to adjust the test both in `xml` and `json` as per requirements.  

need some suggestion towards it @individual-it @phil-davis

tests/acceptance/features/apiCapabilities/capabilitiesWithNormalUser.feature
```diff
@@ -12,40 +12,310 @@ Feature: default capabilities for normal user
...
+#      | files_sharing | share_with_group_members_only             | EMPTY             |
+#      | files_sharing | share_with_membership_groups_only         | EMPTY             |
+#      | files_sharing | auto_accept_share                         | 1                 |
+#      | files_sharing | user_enumeration@@@enabled                | 1                 |
+#      | files_sharing | user_enumeration@@@group_members_only     | EMPTY             |
+#      | files_sharing | user@@@send_mail                          | EMPTY             |
+#      | files         | bigfilechunking                           | 1                 |
+#      | files         | privateLinks                              | 1                 |
+#      | files         | privateLinksDetailsParam                  | 1                 |
+    And the JSON data of the response should match
```
ScharfViktor, line 52: My question (maybe stupid):
Can I compare not all fields? Or `the JSON data of the response` be identical to the response?

tests/acceptance/features/apiCapabilities/capabilitiesWithNormalUser.feature
```diff
@@ -12,40 +12,310 @@ Feature: default capabilities for normal user
...
+#      | files_sharing | group_sharing                             | 1                 |
+#      | files_sharing | share_with_group_members_only             | EMPTY             |
+#      | files_sharing | share_with_membership_groups_only         | EMPTY             |
+#      | files_sharing | auto_accept_share                         | 1                 |
+#      | files_sharing | user_enumeration@@@enabled                | 1                 |
+#      | files_sharing | user_enumeration@@@group_members_only     | EMPTY             |
+#      | files_sharing | user@@@send_mail                          | EMPTY             |
+#      | files         | bigfilechunking                           | 1                 |
+#      | files         | privateLinks                              | 1                 |
+#      | files         | privateLinksDetailsParam                  | 1                 |
```
ScharfViktor, line 51: you left that for comparison, right?

SagarGi: yes !!

tests/acceptance/features/apiCapabilities/capabilitiesWithNormalUser.feature
SagarGi: yes we can send the expected JSON according to our requirement. It is not mandatory we check all the json response.

phil-davis:
For test scenarios where we check all (or almost all) items in the JSON response, using `assertJsonDocumentMatchesSchema` seems OK to me. The schema that goes in the test scenario is a bit of a long thing. But these tests are technical tests of the content of API responses anyway. The scenarios in the feature files only need to be readable by reasonably-technical people anyway, not "mere mortal" users.

But we also have various tests that individually check that the system-under-test can change something, and the change is seen in the API response. This sort of thing:
```
  Scenario: lock file action can be enabled
    Given parameter "enable_lock_file_action" of app "files" has been set to "yes"
    When the administrator retrieves the capabilities using the capabilities API
    Then the OCS status code should be "100"
    And the capabilities should contain
      | capability | path_to_element                 | value |
      | files      | file_locking_support            | 1     |
      | files      | file_locking_enable_file_action | 1     |
```
For these type of scenarios, we only care that the particular items in the response exist and have the expected values.
We could use `assertJsonValueMatches` for that, looping through each item in "the capabilities should contain".
And the table used for the step could have the items in JSON-path format
```
    And the capabilities should contain
      | JSONpath                                            | value |
      | $.files.file_locking_support                  | 1     |
      | $.files.file_locking_enable_file_action | 1     |
```

phil-davis:
Actually, we can probably keep the scenarios in close to their current format, and just replace `capability` and `path_to_element` columns with `JSONpath`. And write the JSON path of each element.

And the underlying code can just check each JSON path using `assertJsonValueMatches`

And adjust the words and underlying code for steps like:
```
Then the capabilities setting of "files_sharing" path "group_sharing" should be "1"
```
to something like:
```
Then the capabilities setting "$.files_sharing.group_sharing" should be "1"
```
That would be just an alternative step format for:
```
Then the capabilities should contain
  | JSONpath | value |
  | $.files_sharing.group_sharing | 1 |
```

I am not sure if we actually really need or want this alternative step. It might be easiest to just refactor those so that they are all using `Then the capabilities should contain`

individual-it:
Anyway we don't have to write down every detail of the structure into the schema, only what we care about. 
```
{
"name": "Phil",
"surname": "Davis"
}
```
would be matched by the schema:
```
{
   "type": "object",
   "required": [ "name"],
   "properties": [
        "name": [ "type": "string", "enum": ["Phil"] ]
    ]
}
```

phil-davis:
> Anyway we don't have to write down every detail of the structure into the schema, only what we care about

If that is true, then great. How fussy is `assertJsonDocumentMatchesSchema` ?

If the JSON has all the things mentioned in the expected schema, plus more, does `assertJsonDocumentMatchesSchema` still pass?

phil-davis:
OK - then `assertJsonDocumentMatchesSchema` can do enough flexible stuff for us to just use that.

SagarGi:
> OK - then `assertJsonDocumentMatchesSchema` can do enough flexible stuff for us to just use that.

@phil-davis @individual-it If so then i will be make some refactor and make this as a sample PR for suite `apiCapabilities` we can check other suites where such drilling is done and adjust accordingly.

individual-it (commented):
LGTM. @ScharfViktor  @phil-davis what do you think? Should we merge this and refactor other tests also?
Or rather just use that for new tests?

ScharfViktor:
> LGTM. @ScharfViktor @phil-davis what do you think? Should we merge this and refactor other tests also? Or rather just use that for new tests?

Are we introducing new tests into the core?
For ocis, I would use this approach for contract tests, get rid of "@@@" and use in new tests

individual-it:
No, we are not adding any tests to core (except if there are new features).
I meant to say: merge this here, bring the same change to ocis, and use it in ocis for new tests

ScharfViktor:
> No, we are not adding any tests to core (except if there are new features). I meant to say: merge this here, bring the same change to ocis, and use it in ocis for new tests

I totally agree.

phil-davis (approved):
LGTM
I think that the commits should be squashed, it will be a bit confusing in future if we have such a long list of commits merged in master, if someone is trying to work out what happened.

And then we can:
- do similar in ocis
look at the other uses of `@@@` in the acceptance tests and design a way to refactor those away

phil-davis:
@individual-it please review again. I think that we can squash and then merge this.

individual-it (approved):

PR #40643, opened 2023-02-17
Flatten folder structure from drag & drop for public file upload

jnweiger:
This allows drag & drop to work with folders too. The code change is minimal, as all the code top pick up files from subdirectories is already there. It only missed the step to actually remove subdirecory levels from the path before uploading.

## Description
Add missing piece for folder drag & drop with public file drop.
Drag & drop uses javascript code to generate a list of files.
With one file or multiple files selected, this is a list of filenames and disregards the folders where they came from.
With one ore multiple folders selected, the list of filenames includes relative pathnames, the selected folder and potentially subdorectories within that folder.

In case of normal file view the uploader implements creating of the subdirectory levels so that the structure of the selected tree is faithfully reproduced on the server.
In case of a public file drop, subdirectories cannot be created. And as agreed in https://github.com/owncloud/enterprise/issues/5489#issuecomment-1430908616 it is also not wanted.

File Drop as implemented in PublicFileView.js already uses the same mechanics as normal drag drop fir files view, so 99% of the code for folder drop is already present. The missing piece is the subfolder problem. 
It is actually sufficient, to remove the subfolders (as stored in file.relativePath) and the upload succeeds.

## Related Issue

- Fixes https://github.com/owncloud/enterprise/issues/5489

## Motivation and Context

Folder upload never worked on Public file Drop. 

## How Has This Been Tested?

Manual testing with files, folders, multiple files, multiple folders selected in drag and drop; 
both with public file drop view and normal file view.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/files_sharing/js/PublicUploadView.js
+++ b/apps/files_sharing/js/PublicUploadView.js
@@ -73,6 +73,15 @@
 		},
 
 		_onUploadBeforeAdd: function(upload) {
+			// Flatten out any subdirectories.
+			// When the user wants to upload folder via drag and drop, the foldername
+			// (and possible further subdirectores) are in relativePath.
+			// This will cause uploader.ensureFolderExists() to fail, as it picks up the
+			// extra path components via getFullPath().
+			// Subdirectories are not needed here. We want all files in the upload folder, flat.
+			// To allow folder upload via drag and drop, we only need to clear out relativePath.
+			// This works nicely together with the autorename feature.
+			upload.getFile().relativePath = '';
 			// add autorename header to deduplicate names on the server in case of conflict
 			upload.setConflictMode(OC.FileUpload.CONFLICT_MODE_AUTORENAME_SERVER);
 		},
--- a/apps/files_sharing/js/PublicUploadView.js
+++ b/apps/files_sharing/js/PublicUploadView.js
@@ -73,6 +73,16 @@
 		},
 
 		_onUploadBeforeAdd: function(upload) {
+                       var file = upload.getFile();
+                        if (file.isDirectory) {
+                               // We don't create folders in PublicUploadView.
+                                // One case has been observed, where we enter here:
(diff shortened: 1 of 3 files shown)


jvillafanez (approved):

mmattel:
@phil-davis we possibly need a test and a changelog, could you assist? 😄

phil-davis:
> @phil-davis we possibly need a check if the test is sufficient and a changelog, could you or someone of your team assist? smile

We can't do drag-and-drop in automated tests, because the browser-test functionality can't "get outside the browser" to do that sort of thing. So we need to manually test this sort of thing.

The change is going to work. My question is "is this behavior the required/acceptable behavior?" "public" users will drop a folder into a public upload-only link browser window, and the whole folder structure will be flattened into the public-link folder. If it is just a few files then I suppose that will be OK for the real user that made the public upload-only link. But when someone drops a folder that has a lot of files in it, the result will be difficult for the real user to realize what happened. And specially confusing if multiple "public users" do this.

It needs someone (Product Manager?) to confirm that this solution is acceptable.

jnweiger:
It is the agreed solution. Customer expectation is already set in this way.

phil-davis:
Works as designed. When I drag-and-drop a folder that has various files and sub-folders then all the files throughout the tree are uploaded "flat" into the the "file drop" folder on the server. If I do different uploads and file names are duplicated, then the 2nd files with the same name all get "(2)" on the end of their names.

phil-davis (approved):

phil-davis:
Note: when you "click to upload" and the file explorer window opens for you to find the file to upload, when you click folders to navigate, they open, just like they did before. You can't upload a whole folder using "click to upload", which is fine.

Issue #40649, opened 2023-02-21
Compatibility issue after update to OC10.11


lukasmatusiewicz:
The new "guzzle/guzzle" version doesn't allow adding arrays into the "body" request option anymore. This dependency update causes a compatibility issue for the apps, that implement the IClientService and are sending the requests using the `body` option.
E.g.: https://github.com/privacyidea/privacyidea-owncloud-app/issues/93.

Using the "form_params" option fixes this problem for us, but I found no information about it in OC documentation.

This is an example of the property-built POST request placed in the PHPDoc comment in "oc-core/lib/public/Http/Client/IClient.php" :
```
'body' => [
    'field' => 'abc',
    'other_field' => '123',
    'file_name' => fopen('/path/to/file', 'r'),
    ],
```
This way to forward the arrays doesn't work anymore (from v10.11).

phil-davis:
Hmmm, yes. `guzzlehttp/guzzle/src/Client.php` `public function requestAsync()` has:
```
        if (\is_array($body)) {
            throw $this->invalidBody();
        }
```
Looking at other places like `guzzlehttp/psr7/src/ServerRequest.php` the body needs to be passed as `string|resource|StreamInterface|null` - that is, something that is a "string/stream of bytes" that can be directly sent in the request.

phil-davis:
I can see these existing bits of code that I suspect will n ot work:
https://github.com/owncloud/core/blob/master/apps/federation/lib/BackgroundJob/RequestSharedSecret.php#L143
```
			$result = $this->httpClient->post(
				$target . $this->endPoint,
				[
					'body' => [
						'url' => $source,
						'token' => $token,
					],
					'timeout' => 3,
					'connect_timeout' => 3,
				]
			);
```

https://github.com/owncloud/core/blob/master/lib/private/HTTPHelper.php#L107
```
			$response = $client->post(
				$url,
				[
					'body' => $fields,
					'connect_timeout' => 10,
				]
			);
```

https://github.com/owncloud/market/blob/master/lib/HttpService.php#L304
```
	private function httpPost($path, $options) {
		$ca = $this->config->getSystemValue('marketplace.ca', null);
		if ($ca !== null) {
			$options = \array_merge(
				[
					'verify' => $ca
				],
				$options
			);
		}
		$client = $this->httpClientService->newClient();

		try {
			$response = $client->post($path, $options);
		} catch (TransferException $e) {
			throw new AppManagerException(
				$this->l10n->t(
[...]

PR #40652, opened 2023-02-22
Adjust parameters for guzzle7

jvillafanez:
## Description
Fixes some bugs caused by the guzzle upgrade
* I'm not sure about the federated sharing notifications. Request should reach the server, but I had problems with the share. I don't know if there is a bug lurking around, or it was caused by any of my changes in order to trigger the behavior. Anyway, it seems out of scope.
  I assume this was left because the default behavior is to use OCM, not the old mechanism. With OCM, the parameters use "json", so there is no problem.
* The federation background job seems awkward and likely untested. That's a bug that is now fixed
* For the HTTPHelper, I think it's deprecated and it's only used in the old sharing 1.0. I don't think we use sharing 1.0 anywhere. It seems dead code, but anyway, the issue with guzzle should be fixed, although not tested.

So it seems the only real bug, at least in core, is with the background job of the federation app. The other 2 shouldn't trigger.

## Related Issue
https://github.com/owncloud/core/issues/40649

## Motivation and Context
Issue with the background job needs to be fixed. There is also comments that needs to be updated.

## How Has This Been Tested?
Manually tested except the one for the HTTPHelper, which is a blind change.

## Types of changes

- [ ] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/lib/Notifications.php
+++ b/apps/federatedfilesharing/lib/Notifications.php
@@ -355,7 +355,7 @@ class Notifications {
 					'timeout' => 10,
 					'connect_timeout' => 10,
 				];
-				$sendAs = $useOcm === true ? 'json' : 'body';
+				$sendAs = $useOcm === true ? 'json' : 'form_params';
 				$options[$sendAs] = $fields;
 				$response = $client->post($endpoint, $options);
 				$result['result'] = $response->getBody();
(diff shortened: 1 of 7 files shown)


jvillafanez:
I don't think we need a changelog item for this.

jnweiger (approved):
Looks good to me. Although I am half blind.

phil-davis (commented):
Looks good

Issue #40665, opened 2023-03-03
[QA] public links do not work with files_primary_s3


jnweiger:
Seen with core 10.12.0-rc.2

* Install server with files_primary_s3
* Create a public link. (file drop, read/write, read only does not matter.)
* paste the public link into a private browser window.
* An error occurs:
![grafik](https://user-images.githubusercontent.com/1108546/222854110-76033dd0-d450-46dd-850c-24d44b2c7824.png)

The log says:
```
{
  "reqId": "ZAJ-mnI5zr-f5hMlA2iXnQAAAAg",
  "level": 3,
  "time": "2023-03-03T23:15:38+00:00",
  "remoteAddr": "46.114.208.108",
  "user": "--",
  "app": "index",
  "method": "GET",
  "url": "/s/KuoxwuxY0uiyBkh",
  "message": "Exception: {
\"Exception\":\"Error\",
\"Message\":\"Call to a member function getFileInfo() on null\",
\"Code\":0,
\"Trace\":\"#0 \\/var\\/www\\/owncloud\\/lib\\/private\\/Files\\/ObjectStore\\/ObjectStoreStorage.php(438): OC\\\\Files\\\\Filesystem::getFileInfo()\\n
#1 \\/var\\/www\\/owncloud\\/lib\\/private\\/Files\\/ObjectStore\\/ObjectStoreStorage.php(454): OC\\\\Files\\\\ObjectStore\\\\ObjectStoreStorage->getTotalUsedStorage()\\n
#2 \\/var\\/www\\/owncloud\\/lib\\/private\\/Files\\/Storage\\/Wrapper\\/Wrapper.php(335): OC\\\\Files\\\\ObjectStore\\\\ObjectStoreStorage->free_space()\\n
#3 \\/var\\/www\\/owncloud\\/lib\\/private\\/Files\\/Storage\\/Wrapper\\/Availability.php(350): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Wrapper->free_space()\\n
#4 \\/var\\/www\\/owncloud\\/lib\\/private\\/Files\\/Storage\\/Wrapper\\/Wrapper.php(335): OC\\\\Files\\\\Storage\\\\Wrapper\\\\Availability->free_space()\\n
[...]

PR #40669, opened 2023-03-06
revert logic to expose free_space, but keep availableStorage config

mrow4a:
- fixes https://github.com/owncloud/core/issues/40665
- related https://github.com/owncloud/enterprise/issues/5006 , https://github.com/owncloud/core/pull/40389
- needed for https://github.com/owncloud/metrics/pull/159

The issue here is that originally when doing https://github.com/owncloud/core/pull/40389 I somehow mixed the storage_id for S3 bucket with the user urns storage_id's . So when matching with filecache I though it represents that one. 

However I was wrong, and used the wrong storage .. also the correct filecache entry for the entire bucket is always showing 0 size. 

In this PR, I just expose the config entry for availableStorage, so that apps could use it. Please note that this PR reverts logic to expose this on storage level, so all "quota" calculations wont work. If we want to make it work on storage level, something like background job that calculates this size would need to be done. However, metrics app does not require that complexity and it can simply just use that parameters as in https://github.com/owncloud/metrics/pull/159

PR #40670, opened 2023-03-06
revert logic to expose free_space, but keep availableStorage config

phil-davis:
This is a copy of PR #40669 with the conflict resolved.

The original PR text was:

- fixes https://github.com/owncloud/core/issues/40665
- related https://github.com/owncloud/enterprise/issues/5006 , https://github.com/owncloud/core/pull/40389
- needed for https://github.com/owncloud/metrics/pull/159

The issue here is that originally when doing https://github.com/owncloud/core/pull/40389 I somehow mixed the storage_id for S3 bucket with the user urns storage_id's . So when matching with filecache I though it represents that one. 

However I was wrong, and used the wrong storage .. also the correct filecache entry for the entire bucket is always showing 0 size. 

In this PR, I just expose the config entry for availableStorage, so that apps could use it. Please note that this PR reverts logic to expose this on storage level, so all "quota" calculations wont work. If we want to make it work on storage level, something like background job that calculates this size would need to be done. However, metrics app does not require that complexity and it can simply just use that parameters as in https://github.com/owncloud/metrics/pull/159


mrow4a:
@phil-davis why do we duplicate this? I fixed my PR..

phil-davis:
> @phil-davis why do we duplicate this? I fixed my PR..

GitHub reports a conflict with your PR. You can rebase your PR and resolve the conflict.

phil-davis:
I see that you have rebased a few minutes ago - good.

jnweiger:
I'd like to have this in the release-10.12.0-branch -- 
Initially, i wanted to retarget the original PR, but then I saw that @phil-davis already made a 'duplicate' tpointing into the release branch. 

So fine with me to have them both, one into master, an one into the release branch.

phil-davis:
Sorry for the confusion. @mrow4a changed his PR to target `release-10.12.0` just after I made this PR.

But I see that @jnweiger changed that PR #40669 back to target `master` late yesterday.

IMO we should just merge something to `release-10.12.0` and it will get merged into `master` later. If we merge PRs into both branches then we are going to get annoying conflicts later, because the changelog files might be in different places.

phil-davis:
Somebody please review, and we can merge...

mrow4a:
@phil-davis I did not change branch in my PR, I just rebased when seeing conflicts.

mrow4a:
@phil-davis @jnweiger please note that I also squashed commits. Be careful not to diverge PRs too much, because it could lead to mismatch between oc-master and oc-10-12

phil-davis:
@jnweiger @pako81 please decide what is happening and merge something.

jvillafanez:
Changes in the l10n files don't look good.

It might be better to start over. Get a fresh branch from the latest release-10.12.0 branch, put the commits there and open a new PR against the release branch.

The changes in the logic are ok since we need to fix the issue.

PR #40674, opened 2023-03-07
revert logic to expose free_space, but keep availableStorage config

phil-davis:
This is a copy of PR #40669 with just 1 commit and targeted to `release-10.12.0`

The original PR text was:

- fixes https://github.com/owncloud/core/issues/40665
- related https://github.com/owncloud/enterprise/issues/5006 , https://github.com/owncloud/core/pull/40389
- needed for https://github.com/owncloud/metrics/pull/159

The issue here is that originally when doing https://github.com/owncloud/core/pull/40389 I somehow mixed the storage_id for S3 bucket with the user urns storage_id's . So when matching with filecache I though it represents that one. 

However I was wrong, and used the wrong storage .. also the correct filecache entry for the entire bucket is always showing 0 size. 

In this PR, I just expose the config entry for availableStorage, so that apps could use it. Please note that this PR reverts logic to expose this on storage level, so all "quota" calculations wont work. If we want to make it work on storage level, something like background job that calculates this size would need to be done. However, metrics app does not require that complexity and it can simply just use that parameters as in https://github.com/owncloud/metrics/pull/159

### DIFF
--- a/changelog/10.12.0_2023-02-24/40389
+++ b/changelog/10.12.0_2023-02-24/40389
@@ -1,6 +1,9 @@
 Change: Allow specifying available space for objectstorages
 
-Before this change, objectstorages were reporting only infinite storage space. This could have caused problems in other apps that rely on this storage method, e.g. metrics app that monitors available space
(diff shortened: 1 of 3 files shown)


jvillafanez (approved):

mrow4a (approved):

pako81:
nothing more to be done, right? merge this?

PR #40677, opened 2023-03-08
fix: in case root on CacheJail is an empty string we should not prefi…

DeepDiver1975:
…x the path with a /

## Description

- In case `root` is an empty string this method creates a path with a leading `/`
- this path is used when querying the table `oc_file_cache`
- paths in `oc_file_cache` a never store with a leading `/`
- as a result queries never hit entries which results in chaos
- `class Jail` is already using the same implementation: https://github.com/owncloud/core/blob/0efca5ee4e44340907925d58527e65eda7d63f03/lib/private/Files/Storage/Wrapper/Jail.php#L55

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

Screenshot from 2023-03-08 09-40-21

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Cache/Wrapper/CacheJail.php
+++ b/lib/private/Files/Cache/Wrapper/CacheJail.php
@@ -49,9 +49,12 @@ class CacheJail extends CacheWrapper {
 	protected function getSourcePath($path) {
 		if ($path === '') {
 			return $this->root;
-		} else {
-			return $this->root . '/' . \ltrim($path, '/');
 		}
+		if ($this->root === '') {
+			return \ltrim($path, '/');
+		}
+
+		return $this->root . '/' . \ltrim($path, '/');
 	}
 
 	/**


IljaN (approved):
LGTM, @jvillafanez could you please also take a quick look for potential side-effects?

jvillafanez:
Are we ensuring that the `$this->root` never starts with "/"? We probably should add a comment saying that it's expected.

Maybe we should switch to a string building approach instead of returning different things. Something like:
```
$root = \trim($this->root, '/');
if ($root === '') {
  return \trim($path, '/');
}
return $root . "/" . \trim($path, '/');
```
This seems easier to follow to me.

In any case, other than fixing things, it shouldn't cause new bugs. I'm not aware of any wrapper modifying paths, maybe the encoding wrapper, but it shouldn't be related with this issue. We probably should double-check all the wrappers anyway to ensure they don't require any change and that they follow the same approach.

jnweiger:
CI fails with
```
...
Default module: OC_DEFAULT_MODULE
+ php occ encryption:select-encryption-type masterkey --yes
  Command "encryption:select-encryption-type" is not defined.  
...
```

DeepDiver1975:
All core PRs will fail due to recent changes in encryption app

IljaN:
@DeepDiver1975 Which changes exactly? Maybe we can take a look and fix?

DeepDiver1975:
> Maybe we can take a look and fix?

I am on it ....

jnweiger:
@DeepDiver1975 what do we do with
https://github.com/owncloud/core/blob/fix/jail-cache-source-path-root-empty/tests/acceptance/features/bootstrap/EncryptionContext.php#L93-L95

DeepDiver1975:
-> https://github.com/owncloud/core/pull/40712

jnweiger:
@DeepDiver1975 @phil-davis 
Drone says 'Killed', but I cannot see any error.

DeepDiver1975:
> Drone says 'Killed', but I cannot see any error.

not relevant until #40712 is merged

phil-davis:
> Do we need a rebase here?

Yes, and the resulting new CI run should pass.

pako81:
CI is now green. @DeepDiver1975 do we need a changelog here?

DeepDiver1975:
> do we need a changelog here?

I don't think so

pako81:
should we still consider what JP suggested here https://github.com/owncloud/core/pull/40677#issuecomment-1473625823 or is it ok to merge this like it is?

DeepDiver1975:
@jvillafanez you are welcome to change the code as you prefer - but let's get this merged now.

THX

PR #40693, opened 2023-03-21
Use getHome in ChunkLocationProvider.php

pako81:
## Description
Use `getHome()` in ChunkLocationProvider.php

## Related Issue
- https://github.com/owncloud/enterprise/issues/5647

## Motivation and Context
When using the `home_folder_naming_rule` attribute for defining the home directories for LDAP users (configurable over the LDAP wizard) chunks of users' uploads are wrongly created under the default data directory rather than inside the configured home directory.

This is because currently  https://github.com/owncloud/core/blob/v10.12.0/apps/dav/lib/Upload/ChunkLocationProvider.php#L67-L70 is missing a check for such cases. It seems therefore better to rely on the `getHome()` method for getting the user's home.

## How Has This Been Tested?
Manually by setting the `home_folder_naming_rule` attribute for defining the home directories for LDAP users and observing chunks are now uploaded inside the configured home directory. 

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Changelog item

### DIFF
--- a/apps/dav/lib/Upload/ChunkLocationProvider.php
+++ b/apps/dav/lib/Upload/ChunkLocationProvider.php
@@ -63,7 +63,7 @@ class ChunkLocationProvider implements IMountProvider {
 		$chunkBaseDir = $this->config->getSystemValue('dav.chunk_base_dir', '');
 		if ($chunkBaseDir === '') {
 			// this is needed as otherwise the returned mount point will be in the home storage rather
-			// than in the "local" external storage and this would cause troubles when applying quota.
+			// than in the "local" external storage and this would cause trouble when applying quota.
 			$cacheDir = $user->getHome();
 			return [
 				new MountPoint(Local::class, '/' . $user->getUID() . '/uploads', ['datadir' => $cacheDir . '/uploads'], $loader)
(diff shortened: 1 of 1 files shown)


jvillafanez (approved):

phil-davis:
> @phil-davis I guess this does not require changelog and/or tests as it relates to a change introduced in #40567 ?

That change is already released in 10.12.0. So this is a fix to already-released code. We should put a changelog about the fix, because the behavior in 10.12.0 and the next release (10.12.1 or 10.13.0) will be different.

Also, @pako81 if/when we get a few fixes related to 10.12.0 you can decide if we make a 10.12.1 release to get those "out the door" or if they wait until 10.13.0

pako81:
> That change is already released in 10.12.0. So this is a fix to already-released code. We should put a changelog about the fix, because the behavior in 10.12.0 and the next release (10.12.1 or 10.13.0) will be different.

makes sense - will do.

> Also, @pako81 if/when we get a few fixes related to 10.12.0 you can decide if we make a 10.12.1 release to get those "out the door" or if they wait until 10.13.0

Yes, actually we are evaluating together with @jnweiger if this fix would already justify a 10.12.1 release or if we should rather wait. I would probably already go with a 10.12.1.

grgprarup:
@pako81 The WebUI tests are failing in the firewall app after this PR merged. Could you have a look into it?
CC @phil-davis

PR #40701, opened 2023-03-24
fix permission bits when enforcing rw links

### LINKED ISSUE #40699: [QA] Public link on a file with Download/View/Edit ignores password enforcement
Seen with 10.12.0 (and also 10.11.0, not a regression)

* Visit the admin UI -> sharing, and check all four available password enforcements.
grafik

----
* Create a file public.txt and add a public share with read only permissions.
* Try to leave the password field empty. An error occurs. OK
grafik

----
* Create a file public.txt and add a public share with download edit permissions.
* Leave the password field empty. The share is created. BAD.
* Access the share through the public link, the file can be downloaded. BAD.
grafik

(Public links on folders work fine. Each of the four link types has its own enforce setting.)

jnweiger:
## Description
The bits READ UPDATE CREATE DELETE are used to differentiate the differnrt types of shares.
passwordMustBeEnforced() maps these bits to the read-only, read-write, read-write-delete, write-only settings.
This code is used for both, files and foldes. with files, the logic was wrong, in such a way, that no setting would enforce passwords on a Download/View/Edit type file share.

## Related Issue

- Fixes #40699 

## How Has This Been Tested?
* Tested wit no enforcement: all file and folder link types can be set without a password.
* Tested with exactly one type enforced, (all four after another): the correct folder link type is enforced; the two file link types correspond to the first two enforcement settings.
* Tested with all types enforced: No file or folder link can be created without a password.

- test environment:
keyboard + hands + eyes

## Types of changes
- [ ] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket raised: <link> 

### DIFF
--- a/core/js/sharedialoglinkshareview.js
+++ b/core/js/sharedialoglinkshareview.js
@@ -116,14 +116,31 @@
 			return (permissions) ? parseInt(permissions, 10) : OC.PERMISSION_READ;
 		},
 
+		/**
+		 * These are all cases as per OCA.Share.ShareDialogLinkShareView.render,
+		 * matching passwordMustBeEnforced from server side
+		 * 
+		 * @returns {boolean}
+		 */
 		_shouldRequirePassword: function() {
-			// matching passwordMustBeEnforced from server side
 			var permissions = this._getPermissions();
-			var roEnforcement = permissions === OC.PERMISSION_READ && this.configModel.get('enforceLinkPasswordReadOnly');
-			var woEnforcement = permissions === OC.PERMISSION_CREATE && this.configModel.get('enforceLinkPasswordWriteOnly');
-			var rwEnforcement = (permissions === (OC.PERMISSION_READ | OC.PERMISSION_CREATE) && this.configModel.get('enforceLinkPasswordReadWrite'));
-			var rwdEnforcement = (permissions === (OC.PERMISSION_READ | OC.PERMISSION_UPDATE | OC.PERMISSION_CREATE | OC.PERMISSION_DELETE) && this.configModel.get('enforceLinkPasswordReadWriteDelete'));
-			if (roEnforcement || woEnforcement || rwEnforcement || rwdEnforcement) {
+
+			// Download / View (file and folder)
+			var publicRead = permissions === OC.PERMISSION_READ && this.configModel.get('enforceLinkPasswordReadOnly');
+
+			// Upload only (File Drop folder)
+			var publicUploadFolder = permissions === OC.PERMISSION_CREATE && this.configModel.get('enforceLinkPasswordWriteOnly');
+
+			// Download / View / Upload (folder)
+			var publicReadUploadFolder = (permissions === (OC.PERMISSION_READ | OC.PERMISSION_CREATE) && this.configModel.get('enforceLinkPasswordReadWrite'));
+
+			// Download / View / Upload / Edit (folder)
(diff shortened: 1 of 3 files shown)


ownclouders:
:boom: Acceptance tests pipeline <strong>webUISettingsMenu-chrome-mariadb10.2-php7.4</strong> failed. The build has been cancelled.

https://drone.owncloud.com/owncloud/core/38131/133

phil-davis:
I restarted CI. At least one of the pipelines had a problem fetching its docker images.

phil-davis:
https://drone.owncloud.com/owncloud/core/38138/9/7
```
There was 1 failure:

1) Test\Share20\ManagerTest::testPasswordMustBeEnforcedForReadWrite

/drone/src/tests/lib/Share20/ManagerTest.php:667
```

pako81:
This will probably also need changelog.

mrow4a:
@jnweiger let me also review this, just to make sure it wont introduce some more regressions that are not covered with enough tests.

pako81:
@mrow4a CI is now happy. Review it?

mrow4a:
@jnweiger @pako81 ok, so the logic must match both server side and client side:
https://github.com/owncloud/core/blob/master/core/js/sharedialoglinkshareview.js#L121-L130
https://github.com/owncloud/core/blob/master/lib/private/Share20/Manager.php#L221-L232

Also we need to make sure all cases are covered (both folder and file). I will add commits to this change.

jnweiger:
@mrow4a you can see in the intial post, 'How Has This Been Tested' what I tested.
Now when looking at [sharedialoglinkshareview.js#L121-L130](https://github.com/owncloud/core/blob/master/core/js/sharedialoglinkshareview.js#L121-L130) I am surprised that all this actually worked correctly... hmm.

mrow4a:
it works, because this checks what should NOT require password. I am also investigation whether something else should be considered.

mrow4a:
@jnweiger @pako81 so the logic that was there was too simple, also too little tests. It was somehow missed when adding more "public link roles". Now I list all cases possible and do not leave out "default". The same with tests

jnweiger:
@mrow4a Thank you! That is much cleaner now: Better and consistent naming, js and php "obviously" do the same logic now. All cases explicit (that always puzzled me). 
From reading code and comments, I think I really understand now, what is going on.
Now if I cannot break it by manual chaos monkey testing, then I am convinced it is correct.

jnweiger:
Tested all my combinations from the initial comment, and a few more.

We now have:
Password for File Public link with Download View Edit could be enforced by the setting [x] Enforce password protection for read + write + delete links

I'd argue that controlling this case with the setting [x] Enforce password protection for read + write links
is more logical: A public file link even when writable cannot do a delete at all. So, when I want to enforce passwords for a writable file link, I would not try anything that has 'delete' in it.

Everything else works as expected.

mrow4a:
@jnweiger so one detail, for the shares on a file, delete permission is always there. As you can delete a share which is basically unmount. I understand this is confusing, but it is as is.

jnweiger:
Ah, understand. Deleting the share is a permission. Well, that is not exactly deleting the file. But makes sense from an end user perspective.

pako81 (approved):

pako81:
@jnweiger @mrow4a can we have a changelog for this? Thx!

PR #40702, opened 2023-03-25
occ encryption:encrypt-all should not auto-enable user-key encryption

mrow4a:
## Description
With OC10.7 the user-key encyption has been deprecated. This PR makes it not possible to auto-enable user-key encryption when no module is selected when executing `occ encryption:encrypt-all`

## Related Issue
- Implements https://doc.owncloud.com/docs/next/server_release_notes.html#deprecation-note-for-user-key-storage-encryption
- Implements https://github.com/owncloud/enterprise/issues/4939

## Types of changes
- [x] Technical debt

## Checklist:
- [x] Code changes
- [x] Unit tests added
- [x] Acceptance tests added
- [x] Documentation ticket (not needed)

### DIFF
--- a/changelog/unreleased/40702
+++ b/changelog/unreleased/40702
@@ -1,0 +1,7 @@
+Change: Do not auto-enable user-key ecryption
+
+Executing occ encryption:encrypt-all would not auto-enable user-key encryption.
+
+https://github.com/owncloud/core/pull/40702
+https://github.com/owncloud/enterprise/issues/4939
+https://doc.owncloud.com/docs/next/server_release_notes.html#deprecation-note-for-user-key-storage-encryption
--- a/changelog/unreleased/40702
+++ b/changelog/unreleased/40702
@@ -1,6 +1,6 @@
 Change: Do not auto-enable user-key ecryption
 
-Executing occ encryption:encrypt-all would not auto-enable user-key encryption.
+Executing occ encryption:encrypt-all will no longer auto-enable user-key encryption.
 
 https://github.com/owncloud/core/pull/40702
 https://github.com/owncloud/enterprise/issues/4939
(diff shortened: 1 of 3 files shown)


pako81:
Makes sense - @mrow4a maybe this is also a good time to start thinking about removing user-key encryption as an option in the webUI.

pako81 (approved):

mrow4a:
> Makes sense - @mrow4a maybe this is also a good time to start thinking about removing user-key encryption as an option in the webUI.

yes, I work on bigger change in encryption app to make it work. but not yet finished.

jvillafanez:
I think it's better to point the user to what he should do instead of just saying "None of the encryption modules is enabled". You can also add something like "Set master encryption through the web UI" or something like that (not sure if it's possible through command line).

mrow4a:
> I think it's better to point the user to what he should do instead of just saying "None of the encryption modules is enabled". You can also add something like "Set master encryption through the web UI" or something like that (not sure if it's possible through command line).

it is core repo, this should not be aware of any encryption modules or even the apps that are used, right ?

changelog/unreleased/40702
```diff
@@ -0,0 +1,7 @@
+Change: Do not auto-enable user-key ecryption
+
+Executing occ encryption:encrypt-all would not auto-enable user-key encryption.
```
phil-davis, line 3:
```suggestion
Executing occ encryption:encrypt-all will no longer auto-enable user-key encryption.
```

phil-davis (approved):
LGTM - I suppose that the encryption app should be the place that tries to guide admins if they enable the app but have not yet chosen the master or user key encryption.

jvillafanez:
> it is core repo, this should not be aware of any encryption modules or even the apps that are used, right ? What I did here was to more maintain backwards compatibility with previous development here.

That's right. It shouldn't but now it is checking if user key or master key encryption is enabled. I'm ok if this is addressed in the future, otherwise the easiest solution is to adjust the error message (making the code module-agnostic will likely take a while)

DeepDiver1975 (approved):

DeepDiver1975:
related: https://github.com/owncloud/encryption/pull/389

new installations will only run master key encryption ...

PR #40511, opened 2022-11-23
[full-ci] Show Legal statement by personal settings

T0mWz:
## Description
When your own legal statements are published, handy to show them at your personal settings page.

## How Has This Been Tested?
 - Applied within own environment

## Screenshots (if appropriate):
Screenshot 2022-11-23 at 10 51 48

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Database schema changes (next release will require increase of minor version instead of patch)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Technical debt
- [ ] Tests only (no source changes)

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [ ] Documentation ticket raised: <link> 
- [ ] Changelog item, see [TEMPLATE](https://github.com/owncloud/core/blob/master/changelog/TEMPLATE)

## ToDo:
 - Translation option added

PR #40704, opened 2023-03-26
Fix changelog typo #40511

ho4ho:
Fix typo problem by changelog by #40511

https://owncloud.com/changelog/server#10.12.0 also saying release by `2023-02-24` but release more later?:

https://central.owncloud.org/t/owncloud-server-10-12-released-persistent-versions-introduced/41598


phil-davis (approved):
I think that the changelog-generation tools will take this fix an apply it correctly to the changelog etc.

Issue #40703, opened 2023-03-26
32 bit Server: Upload of … exceeds the quota for the folder, when server has plenty of remaining storage


sgparry:
** I have tried my best to format this bug report as requested - if I have messed up anywhere please let me know in the comments; I posted a request for help on central and got no feedback at all **

I have been trying to solve this for several days. After extensive housekeeping, My VPS has 17GB out of 74GB still free, but ownCloud still refuses to sync. I suspect it is something to do with the 32 bitness of my server or the external storage.

### Expected behaviour
Files should be synchronized to server without error

### Actual behaviour
Client complains "There is insufficient space available on the server" and shows "Upload of ... exceeds the quota for the folder" in not synced tab.  Running ```occ files:scan --all``` and ```occ trashbin:cleanup``` makes no difference, nor does retry sync in the client. The error seems to be a conspiracy between the two sides, client and server. The web server logs have shown genuine 507 errors from remote.php, typically on MOVE commands, but also the client seems to retain any error status permanently.

### Steps to reproduce
1. Ensure the server has the following space usage for target user, and is using some in-built storage and some external local storage on the same drive as owncloud's data directory:
```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        74G   55G   17G  78% /

$ du -d0 -h /var/www/owncloud/data/
17G     /var/www/owncloud/data/
# External local storage 1
$ du -d0 -h /srv/moodle/**redacted**/vle/moodledata/repository/ 
2.6G    /srv/moodle/**redacted**/vle/moodledata/repository/

# External local storage 2
$ du -d0 -h /srv/www/**redacted**/html
124M    /srv/www/**redacted**/html
```
from web interface:
You are using 14.1 GB of 30.1 GB (46.79 %)

2. Install client with non-virtual file mode and selected files.
3. Connect client to user area - choose about 50 - 60 % of the folders to sync.
4. Configure client to synchronize hidden files. Turn off both "Ask for..." confirmations
5. Allow files to sync (if possible).
6. Add one file of 13,496,320 bytes called "%HOME%\ownCloud\pyspace repository\CA-GCE-COMPSCI\assessments\2022-2023\Year13\AP4-Mock-2\paper-2-raw\2.dat".

### Server configuration
**Operating system:** Ubuntu 18.04.6 LTS (32 bit)

**Web server:** apache2  2.4.29-1ubuntu4.27

**Database:** MySQL 5.7.41

**PHP version:** 7.4.33

[...]

Issue #40695, opened 2023-03-22
507 Insufficient Storage when adding a file or syncing


Jeansen:
### Steps to reproduce
1. Put a file on the server, e.g. pdf, 400KB
2.
3.

### Expected behaviour
File should be accepted, synced

### Actual behaviour
I get a message telling me no space left on the server. Happens in any client, browser, application or app (iOS)

### Server configuration
**Operating system**: 
Debian 11

**Web server:** 
Apache2

**Database:** Mysql 15

**PHP version:**
PHP 7.4.33 (cli) (built: Nov  8 2022 11:40:37) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies  with Zend OPcache v7.4.33, Copyright (c), by Zend Technologies

**ownCloud version:** (see ownCloud admin page) 10.12.0.6

**Updated from an older ownCloud or fresh install:** update

**Where did you install ownCloud from:** owncloud repository
deb http://download.opensuse.org/repositories/isv:/ownCloud:/server:/10/Debian_11/ /

**Signing status (ownCloud 9.0 and above):**

```
No errors have been found.
```

**The content of config/config.php:**

```
{
    "system": {
        "updatechecker": false,
        "instanceid": "oce9dczbeay4",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "192.168.178.35",
            "cloud"
        ],
        "datadirectory": "\/var\/www\/owncloud\/data",
        "overwrite.cli.url": "https:\/\/192.168.178.42\/owncloud",
        "dbtype": "mysql",
        "version": "10.12.0.6",
        "dbname": "owncloud",
        "dbhost": "localhost:3306",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "logtimezone": "Europe\/Berlin",
        "installed": true,
        "memcache.local": "\\OC\\Memcache\\APCu",
        "memcache.locking": "\\OC\\Memcache\\Redis",
        "redis": {
            "host": "localhost",
            "port": 6379
        },
        "mail_smtpmode": "php",
        "maintenance": false,
        "theme": "",
        "loglevel": 3,
        "log_rotate_size": 104857600,
        "singleuser": false,
[...]

PR #40709, opened 2023-03-28
Prevent 507 Insufficient Storage on 32-bit systems

pako81:
## Description
Prevent 507 Insufficient Storage on 32-bit systems

## Related Issue
- https://github.com/owncloud/core/issues/40695
- https://github.com/owncloud/enterprise/issues/5654

## Motivation and Context
With the introduction of https://github.com/owncloud/core/commit/c3687864a3f820445be2a5ef9db543d4fb6e7ed9 we apparently broke compatibility to 32-bit systems as we are now casting `$freeSpace` to int and this is causing an integer overflow on such systems when the free space is above the max supported value. We added therefore an additional check for 32-bit systems in QuotaPlugin.php.

## How Has This Been Tested?
- Feedback from users reporting this in https://github.com/owncloud/core/issues/40695

## Types of changes
- [ ] Technical debt

## Checklist:
- [X] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added
- [X] Changelog item

### DIFF
--- a/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
@@ -204,6 +204,17 @@ class QuotaPlugin extends \Sabre\DAV\ServerPlugin {
 				$path = \rtrim($parentPath, '/') . '/' . $info['name'];
 			}
 			$freeSpace = $this->getFreeSpace($path);
+			// workaround to guarantee compatibility on 32-bit systems as otherwise this would cause an int overflow on such systems
+			// in case $freeSpace is above the max supported value. $freeSpace should be a float so we are using the <= 0.0 comparison
+			if (PHP_INT_SIZE === 4) {
+				$availableSpace = $freeSpace + $extraSpace;
+				if ($freeSpace !== FileInfo::SPACE_UNKNOWN && $freeSpace !== FileInfo::SPACE_UNLIMITED && (($length > $availableSpace) || ($availableSpace <= 0.0))) {
+					if (isset($chunkHandler)) {
+						$chunkHandler->cleanup();
+					}
+					throw new InsufficientStorage();
+				}
+			}
 			// freeSpace might be false, or an int. Anyway, make sure that availableSpace will be an int.
 			$availableSpace = (int) $freeSpace + $extraSpace;
 			if ($freeSpace !== FileInfo::SPACE_UNKNOWN && $freeSpace !== FileInfo::SPACE_UNLIMITED && (($length > $availableSpace) || ($availableSpace === 0))) {
--- a/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
@@ -214,14 +214,15 @@ class QuotaPlugin extends \Sabre\DAV\ServerPlugin {
 					}
 					throw new InsufficientStorage();
 				}
-			}
-			// freeSpace might be false, or an int. Anyway, make sure that availableSpace will be an int.
-			$availableSpace = (int) $freeSpace + $extraSpace;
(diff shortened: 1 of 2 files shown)


heppel:
Can't test this now. But I think the original code with the (int) cast should be in the "else" clause of the PHP_INT_SIZE test or the overflow and the incorrect error will still happen.

apps/dav/lib/Connector/Sabre/QuotaPlugin.php
```diff
@@ -204,6 +204,17 @@ public function checkQuota($path, $length = null, $extraSpace = 0) {
...
+			// in case $freeSpace is above the max supported value. $freeSpace should be a float so we are using the <= 0.0 comparison
+			if (PHP_INT_SIZE === 4) {
+				$availableSpace = $freeSpace + $extraSpace;
+				if ($freeSpace !== FileInfo::SPACE_UNKNOWN && $freeSpace !== FileInfo::SPACE_UNLIMITED && (($length > $availableSpace) || ($availableSpace <= 0.0))) {
+					if (isset($chunkHandler)) {
+						$chunkHandler->cleanup();
+					}
+					throw new InsufficientStorage();
+				}
+			}
```
jvillafanez, line 217: Better include an `else` just in case.
If `PHP_INT_SIZE === 4` but it doesn't trigger the insufficient storage, the code will still go through the 64 bit check.

phil-davis (approved):
LGTM now. There is code duplication, but minimizing that might obfuscate the separate 32-bit and not-32-bit code blocks.

jvillafanez (approved):

jnweiger (commented):
I am not 100% convinced, all edge cases are covered. But as the immediate problem of the user is nicely fixed. I am happy with the solution too.

jvillafanez:
> I am not 100% convinced, all edge cases are covered.

As in "there could be more issues caused by integer overflow in 32 bits"?. Yeah, probably. If it isn't officially supported and we aren't testing with 32 bits, there could be issues lurking around.
For this particular issue with the quota, I don't think we're missing anything.

ho4ho:
Also fix #40703

Valentin-Bleuse:
I changed apps/dav/lib/Connector/Sabre/QuotaPlugin.php still having this issue : there is insufficient space available on the server for some upload. Nothing have changed even with your new code. No idea why it doesn't work properly.

jvillafanez:
If you're using a 32 bit system, you might still be limited to uploads of less than 2GB. I'm not sure about the exact behavior, but there could be issues.
Other than that, I guess you have a regular "not enough quota" error, so check how much quota your user has and ensure the file fits in it.

jvillafanez:
@pako81 we might need a different fix. It seems that the `$extraSpace` variable can also be a float.
If we're copying to an already existing file, the `$extraSpace` will be the size of that file. Doctrine should return the size of the file as string in https://github.com/owncloud/core/blob/master/lib/private/Files/Cache/Cache.php#L162 and `0 + <string>` could return a float

```
php > var_dump(0 + "12378974983729847892798758974897389578934759876978");
float(1.237897498373E+49)
php > var_dump(0 + "345");
int(345)
```
That's "consistency" in PHP....

I think the reproduction steps should be:
1. Upload a 3GB file to ownCloud
2. Copy any file into the 3GB file in order to replace it.

Maybe it works for 64 bits because the bigint coming from the DB fits in a 64 bits integer

I think we'll have to keep the "32 bits block" and remove the other one. Need to check all the places where the `getSize` method says it returns an integer...
For now, we'll have to assume that both `$freeSpace` and `$extraSpace` can be floats.

Valentin-Bleuse:
It's not from the quotas for sure and even when I put a single photo, it told me I don't have any space left

jvillafanez:
@Valentin-Bleuse we might need a new ticket for that. I'd recommend you to try to reproduce the issue either in a fresh environment or in https://demo.owncloud.com . It might be a different issue, and without steps to reproduce we can't fix the problem.

In addition, make sure you have enough space everywhere:
* The web server uses the "/tmp" folder (or whatever temporary directory is configured), and that folder might be limited.
* There could be also quota limitations for the user configured by the ownCloud's admin.
* Your host might not have enough free space

PR #40726, opened 2023-04-04
fix: do not accept federated shares where the name is too long

DeepDiver1975:
## Description
Federated shares with a too long name can result in in accessible shares on the receiving server

## Related Issue
- Fixes <issue_link>

## How Has This Been Tested?

- test environment:
- test case 1:
- test case 2:
- ...

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
+++ b/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
@@ -26,6 +26,7 @@
 
 namespace OCA\FederatedFileSharing\Controller;
 
+use OC\Files\View;
 use OC\OCS\Result;
 use OCA\FederatedFileSharing\Address;
 use OCA\FederatedFileSharing\AddressHandler;
@@ -37,6 +38,7 @@ use OCA\FederatedFileSharing\Ocm\Exception\NotImplementedException;
 use OCA\FederatedFileSharing\Ocm\Exception\OcmException;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\OCSController;
+use OCP\Files\InvalidPathException;
 use OCP\IRequest;
 use OCP\IUserManager;
 
@@ -120,7 +122,7 @@ class RequestHandlerController extends OCSController {
 				]
 			);
 
-			if (!\OCP\Util::isValidFileName($name)) {
+			if (!$this->isFileNameValid($name)) {
 				throw new BadRequestException(
 					'The mountpoint name contains invalid characters.'
 				);
@@ -405,5 +407,19 @@ class RequestHandlerController extends OCSController {
 		}
 
 		return new Result();
+	}
+
+	private function isFileNameValid(?string $name): bool {
+		if ($name === null) {
+			return false;
+		}
+		$v = new View();
+		try {
+			# new shares will show up in user home - therefore we test with /
+			$v->verifyPath('/', $name);
+		} catch (InvalidPathException $e) {
+			return false;
+		}
+		return true;
 	}
 }
--- a/apps/federatedfilesharing/tests/Controller/RequestHandlerTest.php
+++ b/apps/federatedfilesharing/tests/Controller/RequestHandlerTest.php
@@ -42,6 +42,7 @@ use OCP\IRequest;
 use OCP\IUserManager;
 use OCP\Share;
 use OCP\Share\IShare;
+use PHPUnit\Framework\MockObject\MockObject;
 
 /**
  * Class RequestHandlerControllerTest
@@ -53,7 +54,7 @@ class RequestHandlerTest extends TestCase {
 	public const DEFAULT_TOKEN = 'abc';
 
 	/**
-	 * @var IRequest | \PHPUnit\Framework\MockObject\MockObject
+	 * @var IRequest | MockObject
 	 */
 	private $request;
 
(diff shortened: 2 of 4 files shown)


apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -406,4 +408,17 @@ public function updatePermissions($id) {
...
 		return new Result();
 	}
+
+	private function isFileNameValid(?string $name): bool {
+		if ($name === null) {
+			return false;
+		}
+		$v = new View();
+		try {
+			$v->verifyPath('/', $name);
```
jvillafanez, line 418: It might need some comment.
I assume that, even if `$v->verifyPath('/', $name)` passes, `$v->verifyPath('/abc/cde', $name)` might throw an exception. Note that the name is the same but we're checking against a different folder. It might also depend on the view (if the root is "/users/files", for example).

I guess this could be fine because the share will be placed there, but we should explain this in a comment.

DeepDiver1975:
@jvillafanez please review again - thx

jvillafanez (approved):

phil-davis:
@jnweiger @pako81 is this something that is wanted for 10.12.1 release?
If so, then you will need to cherry-pick it into `release-10.12.1` branch.

saw-jan:
I don't know how to confirm this.
Steps that I tried:
1. server1, mount an external storage (SMB) - since creating long name file is not possible from server UI
2. create a file with long name in the SMB share folder
`aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFo.txt`
3. server1, do a federate share of that file to another server (10.13.0) - admin@\<latest-server-url\>
4. server2, accept the share - (accept works)
5. server2, open the file - (content is there)

CC @DeepDiver1975

jnweiger:
Not exaclty sure, where the length limit is. drone CI uses a 290 char string to trigger this. Yours is shorter: 233 chars.

pako81:
The `verifyPath` method in View.php is used https://github.com/owncloud/core/blob/master/lib/private/Files/View.php#L1955 which is then calling https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Common.php#L505-L511 where a `FileNameTooLongException` exception is triggered if the path is longer than 255 chars

jesmrec:
> where a FileNameTooLongException exception is thrown if the path is longer than 255 chars.

but, clients (including web) do not allow such amount of characters in the filename. So, that exception is never thrown.

jnweiger:
> The `verifyPath` method in View.php is used:
> https://github.com/owncloud/core/blob/master/lib/private/Files/View.php#L1921> 
> which is then calling:
> https://github.com/owncloud/core/blob/master/lib/private/Files/Storage/Common.php#L505-L511
> where a `FileNameTooLongException` exception is thrown if the path is longer than 255 chars.

Ouch. This will explode, when versioning or trashcan add their suffixes to the filename.
```
./files_versions/Documents/LoremIpsum.txt.v1692195415
./files_trashbin/files/Diagramm.drawio.d1692184476
```

pako81:
> Ouch. This will explode, when versioning or trashcan add their suffixes to the filename.
./files_versions/Documents/LoremIpsum.txt.v1692195415
./files_trashbin/files/Diagramm.drawio.d1692184476

not related to this specific issue but, yes, we need to check what happens with occ `files:scan` when scanning version and trashbin paths longer than 255 chars.

jnweiger:
The code path with these checks is not executed. 
Patched

```
        public function createShare() {
                \OCP\Util::writeLog('federatedfilesharing', 'JW was here createShare ', \OCP\Util::ERROR);
                throw new BadRequestException('JW was here. createShare');
                try {

...

       public function acceptShare($id) {
                \OCP\Util::writeLog('federatedfilesharing', 'JW was here acceptShare ', \OCP\Util::ERROR);
                throw new BadRequestException('JW was here. acceptShare');
                try {

...
```

* Incoming federated shares can still be accepted.

DeepDiver1975:
> I don't know how to confirm this.

In a regular use case this will not happen because files on all servers have the correct file name length.

This is a safety net against wrong usage of the http request. Submit an http request using curl.

saw-jan:
Tried to re-check:
1. Created file `aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBr.txt` (80 chars)
2. Edited here to have only `55` https://github.com/owncloud/core/blob/71e0ebe2e922ade003be5a0cfb2a5cfe15820986/lib/private/Files/Storage/Common.php#L505-L511
3. Create a fed share
    ```bash
    curl -XPOST "http://localhost/oc10/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json" \
    -d"shareType=6&shareWith=admin@localhost/core&permissions=19&path=/aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBrownFoxJumpsOverAVeryLazyDog-aQuickBraisdiusadsagdiasdasd.txt" \
    -uadmin:admin
    ```
4. Share is created (:exclamation:)

NOTE: The share request doesn't hit the `createShare` controller in https://github.com/owncloud/core/blob/71e0ebe2e922ade003be5a0cfb2a5cfe15820986/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php#L99-L100
but hits https://github.com/owncloud/core/blob/71e0ebe2e922ade003be5a0cfb2a5cfe15820986/apps/files_sharing/lib/Controller/Share20OcsController.php#L372-L373

DeepDiver1975:
> ```shell
> curl -XPOST "http://localhost/oc10/ocs/v1.php/apps/files_sharing/api/v1/shares?format=json" \
> ```

federatedfilesharing is the app -> not files_sharing ....

DeepDiver1975:
> 4\. Share is created (❗)

worth an issue for further investigation

PR #40727, opened 2023-04-05
Fix quota for 32 bits. Some vars could be float instead of integer

jvillafanez:
## Description
The type of some variables seems to be different than the ones we expect based on some conditions.
In particular, 32 bits systems seem to use floats instead of integers in some of our "getSize" methods, which was causing problems because we expected only integers. Quota calculation was affected by this.

The problem with the quota is fixed with this PR also for 32 bits systems because we've taken into consideration that the size could be either an integer or a float.
Most of the "getSize" methods have been adjusted so the PHPDoc shows the right information. The behavior of the methods hasn't been touched (no new issues should happen)

## Related Issue
https://github.com/owncloud/core/pull/40709

## How Has This Been Tested?
Not tested with 32 bits but with 64 bits. I don't see any problem

Test1
1. Create a user with 0B quota
2. The user can't create a folder.
3. The user can't upload files.

Test2
1. Create a user with 1.5GB quota
2. Upload a 1GB file. The upload works
3. Upload another 1GB file. Error shows not enough free space.

Test3
1. Create a user with 2.5GB quota
2. Upload a 1GB file to "/file.bin"
3. Upload another 1GB file to "/t1/file.bin"
4. Copy "/file.bin" to "/zzz/file.bin". Error shows not enough quota

Test4
1. Create a user with 2.5GB quota
2. Upload a 1GB file to "/file.bin"
3. Upload another 1GB file to "/t1/file.bin"
4. Copy "/file.bin" into "/t1/file.bin". This is successful because the file is overwritten.

## Types of changes

- [ ] Technical debt

## Checklist:

- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Node.php
+++ b/apps/dav/lib/Connector/Sabre/Node.php
@@ -220,7 +220,7 @@ abstract class Node implements \Sabre\DAV\INode {
 	 * size because the actual size value isn't returned by google. This function
 	 * will return null in this case)
 	 *
-	 * @return integer|null
+	 * @return integer|float|null
 	 */
 	public function getSize() {
 		$size = $this->info->getSize();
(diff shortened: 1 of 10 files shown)


jvillafanez:
Closing in favor of https://github.com/owncloud/core/pull/40729 , which targets the release branch.

PR #40730, opened 2023-04-05
Do not correct parent folders if the target isn't fully scanned

jvillafanez:
## Description
If the folder we want to correct the size isn't fully scanned, there is no need to try to correct the parent folders: we can't get a proper size so they should stay as "need to scan" or "shallow scanned".

This should improve the performance a bit when the system isn't fully scanned, specially in deep folders, because we won't try to update all the parent folders unless it's needed.

## Related Issue
https://github.com/owncloud/enterprise/issues/5605

## Motivation and Context
This should reduce the DB load in some scenarios because there will be less queries to be done.

## How Has This Been Tested?
1. Prepare an external storage (SFTP for example) with "/f1/f2/f3/f4/file.txt"
2. Connect ownCloud to the external storage
3. Check that the "/SFTP" folder shows "pending" as size.
4. Traverse the directories until you reach "/SFTP/f1/f2/f3/f4/file.txt"
5. From there, traverse the directories backwards. Old "pending" sizes for the folders should show now the right size (assuming there are no additional folders you haven't traversed, such as "/SFTP/f1/f2/side_quest/")

## Types of changes

- [x] Technical debt

## Checklist:

- [x] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -849,7 +849,14 @@ class Cache implements ICache {
 	 * @param array $data (optional) meta data of the folder
 	 */
 	public function correctFolderSize($path, $data = null) {
-		$this->calculateFolderSize($path, $data);
+		$newSize = $this->calculateFolderSize($path, $data);
+		if ($newSize < 0) {
+			// newSize < 0 implies the folder needs a scan or it's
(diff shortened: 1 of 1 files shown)


pako81 (approved):

pako81:
@jvillafanez merge this?

PR #40739, opened 2023-04-17
Prepare merge back of the release branch to master .

jnweiger:
We undo
1) a .drone.star hack that was needed tp make the release-branch pass ci.
2) revert of 40702 and 40712, which we did not want to have in the release, but want to keep them in master.

@phil-davis Not sure, if there a better way to do this?
Last time you did some cherry picking for a merge back branch, that also lead to confusion...

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -285,7 +285,7 @@ config = {
                 "cliEncryption",
             ],
             "extraApps": {
-                "encryption": "git checkout v1.5.3; composer install",
+                "encryption": "composer install",
             },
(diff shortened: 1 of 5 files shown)


phil-davis:
> Last time you did some cherry picking for a merge back branch, that also lead to confusion...

I agree. I quickly tried some cherry-pick locally and it is a real pain.

phil-davis (approved):
Seems reasonable if CI passes.

(then we will see what a merge-back PR to master does!)

PR #40747, opened 2023-04-19
[tests-only] [full-ci] Run some CI pipelines with Ubuntu 22

phil-davis:
## Description
We support oC10.12 with PHP 7.4 running on Ubuntu 20.04 or 22.04. We have been running all the tests (unit, litmus, acceptance...) with Ubuntu 20.04. We also want to know that there are not regressions when running on Ubuntu 22.04. There is an `owncloud-ci/php` docker image that is based on Ubuntu 22.04 with the necessary things done to install PHP 7.4 (rather than PHP 8), and to set up openssl so that legacy encryption methods are enabled. These are the things that also have to be done on 
a production Ubuntu 22.04 system to run oC10.12.

This PR adds some test pipelines to CI that run on Ubuntu 22.04:
- PHP unit tests with sqlite and with Maria DB (note that the existing scripts automagically run all tests for the sqlite case and only the tests tagged with group DB in the MariaDB case, so that is why I have added both pipelines)
- cliEncryption test suite (because that verifies that encryption+openssl are working OK)
- API acceptance smoke tests

I think that is enough test pipelines to run to help notice any regressions due to the Ubuntu 22.04 settings.

## How Has This Been Tested?
CI

## Types of changes
- [ ] Technical debt

## Checklist:
- [ ] Code changes
- [ ] Unit tests added
- [ ] Acceptance tests added

### DIFF
--- a/.drone.star
+++ b/.drone.star
@@ -91,6 +91,18 @@ config = {
                 "oracle",
             ],
         },
+        "ubuntu22": {
+            "phpVersions": [
+                "7.4-ubuntu22.04",
+            ],
+            # These pipelines are run just to help avoid any obscure regression
+            # on Ubuntu 22.04. We do not need coverage for this.
+            "coverage": False,
+            "databases": [
+                "sqlite",
+                "mariadb:10.6",
+            ],
+        },
         "external-samba": {
             "phpVersions": [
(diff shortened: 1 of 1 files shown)


SwikritiT (approved):
LGTM

